PR #47019, opened 2026-06-18
[dom-webview][ios] Fix DOM component props containing double quotes

### LINKED ISSUE #47016: @expo/dom-webview (iOS) corrupts DOM-component props containing a double quote, blanking the component
> [!NOTE]
> Yes, AI wrote the title and description, but there's a real person who instructed AI to clearly do a lay of the land for this quotes bug. It ended up surfacing itself because we were passing in a font face string in as a prop and then the webview borked, threw an error, and then failed to render. 
> I love all the stuff y'all (Expo) are doing with RN and Expo DOM components. This has been a big unlock for us at YouVersion (I am a subcontractor, not an employee) working on our React Native Expo SDK (to be released soon). 

https://github.com/cameronapak/expo-dom-webview-ios-quote-repro

```bash
npm install
npm run ios       # BLANK; Metro logs "Top OS ($$EXPO_DOM_HOST_OS) is not defined"
npm run android   # renders correctly
```

A blank-template Expo SDK 56 app with one `'use dom'` component that receives a single string prop containing a `"`.

## Summary

On iOS, a DOM component (`'use dom'`) renders **blank** when any prop value contains a double quote. The visible error is misleading:

```
Web ERROR [Error: Top OS ($$EXPO_DOM_HOST_OS) is not defined. This is a bug in the DOM Component runtime.]
    registerDOMComponent@.../node_modules/expo/dom/entry.bundle
```

It's a symptom: `@expo/dom-webview`'s iOS injection produces invalid JSON, so the DOM HTML wrapper's `JSON.parse(...)` throws before it can set `window.$$EXPO_DOM_HOST_OS`. Android is unaffected. Regressed in SDK 56, which made `@expo/dom-webview` the default backing WebView (`react-native-webview` escaped the payload safely).

## Root cause

`ios/DomWebView.swift` → `setInjectedJavaScriptObject` embeds the props JSON in a JS **template literal**:

```swift
window.ReactNativeWebView.injectedObjectJson = function () {
  return `\(source)`;   // source = JSON.stringify(injectedJavaScriptObject)
}
```

`JSON.stringify` escapes `"` → `\"`, but the template literal un-escapes `\"` → `"` when the function runs, so `injectedObjectJson()` returns malformed JSON (e.g. `..."fontFamily":""Source Serif 4", serif"...`). `JSON.parse` throws, the `window.$$EXPO_*` globals are never set, and `registerDOMComponent` throws. The same hazard applies to backtick and `${` in any prop value. Android returns the raw string via a real `@JavascriptInterface` (`RNCWebViewBridge.injectedObjectJson()`), so it's unaffected.

## Expected behavior

The prop crosses the bridge intact (as on Android) and the component renders.

## Suggested fix

Don't interpolate `source` into a template literal — base64-encode it (UTF-8 safe) or escape `` ` ``, `\`, and `$`. Base64 is the more robust option.

## Environment

- `@expo/dom-webview` 56.0.5 (current `latest`) · `expo` 56.0.12 · `react-native` 0.85.3 · `react` 19.2
- iOS (WKWebView), Simulator + physical device · dev build (not Expo Go)
- Likely introduced in #43155 (added `injectedJavaScriptObject` support). Distinct from #46374 (size-dependent content-process kill, also hits `react-native-webview`). Supersedes #47014 (filed without an inline repro).

intergalacticspacehighway:
Fixes #47016 

# Why?

Here's a minimal repro to test it in browser console.

```
https://github.com/expo/expo/blob/9be13fd21712607e1d6b6031c4df7b2dbc258952/packages/%40expo/dom-webview/src/DomWebView.tsx#L44
a = JSON.stringify({ font: '"Inter"' });  

// https://github.com/expo/expo/blob/9be13fd21712607e1d6b6031c4df7b2dbc258952/packages/%40expo/dom-webview/ios/DomWebView.swift#L228
b = new Function('return `' + source + '`;');    

// https://github.com/expo/expo/blob/9be13fd21712607e1d6b6031c4df7b2dbc258952/packages/%40expo/cli/src/start/server/middleware/DomComponentsMiddleware.ts#L122
JSON.parse(b())
// throws error
```

## How?

Add `JSON.stringify` to function returned object so it stays compatible with `JSON.parse`.

## Test Plan

Tested the original user repro.

### DIFF
--- a/packages/@expo/dom-webview/ios/DomWebView.swift
+++ b/packages/@expo/dom-webview/ios/DomWebView.swift
@@ -207,7 +207,7 @@ internal final class DomWebView: ExpoView, UIScrollViewDelegate, WKUIDelegate, W
       let script = """
       window.ReactNativeWebView = window.ReactNativeWebView || {};
       window.ReactNativeWebView.injectedObjectJson = function () {
-        return `\(source)`;
+        return JSON.stringify(\(source));
       }
       true;
       """


packages/@expo/dom-webview/ios/DomWebView.swift
```diff
@@ -225,7 +207,7 @@ internal final class DomWebView: ExpoView, UIScrollViewDelegate, WKUIDelegate, W
       let script = """
       window.ReactNativeWebView = window.ReactNativeWebView || {};
       window.ReactNativeWebView.injectedObjectJson = function () {
-        return `\(source)`;
+        return JSON.stringify(\(source));
```
intergalacticspacehighway, line 210: Here the `source` gets injected as literal so JS engine parses it as an object, `JSON.stringify` converts it back to string.

intergalacticspacehighway: alternate fix would be to remove `JSON.stringify` from here and `JSON.parse` from `DomComponentsMiddleware.ts` but then it won't work for Android, because android returns a string in `RNCWebViewBridge.kt` `injectedObjectJson`

Kudo (approved):
that's a correct fix. thanks for helping on this

PR #47032, opened 2026-06-18
chore(precompile) change gz -> xz compression to reduce size

chrfalch:
# Why

We distribute our XCFrameworks as tar.gz files today, and include them in our npm packages for easy downloads and caching.

To reduce size we have decided to change the compression type to xz since it reduces the size of the packages and makes downloads smaller.

Closes #ENG-21969

**Size reduction:**

  |                | Codec          | Archive size | Reduction |
  | -------------- | -------------- | ------------ | --------- |
  | Before this PR | gzip (level 6) | 14.82 MB     | —         |
  | After this PR  | xz (-9 -T0)    | 5.56 MB      | ~62%      |

# How

- Change how precompiled_modules.rb reads packages - now looking for.xz files.
- Updated the debug/release switch scripts to look for .xz files as well
- Update the prebuild-external-xcframeworks workflow to look for xz files as well

Updated prebuilds tools:
- Updated Frameworks.ts to generate xz compressed packages
- Updated tests
- Updated bundleIOSPrebuilds.ts to test for .xz files.

## Test-plan

✅ Rebuild local XCFrameworks and verify they're packaged using .xz files 
✅ Run pod install on BareExpo and verify it is using .xz files 
✅ Run npm packager to ensure it includes the correct files
✅ expo-brownfield build:ios --release
✅ generated Swift Package

# Checklist

### DIFF
--- a/packages/expo-modules-autolinking/scripts/ios/precompiled_modules.rb
+++ b/packages/expo-modules-autolinking/scripts/ios/precompiled_modules.rb
@@ -8,7 +8,7 @@
 # When EXPO_USE_PRECOMPILED_MODULES=1 is set, packages with matching XCFrameworks
 # in the centralized build output will be linked as vendored frameworks.
 #
-# Build output:     packages/precompile/.build/<pkg>/output/<flavor>/xcframeworks/<Product>.tar.gz
+# Build output:     packages/precompile/.build/<pkg>/output/<flavor>/xcframeworks/<Product>.tar.xz
 # Custom path:      Set EXPO_PRECOMPILED_MODULES_PATH to override the base directory
 #                   (replaces packages/precompile/.build/), keeping <pkg>/output/... structure
 #
@@ -22,8 +22,8 @@
 #
 # Resulting layout in Pods/<PodName>/:
 #   <Product>.xcframework/                    (extracted by CocoaPods from source tarball)
-#   artifacts/<Product>-debug.tar.gz          (copied by prepare_command)
-#   artifacts/<Product>-release.tar.gz        (copied by prepare_command)
+#   artifacts/<Product>-debug.tar.xz          (copied by prepare_command)
+#   artifacts/<Product>-release.tar.xz        (copied by prepare_command)
 #   artifacts/.last_build_configuration       (written by prepare_command / switch script)
 
 require 'fileutils'
@@ -97,7 +97,7 @@ module Expo
       # Returns custom base path for precompiled module output, if set.
       # When set, replaces the default `packages/precompile/.build/` base directory.
       # The directory structure under this path must match:
-      #   <npm_package>/output/<flavor>/xcframeworks/<Product>.tar.gz
+      #   <npm_package>/output/<flavor>/xcframeworks/<Product>.tar.xz
       def custom_modules_path
         ENV[MODULES_PATH_ENV_VAR]
       end
@@ -773,7 +773,7 @@ module Expo
           # Copy both flavor tarballs
           ['debug', 'release'].each do |flavor|
             src = resolve_prebuilt_tarball(info, product_name, flavor, pod_name)
-            dst = File.join(artifacts_dir, "#{product_name}-#{flavor}.tar.gz")
+            dst = File.join(artifacts_dir, "#{product_name}-#{flavor}.tar.xz")
             FileUtils.cp(src, dst) if File.exist?(src) && !File.exist?(dst)
           end
 
@@ -787,7 +787,7 @@ module Expo
             tarball = resolve_prebuilt_tarball(info, product_name, build_flavor, pod_name)
             if File.exist?(tarball)
               Pod::UI.info "#{'[Expo-precompiled] '.blue}Extracting #{product_name}.xcframework (cache miss)"
-              system("tar", "xzf", tarball, "-C", pod_dir)
+              system("tar", "xf", tarball, "-C", pod_dir)
             end
           end
         end
@@ -1274,10 +1274,10 @@ module Expo
         <<~SH
           # Self-healing: extract xcframework from tarball if CocoaPods cache was stale/empty
           if [ ! -d "#{product_name}.xcframework" ]; then
-            TARBALL="#{build_output_dir}/#{build_flavor}/xcframeworks/#{product_name}.tar.gz"
+            TARBALL="#{build_output_dir}/#{build_flavor}/xcframeworks/#{product_name}.tar.xz"
             if [ -f "$TARBALL" ]; then
               echo "[Expo XCFramework] #{product_name}: Extracting xcframework from build output (cache miss)"
-              tar xzf "$TARBALL"
+              tar xf "$TARBALL"
             fi
           fi
         SH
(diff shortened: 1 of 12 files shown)


kitten:
just a general question, should we choose a common denominator that Node.js also supports natively? This may not seem super relevant right now, but forking into a separate process may become something we can optimise for in the future, or if we restructure, would `xz` cause a few challenges here?

Natively, zlib in Node, as far as I remember, supports Brotli and zstd too, if those are valid alternatives here, by any chance

chrfalch:
> just a general question, should we choose a common denominator that Node.js also supports natively? This may not seem super relevant right now, but forking into a separate process may become something we can optimise for in the future, or if we restructure, would `xz` cause a few challenges here?

Node compresses the package, but Ruby is decompressing them and supports `xz` natively. Neither Brotli nor zstd is supported by ruby in the stdlib.

tools/src/prebuilds/Frameworks.ts
```diff
@@ -631,6 +632,93 @@ const copySPMDependencyXCFrameworksAsync = async (
...
+        new Error(
+          `Failed to run xz while compressing ${path.basename(tarballPath)}. The xz binary is ` +
+            `required to build precompiled XCFrameworks; install it (e.g. \`brew install xz\`) and ` +
+            `retry. Underlying error: ${e.message}`
+        )
+      )
+    );
+    out.on('error', fail);
+
+    tar.stdout.pipe(xz.stdin);
```
tsapeta: [💡 suggestion] `tar.stdout.pipe(xz.stdin)` has no error handler on `xz.stdin`. If `xz` exits or is killed before `tar` finishes (disk full, OOM-kill, SIGPIPE), `tar` keeps writing into a closed stdin and Node emits an `'error'` (EPIPE) event on `xz.stdin`. With no listener, an unhandled stream `'error'` throws as an uncaught exception and takes down the entire `et prebuild` process with a raw `EPIPE` stack trace — defeating the carefully-worded error messages this function otherwise produces.

I reproduced this: piping a fast producer into a child that exits immediately, with no handler on the writable side, yields `UNCAUGHT: EPIPE write EPIPE` and a non-zero crash.

Suggest adding `xz.stdin.on('error', fail)` (and arguably `tar.stdout.on('error', fail)`) so a downstream death is funneled through `fail()` like the other failure modes.

tools/src/prebuilds/Frameworks.ts
```diff
@@ -631,6 +632,93 @@ const copySPMDependencyXCFrameworksAsync = async (
...
+    const xz = spawn('xz', ['-9', '-T0', '-c'], { stdio: ['pipe', 'pipe', 'pipe'] });
+    const out = fs.createWriteStream(tarballPath);
+
+    let stderr = '';
+    let tarDone = false;
+    let xzDone = false;
+    let outDone = false;
+    let settled = false;
+
+    const fail = (error: Error) => {
```
tsapeta: [💡 suggestion] On failure `fail()` kills both processes and rejects, but leaves the partially-written `tarballPath` on disk. Since archive reads here go through `tar -xf` with no integrity/size check, a truncated `.tar.xz` left behind by a crashed run could later be treated as a valid cache hit (e.g. `resolve_prebuilt_tarball` returns it because `File.exist?` is true). Consider unlinking the partial output in `fail()` (best-effort `fs.unlink(tarballPath, () => {})`), or writing to a temp path and renaming on success so only complete archives ever appear at the final path.

packages/expo-modules-autolinking/scripts/ios/precompiled_modules.rb
```diff
@@ -1978,7 +1978,7 @@ def resolve_prebuilt_tarball(pod_info, product_name, flavor, pod_name = nil)
           (idx ? pod_info[:build_output_dir][idx..] : output_prefix),
           flavor,
           'xcframeworks',
-          "#{product_name}.tar.gz"
+          "#{product_name}.tar.xz"
```
tsapeta, line 1981: [🐧 nit] Worth confirming the GCS transition: after this lands, the download path requests `.tar.xz` URLs, while any artifacts published to `gs://eas-build-precompiled-modules/` before this PR exist only as `.tar.gz`. For version-prefixed paths that won't be republished, a consumer on a cache miss will 404 on the `.tar.xz` and fall back to building from source. That's a safe degradation, not a break — but if you want to avoid the source-build penalty for already-published SDK/RN combinations, the publish workflow may need a one-time backfill/recompress, or the bucket entries for active prefixes need regenerating. Flagging so it's a conscious decision rather than a surprise.

gabrieldonadel (approved):
Great idea Chris! LGTM

tools/src/prebuilds/Frameworks.ts
```diff
@@ -631,6 +632,93 @@ const copySPMDependencyXCFrameworksAsync = async (
...
+      if (settled) return;
+      settled = true;
+      tar.kill();
+      xz.kill();
+      reject(error);
+    };
```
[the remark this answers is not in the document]
chrfalch: Fixed

tools/src/prebuilds/Frameworks.ts
```diff
@@ -232,7 +233,7 @@ export const Frameworks = {
   ): string => {
     return path.join(
       Frameworks.getFrameworksOutputPath(buildPath, buildType, versionPrefix),
-      `${productName}.tar.gz`
+      `${productName}.tar.xz`
     );
```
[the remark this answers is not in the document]
chrfalch: Fixed

packages/expo-modules-autolinking/scripts/ios/precompiled_modules.rb
chrfalch: I think we should be good by just running the workflow for rebuilding them.

tools/src/prebuilds/Frameworks.ts
kitten: this also usually is a good sign to modularise into passthrough streams and use `pipeline` imo

tools/src/prebuilds/Frameworks.ts
```diff
@@ -631,6 +632,126 @@ const copySPMDependencyXCFrameworksAsync = async (
...
+ *
+ * tar and xz run as two piped processes writing to a temp file; the archive is renamed into
+ * place only once tar exits 0, xz exits 0, and the output fd has closed — so a truncated
+ * `.tar.xz` can never appear at the final path (even on a crash) and be mistaken for a complete
+ * one. Pipe-level stream errors (e.g. EPIPE when xz dies before tar finishes) are funneled
+ * through the same failure path as process spawn/exit errors.
+ */
+const createXzTarballAsync = (tarballPath: string, cwd: string, entries: string[]): Promise<void> =>
+  new Promise<void>((resolve, reject) => {
+    const tar = spawn('tar', ['-cf', '-', '-C', cwd, ...entries], {
```
kitten: the spread here and shelling out into the `tar` binary from PATH feels a bit precarious

tsapeta (approved):

PR #46801, opened 2026-06-11
refactor: Drop committed build outputs and migrate to Turborepo

kitten:
Resolves [ENG-20741](https://linear.app/expo/issue/ENG-20741)

## Summary

This PR implements four groups of refactor tasks:
- Cross-repo JS build tasks with Turbobuild
- `**/build/**` output folders removed from tracked files
- JS build scripts switched to an `expo-build` script (SWC-based)
- `package.json:exports` added with `source` conditions for `src` TS files

This, as a consequence means:
- `turbo build` is necessary to build source files for many tasks; to make this easier it's been added as a repo `prepare` task (on `pnpm i`) for now
- TypeScript `tsc` is not used as a pre-publish build tool anymore, and we're free to use it for type-checking only
- The `expo-build` script loosely encompasses all of our usual build configurations and skips type-checking during `turbo build` (manual `pnpm build` invocations will still type check)
- Many changes for type checks reflect immediately due to `customConditions: ["source"]` in the `tsconfig.base.json`
- "Sub-task" builds are now gone, `expo-module build plugin` et al are replaced with focused `package.json:scripts` in the root package
- Many packages that published raw TypeScript source files either have already been switched over to use `build` outputs, or should be switched over

As a secondary consequence for future refactors:
- We can update `tsconfig.json` files to include test files
- We can update Jest's handling of TypeScript files to not type check (and update it to Jest 30, and do other various refactors)
- We can automate secondary task tooling, like updating lint runs, replacing `check-packages`, publishing, etc more easily
- We can potentially cut `main`'s git history to speed up the repo again (needs to be coordinated with GitHub Support most likely in ~3 SDK cycles)

### Usage Guide

`pnpm i` will (for now) run `turbo build` automatically and will build all sources. The resulting repository structure should be vaguely the same as before the PR (except build outputs won't be tracked by Git)

You can also run focused builds with `turbo build --filter [pkg]` or with `pnpm turbo build` from an individual package, which not only builds that individual package but all its transitive dependencies too.

If you have direnv set up and activated, the read-only remote build cache for Turborepo is already activated, and most packages won't build on your machine. Turborepo will instead pull cached build outputs that CI has produced from the remote build cache on Cloudflare. (Check `.env` for the public env vars that activate the remote build cache)

## Set of changes

- Add Turbo repo
- Add `expo-build` script
- Update various `package.json` scripts, exports, entrypoints
- Add `source` condition to various `exports` on packages that are used as dependencies
[...]

PR #47034, opened 2026-06-18
refactor: Update `tsconfig.json` / tsc practices and Jest runners (composites)

kitten:
Follow-up after #46801

## Summary

**We'd like:**
- A unified `pnpm run typecheck` script to run type checks per package
- A unified `pnpm run test` script to run *unit* tests per package
- Runnable via `turbo typecheck` and `turbo test`
- Type-checked test files

**We must:**
- Align `tsconfig.json`s not to exclude any test files
- Create composite `references` in `tsconfig.all.json` where multiple `tsconfig.json`s are present for one package
- Create composite Jest configs with `projects` where multiple Jest configs are present
- Remove redundant watch scripts and ensure `test` only has one meaning

**We can:**
- Drop `ts-jest`, as type validation isn't necessary as part of a Jest run
- Drop `babel-jest` in many cases where SWC is appropriate instead
- Drop Babel configs that are redundant

### Notes

1. `expo-brownfield` has an uncovered `e2e` folder that also cross-imports. This is causing issues, and some tests feel like they were meant to be unit tests. Others are running as part of E2E, but both run in check-packages. Fixing this is out of scope for this PR.

2. This does not yet update/replace `check-packages`, which can now be slimmed down, so next steps after this PR are:
- extract the dependencies check into a per-package script we can run via Turbo
- run the checks we want via Turbo instead
- see if we still need `check-packages` after (likely, yes, temporarily)

3. We may have to pull `expo-module-scripts` (as of the `sdk-56` state) out of `expo/expo` if we want to "pin" it in place for external consumers. We may then want to rename our own to avoid any confusion.

## Set of changes

- Remove `exclude` entries from `tsconfig.json`s
- Create aggregate `tsconfig.all.json` files where multiple `tsconfig.json`s must be checked
- Remove redundant npm scripts
- Consistently define `typecheck` npm scripts
- Switch Jest over to swc where we can
- Align Jest configs where the wrong preset was used
- Update how Jest's Babel options are created when config is missing (match `@expo/metro-config`)
- Rename `source` condition to `expo-source` to avoid conflicts
- Fix various type errors
- Fix various test errors
- Limit concurrency under Turbo for Jest


Ubax (approved):
The changes in router, app-metrics, and observe look good to me

hassankhan (approved):

tsapeta (approved):

PR #37422, opened 2025-06-13
[menu][Android] Rewrite using jetpack compose

lukmccall:
# Why

Rewrites dev-menu using Jetpack Compose

# How
This is the first part of the dev menu rewrite using Jetpack Compose. 
In this PR, I've revamped the entire UI with native code. While not everything is styled exactly as before, some parts are still being polished. 

The theme was generated based on our current solution in JS and contains some unused code. I will clean that up after I’ve rewritten the launcher. 

# Test Plan

- bare-expo ✅ 
- compose preview ✅ 

image

PR #47047, opened 2026-06-19
[android][dev-menu] Restore "Open React Native dev menu" entry

lindboe:
# Why

We noticed after an SDK 53->55 upgrade, pn Android, the dev menu no longer has an "Open React Native dev menu" entry, even though the underlying action (`DevMenuAction.OpenReactNativeDevMenu`) and its handler — which calls `reactHost?.devSupportManager?.showDevOptionsDialog()` — still exist.

That menu item does still exist on iOS. The iOS menu renders [`DevMenuRNDevMenu`](https://github.com/expo/expo/blob/0b840f12505298c6c95ef7cc700318719b19704a/packages/expo-dev-menu/ios/SwiftUI/DevMenuRNDevMenu.swift#L10) — a button labeled "Open React Native dev menu" — from [`DevMenuMainView`](https://github.com/expo/expo/blob/0b840f12505298c6c95ef7cc700318719b19704a/packages/expo-dev-menu/ios/SwiftUI/DevMenuMainView.swift#L38-L39).

**This entry also used to exist on Android.** The Jetpack Compose rewrite ([#37422](https://github.com/expo/expo/pull/37422)) included the button and dispatched the action here: [`DevMenuScreen.kt#L109`](https://github.com/expo/expo/blob/a062b48a6d0a6749908a7baf56811ec22babb42b/packages/expo-dev-menu/android/src/main/java/expo/modules/devmenu/compose/DevMenuScreen.kt#L109). The later UI refresh ([#38759](https://github.com/expo/expo/pull/38759)) reorganized the menu into `ToolsSection` and dropped that row, while leaving the `DevMenuAction.OpenReactNativeDevMenu` action and its handler in place — making it unreachable.

Our team asked about it missing after it was removed, so we did find developers were using it.

# How

- Re-add a row to `ToolsSection` (after "Open DevTools") that dispatches the existing `DevMenuAction.OpenReactNativeDevMenu`.
- Add a `MenuIcons.Gear` composable backed by the existing `gear_fill.xml` drawable (no new asset).
The entry shows unconditionally, matching the previous Android behavior. (On iOS it's gated behind the `devMenuShouldShowReactNativeDevMenu` host delegate; Android has no equivalent delegate, so this restores the prior unconditional behavior.)

# Test Plan

- Built `apps/bare-expo` before and after the change.
- Confirmed after changes:
  - the "Open React Native dev menu" row appears under TOOLS, below "Open DevTools".
  - Tap it → the React Native dev options dialog appears and the Expo dev menu closes 

### Screenshots (Android)
| Before | After |
| --- | --- |
|  | <video src="https://github.com/user-attachments/assets/c6676837-512b-46e1-a03b-6f307a8d1de8" width=300 /> |

# Checklist
- [x] I added a `changelog.md` entry and rebuilt the package sources according to this short guide

### DIFF
--- a/packages/expo-dev-menu/CHANGELOG.md
+++ b/packages/expo-dev-menu/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ### 🐛 Bug fixes
 
+- [Android] Restore the "Open React Native dev menu" entry in the dev menu's Tools section (regressed in [#38759](https://github.com/expo/expo/pull/38759)). ([#PR](https://github.com/expo/expo/pull/PR) by [@lindboe](https://github.com/lindboe))
 - [iOS] Show the floating Tools button at its final position instead of animating in from the top-left corner. ([#46762](https://github.com/expo/expo/pull/46762) by [@alanjhughes](https://github.com/alanjhughes))
 
 ### 💡 Others
--- a/packages/expo-dev-menu/CHANGELOG.md
+++ b/packages/expo-dev-menu/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ### 🐛 Bug fixes
 
-- [Android] Restore the "Open React Native dev menu" entry in the dev menu's Tools section (regressed in [#38759](https://github.com/expo/expo/pull/38759)). ([#PR](https://github.com/expo/expo/pull/PR) by [@lindboe](https://github.com/lindboe))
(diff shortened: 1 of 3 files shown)


lindboe:
Was unsure if this should be backported? Let me know if that's something I should do.

lukmccall (approved):
Looks good!

PR #47057, opened 2026-06-19
test(expo-sqlite): Replace `better-sqlite3` with `node:sqlite` built-in

kitten:
## Summary

We don't have to keep these tests on `better-sqlite3` and there's different candidates to replace it (e.g. libsql and `node:sqlite`). Since `node:sqlite` is now a built-in (as of Node 22+), it's probably the best candidate to replace `better-sqlite3`.

The problem with keeping `better-sqlite3` is a minor one, but the compilaton slows down CI on every install (since the build output isn't cached in CI), since there's no prebuilts (fresh `node-gyp` per install)

## Set of changes

- Drop `better-sqlite3`
- Update mock to `node:sqlite`
- Conditioanlly skip `serialize` test (it's just a passthrough and `node:sqlite` on lowest supported Node 22 doesn't yet support `serialize`)
- Reduce delays since tests are just artificially slow

### DIFF
--- a/packages/expo-sqlite/package.json
+++ b/packages/expo-sqlite/package.json
@@ -66,11 +66,9 @@
   },
   "devDependencies": {
     "@testing-library/react-native": "^13.3.0",
-    "@types/better-sqlite3": "^7.6.6",
     "@types/jest": "^29.2.1",
     "@types/node": "^22.14.0",
     "@types/react": "~19.2.0",
-    "better-sqlite3": "^11.6.0",
     "expo": "workspace:*",
     "expo-module-scripts": "workspace:*",
     "react-error-boundary": "^4.0.11"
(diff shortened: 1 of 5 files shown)


Kudo:
but expo-sqlite check-packages is failing

kitten:
oops, sorry didn't see the failure 😅 I assumed that testing locally against Node 24 would result in good coverage, but my version of Node 24 didn't have `serialize`, so I retested against Node 26, which reproduces the issue (Incorrect implementation of that code-path/test that is skipped in 22)

Also pushed a commit to drive-by fix those `act` warnings, since I restructured those since I wasn't sure whether they're expected (the warnings aren't but the tests don't fail, luckily, after fixing them)

Edit(Note): the `check-packages` run (based on the updates on main) doesn't list `sqlite:test` because the `test` outputs are set to `"outputLogs": "errors-only"` (that's a todo until I get the jest reporter to not be as noisy), but the lack of error reporting means it has run and passed (it not being shown isn't ideal, but the alternative of showing all default jest output is quite noisy locally)

Kudo (approved):
nice to use builtin sqlite. thanks for helping on that and fixing the test errors

PR #47069, opened 2026-06-19
test(router-server,router): Drop `tsd`, `tsd-lite`, and `jest-runner-tsd` for type tests

kitten:
## Summary

We can simplify (and modernise) the type tests a lot by applying one of two patterns each:
- Use raw `expect-type` covered by the usual `tsc`/`pnpm run typecheck` type checks
- Use `@typescript/vfs` and programmatic project creation where we need to evaluate generated types

## Set of changes

- Migrate expo-router to raw `expect-type`
- Migrate router-server to generated type tests via `@typescript/vfs`


packages/@expo/router-server/package.json
```diff
@@ -24,7 +24,7 @@
     "clean": "expo-module clean",
     "lint": "expo-module lint",
     "test": "jest",
-    "test:tsd": "EXPORT_ROUTER_JEST_TSD=true jest",
+    "test:rsc": "jest --config jest-rsc.config.js",
```
Ubax: This file seems to be missing

Ubax (approved):
Please add the rsc config or let me know why current approach works

packages/@expo/router-server/package.json
kitten: Ah, that's a merge conflict that I missed after a rebase, I believe. The `jest-rsc.config.js` was a bit hard to follow but was merged into project lists in #47063 and became parallel `projects` entries there. Since it only applies to `__rsc_tests__` (if we go back a commit before, `expo-router` can be run with them and lists only `__rsc_tests__` tests, and `@expo/router-server` lists none) it's safe to remove, since `@expo/router-server` has no RSC tests / `__rsc_tests__` folders

hassankhan (approved):

packages/@expo/router-server/package.json
kitten: merging, since the above relates to the other PR (#47063). Can catch up on that next week though separately

PR #47070, opened 2026-06-19
[docs] Remove duplicate UI component entries

krolyk:
Removed duplicate entries for Jetpack Compose and SwiftUI.

# Why


amandeepmittal (approved):
Thank you for the fix!

PR #47071, opened 2026-06-19
chore(log-box): Drop typescript-plugin-css-modules & update `tsconfig.json`

kitten:
## Summary

The amount of dependencies (many outdated/unmaintained) it's pulling in seems like a bad trade-off for the value provided. The package itself hasn't been updated in a bit, and we're probably safe to drop this for now.

## Set of changes

- Drop `typescript-plugin-css-modules`
- Re-merge `tsconfig.json` and add missing `./app` include


krystofwoldrich (approved):

krystofwoldrich:
Looks good, ~we should revisit the log-box type check later~ seems like my local issue.

PR #47085, opened 2026-06-20
[android][compose] Skip onLayout when view is detached from window

roitium:
# Why

When an ExpoUI <Host/> is detached from the window — typically during navigation transitions — a pending `requestLayout()` runnable posted to the main thread handler can execute after detachment. This causes `onLayout()` to call through to Compose's internal layout, which then tries to access a LayoutNode whose owner has already been nullified by the composition disposal, resulting in:

> IllegalStateException: LayoutNode should be attached to an owner.

It's a race condition, and the `onMeasure()` already had the guard. But `onLayout()` doesn't have, so this PR add this guard to `onLayout()`.

# How

Added the same shouldUseAndroidLayout && !isAttachedToWindow guard to onLayout() that already exists in onMeasure().

# Test Plan

I build my own app on the new code, and it succeed. 

# Checklist

- [-] I added a `changelog.md` entry and rebuilt the package sources according to [this short guide](https://github.com/expo/expo/blob/main/CONTRIBUTING.md#-before-submitting)


intergalacticspacehighway (approved):
Thanks for fixing!

Kudo (approved):

PR #47089, opened 2026-06-20
test(file-system): Remove re-entry import to fix type check and adjust type root

kitten:
## Summary

The test fails the type-check here since it re-enters the package with `../..` i.e. the package root itself. This _can_ in theory work with the custom condition of `expo-source`, but is redundant, since it means to just get common shared modules.

## Set of changes

- Replace `../..` (package root) with `../index` in test
- ~~Fix accidentally changed `rootDir` on `tsconfig.json`~~ (See #47094 which overrides `rootDir` and makes this non-loadbearing and not a footgun)

### DIFF
--- a/packages/expo-file-system/src/__tests__/FileSystem-test.native.ts
+++ b/packages/expo-file-system/src/__tests__/FileSystem-test.native.ts
@@ -1,6 +1,6 @@
-import { DownloadTask, File, Directory, Paths, UploadTask } from '../..';
 import { __resetMockFileSystem } from '../../mocks/FileSystem';
 import { FileMode } from '../File.types';
+import { DownloadTask, File, Directory, Paths, UploadTask } from '../index';
 
 beforeEach(() => {
   __resetMockFileSystem();


packages/expo-file-system/src/__tests__/FileSystem-test.native.ts
```diff
@@ -1,4 +1,4 @@
-import { DownloadTask, File, Directory, Paths, UploadTask } from '../..';
+import { DownloadTask, File, Directory, Paths, UploadTask } from '../index';
```
tsapeta: Warning here seems valid
```
`../index` import should occur after import of `../File.types`
```

tsapeta (approved):

packages/expo-file-system/src/__tests__/FileSystem-test.native.ts
kitten: oops 😅

hassankhan (approved):

kitten:
I'll have to check the convention here. `expo-file-system` re-imports from `mocks/` but that's not under `src/`. However, that means it'll have to be included by `rootDir`, which changes the output format. We don't have a convention that says only `src` gets built (and sometimes even avoid that), so we'll have to either move `mocks` (which requires a `jest-expo` change, afaik), or find another path

kitten:
See: #47094 (checks here will only pass with its changes, fixing the `build` output)

We need to force `rootDir` on the build script to accommodate the custom `rootDir` on `expo-file-system` (and avoid footguns here)

PR #47092, opened 2026-06-20
[test-suite] Remove Firebase tests and Cloud Functions package

tsapeta:
# Why

`apps/test-suite` carried a full Firebase footprint that we don't actually need:

- A standalone Firebase **Cloud Functions** package (`apps/test-suite/functions`) with its own `firebase-functions` / `firebase-admin` dependencies and a `yarn.lock`.
- The **Firebase JS SDK compatibility test** (`tests/FirebaseJSSDK.js`) plus the `firebase` client dependency and the Firebase project config (`firebase.json`, `.firebaserc`, `firebase/` rules).

These dependencies trigger frequent Dependabot alerts, and the tests weren't hermetic anyway: they ran against live data in an external `expo-payments` Firebase project (a deployed `echoMessage` function, seeded Firestore docs, files in storage), so they were never a reliable local or CI signal.

# How

Removed the whole Firebase surface from `test-suite`:

- Deleted the `functions/` Cloud Functions package.
- Deleted `tests/FirebaseJSSDK.js` and unregistered it from `TestModules.ts`.
- Dropped the `firebase` dependency from `package.json` and regenerated `pnpm-lock.yaml` (~750 fewer lines).
- Removed the now-unused Firebase project config: `firebase.json`, `.firebaserc`, and the `firebase/` rules directory.

The `expo-payments` Firebase project and its deployed resources are left untouched; nothing in the repo references them anymore, so they can be torn down separately.

# Test Plan

- `pnpm install --frozen-lockfile` passes (committed lockfile is self-consistent; `firebase` no longer appears in it).
- `apps/test-suite` typecheck (`pnpm tsc`) introduces no new errors. The pre-existing baseline errors in unrelated files (`Video.tsx`, `Audio.ts`, etc.) are unchanged; the change only removes one `require()` line from a list in `TestModules.ts`.


kitten (approved):
❤️😍

PR #47095, opened 2026-06-20
[metro-config] Fix findUpPackageJson infinite recursion at the Windows drive root

patrickwehbe:
# Why

`findUpPackageJson` in `@expo/metro-config` (used by `getPkgVersion` to read a dependency's installed version) walks up the directory tree until it reaches the root. Its stop condition only checks the current path against `'.'` and `path.sep`:

```ts
if (['.', path.sep].includes(cwd)) return null;
// ...
return findUpPackageJson(path.dirname(cwd));
```

On Windows the drive root is neither of those. `path.dirname('D:\\')` returns `'D:\\'` (a fixed point), which does not equal `'.'` and does not equal `path.sep` (`'\\'`). So if no `package.json` exists anywhere up the tree, the walk reaches `D:\` and keeps recursing on the same value forever, ending in:

```
RangeError: Maximum call stack size exceeded
```

On POSIX this never triggers because the walk lands on `path.sep` (`'/'`) and stops, which is why it has gone unnoticed. The real-world trigger is narrow: it requires a package to resolve to a location with no `package.json` in any ancestor directory on Windows, but when it happens the error is a confusing stack overflow rather than a clean `null`.

# How

Stop when a directory's parent equals itself (`path.dirname(cwd) === cwd`), which is true for both the POSIX root `/` and any Windows drive root such as `D:\`. The `cwd === '.'` case is kept for the relative-path entry.

This is the same idiom the two sibling helpers in this repo already use:

- `packages/@expo/metro-config/src/serializer/findUpPackageJsonPath.ts` — `for (let dir = root; path.dirname(dir) !== dir; dir = path.dirname(dir))`
- `packages/install-expo-modules/src/utils/projectRoot.ts` — same loop condition

# Test Plan

Added a regression test to `getPkgVersion.test.ts` that forces Windows path semantics (`path.win32.dirname`) so it runs on CI's POSIX environment, points `findUpPackageJson` at a drive-root path with no resolvable `package.json`, and asserts it returns `null`.

- Against the current code the new test fails with `RangeError: Maximum call stack size exceeded`.
- With the fix all tests pass:

```
  findUpPackageJson
    √ should return package.json path when found in the given directory
    √ should walk up directories until package.json is found
    √ should return null when reaching the root directory
    √ should return null for relative root
    √ should return null when reaching a Windows drive root without recursing forever

Tests:       12 passed, 12 total
```

The two pre-existing base-case tests (`path.sep` and `'.'`) still pass, so the change is backward compatible.

# Checklist

- [ ] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: updated a module plugin). — n/a, no plugin/prebuild surface touched
- [ ] Conforms with the [Documentation Writing Style Guide](https://github.com/expo/expo/blob/main/guides/Expo%20Documentation%20Writing%20Style%20Guide.md) — n/a, no docs changes

### DIFF
--- a/packages/@expo/metro-config/src/utils/getPkgVersion.ts
+++ b/packages/@expo/metro-config/src/utils/getPkgVersion.ts
@@ -20,11 +20,18 @@ export function getPkgVersionFromPath(packageJsonPath: string): string | null {
 }
 
 export function findUpPackageJson(cwd: string): string | null {
-  if (['.', path.sep].includes(cwd)) return null;
+  // Stop when we reach a directory whose parent is itself, e.g. the POSIX root `/` or a
+  // Windows drive root like `D:\`. `path.dirname('D:\\')` returns `'D:\\'`, which is neither
+  // `.` nor `path.sep`, so checking only against those values would recurse forever on Windows
+  // when no package.json exists up the tree. The sibling helpers in
+  // serializer/findUpPackageJsonPath.ts and install-expo-modules/src/utils/projectRoot.ts use
+  // this same `path.dirname(dir) !== dir` idiom.
+  const parent = path.dirname(cwd);
+  if (parent === cwd || cwd === '.') return null;
 
   const found = resolveFrom.silent(cwd, './package.json');
   if (found) {
     return found;
   }
-  return findUpPackageJson(path.dirname(cwd));
+  return findUpPackageJson(parent);
 }
--- a/packages/@expo/metro-config/CHANGELOG.md
+++ b/packages/@expo/metro-config/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### 🐛 Bug fixes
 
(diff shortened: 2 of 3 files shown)


packages/@expo/metro-config/src/utils/getPkgVersion.ts
```diff
@@ -20,11 +20,18 @@ export function getPkgVersionFromPath(packageJsonPath: string): string | null {
 }
 
 export function findUpPackageJson(cwd: string): string | null {
```
kitten, line 22: There's a fixed up / duplicate version of this elsewhere in the codebase. Not saying this isn't correct, but we have 3+ versions of this, and they'll all likely need a review since this was already fixed before for another

I'll have a look at this when I have a minute to check this

Edit: also please keep in mind that submitting PRs (especially LLM derived) without a prior issues and details on the exact reproduction paths do cause extra work, due to lack of cases/repros/steps to test against (for context on why this would need an extra manual check)

kitten:
I've reviewed all similar utilities in this repo and there's quite a few more that had to be fixed up unfortunately. Thanks for flagging them up! https://github.com/expo/expo/pull/47127

PR #47096, opened 2026-06-20
chore: use `oxfmt` and `oxlint`

hassankhan:
# Why

We want to evaluate replacing ESLint with [oxlint](https://oxc.rs) and Prettier with [oxfmt](https://oxc.rs/docs/guide/usage/formatter.html) for speed/performance improvements.

# How

- Added `oxlint` and `oxfmt` and per-package `lint`/`format` scripts; lint uses [`oxlint-config-universe`](https://github.com/expo/oxlint-config-universe).
- Created a shared base config (`expo-module-scripts/oxlint.config.base`) extending the preset and documented any differences from it.
- Replaced ESLint's `import/order` with Oxfmt's `sortImports`
- Removed ESLint and Prettier from nearly all packages (`apps/`, `docs/` and `tools/` haven't been changed)
- Added an `expo-module format` command to wrap `oxfmt`; this only performs checks in a CI environment, but modifies files when run locally
- Ran `oxfmt --write` across the monorepo
- Updated CI workflows to run linter + formatter (in that order)
- Fixed a few minor bugs/issues surfaced by linter
  - `@expo/cli`: `maxWarnings` evaluating to `NaN` when `undefined`
  - `expo-router`: missing `key` props on `<Screen>` components
  - `native-component-list`: a JSX `style` using the comma operator instead of an array, resulting in one style silently never being applied
  - `expo-camera`: a minor improvement for `no-unsafe-finally` (thanks @zoontek!)
  - smaller style/refactor cleanups across `@expo/cli`, `@expo/metro-file-map`, `expo-image-picker` and `expo-network`

```
## on main
$ turbo run lint --filter="./packages/**"
 Tasks:    130 successful, 130 total
Cached:    0 cached, 130 total
  Time:    33.358s

## on this PR
$ turbo run format lint --filter="./packages/**"
 Tasks:    258 successful, 258 total
Cached:    0 cached, 258 total
  Time:    19.437s
```

# Test Plan

- Run `turbo run format lint` locally (use the `--output-logs=errors-only` flag to only see errors)
- Run `pnpm --filter <PACKAGE_NAME> lint` and/or `pnpm --filter <PACKAGE_NAME> format` to run oxlint/oxfmt on a particular package

# Checklist


packages/expo-module-scripts/oxlint.config.base.js
```diff
@@ -0,0 +1,82 @@
...
+    // Empty leading branches (e.g. `if (x) {} else if ...`) are used as intentional guard clauses.
+    'no-lone-blocks': 'off',
+    // We declare properties on `globalThis`.
+    'no-shadow-restricted-names': 'off',
+    // Triple-slash references are an accepted pattern in this codebase.
+    'typescript/triple-slash-reference': 'off',
+    // Empty files are intentional platform-specific no-op stubs (e.g. `*.web.ts` overrides).
+    'unicorn/no-empty-file': 'off',
+    // `new Array(n)` is used intentionally to pre-allocate arrays.
+    'unicorn/no-new-array': 'off',
```
hassankhan, line 94: Most of these could be replaced by `Array.from({ length: n })`

packages/expo-module-scripts/oxlint.config.base.js
```diff
@@ -0,0 +1,94 @@
...
+  rules: {
+    // fbjs is a Facebook-internal package not intended to be a public API.
+    'no-restricted-imports': ['warn', { patterns: ['fbjs/*', 'fbjs'] }],
+
+    // -----------------------------
+    // --- Intentional patterns  ---
+    // -----------------------------
+
+    // @kitten/@zoontek should disable this rule and audit resulting warnings/errors
+    'no-async-promise-executor': 'off',
```
hassankhan: /cc @kitten @zoontek

packages/expo-module-scripts/oxlint.config.base.js
```diff
@@ -0,0 +1,94 @@
...
+    // Empty leading branches (e.g. `if (x) {} else if ...`) are used as intentional guard clauses.
+    'no-lone-blocks': 'off',
+    // We declare properties on `globalThis`.
+    'no-shadow-restricted-names': 'off',
+    // Throwing non-Error values is an accepted pattern in this codebase.
+    'no-throw-literal': 'off',
+    // @kitten/@zoontek should disable this rule and audit resulting warnings/errors
+    'no-unsafe-optional-chaining': 'off',
+    // @kitten/@zoontek should disable this rule and audit resulting warnings/errors
+    'typescript/no-non-null-asserted-optional-chain': 'off',
```
hassankhan: /cc @kitten @zoontek

packages/create-expo-nightly/oxlint.config.mjs
```diff
@@ -0,0 +1,2 @@
+// @generated by expo-module-scripts
```
kitten, line 1: this comment is potentially misleading if we haven't updated `expo-modules-scripts`'s `configure` / template scripting? either way, can remove across files most likely

packages/uri-scheme/package.json
```diff
@@ -30,6 +30,9 @@
     "build:prod": "ncc build ./src/index.ts -o build/ --minify --no-cache --no-source-map-register",
     "clean": "expo-module clean",
     "lint": "expo-module lint",
+    "lint:oxlint": "oxlint --config oxlint.config.mjs src",
+    "format": "oxfmt --write src",
+    "format:check": "oxfmt --check src",
     "test": "jest",
```
kitten: two things re. the `package.json` setup,

- it'd be interesting to check if we need an `oxlint.config.mjs` file (or can make it a dotfile for npmignore simplicity, since in many cases we don't use `files`)
- could this config be a "global" config in the repo root and repeated in `package.json` for isolation? not sure if oxlint supports a `package.json` config entry
- could the `package.json` scripts be reduced? I assume `lint:oxlint` is temporary, but keeping track of `src` folders could be tedious. Some packages have many sub-folders, so if we can make this auto-discover (even if we have to wrap like with `expo-module lint`), that'd reduce the amount of mental overhead
- re above; I'd above `format` / `format:check` if arguments overlap. `format:check` feels redundant here

packages/expo-updates/package.json
```diff
@@ -29,6 +29,9 @@
     "build:cli": "expo-build --base cli cjs:src",
     "clean": "expo-module clean",
     "lint": "expo-module lint",
+    "lint:oxlint": "oxlint --config oxlint.config.mjs src",
```
kitten: to illustrate the other comment, here we'd have to check `cli`, `e2e-cli`, `e2e`, `plugin`, and `utils`. Might be worth it to check if we can somehow add ignores to the oxlint preset, so we can just do `.`, if there's no auto-discovery default?

packages/expo-module-scripts/oxlint.config.base.js
kitten: not a big deal, let's just disable this. I'd be interested why it's disabled, but as discussed in DMs, this overlaps with TypeScript, so if the intention of code adds a `non-null` assertion, we should assume that assertion is fine, over letting oxlint question it again

packages/expo-module-scripts/oxlint.config.base.js
kitten: imo, shouldn't be disabled, and it's unlikely we care because of TypeScript. It's protecting against `new Array(elem)` vs `new Array(length)`. In practice, tests should always catch this, and I don't think there's a direct need to yield to a rule that protects us from a rare case over an established JS pattern that engines optimise for

kitten:
**tangential / re. rules:** (Not related to the review, but would like to dump thoughts on this somewhere)

Personally, I'd never activate rules speculatively that are stylistic or disrupt code that's been intentionally written a certain way. Lints are useful if they're activated in place where mistakes have happened, and where we have reason to believe they'll happen again (the latter is an important distinction since one-off issues are possible)

Meant by that are rules that align patterns for the sake of aligning them. Often this overlaps with micro-optimisations, older code that doesn't warrant an edit (as in, it's fine, and the edit is arguably a trade-off or dehancement in one way), or stylistic differences.

For example, I suspect `no-async-promise-executor` is useful in many cases across "all code" that's written outside of Expo inclusive. However, within `expo/expo`, the trade-off is how many times it'd disrupt code, versus how many mistakes it'd catch. As in, what's the amount of disruptive warnings it'd flag for for a single mistake it catches. That ratio is more important to me, then judging rules individually by their theoretically correctness.

If we have to leave a disabling comment on code (or worsen it / write it in a way in which we usually wouldn't) to shut a rule up, that's potentially a bad trade-off if it almost never catches a related mistake.

Another good example of this is `no-control-regex` which is (as far as I'm aware) not a mistake we've made — hence it's disabled now.

I'd also consider a second category of rules that may be annoying if we enable editor integrations. For example, `no-lone-blocks` is potentially to nagging / annoying in development, when code is still being changed.

---

tl;dr: Just to have it captured, imho, we should always aim to enable fewer rules rather than adding more speculatively, and in the interest of being able to get this merged quickly, if we have disagreements on rules, disabling one should be prioritised and the default decision

packages/expo-module-scripts/oxlint.config.base.js
hassankhan: Fair, my opinion was that `Array.from()` reads a bit nicer, but its not a good enough reason to make the change.

packages/create-expo-nightly/oxlint.config.mjs
hassankhan: I was planning to add it to `expo-module configure`, but can remove it in any case.

packages/uri-scheme/package.json
hassankhan: > it'd be interesting to check if we need an `oxlint.config.mjs` file (or can make it a dotfile for npmignore simplicity, since in many cases we don't use `files`)

We can rename the file to `.oxlintrc.json`, I kept it as an `.mjs` file temporarily in case I needed to add comments explaining rationale but as it turns out, the `.oxlintrc.json` is JSONC so I'll try switching them over in this PR.

> could this config be a "global" config in the repo root and repeated in `package.json` for isolation? not sure if oxlint supports a `package.json` config entry

We could lift the "base" configuration from `expo-module-scripts` into the repo root, but I don't think there's a way of configuring oxlint via `package.json` config entries.

> could the `package.json` scripts be reduced? I assume `lint:oxlint` is temporary, but keeping track of `src` folders could be tedious. Some packages have many sub-folders, so if we can make this auto-discover (even if we have to wrap like with `expo-module lint`), that'd reduce the amount of mental overhead

`lint:oxlint` is indeed temporary, I just wanted to gauge feedback on the approach before switching out ESLint + Prettier completely. I haven't yet looked into subfolder autodiscovery, but it was on the list.

> re above; I'd above `format` / `format:check` if arguments overlap. `format:check` feels redundant here

Good point, I'd added `check` to make it easy to run via `turbo run format:check` as a quick way to run it via `check-packages` eventually, but I'll take another look.

packages/expo-camera/src/web/useWebBarcodeScanner.ts
```diff
@@ -97,6 +97,7 @@ export function useWebBarcodeScanner(
     } finally {
       if (interval === 0) {
         stop();
+        // oxlint-disable-next-line no-unsafe-finally
```
zoontek: Instead of silencing it, use an `else`:

```ts
} finally {
      if (interval === 0) {
        stop();
      } else {
        const intervalToUse = !interval || interval < 0 ? 16 : interval;
        timeout.current = setTimeout(() => {
          scanAsync();
        }, intervalToUse);
      }
    }
```

turbo.json
```diff
@@ -36,6 +36,9 @@
     "lint": {
       "cache": false
     },
+    "lint:oxlint": {
+      "cache": false
+    },
```
zoontek, line 44: I think you also need to add `format:check` and `format:write` here

packages/expo-module-scripts/package.json
```diff
@@ -83,6 +91,7 @@
     "commander": "^12.1.0",
     "eslint-config-universe": "workspace:^15.2.0",
     "glob": "^13.0.0",
+    "oxlint-config-universe": "^0.0.3",
```
tsapeta: [🐧 nit] Nit: `oxlint-config-universe` is inserted between `glob` and `jest-expo`, which breaks the alphabetical ordering of `devDependencies` (it should sort after `jest-watch-typeahead`). Minor, but worth fixing before the final PR since `expo-module depscheck`/sort tooling may flag it.

packages/expo-module-scripts/oxlint.config.base.js
```diff
@@ -0,0 +1,114 @@
...
+      env: { node: true },
+      globals: {
+        // Tests reassign `__DEV__`
+        __DEV__: 'writable',
+      },
+      rules: {
+        // Tests need to re-assign imported bindings to install mocks and spies.
+        'no-import-assign': 'off',
+        //
+        'no-multi-assign': 'off',
```
tsapeta, line 29: [🐧 nit] Stray empty comment (`//`) above `'no-multi-assign': 'off'` — looks like a rationale line that got dropped. Every other disabled rule in this block has an explanation, so either add the reason or remove the orphaned comment marker.

packages/expo-module-scripts/oxlint.config.base.js
```diff
@@ -0,0 +1,114 @@
+import native from 'oxlint-config-universe/native';
+import { defineConfig } from 'oxlint';
```
tsapeta: Wrong order
```suggestion
import { defineConfig } from 'oxlint';
import native from 'oxlint-config-universe/native';
```

packages/expo-module-scripts/oxlint.config.base.js
```diff
@@ -0,0 +1,114 @@
...
+    // eslint-config-prettier. Kept off here so the switch to oxlint is behavior-neutral; the final
+    // migration commit flips this to 'warn' and runs `oxlint --fix` to add the braces.
+    curly: 'off',
+    'no-useless-computed-key': 'off',
+    'unicorn/prefer-string-starts-ends-with': 'off',
+    'unicorn/no-useless-fallback-in-spread': 'off',
+    'unicorn/no-useless-spread': 'off',
+
+    // React props that pass JSX should always be wrapped by curly braces.
+    'react/jsx-curly-brace-presence': ['warn', { props: 'never', children: 'never', propElementValues: 'always' }],
```
tsapeta: SDK checks fail here too, needs to be formatted

packages/expo-module-scripts/package.json
hassankhan: This will be handled by running `oxfmt` in this PR as a final, pre-merge commit

packages/expo-module-scripts/oxlint.config.base.js
hassankhan: This will be handled by running `oxfmt` in this PR as a final, pre-merge commit

packages/expo-module-scripts/oxlint.config.base.js
hassankhan: This will be handled by running `oxfmt` in this PR as a final, pre-merge commit

byCedric:
@hassankhan are there any vscode extentions we have to install (or would be nice to have) for oxlint/oxfmt? If so, can this be updated too? https://github.com/expo/expo/blob/main/.vscode/extensions.json

kitten (approved):

zoontek (approved):

PR #47097, opened 2026-06-20
[docs] Correct defaultLaunchURL typo in dev-client docs for Android

dmontag23:
The default Android emulator address is 10.0.2.2, see https://developer.android.com/studio/run/emulator-networking-address?_gl=1*1a38ipu*_up*MQ..*_ga*MjYxODQ5NTEzLjE3ODE5NTcyNDY.*_ga_6HH9YJMN9M*czE3ODE5NTcyNDYkbzEkZzAkdDE3ODE5NTcyNDYkajYwJGwwJGgxMDIzMzY0OTE1

# Why

The default Android emulator address has a typo in the docs, it should be `10.0.2.2` instead of `10.0.0.2`.

# How

Only changed the default emulator address.

# Test Plan

Only a doc change.

# Checklist


amandeepmittal (approved):
Thanks for the fix!

PR #47106, opened 2026-06-21
[core] Introduce unified `ArrayBuffer` type

barthap:
# Why

Expo Modules had separate `JavaScriptArrayBuffer` and `NativeArrayBuffer` types, but no single safe module-facing `ArrayBuffer`. This made binary APIs confusing: module authors had to choose between direct JS-runtime-backed access and native-owned/native-retained memory instead of using one preferred type for normal arguments and return values.

# How

- Introduced `ArrayBuffer` as the preferred safe module-facing buffer type on iOS and Android.
- Preserved compatibility: `NativeArrayBuffer` is deprecated, while `JavaScriptArrayBuffer` remains the explicit JS-thread-only zero-copy escape hatch via `JavaScriptObject.getArrayBuffer()`.
- Added explicit owned/borrowed storage semantics. JS-heap `ArrayBuffer` and TypedArray inputs are copied; native-backed ArrayBuffers and TypedArray views borrow retained `MutableBuffer` storage when possible.
- Preserved TypedArray view semantics by exposing only the viewed byte range to native callers.
- On iOS, `NativeArrayBuffer` is now a deprecated typealias to `ArrayBuffer`, and `ArrayBuffer` uses `JavaScriptCodable` for safe encode/decode.
- On Android, added a concrete `ArrayBuffer` HybridClass and conversion/return plumbing while keeping the legacy `NativeArrayBuffer` HybridClass separate for JNI compatibility.
- Added `isOwned` / `isOwned()` and Android `toDirectBuffer(copyBorrowed:)` so callers can distinguish independent storage from retained borrowed views.

This is the first migration step toward one normal `ArrayBuffer` API. Future direction:

- Remove the deprecated `NativeArrayBuffer` compatibility layer once downstream usage has migrated to `ArrayBuffer`.
- Deduplicate Android internals by replacing the separate legacy `NativeArrayBuffer` HybridClass with the new `ArrayBufferStorage` model when JNI compatibility constraints allow it.
- Keep `JavaScriptArrayBuffer` as the explicit zero-copy JS-thread escape hatch, but avoid making it the normal module argument type.
- Explore scoped JS-thread-only zero-copy access on `ArrayBuffer` for performance-sensitive sync paths without exposing long-lived JS-heap pointers.

# Test Plan

- iOS: ExpoModulesCore unit tests; verify owned/borrowed semantics, encode/decode, return conversion, and TypedArray views.
- Android: instrumented tests; verify JS-backed copy semantics, native-backed sharing, TypedArray view offsets, return conversion, and `toDirectBuffer(copyBorrowed:)`.
- Verify existing `NativeArrayBuffer` and `JavaScriptArrayBuffer` compatibility tests still pass.
- Manual NCL testing on Crypto AES

> [!WARNING]
> Android CI will fail due to expo-sqlite breaking changes that I fix in downstack PR #47168. Both PRs must be landed together

# Checklist


barthap:
* **#48028** <a href="https://app.graphite.com/github/pr/expo/expo/48028?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#47262** <a href="https://app.graphite.com/github/pr/expo/expo/47262?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#47261** <a href="https://app.graphite.com/github/pr/expo/expo/47261?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#47168** <a href="https://app.graphite.com/github/pr/expo/expo/47168?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#47106** <a href="https://app.graphite.com/github/pr/expo/expo/47106?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/47106?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/expo-modules-core/ios/Core/ArrayBuffers/NativeArrayBuffer.swift
```diff
@@ -2,152 +2,8 @@
...
-  public static func fromJavaScriptValue(_ value: JavaScriptValue) -> NativeArrayBuffer {
-    fatalError("Creating NativeArrayBuffer from JavaScriptValue is not supported")
-  }
-
-  public func toJavaScriptValue(in runtime: JavaScriptRuntime) -> JavaScriptValue {
-    return asJavaScriptArrayBuffer(runtime: runtime).asValue()
-  }
-}
+@available(*, deprecated, renamed: "ArrayBuffer", message: "Use ArrayBuffer instead.")
+public typealias NativeArrayBuffer = ArrayBuffer
```
tsapeta, line 6: [💡 suggestion] `NativeArrayBuffer` was a released `public final class` (since ~SDK with #46448's zero-copy work) and is now collapsed to `typealias NativeArrayBuffer = ArrayBuffer`. For the normal source-build flow this is fully compatible — all of `allocate`/`copy`/`wrap`/`copy(data:)` survive on `ArrayBuffer`, and the typealias keeps existing call sites compiling (with a deprecation warning).

The one caveat is the prebuilt-xcframework path for `expo-modules-core` (`try_link_with_prebuilt_xcframework` in the podspec): a consumer linked against an *old* core binary that referenced the `NativeArrayBuffer` class symbols would not resolve them against a new binary, since the mangled names move to `ArrayBuffer`. In practice core + consumers rebuild together, so this is low-risk, but given it removes a released public symbol it's worth confirming the prebuilt flow is fine (or tagging `breaking: ABI` if any consumer ships a separately-prebuilt core). Suggestion, not a blocker.

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBuffer.swift
```diff
@@ -45,4 +91,212 @@ public final class ArrayBuffer: AnyArrayBuffer, @unchecked Sendable {
...
+      return ArrayBuffer.from(jsArrayBuffer: jsObject.getArrayBuffer())
+    }
+    let jsValue = jsObject.asValue()
+    guard jsValue.isTypedArray() else {
+      throw ArrayBufferJavaScriptValueConversionException(.object)
+    }
+    return try ArrayBuffer.from(typedArray: jsValue.getTypedArray())
+  }
+
+  private static func from(typedArray: consuming JavaScriptTypedArray) throws -> ArrayBuffer {
```
tsapeta, line 191: [💡 suggestion] `isOwned` ends up meaning different things across the two ways a view can reach native:

- JS-heap-backed typed-array view → copied → `isOwned == true` (storage is independent)
- native-backed typed-array view → borrowed slice → `isOwned == false`

That's internally consistent (owned == "has its own storage"), but as a public API surface `isOwned` doesn't tell a module author whether mutations are visible to JS — both an owned copy and a borrowed *non-shared* buffer read as distinct concepts, while the thing they probably care about is "does writing here alias the JS buffer." Worth a doc comment on the public `isOwned` clarifying it means "storage is independently allocated (writes are not observed by JS)" so callers don't misread it as a general mutability/aliasing flag.

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBuffer.swift
```diff
@@ -45,4 +91,212 @@ public final class ArrayBuffer: AnyArrayBuffer, @unchecked Sendable {
...
+      retained.release()
+    }
+  }
+
+  // MARK: - JavaScript conversion
+
+  /**
+   Converts a JavaScript ArrayBuffer or TypedArray to Expo's safe native ArrayBuffer representation.
+   */
+  internal static func from(jsValue: JavaScriptValue) throws -> ArrayBuffer {
```
tsapeta: [🐧 nit] `from(jsValue:)`/`from(unownedValue:)` only special-case `ArrayBufferJavaScriptValueConversionException` for the "not an ArrayBuffer/TypedArray" case, but the inner `from(jsObject:)` path can also `throw` from `getTypedArray()` / underlying JSI access. In `DynamicArrayBufferType.cast` the new `catch is ...Conversion... { throw NotArrayBufferException } catch { throw error }` correctly rethrows those — just confirming the intent is that a genuine JSI failure surfaces as-is rather than being remapped to `NotArrayBufferException`. If so, the explicit `catch { throw error }` is redundant (a bare `do`/single typed `catch` rethrows automatically); you can drop it for brevity.

packages/expo-modules-core/android/src/main/cpp/ArrayBuffer.h
```diff
@@ -0,0 +1,112 @@
...
+
+  [[nodiscard]] std::shared_ptr<jsi::MutableBuffer> jsiMutableBuffer();
+
+  [[nodiscard]] jni::local_ref<jni::JByteBuffer> toDirectBuffer(bool copyBorrowed);
+
+  [[nodiscard]] bool isOwned();
+
+  template<class T>
+  T read(int position) {
+    return *reinterpret_cast<T *>(storage->data() + position);
```
tsapeta, line 105: [🐧 nit] `read<T>(int position)` does an unchecked `*reinterpret_cast<T*>(storage->data() + position)` with no bounds validation against `storage->size()`. This matches the existing `NativeArrayBuffer`/`JavaScriptArrayBuffer` primitives so it's not a regression, but since `ArrayBuffer` is now the *recommended* type a wider audience will hit these. An out-of-range `position` from Kotlin is an OOB read. Not blocking given parity with the existing types, but worth a follow-up to add a bounds check (or a doc note that callers own the bounds contract).

packages/expo-modules-core/android/src/main/java/expo/modules/kotlin/jni/NativeArrayBuffer.kt
```diff
@@ -9,9 +9,10 @@ import java.nio.ByteBuffer
  * A Kotlin representation of native-owned ArrayBuffer.
  * Can be created on any thread and safely returned to JavaScript.
  */
+@Deprecated("Use ArrayBuffer instead", ReplaceWith("ArrayBuffer"))
```
tsapeta, line 12: [🏗️ **design**] Since `NativeArrayBuffer` no longer implements the (now-class) `ArrayBuffer`, `copyOf(other: ArrayBuffer)` here takes the concrete `ArrayBuffer` class rather than the old interface. Combined with the new `copyOf(other: JavaScriptArrayBuffer)` / `copyOf(other: NativeArrayBuffer)` overloads, the type that used to be accepted polymorphically (any `ArrayBuffer`) now needs an explicit overload per concrete type — which is exactly the source of the expo-sqlite breakage noted in the PR description. Just flagging that any out-of-tree consumer that declared `fun foo(buf: ArrayBuffer)` expecting the old interface will break the same way; the deprecation on `NativeArrayBuffer` doesn't cover that interface-removal. Worth calling out explicitly in the CHANGELOG as a breaking change for Android, not just a deprecation.

packages/expo-modules-core/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 - [iOS] Added `Module.emit` that sends an event directly to the module's own JavaScript object, mirroring `SharedObject.emit`. Both now share an `EventEmitter` protocol. ([#46555](https://github.com/expo/expo/pull/46555) by [@tsapeta](https://github.com/tsapeta))
 - Add `useReleasingSharedObjectWithLifecycle` hook. ([#46494](https://github.com/expo/expo/pull/46494) by [@behenate](https://github.com/behenate))
 - [iOS] Added `SharedObject.native(from:)` that recovers the native shared object paired with a JavaScript object, with a generic overload that returns the concrete subclass directly. ([#47054](https://github.com/expo/expo/pull/47054) by [@tsapeta](https://github.com/tsapeta))
+- Added `ArrayBuffer` as the preferred safe native module argument and return type, and deprecated `NativeArrayBuffer` in favor of it. ([#47106](https://github.com/expo/expo/pull/47106) by [@barthap](https://github.com/barthap))
```
tsapeta, line 20: [💡 suggestion] This entry is under the "🎉 New features" section and frames the change as additive ("Added ArrayBuffer ... deprecated NativeArrayBuffer"). On Android this is also a source-breaking change: the `ArrayBuffer` interface became a class, so `NativeArrayBuffer`/`JavaScriptArrayBuffer` no longer share a common `ArrayBuffer` supertype (the reason #47168 is required). Consider adding a "🛠 Breaking changes" entry for the Android interface→class removal so downstream native-module authors aren't surprised by a compile break that the `@Deprecated` annotation alone doesn't convey.

packages/expo-modules-core/ios/Tests/ArrayBufferTests.swift
```diff
@@ -5,7 +5,7 @@ import Testing
 
 @testable import ExpoModulesCore
 
-@Suite("NativeArrayBuffer")
+@Suite("ArrayBuffer")
```
tsapeta, line 8: [🐧 nit] The suite display name is updated to `@Suite("ArrayBuffer")` but the struct is still `NativeArrayBufferTests` and the file is `NativeArrayBufferTests.swift`. Since `NativeArrayBuffer` is now just a deprecated alias and these tests exercise `ArrayBuffer`, renaming the file/struct to `ArrayBufferTests` would avoid future confusion (and the new `JavaScriptCodableArrayBufferTests.swift` already uses the new name). Minor.

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBuffer.swift
tsapeta: Is it owned by whom/what? Definitely need to make it more clear.

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBuffer.swift
tsapeta: I think it's better to avoid `js` prefix if the type already says so and there is no conflict with the other "value". Moreover, it should be consistent with `from(unownedValue:)` and `from(typedArray:)`.

```suggestion
  internal static func from(value: JavaScriptValue) throws -> ArrayBuffer {
```

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBuffer.swift
```diff
@@ -45,4 +91,212 @@ public final class ArrayBuffer: AnyArrayBuffer, @unchecked Sendable {
...
+    }
+    return ArrayBuffer.copy(of: UnsafeRawPointer(jsArrayBuffer.data()), count: jsArrayBuffer.size)
+  }
+}
+
+extension ArrayBuffer: JavaScriptDecodable, JavaScriptEncodable {
+  // MARK: - JavaScriptDecodable
+  @JavaScriptActor
+  public static func decode(
+    _ value: JavaScriptValue,
```
tsapeta: Since #47178, the value needs to be borrowed so we avoid retain/release.

```suggestion
    _ value: borrowing JavaScriptValue,
```

tsapeta: Also make these functions `@inlinable` if possible. Otherwise they won't be optimized by the compiler when `expo-modules-core` is consumed from a prebuilt xcframework.

packages/expo-modules-core/ios/Tests/NativeArrayBufferTests.swift
```diff
@@ -268,6 +273,106 @@ struct NativeArrayBufferTests {
       #expect(processedValues.allSatisfy { $0 == 99 } == true)  // Processed has new pattern
     }
 
+    @Test
+    func `copies JS-backed ArrayBuffer when using ArrayBuffer argument`() throws {
+      let processedBuffer = try runtime.eval([
```
tsapeta: `eval(lines: [String])` is deprecated. Use multi-line string literal (`"""`).

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBuffer.swift
```diff
@@ -3,35 +3,81 @@
 import ExpoModulesJSI
 
 /**
- Represents a fixed-length raw binary data buffer backed by a `JavaScriptArrayBuffer`.
- Provides access to the underlying memory owned by the JavaScript runtime.
+ An array buffer that manages its own native memory or borrows native-backed memory.
+ Does not require a JavaScript runtime at creation time — the JSI backing buffer
+ is created on demand when the buffer needs to be returned to JavaScript.
+
+ - Note: Sendable conformance is `@unchecked` because `UnsafeMutableRawPointer` isn't `Sendable`.
```
tsapeta: swift-format is not enabled in `expo-modules-core` yet, but it'll enforce using `///` for doc comments (it's more common in Swift). I think it's good to already use `///` in the code that we touch to reduce the amount of changes when we enable the formatter.

tsapeta (changes requested):

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBuffer.swift
barthap: Clarified this in the `isOwned` doc comment.

The meaning is: owned by this `ArrayBuffer` storage itself, i.e., visible byte range is independent native storage, not a retained borrowed view into another/native-backed JSI `MutableBuffer`.

So:
- JS-heap `ArrayBuffer` / `TypedArray` inputs are copied -> `isOwned == true`.
- Native-backed `ArrayBuffer` / `TypedArray` inputs can borrow retained backing storage -> `isOwned == false`.

packages/expo-modules-core/android/src/main/cpp/ArrayBuffer.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+jni::local_ref<ArrayBuffer::jhybriddata>
+ArrayBuffer::initHybrid(jni::alias_ref<JavaPart::javaobject>,
+                        jni::alias_ref<jni::JByteBuffer> byteBuffer) {
+  return makeCxxInstance(byteBuffer);
+}
+
+jni::local_ref<ArrayBuffer::javaobject>
+ArrayBuffer::newInstance(JSIContext *jsiContext, jsi::Runtime &runtime,
+                         jsi::ArrayBuffer &arrayBuffer) {
```
lukmccall:
```suggestion
jni::local_ref<ArrayBuffer::javaobject> ArrayBuffer::newInstance(
  JSIContext *jsiContext, 
  jsi::Runtime &runtime,
  jsi::ArrayBuffer &arrayBuffer
) {
```

packages/expo-modules-core/android/src/main/cpp/ArrayBuffer.cpp
```diff
@@ -0,0 +1,162 @@
...
+  memcpy(byteBuffer->getDirectAddress(), arrayBuffer.data(runtime), size);
+
+  auto value = ArrayBuffer::newObjectCxxArgs(byteBuffer);
+  jsiContext->jniDeallocator->addReference(value);
+  return value;
+}
+
+jni::local_ref<ArrayBuffer::javaobject>
+ArrayBuffer::newInstance(JSIContext *jsiContext, jsi::Runtime &runtime,
+                         expo::TypedArray& typedArray) {
```
lukmccall:
```suggestion
jni::local_ref<ArrayBuffer::javaobject> ArrayBuffer::newInstance(
  JSIContext *jsiContext, 
  jsi::Runtime &runtime,
  expo::TypedArray& typedArray
) {
```

lukmccall (approved):
Android looks good!

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBuffer.swift
```diff
@@ -2,36 +2,84 @@
...
+/// Does not require a JavaScript runtime at creation time — the JSI backing buffer
+/// is created on demand when the buffer needs to be returned to JavaScript.
+///
+/// - Note: Sendable conformance is `@unchecked` because `UnsafeMutableRawPointer` isn't `Sendable`.
 public final class ArrayBuffer: AnyArrayBuffer, @unchecked Sendable {
-  /**
-   The underlying JSI array buffer that manages the memory and JS runtime reference.
-   */
-  internal let backingBuffer: JavaScriptArrayBuffer
+  private enum Storage {
```
tsapeta: Maybe we can put the enum at the end of the class as it's private and not the most important part. `private let storage: Storage` at the top is fine though.

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBuffer.swift
```diff
@@ -2,36 +2,84 @@
...
 
-/**
- Represents a fixed-length raw binary data buffer backed by a `JavaScriptArrayBuffer`.
- Provides access to the underlying memory owned by the JavaScript runtime.
- */
+/// An array buffer that manages its own native memory or borrows native-backed memory.
+/// Does not require a JavaScript runtime at creation time — the JSI backing buffer
+/// is created on demand when the buffer needs to be returned to JavaScript.
+///
+/// - Note: Sendable conformance is `@unchecked` because `UnsafeMutableRawPointer` isn't `Sendable`.
```
tsapeta: We might be able to check it by making the pointer nonisolated and expose it through a function instead of a property 🤔

tsapeta: Can also investigate if we can make `ArrayBuffer` a struct, maybe even non-copyable for better performance and ownership.

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBuffer.swift
```diff
@@ -2,36 +2,84 @@
...
+  /// JavaScript can still be shared with that new JavaScript wrapper.
+  public var isOwned: Bool {
+    return storage.isOwned
+  }
+
+  init(wrapping data: UnsafeMutableRawPointer, count: Int, cleanup: @escaping () -> Void) {
+    self.storage = .owned(pointer: data, count: count, cleanup: cleanup)
+  }
+
+  init(borrowing data: UnsafeMutableRawPointer, count: Int, cleanup: @escaping () -> Void) {
```
tsapeta, line 34: I still feel like it's a bit unclear what's the difference between `borrowing` and `wrapping`. Or, in other words, `wrapping` doesn't sound like it's the "owned" case.

Maybe, for the owned storage, static `allocate` functions should become inits?

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBuffer.swift
barthap: Nonisolated/function didn't work that easily. We must store the raw pointer somewhere, so we can't get rid of `@unchecked` entirely, but I managed to encapsulate it into a private wrapper, so `ArrayBuffer` can now be checked.

Regarding struct: I'd keep it as a `final class`. It behaves like a reference-owning buffer handle: it owns/retains native memory, exposes mutable byte access, and is captured by the JS ArrayBuffer finalizer to keep the memory alive.
Making it a copyable `struct` would make it weird: copies would look value-like while sharing the same mutable bytes.
A non-copyable struct would better express unique ownership, but it would be a big API/ergonomics change, and I'm afraid it would make usage more awkward (for module arguments, records, arrays, and Codable conversions). I'll try it out, but rather in a follow-up PR

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBuffer.swift
barthap: > I still feel like it's a bit unclear what's the difference between borrowing and wrapping. Or, in other words, wrapping doesn't sound like it's the "owned" case.

I agree, this sounds confusing. I'd replace `wrapping` with `owning`, so the two cases read as “own native storage” vs “borrow native-backed storage”.

The final model will be three cases (I rename the storage enum variants to the parentheses):
1. (`ownedNative`) owned native storage: memory allocated/retained by this `ArrayBuffer`;
2. (`nativeBacked`) native-backed borrowed storage: memory comes from a retained JSI `MutableBuffer`, safe to access from native threads, but not allocation-owned by this object;
3. (`jsBacked`) JS-backed borrowed storage: memory belongs to the JS heap and is only safe through scoped JS-thread access.


For the sake of this PR, I kept only the first two cases to keep parity with the previous `NativeArrayBuffer` capabilities and avoid mixing in the scoped JS-backed behavior here. In the follow-up PR (#47262), where JS-backed scoped access is added, I think the `Storage` enum will make it clearer.

> Maybe, for the owned storage, static allocate functions should become inits?

Not a bad idea, I'll do it. The only place where we use `ArrayBuffer.allocate()` is `expo-sqlite`, so I'll update the upstack PR

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBuffer.swift
barthap: Decided to rename this to `isNativeBacked` - it will be more useful from module author's perspective, because it tells "is this memory safe to access outside JS thread". Two current cases resolve to `true`, but in #47262 the JS case will be `false`

tsapeta (approved):

barthap:
### Merge activity

* **Jun 30, 9:23 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/expo/expo/47106).
* **Jun 30, 9:23 AM UTC**: @barthap merged this pull request with [Graphite](https://app.graphite.com/github/pr/expo/expo/47106).

PR #47107, opened 2026-06-21
[ios][core] Bind `@SharedObject` `@JS` members directly into JS

tsapeta:
# Why

`@SharedObject` binds its `@JS` members directly onto the class prototype (the direct-JSI path), mirroring what `@ExpoModule` does for modules. A `@SharedObject`'s `@JS` methods, properties and `@JS init` no longer need `Function(…)` / `Property(…)` / `Constructor { … }` definition entries.

# How

`expo-modules-core` consumes `@expo/expo-modules-macros-plugin` 0.6.1. The macro emits `_decorateSharedObject` (binds `@JS` methods/properties onto the prototype, recovering the per-instance receiver from `this`) and `_constructSharedObject` (builds the native instance from the JS constructor arguments) as `override class func`s.

These live on the base `SharedObject` class, **not** as `AnySharedObject` protocol requirements: every `@SharedObject` inherits the conformance from the base, which locks the witness table to the base, so a subclass's static method is never reached through protocol-witness dispatch. `class func` + `override` gives the per-subclass dispatch, and non-macro classes fall through to the base defaults.

`ClassDefinition.build()` recovers the concrete `SharedObject` subclass from the associated `DynamicSharedObjectType` and calls `_constructSharedObject` in the constructor closure and `_decorateSharedObject` on the prototype.

# Test Plan

`MacroSharedObjectTests` (run via `et native-unit-tests -p ios --packages expo-modules-core`) covers a `@SharedObject` exposed through `@ExpoModule(classes:)`:

- class exposure and the `@SharedObject("Name")` override;
- construction via `@JS init`; `@JS` method binding with per-instance native mutation across calls;
- gettable and settable `@JS` properties;
- a value-returning method with an omittable trailing optional;
- non-primitive decode/encode across `@JS` methods: `@Record`, optional return (value and `null`), and another `SharedObject`;
- a throwing `@JS` method surfacing a coded JS error.

### DIFF
--- a/packages/expo-modules-core/ios/Core/Classes/ClassDefinition.swift
+++ b/packages/expo-modules-core/ios/Core/Classes/ClassDefinition.swift
@@ -51,17 +51,20 @@ public final class ClassDefinition: ObjectDefinition {
 
   @JavaScriptActor
   public override func build(appContext: AppContext) throws -> JavaScriptObject {
+    // The concrete `SharedObject` subclass backing this class, used to dispatch to the
+    // `@SharedObject` macro's `_constructSharedObject` / `_decorateSharedObject` overrides.
+    let sharedObjectType = ((associatedType as? DynamicSharedObjectType)?.innerType) as? SharedObject.Type
+
     let constructorClosure: JavaScriptRuntime.SyncFunctionClosure = { [weak self, weak appContext] this, arguments in
       guard let self, let appContext else {
         throw Exceptions.AppContextLost()
       }
-      // Call the native constructor when defined. Any thrown `Exception` (which
-      // conforms to `JavaScriptThrowable`) propagates to the host function closure,
-      // which converts it to a JS `Error` preserving `message` and `code`.
-      if let constructor {
-        // Run the body natively so we can pair a returned `SharedObject` with `this`
-        // directly. `this` was created by `new` with the correct class prototype; returning
-        // a different JS object would replace `this` and drop that prototype chain.
+      // Build the native instance and pair it with `this` (returning a different object would drop
+      // the prototype chain `new` set up). Prefer the macro-synthesized `_constructSharedObject`,
+      // falling back to the DSL `Constructor` body.
+      if let sharedObject = try sharedObjectType?._constructSharedObject(this: this, arguments: arguments, in: appContext.runtime) {
+        appContext.sharedObjectRegistry.add(native: sharedObject, javaScript: this.getObject())
+      } else if let constructor {
         let result = try constructor.runBody(appContext, in: appContext.runtime, this: this, arguments: arguments)
 
         if let sharedObject = result as? SharedObject {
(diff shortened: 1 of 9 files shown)


packages/expo-modules-core/ios/Core/Classes/ClassDefinition.swift
```diff
@@ -51,17 +51,20 @@ public final class ClassDefinition: ObjectDefinition {
 
   @JavaScriptActor
   public override func build(appContext: AppContext) throws -> JavaScriptObject {
+    // The concrete `SharedObject` subclass backing this class, used to dispatch to the
+    // `@SharedObject` macro's `_constructSharedObject` / `_decorateSharedObject` overrides.
+    let sharedObjectType = ((associatedType as? DynamicSharedObjectType)?.innerType) as? SharedObject.Type
```
tsapeta, line 56: I know it's not the best approach, but it's temporary. The goal is that `@SharedObject` macro doesn't use `ClassDefinition` and dynamic types at all. They will be registered directly in the synthesized code.

packages/expo-modules-core/ios/Core/SharedObjects/SharedObject.swift
```diff
@@ -83,6 +83,20 @@ open class SharedObject: AnySharedObject {
...
+  /// Binds the class's `@JS` members onto its JavaScript prototype. Overridden by the `@SharedObject`
+  /// macro; a `class func` (not a protocol requirement) so dispatch resolves per subclass.
+  @JavaScriptActor
+  open class func _decorateSharedObject(prototype: borrowing JavaScriptObject, in runtime: JavaScriptRuntime) throws {}
+
+  /// Builds a native instance from the JS constructor arguments, or `nil` to use the DSL `Constructor`
+  /// path. Overridden by the `@SharedObject` macro from the class's `@JS init`.
+  @JavaScriptActor
+  open class func _constructSharedObject(this: JavaScriptValue, arguments: borrowing JavaScriptValuesBuffer, in runtime: JavaScriptRuntime) throws -> SharedObject? {
+    return nil
```
tsapeta, line 98: In theory this is not correct, it should return `Self()`, but it's currently not possible in subclasses.
`init()` is not `required` so there is no guarantee that subclasses provide this initializer. Adding this keyword to the base class would be a breaking change for the existing modules.
To fix this, the macro should make sure there is at least one constructor with `@JS` macro or emit `public init() {}` if not.

packages/expo-modules-core/ios/Core/Classes/ClassDefinition.swift
```diff
@@ -75,6 +78,12 @@ public final class ClassDefinition: ObjectDefinition {
 
     try decorate(object: klass, appContext: appContext)
 
+    // Bind the macro's `@JS` members onto the prototype.
+    if let sharedObjectType {
+      let prototype = klass.getProperty("prototype").getObject()
+      try sharedObjectType._decorateSharedObject(prototype: prototype, in: appContext.runtime)
```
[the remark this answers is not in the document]
tsapeta: @codex review again

Kudo (approved):

PR #47110, opened 2026-06-21
refactor(ncl): modernize the `withGradleProperties` and `withSettingsImport` plugins

byCedric:
# Why

I was looking for good examples for the `withGradleProperties` config plugin, and spotted this in NCL. Modernized it a bit so it can function as an example.

# How

- Swapped `@expo/config-plugins` for `expo/config-plugins`
- Swapped `withDangerousMod` for `withGradleProperties`
- Replace existing property node or append new property node at the end

# Test Plan

- `cd ./apps/native-component-list`
- `pnpm expo prebuild -p android`
- Run prebuild a few times, while modifying **./apps/native-component-list/app.config.js** with different values
  - There should be no duplicates
  - All changes should be reflected properly 

# Checklist

### DIFF
--- a/apps/native-component-list/plugins/withGradleProperties.js
+++ b/apps/native-component-list/plugins/withGradleProperties.js
@@ -1,42 +1,24 @@
-const { withDangerousMod } = require('@expo/config-plugins');
-const assert = require('assert');
-const fs = require('fs/promises');
-const path = require('path');
+const { withGradleProperties } = require('expo/config-plugins');
 
-// Use this improve gradle builds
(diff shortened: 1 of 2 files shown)


byCedric:
Ugh, its still using the old Calendar API causing some linting issues. This is an unrelated clean up task we have to do.

kitten (approved):

kitten:
clean up here is to drop the `import/*` rules. But this isn't inheriting from the same eslint config but from `eslint-config-universe`

Issue #39752, opened 2025-09-17
RTL not working when switch to another RTL language


gor3a:
### Minimal reproducible example

https://github.com/gor3a/tech-art

### Steps to reproduce

# RTL Language Switching Issue Report

## Title
`I18nManager.allowRTL(true)` and `I18nManager.forceRTL(true)` don't change app direction to RTL when switching languages dynamically

## Description

When dynamically switching to an RTL language (Arabic) in an Expo app, calling `I18nManager.allowRTL(true)` and `I18nManager.forceRTL(true)` doesn't change the app's layout direction to RTL. The text direction and layout remain LTR despite the RTL settings being applied.

## Environment

- **Expo SDK**: 54.0.8
- **React Native**: 0.81.4
- **Platform(s)**: iOS, Android, Web
- **Environment(s)**: Expo Go, development build, standalone app
- **Package Manager**: npm
- **Node.js**: 18+

### 1. Create a new Expo project

```bash
npx create-expo-app RTLTestApp
cd RTLTestApp
npm install i18n-js expo-localization @react-native-async-storage/async-storage
```

### 2. Create the following file structure

```
RTLTestApp/
├── locales/
│   ├── en.json
│   └── ar.json
├── i18n.js
└── App.js
```

### 3. Add translation files

**`locales/en.json`:**
```json
{
  "welcome": "Welcome",
  "switchLanguage": "Switch to Arabic",
  "testText": "This is a test text to see RTL behavior"
}
```

**`locales/ar.json`:**
```json
{
  "welcome": "مرحبا",
  "switchLanguage": "التبديل إلى الإنجليزية",
  "testText": "هذا نص تجريبي لرؤية سلوك RTL"
}
```

### 4. Create `i18n.js` file

```javascript
import { I18n } from 'i18n-js';
import * as Localization from 'expo-localization';
import { I18nManager } from 'react-native';

import en from './locales/en.json';
import ar from './locales/ar.json';

const i18n = new I18n();
i18n.translations = { en, ar };
i18n.locale = Localization.getLocales()[0].languageCode ?? 'en';
i18n.enableFallback = true;
i18n.defaultLocale = 'en';

const RTL_LANGUAGES = ['ar', 'he', 'fa', 'ur'];

export const changeLanguage = (languageCode) => {
  i18n.locale = languageCode;
  
  // Handle RTL for Arabic and other RTL languages
  const isRTLLanguage = RTL_LANGUAGES.includes(languageCode);
  
[...]

PR #47115, opened 2026-06-21
docs: fix `textDirection` → `textAlign` in localization RTL text alignment section

gor3a:
## Why

The **Text alignment** section of the Localization guide refers to a `textDirection` style property, but the property it describes (and uses in the accompanying code samples) is `textAlign`. `textDirection` is not a React Native style prop, so the prose is incorrect and confusing.

The behavior described is correct — under RTL, `textAlign: 'left'` acts as the start edge — only the property name was wrong.

## What

In the "Text alignment" section only, rename the prose references `textDirection` → `textAlign`. The legitimate `getLocales()[0].textDirection` (locale field) and the Firefox `textDirection` browser note are left untouched.

No code/behavior change — docs only.

## Related

- Issue: #39752 (RTL behavior confusion)
- Background on the `textAlign` vs `textDirection` naming and RTL semantics: [this comment](https://github.com/expo/expo/issues/39752#issuecomment-4762160953)

### DIFF
--- a/docs/pages/guides/localization.mdx
+++ b/docs/pages/guides/localization.mdx
@@ -327,9 +327,9 @@ return <View dir={getLocales()[0].textDirection || 'ltr'}>...</View>;
 
 ### Text alignment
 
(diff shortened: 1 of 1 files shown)


amandeepmittal (approved):
Thanks!

PR #47117, opened 2026-06-21
[expo-router] Fix withAnchor seeding a param-less anchor for a dynami…

### LINKED ISSUE #47114: expo-router: Stack with a single dynamic route ([param]) as its initial screen renders an empty param-less instance as the anchor

https://github.com/Elalfy74/withAnchor-dynamic-route

### Steps to reproduce

When a nested Stack layout declares only a dynamic route (e.g. [userId]) as its initial/only screen, navigating to that route with params causes the navigator to first instantiate the dynamic screen without params (as the stack anchor), then push the real screen with params on top.

The result is a phantom screen with empty params sitting underneath. Pressing back (or router.dismiss()) pops the real screen and lands on the empty-param anchor instead of the parent screen.

A sibling stack whose initial screen is a static route (e.g. security) behaves correctly, which makes the dynamic-route case look like a bug.

https://github.com/user-attachments/assets/84200eaa-98f0-4850-9be9-aff6822fcada

Reproduction
File structure:

```
app/
  (tabs)/
    social/
      _layout.tsx          // Stack: index + users
      index.tsx            // list, links to /social/users/1
      users/
        _layout.tsx        // Stack with ONLY name="[userId]"
        [userId].tsx       // reads useLocalSearchParams()
social/users/_layout.tsx:
```

```
import { Stack } from "expo-router";

export default function UsersLayout() {
  return (
    <Stack screenOptions={{ headerShown: false }}>
      <Stack.Screen name="[userId]" />
    </Stack>
  );
}
social/users/[userId].tsx:
```

```
import { router, useLocalSearchParams } from "expo-router";
import { Pressable, Text, View } from "react-native";

export default function UserDetail() {
  const { userId } = useLocalSearchParams<{ userId: string }>();
  return (
    <View style={{ flex: 1, justifyContent: "center", alignItems: "center" }}>
      <Text>userId: {userId}</Text>
      <Pressable onPress={() => router.back()}>
        <Text>Back</Text>
      </Pressable>
    </View>
  );
}
```
Steps:

Navigate to /social/users/1.
Observe the detail screen shows userId: 1 ✅.
Press Back.
Current behavior
Back lands on a second [userId] screen rendering userId: (empty) instead of returning to /social (the list). The dynamic route was instantiated twice: once param-less as the stack's anchor, once with userId=1.

Expected behavior
Back should return to the parent screen (/social). The stack should not create an extra param-less instance of its only dynamic route.

Comparison that works (static initial route)
An equivalent nested stack whose only screen is static works as expected:

```
profile/settings/
  _layout.tsx    // Stack with ONLY name="security"
  security.tsx
```
Navigating to /profile/settings/security creates exactly one instance, and back returns to the parent correctly. The only difference between the working and broken cases is static vs. dynamic route name for the stack's initial screen

### Environment

```text
    System:
      OS: macOS 26.5.1
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 25.9.0 
      npm: 11.12.1 
    Managers:
      CocoaPods: 1.16.2 
    SDKs:
      iOS SDK:
    IDEs:
      Android Studio: 2025.3 AI-253.32098.37.2534.15336583
    npmPackages:
      expo: ^56.0.12 => 56.0.12 
      expo-router: ~56.2.11 => 56.2.11 
      react: 19.2.3 => 19.2.3 
      react-dom: 19.2.3 => 19.2.3 
    npmGlobalPackages:
      eas-cli: 18.11.0
    Expo Workflow: bare
```

### Expo Doctor Diagnostics

```text

✖ Check Expo config (app.json/ app.config.js) schema
Errors validating fields in /my-app/app.json:
 should NOT have additional property 'newArchEnabled'.
 Field: android - should NOT have additional property 'edgeToEdgeEnabled'.
Advice:
Resolve schema errors in your app config. Learn more: https://docs.expo.dev/workflow/configuration/

✖ Check that @react-navigation packages are not installed alongside expo-router
As of SDK 56, expo-router is no longer compatible with react-navigation. The following @react-navigation packages are installed as direct dependencies and should be removed: "@react-navigation/bottom-tabs", "@react-navigation/elements", "@react-navigation/native".
Advice:
Remove these packages from your package.json and replace any direct `@react-navigation/*` imports in your code with their expo-router equivalents. See https://docs.expo.dev/router/migrate/sdk-55-to-56/ for more information.

```

Elalfy74:
Why

Fixes #47114. Pushing withAnchor into a nested stack whose only screen is a dynamic route (e.g. [userId]) instantiated the route twice — once param-less as an anchor, once with params — so Back landed on the empty screen instead of the parent.

While fixing it, a closely related case surfaced: pushing withAnchor to a route nested under a dynamic segment (e.g. /posts/1/details, where posts/[id]/index is the anchor) seeded the anchor without the dynamic param, so Back landed on a param-less /posts/1.

How

withAnchor set initial: false on every nested level, making React Navigation seed each navigator's initial route as an anchor. Two problems followed:

Self-anchor (the param-less duplicate). When a navigator's initial route is the same (dynamic) route being navigated to, the seeded anchor is a param-less duplicate. The fix resolves each navigator's effective initial route (explicit initialRouteName, else the runtime-sorted first child via sortRoutesWithInitial, matching routeNames[0]) and skips initial: false for such self-anchor levels — mirroring the rule findInitialRoute already applies in getStateFromPath. Legitimate anchors (a distinct initial route, e.g. index in an index + [id] stack) are still seeded.
Anchors under a dynamic segment lost their params. When an anchor is seeded (initial: false), React Navigation's getInitialState only copies params to the navigated screen, so a seeded anchor under a dynamic segment (e.g. [id]) never received the dynamic param. useNavigationBuilder now propagates the navigator route's loose path params (everything except the reserved screen/params/initial/state keys) to the seeded initial route. This is narrowly gated to the initial: false seeding path, so normal navigation is unaffected.
Test Plan

Added regression tests in issues.test.ios.tsx:

Reproduces #47114 — a dynamic-only nested stack pushed with withAnchor instantiates the screen exactly once and Back returns to the parent.
Guard — a multi-screen stack ([id] declared before index) still seeds its real runtime initial route (index) as the anchor.
A seeded anchor nested under a dynamic segment (/posts/1/details) keeps its dynamic param, so Back lands on /posts/1 with id intact.
Plus a unit test in getNavigationAction.test.ios.ts for the self-anchor skip.

pnpm test, pnpm build:lib, and pnpm lint pass.

### DIFF
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -17,7 +17,7 @@
 - Allow async routes to rehydrate synchronously by carrying through preloaded modules preventing FOUC in production output ([#46539](https://github.com/expo/expo/pull/46539) by [@kitten](https://github.com/kitten))
 - Fix type inference for `unstable_createStandardRouterNavigator` ([#46737](https://github.com/expo/expo/pull/46737) by [@Ubax](https://github.com/Ubax))
 - Preserve a system time mocked with `jest.setSystemTime` across `renderRouter` ([#46978](https://github.com/expo/expo/pull/46978) by [@Ubax](https://github.com/Ubax))
-- Fix `withAnchor` navigation seeding a duplicate, param-less screen when a nested stack's only (initial) route is a dynamic route (e.g. `[userId]`). ([#XXXXX](https://github.com/expo/expo/pull/47117) by [@Elalfy74](https://github.com/Elalfy74))
+- Fix `withAnchor` navigation seeding bad anchors for dynamic routes: no longer creates a param-less duplicate when a nested stack's only (initial) route is a dynamic route (e.g. `[userId]`), and a seeded anchor nested under a dynamic segment now keeps its dynamic params. ([#47117](https://github.com/expo/expo/pull/47117) by [@Elalfy74](https://github.com/Elalfy74))
 
 ### 💡 Others
 
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -17,6 +17,7 @@
 - Allow async routes to rehydrate synchronously by carrying through preloaded modules preventing FOUC in production output ([#46539](https://github.com/expo/expo/pull/46539) by [@kitten](https://github.com/kitten))
 - Fix type inference for `unstable_createStandardRouterNavigator` ([#46737](https://github.com/expo/expo/pull/46737) by [@Ubax](https://github.com/Ubax))
 - Preserve a system time mocked with `jest.setSystemTime` across `renderRouter` ([#46978](https://github.com/expo/expo/pull/46978) by [@Ubax](https://github.com/Ubax))
+- Fix `withAnchor` navigation seeding a duplicate, param-less screen when a nested stack's only (initial) route is a dynamic route (e.g. `[userId]`). ([#XXXXX](https://github.com/expo/expo/pull/XXXXX) by [@Elalfy74](https://github.com/Elalfy74))
 
 ### 💡 Others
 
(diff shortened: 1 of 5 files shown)


brentvatne:
@expo-bot check

chrfalch:
Already fixed by #48783; patched code removed from package

PR #44087, opened 2026-03-19
[workspace] cleanup package.json files, reorder dependencies

Simek:
# Why

Run https://github.com/Thinkmill/manypkg check on the monorepo, address part of the warnings which are safe to tackle, and should not be controversial. 

Screenshot 2026-03-19 at 17 58 11

# How

Summary of changes:
* remove empty dependency/devDependency objects
* sort dependencies alphabetically
* remove packages from devDependencies which are already defined in dependencies
* use only devDependencies at workspace root, since dependencies are disallowed, and there is no difference between the dependency categories in the main monorepo file
* use short `bin` definition for cases where binary name matches package name - https://docs.npmjs.com/cli/v10/configuring-npm/package-json#bin

# Test Plan

`yarn et cp` checks on modified packages are passing. Verified that there are no side-effects on the packages due to the performed `package.json` changes.

PR #47129, opened 2026-06-22
chore(router,brownfield-tester,test-suite): Align deps

kitten:
Minor deps alignment of:
- react-native-safe-area-context
- typescript


hassankhan (approved):

hassankhan:
Semi-related, we should probably revive #44087 and #45089

PR #47137, opened 2026-06-22
[expo-doctor] Fix local project module loading in bundled builds

entiendoNull:
# Why

Screenshot 2026-06-22 at 13 41 16

Built and ran `expo-doctor` locally against a fresh project and got these failures.

- `Cannot find module '<project>/node_modules/expo/metro-config.js'`
- `The package "expo-modules-autolinking" is out-of-date and isn't available to check dependency conflicts.`

The published version was working fine.

# How

Use `Module.createRequire(...)` when `expo-doctor` loads modules from the project it is checking, instead of dynamic `require(...)`. The local bundled build rewrote those dynamic project module imports in a way that made them fail at runtime.

Also snuck this in; removed the stale `pnpm watch` instruction from the `expo-doctor` README.

# Test Plan

# Checklist

Tests passed.

Also made manual tests:
- Ran against fresh project (`bun create expo --template default@sdk-56`)
- No errors; all checks passed
- Introduced errors; accurate warnings/errors displayed
- Compared results with `npx expo-doctor@latest`

### DIFF
--- a/packages/expo-doctor/src/utils/autolinkingResolutions.ts
+++ b/packages/expo-doctor/src/utils/autolinkingResolutions.ts
@@ -1,4 +1,6 @@
 import type { DependencyResolution } from 'expo-modules-autolinking/exports';
+import * as Module from 'node:module';
+import path from 'path';
 import resolveFrom from 'resolve-from';
 
 import type { VersionedNativeModuleNamesCache } from './versionedNativeModules';
@@ -17,7 +19,9 @@ export function importAutolinkingExportsFromProject(
     );
   }
   try {
-    const mod = require(autolinkingExportsResolved);
+    const mod = Module.createRequire(path.join(projectDir, 'package.json'))(
+      autolinkingExportsResolved
+    );
     if (
       typeof mod.makeCachedDependenciesLinker !== 'function' ||
       typeof mod.scanDependencyResolutionsForPlatform !== 'function'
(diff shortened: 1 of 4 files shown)


packages/expo-doctor/src/utils/metroConfigLoader.ts
```diff
@@ -17,21 +18,21 @@ function importMetroConfigFromProject(
...
   }
+  const projectRequire = Module.createRequire(path.join(projectDir, 'package.json'));
   try {
     // NOTE(@kitten): We need to use the version of metro-config that Expo uses
     // Luckily, we can import `@expo/metro` via `expo` to get to the same version
-    const expoMetro = require.resolve('@expo/metro/metro-config', {
-      paths: [path.dirname(expoResolved)],
-    });
-    return require(expoMetro);
+    const expoRequire = Module.createRequire(expoResolved);
```
kitten, line 25: This is in theory not necessary on resolved files. If we can't resolve a file that's maybe more an indication that we can't load the internal file we're after?

This import here is already a fallback and shouldn't be used directly in favor of the actual config loader, so in theory, we shouldn't fall through to here

packages/expo-doctor/src/utils/autolinkingResolutions.ts
kitten, line 12: If this fails, we should probably check the `expo/package.json:exports` list against this

PR #47139, opened 2026-06-22
[cli] fix: use loadModule in MCP and DevTools

krystofwoldrich:
- `loadModule` async import was transformed and thus now working for top level async esm modules
- `loadModule` was not used in MCP and DevTools which had custom `add-module.js` impl of the async import
- note that in modern Node require loads esm with exceptions for example top level async

## Tests

- updated devtools test to use async esm to ensure it's loaded correctly

### DIFF
--- a/apps/router-e2e/modules/devtools-e2e/server.mjs
+++ b/apps/router-e2e/modules/devtools-e2e/server.mjs
@@ -1,3 +1,8 @@
+
+// Top-level async to ensure esm modules are loaded correctly
+// require of esm in modern Node doesn't throw error (it "just" works)
+await asyncLog('E2E DevTools Server started.');
+
 export default async function handler(request) {
   const url = new URL(request.url);
 
@@ -25,3 +30,7 @@ export const webSocketHandlers = {
     });
   },
 };
+
+async function asyncLog(message) {
+  console.log(message);
+}
--- a/apps/router-e2e/modules/devtools-e2e/server.mjs
+++ b/apps/router-e2e/modules/devtools-e2e/server.mjs
@@ -1,7 +1,7 @@
 
 // Top-level async to ensure esm modules are loaded correctly
-// require of esm in modern Node doesn't throw error (it "just" works)
-await asyncLog('E2E DevTools Server started.');
+// require of esm in modern Node works(*)
+await undefined;
 
 export default async function handler(request) {
   const url = new URL(request.url);
@@ -30,7 +30,3 @@ export const webSocketHandlers = {
     });
   },
 };
-
-async function asyncLog(message) {
-  console.log(message);
-}
--- a/packages/@expo/cli/add-module.js
+++ b/packages/@expo/cli/add-module.js
@@ -1,8 +1,0 @@
-const { loadModule } = require('@expo/require-utils');
-
-/**
- * @deprecated Will be removed in a future release.
- */
-module.exports = function importESM(name) {
-  return loadModule(name);
-};
--- a/packages/@expo/cli/add-module.js
+++ b/packages/@expo/cli/add-module.js
@@ -1,8 +1,8 @@
-const path = require('node:path');
-const { pathToFileURL } = require('node:url');
+const { loadModule } = require('@expo/require-utils');
 
-// A wrapper that allows to import an ESM module from a CJS module.
-// This works because the `import` in this wrapper is not transpiled by SWC.
-module.exports = function (name) {
-  return import(path.isAbsolute(name) ? pathToFileURL(name).href : name);
+/**
+ * @deprecated Will be removed in a future release.
+ */
+module.exports = function importESM(name) {
+  return loadModule(name);
 };
--- a/packages/@expo/cli/src/start/server/DevToolsPluginServerHelpers.ts
+++ b/packages/@expo/cli/src/start/server/DevToolsPluginServerHelpers.ts
@@ -2,6 +2,8 @@ import { loadModule } from '@expo/require-utils';
 import type { IncomingMessage } from 'node:http';
 import { type WebSocket, WebSocketServer } from 'ws';
 
+const debug = require('debug')('expo:start:server:devtools') as typeof console.log;
+
 /**
  * Handler default-exported by a plugin's `serverEntryPoint`. Receives a fetch API `Request`
  * with the plugin endpoint prefix stripped from the URL. Returning `null`/`undefined` falls
@@ -30,6 +32,7 @@ export async function loadRequestHandlerAsync({
   packageName: string;
   serverEntryPoint: string;
 }): Promise<DevToolsPluginRequestHandler> {
+  debug('Loading DevTools plugin server module: %s', serverEntryPoint);
   const serverModule = (await loadModule(serverEntryPoint)) as
     | DevToolsPluginRequestHandler
     | { default?: DevToolsPluginRequestHandler };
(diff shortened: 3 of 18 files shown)


packages/@expo/require-utils/src/load.ts
```diff
@@ -336,15 +340,18 @@ function evalModule(
...
+    debug('require %s', filename);
     return require(filename);
   } catch {
-    return await import(
-      path.isAbsolute(filename) ? url.pathToFileURL(filename).toString() : filename
-    );
+    const importFilename = path.isAbsolute(filename)
+      ? url.pathToFileURL(filename).toString()
+      : filename;
+    debug('import %s', importFilename);
```
kitten: there's a lot of debug statements here that shouldn't be needed?

Edit: meant, more as in, I'm not sure these add continued value, if we're logging that a file is being loaded. Arguably, if everything already used these methods, it's understandable, but I'm not sure if adding this right now makes things easier, especially with a custom debug.ts harness

packages/@expo/require-utils/src/load.ts
```diff
@@ -336,15 +340,18 @@ function evalModule(
...
+    return await importAsync(importFilename);
   }
 }
 
-async function loadModule(filename: string) {
+async function loadModule<T = any>(filename: string): Promise<T> {
```
kitten: there was a bit of a conscious choice not to add a generic here (over a cast). I think defaulting to `any` over `unknown` isn't a great choice, but the latter wasn't done since there's dependents that switched over to this API but didn't yet have casts typings, so it was meant to encourage casting over passing a generic.

Both is fine, but just noting since it's an unrelated change

packages/@expo/cli/add-module.js
```diff
@@ -1,8 +1,8 @@
...
 
-// A wrapper that allows to import an ESM module from a CJS module.
-// This works because the `import` in this wrapper is not transpiled by SWC.
-module.exports = function (name) {
-  return import(path.isAbsolute(name) ? pathToFileURL(name).href : name);
+/**
+ * @deprecated Will be removed in a future release.
+ */
+module.exports = function importESM(name) {
+  return loadModule(name);
```
kitten: Removal can be squashed in, since it's an internal dep

apps/router-e2e/modules/devtools-e2e/server.mjs
```diff
@@ -1,3 +1,8 @@
+
+// Top-level async to ensure esm modules are loaded correctly
+// require of esm in modern Node doesn't throw error (it "just" works)
+await asyncLog('E2E DevTools Server started.');
```
kitten: redundant log?

packages/@expo/cli/src/start/server/DevToolsPluginServerHelpers.ts
```diff
@@ -117,14 +45,12 @@ export async function loadRequestHandlerAsync({
...
   packageName,
-  packageRoot,
   serverEntryPoint,
 }: {
   packageName: string;
-  packageRoot: string;
   serverEntryPoint: string;
 }): Promise<Record<string, WebSocketServer>> {
-  const serverModule = await loadServerModuleAsync(serverEntryPoint, packageRoot);
+  const serverModule = await loadModule(serverEntryPoint);
```
kitten: nit/haven't checked this. Do we need to do a resolution first? I'm not sure what the constraints are for `serverEntryPoint`, but I'm assuming we may want to do a `resolveFrom` (same package, so not too inconvenient) on `packageRoot` first?

kitten:
unrelated, but leaving for future readers
> `loadModule` async import was transformed and thus now working for top level async esm modules

Was this meant to say something like "direct async `import` was not possible, due to transform and Node previously not supporting ESM via CJS in Node 18"? Afaik, the main reason this probably wasn't used was that depending on who's importing and ESM via CJS support was tricky and had some caveats in Node 18 (and others that are smoothed over by `@expo/require-utils` especially for the eval case), but afaik this shouldn't relate to async import directly

packages/@expo/require-utils/src/load.ts
krystofwoldrich: The logs are in from of every call which tries to load a module, to make it clear where the load failed, what it tried and what it loaded.

packages/@expo/require-utils/src/load.ts
krystofwoldrich: Got it, I'll revert it, it's unfortunate because the any won't force anyone to cast.

packages/@expo/cli/src/start/server/DevToolsPluginServerHelpers.ts
krystofwoldrich: It's restricted to be a path inside of the package.  https://github.com/expo/expo/blob/ab2e222da508bc1f60228cf692c0ca30d6eba3f8/packages/expo-modules-autolinking/src/platforms/devtools.ts#L25

apps/router-e2e/modules/devtools-e2e/server.mjs
krystofwoldrich: It's just a E2E test module, but yeah, can be removed.

packages/@expo/require-utils/src/load.ts
kitten: yes, I'm mostly concerned though that while it's useful to temporarily add, it's not useful to keep. However, if kept, it basically doesn't use `debug` (or the event logs, if/when we add those), so it's a one-off pattern for debugging that's basically immediately obsoleted

as in, assuming things go wrong, with a repro, it could be pinned down easily (but the known cases of what could go wrong are bound to the ESM/CJS switch, which should be mostly the only switch over that's interesting here?)
But assuming things go right, it's unused debugging code that's unique here and never done in this same way
(and it also uses, `console.log`, not debug)

I basically would find it hard to keep track of this, so I'd prefer a hoisted debug in the CLI calls. Since those may need an added `resolveFrom` that'd also give us the case of what loads, narrowed down

so at that point the error annotations we have for the catch case should kick in, and the remaining errors are most likely to be down to user code, right?

Edit: tl;dr is a one-off debug that few people know about expected here?

packages/@expo/cli/src/start/server/DevToolsPluginServerHelpers.ts
kitten: the main thing I think that this misses (which I missed before) is that if this path is not a file path (realpath fails), we default to `resolvedPath`. This then means that we perform further Node resolution on it, no matter if that's pointing at a file, so it could do `/index[ext] | [ext]` resolution on it implicitly, which I think is a bit unexpected.

If we were to slot in another check (`resolveFrom` or otherwise) we narrow down the supported load cases potentially 🤔

packages/@expo/require-utils/src/load.ts
krystofwoldrich: It's larger switch -> esm/cjs/eval/eval-ts. 

Not using debug was intentional, since the package is nice and compact, I didn't want to add extra deps for 4 logs. 

It's enabled like other debug in the code base.

Yes, with repro it's easy to debug, but these logs would let us in case without repro. For example like now with the transformed `import` which is not user code.

But saying that it's ok for me to add the debug logs to the callers.

packages/@expo/cli/src/start/server/DevToolsPluginServerHelpers.ts
krystofwoldrich: Updated in [`f957cb2` (this PR)](https://github.com/expo/expo/pull/47139/changes/f957cb2e1c4b652e69faf7d750e1bfebb84ed9e1)

Kudo (approved):
looks good for the mcp changes. will delegate require-utils review to phil

packages/@expo/require-utils/static/import-util.js
```diff
@@ -0,0 +1,3 @@
+exports.importAsync = function importAsync(moduleName) {
+  return import(moduleName);
```
kitten: This seems like a regression in the SWC transpilation which was not caught because the `requireOrImport` code path isn't yet taken by other modules, I think, addressed by: #47170

kitten (approved):

PR #47140, opened 2026-06-22
[expo/cli] Remove add-module entry point

krystofwoldrich:
Stacked on #47139.

Separate because removing `@expo/cli/add-module` is breaking if external consumers import that subpath.


kitten (commented):
this isn't breaking as `@expo/cli` isn't a public API 😅 also, moving this for the breaking change instead of folding this in won't be needed. I think it'll be fine to drop in the same PR, plus it's not a breaking change, as by our own constraints `@expo/cli` can't/shouldn't be depended on directly

PR #47144, opened 2026-06-22
[tracking-transparency] add getAdvertisingIdAsync (prevent long JS thread blocking)

greg-schrammel:
# Why

`getAdvertisingId` is synchronous, so reading the advertising ID blocks the JS thread until the native lookup returns. On Android that lookup is an IPC call to Google Play Services — Google's docs warn it ["cannot be called in the main thread as it may block leading to ANRs"](https://developers.google.com/android/reference/com/google/android/gms/ads/identifier/AdvertisingIdClient#public-static-advertisingidclient.info-getadvertisingidinfo-context-context), and in my profile it stalled the JS thread for ~280ms. This adds an async variant that moves the lookup off the JS thread.

# How

Added `getAdvertisingIdAsync` alongside the existing sync getAdvertisingId. Both call the same native lookup, and the JS wrapper applies the same all-zeros-UUID → null mapping, so the only difference is threading.

# Test Plan

Profiled with `react-native-release-profile`

# Checklist


packages/expo-tracking-transparency/src/TrackingTransparency.ts
```diff
@@ -58,6 +58,31 @@ export function getAdvertisingId(): string | null {
   return advertisingId;
 }
 
+/**
+ * Asynchronously gets the advertising ID. This is the recommended way to read the advertising ID:
+ * see [`getAdvertisingId()`](#getadvertisingid) for details on the returned value.
+ *
+ * On Android, reading the advertising ID requires a IPC call to Google Play Services.
+ * The synchronous [`getAdvertisingId()`](#getadvertisingid) can block the JavaScript thread for hundreds of milliseconds.
+ * This function offloads the lookup from the JavaScript thread
```
amandeepmittal:
```suggestion
 * This function offloads the lookup from the JavaScript thread.
```

Missing period

amandeepmittal (changes requested):

amandeepmittal (approved):
Thanks for adding docs!

PR #47145, opened 2026-06-22
[expo-image] add expo-observe integration

Ubax:
# Why

Using the image loaded event from previous PR, this PR logs event when image is too big

# How

1. Subscribe to image loaded
2. When `img.w * img.h > screen.w * screen.h * pixelRatio * userDefinedRation` log `expo-image.oversized` event (user defined ratio defaults to `2`)
3. Extend observe configuration in ts to include `expo-image` property

# Test Plan

# Checklist

### DIFF
--- a/apps/observe-tester/app/_layout.tsx
+++ b/apps/observe-tester/app/_layout.tsx
@@ -9,6 +9,9 @@ Observe.configure({
   dispatchInDebug: true,
   integrations: {
     'expo-router': true,
+    'expo-image': {
+      oversizeThreshold: 1.5,
+    },
   },
 });
 
--- a/packages/expo-image/src/observe.ts
+++ b/packages/expo-image/src/observe.ts
@@ -1,0 +1,172 @@
+'use client';
+
+import { requireOptionalNativeModule } from 'expo';
+import type { ExpoAppMetricsModuleType } from 'expo-app-metrics';
+import type { ObserveIntegrationsConfig, ObserveModule } from 'expo-observe';
+import { Dimensions, PixelRatio } from 'react-native';
+
+import type { ImageNativeModule } from './Image.types';
+
+/**
+ * Configuration for the `expo-observe` integration, set through
+ * `Observe.configure({ integrations: { 'expo-image': ... } })`. Passing `true` enables it with
+ * defaults; the object form tunes the behavior.
+ *
+ * The `declare module 'expo-observe'` augmentation that registers the `'expo-image'` key lives in
+ * `Image.types.ts` (always in the package's public type graph via `export *`, so it is picked up
+ * whenever expo-image is imported). `Image.types.ts` `import type`s this from here. It is exported
+ * from this module but not from the package entry, so it is not part of the public API.
+ */
+export type ExpoImageIntegrationConfig = {
+  /**
+   * An image is reported as oversized when its decoded pixel area exceeds the screen's physical
+   * pixel count (its point area times the square of the device pixel ratio) by more than this
+   * factor. For example, `1.5` flags an image decoded at more than 1.5× the pixels the screen
+   * physically has — a full-screen image plus 50% headroom.
+   *
+   * @default 1.5
+   */
+  oversizeThreshold?: number;
+};
+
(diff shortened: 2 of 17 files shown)


Ubax:
* **#47145** <a href="https://app.graphite.com/github/pr/expo/expo/47145?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/47145?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#47388** <a href="https://app.graphite.com/github/pr/expo/expo/47388?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

apps/observe-tester/app/_layout.tsx
```diff
@@ -9,6 +9,9 @@ Observe.configure({
   dispatchInDebug: true,
   integrations: {
     'expo-router': true,
+    'expo-image': {
+      ratio: 1.5,
```
tsapeta: `ratio` is not very descriptive, what about `oversizeThreshold`?
```suggestion
      oversizeThreshold: 1.5,
```

packages/expo-image/src/observe.ts
```diff
@@ -0,0 +1,168 @@
...
+  }
+  // Screen area is in points; the decoded image is in pixels, so convert with pixelRatio² to get
+  // the screen's physical pixel count.
+  const budget = screenWidth * screenHeight * pixelRatio * pixelRatio;
+  if (!(budget > 0) || width * height <= budget * state.threshold) {
+    return;
+  }
+  state.reported.add(url);
+  try {
+    state.appMetrics.logEvent('expo-image.oversized', {
```
tsapeta, line 84: Just to confirm – we don't want `expo.image.oversized`, right? It would actually be a bit harder since it goes through validation 😅

packages/expo-image/src/observe.ts
```diff
@@ -0,0 +1,168 @@
+import type { ExpoAppMetricsModuleType } from 'expo-app-metrics';
+import { requireOptionalNativeModule } from 'expo-modules-core';
```
tsapeta: It's better to import public Expo Modules APIs from `expo`. Modules don't depend on `expo-modules-core` directly, they do have `expo` in peer dependencies.
```suggestion
import { requireOptionalNativeModule } from 'expo';
```

packages/expo-image/src/observe.ts
```diff
@@ -0,0 +1,168 @@
+import type { ExpoAppMetricsModuleType } from 'expo-app-metrics';
```
tsapeta, line 4: Are you sure this works in projects without `expo-app-metrics` installed? It's a dev dependency.

packages/expo-image/src/observe.ts
```diff
@@ -0,0 +1,168 @@
...
+   * pixel count (its point area times the square of the device pixel ratio) by more than this
+   * factor. For example, `1.5` flags an image decoded at more than 1.5× the pixels the screen
+   * physically has — a full-screen image plus 50% headroom.
+   *
+   * @default 1.5
+   */
+  ratio?: number;
+};
+
+const DEFAULT_RATIO = 1.5;
```
tsapeta: If you decide to change the config name to `oversizeThreshold`, then change this one too.

```suggestion
const DEFAULT_OVERSIZE_THRESHOLD = 1.5;
```

packages/expo-image/src/observe.ts
Ubax: That's what I thought. I want to use it as an example for external libraries, so this prefix is I think a bit more suitable.

packages/expo-image/src/observe.ts
Ubax: It's `type` only import so it should. I checked in router-e2e app which has no dependency on observe or app-metrics. Once this is merged I'll test it again.

We use the same pattern in observe for router or react-navigation integration and no complaints so far

tsapeta (approved):

Issue #46242, opened 2026-05-25
[expo-modules-jsi] Build fails on Xcode 26 / Swift 6.2 (weak let + Sendable mutable property)


Deveshb15:
## Summary

`expo-modules-jsi@56.0.7` (pinned by Expo SDK 56) fails to compile with Xcode 26.0.1 / Swift 6.2 because of two stacked strictness rules that earlier Swift compilers did not enforce.

The bug reproduces from a zero-modification `npx create-expo-app --template default` scaffold — no project code is involved.

## Minimal reproduction

**Repo:** https://github.com/Deveshb15/earshot-jsi-repro

```bash
git clone https://github.com/Deveshb15/earshot-jsi-repro
cd earshot-jsi-repro
npm install
npx expo prebuild --clean
npx expo run:ios --device   # any iOS build with Xcode 26
```

The repo is literally `npx create-expo-app@latest --template default` + a `README.md`. The bug lives entirely inside `node_modules/expo-modules-jsi/`.

## Environment

| Tool | Version |
|---|---|
| Xcode | 26.0.1 (Build 17A400) |
| Swift compiler | 6.2 (swiftlang-6.2.0.19.9 clang-1700.3.19.1) |
| Expo SDK | 56.0.4 |
| `expo-modules-jsi` | 56.0.7 |
| React Native | 0.85.3 |
| iOS deployment target | 18.0 |

## Round 1 — `weak let` rejected (15 errors)

> `'weak' must be a mutable variable, because it may change at runtime`

15 sites across 14 files in `apple/Sources/ExpoModulesJSI/`:

- `Contexts/HostFunctionContext.swift:5`
- `Contexts/HostObjectContext.swift:10`
- `Runtime/JavaScriptActor.swift:96`
- `Runtime/JavaScriptActor.swift:117`
- `Runtime/JavaScriptPropNameID.swift:7`
- `Runtime/Values/JavaScriptArray.swift:10`
- `Runtime/Values/JavaScriptArrayBuffer.swift:11`
- `Runtime/Values/JavaScriptBigInt.swift:42`
- `Runtime/Values/JavaScriptError.swift:5`
- `Runtime/Values/JavaScriptFunction.swift:7`
- `Runtime/Values/JavaScriptObject.swift:11`
- `Runtime/Values/JavaScriptPromise.swift:15`
- `Runtime/Values/JavaScriptTypedArray.swift:7`
- `Runtime/Values/JavaScriptValue.swift:12`
[...]

PR #47157, opened 2026-06-22
[expo-modules-jsi] Fix Xcode 26 / Swift 6.2 build (weak let → nonisolated(unsafe) weak var)

higor-leodoro:
# Why

`expo-modules-jsi` fails to compile with **Xcode 26 / Swift 6.2** (tracked in #46242). Two stricter Swift 6.2 rules, stacked:

1. `weak let` is no longer allowed — a `weak` reference may become `nil` at runtime, so it must be a `var`.
2. After switching to `weak var`, the **Sendable** types (`HostFunctionContext`, `HostObjectContext`, `JavaScriptPropNameID`, `JavaScriptValue`) fail: a mutable stored property isn't allowed in a `Sendable`-conforming type.

This blocks `expo run:ios` / building from source on the current toolchain for SDK 56 (and `main`).

# How

Prefix every `weak let runtime` with `nonisolated(unsafe)` and make it `weak var` — i.e. `nonisolated(unsafe) weak var runtime: JavaScriptRuntime?`. `nonisolated(unsafe)` (Swift 5.10+) is the sanctioned escape hatch for storage mutated outside any concurrency domain. It resolves **both** rules uniformly for the class and struct cases, with **no public API change** and no behavior change (the references were already effectively single-assignment).

16 sites across 14 files in `packages/expo-modules-jsi/apple/Sources/ExpoModulesJSI/`.

# Test Plan

- Building `expo-modules-jsi` with Xcode 26.0.1 / Swift 6.2 succeeds (previously 15–16 compile errors: `'weak' must be a mutable variable` and `stored property 'runtime' of 'Sendable'-conforming class … is mutable`).
- Verified end-to-end: a dev build (`expo run:ios`) on an SDK 56 app compiles, installs, and runs on the iOS Simulator with the equivalent patch applied.

Refs #46242.

### DIFF
--- a/packages/expo-modules-jsi/apple/Sources/ExpoModulesJSI/Contexts/HostFunctionContext.swift
+++ b/packages/expo-modules-jsi/apple/Sources/ExpoModulesJSI/Contexts/HostFunctionContext.swift
@@ -1,6 +1,6 @@
 /// Context that captures Swift values to pass them to JSI host function as an unmanaged pointer for interoperability with C++.
 internal final class HostFunctionContext: Sendable {
-  weak let runtime: JavaScriptRuntime?
+  nonisolated(unsafe) weak var runtime: JavaScriptRuntime?
   let name: String?
   let call: JavaScriptRuntime.SyncFunctionClosure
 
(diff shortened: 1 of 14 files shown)


tsapeta:
In the issue that you linked to, I mentioned many times that this is not the proper solution. SDK 56 requires Xcode 26.4 (Swift 6.3), it was mentioned in the blog post, it is mentioned in the docs.
`nonisolated weak var` is not fundamentally the same as `weak let`, `weak let` is there on purpose. Besides that, many of other SDK modules are now shipped as prebuilt xcframeworks which were compiled with Xcode 26.4. Using an older version to link them may lead to crashes at runtime. Thus it's highly not recommended to use any version before 26.4.

PR #47159, opened 2026-06-23
[observe][ios] Correct OTLP retry behavior for dispatches

douglowder:
# Why

Our client side code in expo-observe does not comply with the Open Telemetry spec for retrying requests that fail. This PR fixes that for iOS (Android changes will be in a separate PR).

# How

Classify responses into successful, retryable, and failed types, and handle them as appropriate. Non-retryable errors are handled by moving the cursor, to prevent the request being retried, which means that those metrics or logs will be dropped. In this PR, dropped requests are logged as warnings.

# Test Plan

- Testing with a real Expo app against a local server to send back different responses
- New iOS unit tests added.

# Checklist

### DIFF
--- a/packages/expo-observe/ios/DispatchUtils.swift
+++ b/packages/expo-observe/ios/DispatchUtils.swift
@@ -1,0 +1,170 @@
+// Copyright 2025-present 650 Industries. All rights reserved.
+
+import ExpoModulesCore
+
+/// Outcome of a single dispatch attempt to the OTLP endpoint. Modeled after the OTLP retry
+/// guidance: a 3-way classification distinguishes transient failures (retry the same batch)
+/// from permanent ones (drop the batch and advance the cursor so it can't wedge the loop).
+///
+/// `retryable.retryAfter` carries the parsed `Retry-After` header value in seconds (or `nil`
+/// when the server didn't supply one). `nonRetryable.reason` carries a short diagnostic string
+/// for the `warn`-level log line.
+internal enum DispatchResult: Equatable, Sendable {
+  case success
+  case retryable(retryAfter: TimeInterval?)
+  case nonRetryable(reason: String)
+}
+
+extension Data {
+  /// First ~512 bytes of the response body as UTF-8, for inclusion in error log lines. Bounded
+  /// so a giant HTML error page doesn't flood the log; truncated mid-codepoint is acceptable
+  /// for a diagnostic excerpt.
+  fileprivate func bodyExcerpt(limit: Int = 512) -> String {
+    let slice = self.prefix(limit)
+    return String(data: slice, encoding: .utf8) ?? "<unreadable>"
+  }
+}
+
+/// HTTP transport + response classification for `ObservabilityManager`. Lives in its own file
+/// to keep `Observability.swift` focused on the dispatch lifecycle (cursors, batching, decide
+/// whether to dispatch at all). All functions are pure / `nonisolated` so they can be
+/// unit-tested without instantiating the manager or hitting the network.
+internal enum DispatchUtils {
+  /// POSTs the encoded `body` to `endpointUrl` and classifies the response. Catches both local
+  /// encoding errors (non-retryable — same bytes will fail again) and transport errors (DNS,
+  /// TLS, timeout, connection reset — retryable per OTLP).
+  internal static func sendRequest(
+    to endpointUrl: URL,
+    body: any Encodable
+  ) async -> DispatchResult {
+    var request = URLRequest(url: endpointUrl)
+    request.httpMethod = "POST"
+    request.allHTTPHeaderFields = [
+      "Content-Type": "application/json",
+      // Tells `NetworkRequestURLProtocol` to skip observation so our own telemetry uploads don't
+      // get logged back into the network-request stream. The header reaches o.expo.dev unchanged
+      // (we control that endpoint, so the harmless overhead is fine). The name is duplicated here
+      // rather than imported: expo-observe must not depend on expo-app-metrics internals. Keep it
+      // in sync with `NetworkRequestURLProtocol.internalHeaderName` in expo-app-metrics.
+      "Expo-AppMetrics-Skip": "1",
+    ]
+    do {
+      request.httpBody = try body.toJSONData([])
+    } catch {
+      return .nonRetryable(reason: "encoding error: \(error.localizedDescription)")
+    }
+
+    #if DEBUG
+    observeLogger.debug("[EAS Observe] Sending the request to \(endpointUrl) with body:")
+    // Use `print` so the JSON can be copied without including the log level emojis. Wrapped in
+    // `#if DEBUG` so release builds don't pay for a second pretty-printed encode of the payload.
+    if let prettyBody = try? body.toJSONString(.prettyPrinted) {
+      print(prettyBody)
+    }
+    #endif
+
+    let responseData: Data
+    let urlResponse: URLResponse
+    do {
+      (responseData, urlResponse) = try await URLSession.shared.data(for: request)
+    } catch {
+      observeLogger.warn(
+        "[EAS Observe] Transport error talking to \(endpointUrl): \(error.localizedDescription)"
+      )
+      return .retryable(retryAfter: nil)
+    }
+
+    guard let urlResponse = urlResponse as? HTTPURLResponse else {
+      // Non-HTTP response — exotic but not impossible (proxies). Treat as transient.
+      return .retryable(retryAfter: nil)
+    }
+
+    let retryAfterHeader = urlResponse.value(forHTTPHeaderField: "Retry-After")
+    let partialSuccess = try? JSONDecoder().decode(OTServiceResponse.self, from: responseData)
+      .partialSuccess
+    let result = classifyResponse(
+      statusCode: urlResponse.statusCode,
+      retryAfterHeader: retryAfterHeader,
+      partialSuccess: partialSuccess,
+      bodyExcerpt: { responseData.bodyExcerpt() }
+    )
+
+    switch result {
+    case .success:
+      observeLogger.debug(
+        "[EAS Observe] Server responded successfully with \(urlResponse.statusCode) and data: "
+          + "\(String(data: responseData, encoding: .utf8) ?? "<unreadable>")"
+      )
+    case .retryable:
+      observeLogger.warn(
+        "[EAS Observe] Server responded with \(urlResponse.statusCode) (retryable) and data: "
+          + "\(String(data: responseData, encoding: .utf8) ?? "<unreadable>")"
+      )
+    case .nonRetryable(let reason):
+      observeLogger.warn(
+        "[EAS Observe] Server responded with \(urlResponse.statusCode) (non-retryable, "
+          + "\(reason)) and data: \(String(data: responseData, encoding: .utf8) ?? "<unreadable>")"
+      )
+    }
+    return result
+  }
+
+  /// Pure classifier that maps an HTTP response into one of three retry outcomes. Extracted
+  /// from `sendRequest` so the OTLP-spec rules can be unit-tested without a real network call.
+  ///
+  /// `bodyExcerpt` is invoked lazily, only when the result is `.nonRetryable` and the reason
+  /// string benefits from including a peek at the response body. Lets the caller bound how much
+  /// data we slurp into the log line.
+  internal static func classifyResponse(
+    statusCode: Int,
+    retryAfterHeader: String?,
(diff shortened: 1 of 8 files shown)


douglowder:
* **#47160** <a href="https://app.graphite.com/github/pr/expo/expo/47160?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#47159** <a href="https://app.graphite.com/github/pr/expo/expo/47159?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/47159?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/expo-observe/ios/Observability.swift
```diff
@@ -9,7 +9,23 @@ internal struct ObservabilityManager {
   private static var logsEndpointUrl: URL? = nil
   private static var projectId: String? = nil
 
+  /// In-memory retry-gate state. Reset implicitly when the process restarts — a relaunch
+  /// usually means enough time passed that the transient cause has cleared anyway, and
+  /// persisting the gate would mean an extra UserDefaults write on every retryable response.
+  private static var retryGateState: DispatchUtils.RetryGateState = .initial
```
Ubax: [🏗️ **design**] A single global gate is shared across two independent OTLP endpoints (`/v1/metrics` and `/v1/logs`). Because `dispatch()` runs metrics then logs sequentially and each calls `applyRetryOutcome` on this same field, the two signals' health gets conflated. Three concrete consequences:

1. **Backoff never escalates.** If metrics persistently returns retryable (503) but logs persistently succeeds: metrics sets `counter=1`, then logs' `.success` resets `counter=0` (see `nextRetryGateState`). Every subsequent round repeats from 0, so the delay is pinned at `backoff(1)` and the exponential schedule never engages — defeating the PR's stated goal.
2. **Last-writer-wins on the deadline.** A metrics 503 computing a 60s gate can be overwritten by a logs `429 Retry-After: 5`, silently discarding the longer backoff.
3. **A healthy signal is suppressed by an unhealthy one**, since the gate check at the top of `dispatch()` short-circuits both.

Consider per-signal gate state (keyed by signal, or two fields) so metrics and logs back off independently.

packages/expo-observe/ios/Observability.swift
```diff
@@ -9,7 +9,23 @@ internal struct ObservabilityManager {
...
+
   internal static func dispatch() async {
+    // Honor the retry gate first: if a previous round told us to back off (server-supplied
+    // `Retry-After` or computed exponential backoff after a transient failure), short-circuit
+    // until the deadline elapses. Skips both signals together so we don't hammer the same
+    // server endpoint with logs traffic while it's asking us to slow down metrics.
+    if let until = retryGateState.dispatchAfterDate, until > Date() {
```
Ubax: [💡 suggestion] The comment says both signals are "skipped together," but the gate is only checked once here, at round entry. If `dispatchMetrics` sets the gate mid-round (retryable response), `dispatchLogs` still fires in this same round — it isn't re-checked against the gate metrics just set. So the "skip together" behavior only holds on the *next* round. Worth tightening the comment, or re-checking the gate before `dispatchLogs` if same-round suppression is actually intended.

packages/expo-observe/ios/Observability.swift
```diff
@@ -49,15 +78,24 @@ internal struct ObservabilityManager {
...
+      highestId: highestId
+    )
+    switch result {
+    case .success:
+      ObserveUserDefaults.lastDispatchDate = Date.now
+    case .retryable:
+      break
+    case .nonRetryable(let reason):
+      observeLogger.warn(
+        "[EAS Observe] Dropping batch of \(events.count) metric event(s) past id "
```
Ubax, line 120: [💡 suggestion] When `classifyResponse` returns `.nonRetryable` for a 2xx `partial_success` (some records rejected, most accepted), this logs "Dropping batch of N metric event(s)" using the full `events.count`. That overstates the loss — only the rejected subset was dropped; the rest landed and the cursor correctly advances past all of them. The `rejectedCount` from the partial-success path is the accurate number to surface here. Same applies to the logs branch below.

packages/expo-observe/ios/Tests/ObservabilityClassifyResponseTests.swift
```diff
@@ -0,0 +1,292 @@
...
+    let result = DispatchUtils.classifyResponse(
+      statusCode: 503,
+      retryAfterHeader: "5",
+      partialSuccess: nil
+    )
+    #expect(result == .retryable(retryAfter: 5))
+  }
+
+  @Test
+  func `408 412 502 504 are retryable`() {
```
Ubax: [🐧 nit] Test name claims `412` is retryable, but the loop only iterates `[408, 502, 504]` — 412 is never exercised, and 412 (Precondition Failed) is in fact NOT in OTLP's retryable set (it would correctly fall through to `.nonRetryable`). Misleading as documentation. Rename to match the codes actually tested (e.g. `408 502 504 are retryable`), or add the missing codes the name promises (429 is also missing).

packages/expo-observe/ios/Observability.swift
douglowder: Will investigate and come up with a better design, see my comments in the Android PR.

douglowder: We now have separate gates and separate failure counts for the two endpoints.

packages/expo-observe/ios/Observability.swift
douglowder: Gates are checked separately now.

packages/expo-observe/ios/Tests/ObservabilityClassifyResponseTests.swift
douglowder: Fixed.

packages/expo-observe/ios/DispatchUtils.swift
```diff
@@ -0,0 +1,307 @@
...
+      if let partial = partialSuccess, partial.rejectedCount > 0 {
+        return .partialSuccess(partial)
+      }
+      return .success
+    }
+
+    // Retryable per OTLP. `408` lacks a real-world Retry-After in practice but the spec admits
+    // it; carry the header through unchanged so server can override us.
+    switch statusCode {
+    case 408, 429, 502, 503, 504:
```
tsapeta: [💡 suggestion] Spec divergence on `408`. The OTLP/HTTP spec's retryable set is exactly `{429, 502, 503, 504}`, and it states "All other `4xx` or `5xx` response status codes MUST NOT be retried." `408` isn't mentioned anywhere in the spec, so it falls under that catch-all MUST NOT. Including it here makes a `408` retryable, which is a literal violation of the rule this PR is meant to bring us into compliance with.

The adjacent comment ("`408` lacks a real-world Retry-After in practice but the spec admits it") is the opposite of what the spec says: it doesn't admit `408`, it implicitly forbids retrying it. Retrying a `408` (Request Timeout) is arguably reasonable engineering, but it's a deliberate departure from the spec rather than compliance, so it should be a conscious choice with an accurate comment. Either drop `408` from the retryable set to match the spec exactly, or keep it and rewrite the comment to say "we retry `408` even though OTLP doesn't list it, because <reason>."

Ref: https://opentelemetry.io/docs/specs/otlp/#otlphttp-throttling and the failure-handling section just above it.

packages/expo-observe/ios/Observability.swift
```diff
@@ -49,19 +95,42 @@ internal struct ObservabilityManager {
...
+    ObserveUserDefaults.lastDispatchedMetricId = DispatchUtils.nextCursor(
+      for: result,
+      currentCursor: cursor,
+      highestId: highestId
+    )
+    switch result {
+    case .success:
+      ObserveUserDefaults.lastDispatchDate = Date.now
+    case .partialSuccess(let partial):
+      // The batch reached the server; bump `lastDispatchDate` so stale-cursor repair
```
tsapeta: [💡 suggestion] This comment claims bumping `lastDispatchDate` keeps "stale-cursor repair" from tripping, but `repairCursorIfStale` (CursorRepair.swift) keys purely on `cursor > maxId` and never reads `lastDispatchDate`. In fact `lastDispatchDate` is write-only across the whole package: nothing in expo-observe (or app-metrics, or JS) ever reads it back. So this write has no functional effect, and the justification describes a mechanism that doesn't exist.

Not a correctness bug (the write is harmless), but the comment will mislead a future maintainer into thinking the two are coupled. Either drop the `lastDispatchDate` write plus comment here, or correct the comment to say what the write actually does (currently: nothing observable). It may be worth deciding whether `lastDispatchDate` should be removed package-wide in a follow-up, since the pre-existing `.success` write is equally dead.

packages/expo-observe/ios/Observability.swift
```diff
@@ -98,14 +167,28 @@ internal struct ObservabilityManager {
...
+    applyRetryOutcome(result, to: &logsRetryGate)
+    ObserveUserDefaults.lastDispatchedLogId = DispatchUtils.nextCursor(
+      for: result,
+      currentCursor: cursor,
+      highestId: highestId
+    )
+    switch result {
+    case .success, .retryableFailure:
+      break
+    case .partialSuccess(let partial):
```
tsapeta, line 178: [🐧 nit] Asymmetry worth a sanity check: the metrics `.partialSuccess` arm sets `lastDispatchDate`, but the logs `.partialSuccess` arm (and the logs `.success` arm) doesn't. Given `lastDispatchDate` is currently dead state (see the other comment), this has no runtime impact today, but if `lastDispatchDate` ever gains a real reader, metrics and logs would behave differently for no documented reason. Pick one convention for both signals.

packages/expo-observe/ios/DispatchUtils.swift
```diff
@@ -0,0 +1,307 @@
...
+      return nil
+    }
+    if let seconds = TimeInterval(raw) {
+      guard seconds.isFinite else { return nil }
+      return clampToBounds(seconds, base: base, cap: cap)
+    }
+    let formatter = DateFormatter()
+    formatter.locale = Locale(identifier: "en_US_POSIX")
+    formatter.timeZone = TimeZone(identifier: "GMT")
+    formatter.dateFormat = "EEE, dd MMM yyyy HH:mm:ss zzz"
```
tsapeta, line 194: [🐧 nit] `parseRetryAfter` only handles the IMF-fixdate format (`EEE, dd MMM yyyy HH:mm:ss zzz`). RFC 7231 permits two other legacy date forms (RFC 850 and asctime). In practice IMF-fixdate is the only one a conforming server MUST send, and an unparsed date falls through to `nil` then to `computeBackoffDelay`, which is a safe fallback, so this is acceptable. Just flagging it so it's a conscious choice rather than an oversight; a one-line comment noting "IMF-fixdate only; others fall through to backoff" would document the intent.

packages/expo-observe/ios/DispatchUtils.swift
```diff
@@ -0,0 +1,307 @@
...
+    cap: TimeInterval = backoffCapSeconds
+  ) -> TimeInterval? {
+    guard let raw = header?.trimmingCharacters(in: .whitespacesAndNewlines), !raw.isEmpty else {
+      return nil
+    }
+    if let seconds = TimeInterval(raw) {
+      guard seconds.isFinite else { return nil }
+      return clampToBounds(seconds, base: base, cap: cap)
+    }
+    let formatter = DateFormatter()
```
tsapeta, line 190: [🐧 nit] Minor: a fresh `DateFormatter` is allocated on every call that hits the date branch, and `DateFormatter` construction is comparatively expensive. This path is low-frequency (only on a retryable response carrying a non-numeric `Retry-After`), so it's not a real perf concern, but since the formatter is fully static a `static let` cached formatter would be cleaner and free. Take it or leave it.

tsapeta (approved):
Looks good, just a few nits 🙂

packages/expo-observe/ios/Observability.swift
douglowder: We have separate states for partial success and nonretryable now, so this issue will not occur.

Ubax (approved):

douglowder:
### Merge activity

* **Jun 26, 9:12 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/expo/expo/47159).
* **Jun 26, 9:12 AM UTC**: @douglowder merged this pull request with [Graphite](https://app.graphite.com/github/pr/expo/expo/47159).

PR #47160, opened 2026-06-23
[observe][android] Correct OTLP retry behavior for dispatches

douglowder:
# Why

Our client side code in expo-observe does not comply with the Open Telemetry spec for retrying requests that fail. This PR fixes that for Android (iOS changes are in the downstack PR).

# How

Classify responses into successful, retryable, and failed types, and handle them as appropriate. Non-retryable errors are handled by moving the cursor, to prevent the request being retried, which means that those metrics or logs will be dropped. In this PR, dropped requests are logged as warnings.

# Test Plan

- Testing with a real Expo app against a local server to send back different responses
- New Android unit tests added.

# Checklist

### DIFF
--- a/packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
+++ b/packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
@@ -20,7 +20,7 @@ import kotlin.random.Random
  *   pending-ID removal and counter reset match `Success`; the case stays distinct so the
  *   call site can log the rejected count and `errorMessage` clearly rather than describing
  *   it as a drop.
- * - `RetryableFailure` — transient failure (408/429/502/503/504 or transport error); retry
+ * - `RetryableFailure` — transient failure (429/502/503/504 or transport error); retry
  *   the same batch after `retryAfterMs` or a client-computed backoff.
  * - `NonRetryableFailure` — permanent failure (4xx/5xx outside the retryable set, encoding
  *   error); drop the batch so it can't wedge the queue.
@@ -74,7 +74,7 @@ object DispatchUtils {
     }
 
     return when (statusCode) {
-      408, 429, 502, 503, 504 -> DispatchResult.RetryableFailure(retryAfter)
+      429, 502, 503, 504 -> DispatchResult.RetryableFailure(retryAfter)
       else -> {
         val excerpt = bodyExcerpt()
         val suffix = if (excerpt.isEmpty()) "" else ": $excerpt"
--- a/packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
+++ b/packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
@@ -117,8 +117,7 @@ object DispatchUtils {
   fun parseRetryAfter(
     header: String?,
     base: Long = backoffBaseMs,
-    cap: Long = backoffCapMs,
-    now: Long = TimeUtils.getWallClockMillis()
+    cap: Long = backoffCapMs
   ): Long? {
     val raw = header?.trim() ?: return null
     if (raw.isEmpty()) return null
@@ -133,7 +132,7 @@ object DispatchUtils {
     formatter.timeZone = TimeZone.getTimeZone("GMT")
     val parsed = runCatching { formatter.parse(raw) }.getOrNull()
     if (parsed != null) {
-      return clampToBounds(parsed.time - now, base, cap)
+      return clampToBounds(TimeUtils.millisUntil(parsed), base, cap)
     }
     return null
   }
--- a/packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
+++ b/packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
@@ -132,6 +132,8 @@ object DispatchUtils {
       return clampToBounds((it * 1000L).toLong(), base, cap)
     }
 
+    // IMF-fixdate only; others fall through to backoff
+    // https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Date#syntax
     val formatter = SimpleDateFormat("EEE, dd MMM yyyy HH:mm:ss zzz", Locale.US)
     formatter.timeZone = TimeZone.getTimeZone("GMT")
     val parsed = runCatching { formatter.parse(raw) }.getOrNull()
--- a/packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
+++ b/packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
@@ -6,6 +6,8 @@ import kotlinx.serialization.json.Json
 import java.text.SimpleDateFormat
 import java.util.Locale
 import java.util.TimeZone
+import kotlin.math.pow
+import kotlin.random.Random
 
 /**
  * Outcome of a single dispatch attempt to the OTLP endpoint. Modeled after the OTLP retry
@@ -124,5 +126,83 @@ object DispatchUtils {
     return runCatching {
       responseJson.decodeFromString(OTServiceResponse.serializer(), body).partialSuccess
     }.getOrNull()
+  }
+
+  // MARK: -- Retry gate + exponential backoff
+
+  /**
+   * Base delay (in ms) for the exponential backoff when the server doesn't supply a
+   * `Retry-After`. The cap (`backoffCapMs`) bounds the worst-case wait so we don't end up
+   * snoozing for hours after a long string of failures.
+   */
+  const val backoffBaseMs: Long = 60_000L
+  const val backoffCapMs: Long = 900_000L
+
+  /**
+   * Computes a backoff delay (in ms) for the next dispatch attempt when the server didn't
+   * supply `Retry-After`. Exponential growth (base × 2^(attempt-1)), capped, with full jitter
+   * so a fleet of devices recovering from the same transient backend outage doesn't
+   * thunder-herd the recovery.
+   *
+   * `attempt` is 1-based; the helper is defensive for `0` / negative inputs and returns 0
+   * rather than producing a negative exponent.
+   */
+  fun computeBackoffDelay(
+    attempt: Int,
+    base: Long = backoffBaseMs,
+    cap: Long = backoffCapMs,
+    random: () -> Double = { Random.nextDouble() }
+  ): Long {
+    if (attempt < 1) return 0L
+    val unjittered = minOf((base.toDouble() * 2.0.pow(attempt - 1)).toLong(), cap)
+    return (unjittered.toDouble() * random()).toLong()
+  }
+
+  /**
+   * Snapshot of the retry-gate state carried across dispatch rounds. `dispatchAfterMs` is the
+   * wall-clock deadline (ms since epoch) before which the dispatch entry point should
+   * short-circuit (set by a retryable response, naturally expires); `consecutiveRetryableFailures`
+   * is the counter that drives `computeBackoffDelay` when the server didn't supply a
+   * `Retry-After`.
+   */
+  data class RetryGateState(
+    val dispatchAfterMs: Long?,
+    val consecutiveRetryableFailures: Int
+  ) {
+    companion object {
+      val initial = RetryGateState(dispatchAfterMs = null, consecutiveRetryableFailures = 0)
+    }
+  }
+
+  /**
+   * Pure helper that computes the next `RetryGateState` after a single dispatch result.
+   *
+   * - `Success` resets the counter to 0 and leaves the gate alone. The gate either already
+   *   expired (we wouldn't have dispatched otherwise) or was never set — either way, success
+   *   doesn't introduce a new pause.
+   * - `NonRetryable` also resets the counter. A permanent drop isn't a sign that the server
+   *   is unhealthy and shouldn't pause subsequent rounds.
+   * - `Retryable` increments the counter and sets the gate to `now + delay`, where `delay`
+   *   is the server-supplied `retryAfterMs` if present, otherwise `backoff(nextCount)`.
+   *
+   * `backoff` is injected so tests can drive it deterministically without going through
+   * `computeBackoffDelay`'s random source.
+   */
+  fun nextRetryGateState(
+    result: DispatchResult,
+    currentState: RetryGateState,
+    now: Long,
+    backoff: (Int) -> Long
+  ): RetryGateState = when (result) {
+    is DispatchResult.Success, is DispatchResult.NonRetryable ->
+      currentState.copy(consecutiveRetryableFailures = 0)
+    is DispatchResult.Retryable -> {
+      val nextCount = currentState.consecutiveRetryableFailures + 1
+      val delay = result.retryAfterMs ?: backoff(nextCount)
+      RetryGateState(
+        dispatchAfterMs = now + delay,
+        consecutiveRetryableFailures = nextCount
+      )
+    }
   }
 }
--- a/packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
+++ b/packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
@@ -71,6 +71,21 @@ object DispatchUtils {
   }
 
   /**
+   * Whether the dispatch caller should remove the just-sent pending IDs from the queue.
+   *
+   * - `Success` removes them — the rows have been accepted by the server.
+   * - `NonRetryable` ALSO removes them — the server has refused these rows permanently, so
+   *   retrying would produce the same answer; removing them drops the batch so it can't
+   *   wedge subsequent rounds. This is the acceptance-criterion behavior: a 400/403 must
+   *   not be re-sent on the next cycle.
+   * - `Retryable` keeps them so the next dispatch round picks the same rows up again.
+   */
+  fun shouldRemovePending(result: DispatchResult): Boolean = when (result) {
+    is DispatchResult.Success, is DispatchResult.NonRetryable -> true
+    is DispatchResult.Retryable -> false
+  }
+
+  /**
    * Parses an HTTP `Retry-After` header into a delay in milliseconds from now. Accepts both
    * formats permitted by RFC 7231: an integer delta-seconds, or an HTTP-date.
    * Returns `null` if the header is absent or unparseable. Clamps negative / past values to 0
--- a/packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
+++ b/packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
@@ -2,6 +2,7 @@
 
 package expo.modules.observe
 
+import expo.modules.appmetrics.utils.TimeUtils
 import kotlinx.serialization.json.Json
 import java.text.SimpleDateFormat
 import java.util.Locale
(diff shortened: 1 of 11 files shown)


douglowder:
* **#47160** <a href="https://app.graphite.com/github/pr/expo/expo/47160?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/47160?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#47159** <a href="https://app.graphite.com/github/pr/expo/expo/47159?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
```diff
@@ -0,0 +1,208 @@
...
+   * `computeBackoffDelay`'s random source.
+   */
+  fun nextRetryGateState(
+    result: DispatchResult,
+    currentState: RetryGateState,
+    now: Long,
+    backoff: (Int) -> Long
+  ): RetryGateState = when (result) {
+    is DispatchResult.Success, is DispatchResult.NonRetryable ->
+      currentState.copy(consecutiveRetryableFailures = 0)
```
Ubax, line 236: [🏗️ **design**] The cross-signal interaction here is worth a hard look. `metrics` and `logs` share one `consecutiveRetryableFailures` counter (single `retryGateState` in `BaseObservabilityManager`) and dispatch sequentially — metrics first, then logs.

Trace "metrics healthy, logs endpoint degraded":
- Cycle N: metrics → `Success` → this branch resets counter to 0. Then logs → `Retryable` → counter = 1, `backoff(1)`.
- Gate expires; cycle N+1: metrics runs first → `Success` → counter back to 0. Then logs → `Retryable` → counter = 1 again, `backoff(1)`.

The logs signal is stuck at `backoff(1)` forever — the exponential escalation that the OTLP spec asks for never happens, because the other signal's success keeps zeroing the counter. (A *full* outage hides this, since the later signal is gate-blocked before it can run; the masking specifically bites partial degradation where one endpoint is healthy and the other isn't.)

Sharing the gate *deadline* across signals is defensible ("server told us to slow down, so pause both"). Sharing the failure *counter* — and letting either signal's success reset it — is the part that breaks backoff. Consider a per-signal counter (or only resetting on success of the signal that last set the gate). iOS has the same structure, so this likely wants fixing in both.

packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
```diff
@@ -0,0 +1,208 @@
...
+    is DispatchResult.Retryable -> {
+      val nextCount = currentState.consecutiveRetryableFailures + 1
+      val delay = result.retryAfterMs ?: backoff(nextCount)
```
Ubax, line 239: [💡 suggestion] The computed backoff is capped at `backoffCapMs` (900 s), but a server-supplied `retryAfterMs` flows straight through here uncapped. `parseRetryAfter` only clamps the *lower* bound to 0; there's no upper bound. A server returning `Retry-After: 99999999` (or an HTTP-date far in the future) pins `dispatchAfterMs` arbitrarily far out, and since the gate is in-memory with no ceiling, dispatch is effectively dead until the process restarts.

This is the one spot where untrusted server input directly controls how long the client goes silent. Recommend clamping the honored delay to `backoffCapMs` (or a dedicated max): `min(result.retryAfterMs ?: backoff(nextCount), backoffCapMs)`. Same gap likely exists on iOS.

packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
```diff
@@ -76,20 +77,61 @@ class BaseObservabilityManager(
...
   )
 
+  /**
+   * In-memory retry-gate state. Reset implicitly when the process restarts — a relaunch
+   * usually means enough time passed that the transient cause has cleared anyway, and
+   * persisting the gate would mean an extra disk write on every retryable response. The gate
+   * is shared between the metrics and logs paths so when the server asks us to slow down on
+   * one signal, the other one waits too.
+   */
+  private var retryGateState: DispatchUtils.RetryGateState = DispatchUtils.RetryGateState.initial
```
Ubax: [🏗️ **design**] `retryGateState` is new shared mutable state, and `applyRetryOutcome` does a read-modify-write on it, but there's no synchronization. `EventDispatcher` still carries `// TODO: Make this safe for concurrent usage`, and the two dispatch entry points here can genuinely overlap: `dispatchEvents` is an `AsyncFunction Coroutine` callable from JS at any time, while `scheduleBackgroundDispatch` enqueues a WorkManager job — an app foreground/background transition can have both in flight against the same manager instance.

Before this PR the only shared state was the pending-IDs stores (each presumably DB-backed); a torn/last-writer-wins `retryGateState` is a new hazard. Worst case is benign (a dropped gate update, a double dispatch) given telemetry is best-effort — but please either guard it (the existing `currentTimeMs` seam makes a small `@Synchronized` accessor or `Mutex` easy) or extend the TODO to explicitly cover the gate so the assumption is documented rather than silent.

packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
```diff
@@ -0,0 +1,208 @@
...
+   * formats permitted by RFC 7231: an integer delta-seconds, or an HTTP-date.
+   * Returns `null` if the header is absent or unparseable. Clamps negative / past values to 0
+   * so a misbehaving server can't make us schedule the next dispatch in the past.
+   */
+  fun parseRetryAfter(header: String?, now: Long = System.currentTimeMillis()): Long? {
+    val raw = header?.trim() ?: return null
+    if (raw.isEmpty()) return null
+
+    raw.toLongOrNull()?.let { return maxOf(it * 1000L, 0L) }
+    raw.toDoubleOrNull()?.let { return maxOf((it * 1000L).toLong(), 0L) }
```
Ubax: [🐧 nit] Minor robustness: `raw.toDoubleOrNull()` accepts `"NaN"`, `"Infinity"`, and scientific notation (`"1e9"`). `NaN` collapses to 0 (harmless), but `"1e30"` parses to a huge delay — another path to the uncapped-Retry-After issue flagged on `nextRetryGateState`. If you add the upper clamp there, this is covered; otherwise consider rejecting non-finite / out-of-range doubles here. Not blocking.

packages/expo-observe/android/src/main/java/expo/modules/observe/EventDispatcher.kt
```diff
@@ -96,21 +97,43 @@ class EventDispatcher(
...
+      retryAfterHeader = retryAfterHeader,
+      responseBody = responseBody,
+      bodyExcerpt = { DispatchUtils.bodyExcerpt(responseBody) }
+    )
 
-    continuation.resume(response.code in 200..299)
-  }
+    when (result) {
+      is DispatchResult.Success ->
+        Log.d(OBSERVE_TAG, "Server responded successfully with ${response.code} and data: $responseBody")
```
Ubax, line 126: [🐧 nit] nit: the `Success` and `Retryable` branches log the full `responseBody` (only `NonRetryable` goes through the 512-byte `bodyExcerpt`). For a chatty collector or a large partial_success body these are unbounded log lines, and `Log.d(OBSERVE_TAG, body)` above already logs the entire request body. Probably fine for a debug-level diagnostic, but worth bounding the response body in the success/retryable lines too for consistency with the excerpt you deliberately added for errors.

packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
```diff
@@ -0,0 +1,208 @@
...
+
+/**
+ * Outcome of a single dispatch attempt to the OTLP endpoint. Modeled after the OTLP retry
+ * guidance: a 3-way classification distinguishes transient failures (retry the same batch)
+ * from permanent ones (drop the batch so it can't wedge the queue).
+ *
+ * `Retryable.retryAfterMs` carries the parsed `Retry-After` header value in milliseconds (or
+ * `null` when the server didn't supply one). `NonRetryable.reason` carries a short diagnostic
+ * string for the `warn`-level log line.
+ */
```
Ubax, line 31: Can you add link to the specification describing this?

I think this is the one - https://opentelemetry.io/docs/specs/otlp/#otlphttp-response

packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
```diff
@@ -0,0 +1,208 @@
...
+sealed class DispatchResult {
```
Ubax, line 32: What do you think about having 4 way model here to map fully to otel spec:
- Success
- PartialSuccess
- Retryable
- NonRetryable

This would make it more obvious for the consumer, that there was partial success

packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
```diff
@@ -0,0 +1,208 @@
...
+    return if (body.length <= limit) body else body.substring(0, limit)
+  }
+
+  private val responseJson = Json { ignoreUnknownKeys = true }
+
+  private fun parsePartialSuccess(body: String): OTPartialSuccess? {
+    if (body.isBlank()) return null
+    return runCatching {
+      responseJson.decodeFromString(OTServiceResponse.serializer(), body).partialSuccess
+    }.getOrNull()
```
Ubax: This won't be needed when we have the ParialSuccess class. We can the rely fully on the kotlin compiler rather then parsing strings

packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
```diff
@@ -0,0 +1,208 @@
...
+   * Base delay (in ms) for the exponential backoff when the server doesn't supply a
+   * `Retry-After`. The cap (`backoffCapMs`) bounds the worst-case wait so we don't end up
+   * snoozing for hours after a long string of failures.
+   */
+  const val backoffBaseMs: Long = 60_000L
+  const val backoffCapMs: Long = 900_000L
+
+  /**
+   * Computes a backoff delay (in ms) for the next dispatch attempt when the server didn't
+   * supply `Retry-After`. Exponential growth (base × 2^(attempt-1)), capped, with full jitter
```
Ubax, line 176: Can we add a link to the description of the full jitter algorithm - https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/ ?

I didn't know that this is a well-known algorithm, so having the link here will make it more obvious to other readers

packages/expo-observe/android/src/main/java/expo/modules/observe/EventDispatcher.kt
```diff
@@ -28,60 +27,62 @@ class EventDispatcher(
...
+   * POSTs `events` to the metrics endpoint and classifies the response per the OTLP retry
+   * spec. Non-throwing: payload-encoding errors collapse to `NonRetryable` (same bytes will
+   * fail again), transport errors collapse to `Retryable(null)` (transient by definition).
+   */
+  suspend fun dispatch(events: List<Event>): DispatchResult =
     suspendCancellableCoroutine { continuation ->
       if (events.isEmpty()) {
-        continuation.resume(false)
+        // Empty input isn't a server response — keep the old "nothing to do" semantics by
+        // signaling success so the call site removes (nothing) and moves on.
```
Ubax:
```suggestion
        // Empty input isn't a server response - signal success.
```

packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
```diff
@@ -76,20 +77,61 @@ class BaseObservabilityManager(
   private val isDebugBuild: Boolean = false,
   private val deterministicUniformValueProvider: () -> Double = {
     EASClientID.deterministicUniformValue(EASClientID(context).uuid)
-  }
+  },
+  private val currentTimeMs: () -> Long = { System.currentTimeMillis() }
```
Ubax: Can you add `System.currentTimeMillis()` to `TimeUtils`, so that all time logic is in a single utility?

packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
```diff
@@ -76,20 +77,61 @@ class BaseObservabilityManager(
...
+   * one signal, the other one waits too.
+   */
+  private var retryGateState: DispatchUtils.RetryGateState = DispatchUtils.RetryGateState.initial
+
+  /**
+   * Returns true and logs when an active retry gate suppresses this dispatch round. Mirrors
+   * the `dispatch()` entry-point gate check on the iOS side.
+   */
+  private fun retryGateBlocks(): Boolean {
+    val until = retryGateState.dispatchAfterMs ?: return false
```
Ubax: What do you think about having separate gates for metrics and logs. Theoretically one endpoint can work when the other does not

packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
douglowder: Sure, no problem

packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
douglowder: Good idea. Even though we control the server endpoints, we should still guard against bad values in this header. I'll fix this.

packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
douglowder: This is a good catch, but I'm not sure yet what the correct design is then. I'll map out some different failure scenarios and make sure we don't have any deadlocks like this.

packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
douglowder: In this case, we trust the server, but still, we should impose sensible limits here, I agree.

packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
douglowder: Agreed.

packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
douglowder: Fixed

packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
douglowder: Done on both platforms

packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
douglowder: Fixed

packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
douglowder: Fixed

packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
```diff
@@ -0,0 +1,246 @@
...
+import kotlinx.serialization.json.Json
+import java.text.SimpleDateFormat
+import java.util.Locale
+import java.util.TimeZone
+import kotlin.math.pow
+import kotlin.random.Random
+
+/**
+ * Outcome of a single dispatch attempt to the OTLP endpoint. Four cases, modeled after the
+ * OTLP retry guidance:
```
douglowder:
```suggestion
 * OTLP retry guidance (see https://opentelemetry.io/docs/specs/otlp/#otlphttp-response):
```

packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
```diff
@@ -0,0 +1,246 @@
...
+   */
+  const val backoffBaseMs: Long = 60_000L
+  const val backoffCapMs: Long = 900_000L
+
+  /**
+   * Computes a backoff delay (in ms) for the next dispatch attempt when the server didn't
+   * supply `Retry-After`. Exponential growth (base × 2^(attempt-1)), capped, with full jitter
+   * so a fleet of devices recovering from the same transient backend outage doesn't
+   * thunder-herd the recovery.
+   *
```
douglowder, line 179:
```suggestion
   *
   * https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
   *
```

packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
```diff
@@ -0,0 +1,248 @@
...
+    val raw = header?.trim() ?: return null
+    if (raw.isEmpty()) return null
+
+    raw.toLongOrNull()?.let { return clampToBounds(it * 1000L, base, cap) }
+    raw.toDoubleOrNull()?.let {
+      if (!it.isFinite()) return null
+      return clampToBounds((it * 1000L).toLong(), base, cap)
+    }
+
+    val formatter = SimpleDateFormat("EEE, dd MMM yyyy HH:mm:ss zzz", Locale.US)
```
Ubax, line 137:
```suggestion
    // Following the HTTP Date format
    // https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Date#syntax
    val formatter = SimpleDateFormat("EEE, dd MMM yyyy HH:mm:ss 'GMT'", Locale.US)
```

packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
```diff
@@ -0,0 +1,248 @@
...
+    formatter.timeZone = TimeZone.getTimeZone("GMT")
+    val parsed = runCatching { formatter.parse(raw) }.getOrNull()
+    if (parsed != null) {
+      return clampToBounds(parsed.time - now, base, cap)
```
Ubax: Can you move this implementation to TimeUtils `parsed.time - now`?

Something like `fun millisUntil(to: Date, from: Date = Date()): Long = to.time - from.time`.

Then you can remove the `now` argument from this class and in the test you can 

```kt
mockkObject(TimeUtils)
....
val capturedTo = slot<Date>()
every { TimeUtils.millisUntil(capture(capturedTo), any()) } returns 123L
....
verify {
    TimeUtils.millisUntil(any(), any())
}

assertEquals(targetDate, capturedTo.captured)
```

packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
douglowder: I added a similar comment here.

packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
douglowder: Done.

packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
douglowder: Fixed.

packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
douglowder: Removed.

packages/expo-observe/android/src/main/java/expo/modules/observe/EventDispatcher.kt
```diff
@@ -96,21 +96,50 @@ class EventDispatcher(
...
+
+    when (result) {
+      is DispatchResult.Success ->
+        Log.d(OBSERVE_TAG, "Server responded successfully with ${response.code} and data: $responseBody")
+      is DispatchResult.PartialSuccess ->
+        Log.w(
+          OBSERVE_TAG,
+          "Server responded with ${response.code} (partial success, rejected " +
+            "${result.partial.rejectedCount}: ${result.partial.errorMessage ?: "no error message"}) " +
+            "and data: $responseBody"
```
douglowder:
```suggestion
            "${result.partial.rejectedCount}: ${result.partial.errorMessage ?: "no error message"}) "
```

packages/expo-observe/android/src/main/java/expo/modules/observe/EventDispatcher.kt
douglowder: I think for the partial success, we can just omit the response body in the warning

Ubax (approved):

PR #46969, opened 2026-06-16
[autolinking] Add support for compile-only inline module files

behenate:
# Why

This feature is needed for adding AppIntent support to expo-sdk - we need to add native files, which won't be treated as ExpoModule holders.  Currently inline modules treat every kotlin/swift file as a module holder.

# How

Autolink non-module files without treating them as a module. Use a regular expression to determine whether the file has a module definition

# Test Plan

Tested in BareExpo in my AppIntent PoC

PR #47169, opened 2026-06-23
[expo-type-information] Update CLI to work with compile only inline modules

HubertBer:
# Why
In #46969 compile online inline modules support is added (inline modules without NativeModule) and it doesn't work with the current `inline-modules-interface` command.

# How
Updated the command to track compile onlt files and use them for generating types.

# Test Plan
- [x] Checked the command manually

# Checklist


behenate (approved):
Just the thing we discussed offline

PR #47180, opened 2026-06-23
[gh][docs] Add docs upstream script

amandeepmittal:
# Why

Fix ENG-22049

# How

- Adds a scheduled `Docs Upstream Sync` GitHub Action that runs the CI-safe generators daily through a new `pnpm upstream-sync` script (`docs/scripts/upstream-sync.ts`), which reverts `fetchedAt`-only timestamp noise and prints a JSON summary.

- The `expo-mcp-sync` generator is excluded since it needs an interactive OAuth flow. 

- The generated `permissions/data/android.json` is added to the `oxfmt`/`oxlint` ignore lists.

# Test Plan

N/A

# Checklist

- [ ] I added a `changelog.md` entry and rebuilt the package sources according to [this short guide](https://github.com/nicknisi/dotfiles/wiki/Pull-Request-Guidelines).

### DIFF
--- a/docs/scripts/upstream-sync.ts
+++ b/docs/scripts/upstream-sync.ts
@@ -47,12 +47,14 @@ for (const { label, script } of SYNCS) {
   }
 }
 
+const TIMESTAMP_FIELDS = ['fetchedAt', 'scrapedAt'];
+
 const changed: string[] = [];
 for (const { code, file } of status()) {
   const realDiff = git('diff', '--', file)
     .split('\n')
     .filter(line => /^[+-]/.test(line) && !/^(\+{3}|-{3})/.test(line))
-    .some(line => !line.includes('fetchedAt'));
+    .some(line => !TIMESTAMP_FIELDS.some(field => line.includes(field)));
   if (code.includes('?') || realDiff) {
     changed.push(file);
   } else {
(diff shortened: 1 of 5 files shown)


docs/scripts/upstream-sync.ts
```diff
@@ -0,0 +1,78 @@
...
+    failed.push(label);
+  }
+}
+
+const changed: string[] = [];
+for (const { code, file } of status()) {
+  const realDiff = git('diff', '--', file)
+    .split('\n')
+    .filter(line => /^[+-]/.test(line) && !/^(\+{3}|-{3})/.test(line))
+    .some(line => !line.includes('fetchedAt'));
```
kadikraman: The Android permissions generator's timestamp field is `scrapedAt`, but the noise filter only matches `fetchedAt`, so it looks like this will raise a no-op PR every single day. Not sure if any of the others could have the same issue.

amandeepmittal: Good catch!  I checked the rest: fetchedAt and scrapedAt are the only two timestamp fields across all the generators , so I pulled the names into a `TIMESTAMP_FIELDS` list and added `scrapedAt`.

kadikraman (approved):

PR #47192, opened 2026-06-23
[cli] Guard DevTools plugin request error handler against started responses

krystofwoldrich:
# Why

When a DevTools plugin's fetch `requestHandler` threw *after* the response had already started streaming (headers sent, stream ended/destroyed, or the request aborted), the error handler tried to write a 500 fallback onto an already-committed response, throwing a secondary error.

# How

Before writing the 500 fallback, the error handler now checks `res.headersSent`, `res.writableEnded`, `res.destroyed`, and `request?.signal.aborted`. If any of those hold, the response is already committed/gone and the handler returns without touching it.

### DIFF
--- a/packages/@expo/cli/src/start/server/middleware/DevToolsPluginMiddleware.ts
+++ b/packages/@expo/cli/src/start/server/middleware/DevToolsPluginMiddleware.ts
@@ -77,9 +77,10 @@ export class DevToolsPluginMiddleware extends ExpoMiddleware {
     urlInPluginRoot: string
   ): Promise<boolean> {
     const originalUrl = req.url;
+    let request: Request | undefined;
     try {
       req.url = urlInPluginRoot;
-      const request = convertRequest(req as http.IncomingMessage, res as http.ServerResponse);
+      request = convertRequest(req as http.IncomingMessage, res as http.ServerResponse);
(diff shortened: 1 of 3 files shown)


packages/@expo/cli/src/start/server/middleware/DevToolsPluginMiddleware.ts
kitten, line 89: unrelated, but haven't seen this cast here before. This looks a bit suspicious, since we're casting same-to-same?

packages/@expo/cli/src/start/server/middleware/DevToolsPluginMiddleware.ts
```diff
@@ -89,6 +90,9 @@ export class DevToolsPluginMiddleware extends ExpoMiddleware {
       return true;
     } catch (error: any) {
       debug('DevTools plugin server request failed: %O', error);
+      if (res.headersSent || res.writableEnded || res.destroyed || request?.signal.aborted) {
```
kitten, line 93: fwiw, it feels like we're trying to actually separate `await handler` from `await respond`. While streaming can play a role here, it feels potentially cleaner to separate this into two catch handlers. Not necessarily immediately necessary but would potentially be cleaner.

kitten (approved):

PR #47193, opened 2026-06-23
[core][android] Precompile and ship native libraries

lukmccall:
# Why

`libexpo-modules-core.so` is the most expensive part of an Android build that uses Expo - every consumer recompiles this C++ from source on every clean build, CI run and EAS Build. The code is identical for a given package + React Native version, so there's no reason to compile it on each machine.

This ships precompiled, stripped, and compressed `.so` files inside the npm package, so consumers import them instead of building them. The worklets integration is the one piece that must stay source-built.

# How

**CMake - import instead of compile**
- New `android/cmake/prebuilt.cmake`: when Gradle passes `-DEXPO_PREBUILT_NATIVE_LIBS_DIR`, it defines `expo-modules-jsi` and `expo-modules-core` as `SHARED IMPORTED` targets and sets `EXPO_JSI_PREBUILT` / `EXPO_CORE_PREBUILT`.
- `CMakeLists.txt` includes `prebuilt.cmake` first, then only includes `jsi.cmake` / `main.cmake` when the corresponding target wasn't imported - so the heavy compilation is skipped per-target, per-ABI, with a clean fall back to source when an artifact is missing.
- `expo-modules-jsi` is now `SHARED` (was `STATIC`), so it can be shipped and imported on its own.
- `expo-modules-worklets` is never prebuilt - it always compiles from source and links the imported core/jsi.

**Producer - `./gradlew :expo-modules-core:cachePrebuiltNativeLibs`**
- Builds debug + release from source, then reads the deterministic `intermediates/merged_native_libs/<variant>/…/out/lib` output.
- Strips each `.so` with `llvm-strip --strip-all -o`.
- Bundles every ABI x variant into a single **`prebuilt/native-libs.tar.xz` + `prebuilt/metadata.json`.

**Consumer - `decompressPrebuiltNativeLibs`**
- Unpacks the archive into the build dir, then CMake imports the `.so` and AGP packages them via per-variant `jniLibs` source sets.
- Wired off the single `preBuild` lifecycle anchor.

**Size:** all 4 ABIs x debug + release ship as one **~1.5 MB** `tar.xz` (vs ~16 MB raw).

# Test Plan

- bare-expo


lukmccall:
I've committed the precompiled archive because it's relatively small and I'd love to test it. Later, we can remove it.

packages/expo-modules-core/android/cmake/jsi.cmake
```diff
@@ -3,7 +3,7 @@ file(GLOB android_sources_jsi "${ANDROID_SRC_DIR}/jsi/*.cpp")
 
 add_library(
   expo-modules-jsi
-  STATIC
+  SHARED
```
tsapeta, line 6: [🏗️ **design**] `expo-modules-jsi` goes from STATIC to SHARED here, which applies to **all** builds, not just the prebuilt path. Previously its code was linked into `libexpo-modules-core.so`; now it ships as a standalone `libexpo-modules-jsi.so` that has to be loaded at runtime.

Nothing in the codebase calls `SoLoader.loadLibrary("expo-modules-jsi")` — loading relies entirely on SoLoader resolving the new `DT_NEEDED` entry transitively when `expo-modules-core` is loaded (`MainRuntimeInstaller.kt:63`, `FabricComponentsRegistry.kt:59`). SoLoader does recursive dependency resolution, so this is expected to work *provided the `.so` is packaged in `jniLibs`*, but it's the riskiest behavior change in the PR and it now hits every consumer. Worth explicitly confirming on-device in both a debug and a **release** build (the test plan says bare-expo — please make sure release packaging + load is part of it). If transitive resolution ever proves flaky, the safe fix is an explicit `SoLoader.loadLibrary("expo-modules-jsi")` before the core load.

packages/expo-modules-core/android/build.gradle
```diff
@@ -78,6 +92,36 @@ if (workletsProject != null) {
...
+def isPrebuiltUsable = {
+  if (!prebuiltMetadataFile.exists() || !prebuiltNativeLibsArchive.exists()) {
+    return false
+  }
+  def metadata = new JsonSlurper().parseText(prebuiltMetadataFile.text)
+  if (metadata.libraryVersion != version) {
+    logger.warn("[expo-modules-core] Ignoring prebuilt libs: built for v${metadata.libraryVersion}, package is v${version} - building from source")
+    return false
+  }
+  if (metadata.reactNativeVersion != reactNativeVersionString) {
```
tsapeta, line 116: [💡 suggestion] The RN-version gate is an exact `major.minor.patch` match (via the new `Version.toString()`), so an RN **patch** bump (e.g. `0.86.0` → `0.86.1`) invalidates the prebuilt and silently falls back to compiling from source — losing the speedup for that case. That's the safe direction (no risk of an ABI mismatch), so this is fine as-is; just flagging it as a deliberate tradeoff worth being aware of, since C++ ABI is usually stable across RN patch releases and you could widen the gate to minor-only later if the fallback rate turns out to be high.

Separately: `Version.fromString` strips pre-release suffixes, so `0.86.0-rc.1` and `0.86.0` both normalize to `0.86.0` and compare equal. An RC can differ in ABI from the final, so a prebuilt cached against an RC could be considered "usable" against the release (and vice versa). The metadata RN string is produced the same way, so it's self-consistent within one cache, but cross-prerelease reuse is theoretically possible.

packages/expo-modules-core/android/build.gradle
```diff
@@ -188,6 +236,14 @@ android {
         }
       }
     }
+
+    // Package the decompressed prebuilt .so into each variant. The layout `<variant>/<abi>/lib*.so`
+    // matches what AGP expects from a jniLibs source set.
+    if (usePrebuiltNativeLibs) {
+      prebuiltVariants.each { variant ->
+        getByName(variant).jniLibs.srcDir(new File(decompressedNativeLibsDir, variant))
```
tsapeta, line 244: [💡 suggestion] `jniLibs.srcDir(decompressedNativeLibsDir/<variant>)` adds *all* ABIs present in the archive (the metadata ships all 4). The `abiFilters` in the `externalNativeBuild.cmake` block only constrains the CMake build's ABIs — it does not filter `jniLibs` packaging. So a consumer that narrows ABIs for size (e.g. `reactNativeArchitectures=arm64-v8a`, common in production EAS builds) still gets the prebuilt `armeabi-v7a/x86/x86_64` `.so`s pulled into this AAR's jniLibs.

In practice the app-level abi split/filter usually strips the unused ABIs from the final APK, so this is intermediate bloat rather than a shipped-size regression — but it's worth confirming, and if you want to be tight about it you could intersect the unpacked ABIs with `reactNativeArchitectures()` when wiring the source sets.

packages/expo-modules-core/android/build.gradle
```diff
@@ -318,3 +374,206 @@ def generateStubPCHTask = tasks.register("generateStubPCH") {
...
+    cacheStagingDir.deleteDir()
+
+    prebuiltMetadataFile.text = JsonOutput.prettyPrint(JsonOutput.toJson([
+      libraryVersion    : libraryVersion,
+      reactNativeVersion: reactNativeVersionString,
+      variants          : variantDescriptors.collect { it.name },
+      abis              : allAbis as List,
+      soFiles           : prebuiltNativeLibNames as List,
+      stripped          : stripBinary != null,
+      createdAt          : Instant.now().toString(),
```
tsapeta, line 537: [🐧 nit] `createdAt: Instant.now()` makes the producer output non-deterministic: re-running `cachePrebuiltNativeLibs` on identical inputs produces a `metadata.json` that differs only by timestamp, churning a committed file in every regeneration diff. Since this field isn't part of the usability check (only `libraryVersion` + `reactNativeVersion` are), consider dropping it (or moving it out of the committed artifact) so regenerations are reproducible and don't create noise in `git diff`.

packages/expo-modules-core/android/build.gradle
```diff
@@ -318,3 +374,206 @@ def generateStubPCHTask = tasks.register("generateStubPCH") {
 tasks.register("prepareKotlinBuildScriptModel") {
   dependsOn(generateStubPCHTask)
 }
+
+def prebuiltNativeLibNames = ["libexpo-modules-core.so", "libexpo-modules-jsi.so"]
+
+// Helpers for the cachePrebuiltNativeLibs task. They are `static` so the task action can call them
+// without capturing the build script, which keeps the task configuration-cache compatible.
+class PrebuiltNativeLibs {
```
tsapeta, line 382: [🐧 nit] The comment says these helpers are `static` to keep the task "configuration-cache compatible," but the `doLast` action of `cachePrebuiltNativeLibs` still captures script-level `def`s (`variantDescriptors`, `stripBinary`, `prebuiltNativeLibsDir`, `cacheStagingDir`, etc.). Capturing build-script state in a task action is the usual thing that *breaks* the configuration cache, so the `static` helpers alone don't guarantee compatibility. This is a producer-only task (not run in normal consumer builds), so the impact is low — but if config-cache compatibility for this task is actually a goal, it's worth verifying with `--configuration-cache` rather than relying on the helper being static.

chrfalch (commented):
I see we're committing the precompiled libs - why aren't these just built on CI or when creating npm packages? Maybe I'm missing something - just wanted to ask :)

lukmccall:
> I see we're committing the precompiled libs - why aren't these just built on CI or when creating npm packages? Maybe I'm missing something - just wanted to ask :) 


I've committed the precompiled archive because it's relatively small and I'd love to test it. Later, we can remove it.

packages/expo-modules-core/android/cmake/jsi.cmake
lukmccall: The `expo-modules-jsi` has to be shared because otherwise it can't be consumed by the `expo-modules-worklet` target.

We don't have to load it manually, as it will be loaded by the `expo-modules-core` automatically. It's not risky.

packages/expo-modules-core/android/build.gradle
lukmccall: RN isn't ABI compatible. We don't know if patches will work with our precompiled artifact. We have to build it for each RN version

packages/expo-modules-core/android/build.gradle
lukmccall: That's not true, because those libraries are added under `<abi>/lib...`, so if you're building for only `arm64-v8a`, only those files will be bundled into the final APK.

packages/expo-modules-core/android/build.gradle
lukmccall: I'm okay with that

packages/expo-modules-core/android/build.gradle
lukmccall: I've tested it with the configuration cache enabled and it was working fine.

packages/expo-modules-core/android/build.gradle
lukmccall, line 537:
```suggestion
      createdAt         : Instant.now().toString(),
```

tsapeta (approved):

Kudo:
- could we store debug symbols somewhere rather than just striping? it would help to debug native crash
- committing precompiled archive one time is fine but if that's something we will keep updating (for instance, each react-native upgrade). it would be better to generate precompiled archive from publishing pipeline

lukmccall:
> * could we store debug symbols somewhere rather than just striping? it would help to debug native crash

I agree, but that would require a lot of changes in our publishing process. We can iterate on that. 

> * committing precompiled archive one time is fine but if that's something we will keep updating (for instance, each react-native upgrade). it would be better to generate precompiled archive from publishing pipeline

It's a temporary state for testing the logic internally on a larger scale. We will quickly migrate to building via the publish script. I just want to see if we will face any issues with the current setup.

Kudo:
sounds good as you know these concern and may address later. i'm approving this pr

Kudo (approved):

chrfalch (approved):

alanjhughes (approved):

PR #47197, opened 2026-06-24
[ui] Fix scrollable inside community BottomSheet not scrolling when nested in a list

intergalacticspacehighway:
# Why?

Reported on slack - https://exponent-internal.slack.com/archives/C01D24DMP98/p1782232486941459?thread_ts=1782162662.014889&cid=C01D24DMP98

Nesting a BottomSheet inside a FlatList breaks scroll view scrolling.

# How?

When virtualised list (RN Flatlist) is nested within another FlatList, it [renders](https://github.com/react/react-native/blob/5c197fb303ed0d975482757fefb7ed38349601b6/packages/virtualized-lists/Lists/VirtualizedList.js#L1291) a regular View as Scroll component which breaks scrolling. Nested same direction FlatLists are considered bad as it can break virtualisation but the react context check that RN is doing to identify nesting can break some genuine cases. e.g. User has a FlatList with a button child that opens BottomSheet with a FlatList. In such cases, FlatList is not exactly nested inside a FlatList as natively sheets are mounted in a separate window but the JS context check makes it false positive.

# Test

Tested the repro shared in the report.

### DIFF
--- a/packages/expo-ui/CHANGELOG.md
+++ b/packages/expo-ui/CHANGELOG.md
@@ -34,6 +34,7 @@
 
 ### 🐛 Bug fixes
 
+- [iOS][android] Fix a scrollable (e.g. `FlatList`) inside `community/bottom-sheet` not scrolling when the `<BottomSheet>` is mounted inside another same-orientation list (such as a `FlatList` header). ([#47197](https://github.com/expo/expo/pull/47197) by [@nishan](https://github.com/intergalacticspacehighway))
 - [iOS] Fix the `community/datetime-picker` field collapsing to zero width — invisible and untappable — when its parent doesn't stretch it (e.g. `alignItems: 'center'`). ([#47033](https://github.com/expo/expo/pull/47033) by [@nishan](https://github.com/intergalacticspacehighway))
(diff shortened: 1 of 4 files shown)


packages/expo-ui/CHANGELOG.md
```diff
@@ -34,6 +34,7 @@
 
 ### 🐛 Bug fixes
 
+- [iOS][android] Fix a scrollable (e.g. `FlatList`) inside `community/bottom-sheet` not scrolling when the `<BottomSheet>` is mounted inside another same-orientation list (such as a `FlatList` header). ([#47197](https://github.com/expo/expo/pull/47197) by [@nishan](https://github.com/intergalacticspacehighway))
```
Kudo:
```suggestion
- [iOS][Android] Fix a scrollable (e.g. `FlatList`) inside `@expo/ui/community/bottom-sheet` not scrolling when the `<BottomSheet>` is mounted inside another same-orientation list (such as a `FlatList` header). ([#47197](https://github.com/expo/expo/pull/47197) by [@nishan](https://github.com/intergalacticspacehighway))
```

Kudo (approved):

PR #47207, opened 2026-06-24
[app-intents][ios][1/n] Create `expo-app-intents` module

behenate:
# Why

Adds `expo-app-intents module`, which provides a framework for adding App Intent support to Expo apps

# How

We can't really provide this functionality fully from JS. App Intents are compiled from static swift classes at build time. Therefore our best shot at adding App Intent and Apple Intelligence support with minimal hassle for the users is allowing them to define the Intents via inline modules. 

We have to define the intents natively, but the data used by the intents can be provided from JS. 
For example: In a restaurant order use case - the Entities, Shortcuts and Intents "shapes" have to be defined natively, but the individual items can be provided at runtime from JS via `setEntityCatalogAsync`.  

The general flow of a received AppIntent:

- An App Intent runs in the app target and calls `AppIntentDispatcher.shared.dispatch(name:params:)`.
- `AppIntentDispatcher` wraps the call as an `AppIntentInvocation` with an id, name, params, and timestamp.
- The dispatcher saves it through `AppIntentInvocationStore` first, so the invocation survives even if JS is not running.
- If JS is alive, `ExpoAppIntentsModule` receives the invocation from the dispatcher’s async stream and emits an onIntent event.
- JS handles live events via `addAppIntentListener()` or `useAppIntents()`, or later reads stored ones with `getPendingInvocationsAsync()`.
- After handling, JS calls `removePendingInvocationAsync(id)` or `clearPendingInvocationsAsync()` to prevent reprocessing.
- For parameterized intents, JS manages searchable values with `setEntityCatalogAsync()`, which stores entities in `AppIntentEntityStore` and refreshes App Shortcuts.

The module provides functions for communication between the native and js side:

- addAppIntentListener(listener) - subscribes to live App Intent invocations while JS is running.
- useAppIntents(handler) - reads pending invocations on mount and listens for new invocations afterward.
- getPendingInvocationsAsync() - returns queued intent invocations that have not been removed yet.
- removePendingInvocationAsync(id) - marks one handled invocation as removed from the pending queue.
- clearPendingInvocationsAsync() clears all queued pending invocations.
- setEntityCatalogAsync(kind, entities) - replaces an entity catalog used by App Intents parameter queries and refreshes shortcuts.
- getEntityCatalogAsync(kind) reads the current entity catalog for a given kind.
- refreshShortcutsAsync() asks iOS to re-evaluate App Shortcut phrases and parameter values.

# Things coming in subsequent PRs

- cli tool for easy setup in an existing project. It will also allow setup with a few examples for learning.
- Docs
- Examples for NCL

# Things we may add in the future

- Apple on-screen intelligence support for ExpoUI via a modifier - I think this should be possible, but I didn't have a chance to try implement it because I'm still on the waitlist for the new Siri, so I would have no way of testing it.
- Non main-target App Intents - For some quick actions that don't require an immediate action of the app it's not worth it to spin up the main app target

# Test Plan

Tested on iPhone 13 and iPhone 17 in BareExpo and a separate test app


packages/expo-app-intents/ios/ExpoAppIntentsModule.swift
```diff
@@ -0,0 +1,69 @@
...
+    if !(await AppIntentDispatcher.shared.requestShortcutsRefresh()) {
+      throw ShortcutsRefreshUnavailableException()
+    }
+  }
+
+  @MainActor
+  private func sendIntentEvent(_ invocation: AppIntentInvocation) {
+    sendEvent("onIntent", invocation.toDict())
+  }
+}
```
tsapeta, line 69: Would you like to try out Expo Modules v2 in this module? 😏

behenate: image

packages/expo-app-intents/ios/Tests/AppIntentDispatcherTests.swift
```diff
@@ -0,0 +1,100 @@
+import XCTest
```
tsapeta, line 1: Why the old `XCTest`? 😞 Use Swift Testing framework instead

packages/expo-app-intents/ios/ExpoAppIntents.podspec
```diff
@@ -0,0 +1,33 @@
...
+Pod::Spec.new do |s|
+  s.name           = 'ExpoAppIntents'
+  s.version        = package['version']
+  s.summary        = package['description']
+  s.description    = package['description']
+  s.license        = package['license']
+  s.author         = package['author']
+  s.homepage       = package['homepage']
+  s.platforms      = {
+    :ios => '16.4'
```
tsapeta, line 14: It seems that App Intents are also supported on tvOS and macOS. Can you check it? Maybe it's effortless to support them 😄

packages/expo-app-intents/ios/ExpoAppIntents.podspec
```diff
@@ -0,0 +1,33 @@
...
+
+  s.source_files = "**/*.{h,m,swift}"
+  s.pod_target_xcconfig = {
+    'DEFINES_MODULE' => 'YES',
+    'SWIFT_COMPILATION_MODE' => 'wholemodule'
+  }
+
+  s.exclude_files = 'Tests/'
+  s.test_spec 'Tests' do |test_spec|
+    test_spec.dependency 'ExpoModulesTestCore'
```
tsapeta, line 30: If you use the Testing framework, you won't need `ExpoModulesTestCore`

packages/expo-app-intents/CHANGELOG.md
```diff
@@ -0,0 +1,13 @@
+# Changelog
+
+## Unpublished
+
+### 🛠 Breaking changes
+
+### 🎉 New features
+
+- 🔥🎉Initial release. ([#47207](https://github.com/expo/expo/pull/47207) by [@behenate](https://github.com/behenate))
```
tsapeta, line 9:
```suggestion
- 🔥🎉 Initial release. ([#47207](https://github.com/expo/expo/pull/47207) by [@behenate](https://github.com/behenate))
```

tsapeta (changes requested):
Since this is a new package, can you already enable `swift-format`? See https://github.com/expo/expo/pull/46780

packages/expo-app-intents/ios/ExpoAppIntents.podspec
behenate: A few tests run the module through a real runtime, so it's necessary there. For example `publishing a catalog succeeds when no refresh handler is registered`

packages/expo-app-intents/ios/AppIntentDispatcher.swift
```diff
@@ -0,0 +1,108 @@
+import Foundation
+
+/// Ownership token for a single `AppIntentDispatcher.invocationEvents(for:)` subscription.
+///
+/// Cancelling the read task alone still lets an old module receive invocations after a JavaScript
+/// reload replaces it, so the token prevents it from subscribing at all.
+internal final class AppIntentEventSubscription: @unchecked Sendable {
+  private let lock = NSLock()
```
tsapeta: Use more modern `Mutex` instead

packages/expo-app-intents/ios/AppIntentEntityStore.swift
```diff
@@ -0,0 +1,154 @@
+import ExpoModulesCore
+import Foundation
+
+/// An entity exposed to App Intents parameter queries. JS populates catalogs with
+/// `setEntityCatalogAsync`; app-target `EntityQuery` implementations read them through
+/// `AppIntentEntityStore.shared`.
+///
+/// Catalogs are stored in UserDefaults, so they should stay compact. Apps with large
+/// datasets should publish only the subset needed for Siri and Shortcuts resolution.
+public struct AppIntentEntityRecord: Codable, Record {
```
tsapeta: We could try using `@Record` macro here 😉

packages/expo-app-intents/ios/AppIntentEntityStore.swift
```diff
@@ -0,0 +1,154 @@
...
+    try container.encode(synonyms, forKey: .synonyms)
+  }
+}
+
+public actor AppIntentEntityStore {
+  public static let shared = AppIntentEntityStore()
+
+  private let defaults: UserDefaults
+
+  internal init(defaults: UserDefaults = .standard) {
```
tsapeta: Consider using a custom user defaults suite (`dev.expo.appintents`) instead of the standard one.

packages/expo-app-intents/ios/AppIntentEntityStore.swift
```diff
@@ -0,0 +1,154 @@
...
+          + "setEntityCatalogAsync again."
+      )
+    }
+
+    // Two entities with one id cannot both be resolved, so the id no longer names one entity.
+    var seenIds = Set<String>()
+    for entity in entities {
+      if !seenIds.insert(entity.id).inserted {
+        throw AppIntentEntityInvalidException(
+          "expo-app-intents rejected the '\(kind)' entity catalog because more than one entity has "
```
tsapeta: Maybe using three double quotes (`"""`) instead of string concatenation would look cleaner

packages/expo-app-intents/ios/AppIntentEntityStore.swift
```diff
@@ -0,0 +1,154 @@
...
+      // Every field of `AppIntentEntityRecord` is a string, so this should never happen.
+      throw AppIntentEntityInvalidException(
+        "expo-app-intents could not save the '\(kind)' entity catalog. Encoding error: "
+          + "\(error.localizedDescription)"
+      )
+    }
+  }
+}
+
+internal final class AppIntentEntityInvalidException: GenericException<String>, @unchecked Sendable {
```
tsapeta: Consider splitting this exception class into more specialized variants. Currently there is just one exception name, one code, but multiple messages.

packages/expo-app-intents/ios/ExpoAppIntents.podspec
```diff
@@ -0,0 +1,35 @@
...
+  s.description    = package['description']
+  s.license        = package['license']
+  s.author         = package['author']
+  s.homepage       = package['homepage']
+  s.platforms      = {
+    :ios => '16.4',
+    :osx => '13.4',
+    :tvos => '16.4'
+  }
+  s.swift_version  = '5.9'
```
tsapeta:
```suggestion
  s.swift_version  = '6.0'
```

tsapeta (approved):

PR #47209, opened 2026-06-24
[cli] Make prebuild clean by default

alanjhughes:
# Why
Makes `expo prebuild` clean by default.

# How
- Added the --no-clean flag to the prebuild arg spec and documented it in --help. --clean is still accepted for backwards compatibility but is now a no-op.
- Gated the the git status and native module check in prebuildAsync behind whether a native folder actually exists, so a first prebuild on a project with no native folders no longer prompts about git status.

# Test Plan
Added new tests

### DIFF
--- a/packages/@expo/cli/e2e/__tests__/prebuild-test.ts
+++ b/packages/@expo/cli/e2e/__tests__/prebuild-test.ts
@@ -53,7 +53,7 @@ it('runs `npx expo prebuild --help`', async () => {
       Options
         <dir>                                    Directory of the Expo project. Default: Current working directory
         --no-install                             Skip installing npm packages and CocoaPods
-        --no-clean                               Keep the existing native folders instead of deleting and regenerating them
+        --no-clean                               Apply changes to the existing native folders instead of recreating them
         --npm                                    Use npm to install dependencies. Default when package-lock.json exists
         --yarn                                   Use Yarn to install dependencies. Default when yarn.lock exists
         --bun                                    Use bun to install dependencies. Default when bun.lock or bun.lockb exists
--- a/packages/@expo/cli/e2e/__tests__/prebuild-test.ts
+++ b/packages/@expo/cli/e2e/__tests__/prebuild-test.ts
@@ -53,7 +53,7 @@ it('runs `npx expo prebuild --help`', async () => {
       Options
         <dir>                                    Directory of the Expo project. Default: Current working directory
         --no-install                             Skip installing npm packages and CocoaPods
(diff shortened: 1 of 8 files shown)


packages/@expo/cli/e2e/__tests__/prebuild-test.ts
```diff
@@ -53,7 +53,7 @@ it('runs `npx expo prebuild --help`', async () => {
       Options
         <dir>                                    Directory of the Expo project. Default: Current working directory
         --no-install                             Skip installing npm packages and CocoaPods
-        --clean                                  Delete the native folders and regenerate them before applying changes
+        --no-clean                               Keep the existing native folders instead of deleting and regenerating them
```
kitten:
```suggestion
        --no-clean                               Apply changes to the existing native folders instead of recreating them
```

This isn't necessarily the precise phrasing I'd like to see here. I think it'd be good to imply that this isn't about "keeping" but about "applying", since we'd like to ensure that people are aware this means the config-plugins will modify existing files (the deletion is secondary to why this will be considered less preferable, basically)

packages/@expo/cli/src/prebuild/resolveOptions.ts
```diff
@@ -37,6 +37,16 @@ export function resolvePackageManagerOptions(args: any) {
   return managers;
 }
 
+/** Resolves whether prebuild should clear and regenerate the native folders. Defaults to `true`. */
+export function resolveCleanOption(args: any): boolean {
+  if (args['--clean']) {
+    Log.warn(
+      'The --clean flag is deprecated. Prebuild now clears and regenerates the native folders by default. Pass --no-clean to keep the existing folders instead.'
```
kitten:
```suggestion
      'The --clean flag is deprecated and now enabled by default.'
```

I think we can skip the explanation here, sine we can assume that if it's passed the user had an intent and knows what is supposed to happen.

Side-note: We could also not warn, since it's not relevant to the outcome, but I could be convinced of either approaches 🤷

kitten (approved):
mainly just nits on messages

packages/@expo/cli/src/prebuild/resolveOptions.ts
alanjhughes: I don't feel strongly about the warning. I'm fine with removing. I wasn't sure how you like to handle deprecations in the cli so went with what we typically do in sdk packages.

alanjhughes: I'll remove. The function is trivial so we can just inline `!args['--no-clean']`

PR #47210, opened 2026-06-24
test(module-scripts): Narrow transforms to speed up tests

kitten:
# Why

The recent Jest changes (#47034) should make Node / non-module/RN tests faster via an SWC transform. However, we still apply the Babel transforms pretty aggressively and have not necessarily kept the excludes up-to-date to ensure that we don't over-transform with Babel.

Since `babel-jest` is a large and slow part of running Jest in CI, tightening this should speed up the test runs in CI.

# How

- Add ESM-only SWC transform for ESM-only packages (for now `@react-navigation` and `standard-navigation`)
- Add overridden selectively include list for applying transforms to `node_modules` (mostly needed for react-native packages and `@expo-google-fonts`), but we merge with `jest-expo`'s list

# Test Plan

- [Check CI run of `check-packages` and compare to prior runs](https://github.com/expo/expo/pull/47210#issuecomment-4789345066)

### DIFF
--- a/packages/expo-module-scripts/createJestPreset.cjs
+++ b/packages/expo-module-scripts/createJestPreset.cjs
@@ -1,5 +1,33 @@
 'use strict';
 
+const { createModulesTransform } = require('./jest-modules-transform.cjs');
+
+// NOTE: Many modules ship as ESM-only but are otherwise ready to be used directly
+// We can manually list them here to extend support to them and only transpile to CJS with SWC
+const cjsTransformIncludeNames = [
+  '@react-navigation',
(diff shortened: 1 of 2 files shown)


kitten:
`sdk.yml`'s "Check changed packages" (which should include all relevant packages due to the change in `expo-module-scripts` cascading via `--affected`) went from 10min to 7m50s compared to this PR's run: #47170 (https://github.com/expo/expo/actions/runs/28016704337/job/82922696246)

alanjhughes (approved):

PR #47212, opened 2026-06-24
[expo-image][android] Prevent Canvas "trying to draw too large bitmap" crash

roitium:
# Why

When loading very large images (e.g., a 7200×7200, ~200mb image), the app crashed with:

> RuntimeException: Canvas: trying to draw too large(207360000bytes) bitmap.

`ContentFitDownsampleStrategy` only considers the view dimensions for cover/contain modes — if the view size is unknown  , it falls back to 1f (no downscaling), decoding the full image. The `SafeDownsampleStrategy` already exists to enforce the hardware bitmap limit (~100 MB), but it was only wired into Fill/None contentFit modes. Additionally, `loadAsync` (used by `useImage`) decodes at full resolution with `NoopDownsampleStrategy`, bypassing downsampling entirely.

# How

- Added `decodeFormat` parameter to `ContentFitDownsampleStrategy` and internally delegated to `SafeDownsampleStrategy` as a secondary limit, taking `min(viewScale, safeScale)` in both branches (SIZE_ORIGINAL and normal).
- Updated `ExpoImageViewWrapper.createDownsampleStrategy` to pass `decodeFormat` through.
- In `ImageLoadTask`, when no explicit `maxWidth`/`maxHeight` is set, replaced `NoopDownsampleStrategy` with `SafeDownsampleStrategy` to protect the `useImage` / `loadAsync` preload path.

# Test Plan

I rebuilt my app with this new code and tried to load the same image; it displayed successfully. 

# Checklist


roitium:
Bump. Still interested in getting this reviewed. 👀

packages/expo-image/android/src/main/java/expo/modules/image/CustomDownsampleStrategy.kt
```diff
@@ -95,8 +101,8 @@ class ContentFitDownsampleStrategy(
       requestedHeight.toFloat()
     )
 
-    // We don't want to upscale the image
-    return min(1f, aspectRation)
+    // We don't want to upscale the image or Canvas crashes on large images
+    return min(min(1f, aspectRation), safeScale)
```
tsapeta, line 105: [🏗️ **design**] This now folds in `safeScale` to keep the decoded bitmap under the hardware limit, but `getSampleSizeRounding` (below, unchanged) still returns `SampleSizeRounding.QUALITY`, while `SafeDownsampleStrategy` intentionally returns `MEMORY`. The rounding mode isn't cosmetic: Glide converts this float scale factor into a power-of-two sample size, and QUALITY rounds toward a *larger* bitmap while MEMORY rounds toward a *smaller* one. So for a source whose safe scale lands between two sample-size steps, the QUALITY path can decode a bitmap larger than the safe cap `SafeDownsampleStrategy` was designed to guarantee, and the Canvas crash can still occur near the boundary. Consider returning `MEMORY` rounding when the safe cap is the binding constraint (i.e. when `safeScale < min(1f, aspectRation)`), so the guarantee that `SafeDownsampleStrategy` provides isn't lost when its scale factor is routed through this strategy.

packages/expo-image/android/src/main/java/expo/modules/image/ImageLoadTask.kt
```diff
@@ -1,8 +1,10 @@
 package expo.modules.image
 
 import android.graphics.drawable.BitmapDrawable
+import com.bumptech.glide.load.resource.bitmap.DownsampleStrategy
```
tsapeta: [🐧 nit] This `DownsampleStrategy` import is unused — the `.downsample(...)` call passes `SafeDownsampleStrategy`/`NoopDownsampleStrategy` directly, and both live in this same package, so the type name is never referenced. ktlint/detekt's `no-unused-imports` will fail in CI on this. Remove it.

packages/expo-image/android/src/main/java/expo/modules/image/ImageLoadTask.kt
```diff
@@ -29,6 +31,14 @@ open class ImageLoadTask(
           .asDrawable()
           .load(model)
           .centerInside()
+          .downsample(
+            if (options.maxWidth <= 0 && options.maxHeight <= 0) {
+              // ARGB_8888 is the default format, but is it safe to hardcode it here?
```
tsapeta: [💡 suggestion] The comment `// ARGB_8888 is the default format, but is it safe to hardcode it here?` is an open question left in shipped code. It reads as a real concern, and it is worth resolving: `ImageLoadTask` (the `loadAsync`/`useImage` path) has no user-configured `decodeFormat`, so hardcoding the default `ARGB_8888` is a safe upper bound — RGB_565 is half the bytes, so a cap computed for ARGB_8888 is conservative for it and never under-downsamples. Replace the question with a statement of that rationale, e.g. `// loadAsync has no configurable decode format; ARGB_8888 is the widest, so its size cap is a safe upper bound.`

roitium:
Bump👀

roitium:
Can someone review this pr?

packages/expo-image/android/src/main/java/expo/modules/image/ImageLoadTask.kt
```diff
@@ -29,6 +30,14 @@ open class ImageLoadTask(
...
           .load(model)
           .centerInside()
+          .downsample(
+            if (options.maxWidth <= 0 && options.maxHeight <= 0) {
+              // loadAsync has no configurable decode format; ARGB_8888 is the widest, so its size cap is a safe upper bound.
+              SafeDownsampleStrategy(DecodeFormat.ARGB_8888)
+            } else {
+              NoopDownsampleStrategy
+            }
+          )
```
intergalacticspacehighway:
```suggestion
            // loadAsync has no configurable decode format; ARGB_8888 is the widest, so its size cap is a safe upper bound.
            .customize(`when` = options.maxWidth <= 0 && options.maxHeight <= 0) {
              downsample(SafeDownsampleStrategy(DecodeFormat.ARGB_8888))
            }
```

or else it will replace the downsample strategy added by `centerIndide`

intergalacticspacehighway (changes requested):

tsapeta:
@expo-bot review

intergalacticspacehighway (approved):

roitium:
I think these ci failures weren't my fault?

lukmccall (approved):

PR #47219, opened 2026-06-24
[fingerprint] Extend SourceSkips.ExpoConfigVersions to ios.version/android.version

TLenahan:
# Why

When versioning iOS and Android independently via `ios.version` / `android.version` (which take precedence over the top-level `version`), a bump on one platform currently invalidates the fingerprint on the other.

The existing `SourceSkips.ExpoConfigVersions` flag is named for "versions in app.json" and already strips the top-level `version`, `ios.buildNumber`, and `android.versionCode`. The platform-specific `version` overrides predate the flag and were never added to it — this PR closes that gap so the flag's coverage matches its name.

# How

Extended `normalizeExpoConfig` in `@expo/fingerprint` so that the existing `ExpoConfigVersions` branch also deletes `ios.version` and `android.version` from the cloned ExpoConfig before hashing. Updated the flag's docstring to enumerate the new fields, and updated the existing unit test to seed and assert all five version-related fields are stripped.

I initially considered adding a separate `ExpoConfigPlatformVersions` flag (visible in the first commit on this branch) but folding into the existing flag is cleaner: a static `fingerprint.config.js` can't branch per-platform, so the only useful surface for either flag is "strip all of them together" — which is what `ExpoConfigVersions` is already shaped to express.

# Test Plan

- Updated the existing `should not container version when SourceSkips.ExpoConfigVersions` test in `Expo-test.ts` to also seed `ios.version` / `android.version` and assert both are stripped from the hashed ExpoConfig contents.
- I was unable to run `et check-packages @expo/fingerprint` locally (no bootstrapped `node_modules` in this worktree); CI will validate build, type-check, lint, and the new assertions.

# Note on breaking change

Projects already opting into `SourceSkips.ExpoConfigVersions` and setting `ios.version` / `android.version` will see their fingerprint hash shift after upgrading, because those fields are now stripped before hashing. Called out under 🛠 Breaking changes in the CHANGELOG entry.

# Checklist

- [ ] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: updated a module plugin). _Not applicable — change is internal to `@expo/fingerprint`._

### DIFF
--- a/packages/@expo/fingerprint/CHANGELOG.md
+++ b/packages/@expo/fingerprint/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### 🎉 New features
 
+- Added `SourceSkips.ExpoConfigPlatformVersions` to exclude `ios.version` and `android.version` from the fingerprint, so per-platform version bumps do not invalidate the fingerprint.
+
 ### 🐛 Bug fixes
 
 - Fixed unstable fingerprint for iOS precompiled modules. ([#46466](https://github.com/expo/expo/pull/46466) by [@kudo](https://github.com/kudo))
--- a/packages/@expo/fingerprint/CHANGELOG.md
+++ b/packages/@expo/fingerprint/CHANGELOG.md
@@ -4,9 +4,9 @@
 
 ### 🛠 Breaking changes
 
-### 🎉 New features
+- Extended `SourceSkips.ExpoConfigVersions` to also strip the platform-specific version overrides `ios.version` and `android.version` (which take precedence over the top-level `version`). Projects already setting this flag will see their fingerprint hash change after upgrading.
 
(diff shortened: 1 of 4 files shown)


Kudo:
hi there! thanks for sharing the pr. what do you think to reuse the `SourceSkips.ExpoConfigVersions`?

TLenahan:
@Kudo I can change this to reuse if you'd like. I wasn't sure if it should be added or reused

Kudo:
> @Kudo I can change this to reuse if you'd like. I wasn't sure if it should be added or reused

yes, please help to reuse the flag. it feels correct to reuse it because it's "ExpoConfigVersions"

Kudo (approved):
thanks for having the fix!

TLenahan:
@Kudo is it possible to get this published?

PR #47221, opened 2026-06-24
proposal: Add pre-commit hook running `oxfmt`

kitten:
# Why

**Proposal:** We know we'd like to format all files with `oxfmt`. With eslint+prettier this is usually a slow process, but with `oxfmt`, this can be quite fast, and shouldn't be disruptive. If we don't include lints in this pre-commit hook it should be relatively uncontroversial and not very disruptive.

This can still be skipped with `--no-verify`, as usual, and formatting on pre-commit is a pretty common pattern. Assuming that `oxfmt` basically almost never fails/exits with an error, this shouldn't be all that disruptive and just ensure that formatting is unlikely to fail in PRs. It specifically should reduce the work agents would have to do, and ensure we won't have to add a note for agents for this.

# How

- Add a `scripts/prepare.js` hook that sets up the `pre-commit` hook
- Add direnv tweak to enable `core.untrackedCache` and `core.fsmonitor` for Git

The direnv tweak is to speed up the `lint-staged` run which invokes a few git commands. Git in general is slowing down already on our repo, since the history and the blobs it tracks are quite large. These two options should keep this within reason. We're activating them in direnv, since they're not always safe on certain filesystems. For expo internally, they should be fine.

The `.git/hooks/pre-commit` script is written in `preprepare` via `pnpm` in a script, since it's not guaranteed that external contributors will have direnv set up.

# Test Plan

- Manually checked against commits
- Validated that editing nested files invokes the formatter correctly

# Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
     "lint": "turbo lint",
     "format": "turbo format",
     "test": "turbo test",
+    "preprepare": "node scripts/prepare.js",
     "prepare": "turbo build"
   },
   "workspaces": {
@@ -24,6 +25,9 @@
       "packages/*",
       "packages/@expo/*"
     ]
+  },
+  "lint-staged": {
+    "*.{ts,tsx,js,jsx}": "oxfmt --no-error-on-unmatched-pattern --write"
   },
   "resolutions": {
     "react-native": "0.86.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -25,6 +26,9 @@
       "packages/@expo/*"
     ]
   },
+  "lint-staged": {
+    "*.{ts,tsx,js,jsx}": "oxfmt --no-error-on-unmatched-pattern --write"
```
hassankhan, line 30: Oxfmt currently handles a few more filetypes, OTOH `.json` and `.md`

kitten: I think I was conservative on those because it'd affect `package.json` and `docs/**` respectively

hassankhan (approved):

gabrieldonadel (approved):

PR #47230, opened 2026-06-24
[expo-video] override getAllocator(PlayerId) for media3 1.8.0 compatibility

### LINKED ISSUE #47043: react-native-vision-camera forced Media3 to 1.8.0 → expo-video's VideoPlayerLoadControl was missing getAllocator(PlayerId) which became required in 1.8.0
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `expo-video@3.0.16` for the project I'm working on.

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/expo-video/android/src/main/java/expo/modules/video/player/VideoPlayerLoadControl.kt b/node_modules/expo-video/android/src/main/java/expo/modules/video/player/VideoPlayerLoadControl.kt
index 4ffe72b..8b1cf1d 100644
--- a/node_modules/expo-video/android/src/main/java/expo/modules/video/player/VideoPlayerLoadControl.kt
+++ b/node_modules/expo-video/android/src/main/java/expo/modules/video/player/VideoPlayerLoadControl.kt
@@ -54,4 +54,9 @@ class VideoPlayerLoadControl : DefaultLoadControl() {
 
     updateAllocator()
   }
+
+  // Fix: media3 1.8.0 made getAllocator(PlayerId) abstract in LoadControl
+  override fun getAllocator(playerId: androidx.media3.exoplayer.analytics.PlayerId): androidx.media3.exoplayer.upstream.Allocator {
+    return super.getAllocator()
+  }
 }
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

tsushanth:
Starting from media3 1.8.0, `LoadControl` gains an overloaded `getAllocator(PlayerId)` method that ExoPlayer calls instead of the no-arg `getAllocator()`. Any subclass that only overrides the no-arg form causes an `AbstractMethodError` at runtime because the new overload remains abstract.

`VideoPlayerLoadControl` overrides `getAllocator()` but not `getAllocator(PlayerId)`, so apps built against media3 1.8.0+ crash on playback initialisation. This commit adds the missing override, delegating to `super.getAllocator()` to preserve the existing allocator selection logic.

Fixes #47043

### DIFF
--- a/packages/expo-video/android/src/main/java/expo/modules/video/player/VideoPlayerLoadControl.kt
+++ b/packages/expo-video/android/src/main/java/expo/modules/video/player/VideoPlayerLoadControl.kt
@@ -54,4 +54,8 @@ class VideoPlayerLoadControl : DefaultLoadControl() {
 
     updateAllocator()
   }
+
+  override fun getAllocator(playerId: androidx.media3.exoplayer.analytics.PlayerId): androidx.media3.exoplayer.upstream.Allocator {
+    return super.getAllocator()
+  }
 }


behenate:
What sdk version were you using for testing? Since SDK-56 we are using `1.9.0` version of `media3` so this should not be necessary

tsushanth:
Thanks @behenate — that's a good catch. The PR was originally tested against SDK 55 which bundles media3 1.8.0. If SDK 56 bumped to 1.9.0, where the `getAllocator` overload is available under the new signature, this workaround would indeed be unnecessary.

Could you confirm which version of media3 SDK 56 ships with? If 1.9.0 already handles this correctly, I'm happy to close the PR as moot.

tsushanth:
After looking into this more, I agree with @behenate — since SDK 56 ships media3 1.9.0 which already includes the correct `getAllocator(PlayerId)` API, this workaround is not needed for the current target. I'll close the PR. Thanks for the review!

PR #47233, opened 2026-06-24
[expo-module-scripts] Build packages via turbo run build in prepublishOnly

krystofwoldrich:
The canary publish job fails ([example run](https://github.com/expo/expo/actions/runs/28059346330/job/83069576469)) with:

```
[tsl] ERROR in packages/expo-doctor/src/checks/AutolinkingDependencyDuplicatesCheck.ts(5,8)
  TS2307: Cannot find module 'expo-modules-autolinking/exports' or its corresponding type declarations.
```

- Packages now build with the SWC-based `expo-build` and set `noEmit: true` in their `tsconfig.json` (tsc is only used for type-checking).
- But `expo-module prepublishOnly` still ran `expo-module-clean` + the tsc-based `expo-module-build`. With `noEmit: true`, that `tsc` invocation emits **nothing** — so `prepublishOnly` cleaned `build/` and then left it empty.

# How

- `prepublishOnly` scripts run `turbo run build` except (`pnpm run clean && pnpm run build:prod`):
	- packages/create-expo/package.json
	- packages/create-expo-module/package.json
	- packages/expo-doctor/package.json
	- packages/expo-env-info/package.json
	- packages/install-expo-modules/package.json
	- packages/pod-install/package.json
	- packages/uri-scheme/package.json

### DIFF
--- a/packages/@expo/config-plugins/package.json
+++ b/packages/@expo/config-plugins/package.json
@@ -83,7 +83,7 @@
     "clean": "expo-module clean",
     "lint": "oxlint --config oxlint.config.mjs .",
     "format": "expo-module format",
-    "prepublishOnly": "pnpm run clean && pnpm run build",
+    "prepublishOnly": "pnpm run clean && turbo run build",
     "test": "jest",
     "depscheck": "expo-module depscheck",
     "typecheck": "tsc -p tsconfig.json"
--- a/packages/@expo/config-plugins/package.json
+++ b/packages/@expo/config-plugins/package.json
@@ -83,7 +83,7 @@
     "clean": "expo-module clean",
     "lint": "oxlint --config oxlint.config.mjs .",
     "format": "expo-module format",
-    "prepublishOnly": "pnpm run clean && turbo run build",
+    "prepublishOnly": "expo-module prepublishOnly",
     "test": "jest",
     "depscheck": "expo-module depscheck",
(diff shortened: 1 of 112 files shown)


kitten:
Ah, nice catch. To give the full context here, the publishing workflow likely will also need to be updated. I've left that for now, since that's work that's blocked on prebuild/xcframework potentially (the trade-offs are tricky). We may have to revisit the canary publishing in the meantime.

However, it seems beneficial to me to use `turbo` for this too, as in, ideally we'd have `prepublishOnly` somehow just check whether `turbo` previously ran the build, as required. That way we can keep the speed benefits on publishing, rather than running the `build` task uncached individually.

Haven't fully looked into that yet, but would be neat to somehow have `prepublishOnly` (somehow) enforce that the appropriate turbo tasks already ran

krystofwoldrich:
Publish passes with empty `prepublishOnly`.

https://github.com/expo/expo/actions/runs/28127951026

krystofwoldrich:
Thanks @kitten for the comment, I was thinking the same, that's why I've just made it empty now, because rebuilding without cache seems super wasteful.

krystofwoldrich:
Last test publish: https://github.com/expo/expo/actions/runs/28242635215

krystofwoldrich:
Bumping the versions invalidates the cache, closing for now.

PR #47242, opened 2026-06-25
[expo-type-information] Add suport for generating multiple modules at once in module-interface command

HubertBer:
# Why
For now the type information CLI supported emitting types for only a single native module, however in many packages (for example expo-contacts) there are multiple modules defined and the default to only print the first provided module leads to bad workflow.

# How
Updated `module-interface` command to generate module files for all modules defined in provided files. Now if there are multiple native modules a new `Common.types.ts` file is defined, where all records, enums and missing types are defined.

# Test Plan
- [x] Added another module to tests, updated snapshots
- [x] Checked manually for `expo-contacts`

# Checklist


HubertBer:
@expo-bot review

PR #47248, opened 2026-06-25
[cli] add agent detection to telemetry

davidmokos:
Adds `agent-cli-detector` to Expo CLI telemetry so events recorded from known coding agents include lightweight agent context. The CLI uses `agent-cli-detector@0.1.2`, which exposes a CommonJS entry point, so detection stays synchronous and does not participate in telemetry initialization.

When a known agent is detected, telemetry event context includes:

```ts
context.agent = {
  id: string,
  sessionId: string | undefined,
}
```

When no known agent is detected, `context.agent` is omitted to keep RudderStack and BigQuery payloads sparse. CLI version is already available in telemetry, so analysis can distinguish newer CLI versions with no detected agent from older CLI versions without this integration.

Tests cover detected and undetected telemetry payloads. Locally verified with `format`, `lint`, `depscheck`, and the focused telemetry Jest tests.

### DIFF
--- a/packages/@expo/cli/src/utils/telemetry/utils/agent.ts
+++ b/packages/@expo/cli/src/utils/telemetry/utils/agent.ts
@@ -1,0 +1,31 @@
+import { detectAgent } from 'agent-cli-detector';
+
+const debug = require('debug')('expo:telemetry:agent') as typeof console.log;
+
+export type AgentTelemetryContext = {
+  id: string;
+  sessionId: string | undefined;
+};
+
+let agentTelemetryContext: AgentTelemetryContext | null | undefined;
+
+export function getAgentTelemetryContext(): AgentTelemetryContext | null {
+  if (agentTelemetryContext === undefined) {
+    agentTelemetryContext = resolveAgentTelemetryContext();
+  }
+
+  return agentTelemetryContext;
+}
+
+function resolveAgentTelemetryContext(): AgentTelemetryContext | null {
+  try {
+    const { agent, detected } = detectAgent();
+    if (!detected || agent == null) {
+      return null;
+    }
+    return { id: agent.id, sessionId: agent.sessionId };
+  } catch (error: any) {
+    debug('Failed to detect coding agent: %s', error?.message ?? error);
+    return null;
+  }
+}
--- a/packages/@expo/cli/package.json
+++ b/packages/@expo/cli/package.json
@@ -78,6 +78,7 @@
     "@expo/xcpretty": "^4.4.4",
     "@react-native/dev-middleware": "0.86.0",
     "accepts": "^1.3.8",
+    "agent-cli-detector": "^0.1.2",
     "arg": "^5.0.2",
     "bplist-creator": "0.1.0",
     "bplist-parser": "^0.3.1",
--- a/packages/@expo/cli/src/utils/telemetry/Telemetry.ts
+++ b/packages/@expo/cli/src/utils/telemetry/Telemetry.ts
@@ -5,6 +5,7 @@ import { env } from '../env';
 import { FetchClient } from './clients/FetchClient';
 import { FetchDetachedClient } from './clients/FetchDetachedClient';
 import type { TelemetryClient, TelemetryClientStrategy, TelemetryRecord } from './types';
+import { getAgentTelemetryContext } from './utils/agent';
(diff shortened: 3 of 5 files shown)


packages/@expo/cli/src/utils/telemetry/utils/agent.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export function getAgentTelemetryContext(): AgentTelemetryContext {
+  agentTelemetryContext ??= resolveAgentTelemetryContext();
+  return agentTelemetryContext;
+}
+
+function resolveAgentTelemetryContext(): AgentTelemetryContext {
+  try {
+    const { agent } = detectAgent();
+    return { id: agent?.id, sessionId: agent?.sessionId };
```
Kudo: what do we want to analyze for the sessionId? not sure if people seeing we are collecting the session id would complain. if that's not useful for us, maybe remove sessionId first

Kudo:
```suggestion
    const { agent, detected } = detectAgent();
    if (!detected || agent == null) {
      return null;
    }
    return { id: agent?.id, sessionId: agent?.sessionId };
```

maybe save rudderstack/bigquery payload if we cannot detect the agent.

in the current implemention, we send `{ agent: { id: undefined, sessioniD: undefined }}`. maybe only send it when agent data is not null. 

maybe change
```
let agentTelemetryContext: AgentTelemetryContext | null | undefined = undefined;
```

- undefined: initial undefined state
- null: not by any known agents
- non-null: by agents

packages/@expo/cli/src/utils/telemetry/index.ts
```diff
@@ -24,8 +24,8 @@ export function getTelemetry(): Telemetry | null {
 
     // Initialize the telemetry
     getUserAsync()
-      .then((actor) => telemetry?.initialize({ userId: actor?.id ?? null }))
-      .catch(() => telemetry?.initialize({ userId: null }));
+      .catch(() => null)
+      .then((actor) => telemetry?.initialize({ userId: actor?.id ?? null }));
```
Kudo: not sure what's the intention for the change, especially remove the catch handler?

Kudo (changes requested):
main concern for not sending out `{ agent: { id: undefined, sessionId: undefined }}` to waste traffic/disk size

packages/@expo/cli/src/utils/telemetry/index.ts
davidmokos: thanks for catching that - it was added by mistake. I wanted to put the agent data in the initialize block before and then decided to remove it.

packages/@expo/cli/src/utils/telemetry/utils/agent.ts
davidmokos: if we have the sessionId of the current agent we can group together the commands invoked by the agent in a single run. That way we can see what commands the agent runs and what it struggles with

davidmokos: when storing in bigquery the data is flatten into:
`context_agent_id`
`context_agent_session_id`

so I think when it's undefined it's fine. But will add your suggestion as it's cleaner

davidmokos: by itself the sessionId is useless and doesn't give us any sensitive data. It's exposed via env vars by the coding agents

Kudo: good to know and that makes sense. let's keep the sessionId!

Kudo: good call! as it's flattened and won't increase cost. that's fine as-is.

Kudo (approved):
thanks for explaining. it's all good now except the change:
```
    getUserAsync()
      .catch(() => null)
      .then((actor) => telemetry?.initialize({ userId: actor?.id ?? null }));
```

krystofwoldrich (approved):
Looks good! Looking forward for the data.

krystofwoldrich:
To add this to the current SDK 57, we need to pick the commit to `sdk-57` branch.

Maybe we should also add this to SDK 56 and/or older to get reasonable coverage?

PR #47253, opened 2026-06-25
Add packageManager property in root package.json

zoontek:
# Why

Add `"packageManager": "pnpm@10.33.0"` to the root package.json, as `pnpm` is able to pick the correct version automatically (avoid conflict with v11, when installed)

Note that the package.json has been formatted by `oxfmt`, because [`sortPackageJson`](https://oxc.rs/docs/guide/usage/formatter/sorting.html#sort-package-json-fields) is `true` (default)

# How

- Add `"packageManager": "pnpm@10.33.0"` to the root package.json

# Test Plan

- Run `pnpm install`, even if `pnpm` v11 is installed. It will use the specified version.

# Checklist

### DIFF
--- a/.github/workflows/android-instrumentation-tests.yml
+++ b/.github/workflows/android-instrumentation-tests.yml
@@ -61,8 +61,6 @@ jobs:
         uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
       - name: 🔨 Install pnpm
         uses: pnpm/action-setup@0ebf47130e4866e96fce0953f49152a61190b271 # v6.0.9
-        with:
-          version: 10.33.0
       - name: ⬢ Setup Node
         uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
--- a/.github/workflows/android-instrumentation-tests.yml
+++ b/.github/workflows/android-instrumentation-tests.yml
@@ -60,9 +60,9 @@ jobs:
       - name: 👀 Check out repository
         uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
       - name: 🔨 Install pnpm
-        uses: pnpm/action-setup@0e279bb959325dab635dd2c09392533439d90093 # v6.0.8
+        uses: pnpm/action-setup@0ebf47130e4866e96fce0953f49152a61190b271 # v6.0.9
         with:
-          version: 10
+          version: 10.33.0
       - name: ⬢ Setup Node
         uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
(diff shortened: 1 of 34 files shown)


kitten:
We're currently explicitly avoiding this. @byCedric has more elaborate explanations of why this is problematic. On a side-note, this does also conflict with the workflows `setup-pnpm` action. Alternatively, I did add `package.json:engines.pnpm` with a minimum version in a recent PR

zoontek:
@kitten Unfortunately setting `engines` does not prevent this:



I understand that it's always possible to uninstall `pnpm` and install the required version, but this is (IMHO) not great DX. Also, it's already set in `docs`.

kitten:
That's a tooling problem though, and the argument here is circular. The field works only if `corepack` is used and installed, and the field itself is specific to `corepack`. As such, arguing to add it argues for `corepack` while the `engines` protection is just a protection. This is basically a local tooling problem/question (which can also itself be solved with corepack btw, which doesn't necessarily require the field and can be used for version switching)

On its own, the problems here imho outweigh the benefit. This causes tooling conflicts because `packageManager` was both overly restrictive and underspecified. It's since been misused outside of specification (e.g. even by Bun), and is overly locked down due to it pointing to a pinned version (e.g. by `setup-pnpm` and other CI actions). It's also since been removed as a spec and from Node.js, and also retracted without a replacement/interim specification.

I don't mind rolling out a discussion on Slack internally, but there's no consensus here, and from my perspective, the problems outweigh the benefits.

Unrelated, from the perspective of `pnpm@11` in general, we should start a separate workstream to upgrade.

byCedric:
I would strongly discourage the use of Corepack anywhere in an open-source project. It basically forces everyone to use Corepack, while some (like me) absolutely don't want to use it.

The main reason is that it gets in people's way most of the time. While having a different minor or patch version of `pnpm` shouldn't be problematic, Corepack turns it into a nice little absolute blocking issue. Without, pnpm can say if a feature is used that's only available in a later version.

It also has implications for CI, as now Corepack also needs to be downloaded and installed everywhere where expo/expo is executed, e.g. GitHub Actions _and_ EAS. (It's not part of Node code - [link](https://www.reddit.com/r/node/comments/1jk3vhn/what_are_the_ramifications_of_corepack_being/))

TL;DR; if there is absolutely no other way, ok. But, ideally not.

zoontek:
@kitten @byCedric No it's not specific to `corepack` (I'm a `corepack`  hater myself 😬). This is a feature of `pnpm` since v10, and it's [enabled by default](https://pnpm.io/10.x/settings#managepackagemanagerversions). Check the [release changelog](https://github.com/pnpm/pnpm/releases/tag/v10.0.0):
> `pnpm` now manages its own version based on the `packageManager` field in `package.json` by default.

This means someone installing pnpm via homebrew now (so, v11) can `cd` to the expo repo, run `pnpm install` and `pnpm` is able to download `v10.33.0` and use it automatically instead, without `corepack` even installed on the computer.

So for me, the good practice would be to use this feature, and just fix the Github action conflict (we only need to remove the `v10` version number in the workflow - it will use the version specified instead)

PR #47256, opened 2026-06-25
chore(precompile) added support for precompile/pod install without VFS overlay

chrfalch:
## Summary

In RN 0.87 we've released a new version of our precompiled binaries that works without the VFS overlay:

- https://github.com/react/react-native/pull/57285 - removal of VFS overlay
- https://github.com/react/react-native/pull/57305 - moved resource bundles to xcframework

Giving us completely self-contained XCFrameworks from RN which is a prerequisite for SwiftPM to not become too complex.

Expo needs to detect wether we have a VFS overlay or not.

## How

This PR addds support for detecting wether we're running with an RN version that uses these new frameworks or not - making sure we fallback to the old version when running with pre 0.87.

In addition it updates the pod install pipeline with the same tests.

NOTE: When merging 0.87 into Expo (as our current React Native), we can remove a lot of code in a follow-up PR that removes the VFS support all-over.

There are a couple of 3rd party packages not compatible with RN 0.87 yet, so I tested with some patches for these packages. In addition a small binary fix was added to fix a missing method in 0.87 in EXpo's code.

## Test-plan

- ✅ Built Bare-Expo, minimal-tester and precompiled with the new binaries from React native
- ✅ Verified that the same versions are using VFS and precompiling correctly when using RN 0.86

### DIFF
--- a/packages/expo-modules-autolinking/scripts/ios/precompiled_modules.rb
+++ b/packages/expo-modules-autolinking/scripts/ios/precompiled_modules.rb
@@ -989,11 +989,22 @@ module Expo
           update_xcconfig_header_search_paths(xcconfig_path, paths_string)
         end
 
-        # Also update the main project targets' build settings directly
+        # Also update the main project targets' build settings directly.
+        # NOTE: Xcodeproj build settings can be either a String or an Array — React Native's
+        # post_install sets HEADER_SEARCH_PATHS as an Array (e.g. ["$(inherited)",
+        # "\"${PODS_CONFIGURATION_BUILD_DIR}/ReactCodegen/...\""]) on every pod that depends on
+        # ReactCodegen. Interpolating an Array into a String serializes its inspect form
+        # (`["$(inherited)", ...]`), which clang receives as a single bogus `-I[...` argument that
+        # swallows all inherited header search paths for the target ("ReactCodegen/... file not
+        # found" under use_frameworks). Append per-type instead.
+        quoted_paths = header_search_paths.map { |p| "\"#{p}\"" }
         installer.pods_project.targets.each do |target|
           target.build_configurations.each do |config|
             existing = config.build_settings['HEADER_SEARCH_PATHS'] || '$(inherited)'
-            unless existing.include?(paths_string)
+            if existing.is_a?(Array)
+              missing = quoted_paths.reject { |path| existing.include?(path) }
+              config.build_settings['HEADER_SEARCH_PATHS'] = existing + missing if missing.any?
+            elsif !existing.include?(paths_string)
               config.build_settings['HEADER_SEARCH_PATHS'] = "#{existing} #{paths_string}"
             end
           end
--- a/packages/expo-modules-autolinking/scripts/ios/precompiled_modules.rb
+++ b/packages/expo-modules-autolinking/scripts/ios/precompiled_modules.rb
@@ -847,6 +847,23 @@ module Expo
       # build time when switching Debug↔Release configurations.
       def configure_use_frameworks(installer)
         return unless prebuilt_react_active?
+
+        # Modern modular prebuilt React (React.xcframework + flattened ReactNativeHeaders, no VFS
+        # overlay): React Native's rncore.rb wires the header resolution — <React/...> and
+        # `@import React` via the framework, and the relocated namespaces (<react/...>, <yoga/...>,
+        # RCTDeprecation/, ...) via a clang module map at React-Core-prebuilt/Headers/module.modulemap.
+        # But RN's pod-target loop only covers the React-* pods; it misses every Expo pod (Expo,
+        # ExpoModulesCore, ExpoImage, ...). Re-apply that module-map coverage to the pods RN missed so
+        # their <React/...> explicit-module precompile resolves the lowercase namespaces modularly
+        # instead of tripping -Wnon-modular-include-in-framework-module. (The coverage self-skips if the
+        # prebuilt did not actually ship the module map — see ensure_modular_react_header_flags.)
+        unless prebuilt_react_uses_vfs?(installer)
+          ensure_modular_react_header_flags(installer)
+          return
+        end
+
+        # Legacy VFS prebuilt React: the use_frameworks! compat below only applies when use_frameworks!
+        # is active.
         return if linkage(installer).nil?
 
         react_prebuilt_dir = File.join(installer.sandbox.root, 'React-Core-prebuilt')
@@ -861,6 +878,86 @@ module Expo
         inject_isystem_flags(installer, target_support_dir)
 
         Pod::UI.puts "[Expo] ".blue + "Created non-framework React modulemap for use_frameworks! compatibility"
+      end
+
+      # Central discriminator for how the prebuilt React-Core-prebuilt pod exposes its headers. Two
+      # layouts ship today; everything that branches on "are we modular?" should route through here
+      # rather than re-deriving it (the signals diverged once before — keying on the transient
+      # Headers/module.modulemap — and silently broke the build):
+      #
+      #   - Legacy VFS: a single React.xcframework whose headers live at React.xcframework/Headers;
+      #     <react/...>/<yoga/...> are made resolvable via a generated clang VFS overlay. Expo must
+      #     install use_frameworks! compatibility (non-framework modulemap + -isystem) for it.
+      #   - Modern framework-resolved: ReactNativeHeaders' headers are flattened into a top-level
+      #     React-Core-prebuilt/Headers/ directory and vended as ordinary public pod headers, while
+      #     <React/...> resolves through the framework. No VFS, no module-map flag, no Expo compat.
+      #
+      # The presence of the flattened top-level Headers/ directory is the stable signal: it is the
+      # pod's header_mappings_dir output (re-created by prepare_command on every build-time re-extract),
+      # unlike Headers/module.modulemap which the modern layout copies in transiently and then discards.
+      def prebuilt_react_uses_vfs?(installer)
+        return false unless prebuilt_react_active?
+
+        react_prebuilt_dir = File.join(installer.sandbox.root, 'React-Core-prebuilt')
+        !Dir.exist?(File.join(react_prebuilt_dir, 'Headers'))
+      end
+
+      # Mirrors React Native's rncore.rb add_prebuilt_header_search_paths for the pods RN's own loop
+      # misses (it only covers the React-* pod targets, not the Expo pods). Re-applies the flattened
+      # ReactNativeHeaders module map + header search path idempotently, skipping any xcconfig that
+      # already carries the flag (those RN already covered).
+      #
+      # Gated on the module map actually existing: the Maven prebuilt path (RCT_USE_PREBUILT_RNCORE
+      # without RCT_TESTONLY_RNCORE_TARBALL_PATH) can install a React-Core-prebuilt that does not ship
+      # Headers/module.modulemap. Pointing -fmodule-map-file at a missing file fails `ScanDependencies`
+      # ("module map file ... not found"), so when it is absent we inject nothing and warn — the
+      # faithful path is a tarball that ships ReactNativeHeaders via RCT_TESTONLY_RNCORE_TARBALL_PATH.
+      def ensure_modular_react_header_flags(installer)
+        module_map_file = File.join(installer.sandbox.root, 'React-Core-prebuilt', 'Headers', 'module.modulemap')
+        unless File.exist?(module_map_file)
+          Pod::UI.warn "[Expo] Prebuilt React-Core-prebuilt is missing Headers/module.modulemap — " \
+            "skipping Expo module-map coverage. If the build fails with a missing module map or " \
+            "-Wnon-modular-include, the installed prebuilt React does not match React Native's " \
+            "CocoaPods scripts; install a tarball that ships ReactNativeHeaders via " \
+            "RCT_TESTONLY_RNCORE_TARBALL_PATH."
+          return
+        end
+
+        module_map = '$(PODS_ROOT)/React-Core-prebuilt/Headers/module.modulemap'
+        header_path = '"$(PODS_ROOT)/React-Core-prebuilt/Headers"'
+        cflags_flag = "-fmodule-map-file=#{module_map}"
+        swift_flag = "-Xcc -fmodule-map-file=#{module_map}"
+        swift_flags = "-Xcc -Wno-incomplete-umbrella #{swift_flag}"
+
+        patched = 0
+        Dir.glob(File.join(installer.sandbox.root, 'Target Support Files', '**', '*.xcconfig')).each do |xcconfig_path|
+          content = File.read(xcconfig_path)
+          original = content.dup
+
+          # Each flag is ensured INDEPENDENTLY (not "skip the pod if the module-map flag is present").
+          # A pod can carry the -fmodule-map-file flag from its own podspec (e.g. react-native-reanimated
+          # 4.x adds it) yet still lack the flattened-headers search path — in which case the module map
+          # activates the yoga/react modules but <yoga/style/Style.h> & co. textually fail to resolve
+          # (the modules' headers live under React-Core-prebuilt/Headers/). Add whichever piece is absent.
+          content = append_xcconfig_flag(content, 'HEADER_SEARCH_PATHS', header_path) unless content.include?(header_path)
+          content = append_xcconfig_flag(content, 'OTHER_CFLAGS', cflags_flag) unless content.include?(cflags_flag)
+          content = append_xcconfig_flag(content, 'OTHER_SWIFT_FLAGS', swift_flags) unless content.include?(swift_flag)
+
+          next if content == original
+          File.write(xcconfig_path, content)
+          patched += 1
+        end
+
+        Pod::UI.puts "[Expo] ".blue + "Ensured modular React header flags on #{patched} xcconfig(s)"
+      end
+
+      # Appends `value` to an xcconfig `key` line (preserving $(inherited)), or adds the key if absent.
+      def append_xcconfig_flag(content, key, value)
+        if content =~ /^#{Regexp.escape(key)}\s*=.*$/
+          content.sub(/^(#{Regexp.escape(key)}\s*=.*)$/) { "#{$1} #{value}" }
+        else
+          (content.end_with?("\n") ? content : content + "\n") + "#{key} = $(inherited) #{value}\n"
+        end
       end
 
       # TODO(ExpoModulesJSI-xcframework): Remove this method when ExpoModulesJSI.xcframework
--- a/packages/expo-modules-autolinking/scripts/ios/precompiled_modules.rb
+++ b/packages/expo-modules-autolinking/scripts/ios/precompiled_modules.rb
@@ -185,6 +185,7 @@ module Expo
         configure_header_search_paths(installer)
         configure_codegen_for_prebuilt_modules(installer)
         stub_bundled_pod_targets(installer)
+        quote_swift_compatibility_header_search_paths(installer)
       end
 
       # Runs all precompiled module pre-install steps.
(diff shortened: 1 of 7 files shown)


chrfalch:
* **#47654** <a href="https://app.graphite.com/github/pr/expo/expo/47654?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#47651** <a href="https://app.graphite.com/github/pr/expo/expo/47651?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#47650** <a href="https://app.graphite.com/github/pr/expo/expo/47650?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#47649** <a href="https://app.graphite.com/github/pr/expo/expo/47649?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#47648** <a href="https://app.graphite.com/github/pr/expo/expo/47648?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#47647** <a href="https://app.graphite.com/github/pr/expo/expo/47647?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#47646** <a href="https://app.graphite.com/github/pr/expo/expo/47646?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#47256** <a href="https://app.graphite.com/github/pr/expo/expo/47256?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/47256?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/expo-modules-autolinking/scripts/ios/precompiled_modules.rb
```diff
@@ -863,6 +881,106 @@ def configure_use_frameworks(installer)
...
+          content = File.read(xcconfig_path)
+          original = content.dup
+
+          # Each flag is ensured INDEPENDENTLY (not "skip the pod if the module-map flag is present").
+          # A pod can carry the -fmodule-map-file flag from its own podspec (e.g. react-native-reanimated
+          # 4.x adds it) yet still lack the flattened-headers search path — in which case the module map
+          # activates the yoga/react modules but <yoga/style/Style.h> & co. textually fail to resolve
+          # (the modules' headers live under React-Core-prebuilt/Headers/). Add whichever piece is absent.
+          content = append_xcconfig_flag(content, 'HEADER_SEARCH_PATHS', header_path) unless content.include?(header_path)
+          content = append_xcconfig_flag(content, 'OTHER_CFLAGS', cflags_flag) unless content.include?(cflags_flag)
```
tsapeta: [🚨 **critical**] `OTHER_CPLUSPLUSFLAGS` is missing here, and I think this makes the modular path incomplete for exactly the pods this method exists to serve.

Xcode does not concatenate `OTHER_CFLAGS` into `OTHER_CPLUSPLUSFLAGS` — when a target sets `OTHER_CPLUSPLUSFLAGS`, that value *replaces* `OTHER_CFLAGS` for `.cpp`/`.mm`/`.cc` translation units. React Native's own VFS integration accounts for this and writes all three keys:

```ruby
# rncore.rb#add_vfs_overlay_flags
add_flag_to_map_with_inheritance(attributes, "OTHER_CFLAGS", vfs_overlay_flag)
add_flag_to_map_with_inheritance(attributes, "OTHER_CPLUSPLUSFLAGS", vfs_overlay_flag)
add_flag_to_map_with_inheritance(attributes, "OTHER_SWIFT_FLAGS", swift_vfs_overlay_flag)
```

Checking the generated pods in `apps/bare-expo/ios`, both `ExpoModulesCore.debug.xcconfig` and `Expo.debug.xcconfig` set `OTHER_CPLUSPLUSFLAGS`, and `packages/expo-modules-core/ios` has 111 `.mm`/`.cpp` sources. So on the modular path those files get the module map via neither key, and the `<react/...>` / `<yoga/...>` includes they rely on would fall back to non-modular textual resolution (or fail) — the very thing the method is meant to prevent.

Suggest adding a third `append_xcconfig_flag(content, 'OTHER_CPLUSPLUSFLAGS', cflags_flag)` with the same independent-guard treatment. Since these files are what actually consume the relocated `react/`/`yoga/` namespaces, this may be the key that matters most.

tools/src/prebuilds/SPMPackage.ts
```diff
@@ -1238,6 +1283,40 @@ function collectVfsAndHeaderMapFlags(
...
+      const artifactConfig = ARTIFACT_RELATIVE_PATHS[lowerName];
+
+      // Modern modular React: when the artifact ships ReactNativeHeaders.xcframework, the lowercase
+      // `react/`, `yoga/`, … namespaces are served by its flattened clang module map instead of a
+      // VFS overlay. Activate the module map (so the includes are modular) and add the headers dir
+      // to the search path (so they resolve). `<React/X.h>` keeps resolving via the React.framework
+      // binary target. This replaces the entire -ivfsoverlay + -I roots dance below.
+      if (lowerName === 'react' && artifactConfig?.moduleMapXcframework) {
+        const isModular =
+          reactArtifactIsModular(config.debugBasePath) ||
```
tsapeta, line 1303: [🏗️ **design**] This `||` plus the `continue` below creates a state where a dependency ends up with **no** React header flags at all.

If only one flavor's cache is modular — e.g. `debug/` was populated from an 0.87 artifact and `release/` still holds a stale 0.86 one, which happens when switching RN versions without clearing the cache — `isModular` is `true`, so we enter the branch. `pushModularFlags` then no-ops for the legacy flavor (`findModularHeadersDir` returns `null`), and the `continue` skips the VFS fallback that flavor actually needs. The result is a `Package.swift` where one configuration silently has neither `-fmodule-map-file` nor `-ivfsoverlay`, surfacing later as a confusing header-not-found rather than a clear cache-mismatch error.

Since the flags are already emitted per-flavor, the detection could be per-flavor too: decide modular-vs-VFS inside a helper per `basePath` and let each flavor take its own path. Alternatively, if a mixed cache is considered impossible, `&&` plus an explicit throw on disagreement would at least fail loudly.

tools/src/prebuilds/SPMPackage.ts
```diff
@@ -1238,6 +1283,40 @@ function collectVfsAndHeaderMapFlags(
...
+              // clang requires the joined `-fmodule-map-file=<path>` form (unlike `-ivfsoverlay`,
+              // it rejects the space-separated variant). `-I` adds the headers dir to the search
+              // path so the `<react/…>`, `<yoga/…>` includes resolve.
+              flags.push(`-fmodule-map-file=${path.join(headersDir, 'module.modulemap')}`);
+              flags.push('-I', headersDir);
+            }
+          };
+          pushModularFlags(debug, config.debugBasePath);
+          pushModularFlags(release, config.releaseBasePath);
+          continue;
```
tsapeta, line 1323: [💡 suggestion] This `continue` also skips the `config.includeDirectories` block below, which on the legacy path adds `-I .../React.xcframework/Headers` and `-I .../React.xcframework/React_Core`.

I think that's deliberate given the comment says `<React/X.h>` resolves through the framework binary target, but it's worth stating explicitly — `React_Core/` is where the umbrella header and the `React/` subdirectory live, so it's the textual-fallback root for `<React/...>`. If the modular artifact genuinely makes both redundant, a short line to that effect in the comment ("the `includeDirectories` -I roots below are also unnecessary because …") would save the next reader from re-deriving it. If it turns out one of them is still needed, this is a silent drop rather than a compile error at generation time.

packages/expo-modules-autolinking/scripts/ios/precompiled_modules.rb
```diff
@@ -863,6 +881,106 @@ def configure_use_frameworks(installer)
...
+      #     <React/...> resolves through the framework. No VFS, no module-map flag, no Expo compat.
+      #
+      # The presence of the flattened top-level Headers/ directory is the stable signal: it is the
+      # pod's header_mappings_dir output (re-created by prepare_command on every build-time re-extract),
+      # unlike Headers/module.modulemap which the modern layout copies in transiently and then discards.
+      def prebuilt_react_uses_vfs?(installer)
+        return false unless prebuilt_react_active?
+
+        react_prebuilt_dir = File.join(installer.sandbox.root, 'React-Core-prebuilt')
+        !Dir.exist?(File.join(react_prebuilt_dir, 'Headers'))
```
tsapeta, line 903: [🏗️ **design**] The doc comment above rightly warns that "the signals diverged once before … and silently broke the build", but this PR reintroduces a divergence across the two halves of the change: CocoaPods keys on `React-Core-prebuilt/Headers/` existing, while the prebuild tooling (`Dependencies.ts:255`, `SPMPackage.ts:103`) keys on `ReactNativeHeaders.xcframework` existing.

Both are defensible in isolation — the pods sandbox has the headers already flattened, the artifact cache still has the xcframework — so this isn't a bug today. But they can disagree if RN changes either layout independently, and a reader comparing the Ruby and TS sides has no way to know they're meant to describe the same fact. Cross-referencing them in both comments ("the TS-side equivalent is `reactArtifactIsModular`, keyed on the xcframework because the cache is pre-flatten") would make the coupling discoverable.

tools/src/prebuilds/Dependencies.ts
```diff
@@ -325,8 +340,11 @@ export const Dependencies = {
         spinner.succeed('Artifacts already up-to-date in local .dependencies folder');
       }
 
-      // Resolve the VFS overlay template with the correct paths
-      await resolveVFSOverlayTemplate(rnDest);
+      // Legacy artifacts only: resolve the VFS overlay template with the correct paths. Modular
+      // artifacts (ReactNativeHeaders.xcframework) ship a module map and need no overlay.
+      if (!fs.existsSync(path.join(rnDest, 'ReactNativeHeaders.xcframework'))) {
```
tsapeta, line 345: [💡 suggestion] This inlines the same `ReactNativeHeaders.xcframework` existence check that lines 254-256 do, and that `SPMPackage.ts:96` exposes as `reactArtifactIsModular`. Three copies of one predicate across two files.

Worth hoisting a single shared helper (the `SPMPackage.ts` one is already written and documented) and calling it from all three sites, so the eventual VFS-removal follow-up has one place to delete rather than three to find.

tools/src/prebuilds/SPMPackage.test.ts
```diff
@@ -192,3 +206,64 @@ describe('buildSwiftSettings ExpoModulesMacros plugin flags', () => {
     assert.equal(hasMacroPluginFlags(settings), false);
   });
 });
+
+describe('React header flags: modular module map vs legacy VFS overlay', () => {
+  const version = '1000.0.0';
+
+  function withCache(populate: (debugBase: string) => void): string {
+    const cachePath = fs.mkdtempSync(path.join(os.tmpdir(), 'spm-react-flags-'));
```
tsapeta, line 221: [🐧 nit] `withCache` creates a temp dir per test and never removes it, so each run leaks two directories under `os.tmpdir()`.

Every other temp-dir test in this directory cleans up — `VersionStamp.test.ts:19`, `TransformReactXCFramework.test.ts:26`, and `resolvePackage.test.ts:20` all pair `mkdtempSync` with `fs.rmSync(dir, { recursive: true, force: true })` in `afterEach`/`after`. Tracking the created paths and rming them in an `after` hook would match that convention.

tools/src/prebuilds/Utils.test.ts
```diff
@@ -176,4 +176,17 @@ describe('getVersionsInfoAsync — Hermes V1 polarity', () => {
     const { hermesVersion } = await getVersionsInfoAsync({});
     assert.equal(hermesVersion, classicVersion);
   });
+
+  // React Native 0.87+ collapsed the dual engines back to a single HERMES_VERSION_NAME.
+  // With V1 enabled by default, resolution must fall back to HERMES_VERSION_NAME instead
+  // of throwing because the (now absent) HERMES_V1_VERSION_NAME key can't be read.
+  it('falls back to HERMES_VERSION_NAME when there is no V1 key (single-engine RN)', async (t) => {
+    if (!(classicVersion && !v1Version)) {
```
tsapeta, line 239: [💡 suggestion] This guard makes the test a permanent skip on the currently pinned RN. `react-native@0.86.0`'s `sdks/hermes-engine/version.properties` contains both keys:

```
HERMES_VERSION_NAME=0.17.0
HERMES_V1_VERSION_NAME=250829098.0.14
```

so `v1Version` is always truthy and the `!v1Version` condition never holds — the single-engine fallback that the production change at `Utils.ts:507` introduces gets no coverage until the RN bump lands.

The fallback logic is pure and doesn't need the real RN tree to test: factoring the `useV1`/`version`/`tag` selection into a small exported function taking `(properties, tags, env)` would let you assert all four combinations directly, including the single-engine case this test is reaching for. As written, the test documents intent but won't catch a regression.

packages/expo-modules-autolinking/scripts/ios/precompiled_modules.rb
```diff
@@ -863,6 +881,106 @@ def configure_use_frameworks(installer)
...
+      #
+      # Gated on the module map actually existing: the Maven prebuilt path (RCT_USE_PREBUILT_RNCORE
+      # without RCT_TESTONLY_RNCORE_TARBALL_PATH) can install a React-Core-prebuilt that does not ship
+      # Headers/module.modulemap. Pointing -fmodule-map-file at a missing file fails `ScanDependencies`
+      # ("module map file ... not found"), so when it is absent we inject nothing and warn — the
+      # faithful path is a tarball that ships ReactNativeHeaders via RCT_TESTONLY_RNCORE_TARBALL_PATH.
+      def ensure_modular_react_header_flags(installer)
+        module_map_file = File.join(installer.sandbox.root, 'React-Core-prebuilt', 'Headers', 'module.modulemap')
+        unless File.exist?(module_map_file)
+          Pod::UI.warn "[Expo] Prebuilt React-Core-prebuilt is missing Headers/module.modulemap — " \
```
tsapeta, line 919: [💡 suggestion] Nice error message — states what was missing, why it's likely happening, and the concrete next step. One small thing: this is `Pod::UI.warn` on a path that the comment says will probably fail the build anyway (`ScanDependencies` "module map file not found", or `-Wnon-modular-include`).

Since the method returns without injecting anything, the build proceeds and the user hits a clang error far from this warning. If a missing `Headers/module.modulemap` alongside a present `Headers/` directory is genuinely an inconsistent-install state rather than a supported configuration, failing here with this same text would point directly at the cause. If it *is* a supported configuration (the Maven prebuilt path mentioned), then a warning is right and this is fine as-is — just want to confirm which it is.

packages/expo-modules-autolinking/scripts/ios/precompiled_modules.rb
chrfalch: Fixed — and thanks, this was right and the fix was bigger than one key.

Confirmed the mechanism on `apps/bare-expo` with RN 0.87.0-rc.3 (Maven-resolved prebuilt), counting pods whose xcconfig carries the module map per key:

| | `OTHER_CFLAGS` | `OTHER_CPLUSPLUSFLAGS` | `OTHER_SWIFT_FLAGS` |
|---|---|---|---|
| before | 211 | **0** | 211 |
| after | 211 | 211 | 211 |

One correction to the comment: `expo-modules-core/ios` has 21 `.mm` and 0 `.cpp`; with `common/cpp/**/*.cpp` from the podspec that is ~34 C++/ObjC++ TUs, not 111.

A guarded fourth `append_xcconfig_flag` call would have been **dead code**, which is why the change is shaped differently than suggested. The guards were whole-file `content.include?` checks, so by the time a `OTHER_CPLUSPLUSFLAGS` line ran, line 944 had already inserted `cflags_flag` somewhere in the file and the check short-circuited. I moved the already-present check *into* `append_xcconfig_flag` and scoped it to the key own line. That also fixed two things not in the report:

- the `OTHER_CFLAGS` guard false-positived when a pod podspec supplied `-Xcc -fmodule-map-file=<same path>` in `OTHER_SWIFT_FLAGS` — exactly the reanimated 4.x case the comment above it cites;
- line 945 appended `swift_flags` while guarding on `swift_flag`, so `-Xcc -Wno-incomplete-umbrella` was silently dropped for those same pods. It is now ensured independently and reaches all 476 xcconfigs.

Separately, the flag is now quoted exactly as RN writes it. The unquoted spelling never matched RN own injection (the `"` sits between `-Xcc` and `-fmodule-map-file`), so Expo was appending a duplicate to every pod — 256 of them. Now 0.

Verified with a full `xcodebuild`: BUILD SUCCEEDED, 0 errors, 0 non-modular diagnostics.

Worth noting the gap is upstream too: RN `add_prebuilt_header_search_paths` (`rncore.rb:560`) also omits `OTHER_CPLUSPLUSFLAGS`, while its VFS predecessor `add_vfs_overlay_flags` wrote all three keys. Since RN loop covers every pod target, an Expo-only fix leaves ~209 of the 211 short. Filing that separately.

tools/src/prebuilds/SPMPackage.ts
chrfalch: Addressed — and the answer is that both roots are genuinely redundant, now verified rather than assumed.

I prebuilt `expo-modules-core` and `expo-crypto` against real 0.87.0-rc.3 artifacts (`et prebuild --react-native-version 0.87.0-rc.3`, 8/8 Gen/Build/Compose/Verify). The generated `Package.swift` took the modular path (10 `-fmodule-map-file`, 0 `-ivfsoverlay`) and contains **zero** occurrences of `React.xcframework/Headers` and **zero** of `React.xcframework/React_Core` — and both flavors compose and verify. `ExpoModulesCore` is the right witness since it has a `cpp` target consuming the `react/` and `yoga/` namespaces.

So the silent drop is correct behaviour. I have recorded that in the comment, plus the thing that actually matters for the follow-up: when the VFS branch is deleted the `continue` goes with it, and that `includeDirectories` block would then *start* running for `react`, silently reintroducing both roots. The comment now says to keep them out deliberately.

tools/src/prebuilds/SPMPackage.test.ts
chrfalch: Fixed — tracked the created dirs and added an `afterEach` with `fs.rmSync`, matching the `VersionStamp.test.ts:19` convention you pointed at.

Verified empirically rather than by inspection: counted `spm-react-flags-*` entries in `TMPDIR` before and after a run — `before=0 after=0`.

Minor correction for the record: `resolvePackage.test.ts` cleans up via `try/finally` inside `withTempRepo`, not `afterEach`.

tools/src/prebuilds/Utils.test.ts
chrfalch: Fixed, and you were right that the test as written could not catch a regression — though the situation is slightly worse than described: **exactly one group is always dead, whichever way the RN pin points.** On the pinned 0.86 the single-engine fallback skips (both keys present, as you noted); on a 0.88 nightly, which ships only `HERMES_VERSION_NAME`, the three V1-polarity tests at :149/:160/:170 skip instead.

Took the suggested route: extracted `resolveHermesVersion(properties, tags, env)` as a pure exported function, so `getVersionsInfoAsync` just gathers inputs and delegates. Seven tests now cover all four polarity combinations unconditionally, plus three cases the integration tests could not reach at all — tag-preferred-over-properties with normalization, the V1 tag correctly ignored when falling back to the classic key, and the throw when neither key resolves.

Developed red-first: the tests failed with `TS2305: Module "./Utils" has no exported member "resolveHermesVersion"` before the extraction existed. Suite now 261 tests / 258 pass / 0 fail.

Also a small correction: the fallback lives at `Utils.ts:504-506`, not :507 (:507 was the `tag` line).

packages/expo-modules-autolinking/scripts/ios/precompiled_modules.rb
chrfalch: Good question, and it is now settled empirically: **the Maven prebuilt does ship the module map.**

Verified on 0.87.0-rc.3 with `RCT_USE_PREBUILT_RNCORE=1` and no `RCT_TESTONLY_RNCORE_TARBALL_PATH` — i.e. the pure Maven path — `Pods/React-Core-prebuilt/Headers/module.modulemap` was present (5629 bytes, freshly written by that install).

So the answer to "which is it" is neither of the two options I originally documented. It is not a supported-but-degraded configuration; the configurations that reach this warn are broken installs. `Pod::UI.warn` is still the right call *today* only because the VFS overlay is there to carry the build when injection is skipped.

Once the VFS support is removed there is no fallback left, so warn-and-continue would guarantee a clang failure far from its cause — exactly your point. I have left it as a warning for now with a TODO tied to the removal, where it should become a hard failure with this same text.

packages/expo-modules-autolinking/scripts/ios/precompiled_modules.rb
chrfalch: Fair, and slightly pointed given the doc comment right above it. Not fixing it, because the predicate exists *only* to choose between the modular and VFS paths — with one path there is nothing to detect, and both signals disappear along with the VFS support.

To be concrete about the timing: 0.87 already carries the VFS removal (`add_vfs_overlay_flags` gone, `add_prebuilt_header_search_paths` present), and its `scripts/cocoapods` tree is byte-identical to RN `main`. So the removal follow-up is near, not hypothetical.

Cross-referencing the two comments now would be churn deleted in the same PR that deletes the predicate. I would rather the removal be one deliberate change than leave a breadcrumb pointing at code on its way out — but if you would prefer the cross-reference in the meantime, happy to add it.

tools/src/prebuilds/Dependencies.ts
chrfalch: Agreed on the observation, but I am deliberately not hoisting — your own justification is the reason. If the point is to give the VFS-removal follow-up "one place to delete rather than three to find", hoisting now creates a helper that follow-up deletes immediately. `grep ReactNativeHeaders.xcframework` finds all three sites fine.

Confirmed the nuance you raised: `reactArtifactIsModular` is module-local (not exported), so nothing can reuse it today, and the two `Dependencies.ts` sites test different roots — :254 the cache dir, :345 the per-package `.dependencies/React-Core-prebuilt` copy. Same predicate, different roots.

All three go away with the VFS support, so I would rather delete than refactor-then-delete.

tools/src/prebuilds/SPMPackage.ts
chrfalch: Real bug, correctly diagnosed — and also deleted by the VFS removal, so not fixing it.

The whole failure mode is "one flavor takes the modular path and the `continue` robs the other of its `-ivfsoverlay`". Remove the VFS block and there is no fallback to lose; the `isModular` check and the `continue` go with it.

Two notes on reachability while it stands. The specific "0.87 debug / stale 0.86 release" framing is weaker than stated, since cache paths are version-keyed (`cachePath/react/<version>/<flavor>`) — it needs the same version string carrying different content, so a local-tarball swap or partial download, not a plain version bump. But flavors genuinely *are* downloaded independently, one per flavor, so a per-flavor mismatch is structurally possible and per-flavor detection would be the right shape if this code were staying.

For what it is worth I confirmed the healthy case works: prebuilding against 0.87.0-rc.3 emitted symmetric per-flavor flags, 9 `.when(configuration: .debug)` and 9 `.when(configuration: .release)`.

tsapeta (approved):

gabrieldonadel (approved):

PR #47257, opened 2026-06-25
[ios][audio] Fix playlist track index

### LINKED ISSUE #47247: [SDK-56] expo-audio `useAudioPlaylistStatus` doesn't update currentIndex on iOS

https://github.com/JakeYallop/expo-audio-playlist-repro

### Steps to reproduce

```jsx
import { useEffect, useRef } from "react";
import { Button, SafeAreaView, Text, View } from "react-native";
import { StatusBar } from "expo-status-bar";
import { useAudioPlaylist, useAudioPlaylistStatus } from "expo-audio";

const SOURCES = [
  { uri: "https://download.samplelib.com/mp3/sample-3s.mp3" },
  { uri: "https://download.samplelib.com/mp3/sample-5s.mp3" },
  { uri: "https://download.samplelib.com/mp3/sample-9s.mp3" },
];

  const playlist = useAudioPlaylist({
    sources: SOURCES,
    loop: "none",
    updateInterval: 10,
  });
  const status = useAudioPlaylistStatus(playlist);

  // Log every distinct native status change + every trackChanged event.
  const lastSig = useRef("");
  useEffect(() => {
    const statusSub = playlist.addListener("playlistStatusUpdate", (s) => {
      const sig = `${s.currentIndex}|${s.playing}|${s.didJustFinish}|${s.isLoaded}`;
      if (sig === lastSig.current) return;
      lastSig.current = sig;
      console.log("[status]", {
        currentIndex: s.currentIndex,
        trackCount: s.trackCount,
        playing: s.playing,
        didJustFinish: s.didJustFinish,
        isLoaded: s.isLoaded,
      });
    });
    // trackChanged fires once for the first auto-advance (0 -> 1) but never
    // again — playback stalls at track 1 and never advances to 2. That's the bug.
    const trackSub = playlist.addListener("trackChanged", (d) =>
      console.log("[trackChanged]", d),
    );
    return () => {
      statusSub.remove();
      trackSub.remove();
    };
  }, [playlist]);

  return (
    <SafeAreaView
      style={{ flex: 1, justifyContent: "center", gap: 24, padding: 24 }}
    >
      <StatusBar style="auto" />
      <View style={{ gap: 4 }}>
        <Text>currentIndex: {status.currentIndex}</Text>
        <Text>trackCount: {status.trackCount}</Text>
        <Text>playing: {String(status.playing)}</Text>
        <Text>didJustFinish: {String(status.didJustFinish)}</Text>
      </View>

      <Button
        title="Play (from track 0)"
        onPress={() => {
          playlist.skipTo(0);
          playlist.play();
        }}
      />
      <Button title="Pause" onPress={() => playlist.pause()} />
      <Button title="Manual next()" onPress={() => playlist.next()} />
    </SafeAreaView>
  );
}
```

Observe that the currentIndex increments from 0 -> 1, but never from 1 -> 2, even while playback continues to the third track successfully.

This is not an issue on web or android.

This is present in SDK 55 as well (which my real project is still on).

Related: https://github.com/expo/expo/issues/43819 (closed with as it didn't have a repro)

### Environment

```text
    System:
      OS: macOS 15.7.7
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 22.18.0 - /Users/jakeyallop/.nvm/versions/node/v22.18.0/bin/node
      Yarn: 1.22.22 - /Users/jakeyallop/Library/pnpm/yarn
      npm: 10.9.3 - /Users/jakeyallop/.nvm/versions/node/v22.18.0/bin/npm
      Watchman: 2026.05.18.00 - /opt/homebrew/bin/watchman
    Managers:
    SDKs:
      iOS SDK:
    IDEs:
      Android Studio: 2025.2 AI-252.27397.103.2522.14617522
    npmPackages:
      expo: 56.0.12 => 56.0.12
      react: 19.2.3 => 19.2.3
      react-dom: 19.2.3 => 19.2.3
      react-native-web: ~0.21.2 => 0.21.2
    npmGlobalPackages:
      eas-cli: 16.18.1
    Expo Workflow: bare
```

### Expo Doctor Diagnostics

```text
```

alanjhughes:
# Why
Closes #47247

# How
The iOS playlist kept its own currentTrackIndex and a currentItem reference, and updated them from inside the `AVPlayerItemDidPlayToEndTime` notification. That read races the queue player, which has not reliably advanced its current item at the moment the notification fires, so currentItem got pinned to the track that just finished. Every end notification after then failed the item == currentItem guard and bailed out, leaving the index stuck.

This reworks the iOS side, where the index comes from the player rather than managing it manually ourselves.

- currentTrackIndex is now derived from the identity of ref.currentItem within an ordered playerItems map, so it cannot drift out of sync with playback.
- trackChanged is emitted from the currentItem KVO observer.
- The end-of-track notification is reduced to the cases the player cannot handle on its own, replaying in single mode, wrapping to the start in all mode, and reporting didJustFinish at the end in none mode.

# Test Plan
Bare expo

### DIFF
--- a/packages/expo-audio/ios/AudioPlaylist.swift
+++ b/packages/expo-audio/ios/AudioPlaylist.swift
@@ -180,7 +180,12 @@ public class AudioPlaylist: SharedRef<AVQueuePlayer>, Playable, LockScreenPlayab
     }
 
     let wasPlaying = isPlaying
-    let resumeIndex = index <= currentTrackIndex ? currentTrackIndex + 1 : currentTrackIndex
+    let resumeIndex: Int
+    if ref.currentItem != nil {
+      resumeIndex = index <= currentTrackIndex ? currentTrackIndex + 1 : currentTrackIndex
+    } else {
+      resumeIndex = index
+    }
     sources.insert(source, at: index)
 
     previousTrackIndex = resumeIndex
@@ -214,6 +219,11 @@ public class AudioPlaylist: SharedRef<AVQueuePlayer>, Playable, LockScreenPlayab
       if wasPlaying {
         play(at: currentRate)
       }
+    } else if index < playerItems.count {
+      if let item = playerItems[index] {
+        ref.remove(item)
+      }
+      playerItems.remove(at: index)
     }
 
     updateStatus(with: ["trackCount": trackCount])
--- a/packages/expo-audio/ios/AudioPlaylist.swift
+++ b/packages/expo-audio/ios/AudioPlaylist.swift
@@ -82,13 +82,13 @@ public class AudioPlaylist: SharedRef<AVQueuePlayer>, Playable, LockScreenPlayab
     sources.count
   }
 
-  init(_ ref: AVQueuePlayer, sources: [AudioSource], interval: Double, loopMode: LoopMode) {
+  init(_ ref: AVQueuePlayer, sources: [AudioSource], items: [AVPlayerItem?], interval: Double, loopMode: LoopMode) {
     self.interval = interval
     self.loopMode = loopMode
     self.sources = sources
+    self.playerItems = items
     super.init(ref)
 
-    self.playerItems = ref.items().map { $0 as AVPlayerItem? }
     ref.actionAtItemEnd = loopMode == .single ? .pause : .advance
     setupPublisher()
     addPlaybackEndNotification()
--- a/packages/expo-audio/ios/AudioPlaylist.swift
+++ b/packages/expo-audio/ios/AudioPlaylist.swift
@@ -19,13 +19,22 @@ public class AudioPlaylist: SharedRef<AVQueuePlayer>, Playable, LockScreenPlayab
   var wasPlaying = false
 
   private(set) var sources: [AudioSource] = []
-  private(set) var currentTrackIndex: Int = 0
+
+  private var playerItems: [AVPlayerItem?] = []
+  private var previousTrackIndex = 0
+
+  var currentTrackIndex: Int {
+    guard let current = ref.currentItem,
+          let index = playerItems.firstIndex(where: { $0 === current }) else {
+      return min(max(previousTrackIndex, 0), max(sources.count - 1, 0))
+    }
+    return index
+  }
 
   func getSourceInfo() -> [AudioSource] {
     return sources
   }
 
-  private var currentItem: AVPlayerItem?
   private var timeToken: Any?
   private var cancellables = Set<AnyCancellable>()
   private var endObserver: NSObjectProtocol?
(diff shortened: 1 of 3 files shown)


packages/expo-audio/ios/AudioPlaylist.swift
```diff
@@ -150,13 +145,11 @@ public class AudioPlaylist: SharedRef<AVQueuePlayer>, Playable {
     }
 
     let wasPlaying = isPlaying
+    let resumeIndex = index <= currentTrackIndex ? currentTrackIndex + 1 : currentTrackIndex
```
behenate: When the playlist is empty, `currentTrackIndex` falls back to `0`, so `insert(source, at: 0)` sets `resumeIndex` to `1`. After inserting the source, `rebuildPlaylist(startingAt: 1)` creates no `AVPlayerItem`, leaving `trackCount == 1` but no current item to play.

packages/expo-audio/ios/AudioPlaylist.swift
```diff
@@ -254,14 +249,18 @@ public class AudioPlaylist: SharedRef<AVQueuePlayer>, Playable {
...
+    }
+    playerItems = items
 
+    ref.removeAllItems()
     for i in index..<sources.count {
-      if let item = AudioUtils.createAVPlayerItem(from: sources[i]) {
+      if let item = items[i] {
         ref.insert(item, after: nil)
       }
     }
```
behenate, line 323: This is out of scope of this PR, but creating new AVPlayerItems every time we insert or remove a single source seems inefficient. It may lead to more buffering and data usage. I think we should re-use existing items whenever possible.

packages/expo-audio/ios/AudioPlaylist.swift
```diff
@@ -169,19 +162,20 @@ public class AudioPlaylist: SharedRef<AVQueuePlayer>, Playable {
...
       if wasPlaying && !sources.isEmpty {
         play(at: currentRate)
       }
-    } else if index < currentTrackIndex {
-      currentTrackIndex -= 1
-      rebuildPlaylist(startingAt: currentTrackIndex)
+    } else if index < current {
+      let resumeIndex = current - 1
+      previousTrackIndex = resumeIndex
+      rebuildPlaylist(startingAt: resumeIndex)
```
behenate, line 218: It looks like we don't call `rebuildPlaylist` then `index > current` which will lead to `playerItems` becoming desynchronised from `sources`.

behenate (changes requested):

packages/expo-audio/ios/AudioPlaylist.swift
alanjhughes: `AVQueuePlayer` only moves forward, that's why any edit at or before the the current index has to rebuild the queue. Only the items ahead of the current that haven't been played can be reused. An item that has played to its end won't replay if it's re-inserted, so the current item and everything behind it have to be recreated regardless. I don't think it's worth the manual bookkeeping to try do this.

PR #47259, opened 2026-06-25
[ios][jsi] Preserve error code on async function promise rejections

wwdrew:
## Why

`JavaScriptPromise.reject` stringified any non-`JavaScriptError`, dropping the structured `code` carried by `JavaScriptThrowable` errors (e.g. `Exception`). As a result, async functions on the new architecture lost the `code` on the JS error, even though the synchronous throw path (`forwardingSwiftErrorsToJS`) preserved it.

## How

- Route `JavaScriptThrowable` errors through the code-preserving `JavaScriptError(from:)` initializer in `reject`, mirroring the sync path.
- This completes the async half of expo#27959, whose async coverage in expo#27960 was left commented out because there was no way to evaluate async code in tests at the time.
- Add async error-code tests (throw / reject / concurrent) to `ExceptionsTests` using the now-available `runtime.evalAsync` harness.

## Test Plan

- New async tests in `ExceptionsTests.swift` assert the `code` is preserved across the async throw, reject, and concurrent paths.

### DIFF
--- a/packages/expo-modules-jsi/apple/Sources/ExpoModulesJSI/Runtime/Values/JavaScriptPromise.swift
+++ b/packages/expo-modules-jsi/apple/Sources/ExpoModulesJSI/Runtime/Values/JavaScriptPromise.swift
@@ -102,21 +102,11 @@ public struct JavaScriptPromise: JavaScriptType, ~Copyable {
       guard let rejecter = rejectFunction.take() else {
         return
       }
-      // A `JavaScriptError` already carries the value to reject with (which may be an arbitrary JS
-      // value rather than an `Error`), so reuse it. Errors conforming to `JavaScriptThrowable`
-      // (e.g. expo-modules-core's `Exception`) carry a structured `code`, so route them through the
-      // code-preserving initializer to mirror the synchronous throw path in
-      // `forwardingSwiftErrorsToJS`; otherwise the `code` is lost and JS only sees the message.
-      // Any other native error is stringified into a generic `Error`.
-      let jsError: JavaScriptError
-      if let javaScriptError = error as? JavaScriptError {
-        jsError = javaScriptError
-      } else if let throwable = error as? JavaScriptThrowable {
-        jsError = JavaScriptError(runtime, from: throwable)
-      } else {
-        jsError = JavaScriptError(runtime, message: String(describing: error))
-      }
-      let errorValue = jsError.toValue()
+      // Convert the error to its JavaScript representation. This preserves an existing
+      // `JavaScriptError`'s wrapped value and a `JavaScriptThrowable`'s structured `code`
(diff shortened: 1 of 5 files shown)


packages/expo-modules-jsi/apple/Sources/ExpoModulesJSI/Runtime/Values/JavaScriptPromise.swift
```diff
@@ -103,9 +103,19 @@ public struct JavaScriptPromise: JavaScriptType, ~Copyable {
...
       // A `JavaScriptError` already carries the value to reject with (which may be an arbitrary JS
-      // value rather than an `Error`), so reuse it. Any other native error is stringified into a
-      // generic `Error`.
-      let jsError = error as? JavaScriptError ?? JavaScriptError(runtime, message: String(describing: error))
+      // value rather than an `Error`), so reuse it. Errors conforming to `JavaScriptThrowable`
+      // (e.g. expo-modules-core's `Exception`) carry a structured `code`, so route them through the
+      // code-preserving initializer to mirror the synchronous throw path in
+      // `forwardingSwiftErrorsToJS`; otherwise the `code` is lost and JS only sees the message.
+      // Any other native error is stringified into a generic `Error`.
+      let jsError: JavaScriptError
```
tsapeta: Please extract this `if/else if` cascade onto `JavaScriptError` and call it from `reject`. It has to be a `static func`, not an `init`, because the `JavaScriptError` branch returns the *input* unchanged to preserve its wrapped value's identity.

```swift
// JavaScriptError.swift, somewhere at the end

/// Returns the `JavaScriptError` representing an arbitrary native error. An existing
/// `JavaScriptError` is returned unchanged so its wrapped value reaches JS as-is; a
/// `JavaScriptThrowable` is routed through the code-preserving initializer above; any
/// other error is stringified into a generic `Error`.
@inlinable
public static func from(_ error: any Error, in runtime: JavaScriptRuntime) -> JavaScriptError {
  if let jsError = error as? JavaScriptError {
    return jsError
  }
  if let throwable = error as? JavaScriptThrowable {
    return JavaScriptError(runtime, from: throwable)
  }
  return JavaScriptError(runtime, message: String(describing: error))
}
```

```swift
// here in reject
let errorValue = JavaScriptError.from(error, in: runtime).toValue()
```

And since this is an addition to `JavaScriptError`, please also add tests for this particular helper.

tsapeta (changes requested):
Looks good, just one change request to make the call site cleaner 🙂

wwdrew:
Thanks @tsapeta for taking a look at this so quickly! I've updated with your changes, hope it's okay now.

tsapeta (approved):

PR #47262, opened 2026-06-25
[core][ios] Add closures for scoped ArrayBuffer access

barthap:
# Why

The Android PR #47261 added scoped `withJSBytes` / `withMutableJSBytes` APIs for safe cross-thread ArrayBuffer access. This PR brings the same capability to iOS, plus adds a `JavaScriptBackedArrayBufferStorage` mode so JS-allocated buffers are no longer eagerly copied into native memory.

# How

- Added `.javaScriptBacked` storage variant to `ArrayBufferStorage` — wraps a `JavaScriptValue` reference with byte offset/length, deferring byte access to the JS thread via `JavaScriptRuntime.execute`.
- JS-allocated ArrayBuffers and TypedArray views are now kept as JS-backed by default (no eager copy). `data` or `withUnsafeMutableBytes` lazily materializes into `.ownedNative` on first access.
- Added `withJSBytes` / `withMutableJSBytes` (sync `@available(*, noasync)` + async overloads) for scoped zero-copy access to the current backing storage.
- Wrapped `ArrayBufferStorage` in `SynchronizedArrayBufferStorage`, using `Mutex` to serialize lazy materialization. ArrayBuffer byte-access callbacks run after the storage lock is released.
- `asJavaScriptArrayBuffer` preserves JS identity when returning a full JS-backed buffer to the same runtime.

# Test Plan

- Run iOS native tests for expo-modules-core (`et native-unit-tests -p ios --packages expo-modules-core`).
- New tests: `withJSBytes` reads JS-backed buffer without detaching, reads typed array view ranges, `withMutableJSBytes` mutates original JS buffer and typed array view ranges, both APIs work for native-backed storage.
- Updated existing tests to expect `isNativeBacked == false` for JS-allocated buffers (lazy behavior change).
- verified that sync functions do not deadlock: `Function("foo") { (b: ArrayBuffer) in b.withJSBytes { ... } }`

# Checklist

### DIFF
--- a/packages/expo-modules-core/CHANGELOG.md
+++ b/packages/expo-modules-core/CHANGELOG.md
@@ -19,7 +19,7 @@
 - [Android] Support `PlatformColor` as Color. ([#47632](https://github.com/expo/expo/pull/47632) by [@jakex7](https://github.com/jakex7))
 - [Android] Added `ArrayBuffer.withJSBytes` for safe scoped access to underlying bytes from any thread. ([#47261](https://github.com/expo/expo/pull/47261) by [@barthap](https://github.com/barthap))
 - Added `getNativeRef()` to views created with `requireNativeViewManager`, which returns the underlying host component. ([#48788](https://github.com/expo/expo/pull/48788) by [@intergalacticspacehighway](https://github.com/intergalacticspacehighway))
-- [iOS] Added `ArrayBuffer.withJSBytes` for safe scoped access to underlying bytes from any thread. ([#47262](https://github.com/expo/expo/pull/47262) by [@barthap](https://github.com/barthap))
+- [iOS] Added `ArrayBuffer.withJSBytes` for safe scoped access to underlying bytes from any thread. JavaScript-allocated buffers are no longer eagerly copied: reads see the current JS bytes, and the first `data` or `withUnsafeMutableBytes` access makes a permanent native copy. ([#47262](https://github.com/expo/expo/pull/47262) by [@barthap](https://github.com/barthap))
 
 ### 🐛 Bug fixes
 
--- a/packages/expo-modules-core/CHANGELOG.md
+++ b/packages/expo-modules-core/CHANGELOG.md
@@ -19,6 +19,7 @@
 - [Android] Support `PlatformColor` as Color. ([#47632](https://github.com/expo/expo/pull/47632) by [@jakex7](https://github.com/jakex7))
 - [Android] Added `ArrayBuffer.withJSBytes` for safe scoped access to underlying bytes from any thread. ([#47261](https://github.com/expo/expo/pull/47261) by [@barthap](https://github.com/barthap))
 - Added `getNativeRef()` to views created with `requireNativeViewManager`, which returns the underlying host component. ([#48788](https://github.com/expo/expo/pull/48788) by [@intergalacticspacehighway](https://github.com/intergalacticspacehighway))
+- [iOS] Added `ArrayBuffer.withJSBytes` for safe scoped access to underlying bytes from any thread. ([#47262](https://github.com/expo/expo/pull/47262) by [@barthap](https://github.com/barthap))
 
 ### 🐛 Bug fixes
 
--- a/packages/expo-modules-core/CHANGELOG.md
+++ b/packages/expo-modules-core/CHANGELOG.md
@@ -24,7 +24,7 @@
 - [iOS] Added module lifecycle hooks that replace the `OnCreate`, `OnDestroy`, `OnStartObserving` and `OnStopObserving` DSL components: `didCreate`, `willDestroy`, `didStartListening` and `didStopListening`. ([#47542](https://github.com/expo/expo/pull/47542) by [@tsapeta](https://github.com/tsapeta))
 - [Android] Support `PlatformColor` as Color. ([#47632](https://github.com/expo/expo/pull/47632) by [@jakex7](https://github.com/jakex7))
 - [Android] Added `ArrayBuffer.withJSBytes` for safe scoped access to underlying bytes from any thread. ([#47261](https://github.com/expo/expo/pull/47261) by [@barthap](https://github.com/barthap))
-- [iOS] Added `ArrayBuffer.withJSBytes` for safe scoped access to underlying bytes from any thread. ([#47262](https://github.com/expo/expo/pull/47262) by [@barthap](https://github.com/barthap))
+- [iOS] Added `ArrayBuffer.withJSBytes` for safe scoped access to underlying bytes from any thread. JavaScript-allocated buffers are no longer eagerly copied: reads see the current JS bytes, and the first `data` or `withUnsafeMutableBytes` access makes a permanent native copy. ([#47262](https://github.com/expo/expo/pull/47262) by [@barthap](https://github.com/barthap))
 
 ### 🐛 Bug fixes
 
--- a/packages/expo-modules-core/CHANGELOG.md
+++ b/packages/expo-modules-core/CHANGELOG.md
@@ -24,6 +24,7 @@
 - [iOS] Added module lifecycle hooks that replace the `OnCreate`, `OnDestroy`, `OnStartObserving` and `OnStopObserving` DSL components: `didCreate`, `willDestroy`, `didStartListening` and `didStopListening`. ([#47542](https://github.com/expo/expo/pull/47542) by [@tsapeta](https://github.com/tsapeta))
 - [Android] Support `PlatformColor` as Color. ([#47632](https://github.com/expo/expo/pull/47632) by [@jakex7](https://github.com/jakex7))
 - [Android] Added `ArrayBuffer.withJSBytes` for safe scoped access to underlying bytes from any thread. ([#47261](https://github.com/expo/expo/pull/47261) by [@barthap](https://github.com/barthap))
+- [iOS] Added `ArrayBuffer.withJSBytes` for safe scoped access to underlying bytes from any thread. ([#47262](https://github.com/expo/expo/pull/47262) by [@barthap](https://github.com/barthap))
 
 ### 🐛 Bug fixes
 
--- a/packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBuffer.swift
+++ b/packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBuffer.swift
@@ -98,11 +98,21 @@ public final class ArrayBuffer: AnyArrayBuffer, Sendable {
   /// exposing JavaScript heap memory directly. Use `withJSBytes(_:)` when you need
   /// scoped zero-copy access to the current JavaScript backing storage.
   public func withUnsafeBytes<R>(_ body: (UnsafeRawBufferPointer) throws -> R) rethrows -> R {
-    let storage = materializedNativeStorageOrFail()
-    guard let nativePointer = storage.nativePointer else {
-      preconditionFailure("ArrayBuffer storage should have been materialized before unsafe access")
+    switch storageBox.currentStorage() {
+    case .ownedNative(let nativeStorage), .nativeBacked(let nativeStorage):
+      return try body(UnsafeRawBufferPointer(start: nativeStorage.pointer, count: nativeStorage.byteLength))
+    case .javaScriptBacked(let view):
+      do {
+        let storage = try view.makeOwnedNativeStorageCopy()
+        defer { storage.cleanup() }
+        guard let nativePointer = storage.nativePointer else {
+          preconditionFailure("ArrayBuffer storage copy should be native-backed")
+        }
+        return try body(UnsafeRawBufferPointer(start: nativePointer, count: storage.byteLength))
+      } catch {
+        materializationFailure(error)
+      }
     }
-    return try body(UnsafeRawBufferPointer(start: nativePointer, count: storage.byteLength))
   }
 
   /// Provides mutable access to native-accessible bytes.
--- a/packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBuffer.swift
+++ b/packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBuffer.swift
@@ -209,7 +209,7 @@ public final class ArrayBuffer: AnyArrayBuffer, Sendable {
   func asJavaScriptArrayBuffer(runtime: JavaScriptRuntime) -> JavaScriptArrayBuffer {
     switch storageBox.currentStorage() {
     case .javaScriptBacked(let view):
-      if runtime.isOnJavaScriptThread() || ProcessInfo.processInfo.processName == "xctest",
+      if runtime.isOnJavaScriptThread(),
         let arrayBuffer = JavaScriptActor.assumeIsolated({ view.asJavaScriptArrayBuffer(runtime: runtime) })
       {
         return arrayBuffer
--- a/packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBuffer.swift
+++ b/packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBuffer.swift
@@ -65,18 +65,17 @@ public final class ArrayBuffer: AnyArrayBuffer, Sendable {
 
   /// Creates a native-owned copy of this ArrayBuffer.
   ///
-  /// JavaScript-backed storage is permanently materialized before copying, so this instance no
-  /// longer observes JavaScript mutations and `isNativeBacked` becomes `true`. If the JavaScript
-  /// runtime is unavailable or the backing range is invalid, this nonthrowing API terminates with
-  /// a diagnostic instead of substituting an empty buffer.
+  /// JavaScript-backed storage is copied from its current bytes without materializing this
+  /// instance, so the source remains JavaScript-backed and observes later JavaScript mutations.
+  /// Native-backed storage is copied into new native storage. If the JavaScript runtime is
+  /// unavailable or the backing range is invalid, this nonthrowing API terminates with a
+  /// diagnostic instead of substituting an empty buffer.
   public func copy() -> ArrayBuffer {
-    let storage = materializedNativeStorageOrFail()
-    guard let nativeStorage = storage.nativeStorage else {
-      preconditionFailure("ArrayBuffer storage should have been materialized before copying")
+    do {
+      return ArrayBuffer(storage: try storageBox.currentStorage().makeOwnedNativeStorageCopy())
+    } catch {
+      materializationFailure(error)
     }
-    return ArrayBuffer(
-      storage: ArrayBufferStorage.makeOwnedNativeStorageCopy(
-        of: UnsafeRawPointer(nativeStorage.pointer), count: nativeStorage.byteLength))
   }
 
   /// Wraps native-backed storage in a `Data` instance without copying.
--- a/packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBuffer.swift
+++ b/packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBuffer.swift
@@ -75,19 +75,15 @@ public final class ArrayBuffer: AnyArrayBuffer, Sendable {
   /// observes or mutates the original JavaScript backing storage, and `isNativeBacked`
   /// returns `true`.
   public var data: Data {
-    return storageBox.withMutableStorage { storage in
-      storage.materializeNativeStorageIfNeeded()
-
-      guard let nativeStorage = storage.nativeStorage else {
-        preconditionFailure("ArrayBuffer storage should have been materialized before Data access")
-      }
-
-      let retained = Unmanaged.passRetained(self)
-      return Data(
-        bytesNoCopy: nativeStorage.pointer,
-        count: nativeStorage.byteLength,
-        deallocator: .custom({ _, _ in retained.release() }))
+    guard let nativeStorage = materializedNativeStorage().nativeStorage else {
+      preconditionFailure("ArrayBuffer storage should have been materialized before Data access")
     }
+
+    let retained = Unmanaged.passRetained(self)
+    return Data(
+      bytesNoCopy: nativeStorage.pointer,
+      count: nativeStorage.byteLength,
+      deallocator: .custom({ _, _ in retained.release() }))
   }
 
   /// Provides read-only access to native-accessible bytes.
@@ -96,7 +92,7 @@ public final class ArrayBuffer: AnyArrayBuffer, Sendable {
   /// exposing JavaScript heap memory directly. Use `withJSBytes(_:)` when you need
   /// scoped zero-copy access to the current JavaScript backing storage.
   public func withUnsafeBytes<R>(_ body: (UnsafeRawBufferPointer) throws -> R) rethrows -> R {
-    switch storageBox.withStorage({ $0 }) {
+    switch storageBox.currentStorage() {
     case .ownedNative(let nativeStorage), .nativeBacked(let nativeStorage):
       return try body(UnsafeRawBufferPointer(start: nativeStorage.pointer, count: nativeStorage.byteLength))
     case .javaScriptBacked(let view):
@@ -116,13 +112,11 @@ public final class ArrayBuffer: AnyArrayBuffer, Sendable {
   /// storage, not to the original JavaScript `ArrayBuffer`. Use `withMutableJSBytes(_:)` when
   /// mutations must affect the current JavaScript backing storage.
   public func withUnsafeMutableBytes<R>(_ body: (UnsafeMutableRawBufferPointer) throws -> R) rethrows -> R {
-    return try storageBox.withMutableStorage { storage in
-      storage.materializeNativeStorageIfNeeded()
-      guard let nativePointer = storage.nativePointer else {
-        preconditionFailure("ArrayBuffer storage should have been materialized before mutable access")
-      }
-      return try body(UnsafeMutableRawBufferPointer(start: nativePointer, count: storage.byteLength))
+    let storage = materializedNativeStorage()
+    guard let nativePointer = storage.nativePointer else {
+      preconditionFailure("ArrayBuffer storage should have been materialized before mutable access")
     }
+    return try body(UnsafeMutableRawBufferPointer(start: nativePointer, count: storage.byteLength))
   }
 
   // MARK: - Scoped JavaScript access
@@ -136,7 +130,7 @@ public final class ArrayBuffer: AnyArrayBuffer, Sendable {
   public func withJSBytes<R: Sendable>(
     _ body: (UnsafeRawBufferPointer) throws -> R
   ) throws -> R {
-    switch storageBox.withStorage({ $0 }) {
+    switch storageBox.currentStorage() {
     case .ownedNative(let nativeStorage), .nativeBacked(let nativeStorage):
       return try body(UnsafeRawBufferPointer(start: nativeStorage.pointer, count: nativeStorage.byteLength))
     case .javaScriptBacked(let view):
@@ -154,7 +148,7 @@ public final class ArrayBuffer: AnyArrayBuffer, Sendable {
   public func withJSBytes<R: Sendable>(
     _ body: @escaping (UnsafeRawBufferPointer) throws -> R
   ) async throws -> R {
-    switch storageBox.withStorage({ $0 }) {
+    switch storageBox.currentStorage() {
     case .ownedNative(let nativeStorage), .nativeBacked(let nativeStorage):
       return try body(UnsafeRawBufferPointer(start: nativeStorage.pointer, count: nativeStorage.byteLength))
     case .javaScriptBacked(let view):
@@ -171,7 +165,7 @@ public final class ArrayBuffer: AnyArrayBuffer, Sendable {
   public func withMutableJSBytes<R: Sendable>(
     _ body: (UnsafeMutableRawBufferPointer) throws -> R
   ) throws -> R {
-    switch storageBox.withStorage({ $0 }) {
+    switch storageBox.currentStorage() {
     case .ownedNative(let nativeStorage), .nativeBacked(let nativeStorage):
       return try body(UnsafeMutableRawBufferPointer(start: nativeStorage.pointer, count: nativeStorage.byteLength))
     case .javaScriptBacked(let view):
@@ -189,7 +183,7 @@ public final class ArrayBuffer: AnyArrayBuffer, Sendable {
   public func withMutableJSBytes<R: Sendable>(
     _ body: @escaping (UnsafeMutableRawBufferPointer) throws -> R
   ) async throws -> R {
-    switch storageBox.withStorage({ $0 }) {
+    switch storageBox.currentStorage() {
     case .ownedNative(let nativeStorage), .nativeBacked(let nativeStorage):
       return try body(UnsafeMutableRawBufferPointer(start: nativeStorage.pointer, count: nativeStorage.byteLength))
     case .javaScriptBacked(let view):
@@ -203,9 +197,11 @@ public final class ArrayBuffer: AnyArrayBuffer, Sendable {
   /// The native buffer is retained for the lifetime of the `JavaScriptArrayBuffer`.
   @usableFromInline
   func asJavaScriptArrayBuffer(runtime: JavaScriptRuntime) -> JavaScriptArrayBuffer {
-    switch storageBox.withStorage({ $0 }) {
+    switch storageBox.currentStorage() {
     case .javaScriptBacked(let view):
-      if let arrayBuffer = JavaScriptActor.assumeIsolated({ view.asJavaScriptArrayBuffer(runtime: runtime) }) {
+      if runtime.isOnJavaScriptThread() || ProcessInfo.processInfo.processName == "xctest",
+        let arrayBuffer = JavaScriptActor.assumeIsolated({ view.asJavaScriptArrayBuffer(runtime: runtime) })
+      {
         return arrayBuffer
       }
       return copy().asJavaScriptArrayBuffer(runtime: runtime)
@@ -284,8 +280,12 @@ public final class ArrayBuffer: AnyArrayBuffer, Sendable {
   // MARK: - JavaScript conversion
 
   /// Converts a JavaScript ArrayBuffer or TypedArray to Expo's safe native ArrayBuffer representation.
+  @JavaScriptActor
   @usableFromInline
-  internal static func from(value: borrowing JavaScriptValue, in runtime: borrowing JavaScriptRuntime) throws
+  internal static func from(
+    value: borrowing JavaScriptValue,
+    in runtime: borrowing JavaScriptRuntime
+  ) throws
     -> ArrayBuffer
   {
     guard value.isObject() else {
@@ -295,6 +295,7 @@ public final class ArrayBuffer: AnyArrayBuffer, Sendable {
   }
 
   /// Converts a borrowed JavaScript ArrayBuffer or TypedArray without materializing an owning JavaScriptValue.
+  @JavaScriptActor
   @usableFromInline
   internal static func from(
     unownedValue: borrowing JavaScriptUnownedValue,
@@ -306,13 +307,20 @@ public final class ArrayBuffer: AnyArrayBuffer, Sendable {
     return try ArrayBuffer.from(jsObject: unownedValue.getObject(in: runtime), in: runtime)
   }
 
-  private static func from(jsObject: consuming JavaScriptObject, in runtime: borrowing JavaScriptRuntime) throws
+  @JavaScriptActor
+  private static func from(
+    jsObject: consuming JavaScriptObject,
+    in runtime: borrowing JavaScriptRuntime
+  ) throws
     -> ArrayBuffer
   {
     if jsObject.isArrayBuffer() {
       let backingValue = jsObject.asValue()
       return ArrayBuffer.from(
-        jsArrayBuffer: jsObject.getArrayBuffer(), backingValue: backingValue, byteOffset: 0, in: runtime)
+        jsArrayBuffer: jsObject.getArrayBuffer(),
+        backingValue: backingValue,
+        byteOffset: 0,
+        in: runtime)
     }
     let jsValue = jsObject.asValue()
     guard jsValue.isTypedArray() else {
@@ -321,6 +329,7 @@ public final class ArrayBuffer: AnyArrayBuffer, Sendable {
     return try ArrayBuffer.from(typedArray: jsValue.getTypedArray(), in: runtime)
   }
 
+  @JavaScriptActor
   private static func from(typedArray: consuming JavaScriptTypedArray, in runtime: borrowing JavaScriptRuntime) throws
     -> ArrayBuffer
   {
@@ -348,6 +357,7 @@ public final class ArrayBuffer: AnyArrayBuffer, Sendable {
     )
   }
 
+  @JavaScriptActor
   private static func from(
     jsArrayBuffer: consuming JavaScriptArrayBuffer,
     backingValue: JavaScriptValue,
@@ -378,9 +388,17 @@ public final class ArrayBuffer: AnyArrayBuffer, Sendable {
   }
 
   private func makeOwnedNativeStorageCopy() -> ArrayBufferStorage {
-    return storageBox.withStorage { storage in
-      storage.makeOwnedNativeStorageCopy()
+    return storageBox.currentStorage().makeOwnedNativeStorageCopy()
+  }
+
+  private func materializedNativeStorage() -> ArrayBufferStorage {
+    let storage = storageBox.currentStorage()
+    if storage.nativeStorage != nil {
+      return storage
     }
+
+    let materializedStorage = storage.makeOwnedNativeStorageCopy()
+    return storageBox.publishMaterializedStorage(materializedStorage)
   }
 }
 
--- a/packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBuffer.swift
+++ b/packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBuffer.swift
@@ -33,10 +33,6 @@ public final class ArrayBuffer: AnyArrayBuffer, Sendable {
       .ownedNative(NativeArrayBufferStorage(pointer: data, byteLength: count, cleanup: cleanup)))
   }
 
-  convenience init(wrapping data: UnsafeMutableRawPointer, count: Int, cleanup: @escaping () -> Void) {
-    self.init(owning: data, count: count, cleanup: cleanup)
-  }
-
   init(borrowing data: UnsafeMutableRawPointer, count: Int, cleanup: @escaping () -> Void) {
     self.storageBox = SynchronizedArrayBufferStorage(
       .nativeBacked(NativeArrayBufferStorage(pointer: data, byteLength: count, cleanup: cleanup)))
@@ -54,7 +50,9 @@ public final class ArrayBuffer: AnyArrayBuffer, Sendable {
   }
 
   /// Allocates a new native ArrayBuffer of the given size.
-  @available(*, deprecated, renamed: "init(size:initializeToZero:)", message: "Use ArrayBuffer(size:initializeToZero:) instead.")
+  @available(
+    *, deprecated, renamed: "init(size:initializeToZero:)", message: "Use ArrayBuffer(size:initializeToZero:) instead."
+  )
   public static func allocate(size: Int, initializeToZero: Bool = true) -> ArrayBuffer {
     return ArrayBuffer(size: size, initializeToZero: initializeToZero)
   }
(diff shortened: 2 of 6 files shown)


barthap:
* **#48028** <a href="https://app.graphite.com/github/pr/expo/expo/48028?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#47262** <a href="https://app.graphite.com/github/pr/expo/expo/47262?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/47262?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#47261** <a href="https://app.graphite.com/github/pr/expo/expo/47261?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#47168** <a href="https://app.graphite.com/github/pr/expo/expo/47168?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#47106** <a href="https://app.graphite.com/github/pr/expo/expo/47106?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBuffer.swift
```diff
@@ -48,36 +50,151 @@ public final class ArrayBuffer: AnyArrayBuffer, Sendable {
...
+    }
   }
 
+  /// Provides mutable access to native-accessible bytes.
+  ///
+  /// If this buffer is JavaScript-backed, this method materializes the visible byte range into
+  /// native storage before calling `body`. Mutations then apply to the materialized native
+  /// storage, not to the original JavaScript `ArrayBuffer`. Use `withMutableJSBytes(_:)` when
+  /// mutations must affect the current JavaScript backing storage.
   public func withUnsafeMutableBytes<R>(_ body: (UnsafeMutableRawBufferPointer) throws -> R) rethrows -> R {
```
barthap, line 144: I considered making `withUnsafeMutableBytes` use the JS-scoped path instead of adding `withMutableJSBytes`, but I think separating them is clearer.

IMO, `withUnsafeBytes` / `withUnsafeMutableBytes` should preserve Swift-native `ContiguousBytes` semantics: they operate on native or materialized storage. For JS-backed buffers, that means copying/materializing first, so callers don’t accidentally access JS heap memory directly.
`withJSBytes` / `withMutableJSBytes` are then the explicit APIs for safe scoped access to the current JS backing storage, including mutating the original JS `ArrayBuffer`.

This also keeps `data` aligned with the Swift-native side. Since `Data` is copy-on-write, it works well as a byte export/view, but it isn’t an obvious mutation API for the `ArrayBuffer` itself. For mutation, callers can choose the behavior they want: `withUnsafeMutableBytes` to mutate native/materialized storage, or `withMutableJSBytes` to mutate the original JS backing storage.

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBufferStorage.swift
```diff
@@ -0,0 +1,233 @@
...
+    self.storage = storage
+  }
+
+  deinit {
+    storage.cleanup()
+  }
+
+  func withStorage<R>(_ body: (ArrayBufferStorage) throws -> R) rethrows -> R {
+    lock.lock()
+    defer { lock.unlock() }
```
tsapeta: Use `Mutex` instead

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBufferStorage.swift
```diff
@@ -0,0 +1,233 @@
...
+  }
+}
+
+/// Retains a JavaScript ArrayBuffer value and exposes a byte range through scoped runtime access.
+///
+/// This storage does not expose raw pointers directly. Callers must enter the JavaScript runtime
+/// to read or mutate the current backing bytes, or materialize the view into native storage first.
+final class JavaScriptBackedArrayBufferView: @unchecked Sendable {
+  let runtime: JavaScriptRuntime
+  let backingValue: JavaScriptValue
```
tsapeta: [🚨 **critical**] This retained `JavaScriptValue` is the crux of a cross-thread lifetime problem. `JavaScriptValue.pointee` is a `facebook.jsi.Value` holding a strong ref to the JS ArrayBuffer; it has no custom `deinit`, so the C++ `jsi::Value` destructor runs on whatever thread this view (and the enclosing `ArrayBuffer` / `SynchronizedArrayBufferStorage`) deallocates on. Releasing a `jsi::Value` off the JS thread is the off-thread-release UAF footgun from #47454 (and JSC is unsafe even when the runtime is still alive).

Before this PR, JS-backed inputs were eagerly copied into native memory, so an `ArrayBuffer` never retained a JS value and was safe to drop from any thread. This PR makes `.javaScriptBacked` the default, and the decoded `ArrayBuffer` is routinely handed to native code that drops it off-thread. Concrete path: `expo-sqlite`'s `AsyncFunction("runAsync") { ... bindBlobParams: [String: ArrayBuffer] ... }` (SQLiteModule.swift:199) — argument decoding runs on the JS thread and now yields `.javaScriptBacked`, but the `AsyncFunction` body runs on a background queue, so when that dictionary/closure is torn down the `jsi::Value` destructor fires off the JS thread.

Android avoids this because it retains via a JNI global ref whose destructor hops through `ThreadScope::WithClassLoader` and is thread-safe. The iOS equivalent would be to guarantee the `JavaScriptValue` is released on the JS thread — e.g. give the view a `deinit` that hands the value off to the runtime's JS thread for release (as the promise-ownership work in #47521 does via `LongLivedState`), rather than letting ARC drop it inline. At minimum this needs an explicit argument for why off-thread release is safe here.

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBufferStorage.swift
```diff
@@ -0,0 +1,233 @@
...
+    }
+  }
+
+  func makeOwnedNativeStorageCopy() -> ArrayBufferStorage {
+    do {
+      return try withUnsafeBytes { bytes in
+        return ArrayBufferStorage.makeOwnedNativeStorageCopy(of: bytes.baseAddress, count: bytes.count)
+      }
+    } catch {
+      preconditionFailure("Failed to materialize JavaScript-backed ArrayBuffer storage: \(error)")
```
tsapeta: [🏗️ **design**] `makeOwnedNativeStorageCopy()` turns a caught error into `preconditionFailure`, i.e. an unconditional crash. This is reachable from `withUnsafeBytes` / `withUnsafeMutableBytes` / `data` — all of which are how a native module reads a decoded ArrayBuffer. If the JS buffer was detached or resized to be out of bounds after decoding, `validateBounds` throws `ArrayBufferJSBytesAccessException`, and this converts that recoverable condition into a hard crash. Since the unscoped accessors are non-throwing this is awkward, but consider at least returning an empty/zero-length native storage on the bounds-error case rather than crashing the app, or documenting precisely which failures are considered unrecoverable.

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBuffer.swift
```diff
@@ -86,13 +203,21 @@ public final class ArrayBuffer: AnyArrayBuffer, Sendable {
...
-    ) { [self] in
-      // Retain `self` until the JS engine releases the ArrayBuffer. When this
-      // closure is dropped, the buffer deinits and deallocates its memory.
-      _ = self
+    switch storageBox.withStorage({ $0 }) {
+    case .javaScriptBacked(let view):
+      if let arrayBuffer = JavaScriptActor.assumeIsolated({ view.asJavaScriptArrayBuffer(runtime: runtime) }) {
+        return arrayBuffer
+      }
+      return copy().asJavaScriptArrayBuffer(runtime: runtime)
```
tsapeta, line 253: [🐧 nit] When the JS-backed view can't preserve identity (different runtime, or a partial view where `byteOffset != 0` / `byteLength != size`), this falls back to `copy().asJavaScriptArrayBuffer(...)`. `copy()` calls `makeOwnedNativeStorageCopy()` which for `.javaScriptBacked` runs `withUnsafeBytes` → `runtime.execute`. Since `asJavaScriptArrayBuffer` is already on the JS thread here, `execute` runs inline (fine), but it's worth confirming the recursive `.asJavaScriptArrayBuffer` on the freshly-copied `.ownedNative` buffer can't itself re-enter — it can't (the copy is native-backed so it hits the `ownedNative` branch), so this is just a note to keep the invariant in mind. No change strictly required.

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBufferStorage.swift
barthap: Switched this to Mutex.
A recursive lock was only needed by the earlier design that invoked a callback while holding the storage lock. That helper is gone. The current locking scopes are deliberately short: currentStorage() snapshots the enum, materialization and JavaScript access happen after releasing the lock, and publishMaterializedStorage() only serializes the final publish-or-clean-up decision.
In particular, withUnsafeMutableBytes does not hold the lock while executing the caller’s closure, so reentrant access to the same ArrayBuffer remains valid without requiring recursive locking. A non-recursive mutex now expresses the actual invariant and makes accidental lock-held callbacks easier to catch.

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBufferStorage.swift
barthap: Addressed by moving the retained JSI value behind `LongLivedState`.

`JavaScriptBackedArrayBufferView` is no longer the last owner that can release the `JavaScriptValue.Ref` on an arbitrary ARC thread. The state is registered in `runtime.longLivedObjects`, which keeps it alive until it is explicitly emptied on the JavaScript thread.

- On the JavaScript thread, deinit removes the state and releases the ref immediately.
- With an async scheduler, deinit schedules that same removal/release on the JavaScript thread.
- For schedulerless runtimes, deinit intentionally leaves the state registered. Runtime teardown performs the existing JavaScript-thread sweep and releases the ref there.

This also makes dropped scheduler work safe: a dropped cleanup closure is not the last owner of a live JSI value, since the long-lived-object collection retains the state until its teardown sweep. The regression tests cover worker-thread deinit and the schedulerless teardown sweep.

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBufferStorage.swift
barthap: This is intentional. The nonthrowing native access APIs promise access to the visible bytes of the `ArrayBuffer`; returning an empty buffer after runtime loss or an invalid captured range would report success while silently replacing the payload.

That is especially problematic because a valid empty `ArrayBuffer` is indistinguishable from a fabricated fallback. A caller could continue with corrupted input and have no way to detect it.

The recoverable API is the scoped `withJSBytes` / `withMutableJSBytes` path, which throws `ArrayBufferJSBytesAccessException` for runtime loss or an invalid captured range. The nonthrowing `copy()`, `data`, `withUnsafeBytes`, and `withUnsafeMutableBytes` APIs instead fail with an explicit diagnostic; their documentation now states that contract. This keeps the two API families semantically honest instead of silently changing bytes.

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBuffer.swift
barthap: Confirmed. This path does not recurse.

When identity cannot be preserved, `copy()` creates a fresh buffer with owned native storage. The subsequent `asJavaScriptArrayBuffer(...)` call therefore takes the native-storage branch immediately; it does not enter the JS-backed conversion path again.

The source remains JS-backed, while the fallback result is a one-copy native snapshot. Full-range identity and partial-view fallback are covered separately.

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBufferStorage.swift
```diff
@@ -0,0 +1,301 @@
...
+    guard byteOffset == 0, byteLength == arrayBuffer.size else {
+      return nil
+    }
+    return arrayBuffer
+  }
+
+  @JavaScriptActor
+  private func validateBounds(_ arrayBuffer: borrowing JavaScriptArrayBuffer) throws {
+    let size = arrayBuffer.size
+    guard byteOffset >= 0, byteLength >= 0, byteOffset <= size, byteLength <= size - byteOffset else {
```
tsapeta, line 267: [🚨 **critical**] `validateBounds` doesn't catch a detached buffer when the captured range is zero-length. A detached `ArrayBuffer` reports `size == 0`, so for `byteOffset == 0, byteLength == 0` every clause passes (`0 >= 0`, `0 >= 0`, `0 <= 0`, `0 <= 0`) and execution falls through to `arrayBuffer.data()` on line 224 / 234.

`data()` is an unguarded pass-through to `jsi::ArrayBuffer::data` (`JSIUtils.h:182-184`) with no `expo::CppError::tryCatch` wrapper. If Hermes raises `jsi::JSError` for a detached buffer, that C++ exception unwinds into Swift with no handler — `std::terminate`, not a catchable Swift error. Nothing in either package ever calls jsi's `detached()` predicate.

This is reachable from plain JS: pass a zero-length view (`new Uint8Array(buffer, 0, 0)`, or a buffer created as `new ArrayBuffer(0)`), then transfer or detach the backing buffer, then trigger the native read.

The test at `ArrayBufferTests.swift:602-604` only covers the branch that *does* throw (size 0 vs. stale length 3), and its comment notes real detachment couldn't be exercised on this Hermes runtime. Suggest adding an explicit detached check before `data()` — and treating `byteLength == 0` as "nothing to read, skip `data()` entirely", which also sidesteps the question of whether `data()` is even well-defined for an empty buffer. Note the pre-PR code guarded this by returning early for `size == 0` before any `data()` call.

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBufferStorage.swift
```diff
@@ -0,0 +1,301 @@
...
+        runtime.longLivedObjects.remove(longLivedState)
+        longLivedState.allowRelease()
+      }
+      return
+    }
+    guard runtime.supportsAsyncScheduling else {
+      // Schedulerless runtimes defer release to the JavaScript-thread teardown sweep.
+      return
+    }
+    runtime.schedule(priority: .immediate) { [weak runtime, longLivedState] in
```
tsapeta: [🏗️ **design**] This deviates from the `JavaScriptPromise` precedent in a way that can silently skip deregistration. `JavaScriptPromise.deinit` (`JavaScriptPromise.swift:56-60`) captures the *collection*, with the rationale spelled out in a comment:

```swift
// Capture the collection, not the runtime, so a wrapper's deinit can't prolong the runtime's
// lifetime by keeping it alive until the scheduled job drains.
let longLivedObjects = runtime.longLivedObjects
runtime.schedule { [longLivedState] in
  longLivedObjects.remove(longLivedState)
```

Capturing `[weak runtime]` and re-reading `runtime.longLivedObjects` inside the closure avoids that lifetime problem too, but adds a new failure mode: if the runtime wrapper deallocates between `schedule` and the task draining, the `guard let runtime else { return }` bails and `backingValue` is never released. It's bounded by the teardown sweep, so it's a leak-until-teardown rather than a permanent one — but the established pattern can't miss at all. Capturing the collection strongly (as the runtime's own teardown installer does, `JavaScriptRuntime.swift:729`) gets the same protection for free.

Separately, the `guard runtime.supportsAsyncScheduling else { return }` path above: that's false for every runtime except the one built by `init(unsafePointer:scheduler:dispatch:)`. Real RN apps take the scheduler path, so no production leak there, but `init(unsafePointer:)` is public — an embedder with a secondary/worklet runtime that drops a JS-backed ArrayBuffer off-thread leaks the retained JS value until teardown. The comment says "defer to the teardown sweep", which is accurate; it'd help to say explicitly that this is a deliberate leak-until-teardown for schedulerless runtimes rather than a no-op cleanup.

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBufferStorage.swift
```diff
@@ -0,0 +1,301 @@
...
+    }
+  }
+}
+
+/// Serializes access to mutable `ArrayBufferStorage`.
+///
+/// The storage can be materialized from `.javaScriptBacked` into `.ownedNative` during unscoped
+/// access, so the mutable enum value is isolated behind this lock-protected wrapper.
+final class SynchronizedArrayBufferStorage: @unchecked Sendable {
+  private let storage: Mutex<ArrayBufferStorage>
```
tsapeta, line 278: [🏗️ **design**] The PR description says this wraps storage in a `SynchronizedArrayBufferStorage` using **NSRecursiveLock**, but `ExpoModulesCore.Mutex` is `NSLock`-based and non-recursive (`packages/expo-modules-core/ios/Utilities/Mutex.swift:11`). Re-entering `withLock` on the same thread deadlocks.

No live bug today: every accessor in `ArrayBuffer.swift` uses `currentStorage()`, which copies the enum out and releases the lock before `body` runs, so the lock is never held across user code. But `withStorage` (line 279) *does* hold it across its closure, and it's only ever passed trivial bodies (`byteLength`, `isNativeBacked`). Since `ArrayBuffer` is public and `Sendable`, a future `withStorage` body that touches the same buffer, or a refactor that holds the lock across `body`, turns this into a hard deadlock.

The test named `withUnsafeMutableBytes allows reentrant access to the same buffer` (`ArrayBufferTests.swift:536`) doesn't guard against that — it uses `ArrayBuffer(size: 4)`, which is already `.ownedNative`, so `materializedNativeStorage()` returns at the `storage.nativeStorage != nil` early exit and the nested `byteLength`/`data` calls take the lock only after the outer call released it. It passes with a non-recursive lock, and would pass with no lock at all. Either make the lock actually recursive, or rename the test and drop the reentrancy claim from the description.

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBuffer.swift
```diff
@@ -48,36 +50,182 @@ public final class ArrayBuffer: AnyArrayBuffer, Sendable {
...
+  /// storage, not to the original JavaScript `ArrayBuffer`. Use `withMutableJSBytes(_:)` when
+  /// mutations must affect the current JavaScript backing storage. Materialization is permanent:
+  /// later access uses the same native storage and `isNativeBacked` becomes `true`. If
+  /// materialization fails, this nonthrowing API terminates with a diagnostic instead of
+  /// substituting an empty buffer.
   public func withUnsafeMutableBytes<R>(_ body: (UnsafeMutableRawBufferPointer) throws -> R) rethrows -> R {
-    return try body(UnsafeMutableRawBufferPointer(start: rawPointer, count: byteLength))
+    let storage = materializedNativeStorageOrFail()
+    guard let nativePointer = storage.nativePointer else {
+      preconditionFailure("ArrayBuffer storage should have been materialized before mutable access")
```
tsapeta, line 147: [💡 suggestion] `withUnsafeBytes`, `withUnsafeMutableBytes`, `data`, and `copy()` all terminate the process via `preconditionFailure` when materialization fails, and the doc comments justify this as preferable to "substituting an empty buffer".

Agreed that a silent empty buffer would be worse, but crashing isn't the only alternative — and two of the failure causes are triggerable from JS rather than being programmer errors: the runtime going away (`"JavaScript runtime is no longer available"`) and a detached/resized backing buffer (`validateBounds`). Pre-PR, none of these APIs could fail at all, because the bytes were already copied into native memory by the time a consumer touched them.

So this converts "JS did something unusual with a buffer it handed to a native module" into an app crash, in the same call sites that previously could not fail. `data` and `withUnsafeBytes` come from `AnyArrayBuffer`/`ContiguousBytes` so they can't be made `throws`, which is presumably why it landed this way — that's a good argument for not making `.javaScriptBacked` the default for buffers reaching those APIs, rather than for crashing when it happens.

At minimum, the error messages reaching `preconditionFailure` should say what a developer can do about it (per the repo's error-message guidance: what / why / how). `"ArrayBuffer materialization failed: <error>"` doesn't tell anyone that their JS code detached a buffer while native code held it, or that they should use `withJSBytes` instead.

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBuffer.swift
```diff
@@ -48,36 +50,182 @@ public final class ArrayBuffer: AnyArrayBuffer, Sendable {
...
+  ///
+  /// If this buffer is JavaScript-backed, every call creates a fresh full-range transient native
+  /// copy that observes the current JavaScript bytes. The copy is not published: this buffer
+  /// remains JavaScript-backed and later calls observe later JavaScript mutations. Use
+  /// `withJSBytes(_:)` when you need scoped zero-copy access to the current JavaScript backing
+  /// storage. If the transient copy cannot be made, this nonthrowing API terminates with a
+  /// diagnostic instead of substituting an empty buffer.
   public func withUnsafeBytes<R>(_ body: (UnsafeRawBufferPointer) throws -> R) rethrows -> R {
-    return try body(UnsafeRawBufferPointer(start: rawPointer, count: byteLength))
+    switch storageBox.currentStorage() {
```
tsapeta, line 117: [💡 suggestion] The read/write asymmetry for JS-backed storage is going to surprise people: `withUnsafeBytes` makes a fresh transient copy per call and therefore *observes* later JS mutations, while `withUnsafeMutableBytes` permanently materializes and *stops* observing them (and its writes never reach JS).

So for a single buffer, `read → write → read` and `write → read` return different things than `read → read`, and nothing in the API surface signals the transition. The doc comments describe each method's behavior accurately in isolation, but the interaction is only discoverable by reading both.

No current consumer mixes the two on a JS-supplied buffer, so this is latent. Still, if the lazy default ships it'd be worth stating the state machine in the type-level doc comment on `ArrayBuffer` (JS-backed → materialized-on-first-mutable-access, one-way), rather than only per-method.

packages/expo-modules-core/ios/Core/DynamicTypes/DynamicArrayBufferType.swift
```diff
@@ -19,15 +19,21 @@ internal struct DynamicArrayBufferType: AnyDynamicType {
...
   func cast(jsValue: JavaScriptValue, appContext: AppContext) throws -> Any {
     do {
-      return try ArrayBuffer.from(value: jsValue)
+      return try ArrayBuffer.from(value: jsValue, in: try appContext.runtime)
     } catch is ArrayBufferJavaScriptValueConversionException {
       throw NotArrayBufferException(innerType)
     }
   }
 
   func castToJS<ValueType>(_ value: ValueType, appContext: AppContext) throws -> JavaScriptValue {
```
tsapeta, line 28: [🐧 nit] Good catch adding the runtime-aware overload — this is live, not dead code. `Conversions.anyToJavaScriptValue(_:appContext:in:)` (`Conversions.swift:152`), `EventEmitter.swift:84`, `JSValueEncoder.swift:111`, and the per-element recursion in `DynamicArrayType`/`DynamicDictionaryType` all call the 3-arg form, which previously fell through to the default that drops `runtime` and re-reads `appContext.runtime`. Matters for the multi-runtime/worklet case, and matches how `DynamicArrayType`, `DynamicDictionaryType`, `DynamicCodableType`, `DynamicRawType`, and `DynamicStringType` already do it.

One thought: `DynamicTypedArrayType` has the same 2-arg-only shape and the same JS-identity concern. Might be worth the same treatment in a follow-up.

packages/expo-modules-core/CHANGELOG.md
```diff
@@ -23,6 +23,7 @@
 - [iOS] Added module lifecycle hooks that replace the `OnCreate`, `OnDestroy`, `OnStartObserving` and `OnStopObserving` DSL components: `didCreate`, `willDestroy`, `didStartListening` and `didStopListening`. ([#47542](https://github.com/expo/expo/pull/47542) by [@tsapeta](https://github.com/tsapeta))
 - [Android] Support `PlatformColor` as Color. ([#47632](https://github.com/expo/expo/pull/47632) by [@jakex7](https://github.com/jakex7))
 - [Android] Added `ArrayBuffer.withJSBytes` for safe scoped access to underlying bytes from any thread. ([#47261](https://github.com/expo/expo/pull/47261) by [@barthap](https://github.com/barthap))
+- [iOS] Added `ArrayBuffer.withJSBytes` for safe scoped access to underlying bytes from any thread. ([#47262](https://github.com/expo/expo/pull/47262) by [@barthap](https://github.com/barthap))
```
tsapeta: [🐧 nit] The entry covers the new closures but not the storage-default change, which is the part with observable consequences for existing code: JS-allocated ArrayBuffer arguments are no longer eagerly copied, `isNativeBacked` now returns `false` for them (the PR's own test updates confirm this), `withUnsafeBytes` can now block and can now observe later JS mutations, and `data`/`withUnsafeBytes`/`copy()` can now abort the process where they previously could not fail.

That deserves its own entry — arguably under a breaking or behavior-change heading rather than a feature bullet, since anyone relying on snapshot semantics is affected without touching their code.

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBufferStorage.swift
```diff
@@ -0,0 +1,301 @@
...
+    let backingValue = JavaScriptValue.Ref()
+
+    func allowRelease() {
+      backingValue.release()
+    }
+
+    func getArrayBuffer() throws -> JavaScriptArrayBuffer {
+      guard
+        let arrayBuffer = try backingValue.withValue({
+          value in try value?.getArrayBuffer()
```
tsapeta: **suggestion:** This reads awkwardly, and the shape is hiding two real problems rather than just being verbose.

```swift
guard
  let arrayBuffer = try backingValue.withValue({
    value in try value?.getArrayBuffer()
  })
else {
```

1. **Doubled optionality collapses two distinct failures into one message.** `withValue`'s closure returns `R?` *and* the ref yields `nil` when empty, so `withValue` itself returns `R?` too. The `guard let` can't tell "ref was released" from "closure returned nil" — both surface as `"JavaScript-backed ArrayBuffer was released"`. That's accurate today only because the closure never returns nil for a live value.

2. **Neither `try` does anything.** `getArrayBuffer()` doesn't throw — it's the non-throwing accessor whose type check is a debug-only `assert` (`JavaScriptValue.swift:303-309`), compiled out in Release, after which a non-ArrayBuffer value reaches `jsi::Object::getArrayBuffer` and hits a precondition failure / UB. So the inner `try` guards nothing, and the outer `try` exists *only* because `rethrows` sees a `try` in the closure body. Remove the inner one and the outer becomes a compile error: the throwing appearance is self-referential.

The repo convention is to prefer the throwing `as*` accessors over `get*` on decode paths for exactly the Release-assert reason. Using `asArrayBuffer()` (if available) would make the `try` genuine, let the two failure modes carry distinct messages, and keep the type check alive in Release.

Even leaving the accessor alone, this collapses to one readable line:

```swift
guard let arrayBuffer = backingValue.withValue({ $0?.getArrayBuffer() }) else {
  throw ArrayBufferJSBytesAccessException("JavaScript-backed ArrayBuffer was released")
}
return arrayBuffer
```

Same behavior, no phantom `try`, and the `value in`-on-its-own-line reflow goes away (that reflow is itself a hint the expression was doing too much for one statement).

For context, reading through `withValue` rather than `take()` is the right call and matches `JavaScriptPromise.asValue()` — it deliberately avoids consuming the long-lived ref. It's the doubled optionality plus the decorative `try` that make this instance read badly, not the use of `withValue`.

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBufferStorage.swift
barthap: Added zero-length short-circuits in both JS-thread access paths, so an empty range no longer calls `data()`. I couldn't reproduce the detached-buffer case because this test runtime has no detach API. The new test covers the empty scoped read/write contract, not detachment; the production guards avoid calling `data()` in either case.

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBufferStorage.swift
barthap: Changed the queued cleanup to capture the collection instead of the runtime, matching `JavaScriptPromise.deinit`. The collection is stored during actor-isolated initialization because this cross-module `deinit` cannot read the actor-isolated property through the runtime. Runtime-wrapper deallocation can no longer make the cleanup job return early; if teardown drops the job, the teardown sweep releases the registered state. I also clarified the schedulerless-runtime comment.

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBufferStorage.swift
barthap: The PR description was wrong: `Mutex` is `NSLock`-based, and that is intentional. `withUnsafeMutableBytes` calls its body after materialization and after releasing the storage lock. I corrected the description and renamed the test to state that contract.

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBuffer.swift
barthap: Fail-loud is intentional for storage-access failures. `copy()` and `data` satisfy nonthrowing `AnyArrayBuffer` requirements, while `withUnsafeBytes` is constrained by `ContiguousBytes`; returning empty bytes would silently replace the payload. I narrowed the catch so errors thrown by the `rethrows` body still propagate. The scoped APIs remain the recoverable path, and the diagnostic now names the likely causes and points to them.

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBuffer.swift
barthap: Added type-level documentation for the one-way transition. `withUnsafeBytes` uses a temporary copy and `copy()` returns an independent snapshot without materializing the source. The first `data` or mutable native access materializes permanent native storage, and all later source access uses that storage.

packages/expo-modules-core/ios/Core/DynamicTypes/DynamicArrayBufferType.swift
barthap: Agreed, `DynamicTypedArrayType` has the same shape. I'll handle it in a follow-up PR, it's unrelated to scoped `ArrayBuffer` access.

packages/expo-modules-core/CHANGELOG.md
barthap: Extended the entry with the lazy-storage behavior. I kept it under New features because the unified `ArrayBuffer` API and this behavior are part of the same unpublished release, so there is no released eager-copy contract to break.

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBufferStorage.swift
barthap: Simplified the optional handling and removed the decorative `try`. This private reference is initialized only from a verified ArrayBuffer and can later become empty, but cannot change type, so `getArrayBuffer()` preserves the invariant without adding a Release-mode type check to every scoped access.

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBuffer.swift
```diff
@@ -48,36 +58,183 @@ public final class ArrayBuffer: AnyArrayBuffer, Sendable {
...
 
+  /// Provides read-only access to native-accessible bytes.
+  ///
+  /// If this buffer is JavaScript-backed, every call creates a fresh full-range transient native
+  /// copy that observes the current JavaScript bytes. The copy is not published: this buffer
+  /// remains JavaScript-backed and later calls observe later JavaScript mutations. Use
+  /// `withJSBytes(_:)` when you need scoped zero-copy access to the current JavaScript backing
+  /// storage. If the transient copy cannot be made, this nonthrowing API terminates with a
+  /// diagnostic instead of substituting an empty buffer.
   public func withUnsafeBytes<R>(_ body: (UnsafeRawBufferPointer) throws -> R) rethrows -> R {
```
tsapeta, line 116: [🚨 **critical**] Making `.javaScriptBacked` the decode default turns this into a **blocking JS-thread hop**, and that deadlocks `expo-sqlite` today.

The chain: for JS-backed storage this calls `view.makeOwnedNativeStorageCopy()` → the sync `withUnsafeBytes` → `runtime.execute`, which off the JS thread does `scheduler.scheduleTask(.ImmediatePriority)` then spins `CFRunLoopRunInMode` until the JS thread drains it (`JavaScriptRuntime.swift:493-527`).

Now the consumer side. `SQLiteModule.run(statement:...)` takes `statement.lock.wait()` — a `DispatchSemaphore(value: 1)` (`NativeStatement.swift:9`) — at `SQLiteModule.swift:394` and holds it via `defer` across the whole function, including the `bindBlobParams` loop at `:407-412` → `bindStatementParam` → `param.withUnsafeBytes { exsqlite3_bind_blob(...) }` at `:643`.

So an `AsyncFunction` body on `moduleQueue` holds the semaphore while blocking on the JS thread. If the JS thread then calls `executeSync` on the *same* statement, it hits `lock.wait()` — and `DispatchSemaphore.wait()` does not pump the run loop, so the JS thread can never service the pending `scheduleTask`. Both threads stay stuck forever.

Repro, no contrived code:
```js
const stmt = await db.prepareAsync('INSERT INTO t (blob) VALUES (?)');
const blob = new Uint8Array(1024);   // JS-heap, no MutableBuffer → .javaScriptBacked
stmt.executeAsync(blob);             // not awaited → moduleQueue
stmt.executeSync(blob);              // same statement, JS thread
```
A forgotten `await`, or fire-and-forget write then sync read on a shared prepared statement.

Two scoping notes so this isn't overstated: `db.runSync` vs `db.runAsync` **cannot** hit it (each prepares a fresh `NativeStatement` with its own semaphore — `SQLiteDatabase.ts:246-248`, `:358-359`, `:461-462`), and two concurrent `runAsync` calls **cannot** either (the JS thread is free to service the hop). It needs a user-held `SQLiteStatement` with both sync and async execution, plus a blob param. Narrow, but reachable and permanent.

`tryBorrowMutableBuffer()` fails for JS-allocated `Uint8Array` (`JSIUtils.h:230`), and `paramUtils.ts:41` routes every `Uint8Array`/`ArrayBuffer` into `blobParams`, so all binary binding takes this path.

The API in this file is fine. The problem is the *default*: any existing `AsyncFunction` that reads an `ArrayBuffer` argument under a lock now inverts that lock against the JS thread, silently, with no source change. Options, roughly in order of preference:

1. Keep eager copy as the decode default and let callers opt into JS-backed storage. That preserves the pre-PR contract for `AsyncFunction` bodies and still ships the scoped API.
2. Land a paired expo-sqlite change that copies blobs before `lock.wait()` (or moves the bind loop out of the critical section), and audit for other lock-across-hop sites.

Per-blob JS round trips are also a real perf regression for bulk inserts (`db.runAsync('… VALUES (?,?,?)', b1, b2, b3)` becomes 3 blocking hops), and `FileSystemFile.swift:135` now costs a hop plus a second full copy of the payload. Those are stalls, not hangs — but they argue the same way.

Related: `sessionApplyChangeset` (`SQLiteModule.swift:790-792`) does `UnsafeMutableRawPointer(mutating: $0.baseAddress)` on the buffer from `withUnsafeBytes`. That now aliases a **transient copy**, so any in-place mutation `exsqlite3changeset_apply` performs is silently discarded. Same cast at `:809-811`. No lock held there, so no deadlock, but it is a behavior change worth checking.

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBuffer.swift
```diff
@@ -48,36 +58,183 @@ public final class ArrayBuffer: AnyArrayBuffer, Sendable {
...
+  /// byte range is invalid.
+  @available(*, noasync)
+  public func withJSBytes<R: Sendable>(
+    _ body: (UnsafeRawBufferPointer) throws -> R
+  ) throws -> R {
+    switch storageBox.currentStorage() {
+    case .ownedNative(let nativeStorage), .nativeBacked(let nativeStorage):
+      return try body(UnsafeRawBufferPointer(start: nativeStorage.pointer, count: nativeStorage.byteLength))
+    case .javaScriptBacked(let view):
+      return try withoutActuallyEscaping(body) { escapingBody in
```
tsapeta: [🏗️ **design**] This `withoutActuallyEscaping` is not sound by construction, and the same pattern repeats at the `withMutableJSBytes` sync overload.

`withoutActuallyEscaping` requires the closure not to outlive the call. Here `escapingBody` goes into the view's sync `withUnsafeBytes`, which boxes it in `NonisolatedUnsafeVar` and passes it to `runtime.execute`. `execute`'s only exit condition is `result != nil`, which is assigned *inside* the scheduled block (`JavaScriptRuntime.swift:503-517`). Nothing after that assignment is synchronized with the caller: the block finishes setting `result`, wakes the caller, and only then does the JS thread return into RN's task loop, pop the `shared_ptr<Task>`, and destroy the `RawCallback` that holds the closure. So the caller can pop the `withoutActuallyEscaping` frame while the closure context is still alive on another thread.

Whether RN frees `Task::callback` before `execute` returns isn't determinable from this repo (headers only, no `.cpp`). But there is an in-repo `ScheduleFn` that retains blocks indefinitely by design — `holdSchedulerTask` in `JavaScriptRuntimeTests.swift:1103-1116`, simulating the #47716 reload scenario. So the precondition rests on scheduler-implementation behavior the type system doesn't enforce.

In practice this is benign, because the closure is heap-boxed and is only *released* after the scope, never *invoked*. But that is a different guarantee than the one `withoutActuallyEscaping` expresses, and the laundering sits two frames above where the escape actually happens.

Suggestion: move the `withoutActuallyEscaping` down into `JavaScriptRuntime.execute`'s cross-thread branch, where the escape occurs and the lifetime is known. Failing that, document the real invariant here rather than leaving `withoutActuallyEscaping` to imply a stronger one.

Secondary: `R: Sendable` on the sync overloads constrains the wrong thing. The caller is blocked, so nobody else can observe `R` — it needn't be `Sendable`. Meanwhile the *captures* are unconstrained and do cross threads. And the caller isn't fully idle: `execute` pumps `CFRunLoopRunInMode(.commonModes, 0.1, false)` on the caller's run loop (`JavaScriptRuntime.swift:525`), so run-loop callouts can touch captured state concurrently with the body on the JS thread. Worth stating that constraint in the doc comment ("the body must not touch state the caller's run loop can reach").

packages/expo-modules-core/ios/Core/ArrayBuffers/ArrayBufferStorage.swift
```diff
@@ -0,0 +1,312 @@
...
+      return
+    }
+    runtime.schedule(priority: .immediate) { [longLivedState, longLivedObjects] in
+      longLivedObjects.remove(longLivedState)
+      longLivedState.allowRelease()
+    }
+  }
+
+  @available(*, noasync)
+  func withUnsafeBytes<R: Sendable>(
```
tsapeta, line 164: [💡 suggestion] Reply on the fail-loud contract is accurate for what it defends, but it doesn't cover the case that actually motivated the concern.

Agreed that `copy()`, `data`, and `withUnsafeBytes` can't be made `throws` (`AnyArrayBuffer` / `ContiguousBytes`), and agreed that substituting an empty buffer would be worse than terminating. The diagnostic is also much better now — it names the likely causes and points at `withJSBytes`. No argument with any of that.

What remains: pre-PR those three APIs **could not fail at all**, because the bytes were already native by the time a consumer touched them. The lazy default is what introduces the abort, and two of the triggers are reachable from ordinary JS rather than being programmer errors — runtime torn down, and a detached/transferred backing buffer.

Concretely, `applyChangesetAsync` / `invertChangesetAsync` (`SQLiteModule.swift:287`, `:294`) run on `moduleQueue` and read the buffer well after decode. Under app teardown, or if JS transfers the buffer in between, `withUnsafeBytes` now hard-terminates where it previously could not fail. Changesets are the largest buffers in that API, so the decode-to-read window is widest there.

That isn't an argument for returning empty bytes. It's an argument that the *default* shouldn't route buffers that reach non-throwing APIs into storage that can fail (same conclusion as the deadlock comment). If the default stays, this is worth calling out explicitly in the CHANGELOG as a new abort surface, not only in the doc comments.

packages/expo-modules-core/ios/Tests/ArrayBufferTests.swift
```diff
@@ -382,11 +418,282 @@ struct ArrayBufferTests {
...
+      let encoded = buffer.asJavaScriptArrayBuffer(runtime: runtime).asValue()
+
+      runtime.global().setProperty("encodedBuffer", value: encoded)
+      let bytes = try runtime.eval("Array.from(new Uint8Array(encodedBuffer))").asArray().map { try $0.asInt() }
+
+      #expect(bytes == [2, 3])
+    }
+
     @Test
     func `shares native-backed buffer when using legacy NativeArrayBuffer argument`() throws {
```
tsapeta, line 698: [💡 suggestion] These three "legacy `NativeArrayBuffer` argument" tests don't test a legacy path — there isn't one.

`NativeArrayBuffer` is a deprecated typealias for `ArrayBuffer` (`NativeArrayBuffer.swift:5-6`), so `processNativeBuffer` (`:1039`) takes the identical type and runs the identical code as the non-legacy tests. Same for `shares native-backed typed array view when using legacy NativeArrayBuffer argument` (`:720`) and `copies JS-backed typed array view when using legacy NativeArrayBuffer argument` (`:744`). The names promise coverage of a distinct path that doesn't exist, which is the kind of thing that reads as a safety net later and isn't one.

Same issue with the ~25 `NativeArrayBuffer.allocate(...)` / `.copy(...)` / `.wrap(...)` call sites the diff introduces across this file: they route through the deprecated alias *and* the deprecated `allocate` static, so the suite now compiles with deprecation warnings while testing exactly what `ArrayBuffer(size:)` tested before.

Suggest reverting the mechanical `ArrayBuffer` → `NativeArrayBuffer` rename in the tests and either dropping the three "legacy" tests or renaming them to say what they actually cover (native-backed sharing, JS-backed copy-on-mutable-access). If the intent was to pin the deprecated alias's behavior, one dedicated test asserting the alias resolves and forwards would say that far more clearly.

barthap:
## Moving this PR to draft

The deadlock is real (although unlikely). Sync functions run inline on the JS thread, so `executeSync` blocks it on the statement semaphore. The async side then waits forever for a scheduled task that can never run. RN's modern sync access (`executeNowOnTheSameThread`) doesn't help either; it also needs the JS thread to reach a yield point.

I looked at both suggested options:

**Eager copy on decode.** This removes the deadlock and keeps the pre-PR contract, but it also removes the reason this PR exists. With an eager default, nothing creates JS-backed storage from user input. `withJSBytes` never sees live JS bytes and becomes `withUnsafeBytes` with a different name. There would be no point in shipping the scoped API like that.

**Paired expo-sqlite change.** Out of scope, #47168 is already merged, and I don't want this PR to depend on rewriting consumers.

Instead, I did the same thing Android already does. Android's `JSHeapAccessExecutor.runOnQueueSync` waits 5 seconds, then cancels and throws, but only if the body has not started yet. The same lock inversion there turns into a 5s stall and a rejected promise instead of a permanent hang. This branch now does the same on iOS: off-thread sync access to JS-backed storage waits at most 5 seconds, with the same cancel-if-not-started rule. The deadlock window is narrow (a held prepared statement, a JS-heap blob, an unawaited async call and a concurrent sync one), but now it's at least visible and recoverable on both platforms.

That doesn't answer the design objection, though. Changing the decode default silently changes the contract for every existing `ArrayBuffer` consumer, and I don't want to merge that as-is. So:

- This PR moves to draft. The scoped-access API and the timeout are worth keeping for reference. The timeout is an isolated commit, so it can be cherry-picked later.
- Android keeps its merged behavior (lazy default + timeout). If the inversion happens in real apps, it will show up as `IllegalStateException: Timed out waiting to run synchronous JavaScript task`. We'll see if such reports show up.
- Once the `@JS` (v2) API settles, I want to try a different approach: keep eager copy as the safe decode default and add a per-function or per-parameter annotation that opts a specific `ArrayBuffer` argument into lazy zero-copy behavior. There is still one buffer type, and the risky behavior becomes an explicit local decision instead of a global default.

Also fixed other review nits: tests no longer use the deprecated `NativeArrayBuffer` alias, and the former "legacy" test names now say what they actually cover.

PR #47264, opened 2026-06-25
[docs] fix HorizontalFloatingToolbar example using wrong prop name

GautamKumarOffical:
Fixes #44571

The floating toolbar scrollable content example had `onClick` on `HorizontalFloatingToolbar.FloatingActionButton`, but the component actually uses `onPress`. Fixed in both unversioned and v56 docs.


amandeepmittal (approved):
Thank you for the fix!

PR #47283, opened 2026-06-26
[ios][updates] Add copyEmbeddedAssets property

alanjhughes:
# Why

https://exponent-internal.slack.com/archives/C013ZK4SA12/p1782392708325469

On iOS, expo-updates reads and hashes every embedded asset and inserts a database row for it on first launch. That work synchronously gates app startup but isn't used by the launch it blocks, the embedded update serves its JS bundle and assets directly from the app binary. The copy only pre-seeds the updates cache so a later OTA update that reuses an embedded asset key can skip a re-download.

Android already exposes this as an opt-out via `EX_UPDATES_COPY_EMBEDDED_ASSETS` #36059 iOS had no equivalent.

# How

- updates.copyEmbeddedAssets in app config
- EXUpdates.podspec reads updatesCopyEmbeddedAssets and sets the compile flag for Debug and Release.
- UpdatesUtils.shouldCopyEmbeddedAssets() exposes the macro to Swift.

# Test Plan

- New BuildProperties-test.ts cases assert the plugin writes `updatesCopyEmbeddedAssets` correctly.
- ExpoConfig test passes

PR #47284, opened 2026-06-26
[ios][updates] Skip embedded asset copying on first launch by default

alanjhughes:
# Why

This makes no-copy the default on iOS, matching Android. The embedded asset copy/hash/insert work was gating first launch without benefiting it, since embedded updates resolve their bundle and assets directly from the .app binary. Removing it from the default launch path is a large startup win that scales with bundled asset count and size.

# How

- EmbeddedAppLoader.swift takes an `shouldCopyEmbeddedAssets` boolean. In the default no-copy path it registers the embedded update row without reading or hashing the non-launch assets.
- UpdatesDatabase.swift gates markUpdateFinished so that in no-copy mode embedded updates stay StatusEmbedded rather than being promoted to StatusReady by the reaper. This keeps every launch on the bundle-resolution branch. Without it, the second launch took the non-embedded ensureAllAssetsExist path with zero registered asset rows, so the completion never fired and the JS bundle never loaded.
- Docs: clarify that asset selection controls OTA eligibility, not what's bundled into the binary, and document the updates.copyEmbeddedAssets opt-in and its first-launch cost tradeoff.

# Test Plan

- Native tests are passing
- Local e2e pass
- Benchmark bare-expo in Release. Updates.launchDuration on first launch:
    - copy ON: 303.23 ms
    - copy OFF (new default): 19.21 ms

### DIFF
--- a/docs/pages/eas-update/asset-selection.mdx
+++ b/docs/pages/eas-update/asset-selection.mdx
@@ -47,6 +47,8 @@ After adding this configuration all **.png** files in all subdirectories of **ap
 
 If `assetPatternsToBeBundled` isn't included in the app config, all assets resolved by the bundler will be included in updates (as per SDK 49 and earlier behavior).
 
+> **info** Asset selection controls which assets are eligible for over-the-air updates. It does not change which assets are bundled into the native binary, so it does not reduce app startup time.
+
 ## Verifying that an update includes all required app assets
 
 When using the asset selection, assets that do not match any file patterns will resolve in the Metro bundler. However, these assets will not be uploaded to the updates server. You have to be certain that assets not included in updates are built into the native build of the app.
--- a/apps/bare-expo/ios/Podfile.lock
+++ b/apps/bare-expo/ios/Podfile.lock
@@ -4121,7 +4121,7 @@ SPEC CHECKSUMS:
   ExpoVideoThumbnails: 116c2563d2bd3a1e98326a267020e25fea8af79e
   ExpoWebBrowser: 6e3d90e3fe1952d21a8494872b1e1a485cd50e2f
   EXStructuredHeaders: 9e89bcdd636ae2ecb59995cfba3230f5d7547c08
-  EXUpdates: c6488f0bebe92d6decb5f9b685d33ee91feeccf0
+  EXUpdates: 165c09b63ed318b21fdde13d8f0796ea09e350d9
   EXUpdatesInterface: 25408a97d682355eb9fb37e5aa6e22caece1881f
   FBLazyVector: b3e7ad108f0d882e30445c5527d774e3fd432f3d
   hermes-engine: 4c998771d5218e20701b63d8358199a520a54447
--- a/packages/expo-updates/CHANGELOG.md
+++ b/packages/expo-updates/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### 🎉 New features
 
+- [iOS] Skip reading and hashing embedded assets on first launch by default, serving them from the app binary instead of copying them into the updates cache. ([#47284](https://github.com/expo/expo/pull/47284) by [@alanjhughes](https://github.com/alanjhughes))
+
 ### 🐛 Bug fixes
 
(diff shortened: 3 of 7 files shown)


alanjhughes:
* **#47284** <a href="https://app.graphite.com/github/pr/expo/expo/47284?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/47284?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#47283** <a href="https://app.graphite.com/github/pr/expo/expo/47283?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

docs/pages/eas-update/asset-selection.mdx
```diff
@@ -47,6 +47,9 @@ After adding this configuration all **.png** files in all subdirectories of **ap
 
 If `assetPatternsToBeBundled` isn't included in the app config, all assets resolved by the bundler will be included in updates (as per SDK 49 and earlier behavior).
 
+> **info** Asset selection controls which assets are eligible for over-the-air updates. It does not change which assets are bundled into the native binary, so it does not reduce app startup time. On iOS, embedded assets are served directly from the app binary on first launch and are not copied into the updates cache by default.
+> If you want embedded assets pre-seeded into the cache so a later update that reuses them avoids a re-download, set `updates.copyEmbeddedAssets` to `true`. Note that this runs the copy while the app launches and can slow first launch for apps with many bundled assets.
```
Kudo:
```suggestion
> **info** Asset selection controls which assets are eligible for over-the-air updates. It does not change which assets are bundled into the native binary, so it does not reduce app startup time.
```

not if i misunderstood
- keep android/ios parity
- keep `copyEmbeddedAssets` as internal. not explicitly doc it

Kudo (changes requested):
there's one major difference compared to android implementation.

currently in this pr, we don't go through the downloadAsset phase. we don't save asset metadata to the updates db. if people calling `Updates.localAssets`, it would be empty.

i think that would be good to keep the flow aligned with normal updates. just skip the copying/hashing operations.

alanjhughes:
@Kudo we can't really do the same as android here as the hash and all other columns are required on iOS. They are nullable on android. We can't skip generating the hash if we want to populate the row. I'd prefer to avoid a database migration. We could stop using `assetFilesMap` as a signal we are using the embedded update and instead populate it so `localAssets` are still populated.

Kudo:
migration is fine for new sdk. if you want to avoid migration, we could use empty string for the `hash` column. what're the other non-nullable columns on ios that we cannot support?

alanjhughes:
@Kudo we need to be able to backport this to 54. It's not a new sdk feature.

I'm not sure I understand why we want the db rows inserted? What does it gain us only that it's the same as android? It makes no functional difference on iOS. Putting "fake" or bad data in there doesn't seem like the right approach. We can provide the `localAssets` without the db rows

Kudo:
> we need to be able to backport this to 54. It's not a new sdk feature.

can we use empty `hash` for backporting and add nullable hash db migration on main?

> I'm not sure I understand why we want the db rows inserted? What does it gain us only that it's the same as android? It makes no functional difference on iOS. Putting "fake" or bad data in there doesn't seem like the right approach. We can provide the localAssets without the db rows

doesn't mean to align android design. it could be me not familiar expo-updates enough and not confident to introduce a new flow. i feel expo-updates is designed upon on database data. if we skip some steps, not sure if that would break something.

don't want to block by my skeptic. i'll leave you to make the call

Kudo (approved):

alanjhughes:
As far as I can tell it should be fine. But I'll try investigate some more to see if we have any better options 👍

Kudo:
thanks Alan!

alanjhughes:
### Merge activity

* **Jun 29, 12:57 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/expo/expo/47284).
* **Jun 29, 12:59 PM UTC**: [Graphite](https://app.graphite.com/github/pr/expo/expo/47284) rebased this pull request as part of a merge.
* **Jun 29, 12:59 PM UTC**: @alanjhughes merged this pull request with [Graphite](https://app.graphite.com/github/pr/expo/expo/47284).

PR #47287, opened 2026-06-26
[expo-router] Fix `requireContextWithOverrides` ignoring overrides and duplicating route keys

wwdrew:
# Why

`renderRouter({ appDir, overrides })` silently ignores `overrides` whenever the override key matches a file that already exists in `appDir`. The override component is never rendered — the real file is loaded instead — and `keys()` returns the same route under two different key shapes, so expo-router sees a duplicate route.

The testing docs state that override keys use the same extension-free, no-leading-notation format as `inMemoryContext` (e.g. `_layout`, not `./_layout.tsx`):

> The `overrides` parameter is an inline mock that can be used to override specific paths within the `appDir`. Do not use leading relative (`./`) or absolute (`/`) notation when defining these paths and exclude file extension.
> — https://docs.expo.dev/router/reference/testing/

But `require-context-ponyfill` produces keys in the `./name.ext` form. When expo-router calls `ctx('./_layout.tsx')` to load a route, `requireContextWithOverrides` checks `if (id in overrides)` — `'./_layout.tsx' in { _layout: … }` is always `false` — so the override is skipped and the real file is loaded. `keys()` then concatenates `['_layout']` with `['./_layout.tsx', …]`, listing the same file twice.

# How

In `requireContextWithOverrides`, normalize each override key to the matching require-context id before performing the lookup: for an override key `name`, find the existing context key of the form `./name.<ext>` (falling back to `./name` when no file matches). The function lookup and `keys()` then operate on the normalized map, and `keys()` filters out any existing key replaced by an override so a route is never listed twice.

# Test Plan

Added `packages/expo-router/src/__tests__/context-stubs.test.node.ts` covering:

- an override registered with an extension-free key (`_layout`) resolves when the context is queried with the require-context id (`./_layout.tsx`);
- files without an override still fall through to the existing context;
- `keys()` lists the require-context id once and does not include the bare override key;
- an override with no matching file is kept under a normalized `./name` key.

# Checklist


packages/expo-router/src/__tests__/context-stubs.test.node.ts
```diff
@@ -0,0 +1,44 @@
...
+  it('resolves an override registered with an extension-free key to the require-context id', () => {
+    const ctx = requireContextWithOverrides(appDir, { _layout: MockLayout });
+
+    expect(ctx('./_layout.tsx').default).toBe(MockLayout);
+  });
+
+  it('falls back to the existing context for files without an override', () => {
+    const ctx = requireContextWithOverrides(appDir, { _layout: MockLayout });
+
+    expect(ctx('./index.tsx').default.name).toBe('Index');
```
Ubax: Please assert these using `import IndexFixture from "./fixtures/context-stubs/index"`

packages/expo-router/src/__tests__/context-stubs.test.node.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+describe('requireContextWithOverrides', () => {
+  it('resolves an override registered with an extension-free key to the require-context id', () => {
+    const ctx = requireContextWithOverrides(appDir, { _layout: MockLayout });
+
+    expect(ctx('./_layout.tsx').default).toBe(MockLayout);
+  });
+
+  it('falls back to the existing context for files without an override', () => {
```
Ubax, line 83: Please also add tests checking if Index can be overridden as well

And one more when there is no override

packages/expo-router/src/__tests__/context-stubs.test.node.ts
```diff
@@ -0,0 +1,44 @@
...
+
+import { requireContextWithOverrides } from '../testing-library/context-stubs';
+
+const appDir = path.join(__dirname, 'fixtures', 'context-stubs');
+
+function MockLayout() {
+  return null;
+}
+
+describe('requireContextWithOverrides', () => {
```
Ubax, line 67: Please also add a nested route fxiture and test with and without override

packages/expo-router/src/testing-library/context-stubs.ts
```diff
@@ -46,18 +46,38 @@ export function inMemoryContext(context: MemoryContext) {
 
 export function requireContextWithOverrides(dir: string, overrides: MemoryContext) {
   const existingContext = requireContext(path.resolve(process.cwd(), dir));
+  const existingKeys = existingContext.keys();
```
Ubax: Please do this the other way around:
1. Create a function - normalize keys which will map files with extensions to normalized keys. In this function when you find duplicates e.g. both `.jsx` and `.tsx` for the same file, then throw an error 
2. Use this function to normalize the `existingContext`


Add unit tests for the new function, and adjust existing ones for the new implementation

Ubax (changes requested):
Thank you for opening the PR. I'd like you to do a small change. Otherwise looks good 👍

wwdrew:
Hi @Ubax, thank you so much for the review. I've updated the PR to address your comments, hope it's ok now.

packages/expo-router/src/testing-library/context-stubs.ts
```diff
@@ -44,20 +44,70 @@ export function inMemoryContext(context: MemoryContext) {
...
+    const existing = normalized.get(normalizedKey);
+    if (existing !== undefined) {
+      throw new Error(
+        `Multiple files resolve to the same route "${normalizedKey}": "${existing}" and "${key}". Remove or rename one of these files.`
+      );
+    }
+    normalized.set(normalizedKey, key);
+  }
+  return normalized;
+}
```
Ubax:
```suggestion
function normalizeKey(key: string): string {
  const withoutPrefix = key.replace(/^\.\//, '');
  const ext = path.extname(withoutPrefix);

  return validExtensions.includes(ext)
    ? withoutPrefix.slice(0, -ext.length)
    : withoutPrefix;
}

function findDuplicateKeys(normalizedKeys: readonly string[]): string[] {
  return normalizedKeys.filter(
    (normalizedKey, index) => normalizedKeys.indexOf(normalizedKey) !== index
  );
}

export function normalizeKeys(keys: string[]): Record<string, string> {
  const normalizedKeys = keys.map(normalizeKey);
  const duplicateKeys = findDuplicateKeys(normalizedKeys);

  if (duplicateKeys.length > 0) {
    throw new Error(
      `Multiple routes resolved to the same route: ${duplicateKeys.join(', ')}`
    );
  }

  return Object.fromEntries(
    normalizedKeys.map((normalizedKey, index) => [normalizedKey, keys[index]])
  );
}
```

Please also add tests for `findDuplicateKeys` and `normalizeKey`

packages/expo-router/src/testing-library/context-stubs.ts
```diff
@@ -44,20 +44,70 @@ export function inMemoryContext(context: MemoryContext) {
...
+
 export function requireContextWithOverrides(dir: string, overrides: MemoryContext) {
   const existingContext = requireContext(path.resolve(process.cwd(), dir));
 
+  // Normalize the existing context keys so they align with the extension-free
+  // override keys documented for `inMemoryContext` (e.g. `_layout`). This also
+  // surfaces ambiguous files (e.g. both `index.jsx` and `index.tsx`) up front.
```
Ubax:
```suggestion
```

packages/expo-router/src/testing-library/context-stubs.ts
```diff
@@ -44,20 +44,70 @@ export function inMemoryContext(context: MemoryContext) {
...
+  const normalizedExistingKeys = normalizeKeys(existingContext.keys());
+
+  // Resolve each override to the require-context id it replaces, falling back to
+  // `./name` for overrides that introduce a file not present in `dir`.
+  const overridesByContextKey = new Map<string, FileStub | NativeIntentStub>();
+  for (const [key, value] of Object.entries(overrides)) {
+    const contextKey = normalizedExistingKeys.get(key) ?? `./${key}`;
+    overridesByContextKey.set(contextKey, value);
+  }
+
```
Ubax:
```suggestion
```

I think that this is not necessarily needed.

packages/expo-router/src/testing-library/context-stubs.ts
```diff
@@ -44,20 +44,70 @@ export function inMemoryContext(context: MemoryContext) {
...
     function (id: string) {
-      if (id in overrides) {
-        const route = overrides[id];
+      if (overridesByContextKey.has(id)) {
+        const route = overridesByContextKey.get(id)!;
         return typeof route === 'function' ? { default: route } : route;
-      } else {
-        return existingContext(id);
       }
+      return existingContext(id);
```
Ubax:
```suggestion
      if (id in overrides) {
        const route = overrides[id];
        return typeof route === 'function' ? { default: route } : route;
      } else {
        return existingContext(id);
      }
```

Then we can revert back here

packages/expo-router/src/testing-library/context-stubs.ts
```diff
@@ -44,20 +44,70 @@ export function inMemoryContext(context: MemoryContext) {
...
-      keys: () => [...Object.keys(overrides), ...existingContext.keys()],
+      // Override keys take precedence, so omit any existing key they replace to
+      // avoid listing the same route twice.
+      keys: () => {
+        const overrideKeys = [...overridesByContextKey.keys()];
+        const dedupedExisting = existingContext
+          .keys()
+          .filter((key) => !overridesByContextKey.has(key));
+        return [...overrideKeys, ...dedupedExisting];
+      },
```
Ubax:
```suggestion
    keys: () => [...Object.keys(overrides), ...existingContext.keys()],
```

This can be reverted as well, since the keys are normalized now

Ubax (changes requested):
Thanks for addressing previous comments!

I added some improvements ideas 👍

wwdrew:
@Ubax No worries, hopefully this takes care of it all!

packages/expo-router/src/__tests__/context-stubs.test.node.ts
```diff
@@ -0,0 +1,129 @@
...
+
+function MockLayout() {
+  return null;
+}
+
+describe('normalizeKey', () => {
+  it('strips the leading `./` and a valid extension', () => {
+    expect(normalizeKey('./_layout.tsx')).toBe('_layout');
+    expect(normalizeKey('./index.jsx')).toBe('index');
+    expect(normalizeKey('./nested/route.ts')).toBe('nested/route');
```
Ubax, line 23: nit: Can you add one more for `.js`?

packages/expo-router/src/testing-library/context-stubs.ts
```diff
@@ -44,8 +44,52 @@ export function inMemoryContext(context: MemoryContext) {
...
+  const rawContext = requireContext(path.resolve(process.cwd(), dir));
+
+  // Normalize the require-context keys (`./name.ext`) to the extension-free form
+  // used by override keys, so `overrides` can be matched directly. Keys an
+  // override replaces are dropped here so each route is only listed once.
+  const normalizedKeys = normalizeKeys(rawContext.keys());
+  const existingContext = Object.assign(
+    (id: string) => rawContext(normalizedKeys[id] ?? id),
+    {
+      keys: () => Object.keys(normalizedKeys).filter((key) => !(key in overrides)),
```
Ubax:
```suggestion
      keys: () => Object.keys(normalizedKeys),
```

packages/expo-router/src/testing-library/context-stubs.ts
```diff
@@ -44,8 +44,52 @@ export function inMemoryContext(context: MemoryContext) {
...
+      resolve: (key: string) => key,
+      id: '0',
```
Ubax:
```suggestion
```

Ubax (approved):
Three small addition and we are good to merge! 🎉 

Thank you

packages/expo-router/src/testing-library/context-stubs.ts
```diff
@@ -57,7 +95,12 @@ export function requireContextWithOverrides(dir: string, overrides: MemoryContex
       }
     },
     {
-      keys: () => [...Object.keys(overrides), ...existingContext.keys()],
+      // Overrides take precedence, so drop any existing key they replace to keep
+      // each route listed once (a duplicate key makes route resolution throw).
+      keys: () => [
+        ...Object.keys(overrides),
+        ...existingContext.keys().filter((key) => !(key in overrides)),
+      ],
```
Ubax:
```suggestion
      keys: () => [...Object.keys(overrides), ...existingContext.keys()],
```

Ubax:
@wwdrew thanks for your contribution! I'll merge this when the CI is done

wwdrew:
Heads up — the `Apply suggestion from @Ubax` commit (`6b2b8fd`) reverting `keys()` to `[...Object.keys(overrides), ...existingContext.keys()]` reintroduces the bug this PR fixes. 🙁

Since the existing context keys are now normalized, an override key and the file it replaces collide (e.g. `_layout` from `overrides` **and** `_layout` from `existingContext.keys()`), so `keys()` lists the route twice. That duplicate then throws during route resolution:

```
The layouts "_layout" and "_layout" conflict on the route "/_layout". Remove or rename one of these files.
```

Repro (fails on the current branch):

```ts
const ctx = requireContextWithOverrides(appDir, { _layout: MockLayout });
getExactRoutes(ctx); // throws "layouts ... conflict"
```

It also fails the `lists normalized keys and does not duplicate an overridden file` unit test.

The dedup is still needed — I'd previously moved it to the outer `keys()` for exactly this reason:

```ts
keys: () => [
  ...Object.keys(overrides),
  ...existingContext.keys().filter((key) => !(key in overrides)),
],
```

Happy to re-apply that (or handle the dedup however you prefer), just didn't want to revert your commit without checking.

packages/expo-router/src/testing-library/context-stubs.ts
```diff
@@ -44,8 +44,46 @@ export function inMemoryContext(context: MemoryContext) {
...
 export function requireContextWithOverrides(dir: string, overrides: MemoryContext) {
-  const existingContext = requireContext(path.resolve(process.cwd(), dir));
+  const rawContext = requireContext(path.resolve(process.cwd(), dir));
+
+  // Normalize the require-context keys (`./name.ext`) to the extension-free form
+  // used by override keys, so `overrides` can be matched directly.
+  const normalizedKeys = normalizeKeys(rawContext.keys());
+  const existingContext = Object.assign((id: string) => rawContext(normalizedKeys[id] ?? id), {
+    keys: () => Object.keys(normalizedKeys),
+  });
```
Ubax, line 87:
```suggestion

const uniqueKeys = Array.from(
  new Set([...Object.keys(overrides), ...existingContext.keys()])
);
```

packages/expo-router/src/testing-library/context-stubs.ts
Ubax:
```suggestion
      keys: () => [...uniqueKeys],
```

Ubax (commented):
@wwdrew thanks for clarification. By mistake I took the spread as object with the uniqueness. 

I added suggestion which should work with the new approach and make the key uniqueness clear.

skdrew:
@Ubax No worries, taken care of that one now and tests are looking happy again!

Thank you again for your help on this, really appreciated!

PR #47290, opened 2026-06-26
[ios][updates] Add `excludeFromBackup` option

alanjhughes:
# Why
By default the updates directory `.expo-internal`, is included in device backups. It holds downloaded update bundles, their assets, and the updates database, so for apps with large updates it can dominate the iCloud backup. Update bundles are re-downloadable, so backing them up is wasted space.

# How
Added a new `updates.excludeFromBackup` config option. When enabled, we set `isExcludedFromBackup` on the updates directory.

# Test Plan
CI
Added new tests


Kudo (approved):

brentvatne:
@expo-bot verify --fable

PR #47293, opened 2026-06-26
[dev-launcher][iOS] Add build's expiration date to Settings 

gabrieldonadel:
# Why

Development and internal distribution build embed a provisioning profile with an `ExpirationDate`, once that date passes, iOS refuses to launch the app. Currently, we don't display this information anywhere.

# How

Update dev-launcher and Expo Go to display how many days the app will be available before it expires.

We read that directly from `embedded.mobileprovision`, which is present only on development, ad hoc, and enterprise builds, causing this not to show up in App Store and Simulator builds.

The change spans three packages:

Also had to update expo-application to expose `ExpirationDate` for Expo Go, it already parses the embedded profile plist to determine the app release type, so it is the natural home for this.

# Test Plan

Build a BareExpo and Expo Go to a physical device, then open Settings and confirm that "Expires in" row is displayed correctly.

Then cross-check the value against the profile itself:

```sh
security cms -D -i "<App>.app/embedded.mobileprovision" | plutil -extract ExpirationDate raw -
```

### DIFF
--- a/packages/expo-dev-launcher/ios/EXDevLauncherController.m
+++ b/packages/expo-dev-launcher/ios/EXDevLauncherController.m
@@ -685,7 +685,7 @@ static const NSTimeInterval EXDevLauncherDefaultRequestTimeout = 10.0;
 
 -(nullable NSString *)getAppExpirationDate
 { 
-  // App Store and Simulator builds have don't have mobileprovision
+  // App Store and Simulator builds don't have mobileprovision
   NSString *path = [[NSBundle mainBundle] pathForResource:@"embedded" ofType:@"mobileprovision"];
   if (!path) {
     return nil;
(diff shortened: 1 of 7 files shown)


packages/expo-dev-launcher/ios/EXDevLauncherController.m
```diff
@@ -675,9 +675,71 @@ -(NSDictionary *)getBuildInfo
...
+  if (appExpirationDate) {
+    [buildInfo setObject:appExpirationDate forKey:@"appExpirationDate"];
+  }
+
   return buildInfo;
 }
 
+-(nullable NSString *)getAppExpirationDate
+{ 
+  // App Store and Simulator builds have don't have mobileprovision
```
alanjhughes:
```suggestion
  // App Store and Simulator builds don't have mobileprovision
```

alanjhughes (approved):

PR #47298, opened 2026-06-26
[docs] Lazy-load Kapa Ask AI SDK out of the shared _app bundle

amandeepmittal:
# Why

Fix ENG-22178

# How

- Remove the app-root `KapaProvider` from `pages/_app.tsx`.
- Add `ui/components/AskPageAI/AskPageAILazyMount.tsx` to wrap `KapaProvider` and the Ask AI overlay, isolating `@kapaai/react-sdk` into its own chunk.
- Mount the overlay in `DocumentationPage` via `next/dynamic({ ssr: false })`, gated behind a `hasActivatedAskAI` latch so an SDK page loads only on first Ask AI open or `sessionStorage` restore, and stays mounted to preserve the conversation.
- Import the Ask AI triggers in `PageHeader` directly from `./AskPageAITrigger` instead of the `AskPageAI` barrel, so the chat and overlay are no longer pulled into `_app`.

# Test Plan

Run docs app locally and confirm Ask AI widget and overlay are working as expected.

# Checklist

### DIFF
--- a/docs/components/DocumentationPage.tsx
+++ b/docs/components/DocumentationPage.tsx
@@ -22,6 +22,7 @@ import { buildBreadcrumbListSchema, buildTechArticleSchema } from '~/constants/s
 import { usePageApiVersion } from '~/providers/page-api-version';
 import versions from '~/public/static/constants/versions.json';
 import { PageMetadata } from '~/types/common';
+import { AskPageAILoading } from '~/ui/components/AskPageAI/AskPageAILoading';
 import { Footer } from '~/ui/components/Footer';
 import { Header } from '~/ui/components/Header';
 import { InlineHelp } from '~/ui/components/InlineHelp';
@@ -39,6 +40,7 @@ const { LATEST_VERSION } = versions;
 
 const AskPageAILazyMount = dynamic(() => import('~/ui/components/AskPageAI/AskPageAILazyMount'), {
   ssr: false,
+  loading: () => <AskPageAILoading />,
 });
 
 export type DocPageProps = PropsWithChildren<PageMetadata>;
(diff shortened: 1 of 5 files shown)


docs/ui/components/AskPageAI/AskPageAILazyMount.tsx
```diff
@@ -0,0 +1,23 @@
+import { KapaProvider } from '@kapaai/react-sdk';
+
+import { AskPageAIOverlay } from './AskPageAIOverlay';
+
+const KAPA_INTEGRATION_ID = '2063233f-1e70-45e8-b1b5-a872c9887afc';
+
+type AskPageAILazyMountProps = {
```
kadikraman: Should be able to use `type AskPageAILazyMountProps = ComponentProps<typeof AskPageAIOverlay>` if the props are identical.

docs/components/DocumentationPage.tsx
```diff
@@ -37,6 +37,10 @@ import { A } from '~/ui/components/Text';
 
 const { LATEST_VERSION } = versions;
 
+const AskPageAILazyMount = dynamic(() => import('~/ui/components/AskPageAI/AskPageAILazyMount'), {
+  ssr: false,
+});
```
kadikraman, line 44: Might be ok if it's fast to load in but FYI there's also the option to pass in a `loading` component https://nextjs.org/docs/pages/guides/lazy-loading

kadikraman (approved):

docs/components/DocumentationPage.tsx
amandeepmittal: Done. The Loading stater will be noticeable on slow network only. 


https://github.com/user-attachments/assets/26f4f4c5-be59-4ad8-a336-bde83809741f

PR #47305, opened 2026-06-26
refactor(expo-audio): implement iOS AVAudioEngineBackend for local file playback

Capstan:
# Why

Refactors the iOS player engine to lay down a pluggable backend interface. This is required to support low-level DSP nodes (such as pitch-shifters and audio taps) for local playback files via `AVAudioEngine` without breaking standard remote HLS fallbacks.

# How

- **AudioPlayerBackendProtocol**: Defines the Swift playback interface contract.
- **AVPlayerBackend**: Core remote HLS audio fallback wrapper. *Note: Extracted directly from the original `AVPlayer` code in `AudioPlayer.swift` with minimal changes.*
- **AVAudioEngineBackend**: Local audio engine graph setup containing player, time-pitch, and mixer nodes.
- **AudioPlayer**: Swaps backends and synchronizes player state (rate, volume, loop, status) dynamically based on loaded source. It now acts as a routing shell that delegates to the active backend.

# Test Plan

- Verified with all existing integration tests in `test-suite`. Run in the iOS simulator:
  `xcrun simctl openurl booted "bareexpo://test-suite/run?tests=Audio"`

# Checklist


behenate:
Hi, thanks for the PR, we have discussed this internally and came to a conclusion that, as of now, this feature is out of scope for expo-audio. We try to keep our modules relatively straightforward as there is a lot of them and we have limited time for maintenance. This feature increases complexity while providing functionality that would be used by a slim margin of users. Thank you again for your effort!

PR #47317, opened 2026-06-27
[crypto][android] Fix AESSealedData.fromCombined() throwing when given a base64 string

### LINKED ISSUE #47274: [expo-crypto] `AESSealedData.fromCombined(sealedDataBase64String)` fails on Android

https://github.com/MC-DS/expo-crypto-aes-sealed-data-from-combined-bug

### Steps to reproduce

- Clone the project
- Open an Android emulator
- Run `npm ci`
- Run `npm run android`
- Click the "Encrypt and Decrypt" button

Expected behaviour:
- The encrypted and decrypted values are logged to the console

Actual behaviour:
- An error is thrown in the `decrypt` function when calling `AESSealedData.fromCombined`

The [TypeScript function](https://github.com/expo/expo/blob/b1bf627f01772077dd73a4976c3e8663f8dce465/packages/expo-crypto/src/aes/index.ts#L68) claims to accept `BinaryData`, but the [native Kotlin function](https://github.com/expo/expo/blob/b1bf627f01772077dd73a4976c3e8663f8dce465/packages/expo-crypto/android/src/main/java/expo/modules/crypto/aes/AesCryptoModule.kt#L80) has a `ByteArray` parameter rather than a `BinaryInput` one.

### Environment

```text
    System:
      OS: Windows 11 10.0.26200
    Binaries:
      Node: 24.11.1 - C:\Program Files\nodejs\node.EXE
      Yarn: 1.22.22 - C:\Program Files\nodejs\yarn.CMD
      npm: 11.17.0 - C:\Program Files\nodejs\npm.CMD
    IDEs:
      Android Studio: AI-261.23567.138.2611.15613797
    npmPackages:
      expo: ~56.0.12 => 56.0.12 
      react: 19.2.3 => 19.2.3 
    Expo Workflow: bare
```

### Expo Doctor Diagnostics

```text
```

gabrieldonadel:
# Why

Fixes https://github.com/expo/expo/issues/47274

`AESSealedData.fromCombined()` is documented to accept a `BinaryInput`, where "when providing a string, it must be base64-encoded." That works on iOS and Web, but on Android passing a base64 string throws inside `fromCombined`.

# How

Changed the Android `fromCombined` parameter from `ByteArray` to `BinaryInput` and decoded it with the module's existing `toBytes()` helper, matching iOS. This is purely a widening of accepted input: a `ByteArray` still flows through unchanged (`toBytes()` returns it as-is), and base64 strings are now decoded instead of throwing. The public TypeScript type and JS signature are unchanged, so there is no API break.

# Test Plan

The repro from the issue (encrypt, then `AESSealedData.fromCombined(base64String)` and decrypt) now succeeds on Android 

# Checklist

### DIFF
--- a/packages/expo-crypto/CHANGELOG.md
+++ b/packages/expo-crypto/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### 🐛 Bug fixes
 
+- Fix `AESSealedData.fromCombined()` throwing on Android when given a base64-encoded string. ([#47274](https://github.com/expo/expo/issues/47274))
+
 ### 💡 Others
 
 ## 56.0.4 — 2026-05-26
--- a/packages/expo-crypto/CHANGELOG.md
+++ b/packages/expo-crypto/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### 🐛 Bug fixes
 
-- Fix `AESSealedData.fromCombined()` throwing on Android when given a base64-encoded string. ([#47274](https://github.com/expo/expo/issues/47274))
+- Fix `AESSealedData.fromCombined()` throwing on Android when given a base64-encoded string. ([#47317](https://github.com/expo/expo/pull/47317) by [@gabrieldonadel](https://github.com/gabrieldonadel))
 
 ### 💡 Others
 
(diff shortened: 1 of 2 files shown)


barthap (approved):

packages/expo-crypto/CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 ### 🐛 Bug fixes
 
+- Fix `AESSealedData.fromCombined()` throwing on Android when given a base64-encoded string. ([#47274](https://github.com/expo/expo/issues/47274))
```
gabrieldonadel:
```suggestion
- Fix `AESSealedData.fromCombined()` throwing on Android when given a base64-encoded string. ([#47317](https://github.com/expo/expo/pull/47317) by [@gabrieldonadel](https://github.com/gabrieldonadel))
```

PR #47318, opened 2026-06-27
[docs] Migrate Tabs from @reach/tabs to @radix-ui/react-tabs

amandeepmittal:
# Why

Fix ENG-22070

# How

- Replace `@reach/tabs` with `@radix-ui/react-tabs` and migrate `Tabs`/`TabButton`/`Tab` to Radix `Root`/`List`/`Trigger`/`Content`, bridging the index-based API to Radix's string `value` via `String(index)`.
- Render `Content` with `forceMount` + `hidden` so every panel stays in the server HTML (preserves SSR/SEO, avoids CLS), and keep the framer-motion `layoutId` active indicator.
- Add a `data-md="tabs"` marker on `Root` so the markdown export can locate tab groups without Reach internals, and rely on Radix for roles, roving tabindex, and arrow-key navigation.
- Retarget `cleanHtml` and `check-markdown-pages.ts` from `data-reach-*` to `data-md="tabs"`/`role="tab"`/`role="tabpanel"`, scoping label collection to each group's own `tablist` so nested `role="tab"` Terminals do not pollute panel headings.
- Update the markdown export test fixtures and add a nested-`role="tab"` regression test.
- Add `TabsGroup.test.tsx` covering synced groups, standalone independence, and arrow-key navigation.

# Test Plan

See preview.

# Checklist


kadikraman (approved):
Tested on the PR preview - LGTM!

Issue #36831, opened 2025-05-12
Broken Tests In SDK 53 Which Passed in SDK 52


pachun:
### Minimal reproducible example

https://github.com/pachun/expo-sdk-53-test-failures

### Steps to reproduce

# Minimal Reproduction of Expo SDK 53 Test Failures Which Pass in SDK 52

There are two test cases in this MRE app, in a single file. Both test cases
run the exact same test. When either test is run individually, they each
pass. When they're run together, the first test passes and causes all
remaining tests to fail in Expo SDK 53.

These two tests pass when run together if run in an Expo SDK 52 project.

The test case being reused here taps a button, which makes an API call
that's mocked with nock and then pushes a new screen onto the expo
router's stack.

You can see from the console.log's that subsequent runs of the test case
cause the `await response.json()` line to hang; because no logs are
shown after that line in subsequent test runs.

I'm not sure if this is because RN replaced the old node fetch with a
native fetch, which is what nock is capable of mocking.

I tried adding `yarn add node-fetch@2` and then importing it in the
app/index.tsx screen with `import fetch from "node-fetch"`.

However, I continued to see the same type of successes and failures (an
individual test passes and multiple tests being run together causes all
subsequent tests to fail).

That leads me to believe the problem is elsewhere, in an upgrade to
either react, react-native, expo-router, expo itself, or jest-expo
which all need to be upgraded in lock step with each other, so I'm
unsure how to narrow down the problem from here.

The major changes that's causing this problem (I've found, as a result
of creating this minimal reproducible example), must be caused by one
of the following upgrades:

---

expo (from) 52.0.46 (to) 53.0.9

react (from) 18.3.1 (to) 19.0.0

react-native (from) 0.76.9 (to) 0.79.2

jest-expo (from) 52.0.6 (to) 53.0.5

---

## Steps to Reproduce

1. Clone and cd into the MRE repo (https://github.com/pachun/expo-sdk-53-test-failures)
2. `yarn`
3. Run a single test to see it pass with: `node_modules/.bin/jest --runTestsByPath -t '^the same two tests pass when run individually$' -- "__tests__/test.test.tsx"`
4. Run both tests to see the latter one fail and note the where the console.logs get cut off after the `await response.json()` line: `node_modules/.bin/jest --runTestsByPath -- "__tests__/test.test.tsx"`

Downgrading to Expo SDK 52 allows running the command in step 4 to see both tests pass.

### Environment

```text
expo-env-info 1.3.3 environment info:
[...]

PR #47332, opened 2026-06-29
[expo][jest-expo] Fix Jest 30 "import outside scope" teardown crash

janpe:
_Based on @netconomy-stephan-dum's #43535 — this rebases that fix onto `main` and adds the root-cause writeup, changelog, and verification. Happy to close this in favor of #43535 if it moves forward instead._

## Why

On **Jest 30**, every `jest-expo` test fails during environment teardown — even a trivial passing one — with:

```
ReferenceError: You are trying to `import` a file outside of the scope of the test code.
```

`expo/src/winter/runtime.native.ts` registers `structuredClone` and `__ExpoImportMetaRegistry` as **lazy** globals (the backing module is `require`d only on first read). Jest 30's teardown (`resetModules`) now reads every global to clear mocks, firing those lazy `require`s after the test scope has closed — and loading a module then throws. The other winter globals are non-enumerable, so teardown skips them; only these two are hit.

Refs #36831, #37261.

## How

Resolve the two globals eagerly at module load instead of lazily `require`-ing inside the getter:
- `__ExpoImportMetaRegistry` → top-level `import`.
- `structuredClone` → a small `structuredClone.ts` wrapper over `@ungap`'s `serialize`/`deserialize` (its default export delegates to the global `structuredClone`, which would recurse once we install it).

The globals stay enumerable; only their resolution timing moves to module load.

## Alternative

A smaller fix is possible: mark these two globals **non-enumerable**, so Jest's teardown loop skips them and the lazy `require` never fires. That drops the wrapper and the `@types` dependency and keeps the existing lazy loading — at the cost of changing an observable attribute (enumerability) of those globals. I went with eager resolution because it's behavior-transparent, but happy to switch to the non-enumerable version if you'd prefer the smaller surface.

## Test Plan

- `et check-packages expo` passes (build, typecheck, oxlint, format); `structuredClone.test.ios.ts` 2/2.
- Jest-30 repro (can't run in this repo's Jest-29 CI): https://github.com/netconomy-stephan-dum/jest-expo-with-jest-30 — fails on `main`, passes with this change.

## Checklist
- [x] `CHANGELOG.md` entry
- [x] `et check-packages expo` green
- [x] No `build/` staged

### DIFF
--- a/packages/expo/CHANGELOG.md
+++ b/packages/expo/CHANGELOG.md
@@ -21,6 +21,7 @@
 - [iOS] Add ExpoBundleConfiguration to derive RCTBundleConfiguration from the normalized bundle URL instead of default shared settings singleton ([#48010](https://github.com/expo/expo/pull/48010) by [@kitten](https://github.com/kitten))
 - [iOS] Resolve the dev server port from the `RCTMetroPort` Info.plist key at runtime so bare projects without expo-dev-client connect to their own Metro instance instead of defaulting to 8081. ([#48098](https://github.com/expo/expo/pull/48098) by [@alanjhughes](https://github.com/alanjhughes))
 - Fix async imports (`import(...)`) via `asyncRequireModule` not a thenable instead of a full promise shape ([#48550](https://github.com/expo/expo/pull/48550) by [@kitten](https://github.com/kitten))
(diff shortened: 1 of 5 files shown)


janpe:
@brentvatne I saw that you closed #43535 in favor of this one. I've rebased this one now.

PR #47337, opened 2026-06-29
[expo-image] add `imageLoaded` event

Ubax:
# Why

In order to integrate expo-image with expo-observe we need information when image is loaded.

# How

Add `onImageLoaded` event on iOS and Android and emit it when an image is loaded

# Test Plan

Tested in the follow-up PR

# Checklist

### DIFF
--- a/packages/expo-image/ios/ImageModule.swift
+++ b/packages/expo-image/ios/ImageModule.swift
@@ -8,8 +8,17 @@ internal import SDWebImageSVGCoder
 public final class ImageModule: Module {
   lazy var prefetcher = SDWebImagePrefetcher.shared
 
+  // Whether JS subscribed to `imageLoaded`. Skips the per-image-load bridge hop when nothing is
+  // listening
+  private var hasImageLoadedListener = false
+
   public func definition() -> ModuleDefinition {
     Name("ExpoImage")
+
+    Events("imageLoaded")
+
+    OnStartObserving("imageLoaded") { self.hasImageLoadedListener = true }
+    OnStopObserving("imageLoaded") { self.hasImageLoadedListener = false }
 
     OnCreate {
       ImageModule.registerCoders()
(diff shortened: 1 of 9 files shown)


Ubax:
* **#47145** <a href="https://app.graphite.com/github/pr/expo/expo/47145?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#47337** <a href="https://app.graphite.com/github/pr/expo/expo/47337?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/47337?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/expo-image/ios/ImageModule.swift
```diff
@@ -304,6 +320,17 @@ public final class ImageModule: Module {
     }
   }
 
+  func emitImageLoaded(url: String, width: CGFloat, height: CGFloat) {
+    guard hasImageLoadedListener, width > 0, height > 0, !url.isEmpty else {
+      return
+    }
+    sendEvent("onImageLoaded", [
```
tsapeta: Use `emit(event: "imageLoaded", payload: [...])` instead – newer, unified API in both modules and shared objects

packages/expo-image/src/Image.types.ts
```diff
@@ -792,9 +792,21 @@ export declare class ImageRef extends SharedRef<'image'> {
...
+ * Module-level events emitted by the native `ExpoImage` module.
  * @hidden
  */
-export declare class ImageNativeModule extends NativeModule {
+export type ImageModuleEvents = {
+  /**
+   * Fires from every relevant load path (`loadAsync`, `useImage`, and the rendered `<Image>` view)
+   * with the decoded pixel size.
+   */
+  onImageLoaded: (event: { url: string; width: number; height: number }) => void;
```
tsapeta: We're going to unify the event names at some point. Most of the events in the SDK don't have the `on` prefix (except views where it's actually required by React).
```suggestion
  imageLoaded: (event: { url: string; width: number; height: number }) => void;
```

packages/expo-image/ios/ImageView.swift
```diff
@@ -270,6 +270,12 @@ public final class ImageView: ExpoView {
         ]
       ])
 
+      (appContext?.moduleRegistry.get(moduleWithName: "ExpoImage") as? ImageModule)?.emitImageLoaded(
```
tsapeta: This reads a bit better (still not ideal though)
```suggestion
      appContext?.moduleRegistry.getModule(implementing: ImageModule.self)?.emitImageLoaded(
```

tsapeta (approved):

PR #47341, opened 2026-06-29
[app-metrics] Capture React render-phase errors via an error boundary

tsapeta:
## Why

React render-phase errors never reach the global `ErrorUtils` handler (React routes them through error boundaries), so they weren't captured at all. The React component stack, which is the most useful signal for locating the failure, is only available inside an error boundary.

Details and design notes: [ENG-21894](https://linear.app/expo/issue/ENG-21894/sdk-handle-react-render-errors-errorboundary).

## How

- Add `AppMetricsErrorBoundary`, re-exported from `expo-observe` as `ObserveErrorBoundary`, that records render-phase errors as non-fatal `exception` log events (OpenTelemetry conventions) with the React component stack attached as `expo.error.component_stack`. Reported non-fatal because the boundary renders a fallback and the app keeps running, like Sentry's handled error-boundary captures.
- `fallback` is required (pass an element, or explicit `null` to render nothing). The boundary always renders something, so it never re-propagates; there is no re-throw mode, since a boundary can't faithfully reproduce React Native's uncaught-error path.
- `AppMetricsRoot` / `ObserveRoot` mount a boundary only when given a non-null `errorBoundaryFallback`. Without it, no boundary is mounted and render errors keep React Native's default behavior, so an app rendering the root with no props is unaffected. Those errors are still recorded by the global handler.
- Add the `errorBoundary` value to the `ErrorReport` source enum on JS, iOS, and Android. `expo.error.component_stack` is omitted when absent, consistently on both platforms.

## Test Plan

- `et check-packages expo-app-metrics expo-observe` passes.
- Jest: boundary renders children when healthy, renders the element or `null` fallback on catch, reports a non-fatal `errorBoundary` exception with the component stack, handles a falsy `throw null` without looping, and never lets a `reportError` failure escape; `AppMetricsRoot` mounts no boundary without a fallback; `ObserveErrorBoundary` aliases and works through the Observe entry point.
- Android and iOS native unit tests pass (`ErrorReport` includes `expo.error.component_stack` when present, omits it when absent).
- Verified on iOS in observe-tester: render-error triggers record `errorBoundary` exception events with the expected attributes ([example session](https://expo.dev/accounts/expo/projects/observability/observe/sessions/c02978e0-b9fb-4244-9b09-2201f5c6db13)).

### DIFF
--- a/packages/expo-app-metrics/CHANGELOG.md
+++ b/packages/expo-app-metrics/CHANGELOG.md
@@ -12,8 +12,7 @@
 - Add android crash reports ([#46869](https://github.com/expo/expo/pull/46869) by [@Ubax](https://github.com/Ubax))
 - Record an `expo.memory.warning` log event on iOS when the system delivers a low-memory warning, carrying the memory usage snapshot (`expo.memory.*`) taken at warning time. ([#47108](https://github.com/expo/expo/pull/47108) by [@tsapeta](https://github.com/tsapeta))
 - Add an optional `displayName` to `logEvent` ([#47289](https://github.com/expo/expo/pull/47289) by [@Ubax](https://github.com/Ubax))
-||||||| parent of 63fe4cf1fa5 ([app-metrics][observe] Capture React render-phase errors via an error boundary)
-- Capture React render-phase errors via the new `AppMetricsErrorBoundary`, recorded as non-fatal `exception` log events with the React component stack (`expo.error.component_stack`). Place it around any subtree with a `fallback`, or pass `errorBoundaryFallback` to `AppMetricsRoot` to mount one at the root. ([#47038](https://github.com/expo/expo/pull/47038) by [@tsapeta](https://github.com/tsapeta))
+- Capture React render-phase errors via the new `AppMetricsErrorBoundary`, recorded as non-fatal `exception` log events with the React component stack (`expo.error.component_stack`). Place it around any subtree with a `fallback`, or pass `errorBoundaryFallback` to `AppMetricsRoot` to mount one at the root. ([#47341](https://github.com/expo/expo/pull/47341) by [@tsapeta](https://github.com/tsapeta))
 
 ### 🐛 Bug fixes
 
--- a/packages/expo-app-metrics/CHANGELOG.md
+++ b/packages/expo-app-metrics/CHANGELOG.md
@@ -12,7 +12,7 @@
 - Add android crash reports ([#46869](https://github.com/expo/expo/pull/46869) by [@Ubax](https://github.com/Ubax))
 - Record an `expo.memory.warning` log event on iOS when the system delivers a low-memory warning, carrying the memory usage snapshot (`expo.memory.*`) taken at warning time. ([#47108](https://github.com/expo/expo/pull/47108) by [@tsapeta](https://github.com/tsapeta))
 - Add an optional `displayName` to `logEvent` ([#47289](https://github.com/expo/expo/pull/47289) by [@Ubax](https://github.com/Ubax))
-- Capture React render-phase errors via the new `AppMetricsErrorBoundary`, recorded as non-fatal `exception` log events with the React component stack (`expo.error.component_stack`). Place it around any subtree with a `fallback`, or pass `errorBoundaryFallback` to `AppMetricsRoot` to mount one at the root. ([#47341](https://github.com/expo/expo/pull/47341) by [@tsapeta](https://github.com/tsapeta))
+- Capture React render-phase errors via `AppMetricsErrorBoundary`. ([#47341](https://github.com/expo/expo/pull/47341) by [@tsapeta](https://github.com/tsapeta))
 
 ### 🐛 Bug fixes
 
--- a/packages/expo-app-metrics/CHANGELOG.md
+++ b/packages/expo-app-metrics/CHANGELOG.md
@@ -12,6 +12,8 @@
 - Add android crash reports ([#46869](https://github.com/expo/expo/pull/46869) by [@Ubax](https://github.com/Ubax))
 - Record an `expo.memory.warning` log event on iOS when the system delivers a low-memory warning, carrying the memory usage snapshot (`expo.memory.*`) taken at warning time. ([#47108](https://github.com/expo/expo/pull/47108) by [@tsapeta](https://github.com/tsapeta))
 - Add an optional `displayName` to `logEvent` ([#47289](https://github.com/expo/expo/pull/47289) by [@Ubax](https://github.com/Ubax))
+||||||| parent of 63fe4cf1fa5 ([app-metrics][observe] Capture React render-phase errors via an error boundary)
+- Capture React render-phase errors via the new `AppMetricsErrorBoundary`, recorded as non-fatal `exception` log events with the React component stack (`expo.error.component_stack`). Place it around any subtree with a `fallback`, or pass `errorBoundaryFallback` to `AppMetricsRoot` to mount one at the root. ([#47038](https://github.com/expo/expo/pull/47038) by [@tsapeta](https://github.com/tsapeta))
 
 ### 🐛 Bug fixes
 
--- a/packages/expo-app-metrics/src/AppMetricsErrorBoundary.tsx
+++ b/packages/expo-app-metrics/src/AppMetricsErrorBoundary.tsx
@@ -1,0 +1,86 @@
+import React from 'react';
+
+import AppMetrics from './module';
+
+export type AppMetricsErrorBoundaryProps = {
+  children: React.ReactNode;
+  /**
+   * What to render in place of the failed subtree after an error is caught. Pass an element to
+   * render it instead of the broken children, keeping the app usable.
+   *
+   * Required, but explicitly pass `null` to opt out and render nothing in place of the subtree. The
+   * boundary always renders something (the element or nothing), so a render error never propagates
+   * past it; React boundaries can't re-throw to reproduce the original crash, so capture-only
+   * boundaries aren't supported. Render-phase errors that no boundary catches are still recorded by
+   * the global `ErrorUtils` handler `expo-app-metrics` installs on import.
+   */
+  fallback: React.ReactElement | null;
+};
+
+type State = {
+  /**
+   * Whether an error has been caught. Tracked separately from the value because a thrown value can
+   * itself be falsy (e.g. `throw null`); keying "caught" off the value would make such a throw look
+   * like a healthy state and re-render the children into an infinite loop.
+   */
+  hasError: boolean;
+  /**
+   * The caught value. Whatever was thrown, which is not necessarily an `Error` (code can throw a
+   * string, `null`, or any value); only meaningful when `hasError` is `true`.
+   */
+  error: unknown;
+};
+
+/**
+ * Error boundary that records React render-phase errors as `exception` log events (following
+ * OpenTelemetry's exception conventions, same as the global `ErrorUtils` handler) and renders a
+ * `fallback` in place of the subtree that threw.
+ *
+ * Render-phase errors don't reach `global.ErrorUtils` — React routes them through error boundaries —
+ * so a boundary is the only way to observe them with the React component stack attached. Place one
+ * around any subtree you want to keep capturing and recoverable; `AppMetricsRoot` mounts one
+ * automatically when given an `errorBoundaryFallback`.
+ *
+ * The error is reported as non-fatal: the boundary renders the fallback and the app keeps running,
+ * so it's a handled error, not a crash. Reporting runs in a `try/catch` so a failure inside
+ * `reportError` can never become a new app-facing error.
+ */
+export class AppMetricsErrorBoundary extends React.Component<AppMetricsErrorBoundaryProps, State> {
+  state: State = { hasError: false, error: null };
+
+  static getDerivedStateFromError(error: unknown): State {
+    return { hasError: true, error };
+  }
+
+  componentDidCatch(error: unknown, errorInfo: React.ErrorInfo): void {
+    // The thrown value may not be an `Error` (code can throw anything), so read its fields
+    // defensively and fall back to `String(error)` for the message.
+    const caught = error instanceof Error ? error : undefined;
+    // React's component stack has a leading newline and indentation on each frame; trim the
+    // surrounding whitespace so it reads cleanly.
+    const componentStack = errorInfo?.componentStack?.trim();
+
+    try {
+      AppMetrics.reportError({
+        source: 'errorBoundary',
+        type: caught?.name,
+        message: caught?.message ?? String(error),
+        stacktrace: caught?.stack,
+        componentStack,
+        // The boundary renders the fallback and the app keeps running, so the error is handled, not
+        // fatal (mirrors how Sentry marks error-boundary captures as `handled`).
+        isFatal: false,
+      });
+    } catch {
+      // Never let a reporting failure (e.g. a native module error) escape the boundary and become a
+      // new app-facing error. Losing one report is strictly better than crashing.
+    }
+  }
+
+  render(): React.ReactNode {
+    if (!this.state.hasError) {
+      return this.props.children;
+    }
+    return this.props.fallback;
+  }
+}
--- a/packages/expo-app-metrics/src/AppMetricsErrorBoundary.tsx
+++ b/packages/expo-app-metrics/src/AppMetricsErrorBoundary.tsx
@@ -85,9 +85,13 @@ export class AppMetricsErrorBoundary extends React.Component<AppMetricsErrorBoun
         // fatal (mirrors how Sentry marks error-boundary captures as `handled`).
         isFatal: false,
       });
-    } catch {
+    } catch (reportingError) {
       // An observability boundary must never become the source of a new error, so swallow a failure
-      // inside `reportError`. Losing one report is better than crashing.
+      // inside `reportError`. Losing one report is better than crashing. Surface it in dev so a
+      // dropped report is at least noticeable while developing.
+      if (__DEV__) {
+        console.warn('[expo-app-metrics] Failed to report a caught error:', reportingError);
+      }
     }
   }
 
--- a/packages/expo-app-metrics/src/AppMetricsErrorBoundary.tsx
+++ b/packages/expo-app-metrics/src/AppMetricsErrorBoundary.tsx
@@ -2,48 +2,59 @@ import React from 'react';
 
 import AppMetrics from './module';
 
+/**
+ * Arguments passed to a `fallback` render function.
+ */
+export type AppMetricsErrorBoundaryFallbackProps = {
+  /**
+   * The value the subtree threw. Usually an `Error`, but any value can be thrown.
+   */
+  error: unknown;
+  /**
+   * Clears the caught error and re-renders the children. Use it to offer a "try again" action;
+   * the children re-mount, so they run from a clean state.
+   */
+  resetError: () => void;
+};
+
 export type AppMetricsErrorBoundaryProps = {
   children: React.ReactNode;
   /**
-   * What to render in place of the failed subtree after an error is caught. Pass an element to
-   * render it instead of the broken children, keeping the app usable.
+   * Rendered in place of the subtree after an error is caught. Provide one of:
    *
-   * Required, but explicitly pass `null` to opt out and render nothing in place of the subtree. The
-   * boundary always renders something (the element or nothing), so a render error never propagates
-   * past it; React boundaries can't re-throw to reproduce the original crash, so capture-only
-   * boundaries aren't supported. Render-phase errors that no boundary catches are still recorded by
-   * the global `ErrorUtils` handler `expo-app-metrics` installs on import.
+   * - a React element to render as-is,
+   * - a function receiving the `error` and a `resetError` callback (to show details and offer retry),
+   * - `null` to render nothing.
+   *
+   * A boundary can't re-throw to reproduce React Native's default crash, so it always renders one of
+   * the above; there's no capture-only mode. Errors no boundary catches are still recorded by the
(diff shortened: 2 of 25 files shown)


packages/expo-app-metrics/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
...
 - Add android crash reports ([#46869](https://github.com/expo/expo/pull/46869) by [@Ubax](https://github.com/Ubax))
 - Record an `expo.memory.warning` log event on iOS when the system delivers a low-memory warning, carrying the memory usage snapshot (`expo.memory.*`) taken at warning time. ([#47108](https://github.com/expo/expo/pull/47108) by [@tsapeta](https://github.com/tsapeta))
+- Capture React render-phase errors via the new `AppMetricsErrorBoundary`, recorded as non-fatal `exception` log events with the React component stack (`expo.error.component_stack`). Place it around any subtree with a `fallback`, or pass `errorBoundaryFallback` to `AppMetricsRoot` to mount one at the root. ([#47038](https://github.com/expo/expo/pull/47038) by [@tsapeta](https://github.com/tsapeta))
```
Ubax: [🚨 **critical**] This links `#47038`, which returns 404 — that PR doesn't exist. This change ships as `#47341` (verified: that's this PR, by @tsapeta). The link will permanently point at nothing once merged. Update to `#47341`.

packages/expo-observe/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 ### 🎉 New features
 
+- Expose `ObserveErrorBoundary`, a React error boundary that records render-phase errors as `exception` log events and renders a `fallback` in place of the crashed subtree. ([#47340](https://github.com/expo/expo/pull/47340) by [@tsapeta](https://github.com/tsapeta))
```
Ubax: [🚨 **critical**] This links `#47340`, which is an unrelated PR

packages/expo-app-metrics/src/AppMetricsRoot.tsx
```diff
@@ -1,17 +1,38 @@
...
+  errorBoundaryFallback,
+}: {
+  children: React.ReactNode;
+  errorBoundaryFallback?: React.ReactElement | null;
+}) {
   useEffect(() => {
     AppMetrics.markFirstRender();
   }, []);
 
+  if (errorBoundaryFallback != null) {
```
Ubax: [🏗️ **design**] `null` has inverted meanings across the two layers: here `errorBoundaryFallback={null}` mounts **no** boundary (render errors crash with RN default behavior), but `AppMetricsErrorBoundary` treats `fallback={null}` as "mount a boundary that catches and renders nothing." A developer who writes `<ObserveRoot errorBoundaryFallback={null}>` reasonably expecting "capture silently, show blank" gets the exact opposite — no capture at all. Two cheap fixes: gate on `errorBoundaryFallback !== undefined` so an explicit `null` still mounts the boundary (render-nothing), matching the boundary's own semantics; or split capture-opt-in from the fallback element. At minimum, the TSDoc should call out that explicit `null` opts out entirely.

packages/expo-app-metrics/src/AppMetricsErrorBoundary.tsx
```diff
@@ -0,0 +1,86 @@
...
+      // Never let a reporting failure (e.g. a native module error) escape the boundary and become a
+      // new app-facing error. Losing one report is strictly better than crashing.
+    }
+  }
+
+  render(): React.ReactNode {
+    if (!this.state.hasError) {
+      return this.props.children;
+    }
+    return this.props.fallback;
```
Ubax: What do you think about aligning this with https://github.com/bvaughn/react-error-boundary/blob/main/lib/components/ErrorBoundary.tsx?

This would surface error to the user and give them an option to reset

packages/expo-app-metrics/src/AppMetricsErrorBoundary.tsx
```diff
@@ -0,0 +1,86 @@
...
+   * Whether an error has been caught. Tracked separately from the value because a thrown value can
+   * itself be falsy (e.g. `throw null`); keying "caught" off the value would make such a throw look
+   * like a healthy state and re-render the children into an infinite loop.
+   */
+  hasError: boolean;
+  /**
+   * The caught value. Whatever was thrown, which is not necessarily an `Error` (code can throw a
+   * string, `null`, or any value); only meaningful when `hasError` is `true`.
+   */
+  error: unknown;
```
Ubax, line 48: [🏗️ **design**] `state.error` is written by `getDerivedStateFromError` but never read: `render` only consults `hasError`, and `componentDidCatch` receives `error` as its own argument. Per the package's minimalism bias, either drop it (`getDerivedStateFromError` returns `{ hasError: true }`, `State` becomes `{ hasError: boolean }`) or make it earn its place by passing it to a function-fallback (see the reset/retry comment). The separate-`hasError`-from-value rationale in the comment is correct, but it justifies tracking `hasError`, not storing the value nothing consumes.

packages/expo-app-metrics/src/AppMetricsErrorBoundary.tsx
```diff
@@ -0,0 +1,86 @@
...
+    return { hasError: true, error };
+  }
+
+  componentDidCatch(error: unknown, errorInfo: React.ErrorInfo): void {
+    // The thrown value may not be an `Error` (code can throw anything), so read its fields
+    // defensively and fall back to `String(error)` for the message.
+    const caught = error instanceof Error ? error : undefined;
+    // React's component stack has a leading newline and indentation on each frame; trim the
+    // surrounding whitespace so it reads cleanly.
+    const componentStack = errorInfo?.componentStack?.trim();
```
Ubax: [💡 suggestion] Two small things on this line. (a) The `errorInfo?.` guard is dead — React always passes a non-null `ErrorInfo` to `componentDidCatch`; only `componentStack` itself is nullable, so `errorInfo.componentStack?.trim()` is the precise form. (b) `"".trim()` is `""`, and the native side only omits `expo.error.component_stack` on `null`/`nil` — an empty component stack would be sent as an empty attribute. Coalesce to `undefined`: `errorInfo.componentStack?.trim() || undefined`.

packages/expo-app-metrics/src/AppMetricsErrorBoundary.tsx
```diff
@@ -0,0 +1,86 @@
...
+ * `reportError` can never become a new app-facing error.
+ */
+export class AppMetricsErrorBoundary extends React.Component<AppMetricsErrorBoundaryProps, State> {
+  state: State = { hasError: false, error: null };
+
+  static getDerivedStateFromError(error: unknown): State {
+    return { hasError: true, error };
+  }
+
+  componentDidCatch(error: unknown, errorInfo: React.ErrorInfo): void {
```
Ubax, line 66: Not necessarily for this PR, but maybe it would be good to add TODO

[💡 suggestion] No throttling/dedup on the report path. A component that throws on every render (a screen that auto-retries, a polling query that errors, or exactly the keyed-remount pattern the demo uses) emits one `reportError` per catch, unbounded — and a boundary with a non-null fallback recovers and can be re-entered immediately, so a tight remount loop produces a torrent of large payloads (the JSONView comment notes stacks can be ~40KB). The global `ErrorUtils` path shares this gap. Ideally dedup/rate-limit natively on a fingerprint (type+message+first frame) so both `global` and `errorBoundary` sources are covered, recording a count rather than N events. Probably a follow-up, but flag it before this becomes a quota/billing surprise.

packages/expo-app-metrics/src/__tests__/AppMetricsErrorBoundary.test.native.tsx
```diff
@@ -0,0 +1,147 @@
...
+      render(
+        <AppMetricsErrorBoundary fallback={<Text testID="fallback">fallback</Text>}>
+          <Boom />
+        </AppMetricsErrorBoundary>
+      )
+    ).not.toThrow();
+    expect(screen.getByTestId('fallback')).toBeVisible();
+  });
+
+  it('reports a caught error only once despite React dev-mode double rendering', () => {
```
Ubax: [🐧 nit] This test name overstates what it proves. `@testing-library/react-native`'s `render` doesn't wrap children in `StrictMode`, so the dev-mode double-invocation of `getDerivedStateFromError` never occurs here — and reporting lives in `componentDidCatch` (called once regardless). The "once" guarantee comes from React's contract, not from anything this test exercises, and it duplicates the earlier `reports a caught error as a non-fatal errorBoundary exception` test minus the payload assertions. Either delete it as redundant, or wrap the tree in `<StrictMode>` so the claim is actually tested.

packages/expo-app-metrics/src/__tests__/AppMetricsErrorBoundary.test.native.tsx
```diff
@@ -0,0 +1,147 @@
...
+      </AppMetricsErrorBoundary>
+    );
+
+    // If the boundary mistook the caught `null` for a healthy state, it would re-render the children
+    // and loop; reaching the fallback proves it stopped.
+    expect(screen.getByTestId('fallback')).toBeVisible();
+    expect(reportError).toHaveBeenCalledTimes(1);
+  });
+
+  it('reports a caught error as a non-fatal errorBoundary exception', () => {
```
Ubax, line 108: Can you add similar test but with fallback defined?

packages/expo-app-metrics/src/AppMetricsRoot.tsx
```diff
@@ -1,17 +1,38 @@
...
  * without manually calling `AppMetrics.markFirstRender()`.
+ *
+ * Pass `errorBoundaryFallback` to also wrap the app in an `AppMetricsErrorBoundary` that records
+ * React render-phase errors and renders the fallback in place of the crashed app. Without it no
+ * boundary is mounted, so render errors keep React Native's default behavior unchanged (an existing
+ * app that renders `AppMetricsRoot` with no props is unaffected). Those errors are still recorded by
+ * the global `ErrorUtils` handler; only the boundary-only React component stack is unavailable.
+ *
+ * To render an error boundary deeper in the tree instead of at the root, use `AppMetricsErrorBoundary`
+ * directly.
```
Ubax: Can you make this comment a little shorter?

I think it would be better to add `AppMetricsRootProps` type and add doc comments there

packages/expo-app-metrics/src/AppMetricsErrorBoundary.tsx
```diff
@@ -0,0 +1,86 @@
...
+  children: React.ReactNode;
+  /**
+   * What to render in place of the failed subtree after an error is caught. Pass an element to
+   * render it instead of the broken children, keeping the app usable.
+   *
+   * Required, but explicitly pass `null` to opt out and render nothing in place of the subtree. The
+   * boundary always renders something (the element or nothing), so a render error never propagates
+   * past it; React boundaries can't re-throw to reproduce the original crash, so capture-only
+   * boundaries aren't supported. Render-phase errors that no boundary catches are still recorded by
+   * the global `ErrorUtils` handler `expo-app-metrics` installs on import.
```
Ubax: Nit: I feel like this is a little to complicated explanation. Maybe splitting it into bullet points would help

packages/expo-app-metrics/src/AppMetricsErrorBoundary.tsx
```diff
@@ -0,0 +1,86 @@
...
+   */
+  fallback: React.ReactElement | null;
+};
+
+type State = {
+  /**
+   * Whether an error has been caught. Tracked separately from the value because a thrown value can
+   * itself be falsy (e.g. `throw null`); keying "caught" off the value would make such a throw look
+   * like a healthy state and re-render the children into an infinite loop.
```
Ubax: Nit: I am not sure if I understand this correctly. Can you write this in a bit simpler language?

packages/expo-app-metrics/src/AppMetricsErrorBoundary.tsx
```diff
@@ -0,0 +1,86 @@
...
+ * `fallback` in place of the subtree that threw.
+ *
+ * Render-phase errors don't reach `global.ErrorUtils` — React routes them through error boundaries —
+ * so a boundary is the only way to observe them with the React component stack attached. Place one
+ * around any subtree you want to keep capturing and recoverable; `AppMetricsRoot` mounts one
+ * automatically when given an `errorBoundaryFallback`.
+ *
+ * The error is reported as non-fatal: the boundary renders the fallback and the app keeps running,
+ * so it's a handled error, not a crash. Reporting runs in a `try/catch` so a failure inside
+ * `reportError` can never become a new app-facing error.
```
Ubax: Nit: is all of this information relevant to the user?

packages/expo-app-metrics/src/AppMetricsErrorBoundary.tsx
```diff
@@ -0,0 +1,86 @@
...
+  componentDidCatch(error: unknown, errorInfo: React.ErrorInfo): void {
+    // The thrown value may not be an `Error` (code can throw anything), so read its fields
+    // defensively and fall back to `String(error)` for the message.
+    const caught = error instanceof Error ? error : undefined;
+    // React's component stack has a leading newline and indentation on each frame; trim the
+    // surrounding whitespace so it reads cleanly.
+    const componentStack = errorInfo?.componentStack?.trim();
+
+    try {
+      AppMetrics.reportError({
```
Ubax, line 78: What do you think about adding an attribute which will note what is the scope of the boundary - root or standalone

packages/expo-app-metrics/src/AppMetricsErrorBoundary.tsx
```diff
@@ -0,0 +1,86 @@
...
+        message: caught?.message ?? String(error),
+        stacktrace: caught?.stack,
+        componentStack,
+        // The boundary renders the fallback and the app keeps running, so the error is handled, not
+        // fatal (mirrors how Sentry marks error-boundary captures as `handled`).
+        isFatal: false,
+      });
+    } catch {
+      // Never let a reporting failure (e.g. a native module error) escape the boundary and become a
+      // new app-facing error. Losing one report is strictly better than crashing.
```
Ubax: Can you add `console.warn` in dev?

packages/expo-app-metrics/src/AppMetricsErrorBoundary.tsx
tsapeta: Follow-up task: https://linear.app/expo/issue/ENG-22566/sdk-throttlededup-the-exception-report-path-global-errorboundary

packages/expo-app-metrics/src/AppMetricsErrorBoundary.tsx
tsapeta: Went with the second option — `state.error` now earns its place. `fallback` accepts `({ error, resetError }) => ReactNode`, and `render` passes the caught value through:

```ts
return fallback({ error: this.state.error, resetError: this.resetError });
```

So it also resolves your reset/retry suggestion in the same change. `hasError` stays as the separate "caught" flag (the `throw null` case).

packages/expo-app-metrics/src/__tests__/AppMetricsErrorBoundary.test.native.tsx
```diff
@@ -0,0 +1,190 @@
...
+  it('renders children unchanged when nothing throws', () => {
+    render(
+      <AppMetricsErrorBoundary fallback={null}>
+        <Text testID="child">healthy</Text>
+      </AppMetricsErrorBoundary>
+    );
+
+    expect(screen.getByTestId('child')).toBeVisible();
+    expect(reportError).not.toHaveBeenCalled();
+  });
```
Ubax, line 59: nit: Can you add the same test but with fallback defined and assert that its `testId` is not visible?

packages/expo-app-metrics/src/__tests__/AppMetricsErrorBoundary.test.native.tsx
```diff
@@ -0,0 +1,190 @@
...
+
+  it('renders nothing when the fallback is explicitly null', () => {
+    render(
+      <AppMetricsErrorBoundary fallback={null}>
+        <Boom />
+        <Text testID="child">healthy</Text>
+      </AppMetricsErrorBoundary>
+    );
+
+    expect(screen.queryByTestId('child')).toBeNull();
```
Ubax: Can you match snapshot here?

Just so we assert that what we expect is rendered

packages/expo-app-metrics/src/__tests__/AppMetricsErrorBoundary.test.native.tsx
```diff
@@ -0,0 +1,190 @@
...
+      expect.objectContaining({
+        source: 'errorBoundary',
+        type: 'Error',
+        message: 'render exploded',
+        isFatal: false,
+      })
+    );
+  });
+
+  it('renders a fallback render function with the error and resets on retry', () => {
```
Ubax, line 145: nit: I'm not sure if it is worth it, but maybe we can add a test which will check wether the boundary still catches the error after reset

packages/expo-app-metrics/src/__tests__/AppMetricsErrorBoundary.test.native.tsx
```diff
@@ -0,0 +1,190 @@
...
+
+  it('reports the React component stack of the subtree that threw', () => {
+    render(
+      <AppMetricsErrorBoundary fallback={null}>
+        <NestedBoom />
+      </AppMetricsErrorBoundary>
+    );
+
+    const { componentStack } = reportError.mock.calls[0][0];
+    expect(componentStack).toEqual(expect.stringContaining('Boom'));
```
Ubax, line 209: Can we also add a `expect.toMatchSnapshot` here?

Just so that we can see what is reported in componentStack

packages/expo-app-metrics/src/AppMetricsErrorBoundary.tsx
```diff
@@ -0,0 +1,109 @@
...
+        message: caught?.message ?? String(error),
+        stacktrace: caught?.stack,
+        componentStack,
+        // The boundary renders the fallback and the app keeps running, so the error is handled, not
+        // fatal (mirrors how Sentry marks error-boundary captures as `handled`).
+        isFatal: false,
+      });
+    } catch {
+      // An observability boundary must never become the source of a new error, so swallow a failure
+      // inside `reportError`. Losing one report is better than crashing.
```
Ubax: I wonder wether we should show a warning here

packages/expo-app-metrics/CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 - Add android crash reports ([#46869](https://github.com/expo/expo/pull/46869) by [@Ubax](https://github.com/Ubax))
 - Record an `expo.memory.warning` log event on iOS when the system delivers a low-memory warning, carrying the memory usage snapshot (`expo.memory.*`) taken at warning time. ([#47108](https://github.com/expo/expo/pull/47108) by [@tsapeta](https://github.com/tsapeta))
 - Add an optional `displayName` to `logEvent` ([#47289](https://github.com/expo/expo/pull/47289) by [@Ubax](https://github.com/Ubax))
+- Capture React render-phase errors via the new `AppMetricsErrorBoundary`, recorded as non-fatal `exception` log events with the React component stack (`expo.error.component_stack`). Place it around any subtree with a `fallback`, or pass `errorBoundaryFallback` to `AppMetricsRoot` to mount one at the root. ([#47341](https://github.com/expo/expo/pull/47341) by [@tsapeta](https://github.com/tsapeta))
```
Ubax:
```suggestion
- Capture React render-phase errors via `AppMetricsErrorBoundary`. ([#47341](https://github.com/expo/expo/pull/47341) by [@tsapeta](https://github.com/tsapeta))
```

packages/expo-observe/src/ObserveErrorBoundary.tsx
```diff
@@ -0,0 +1,5 @@
+export {
+  AppMetricsErrorBoundary as ObserveErrorBoundary,
+  type AppMetricsErrorBoundaryProps as ObserveErrorBoundaryProps,
+  type AppMetricsErrorBoundaryFallbackProps as ObserveErrorBoundaryFallbackProps,
```
Ubax, line 4: This may feel a little weird. The component accepts `AppMetricsErrorBoundaryFallbackProps`, but we export `ObserveErrorBoundaryFallbackProps`

packages/expo-observe/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 ### 🎉 New features
 
+- Expose `ObserveErrorBoundary`, a React error boundary that records render-phase errors as `exception` log events and renders a `fallback` in place of the crashed subtree. ([#47341](https://github.com/expo/expo/pull/47341) by [@tsapeta](https://github.com/tsapeta))
```
Ubax: Isn't it too long?

Ubax (approved):

packages/expo-app-metrics/src/__tests__/AppMetricsErrorBoundary.test.native.tsx
tsapeta: I'd rather not snapshot this one. The component stack is environment-dependent: in dev each frame carries a full Metro bundle URL (`http://<lan-ip>:8081/...&platform=ios&...`), so the string runs to tens of KB and varies by machine/host/transform flags. A snapshot would be noisy to review and would churn on unrelated env changes. The explicit `stringContaining('Boom')` + `toBe(componentStack.trim())` assertions already pin the two things we actually care about here: that the throwing component is named, and that we trim React's leading newline/indentation. Happy to add more targeted assertions if there's a specific property you want covered.

PR #47347, opened 2026-06-29
[dev-launcher][Android][Fix] Fix onUserLeaveHint NPE and lost EAS sign-in redirect

vicprz:
# Why

Two Android-only `expo-dev-launcher` bugs. The first surfaces once `androidx.activity` ≥ 1.8.0 is in the graph (e.g. via `react-native-video` ≥ 6.19); the second is independent, but was masked by the first until it was fixed.

### 1. Crash loop on `onUserLeaveHint`.
Opening the launcher crashes with an NPE when Android delivers `onUserLeaveHint` (Home press / PiP transition):

```bash
NullPointerException at ReactActivityDelegate.onUserLeaveHint
  at DevLauncherReactActivityNOPDelegate
```

The NOP delegate is built with a null host, so `mReactDelegate` is always null and every lifecycle method is a no-op, except `onUserLeaveHint`, which falls through to the base impl's `requireNonNull(mReactDelegate)`. It's latent because `onUserLeaveHint` is only delivered once `androidx.activity` ≥ 1.8.0 registers an `OnUserLeaveHintProvider`.

### 2. EAS sign-in lost on redirect.
Signing in to an EAS account leaves the user logged out on a blank launcher. `AuthActivity` only handles the `expo-dev-launcher://auth?session_secret=…` redirect in `onNewIntent`, but the browser launches it `NEW_TASK`, so it lands on a fresh `AuthActivity` that runs `onCreate` (never `onNewIntent`). The session is never parsed and `onResume` cancels it. Also reproduces when the backgrounded process is reaped mid-flow.

# How

Fix 1: add the missing no-op `override fun onUserLeaveHint() {}` to `DevLauncherReactActivityNOPDelegate`.

Fix 2: extract redirect parsing into one `consumeAuthRedirect(intent)` called from both `onCreate` and `onNewIntent`. Deliver via the process-wide `SessionService` (the launcher observes its `StateFlow`) instead of relying only on the Activity result, which doesn't reach the launcher across tasks. Teardown differs by entry point: `onCreate` re-foregrounds the existing `singleTask` launcher + `finishAndRemoveTask()`; `onNewIntent` (same task as the launcher) just `finish()`.

# Test Plan

Verified on a physical Android device.

- Crash: with `androidx.activity` ≥ 1.8.0, open launcher → press Home. Before: NPE crash loop. After: no crash.
- Auth: launcher → sign in to EAS → complete in browser → return. Before: logged out, blank launcher. After: logged in, no blank screen.

# Checklist


lukmccall (approved):
Looks good! Thanks for your contribution

Issue #47272, opened 2026-06-26
Expo 56: weird error at loading custom config plugins


yuhr:
### Minimal reproducible example

https://github.com/yuhr/repro-expo-config-plugins

### Steps to reproduce

- `pnpm install`
- `pnpm expo prebuild --platform android`

This results in the following error:

```
PluginError: exports is not defined in ES module scope
ReferenceError: exports is not defined in ES module scope
    at file:///Users/yuhr/.ghr/github.com/yuhr/repro-expo-config-plugins/plugins/withCopyDirectoryAndroid.js:3:23
    at ModuleJobSync.runSync (node:internal/modules/esm/module_job:543:37)
    at ModuleLoader.importSyncForRequire (node:internal/modules/esm/loader:346:47)
    at loadESMFromCJS (node:internal/modules/cjs/loader:1692:24)
    at Module._compile (node:internal/modules/cjs/loader:1855:5)
    at compileModule (/Users/yuhr/.ghr/github.com/yuhr/repro-expo-config-plugins/node_modules/.pnpm/@expo+require-utils@56.1.3_typescript@6.0.3/node_modules/@expo/require-utils/build/load.js:220:9)
    at evalModule (/Users/yuhr/.ghr/github.com/yuhr/repro-expo-config-plugins/node_modules/.pnpm/@expo+require-utils@56.1.3_typescript@6.0.3/node_modules/@expo/require-utils/build/load.js:310:17)
    at loadModuleSync (/Users/yuhr/.ghr/github.com/yuhr/repro-expo-config-plugins/node_modules/.pnpm/@expo+require-utils@56.1.3_typescript@6.0.3/node_modules/@expo/require-utils/build/load.js:372:10)
    at resolveConfigPluginFunctionWithInfo (/Users/yuhr/.ghr/github.com/yuhr/repro-expo-config-plugins/node_modules/.pnpm/@expo+config-plugins@56.0.9_typescript@6.0.3/node_modules/@expo/config-plugins/build/utils/plugin-resolver.js:138:49)
    at resolveConfigPluginFunction (/Users/yuhr/.ghr/github.com/yuhr/repro-expo-config-plugins/node_modules/.pnpm/@expo+config-plugins@56.0.9_typescript@6.0.3/node_modules/@expo/config-plugins/build/utils/plugin-resolver.js:126:7)
```

[...]

PR #47358, opened 2026-06-29
fix(require-utils): Throw a clear error for import.meta in CommonJS-loaded .ts/.cts modules

jiunshinn:
# Why

Related to #47272.

A `.ts`/`.cts` config plugin that uses `import.meta` (e.g. `import.meta.dirname`) fails during `expo prebuild` with a confusing `exports is not defined in ES module scope` that never mentions `import.meta` — because, as @kitten noted in the issue, these files are transpiled to CommonJS and `import.meta` is left untouched in the output.

The failure is also Node-version dependent: on Node 24 `import.meta` silently evaluates to `undefined`, while on Node 26 it throws the error above. Either way it never works, and nothing points the user at the cause.

This PR only improves the diagnostic — it does not add native ESM support for `.ts` plugins, which is a separate design decision.

# How

After a `.ts`/`.cts` file is transpiled to CommonJS, detect an `import.meta` meta-property and throw an actionable error instead of letting it reach Node:

```
Cannot use "import.meta" in "withCopyDirectoryAndroid.ts". ".ts" and ".cts" files are
loaded as CommonJS for backwards compatibility, and "import.meta" only exists in ES modules.
Replace "import.meta.dirname" with "__dirname" and "import.meta.url" with
"require('node:url').pathToFileURL(__filename).href", or rename the file to ".mts" to load
it as an ES module.
```

Detection is gated by a cheap regex and confirmed via the TypeScript AST, so `import.meta` in a string or comment is not misdetected. `.mts` files are untouched, since they load as ES modules where `import.meta` works.

# Test Plan

Added unit tests in `src/__tests__/load-test.ts` (`.ts` with `import.meta` throws; string/comment occurrence is ignored; `.mts` still loads). `build`, `typecheck`, `lint`, and `test` pass via `turbo run … --filter=@expo/require-utils`.

Happy to adjust the approach or wording, or close this if you'd prefer to handle it differently.

# Checklist

### DIFF
--- a/packages/@expo/require-utils/CHANGELOG.md
+++ b/packages/@expo/require-utils/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### 🐛 Bug fixes
 
+- Throw a clear error when `import.meta` is used in a `.ts`/`.cts` module loaded as CommonJS, instead of failing with a cryptic "exports is not defined in ES module scope" or silently evaluating to `undefined`. ([#PR](https://github.com/expo/expo/pull/PR) by [@jiunshinn](https://github.com/jiunshinn))
+
 ### 💡 Others
 
 ## 56.1.3 — 2026-05-23
--- a/packages/@expo/require-utils/CHANGELOG.md
+++ b/packages/@expo/require-utils/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### 🐛 Bug fixes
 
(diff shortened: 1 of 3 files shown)


kitten:
Thanks for submitting! However, I'll likely go for a slightly different fix here

Detecting the issue isn't necessarily comprehensive or perfectly what a user may want to see, as it leaves them in an unintended mode, as in, it throws another error that's avoidable. If `import.meta` is used, we definitively know that a user cannot be using mixed ESM/CJS and the error is redundant. This PR also doesn't handle the pure-JS case, since this isn't isolated to TypeScript and we also have the `toCommonjs` codepath.

#47441 should supersede this, which automatically switches to the ESM output when the legacy-CJS mode isn't applicable

jiunshinn:
Thanks for taking a look and for the quick follow-up! 🙏

Issue #47363, opened 2026-06-29
[expo-video] Docs say audioMixingMode defaults to auto, but iOS implementation defaults to doNotMix


andymatuschak:
### Minimal reproducible example

https://github.com/andymatuschak/expo-video-audiomixingmode-bug

### Steps to reproduce

1. On iOS, play audio in some app.
2. Open the repro app (which uses the default `audioMixingMode`)
3. Note that the muted video causes your audio to stop playing; it should not.

The [documented default](https://docs.expo.dev/versions/latest/sdk/video/#audiomixingmode) for `audioMixingMode` is `auto`, which should not interrupt music when the video player is muted. But the [iOS implementation sets the default to `doNotMix`](https://github.com/expo/expo/blob/main/packages/expo-video/ios/VideoPlayer.swift#L18), not `auto`. The Android and Web implementations set the default to `auto` as documented.

### Environment

```text
expo-env-info 2.0.14 environment info:
    System:
      OS: macOS 26.5.1
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 24.11.0 - /Users/andym/.local/state/fnm_multishells/97314_1782764679397/bin/node
      Yarn: 1.22.19 - /Users/andym/.yarn/bin/yarn
      npm: 11.6.1 - /Users/andym/.local/state/fnm_multishells/97314_1782764679397/bin/npm
      Watchman: 2026.01.12.00 - /opt/homebrew/bin/watchman
    Managers:
      CocoaPods: 1.16.2 - /opt/homebrew/bin/pod
    SDKs:
      iOS SDK:
        Platforms: DriverKit 25.5, iOS 26.5, macOS 26.5, tvOS 26.5, visionOS 26.5, watchOS 26.5
    IDEs:
      Xcode: 26.5/17F42 - /usr/bin/xcodebuild
    npmPackages:
      expo: ~56.0.12 => 56.0.12 
      react: 19.2.3 => 19.2.3 
      react-dom: 19.2.3 => 19.2.3 
      react-native: 0.85.3 => 0.85.3 
      react-native-web: ^0.21.2 => 0.21.2 
    npmGlobalPackages:
      eas-cli: 18.4.0
    Expo Workflow: managed
```

### Expo Doctor Diagnostics

```text
Running 21 checks on your project...
21/21 checks passed. No issues detected!
```

alanjhughes:
Closed by #47364

PR #47364, opened 2026-06-29
[ios][video] Change iOS audioMixingMode default to .auto (as documented)

andymatuschak:
# Why

See #47363: the [documented default](https://docs.expo.dev/versions/latest/sdk/video/#audiomixingmode) for audioMixingMode is auto, which should not interrupt music when the video player is muted. But the [iOS implementation sets the default to doNotMix](https://github.com/expo/expo/blob/main/packages/expo-video/ios/VideoPlayer.swift#L18), not auto.

The Android and Web implementations set the default to auto as documented.

# How

I just changed the iOS default to `.auto` from `.doNotFix`.

# Test Plan

1. On iOS, use some app to play audio.
2. Run [the repro app from #47363](https://github.com/andymatuschak/expo/tree/patch-2) against these changes.
3. Without this patch, the playing audio stops when the app is launched. With the patch, it does not.

# Checklist

- [X] I added a `changelog.md` entry and rebuilt the package sources according to [this short guide](https://github.com/expo/expo/blob/main/CONTRIBUTING.md#-before-submitting)
- [X] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: updated a module plugin).
- [X] Conforms with the [Documentation Writing Style Guide](https://github.com/expo/expo/blob/main/guides/Expo%20Documentation%20Writing%20Style%20Guide.md)

### DIFF
--- a/packages/expo-video/CHANGELOG.md
+++ b/packages/expo-video/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ### 🐛 Bug fixes
 
+- [iOS] Set the default `audioMixingMode` to `auto`, [as documented](https://docs.expo.dev/versions/latest/sdk/video/#audiomixingmode); was `doNotMix`. ([#47363](https://github.com/expo/expo/issues/47363) by [@andymatuschak](https://github.com/andymatuschak))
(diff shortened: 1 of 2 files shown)


brentvatne (approved):
thanks Andy!

alanjhughes (approved):

PR #47377, opened 2026-06-30
[expo-build-properties] Add android.cmakeVersion support

zoontek:
# Why

Native code built with CMake currently uses the version the Android Gradle plugin defaults to (CMake 3.22.1). That version's bundled Ninja breaks on Windows when build paths exceed the 260 character limit, which is easy to hit in `node_modules`.

See [#36274](https://github.com/expo/expo/issues/36274).

# How

Added an `android.cmakeVersion` option to `expo-build-properties`:

- Declared the `cmakeVersion` field (and its JSON schema entry) in `pluginConfig.ts`.
- Registered `android.cmakeVersion` in `android.ts` so the value is written to `gradle.properties`, following the same pattern as `compileSdkVersion`, `buildToolsVersion`, etc.

Made the property actually take effect in `expo-modules-autolinking`'s `ExpoRootProjectPlugin`:

- Added `overrideCmakeVersion()`, which reads the `android.cmakeVersion` property and, when set, applies it to `externalNativeBuild.cmake.version` on the app and every autolinked `com.android.application` / `com.android.library` subproject.

Applying it from the root project plugin (rather than having the config plugin rewrite `build.gradle`) is what lets a single property reach every autolinked native module instead of just the app module, which is where the long-path problem actually bites. The plugin logs the override so it's visible in the Gradle build output.

# Test Plan

- Added a unit test in `withBuildProperties-test.ts` asserting that `android: { cmakeVersion: '3.31.6' }` produces the `android.cmakeVersion` gradle property. Verified with `et check-packages expo-build-properties expo-modules-autolinking`.
- Built the bare-expo Android app with `android.cmakeVersion` set and confirmed the configuration-time log line appears:

  ```
  [ExpoRootProject] Overriding CMake version: 3.31.6
  ```

  and that the app and autolinked modules compile with the selected CMake version.

# Checklist


packages/expo-modules-autolinking/android/expo-gradle-plugin/expo-autolinking-plugin/src/main/kotlin/expo/modules/plugin/ExpoRootProjectPlugin.kt
```diff
@@ -18,11 +19,39 @@ class ExpoRootProjectPlugin : Plugin<Project> {
...
 
+/**
+ * Applies the `android.cmakeVersion` property (set through `expo-build-properties`) to the app and
+ * every autolinked native module that builds C++ code with CMake.
+ *
+ * The Android Gradle plugin otherwise defaults to CMake 3.22.1, whose bundled Ninja breaks on
+ * Windows with paths longer than 260 characters. Letting users pick a newer CMake version lifts
+ * that limit without having to edit each module's `build.gradle`.
+ */
+internal fun Project.overrideCmakeVersion() {
```
Kudo:
```suggestion
private fun Project.maybeOverrideCmakeVersion() {
```

- use private
- "maybe" prefix as it's not always overriding

packages/expo-modules-autolinking/android/expo-gradle-plugin/expo-autolinking-plugin/src/main/kotlin/expo/modules/plugin/ExpoRootProjectPlugin.kt
```diff
@@ -18,11 +19,39 @@ class ExpoRootProjectPlugin : Plugin<Project> {
...
   }
 }
 
+/**
+ * Applies the `android.cmakeVersion` property (set through `expo-build-properties`) to the app and
+ * every autolinked native module that builds C++ code with CMake.
+ *
+ * The Android Gradle plugin otherwise defaults to CMake 3.22.1, whose bundled Ninja breaks on
+ * Windows with paths longer than 260 characters. Letting users pick a newer CMake version lifts
+ * that limit without having to edit each module's `build.gradle`.
```
Kudo:
```suggestion
 * Maybe override the `android.externalNativeBuild.cmake.version` for all subprojects. Set via `android.cmakeVersion` in **gradle.properties**.
```

simplify the comment a little bit
- cmake 3.22.1 is for current AGP. may not be true in the future
- autolinking see gradle.properties, not expo-build-properties

packages/expo-modules-autolinking/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 ### 🎉 New features
 
+- [Android] Apply the `android.cmakeVersion` build property to the app and all autolinked native modules. ([#47377](https://github.com/expo/expo/pull/47377) by [@zoontek](https://github.com/zoontek))
```
Kudo:
```suggestion
- [Android] Apply the `android.cmakeVersion` build property to the app and all library subprojects. ([#47377](https://github.com/expo/expo/pull/47377) by [@zoontek](https://github.com/zoontek))
```

Kudo (approved):

PR #47379, opened 2026-06-30
[expo-type-information] Simple static and global values resolution

HubertBer:
# Why

Sometimes names in expo modules are provided as static / global variable instead of raw strings, and the expo-type-information parser expects them to be strings right now. So if you define
```
Events(someStaticVar)
``` 
The parser will find `someStaticVar` and later in typescript generate
```
type Events = {
	someStaticVar: () => void
}
```

But we want it to actually generate the event name in there! Especially that If the static variable is a static property so you have `Events(SomeModule.SomeVariable)` it will generate:
```
type Events = {
	SomeModule.SomeVariable: () => void
}
```

Which isn't proper typescript so it won't parse.

# How
Added a simple static / global variable storage to the sourcekitten parser. It firsts, before parsing module classes, finds all static / global variables and saves them to a `namespaces` object. This for now is only used when parsing Events, so when parsing event name we instead of just trimming `"` characters, we resolve the variable using `namespaces` object if the event name is not a string literal.

# Test Plan
- [x] tested manually in `expo-contacts`
- [x] added tests, updated snapshot

# Checklist

### DIFF
--- a/packages/expo-type-information/__tests__/TestModule.swift
+++ b/packages/expo-type-information/__tests__/TestModule.swift
@@ -3,7 +3,20 @@ import WebKit
 
 public class TestModule: Module {
   public func definition() -> ModuleDefinition {
-    Events("event1", "event2", "event3")
+    Events(
+      "event1",
+      "event2",
+      "event3",
+      globalEventName,
+      privateGlobalEventName,
+      EventNames.staticLetEvent,
+      EventNames.staticVarEvent,
+      EventStructNamespace.structEvent,
+      EventClassNamespace.classEvent,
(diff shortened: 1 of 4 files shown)


chrfalch (approved):
**nit:** getInitializerValue uses split('=', 2)[1], which drops everything after a second = — so a computed initializer or an interpolated string with a = in it truncates. slice(indexOf('=') + 1).trim() would be safer?

**Heads-up:** this conflicts with main and shares TestModule.swift + the snapshot with your other open type-info PRs, so we'll want to pick a merge order — a rebase will be needed after the first of them lands.

PR #46424, opened 2026-05-30
Fix ExpoModulesProvider lookup for dotted bundle names

shawnthye-guru:
# Why

Apps can set an iOS `CFBundleName` that is not a valid Swift module identifier (e.g. `Universal.internal`, or a product name starting with a digit like `123App`). Swift compiles the generated `ExpoModulesProvider` under the `:c99extidentifier`-sanitized module name (e.g. `Universal_internal`, `_23App`), but `AppContext.modulesProvider()` only looked up `CFBundleName.ExpoModulesProvider`.

With Expo SDK 56 precompiled modules, that lookup fails at runtime and leaves Expo native modules unregistered:

```text
[runtime not ready]: Error: Cannot find native module 'ExpoAsset'
```

Fixes #46423.

# How

- Reads `CFBundleExecutable` alongside `CFBundleName` from `Info.plist` in both the main bundle and framework fallback paths.
- Prioritizes `CFBundleExecutable` — it equals `$(PRODUCT_NAME:c99extidentifier)` and **is** the Swift module name by construction, so no runtime sanitization is needed.
- `CFBundleName` is kept as a secondary candidate for backwards compatibility.
- Removes `swiftModuleName(from:)` and its runtime sanitization logic entirely — it was redundant (since `CFBundleExecutable` already carries the correct value) and incorrect for digit-leading names (it prefixed `_` instead of replacing the first character, e.g. `_123App` instead of `_23App`).
- Simplifies `moduleProviderClassNames` to pure deduplicating mapping.
- Updates Swift Testing coverage to reflect the simpler contract.

# Test Plan

- `git diff --check`
- Verified in an Expo SDK 56 app with `CFBundleName = Universal.internal`; the app no longer throws `Cannot find native module 'ExpoAsset'` on simulator launch.

PR #47383, opened 2026-06-30
[ios][config-plugins] Sanitize CFBundleName to match PRODUCT_NAME

rolfb:
## Summary

`CFBundleName` is used as a fallback to resolve the app's `ExpoModulesProvider` class name (`AppContext.modulesProvider` on iOS), so it must be a valid identifier. `setProductName` already sanitizes `PRODUCT_NAME` for the same lookup via the shared `sanitizedName` helper, but `setName` wrote `config.name` raw into `CFBundleName` — so a name like `\"My Appæ\"` produced a candidate (`My Appæ.ExpoModulesProvider`) that can never resolve via `NSClassFromString`.

- Route `setName` through the existing `sanitizedName` helper so `CFBundleName` and `PRODUCT_NAME` stay consistent.
- `CFBundleDisplayName` (set raw by `setDisplayName`) remains the user-visible name, so the home screen is unaffected.
- Added a test mirroring the existing `setProductName` test, covering `æ`+whitespace and other non-identifier characters.

## Test plan

- [x] `jest` in `@expo/config-plugins` — `Name-test` (10/10) and `withDefaultPlugins` snapshot (3/3, no churn)
- [x] `ExpoModulesCore-Unit-Tests/AppContextTests` iOS simulator — 15/15 pass

### DIFF
--- a/packages/@expo/config-plugins/src/ios/Name.ts
+++ b/packages/@expo/config-plugins/src/ios/Name.ts
@@ -80,6 +80,7 @@ export function setName(
   return {
     ...infoPlist,
     CFBundleName: sanitizedName(name),
+    CFBundleExecutable: sanitizedName(name),
   };
 }
 
--- a/packages/@expo/config-plugins/src/ios/Name.ts
+++ b/packages/@expo/config-plugins/src/ios/Name.ts
@@ -77,9 +77,13 @@ export function setName(
     return infoPlist;
   }
 
+  // `CFBundleName` is used as a fallback to resolve the app's `ExpoModulesProvider`
+  // class name (`AppContext.modulesProvider`), so it must be a valid identifier —
+  // mirroring `setProductName`, which sanitizes `PRODUCT_NAME` for the same lookup.
+  // `CFBundleDisplayName` (set raw by `setDisplayName`) remains the user-visible name.
   return {
     ...infoPlist,
-    CFBundleName: name,
+    CFBundleName: sanitizedName(name),
   };
 }
 
--- a/packages/@expo/config-plugins/src/ios/Name.ts
+++ b/packages/@expo/config-plugins/src/ios/Name.ts
@@ -77,10 +77,6 @@ export function setName(
     return infoPlist;
   }
 
-  // `CFBundleName` is used as a fallback to resolve the app's `ExpoModulesProvider`
-  // class name (`AppContext.modulesProvider`), so it must be a valid identifier —
-  // mirroring `setProductName`, which sanitizes `PRODUCT_NAME` for the same lookup.
-  // `CFBundleDisplayName` (set raw by `setDisplayName`) remains the user-visible name.
   return {
     ...infoPlist,
     CFBundleName: sanitizedName(name),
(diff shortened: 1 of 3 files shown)


packages/@expo/config-plugins/src/ios/Name.ts
```diff
@@ -77,9 +77,13 @@ export function setName(
...
   }
 
+  // `CFBundleName` is used as a fallback to resolve the app's `ExpoModulesProvider`
+  // class name (`AppContext.modulesProvider`), so it must be a valid identifier —
+  // mirroring `setProductName`, which sanitizes `PRODUCT_NAME` for the same lookup.
+  // `CFBundleDisplayName` (set raw by `setDisplayName`) remains the user-visible name.
   return {
     ...infoPlist,
-    CFBundleName: name,
+    CFBundleName: sanitizedName(name),
```
Kudo, line 82:
```suggestion
    CFBundleName: sanitizedName(name),
    CFBundleExecutable: sanitizedName(name),
```

we may need to sanitize name for CFBundleExecutable based on my experience on repack-app.

packages/@expo/config-plugins/src/ios/__tests__/Name-test.ts
```diff
@@ -28,6 +28,21 @@ describe(setName, () => {
   it(`makes no changes to the infoPlist no name is provided`, () => {
     expect(setName({} as any, {})).toMatchObject({});
   });
+
+  it(`sanitizes CFBundleName to a valid identifier, mirroring PRODUCT_NAME`, () => {
+    // CFBundleName feeds the ExpoModulesProvider class-name lookup as a fallback
+    // for CFBundleExecutable, so it must be identifier-safe. The user-visible name
+    // is CFBundleDisplayName (set raw by setDisplayName), not CFBundleName.
```
Kudo:
```suggestion
```

drop the comment. the test case title expresses well

packages/@expo/config-plugins/src/ios/Name.ts
```diff
@@ -77,9 +77,13 @@ export function setName(
     return infoPlist;
   }
 
+  // `CFBundleName` is used as a fallback to resolve the app's `ExpoModulesProvider`
+  // class name (`AppContext.modulesProvider`), so it must be a valid identifier —
+  // mirroring `setProductName`, which sanitizes `PRODUCT_NAME` for the same lookup.
+  // `CFBundleDisplayName` (set raw by `setDisplayName`) remains the user-visible name.
```
Kudo:
```suggestion
```

i think we don't need the comments

packages/@expo/config-plugins/src/ios/Name.ts
rolfb: The CFBundleExecutable should get sanitized elsewhere, but it's probably fine to add this as well.

Kudo:
could you also help us to add an item in CHANGELOG.md? it's under **packages/@expo/config-plugins/CHANGELOG.md**

rolfb:
This PR fixes an issue where `No script URL provided. Make sure the packager is running or you have embedded a JS bundle in your application bundle.` is thrown in in development, and `Can't find native module 'ExpoAsset'` indicated by a blank screen after launch in production / release builds when apps had spaces or non-ascii letters in their name.

(just leaving this as more searchable comment)

rolfb:
Closing as it's probably solved by #46424

PR #47393, opened 2026-06-30
[dev-launcher] Debounce reloads

alanjhughes:
# Why
On iOS, reloading a dev client could leave the app in a broken state. Depending on timing it would do one of three things: spin in an endless reload loop, freeze completely, or come back up with a dev menu that could no longer be opened and the CLI commands no longer responded. This reproduced on both SDK 56 and 57.

# How
1.  DevMenuManager.reload() now ignores reloads while one is already in flight plus a short window after, which swallows the repeat burst without throttling intentional reloads.

2. Stop the reload deadlock. addressed in #47392

3. Stop the dev menu from being unable to open after a reload. During reload the incoming context's OnCreate can run before the outgoing context's OnDestroy. The unconditional setAppContext(nil) in OnDestroy then wiped the live context, leaving isReactAppRunning false so the menu refused to open. OnDestroy now clears the context only if it is still the active one.

The deadlock is a RCTHost reload race in React Native itself, where the legacy RCTNativeAnimatedTurboModule invalidation overlaps new-instance startup. Running the reload inline removes the interleaving that triggers it.

# Test Plan
New 57 project

### DIFF
--- a/packages/expo-dev-menu/ios/DevMenuManager.swift
+++ b/packages/expo-dev-menu/ios/DevMenuManager.swift
@@ -110,6 +110,9 @@ open class DevMenuManager: NSObject {
 
   private var isNavigatingHome = false
 
+  private var isReloading = false
+  private var lastReloadEventAt: Date?
+
   weak var hostDelegate: DevMenuHostDelegate?
 
   @objc
@@ -214,6 +217,8 @@ open class DevMenuManager: NSObject {
   public func setAppContext(_ appContext: AppContext?) {
     currentAppContext = appContext
     if appContext != nil {
+      isReloading = false
+      lastReloadEventAt = Date()
       isNavigatingHome = false
       isReactAppRunning = true
       // Re-run packager connection setup now that the app context (and devSettings) is available.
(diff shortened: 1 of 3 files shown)


tsapeta (approved):

packages/expo-dev-menu/ios/DevMenuManager.swift
```diff
@@ -524,6 +543,17 @@ open class DevMenuManager: NSObject {
   }
 
   func reload() {
```
Kudo, line 545: [🏗️ **design**] This debounce doesn't cover every `reload` trigger. `DevMenuPackagerConnectionHandler.sendDevCommandNotificationHandler` (the handler for Metro/CLI `reload` commands) calls `DevMenuDevOptionsDelegate.reload()` directly at `DevMenuPackagerConnectionHandler.swift:95`, never touching `DevMenuManager.reload()`. That's exactly the "CLI commands" path called out in the PR description, so a duplicate/rapid `reload` notification from Metro would still hit `AppContext.reloadAppAsync()` unguarded.

Worth routing that call site through `DevMenuManager.shared.reload()` too so all three entry points (key command, dev menu UI, packager connection) share the same guard.

packages/expo-dev-menu/ios/DevMenuManager.swift
```diff
@@ -524,6 +543,17 @@ open class DevMenuManager: NSObject {
...
+    let now = Date()
+    if isReloading || (lastReloadEventAt.map { now.timeIntervalSince($0) < 0.5 } ?? false) {
+      lastReloadEventAt = now
+      return
+    }
+    isReloading = true
```
Kudo, line 554: [💡 suggestion] `isReloading` is set before we know a reload will actually be attempted. `getDevToolsDelegate()` below can return nil (it already explicitly handles `currentAppContext == nil` and `devSettings == nil`), in which case `devToolsDelegate?.reload()` is a silent no-op — but the guard still blocks every reload attempt for the next 5 seconds until the fallback timer fires, since no new `setAppContext` call is coming to reset it early.

Could move the `isReloading = true` / timer scheduling after confirming `getDevToolsDelegate()` returns non-nil, so a reload that never starts doesn't lock out the next real one.

Kudo (approved):
left some comments from agents which looks like something

PR #47401, opened 2026-07-01
[tools] Add label author to cherry picks

alanjhughes:
# Why
When triaging what still needs cherry-picking to an SDK branch, it helps to know who put each PR on the list, not just that it's there.

# How
Added a --who flag to et csp.
I also changed the table formatting so it's doesn't take up unnecessary space

# Test Plan

Screenshot 2026-07-01 at 09 51 31

### DIFF
--- a/tools/src/commands/CheckSdkPrs.test.ts
+++ b/tools/src/commands/CheckSdkPrs.test.ts
@@ -1,0 +1,95 @@
+import assert from 'node:assert/strict';
+import { describe, it } from 'node:test';
+
+import { getLabelAuthor, renderRows } from './CheckSdkPrsCommand';
+
+describe('getLabelAuthor', () => {
+  it('returns the actor of a matching labeled event', () => {
+    const events = [
+      {
+        event: 'labeled',
+        actor: { login: 'brentvatne' },
+        label: { name: 'sdk-55' },
+        created_at: '2026-06-01T00:00:00Z',
+      },
+    ];
+    assert.equal(getLabelAuthor(events, 'sdk-55'), 'brentvatne');
+  });
+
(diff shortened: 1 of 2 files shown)


tools/src/commands/CheckSdkPrs.test.ts
```diff
@@ -0,0 +1,95 @@
...
+import { describe, it } from 'node:test';
+
+import { getLabelAuthor, renderRows } from './CheckSdkPrsCommand';
+
+describe('getLabelAuthor', () => {
+  it('returns the actor of a matching labeled event', () => {
+    const events = [
+      {
+        event: 'labeled',
+        actor: { login: 'brentvatne' },
```
tsapeta, line 11: This tells me it wasn't AI-generated 😂

tools/src/commands/CheckSdkPrsCommand.ts
```diff
@@ -383,6 +470,11 @@ export default (program: Command) => {
   A PR is "published" if all packages it modified have appeared in a publish commit.`
     )
     .option('-a, --all', 'Show all PRs grouped by status', false)
+    .option(
+      '--who',
+      'Show who applied the SDK label to each PR (one extra API request per PR)',
+      false
```
tsapeta, line 476: Why not just make it on by default? 🙂

tsapeta (approved):

tools/src/commands/CheckSdkPrsCommand.ts
alanjhughes: There's an extra api call involved so it's slightly slower. Not a big deal, in theory we could hit rate limits but it's unlikely.

PR #47404, opened 2026-07-01
[expo-blur] Fix gestures being cancelled under a BlurTargetView on Android

### LINKED ISSUE #47402: [expo-blur]  ExpoBlurTargetView.indexOfChild() returns -1 for its own child on Android → cancels all react-native-gesture-handler gestures rendered inside a BlurTargetView

https://github.com/efstathiosntonas/rngh-long-press-repro

## Summary

On Android, any [`react-native-gesture-handler`](https://github.com/software-mansion/react-native-gesture-handler) gesture (long-press, pan, tap…) rendered **inside** an `expo-blur` `<BlurTargetView>` is silently cancelled — it is recorded and then killed on its first delivered event, so it never activates. iOS is unaffected.

The root cause is in `expo-blur`: `ExpoBlurTargetView` overrides the child operations (including `indexOfChild`) and redirects them to the inner `blurTargetView`. That makes `parent.indexOfChild(child)` return `-1` for a view that is genuinely attached, violating the Android `View`/`ViewGroup` parent↔child contract.

## Root cause

Diagnosed by @j-piasecki on https://github.com/software-mansion/react-native-gesture-handler/pull/4177#issuecomment-4851572523 :

> I think this is an issue with `expo-blur`. [`ExpoBlurTargetView`](https://github.com/expo/expo/blob/63f608c6084e1d77ecc6243c480bc32c98a50f51/packages/expo-blur/android/src/main/java/expo/modules/blur/ExpoBlurTargetView.kt#L98) redirects all operations on children (including `indexOfChild` which RNGH now uses) to the `blurTargetView`.
>
> We do the traversal bottom-up in this case, so what RNGH sees is some view which has the parent of `blurTargetView`, which has the parent of `ExpoBlurTargetView`. When we ask `ExpoBlurTargetView` for the index of `blurTargetView` it's redirected to `blurTargetView` so it effectively ends up as "what is the index of blurTargetView in blurTargetView's list of children", which returns -1 since a view cannot be its own child.

Concretely: `blurTargetView.getParent() === ExpoBlurTargetView`, yet `ExpoBlurTargetView.indexOfChild(blurTargetView) === -1`, because the call is delegated to `blurTargetView.indexOfChild(blurTargetView)`.

The current `ExpoBlurTargetView` implementation is from #43595.

## Impact

`react-native-gesture-handler` 3.0.x (version 2 works fine) uses `indexOfChild` in `GestureHandlerOrchestrator.isViewAttachedUnderWrapper` (introduced in [software-mansion/react-native-gesture-handler#4177](https://github.com/software-mansion/react-native-gesture-handler/pull/4177)) to detect detached / "disappearing" views and stop delivering events to them. Because `ExpoBlurTargetView` returns `-1` for its own attached child, RNGH concludes the on-screen gesture view is detached and cancels every gesture under the blur target.

Real-world effect: apps that wrap content in a `BlurTargetView` (e.g. an app-wide blur shell — the documented "single shared target for multiple `BlurView`s" pattern) lose **all** gesture-handler gestures on Android for every screen under the target — long-press (longPress) menus, pans, swipeables, etc. — even though the views are fully visible and interactive.

## Reproduction

Minimal repro: https://github.com/efstathiosntonas/rngh-long-press-repro

1. `npm install && npx expo prebuild --clean && npx expo run:android`
2. On the **Minimal** tab, long-press a `FlashList` row → nothing happens (gesture cancelled).
3. Long-press the **control row** rendered *outside* the `FlashList` → works.
4. Delete the single `<BlurTargetView>` wrapper in `App.tsx`, refresh client → the list long-press works again.

The whole repro is `GestureHandlerRootView → BlurTargetView → FlashList (rows holding a useLongPressGesture)`. The `<BlurTargetView>` is the only required ingredient.

## Evidence

Instrumenting RNGH's attach check to log the ancestor chain at the point it decides "detached" (long-pressing a list row):

```
attach=FALSE detached='UIManagerCompatibleBlurTarget' in='ExpoBlurTargetView'
chain: RNGestureHandlerDetectorView -> ReactViewGroup -> ReactViewGroup -> ReactScrollView
  -> ReactSwipeRefreshLayout -> ClippingScrollViewDecoratorView -> ScreenContentWrapper
  -> Screen -> ScreensCoordinatorLayout -> ScreenStack -> ReactViewGroup -> ReactViewGroup
  -> RNGestureHandlerRootView -> ReactViewGroup -> ReactViewGroup
  -> UIManagerCompatibleBlurTarget -> [ExpoBlurTargetView]
```

`ExpoBlurTargetView.indexOfChild(<its UIManagerCompatibleBlurTarget child>)` returns `-1`, matching the j-piasecki's analysis.

## Expected vs actual

- **Expected:** `parent.indexOfChild(child)` returns a valid index (`>= 0`) for a child whose `getParent()` is that parent, so consumers that rely on the `View` contract (RNGH, and the Android framework itself) see the view as attached.
- **Actual:** `ExpoBlurTargetView.indexOfChild(blurTargetView)` returns `-1` for its own attached child, so gestures under the blur target are treated as detached and cancelled.

## References

- RNGH `isViewAttachedUnderWrapper` change that started using `indexOfChild`: software-mansion/react-native-gesture-handler#4177
- Current `ExpoBlurTargetView`: expo/expo#43595

### Environment

```text
- `expo-blur`: 56.0.x (Expo SDK 56); also reproduces on 55.0.x (SDK 55)
- `react-native`: 0.85.x (new architecture)
- `react-native-gesture-handler`: 3.0.2
- `react-native-screens`: 4.25.2
- Platform: **Android** (tested on a Pixel 7a); iOS unaffected
```

### Expo Doctor Diagnostics

```text
all good here
```

efstathiosntonas:
# Why

Fixes #47402.

On Android, `ExpoBlurTargetView` proxies child operations to its inner `blurTargetView`. The `indexOfChild` override did this unconditionally:

```kotlin
override fun indexOfChild(child: View?): Int = blurTargetView.indexOfChild(child)
```

`blurTargetView` is itself a real child of `ExpoBlurTargetView` (added via `super.addView` in `init`). So `ExpoBlurTargetView.indexOfChild(blurTargetView)` is delegated to `blurTargetView.indexOfChild(blurTargetView)`, which is always `-1` (a view cannot be its own child) — even though `blurTargetView.parent === ExpoBlurTargetView`.

This breaks the `View`/`ViewGroup` parent↔child contract. `react-native-gesture-handler` 3.0.x relies on that contract (v2 works fine) in `GestureHandlerOrchestrator.isViewAttachedUnderWrapper` to detect detached views (software-mansion/react-native-gesture-handler#4177). Its walk goes bottom-up, reaches `blurTargetView`, asks its parent (`ExpoBlurTargetView`) for the index, gets `-1`, and concludes the fully on-screen view is detached — so it cancels **every** gesture rendered under a `BlurTargetView` on Android (long-press, pan, swipeables, …).

# How

Give `indexOfChild` the same `child === blurTargetView` guard that the `addView`/`removeView` overrides already use: return the real index of the blur target view (`super.indexOfChild`) instead of delegating to it. Content children keep resolving through `blurTargetView`, so the proxy stays transparent.

# Test Plan

- Minimal repro (`react-native-gesture-handler` `useLongPressGesture` inside a `FlashList` inside a `BlurTargetView`): long-press was cancelled on Android before this change and works after — https://github.com/efstathiosntonas/rngh-long-press-repro
- Verified in a production app: with stock `react-native-gesture-handler` 3.0.2 (no patches), gestures under the app's app-wide blur target were dead on Android before this change and work after applying it.
- iOS is unaffected

# Notes

After this change `indexOfChild(blurTargetView) == 0` while `getChildAt(0)` returns the first *content* child (the intentionally-proxied set). This minor internal inconsistency is harmless — nothing does `getChildAt(indexOfChild(blurTargetView))`, and contract walkers only check the sign — and it matches the existing `addView`/`removeView` guard behavior.

### DIFF
--- a/packages/expo-blur/CHANGELOG.md
+++ b/packages/expo-blur/CHANGELOG.md
@@ -9,7 +9,7 @@
 ### 🐛 Bug fixes
 
 - Fix `BlurView` not applying border radius styles to the native blur layer. ([#45691](https://github.com/expo/expo/pull/45691) by [@mvincentong](https://github.com/mvincentong)) and [@behenate](https://github.com/behenate)
-- [Android] Fixed `ExpoBlurTargetView.indexOfChild()` returning `-1` for its own child, which made gesture handlers (e.g. `react-native-gesture-handler`) treat views under a `BlurTargetView` as detached and cancel their gestures. ([#47402](https://github.com/expo/expo/issues/47402) by [@efstathiosntonas](https://github.com/efstathiosntonas))
+- [Android] Fix gesture handlers treating views under a `BlurTargetView` as detached and cancelling their gestures. ([#47404](https://github.com/expo/expo/pull/47404) by [@efstathiosntonas](https://github.com/efstathiosntonas))
 
 ### 💡 Others
 
--- a/packages/expo-blur/CHANGELOG.md
+++ b/packages/expo-blur/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### 🐛 Bug fixes
 
 - Fix `BlurView` not applying border radius styles to the native blur layer. ([#45691](https://github.com/expo/expo/pull/45691) by [@mvincentong](https://github.com/mvincentong)) and [@behenate](https://github.com/behenate)
+- [Android] Fixed `ExpoBlurTargetView.indexOfChild()` returning `-1` for its own child, which made gesture handlers (e.g. `react-native-gesture-handler`) treat views under a `BlurTargetView` as detached and cancel their gestures. ([#47402](https://github.com/expo/expo/issues/47402) by [@efstathiosntonas](https://github.com/efstathiosntonas))
 
 ### 💡 Others
 
--- a/packages/expo-blur/android/src/main/java/expo/modules/blur/ExpoBlurTargetView.kt
+++ b/packages/expo-blur/android/src/main/java/expo/modules/blur/ExpoBlurTargetView.kt
@@ -95,7 +95,19 @@ class ExpoBlurTargetView(context: Context, appContext: AppContext) : ExpoView(co
 
   override fun getChildAt(index: Int): View? = blurTargetView.getChildAt(index)
 
-  override fun indexOfChild(child: View?): Int = blurTargetView.indexOfChild(child)
+  override fun indexOfChild(child: View?): Int {
+    // `blurTargetView` is our own real child (added via `super.addView` in `init`); every other
+    // child is proxied to it. Without this guard `indexOfChild(blurTargetView)` delegates to
+    // `blurTargetView.indexOfChild(blurTargetView)`, which is always -1 (a view can't be its own
+    // child) even though `blurTargetView.parent === this`. Consumers relying on the View
+    // parent/child contract — e.g. react-native-gesture-handler's attached-view walk — then treat
+    // everything under the blur target as detached and cancel their gestures. Mirror the
+    // add/remove overrides above.
+    if (child === blurTargetView) {
+      return super.indexOfChild(child)
+    }
+    return blurTargetView.indexOfChild(child)
+  }
 
   override fun onMeasure(widthMeasureSpec: Int, heightMeasureSpec: Int) {
     val width = MeasureSpec.getSize(widthMeasureSpec)


efstathiosntonas:
@behenate hi, can you please take a look into this one? It literally breaks `react-native-gesture-handler@3` when `expo-blur` is used. Thanks

behenate (approved):
I fixed the CHANGELOG entry 😄

efstathiosntonas:
@behenate thanks! Keep it up!!

efstathiosntonas:
@behenate hi, I noticed that you tagged this for sdk-56 and 57. I think it should also ported to sdk-55, many developers cannot bump to 55+ because of iOS 16.4 bump.

PR #47406, opened 2026-07-01
[docs] Rename "Bundled version" label to "Compatible version"

amandeepmittal:
# Why

Fix ENG-22382

# How

- Rename the SDK page label from `Bundled version` to `Compatible version` in `PagePackageVersion.tsx`.
- Add an accessible Radix tooltip that explains the version, appending an "included in Expo Go" line only when the library ships in Expo Go.
- Pass `platforms` from `PageHeader` into `PagePackageVersion` so the tooltip mirrors the existing "Included in Expo Go" tag.
- Update `PageHeader.test.tsx` for the new label, the conditional copy, and tooltip a11y (`role="tooltip"`, `aria-describedby`).

# Test Plan

Open an SDK reference page such as `/versions/latest/sdk/router/` and confirm the header shows "Compatible version" with an info tooltip explaining the version. Also, verify the tooltip adds the Expo Go line only for libraries included in Expo Go since `expo-router` is compatible with Expo Go.

CleanShot 2026-07-01 at 15 36 05@2x

Check a page like `/versions/latest/sdk/devclient/` and confirm the header shows "Compatible version", but without the Expo Go wording.

 
CleanShot 2026-07-01 at 15 36 22@2x

# Checklist

### DIFF
--- a/docs/ui/components/PageHeader/PagePackageVersion.tsx
+++ b/docs/ui/components/PageHeader/PagePackageVersion.tsx
@@ -1,6 +1,8 @@
 import { mergeClasses } from '@expo/styleguide';
+import { InfoCircleDuotoneIcon } from '@expo/styleguide-icons/duotone/InfoCircleDuotoneIcon';
 import { PackageIcon } from '@expo/styleguide-icons/outline/PackageIcon';
 import { useRouter } from 'next/compat/router';
+import { useState } from 'react';
 
 import { usePageApiVersion } from '~/providers/page-api-version';
 import versions from '~/public/static/constants/versions.json';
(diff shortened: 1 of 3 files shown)


docs/ui/components/PageHeader/PagePackageVersion.tsx
```diff
@@ -43,15 +50,41 @@ export function PagePackageVersion({
...
+
   return (
     <div className={mergeClasses('flex items-center gap-2', className)}>
       {versionRange && (
         <div
           data-md="skip"
           className="flex items-center justify-center gap-1.5 text-sm text-secondary">
           <PackageIcon aria-hidden="true" className="icon-sm text-icon-secondary" />
-          Bundled version:
+          Compatible version:
```
kadikraman: Maybe it would be more accurate to say "Recommended version"?

kadikraman (approved):

PR #47410, opened 2026-07-01
[cli] Convert DevTools plugin WebSocket upgrade request to fetch Request

krystofwoldrich:
# Why

Plugin WebSocket handlers received the raw Node.js `IncomingMessage` as their `request` argument, while the fetch `requestHandler` received a web-style fetch API `Request`. This inconsistency forced plugin authors to handle two different request shapes depending on whether they were serving HTTP or WebSocket traffic.

# How

Added a `convertUpgradeRequest` helper in `DevToolsPluginServerHelpers.ts` that converts the upgrade `IncomingMessage` into a fetch API `Request`, carrying over the URL and headers. WebSocket upgrade requests are always bodyless `GET`s, so no body/response wiring is needed (unlike the fetch `requestHandler`'s `convertRequest`, which requires a `ServerResponse` to wire up its abort signal). The `'connection'` handler now passes the converted `Request` to plugin handlers, and the `DevToolsPluginWebSocketHandler` type was updated accordingly.

# Test Plan

Update e2e test to reflect the new behaviour.


packages/@expo/cli/src/start/server/DevToolsPluginServerHelpers.ts
```diff
@@ -21,10 +21,27 @@ export type DevToolsPluginRequestHandler = (
  */
 export type DevToolsPluginWebSocketHandler = (
   socket: WebSocket,
-  request: IncomingMessage,
+  request: Request,
```
kitten, line 24: didn't we want to convert this to connect-style and doesn't this do the opposite? I'm a bit confused which way we're going then 😅

packages/@expo/cli/src/start/server/DevToolsPluginServerHelpers.ts
```diff
@@ -21,10 +21,27 @@ export type DevToolsPluginRequestHandler = (
...
+  const headers = new Headers();
+  const { rawHeaders } = request;
+  for (let index = 0; index < rawHeaders.length; index += 2) {
+    const name = rawHeaders[index];
+    const value = rawHeaders[index + 1];
+    if (name != null && value != null) {
+      headers.append(name, value);
+    }
+  }
+  return new Request(url.href, { method: request.method, headers });
```
kitten, line 42: Feels like the only original part is the `route` remapping, but this looks suspiciously redundant 🤔

If we're guarding `route` being the handler's activation, and assuming we're only activating said handler for this `route`, the remapping would always be redundant, and we can trust the two to be identical, which then makes the above remapping redundant too

Does this not hold?

packages/@expo/cli/src/start/server/DevToolsPluginServerHelpers.ts
krystofwoldrich: Yeah, I see, the idea to convert to connect style, assuming it will make it pretty much config file only change for libs, but that's not the case as there's no default way of exporting the middleware. Meaning there will still be some wrapper file required to configure the middleware options.

In the initial impl, I've missed that we pass IncommingMessage to the WebSocket handlers. Compared to the request handler which already receives Request.

packages/@expo/cli/src/start/server/DevToolsPluginServerHelpers.ts
krystofwoldrich: Yes this holds, the reason for it is matching the expected behavior -> request handler already gets the path stripped, websocket should too even when it's guaranteed that handler only activated for given route.

krystofwoldrich: We have a bit more universal convert in the `expo-server` but that doesn't apply here because it expects also `outgoingMessage` and we don't have that for the upgrade request.

packages/@expo/cli/src/start/server/DevToolsPluginServerHelpers.ts
kitten: gotcha, let's figure something out iteratively then 👍

kitten (approved):

packages/@expo/cli/src/start/server/DevToolsPluginServerHelpers.ts
kitten: I think re. the second part of this comment then, maybe assigning `pathname` is a bit simpler then

Issue #44144, opened 2026-03-22
[iOS][expo-ui] SwiftUI Menu/ContextMenu dismiss goes behind screen content (z-index)


FroostySnoowman:
### Minimal reproducible example

https://github.com/FroostySnoowman/expo-repro-swiftui-menu-glass

### Steps to reproduce

1. Run the app on iOS with a dev build that includes `@expo/ui` and `expo-glass-effect` (use a device/OS where liquid glass is available if the artifact only shows with real glass).
2. Open the repro screen at route `/repro-expo-ui-glass-menu` (SwiftUI `Menu` with a circular `GlassView` label — same interaction pattern as a header / toolbar menu button).
3. Tap the round glass button once to open the native menu.
4. Dismiss the menu without selecting an item (e.g. tap outside the menu).
5. Observe the UI for ~1 second after dismiss (when you try to interact with the button, it's behind screen content).

### Environment

```text
jbeal@Jacobs-MacBook-Pro expo-repro-swiftui-menu-glass % npx expo-env-info

  expo-env-info 2.0.11 environment info:
    System:
      OS: macOS 26.4
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 21.7.3 - ~/.nvm/versions/node/v21.7.3/bin/node
      Yarn: 1.22.22 - /usr/local/bin/yarn
      npm: 10.9.2 - /usr/local/bin/npm
    Managers:
      CocoaPods: 1.16.2 - /Users/jbeal/.rbenv/shims/pod
    SDKs:
      iOS SDK:
        Platforms: DriverKit 25.2, iOS 26.2, macOS 26.2, tvOS 26.2, visionOS 26.2, watchOS 26.2
    IDEs:
      Android Studio: 2025.3 AI-253.29346.138.2531.14876573
      Xcode: 26.3/17C529 - /usr/bin/xcodebuild
    npmPackages:
      expo: ~55.0.8 => 55.0.8 
      react: 19.2.0 => 19.2.0 
      react-native: 0.83.2 => 0.83.2 
    npmGlobalPackages:
      eas-cli: 16.32.0
    Expo Workflow: bare
```

### Expo Doctor Diagnostics

```text
jbeal@Jacobs-MacBook-Pro expo-repro-swiftui-menu-glass % npx expo-doctor@latest
17/17 checks passed. No issues detected!
```

FroostySnoowman:
[...]

Issue #42068, opened 2026-01-12
[expo-ui] ContextMenu: Add lifecycle callbacks for menu presentation state


gwlortscher:
## Summary
When using `@expo/ui/swift-ui` ContextMenu in a React Native app, tapping outside to dismiss the menu also triggers touch handlers on underlying React Native views. This differs from native iOS behavior where dismiss taps are consumed.

## Expected Behavior
Tapping outside an open ContextMenu should dismiss the menu without activating underlying touch handlers (consistent with Apple Music and other iOS apps).

## Current Behavior
The tap both dismisses the menu AND triggers the underlying Pressable/Touchable component.

## Requested Feature
Add lifecycle callbacks to ContextMenu:
```typescript
type ContextMenuProps = {
  activationMethod?: 'singlePress' | 'longPress';
  onMenuWillShow?: () => void;
  onMenuDidShow?: () => void;
  onMenuWillHide?: () => void;
  onMenuDidHide?: () => void;
  children: ReactNode;
};
```

These callbacks would allow developers to:
1. Track menu open/close state
2. Implement touch blocking during presentation
3. Synchronize UI state with menu visibility

## Technical Context
The underlying SwiftUI `contextMenuInteraction` delegate provides:
- `contextMenuInteractionWillPresent`
- `contextMenuInteractionDidEnd`

Exposing these to React Native would solve the touch passthrough issue.

## Root Cause Analysis
The `@expo/ui/swift-ui` ContextMenu renders via `UIHostingController`. When the native iOS context menu opens, it creates a presentation layer that properly blocks UIKit touches. However, React Native's gesture/touch system processes touches independently from UIKit's responder chain, so both:
1. The context menu receives the tap (and dismisses)
2. React Native's touch handlers receive the same tap (and activate underlying elements)

The iOS 18 fix ("SwiftUI gestures no longer errantly trigger when dismissing a menu") only applies to SwiftUI-to-SwiftUI touch conflicts, not SwiftUI-to-React Native.

## Alternative Solution (for consideration)
Implement touch blocking automatically during context menu presentation by:
1. Detecting when the system `_UIContextMenuContainerView` is visible
2. Blocking React Native touch events while active

## Environment
- @expo/ui: ^0.2.0-beta.9
- expo: ~54.0.29
- iOS 18.2
- Device: Physical device

PR #41087, opened 2025-11-18
[ios][ui] add menuActionDismissBehavior to ViewModifiers

starsky-nev:
# Why

ContextMenu's from SwiftUI currently have no way of controlling the dismiss behaviour. This is a common requirement allowing for multi-select filters etc. within a ContextMenu.

# How

Added a new scrollDismissesKeyboard prop to the ContextMenu and Button components that maps to SwiftUI's `. menuActionDismissBehavior()` modifier. The implementation follows the existing pattern in the ContextMenu and Button components and supports all available options that SwiftUI has while adhering to their naming standards.

# Test Plan

I updated the the ContextMenu example in the UI to have this as a setting to play with.

# Checklist

- [x] I added a `changelog.md` entry and rebuilt the package sources according to [this short guide](https://github.com/expo/expo/blob/main/CONTRIBUTING.md#-before-submitting)
- [x] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: updated a module plugin).
- [x] Conforms with the [Documentation Writing Style Guide](https://github.com/expo/expo/blob/main/guides/Expo%20Documentation%20Writing%20Style%20Guide.md)

PR #47413, opened 2026-07-01
[ios][expo-ui] Fix SwiftUI Menu/ContextMenu outside-tap dismiss passing through to React Native views

damselem:
# [ios][expo-ui] Fix SwiftUI Menu/ContextMenu outside-tap dismiss passing through to React Native views

## Demo

Before/after with the exact same taps. On published `@expo/ui` the outside-tap dismiss also fires the row underneath (see the log line); with this PR the tap only dismisses the menu.

https://github.com/user-attachments/assets/277f9310-f4f3-4f1d-8dd8-1dd4db710430

## Why

When an `@expo/ui/swift-ui` `Menu` (or `.contextMenu` via `ContextMenu`) is open, tapping **outside** the menu to dismiss it also fires the React Native view under the tap — e.g. it navigates into the list row you tapped. On native iOS the dismiss tap is *consumed* (tapping outside an open menu only closes it; the control underneath is not triggered — see Apple Music, Photos "Filter" menu).

This is the last gap keeping `@expo/ui` menus from matching native menu UX. Keep-open multi-select already works (`menuActionDismissBehavior`, #41087/#41109); this outside-tap pass-through is the remaining difference.

Prior reports: #42068 described this exact behavior but was auto-closed for a missing repro. Distinct from #44144 (a z-index/visual dismiss artifact).

### Root cause

When the SwiftUI `Menu` opens, iOS inserts a `_UIContextMenuContainerView` at the window level. UIKit's event routing blocks other **UIKit** controls behind it, but does **not** block React Native's touch handler (`RCTSurfaceTouchHandler` on Fabric / `RCTTouchHandler` on Paper) — it is a `UIGestureRecognizer` on the RN surface, so the dismiss tap still reaches RN's responder chain and fires the view. (Apple fixed the pure-SwiftUI-gesture case in iOS 18, but RN-hosted touch handling still leaks. Arguably the root deficiency is in React Native's touch handler — it ignores UIKit presentation exclusivity — so this guard is a compensating fix at the expo-ui level; happy to file/link an upstream RN issue if you'd like that tracked.)

## How

SwiftUI `Menu` exposes no present/dismiss callback, so we detect the system context-menu container at the window level and suspend RN touch handling exactly while the menu is interactively presented:

- New file `ios/ExpoUIMenuDismissTouchGuard.swift`: installs an **isa-scoped** swizzle of `UIWindow.didAddSubview(_:)` / `willRemoveSubview(_:)` (only `UIWindow` is affected — no overhead on ordinary view mutations). When a `_UIContextMenuContainerView` appears, it disables the React Native touch handlers found in that window, identified by a **typed predicate** shared with `ExpoUITouchHandlerHelper` (`RCTSurfaceTouchHandler` by class; Paper's `RCTTouchHandler` via cached `NSClassFromString`) — no class-name string matching for the recognizers.
- **Re-enable at dismissal start, not container removal.** The container is only removed at the *end* of the dismissal animation (up to ~1s after the user's tap); re-enabling only then would silently drop every tap in the window for that whole period (e.g. tap a menu item, then immediately tap a button → the button tap is eaten). UIKit flips the container's `isUserInteractionEnabled` to `false` the instant dismissal commits, so a self-terminating per-frame poll re-enables the handlers on that flip. The dismiss tap itself still never reaches RN: it began while the handlers were disabled, and a recognizer re-enabled mid-sequence is never handed an in-flight touch sequence.
- `ios/ExpoUIModule.swift`: one `OnCreate` call, `#if os(iOS)` (the podspec also builds tvOS, where the bug doesn't manifest). `install()` hops to the main queue — module `OnCreate` verifiably runs off the main thread.

No changes to `MenuView`/`ContextMenu`/`Host`, no JS changes, no new public API. Fixes both the tap `Menu` and the long-press `.contextMenu`. A no-op when no menu is open.

Robustness details:
- **Per-window, self-healing state**: on each container add/remove — and on every poll tick — the guard re-derives from the live hierarchy whether the window still contains an *interactive* context-menu container, and drives the handlers to match (no running counter that could desync; robust to overlapping presentations). State is keyed by window in a weak-keyed `NSMapTable`; container removal remains a backstop re-enable path.
- **Opening/dismissing latch**: the poll only honors the all-non-interactive condition after having observed an interactive tick, so an OS that adds the container non-interactive first cannot trigger a premature re-enable mid-open.
- **DEBUG tripwire**: if handlers ever re-enable only via the removal backstop, the dismissal-timing heuristic went inert (private API changed) — a dev-build log makes that fail loudly instead of silently regressing.
- **Weak references** to the disabled recognizers, so a view/window torn down mid-menu is never retained or left as a stale, permanently-disabled handler.
- **Install is idempotent** via `static let` (`dispatch_once` semantics).
- **Known limitations** (documented in the file header): disabling cancels an in-flight RN touch (a transient stuck pressed state is possible if a finger is down at menu-open); re-enable restores exactly what the guard disabled, which would override a third party's deliberate mid-presentation disable of the same recognizer; and the guard deliberately fires for *any* context-menu container (UIKit `UIContextMenuInteraction` included), since the RN pass-through exists for all of them.

The disable/re-enable-the-RN-touch-handler technique matches what [react-native-ios-context-menu fixed for its UIKit menu](https://github.com/dominicstop/react-native-ios-context-menu/pull/141); here it is driven by window-level container detection because SwiftUI `Menu` has no interaction delegate, and re-enables at dismissal start rather than dismissal end.

## Test Plan

Minimal repro (Host + SwiftUI `Menu` over a list of RN `Pressable` rows, with an on-screen "last onPress" log): https://github.com/damselem/expo-ui-menu-passthrough-repro

Validated on **iOS 26 / New Architecture (Fabric)**: on device (iPhone 17 Pro, SDK 57 / RN 0.86) via the repro app, and in a production app under an automated tap harness (simulator, SDK 56 / RN 0.85).

| Step | Before | After (this PR) |
|------|--------|-----------------|
| Open menu, tap a row to dismiss | menu closes **and** row `onPress` fires ❌ | menu closes, row `onPress` does **not** fire ✅ (0/6 leaks, automated) |
| Tap a row directly (menu closed) | fires ✅ | fires ✅ (handler correctly re-enabled — no freeze) |
| Open menu, tap a menu item | fires ✅ | fires ✅ (selection unaffected) |
| Tap a menu item, then a header button 150–300ms later | button tap silently **eaten** for the dismissal duration ❌ | lands first try ✅ (0/36 eaten, automated) |

Also verified:
- The DEBUG tripwire stayed silent across all presentations (the dismissal-start heuristic fired every time; the removal backstop was never needed).
- Multiple menus on one screen — opening, outside-tap-dismiss, and selection all behave, and each menu's per-window state stays independent (no cross-talk, no stuck handlers).
- **Keep-open multi-select** menu (`menuActionDismissBehavior('disabled')` + `Toggle`s): toggling multiple items keeps the menu open with native checkmarks, then an outside tap dismisses it without firing the row behind, and the row is tappable again afterward. This is the primary real-world case, since a keep-open menu is closed *by* an outside tap.

## Coverage / where review is welcome

The mechanism depends on two private behaviors: the `_UIContextMenuContainerView` class name (detection) and its `isUserInteractionEnabled` flip at dismissal start (re-enable timing). If detection fails on some configuration, the guard **degrades to a no-op** (the pre-existing pass-through returns); if only the timing flip changes, re-enable falls back to container removal (taps dropped for the dismissal animation, dev tripwire fires). I have **not** yet verified:
- iOS 16–18,
- iPad (where a `Menu` can present as a popover),
- old-architecture Paper (`RCTTouchHandler`).

Happy to extend testing or adjust the detection for these.

## Checklist

Notes:
- Bug fix; iOS only; no breaking changes, no new public API.
- Maintainer note: happy to adjust the detection hook if you prefer a different mechanism (e.g. a `Host`-level present/dismiss lifecycle, which would also address the original #42068 request for menu lifecycle callbacks).

### DIFF
--- a/packages/expo-ui/CHANGELOG.md
+++ b/packages/expo-ui/CHANGELOG.md
@@ -39,6 +39,7 @@
 
 ### 🐛 Bug fixes
 
+- [iOS] Fix tapping outside a SwiftUI `Menu`/`ContextMenu` to dismiss it passing the touch through to the React Native view behind it. ([#47413](https://github.com/expo/expo/pull/47413) by [@damselem](https://github.com/damselem))
 - [iOS] Fix the graphical `DatePicker` style (`community/datetime-picker` `display="inline"`) rendering undersized on mount and jumping to its correct size on the first tap. ([#47062](https://github.com/expo/expo/issues/47062) by [@etiennetalbot](https://github.com/etiennetalbot)) ([#47465](https://github.com/expo/expo/pull/47465) by [@nishan](https://github.com/intergalacticspacehighway))
 - [iOS][android] Fix `community/masked-view` doubling offsets and transforms by re-applying the user-supplied `style` to the inner mask/content `View` wrappers. ([#47067](https://github.com/expo/expo/issues/47067) by [@tsushanth](https://github.com/tsushanth))
 - [iOS] Fix the SwiftUI `TextField` crashing with "String index is out of bounds" on iOS 18 when JS writes new text into a focused field that holds an active text selection. ([#47447](https://github.com/expo/expo/pull/47447) by [@nishan](https://github.com/intergalacticspacehighway))
--- a/packages/expo-ui/ios/ExpoUIMenuDismissTouchGuard.swift
+++ b/packages/expo-ui/ios/ExpoUIMenuDismissTouchGuard.swift
@@ -1,0 +1,270 @@
+// Copyright 2025-present 650 Industries. All rights reserved.
+
+// iOS-only: ExpoUI.podspec also builds tvOS, and sibling files use `#if os(...)` for
+// platform-divergent behavior. The RN touch pass-through this guard fixes is touch-specific,
+// so the whole guard compiles out on other platforms.
+#if os(iOS)
+
+import UIKit
+import ObjectiveC.runtime
+
+/**
+ Fixes: tapping OUTSIDE an open SwiftUI `Menu` / `.contextMenu` to dismiss it also
+ fires the React Native view under the tap (the dismiss tap "passes through").
+
+ Root cause: when the menu opens, iOS inserts a `_UIContextMenuContainerView` at the
+ window level. UIKit blocks other UIKit controls behind it, but does NOT block React
+ Native's touch handler (`RCTSurfaceTouchHandler` on Fabric / `RCTTouchHandler` on
+ Paper) — it is a `UIGestureRecognizer` on the RN surface, so the tap still reaches RN.
+
+ SwiftUI `Menu` exposes no present/dismiss callback, so we detect the context-menu
+ container being added to the window and disable the RN touch handler(s) in that window
+ while it is presented, re-enabling them the moment the menu commits to dismissal.
+
+ Re-enable timing (why not just wait for the container to be removed): the
+ `_UIContextMenuContainerView` is only removed at the END of the dismissal animation —
+ up to ~1s after the user tapped. If we kept the RN handlers disabled until then, every
+ tap in the window would be silently dropped for that whole animation (e.g. tapping a
+ menu item and then immediately tapping a button would eat the button tap). Instead we
+ re-enable at the earliest SAFE moment: when the container stops being interactive.
+ UIKit sets `container.isUserInteractionEnabled = false` (and disables the container's
+ own pan recognizer) at the very start of the dismissal animation, which is exactly when
+ the menu is no longer accepting input. Re-enabling there is safe for the original
+ pass-through bug: the touch that triggered the dismissal began while our handlers were
+ disabled, and a gesture recognizer re-enabled mid-sequence is never handed a touch
+ sequence already in flight — so that dismiss tap still never reaches RN. Only brand-new
+ touches (which the user genuinely intends) get through.
+
+ Detection is scoped to `UIWindow` only (via isa-scoped method swizzling), so no
+ overhead is added to ordinary view-hierarchy mutations.
+
+ State is tracked PER-WINDOW and re-derived from the actual hierarchy on every container
+ add/remove and on every dismissal-poll tick (self-healing): whether a window's handlers
+ should be disabled is decided by live UIKit state, never by a running counter that could
+ desync. This keeps multiple windows/scenes independent and recovers automatically if an
+ add/remove event is ever missed. `willRemoveSubview` remains a backstop: even if the
+ dismissal poll never observed the interactivity flip, container removal still re-derives
+ and re-enables.
+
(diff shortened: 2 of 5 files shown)


damselem:
Pushed an important refinement (2ba0e50): the original version re-enabled the RN touch handlers only when the `_UIContextMenuContainerView` was removed — but UIKit removes it at the **end** of the dismissal animation, up to ~1s after the user's tap. That left a window after every menu dismissal where all taps in the window were silently dropped (e.g. tap a menu item, then quickly tap a button → the button tap was eaten).

UIKit flips `container.isUserInteractionEnabled` to `false` the instant dismissal commits (both the item-tap and outside-tap paths), so the guard now re-enables the handlers on that flip via a self-terminating main-queue poll, with container removal kept as the self-healing backstop. The pass-through fix is preserved by construction — the dismiss touch began while the handlers were disabled, and a recognizer re-enabled mid-sequence is never handed an in-flight touch sequence.

Re-verified on iOS 26 / RN 0.85 New Architecture: 0/36 dropped taps at pacing that previously dropped taps reliably, 0/6 pass-through regressions with the menu open, and the early re-enable observed in the unified log on every dismissal.

damselem:
Follow-up hardening in 7191c1f after self-review: the dismissal poll now re-derives "is any context-menu container in this window still interactive?" from the live hierarchy on every tick instead of watching only the container that scheduled it — a container-keyed check could re-enable the handlers during an overlapping second presentation. Also made the poll's window capture genuinely weak (matching the doc comment) and replaced the bare `0.016` with a named `dismissalPollInterval = 1.0/60.0` constant with rationale. Re-verified the full behavior matrix on iOS 26 / RN 0.85: no dropped taps after dismissal, no pass-through with the menu open.

damselem:
Added a short before/after screen recording to the PR description demonstrating the fix (the exact same taps on published `@expo/ui` vs. this PR).

Melzmr:
Amazing! That's exactly what blocked us from using expo-ui/Menu for our project. All outside clicks were propagated to other views. I am sad this thing isn't still merged into main and released.

fobos531:
@intergalacticspacehighway Do you think we could get some eyes on this? 👀

intergalacticspacehighway:
@fobos531 @damselem made a simpler version of this here https://github.com/expo/expo/pull/48463

brentvatne:
Closing as requested/superseded — Ubax landed a simpler version in https://github.com/expo/expo/pull/48463 (`SystemMenuTouchGate` in modules-core, no swizzling). Thanks for the investigation; that PR calls this one out.

PR #47414, opened 2026-07-01
[cli] add standalone submit feedback command

davidmokos:
Adds a new `submit-expo-feedback` package for submitting CLI feedback to `/v2/feedback/cli-send`, the endpoint added in https://github.com/expo/universe/pull/28476.

The goal of this CLI is for agents to have a simple way of giving us feedback for things like docs, skills, MCP server, CLIs, and any issues they encounter. It is better than providing a curl command because we can capture more data about the environment by using the CLI.

Once this is merged we can show a helpful line in all skills, docs, and in the CLI output when we detect an agent (using [agent-cli-detector](https://www.npmjs.com/package/agent-cli-detector)) that would read like:

```
If you encounter errors, misleading or outdated information, report it so Expo can be improved:
npx --yes submit-expo-feedback --category <category> <message>
Categories: skills, expo-cli, eas-cli, mcp, docs, unknown.
Only submit when you have something specific and actionable to report. Try to give the most context.
```

The package was renamed from `expo-feedback` because, per Jon Samp, "we will likely have an expo-feedback library and EAS Feedback service in the next 6-12 months. meant for end-users to send feedback to our customers".

Usage:

```sh
npx submit-expo-feedback
npx submit-expo-feedback <message>
npx --yes submit-expo-feedback --category docs <message>
npx --yes submit-expo-feedback -c mcp <message>
```

`--category` and `-c` accept `skills`, `expo-cli`, `eas-cli`, `mcp`, `docs`, or `unknown`. The category defaults to `unknown` when omitted from a non-interactive submission.

The command sends agent environment and local telemetry metadata, and includes Expo project/auth metadata when available. The `www` endpoint sends the data to PostHog and Slack channel `#cli-feedback-feed`.

Example metadata:

```json
{
  "category": "docs",
  "cli": {
    "name": "submit-expo-feedback",
    "version": "0.0.0"
  },
  "agentEnvironment": {
    "detected": true,
    "agent": {
      "id": "codex",
      "name": "Codex",
      "sessionId": "019efa08-efca-7091-a547-38ef97bc37be"
    }
  },
  "device": {
    "arch": "arm64",
    "platform": "darwin"
  },
  "node": {
    "version": "24.11.1"
  },
  "packageManager": "npm",
  "project": {
    "isExpoProject": true,
    "name": "friend-draw",
    "slug": "friend-draw",
    "sdkVersion": "56.0.0",
    "platforms": ["ios", "android"],
    "expoPackageVersion": "56.0.12",
    "reactNativePackageVersion": "0.85.3"
  },
  "user": {
    "id": "4e8c93ed-80bd-4631-92dd-5b4f366bef49",
    "username": "testuser",
    "authType": "session"
  },
  "isAuthenticated": true,
  "actorId": "4e8c93ed-80bd-4631-92dd-5b4f366bef49",
  "actorType": "person"
}
```

Tested with `pnpm --filter submit-expo-feedback test`, `pnpm turbo run depscheck lint format --filter=submit-expo-feedback --continue=dependencies-successful`, and `pnpm turbo run build typecheck test --filter=submit-expo-feedback --continue=dependencies-successful`.


packages/submit-expo-feedback/src/cli.ts
```diff
@@ -0,0 +1,431 @@
...
+  return 'https://api.expo.dev';
+}
+
+function checkNodeVersion(): void {
+  const nodeVersion = process.version.slice(1).split('.', 3).map(Number);
+  if (
+    nodeVersion[0]! < NODE_MIN[0] ||
+    (nodeVersion[0] === NODE_MIN[0] && nodeVersion[1]! < NODE_MIN[1])
+  ) {
+    console.error(
```
chrfalch: Should we throw here so that we don't just return to the callsite without stopping?

packages/submit-expo-feedback/src/cli.ts
```diff
@@ -0,0 +1,431 @@
...
+  console.log(chalk.green('Thanks for the feedback!'));
+}
+
+export async function resolveFeedbackAsync(messageParts: string[]): Promise<string> {
+  const feedback = messageParts.join(' ').trim();
+  if (feedback) {
+    return feedback;
+  }
+
+  const response = await prompts(
```
chrfalch, line 145: Should we check if we're in a CI environment or test environment here - to avoid asking for feedback without a terminal that handles it?

packages/submit-expo-feedback/src/cli.ts
```diff
@@ -0,0 +1,431 @@
...
+export async function sendFeedbackAsync({
+  feedback,
+  metadata,
+  session,
+}: {
+  feedback: string;
+  metadata: FeedbackMetadata;
+  session?: UserSession | null;
+}): Promise<void> {
+  const response = await fetch(new URL('/v2/feedback/cli-send', getExpoApiBaseUrl()).toString(), {
```
chrfalch, line 345: Should we add a timeout here?

chrfalch (changes requested):
A few nits - and a more general one: We might want to add a disclosure text somewhere to inform agent what kind of data we collect? Not sure how to do this - but we might need to consider what we post in a privacy context.

packages/submit-expo-feedback/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "typecheck": "tsc -p tsconfig.json",
+    "test": "jest",
+    "prepublishOnly": "pnpm run clean && pnpm run build:prod"
+  },
+  "devDependencies": {
+    "@expo/config": "workspace:*",
+    "@expo/package-manager": "workspace:*",
+    "@types/jest": "^29.2.1",
+    "@types/prompts": "^2.0.6",
+    "@vercel/ncc": "^0.38.3",
```
Kudo, line 39: the tool is simple enough. we could drop ncc

packages/submit-expo-feedback/src/cli.ts
```diff
@@ -0,0 +1,431 @@
...
+  if (feedback) {
+    return feedback;
+  }
+
+  const response = await prompts(
+    {
+      type: 'text',
+      name: 'feedback',
+      message: 'Share feedback with Expo',
+      validate: (value) => (value.trim() ? true : 'Feedback cannot be empty.'),
```
Kudo: the prompt is just a single question/answer. do we really need to use prompt() here? reducing deps as much as possible would be neat

packages/submit-expo-feedback/src/cli.ts
Kudo: fetch has default timeout fwiw

packages/submit-expo-feedback/src/cli.ts
```diff
@@ -0,0 +1,431 @@
...
+    };
+    return contents.auth ?? null;
+  } catch {
+    return null;
+  }
+}
+
+export function getExpoHomeDirectory(): string {
+  if (process.env.__UNSAFE_EXPO_HOME_DIRECTORY) {
+    return process.env.__UNSAFE_EXPO_HOME_DIRECTORY;
```
Kudo: the __UNSAFE_EXPO_HOME_DIRECTORY was to address the dotenv hijacking. since we dont load dotenv here. maybe we don't need this?

packages/submit-expo-feedback/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "agent-cli-detector": "^0.1.2",
+    "arg": "^5.0.2",
+    "chalk": "^4.0.0",
+    "ci-info": "^3.3.0",
+    "expo-module-scripts": "workspace:*",
+    "prompts": "^2.3.2"
+  },
+  "engines": {
+    "node": "^22.13.0 || ^24.3.0 || ^26.0.0 || >=27.0.0"
```
Kudo, line 48: we have this version here. do you think if we still need `checkNodeVersion`?
btw, the version range here is limited

packages/submit-expo-feedback/src/cli.ts
davidmokos: I'd keep prompt() and I'll extend it with a category (skills, expo-cli, eas-cli, mcp, docs, unknown) so we gather more info

packages/submit-expo-feedback/src/cli.ts
davidmokos: agents will hopefully use it like this: `npx --yes submit-expo-feedback <message>``  
``  
`but I'll test what'll happen when I don't provide the message in a non-interactive mode, just to have it handled

packages/submit-expo-feedback/package.json
davidmokos: I removed `checkNodeVersion`; this package is simple enough to attempt running outside the declared engine range. The `engines` entry remains as guidance, and we still collect the actual Node version in metadata for diagnostics.

packages/submit-expo-feedback/src/cli.ts
davidmokos: the default timeout is 300 seconds, which is too long. setting explicitely to 15s

packages/submit-expo-feedback/package.json
davidmokos: I like keeping `ncc` for consistency with our other CLIs and to keep workspace/runtime dependencies bundled. The published bundle is about 1.8 MB gzipped which feels reasonable?

davidmokos: but open to suggestions

davidmokos:
Added a disclosure before submission and a detailed data-collection section to `--help`

packages/submit-expo-feedback/package.json
krystofwoldrich: 1.8 MB sounds suspicious, I would expect hundreths of kB

davidmokos: you're absolutely right to push back :D

davidmokos: `typescript` was being bundled through `@expo/config`’s optional loader. I externalized it from `ncc`  
  
it iis now ~626 kB

Kudo: oh i misunderstood the usage of ncc. it's to prevent downloading extra deps and bundle as single executable. thanks for reducing the size. that's great!

Kudo (approved):

chrfalch (approved):

PR #46211, opened 2026-05-23
[docs][expo-build-properties] Document useHermesV1 default of true

yaminyassin:
# Why

`useHermesV1` defaults to `true` now, but the docs still say `@default false`. Hermes V1 became the default engine in SDK 56 (see the [SDK 56 changelog](https://expo.dev/changelog/sdk-56), which mentions you can opt out via `useHermesV1`).

The description is also out of date. It says you need `buildReactNativeFromSource` to turn Hermes V1 on. It's the other way around now: you need it to turn Hermes V1 off. The plugin resolves an unset value to `true` (`resolveConfigValue(...) ?? true`), and `validateConfig` throws if `useHermesV1` is `false` without `buildReactNativeFromSource`.

# How

- Updated the JSDoc for `useHermesV1` in `packages/expo-build-properties/src/pluginConfig.ts`, since the reference docs are generated from it. Changed `@default false` to `@default true` and fixed the description.
- Regenerated the `unversioned` reference data in `docs/public/static/data/unversioned/expo-build-properties.json`.

I only updated `unversioned`, following the note in `docs/README.md` about fixing docs after an SDK is released. Happy to update the `v56.0.0` data too if you'd prefer that here instead of a cherry-pick.

# Test Plan

- Checked the resolution and validation logic in `pluginConfig.ts`: unset resolves to `true`, and `false` requires `buildReactNativeFromSource`.
- The JSON change only touches the `useHermesV1` comment blocks. Worth re-running `et gdad -p expo-build-properties` to confirm it matches.

# Checklist

- [x] I added a `changelog.md` entry and rebuilt the package sources according to [this short guide](https://github.com/expo/expo/blob/main/CONTRIBUTING.md#-before-submitting)
- [x] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: updated a module plugin).
- [x] Conforms with the [Documentation Writing Style Guide](https://github.com/expo/expo/blob/main/guides/Expo%20Documentation%20Writing%20Style%20Guide.md)

PR #47415, opened 2026-07-01
[docs][expo-build-properties] Align SDK 56 `useHermesV1` docs

shubh73:
# Why

SDK 56 still has the stale generated `useHermesV1`, while SDK 57 is correct.

#46211 fixed the source JSDoc and unversioned docs, but SDK 56 was left behind.

Related:
- #43625
- #46211
- #47275

# How

Updated `docs/public/static/data/v56.0.0/expo-build-properties.json` so the SDK 56 `useHermesV1` reference entries match the SDK 57 generated docs.

# Checklist


amandeepmittal (approved):
Thanks for the fix!

Issue #44879, opened 2026-04-17
Stale linking state reused on Android relaunch when activity is rebuilt but process hasn't exited


iffa:
### Minimal reproducible example

https://github.com/iffa/expo-router-android-linking-stale-initial-data-bug

### Steps to reproduce

On Android 12+ the OS keeps the process alive after the user backs out of an
app. When the user relaunches from the launcher, Android destroys and recreates
the `MainActivity`, but the JS context (and all module-level JS variables)
**survives**. The recreated activity's intent has `action=ACTION_MAIN` and
`data=null` — no deep-link URL.

PR #42644 added the following block to `useStore()` in
`router-store.tsx`:

```tsx
  if (Platform.OS === 'android' && storeRef.current.state && storeRef.current.context === context) {
    initialState = storeRef.current.state;
  }
```

`storeRef` is a **module-level variable** — it lives for as long as the JS
context does (i.e., the entire process lifetime). `onStateChange` writes every
navigation state transition into it:

```js
onStateChange(newState) {
    storeRef.current.state = newState;   // always saved
    ...
}
```

The condition `storeRef.current.context === context` was meant to detect
"same bundle, same require context → this is an activity recreation, not a
fresh launch". In practice the reference is **always** the same object within
a surviving process, so the condition is always true.

The result: every time `useStore()` runs on Android (including on a plain
launcher relaunch), `initialState` is set to whatever navigation state the user
was in last time. That stale state — still pointing at `/explore` — is passed
directly to `NavigationContainer` as `initialState`.

Commenting the 3 lines in question out and testing again shows that no stale linking data is shown, which validates the root cause.

See reproduce repo's README for full steps, here's the key bits:

1. Open via deep link

   ```bash
   npx uri-scheme open linkingbug://explore -a
   ```

2. Back out to the launcher (press back button twice)

3. Relaunch the app from the launcher icon

4. **Bug:** app opens at `/explore` instead of the home screen

5. Back out and relaunch again — this time it correctly shows the home screen,
   because on the second relaunch there is no stale state left to restore from
   the previous cycle

### Environment

```text
expo-env-info 2.0.12 environment info:
    System:
      OS: macOS 26.4
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 24.13.0 - ~/.nvm/versions/node/v24.13.0/bin/node
      Yarn: 4.14.1 - ~/.nvm/versions/node/v24.13.0/bin/yarn
[...]

Issue #42500, opened 2026-01-24
[SDK 55][Android]  Unexpected app restart and linking error after wallpaper update


androidanimation:
### Minimal reproducible example

https://github.com/androidanimation/Expo-app-restarts-after-wallpaper-change

### Steps to reproduce

After changing the system wallpaper and returning to the app, the app restarts and reports a linking configuration issue.

Steps to reproduce
1  Run bun create expo-app --template default@next
2  Build with eas build --profile development --platform android
3  Open app on Android 13
4  Change wallpaper
5  Switch back to the app

Expected behavior
The app should resume normally.

Actual behavior
The app restarts and shows a linking error.

https://github.com/user-attachments/assets/9f3af129-7728-4088-a0e0-183e7306b6a5

### Environment

```text
expo-env-info 2.0.8 environment info:
    System:
      OS: Linux 5.10 Debian GNU/Linux 11 (bullseye) 11 (bullseye)
      Shell: 5.1.4 - /bin/bash
    Binaries:
      Node: 24.13.0 - /run/user/1000/fnm_multishells/5644_1769265571743/bin/node
      npm: 11.6.2 - /run/user/1000/fnm_multishells/5644_1769265571743/bin/npm
    SDKs:
      Android SDK:
        API Levels: 34
        Build Tools: 34.0.0
    npmPackages:
      expo: ~55.0.0-preview.4 => 55.0.0-preview.5 
      expo-router: ~55.0.0-beta.1 => 55.0.0-beta.2 
      react: 19.2.0 => 19.2.0 
      react-dom: 19.2.0 => 19.2.0 
      react-native: 0.83.1 => 0.83.1 
      react-native-web: ~0.21.0 => 0.21.2 
    Expo Workflow: bare
```

### Expo Doctor Diagnostics

```text
17/17 checks passed. No issues detected!
```

androidanimation:
Hey @Ubax @EvanBacon , I tested this again after the recent changes and I am still running into a related problem.

After changing the Android wallpaper and switching back to the app, the app ends up stuck on a blue screen. It does not crash with the same linking error anymore, but it also does not recover or resume correctly.

I see the same behavior whether NativeTabs has a custom backgroundColor or whether it is left empty . This is happening for me on SDK 55 using version 55.0.0 preview.10.

Is there any supported way to disable dynamic colors coming from the Android wallpaper for NativeTabs?

Ubax:
Hi @androidanimation 
[...]

PR #47417, opened 2026-07-01
[android][expo-router] Remove navigation state restoration across activity recreation

jakub-agent:
# Why

On Android, when the OS keeps the JS process alive but destroys and recreates `MainActivity` (e.g. relaunch from the launcher), the state-restoration block added in #42644 reopens the app at a stale route instead of the home screen. See #44879.

The block reused `storeRef.current.state` as `initialState` whenever `storeRef.current.context === context`. Within a surviving process the require context is the same object on every mount, so that condition is *always* true — it can't distinguish a genuine activity recreation from a fresh launch, and stale navigation state leaks into `NavigationContainer` as `initialState`.

# How

- Remove the Android state-reuse block in `useStore`.
- Remove the now-unused `context` field from the store ref (and its `RequireContext` import).
- Rebuild `build/` and add a CHANGELOG entry.

The linking-conflict warning change from #42644 is intentionally left in place; only the state restoration is removed.

> Note: this removes the incorrect restore. Correctly *preserving* navigation across a genuine recreation (e.g. Material You dynamic-color change, #42500) is better addressed separately — either by gating the restore on `savedInstanceState` (only non-null on a real recreation) or, on compileSdk 36+, by adding `assetsPaths` to the activity's `configChanges` so the overlay change no longer recreates the activity at all.

# Test Plan

- `packages/expo-router`: `pnpm test src/global-state` → 107 passed; `pnpm build` regenerates the committed `build/` output.
- Manual (Android): navigate to a pushed screen, background to the launcher, relaunch from the icon → now returns to the pushed/expected state driven by the intent rather than reopening a stale route.

# Checklist

- [x] Rebuilt the package sources (`build/`).
- [ ] N/A: docs.


jakub-agent:
Superseded by #47422 — same change, branch moved into expo/expo directly instead of a fork.

PR #47419, opened 2026-07-01
[config-plugins][updates] Honor ios.version/android.version in Updates helpers (sdk-54 backport)

TLenahan:
# Why

Backport of [expo#47416](https://github.com/expo/expo/pull/47416) to `sdk-54`.

The runtime-version helpers in `@expo/config-plugins` silently ignore the platform-specific version overrides `ios.version` and `android.version`. Projects that put their versions **only** in the platform keys (with no top-level `version` in `app.json`) fall into the `pkg.version` fallback in `ensureConfigHasDefaultValues` — and if `package.json` has no `version` either, the fallback resolves to `"1.0.0"`.

Every consumer downstream of these helpers then sees `"1.0.0"` (or the `package.json` version) rather than the platform-specific override the user configured — including the `appVersion` runtime version policy and `Updates.getNativeVersion`. On Android, `getNativeVersion` also had a latent bug: it used the iOS version for the `${version}` component of the returned `${version}(${versionCode})` string.

The `ios.version` and `android.version` fields are documented as taking precedence over top-level `version` — and the prebuild writers (`IOSVersion.getVersion`, `AndroidVersion.getVersionName`) already honor them. This PR aligns the `Updates` helpers with that documented contract.

# How

Three targeted changes in `packages/@expo/config-plugins/src/utils/Updates.ts`:

1. **`getAppVersion` becomes platform-aware.** Adds an optional `platform` argument. When provided and the corresponding `ios.version` / `android.version` is set, that value is returned. When absent, the previous top-level-only behavior is preserved, so existing external callers keep working unchanged.
2. **`getNativeVersion` uses per-platform helpers.** The ios branch keeps `IOSVersion.getVersion` (which already reads `ios.version`), and the android branch now uses `AndroidVersion.getVersionName` (which reads `android.version`). This also fixes the latent Android-uses-iOS-version bug.
3. **`resolveRuntimeVersionPolicyAsync` threads `platform` into `getAppVersion`.** With the appVersion policy, `ios.version`/`android.version` is now honored instead of quietly falling back.

The `Pick<Android, …>` / `Pick<IOS, …>` types in the affected signatures are widened to also read the `version` field. This is additive — existing callers passing a superset (like the full `ExpoConfig`) continue to typecheck.

# Test Plan

This PR is a clean cherry-pick of the `main`-targeting change (commit `2619c4676cf`). The starting file contents in `packages/@expo/config-plugins/src/utils/Updates.ts` were byte-identical between `main` and `sdk-54` before the fix, so the resulting code is the same on both branches. The dependency helpers (`IOSVersion.getVersion`, `IOSVersion.getBuildNumber`, `AndroidVersion.getVersionName`, `AndroidVersion.getVersionCode`) and the `ios.version` / `android.version` fields in `@expo/config-types` exist with the same signatures on this branch as on `main`.

On the `main`-targeting PR, verification included:

- `turbo test --filter=@expo/config-plugins` → **68 suites, 437 passed, 2 skipped, 0 failed**
- `turbo typecheck --filter=@expo/config-plugins` → clean
- `turbo lint --filter=@expo/config-plugins` → clean (0 warnings, 0 errors)
- `turbo depscheck --filter=@expo/config-plugins` → clean
- `turbo typecheck --filter=@expo/prebuild-config` → clean (downstream sanity check)
- `turbo typecheck --filter=@expo/cli` → clean (downstream sanity check — calls `Updates.getRuntimeVersionAsync` from server middleware)
- `turbo typecheck --filter=expo-updates` → clean (downstream sanity check — calls `Updates.resolveRuntimeVersionPolicyAsync`)

Included 9 new unit tests in `packages/@expo/config-plugins/src/utils/__tests__/Updates-test.ts` (5 for `getAppVersion`, 3 for `getNativeVersion`, and 2 for the `appVersion` runtime policy honoring `ios.version`/`android.version`).

# Checklist

### DIFF
--- a/packages/@expo/config-plugins/CHANGELOG.md
+++ b/packages/@expo/config-plugins/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### 🐛 Bug fixes
 
+- Honor `ios.version` and `android.version` in `Updates.getAppVersion`, `Updates.getNativeVersion`, and the `appVersion` runtime version policy. Previously the platform-specific overrides were ignored, so projects that used only `ios.version`/`android.version` (with no top-level `version` in `app.json`) received the `package.json` fallback (or `"1.0.0"`) wherever these helpers were consumed. `Updates.getAppVersion` gains an optional `platform` argument; passing it prefers the platform-specific override, and calls without a platform keep the previous behavior. Also fixes `Updates.getNativeVersion` on Android, which previously used the iOS version for the `${version}` component. ([#TBD](https://github.com/expo/expo/pull/TBD) by [@tlenahan](https://github.com/tlenahan))
+
 ### 💡 Others
 
 ## 54.0.4 — 2025-12-12
--- a/packages/@expo/config-plugins/src/utils/Updates.ts
+++ b/packages/@expo/config-plugins/src/utils/Updates.ts
@@ -29,24 +29,37 @@ export function getUpdateUrl(config: Pick<ExpoConfigUpdates, 'updates'>): string
   return config.updates?.url ?? null;
 }
 
-export function getAppVersion(config: Pick<ExpoConfig, 'version'>): string {
+export function getAppVersion(
+  config: Pick<ExpoConfig, 'version'> & {
+    android?: Pick<Android, 'version'>;
+    ios?: Pick<IOS, 'version'>;
+  },
+  platform?: 'android' | 'ios'
+): string {
+  if (platform === 'ios' && config.ios?.version) {
+    return config.ios.version;
+  }
+  if (platform === 'android' && config.android?.version) {
+    return config.android.version;
+  }
   return config.version ?? '1.0.0';
 }
 
 export function getNativeVersion(
   config: Pick<ExpoConfig, 'version'> & {
-    android?: Pick<Android, 'versionCode'>;
-    ios?: Pick<IOS, 'buildNumber'>;
+    android?: Pick<Android, 'version' | 'versionCode'>;
+    ios?: Pick<IOS, 'version' | 'buildNumber'>;
   },
   platform: 'android' | 'ios'
 ): string {
-  const version = IOSVersion.getVersion(config);
   switch (platform) {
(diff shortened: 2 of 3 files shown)


packages/@expo/config-plugins/CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 ### 🐛 Bug fixes
 
+- Honor `ios.version` and `android.version` in `Updates.getAppVersion`, `Updates.getNativeVersion`, and the `appVersion` runtime version policy. Previously the platform-specific overrides were ignored, so projects that used only `ios.version`/`android.version` (with no top-level `version` in `app.json`) received the `package.json` fallback (or `"1.0.0"`) wherever these helpers were consumed. `Updates.getAppVersion` gains an optional `platform` argument; passing it prefers the platform-specific override, and calls without a platform keep the previous behavior. Also fixes `Updates.getNativeVersion` on Android, which previously used the iOS version for the `${version}` component. ([#TBD](https://github.com/expo/expo/pull/TBD) by [@tlenahan](https://github.com/tlenahan))
```
Kudo, line 11:
```suggestion
- Honor `ios.version` and `android.version` in `Updates.getAppVersion`, `Updates.getNativeVersion`, and the `appVersion` runtime version policy. Previously the platform-specific overrides were ignored, so projects that used only `ios.version`/`android.version` (with no top-level `version` in `app.json`) received the `package.json` fallback (or `"1.0.0"`) wherever these helpers were consumed. `Updates.getAppVersion` gains an optional `platform` argument; passing it prefers the platform-specific override, and calls without a platform keep the previous behavior. Also fixes `Updates.getNativeVersion` on Android, which previously used the iOS version for the `${version}` component. ([#47419](https://github.com/expo/expo/pull/47419) by [@tlenahan](https://github.com/tlenahan))
```

Kudo (approved):

Kudo:
closing and will backport #47416 manually because i need to build js files

PR #47429, opened 2026-07-01
feat(expo-router): allow specifying path-specific headers

hassankhan:
# Why

Allow users to specify headers that only apply to specific paths via the `expo-router` config plugin options via a pattern similar to redirects/rewrites:

```js
{
  pageHeaders: [
    { 
      source: '/',
      headers: { 'X-Page-Rule': 'index', 'X-Powered-By': 'page-override' } 
    },
  ]
}
```

The precedence order for header application is:

- Route-level (typically declared in `expo-server` to set the `Content-Type`)
- `pageHeaders` from the config plugin
- `headers` from the config plugin

If multiple `pageHeaders` rules are found for a path, then the last-declared rule will take precedence:

```js
{
  pageHeaders: [
    { 
      source: '/',
      headers: { 'X-Page-Rule': 'index', 'X-Powered-By': 'page-override' } 
    },
    { 
      source: '/',
      headers: { 'X-Page-Rule': 'second'}  // For a request to `/`, the `X-Page-Rule` header will have the value `second`
    },
  ]
}
```

# How

- Added a new `pageHeaders` property to the `expo-router` config plugin options
- Thread the `pageHeaders` through to the export manifests for both SSG and SSR
- Make `expo-server` handle `pageHeaders` with the correct precedence order

# Test Plan

- Added unit tests to cover header merging precedence and behavior
- Added E2E tests to ensure headers are set correctly when served

# Checklist

### DIFF
--- a/apps/router-e2e/__e2e__/server-headers/workerd/config.capnp
+++ b/apps/router-e2e/__e2e__/server-headers/workerd/config.capnp
@@ -11,10 +11,9 @@ const serverConfig :Workerd.Config = (
 const server :Workerd.Worker = (
   modules = [
     (name = "worker", esModule = embed "workerd.js"),
-    (name = "_sitemap.html", text = embed "_sitemap.html"),
-    (name = "index.html", text = embed "index.html"),
-    (name = "+not-found.html", text = embed "+not-found.html"),
+    (name = "_expo/server/render.js", commonJsModule = embed "_expo/server/render.js"),
     (name = "_expo/functions/api+api.js", commonJsModule = embed "_expo/functions/api+api.js"),
+    (name = "_expo/loaders/blog.js", commonJsModule = embed "_expo/loaders/blog.js"),
     (name = "_expo/routes.json", text = embed "_expo/routes.json"),
   ],
   compatibilityDate = "2025-05-05",
--- a/docs/pages/versions/unversioned/sdk/router/index.mdx
+++ b/docs/pages/versions/unversioned/sdk/router/index.mdx
@@ -114,6 +114,12 @@ If you are using the [default](/more/create-expo/#--template) template to create
       default: 'undefined',
     },
     {
+      name: 'pageHeaders',
+      description:
+        "An array of header rules that are set on a specific page's response from the server. Each rule should have `source` and `headers`. Rules only apply to page responses, not API routes or data loaders.",
+      default: 'undefined',
+    },
+    {
       name: 'disableSynchronousScreensUpdates',
       description:
         'Disable synchronous layout updates for native screens. This can help with performance in some cases.',
--- a/packages/expo-router/plugin/src/withRouter.ts
+++ b/packages/expo-router/plugin/src/withRouter.ts
@@ -56,6 +56,13 @@ type RewriteConfig = {
   methods?: HttpMethod[];
 };
 
+type PageHeadersConfig = {
+  /** The path to match for the headers to apply. */
+  source: string;
+  /** Response headers to apply for matching paths. */
+  headers: Record<string, string | string[]>;
+};
+
 export type Props = {
   /** Production origin URL where assets in the public folder are hosted. The fetch function is polyfilled to support relative requests from this origin in production, development origin is inferred using the Expo CLI development server. */
   origin?: string | boolean;
@@ -84,6 +91,8 @@ export type Props = {
   rewrites?: RewriteConfig[];
   /** A list of headers that are set on every route response from the server */
   headers?: Record<string, string | string[]>;
+  /** A list of headers that are set on a specific path's response from the server. */
+  pageHeaders?: PageHeadersConfig[];
   /** Enable experimental server middleware support with a `+middleware.ts` file. Requires `web.output: 'server'` to be set in app config. */
   unstable_useServerMiddleware?: boolean;
   /** Enable experimental data loader support. Requires `web.output: 'static' | 'server'` to be set in app config. */
--- a/packages/expo-server/src/utils/headers.ts
+++ b/packages/expo-server/src/utils/headers.ts
@@ -15,3 +15,33 @@ export const appendHeadersRecord = (
     }
   }
 };
+
+export const mergeHeaderInputs = (
+  base: Record<string, string | string[]>,
+  update: Record<string, string | string[]>
+): Record<string, string | string[]> => {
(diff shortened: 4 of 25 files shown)


packages/expo-router/plugin/src/withRouter.ts
```diff
@@ -56,6 +56,13 @@ type RewriteConfig = {
   methods?: HttpMethod[];
 };
 
+type PageHeadersConfig = {
+  /** The path to match for the headers to apply */
```
amandeepmittal:
```suggestion
  /** The path to match for the headers to apply. */
```

Missing period.

packages/expo-router/plugin/src/withRouter.ts
```diff
@@ -56,6 +56,13 @@ type RewriteConfig = {
...
+  source: string;
+  /** Response headers to apply for matching paths */
```
amandeepmittal:
```suggestion
  /** Response headers to apply for matching paths. */
```

packages/expo-router/plugin/src/withRouter.ts
```diff
@@ -84,6 +91,8 @@ export type Props = {
   rewrites?: RewriteConfig[];
   /** A list of headers that are set on every route response from the server */
   headers?: Record<string, string | string[]>;
+  /** A list of headers that are set on a specific path's response from the server */
```
amandeepmittal:
```suggestion
  /** A list of headers that are set on a specific path's response from the server. */
```

amandeepmittal (approved):

docs/pages/versions/unversioned/sdk/router/index.mdx
```diff
@@ -113,6 +113,12 @@ If you are using the [default](/more/create-expo/#--template) template to create
         'A list of headers that are set on every route response from the server. The value can be a string or an array of strings.',
       default: 'undefined',
     },
+    {
+      name: 'pageHeaders',
+      description:
+        "An array of header rules that are set on a specific path's response from the server. Each rule should have `source` and `headers`.",
```
Ubax: Should we add an example of what `source` means?

I can think of different ways people may understand path, e.g. `/abc/index.tsx`, `/abc`, `/abc/index`.

docs/pages/versions/unversioned/sdk/router/index.mdx
Ubax, line 1: Just to check if this is on your radar, do you plan to update https://docs.expo.dev/router/web/server-headers/ in a follow-up?

apps/router-e2e/__e2e__/server-headers/workerd/config.capnp
```diff
@@ -11,10 +11,9 @@ const serverConfig :Workerd.Config = (
 const server :Workerd.Worker = (
   modules = [
     (name = "worker", esModule = embed "workerd.js"),
-    (name = "_sitemap.html", text = embed "_sitemap.html"),
-    (name = "index.html", text = embed "index.html"),
-    (name = "+not-found.html", text = embed "+not-found.html"),
+    (name = "_expo/server/render.js", commonJsModule = embed "_expo/server/render.js"),
     (name = "_expo/functions/api+api.js", commonJsModule = embed "_expo/functions/api+api.js"),
+    (name = "_expo/loaders/blog.js", commonJsModule = embed "_expo/loaders/blog.js"),
```
Ubax, line 16: Just asking, why is this changed?

Ubax (approved):

apps/router-e2e/__e2e__/server-headers/workerd/config.capnp
hassankhan: I switched the E2E project to use SSR rendering with the `unstable_useServerRendering: true` flag

docs/pages/versions/unversioned/sdk/router/index.mdx
hassankhan: Yep this will be addressed in a followup PR

docs/pages/versions/unversioned/sdk/router/index.mdx
hassankhan: My plan was to clarify it in a followup addition to https://docs.expo.dev/router/web/server-headers/

packages/expo-server/src/utils/headers.ts
```diff
@@ -15,3 +15,33 @@ export const appendHeadersRecord = (
...
+    const value = update[headerName];
+    if (value == null) {
+      continue;
+    }
+    const key = headerName.toLowerCase();
+    if (Array.isArray(value)) {
+      const existing = merged[key];
+      merged[key] =
+        existing != null
+          ? [...(Array.isArray(existing) ? existing : [existing]), ...value]
```
kitten, line 39: This is a slight behavioral quirk. We have hidden semantics right now that say that arrays always use append and single-values always use set. I do think that's somewhat intuitive and covers more cases, but it also means that if we have a single value and an array, they stand in conflict a bit.

We're basically going from "set + single value" to "append + single value + multi value"

Just noting this down since I don't think we have a perfect answer for this here

packages/expo-server/src/vendor/abstract.ts
```diff
@@ -161,11 +165,11 @@ export function createRequestHandler({
           // NOTE(@hassankhan): Relocate the request rewriting logic from here
           url.pathname = matchedPath;
           const loaderRequest = new Request(url, request);
-          return createResponseFrom('api', route, await getLoaderData(loaderRequest, route));
+          return createResponseFrom('api', route, await getLoaderData(loaderRequest, route), { headers: headersToApply });
```
kitten: question: do we think it makes sense to apply this to loaders? Loaders have runtime APIs that can alter headers, and I'm not sure there's cases where I'd want headers to apply to a loader response and a page response simultaneously, that wouldn't be either "default all" or solvable via middleware/loaders.

Just checking; not really a blocker

kitten (approved):

packages/expo-server/src/vendor/abstract.ts
hassankhan: Good question, agree that it doesn't really make sense to apply to loaders (or likely any server routes). I've narrowed down `pageHeaders` to only apply to HTML routes. The global `headers` continues to apply to HTML, API and Loader data routes though.

PR #47430, opened 2026-07-01
[cli] Support trusted dev server proxy CIDRs

kebr0m:
## Summary

Adds an unstable `EXPO_UNSTABLE_DEV_SERVER_TRUSTED_PROXY_CIDRS` option for native development server control endpoints when an Expo development server is intentionally accessed through a trusted reverse proxy.

Today, several native development endpoints rely on `isLocalSocket(req.socket)` so React Native debugger middleware, `/inspector/debug`, and trusted `/message` broadcasts only work when the TCP peer is loopback. That is a good default, but it means custom-domain development setups that terminate through a local reverse proxy can serve the bundle while still losing parts of the native dev control plane because the final peer address is the proxy instead of `127.0.0.1`.

This is intended for local development setups where the proxy address range is controlled by the developer, such as a local reverse proxy or container bridge network.

This keeps `isLocalSocket()` local-only and adds a separate trusted-dev-server socket check for the specific native dev-server surfaces that need proxy awareness:

- `/message` trusted client broadcasts
- React Native debug middleware
- `/inspector/debug` websocket gating

CIDR matching uses Node's native `BlockList`, supporting IPv4 and IPv6 CIDRs. Same-device-only surfaces that still call `isLocalSocket()` directly, such as `POST /_expo/open`, remain loopback-only.

## Tests Added

The new tests cover both the network utility boundary and the dev-server call sites that opt into trusted proxy handling:

- `isLocalSocket()` remains loopback-only.
- `isTrustedDevServerSocket()` accepts configured IPv4 and IPv6 proxy CIDRs and rejects malformed or untrusted CIDRs.
- `/message` still drops privileged client broadcasts from untrusted proxy sockets and allows them only when the proxy CIDR is trusted.
- React Native debug middleware and `/inspector/debug` websocket gating follow the same trusted-proxy behavior.

## Test Plan

Fully bootstrapped the workspace with:

```bash
pnpm install
```

Then ran the package check required by the contributing guide:

```bash
et check-packages @expo/cli
```

### DIFF
--- a/packages/@expo/cli/CHANGELOG.md
+++ b/packages/@expo/cli/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Add option to specify targets to use with inline modules ([#46698](https://github.com/expo/expo/pull/46698) by [@HubertBer](https://github.com/HubertBer))
 - Support Bundler-managed CocoaPods installations ([#43605](https://github.com/expo/expo/pull/43605) by [@tiwari91](https://github.com/tiwari91), [@kitten](https://github.com/kitten))
 - Support Device Hub as Simulator replacement for Xcode 27+ ([#46757](https://github.com/expo/expo/pull/46757) by [@byCedric](https://github.com/byCedric), [@GersonRocha9](https://github.com/GersonRocha9))
+- Support trusted reverse proxy CIDRs for native development server control endpoints. ([#47430](https://github.com/expo/expo/pull/47430) by [@kebr0m](https://github.com/kebr0m))
 
 ### 🐛 Bug fixes
 
--- a/packages/@expo/cli/src/start/server/metro/debugging/__tests__/createDebugMiddleware.test.ts
+++ b/packages/@expo/cli/src/start/server/metro/debugging/__tests__/createDebugMiddleware.test.ts
@@ -1,0 +1,137 @@
+import type { NextHandleFunction } from 'connect';
+import { IncomingMessage, ServerResponse } from 'node:http';
+import { Socket } from 'node:net';
+import { WebSocketServer } from 'ws';
+
+import type { SocketRemoteFamily } from '../../../../../utils/net';
+import { createDebugMiddleware } from '../createDebugMiddleware';
+
+const devMiddleware = jest.fn<ReturnType<NextHandleFunction>, Parameters<NextHandleFunction>>(
+  (_req, _res, next) => next()
+);
+let debuggerWebsocketEndpoint: WebSocketServer;
+
+jest.mock('@react-native/dev-middleware', () => ({
+  createDevMiddleware: jest.fn(() => {
+    debuggerWebsocketEndpoint = new WebSocketServer({ noServer: true });
+
+    return {
+      middleware: devMiddleware,
+      websocketEndpoints: {
+        '/inspector/debug': debuggerWebsocketEndpoint,
+      },
+    };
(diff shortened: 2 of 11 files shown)


kitten:
We assume that proxies locally should rewrite `Host` and filter or reset `Origin`. Since that's a process entirely controlled by the proxy, it's not too unreasonable to stay within the trust-boundary by expecting the proxy to rewrite/alter the request. That's also how our tunnel/proxy code works.

kebr0m:
So I tested the Host/Origin approach in our topology to make sure we were not missing something. Our dev path is:

iPhone -> Cloudflare -> Envoy Gateway in kind -> host bridge IP -> Metro

I temporarily updated the Envoy Gateway `HTTPRoute` to:

- rewrite `Host` to `localhost`
- remove `Origin`

The route was accepted, and a separate echo backend confirmed Envoy was applying those changes correctly:

- `host: localhost`
- `origin: null`
- `remoteAddress: <bridge-network-address>`

However, with the real Metro backend, `https://app.example.test/json` still returned the Expo Router web 404 page instead of the React Native debugger middleware response.

So the issue appears to be that `createDebugMiddleware`/related native dev control endpoints are gated on the actual socket peer via `isLocalSocket(req.socket)`. In our case, even with normalized Host/Origin, Metro still sees the TCP peer as the Envoy/kind bridge address, not loopback.

That is why I proposed the explicit trusted proxy CIDR option: it keeps the existing local-socket behavior as the default, but gives development setups with a trusted reverse proxy a way to opt into treating specific proxy socket peers as trusted.

Open to adjusting the shape/API, or taking a different approach, if there is a better fit for Expo’s conventions.

kitten:
That's the other check (`isLocalSocket`), yes. Select endpoints will apply this to ensure that we're not exposing the CDP's (and similar) capabilties, which can include sending arbitrary instructions / JS execution, to a client when the user cannot reasonably expect this to be allowed.

In our tunnel case, we run the proxy locally to circumvent this, not as a remote machine (since realistically this is an all or nothing change). We could disable this in some circumstances but it'd have to be extremely limited. As in, the CIDR feels a bit misleading since realistically you're after proxying CDP purposefully so the CIDR is redundant (you're basically disabling the check on the affected ingress anyway)

As such, if we were to add an env flag for this it'd have to be blocked in `@expo/env` from being read and be heavily restricted. Either way though, Id have to check internally and with the RN team, since this wasn't meant to be exposed in this way by the CLI, so it's a bit security sensitive to introduce a flag to re-disable the check

kebr0m:
Got it, that makes sense. I see why the CIDR framing is misleading here if the practical effect is enabling non-local access to CDP/debug-control endpoints.

I’ll hold off until you’ve had a chance to check with the RN team.

kitten:
We're separately now working on changes to make proxies work reliably for dev-builds, which had a few caveats depending on what the entry was set to, which will be mitigated with a series of changes for the next SDK release.

Until then though, the change here is/has become obsolete, and a general allowlist here is an inelegant and almost impossible to document workaround, which is workflow-specific and not a generic solution.

In other words, by the XY-problem, it's a personal fix to a problem that's in turn caused by a root cause / underlying problem that's different, so I'll close this for now.

Still happy to accept other issues for the proxying, but as said, things are changing on `main` to accommodate more Bonjour/proxying cases here

Issue #47449, opened 2026-07-02
[iOS][SDK 57] EXC_BAD_ACCESS in jsi::Pointer::~Pointer — JavaScriptPromise's JSI values destroyed off the JS thread / after reloadAsync teardown


jmalmo:
### Minimal reproducible example

Single-file repro on a fresh SDK 57 project (crash is a race, so it triggers within a few taps, not deterministically on the first):

```bash
npx create-expo-app@latest jsi-teardown-repro --template blank-typescript
cd jsi-teardown-repro
npx expo install expo-updates expo-file-system
# configure expo-updates (any runtimeVersion/channel), then build a RELEASE build to a device:
eas build --platform ios --profile preview   # or npx expo run:ios --configuration Release
```

```tsx
// App.tsx
import * as FileSystem from "expo-file-system/legacy";
import * as Updates from "expo-updates";
import { Button, View } from "react-native";

// Kick off a burst of expo async-module work, then reload while it is in flight.
function reloadWithInFlightWork() {
  for (let i = 0; i < 30; i++) {
    // On SDK 57 global fetch is expo/fetch, so this exercises NativeResponse.
    void fetch(`https://picsum.photos/1200?r=${Math.random()}`)
      .then((r) => r.arrayBuffer())
      .catch(() => {});
    void FileSystem.createDownloadResumable(
      `https://picsum.photos/1200?d=${Math.random()}`,
      `${FileSystem.cacheDirectory}repro-${i}.jpg`
    )
      .downloadAsync()
      .catch(() => {});
  }
  setTimeout(() => void Updates.reloadAsync(), 150);
}

export default function App() {
  return (
    <View style={{ flex: 1, justifyContent: "center" }}>
      <Button title="Reload with in-flight work" onPress={reloadWithInFlightWork} />
    </View>
  );
}
```

Tap the button repeatedly. Within a few attempts the app dies with `EXC_BAD_ACCESS` in `facebook::jsi::Pointer::~Pointer`.

### Summary

`JavaScriptPromise` (expo-modules-jsi) stores its `jsi::Object`s inline in a `~Copyable` struct: the promise object plus the resolver/rejecter `JavaScriptValue.Ref`s, and the settled result value inside its `DeferredPromise` actor. Every expo-modules async function captures the promise into Swift closures (`AsyncFunctionDefinition.build` captures `[promise]`) that end up retained by URLSession delegates and dispatch queues.

[...]

PR #47450, opened 2026-07-02
[ios][modules-jsi] Release JavaScriptPromise's JSI values on the JS thread

### LINKED ISSUE #47454: [iOS][SDK 57] EXC_BAD_ACCESS in jsi::Pointer::~Pointer — JavaScriptPromise's JSI values destroyed off the JS thread / after reloadAsync teardown

**Repository: https://github.com/jmalmo/jsi-teardown-repro**

Fresh `create-expo-app` (blank-typescript, SDK 57) plus `npx expo install expo-updates expo-file-system` and a single-button `App.tsx`. The crash is a race, so it triggers within a few taps rather than deterministically on the first:

```tsx
// App.tsx — full file in the repo
function reloadWithInFlightWork() {
  for (let i = 0; i < 30; i++) {
    // On SDK 57 global fetch is expo/fetch, so this exercises NativeResponse.
    void fetch(`https://picsum.photos/1200?r=${Math.random()}`)
      .then((r) => r.arrayBuffer())
      .catch(() => {});
    void FileSystem.createDownloadResumable(
      `https://picsum.photos/1200?d=${Math.random()}`,
      `${FileSystem.cacheDirectory}repro-${i}.jpg`
    )
      .downloadAsync()
      .catch(() => {});
  }
  setTimeout(() => void Updates.reloadAsync(), 150);
}
```

> Note: this issue was previously filed as #47449 and auto-closed for a missing repro link — this filing adds the repository above. Proposed fix: #47450.

### Summary

`JavaScriptPromise` (expo-modules-jsi) stores its `jsi::Object`s inline in a `~Copyable` struct: the promise object plus the resolver/rejecter `JavaScriptValue.Ref`s, and the settled result value inside its `DeferredPromise` actor. Every expo-modules async function captures the promise into Swift closures (`AsyncFunctionDefinition.build` captures `[promise]`) that end up retained by URLSession delegates and dispatch queues.

Whichever thread drops the **last reference** to such a closure destroys those JSI values **on that thread** — and after `Updates.reloadAsync()` tears down Hermes, **against freed memory**. `resolve()`/`reject()` correctly hop to the JS thread via `runtime.schedule`, but destruction has no such path.

We are seeing this in production (app on SDK 57 / expo-modules-core 57.0.1 / expo-modules-jsi 57.0.0) with two distinct entry points that share the same fatal tail:

Crash 1 — expo-file-system legacy download completing on the main queue after a reload:

```
at __92-[__NSCFURLSessionDelegateWrapper downloadTask:didFinishDownloadingToURL:completionHandler:]_block_invoke
at -[EXSessionTaskDispatcher URLSession:downloadTask:didFinishDownloadingToURL:] (EXSessionTaskDispatcher.m:48)
at _Block_release
...
at closure in AsyncFunctionDefinition.call
at (anonymous namespace)::destroyGenericBox
at JavaScriptPromise (<compiler-generated>)
at facebook::jsi::Object::~Object (jsi.h:1154)
at facebook::jsi::Object::~Object (jsi.h:1154)
at facebook::jsi::Pointer::~Pointer (jsi.h:833)
```

Crash 2 — `expo/fetch`'s `NativeResponse.state.didSet` clearing its listener array on a background dispatch queue (`stateChangeOnceListeners.removeAll` releases listener closures that captured the promise):

```
at closure in NativeResponse.state.didset (NativeResponse.swift:19)
at RangeReplaceableCollection.removeAll
...
at closure in AsyncFunctionDefinition.call
at (anonymous namespace)::destroyGenericBox
at JavaScriptPromise (<compiler-generated>)
at facebook::jsi::Object::~Object (jsi.h:1154)
at facebook::jsi::Object::~Object (jsi.h:1154)
at facebook::jsi::Pointer::~Pointer (jsi.h:833)
```

In both production events the crash landed seconds after the app downloaded a fresh OTA update and called `Updates.reloadAsync()` while a fetch POST and several image downloads were mid-flight. We have also observed a steady-state occurrence without any reload (off-thread release racing Hermes GC), which matches JSI's documented single-threaded ownership rules.

Related: #47051 was fixed in 57.0.0 (#47080/#47098) by tying `AppContext`-cached JSI objects to the runtime lifetime — that fix is present in our build and does **not** cover promises captured in module/URLSession closures, which never go through the AppContext cache.

### Steps to reproduce

1. Clone https://github.com/jmalmo/jsi-teardown-repro, `npm install`.
2. Set your own EAS `projectId`/`updates.url` in `app.json` (any channel — `reloadAsync` just needs expo-updates enabled), then build Release to a device: `npx expo run:ios --configuration Release` (or an EAS preview build).
3. Tap "Reload with in-flight work" a few times.
4. Observe `EXC_BAD_ACCESS` / `SIGSEGV` in `facebook::jsi::Pointer::~Pointer` with the `AsyncFunctionDefinition.call` → `destroyGenericBox` → `JavaScriptPromise` release chain.

A fix proposal is open in #47450: move the promise's JSI-owning members behind a reference-counted `State` whose deinit releases them on the JS thread (or parks them when the runtime is already gone).

### Environment

```
  System:
    OS: macOS 15.6.1
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.13.1 - /usr/local/bin/node
    npm: 11.1.0 - /usr/local/bin/npm
  Managers:
  SDKs:
    iOS SDK:
  IDEs:
    Xcode: 26.3/17C529 - /usr/bin/xcodebuild
  npmGlobalPackages:
    eas-cli: 16.32.0
  Expo Workflow: bare
```

Affected packages: `expo-modules-jsi@57.0.0`, `expo-modules-core@57.0.1`, `expo@57.0.1`, `expo-file-system@57.0.0`. Devices observed: iPhone14,7 (iOS 26.5) and iPhone18,1 (iOS 26.5.1), production App Store build.

### Expo Doctor Diagnostics

```
```

jmalmo:
# Why

Fixes #47454 (repro repository: https://github.com/jmalmo/jsi-teardown-repro; originally filed as #47449) — production `EXC_BAD_ACCESS` in `facebook::jsi::Pointer::~Pointer`.

`JavaScriptPromise` stores its JSI values inline: the promise `JavaScriptObject`, the resolver/rejecter `JavaScriptValue.Ref`s, and (once settled) the result value inside its `DeferredPromise` actor. Async-function promises get captured into Swift closures (`AsyncFunctionDefinition` captures `[promise]`) that are retained by URLSession delegates and dispatch queues. `resolve()`/`reject()` hop to the JS thread via `runtime.schedule`, but **destruction doesn't**: whichever thread drops the last closure reference destroys the `jsi::Object`s on that thread — and after `Updates.reloadAsync()` tears down the runtime, against freed memory. #47080/#47098 fixed this for AppContext-cached objects; closure-captured promises never go through that cache.

# How

- `JavaScriptPromise`'s JSI-owning members (object ref, resolver/rejecter refs, and the `DeferredPromise` actor) move behind a single reference-counted `State` class, so their destruction thread is decided in exactly one place: `State.deinit`.
- `State.deinit`: on the JS thread (or without async scheduling) the members release inline right after it, unchanged. Off the JS thread with a live runtime, the refs are parked in a new `JSIReleasePool` and a drain is scheduled — the scheduled block captures **no JSI values** (only a weak runtime), so the scheduler dropping it un-run during teardown is harmless. When the weak runtime is already gone, the entry is parked forever: releasing it would dereference freed runtime memory, and the cost is a few dozen bytes per promise abandoned mid-teardown.
- `JSIReleasePool.drain(on:)` matches entries by weak identity (`===`), never by address, so runtime address reuse across reloads cannot release another runtime's husks.
- `resolve`/`reject` now capture `[state]`, so a settle closure dropped un-run at teardown also funnels into the safe deinit.
- `asValue()`/`setUpCallbacks()` precondition on an empty object ref (unreachable today) instead of silently producing `undefined`, and use the existing `JavaScriptRef.withValue` for non-consuming access.

No public API changes.

# Test Plan

- Full release build of the package (SPM via `build-xcframework.sh`) for `iphoneos` and `iphonesimulator` (arm64 + x86_64) against an SDK 57 host app's Pods.
- Shipped as a source patch on expo-modules-jsi@57.0.0 in our production app to validate against the crash in #47449 (repro: burst of `fetch`/`expo-file-system` downloads, then `reloadAsync()`; see the issue for stacks and a single-file MRE).
- Existing `JavaScriptPromiseTests` semantics preserved: deferred create/resolve/reject ordering, wrapping already-settled objects, `await()` (now backed by the state-owned actor), `isDeferred`.
- swift-format (Xcode toolchain) run on the changed file; changelog entry added.

# Checklist

### DIFF
--- a/packages/expo-modules-jsi/CHANGELOG.md
+++ b/packages/expo-modules-jsi/CHANGELOG.md
@@ -17,6 +17,8 @@
 
 ### 🐛 Bug fixes
 
+- [iOS] Fix `EXC_BAD_ACCESS` in `jsi::Pointer::~Pointer` when the last reference to a `JavaScriptPromise` is dropped off the JS thread (URLSession completion handlers, dispatch queues) or after `reloadAsync()` destroyed the runtime. The promise's JSI-owning members now live behind a reference-counted state whose deinit releases them on the JS thread, or parks them when the runtime is already gone. ([#47454](https://github.com/expo/expo/issues/47454)) ([#47450](https://github.com/expo/expo/pull/47450) by [@jmalmo](https://github.com/jmalmo))
+
 - [iOS] Fixed `Build ExpoModulesJSI xcframework` build phase failing on Xcode 26 because the nested SwiftPM build ignored `-derivedDataPath` and wrote products outside the expected location. ([#46326](https://github.com/expo/expo/issues/46326) by [@Kurogoma4D](https://github.com/Kurogoma4D))
 - [iOS] Fixed the xcframework build failing with a `sed` error when building in an environment that uses GNU `sed` instead of BSD `sed` (e.g. a Nix shell). ([#46389](https://github.com/expo/expo/pull/46389) by [@niteshbalusu11](https://github.com/niteshbalusu11))
 - [iOS] Propagate `JavaScriptPromise` setup failures instead of trapping the app. ([#46106](https://github.com/expo/expo/issues/46106) by [@qutrek](https://github.com/qutrek)) ([#46145](https://github.com/expo/expo/pull/46145) by [@mvincentong](https://github.com/mvincentong))
--- a/packages/expo-modules-jsi/apple/Sources/ExpoModulesJSI/Runtime/Values/JavaScriptPromise.swift
+++ b/packages/expo-modules-jsi/apple/Sources/ExpoModulesJSI/Runtime/Values/JavaScriptPromise.swift
@@ -1,4 +1,5 @@
 internal import ExpoModulesJSI_Cxx
+import Foundation
 internal import jsi
 
 /// A Swift representation of a JavaScript Promise.
@@ -11,20 +12,20 @@ public struct JavaScriptPromise: JavaScriptType, ~Copyable {
   private typealias PromiseContinuation = CheckedContinuation<JavaScriptValue.Ref, any Error>
 
   private weak let runtime: JavaScriptRuntime?
-  private var object: JavaScriptObject
-  private let deferredPromise = DeferredPromise()
 
-  // Create refs for resolve and reject functions.
-  // They will be set in the Promise setup function.
-  private let resolveFunction = JavaScriptValue.Ref()
-  private let rejectFunction = JavaScriptValue.Ref()
+  // All JSI-owning members live behind `state`, whose deinit releases them on the JS thread.
+  // Promises get captured by arbitrary native closures (URLSession completion handlers, dispatch
+  // queues) whose last release can happen on any thread — and, after `reloadAsync()`, after the
+  // runtime is gone. Destroying `jsi::Object`s there crashes in `jsi::Pointer::~Pointer`.
+  private let state: State
 
   /// Initializes a promise from the existing object. The promise may already be settled.
   /// It cannot be resolved/rejected from the outside, i.e. `resolve` and `reject` functions are no-op.
   @JavaScriptActor
   public init(_ runtime: JavaScriptRuntime, _ object: consuming JavaScriptObject) throws {
     self.runtime = runtime
-    self.object = object
+    self.state = State(runtime: runtime)
+    state.object.reset(object)
     try setUpCallbacks()
   }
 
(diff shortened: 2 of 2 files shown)


jmalmo:
One known residual worth calling out for review: the settle Task created in `setUpCallbacks` retains the `DeferredPromise` actor until it completes, so if the promise's owning box dies concurrently with settlement, the actor's last release (and its stored settled payload) can land on a concurrency-pool thread — a sub-ms window that exists today for actor-held JSI values generally. This PR confines the off-thread destruction problem to that window; closing it fully likely needs an executor-level decision about where `DeferredPromise` (and actor-held JSI values in general) deallocate, which felt like a maintainer call rather than something to bundle here.

bayasdev:
This fixes iOS crashing when Updates.reloadAsync is called on my app

tsapeta:
Thank you for the thorough analysis and this well-considered fix! The `State`-class approach is good, but we ended up addressing the same root cause via a different mechanism in #47521, released in `expo@57.0.6`: promise JSI values are now owned by a `LongLivedState` registered with the runtime's `LongLivedObjectCollection`, so the teardown sweep releases them on the JS thread before the runtime is destroyed, and `deinit` schedules a hop back to the JS thread when the wrapper is dropped off-thread. Closing in favor of that, but the investigation here was very helpful.

Issue #47372, opened 2026-06-30
[iOS][SDK 56] CodedError code is not forwarded to JavaScript on iOS


fknop:
### Minimal reproducible example

https://github.com/fknop/missing-error-code-ios

### Steps to reproduce

I've implemented a basic native module in ios & android. This module throws a coded exception in an async function.
Run the app with `npm run android` and `npm run ios`.
When pressing the button on the home screen, Android should log `ERR_INVALID_INPUT` in the console. ios  does not log anything. The error code is missing

### Environment

```text
expo-env-info 2.0.14 environment info:
    System:
      OS: macOS 26.3.1
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 24.14.1 - /Users/fknop/.local/share/mise/installs/node/24.14.1/bin/node
      npm: 11.11.0 - /Users/fknop/.local/share/mise/installs/node/24.14.1/bin/npm
      Watchman: 2026.05.04.00 - /opt/homebrew/bin/watchman
    Managers:
      CocoaPods: 1.16.2 - /opt/homebrew/bin/pod
    SDKs:
      iOS SDK:
        Platforms: DriverKit 25.5, iOS 26.5, macOS 26.5, tvOS 26.5, visionOS 26.5, watchOS 26.5
    IDEs:
      Android Studio: 2026.1 AI-261.23567.138.2611.15503007
      Xcode: 26.5/17F42 - /usr/bin/xcodebuild
    npmPackages:
      expo: ~56.0.12 => 56.0.12
      expo-router: ~56.2.11 => 56.2.11
      react: 19.2.3 => 19.2.3
      react-dom: 19.2.3 => 19.2.3
      react-native: 0.85.3 => 0.85.3
      react-native-web: ~0.21.0 => 0.21.2
    npmGlobalPackages:
      eas-cli: 20.3.0
    Expo Workflow: managed
```

### Expo Doctor Diagnostics

```text
21/21 checks passed. No issues detected!
```

HubertBer:
Hey, thanks for filing this issue! I've added a change that correctly propagates error codes to JS. Once new expo package version is published this should be fixed!

PR #47453, opened 2026-07-02
[JSI][iOS] Fix JavaScriptPromise not propagating error code correctly to JS

HubertBer:
# Why
In the issue https://github.com/expo/expo/issues/47372 there is a problem that in the newest expo release error codes are not propagated through promise reject to the JS from the Swift modules. This was fixed on main in #47259, however cherry-picking that change would require cherry-picking other changes as well like #47154 as it depends on `JavaScriptError` changes, which are not on sdk-56. 

# How
I've added a minimal change required for this fix, that is construct a JavaScriptError differently based on whether the error was a JavaScriptThrowable or not, similar to the synchronous path.

# Test Plan
I've checked it manually in an example modue taken from the #47372 repro.

# Checklist

### DIFF
--- a/packages/expo-modules-jsi/CHANGELOG.md
+++ b/packages/expo-modules-jsi/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### 🐛 Bug fixes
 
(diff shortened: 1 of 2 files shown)


tsapeta (approved):
We could cherry pick https://github.com/expo/expo/pull/47154, but on the other hand this change is less risky to backport, so it's probably a better decision. Thanks!

PR #47455, opened 2026-07-02
[ci] Add gradle plugins to ccache key

Wenszel:
# Why

Changes in Gradle plugins cause the Android ccache hit rate to drop: https://github.com/expo/expo/actions/runs/28445779939/job/84294761381?pr=47377

Gradle plugin sources are compiled into JARs. When they change, Gradle/AGP can place transformed dependency artifacts under paths with new hashes:
`~/.gradle/caches/.../transforms/<hash>/transformed/...`
Those paths are passed to `clang++` as include paths, causing ccache to miss

# How
Added `.kt/.kts` paths for the main `expo-gradle-plugin` (in expo-autolinking) and for every package declaring `gradlePlugins` in its `expo-module.config.json`, which `expo-gradle-plugin` includeBuilds during the Android compilation

# Test Plan

Green CI

### DIFF
--- a/.github/actions/expo-caches/action.yml
+++ b/.github/actions/expo-caches/action.yml
@@ -171,7 +171,8 @@ runs:
         # ccache tracks these files without any sloppiness issues.
         # - C/C++ source files
         # - Gradle files: see https://github.com/expo/expo/pull/44884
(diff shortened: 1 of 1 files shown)


Wenszel:
* **#47455** <a href="https://app.graphite.com/github/pr/expo/expo/47455?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/47455?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

chrfalch (approved):
LGTM. 

Nit: packages/expo-brownfield/gradle-plugins also implements org.gradle.api.Plugin and isn't covered — worth adding if brownfield hits the same transform-path churn.

Wenszel:
Hey, good catch! I should have mentioned that these plugins have to be included by the main autolinking gradle plugin; that's why brownfield isn't there - it doesn't declare any gradlePlugins in its `expo-module.config.json`. I've changed the PR description to be more accurate.

PR #47466, opened 2026-07-02
[expo-router] Forward disableTopInsetApplication to the native stack header config

idoyana:
# Why

On Android, a native stack header rendered below a top-positioned element that already consumes the status-bar inset (e.g. `Tabs` from `expo-router/js-tabs` with `tabBarPosition: "top"` and a tab bar padded by `useSafeAreaInsets().top`) is padded by the status-bar height a second time, producing a large empty band above the header title.

| Before | After (`disableTopInsetApplication: true`) |
| --- | --- |
| Header shows a status-bar-height gap between the tab bar and the title | Header sits flush below the tab bar |

Root cause chain:

- Since the edge-to-edge enforcement, `react-native-screens` hard-codes the header's top-inset application on Android (`shouldApplyTopInset = true` in `CustomToolbar.kt`) — the `topInsetEnabled` prop that `useHeaderConfigProps` passes is ignored there.
- The replacement coordination (`TopInsetApplicationContext`) only sees other `react-native-screens` headers, so a JS tab bar above the stack can't mark the inset as consumed.
- `react-native-screens` added an explicit opt-out for exactly this case — `disableTopInsetApplication` on `ScreenStackHeaderConfigProps` (software-mansion/react-native-screens#3835, extended in software-mansion/react-native-screens#4220). That PR's description notes it *requires changes in react-navigation* to be reachable from navigator options; those changes never landed in react-navigation, and expo-router's vendored native stack doesn't forward the prop either.

This PR adds that missing plumbing to expo-router's vendored native stack, following the exact integration sketched in software-mansion/react-native-screens#3835.

# How

- Added `disableTopInsetApplication?: boolean` to `NativeStackNavigationOptions` (Android-only, documented, default `undefined` — behavior is unchanged unless set).
- `useHeaderConfigProps` now destructures the option and includes it in the returned `ScreenStackHeaderConfigProps`. `react-native-screens` already consumes it from there: `ScreenStackItem` reads `headerConfig?.disableTopInsetApplication` for the inset-coordination context, and `ScreenStackHeaderConfig` receives it via the `{...headerConfig}` spread.

No behavior change on iOS (the prop is Android-only in `react-native-screens`) or for anyone not setting the option.

# Test Plan

Validated in a production app (expo-router 56.2.11 with this exact change applied as a pnpm patch, react-native-screens 4.25.2, RN 0.85.3, Android emulator, edge-to-edge):

1. Layout: `Tabs` from `expo-router/js-tabs` with `tabBarPosition: "top"` and a custom tab bar that pads itself with `useSafeAreaInsets().top`; each tab hosts a nested `Stack` with visible native headers.
2. Before: every screen with a visible header shows an extra status-bar-height white band between the tab bar and the header title; screens with `headerShown: false` are unaffected (control).
3. After adding `disableTopInsetApplication: true` to the stack's `screenOptions`: headers sit flush below the tab bar; back button, title, and header styling unchanged.
4. Without the option set, behavior is identical to before the change.

Repro layout:

```tsx
// (main)/_layout.tsx
<Tabs tabBar={(props) => <TopTabBar {...props} />} screenOptions={{ headerShown: false, tabBarPosition: 'top' }}>
  ...
</Tabs>

// (main)/settings/_layout.tsx
<Stack screenOptions={{ disableTopInsetApplication: true }}>
  <Stack.Screen name="index" options={{ title: 'Settings' }} />
</Stack>
```

# Checklist


Ubax:
Hi @idoyana 
Thanks for making us aware of this issue and opening a PR.

Internally we agreed to take a different approach here - https://github.com/expo/expo/pull/47482. We will expose `nativeProps` option which you can use to pass any property directly to `react-native-screens`. Will this solution work for you?

idoyana:
Verified #47482 against our case — the `headerConfig` merge lands in the same returned `ScreenStackHeaderConfigProps` object that `ScreenStackItem` reads `disableTopInsetApplication` from, so the top-tab-bar double-inset scenario is fully covered by `unstable_nativeProps: { headerConfig: { disableTopInsetApplication: true } }`. That works for us; closing this PR in favor of #47482. We're on SDK 56 with this change applied as a local patch until we upgrade, so no backport needed.

PR #47476, opened 2026-07-02
[router] Export `Theme` type

eliotgevers:
# Why

`expo-router` already exports the values used to customize router themes from the root entry point:

```ts
import { DarkTheme, DefaultTheme, ThemeProvider } from 'expo-router';
```

The matching `Theme` type exists internally, but it is not exported from the root entry point. That makes custom themes a little awkward to type without reaching for the deprecated React Navigation compatibility surface or the global namespace.

# What changed

This exports `Theme` from the root `expo-router` entry point, so custom themes can use one import:

```ts
import { DefaultTheme, ThemeProvider, type Theme } from 'expo-router';

const theme: Theme = {
  ...DefaultTheme,
  colors: {
    ...DefaultTheme.colors,
    primary: '#2563eb',
  },
};
```

Without this, users need a workaround like:

```ts
const theme = {
  ...DefaultTheme,
  colors: {
    ...DefaultTheme.colors,
    primary: '#2563eb',
  },
} satisfies ReactNavigation.Theme;
```

# Test plan

- Ran `git diff --check`.
- Confirmed the exported type already exists at `src/react-navigation/native/types.tsx` and is used by `DefaultTheme` / `DarkTheme`.


Ubax:
Hi @eliotgevers 
Thanks for opening the PR. The code looks good 🎉 

Can you please rebase on the latest `main`?

hassankhan (approved):

eliotgevers:
@Ubax Done!

Ubax (approved):

Ubax:
Thanks for your contribution 🎉

PR #47477, opened 2026-07-02
[ios][camera] Fix UI hang and oversized files during photo capture

boojamya:
# Why

On iOS, `takePictureAsync` processes the captured still (full-resolution decode, crop, orientation normalization, re-encode) synchronously inside the `AVCapturePhotoCaptureDelegate` callback, which is often delivered on the main queue. On older devices this freezes the UI for several seconds per capture — the main-thread hang stacks bottom out in `UIImageJPEGRepresentation` → `CameraPhotoCapture.processImageData`, running under `_dispatch_main_queue_drain`.

Two related issues in the same path:

- `ExpoCameraUtils.normalizeOrientation` renders with the default `UIGraphicsImageRenderer` format (screen scale, 2x/3x), upscaling every rotated capture to 4–9x its native pixel count. The saved file is larger than the sensor resolution and its dimensions no longer match the returned `width`/`height`.
- `ExpoCameraUtils.data` re-encodes a throwaway full-quality JPEG just to copy its structural metadata, then encodes again — a redundant full-resolution pass per photo.

# How

- Move processing off the delegate callback onto a serial background queue, extracting the photo's data, metadata, and preview size first so `AVCapturePhoto` can be released promptly. The queue is serial with per-work-item autorelease draining so overlapping full-resolution pipelines can't accumulate into a memory termination.
- Pin `normalizeOrientation` to the image's own scale.
- Encode once directly from the `CGImage`, using the caller's metadata and the image's own color space.

# Test Plan

Built into a dev client and captured on device (iPhone SE, iPhone 11) with `exif: true`:

- UI stays responsive during capture; the multi-second freeze is gone.
- Saved photos are at native sensor resolution (previously ~4x larger) and the returned `width`/`height` match the file.
- Exif metadata in the saved file is intact.
- Portrait, landscape, front camera, and rapid successive captures behave correctly.

# Checklist

- [x ] I added a `changelog.md` entry and rebuilt the package sources according to [this short guide](https://github.com/expo/expo/blob/main/CONTRIBUTING.md#-before-submitting)
- [x ] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: updated a module plugin).
- [ x] Conforms with the [Documentation Writing Style Guide](https://github.com/expo/expo/blob/main/guides/Expo%20Documentation%20Writing%20Style%20Guide.md)


packages/expo-camera/ios/Common/ExpoCameraUtils.swift
```diff
@@ -188,38 +192,21 @@ struct ExpoCameraUtils {
...
     updatedMetadata.setObject(NSNumber(value: quality), forKey: kCGImageDestinationLossyCompressionQuality as NSString)
     let processedImageData = NSMutableData()
 
-    guard let sourceType = CGImageSourceGetType(sourceCGImageSourceRef) else {
-      return nil
-    }
-
     guard let destinationCGImageRef =
-      CGImageDestinationCreateWithData(processedImageData, sourceType, 1, nil) else {
+      CGImageDestinationCreateWithData(processedImageData, "public.jpeg" as CFString, 1, nil) else {
```
alanjhughes:
```suggestion
      CGImageDestinationCreateWithData(processedImageData, UTType.jpeg.identifier as CFString, 1, nil) else {
```

packages/expo-camera/ios/Current/CameraPhotoCapture.swift
```diff
@@ -124,21 +131,45 @@ class CameraPhotoCapture: NSObject, AVCapturePhotoCaptureDelegate {
...
+  }
+
+  private func currentPreviewSize() -> CGSize {
+    guard let captureDelegate else {
+      return .zero
     }
+    let size = captureDelegate.previewLayer.frame.size
+    return captureDelegate.deviceOrientation == .portrait
+      ? CGSize(width: size.height, height: size.width)
+      : CGSize(width: size.width, height: size.height)
```
alanjhughes:
```suggestion
    return captureDelegate.deviceOrientation == .portrait
      ? CGSize(width: size.height, height: size.width)
      : size
```

alanjhughes (approved):
Thank you for your contribution. Just a couple of small nits

PR #47488, opened 2026-07-03
[expo-observe] ENG-22118: Add filteredParams configuration option to navigation integrations

Ubax:
# Why

Route and query params can include sensitive information, which apps may want to ignore when sending data to observe

# How

1. Add `filteredParams` property to specify which params should be filtered. When params are filtered then the `url` for that entry will not be sent. 

# Test Plan

1. CI
2. Observe tester

image

3. Bare expo

image

`p2` and `q1` are filtered

# Checklist

### DIFF
--- a/apps/bare-expo/App.tsx
+++ b/apps/bare-expo/App.tsx
@@ -1,9 +1,9 @@
 import { ThemeProvider } from 'ThemeProvider';
 import BenchmarkHelper from 'benchmark-helper';
+import * as DevMenu from 'expo-dev-menu';
 import { Observe, ObserveRoot } from 'expo-observe';
 import * as Splashscreen from 'expo-splash-screen';
 import React from 'react';
-import * as DevMenu from 'expo-dev-menu';
 
 import MainNavigator, { optionalRequire } from './MainNavigator';
 
@@ -70,7 +70,7 @@ Observe.configure({
   dispatchingEnabled: true,
   sampleRate: 0.9,
   integrations: {
-    'react-navigation': true,
+    'react-navigation': { filteredParams: ['accountId', 'firstName'] },
   },
 });
 
--- a/packages/expo-observe/CHANGELOG.md
+++ b/packages/expo-observe/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 - Add ObserveInteractiveMarker component ([#46909](https://github.com/expo/expo/pull/46909) by [@Ubax](https://github.com/Ubax))
 - Expose configure event ([#47388](https://github.com/expo/expo/pull/47388) by [@Ubax](https://github.com/Ubax))
+- Add `filteredParams` configuration option to navigation integrations ([#47488](https://github.com/expo/expo/pull/47488) by [@Ubax](https://github.com/Ubax))
 
 ### 🐛 Bug fixes
 
--- a/packages/expo-observe/src/integrations/expo-router/__tests__/navigationConfig.test.ts
+++ b/packages/expo-observe/src/integrations/expo-router/__tests__/navigationConfig.test.ts
@@ -1,0 +1,49 @@
+import {
+  getNavigationMetricParams,
+  getNavigationRouteParams,
+} from '../../navigationConfig';
+
+describe('expo-router navigation config', () => {
+  it('filters configured route and query param keys', () => {
+    expect(
+      getNavigationRouteParams(
+        { filteredParams: ['userId', 'token', 42 as unknown as string] },
+        { userId: '1', tab: 'posts', token: 'secret' }
+      ).routeParams
+    ).toEqual({ tab: 'posts' });
+  });
+
+  it('returns an empty object when every param is filtered', () => {
+    expect(
+      getNavigationRouteParams({ filteredParams: ['userId'] }, { userId: '1' }).routeParams
+    ).toEqual({});
+  });
+
+  it('keeps params and URL visible by default', () => {
+    const params = { userId: '1' };
+    expect(getNavigationRouteParams(true, params).routeParams).toBe(params);
+    expect(getNavigationMetricParams(true, params, '/users/1')).toEqual({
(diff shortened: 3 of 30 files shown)


Ubax:
* **#47497** <a href="https://app.graphite.com/github/pr/expo/expo/47497?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#47488** <a href="https://app.graphite.com/github/pr/expo/expo/47488?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/47488?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/expo-observe/src/integrations/expo-router/__tests__/navigationConfig.test.ts
```diff
@@ -0,0 +1,62 @@
+import {
+  filterNavigationParams,
+  getNavigationMetricParams,
+} from '../../navigationConfig';
+
+const mockNative = {
+  getIntegrations: jest.fn(() => ({})),
+};
+
+jest.mock('../../../nativeModule', () => ({ __esModule: true, default: mockNative }));
```
Ubax: [🚨 **critical**] This suite fails to run: `ReferenceError: Cannot access 'mockNative' before initialization`. Jest hoists this `jest.mock` factory above everything, but the `import` of `../../navigationConfig` (also hoisted) eagerly `require`s `../nativeModule`, which fires this factory *before* the `const mockNative` initializer executes — so the factory dereferences `mockNative` in its TDZ.

I reproduced this at the PR head commit with the repo's `yarn test`: all 14 new/modified suites fail to load (0 tests run), so CI is red and none of the new coverage is actually exercised. The same pattern is in every new test file that mocks `nativeModule` (`react-navigation/__tests__/navigationConfig.test.ts`, and the added blocks in `expo-router/__tests__/init.test.native.ts`, `useObserveForRouter.test.native.tsx`, `react-navigation/__tests__/handleStateChange.test.native.ts`, `integration.test.native.tsx`, `useObserveForReactNavigation.test.native.tsx`).

Fix by not referencing an outer var at factory time — e.g. define the mock inline and grab the handle after: `jest.mock('../../../nativeModule', () => ({ __esModule: true, default: { getIntegrations: jest.fn(() => ({})) } }))` then `import native from '../../../nativeModule'; const mockNative = native as unknown as { getIntegrations: jest.Mock }`. Or defer the deref: `default: { getIntegrations: (...a) => mockNative.getIntegrations(...a) }`.

packages/expo-observe/src/integrations/navigationConfig.ts
```diff
@@ -0,0 +1,86 @@
...
+
+function getFilteredParamKeys(integration: NavigationIntegrationName): Set<string> | null {
+  const filteredParams = getIntegrationConfig(integration)?.filteredParams;
+  if (!Array.isArray(filteredParams)) return null;
+
+  const keys = new Set(filteredParams.filter((key): key is string => typeof key === 'string'));
+  return keys.size > 0 ? keys : null;
+}
+
+export function filterNavigationParams<T extends object | undefined>(
```
Ubax: [🏗️ **design**] `filterNavigationParams` has no production callers — it's only referenced by the two `navigationConfig.test.ts` files. It's a thin wrapper over `getFilteredNavigationParams(...).routeParams`. Consider dropping it and asserting via the functions the code actually uses (`getNavigationRouteParams` / `getNavigationMetricParams`) so the tests exercise real surface rather than a test-only export.

packages/expo-observe/src/integrations/navigationConfig.ts
```diff
@@ -0,0 +1,86 @@
...
+  routeParams: T | Record<string, never>;
+  ignoredParam: boolean;
+};
+type NavigationRouteParams<T extends object | undefined> = {
+  routeParams: T | Record<string, never>;
+  urlHidden?: true;
+};
+
+function getIntegrations(): ObserveIntegrationsConfig {
+  return native.getIntegrations();
```
Ubax: [💡 suggestion] `getIntegrations()` is a synchronous native (JSI) call and this runs on every navigation event — often more than once per event (e.g. the expo-router path calls `getNavigationMetricParams` in `init.ts`, then `emitTTI` calls it again → two native round-trips for the same event). The integrations config only changes on `configure`, so this could be read once and cached (invalidated on the `configure` event the module already broadcasts). Minor, but it's a hot path.

packages/expo-observe/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 - Add ObserveInteractiveMarker component ([#46909](https://github.com/expo/expo/pull/46909) by [@Ubax](https://github.com/Ubax))
 - Expose configure event ([#47388](https://github.com/expo/expo/pull/47388) by [@Ubax](https://github.com/Ubax))
+- Add filteredParams configuration option to navigation integrations ([#47488](https://github.com/expo/expo/pull/47488) by [@Ubax](https://github.com/Ubax))
```
tsapeta:
```suggestion
- Add `filteredParams` configuration option to navigation integrations ([#47488](https://github.com/expo/expo/pull/47488) by [@Ubax](https://github.com/Ubax))
```

apps/bare-expo/App.tsx
```diff
@@ -70,7 +70,7 @@ Observe.configure({
   dispatchingEnabled: true,
   sampleRate: 0.9,
   integrations: {
-    'react-navigation': true,
+    'react-navigation': { filteredParams: ['p2', 'q1'] },
```
tsapeta: Could we use less cryptic names? Like `userId`, `firstName`, `email` etc.

tsapeta (approved):

PR #47496, opened 2026-07-03
[jest-expo] Migrate module mock generation to bare-expo

alanjhughes:
# Why
Expo go excludes a lot of modules and it was annoying to need to use it for this so I'm moving it into bare-expo

# How

- Added a local module in bare-expo that exposes getModulesSchema.
- Added a NCL screen that generates the mocks like the old mock generator app.
- Removed getModulesSchema function from the ExpoGoModule and deleted the jest-expo-mock-generator app.
- Regenerated `expoModules.js` from bare-expo, which correctly drops the ExpoGo module and picks up the wider SDK set that bare-expo links.
- Added expo-router to bare-expo so its native modules are linked and reflected. ExpoRouter is Android only and cannot be reflected by the iOS generator, so its mock is preserved by hand in internalExpoModules.js, and the redundant ExpoSplashScreen entry was removed there since bare-expo generates it identically.

# Test Plan
Bare-expo, regenerate the module mocks


tsapeta (approved):
Good idea 👍 We should actually find a better way to generate these mocks – the schema was designed for the legacy modules system. It's one of the reasons why we can't remove some legacy code...

alanjhughes:
@tsapeta cool, makes sense. I'll add a task to linear 👍

PR #47497, opened 2026-07-03
[expo-observe] ENG-22120 Fix non-serializable route params issue

Ubax:
# Why

See https://linear.app/expo/issue/ENG-22120/investigate-crash-with-non-serializable-navigation-event-params

# How

1. `JSON.parse(JSON.stringify(value))` value before passing to native. Catch and don't save on exception

# Test Plan

1. CI
2. Bare expo

# Checklist

### DIFF
--- a/packages/expo-observe/src/integrations/navigationConfig.ts
+++ b/packages/expo-observe/src/integrations/navigationConfig.ts
@@ -1,15 +1,11 @@
 import type { ObserveNavigationIntegrationConfig } from '../types';
 
 type NavigationIntegrationConfig = boolean | ObserveNavigationIntegrationConfig | undefined;
-type NavigationMetricParams<T extends object | undefined> = {
-  routeParams: T | Record<string, never>;
+type NavigationMetricParams = {
+  routeParams: Record<string, unknown>;
 } & ({ url: string | undefined; urlHidden?: false } | { url?: undefined; urlHidden: true });
-type FilteredNavigationParams<T extends object | undefined> = {
-  routeParams: T | Record<string, never>;
-  ignoredParam: boolean;
-};
-type NavigationRouteParams<T extends object | undefined> = {
-  routeParams: T | Record<string, never>;
+type NavigationRouteParams = {
+  routeParams: Record<string, unknown>;
   urlHidden?: true;
 };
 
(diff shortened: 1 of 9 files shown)


Ubax:
* **#47497** <a href="https://app.graphite.com/github/pr/expo/expo/47497?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/47497?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#47488** <a href="https://app.graphite.com/github/pr/expo/expo/47488?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/expo-observe/src/integrations/navigationConfig.ts
```diff
@@ -37,47 +36,60 @@ function getFilteredParamKeys(integration: NavigationIntegrationName): Set<strin
...
+  const routeParams = Object.fromEntries(
+    Object.entries(params ?? {})
+      .filter(([key]) => {
+        const keep = !filteredKeys?.has(key);
+        ignoredParam ||= !keep;
+        return keep;
+      })
+      .map(([key, value]) => {
+        try {
+          return [key, JSON.parse(JSON.stringify(value))] as const;
```
Ubax: [💡 suggestion] The round-trip does double duty here — it both *tests* serializability and *is* the sanitized value. That produces an asymmetry worth confirming is intentional:

- A top-level non-serializable value (`callback: () => {}`) throws → the whole param is dropped and `ignoredParam`/`urlHidden` is set.
- A nested non-serializable value (`nested: { ok: true, fn: () => {} }`) does *not* throw — `JSON.stringify` silently omits `fn`, so the param survives as `{ ok: true }` and neither `ignoredParam` nor `urlHidden` is set.

So the same kind of unexportable data leaks or hides the URL depending only on nesting depth. If the intent is "any dropped data hides the URL", nested stripping should also flip `ignoredParam`.

Separately, `undefined` and symbol values throw the same way functions do (`JSON.parse(JSON.stringify(undefined))` → SyntaxError), so a param that just happens to carry an `undefined` value is treated as non-serializable and hides the URL. Optional/omitted params with `undefined` values are common enough from navigation that this may fire more often than intended — consider skipping `undefined`/`null` before the round-trip.

packages/expo-observe/src/integrations/navigationConfig.ts
```diff
@@ -37,47 +36,60 @@ function getFilteredParamKeys(integration: NavigationIntegrationName): Set<strin
...
-  const { routeParams, ignoredParam } = getFilteredNavigationParams(integration, params);
   return {
     routeParams,
+    ignoredParam,
     ...(ignoredParam ? { urlHidden: true as const } : {}),
   };
 }
 
-function getFilteredNavigationParams<T extends object | undefined>(
+export function prepareRouteParams(
```
Ubax: [🐧 nit] `prepareRouteParams` is exported but only referenced from tests — the production paths use `getNavigationRouteParams`/`getNavigationMetricParams`. The two navigationConfig test files could assert against `getNavigationRouteParams(...).routeParams` instead, letting this test-only export go. Minor, take it or leave it.

packages/expo-observe/src/integrations/navigationConfig.ts
Ubax: How to address:
- Don't set `urlHidden` on non-serializable
- Check for `null` and `undefined` in `JSON.....`

tsapeta (approved):

Issue #418, opened 2017-07-26
I get an 'xcrun simctl' error when in the Expo SDK when I open my project in the iOS simulator.


MarcusHurney:
For the lifetime of my project and other projects I've made using the Expo SDK, I've gotten this error every single time I open my project in the iOS simulator. It doesn't affect my application in any way, but I do wonder why I am seeing it. Could you please help me understand this error?

<img width="573" alt="screen shot 2017-07-25 at 5 32 17 pm" src="https://user-images.githubusercontent.com/13347886/28599463-a1718274-715f-11e7-823e-e4fbf612ef75.png">

lhp--:
i too

lhp--:
1.sudo gem install cocoapods
2.pod install(in /ios)
3.react-native run-ios
it work :)

ide:
You don't want to use CocoaPods for this, it's unrelated. I suspect that there's either an issue with your iOS simulator or your development server is not running.

First verify the development server is running by opening localhost:19000 (or whatever URL you see in the logs) in your browser. If you get back a JSON response, it's running. If not, click Restart.

Then try uninstalling the Expo app from the iOS simulator and opening your project in the simulator again from XDE. Make sure you're using the simulator with Xcode 8 (Xcode 9 beta is not supported). You also can try Reset Content & Settings in the simulator since it sometimes can get into a bad state, but this is less common.

AlberZong:
Reset Ios content and settings your ios Simulator
I fixed this error

skevy:
Please let us know if resetting the simulator didn't work for you.

Thanks!

iamclaytonray:
@skevy @ide - 

I'm currently running into a similar issue as above. I do see a huge difference in the app. (Tested against several.) Even outside of Expo on the Simulator, everything is running incredibly slow. Every transition takes several seconds. I've deleted the app on the Simulator and my phone. I deleted Expo XDE from my Macbook Pro. I've reset Content & Settings. I've verified that the server is running.

Moving forward, what should I do that hasn't been recommended yet? I'm not really a mobile developer. I'm more of a web guy. In fact, this past month has been the first time I've used Expo.

As far as a repro goes, I can't share either of the apps I am currently working on but I did clone another app (not mine) that is also returning this error. I haven't touched anything in XCode or my `app.json` besides some basic things in the beginning, about a month ago. I changed a few styles and then out of nowhere, this error hit me.

Thanks for any help you can give!

[...]

PR #47498, opened 2026-07-03
[docs] Update markdown generation to account native upgrade helper diffs

amandeepmittal:
# Why

Fix ENG-22652

# How

- Add `scripts/generate-upgrade-diff-pages.ts` to write a markdown page per SDK version pair (`out/bare/upgrade/<from>-to-<to>/index.md`) from `diffInfo.json` during `generate-markdown-pages`, skipping `unversioned` pairs when the environment hides them
- Resolve version-pinned markdown requests in `public/_worker.js` (`?fromSdk`/`?toSdk`, including a trailing `.md` on the value) to the pair file, falling back to the default `index.md`
- Show the Copy page dropdown on `/bare/upgrade` once a pair is selected and point its actions at the pair file via `getVersionedMarkdownPath` in `MarkdownActions/paths.ts`
- Export the prompt instructions from `buildUpgradePrompt.ts` for reuse, and cover the above in jest and `test-worker.ts` (fixture now runs against the real `_redirects`)

# Test Plan

On the preview:

- Open **/bare/upgrade/**, select from SDK 52 to SDK 57, and confirm the Copy page dropdown appears in the header
- Use **View markdown** from the dropdown and confirm it opens the markdown titled "Upgrade native projects from SDK 52 to SDK 57", then change the selectors and confirm the link follows
- Use **Copy markdown** and paste somewhere to confirm the prompt and diff match the selected pair
- Visit **/bare/upgrade/52-to-57.md** directly and confirm the same pair markdown renders
- Visit **/bare/upgrade/?fromSdk=52&toSdk=57.md** (the append-.md gesture on a pinned URL) and confirm the pair markdown renders

# Checklist

- [ ] I added a `changelog.md` entry and rebuilt the package sources according to [this short guide](https://github.com/nicknisi/dotfiles/wiki/Pull-Request-Guidelines).

### DIFF
--- a/docs/public/_worker.js
+++ b/docs/public/_worker.js
@@ -1,24 +1,48 @@
+function upgradeHelperPairPath(url) {
+  if (!/^\/bare\/upgrade\/?$/.test(url.pathname)) return null;
+
+  const from = (url.searchParams.get("fromSdk") || "").replace(/\.md$/, "");
+  const to = (url.searchParams.get("toSdk") || "").replace(/\.md$/, "");
+  const version = /^(\d+|unversioned)$/;
+  if (!from || !to || from === to || !version.test(from) || !version.test(to)) {
+    return null;
+  }
+
+  return `/bare/upgrade/${from}-to-${to}/index.md`;
+}
+
 export default {
   async fetch(request, env) {
     const accept = request.headers.get("Accept") || "";
+    const url = new URL(request.url);
+    const pairPath = upgradeHelperPairPath(url);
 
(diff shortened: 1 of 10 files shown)


entiendoNull (changes requested):
I started this up locally at first and immediately ran into this hydration issue. The markdown actions visibility depend on the URL query params, which the server doesn't know about... so the server and first client render disagree. 

Same error shows up on the preview console as React error #418.

Image

Image

https://react.dev/errors/418

entiendoNull (approved):

PR #47508, opened 2026-07-03
[ncl][bare-expo] Prepare screens for expo-router migration

Kudo:
# Why

First of two PRs migrating `bare-expo`, `native-component-list`, and `test-suite` from react-navigation to expo-router (SDK 56 apps should no longer depend on `@react-navigation/*`). This one is behavior-preserving prep so the actual flip PR stays reviewable.

close ENG-23053

# How

- Extract the api/component screen lists out of the stack navigators into pure registry modules (`apiScreens.ts`, `componentScreens.ts`) plus id-based lookups (`screenRegistry.ts`), so file-based routes can later resolve screens by their deep-link id.
- Make route params serializable, since expo-router params travel through URLs: MediaLibrary passes `assetId`/`albumId` and refreshes on focus instead of an `onGoBack` callback, Calendars (legacy and @Next) pass `calendarId` and refetch, SVG passes the example key.
- Flatten inner stack navigators: `SearchScreen` now exports a plain screen and the host stack applies its header options; the react-native-screens demo registers its sub-screens as regular routes (`components/screens/*`) instead of embedding navigators.
- Drop the opt-in nav-state persistence (lock toggle + AsyncStorage restore) — expo-router owns navigation state, so this won't carry over.

# Test Plan

- `yarn tsc`, `yarn lint`, and `yarn format --check` pass in `apps/native-component-list`.
- `expo export -p web` (NCL) and `expo export -p ios` (bare-expo) bundle successfully.
- bare-expo Maestro e2e in CI.


Kudo:
* **#47890** <a href="https://app.graphite.com/github/pr/expo/expo/47890?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#47882** <a href="https://app.graphite.com/github/pr/expo/expo/47882?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#47510** <a href="https://app.graphite.com/github/pr/expo/expo/47510?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#47880** <a href="https://app.graphite.com/github/pr/expo/expo/47880?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#47508** <a href="https://app.graphite.com/github/pr/expo/expo/47508?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/47508?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Kudo:
@codex review

Kudo:
@codex review

apps/native-component-list/src/navigation/screenRegistry.ts
```diff
@@ -0,0 +1,33 @@
+import { getScreenIdForLinking } from 'test-suite/screens/getScreenIdForLinking';
```
Kudo, line 1: Note for reviewers: nothing consumes this registry in this PR yet — the stacked expo-router flip uses `findApiScreen`/`findComponentScreen` to resolve screens from dynamic route segments. It reuses test-suite's `getScreenIdForLinking` (already used by `ComponentListScreen` and `MainNavigators` for the linking config) so the registry keys are guaranteed to match the deep-link URLs.

apps/native-component-list/src/navigation/RootNavigation.tsx
```diff
@@ -29,17 +29,13 @@ export const linking: LinkingOptions<object> = {
           components: MainNavigators.components.linking,
         },
       },
-      search: {
-        screens: {
-          search: 'search',
-        },
-      },
+      searchNavigator: 'search',
```
Kudo, line 32: This also fixes a pre-existing mismatch: the old key was `search`, but the screen is registered as `searchNavigator`, so the `/search` path never actually matched. With `SearchScreen` no longer wrapping its own stack, the path now maps directly to the screen.

apps/bare-expo/MainNavigator.tsx
```diff
@@ -130,64 +135,32 @@ function TabNavigator() {
     </Tab.Navigator>
   );
 }
-const PERSISTENCE_KEY = 'NAVIGATION_STATE_V1';
-
 export default function MainNavigator() {
```
Kudo, line 134: The opt-in navigation-state persistence (the lock toggle + AsyncStorage snapshot) is removed intentionally: after the expo-router flip the router owns navigation state, and a persisted react-navigation state blob can't be restored into it. This also lets `markInteractive` fire directly instead of waiting for the async restore.

What happens to this feature (from #30360: get back to the screen you were testing after a native rebuild) after the migration:

- **Fast Refresh** keeps you on the current screen while editing JS — no restore needed.
- **JS reloads** in the same native session re-derive state from the launch URL, so an app opened via `bareexpo://components/blurview` keeps returning there.
- **Cold start / native rebuild**: every screen becomes URL-addressable, so you relaunch straight into the screen with `yarn open ios <screen>` or `xcrun simctl openurl` — reproducible and scriptable, unlike the persisted state blob that was coupled to the navigator's exact shape (hence `NAVIGATION_STATE_V1`) and silently rotted on refactors.

If automatic resume turns out to be missed in practice, it can come back as a small follow-up on top of expo-router: persist the `usePathname()` string and `router.replace()` it on startup when no deep link is present. A path string survives navigator refactors and — now that all params are serializable and in the URL — captures the full screen state, so that version would be strictly more robust than the old blob.

apps/native-component-list/src/screens/Screens/index.tsx
```diff
@@ -1,42 +1,55 @@
...
+import { type ScreenConfig } from '../../types/ScreenConfig';
 
-const SCREENS: Record<string, { component: any; options: { title: string } }> = {
-  Container: { component: Container, options: { title: 'ScreenContainer example' } },
-  NativeStack: { component: NativeStack, options: { title: 'ScreenStack example' } },
-  Navigation: { component: Navigation, options: { title: 'React Navigation example' } },
-};
-
-type Links = { Container: undefined; NativeStack: undefined; Navigation: undefined };
+export const ScreensExampleScreens: ScreenConfig[] = [
```
Kudo, line 12: The react-native-screens examples used to live in their own nested stack navigator inside this screen. That ad-hoc navigator can't be expressed as an expo-router route, so the three examples are flattened into regular screen entries (`screens/container`, `screens/native-stack`, `screens/navigation`) registered on the Components stack — this screen is now just a list linking to them. Behavior change: the examples push onto the parent stack instead of a nested one, and each becomes individually deep-linkable.

apps/native-component-list/src/screens/MediaLibrary/MediaLibraryScreen.tsx
```diff
@@ -257,8 +260,16 @@ function MediaLibraryView({ navigation, route, accessPrivileges }: Props) {
   }, []);
 
   // Subscribe to state changes
+  const isFirstFocus = React.useRef(true);
```
Kudo, line 263: This refresh-on-focus replaces the removed `onGoBack` callback param: route params must stay serializable, so the details screen can no longer call back into this screen after deleting/favoriting an asset. Skipping the first focus avoids a duplicate fetch on initial mount.

apps/native-component-list/src/screens/SearchScreen.tsx
```diff
@@ -1,16 +1,35 @@
...
+import { screenApiItems as ApiScreenApiItems } from '../navigation/apiScreens';
+import { screenApiItems as ComponentScreenApiItems } from '../navigation/componentScreens';
 import ComponentListScreen from './ComponentListScreen';
 
 const fuse = new Fuse(ApiScreenApiItems.concat(ComponentScreenApiItems), { keys: ['name'] });
 
-function SearchScreen({ navigation }: NativeStackScreenProps<SearchStack, 'search'>) {
+// The header (with the native search bar) comes from the stack that renders this screen,
+// so hosts must apply `getSearchScreenOptions` to that stack screen.
+export function getSearchScreenOptions(theme: ThemeType): NativeStackNavigationOptions {
```
Kudo, line 18: The single-screen wrapper stack is dissolved so the search screen can be hosted directly by the root stack (a nested navigator would also break the direct `/search` linking path).

Why export `getSearchScreenOptions` instead of letting the screen configure itself: the host navigators default to `headerShown: false`, so the themed header must be declared on the host's `Stack.Screen` entry. Options applied at registration are known before the screen is pushed, so the header is correct from the first frame of the transition; setting them from inside the component (`navigation.setOptions` / inline `<Stack.Screen options>` after the flip) only lands after the screen renders, which can pop the header in mid-animation. It's a function because the header colors depend on the current theme, which lives in the host's context. Both hosts (bare-expo `MainNavigator` and NCL's `RootNavigation`) apply it, and the expo-router flip reuses the same function in `app/_layout.tsx`.

Once the react-navigation hosts are deleted, moving these options inline into the `app/search.tsx` route (via `<Stack.Screen options>`) would be a reasonable simplification if the header pop-in turns out not to matter in practice.

apps/native-component-list/src/navigation/apiScreens.ts
```diff
@@ -0,0 +1,531 @@
...
+import { ContactsScreens } from '../screens/Contacts/ContactsScreen';
+import { ContactsNextScreens } from '../screens/Contacts@Next/ContactsNextScreen';
+import { CryptoScreens } from '../screens/Crypto/CryptoScreen';
+import { MediaLibraryScreens } from '../screens/MediaLibrary@Next/MediaLibraryScreens';
+import { ModulesCoreScreens } from '../screens/ModulesCore/ModulesCoreScreen';
+import { WorkletsScreens } from '../screens/Worklets/WorkletsScreen';
+import { type ScreenConfig } from '../types/ScreenConfig';
+import { optionalRequire } from './routeBuilder';
+
+export const ScreensList: ScreenConfig[] = [
```
Kudo, line 17: Reviewer note: this file and `componentScreens.ts` are a mechanical extraction of `ScreensList`/`Screens`/`screenApiItems` out of `ExpoApisStackNavigator`/`ExpoComponentsStackNavigator` — the entries themselves are unchanged. The split lets the react-navigation stacks (still used here) and the expo-router route registry (next PR) consume the same screen lists without importing navigator code. The only additions are the flattened `ScreensExampleScreens` entries in `componentScreens.ts`.

apps/native-component-list/src/screens/SVG/SVGScreen.tsx
```diff
@@ -18,7 +18,7 @@ export const SVGScreens = [
 ];
 
 type StackParams = {
-  SVGExample: { title: string; key: string };
+  SVGExample: { exampleKey: string };
```
Kudo, line 21: Params become visible URL query strings after the router flip, so the two duplicated params (`title` was always equal to `key`) collapse into one self-describing `exampleKey` — also avoiding `key`, which reads as React's special prop / react-navigation's internal `route.key`.

apps/bare-expo/MainNavigator.tsx
```diff
@@ -57,9 +56,15 @@ const NativeComponentList: NativeComponentListExportsType = optionalRequire(() =
...
-const Search = optionalRequire(() =>
-  require('native-component-list/src/screens/SearchScreen')
-) as any;
+const SearchScreenModule = (() => {
+  try {
+    return require('native-component-list/src/screens/SearchScreen');
+  } catch {
+    return null;
+  }
+})();
```
Ubax: Why do you need to change from `optionalRequire` here?

apps/native-component-list/src/screens/Screens/index.tsx
Ubax, line 1: Does it make sense to keep it here? Do we have any tests for screens in test-suite?

I would mostly consider this, because the app itself will become one big screens test

Ubax (approved):

apps/native-component-list/src/navigation/screenRegistry.ts
```diff
@@ -0,0 +1,33 @@
...
+
+import { type ScreenConfig } from '../types/ScreenConfig';
+import { Screens as apiScreens } from './apiScreens';
+import { Screens as componentScreens } from './componentScreens';
+
+function buildRegistry(screens: ScreenConfig[]): Map<string, ScreenConfig> {
+  const registry = new Map<string, ScreenConfig>();
+  for (const screen of screens) {
+    registry.set(getScreenIdForLinking(screen), screen);
```
Ubax: [💡 suggestion]  `Map.set` silently overwrites on duplicate ids. With ~150 screens whose ids are mostly *derived* from display names, two names can collapse to the same id (the derivation strips `@`, parentheses, etc.) and one screen becomes unreachable by deep link with no signal. A dev-mode `if (registry.has(id)) console.warn(...)` would make collisions visible.

apps/native-component-list/src/screens/MediaLibrary/MediaDetailsScreen.tsx
```diff
@@ -163,10 +158,9 @@ export default class MediaDetailsScreen extends React.Component<Props> {
           />
         )}
 
-        <View style={styles.imageContainer}>{this.renderAsset(asset)}</View>
-
-        <HeadingText>Base asset data</HeadingText>
-        <MonoText>{JSON.stringify(asset, null, 2)}</MonoText>
```
Ubax, line 169: Just checking, did you mean to remove this section?

Ubax (approved):
More nits after automated review

apps/bare-expo/MainNavigator.tsx
Kudo: Done in 5791984. I reused `optionalRequire` and cast the module, so no new helper. The reason for the change: I need the named export `getSearchScreenOptions`, which `optionalRequire` alone drops (it returns only `.default`).

apps/native-component-list/src/screens/Screens/index.tsx
Kudo: Good point. I kept it for now: expo-router uses react-native-screens indirectly and does not exercise these raw `ScreenContainer`/`ScreenStack` primitives, so the app-wide usage does not fully replace this demo. I added a `TODO(kudo,20260802)` in 5791984 to revisit and likely remove it in a follow-up PR.

apps/native-component-list/src/navigation/screenRegistry.ts
Kudo: Good catch. Added a dev-mode `console.warn` in `buildRegistry` (5791984) that fires when two screens resolve to the same id, naming both screens so the collision is visible.

apps/native-component-list/src/screens/MediaLibrary/MediaDetailsScreen.tsx
Kudo: Yes, intentional. Route params are now just `assetId` (they must stay serializable), so the raw `asset` object is no longer passed in. The image still renders from the refetched `previewDetails`, and the full data shows in the `getAssetInfoAsync(assetId)` section right below — a superset of the old "Base asset data". So nothing is lost.

PR #47510, opened 2026-07-03
[bare-expo][ncl][test-suite] Migrate from react-navigation to expo-router

Kudo:
# Why

In SDK 56 expo-router replaced react-navigation, and apps should no longer have `@react-navigation/*` in their dependencies. bare-expo, native-component-list, and test-suite were the last unmigrated apps. They're coupled (bare-expo embeds navigators from the other two), so all three flip together here.

Stacked on #47508 (behavior-preserving prep) — will retarget to `main` once that lands.

close ENG-23053

# How

- **test-suite**: `app/` routes replace `App.js`/`TestStackNavigator` (index select screen, `run`, plus redirect routes for the legacy `/all` and `/select/<tests>` deep-link forms). SelectScreen's manual cold-start `Linking` handler is deleted — the router owns URL mapping, so the old race is gone. `navigationConfig.tsx` exports the shared themed stack options that bare-expo reuses.
- **native-component-list**: `src/app/` routes replace the navigation containers. Tabs are **NativeTabs** on native and narrow web; web desktop keeps the permanent Drawer. The ~330 example screens resolve through registry catch-all routes (`apis/[...id]`, `components/[...id]`) by deep-link id instead of one route file per screen. Imperative `navigate('Name')` call sites now push router paths, and remaining imports use the vendored equivalents (`expo-router`, `expo-router/js-tabs`, `expo-router/react-navigation`). `react-native-paper` is removed (only used for Android material bottom tabs).
- **bare-expo**: `app/` routes replace `MainNavigator` with NativeTabs (test-suite / playground / apis / components). The NCL-backed routes keep the `optionalRequire` tolerance — the apis/components triggers hide when NCL isn't bundled. The expo-observe integration switches to `'expo-router'`. `EXPO_ROUTER_DISABLE_RN_NAVIGATION_CHECK=1` is removed from all scripts, so Metro itself now enforces that no `@react-navigation/*` import sneaks back into the bundle.
- **expo-router**: adds the missing `./react-navigation` subpath to the exports map — the SDK 56 codemod rewrites imports to it, but TypeScript's bundler module resolution couldn't resolve it.

The `bareexpo://` URL contract asserted by Maestro e2e is unchanged: `test-suite/run?tests=…`, `apis/<id>`, `components/<id>` (including nested ids like `components/image/comparison`).

Dropped per prior discussion: the opt-in nav-state persistence (lock toggle) and the react-native-paper material tabs.

Screenshot 2026-07-04 at 3 09 04 AM

# Test Plan

- `tsc`, `lint`, `format --check` pass in native-component-list and test-suite; test-suite jest passes (deep-link id tests intact).
- `expo export -p ios` succeeds for bare-expo **without** the RN-navigation-check bypass, and for test-suite; `expo export -p web` succeeds for NCL.
- Grep gate: no `@react-navigation` references left in the three apps.
- bare-expo Maestro e2e (CI) exercises the cold-start deep links and `test_suite_*` testIDs.


Kudo:
* **#47890** <a href="https://app.graphite.com/github/pr/expo/expo/47890?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#47882** <a href="https://app.graphite.com/github/pr/expo/expo/47882?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#47510** <a href="https://app.graphite.com/github/pr/expo/expo/47510?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/47510?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#47880** <a href="https://app.graphite.com/github/pr/expo/expo/47880?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#47508** <a href="https://app.graphite.com/github/pr/expo/expo/47508?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

apps/bare-expo/index.js
```diff
@@ -1 +1 @@
-import './index.tsx';
+import 'expo-router/entry';
```
Kudo: The app now boots through expo-router's entry, which mounts the routes in `app/`. This replaces the old `registerRootComponent` setup, so `index.tsx` and `MainNavigator.tsx` are gone.

apps/bare-expo/app/(tabs)/_layout.tsx
```diff
@@ -0,0 +1,33 @@
...
+    <NativeTabs>
+      <NativeTabs.Trigger name="test-suite">
+        <NativeTabs.Trigger.Icon sf="checklist" md="checklist" />
+        <NativeTabs.Trigger.Label>Tests</NativeTabs.Trigger.Label>
+      </NativeTabs.Trigger>
+      <NativeTabs.Trigger name="playground">
+        <NativeTabs.Trigger.Icon sf="flask" md="science" />
+        <NativeTabs.Trigger.Label>Playground</NativeTabs.Trigger.Label>
+      </NativeTabs.Trigger>
+      <NativeTabs.Trigger name="apis" hidden={!hasNativeComponentList}>
```
Kudo, line 23: bare-expo can build without native-component-list (the macOS app does), so the APIs/Components tabs hide themselves when the package isn't bundled — same behavior the old JS tab bar had.

apps/bare-expo/app/(tabs)/apis/[...id].tsx
```diff
@@ -0,0 +1,17 @@
...
+import { optionalRequire } from '../../../optionalRequire';
+
+const RegistryScreenRoute = optionalRequire(() =>
+  require('native-component-list/src/navigation/RegistryScreenRoute')
+)?.default;
+const findApiScreen = optionalRequire(() =>
+  require('native-component-list/src/navigation/screenRegistry')
+)?.findApiScreen;
+
+export default function ApiScreenRoute() {
```
Comment (Kudo) [inline apps/bare-expo/app/(tabs)/apis/[...id].tsx:5]: This single catch-all serves every NCL API screen, including nested ids — the URL segments join back into a registry key. The alternative was generating ~330 route files; the registry already knows each screen's component, title, and options, so one route plus a lookup is much easier to maintain.

apps/native-component-list/src/navigation/RegistryScreenRoute.tsx
```diff
@@ -0,0 +1,34 @@
...
+    | undefined;
+
+  if (!config || !Component) {
+    return <Unmatched />;
+  }
+  return (
+    <>
+      <Stack.Screen options={{ title: config.name, ...config.options }} />
+      {/* Pass react-navigation style props for screens that still read them. */}
+      <Component {...{ navigation, route: { params } }} />
```
Kudo, line 31: Many NCL screens still read `props.navigation` and `props.route.params` the react-navigation way. Handing them these props means none of those screens had to change in this PR — they can move to router hooks one by one later.

apps/bare-expo/app/_layout.tsx
```diff
@@ -70,11 +83,36 @@ Observe.configure({
   dispatchingEnabled: true,
   sampleRate: 0.9,
   integrations: {
-    'react-navigation': { filteredParams: ['accountId', 'firstName'] },
+    'expo-router': { filteredParams: ['accountId', 'firstName'] },
```
Kudo, line 86: expo-observe switched from its react-navigation integration to the expo-router one. `filteredParams` is carried over from main — it landed there while this branch was in flight.

apps/bare-expo/app/_layout.tsx
```diff
@@ -70,11 +83,36 @@ Observe.configure({
...
+    });
+  }, []);
+
+  return (
+    <GestureHandlerRootView style={{ flex: 1 }}>
+      <NavigationThemeProvider value={themeName === 'dark' ? DarkTheme : DefaultTheme}>
+        <Stack screenOptions={{ headerShown: false }}>
+          <Stack.Screen name="(tabs)" />
+          <Stack.Screen name="redirect" />
+          <Stack.Screen name="search" options={getSearchScreenOptions?.(theme)} />
```
Kudo, line 108: Search is a root-level route so it presents as a modal over the tabs. Its options (`headerSearchBarOptions` etc.) have to be set here at the navigator level — set inside the screen, the search bar pops in after mount. They come from NCL via `optionalRequire`, so this still works when NCL isn't bundled.

apps/bare-expo/app/_layout.tsx
```diff
@@ -14,22 +22,24 @@ try {
   // do nothing
 }
 
-DevMenu.registerDevMenuItems([
-  {
-    name: 'Action 1',
-    callback: () => {
-      console.log('Action 1 executed');
+if (process.env.EXPO_OS !== 'web') {
```
Kudo, line 25: expo-dev-menu is native-only, so registering menu items crashed the web build. Skipping on web fixes it.

apps/bare-expo/app/(tabs)/test-suite/select/[tests].tsx
```diff
@@ -0,0 +1,11 @@
+import { Redirect, useLocalSearchParams } from 'expo-router';
+import * as React from 'react';
+import { createQueryString, getSelectedTestNames } from 'test-suite/screens/getScreenIdForLinking';
+
+// Legacy deep-link form that runs a set of tests, e.g. `bareexpo://test-suite/select/basic,crypto`.
+export default function SelectTests() {
```
Comment (Kudo) [inline apps/bare-expo/app/(tabs)/test-suite/select/[tests].tsx:6]: Keeps the legacy `test-suite/select/basic,crypto` deep-link form working (CI's Maestro flows still use it). It normalizes the list and redirects into `/test-suite/run?tests=…`.

apps/bare-expo/app/index.tsx
```diff
@@ -0,0 +1,6 @@
+import { Redirect } from 'expo-router';
+import * as React from 'react';
+
+export default function Index() {
+  return <Redirect href="/test-suite" />;
```
Kudo, line 5: Opening the plain `bareexpo://` root lands on the Tests tab, matching the old initial tab.

apps/bare-expo/package.json
```diff
@@ -4,14 +4,14 @@
   "private": true,
   "main": "index.js",
   "scripts": {
-    "android": "NODE_ENV=\"development\" EXPO_ROUTER_DISABLE_RN_NAVIGATION_CHECK=1 ./scripts/start-emulator.sh",
+    "android": "NODE_ENV=\"development\" ./scripts/start-emulator.sh",
```
Kudo, line 7: This env var was the escape hatch for using react-navigation directly alongside the router. Nothing in bare-expo does that anymore, so the router's check can stay on.

apps/native-component-list/src/app/(main)/_layout.tsx
```diff
@@ -0,0 +1,46 @@
...
+  const isMobile = width <= 640;
+  const isTablet = !isMobile && width <= 960;
+  const isLargeScreen = !isTablet && !isMobile;
+
+  // Use native tabs on all except web desktop.
+  // NOTE(brentvatne): if you navigate to an example screen and then resize your
+  // browser such that the navigator changes from tab to drawer or drawer to tab
+  // then it will reset to the list because the navigator has changed and the state
+  // of its children will be reset.
+  if (Platform.OS !== 'web' || isMobile) {
```
Kudo, line 19: Same responsive split the old MainNavigators had: tabs on native and small web windows, a permanent drawer on desktop web. The only difference is the tabs are native now.

apps/native-component-list/src/app/(main)/apis/expo-observe/filtered/[userId]/[accountId].tsx
```diff
@@ -0,0 +1,56 @@
+import { useTheme } from 'ThemeProvider';
+import { useObserve } from 'expo-observe';
+import { Stack, useLocalSearchParams } from 'expo-router';
+import { useEffect } from 'react';
+import { StyleSheet, Text, View } from 'react-native';
+
+export default function FilteredParamsScreen() {
```
Comment (Kudo) [inline apps/native-component-list/src/app/(main)/apis/expo-observe/filtered/[userId]/[accountId].tsx]: The one NCL screen that's a real route instead of a registry entry: the demo needs actual path params for observe's `filteredParams` to have something to filter. The old version created its own nested native-stack, which can't exist under the no-react-navigation rule. Explicit routes win over the `[...id]` catch-all, so `apis/expo-observe/filtered/1/2` resolves here.

apps/native-component-list/src/utilities/useOptionalBottomTabBarHeight.ts
```diff
@@ -0,0 +1,10 @@
+import { BottomTabBarHeightContext } from 'expo-router/js-tabs';
+import { useContext } from 'react';
+
+/**
+ * Native tabs don't provide the JS bottom tab bar height context,
+ * so fall back to 0 and let the native content insets handle the tab bar.
+ */
+export default function useOptionalBottomTabBarHeight(): number {
+  return useContext(BottomTabBarHeightContext) ?? 0;
```
Kudo, line 9: react-navigation's `useBottomTabBarHeight` throws when no JS tab bar is mounted above you. This version returns 0 in that case — with native tabs the bar height is handled through native content insets, so screens don't need it.

apps/common/ThemeProvider.tsx
```diff
@@ -19,15 +25,26 @@ export const ThemeContext = createContext<ThemeContextType>({
...
+  const [themeOverride, setThemeOverride] = useState<ThemeName | null>(null);
+  const currentThemeName =
+    themeOverride ?? (systemColorScheme !== 'unspecified' ? systemColorScheme : 'light');
   const currentTheme = currentThemeName === 'dark' ? darkTheme : lightTheme;
 
+  const setTheme = useCallback((themeName: ThemeName) => {
+    if (typeof Appearance.setColorScheme === 'function') {
+      Appearance.setColorScheme(themeName);
+    } else {
+      setThemeOverride(themeName);
```
Kudo, line 38: react-native-web's `Appearance` has no `setColorScheme`, so the theme toggle crashed on web. Falling back to plain state keeps the toggle working there.

apps/bare-expo/app.json
```diff
@@ -49,6 +49,7 @@
       "url": "https://u.expo.dev/2c28de10-a2cd-11e6-b8ce-59d1587e6774"
     },
     "plugins": [
+      "expo-router",
```
Kudo, line 52: Prebuild needs the router's config plugin now that the app uses expo-router.

Kudo (commented):
Left inline notes on the less obvious parts to make reviewing easier.

apps/native-component-list/src/navigation/StackConfig.tsx
Ubax, line 37: WDYT about adding a TODO here to convert this into `Stack.Toolbar.Right` later?

apps/bare-expo/app/(tabs)/apis/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+  require('native-component-list/src/navigation/apiScreens')
+)?.screenApiItems;
+
+export default function ApisIndex() {
+  if (!ExpoApis) {
+    return null;
+  }
+  return (
+    <>
+      <Stack.Screen options={{ title: 'APIs in Expo SDK' }} />
```
Ubax:
```suggestion
      <Stack.Title>APIs in Expo SDK</Stack.Title>
```

Ubax: This can also be moved into `_layout.tsx` to simplify this file into

```tsx
return ExpoApis ? <ExpoApis apis={screenApiItems} /> : null;
```

apps/bare-expo/app/(tabs)/components/[...id].tsx
```diff
@@ -0,0 +1,17 @@
+import * as React from 'react';
+
+import { optionalRequire } from '../../../optionalRequire';
+
+const RegistryScreenRoute = optionalRequire(() =>
+  require('native-component-list/src/navigation/RegistryScreenRoute')
+)?.default;
```
Comment (Ubax) [inline apps/bare-expo/app/(tabs)/components/[...id].tsx]: WDYT about extracting this into a standalone component in `bare-expo`?

```tsx
import { optionalRequire } from '../../../optionalRequire';

const OptionalRegistryScreenRoute = optionalRequire(() =>
  require('native-component-list/src/navigation/RegistryScreenRoute')
)?.default;

export function RegistryScreenRoute(props: typeof ComponentProps<....>){
  if (!RegistryScreenRoute || !props.findScreen) {
    return null;
  }
  return <RegistryScreenRoute ...props/ >
}
```

apps/bare-expo/app/(tabs)/components/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+  require('native-component-list/src/navigation/componentScreens')
+)?.screenApiItems;
+
+export default function ComponentsIndex() {
+  if (!ExpoComponents) {
+    return null;
+  }
+  return (
+    <>
+      <Stack.Screen options={{ title: 'Components in Expo SDK' }} />
```
Ubax: Same here about title

apps/bare-expo/app/(tabs)/test-suite/select/[tests].tsx
Comment (Ubax) [reply apps/bare-expo/app/(tabs)/test-suite/select/[tests].tsx:6]: WDYT about using `+native-intent` here?

https://docs.expo.dev/router/advanced/native-intent/#rewrite-incoming-native-deep-links

apps/bare-expo/app/(tabs)/test-suite/all.tsx
```diff
@@ -0,0 +1,10 @@
+import { Redirect } from 'expo-router';
+import * as React from 'react';
+import { getTestModules } from 'test-suite/TestModules';
+import { createQueryString } from 'test-suite/screens/getScreenIdForLinking';
+
+// Legacy deep-link form that runs every test, e.g. `bareexpo://test-suite/all`.
+export default function All() {
+  const tests = createQueryString(getTestModules().map((module) => module.name));
```
Ubax, line 8: Same thing about +native-intent

apps/bare-expo/app/_layout.tsx
```diff
@@ -70,11 +83,36 @@ Observe.configure({
...
+    AppMetrics.markInteractive({
+      params: {
+        theme: themeName,
+      },
+    });
+  }, []);
+
+  return (
+    <GestureHandlerRootView style={{ flex: 1 }}>
+      <NavigationThemeProvider value={themeName === 'dark' ? DarkTheme : DefaultTheme}>
```
Ubax, line 104: WDYT about adding a todo to unify these themes with `ThemeProvider`?

apps/bare-expo/index.js
Ubax: Can we remove this file and use `entry` in package.json?

apps/native-component-list/src/app/(main)/apis/expo-observe/filtered/[userId]/[accountId].tsx
Comment (Ubax) [reply apps/native-component-list/src/app/(main)/apis/expo-observe/filtered/[userId]/[accountId].tsx]: I don't think this is needed anymore. We have an observe-tester app, which also uses expo-router. I added this screen to bare-expo, only to test the react-navigation integration. We can safely remove it

apps/native-component-list/src/app/(main)/components/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { Layout } from '../../../constants';
+import { screenApiItems } from '../../../navigation/componentScreens';
+import ExpoComponents from '../../../screens/ExpoComponentsScreen';
+
+export default function ComponentsIndex() {
+  return (
+    <>
+      <Stack.Screen
+        options={{
+          title: Layout.isSmallDevice ? 'Expo SDK Components' : 'Components in Expo SDK',
```
Ubax, line 13: Just curious, do you know why do we need this check?

apps/native-component-list/src/app/+not-found.tsx
Ubax, line 1: I think this is not needed. This should be displayed by default if I am not mistaken

Ubax (approved):

apps/native-component-list/src/screens/ExpoObserveScreen.tsx
```diff
@@ -44,65 +25,13 @@ function IndexScreen({
...
-          navigation.navigate('filteredParams', {
-            p1: 'p1',
-            p2: 'p2',
-            onPress: () => {},
-          })
-        }>
-        <Text style={[styles.button, { color: theme.text.link }]}>
-          Navigate with function param
-        </Text>
+        onPress={() => router.push('/apis/expo-observe/filtered/7/acct-99?p1=p1&p2=p2')}>
```
Ubax: [🏗️ **design**] Both buttons on this screen regress in bare-expo. The dedicated route `apis/expo-observe/filtered/[userId]/[accountId]` only exists in NCL's own `src/app` tree — bare-expo's app dir has just the catch-all `apis/[...id]`, so this push (and the `Linking.openURL` above) resolves to `findApiScreen('expo-observe/filtered/7/acct-99')` → `undefined` → `<Unmatched />`. Previously this worked in bare-expo because the screen embedded its own stack navigator and the registry entry carried the nested `linking` config (both removed here). Since bare-expo is where the observe integration is primarily exercised and there's no e2e coverage of this link, consider adding a mirror route under `apps/bare-expo/app/(tabs)/apis/expo-observe/filtered/[userId]/[accountId].tsx` (with the usual `optionalRequire` wrapper).

apps/test-suite/app/select/[tests].tsx
```diff
@@ -0,0 +1,12 @@
+import { Redirect, useLocalSearchParams } from 'expo-router';
+import * as React from 'react';
+
+import { createQueryString, getSelectedTestNames } from '../../screens/getScreenIdForLinking';
+
+// Legacy deep-link form that runs a set of tests, e.g. `test-suite://select/basic,crypto`.
+export default function SelectTests() {
+  const { tests } = useLocalSearchParams<{ tests?: string | string[] }>();
+  const raw = Array.isArray(tests) ? tests.join(',') : (tests ?? '');
+  const query = createQueryString(getSelectedTestNames(decodeURIComponent(raw)));
```
Comment (Ubax) [inline apps/test-suite/app/select/[tests].tsx]: [💡 suggestion] `useLocalSearchParams` already returns percent-decoded values, so this is a double decode — and `decodeURIComponent` throws `URIError` on stray `%` (e.g. `test-suite://select/100%`), crashing the route render on attacker/user-controlled deep-link input. The old handler only decoded the `tests=` query match, never the `/select/` path segment. `getSelectedTestNames(raw)` should be enough here. Same pattern in `apps/bare-expo/app/(tabs)/test-suite/select/[tests].tsx`.

apps/bare-expo/app/(tabs)/_layout.tsx
```diff
@@ -0,0 +1,33 @@
+import { NativeTabs } from 'expo-router/unstable-native-tabs';
+import * as React from 'react';
+
+import { optionalRequire } from '../../optionalRequire';
+
+// The apis/components tabs render screens from `native-component-list`;
+// hide them when the package isn't bundled.
+const hasNativeComponentList = !!optionalRequire(() =>
```
Ubax, line 8: [🐧 nit] The NCL-optional tolerance this preserves is undermined by the always-visible playground tab: `app/(tabs)/playground.tsx` re-exports `Playground.tsx`, which statically imports `native-component-list/src/components/TabIcon`. Without NCL that import fails at bundle time regardless of these hidden triggers. Pre-existing (the old `MainNavigator` had the same chain), so fine to punt — but if NCL-less builds are meant to work, Playground should render its icon from `@expo/vector-icons` directly.

apps/native-component-list/src/screens/ExpoObserveScreen.tsx
Ubax: I would suggest just removing this screen entirely. I can do it as a PR before this one, if this helps

Ubax (commented):
More comments after automated review

apps/bare-expo/index.js
Kudo: Done in 8145ea8 — `main` is now `expo-router/entry` and `index.js` is gone. The native side loads through `.expo/.virtual-metro-entry`, which reads the `main` field, so no native change was needed.

apps/bare-expo/app/(tabs)/test-suite/select/[tests].tsx
Comment (Kudo) [reply apps/bare-expo/app/(tabs)/test-suite/select/[tests].tsx:6]: Went back and forth on this. `redirectSystemPath` receives the raw launch URL (scheme + authority), before the router normalizes it. The legacy keyword also lands in a different URL part per app: standalone `test-suite://select/basic,crypto` puts `select` in the **host**, while bare-expo `bareexpo://test-suite/select/basic,crypto` puts it in a **path segment**. A single `+native-intent` would need per-app, per-form string parsing plus manual percent-decoding (the same double-decode footgun fixed in the thread below). The `select/[tests].tsx` redirect route lets the router normalize the URL, read typed params, and stay covered by the deep-link jest tests + Maestro e2e. So I'd keep the redirect routes for both bare-expo and standalone test-suite. Happy to revisit if we want to centralize the rewriting later.

apps/bare-expo/app/(tabs)/test-suite/all.tsx
Kudo: Same reasoning as the `select` thread — keeping this redirect route. `redirectSystemPath` gets the raw launch URL before normalization, and the `all` keyword sits in the host (standalone test-suite) vs a path segment (bare-expo), so a shared `+native-intent` would need per-app string parsing. The redirect route stays typed and e2e-covered.

apps/native-component-list/src/app/(main)/apis/expo-observe/filtered/[userId]/[accountId].tsx
Comment (Kudo) [reply apps/native-component-list/src/app/(main)/apis/expo-observe/filtered/[userId]/[accountId].tsx]: Done in 8145ea8 — removed the ExpoObserve screen, this filtered-params route, and the registry entry entirely. observe-tester covers the integration now, as you noted.

apps/native-component-list/src/navigation/StackConfig.tsx
Kudo: Added `TODO(kudo,20260802)` in 8145ea8 to convert this into `Stack.Toolbar.Right` once we adopt the native toolbar API.

apps/bare-expo/app/(tabs)/apis/index.tsx
Kudo: Went with the `_layout.tsx` route — 8145ea8 sets the title via `<Stack.Screen name="index" options={{ title: 'APIs in Expo SDK' }} />` in `apis/_layout.tsx`, so `index.tsx` is now just `return ExpoApis ? <ExpoApis apis={screenApiItems} /> : null`.

apps/bare-expo/app/(tabs)/components/[...id].tsx
Comment (Kudo) [reply apps/bare-expo/app/(tabs)/components/[...id].tsx]: Done in 8145ea8 — extracted `OptionalRegistryScreenRoute` in bare-expo. Both catch-all routes are now one-liners, e.g. `return <OptionalRegistryScreenRoute find="findApiScreen" />`.

apps/bare-expo/app/(tabs)/components/index.tsx
Kudo: Same as APIs — 8145ea8 moves the title into `components/_layout.tsx` and simplifies `index.tsx`.

apps/bare-expo/app/_layout.tsx
Kudo: Added `TODO(kudo,20260802)` in 8145ea8 to unify these navigation themes with `ThemeProvider`.

apps/native-component-list/src/app/(main)/components/index.tsx
Kudo: It's carried over from the pre-expo-router NCL header — this conditional has been there since NCL moved into the repo. The short "Expo SDK Components" fits the header on small devices; larger devices get "Components in Expo SDK". I kept it to preserve the existing look.

apps/native-component-list/src/app/+not-found.tsx
Kudo: Done in 8145ea8 — removed. expo-router auto-appends a default `+not-found` (the `Unmatched` screen) via `appendNotFoundRoute`, so the re-export was redundant. You were right.

apps/native-component-list/src/screens/ExpoObserveScreen.tsx
Kudo: Done in 8145ea8 — removed the ExpoObserveScreen entirely (screen, filtered route, and registry entry), so this regression is gone. observe-tester covers the integration now.

apps/test-suite/app/select/[tests].tsx
Comment (Kudo) [reply apps/test-suite/app/select/[tests].tsx]: Good catch — fixed in 8145ea8. Dropped `decodeURIComponent` in both `apps/test-suite/app/select/[tests].tsx` and `apps/bare-expo/app/(tabs)/test-suite/select/[tests].tsx`. `useLocalSearchParams` already returns decoded values.

apps/bare-expo/app/(tabs)/_layout.tsx
Kudo: Fixed in 8145ea8 — Playground no longer imports NCL's `TabIcon`; it renders a plain emoji, so the always-present Playground tab has no native-component-list dependency. (bare-expo doesn't depend on `@expo/vector-icons` directly, so an icon font would have meant a new dependency.)

apps/native-component-list/src/screens/ExpoObserveScreen.tsx
```diff
@@ -44,65 +25,13 @@ function IndexScreen({
...
+        <Text style={[styles.button, { color: theme.text.link }]}>Navigate with params</Text>
       </TouchableOpacity>
```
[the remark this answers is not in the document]
Kudo: Resolved in 8145ea8 by removing the ExpoObserve screen entirely rather than mirroring the route into bare-expo. observe-tester exercises the observe integration now, so the demo screen isn't needed in bare-expo/NCL.

PR #47526, opened 2026-07-05
[expo-router] Guard deep link decode against malformed percent-encoding

### LINKED ISSUE #47525: expo-router: malformed percent-encoding in deep link throws URIError from extractPathFromURL (unhandled rejection / silently dropped link)
### Summary

`extractExactPathFromURL` in `expo-router` decodes deep-link query values with a **bare `decodeURIComponent` / `decodeURI`** — no try/catch. A deep link whose query contains malformed percent-encoding (e.g. `myapp:///path?q=%GG` or a truncated multi-byte sequence `?q=%E0%A4%A`) throws `URIError: Malformed decodeURI input`.

Because the link is processed inside expo-router's **async** `'url'` event listener (and the async `getInitialURL().then(...)` cold-start path), the throw surfaces as an **unhandled promise rejection**, not a catchable error at the call site:

- **Dev:** a LogBox error toast (`URIError: Malformed decodeURI input`).
- **Release:** no visible crash, but the deep link is **silently dropped** — the app foregrounds and never navigates, and nothing is reported (RN release builds don't track unhandled rejections). This is arguably worse than the old hard crash because it's invisible.

Deep links arrive from external, un-sanitized sources — widgets, push notifications, other apps, hand-typed URLs — so a malformed one should not be able to trigger a dev toast or a silent no-op. The whole app's linking is affected, not one route.

### Which package

`expo-router`

Any expo-router app with a `[param]` or `?query`-consuming route. With the app installed on a simulator:

```sh
# malformed percent-encoding — throws URIError
xcrun simctl openurl booted "myapp:///somepath?q=%GG"
# truncated multi-byte sequence — also throws
xcrun simctl openurl booted "myapp:///somepath?q=%E0%A4%A"

# Android
adb shell am start -W -a android.intent.action.VIEW -d "myapp://somepath?q=%GG"
```

Expected: the link resolves (ideally passing the raw/undecodable value through to the route), no error.
Actual: `URIError: Malformed decodeURI input` — LogBox toast in dev, silently dropped link in release.

### Root cause

Two bare decode calls in `packages/expo-router/src/fork/extractPathFromURL.ts` (permalinks at current `main`, `64f1459`):

- Query values: https://github.com/expo/expo/blob/64f145964a34740707330aa4cbae0bd9fc12cdad/packages/expo-router/src/fork/extractPathFromURL.ts#L118
  ```ts
  [...res.searchParams.entries()].map(([k, v]) => `${k}=${decodeURIComponent(v)}`).join('&');
  ```
- Dev-client `?url=` path: https://github.com/expo/expo/blob/64f145964a34740707330aa4cbae0bd9fc12cdad/packages/expo-router/src/fork/extractPathFromURL.ts#L102
  ```ts
  return extractExactPathFromURL(decodeURI(incomingUrl));
  ```

A malformed sequence makes either call throw. Since a version bump doesn't help (the bare calls are on the current `main`), apps can only work around it with a `patch-package` patch.

### Inconsistency with the package's own code

`useLocalSearchParams` already guards the *identical* decode defensively — malformed values fall through to the raw string instead of throwing:

https://github.com/expo/expo/blob/64f145964a34740707330aa4cbae0bd9fc12cdad/packages/expo-router/src/hooks/useLocalSearchParams.ts#L63-L77

```ts
try {
  return [key, decodeURIComponent(value as string)];
} catch {
  return [key, value];
}
```

`extractPathFromURL.ts` just never got the same treatment.

### Suggested fix

Wrap both decode sites in the same try/catch-with-raw-fallback pattern `useLocalSearchParams` already uses:

```ts
function decodeURIComponentOrRaw(value: string): string {
  try { return decodeURIComponent(value); } catch { return value; }
}
function decodeURIOrRaw(value: string): string {
  try { return decodeURI(value); } catch { return value; }
}
```

This makes malformed links non-throwing regardless of the separate correctness question raised in #42075 (whether `decodeURIComponent` on `searchParams.entries()` — which are already decoded — belongs there at all, given it was changed away from `searchParams.toString()` in #25198). Even if that line is ultimately reworked, the robustness guard is orthogonal and worth having on both decode paths.

### Related prior issues

- #25323 (`Issue accepted`, closed as completed) — the same `%25` deep-link crash; the SDK-49-era hard RedBox has softened, but the underlying bare decode remains.
- #26034 — same `URIError: Malformed decodeURI input`, via the `push` function.
- #42075 (closed stale, 2026) — flags this exact line 118; a maintainer noted `decodeURIComponent` there "seems wrong" and wanted to understand #25198 before changing it, then it lapsed for inactivity. The auto-close comment invited opening a new issue for related problems — hence this report, framed around robustness rather than the double-decode correctness angle.

### Environment

```
expo-router: 55.0.12 (also present on main / 57.x tip)
expo:        55.0.14 (SDK 55)
react-native: 0.83.2 (New Architecture)
Platforms:   iOS + Android (native deep links)
```

momomuchu:
decodeURI/decodeURIComponent in extractExactPathFromURL throw on malformed percent-encoding (e.g. %GG), which surfaces as an unhandled rejection in the url listener and silently drops the link in release builds. Wraps both decode sites in the same try/catch-with-raw-fallback useLocalSearchParams already uses.

Fixes #47525

### DIFF
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -24,6 +24,7 @@
 - [ios] Fix white flash behind screens during the interactive swipe-back gesture by forwarding the theme background to the native stack container. ([#47121](https://github.com/expo/expo/pull/47121) by [@kevenleone](https://github.com/kevenleone))
 - [ios] Fix memory leaks in the native toolbar and link preview from strong-reference cycles and closures that strongly captured `self`. ([#47378](https://github.com/expo/expo/pull/47378) by [@Ubax](https://github.com/Ubax))
 - [android] Remove navigation state restoration across activity recreation. ([#47422](https://github.com/expo/expo/pull/47422) by [@Ubax](https://github.com/Ubax))
+- Guard the deep link decode in `extractExactPathFromURL` against malformed percent-encoding, falling back to the raw value instead of throwing `URIError` and dropping the link. ([#47526](https://github.com/expo/expo/pull/47526) by [@momomuchu](https://github.com/momomuchu))
 
 ### 💡 Others
 
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -24,7 +24,7 @@
 - [ios] Fix white flash behind screens during the interactive swipe-back gesture by forwarding the theme background to the native stack container. ([#47121](https://github.com/expo/expo/pull/47121) by [@kevenleone](https://github.com/kevenleone))
 - [ios] Fix memory leaks in the native toolbar and link preview from strong-reference cycles and closures that strongly captured `self`. ([#47378](https://github.com/expo/expo/pull/47378) by [@Ubax](https://github.com/Ubax))
 - [android] Remove navigation state restoration across activity recreation. ([#47422](https://github.com/expo/expo/pull/47422) by [@Ubax](https://github.com/Ubax))
-- Guard the deep link decode in `extractExactPathFromURL` against malformed percent-encoding, falling back to the raw value instead of throwing `URIError` and dropping the link. ([#47526](https://github.com/expo/expo/pull/47526) by [@momomuchu](https://github.com/momomuchu))
+- Guard the deep link decode in `extractExactPathFromURL` against malformed percent-encoding. ([#47526](https://github.com/expo/expo/pull/47526) by [@momomuchu](https://github.com/momomuchu))
 
 ### 💡 Others
 
--- a/packages/expo-router/src/fork/extractPathFromURL.ts
+++ b/packages/expo-router/src/fork/extractPathFromURL.ts
@@ -99,7 +99,14 @@ function fromDeepLink(url: string): string {
       return '';
     }
     const incomingUrl = res.searchParams.get('url')!;
-    return extractExactPathFromURL(decodeURI(incomingUrl));
+    let decodedIncomingUrl: string;
+    try {
+      decodedIncomingUrl = decodeURI(incomingUrl);
+    } catch {
+      // Malformed percent-encoding (e.g. `%GG`) would otherwise throw and drop the link.
+      decodedIncomingUrl = incomingUrl;
+    }
+    return extractExactPathFromURL(decodedIncomingUrl);
   }
 
   let results = '';
@@ -115,7 +122,16 @@ function fromDeepLink(url: string): string {
   const qs = !res.search
     ? ''
     : // @ts-ignore: `entries` is not on `URLSearchParams` in some typechecks.
-      [...res.searchParams.entries()].map(([k, v]) => `${k}=${decodeURIComponent(v)}`).join('&');
+      [...res.searchParams.entries()]
+        .map(([k, v]) => {
+          try {
+            return `${k}=${decodeURIComponent(v)}`;
+          } catch {
+            // Malformed percent-encoding (e.g. `%GG`) would otherwise throw and drop the link.
+            return `${k}=${v}`;
+          }
+        })
+        .join('&');
 
   if (qs) {
     results += '?' + qs;
--- a/packages/expo-router/src/fork/extractPathFromURL.ts
+++ b/packages/expo-router/src/fork/extractPathFromURL.ts
@@ -1,3 +1,5 @@
+import { safeDecodeURI, safeDecodeURIComponent } from '../utils/url';
+
 export function parsePathAndParamsFromExpoGoLink(url: string): {
   pathname: string;
   queryString: string;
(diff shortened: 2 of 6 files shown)


packages/expo-router/src/fork/extractPathFromURL.ts
```diff
@@ -99,7 +99,14 @@ function fromDeepLink(url: string): string {
...
     }
     const incomingUrl = res.searchParams.get('url')!;
-    return extractExactPathFromURL(decodeURI(incomingUrl));
+    let decodedIncomingUrl: string;
+    try {
+      decodedIncomingUrl = decodeURI(incomingUrl);
+    } catch {
+      // Malformed percent-encoding (e.g. `%GG`) would otherwise throw and drop the link.
+      decodedIncomingUrl = incomingUrl;
+    }
```
Ubax: Can you extract this into a function - `safelyDecodeURI`?

packages/expo-router/src/fork/extractPathFromURL.ts
```diff
@@ -115,7 +122,16 @@ function fromDeepLink(url: string): string {
...
     : // @ts-ignore: `entries` is not on `URLSearchParams` in some typechecks.
-      [...res.searchParams.entries()].map(([k, v]) => `${k}=${decodeURIComponent(v)}`).join('&');
+      [...res.searchParams.entries()]
+        .map(([k, v]) => {
+          try {
+            return `${k}=${decodeURIComponent(v)}`;
+          } catch {
+            // Malformed percent-encoding (e.g. `%GG`) would otherwise throw and drop the link.
+            return `${k}=${v}`;
+          }
```
Ubax: Can you extract try/catch into a `safelyDecodeURIComponent` function?

Ubax (commented):
Thanks for opening the PR to solve this issue. I added two small nit comments. 

When you add them please add unit tests for them as well

packages/expo-router/src/fork/extractPathFromURL.ts
```diff
@@ -69,6 +69,23 @@ function isExpoDevelopmentClient(url: URL): boolean {
...
+  }
+}
+
+export function safelyDecodeURIComponent(uriComponent: string): string {
+  try {
+    return decodeURIComponent(uriComponent);
+  } catch {
+    return uriComponent;
+  }
+}
```
Ubax: Sorry for the confusion. Can you please use `safeDecodeURIComponent` which already exists in `packages/expo-router/src/global-state/getRouteInfoFromState.ts`? 

Please extract it to `packages/expo-router/src/utils/url.ts` and add `safeDecodeURI` there as well.

Ubax (commented):
Sorry for the confusion. I only now realized that `safeDecodeURIComponent` already exists in the code-base.

packages/expo-router/src/fork/extractPathFromURL.ts
momomuchu: Done in `520ab69`. Moved `safeDecodeURIComponent` out of `getRouteInfoFromState.ts` into `packages/expo-router/src/utils/url.ts`, added `safeDecodeURI` next to it, and pointed `extractPathFromURL.ts` at the shared helpers instead of its own copies. Tests moved to `utils/__tests__/url.test.ios.ts`.

packages/expo-router/CHANGELOG.md
```diff
@@ -24,6 +24,7 @@
 - [ios] Fix white flash behind screens during the interactive swipe-back gesture by forwarding the theme background to the native stack container. ([#47121](https://github.com/expo/expo/pull/47121) by [@kevenleone](https://github.com/kevenleone))
 - [ios] Fix memory leaks in the native toolbar and link preview from strong-reference cycles and closures that strongly captured `self`. ([#47378](https://github.com/expo/expo/pull/47378) by [@Ubax](https://github.com/Ubax))
 - [android] Remove navigation state restoration across activity recreation. ([#47422](https://github.com/expo/expo/pull/47422) by [@Ubax](https://github.com/Ubax))
+- Guard the deep link decode in `extractExactPathFromURL` against malformed percent-encoding, falling back to the raw value instead of throwing `URIError` and dropping the link. ([#47526](https://github.com/expo/expo/pull/47526) by [@momomuchu](https://github.com/momomuchu))
```
```suggestion
- Guard the deep link decode in `extractExactPathFromURL` against malformed percent-encoding. ([#47526](https://github.com/expo/expo/pull/47526) by [@momomuchu](https://github.com/momomuchu))
```

Ubax (approved):

Ubax:
Thanks for your contribution 🎉 

I'll merge the PR when the CI is green

packages/expo-router/CHANGELOG.md
Ubax: I'll apply the suggestion before merging

PR #47533, opened 2026-07-05
[widgets][iOS] Changing widget configuration at runtime

jakex7:
# Why

Some apps may require widget configuration to be provided at runtime. For example, Slack can provide the list of workspaces after the user has authenticated.

# How

Add a `dynamic` option to the `app.json` configuration, that will generate a `DynamicOptionsProvider` instead of a standard enum.

The available values can then be provided at runtime using `setConfigurationParameterEnum(parameterName, options)`.

For example:

```tsx
MyWidget.setConfigurationParameterEnum('workspace', [
  { name: 'Current Workspace', value: 'current', subtitle: 'Follows the app' },
  { name: 'Personal', value: 'personal' },
  { name: 'Expo', value: 'expo' },
  { name: 'SWM.', value: 'swm' },
]);
```

# Test Plan

https://github.com/user-attachments/assets/0a8ad5c6-8702-4f0c-aadb-5ff3fa4ad5c4

# Checklist

### DIFF
--- a/docs/pages/versions/unversioned/sdk/widgets.mdx
+++ b/docs/pages/versions/unversioned/sdk/widgets.mdx
@@ -126,7 +126,7 @@ You can configure `expo-widgets` using its built-in [config plugin](/config-plug
         'Makes the widget [configurable](#configurable-widgets). The values the user picks are passed to your widget at runtime through `environment.configuration`. An object with:',
         '* `title` - The title shown when the user edits the widget.',
         '* `description` - An optional description shown when the user edits the widget.',
-        '* `parameters` - A map of parameter keys to parameter definitions. Each parameter has a `title`, a `type` (`string`, `number`, `boolean`, or `enum`), and a `default`. An `enum` parameter additionally takes a `values` array of `{ name, value }` options.',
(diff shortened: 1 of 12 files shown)


docs/pages/versions/unversioned/sdk/widgets.mdx
```diff
@@ -464,6 +465,18 @@ const WeatherWidget = (
 export default createWidget<WeatherProps, WeatherConfiguration>('WeatherWidget', WeatherWidget);
 ```
 
+#### Dynamic enum options
+
+Set `dynamic: true` on an enum parameter when the list of choices is only known at runtime, such as workspaces loaded after sign-in. The `values` array in app config is used as the fallback list until the app provides runtime options.
```
amandeepmittal:
```suggestion
Set `dynamic: true` on an enum parameter when the list of choices is only known at runtime, such as workspaces loaded after sign-in. The `values` array in app config is used as the fallback list until the app provides runtime options as shown below:
```

It'll be good to mention that `dynamic` is set in app config file here.

amandeepmittal (changes requested):

lukmccall (approved):

amandeepmittal (approved):

PR #47540, opened 2026-07-05
fix(expo-sqlite): make SQLiteStorage migration idempotent to heal sync/async race

### LINKED ISSUE #47448: expo-sqlite/kv-store: getDbSync() bypasses the awaitLock, racing async access and corrupting the migration (no such table: storage)

https://github.com/ayozebarrera/expo-kvstore-race-repro

### Steps to reproduce

Platform(s): iOS (kv-store used as an AsyncStorage-style store).

What happens. SQLiteStorage serializes its async init with awaitLock (added in #33834, which closed #33754), but the sync path does not. From main:

```ts
private getDbSync(): SQLiteDatabase {
  if (!this.db) {
    const db = openDatabaseSync(this.databaseName);
    this.maybeMigrateDbSync(db);   // no awaitLock, unlike getDbAsync()
    this.db = db;
  }
  return this.db;
}
```

So getDbSync() and getDbAsync() can run their migration transactions concurrently on the same file with no shared lock. Since maybeMigrateDb* only creates the table while user_version === 0 and then bumps the version, a raced/interrupted first run can commit user_version = 1 without an effective CREATE TABLE storage. After that the table is never recreated, so every getItemSync/setItemSync throws no such table: storage.

Image

In our app the default Storage instance was shared by an i18n cache (async) and a small env store (sync getItemSync) that runs on the first request at cold start; that was enough to corrupt it. It only affected 

In our app the default Storage instance was shared by an i18n cache (async) and a small env store (sync getItemSync) that runs on the first request at cold start; that was enough to corrupt it. It only affected users opening the app with an existing session (which forces the early sync call), was intermittent, and survived reinstalls on iOS (backup restore of the file + auth token surviving in the Keychain re-triggering the early sync call).

Steps:
1. Use a SQLiteStorage (or default Storage) whose database is touched by both the sync and async APIs during first launch.
2. Cold-start so a sync getItemSync and an async getItem race the first-run migration.
3. (Deterministic equivalent: run the snippet above.)

Expected: sync/async init are mutually exclusive (or migration is idempotent/self-healing), so a raced first launch can't leave user_version bumped without the storage table.

Actual: DB left with user_version = 1 and no storage table → permanent SQLiteErrorException: Error code 1: no such table: storage (surfaced via prepareSync/finalizeSync → FunctionCallException).

Suggested fixes (any one):
1. Guard getDbSync() init against a concurrent async init (sync-visible flag or sync mutex around init + migration).
2. Make maybeMigrateDb* resilient: even when user_version >= DATABASE_VERSION, verify the storage table exists and CREATE TABLE IF NOT EXISTS it if missing (self-heal).
3. Eagerly init the connection in the constructor (as suggested in #33754).

Related: #33754 (closed via #33834) fixed the async path only; getDbSync remains unlocked on main.

### Environment

```text
❯ npx expo-env-info

    System:
      OS: macOS 26.5.1
    Binaries:
      Node: 24.16.0 - /Users/abarrera/.nvm/versions/node/v24.16.0/bin/node
      npm: 11.13.0 - /Users/abarrera/.nvm/versions/node/v24.16.0/bin/npm
      Watchman: 2025.02.17.00 - /opt/homebrew/bin/watchman
    Managers:
    SDKs:
      iOS SDK:
    IDEs:
      Android Studio: 2024.2 AI-242.23726.103.2422.12816248
    npmPackages:
      expo: ~57.0.1 => 57.0.1 
      react: 19.2.3 => 19.2.3 
    Expo Workflow: managed
```

### Expo Doctor Diagnostics

```text
❯ npx expo-doctor@latest
```

tsushanth:
Fixes #47448.

## Root cause

`getDbSync()` and `getDbAsync()` can both trigger the first-run migration concurrently. `getDbAsync()` serializes through `awaitLock` (added in #33834), but `getDbSync()` has no shared lock with the async path. When they race:

1. `getDbAsync()` starts `withTransactionAsync`: reads `user_version = 0`, executes `CREATE TABLE IF NOT EXISTS storage`, bumps `user_version = 1`, commits.
2. `getDbSync()` starts `withTransactionSync` concurrently: depending on timing it may read `user_version = 1` from the snapshot already committed by step 1, hit the early-return guard (`currentDbVersion >= DATABASE_VERSION`), **skip `CREATE TABLE`**, and complete.

This leaves the database with `user_version = 1` but no `storage` table. All subsequent `getItemSync`/`setItemSync`/`getItemAsync`/… calls throw `no such table: storage` permanently — the version guard ensures migration is never retried.

## Fix

Remove the early-return guard and unconditionally run `MIGRATION_STATEMENT_0`. The statement uses `IF NOT EXISTS`, so it is always a no-op when the table already exists. The `PRAGMA user_version` bump remains conditional on being below `DATABASE_VERSION`.

Both `maybeMigrateDbSync` and `maybeMigrateDbAsync` receive the same treatment:
- **Sync**: prevents future races.
- **Async**: self-heals any database already corrupted by the race on a user's device (avoids requiring a reinstall).

### DIFF
--- a/packages/expo-sqlite/src/Storage.ts
+++ b/packages/expo-sqlite/src/Storage.ts
@@ -384,14 +384,15 @@ export class SQLiteStorage {
     return db.withTransactionAsync(async () => {
       const result = await db.getFirstAsync<{ user_version: number }>('PRAGMA user_version');
       let currentDbVersion = result?.user_version ?? 0;
-      if (currentDbVersion >= DATABASE_VERSION) {
-        return;
+      // Always run the migration statement even when user_version is already at
+      // DATABASE_VERSION. MIGRATION_STATEMENT_0 uses IF NOT EXISTS, so it is a
+      // no-op when the table exists. This self-heals a DB that was left with
+      // user_version = 1 but no storage table (possible when getDbSync() raced
+      // a concurrent getDbAsync() call before this lock was introduced).
+      await db.execAsync(MIGRATION_STATEMENT_0);
+      if (currentDbVersion < DATABASE_VERSION) {
+        await db.execAsync(`PRAGMA user_version = ${DATABASE_VERSION}`);
       }
-      if (currentDbVersion === 0) {
-        await db.execAsync(MIGRATION_STATEMENT_0);
-        currentDbVersion = 1;
-      }
-      await db.execAsync(`PRAGMA user_version = ${DATABASE_VERSION}`);
     });
   }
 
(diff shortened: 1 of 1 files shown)


tsushanth:
Closing — this has been open without activity long enough that it's more useful to withdraw and let a fresh PR reflect current `main`/`master` than to keep pinging for review on stale state. Happy to reopen or resubmit if there's still interest.

Issue #866, opened 2017-10-27
Building for iOS errors out without information


pouriaa:
When building an expo project for iOS, the build initiates sucessfully using `exp build:ios`. When checking on the status after a few seconds using `exp build:status`, I get the following response:

```
[exp] Checking if current build exists...

[exp] ============
[exp] Build Status
[exp] ============

[exp] iOS: There was an error with this build.
```

I tried the suggested solution in this issue with no success https://github.com/expo/expo/issues/836

More info:
My Build ID: 4ac607d1-87ae-4f2c-8128-a6aa4b4e7102

package.json
```
{
  "name": "myapp",
  "version": "0.1.0",
  "private": true,
  "devDependencies": {
    "babel-plugin-transform-async-to-generator": "^6.24.1",
    "jest-expo": "~20.0.0",
    "react-native-scripts": "1.3.1",
    "react-test-renderer": "16.0.0-alpha.12"
  },
  "main": "./node_modules/react-native-scripts/build/bin/crna-entry.js",
  "scripts": {
    "start": "react-native-scripts start",
    "eject": "react-native-scripts eject",
    "android": "react-native-scripts android",
    "ios": "react-native-scripts ios",
    "test": "node node_modules/jest/bin/jest.js --watch"
  },
  "jest": {
    "preset": "jest-expo"
  },
  "dependencies": {
    "expo": "^21.0.0",
    "react": "16.0.0-alpha.12",
    "react-native": "^0.48.0",
    "react-native-elements": "^0.17.0",
    "react-navigation": "^1.0.0-beta.11"
  }
}
```

app.json
```
{
   "expo": {
    "name": "My app",
    "icon": "./app/assets/img/logo.png",
    "version": "1.0.0",
    "slug": "tolio",
    "sdkVersion": "21.0.0",
    "ios": {
      "bundleIdentifier": "com.pouriaassadipour.tolio"
    },
    "android": {
      "package": "com.yourcompany.yourappname"
    }
   }
 }
```

tcdavis:
Can you use imagemagick or something to check your app icon / other assets and ensure that the formats match the file extensions?

pouriaa:
@tcdavis I sent the app icon that I'm using to Adam on Slack and he couldn't see anything wrong with it. He said he forwarded my case to the iOS engineer last night. Haven't heard back yet though.

pouriaa:
using imagemagick's `identify` command I found the following details:
```
[...]

Issue #859, opened 2017-10-26
Audio loop not behaving how it should 


shinedark:
Need some help with expo audio setIsLoopingAsync. it works but it skips part of the audio track after the 1st loop. so let’s say i have a loop of a kick drum. It has 8 hits in the loop. when it loops it only plays 7 of them.Then their is the issue when running multiple sounds at a time they don't play how they should they get stop and start at random times while still trying to play.

i test it in ios simulator and iphone. both had the same issues 

if some one can help would be great thanks.

here is some basic code to recreate the issue and the repo 
https://github.com/shinedark/LoopNotWorking

`import React from 'react';
import { Text, View  } from 'react-native';
import { CardSection, Button, Button2 } from './src/components/common';
import { Audio } from 'expo';

export default class App extends React.Component {
  
    state = {
      playing1: false
    };

    async componentWillMount(){
      
      try{
        await Expo.Audio.setIsEnabledAsync(true);
      } catch(error) {console.log(error);}
    }

    

    handlePLay1 = async () => {
      const { playing1 } = this.state;
      this.setState({ playing1: true});
        const soundObject = new Expo.Audio.Sound();
          try {
            await soundObject.loadAsync(require('./sounds/drums/kick.wav'));
          { shouldPlay: true }
          this.audioPlayer1  = soundObject;
             this.audioPlayer1.setIsLoopingAsync(true);
             this.audioPlayer1.playAsync();
             this.audioPlayer1.setVolumeAsync(0.8);
           // Your sound is playing!
          } catch (error) {
          // An error occurred!
          
          } 
    }

    stopMusic1 = async () => {
      const { playing1 } = this.state;
      this.setState({ playing1: false});
        await this.audioPlayer1.stopAsync();

     }  

    renderButton1 () {
       
      if (this.state.playing1){
        return  (
          <View style={styles.col}>
            <Button2 style={styles.bol} onPressIn={this.stopMusic1.bind(this)}>
              ||
            </Button2>
          </View>
        );

      }
      return(
        
        <View style={styles.col}> 
          <Button style={styles.col1} onPressIn={this.handlePLay1.bind(this)}>
            |>
          </Button>
        </View>
      );
     }
    render(){
      
      return (
        <View style={{flex:1}}>
          <View style={styles.col}>
            {this.renderButton1()}
          </View>
        </View>

      );

    }
}

const styles =  {
  col: {
    flex: 2,
    flexDirection: 'row'
  }, 
  col1:{
[...]

Issue #851, opened 2017-10-25
Video component not loading files from CameraRoll.getPhotos


mfoncho:
The Video component mounts but doesn't load the video file from the file system
after inspecting the component i found no errors were thrown, this issue persisted across platforms 
(both IOS & Android) 
Tried the same code with the Image component and mimeTypes of ['image/jpeg'] and it worked just fine
Can someone please tell me how to get this rolling / How to pass a local resource form CameraRoll.getPhotos to the Video component for playback

-> Expo sdk : 21.0.0

-----------------------------------------    BUGGY SAMPLE APP   --------------------------------------------

```javascript
import React, { Component } from 'react'
import {Video} from 'expo';
import { CameraRoll, StyleSheet, FlatList, View } from 'react-native'
const styles = StyleSheet.create({
    video: {
        width: 300,
        heigh: 300,
    },
    screen:{
         flex:1,
    }
  
});

class SampleApp extends Component {
    state = {
        videos:[],
    }
   componentWillMount(){
        CameraRoll.getPhotos({ first:100, asssetType:'Videos', mimeTypes:['video/mp4']}).then( (data) =>{
            this.setState({videos:data.edges})
        })
    }
    render(){
        return(
            <View style={styles.screen}>
                <FlatList
                    data={this.state.videos}
                    keyExtractor={ (n, i) => i }
                    renderItem={ ({item})=>(
                        <Video 
                            source={{uri:item.node.image.uri}}  
                            style={styles.video} 
                            resizeMode={Video.RESIZE_MODE_COVER}/>
                    )}
                />
           </View>
        );
    }
}
```

sjchmiela:
Hi, @mfoncho! I've just tested it on SDK 23 and it looks like it works just fine! Try the following application:

```js
import React from 'react';
import { Video } from 'expo';
import { StyleSheet, CameraRoll, FlatList } from 'react-native';

export default class App extends React.Component {
  state = {
    videoNodes: [],
  };

  componentDidMount() {
    CameraRoll.getPhotos({ first: 10, assetType: CameraRoll.AssetTypeOptions.Videos }).then(
      this._videosLoaded
    );
  }

  _videosLoaded = ({ edges }) => this.setState({ videoNodes: edges });

  _keyExtractor = (item, index) => item.node.image.uri;

  _renderVideo = ({ item }) => (
    <Video source={{ uri: item.node.image.uri }} useNativeControls style={styles.video} />
  );

  render() {
    return (
      <FlatList
[...]

Issue #849, opened 2017-10-25
Question: Is it possible to hook react-devtools to debug CSS? if so how?


Eyesonly88:
One of the tools I regularly use to debug CSS and dynamically interact with components during development is the `react-devtools` 👉 https://github.com/facebook/react-devtools/tree/master/packages/react-devtools

I can't figure out how to hook it up with expo. It's just stuck like this:
![image](https://user-images.githubusercontent.com/767013/31980711-dc9f10f8-b9aa-11e7-9a12-319082e0dd30.png)

I have remote debugging enabled and try reloading the app and stop/start debugging but didn't work.

Is it even possible to make this work? if so I would be super grateful to know how 🙏 
P.S: it works fine using plain react-native project (i.e. not through CRNA/Expo)

skevy:
Hi there!

Can you try setting the url (using exp or XDE) to "localhost", and make sure you're using the simulator, and then try using react-devtools again?

brentvatne:
<img width="1113" alt="screen shot 2017-10-25 at 1 04 21 pm" src="https://user-images.githubusercontent.com/90494/32020454-0fe127b0-b985-11e7-9c35-77a8ecf5da7c.png">

confirmed it works as expected in simulator. the problem is on device, where react-native isn't able to do this yet (on ios). it should work fine on device on android. feel free to contribute a fix to this issue upstream https://github.com/facebook/react-native/issues/14431

Eyesonly88:
Ah I see, thanks. 
I'll try to contribute a fix upstream then. Cheers.

happy-thorny:
I managed to get it working on iOS device, but I guess it is universal solution. According to [react-devtools-core documentation](https://www.npmjs.com/package/react-devtools-core), add this before any react-related import in your main file (App.js or something that provides the topmost component):
`require('react-devtools-core').connectToDevTools({host: 'ip-of-machine-with-devtools-running'})`

By default [CRNA](https://github.com/react-community/create-react-native-app)'s main file is crna-entry.js which is located somewhere in node_modules. I didn't test the solution with crna-entry as main because i have it's copy as main which looks like that:
```
// src/index.js
require('react-devtools-core').connectToDevTools({host: '-ip-'})

import Expo from 'expo'
import App from './App'
import React from 'react'
import { View } from 'react-native'

class AwakeInDevApp extends React.Component {
  render() {
    return (
      <View style={{flex: 1}}>
        {React.createElement(App, this.props)}
[...]

PR #47543, opened 2026-07-05
[ios][core] Fix CSS color regexes not matching on the iOS 27.0 beta runtime

tsapeta:
# Why

All nine `hsl()`/`hsla()` tests in `ColorConvertiblesTests` fail when the native test suite runs on the iOS 27.0 beta simulator runtime, throwing `InvalidHSLColorException` for valid inputs like `hsl(0, 100%, 50%)`. CI does not catch this because it runs on a stable runtime, and the Swift regex engine ships with the OS rather than with the app, so the same binary behaves differently per runtime. If the regression ships in the iOS 27 release, `hsl()` color parsing breaks for real apps.

# How

The root cause is a regression in the regex engine of the iOS 27.0 beta: regex-literal fragments composed inside `RegexBuilder` blocks (e.g. `Regex { /hsla?\(\s*/ ... }`) miscompile and silently stop matching. Only some shapes are affected, which is why the `rgb()` and `hwb()` patterns (built the same way) happened to survive: a bisect showed failing shapes directly adjacent to theirs, e.g. `Regex { Capture { /\d+/ }; /%\s*,\s*/ }` no longer matches `100%,`. Flat regex literals and pure builder DSL both work correctly.

Upstream context: this belongs to a known regression family in the Swift 6.4 regex engine (the stdlib the iOS 27 beta ships), tracked in [swiftlang/swift-experimental-string-processing#865](https://github.com/swiftlang/swift-experimental-string-processing/issues/865) and traced there to the DSL flattening/auto-possessification rework ([#849](https://github.com/swiftlang/swift-experimental-string-processing/pull/849)), with related piecemeal fixes in [#851](https://github.com/swiftlang/swift-experimental-string-processing/pull/851), [#859](https://github.com/swiftlang/swift-experimental-string-processing/pull/859) and [#866](https://github.com/swiftlang/swift-experimental-string-processing/pull/866). How builder concatenations are flattened and coalesced is clearly in flux, so this PR moves the patterns off that code path rather than dodging individual broken shapes.

All five CSS color patterns are now built from `RegexBuilder` primitives only (string literals, `Capture`, `Optionally`, `ZeroOrMore(.whitespace)`, `OneOrMore(.whitespace)`), sharing small `cssNumber`, `cssNumberOrPercent` and `commaSeparator` components. A comment on the section documents the constraint so the fragments don't sneak back in.

Isolation of the root cause: one identical test binary spawned via `simctl` passes on the iOS 26.1 and 26.5 simulator runtimes and on macOS 26.5, and fails only on the iOS 27.0 beta (24A5355p). Byte-for-byte copies of the new patterns were verified against the full test input set (including capture contents and negative cases) on all four runtimes before porting.

# Test Plan

`et native-unit-tests -p ios --packages expo-modules-core` on the iOS 27.0 beta simulator: all 696 tests pass, including the nine previously failing `hsl()`/`hsla()` tests and the existing `rgb()`/`hwb()` coverage. Before the fix, the same run failed exactly those nine tests.

### DIFF
--- a/packages/expo-modules-core/CHANGELOG.md
+++ b/packages/expo-modules-core/CHANGELOG.md
@@ -21,6 +21,7 @@
 
 ### 🐛 Bug fixes
 
+- [iOS] Fix `hsl()`/`hsla()` color strings failing to parse on some OS versions (observed on the iOS 27.0 beta), caused by regex-literal fragments composed in `RegexBuilder` blocks. The CSS color patterns are now built from `RegexBuilder` primitives only. ([#47543](https://github.com/expo/expo/pull/47543) by [@tsapeta](https://github.com/tsapeta))
 - [iOS] Fix a reload deadlock on the new architecture where reloading (from pressing `r` in the iOS simulator) would freeze the app until the watchdog killed it. `reloadAppAsync` now triggers the reload synchronously when already on the main thread instead of always deferring via `DispatchQueue.main.async`. ([#47392](https://github.com/expo/expo/pull/47392) by [@brentvatne](https://github.com/brentvatne))
 - [iOS] Pair a native shared object with one JavaScript object per runtime, so a shared object exposed to several runtimes (e.g. the main and UI runtimes, or worklet contexts) keeps a live pairing in each instead of only the one paired last. ([#47238](https://github.com/expo/expo/pull/47238) by [@tsapeta](https://github.com/tsapeta))
 - [Android][compose] Guard `onLayout` against detached window to prevent `LayoutNode should be attached to an owner` crash. ([#47085](https://github.com/expo/expo/pull/47085) by [@roitium](https://github.com/roitium))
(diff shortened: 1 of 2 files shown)


packages/expo-modules-core/CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
 
 ### 🐛 Bug fixes
 
+- [iOS] Fix `hsl()`/`hsla()` color strings failing to parse on some OS versions (observed on the iOS 27.0 beta), caused by regex-literal fragments composed in `RegexBuilder` blocks. The CSS color patterns are now built from `RegexBuilder` primitives only. ([#47543](https://github.com/expo/expo/pull/47543) by [@tsapeta](https://github.com/tsapeta))
```
alanjhughes, line 24: I don't know why claude adds these at the top 🤦‍♂️

alanjhughes (approved):

packages/expo-modules-core/CHANGELOG.md
tsapeta: Yeah 🤷‍♂️ I'm going to improve our `.md` files for agents this week. It's a good point to make it more explicit to put changelog entries at the end.

PR #47545, opened 2026-07-05
fix(expo-modules-autolinking): Honor explicit `null` platform overrides

### LINKED ISSUE #47544: [expo-modules-autolinking] Project react-native.config.js "platforms.android: null" is silently ignored when the library ships its own react-native.config.js

The reproduction is fully inline below: 3 small files + 2 commands from an empty directory, plus a dependency-free 20-line demonstration of the exact faulty function. I kept it inline because the whole repro is smaller than a repo README — but I'm happy to push it as a standalone repository, and/or open a PR (the fix is a one-line guard change), if preferred.

### Summary

A project-level `react-native.config.js` with

```js
module.exports = {
  dependencies: {
    'react-native-maps': { platforms: { android: null } },
  },
};
```

is the React Native CLI's documented mechanism to disable a library's autolinking for one platform ([cli `docs/autolinking.md`, "How can I disable autolinking for unsupported library?"](https://github.com/react-native-community/cli/blob/main/docs/autolinking.md) — `android: null, // disable Android platform, other platforms will still autolink if provided`).

expo-modules-autolinking's react-native-config integration clearly intends to support this semantics:

- `resolveDependencyConfigImplAndroidAsync` starts with an explicit skip: `if (reactNativeConfig === null) { // Skip autolinking for this package. return null; }` ([androidResolver.ts#L17-L23](https://github.com/expo/expo/blob/720e7a2c022c59bf2e63973605c0b7b24e8677bf/packages/expo-modules-autolinking/src/reactNativeConfig/androidResolver.ts#L17-L23))
- the tvos/macos branch documents it in prose: *"Use the platform-specific `react-native.config` entry when it's set — including an explicit `null`, which disables autolinking for that platform"* ([reactNativeConfig.ts#L137-L144](https://github.com/expo/expo/blob/720e7a2c022c59bf2e63973605c0b7b24e8677bf/packages/expo-modules-autolinking/src/reactNativeConfig/reactNativeConfig.ts#L137-L144))

But the project-level `null` never reaches those resolvers **whenever the library ships its own `react-native.config.js`** (react-native-maps, react-native-svg, react-native-webview, …). The project-override-over-library-config merge swallows it — silently: no warning, and the library is autolinked anyway. Besides breaking the disable-unsupported-package use case, this is a silent behavioral divergence from `@react-native-community/cli` for the same config file.

#### Root cause

`deepObjectMerge` ([reactNativeConfig.ts#L36-L53](https://github.com/expo/expo/blob/720e7a2c022c59bf2e63973605c0b7b24e8677bf/packages/expo-modules-autolinking/src/reactNativeConfig/reactNativeConfig.ts#L36-L53), applied to the project override at [#L100-L104](https://github.com/expo/expo/blob/720e7a2c022c59bf2e63973605c0b7b24e8677bf/packages/expo-modules-autolinking/src/reactNativeConfig/reactNativeConfig.ts#L100-L104)): the merge-branch guard checks `source !== undefined && typeof source === 'object' && !Array.isArray(source)` — but never `source !== null`. Since `typeof null === 'object'`, an explicit `null` override enters the merge branch, `for (const key in null)` iterates zero keys, and the function returns the unmodified library config.

Dependency-free demonstration (function copied verbatim from the published package, `expo-modules-autolinking@57.0.4` `build/reactNativeConfig/reactNativeConfig.js#L18-L33` — identical logic to the TS source on `main`):

```js
const deepObjectMerge = (target, source) => {
    if (source !== undefined &&
        typeof target === 'object' &&
        target != null &&
        !Array.isArray(target) &&
        (!target.constructor || target.constructor === Object) &&
        typeof source === 'object' &&
        !Array.isArray(source)) {
        target = { ...target };
        for (const key in source) {
            target[key] = deepObjectMerge(target[key], source[key]);
        }
        return target;
    }
    return source !== undefined ? source : target;
};

// react-native-maps' own react-native.config.js (`dependency` key):
const libraryConfig = { platforms: { android: { cmakeListsPath: 'src/main/jni/CMakeLists.txt' } } };
// project react-native.config.js (`dependencies['react-native-maps']`):
const projectOverride = { platforms: { android: null } };

console.log(deepObjectMerge(libraryConfig, projectOverride).platforms.android);
// Actual:   { cmakeListsPath: 'src/main/jni/CMakeLists.txt' }   <-- explicit null swallowed
// Expected: null                                                <-- so the android resolver skips the package
```

The asymmetry that pins the bug to the merge (verified in the steps below): the **identical** override *works* for a library that ships *no* `react-native.config.js` — there `target` is `undefined` at that path, the guard's `typeof target === 'object' && target != null` fails, and the `null` falls through `return source !== undefined ? source : target` and lands in the resolver. The `null` is swallowed precisely when the library's value at the same path is a plain object — which the platform-disable case always is.

#### Suggested fix

Treat explicit `null` as an override, not a merge target:

```diff
 const deepObjectMerge = (target: any, source: any): any => {
   if (
     source !== undefined &&
+    source !== null &&
     typeof target === 'object' &&
```

With that, `deepObjectMerge(<library platform config>, null)` falls through to `return source !== undefined ? source : target;` → returns `null`, which `resolveDependencyConfigImplAndroidAsync` already handles correctly. Happy to open a PR.

#### Workaround

For anyone else hitting this: `"expo": { "autolinking": { "android": { "exclude": ["react-native-maps"] } } }` in `package.json` works — the exclude set is checked before any config merging ([reactNativeConfig.ts#L75](https://github.com/expo/expo/blob/720e7a2c022c59bf2e63973605c0b7b24e8677bf/packages/expo-modules-autolinking/src/reactNativeConfig/reactNativeConfig.ts#L75)).

#### Versions

Reproduced with `expo-modules-autolinking@57.0.4` (Expo SDK 57 / `expo@57.0.2`, the current stable). The code is unchanged on `main` as of [`720e7a2c`](https://github.com/expo/expo/commit/720e7a2c022c59bf2e63973605c0b7b24e8677bf) (all permalinks above point at that commit).

### Steps to reproduce

From an empty directory, create three files. `react-native-maps` ships its own `react-native.config.js` (`dependency.platforms.android.cmakeListsPath`) → the bug case; `react-native-gesture-handler` ships none → the control.

**`package.json`**

```json
{
  "name": "repro",
  "version": "1.0.0",
  "private": true,
  "dependencies": {
    "expo": "~57.0.2",
    "react": "19.2.3",
    "react-native": "0.86.0",
    "react-native-gesture-handler": "~2.32.0",
    "react-native-maps": "1.27.2"
  }
}
```

**`react-native.config.js`** — disable **both** packages for Android:

```js
module.exports = {
  dependencies: {
    'react-native-maps': { platforms: { android: null } },          // ships its own react-native.config.js
    'react-native-gesture-handler': { platforms: { android: null } }, // ships none (control)
  },
};
```

**`app.json`**

```json
{ "expo": { "name": "repro", "slug": "repro" } }
```

Then:

```bash
npm install
npx expo-modules-autolinking react-native-config --platform android --json \
  | node -e "const c=JSON.parse(require('fs').readFileSync(0,'utf8'));for(const n of['react-native-maps','react-native-gesture-handler'])console.log(n+':',c.dependencies[n]?'AUTOLINKED':'skipped');"
```

**Expected** (and what `@react-native-community/cli` does for the same config): both packages skipped.

**Actual** (expo-modules-autolinking 57.0.4, fresh install, macOS, npm):

```
react-native-maps: AUTOLINKED     <-- BUG: explicit null ignored (library ships its own config)
react-native-gesture-handler: skipped     <-- control: same override honored (library ships no config)
```

The returned `react-native-maps` entry is the library's full Android config, as if the override didn't exist:

```jsonc
{
  "root": "<projectRoot>/node_modules/react-native-maps",
  "name": "react-native-maps",
  "platforms": {
    "android": {
      "sourceDir": "<projectRoot>/node_modules/react-native-maps/android",
      "packageImportPath": "import com.rnmaps.maps.MapsPackage;",
      "packageInstance": "new MapsPackage()",
      "libraryName": "RNMapsSpecs",
      "componentDescriptors": ["RNMapsCalloutComponentDescriptor", "…"],
      "cmakeListsPath": "<projectRoot>/node_modules/react-native-maps/android/src/main/jni/CMakeLists.txt",
      // …
    }
  }
}
```

`createReactNativeConfigAsync` is the same entry point the Gradle autolinking / `npx expo prebuild` integration consumes, so the excluded library's Android AAR + New Architecture codegen are compiled into builds that were configured to drop it.

Reproduced on Android; the merge is platform-agnostic, so `platforms.ios: null` (and tvos / macos / web) are affected the same way whenever the library ships its own config file.

Context: found while making react-native-maps iOS-only (MapKit on iOS, `expo-maps` on Android) in a production Expo SDK 57 app. Related reports previously filed from the same project: #46418 / #46421 (expo-maps, unrelated code path).

### Environment

```
    System:
      OS: macOS 26.5.1
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 22.14.0 - ~/.nvm/versions/node/v22.14.0/bin/node
      Yarn: 4.10.3 - ~/.nvm/versions/node/v22.14.0/bin/yarn
      npm: 11.3.0 - ~/.nvm/versions/node/v22.14.0/bin/npm
      Watchman: 2026.04.20.00 - /opt/homebrew/bin/watchman
    Managers:
    SDKs:
      iOS SDK:
      Android SDK:
        API Levels: 30, 34, 35, 36, 37
        Build Tools: 30.0.3, 34.0.0, 35.0.0, 36.0.0, 37.0.0
    IDEs:
      Android Studio: 2025.2 AI-252.27397.103.2522.14617522
    npmPackages:
      expo: ~57.0.2 => 57.0.2
      react: 19.2.3 => 19.2.3
    npmGlobalPackages:
      eas-cli: 20.3.0
    Expo Workflow: managed
```

### Expo Doctor Diagnostics

Run on the minimal reproduction above (`npx expo-doctor@latest`, v1.20.0):

```
```

chownation:
# Why

Fixes #47544.

Project-level `react-native.config.js` can use `platforms.<platform>: null` to disable autolinking for one platform. `expo-modules-autolinking` already treats an explicit `null` as a skip once it reaches the platform resolver, but the deep merge between a library config and a project override swallowed `null` when the library already had a plain object at that path.

# How

Treat explicit `null` as a leaf override in `deepObjectMerge` instead of entering the object-merge branch. This lets the project override replace the library platform config with `null`, while preserving the existing deep-merge behavior for object overrides.

I also updated the existing regression test that encoded the previous behavior so Android `platforms.android: null` is passed through to the resolver as `null`, and left the adjacent nested-object merge coverage intact.

# Test Plan

- `pnpm run test -- src/reactNativeConfig/__tests__/reactNativeConfig-test.ts`
  - `Test Suites: 1 passed, 1 total`
  - `Tests: 21 passed, 21 total`
- `pnpm run test`
  - `Test Suites: 21 passed, 21 total`
  - `Tests: 241 passed, 241 total`
- `pnpm run typecheck`
- `pnpm run lint`

Before opening this PR, I also checked open PRs for `deepObjectMerge expo-modules-autolinking`, `platforms.android null autolinking`, and `react-native.config null autolinking`; all returned no matches.

# Checklist

### DIFF
--- a/packages/expo-modules-autolinking/CHANGELOG.md
+++ b/packages/expo-modules-autolinking/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 ### 🐛 Bug fixes
 
+- Respect explicit per-platform null overrides from project `react-native.config.js` files when merging dependency config from packages that ship their own config. ([#47545](https://github.com/expo/expo/pull/47545) by [@chownation](https://github.com/chownation))
 - [iOS] Fix a duplicate pod error (`multiple dependencies with different sources`) for precompiled modules in a Podfile with multiple targets. `prebuilt_react_active?` now mirrors React Native's default for `RCT_USE_PREBUILT_RNCORE` (prebuilt unless explicitly `0`), since `use_react_native!` only sets it after `use_expo_modules!` has run. ([#47329](https://github.com/expo/expo/pull/47329) by [@janicduplessis](https://github.com/janicduplessis))
 - [iOS] Align the `react-native-reanimated` precompile config with `react-native-reanimated@4.4.1` upstream defaults, adding the `CSS/core/transition` and `PseudoStyles` header mappings along with the missing `IOS_CSS_CORE_ANIMATION` and `USE_ANIMATION_BACKEND` feature flags so its XCFramework builds. ([#46950](https://github.com/expo/expo/pull/46950) by [@zoontek](https://github.com/zoontek))
 - Fixed build error for unresolvable `expo-modules-macros-plugin`. ([#46294](https://github.com/expo/expo/pull/46294) by [@kudo](https://github.com/kudo))
--- a/packages/expo-modules-autolinking/CHANGELOG.md
+++ b/packages/expo-modules-autolinking/CHANGELOG.md
@@ -22,6 +22,7 @@
 - [iOS] Stop discarding xcconfig changes made by other `post_install` hooks, which broke builds against React Native nightlies with `include of non-modular header inside framework module 'React.…'`. ([#49038](https://github.com/expo/expo/pull/49038) by [@kudo](https://github.com/kudo))
 - [iOS] Fix `'React/RCTBridge.h' file not found` with `ios.useFrameworks: "dynamic"` by reverting the dynamic-framework linkage guard from [#47500](https://github.com/expo/expo/pull/47500). Under `:dynamic` linkage every pod target is already a dynamic framework, so the guard skipped the whole `USE_FRAMEWORKS` downgrade. `@rnmapbox/maps` 10.3.2 no longer needs the guard because it skips its own dynamic flip when precompiled modules are enabled. ([#48869](https://github.com/expo/expo/pull/48869) by [@kudo](https://github.com/kudo))
 - [iOS] Pull a 3rd-party pod's prebuilt XCFramework to source when a dependent 3rd-party pod builds from source, fixing `'worklets/Compat/StableApi.h' file not found`. ([#49147](https://github.com/expo/expo/pull/49147) by [@chrfalch](https://github.com/chrfalch))
+- Respect explicit per-platform `null` overrides from project `react-native.config.js` files ([#47545](https://github.com/expo/expo/pull/47545) by [@chownation](https://github.com/chownation))
 
 ### 💡 Others
 
--- a/packages/expo-modules-autolinking/src/reactNativeConfig/__tests__/reactNativeConfig-test.ts
+++ b/packages/expo-modules-autolinking/src/reactNativeConfig/__tests__/reactNativeConfig-test.ts
@@ -516,13 +516,19 @@ describe(resolveReactNativeModule, () => {
   });
 
   itWithMemoize(
-    'should preserve library config when project config sets platform to null (deep merge)',
+    'should honor project config null platform overrides when library config has platform data',
     async () => {
+      const androidResolver = require('../androidResolver');
+      const mockPlatformResolverAndroid = jest.spyOn(
+        androidResolver,
+        'resolveDependencyConfigImplAndroidAsync'
+      );
+      mockPlatformResolverAndroid.mockResolvedValueOnce(null);
       const projectConfig: RNConfigReactNativeProjectConfig = {
         dependencies: {
           'react-native-test': {
             platforms: {
-              ios: null,
+              android: null,
             },
           },
         },
@@ -530,9 +536,9 @@ describe(resolveReactNativeModule, () => {
       const libraryConfig: RNConfigReactNativeLibraryConfig = {
         dependency: {
           platforms: {
-            ios: {
-              configurations: ['Debug'],
-              scriptPhases: [{ name: 'test', path: './test.js' }],
+            android: {
+              sourceDir: './android',
+              cmakeListsPath: './src/main/jni/CMakeLists.txt',
             },
           },
         },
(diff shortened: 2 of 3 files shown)


kitten (approved):
validated against the community CLI and looks to match the expected cases closely. Nice catch!

PR #47556, opened 2026-07-06
[docs] Expand contributor guidance in .claude/CLAUDE.md

tsapeta:
# Why

We receive a growing number of agent-assisted pull requests, and the repo's `.claude/CLAUDE.md` didn't cover most of the conventions we expect them to follow. This expands it so Claude Code sessions working in this repo pick up those conventions in every session instead of relying on each contributor's setup.

# How

- Reordered the file: repo-wide rules and a repository layout overview come first, the `expotools` reference moved near the end.
- Added a Changelogs section based on [Updating Changelogs](https://github.com/expo/expo/blob/main/guides/contributing/Updating%20Changelogs.md): categories, entry format and phrasing, appending new entries at the end of a category, and predicting the PR number for the entry link.
- Expanded the Creating PRs section: a local (non-posting) `/deep-code-review` self-review whose findings the author decides on, a short comprehension check with the author before opening non-trivial PRs, draft-first PR creation, and description guidelines.
- Documented smaller conventions: pnpm as the package manager, branch naming, commit message tags, scoping `et check-packages` to changed packages.
- Emphasized the most consequential rules with bold keywords and removed duplicated guidance.

Here is the [rendered `.claude/CLAUDE.md` on this branch](https://github.com/expo/expo/blob/tsapeta/update-claude-md/.claude/CLAUDE.md).

# Test Plan

Docs-only change: proofread the rendered Markdown and verified the referenced commands, paths, and guide links exist in the repo.

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -1,23 +1,23 @@
-## expotools (`et`)
-
-This repo ships its own CLI, `expotools`, available directly as `et <command> <...args>`.
-Call it as `et`, not via `npx`, `bunx`, or `pnpm run`. It's put on the PATH by `direnv`, so run
-`direnv allow` once in the repo root if `et` isn't found; otherwise invoke it directly with
-`node ./tools/bin/expotools.js <command>`. Many of our processes (native unit tests, prebuild,
-and more) run through it.
-
-Commands you'll reach for during normal development:
-
-- `et check-packages` — verify that packages build and their tests pass.
-
-Run `et --help` to discover the rest (publishing, changelogs, on-call dashboards, and more).
-
 ## Rules
 
-- **Red/green:** write the test first and watch it fail, then implement the feature or fix until
-  it passes. Don't write the implementation before the failing test exists.
+- **Package manager:** this is a pnpm workspace. Use `pnpm` for installing, running scripts, and
+  managing dependencies; **NEVER** use `npm` or `yarn`. Run `pnpm install` at the repo root.
+
+## Repository layout
+
+- `packages/`: the Expo SDK packages, each with its own `CHANGELOG.md`.
+- `apps/`: development and test apps: `bare-expo` (manual testing of packages), `native-tests`
+  (host app for native unit tests), `native-component-list` (component demos), `test-suite`
+  (JS test suite), `expo-go`.
+- `tools/`: source of the `expotools` CLI (see the expotools section below).
+- `docs/`: the docs site (docs.expo.dev).
+- `templates/`: project templates published to npm.
+- `guides/`: contributor guides referenced throughout this file.
 
 ## Writing tests
+
+**Red/green:** write the test first and watch it fail, then implement the feature or fix until
+it passes. **DON'T** write the implementation before the failing test exists.
 
 ### Unit tests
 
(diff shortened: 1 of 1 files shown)


.claude/CLAUDE.md
```diff
@@ -40,32 +38,136 @@ one package with `--packages <name>` (e.g. `et native-unit-tests -p ios --packag
...
 
 ## Creating PRs
 
-Follow the contribution guide: https://github.com/expo/expo/blob/main/CONTRIBUTING.md and the PR template:
+Follow the contribution guide: https://github.com/expo/expo/blob/main/CONTRIBUTING.md and the PR template.
+
+**Self-review:** before opening a non-trivial PR, and before adding the changelog entry, run
+`/deep-code-review` with no arguments (local mode: it reviews the branch against `main` and
+reports findings in the conversation without posting anything to GitHub). **DON'T** fix the
+findings straightaway: present them to the author and let them decide which ones to address.
```
tsapeta, line 115: @Kudo, are you ok with running the deep code review before opening the PR?

alanjhughes (approved):

.claude/CLAUDE.md
```diff
@@ -40,32 +38,136 @@ one package with `--packages <name>` (e.g. `et native-unit-tests -p ios --packag
...
+after the existing entries: the order of changes sometimes matters.
+
+- `🛠 Breaking changes`: API changes that require users to update their code or configuration
+  (major bump).
+- `🎉 New features`: non-breaking additions to the public API (at least a minor bump). Features
+  that are internal-only belong in Others instead.
+- `🐛 Bug fixes`: bug fixes and documentation clarifications.
+- `⚠️ Notices`: deprecations and corner-case behavior changes that stay backwards compatible.
+- `💡 Others`: internal changes, refactors, build tooling, routine work.
+- `📚 3rd party library updates`: upgrades of libraries shipped in Expo Go (root changelog only).
```
Kudo: these sections feel duplicated as the **Updating Changelogs.md**. maybe reuse the link above if that's enough

.claude/CLAUDE.md
```diff
@@ -40,32 +38,136 @@ one package with `--packages <name>` (e.g. `et native-unit-tests -p ios --packag
...
+  ```
+  - [iOS] Fixed video playback stalling on older devices. ([#NNNNN](https://github.com/expo/expo/pull/NNNNN) by [@username](https://github.com/username))
+  ```
+
+The PR number doesn't exist until the PR is opened, so predict it right before creating the PR:
+fetch the most recent issue and PR numbers and take `max + 1`.
+
+```
+gh issue list -R expo/expo -L 1 --state all --json number
+gh pr list   -R expo/expo -L 1 --state all --json number
```
Kudo, line 96: should consider if `gh` isn't installed

.claude/CLAUDE.md
```diff
@@ -40,32 +38,136 @@ one package with `--packages <name>` (e.g. `et native-unit-tests -p ios --packag
...
+  answer options (one correct, the others plausible but wrong), plus a separate "walk me through
+  it" escape option.
+- Ask about the non-obvious parts: design decisions and tradeoffs, assumptions and edge cases the
+  change depends on, details a maintainer might trip on. Don't ask about things obvious from the
+  diff.
+- If an answer reveals a real problem, fix it before opening the PR. If the author can't answer
+  confidently, walk through that part of the code together first; a shaky answer means the code
+  isn't ready.
+- Contributors working on a fork (check `git remote -v` against `expo/expo`) get a higher bar:
+  **DON'T** open the PR until they can explain what the change does and why.
```
Kudo, line 132: this is super interesting! i'm wondering what the actual workflow looks like? did you try create a pr with the new CLAUDE.md?

.claude/CLAUDE.md
```diff
@@ -40,32 +38,136 @@ one package with `--packages <name>` (e.g. `et native-unit-tests -p ios --packag
...
+how it evolved while you worked on it (approaches you abandoned, bugs you introduced and fixed
+along the way, earlier revisions of the diff). **DON'T** use em dashes (—); rewrite with commas,
+parentheses, colons, or separate sentences instead.
+
+**Creating the PR:** **ALWAYS** create it as a draft; mark it ready for review once CI passes. Pass the
+description through stdin with a quoted heredoc so backticks and newlines survive unescaped:
 
-**See also:**
+```
+gh pr create --draft --title "..." --body-file - <<'EOF'
```
Kudo, line 156: ditto when no `gh` available

.claude/CLAUDE.md
Kudo: > **Self-review:** before opening a non-trivial PR, and before adding the changelog entry, run
`/deep-code-review` with no arguments (local mode: it reviews the branch against `main` and

does `/deep-code-review` work before creating the pr?

Kudo (changes requested):

.claude/CLAUDE.md
tsapeta: We had a link before, but I noticed it wasn't fully respected. Having it here makes it more explicit for agents and the maintenance cost for us to keep it in two places is small.

.claude/CLAUDE.md
tsapeta: I had this in my project memories for a while, but I've been asked just a few times. I think Claude sometimes forgets/doesn't load memories, so it's better to test whether it works more reliably when it's in CLAUDE.md.
I'll try it locally next time and post a screenshot here.

.claude/CLAUDE.md
tsapeta: Yes, that's its local mode: run with no arguments it diffs the current branch against main and prints findings in the conversation only, nothing is posted to GitHub. No PR needs to exist yet. I used exactly that flow on this PR before opening it.

.claude/CLAUDE.md
tsapeta: Screenshot 2026-07-07 at 15 27 53

Kudo (approved):

.claude/CLAUDE.md
```diff
@@ -31,41 +31,150 @@ Swift/Kotlin unit tests live in `packages/<pkg>/ios/Tests/` and `android/`, and
...
+  the implementation.
+- Prefix platform-specific entries with `[iOS]`, `[Android]`, or `[Web]`, matching existing
+  entries in the file.
+- End with links to the PR and the author's GitHub profile:
+
+  ```
+  - [iOS] Fixed video playback stalling on older devices. ([#NNNNN](https://github.com/expo/expo/pull/NNNNN) by [@username](https://github.com/username))
+  ```
+
+The PR number doesn't exist until the PR is opened, so predict it right before creating the PR:
```
chrfalch, line 90: We're a pretty busy repo, maybe it is ok to not predict the next number? It seems to only add a few extra steps - checking and verifying after creating the PR etc. that the CHANGELOG job will fix later? To me this extra rule won't give us any extra value/quality.

chrfalch (approved):

.claude/CLAUDE.md
tsapeta: I'm doing it that way for quite a long time and that worked perfectly 😉 Usually the prediction is correct because it's checked and the changelog entry is added just before opening the PR, so in most cases the window when another issue/PR can be opened is pretty short.
Honestly I'd like to avoid triggering unnecessary CI checks between opening a PR and pushing a changelog entry.

.claude/CLAUDE.md
```diff
@@ -1,24 +1,24 @@
...
-Run `et --help` to discover the rest (publishing, changelogs, on-call dashboards, and more).
-
-## Rules
-
-- **Red/green:** write the test first and watch it fail, then implement the feature or fix until
-  it passes. Don't write the implementation before the failing test exists.
+## Repository layout
+
+This is a pnpm workspace. Use `pnpm` for installing, running scripts, and managing dependencies;
+**NEVER** use `npm` or `yarn`. Run `pnpm install` at the repo root (except for `docs/`, see below).
```
kitten, line 4:
```suggestion
Run `pnpm install` at the repo root (except for `docs/`, see below).
```

Generally, these two feel a bit speculative. From hundreds of sessions, I've only seen Claude accidentally reach for `yarn` once. I think having a more aggressive instruction here may lead to issues in some cases.

I think saying at the repo root (without specifying `only` there, and then calling out a specific `docs` instruction and having it below is a bit redundant too)

.claude/CLAUDE.md
```diff
@@ -1,24 +1,24 @@
...
+- `docs/`: the docs site (docs.expo.dev). It's a standalone pnpm project outside the workspace,
+  with its own lockfile: the root `pnpm install` doesn't cover it, so for docs work also run
+  `pnpm install` inside `docs/`.
+- `templates/`: project templates published to npm.
+- `guides/`: contributor guides referenced throughout this file.
 
 ## Writing tests
 
+**Red/green:** write the test first and watch it fail, then implement the feature or fix until
+it passes. **DON'T** write the implementation before the failing test exists.
```
kitten, line 20: Let's please not add this, because I'm not looking forward to having to delete this on my own end when it's not applicable or not a good idea for non-Expo modules/native work or devtools work, where this doesn't always make sense or isn't always applicable.

For feature work and external PRs, completely agree that this can be a value add, but we'll have to keep in mind that this by default will also apply to us and can be in the way of our own work.

Also worth noting that `**DON'T**` and more aggressive qualifiers (like above) don't always yield great results.

.claude/CLAUDE.md
```diff
@@ -31,41 +31,150 @@ Swift/Kotlin unit tests live in `packages/<pkg>/ios/Tests/` and `android/`, and
...
 
 Running native tests:
-`et native-unit-tests` — run native unit tests for all packages that provide them. Scope to
+`et native-unit-tests` runs native unit tests for all packages that provide them. Scope to
 one package with `--packages <name>` (e.g. `et native-unit-tests -p ios --packages @expo/ui`).
 
 ## Committing
 
-Before committing, you may use `turbo run <task>` to run an npm script for a given task on all dependents, for example `build`, `typecheck`, `depscheck`, `test`, and `lint`. This may also be run via `et check-packages <...packages>` with the names of the packages that changed. This matches how CI checks packages.
+Before committing, you may use `turbo run <task>` to run an npm script for a given task on all dependents, for example `build`, `typecheck`, `depscheck`, `test`, and `lint`. This may also be run via `et check-packages <...packages>`, which matches how CI checks packages. Keep the loop fast by scoping it to the packages you changed, e.g. `et check-packages expo-location`, rather than running repo-wide tasks.
```
kitten, line 43: Curious about this:
> Keep the loop fast by scoping it to the packages you changed, e.g. `et check-packages expo-location`, rather than running repo-wide tasks.

I haven't seen an agent over-scope this just yet. Not saying it can't happen, but haven't seen it myself yet

.claude/CLAUDE.md
```diff
@@ -31,41 +31,150 @@ Swift/Kotlin unit tests live in `packages/<pkg>/ios/Tests/` and `android/`, and
...
+
+The compiled `build/` output is gitignored and is **not** committed. Turborepo regenerates and caches it on demand.
```
kitten, line 45: Haven't seen this be an issue. We should maybe check if there's conflicting info on this, but ever since we deleted build output and some prompts have been updated, this has been fine and obvious for agents, at least on my end, so probably redundant

.claude/CLAUDE.md
```diff
@@ -31,41 +31,150 @@ Swift/Kotlin unit tests live in `packages/<pkg>/ios/Tests/` and `android/`, and
...
+
+**Commit messages** use the format `[platform][api] Title`, e.g. `[ios][video] Fix black screen
+on older devices`. The API tag drops the `expo-` prefix: `[video]`, not `[expo-video]`.
+
+**Branches** are named `<github-username>/<short-description>`, e.g. `username/fix-video-playback`.
```
kitten, line 50: This isn't consistent, and the main convention is `@<username>/<verb>/<description>`

Would imho be better to leave this to personal prompts too. I think the commit convention above is also borderline personal preference

.claude/CLAUDE.md
```diff
@@ -31,41 +31,150 @@ Swift/Kotlin unit tests live in `packages/<pkg>/ios/Tests/` and `android/`, and
...
+## Changelogs
+
+Most packages in `packages/` have their own `CHANGELOG.md`, and every PR that changes a package
+should add an entry to it. A PR that changes several packages adds an entry to each changed
+package's changelog. Changelogs exist only for packages: PRs limited to `apps/`, `docs/`,
+`templates/`, `tools/`, or `guides/` don't need an entry anywhere. See
+[Updating Changelogs](https://github.com/expo/expo/blob/main/guides/contributing/Updating%20Changelogs.md)
+for the full guide.
+
+### Where the entry goes
```
kitten, line 61: Just noting, I haven't seen it mess this up yet. Could be something on my end, but I think, like above, or like with keywords such as `**ALWAYS**`, I'm wary of adding things (and increasing the context with many tokens) for things that it already does correctly

.claude/CLAUDE.md
kitten: To add another note to this, since it's a bit of a theme. Unless we delete the file for ourselves locally, there's some issues with assuming the external contributor flow, and taking up more and more tokens at the system prompt level when working on `expo/expo`, so I'd be a bit wary of adding this much text for the changelog instructions in the first place to be honest 😅 

to me, this all falls into a personal preference kind of area

.claude/CLAUDE.md
```diff
@@ -31,41 +31,150 @@ Swift/Kotlin unit tests live in `packages/<pkg>/ios/Tests/` and `android/`, and
...
+if they differ. The review bot also suggests these links in a code review, but **DON'T** rely on
+it: a silent mismatch leaves a dead link.
 
 ## Creating PRs
 
-Follow the contribution guide: https://github.com/expo/expo/blob/main/CONTRIBUTING.md and the PR template:
+Follow the contribution guide: https://github.com/expo/expo/blob/main/CONTRIBUTING.md and the PR template.
+
+The PR workflow relies on the GitHub CLI (`gh`). If it's not installed or authenticated, stop and
+ask the user to set it up (`gh auth login`); **DON'T** fall back to raw GitHub API requests or ask
```
kitten, line 108: Again, good for external contributors, personal preference for us. As in, this could lead to annoying forced flows on our end

kitten (commented):
I think there's a lot of good things in here, but overall, there's a theme of diminishing returns. Main things are:

- overuse of instructions and `**ALWAYS**`, `**DON'T**`, etc keywords, where it's not proven that they're necessary (in my experience there's a limit to how well Claude specifically responds to them, especially as their numbers increases)
- a few forced flows in here are nice for externals but can be annoying internally
- a few forced flows in here are nice for personal preference but not necessarily great if every external contributor or some people at Expo have them

overall, I think there's a main question here around whether everything here passes a sniff test of usefulness (does it have to be in here, or else, something goes wrong) or alignment (do we want this to happen 100% for everyone all the time)

PR #47567, opened 2026-07-06
[expo][jest-expo] remove vector-icons aliasing from jest-preset

matinzd:
**BREAKING CHANGE**
 
Removed the `react-native-vector-icons` → `@expo/vector-icons` alias from the Jest preset.

# Why

`@expo/vector-icons` [is being deprecated and is no longer recommended](https://docs.expo.dev/guides/icons/), with Expo recommending a migration to `react-native-vector-icons` directly (see the [migration blog post](https://expo.dev/blog/moving-away-from-expo-vector-icons)).

The `jest-preset.js` alias that mapped `react-native-vector-icons` imports to `@expo/vector-icons` causes problems in monorepos, particularly with Nx. Nx resolves module aliases eagerly across the workspace graph, so it tries to resolve `@expo/vector-icons` even for projects that do not depend on either `@expo/vector-icons` or `react-native-vector-icons`. This breaks the build/test graph for those projects.

Since this alias only emulated Metro's resolver behavior for a package that's now on its way out, keeping it in the Jest preset does more harm than good.

There are more changes introduced in https://github.com/expo/expo/pull/25512 which we need to address now that `@expo/vector-icons` will not be around. Personally I feel that all changes introduced in that PR needs to be reverted.

# How

Removed the `moduleNameMapper` entries in `packages/jest-expo/jest-preset.js` that aliased `react-native-vector-icons` (and its subpaths) to `@expo/vector-icons`. Projects using `react-native-vector-icons` will now resolve it directly, matching the non-deprecated, recommended path.

# Test Plan

Ran `et check-packages jest-expo` to confirm the preset still builds, type-checks, and passes existing tests.

# Checklist

### DIFF
--- a/packages/jest-expo/CHANGELOG.md
+++ b/packages/jest-expo/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### 🐛 Bug fixes
 
+- Remove `react-native-vector-icons` → `@expo/vector-icons` aliasing from the Jest preset, since it broke module resolution in monorepos (e.g. Nx) for projects without `@expo/vector-icons` installed. ([#47567](https://github.com/expo/expo/pull/47567) by [@matinzd](https://github.com/matinzd))
+
 ### 💡 Others
 
 ## 56.0.4 — 2026-05-20
--- a/packages/jest-expo/jest-preset.js
+++ b/packages/jest-expo/jest-preset.js
@@ -26,13 +26,6 @@ jestPreset = cloneDeep(jestPreset);
 const { withTypescriptMapping } = require('./src/preset/withTypescriptMapping');
 const { resolveBabelOptions } = require('./src/resolveBabelOptions');
 
-// Emulate the alias behavior of Expo's Metro resolver.
-jestPreset.moduleNameMapper = {
-  ...(jestPreset.moduleNameMapper || {}),
-  '^react-native-vector-icons$': '@expo/vector-icons',
-  '^react-native-vector-icons/(.*)': '@expo/vector-icons/$1',
-};
-
 const upstreamBabelJest = Object.keys(jestPreset.transform).find(
   (key) => jestPreset.transform[key] === 'babel-jest'
 );


packages/jest-expo/CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
 
 ### 🐛 Bug fixes
 
+- Remove `react-native-vector-icons` → `@expo/vector-icons` aliasing from the Jest preset, since it broke module resolution in monorepos (e.g. Nx) for projects without `@expo/vector-icons` installed. ([#47567](https://github.com/expo/expo/pull/47567) by [@matinzd](https://github.com/matinzd))
```
kitten, line 13:
```suggestion
- Remove `react-native-vector-icons` → `@expo/vector-icons` aliasing from the Jest preset, as it's obsolete ([#47567](https://github.com/expo/expo/pull/47567) by [@matinzd](https://github.com/matinzd))
```

I can't trace the full resolution path here without a repro, but I'd wager that whenever Nx is mentioned that, no, it's not a general monorepo issue, but a config issue with Nx and/or the lack of workspaces or a specific expected setup.

In this case, the remap here can clearly be an isolated modules / dependency graph violation, but I'd say (a) it's not a general monorepo issue, as in, more generally a deps issue, and (b) not directly related to Nx.

So stepping back the changelog entry, I'd just say it's obsolete, as there's another migration path for `@expo/vector-icons` already

cc @brentvatne

matinzd: I am not really sure what's causing this but I don't have vector-icons installed and my tests are failing.

```
 NX   The externalDependency '@expo/vector-icons' for 'lendo-app:test' could not be found
```

Not sure if it's nx only. I try to reproduce it outside of our repository.

kitten:
Sorry, coming back to this. There's multiple issues that make this difficult to merge.

The first is that `jest-expo` needs an internal owner, but Jest is in an awkward spot where we're hesitant to invest the time in upgrading to Jest 30 when none of the changes actually help us or push anything forward. Unfortunately, due to the coupling that also affects `jest-expo`. But that means we're conservative on merging changes to it right now.

Regarding the alias itself, the alias is still valid, as far as I know, since the `react-native-vector-icons` deprecation and alias still apply, and will still apply, with only the latter now having a "manual migration guide" which doesn't invalidate this alias.

Lastly, if there's a separate issue re. resolutions, as said, that's complicated by the mention of Nx without having a reproduction or separate issue to track that, which likely wouldn't be resolved by changing semantics here, but by knowing what the underlying root cause is.

Closing for now, sorry!

PR #44909, opened 2026-04-19
[expo] fix expo/fetch fatal close race after stream cancel

safaiyeh:
# Why

`expo/fetch` can surface a fatal JS crash in release builds:

```
TypeError: The stream is not in a state that permits close
```

Native is free to deliver `didComplete` (or `didFailWithError`) *after* the JS consumer has already canceled the body stream — e.g. when the app is backgrounded mid-stream, or a consumer calls `reader.cancel()` while the native request is about to finish. Today the `didComplete` listener always calls `controller.close()`, which throws on an already-closed controller. The throw escapes the listener and becomes an unhandled exception.

This is the iOS-side counterpart to #34804 (previously reported on Android). I hit it in a production iOS app via Firebase Crashlytics:

- Error: `JavaScriptError - The stream is not in a state that permits close`
- Context: user sent a chat message, app transitioned to background, request was in flight
- Expo SDK 55, iOS 18.7.7

#34892 added the `isControllerClosed` guard for `didReceiveResponseData` but left `didComplete` and `didFailWithError` unguarded, so the same race still trips the fatal close.

# How

Extend the existing `isControllerClosed` guard to the two remaining listeners in `FetchResponse`'s body source:

```ts
response.addListener('didComplete', () => {
  if (isControllerClosed) return;
  isControllerClosed = true;
  controller.close();
});
```

Also flip the ordering — set `isControllerClosed = true` *before* `controller.close()` / `controller.error()` — so any re-entrant listener that fires during close won't see a not-yet-closed flag.

No public API change, no behavior change on the happy path. Guard only activates when cancel + native completion race.

# Test Plan

Added three unit tests in `FetchResponse-test.ts` covering:
1. `didComplete` arriving after `reader.cancel()` (the iOS repro)
2. `didFailWithError` arriving after `reader.cancel()` (symmetric case)
3. `didComplete` firing twice (defensive — native is free-threaded)

The existing test stub for `NativeResponse` was a no-op class, so I extended it with a minimal event emitter surface (`addListener`/`removeListener`/`removeAllListeners`/`emit` + stub `startStreaming`/`cancelStreaming`) so the tests can simulate native events hitting the JS listeners. No changes to the public `ExpoFetchModule` shape.

Without the fix, each of those three tests throws with `TypeError: The stream is not in a state that permits close` out of the listener. With the fix, they pass silently.

# Checklist

[...]

Issue #34804, opened 2025-02-10
[Android][Fetch] `break` statement in the AsyncIterator of streaming response causes enqueue-after-cancel error


yukukotani:
### Minimal reproducible example

https://github.com/expo/expo/pull/34803

### What platform(s) does this occur on?

Android

### Where did you reproduce the issue?

in a standalone app

#### TL;DR;

[This code](https://github.com/yukukotani/expo/blob/477c3c3a103437fedf97096ac756ef5b6508f0d8/apps/test-suite/tests/Fetch.ts#L297-L315) occurs `The stream is not in a state that permits enqueue` error and `The stream is not in a state that permits close` error on the `break` statement at L310.

https://github.com/user-attachments/assets/c231d133-de8d-4407-9da2-2ea72300f4d6

#### Details

`expo/fetch`'s response can be treated as an AsyncIterator, and we can iterate it using for-await statement. It's ok in the case of we iterate it to the end.

However, when we break it, the stream is going to be canceled. This is the right behavior [according to the spec](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream#async_iteration). In spite of that, `expo/fetch` keeps emitting `didReceiveResponseData` and `didComplete` events, which triggers [enqueue](https://github.com/yukukotani/expo/blob/477c3c3a103437fedf97096ac756ef5b6508f0d8/packages/expo/src/winter/fetch/FetchResponse.ts#L30-L32) and [close](https://github.com/yukukotani/expo/blob/477c3c3a103437fedf97096ac756ef5b6508f0d8/packages/expo/src/winter/fetch/FetchResponse.ts#L34-L36). This results in enqueue-after-cancel or cancel-after-cancel and causes `The stream is not in a state that permits enqueue` or `The stream is not in a state that permits close` error.

### Environment

```text
`./apps/bare-expo`

  expo-env-info 1.2.2 environment info:
    System:
      OS: macOS 14.5
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 22.13.1 - ~/.volta/tools/image/node/22.13.1/bin/node
      Yarn: 1.22.19 - ~/.volta/tools/image/yarn/1.22.19/bin/yarn
      npm: 10.9.2 - ~/.volta/tools/image/node/22.13.1/bin/npm
      Watchman: 2024.12.02.00 - /opt/homebrew/bin/watchman
    Managers:
[...]

Issue #33553, opened 2024-12-10
Fetch doesn't cancel the stream


hollanderbart:
### Minimal reproducible example

https://github.com/hollanderbart/BusinessCards-Connect

### What platform(s) does this occur on?

Android, iOS

### Where did you reproduce the issue?

in a standalone app

### Summary

When I've created a server fetch stream and I want to cancel the stream using an AbortController, the stream isn't cancelled. The stream still continues after calling the abort() function. I expect the current stream to be cancelled.

In my reproduction repo the stream is started but keeps returning messages, although abortController.abort() is called after the first message.

```
(NOBRIDGE) LOG  LOG: Response: {"redirected": false, "status": 200, "statusText": "no error", "url": "https://httpbin.org/drip?numbytes=10&duration=10&delay=1&code=200"}
 (NOBRIDGE) LOG  LOG: Reader: {"_closedPromise": {"_h": 0, "_i": 0, "_j": null, "_k": null}, "_closedPromise_reject": [Function anonymous], "_closedPromise_resolve": [Function anonymous], "_ownerReadableStream": {"_disturbed": false, "_readableStreamController": {"_cancelAlgorithm": [Function anonymous], "_closeRequested": false, "_controlledReadableStream": [Circular], "_pullAgain": false, "_pullAlgorithm": [Function anonymous], "_pulling": false, "_queue": [e], "_queueTotalSize": 0, "_started": false, "_strategyHWM": 1, "_strategySizeAlgorithm": [Function anonymous]}, "_reader": [Circular], "_state": "readable", "_storedError": undefined}, "_readRequests": {"_back": {"_elements": [Array], "_next": undefined}, "_cursor": 0, "_front": {"_elements": [Array], "_next": undefined}, "_size": 0}}
 (NOBRIDGE) LOG  Received chunk: *
 (NOBRIDGE) LOG  Trying to abort the stream
 (NOBRIDGE) LOG  Received chunk: *
 (NOBRIDGE) LOG  Trying to abort the stream
 (NOBRIDGE) LOG  Received chunk: *
 (NOBRIDGE) LOG  Trying to abort the stream
 (NOBRIDGE) LOG  Received chunk: *
 (NOBRIDGE) LOG  Trying to abort the stream
 (NOBRIDGE) LOG  Received chunk: *
 (NOBRIDGE) LOG  Trying to abort the stream
 (NOBRIDGE) LOG  Received chunk: *
 (NOBRIDGE) LOG  Trying to abort the stream
[...]

Issue #33549, opened 2024-12-09
Fetch doesn't cancel the stream


hollanderbart:
### Minimal reproducible example

https://github.com/hollanderbart/BusinessCards-Connect

### What platform(s) does this occur on?

iOS

### Where did you reproduce the issue?

in a development build

### Describe the bug

When I've created a server fetch stream and I want to cancel the stream using an AbortController, the stream isn't cancelled. The stream still continues after calling the `abort()` function. I expect the current stream to be cancelled.

In my reproduction repo the stream is started but keeps returning messages, although `abortController.abort()` is called after the first message.

```
(NOBRIDGE) LOG  LOG: Response: {"redirected": false, "status": 200, "statusText": "no error", "url": "https://httpbin.org/drip?numbytes=10&duration=10&delay=1&code=200"}
 (NOBRIDGE) LOG  LOG: Reader: {"_closedPromise": {"_h": 0, "_i": 0, "_j": null, "_k": null}, "_closedPromise_reject": [Function anonymous], "_closedPromise_resolve": [Function anonymous], "_ownerReadableStream": {"_disturbed": false, "_readableStreamController": {"_cancelAlgorithm": [Function anonymous], "_closeRequested": false, "_controlledReadableStream": [Circular], "_pullAgain": false, "_pullAlgorithm": [Function anonymous], "_pulling": false, "_queue": [e], "_queueTotalSize": 0, "_started": false, "_strategyHWM": 1, "_strategySizeAlgorithm": [Function anonymous]}, "_reader": [Circular], "_state": "readable", "_storedError": undefined}, "_readRequests": {"_back": {"_elements": [Array], "_next": undefined}, "_cursor": 0, "_front": {"_elements": [Array], "_next": undefined}, "_size": 0}}
 (NOBRIDGE) LOG  Received chunk: *
 (NOBRIDGE) LOG  Trying to abort the stream
 (NOBRIDGE) LOG  Received chunk: *
 (NOBRIDGE) LOG  Trying to abort the stream
 (NOBRIDGE) LOG  Received chunk: *
 (NOBRIDGE) LOG  Trying to abort the stream
 (NOBRIDGE) LOG  Received chunk: *
 (NOBRIDGE) LOG  Trying to abort the stream
 (NOBRIDGE) LOG  Received chunk: *
 (NOBRIDGE) LOG  Trying to abort the stream
 (NOBRIDGE) LOG  Received chunk: *
 (NOBRIDGE) LOG  Trying to abort the stream
[...]

PR #47573, opened 2026-07-07
[expo][fetch] Propagate AbortSignal cancellation into the body stream

idoyana:
# Why

`expo/fetch` cancellation via an `AbortSignal` is incomplete: `fetch()`'s abort listener calls `request.cancel()` (native only) and never propagates the cancellation to the JS `ReadableStream` controller. Two consequences:

1. **The in-flight body read hangs forever** — the controller is never closed or errored, so a pending `reader.read()` (or `res.text()`/`res.json()` reading the stream) never settles, and it certainly never rejects with a spec `AbortError`.
2. **The enqueue-after-teardown race** — because the teardown guard is never set on the abort path, a late multithreaded native `didReceiveResponseData` is enqueued onto the abandoned controller, throwing `TypeError: The stream is not in a state that permits enqueue` (or `…permits close` on the completion event) from inside the native `addListener` callback. The throw is on a detached async stack, so the caller's `try/catch` cannot catch it — it surfaces as a fatal unhandled error.

This is the same teardown race tracked in #34804, #33549, and #33553. #44909 fixed it for cancellation via the JS stream (`reader.cancel()`), but the **`AbortSignal` path is still uncovered** — and `AbortSignal` is the common trigger (e.g. RTK Query aborts superseded/unmounted queries, `for await (…of res.body) break`, request timeouts).

We hit both failure modes in production (an aborted RTK boot query): the loud `permits enqueue`/`permits close` `TypeError` intermittently, and — because a clean native cancel drops the response silently — dropped cold-start queries with no error at all. Our only mitigation was to opt the whole app out of the global `expo/fetch` (`EXPO_PUBLIC_USE_RN_FETCH=1`), which this fix lets us remove.

## A second hole in the same class (added 2026-08-04)

We shipped this patch in our production app while the PR sat in review, and Sentry caught the teardown race **recurring through a path the flag cannot see**: when `pull()` rejects (a failed native `startStreaming` call), the streams implementation errors the stream through its *internal* error routine — `controller.error()` is never invoked, so `bodyStreamClosed` is never set. The trailing `didComplete` then passes the stale guard and calls `close()` on an errored stream: `TypeError: The stream is not in a state that permits close`, again out of the native event listener where no consumer `try/catch` can reach.

A boolean mirror cannot arbitrate a state machine it doesn't own; the guard has to ask the controller. Reproduced deterministically against both `web-streams-polyfill@3.3.3` (what Expo Go / `expo/virtual/streams.js` runs) and Node's `node:stream/web`: one rejected `pull` produces the `permits enqueue` + `permits close` pair verbatim on the old logic, and zero escaped throws on the new logic — with the reader still rejecting with the real underlying error.

# How

- Add `FetchResponse.abort(reason?)`, invoked from `fetch()`'s abort listener **before** `request.cancel()`. It errors the body-stream controller with the signal's reason (a spec `AbortError` when the signal provides none) and closes the teardown guard, so:
  - the pending read rejects with the `AbortError`, and
  - any subsequent native event (`didReceiveResponseData`/`didComplete`/`didFailWithError`) is a guarded no-op.
- Hoist the teardown guard (`isControllerClosed`) from a per-getter closure to instance state (`bodyStreamClosed` + a captured `bodyStreamController`), so `abort()` — which runs outside the `body` getter closure — can reach the live controller. `abortReason` also covers an abort that lands **before** the body stream is created (the getter errors the controller immediately on first access).
- Throw the signal's `reason` from the pre-flight `signal.aborted` early-return too, for consistent `AbortError` semantics.
- **(2026-08-04)** Route every controller call from the native listeners and `pull` through guarded helpers (`closeBodyStream`/`errorBodyStream`/`enqueueBodyChunk`): `bodyStreamClosed` stays as the fast path, `controller.desiredSize == null` detects the internally-errored state the spec exposes, and `closed`/`closeRequested` — observable only by trying — are absorbed by a `try/catch`. The `enqueue` guard removes the *cause*, not just the symptom: throwing from a late enqueue is precisely what rejects the in-flight `pull()` and errors the stream behind the flag's back. In `cancel()`, set the flag before `cancelStreaming()` — `ReadableStreamCancel` has already moved the stream to `closed` when the callback runs.

No behavior change for the non-abort paths; the existing `reader.cancel()`/`didComplete`-after-cancel guards (#44909) are preserved.

# Test Plan

`FetchResponse-test.ts` — added an `abort()` describe block covering:
- rejects the in-flight read with an `AbortError`,
- rejects a read started after abort with an `AbortError`,
- does not deliver or throw on native data arriving after abort (the teardown race),
- preserves a provided abort reason,
- idempotent / safe to call twice.

**(2026-08-04)** — added an `internally errored stream (rejected pull)` describe block covering:
- the reader still surfaces the underlying native error,
- no throw on `didReceiveResponseData`/`didComplete`/`didFailWithError` arriving after `pull()` rejected (the stale-guard race; both fail against the pre-guard code).

All existing fetch tests continue to pass (`yarn jest FetchResponse` in `packages/expo`). The change also removes the loud `permits enqueue`/`permits close` errors and the silent dropped-query behavior in our production app.


idoyana:
Hi @Kudo 👋 — you reviewed and merged #44909 (the `reader.cancel()` teardown fix), and this PR extends that same guard to the `AbortSignal` path, so you may be the right person to look.

The short version: #44909 stops the enqueue-after-teardown race when the *consumer cancels the reader*, but an `AbortSignal` abort still only cancels the native request — the JS `ReadableStream` controller is left open. So the in-flight body read hangs forever, and a late native `didReceiveResponseData` gets enqueued onto an abandoned controller (`The stream is not in a state that permits enqueue`) — the crashes tracked in #34804 / #33549 / #33553.

This adds `FetchResponse.abort()`, called from `fetch()`'s abort listener: it errors the body controller with the signal's reason (a spec `AbortError`) and closes the same teardown guard #44909 introduced — hoisted to instance state so it also covers an abort that lands *before* the body stream is created. Regression tests for the abort path are added to `FetchResponse-test.ts`.

Would you be up for reviewing, or pointing me to the right owner? Happy to adjust anything. Thanks! 🙏

idoyana:
Hi @Kudo 👋 — gentle nudge on this one, ~2 weeks since you added yourself. It's still green and conflict-free against main. If the diff is bigger than you'd like or anything about the approach gives you pause, happy to split it or rework — just say the word. Thanks! 🙏

idoyana:
Hi @zoontek 👋 — thanks for landing #47796! With that merged, this PR is the last remaining piece of the fetch delivery/teardown story: #42161 fixed the Android pump race, #47796 the iOS delegate-queue race, #44909 the consumer-cancel teardown — and this one closes the `AbortSignal` path on the JS side (abort currently cancels only the native request, leaving the `ReadableStream` controller open: the pending read hangs forever, and a late `didReceiveResponseData` hits an abandoned controller — the crash in #34804 / #33549 / #33553). Rebased yesterday, green and conflict-free, with regression tests for the abort path. 🙏

idoyana:
@zoontek heads-up before you look — I pushed one commit (4443d4e) extending this to a second hole in the same teardown class, caught by Sentry in our production app while the PR sat in review.

**The hole:** when `pull()` rejects (failed native `startStreaming` call), the streams implementation errors the stream through its *internal* error routine — `controller.error()` is never invoked, so `bodyStreamClosed` stays stale. The trailing `didComplete` then passes the guard and calls `close()` on an errored stream: `TypeError: The stream is not in a state that permits close`, thrown from the native event listener where no consumer `try/catch` can reach.

**The fix:** the listeners and `pull` now go through three small guarded helpers instead of trusting the boolean — `desiredSize == null` detects the internally-errored state, `try/catch` absorbs the states the spec doesn't expose. The `enqueue` guard is the one that removes the cause: throwing from a late enqueue is exactly what rejects the in-flight `pull()` and errors the stream behind the flag's back.

Reproduced deterministically on both `web-streams-polyfill@3.3.3` and `node:stream/web` — one rejected `pull` yields the `permits enqueue` + `permits close` pair on the old logic, zero escaped throws on the new, reader still rejects with the real error. Two regression tests added; PR description updated with the details.

Same shape as the rest of the series (#42161 → #47796 → #44909 → this): the JS controller has to be told what native did — and asked what state it's actually in.

zoontek:
@idoyana Please fix the formatting issue

idoyana:
Fixed in 943e5467 — `oxfmt` wanted the `bodyStreamController` declaration wrapped (line width). `pnpm oxfmt --check .` passes clean on the package now. The remaining `macos-build` failure is the repo-wide one from yesterday, unrelated to this branch.

packages/expo/src/winter/fetch/fetch.ts
```diff
@@ -75,9 +75,14 @@ export async function fetch(
   };
 
   if (signal && signal.aborted) {
-    throw new FetchError('The operation was aborted.');
+    throw signal.reason ?? new FetchError('The operation was aborted.');
```
zoontek: @idoyana Callers that check `error instanceof FetchError` on this path stop matching, silently, because the thrown value changes from a `FetchError` to the raw `signal.reason`.

Replaces it with `throw new FetchError('The operation was aborted.', { cause: signal.reason });`

zoontek (dismissed):

zoontek:
@idoyana Thanks for the PR. While you are handling the requested change, could you also trim comments down? They are great but a bit long (AI tends to be verbose)

packages/expo/src/winter/fetch/fetch.ts
idoyana: Fixed in 1a0649dd, exactly as suggested. Verified the break locally first: an aborted signal always carries a `reason` (default `AbortError` DOMException), so the `?? new FetchError(...)` fallback never fired — this path always threw the raw reason, and `instanceof FetchError` was `false`. With the change it's `true` again, the `fetch failed:` message contract holds, and `cause` preserves the reason (checked for both the default AbortError and a custom abort reason). The body-stream read rejections keep their `AbortError` semantics — untouched, matching the scope of your comment.

idoyana:
Trimmed in e4a6c470 — comment volume cut roughly in half (−24 lines), keeping one line per invariant the code can't show. The `instanceof FetchError` fix from your review is in 1a0649dd.

zoontek (approved):

PR #47575, opened 2026-07-07
[core][android] - Add fallback Color type converter below API 26

### LINKED ISSUE #47546: [ui][Android] @expo/ui is unusable on Android 7 (API 24/25): no type converter for android.graphics.Color

https://github.com/anasvemmully/expo-ui-android7-repro

### Steps to reproduce

1. Clone https://github.com/anasvemmully/expo-ui-android7-repro (a fresh `create-expo-app` blank template on SDK 57 with `@expo/ui` added; `App.js` renders one `@expo/ui/community/slider` component).
2. `npm install`, then build with `npx expo run:android` (dev) or `cd android && ./gradlew assembleRelease` (release) — the failure reproduces in both.
3. Run it on an **Android 7.0 / 7.1 device or emulator (API 24 or 25)**, e.g. an AVD from the `android-24 | Google APIs ARM 64` system image. Android 8+ (API 26+) is unaffected.
4. Open the app. In a release build it crashes immediately on launch; in a dev build the screen fails with a component error.

**Expected:** the component renders, since the Expo SDK's `minSdkVersion` is 24 and `@expo/ui` documents no higher requirement.

**Actual:** two failures, both rooted in the same missing type converter:

a) `Host` calls `getMaterialColors()` on mount, which rejects and (being called synchronously in a `useMemo` with no error handling) takes down the whole screen. **In a release build this is a fatal, app-killing crash** — verified on an API 24 emulator:

```
E AndroidRuntime: FATAL EXCEPTION: mqt_v_native
E AndroidRuntime: com.facebook.react.common.JavascriptException: Error: Call to function 'ExpoUI.getMaterialColors' has been rejected.
E AndroidRuntime: → Caused by: Cannot find type converter for 'class android.graphics.Color?'. Make sure the class implements `expo.modules.kotlin.records.Record` (i.e. `class MyObj : Record`).
E AndroidRuntime: getMaterialColors@1:539885
```

In a development build the same failure surfaces as a component error:

```
ERROR  [Error: Call to function 'ExpoUI.getMaterialColors' has been rejected.
→ Caused by: The 1st argument cannot be cast to type class expo.modules.ui.colors.MaterialColorsOptions? (received class com.facebook.react.bridge.ReadableNativeMap)
→ Caused by: Cannot find type converter for 'class android.graphics.Color?'. Make sure the class implements `expo.modules.kotlin.records.Record` (i.e. `class MyObj : Record`).]
```

b) Any `Color`-typed prop (e.g. the Slider's `colors` record) is rejected on every prop set:

```
ERROR  Cannot set prop 'color' on view 'class expo.modules.kotlin.views.ComposeFunctionHolder'
→ Caused by: Cannot find type converter for 'class android.graphics.Color?'. Make sure the class implements `expo.modules.kotlin.records.Record` (i.e. `class MyObj : Record`).
```

**Root cause:** expo-modules-core's converter for `android.graphics.Color` relies on the class-based Color API introduced in API 26 (`Color.valueOf`, etc.), so the converter is unavailable on API 24/25. Since the SDK officially supports `minSdkVersion 24`, every `Color?`-typed argument or prop in `@expo/ui`'s Android implementation fails on those devices.

**Workaround** (patch-package on `@expo/ui`):
- wrap `getMaterialColors()` in try/catch and return the static Material 3 baseline palette on failure, and
- omit `Color`-typed props on `Platform.Version < 26` so components fall back to default Material colors.

This makes `@expo/ui` degrade gracefully on Android 7 (default palette instead of custom colors). Happy to share the patch. A converter fallback to int-based colors on the native side, or an explicit documented minimum of API 26 for `@expo/ui`, would both resolve the ambiguity.

### Environment

```text
    System:
      OS: macOS 26.5.1
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 24.16.0 - /Users/kitchen/.nvm/versions/node/v24.16.0/bin/node
      Yarn: 1.22.22 - /Users/kitchen/.nvm/versions/node/v24.16.0/bin/yarn
      npm: 11.13.0 - /Users/kitchen/.nvm/versions/node/v24.16.0/bin/npm
    Managers:
    SDKs:
      iOS SDK:
      Android SDK:
        API Levels: 33, 34, 36, 36
        Build Tools: 35.0.0, 36.0.0, 36.1.0, 37.0.0
        System Images: android-24 | Google APIs ARM 64 v8a, android-37.0 | 16 KB Page Size Google Play ARM 64 v8a, android-37.1 | 16 KB Page Size Google Play ARM 64 v8a
    IDEs:
      Android Studio: 2026.1 AI-261.23567.138.2611.15613797
    npmPackages:
      expo: ^57.0.2 => 57.0.2
      react: 19.2.3 => 19.2.3
    Expo Workflow: bare

@expo/ui: 57.0.3
Device: Android 7.0 emulator (API 24, Google APIs ARM64) — also reproduced on a
physical Redmi Note 4 (Android 7.0). Not reproducible on Android 8+ devices.
```

### Expo Doctor Diagnostics

```text
```

intergalacticspacehighway:
# Why

`@expo/ui` is unusable on Android 7.0/7.1 (API 24/25) because expo-modules-core doesn't register a type converter for `android.graphics.Color` below API 26, so every `Color?` arg/prop throws `MissingTypeConverter` and `Host`'s `getMaterialColors` call takes down the screen.

Fixes - https://github.com/expo/expo/issues/47546

# How

- The class-based `android.graphics.Color` API (`Color.valueOf`, the float accessors) only exists on API 26+, so a real `Color` can't be built below it. Add a fallback color compat convertor that works on API level 24. https://github.com/expo/expo/pull/47575/#issuecomment-4903182492

# Test Plan

Tested running the bare-expo and Expo UI screens in Android 7 emulator.

# Checklist


Kudo:
let me find if there's any existing polyfill for `Color.valueOf`

docs/pages/versions/unversioned/sdk/ui/jetpack-compose/colors.mdx
```diff
@@ -15,7 +15,9 @@ The palette source depends on the options you pass:
 
 - **Wallpaper-derived:** Default on Android 12+ when no `seedColor` is given. Uses [Material 3 Dynamic Colors](https://m3.material.io/styles/color/dynamic-color/overview) (Material You).
 - **Static [Material 3 baseline](https://m3.material.io/styles/color/roles):** Default fallback on Android 11 and below when no `seedColor` is given.
-- **Seeded from a color:** When you pass a `seedColor`, the full palette is derived using the same algorithm that Material 3 Dynamic Colors use for wallpaper-based colors. Works on every Android API level and is independent of the wallpaper.
+- **Seeded from a color:** When you pass a `seedColor`, the full palette is derived using the same algorithm that Material 3 Dynamic Colors use for wallpaper-based colors. It is independent of the wallpaper and works on Android 8.0 (API 26) and above.
+
+> **Note:** Custom colors require Android 8.0 (API 26) or newer. On Android 7 (API 24 and 25), a `seedColor` and component color props are ignored. The palette falls back to the Material 3 baseline, and components use their default colors.
```
amandeepmittal:
```suggestion
> **Note:** Custom colors require Android 8.0 (API 26) or newer. On Android 7 (API 24 and 25), since `seedColor` and component color props are ignored, the palette falls back to the Material 3 baseline, and components use their default colors.
```

the last sentence should be part of the _what_ happens when seed color and other props are ignored.

amandeepmittal (changes requested):

Kudo:
i've changed my mind a little bit and thinked differently. how about add a `ColorCompat` wrapper instead?
this is the change https://gist.github.com/Kudo/a9f7d37f9035aefa429c083f550887ab
cc @lukmccall for your input

intergalacticspacehighway:
> how about add a ColorCompat wrapper instead?

I like it!

lukmccall:
I think that's a good idea! We can introduce a color wrapper that will work on older devices.

intergalacticspacehighway:
@Kudo @lukmccall added the ColorCompat wrapper. Kudo's patch https://github.com/expo/expo/pull/47575/commits/8e809e117e8b4f12c8a7272c6c0943d8b004423c

docs/pages/versions/unversioned/sdk/ui/jetpack-compose/colors.mdx
intergalacticspacehighway: Thanks. Reverted the docs changes as they are no longer needed with the new approach.

Kudo (approved):

amandeepmittal (approved):

intergalacticspacehighway:
[Added](https://github.com/expo/expo/pull/47575/commits/a0fab40f55eae7504a35a88a60b45819791f91a7) PlatformColor resolver for < API 26 . Should help here - https://github.com/expo/expo/pull/47739

intergalacticspacehighway:
CI is fixed here - https://github.com/expo/expo/pull/47742

PR #44967, opened 2026-04-21
[iOS][calendar][next] Add support for writeOnly permissions

Wenszel:
# Why

Adds an option to request only write permissions, requested in this issue: #33002

# How

- Added a `writeOnlyAccess` prop to the config plugin
- Added `CalendarWriteOnlyNextPermissionsRequester` and added other, corresponding requesters to the next package
-  Renamed: ExpoCalendarPermissions to `CalendarAccessGuard`, and added `CalendarPermissionRequester` to move the responsibilities from the module class.
- Removed throwing RCTFatal in the legacy module when FullAccess permissions are not defined in the infoPlist   

# Test Plan
- Added a new test screen to NCL
- Tested the config plugin using: 
`et gba --useLocalTemplate -n test-app expo-calendar` 
with this screen: [calendar.tsx](https://github.com/user-attachments/files/27474671/calendar.tsx)

Results: 
`writeOnlyAccess`: `true`:

https://github.com/user-attachments/assets/09c6bada-fa29-4fd9-a600-93f4b76ffcc2

`writeOnlyAccess`: `false`:

https://github.com/user-attachments/assets/aaff4658-6c3b-4aa8-a6b1-2525046fb5a9

PR #47582, opened 2026-07-07
[docs] revise iOS permissions for calendar access

msabaeian:
**Note:** PR description is AI-generated with human check and verification.

Updated iOS permissions documentation for calendar and reminders access, including new keys for iOS 17+.

# Why

Apps using `expo-calendar` crash immediately on launch on iOS 17+ with the error: `ExpoCalendar.MissingCalendarPlistValueException error 1`

The current documentation is missing `NSRemindersFullAccessUsageDescription`, which is **required** on iOS 17+ and checked during module initialization (not just when calendar methods are called). This causes apps to crash on startup before any user interaction.

**Source code reference:**

[`RemindersPermissionRequester.swift` (line 20-32)](https://github.com/expo/expo/blob/main/packages/expo-calendar/ios/Requesters/RemindersPermissionsRequester.swift#L20-L32) checks for `NSRemindersFullAccessUsageDescription` on iOS 17+ and if the key is missing, it calls `RCTFatal()` during the `OnCreate` lifecycle event

This issue affects any app using `expo-calendar` on iOS 17+ that accesses reminders or initializes the calendar module.

# How

1. Added `NSRemindersFullAccessUsageDescription` to the manual configuration examples for both full access and write-only access scenarios
2. Added prominent warnings explaining that apps will crash on iOS 17+ if these keys are missing
3. Updated the iOS Permissions section to include `NSRemindersFullAccessUsageDescription` with proper description
4. Clarified that permission checks happen during module initialization, not just when methods are called

# Test Plan

1. Create a React Native app with `expo-calendar` v55.x
2. Add only the old permissions: `NSCalendarsUsageDescription`, `NSCalendarsFullAccessUsageDescription`, `NSRemindersUsageDescription`
3. Run on iOS 17+ simulator/device
4. App crashes immediately on launch with a release build and shows a red error modal in debug mode with `MissingCalendarPlistValueException`
SCR-20260707-kvsj

**Verification of the fix:**
1. Add `NSRemindersFullAccessUsageDescription` to Info.plist
2. Rebuild the app
3. App launches successfully without crashes

# Checklist

### DIFF
--- a/docs/pages/versions/unversioned/sdk/calendar.mdx
+++ b/docs/pages/versions/unversioned/sdk/calendar.mdx
@@ -83,7 +83,11 @@ You can configure `expo-calendar` using its built-in [config plugin](/config-plu
 
 If you're not using Continuous Native Generation ([CNG](/workflow/continuous-native-generation/)) (you're using native **ios** project manually), then you need to configure following permissions in your native project:
 
-- For iOS, add `NSCalendarsUsageDescription`, `NSCalendarsFullAccessUsageDescription`, and `NSRemindersUsageDescription` to your project's **ios/[app]/Info.plist**:
+- For iOS, add the following keys to your project's **ios/[app]/Info.plist**:
+
+  > **warning** On iOS 17+, the app will crash on startup if these keys are missing. The module checks for these permissions during initialization, not just when calendar methods are called.
+
+  **For full calendar and reminders access:**
 
   ```xml
   <key>NSCalendarsUsageDescription</key>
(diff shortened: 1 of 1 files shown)


docs/pages/versions/unversioned/sdk/calendar.mdx
```diff
@@ -83,7 +83,11 @@ You can configure `expo-calendar` using its built-in [config plugin](/config-plu
 
 If you're not using Continuous Native Generation ([CNG](/workflow/continuous-native-generation/)) (you're using native **ios** project manually), then you need to configure following permissions in your native project:
 
-- For iOS, add `NSCalendarsUsageDescription`, `NSCalendarsFullAccessUsageDescription`, and `NSRemindersUsageDescription` to your project's **ios/[app]/Info.plist**:
+- For iOS, add the following keys to your project's **ios/[app]/Info.plist**:
+
+  > **warning** On iOS 17+, the app will crash on startup if these keys are missing. The module checks for these permissions during initialization, not just when calendar methods are called.
+
+  **For full calendar and reminders access:**
```
amandeepmittal, line 90: ✏️ **style** `mechanics.md#file-and-directory-names-in-bold`

Bold marks file and directory names on these pages, so a bold line reads as one rather than as a heading. Fold it into the sentence that introduces the block, such as "For full calendar and reminders access, add:", and do the same for `**For write-only calendar access (iOS 17+):**` below.

docs/pages/versions/unversioned/sdk/calendar.mdx
```diff
@@ -195,5 +209,10 @@ The following usage description keys are used by this library:
         "A message that tells the user why the app is requesting write-only access to the user's calendar data.",
     },
     'NSRemindersUsageDescription',
+    {
+      name: 'NSRemindersFullAccessUsageDescription',
+      description:
+        "A message that tells the user why the app is requesting full access to the user's reminders data. Required on iOS 17+.",
```
amandeepmittal, line 215: 💡 **nit** `mechanics.md#sdk-version-ranges`

Write "Required on iOS 17 and later when your app accesses reminders", since the docs spell version ranges out rather than using `+`, and the `NSCalendarsFullAccessUsageDescription` entry above carries no unconditional requirement note.

docs/pages/versions/unversioned/sdk/calendar.mdx
```diff
@@ -83,7 +83,11 @@ You can configure `expo-calendar` using its built-in [config plugin](/config-plu
 
 If you're not using Continuous Native Generation ([CNG](/workflow/continuous-native-generation/)) (you're using native **ios** project manually), then you need to configure following permissions in your native project:
 
-- For iOS, add `NSCalendarsUsageDescription`, `NSCalendarsFullAccessUsageDescription`, and `NSRemindersUsageDescription` to your project's **ios/[app]/Info.plist**:
+- For iOS, add the following keys to your project's **ios/[app]/Info.plist**:
+
+  > **warning** On iOS 17+, the app will crash on startup if these keys are missing. The module checks for these permissions during initialization, not just when calendar methods are called.
```
amandeepmittal, line 88: ⚠️ **correctness** `packages/expo-calendar/ios/Requesters/RemindersPermissionsRequester.swift`

The crash is gone from the SDK this file documents. `RCTFatal(MissingCalendarPListValueException(...))` exists only up to the sdk-55 branch and was removed by #44967 for SDK 56, so on main a missing key makes `getPermissions()` return `denied` instead: the app starts, and the calls that need the permission fail. Reword the callout around denied permissions, and keep the crash warning for the SDK versions where it still applies.

---

"the app will crash" is future tense. Docs describe behavior in the present, so write what happens now rather than what will happen.

```suggestion
  > **warning** On iOS 17 and later, `expo-calendar` reports calendar and reminders permissions as denied when the matching full access key is missing, so any call that needs them fails. The module reads these keys when it initializes, not only when you call a calendar method.
```

amandeepmittal (changes requested):
Thanks for updating docs!

Left a few comments.

PR #47586, opened 2026-07-07
[tools] Add prepare-branch command

alanjhughes:
# Why
Before publishing from an SDK branch, for most consistent results, I've found these steps necessary. Clean install node_modules, a clean react-native-lab checkout, resync pods in the native projects, and no stale build caches. Doing this by hand is a series of easy-to-forget commands, so this adds `et prepare-branch` to automate it.

# How
The command runs six steps in order, stopping on the first failure:

1. Remove the root `node_modules` and run `git checkout .` to reset tracked files.
2. `git clean -fdx` in `react-native-lab/react-native`, then `yarn`.
3. `pnpm install` at the repo root.
4. Remove `Pods` and `build` for bare-expo and Expo Go, then force a pod install to resync them with the native projects.
5. Remove `.DerivedData`, `.generated`, `.swiftpm` and `Products` from `packages/expo-modules-jsi/apple`.
6. Remove `.build` and `.cache` from `packages/precompile`.

Since step 1 discards uncommitted changes, the command aborts up front if the working tree is dirty and lists the offending files. Pass `-f` / `--force` to proceed anyway. It also warns (but continues) when the current branch is not an SDK branch.

To reuse the pod install logic in-process, `PodInstallCommand`'s action is now exported and returns a boolean instead of swallowing failures. The `et pods` behavior is unchanged.

# Test Plan
- Unit tests for the guard logic
- On a dirty tree, `et prepare-branch` aborts with the list of dirty files and deletes nothing.
- Ran the full command end to end on this branch: all six steps completed, node_modules reinstalled, react-native-lab clean with dependencies restored, `Podfile.lock` updated in both apps, and all build caches removed.

### DIFF
--- a/tools/src/commands/PrepareBranchCommand.ts
+++ b/tools/src/commands/PrepareBranchCommand.ts
@@ -1,0 +1,180 @@
+import { Command } from '@expo/commander';
+import chalk from 'chalk';
+import fs from 'fs-extra';
+import path from 'path';
+
+import { EXPO_DIR } from '../Constants';
+import * as Directories from '../Directories';
+import Git, { GitDirectory } from '../Git';
+import logger from '../Logger';
+import { spawnAsync } from '../Utils';
+import { action as podInstallAsync } from './PodInstallCommand';
+
+type PreconditionsInput = {
+  branchName: string;
+  dirtyFiles: string[];
+  force: boolean;
+};
+
+type PreconditionsResult = {
+  warning: string | null;
+  error: string | null;
+};
+
+export function validatePreconditions({
+  branchName,
+  dirtyFiles,
+  force,
+}: PreconditionsInput): PreconditionsResult {
+  const warning = /^sdk-\d+/.test(branchName)
+    ? null
+    : `Current branch "${branchName}" is not an SDK branch. Proceeding anyway.`;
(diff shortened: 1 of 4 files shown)


tools/src/commands/PrepareBranchCommand.ts
```diff
@@ -0,0 +1,170 @@
...
+      },
+    },
+    {
+      title: 'Removing expo-modules-jsi build artifacts',
+      async runAsync() {
+        for (const name of ['.DerivedData', '.generated', '.swiftpm', 'Products']) {
+          await fs.remove(path.join(jsiAppleDir, name));
+        }
+      },
+    },
```
tsapeta: Might be a good idea to make a dedicated script in `expo-modules-jsi` that clears its caches and artifacts. Also, I think we should run `pod install` after this step to create a stub `xcframework` (https://github.com/expo/expo/blob/main/packages/expo-modules-jsi/apple/scripts/create-stub-xcframework.sh)

alanjhughes: Makes sense. I'll follow up with that script 👍

tsapeta (approved):

PR #47590, opened 2026-07-07
fix(require-utils): drop `dynamic-import` babel plugin

krystofwoldrich:
Fixes the same issue as https://github.com/expo/expo/pull/47170 but for `sdk-57` branch which uses babel.

Purposely limited to require-utils, as it's not know to cause issues anywhere else.


packages/@expo/require-utils/babel.config.js
```diff
@@ -1,2 +1,11 @@
-// @generated by expo-module-scripts
-module.exports = require('expo-module-scripts/babel.config.cli');
+// ensure dynamic import are not transformed for require-utils
+// this is required to dynamically load ESM modules using `await import()`
+// fix on main https://github.com/expo/expo/pull/47170
+module.exports = function (api) {
+  const config = require('expo-module-scripts/babel.config.cli')(api);
+  config.plugins = config.plugins.filter((plugin) => {
+    const name = Array.isArray(plugin) ? plugin[0] : plugin;
+    return name !== 'babel-plugin-dynamic-import-node';
```
kitten, line 8: Should we remove this plugin from the config entirely? Haven't checked if it'll cause issues though

kitten (approved):
will add some backport labels 👀

packages/@expo/require-utils/babel.config.js
krystofwoldrich: Same thought, but since sdk-58 won't have it and the swc rule is disabled globally, I thought let's play it safe, minimal changes.

PR #47592, opened 2026-07-07
[app-intents][ios][5/n] Add visual intelligence support, extend examples in ncl

behenate:
# Why

Adds support for visual intelligence modifier, extends NCL examples.

# How

- Create a new ExpoUI modifier, which translates into a native swift modifier
- The modifier becomes available once `expo-app-intents` is installed in the project

# Test Plan

Tested in BareExpo on iOS 27


packages/expo-app-intents/ios/ExpoAppIntents.podspec
```diff
@@ -20,6 +20,7 @@ Pod::Spec.new do |s|
   s.static_framework = true
 
   s.dependency 'ExpoModulesCore'
+  s.dependency 'ExpoUI'
```
tsapeta: [🏗️ **design**] This podspec declares `:osx => '13.4'`, but `ExpoUI.podspec` declares only `:ios` and `:tvos`. CocoaPods refuses a dependency that does not support the target's platform, so any macOS target that autolinks `ExpoAppIntents` now fails `pod install` (and `import ExpoUI` in `ExpoAppIntentsModule.swift` would not compile there either). No app in this repo installs the pod on macOS, so CI won't catch it, but the podspec advertises macOS support and the new Swift carries `macOS 15.0/15.4` availability annotations.

Either drop `:osx` from the platforms, or scope the dependency (`s.ios.dependency 'ExpoUI'` / `s.tvos.dependency 'ExpoUI'`) and guard the ExpoUI-touching code with `#if canImport(ExpoUI)`.

packages/expo-app-intents/package.json
```diff
@@ -49,6 +49,7 @@
     "expo-module-scripts": "workspace:*"
   },
   "dependencies": {
+    "@expo/ui": "workspace:*",
```
tsapeta: [💡 suggestion] `workspace:*` publishes as an exact pinned version, so each `expo-app-intents` release would demand one exact `@expo/ui` version and could conflict with the version the app itself installs. The other packages that depend on `@expo/ui` use a range: `expo-router` has `workspace:^56.0.14` and `expo-widgets` has `workspace:~56.0.14`.

packages/expo-app-intents/ios/Tests/ExpoAppIntentsModuleTests.swift
```diff
@@ -61,4 +69,159 @@ struct ExpoAppIntentsModuleTests {
...
+  }
+}
+
+@Suite("AppEntityIdentifierRegistry")
+struct AppEntityIdentifierRegistryTests {
+  @Test
+  func `reindexing ignores a kind that is not registered for indexing`() async throws {
+    let kind = "testUnregistered-\(UUID().uuidString)"
+    let key = "dev.expo.appintents.entities.\(kind)"
+    UserDefaults.standard.set(Data("not json".utf8), forKey: key)
```
tsapeta: [💡 suggestion] This seeds the garbage into `UserDefaults.standard`, but `AppIntentEntityStore` reads from the `dev.expo.appintents` suite (`requireDefaults()` uses `userDefaultsSuiteName`). The store never sees the invalid data, so the test passes even if the `hasIndexer` guard in `replaceIndexFromCatalog` is removed: the suite read finds nothing and returns an empty catalog. Write through `UserDefaults(suiteName: AppIntentEntityStore.userDefaultsSuiteName)` so the test fails when the guard is gone.

packages/expo-app-intents/src/index.ts
```diff
@@ -233,3 +263,19 @@ export async function refreshShortcutsAsync(): Promise<void> {
   }
   return ExpoAppIntents.refreshShortcutsAsync();
 }
+
+/**
+ * Returns an ExpoUI SwiftUI modifier config that ties a view to an AppEntity identifier.
+ *
+ * The `entity` value must be registered from app-target Swift with
+ * `AppEntityIdentifierRegistry.shared.register(_:as:)`.
```
tsapeta: [🐧 nit] The visual-intelligence flow this PR documents registers the kind with `registerIndexed(_:as:)`, not `register(_:as:)`. Consider "with `register(_:as:)` or `registerIndexed(_:as:)`" so a reader following the scaffolded example doesn't think a second registration is required.

docs/pages/versions/unversioned/sdk/app-intents.mdx
```diff
@@ -196,12 +196,68 @@ Pass `--dir` to generate the module in a directory other than **app-intents**. T
 
 Keep the App Intents directory at the project root, like the default **app-intents** directory. This location keeps the native declarations easy to find and separate from the rest of the app.
 
+Pass `--visual-intelligence` with the mail example to scaffold its Spotlight integration. This adds native code that makes drafts searchable, lets the system transfer and open a draft, and registers the `mailDraft` entity kind so [`appEntityIdentifier()`](#appintentsappentityidentifierentity-id) can identify the draft represented by a view. The generated code extends the base mail types, so the example works the same way with or without this flag.
```
amandeepmittal: ✏️ **style** `prose.md#sentence-construction`

The second sentence runs 30 words across three coordinated clauses. Split the Spotlight behavior from the entity registration.

```suggestion
Pass `--visual-intelligence` with the mail example to scaffold its Spotlight integration. This adds native code that makes drafts searchable and lets the system transfer and open a draft. It also registers the `mailDraft` entity kind so [`appEntityIdentifier()`](#appintentsappentityidentifierentity-id) can identify the draft represented by a view. The generated code extends the base mail types, so the example works the same way with or without this flag.
```

docs/pages/versions/unversioned/sdk/app-intents.mdx
```diff
@@ -409,6 +465,17 @@ await AppIntents.setEntityCatalogAsync('dish', [
 
 The generated entity queries read these values through `await AppIntentEntityStore.shared.entities(ofKind:)`. Siri and Shortcuts can then offer and resolve them as intent parameters.
 
+Set `hideInSpotlight` on an entity to keep it out of the Spotlight index while leaving it resolvable, so Siri can still offer it as a parameter and open it by identifier:
```
amandeepmittal: ✏️ **style** `prose.md#core-principles`

This is one 30-word sentence carrying two ideas: what the flag does to the index, and what Siri can still do. Split it.

```suggestion
Set `hideInSpotlight` on an entity to keep it out of the Spotlight index while leaving it resolvable. Siri can still offer it as a parameter and open it by identifier:
```

docs/pages/versions/unversioned/sdk/app-intents.mdx
```diff
@@ -409,6 +465,17 @@ await AppIntents.setEntityCatalogAsync('dish', [
...
+
+```ts
+await AppIntents.setEntityCatalogAsync('mailDraft', [
+  { id: 'release-notes', title: 'Release notes for review' },
+  { id: 'salary-review', title: 'Salary review', hideInSpotlight: true },
+]);
+```
+
+The flag only applies to entities registered natively with `registerIndexed`. Flipping it takes effect the next time the catalog is published, and an entity that was already indexed is removed. To keep an entity away from Siri altogether, leave it out of the catalog instead — an entity that isn't published can't be offered, matched, or resolved.
```
amandeepmittal: 💡 **nit** `prose.md#sentence-construction`

A full stop instead of the em dash keeps the last sentence to one thought.

```suggestion
The flag only applies to entities registered natively with `registerIndexed`. Flipping it takes effect the next time the catalog is published, and an entity that was already indexed is removed. To keep an entity away from Siri altogether, leave it out of the catalog instead. An entity that isn't published can't be offered, matched, or resolved.
```

docs/pages/versions/unversioned/sdk/app-intents.mdx
```diff
@@ -19,9 +19,9 @@ import { Terminal } from '~/ui/components/Snippet';
...
 
-## Limitations
+## Apple Intelligence support
 
-With `expo-app-intents`, you can implement many Apple Intelligence features. The package does not currently support [on-screen intelligence](https://developer.apple.com/documentation/appintents/providing-contextual-cues-to-apple-intelligence-and-siri).
+With `expo-app-intents`, you can implement Apple Intelligence features, including [on-screen intelligence](https://developer.apple.com/documentation/appintents/providing-contextual-cues-to-apple-intelligence-and-siri). [`appEntityIdentifier()`](#appintentsappentityidentifierentity-id) marks the entity represented by a view so Siri can act on the visible content. You can apply it to [`@expo/ui`](/versions/latest/sdk/ui/) views on iOS 18.4 and later.
```
amandeepmittal: ⚠️ **correctness** `review-ai-text.md#t7-overclaiming`

`appEntityIdentifier()` also needs the project compiled with Xcode 27 or later, because below Swift 6.4 the modifier is registered as a no-op that logs a warning. This paragraph gives only the iOS 18.4 requirement, so a reader building with Xcode 26 has nothing that explains the missing association.

```suggestion
With `expo-app-intents`, you can implement Apple Intelligence features, including [on-screen intelligence](https://developer.apple.com/documentation/appintents/providing-contextual-cues-to-apple-intelligence-and-siri). [`appEntityIdentifier()`](#appintentsappentityidentifierentity-id) marks the entity represented by a view so Siri can act on the visible content. You can apply it to [`@expo/ui`](/versions/latest/sdk/ui/) views on iOS 18.4 and later, in a project compiled with Xcode 27 or later. In a project built with an earlier Xcode, the modifier does not apply the identifier and `expo-app-intents` logs a warning once.
```

packages/expo-app-intents/ios/AppEntityIdentifierModifier.swift
```diff
@@ -0,0 +1,525 @@
...
+  ///
+  /// On disk rather than in memory because the drift outlives the process that caused it: indexing
+  /// fails, the app is killed, and the next launch republishes the same catalog — which the entity store
+  /// short-circuits, because the catalog itself did not change. An in-memory flag would be gone by
+  /// then, and the index would stay stale until something called `reindexEntitiesAsync`.
+  ///
+  /// A key of the registry's own, not one `AppIntentEntityStore` hands out, because staleness is a fact
+  /// about the Spotlight index and the store knows nothing about Spotlight. It does mean that moving
+  /// the store to an App Group suite has to move this too.
+  private let defaults = UserDefaults.standard
```
tsapeta: [💡 suggestion] The stale flag lives in `UserDefaults.standard` while the catalog it describes lives in the `dev.expo.appintents` suite. The doc comment already notes that moving the store to an App Group would have to move this too, which is the coupling to avoid: anything that clears or migrates the suite (tests included, the module tests clear the suite between runs) leaves this flag behind in the app's own defaults. Writing it to `UserDefaults(suiteName: AppIntentEntityStore.userDefaultsSuiteName)` keeps every piece of package state in one place without the store having to know about Spotlight; the key prefix already distinguishes it.

packages/expo-app-intents/ios/AppEntityIdentifierModifier.swift
```diff
@@ -0,0 +1,525 @@
...
+      // as long as JavaScript kept republishing the same records - and republishing an identical
+      // catalog on every app start is the normal case.
+      guard didChangeCatalog || self.isIndexStale(kind: kind) else {
+        return didChangeCatalog
+      }
+
+      do {
+        try await self.performIndex(kind: kind, records: records, update: .replaceEverything)
+      } catch {
+        log.error("expo-app-intents: could not update the Spotlight index for '\(kind)': \(error)")
```
tsapeta: [💡 suggestion] This is the one failure a developer most needs to see (the catalog was stored but Spotlight was not updated), and it goes to the global `log`, which the `AppEntityIdentifierDiagnostics` comment in this same file describes as the place where the developer would never see it. `setEntityCatalogAsync` has `appContext?.jsLogger` at hand; passing a `Logger` into `publishCatalog` (defaulting to `log` for the `IndexedEntityQuery` callers that have no context) would put this line in Metro and LogBox, consistent with the rest of the file's reasoning.

packages/expo-app-intents/ios/AppEntityIdentifierModifier.swift
```diff
@@ -0,0 +1,525 @@
...
+  /// Only ever called from inside the kind's `indexing` section, because the flag has to be maintained
+  /// there too and not merely around it. Marking outside the section lets two updates interleave their
+  /// mark/clear pairs, so a short update that succeeds can clear the flag a long one that failed had
+  /// set - reporting a half-written index as matching its catalog, permanently, because an unchanged
+  /// catalog short-circuits every later publish.
+  ///
+  /// The kind is marked stale before the work starts rather than after a failure, so that a process
+  /// killed part way through an update - after the delete, before the index - is caught on the next
+  /// launch as well.
+  private func performIndex(
```
tsapeta, line 319: [💡 suggestion] The stale-flag state machine here and in `publishCatalog` is the most intricate part of the PR (mark before work, clear only on full replace or when the refresh did not inherit an earlier failure, retry an unchanged catalog when stale), and none of it is exercised by the tests. `KeyedSerialQueue`, the claims, and the diagnostics are covered; this is not. The indexer closure is built inside `registerIndexed`, so there is no seam to fail it on purpose. An internal `register(kind:indexer:)` overload that takes an `EntityIndexer` directly (with `registerIndexed` calling it) would let a test assert: index throws → flag set → republishing the identical catalog reindexes → flag cleared; and a `.refreshOnly` that succeeds while the kind was already stale leaves the flag set.

packages/expo-app-intents/ios/AppEntityIdentifierModifier.swift
```diff
@@ -0,0 +1,525 @@
...
+/// so a delete from the second could wipe what the first had just put in, and the index would end up
+/// matching neither catalog. Updates to different kinds touch disjoint parts of the index, so they are
+/// free to run at the same time.
+internal final class KeyedSerialQueue: @unchecked Sendable {
+  private struct Tail {
+    let id: UUID
+    let task: Task<Void, Never>
+  }
+
+  private let lock = NSLock()
```
tsapeta: [🐧 nit] Two locking primitives in one file: `KeyedSerialQueue`, the registry, the claims and the registration use `NSLock` + `withLock`, while `AppEntityIdentifierDiagnostics` a few hundred lines down uses `ExpoModulesCore.Mutex`, which is also what `AppIntentEventSubscription` in this package already uses. `Mutex<Value>` would also let the guarded state be the value itself (`Mutex<[String: Tail]>`) instead of a lock sitting next to a `var`. Pick one; `Mutex` matches the existing code.

packages/expo-app-intents/ios/ExpoAppIntents.podspec
```diff
@@ -31,5 +33,8 @@ Pod::Spec.new do |s|
   s.test_spec 'Tests' do |test_spec|
     test_spec.dependency 'ExpoModulesTestCore'
     test_spec.source_files = 'Tests/**/*.{m,swift}'
+    # ExpoUI pulls in C++ through React, so the test target has to link the C++ runtime or the
+    # link step fails on operator new and the __cxa_* personality symbols.
+    test_spec.libraries = 'c++'
```
tsapeta, line 39: [💡 suggestion] No other `test_spec` in the repo sets `libraries = 'c++'`, and the two other pods that depend on `ExpoUI` (`ExpoWidgets`, the module template) have no test spec to compare against. Did the `operator new` / `__cxa_*` link failure reproduce from a clean `/tmp/ExpoUnitTestsDerivedData`? I have seen the same undefined-symbol wave in this repo come from a stale `React-Core-prebuilt` build configuration rather than a missing libc++, and disappear without linker changes. If a clean run links without this line, drop it so the comment does not codify a machine-local artifact. If it is real, keep it, and the comment is a good record of why.

packages/expo-app-intents/ios/ExpoAppIntentsModule.swift
```diff
@@ -1,17 +1,29 @@
...
+
 internal final class ShortcutsRefreshUnavailableException: Exception, @unchecked Sendable {
   override var reason: String {
     "App Shortcuts could not be refreshed because no refresh handler is registered. "
-      + "The app target must contain an 'AppIntentsSetup' inline module that sets "
-      + "AppIntentDispatcher.shared.setShortcutsRefreshHandler(...). Run `npx expo-app-intents init` "
-      + "to generate it, or add it manually as described in the expo-app-intents documentation."
+      + "Only your app target can call AppShortcuts.updateAppShortcutParameters(), so the "
+      + "'AppIntentsSetup' inline module has to hand that call to "
+      + "AppIntentDispatcher.shared.setShortcutsRefreshHandler(...) — and it only does so for an "
```
tsapeta: [🐧 nit] Em dash inside a user-facing error string ("setShortcutsRefreshHandler(...) — and it only does so"). The docs review on this PR already asks for full stops instead of em dashes in prose; the same reads better in an error the developer sees in LogBox, and this one is long enough to benefit from a sentence break here anyway.

packages/expo-app-intents/ios/Tests/AppIntentEntityStoreTests.swift
```diff
@@ -33,7 +39,104 @@ struct AppIntentEntityStoreTests {
...
+      )
+    ]
+
+    let firstWrite = try await store.setCatalog(kind: "trail", entities: records)
+    #expect(firstWrite, "the first write introduces a catalog")
+
+    let identicalWrite = try await store.setCatalog(kind: "trail", entities: records)
+    #expect(!identicalWrite, "republishing the same catalog is not a change")
+
+    // Built from scratch rather than copied and mutated: `@Field` is a class, so assigning through
```
tsapeta: [🐧 nit] Is this comment accurate? `AppIntentEntityRecord` declares plain `var`s under `@Record`, with no `@Field` wrappers, and the core tests build `@Record` structs that are `Equatable`, which would not synthesize over class-typed wrappers. If a copy of the array can be mutated without touching `records`, the comment sends the next reader looking for reference semantics that are not there. Either verify and drop it, or keep the from-scratch construction with a comment that says it is just for clarity.

packages/expo-app-intents/ios/AppEntityIdentifierModifier.swift
behenate: Tests have been added

packages/expo-app-intents/ios/ExpoAppIntents.podspec
behenate: this is necessary and reproduced in a clean build. We already link lc++ (albeit slightly differently) in [expo-camera](https://github.com/expo/expo/blob/7e1e1cec9bc679f987583a810de5a7f560ad62de/packages/expo-camera/ios/ExpoCamera.podspec#L40-L42)
[expo-sqlite](https://github.com/expo/expo/blob/7e1e1cec9bc679f987583a810de5a7f560ad62de/packages/expo-sqlite/ios/ExpoSQLite.podspec#L77-L80)

tsapeta (approved):

amandeepmittal (approved):

PR #47593, opened 2026-07-07
[docs] Exclude redirect stub pages from sitemap.xml

amandeepmittal:
# Why

Fix ENG-23559

The generated `sitemap.xml` lists 19 redirect stub pages (`/develop`, `/get-started`, `/build`, ...) that serve an empty shell with a meta refresh, so crawlers that sample them see no content and the pages inflate the llms.txt coverage denominator in agent-docs measurements. Sitemaps should list canonical content URLs only. 

# How

- Add `findRedirectStubUrls` to `scripts/create-sitemap.js` to detect pages backed by a `common/redirect` `index.js` stub or an MDX page importing `components/plugins/Redirect`, and filter them from sitemap entries. The import path is the marker, not the `<Redirect>` tag, since content pages embed that tag in code examples.
- Pass `pagesDirectory` from `next.config.ts` so the export build applies the filter.
- Add tests in `scripts/create-sitemap.test.js` covering both stub kinds, content pages that only mention `<Redirect>` in examples, and the stub-free written XML.

# Test Plan

Run `node --experimental-strip-types --experimental-vm-modules ./node_modules/jest/bin/jest.js scripts/create-sitemap.test.js` (11 passing). 

On the preview deploy, `/sitemap.xml` drops from 1488 to 1469 entries and no longer lists `/develop/` or `/build/` while their redirect targets remain: https://pr-47593.expo-docs.pages.dev/sitemap.xml

# Checklist

### DIFF
--- a/docs/next.config.ts
+++ b/docs/next.config.ts
@@ -206,6 +206,7 @@ const nextConfig: NextConfig = {
       pathMap,
       domain: `https://docs.expo.dev`,
       output: join(outDir, `sitemap.xml`),
+      pagesDirectory: pagesDir,
       // Some of the search engines only track the first N items from the sitemap,
       // this makes sure our starting and general guides are first, and API index last (in order from new to old)
       pathsPriority: [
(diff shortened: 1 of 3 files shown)


entiendoNull (approved):
Nice!

Issue #47316, opened 2026-06-27
EXC_BAD_ACCESS in `EXTaskService _runTasksSupportingLaunchReason` on background relaunch (long-lived `startLocationUpdatesAsync` task)


ctmayhew:
## Summary

On iOS, an app that keeps a background **location** task registered via
`Location.startLocationUpdatesAsync` for many hours intermittently crashes with
`EXC_BAD_ACCESS (SIGSEGV)` inside expo-task-manager's own request-completion block —
**not** in any JS task handler. The crash happens while the app is launched in the
background (`Role: Non UI`) to service the task. It recurs across nights for users who
leave the task registered overnight (the OS cold-relaunches the app on movement
repeatedly, and one of those relaunches segfaults).

The faulting frame is the completion block created in
`-[EXTaskService _runTasksSupportingLaunchReason:userInfo:callback:]`, at the line that
removes the finished request from `_requests`:

```objc
// EXTaskService.m  (expo-task-manager 55.0.16)
request = [[EXTaskExecutionRequest alloc] initWithCallback:^(NSArray * _Nonnull results) {
    if (callback != nil) { callback(results); }
    [self->_requests removeObject:request];   // <-- line 531, faulting frame
    request = nil;
}];
```

`KERN_INVALID_ADDRESS at 0x0000000000000008` is an ivar access on a freed/near-null
object, i.e. the request (or the service) backing this block has been torn down by the
time the completion block runs — a native lifecycle race in the background-launch path.

## Environment

| | |
|---|---|
| expo | 55.0.26 |
| expo-task-manager | 55.0.16 |
| expo-location | 55.1.10 |
| react-native | 0.83.6 (New Architecture: Fabric + TurboModules, Hermes) |
| Platform | iOS 26.5 (23F77), iPhone 15 Pro (iPhone16,1), arm64e |
| Build | Release / TestFlight |

## Reproduction

1. Define a task with `TaskManager.defineTask(TASK_NAME, …)` at module scope.
2. `Location.startLocationUpdatesAsync(TASK_NAME, { accuracy, foregroundService, … })`
   with **Always** location permission.
3. Leave the task registered for many hours (e.g. overnight) so iOS repeatedly
   terminates and **cold-relaunches the app in the background** on movement / new fixes.
4. Intermittently, one background relaunch crashes in the task-completion block above
   rather than delivering the location event.

It is timing/lifecycle dependent (not every relaunch), so it reproduces statistically
over a long-lived registration, not on a single launch.

[...]

Issue #44709, opened 2026-04-12
expo-background-task: SIGSEGV in EXTaskExecutionRequest._maybeExecuteCallback when task completes during foreground transition


waltthirion:
## Summary

When a background task is in-flight and the user brings the app back to foreground, a native race condition causes a `SIGSEGV` crash (`KERN_INVALID_ADDRESS at 0x0000000000000008`) on the `expo.modules.ExpoTaskManagerQueue`.

## Environment

| | |
|---|---|
| `expo-background-task` | 1.0.8 |
| Expo SDK | 54 |
| Platform | iPadOS 26.4 (23E246) |
| Device | iPad Air 5th gen M1 — `iPad13,18` |
| Architecture | ARM64 |

## Reproduction steps

1. Register a background task with `BackgroundTask.registerTaskAsync`
2. Send the app to background (swipe up to home)
3. Background task fires and begins executing
4. Bring the app back to foreground before the task completes
5. App crashes with SIGSEGV — no JS error, no console output

In our case the task body includes a `Barometer.addListener` call with a 3-second timeout, creating a window during which the foreground transition can race with task completion.

## Crash stack (faulting thread — `expo.modules.ExpoTaskManagerQueue`)

```
EXC_BAD_ACCESS (SIGSEGV) — KERN_INVALID_ADDRESS at 0x0000000000000008

__67-[EXTaskService _runTasksSupportingLaunchReason:userInfo:callback:]_block_invoke
-[EXTaskExecutionRequest _maybeExecuteCallback]     <- crashes here (nil ivar at +8)
-[EXTaskExecutionRequest maybeEvaluate]
-[EXTaskExecutionRequest task:didFinishWithResult:]
-[EXTaskService notifyTaskWithName:forAppId:didFinishWithResponse:]
-[EXTaskManager notifyTaskFinished:withResponse:resolve:reject:]
-[EXExportedModule callExportedMethod:withArguments:resolver:rejecter:]
__79-[EXNativeModulesProxy callMethod:...]_block_invoke
_dispatch_lane_serial_drain / _dispatch_workloop_worker_thread
```

## Main thread at time of crash

```
BackgroundTaskAppDelegateSubscriber.reschedule()   <- running simultaneously
Logger.debug(_:) / Logger.log(type:_:)
```

## Analysis

`0x8` = null pointer dereference at offset +8 (first ivar past the ObjC `isa` pointer) — an object has been deallocated while still in use.

The race is between:
[...]

PR #47594, opened 2026-07-07
[ios][task-manager] Make task request callback fire exactly once

### LINKED ISSUE #47572: [iOS] expo-task-manager: EXTaskExecutionRequest invokes _callback before clearing it — re-entrant evaluation double-fires a freed callback (EXC_BAD_ACCESS)
### Summary

In `EXTaskExecutionRequest._maybeExecuteCallback` ([`EXTaskExecutionRequest.m` on `main`](https://github.com/expo/expo/blob/main/packages/expo-task-manager/ios/EXTaskManager/EXTaskExecutionRequest.m)), the callback is invoked **before** the guard is cleared:

```objc
- (void)_maybeExecuteCallback
{
  if (_callback) {
    // Make a strong pointer to self before executing a callback as the request may be deallocated there,
    // due to this fact `_callback = nil;` was crashing on older versions of iOS (below 12.0).
    __strong EXTaskExecutionRequest *strongSelf = self;

    _callback(_results);
    _callback = nil;
    _tasks = nil;
    _results = nil;
    strongSelf = nil;
  }
}
```

Invoking `_callback` runs `EXTaskService`'s per-request completion block, which unregisters the request and breaks the retain cycle keeping that block (and, transitively, the `EXTaskService` machinery for this request) alive. Because `_callback` is still non-nil at that point, a re-entrant or duplicated evaluation of the request — e.g. `task:didFinishWithResponse:` arriving twice, or `taskDidFinish` racing a foreground transition that also drains pending requests — passes the `if (_callback)` guard a second time and invokes a callback whose captured state has already been freed, crashing with `EXC_BAD_ACCESS` / `SIGSEGV`.

The `__strong strongSelf` dance in the current code protects `self`, but not the callback's captured environment, and it does nothing to prevent the second entry.

This matches previously reported crashes that were closed for lacking a repro, e.g.:
- #44709 — `expo-background-task: SIGSEGV in EXTaskExecutionRequest._maybeExecuteCallback when task completes during foreground transition`
- #47316 — `EXC_BAD_ACCESS in EXTaskService _runTasksSupportingLaunchReason on background relaunch (long-lived startLocationUpdatesAsync task)`

### Reproducible example

There is no deterministic MRE — this is a use-after-free race (a background task finishing while the app transitions to foreground makes it most likely; we observed it in a production app using `startLocationUpdatesAsync`). Filing anyway because the defect and the fix are fully visible at the source level: the callback is a one-shot completion, but nothing enforces one-shot semantics.

### Suggested fix

Capture the callback and results locally, clear all state, then invoke — making the callback strictly fire-once regardless of re-entrancy:

```diff
     __strong EXTaskExecutionRequest *strongSelf = self;
 
-    _callback(_results);
+    void (^callback)(NSArray *) = _callback;
+    NSArray *results = _results;
     _callback = nil;
     _tasks = nil;
     _results = nil;
+
+    callback(results);
+
     strongSelf = nil;
```

We ship exactly this reordering as a local patch of `expo-task-manager@57.0.2` and would love to drop it. Happy to open a PR if useful.

### Environment

- `expo`: ~57.0.1 (SDK 57)
- `expo-task-manager`: 57.0.2 (code unchanged on `main`)
- `react-native`: 0.86.0
- Standalone production build (CNG/prebuild), not Expo Go
- Platform: iOS

tsapeta:
## Why

Fixes #47572. In production apps using background tasks (e.g. `startLocationUpdatesAsync`), we've seen intermittent `EXC_BAD_ACCESS` / `SIGSEGV` crashes in `EXTaskExecutionRequest._maybeExecuteCallback`. This also plausibly explains the earlier crash reports #44709 and #47316, which were closed for lacking a deterministic repro.

`_maybeExecuteCallback` invoked the completion callback *before* clearing `_callback`:

```objc
_callback(_results);
_callback = nil;
_tasks = nil;
_results = nil;
```

The `if (_callback)` guard is meant to give the callback one-shot semantics, but because the state was cleared only *after* the call returned, a re-entrant or duplicated evaluation of the same request that reached `_maybeExecuteCallback` while `_callback(...)` was still on the stack passed the guard a second time. The callback is `EXTaskService`'s per-request completion block, which unregisters the request and breaks the retain cycle keeping its captured state alive, so the second invocation fired against already-freed memory. The re-entry is reachable when a background task finishes during a foreground transition (the completion block's work drains more task-finish notifications for a task still in the same request's set).

## How

Capture the callback and results into locals, clear all request state (`_callback`, `_tasks`, `_results`), then invoke. Now the `if (_callback)` guard genuinely fires once: a re-entrant `_maybeExecuteCallback` sees `_callback == nil` and no-ops instead of firing against freed memory. The `__strong strongSelf` reference is kept, since the callback removes the request from `EXTaskService._requests` and can drop the last strong reference to `self` mid-call.

This matches the fix reported in #47572, which the reporter has been shipping as a local patch of `expo-task-manager`.

## Test Plan

There is no automated test: the crash is a use-after-free race in Objective-C completion plumbing with no JSI surface, and `EXTaskExecutionRequest` has no native unit-test target. The fix is verified by source inspection (the reorder makes the guard fire-once) and by the reporter running the identical reorder in a production app that previously crashed. `et check-packages expo-task-manager` passes (build, lint, format).

### DIFF
--- a/packages/expo-task-manager/CHANGELOG.md
+++ b/packages/expo-task-manager/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### 🐛 Bug fixes
 
+- [iOS] Fix a crash when a task execution request is evaluated re-entrantly, by making its completion callback fire exactly once. ([#47594](https://github.com/expo/expo/pull/47594) by [@tsapeta](https://github.com/tsapeta))
+
 ### 💡 Others
 
 ## 56.0.15 — 2026-05-26
--- a/packages/expo-task-manager/ios/EXTaskManager/EXTaskExecutionRequest.m
+++ b/packages/expo-task-manager/ios/EXTaskManager/EXTaskExecutionRequest.m
@@ -56,10 +56,17 @@
     // due to this fact `_callback = nil;` was crashing on older versions of iOS (below 12.0).
     __strong EXTaskExecutionRequest *strongSelf = self;
 
-    _callback(_results);
+    // Capture the callback and results, then clear the request state before invoking. The callback is a
+    // one-shot completion that unregisters this request and breaks the retain cycle keeping its captured
+    // state alive, so a re-entrant or duplicated evaluation (e.g. a task finishing during a foreground
+    // transition) must not pass the `_callback` guard again and fire against freed memory.
+    void (^callback)(NSArray *) = _callback;
+    NSArray *results = _results;
     _callback = nil;
     _tasks = nil;
     _results = nil;
+
+    callback(results);
     strongSelf = nil;
   }
 }


alanjhughes (approved):

nickcernera:
We're hitting this on SDK 55 (`expo-task-manager@55.0.15`, `react-native` 0.83.6, iOS 26.5.2, App Store build) — same stack as #47316, `EXC_BAD_ACCESS` / `KERN_INVALID_ADDRESS at 0x8` in the `_runTasksSupportingLaunchReason` completion block via `_maybeExecuteCallback` → `notifyTaskWithName:forAppId:didFinishWithResponse:`.

The `sdk-55` branch still has the pre-fix ordering, and the changelog entry for this PR is still under `Unpublished`, so there's currently no released version carrying the fix.

Is a cherry-pick to `sdk-55` on the table, or should we plan on carrying the patch locally until the next release? Happy to open the backport PR if that would help.

tsapeta:
@nickcernera This fix has been published from the branch dedicated for SDK 57. Here is its changelog: https://github.com/expo/expo/blob/sdk-57/packages/expo-task-manager/CHANGELOG.md
Changelogs on the main branch are usually out of sync because we're publishing packages from the release branches. There are no plans to backport it to SDK 55 so I recommend to upgrade to SDK 57 or use patch-package to patch it locally.

nickcernera:
Thanks @tsapeta!

PR #47596, opened 2026-07-07
[ios][autolinking] Fix ccache build settings for targets not integrated with CocoaPods

AbbanMustafa:
# Why

When React Native's ccache integration is active (`USE_CCACHE=1` or `:ccache_enabled => true`), iOS builds fail for projects with custom app extension targets (share extensions, widgets): for ex https://exponent-internal.slack.com/archives/C03KMUK9J10/p1783418122028839

```
error: unable to spawn process '/../../node_modules/react-native/scripts/xcode/ccache-clang.sh'
(No such file or directory) (in target 'expo-sharing-extension' from project 'workvivo')
```

`react_native_post_install` sets `CC`/`LD`/`CXX`/`LDPLUSPLUS` at the **project level** of the user's Xcode project, pointing at `$(REACT_NATIVE_PATH)/scripts/xcode/ccache-clang.sh`, and sets `REACT_NATIVE_PATH` itself (also project level) to `${PODS_ROOT}/../<react-native>`. Project-level settings are inherited by **every** target in the project, but `PODS_ROOT` is only defined by the CocoaPods-generated xcconfigs. For targets not integrated with CocoaPods the variable expands to nothing, the wrapper path resolves to `/../../node_modules/...`, and the link step fails.

# How

In our `perform_post_install_actions` override (runs after the user's `post_install`, i.e. after `react_native_post_install`), when the ccache wrapper settings are present on a user project, re-anchor the project-level `REACT_NATIVE_PATH` from `${PODS_ROOT}` to `$(SRCROOT)/<pods-dir-relative-to-project>`. `SRCROOT` is an Xcode built-in defined for every target, so the setting now resolves everywhere:

- CocoaPods-integrated targets resolve to the byte-identical path as before.
- Non-integrated targets go from a broken linker invocation to working ccache.
- No-op when ccache is not enabled; idempotent across repeated `pod install`.

The underlying bug is upstream in `react_native_pods.rb` (a project-wide setting anchored to a pods-only, per-target variable) — will file there separately; this ships the fix to SDK users via a patch release without waiting on an RN version bump.

# Test Plan

- Harness against real `Xcodeproj` + CocoaPods 1.16.2: a project with an app target + share-extension target and RN's exact ccache-style project-level settings. Verified `REACT_NATIVE_PATH` is rewritten to `$(SRCROOT)/Pods/../../node_modules/react-native` with `CC`/`LD` untouched; untouched when ccache settings are absent; idempotent on a second run.
- `ruby -c` on `installer.rb`.


AbbanMustafa:
Validated end-to-end on EAS Build with a repro app (SDK 54, managed) containing a WidgetKit extension target added straight to the pbxproj, i.e. not integrated with CocoaPods — same shape as the customer's share extension. Both builds use the same profile with `EAS_USE_CACHE=1` (iOS simulator); the only difference is this PR's `installer.rb` change applied via patch-package:

- **Without the fix** — fails with the exact error from the report, at the extension's link step: https://expo.dev/accounts/mustafaexpo/projects/worldcup-nyc-planner/builds/1182689f-a8c8-40ba-ac50-83b9f1efe226
  ```
  ❌ error: unable to spawn process '/../../node_modules/react-native/scripts/xcode/ccache-clang.sh' (No such file or directory) (in target 'HelloWidget' from project 'WorldCupNYC')
  ```
- **With the fix** — build succeeds, and the pod install logs show the new code path firing: https://expo.dev/accounts/mustafaexpo/projects/worldcup-nyc-planner/builds/d3a5f624-e359-4374-a573-0a5c3b5db27c
  ```
  [Expo] Re-anchored REACT_NATIVE_PATH to $(SRCROOT) in WorldCupNYC.xcodeproj so ccache build settings resolve for all targets
  ```

Also verified locally with `xcodebuild -showBuildSettings` on the extension target: `LD` resolves to `/../../node_modules/react-native/scripts/xcode/ccache-clang.sh` (nonexistent) before the fix, and to `<project>/ios/Pods/../../node_modules/react-native/scripts/xcode/ccache-clang.sh` (exists) after, while the CocoaPods-integrated app target resolves identically before and after.

The original report was on SDK 57 / RN 0.86 and this repro is SDK 54 / RN 0.81, so the fix holds across both anchoring generations of `react_native_pods.rb`.

sjchmiela (approved):
not an expert here, but overall it looks good to me.

this feels like an upstreamable thing too?

packages/expo-modules-autolinking/scripts/ios/cocoapods/installer.rb
```diff
@@ -223,6 +227,48 @@ def reconcile_resource_bundle_deployment_targets
...
+    def fix_react_native_path_for_non_cocoapods_targets
+      user_projects = self.aggregate_targets.map { |t| t.user_project }.compact.uniq { |p| p.path }
+      user_projects.each do |project|
+        pods_root_from_srcroot = File.join('$(SRCROOT)', self.sandbox.root.relative_path_from(project.path.dirname).to_s)
+        changed = false
+        project.build_configurations.each do |config|
+          ccache_in_use = ['CC', 'LD', 'CXX', 'LDPLUSPLUS'].any? do |key|
+            config.build_settings[key].is_a?(String) && config.build_settings[key].include?('ccache-clang')
+          end
+          next unless ccache_in_use
```
chrfalch, line 254: Make the check for cc-cache a bit narrower (like in [React Native here](https://github.com/facebook/react-native/blob/9160010d1f35266ddc18004f7fb945958aa16be6/packages/react-native/scripts/cocoapods/utils.rb#L104-L130))

PS.: Move the declaration of `react_native_ccache_settings` outside of the loop - I just added it here to show.

```suggestion
          react_native_ccache_settings = {
            'CC' => '$(REACT_NATIVE_PATH)/scripts/xcode/ccache-clang.sh',
            'LD' => '$(REACT_NATIVE_PATH)/scripts/xcode/ccache-clang.sh',
            'CXX' => '$(REACT_NATIVE_PATH)/scripts/xcode/ccache-clang++.sh',
            'LDPLUSPLUS' => '$(REACT_NATIVE_PATH)/scripts/xcode/ccache-clang++.sh',
          }

          ccache_in_use = react_native_ccache_settings.any? do |key, expected|
            config.build_settings[key] == expected
          end
          next unless ccache_in_use
```

chrfalch (approved):
A small comment - I don't think it is super important to fix, it is up to you!

Kudo:
christian is expert for the cocoapods. i'm removing myself from review

AbbanMustafa:
Revalidated again end-to-end with a fresh repro

  - Without the fix, fails at the widget link step with the original error:
    https://expo.dev/accounts/mustafaexpo/projects/plantpal/builds/89b9e850-7f20-4432-b437-9eaa1366369c

```
    unable to spawn process '/../../node_modules/react-native/scripts/xcode/ccache-clang.sh'
    (No such file or directory) (in target 'PlantPalWidget' from project 'PlantPal')
```

  - With the current PR fix, succeeds and produces the simulator artifact:
    https://expo.dev/accounts/mustafaexpo/projects/plantpal/builds/7032709f-50f7-4c9b-9e9d-19c62f0cb440

    The pod-install logs confirm the new path ran:

```
    [Expo] Re-anchored REACT_NATIVE_PATH to $(SRCROOT) in PlantPal.xcodeproj so ccache build settings resolve for all targets
```

  The two build archives used the same app commit, profile, environment, and native extension setup; they differed only by the expo-modules-autolinking patch.

PR #47601, opened 2026-07-07
[bare-expo] Run e2e flows in one Maestro invocation and retry only failures

tsapeta:
# Why

The Test Suite e2e jobs are slow and flaky-retry-expensive. Every `maestro test` invocation pays 60-110s of JVM + XCTest driver startup, plus an app reinstall and relaunch, and the runner invoked Maestro once per flow and once per retry attempt. In a typical flaky run (e.g. [this job](https://github.com/expo/expo/actions/runs/28882152638/job/85677328862)) that's 8 invocations, roughly 10 of 23 minutes spent on pure setup. Retries also reran the whole flow set from a cold start.

# How

- All custom flows now run in a single `maestro test` invocation (flows execute in argument order and a failed flow doesn't stop the following ones).
- Simulator boot, app install, and the inspector-dylib launch happen once per job instead of once per flow. On Android the APK is installed once.
- Failed flows are read from the JUnit report (`--format junit`) via a new `maestro-junit-report.ts` parser, so each retry attempt reruns only the flows that actually failed. When Maestro fails without reporting any failed flow (e.g. it can't reach the device), the error propagates instead of burning retries.
- Retry attempts relaunch the app (iOS: `simctl launch` with the inspector dylib re-injected, Android: `am force-stop`) to reset any state left behind by failed flows, and pass `--no-reinstall-driver` to skip the driver reinstall.
- The generated native-modules suite keeps its retry count but no longer reinstalls or relaunches anything between attempts.
- The CI `bun test` step now also covers `scripts/lib`, so the parser tests run in the `android-build` job.

Expected effect on the linked run: ~23.5 min → ~12-13 min with identical failures; a fully green run saves ~6-7 min.

# Test Plan

- `bun test e2e/image-comparison scripts/lib` passes (22 tests, including the new parser tests).
- Verified the orchestration live against a local simulator with probe flows:
  - green path: one invocation, no retries;
  - flaky path: attempt 1 runs all flows, attempts 2-3 rerun only the failed flow, and the runner throws after 3 attempts listing it;
  - infra path (bogus device id): the Maestro error propagates immediately instead of retrying.
- Verified on Maestro 2.3.0 that multiple flow files per invocation, CLI-order execution, continue-on-failure, JUnit output, and `--no-reinstall-driver` all behave as relied upon (CI uses 2.4.0/latest).
- This PR touches `test-suite.yml` and `apps/bare-expo/**`, so the full Test Suite e2e workflow runs on both platforms here.

### DIFF
--- a/apps/bare-expo/scripts/lib/e2e-common.ts
+++ b/apps/bare-expo/scripts/lib/e2e-common.ts
@@ -244,18 +244,30 @@ export interface RunMaestroOptions {
    * invocation in a session needs it; skipping the reinstall saves about a minute per invocation.
    */
   reinstallDriver: boolean;
+  /** How long a wedged maestro may outlive its own JUnit report before it gets killed. */
+  reportGracePeriodMs?: number;
+  /** Hard ceiling on the whole invocation, for runs that never produce a report. */
+  invocationTimeoutMs?: number;
 }
 
 /**
  * Runs the given flows in a single maestro invocation (flows execute in the given order and
  * a failed flow doesn't stop the following ones) and returns the relative paths of failed flows.
  * Throws if maestro fails without producing flow results, e.g. when it can't reach the device.
+ *
+ * Maestro writes the JUnit report only after all flows have finished, so a process that
+ * outlives the report is just wedged on exit (Maestro 2.4.0 can crash its main thread while
+ * finalizing logs and then hang forever on a non-daemon thread). A watchdog kills the process
+ * once the report has been around for the grace period and the results are read as usual;
+ * the invocation timeout backstops runs that never write a report at all.
  */
 export async function runMaestroAsync({
   deviceArgs,
   flowRelativePaths,
   e2eDir,
   reinstallDriver,
+  reportGracePeriodMs = 60_000,
+  invocationTimeoutMs = 20 * 60_000,
 }: RunMaestroOptions): Promise<string[]> {
   const reportPath = path.join(
     await fs.mkdtemp(path.join(os.tmpdir(), 'maestro-report-')),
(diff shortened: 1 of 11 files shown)


apps/bare-expo/scripts/start-android-e2e-test.ts
```diff
@@ -53,25 +54,47 @@ const __dirname = dirname(__filename);
...
           deviceId,
-          appBinaryPath,
           adbPath,
           e2eDir
         );
+        if (failedFlows.length > 0) {
+          annotate(
+            retryNumber + 1 >= NUM_OF_RETRIES ? 'error' : 'warning',
+            'Native modules test suite failed',
+            `attempt  of  failed`
```
Kudo: [💡 suggestion] The `${retryNumber + 1}` and `${NUM_OF_RETRIES}` interpolations got dropped, so this annotation always reads a literal `attempt  of  failed`. The iOS script has the intended version.

```suggestion
            `attempt ${retryNumber + 1} of ${NUM_OF_RETRIES} failed`
```

apps/bare-expo/scripts/lib/e2e-common.ts
```diff
@@ -52,8 +56,15 @@ jsEngine: graaljs
   ];
   if (confirmFirstRunPromptIOS) {
     contents.push(`\
-# Run once to approve the first time deeplinking prompt on iOS (clearState above resets this)
+# Run once to approve the first time deeplinking prompt on iOS (clearState above resets this).
+# The prompt can take a while to show up while the app cold-starts after clearState, so wait
+# for it instead of racing it with an immediate tap; otherwise the approval silently doesn't
+# happen and the first test case's deep link opens the prompt instead of the test suite.
```
Kudo: nit: comment too long. extendedWaitUntil timeout 20000 explains well

apps/bare-expo/scripts/lib/maestro-junit-report.ts
```diff
@@ -0,0 +1,44 @@
...
+        `Flows '${existingFlow}' and '${flowRelativePath}' would both report as '${flowName}'. ` +
+          `Rename one of them so that Maestro results can be mapped back to flow files.`
+      );
+    }
+    flowsByName.set(flowName, flowRelativePath);
+  }
+
+  const failedFlows: string[] = [];
+  for (const [, testcase] of reportContents.matchAll(/<testcase\b([^>]*)>/g)) {
+    const name = testcase.match(/\bname="([^"]*)"/)?.[1];
```
Kudo, line 29: maybe we can use bun builtin xml parser later when it's available https://x.com/bunjavascript/status/2042868714585141520

Kudo (approved):
great improvement and with some cool stuffs:
- using pipe for the inspector
- pass junit failed report

PR #47603, opened 2026-07-07
[expo-modules-jsi] Add clear caches script

alanjhughes:
# Why
Follow up of #47586. Instead of `et prepare-branch` hardcoding which directories count as expo-modules-jsi build caches, the package should own that with a dedicated script.

# How
- Added `apple/scripts/clear-caches.sh` to `expo-modules-jsi`. It runs `git clean -fdX` scoped to the `apple` directory, so it removes everything gitignored there (`.build`, `.DerivedData`, `.generated`, `.swiftpm`, `.xcframework-slices`, `.test-frameworks`, `Products`) and stays in sync with `.gitignore` automatically. Capital `X` is deliberate, it only touches ignored files so untracked work in progress survives.
- `et prepare-branch` now calls the script in its expo-modules-jsi step instead of removing a hardcoded list of four directories.

# Test Plan
- Ran the script against the real package with all cache directories present plus an untracked decoy file in `Sources`: the ignored directories were removed, the decoy survived, and a second run exited cleanly with nothing to delete.
- `pnpm run clean` from the package directory invokes the script correctly.


tsapeta (commented):
Could we make something that would be reusable in 👇? It could be a function in `xcframework-helpers.sh`.
https://github.com/expo/expo/blob/6863941cf7f848ce6af55f5e1394c98631823bca/packages/expo-modules-jsi/apple/scripts/build-xcframework.sh#L402-L409

tsapeta (approved):

PR #47604, opened 2026-07-07
[bare-expo][video] Make the expo-video e2e tests deterministic

tsapeta:
# Why

The expo-video flows are the least reliable part of the Test Suite e2e workflow. Most of it traces back to one root cause, streaming remote videos in CI, plus one screenshot-geometry race:

- `player-output-test` was the flakiest flow. In a single recent run ([example](https://github.com/expo/expo/actions/runs/28882152638/job/85677328862)) it failed three attempts in three different ways: "Players ready" timing out, a 37.83% screenshot diff, and the `viewshot-ready` marker never appearing.
- `playback-test` streamed a remote 720p Big Buck Bunny and asserted playback metadata against it, so its duration and reliability degraded with the runner's network and load (54s on a healthy runner, 3m30s on an overloaded one).
- `surface-type-test.android` streamed the same remote video with four players at once under a 10-second readiness timeout, and failed both retry attempts in recent runs.
- `picture-in-picture-test.android` screenshotted a container spanning the whole screen, whose bounds race the system bar insets while the PiP-exit transition settles (CI captured 525x1244 and 525x1274 about evenly across attempts, so no fixed baseline could pass reliably). A failed attempt also left the app as a PiP window, poisoning the three flows that run after it in the same Maestro invocation.

After this PR, no expo-video flow touches the network and every viewshot targets a fixed-size view.

# How

## player-output-test

The flow now drives `VideoPlayerOutputE2EScreen`, which replaces `VideoChangePlayerOutputScreen` (the old screen had accumulated near-identical e2e scaffolding, so only one screen remains, labeled "(e2e)" in the list). It exercises the same behavior but is built for determinism:

- Bundled videos (`ace.mp4` / `star.mp4`, distinct content and aspect ratios so misplaced players stay visually detectable), seeked to a fixed frame once loaded.
- Placement, per-view readiness, and a move counter live in one state object updated functionally, so Maestro's rapid `repeat:` taps can't interleave and drop a move.
- The ready marker's testID carries the move counter (`viewshot-ready-<N>`), so each checkpoint waits for the exact post-tap state instead of matching a stale marker. A view is only gated on `onFirstFrameRender` when its player actually changes, since re-assigning the same player never repaints.

## playback-test

`VideoEventsScreen` (now labeled "(e2e)" in the list) plays the bundled `star.mp4`, with `ace.mp4` as the replacement source so the source toggle stays offline too. The seek target moves from 30s to 20s to fit the 28-second video and the metadata assertions match the new source.

## fullscreen-test and surface-type-test

Both screens switch to the bundled `star.mp4`. Their existing e2e pause/setup buttons already seek within its duration, so the flows needed no changes.

## picture-in-picture-test

- The viewshot wraps only the fixed-size video view, making the capture geometrically indifferent to the inset race.
- The flow waits for the restore animation to settle before capturing, mirroring playback-test.
- The flow stops the app when it completes, even on failure (`onFlowComplete`), so a failed attempt can't poison the flows after it. That hook exposed that both Android-only flows declared the iOS bundle id as their `appId`, making `stopApp` silently target a nonexistent package; they now use the Android package name.

## Baselines

Each capture change regenerates its baselines in a dedicated commit, using the viewshots CI uploads to its workflow artifacts. Captures from a local device don't work as baselines because its resolution and density differ from the CI devices; the comparison server's dimension mismatch error now explains that directly.

# Test Plan

- `tsc --noEmit` and `oxlint` pass in `apps/native-component-list`.
- All flows verified locally end to end on the iOS simulator and Android emulator (playback on both platforms; picture-in-picture and surface-type on Android, including that the app is stopped after the PiP flow even when it fails).
- Test Suite e2e on this PR is green on both platforms, with every flow passing on its first attempt.


Kudo (approved):
that's ios build error which seems to be unrelated. may double check before merge

PR #47607, opened 2026-07-07
[package-manager] Add Deno package manager support

yyq1025:
# Why

Deno is now a viable package manager for Expo projects: `nodeModulesDir: "auto"` produces a real `node_modules` directory that works with autolinking, `deno add` defaults to the npm registry (Deno ≥2.6) and writes dependencies to `package.json` when it is the nearest manifest, and Deno can migrate an existing `pnpm-lock.yaml` into `deno.lock` (denoland/deno#35346). There is also a long-running tracking issue on the Deno side: denoland/deno#22970.

This is a frequently requested integration ([Deno on the feature request board](https://expo.canny.io/feature-requests/p/deno), and Brent mentioned on Reddit that PRs improving Deno support would be welcome). `@expo/package-manager` is the natural place to start: `npx expo install` picks it up through `createForProject`, and EAS CLI consumes `resolvePackageManager` from this package — expo/eas-cli#3951 adds the EAS Build side and currently has to hardcode the `deno.lock` filename because this package does not export it yet.

# How

Added `DenoPackageManager`, mirroring the existing `BunPackageManager`:

- `add` / `add --dev` / `remove` pass package names through unchanged — since Deno 2.6, bare names default to the npm registry, and npm dependencies are written to `package.json` when it is the closest manifest (which is the standard Expo project layout).
- `addGlobalAsync` uses `deno install --global` and prefixes bare names with `npm:`, because global installs take module specifiers rather than package names. `removeGlobalAsync` uses `deno uninstall --global`.
- `workspaceRoot()` resolves through `resolve-workspace-root`, same as the other managers.

Detection: `DENO_LOCK_FILE` (`deno.lock`) is exported and wired into `resolvePackageManager` / `createForProject`. Deno is resolved first in `RESOLUTION_ORDER`: `deno.lock` is only ever written by Deno itself, and projects migrated to Deno may retain the lockfile of their previous package manager (Deno can even seed `deno.lock` from `pnpm-lock.yaml`), so its presence is the strongest signal. Projects without a `deno.lock` are unaffected.

Possible follow-ups (kept out of this PR to keep it reviewable): a `--deno` flag for `npx expo install` in `@expo/cli`, and consuming the new export from EAS CLI.

# Test Plan

- Added a `DenoPackageManager` unit test suite mirroring the Bun one, including the `npm:` prefixing behavior for global installs.
- Added `nodeManagers` cases: `{ deno: true }` option, `deno.lock` detection (project + monorepo workspace), and lockfile coexistence (`deno.lock` + `pnpm-lock.yaml` resolves to Deno, while explicitly preferring `pnpm` still resolves pnpm).
- `pnpm run test` (241 passed), `pnpm run typecheck`, `pnpm run lint`, `pnpm run format --check`, and `pnpm run build` all pass in `packages/@expo/package-manager`.
- Real-world validation: an Expo SDK 57 monorepo (app + native modules) fully managed by Deno — `deno install`, `npx expo install`, and `npx expo run:ios` work with this resolution logic.

# Checklist


TheJanitor337:
If it helps anyone else, I've been patching this in for two Expo projects, for a couple of weeks now, without issue. [Expo patch gist](https://gist.github.com/TheJanitor337/1be33312b5870c3b6c62d153cec1a7d9)

PR #47611, opened 2026-07-08
[expo-router][android] Collapse tab content bottom inset when native tab bar is hidden

debitan:
## Why

On Android, each native-tab screen's content is wrapped in a react-native-screens `SafeAreaView` with `edges={{ bottom: true }}`. That bottom inset accounts for the tab bar. When the tab bar is hidden via `NativeTabs`' `hidden` prop (`Tabs.Host` `tabBarHidden`), the bar disappears but the content keeps the inset, leaving an empty band the height of the tab bar below the content. iOS reclaims the space; Android does not.

## How

`NativeTabsView` already receives `hidden`; thread it to each `Screen` as `tabBarHidden` and drive the `SafeAreaView`'s bottom edge with it (`edges={{ bottom: !tabBarHidden }}`). The `SafeAreaView` stays mounted and only its `edges` prop changes, so there is no remount of the tab content. When the bar is visible (`hidden` falsy) behaviour is unchanged.

## Test plan

App with `NativeTabs` where a tab holds a nested stack that hides the tab bar on a screen (`hidden` toggled on focus).
- Before: a blank strip (~ tab bar height) below the screen content on Android when the bar is hidden.
- After: content fills the space; toggling the bar back restores the normal inset, with no layout flicker or remount.
- iOS unaffected (Android-only file).


packages/expo-router/CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 
 ### 🐛 Bug fixes
 
+- [android] Collapse the tab content bottom inset when the native tab bar is hidden, so screens that hide the bar (e.g. a nested stack) no longer leave an empty band the height of the tab bar. ([#47611](https://github.com/expo/expo/pull/47611) by [@debitan](https://github.com/debitan))
```
```suggestion
- [android] Disable safe area insets in Native Tabs on Android when tab bar is hidden. ([#47611](https://github.com/expo/expo/pull/47611) by [@debitan](https://github.com/debitan))
```

Ubax (commented):
Can you please add tests for the behavior?

chrfalch:
@debitan - thanks for your contribution! Please approve the suggestion from @Ubax before we move on.

`edges={{ bottom: false }}` drops the whole bottom inset. 

Could you confirm that on a hidden-bar screen the content still sits above the system **navigation bar** on 3-button nav (not just gesture nav), and check behavior with the keyboard open? Want to make sure we're reclaiming the tab-bar band without pushing bottom-anchored content under the nav/gesture bar. 

@Ubax — is a full bottom: false the intended semantics here, or should it still respect the system nav inset when the bar is hidden?

Ubax:
Leaving a note that this is already fixed with react-native-screens 4.26.0, so I won't cherry-pick this to sdk 56 or 57

Ubax (approved):

PR #44766, opened 2026-04-14
refactor(expo-camera): extract barcode scanning into optional companion pod

chrfalch:
# Why

When expo-camera is precompiled, `#if canImport(ZXingObjC)` is resolved at prebuild time — users cannot disable barcode scanning to save binary size (~1-2MB for ZXingObjC). Setting `barcodeScannerEnabled: false` had no effect on precompiled builds.

# How

Move ZXingObjC barcode scanning into a separate ExpoCameraBarcodeScanning provider module. ExpoCamera defines a protocol and discovers the provider at runtime via NSClassFromString — if the companion pod is linked, scanning works; if not, it's gracefully disabled.

The prebuild system gains two generic capabilities:

- Intra-package product dependencies: sibling products in the same spm.config.json can depend on each other's xcframeworks, with transitive external dependencies resolved automatically.
- Companion pod auto-registration: products can declare `autolinkWhen` in spm.config.json to be conditionally included based on Podfile.properties.json values.

# Test plan

Bare-Expo:
✅ Run with ZXingObjC disabled (expo.camera.barcode-scanner-enabled = false) with regular pod install - barcode scanner not loaded 
✅ Run with ZXingObjC enabled (expo.camera.barcode-scanner-enabled = true) with regular pod install - barcode scanner IS found 
✅ Run with ZXingObjC disabled (expo.camera.barcode-scanner-enabled = false) with precompiled modules - barcode scanner not loaded 
✅ Run with ZXingObjC enabled (expo.camera.barcode-scanner-enabled = true) with precompiled modules - barcode scanner IS found

# Checklist

- [x] I added a `changelog.md` entry and rebuilt the package sources according to [this short guide](https://github.com/expo/expo/blob/main/CONTRIBUTING.md#-before-submitting)
- [x] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: updated a module plugin).

PR #44726, opened 2026-04-13
[ios][camera] Fix inconsistent barcode type string returned by ZXing fallback scanner

jensdev:
## Summary

When scanning `code39`, `pdf417`, or `codabar` barcodes on iOS, the `type` field in the `onBarcodeScanned` event was returning the raw `AVMetadataObject.ObjectType` string (e.g. `"org.iso.Code39"`) instead of the `BarcodeType` enum value (e.g. `"code39"`).

This only affected the ZXing fallback scanner path, which handles barcode types not natively supported by AVFoundation. The AVFoundation path was already using `BarcodeType.toBarcodeType(...).rawValue` and returning the correct format.

### Root cause

`BarcodeScannerUtils.zxingFormatToString` was returning `AVMetadataObject.ObjectType.rawValue` directly, while the AVFoundation path was going through `BarcodeType.toBarcodeType()` first.

### Fix

Replaced `zxingFormatToString(_ format:) -> String` with `zxingFormatToBarcodeType(_ format:) -> BarcodeType?`, aligning the ZXing path with the AVFoundation path. As a side effect, the redundant `#available(iOS 15.4, *)` guard for `codabar` inside the format mapper was removed — the codabar ZXing reader is only registered on iOS 15.4+ in BarcodeScanner.swift`, so the case can never be reached on earlier versions.

## Test Plan

- [ ] Scan a **Code39** barcode using `<CameraView onBarcodeScanned={...} />` on a
      physical iOS device and verify `event.data` is the expected string and
      `event.type === "code39"` (previously returned `"org.iso.Code39"`)
- [ ] Scan a **PDF417** barcode and verify `event.type === "pdf417"`
      (previously returned `"org.iso.PDF417"`)
- [ ] Scan a **Codabar** barcode on iOS 15.4+ and verify `event.type === "codabar"`
- [ ] Verify other barcode types (QR, EAN-13, Code128, etc.) still work correctly —
      those go through the AVFoundation path and are unaffected by this change

PR #47613, opened 2026-07-08
[ios][camera] Fix ZXing barcode fallback emitting raw AVFoundation ty…

jensdev:
On iOS, expo-camera has two barcode scan paths, and they disagreed on the type value for three formats:

- The native AVFoundation path (avMetadataCodeObjectToDictionary) emits the short expo BarcodeType strings — pdf417, code39, codabar.
- The ZXing fallback provider (ExpoCameraZXingProvider) emitted the raw AVMetadataObject.ObjectType values — org.iso.PDF417, org.iso.Code39, org.ansi.Codabar.

Because either path can fire for the same barcode, consumers treating event.type as an expo BarcodeType saw flaky, path-dependent values for these formats.

This is a regression. The normalization was originally added in #44726, but was reverted by #44766, which extracted the ZXing logic into the separate ExpoCameraBarcodeScanning companion pod. That pod imports ExpoCamera as an external module and can no longer see the internal BarcodeType / toBarcodeType(type:) mapping, so it fell back to emitting raw values.

How

- Add a small expoType(for:) helper in ExpoCameraZXingProvider that maps the three raw ObjectType keys to their short expo strings, applied on both return paths (the upright decode and the rotated retry).
- Any type the provider doesn't handle is passed through unchanged.
- Document the ExpoBarcodeScannerProvider protocol contract: type must be the short expo BarcodeType string, matching the native path.

This keeps the fix self-contained in the companion pod, touches no shared/public API, and leaves the native path and BarcodeType untouched.

Test Plan

Native iOS build/tests can't run in my environment (Linux; AVFoundation/ZXingObjC are Apple-only). Reviewer verification on macOS:

- pod install in an app that links ExpoCameraBarcodeScanning, then build expo-camera.
- Scan a PDF417, a Code 39, and a Codabar barcode via the ZXing fallback and confirm event.type is pdf417 / code39 / codabar (not org.iso.PDF417 etc.), matching the native AVFoundation path.

Checklist

- [x] Added a CHANGELOG.md entry.
- [x] Change is limited to iOS native source; no config plugin / prebuild impact.
- [x] Conforms to the documentation style guide.

References: restores #44726, reverted by #44766.

### DIFF
--- a/packages/expo-camera/CHANGELOG.md
+++ b/packages/expo-camera/CHANGELOG.md
@@ -16,7 +16,7 @@
 - Replace the deprecated `videoOrientation` API with `AVCaptureDevice.RotationCoordinator` for the iOS camera preview. ([#47172](https://github.com/expo/expo/pull/47172) by [@alanjhughes](https://github.com/alanjhughes))
 - Fix dark frames and the preview rotating into place on iOS launch by fully configuring the camera session before it starts running. ([#47173](https://github.com/expo/expo/pull/47173) by [@alanjhughes](https://github.com/alanjhughes))
 - Default the iOS camera `pictureSize` to `photo` instead of `high`. ([#47173](https://github.com/expo/expo/pull/47173) by [@alanjhughes](https://github.com/alanjhughes))
-- [iOS] Fix the ZXing barcode fallback scanner returning raw AVFoundation type strings (e.g. `org.iso.PDF417`) instead of short expo `BarcodeType` values (e.g. `pdf417`) for `pdf417`, `code39`, and `codabar`, restoring the fix from [#44726](https://github.com/expo/expo/pull/44726) that was reverted by the `ExpoCameraBarcodeScanning` pod extraction in [#44766](https://github.com/expo/expo/pull/44766). ([#44726](https://github.com/expo/expo/pull/44726) by [@jensdev](https://github.com/jensdev))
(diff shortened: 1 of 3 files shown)


packages/expo-camera/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 
 ### 🐛 Bug fixes
 
+- [iOS] Fix the ZXing barcode fallback scanner returning raw AVFoundation type strings (e.g. `org.iso.PDF417`) instead of short expo `BarcodeType` values (e.g. `pdf417`) for `pdf417`, `code39`, and `codabar`, restoring the fix from [#44726](https://github.com/expo/expo/pull/44726) that was reverted by the `ExpoCameraBarcodeScanning` pod extraction in [#44766](https://github.com/expo/expo/pull/44766). ([#44726](https://github.com/expo/expo/pull/44726) by [@jensdev](https://github.com/jensdev))
```
alanjhughes: Please add this at the bottom of the list

alanjhughes (approved):

packages/expo-camera/CHANGELOG.md
jensdev: Done

alanjhughes:
Thanks for your contribution @jensdev!

PR #47614, opened 2026-07-08
[bare-expo] Pre-approve the iOS deep link prompt instead of tapping it

tsapeta:
# Why

The iOS e2e setup approved the "Open in BareExpo?" deep link confirmation prompt by tapping it: once in a dedicated `confirm-app-open` maestro invocation before the custom flows, and again inside the generated native-modules flow after `clearState`. The tap raced the app's cold start: when the prompt rendered after the optional tap had already given up, the approval silently didn't happen, the first test case's deep link opened the prompt instead of the test suite, and the whole suite attempt failed on the selection header wait (the recurring `"AppMetrics" ... test_suite_selection_query_text is visible` failures, which regularly burned 2-3 of the suite's retry attempts).

# How

- The runner writes the approval directly into the simulator's LaunchServices scheme-approval store (`com.apple.launchservices.schemeapproval`, key `com.apple.CoreSimulator.CoreSimulatorBridge-->bareexpo`) right after booting the simulator. CoreSimulatorBridge is the requesting process for both `simctl openurl` and maestro's `openLink`, and the store lives in the simulator's global preferences, so the approval survives `clearState` and the prompt never appears at all.
- The `confirm-app-open` flow and all prompt handling in the generated test-suite flow are removed. What remains is a cold-start warm-up: a deep link that launches the app can get dropped before the JS side is ready to handle it, so the generated flow opens the first link once up front and waits for the app UI before the per-test links.

# Test Plan

Verified locally against the same iOS 26 simulator runtime CI uses, after wiping `com.apple.launchservices.schemeapproval`:

- `clearState` → `openLink` shows no prompt with the pre-approval written.
- The full generated-flow shape (`clearState` → warm-up link → wait → per-test link → header → results) passes with the AppMetrics test case reporting "All tests passed!".
- Also reproduced both failure modes this replaces: without pre-approval the cold-start link lands on the prompt, and without the warm-up wait, back-to-back links both get dropped during the cold start (previously the tap block accidentally served as that delay).
- Test Suite e2e runs on this PR.


apps/bare-expo/scripts/start-ios-e2e-test.ts
```diff
@@ -197,6 +216,7 @@ async function testAsync(
 
   try {
     await startSimulatorAsync(deviceId);
+    await preApproveDeepLinkPromptAsync(deviceId);
```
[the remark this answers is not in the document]
tsapeta: The premise doesn't hold up empirically. The CI jobs erase and shut down all simulators before the test script runs (the "Cleanup unused simulator runtimes" step), so in both validation runs of this change the approval was written after booting a freshly-erased simulator, and every `openLink` navigated without the prompt, including the generated test-suite flow passing on its first attempt:

- https://github.com/expo/expo/actions/runs/28933947801 (this PR's variant)
- https://github.com/expo/expo/actions/runs/28930946203 (the same mechanism on a refactored runner)

If LaunchServices only read the store at boot, both runs would have failed all suite retry attempts on the first deep link. Preferences inside the simulator are mediated by `cfprefsd`, which propagates changes to running consumers, so a post-boot write is visible to the next query.

The boot-retry path in `startSimulatorAsync` (which erases the simulator) also isn't affected: the approval is written after `startSimulatorAsync` returns, so it always post-dates any erase.

alanjhughes (approved):

apps/bare-expo/scripts/lib/e2e-common.ts
```diff
@@ -50,12 +50,18 @@ jsEngine: graaljs
...
-  if (confirmFirstRunPromptIOS) {
+  if (warmUpDeepLinkIOS) {
     contents.push(`\
-# Run once to approve the first time deeplinking prompt on iOS (clearState above resets this)
+# A deep link that cold-starts the app after clearState can get dropped before the JS side is
+# ready to handle it, so open the first link once up front and wait for the app UI to come up
+# before the per-test links below, which then hit a running app. Optional because when the
+# cold-start link does get handled, the app shows the test run screen without this header.
+# (The deep link confirmation prompt doesn't show up at all: the runner pre-approves the
+# scheme before the flows.)
```
Kudo: - the comment is too long.
- if js side may drop the deeplinking from cold-start, if feels like a bug. did you test with `warmUpDeepLinkIOS=false`, it would actually fail?

apps/bare-expo/scripts/start-ios-e2e-test.ts
```diff
@@ -186,6 +186,25 @@ async function startSimulatorAsync(deviceId: string, timeout: number = 180_000)
   }, 3);
 }
 
+async function preApproveDeepLinkPromptAsync(deviceId: string): Promise<void> {
+  // Approve the bareexpo:// scheme up front so the "Open in BareExpo?" confirmation prompt
+  // never shows up. The approval store lives in the simulator's global preferences (keyed by
+  // the requesting process, CoreSimulatorBridge for both simctl openurl and maestro openLink),
+  // so unlike a tapped approval it also survives clearState.
```
Kudo: ditto: shorter comment

apps/bare-expo/scripts/lib/e2e-common.ts
tsapeta: Yes, I tested that locally as well. On the cold start, the deep link doesn't seem to be restored by the JS side. I think this was pre-existing issue, but we've never seen it because the app was always launched with `bareexpo://apis` link and then redirected to `bareexpo://test-suite/run` (JS already ran).
I think it should be a follow-up investigation.

tsapeta: I confirm, it regressed when we introduced `ExpoAppSceneDelegate` which doesn't pass the initial URL to `factory.startReactNative` through `launchOptions` (it's always nil). I'll follow up on that.

tsapeta: Follow-up: https://github.com/expo/expo/pull/47628

Kudo (approved):

tsapeta:
@codex this was the behavior before #47628, but that PR (the scene-delegate cold-start fix) merged into `main` and this branch is rebased on top of it, so cold-start deep links are no longer dropped: `Linking.getInitialURL()` now returns the URL that launched the app under the scene life cycle. The warm-up link was only ever needed to work around that drop.

I verified the exact flow this PR generates on a Release build containing the fix: `clearState` → a single `openLink` for the first test case cold-starts straight onto the Test Runner and runs the suite to completion, no warm-up. As a safety net, the `test_suite_selection_query_text` step is an `extendedWaitUntil` (30s) inside a 6-attempt suite retry, so a one-off drop would retry rather than hang. The old `confirmFirstRunPromptIOS` block had no wait, just an immediate optional `tapOn`, so it wasn't a reliable warm-up anyway.

PR #47618, opened 2026-07-08
[docs] Add a guide on controlled components

amandeepmittal:
# Why

Fix ENG-21509

# How

- Add `pages/guides/controlled-components.mdx`, a guide to controlled and uncontrolled `TextInput`, differences from controlled inputs on web, formatting and restricting input, cursor behavior, and worklet-based formatting with the universal `TextInput` from `@expo/ui`.
- Add `UpdateLoopDiagram` and `WorkletPathDiagram` under `scenes/guides/controlled-components/` to render the update loop and worklet path as theme-aware components instead of static images.

# Test Plan

See preview: https://pr-47618.expo-docs.pages.dev/guides/controlled-components/

# Checklist

### DIFF
--- a/docs/pages/guides/controlled-components.mdx
+++ b/docs/pages/guides/controlled-components.mdx
@@ -3,6 +3,7 @@ title: Controlled components
 description: Learn how to use controlled components in React Native to control a TextInput, format text as app users type, restrict input, and keep the cursor stable.
 ---
 
+import { Collapsible } from '~/ui/components/Collapsible';
 import { UpdateLoopDiagram } from '~/scenes/guides/controlled-components/UpdateLoopDiagram';
 import { WorkletPathDiagram } from '~/scenes/guides/controlled-components/WorkletPathDiagram';
 
@@ -33,7 +34,7 @@ The following diagram shows how a keystroke travels through the update loop:
 
 <UpdateLoopDiagram />
 
-## Differences from controlled inputs on web
+<Collapsible summary="Differences from controlled inputs on web">
 
 If you have written [controlled inputs for the web](https://react.dev/reference/react-dom/components/input#controlling-an-input-with-a-state-variable), three things change in React Native:
 
@@ -50,6 +51,8 @@ The following example shows the same input on both platforms:
 // React Native
 <TextInput value={text} onChangeText={setText} />
 ```
+
+</Collapsible>
 
 ## Choose between controlled and uncontrolled inputs
 
@@ -82,7 +85,7 @@ The following table compares controlled and uncontrolled inputs:
 | Re-renders per keystroke | Yes                                       | No                                                                                                       |
 | Use when                 | Live validation, formatting, dependent UI | Search boxes, read-on-submit forms                                                                       |
 
-## Format input as users type
+### Format input as users type
 
 To format text such as a phone number, a currency amount, or a date as the user types, transform the string inside `onChangeText` and store the formatted result back in state.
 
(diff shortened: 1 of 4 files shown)


docs/pages/guides/controlled-components.mdx
```diff
@@ -0,0 +1,270 @@
+---
+title: Controlled components
+description: Learn how to use controlled components in React Native to control a TextInput, format text as app users type, restrict input, and keep the cursor stable.
+---
+
+import { Collapsible } from '~/ui/components/Collapsible';
+import { UpdateLoopDiagram } from '~/scenes/guides/controlled-components/UpdateLoopDiagram';
+import { WorkletPathDiagram } from '~/scenes/guides/controlled-components/WorkletPathDiagram';
+
+Every input has a current value, such as the text in an input field, the position of a switch, or the selection in a picker. A **controlled component** keeps the value in your component's state. The input displays whatever the state holds and every change the user makes updates that state. That state is the single source of truth.
```
entiendoNull: I read this sentence as the state has to live in *that* component.

```suggestion
Every input has a current value, such as the text in an input field, the position of a switch, or the selection in a picker. A **controlled component** keeps the value in state you manage. The input displays whatever the state holds and every change the user makes updates that state. That state is the single source of truth.
```

docs/pages/guides/controlled-components.mdx
```diff
@@ -0,0 +1,270 @@
...
+
+The following diagram shows how a keystroke travels through the update loop:
+
+<UpdateLoopDiagram />
+
+<Collapsible summary="Differences from controlled inputs on web">
+
+If you have written [controlled inputs for the web](https://react.dev/reference/react-dom/components/input#controlling-an-input-with-a-state-variable), three things change in React Native:
+
+- The change handler is `onChangeText` and receives the new text directly instead of an event object.
```
entiendoNull: `<TextInput />` has onChange too

```
<TextInput
  onChange={(ev) => setFoo(ev.nativeEvent.text)}
  style={{ borderColor: "red", borderWidth: 1, width: 100, height: 32 }}
  placeholder="Foo"
  value={foo}
/>
```

```suggestion
- The usual change handler is `onChangeText` and receives the new text directly instead of an event object.
```

docs/pages/guides/controlled-components.mdx
```diff
@@ -0,0 +1,270 @@
...
+/>
+```
+
+### Keep the cursor in place while formatting
+
+Before React Native's [New Architecture](/guides/new-architecture/), reformatting the value inside `onChangeText` snapped the cursor to the end of the field, which made edits in the middle of the input difficult.
+
+In React Native and Expo apps that use the New Architecture, inserting text in the middle of the field keeps the cursor right after the inserted text.
+
+If the cursor still ends up in the wrong place (on an older version of React Native, or with a transform that inserts or removes characters before the cursor):
```
entiendoNull: This sounds like a guarantee
> In React Native and Expo apps that use the New Architecture, inserting text in the middle of the field keeps the cursor right after the inserted text.

And the next sentence begins like a contradiction:
> If the cursor still ends up in the wrong place

entiendoNull (approved):

docs/pages/guides/controlled-components.mdx
amandeepmittal: Good catch! The cursor can still move on the New Architecture when `onChangeText` returns different text than what the user typed. @entiendoNull I've updated this to not self-contradict each other here: https://github.com/expo/expo/pull/47618/changes#diff-ee5a5e92b8c521c5b77904c04a54242a5f4d637043dadc0d9c593ed180b3ffddR166

entiendoNull: Looks great!

entiendoNull (approved):

PR #47619, opened 2026-07-08
[ui][ios] Add `container` value to Host's `ignoreSafeArea` prop

intergalacticspacehighway:
# Why 

Issue reported on discord - https://discord.com/channels/695411232856997968/1504711019199467550/1524336586039361636

A `matchContents` `Host` pinned to a screen edge can render its content outside the host's tappable bounds because of the safe area inset, and there was no way to ignore only the device/container safe area while keeping keyboard avoidance.

# How

- User can pass `ignoreSafeAre="container"` to resolve the issue. But we need better defaults, looking into it if we can convey yoga about the applied safe area inset.
- Add `ignoreSafeArea="container"` to `Host`: ignores the device and container safe area (notch, home indicator, status and navigation bars) while still avoiding the keyboard. 
- Also made `ignoreSafeArea` reactive: changing or clearing the prop updates the safe area without an app reload (was applied once on mount).

# Test Plan

- [ ] iOS: a top-pinned `matchContents` `Host` stays interactive with `ignoreSafeArea="container"`.
- [ ] Toggling `ignoreSafeArea` updates the safe area without an app reload.


packages/expo-modules-core/ios/Core/Views/SwiftUI/SwiftUIHostingView.swift
```diff
@@ -260,45 +252,62 @@ extension ExpoSwiftUI {
...
-      // https://developer.apple.com/forums/thread/658432
-      guard let viewClass = object_getClass(view) else { return }
-
-      let suffix = mode == .all ? "_IgnoresSafeArea" : "_IgnoresKeyboard"
+      guard let viewClass = object_getClass(view) else {
+        return
+      }
+      let ignoresContainer = mode == .all || mode == .container
+      let suffix = ignoresContainer ? "_IgnoresSafeArea" : "_IgnoresKeyboard"
       let viewSubclassName = String(cString: class_getName(viewClass)).appending(suffix)
```
Kudo: [🏗️ **design**] `updateProps` now runs this fallback on every prop change (the one-shot `hasSafeAreaBeenConfigured` guard was dropped for reactivity). On iOS < 16.4 the suffix is derived from the view's *current* class — which is already the swapped subclass after the first run — so each update registers a fresh nested `..._IgnoresSafeArea` class, growing the ObjC runtime class table without bound. Reachable on iOS 16.0–16.3 precompiled builds. A short-circuit fixes it:

```suggestion
      let currentClassName = String(cString: class_getName(viewClass))
      // Already subclassed for this mode; skip so repeated prop updates don't keep allocating nested subclasses.
      if currentClassName.hasSuffix(suffix) {
        return
      }
      let viewSubclassName = currentClassName.appending(suffix)
```

Side note: this fallback still overrides `keyboardWillShowWithNotification:` unconditionally, so on iOS < 16.4 `container` also disables keyboard avoidance (i.e. `container` behaves like `all` here), which contradicts the documented "keyboard safe area still applies."

Kudo (approved):

packages/expo-modules-core/ios/Core/Views/SwiftUI/SwiftUIHostingView.swift
intergalacticspacehighway: I [removed](https://github.com/expo/expo/pull/47619/changes/901164b016b93cbb9321cd012ba263bbc809e929) the fallback since we support min ios 16.4 now. I still had to keep the check because our prebuild is using 16.0 instead of 16.4. Looking into that.

PR #42941, opened 2026-02-06
[babel-preset][widgets] Add babel plugin for new expo-widgets API

jakex7:
# Why

With new expo-widgets, widget components are evaluated in the JSC context in widget itself instead of the app.

# How

To get that function, we need to transform the component expression to string containing the component function. The plugin works similar to worklets plugin and changes code only when `'widget';` directive is present.

Widgets has to be "referentially free" functions (they can use only functions/components imported from `@expo/ui` so there is no need to collect the context.  

# Test Plan

Tests should pass.

# Checklist

- [x] I added a `changelog.md` entry and rebuilt the package sources according to [this short guide](https://github.com/expo/expo/blob/main/CONTRIBUTING.md#-before-submitting)
- [x] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: updated a module plugin).
- [x] Conforms with the [Documentation Writing Style Guide](https://github.com/expo/expo/blob/main/guides/Expo%20Documentation%20Writing%20Style%20Guide.md)

PR #47626, opened 2026-07-08
[babel-preset-expo][widgets] Escape backslashes when serializing widget layouts

alecmolloy:
# Why

The `expo-widgets` babel plugin serializes each `'widget'` function into a template literal. At runtime the app evaluates that template literal to get the layout **string**, stores it in the app group, and the widget extension renders it via a JS-runtime `eval('(' + layout + ')')`.

`escapeTemplateLiteral` escapes `` ` `` and `${` but **not backslashes**. So any escape sequence inside a *string literal* in the widget source (e.g. `'a\nb'`) gets re-cooked when the outer template literal is evaluated: the `\n` becomes a real newline, turning the serialized function into invalid source. On device this surfaces as a red error box reading **`SyntaxError: Unexpected EOF`**, and the widget never renders.

This affects any widget or Live Activity whose layout contains a string literal with an escape sequence — multi-line text, tabs, `\uXXXX`, `\\`, etc. It's been present since the plugin was introduced in #42941.

# How

Escape backslashes first in `escapeTemplateLiteral`, so escape sequences survive the round-trip through the template literal:

```ts
value
  .replace(/\\/g, '\\\\') // must run first
  .replace(/`/g, '\\`')
  .replace(/\$\{/g, '\\${')
```

# Test Plan

Minimal repro (fails on `main`, passes with this change):

```js
const babel = require('@babel/core')
const { widgetsPlugin } = require('babel-preset-expo/build/plugins/widgets-plugin')

const src = `export function MyWidget(props, environment) {
  'widget'
  const text = 'line one\\nline two'
  return null
}`

const { code } = babel.transformSync(src, {
  filename: 'MyWidget.tsx', babelrc: false, configFile: false,
  presets: ['@babel/preset-typescript'], plugins: [widgetsPlugin],
})

// Extract the emitted template literal and cook it exactly as the runtime does:
const raw = code.slice(code.indexOf('`') + 1, code.lastIndexOf('`'))
const layout = eval('`' + raw + '`')
new Function('return (' + layout + ')') // main: throws SyntaxError; fixed: OK
```

- **Before:** `layout` contains a raw newline inside `'line one<newline>line two'` → `SyntaxError` (JavaScriptCore: `Unexpected EOF`).
- **After:** `layout` contains `'line one\nline two'` with the escape preserved → parses fine, widget renders.

### DIFF
--- a/packages/babel-preset-expo/CHANGELOG.md
+++ b/packages/babel-preset-expo/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 - Inline `EXPO_PUBLIC_USE_RN_FETCH` inside `node_modules` so the `expo/fetch` opt-out works in production builds. ([#46986](https://github.com/expo/expo/pull/46986) by [@intergalacticspacehighway](https://github.com/intergalacticspacehighway))
 - Bump `babel-plugin-syntax-hermes-parser` to `^0.36.0` to parse newer Flow syntax (e.g. `readonly` property modifiers) shipped in recent React Native versions ([#46636](https://github.com/expo/expo/pull/46636) by [@zoontek](https://github.com/zoontek))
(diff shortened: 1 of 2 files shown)


jakex7 (approved):
Good catch, thank you!

PR #46733, opened 2026-06-10
[ios][expo] Add ExpoAppSceneDelegate base class for the UIScene life cycle

alanjhughes:
# Why

iOS 27 requires us to adopt the scene-based life cycle. React Native core hasn't migrated so we need to provide the scene infrastructure itself.

# How
- Adds ExpoAppSceneDelegate, a UIWindowSceneDelegate base class. On scene(_:willConnectTo:) it creates the UIWindow from the connecting UIWindowScene and hands it to the factory.
- Bridges scene events into the existing `ExpoAppDelegateSubscriberManager`.
- Adds ExpoReactNativeFactoryProvider, a small protocol the app delegate conforms to so the scene delegate can retrieve the factory it created in `didFinishLaunchingWithOptions`.

# Test Plan
Bare expo, app launches, deep links, and AppState transitions verified there.

PR #47628, opened 2026-07-08
[ios][expo] Deliver cold-start deep links under the scene life cycle

tsapeta:
## Why

Since the app adopted the UIKit scene life cycle in #46733, a deep link (or universal link) that **cold-starts** the app on iOS is dropped: nothing navigates.

`Linking.getInitialURL()` returns `null`, and the `url` event never reaches JS either. This affects every app built with the scene life cycle, not just a test harness. I found it while removing the deep-link confirmation prompt from the bare-expo e2e flows (#47614): once the prompt stopped acting as an accidental warm-up, the first deep link in the flow started failing 100% of the time, and the failure reproduces deterministically with a single `simctl openurl` against a terminated app (the app launches to its default screen and stays there).

## How

`ExpoAppSceneDelegate.scene(_:willConnectTo:options:)` started React Native with `launchOptions: nil`. Two problems flow from that:

- React Native's `RCTLinkingManager.getInitialURL` reads the initial URL only from `bridge.launchOptions[UIApplicationLaunchOptionsURLKey]`. With `nil` launch options that key is never present, so `getInitialURL()` always resolves `null`.
- The URL is separately routed to `RCTLinkingManager.application(_:open:options:)`, which posts the `url` event. At cold start that fires milliseconds into launch, before the JS bundle has registered a listener, so it's lost.

Under the scene life cycle UIKit hands the cold-start URL and user activities to the delegate in `connectionOptions`, not in the app delegate's launch options. The fix rebuilds the launch options from `connectionOptions` (the URL and any browsing-web `NSUserActivity`) and passes them to `startReactNative`, so `getInitialURL()` behaves exactly as it did under the app-delegate life cycle this replaced. Warm-path `url` routing is unchanged.

The keys are built from their underlying constant strings rather than the `UIApplication.LaunchOptionsKey` accessors, because those accessors are deprecated as of iOS 26 in favor of the scene APIs, while React Native still reads the legacy keys.

## Test Plan

- New `ExpoAppSceneDelegateTests` cases cover the launch-options reconstruction (URL, browsing-web activity, and the empty case). `et native-unit-tests -p ios --packages expo` passes.
- Manual: with a bare-expo build that includes this change, a cold-start deep link (`xcrun simctl openurl <udid> "bareexpo://test-suite/run?tests=AppMetrics"` against a terminated app) navigates straight to the test runner instead of stalling on the selection screen. Before the fix the same command leaves the app idle on its default screen.

### DIFF
--- a/packages/expo/ios/AppDelegates/ExpoAppSceneDelegate.swift
+++ b/packages/expo/ios/AppDelegates/ExpoAppSceneDelegate.swift
@@ -44,12 +44,20 @@ open class ExpoAppSceneDelegate: UIResponder, UIWindowSceneDelegate {
     // `UIApplication.shared.delegate?.window` keeps working (e.g. expo-system-ui).
     provider.window = window
 
-    // launchOptions is nil under the scene life cycle; cold-start URLs/activities arrive via
-    // `connectionOptions` and are routed below.
+    // Under the scene life cycle UIKit passes cold-start URLs and activities in `connectionOptions`
+    // rather than in the app delegate's launch options. React Native's `Linking.getInitialURL()`
+    // only reads them from launch options, so rebuild them here; otherwise a link that cold-starts
+    // the app is delivered to no one, because the `url` event routed below fires before JS is ready.
+    let browsingWebActivity = connectionOptions.userActivities.first {
+      $0.activityType == NSUserActivityTypeBrowsingWeb
+    }
     factory.startReactNative(
       withModuleName: provider.reactNativeFactoryModuleName,
       in: window,
-      launchOptions: nil
+      launchOptions: Self.launchOptions(
+        url: connectionOptions.urlContexts.first?.url,
+        userActivity: browsingWebActivity
+      )
     )
 
     // Deep links / universal links.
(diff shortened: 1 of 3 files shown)


packages/expo/ios/AppDelegates/ExpoAppSceneDelegate.swift
```diff
@@ -124,6 +150,20 @@ open class ExpoAppSceneDelegate: UIResponder, UIWindowSceneDelegate {
...
+    var options: [UIApplication.OpenURLOptionsKey: Any] = [:]
+    if let sourceApplication = sceneOptions.sourceApplication {
+      options[.sourceApplication] = sourceApplication
+    }
+    if let annotation = sceneOptions.annotation {
+      options[.annotation] = annotation
+    }
+    options[.openInPlace] = sceneOptions.openInPlace
+    return options
+  }
```
tsapeta, line 166: I only moved this function just so launch options and routing helpers are within one extension. Implementation is the same as it was.

packages/expo/ios/AppDelegates/ExpoAppSceneDelegate.swift
```diff
@@ -87,6 +95,38 @@ open class ExpoAppSceneDelegate: UIResponder, UIWindowSceneDelegate {
...
+// MARK: - Launch options & routing helpers
+
+extension ExpoAppSceneDelegate {
+  /// Rebuilds the launch options that `Linking.getInitialURL()` reads from a scene's connection
+  /// options. Returns `nil` when the app wasn't cold-started by a URL or a browsing-web activity,
+  /// so it can be forwarded to `startReactNative` as-is.
+  static func launchOptions(
+    url: URL?,
+    userActivity: NSUserActivity?
+  ) -> [UIApplication.LaunchOptionsKey: Any]? {
```
tsapeta, line 109: Open question: we could make it a helper that translates `connectionOptions` to `launchOptions`, then finding the `browsingWebActivity` would be moved here and `willConnectTo` would be cleaner. It makes it a bit less testable as `ConnectionOptions` have no public initializer, but I can find another way to test it 😉

packages/expo/ios/AppDelegates/ExpoAppSceneDelegate.swift
```diff
@@ -87,6 +95,38 @@ open class ExpoAppSceneDelegate: UIResponder, UIWindowSceneDelegate {
...
+      let urlKey = UIApplication.LaunchOptionsKey(rawValue: "UIApplicationLaunchOptionsURLKey")
+      launchOptions[urlKey] = url
+    }
+    if let userActivity {
+      let userActivityDictionaryKey = UIApplication.LaunchOptionsKey(
+        rawValue: "UIApplicationLaunchOptionsUserActivityDictionaryKey"
+      )
+      launchOptions[userActivityDictionaryKey] = [
+        "UIApplicationLaunchOptionsUserActivityTypeKey": userActivity.activityType,
+        "UIApplicationLaunchOptionsUserActivityKey": userActivity,
```
tsapeta, line 125: I tried to avoid hardcoding these keys as strings, but while `UIApplicationLaunchOptionsURLKey` and `UIApplicationLaunchOptionsUserActivityTypeKey` can be accessed as enum's raw value (`UIApplication.LaunchOptionsKey.url.rawValue`), `UIApplicationLaunchOptionsUserActivityKey` is not exposed that way 😞 So I preferred consistency.
In addition, these raw values would produce deprecation warnings.

alanjhughes (approved):

Kudo (approved):
great catching for the problem!

PR #46667, opened 2026-06-08
feat(cli): Replace event logging with extracted package

kitten:
## Summary

This is an experimental change to "productionise" the `src/events` logic into an external package, called `2g`. This is unfinished, and pushed for testing. `2g` ships its own CLI and agents can be instructed to use it to diagnose running `expo start` tasks.

Eventually, we'll be able to expand this to more features and replace more/all `debug` calls. The events we create can be split into "debug" and normal events. The former aren't by default presented to users with `2g tap`.

An agent can be instructed to perform a task while introspecting an `expo start` background process with the `2g` CLI. Based on the `--help` output (e.g. `pnpx 2g --help`) the agent should have enough information to use it effectively.

https://github.com/kitten/2g

### Out of scope

Some sub-tasks of this effort are out of scope:
- Answering whether the `2g` CLI is used effectively for debug events (once we use `events.debug`)
- Recategorising some existing events as `events.debug`
- Adding missing events and replacing more `debug` calls
- Some testing on sub-workers, on Windows, and of `2g export`
- Implementing the "freestanding" `expo logs` sub-command

## Set of changes

- Add `2g` package
- Migrate to slightly updated events and typings API

PR #47655, opened 2026-07-09
feat(cli,metro-config,metro-file-map): Add/Migrate debug events to event logs

kitten:
# Why

Stacked on #46667

This is an automated migration to replace more `debug` usages with event logs. We're only replacing `debug` in non-forked files and paths the the `@expo/cli` can reach directly. We're not yet migrating all calls that sit in the `@expo/cli` execution path and are limiting ourselves to `@expo/metro-config` and `@expo/metro-file-map` (partial, where we've already diverged / haven't forked)

The events should give us a general debug surface for several spans and events that were previously uncovered or only covered by `debug` events.

# How

- Migrate several CLI events
- Ensure we have good transform worker coverage
- Add `devserver:url` event as a non-debug event
- ... (too many to list, but mostly `debug` migration)

# Test Plan

Tell Claude:
> With the 2g CLI, record a trace for `expo start` (with `--clear`) on `apps/router-e2e` making a bundle request, and then export it to a Chrome trace

The `2g typegen` command can give an overview of all available events when run in `@expo/cli`.

# Checklist


packages/@expo/cli/src/start/interface/interactiveActions.ts
```diff
@@ -56,7 +55,7 @@ export async function printDevToolsPluginCliBannersAsync(
 
     return bannerItems.length;
   } catch (error: any) {
-    debug(`Failed to print DevTools plugin CLI banners: ${error.toString()}`);
+    event('banners_failed', { error: event.error(error as Error) });
```
hassankhan, line 58: Definitely makes more sense to use a regular event here 💯

hassankhan (approved):

PR #47656, opened 2026-07-09
[inline-modules] Merge package and module definition search

HubertBer:
# Why
Kotlin Inline modules silently fail to be resolved when the package declaration is not at the beginning of the file but after some comment which leads to them not being autolinked.

When inline modules were added we've opted to look for package declaration in the file header - first 512 bytes, however right now we either way have to match entire file for ModuleDefinition to check whether the inline modules file is compile only or not, so both can be merged into a one full file read.

# How
Merge `getKotlinFileNameWithItsPackage` and `hasKotlinModuleDefinition` into one `getKotlinInlineModuleInfo`, now there is no HEADER_SIZE, as `hasKotlinModuleDefinition` already checked the whole file.

# Test Plan
- [x] updated tests
- [x] checked bare-expo

# Checklist

### DIFF
--- a/packages/expo-modules-autolinking/src/inlineModules/iosInlineModules.ts
+++ b/packages/expo-modules-autolinking/src/inlineModules/iosInlineModules.ts
@@ -5,7 +5,8 @@ import { getMirrorStateObject } from './inlineModules';
 export async function getIosInlineModulesClassNames(
   options: InlineModulesScanOptions
 ): Promise<ModuleIosConfig[]> {
-  return (await getMirrorStateObject(options)).swiftModuleClassNames.map((className: string) => {
+	const stateObject = await getMirrorStateObject(options);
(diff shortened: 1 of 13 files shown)


packages/expo-modules-autolinking/src/inlineModules/iosInlineModules.ts
```diff
@@ -1,18 +1,16 @@
...
-        name: null,
-      };
-    }
-  );
+  return (await getMirrorStateObject(options)).swiftModuleClassNames.map((className: string) => {
+    return {
+      class: className,
+      name: null,
+    };
+  });
```
lukmccall:
```suggestion
	const stateObject = await getMirrorStateObject(options);
  return stateObject.swiftModuleClassNames.map((className: string) => {
    return {
      class: className,
      name: null,
    };
  });
```

lukmccall (approved):

PR #47148, opened 2026-06-22
[cellular][Android] Remove deprecated `allowsVoipAsync`

Wenszel:
# Why

`allowsVoipAsync` was deprecated in SDK 56 because `SipManager` has been deprecated since API level 31. It can now be removed.

# How

Removes the `allowsVoipAsync` function.

# Test Plan

BareExpo

PR #47663, opened 2026-07-09
[docs] Regenerate unversioned API data

ramonclaudio:
Regenerates the unversioned API data for every package where the committed JSON no longer matches a clean `et generate-docs-api-data` run (32 files), and drops the cellular page intro's mention of VoIP support, whose API is gone.

The last full refresh was the SDK 56 cut-off (#45215). The TypeScript 6 + TypeDoc bump (#45370) landed a day later without a regen, and nothing has been regenerated in bulk since. The result: `expo-cellular.json` still documents `allowsVoipAsync` (removed in #47148), `expo-location.json` is missing the Motion Activity API, and ~25 more files differ only by TypeScript 6 union ordering. Production is unaffected (the deployed site regenerates this data at build time), but local docs dev renders stale content and any PR that regenerates a single package inherits all of this in its diff.

Same as #42376, #41937, and #38824. `expo-speech.json` is deliberately absent: its mapping entry points at a moved source file, fixed in #47670.

# Test Plan

- Reproducible: `et generate-docs-api-data` on `main` produces exactly this diff, byte-identical across three runs.
- `docs`: `pnpm test`, `pnpm test:worker`, `pnpm lint --max-warnings 0`, `pnpm lint-links`, and `pnpm lint-prose` all pass. `pnpm export` builds the full site cleanly.
- Spot-checked locally against production: the cellular page no longer documents `allowsVoipAsync` and the location page documents the Motion Activity API, matching https://docs.expo.dev/versions/unversioned/ (which already renders regenerated data).

# Checklist

  - No package sources changed, generated docs data only.
  - Docs-only change, nothing consumed by prebuild or EAS Build.


amandeepmittal (approved):
Thank you!

PR #47666, opened 2026-07-09
[fingerprint] Fix config-plugin module capture with a compile hook

Kudo:
# Why

Fingerprinting crashed on projects with a local TypeScript config plugin — the loader recorded a `.js` path for a `.ts` source and then couldn't open it. It also silently dropped virtual (compiled-in-memory) modules. Rework upon #46092

close ENG-18201

# How

Replaced the post-hoc `module._cache` diff with a `Module.prototype._compile` hook, so we keep each module's real filename and source content. Real files are hashed from disk (so the hash stays transpiler-independent); modules with no file on disk are hashed from their captured contents. This finishes the long-standing loader-hook TODO in `ExpoConfigLoader`.

# Test Plan

`et check-packages @expo/fingerprint`. Added unit tests for the capture hook and source classification, and ran the built loader against `apps/bare-expo` to confirm the crash is gone.

# Checklist

- [x] Added a `CHANGELOG.md` entry.
- [x] Builds, type-checks, lints, and tests via `et check-packages`.

Issue #44808, opened 2026-04-15
[@expo/fingerprint] dir source hash changes when pnpm peer-dep paths reshuffle, even if file contents are unchanged


aamagda:
### Minimal reproducible example

https://github.com/aamagda/expo-fingerprint-pnpm-repro/tree/master

### Steps to reproduce                     
              
1. Clone https://github.com/aamagda/expo-fingerprint-pnpm-repro
2. Run: `./reproduce.sh` that
  - bumps `expo@52.0.48 -> expo@52.0.49` -> reshuffles peer-dep in pnpm virtual paths
  - regenerates fingerprint
  - reports changes in `dir` sources despite no native file modification 

See code changes by commit:
- https://github.com/aamagda/expo-fingerprint-pnpm-repro/tree/master - initial state before peer-dep reshuffles
- https://github.com/aamagda/expo-fingerprint-pnpm-repro/commit/42a7491cdd45396f0e40f3208a505a2651c750ed - updated state after peer-dep reshuffles

### Expected behavior                                                                                                              
                                                                                                                                     
  `fingerprint diff` reports no changes when only pnpm's virtual-store path suffixes changed (peer-dep reshuffling), and actual      
  native file content is identical.
                                                                                                                                     
### Actual behavior                            
              
  `fingerprint diff` reports changes for every `dir` source whose pnpm virtual-store path was affected by the peer-dep reshuffle,    
  even though:
  - all file content hashes (`result.hex`) are identical                                                                             
  - only `result.id` (the path string) changed           

### Environment                                                                                                                    

  - `@expo/fingerprint`: `0.16.6`                                                                                                    
  - Package manager: `pnpm` with default `node-linker=isolated` (virtual store)
  - Platform: any (iOS / Android)                                                                                                    
                                                                                                                                     
  ### Root cause (with code reference)                                                                                               
                                                                                                                                     
  In `@expo/fingerprint/build/hash/Hash.js`, the `createDirHashResultsAsync` function builds the directory hash by feeding **both the
   file path and the file content hash** into the hasher:                                                                          
                                                                                                                                     
  ```js                                          
  for (const result of results) {                                                                                                    
      hasher.update(result.id);   // ← relative file path, includes full pnpm virtual-store path
      hasher.update(result.hex);  // ← content hash                                                                                  
  }                                              
  ```                                                                                                                                
                                                 
  `result.id` is produced by `createFileHashResultsAsync`:                                                                           
  ```js
[...]

PR #47667, opened 2026-07-09
[fingerprint] Add a package source type

Kudo:
# Why

The fingerprint presets (next PR in the stack) need a way to hash a native dependency by its identity rather than its whole directory, so patch-level or cross-machine churn inside `node_modules` doesn't flip the fingerprint.

close ENG-18201

# How

Added a `package` source type that hashes only a package's `name` and `version` from its `package.json`, with matching sort, dedup, and debug-info handling. It's inert on its own — nothing emits `package` sources yet — so this changes no fingerprints by itself.

# Test Plan

`et check-packages @expo/fingerprint`. Added unit tests: the hash is stable across non-version churn, changes on a version bump, and is null when the `package.json` is missing.

# Checklist


Kudo:
* **#47668** <a href="https://app.graphite.com/github/pr/expo/expo/47668?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#47667** <a href="https://app.graphite.com/github/pr/expo/expo/47667?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/47667?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#47678** <a href="https://app.graphite.com/github/pr/expo/expo/47678?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/@expo/fingerprint/src/Fingerprint.types.ts
```diff
@@ -246,7 +246,27 @@ export interface HashSourceContents {
...
 }
 
-export type HashSource = HashSourceFile | HashSourceDir | HashSourceContents;
+export interface HashSourcePackage {
+  type: 'package';
+  /**
+   * Path to the package's `package.json`. The hash is derived from the package `name` and
+   * `version` only, so unrelated churn inside the package (or across machines) is ignored.
+   */
+  filePath: string;
```
kitten, line 270: nit: Assuming we're sourcing this from multiple sources, even then, it might be beneficial to embed the information here and front-load it. Mainly for two reasons:
- it'd be easier to spot in the output JSON for debugging (`name` and `version` embedded)
- it'd be directly sourcable from `expo-modules-autolinking` output

packages/@expo/fingerprint/src/hash/Hash.ts
```diff
@@ -287,6 +292,45 @@ export async function createContentsHashResultsAsync(
...
+  let name = '';
+  let version = '';
+  try {
+    const raw = await fs.readFile(path.join(projectRoot, source.filePath), 'utf8');
+    const packageJson = JSON.parse(raw);
+    name = packageJson.name ?? '';
+    version = packageJson.version ?? '';
+  } catch (error: any) {
+    if (error.code === 'ENOENT') {
+      return null;
```
kitten: nit: We should account for this potentially and fallback to a special ENOENT entry on the relative file path maybe? That'd ensure that two different ways of arriving at this error hash differently.

That's a minor detail and I'm not 100% sure if it'll ever come into play, but it'd be a "cheap" safety change

packages/@expo/fingerprint/src/hash/Hash.ts
```diff
@@ -287,6 +292,45 @@ export async function createContentsHashResultsAsync(
...
+    throw error;
+  }
+
+  const hex = createHash(options.hashAlgorithm)
+    .update(JSON.stringify({ name, version }))
+    .digest('hex');
+  const debugInfo = options.debug ? { path: source.filePath, name, version, hash: hex } : undefined;
+  return {
+    type: 'package',
+    id: createSourceId(source),
```
kitten, line 316: This is a bit of a caveat, we likely don't want? I think, we should avoid leaking the source path into the hash here. It's basically just a source of instability and once we're trusting the package name+version (and patches are sourced separately), the source path from `node_modules` (and/or otherwise) only serves as indirection.

This could come into play with isolated modules changing the path. That can sometimes be beneficial, but also can sometimes be incorrect. Generally, we should probably avoid divergence depending on install behaviour once we use the `package` type, so I'd strip the path information here

kitten: As per: https://github.com/expo/expo/issues/44808

Kudo: good catch! the id is now hashed by name+version.

packages/@expo/fingerprint/src/Fingerprint.types.ts
Kudo: as your third comment, now we hashing by package name+version. for traceability, we still keep filePath in the fingerprint but it does not affect hash result

packages/@expo/fingerprint/src/hash/Hash.ts
Kudo: i'd keep ENOENT returning null. using filePath as fallback may result in unstable fingerprint. the `type: package` is for balanced/relaxed preset, it's fine to keep the fingerprint loose.

kitten (approved):

Kudo:
### Merge activity

* **Jul 17, 3:42 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/expo/expo/47667).
* **Jul 17, 3:46 PM UTC**: [Graphite](https://app.graphite.com/github/pr/expo/expo/47667) rebased this pull request as part of a merge.
* **Jul 17, 3:46 PM UTC**: @Kudo merged this pull request with [Graphite](https://app.graphite.com/github/pr/expo/expo/47667).

PR #47668, opened 2026-07-09
[fingerprint] Add fingerprint presets

Kudo:
# Why

Fingerprint has always been strict, so people who just bump their app version or name kept getting a new fingerprint and an unnecessary rebuild. Presets give a friendlier default without having to hand-tune `sourceSkips`.

close ENG-18201

# How

Added `strict` / `balanced` / `relaxed` presets (set via `preset` in **fingerprint.config.js** or the `--preset` CLI flag), and made `balanced` the default:

- `balanced` skips app version and string runtime version churn, hashes autolinked packages by their version, and collapses `node_modules` config plugins to their package version (in-repo plugins are still hashed from their files).
- `relaxed` additionally ignores names, identifiers, schemes, and assets, for building multiple variants from one native project.

An explicit `sourceSkips` still overrides the preset. This changes the default fingerprint once on SDK upgrade; `preset: 'strict'` restores the previous behavior.

# Test Plan

`et check-packages @expo/fingerprint` and regenerated e2e snapshots. Added unit tests for preset resolution/precedence, the package-mode autolinking wiring (with a `dir` fallback when there's no `package.json`), and the scoped config-plugin trace.

# Checklist


Kudo:
* **#47668** <a href="https://app.graphite.com/github/pr/expo/expo/47668?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/47668?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#47667** <a href="https://app.graphite.com/github/pr/expo/expo/47667?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#47678** <a href="https://app.graphite.com/github/pr/expo/expo/47678?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/@expo/fingerprint/src/Presets.ts
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * - `strict`: highest fidelity - the historical default. Only skips prebuild-mutated package.json
+ *   scripts so a fingerprint stays consistent before and after prebuild.
+ * - `balanced`: the default. Also ignores app version and string runtime version churn, and hashes
+ *   autolinked packages and node_modules config-plugin modules by their `package.json` version.
+ *   Best first-time experience.
+ * - `relaxed`: for building multiple variants from one native project. Additionally ignores app
+ *   names, bundle identifiers, schemes, and assets, while still hashing the config-plugins list so
+ *   adding a plugin still changes the fingerprint.
+ */
```
kitten, line 30: This is more of a "conceptual" nit, but I'd say that the best DX here would be if the default (balanced) does the right thing 99% of the time.

That also implies though that the other times people will want to change this. Ideally, they'd then only have to pick a preset. However, I think offering two here implies that we're maybe not too confident about the `balanced` preset?

- Can we detect variant switching and/or always ignore some of it? (I'd assume that we'd want different hashes across variants in many cases anyway, since mixing prod/staging might be undesirable, so there may be a middleground here)
- The note states that this "still hashes" config-plugins, so the difference here feels a bit redundant. If it's not fully relaxed, does this only serve as a variant distinguisher? I'd say that doesn't potentially merit the preset, and we should maybe see if we can combine `balanced` and `relaxed`

That's a bit vague, sorry 😅 But mainly, I'd expect most people not to dive too deep into configuration here, unless things have gone very wrong. If we can make more decisions for them, that'd be nice

In essence/tl;dr:
- Would this mean we'd add more and more presets? or does that mean we stop here arbitrarily?
- If we stopped at having three, what stops us from having just two?

Kudo: yes i would have a good default and expect people not to configure presets or other advanced settings. i'm just not quite confident and clear about what's a good default. build variants could have quite different setup. i do see people have different appName, appId, googleServiceFile, or even sentry project inside the `plugins`. i was not confident enough to ignore all the config-plugins for relaxed preset yet. that's still a limitation for relaxed preset.

all in all, the refactoring this time is mainly to resolve the autolinked node_modules problems and provide "preset" to allow people configuring the behavior. we may fine tune the behavior in the relaxed preset in the future or even introduce new advanced preset.

kitten: I suppose what could be a good follow-up would be:

- differentiate between static & dynamic config when both are present (when both are present we could ignore the dynamic config as a "you know what you're doing" case, or scope down what we use from it _heavily_
- separate out plugin resolution and limit what's done depending on if that's a package or local module entirely separate from the app config fingerprinting


Not sure what you think about that. But those are separate ideas from the existing PRs anyway

kitten (approved):

packages/@expo/fingerprint/src/Presets.ts
Kudo: if `@expo/config` could offer better detection, that would be great. i would leave it as a follow up once we have clear idea how to achieve that either in `@expo/config` or `@expo/require-utils`.

Kudo (commented):
going to merge the prs. thanks again for the great review!

Kudo:
### Merge activity

* **Jul 17, 3:42 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/expo/expo/47668).
* **Jul 17, 3:48 PM UTC**: [Graphite](https://app.graphite.com/github/pr/expo/expo/47668) rebased this pull request as part of a merge.
* **Jul 17, 3:48 PM UTC**: @Kudo merged this pull request with [Graphite](https://app.graphite.com/github/pr/expo/expo/47668).

PR #47670, opened 2026-07-09
[tools][docs] Fix dead docs data mapping entries, report all failures

ramonclaudio:
Two entries in the docs data generator's package mapping point at files that no longer exist: `expo-ui/jetpack-compose/filterchip` (FilterChip merged into `Chip` in #43900) and `expo-speech` (`Speech/Speech.ts` flattened to `Speech.ts` in #44791). Only the first `Promise.all` rejection got logged, so the FilterChip error hid the speech one, and `expo-speech.json` has been unregenerable since April, its content frozen at a February refresh.

This fixes both entries, reports every failure by package name (`Promise.allSettled`, exit semantics unchanged), and adds a unit test asserting each mapping entry resolves to an existing source file, so the next source move fails `tools` tests instead of silently breaking generation. Includes the regenerated `expo-speech.json`, which only generates with the fixed mapping. The rest of the drifted unversioned data was refreshed in #47663.

# Test Plan

- `et generate-docs-api-data -p expo-speech` fails silently on `main` (no output file, exit 0) and generates the file with this change.
- A forced failure (`et gdad -p expo-speech -s 999`) reports the failing package by name.
- `cd tools && pnpm test` (322 tests, including the new mapping guard) and `pnpm lint` pass.
- Full `et generate-docs-api-data` run completes with every mapping key generating output.

# Checklist

  - No package sources changed, `tools` and generated docs data only.
  - Repo tooling only, nothing consumed by prebuild or EAS Build.
  - No documentation prose changes.


amandeepmittal (approved):
This is amazing! Thank you for catching these errors in the docs generation pipeline and `et gdad` improvements!

ramonclaudio:
> This is amazing! Thank you for catching these errors in the docs generation pipeline and `et gdad` improvements!

No problem! 🙏🏽

PR #47671, opened 2026-07-09
[bare-expo] Verify the Maestro install and fail honestly when it is missing

tsapeta:
# Why

This [Test Suite e2e run on main](https://github.com/expo/expo/actions/runs/29032108430/job/86170497034) failed in a misleading way. The chain of events:

1. The Maestro download from `get.maestro.mobile.dev` hit a CDN 504. The installer printed "Downloaded zip archive is corrupt" but exited 0, so the install step went green with no `maestro` binary installed.
2. Every Maestro invocation then failed instantly with `ENOENT` (the per-attempt `test duration` logs read 3.02ms, 1.52ms, 0.84ms).
3. The runner's crash heuristic found no app process (it was never launched) and printed "The runner app has probably crashed" with empty logs, pointing anyone reading the log at a crash that never happened.
4. All retries burned in minutes and the job failed.

Separately, the iOS job installed whatever the latest Maestro release was at run time, so a new release could break iOS CI overnight with no change in this repo (this already forced pinning on Android, where 2.5.0+ breaks the emulator ADB connection).

# How

- The install moves into a `setup-maestro` composite action that verifies `maestro --version` runs after the installer finishes (the only reliable check, since the installer swallows download failures) and retries the download up to 3 times before failing the step. All four e2e jobs use it, including the two in `test-suite-nightly.yml` that were still unpinned and unhardened.
- iOS is pinned to Maestro 2.6.1, the version yesterday's passing runs resolved to; Android stays on 2.4.0 for the known gRPC/ADB reason.
- `runMaestroAsync` recognizes an `ENOENT` spawn failure and throws "The maestro binary was not found on PATH..." instead of falling through to flow-failure handling, so the job fails fast on the first invocation with the real cause instead of retrying and misreporting an app crash.

# Test Plan

- `bun test scripts/lib/` in `apps/bare-expo` passes (15 tests).
- Workflow YAML parses.
- The Test Suite e2e runs on this PR exercise both modified install steps end to end.

### DIFF
--- a/.github/workflows/test-suite-nightly.yml
+++ b/.github/workflows/test-suite-nightly.yml
@@ -8,10 +8,12 @@ on:
     branches: [main]
     paths:
       - .github/workflows/test-suite-nightly.yml
+      - .github/actions/setup-maestro/action.yml
       - tools/src/commands/SetupReactNativeNightly.ts
   pull_request:
     paths:
       - .github/workflows/test-suite-nightly.yml
+      - .github/actions/setup-maestro/action.yml
       - tools/src/commands/SetupReactNativeNightly.ts
 
 concurrency:
--- a/.github/workflows/test-suite-nightly.yml
+++ b/.github/workflows/test-suite-nightly.yml
@@ -246,7 +246,7 @@ jobs:
         with:
           # Pinned to 2.4.0 — 2.5.0+ replaces TCP port forwarding with a direct ADB
           # socket gRPC connection that fails (UNAVAILABLE / tcp:7001 closed) against
-          # the Android emulator in CI.
+          # the Android emulator in CI; verified still broken in 2.6.1.
           version: 2.4.0
       - name: 🔨 Install pnpm
         uses: pnpm/action-setup@0e279bb959325dab635dd2c09392533439d90093 # v6.0.8
--- a/.github/workflows/test-suite-nightly.yml
+++ b/.github/workflows/test-suite-nightly.yml
@@ -244,10 +244,10 @@ jobs:
       - name: 🍺 Install Maestro
         uses: ./.github/actions/setup-maestro
         with:
-          # Trying 2.6.1: 2.5.0 replaced TCP port forwarding with a direct ADB socket
-          # gRPC connection that failed (UNAVAILABLE / tcp:7001 closed) against the
-          # Android emulator in CI; checking whether 2.6.x still has the problem.
-          version: 2.6.1
+          # Pinned to 2.4.0 — 2.5.0+ replaces TCP port forwarding with a direct ADB
+          # socket gRPC connection that fails (UNAVAILABLE / tcp:7001 closed) against
+          # the Android emulator in CI.
+          version: 2.4.0
       - name: 🔨 Install pnpm
         uses: pnpm/action-setup@0e279bb959325dab635dd2c09392533439d90093 # v6.0.8
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/test-suite-nightly.yml
```diff
@@ -241,12 +242,12 @@ jobs:
...
       - name: 🍺 Install Maestro
-        run: |
+        uses: ./.github/actions/setup-maestro
+        with:
           # Pinned to 2.4.0 — 2.5.0+ replaces TCP port forwarding with a direct ADB
           # socket gRPC connection that fails (UNAVAILABLE / tcp:7001 closed) against
-          # the Android emulator in CI.
-          curl -Ls "https://get.maestro.mobile.dev" | MAESTRO_VERSION=2.4.0 bash
-          echo "${HOME}/.maestro/bin" >> $GITHUB_PATH
+          # the Android emulator in CI; verified still broken in 2.6.1.
```
Kudo:
```suggestion
```

the comment above explains well already

.github/workflows/test-suite-nightly.yml
```diff
@@ -116,9 +116,10 @@ jobs:
           xcrun simctl shutdown all
           sudo xcodebuild -runFirstLaunch
       - name: 🍺 Install Maestro
-        run: |
-          curl -Ls "https://get.maestro.mobile.dev" | bash
-          echo "${HOME}/.maestro/bin" >> $GITHUB_PATH
+        uses: ./.github/actions/setup-maestro
+        with:
+          # Pinned so that a new Maestro release can't break CI without a change in this repo.
```
Kudo:
```suggestion
```

pinning version is common and we don't need comment

.github/workflows/test-suite.yml
```diff
@@ -268,9 +268,10 @@ jobs:
           xcrun simctl shutdown all
           sudo xcodebuild -runFirstLaunch
       - name: 🍺 Install Maestro
-        run: |
-          curl -Ls "https://get.maestro.mobile.dev" | bash
-          echo "${HOME}/.maestro/bin" >> $GITHUB_PATH
+        uses: ./.github/actions/setup-maestro
+        with:
+          # Pinned so that a new Maestro release can't break CI without a change in this repo.
```
Kudo:
```suggestion
```

.github/workflows/test-suite.yml
```diff
@@ -465,12 +466,12 @@ jobs:
...
       - name: 🍺 Install Maestro
-        run: |
+        uses: ./.github/actions/setup-maestro
+        with:
           # Pinned to 2.4.0 — 2.5.0+ replaces TCP port forwarding with a direct ADB
           # socket gRPC connection that fails (UNAVAILABLE / tcp:7001 closed) against
-          # the Android emulator in CI.
-          curl -Ls "https://get.maestro.mobile.dev" | MAESTRO_VERSION=2.4.0 bash
-          echo "${HOME}/.maestro/bin" >> $GITHUB_PATH
+          # the Android emulator in CI; verified still broken in 2.6.1.
```
Kudo:
```suggestion
```

Kudo (approved):

.github/workflows/test-suite-nightly.yml
tsapeta: Well, it wasn't pinned on iOS so not that common 😅 And on Android only because 2.5 broke it.

.github/workflows/test-suite-nightly.yml
tsapeta: I think it's still worth noting that I tried 2.6.1 and it's still broken.

Kudo (approved):

PR #47678, opened 2026-07-10
[fingerprint] Derive config-plugin modules by diffing a plugins-skipped pass

Kudo:
# Why

Capturing config-plugin modules relied on a large hand-maintained allowlist (`DEFAULT_CONFIG_LOADING_IGNORE_PATHS`, ~50 packages) to exclude the config-loading framework (Babel, TypeScript, `@expo/config`, …). It drifts every time those dependencies change.

close ENG-18201

# How

Load the config twice in parallel — once fully, once with `skipPlugins` — and keep only the modules that applying plugins added (`B − A`). The framework loads in both passes, so diffing removes it automatically; in-repo files are always kept, so a local plugin imported at the top of `app.config` is never lost. This retires the allowlist. Two separate processes keep each module cache clean, and `Promise.all` makes the second pass ~free in wall-clock time.

# Test Plan

`et check-packages @expo/fingerprint`. Added unit tests for the diff (framework dropped, plugin modules kept, in-repo always kept). Verified end-to-end against `apps/bare-expo`: 138 plugin modules captured, both local plugins kept, framework excluded, ~186 ms for both parallel passes.

# Checklist

### DIFF
--- a/packages/@expo/fingerprint/src/ExpoConfig.ts
+++ b/packages/@expo/fingerprint/src/ExpoConfig.ts
@@ -18,7 +18,10 @@ export async function getExpoConfigAsync(
   config: ProjectConfig | null;
   loadedModules: LoadedModuleSource[] | null;
 }> {
-  const result = {
+  const result: {
+    config: ProjectConfig | null;
+    loadedModules: LoadedModuleSource[] | null;
+  } = {
     config: null,
     loadedModules: null,
   };
@@ -30,14 +33,15 @@ export async function getExpoConfigAsync(
   const tmpDir = await fs.mkdtemp(path.join(os.tmpdir(), 'expo-fingerprint-'));
   const ignoredFile = await createTempIgnoredFileAsync(tmpDir, options);
   try {
-    const { message } = await spawnWithIpcAsync(
-      'node',
-      [getExpoConfigLoaderPath(), path.resolve(projectRoot), ignoredFile],
-      { cwd: projectRoot }
-    );
-    const stdoutJson = JSON.parse(message);
-    result.config = stdoutJson.config;
-    result.loadedModules = stdoutJson.loadedModules;
+    const [
+      { config, loadedModules: modulesWithPlugins },
+      { loadedModules: modulesWithoutPlugins },
+    ] = await Promise.all([
+      spawnConfigLoaderAsync(projectRoot, ignoredFile, /* skipPlugins */ false),
+      spawnConfigLoaderAsync(projectRoot, ignoredFile, /* skipPlugins */ true),
+    ]);
+    result.config = config;
+    result.loadedModules = diffLoadedModules(modulesWithPlugins ?? [], modulesWithoutPlugins ?? []);
   } catch (e: unknown) {
     if (e instanceof Error) {
       console.warn(`Cannot get Expo config from an Expo project - ${e.message}: `, e.stack);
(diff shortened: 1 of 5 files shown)


Kudo:
* **#47668** <a href="https://app.graphite.com/github/pr/expo/expo/47668?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#47667** <a href="https://app.graphite.com/github/pr/expo/expo/47667?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#47678** <a href="https://app.graphite.com/github/pr/expo/expo/47678?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/47678?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/@expo/fingerprint/src/ExpoConfig.ts
```diff
@@ -51,6 +55,48 @@ export async function getExpoConfigAsync(
...
+/** Stable identity of a loaded module: its file path, or its id for a virtual module. */
+function getLoadedModuleKey(source: LoadedModuleSource): string {
+  return source.type === 'file' ? source.path : source.id;
+}
+
+/** Whether a loaded module is a project-local file (e.g. a local config plugin) rather than a dependency. */
+function isInRepoLoadedModule(source: LoadedModuleSource): boolean {
+  const key = getLoadedModuleKey(source);
+  // `..` means the path escaped the project root (e.g. a hoisted or linked dependency).
+  return !key.startsWith('..') && !key.includes('node_modules');
```
kitten, line 97: This isn't always going to be accurate, since we can keep a "local plugin" in the source root. That's partially fine, but I think we could try to avoid the `--skipPlugins` route with some purpose-written logic in `@expo/config` in the future. There, we have more information/local-knowledge on the control flow, and it'd be nice to simply separate out what we need
(on an internal entrypoint, if need be maybe)

packages/@expo/fingerprint/src/ExpoConfigLoader.ts
```diff
@@ -208,69 +211,37 @@ function setNodeEnv(mode: 'development' | 'production') {
...
-    'ts-interface-checker',
-    'write-file-atomic',
-    'xml2js',
-    'xmlbuilder',
-    'which',
     'uuid',
+    'which',
     'xcode',
+    'xml2js',
+    'xmlbuilder',
```
kitten, line 245: Re prior comment: Some of this basically will still be a bit of an overreach, since we don't differentiate between config-plugin boundaries. Ultimately, it's hard to differentiate what's for a config plugin and what isn't. However, as per the planned change, if we take a similar approach to modules, then we could try to interrupt loading `node_modules` config plugin and replace them with their version from npm, depending on the plugins entry (by name / how it'll be loaded)

kitten (approved):

packages/@expo/fingerprint/src/ExpoConfig.ts
Kudo: yes if `@expo/config` could provide the logic of what plugins were being loaded, it would be better. i was thinking that we probably need to move `module._compile` hook from fingerprint to `@expo/config`. not sure if that's worth to do now.

for `isInRepoLoadedModule`, it's fine for not being precise. we mainly want to ignore 3rd party modules from node_modules.

packages/@expo/fingerprint/src/ExpoConfigLoader.ts
Kudo: i agree that not ideal. if we could find a proper way to implement the detection in `@expo/config`. that would help. it's still unclear how/when to implement the detection properly

packages/@expo/fingerprint/src/ExpoConfig.ts
kitten: We could move a helper into `@expo/require-utils` since it has similar responsibilities then piece it together in an internal config CLI maybe? There's a few options, but definitely not pressing now.

Mainly, I'm just concerned about us taking the full "config path" while we're mainly interested in the plugins.

That depends on my comment in the other PR though, and whether we're interested in separating our special dynamic/static handling and plugin resolution

I think, personally, that's worth it, but a separate change we can layer on top later on in a light refactor

Kudo: replied in the other pr's comment. let's keep it as a follow up once we have clear picture how to achieve it

Kudo:
### Merge activity

* **Jul 17, 3:42 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/expo/expo/47678).
* **Jul 17, 3:44 PM UTC**: [Graphite](https://app.graphite.com/github/pr/expo/expo/47678) rebased this pull request as part of a merge.
* **Jul 17, 3:44 PM UTC**: @Kudo merged this pull request with [Graphite](https://app.graphite.com/github/pr/expo/expo/47678).

PR #47638, opened 2026-07-09
[ios][dev-client] Remove swizzling

alanjhughes:
# Why

On iOS, expo-dev-launcher and expo-dev-menu swizzled and reflected into React Native internals to: suppress RN's own dev menu, gate packager server access, and render a custom RedBox. Under the New Architecture these hacks had become the *only* thing holding some behavior together (e.g. the RN-dev-menu suppression), and React Native 0.86 now exposes public APIs that replace them.

# How

- **RN dev-menu suppression** — replaced the `RCTDevMenu.show` swizzle and the `RCTDidInitializeModule` observer with the public `RCTDevMenu.devMenuEnabled` / `keyboardShortcutsEnabled` properties, re-applied per reload in `DevMenuPackagerConnectionHandler.setup()`, and re-asserting our own key commands so they win over RN's.
- **Packager server access** — use the public `RCTBundleURLProviderAllowPackagerServerAccess(...)` instead of the `guessPackagerHost` swizzle; removed `EXDevLauncherEarlySetup` and `RCTBundleURLProvider+Private.h`.
- **RedBox** — dropped the custom RedBox built on an `RCTCxxBridge` swizzle (`EXDevLauncherRedBox`, `EXDevLauncherRedBoxInterceptor`, `RCTCxxBridge+Private.h`) in favor of RN's default error UI; added `EXDevLauncherReactImports.h` for the one remaining public symbol needed.
- **Instance teardown** — annotated the `reactHost` KVC reset used to relaunch the app.

# Test Plan

- iOS dev build on device: open the dev menu (shake, Cmd+D/Ctrl+D), and confirm only our menu shows — not RN's.
- Reload repeatedly; confirm no crashes and that key commands (reload, inspector, perf monitor) keep working after a reload.
- Trigger a JS error and confirm the redbox appears.
- Confirm the Metro packager reconnects on a physical device (dev commands from the CLI reach the app).

PR #47688, opened 2026-07-10
[ios][dev-launcher] Fix compile error in Release builds

ramonclaudio:
Guards the two `RCTBundleURLProviderAllowPackagerServerAccess` calls added in #47638 with the same `#if` React Native uses to declare that function, so `expo-dev-launcher` compiles in Release again.

The function is only declared under `RCT_DEV_MENU | RCT_PACKAGER_LOADING_FUNCTIONALITY`, both off in Release, so the unguarded calls failed with `call to undeclared function` and broke the `ios-build` CI job on every PR that triggers it. The nearby `RCTDevLoadingViewSetEnabled` calls need no guard, they are declared unconditionally.

# Test Plan

`packages/expo-dev-launcher`, green via `et check-packages expo-dev-launcher` (build, typecheck, lint, format, test):

- `apps/bare-expo` builds and runs in Release on the iPhone 17 Pro simulator (Xcode 26.6), the exact `ios-build` scenario that regressed. `et native-unit-tests -p ios --packages expo-dev-launcher` passes, and a Debug build still discovers and loads dev servers.
- Reproduced the break and the fix standalone against the bare-expo pods headers: the call fails without `DEBUG` and compiles with `-DDEBUG=1`, matching Release and Debug.

# Checklist

  - Not applicable: no module plugin touched.
  - No documentation changes needed.


alanjhughes (approved):
Thank you @ramonclaudio!

tsapeta (approved):

ramonclaudio:
> Thank you @ramonclaudio!

No problem @alanjhughes!

PR #47693, opened 2026-07-10
[macos][dev-menu] Fix build with react-native-macos's older RCTDevMenu

ramonclaudio:
Guards the `RCTDevMenu.devMenuEnabled` / `keyboardShortcutsEnabled` writes added in #47638 with `#if !os(macOS)`, so `expo-dev-menu` compiles on macOS again.

react-native-macos trails RN core, and its `RCTDevMenu` has neither member (they land in core 0.83), so the writes failed with `value of type 'RCTDevMenu' has no member 'devMenuEnabled'` and broke the `test-suite-macos` CI job. `openRNDevMenu` guards only the toggle: macOS `RCTDevMenu` still has `show()` with no enabled gate, so the "Open RN dev menu" button keeps working by calling `show()` directly. iOS and tvOS use RN core and keep the writes. The guard mirrors the `// TODO(gabrieldonadel)` block already in the same file.

# Test Plan

`packages/expo-dev-menu`, green via `et check-packages expo-dev-menu` (build, typecheck, lint, format, test):

- `apps/bare-expo` builds the `ExpoMacOS-macOS` scheme against `react-native-macos@0.81.8`: fails without the guard (`has no member 'devMenuEnabled'`), succeeds with it. The exact `test-suite-macos` scenario that regressed.
- iOS unaffected: `et native-unit-tests -p ios --packages expo-dev-menu` passes. The guard excludes only macOS, and iOS and tvOS use RN core, which has the members.

# Checklist

  - Not applicable: native-only guard, no module plugin touched.
  - No documentation changes needed.


gabrieldonadel (approved):
LGTM, thanks for the fix

Issue #47703, opened 2026-07-11
Proposal: recognize react-native-windows as an out-of-tree platform (extends #46344)


FaithfulAudio:
## Summary

Proposal: extend the `tvos`/`macos` out-of-tree platform support landed in #46344 to also recognize `windows` (react-native-windows), so an Expo app can target Windows from the same Metro instance and the same tree as iOS/Android/web — no `metro.config.js` platform hacks required.

This continues the discussion in #30820 (closed by the stale bot, not resolved — @acoates-ms's original report and the RN CLI links there are exactly the mechanism this proposal wires up).

## Production motivation

We maintain a shipping Expo Router app (Facilitron FIT) that bundles **iOS, Android, web, and Windows from one Metro tree** — `expo start` / EAS Build for iOS/Android, Expo export for web, and a react-native-windows 0.83 new-architecture app served by the *same* Metro dev server for Windows. Running this way in production since July 2026.

Getting Windows working required reimplementing, in userland `metro.config.js`, the two behaviors a standalone RNW app gets for free from `@react-native/community-cli-plugin`:

1. a **platform resolver** redirecting `react-native` → `react-native-windows` for `platform === 'windows'` (`metroPlatformResolver.js`), and
2. **out-of-tree `InitializeCore`/setup-env inclusion** in `getModulesRunBeforeMainModule` (`loadMetroConfig.js`).

We extracted that userland workaround into a standalone package, [`@facilitronworks/expo-metro-windows`](https://github.com/FacilitronWorks/expo-metro-windows), which documents the four failure modes and fixes in detail. It's explicitly framed as an interim shim these upstream PRs should obsolete.

While implementing this, we found `packages/expo-modules-autolinking/src/platforms/index.ts` already has:

```ts
case 'windows':
  return 'react-native-windows';
```

— i.e. `windows` support for this mechanism was already anticipated, just not wired all the way through `@expo/config`'s `Platform` type, `@expo/cli`'s `platformBundlers`/`AUTOLINKING_PLATFORMS`, or the bare-specifier redirect in `withMetroMultiPlatform.ts`.

## Proposed PRs (draft, attached)

[...]

PR #46344, opened 2026-05-27
feat(cli,autolinking,config,metro-config): Add experimental out-of-tree `tvos` and `macos` platforms

kitten:
## Summary

Adds experimental CLI integrations for `react-native-tvos` (tvOS-only) and `react-native-macos` (untested). This auto-activates autolinking module resolution, extended for `reactNativePath` remapping, adds the appropriate autolinking entrypoints, and updates `@expo/config` with new helpers, as necessary.

The goal is to be able to launch a tvOS or macOS project without remapping the `react-native` package or Metro config manually, out of the box.

**Note:** Assumes `RCTPlatformName` is set to `"macos"` or `"tvos"` respectively.

> [!WARNING]
> This only currently adds `tvos` (meaning Apple tvOS, for `react-native-tvos`) and `macos` (for `react-native-macos`) for testing. The chosen extensions temporarily match this (i.e. `.tvos` and `.macos`)
>
> This does not yet add `run:*` commands or fix/change assumptions about the native folders. This isn't the final state and further modifications are expected. This should eventually be picked to `sdk-56` since OOTs lag behind and the assumption is that future React Native changes may make them incompatible with `sdk-57`

The main changes that make this work are:
- `expo`'s native runtime entrypoint was changed to add `react-native/Libraries/Core/InitializeCore` explicitly, which removes the need for `getModulesRunBeforeMainModule` to be accurate for this specific module
- Configs in `@expo/metro-config` don't hardcode `react-native` paths anymore but refer to canonical react-native support package paths (OOT) manually (remapping doesn't work for `getPolyfills` unless we make another change, which we could, but for now, this is the more explicit change)
- The autolinking resolver now supports remapping and can redirect `react-native` to an OOT package, such as `react-native-tvos`, and the autolinking resolver supports the two experimental platforms
- We override extension resolutions in the CLI via `config.resolver.unstable_platformExtensions` (or some default values)

## Set of changes

- Add explicit `react-native/Libraries/Core/InitializeCore` ordered import
- Alter `getPolyfills` as per support package OOT path
- Add `expo-modules-autolinking` overrides for supported OOTs per platform
- Update autolinking resolution for redirects for `react-native`
- Auto-enable autolinking resolution for OOT
- Expand `start` and `export` commands
- Add `getPlatformExtensions` to `@expo/config`
- Use `getPlatformExtensions` to override native extensions for OOT
- Add `experiments.outOfTreePlatforms` flag (required to enable `tvos`/`macos` overrides and platforms)
[...]

Issue #30820, opened 2024-08-05
Support additional platforms such as react-native-windows in expo CLI


acoates-ms:
### Summary

@keith-kurak we talked a little about doing something about this.

RN CLI supports detecting additional out of tree platforms and automatically adding them to the metro config.  The 2 most important things required for metro's config is redirecting import to `react-native` to the appropriate platform package.  And the inclusion of that platforms InitializeCore.

See https://github.com/microsoft/react-native-windows/issues/13534 for an example of a metro config that adds react-native-windows support to expo. - As well as some other comments about issues using expo with react-native-windows.

### Some relevant RN CLI links:
How RN CLI determines what platforms/packages to redirect to:
https://github.com/facebook/react-native/blob/9ee50e5db9f4672b4f7019aa9c75d75465ac3a3e/packages/community-cli-plugin/src/utils/loadMetroConfig.js#L38
metro config that adds InitializeCore from out of tree platforms to the list of initialModules:
https://github.com/facebook/react-native/blob/9ee50e5db9f4672b4f7019aa9c75d75465ac3a3e/packages/community-cli-plugin/src/utils/loadMetroConfig.js#L67
custom metro resolver that redirects to out of tree platforms:
https://github.com/facebook/react-native/blob/main/packages/community-cli-plugin/src/utils/metroPlatformResolver.js

keith-kurak:
Hi @acoates-ms ! I was trying to find a way to direct message you to continue our discussion, but struck out 😅 . Can you send me a note at keith [at] expo.dev ?

listepo:
Yes pls support windows and mac os

ianzone:
would love to have this feature.

PR #47704, opened 2026-07-11
[cli][config][metro-config] Recognize react-native-windows as an out-of-tree platform

FaithfulAudio:
# Why

Proposal/context: #47703 (continues the #30820 discussion). Extends the `tvos`/`macos` out-of-tree (OOT) platform support landed in #46344 to also recognize `windows` (react-native-windows), so an Expo app can target Windows from the same Metro instance and the same tree as iOS/Android/web.

**Production story:** we run a shipping Expo Router app (Facilitron FIT) bundling iOS, Android, web, and Windows from one Metro tree — `expo start`/EAS for iOS/Android, Expo export for web, and a react-native-windows 0.83 new-arch app served by the same Metro dev server for Windows, in production since July 2026. Getting there required reimplementing in userland `metro.config.js` two behaviors a standalone RNW app gets for free from `@react-native/community-cli-plugin`: the `react-native` → `react-native-windows` platform redirect (`metroPlatformResolver.js`) and out-of-tree `InitializeCore`/`setup-env` inclusion (`loadMetroConfig.js`). We packaged that workaround as [`@facilitronworks/expo-metro-windows`](https://github.com/FacilitronWorks/expo-metro-windows) — its README documents all four failure modes in detail and is explicitly framed as an interim shim this PR should obsolete.

While building this we noticed `expo-modules-autolinking/src/platforms/index.ts` already has:
```ts
case 'windows':
  return 'react-native-windows';
```
`windows` was already anticipated in `getSupportPackageForPlatform`, just not threaded through `@expo/config`'s `Platform` type, `@expo/cli`'s `platformBundlers`/`AUTOLINKING_PLATFORMS`, or a bare-specifier redirect. This PR closes that loop.

# How

Mirrors #46344's `tvos`/`macos` pattern file-for-file, plus the one extra piece `windows` structurally needs:

- **`@expo/config`**
  - `Platform` type: add `'windows'`.
  - `getSupportedPlatforms`: detect `windows` via `resolveFrom(projectRoot, 'react-native-windows/package.json')`, gated behind the existing `experiments.outOfTreePlatforms` flag (same as `tvos`/`macos`).
  - `getPlatformExtensions`: add `windows: ['windows', 'native']` (no `ios`/`apple` fallback, unlike `tvos`/`macos`).

- **`@expo/metro-config`**
  - `getReactNativeHostPackage`: add the `windows` → `react-native-windows` case (used by `getPolyfills`, which already receives `platform`).
  - Default `resolver.platforms`: add `'windows'` (inert until a project actually targets it, same as `tvos`/`macos` today).
  - `unstable_conditionsByPlatform.windows = ['react-native']`.

- **`@expo/cli`**
  - `platformBundlers` / `AUTOLINKING_PLATFORMS`: add `windows: 'metro'`.
  - `withMetroMultiPlatform.ts`: add a `react-native` → `react-native-windows` bare-specifier + submodule resolver (`requestReactNativeWindowsRedirect`), gated on `react-native-windows` actually being installed (`hasReactNativeWindows`, computed once via `resolveFrom`).

**Why `windows` needs the redirect but `tvos`/`macos` don't:** `react-native-tvos`/`react-native-macos` are conventionally installed by *aliasing the `react-native` package name itself* (e.g. `"react-native": "npm:react-native-macos@..."` in `package.json`), so `resolveFrom(projectRoot, 'react-native/package.json')` already resolves into the OOT package — Metro's default resolution just works, and the existing `getReactNativeHostPackage`-based path-suffix rewrites in `requestPostRewrites` (InitializeCore/HMRClient/LogBox swaps) already fire correctly against the real (symlink-resolved, pnpm-realpath) package name. `react-native-windows` is conventionally installed as an *additional* dependency alongside a real, un-aliased `react-native` (the standard RNW docs/templates), so nothing redirects the bare specifier without this change — `import { Platform } from 'react-native'` would resolve into RN core, whose backwards-compat `Libraries/Utilities/Platform.js` does `import Platform from './Platform'`, which — with no `Platform.windows.js` in RN core — is a self-referencing circular import, giving `undefined` and `Cannot read property 'OS' of undefined` at boot.

The new resolver is a re-implementation, for Expo's resolver pipeline, of `@react-native/community-cli-plugin`'s [`reactNativePlatformResolver`](https://github.com/facebook/react-native/blob/main/packages/community-cli-plugin/src/utils/metroPlatformResolver.js) — falls back to un-redirected resolution (via the existing `getOptionalResolver` helper already used elsewhere in this file) for the rare submodule `react-native-windows` doesn't override, rather than hard-failing.

Once the bare-specifier redirect lands, several already-windows-aware code paths in `withMetroMultiPlatform.ts` that were previously unreachable for `windows` (because nothing ever resolved a `react-native-windows`-rooted path) start working: `getReactNativeHostPathPattern`, and the `hostPackage`-suffix swaps in `requestPostRewrites` (HMRClient, LogBox).

Everything is feature-detected/gated — zero behavior change for any project without `react-native-windows` installed, and `tvos`/`macos`/`ios`/`android`/`web` code paths are untouched (only additive `case`/map entries).

# Test Plan

- Added `getPlatformExtensions('windows', ...)` unit test in `packages/@expo/config/src/paths/__tests__/extensions-test.ts` (mirrors the existing `tvos` test).
- The rest of the diff is additive `case`/array/map entries following the exact shape of existing `tvos`/`macos` entries — no new logic branches beyond `requestReactNativeWindowsRedirect`, which is gated behind `platform === 'windows' && hasReactNativeWindows` and therefore untestable without a `react-native-windows`-having fixture project; happy to add an integration test if maintainers can point me at the right harness (`apps/bare-expo`'s OOT platform tests, if any exist, or a synthetic resolver-level test alongside the existing tests in `withMetroMultiPlatform`'s test suite).
- **Not yet run against our production app or Expo's CI** — this is a fresh reimplementation against current `main` following the #46344 architecture, not a copy-paste of our userland `metro.config.js` workaround (which predates and does not use the OOT-platform machinery this PR extends). Flagging honestly: needs your CI, and ideally a follow-up smoke test swapping `@facilitronworks/expo-metro-windows` for this PR in our app to confirm parity before merge.

# Checklist

- [x] I added a changelog entry to each touched package (`@expo/config`, `@expo/cli`, `@expo/metro-config`) — will fill in the real PR number/link once this PR is open.
- [ ] This diff will work correctly for `npx expo prebuild` & EAS Build — untested for Windows specifically since there's no EAS Build Windows target; no impact expected on existing platforms.
- [x] Conforms with the Documentation Writing Style Guide (n/a — no docs/ changes).

### DIFF
--- a/packages/@expo/cli/CHANGELOG.md
+++ b/packages/@expo/cli/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ### 🎉 New features
 
+- Add `windows` (react-native-windows) to the experimental out-of-tree platform resolver: registers the Metro platform, autolinking, and a `react-native` -> `react-native-windows` bare-specifier redirect gated on the package being installed ([#TBD](https://github.com/expo/expo/pull/TBD) by [@FaithfulAudio](https://github.com/FaithfulAudio))
 - Add option to specify targets to use with inline modules ([#46698](https://github.com/expo/expo/pull/46698) by [@HubertBer](https://github.com/HubertBer))
 - Support Bundler-managed CocoaPods installations ([#43605](https://github.com/expo/expo/pull/43605) by [@tiwari91](https://github.com/tiwari91), [@kitten](https://github.com/kitten))
 - Support Device Hub as Simulator replacement for Xcode 27+ ([#46757](https://github.com/expo/expo/pull/46757) by [@byCedric](https://github.com/byCedric), [@GersonRocha9](https://github.com/GersonRocha9))
--- a/packages/@expo/cli/CHANGELOG.md
+++ b/packages/@expo/cli/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ### 🎉 New features
 
-- Add `windows` (react-native-windows) to the experimental out-of-tree platform resolver: registers the Metro platform, autolinking, and a `react-native` -> `react-native-windows` bare-specifier redirect gated on the package being installed ([#TBD](https://github.com/expo/expo/pull/TBD) by [@FaithfulAudio](https://github.com/FaithfulAudio))
+- Add `windows` (react-native-windows) to the experimental out-of-tree platform resolver: registers the Metro platform, autolinking, and a `react-native` -> `react-native-windows` bare-specifier redirect gated on the package being installed ([#47704](https://github.com/expo/expo/pull/47704) by [@FaithfulAudio](https://github.com/FaithfulAudio))
 - Add option to specify targets to use with inline modules ([#46698](https://github.com/expo/expo/pull/46698) by [@HubertBer](https://github.com/HubertBer))
 - Support Bundler-managed CocoaPods installations ([#43605](https://github.com/expo/expo/pull/43605) by [@tiwari91](https://github.com/tiwari91), [@kitten](https://github.com/kitten))
 - Support Device Hub as Simulator replacement for Xcode 27+ ([#46757](https://github.com/expo/expo/pull/46757) by [@byCedric](https://github.com/byCedric), [@GersonRocha9](https://github.com/GersonRocha9))
--- a/packages/@expo/cli/src/start/server/metro/createExpoAutolinkingResolver.ts
+++ b/packages/@expo/cli/src/start/server/metro/createExpoAutolinkingResolver.ts
@@ -39,7 +39,14 @@ const KNOWN_STICKY_DEPENDENCIES = [
   '@react-navigation/native',
 ];
 
-const AUTOLINKING_PLATFORMS: readonly Platform[] = ['android', 'ios', 'web', 'tvos', 'macos'];
+const AUTOLINKING_PLATFORMS: readonly Platform[] = [
+  'android',
+  'ios',
+  'web',
+  'tvos',
+  'macos',
+  'windows',
+];
 
 export type AutolinkingPlatform = Platform;
 
--- a/packages/@expo/cli/src/start/server/metro/withMetroMultiPlatform.ts
+++ b/packages/@expo/cli/src/start/server/metro/withMetroMultiPlatform.ts
@@ -248,6 +248,8 @@ export function getNodejsExtensions(srcExts: readonly string[]): string[] {
  * - Alias `react-native` to `react-native-web` on web.
  * - Redirect `react-native-web/dist/modules/AssetRegistry/index.js` to `@react-native/assets/registry.js` on web.
  * - Add support for `tsconfig.json`/`jsconfig.json` aliases via `compilerOptions.paths`.
+ * - Alias `react-native` (and its submodules) to `react-native-windows` when targeting
+ *   the Windows out-of-tree platform and `react-native-windows` is installed.
  */
 export function withExtendedResolver(
   config: ConfigT,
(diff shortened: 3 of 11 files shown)


FaithfulAudio:
A gentle status update with some evidence that this PR's use case is real, plus a direct question about what would make it acceptable.

**What exists today, running against this change:**

- A production Windows app in the Microsoft Store — Facilitron FIT, a React Native **new-architecture** app on react-native-windows 0.83.2, built with Expo SDK 55, expo-router, and Metro. This PR (plus #47705) is the difference between that working out of the box and the pile of resolver workarounds we currently vendor.
- Six published packages that mirror Expo/community module APIs on Windows (`@facilitronworks/react-native-windows-camera`, `-image-picker`, `-sync-kv`, `-modal`, `expo-metro-windows`, `react-native-webview-windows`) — i.e. the module side of the platform is being carried by the community, not requested of Expo.
- An independent third-party consumer testing those packages on RNW 0.84.1 Fabric and reporting results upstream.

**What this PR is not asking for:** Expo support for Windows. No CI burden, no module work, no docs commitment — the platform-support policy is clear and we're not contesting it. The 88-line change only makes the CLI/config/metro-config *recognize* `react-native-windows` as an out-of-tree platform the way the macOS/tvOS packages are recognized, so that people who do carry the platform don't have to patch `@expo/cli` to use Expo.

**The question:** is there a shape of this change you'd accept? If the current diff is wrong in approach — an allowlist that shouldn't grow, a config-driven escape hatch you'd prefer, tests you'd want, or a reason recognition itself is off the table — I'd genuinely rather hear that and re-cut it than have it sit. Happy to split, gate behind explicit app config, or add coverage as needed.

(CI here is green; the PR is mergeable and has no conflicts. #47705 is the runtime companion — a JS polyfill fallback in `expo-modules-core` when no native module exists — and stands alone if this one's approach needs rework.)

kitten:
This requires internal alignment at the foundation. react-native-tvos is an exception since we have an internal contributor. react-native-macos is an exception because we have internal expertise and a line to the maintainer. We won't yet be able to accept this until the React Foundation's RN Platform team aligns on this. However, at that point, it'd be better for us to align as per a spec, since the underlying code is marked experimental and just meant as a proof of concept for now.

As such, we currently can't accept PRs with proposals or requests for changes for this feature yet. Sorry!

As per the issue, I'll also close that and this PR for now until we have a proposal from the React Foundation Platforms team to work against, so this doesn't run away or ahead of them. Since that's an effort that's easier tracked on the foundation level, from our end it's easier to close this for now unfortunately.

FaithfulAudio:
Thanks for the straight answer — the foundation-alignment reasoning makes sense, and aligning to a spec beats growing an allowlist case by case. Not contesting the close.

Three small things:

1. Is there somewhere to follow the RN Platform team's work on this? A repo, working group, or RFC tracker — I'd rather watch that than periodically re-ping here. Any rough sense of timing (this quarter, next, genuinely unknown) would help us plan; not asking for a commitment, just a horizon.

2. Can we be a data point when the spec work starts? We're running a shipping new-arch RNW 0.83.2 app on Expo SDK 55 in the Microsoft Store, six community modules mirroring Expo/community APIs, and an outside consumer testing on 0.84.1 Fabric. Whoever writes that spec will want real-world constraints from someone who hit them — happy to supply ours, review a draft, or just answer questions. Point me at where and I'll show up.

3. #47705 is orthogonal to this. The expo-modules-core JS polyfill fallback isn't OOT-platform machinery — it's a defensive fallback when no native ExpoModulesCore exists, useful independent of whether windows is ever recognized. Could it be evaluated on its own merits rather than held behind the same gate?

On the cost, briefly: with recognition off the table we keep patching @expo/cli in userland, which re-breaks on every SDK bump. That's ours to carry and we're not asking Expo to carry it — flagging it only because it's exactly the tax a spec eventually removes, and the longer the gap the more projects independently reinvent the same shim.

douglowder:
> 1. Is there somewhere to follow the RN Platform team's work on this? A repo, working group, or RFC tracker — I'd rather watch that than periodically re-ping here. Any rough sense of timing (this quarter, next, genuinely unknown) would help us plan; not asking for a commitment, just a horizon.

@FaithfulAudio the RFC discussion is here: https://github.com/react-native-community/discussions-and-proposals/discussions/1009 . Timing is uncertain at this point. I've been attempting to plan out a refactor of expo/expo code to make it easier to add new platforms and have them properly handled by various systems (Metro, prebuild, config-plugins), but this is very complex and there is no definite timeline yet.

PR #47705, opened 2026-07-11
[core] Fall back to the JS polyfill when no native ExpoModulesCore exists

FaithfulAudio:
# Why

Proposal/context: #47703. Companion to the `@expo/metro-config`/`@expo/cli`/`@expo/config` out-of-tree platform PR for `windows` — that PR gets Metro to bundle for a platform with no native React Native host modification; this PR is the runtime-side half: making `expo-modules-core` degrade gracefully when a platform has no native `ExpoModulesCore` module, instead of crashing at import time.

**Production story:** in our shipping Expo Router app bundling iOS/Android/web/Windows from one tree (see the companion PR for details), the very first thing that broke on react-native-windows — before any individual native module mattered — was importing `expo-modules-core` itself (transitively, via `expo-router`):

```
TypeError: Cannot read property 'EventEmitter' of undefined
```

Root cause: `EventEmitter.ts`, `NativeModule.ts`, `SharedObject.ts`, and `SharedRef.ts` each do `export const X = globalThis.expo.X` at **module scope**. `globalThis.expo` is installed by native code on iOS/Android and by a JS polyfill on web (`polyfill/index.web.ts`), but on a platform with a JS-only React Native host and no native `ExpoModulesCore` port (react-native-windows today; potentially others later), `ensureNativeModulesAreInstalled.native.ts`'s `TurboModuleRegistry.get('ExpoModulesCore')` returns `null`, and the optional-chained `?.installModules()` silently no-ops — so `globalThis.expo` is simply never set, and the very first `expo-modules-core` import in the app crashes before a single line of app code runs.

Our interim workaround (documented in [`@facilitronworks/expo-metro-windows`](https://github.com/FacilitronWorks/expo-metro-windows)'s README) is a manual side-effect import of `expo-modules-core/src/polyfill/dangerous-internal`'s `installExpoGlobalPolyfill()` as the very first line of our Windows entry point (`index.windows.js`), before anything else imports `expo-modules-core`/`expo-router`. It works, but it's fragile (import-order-dependent, easy to break by adding an earlier import) and every consuming app would need to reinvent it. This PR moves the fallback to the one place that actually needs to own it.

# How

One conditional branch in `ensureNativeModulesAreInstalled.native.ts`: when `TurboModuleRegistry.get('ExpoModulesCore')` returns `null` (no native module registered), call `installExpoGlobalPolyfill()` — the same pure-JS polyfill `web` already uses (exported from `./polyfill/dangerous-internal` specifically for this kind of call site; `CoreModule.ts`'s `EventEmitter`/`NativeModule`/`SharedObject`/`SharedRef` are plain JS classes with no DOM dependency).

iOS/Android are unaffected: the native module is always found there today, so the new `else` branch is unreachable — this is a pure addition, not a behavior change, on every platform that currently works.

Once `globalThis.expo` exists, the *existing* `requireNativeModule`/`requireOptionalNativeModule` machinery already degrades individual unsupported native modules correctly (`requireOptionalNativeModule` → `null`, `requireNativeModule` → a clear `Cannot find native module 'X'` Error) — that path is completely untouched by this PR. This PR only fixes the crash that currently happens *before* any of that machinery gets a chance to run.

# Test Plan

Added `packages/expo-modules-core/src/__tests__/ensureNativeModulesAreInstalled-test.native.ts` (no prior test coverage existed for this function). Mocks `TurboModuleRegistry.get`, following the existing `jest.mock('react-native', ...)` pattern from `NativeViewManagerAdapter-test.native.tsx`:
- falls back to the polyfill and populates `globalThis.expo.{EventEmitter,NativeModule,SharedObject}` when no native module is registered
- `new globalThis.expo.EventEmitter()` doesn't throw afterward (this is the exact line that crashed before the fix)
- still prefers a registered native module's `installModules()` and does **not** run the polyfill when one exists (guards the iOS/Android path)

**Honest status:** this is a fresh implementation against current `main`, written for this PR — not a copy-paste of our app's `dangerous-internal` import workaround (which sidesteps this function entirely rather than fixing it). I have not been able to run `pnpm test` in this environment (no local install), so the new test file needs your CI to confirm it passes as written; happy to iterate on the jest-expo mock setup if it doesn't match your harness exactly.

# Checklist

- [x] Added a changelog entry to `packages/expo-modules-core/CHANGELOG.md` — will fill in the real PR number/link once this PR is open.
- [x] This diff has no `npx expo prebuild`/EAS Build surface (pure JS runtime change, no plugin/config).
- [x] Conforms with the Documentation Writing Style Guide (n/a — no docs/ changes).


packages/expo-modules-core/src/__tests__/ensureNativeModulesAreInstalled-test.native.ts
```diff
@@ -0,0 +1,65 @@
...
+
+describe('ensureNativeModulesAreInstalled (native, no `ExpoModulesCore` turbo module)', () => {
+  const originalExpo = globalThis.expo;
+
+  beforeEach(() => {
+    jest.resetModules();
+    // @ts-expect-error - deliberately tearing down the global to simulate a runtime
+    // that never installed it (e.g. no native `ExpoModulesCore` on this platform).
+    delete globalThis.expo;
+    (TurboModuleRegistry.get as jest.Mock).mockReturnValue(null);
```
Kudo: **Medium:** `jest.resetModules()` in `beforeEach` will re-run the `react-native` mock factory and create a **new** `TurboModuleRegistry.get` `jest.fn()`. The `TurboModuleRegistry` imported at module scope stays bound to the *previous* mock.

So `(TurboModuleRegistry.get as jest.Mock).mockReturnValue(...)` here (and in the prefer-native test below) does not configure the `get` that `ensureNativeModulesAreInstalled.native` actually calls after `require(...)`.

- Null-fallback tests may pass accidentally because a fresh `jest.fn()` returns `undefined` (still falsy → else branch).
- The prefer-native test is likely to fail or not exercise `installModules` at all.

**Fix:** close over a stable `const mockGet = jest.fn()` in the factory, or re-`import`/`require` `TurboModuleRegistry` after `resetModules`. Also spy/`jest.mock` `installExpoGlobalPolyfill` so the prefer-native case asserts the polyfill was **not** invoked.

packages/expo-modules-core/src/ensureNativeModulesAreInstalled.native.ts
```diff
@@ -13,9 +15,27 @@ export function ensureNativeModulesAreInstalled(): void {
...
+      // installation problem. Fall back to the same pure-JS polyfill `web`
+      // already uses for the same reason (see `polyfill/index.web.ts`), so the
+      // `EventEmitter`/`NativeModule`/`SharedObject` base classes this package
+      // itself exports exist. Without this, `globalThis.expo` is left
+      // `undefined` and importing this package crashes immediately with
+      // `Cannot read properties of undefined (reading 'EventEmitter')` --
+      // instead of the individual `requireNativeModule()` call for a specific
+      // module throwing/returning `null`, which callers can already handle.
+      installExpoGlobalPolyfill();
+    }
```
Kudo: **Medium:** Static import always wires `dangerous-internal` (and thus `uuid/index.web` + web-shaped `reloadAppAsync: () => window.location.reload()`) into the native ensure module graph, even on iOS/Android where this `else` is unreachable.

Prefer lazy-loading inside the branch so shipping native bundles do not pay for the web polyfill:

```ts
} else {
  // eslint-disable-next-line @typescript-eslint/no-var-requires
  require('./polyfill/dangerous-internal').installExpoGlobalPolyfill();
}
```

Also worth a clear `console.warn` here: on iOS/Android a missing `ExpoModulesCore` turbo module is an install bug; silently installing the JS polyfill changes the failure mode from import-time crash to later “Cannot find native module”.

Kudo:
@expo-bot verify --fable

Issue #42225, opened 2026-01-15
[SDK55 Canary] `@expo/ui`: swift-ui Menu with react node label disappears after opening


e-simpson:
### Minimal reproducible example

See code below

### Steps to reproduce

1. Update to latest expo sdk v55 beta packages
```JSON
    "expo": "55.0.0-canary-20260114-d8e19f5",
     "@expo/ui": "^0.2.0-canary-20260114-d8e19f5",
```

2. Use swift-ui Menu with a react node label
```JSX
import { Host, Menu, Button } from '@expo/ui/swift-ui';
import { View } from 'react-native';

export default function SimpleMenuExample() {
  return (
    <Host matchContents>
      <Menu 
	  	label={<View style={{ width: 80, height: 30, backgroundColor: 'red' }}/>}
	  >
        <Button label="Option 1" onPress={() => console.log('Option 1')} />
        <Button label="Option 2" onPress={() => console.log('Option 2')} />
        <Button label="Option 3" onPress={() => console.log('Option 3')} />
      </Menu>
    </Host>
  );
}
```

3. Tap on the Menu element, to see the react node does not come back

### Environment

```text
expo-env-info 2.0.8 environment info:
    System:
      OS: macOS 26.1
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 22.16.0 - ~/.nvm/versions/node/v22.16.0/bin/node
      Yarn: 1.22.19 - /opt/homebrew/bin/yarn
      npm: 10.9.2 - ~/.nvm/versions/node/v22.16.0/bin/npm
    Managers:
      CocoaPods: 1.16.2 - /opt/homebrew/bin/pod
    SDKs:
      iOS SDK:
        Platforms: DriverKit 25.1, iOS 26.1, macOS 26.1, tvOS 26.1, visionOS 26.1, watchOS 26.1
    IDEs:
      Android Studio: 2024.3 AI-243.26053.27.2432.13536105
      Xcode: 26.1/17B55 - /usr/bin/xcodebuild
    npmPackages:
      expo: 55.0.0-canary-20260114-d8e19f5 => 55.0.0-canary-20260105-6b962e6 
      expo-router: ~7.0.0-canary-20260114-d8e19f5 => 7.0.0-canary-20260114-d8e19f5 
      expo-updates: ~29.1.0-canary-20260114-d8e19f5 => 29.1.0-canary-20260114-d8e19f5 
      react: 19.2.3 => 19.2.3 
[...]

Issue #40604, opened 2025-10-24
[ios][ui] `@expo/ui` Context Menu causes glitchy UI when `ScrollView` is present


bnussman:
## Overview

Using an `@expo/ui` iOS Context Menu causes glitchy UI when rendering React Native components as the Trigger and when a `ScrollView` is conditionally rendered.

## Minimal reproducible example

https://github.com/bnussman/expo-ui-glitchy-ui

## Steps to Reproduce

- Open the minimal reproducible example (I think you'll need a development build with `@expo/ui` installed)
- Click the "Go to other page" button
- Click the "Go home" button
- Repeat those last two steps
- Observe the UI glitch

https://github.com/user-attachments/assets/364fc247-77b5-4430-bccf-656dbea61e04

## Other reports of this bug

* https://x.com/bnussman/status/1981161305173348497
* https://x.com/garyautomates/status/1980819632832540790
* https://github.com/expo/expo/issues/36147#issuecomment-2805374513
* https://github.com/expo/expo/issues/36147#issuecomment-3376858967
* https://github.com/expo/expo/issues/36147#issuecomment-3429581445

### Environment

```text
expo-env-info 2.0.7 environment info:
    System:
      OS: macOS 26.0.1
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 24.5.0 - ~/.local/state/fnm_multishells/61305_1761324862947/bin/node
      npm: 11.5.1 - ~/.local/state/fnm_multishells/61305_1761324862947/bin/npm
    SDKs:
      iOS SDK:
        Platforms: DriverKit 25.0, iOS 26.0, macOS 26.0, tvOS 26.0, visionOS 26.0, watchOS 26.0
    IDEs:
      Android Studio: 2025.1 AI-251.26094.121.2513.14007798
      Xcode: 26.0/17A324 - /usr/bin/xcodebuild
    npmPackages:
      expo: ~54.0.20 => 54.0.20 
      expo-router: ~6.0.13 => 6.0.13 
      react: 19.1.0 => 19.1.0 
      react-dom: 19.1.0 => 19.1.0 
      react-native: 0.81.5 => 0.81.5 
      react-native-web: ~0.21.0 => 0.21.2 
    npmGlobalPackages:
      eas-cli: 16.20.4
    Expo Workflow: managed
```

### Expo Doctor Diagnostics

```text
17/17 checks passed. No issues detected!
```

intergalacticspacehighway:
[...]

PR #47707, opened 2026-07-11
[core][ios] Isolate SwiftUI-hosted React views behind a disposable container

### LINKED ISSUE #47706: [iOS] SwiftUI-hosted React views (ExpoSwiftUI.UIViewHost): leaked autoresizingMask/frame mutations corrupt Fabric rendering after unmount (zero-height/invisible views)

Complete single-file repro inline below (Steps to reproduce) — it's a fresh `create-expo-app` blank-typescript template plus `@expo/ui`, with `App.tsx` replaced. Happy to push it as a standalone repo on request.

### Summary

`ExpoSwiftUI.UIViewHost` hands SwiftUI the **React-managed UIView directly** (`makeUIView` returns `view`). SwiftUI's platform-view hosting mutates that view — it sets `autoresizingMask = [.flexibleWidth, .flexibleHeight]`, and `Menu`/`ContextMenu` additionally mutate frame/visibility **asynchronously, after the surrounding SwiftUI hierarchy has been torn down**.

#40604 mitigated one instance of this class by restoring `autoresizingMask` in `dismantleUIView` — but **SwiftUI does not reliably call `dismantleUIView` on wholesale teardown** (e.g. when the hosting view is deallocated because a React subtree unmounted), and the deferred `Menu`/`ContextMenu` mutations can land *after* Fabric has unmounted (or recycled) the view.

The result on the New Architecture: React views end up with leaked `autoresize = W+H`, so the next time their superview resizes, UIKit's autoresizing math mangles the Fabric-assigned frame — views collapse to **zero height**, balloon, or render **invisible while still occupying layout space**. Because the mutations can arrive after the view has been recycled and remounted elsewhere, arbitrary unrelated components inherit the corruption — resetting state in `prepareForRecycle` does **not** fix it (verified).

`lldb` view-hierarchy dumps of the blank state show plain `RCTViewComponentView`s (not SwiftUI-hosted at that moment) with Fabric-correct positions but corrupted UIKit state:

```
<RCTViewComponentView: 0x11c65e700; frame = (16 57; 370 0); clipsToBounds = YES; autoresize = W+H; ...>
<RCTViewComponentView: 0x13c7c0300; frame = (16 131; 370 0); clipsToBounds = YES; ...>
<RCTViewComponentView: 0x11c65f600; frame = (16 205; 370 0); clipsToBounds = YES; autoresize = W+H; ...>
```

`frame = (…; 370 0)` — width correct, **height zero** — with `autoresize = W+H` that no React code ever set. The children inside are all present with correct frames; the container has a zeroed height and clips them.

**Suggested fix** (PR incoming): `makeUIView` should return a **disposable isolation container** holding the React view as a subview, so SwiftUI's mutations (including deferred ones) only ever reach the throwaway container. Two details: the React child must keep its Yoga frame at `makeUIView` time (the `matchContents` sizing path KVO-observes the React child's bounds — zeroing it to the empty container deadlocks that feedback loop at zero), and empty container bounds must not propagate during layout (transient zero sizes during teardown). This fully resolved the corruption in our production app.

Related: #40604, #42225 (Menu label content becoming invisible — consistent with the same leaked-visibility mechanism), #44579 (Host/Yoga async sizing; adjacent but distinct).

### Steps to reproduce

1. `npx create-expo-app@latest menuview-repro --template blank-typescript`
2. `npx expo install @expo/ui expo-build-properties`
3. Replace `App.tsx` with the file below.
4. `npx expo run:ios` (dev client — `MenuView` is not available in Expo Go).
5. Watch it auto-page every 2 seconds (each page remounts the keyed subtree, like paging between days in a planner). Within a handful of pages, rows render invisible while still occupying layout space (worst on "menu-only" pages, where the whole list can go blank) and/or collapse to zero height / balloon and overlap.

```tsx
import { MenuView } from "@expo/ui/community/menu";
import { useEffect, useState } from "react";
import { ScrollView, StyleSheet, Text, View } from "react-native";

const ACTIONS = [
  { id: "a", title: "Option A" },
  { id: "b", title: "Option B" },
];

function MenuTrigger() {
  return (
    <MenuView actions={ACTIONS} style={styles.menu}>
      <View style={styles.trigger}>
        <Text>○</Text>
      </View>
    </MenuView>
  );
}

function Row({ label, contextMenu }: { label: string; contextMenu?: boolean }) {
  const row = (
    <View style={styles.row}>
      <MenuTrigger />
      <Text style={styles.label}>{label}</Text>
    </View>
  );
  if (!contextMenu) return row;
  return (
    <MenuView actions={ACTIONS} shouldOpenOnLongPress style={styles.rowMenu}>
      {row}
    </MenuView>
  );
}

  const [page, setPage] = useState(0);

  useEffect(() => {
    const t = setInterval(() => setPage((p) => p + 1), 2000);
    return () => clearInterval(t);
  }, []);

  // Alternate: pages with long-press ContextMenu-wrapped rows, and pages
  // where every row's only fixed-height child is the MenuView trigger.
  const menuOnly = page % 2 === 1;

  return (
    <View key={page} style={styles.container}>
      <Text style={styles.header}>
        page {page} ({menuOnly ? "menu-only rows" : "mixed rows"})
      </Text>
      <ScrollView contentContainerStyle={styles.list}>
        <Text style={styles.marker}>START MARKER</Text>
        {Array.from({ length: 6 }, (_, i) => (
          <Row key={i} label={`Row ${i}`} contextMenu={!menuOnly && i < 3} />
        ))}
        <Text style={styles.marker}>END MARKER</Text>
      </ScrollView>
    </View>
  );
}

  container: { flex: 1, paddingTop: 80 },
  header: { fontWeight: "700", textAlign: "center", padding: 8 },
  list: { gap: 8, padding: 16 },
  marker: { backgroundColor: "#c00", color: "#fff", padding: 8 },
  row: {
    alignItems: "center",
    backgroundColor: "#eee",
    borderRadius: 12,
    flexDirection: "row",
    gap: 8,
    padding: 16,
  },
  rowMenu: { alignSelf: "stretch" },
  menu: { height: 32, width: 32 },
  trigger: {
    alignItems: "center",
    borderColor: "#999",
    borderRadius: 999,
    borderWidth: 1,
    height: 32,
    justifyContent: "center",
    width: 32,
  },
  label: { flex: 1 },
});
```

Note for anyone testing a source-level fix on SDK 57: set `expo-build-properties` → `ios.usePrecompiledModules: false`, since `expo-modules-core` otherwise links as a prebuilt XCFramework and silently ignores source patches.

### Environment

```
    System:
      OS: macOS 26.5.1
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 24.3.0 - ~/.nvm/versions/node/v24.3.0/bin/node
      npm: 11.10.1 - ~/.nvm/versions/node/v24.3.0/bin/npm
    Managers:
    SDKs:
      iOS SDK:
    IDEs:
    npmPackages:
      expo: ^57.0.4 => 57.0.4
      react: 19.2.3 => 19.2.3
      react-dom: 19.2.3 => 19.2.3
    npmGlobalPackages:
      eas-cli: 20.0.0
    Expo Workflow: bare
```

Also reproduced with `expo-modules-core@57.0.3` and `@expo/ui@57.0.4` on the iOS 26.0 simulator and an iOS 26 device.

### Expo Doctor Diagnostics

```
```

### LINKED ISSUE #47934: [iOS][@expo/ui] Native-stack `replace` while a SwiftUI `BottomSheet` is presented: uncoordinated presentation teardown corrupts the incoming screen's native geometry (invisible to the Fabric shadow tree)

https://github.com/Xelson/expo-sheet-replace-repro

Fully autonomous — no taps needed: Screen A presents an `@expo/ui/swift-ui` `BottomSheet` at t=1.5s, then fires `router.replace('/form')` at t=3s **without closing the sheet** (exactly what happens in a real app when the auth/navigation gate reacts to a server event while a sheet-hosted flow is still on screen). A `CLOSE_SHEET_FIRST` constant flips it into the clean control case. Screenshots of every phase are committed under `docs/`.

iOS

Development build (`npx expo run:ios`), iPhone 17 Pro simulator, iOS 26.5, Xcode 26.5 (17F42), macOS 26.4.

### Environment

- expo 57.0.7 (SDK 57), @expo/ui 57.0.7, expo-router 57.0.7
- react-native 0.86.0, react 19.2.3, react-native-screens 4.25.2
- New Architecture (SDK 57 default)
- (`npm install --legacy-peer-deps` needed: @expo/ui's `vaul`/`react-dom` optional peers conflict with react 19.2.3)

### Summary

When a react-native-screens native-stack `replace` removes the screen that hosts a still-presented `@expo/ui/swift-ui` `BottomSheet`, UIKit tears the live sheet presentation apart without coordination. Two artifacts on the INCOMING screen:

1. **Uncoordinated presentation teardown.** For several frames the incoming screen renders *inside the sheet presentation's container* — rounded bottom corners with the presentation's gray backdrop layer visible behind it (`docs/03-replace-transition-b-in-sheet-container.png`). In our production app this shows as a gray rectangle flashing over the incoming screen just below the status bar.

2. **Persistent native frame corruption.** The incoming screen's round 94×94 `Pressable` is actually rendered as a **94×~209pt capsule**, covering the `TextInput` below it, and it stays corrupted 12+ seconds later (`docs/04`, `docs/05`). Reproduced identically on 3 consecutive runs.

The nasty part: **the corruption is invisible to JS.** The repro renders live instrumentation on screen — `onLayout` (Yoga) and a 500ms `measureInWindow` poll. Both report `94x94 @ (154,216)` and the "yoga/native agree" label stays green *while the pixels clearly show a 94×209 capsule*: on Fabric, `measureInWindow` answers from the shadow tree, so the shadow tree and the actual UIKit layer disagree with no programmatic way to detect it.

Relationship to #47708 / #47713: that fix (shipped in 57.0.7, thanks!) detaches the stale `RCTSurfaceTouchHandler` on host disappear and resolved the swallowed-touches half of this failure mode in our production app. The uncoordinated presentation teardown and the frame corruption reported here remain on 57.0.7.

### Steps to reproduce

```sh
git clone https://github.com/Xelson/expo-sheet-replace-repro
cd expo-sheet-replace-repro
npm install --legacy-peer-deps
npx expo run:ios
# just watch: Screen A -> sheet slides up -> replace -> Screen B
```

Control case: set `CLOSE_SHEET_FIRST = true` in `app/index.tsx` (dismiss + 700ms settle before the replace) → everything is clean (`docs/06`).

### Expected behavior

`replace` while a sheet is presented coordinates with the presentation (dismisses it) before the outgoing screen's view controller is torn down. The incoming screen renders with no leftover presentation layers and correct native geometry.

### Actual behavior

See summary; screenshots in `docs/` of the repro repo:

- `03-replace-transition-b-in-sheet-container.png` — incoming screen rendered inside the leftover presentation container
- `04-screen-b-frame-corrupted.png` — 94×94 Pressable drawn as a 94×209 capsule while on-screen instrumentation reports `yoga: 94x94`, `native: 94x94`, "yoga/native agree"
- `05-screen-b-still-corrupted-12s-later.png` — not transient
- `06-control-close-sheet-first-clean.png` — control case clean

### Workarounds we verified in a production app

- **Anchor the `SwiftUI.Host` OUTSIDE the screens** — render it at the navigator layout level as a sibling of the `<Stack>`, retain its content across the route unmatch, and drive only `isPresented`. The presenting view controller then survives the replace and the sheet plays its normal dismissal animation over the incoming screen. Fully fixes both artifacts with no delays.
- Alternatively, dismiss the sheet and wait for the dismissal animation to settle (~700ms) before navigating. Merely *initiating* the dismissal before the replace is NOT enough — the animation must finish.

Removing the dim (`presentationBackgroundInteraction('enabled')`) does NOT help — the artifact is the presentation container itself, not the dimming view.

cvburgess:
# Why

Fixes #47706 (related: #40604, #42225).

`ExpoSwiftUI.UIViewHost.makeUIView` returns the React-managed UIView directly, so SwiftUI's platform-view hosting mutates that view: it sets `autoresizingMask = [.flexibleWidth, .flexibleHeight]`, and `Menu`/`ContextMenu` additionally mutate frame/visibility **asynchronously, after their hierarchy has been torn down**.

#40604 mitigated one instance by restoring `autoresizingMask` in `dismantleUIView`, but SwiftUI does not reliably call `dismantleUIView` on wholesale teardown, and the deferred `Menu`/`ContextMenu` mutations can land **after Fabric has unmounted (or recycled) the view**. On the New Architecture the leaked `autoresize = W+H` then mangles Fabric-assigned frames whenever the superview resizes — views collapse to zero height, balloon, or render invisible while still occupying layout space. Because the mutations can arrive after the view has been recycled into an unrelated component, arbitrary views inherit the corruption (which also means resetting state at the recycle boundary cannot fix it — verified).

Repro and lldb view-hierarchy evidence in #47706: plain `RCTViewComponentView`s with `frame = (16 57; 370 0)` (zero height) and `autoresize = W+H` that no React code ever set.

# How

`makeUIView` now returns a **disposable isolation container** (`ReactViewIsolationContainer`) holding the React view as a subview:

- SwiftUI's mutations — synchronous or deferred — only ever reach the throwaway container, never the React-managed view.
- The container syncs the React view's frame to its bounds in `layoutSubviews`, preserving the previous fill behavior.
- The React view keeps its Yoga-assigned frame at creation: the `matchContents` sizing path (`ApplySizeFromYogaNode` in `@expo/ui`) KVO-observes the React child's bounds to size the SwiftUI content, so zeroing the child to the initially-empty container would deadlock that feedback loop at zero.
- Empty container bounds are not propagated (transient zero sizes during initial layout/teardown).
- `dismantleUIView` no longer needs to restore `autoresizingMask` — SwiftUI never touches the React view.

macOS (`makeNSView`) is intentionally left unchanged in this PR.

# Test Plan

- Reproduced the corruption with the repro in #47706 (SDK 57, RN 0.86, New Architecture, iOS 26 simulator + device): paging a keyed subtree containing `MenuView`s corrupts row rendering within a handful of pages.
- With this change applied (via `usePrecompiledModules: false` so `expo-modules-core` compiles from source): dozens of paging cycles across both row mixes render correctly every time — correct heights, no invisible/blank lists; `Menu` tap triggers and `ContextMenu` long-press previews keep working.
- Verified `matchContents` sizing still resolves (32×32 menu triggers render at their intrinsic size, not zero).

# Checklist

- [x] Documentation is up to date to reflect these changes (no doc changes needed — internal hosting behavior).


intergalacticspacehighway (changes requested):
Great fix. Thanks! Can we just gate it for macOS?

```swift
func makeUIView(context: Context) -> UIView {
#if os(macOS)
  return view
#else
  let container = ReactViewIsolationContainer()
  container.hostedView = view
  container.addSubview(view)
  return container
#endif
}
```

```swift
#if !os(macOS)
private final class ReactViewIsolationContainer: UIView {
  // … layoutSubviews() …
}
#endif
```

### 🟡 Minor (non-blocking)

`Coordinator.originalAutoresizingMask` is now dead — set only by `makeNSView`, and no longer read by anyone after `dismantleUIView` dropped it. The `Coordinator` payload can be trimmed.

cvburgess:
@intergalacticspacehighway this was happening on iOS _and_ macOS - let me see if that gate would cause a regression on iOS.

I'll admit, swift is not my strong suit!

intergalacticspacehighway:
@cvburgess lmk if you face any issues. I meant to just add the new UIView wrapper for iOS. Did you manage to repro it on macOS too!?

cvburgess:
Oh duh, my bad - i read `os(macOS)` - that `!` is important 😂 I'll update this PR now

cvburgess:
@intergalacticspacehighway I think I got it done, let me know!

packages/expo-modules-core/ios/Core/Views/SwiftUI/SwiftUIViewHost.swift
```diff
@@ -23,28 +22,47 @@ extension ExpoSwiftUI {
...
+      // teardown (see https://github.com/expo/expo/issues/40604 for a prior
+      // mitigation), and deferred mutations can land after the view has been
+      // unmounted or recycled into an unrelated component, which then
+      // renders zero-sized or invisible
+      // (https://github.com/expo/expo/issues/47706). The container absorbs
+      // all SwiftUI mutations; the React view tracks the container's bounds
+      // while mounted. It keeps its Yoga-assigned frame here — the
+      // matchContents sizing path observes the React view's bounds, so
+      // zeroing it to the empty container would deadlock that feedback loop
+      // at zero.
```
intergalacticspacehighway: Thanks for the changes. LGTM. This comment feels too big, can we shorten it a bit? 😅

intergalacticspacehighway (approved):

packages/expo-modules-core/ios/Core/Views/SwiftUI/SwiftUIViewHost.swift
cvburgess: Done!

Issue #47680, opened 2026-07-10
[@expo/ui][iOS] TextField nested in a container inside a Form crashes (SIGABRT, _resignOrRebaseFirstResponderView) after pushing a route and unmounting


andreavrr:
## Summary

On iOS, a SwiftUI `TextField` that is **nested inside a container** (e.g. `LabeledContent`, `HStack`) inside a `Form` crashes the app with `SIGABRT` when:

1. the field is focused (keyboard up), then
2. another route is pushed on top of the screen (react-native-screens), then
3. you return and the form screen is **unmounted** (e.g. dismissing a modal on save).

The exact same field placed as a **direct child of a `Section`** does **not** crash. Saving/unmounting **without** first pushing a route does **not** crash either. So the crash needs the combination: *nested field + focused + a screen pushed over it + unmount*.

## Minimal reproducible example

Native stack with two screens. Screen A has a `TextField` nested in a container inside a `Form`.

```tsx
// app/_layout.tsx
import { Stack } from 'expo-router';
export default function Layout() {
  return <Stack />;
}
```

```tsx
// app/index.tsx  (Screen A — the form)
import { router } from 'expo-router';
import { Host, Form, Section, LabeledContent, HStack, TextField, Button, Text } from '@expo/ui/swift-ui';

export default function A() {
  return (
    <Host style={{ flex: 1 }}>
      <Form>
        <Section>
          {/* TextField nested inside a container, NOT a direct Section child */}
          <LabeledContent label="Amount">
            <HStack>
              <TextField placeholder="0" />
            </HStack>
          </LabeledContent>
        </Section>
        <Section>
          <Button onPress={() => router.push('/b')}>
            <Text>Go to B</Text>
          </Button>
          <Button onPress={() => router.back()}>
            <Text>Close A</Text>
          </Button>
        </Section>
      </Form>
    </Host>
  );
}
```

```tsx
// app/b.tsx  (Screen B — any pushed screen)
import { router } from 'expo-router';
import { Host, Button, Text } from '@expo/ui/swift-ui';
export default function B() {
  return (
    <Host style={{ flex: 1 }}>
      <Button onPress={() => router.back()}><Text>Back to A</Text></Button>
    </Host>
  );
}
```

### Steps
1. From the root, navigate to A.
2. On A, **tap the "Amount" field** so it becomes first responder (keyboard shows).
3. Tap **"Go to B"** (pushes B over A **while the field is focused**).
4. Tap **"Back to A"**.
[...]

PR #47709, opened 2026-07-11
[core][ios] Fix crash unmounting a SwiftUI tree with a focused nested field

### LINKED ISSUE #47682: [@expo/ui] iOS SwiftUI TextField nested in a Form container crashes (SIGABRT) on unmount after a route is pushed while focused

https://github.com/andreavrr/expo-ui-textfield-crash

### Summary

On iOS, a SwiftUI `TextField` **nested inside a container** (`LabeledContent`, `HStack`, …) inside a `Form` crashes the app with `SIGABRT` when the field is focused, another route is pushed on top of the screen, and the form screen is then unmounted (in the repro: dismissing a modal from a **native header button**).

This is a fresh, runnable repro for the same problem I reported in #47680 (which was auto-closed for a missing reproducible example — my mistake, that one only had code in the description).

### Steps to reproduce

The whole repro is 4 route files under `src/app/`. Run it as a dev build (`@expo/ui` is native, so Expo Go won't work):

```bash
npm install
npx expo run:ios      # physical device or simulator, iOS 18+
```

1. Tap **"Open form (modal)"** — the form opens as a modal.
2. The **"Amount"** field auto-focuses (keyboard shows).
3. Tap **"1. Push screen B (while the field is focused)"**.
4. On B, tap **"2. Back to A"**.
5. Tap the native **"Save"** button in the header. → **crash**.

### Why each ingredient matters

The crash needs all of these together:

- **Nested field.** The `TextField` is inside `LabeledContent` + `HStack`, not a direct `Section` child. Making it a direct child of the `Section` → no crash.
- **A route pushed over the focused form, then popped** (steps 3–4). Skipping it and saving directly → no crash. The push detaches the form's native view while the field is first responder, leaving the SwiftUI `List`'s `UICollectionView` first-responder tracking stale.
- **Unmount via a control that does NOT resign the field.** Save is a native header button (`unstable_headerRightItems`), outside the SwiftUI `Host`, so tapping it does not blur the `TextField`. The field is still first responder when `router.back()` dismisses the modal and unmounts the `Form`.

### Actual crash (from a physical-device `.ips`)

```
SIGABRT
-[NSAssertionHandler handleFailureInMethod:object:file:lineNumber:description:]
-[UICollectionView _resignOrRebaseFirstResponderViewWithIndexPathMapping:]
-[UICollectionView _updateWithItems:tentativelyForReordering:...]
-[UICollectionView _endItemAnimations...] / _performBatchUpdates
SwiftUI UICollectionViewListCoordinatorBase.performUpdates
SwiftUI _UIHostingView.layoutSubviews()

// triggered by the RN unmount:
-[RCTMountingManager performTransaction:]
RCTPerformMountInstructions(...)
-[RCTViewComponentView unmountChildComponentView:index:]
-[UIView removeFromSuperview]
```

### Likely cause (hypothesis)

`expo-modules-core/ios/Core/Views/SwiftUI/SwiftUIVirtualView.swift` resigns the first responder on unmount only when the removed `contentView` **itself** is a focusable field:

```swift
private func virtualViewRemoveFromSuperview<ContentView: SwiftUI.View>(contentView: ContentView) {
  // When the view is unmounted, the focus on TextFieldView stays active and it causes a crash, so we blur it here
  if let focusableView = contentView as? any ExpoSwiftUI.FocusableView {
    focusableView.forceResignFirstResponder()
  }
}
```

When the `TextField` is nested inside a container (`LabeledContent`, `HStack`, …), the `contentView` is the **container**, the `as? FocusableView` cast fails, and the nested field is never resigned. Combined with the screen being detached by the earlier push (which leaves the `UICollectionView`'s first-responder tracking stale), the batch update at unmount asserts in `_resignOrRebaseFirstResponderViewWithIndexPathMapping`.

A fix would likely need `forceResignFirstResponder` to recurse into focusable descendants, and/or to clean the collection view's first-responder tracking when the hosting screen is detached.

### Workaround (app side)

Resigning the field's first responder (`TextFieldRef.blur()`) when its screen loses focus (e.g. via `useIsFocused` from `expo-router`) avoids the crash. Note: React Native's `Keyboard.dismiss()` does not work — it does not resign the SwiftUI `TextField`'s first responder.

### Environment

- `expo`: 57.0.4
- `@expo/ui`: 57.0.4
- `expo-router`: 57.0.4
- `react-native`: 0.86.0
- `react-native-screens`: 4.25.2
- `react`: 19.2.3
- iOS 26 / iOS 18+, physical device (originally seen on iPhone 16 Pro Max)

intergalacticspacehighway:
# Why

On iOS, when a view unmounts and a SwiftUI TextField is in focus we have to manually remove the focus or it leads to a crash. We fixed it by using `FocusableView` protocol and removing focus on unmount. But it is breaking when `TextField` is nested in a SwiftUI hierarchy. We need to recursively check if a TextField is present and remove the focus or else it leads to a crash. This looks like something handled by the UIView itself, but since we have virtual view we need to copy this behaviour.

Fixes #47682 (originally reported as #47680).

# How

Iterate children and remove the focus if it is a `FocusableView`.

# Test Plan

Tested on user shared repro.

# Checklist

- [x] I added a [changelog entry](https://github.com/expo/expo/blob/main/guides/contributing/Updating%20Changelogs.md) to the package's `CHANGELOG.md`.
- [x] Verified the change builds (Xcode Debug, iOS simulator).
- [ ] Conforms with the [Documentation Writing Style Guide](https://github.com/expo/expo/blob/main/guides/Expo%20Documentation%20Writing%20Style%20Guide.md) (no docs change).


intergalacticspacehighway:
Workflows are failing due to mismatched pnpm version between our repo and the one used by EAS. Looking into it!

```
Your pnpm version is incompatible with "/Users/expo/workingdir/build".
Expected version: ^10.33.0
Got: 11.9.0
```

Kudo (approved):

PR #47711, opened 2026-07-12
[config-plugins] fix `getApplicationIdAsync` and `setPackageInBuildGradle` failing with Gradle assignment syntax 

idoyana:
# Why

`AndroidConfig.Package.getApplicationIdAsync()` returns `null` when `android/app/build.gradle` declares the application id with the Gradle **assignment** syntax:

```gradle
defaultConfig {
    applicationId = "com.example.app"   // note the `=`
}
```

The current regex only matches the method-call form (`applicationId "com.example.app"`), so it fails to read the id from projects whose `build.gradle` uses the assignment form — e.g. files modernized to the new AGP DSL style (Android Studio's AGP Upgrade Assistant emits `applicationId = '…'` / `namespace = '…'` in Groovy), or any `build.gradle.kts` project, where assignment is the only valid syntax (`getAppBuildGradleFilePath` resolves `.gradle.kts` too).

This breaks `expo run:android`. `AndroidAppIdResolver.resolveAppIdFromNativeAsync()` calls `getApplicationIdAsync()` (→ `null`), then falls back to `androidManifest.manifest.$.package` — which no longer exists under AGP 8 (the package moved to `namespace` in `build.gradle`). Both resolve to `null`, so the run aborts before Gradle with:

> CommandError: Failed to locate the android application identifier in the "android/" folder. This is required to open the app.

Hit in a production **Expo SDK 57 / React Native 0.86** app (CNG — `android/` gitignored and periodically prebuilt): its `android/app/build.gradle` had been modernized to the AGP DSL style (`applicationId = '…'`, `namespace = '…'`) during an AGP 8 upgrade pass, and non-`--clean` prebuilds preserve the file. The file itself shows the silent write-path failure: `setVersionCode`'s form-agnostic regex kept rewriting its line back to method form on every prebuild, while `setPackageInBuildGradle`'s method-form-only regex silently no-oped on the assignment lines.

# How

Widen the regex in `getApplicationIdAsync` to allow an optional `=` between `applicationId` and the quoted value:

```diff
- const matchResult = buildGradle.match(/applicationId ['"](.*)['"]/);
+ const matchResult = buildGradle.match(/applicationId\s*=?\s*['"](.*)['"]/);
```

It still matches the legacy `applicationId "…"` form, so the change is backward compatible.

Per review, `setPackageInBuildGradle` is fixed as well: its pattern now matches either syntax and the replacement preserves the original separator, so method-call files stay `applicationId '…'` and assignment files stay `applicationId = '…'`. Tests cover both syntaxes for both functions via `describe.each`.

# Test Plan

- Added a regression test in `Package-test.ts` covering the assignment syntax (`applicationId = "com.helloworld"`), next to the existing method-call test.
- Both pass locally.
- Verified against a real SDK 57 project: with the fix, `getApplicationIdAsync` returns the correct id (previously `null`), and `expo run:android` resolves the launch activity instead of throwing.

# Checklist

- [x] I have added tests to cover my changes.
- [x] I have added a changelog entry (`packages/@expo/config-plugins/CHANGELOG.md`).
- [x] This is a backward-compatible bug fix.

### DIFF
--- a/packages/@expo/config-plugins/CHANGELOG.md
+++ b/packages/@expo/config-plugins/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### 🐛 Bug fixes
 
 - Honor `ios.version` and `android.version` in `Updates.getAppVersion`, `Updates.getNativeVersion`, and the `appVersion` runtime version policy. Previously the platform-specific overrides were ignored, so projects that used only `ios.version`/`android.version` (with no top-level `version` in `app.json`) received the `package.json` fallback (or `"1.0.0"`) wherever these helpers were consumed. `Updates.getAppVersion` gains an optional `platform` argument; passing it prefers the platform-specific override, and calls without a platform keep the previous behavior. Also fixes `Updates.getNativeVersion` on Android, which previously used the iOS version for the `${version}` component. ([#47416](https://github.com/expo/expo/pull/47416) by [@tlenahan](https://github.com/tlenahan))
+- Fix `getApplicationIdAsync` failing to read `applicationId` when it is declared with the Gradle assignment syntax (`applicationId = '...'`) used by recent React Native Android templates. This caused `expo run:android` to abort with "Failed to locate the android application identifier" on SDK 57 / React Native 0.86 projects. ([#47711](https://github.com/expo/expo/pull/47711) by [@idoyana](https://github.com/idoyana))
 
 ### 💡 Others
 
--- a/packages/@expo/config-plugins/CHANGELOG.md
+++ b/packages/@expo/config-plugins/CHANGELOG.md
@@ -9,7 +9,7 @@
 ### 🐛 Bug fixes
 
 - Honor `ios.version` and `android.version` in `Updates.getAppVersion`, `Updates.getNativeVersion`, and the `appVersion` runtime version policy. Previously the platform-specific overrides were ignored, so projects that used only `ios.version`/`android.version` (with no top-level `version` in `app.json`) received the `package.json` fallback (or `"1.0.0"`) wherever these helpers were consumed. `Updates.getAppVersion` gains an optional `platform` argument; passing it prefers the platform-specific override, and calls without a platform keep the previous behavior. Also fixes `Updates.getNativeVersion` on Android, which previously used the iOS version for the `${version}` component. ([#47416](https://github.com/expo/expo/pull/47416) by [@tlenahan](https://github.com/tlenahan))
-- Fix `getApplicationIdAsync` failing to read `applicationId` when it is declared with the Gradle assignment syntax (`applicationId = '...'`) used by recent React Native Android templates. This caused `expo run:android` to abort with "Failed to locate the android application identifier" on SDK 57 / React Native 0.86 projects. ([#47711](https://github.com/expo/expo/pull/47711) by [@idoyana](https://github.com/idoyana))
+- Fix `getApplicationIdAsync` failing to read `applicationId` when it is declared with the Gradle assignment syntax (`applicationId = '...'`) used by recent React Native Android templates. Also fix `setPackageInBuildGradle` to match (and preserve) the assignment syntax when replacing the package. ([#47711](https://github.com/expo/expo/pull/47711) by [@idoyana](https://github.com/idoyana))
 
 ### 💡 Others
 
--- a/packages/@expo/config-plugins/CHANGELOG.md
+++ b/packages/@expo/config-plugins/CHANGELOG.md
@@ -9,7 +9,7 @@
 ### 🐛 Bug fixes
 
 - Honor `ios.version` and `android.version` in `Updates.getAppVersion`, `Updates.getNativeVersion`, and the `appVersion` runtime version policy. Previously the platform-specific overrides were ignored, so projects that used only `ios.version`/`android.version` (with no top-level `version` in `app.json`) received the `package.json` fallback (or `"1.0.0"`) wherever these helpers were consumed. `Updates.getAppVersion` gains an optional `platform` argument; passing it prefers the platform-specific override, and calls without a platform keep the previous behavior. Also fixes `Updates.getNativeVersion` on Android, which previously used the iOS version for the `${version}` component. ([#47416](https://github.com/expo/expo/pull/47416) by [@tlenahan](https://github.com/tlenahan))
-- Fix `getApplicationIdAsync` failing to read `applicationId` when it is declared with the Gradle assignment syntax (`applicationId = '...'`) used by recent React Native Android templates. Also fix `setPackageInBuildGradle` to match (and preserve) the assignment syntax when replacing the package. ([#47711](https://github.com/expo/expo/pull/47711) by [@idoyana](https://github.com/idoyana))
+- Fix `getApplicationIdAsync` and `setPackageInBuildGradle` failing with the Gradle assignment syntax (`applicationId = '...'`). ([#47711](https://github.com/expo/expo/pull/47711) by [@idoyana](https://github.com/idoyana))
 
 ### 💡 Others
 
--- a/packages/@expo/config-plugins/src/android/Package.ts
+++ b/packages/@expo/config-plugins/src/android/Package.ts
@@ -248,7 +248,10 @@ export async function getApplicationIdAsync(projectRoot: string): Promise<string
     return null;
   }
   const buildGradle = await fs.promises.readFile(buildGradlePath, 'utf8');
-  const matchResult = buildGradle.match(/applicationId ['"](.*)['"]/);
+  // Match both the legacy method-call form (`applicationId "com.app"`) and the
+  // Gradle assignment form (`applicationId = "com.app"`) emitted by recent
+  // React Native Android templates.
+  const matchResult = buildGradle.match(/applicationId\s*=?\s*['"](.*)['"]/);
   // TODO add fallback for legacy cases to read from AndroidManifest.xml
   return matchResult?.[1] ?? null;
 }
--- a/packages/@expo/config-plugins/src/android/Package.ts
+++ b/packages/@expo/config-plugins/src/android/Package.ts
@@ -238,8 +238,11 @@ export function setPackageInBuildGradle(config: Pick<ExpoConfig, 'android'>, bui
     return buildGradle;
   }
 
-  const pattern = new RegExp(`(applicationId|namespace) ['"].*['"]`, 'g');
-  return buildGradle.replace(pattern, `$1 '${packageName}'`);
+  // Match either declaration syntax, preserving the original separator so the
+  // method-call form stays `applicationId '...'` and the assignment form stays
+  // `applicationId = '...'`.
+  const pattern = new RegExp(`(applicationId|namespace)(\\s*=\\s*|\\s+)['"].*['"]`, 'g');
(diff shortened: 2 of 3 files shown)


idoyana:
Hi @kitten 👋 — you've been doing most of the recent config-plugins work, so flagging this small fix in case it slipped past the auto-mentions.

`getApplicationIdAsync` only matches the method-call form (`applicationId "com.app"`) and misses the assignment syntax (`applicationId = "com.app"`) that the current RN Android template generates. Under AGP 8 the manifest-package fallback is gone too, so `expo run:android` aborts with "Failed to locate the android application identifier" on a clean SDK 57 prebuild.

The fix just makes the `=` optional in the regex — backward compatible, regression test added. Tiny diff; happy to adjust anything. Thanks! 🙏

idoyana:
Hi @vonovak 👋 — would you be able to take a look at this small config-plugins fix? `getApplicationIdAsync` misses the Gradle assignment syntax (`applicationId = "…"`) that the current RN Android template generates, and under AGP 8 the manifest fallback is gone too — so `expo run:android` fails on a clean SDK 57 prebuild. One-character regex fix, tests included. Thanks! 🙏

packages/@expo/config-plugins/src/android/__tests__/Package-test.ts
vonovak, line 15: @idoyana can you add another example build gradle with the new syntax and then use `describe.each` to test `getApplicationIdAsync` and `setPackageInBuildGradle` as well, thank you!

packages/@expo/config-plugins/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
...
+- Fix `getApplicationIdAsync` failing to read `applicationId` when it is declared with the Gradle assignment syntax (`applicationId = '...'`) used by recent React Native Android templates. This caused `expo run:android` to abort with "Failed to locate the android application identifier" on SDK 57 / React Native 0.86 projects. ([#47711](https://github.com/expo/expo/pull/47711) by [@idoyana](https://github.com/idoyana))
```
vonovak:
```suggestion
- Fix `getApplicationIdAsync` failing to read `applicationId` when it is declared with the Gradle assignment syntax (`applicationId = '...'`) used by recent React Native Android templates. ([#47711](https://github.com/expo/expo/pull/47711) by [@idoyana](https://github.com/idoyana))
```

vonovak (changes requested):
thank you for the PR, see my comments :)

idoyana:
Thanks for the review @vonovak! Done in 0d795a08:

- Applied your changelog suggestion (with a short added clause, since the scope grew — see below).
- Added an assignment-syntax `build.gradle` fixture and converted the tests to `describe.each` over both syntaxes, covering `getApplicationIdAsync` **and** `setPackageInBuildGradle`.
- Testing `setPackageInBuildGradle` against the assignment fixture meant fixing it too (its regex had the same blind spot), so the write path now matches either syntax and **preserves the original separator** when replacing — method-call files stay `applicationId '…'`, assignment files stay `applicationId = '…'`.

All 14 tests in `Package-test.ts` pass locally. Happy to adjust anything else!

packages/@expo/config-plugins/src/android/Package.ts
```diff
@@ -248,7 +251,10 @@ export async function getApplicationIdAsync(projectRoot: string): Promise<string
     return null;
   }
   const buildGradle = await fs.promises.readFile(buildGradlePath, 'utf8');
-  const matchResult = buildGradle.match(/applicationId ['"](.*)['"]/);
+  // Match both the legacy method-call form (`applicationId "com.app"`) and the
+  // Gradle assignment form (`applicationId = "com.app"`) emitted by recent
+  // React Native Android templates.
```
vonovak:
```suggestion
  // Gradle assignment form (`applicationId = "com.app"`)
```

"recent" isn't gonna read well 2 years from now

you can also drop the whole comment

packages/@expo/config-plugins/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### 🐛 Bug fixes
 
 - Honor `ios.version` and `android.version` in `Updates.getAppVersion`, `Updates.getNativeVersion`, and the `appVersion` runtime version policy. Previously the platform-specific overrides were ignored, so projects that used only `ios.version`/`android.version` (with no top-level `version` in `app.json`) received the `package.json` fallback (or `"1.0.0"`) wherever these helpers were consumed. `Updates.getAppVersion` gains an optional `platform` argument; passing it prefers the platform-specific override, and calls without a platform keep the previous behavior. Also fixes `Updates.getNativeVersion` on Android, which previously used the iOS version for the `${version}` component. ([#47416](https://github.com/expo/expo/pull/47416) by [@tlenahan](https://github.com/tlenahan))
+- Fix `getApplicationIdAsync` failing to read `applicationId` when it is declared with the Gradle assignment syntax (`applicationId = '...'`) used by recent React Native Android templates. Also fix `setPackageInBuildGradle` to match (and preserve) the assignment syntax when replacing the package. ([#47711](https://github.com/expo/expo/pull/47711) by [@idoyana](https://github.com/idoyana))
```
vonovak:
```suggestion
- Fix `getApplicationIdAsync` and `setPackageInBuildGradle` failing with the Gradle assignment syntax (`applicationId = '...'`). ([#47711](https://github.com/expo/expo/pull/47711) by [@idoyana](https://github.com/idoyana))
```

packages/@expo/config-plugins/src/android/Package.ts
```diff
@@ -238,8 +238,11 @@ export function setPackageInBuildGradle(config: Pick<ExpoConfig, 'android'>, bui
     return buildGradle;
   }
 
-  const pattern = new RegExp(`(applicationId|namespace) ['"].*['"]`, 'g');
-  return buildGradle.replace(pattern, `$1 '${packageName}'`);
+  // Match either declaration syntax, preserving the original separator so the
+  // method-call form stays `applicationId '...'` and the assignment form stays
+  // `applicationId = '...'`.
```
vonovak:
```suggestion
```
git history will explain this

packages/@expo/config-plugins/src/android/__tests__/Package-test.ts
```diff
@@ -47,17 +65,37 @@ describe('package', () => {
...
-    ).toMatch("applicationId 'my.new.app'");
-  });
-
-  it(`sets the namespace in build.gradle if package is given`, () => {
-    expect(
-      setPackageInBuildGradle({ android: { package: 'my.new.app' } }, EXAMPLE_BUILD_GRADLE)
-    ).toMatch("namespace 'my.new.app'");
-  });
+  describe.each([
+    ['method-call', EXAMPLE_BUILD_GRADLE, "applicationId 'my.new.app'", "namespace 'my.new.app'"],
```
vonovak: no point in repeating `applicationId` and `namespace` is they are the same. Just use them in the tests directly

packages/@expo/config-plugins/src/android/__tests__/Package-test.ts
```diff
@@ -47,17 +65,37 @@ describe('package', () => {
     expect(getApplicationIdAsync(projectRoot)).resolves.toBe('com.helloworld');
   });
```
vonovak:
```suggestion
```

delete this individual test case, the functionality is covered by what you added

packages/@expo/config-plugins/src/android/__tests__/Package-test.ts
```diff
@@ -28,6 +28,24 @@ const EXAMPLE_BUILD_GRADLE = `
   }
   `;
 
+// Recent React Native Android templates define the applicationId and namespace
+// using the Gradle assignment form rather than the legacy method-call form.
+const EXAMPLE_BUILD_GRADLE_ASSIGNMENT = `
```
vonovak:
```suggestion
const EXAMPLE_BUILD_GRADLE_ASSIGNMENT_FORM = `
```

nicer name

vonovak (changes requested):
let's fine tune this a bit more, thanks!

packages/@expo/config-plugins/src/android/Package.ts
idoyana: Applied in 420a6027.

packages/@expo/config-plugins/CHANGELOG.md
idoyana: Applied in 420a6027.

packages/@expo/config-plugins/src/android/Package.ts
idoyana: Removed in 420a6027.

packages/@expo/config-plugins/src/android/__tests__/Package-test.ts
idoyana: Restructured in 420a6027 — the table now carries just the separator (' ' / ' = ') and the tests build the `applicationId`/`namespace` expectations inline.

packages/@expo/config-plugins/src/android/__tests__/Package-test.ts
idoyana: Deleted in 420a6027.

packages/@expo/config-plugins/src/android/__tests__/Package-test.ts
idoyana: Renamed in 420a6027. Also dropped the comment above the fixture — it had the same "recent templates" wording you flagged in Package.ts.

vonovak (approved):

vonovak (approved):

packages/@expo/config-plugins/src/android/__tests__/Package-test.ts
vonovak: haha I missed that they (applicationId and namespace) were different before. But this is okay

vonovak:
@idoyana you mentioned

> getApplicationIdAsync misses the Gradle assignment syntax (applicationId = "…") that the current RN Android template generates

what template was that? Can you share it?

Thanks!

idoyana:
> what template was that? Can you share it?

Good question — I went back and verified, and my "current RN Android template" phrasing was wrong: I checked `react-native-community/template` (`main` / `0.86-stable` / `0.87-stable`) and `expo-template-bare-minimum` (sdk-55 through main, including the npm tarballs actually published at the time) — all of them emit the method-call form.

The real provenance, from the production app where this bit us (Expo CNG — `android/` gitignored and periodically prebuilt): prebuild originally generated the file from `expo-template-bare-minimum` (method-call), and it was later modernized to the new AGP DSL style (`compileSdk =`, `minSdk =`, `namespace =`, `applicationId = '…'`) during an AGP 8 upgrade pass. Non-`--clean` prebuilds preserve the file — and the file itself records why the drift stuck: `setVersionCode`'s form-agnostic regex (`versionCode.*`) kept rewriting its line back to method form on every prebuild, while `setPackageInBuildGradle`'s method-form-only regex silently no-oped on the assignment lines. So the write path you had me fix here was already silently failing in the wild, and `expo run:android` then broke via the read path once the AGP 8 manifest fallback was gone.

Also independent of any template: `getAppBuildGradleFilePath` resolves `build.gradle.kts` too, where assignment is the only valid syntax. I've updated the PR description accordingly.

PR #47713, opened 2026-07-12
[ui][ios] Fix touches broken after BottomSheet dismiss

### LINKED ISSUE #47708: [iOS][@expo/ui] BottomSheet unmounted during JS-driven dismiss breaks all subsequently presented RN Modals (touches swallowed, missing from accessibility tree)

https://github.com/oeddyo/re-produce-active-sheet-issue

Created with `npx create-expo-app@latest --template blank-typescript` + `@expo/ui`. One ~150-line `App.tsx`, no navigation library. Includes an optional Maestro flow (`maestro/repro.yaml`) that is written to pass when the bug is fixed and fail while it exists, plus its screenshot artifacts.

### Steps to reproduce

**Platform:** iOS (simulator, iOS 18.x). **Environment:** development build (`npx expo run:ios`) — not reproducible in Expo Go since `@expo/ui` is a native module. **Package manager:** npm. New Architecture (SDK 57 default).

```
npm install
npx expo run:ios
```

1. Sanity check: tap **"Open RN Modal"**, tap **"Tap me (in modal)"** (counter increments), tap **"Close modal"**. Everything works.
2. Tap **"Open sheet"** — an `@expo/ui/community/bottom-sheet` opens.
3. Tap **"Dismiss + unmount now (BUG)"** inside it. This sets `index` to `-1` **and unmounts the `<BottomSheet>` in the same tick** — which is exactly what happens in a real app whenever selecting an item in a sheet triggers navigation (`router.replace`/`push`), the hosting screen closes, or the data backing the sheet drains.
4. The sheet is gone and the base screen still works fine (the bottom "Tap me" button increments).
5. Tap **"Open RN Modal"** again. The modal appears and looks completely normal.
6. Tap **"Tap me (in modal)"**, then try **"Close modal"**.

**Expected:** the modal behaves exactly as in step 1 — counter increments, close works.

**Actual:** every touch inside the modal is silently swallowed — the counter never moves and the modal cannot be closed (the app must be killed). The modal's content is also entirely missing from the accessibility tree (verified with XCUITest-based tooling; the same query that passed in step 1 finds nothing in step 6), so VoiceOver users and any UI-test framework are equally locked out. The base window underneath keeps working, which makes the breakage easy to miss in development.

Once triggered, the state is permanent for the app session: **every** RN `<Modal>` presented afterwards is broken the same way.

Control cases in the same repro, all fine:
- "Dismiss (stay mounted)" — plain `index` → `-1` with the wrapper kept mounted.
- Gesture dismissal (swipe down / backdrop tap).
- "Dismiss, unmount in 600ms (workaround)" — unmounting only after the dismissal animation completes avoids the bug entirely. This is the app-side workaround today, but it forces every navigate-on-selection to wait for the animation, and the fixed delay races the animation duration.

Relationship to existing issues: this is the same family as #45812 ("BottomSheet on iOS leaves touch-blocking host mounted after programmatic dismiss"). The fix from #45846 (in ≥ 56.0.9; we're on 57.0.4) does not cover this path — consistent with the reporter's follow-up comment there that the JS-driven dismiss flow still leaves the blocking host and only a delayed unmount of the wrapper clears it. The new information in this report: the leaked state doesn't just cover the sheet's former frame — it breaks all subsequently presented `UIWindow`s (plain RN Modals), both for touch delivery and for the accessibility snapshot. A fix direction that matches the evidence: defer the native host teardown until dismissal completes (a native `onDismiss` event already exists) when the wrapper unmounts mid-dismissal, instead of orphaning the presentation state.

Screenshots (also in the repo): `after-blind-taps.png` shows the modal still open with "Modal taps: 1" after three synthesized touches on its buttons.

### Environment

```
    System:
      OS: macOS 26.5
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 22.13.0 - /Users/eddiexie/.nvm/versions/node/v22.13.0/bin/node
      Yarn: 1.22.19 - /opt/homebrew/bin/yarn
      npm: 10.9.2 - /Users/eddiexie/.nvm/versions/node/v22.13.0/bin/npm
    Managers:
    SDKs:
      iOS SDK:
      Android SDK:
        API Levels: 35, 36
        System Images: android-35 | Google APIs ARM 64 v8a, android-35 | Google Play ARM 64 v8a
    IDEs:
      Android Studio: 2024.2 AI-242.23726.103.2422.13103373
    npmPackages:
      expo: ~57.0.4 => 57.0.4
      react: 19.2.3 => 19.2.3
    npmGlobalPackages:
      eas-cli: 20.5.1
    Expo Workflow: managed
```

Additional relevant package: `@expo/ui` ~57.0.4.

### Expo Doctor Diagnostics

```
```

intergalacticspacehighway:
# Why

Fixes #47708. After a `community/bottom-sheet` `BottomSheet` is dismissed, RN `Modal`s presented afterwards swallow all touches and go missing from the accessibility tree. This is not an issue on android as RN touch handler is attached to a new wrapper View and not the one getting recycled by Fabric.

# How

`RNHostView` attaches an `RCTSurfaceTouchHandler` to the hosted RN view but never detaches it. Fabric recycles component views across surfaces, so the handler rides the recycled view into the next surface (the `Modal`) and terminates its touch responder. Now detached on `.onDisappear`.

# Test Plan

Tested repro.


Kudo (approved):
sweet!

oeddyo:
Thank you!

oeddyo:
@intergalacticspacehighway is it possible to publish a 57.0.5 including this change please? We have been waiting to test this for a bit. Would love to have it and try it out!

intergalacticspacehighway:
Hi @oeddyo. I have cherry-picked it in SDK 57 and 56. It will be up in the next release (should happen probably today or tomorrow!)

oeddyo:
Hi @intergalacticspacehighway 
Double checked with Phil today and he mentioned the release is on an ad-hoc basis. No auto-release scheduled... I'm wondering could you spare 2 minutes to help create a release? We have a patch but really want to minimize the deviation from expo core.. thank you in advance!

PR #47714, opened 2026-07-12
[docs][map-view] Add .easignore troubleshooting tip for API key injection

Bert0ns:
# Why

When using a **.env** file to supply the Google Maps API key via the config plugin, the key silently fails to load in EAS Build if the project's **.gitignore** excludes **.env** and no **.easignore** file is present. 
* Common pitfall for inexperienced devs, and there is currently no mention of it in the docs.

# How

Added an `> **info**` callout in both the Android (step 4) and iOS (step 3) "Add the API key to your project" sections, explaining that a **.easignore** file (mirroring **.gitignore** but not excluding **.env**) may be needed for the API key to be available at build time.

Changes applied to both `unversioned` and `v57.0.0` copies of the file.

# Test plan

Documentation-only change. Verified that:
- The MDX syntax is consistent with existing callouts in the file (for example, `> **important**` on line 15).
- File names use **bold** formatting per the [Expo Documentation Writing Style Guide](https://github.com/expo/expo/blob/main/guides/Expo%20Documentation%20Writing%20Style%20Guide.md).
- Copy uses present tense, active voice, and second person per style guide conventions.
- Both `unversioned` and `v57.0.0` copies are kept in sync.

# Checklist

- [ ] I added a `changelog.md` entry and rebuilt the package sources according to [this short guide](https://github.com/expo/expo/blob/main/CONTRIBUTING.md#-before-submitting) — *N/A, docs-only change*
- [ ] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: updated a module plugin) — *N/A, docs-only change*

### DIFF
--- a/docs/pages/versions/unversioned/sdk/map-view.mdx
+++ b/docs/pages/versions/unversioned/sdk/map-view.mdx
@@ -125,6 +125,8 @@ Since you are using Google as the map provider, you need to add the API key to t
 }
 ```
 
+> **info** If EAS Build does not pick up the API key, make sure your project contains a [.easignore](/build-reference/easignore/) file. This file should mirror your **.gitignore** but must not exclude **.env**, so that EAS Build can access your environment variables.
+
 - In your code, import `{ PROVIDER_GOOGLE }` from `react-native-maps` and add the property `provider={PROVIDER_GOOGLE}` to your `<MapView>`. This property works on both Android and iOS.
 - Rebuild the app binary (or re-submit to the Google Play Store in case your app is already uploaded). An easy way to test if the configuration was successful is to do an [emulator build](/develop/development-builds/introduction/#how-would-you-like-to-build-your-development-build).
 
--- a/docs/pages/versions/v57.0.0/sdk/map-view.mdx
+++ b/docs/pages/versions/v57.0.0/sdk/map-view.mdx
@@ -125,6 +125,8 @@ Since you are using Google as the map provider, you need to add the API key to t
 }
 ```
 
+> **info** If EAS Build does not pick up the API key, make sure your project contains a [.easignore](/build-reference/easignore/) file. This file should mirror your **.gitignore** but must not exclude **.env**, so that EAS Build can access your environment variables.
+
 - In your code, import `{ PROVIDER_GOOGLE }` from `react-native-maps` and add the property `provider={PROVIDER_GOOGLE}` to your `<MapView>`. This property works on both Android and iOS.
 - Rebuild the app binary (or re-submit to the Google Play Store in case your app is already uploaded). An easy way to test if the configuration was successful is to do an [emulator build](/develop/development-builds/introduction/#how-would-you-like-to-build-your-development-build).
 


docs/pages/versions/unversioned/sdk/map-view.mdx
```diff
@@ -16,6 +16,8 @@ import { Tabs, Tab } from '~/ui/components/Tabs';
 
 `react-native-maps` provides a Map component that uses Google Maps on Android and Apple Maps or Google Maps on iOS.
 
+> **warning** `react-native-maps` does not support the web platform. If your app targets web, you need to provide a platform-specific alternative or use a web-compatible map library.
```
amandeepmittal:
```suggestion
```

The eligible platform already does not mention web as a platform. No need for a callout since the description and platform tags are enough to indicate which platforms are supported for this library.

Image

amandeepmittal (changes requested):

Bert0ns:
i made the commit messages and pr title compliant with the repository standards.

docs/pages/versions/unversioned/sdk/map-view.mdx
```diff
@@ -110,6 +110,8 @@ const styles = StyleSheet.create({
 
 Since you are using Google as the map provider, you need to add the API key to the `react-native-maps` [config plugin](/config-plugins/introduction/). Copy your **API Key** into your project to either a **.env** file or copy it directly and then add it to your app config under the `plugins.react-native-maps.androidGoogleMapsApiKey` field like:
 
+> **info** If the API key is not injected at build time, make sure your project contains a **.easignore** file. This file should mirror your **.gitignore** but **must not exclude .env**, so that EAS Build can access your environment variables.
```
amandeepmittal: ⚠️ **correctness** `links-and-headings.md#internal-links`

This restates guidance that already lives on [Using environment variables without EAS](/eas/environment-variables/without-eas/#how-environment-variables-are-loaded), and it offers shipping **.env** to the builder as the only fix when EAS environment variables are the recommended path for a build-time value like this key. Link to that page from the paragraph above instead of repeating the guidance in four places (both platform sections, in both copies of this file). If the callout stays, move it below the **app.json** block so the sentence ending in "like:" still runs into the example it introduces, and link **.easignore** to [Ignore files via .easignore](/build-reference/easignore/).

---

"the API key is not injected at build time" is passive. Write it actively, such as "If EAS Build does not pick up the API key".

---

💡 **nit** `mechanics.md#file-and-directory-names-in-bold`

Bold marks file and directory names in this sentence, so "**must not exclude .env**" reads like one: keep the bold on **.env** only.

docs/pages/versions/unversioned/sdk/map-view.mdx
```diff
@@ -159,6 +161,8 @@ Since you are using Google as the map provider, you need to add the API key to t
 
 Since you are using Google as the map provider, you need to add the API key to the `react-native-maps` [config plugin](/config-plugins/introduction/). Copy your **API Key** into your project to either a **.env** file or copy it directly and then add it to your app config under the `plugins.react-native-maps.iosGoogleMapsApiKey` field like:
 
+> **info** If the API key is not injected at build time, make sure your project contains a **.easignore** file. This file should mirror your **.gitignore** but **must not exclude .env**, so that EAS Build can access your environment variables.
```
amandeepmittal:
```suggestion
```

I think the callout once referenced is enough per page.

amandeepmittal (changes requested):
Thanks for updating docs!

Left some suggestions.

Bert0ns:
Hi, i updated the pr following your review

amandeepmittal (approved):

PR #47717, opened 2026-07-12
[expo-modules-core] Fix crash when async function calls outlive a runtime reload

### LINKED ISSUE #47716: [expo-modules-core] Async function crash on reload while native calls are suspended (JavaScriptValuesBuffer.deinit)

https://github.com/AlirezaHadjar/expo-values-buffer-repro

### Steps to reproduce

Just to make sure that the fix isn't covered by the merged but unreleased commits, I've applied them in my repro to prove that the app still crashes

1. `bun install` (the #47492 and #47521 fixes are applied automatically via `patchedDependencies`, see `patches/`)
2. `bunx expo prebuild -p ios`
3. `bun run ios:release`
4. Tap "Stress then reload"

### Environment

```text
    System:
      OS: macOS 26.5.2
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 22.15.0 - /Users/alirezahadjar/.nvm/versions/node/v22.15.0/bin/node
      Yarn: 1.22.22 - /Users/alirezahadjar/.nvm/versions/node/v22.15.0/bin/yarn
      npm: 11.13.0 - /Users/alirezahadjar/.nvm/versions/node/v22.15.0/bin/npm
      Watchman: 2026.06.01.00 - /opt/homebrew/bin/watchman
    Managers:
      CocoaPods: 1.16.2 - /Users/alirezahadjar/.rbenv/shims/pod
    SDKs:
      iOS SDK:
    IDEs:
      Android Studio: 2025.1 AI-251.25410.109.2511.13752376
    npmPackages:
      expo: 57.0.4 => 57.0.4
      expo-updates: 57.0.6 => 57.0.6
      react: 19.2.3 => 19.2.3
    npmGlobalPackages:
      eas-cli: 20.5.0
    Expo Workflow: managed
```

### Expo Doctor Diagnostics

```text
```

AlirezaHadjar:
# Why

Reloading the app while Swift concurrency `AsyncFunction` calls are still suspended in their native body crashes with `EXC_BAD_ACCESS`:

`ConcurrentFunctionDefinition.build` goes through `createAsyncFunction`, which captures `this` and a copy of the arguments buffer into the scheduled async task. Those wrappers own `jsi::Value`s and stay alive in the suspended task frame. `JavaScriptActor` runs its jobs synchronously on the calling thread, so when the body resumes on the cooperative pool, the frame teardown also runs there. If the runtime was destroyed in the meantime, the destructors run against a freed runtime.

Still reproduces with #47492 and #47521 applied, so this is a separate gap from the scheduler and promise lifetime fixes. We hit it in a production App Store build on iOS 26 right after an expo-updates check.

Closes #47716

Minimal repro: https://github.com/AlirezaHadjar/expo-values-buffer-repro (500 async module calls with a 2.5s native delay, reload 20ms later, crashes on the first tap every time)

# How

Build the host function with the synchronous `createFunction` instead, mirroring the shape `AsyncFunctionDefinition` already uses: create the promise and decode `this` plus arguments into native values inside the host call, then schedule only the converted native tuple into the async task. No JSI owned wrapper crosses the suspension boundary. The arguments copy is no longer needed since decoding now happens while the caller owned buffer is still valid.

# Test Plan

- Repro app above, Release simulator build with `expo-modules-core`, `expo-modules-jsi` and `expo` built from source: crashes on the first tap without this change, survived 11 consecutive stress plus reload rounds with it
- Return values verified end to end: the repro's "Stress without reload" logs the resolved value (`label:payloadCount:itemsCount`) correctly
- `requiresMainActor`, `takesOwner` and argument validation paths are structurally unchanged, decoding uses the same helpers on the JS thread

# Checklist


tsapeta (approved):
Looks good to me as a safe and quick fix for SDK 57. Thanks!
Ultimately we should rather fix this in `createAsyncFunction` on `expo-modules-jsi` side, but it might be a little bigger change that I wouldn't backport to SDK 57. I'll follow up on this.

PR #47724, opened 2026-07-12
fix(babel-preset-expo): Fix legacy decorators on class properties for hermes-v1 and web profiles

### LINKED ISSUE #47722: [SDK 57] Legacy decorators on class properties throw at runtime: hermes-v1 profile lacks class-properties transform (breaks WatermelonDB, MobX 5)

https://github.com/Gitarcitano/expo-sdk57-legacy-decorators-repro

Android, iOS (any Hermes target using the modern transform profile)

In a development build and in Expo Go, on iOS simulator (also reproducible by compiling with Babel directly, see below)

### Summary

`babel-preset-expo` enables `@babel/plugin-proposal-decorators` (`{ legacy: true }`) by default (`lazyDecoratorsPlugin` in `configs/expo.js`), but the modern Hermes transform profile no longer applies `@babel/plugin-transform-class-properties`: `configs/hermes-v1.js` does not include it; only `hermes-v0.js` and `webview.js` do.

The legacy decorators transform requires the class-properties transform to run after it. When it is absent, Babel compiles every decorated class property into a call to `initializerWarningHelper`, which throws at runtime in the class constructor:

```
Error: Decorating class property failed. Please ensure that transform-class-properties
is enabled and runs after the decorators transform.
```

So the preset's default configuration is internally inconsistent on SDK 57 / RN 0.86: decorators are on by default, but the transform they depend on is gone. Upstream `@react-native/babel-preset` also skips class-properties when `preserveClasses` is set for Hermes V1, but it does not enable decorators by default, so upstream the combination stays consistent.

This breaks every library using legacy decorators on class properties, notably WatermelonDB models (`@field`, `@text`, ...) and MobX 5. The WatermelonDB failure mode is particularly hard to trace: the Model constructor throws during record materialization after the native adapter has already marked the record as sent, permanently desyncing its record cache - the app then floods with `Record ID xxx was sent over the bridge, but it's not cached` errors while the real error above surfaces only once and is easily swallowed.

### Reproduction (no device needed)

The repro is `create-expo-app --template blank-typescript` plus one decorated class:

```ts
function watch(target: unknown, key: string, descriptor?: PropertyDescriptor) {
  return descriptor;
}

export class Model {
  @watch greeting = 'decorated field works';
}
```

Compile it with the same configuration Metro uses:

```bash
node -e "
const babel = require('@babel/core');
const presetPath = require('path').resolve('node_modules/expo/node_modules/babel-preset-expo');
const caller = { name: 'metro', bundler: 'metro', platform: 'ios', engine: 'hermes', isDev: false, projectRoot: process.cwd() };
console.log(babel.transformFileSync('Model.ts', { caller, babelrc: false, presets: [presetPath] }).code);
"
```

Output (note the field initializer):

```js
var Model=exports.Model=(_class=class Model{greeting=(0,_initializerWarningHelper2.default)(_descriptor,this);},_descriptor=(0,_applyDecoratedDescriptor2.default)(_class.prototype,"greeting",[watch],{...}),_class);
```

`new Model()` throws. Running the app shows `THROWS: Decorating class property failed. ...` instead of `decorated field works`.

### Expected behavior

Either the default decorators support produces working output on the modern Hermes profile (e.g. `hermes-v1.js` conditionally includes the class-feature transforms when decorators are enabled), or decorators default to off on profiles where the preset cannot support them, with a clear build-time error when decorator syntax is encountered.

### Workaround

Documented in the repro README: a custom preset with `@babel/plugin-proposal-decorators` + the class-feature transforms placed before `babel-preset-expo` in `presets`, combined with `['babel-preset-expo', { decorators: false }]`. Note for anyone landing here: the plugins must be `@babel/*@^7` (the `8.x` versions silently produce the same broken output because their `helper-create-class-features-plugin` cannot cooperate with the preset's 7.x copy), and the extra preset must not be applied under Jest, where the `hermes-v0` profile already handles everything.

### Environment

```
    System:
      OS: macOS 26.5.2
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 24.11.1
      Yarn: 4.12.0
      npm: 11.13.0
      Watchman: 2025.11.10.00
    Managers:
      CocoaPods: 1.16.2
    SDKs:
      iOS SDK:
    IDEs:
      Android Studio: 2025.1
      Xcode: 26.6/17F113
    npmPackages:
      expo: ~57.0.4 => 57.0.4
      react: 19.2.3 => 19.2.3
    npmGlobalPackages:
      eas-cli: 18.13.0
    Expo Workflow: managed
```

Gitarcitano:
# Why

Resolve #47722

The Babel preset enables decorators lazily when detecting decorator-like syntax, which works out as overall cheaper. However, not all engine presets enable class transforms (properties, private properties, private methods), which the decorators preset relies on.

This specifically breaks decorators when used with the Hermes v1 or Modern Web engine presets.

# How

- Add lazy plugin variants of class properties, private properties, private methods; additionally gated by an AST check
- Apply decorators plugins with the three class transforms as a set, as needed
- Update main preset to select these sub-options as needed

# Test Plan

- Preset capturing snapshot with plugin set updated
- Ensure web preset doesn't pull in general private props/methods plugins, as those are already added

# Checklist

- [x] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: updated a module plugin). (No config plugin or native changes; the transform behaves identically in local and EAS bundling.)
- [ ] Conforms with the [Documentation Writing Style Guide](https://github.com/expo/expo/blob/main/guides/Expo%20Documentation%20Writing%20Style%20Guide.md) (no `docs/` changes)


packages/babel-preset-expo/src/plugins/lazy-decorators-plugin.ts
```diff
@@ -54,7 +51,7 @@ export function lazyDecoratorsPlugin(
   }
 
   return {
-    name: 'expo-lazy-decorators',
+    name,
     inherits: realPlugin.inherits,
     pre(file) {
       this.decoratorsDetected = DECORATOR_PATTERN.test(file.code);
```
kitten: This isn't quite correct in the context of this branch, since we're doing a heuristic `@\w+` check here. Basically, any time something looks like a decorator, we activate the decorators plugin, since it's expensive.

However, activating the other class plugins anytime something looks like a decorator isn't correct. `@\w+` can occur in JSDoc/TSDoc comments too (or otherwise). Meaning, if anything _looks_ like a decorator, rather than just trying to find and transform decorators, we instead also transform classes and related patterns.

That basically makes output unstable based on the heuristic, since the transforms themselves are unrelated

packages/babel-preset-expo/src/index.ts
```diff
@@ -166,6 +166,15 @@ function babelPresetExpo(api: ConfigAPI, options: BabelPresetExpoOptions = {}):
...
       : platformOptions.enableBabelRuntime;
 
+  // Whether the engine preset selected below already compiles class features (class
+  // properties, private methods): `hermes-v0` and the webview preset do, while the
+  // `hermes-v1` and web presets preserve them natively.
+  const engineSuppliesClassFeatures =
+    isDomComponent ||
+    (!isModernEngine &&
+      platformOptions.unstable_transformProfile !== 'hermes-stable' &&
+      platformOptions.unstable_transformProfile !== 'hermes-canary');
```
kitten: nit: This isn't the best way of doing this imho. It'd be better to couple this to the second preset below, since there, we already know if we're applying the plugins, since we know it's only contained within `hermes-v0` and that's the main exception.

Hence, we know, that when the `case` triggers, we need to enable this condition.

kitten (commented):
This isn't quite the right fix, since we shouldn't trigger unrelated transforms on the heuristic of decorators. However, until decorators are by default disabled, it's not safe for us to assume we can run them non-lazily (since the plugin was simply too expensive — which can be repro'd using the profiler script in this package's scripts folder)

I'd prefer a narrower transform that targets the exact failure case without re-transforming class properties/methods/etc via the dedicated transforms. Babel has a requirement on this that we can more cleanly circumvent.

I suspect that requirement is in place exactly because this is a legacy transform and abandoned proposal, so it hasn't been adapted.

We can either re-target on the decorator plugin's `_initializerWarningHelper` additions (like, I believe, it might do itself when the other transforms are present)

Or, alternatively, we need to add an actual `path.traverse` on `Program` that checks `ClassProperty` and `ClassPrivateProperty` for decorators. When detected, it'd then escalate to the case you have today, to narrow these transforms to when there's actually a decorator on a class property. That makes the heuristic narrowed again.

(Edit: to show a quick draft of this: https://gist.github.com/kitten/a2c5a20ee1ab574800695f21fca33ebb [llm derived, but just to demo the principle])

Either works, but if you'd like us to take this over, let me know

ide:
Taking a step back, it seems to me to be in the best interest of WatermelonDB, MobX, etc. to publish code without legacy decorator syntax, and for Expo to define a contract of what code is accepted - which IMO should be a subset of ratified JS features and optional opt-ins for earlier TC39 proposals with the understanding that support is not stable.

If there are a small enough number of libraries affected, it seems tractable to provide clear instructions for those library maintainers, or to hardcode those library names in babel-preset-expo, similar to how programs like Windows and Safari/WebKit will detect a small list of specific apps and websites and run workarounds.

packages/babel-preset-expo/src/plugins/lazy-decorators-plugin.ts
Gitarcitano: Good catch, gating on @\w was too broad, a JSDoc tag would end up downleveling every class field in the file. Ended up using the traversal approach from your gist: the regex is just a fast path now, and the transforms only kick in when a ClassProperty/ClassPrivateProperty actually has decorators. Decorated methods and class-level decorators don't need the transforms, so I scoped detection to properties only. Fixed in [1c02091](https://github.com/expo/expo/pull/47724/commits/1c020911db3b2bca1780e57f0e19775f63783428) and added regression tests for the JSDoc case.

packages/babel-preset-expo/src/index.ts
Gitarcitano: The flag is now returned by the same block that selects the engine preset, so there's a single source of truth on the commit [1c02091](https://github.com/expo/expo/pull/47724/commits/1c020911db3b2bca1780e57f0e19775f63783428)

Gitarcitano:
> Taking a step back, it seems to me to be in the best interest of WatermelonDB, MobX, etc. to publish code without legacy decorator syntax, and for Expo to define a contract of what code is accepted - which IMO should be a subset of ratified JS features and optional opt-ins for earlier TC39 proposals with the understanding that support is not stable.
> 
> If there are a small enough number of libraries affected, it seems tractable to provide clear instructions for those library maintainers, or to hardcode those library names in babel-preset-expo, similar to how programs like Windows and Safari/WebKit will detect a small list of specific apps and websites and run workarounds.

@ide 
Agreed on the long-term direction. A contract of ratified JS plus explicit opt-ins for older proposals would be healthier.

One nuance: here the decorated code isn't in the libraries' published output, it's in application code. WatermelonDB models and MobX 5 stores are written by the app developer using `@field`/`@observable`, that's those libraries documented API. So publishing compiled code or hardcoding library names wouldn't cover the main case. The opt-in would need to be app-level, which is what disabling legacy decorators by default (with an explicit flag) would give us.

Until that default changes, though, the preset ships with legacy decorators enabled, so IMO the default output needs to run. That's what this PR guarantees, without touching undecorated code on the modern profiles.

kitten (approved):

PR #46431, opened 2026-05-30
[cli][metro-file-map] Preserve Watchman default for file map

maxvaljan:
# Why

Fixes #46429.

Expo SDK 56 intends to use Metro's `resolver.useWatchman` third state: `null | undefined` should still allow Watchman, while avoiding Metro's native `find` crawler path when Watchman is unavailable.

That state is currently lost in two places:

- `createFileMap-fork.ts` passes `config.resolver.useWatchman ?? false`
- `@expo/metro-file-map` stores `options.useWatchman ?? false`

As a result, the `null` value set by `@expo/metro-config` is collapsed to `false`, so projects that previously would have used Watchman now crawl with Node and watch with `NativeWatcher` on macOS. In Bun workspaces this can cause missed Fast Refresh updates and noisy refresh behavior.

# How

- Pass `resolver.useWatchman` through from Expo CLI to `@expo/metro-file-map` without nullish coalescing.
- Restore Metro's file-map semantics in the Expo fork: `useWatchman == null ? true : useWatchman`.
- Add tests for both the CLI handoff and `@expo/metro-file-map` behavior.

# Test Plan

- `mise x node@22.14.0 -- npm exec --yes pnpm@10 -- pnpm --filter @expo/cli test -- src/start/server/metro/__tests__/createFileMap-fork.test.ts`
- `mise x node@22.14.0 -- npm exec --yes pnpm@10 -- pnpm --filter @expo/metro-file-map test -- src/__tests__/index.test.ts`
- `mise x node@22.14.0 -- npm exec --yes pnpm@10 -- pnpm --filter @expo/cli typecheck`
- `mise x node@22.14.0 -- npm exec --yes pnpm@10 -- pnpm --filter @expo/metro-file-map typecheck`
- `mise x node@22.14.0 -- npm exec --yes pnpm@10 -- pnpm --filter @expo/cli build`
- `mise x node@22.14.0 -- npm exec --yes pnpm@10 -- pnpm --filter @expo/metro-file-map build`

PR #47726, opened 2026-07-12
fix(metro-config): Preserve explicit `resolver.useWatchman: true`

### LINKED ISSUE #47662: resolver.useWatchman: true can never re-enable Watchman on SDK 56 — rewritten to null by metro-config, then coerced to false by the CLI

https://github.com/jumptransit/expo-watchman-repro

### Steps to reproduce

The bug is in the dev server's file-map configuration, so it is platform-independent; observed on Windows 11 with npm and Watchman 20250223.093309.0.

1. Install Watchman and confirm `watchman version` works.
2. `npm install`
3. `watchman watch-del-all` (start from a clean state)
4. `npx expo start`
5. In another terminal, run `watchman watch-list`.

**Expected:** The project root is watched. The repro's `metro.config.js` sets `config.resolver.useWatchman = true`, which the [SDK 56 changelog](https://expo.dev/changelog/sdk-56) documents as the way to opt back in: *"You can switch back to Watchman with `resolver.useWatchman` in a Metro config, but it's no longer recommended."*

**Actual:** `"roots": []` — Watchman is never contacted. Metro silently uses the Node watcher; nothing is logged (the disabled state only surfaces in internal perf-logger annotations).

**Control experiment:** in `metro.config.js`, replace `useWatchman = true` with `useWatchman = 1` (truthy, but not `=== true`) and restart. `watchman watch-list` now shows the project root. So the Watchman capability check and `WatchmanWatcher` both work — only the configured value is being lost in config plumbing.

### Root cause

The `true` value is destroyed in two steps (verified on `expo@56.0.15` / `@expo/cli@56.1.19` / `@expo/metro-config@56.0.16` / `@expo/metro-file-map@56.0.3`, and still present on `main`):

1. `@expo/metro-config` `loadUserConfig` rewrites `useWatchman === true` to `null` — the comment above it assumes upstream Metro semantics, where `null` still enables Watchman:
   https://github.com/expo/expo/blob/416d06128a27fba9a89868b69c12d0f8227d83e8/packages/%40expo/metro-config/src/config/loadUserConfig.ts#L41-L42
2. `@expo/cli`'s file-map fork then coerces `null`/`undefined` to `false`:
   https://github.com/expo/expo/blob/416d06128a27fba9a89868b69c12d0f8227d83e8/packages/%40expo/cli/src/start/server/metro/createFileMap-fork.ts#L99
   and `@expo/metro-file-map` does the same in its constructor:
   https://github.com/expo/expo/blob/416d06128a27fba9a89868b69c12d0f8227d83e8/packages/%40expo/metro-file-map/src/index.ts#L325

Net effect: `true` → `null` → `false`. No value a user can set in `resolver.useWatchman` enables Watchman, except a truthy non-`true` value (e.g. `1`) that happens to slip past both checks.

To be clear, this is **not** about the SDK 56 default. Defaulting to the Node watcher is documented and intentional (also confirmed by @kitten in #46429). This report is only that the documented opt-in cannot work. It also looks distinct from the closed draft PR #46431, which changed the default back to Watchman; a fix here could keep the Node-watcher default — e.g. dropping the `true → null` rewrite in `loadUserConfig` would let an explicit `true` survive `?? false` while leaving unset configs on the Node watcher.

Why the opt-in matters (motivation, not part of the bug): on Windows the Node watcher opens one OS handle per watched directory. In a monorepo this exhausts the handle pool (EMFILE), and directories deleted mid-build (e.g. Gradle output) crash the watcher with ENOENT. Watchman avoids both with a single recursive watch per root, so it remains the only workable watcher for some Windows monorepo setups.

### Environment

```text
    System:
      OS: Windows 11 10.0.26200
    Binaries:
      Node: 22.22.2 - C:\Users\Justin\scoop\apps\nodejs22\current\node.EXE
      Yarn: 4.17.0 - C:\Users\Justin\scoop\apps\nodejs22\current\yarn.CMD
      npm: 10.9.7 - C:\Users\Justin\scoop\apps\nodejs22\current\npm.CMD
      Watchman: 20250223.093309.0 - C:\Users\Justin\scoop\shims\watchman.EXE
    IDEs:
      Android Studio: AI-253.30387.90.2532.14935130
    npmPackages:
      expo: ^56.0.0 => 56.0.15
      react: 19.2.3 => 19.2.3
    Expo Workflow: managed
```

### Expo Doctor Diagnostics

```text
```

isaka1022:
# Why

Fixes #47662.

On SDK 56 the documented opt-in for Watchman is broken: setting
`config.resolver.useWatchman = true` in `metro.config.js` never re-enables
Watchman. The [SDK 56 changelog](https://expo.dev/changelog/sdk-56) states this
is the supported way to switch back ("You can switch back to Watchman with
`resolver.useWatchman` in a Metro config"), but no value a user can set actually
turns it on — Watchman is silently never contacted (`watchman watch-list` shows
no roots) and Metro falls back to the Node watcher with nothing logged.

This is **not** about the SDK 56 default. Defaulting to the Node watcher is
intentional and documented; this PR keeps that default. The only bug is that the
explicit opt-in is unreachable, which matters for Windows monorepos where the
Node watcher opens one OS handle per directory and exhausts the handle pool
(EMFILE), and where directories deleted mid-build (e.g. Gradle output) crash the
watcher with ENOENT. Watchman uses a single recursive watch per root and avoids
both, so it is the only workable watcher for some setups.

# How

Two facts collide:

- Metro's base config defaults `resolver.useWatchman` to `true`, while Expo
  CLI's `@expo/metro-file-map` fork coalesces a nullish value to `false`
  (`useWatchman: config.resolver.useWatchman ?? false` in
  `createFileMap-fork.ts`, and `options.useWatchman ?? false` in the file-map
  constructor). The Node-watcher default therefore relies on `useWatchman`
  reaching the file map as a nullish value.
- To bridge that, `loadUserConfig` rewrote the **merged** config's
  `useWatchman === true` to `null`. But post-merge, Metro's default `true` and
  a user's explicit `true` are indistinguishable, so the rewrite also destroys
  the documented opt-in: `true → null → ?? false → false`, and
  `#shouldUseWatchman()` never runs. (The reason a truthy non-`true` value like
  `1` "works" in the issue's control experiment is that it slips past the
  `=== true` check and survives `?? false`.)

The fix neutralizes Metro's default at its entry point instead of rewriting the
merged value: `getDefaultConfig` now sets `resolver.useWatchman = null` on the
default config it returns, and the post-merge rewrite in `loadUserConfig` is
removed. This must happen inside `getDefaultConfig` (not pre-merge in
`loadUserConfig`), because the typical `metro.config.js` passes
`getDefaultConfig(__dirname)` through as the user config, which would otherwise
re-introduce Metro's `true` as an apparent user setting.

Resulting behavior:

- Unset (typical template config) → `null` → `?? false` → Node watcher stays
  the default (unchanged).
- Explicit `resolver.useWatchman = true` → survives the merge → Watchman is
  enabled (the documented opt-in works again).
- Explicit `false` → `false`, unchanged.

Two deliberate choices, both informed by earlier attempts:

- `null` rather than `undefined`/deletion: for consumers on upstream Metro's
  file map, `null` still opts into Watchman while bypassing the slower native
  "find" crawler codepath (the same reason the original rewrite chose `null`),
  so non-Expo-CLI consumers of `getDefaultConfig` keep working Watchman.
- The value is assigned at runtime rather than typed as a property, so the
  inferred return type of `getDefaultConfig` is unchanged. #43251 made a
  similar change via a typed property and was reverted in #43319 because the
  type change broke `expo-doctor`'s type checks; this approach avoids that.

Unlike the earlier closed draft #46431, this does not change the default back
to Watchman — the Node watcher remains the default.

# Test Plan

## Unit tests (red/green)

Added a `useWatchman` test to
`packages/@expo/metro-config/src/__tests__/ExpoMetroConfig-test.ts` and a new
`packages/@expo/metro-config/src/config/__tests__/loadUserConfig.test.ts`, and
verified them red/green against the fix.

Red on `main`'s code (only the two bug-pinning tests fail; the failures also
demonstrate both halves of the root cause):

```
● getDefaultConfig › neutralizes Metro's `useWatchman` default so the Node watcher is used by default
  expect(received).toBeNull()
  Received: true        ← Metro's default leaks through getDefaultConfig

● loadUserConfig › preserves an explicit `resolver.useWatchman: true` so Watchman can be re-enabled
  expect(received).toBe(expected)
  Expected: true
  Received: null        ← the post-merge rewrite destroys the opt-in

Tests: 2 failed, 8 passed, 10 total
```

Green on the fix:

```
$ npx jest src/__tests__/ExpoMetroConfig-test.ts src/config/__tests__/loadUserConfig.test.ts
Tests:       10 passed, 10 total
```

The rest of the package's jest suite is unchanged by this diff (identical
pass/fail sets with and without the change applied). `tsc -p tsconfig.json
--noEmit` → 0 errors; `oxlint` on the changed files → no errors or warnings.

## End-to-end against a live Watchman (macOS, Watchman 2026.06.08.00)

Created a fresh SDK 56 project (`npx create-expo-app`), started
`npx expo start --clear`, and inspected `watchman watch-list` after startup
(each run started from an empty watch list via `watchman watch-del-all`).

Before (stock `@expo/metro-config` 56.0.16), with the documented opt-in
`config.resolver.useWatchman = true`, Watchman is never contacted — this is the
bug:

```
=== watchman watch-list (BEFORE, useWatchman:true, unpatched) ===
{ "version": "2026.06.08.00", "roots": [] }
```

After (this fix applied to the installed package), the same config enables
Watchman:

```
=== watchman watch-list (AFTER, useWatchman: true) ===
{ "version": "2026.06.08.00", "roots": ["/…/watchman-e2e"] }
```

After, with the unmodified template config (`useWatchman` unset), the Node
watcher remains the default and Watchman is not contacted:

```
=== watchman watch-list (AFTER, default template config) ===
{ "version": "2026.06.08.00", "roots": [] }
```

The default-unset E2E case is also what rules out the simpler pre-merge
variants of this fix: because the template `metro.config.js` passes
`getDefaultConfig(__dirname)` through as the user config, any fix that leaves
`getDefaultConfig` returning `useWatchman: true` re-enables Watchman for every
default project.

# Checklist

- [x] I added a `CHANGELOG.md` entry (`@expo/metro-config`).
- [x] This diff will work correctly for `npx expo prebuild` & EAS Build (no
      native or plugin changes; config-plumbing only).
- [x] Conforms with the Documentation Writing Style Guide.


packages/@expo/metro-config/src/config/loadUserConfig.ts
```diff
@@ -33,14 +33,5 @@ export async function loadUserConfig(params: LoadMetroConfigParams): Promise<Con
       ? resolveBabelrcName(params.projectRoot)
       : undefined;
 
-  // NOTE(@kitten): `useWatchman` is currently enabled by default, but it also disables `forceNodeFilesystemAPI`.
```
kitten, line 36: I forgot to remove this. Happy to merge this, however, the test and the comment are redundant, and mostly an artifact of Claude's thinking output. Can you remove both please, as they add no lasting value?

packages/@expo/metro-config/src/ExpoMetroConfig.ts
```diff
@@ -452,6 +452,19 @@ export function getDefaultConfig(
...
+  // watcher the default watcher. Neutralize Metro's default here, so that only an explicit
+  // `useWatchman: true` set in a user config re-enables Watchman. This must happen here rather
+  // than after the user config is merged, because the typical `metro.config.js` passes this
+  // default config through, which would otherwise re-introduce Metro's `true` as an apparent
+  // user setting. `null` (rather than `undefined`) keeps Watchman enabled for consumers on
+  // upstream Metro's file map, where `null` still opts into Watchman but bypasses the slower
+  // "native find" crawler codepath. The value is assigned at runtime instead of typed as a
+  // property to keep the inferred config type unchanged (see expo/expo#43319).
+  // See: https://github.com/expo/expo/issues/47662
+  (metroConfig.resolver as { useWatchman?: boolean | null }).useWatchman = null;
```
kitten, line 459: To clarify, as per the other comments, this and the comment (plus the test in the separate file) should be removable

isaka1022: Done in 975809d — condensed this to just the non-obvious intent behind the `null` assignment. Thanks for the review!

packages/@expo/metro-config/src/config/loadUserConfig.ts
isaka1022: Done in 975809d — dropped the redundant `loadUserConfig` test file (its logic now lives in `getDefaultConfig`). Kept the single `getDefaultConfig` regression test.

packages/@expo/metro-config/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 - Fix stack frame collapsing for Windows paths. ([#46645](https://github.com/expo/expo/pull/46645) by [@krystofwoldrich](https://github.com/krystofwoldrich))
 - Bump `hermes-parser` to `^0.36.0` to parse newer Flow syntax (e.g. `readonly` property modifiers) shipped in recent React Native versions ([#46636](https://github.com/expo/expo/pull/46636) by [@zoontek](https://github.com/zoontek))
 - Treat dynamic imports with rejection handlers as optional dependencies, ported from [react/metro#1697](https://github.com/react/metro/pull/1697) ([#47334](https://github.com/expo/expo/pull/47334) by [@kitten](https://github.com/kitten))
+- Fix `resolver.useWatchman: true` never re-enabling Watchman: neutralize Metro's `useWatchman` default before merging the user config instead of rewriting the merged value, so an explicit opt-in survives while the Node watcher stays the default. ([#47662](https://github.com/expo/expo/issues/47662) by [@isaka1022](https://github.com/isaka1022))
```
kitten:
```suggestion
- Fix `resolver.useWatchman: true` not re-enabling watchman as intended ([#47662](https://github.com/expo/expo/issues/47662) by [@isaka1022](https://github.com/isaka1022))
```

packages/@expo/metro-config/src/ExpoMetroConfig.ts
```diff
@@ -452,6 +452,10 @@ export function getDefaultConfig(
     expoMetroConfig
   );
 
+  // Neutralize Metro's `useWatchman: true` default to `null` so the `@expo/metro-file-map` fork
+  // keeps the Node watcher by default while an explicit `useWatchman: true` can re-enable Watchman.
```
kitten:
```suggestion
  // NOTE(@kitten): `useWatchman` is currently defaulting to `false`
  // However, in standard Metro, it defaults to `true`, and we must override this to `null` to disable the slow "native find" codepath
  // See: https://github.com/facebook/metro/blob/b9c243f/packages/metro-file-map/src/index.js#L326
  // See: https://github.com/facebook/metro/blob/b9c243f/packages/metro/src/node-haste/DependencyGraph/createFileMap.js#L109
```

kitten (approved):

PR #47737, opened 2026-07-13
[expo-router] Add Agents.md and require et check-packages before commiting

Ubax:
# Why

There is only CLAUDE.md file in expo-router, so other agents may not read it. 

# How

1. Add AGENTS.md with the content of existing CLAUDE.md
2. Reference AGENTS.md in CLAUDE.md
3. Add a new paragraph to the md file - https://github.com/expo/expo/pull/47737/changes/40daec2a1e512a248ff1920548734f2729eb7a40

> **Required:** before committing, run `et check-packages expo-router` to build, type-check, lint, and test the package the same way CI does. See the repo root [`.claude/CLAUDE.md`](../../.claude/CLAUDE.md) for how to use `et` (expotools) and other repo tooling.

# Test Plan

# Checklist


Ubax:
* **#47737** <a href="https://app.graphite.com/github/pr/expo/expo/47737?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/47737?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/expo-router/CLAUDE.md
```diff
@@ -1,355 +1,3 @@
...
-
-- Always use latest React 19 hooks and patterns - `use` instead of `useContext`, `useId`, etc.
-- Make sure the code works with and without React Compiler enabled.
-- Don't use `any` types, unless strictly necessary. Use `unknown` instead and narrow types as much as possible.
-- Never import files with platform-specific extensions directly. Always import from the base path and let the bundler resolve the correct file. Correct `import { Component } from './Component'`, not `import { Component } from './Component.ios'`.
-
-## Maintaining This Document
-
-When developing or planning features, document missing behaviors in this file. Update sections when implementations change or new patterns emerge.
+See [`AGENTS.md`](./AGENTS.md) for guidance on working in this package.
```
chrfalch: Claude Code doesn't follow markdown links in `CLAUDE.md` — it only auto-loads content referenced with the `@` import syntax. As written, the expo-router guidance goes from "always in context" to "read only if the agent decides to open the file", which undermines the new **Required** `et check-packages` gate.

```suggestion
See @AGENTS.md for guidance on working in this package.
```

(A `CLAUDE.md → AGENTS.md` symlink would also work, but symlinks in git are awkward on Windows checkouts, so the `@` import is the safer option.)

chrfalch (changes requested):

packages/expo-router/CLAUDE.md
Ubax: Thanks for letting me know!

chrfalch (approved):
Thanks!

packages/expo-router/AGENTS.md
```diff
@@ -0,0 +1,357 @@
...
+pnpm test
+
+# Run specific test file
+pnpm test src/__tests__/navigation.test.ios.tsx
+```
+
+To verify if the types used in tests are correct, run:
+
+```bash
+pnpm test:types
```
kitten: This doesn't exist anymore

packages/expo-router/AGENTS.md
```diff
@@ -0,0 +1,357 @@
...
+**Conventions:**
+
+- Use `@Test` / `@Suite` from Swift Testing (not XCTest)
+- Backtick-quoted test names for readability (e.g., ``@Test func `converts options correctly`()``)
+- Inner structs for grouping related tests within a `@Suite`
+- `#expect` / `#require` for assertions
+
+### Testing Patterns
+
+Tests use the custom `renderRouter` testing utility:
```
kitten: Are we sure this didn't need a qualifier? ("some tests", "router tests", etc)

packages/expo-router/AGENTS.md
```diff
@@ -0,0 +1,357 @@
...
+
+**Spies and console mocks:** Use `beforeEach`/`afterEach` with `mockRestore()`:
+
+```ts
+let spy: jest.SpyInstance;
+beforeEach(() => {
+  spy = jest.spyOn(Module, 'fn');
+}); // or jest.spyOn(console, 'warn').mockImplementation(() => {})
+afterEach(() => {
+  spy.mockRestore();
```
kitten, line 234: Worth checking (which has been on my list for a while), whether we can change the Jest settings here (if we haven't yet, but I believe we haven't), to do this for all mocks by default, as that's just a better default.

That'd make this note and gotcha redundant

packages/expo-router/AGENTS.md
```diff
@@ -0,0 +1,357 @@
...
+
+**Manual Android testing:** Use the `/android-e2e-testing` skill for step-by-step guidance on testing Expo Router screens on Android emulators using ADB. This covers launching E2E apps, navigating via UI dumps, interacting with the app, and verifying results.
+
+## Verification
+
+After developing a feature, run these commands in `packages/expo-router`:
+
+1. `CI=1 pnpm test` - Run all tests, including the RSC `__rsc_tests__` (they run as the `rsc/<platform>` Jest projects). During development use `pnpm test [test file]` for efficiency, or `pnpm test --selectProjects rsc/web` to run only RSC tests.
+2. `pnpm build` - Build and verify TypeScript correctness. If you moved or deleted files, run `pnpm clean` first.
+3. `pnpm test:types` - Verify type correctness in tests
```
kitten: as before, doesn't exist anymore (did this mean `typecheck`, i.e. running tsc)?

packages/expo-router/AGENTS.md
Ubax: https://linear.app/expo/issue/ENG-24657/enable-automatic-jest-spy-restoration-in-expo-router

Added a ticket for that in router

PR #47745, opened 2026-07-13
[cli] Add dynamic WebSocket upgrades to DevTools plugin request handlers

krystofwoldrich:
## Why

DevTools plugin servers could only open WebSockets at **static, exact-path routes** declared up front through the `webSocketHandlers` export — each route is mounted into Metro's `websocketEndpoints` map before the server starts and matched by literal pathname equality. Dynamic routes (e.g. `/ws/sessions/:id`) were impossible, and the fetch-based request handler had no way to participate in upgrades at all.

This PR lets the plugin's request handler handle WebSocket upgrades on any route, inspired by the [Next.js WebSocket RFC](https://github.com/vercel/next.js/discussions/95514) (crossws-based, "similar API as Bun").

## How

- `runServer-fork.ts` gains a `websocketUpgradeHandler` fallback, called for Upgrade requests that no exact-path endpoint claimed (previously: unconditional `socket.destroy()`).
- The new dispatcher (`DevToolsPluginUpgradeHandler.ts`) resolves the plugin lazily at upgrade time, converts the raw upgrade to a fetch `Request`, and invokes the plugin's request handler with a `context` as second argument:

```js
export default function handler(request, context) {
  const url = new URL(request.url);
  const channel = url.pathname.match(/^\/ws\/(.+)$/)?.[1];
  if (channel) {
    const response = context.upgrade({
      onopen(peer) { peer.send({ hello: channel }); },
      onmessage(peer, message) { peer.send(message.text()); },
    });
    response.headers.set('x-channel', channel); // written into the 101 handshake
    return response;
  }
  return null; // falls through to static webpageRoot serving
}
```

- **Return value is the contract**: `context.upgrade(hooks)` returns a marker `Response` (status 101) that *commits the handshake only when returned*. Returning a plain `Response` (e.g. a 401) rejects the upgrade with that response written raw onto the socket; `null`/`undefined` declines and the socket is destroyed. Headers/cookies set on the upgrade response land on the `101 Switching Protocols` reply.
- The static `webSocketHandlers` path is unchanged; plain HTTP requests get a context whose `upgrade()` throws with guidance.

## Approaches considered

**1. Bun / Vercel (crossws) hooks style — implemented.** `context.upgrade({ onopen, onmessage, onclose, onerror })` where hooks receive a `peer` (send with object→JSON convenience, `close`, `terminate`, raw `ws` socket as escape hatch) and a lazy `message` wrapper (`text()`, `json()`, `uint8Array()`, `arrayBuffer()`, mirroring fetch `Body`). Chosen because:
- the handshake hasn't happened while the handler runs, and hooks make pre-open socket access *unrepresentable* — there is no socket object to misuse before `onopen`;
- commit-on-return keeps one exhaustive contract (`Response | null`) for HTTP and WS alike, and rejection is just a normal `Response`;
- it matches the Next.js RFC and Bun, so the shape is familiar and portable.

**2. Deno style (`const { socket, response } = context.upgrade()`).** Returns a standard-flavored socket synchronously in `CONNECTING` state; the handshake commits when `response` is returned and the facade binds to the real socket. Browser-symmetric (`onopen`/`onmessage` like client code), but it forces a facade with pre-open semantics (`send()` must throw before open, buffered `close()`, `readyState` bookkeeping), and the facade type is either a thin subset (consumers lose `ping`/`pong` liveness, `terminate`, multi-listener registration, backpressure signals) or a full `ws`-compatible implementation — prototyped at ~290 lines, ~3× the hooks implementation, with half the code servicing the dual browser-compat event API.

Not addressed in either variant: `Sec-WebSocket-Protocol` negotiation (a client requesting a subprotocol fails the handshake — would need a `handleProtocols` option at upgrade time).

## Test Plan

- unit, e2e tests

https://claude.ai/code/session_013qcxr3fWQXxDEft4PUpkwM

### DIFF
--- a/packages/@expo/cli/src/start/server/DevToolsPluginSocketHelpers.ts
+++ b/packages/@expo/cli/src/start/server/DevToolsPluginSocketHelpers.ts
@@ -1,0 +1,200 @@
+import { loadModule } from '@expo/require-utils';
+import type { IncomingMessage } from 'node:http';
+import { type RawData, type WebSocket, WebSocketServer } from 'ws';
+
+const debug = require('debug')('expo:start:server:devtools') as typeof console.log;
+
+/**
+ * Incoming WebSocket message passed to the `message` hook, with lazy conversion helpers
+ * mirroring the fetch API `Body` readers.
+ */
+type WebSocketMessage = {
+  readonly rawData: RawData;
+  readonly isBinary: boolean;
+  text(): string;
+  json<T = unknown>(): T;
+  uint8Array(): Uint8Array;
+  arrayBuffer(): ArrayBuffer;
+};
+
+/**
+ * Connected WebSocket peer passed to the lifecycle hooks. `send` serializes plain objects
+ * as JSON; strings and binary payloads are sent as-is. The raw `ws` socket is exposed as
+ * an escape hatch.
+ */
+type WebSocketPeer = {
+  readonly socket: WebSocket;
+  send(data: string | Buffer | ArrayBufferLike | ArrayBufferView | object): void;
+  close(code?: number, reason?: string): void;
+  terminate(): void;
+};
+
+/** Lifecycle hooks passed to `context.upgrade()`, wired to the socket once the handshake commits. */
+type WebSocketHooks = {
+  onopen?(peer: WebSocketPeer): void;
+  onmessage?(peer: WebSocketPeer, message: WebSocketMessage): void;
+  onclose?(peer: WebSocketPeer, details: { code: number; reason: string }): void;
+  onerror?(peer: WebSocketPeer, error: Error): void;
+};
+
+/**
+ * Per-request context passed to a plugin's request handler as the second argument.
+ * `upgrade()` does not perform the WebSocket handshake — it returns a marker `Response`
+ * that commits the handshake only when the handler returns it. Headers set on that
+ * response (including cookies) are written into the `101 Switching Protocols` handshake.
+ */
+export type RequestContext = {
+  upgrade(hooks: WebSocketHooks): Response;
+};
+
+const upgradeHooksRegistry = new WeakMap<Response, WebSocketHooks>();
+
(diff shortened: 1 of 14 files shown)


packages/@expo/cli/src/start/server/DevToolsPluginSocketHelpers.ts
```diff
@@ -0,0 +1,200 @@
...
+    }
+  };
+  socket.on('message', (data, isBinary) =>
+    invoke(() => hooks.onmessage?.(peer, createWebSocketMessage(data, isBinary)))
+  );
+  socket.on('close', (code, reason) =>
+    // https://www.rfc-editor.org/info/rfc6455/#section-5.5.1 -> defined UTF-8-encoded data
+    invoke(() => hooks.onclose?.(peer, { code, reason: reason.toString() }))
+  );
+  socket.on('error', (error) => hooks.onerror?.(peer, error));
```
Kudo, line 132: [💡 suggestion] A throwing `onerror` hook isn't isolated here. `onmessage`/`onclose`/`onopen` all run through `invoke`, but a socket `'error'` event calls the plugin's `onerror` directly — so if that hook throws, it surfaces as an uncaught exception on the socket's error listener and can take down the dev server. Give it its own try/catch (not `invoke`, which would re-call `onerror` on throw):

```suggestion
  socket.on('error', (error) => {
    try {
      hooks.onerror?.(peer, error);
    } catch (hookError) {
      debug('DevTools plugin WebSocket onerror hook failed: %O', hookError);
    }
  });
```

Kudo (approved):

packages/@expo/cli/src/start/server/DevToolsPluginSocketHelpers.ts
krystofwoldrich: thanks, make sense. will add it.

PR #47750, opened 2026-07-13
feat(cli,metro-config,babel-preset): Implement cache-vary capturing in transformer to selectively redirect cache entries

kitten:
# Why

We currently have really narrow and few strategies to deal with impure changes on the transformer. The cache considers the input coupled to the output purely. The input considers a cache key (mostly made up of a predetermined set of file inputs, and config shape) as an additional hash discriminator.

This means that we always have a choice of:
- discriminating the cache/transform purity by adding a Metro config entry
- pushing impure modifications to the uncached serializer

However, even if we cleanly did both of these refactors above, sometimes this isn't ideal or possible. The best (and currently most important) example are arbitrary envirionment variables.

Per module, we cannot predict which values are important per-module, however the `babel-preset-expo`'s inliner means that module outputs become impure.

This PR implements a **cache-vary feature** on top of Metro to address this.

# How

The cache-vary strategy is split into three parts:
1. CLI patches for Metro in `@expo/cli`
2. The transform-worker outputs and the `VaryingCacheStore` in `@expo/metro-config`
3. The cache vary entry additions in `babel-preset-expo`

This feature can be disabled using `EXPO_NO_CACHE_VARY=1`

The CLI makes three distinct changes:
- the cache store(s) are wrapped in the `VaryingCacheStore`
- the `transformFile` method modifies the `unstable_transformResultKey` returned entry to add cache-vary hashes (and track cache-vary dims)
- the `getDelta` method is modified to mark modules as invalidated when their dependent ambient values have changed

This cascades upwards in reverse; on a delta update (HMR), `getDelta` is consulted and checks the dims that have been tracked by the previous build's `transformFile` calls. It then notices changes to the ambient values and marks modules as invalidated. It re-checks `transformFile`, and since the `unstable_transformResultKey` changes, the cache is re-consulted. The cache store forks by ambient cache-vary values using the `VaryingCacheStore`.

The `VaryingCacheStore` itself has a simple 3 option output that minimises overhead. It assumes that the **cache-vary dimensionss per module are stable** (purity of module to vary dims).
- If there are no cache-vary dims in a module, an entry is still the raw cache entry, as it was before (zero overhead).
- If the cache entry is replaced by a "head value", the target variant is computed with the cache-vary entries against the ambient values
- If the cache entry is a "head value" with varying cache-dims (which breaks the assumption) it checks multiple in order of specificity, basically

The two most common cases keep overhead low, and the performance impact should be minimal.

The changes in `babel-preset-expo` only currently address the `env` scheme for cache-vary dimensions:
- For production env inlining of `EXPO_PUBLIC_*` variables, in the `inline-env-vars-plugin`, a cache-vary marker is added
- For inlining of `EXPO_PUBLIC_*` variables in the `define-plugin`, a cache-vary marker is added too (this only affects `EXPO_PUBLIC_USE_RN_FETCH` currently)

The cache-vary markers are returned via the Babel metadata in a `cacheVary` property. This is computed into a `expoCacheVary` output on our Metro transform worker. That's the vary output that's used by all the other logic above.

> [!WARNING]
> The implementation in `babel-preset-expo` (and of cache-vary schemes) isn't complete, but has reached a nice stopping point. Currently, we do not synchronise environment variables from the main thread to the transform worker threads. This will be done separately. Meaning, in development, when `.env` changes, these changes aren't synchronised to the transform-workers.

**Note:** We should apply this change to other Babel plugins where appropriate in subsequent PRs. This might also allow us to remove `ExpoMetroConfig` entries that were only added for cache invalidation (e.g. `expo-router` base URL or root app-dir path).

# Test Plan

- Unit tests added
- E2E test added
- Manually verified against `expo export`

# Checklist

### DIFF
--- a/apps/router-e2e/__e2e__/cache-vary/app/index.js
+++ b/apps/router-e2e/__e2e__/cache-vary/app/index.js
@@ -1,0 +1,5 @@
+import { Text } from 'react-native';
+
+export default function Page() {
+  return <Text testID="env-value">{`env-value:${process.env.EXPO_PUBLIC_CACHE_VARY_VALUE}`}</Text>;
+}
--- a/packages/@expo/cli/CHANGELOG.md
+++ b/packages/@expo/cli/CHANGELOG.md
@@ -23,6 +23,7 @@
 
 ### 🐛 Bug fixes
 
+- Fix cache pollution in production by inlined environment variable values with a vary-cache approach ([#47750](https://github.com/expo/expo/pull/47750) by [@kitten](https://github.com/kitten))
 - Fail when `--private-key-path` is passed without `updates.codeSigningCertificate` in the resolved app config, instead of ignoring the flag and continuing without signing.
 - Show the Xcode build log path when `run:ios` fails. ([#48624](https://github.com/expo/expo/pull/48624) by [@ramonclaudio](https://github.com/ramonclaudio))
 - [Internal] Fix `LogStream.destroy()` racing a pending write and dropping log data ([#47181](https://github.com/expo/expo/pull/47181) by [@kitten](https://github.com/kitten))
@@ -38,6 +39,7 @@
 
 ### 💡 Others
 
+- Add `VaryingCacheStore` and embed `expoCacheVary` fingerprints into transform results so a cache output never holds on to stale values inline ([#47750](https://github.com/expo/expo/pull/47750) by [@kitten](https://github.com/kitten))
 - Add sandbox detection to telemetry context ([#47928](https://github.com/expo/expo/pull/47928) by [@davidmokos](https://github.com/davidmokos))
 - [Internal] Remove the unreachable port fallbacks and increase consistency in port selection logic ([#47771](https://github.com/expo/expo/pull/47771) by [@kitten](https://github.com/kitten))
 - Add experimental `tvos` and `macos` autolinking gated by `expriments.outOfTreePlatforms` ([#46344](https://github.com/expo/expo/pull/46344) by [@kitten](https://github.com/kitten))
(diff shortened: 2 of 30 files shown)


hassankhan (approved):

robhogan:
High level thoughts - "dealing with impure changes" / "module outputs become impure" shouldn't be the framing, that's build system nightmare fuel. The way I'd frame this is we want to provide the inputs needed safely, and then additionally, we want to know which inputs have been used (the vary system) so that we're not combinatorially exploding the cache.

Whatever the approach is should be eventually (after upstreaming) extensible to the inputs we currently expose but which rarely actually change the output - `platform` is the big one - it'd be great if iOS and Android (at least) could share most of their transformed modules.

On the consumer side, I'd aim for something safe by construction - rather than plugins declaring that they've used something out of band, they should be required to read the needed value in a way we can observe. Something in the shape of Babel's `api.caller()`

So rather than:
```js
addCacheVary(state, { scheme: 'env', name: 'MYVAR' });
if (process.env.MYVAR) {}
else if (process.env.SOMETHING_ELSE_I_FORGOT_TO_DECLARE) {}
```
I'd either go explicit, and ban `process.env` reads:
```js
if(api.getEnv('MYVAR')){}
```
or do the same thing with a proxy object substituting `process.env`, which has the advantage of the plugin not needing to know about Expo/Metro, and works as long as module transforms aren't overlapping within a worker.

All that said, we can iterate on `main` as long as we're not publicising `addCacheVary` yet.

kitten:
> they should be required to read the needed value in a way we can observe

Agreed, I deferred this and didn't abstract further since this branch was made in parallel with noxcturnal, and I didn't have that shape nailed down yet. The other reason I've deferred this is that it's unclear what the `@expo/config` shape here will look like (for `expo-constants`' expo config embedding). There's a chance that one will have to share the value via a temporary file, I'm guessing.

Since that means we currently only abstract over env vars, I didn't yet abstract this, and left that for a follow-up PR, but agreed that the final shape should have a side-effectful read that records the use, since that basically makes mistakes almost impossible.

> The way I'd frame this is we want to provide the inputs needed safely

I largely agree with this, but would add a caveat (which partially connects to why I've phrased it this way for now in the explanation). The problem here is that there's an additional constraint in that we don't necessarily will know what values to pass. There's a few cases that will potentially become tricky:
- We're updating dotenv values (via `@expo/env`), which means, we'll have to keep the workers in sync with changing env vars, if we want to preserve this
- However, sending all values/inputs ahead of time or alongside transformers means we either have to know all used values (which we don't, as in, no fixed arity, number of inputs is virtually unlimited)

This means it's slightly phrased this way since I'm expecting that either:
- The "ambient" values are updated proactively by the main thread as needed, and shared, or we have another system in parallel to transformers to keep them in sync (but the transformer calls are too constrained to make that a parallel call, so would have to use files or another sync mechanism, or no sync at all)
- They're impure in that there's a chance that values like `@expo/config` may actually end up executing changes. It'd be great to avoid that, but conceptually I'm thinking of that as the transformer's ambient values happening to match the main thread's one. That's just the mental model until we constrain this more

kitten:
Note on outstanding comments. Follow-up PRs will change and refine the internal implementation and expand on the cache-vary implementation's scope

PR #47755, opened 2026-07-13
[iOS][JSI] Make async host functions two-phase so no JSI values cross the async boundary

tsapeta:
# Why

#47716 was fixed at the `expo-modules-core` layer by #47717, which stopped `ConcurrentFunctionDefinition` from sending JSI-owned wrappers into the scheduled async task. The underlying gap remained in `expo-modules-jsi` though: `createAsyncFunction` (public API) still copied `this` and the arguments buffer into the task it schedules, so any consumer of that API kept the same use-after-free when the runtime is reloaded while calls are queued or suspended. This PR fixes it at the source, as the API design rather than a call-site workaround.

# How

**Two-phase `AsyncFunctionClosure`.** The closure now runs synchronously within the host function call, with the same parameter shape as sync closures (a borrowed unowned `this` and a consumed arguments buffer), and returns the async body (`AsyncFunctionBody`): sync and async closures differ only in what they return. Decoding happens while the JS caller's values are still valid, and only decoded values cross the asynchronous boundary, so an abandoned task frame holds nothing to destroy against a freed runtime. (The buffer is a non-owning view valid only for the duration of the host call; moving it into the returned body is documented as forbidden.) This also removes the per-call `jsi::Value` copies of the arguments and the owning-`this` allocation, and an error thrown before the first suspension now rejects the returned promise, matching JavaScript's async-function semantics.

**`ConcurrentFunctionDefinition` builds on `createAsyncFunction` again.** The decode phase promotes `this` (only when the function takes an owner) and converts arguments to native values; the returned body invokes the function's implementation. This drops the promise wiring the core had to duplicate in #47717 and restores rejecting the returned promise (instead of throwing synchronously) when the app context is lost or decoding fails.

**`expo-modules-macros` 0.8.0.** The macros release that emits the two-phase shape ships separately (expo/expo-modules-macros-plugin#35); this PR bumps `expo-modules-core`'s plugin pin to it, so `@JS` async bindings generated in this repo use the new shape. Since nothing generates the previous owned-values async closure anymore and none of this has shipped, no compatibility overload is kept.

# Test Plan

- `expo-modules-jsi` test suite: 566 tests in 30 suites pass, including a regression test that emulates a dying React scheduler with a task-holding `dispatch` trampoline: an async call is scheduled, the runtime is destroyed, and the dropped task is released afterwards, which crashed in `JavaScriptValuesBuffer.deinit` before this change.
- `et native-unit-tests -p ios --packages expo-modules-core`: all suites pass, including the macro tests whose bindings are generated by the 0.8.0 plugin in the two-phase shape, and the function-definition suites exercising the migrated `ConcurrentFunctionDefinition`.
- Manual end-to-end repro of #47716 on the NCL "Runtime teardown" screen (#47776): a batch of suspended async module calls holding large arguments while the runtime reloads. Reloads cleanly with this change; crashes in `JavaScriptValuesBuffer.deinit` on a Release device build without it.

- New `addStringsAsync(a, b)` benchmark on the NCL benchmarks screen (`AsyncFunction` and `@JS` rows; strings are where the removed per-call `jsi::Value` argument copies actually cost something). Measured on iOS with the two-phase path and the migrated macro codegen: `AsyncFunction` 25 819 ns/op, `@JS` 11 020 ns/op (100k iterations). The remaining gap between the rows is the `[Any]`/`toTuple` dynamic conversion machinery that `ConcurrentFunctionDefinition` still uses, which the macro bindings skip.


packages/expo-modules-jsi/apple/Sources/ExpoModulesJSI/Runtime/Values/JavaScriptObject.swift
```diff
@@ -299,6 +300,18 @@ public struct JavaScriptObject: JavaScriptType, Sendable, ~Copyable {
...
+  /// ``JavaScriptRuntime/LegacyAsyncFunctionClosure``. A transitional overload kept for
+  /// compatibility with code generated by the current `expo-modules-macros` release; the `async`
+  /// closure body selects it. See ``JavaScriptRuntime/createAsyncFunction(_:_:)-(String,LegacyAsyncFunctionClosure)``.
+  @JavaScriptActor
+  public func setProperty(_ name: String, function: sending @escaping JavaScriptRuntime.LegacyAsyncFunctionClosure) {
+    guard let runtime else {
+      FatalError.runtimeLost()
+    }
+    setProperty(name, value: runtime.createAsyncFunction(name, function))
+  }
```
tsapeta: I'm going to remove this legacy path once I bump the macros plugin. I don't think we need both, especially that we're the only ones using it.

packages/expo-modules-jsi/apple/Sources/ExpoModulesJSI/Runtime/JavaScriptRuntime.swift
```diff
@@ -392,43 +392,130 @@ open class JavaScriptRuntime: Equatable, Identifiable, @unchecked Sendable {
...
+      longLivedObjects.add(pendingCall)
+
+      return {
+        // The task runs on the JS thread while the runtime is alive, so take the values back out
+        // and deregister the box: a stream of calls must not pin their arguments until teardown.
+        longLivedObjects.remove(pendingCall)
+        return try await function(pendingCall.this.take(), pendingCall.arguments.take())
+      }
+    }
+  }
```
tsapeta: Also to be removed once https://github.com/expo/expo-modules-macros-plugin/pull/35 lands and is bumped in `expo/expo`.

Kudo (approved):

PR #47757, opened 2026-07-13
[docs] Increase SDK 57 support range due to off-cycle release cadence

kitten:
# Why

This was a release that was off-cadence, so SDK 54 should technically still be in range.

Also added a small drive-by fix since we technically didn't bump out of Node 20.19.x for SDK 56. The React Native and Metro teams didn't either. That's because the branch cut date of this SDK and React Native release was 2026-03-02, before LTS ended.

# How

# Test Plan

# Checklist

### DIFF
--- a/docs/ui/components/SDKTables/utils.ts
+++ b/docs/ui/components/SDKTables/utils.ts
@@ -12,7 +12,11 @@ export const getThreeVersions = (currentVersion: string) => {
     return [];
   }
 
-  const endIndex = Math.min(sdkData.sdkVersions.length, currentIndex + 3);
+  // NOTE(@kitten): This is a temporary exception for SDK 57, which keeps SDK 54 in range
+  // as it was an "off-cycle" release
+  const inRange = normalizedVersion === '57.0.0' ? 4 : 3;
+
+  const endIndex = Math.min(sdkData.sdkVersions.length, currentIndex + inRange);
   return sdkData.sdkVersions.slice(currentIndex, endIndex);
 };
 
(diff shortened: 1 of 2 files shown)


brentvatne (approved):

amandeepmittal (approved):

docs/ui/components/SDKTables/utils.ts
```diff
@@ -12,7 +12,11 @@ export const getThreeVersions = (currentVersion: string) => {
     return [];
   }
 
-  const endIndex = Math.min(sdkData.sdkVersions.length, currentIndex + 3);
+  // NOTE(@kitten): This is a temporary exception for SDK 57, which keeps SDK 54 in range
+  // as it was an "off-cycle" release
+  const inRange = normalizedVersion === '57.0.0' ? 4 : 3;
```
chrfalch, line 17: Wouldn't this also happen for `58.0.0` and later since we'll be releasing at a different cadence also in the future?

kitten: If the cadence officially changes, more changes will be needed than this, and we should likely just decide at that time what happens. From my end, I don't think that it's very likely we'd move up to 6 or more releases a year, but if things change, there's likely more updates that need to be done than this per-release tweak.

As for SDK 58, under the current policy, once 58 lands 54 will leave the support window (unless we revise that decision), and we'll have to decide if we're increasing the window to include SDK 55 or not, so the default decision there is still correct, unless we further make another one-off decision for 58 imho

chrfalch (approved):

PR #47764, opened 2026-07-14
[expo-widgets] Fix typecheck

kadikraman:
# Why

Fix typecheck issue in a test introduced in https://github.com/expo/expo/commit/44c4e97fec8151b48956c1cad226f9612657b473

I'm slightly mystified why this didn't show up on the PR or in main after merging it, but I can repro it locally so this should stort it.

# How

Type the mock objects with `InfoPlist` instead of `Record<string, unknown>`.

# Test Plan

Screenshot 2026-07-14 at 10 04 24

# Checklist


chrfalch (approved):
✅

jakex7 (approved):

PR #47769, opened 2026-07-14
fix(cli): reset Metro cache when bundling production apps in local EAS builds

tjzel:
# Why

Local EAS production builds re-use Metro cache, which can lead to invalid bundles being produced on subsequent runs, where the EAS build sits in a new temporary directory.

This was especially breaking Worklets' Bundle Mode, where files are generated in the bundling process, but are skipped that way.

EAS builds don't reset cache by default because they assume a CI run, where the cache isn't persistent anyway.

# How

I'm resetting the Metro cache for local production bundles.

# Test Plan

I checked that running local eas production builds with Worklets in an Expo app doesn't fail in subsequent runs anymore.


packages/@expo/cli/src/export/embed/exportEmbedAsync.ts
```diff
@@ -66,7 +66,12 @@ function guessCopiedAppleBundlePath(bundleOutput: string) {
 export async function exportEmbedAsync(projectRoot: string, options: Options) {
   // The React Native build scripts always enable the cache reset but we shouldn't need this in CI environments.
   // By disabling it, we can eagerly bundle code before the build and reuse the cached artifacts in subsequent builds.
```
kitten, line 68: As per: By disabling it, we can eagerly bundle code before the build and reuse the cached artifacts in subsequent builds.

This breaks other functionality, due to the pre-build nature of `export:embed` being invoked twice, and isn't desirable

tjzel: How about invalidating the cache only for the first invocation? Would that be non-breaking?

tjzel:
@kitten I made it so only the first run of export:embed comes with a cache reset and then for the native build the cache is re-used.

PR #47772, opened 2026-07-14
fix(log-box): Remove deep import warning from web bundle

ramonclaudio:
# Why

Fixes the LogBox web overlay printing a `Deep imports from the 'react-native' package are deprecated` warning.

# How

- Moves the native-only import into a native file and adds a web version.
- Rebuilds the LogBox bundle.

# Test Plan

- `et check-packages @expo/log-box`
- `dev-console-errors` Playwright tests
- Manually verified the warning is gone and the overlay still works.

# Checklist


krystofwoldrich (approved):
Thank you for the improvement! Looks good. 

Let's push the updated bundle so it matched the source in case someone runs the commited bundle without rebuild.

(This has not shown in users apps because warnings from `node_modules` are filtered.)

packages/@expo/log-box/src/utils/getDevServer.ts
```diff
@@ -0,0 +1,15 @@
+export const getDevServer = () => {
+  // Disable for SSR
+  if (typeof window === 'undefined') {
+    return {
+      bundleLoadedFromServer: true,
+      url: '',
+    };
+  }
+
+  return {
```
kitten, line 10: Once merged, we'll probably want to check if we want to do the same URL dance as here: https://github.com/expo/expo/blob/@kitten/refactor/derive-consistent-bundle-url/packages/expo/src/utils/getDevServer.native.ts

Problem with `location.origin`, as far as I recall from this stack of PRs, is that it relies on `@expo/metro-runtime` which in turn is sometimes optional (in practice not always in web bundles). However, given that I'm not 100% confident that this code here isn't reused outside of the bundle (since log-box tended to have multiple entry points), it should probably switch once those PRs are merged as a precaution

ramonclaudio: Ah okay, that makes sense. I’ll circle back on this once those PRs are merged.

PR #47774, opened 2026-07-14
[cli] Create `pageHeaders` rules from loader-declared `Cache-Control` headers for SSG

hassankhan:
# Why

Loaders can declare `Cache-Control` by returning a `Response`, but these were initially ignored in SSG exports as there was no mechanism to preserve headers from build-time evaluation of a `loader()` export. This PR adds that mechanism, building on the `pageHeaders` rules from #47429.

# How

- During pre-rendering, read allowlisted headers (currently `Cache-Control`) from each loader's `Response` and generate two `pageHeaders` rules; one for the page's path and one for the loader data file's path. Both are appended after config rules, so the loader-declared values take precedence over config-declared ones.
- Write the SSG export manifest _after_ pre-rendering.
- Preserve headers when a loader `Response` is re-wrapped for SSG in `executeServerDataLoaderAsync()` to maintain parity between development and production.

# Test Plan

- CI
- Manual testing locally with the `static-headers` and `static-loader` E2E fixtures in `apps/router-e2e`
- Manual testing on EAS Hosting

# Checklist

### DIFF
--- a/packages/@expo/cli/src/export/__tests__/exportStaticAsync.test.ts
+++ b/packages/@expo/cli/src/export/__tests__/exportStaticAsync.test.ts
@@ -2,6 +2,7 @@ import { getMockConfig as getMockConfigUntyped } from 'expo-router/build/testing
 
 import type { ExpoRouterRuntimeManifest } from '../../start/server/metro/MetroBundlerDevServer';
 import {
+  getExactPathNamedRegex,
   getHtmlFiles,
   getPathVariations,
   getFilesToExportFromServerAsync,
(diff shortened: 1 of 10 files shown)


packages/@expo/cli/src/export/exportStaticAsync.ts
```diff
@@ -86,11 +86,16 @@ function matchGroupName(name: string): string | undefined {
   return name.match(/^\(([^/]+?)\)$/)?.[1];
 }
 
+/**
+ * An allowlist of loader `Response` headers that are written to the export manifest when
+ * pre-rendering.
+ */
+const LOADER_DECLARED_HEADERS = ['Cache-Control'];
```
Ubax: nit, WDYT about a bit more descriptive name like `SSG_LOADER_HEADER_ALLOWLIST`?

packages/@expo/cli/src/export/exportStaticAsync.ts
```diff
@@ -313,6 +317,30 @@ export async function exportFromServerAsync(
...
+  if (!exportServer) {
+    // When we're not exporting a `server` output, we provide a `_expo/.routes.json` for
+    // EAS Hosting to recognize the `headers`, `pageHeaders`, and `redirects` configs
+    const subsetServerManifest: StaticManifest = {
+      headers: serverManifest.headers,
+      pageHeaders: serverManifest.pageHeaders,
+      redirects: serverManifest.redirects,
+    };
+    files.set('_expo/.routes.json', {
+      contents: JSON.stringify(subsetServerManifest, null, 2),
```
Ubax, line 337: nit, do we need `, null, 2)` here? Is the `.routes.json` meant to be read by humans?

packages/@expo/cli/src/export/__tests__/exportStaticAsync.test.ts
```diff
@@ -407,3 +408,30 @@ describe(getFilesToExportFromServerAsync, () => {
     expect([...files.keys()]).toEqual(['(a)/index.html', '(b)/index.html']);
   });
 });
+
+describe(getExactPathNamedRegex, () => {
```
Ubax, line 412: I see that we are escaping more characters. Is it worth adding the test cases for them as well?

Ubax (approved):

packages/@expo/cli/src/export/exportStaticAsync.ts
hassankhan: We were already prettifying it before (see lines 127-130 in this file); separately I think there's value in these manifests being human-readable even if they are ultimately consumed by EAS/`expo-server`.

packages/@expo/cli/src/start/server/metro/MetroBundlerDevServer.ts
```diff
@@ -1906,19 +1906,21 @@ export class MetroBundlerDevServer extends BundlerDevServer {
...
-          // In SSG, extract body
+          // In SSG, extract the body and the headers
           data = await maybeResponse.json();
+          headers = maybeResponse.headers;
         } else {
           data = maybeResponse;
         }
 
-        return Response.json(data ?? null);
+        return Response.json(data ?? null, { headers });
```
Ubax, line 1930: [💡 suggestion] This forwards *all* loader `Response` headers in development, while the export path only propagates the `LOADER_DECLARED_HEADERS` allowlist (`Cache-Control`). So a loader setting e.g. `X-Custom-Header` sees it on the loader endpoint in dev but silently loses it in the static export — the opposite of the parity this change aims for. Consider applying the same allowlist here (sharing the constant), or leaving a comment noting that only allowlisted headers survive SSG export so the divergence is deliberate.

packages/@expo/cli/src/export/exportStaticAsync.ts
```diff
@@ -285,6 +277,18 @@ export async function exportFromServerAsync(
...
+          const declaredHeaders: Record<string, string> = {};
+          for (const name of LOADER_DECLARED_HEADERS) {
+            const value = loaderResponse.headers.get(name);
+            if (value) {
+              declaredHeaders[name] = value;
+            }
+          }
+          if (Object.keys(declaredHeaders).length) {
+            loaderHeadersByPage.set(normalizedPathname, declaredHeaders);
+            loaderHeadersByFile.set(`/${fileSystemPath}`, declaredHeaders);
```
Ubax, line 287: [🐧 nit] Group variations render concurrently under `Promise.all`, so when several pathnames share one loader file this is last-write-wins in nondeterministic order. Today that's harmless — SSG loaders get no request, so every variation should produce identical headers — but a loader with nondeterministic output (e.g. time-based `Cache-Control`) would make the file rule flap between exports. Not worth restructuring for; a short note here would do.

Ubax (approved):
Two more comments form automated review. Nothing serious, just nits

kitten (approved):

PR #47785, opened 2026-07-14
[docs] Revise TypeScript type definition instructions for dom-to-image in Expo tutorial 

johnathanesanders:
Updated instructions for adding TypeScript type definitions for the dom-to-image library, including installation commands for npm, yarn, pnpm, and bun.

# Why

Adding an empty type definition is of little value in TypeScript outside of simply removing the error. As the proper type definitions are (now) available via @types/dom-to-image, it should be used in lieu of an empty type definition.

# How

Simply revised the verbiage and added a Terminal tag with the proper package manager commands.

# Test Plan

Preview the change, as it is documentation only.

# Checklist

- [] I added a `changelog.md` entry and rebuilt the package sources according to [this short guide](https://github.com/expo/expo/blob/main/CONTRIBUTING.md#-before-submitting)
- [] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: updated a module plugin).
- [*] Conforms with the [Documentation Writing Style Guide](https://github.com/expo/expo/blob/main/guides/Expo%20Documentation%20Writing%20Style%20Guide.md)

### DIFF
--- a/docs/pages/tutorial/platform-differences.mdx
+++ b/docs/pages/tutorial/platform-differences.mdx
@@ -210,11 +210,16 @@ const styles = StyleSheet.create({
 
 <Collapsible summary={<>Fix <CODE>dom-to-image</CODE> TypeScript module error</>}>
 
-We need to add a type definition after importing the `domtoimage` library since we're using TypeScript. We can do this by creating a file **types.d.ts** in the root of our project directory and adding the declaration statement:
+We need to add the type definitions after importing the `domtoimage` library since we're using TypeScript. We can do this by adding `@types/dom-to-image` with:
 
-```tsx types.d.ts
-declare module 'dom-to-image';
-```
+  <Terminal
+    cmd={{
+      npm: ['$ npm install @types/dom-to-image --save-dev'],
+      yarn: ['$ yarn add @types/dom-to-image --dev'],
+      pnpm: ['$ pnpm add @types/dom-to-image --save-dev'],
+      bun: ['$ bun add @types/dom-to-image --dev'],
+    }}
+  />
 
 </Collapsible>
 


amandeepmittal (changes requested):
Thanks for the update @johnathanesanders!

Please update the translated part of this tutorial (you can ask your agent) and run `pnpm ja:stamp`.

amandeepmittal (changes requested):
Thanks for digging into this, @johnathanesanders.

You're right on the main point: `@types/dom-to-image` does exist (2.6.7, from DefinitelyTyped), and `declare module 'dom-to-image'` is a blunt instrument that silences the module rather than typing it. Worth raising.

The trouble is what happens when the tutorial's code meets those types. The tutorial creates the ref as `useRef<View>(null)` and passes `imageRef.current` straight into `toJpeg()`, but `@types/dom-to-image` types that parameter as the DOM `Node`. I ran four states against the tutorial's companion app ([expo/examples/stickersmash](https://github.com/expo/examples/tree/master/stickersmash), SDK 54, React Native 0.81.4, `expo/tsconfig.base`, `strict: true`):

| State                                                        | `tsc --noEmit` |
| ------------------------------------------------------------ | -------------- |
| Current docs (`declare module "dom-to-image";`)              | clean          |
| This PR (install `@types/dom-to-image`, drop **types.d.ts**) | 1 error        |
| This PR + `imageRef.current!`                                | 1 error        |
| This PR + `imageRef.current as unknown as Node`              | clean          |

Here's the error from the third row, with the non-null assertion already applied, so this isn't about `null`:

```
app/(tabs)/index.tsx(82,49): error TS2345: Argument of type 'View' is not assignable to parameter of type 'Node'.
  Type 'NativeMethods & ViewComponent' is missing the following properties from type 'Node': baseURI, childNodes, firstChild, isConnected, and 46 more.
```

React Native's `View` is missing 50 members of the DOM `Node` interface. The code works at runtime because React Native Web renders a real DOM node underneath, but TypeScript has no way to know that.

So as written, the change makes the tutorial fail to type-check at exactly the step whose collapsible is titled "Fix `dom-to-image` TypeScript module error". To land it, the tutorial would also have to teach `imageRef.current as unknown as Node` and explain why a React Native ref is being cast to a DOM node, which is a lot to put in front of someone building their first Expo app. The `declare module` line stays, not because it's elegant, but because the library's DOM-oriented types don't describe this usage.

I'm going to close this one.

johnathanesanders:
@amandeepmittal This makes sense in the context of someone building their first Expo app, especially if also learning TypeScript and the differences in the DOM vs. Virtual DOM etc.

I appreciate the review and consideration!

PR #47775, opened 2026-07-14
[docs] Fix a11y audits: labels, contrast, target size, heading order

amandeepmittal:
# Why

Fix ENG-24929

# How

**Components**

- Fix Ask AI trigger labels in `AskPageAITrigger.tsx` so the accessible name contains the visible "Ask AI" text (`label-content-name-mismatch`).
- Add `aria-label="Open in Snack"` to the Snack submit button in `SnackInline.tsx`, whose visible text is hidden at mobile widths (`button-name`).
- Stop tippy from setting `aria-expanded` on code annotation spans in `components/base/code.tsx` (`aria-allowed-attr`).
- Darken links inside callouts to `#0c6ec4` in `InlineHelp` to reach WCAG 4.5:1 on all tinted callout backgrounds (`color-contrast`).
- Mark `react-diff-view` tables as `role="presentation"` in `DiffBlock.tsx`, they are code layout without headers (`td-has-header`).

**Table of contents**

- Enlarge TOC links and section expand buttons to 24px touch targets and remove the button/link overlap in `TableOfContents` (`target-size`).

**Content headings** (`heading-order`)

- Move intro-positioned "Known issues" sections under `## Installation` as `###` on the `video`, `blur-view`, `background-fetch`, and `glass-effect` pages, matching the existing Notifications and ImagePicker pattern.
- Promote skipped `####` headings to `###` across SDK pages (v54 to unversioned) and topic pages (guides, EAS Update, modules, workflow, submit, and others).
- Promote `guides/overview` sections to `##` and change the Image page "Supported image formats" heading to bold text, matching unversioned.

**Dependencies**

- Bump `@expo/styleguide-search-ui` to 9.2.1 and `@expo/styleguide-cookie-consent` to 2.1.6, pulling in the search trigger icon fix (`svg-img-alt`) and the nofollow cookie banner link (SEO `link-text`).
# Test Plan

- On `/versions/latest/sdk/notifications/`: inspect the "Ask AI" trigger, its `aria-label` should be "Ask AI about this page".
- Same page, intro warning callout: the "development build" link renders darker blue (`#0c6ec4`) and stays legible in light and dark themes.
- Same page, narrow the viewport to mobile width: the icon-only Snack submit button ("Open in Snack") has an `aria-label`.
- Same page, hover a code annotation in the Usage example: tooltip still appears, and the annotation `span` carries no `aria-expanded` attribute.
[...]

PR #47786, opened 2026-07-14
[docs] Add `jest-axe` accessibility assertions to component tests

amandeepmittal:
# Why

The accessibility audit failures fixed in ENG-24929 sat in production because the only detection was the bi-weekly/monthly PageSpeed Insights capture. 

**Proposal**: Running `axe-core` inside the component test suites catches the same class of defects (missing accessible names, invalid ARIA, headerless tables) at PR time instead.

Stacked on https://github.com/expo/expo/pull/47775

# How

- Add `jest-axe`, registered globally via `jest.axe-setup.ts` in `setupFilesAfterEnv`.
- Add a shared configured `axe` helper to `common/test-utilities.tsx` that disables the `region` and `color-contrast` rules, which require real layout that jsdom cannot provide.
- Add a `has no axe violations` assertion to 16 existing component test suites (InlineHelp, TableOfContents, DiffBlock, Navigation, Footer, Tabs, Collapsible, FAQ, PageHeader, Prerequisites, Terminal, AppConfigSchemaTable, NativeUpgradePromptCallout, and the API section plugins).
- Add new test files for the previously untested `SnackInline` and `AskPageAITrigger`.
- Disable only the `nested-interactive` rule in the four suites that render `Collapsible`: the sweep surfaced a real pre-existing violation (the permalink anchor is nested inside the `<summary>` control) that needs its own component fix. Re-enable the rule there once it lands.

# Test Plan

Run `pnpm test`: all 486 tests pass.

To see the guards work, in `AskPageAITrigger`:

- Remove the `aria-label`: the accessible name assertion fails (the name falls back to the visible "Ask AI" text, so axe still passes).
- Remove both the `aria-label` and the visible "Ask AI" text: the axe assertion fails with "Buttons must have discernible text (button-name)".

# Checklist

- [ ] I added a `changelog.md` entry and rebuilt the package sources according to [this short guide](https://github.com/expo/expo/blob/main/CONTRIBUTING.md#-before-submitting).


entiendoNull (commented):
Small heads-up on the Test Plan repro.

> remove the aria-label from AskPageAITrigger and rerun: the axe assertion fails

Removing just the aria-label didn't fail the axe assertion. I needed to remove the aria-label and the visible "Ask AI" text, so the button has no accessible name at all, and it'd give me this:

> AskPageAITrigger › has no axe violations

Wanted to double-check whether you wanna change something in the tests or just update the Test Plan... from what I can tell the tests though!

amandeepmittal:
> Small heads-up on the Test Plan repro.
> 
> > remove the aria-label from AskPageAITrigger and rerun: the axe assertion fails
> 
> Removing just the aria-label didn't fail the axe assertion. I needed to remove the aria-label and the visible "Ask AI" text, so the button has no accessible name at all, and it'd give me this:
> 
> > AskPageAITrigger › has no axe violations
> 
> Wanted to double-check whether you wanna change something in the tests or just update the Test Plan... from what I can tell the tests though!

Thanks for catching that! I have updated the test plan. The `button-name` is the actual property that gets logged by pagespeed insights score. With the `aria-label` gone the button's accessible name falls back to the visible "Ask AI" text, so axe stays green and only the explicit `getByRole` name assertion fails. Removing both is what triggers `button-name`.

entiendoNull (approved):

kadikraman (approved):

PR #47788, opened 2026-07-14
[build-properties][Android] Fix PCH dialect mismatch for libraries that set CMAKE_CXX_STANDARD

gabrieldonadel:
# Why

First noticed this while bumping react-native-gesture-handler in https://github.com/expo/expo/pull/47783

Gesture Handler 3.x ships its own `android/CMakeLists.txt` for its codegen target and calls `set(CMAKE_CXX_STANDARD 20)` without `CXX_EXTENSIONS OFF`. Since CMake defaults extensions to ON, that target compiles with `-std=gnu++20`, while the shared PCH owner (`appmodules_pch`) is built with React Native's `-std=c++20`. Clang refuses to reuse a precompiled header across C++ dialects, so the build fails. Any project using `usePrecompiledHeaders` together with a library that sets `CMAKE_CXX_STANDARD` in its own CMakeLists hits the same error.

# How

In `add_pch_if_eligible`, pin every PCH consumer target to the owner's dialect before attaching the header. This is safe for targets that do not set a standard themselves given they already compile with -std=c++20 via target_compile_reactnative_options.

# Test Plan

Run BareExpo locally 

# Checklist

### DIFF
--- a/packages/expo-build-properties/CHANGELOG.md
+++ b/packages/expo-build-properties/CHANGELOG.md
@@ -11,6 +11,7 @@
 ### 🐛 Bug fixes
 
 - [Android] Make the precompiled header reusable by passing `-Xclang -fno-pch-timestamp`, so ccache can reuse it across builds. ([#46915](https://github.com/expo/expo/pull/46915) by [@lukmccall](https://github.com/lukmccall))
+- [Android] Pin the C++ dialect of PCH consumer targets to `-std=c++20`.
 
 ### 💡 Others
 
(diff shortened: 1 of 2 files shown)


lukmccall (approved):
Looks good! We need to add this to bare-expo's `CMakeList` too.

Kudo (approved):

gabrieldonadel:
> Looks good! We need to add this to bare-expo's `CMakeList` too.

Thanks for the heads up, I added it into https://github.com/expo/expo/pull/47783

PR #47790, opened 2026-07-14
[media-library][iOS] Allow network access when resolving iCloud-offloaded assets in the modern API

### LINKED ISSUE #47789: [media-library][iOS] New class-based API can't resolve iCloud-offloaded assets: getUri()/getInfo() throw — isNetworkAccessAllowed never enabled (legacy downloads by default)

https://github.com/oeddyo/rn-icloud-download

### Steps to reproduce

**Summary:** On iOS, the new class-based `expo-media-library` API cannot resolve any asset whose original has been offloaded to iCloud (Settings → Photos → **Optimize iPhone Storage**). `Asset.getUri()`, `Asset.getInfo()`, and `Asset.getExif()` throw, and `Asset.getOrientation()` silently returns `null`. The legacy API succeeds on the exact same assets because it downloads the original from iCloud by default.

**Cause:** the new API's `UriExtractor` never enables `isNetworkAccessAllowed` on its PhotoKit request options, and the default is `false` for both `PHContentEditingInputRequestOptions` and `PHVideoRequestOptions` (verified by instantiating both against the Photos framework):

- Images: [`UriExtractor.swift#L17`](https://github.com/expo/expo/blob/70d4ef855118/packages/expo-media-library/ios/next/objects/asset/UriExtractor.swift#L17) calls `requestContentEditingInput()` with default options → for an offloaded asset this throws `Missing content editing input for image` / `Missing fullSizeImageURL for image`.
- Videos: [`UriExtractor.swift#L28`](https://github.com/expo/expo/blob/70d4ef855118/packages/expo-media-library/ios/next/objects/asset/UriExtractor.swift#L28) constructs `PHVideoRequestOptions()` without setting the flag → throws `Missing AVAsset for video`.
- [`Asset.getOrientation()`](https://github.com/expo/expo/blob/70d4ef855118/packages/expo-media-library/ios/next/objects/asset/Asset.swift#L115) has the same problem and silently returns `null` instead.

**Why this is a parity regression vs the legacy API:** legacy `getAssetInfoAsync` exposes `shouldDownloadFromNetwork`, which **defaults to `true`** ([`MediaLibraryRecords.swift#L90`](https://github.com/expo/expo/blob/70d4ef855118/packages/expo-media-library/ios/MediaLibraryRecords.swift#L90)) and is applied to both the image path ([`MediaLibraryModule.swift#L442`](https://github.com/expo/expo/blob/70d4ef855118/packages/expo-media-library/ios/MediaLibraryModule.swift#L442)) and the video path ([`MediaLibraryModule.swift#L468`](https://github.com/expo/expo/blob/70d4ef855118/packages/expo-media-library/ios/MediaLibraryModule.swift#L468)). So apps migrating from `getAssetInfoAsync().localUri` to `asset.getUri()` / `getInfo().uri` start hard-failing for every "Optimize iPhone Storage" user — a very common configuration — on exactly the assets the legacy API handled transparently.

The new module is also internally inconsistent: its own Live Photo extractor already sets [`options.isNetworkAccessAllowed = true`](https://github.com/expo/expo/blob/70d4ef855118/packages/expo-media-library/ios/next/objects/asset/LivePhotoVideoUriExtractor.swift#L23), so paired Live Photo videos download from iCloud while plain images/videos throw. (`Asset.getIsInCloud()` correctly keeps the flag `false` — it's the cheap "is this offloaded?" pre-check and shouldn't trigger a download.)

**Suggested fix:** set `isNetworkAccessAllowed = true` in `UriExtractor.extract(fromImage:)`, `UriExtractor.extract(fromVideo:)`, and `Asset.getOrientation()` — restoring legacy default behavior and matching `LivePhotoVideoUriExtractor`. Callers who don't want a potentially slow download can check `getIsInCloud()` first, which makes for a coherent API. I'm happy to submit a PR.

**Steps (using the repro app):**

1. Use a **physical iPhone** signed into iCloud Photos with **Optimize iPhone Storage** enabled, where at least one original has been evicted (old large photos/videos in a near-full library are reliable candidates). This cannot be reproduced on a simulator (no iCloud Photos).
2. Clone https://github.com/oeddyo/rn-icloud-download, `npm install`, then `npx expo run:ios --device` (development build; package manager: npm; platform: iOS).
3. Tap **Reproduce**. The app scans the 500 newest assets via the new API, uses `asset.getIsInCloud()` to find an offloaded one, then calls the new API and the legacy API on the same asset.
4. **Expected:** `asset.getUri()` / `asset.getInfo()` resolve the asset (downloading from iCloud if needed), like the legacy API does.
   **Actual:** the new API throws (`Missing content editing input for image` / `Missing fullSizeImageURL for image` for images, `Missing AVAsset for video` for videos), while legacy `getAssetInfoAsync()` on the same asset succeeds and returns a `localUri`.

Affected: `expo-media-library@57.0.1` (latest published) and current `main` (`70d4ef855118`). iOS only — Android's MediaStore files are always local.

### Environment

```
    System:
      OS: macOS 26.5
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 22.13.0 - /Users/eddiexie/.nvm/versions/node/v22.13.0/bin/node
      Yarn: 1.22.19 - /opt/homebrew/bin/yarn
      npm: 10.9.2 - /Users/eddiexie/.nvm/versions/node/v22.13.0/bin/npm
    Managers:
    SDKs:
      iOS SDK:
      Android SDK:
        API Levels: 35, 36
        System Images: android-35 | Google APIs ARM 64 v8a, android-35 | Google Play ARM 64 v8a
    IDEs:
      Android Studio: 2024.2 AI-242.23726.103.2422.13103373
    npmPackages:
      expo: ~57.0.4 => 57.0.4
      react: 19.2.3 => 19.2.3
    npmGlobalPackages:
      eas-cli: 20.5.1
    Expo Workflow: managed
```

### Expo Doctor Diagnostics

```
```

oeddyo:
# Why

Fixes #47789.

On iOS, the modern class-based `expo-media-library` API cannot resolve any asset whose original has been offloaded to iCloud (Settings → Photos → **Optimize iPhone Storage**): `Asset.getUri()`, `getInfo()`, and `getExif()` throw (`Missing content editing input for image` / `Missing fullSizeImageURL for image` / `Missing AVAsset for video`), and `getOrientation()` silently returns `null`.

The legacy API handles the same assets transparently because `getAssetInfoAsync` applies `shouldDownloadFromNetwork` (default `true`) to its PhotoKit request options. The modern API never sets `isNetworkAccessAllowed`, and the platform default is `false` for both `PHContentEditingInputRequestOptions` and `PHVideoRequestOptions` — so apps migrating from `getAssetInfoAsync().localUri` to `asset.getUri()` start hard-failing for every "Optimize iPhone Storage" user.

# How

Set `isNetworkAccessAllowed = true` at the three PhotoKit request sites in the modern API:

- `UriExtractor.extract(fromImage:)` — used by `getUri()`, `getInfo()`, and `getExif()`
- `UriExtractor.extract(fromVideo:)` — used by `getUri()` and `getInfo()`
- `Asset.getOrientation()`

This restores the legacy default and makes the module internally consistent — `LivePhotoVideoUriExtractor` already sets `isNetworkAccessAllowed = true` for paired Live Photo videos. `Asset.getIsInCloud()` intentionally keeps the flag `false` (it is the cheap "is this offloaded?" probe and must not trigger a download); this PR does not touch it.

**On API parity with legacy's `shouldDownloadFromNetwork`:** the only use case for passing `false` in the legacy API was detecting an iCloud-offloaded asset without downloading it (checking `isNetworkAsset` in the result). The modern API already covers that with the dedicated `getIsInCloud()` method, so this PR restores just the default rather than adding a per-call boolean. If per-call download control is wanted in the modern API, an explicit `asset.download({ onProgress })` (PhotoKit's `progressHandler` is available on these request options) would compose better with `getIsInCloud()` than a throwing opt-out on `getUri()` — happy to follow up on that separately if there's interest.

# Test Plan

- Repro app from #47789: https://github.com/oeddyo/rn-icloud-download — on a physical iPhone with iCloud Photos + Optimize iPhone Storage, it finds an offloaded asset via `getIsInCloud()`, then demonstrates `getUri()`/`getInfo()` throwing while legacy `getAssetInfoAsync()` succeeds on the same asset.
- The failure mechanism and this fix were verified against the PhotoKit request-option defaults (`isNetworkAccessAllowed == false` for `PHContentEditingInputRequestOptions`, `PHVideoRequestOptions`, and `PHImageRequestOptions`, checked empirically against the Photos framework) and against the legacy implementation's use of `shouldDownloadFromNetwork` (`MediaLibraryModule.swift` image and video paths).
- Behavior is not reproducible on simulators (no iCloud Photos), so CI cannot exercise it; I can provide before/after logs from a physical device run of the repro app with this patch applied if helpful.

# Checklist

- [x] I added a `changelog.md` entry and rebuilt the package sources according to [this short guide](https://github.com/expo/expo/blob/main/CONTRIBUTING.md#-before-submitting) (iOS-only change; no TypeScript sources affected, so no rebuild output)
- [x] This diff will work correctly for `npx expo prebuild` & EAS Build (no config plugin changes)
- [ ] Conforms with the [Documentation Writing Style Guide](https://github.com/expo/expo/blob/main/guides/Expo%20Documentation%20Writing%20Style%20Guide.md) (no docs changes)

### DIFF
--- a/packages/expo-media-library/CHANGELOG.md
+++ b/packages/expo-media-library/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 ### 🐛 Bug fixes
 
+- [iOS] Fix resolving iCloud-offloaded assets ("Optimize iPhone Storage"): allow network access when extracting asset URIs so `Asset.getUri()`, `getInfo()`, `getExif()` and `getOrientation()` can download originals from iCloud, matching the legacy API's `shouldDownloadFromNetwork` default. ([#47790](https://github.com/expo/expo/pull/47790) by [@oeddyo](https://github.com/oeddyo))
 - Add `accessPrivileges` to `PermissionResponse` type ([#47177](https://github.com/expo/expo/pull/47177) by [@Wenszel](https://github.com/Wenszel))
 - [iOS] Fix permission guards for limited and write-only photo library access. ([#47216](https://github.com/expo/expo/pull/47216) by [@Wenszel](https://github.com/Wenszel))
 
--- a/packages/expo-media-library/ios/next/objects/asset/Asset.swift
+++ b/packages/expo-media-library/ios/next/objects/asset/Asset.swift
@@ -110,9 +110,11 @@ class Asset: SharedObject {
 
   func getOrientation() async throws -> Int? {
     let phAsset = try await requirePHAsset()
+    let options = PHContentEditingInputRequestOptions()
+    options.isNetworkAccessAllowed = true
     guard
       phAsset.mediaType == .image,
-      let contentEditingInput = try await phAsset.requestContentEditingInput().input
+      let contentEditingInput = try await phAsset.requestContentEditingInput(options: options).input
     else {
       return nil
     }
--- a/packages/expo-media-library/ios/next/objects/asset/UriExtractor.swift
+++ b/packages/expo-media-library/ios/next/objects/asset/UriExtractor.swift
@@ -14,7 +14,9 @@ class UriExtractor {
   }
 
   private static func extract(fromImage phAsset: PHAsset) async throws -> URL {
-    let result = try await phAsset.requestContentEditingInput()
+    let options = PHContentEditingInputRequestOptions()
+    options.isNetworkAccessAllowed = true
+    let result = try await phAsset.requestContentEditingInput(options: options)
     guard let contentEditingInput = result.input else {
       throw FailedToExtractUri("Missing content editing input for image")
     }
@@ -27,6 +29,7 @@ class UriExtractor {
   private static func extract(fromVideo phAsset: PHAsset) async throws -> URL {
     let options = PHVideoRequestOptions()
     options.version = .original
+    options.isNetworkAccessAllowed = true
     let result = try await PHImageManager.default()
       .requestAVAsset(forVideo: phAsset, options: options)
     guard let avAsset = result.asset else {


Wenszel (approved):
Hey, thanks for taking care of this issue! I checked on a physical device with Optimize Storage on - LGTM 👍

PR #47791, opened 2026-07-15
[android] Set CMAKE_OBJECT_PATH_MAX=1024 default for CMake modules

ide:
Why
===
CMake limits object file paths to 250 characters on Windows and 1000 elsewhere; builds that exceed the limit get a warning and eventually fail. The limit is CMake's guess at the longest path the downstream tools (Ninja, the compiler, the archiver) can handle; CMake can't know their real limits, so it guesses conservatively and fails preemptively, even when the tools support longer paths.

Deeply nested paths like a pnpm virtual store commonly exceed the Windows limit and `react-native-reanimated` and `react-native-worklets` had this problem, too.

Each Android module that builds native code runs its own CMake configure, so fixing this per library (as #39662 proposed for `expo-modules-core`) would require the same change from every affected library author. Every SDK 53+ project applies `expo-root-project` in its root `build.gradle`, so modern projects will pick this up with a package upgrade and library authors don't need to do anything.

There are two parts to this Windows long-path problem (#36274): the toolchain programs (like Ninja) need to support longer file paths, and CMake needs to stop blocking longer file paths:
- PR #47377 addressed the toolchain part. AGP's default CMake 3.22.1 bundles a Ninja without long-path support, and the opt-in `android.cmakeVersion` build property selects a newer CMake.
- This PR addresses CMake's check.

How
===
A new `expo.android.cmakeObjectPathMax` Gradle property controls CMake's `CMAKE_OBJECT_PATH_MAX`:

| `expo.android.cmakeObjectPathMax` | Behavior                                                  |
| --------------------------------- | --------------------------------------------------------- |
| Unset or empty                    | Uses the default of `1024`                                |
| `0`                               | Opts out; keeps CMake's platform defaults                 |
| Integer ≥ `128` (CMake's minimum) | Uses that value                                           |
| Anything else                     | Logs a warning and falls back to the default of `1024`    |

`ExpoRootProjectPlugin` adds `-DCMAKE_OBJECT_PATH_MAX=1024` to `defaultConfig.externalNativeBuild.cmake.arguments` for every subproject that applies an Android plugin. PR #47377 overrides the CMake version the same way.

The injection runs when a module applies the Android plugin, before the module's own `android` block. This means a library that passes its own `-DCMAKE_OBJECT_PATH_MAX` argument can override our default since the library's argument comes later on the CMake command line.

A note on 1024: macOS limits paths to 1024 bytes (`PATH_MAX`), so we can't go higher there. CMake already allowed 1000 on Mac and Linux, so this barely changes anything for them; the real point is to raise the 250-character limit on Windows.

Test Plan
===
Ran `configureCMakeDebug[arm64-v8a]` in bare-expo and confirmed `-DCMAKE_OBJECT_PATH_MAX=1024` appears in the generated `.cxx/**/metadata_generation_command.txt` for `expo-modules-core`, `react-native-worklets`, and the app module, none of which set the value themselves.

Verified `-Pexpo.android.cmakeObjectPathMax=2048` changes the injected value, `-Pexpo.android.cmakeObjectPathMax=0` omits the argument entirely, and `-Pexpo.android.cmakeObjectPathMax=abc` logs a warning and falls back to the default of 1024.

Ran `et check-packages expo-modules-autolinking`, which passed.

### DIFF
--- a/packages/expo-modules-autolinking/CHANGELOG.md
+++ b/packages/expo-modules-autolinking/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ### 🎉 New features
 
+- [Android] Set `CMAKE_OBJECT_PATH_MAX=1024` by default for the app and all library subprojects that build native code with CMake, so long object file paths (for example in pnpm monorepos on Windows) no longer fail the build. Configurable with the `expo.android.cmakeObjectPathMax` Gradle property. ([#47791](https://github.com/expo/expo/pull/47791) by [@ide](https://github.com/ide))
 - [Android] Apply the `android.cmakeVersion` build property to the app and all library subprojects. ([#47377](https://github.com/expo/expo/pull/47377) by [@zoontek](https://github.com/zoontek))
 - Add option to specify targets to use with inline modules ([#46698](https://github.com/expo/expo/pull/46698) by [@HubertBer](https://github.com/HubertBer))
 - Added support for compile-only inline module files, which are compiled into the target without being registered as Expo modules. ([#46969](https://github.com/expo/expo/pull/46969) by [@behenate](https://github.com/behenate))
--- a/packages/expo-modules-autolinking/android/expo-gradle-plugin/README.md
+++ b/packages/expo-modules-autolinking/android/expo-gradle-plugin/README.md
@@ -21,6 +21,28 @@ Responsibilities:
 - Add previously linked modules to the dependency graph.
 - Create a task that will generate the package list file.
 
+### `expo-root-project`
+
+This plugin should be applied to the root `build.gradle` file of the application (the project templates already do this). It configures defaults that apply to the app and all of its modules.
+
+Responsibilities:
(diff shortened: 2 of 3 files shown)


chrfalch (approved):

lukmccall (approved):

kitten (approved):
tested and verified on Windows that this raises the path length for builds and then succeeds. As expected, with the default CMake version, the bundled Ninja (1.10.2) can still cause problems, unless the Cmake version is raised via `android.cmakeVersion`, but that's expected (and now within user's reach of arriving at a full workaround)

However, since I last tested this, an app may now contain scripts and `bat` files that can cause the working directory (for `CreateProcess`) to be above the limit.

Effectively this means that we're unlocking _slightly_ longer paths (which is still worth it), but there's problems that still remain (and workarounds for those are trickier, but either way, are candidates for follow-up PRs)

The good news is that those errors are more easily detectable and readable, and not as unexpected as the "late" Cmake errors

ide:
In a follow up I wonder if we should check for Ninja's capabilities (or version) on Windows and show a clearer error message. Ultimately the way everyone is going to debug this issue is asking some agent.

PR #47795, opened 2026-07-15
chore(cli): Fix misleading `--private-key-path` help text

### LINKED ISSUE #47812: [cli] `expo start` ignores the documented default for `--private-key-path` (signed dev manifests fail with 500)

https://github.com/ciospettw/repro-expo-cli-privatekey

A blank Expo app (`create-expo-app --template blank`) with expo-updates code signing configured via `expo-updates codesigning:generate` — `certs/certificate.pem` and `certs/private-key.pem` sit in the same directory, which is exactly the default location the `--private-key-path` help text describes. Throwaway keys are committed on purpose so the repro is self-contained.

### Steps to reproduce

Platform: any (this is a `@expo/cli` dev-server issue, reproduced on macOS with npm). Environment: development server (`expo start`).

1. `git clone https://github.com/ciospettw/repro-expo-cli-privatekey && cd repro-expo-cli-privatekey && npm install`
2. Start the dev server **without** `--private-key-path`:
   ```bash
   npx expo start --port 8099
   ```
3. Request the manifest the way a code-signing-enabled client does (with an `expo-expect-signature` header):
   ```bash
   curl -s -D - \
     -H 'expo-platform: ios' \
     -H 'expo-dev-client: true' \
     -H 'expo-expect-signature: sig, keyid="main", alg="rsa-v1_5-sha256"' \
     http://localhost:8099/
   ```

**Expected:** `200 OK` with an `expo-signature` header, because `expo start --help` documents that `--private-key-path` defaults to *"private-key.pem in the same directory as the certificate specified by the expo-updates configuration in app.json"* — and that file exists.

**Actual:**
```
HTTP/1.1 500 Internal Server Error
{"error":"CommandError: Must specify --private-key-path argument to sign development manifest for requested code signing key"}
```

Proof the key/cert are valid and only the default is missing — passing the flag explicitly works:
```bash
npx expo start --port 8099 --private-key-path certs/private-key.pem
# the same curl now returns 200 with: expo-signature: keyid="main", sig="…", alg="rsa-v1_5-sha256"
```

Root cause is in `packages/@expo/cli/src/utils/codesigning.ts`, `getProjectCodeSigningCertificateAsync`, which throws when `privateKeyPath` is undefined instead of resolving the documented default. Fix proposed in https://github.com/expo/expo/pull/47795.

### Environment

```
    System:
      OS: macOS 26.0.1
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 22.17.0 - /usr/local/bin/node
      npm: 10.9.2 - /usr/local/bin/npm
    Managers:
    SDKs:
      iOS SDK:
    IDEs:
      Xcode: 26.3/17C529 - /usr/bin/xcodebuild
    npmPackages:
      expo: ~57.0.6 => 57.0.6
      expo-updates: ~57.0.7 => 57.0.7
      react: 19.2.3 => 19.2.3
    Expo Workflow: managed
```

### Expo Doctor Diagnostics

```
```

ciospettw:
Closes #47812.

# Why

`expo start --help` documents a default for `--private-key-path`:

```
--private-key-path <path>  Path to private key for code signing.
                           Default: "private-key.pem" in the same directory as the
                           certificate specified by the expo-updates configuration in app.json.
```

But `expo start` never applies that default. For project code signing, `getProjectCodeSigningCertificateAsync` throws when the flag is absent — and that throw is intentional and covered by an existing test (`codesigning-test.ts` → *"throws when private key path is not supplied"*). The conventional setup also keeps keys and certs in separate directories (the tests use `keys/private-key.pem` + `certs/cert.pem`), so there's no reliable "next to the certificate" file to fall back to.

So the behavior is deliberate; the **help text is what's inaccurate** — it promises a default that doesn't exist, which sends users debugging a 500 (`Must specify --private-key-path…`) they didn't expect.

# How

Correct the `--private-key-path` help text to describe the actual behavior (the flag is required to sign development manifests for project code signing) instead of documenting a non-existent default. No behavior change, no test changes.

> If the maintainers would rather *implement* the documented default instead (resolve `private-key.pem` next to the certificate when the flag is omitted), I'm happy to switch this PR to that — it would just also need the existing "throws when private key path is not supplied" test updated. Fixing the docs seemed like the lower-risk change given the throw is intentional and tested.

# Test Plan

- `npx expo start --help` → the `--private-key-path` line no longer claims a default.
- No code paths changed; existing `codesigning-test.ts` suite is untouched and still passes.

### DIFF
--- a/packages/@expo/cli/CHANGELOG.md
+++ b/packages/@expo/cli/CHANGELOG.md
@@ -32,6 +32,7 @@
 - [Internal] Migrate structured event logging to the `2g` package ([#46667](https://github.com/expo/expo/pull/46667) by [@kitten](https://github.com/kitten))
 - [Internal] Migrate an initial set of events to `2g` ([#47655](https://github.com/expo/expo/pull/47655) by [@kitten](https://github.com/kitten))
 - [Internal] Resolve the dev server port once instead of re-deriving it, and read the URL environment variables outside the `UrlCreator` ([#48236](https://github.com/expo/expo/pull/48236) by [@ramonclaudio](https://github.com/ramonclaudio))
+- Update `expo start --private-key-path` help text ([#47795](https://github.com/expo/expo/pull/47795) by [@kitten](https://github.com/kitten))
 
 ## 57.0.11 - 2026-07-29
 
@@ -70,7 +71,6 @@ _This version does not introduce any user-facing changes._
 
 ### 🐛 Bug fixes
 
-- Fix the `expo start --private-key-path` help text, which documented a `private-key.pem`-next-to-certificate default that `expo start` does not implement (the flag is required to sign development manifests for project code signing). ([#TBD](https://github.com/expo/expo/pull/TBD) by [@ciospettw](https://github.com/ciospettw))
 - Avoid writing a 500 response when a DevTools plugin server request has already started or aborted. ([#47192](https://github.com/expo/expo/pull/47192) by [@krystofwoldrich](https://github.com/krystofwoldrich))
 - Pass a fetch API `Request` to DevTools plugin WebSocket handlers instead of `IncomingMessage`. ([#47410](https://github.com/expo/expo/pull/47410) by [@krystofwoldrich](https://github.com/krystofwoldrich))
 - Use `loadModule` for MCP and DevTools plugin server module loading. ([#47139](https://github.com/expo/expo/pull/47139) by [@krystofwoldrich](https://github.com/krystofwoldrich))
--- a/packages/@expo/cli/CHANGELOG.md
+++ b/packages/@expo/cli/CHANGELOG.md
@@ -70,6 +70,7 @@ _This version does not introduce any user-facing changes._
 
 ### 🐛 Bug fixes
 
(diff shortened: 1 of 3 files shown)


kitten (approved):

packages/@expo/cli/CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
 
 ### 🐛 Bug fixes
 
+- Fix the `expo start --private-key-path` help text, which documented a `private-key.pem`-next-to-certificate default that `expo start` does not implement (the flag is required to sign development manifests for project code signing). ([#TBD](https://github.com/expo/expo/pull/TBD) by [@ciospettw](https://github.com/ciospettw))
```
kitten:
```suggestion
- Fix the `expo start --private-key-path` help text, which documented a `private-key.pem`-next-to-certificate default that `expo start` does not implement (the flag is required to sign development manifests for project code signing). ([#47795](https://github.com/expo/expo/pull/47795) by [@ciospettw](https://github.com/ciospettw))
```

PR #42161, opened 2026-01-14
[fetch] Fix streaming race condition causing out-of-order chunk delivery

matthieugicquel:
# Why

Fixes #41783 on Android.

I also observed this issue of out-of-order delivery in production in a significant amount of requests.

My understanding is that there's a race between startStreaming() (called from JS) and the network callbacks delivering chunks:

In some cases it causes a chunk loss:
```
     IO Thread                         JS Thread
        │                                  │
   sink.append(A)                          │
   sink.append(B)                          │
        │                                  │
        ├─ read state ─────────────────────┤
        │  (RESPONSE_RECEIVED)             │
        │                             state = STREAMING
        │                             data = sink.finalize() → [A,B]
        │                             emit([A,B])
        │                                  │
   sink.append(C) ←── sink is finalized, C goes nowhere
        │                                  │
        ├─ read state ─────────────────────┤
        │  (STREAMING)                     │
   emit(D)                                 │
        │                                  │
        ▼                                  ▼
JS receives: [A,B], [D]
C is lost
```

And in other ones out-of-order delivery:

```
     IO Thread                         JS Thread
        │                                  │
   sink.append(A)                          │
   sink.append(B)                          │
        │                                  │
        ├─ read state ─────────────────────┤
        │  (RESPONSE_RECEIVED)             │
        │                             state = STREAMING
   sink.append(C) ←── C wins race, added before finalize
        │                                  │
        ├─ read state ─────────────────────┤
        │  (STREAMING)                     │
   emit(D)                            data = sink.finalize() → [A,B,C]
        │                             emit([A,B,C])
        ▼                                  ▼
JS receives: [D], [A,B,C]
```

# How

Add a `synchronised` block. This seems safe, but I'm not 100% sure about the perf impact / whether that's the best way to handle this.

_Note: a previous version of this PR included an iOS fix, but as I haven't been able to reproduce the issue there, I dropped this part for now (though production logs indicate the issues exists on iOS too)._

# Test Plan

I've pushed a repro to [this commit](https://github.com/matthieugicquel/expo/commit/e83d7c11d2fb16df19ed51dd4c8763f22c6d8a43) of my fork on the minimal-tester app. It triggers the race condition ~10% of the time on the device I have before the fix, and 0% after the fix.

_Note: this repro also frequently triggers a "stream is not in a state that permits close" error that I'm not trying to fix in this PR._

# Checklist

[...]

PR #47796, opened 2026-07-15
[ios][fetch] Fix streaming race causing dropped chunks and empty 200 responses

idoyana:
# Why

`expo/fetch` on iOS can deliver an **empty body with HTTP status 200**, drop body chunks, or (rarely) leave the JS `ReadableStream` open forever.

`NativeResponse`'s `URLSessionDataDelegate` callbacks run on URLSession's internal serial delegate queue, while `startStreaming()` / `cancelStreaming()` run on `fetchRequestQueue` (`.runOnQueue` in `ExpoFetchModule`). `state` and `ResponseSink` are shared across both queues with no synchronization (`ResponseSink.bodyQueue` is a plain array; the class is `@unchecked Sendable`), so chunk routing races concurrent state transitions:

1. **Lost in sink** — `didReceive(data:)` observes `.responseReceived`, then `startStreaming()` transitions + `finalize()`s the sink before `appendBufferBody` lands. The chunk is appended into a finalized sink and is never delivered.
2. **Dropped after close** — `didCompleteWithError` observes `.bodyStreamingStarted` mid-`startStreaming()` and emits `didComplete` before the flushed data event. JS (`FetchResponse.ts`) closes the stream on `didComplete` and discards the late data event — the widest window in practice.
3. **Never closed** — `startStreaming()` and completion interleave so `state = .bodyStreamingStarted` overwrites `.bodyCompleted` after completion already ran; `didComplete` is never emitted and the pending JS read hangs.

For a small single-chunk response (e.g. a JSON API reply), 1 and 2 both surface as a *successful* response with an empty body — no error anywhere. We hit this in production in a React Native app: an RTK Query mutation received a 200 whose body parsed to zero bytes, observed as `data: null` downstream.

This is the iOS counterpart of #42161 (Android: chunk loss / out-of-order delivery between the OkHttp pump and `startStreaming()`). It is distinct from the abort-path stream teardown addressed in #47573 — no `AbortSignal` is involved here, and the failure reproduces with the #47573 guard in place.

# How

Hop every `ExpoURLSessionTaskDelegate` callback onto `dispatchQueue` (`fetchRequestQueue`) — the queue the JS-driven methods already run on — so the entire state machine is confined to one serial queue. This mirrors the module's existing intent: state-change listeners, `waitFor`, and the file-404 completion re-entry already dispatch there; the URLSession entry points were the only writers left on a foreign queue.

- `didCompleteWithError` body is extracted into `handleDidCompleteWithError` so the existing file-404 next-loop re-entry stays a single hop instead of double-hopping through the public method.
- `willPerformHTTPRedirection` calls `completionHandler(nil)` in the `guard let self` failure branch so URLSession never strands the task.
- Ordering is preserved end-to-end: URLSession's delegate queue is serial and `dispatchQueue.async` preserves enqueue order.

# Test Plan

Standalone race harness (macOS, `swiftc`) that compiles `NativeResponse.swift`, `ResponseSink.swift`, and `ResponseState.swift` **verbatim** against a minimal `SharedObject` stub that records emitted events, reproducing the production queue topology: delegate callbacks on one serial queue, `startStreaming()` on `fetchRequestQueue`. Each iteration delivers one 512-byte chunk + completion while `startStreaming()` races on the other queue; the spy applies `FetchResponse.ts` semantics (data after `didComplete` is discarded).

| | before | after |
|---|---|---|
| Behavioral, 40,000 iterations | **541 lossy** (472 dropped-after-close, 84 lost-in-sink, 35 never-closed) | **0 — 40,000/40,000 clean** |
| ThreadSanitizer (`-sanitize=thread`), 2,000 iterations | **6–8 data-race reports** (`state` setter in `startStreaming()` vs. delegate reads) | **0 reports** |

Happy to contribute the harness or a CI-runnable variant if useful.


idoyana:
Hi @Kudo 👋 — one more in the same area as #47573, but on the native iOS side this time (it's close to your earlier streaming fixes #33756 / #30604, so this should be familiar territory).

`ExpoURLSessionTaskDelegate` callbacks run on URLSession's delegate queue while `startStreaming()` runs on `fetchRequestQueue`, with `state` and `ResponseSink` shared unsynchronized — so chunk delivery races the state machine, and a small single-chunk response can surface as a **200 with an empty body** (plus rarer dropped-chunk / never-closed variants). The fix hops every delegate callback onto `fetchRequestQueue`, confining the whole state machine to the one serial queue the JS-driven methods already use. A 40k-iteration race harness goes from 541 lossy runs to 0, and TSan reports drop to 0 (details in the PR body).

It's independent of #47573 (no AbortSignal involved), but since you already have that one queued, it might be efficient to look at both together — they share the fetch-teardown context. Thanks! 🙏

idoyana:
Rebased on main to resolve the changelog conflict with #42161 — that PR fixed the Android side of this same delivery race (native pump vs. `startStreaming()`); this PR is the iOS side of it. No code changes in the rebase.

ottob:
This seems rather important. Any updates on when this will be fixed for sdk 57?

ottob:
@Kudo gentle ping

idoyana:
Hi @zoontek 👋 — since Kudo seems to be away, would you be able to take a look? You have the freshest context here: your #48230 un-strands `text()`/`arrayBuffer()` when a request fails mid-response; this PR fixes the delegate-queue race in `NativeResponse.swift` that corrupts the delivery in the first place (empty-200s, dropped chunks, never-closed streams) — the iOS side of the Android fix in #42161. Still green and conflict-free against main, with a 40k-iteration race harness + TSan runs in the test plan.

Its JS sibling #47573 (propagating `AbortSignal` into the body stream) is also awaiting a reviewer, if you have appetite for both. Thanks! 🙏

idoyana:
@ottob thanks for the interest! From my side the PR is ready — rebased over #42161 (the Android side of this same race), green CI, and no conflicts with the recent #48230 (different file, complementary fix). Just awaiting maintainer review.

idoyana:
The two failing checks look unrelated to this PR: **macos-build** is currently failing on every recent branch (including maintainers' own PRs), and **iOS Unit Tests / build** failed on `main` itself at 08:46 UTC today — this branch's run (10:23) fell in the same window, with every individually-listed test in its log passing before the runner declared all ~25 packages failed. Runs after ~12:10 UTC are green across other PRs, so a re-run should clear it. The branch is as you left it — nothing changed on my side.

(Also rebased the JS sibling #47573 over its changelog conflict — green and conflict-free again.)

zoontek (approved):

PR #47798, opened 2026-07-15
[dev-launcher] Use `https` for `exp://` tunnel URLs (`*.exp.direct`)

### LINKED ISSUE #47815: [expo-dev-launcher] `exp://` tunnel URLs (`*.exp.direct`) are downgraded to cleartext http and blocked by ATS

https://github.com/ciospettw/repro-expo-cli-privatekey

Blank Expo app with `expo-dev-client` (see "Bug 3" in the repo README). Throwaway keys committed for a self-contained repro.

### Steps to reproduce

Platform: iOS. Environment: development build (`expo-dev-client`).

1. `git clone https://github.com/ciospettw/repro-expo-cli-privatekey && cd repro-expo-cli-privatekey && npm install`
2. Build a development build and install on an iOS device: `npx expo run:ios --device`.
3. Start a tunnel: `npx expo start --dev-client --tunnel --private-key-path certs/private-key.pem`. Note the tunnel host, e.g. `xxxxx.exp.direct`.
4. On the device, open `exp://xxxxx.exp.direct` (manual entry or a deep link).

**Expected:** loads — `exp://` tunnel hosts should resolve to `https`, which is what Expo CLI already uses for tunnel manifest URLs (`UrlCreator.constructDevClientUrl`: `scheme: hostType === 'tunnel' ? 'https' : 'http'`). Opening the same host as `https://xxxxx.exp.direct` works.

**Actual (iOS):** fails to connect. `DevLauncherUrl` rewrites `exp://` → `http://` unconditionally (`EXDevLauncherURLHelper.swift` / `DevLauncherURLHelper.kt`), producing a cleartext request to a TLS-only host.

### Why it is blocked (App Transport Security)

Expo's default `Info.plist` (both `templates/expo-template-bare-minimum` and the `expo prebuild` output) sets:

```xml
<key>NSAppTransportSecurity</key>
<dict>
  <key>NSAllowsArbitraryLoads</key><false/>
  <key>NSAllowsLocalNetworking</key><true/>
</dict>
```

So cleartext HTTP is allowed to **local** hosts (LAN/localhost) but **blocked to remote hosts**. A tunnel host is remote, so the downgraded `http://xxxxx.exp.direct` request is blocked by ATS — which is exactly why LAN `exp://` URLs keep working while tunnel `exp://` URLs fail. Verified against the built app's `Info.plist` on an iOS 26.5 device.

Fix proposed in https://github.com/expo/expo/pull/47798.

### Environment

```
    System: OS: macOS 26.0.1
    Binaries: Node 22.17.0, npm 10.9.2
    Managers: CocoaPods 1.16.2
    IDEs: Xcode 26.3/17C529
    npmPackages: expo ~57.0.6, expo-updates ~57.0.7, expo-dev-client ~57.0.6, react 19.2.3, react-native 0.86.0
    Expo Workflow: managed
```

### Expo Doctor Diagnostics

```
```

ciospettw:
Closes #47815.

Minimal reproducible example: https://github.com/ciospettw/repro-expo-cli-privatekey

# Why

`EXDevLauncherUrl` (iOS) and `DevLauncherUrl` (Android) rewrite `exp://` to `http://` unconditionally:

```swift
self.url = EXDevLauncherURLHelper.replaceEXPScheme(urlFromParam, to: "http")
```

For LAN/localhost this is correct, but tunnel hosts (`*.exp.direct`, from `expo start --tunnel`) are served over TLS. Expo CLI itself generates **https** manifest URLs for tunnels:

```ts
// UrlCreator.constructDevClientUrl
scheme: this.defaults?.hostType === 'tunnel' ? 'https' : 'http',
```

so the two sides disagree on the scheme for the same host.

On iOS this is not just aesthetic — it fails. Expo's default `Info.plist` (both `templates/expo-template-bare-minimum` and the `expo prebuild` output) sets:

```xml
<key>NSAppTransportSecurity</key>
<dict>
  <key>NSAllowsArbitraryLoads</key><false/>
  <key>NSAllowsLocalNetworking</key><true/>
</dict>
```

Cleartext HTTP is allowed to local hosts but blocked to remote hosts. A tunnel host is remote, so the downgraded `http://xxxxx.exp.direct` request is blocked by App Transport Security. This is exactly why LAN `exp://` URLs keep working while tunnel `exp://` URLs fail.

# How

Add a `packagerScheme` helper that returns `https` for `*.exp.direct` hosts and `http` otherwise, and use it where `DevLauncherUrl` currently hardcodes `"http"`, on both platforms. LAN/localhost behavior is unchanged.

# Test Plan

- `npx expo start --dev-client --tunnel`; open `exp://<host>.exp.direct` on device.
  - Before: fails on iOS (cleartext to a TLS-only remote host, blocked by ATS given the default `NSAllowsArbitraryLoads=false` / `NSAllowsLocalNetworking=true`).
  - After: resolves to `https://<host>.exp.direct` and loads.
- LAN: `exp://192.168.1.10:8081` still maps to `http://192.168.1.10:8081` (unchanged; allowed by `NSAllowsLocalNetworking`).

# Checklist

- Changelog entry added under **Unpublished → 🐛 Bug fixes** in `packages/expo-dev-launcher/CHANGELOG.md`.
- Both iOS and Android covered. Follow-up: extend `EXDevLauncherURLHelperTests.swift` / `DevLauncherURLHelperTest.kt` with a tunnel-host case.

### DIFF
--- a/packages/expo-dev-launcher/android/src/debug/java/expo/modules/devlauncher/helpers/DevLauncherURLHelper.kt
+++ b/packages/expo-dev-launcher/android/src/debug/java/expo/modules/devlauncher/helpers/DevLauncherURLHelper.kt
@@ -4,6 +4,11 @@ import android.net.Uri
 
 fun replaceEXPScheme(uri: Uri, scheme: String): Uri = if (uri.scheme == "exp") uri.buildUpon().scheme(scheme).build() else uri
 
+// Tunnel hosts (e.g. `*.exp.direct`) are served over TLS, so `exp://` tunnel URLs
+// must map to `https` — matching the scheme the Expo CLI uses for tunnel manifest
+// URLs — while LAN/localhost dev servers keep using `http`.
+fun packagerScheme(uri: Uri): String = if (uri.host?.endsWith(".exp.direct") == true) "https" else "http"
+
 fun isDevLauncherUrl(uri: Uri) = uri.host == "expo-development-client"
 
 fun hasUrlQueryParam(uri: Uri): Boolean {
@@ -21,10 +26,10 @@ class DevLauncherUrl(var url: Uri) {
     if (isDevLauncherUrl(url)) {
       if (queryParams["url"] != null) {
         val queryUrl = Uri.parse(queryParams["url"])
-        url = replaceEXPScheme(queryUrl, "http")
+        url = replaceEXPScheme(queryUrl, packagerScheme(queryUrl))
       }
     } else {
-      url = replaceEXPScheme(url, "http")
+      url = replaceEXPScheme(url, packagerScheme(url))
     }
   }
 }
--- a/packages/expo-dev-launcher/CHANGELOG.md
+++ b/packages/expo-dev-launcher/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 ### 🐛 Bug fixes
 
+- Map `exp://` URLs pointing at tunnel hosts (`*.exp.direct`) to `https` instead of cleartext `http`, matching the scheme Expo CLI uses for tunnel manifest URLs and avoiding iOS App Transport Security blocks. ([#TBD](https://github.com/expo/expo/pull/TBD) by [@ciospettw](https://github.com/ciospettw))
 - [iOS] Use `RCTPlatformName` instead of hardcoding `ios` when requesting bundles from Metro. ([#46443](https://github.com/expo/expo/pull/46443) by [@gabrieldonadel](https://github.com/gabrieldonadel))
 - [iOS] Cleared the deep-link URL from cached `launchOptions` after it is consumed ([#46265](https://github.com/expo/expo/pull/46265) by [@gabrieldonadel](https://github.com/gabrieldonadel))
 - [Android] Fixed a crash when cold-launching a development build from a deep link that carries intent categories (e.g. an App Link opened from a browser). ([#46314](https://github.com/expo/expo/pull/46314) by [@lilianchiassai-fc](https://github.com/lilianchiassai-fc) & [#46328](https://github.com/expo/expo/pull/46328) by [@lukmccall](https://github.com/lukmccall))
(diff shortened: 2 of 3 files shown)


packages/expo-dev-launcher/android/src/debug/java/expo/modules/devlauncher/helpers/DevLauncherURLHelper.kt
```diff
@@ -4,6 +4,11 @@ import android.net.Uri
 
 fun replaceEXPScheme(uri: Uri, scheme: String): Uri = if (uri.scheme == "exp") uri.buildUpon().scheme(scheme).build() else uri
 
+// Tunnel hosts (e.g. `*.exp.direct`) are served over TLS, so `exp://` tunnel URLs
+// must map to `https` — matching the scheme the Expo CLI uses for tunnel manifest
+// URLs — while LAN/localhost dev servers keep using `http`.
+fun packagerScheme(uri: Uri): String = if (uri.host?.endsWith(".exp.direct") == true) "https" else "http"
```
kitten, line 10: This isn't exhaustive or correct, since this just adds a magic string for one case. #47255 is probably a good opportunity to make sure we have the info we need here, and accept that behaviour is slightly breaking/changed

cc @alanjhughes @tsapeta: Any opinions on this?

ciospettw: You're right — I looked into this more closely and I'm walking it back.

Two things I'd missed: `getNativeRuntimeUrl` already routes dev clients through `constructDevClientUrl`, which uses `https` for tunnels, so the standard dev-client tunnel flow already works on iOS without this — it only affected a manually-entered `exp://` (Expo-Go-style) URL. And forcing `exp://` tunnel hosts to `https` steps on the intentional `exp://` → `http` convention, which `EXPO_TUNNEL_SUBDOMAIN` already gates behind an opt-in precisely because it "may cause issues with expo-linking and Expo Go". So it's the wrong layer, and #47255 is the right place to carry the scheme through.

Closing this — thanks for the pointer. (The sibling PR #47797, surfacing the dev server's actual error body instead of a generic "failed to connect", is orthogonal to the scheme question, so I've left that one open.)

ciospettw:
Closing in favour of handling the scheme in #47255 (see review thread) — this was the wrong layer. The orthogonal error-surfacing fix stays in #47797.

PR #47802, opened 2026-07-15
[web][expo] Fix getBundleUrl returning null after script execution

### LINKED ISSUE #47754: import.meta.url is null on web when read outside synchronous bundle execution

https://github.com/tyrauber/expo-import-meta-url-web

### Steps to reproduce

Platform: web. Environment: dev server (Metro). Package manager: npm.

1. Clone https://github.com/tyrauber/expo-import-meta-url-web and `npm install` (or `npx create-expo-app@latest --template blank`, add react-dom + react-native-web, and apply the App.js diff from the second commit).
2. `npx expo start --web` and open the app in a browser.
3. App.js reads `import.meta.url` inside a `useEffect` (i.e. after load) and then does `new URL('../assets/', import.meta.url)`.

Expected: `import.meta.url` is the bundle URL, so `new URL(relative, import.meta.url)` resolves. This is what ESM packages assume at module scope — three's DRACOLoader/KTX2Loader compute `new URL('../libs/draco/', import.meta.url)` this way.

Actual (rendered on screen and logged):
    import.meta.url = null
    new URL('../assets/', import.meta.url)
      THREW: Failed to construct 'URL': Invalid base URL

Root cause: Metro rewrites `import.meta` to `globalThis.__ExpoImportMetaRegistry`, whose `url` getter calls `getBundleUrl()`. On web (`expo/src/utils/getBundleUrl.web.ts`) that derives the URL from `document.currentScript?.src`. `document.currentScript` is only set during a script's synchronous execution, so any read after load — a lazy `require`, a dynamic import, anything async — sees `null`, and `import.meta.url` is `null`. When the access is at module scope, the whole web bundle crashes.

This is in `expo` (the winter runtime / getBundleUrl) plus the Metro import.meta transform, not expo-router — the crash just surfaces in router apps because that's what renders.

> Transparency note: this issue was drafted with AI assistance. I'm a human, I've personally verified the bug in the linked reproduction, and I'm happy to answer any questions or concerns directly. Thanks for the great tooling.

### Environment

```text
    System:
      OS: macOS 26.5.1
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 22.16.0
      Yarn: 1.22.15
      npm: 10.9.2
      Watchman: 2025.09.15.00
    Managers:
      CocoaPods: 1.16.2
    SDKs:
      iOS SDK:
        Platforms: DriverKit 25.0, iOS 26.0, macOS 26.0, tvOS 26.0, visionOS 26.0, watchOS 26.0
    IDEs:
      Android Studio: 2025.3
      Xcode: 26.0.1/17A400
    npmPackages:
      expo: ~57.0.4 => 57.0.4
      react: 19.2.3 => 19.2.3
      react-dom: 19.2.3 => 19.2.3
    Expo Workflow: managed
```

### Expo Doctor Diagnostics

```text
```

zoontek:
# Why

On web, `import.meta.url` was `null` when read after the bundle's synchronous execution (effects, async code, dynamic imports). It comes from `getBundleUrl()`, which read `document.currentScript` on every call, but that's only set while the script runs synchronously, so later reads returned `null`.

Fixes [#47754](https://github.com/expo/expo/issues/47754).

# How

Read `document.currentScript?.src` once at module load and cache it; reuse it on every `getBundleUrl()` call. The server branch is unchanged.

# Test Plan

Added `getBundleUrl.test.web.ts`. The key case clears `document.currentScript` before calling `getBundleUrl()` - it fails against the old per-call read, passes now.

```
$ jest src/utils/__tests__/getBundleUrl.test.web.ts
Tests:       3 passed, 3 total
```

# Checklist

### DIFF
--- a/packages/expo/CHANGELOG.md
+++ b/packages/expo/CHANGELOG.md
@@ -22,7 +22,7 @@
 - Fix `asyncRoutes` failing on Android and iOS with `Requiring unknown module` ([#46870](https://github.com/expo/expo/pull/46870) by [@hassankhan](https://github.com/hassankhan))
 - Fix `Response.blob()` in `expo/fetch` throwing when the global `Blob` is react-native's implementation. ([#47538](https://github.com/expo/expo/pull/47538) by [@kudo](https://github.com/kudo))
 - [iOS] Fix `Linking.getInitialURL()` returning `null` and deep links being dropped when a URL cold-starts an app on the UIKit scene life cycle. ([#47628](https://github.com/expo/expo/pull/47628) by [@tsapeta](https://github.com/tsapeta))
-- Fix `import.meta.url` being `null` on web when read after the bundle's synchronous execution (from effects, async code, or dynamic imports).
+- Fix `import.meta.url` being `null` on web when read after the bundle's synchronous execution (from effects, async code, or dynamic imports). ([#47802](https://github.com/expo/expo/pull/47802) by [@zoontek](https://github.com/zoontek))
 
 ### 💡 Others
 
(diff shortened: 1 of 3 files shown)


kitten (approved):
probably worth backporting

PR #47804, opened 2026-07-15
[expo-calendar][ios] Fix MissingPermissionsException typo

conanm:
# Why

The internal iOS exception class is misspelled: `MissionPermissionsException` (should be `Missing`). `expo-modules-core` derives the JS error `code` from the exception class name (`errorCodeFromString`), so this typo leaks all the way to app code — the permission error surfaces as **`ERR_MISSION_PERMISSIONS`** instead of `ERR_MISSING_PERMISSIONS`. Apps matching on that code have to hardcode the misspelling.

# How

Rename the class and its two throw sites (`CalendarModule.swift`, `Next/CalendarAccessGuard.swift`). The class is `internal`, so the Swift surface is unaffected — but the rename also corrects the derived JS error code from `ERR_MISSION_PERMISSIONS` → `ERR_MISSING_PERMISSIONS`.

Since that error-code string is observable from JS, I noted the change in the changelog under **🐛 Bug fixes**. Happy to move it to **🛠 Breaking changes** if you consider the derived code part of the public contract.

# Test Plan

Mechanical rename — one class definition + two throw sites, verified there are no remaining `Mission` references in `packages/expo-calendar/ios`. No behavior change beyond the corrected code string.

# Checklist

- [x] Updated the changelog for [any user-facing changes](https://github.com/expo/expo/blob/main/guides/contributing/Updating%20Changelog.md).


Wenszel (approved):
Hey, thanks for the fix! I'll get this merged when CI is green 🐆

PR #47811, opened 2026-07-15
[android][media-library] Fix createAssetAsync failing for files larger than ~2 GB

### LINKED ISSUE #47767: [expo-media-library] createAssetAsync fails for files larger than ~2.147 GB on Android: "Unable to copy file into external storage" (single unlooped FileChannel.transferTo)

https://github.com/wekor/expo-media-library-large-file-repro

### Summary

On Android, `MediaLibrary.createAssetAsync()` deterministically fails for any file larger than **~2.147 GB** with:

```
Call to function 'ExpoMediaLibrary.createAssetAsync' has been rejected.
→ Caused by: java.io.IOException: Unable to copy file into external storage.
```

even when the device has far more than enough free storage. Files below the threshold save fine. We hit this in production with 4 GB videos; free space on the affected devices was verified to be sufficient.

**Root cause** (present in `main` as of today): [`CreateAsset.kt` → `writeFileContentsToAsset`](https://github.com/expo/expo/blob/main/packages/expo-media-library/android/src/main/java/expo/modules/medialibrary/assets/CreateAsset.kt) performs a **single unlooped** `FileChannel.transferTo` and treats a partial transfer as out-of-space:

```kotlin
val transferred = input.transferTo(0, input.size(), output)
if (transferred != input.size()) {
  contentResolver.delete(assetUri, null, null)
  throw IOException("Could not save file to $assetUri Not enough space.")
}
```

`FileChannel.transferTo` is [documented](https://developer.android.com/reference/java/nio/channels/FileChannel#transferTo(long,%20long,%20java.nio.channels.WritableByteChannel)) to possibly transfer fewer bytes than requested and must be called in a loop. A single call transfers at most `Integer.MAX_VALUE` bytes at the Java layer, and the underlying `sendfile(2)` syscall caps a single call at `0x7ffff000` (2,147,479,552) bytes. So for any file above ~2.147 GB, `transferred != input.size()` is guaranteed, the asset entry is deleted, and the misleading "Not enough space" `IOException` is thrown — wrapped by `execute()` into "Unable to copy file into external storage".

The pass/fail boundary in the reproduction sits exactly at this ceiling (2.10 GB saves, 2.20 GB fails), which rules out actual storage exhaustion.

**Suggested fix**: loop `transferTo` until all bytes are transferred (treating repeated zero-progress iterations as a genuine failure). Happy to submit a PR.

iOS is unaffected (it goes through `PHAssetChangeRequest.creationRequestForAssetFromVideo`, no manual byte copy).

### Steps to reproduce

1. Clone the reproduction repo, `npm install`, `npx expo run:android` (development build; real device or emulator with ~10 GB free storage; package manager: npm).
2. Grant media permission when prompted.
3. Tap the four probe buttons top to bottom. Each one generates a dummy `.mp4` of the given size in the app cache (content is irrelevant — MediaStore copies bytes without inspecting them), then calls `MediaLibrary.createAssetAsync(file.uri)`. Free disk space is pre-checked before each probe so results are not confounded by real space pressure.

Expected: all four probes save to the media library.

Actual (Android, any API 30+ device):

| Probe | Result |
| --- | --- |
| 1.00 GB | ✅ saves |
| 2.10 GB | ✅ saves |
| 2.20 GB | ❌ `Unable to copy file into external storage` |
| 4.00 GB | ❌ `Unable to copy file into external storage` |

### Environment

```
    System:
      OS: macOS 26.5
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 20.20.2 - /Users/zephyr/.nvm/versions/node/v20.20.2/bin/node
      Yarn: 1.22.22 - /Users/zephyr/.nvm/versions/node/v20.20.2/bin/yarn
      npm: 10.8.2 - /Users/zephyr/.nvm/versions/node/v20.20.2/bin/npm
      Watchman: 2026.03.30.00 - /opt/homebrew/bin/watchman
    Managers:
      CocoaPods: 1.16.2 - /opt/homebrew/lib/ruby/gems/3.3.0/bin/pod
    SDKs:
      iOS SDK:
    IDEs:
      Android Studio: 2025.3 AI-253.31033.145.2533.15113396
    npmPackages:
      expo: ~55.0.27 => 55.0.27
      react: 19.2.0 => 19.2.0
    npmGlobalPackages:
      eas-cli: 18.6.0
    Expo Workflow: managed
```

expo-media-library: ~55.0.18

### Expo Doctor Diagnostics

```
Running 19 checks on your project...
```

jiunshinn:
# Why

Fixes #47767.

On Android, `MediaLibrary.createAssetAsync()` deterministically fails for any file larger than **~2.147 GB** with `java.io.IOException: Unable to copy file into external storage`, even when the device has plenty of free space. This was hit in production with 4 GB videos.

Root cause (confirmed on `main` by @wekor's report and @chrfalch): the file copy performs a single, unlooped `FileChannel.transferTo`. `transferTo` may copy fewer bytes than requested — the underlying `sendfile(2)` syscall caps one transfer at `0x7ffff000` (2,147,479,552) bytes — so for any file above ~2.147 GB the copy is guaranteed to be partial. The code treated the partial transfer as out-of-space, deleted the pending asset, and threw the misleading "Not enough space" `IOException`.

# How

The same unlooped-`transferTo` bug exists in **three** copy helpers in this package, all reachable from `createAssetAsync`/album copies:

- `assets/CreateAsset.kt` → `writeFileContentsToAsset` (content-resolver path, API 30+ — the exact site in the report)
- `MediaLibraryUtils.safeCopyFile` (legacy file copy, API 29 and below + album copies)
- `next/extensions/FileExtensions.kt` → `File.safeCopy` (the object-oriented API's copy path, now the default root API)

Rather than patch one site, I extracted a shared `FileChannel.transferAllTo` extension that loops `transferTo` until every byte is copied, treating a zero-progress iteration (a genuinely stuck transfer, e.g. the volume really is full) as a short return. Each call site keeps its existing `transferred != size` check, cleanup, and error message unchanged — the only change per site is swapping the single call for `transferAllTo`.

(The modern `next` content-resolver path already uses a buffered `InputStream.copyTo` loop and is unaffected.)

If the maintainers prefer to scope this to only the reported `CreateAsset.kt` site, I'm happy to narrow it.

# Test Plan

**Unit tests** (JVM/Robolectric, `:expo-media-library:testDebugUnitTest`): added `FileChannelExtensionsTests` covering
- a small single-call copy,
- the >2 GB case: `transferTo` returning the `0x7ffff000` cap on the first call must loop and copy the whole file (fails against the old single-call code),
- a stalled transfer returning a short count,
- a **non-mocked** case that wraps a real `FileChannel` destination so the real `transferTo` reports partial writes on a tiny file, verifying the loop copies all bytes and the content matches.

```
> Task :expo-media-library:testDebugUnitTest
BUILD SUCCESSFUL — 4 tests passed
> Task :expo-media-library:spotlessCheck
BUILD SUCCESSFUL
```

**Device reproduction** (Galaxy S21 5G, Android 15 / API 35): a local instrumented test copied a real ~3 GB file through the patched `MediaLibraryUtils.safeCopyFile`, exercising the real Android `sendfile` path.

Before the fix (single unlooped `transferTo`) — reproduces the reported failure:

```
SafeCopyLargeFileInstrumentedTest > safeCopyFile_copiesFilesLargerThan2GB[SM-G991N - 15] FAILED
  java.io.IOException: Could not save file to .../large-file-repro/dest Not enough space.
  at expo.modules.medialibrary.MediaLibraryUtils.safeCopyFile(MediaLibraryUtils.kt:76)
> Task :expo-media-library:connectedDebugAndroidTest FAILED
```

After the fix (looped `transferAllTo`) — the whole 3 GB file is copied:

```
Starting 1 tests on SM-G991N - 15
Finished 1 tests on SM-G991N - 15
> Task :expo-media-library:connectedDebugAndroidTest
BUILD SUCCESSFUL in 1m 27s
```

The instrumented test is not included in this PR — a >2 GB copy is too slow for CI; it was used only to gather the before/after evidence above.

# Checklist

- [ ] This diff will work correctly for `npx expo prebuild` & EAS Build — N/A, no config-plugin change.

### DIFF
--- a/packages/expo-media-library/CHANGELOG.md
+++ b/packages/expo-media-library/CHANGELOG.md
@@ -13,6 +13,7 @@
 ### 🐛 Bug fixes
 
 - [iOS] Fix resolving iCloud-offloaded assets ("Optimize iPhone Storage"): allow network access when extracting asset URIs so `Asset.getUri()`, `getInfo()`, `getExif()` and `getOrientation()` can download originals from iCloud, matching the legacy API's `shouldDownloadFromNetwork` default. ([#47790](https://github.com/expo/expo/pull/47790) by [@oeddyo](https://github.com/oeddyo))
+- [Android] Fix saving files larger than ~2 GB (e.g. `createAssetAsync` with large videos) failing with "Unable to copy file into external storage" by looping `FileChannel.transferTo` until the whole file is copied. ([#47811](https://github.com/expo/expo/pull/47811) by [@jiunshinn](https://github.com/jiunshinn))
 - Add `accessPrivileges` to `PermissionResponse` type ([#47177](https://github.com/expo/expo/pull/47177) by [@Wenszel](https://github.com/Wenszel))
 - [iOS] Fix permission guards for limited and write-only photo library access. ([#47216](https://github.com/expo/expo/pull/47216) by [@Wenszel](https://github.com/Wenszel))
 
--- a/packages/expo-media-library/android/src/main/java/expo/modules/medialibrary/FileChannelExtensions.kt
+++ b/packages/expo-media-library/android/src/main/java/expo/modules/medialibrary/FileChannelExtensions.kt
@@ -1,0 +1,19 @@
+package expo.modules.medialibrary
+
+import java.nio.channels.FileChannel
+import java.nio.channels.WritableByteChannel
+
+// A single transferTo copies at most ~2.147 GB (the sendfile(2) 0x7ffff000 cap), so loop until the
+// whole file is copied or the transfer stalls; otherwise larger files silently truncate (expo/expo#47767).
+fun FileChannel.transferAllTo(destination: WritableByteChannel): Long {
+  val size = size()
+  var position = 0L
+  while (position < size) {
+    val transferred = transferTo(position, size - position, destination)
+    if (transferred <= 0L) {
+      break
+    }
+    position += transferred
+  }
+  return position
+}
--- a/packages/expo-media-library/android/src/main/java/expo/modules/medialibrary/MediaLibraryUtils.kt
+++ b/packages/expo-media-library/android/src/main/java/expo/modules/medialibrary/MediaLibraryUtils.kt
@@ -70,7 +70,7 @@ object MediaLibraryUtils {
     }
     FileInputStream(src).channel.use { input ->
       FileOutputStream(newFile).channel.use { output ->
-        val transferred = input.transferTo(0, input.size(), output)
+        val transferred = input.transferAllTo(output)
         if (transferred != input.size()) {
           newFile.delete()
           throw IOException("Could not save file to $destDir Not enough space.")
--- a/packages/expo-media-library/android/src/main/java/expo/modules/medialibrary/assets/CreateAsset.kt
+++ b/packages/expo-media-library/android/src/main/java/expo/modules/medialibrary/assets/CreateAsset.kt
@@ -17,6 +17,7 @@ import expo.modules.medialibrary.MediaLibraryUtils
 import expo.modules.medialibrary.UnableToLoadPermissionException
 import expo.modules.medialibrary.UnableToSaveException
 import expo.modules.medialibrary.albums.getAlbumFileOrNull
+import expo.modules.medialibrary.transferAllTo
 import kotlinx.coroutines.Dispatchers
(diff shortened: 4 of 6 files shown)


chrfalch (approved):
Thanks, @jiunshinn - appreciate the fix!

PR #47825, opened 2026-07-15
[router] Improve standard-navigation types for createProps

Ubax:
# Why

Standard navigation integration distinguishes two types of props:
- `NavigatorProps` - these are the props which user needs to pass to the navigator, for example `<Stack prop1={v}>`
- `CreateProps` - these are the props created with `createProps` function. The main reason to use these, is to use expo-router internals to pass framework specific values. For example if someone wants to call a custom action not available in the standard-navigation contract then they need to add this as custom prop, e.g `preload: () => void`, and in both react-navigaton and expo-router add their own implementation, e.g. `createProps: ({navigation}) => ({ preload: navigation.preload })`. 

Since developer may not need the custom props, they don't need to specify any and they don't need to pass `createProps` function. However the typescript layer is too loose at the moment. Since `createProps` is optional, then all the `CreateProps` are marked as `Partial`. 

This PR fixes it - when `CreateProps` are specified then TS will require the developer to pass `createProps`

# How

1. Improve typescript for standard navigators - mark `createProps` required when custom props need to be passed and optional when they don't (`[keyof CreateProps] extends [never]`)
2. Fix TS errors in the tests and integrations

> I know that using `...[options]: IntegrateWithRouterOptionsTuple<State, CreateProps>` is not the cleanest syntax, but I was not able to find a better way to make this optional without adding 40 lines of function declaration overload.

# Test Plan

CI

# Checklist

### DIFF
--- a/packages/expo-router/src/standard-navigation/__tests__/types.test.tsx
+++ b/packages/expo-router/src/standard-navigation/__tests__/types.test.tsx
@@ -11,12 +11,17 @@ import {
 } from '../../react-navigation/routers';
 import type { GoBackAction, NavigateAction } from '../../react-navigation/routers/CommonActions';
 import { unstable_createStandardRouterNavigator, unstable_integrateWithRouter } from '../index';
-import type { NavigatorContentProps, StandardNavigationAction } from '../types';
+import type {
+  IntegrateWithRouterOptions,
+  NavigatorContentProps,
+  StandardNavigationAction,
+} from '../types';
 
 // Type-equality helpers
 type Expect<T extends true> = T;
 type Equal<A, B> =
   (<T>() => T extends A ? 1 : 2) extends <T>() => T extends B ? 1 : 2 ? true : false;
+function typecheck(_value: unknown) {}
 
 type Opts = { title?: string };
 type EventMap = { tabPress: { data: undefined; canPreventDefault: true } };
@@ -92,7 +97,7 @@ export const _options: OptionsFn = ({ route, theme }) => {
 
 type NavProps = { tintColor?: string };
 type CreateProps = { routeNames: string[]; preload: (name: string) => void };
-type SplitContentProps = NavigatorContentProps<Opts, EventMap, NavProps & CreateProps>;
+type SplitContentProps = NavigatorContentProps<Opts, EventMap, NavProps, CreateProps>;
 type RequiredNavProps = { label: string };
 
 function SplitContent(_props: SplitContentProps) {
@@ -140,6 +145,83 @@ export type _InferredElementRequiresPublicProp = Expect<
   Equal<InferredPublicElementProps['label'], string>
 >;
 
+const InferredSplitNav = unstable_createStandardRouterNavigator(SplitContent, TabRouter, {
+  createProps: () => ({ routeNames: [], preload: () => {} }),
+});
+type InferredSplitElementProps = ComponentProps<typeof InferredSplitNav>;
+export type _InferredElementAcceptsNavigatorProps = Expect<
+  Equal<InferredSplitElementProps['tintColor'], string | undefined>
+>;
+export type _InferredElementLacksRouteNames = Expect<
+  Equal<'routeNames' extends keyof InferredSplitElementProps ? true : false, false>
+>;
+export type _InferredElementLacksPreload = Expect<
+  Equal<'preload' extends keyof InferredSplitElementProps ? true : false, false>
+>;
+
+unstable_createStandardRouterNavigator(RequiredPublicContent, TabRouter, {
+  // @ts-expect-error This content does not declare any injected props.
+  createProps: () => ({ label: 1 }),
+});
+
+type CarrierCreateProps = { x: string };
+function CarrierContent(
+  _props: NavigatorContentProps<Opts, EventMap, object, CarrierCreateProps>
+) {
+  return null;
+}
+
+unstable_createStandardRouterNavigator<
+  Opts,
+  TabNavigationState<ParamListBase>,
+  EventMap,
+  object,
+  TabRouterOptions,
+  { x: number }
+  // @ts-expect-error Explicit CreateProps must match the content's declared CreateProps.
+>(CarrierContent, TabRouter, { createProps: () => ({ x: 1 }) });
+
+unstable_createStandardRouterNavigator<
+  Opts,
+  TabNavigationState<ParamListBase>,
+  EventMap,
+  RequiredNavProps,
+  TabRouterOptions
+  // @ts-expect-error Five explicit generics declare no injected props, so `createProps` is forbidden.
+>(RequiredPublicContent, TabRouter, { createProps: () => ({ label: 1 }) });
+
+const broadlyAnnotatedFactoryOptions: IntegrateWithRouterOptions = {
+  // @ts-expect-error Bare options do not declare injected props, so `createProps` is forbidden.
+  createProps: () => ({ label: 1 }),
+};
+unstable_createStandardRouterNavigator(
+  RequiredPublicContent,
+  TabRouter,
+  broadlyAnnotatedFactoryOptions
+);
+
+// @ts-expect-error Inferred non-empty CreateProps require the options argument.
+unstable_createStandardRouterNavigator(SplitContent, TabRouter);
+
+// `NoInfer` keeps a zero-argument factory from declaring injected props for content that has none.
+unstable_createStandardRouterNavigator(PublicContent, TabRouter, {
+  // @ts-expect-error `PublicContent` does not declare any injected props.
+  createProps: () => ({ injected: true }),
+});
+
+type OptionalCreateProps = { a?: string };
+function OptionalCreateContent(
+  _props: NavigatorContentProps<Opts, EventMap, object, OptionalCreateProps>
+) {
+  return null;
+}
+
+// @ts-expect-error CreateProps with optional keys still require options.
+unstable_createStandardRouterNavigator(OptionalCreateContent, TabRouter);
+unstable_createStandardRouterNavigator(OptionalCreateContent, TabRouter, {
+  createProps: () => ({}),
+});
+
 // @ts-expect-error The options argument is required when `CreateProps` is non-empty.
 unstable_createStandardRouterNavigator<
   Opts,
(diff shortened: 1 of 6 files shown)


Ubax:
* **#47825** <a href="https://app.graphite.com/github/pr/expo/expo/47825?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 3  dependent PRs ([#46536](https://github.com/expo/expo/pull/46536) <a href="https://app.graphite.com/github/pr/expo/expo/46536?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#46592](https://github.com/expo/expo/pull/46592) <a href="https://app.graphite.com/github/pr/expo/expo/46592?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#47839](https://github.com/expo/expo/pull/47839) <a href="https://app.graphite.com/github/pr/expo/expo/47839?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/expo/expo/47825?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/expo-router/src/standard-navigation/index.tsx
```diff
@@ -162,7 +199,7 @@ export function unstable_integrateWithRouter<
           // its type, so the TS contract keeps users from reading router options off `NavigatorContent`
           // in the common case, even though they are physically present on the object.
           {...(extraProps as unknown as NavigatorProps)}
-          {...derivedProps}
+          {...(derivedProps as CreateProps)}
```
Ubax, line 205: Add a comment here explaining the casting

packages/expo-router/src/standard-navigation/index.tsx
```diff
@@ -106,13 +132,24 @@ export function unstable_integrateWithRouter<
...
-  navigator: StandardNavigator<NavigatorOptions, EventMap, NavigatorProps>,
+  navigator: StandardNavigator<
+    NavigatorOptions,
+    EventMap,
+    NavigatorPropsWithCreateProps<NavigatorProps, CreateProps>
+  >,
   router: RouterFactory<State, NavigationAction, RouterOptions>,
-  options?: IntegrateWithRouterOptions<State, NavigatorProps>
+  // A rest tuple is the only way to make the whole argument optional for empty `CreateProps` and
+  // required otherwise; a normal optional parameter would allow it in both cases.
```
Ubax:
```suggestion
  // A rest tuple is the only way to make the whole argument optional for empty `CreateProps` and
  // required otherwise; a normal optional parameter would accept `undefined` in both cases.
```

packages/expo-router/src/standard-navigation/index.tsx
```diff
@@ -61,23 +75,35 @@ export function unstable_createStandardRouterNavigator<
...
   return unstable_integrateWithRouter<
     NavigatorOptions,
     State,
     EventMap,
     NavigatorProps,
-    RouterOptions
-  >(navigator, router, options);
+    RouterOptions,
+    CreateProps
+  >(navigator, router, options!);
```
Ubax: Instead of `options!` do `options ?? {}`

packages/expo-router/src/standard-navigation/index.tsx
```diff
@@ -61,23 +75,35 @@ export function unstable_createStandardRouterNavigator<
...
+    NavigatorContentProps<
+      NavigatorOptions,
+      EventMap,
+      NavigatorPropsWithCreateProps<NavigatorProps, CreateProps>
+    >
   >,
   router: RouterFactory<State, NavigationAction, RouterOptions>,
-  options?: IntegrateWithRouterOptions<State, NavigatorProps>
+  // A rest tuple is the only way to make the whole argument optional for empty `CreateProps` and
+  // required otherwise; a normal optional parameter would allow it in both cases.
```
Ubax: Same comment change here

packages/expo-router/src/standard-navigation/index.tsx
```diff
@@ -36,6 +36,20 @@ export type {
...
 const STANDARD_NAVIGATOR_TYPE = 'standard';
 
+type NavigatorPropsWithCreateProps<NavigatorProps extends object, CreateProps extends object> = [
+  keyof CreateProps,
+] extends [never]
+  ? NavigatorProps
+  : NavigatorProps & CreateProps;
+
+// A rest tuple is the only way to make the whole argument optional for empty `CreateProps` and
+// required otherwise; a normal optional parameter would allow it in both cases.
```
Ubax: And here as well

packages/expo-router/src/standard-navigation/__tests__/types.test.tsx
```diff
@@ -86,6 +86,190 @@ export const _options: OptionsFn = ({ route, theme }) => {
   return { title: `${route.name}-${theme.dark}` };
 };
 
+// ---------------------------------------------------------------------------
+// CreateProps: content-only props injected by createProps
+// ---------------------------------------------------------------------------
```
Ubax: Can you have two sections here:
1. CreateProps - unstable_createStandardRouterNavigator: content-only props injected by createProps
2. CreateProps - unstable_integrateWithRouter: content-only props injected by createProps

packages/expo-router/src/standard-navigation/index.tsx
```diff
@@ -61,23 +75,33 @@ export function unstable_createStandardRouterNavigator<
   EventMap extends StandardNavigatorEventMapBase,
   NavigatorProps extends object,
   RouterOptions extends DefaultRouterOptions,
+  CreateProps extends object = object,
```
Ubax, line 91: [🏗️ **design**] The fix only takes effect when all six type arguments are passed explicitly. I verified with a `tsc` probe against this PR's head: using the docs-guide call form (full inference, content declaring the injected props in `NavigatorContentProps`'s third argument, `createProps` supplied) —

```tsx
const Tabs = unstable_createStandardRouterNavigator(TabsContent, TabRouter, {
  createProps: ({ state, dispatch }) => ({ activeRouteKey: ..., preload: ... }),
});
```

— `activeRouteKey` and `preload` end up as **required** props on the `<Tabs>` element (`ComponentProps<typeof Tabs>['activeRouteKey']` is `string`, required). 

Make inference split correctly (e.g. inferring `NavigatorProps` from the content minus the `createProps` return via an extra overload),

packages/expo-router/src/standard-navigation/__tests__/types.test.tsx
```diff
@@ -86,6 +86,194 @@ export const _options: OptionsFn = ({ route, theme }) => {
...
+  Equal<SplitContentProps['preload'], (name: string) => void>
+>;
+export type _ElementLacksRouteNames = Expect<
+  Equal<'routeNames' extends keyof SplitElementProps ? true : false, false>
+>;
+export type _ElementLacksPreload = Expect<
+  Equal<'preload' extends keyof SplitElementProps ? true : false, false>
+>;
+
+const InferredPublicNav = unstable_createStandardRouterNavigator(RequiredPublicContent, TabRouter);
```
Ubax, line 173: [💡 suggestion] This is the only inference-form test, and it has no `createProps`. The docs guide's "Options" example uses inference *with* `createProps`, and in that combination the injected props leak onto the element as required props (see comment in `index.tsx`). Add an inference test mirroring the docs pattern (content declaring injected props, `createProps` supplied, no explicit generics) asserting the element does not require them — right now the guide-mirror block below also diverges from the docs, since its `TabsContent` never declares the injected props.

packages/expo-router/src/standard-navigation/index.tsx
```diff
@@ -106,13 +130,24 @@ export function unstable_integrateWithRouter<
...
 >(
-  navigator: StandardNavigator<NavigatorOptions, EventMap, NavigatorProps>,
+  navigator: StandardNavigator<
+    NavigatorOptions,
+    EventMap,
+    NavigatorPropsWithCreateProps<NavigatorProps, CreateProps>
+  >,
   router: RouterFactory<State, NavigationAction, RouterOptions>,
-  options?: IntegrateWithRouterOptions<State, NavigatorProps>
+  // A rest tuple is the only way to make the whole argument optional for empty `CreateProps` and
```
Ubax: [🐧 nit] This same rest-tuple comment appears three times (here and on both function signatures). One explanation on the type definition is enough — the parameter sites can just say `// see IntegrateWithRouterOptionsTuple`.

packages/expo-router/src/standard-navigation/index.tsx
```diff
@@ -106,10 +143,11 @@ export function unstable_integrateWithRouter<
   EventMap extends StandardNavigatorEventMapBase,
   NavigatorProps extends object,
   RouterOptions extends DefaultRouterOptions,
+  CreateProps extends object = object,
 >(
-  navigator: StandardNavigator<NavigatorOptions, EventMap, NavigatorProps>,
+  navigator: StandardNavigator<NavigatorOptions, EventMap, NavigatorProps & CreateProps>,
   router: RouterFactory<State, NavigationAction, RouterOptions>,
-  options?: IntegrateWithRouterOptions<State, NavigatorProps>
+  ...[options]: IntegrateWithRouterOptionsTuple<State, NoInfer<CreateProps>>
```
kitten, line 150: I don't think this is very idiomatic. Are we doing this to preserve `.length` on the function?

> A rest tuple is the only way to make the whole argument optional for empty `CreateProps` and required otherwise

Not necessarily. If we structure this as an `extends ... ? A : B` case, we can potentially just use two interfaces, where one has the appropriate overload.

I think the comment having to explain this might merit that, but I think a direct comparison could help here to see if it works? It might end up being cleaner

packages/expo-router/src/standard-navigation/index.tsx
```diff
@@ -61,23 +88,33 @@ export function unstable_createStandardRouterNavigator<
   EventMap extends StandardNavigatorEventMapBase,
   NavigatorProps extends object,
   RouterOptions extends DefaultRouterOptions,
+  CreateProps extends object = object,
```
kitten, line 91: I'm not sure of the function of `CreateProps` yet (didn't read all of it just yet) but isn't the idiomatic `extends`-type constraint instead `{ [key: string | number]: any }`?

packages/expo-router/src/standard-navigation/index.tsx
Ubax: 1. CreateProps -> createProps function should bre required
2. navigator -> infer CreateProps and make function required
3. createProps should be optional

kitten (approved):

packages/expo-router/src/standard-navigation/index.tsx
Ubax: Added a ticket as a follow-up - https://linear.app/expo/issue/ENG-25346/simplify-and-refactor-standard-navigation-types

PR #47828, opened 2026-07-15
[video][ios] Fix player not releasing in SDK 57

### LINKED ISSUE #47569: [expo-video] SDK 57: audio keeps playing after the player is released on unmount (regression vs SDK 56)

https://github.com/oeddyo/reproducible-sound-continues

### Steps to reproduce

Created with `npx create-expo-app@latest --template blank` (npm) + `npx expo install expo-video` and a bundled 30s sample video with a clearly audible 440 Hz tone. Nothing else.

1. `npm install && npx expo run:ios` (**iOS development build**; observed on both the iOS 18 simulator and an iOS 26.5 device)
2. Tap **Open video** — a fullscreen `Modal` mounts a component that calls `useVideoPlayer(source, (p) => p.play())` and renders a `VideoView`. The tone is clearly audible.
3. Tap **Close** — the modal's content unmounts, so `useVideoPlayer` releases the player.
4. **What I expected:** playback stops, as it did on SDK 56 (same code, expo-video 56.1.4).
5. **What actually happens:** the JS player is released — any subsequent call throws `FunctionCallException → NotFoundException: Unable to find the native shared object` (`SyncFunctionDefinition.swift:94` → `DynamicSharedObjectType.swift:60`) — **but the native AVPlayer keeps playing; the tone continues in the background indefinitely.**

The same happens with expo-audio's `useAudioPlayer` (a looping music player keeps playing after unmount), which suggests this is not specific to expo-video.

## Why this looks like an expo-modules-core regression

- The **iOS sources of expo-video 56.1.4 and 57.0.0 are byte-identical** (verified by diffing the published npm tarballs), so the change in behavior does not come from expo-video itself.
- `VideoPlayer` stops playback only in `deinit` (`DispatchQueue.main.async { ref.replaceCurrentItem(with: nil) }`). Nothing on the release path pauses: `sharedObjectWillRelease()` / `sharedObjectDidRelease()` are not overridden by `VideoPlayer`.
- expo-modules-core 57 introduced `SharedObjectNativeState` (`ios/Core/SharedObjects/SharedObjectNativeState.swift`), which holds a **strong** `let native: SharedObject` reference attached to the JS object and retained (via `Unmanaged.passRetained`) until the JS object's C++ native state is destroyed — i.e. until **JS garbage collection**, not until `release()`.
- `SharedObjectRegistry.delete()` removes the id pair but does **not** detach that native state. So after `release()` the native player can stay alive until Hermes collects the JS wrapper — and during that window nothing has stopped the AVPlayer. In core 56, the registry pair itself held the strong reference, so `release()` → prompt `deinit` → playback stopped.
- Side effect: since `getAdditionalMemoryPressure()` defaults to 0, each released-but-uncollected player invisibly pins an AVPlayer + buffered AVPlayerItem.

Related: #33804 / #34050 (identical symptom in 2024/2025, fixed in expo-video 2.0.6), #39427 (the NotFoundException after release), #47458 (expo-audio sibling report, possibly the same root cause).

**Workaround** we're using in production — own the player lifecycle so a cleanup can pause before releasing (a pause-cleanup declared after `useVideoPlayer` runs too late: same-component cleanups run in declaration order and the hook's internal release wins):

```tsx
const [player] = useState(() => createVideoPlayer(source));
useEffect(() => {
  player.play();
  return () => {
    player.pause();
    player.release();
  };
}, [player]);
```

### Environment

```
    System:
      OS: macOS 26.5
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 22.13.0 - ~/.nvm/versions/node/v22.13.0/bin/node
      npm: 10.9.2 - ~/.nvm/versions/node/v22.13.0/bin/npm
    Managers:
    SDKs:
      iOS SDK:
    IDEs:
    npmPackages:
      expo: ~57.0.2 => 57.0.2
      expo-video: ~57.0.0 => 57.0.0
      react: 19.2.3 => 19.2.3
    Expo Workflow: managed
```

### Expo Doctor Diagnostics

```
(expo-doctor@1.20.0)
```

behenate:
# Why

Fixes https://github.com/expo/expo/issues/47569
SDK57 introduced some changes to SharedObject lifecycle - the information of the shared object releasing is delivered in `sharedObjectWillRelease` while the `deinit` itself is run later. 

# How

Implement `sharedObjectWillRelease` and move the release logic there. Also some threading improvements

# Test Plan

Tested in BareExpo on iOS

### DIFF
--- a/packages/expo-video/ios/VideoPlayer.swift
+++ b/packages/expo-video/ios/VideoPlayer.swift
@@ -13,6 +13,7 @@ internal final class VideoPlayer: SharedRef<AVPlayer>, Hashable, VideoPlayerObse
   lazy var audioTracks: VideoPlayerAudioTracks = VideoPlayerAudioTracks(owner: self)
   lazy var seeker: VideoPlayerSeeker = VideoPlayerSeeker(player: self)
   private var tracksLoadingTask: Task<(), Never>?
+  private let didRelease = Mutex(false)
 
   var loop = false
   var audioMixingMode: AudioMixingMode = .auto {
@@ -181,19 +182,18 @@ internal final class VideoPlayer: SharedRef<AVPlayer>, Hashable, VideoPlayerObse
   }
 
   deinit {
-    observer?.notifyPlayerDeinit(player: self)
-    observer?.cleanup()
-    NowPlayingManager.shared.unregisterPlayer(self)
-    VideoManager.shared.unregister(videoPlayer: self)
+    releasePlayer()
+  }
 
-    videoSourceLoader.cancelCurrentTask()
-    tracksLoadingTask?.cancel()
-
-    // We have to replace from the main thread because of KVOs (see comment in VideoSourceLoader).
-    // Moreover, in this case we have to keep a strong reference to AVPlayer and remove its item
-    // If we don't do this AVPlayer doesn't get deallocated
-    DispatchQueue.main.async { [ref] in
-      ref.replaceCurrentItem(with: nil)
+  override func sharedObjectWillRelease() {
+    if Thread.isMainThread {
+      releasePlayer()
+    } else {
+      // Strong self capture is intentional: it keeps the player alive until the teardown
+      // has run on the main thread, so `deinit` can never race with it.
+      DispatchQueue.main.async {
+        self.releasePlayer()
+      }
     }
   }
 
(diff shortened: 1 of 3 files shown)


packages/expo-video/ios/VideoPlayer.swift
```diff
@@ -181,19 +182,18 @@ internal final class VideoPlayer: SharedRef<AVPlayer>, Hashable, VideoPlayerObse
...
+  override func sharedObjectWillRelease() {
+    if Thread.isMainThread {
+      releasePlayer()
+    } else {
+      // Strong self capture is intentional: it keeps the player alive until the teardown
+      // has run on the main thread, so `deinit` can never race with it.
+      DispatchQueue.main.async {
+        self.releasePlayer()
+      }
     }
```
tsapeta: Please, no... 😅 We're repeating this bad pattern in too many places. It should be defined in one place instead. Sadly it's currently not exported publicly by `Utilities.swift` from `expo-modules-core`, maybe it should?
Since we don't want to break the compatibility (core ↔ video) and you probably plan to backport it, can you at least extract it in `expo-video`? You also repeat it in `releasePlayer`.

tsapeta (approved):

PR #47830, opened 2026-07-15
[updates] Don't set isUpdatePending when a download completes with no new update

Kudo:
# Why

Fixes [ENG-23675](https://linear.app/expo/issue/ENG-23675/investigate-isupdatepending-after-channel-switch-reload).

After a channel switch (`setUpdateRequestHeadersOverride` → `fetchUpdateAsync` → `reloadAsync`), `isUpdatePending` could come back `true` on the next load even though no new update had been downloaded, surfacing a spurious "An update is available" banner.

Root cause: the updates state machine's bare `downloadComplete` event unconditionally set `isUpdatePending = true`. That event is emitted on paths where the download finished **without** producing a new launchable update — a `fetchUpdateAsync` that finds nothing new (server had nothing, or the update was rejected by the selection policy), and the startup no-update-while-downloading path. So a no-op fetch (which the demo's overlay triggers automatically after a check reports availability) flipped `isUpdatePending` on with nothing actually pending.

This contradicts the documented contract of `isUpdatePending`: *"True if a new available update is available and has been downloaded."*

# How

`downloadComplete` now sets `isUpdatePending = false`. Nothing was downloaded on that path, so nothing is pending. A genuinely downloaded update or a rollback continues to flow through `downloadCompleteWithUpdate` / `downloadCompleteWithRollback`, which set `isUpdatePending = true` — so real pending updates are unaffected.

- **iOS** – `UpdatesStateMachine` reducer flips the flag; `FetchUpdateProcedure`'s rollback branch now emits `.downloadCompleteWithRollback` (a pending rollback) instead of the generic `.downloadComplete`. This matches what Android already does and revives the previously-unused `.downloadCompleteWithRollback` event on iOS.
- **Android** – one-line reducer mirror. Its procedures already emit `DownloadCompleteWithRollback()` / bare `DownloadComplete()` correctly, so no procedure changes were needed.

`downloadProgress` behavior is intentionally left as-is (`1.0`) to keep both platforms identical.

# Test Plan

Verified against Keith's [pancake-theory](https://github.com/keith-kurak/pancake-theory) repro: channel surf → reload no longer shows the "An update is available" banner, and `isUpdatePending` stays `false`.

# Checklist

- [x] I added a changelog.md entry and rebuilt the package sources according to [this short guide](https://github.com/expo/expo/blob/main/CONTRIBUTING.md#-before-submitting)
- [x] This diff will work correctly for npx expo prebuild & EAS Build (eg: updated a module plugin).

### DIFF
--- a/packages/expo-updates/CHANGELOG.md
+++ b/packages/expo-updates/CHANGELOG.md
@@ -13,6 +13,7 @@
 - [Android] Widen `UpdatesLogEntry.create`'s catch from `JSONException` to `Exception` so log-line parse failures consistently degrade to "skip the entry" instead of propagating. ([#46182](https://github.com/expo/expo/pull/46182) by [@jakequade-pc](https://github.com/jakequade-pc))
 - [Android] Correct `UpdatesLogReader.ONE_DAY_MILLISECONDS` from `86400` (seconds) to `86_400_000` (milliseconds), so the "older than one day" purge filter actually retains a day's worth of entries instead of ~86 seconds' worth. ([#46182](https://github.com/expo/expo/pull/46182) by [@jakequade-pc](https://github.com/jakequade-pc))
 - [iOS] Isolate UpdatesLogReaderTests from concurrent suites. ([#47082](https://github.com/expo/expo/pull/47082) by [@douglowder](https://github.com/douglowder))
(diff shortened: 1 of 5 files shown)


alanjhughes (approved):

douglowder (approved):

Kudo:
apply this patch to UpdatesAPIDemo and confirm it doesn't have regression. going to merge it

PR #47835, opened 2026-07-16
[JSI] Avoid per-call refcount churn on JavaScriptValuesBuffer.runtime

tsapeta:
## Why

`JavaScriptValuesBuffer` is constructed for every synchronous host-function call. Its `runtime` field was a plain `unowned` reference, so each call paid an `incrementUnowned` on construction and a side-table `decrementUnownedShouldFree` on destruction. Profiling a no-op `@JS` host call (release, on device) showed this unowned reference-counting churn as a meaningful part of the Swift-side per-call floor.

## How

Change `JavaScriptValuesBuffer.runtime` from `unowned` to `unowned(unsafe)`, dropping the reference-counting traffic.

There are exactly two places a buffer is created, and both are call-scoped with the runtime held alive by the caller for the whole duration:

1. The host-function trampolines (`createFunctionClosure`) build a buffer from the raw C++ argument pointer and consume it synchronously within the host call; the runtime is invoking the function, so it is necessarily alive. This became strictly call-scoped only after #47755 made async host functions two-phase so the arguments buffer no longer crosses the asynchronous boundary (that PR also removed the last production use of `copy()`). Before it, an async binding could carry the buffer into an async body that outlives the synchronous call, which is exactly the case where an `unowned(unsafe)` runtime could dangle; that path no longer exists.
2. `JavaScriptFunction.call` / `callAsConstructor` allocate an owning buffer, pass it into the synchronous call, and drop it when the call returns; the enclosing `JavaScriptFunction` holds the runtime throughout.

So in both cases the runtime outlives the buffer and the liveness re-check is unnecessary. The API also can't be misused from outside: `JavaScriptValuesBuffer` has no public initializer (all `init`s are `internal`); the only public ways to obtain one are the `allocate`/`copying` factories, which all take `in runtime:` and produce a buffer meant to invoke a JS function synchronously.

Performance-wise, the hot decode path (`unownedValue(at:)`, `set`) reads the refcount-free `iRuntime`, not this wrapper; only the cold `subscript`/`copy` paths touch the wrapper. So on the hot path the reference was stored and destroyed without ever being read. No behavior change; only the refcount traffic is removed on a path where the runtime is provably alive.

## Test Plan

- `swift build` of `expo-modules-jsi` is clean.
- On device (release), the affected argument-decode host-call rows in the bare-expo modules-benchmarks screen improved with no behavior change.


tsapeta:
Actually I'm dropping this idea. Codex's review is correct. It would be safe in our hot path, but not necessarily in the cold path. I'll need to think more how to redesign the API of `JavaScriptValuesBuffer` to make it not store the `JavaScriptRuntime` at all.
`let runtime: Unmanaged<JavaScriptRuntime>` would avoid retain/release, but I'm a bit hesitant to do this. Perf gains from `unowned(unsafe)` are not that big to make it worth it.

PR #47837, opened 2026-07-16
[JSI] Stop `JavaScriptActor.assumeIsolated` from boxing its operation

tsapeta:
## Why

The synchronous host-function trampolines wrap their body in `JavaScriptActor.assumeIsolated`. Profiling a no-op `@JS` host call (release, on device) showed a per-call heap allocation (`swift_allocObject`) plus its retain/release dominating the Swift-side cost. The cause: `assumeIsolated` routed its `operation` through `withoutActuallyEscaping`, which defeats closure stack promotion and forces the operation's captured context onto the heap every call.

## How

Keep `operation` non-escaping in both overloads:

- The non-throwing overload casts a capture-free static runner (`runIsolated`) from isolated to non-isolated and passes `operation` to it as an ordinary non-escaping argument. With no reabstraction thunk over `operation`, its captures stay on the caller's stack and no box is allocated.
- The throwing overload becomes typed-throws (`throws(E)`) and delegates to the non-throwing fast path via `Result(catching:)`, avoiding `withoutActuallyEscaping` while preserving the exact error type. Typed throws also sidesteps the newer-OS runtime support that casting a `throws` function value would require.

Both are `@_alwaysEmitIntoClient @inline(__always)`, so the runner indirection and casts fold away at `-O`. This benefits every `assumeIsolated { … }` caller, not just the host-function trampolines. No behavior change: `checkIsolated()` keeps the same thread-name assertion.

## Test Plan

- `swift build` of `expo-modules-jsi` is clean; added `JavaScriptActorTests` coverage (runs under the package's iOS `xcodebuild test` suite).
- On device (release), the no-op `@JS` host call dropped from ~168 ns/call to ~103 ns/call after this change (~1.6x).

### DIFF
--- a/packages/expo-modules-jsi/apple/Sources/ExpoModulesJSI/Runtime/JavaScriptActor.swift
+++ b/packages/expo-modules-jsi/apple/Sources/ExpoModulesJSI/Runtime/JavaScriptActor.swift
@@ -1,8 +1,5 @@
 import Foundation
 
-/// Name of the JavaScript thread created by React Native. Copied from `RCTJSThreadManager.mm`.
-private let jsThreadName = "com.facebook.react.runtime.JavaScript"
-
 /// Global actor that is used to isolate the code that should only be executed from the JavaScript thread.
 /// Theoretically it does not act as a real actor; it uses a serial executor that executes jobs **synchronously**
 /// without hopping to the proper thread. Meaning that running these jobs on the JavaScript thread must be ensured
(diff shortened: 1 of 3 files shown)


packages/expo-modules-jsi/apple/Sources/ExpoModulesJSI/Runtime/JavaScriptActor.swift
```diff
@@ -23,25 +20,56 @@ public actor JavaScriptActor: GlobalActor {
...
-    shared.executor.checkIsolated()
+    // Using `assert` instead of `precondition` because this check is a heuristic based on
+    // thread name, not a precise isolation guarantee. Worklet runtimes legitimately run on
+    // the UI thread, which would cause a false-positive crash with `precondition`.
+    assert(
+      // JavaScript thread name copied from `RCTJSThreadManager.mm`.
+      Thread.current.name == "com.facebook.react.runtime.JavaScript" || !Thread.isMultiThreaded()
+        || ProcessInfo.processInfo.processName == "xctest",
+      "JavaScriptActor operations must be run on the JavaScript thread"
+    )
```
tsapeta, line 66: I didn't really like this assert and I'm going to remove it in a follow-up PR. It's always stripped as `expo-modules-jsi` is always prebuilt in the release mode. Even if someone would force to use debug, it will trap for worklets. This PR just moves it around.

Kudo (approved):

PR #47838, opened 2026-07-16
Add jsBundleFilePathLoader parameter to ExpoReactHostFactory

NiuGuohui:
# Why

Sometimes, applications developed in brownfield mode require dynamic modification of the file path of the JS bundle. By adding `jsBundleFilePathLoader`, developers can more easily control the source of the Bundle.

# Checklist


chrfalch:
Hi, @NiuGuohui - thanks for your contribution. Please add changelog entries, and also - if possible - maybe a test verifying that this is working, or a way to test it if possible. Also consider if this should be added to iOS as well?

PR #47848, opened 2026-07-16
[router] add useRouterColor hook

Ubax:
# Why

At the moment there is no way to react to material you scheme changes in react-native. This is a follow-up to https://github.com/expo/expo/pull/47423, which prevented recreation of an activity when material you scheme changed.

# How

1. Add native listener to listen to configuration changes on Android (`CONFIG_ASSETS_PATHS`)
2. Add a hook returning new colors when material or theme scheme changes

# Test Plan

# Checklist


Ubax:
* **#47848** <a href="https://app.graphite.com/github/pr/expo/expo/47848?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/47848?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/expo-router/src/color/useRouterColor.ts
```diff
@@ -0,0 +1,58 @@
...
+ */
+export function useRouterColor(): ColorType {
+  const colorScheme = useColorScheme();
+  const [color, setColor] = useState<ColorType>(createColor);
+  // Adjust state during render on scheme changes (avoids an extra effect pass).
+  const [previousScheme, setPreviousScheme] = useState<ColorSchemeName>(colorScheme);
+  if (previousScheme !== colorScheme) {
+    setPreviousScheme(colorScheme);
+    setColor(createColor());
+  }
```
Ubax: Make this an effect as well

packages/expo-router/src/color/useRouterColor.ts
```diff
@@ -0,0 +1,58 @@
...
+ * Re-renders on light/dark scheme changes and on native Material You palette
+ * change events, without exposing any value.
+ *
+ * @internal
+ */
+export function useColorPaletteVersion(): void {
+  useColorScheme();
+  const [, increment] = useReducer((current: number) => current + 1, 0);
+  useEffect(() => addColorPaletteListener(increment), []);
+}
```
Ubax: Remove this hook. It's not needed

packages/expo-router/android/src/main/java/expo/modules/router/ExpoRouterModule.kt
```diff
@@ -9,20 +11,74 @@ import expo.modules.kotlin.modules.ModuleDefinition
 import androidx.appcompat.view.ContextThemeWrapper
 import com.google.android.material.color.MaterialColors
 
+private const val COLOR_PALETTE_CHANGED_EVENT = "onColorPaletteChanged"
+
+// `ActivityInfo.CONFIG_ASSETS_PATHS` — inlined because the constant is public API
+// only since compileSdk 36, while the framework sets the bit since API 26.
+private const val CONFIG_ASSETS_PATHS = 0x80000000.toInt()
```
Ubax: Can I use it here? What is the compileSdk for expo sdk 56?

packages/expo-router/src/color/useRouterColor.ts
```diff
@@ -0,0 +1,45 @@
...
+ * @platform android
+ * @platform ios
+ */
+export function useRouterColor(): ColorType {
+  const colorScheme = useColorScheme();
+  const [color, setColor] = useState<ColorType>(createColor);
+
+  useEffect(() => {
+    setColor(createColor());
+  }, [colorScheme]);
```
Kudo: [💡 suggestion] The `[colorScheme]` effect also runs on mount, so `setColor(createColor())` fires once right after the first render — every consumer re-renders and `color` gets a new identity even though nothing changed. Since the hook is documented as producing a new identity only “whenever colors may have changed” and as safe in a `useMemo`/`useEffect` dependency array, a consumer effect keyed on `color` will run twice on mount. Consider guarding the first run (e.g. a `useRef` first-render flag) so the identity only changes on an actual scheme change.

packages/expo-router/android/src/main/java/expo/modules/router/ExpoRouterModule.kt
```diff
@@ -9,20 +12,64 @@ import expo.modules.kotlin.modules.ModuleDefinition
...
+
 class ExpoRouterModule : Module() {
   private val context: Context
     get() = appContext.reactContext ?: throw Exceptions.ReactContextLost()
 
+  @Volatile
+  private var lastConfiguration: Configuration? = null
+
+  // Retained for cleanup after the React context is destroyed.
+  private var registeredApplication: Application? = null
```
Kudo:
```suggestion
  private val registeredApplication: Application?
    get() = appContext.reactContext?.applicationContext as? Application
```

do we need to retain the Application?

docs/pages/router/reference/color.mdx
```diff
@@ -97,27 +97,27 @@ The available colors are the same as [Material 3 static colors](#material-design
 
 ### Responding to theme changes on Android
 
-Android Material colors (both static and dynamic) respond to the system's light/dark mode. To ensure your component re-renders when the theme changes, use the `useColorScheme()` hook from React Native.
+Android Material colors respond to the system's light/dark mode, and dynamic colors additionally change when the user updates their Material You colors (for example, by changing the wallpaper). To make your component re-render in both cases, use the `useRouterColor()` hook and read colors from the object it returns.
 
 ```tsx
-import { Color } from 'expo-router';
-import { View, Text, useColorScheme } from 'react-native';
+import { useRouterColor } from 'expo-router';
```
Kudo: the "Router" in useRouterColor feels redundant. may use different names, e.g. just `useColor`

packages/expo-router/android/src/main/java/expo/modules/router/ExpoRouterModule.kt
```diff
@@ -9,20 +12,64 @@ import expo.modules.kotlin.modules.ModuleDefinition
...
+
+  // Dynamic color overlays change the assets path.
+  private val componentCallbacks = object : ComponentCallbacks {
+    override fun onConfigurationChanged(newConfig: Configuration) {
+      val previous = lastConfiguration
+      lastConfiguration = Configuration(newConfig)
+      if (previous != null && previous.diff(newConfig) and ActivityInfo.CONFIG_ASSETS_PATHS != 0) {
```
Kudo:
```suggestion
      if (previous?.diff(newConfig) && ActivityInfo.CONFIG_ASSETS_PATHS != 0) {
```

- use `?.` if possible
- not mixing `and`, `&&`

packages/expo-router/android/src/main/java/expo/modules/router/ExpoRouterModule.kt
```diff
@@ -9,20 +12,64 @@ import expo.modules.kotlin.modules.ModuleDefinition
...
       materialColor(name, scheme)
     }
 
     Function("Material3DynamicColor") { name: String, scheme: String ->
       dynamicColor(name, scheme)
     }
+
+    OnStartObserving(COLOR_PALETTE_CHANGED_EVENT) {
+      unregisterComponentCallbacks()
+      registeredApplication = (appContext.reactContext?.applicationContext as? Application)?.also {
```
Kudo: if registeredApplication can be a computed prop, we can rewrite this

packages/expo-router/src/color/materialColor.android.ts
```diff
@@ -53,22 +53,29 @@ type SchemeName = 'light' | 'dark' | 'unspecified';
...
-  if (!AndroidExpoRouterModule) {
-    AndroidExpoRouterModule = requireNativeModule<AndroidExpoRouterModuleType>('ExpoRouter');
-  }
-  return AndroidExpoRouterModule.Material3Color(name, scheme);
+  return getNativeModule().Material3Color(name, scheme);
+}
+
+export function addColorPaletteListener(listener: () => void): () => void {
+  const subscription = getNativeModule().addListener('onColorPaletteChanged', listener);
+  return () => subscription.remove();
```
Kudo, line 78: it feels we adding a listener (which with some native call overhead) in each useRouterColor hook.

would be good if expo-router has some root layout component that we can register the listener and useContext there, so we can share the listener

Kudo (changes requested):
leaving some comments if anything of these makes sense to you

PR #47856, opened 2026-07-16
[docs] Document non-interactive `eas init` for access token flows

amandeepmittal:
# Why

Fix ENG-25190

# How

Document non-interactive `eas init` for access token flows under Access tokens usage section.

# Test Plan

CleanShot 2026-07-16 at 19 29 00@2x

# Checklist

### DIFF
--- a/docs/pages/accounts/programmatic-access.mdx
+++ b/docs/pages/accounts/programmatic-access.mdx
@@ -27,7 +27,7 @@ For example, once you obtain a token, you can run the following EAS CLI command 
 
 <Terminal cmd={['$ EXPO_TOKEN=my_token eas build']} />
 
-> **info** Commands that run with an access token require the project to already be linked to an EAS project. If `extra.eas.projectId` is not set in your [app config](/workflow/configuration), commands such as `eas build` fail with an `EAS project not configured` error. To configure it, run `EXPO_TOKEN=my_token eas init --force --non-interactive` (or `EXPO_TOKEN=my_token eas init --id <your-project-id>`) first.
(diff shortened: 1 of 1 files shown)


entiendoNull (approved):
LGTM! 
This addresses point 2 of the Linear ticket. Point 1 looks inaccurately reported (no matching build:status reference in the docs as far as I can see), so scoping this PR to point 2 makes sense.

docs/pages/accounts/programmatic-access.mdx
```diff
@@ -27,6 +27,8 @@ For example, once you obtain a token, you can run the following EAS CLI command
 
 <Terminal cmd={['$ EXPO_TOKEN=my_token eas build']} />
 
+> **info** Commands that run with an access token require the project to already be linked to an EAS project. If `extra.eas.projectId` is not set in your [app config](/workflow/configuration), commands such as `eas build` fail with an `EAS project not configured` error. To configure it, run `EXPO_TOKEN=my_token eas init --force --non-interactive` (or `EXPO_TOKEN=my_token eas init --id <your-project-id>`) first.
```
kadikraman: I don't think the second example might confuse things further since that's only for the case where the project is created manually, just the local code is unlinked. I doubt agents would even run into that one.

```suggestion
> **info** Commands that run with an access token require the project to already be linked to an EAS project. If `extra.eas.projectId` is not set in your [app config](/workflow/configuration), commands such as `eas build` fail with an `EAS project not configured` error. To configure it, run `EXPO_TOKEN=my_token eas init --force --non-interactive` first.
```

kadikraman (approved):

PR #47857, opened 2026-07-16
[expo-modules-jsi][fix] Change abs to magnitude, partially fix bare-expo build

HubertBer:
# Why
`bare-expo` doesn't build, because `abs` call is ambiguous.

# How
Changed `abs` to `magnitude`.

# Test Plan
- [x] compiled bare-expo (though it needed different unrelated fixes) 

# Checklist

### DIFF
--- a/packages/expo-modules-jsi/apple/Sources/ExpoModulesJSI/Coding/JavaScriptCodable+Date.swift
+++ b/packages/expo-modules-jsi/apple/Sources/ExpoModulesJSI/Coding/JavaScriptCodable+Date.swift
@@ -50,7 +50,7 @@ let maxJavaScriptDateMilliseconds: Double = 8_640_000_000_000_000
 /// faithful to `new Date(number)`; the `Date`/string branches pass an already-clipped `getTime()` through.
 @usableFromInline
 func dateFromMilliseconds(_ milliseconds: Double) throws -> Date {
-  guard milliseconds.isFinite, abs(milliseconds) <= maxJavaScriptDateMilliseconds else {
+  guard milliseconds.isFinite, milliseconds.magnitude <= maxJavaScriptDateMilliseconds else {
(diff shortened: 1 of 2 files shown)


packages/expo-modules-jsi/apple/Sources/ExpoModulesJSI/Coding/JavaScriptCodable+Date.swift
```diff
@@ -50,7 +50,7 @@ let maxJavaScriptDateMilliseconds: Double = 8_640_000_000_000_000
 /// faithful to `new Date(number)`; the `Date`/string branches pass an already-clipped `getTime()` through.
 @usableFromInline
 func dateFromMilliseconds(_ milliseconds: Double) throws -> Date {
-  guard milliseconds.isFinite, abs(milliseconds) <= maxJavaScriptDateMilliseconds else {
+  guard milliseconds.isFinite, milliseconds.magnitude <= maxJavaScriptDateMilliseconds else {
```
tsapeta, line 53: It would be good to understand why your Xcode considers it as ambigious. It works for me on both Xcode 26 and 27, same on the CI.

tsapeta (approved):

packages/expo-modules-jsi/apple/Sources/ExpoModulesJSI/Coding/JavaScriptCodable+Date.swift
alanjhughes: Working for me on both too

HubertBer: So I've actually had an old XCode version. It works after the update, sorry for the confusion!

PR #47862, opened 2026-07-16
[JSI] Resolve a `JavaScriptPromise` with a `JavaScriptEncodable` value

tsapeta:
# Why

Settling a `JavaScriptPromise` currently requires a `JavaScriptRepresentable` value. To resolve with a native type that only conforms to `JavaScriptEncodable` (a `Record`, a `SharedObject`, or a `Task`, see the stacked PR), the caller had to hop to the JavaScript thread and encode by hand first.

# How

Add a `resolve` overload taking a `JavaScriptEncodable` value. Like the existing overload it is non-isolated and hops to the JavaScript thread via `runtime.schedule`; the `encode` (which is `@JavaScriptActor`-isolated and may touch the runtime) runs inside that scheduled job, and a thrown encode rejects the promise instead.

The existing `JavaScriptRepresentable` overload is marked `@_disfavoredOverload`, so a value conforming to *both* protocols (a primitive, a container, `JavaScriptValue`) resolves through the encodable path, while the JSI value wrappers that are representable-only (`JavaScriptObject`, `JavaScriptFunction`, `JavaScriptBigInt`) keep using it. No call-site label is needed.

The two protocols stay complementary, and `JavaScriptRepresentable` is not being phased out. It keeps its JS to native direction (`fromJavaScriptValue`), its non-isolated/non-throwing callers (e.g. `promise.resolve` from a `@Sendable` async-function callback), and the `JSIRepresentable` fast path used in variadic argument packing.

# Test Plan

Added a test resolving with an encodable container (`[String: Int]`), which routes through the new overload via the disfavored ranking. Full `JavaScriptPromiseTests` (36) pass, and the full `expo-modules-core` iOS suite passes, confirming the disfavored ranking leaves every existing `resolve(...)` caller (primitives, `Record`, `SharedObject` subclasses, `JavaScriptValue`) unambiguous.

### DIFF
--- a/packages/expo-modules-jsi/apple/Sources/ExpoModulesJSI/Runtime/Values/JavaScriptPromise.swift
+++ b/packages/expo-modules-jsi/apple/Sources/ExpoModulesJSI/Runtime/Values/JavaScriptPromise.swift
@@ -121,6 +121,12 @@ public struct JavaScriptPromise: JavaScriptType, ~Copyable {
     } ?? .undefined
   }
 
+  /// Resolves the promise with a value that has a direct JavaScript representation.
+  ///
+  /// Preferred over the encodable overload for a type that is both ``JavaScriptRepresentable`` and
+  /// ``JavaScriptEncodable``, so existing values (primitives, containers, JSI value wrappers) keep
+  /// their established representation. For example a 64-bit integer stays a JS `number` here rather
+  /// than encoding to a `bigint` or rejecting for exceeding the safe-integer range.
   public func resolve<V: JavaScriptRepresentable>(_ value: V) {
     guard let runtime else {
       return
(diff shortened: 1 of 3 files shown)


packages/expo-modules-jsi/apple/Sources/ExpoModulesJSI/Runtime/Values/JavaScriptPromise.swift
```diff
@@ -142,6 +148,39 @@ public struct JavaScriptPromise: JavaScriptType, ~Copyable {
...
+    // `resolve` is not isolated, so make sure to jump to JS thread; the encode happens there too.
+    runtime.schedule(priority: .immediate) { [longLivedState] in
+      // If the promise is already settled, do nothing.
+      guard let resolver = longLivedState.resolveFunction.take() else {
+        return
+      }
+      do {
+        let encoded = try V.encode(value, in: runtime)
+        // Call the actual resolver, which also calls `deferredPromise.resolve` in the `then` handler.
+        _ = try! resolver.getFunction().call(arguments: encoded)
```
alanjhughes:
```suggestion
        _ = try resolver.getFunction().call(arguments: encoded)
```
Is the `!` needed?

packages/expo-modules-jsi/apple/Sources/ExpoModulesJSI/Runtime/Values/JavaScriptPromise.swift
```diff
@@ -142,6 +148,39 @@ public struct JavaScriptPromise: JavaScriptType, ~Copyable {
...
+        // The state stays registered so it keeps owning the object until the wrapper is dropped.
+        longLivedState.rejectFunction.release()
+      } catch {
+        // Encoding failed; reject with the error and release the resolver instead.
+        let errorValue = JavaScriptError.from(error, in: runtime).toValue()
+        _ = try! longLivedState.rejectFunction.take()?.getFunction().call(arguments: errorValue)
```
alanjhughes: This is a fatal if it throws?

packages/expo-modules-jsi/apple/Sources/ExpoModulesJSI/Runtime/Values/JavaScriptPromise.swift
tsapeta: Now no, it seems that Claude took it as is in the other overloads. Changing to `try`, but it's safe to assume this doesn't throw – it's a `resolve` function passed to `Promise` constructor which doesn't throw.

packages/expo-modules-jsi/apple/Sources/ExpoModulesJSI/Runtime/Values/JavaScriptPromise.swift
tsapeta: Yes, but we can't do much if `reject` passed to `Promise` constructor throws 😅

tsapeta:
@alanjhughes, I addressed both comments, so it's ready for re-review.

- Dropped the `try!` on the resolver call as suggested; a throwing resolver call now rejects the promise instead of trapping.
- Your second comment convinced me to remove the remaining `try!`s from the settle paths too: they can't realistically throw (replaced `globalThis.Promise` implementation?), so dropping the settle is harmless https://github.com/expo/expo/pull/47862/changes/a6fe6e450b074c24527d712a7538a48eeac8a236

alanjhughes (approved):

PR #47869, opened 2026-07-16
[expo-router] Fix TabRouter crash on partial stale:false state (missing history/preloadedRouteKeys)

### LINKED ISSUE #47868: [expo-router] Vendored TabRouter.getStateForAction crashes on a stale:false state missing history/preloadedRouteKeys ("Cannot read properties of undefined (reading 'filter')")
## Summary

The vendored react-navigation `TabRouter` in expo-router (`packages/expo-router/src/react-navigation/routers/TabRouter.tsx`) crashes in `getStateForAction` when it receives a `stale: false` navigation state that does **not** carry `history` and `preloadedRouteKeys`:

```
TypeError: Cannot read properties of undefined (reading 'filter')
    at Object.getStateForAction (TabRouter)
    at useTabsWithTriggers
    at Tabs
```

Two compounding issues let a partial state reach the unguarded reads:

1. **`getRehydratedState` adopts foreign states verbatim.** Its fast path returns any `stale: false` state as-is, without checking `state.type`. A sibling navigator's state (e.g. `type: 'stack'`, which has neither `history` nor `preloadedRouteKeys`) is accepted wholesale as if it were a complete tab state.
2. **`getStateForAction` reads `state.history` / `state.preloadedRouteKeys` unguarded**, so a state that arrives (or was rehydrated) without those fields throws.

The same file is already internally inconsistent about whether these fields can be absent: `getStateForRouteNamesChange` guards them (`state.history?.filter(...) ?? []`, `state.preloadedRouteKeys?.filter(...) ?? []`), and `ExpoTabRouter` in `src/ui/TabRouter.ts` optional-chains `state.history?.some(...)`. Only `getStateForAction` is missing the guard.

Reproduced on expo-router **56.2.14**, and the unguarded code is unchanged on current **main**. The crash path is platform-agnostic; we hit it on web (static export).

https://github.com/kimchouard/expo-router-tabrouter-partial-state-repro

The router unit itself is enough — no app, renderer, or platform is needed. The repo dispatches the exact failing action shapes against the real published router (`expo-router/build/react-navigation/routers/TabRouter`) and asserts the crash, so the tests **pass** by expecting the throw.

## Steps to reproduce

Package manager: npm. Platform: platform-independent (the reproduction is at the router-unit level; the real-world hit is on **web**, static export). Environment: Node — no Expo Go / dev build / standalone app required.

```bash
git clone https://github.com/kimchouard/expo-router-tabrouter-partial-state-repro
cd expo-router-tabrouter-partial-state-repro
npm install
npm test
```

**Expected:** dispatching a `JUMP_TO` / `NAVIGATE` / `GO_BACK` action against a `stale: false` tab state should either navigate or be handled gracefully.

**Actual:** `getStateForAction` throws. `JUMP_TO` and `NAVIGATE` throw `TypeError: Cannot read properties of undefined (reading 'filter')` (on `state.preloadedRouteKeys.filter`); `GO_BACK` throws `TypeError: Cannot read properties of undefined (reading 'length')` (on `state.history.length`). The four tests pass because they assert exactly these throws, plus one test showing a complete state navigates normally.

Test output:

```
 ✓ tabRouterPartialState.test.ts (4 tests) 2ms

 Test Files  1 passed (1)
      Tests  4 passed (4)
```

## Environment

```
    System:
      OS: macOS 26.5.2
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 26.0.0 - /opt/homebrew/bin/node
      npm: 11.12.1 - /opt/homebrew/bin/npm
      Watchman: 2026.05.25.00 - /opt/homebrew/bin/watchman
    Managers:
      CocoaPods: 1.14.2
    SDKs:
      iOS SDK:
    IDEs:
      Xcode: 26.6/17F113
    npmPackages:
      expo-router: 56.2.14 => 56.2.14
    npmGlobalPackages:
      eas-cli: 16.28.0
    Expo Workflow: managed
```

## Expo Doctor Diagnostics

Not applicable to the minimal reproduction: the crash is in a pure router unit, so the repro is a Node/Vitest harness with no Expo app installed (`expo-doctor` reports `Cannot determine the project's Expo SDK version`). Deliberately keeping the repro at router-unit level makes it fully deterministic and platform-independent — no app, renderer, Metro, or simulator involved. Happy to provide a full app-level repro + `expo-doctor` output if a maintainer prefers, but the unit repro isolates the bug precisely.

## What happens

`getStateForAction` reaches a branch that reads `.filter` / `.length` on `state.history` or `state.preloadedRouteKeys`, both of which are `undefined` on the incoming partial state, and throws during render (inside `useTabsWithTriggers` → `Tabs`).

## Why

Line references and code below are from `main` at commit `9758cc093814b0c8c58b547906dc2b1640bb27a1` (`packages/expo-router/src/react-navigation/routers/TabRouter.tsx`).

**1. The rehydration fast path returns partial/foreign states verbatim — no `type` check (lines 225-227):**

```ts
getRehydratedState(partialState, { routeNames, routeParamList }) {
  const state = partialState;

  if (state.stale === false) {
    return state;                    // ← returned as-is; never checks state.type === 'tab'
  }
```

A `stale: false` state belonging to another navigator (e.g. `type: 'stack'`) is adopted unchanged, so it flows into `getStateForAction` without `history` / `preloadedRouteKeys`.

**2. `getStateForAction` then reads those fields unguarded.**

JUMP_TO / NAVIGATE (line 379):

```ts
return {
  ...updatedState,
  preloadedRouteKeys: updatedState.preloadedRouteKeys.filter(   // ← undefined.filter → throw
    (key) => key !== state.routes[updatedState.index]!.key
  ),
};
```

(`updatedState` is `changeIndex({ ...state, ... })`, which spreads the incoming `state`, so a missing `preloadedRouteKeys` stays missing.)

GO_BACK (lines 419, 423, 447, 450):

```ts
case 'GO_BACK': {
  if (state.history.length === 1) {                            // ← 419: undefined.length → throw
    return null;
  }

  const previousHistoryItem = state.history[state.history.length - 2];  // ← 423
  // ...
  return {
    ...state,
    routes,
    preloadedRouteKeys: state.preloadedRouteKeys.filter(       // ← 447
      (key) => key !== state.routes[index]!.key
    ),
    history: state.history.slice(0, -1),                       // ← 450
    index,
  };
}
```

**Inconsistency in the same file** — `getStateForRouteNamesChange` already treats these fields as possibly-absent (lines 257, 269):

```ts
const history = state.history?.filter((it) => routeKeys.includes(it.key)) ?? [];
// ...
preloadedRouteKeys:
  state.preloadedRouteKeys?.filter((key) => routeKeys.includes(key)) ?? [],
```

and `ExpoTabRouter` (`src/ui/TabRouter.ts`) optional-chains `state.history?.some(...)`. So the codebase half-acknowledges the fields can be undefined; `getStateForAction` is the gap.

## Real-world trigger (production)

On web (static export): a user is on a screen inside a sibling stack and triggers `router.dismissTo('/songs/<id>')`, where the target route belongs to a **headless `expo-router/ui` `<Tabs>`** navigator (`useTabsWithTriggers`). expo-router's `getNavigationAction` converts the `POP_TO` into a `JUMP_TO` targeted at the tabs navigator. That tabs navigator's current state is a `stale: false` partial (no `history` / `preloadedRouteKeys`), so the `JUMP_TO` branch above throws during render. Observed frames: `Object.getStateForAction (TabRouter)` ← `useTabsWithTriggers` ← `Tabs`.

## Suggested fix

Minimal, behavior-preserving (complete states are unaffected): normalize at the top of `getStateForAction` so the downstream `.filter` / `.length` reads always see arrays:

```ts
getStateForAction(state, action, options) {
  if (state.history === undefined || state.preloadedRouteKeys === undefined) {
    state = {
      ...state,
      history: state.history ?? [],
      preloadedRouteKeys: state.preloadedRouteKeys ?? [],
    };
  }
  // ...existing switch
}
```

This mirrors the guards `getStateForRouteNamesChange` already uses and adds zero behavior change for complete states. A unit test dispatching the three failing shapes against the real shipped router passes with it and fails without.

**Complementary (behavior-affecting) option, maintainer's call:** also make `getRehydratedState`'s fast path require `state.type === this.type` before returning verbatim, so a foreign navigator's state is re-normalized instead of adopted. This matches the documented `stale: false` completeness contract (https://reactnavigation.org/docs/navigation-state/), but it changes which states take the fast path, so it deserves its own review; the `getStateForAction` guard above is the safe, targeted fix on its own.

I'm happy to open a PR for the minimal `getStateForAction` guard.

## Related issues

- expo/expo#37543 — "Expo Router breaks on 404 routes" — same partial-state-on-web family; fixed in expo/expo#37747 but only at `src/fork/useLinking.ts`, so `TabRouter` remains uncovered.
- expo/expo#29865 — closed as needs-repro/outdated — same error string via `router.replace`; likely the same underlying gap, undiagnosed.
- react-navigation/react-navigation#13011 — related — `getRehydratedState(undefined)` crash from the same fast-path fragility upstream. Noting it for context only; expo-router fully vendors these routers, so this report and fix are scoped to expo's fork.

kimchouard:
# Why

Fixes #47868.

The vendored `TabRouter.getStateForAction` (`packages/expo-router/src/react-navigation/routers/TabRouter.tsx`) crashes when it receives a `stale: false` navigation state that lacks `history` and `preloadedRouteKeys`:

```
TypeError: Cannot read properties of undefined (reading 'filter')
    at Object.getStateForAction (TabRouter)
    at useTabsWithTriggers
    at Tabs
```

`getRehydratedState` returns any `stale: false` state verbatim (its fast path never checks `state.type`), so a state that skipped rehydration — e.g. a sibling navigator's state adopted as-is — reaches the action handlers without those two collections. The `JUMP_TO` / `NAVIGATE` branch then does `updatedState.preloadedRouteKeys.filter(...)` and `GO_BACK` does `state.history.length` / `state.history.slice(...)` / `state.preloadedRouteKeys.filter(...)`, all on `undefined`.

Real-world trigger (web static export): a user on a screen inside a sibling stack calls `router.dismissTo('/…')` where the target belongs to a headless `expo-router/ui` `<Tabs>` navigator. `getNavigationAction` converts the `POP_TO` into a `JUMP_TO` targeted at the tabs navigator, whose current state is a `stale: false` partial — so the read throws during render.

# How

Normalize the two optional collections at the top of `getStateForAction`, mirroring the way `getRehydratedState` already tolerates them (`state.history?.filter(...) ?? []`, `state.preloadedRouteKeys?.filter(...) ?? []`):

```ts
if (state.history === undefined || state.preloadedRouteKeys === undefined) {
  state = {
    ...state,
    history: state.history ?? [],
    preloadedRouteKeys: state.preloadedRouteKeys ?? [],
  };
}
```

This is minimal and behavior-preserving: complete states never enter the branch, and every downstream branch stays safe (`JUMP_TO`/`NAVIGATE` rebuild history via `changeIndex`; `GO_BACK` on an empty history returns `null`, handled upstream).

The issue also notes a complementary, behavior-affecting option — making `getRehydratedState`'s fast path require `state.type === this.type` before adopting a state verbatim. That changes which states take the fast path, so it deserves its own review and is deliberately **not** included here; the `getStateForAction` guard is the safe, targeted fix on its own.

# Test Plan

Added regression tests in `packages/expo-router/src/react-navigation/routers/__tests__/TabRouter.test.tsx` that dispatch `JUMP_TO`, `NAVIGATE`, and `GO_BACK` against a `stale: false` state with `history`/`preloadedRouteKeys` absent, asserting no throw and the expected navigated/`null` result, plus a complete-state case proving no behavior change. These run under the package's existing `jest` suite in CI.

The expected values in those tests were validated against the real published (patched) router: `JUMP_TO`/`NAVIGATE` → `index: 1`, `history: [baz, bar]`, `preloadedRouteKeys: []`; `GO_BACK` → `null`.

Additional executable evidence — a standalone published-package reproduction that dispatches the exact failing action shapes against the shipped router and asserts the pre-fix crash: https://github.com/kimchouard/expo-router-tabrouter-partial-state-repro

# Checklist

- [x] I added a `changelog.md` entry and rebuilt the package sources according to [this short guide](https://github.com/expo/expo/blob/main/CONTRIBUTING.md#-before-submitting) (`build/` is generated and gitignored; nothing to commit).
- [x] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: updated a module plugin). (Pure JS router logic; no native or plugin surface touched.)
- [x] Conforms with the [Documentation Writing Style Guide](https://github.com/expo/expo/blob/main/guides/Expo%20Documentation%20Writing%20Style%20Guide.md) (no docs changes).

### DIFF
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -30,6 +30,7 @@
 - [android] Remove navigation state restoration across activity recreation. ([#47422](https://github.com/expo/expo/pull/47422) by [@Ubax](https://github.com/Ubax))
 - Fix `renderRouter` ignoring `overrides` and listing duplicate routes when an override key matches a file in `appDir`. ([#47287](https://github.com/expo/expo/pull/47287) by [@wwdrew](https://github.com/wwdrew))
 - Guard the deep link decode in `extractExactPathFromURL` against malformed percent-encoding. ([#47526](https://github.com/expo/expo/pull/47526) by [@momomuchu](https://github.com/momomuchu))
+- Fix `TabRouter.getStateForAction` crashing on a `stale: false` state that is missing `history`/`preloadedRouteKeys`. ([#47869](https://github.com/expo/expo/pull/47869) by [@kimchouard](https://github.com/kimchouard))
 
 ### 💡 Others
 
--- a/packages/expo-router/src/react-navigation/routers/TabRouter.tsx
+++ b/packages/expo-router/src/react-navigation/routers/TabRouter.tsx
@@ -315,6 +315,19 @@ function BaseTabRouter({ initialRouteName, backBehavior = 'firstRoute' }: TabRou
     },
 
     getStateForAction(state, action, { routeParamList, routeGetIdList }) {
+      // A `stale: false` state is returned verbatim by `getRehydratedState`, so a
+      // state that skipped rehydration (e.g. a foreign navigator's state adopted
+      // as-is) can reach here without `history`/`preloadedRouteKeys`. Normalize
+      // them the same way `getRehydratedState` already does so the `.filter`/
+      // `.length` reads below don't throw. Fixes #47868.
+      if (state.history === undefined || state.preloadedRouteKeys === undefined) {
+        state = {
+          ...state,
+          history: state.history ?? [],
+          preloadedRouteKeys: state.preloadedRouteKeys ?? [],
+        };
+      }
+
       switch (action.type) {
         case 'JUMP_TO':
         case 'NAVIGATE':
--- a/packages/expo-router/src/react-navigation/routers/__tests__/TabRouter.test.tsx
+++ b/packages/expo-router/src/react-navigation/routers/__tests__/TabRouter.test.tsx
@@ -2592,3 +2592,138 @@ test('handles screen preloading', () => {
     history: [{ type: 'route', key: 'qux-test' }],
   });
 });
+
+// Regression tests for #47868: a `stale: false` state can reach
+// getStateForAction without `history`/`preloadedRouteKeys` (e.g. a foreign
+// navigator's state returned verbatim by getRehydratedState). The unguarded
+// `.filter`/`.length` reads used to throw
+// "Cannot read properties of undefined (reading 'filter'/'length')".
+
+const partialState = () =>
+  ({
+    stale: false,
+    type: 'tab',
+    key: 'root',
+    index: 0,
+    routeNames: ['baz', 'bar'],
+    routes: [
+      { key: 'baz', name: 'baz' },
+      { key: 'bar', name: 'bar' },
+    ],
+    // `history` and `preloadedRouteKeys` intentionally absent
+  }) as unknown as TabNavigationState<ParamListBase>;
+
+test('handles jump to on a state without history/preloadedRouteKeys (#47868)', () => {
+  const router = TabRouter({});
+  const options: RouterConfigOptions = {
+    routeNames: ['baz', 'bar'],
+    routeParamList: {},
+    routeGetIdList: {},
+  };
+
+  expect(() =>
+    router.getStateForAction(partialState(), TabActions.jumpTo('bar'), options)
+  ).not.toThrow();
+
+  expect(router.getStateForAction(partialState(), TabActions.jumpTo('bar'), options)).toEqual({
+    stale: false,
+    type: 'tab',
+    preloadedRouteKeys: [],
+    key: 'root',
+    index: 1,
+    routeNames: ['baz', 'bar'],
+    routes: [
+      { key: 'baz', name: 'baz' },
+      { key: 'bar', name: 'bar' },
+    ],
+    history: [
+      { type: 'route', key: 'baz' },
+      { type: 'route', key: 'bar' },
+    ],
+  });
+});
+
+test('handles navigate on a state without history/preloadedRouteKeys (#47868)', () => {
+  const router = TabRouter({});
+  const options: RouterConfigOptions = {
+    routeNames: ['baz', 'bar'],
+    routeParamList: {},
+    routeGetIdList: {},
+  };
+
+  expect(() =>
+    router.getStateForAction(partialState(), CommonActions.navigate('bar'), options)
+  ).not.toThrow();
+
+  expect(router.getStateForAction(partialState(), CommonActions.navigate('bar'), options)).toEqual({
+    stale: false,
+    type: 'tab',
+    preloadedRouteKeys: [],
+    key: 'root',
+    index: 1,
+    routeNames: ['baz', 'bar'],
+    routes: [
+      { key: 'baz', name: 'baz' },
+      { key: 'bar', name: 'bar' },
+    ],
+    history: [
+      { type: 'route', key: 'baz' },
+      { type: 'route', key: 'bar' },
+    ],
+  });
+});
+
+test('handles back on a state without history/preloadedRouteKeys (#47868)', () => {
+  const router = TabRouter({});
+  const options: RouterConfigOptions = {
+    routeNames: ['baz', 'bar'],
+    routeParamList: {},
+    routeGetIdList: {},
+  };
+
+  expect(() =>
+    router.getStateForAction(partialState(), CommonActions.goBack(), options)
+  ).not.toThrow();
+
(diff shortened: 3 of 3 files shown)


Ubax:
Please see the question in the issue

brentvatne:
Closing — TabRouter crash on partial `stale:false` state was fixed on main by https://github.com/expo/expo/pull/48709 (`ensureStateHistory`) and https://github.com/expo/expo/pull/48718. Thanks!

PR #47870, opened 2026-07-16
[router] Fix `expo-router/head` and `expo-router/stack` resolution

hassankhan:
# Why

As part of the changes in #46801, an exports map was added to `expo-router/package.json` which included `./head` and `./stack`. When Metro resolves through an exports map, it doesn't apply platform extension resolution.

This caused the web implementation (the default, non-platform-suffixed `ExpoHead.tsx`) to be bundled for Android/iOS. Since there is no `<HelmetProvider>` on native, any route rendering `<Head>` crashed at startup with `TypeError: Cannot read property 'add' of undefined`.

The `./stack` export had the same issue in reverse; web was bundling the native `<Stack>` instead of `Stack.web.tsx`.

# How

Both exports in the exports map now point to a barrel file whose extension-less relative import Metro _can_ platform-resolve. This is the same way the legacy root `head.js`/`stack.js` shims worked before the exports map was added. Additionally, the `./head` barrel explicitly re-exports `Head` as the default export, mirroring the old root export.

As a follow-up, we should consider whether the legacy root shims (`head.js`, `stack.js`, etc.) can be removed now that the exports map covers these subpaths.

# Test Plan

- `cd apps/router-e2e && pnpm expo prebuild && E2E_ROUTER_SRC=static-rendering pnpm expo run:ios` should run the app without `TypeError: Cannot read property 'add' of undefined`.
- `E2E_ROUTER_SRC=static-rendering pnpm expo export -p ios` should produce an iOS bundle that includes `ExpoHead.ios.js` instead of the web helmet implementation.

# Checklist

### DIFF
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -30,6 +30,7 @@
 - [android] Remove navigation state restoration across activity recreation. ([#47422](https://github.com/expo/expo/pull/47422) by [@Ubax](https://github.com/Ubax))
 - Fix `renderRouter` ignoring `overrides` and listing duplicate routes when an override key matches a file in `appDir`. ([#47287](https://github.com/expo/expo/pull/47287) by [@wwdrew](https://github.com/wwdrew))
 - Guard the deep link decode in `extractExactPathFromURL` against malformed percent-encoding. ([#47526](https://github.com/expo/expo/pull/47526) by [@momomuchu](https://github.com/momomuchu))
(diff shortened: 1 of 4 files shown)


Ubax:
~~When I test it on Android I can see this error~~

~~Screenshot_1784274203~~

~~I'm not sure if this is related~~

Sorry, I forgot to run `pnpm build` in `expo-router` 😅

Ubax (approved):

kitten (approved):

PR #47871, opened 2026-07-16
[observe] Add reportError for caught non-fatal errors

tsapeta:
# Why

Error capture in `expo-app-metrics` is fully automatic today, through two internal paths that both feed the private native `reportError`:

- `source: 'global'` — the React Native `ErrorUtils` global handler (fatal or non-fatal)
- `source: 'errorBoundary'` — the `AppMetricsErrorBoundary` (always non-fatal)

There is no way for a developer to report an error they caught and handled in their own `try`/`catch`. This adds a public API to record those handled, non-fatal errors so they flow to `/v1/logs` alongside the automatically-captured ones.

# How

Adds a public `Observe.reportError(error: unknown): void`.

It normalizes the caught value the same way the global `ErrorUtils` handler does: an `Error`'s `name`, `message`, and `stack` are captured (each only when it's actually a string); any other thrown value (a string, a plain object) is stringified as the message via `String(error)` and carries no stack. The result is recorded as a non-fatal `exception` log event tagged `expo.error.source = 'reportedByUser'`, reusing the existing non-fatal log pipeline end to end (no new dispatch wiring).

- **expo-observe**: an explicit `reportError` case in the `module.ts` `Proxy` (so the public shape stays decoupled from the native payload) and a matching `module.web.ts` implementation, both delegating to a shared `reportCaughtError` helper; plus a typed public declaration on `ObserveModule` with TSDoc + `@example`.
- **expo-app-metrics**: a new `reportedByUser` case added to the `source` closed set across TS, Swift (`ErrorReport.Source`), and Kotlin (`ErrorSource`). `reportError` itself already handles the non-fatal async path on both platforms, so no other native change is needed.

**Robustness** (`reportError` is called from a `catch` block, so it must not make a handled error worse):
- Only a real `Error` contributes `name`/`message`/`stack`, and each only when it's a string — so a non-`Error` object with Error-like non-string fields (e.g. `throw { message: 404 }`) can't reach the native string fields and fail the record decode.
- The whole report flow never throws: a pathological thrown value (a throwing getter or `toString`) or a native call that rejects the payload is swallowed, with a `console.warn` in `__DEV__` only.

A follow-up will add an optional `attributes` argument; that requires the `@Record` macro to accept `[String: Any]` fields, tracked in ENG-25264.

# Test Plan

- `et check-packages expo-observe expo-app-metrics` → all checks pass (build, typecheck, lint, test), no warnings.
- JS: `expo-observe` module tests cover a caught `Error` (name/message/stack, non-fatal, `source: 'reportedByUser'`), a non-`Error` throw (stringified message, no stack), a plain object with non-string Error-like fields, and the never-throws cases (throwing getter, throwing `toString`, native call throwing). 62 passed.
- Native: `ErrorReportTests` (iOS) and `ErrorReportTest` (Android) assert a `reportedByUser`-source error is tagged `expo.error.source = 'reportedByUser'` at `error` severity. Both suites pass.
- Verified end-to-end on an Android emulator via the observe-tester app (`Report caught errors` section on the Debug tab): the recorded session logs show the two `exception` events with `expo.error.source = 'reportedByUser'`.

### DIFF
--- a/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/jserrors/ErrorReport.kt
+++ b/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/jserrors/ErrorReport.kt
@@ -55,7 +55,8 @@ data class ErrorReport(
 /** How the error was captured. A closed set so the `expo.error.source` attribute stays consistent. */
 enum class ErrorSource(val rawValue: String) : Enumerable {
   GLOBAL("global"),
-  ERROR_BOUNDARY("errorBoundary")
+  ERROR_BOUNDARY("errorBoundary"),
+  REPORTED_BY_USER("reportedByUser")
 }
 
 /**
--- a/packages/expo-observe/src/reportCaughtError.ts
+++ b/packages/expo-observe/src/reportCaughtError.ts
@@ -1,0 +1,57 @@
+import AppMetrics from 'expo-app-metrics';
+
+/**
+ * The `reportError` payload shape sent to the native AppMetrics module. The native record types
+ * `type`/`stacktrace` as optional strings and `message` as a required string, so a non-string value
+ * would fail the record decode and drop the report. Every field is normalized to a string here.
+ */
+type NormalizedReportedError = {
+  type?: string;
+  message: string;
+  stacktrace?: string;
+};
+
+/** Returns `value` when it's a string, otherwise `undefined`, so non-string fields never reach native. */
+function stringOrUndefined(value: unknown): string | undefined {
+  return typeof value === 'string' ? value : undefined;
+}
+
+/**
+ * Normalizes an arbitrary caught value into the fields the native `reportError` expects, the way the
+ * global `ErrorUtils` handler does.
+ */
+function normalizeReportedError(error: unknown): NormalizedReportedError {
+  // Only a real `Error` contributes name/message/stack, and each only when it's actually a string;
+  // an `Error` with non-string fields would otherwise reach the native string fields (e.g. a mutated
+  // `message`) and fail the record decode.
+  if (error instanceof Error) {
+    return {
+      type: stringOrUndefined(error.name),
+      message: stringOrUndefined(error.message) ?? String(error),
+      stacktrace: stringOrUndefined(error.stack),
+    };
+  }
+  // Any non-Error throw (a string, a plain object, a number) becomes its stringification, no stack.
+  return { message: String(error) };
+}
+
+/**
+ * Reports a caught value as a non-fatal `reportedByUser`-source error through the AppMetrics module,
+ * shared by the native and web `Observe.reportError` implementations.
+ *
+ * Never throws: it's called from a `catch` block, so a failure here (a pathological thrown value, or
+ * a native call that rejects the payload) must not turn a handled error into an unhandled one.
+ */
+export function reportCaughtError(error: unknown): void {
+  try {
+    AppMetrics.reportError({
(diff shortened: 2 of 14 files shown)


packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/jserrors/ErrorReport.kt
```diff
@@ -55,7 +55,8 @@ data class ErrorReport(
 /** How the error was captured. A closed set so the `expo.error.source` attribute stays consistent. */
 enum class ErrorSource(val rawValue: String) : Enumerable {
   GLOBAL("global"),
-  ERROR_BOUNDARY("errorBoundary")
+  ERROR_BOUNDARY("errorBoundary"),
+  CAUGHT("caught")
```
Ubax: WDYT about naming it more like `reportedByUser` or something similar?

In theory this can be anything that user reports, not only error caught in try/catch.

packages/expo-observe/src/reportCaughtError.ts
```diff
@@ -0,0 +1,58 @@
...
+ * would fail the record decode and drop the report. Every field is normalized to a string here.
+ */
+type NormalizedReportedError = {
+  type?: string;
+  message: string;
+  stacktrace?: string;
+};
+
+/** Returns `value` when it's a string, otherwise `undefined`, so non-string fields never reach native. */
+function asString(value: unknown): string | undefined {
```
Ubax: Can we call it `assertString` or `stringOrUndefined`?

`asString` suggests to me that the value will be cast to string, e.g. `"abc" -> "abc"; { 1: 2 } -> "[Object object]`

packages/expo-observe/src/reportCaughtError.ts
```diff
@@ -0,0 +1,58 @@
...
+}
+
+/**
+ * Reports a caught value as a non-fatal `caught`-source error through the AppMetrics module, shared
+ * by the native and web `Observe.reportError` implementations.
+ *
+ * Never throws. `reportError` is called from a `catch` block, so a failure here (a pathological
+ * thrown value with a throwing getter or `toString`, or a native call that rejects the payload) must
+ * not escape and turn a handled error into an unhandled one. Any failure is swallowed, with a
+ * `console.warn` in development so the problem is still visible while debugging.
```
Ubax:
```suggestion
 * Never throws.
```

This comment is basically explaining the next 6 lines of code. I don't think this is needed

packages/expo-observe/src/reportCaughtError.ts
```diff
@@ -0,0 +1,58 @@
...
+  return typeof value === 'string' ? value : undefined;
+}
+
+/**
+ * Normalizes an arbitrary caught value into the fields the native `reportError` expects, the way the
+ * global `ErrorUtils` handler does. Only a real `Error` contributes `name`/`message`/`stack`, and
+ * each is used only when it's actually a string; anything else (a string, a plain object, a number,
+ * an `Error` with non-string fields) falls back to `String(error)` for the message with no type or
+ * stacktrace. This keeps a non-string field (e.g. `throw { message: 404 }`) from reaching native.
+ */
```
Ubax, line 22:
```suggestion
/**
 * Normalizes an arbitrary caught value into the fields the native `reportError` expects, the way the
 * global `ErrorUtils` handler does. 
 */
```

Can you split this into inline comments? Some parts of the comment are self explaining in the code

Ubax (approved):

PR #43725, opened 2026-03-06
[expo-image-manipulator] preserve orientation for local images on iOS

maxsz:
# Why

`expo-image-manipulator` was loading local HEIC images incorrectly on iOS before resizing/compressing them.

In our app, the original HEIC selected from the photo library displayed correctly before upload, but after running through `expo-image-manipulator` and being exported as JPEG, the uploaded image came out rotated incorrectly. 

For local file URLs, `expo-image-manipulator` currently uses React Native's `RCTImageLoader`. For HEIC files on iOS, that path appears to lose the source orientation semantics needed for later manipulation/export. As a result, the manipulator writes a JPEG with the wrong physical pixel orientation.

# How

This change updates `packages/expo-image-manipulator/ios/ImageManipulatorUtils.swift`.

Before:
  - `data:` URLs were decoded directly with `UIImage(data:)`
  - `ph://` and `assets-library://` URLs used the Photos-specific loading path
  - other URLs, including readable local file URLs, went through `appContext.imageLoader` / `RCTImageLoader`

After:
  - `data:` URLs still use `UIImage(data:)`
  - `ph://` and `assets-library://` still use the Photos path
  - readable local file URLs are now loaded directly with `Data(contentsOf:)` + `UIImage(data:)`
  - only non-readable / non-file URLs continue to use `appContext.imageLoader`

The reason for this split is that image manipulation needs the original image orientation preserved before resize/compression/export. For local HEIC files on iOS, direct UIKit decoding preserves that correctly, while the `RCTImageLoader` path does not appear to.

# Test Plan

I tested this manually with a real HEIC sample on iOS.

Reviewer reproduction:
  1. Use a portrait HEIC image from an iPhone photo library.
  2. In any iOS app flow that uses `expo-image-manipulator` to resize/compress a local image to JPEG, run the image through the manipulator.
  3. Before this change, the exported JPEG may be rotated incorrectly.
  4. After this change, the exported JPEG should remain upright.

# Checklist
- [x] I added a `changelog.md` entry and rebuilt the package sources according to [this short guide](https://github.com/expo/expo/blob/main/CONTRIBUTING.md#-before-submitting)
- [x] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: updated a module plugin).
- [x] Conforms with the [Documentation Writing Style Guide](https://github.com/expo/expo/blob/main/guides/Expo%20Documentation%20Writing%20Style%20Guide.md)

PR #47877, opened 2026-07-17
[image-manipulator] Add maxWidth/maxHeight options that downsample large images while decoding

### LINKED ISSUE #40158: expo-image-manipulator: Memory crash (EXC_RESOURCE) when processing large images on iOS

https://github.com/RamboWasReal/expo-image-manipulator-repro

### Steps`
1. Install: `npm install`
2. **Connect physical iOS device via USB**
3. Run on physical device: `npm run ios`
4. In app: Tap "64MB Image" → Tap "Process Image"

When processing large images (50MB+), `expo-image-manipulator` causes excessive memory allocation during the Core Video processing pipeline. A 64MB input
image triggers over 3.3GB of memory usage (50x expansion), leading to iOS terminating the app with:

Thread XX: EXC_RESOURCE (RESOURCE_TYPE_MEMORY: high watermark memory limit exceeded) (limit=3376 MB) #0 0xXXXXXXXX in
vConvert_PermuteChannels_ARGB8888_CV_vec ()

The crash occurs in the `vConvert_PermuteChannels_ARGB8888_CV_vec` function during ARGB8888 pixel format conversion, indicating inefficient memory management
in the image manipulation process. This suggests the library is not properly handling memory allocation for large image operations, causing memory usage to
scale disproportionately with input size.

### Environment

```text
expo-env-info 2.0.7 environment info:
    System:
      OS: macOS 26.0.1
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 22.17.0 - ~/.local/state/fnm_multishells/5667_1759405133461/bin/node
      npm: 10.9.2 - ~/.local/state/fnm_multishells/5667_1759405133461/bin/npm
    Managers:
    SDKs:
      iOS SDK:
        Platforms: DriverKit 25.0, iOS 26.0, macOS 26.0, tvOS 26.0, visionOS 26.0, watchOS 26.0
    IDEs:
      Android Studio: 2025.1 AI-251.26094.121.2513.14007798
      Xcode: 26.0.1/17A400 - /usr/bin/xcodebuild
    npmPackages:
      expo: ~54.0.12 => 54.0.12
      react: 19.1.0 => 19.1.0
      react-native: 0.81.4 => 0.81.4
    npmGlobalPackages:
      eas-cli: 16.18.0
    Expo Workflow: bare
```

### Expo Doctor Diagnostics

```text
17/17 checks passed. No issues detected!
```

jiunshinn:
# Why

Fixes #40158 (together with facebook/react-native#54184).

`expo-image-manipulator` always decodes the source image at full resolution. For large photos this materializes multiple full-sized RGBA bitmaps (decode + the unconditional orientation-fix redraw), so a 64 MB JPEG can transiently need gigabytes of memory and iOS terminates the app with `EXC_RESOURCE (RESOURCE_TYPE_MEMORY)`. The React Native side was improved in facebook/react-native#54184 (included since RN 0.83), but as noted there, the largest images still crash — loading a full-sized large image is inherently risky.

This implements the API suggested by @tsapeta in the issue: an option to downscale the image by specifying the maximum size, as in `useImage` from `expo-image`, so the full-sized bitmap is never created in the first place.

# How

New optional `{ maxWidth, maxHeight }` argument for `ImageManipulator.manipulate` and `useImageManipulator` (both bounds cap the *decoded* size, preserving the aspect ratio; images that already fit are untouched, so there is no upscaling):

```ts
const context = ImageManipulator.manipulate(uri, { maxWidth: 4000, maxHeight: 4000 });
```

- **iOS**: local files and `data:` URLs are decoded with `CGImageSourceCreateThumbnailAtIndex` + `kCGImageSourceThumbnailMaxPixelSize`, which decodes directly to the bounded size without materializing the full bitmap (the same strategy SDWebImage uses). The bounds are applied to the EXIF-oriented dimensions and the orientation is baked in by `kCGImageSourceCreateThumbnailWithTransform`. `ph://` assets pass a bounded `targetSize` to `PHImageManager`. Images loaded through `RCTImageLoader` (remote URLs) and already-decoded `SharedRef<UIImage>` sources are downscaled after loading, since those paths can't decode within bounds.
- **Android**: added a `loadImageForManipulationFromURL(url, maxWidth, maxHeight, listener)` overload to `ImageLoaderInterface` (with a default implementation falling back to the full-sized load, so third-party implementers don't break) and implemented it in `expo-image-loader` with Glide's `centerInside()` + bounded target size, mirroring how `expo-image`'s `useImage` passes `maxWidth`/`maxHeight` to Glide. The manipulator additionally applies an exact bound after loading to guarantee the contract regardless of the loader implementation.
- **Web**: the source canvas is drawn at the bounded size.

When no options are passed, every path is byte-for-byte the existing behavior — including the direct local-file read added in #43725 to preserve HEIC/EXIF orientation.

Deliberately **not** done here, to keep the change scoped: the deprecated `manipulateAsync` API doesn't expose the new options, and the always-on full-size orientation-fix redraw on iOS (`ImageFixOrientationTransformer` runs even when the orientation is already `.up`) is left untouched — happy to address that as a follow-up if you're open to it, since skipping it would roughly halve the peak memory for unrotated images even without `maxWidth`/`maxHeight`.

# Test Plan

New unit tests (written first, red → green):

- iOS (new `Tests` test spec + `apps/native-tests` wiring): 12 Swift Testing cases covering the bound math, bounded file/data decoding, EXIF-rotated sources (bounds apply to oriented dimensions, orientation baked in), no-upscale behavior, and post-load downscaling — `xcodebuild test -workspace apps/native-tests/ios/NativeTests.xcworkspace -scheme ExpoImageManipulator-Unit-Tests -destination 'platform=iOS Simulator,name=iPhone 17 Pro'` → **12/12 passed**.
- Android: 9 Robolectric cases for the bound math and bitmap downscaling — `et native-unit-tests --packages expo-image-manipulator -p android` → **passed**. The Glide decode path itself is not unit-tested (requires instrumentation); it mirrors `expo-image`'s production `ImageLoadTask` and the module re-applies the bounds after loading.
- JS/web: `pnpm --filter expo-image-manipulator test` → **8 suites, 92 tests passed** (includes new `boundedSize` tests).
- `et check-packages expo-image-manipulator` → all checks passed (build, lint, format, test).
- Android compile of the interface/loader changes: `./gradlew :expo-modules-core:compileReleaseKotlin :expo-image-loader:compileReleaseKotlin` → **BUILD SUCCESSFUL**.

Memory measurement (peak RSS via `getrusage`, measured in the iOS test target on the iPhone 17 Pro simulator, 12000×9000 generated JPEG, temporary test not committed):

| Path | Peak RSS delta | Result |
| --- | --- | --- |
| Full decode + full-size redraw (current pipeline) | **+1242 MB** | 12000×9000 |
| `decodeDownsampledImage(maxWidth: 2000)` (this PR) | **+50 MB** | 2000×1500 |

Not run: an on-device end-to-end of the original repro app. The pod test target is tool-hosted, which Xcode can't run on physical devices, and I didn't want to inflate the diff with a demo-app change — the repro from #40158 with `manipulate(uri, { maxWidth: 4000 })` is the straightforward manual verification if desired.

# Checklist

- [x] Documentation is up to date to reflect these changes: TSDoc on the new `ImageLoadOptions` type and `manipulate`/`useImageManipulator` signatures (picked up by the API reference generator).
- [x] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: no addition of native code that would require a config plugin change).
- [x] Added a `CHANGELOG.md` entry (expo-image-manipulator, expo-modules-core, expo-image-loader).


jiunshinn:
**A quick guide to the diff** — 24 files / +630 lines is bigger than the change really is, so here's the breakdown for review:

- **~250 lines are new tests**: 12 Swift Testing cases, 9 Robolectric cases, and 6 Jest cases (written red → green), plus the one-time iOS test-infra setup — the `test_spec` in the podspec and adding `expo-image-manipulator` to `apps/native-tests`, which is also all the `Podfile.lock`/`pnpm-lock.yaml` churn. This package had no iOS unit tests before.
- **The substantive change is ~350 lines in `expo-image-manipulator`**: `ios/ImageDownsampling.swift` and `android/.../ImageDownsampling.kt` contain the actual downsampling logic; the remaining files just thread the new optional argument through `manipulate` on each platform.
- **`expo-modules-core` is +8 lines**: one new `loadImageForManipulationFromURL(url, maxWidth, maxHeight, listener)` overload with a default implementation that falls back to the existing full-size load — no breaking change for existing `ImageLoaderInterface` implementers. It's needed so Android can bound the decode inside Glide instead of downscaling after the full-size bitmap already exists.
- **`expo-image-loader` is +51 lines**: the Glide implementation of that overload (mirroring `expo-image`'s `ImageLoadTask`), plus extracting the previously duplicated `CustomTarget` into a shared factory.

Suggested reading order: `ios/ImageDownsampling.swift` → `ios/ImageManipulatorUtils.swift` → `android/.../ImageLoaderService.kt` → tests.

If you'd prefer this without the Android loader changes (keeping the bounds contract via post-load downscaling only) or split into separate PRs, happy to restructure.

packages/expo-image-manipulator/android/src/main/java/expo/modules/imagemanipulator/ImageManipulatorModule.kt
```diff
@@ -28,50 +28,58 @@ import kotlin.coroutines.resume
...
-            override fun onFailure(cause: Throwable?) {
-              continuation.resumeWithException(ImageLoadingFailedException(url.toString(), cause.toCodedException()))
-            }
+          override fun onFailure(cause: Throwable?) {
+            continuation.resumeWithException(ImageLoadingFailedException(url.toString(), cause.toCodedException()))
           }
-        )
+        }
+        if (maxWidth != null || maxHeight != null) {
+          imageLoader.loadImageForManipulationFromURL(url.toString(), maxWidth ?: Int.MAX_VALUE, maxHeight ?: Int.MAX_VALUE, resultListener)
```
tsapeta: [🏗️ **design**] Prefer Glide's `Target.SIZE_ORIGINAL` over `Int.MAX_VALUE` for the unset axis. `expo-image`'s production path (`records/ImageLoadOptions.kt` defaults both dimensions to `Target.SIZE_ORIGINAL`, then `ImageLoadTask.kt` does `.centerInside().submit(maxWidth, maxHeight)`) uses `SIZE_ORIGINAL` (`Integer.MIN_VALUE`) as the canonical "no bound on this axis" sentinel — Glide special-cases it, and `ImageViewWrapperTarget.isDimensionValid` documents exactly this (`size > 0 || size == Target.SIZE_ORIGINAL`).

`Int.MAX_VALUE` happens to work here only because `.centerInside()` caps the scale at 1.0 and picks the more restrictive axis, so the ~2.1B value never drives an allocation. But it's a literal requested pixel size to Glide, not a sentinel — fragile if the downsample strategy or transform changes, and inconsistent with the sibling module. Passing `Target.SIZE_ORIGINAL` for the missing bound keeps the two paths aligned.

packages/expo-image-manipulator/ios/ImageDownsampling.swift
```diff
@@ -0,0 +1,101 @@
...
+ Returns the image untouched when it already fits within the bounds.
+ */
+internal func downscaledIfExceedsBounds(_ image: UIImage, maxWidth: Double?, maxHeight: Double?) -> UIImage {
+  let width = image.size.width * image.scale
+  let height = image.size.height * image.scale
+  guard let maxPixelSize = downsampledMaxPixelSize(width: width, height: height, maxWidth: maxWidth, maxHeight: maxHeight) else {
+    return image
+  }
+  let scale = maxPixelSize / max(width, height)
+  let targetSize = CGSize(width: (width * scale).rounded(), height: (height * scale).rounded())
```
tsapeta: [💡 suggestion] `targetSize` can round a dimension to 0 for extreme aspect ratios (e.g. a 10000×1 image bounded to `maxWidth: 50` → height `1 * 0.005` rounds to 0), and `UIGraphicsImageRenderer` with a zero-height size produces an invalid/empty image. Android's `boundedSize` and web's `boundedSize` both guard this with `max(1, …)`; this iOS path (and `loadImageFromPhotoLibrary`'s `size` computation in `ImageManipulatorUtils.swift`) don't. Clamping each side to `max(1, …)` would make the three platforms consistent. Edge case, but cheap to close.

tsapeta (changes requested):

packages/expo-image-manipulator/android/src/main/java/expo/modules/imagemanipulator/ImageManipulatorModule.kt
jiunshinn: Switched to the `SIZE_ORIGINAL` sentinel (6d8ede53735, reworked in 608623d5080): the interface overload takes nullable bounds since `expo-image-manipulator` does not depend on Glide, and `ImageLoaderService` maps `null` to `Target.SIZE_ORIGINAL` at the actual Glide call.

packages/expo-image-manipulator/ios/ImageDownsampling.swift
jiunshinn: Fixed in 4a19b745d66 — both spots now clamp with `max(1, …)`, with a regression test: 400×2 bounded to `maxWidth: 50` previously produced a `{0, 0}` image with a nil `cgImage`.

jiunshinn:
Hi @tsapeta, Is there anything else I can do help for this PR get merged?

tsapeta (approved):

PR #47889, opened 2026-07-17
[notifications] prevent crash on notification tap when getLaunchIntentForPackage throws

nunocaseiro:
> _Disclosure: this crash was investigated and the fix prepared with the help of Claude Code, then reviewed before submitting. The root cause and stack trace below are from a real production crash (Firebase Crashlytics)._

# Why

`NotificationForwarderActivity.onCreate` can crash the app on notification tap with:

```
java.lang.RuntimeException: Unable to start activity ComponentInfo{…/expo.modules.notifications.service.NotificationForwarderActivity}: java.lang.NullPointerException: class name is null
Caused by: java.lang.NullPointerException: class name is null
  at android.content.ComponentName.<init>(ComponentName.java:103)
  at android.content.Intent.setClassName(Intent.java:10044)
  at android.app.ApplicationPackageManager.getLaunchIntentForPackage(...)
  at expo.modules.notifications.service.NotificationForwarderActivity.onCreate(NotificationForwarderActivity.kt:27)
```

Line 27 is the **fallback inside the existing `catch` block**. The sequence is:

1. the `try` fails to recover notification data from the intent extras (the case the current `catch (e: IllegalArgumentException)` was added for), then
2. the fallback — `packageManager.getLaunchIntentForPackage(packageName)` — **itself throws**. On some OEM ROMs (observed on a Huawei device, Android 11, no Google Mobile Services) `getLaunchIntentForPackage` calls `Intent.setClassName(pkg, null)` internally when no launcher activity resolves, throwing `NullPointerException: "class name is null"` instead of returning `null`.

Because the fallback isn't guarded, this NPE escapes `onCreate` and crashes the app — the exact situation the fallback was meant to prevent.

This fallback was introduced in #43203 as *"a safety net [so] NotificationForwarderActivity … opens the app instead of crashing."* This PR hardens that safety net against the case where the safety net itself throws.

Seen in production via Firebase Crashlytics on `expo-notifications` 56.0.15 (Huawei / Android 11 / no-GMS device); the code path is unchanged on `main`. No existing open issue or PR covers this failure mode.

# How

Root cause: `packageManager.getLaunchIntentForPackage(...)` can throw (rather than return `null`) on some OEM ROMs when no launcher activity resolves. `expo-notifications` calls it at two unguarded sites, so this PR guards both:

- **`ExpoHandlingDelegate.getMainActivityLauncher`** (the happy path — used by `openAppToForeground` on every notification tap that opens the app): wrap the lookup so a throw returns `null`, routing into the existing "No launch intent found" warning branch instead of crashing.
- **`NotificationForwarderActivity` fallback** (the safety net added in #43203, where the observed crash occurred): wrap the fallback `getLaunchIntentForPackage`/`startActivity` in its own `try/catch`.

The outer `catch (e: IllegalArgumentException)` in the forwarder is intentionally left as-is: with the `NullPointerException` source now guarded at both sites, widening it would be speculative and out of scope.

Net effect: a missing/throwing launcher intent now degrades to "open the app" or a logged no-op + `finish()`, never a crash on tap.

# Test Plan

- Manual: tapping a notification whose forward path fails no longer crashes; the app opens to foreground. On a ROM where `getLaunchIntentForPackage` throws, the app no longer crashes (logged warning + `finish()`).
- No behavior change on the happy path (successful forward + broadcast) or on the previously-handled `IllegalArgumentException` path.

# Checklist

- [x] I added a `changelog.md` entry (Android-only Kotlin change; no generated `build/` sources to rebuild).
- [ ] Documentation is up to date to reflect these changes (n/a — internal behavior, no public API change).
- [x] This diff will work correctly for `npx expo prebuild` & EAS Build (no config changes).


vonovak (approved):
thank you!

PR #47892, opened 2026-07-17
[autolinking][ios] Fix HEADER_SEARCH_PATHS corruption when Xcodeproj normalized the setting to an Array

### LINKED ISSUE #47891: [expo-modules-autolinking][iOS] precompiled_modules.rb interpolates HEADER_SEARCH_PATHS that Xcodeproj may have normalized to an Array — embeds Array#inspect, severs $(inherited)

This is a code-level defect in `expo-modules-autolinking`'s CocoaPods post-install hook that is fully demonstrable below without an app repository; the trigger conditions and a self-contained demonstration script are included. A one-line PR with the fix is attached (see linked PR).

### What happened

`scripts/ios/precompiled_modules.rb` → `configure_header_search_paths` appends ExpoModulesJSI header search paths to every pod target with string interpolation:

```ruby
existing = config.build_settings['HEADER_SEARCH_PATHS'] || '$(inherited)'
unless existing.include?(paths_string)
  config.build_settings['HEADER_SEARCH_PATHS'] = "#{existing} #{paths_string}"
end
```

(verified today at `expo-modules-autolinking@57.0.8`, the latest published version, lines 898–901, and identical on `main`.)

`existing` is assumed to be a String, but Xcodeproj mutates known multi-value build settings — `HEADER_SEARCH_PATHS` among them — from a space-separated String into an **Array in place** whenever the project is serialized (`XCBuildConfiguration#to_hash_as` → `normalize_array_settings`, which mutates the live `build_settings` hash). React Native's own post-install helper does exactly that mid-`post_install`: `ReactNativePodsUtils.update_search_paths` calls `project.save()` when `USE_FRAMEWORKS` is set (`react-native/scripts/cocoapods/utils.rb`). Any other Podfile hook that saves the project, or that assigns an Array itself, produces the same state.

When `existing` is an Array:

- `existing.include?(paths_string)` checks Array *element* equality instead of substring containment, so the guard never matches, and
- `"#{existing} #{paths_string}"` embeds the Array's `inspect` form into the build setting.

The resulting value is a single garbage string like:

```
["$(inherited)", "\"/foo\"", "\"/bar\""] "/expo/jsi/path"
```

`$(inherited)` is now inside a quoted literal, so every xcconfig-level header search path for that target is silently severed — the whole pod graph loses its inherited `-I` flags. With SDK 56/57 precompiled React-Core and dynamic frameworks this surfaces as unrelated-looking "header not found" compile failures in downstream pods, which is painful to trace back here.

### Steps to reproduce

Trigger conditions in a real project: iOS build with precompiled modules enabled (so `configure_header_search_paths` runs) + `use_frameworks!` (so React Native's `update_search_paths` runs `project.save()` before this hook reads the settings). Platform: iOS. Any package manager.

Self-contained demonstration of the mechanism (only needs the `xcodeproj` gem, which CocoaPods ships):

```ruby
require "xcodeproj"
require "tmpdir"
Dir.mktmpdir do |dir|
  proj = Xcodeproj::Project.new(File.join(dir, "T.xcodeproj"))
  target = proj.new_target(:framework, "Demo", :ios)
  cfg = target.build_configurations.first
  cfg.build_settings["HEADER_SEARCH_PATHS"] = "$(inherited) \"/tmp/a\" \"/tmp/b\""
  proj.save  # what ReactNativePodsUtils.update_search_paths does mid-post_install
  existing = cfg.build_settings["HEADER_SEARCH_PATHS"] || "$(inherited)"
  paths_string = "\"/tmp/c\""
  unless existing.include?(paths_string)
    puts "#{existing} #{paths_string}"
  end
end
```

Output (xcodeproj 1.27.0):

```
["$(inherited)", "\"/tmp/a\"", "\"/tmp/b\""] "/tmp/c"
```

Expected: `$(inherited) "/tmp/a" "/tmp/b" "/tmp/c"`.

### Suggested fix

Join the Array back into the space-separated form before interpolating:

```diff
--- a/scripts/ios/precompiled_modules.rb
+++ b/scripts/ios/precompiled_modules.rb
@@ -896,6 +896,7 @@ module Expo
         installer.pods_project.targets.each do |target|
           target.build_configurations.each do |config|
             existing = config.build_settings['HEADER_SEARCH_PATHS'] || '$(inherited)'
+            existing = existing.join(' ') if existing.is_a?(Array)
             unless existing.include?(paths_string)
               config.build_settings['HEADER_SEARCH_PATHS'] = "#{existing} #{paths_string}"
             end
```

We have been running this as a pnpm patch in production builds; it restores `$(inherited)` propagation across the pod graph.

### Environment

Observed with Expo SDK 56 (expo-modules-autolinking 56.0.20) on macOS/Xcode release builds with precompiled React-Core and dynamic frameworks; the defective code is unchanged in 57.0.8 (latest) and on `main`.

### Expo Doctor Diagnostics

Not applicable — library code defect demonstrated above; no app project involved.

joedeleeuw:
# Why

Xcodeproj can normalize `HEADER_SEARCH_PATHS` to an Array; interpolating it corrupts inherited paths. Fixes #47891.

# How

Join Array values before checking or appending paths.

# Test Plan

- the #47891 Ruby repro now outputs `$(inherited) "/tmp/a" "/tmp/b" "/tmp/c"`
- `ruby -c packages/expo-modules-autolinking/scripts/ios/precompiled_modules.rb`
- `git diff --check upstream/main..HEAD`

# Checklist

- [x] added a changelog entry
- [ ] verified `npx expo prebuild` and EAS Build
- [ ] documentation style guide applies

### DIFF
--- a/packages/expo-modules-autolinking/CHANGELOG.md
+++ b/packages/expo-modules-autolinking/CHANGELOG.md
@@ -13,6 +13,7 @@
 
 ### 🐛 Bug fixes
 
+- [iOS] Preserve inherited header search paths when Xcodeproj normalizes `HEADER_SEARCH_PATHS` into an Array. ([#47892](https://github.com/expo/expo/pull/47892) by [@joedeleeuw](https://github.com/joedeleeuw))
 - [iOS] Only stub a source pod bundled inside a prebuilt XCFramework (e.g. `SDWebImage` in `ExpoImage.xcframework`) when one of its consumers also links the owning prebuilt pod. An app extension that depends on such a pod on its own now keeps building it from source instead of failing to link with undefined symbols. ([#47847](https://github.com/expo/expo/pull/47847) by [@chrfalch](https://github.com/chrfalch))
 - [iOS] Pass target name to the generateModulesProviderCommand to match it against inline modules targets, when checking if inline modules should be autolinked with that target. ([#47502](https://github.com/expo/expo/pull/47502) by [@HubertBer](https://github.com/HubertBer))
 - [Android] Scan the whole Kotlin file for its `package` declaration when registering inline modules, so modules with long comments (for example, a license header) before the `package` declaration are no longer silently skipped. ([#47656](https://github.com/expo/expo/pull/47656) by [@HubertBer](https://github.com/HubertBer))
--- a/packages/expo-modules-autolinking/scripts/ios/precompiled_modules.rb
+++ b/packages/expo-modules-autolinking/scripts/ios/precompiled_modules.rb
@@ -896,6 +896,8 @@ module Expo
         installer.pods_project.targets.each do |target|
           target.build_configurations.each do |config|
             existing = config.build_settings['HEADER_SEARCH_PATHS'] || '$(inherited)'
+            # Xcodeproj normalizes known multi-value settings to Arrays when serializing.
+            existing = existing.join(' ') if existing.is_a?(Array)
             unless existing.include?(paths_string)
               config.build_settings['HEADER_SEARCH_PATHS'] = "#{existing} #{paths_string}"
             end


brentvatne:
Closing — same `HEADER_SEARCH_PATHS` Array corruption fix landed in https://github.com/expo/expo/pull/48665. Duplicate of #47552. Thanks!

PR #47902, opened 2026-07-17
[app-metrics][updates] Rename download completion state

Kudo:
# Why

`downloadComplete` is used when no update is available, so its name was unclear.

# How

Renamed it to `downloadCompleteUnavailable` in the native update state machines and app-metrics parsers. Updated tests and the changelog.

# Test Plan

- Ran Swift syntax parsing for the changed Swift files.
- Confirmed the old name is absent from native and TypeScript sources.
- Android Gradle compilation did not start locally because the app could not resolve its React Native Gradle plugin.

# Checklist

### DIFF
--- a/packages/expo-app-metrics/CHANGELOG.md
+++ b/packages/expo-app-metrics/CHANGELOG.md
@@ -22,7 +22,7 @@
 
 ### 💡 Others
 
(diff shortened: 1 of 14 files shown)


Kudo:
the macos ci error feels unrelated

alanjhughes (approved):

douglowder (approved):

PR #47906, opened 2026-07-17
[cli] update agent cli detector

davidmokos:
## Summary
- update agent-cli-detector from 0.1.2 to 0.1.4 for @expo/cli and submit-expo-feedback
- the latest version adds detection signals for Rork, Bolt and Replit


krystofwoldrich (approved):
Sorry for the delay, looks good. Will need backport to all SDK versions where we have the cli detector. I added sdk-57, not sure if it's in older versions as well.

kitten:
This won't necessarily need to be backported or picked, since these are patch updates in range of the semver specifiers, so I'd skip it fwiw

PR #47921, opened 2026-07-17
[brownfield][android] Add support Fused Library

gabrieldonadel:
# Why

In its current state `npx expo-brownfield build:android` publishes ~30 separate Maven coordinates per release (one per autolinked Expo Android module). This PR adds an opt-in single-fat-AAR mode so consumers only need to publish (and depend on) one AAR per variant.

# How

Add new `--fused` flag to `expo-brownfield build:android`. When set, the CLI targets the new sibling Gradle subprojects emitted by the config plugin,`<library>-fused-release` and `<library>-fused-debug`, producing a single AAR containing every autolinked module + each module's transitive Maven dependencies (ExoPlayer for expo-video, CameraX for expo-camera, etc.).
Key pieces:

- **Inert by default.** Prebuild always emits the two fused siblings, but their build scripts only activate when the CLI passes `-Pbrownfield.fused=true`. Normal builds and IDE sync pay no configuration cost, and non-fused publishes can't accidentally trigger a fat-AAR build via Gradle's cross-project task-name matching.
- **One sibling per variant.** `-fused-release` and `-fused-debug` publish distinct coordinates so hosts wire them as `releaseImplementation` / `debugImplementation`. A `BuildTypeAttr` override makes the debug sibling actually fuse debug bytecode (AGP's fused-library resolves everything as release by default).
- **Fused-only side effects.** The same property skips the publish plugin's per-module re-publish loop and force-bumps AGP to 8.13 just for fused builds (8.12's `rewriteClasses` can't read Kotlin sealed-class bytecode).
- **Fused contents are discovered, not hardcoded.** The sibling resolves the autolinked dependency graph and includes every module plus their external AARs (ExoPlayer, CameraX, Glide, ...). Three groups stay external: the brownfield host baseline (RN runtime, Kotlin stdlib, Material/Guava/OkHttp/Okio), `androidx.*` (AGP's class rewriter can't resolve framework attrs in their styleables; `androidx.camera`/`media3` are allowlisted), and auto-detected pom-only KMP umbrella modules. Everything external is re-declared in the published POM/`.module` so consumers resolve it themselves.
- **Variant-safe RN resolution.** Published metadata annotates the `react-android`/`hermes-android` edges with the sibling's build type, so consumers link the same RN variant the fused codegen `.so` files were built against (a mismatch crashes at startup).
- **Escape hatches.** `-Pbrownfield.fused.skip`, `strip-packages`, `androidx-fuse`, and `exclude-transitive` cover dependency graphs the auto-detection can't; documented in the new Fused mode docs section.
- **Runtime fixes along the way.** `BrownfieldActivity` implements `DefaultHardwareBackBtnHandler` and finishes on default back (avoids an infinite native/JS back-press loop), consumer ProGuard rules ship in the AAR so host R8 keeps reflection-loaded Expo classes, and the dev-menu patch passes `useDevSupport` explicitly so dev tooling follows the AAR variant.

# Test Plan

Automated:

- CLI e2e: fused task construction routed to the sibling subprojects, `--all` expanding to one Gradle invocation per variant, `-Pbrownfield.fused=true` forwarding (including for tasks passed via `-t`), plus existing snapshot/dry-run coverage.
- Plugin e2e: both fused siblings emitted with correct group/version/namespace/variant substitutions and the inert-by-default guard, settings.gradle includes, gradle.properties Fused Library opt-ins, and the conditional AGP force block. (Also fixed the suite's `expectFile` helper, which was async-but-unawaited, so its content assertions actually fail tests now.)

Manual, in `apps/brownfield-tester/expo-app`:

```sh
# Default — publishes both siblings
npx expo-brownfield build:android --fused --repo MavenLocal --verbose

# Or per variant
npx expo-brownfield build:android --fused --release --repo MavenLocal --verbose
npx expo-brownfield build:android --fused --debug   --repo MavenLocal --verbose
``

### DIFF
--- a/packages/expo-brownfield/CHANGELOG.md
+++ b/packages/expo-brownfield/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### 🎉 New features
 
 - [iOS] Add hostProvidedFrameworks option to skip Frameworks provided by the host app ([#46355](https://github.com/expo/expo/pull/46355) by [@gabrieldonadel](https://github.com/gabrieldonadel))
+- [android] Add support Fused Library ([#45878](https://github.com/expo/expo/pull/45878) by [@gabrieldonadel](https://github.com/gabrieldonadel))
 
 ### 🐛 Bug fixes
 
--- a/packages/expo-brownfield/CHANGELOG.md
+++ b/packages/expo-brownfield/CHANGELOG.md
@@ -7,7 +7,7 @@
 ### 🎉 New features
 
 - [iOS] Add hostProvidedFrameworks option to skip Frameworks provided by the host app ([#46355](https://github.com/expo/expo/pull/46355) by [@gabrieldonadel](https://github.com/gabrieldonadel))
-- [android] Add support for publishing a single fused AAR via AGP Fused Library ([#45878](https://github.com/expo/expo/pull/45878) by [@gabrieldonadel](https://github.com/gabrieldonadel))
+- [android] Add support for publishing a single fused AAR via AGP Fused Library ([#47921](https://github.com/expo/expo/pull/45878) by [@gabrieldonadel](https://github.com/gabrieldonadel))
 
 ### 🐛 Bug fixes
 
--- a/packages/expo-brownfield/CHANGELOG.md
+++ b/packages/expo-brownfield/CHANGELOG.md
@@ -7,7 +7,7 @@
 ### 🎉 New features
 
 - [iOS] Add hostProvidedFrameworks option to skip Frameworks provided by the host app ([#46355](https://github.com/expo/expo/pull/46355) by [@gabrieldonadel](https://github.com/gabrieldonadel))
-- [android] Add support Fused Library ([#45878](https://github.com/expo/expo/pull/45878) by [@gabrieldonadel](https://github.com/gabrieldonadel))
+- [android] Add support for publishing a single fused AAR via AGP Fused Library ([#45878](https://github.com/expo/expo/pull/45878) by [@gabrieldonadel](https://github.com/gabrieldonadel))
 
 ### 🐛 Bug fixes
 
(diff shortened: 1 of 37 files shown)

### CODE REVIEW

File: packages/expo-brownfield/CHANGELOG.md [hunk 1]
```diff
@@ -7,6 +7,7 @@
 ### 🎉 New features
 
 - [iOS] Add hostProvidedFrameworks option to skip Frameworks provided by the host app ([#46355](https://github.com/expo/expo/pull/46355) by [@gabrieldonadel](https://github.com/gabrieldonadel))
+- [android] Add support for publishing a single fused AAR via AGP Fused Library ([#45878](https://github.com/expo/expo/pull/45878) by [@gabrieldonadel](https://github.com/gabrieldonadel))
```
chrfalch: Change to this PR :) 

```suggestion
- [android] Add support for publishing a single fused AAR via AGP Fused Library ([#47921](https://github.com/expo/expo/pull/45878) by [@gabrieldonadel](https://github.com/gabrieldonadel))
```

packages/expo-brownfield/cli/src/utils/config.ts
```diff
@@ -159,15 +163,59 @@ const resolveLibrary = (options: OptionValues): string => {
...
+const resolveLocalRepositoriesFromAppConfig = (): string[] => {
+  let publishing: unknown;
+  try {
+    const { exp } = getConfig(process.cwd(), { skipSDKVersionRequirement: true });
+    const plugin = exp.plugins?.find(
+      (entry): entry is [string, any] => Array.isArray(entry) && entry[0] === 'expo-brownfield'
+    );
+    publishing = plugin?.[1]?.android?.publishing;
+  } catch {
+    // App config could not be evaluated here — fall through to the plugin's
```
chrfalch, line 201: Should we log (if `--verbose` is enabled) the error if `getConfig` fails?

chrfalch (approved):
Approved with a few small nits :)

packages/expo-brownfield/cli/src/utils/config.ts
gabrieldonadel: Updated

PR #47928, opened 2026-07-18
[cli] Add sandbox telemetry context

davidmokos:
Adds `sandbox-cli-detector` to Expo CLI telemetry so events recorded inside known sandbox environments include lightweight sandbox context. The CLI uses `sandbox-cli-detector@0.2.0`, which exposes a CommonJS entry point, so detection stays synchronous and does not participate in telemetry initialization.

When a known sandbox is detected, telemetry event context includes:

```ts
context.sandbox_provider = string
```

When no known sandbox is detected, `context.sandbox_provider` is omitted to keep RudderStack and BigQuery payloads sparse. CLI version is already available in telemetry, so analysis can distinguish newer CLI versions with no detected sandbox from older CLI versions without this integration.

Tests cover detected and undetected telemetry payloads. Locally verified with `format`, `lint`, `depscheck`, `typecheck`, and the focused telemetry Jest tests.


packages/@expo/cli/src/utils/telemetry/Telemetry.ts
```diff
@@ -103,6 +105,7 @@ export class Telemetry {
         context: {
           ...this.context,
           sessionId: this.actor.sessionId,
+          ...(sandboxId ? { sandbox_id: sandboxId } : {}),
```
sjchmiela: as mention in the other PR i think "name" would suite this better, but i defer to @byCedric's judgment here

pnpm-workspace.yaml
```diff
@@ -62,3 +62,4 @@ minimumReleaseAgeExclude:
   - 'babel-preset-expo'
   - 'jest-expo'
   - '2g'
+  - 'sandbox-cli-detector'
```
sjchmiela: here too i think i would not add this

sjchmiela (approved):

packages/@expo/cli/src/utils/telemetry/Telemetry.ts
davidmokos: renamed to sandbox_provider - good call

PR #47904, opened 2026-07-17
[cli] add feedback subject option

davidmokos:
Adds an optional `--subject` / `-s` argument to `submit-expo-feedback` for identifying the specific docs page, skill, CLI command, MCP server, or other item the feedback is about. The subject is trimmed, omitted when empty, and included in feedback metadata. Adds tests for subject normalization and payload metadata.

Tested with `pnpm --filter submit-expo-feedback test`, `pnpm --filter submit-expo-feedback typecheck`, and `pnpm --filter submit-expo-feedback lint`.

PR #47932, opened 2026-07-18
[cli] add agent feedback nudge to non-interactive output

davidmokos:
Adds a compact feedback command to the end of short-lived Expo CLI output, or once the dev server is ready for `expo start`, when running non-interactively under a detected coding agent.
It reuses `agent-cli-detector`, includes the Expo command name as the feedback subject, and leaves interactive and non-agent executions unchanged.
Tests cover exit-time output, deduplicated dev-server-ready output, and both gating conditions; the package typecheck, lint, focused tests, and built-CLI smoke tests pass.

## Example agent output

Long-running dev server:

```text
Starting project at /Users/davidmokos/Developer/experiments/friend-draw
Metro is running in CI mode, reloads are disabled. Remove CI=true to enable watch mode.
Starting Metro Bundler

Waiting on http://localhost:8082

Logs for your project will appear below.

Expo CLI issue? Report it: npx --yes submit-expo-feedback@latest --category expo-cli --subject "start" "<what happened and how to reproduce>"
(node:39610) [DEP0169] DeprecationWarning: `url.parse()` behavior is not standardized and prone to errors that have security implications. Use the WHATWG URL API instead. CVEs are not issued for `url.parse()` vulnerabilities.
(Use `node --trace-deprecation ...` to show where the warning was created)
```

Successful export:

```text
Starting Metro Bundler
Android Bundled 4821ms index.js (614 modules)
Exporting 42 assets
Exported to: dist

Expo CLI issue? Report it: npx --yes submit-expo-feedback@latest --category expo-cli --subject "export" "<what happened and how to reproduce>"
```

Failed install:

```text
› Installing 1 SDK 56.0.0 compatible native module using npm
npm error code ERESOLVE
npm error unable to resolve dependency tree

Expo CLI issue? Report it: npx --yes submit-expo-feedback@latest --category expo-cli --subject "install" "<what happened and how to reproduce>"
```

Version command:

```text
56.1.12

Expo CLI issue? Report it: npx --yes submit-expo-feedback@latest --category expo-cli --subject "--version" "<what happened and how to reproduce>"
```

> [!IMPORTANT]
> Do not merge this PR until [#47904](https://github.com/expo/expo/pull/47904) has been published to npm; merging #47904 alone is not sufficient.

### DIFF
--- a/packages/@expo/cli/src/utils/agentFeedback.ts
+++ b/packages/@expo/cli/src/utils/agentFeedback.ts
@@ -8,7 +8,7 @@ export function installAgentFeedback(command: string): void {
 
   process.once('exit', () => {
     process.stdout.write(
-      `\nExpo CLI issue? Report it: npx --yes submit-expo-feedback --category expo-cli --subject ${JSON.stringify(command)} "<what happened and how to reproduce>"\n`
+      `\nExpo CLI issue? Report it: npx --yes submit-expo-feedback@latest --category expo-cli --subject ${JSON.stringify(command)} "<what happened and how to reproduce>"\n`
     );
   });
 }
--- a/packages/@expo/cli/src/utils/agentFeedback.ts
+++ b/packages/@expo/cli/src/utils/agentFeedback.ts
@@ -1,14 +1,27 @@
 import { isInteractive } from './interactive';
 import { getAgentTelemetryContext } from './telemetry/utils/agent';
 
+let pendingFeedback: (() => void) | undefined;
+
 export function installAgentFeedback(command: string): void {
   if (isInteractive() || !getAgentTelemetryContext()) {
     return;
   }
 
-  process.once('exit', () => {
+  const print = () => {
+    if (pendingFeedback !== print) {
+      return;
+    }
+    pendingFeedback = undefined;
+    process.removeListener('exit', print);
     process.stdout.write(
       `\nExpo CLI issue? Report it: npx --yes submit-expo-feedback@latest --category expo-cli --subject ${JSON.stringify(command)} "<what happened and how to reproduce>"\n`
     );
-  });
+  };
+  pendingFeedback = print;
+  process.once('exit', print);
+}
+
+export function printAgentFeedback(): void {
+  pendingFeedback?.();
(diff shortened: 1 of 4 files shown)


packages/@expo/cli/src/utils/agentFeedback.ts
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  const print = () => {
+    if (pendingFeedback !== print) {
+      return;
+    }
+    pendingFeedback = undefined;
+    process.removeListener('exit', print);
+    process.stdout.write(
+      `\nExpo CLI issue? Report it: npx --yes submit-expo-feedback@latest --category expo-cli --subject ${JSON.stringify(command)} "<what happened and how to reproduce>"\n`
```
kitten, line 18: This is too broad, as in, I don't think we should do this. There's very little feedback that can apply "to the Expo CLI" directly. There's only really a few outcomes that can happen of which I see few that will be directly useful:
- the agent does submit Expo CLI feedback, but since that's potentially coupled to logs work, I don't see that being directly useful
- an error happened, but without context, this isn't as valuable as an issue with a repro (even if it goes through an agent)
- the prompt gets ignored, and non CLI feedback gets submitted sometimes, which isn't great for triage and could "overload" us with calls

On the side of implementation, we're fixing a CLI API in place here, but instead of `--help`, to keep it short, we just give a full command. Not only is this relatively long (and may lead to extra agent turns for users on every interaction where the agent does choose to use this), we've decoupled `submit-expo-feedback` from this prompt injection, but that means the contract here can never change.

(if anything that either means we should embed the `feedback` command, or inject an up-to-date string from an export on `submit-expo-feedback`. However, the latter also couples us to `@latest`, and since this is `npx` that won't work; which then also means coupling to `npx` makes some assumptions we typically don't make anymore)

tl;dr: Let's first have an internal draft RFC/proposal on changes like these
cc @byCedric

davidmokos: I agree this might be too big of a change - moving back to draft and let's discuss it

PR #47939, opened 2026-07-19
[font] Support font files that define multiple styles for a single font family

bryanmylee:
# Why

`useFonts`/`loadAsync` only support one font file per `fontFamily` key. For a family with multiple weights/styles (e.g. "Wix Madefor Text" with regular, italic, and bold files), the only option was giving each file its own unrelated fontFamily name – which means the web export's generated `@font-face` CSS never defines `font-weight`/`font-style`, so the browser can't select the correct face for a given weight/style the way it can for a normal web font. Every variant then has to be referenced by its own `fontFamily` string in styles instead of `fontWeight`/`fontStyle`.

With this change, users can call `useFonts`/`loadAsync` like:

```tsx
useFonts([
  {
    fontFamily: "Wix Madefor Text",
    fontDefinitions: [
      {
        path: require("@/assets/fonts/WixMadeforText-Regular.woff2"),
      },
      {
        path: require("@/assets/fonts/WixMadeforText-Italic.woff2"),
        style: "italic",
      },
      {
        path: require("@/assets/fonts/WixMadeforText-Bold.woff2"),
        weight: 800,
      },
    ]
  }
])
```

This issue was reported on [canny.io](https://expo.canny.io/feature-requests/p/support-font-files-that-define-multiple-styles-for-a-single-font-family-on-web). It is also a revival of my previous PR #37170 addressing the web font issue, with the suggested improvements to the caller interface.

# How

* Added `FontFaceDefinition` (`path`, `weight`, `style`, `display`, `testString`) and `FontFamilyDefinition` (`fontFamily`, `fontDefinitions`) types. `useFonts`/`loadAsync` now accept `string | Record<string, FontSource> | FontFamilyDefinition[]` – the existing string/map input keeps working unchanged.
* `Font.ts` loads and caches each face independently, keyed by family + weight + style, so multiple faces of the same family aren't silently skipped by the existing per-name "already loaded" dedup (which only ever tracked one file per family name).
* `weight`/`style` default to `400`/`"normal"` when a face doesn't specify one, so every generated `@font-face` rule is unambiguous.
* On web, `_createWebFontTemplate` now emits `font-weight`/`font-style` (sanitized against CSS injection via the existing `CSS_IDENT_RE` check), and the CSS-rule matching/dedup logic (`_matchesFontFaceOptions`) now considers weight/style so each face gets its own `@font-face` rule instead of colliding with others on family name alone.
* On native (iOS/Android), only the first `fontDefinitions` entry for a given family is loaded – there's no native mechanism to select among multiple files for one family name, so this is documented as a platform limitation rather than silently ignored.

# Test Plan

* `et check-packages expo-font` (build, typecheck, depscheck, test, lint, format) passes.
* Added/extended unit tests in `FontHooks-test.ts`, `Font-test.web.ts`, and a new `ExpoFontLoader-test.web.ts`, covering: the array API end-to-end, backward compatibility with the old map/string API, weight/style defaulting, CSS-rule matching, and CSS-injection sanitization.
* Used red/green: verified (via `git stash`) that each new test fails against the pre-fix code before restoring the fix.
* Reproduced the exact scenario from the issue ("Wix Madefor Text" with regular/italic/bold faces) in a real browser engine — jsdom doesn't implement `CSSFontFaceRule`, so this couldn't be verified in Jest alone. Confirmed the three generated `@font-face` strings parse into three distinct `CSSFontFaceRule`s, all with `font-family: "Wix Madefor Text"` but correct, distinct `font-weight`/`font-style` (400/normal, 400/italic, 800/normal).
* Updated the Font SDK reference and Fonts guide with a usage example for the new array API, and regenerated the API data JSON via `et generate-docs-api-data` so the new types show up in the API reference.

# Checklist


amandeepmittal (changes requested):
There is no need to duplicate info on both pages. I'd suggest adding the section to a primary page and then using an internal link to that section from the other page.

Also, please format the prose in the mdx files as they already are.

bryanmylee:
@amandeepmittal I've updated the formatting of the Markdown docs to have no explicit line wrapping and linked the two pages together :)

bryanmylee:
@amandeepmittal Hi Aman, are there any further changes required for this PR? Let me know if there are any issues. I've been using a patch of this successfully on my project.

amandeepmittal (approved):
Thanks for the docs updates!

Looks good to me. Let's wait for the technical review from library authors!

bryanmylee:
@amandeepmittal Any news on this? I've been patching expo-font for a whole year now and would love to have this merged upstream so I don't have to keep patching it after every version upgrade 😅

Please let me know the expected timeline for changes like this and who else I should reach out to.

packages/expo-font/src/Font.ts
```diff
@@ -107,6 +136,80 @@ export function loadAsync(
...
+
+async function loadFontFamilyInNamespaceAsync(definition: FontFamilyDefinition): Promise<void> {
+  const { fontFamily, fontDefinitions } = definition;
+
+  if (loadPromises.hasOwnProperty(fontFamily)) {
+    return loadPromises[fontFamily];
+  }
+
+  const promise = Promise.all(
+    fontDefinitions.map((face) => loadFontFaceInNamespaceAsync(fontFamily, face))
```
tsapeta, line 172: [🚨 **critical**] This loads every face on native too, which contradicts the docs ("On Android and iOS, only the first `fontDefinitions` entry for a given `fontFamily` is loaded"). Each face calls `loadSingleFontAsync(fontFamily, asset)` with the same family name. On Android, `FontLoaderModule` calls `ReactFontManager.setTypeface(fontFamilyName, Typeface.NORMAL, typeface)`, so each face overwrites the previous one. On iOS, `loadAsync` unregisters the existing alias and re-registers the new file. Since `Promise.all` runs the downloads concurrently, the last download to finish wins, so the family can render as bold or italic app-wide, nondeterministically. It also downloads N-1 font files that end up unused.

On native platforms, load only the first face (for example `Platform.OS !== 'web' ? fontDefinitions.slice(0, 1) : fontDefinitions`) and consider a `__DEV__` warning when extra faces are dropped.

packages/expo-font/src/Font.ts
```diff
@@ -107,6 +136,80 @@ export function loadAsync(
...
+  return promise;
+}
+
+// Faces of the same `fontFamily` are cached and awaited individually (keyed by weight/style)
+// since native and the JS-level cache in `memory.ts` can only track loaded state per name.
+async function loadFontFaceInNamespaceAsync(
+  fontFamily: string,
+  face: FontFaceDefinition
+): Promise<void> {
+  const cacheKey = `${fontFamily}\0${face.weight ?? ''}\0${face.style ?? ''}`;
```
tsapeta, line 189: [🏗️ **design**] These per-face cache keys are never purged. `unloadFontInNamespaceAsync` calls `purgeFontFamilyFromCache(fontFamily)`, which deletes only `cache[fontFamily]`. The `` `${fontFamily}\0...` `` keys survive. After `unloadAsync(fontFamily)` deletes the CSS rules, a second `loadAsync` with the same array hits `isLoadedInCache(cacheKey)` and returns without re-injecting anything, so the font is gone until a full reload.

Either purge every key with the `` `${fontFamily}\0` `` prefix in `purgeFontFamilyFromCache`, or store face keys in a structure the purge function knows about. A test for unload-then-reload through the array API would lock this in.

packages/expo-font/src/Font.ts
```diff
@@ -107,6 +136,80 @@ export function loadAsync(
...
+  if (loadPromises.hasOwnProperty(fontFamily)) {
+    return loadPromises[fontFamily];
+  }
+
+  const promise = Promise.all(
+    fontDefinitions.map((face) => loadFontFaceInNamespaceAsync(fontFamily, face))
+  ).then(() => {});
+
+  loadPromises[fontFamily] = promise;
+  promise.finally(() => {
```
tsapeta, line 176: [🏗️ **design**] `promise.finally(...)` creates a new derived promise that nobody awaits. When any face fails to load (a plain network error is enough), `promise` rejects, the caller handles the rejection through the returned `promise`, but the derived `.finally` promise also rejects and stays unhandled. That surfaces as an "Uncaught (in promise)" / LogBox warning for every failed load.

Assign the chained promise instead, matching what the caller awaits:

```ts
const tracked = promise.finally(() => {
  delete loadPromises[fontFamily];
});
loadPromises[fontFamily] = tracked;
return tracked;
```

Or use the async-IIFE + `try/finally` pattern that `loadFontInNamespaceAsync` and `loadFontFaceInNamespaceAsync` already use.

packages/expo-font/src/Font.ts
```diff
@@ -107,6 +136,80 @@ export function loadAsync(
...
+// Merges a face's `weight`/`style`/`display`/`testString` onto its `path`, falling back to
+// values already present on `path` when it's a `FontResource`-shaped object. None of these are
+// given a hardcoded default: leaving a property unset lets the generated `@font-face` rule omit
+// that descriptor entirely, which is required for variable fonts (a single file can cover a
+// range of weights/styles; forcing e.g. `font-weight: 400` on it would incorrectly restrict the
+// face to only that one weight).
+function fontSourceFromFace(face: FontFaceDefinition): FontSource {
+  const { path, weight, style, display, testString } = face;
+
+  if (path instanceof Asset) {
```
tsapeta, line 148: [🏗️ **design**] The early return for `Asset` drops the face's `weight`, `style`, `display`, and `testString`. `path` is typed as `FontSource`, so `{ path: Asset.fromModule(...), weight: 700 }` is valid input, but on web the merged resource loses the weight. The generated rule then has no `font-weight`, and worse: because the rule-matching in `ExpoFontLoader.web.ts` now compares weight/style, a second `Asset`-backed face matches the first face's rule (same family, no weight/style on either side) and is silently skipped.

On web you can safely spread the face options over `{ uri: path.uri ?? path.localUri }`; the `Asset` passthrough is only required for the native loader. Alternatively, reject or document that `Asset` paths cannot carry per-face options.

packages/expo-font/src/ExpoFontLoader.web.ts
```diff
@@ -184,14 +202,31 @@ function getStyleElement(): HTMLStyleElement {
...
+    `src:url(${JSON.stringify(resource.uri)})`,
+  ];
+
+  if (typeof resource.display === 'string' && CSS_IDENT_RE.test(resource.display)) {
+    declarations.push(`font-display:${resource.display}`);
+  }
+
+  if (typeof resource.weight === 'number' && Number.isFinite(resource.weight)) {
+    declarations.push(`font-weight:${resource.weight}`);
+  } else if (typeof resource.weight === 'string' && CSS_IDENT_RE.test(resource.weight)) {
```
tsapeta, line 221: [💡 suggestion] `CSS_IDENT_RE` (`/^[a-zA-Z_-][\w-]*$/`) rejects any string starting with a digit, so `weight: '700'` is silently dropped, and so is a variable-font range like `'100 900'` (contains a space). Both are values the `number | string` type invites, and `'100 900'` is the canonical way to declare a variable font's weight range, which this PR otherwise takes care to support.

Use a weight-specific validation instead, for example accept `/^\d{1,4}(\s+\d{1,4})?$/` plus the CSS keywords (`normal`, `bold`, `bolder`, `lighter`), and keep `CSS_IDENT_RE` for `display`/`style` only.

packages/expo-font/src/Font.ts
```diff
@@ -107,6 +136,80 @@ export function loadAsync(
...
+
+  if (loadPromises.hasOwnProperty(cacheKey)) {
+    return loadPromises[cacheKey];
+  }
+
+  const asset = getAssetForSource(fontSourceFromFace(face));
+  loadPromises[cacheKey] = (async () => {
+    try {
+      await loadSingleFontAsync(fontFamily, asset);
+      markLoaded(fontFamily);
```
tsapeta, line 203: [💡 suggestion] Marking the whole family loaded when the first face finishes makes `isLoaded(fontFamily)` return `true` while the remaining faces are still loading. A component that mounts in that window gets `true` from `isMapLoaded` in `FontHooks.ts` and renders with a missing bold/italic face (the browser falls back or synthesizes it). Consider marking the family only after `Promise.all` in `loadFontFamilyInNamespaceAsync` resolves, and keeping just `markLoaded(cacheKey)` here.

packages/expo-font/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### 🎉 New features
 
 - [web] Added `Server.withServerContext(callback)` which scopes server-side font loads per render. ([#46669](https://github.com/expo/expo/pull/46669) by [@hassankhan](https://github.com/hassankhan))
+- Support loading multiple font weights and styles per `fontFamily` in `useFonts`/`loadAsync`, and generate correct `font-weight`/`font-style` `@font-face` rules on web. (by [@bryanmylee](https://github.com/bryanmylee))
```
tsapeta, line 12: [🐧 nit] Changelog entries in this repo link the PR: `([#47939](https://github.com/expo/expo/pull/47939) by [@bryanmylee](https://github.com/bryanmylee))`.

tsapeta (changes requested):
Thanks for the contribution! I did the first pass of AI review. I discarded a few comments that seemed wrong to me, but let me know if you have any concerns or questions!

packages/expo-font/src/Font.ts
bryanmylee: I've added a warning in DEV mode and selected only the first definition on iOS and Android!

packages/expo-font/src/Font.ts
bryanmylee: I've added a fix to purge all keys with the same font-family prefix.

packages/expo-font/src/Font.ts
bryanmylee: I've updated to track the promise with `.finally()` chained.

packages/expo-font/src/Font.ts
bryanmylee: Added a branch to handle this case!

bryanmylee:
The review makes sense, I've updated the implementation. Let me know if there are any other edge cases I might have missed.

bryanmylee:
@tsapeta Hey team, let me know what I need to do to push this forward :)

tsapeta:
@expo-bot review

vonovak:
@expo-bot investigate why there's no review

brentvatne:
@expo-bot check

vonovak:
@bryanmylee thanks for your time and effort so far, I'm going to take it over the finish line, thanks!

vonovak:
@expo-bot check

PR #47941, opened 2026-07-19
fix(jsi): use UTF-8 byte lengths for property names

Unmesh100:
Fixes non-ASCII JavaScript property names in the iOS JSI layer by passing UTF-8 byte lengths instead
  of Swift grapheme counts. Also corrects array property reads that constructed PropNameIDs with the
  wrong length.


brentvatne:
Closing — UTF-8 property-name lengths landed in https://github.com/expo/expo/pull/48329. Thanks!

PR #47942, opened 2026-07-19
fix(tools): interpolate npm owner validation responses

Unmesh100:
Improves validate-npm-owners error output by interpolating malformed registry responses, making
  validation failures easier to diagnose.


tsapeta:
Good catch, thanks for opening the PR! I'm going to close it in favor of #48295 – it turned out that this command no longer works reliably, so we had to rewrite it.

PR #47944, opened 2026-07-19
[fingerprint] Skip prebuild-generated native dirs for CNG platforms

### LINKED ISSUE #47943: [@expo/fingerprint] Bare native dir sources hashed for CNG projects, ignoring useCNGForPlatforms — fingerprint diverges pre/post-prebuild
### Summary

For a **CNG project** (`ios/` and `android/` are gitignored and regenerated by `expo prebuild`), `@expo/fingerprint` includes the **generated** native directories as `bareNativeDir` sources when they happen to be present on disk. Because those directories only exist *after* a prebuild, the computed fingerprint **diverges depending on whether prebuild has run**:

- Managed fingerprint (e.g. `eas update`, or an EAS Workflows `fingerprint` job — no native dirs present) → excludes the native dirs.
- Build-time fingerprint (computed *after* `expo prebuild` on EAS Build) → includes them.

The two hashes differ for the same commit, which breaks `runtimeVersion: { policy: "fingerprint" }`.

The `useCNGForPlatforms` option (added in #37204) is populated for CNG projects, but the bare-native-dir sourcers in `sourcer/Bare.ts` don't consult it, so the generated directories are hashed anyway.

### Environment

- `@expo/fingerprint` `0.20.5` (Expo SDK 57). Same behavior observed on the SDK 55 line (see #46817).
- Workflow: CNG / managed (no committed `ios/` or `android/`), `runtimeVersion: { policy: "fingerprint" }`.

```bash
npx create-expo-app@latest fp-cng-repro
cd fp-cng-repro

# 1) Fingerprint the pure CNG project (no native dirs on disk)
node -e "require('@expo/fingerprint').createFingerprintAsync('.').then(f => console.log('no-prebuild :', f.hash))"

# 2) Generate the native dirs — the project is still CNG; these are derived/gitignored
npx expo prebuild

# 3) Fingerprint again
node -e "require('@expo/fingerprint').createFingerprintAsync('.').then(f => console.log('post-prebuild:', f.hash, '| bareNativeDir sources:', f.sources.filter(s => JSON.stringify(s).includes('bareNativeDir')).length))"
```

### Current behavior

Reproduced on a **stock `create-expo-app`** (Expo SDK 57, `@expo/fingerprint@0.20.5`):

```
no-prebuild  : 55a61c85ef0f1b4682dc571a747835dac4b2b0a8   bareNativeDir sources: 0
post-prebuild: ede8de396e545e43827de7ce433d5cdc1746d543   bareNativeDir sources: 2   (android, ios)
```

The template's `.gitignore` ignores `/ios` and `/android`, so the project is still CNG after prebuild — the generated directories should not contribute to the fingerprint, yet running `expo prebuild` alone changes the hash. So `eas update` (managed fingerprint) and an EAS Build (fingerprint computed after prebuild) produce different runtime versions for the same commit.

### Expected behavior

For a CNG project, the fingerprint should be **identical whether or not `expo prebuild` has run** — the generated `ios/`/`android/` directories are derived from config and must not contribute to the hash. In other words, `getBareAndroidSourcesAsync` / `getBareIosSourcesAsync` should skip the bare-native-dir source when `useCNGForPlatforms[platform]` is `true`.

### Root cause

In [`packages/@expo/fingerprint/src/sourcer/Bare.ts`](https://github.com/expo/expo/blob/main/packages/%40expo/fingerprint/src/sourcer/Bare.ts), `getBareAndroidSourcesAsync` and `getBareIosSourcesAsync` gate the `bareNativeDir` source only on `options.platforms.includes(platform)`, without checking `options.useCNGForPlatforms[platform]`:

```ts
async function getBareAndroidSourcesAsync(projectRoot, options) {
  if (options.platforms.includes('android')) {            // <- also needs: && !options.useCNGForPlatforms.android
    const result = await getFileBasedHashSourceAsync(projectRoot, 'android', 'bareNativeDir');
    ...
  }
}
```

`useCNGForPlatforms` was introduced in #37204 to signal project workflow, but these two sourcers don't honor it.

### Impact

`runtimeVersion: { policy: "fingerprint" }` on a CNG project can produce a runtime-version mismatch between `eas update` / Workflows `fingerprint` jobs and EAS Build (which fingerprints post-prebuild), so update targeting and `type: repack` build reuse break on every native-affecting build. This is the concrete, reproducible version of #46817 (which was closed for a missing repro).

Oupsla:
# Why

For a **CNG (managed)** project, `@expo/fingerprint` includes the prebuild-generated `ios/` and `android/` directories as `bareNativeDir` sources whenever they happen to exist on disk. Because those directories only exist *after* `expo prebuild`, the fingerprint diverges depending on whether prebuild has run:

- Managed fingerprint (`eas update`, a Workflows `fingerprint` job — no native dirs present) → excludes them.
- Build-time fingerprint (computed *after* `expo prebuild` on EAS Build) → includes them.

The two hashes differ for the same commit, which breaks `runtimeVersion: { policy: "fingerprint" }` (update targeting + `type: repack` reuse).

Fixes #47943 — which has a full repro on a stock `create-expo-app`.

# How

`getBareAndroidSourcesAsync` / `getBareIosSourcesAsync` gated the `bareNativeDir` source only on `options.platforms.includes(platform)`. This adds `&& !options.useCNGForPlatforms[platform]`, so the generated native dirs are skipped for CNG platforms.

This is consistent with `normalizeOptionsAsync`, which already appends `android/**/*` / `ios/**/*` to the ignore paths for CNG platforms — it neutralized the dir *contents* but the dir *source entry* remained, which is the leak fixed here.

# Test Plan

Added a unit test in `Bare-test.ts`: with the project workflow mocked to `managed`, `getBareAndroidSourcesAsync` / `getBareIosSourcesAsync` now return `[]`; the existing bare-workflow test still asserts the dirs are included.

Verified end-to-end on a fresh `create-expo-app` (Expo SDK 57, `@expo/fingerprint@0.20.5`), `/ios` and `/android` gitignored (so still CNG):

```
before prebuild : 55a61c85…   bareNativeDir sources: 0
after  prebuild : ede8de39…   bareNativeDir sources: 2   (android, ios)
```

With this change the pre/post-prebuild fingerprints match.


Oupsla:
Superseded by #47949 — this PR's branch was accidentally force-pushed to a broken (parentless) commit, which blew the diff up to the whole repo. The clean, single-commit version of the same fix lives in #47949.

PR #46280, opened 2026-05-26
[file-system] Make FileHandle.readBytes()/writeBytes async

wh201906:
# Why

Similar to #44359 and #45992, async file IO makes the UI more responsive

# How

1. Replace `FileHandle.readBytes()` and `FileHandle.writeBytes()` with async definitions in both TS and native code
2. Update downstream calls
3. Update test cases

# Test Plan

1. `CI=1 pnpm build`, `pnpm lint --fix` and `CI=1 pnpm test` passed in `packages/expo-file-system`

<details><summary>Output</summary>

```
> expo-module test

PASS Android src/__tests__/FSNetworkTasks-test.native.ts
PASS Android src/__tests__/FileSystem-test.native.ts
PASS Node src/__tests__/FileSystemWatcher-test.ts
PASS Web src/__tests__/FileSystemWatcher-test.ts
PASS Android src/__tests__/FileSystemWatcher-test.ts
PASS Android src/legacy/__tests__/FileSystem-test.native.ts
PASS iOS src/__tests__/FileSystem-test.native.ts
PASS iOS src/__tests__/FSNetworkTasks-test.native.ts
PASS iOS src/__tests__/FileSystemWatcher-test.ts
PASS iOS src/legacy/__tests__/FileSystem-test.native.ts

Test Suites: 10 passed, 10 total
Tests:       14 skipped, 196 passed, 210 total
Snapshots:   0 total
Time:        2.021 s
Ran all test suites in 4 projects.
```

</details> 

2. `pnpm et check expo-file-system` passed in project root (tested on 91dbfb47c8d6f31ae2a86757f32616275d567b92)

<details><summary>Output</summary>

```
 🛠  Rebuilding expotools
 ✨ Successfully built expotools

🔍 Checking expo-file-system package
🏃‍♀️ Running pnpm run clean
🏃‍♀️ Running pnpm run build
🏃‍♀️ Running pnpm run test --watch false --passWithNoTests
🏃‍♀️ Running pnpm run lint --max-warnings 0
✨ expo-file-system checks passed
🔌 Checking expo-file-system plugin
🏃‍♀️ Running pnpm run clean plugin
🏃‍♀️ Running pnpm run build plugin
🏃‍♀️ Running pnpm run test plugin --watch false --passWithNoTests
🏃‍♀️ Running pnpm run lint plugin --max-warnings 0
✨ expo-file-system checks passed

🏁 All packages passed.
```

</details> 

[...]

PR #47945, opened 2026-07-19
[file-system] Apply Dispatchers.IO to FileHandle and sync FileHandle access with lock

wh201906:
# Why

Like #46376, `withContext(Dispatchers.IO)` ensures other unrelated async calls won't be blocked by `FileHandle.readBytes()` and `FileHandle.writeBytes()`, makes it possible to read/write multiple file at the same time on Android.

This PR also fixes a problem introduced in #46280. It's possible that a user accidentally changes the offset when an async read/write is not finished like this:
```ts
const p = handle.writeBytes(...); // This is async after #46280
while (handle.size < 1024 * 1024 * 1024) { // For iOS, accessing handle.size calls handle.seekToEnd() implicitly
    ...
}
```
Rare but possible, a user can get unexpected result on Android with such code:
```ts
const p = handle.readBytes(...); // on Android, this is performed with one or more fileChannel.read() calls.
handle.offset = newOffset;
const result = await p;
```
With the `withContext(Dispatchers.IO)` introduced in this PR, I thinks this problem is more likely on Android because multiple async read/writes may execute concurrently instead of being serialized with the Expo modules queue.
8b16c7371de3cbc495d8e840749e0166d3bb726a fixes this by wrapping most of the operations with a lock on both Android and iOS.

# How

1. Use Dispatchers.IO for FileHandle
2. Add locks for concurrent accesses on both Android and iOS

# Test Plan

1. `pnpm et check expo-file-system` passed
2. Filesystem related tests in `apps/bare-expo` passed on Android

<details><summary>Screenshot</summary>

image

</details>

# Checklist


packages/expo-file-system/ios/FileSystemFileHandle.swift
```diff
@@ -29,6 +29,7 @@ internal final class FileSystemFileHandle: SharedRef<FileHandle> {
   let handle: FileHandle
   private let didAccessSecurityScope: Bool
   private var isClosed = false
+  private let lock = NSLock()
```
barthap: NSLock is non-reentrant — I checked the call paths and no locked method calls another locked method, so we're fine. Just something to keep in mind if anyone adds a helper later that takes the lock, perhaps worth adding a code comment.

packages/expo-file-system/CHANGELOG.md
```diff
@@ -14,11 +14,13 @@
 ### 🐛 Bug fixes
 
 - Fixed `FileHandle` security-scoped access, and non-SAF `content://` URI support. ([#47176](https://github.com/expo/expo/pull/47176) by [@barthap](https://github.com/barthap))
+- Fixed potential file offset races when asynchronous and synchronous `FileHandle` operations overlap on Android and iOS. ([#47945](https://github.com/expo/expo/pull/47945) by [@wh201906](https://github.com/wh201906))
 
 ### 💡 Others
 
 - Improve read/write performance on Android by applying `withContext(Dispatchers.IO)` when possible. ([#46376](https://github.com/expo/expo/pull/46376) by [@wh201906](https://github.com/wh201906))
```
barthap:
```suggestion
- Improve `File` read/write performance on Android by applying `withContext(Dispatchers.IO)` when possible. ([#46376](https://github.com/expo/expo/pull/46376) by [@wh201906](https://github.com/wh201906))
```
to distinguish previous PR from the crrent one

or we can merge these two changelog entries into one:

```suggestion
- Improve read/write performance on Android by applying `withContext(Dispatchers.IO)` when possible. ([#46376](https://github.com/expo/expo/pull/46376), ([#47945](https://github.com/expo/expo/pull/47945) by [@wh201906](https://github.com/wh201906))
```

barthap (commented):
Nice fix 👍 it indeed fixes a real race that already exists on main, and the lock is the correct fix. Two things before merge:

1. The lock gives us atomicity but not ordering. `synchronized` and `NSLock` aren't FIFO, and before #46280 the modules queue serialized same-handle ops for free - `Dispatchers.IO` removes that. So two unawaited `writeBytes()` on the same handle can still land in either order. Neither corrupts, but the order isn't guaranteed. Could you add a docs note that operations on the same handle should be awaited?
2. ~The Test Plan is empty~. A concurrency test-suite test would help but it might be tricky - e.g. kick off an async write, mutate `offset` / do a sync read before it resolves, and assert the file ends up consistent.

packages/expo-file-system/ios/FileSystemFileHandle.swift
wh201906: Fixed in 07ae2dc2a49e2c67e7c19307f48983f4e07f9666. I'm not sure if I comment it concisely.

packages/expo-file-system/CHANGELOG.md
wh201906: I prefer the merged one. Fixed in ab830f14d300df112ec61135304000642331cc70

wh201906:
Thanks for your review!

> Could you add a docs note that operations on the same handle should be awaited?

I've added a comment for it in 07ae2dc2a49e2c67e7c19307f48983f4e07f9666 and updated the generated doc data.

> A concurrency test-suite test would help but it might be tricky

I tried some tests like `async write` -> `handle.offset = x` -> `sync read` on Android, but the `async write` is always finished before the sync operations. I also tried writing 64MB of data to prolong it and make it overlapped with the sync operations, but I got OOM error then.

barthap:
Thanks for adding the docs note; it looks good.

On the test, you're running into the reason these are painful. The async write finishes before your sync calls even start, so there's no overlap to catch, and forcing overlap with a giant buffer just runs you out of memory. A JS-level test would depend on thread timing, so it'd be flaky and wouldn't reliably fail on the pre-fix version anyway. I don't think it's worth chasing.

I'll approve once CI passes and merge later.

wh201906:
The same CI failed again. Maybe merging the latest `main` branch helps.

barthap:
No, this is fine. macOS is failing on main too.

barthap (approved):

wh201906:
Thanks!

PR #47947, opened 2026-07-19
[ios][file-system] Fix preview reopening during dismissal

### LINKED ISSUE #47946: [expo-file-system][iOS] File.preview() rejects when reopened during Quick Look dismissal

https://github.com/eliotgevers/expo-file-preview-reopen-repro

## Steps to reproduce

Platform: iOS 26.5 (iPhone 17 Pro Max and iPad Pro 13-inch Simulators)

Environment: Expo development build using `expo-file-system@57.0.0-canary-20260718-7cbcda5`, built from the current `expo/expo` `main` commit at the time of filing

The minimal reproduction uses the current canary.

Package manager: Bun 1.3.13

1. Install and run the minimal reproduction on iOS:

   ```sh
   bun install
   bun run ios
   ```

2. Tap **Preview file**.
3. Close Quick Look.
4. As soon as the app is visible and interactive, tap **Preview file** again.

Expected: the second preview opens after the in-progress Quick Look dismissal finishes.

Actual: the second call rejects with `ERR_FILE_PREVIEW_IN_PROGRESS` / `File preview is already in progress`.

This is specifically a dismissal-transition race, not a request to allow two previews concurrently. A second call while Quick Look remains open should continue to reject.

The iOS module keeps `isPresentingPreview` set until `previewControllerDidDismiss`. Instrumenting the native code showed that the app can receive the second tap before that callback:

```text
22:03:21.614 previewControllerWillDismiss
22:03:22.463 preview request active=true beingDismissed=true presenting=true window=true coordinator=true
22:03:22.699 previewControllerDidDismiss
```

At the rejected request, the existing `QLPreviewController` reports `isBeingDismissed == true` and has an active transition coordinator. The delegate's `previewControllerDidDismiss` callback arrives about 236 ms later.

The public promise cannot be used to guard this at the app level: by design, `File.preview()` resolves when the viewer is presented rather than when it is dismissed.

I reproduced the failure deterministically with a 500 ms close-to-reopen sequence. Waiting about one second avoids it. The same sequence passes on both iPhone and iPad when one request made during the active dismissal transition is deferred until `previewControllerDidDismiss`.

Source inspection indicates this is iOS-specific. Android hands the file to an `ACTION_VIEW` intent and does not maintain the same in-process Quick Look presentation state. I have only runtime-tested the issue on the iOS 26.5 Simulator; a physical device and older iOS runtime would still be useful confirmation.

## Environment

```text
  System:
    OS: macOS 26.5.1
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 25.5.0 - /opt/homebrew/bin/node
    npm: 11.8.0 - /opt/homebrew/bin/npm
  Managers:
  SDKs:
    iOS SDK:
  IDEs:
    Android Studio: 2025.1 AI-251.26094.121.2513.14007798
  npmPackages:
    expo: 57.0.0-canary-20260718-7cbcda5 => 57.0.0-canary-20260718-7cbcda5
    react: 19.2.3 => 19.2.3
  npmGlobalPackages:
    eas-cli: 20.5.1
  Expo Workflow: managed
```

## Expo Doctor Diagnostics

```text
```

eliotgevers:
# Why

Fixes #47946.

On iOS, `File.preview()` keeps its presentation guard set until Quick Look calls `previewControllerDidDismiss`. During an animated dismissal, however, the underlying app can already be visible and interactive before that delegate callback arrives. A user who taps the preview action in this short window receives `ERR_FILE_PREVIEW_IN_PROGRESS`, even though the previous preview is actively closing.

Apps cannot reliably cover the full viewer lifetime with the returned promise because `preview()` intentionally resolves when the platform preview is presented, not when the user dismisses it.

# How

- Retain a weak reference to the active `QLPreviewController`.
- Continue rejecting true concurrent calls while Quick Look is open.
- If and only if the active controller is being dismissed and has an active transition coordinator, retain one deferred preview request.
- Present that request from `previewControllerDidDismiss`, after UIKit confirms the old viewer has closed.
- Reject the deferred request if an interactive dismissal is cancelled, and reject additional requests while one is already deferred.

This uses UIKit lifecycle state rather than an arbitrary delay. It does not change Android behavior or the public promise semantics.

I also tested two smaller-looking alternatives:

- Clearing the guard as soon as `isBeingDismissed` becomes true allows a new presentation before the old controller has actually finished closing; the second preview did not reliably appear.
- Presenting from the transition coordinator's completion happened to work on iOS 26.5, but it depends on its ordering relative to Quick Look's delegate cleanup. `previewControllerDidDismiss` is Quick Look's documented signal that the preview closed, so it remains the authoritative handoff point.

For an interactive dismissal, `notifyWhenInteractionChanges` is used only to reject and clear the deferred request if UIKit reports that the transition was cancelled.

# Test Plan

Minimal reproduction:

1. Run https://github.com/eliotgevers/expo-file-preview-reopen-repro on iOS.
2. Tap **Preview file**.
3. Close Quick Look.
4. As soon as the app is visible, tap **Preview file** again.

Before this change, a repeatable 500 ms close-to-reopen sequence rejects with `ERR_FILE_PREVIEW_IN_PROGRESS`. Native instrumentation captured the second request while `isBeingDismissed == true` with an active transition coordinator, about 236 ms before `previewControllerDidDismiss`.

After this change:

- Built and runtime-tested the exact patch from source against `expo-file-system@57.0.0-canary-20260718-7cbcda5`. Before opening the PR, rebased the same change onto `expo/expo` `main` at `4e87c54`; the affected preview implementation was unchanged.
- Repeated the exact 500 ms close-to-reopen sequence on an iPhone 17 Pro Max Simulator running iOS 26.5; the unpatched current canary rejected, while the patched build reopened Quick Look. The patched sequence also passed repeated runs.
- Built `expo-file-system` from the patched Swift source for an iPad Pro 13-inch Simulator running iOS 26.5 and confirmed the same immediate close/reopen sequence opened `Preview attempt 2`.
- Ran a scheduled second call while Quick Look remained fully open and confirmed it still rejects with `FilePreviewInProgressException`.
- Verified that the state guard only permits one deferred request during a dismissal.
- `npx tsc --noEmit` passes in the minimal reproduction.
- `npx expo-doctor@latest` passes all 20 checks in the minimal reproduction.
- `git diff --check` passes.

The existing Native Component List file-system screen can also exercise local PDF, image, and text files. No JavaScript unit test is added because the failure depends on the UIKit dismissal lifecycle; the JavaScript test suite mocks the native `preview()` method and cannot observe this transition.

Runtime coverage in this PR is currently limited to iPhone and iPad Simulators on iOS 26.5. Testing a physical device and an older supported iOS runtime before merge would add confidence. Android is unchanged and uses `ACTION_VIEW` rather than `QLPreviewController`.

# Checklist


packages/expo-file-system/ios/FileSystemModule.swift
```diff
@@ -59,6 +61,82 @@ public final class FileSystemModule: Module {
...
+    promise: Promise
+  ) -> Bool {
+    guard deferredPreview == nil,
+      let previewController,
+      previewController.isBeingDismissed,
+      let transitionCoordinator = previewController.transitionCoordinator else {
+      return false
+    }
+
+    deferredPreview = (file, options, promise)
```
barthap, line 125: Once a request is stored here, it only settles through one of two paths: `previewControllerDidDismiss` (which presents it) or the interactive-cancel handler below (which rejects it). I might be wrong, but if the active controller goes away without `previewControllerDidDismiss` firing — for example the presenting view controller stack is torn down mid-animation — it seems `onFinish` never runs and this deferred promise never resolves or rejects, so the JS caller would hang.

Quick Look normally does call the delegate on dismissal, so this is likely an edge case. But unlike the original `preview()` promise, which resolves at presentation time, the deferred one has a window where it might never settle. I'm wondering if it's worth rejecting and clearing `deferredPreview` on module deinit (or `OnDestroy`) as a backstop, so a torn-down preview can't leave the promise pending. Curious what you think.

barthap (approved):
Nice fix. Deferring the request and handing off at `previewControllerDidDismiss`, instead of clearing the guard early, is the right approach, and the interactive-cancel handling covers the tricky case. All the settlement paths and the state transitions are consistent, and seem correct.

One non-blocking thought inline about the deferred promise's lifetime when the preview is torn down without the dismiss callback. Worth considering as a follow-up. I won't block the PR on that, unless you want to address it here directly.

---

P.S. macOS CI failure unrelated, the `#if os(iOS)` guards are correct

packages/expo-file-system/ios/FileSystemModule.swift
eliotgevers: Good catch, thanks! I added an iOS `OnDestroy` backstop and centralized the rejection path so interactive cancellation and module teardown can settle the deferred promise at most once. `previewControllerDidDismiss` remains the normal success path. I also source-built the exact change and tested root-view-controller replacement after deferral. Quick Look still delivered `didDismiss`, while explicit module unregistration fired `OnDestroy` and rejected the pending promise.

Issue #47024, opened 2026-06-18
expo-location: foreground service notification does not reappear after swipe-close on Samsung OneUI (mStopped never reset in onStartCommand)


MarkusAbtion:
## Summary

On Android devices where the process survives task removal (Samsung OneUI, OneUI 6+), the foreground GPS notification does not reappear when the user reopens the app. `startLocationUpdatesAsync` resolves without error, but the notification is silently not shown.

**Minimal reproducible example:** https://github.com/MarkusAbtion/expo-location-mre

## Steps to reproduce

1. Clone and run the MRE:
   ```bash
   git clone https://github.com/MarkusAbtion/expo-location-mre
   cd expo-location-mre
   npm install
   npx expo run:android
   ```
2. Grant **location (always)** permission, tap **Start tracking** — notification appears.
3. Open Recent Apps and **swipe the app closed** (do NOT force-stop).
4. Reopen the app from the launcher.
5. Observe: notification does **not** reappear, despite the app calling `startLocationUpdatesAsync` (visible in the on-screen log).

> **Note:** Reproduction requires a Samsung Galaxy device (OneUI 6+) or any Android device that keeps the process alive after task removal. On stock Android (Pixel, emulator), the process is killed on swipe and the bug does not manifest.

## Logcat evidence

```bash
# Swipe + reopen on Samsung — buggy path (setOptions, no startForegroundService):
adb logcat | grep -E "LocationTaskConsumer|setOptions|didRegister|maybeStartForegroundService"
```

On the buggy path: `setOptions` is called, but `startForegroundService` and `onStartCommand` never fire.

On a fresh start (force-stop + restart): `didRegister` is called, `startForegroundService` fires, notification appears correctly.

## Expected behavior

The foreground notification reappears every time the user reopens the app and `startLocationUpdatesAsync` is called, regardless of whether the process survived the swipe.

## Observed behavior

After swipe-close on Samsung OneUI, reopening the app and calling `startLocationUpdatesAsync` resolves silently with no notification shown. GPS updates continue in the background (the task is still registered), but there is no visible foreground service notification and the service is technically in a stopped state.

## Root cause

The bug is in `LocationTaskService.kt` (compiled into the `expo-location` AAR).

**Call chain that sets `mStopped = true`:**
1. User swipes app closed → Android calls `onTaskRemoved()` on `LocationTaskService`
2. `onTaskRemoved()` calls `stop()`, which calls `stopForeground()` and sets `mStopped = true`

[...]

PR #47958, opened 2026-07-20
[android][task-manager] Clear headless task manager on context destroy

### LINKED ISSUE #47673: Expo Location + Expo Task Manager -- Location updates in task can fail to restart in certain circumstances

https://github.com/realtux/mre-bglocation

### Steps to reproduce

I'm encountering an issue which I strongly believe is a bug in either expo location or expo task manager. I've prepared a minimum reproducible example along with detailed instructions to reproduce the problem.

Bug Summary: An Expo app can enter a state where location updates will no longer be delivered to the task manager task. This issue will persist through app restarts. This appears to be an Android only issue.

This state can be reproduced in one guaranteed way and one eventual way. What both of these methods have in common is the OS is forcibly killing the task:
1. Guaranteed: force quit the app any time after calling `Location.startLocationUpdatesAsync`
2. Eventually: leave app in recent list or terminated for several hours until the OS kills it

My reproducible example is very basic, it has a Start button and Stop button, the former calls `Location.startLocationUpdatesAsync` and the latter calls `Location.stopLocationUpdatesAsync`. A one line task is defined in `index.js` which just logs location updates to the console.

Platforms Affected:
- Android

Environments Affected:
- Development Client
- Production Build

Steps To Reproduce:
1. Open app.
2. Press Start. You will observe location updates starting as expected.
3. Go into app info and force quit the application. At this point, the app state is corrupted.
4. Wait a minute or so. This step is very important because location updates are still accumulating somehow (which you'll see in Step 6).
5. Open app (dev client client only). You will observe the location icon appearing briefly in the notification bar then the app crashes near instantly.
6. Open app. App will start normally. Once it starts you will observe a batch of location updates come in (before even pressing Start). These updates are unexpected. The presence of this location updates batch is what perpetuates this bug forever. Any time these updates appear, the app will not function normally.
7. Press Start. You will observe no location updates starting. You can press Stop and Start over and over and it won't resolve. However, after pressing Start, location updates are accumulating somewhere in the system (same as Step 4).
8. Close or Restart app. You will observe the same behavior as Step 6/7 over and over.

How to clear the corrupted state and return the app to normal operation:
Option 1: Force quit the app, clear the storage cache and storage data for the app, and restart.
Option 2: Within Step 6., rapidly restart the app after the batch of updates show up either with the built in Expo restart button or `BackHandler.exitApp()`. This sometimes fails, depending on whether any location updates were accumulated since the app opened.

I wouldn't mind investigating the native side of this myself, I've so far been unable to run a forked version of expo-location in my project for debugging.

### Environment

```text
    System:
      OS: Linux 6.8 Ubuntu 24.04.4 LTS 24.04.4 LTS (Noble Numbat)
      Shell: 5.2.21 - /bin/bash
    Binaries:
      Node: 20.19.4 - /home/brian/.nvm/versions/node/v20.19.4/bin/node
      Yarn: 1.22.22 - /home/brian/.nvm/versions/node/v20.19.4/bin/yarn
      npm: 10.8.2 - /home/brian/.nvm/versions/node/v20.19.4/bin/npm
    npmPackages:
      expo: ~57.0.4 => 57.0.4 
      react: 19.2.3 => 19.2.3 
    npmGlobalPackages:
      eas-cli: 16.19.2
    Expo Workflow: managed
```

### Expo Doctor Diagnostics

```text
```

Wenszel:
# Why
Fixes: [#47673](https://github.com/expo/expo/issues/47673), https://github.com/expo/expo/issues/47024

There was a problem in `expo-location` which occurred after force-stop on stock Android ([#47673](https://github.com/expo/expo/issues/47673)) and on swipe-close on custom Android ([#47673](https://github.com/expo/expo/issues/47673)) . After going back to the app the event queue was emptied, and since then no other events were handled

It turned out to be an `expo-task-manager` problem: it removes the `headlessTaskManager` after `invalidateApp`, which would be correct if the context weren't alive (then the app would just be reloaded on the next event). But because the context is still alive, no new task manager is created (that only happens on module create / onCreate), and `getTaskManager()` returns null

# How
Moved removing the `headlessTaskManager` from after invalidating the app to `onDestroy` - `setTaskManager(null)` now clears both maps, so the headless manager lives as long as its context, same as the foreground one

There was also an alternative where the `headlessTaskManager` could be moved to `sTaskManagers` when the activity takes it over, but moving it to `onDestroy` is better since it follows the same flow as the regular `taskManager`

# Test Plan 

Tested on BareExpo Playground

### DIFF
--- a/packages/expo-task-manager/CHANGELOG.md
+++ b/packages/expo-task-manager/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### 🐛 Bug fixes
 
 - [iOS] Fix a crash when a task execution request is evaluated re-entrantly, by making its completion callback fire exactly once. ([#47594](https://github.com/expo/expo/pull/47594) by [@tsapeta](https://github.com/tsapeta))
+- [Android] Clear headless task manager on context destroy ([#47958](https://github.com/expo/expo/pull/47958) by [@Wenszel](https://github.com/Wenszel))
 
 ### 💡 Others
 
--- a/packages/expo-task-manager/android/src/main/java/expo/modules/taskManager/TaskService.java
+++ b/packages/expo-task-manager/android/src/main/java/expo/modules/taskManager/TaskService.java
@@ -242,6 +242,7 @@ public class TaskService implements SingletonModule, TaskServiceInterface {
     // It may be called with null when the host activity is destroyed.
     if (taskManager == null) {
       sTaskManagers.remove(appScopeKey);
+      sHeadlessTaskManagers.remove(appScopeKey);
       return;
     }
 
@@ -611,9 +612,7 @@ public class TaskService implements SingletonModule, TaskServiceInterface {
   private void invalidateAppRecord(String appScopeKey) {
     HeadlessAppLoader appLoader = getAppLoader();
     if (appLoader != null) {
-      if (getAppLoader().invalidateApp(appScopeKey)) {
-        sHeadlessTaskManagers.remove(appScopeKey);
-      }
+      appLoader.invalidateApp(appScopeKey);
     }
   }
 


Wenszel:
* **#47958** <a href="https://app.graphite.com/github/pr/expo/expo/47958?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/47958?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tsapeta (approved):

PR #47961, opened 2026-07-20
[expo-router] refactor preloaded routes logic to comply with standard-navigation interface

Ubax:
# Why

Native stack currently handles preloaded routes separately, as `preloadedRoutes` property. 

For the standard-navigation integration we need a single routes array in state where preloaded routes are placed after the focused index.

# How

The change is split into multiple commits to make it easier to review:

1. state index refactor - there were some `state.index || state.routes.lenght - 1` fallbacks which will not work after the change
2. mechanical refactor to reduce changes in the next commit - adding additional variables and functions 
3. `[expo-router] project preloaded routes in native stack` - main changes
4. mechanical tests updated
5. small follow-up fix

# Test Plan

1. CI
2. Router e2e

# Checklist

- [x] I added a `changelog.md` entry and rebuilt the package sources according to the contributing guide.
- [ ] This diff will work correctly for `npx expo prebuild` & EAS Build.
- [ ] Conforms with the Documentation Writing Style Guide.


packages/expo-router/src/fork/native-stack/NativeStackView.tsx
```diff
@@ -15,15 +15,18 @@ export function NativeStackView(props: ComponentProps<typeof RNNativeStackView>)
 function NativeStackViewInner(props: ComponentProps<typeof RNNativeStackView>) {
   const rootModals = use(RootModalContext);
```
Ubax, line 16: I added a follow-up ticket to remove this - https://linear.app/expo/issue/ENG-25301/remove-rootmodalcontext

packages/expo-router/src/layouts/experimental-stack/ExperimentalStackView.tsx
```diff
@@ -29,28 +28,18 @@ type Props = {
...
   const { setNextDismissedKey } = useDismissedRouteError(state);
   const { preventedRoutes } = usePreventRemoveContext();
 
-  const preloadedDescriptors = state.preloadedRoutes.reduce<ExperimentalStackDescriptorMap>(
-    (acc, route) => {
-      acc[route.key] = acc[route.key] || describe(route, true);
-      return acc;
-    },
-    {}
-  );
```
Ubax, line 45: We don't need preloaded descriptors anymore, because they are now normal descriptors declared for routes after `state.index`

packages/expo-router/src/react-navigation/routers/StackRouter.tsx
```diff
@@ -449,23 +473,23 @@ export function StackRouter(options: StackRouterOptions) {
...
+          );
         }
 
         case 'NAVIGATE_DEPRECATED': {
           if (!state.routeNames.includes(action.payload.name)) {
             return null;
           }
 
+          const getId = options.routeGetIdList[action.payload.name];
+          const id = getId?.({ params: action.payload.params });
```
Ubax, line 499: Moved because they are used in 
```ts
            preloadedRoutes.find(
              (route) =>
                route.name === action.payload.name && id === getId?.({ params: route.params })
            )
```

Ubax:
* **#48114** <a href="https://app.graphite.com/github/pr/expo/expo/48114?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#47961** <a href="https://app.graphite.com/github/pr/expo/expo/47961?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/47961?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#46592** <a href="https://app.graphite.com/github/pr/expo/expo/46592?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#46593](https://github.com/expo/expo/pull/46593) <a href="https://app.graphite.com/github/pr/expo/expo/46593?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#47825** <a href="https://app.graphite.com/github/pr/expo/expo/47825?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2 other dependent PRs ([#46536](https://github.com/expo/expo/pull/46536) <a href="https://app.graphite.com/github/pr/expo/expo/46536?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#47839](https://github.com/expo/expo/pull/47839) <a href="https://app.graphite.com/github/pr/expo/expo/47839?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

packages/expo-router/src/fork/useLinking.ts
```diff
@@ -33,8 +34,8 @@ const findMatchingState = <T extends NavigationState>(
   }
 
   // Tab and drawer will have `history` property, but stack will have history in `routes`
-  const aHistoryLength = a.history ? a.history.length : a.routes.length;
-  const bHistoryLength = b.history ? b.history.length : b.routes.length;
+  const aHistoryLength = getHistoryLength(a);
```
Ubax, line 37: [🚨 **critical**] This fork copy was migrated to `getHistoryLength`, but the vendored copy at `src/react-navigation/native/useLinking.tsx` (lines ~33–34 and ~365) still computes stack history length as `state.routes.length`. That copy ships through the public `NavigationContainer` export (`react-navigation/native/index.tsx`, also `createStaticNavigation`), which now runs against the new StackRouter semantics.

Repro: web app on the vendored `NavigationContainer` with linking; state `[a]`, `preload(B)` → `routes.length` goes 1→2 with no navigation; `navigate(B)` promotes B → `routes.length` stays 2 → `historyDelta === 0` → `history.replace` instead of `push`, so browser Back skips `a`. `findMatchingState` in the same file has the same problem. Both call sites should use `getHistoryLength` like this fork does.

packages/expo-router/src/react-navigation/core/useNavigationCache.tsx
```diff
@@ -61,155 +61,105 @@ export function useNavigationCache<
...
+  cache.current = routes.reduce<NavigationCache<State, ScreenOptions, EventMap>>((acc, route) => {
+    const previous = cache.current[route.key];
+
+    type Thunk = NavigationAction | ((state: State) => NavigationAction | null | undefined);
+
+    if (previous) {
+      // If a cached navigation object already exists, reuse it
+      acc[route.key] = previous;
+    } else {
+      const dispatch = (thunk: Thunk) => {
```
Ubax, line 83: [🏗️ **design**] With the placeholder `base` gone, preloaded routes get a fully functional `dispatch`. Screen dispatches carry `source: route.key` but no `target`, and every source-aware branch in `StackRouter.getStateForAction` is gated on `action.target === state.key && action.source` — so a dispatch from a preloaded screen falls back to `currentIndex = state.index`, i.e. the **focused** route.

Concrete scenario: a screen with `useEffect(() => { if (!authed) navigation.goBack() })` is prefetched via `<Link prefetch>`. Its effects run at preload time (they always did), but previously `dispatch` threw "Actions cannot be dispatched from a placeholder screen" — a loud, debuggable error. Now it silently pops the screen the user is looking at. Same for `replace`, `popTo`, `dismiss`.

If functional navigation on preloaded screens is intended (the descriptor-identity test suggests it is), consider ignoring stack actions whose `source` lies in the preloaded segment (or at least a dev-mode warning), so a prefetch can't mutate the visible stack.

packages/expo-router/src/react-navigation/core/useOnPreventRemove.tsx
```diff
@@ -19,18 +19,29 @@ type Options = {
 
 const VISITED_ROUTE_KEYS = Symbol('VISITED_ROUTE_KEYS');
 
+export const getPreventableRoutes = (
```
Ubax, line 22: [🏗️ **design**] Related to preloaded screens gaining real navigation objects: `usePreventRemove(true)` in a preloaded screen now registers successfully (its key is in `getState().routes`, so `PreventRemoveProvider.setPreventRemove` no longer throws). The nested-provider aggregation in `PreventRemoveProvider` computes `isPrevented` over **all** registered screens including preloaded ones and propagates it to the containing route — so an invisible, never-focused prefetched screen can set `preventNativeDismiss` / disable the back-button menu on the visible enclosing screen.

Meanwhile the preloaded route's own removal correctly bypasses `beforeRemove` (excluded here in `getPreventableRoutes`), so its `preventRemove` state and its actual removability disagree. Consider excluding preloaded routes from the prevented-set aggregation as well.

packages/expo-router/src/modal/web/utils.ts
```diff
@@ -76,23 +76,27 @@ export function convertStackStateToNonModalState(
...
   // if that route was filtered out – at the last remaining route.
-  let index = routes.findIndex((r) => r.key === state.routes[state.index]?.key);
+  let index = routes.findIndex((r) => r.key === activeRoutes[state.index]?.key);
   if (index < 0) {
     index = routes.length > 0 ? routes.length - 1 : 0;
   }
 
-  return { routes, index };
+  // Preloaded routes start after the active index, so they cannot exist when no active routes remain.
+  return { routes: routes.length > 0 ? routes.concat(preloadedRoutes) : routes, index };
```
Ubax: [🏗️ **design**] The comment doesn't hold: `routes` is empty whenever every **active** route is modal-presented (all filtered out above) while `preloadedRoutes` is non-empty — active routes exist, they're just all modals. In that case the preloaded tail is dropped here, and the new `newStackState.routes.length > 0` guard in `ModalStack.tsx` then skips rendering `NativeStackView` entirely, so prefetched screens unmount and lose their state. The old code always rendered the view and concatenated `state.preloadedRoutes` regardless.

Repro: web app whose current stack is only modal routes, prefetch a non-modal sibling → the prefetched screen unmounts. Appending `preloadedRoutes` unconditionally (and rendering `NativeStackView` when only preloaded routes exist) restores the old behavior — though `index` handling for the empty-active case needs care.

packages/expo-router/CHANGELOG.md
```diff
@@ -39,6 +39,7 @@
 
 ### 💡 Others
 
+- Project preloaded routes in native-stack. ([#47961](https://github.com/expo/expo/pull/47961) by [@Ubax](https://github.com/Ubax))
```
Ubax: [💡 suggestion] This entry undersells an observable breaking change: `StackNavigationState` no longer has `preloadedRoutes` and is publicly reachable through `expo-router`'s exports (`src/exports.ts`), so user code reading `state.preloadedRoutes` (e.g. via `useRootNavigationState()`) breaks — compile-time if typed, silently `undefined` at runtime. Additionally, `navigation.reset({ routes, index })` with `index < routes.length - 1` changes meaning: the tail used to stay active (topmost visible), now it becomes detached preloaded screens. Worth spelling both out here so it's findable when someone hits it.

packages/expo-router/src/modal/web/utils.ts
Ubax: Fix this

packages/expo-router/src/react-navigation/core/useNavigationCache.tsx
Ubax: Add a test case for this. Throw when dispatch is happening from preloaded screen

packages/expo-router/src/react-navigation/core/useOnPreventRemove.tsx
Ubax: Add test case for this. Then disable preventRemove in preloaded screens

packages/expo-router/CHANGELOG.md
Ubax: Move to breaking changes

kitten (approved):
exact correctness is a bit tricky to review, due to the wide-ranging changes, but as long as it works correctly, it has the same wide-ranging impact of the state refactor, and the individual commits and their logical changes lgtm 👍

PR #47967, opened 2026-07-20
[video][ios] Fix source loading race conditions

behenate:
# Why

Rapid source changes can leave multiple loading tasks in flight. A stale task can clear newer loader state or emit callbacks after the player has been released. This is very rare, but it's good to patch it up.

# How

Synchronize loader state with a Mutex and use task IDs to ignore stale completions. Close the loader when releasing the player, add cancellation checks around player item creation, and keep weak listener identities stable after deallocation.

# Test Plan

Tested in BareExpo on iOS

### DIFF
--- a/packages/expo-video/ios/VideoPlayerItem.swift
+++ b/packages/expo-video/ios/VideoPlayerItem.swift
@@ -29,6 +29,8 @@ class VideoPlayerItem: AVPlayerItem {
   }
 
   init?(videoSource: VideoSource, urlOverride: URL? = nil) async throws {
+    try Task.checkCancellation()
+
     guard let url = urlOverride ?? videoSource.uri else {
       return nil
     }
@@ -43,10 +45,13 @@ class VideoPlayerItem: AVPlayerItem {
       try await asset.prepareForLoadingIfNeeded()
       _ = try await asset.load(.duration, .preferredTransform, .isPlayable)
     } catch {
-        // Catch block is intentionally left empty
+      // Asset errors are surfaced through the player status, but cancellation must stop obsolete work.
+      try Task.checkCancellation()
     }
 
+    try Task.checkCancellation()
     super.init(asset: urlAsset, automaticallyLoadedAssetKeys: nil)
+    try Task.checkCancellation()
     self.createTracksLoadingTask()
   }
 
--- a/packages/expo-video/ios/VideoSourceLoader.swift
+++ b/packages/expo-video/ios/VideoSourceLoader.swift
@@ -129,8 +129,24 @@ internal class VideoSourceLoader {
         return LoadingResult(value: nil, isCancelled: false)
       }
 
-      let safeUrl = try await url.toUrlWithPermissions()
+      // A resolution failure (only possible for `ph://` sources) shouldn't reject the load.
+      // Build the item with the raw, unplayable URL instead, so the player transitions to the
+      // `.error` status the same way it does for any other failing source, with `transportError`
+      // carrying the real resolution error to that status change.
+      var safeUrl: URL?
+      var resolutionError: Error?
+      do {
+        safeUrl = try await url.toUrlWithPermissions()
+      } catch let error as CancellationError {
+        throw error
+      } catch {
+        resolutionError = error
+      }
+
       let playerItem = try await VideoPlayerItem(videoSource: videoSource, urlOverride: safeUrl)
+      if let resolutionError {
+        playerItem?.urlAsset.transportError = resolutionError
+      }
 
       try Task.checkCancellation()
       return LoadingResult(value: playerItem, isCancelled: false)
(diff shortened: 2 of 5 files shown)


packages/expo-video/ios/VideoSourceLoader.swift
```diff
@@ -1,86 +1,217 @@
...
 
-    return LoadingResult(value: playerItem, isCancelled: false)
+  /**
+   Must be called while holding `state` so listener events are enqueued in state transition order.
+   */
+  private func enqueueListenerEventWhileLocked(
+    listeners: Set<WeakVideoSourceLoaderListener>,
+    event: @escaping (VideoSourceLoaderListener, VideoSourceLoader) -> Void
+  ) {
+    DispatchQueue.main.async { [weak self] in
```
tsapeta, line 220: [🏗️ **design**] Listener callbacks now always hop to the main queue via `DispatchQueue.main.async`, whereas previously they were delivered synchronously on whatever context `load()`/`loadImpl()` ran on. This is very likely correct/desirable since `VideoPlayerObserver.status` and `isLoadingAsynchronously` aren't thread-safe and now get consistent main-thread access. Worth calling out explicitly: after `load()` returns to `replaceCurrentItem(async)`, `onLoadingFinished` has *not* fired yet (it's queued behind the current main-queue work). Since the observer's `isLoadingAsynchronously` flip and the item replacement both land on main, ordering should hold, but please confirm no consumer assumed the callback ran before `load()`'s continuation resumes.

packages/expo-video/ios/VideoSourceLoader.swift
```diff
@@ -1,86 +1,217 @@
...
 
-    if !loadingResult.isCancelled {
-      listeners.forEach { listener in
-        listener.value?.onLoadingFinished(loader: self, videoSource: videoSource, result: loadingResult.value)
-      }
+    let loadingResult: LoadingResult
+    do {
+      loadingResult = try await context.task.value
+    } catch {
+      finishLoading(taskId: context.taskId, videoSource: videoSource)
```
tsapeta, line 97: [💡 suggestion] On a non-cancellation throw (e.g. `toUrlWithPermissions()` failing), this calls `finishLoading(taskId:videoSource:)` with no `loadingResult`, which enqueues `onLoadingCancelled` → the observer sets `status = .idle`. The error is still re-thrown to the caller, so this is only a semantics nit: a genuine load *failure* is reported to listeners as a *cancellation* rather than surfacing an error status. Pre-existing behavior didn't fire any listener event on throw, so this isn't a regression, but if listeners should distinguish "failed" from "cancelled" it may be worth an explicit failure path.

packages/expo-video/ios/VideoSourceLoader.swift
```diff
@@ -1,86 +1,217 @@
...
+    }
+    currentTask?.cancel()
+  }
 
-    return LoadingResult(value: playerItem, isCancelled: false)
+  /**
+   Must be called while holding `state` so listener events are enqueued in state transition order.
+   */
+  private func enqueueListenerEventWhileLocked(
+    listeners: Set<WeakVideoSourceLoaderListener>,
```
tsapeta, line 217: [🐧 nit] `enqueueListenerEventWhileLocked` snapshots the `listeners` set at enqueue time and delivers on the next main-queue turn. A listener that is still alive but has called `unregisterListener` in between will still receive the callback. In practice the only listener is `VideoPlayerObserver`, whose lifetime tracks the player, so this is benign today, but it's a latent surprise if more listeners are added. Consider re-reading `state.listeners` inside the async block (under lock) instead of capturing a snapshot.

packages/expo-video/ios/VideoPlayerItem.swift
```diff
@@ -43,10 +45,13 @@ class VideoPlayerItem: AVPlayerItem {
       try await asset.prepareForLoadingIfNeeded()
       _ = try await asset.load(.duration, .preferredTransform, .isPlayable)
     } catch {
-        // Catch block is intentionally left empty
+      // Asset errors are surfaced through the player status, but cancellation must stop obsolete work.
+      try Task.checkCancellation()
     }
 
+    try Task.checkCancellation()
     super.init(asset: urlAsset, automaticallyLoadedAssetKeys: nil)
```
tsapeta, line 53: [🐧 nit] The check right before `super.init` and the one right after are separated only by the synchronous, non-suspending `super.init`, so cancellation can't newly arrive between them. The post-`super.init` check is redundant with the pre-`super.init` one and can be dropped. The meaningful checks are: at entry, and after the async asset `load` (both suspension points). Minor; harmless as-is.

tsapeta (approved):

packages/expo-video/ios/VideoPlayerItem.swift
behenate: Cancellation can be set at any moment, including while super.init is running. So the post-init check can catch a cancel that arrived mid-init and skip spawning the tracks-loading task. It's a tiny window, but the check is one line, so I'd rather keep it.

packages/expo-video/ios/VideoSourceLoader.swift
behenate: Added a comment addressing this

packages/expo-video/ios/VideoSourceLoader.swift
behenate: Fixed!

PR #47975, opened 2026-07-20
[video][ios] Reduce main thread load when loading videos

behenate:
# Why

This PR aims to clean up the source loading flow while slightly reducing the load on the main thread when loading videos.

# How

- Introduce `VideoLoadingActor.swift` - used to load the sources off the main thread in a more organised way
- Preload more asset properties so that the AVPlayer doesn't try to load them later on the main thread as it tries to play
- And avoid reassigning the same player to AVPlayerViewController, which recreates its KVOs
- The `replaceCurrentItem(with:)` still has to run on the main thread.

- The main thread is now slightly less strained when replacing the player item. The difference is not big as we were already loading most possible things on a different thread, but in some cases noticeable.

# Test Plan

Tested in BareExpo on iOS

### DIFF
--- a/packages/expo-video/ios/OrientationAVPlayerViewControllerWrapper.swift
+++ b/packages/expo-video/ios/OrientationAVPlayerViewControllerWrapper.swift
@@ -51,7 +51,10 @@ internal final class OrientationAVPlayerViewControllerWrapper: VideoPlayerObserv
     #endif
 
     guard hasStalePlayerReference, !isPresenting else {
-      controller.player = player?.ref
+      // Skip redundant assignments to avoid re-creating KVOs
+      if controller.player !== player?.ref {
+        controller.player = player?.ref
+      }
       return false
     }
 
--- a/packages/expo-video/ios/OrientationAVPlayerViewControllerWrapper.swift
+++ b/packages/expo-video/ios/OrientationAVPlayerViewControllerWrapper.swift
@@ -51,7 +51,7 @@ internal final class OrientationAVPlayerViewControllerWrapper: VideoPlayerObserv
     #endif
 
     guard hasStalePlayerReference, !isPresenting else {
-      // Skip redundant assignments to avoid re-creating KVOs
+      // Reassigning the same `AVPlayer` tears down and recreates the controller's KVO observers for no benefit.
       if controller.player !== player?.ref {
         controller.player = player?.ref
       }
--- a/packages/expo-video/ios/VideoPlayerItem.swift
+++ b/packages/expo-video/ios/VideoPlayerItem.swift
@@ -42,7 +42,7 @@ class VideoPlayerItem: AVPlayerItem {
     // and cause it to go into .error state triggering the `onStatusChange` event.
     do {
       try await asset.prepareForLoadingIfNeeded()
-      _ = try await asset.load(
+      let (_, _, _, _, _, characteristics) = try await asset.load(
         .duration,
         .preferredTransform,
         .isPlayable,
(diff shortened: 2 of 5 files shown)


packages/expo-video/ios/VideoPlayerItem.swift
```diff
@@ -41,7 +42,17 @@ class VideoPlayerItem: AVPlayerItem {
...
     do {
       try await asset.prepareForLoadingIfNeeded()
-      _ = try await asset.load(.duration, .preferredTransform, .isPlayable)
+      let (_, _, _, _, _, characteristics) = try await asset.load(
+        .duration,
+        .preferredTransform,
+        .isPlayable,
+        .hasProtectedContent,
+        .tracks,
+        .availableMediaCharacteristicsWithMediaSelectionOptions
```
tsapeta, line 51: [🏗️ **design**] This widens the preload from 3 properties to 6 and then serially loads a media selection group per characteristic, all before the item reaches the player. That moves work off the main thread, which is the goal, but it also adds network round trips to the critical path, and the two effects pull in opposite directions on different source types.

For remote HLS specifically: `availableMediaCharacteristicsWithMediaSelectionOptions` needs the master playlist, then each `loadMediaSelectionGroup` is another `await` in a sequential `for` loop (typically `.audible` and `.legible`, sometimes more). `.tracks` is also empty for HLS until playback begins, so that one may be buying nothing there. On a slow connection this shows up as added latency before first frame in exchange for main-thread hitching the player would otherwise pay lazily. That's a defensible trade, but it is a trade, and the PR description reads as though it's free.

Two suggestions:

- Run the characteristic loads concurrently instead of serially — a `TaskGroup`, or `async let` per characteristic — so the cost is one round trip rather than N.
- Consider gating the media-selection preload on `!isHls`, or skipping it when the source declares no audio/subtitle options, so progressive files get the win without HLS paying the latency.

Did you measure time-to-first-frame for a remote HLS source before and after, alongside the main-thread numbers? If it's flat, this is a non-issue and I'd just take the concurrency tweak.

packages/expo-video/ios/OrientationAVPlayerViewControllerWrapper.swift
```diff
@@ -51,7 +51,10 @@ internal final class OrientationAVPlayerViewControllerWrapper: VideoPlayerObserv
     #endif
 
     guard hasStalePlayerReference, !isPresenting else {
-      controller.player = player?.ref
+      // Skip redundant assignments to avoid re-creating KVOs
+      if controller.player !== player?.ref {
```
tsapeta, line 55: [🐧 nit] The change is correct; just the comment. "Skip redundant assignments to avoid re-creating KVOs" states the mechanism but not why the assignment is redundant, and this sits inside a `guard`'s `else` whose surrounding comment (lines 45-47) is about the *rebuild* path, so a reader can misread this as part of that. Something like "Reassigning the same `AVPlayer` tears down and recreates the controller's KVO observers for no benefit" makes the reason self-contained.

tsapeta (approved):

packages/expo-video/ios/VideoPlayerItem.swift
behenate: Since .availableMediaCharacteristicsWithMediaSelectionOptions is loaded in the asset.load(...) call immediately above the loop, every subsequent loadMediaSelectionGroup(for:) is a resident lookup, not a fetch. So we can pretty safely assume we are not doing N round trips, instead we only do one and then we do a local lookup. Adding `!isHLS` wouldn't be beneficial, in hls we have audio and subtitle selection groups, which be loaded lazily on the main thread if we do it. 


I have also run some TTFF tests with 3G network conditioning, I didn't measure a difference that couldn't be attributed to noise

PR #47980, opened 2026-07-20
Fix user facing typos

zoontek:
# Why

Following this https://github.com/expo/expo/pull/47965#event-28220098318
This PR fixes user facing / comments typos.

# How

- Ran a Sonnet prompt on the whole repository

# Test Plan

- Nothing to test here

# Checklist


packages/@expo/metro-config/src/serializer/__tests__/sourceMap.test.ts
```diff
@@ -764,7 +764,7 @@ describe('sourceMapString', () => {
       jest.resetModules();
     });
 
-    it('replaces both stock exports with Expo`s encoder', () => {
+    it('replaces both stock exports with Expo's encoder', () => {
```
hassankhan: Think this might need adjusting

zoontek: Good catch

hassankhan (approved):

PR #47983, opened 2026-07-20
[router] Remove `useOnlyUserDefinedScreens` from `withLayoutContext`

Ubax:
# Why

Remove `useOnlyUserDefinedScreens` from `withLayoutContext` and use `routeSource` to hide native tab when its Trigger is not specified.

# How

1. Remove `useOnlyUserDefinedScreens` and its logic
2. Filter routes in native-tabs to be `descriptor.routeSource === 'layout'`

# Test Plan

CI

# Checklist

---

### DIFF
--- a/packages/expo-router/src/native-tabs/NativeBottomTabsNavigator.tsx
+++ b/packages/expo-router/src/native-tabs/NativeBottomTabsNavigator.tsx
@@ -1,14 +1,19 @@
 'use client';
 
-import React, { use, useCallback, useMemo, useRef } from 'react';
-import type { NavigatorArgs } from 'standard-navigation';
+import React, { use, useCallback, useEffect, useMemo, useRef } from 'react';
 
+import { useRouteNode } from '../Route';
+import { router } from '../imperative-api';
 import type {
   ParamListBase,
   TabNavigationState,
   TabRouterOptions,
 } from '../react-navigation/native';
 import { unstable_createStandardRouterNavigator } from '../standard-navigation';
+import type {
+  StandardNavigatorContentProps,
+  StandardNavigatorDescriptor,
+} from '../standard-navigation/types';
 import { getAllChildrenNotOfType, getAllChildrenOfType } from '../utils/children';
 import { NativeBottomTabsRouter } from './NativeBottomTabsRouter';
 import { NativeTabTrigger } from './NativeTabTrigger';
@@ -46,9 +51,13 @@ function NativeTabsContent({
   rippleColor,
   disableIndicator,
   labelVisibilityMode,
+  redirectToRouteName,
   ...rest
-}: NavigatorArgs<NativeTabOptions, NativeTabNavigationEventMap> &
-  Omit<InternalNativeTabsProps, 'screenListeners'>) {
+}: StandardNavigatorContentProps<
+  NativeTabOptions,
+  NativeTabNavigationEventMap,
+  Omit<InternalNativeTabsProps, 'screenListeners'>
+>) {
   if (use(NativeTabsContext)) {
     throw new Error(
       'Nesting Native Tabs inside each other is not supported natively. Use JS tabs for nesting instead.'
@@ -60,9 +69,14 @@ function NativeTabsContent({
   const visibleTabs = useMemo(
     () =>
       routes
-        // The <NativeTab.Trigger> always sets `hidden` to defined boolean value.
-        // If it is not defined, then it was not specified, and we should hide the tab.
-        .filter((route) => descriptors[route.key]!.options?.hidden !== true)
+        // Every filesystem route is registered in state; only routes declared by a
+        // non-hidden <NativeTabs.Trigger> become tab items.
+        .filter((route) => {
+          const descriptor = descriptors[
+            route.key
+          ]! as StandardNavigatorDescriptor<NativeTabOptions>;
+          return descriptor.routeSource === 'layout' && descriptor.options?.hidden !== true;
+        })
         .map(
           (route): NativeTabsViewTabItem => ({
             options: descriptors[route.key]!.options,
@@ -82,15 +96,27 @@ function NativeTabsContent({
     [visibleTabs]
   );
 
-  if (visibleFocusedTabIndex < 0) {
-    if (process.env.NODE_ENV !== 'production') {
-      const focusedRoute = routes[state.index];
-      throw new Error(
-        `The focused tab in NativeTabsView cannot be displayed. Make sure path is correct and the route is not hidden. Route: "${focusedRoute?.href ?? focusedRoute?.name}"`
-      );
-    }
-  }
   const focusedIndex = visibleFocusedTabIndex >= 0 ? visibleFocusedTabIndex : 0;
+
+  // The focused route can be hidden or have no trigger at all — for example a path pointing at a
+  // route without a tab, or a trigger hidden while focused. Redirect to the router's initial tab,
+  // falling back to the first visible tab. `replace` keeps the unreachable route out of history.
+  // TODO(@ubax): Show a formsheet for hidden tabs which are focused (Tabs + Stack in one).
+  const redirectTabHref = useMemo(() => {
+    const redirectTab =
+      visibleTabs.find(
+        (tab) =>
+          tab.name === redirectToRouteName ||
+          tab.name.replace(/\/index$/, '') === redirectToRouteName
+      ) ?? visibleTabs[0];
+    return routes.find((route) => route.key === redirectTab?.routeKey)?.href;
+  }, [redirectToRouteName, routes, visibleTabs]);
+  useEffect(() => {
+    if (visibleFocusedTabIndex < 0 && redirectTabHref != null) {
+      router.replace(redirectTabHref);
+    }
+  }, [visibleFocusedTabIndex, redirectTabHref]);
+
   const provenanceRef = useRef(0);
 
   const onTabChange = useCallback(
@@ -147,6 +173,10 @@ function NativeTabsContent({
   > &
     Record<Exclude<keyof typeof rest, keyof NativeTabsViewProps>, never> = rest;
 
+  if (visibleTabs.length === 0) {
+    return null;
+  }
+
   return (
     <NativeTabsContext value>
       <NativeTabsView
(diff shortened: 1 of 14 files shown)


docs/pages/router/advanced/custom-navigators.mdx
Ubax, line 1: Revert

packages/expo-router/src/native-tabs/NativeBottomTabsNavigator.tsx
```diff
@@ -60,9 +66,13 @@ function NativeTabsContent({
...
-        // The <NativeTab.Trigger> always sets `hidden` to defined boolean value.
-        // If it is not defined, then it was not specified, and we should hide the tab.
-        .filter((route) => descriptors[route.key]!.options?.hidden !== true)
+        // Every filesystem route is registered in state; only routes declared by a
+        // non-hidden <NativeTabs.Trigger> become tab items.
+        .filter(
+          (route) =>
+            descriptors[route.key]!.routeSource === 'layout' &&
+            descriptors[route.key]!.options?.hidden !== true
+        )
```
Ubax: Cast to StandardNavigationDescriptor

packages/expo-router/src/native-tabs/NativeBottomTabsNavigator.tsx
```diff
@@ -82,15 +92,24 @@ function NativeTabsContent({
...
+  // The focused route can be hidden or have no trigger at all — for example a path pointing at a
+  // route without a tab, or a trigger hidden while focused. Redirect to the router's initial tab,
+  // falling back to the first visible tab. `replace` keeps the unreachable route out of history.
+  // TODO(@ubax): Show a formsheet for hidden tabs which are focused (Tabs + Stack in one).
+  const redirectTab =
+    visibleTabs.find(
+      (tab) =>
+        tab.name === redirectToRouteName || tab.name.replace(/\/index$/, '') === redirectToRouteName
+    ) ?? visibleTabs[0];
+  const redirectTabHref = routes.find((route) => route.key === redirectTab?.routeKey)?.href;
```
Ubax: Can you wrap this in useMemo?

packages/expo-router/src/native-tabs/NativeBottomTabsNavigator.tsx
```diff
@@ -170,16 +193,21 @@ const NativeTabsNavigatorWithContext = unstable_createStandardRouterNavigator<
...
+        (child) => isChildOfType(child, NativeTabTrigger) || isProtectedReactElement(child)
+      ),
     [props.children]
   );
   const nonTriggerChildren = useMemo(
-    () => getAllChildrenNotOfType(props.children, NativeTabTrigger),
+    () =>
+      Children.toArray(props.children).filter(
+        (child) => !isChildOfType(child, NativeTabTrigger) && !isProtectedReactElement(child)
+      ),
```
Ubax: Extract to separate PR as well

packages/expo-router/src/native-tabs/NativeTabs.tsx
```diff
@@ -31,5 +32,5 @@ export const NativeTabs = Object.assign(
   (props: NativeTabsProps) => {
     return <NativeTabsNavigatorWrapper {...props} />;
   },
-  { Trigger: NativeTabTrigger, BottomAccessory }
+  { Trigger: NativeTabTrigger, Protected, BottomAccessory }
```
Ubax: Extract this to separate PR - add protected to native tabs

packages/expo-router/src/native-tabs/types.ts
```diff
@@ -315,6 +315,11 @@ export interface NativeTabsProps extends PropsWithChildren {
    * Can be overridden by icon color and label color for each tab individually.
    */
   tintColor?: ColorValue;
+  /**
+   * The route shown initially and used as the redirect target when the focused route has no visible
+   * tab trigger.
+   */
+  initialRouteName?: string;
```
Ubax: Remove this or change. The initialRouteName should come from `unstable_settings`

packages/expo-router/src/__tests__/stable-route-registration.test.ios.tsx
Ubax, line 1: Revert

packages/expo-router/src/native-tabs/__tests__/navigation.test.ios.tsx
```diff
@@ -86,7 +83,7 @@ describe('Native Bottom Tabs Navigation', () => {
         </View>
       ),
       hidden: () => <View testID="hidden" />,
-      notSpecified: () => <View testID="not-specified" />,
+      'not-specified': () => <View testID="not-specified" />,
```
Ubax: Why did you change this?

packages/expo-router/src/native-tabs/__tests__/navigation.test.ios.tsx
```diff
@@ -137,29 +134,239 @@ describe('Native Bottom Tabs Navigation', () => {
...
+
+describe('Native Bottom Tabs route registration', () => {
+  function tabRouteNames() {
+    const tabState = store.navigationRef.getRootState().routes[0]!.state!;
+    return tabState.routes.map((r) => r.name);
+  }
+
+  function tabState() {
+    return store.navigationRef.getRootState().routes[0]!.state!;
+  }
```
Ubax: Remove these, and change the assertions to what is rendered only. Don't pay attention to the internal state ref. Please update the tests accordingly

packages/expo-router/src/native-tabs/__tests__/navigation.test.ios.tsx
Ubax: I changed the fixture name to match the `/not-specified` URLs added by these tests. That rename was unnecessary, so I restored `notSpecified` and updated the test URLs to `/notSpecified`.

packages/expo-router/src/native-tabs/NativeBottomTabsNavigator.tsx
Ubax: Yes, the redirect tab and href lookup are now wrapped in `useMemo`.

packages/expo-router/src/native-tabs/NativeBottomTabsNavigator.tsx
```diff
@@ -82,15 +96,27 @@ function NativeTabsContent({
...
 
-  if (visibleFocusedTabIndex < 0) {
-    if (process.env.NODE_ENV !== 'production') {
-      const focusedRoute = routes[state.index];
-      throw new Error(
-        `The focused tab in NativeTabsView cannot be displayed. Make sure path is correct and the route is not hidden. Route: "${focusedRoute?.href ?? focusedRoute?.name}"`
-      );
-    }
-  }
   const focusedIndex = visibleFocusedTabIndex >= 0 ? visibleFocusedTabIndex : 0;
```
Ubax, line 99: [🏗️ **design**] While the redirect effect is pending, this falls back to tab `0`, but the effect then replaces to the `redirectToRouteName` tab. When `initialRouteName` is not the first visible tab (e.g. the `respects initialRouteName` test scenario), the first render mounts and shows tab 0's content, then jumps to the initial tab — a visible flash plus an unnecessary mount of the wrong tab. Consider making the fallback the redirect target's index (`visibleTabs.indexOf(redirectTab)`), so the first render already shows the tab the URL is about to become.

packages/expo-router/src/native-tabs/NativeBottomTabsNavigator.tsx
```diff
@@ -82,15 +96,27 @@ function NativeTabsContent({
...
+      visibleTabs.find(
+        (tab) =>
+          tab.name === redirectToRouteName ||
+          tab.name.replace(/\/index$/, '') === redirectToRouteName
+      ) ?? visibleTabs[0];
+    return routes.find((route) => route.key === redirectTab?.routeKey)?.href;
+  }, [redirectToRouteName, routes, visibleTabs]);
+  useEffect(() => {
+    if (visibleFocusedTabIndex < 0 && redirectTabHref != null) {
+      router.replace(redirectTabHref);
```
Ubax, line 116: [🏗️ **design**] Two things here:

1. **UX question:** the old behavior for `Link`-pressing a hidden/undeclared route was a no-op — the user stayed on their current tab. Now they're yanked to the *initial* tab (the test `second-not-specified-link` → `/` encodes this). Redirecting to the previously focused visible tab (available via the tab router's history) would preserve the user's place; redirecting to the initial tab only really makes sense for the deep-link/cold-start case. Was initial-tab-always deliberate?

2. **Lost dev signal:** the removed dev-mode error (`The focused tab in NativeTabsView cannot be displayed…`) was the only thing that surfaced a mistyped trigger `name` or an unintended navigation to a non-tab route. Now it silently redirects. A dev-only `console.warn` here ("route X has no visible tab, redirecting to Y") would keep that debuggability without the crash.

packages/expo-router/src/native-tabs/NativeBottomTabsNavigator.tsx
```diff
@@ -252,6 +283,8 @@ export function NativeTabsNavigatorWrapper(props: NativeTabsProps) {
       children={triggerChildren}
       nonTriggerChildren={nonTriggerChildren}
       screenOptions={screenOptions}
+      initialRouteName={routeNode?.initialRouteName}
+      redirectToRouteName={routeNode?.initialRouteName}
```
Ubax, line 287: [💡 suggestion] Normalization is inconsistent between the two consumers of `routeNode.initialRouteName`. The redirect matcher strips `/index` (so `initialRouteName: 'second'` matches the `second/index` tab), but this raw value goes to `useNavigationBuilder`, where the vendored `TabRouter` requires an exact `routeNames` match and silently falls back to index 0. So with `unstable_settings: { initialRouteName: 'second' }` and `app/second/index.tsx`: cold start focuses the first trigger, but an invalid deep link redirects to `second` — two different "initial" tabs. Consider resolving the name against the route node's children once here (the same `name || name + '/index'` matching `getSortedChildren` uses) and passing the resolved route name to both props.

packages/expo-router/src/native-tabs/NativeBottomTabsNavigator.tsx
Ubax: 1. Yes, in the follow-up, I'll show a formsheet instead
2. Add dev-only console.warń

packages/expo-router/src/native-tabs/NativeBottomTabsNavigator.tsx
Ubax: Share the logic between this and getSortedChildren

packages/expo-router/src/standard-navigation/__tests__/docs-examples.test.tsx
```diff
@@ -80,7 +80,6 @@ function typecheck(_value: unknown) {}
   };
 
   const Tabs = unstable_createStandardRouterNavigator(TabsContent, TabRouter, {
-    useOnlyUserDefinedScreens: true,
```
Ubax, line 83: https://linear.app/expo/issue/ENG-25377/update-standard-navigation-docs closer to SDK release

hassankhan (approved):

PR #47984, opened 2026-07-20
[router] Keep fully guarded navigators mounted

Ubax:
# Why

At the moment when all routes in navigator are protected the filtering passes empty `routeNames` array, so navigator does not register. Since we want to remove the `routeNames` change detection behavior, we need to replace this with redirect to parent instead

# How

1. Pass fallback from parent to children via context
2. Redirect to parent when no redirect is possible in current navigator

# Test Plan

CI

# Checklist

---

### DIFF
--- a/packages/expo-router/src/layouts/GuardContext.tsx
+++ b/packages/expo-router/src/layouts/GuardContext.tsx
@@ -3,59 +3,18 @@
 import { createContext, use, useMemo, type ReactNode } from 'react';
 
 import type { RouteNode } from '../Route';
-import { sortRoutesWithInitial } from '../Route';
-import { getContextKey, stripInvisibleSegmentsFromPath } from '../matchers';
+import { LocalRouteParamsContext, sortRoutesWithInitial } from '../Route';
+import { getContextKey } from '../matchers';
 import type { Href } from '../types';
 
-export const GuardContext = createContext<Map<string, Href> | null>(null);
+export type GuardedRedirects = Map<string, Href | undefined>;
+// `Href` = guarded with a redirect, `null` = guarded with no available destination.
+export type GuardRedirect = Href | null;
+// Per route name; a missing entry means the route is unguarded.
+export type ResolvedGuards = Map<string, GuardRedirect>;
 
-export function useGuardRedirect(routeName: string): Href | undefined {
-  const guards = use(GuardContext);
-  if (!guards) {
-    return undefined;
-  }
-  return guards.get(routeName) ?? guards.get(routeName.replace(/\/index$/, ''));
-}
-
-function matchesRouteName(route: string, name: string) {
-  return route === name || route === `${name}/index` || route.replace(/\/index$/, '') === name;
-}
-
-function routeNodeToHref(route: Pick<RouteNode, 'contextKey'>): Href {
-  return (stripInvisibleSegmentsFromPath(getContextKey(route.contextKey)) || '/') as Href;
-}
-
-function serializeGuardedRedirects(guardedRedirects: Map<string, Href | undefined>): string {
-  return Array.from(guardedRedirects.entries())
-    .map(([name, href]) => `${name}=${href ?? ''}`)
-    .join('|');
-}
-
-function resolveDefaultHref(
-  node: Pick<RouteNode, 'children' | 'initialRouteName'>,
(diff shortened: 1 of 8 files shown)


packages/expo-router/src/useScreens.tsx
```diff
@@ -367,10 +360,13 @@ export function getQualifiedRouteComponent(value: RouteNode) {
     const isRouteType = value.type === 'route';
     const hasRouteKey = !!route?.key;
 
-    if (guardRedirect) {
+    if (hasRedirect) {
+      // A `null` href means guarded with no destination to redirect to (every
+      // candidate is also guarded): render nothing but keep the screen registered
+      // so the navigator and its state stay mounted.
```
Ubax, line 373: Can you add `console.warn` in dev here, to indicate to the user that all routes are protected in the navigator?

packages/expo-router/src/link/__tests__/Link.test.ios.tsx
```diff
@@ -670,6 +670,65 @@ describe('prefetch', () => {
...
+            ],
+            stale: false,
+            type: 'stack',
+          },
+        },
+      ],
+      stale: false,
+      type: 'stack',
+    });
+  });
```
Ubax, line 690: Don't assert the whole state shape. Just check if it includes the `preloadedRoutes` with correct route key

packages/expo-router/src/native-tabs/__tests__/navigation.test.ios.tsx
```diff
@@ -356,4 +357,70 @@ describe('Native Bottom Tabs trigger changes', () => {
...
     expect(TabsScreen).not.toHaveBeenCalled();
   });
+
+  it('keeps the navigator mounted when every trigger is protected', () => {
+    consoleWarnSpy = jest.spyOn(console, 'warn').mockImplementation(() => {});
+
+    renderRouter({
+      _layout: () => (
+        <NativeTabs>
+          <NativeTabs.Protected guard={false}>
```
Ubax: Remove all the new protected tests for native-tabs. The functionality will be added in follow-up pr

packages/expo-router/src/native-tabs/NativeBottomTabsNavigator.tsx
```diff
@@ -65,6 +66,8 @@ function NativeTabsContent({
   }
 
   const { routes } = state;
+  const focusedRoute = routes[state.index]!;
+  const { hasRedirect: focusedRouteHasGuardRedirect } = useGuardRedirect(focusedRoute.name);
```
Ubax: [🚨 **critical**] As written this is dead code: the nearest `GuardContextProvider` is the tabs' own (from `withLayoutContext` wrapping `StandardRouterNavigator`), and its `guardedRedirects` come from the `children` prop — which `NativeTabsNavigatorWrapper` pre-filters to direct `NativeTabTrigger` elements via the shallow `getAllChildrenOfType`. A `Protected` wrapper never reaches `useFilterScreenChildren`, so the guards map is always empty here and `hasRedirect` is always `false`. For this to do anything, the wrapper needs to keep `Protected` elements in the children it forwards (and `NativeTabs` needs to export `Protected` at all).

**Revert this file**

packages/expo-router/src/layouts/GuardContext.tsx
```diff
@@ -7,14 +7,31 @@ import { sortRoutesWithInitial } from '../Route';
...
+// Per route name; a missing entry means the route is unguarded.
+export type ResolvedGuards = Map<string, GuardRedirect>;
 
-export function useGuardRedirect(routeName: string): Href | undefined {
+export const GuardContext = createContext<ResolvedGuards | null>(null);
+
+export function useGuardRedirect(
+  routeName: string
+):
+  | { hasRedirect: true; redirectHref: GuardRedirect }
```
Ubax: [💡 suggestion] `{ hasRedirect: true, redirectHref: null }` reads as a contradiction — the discriminant really means "this route is guarded", not "there is a redirect". Consider `{ isGuarded: true, redirectHref: Href | null }` (and `focusedRouteIsGuarded` at the native-tabs call site). It would also make the `has`-vs-`get` comment less necessary, since the shape itself would say a guarded route may have no destination.

kitten (approved):

PR #47986, opened 2026-07-20
[router] Remove render time set state on `routeName` changes

Ubax:
# Why

In order to use transitions we need to remove all the render time state updates.

# How

1. Remove `navigationKey` API - it is not used in router anyway
2. Remove `getStateForRouteNamesChange` from all the routers
3. Remove render time set state using `getStateForRouteNamesChange`

# Test Plan

1. CI
2. Router e2e
3. Hot module reloading when route files are removed

# Checklist

---


Ubax:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/expo/expo/47986?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#47986** <a href="https://app.graphite.com/github/pr/expo/expo/47986?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/47986?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#48257** <a href="https://app.graphite.com/github/pr/expo/expo/48257?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#48256** <a href="https://app.graphite.com/github/pr/expo/expo/48256?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/expo-router/src/react-navigation/core/BaseNavigationContainer.tsx
```diff
@@ -385,7 +385,7 @@ export function BaseNavigationContainer({
       case 'POP_TO':
       case 'JUMP_TO':
         if (payload?.name) {
-          message += `\n\nDo you have a screen named '${payload.name}'?\n\nIf you're trying to navigate to a screen in a nested navigator, see https://reactnavigation.org/docs/nesting-navigators#navigating-to-a-screen-in-a-nested-navigator.\n\nIf you're using conditional rendering, navigation will happen automatically and you shouldn't navigate manually, see.`;
+          message += `\n\nDo you have a screen named '${payload.name}'?\n\nIf you're trying to navigate to a screen in a nested navigator, see https://reactnavigation.org/docs/nesting-navigators#navigating-to-a-screen-in-a-nested-navigator.`;
```
Ubax, line 388: I created a follow-up ticket to clean messages like this up from references to react-navigation

https://linear.app/expo/issue/ENG-25529/fix-reactnavigationorg-links-in-expo-router

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
```diff
@@ -368,10 +339,10 @@ export function useNavigationBuilder<
   }, {});
 
   const routeNames = routeConfigs.map((config) => config.props.name);
-  const routeKeyList = routeNames.reduce<Record<string, React.Key | undefined>>((acc, curr) => {
-    acc[curr] = screens[curr]!.keys.map((key) => key ?? '').join(':');
-    return acc;
-  }, {});
+  const initialRouteNamesRef = React.useRef(routeNames);
+  if (!isArrayEqual(initialRouteNamesRef.current, routeNames)) {
+    throw new Error('Route names cannot be changed after the navigator has mounted.');
```
Ubax: [🏗️ **design**] This error will be the only signal a user gets when a layout conditionally renders, reorders, or `redirect`-toggles a `Screen` — and it fires during render, so the app hard-crashes. Per the repo's error-message guidance (what/why/how), consider including the old and new route-name lists in the message and pointing at the fix (screens must be statically declared; use `Protected` for conditional availability). Something like: `Route names changed after the navigator mounted (from [a, b] to [b, a]). Screens must be statically defined; wrap conditionally available screens in <Stack.Protected> instead of rendering them conditionally.`

Also note the check is order-sensitive via `isArrayEqual` — merely reordering `<Stack.Screen>` children throws. If that is intentional (it appears to be, per the new test), the message above covers it; if not, compare as sets.

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
```diff
@@ -368,10 +339,10 @@ export function useNavigationBuilder<
   }, {});
 
   const routeNames = routeConfigs.map((config) => config.props.name);
-  const routeKeyList = routeNames.reduce<Record<string, React.Key | undefined>>((acc, curr) => {
-    acc[curr] = screens[curr]!.keys.map((key) => key ?? '').join(':');
-    return acc;
-  }, {});
+  const initialRouteNamesRef = React.useRef(routeNames);
```
Ubax: [💡 suggestion] HMR (route file added/removed in dev) avoids this throw only because `getRoutes` in `useStore.ts` runs unmemoized on every render and creates new `RouteNode` objects, which changes screen component identity and remounts every navigator. That invariant is implicit — if someone later memoizes `getRoutes` on context identity (a plausible optimization), mounted navigators would receive a changed route-name list and every HMR route add/delete would crash the dev app. Worth a short comment here (or next to `getRoutes`) recording that route changes must remount navigators, and ideally a test covering it.

PR #38290, opened 2025-07-24
Fix DOM component metadata overwriting bug causing 404s after EAS upd…

danoc:
# Why

When exporting apps with multiple DOM components (WebViews), only one WebView works correctly after an EAS update while others return 404 errors. This is because the metadata accumulation loop overwrites the `domComponentAssetsMetadata` array for each component instead of accumulating entries, causing only the last processed DOM component to have its metadata preserved in the final metadata.json file.

This regression was introduced in commit 489252a58219cb48d52beef528add184b6143ae2 which refactored DOM component asset handling but inadvertently changed the metadata accumulation from `push()` to direct assignment.

Fixes #37269

# How

Changed the array assignment in the DOM component processing loop to spread the existing metadata before adding new entries:

```diff
domComponentAssetsMetadata[platform] = [
+   ...(domComponentAssetsMetadata[platform] || []),
    ...(await addDomBundleToMetadataAsync(platformDomComponentsBundle)),
    ...transformDomEntryForMd5Filename({
      files,
      htmlOutputName,
    }),
];
```

This ensures all DOM components have their metadata properly preserved instead of being overwritten.

# Test Plan

**Manual Testing:**
1. Create an Expo app with multiple DOM components (WebViews)
2. Build the app with EAS: `eas build --platform ios`
3. Install the build on a device
4. Verify all WebViews load correctly
5. Publish an update
6. Force close and reopen the app to load the update
7. Verify all WebViews still load correctly (previously only one would work)

I tested this fix in a production app with multiple DOM components and confirmed that all components now load correctly after EAS updates, whereas previously only the last one would work. I tested it using `yarn patch` within my project.

**Automated Testing:**
Added comprehensive regression tests in `src/export/__tests__/exportDomComponents-test.ts`:

I also verified that the regression test would have failed without the patch.

# Checklist

- [x] I added a `changelog.md` entry and rebuilt the package sources according to [this short guide](https://github.com/expo/expo/blob/main/CONTRIBUTING.md#-before-submitting)
- [x] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: updated a module plugin).
- [x] Conforms with the [Documentation Writing Style Guide](https://github.com/expo/expo/blob/main/guides/Expo%20Documentation%20Writing%20Style%20Guide.md)

PR #48003, opened 2026-07-21
fix(cli): process DOM components sequentially in expo export to avoid metadata race

### LINKED ISSUE #37269: DOM Components not loading after OTA Update (The requested URL was not found)

**Repository:** https://github.com/fkreinh/dom-components-error

This project was created using `npx create-expo-app@latest --template blank` and includes only the necessary code to reproduce the DOM Components OTA update issue.

### Steps to reproduce

1. **Clone and setup the repository:**
   ```bash
   git clone https://github.com/fkreinh/dom-components-error.git
   cd dom-components-error
   npm install
   ```

2. **Test locally (works correctly):**
   ```bash
   npx expo start
   ```
   - Open on iOS/Android device or simulator
   - Verify DOM component (Lexical rich text editor) renders correctly
   - Confirm CSS styles load properly

3. **Build and deploy initial version:**
   ```bash
   npx expo build:ios  # or build:android
   # Install the built app on device
   ```

4. **Publish OTA update:**
   ```bash
   npx expo publish
   ```

5. **Trigger update on device:**
   - Force close and reopen the app
   - Wait for OTA update to download and apply

**Expected behavior:** DOM components should continue working after OTA updates
**Actual behavior:** DOM components fail to render, showing blank areas or "URL not found" errors

**Platforms affected:** iOS (primary), Android (secondary)
**Environment:** Development build, Standalone app
**Package manager:** npm

A close-up of a cat's face with green eyes
A cat with orange and white fur looking up

### Environment

```text
expo-env-info 1.3.3 environment info:
  System:
    OS: macOS 15.5
  Binaries:
    Node: 20.19.1 - ~/.nvm/versions/node/v20.19.1/bin/node
    npm: 11.3.0 - ~/.nvm/versions/node/v20.19.1/bin/npm
    Watchman: 2025.05.26.00 - /opt/homebrew/bin/watchman
  Managers:
    CocoaPods: 1.16.2 - /Users/fk/.rbenv/shims/pod
  SDKs:
    iOS SDK:
      Platforms: DriverKit 24.2, iOS 18.2, macOS 15.2, tvOS 18.2, visionOS 2.2, watchOS 11.2
  IDEs:
    Android Studio: 2024.3 AI-243.24978.46.2431.13363775
    Xcode: 16.2/16C5032a - /usr/bin/xcodebuild
  npmPackages:
    expo: ~53.0.10 => 53.0.10 
    expo-router: ~5.0.7 => 5.0.7 
    expo-updates: ~0.28.14 => 0.28.14 
    react: 19.0.0 => 19.0.0 
    react-dom: 19.0.0 => 19.0.0 
    react-native: 0.79.3 => 0.79.3 
    react-native-web: ^0.20.0 => 0.20.0 
  npmGlobalPackages:
    eas-cli: 16.7.0
  Expo Workflow: managed
```

### Expo Doctor Diagnostics

```text
14/15 checks passed. 1 checks failed. Possible issues detected:

✖ Validate packages against React Native Directory package metadata
The following issues were found when validating your dependencies against React Native Directory:
  No metadata available: @lexical/react, lexical
Advice:
Update React Native Directory to include metadata for unknown packages. Alternatively, set expo.doctor.reactNativeDirectoryCheck.listUnknownPackages in package.json to false to skip warnings about packages with no metadata, if the warning is not relevant.

```

ammar027:
# Why

#38290 fixed `domComponentAssetsMetadata` being **overwritten** instead of accumulated across DOM components, but the surrounding loop in `exportApp.ts` still processes every DOM component **concurrently** via:

```ts
await Promise.all(
  expoDomComponentReferences.map(async (filePath) => {
    ...
    domComponentAssetsMetadata[platform] = [
      ...(domComponentAssetsMetadata[platform] || []),
      ...(await addDomBundleToMetadataAsync(platformDomComponentsBundle)),
      ...transformDomEntryForMd5Filename({ files, htmlOutputName }),
    ];
  })
);
```

The `await addDomBundleToMetadataAsync(...)` sits **inside the array literal**, between the read of the existing array (the first spread element, evaluated synchronously) and the write (the assignment, which only happens once that `await` resolves). When multiple DOM components are in flight and resolve out of order — e.g. the component that starts first happens to finish last — a later callback's write can be silently clobbered by an earlier-started callback resuming with a stale read of the array. That component's metadata is dropped entirely.

In practice, apps with 2+ `"use dom"` components can end up with some of them missing from the exported `metadata.json` — the corresponding WebView then 404s after loading an EAS Update, even on `@expo/cli` versions that already include the #38290 fix. This reproduced consistently for us in a production app with 2 DOM components: the manifest would non-deterministically ship with 1 or 2 HTML bundles depending on export timing.

Fixes #37269 (the underlying issue is the same one #38290 targeted — that fix was necessary but not sufficient under concurrent execution).

# How

Switched the loop from `Promise.all(expoDomComponentReferences.map(async ...))` to a sequential `for...of` + `await`, so each component's full read-await-write completes before the next one starts. No behavior change other than removing the race — component export order is now deterministic (declaration order) rather than unspecified, and total export time is unaffected in the common case (`Promise.all` didn't meaningfully parallelize this work anyway, since each iteration bundles a whole separate DOM component via Metro).

```diff
-await Promise.all(
-  expoDomComponentReferences.map(async (filePath) => {
-    ...
-  })
-);
+for (const filePath of expoDomComponentReferences) {
+  ...
+}
```

# Test Plan

Added a regression test (`Concurrent DOM component metadata accumulation (race condition)` in `exportDomComponents-test.ts`) that:
1. Reproduces the race by staggering resolution timing so the component that starts first resolves last (mirroring real-world variance in per-component bundling cost) — asserts the concurrent `Promise.all` shape drops entries.
2. Verifies the sequential fix accumulates every entry regardless of per-component latency.

I verified the exact async-ordering mechanics (that the race requires the `await` to sit *inside* the array literal, not merely somewhere before the read-modify-write statement) against a minimal standalone reproduction before writing the test, since a naive "await-then-build-array" version of the test does **not** reproduce the bug — only the "await-inside-the-literal" shape that matches the actual code does.

**Manual verification:** applied the equivalent fix locally (via `pnpm patch`) against a production app with 2 DOM components and confirmed `expo export` now consistently produces the full expected HTML bundle count for both DOM components on both iOS and Android, where it previously non-deterministically dropped one.

# Checklist

- [x] Added a `changelog.md` entry
- [x] This diff will work correctly for `npx expo prebuild` & EAS Build (no native/config-plugin changes — export-time only)

### DIFF
--- a/packages/@expo/cli/CHANGELOG.md
+++ b/packages/@expo/cli/CHANGELOG.md
@@ -19,6 +19,7 @@
 
 ### 🐛 Bug fixes
 
+- Fix a race condition in `expo export` where concurrently-processed DOM components could overwrite each other's entries in `domComponentAssetsMetadata`, causing some DOM components to be missing from `metadata.json` (and 404 in their WebView after an EAS Update) on apps with multiple `"use dom"` components. ([#37269](https://github.com/expo/expo/issues/37269))
 - Avoid writing a 500 response when a DevTools plugin server request has already started or aborted. ([#47192](https://github.com/expo/expo/pull/47192) by [@krystofwoldrich](https://github.com/krystofwoldrich))
 - Pass a fetch API `Request` to DevTools plugin WebSocket handlers instead of `IncomingMessage`. ([#47410](https://github.com/expo/expo/pull/47410) by [@krystofwoldrich](https://github.com/krystofwoldrich))
 - Use `loadModule` for MCP and DevTools plugin server module loading. ([#47139](https://github.com/expo/expo/pull/47139) by [@krystofwoldrich](https://github.com/krystofwoldrich))
--- a/packages/@expo/cli/src/export/__tests__/exportDomComponents-test.ts
+++ b/packages/@expo/cli/src/export/__tests__/exportDomComponents-test.ts
@@ -439,3 +439,158 @@ describe('Multiple DOM components metadata accumulation', () => {
     expect(allPaths).not.toContain('www.bundle/_expo/static/js/web/entry-dom2.js');
   });
 });
+
+describe('Concurrent DOM component metadata accumulation (race condition)', () => {
+  // The spread-accumulation fix (`domComponentAssetsMetadata[platform] =
+  // [...(domComponentAssetsMetadata[platform] || []), ...newEntries]`) is a
+  // non-atomic read-modify-write. `exportApp.ts` processes DOM components via
+  // `Promise.all(expoDomComponentReferences.map(async (filePath) => { ... }))`,
+  // so multiple callbacks can be in flight at once. If callback A reads the
+  // array (to spread it) before callback B has written its own update back,
+  // A's write clobbers B's — even though the accumulation *logic* is correct
+  // in isolation, it is not safe under concurrent execution.
+  //
+  // This suite exercises that scenario directly by staggering resolution
+  // order (the component processed *first* finishes *last*, as commonly
+  // happens when component export cost varies), which the previous
+  // sequential-only tests (see suite above) do not cover.
+  // See: https://github.com/expo/expo/issues/37269
+
+  const makeBundle = (n: number): BundleOutput => ({
+    artifacts: [
+      {
+        filename: `_expo/static/js/web/entry-dom${n}.js`,
+        originFilename: `components/Editor${n}.js`,
+        type: 'js',
+        metadata: {},
+        source: `console.log("dom${n}")`,
+      },
+    ],
+    assets: [],
+  });
+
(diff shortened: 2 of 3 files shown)


bilal1031:
@ammar027 Thank you for the fix; I created a patch using your fix that resolved my issue.

ammar027:
Glad it helped @bilal1031! Good to hear it's not just us hitting this — confirms it's a real race in the export pipeline, not something specific to one app's setup. Hopefully it lands upstream soon so folks don't need the patch.

PR #43018, opened 2026-02-09
[1/n] Upgrade React Native to 0.84.0

chrfalch:
# Why

React-native-0.84 has been released. This PR implements support for it.

## Follow-ups
- [x] Fix CLI tests on Windows
    - RN 0.83 used hermes-compiler@0.14.0, which included win64-bin/hermesc.exe  but RN 0.84.0 uses hermes-compiler@250829098.0.7, which dropped the Windows binaries entirely — it only ships linux64-bin/hermesc and osx-bin/hermesc.
    - Fix for now is forcing Windows tests to use hermes-compiler@0.14.1
- [ ] Fix Expo-Go builds
- [ ] Fix babel-preset tests -> https://github.com/expo/expo/pull/43080
- [x] Fix macOS builds

# How

- Upgraded to react-native 0.84.0
- Fixed issues in the APIs

# Test Plan

✅ Bare Expo iOS
✅ Bare Expo Android

# Checklist

- [x] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: updated a module plugin).

PR #48010, opened 2026-07-21
[iOS] fix(expo,expo-go): Provide `bundleConfiguration` based on final bundle URL

kitten:
Related to #43018
Related to https://github.com/react/react-native/pull/54258
Follow-up to #47424

# Why

React Native 0.84 added `RCTBundleConfiguration` as a new root view (host creation?) argument as `bundleConfiguration`, while `RCTDevSettings` now uses this input to start the `RCTPackagerConnection`.

Internally, `RCTPackagerConnection` is now instance-owned by `RCTDevSettings` and initialized from `RCTBundleManager.bundleConfig`.

Expo was passing the real resolved bundle URL through `bundleURLBlock`, however, hooked up `RCTBundleConfiguration.default()` to the `bundleConfiguration` argument. As a result, the bundle was loading correctly (from the manifest-provided URL), however, the command websocket on `/message` used `RCTBundleURLProvider`'s default shared settings, which is `localhost:8081`.

Previously, this wasn't as noticeable since Expo Go and `expo-dev-client` would reconnect immediately via `devSettings.packagerConnection`. But, this reconnection was removed on Expo Go and this made the issue more obvious. Either way, the `reconnect` path doesn't seem to be desired fix here.

# How

This PR adds `EXBundleConfiguration` / `ExpoBundleConfiguration`, which implements/sub-classes `RCTBundleConfiguration` so it can be derived from an already normalised and resolved bundle URL.

This:
- returns the original bundle URL on `getBundleURL`
- derives the host and scheme from the bundle URL
- preserves file URL behaviour as is on `initWithBundleFilePath`
- falls back to the default configuration as well otherwise

This is then used in `ExpoReactNativeFactory` and in Expo Go's root view.

# Test Plan

- Compiled and ran `apps/expo-go` and connected to `expo start --port 1337 --clear`
  - NOTE: No connections to `8081` are observed and the reload command works

# Checklist

### DIFF
--- a/packages/expo/CHANGELOG.md
+++ b/packages/expo/CHANGELOG.md
@@ -25,7 +25,7 @@
 - [iOS] Fix `Linking.getInitialURL()` returning `null` and deep links being dropped when a URL cold-starts an app on the UIKit scene life cycle. ([#47628](https://github.com/expo/expo/pull/47628) by [@tsapeta](https://github.com/tsapeta))
 - Fix `import.meta.url` being `null` on web when read after the bundle's synchronous execution (from effects, async code, or dynamic imports). ([#47802](https://github.com/expo/expo/pull/47802) by [@zoontek](https://github.com/zoontek))
 - [iOS] Mark `ExpoAppSceneDelegate` as unavailable in `iOSApplicationExtension` for widgets. ([#47894](https://github.com/expo/expo/pull/47894) by [@jakex7](https://github.com/jakex7))
-- [iOS] Add `ExpoBundleConfiguration` derive `RCTBundleConfiguration` from normalized bundle URL instead of default shared settings singleton ([#48010](https://github.com/expo/expo/pull/48010) by [@kitten](https://github.com/kitten))
+- [iOS] Add ExpoBundleConfiguration to derive RCTBundleConfiguration from the normalized bundle URL instead of default shared settings singleton ([#48010](https://github.com/expo/expo/pull/48010) by [@kitten](https://github.com/kitten))
 
 ### 💡 Others
 
--- a/packages/expo/CHANGELOG.md
+++ b/packages/expo/CHANGELOG.md
@@ -25,6 +25,7 @@
 - [iOS] Fix `Linking.getInitialURL()` returning `null` and deep links being dropped when a URL cold-starts an app on the UIKit scene life cycle. ([#47628](https://github.com/expo/expo/pull/47628) by [@tsapeta](https://github.com/tsapeta))
 - Fix `import.meta.url` being `null` on web when read after the bundle's synchronous execution (from effects, async code, or dynamic imports). ([#47802](https://github.com/expo/expo/pull/47802) by [@zoontek](https://github.com/zoontek))
 - [iOS] Mark `ExpoAppSceneDelegate` as unavailable in `iOSApplicationExtension` for widgets. ([#47894](https://github.com/expo/expo/pull/47894) by [@jakex7](https://github.com/jakex7))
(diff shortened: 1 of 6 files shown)


packages/expo/CHANGELOG.md
```diff
@@ -25,6 +25,7 @@
 - [iOS] Fix `Linking.getInitialURL()` returning `null` and deep links being dropped when a URL cold-starts an app on the UIKit scene life cycle. ([#47628](https://github.com/expo/expo/pull/47628) by [@tsapeta](https://github.com/tsapeta))
 - Fix `import.meta.url` being `null` on web when read after the bundle's synchronous execution (from effects, async code, or dynamic imports). ([#47802](https://github.com/expo/expo/pull/47802) by [@zoontek](https://github.com/zoontek))
 - [iOS] Mark `ExpoAppSceneDelegate` as unavailable in `iOSApplicationExtension` for widgets. ([#47894](https://github.com/expo/expo/pull/47894) by [@jakex7](https://github.com/jakex7))
+- [iOS] Add `ExpoBundleConfiguration` derive `RCTBundleConfiguration` from normalized bundle URL instead of default shared settings singleton ([#48010](https://github.com/expo/expo/pull/48010) by [@kitten](https://github.com/kitten))
```
chrfalch:
```suggestion
- [iOS] Add ExpoBundleConfiguration to derive RCTBundleConfiguration from the normalized bundle URL instead of default shared settings singleton ([#48010](https://github.com/expo/expo/pull/48010) by [@kitten](https://github.com/kitten))
```

packages/expo/ios/AppDelegates/EXBundleConfiguration.mm
```diff
@@ -0,0 +1,77 @@
...
+@implementation EXBundleConfiguration {
+  NSURL *_bundleURL;
+}
+
+- (instancetype)initWithBundleURL:(NSURL *)bundleURL
+{
+  if (bundleURL.fileURL) {
+    self = [super initWithBundleFilePath:bundleURL];
+  } else {
+    self = [super init];
```
chrfalch, line 16: Will calling `[super init]` here bypass any important initialization logic in `RCTBundleConfiguration`?

chrfalch (approved):
Couple of comments - thanks for working on this!

gabrieldonadel (approved):
LGTM

packages/expo/ios/AppDelegates/EXBundleConfiguration.mm
kitten: In theory, I think the purpose of `RCTBundleConfiguration` is to just hold values. https://github.com/react/react-native/blob/677f571059cbbd53b943d5b7176c3732bb19aa3f/packages/react-native/React/Base/RCTBundleManager.m#L16-L61

But on a second read I do see that `RCTDevLoadingViewSetEnabled(false);` call and I'm not sure why it's there or what exactly that does

alanjhughes (approved):

kitten:
We'll make more changes/improvements/fixes related to what we found while creating this fix, so I'll hold off on backporting and will revisit the [`RCTDevLoadingViewSetEnabled(false)` call](https://github.com/expo/expo/pull/48010#discussion_r3621676122) once we're ready again to test this end to end

Edit: CI failure looks unrelated (expo-modules-jsi)

packages/expo/ios/AppDelegates/EXBundleConfiguration.mm
kitten: As per the last comment, I'll merge first, so we can work off of `main`, but am assuming that I'll have to revisit the above by testing this end-to-end, once the other changes are ready too

PR #48017, opened 2026-07-21
[android][ui] Add color scheme to community menu

mkaradeniz:
# Why

The Android community `MenuView` creates its own Jetpack Compose `Host`, so consumers cannot force the menu palette when the React Native app theme differs from the device theme. The menu can therefore render light inside an explicitly dark app.

# How

Adds an Android-only `colorScheme` prop to `MenuComponentProps` and forwards it to the internal Compose `Host`. Omitting the prop preserves the existing system-derived appearance.

# Test Plan

Ran from the Expo repository:

- `pnpm --filter @expo/ui format --check`
- `pnpm --filter @expo/ui lint`
- `pnpm --filter @expo/ui typecheck`
- `pnpm --filter @expo/ui build`

A downstream Android app uses the equivalent patch with `colorScheme="dark"`; the menu follows the dark Compose palette while omitted values continue to follow the system theme.

# Checklist


intergalacticspacehighway (approved):
LGTM — clean, well-scoped additive prop. ✅

The community `MenuView` mounts its own Compose `Host` internally, so a consumer's `colorScheme` never reaches it; forwarding the prop is the right fix.

Verified:
- Forwarding target is real — the jetpack-compose `Host` accepts `colorScheme` (`HostProps.colorScheme?: ColorSchemeName`) and threads it to native `HostView.kt` (`ExpoColorScheme` → `toColorScheme`).
- Type matches on both sides (`ColorSchemeName`); `Host` already normalizes `null`/`undefined` back to the system theme, so omitting preserves current behavior.
- iOS unaffected: added to the shared props type but marked `@platform android`, and `MenuView.ios.tsx` doesn't destructure it.
- Changelog formatted correctly under 🎉 New features.

Non-blocking: the swift-ui `Host` likely supports `colorScheme` too, so this could be made universal in a follow-up.

PR #48019, opened 2026-07-21
[image-picker][Android] Fix crop failing for large images

### LINKED ISSUE #48011: [expo-image-picker][Android] Fatal crash in CropImageContract.parseResult (IllegalArgumentException: Required value was null) when the crop activity finishes with an error

https://github.com/aashishshrestha5532/expo-image-picker-crop-crash-repro

One button calling `launchImageLibraryAsync({ allowsEditing: true })`; the README includes a one-liner to generate a reliably-crashing test image (28000×28000 JPEG) and push it to the device gallery.

### Summary

On Android, `launchImageLibraryAsync`/`launchCameraAsync` with `allowsEditing: true` fatally crashes the app when the crop activity fails to process the selected image — typically a decode/OOM failure on very large images. The crash is thrown during native activity-result dispatch, so it cannot be caught from JS. It is deterministic per image (the same large image crashes every time; normal photos work), which makes it look flaky across devices until the Crashlytics signature is spotted.

**Stack trace (Crashlytics, production):**

```
Fatal Exception: java.lang.IllegalArgumentException: Required value was null.
       at expo.modules.imagepicker.contracts.CropImageContract.parseResult(CropImageContract.kt:71)
```

**Root cause:** when the crop activity fails, CanHub's `CropImageActivity` finishes with `CropImage.CROP_IMAGE_ACTIVITY_RESULT_ERROR_CODE` (204) and an `ActivityResult` whose `uriContent` is **null** — the failure is carried in `result.error`. `CropImageContract.parseResult` only checks for `RESULT_CANCELED`, so the error result falls through to the `requireNotNull`:

```kotlin
if (resultCode == Activity.RESULT_CANCELED || result == null) {
  return ImagePickerContractResult.Cancelled
}
val targetUri = requireNotNull(result.uriContent)   // ← throws on error results
```

Two consistency notes:

- Android-Image-Cropper's own README requires checking `result.isSuccessful` before using `uriContent` (else read `result.error`) — that check is missing here.
- `ImageLibraryContract` in this same package already returns `ImagePickerContractResult.Error` when it cannot produce a usable URI, and `ImagePickerModule` already maps `Error` to `FailedToPickMediaException` (a clean JS promise rejection). `CropImageContract` just never got the same handling.

**Affected versions:** reproduced in production on 17.0.10 (SDK 54); the unguarded line is byte-identical in 16.1.4, in the latest published 57.0.5, and on `main` as of 2026-07-21 — the recent `ExpoCropImageActivity` migration (2e0af0a9) does not touch `parseResult`, so the bug survives that rewrite as well.

**Suggested fix** (we run exactly this as a package patch in our production app, verified on-device — the huge image now surfaces a handled error, normal images crop and return as before):

```kotlin
if (resultCode == Activity.RESULT_CANCELED || result == null) {
  return ImagePickerContractResult.Cancelled
}
if (result.uriContent == null) {
  // The crop activity finished with an error (CROP_IMAGE_ACTIVITY_RESULT_ERROR_CODE,
  // e.g. a decode/OOM failure on very large images) — the exception is in result.error.
  return ImagePickerContractResult.Error
}
val targetUri = requireNotNull(result.uriContent)
```

With this change the JS promise rejects with `FailedToPickMediaException` and apps can show a proper error message instead of dying. Happy to open a PR with this change.

### Steps to reproduce

1. Add a very large image to the device gallery. A 28000×28000 solid-color JPEG (~12 MB file) reproduces reliably:
   ```sh
   python3 -c "from PIL import Image; Image.new('RGB',(28000,28000),'#B03A3A').save('huge.jpg',quality=85)"
   adb push huge.jpg /sdcard/Pictures/
   adb shell am broadcast -a android.intent.action.MEDIA_SCANNER_SCAN_FILE -d file:///sdcard/Pictures/huge.jpg
   ```
2. On Android, call `launchImageLibraryAsync({ mediaTypes: ['images'], allowsEditing: true })` (see the repro app).
3. Pick the huge image. The crop screen opens normally (its preview decode is downsampled).
4. Tap **CROP** → the app crashes with the stack above; a `try/catch` around the call never fires.

Reproduced on a Pixel 8 Pro emulator (API 35) and multiple physical devices.

### Environment

Repro project (managed, SDK 57):

```
    System:
      OS: macOS 26.5.2
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 20.18.1
      npm: 10.8.2
    npmPackages:
      expo: ~57.0.7 => 57.0.7
      expo-image-picker: ~57.0.5
      react: 19.2.3 => 19.2.3
    npmGlobalPackages:
      eas-cli: 18.0.5
    Expo Workflow: managed
```

Also hit in our production app: bare workflow, SDK 54 (`expo@54.0.33`, `expo-image-picker@17.0.10`, RN 0.81.5), Android only — iOS uses the system editor and is unaffected.

### Expo Doctor Diagnostics

```
```

Wenszel:
# Why

fixes: https://github.com/expo/expo/issues/48011

Cropping large images (for example 28000x28000px) couldn't be handled by the image cropper, which returned a null URI to `parseResult` which wasn't able to handle it and crashed the app with `IllegalArgumentException`

# How

- added a new result function with error code `CROP_IMAGE_ACTIVITY_RESULT_ERROR_CODE`
- handled the error code in `parseResult`, which now branches solely on the result code
- dropped the unnecessary result intent and replaced it with `input.outputFile`

# Test Plan

BareExpo ✅

### DIFF
--- a/packages/expo-image-picker/CHANGELOG.md
+++ b/packages/expo-image-picker/CHANGELOG.md
@@ -15,6 +15,7 @@
 - [Android] Grant the camera app explicit access to the output URI, so image capture keeps working as Android removes the implicit URI grant for `ACTION_IMAGE_CAPTURE`. ([#46954](https://github.com/expo/expo/pull/46954) by [@alanjhughes](https://github.com/alanjhughes))
 - [Android] Fix `videoMaxDuration` option ([#47504](https://github.com/expo/expo/pull/47504) by [@Wenszel](https://github.com/Wenszel))
 - [iOS] Fix `base64` returning original (non-JPEG) image data when `allowsEditing` is `false`. ([#48005](https://github.com/expo/expo/pull/48005) by [@barthap](https://github.com/barthap))
+- [Android] Fix crop failing for large images ([#48019](https://github.com/expo/expo/pull/48019) by [@Wenszel](https://github.com/Wenszel))
 
 ### 💡 Others
 
--- a/packages/expo-image-picker/android/src/main/java/expo/modules/imagepicker/ExpoCropImageActivity.kt
+++ b/packages/expo-image-picker/android/src/main/java/expo/modules/imagepicker/ExpoCropImageActivity.kt
@@ -1,6 +1,5 @@
 package expo.modules.imagepicker
 
-import android.content.Intent
 import android.content.res.Configuration
 import android.graphics.Color
 import android.graphics.PorterDuff
@@ -333,7 +332,7 @@ class ExpoCropImageActivity :
 
   override fun onSetImageUriComplete(view: CropImageView, uri: Uri, error: Exception?) {
     if (error != null) {
-      setResult(null, error, 1)
+      setResultError()
       return
     }
     if (cropImageOptions.initialCropWindowRectangle != null) {
@@ -350,7 +349,11 @@ class ExpoCropImageActivity :
   }
 
   override fun onCropImageComplete(view: CropImageView, result: CropImageView.CropResult) {
-    setResult(result.uriContent, result.error, result.sampleSize)
+    if (result.error != null) {
+      setResultError()
+      return
+    }
+    setResultOK()
   }
   // endregion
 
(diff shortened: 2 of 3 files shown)


Wenszel:
* **#48019** <a href="https://app.graphite.com/github/pr/expo/expo/48019?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/48019?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Issue #40031, opened 2025-09-27
Error: [ios.dangerous]: withIosDangerousBaseMod: Invalid mimeType for image with source


Code-Parth:
### Minimal reproducible example

using a new icon composer for ios icons and failing due to Invalid mimeType

### Steps to reproduce

instead of using `app.config.json`, I used `app.config.ts`

And here are the logs and information
```ts
ios: {
    supportsTablet: true,
    bundleIdentifier: "com.bundle.identifier",
    infoPlist: {
      NSCameraUsageDescription:
        "This app uses the camera to scan QR codes and deep links.",
    },
    associatedDomains: ["applinks:domain.xyz"],
    icon: "./assets/icon.icon",
  },
```

```bash
✔ Cleared ios code
✔ Created native directory
✔ Updated package.json | no changes
✖ Prebuild failed
Error: [ios.dangerous]: withIosDangerousBaseMod: Invalid mimeType for image with source: ./assets/icon.icon
Error: [ios.dangerous]: withIosDangerousBaseMod: Invalid mimeType for image with source: ./assets/icon.icon
    at ensureImageOptionsAsync (/opt/homebrew/lib/node_modules/expo/node_modules/@expo/image-utils/build/Image.js:180:15)
    at async generateImageAsync (/opt/homebrew/lib/node_modules/expo/node_modules/@expo/image-utils/build/Image.js:219:18)
    at async generateUniversalIconAsync (/opt/homebrew/lib/node_modules/expo/node_modules/@expo/prebuild-config/build/plugins/icons/withIosIcons.js:160:15)
    at async setIconsAsync (/opt/homebrew/lib/node_modules/expo/node_modules/@expo/prebuild-config/build/plugins/icons/withIosIcons.js:94:20)
    at async /opt/homebrew/lib/node_modules/expo/node_modules/@expo/prebuild-config/build/plugins/icons/withIosIcons.js:53:5
    at async action (/opt/homebrew/lib/node_modules/expo/node_modules/@expo/config-plugins/build/plugins/withMod.js:199:23)
    at async interceptingMod (/opt/homebrew/lib/node_modules/expo/node_modules/@expo/config-plugins/build/plugins/withMod.js:104:21)
    at async action (/opt/homebrew/lib/node_modules/expo/node_modules/@expo/config-plugins/build/plugins/createBaseMod.js:60:21)
[...]

PR #48026, opened 2026-07-21
[cli,router-server] feat: support SVG favicons via `web.favicon` and `public/favicon.svg`

### LINKED ISSUE #48024: `expo export -p web` crashes with "Invalid mimeType" when `web.favicon` is an SVG

https://github.com/danfitz36/expo-svg-favicon-repro

Fresh `npx create-expo-app@latest --template blank` with only the changes needed to demonstrate the bug: `assets/favicon.svg` added, `app.json` `expo.web.favicon` pointed at it (and `expo.web.output: "single"` set explicitly), and `react-dom` + `react-native-web` installed (required by `expo export -p web`). `App.js` / `index.js` are untouched from the scaffold.

### Steps to reproduce

Platform: **web**. Environment: **managed workflow**, `expo export -p web`. Package manager: **npm**.

1. `git clone https://github.com/danfitz36/expo-svg-favicon-repro && cd expo-svg-favicon-repro`
2. `npm install`
3. `npx expo export -p web`

**Expected:** the export completes, `dist/index.html` references the SVG favicon (e.g. `<link rel="icon" type="image/svg+xml" href="/favicon.svg">`), and `dist/favicon.svg` is served.

**Observed:** export fails with

```
Web Bundled 1844ms index.js (179 modules)
Error: Invalid mimeType for image with source: ./assets/favicon.svg
    at ensureImageOptionsAsync (.../@expo/image-utils/build/Image.js:181:15)
    at async generateImageAsync (.../@expo/image-utils/build/Image.js:222:18)
    at async getFaviconFromExpoConfigAsync (.../@expo/cli/build/src/export/favicon.js:112:28)
    at async getVirtualFaviconAssetsAsync (.../@expo/cli/build/src/export/favicon.js:69:18)
    at async .../@expo/cli/build/src/export/exportApp.js:283:40
    at async Promise.all (index 0)
    at async exportAppAsync (.../@expo/cli/build/src/export/exportApp.js:193:13)
    at async exportAsync (.../@expo/cli/build/src/export/exportAsync.js:91:5)
```

`dist/` is left empty — no `index.html`, no bundle, no favicon.

**Why this matters:** modern browsers honor `<style>@media (prefers-color-scheme: dark) { ... }</style>` inside an SVG used as a favicon (Chrome, Safari, Firefox). It's the cleanest way to ship a single asset that adapts to a user's light/dark tab strip. Today there's no supported way to do that in an Expo web app.

### Root cause (from source inspection)

`getFaviconFromExpoConfigAsync` in `packages/@expo/cli/src/export/favicon.ts` unconditionally calls `generateImageAsync` from `@expo/image-utils`. That package's `ensureImageOptionsAsync` (`build/Image.js:170`) validates against a hardcoded MIME table that only accepts `png/jpeg/webp/gif` — SVG isn't in the list. It throws. The favicon function's `try/catch` only handles `ENOENT`, so `Invalid mimeType` propagates through `Promise.all` in `exportApp.ts:283` and aborts the whole export.

Even if the rasterization *were* attempted, it would defeat the point of using SVG — the `prefers-color-scheme` media query inside the SVG would be evaluated at build time using whatever default the renderer picks, then frozen as pixels.

I have a draft fix ready — https://github.com/danfitz36/expo/compare/main...svg-favicon-support — spanning `@expo/cli` (`favicon.ts`, `FaviconMiddleware.ts`) and `@expo/router-server` (`createFaviconAsString`, `createFaviconAsNode`) since #46586 recently split the pipeline across those two packages. The change branches on `.svg` extension, copies the file raw, and emits `<link rel="icon" type="image/svg+xml">`. Format is inferred from the extension so `web.favicon` stays typed as `string` — no API change. Happy to open the PR proper once you've had a chance to weigh in on direction (e.g. whether you'd want an ICO fallback auto-generated alongside, whether a `{ svg, ico }` object shape would be preferred, etc.).

Note: this issue was drafted with Claude Code assistance and reviewed by me — flagging so you know.

### Environment

```
    System:
      OS: macOS 26.5.2
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 22.20.0 - /Users/danfitz/.nvm/versions/node/v22.20.0/bin/node
      Yarn: 1.22.22 - /Users/danfitz/.nvm/versions/node/v22.20.0/bin/yarn
      npm: 11.6.2 - /Users/danfitz/.nvm/versions/node/v22.20.0/bin/npm
      Watchman: 2026.07.06.00 - /opt/homebrew/bin/watchman
    Managers:
    SDKs:
      iOS SDK:
      Android SDK:
        API Levels: 31, 34, 36, 36
        Build Tools: 35.0.0, 36.0.0, 36.1.0, 37.0.0
        System Images: android-34 | Google APIs ARM 64 v8a
    IDEs:
      Android Studio: 2025.3 AI-253.32098.37.2534.15336583
    npmPackages:
      expo: ~56.0.8 => 56.0.16 
      react: 19.2.3 => 19.2.3 
      react-dom: 19.2.3 => 19.2.3 
    Expo Workflow: managed
```

### Expo Doctor Diagnostics

```
```

danfitz36:
Fixes #48024.

## Summary

Adds SVG support to the `web.favicon` pipeline. Setting `web.favicon` to an SVG previously crashed `expo export -p web` outright with `Invalid mimeType` from `@expo/image-utils`, because `getFaviconFromExpoConfigAsync` unconditionally rasterizes and that package's MIME table doesn't include SVG. The favicon `try/catch` only handles `ENOENT`, so the error propagated and killed the export, leaving `dist/` empty. This PR branches on the `.svg` extension in `@expo/cli` and copies the file raw — preserving features like `prefers-color-scheme` media queries inside the SVG.

`web.favicon` stays typed as `string`; format is inferred from the extension. PNG/ICO inputs are byte-identical to today.

Two things beyond the crash fix, both needed for SVG favicons to actually work end to end:

- **`expo start` rendered no favicon at all** for an SVG-configured project. The dev HTML carries no icon tag, and browsers auto-discover `/favicon.ico` but never `/favicon.svg` — so a `<link>` has to be emitted. This covers the default SPA dev HTML (`web.output: 'single'`) and the RSC dev path, which share the same template. Non-RSC `web.output: 'static' | 'server'` renders through a different dev SSR path that has never been passed a favicon href; I left that alone rather than widen the diff, happy to wire it up here if you'd prefer.
- **`favicon.ico` generation is preserved alongside a `public/favicon.svg`.** The public SVG takes the `<link>`, but a raster `web.favicon` still rasterizes, so the fallback older browsers auto-discover isn't silently dropped. Generation is skipped only when the public folder already ships a file at that same output path — otherwise a project holding both `public/favicon.svg` and `public/favicon.ico` would have its hand-crafted ICO overwritten by a generated one.

## Changes

Since #46586 split favicon rendering across two packages, the SVG-aware pieces live in both.

**`@expo/cli`**

- `src/export/favicon.ts` — detect `.svg` earlier in `getFaviconFromExpoConfigAsync` and copy the file raw (skip `generateImageAsync`, which rejects SVG). Extend `getUserDefinedFaviconFile` to also detect `public/favicon.svg`, and inject a `<link>` tag for it. New `getSvgFaviconHref` resolves the href the dev server needs without generating anything.
- `src/start/server/middleware/FaviconMiddleware.ts` — intercept both `/favicon.ico` and `/favicon.svg` in dev; pick the response based on the configured source; fall through on format mismatch. Classify on the query-stripped pathname, since `shouldHandleRequest` already dispatches on it and `/favicon.svg?v=2` reaches the handler with the raw URL.
- `src/start/server/middleware/ManifestMiddleware.ts` — emit the SVG favicon `<link>` in the dev server HTML.

**`@expo/router-server`**

- `src/utils/html.ts` — `createFaviconAsString` emits `type="image/svg+xml"` when the href points to an SVG, via a shared `isSvgFaviconHref` that ignores any query string or fragment.
- `src/utils/react.tsx` — `createFaviconAsNode` does the same via a React attribute, staying byte-equivalent to the string form (per the existing doc comment invariant).

## Design notes

- **API stays `string`.** Format-by-extension is the least invasive path — no `ExpoConfig` type change, no caller updates, no docs churn beyond mentioning that SVG is accepted. If a `{ svg, ico }` object shape is preferred, happy to add it — I chose the additive path first.
- **No auto-generated ICO fallback for SVG inputs.** Rasterizing the SVG would freeze its `prefers-color-scheme` media query at build time (defeating the point), and it would require bypassing the `@expo/image-utils` MIME check (touching a shared package). Modern browser support for SVG favicons is high enough that users needing an ICO fallback can drop a hand-crafted `public/favicon.ico` — which the auto-detect already picks up.
- **No `@expo/image-utils` changes.** The MIME-table gate stays intact; SVG bypasses the raster pipeline entirely in `@expo/cli`.

## Tests

- `favicon.test.ts` — the SVG config path, the `public/favicon.svg` auto-detect path, a check that `generateImageAsync` is *not* called on SVG input, and each public-folder/config combination that decides whether generation runs (including the both-public-files case that would otherwise clobber the user's ICO).
- `FaviconMiddleware.test.ts` — SVG serving with the right MIME, a cache-busted `/favicon.svg?v=2`, and format-mismatch fall-through in both directions.
- `ManifestMiddleware-test.ts` — the dev HTML carries the SVG `<link>`.
- `html.test.ts` / `react.test.node.tsx` — the SVG `type` attribute, plus a case comparing `createFaviconAsNode`'s static markup against `createFaviconAsString` across a range of hrefs, so the byte-equivalence invariant can't drift silently.

## Verification

Run locally on this branch:

- `packages/@expo/cli` — jest: 243 suites, 1940 passed / 7 skipped; `tsc --noEmit`: clean.
- `packages/@expo/router-server` — jest: 21 suites, 317 passed.
- `turbo lint format depscheck` for both packages — 0 warnings, 0 errors, no files rewritten.
- **Live end-to-end** (SDK 56 architecture, before this PR's TS-level changes): repro at [danfitz36/expo-svg-favicon-repro](https://github.com/danfitz36/expo-svg-favicon-repro) confirmed the pre-fix crash. Hand-patched the built favicon.js in `node_modules` to mirror the fix; `dist/favicon.svg` byte-identical to source, correct `<link rel="icon" type="image/svg+xml">` in HTML head, PNG path byte-identical to pre-fix output. The dev-server and public-folder behavior above is covered by unit tests, not by that manual run.

## Notes

This branch and issue were drafted with Claude Code assistance and reviewed by me — flagging so you know what you're looking at. The diff has also been reviewed by [@jhhayashi](https://github.com/jhhayashi), a senior engineer on my team.

### DIFF
--- a/packages/@expo/cli/CHANGELOG.md
+++ b/packages/@expo/cli/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ### 🎉 New features
 
+- Support SVG favicons via `web.favicon` and `public/favicon.svg` ([#48026](https://github.com/expo/expo/pull/48026) by [@danfitz36](https://github.com/danfitz36))
 - Add `pageHeaders` to exported routes manifests ([#47429](https://github.com/expo/expo/pull/47429) by [@hassankhan](https://github.com/hassankhan))
 - Apply `pageHeaders` when serving static exports with `expo serve` ([#47781](https://github.com/expo/expo/pull/47781) by [@hassankhan](https://github.com/hassankhan))
 - Create `pageHeaders` rules from loader-declared `Cache-Control` headers for SSG ([#47774](https://github.com/expo/expo/pull/47774) by [@hassankhan](https://github.com/hassankhan))
--- a/packages/@expo/cli/src/export/__tests__/favicon.test.ts
+++ b/packages/@expo/cli/src/export/__tests__/favicon.test.ts
@@ -33,6 +33,24 @@ describe(generateFaviconAssetAsync, () => {
     expect(result).toBeNull();
   });
 
+  it('returns the SVG href when the user supplied a `public/favicon.svg`', async () => {
+    // Browsers don't auto-discover SVG favicons the way they do
+    // `/favicon.ico`, so a `<link>` tag must still be emitted. The SVG file
+    // itself is copied to the output by `copyPublicFolderAsync`, not here.
+    getUserDefinedFile.mockReturnValueOnce('/project/public/favicon.svg');
+
+    const files: ExportAssetMap = new Map();
+    const result = await generateFaviconAssetAsync('/project', {
+      baseUrl: '/app',
+      outputDir: '/out',
+      files,
+      exp: { name: '', slug: '' } as any,
+    });
+
+    expect(result).toEqual({ href: '/app/favicon.svg' });
+    expect(files.size).toBe(0);
+  });
+
   it('returns `null` when `web.favicon` is not set', async () => {
     const result = await generateFaviconAssetAsync('/project', {
       baseUrl: '',
@@ -57,6 +75,37 @@ describe(generateFaviconAssetAsync, () => {
       contents: Buffer.from([1, 2, 3]),
       targetDomain: 'client',
     });
+  });
+
+  it('copies an SVG `web.favicon` raw (no rasterization) and returns the SVG href', async () => {
+    // Rasterizing would defeat the point of an SVG favicon — features like
+    // `prefers-color-scheme` media queries need the original markup to
+    // survive into the served asset.
+    const svgSource = '<svg xmlns="http://www.w3.org/2000/svg"/>';
+    const readFileMock = jest
+      .spyOn(require('fs').promises, 'readFile')
+      .mockResolvedValue(Buffer.from(svgSource));
+
+    try {
+      const files: ExportAssetMap = new Map();
+      const result = await generateFaviconAssetAsync('/project', {
+        baseUrl: '/app',
+        outputDir: '/out',
+        files,
+        exp: { name: '', slug: '', web: { favicon: './icon.svg' } } as any,
+      });
+
+      expect(result).toEqual({ href: '/app/favicon.svg' });
+      expect(files.get('favicon.svg')).toEqual({
+        contents: Buffer.from(svgSource),
+        targetDomain: 'client',
+      });
+      // The raster pipeline must not have been invoked for the SVG input.
+      const { generateImageAsync } = jest.requireMock('@expo/image-utils');
+      expect(generateImageAsync).not.toHaveBeenCalled();
+    } finally {
+      readFileMock.mockRestore();
+    }
   });
 
   it('writes to disk when no asset map is provided', async () => {
--- a/packages/@expo/cli/src/export/__tests__/favicon.test.ts
+++ b/packages/@expo/cli/src/export/__tests__/favicon.test.ts
@@ -54,6 +54,50 @@ describe(generateFaviconAssetAsync, () => {
     expect(generateImageAsync).not.toHaveBeenCalled();
   });
 
+  it('still generates favicon.ico when a public SVG is paired with a raster web.favicon', async () => {
+    // A project that has both used to get a generated `favicon.ico`. The public SVG takes the
+    // `<link>`, but dropping the ICO would silently remove the fallback older browsers
+    // auto-discover at `/favicon.ico`.
+    getUserDefinedFile.mockReturnValueOnce('/project/public/favicon.svg');
+
+    const files: ExportAssetMap = new Map();
+    const result = await generateFaviconAssetAsync('/project', {
+      baseUrl: '',
+      outputDir: '/out',
+      files,
+      exp: { name: '', slug: '', web: { favicon: './icon.png' } } as any,
+    });
+
+    expect(result).toEqual({ href: '/favicon.svg' });
+    expect(files.get('favicon.ico')).toEqual({
+      contents: Buffer.from([1, 2, 3]),
+      targetDomain: 'client',
+    });
+  });
+
+  it('does not double-write favicon.svg when a public SVG is paired with an SVG web.favicon', async () => {
+    getUserDefinedFile.mockReturnValueOnce('/project/public/favicon.svg');
+    const readFileMock = jest
+      .spyOn(require('fs').promises, 'readFile')
+      .mockResolvedValue(Buffer.from('<svg/>'));
+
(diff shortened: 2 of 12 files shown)


danfitz36:
On the `ai-generated` label — some context so you know what you're looking at:

I've been driving this with Claude Code but reviewing every change, and the current diff has also been reviewed by Bizzen engineers. A few other artifacts a skeptical reader can check independently:

- The patch is written against the post-refactor architecture from #46586 (the split across `@expo/cli` and `@expo/router-server`) — not machine-translated from the pre-refactor shape. The invariant that `createFaviconAsString` and `createFaviconAsNode` stay byte-equivalent is honored, with a dedicated regression test on each side.
- Tests live on both sides of the split, plus a "raster path was not invoked" assertion for the SVG branch and a preserved-byte assertion for the SVG source.
- PNG/JPEG hot path is byte-identical — verifiable by diffing the emitted `dist/favicon.ico` from a scaffold before and after.
- End-to-end reproduction verified against a linked minimal repro (referenced from the linked issue #48024), not just the unit tests.

Some context on why this might be worth prioritizing (also on the linked issue):

1. It's a **crash**, not a missing feature — `expo export -p web` aborts entirely and leaves `dist/` empty. Users hit a build that silently fails instead of a warning.
2. Same `@expo/image-utils` `Invalid mimeType` pattern crashed a different pipeline in #40031 (iOS icon composer). The class of bug isn't unique to favicons.
3. Sidesteps [expo/expo-cli#2012](https://github.com/expo/expo-cli/issues/2012) (blurry favicon rasterization), since SVG inputs no longer round-trip through `sharp`.

Happy to walk through any specific piece if it'd help calibrate.

PR #27390, opened 2024-03-01
[fingerprint] Fix inconsistent hashes for autolinking

wschurman:
# Why

Fingerprint hashes in my test project weren't match when run through EAS build. This PR fixes multiple reasons.

# How

- Add `.cxx` folder in any android folder to ignores. It seems that this is sometimes present elsewhere.
- Add `android-annotation` and `android-annotation-processor` build/.cxx folders.
- Add `*-gradle-plugin` source and pertinent ignores. The gradle plugin is sourced from autolinking config and needs to be hashed.
- Normalize `module.plugins.sourceDir` in autolinking config before hashing contents, similar to how we normalize `module.projects.sourceDir` right above it.
- Fix default ignore dirs for managed projects in updates runtime version calculation. These were having issues with different root dirs.

# Test Plan

This is a bit convoluted to test e2e but the gist is to set up a project that uses fingerprint runtime version policy, do a local build with a local eas-build repo and eas-cli repo and verify that when `npx expo-updates runtimeversion:resolve` is run within the project dir it is the same as when run within the `build` intermediate `eas-build` working directory (with the patches from this PR added in to the node_modules manually).

# Checklist

- [ ] Documentation is up to date to reflect these changes (eg: https://docs.expo.dev and README.md).
- [ ] Conforms with the [Documentation Writing Style Guide](https://github.com/expo/expo/blob/main/guides/Expo%20Documentation%20Writing%20Style%20Guide.md)
- [ ] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: updated a module plugin).

PR #40761, opened 2025-10-31
[fingerprint] sort autolinking config

Kudo:
# Why

trying to implement the recursive sorting for autolinking config. that helps to defense the assumption that autolinking config should be in stable order

# How

- added a recursive sorting function `sortConfig`

# Test Plan

- add unit tests
- tried to benchmark the performance impact in bare-expo

before
```
$ hyperfine '../../packages/@expo/fingerprint/bin/cli.js fingerprint:generate'
Benchmark 1: ../../packages/@expo/fingerprint/bin/cli.js fingerprint:generate
  Time (mean ± σ):     76.908 s ±  0.684 s    [User: 51.401 s, System: 107.371 s]
  Range (min … max):   75.977 s … 78.447 s    10 runs
```

after
```
Benchmark 1: ../../packages/@expo/fingerprint/bin/cli.js fingerprint:generate
  Time (mean ± σ):     79.139 s ±  3.107 s    [User: 52.745 s, System: 107.846 s]
  Range (min … max):   76.298 s … 85.064 s    10 runs
```

sounds okay!? at least it's not a bottleneck for the whole fingerprinting

# Checklist

- [x] I added a `changelog.md` entry and rebuilt the package sources according to [this short guide](https://github.com/expo/expo/blob/main/CONTRIBUTING.md#-before-submitting)
- [x] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: updated a module plugin).
- [x] Conforms with the [Documentation Writing Style Guide](https://github.com/expo/expo/blob/main/guides/Expo%20Documentation%20Writing%20Style%20Guide.md)

PR #48027, opened 2026-07-21
[fingerprint] Fix non-deterministic hash from autolinking dependency order

Oupsla:
# Why

React Native core autolinking returns its `dependencies` object in resolution order, which is not stable across runs when a project has local `file:` dependencies. `parseCoreAutolinkingSourcesAsync` serializes that object with a raw `JSON.stringify`, so the key order leaks into the hash and the project fingerprint changes between runs with no source change — which breaks `runtimeVersion: { policy: "fingerprint" }` and prevents EAS build reuse.

Ten consecutive `createFingerprintAsync()` calls on an unchanged tree, `@expo/fingerprint@0.20.5`:

```
runs=10  distinctHashes=4  distinctOrders=2
```

The unstable sources are `rncoreAutolinkingConfig:android` and `rncoreAutolinkingConfig:ios`. Same key set on every run, different key **order**, and the only packages that move are the project's two local `file:` dependencies:

```
run1  react-native-nitro-toast-local | sce-module | @intercom/intercom-react-native
run3  sce-module | react-native-nitro-toast-local | @intercom/intercom-react-native
```

Two orders on android × two on ios = the four distinct fingerprints. The per-dependency sources are unaffected because `Fingerprint.ts` sorts sources via `sortSources()`; only the `contents` blob leaks the ordering.

This came out of #47943 — thanks @Kudo for the counter-repro there, which ruled out the CNG theory I originally filed and led here.

# How

Serialize the autolinking config with `stringifyJsonSorted` instead of `JSON.stringify`. The helper already exists in `sourcer/Utils.ts` for exactly this purpose and is already used for the Expo config in `Expo.ts` (added in #27390, "Fix inconsistent hashes for autolinking"); `Bare.ts` was simply never converted.

# Test Plan

- Added a unit test asserting the emitted `contents` are identical when the autolinking `dependencies` are returned in reverse order. It runs for all three sourcers (`getCoreAutolinkingSourcesFromExpoAndroid`, `...FromExpoIos`, `...FromRncCliAsync`) and fails on `main`.
- Updated the five affected `rncoreAutolinkingConfig` snapshot entries (3 in `Bare-test`, 2 in the `managed` e2e snapshot). Each was re-serialized with the same `stringifyJsonSorted` implementation, verified to round-trip byte-for-byte before sorting, to preserve the parsed value, and to be idempotent.
- On the affected project, this takes the fingerprint from 4 distinct hashes over 10 runs to 12/12 identical.

Two notes, in the interest of full disclosure:

1. I could **not** reduce this to a minimal repro on a stock app — a `create-expo-app` with two synthetic `file:` modules stayed stable, so it appears to need real local native modules (ours are two Nitro/Expo modules with podspecs and android sources). The measurements above are from a real project.
2. `Expo.ts:387` and `Expo.ts:450` serialize `expoAutolinkingConfig:{android,ios}` with the same raw `JSON.stringify` pattern. I left them alone since I only measured instability in the rncore path, but I'm happy to extend this PR if you'd like them covered too.

I was not able to run the fingerprint test suite locally (installing the workspace deps for this package alone cascades through most of the monorepo), so the snapshot updates are the part most worth a careful look.


Kudo:
hi there! sorry being late response from my time off. do you have repro for the unstable autolinking result? i had this poc but was hesitant to introduce it for speed https://github.com/expo/expo/pull/40761

Oupsla:
Hi @Kudo , no worries, hope you get a good time off.

Yes — minimal repro: https://github.com/Oupsla/expo-autolinking-order-repro

Stock create-expo-app (SDK 57 / RN 0.86.2) plus four local RN modules under modules/, nothing else. npm install && npm run check:fingerprint gives 10 distinct fingerprints over 10 runs on an unchanged tree. npm run check:order isolates the autolinking command with no fingerprint involved (20 fresh processes → 4–7 distinct dependency orders).

-------

One thing I'd want your read on: sortConfig sorts arrays as well as keys. Since fingerprint should change when the config changes, sorting scriptPhases / buildTypes / componentDescriptors means two configs differing only in array order hash the same. I haven't found a case where that order carries meaning, so it may be finen but it's a false-negative risk key-sorting doesn't have.

Also: I offered above to extend #48027 to Expo.ts:387/450, I'd retract that. resolveModulesAsync already sorts modules by packageName, so expoAutolinkingConfig:* is order-stable, and the repro confirms only the rncore* sources move.

Happy to open a PR against expo-modules-autolinking for the scanning.ts change, or leave it with you, and equally happy to close #48027 if you'd rather the fix live in autolinking. Your call.

brentvatne:
@expo-bot check

Kudo:
thanks for this contribution. i'll try to resolve it in autolinking first. i still assume autolinking to return stable result.
superseded by #48629

PR #48034, opened 2026-07-22
[metro-config] Resolve polyfills via @react-native/js-polyfills when rn-get-polyfills subpath is gone

alanjhughes:
# Why
React Native's nightly (0.88) removed the `react-native/rn-get-polyfills` subpath in [facebook/react-native#57482](https://github.com/facebook/react-native/pull/57482). metro-configs `getPolyfills `required that subpath directly, so bundling against RN 0.88 fails.

# How

getPolyfills now tries the legacy react-native/rn-get-polyfills subpath first, then falls back to resolving @react-native/js-polyfills from the same React Native host path. Behavior on RN 0.86 and earlier is unchanged, and RN 0.88 uses the fallback.

# Test Plan

- RN 0.86 resolves via the legacy rn-get-polyfills and returns [console.js, error-guard.js].
- A simulated 0.88 layout (subpath absent, @react-native/js-polyfills reachable) falls back and returns the identical set.

### DIFF
--- a/packages/@expo/metro-config/src/ExpoMetroConfig.ts
+++ b/packages/@expo/metro-config/src/ExpoMetroConfig.ts
@@ -386,9 +386,14 @@ export function getDefaultConfig(
           return [];
         }
 
-        return require(
-          path.join(getReactNativeHostPath(projectRoot, platform), 'rn-get-polyfills')
-        )();
+        const reactNativeHostPath = getReactNativeHostPath(projectRoot, platform);
+        const polyfillsModule =
+          resolveFrom.silent(reactNativeHostPath, './rn-get-polyfills') ??
+          resolveFrom.silent(reactNativeHostPath, '@react-native/js-polyfills');
+        if (!polyfillsModule) {
+          return [];
+        }
+        return require(polyfillsModule)();
       },
     },
     server: {
--- a/packages/@expo/metro-config/src/ExpoMetroConfig.ts
+++ b/packages/@expo/metro-config/src/ExpoMetroConfig.ts
@@ -386,14 +386,7 @@ export function getDefaultConfig(
           return [];
         }
 
-        const reactNativeHostPath = getReactNativeHostPath(projectRoot, platform);
-        const polyfillsModule =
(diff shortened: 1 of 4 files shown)


packages/@expo/metro-config/src/ExpoMetroConfig.ts
```diff
@@ -386,9 +386,14 @@ export function getDefaultConfig(
           return [];
         }
 
-        return require(
-          path.join(getReactNativeHostPath(projectRoot, platform), 'rn-get-polyfills')
-        )();
+        const reactNativeHostPath = getReactNativeHostPath(projectRoot, platform);
+        const polyfillsModule =
+          resolveFrom.silent(reactNativeHostPath, './rn-get-polyfills') ??
```
kitten: I'd say, the resolution here would be more predictable against `resolveFrom` in `@expo/require-utils`, since we bypass that re-export here. This isn't something we migrated to yet, but technically more correct.

To go further, it might even be more correct to still do the join and do the resolution first, since with `package.json:exports` that can sometimes matter. By convention it doesn't here, since react-native won't use `exports` for this, but it's still a bit cleaner

i.e. introduce a sub-path for `rn-get-polyfills` on `getReactNativeHostPath` potentially

packages/@expo/metro-config/src/ExpoMetroConfig.ts
```diff
@@ -386,9 +386,14 @@ export function getDefaultConfig(
...
+          resolveFrom.silent(reactNativeHostPath, '@react-native/js-polyfills');
```
kitten: more importantly than the above, I'm not sure what the react-native change achieves here. Basically, the dependency chain has been broken and I'd expect that there's a replacement entrypoint to `@react-native/js-polyfills`.

Basically, either we can get the dependency from React Native, or have a dependency ourselves. However, both stands in conflict with OOT if there's no file like `rn-get-polyfills`.

Maybe that does signal we should ditch and replace `@react-native/js-polyfills` entirely (it's not a large package) but it feels like there's a larger discussion/alignment piece that's missing here

kitten (approved):

PR #48047, opened 2026-07-22
[docs] document build and update job hooks and defaults.hooks

hSATAC:
# Why

EAS Workflows hooks got several additions that the public docs don't cover yet: `before_update` / `after_update` on the update job, install hooks now running on build jobs, and the long-undocumented `defaults.hooks`. ENG-22616

# How

Listing the changes separately since each one can be dropped independently:

1. **Update job hooks** (`before_update` / `after_update`), with an example that uploads sourcemaps after publishing using the built-in `eas/posthog_upload_sourcemaps` function. Sourcemaps only exist on the worker that published the update, which is what the hook is for.
2. **Build job hooks** (`before_install_node_modules` / `after_install_node_modules`) — managed builds now run these.
3. **`defaults.hooks`**, first-time docs: broadcasts to every job (each job runs the keys that apply to it and ignores the rest), a job-level key overrides the default, `[]` opts a job out.
4. **Complete per-job hook lists for all nine hook-capable jobs** (from review feedback): new Hooks sections for deploy, fingerprint, repack, and testflight (checkout and install hooks; the testflight `asc_build_id` variant doesn't support hooks and says so); checkout and install hooks added to submit and update; `after_checkout` added to maestro and maestro-cloud. All of these keys already fire today, the docs just never listed them. The build job intentionally lists only the install pair: managed native builds don't wrap the checkout phase, so `after_checkout` wouldn't fire there.
5. **The hooks overview sentence in the syntax reference now names all nine supported jobs** instead of a three-job example list.

# Test Plan

Docs lint (including Vale prose lint) passes.

# Checklist

### DIFF
--- a/docs/pages/eas/workflows/pre-packaged-jobs.mdx
+++ b/docs/pages/eas/workflows/pre-packaged-jobs.mdx
@@ -38,6 +38,9 @@ jobs:
       profile: string # optional - default: production
       message: string # optional
       refresh_ad_hoc_provisioning_profile: boolean # optional
+    hooks:
+      before_install_node_modules: step[] # optional - steps to run before dependencies are installed.
+      after_install_node_modules: step[] # optional - steps to run after dependencies are installed.
 ```
 
 #### Parameters
@@ -74,6 +77,15 @@ You can reference the following outputs in subsequent jobs:
 | runtime_version   | string | The runtime version used.                                          |
 | sdk_version       | string | The SDK version used.                                              |
 | simulator         | string | Whether the build is for simulator.                                |
+
+#### Hooks
+
+The following hooks are supported for the Build job:
+
+- `before_install_node_modules`: steps to run before the build installs your project's dependencies.
+- `after_install_node_modules`: steps to run after the build installs your project's dependencies.
+
+See [`jobs.<job_id>.hooks`](/eas/workflows/syntax/#jobsjob_idhooks) for the general hooks syntax.
 
 ### Examples
 
@@ -905,6 +917,9 @@ jobs:
       rollout_percentage: number # optional - 0 to 100, defaults to 100
       private_key_path: string # optional
       upload_sentry_sourcemaps: boolean # optional - defaults to "try uploading, but don't fail the job if it fails"
+    hooks:
+      before_update: step[] # optional - steps to run before the update is published.
+      after_update: step[] # optional - steps to run after the update is published.
 ```
 
 #### Environment variables
@@ -933,6 +948,15 @@ You can reference the following outputs in subsequent jobs:
 | --------------------- | ------ | ------------------------------------------------------------- |
 | first_update_group_id | string | The ID of the first update group.                             |
 | updates_json          | string | A JSON string containing information about all update groups. |
+
+#### Hooks
+
+The following hooks are supported for the Update job:
+
+- `before_update`: steps to run before the update is published.
+- `after_update`: steps to run after the update is published. Files produced during publishing (for example, sourcemaps in the **dist** directory) are available to the hook steps.
+
+See [`jobs.<job_id>.hooks`](/eas/workflows/syntax/#jobsjob_idhooks) for the general hooks syntax.
 
 ### Examples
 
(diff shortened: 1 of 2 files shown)


docs/pages/eas/workflows/syntax.mdx
```diff
@@ -480,7 +480,7 @@ jobs:
 
 ### `jobs.<job_id>.hooks`
 
-Some pre-packaged jobs support hooks to run additional steps before or after the main job action (for example, `maestro-cloud`, `maestro`, and `submit`). Hook names are job-specific. See the job's documentation for the available hook keys.
+Some pre-packaged jobs support hooks to run additional steps before or after the main job action (for example, `maestro-cloud`, `maestro`, and `submit`). Hook names are job-specific. See the job's documentation for the available hook keys. To set default hooks for all jobs in the workflow, use [`defaults.hooks`](#defaultshooks).
```
amandeepmittal: **[suggestion]** This PR documents hooks for the build and update jobs, but the example list on this line still names only the three jobs that supported hooks before (`style-guide.md#core-principles`). Naming the newly documented jobs keeps the example list current for the reader who lands here first.

```suggestion
Some pre-packaged jobs support hooks to run additional steps before or after the main job action (for example, `build`, `update`, `maestro-cloud`, `maestro`, and `submit`). Hook names are job-specific. See the job's documentation for the available hook keys. To set default hooks for all jobs in the workflow, use [`defaults.hooks`](#defaultshooks).
```

docs/pages/eas/workflows/syntax.mdx
```diff
@@ -518,6 +518,28 @@ jobs:
...
+
+jobs:
+  build_app:
+    type: build
+    params:
+      platform: ios
+  publish_update:
+    type: update
+    hooks:
+      before_install_node_modules: [] # opt this job out of the default hook
```
amandeepmittal, line 540: **[suggestion]** The opt-out example contradicts the section's own rules (`style-guide.md#core-principles`). The prose above says each job "runs the hook keys that apply to it and ignores the rest", and the update job's documented hooks are only `before_update` and `after_update`. As written, `publish_update` opts out of a hook key that an update job would ignore anyway, so the `[]` reads as a no-op and implies update jobs run install hooks. Demonstrating the opt-out with a second build job shows a case where the key actually applies. If update jobs do run the install hooks, add them to the update job's hooks documentation instead.

```suggestion
  build_preview:
    type: build
    params:
      platform: android
    hooks:
      before_install_node_modules: [] # opt this job out of the default hook
```

docs/pages/eas/workflows/pre-packaged-jobs.mdx
```diff
@@ -1010,6 +1034,31 @@ jobs:
 
 </Collapsible>
 
+<Collapsible summary="Upload sourcemaps after publishing an update">
+
+This workflow publishes an update and then uploads the generated sourcemaps to an error tracking service (PostHog in this example). Publishing an update creates sourcemaps in the **dist** directory. These files only exist on the worker that runs the job, so the upload must happen in an `after_update` hook rather than in a separate job. Hook steps inherit the job's environment, so you can provide credentials such as `POSTHOG_CLI_API_KEY` as [EAS environment variables](/eas/environment-variables/).
```
amandeepmittal: **[suggestion]** PostHog's CLI needs both `POSTHOG_CLI_API_KEY` and `POSTHOG_CLI_PROJECT_ID` set in CI, per PostHog's [React Native sourcemap docs](https://posthog.com/docs/error-tracking/upload-source-maps/react-native), so a reader who copies this example and sets only the named variable gets an authentication failure (`style-guide.md#core-principles`). Naming both keeps the example self-sufficient.

```suggestion
This workflow publishes an update and then uploads the generated sourcemaps to an error tracking service (PostHog in this example). Publishing an update creates sourcemaps in the **dist** directory. These files only exist on the worker that runs the job, so the upload must happen in an `after_update` hook rather than in a separate job. Hook steps inherit the job's environment, so you can provide credentials such as `POSTHOG_CLI_API_KEY` and `POSTHOG_CLI_PROJECT_ID` as [EAS environment variables](/eas/environment-variables/).
```

docs/pages/eas/workflows/syntax.mdx
hSATAC: Went one step further and made it the full list: the sentence now names all nine jobs that support hooks, since this PR now documents hooks for each of them (see the other thread). Dropped the "for example" phrasing accordingly.

docs/pages/eas/workflows/syntax.mdx
hSATAC: You caught a real gap here, one level deeper than it looks: update jobs do run the install hooks (the job checks out the project and installs dependencies before publishing), and the same goes for several other job types. The docs just never said so. So instead of swapping the example, I added the missing hook lists to every hook-capable job (deploy, fingerprint, repack, testflight, submit, update, maestro, maestro-cloud). The example is consistent as written now: the key applies to both jobs shown, and the update job opts out. Thanks for the nudge, it reminded me we had skipped documenting these lists.

docs/pages/eas/workflows/pre-packaged-jobs.mdx
hSATAC: Right, thanks. Added `POSTHOG_CLI_PROJECT_ID`. Double-checking the PostHog specifics also reminded me that we ship a built-in `eas/posthog_upload_sourcemaps` function that reads both variables from the environment and defaults to the **dist** directory, so the example now uses it instead of a raw CLI call.

amandeepmittal (approved):

sjchmiela (approved):

PR #48052, opened 2026-07-22
[sharing][ios] Respect passed UTI/mimeType

### LINKED ISSUE #44887: [expo-sharing] You don't have access to the provided file

https://github.com/SpasiboKojima/expo-sharing-test

### Steps to reproduce

1. Load remote image with `Image.loadAsync`
2. Get its cache path with `Image.getCachePathAsync`
3. Prepend 'file://' and append `.${fileExtension}`
4. Send it to `Sharing.shareAsync`
5.  [Error: Calling the 'shareAsync' function has failed
→ Caused by: You don't have access to the provided file]

Tried it multiple ways (there's another, different path retrieved through File), it never works, and there's no comprehensible example anywhere. Am I doing something wrong?

### Environment

```text
    System:
      OS: Linux 6.6 Ubuntu 24.04.3 LTS 24.04.3 LTS (Noble Numbat)
      Shell: 5.2.21 - /bin/bash
    Binaries:
      Node: 22.19.0 - ~/.nvm/versions/node/v22.19.0/bin/node
      npm: 10.9.3 - ~/.nvm/versions/node/v22.19.0/bin/npm
    npmPackages:
      expo: ~55.0.12 => 55.0.12
      expo-router: ~55.0.11 => 55.0.11
      react: 19.2.0 => 19.2.0
      react-dom: 19.2.0 => 19.2.0
      react-native: 0.83.4 => 0.83.4
    npmGlobalPackages:
      eas-cli: 18.5.0
    Expo Workflow: manage
```

### Expo Doctor Diagnostics

```text
17/17 checks passed. No issues detected!
```

behenate:
# Why

Fixes https://github.com/expo/expo/issues/44887
On iOS, UTI and mimeType are ignored when sharing files without the expected extension, causing an incorrect type or preview.

# How

Resolve the declared UTI, falling back to mimeType, and share a temporary hard link with the preferred extension. Remove the link after the share sheet closes.

# Test Plan

Tested in BareExpo on iOS


packages/expo-sharing/src/Sharing.types.ts
```diff
@@ -1,13 +1,16 @@
 // @needsAudit
 export type SharingOptions = {
   /**
-   * Sets `mimeType` for `Intent`.
+   * The MIME type of the target file, for example `image/jpeg`.
+   * On Android it sets the `mimeType` of the sharing `Intent`. On iOS it determines
```
amandeepmittal:
```suggestion
   * On Android, it sets the `mimeType` of the sharing `Intent`. On iOS, it determines
```

minor nit.

packages/expo-sharing/ios/SharingModule.swift
```diff
@@ -91,6 +104,46 @@ public final class SharingModule: Module {
...
+    }
+
+    guard let ext = declaredContentType(options)?.preferredFilenameExtension else {
+      return url
+    }
+    if url.pathExtension.caseInsensitiveCompare(ext) == .orderedSame {
+      return url
+    }
+
+    let baseName = url.deletingPathExtension().lastPathComponent
```
tsapeta, line 128: [🏗️ **design**] The link name is derived purely from the source basename, so it isn't unique. Two concerns follow from that:

1. **Cross-share clobbering.** Cached filenames are frequently non-unique in practice (`image.jpg`, or a hash reused across directories). If share A is still on screen and share B resolves to the same `linkURL`, the `removeItem` below deletes A's link out from under the live `UIActivityViewController`. The data survives (the original file still holds the inode), but the *name* A handed to the activity controller is gone, so any extension that reads the URL lazily gets ENOENT.
2. **Stale-link reuse.** If a share is dismissed in a way that skips the completion handler, the leftover link sits in `tmp/` until iOS purges it. A later share of a *different* file with the same basename hits the `fileExists` branch and deletes it, which is correct but only by accident.

Suggest a `UUID` in the path, in a dedicated subdirectory so cleanup is greppable. This also lets you drop the `fileExists`/`removeItem` dance entirely, since a fresh UUID never collides:

```swift
let baseName = url.deletingPathExtension().lastPathComponent
let linkDirectory = FileManager.default.temporaryDirectory
  .appendingPathComponent("expo-sharing", isDirectory: true)
  .appendingPathComponent(UUID().uuidString, isDirectory: true)
let linkURL = linkDirectory
  .appendingPathComponent(baseName.isEmpty ? "expo-sharing-item" : baseName)
  .appendingPathExtension(ext)

do {
  try FileManager.default.createDirectory(at: linkDirectory, withIntermediateDirectories: true)
  try FileManager.default.linkItem(at: url, to: linkURL)
  return linkURL
} catch {
  return url
}
```

The per-share subdirectory keeps the user-visible filename intact (share sheets and Files.app show `baseName.ext`, not the UUID), which a `UUID`-in-the-filename approach would lose. Cleanup then removes `linkDirectory` rather than the file.

For reference, `expo-media-library` (`UriExtractor.swift:56`) and `expo-audio` (`AudioUtils.swift:146`) both use `temporaryDirectory` + `UUID().uuidString` for exactly this reason.

packages/expo-sharing/ios/SharingModule.swift
```diff
@@ -18,10 +19,22 @@ public final class SharingModule: Module {
...
+      // the same on-disk data, so this costs no copy and behaves as a real file
+      // for every consumer.
+      let itemURL = shareableURL(for: url, options: options)
+      let createdLink = itemURL != url
+
+      let activityController = UIActivityViewController(activityItems: [itemURL], applicationActivities: nil)
       activityController.title = options.dialogTitle
 
       activityController.completionWithItemsHandler = { _, _, _, _ in
+        if createdLink {
```
tsapeta: [💡 suggestion] The link is only removed from `completionWithItemsHandler`, but two paths below can exit the function after the link has already been created:

- `MissingCurrentViewControllerException` is thrown when `currentViewController()` is nil (line 47).
- `present(_:animated:)` can fail to actually present (for example, another modal is already up), in which case the completion handler never fires.

In both cases the link leaks into `tmp/` until iOS purges it. Cheap fix: move link creation to *after* the `currentViewController` guard, and clean up in a `catch`/`defer` on the throwing path. With the per-share subdirectory suggested below, cleanup becomes `try? FileManager.default.removeItem(at: linkURL.deletingLastPathComponent())`.

Not data loss (the original file is untouched — that's the nice property of a hard link), so this is a hygiene issue rather than a correctness one, but it's easy to get right.

packages/expo-sharing/ios/SharingModule.swift
```diff
@@ -91,6 +104,46 @@ public final class SharingModule: Module {
     }
   }
 
+  private func declaredContentType(_ options: SharingOptions) -> UTType? {
+    if let uti = options.UTI, let type = UTType(uti) {
```
tsapeta, line 105: [🏗️ **design**] `UTType(_:)` resolves against types **registered on the device** — it returns nil for a syntactically valid UTI that no installed app declares. I verified this:

```
public.jpeg         -> public.jpeg  (ext: jpeg)
com.example.custom  -> nil
public.data         -> public.data  (ext: nil)
```

So when a caller passes a custom/vendor UTI that isn't declared anywhere on the device, `declaredContentType` silently falls through to `mimeType` — and if `mimeType` is absent, `shareableURL` returns the original URL and the fix no-ops. That directly contradicts the docs change in this PR ("Takes precedence over `mimeType` when both are provided"), which reads as an unconditional guarantee.

Two options, either is fine:
- Use `UTType(uti) ?? UTType(exportedAs: uti)` so an undeclared-but-well-formed UTI still resolves (note `exportedAs:` traps on a malformed identifier, so guard it with `UTType(filenameExtension:)`-style validation or keep the plain `UTType(uti)` first).
- Or leave the behavior and soften the docs to note that iOS only honors UTIs it can resolve.

Also worth noting for the docs: `preferredFilenameExtension` is nil for abstract types like `public.data` (`application/octet-stream`), so those correctly no-op — that part is fine as written.

packages/expo-sharing/ios/SharingModule.swift
```diff
@@ -91,6 +104,46 @@ public final class SharingModule: Module {
...
+    let baseName = url.deletingPathExtension().lastPathComponent
+    let linkURL = FileManager.default.temporaryDirectory
+      .appendingPathComponent(baseName.isEmpty ? "expo-sharing-item" : baseName)
+      .appendingPathExtension(ext)
+
+    do {
+      if FileManager.default.fileExists(atPath: linkURL.path) {
+        try FileManager.default.removeItem(at: linkURL)
+      }
+      try FileManager.default.linkItem(at: url, to: linkURL)
```
tsapeta: [🐧 nit] `linkItem` fails with `EXDEV` when source and destination are on different filesystems. On iOS the app sandbox and `tmp/` are normally the same volume, so this is rare — but it is reachable for files vended from an App Group container or an iCloud/`NSFileProvider`-backed URL, and `shareAsync` accepts any readable URL.

The `catch` correctly falls back to the original URL, so behavior degrades to today's (no extension → wrong type), which is acceptable. Just flagging that a `copyItem` fallback inside the `catch` would make the fix work for those cases too, at the cost of an actual copy. Your call whether that's worth it; the current behavior is not a regression either way.

packages/expo-sharing/ios/SharingModule.swift
```diff
@@ -91,6 +104,46 @@ public final class SharingModule: Module {
...
+    }
+    if let mimeType = options.mimeType, let type = UTType(mimeType: mimeType) {
+      return type
+    }
+    return nil
+  }
+
+  private func shareableURL(for url: URL, options: SharingOptions) -> URL {
+    var isDirectory: ObjCBool = false
+    FileManager.default.fileExists(atPath: url.path, isDirectory: &isDirectory)
```
tsapeta: [🐧 nit] Minor: the return value of `fileExists(atPath:isDirectory:)` is discarded, so a **nonexistent** path and an existing **file** are treated identically (both leave `isDirectory` false and fall through to the link attempt). That's harmless here — `linkItem` throws for a missing source and the `catch` returns the original URL, which then fails the same way it does today.

Still, checking the return value makes the intent explicit and lets you skip the pointless link attempt:

```swift
var isDirectory: ObjCBool = false
guard FileManager.default.fileExists(atPath: url.path, isDirectory: &isDirectory), !isDirectory.boolValue else {
  return url
}
```

Note this runs after the `isReadableFile` gate, which already requires the file to exist — so this is about clarity, not a real bug.

packages/expo-sharing/src/Sharing.types.ts
```diff
@@ -1,13 +1,16 @@
 // @needsAudit
 export type SharingOptions = {
   /**
-   * Sets `mimeType` for `Intent`.
+   * The MIME type of the target file, for example `image/jpeg`.
```
tsapeta, line 4: [💡 suggestion] Two small doc points:

1. The original text ("Sets `mimeType` for `Intent`") is preserved as the Android sentence, which is right, but the field is still `@needsAudit` at the top of the type and now describes genuinely different semantics per platform. Consider a blockquote note that on iOS the mimeType is used *only* to derive a filename extension for the share sheet — it isn't attached to the item as metadata, which is what an Android reader would assume from the Intent phrasing.

2. "Takes precedence over `mimeType` when both are provided" is only true when the UTI actually resolves on the device (see the comment on `declaredContentType`). If you go with the docs-side fix there, this is the line to qualify.

Also worth confirming: `docs/public/static/data/unversioned/expo-sharing.json` is regenerated output — make sure it was produced by the docs generator rather than hand-edited, so it doesn't drift on the next regen.

amandeepmittal (approved):

packages/expo-sharing/ios/SharingModule.swift
```diff
@@ -18,42 +19,51 @@ public final class SharingModule: Module {
...
-          width: rect?.width ?? 0,
-          height: rect?.height ?? 0
-        )
-        activityController.popoverPresentationController?.sourceView = currentViewcontroller.view
-        activityController.modalPresentationStyle = .pageSheet
-      }
+        activityController.completionWithItemsHandler = { _, _, _, _ in
+          // Resolve unconditionally. UIActivityViewController invokes this once
+          // on dismissal for every (activityType, completed) permutation.
+          session.resolve()
```
tsapeta, line 45: [🚨 **critical**] `completionWithItemsHandler` fires when the sheet is **dismissed**, not when the receiving activity has finished reading the item. Several system activities keep reading the source URL after dismissal (Mail composing in the background, Files "Save to", AirDrop transfers). Since `session.resolve()` calls `cleanup()`, which does `removeItem(at: stagedDirectory)`, the staged link can be deleted out from under an in-flight consumer, producing a truncated or failed share, exactly the class of bug this PR sets out to fix.

The hard link makes this less severe than it sounds: the inode survives while any consumer holds an open file descriptor. But activities that resolve the path lazily rather than opening it eagerly (`NSItemProvider` handing over a URL, out-of-process extensions that re-`open()` by path) will see `ENOENT`.

Worth noting no other package in this repo deletes a staged file in a dismissal/completion handler; the prevailing pattern is to stage into a per-module cache subdir with a UUID name and let it be reclaimed later. Two options:

- Drop the eager delete and instead sweep stale `expo-sharing` staging dirs on next module init. Deterministic-enough, zero race.
- Keep the delete but defer it well past dismissal (e.g. a delayed cleanup), accepting it's best-effort.

Either way this deserves a comment explaining the chosen tradeoff, since the timing is subtle.

packages/expo-sharing/ios/SharingModule.swift
```diff
@@ -91,6 +101,55 @@ public final class SharingModule: Module {
     }
   }
 
+  private func declaredContentType(_ options: SharingOptions) -> UTType? {
```
tsapeta, line 104: [🏗️ **design**] The `UTI`-then-`mimeType` chain returns early as soon as `UTType(uti)` resolves, even when the resolved type yields no `preferredFilenameExtension`. Abstract UTIs resolve fine but have no canonical extension, so a usable `mimeType` sitting right there is never consulted.

Verified with a standalone Swift harness against the real UniformTypeIdentifiers framework:

```
UTI=public.image  mime=image/jpeg     -> resolved=public.image   ext=nil  => no rename
UTI=public.data   mime=image/jpeg     -> resolved=public.data    ext=nil  => no rename
UTI=public.content mime=application/pdf -> resolved=public.content ext=nil  => no rename
```

`UTI: 'public.image'` + `mimeType: 'image/jpeg'` is a perfectly reasonable call, and it silently does nothing. This also contradicts the new TSDoc on `UTI`, which promises "iOS can only resolve types registered by the system or an installed app. Otherwise, `UTI` is ignored and `mimeType` is used instead" — here the `UTI` *is* resolvable, just not concrete, and the fallback never runs.

Suggest selecting on usability rather than mere resolvability, so each candidate is only accepted if it can actually name a file:

```swift
private func declaredContentType(_ options: SharingOptions) -> UTType? {
  let candidates = [options.UTI.flatMap(UTType.init), options.mimeType.flatMap { UTType(mimeType: $0) }]
  return candidates.compactMap { $0 }.first { $0.preferredFilenameExtension != nil }
}
```

That keeps `UTI` precedence while letting `mimeType` actually serve as the documented fallback.

packages/expo-sharing/ios/SharingModule.swift
```diff
@@ -103,3 +162,49 @@ public final class SharingModule: Module {
...
+      promise.reject(exception)
+    }
+  }
+
+  deinit {
+    guard !isSettled else {
+      return
+    }
+    cleanup()
+    promise.reject(FailedToPresentShareSheetException())
```
tsapeta, line 193: [🏗️ **design**] Using `deinit` to reject the promise makes correctness depend on the exact retain graph of a `UIActivityViewController` you don't own, which is fragile in both directions:

- **Leak if UIKit over-retains.** If the controller (or its completion handler) is retained past the sheet's life, or if `completionWithItemsHandler` is never invoked *and* the block outlives the module, the session never deinits and the promise hangs forever. The whole point of the handler comment is that this API's invocation guarantees are quirky.
- **Misleading error if it deinits for an unrelated reason.** Any path that drops the last reference without settling reports `FailedToPresentShareSheetException` ("The share sheet could not be presented") even when the sheet presented fine.

`isSettled` is also read/written without synchronization: `settle` runs on whichever queue invoked it, while `deinit` runs wherever the last release happens. That's a formal data race on a non-atomic `Bool`. The consequence is mild rather than fatal, since `JavaScriptPromise` settles via take-once refs (`resolveFunction.take()`), so a double-settle is a silent no-op rather than a crash. But the same torn read can make `cleanup()` run twice, and more importantly the race is on the flag that decides whether to reject at all.

Since presentation failure is really only knowable at the call site, I'd suggest making it explicit rather than inferred: reject directly on the paths you know failed (missing view controller, and optionally a `presentedViewController` check after `present`), and keep the session purely as a settle-once + cleanup helper. If you keep the `deinit` guard as a backstop, it needs a lock or an atomic around `isSettled`.

packages/expo-sharing/ios/SharingModule.swift
```diff
@@ -91,6 +101,55 @@ public final class SharingModule: Module {
...
+    guard let type = declaredContentType(options), let ext = type.preferredFilenameExtension else {
+      return url
+    }
+
+    if let currentType = UTType(filenameExtension: url.pathExtension), currentType.conforms(to: type) {
+      return url
+    }
+
+    let baseName = url.deletingPathExtension().lastPathComponent
+    let linkDirectory = FileManager.default.temporaryDirectory
```
tsapeta, line 129: [🏗️ **design**] Staging into the global `FileManager.default.temporaryDirectory` bypasses the appContext-scoped cache directory the rest of the codebase uses. `expo-modules-core` exposes `FileSystemUtilities.generatePathInCache(appContext, in:extension:)` (backed by `appContext.config.cacheDirectory`), and ~10 packages follow the "per-module subdir of the appContext cache dir" convention (expo-camera `"Camera"`, expo-contacts `"Contacts"`, expo-image-manipulator `"ImageManipulator"`, etc.).

This matters beyond consistency: `AppContextConfig.cacheDirectory` is overridable, and that override is how Expo Go scopes storage per experience. `FileSystemUtilities.getInternalPathPermissions` grants read+write inside the scoped dirs, while paths outside fall through to `getExternalPathPermissions`, which explicitly denies scoped paths containing `ExponentExperienceData`. Hard-linking a scoped file into the shared global temp dir places readable content outside the experience sandbox.

Using `appContext?.config.cacheDirectory` with an `"expo-sharing"` subdirectory (falling back to `temporaryDirectory` only if unavailable) would both match convention and keep the link inside the scope the source file came from.

packages/expo-sharing/ios/SharingModule.swift
```diff
@@ -91,6 +101,55 @@ public final class SharingModule: Module {
...
+
+    guard let type = declaredContentType(options), let ext = type.preferredFilenameExtension else {
+      return url
+    }
+
+    if let currentType = UTType(filenameExtension: url.pathExtension), currentType.conforms(to: type) {
+      return url
+    }
+
+    let baseName = url.deletingPathExtension().lastPathComponent
```
tsapeta, line 128: [💡 suggestion] `deletingPathExtension()` strips whatever trails the last dot, even when it isn't a real extension, so the user-visible filename in the share sheet loses part of its name. Verified:

```
"Q3 2026 Report.v2" + public.jpeg -> "Q3 2026 Report.jpeg"   (".v2" lost)
"archive.tar.gz"    + gzip        -> "archive.tar.gz"        (fine, conforms)
"report.final"      + public.jpeg -> "report.jpeg"           (".final" lost)
```

Since the branch is only reached when the current extension does *not* conform to the declared type, appending rather than replacing is arguably more faithful to the caller's filename: `url.lastPathComponent` + `.appendingPathExtension(ext)` yields `Q3 2026 Report.v2.jpeg`, keeping the name intact while still giving the sheet the extension it needs. Minor, but it's the name the user actually sees.

packages/expo-sharing/ios/SharingModule.swift
```diff
@@ -91,6 +101,55 @@ public final class SharingModule: Module {
...
+      .appendingPathComponent(UUID().uuidString, isDirectory: true)
+    let linkURL = linkDirectory
+      .appendingPathComponent(baseName.isEmpty ? "expo-sharing-item" : baseName)
+      .appendingPathExtension(ext)
+
+    do {
+      try FileManager.default.createDirectory(at: linkDirectory, withIntermediateDirectories: true)
+      do {
+        try FileManager.default.linkItem(at: url, to: linkURL)
+      } catch {
```
tsapeta, line 140: [💡 suggestion] Removing `.runOnQueue(.main)` correctly moves this I/O off the main thread, but the default `AsyncFunction` queue is a single **serial** queue shared process-wide by every async function in every Expo module (`AsyncFunctionDefinition.swift`: `DispatchQueue(label: "expo.modules.AsyncFunctionQueue", qos: .userInitiated)`).

For the hard-link path that's irrelevant. For the `EXDEV` `copyItem` fallback on a large file (a video from a file provider is the realistic case) this head-of-line-blocks unrelated async calls across the whole app for the duration of a full file copy.

An explicit `.runOnQueue` with a private queue for this function would keep the copy from contending with everything else.

packages/expo-sharing/src/Sharing.types.ts
```diff
@@ -1,13 +1,23 @@
...
```
tsapeta, line 4: [💡 suggestion] Two accuracy points on the new wording, both flowing from the implementation:

1. The `application/octet-stream` caveat is narrower than the actual limitation. Any MIME type that resolves only to an abstract type (`public.data`) or *dynamically* (`application/vnd.*`, or anything with parameters like `text/plain; charset=utf-8`, which resolves to a `dyn.*` type with `preferredFilenameExtension == nil`) is equally ignored. Suggest generalizing: the type is only honored when iOS can map it to a canonical filename extension, with `application/octet-stream` as the example rather than the rule.

2. "When iOS can resolve the identifier, it takes precedence over `mimeType`" on the `UTI` field is currently stronger than the code guarantees: a resolvable-but-abstract `UTI` takes precedence *and* suppresses the `mimeType` fallback entirely, so neither is applied. If you adopt the `declaredContentType` fix, this sentence becomes accurate as written; if not, it needs a caveat.

packages/expo-sharing/ios/SharingModule.swift
```diff
@@ -91,6 +101,55 @@ public final class SharingModule: Module {
...
+    var isDirectory: ObjCBool = false
+    guard FileManager.default.fileExists(atPath: url.path, isDirectory: &isDirectory), !isDirectory.boolValue else {
+      return url
+    }
+
+    guard let type = declaredContentType(options), let ext = type.preferredFilenameExtension else {
+      return url
+    }
+
+    if let currentType = UTType(filenameExtension: url.pathExtension), currentType.conforms(to: type) {
```
tsapeta, line 124: [💡 suggestion] Worth confirming this is the intended precedence: when the declared type and the actual extension disagree concretely, the declaration wins and the file is presented under a mismatched name. Verified `UTI: 'public.png'` on `photo.jpg` stages `photo.png` — a real JPEG advertised as a PNG, which some receiving apps will fail to decode.

Respecting the caller's declaration is a defensible reading of "Respect passed UTI/mimeType," and second-guessing it would require sniffing content. But since the existing extension already resolved to a concrete, *non-conforming* type, that's decent evidence the declaration is the wrong one. At minimum this precedence deserves a brief comment so the behavior is intentional rather than incidental.

packages/expo-sharing/ios/SharingExceptions.swift
```diff
@@ -12,6 +12,12 @@ internal final class MissingCurrentViewControllerException: Exception, @unchecke
   }
 }
 
+internal final class FailedToPresentShareSheetException: Exception, @unchecked Sendable {
+  override var reason: String {
+    "The share sheet could not be presented"
```
tsapeta, line 17: [🐧 nit] Per the repo's error-message guidance (what / why / how), this one states only the what. Because it's currently raised from `deinit` it's also the message users hit for causes unrelated to presentation, which makes the bare wording actively misleading.

Something like "The share sheet could not be presented. This usually means the app had no view controller to present from, for example while it was backgrounded. Try sharing again once the app is in the foreground." gives a cause and a next step.

packages/expo-sharing/ios/SharingModule.swift
behenate: This behavior is correct. If someone explicitly provides a `.png` extension for a file it would be weird not to respect that. I'll add a comment.

packages/expo-sharing/ios/SharingModule.swift
```diff
@@ -17,43 +23,76 @@ public final class SharingModule: Module {
...
+          )
+          activityController.popoverPresentationController?.sourceView = currentViewController.view
+          activityController.modalPresentationStyle = .pageSheet
+        }
+
+        currentViewController.present(activityController, animated: true)
+      }
     }
-    .runOnQueue(.main)
+    .runOnQueue(sharingQueue)
```
tsapeta: It's not clear to me, why do we need to use a custom queue?

packages/expo-sharing/ios/SharingModule.swift
```diff
@@ -91,6 +130,95 @@ public final class SharingModule: Module {
...
+  private func shareableURL(for url: URL, options: SharingOptions) -> URL {
+    var isDirectory: ObjCBool = false
+    guard FileManager.default.fileExists(atPath: url.path, isDirectory: &isDirectory), !isDirectory.boolValue else {
+      return url
+    }
+
+    let stagingDirectory = (appContext?.config.cacheDirectory ?? FileManager.default.temporaryDirectory)
+      .appendingPathComponent("expo-sharing", isDirectory: true)
+    cleanupStaleShareDirectories(in: stagingDirectory)
+
```
tsapeta: [design] Suggest dropping this sweeper entirely and staging into `FileManager.default.temporaryDirectory` instead of the cache directory. The system purges `tmp/` on its own, which removes the need for `cleanupStaleShareDirectories`, `stagedShareLifetime`, and the UUID-validation/resource-key enumeration inside it.

That's also the established pattern in this repo for the same problem: `expo-media-library/ios/MediaLibraryModule.swift:402` writes a full video into `temporaryDirectory` under a UUID with an extension and never sweeps, and `expo-audio/ios/AudioUtils.swift:146` does the same.

As written this is the only thing reclaiming staged files (the success path doesn't delete), and it only runs when someone calls `shareAsync` again — so an app that shares once never reclaims. Since a hard link costs no data blocks, leaving it to the OS purge is both simpler and no worse.

Separately, the call sits before the `declaredContentType` / `conforms(to:)` early returns, so today every `shareAsync` pays a `contentsOfDirectory` plus a `resourceValues` per entry even when nothing gets staged.

packages/expo-sharing/ios/SharingModule.swift
```diff
@@ -91,6 +130,95 @@ public final class SharingModule: Module {
...
+    }
+
+    // An explicitly declared type intentionally takes precedence over a conflicting filename extension.
+    let baseName = url.lastPathComponent
+    let linkDirectory = stagingDirectory
+      .appendingPathComponent(UUID().uuidString, isDirectory: true)
+    let linkURL = linkDirectory
+      .appendingPathComponent(baseName.isEmpty ? "expo-sharing-item" : baseName)
+      .appendingPathExtension(ext)
+
```
tsapeta: [design] `appendingPathExtension` appends to the whole `lastPathComponent`, so the existing extension is kept rather than replaced. The PR's own demo case (`photo.pdf` + `mimeType: 'image/jpeg'`) produces `photo.pdf.jpeg`.

iOS resolves the type from the trailing extension so the fix still works, but that filename is what AirDrop and Save-to-Files write on the receiving end, so the user sees `photo.pdf.jpeg`. Since the comment just above says the declared type intentionally wins over a conflicting extension, the old extension is noise:

```swift
let baseName = url.deletingPathExtension().lastPathComponent
```

keeping the existing empty-name fallback.

packages/expo-sharing/ios/SharingModule.swift
```diff
@@ -17,43 +23,76 @@ public final class SharingModule: Module {
...
+          )
+          activityController.popoverPresentationController?.sourceView = currentViewController.view
+          activityController.modalPresentationStyle = .pageSheet
+        }
+
+        currentViewController.present(activityController, animated: true)
+      }
     }
-    .runOnQueue(.main)
+    .runOnQueue(sharingQueue)
```
tsapeta: [design] Adding to the existing question about the custom queue: as far as I can tell nothing here needs it. `linkItem` is a metadata-only operation, and the one genuinely slow path is the rare `EXDEV` full copy — which would argue for moving just that off the main thread, not the whole function.

The queue also isn't free. It's what forces `isShareSheetActive()` to do a `DispatchQueue.main.sync` hop to read state that lives on main, and it splits the body across two queues so the in-progress check has to happen twice (see the separate note on the first guard, which can't actually reject anything).

More broadly, this plus `activeShareSheet`, the readiness guard, `ShareSheetSession`, and the two new exceptions are a separate concern from respecting `UTI`/`mimeType`, and they change `shareAsync`'s observable contract — the readiness guard is strict enough to reject legitimate cases (`presentedViewController != nil` covers sharing from above an app's own modal; `transitionCoordinator` can be non-nil during unrelated animations). Splitting them into their own PR would let that behavior change be judged on its own, and would make this one revertible without giving up the type fix.

packages/expo-sharing/ios/SharingExceptions.swift
```diff
@@ -12,6 +12,18 @@ internal final class MissingCurrentViewControllerException: Exception, @unchecke
   }
 }
 
+internal final class SharingInProgressException: Exception, @unchecked Sendable {
+  override var reason: String {
```
tsapeta: [critical] If the concurrency work does stay in this PR, both of these need documenting: a `### 🛠 Breaking changes` CHANGELOG entry (the current entry only mentions the UTI/mimeType fix) and a note in the `shareAsync` TSDoc that it can reject while another share is in flight, so callers know to serialize.

Right now a JS caller has no way to learn these rejection paths exist short of hitting them in production.

packages/expo-sharing/ios/SharingModule.swift
```diff
@@ -17,43 +23,76 @@ public final class SharingModule: Module {
...
-      activityController.completionWithItemsHandler = { _, _, _, _ in
-        // Resolve unconditionally. UIActivityViewController invokes this once
-        // on dismissal for every (activityType, completed) permutation. The
-        // previous implementation only resolved two of four cases, leaking
-        // the promise when the user picked an activity and then cancelled
-        // its follow-up dialog (e.g. tapped Print, then cancelled the print
-        // dialog: activityType != nil, completed == false).
-        promise.resolve(nil)
+      guard !isShareSheetActive() else {
+        throw SharingInProgressException()
```
tsapeta: [suggestion] This guard can't reject a concurrent call: `activeShareSheet` is only assigned inside the `DispatchQueue.main.async` block below, so two calls arriving back-to-back both observe `nil` here and both enqueue a main-queue block. The re-check at the top of that block is what actually enforces exclusivity.

So this adds a `DispatchQueue.main.sync` hop (via `isShareSheetActive()`) on every call without providing a guarantee the inner check doesn't already give. If the guards survive the split, one authoritative check on the queue that owns the state is enough.

packages/expo-sharing/ios/SharingModule.swift
```diff
@@ -91,6 +130,95 @@ public final class SharingModule: Module {
...
+    let stagingDirectory = (appContext?.config.cacheDirectory ?? FileManager.default.temporaryDirectory)
+      .appendingPathComponent("expo-sharing", isDirectory: true)
+    cleanupStaleShareDirectories(in: stagingDirectory)
+
+    guard let type = declaredContentType(options), let ext = type.preferredFilenameExtension else {
+      return url
+    }
+
+    if let currentType = UTType(filenameExtension: url.pathExtension), currentType.conforms(to: type) {
+      return url
```
tsapeta: [suggestion] The `conforms(to:)` direction is right — it asks whether the current extension already communicates at least the declared type, so `photo.jpg` + `public.image` correctly skips staging.

One consequence worth a docs line rather than code: when the caller's declaration is wrong, this now propagates the mistake. `photo.heic` + `mimeType: 'image/jpeg'` stages as `.jpeg`, asserting JPEG for HEIC bytes, where before the correct `.heic` extension was used. That's consistent with Android (where `mimeType` goes into the Intent verbatim) and I wouldn't add content sniffing, but the TSDoc could say the declared type is trusted and not validated against file contents.

apps/native-component-list/src/screens/SharingScreen.tsx
```diff
@@ -1,74 +1,415 @@
...
+const REACT_NATIVE_RAW = 'https://raw.githubusercontent.com/facebook/react-native/main';
 
-  state = {
-    loading: true,
-    isAvailable: false,
-  };
+const sources = {
+  extensionlessJpeg:
+    'https://useful-cow-483.convex.cloud/api/storage/042a5375-85f6-4351-b950-bb3b65b7d6a3',
+  jpeg: `${EXPO_RAW}/apps/test-suite/assets/example_image_1.jpg`,
```
tsapeta: [design] I recognize this as the reporter's repro image from #44887, but it points at a Convex bucket nobody on the team controls — the only non-`github.com` asset host in the repo. When it expires the demo fails with an HTTP error and the cause won't be obvious.

Since the point is "URL with no extension," committing a small extensionless fixture under `apps/native-component-list/assets/` would be stable and would also let the screen work offline.

apps/native-component-list/src/screens/SharingScreen.tsx
```diff
@@ -1,74 +1,415 @@
...
 import * as FileSystem from 'expo-file-system/legacy';
 import * as Sharing from 'expo-sharing';
-import React from 'react';
-import { Image, StyleSheet, Text, View } from 'react-native';
+import type { SharingOptions } from 'expo-sharing';
+import React, { useState } from 'react';
+import {
+  ActivityIndicator,
+  Alert,
+  Pressable,
```
tsapeta: [nit] `Alert` is imported but never used (`@typescript-eslint/no-unused-vars` will warn).

apps/native-component-list/src/screens/SharingScreen.tsx
```diff
@@ -1,74 +1,415 @@
...
 
-  state = {
-    loading: true,
-    isAvailable: false,
-  };
+const sources = {
+  extensionlessJpeg:
+    'https://useful-cow-483.convex.cloud/api/storage/042a5375-85f6-4351-b950-bb3b65b7d6a3',
+  jpeg: `${EXPO_RAW}/apps/test-suite/assets/example_image_1.jpg`,
+  png: `${EXPO_RAW}/apps/bare-expo/assets/icon.png`,
```
tsapeta: [nit] These fixtures track whatever is on `main` right now, so an upstream rename or deletion breaks the screen with no signal, and a couple are large for a demo download (the 1MB mp4, the mov). Pinning to a commit SHA instead of `main` would at least make the URLs stable.

packages/expo-sharing/ios/SharingModule.swift
tsapeta: [🏗️ **design**] Suggest dropping this sweeper entirely and staging into `FileManager.default.temporaryDirectory` instead of the cache directory. The system purges `tmp/` on its own, which removes the need for `cleanupStaleShareDirectories`, `stagedShareLifetime`, and the UUID-validation/resource-key enumeration inside it.

That's also the established pattern in this repo for the same problem: `expo-media-library/ios/MediaLibraryModule.swift:402` writes a full video into `temporaryDirectory` under a UUID with an extension and never sweeps, and `expo-audio/ios/AudioUtils.swift:146` does the same.

As written this is the only thing reclaiming staged files (the success path doesn't delete), and it only runs when someone calls `shareAsync` again — so an app that shares once never reclaims. Since a hard link costs no data blocks, leaving it to the OS purge is both simpler and no worse.

Separately, the call sits before the `declaredContentType` / `conforms(to:)` early returns, so today every `shareAsync` pays a `contentsOfDirectory` plus a `resourceValues` per entry even when nothing gets staged.

packages/expo-sharing/ios/SharingModule.swift
tsapeta: [🏗️ **design**] `appendingPathExtension` appends to the whole `lastPathComponent`, so the existing extension is kept rather than replaced. The PR's own demo case (`photo.pdf` + `mimeType: 'image/jpeg'`) produces `photo.pdf.jpeg`.

iOS resolves the type from the trailing extension so the fix still works, but that filename is what AirDrop and Save-to-Files write on the receiving end, so the user sees `photo.pdf.jpeg`. Since the comment just above says the declared type intentionally wins over a conflicting extension, the old extension is noise:

```swift
let baseName = url.deletingPathExtension().lastPathComponent
```

keeping the existing empty-name fallback.

packages/expo-sharing/ios/SharingModule.swift
tsapeta: [🏗️ **design**] Adding to the existing question about the custom queue: as far as I can tell nothing here needs it. `linkItem` is a metadata-only operation, and the one genuinely slow path is the rare `EXDEV` full copy — which would argue for moving just that off the main thread, not the whole function.

The queue also isn't free. It's what forces `isShareSheetActive()` to do a `DispatchQueue.main.sync` hop to read state that lives on main, and it splits the body across two queues so the in-progress check has to happen twice (see the separate note on the first guard, which can't actually reject anything).

More broadly, this plus `activeShareSheet`, the readiness guard, `ShareSheetSession`, and the two new exceptions are a separate concern from respecting `UTI`/`mimeType`, and they change `shareAsync`'s observable contract — the readiness guard is strict enough to reject legitimate cases (`presentedViewController != nil` covers sharing from above an app's own modal; `transitionCoordinator` can be non-nil during unrelated animations). Splitting them into their own PR would let that behavior change be judged on its own, and would make this one revertible without giving up the type fix.

packages/expo-sharing/ios/SharingExceptions.swift
tsapeta: [🚨 **critical**] If the concurrency work does stay in this PR, both of these need documenting: a `### 🛠 Breaking changes` CHANGELOG entry (the current entry only mentions the UTI/mimeType fix) and a note in the `shareAsync` TSDoc that it can reject while another share is in flight, so callers know to serialize.

Right now a JS caller has no way to learn these rejection paths exist short of hitting them in production.

packages/expo-sharing/ios/SharingModule.swift
tsapeta: [💡 suggestion] This guard can't reject a concurrent call: `activeShareSheet` is only assigned inside the `DispatchQueue.main.async` block below, so two calls arriving back-to-back both observe `nil` here and both enqueue a main-queue block. The re-check at the top of that block is what actually enforces exclusivity.

So this adds a `DispatchQueue.main.sync` hop (via `isShareSheetActive()`) on every call without providing a guarantee the inner check doesn't already give. If the guards survive the split, one authoritative check on the queue that owns the state is enough.

packages/expo-sharing/ios/SharingModule.swift
tsapeta: [💡 suggestion] The `conforms(to:)` direction is right — it asks whether the current extension already communicates at least the declared type, so `photo.jpg` + `public.image` correctly skips staging.

One consequence worth a docs line rather than code: when the caller's declaration is wrong, this now propagates the mistake. `photo.heic` + `mimeType: 'image/jpeg'` stages as `.jpeg`, asserting JPEG for HEIC bytes, where before the correct `.heic` extension was used. That's consistent with Android (where `mimeType` goes into the Intent verbatim) and I wouldn't add content sniffing, but the TSDoc could say the declared type is trusted and not validated against file contents.

apps/native-component-list/src/screens/SharingScreen.tsx
tsapeta: [🏗️ **design**] I recognize this as the reporter's repro image from #44887, but it points at a Convex bucket nobody on the team controls — the only non-`github.com` asset host in the repo. When it expires the demo fails with an HTTP error and the cause won't be obvious.

Since the point is "URL with no extension," committing a small extensionless fixture under `apps/native-component-list/assets/` would be stable and would also let the screen work offline.

apps/native-component-list/src/screens/SharingScreen.tsx
tsapeta: [🐧 nit] `Alert` is imported but never used (`@typescript-eslint/no-unused-vars` will warn).

apps/native-component-list/src/screens/SharingScreen.tsx
tsapeta: [🐧 nit] These fixtures track whatever is on `main` right now, so an upstream rename or deletion breaks the screen with no signal, and a couple are large for a demo download (the 1MB mp4, the mov). Pinning to a commit SHA instead of `main` would at least make the URLs stable.

packages/expo-sharing/ios/SharingModule.swift
```diff
@@ -91,6 +130,95 @@ public final class SharingModule: Module {
...
+    return DispatchQueue.main.sync { activeShareSheet != nil }
+  }
+
+  private func shareableURL(for url: URL, options: SharingOptions) -> URL {
+    var isDirectory: ObjCBool = false
+    guard FileManager.default.fileExists(atPath: url.path, isDirectory: &isDirectory), !isDirectory.boolValue else {
+      return url
+    }
+
+    let stagingDirectory = (appContext?.config.cacheDirectory ?? FileManager.default.temporaryDirectory)
```
tsapeta: [🏗️ **design**] Suggest dropping this sweeper entirely and staging into `FileManager.default.temporaryDirectory` instead of the cache directory. The system purges `tmp/` on its own, which removes the need for `cleanupStaleShareDirectories`, `stagedShareLifetime`, and the UUID-validation/resource-key enumeration inside it.

That's also the established pattern in this repo for the same problem: `expo-media-library/ios/MediaLibraryModule.swift:402` writes a full video into `temporaryDirectory` under a UUID with an extension and never sweeps, and `expo-audio/ios/AudioUtils.swift:146` does the same.

As written this is the only thing reclaiming staged files (the success path doesn't delete), and it only runs when someone calls `shareAsync` again — so an app that shares once never reclaims. Since a hard link costs no data blocks, leaving it to the OS purge is both simpler and no worse.

Separately, the call sits before the `declaredContentType` / `conforms(to:)` early returns, so today every `shareAsync` pays a `contentsOfDirectory` plus a `resourceValues` per entry even when nothing gets staged.

packages/expo-sharing/ios/SharingModule.swift
```diff
@@ -91,6 +130,95 @@ public final class SharingModule: Module {
...
+
+    if let currentType = UTType(filenameExtension: url.pathExtension), currentType.conforms(to: type) {
+      return url
+    }
+
+    // An explicitly declared type intentionally takes precedence over a conflicting filename extension.
+    let baseName = url.lastPathComponent
+    let linkDirectory = stagingDirectory
+      .appendingPathComponent(UUID().uuidString, isDirectory: true)
+    let linkURL = linkDirectory
```
tsapeta, line 152: [🏗️ **design**] `appendingPathExtension` appends to the whole `lastPathComponent`, so the existing extension is kept rather than replaced. The PR's own demo case (`photo.pdf` + `mimeType: 'image/jpeg'`) produces `photo.pdf.jpeg`.

iOS resolves the type from the trailing extension so the fix still works, but that filename is what AirDrop and Save-to-Files write on the receiving end, so the user sees `photo.pdf.jpeg`. Since the comment just above says the declared type intentionally wins over a conflicting extension, the old extension is noise:

```swift
let baseName = url.deletingPathExtension().lastPathComponent
```

keeping the existing empty-name fallback.

packages/expo-sharing/ios/SharingModule.swift
```diff
@@ -17,43 +23,76 @@ public final class SharingModule: Module {
...
+            height: rect?.height ?? 0
+          )
+          activityController.popoverPresentationController?.sourceView = currentViewController.view
+          activityController.modalPresentationStyle = .pageSheet
+        }
+
+        currentViewController.present(activityController, animated: true)
+      }
     }
-    .runOnQueue(.main)
```
tsapeta, line 56: [🏗️ **design**] Adding to the existing question about the custom queue: as far as I can tell nothing here needs it. `linkItem` is a metadata-only operation, and the one genuinely slow path is the rare `EXDEV` full copy — which would argue for moving just that off the main thread, not the whole function.

The queue also isn't free. It's what forces `isShareSheetActive()` to do a `DispatchQueue.main.sync` hop to read state that lives on main, and it splits the body across two queues so the in-progress check has to happen twice (see the separate note on the first guard, which can't actually reject anything).

More broadly, this plus `activeShareSheet`, the readiness guard, `ShareSheetSession`, and the two new exceptions are a separate concern from respecting `UTI`/`mimeType`, and they change `shareAsync`'s observable contract — the readiness guard is strict enough to reject legitimate cases (`presentedViewController != nil` covers sharing from above an app's own modal; `transitionCoordinator` can be non-nil during unrelated animations). Splitting them into their own PR would let that behavior change be judged on its own, and would make this one revertible without giving up the type fix.

packages/expo-sharing/ios/SharingExceptions.swift
```diff
@@ -12,6 +12,18 @@ internal final class MissingCurrentViewControllerException: Exception, @unchecke
   }
 }
 
+internal final class SharingInProgressException: Exception, @unchecked Sendable {
```
tsapeta: [🚨 **critical**] If the concurrency work does stay in this PR, both of these need documenting: a `### 🛠 Breaking changes` CHANGELOG entry (the current entry only mentions the UTI/mimeType fix) and a note in the `shareAsync` TSDoc that it can reject while another share is in flight, so callers know to serialize.

Right now a JS caller has no way to learn these rejection paths exist short of hitting them in production.

packages/expo-sharing/ios/SharingModule.swift
```diff
@@ -17,43 +23,76 @@ public final class SharingModule: Module {
...
-      activityController.title = options.dialogTitle
-
-      activityController.completionWithItemsHandler = { _, _, _, _ in
-        // Resolve unconditionally. UIActivityViewController invokes this once
-        // on dismissal for every (activityType, completed) permutation. The
-        // previous implementation only resolved two of four cases, leaking
-        // the promise when the user picked an activity and then cancelled
-        // its follow-up dialog (e.g. tapped Print, then cancelled the print
-        // dialog: activityType != nil, completed == false).
-        promise.resolve(nil)
```
tsapeta, line 31: [💡 suggestion] This guard can't reject a concurrent call: `activeShareSheet` is only assigned inside the `DispatchQueue.main.async` block below, so two calls arriving back-to-back both observe `nil` here and both enqueue a main-queue block. The re-check at the top of that block is what actually enforces exclusivity.

So this adds a `DispatchQueue.main.sync` hop (via `isShareSheetActive()`) on every call without providing a guarantee the inner check doesn't already give. If the guards survive the split, one authoritative check on the queue that owns the state is enough.

packages/expo-sharing/ios/SharingModule.swift
```diff
@@ -91,6 +130,95 @@ public final class SharingModule: Module {
...
+      return url
+    }
+
+    let stagingDirectory = (appContext?.config.cacheDirectory ?? FileManager.default.temporaryDirectory)
+      .appendingPathComponent("expo-sharing", isDirectory: true)
+    cleanupStaleShareDirectories(in: stagingDirectory)
+
+    guard let type = declaredContentType(options), let ext = type.preferredFilenameExtension else {
+      return url
+    }
```
tsapeta, line 136: [💡 suggestion] The `conforms(to:)` direction is right — it asks whether the current extension already communicates at least the declared type, so `photo.jpg` + `public.image` correctly skips staging.

One consequence worth a docs line rather than code: when the caller's declaration is wrong, this now propagates the mistake. `photo.heic` + `mimeType: 'image/jpeg'` stages as `.jpeg`, asserting JPEG for HEIC bytes, where before the correct `.heic` extension was used. That's consistent with Android (where `mimeType` goes into the Intent verbatim) and I wouldn't add content sniffing, but the TSDoc could say the declared type is trusted and not validated against file contents.

apps/native-component-list/src/screens/SharingScreen.tsx
```diff
@@ -1,74 +1,415 @@
...
+const EXPO_RAW = 'https://raw.githubusercontent.com/expo/expo/main';
+const REACT_NATIVE_RAW = 'https://raw.githubusercontent.com/facebook/react-native/main';
 
-  state = {
-    loading: true,
-    isAvailable: false,
-  };
+const sources = {
+  extensionlessJpeg:
+    'https://useful-cow-483.convex.cloud/api/storage/042a5375-85f6-4351-b950-bb3b65b7d6a3',
```
tsapeta: [🏗️ **design**] I recognize this as the reporter's repro image from #44887, but it points at a Convex bucket nobody on the team controls — the only non-`github.com` asset host in the repo. When it expires the demo fails with an HTTP error and the cause won't be obvious.

Since the point is "URL with no extension," committing a small extensionless fixture under `apps/native-component-list/assets/` would be stable and would also let the screen work offline.

apps/native-component-list/src/screens/SharingScreen.tsx
```diff
@@ -1,74 +1,415 @@
-import { Asset } from 'expo-asset';
 import * as FileSystem from 'expo-file-system/legacy';
 import * as Sharing from 'expo-sharing';
-import React from 'react';
-import { Image, StyleSheet, Text, View } from 'react-native';
+import type { SharingOptions } from 'expo-sharing';
+import React, { useState } from 'react';
+import {
+  ActivityIndicator,
+  Alert,
```
tsapeta: [🐧 nit] `Alert` is imported but never used (`@typescript-eslint/no-unused-vars` will warn).

apps/native-component-list/src/screens/SharingScreen.tsx
tsapeta: [🐧 nit] These fixtures track whatever is on `main` right now, so an upstream rename or deletion breaks the screen with no signal, and a couple are large for a demo download (the 1MB mp4, the mov). Pinning to a commit SHA instead of `main` would at least make the URLs stable.

tsapeta (changes requested):

packages/expo-sharing/ios/SharingModule.swift
behenate: Its not really noise. We'd have to also take into accounts cases like image called "edit.final" in that case we'd rename the image to "edit.jpg" instead of "edit.final.jpg". Also I believe that keeping the original extension is more robust in general.

Developers can rename the files using `expo-file-system` if they don't like our behaviour.

packages/expo-sharing/ios/SharingModule.swift
behenate: We need it, because in case we need to override the file type, we need to do a bunch of filesystem operations. We don't want to be doing that on the main thread. 

We could do it on the async modules queue, but adding the queue was suggested in this comment in the last rewiew round https://github.com/expo/expo/pull/48052#discussion_r3660369465

packages/expo-sharing/ios/SharingModule.swift
behenate: https://github.com/expo/expo/pull/48052#discussion_r3660369459

brentvatne:
@expo-bot review

tsapeta (approved):

PR #48053, opened 2026-07-22
test(cli): Drop the router-e2e `customizeFrame` override

ramonclaudio:
# Why

`apps/router-e2e/metro.config.js` adds a second collapse rule on top of `customizeFrame` that matches `expo/packages/.+`. It matches on the path, so it only works when the repo folder is named `expo`. In a fork, in a worktree, or on Windows it doesn't match and the internal frames show up.

So the test was only passing because of the override.

# How

Removed the override and updated the test to match what actually prints.

`expo` and `@expo/log-box` live in `packages/` here, not `node_modules/`, so the collapse patterns miss them and those frames show up. The code frame ends up on `setupHMR.js`. In a real app both are under `node_modules` and get collapsed. Nothing in `@expo/metro-config` changed.

One thing worth flagging: the test now names frames from `packages/expo` and `@expo/log-box`, so if someone renames `captureCurrentStack` or moves it, this test breaks. I left the line numbers out so it only matches the file and function names.

The test had a separate Windows branch because Windows never matched the override anyway. Both platforms print the same thing now, so that branch is gone.

I also added a test for the two `ERROR_IN_ROUTER_GLOBAL_SCOPE` fixtures in `serverLogLikeMetro.test.ts`. They're the same stack, one with `packages/expo-router` and one with `node_modules/expo-router`, and the snapshots keep 8 frames for the first and 1 for the second. Right now you'd only catch that by comparing the two snapshot files. Happy to drop it if it's noise.

# Test Plan

- `et check-packages @expo/cli`
- `pnpm test:playwright dev-console-errors`, 7 passed

# Checklist

### DIFF
--- a/apps/router-e2e/metro.config.js
+++ b/apps/router-e2e/metro.config.js
@@ -42,13 +42,4 @@ const config = getDefaultConfig(
 // Disable Babel's RC lookup, reducing the config loading in Babel - resulting in faster bootup for transformations
 config.transformer.enableBabelRCLookup = false;
 
-const originalCustomizeFrame = config.symbolicator.customizeFrame;
-config.symbolicator.customizeFrame = (frame) => {
-  const newFrame = originalCustomizeFrame(frame);
-  // This will make frames collapsed the same as in use apps, where expo/packages are in node_modules
-  // and thus collapsed by default.
-  newFrame.collapse ||= new RegExp('expo/packages/.+').test(frame.file);
-  return newFrame;
-};
-
 module.exports = config;
--- a/packages/@expo/cli/CHANGELOG.md
+++ b/packages/@expo/cli/CHANGELOG.md
@@ -48,6 +48,7 @@
 - Bump to `dnssd-advertise@^1.1.6` ([#47183](https://github.com/expo/expo/pull/47183) by [@kitten](https://github.com/kitten))
 - [Internal] Migrate structured event logging to the `2g` package ([#46667](https://github.com/expo/expo/pull/46667) by [@kitten](https://github.com/kitten))
 - [Internal] Migrate an initial set of events to `2g` ([#47655](https://github.com/expo/expo/pull/47655) by [@kitten](https://github.com/kitten))
+- [Internal] Remove the `router-e2e` `customizeFrame` override that was hiding stack frames in the dev console tests ([#48053](https://github.com/expo/expo/pull/48053) by [@ramonclaudio](https://github.com/ramonclaudio))
 
 ## 56.1.12 — 2026-05-26
 
--- a/packages/@expo/cli/CHANGELOG.md
+++ b/packages/@expo/cli/CHANGELOG.md
@@ -48,7 +48,6 @@
 - Bump to `dnssd-advertise@^1.1.6` ([#47183](https://github.com/expo/expo/pull/47183) by [@kitten](https://github.com/kitten))
 - [Internal] Migrate structured event logging to the `2g` package ([#46667](https://github.com/expo/expo/pull/46667) by [@kitten](https://github.com/kitten))
 - [Internal] Migrate an initial set of events to `2g` ([#47655](https://github.com/expo/expo/pull/47655) by [@kitten](https://github.com/kitten))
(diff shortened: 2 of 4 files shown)


packages/@expo/metro-config/src/customizeFrame.ts
```diff
@@ -55,7 +58,7 @@ export const INTERNAL_CALLSITES_REGEX = new RegExp(
     // Block all whatwg polyfills
     'node_modules/whatwg-.+\\.js$',
     // Hide expo-router warnings which are often wrapping all routes and imports.
-    'node_modules/expo-router/build/',
+    '(?:node_modules|packages)/expo-router/build/',
```
kitten: ah, for context, we mainly did this to prevent us "leaking" our own concerns into public code, since we're typically not differentiating, so we consciously took the hit here on this not quite matching prod (although there's a few cases like these in a few places, e.g. with how some Babel plugins apply)

@krystofwoldrich is currently the owner of this code (but OOO) but there's an additional ticket that probably correlates well here (overeager collapsing of stack traces)

apps/router-e2e/metro.config.js
```diff
@@ -42,13 +42,4 @@ const config = getDefaultConfig(
 // Disable Babel's RC lookup, reducing the config loading in Babel - resulting in faster bootup for transformations
 config.transformer.enableBabelRCLookup = false;
 
-const originalCustomizeFrame = config.symbolicator.customizeFrame;
```
kitten, line 45: Even if we drop the change below, deleting this is most likely worthwhile, because this makes it harder for us to reason about the rules of which stack frames are displayed manually 👍

packages/@expo/metro-config/src/customizeFrame.ts
ramonclaudio: Got it!

Without the regex the code frame lands on `setupHMR.js:77` with five internal frames above the app frame. Happy to bake that into the e2e, or hold the regex for @krystofwoldrich.

kitten: we have an internal ticket for this (https://linear.app/expo/issue/DVT-137/dont-hide-top-stack-trace-line-item), because I suspect that "five internal frames" here are the ones that'd be hidden in a user app.

What we see is basically not as important (and I think, hiding more stack frames from us to try to imitate user apps will lead to more cases here where we mess something up, and risk affecting user apps in odd ways)

Basically, I think there's two decisions here that interact poorly:
- we try to hide stack traces of `node_modules` broadly
- we assume internal errors should have "shallow stacks" only into user code

I think both doesn't hold and wasn't a great decision. At the worst case (as per the ticket) I suspect that we're accidentally emptying out many stack traces. Hiding the "top-level frame" in general is risky, since we then don't necessarily point at the origin of the error.

I think, to combine this with this error, if we have a reason to suspect that:
- user apps point at files that are internal when that's not where the error originates
- or; we empty out the stack trace in user apps because frames are "internal" (which essentially means that we assume our code/libs issues no error, which is clearly not right)

then follows, that our heuristics for editing stack traces is overall "bad" or "inaccurate" and not helpful

packages/@expo/cli/CHANGELOG.md
```diff
@@ -48,6 +48,7 @@
 - Bump to `dnssd-advertise@^1.1.6` ([#47183](https://github.com/expo/expo/pull/47183) by [@kitten](https://github.com/kitten))
 - [Internal] Migrate structured event logging to the `2g` package ([#46667](https://github.com/expo/expo/pull/46667) by [@kitten](https://github.com/kitten))
 - [Internal] Migrate an initial set of events to `2g` ([#47655](https://github.com/expo/expo/pull/47655) by [@kitten](https://github.com/kitten))
+- [Internal] Remove the `router-e2e` `customizeFrame` override that was hiding stack frames in the dev console tests ([#48053](https://github.com/expo/expo/pull/48053) by [@ramonclaudio](https://github.com/ramonclaudio))
```
kitten:
```suggestion
```

nit: when we make an internal-only modification (that's in no way observable by users) we omit the changelog entry

kitten (approved):

Issue #47625, opened 2026-07-08
[Android][expo-ui] Host with matchContents crashes ReactHost: "performMeasureAndLayout called during measure layout" (synchronous onLayoutContent dispatch during measure pass + reanimated)


ethras:
### Minimal reproducible example

Not yet extracted to a standalone repo (happy to do so if needed) — but the crash is fully explained by the synchronous event dispatch in `HostView.kt` and reproduces reliably in our production app on mid-range Android devices (Samsung A54/A35, Pixel 9a; Android 15/16). Conditions:

1. A screen containing several `@expo/ui/jetpack-compose` `Host` views with `matchContents` (in our case `TextField`s inside `Host`).
2. `react-native-reanimated` (4.5.0) installed — its `FabricEventDispatcher` hook is part of the crash chain.
3. Navigate to that screen with `router.replace(...)` (react-native-screens stack) so the screen mounts while the stack transition/reanimated queue is active.

### What happened?

On Android, mounting `Host` views (with `matchContents`) right after a stack navigation intermittently crashes the app with:

```
IllegalArgumentException: performMeasureAndLayout called during measure layout
mechanism: expoReactHost (unhandled)
```

The ReactHost is torn down and the user is left on a dead, unresponsive blank screen (it survives backgrounding/foregrounding).

**Root cause** (from the stack trace below, reading bottom-up):

1. `ExpoView.requestLayout` posts `measureAndLayout`, which measures the `HostView` → `ExpoComposeView` → `AndroidComposeView` (`measureOnly`).
2. During that measure pass, the Compose layout in `HostView.MaybeMatchContentsLayout` hits `Modifier.onSizeChanged { size -> dispatchOnLayoutContent(size, density) }` (`HostView.kt:155`), which **synchronously** emits the `onLayoutContent` event (`HostView.kt:226`) through `EventDispatcher` → `KModuleEventEmitterWrapper.emit` → `FabricEventDispatcher.dispatchEvent`.
3. react-native-reanimated listens to Fabric event dispatches (`NodesManager.onEventDispatch`) and synchronously runs `performOperations`, which calls `FabricUIManager.scheduleMountItem` and executes mount items **on the same call stack**.
4. One of those mount items (`SurfaceMountingManager.updateLayout`) re-enters `onMeasure` of a `HostView`/`ComposeView` that is still inside its own measure pass → Compose throws `performMeasureAndLayout called during measure layout`.

[...]

Issue #43537, opened 2026-02-28
[SDK 55] Expo UI menu with glass button + native view freezes app on iOS


iankberry:
### Minimal reproducible example

https://github.com/iankberry/expo-ui-menu-hostview-repro

### Steps to reproduce

1. Checkout the above repo
2. Run `yarn ios`
3. The app will freeze on launch (or if you try switching tabs)

I'll explain the issue in a bit more detail since it requires several conditions to be met.

I'm using the `react-navigation` library (not sure if the issue is present in `expo-router`). I have an Expo UI `Menu` component that lives on each screen. This menu contains a glass style `Button` as its label, which contains a RN native view (because I need to render a custom image). When an `animation` param is specified in the tab navigator, the app will completely freeze when rendering the UI. In my app, it renders the initial screen and freezes when switching tabs while in the demo app it freezes on launch. So it might be timing related.

If you remove the `animation` param OR remove the glass `buttonStyle` from the `Button` component, the issue is no longer present.

### Environment

```text
expo-env-info 2.0.11 environment info:
    System:
      OS: macOS 15.7
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 24.10.0 - ~/.nvm/versions/node/v24.10.0/bin/node
      Yarn: 1.22.22 - ~/.nvm/versions/node/v24.10.0/bin/yarn
      npm: 11.6.1 - ~/.nvm/versions/node/v24.10.0/bin/npm
      Watchman: 2025.06.23.00 - /opt/homebrew/bin/watchman
    Managers:
      CocoaPods: 1.16.2 - /Users/iberry/.gem/ruby/3.4.3/bin/pod
    SDKs:
      iOS SDK:
        Platforms: DriverKit 25.2, iOS 26.2, macOS 26.2, tvOS 26.2, visionOS 26.2, watchOS 26.2
    IDEs:
      Android Studio: 2025.2 AI-252.27397.103.2522.14514259
      Xcode: 26.3/17C529 - /usr/bin/xcodebuild
    npmPackages:
      expo: ~55.0.4 => 55.0.4
      react: 19.2.0 => 19.2.0
      react-native: 0.83.2 => 0.83.2
    npmGlobalPackages:
      expo-cli: 6.3.12
    Expo Workflow: managed
```

### Expo Doctor Diagnostics

```text
iberry@Ians-MacBook-Pro-4 expo-ui-repro % npx expo-doctor@latest
Need to install the following packages:
expo-doctor@1.18.10
[...]

PR #40017, opened 2025-09-26
[ui][ios] - set host dimension synchronously on native

intergalacticspacehighway:
# Why

We need to set yoga height/width style props synchronously from native to avoid layout shifts. This PR aims to match the current Host behaviour, move the dimension set call to native and make it sync. 

### Before

https://github.com/user-attachments/assets/82911650-26a0-441a-ab93-f819159ac4c0

### After

https://github.com/user-attachments/assets/1e1fb386-9a93-4eaf-81a8-8930b891f05e

# How
- Adds a new `setStyleSize` method, this behaves more like updating style props rather than setting the dimensions directly.
- It uses an API used by RN for Android [TextInput](https://github.com/facebook/react-native/blob/4f2755bda58ed627fe46376b337364484076b604/packages/react-native/ReactCommon/react/renderer/components/textinput/platform/android/react/renderer/components/androidtextinput/AndroidTextInputComponentDescriptor.h#L146) for some usecase. We could check with meta if this API is safe (it is public though). 
- We also trigger the state in sync using `EventQueue::UpdateMode::unstable_Immediate` that was introduced in RN 0.82.

# Test Plan

Test all examples in Expo UI

# Alternate approach

Alternative approach could be using reanimated like shadow tree update, but it will need UIManager instance. Will need to dig more how to get the instance, but I came across above API which seems suitable for our usecase.

# Note

In this approach dynamically changing `matchContent` might not behave correctly, we can hack it using `key` prop. But i don't think you'd normally have a usecase of changing `matchContents`.

# Checklist

- [x] I added a `changelog.md` entry and rebuilt the package sources according to [this short guide](https://github.com/expo/expo/blob/main/CONTRIBUTING.md#-before-submitting)
- [x] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: updated a module plugin).
- [x] Conforms with the [Documentation Writing Style Guide](https://github.com/expo/expo/blob/main/guides/Expo%20Documentation%20Writing%20Style%20Guide.md)

PR #48059, opened 2026-07-22
[expo-modules-core][iOS] Cap synchronous Host size commits per run-loop turn to prevent an infinite SwiftUI/Yoga layout loop

### LINKED ISSUE #48058: [SDK 57] iOS: @expo/ui Host with matchContents freezes the app (infinite layout loop) when Button Shapes accessibility setting is enabled

https://github.com/focux/expo-ui-sizeloop-repro

### Steps to reproduce

On iOS, mounting `@expo/ui` components that use `<Host matchContents>` (in the repro: `MenuView` from `@expo/ui/community/menu` with React Native children as the trigger) permanently blocks the main thread when the device has the **Button Shapes** accessibility setting enabled (Settings → Accessibility → Display & Text Size), combined with Bold Text / an enlarged Dynamic Type size. The UI freezes on the spot; on a physical device, backgrounding the app then gets it killed by the watchdog (SpringBoard shows the "app crashed" dialog).

1. On the simulator/device: Settings → Accessibility → Display & Text Size → enable **Bold Text** and **Button Shapes**, and set Larger Text to an accessibility size.
   <details><summary>Enable via CLI on a simulator</summary>

   The `defaults` writes only take effect after the simulator reboots; `content_size` applies immediately:

   ```sh
   UDID=XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX # from `xcrun simctl list devices booted`
   xcrun simctl spawn $UDID defaults write com.apple.Accessibility EnhancedTextLegibilityEnabled -bool YES # Bold Text
   xcrun simctl spawn $UDID defaults write com.apple.Accessibility ButtonShapesEnabled -bool YES           # Button Shapes
   xcrun simctl ui $UDID content_size accessibility-extra-extra-extra-large                                # Larger Text
   xcrun simctl shutdown $UDID && xcrun simctl boot $UDID
   ```
   </details>
2. `npm install && npx expo run:ios --configuration Release` (dev build works too — `@expo/ui` is not in Expo Go).
3. Tap "Open form sheet". After a short simulated loading phase the sheet mounts a few `MenuView` rows (trigger = a flex row with a width-constrained label and a right-aligned value) plus some `DateTimePicker` rows.
4. The heartbeat counter at the top of the sheet freezes and the content never renders. Without the accessibility settings from step 1, the same sheet renders normally.

**Expected:** the sheet renders; accessibility display settings only change how content looks.
**Observed:** permanent main-thread hang; watchdog kill on device after backgrounding.

Sampling the hung process shows ~100% of main-thread samples in this cycle, repeating inside a single UIUpdate cycle:

```
GeometryChangeModifier.dispatchOnLayoutContent (expo-ui HostView.swift)
→ -[ExpoFabricViewObjC setStyleSize:height:] (ExpoFabricViewObjC.mm:188)
→ ConcreteState::updateState(..., EventQueue::UpdateMode::unstable_Immediate)
→ synchronous ShadowTree::commit → Yoga layoutTree → updateClippedSubviews
→ host frame changes → SwiftUI re-measures → repeat
```

With these accessibility settings the SwiftUI-measured content size and the Yoga layout that results from committing it never converge, and since the commit is synchronous (`unstable_Immediate`, introduced in #40017), the feedback loop never yields the main thread. This looks like the same failure class as #43537 (different disagreement source — the #43570 refactor does not cover this one) and possibly the Android sibling #47625.

We hit this in production: a user with these accessibility settings got a watchdog kill on every open of a screen that mounts several such Hosts. I have a fix and will open a PR shortly.

### Environment

```

    System:
      OS: macOS 27.0
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 22.16.0 - /Users/leonardo/.nvm/versions/node/v22.16.0/bin/node
      Yarn: 1.22.22 - /Users/leonardo/.nvm/versions/node/v22.16.0/bin/yarn
      npm: 10.9.2 - /Users/leonardo/.nvm/versions/node/v22.16.0/bin/npm
      Watchman: 2025.05.26.00 - /opt/homebrew/bin/watchman
    Managers:
    SDKs:
      iOS SDK:
      Android SDK:
        API Levels: 23, 24, 28, 29, 30, 31, 32, 33, 34, 34, 35
        Build Tools: 28.0.3, 29.0.2, 30.0.2, 30.0.3, 31.0.0, 32.0.0, 33.0.0, 33.0.1, 34.0.0, 35.0.0, 35.0.0
        System Images: android-25 | ARM 64 v8a, android-29 | Intel x86 Atom_64, android-31 | Google Play ARM 64 v8a, android-32 | Google APIs ARM 64 v8a, android-36 | Google Play ARM 64 v8a
    IDEs:
      Android Studio: 2024.2 AI-242.23339.11.2421.12700392
    npmPackages:
      expo: ^57.0.1 => 57.0.1 
      react: 19.2.3 => 19.2.3 
    npmGlobalPackages:
      eas-cli: 20.1.0
    Expo Workflow: managed
```

### Expo Doctor Diagnostics

```
```

focux:
# Why

`ExpoFabricViewObjC`'s `setStyleSize`/`setShadowNodeSize` commit shadow node state synchronously (`EventQueue::UpdateMode::unstable_Immediate`, introduced in #40017 so a `matchContents` Host adopts its measured content size within the current frame, without a one-frame layout shift).

That is only safe while the SwiftUI measurement and the resulting Yoga layout converge. When they persistently disagree — reproducibly triggered by enabling the iOS **Button Shapes** accessibility setting together with Bold Text / an enlarged Dynamic Type size — every synchronous commit re-triggers a measurement in the same run-loop turn:

```
GeometryChangeModifier.dispatchOnLayoutContent → setStyleSize
→ unstable_Immediate ShadowTree commit → Yoga relayout
→ host frame changes → SwiftUI re-measures → repeat
```

The main thread never yields, the UI freezes, and the iOS watchdog kills the app when it's backgrounded. Same failure class as #43537 (the #43570 refactor doesn't cover this trigger); #47625 looks like the Android sibling of the same synchronous-dispatch pattern.

Fixes #48058

# How

Budget the synchronous path instead of removing it:

- Each view may issue 2 `unstable_Immediate` size commits per main-queue turn. The first commits keep the synchronous path, so the common convergent case still adopts the content size within the current frame, exactly as #40017 intended.
- Over-budget updates are **coalesced locally on the view** (latest value per state channel) and only entered into the event queue — as a regular `Asynchronous` update — when the next main-queue turn runs. They can't be enqueued immediately: RN has a single shared `EventQueue`, and an `unstable_Immediate` update from *any other view* flushes the whole queue synchronously (`EventQueue.cpp`), which would pull the "deferred" update right back into the loop being broken.
- State updates now go through the `updateState(oldData)` callback, merging the `width/height` and `styleWidth/styleHeight` channels instead of replacing the whole `ExpoViewState` (previously `setStyleSize` reset the shadow-size channel to NaN and vice versa), and returning `nullptr` to cancel commits when the size hasn't meaningfully changed (±0.05pt), so no-op updates no longer dirty the tree.
- The guard state resets each turn and on `prepareForRecycle`/`invalidate` (with a generation token so a stale reset block can't touch a reused view), and the two setters are annotated `NS_SWIFT_UI_ACTOR` with a main-queue assert — they were already main-thread-only in practice.
- `ExpoViewComponentDescriptor::adopt()` now compares the existing Yoga style dimension before calling `setDimension` + `dirtyLayout()`; previously every node clone re-dirtied layout unconditionally even when the state carried the same value.

No behavior change when layout converges within the budget — which is every case except a runaway feedback loop. `SwiftUIVirtualViewSharedImpl` duplicates the same immediate setters and could get the same guard as a follow-up.

# Test Plan

Manual repro from #48058 (https://github.com/focux/expo-ui-sizeloop-repro): enable Button Shapes + Bold Text + an accessibility Dynamic Type size, open a form sheet that mounts several `matchContents` Hosts (`@expo/ui` `MenuView` rows).

- Before: main thread loops forever in `GeometryChangeModifier → setStyleSize` (~100% of `sample` frames, stack in the issue), sheet frozen showing only a spinner, watchdog kill on device after backgrounding.
- After: the sheet renders and stays responsive across repeated open/close cycles; process samples show the loop gone.
- With default accessibility settings, verified hosts still adopt their content size with no visible one-frame shift (the #40017 behavior), on both the repro app and a production app (BenchApp) with several menus + compact date/time pickers mounting at once.

# Checklist


fobos531:
Following this! I got user reports of issues which sound suspiciously similar to the issue described here, so I'm looking forward to finding out if it fixes it for me.

intergalacticspacehighway:
@focux i think we should merge this PR as a failsafe but in the original repro we can fix the crash by making `Menu` label `Text` not dependent of parent dimensions. Adding `align-self` `flex-start` and removing flex: 1.

<details>
<summary>Repro Code changes</summary>

```jsx
/**
 * Repro: SwiftUI <-> Yoga sizing feedback loop on iOS.
 *
 * Requires iOS Accessibility settings: Bold Text + Button Shapes, and an
 * enlarged Dynamic Type size (Settings -> Accessibility -> Display & Text
 * Size). With those enabled, opening the form sheet below locks the main
 * thread (the heartbeat counter freezes and the pickers never render).
 * Backgrounding the app at that point causes iOS's watchdog to kill it.
 */
import DateTimePicker from '@expo/ui/community/datetime-picker';
import { MenuView } from '@expo/ui/community/menu';
import { useEffect, useState } from 'react';
import { ActivityIndicator, Button, Modal, ScrollView, StyleSheet, Text, View } from 'react-native';

function Heartbeat() {
  const [tick, setTick] = useState(0);
  useEffect(() => {
    const id = setInterval(() => setTick((t) => t + 1), 100);
    return () => clearInterval(id);
  }, []);
  return <Text style={styles.heartbeat}>Heartbeat: {tick}</Text>;
}

function PickerRow({ n }: { n: number }) {
  return (
    <View style={styles.row}>
      <Text style={styles.rowLabel}>Label {n}</Text>
      <View style={styles.rowPicker}>
        <DateTimePicker
          value={new Date()}
          mode={n % 2 ? 'date' : 'time'}
          display="compact"
          presentation="inline"
          style={{ width: 150, height: 36 }}
          onValueChange={() => {}}
        />
      </View>
    </View>
  );
}

function MenuRow({ label, value }: { label: string; value: string }) {
  return (
    <MenuView
      actions={[
        { id: 'a', title: 'Option A' },
        { id: 'b', title: 'Option B' },
      ]}
      onPressAction={() => {}}>
      <View style={styles.menuItem}>
        <View style={styles.menuLabelBox}>
          <Text style={styles.menuLabelText}>{label}</Text>
        </View>
        <View style={styles.menuValueBox}>
          <Text numberOfLines={1} ellipsizeMode="tail" style={styles.menuValueText}>
            {value}
          </Text>
          <Text style={styles.menuChevron}>›</Text>
        </View>
      </View>
    </MenuView>
  );
}

function FormSheetContent() {
  const [loading, setLoading] = useState(true);
  useEffect(() => {
    const id = setTimeout(() => setLoading(false), 1500);
    return () => clearTimeout(id);
  }, []);

  return (
    <View style={styles.sheet}>
      <Heartbeat />
      {loading ? (
        <View style={styles.center}>
          <ActivityIndicator size="large" />
        </View>
      ) : (
        <ScrollView>
          <View style={styles.menuSection}>
            <MenuRow label="Event Type" value="Game (Regular)" />
            <MenuRow label="Home or Away" value="Home" />
          </View>
          {[1, 2, 3, 4].map((n) => (
            <PickerRow key={n} n={n} />
          ))}
          <View style={styles.menuSection}>
            <MenuRow label="Repeat" value="Never" />
          </View>
        </ScrollView>
      )}
    </View>
  );
}

  const [visible, setVisible] = useState(false);

  return (
    <View style={styles.container}>
      <Text style={styles.title}>expo-ui sizeloop repro</Text>
      <Heartbeat />
      <ActivityIndicator style={styles.spinner} />
      <Button title="Open form sheet" onPress={() => setVisible(true)} />
      <Modal
        visible={visible}
        presentationStyle="formSheet"
        animationType="slide"
        onRequestClose={() => setVisible(false)}
      >
        <FormSheetContent />
        <Button title="Close" onPress={() => setVisible(false)} />
      </Modal>
    </View>
  );
}

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
    gap: 12,
  },
  title: {
    fontSize: 20,
    fontWeight: '600',
  },
  spinner: {
    marginVertical: 4,
  },
  heartbeat: {
    fontSize: 17,
    fontVariant: ['tabular-nums'],
  },
  sheet: {
    flex: 1,
    paddingTop: 24,
  },
  center: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
  },
  row: {
    flexDirection: 'row',
    alignItems: 'center',
    paddingHorizontal: 16,
    paddingVertical: 8,
  },
  rowLabel: {
    flex: 1,
    fontSize: 17,
  },
  rowPicker: {
    minWidth: 0,
    flex: 1,
    alignItems: 'flex-end',
  },
  menuSection: {
    marginHorizontal: 16,
    marginBottom: 24,
    backgroundColor: '#fff',
    paddingHorizontal: 16,
    paddingVertical: 8,
    borderRadius: 10,
    rowGap: 8,
  },
  menuItem: {
    flexDirection: 'row',
    alignItems: 'center',
    paddingVertical: 4,
    minHeight: 44,
    // Inside a matchContents Host, the label's width must not derive from the
    // granted width: opt out of Yoga's stretch default and cap wrapping with a
    // constant, so the measured size is grant-independent (no layout loop).
    alignSelf: 'flex-start',
    gap: 12,
    maxWidth: 300,
  },
  menuLabelBox: {
    maxWidth: 100,
    flexDirection: 'row',
    alignItems: 'center',
  },
  menuLabelText: {
    fontSize: 18,
    flexShrink: 0,
  },
  menuValueBox: {
    minWidth: 0,
    flexDirection: 'row',
    justifyContent: 'flex-end',
    alignItems: 'center',
  },
  menuValueText: {
    minWidth: 0,
    flexShrink: 1,
    textAlign: 'right',
    fontSize: 17,
    color: '#888',
  },
  menuChevron: {
    marginLeft: 12,
    marginTop: 2,
    fontSize: 15,
    color: '#aaa',
  },
});

```
</details>

intergalacticspacehighway (approved):
LGTM. Will look into the android equivalent.

fobos531:
@intergalacticspacehighway Do you have an update on this?

intergalacticspacehighway:
@fobos531 i am working on an alternate fix which should prevent such loops. Aiming to get it out by early next week. For now can you try passing `alignSelf: "flex-start"` style to `RNHostView`? The loop happens when `Host` `matchContents` and `RNHostView` `matchContents` are looping because `alignItem: "stretch"` is the default, so they keep on sizing each other.

fobos531:
> @fobos531 i am working on an alternate fix which should prevent such loops. Aiming to get it out by early next week. For now can you try passing `alignSelf: "flex-start"` style to `RNHostView`? The loop happens when `Host` `matchContents` and `RNHostView` `matchContents` are looping because `alignItem: "stretch"` is the default, so they keep on sizing each other.

Hello @intergalacticspacehighway - yes, that works. With the alternate fix that you mentioned, will the alignSelf still be necessary?

intergalacticspacehighway:
> will the alignSelf still be necessary?

@fobos531 no, alignSelf will be implicitly added to prevent such loops. Check [here](https://github.com/expo/expo/pull/48970/changes#diff-ee84659ee4dddb4d2dcae9d4539a4262a1ef060d7a52e18c1fb9f43013adcc6aR54) and https://github.com/expo/expo/pull/48970

EvanBacon:
@brentvatne can we get this landed 🙏

brentvatne:
@expo-bot check --fable

intergalacticspacehighway:
Merging this PR as a failsafe, it will be up in the next 57 release. It stops the bleeding but the loop would still exist, it's just throttled. 

Below case is enough to trigger a loop.
```jsx
<Host matchContents>
  <VStack modifiers={[padding({ all: 8 })]}>
    <RNHostView matchContents>
      <Text style={{ fontSize: 24, color: theme.text.default }}>Hello, world!</Text>
    </RNHostView>
  </VStack>
</Host>
```

`Host` and `RNHostView` both have `matchContents`. `Host` width on 1st pass = 16 (padding) + 100 (text width) = 116. In next pass `RNHostView` expands to `116` because it has `align-self: stretch`. `Host` receives `116 + 16` = `132`, `RNHostView` expands again and it keeps going on.

The main issue here is that `VStack's` shadow node (yoga node) has no idea about the `padding` since it is set by SwiftUI. So `RNHostView` expands fully to the `Host`. 

Passing `align-self: flex-start` also won't break the loop in cases where a child has `flex: 1` and `undefined` parent height (in the repro shared by @focux and fix provided [here](https://github.com/expo/expo/pull/48059#issuecomment-5123507776)).

The fix for above would be to pass `matchContents` `vertical: true` in `Host`. This breaks the loop, `Host` won't size horizontally to the SwiftUI content.

```jsx
<Host matchContents={{vertical: true}}>
  <VStack modifiers={[padding({ all: 8 })]}>
    <RNHostView matchContents>
      <Text style={{ fontSize: 24, color: theme.text.default }}>Hello, world!</Text>
    </RNHostView>
  </VStack>
</Host>
```

Some accessibility API like `Button Shapes` adds a box around SwiftUI items, this added dimension is not known to Shadow node, it works just like the above `padding: 8` case.


For an ideal fix, when we have `RNHostView` with `matchContents`, RNHostView's yoga node should not consult the parent yoga node. The dimension should be solely driven by child. There are a few ways to do it.

1. Using [measure](https://github.com/react/react-native/blob/6c4cf29689d6c330239bdaf1b974eff80552f35f/packages/react-native/ReactCommon/react/renderer/core/LayoutableShadowNode.h#L93) API from `LayoutableShadowNode.h`. We'll have to run it before the actual layout happens, a commit hook can work. Downside is that it adds quite a few lines of code and adds some measure overhead 😅  
2. Supporting width/height `max-content` unit in Yoga. This would be ideal. We can pass width/height as max-content to `RNHostView` and it will stop consulting the parent node. 

Currently I am testing above approaches. Until then if you face any such loops, try setting single axis to `true` in `Host` `matchContents` or removing flexible child in `RNHostView` or setting `alignSelf` `flex-start` in `RNHostView` 😅. One of these should fix the loop!

matinzd:
This PR caused a regression in `expo-modules-core@57.0.12` where swiftui layouts nested inside RN views cannot be calculated properly in our use case. I could not reproduce it in a simple app but in our app at least I am sure that this PR broke the layout measurement.

I guess this comment from the agent makes sense:

> Yoga node can keep a stale size because the guard compares the shared props style, not the node style 

https://github.com/user-attachments/assets/1aed9701-d888-4ca8-8cd0-ef9434d378f9

image


Pinning `expo-modules-core` to `57.0.11` fixes the issue.

intergalacticspacehighway:
@matinzd thanks for reporting, can you try this [patch](https://github.com/expo/expo/pull/49211) and lmk if it fixes the issue?

matinzd:
> @matinzd thanks for reporting, can you try this [patch](https://github.com/expo/expo/pull/49211) and lmk if it fixes the issue?

It works!

PR #48065, opened 2026-07-23
[brownfield][ios] Fix build:ios packaging bugs with precompiled modules

gabrieldonadel:
# Why

While integrating expo-brownfield in https://github.com/expo/expo-brownfield-examples a few bugs surfaced in the CLI that break iOS packaging when precompiled modules (`ios.usePrecompiledModules`) are enabled:
 

1. `build:ios` fails to locate the built brownfield framework because the pod build settings route it under `XCFrameworkIntermediates/<scheme>/` rather than the products-dir root.
2. The generated `Package.swift` can declare the same target twice (when the brownfield target shares its name with a bundled module), producing a manifest Xcode refuses to resolve. The bundled module's xcframework also silently overwrites the wrapper framework on disk, shipping a package without `ReactNativeHostManager`.
3. `ExpoModulesJSI.xcframework` is never bundled even though the prebuilt `ExpoModulesCore` publicly depends on it, so consumers fail to compile with `unable to resolve module dependency: 'ExpoModulesJSI'`.
4. Shared SPM-dep xcframeworks (e.g. `SDWebImage`) are copied as symlinks pointing into `node_modules/<pkg>/prebuilds/spm-deps/<flavor>/`, making the package non-portable and able to ship debug binaries in a release build when the flavor at `pod install` time differs from the requested one.

# How

- **Framework resolution:** `createXCframework` and `enumerateSourceBuiltDeps` now resolve the scheme framework through the existing `findSourceBuiltFramework` helper (products root first, then `XCFrameworkIntermediates/`). A missing framework now fails with an actionable error instead of a raw `xcodebuild` failure. This also revived the source-built dependency scan, which was silently returning empty in precompiled builds.
- **Duplicate targets:** `generatePackageMetadataFile` dedupes targets by name, and a new `validateSchemeCollision` check fails fast (before the long `xcodebuild` step) when the scheme name collides with a bundled framework, telling the user to rename `ios.targetName`.
- **Pod-built vendored xcframeworks:** a new enumeration layer picks up pods that build their vendored xcframework during the app build (`Pods/<Pod>/Products/<Pod>.xcframework`, the `ExpoModulesJSI` pattern) so they are bundled and declared as binary targets alongside the other layers.
- **Symlinked SPM deps:** enumeration resolves staged symlinks to their real paths and swaps to the requested flavor's sibling store directory. If that flavor is not on disk, the build fails with a flavor-mismatch error rather than silently mixing debug and release binaries. Host-provided frameworks are exempt from this check.

Also fixes `build:ios` not awaiting the ship step, and wires up unit testing for the CLI (previously untested) using the repo-standard composite jest preset.

# Test Plan

- Added 13 unit tests in `cli/src/utils/__tests__/` covering all four fixes: framework resolution at both locations plus the error path, `Package.swift` dedup, scheme collision validation, pod-built xcframework enumeration, and symlink flavor correction including the mismatch and host-provided cases. Tests were written red-first against the buggy behavior.
- `et check-packages expo-brownfield` passes (build, typecheck, lint, tests).

### DIFF
--- a/packages/expo-brownfield/CHANGELOG.md
+++ b/packages/expo-brownfield/CHANGELOG.md
@@ -11,7 +11,7 @@
 
 ### 🐛 Bug fixes
 
-- [iOS] Fix `build:ios` with precompiled modules: locate frameworks under `XCFrameworkIntermediates/`, bundle `ExpoModulesJSI`, copy SPM deps as real flavor-matched directories instead of symlinks, and fail fast on duplicate or colliding target names.
+- [iOS] Fix `build:ios` with precompiled modules: locate frameworks under `XCFrameworkIntermediates/`, bundle `ExpoModulesJSI`, copy SPM deps as real flavor-matched directories instead of symlinks, and fail fast on duplicate or colliding target names. ([#48065](https://github.com/expo/expo/pull/48065) by [@gabrieldonadel](https://github.com/gabrieldonadel))
 
 ### 💡 Others
 
--- a/packages/expo-brownfield/CHANGELOG.md
+++ b/packages/expo-brownfield/CHANGELOG.md
@@ -11,6 +11,8 @@
 
 ### 🐛 Bug fixes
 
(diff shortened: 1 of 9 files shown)


chrfalch (approved):
Looks good!

PR #48084, opened 2026-07-23
[android][navigation-bar] Handle rejected declarative updates

magnesae:
# Why

The declarative `NavigationBar` component updates Android through asynchronous native methods, but its internal `setStyle` and `setHidden` helpers do not return those promises to the caller.

If an update races with activity teardown, the native promise can reject after the component has already unmounted. Because this internal best-effort update has no consumer, the rejection becomes an unhandled promise rejection and can surface as a LogBox error.

# How

Consume rejections from the two declarative native updates. The public async visibility API remains unchanged and still propagates its rejection to callers.

Added focused Android unit coverage that verifies both internal update paths attach rejection handlers.

# Test Plan

- `pnpm run lint`
- Targeted Expo formatter check for both changed TypeScript files
- Focused Android Jest suite: 2 tests passed
- `git diff --check`

The focused Jest suite was run with the repository's installed `@swc/jest` transformer to isolate it from unrelated sparse-checkout workspace packages.

# Checklist

- [x] I added a changelog entry.
- [x] This diff does not change a config plugin or prebuild output.
- [x] Conforms with the Expo JavaScript style.

### DIFF
--- a/packages/expo-navigation-bar/src/__tests__/NavigationBar-test.android.ts
+++ b/packages/expo-navigation-bar/src/__tests__/NavigationBar-test.android.ts
@@ -1,0 +1,41 @@
+import ExpoNavigationBar from '../ExpoNavigationBar';
+import { NavigationBar, setStyle } from '../NavigationBar.android';
+
+jest.mock('../ExpoNavigationBar', () => ({
+  __esModule: true,
+  default: {
+    setStyle: jest.fn(),
+    setHidden: jest.fn(),
+  },
+}));
+
+jest.mock('react-native', () => ({
+  Appearance: {
+    getColorScheme: () => 'light',
+  },
+  useColorScheme: () => 'light',
+}));
+
+const mockExpoNavigationBar = jest.mocked(ExpoNavigationBar);
+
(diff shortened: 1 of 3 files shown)


packages/expo-navigation-bar/src/__tests__/NavigationBar-test.android.ts
```diff
@@ -0,0 +1,41 @@
...
+}));
+
+const mockExpoNavigationBar = jest.mocked(ExpoNavigationBar);
+
+describe('NavigationBar', () => {
+  it('handles rejected native style updates', () => {
+    const catchError = jest.fn();
+    mockExpoNavigationBar.setStyle.mockReturnValue({ catch: catchError } as never);
+
+    setStyle('dark');
```
zoontek, line 26: Could you use `NavigationBar.setStyle('dark');` instead?

zoontek (changes requested):

magnesae:
Closing this contribution. It was opened without the repository owner's explicit authorization to submit upstream changes. Sorry for the noise.

Issue #35385, opened 2025-03-09
java.lang.RuntimeException: Unable to start activity ComponentInfo{com.practice.example/com.practice.example.MainActivity}: java.lang.IllegalArgumentException: App react context shouldn't be created before


Ikshit001:
### Summary

seeing random crashes in android app

### Managed or bare workflow?

managed

### What platform(s) does this occur on?

Android

### Environment

```text
expo-env-info 1.2.2 environment info:
    System:
      OS: macOS 14.3.1
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 23.4.0 - /opt/homebrew/bin/node
      Yarn: 3.6.4 - /usr/local/bin/yarn
      npm: 10.9.2 - /opt/homebrew/bin/npm
      Watchman: 2024.12.02.00 - /opt/homebrew/bin/watchman
    Managers:
      CocoaPods: 1.16.2 - /opt/homebrew/bin/pod
    SDKs:
      iOS SDK:
        Platforms: DriverKit 23.2, iOS 17.2, macOS 14.2, tvOS 17.2, visionOS 1.0, watchOS 10.2
      Android SDK:
        API Levels: 29, 30, 31, 32, 33, 34, 35
        Build Tools: 29.0.2, 30.0.2, 30.0.3, 31.0.0, 33.0.0, 34.0.0, 35.0.0
        System Images: android-29 | Google Play Intel x86 Atom, android-30 | Google Play Intel x86 Atom, android-33 | Google APIs ARM 64 v8a, android-35 | Intel x86_64 Atom, android-Baklava | Google Play ARM 64 v8a
    IDEs:
      Android Studio: 2024.2 AI-242.23339.11.2421.12700392
      Xcode: 15.2/15C500b - /usr/bin/xcodebuild
    npmPackages:
      expo: ~52.0.20 => 52.0.37 
      expo-router: ~4.0.15 => 4.0.17 
      react: 18.3.1 => 18.3.1 
      react-dom: 18.3.1 => 18.3.1 
      react-native: 0.76.7 => 0.76.7 
      react-native-web: ~0.19.13 => 0.19.13 
    npmGlobalPackages:
      eas-cli: 14.2.0
    Expo Workflow: bare
```

### Reproducible demo

N/A

### Stacktrace (if a crash is involved)

```text
[...]

PR #48085, opened 2026-07-23
[android][dev-launcher] Reload an existing React context

magnesae:
# Why

`DevLauncherAppLoader` currently requires the shared `ReactHost` to have no React context before the selected app's activity delegate is created.

That invariant does not always hold. A background/headless React task can initialize the process first—for example, an Android home-screen widget update—or a development build can be reopened while a context from the previous launch still exists. The development activity then crashes at startup with:

```text
IllegalArgumentException: App react context shouldn't be created before.
```

The same stack was previously reported in #35385, but that issue was auto-closed for lacking a minimal reproduction; the assertion remains on `main`.

# How

Capture an existing context, install Dev Launcher's React instance listener, then ask the public `ReactHost.reload` API to replace that context with the bundle loader already selected by Dev Launcher.

Registering the listener before reloading ensures the replacement context still follows the normal `onAppLoaded` / `onReactContext` path.

A focused Robolectric/MockK regression test covers the pre-existing-context case and verifies the reload request.

# Test Plan

- Added `DevLauncherAppLoaderTest` under the package's existing `src/testDebug` Robolectric suite.
- Verified the production call against React Native 0.86's public `ReactHost.reload(reason)` API.
- `git diff --check`

I did not run the Android Gradle suite locally because this contribution was prepared from a sparse package checkout; Expo CI has the full native workspace and will run the added test.

# Checklist

- [x] I added a changelog entry.
- [x] This is debug-only Android code and does not affect release builds or prebuild output.
- [x] The fix uses a public React Native API.

### DIFF
--- a/packages/expo-dev-launcher/CHANGELOG.md
+++ b/packages/expo-dev-launcher/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 ### 🐛 Bug fixes
 
+- [Android] Reload an existing React context instead of crashing when launching a selected app.
 - [iOS] Use `RCTPlatformName` instead of hardcoding `ios` when requesting bundles from Metro. ([#46443](https://github.com/expo/expo/pull/46443) by [@gabrieldonadel](https://github.com/gabrieldonadel))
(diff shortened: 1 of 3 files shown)


magnesae:
Closing this contribution. It was opened without the repository owner's explicit authorization to submit upstream changes. Sorry for the noise.

PR #48087, opened 2026-07-23
[router] respect `Cache-Control` headers in `useLoaderData`

hassankhan:
# Why

Currently, data loaders cache their results after the `useLoaderData()` hook is invoked with no method of invalidating the cache.

This PR improves `useLoaderData()` (and associated machinery) to use the platform cache (`NSURLCache` for iOS, `OkHttp` for Android, and the browser cache on web). The hook will now re-fetch data on every mount and let the platform decide whether the request should hit the network or not.

Two deliberate behavior changes come with this: 

1. Loaders without a `Cache-Control` header no longer cache forever (they re-fetch or revalidate as per the defaults below)
2. Revisits render a brief Suspense fallback since the in-memory cache that made them synchronous is gone (this should be resolved by the upcoming support for transitions).

# How

## `expo-router`

- Data persistence has been removed, and a fresh mount will now always result in a re-fetch.
- The injected seed data is consumed once into the suspense store. It serves the initial hydration render, then follows the normal mount lifecycle, so revisits to the initial route re-fetch like any other route. (Previously, the injected global was never consumed, and the initial route never re-fetched at all.)
- Loader errors now live in the suspense store and expire once delivered to the error boundary, so `ErrorBoundary`'s `retry()` now re-fetches instead of re-throwing the cached error.
- On a HMR `loader-invalidate` event, the router appends a versioning query parameter to loader URLs to bust the cache since platform caches don't expose a way to evict entries, and `RequestInit.cache` doesn't work on native.

## `@expo/cli`

- The dev server strips the cache-busting query parameter before request handling, so it never reaches loader code.

# Test Plan

- CI
- Manual testing with the `server-loader` fixture in `apps/router-e2e`

# Checklist


packages/expo-router/src/hooks/useLoaderData.ts
```diff
@@ -35,13 +35,13 @@ type LoaderFunctionResult<T extends LoaderFunction<any>> =
...
   const serverDataLoaderContext = use(ServerDataLoaderContext);
-  const loaderCache = use(LoaderCacheContext);
+  const loaderClient = use(LoaderClientContext);
 
   // Subscribe before any early returns so a later `loader-invalidate` re-renders this hook even
   // when the initial render was satisfied by `ServerDataLoaderContext` or `__EXPO_ROUTER_LOADER_DATA__`.
   // Returning early before subscribing would also change hook order on the next render once
   // invalidation deletes the injected global.
-  useSyncExternalStore(loaderCache.subscribe, loaderCache.getSnapshot, loaderCache.getSnapshot);
+  useSyncExternalStore(loaderClient.subscribe, loaderClient.getSnapshot, loaderClient.getSnapshot);
```
kitten, line 44: Already discussed this, but would prefer to erase `useSyncExternalStore` eventually, provided nothing weird happens to React in a hypothetical v20

packages/expo-router/src/loaders/LoaderClient.ts
```diff
@@ -0,0 +1,159 @@
...
+  private fetchers = new Map<string, LoaderFetcher>();
+  private version = 0;
+  private listeners = new Set<() => void>();
+
+  readonly suspense = new LoaderSuspenseStore();
+
+  // Arrow-bound so `loaderClient.subscribe` returns a stable reference across renders,
+  // which keeps `useSyncExternalStore()` from tearing down and re-attaching every render.
+  subscribe = (listener: () => void): (() => void) => {
+    this.listeners.add(listener);
```
kitten, line 27: I guess this is an interesting one, since, while it still uses `useSyncExternalStore`, it won't actually deliver an immediate message and hence `getSnapshot` and `subscribe` are basically "noop updates" i.e. no-update/no-change on mount.

That's basically another argument I guess for replacing `useSyncExternalStore` apart from the semantic differences when it's used in a component to how React processes updates

packages/expo-router/src/loaders/readLoaderData.ts
```diff
@@ -1,78 +1,35 @@
...
-    (data) => {
-      cache.suspense.set(resolvedPath, { data });
-      return data;
-    },
-    (error) => {
-      // The error is held in the cache; dropping the store entry lets a re-read re-throw it.
-      cache.suspense.clear(resolvedPath);
-      throw error;
+    if ('error' in suspended) {
+      client.suspense.expireError(resolvedPath);
```
kitten: nit: I'd argue, the logic in `suspense` should either happen on `get` or should happen here. The reason is that we already have `suspended`, so it's convenient to invalidate it here. If we don't do this here, then we have to "redo" the entire `entries.get` in `expireError`, which then uncleanly absorbs that detail there. So, it feels a bit between places right now

packages/expo-router/src/loaders/LoaderClient.ts
```diff
@@ -0,0 +1,159 @@
...
+  ending: boolean;
+}
+
+export class LoaderClient {
+  private active = new Map<string, LoaderSource>();
+  private fetchers = new Map<string, LoaderFetcher>();
+  private version = 0;
+  private listeners = new Set<() => void>();
+
+  readonly suspense = new LoaderSuspenseStore();
```
kitten, line 22: nit/follow-up: the reason this was separated in `urql` is to stop indicating in code that the client is ever "allowed" to access the suspense cache, to basically make the isolation and cache stacking clearer. I think with agents this is a nice thing to replicate, since it stops them from ever writing code that misses this constraint

Edit: that's basically the main constraint that can make the hook implementation confusing, so the isolation may be worthwhile to further emphasise the layering

kitten (approved):

packages/expo-router/src/hooks/useLoaderData.ts
hassankhan: Yep, will address this in #48523

packages/expo-router/src/loaders/LoaderClient.ts
hassankhan: Addressed in #48563

packages/expo-router/src/loaders/LoaderClient.ts
hassankhan: Addressed in #48523

PR #48089, opened 2026-07-23
[file-system] Add File.digest() for calculating various hashes and deprecate old md5 usage

wh201906:
# Why

As suggested in https://github.com/expo/expo/pull/46403#pullrequestreview-4753352185, this PR adds an async `File.digest()` to calculate hashes like MD5/SHA-1/SHA-256/SHA-384/SHA-512, which exposes more digest algorithms supported by both Android and iOS. The async implementation also makes the UI more responsive.
This PR also adds `@deprecated` tag to the existing MD5 property/argument/results, as suggested in the same PR comment above.

# How

- Add `FileDigestAlgorithm` at TS side to limit the input to the supported digest algorithms
- Add native implementation of `FileSystemFile.digest()` on both Android and iOS, with chunk reading.
  - For iOS, Add `calculateDigest()` so the type of hasher can be determined during the loop.
- Add `@deprecated` for `FileInfo.md5`, `InfoOptions.md5`, `InfoOptions` itself and `NativeFileSystemFile.md5`
- Add tests
- Update downstream calls

# Test Plan

1. `pnpm et check expo-file-system` passed in a clean worktree
2. Filesystem related tests in `apps/bare-expo` passed on Android

<details><summary>Screenshot</summary>

image

</details>

3. Filesystem related tests in `apps/bare-expo` passed on iOS

<details><summary>Screenshot</summary>

image

</details>

# Checklist

### DIFF
--- a/packages/expo-file-system/android/src/main/java/expo/modules/filesystem/FileSystemFile.kt
+++ b/packages/expo-file-system/android/src/main/java/expo/modules/filesystem/FileSystemFile.kt
@@ -174,12 +174,23 @@ class FileSystemFile(uri: Uri) : FileSystemPath(uri) {
     return file.getContentUri(appContext ?: throw MissingAppContextException())
   }
 
-  @OptIn(ExperimentalStdlibApi::class)
   val md5: String get() {
+    return digest("md5")
+  }
+
+  @OptIn(ExperimentalStdlibApi::class)
+  fun digest(algorithm: String): String {
     val bufferSize = 65536
 
     validatePermission(FilePermissionService.Permission.READ)
-    val md = MessageDigest.getInstance("MD5")
+    val md = when (algorithm) {
+      "md5" -> MessageDigest.getInstance("MD5")
+      "sha-1" -> MessageDigest.getInstance("SHA-1")
+      "sha-256" -> MessageDigest.getInstance("SHA-256")
+      "sha-384" -> MessageDigest.getInstance("SHA-384")
+      "sha-512" -> MessageDigest.getInstance("SHA-512")
+      else -> throw IllegalArgumentException("Unsupported digest algorithm: $algorithm")
+    }
     file.inputStream().use { stream ->
       val buffer = ByteArray(bufferSize)
       var bytesRead: Int
--- a/packages/expo-file-system/android/src/main/java/expo/modules/filesystem/FileSystemFile.kt
+++ b/packages/expo-file-system/android/src/main/java/expo/modules/filesystem/FileSystemFile.kt
@@ -175,7 +175,7 @@ class FileSystemFile(uri: Uri) : FileSystemPath(uri) {
   }
 
   val md5: String get() {
-    return digest("md5")
+    return digest("MD5")
   }
 
   @OptIn(ExperimentalStdlibApi::class)
@@ -184,11 +184,11 @@ class FileSystemFile(uri: Uri) : FileSystemPath(uri) {
 
     validatePermission(FilePermissionService.Permission.READ)
     val md = when (algorithm) {
-      "md5" -> MessageDigest.getInstance("MD5")
-      "sha-1" -> MessageDigest.getInstance("SHA-1")
-      "sha-256" -> MessageDigest.getInstance("SHA-256")
-      "sha-384" -> MessageDigest.getInstance("SHA-384")
-      "sha-512" -> MessageDigest.getInstance("SHA-512")
+      "MD5" -> MessageDigest.getInstance("MD5")
+      "SHA-1" -> MessageDigest.getInstance("SHA-1")
+      "SHA-256" -> MessageDigest.getInstance("SHA-256")
+      "SHA-384" -> MessageDigest.getInstance("SHA-384")
+      "SHA-512" -> MessageDigest.getInstance("SHA-512")
       else -> throw IllegalArgumentException("Unsupported digest algorithm: $algorithm")
     }
     file.inputStream().use { stream ->
--- a/packages/expo-file-system/android/src/main/java/expo/modules/filesystem/FileSystemFile.kt
+++ b/packages/expo-file-system/android/src/main/java/expo/modules/filesystem/FileSystemFile.kt
@@ -182,6 +182,7 @@ class FileSystemFile(uri: Uri) : FileSystemPath(uri) {
   fun digest(algorithm: String): String {
     val bufferSize = 65536
 
+    validateType()
     validatePermission(FilePermissionService.Permission.READ)
     val md = when (algorithm) {
       "MD5" -> MessageDigest.getInstance("MD5")
@@ -189,7 +190,7 @@ class FileSystemFile(uri: Uri) : FileSystemPath(uri) {
       "SHA-256" -> MessageDigest.getInstance("SHA-256")
       "SHA-384" -> MessageDigest.getInstance("SHA-384")
       "SHA-512" -> MessageDigest.getInstance("SHA-512")
-      else -> throw IllegalArgumentException("Unsupported digest algorithm: $algorithm")
+      else -> throw UnsupportedDigestAlgorithmException(algorithm)
     }
     file.inputStream().use { stream ->
       val buffer = ByteArray(bufferSize)
--- a/packages/expo-file-system/src/File.types.ts
+++ b/packages/expo-file-system/src/File.types.ts
@@ -226,15 +226,26 @@ export type FileInfo = {
   creationTime?: number;
   /**
    * Present if the `md5` option was truthy. Contains the MD5 hash of the file.
+   *
+   * @deprecated Use `await file.digest('md5')` instead.
    */
   md5?: string;
 };
 
+/**
+ * Algorithm used to calculate a file digest.
+ */
+export type FileDigestAlgorithm = 'md5' | 'sha-1' | 'sha-256' | 'sha-384' | 'sha-512';
+
+/**
+ * @deprecated Use `await file.digest('md5')` to calculate an MD5 digest.
+ */
 export type InfoOptions = {
   /**
    * Whether to return the MD5 hash of the file.
    *
    * @default false
+   * @deprecated Use `await file.digest('md5')` instead.
    */
   md5?: boolean;
 };
--- a/packages/expo-file-system/src/File.types.ts
+++ b/packages/expo-file-system/src/File.types.ts
@@ -227,7 +227,7 @@ export type FileInfo = {
   /**
    * Present if the `md5` option was truthy. Contains the MD5 hash of the file.
    *
-   * @deprecated Use `await file.digest('md5')` instead.
+   * @deprecated Use `await file.digest('MD5')` instead.
    */
   md5?: string;
 };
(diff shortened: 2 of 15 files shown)


wh201906:
> Limit concurrent digest calculations internally. The concern you raised about concurrency is valid, but I think it should remain an implementation detail.

How do you think about this? The concurrency count limit is like `Dispatchers.Default`, but it has less impact on the pure CPU-intensive tasks scheduled with `Dispatchers.Default`.
```kotlin
private val mixedDispatcher = Dispatchers.IO.limitedParallelism(
  Runtime.getRuntime().availableProcessors().coerceAtLeast(2)
)
...
withContext(mixedDispatcher) {
  // do digest
}
```

We can even use it for `File.base64()` because it also mixes CPU calculation and IO

wh201906:
I just added some test cases and they pass. It's ready for review now.

packages/expo-file-system/src/File.types.ts
```diff
@@ -226,15 +226,26 @@ export type FileInfo = {
...
+   *
+   * @deprecated Use `await file.digest('md5')` instead.
    */
   md5?: string;
 };
 
+/**
+ * Algorithm used to calculate a file digest.
+ */
+export type FileDigestAlgorithm = 'md5' | 'sha-1' | 'sha-256' | 'sha-384' | 'sha-512';
```
barthap: `expo-crypto` already ships `CryptoDigestAlgorithm` with `'MD5'`, `'SHA-1'`, `'SHA-256'` and so on, and Web Crypto's `subtle.digest()` uses the same uppercase names. With lowercase here, you can't pass `CryptoDigestAlgorithm.SHA256` straight into `file.digest()`, which is the first thing I'd reach for.

Let's match it:

```ts
export type FileDigestAlgorithm = 'MD5' | 'SHA-1' | 'SHA-256' | 'SHA-384' | 'SHA-512';
```

That means updating the `when` in `FileSystemFile.kt`, the `switch` in `FileSystemFile.swift`, `digestHexLengths` in the mock, and the tests.
No need to introduce enum, string union is enough. It's only about uppercase
Sorry for the churn, but this is public API we can't quietly change later.

packages/expo-file-system/android/src/main/java/expo/modules/filesystem/FileSystemFile.kt
```diff
@@ -174,12 +174,23 @@ class FileSystemFile(uri: Uri) : FileSystemPath(uri) {
...
 
     validatePermission(FilePermissionService.Permission.READ)
-    val md = MessageDigest.getInstance("MD5")
+    val md = when (algorithm) {
+      "md5" -> MessageDigest.getInstance("MD5")
+      "sha-1" -> MessageDigest.getInstance("SHA-1")
+      "sha-256" -> MessageDigest.getInstance("SHA-256")
+      "sha-384" -> MessageDigest.getInstance("SHA-384")
+      "sha-512" -> MessageDigest.getInstance("SHA-512")
+      else -> throw IllegalArgumentException("Unsupported digest algorithm: $algorithm")
```
barthap: This is the plain `java.lang.IllegalArgumentException`, so expo-modules-core wraps it in `UnexpectedException` and JS sees code `ERR_UNEXPECTED`. iOS throws a proper `Exception` named `UnsupportedDigestAlgorithm`, so the same bad input reports differently on each platform.

Could you add a `CodedException` to `FileSystemExceptions.kt` and throw that instead?

```kotlin
internal class UnsupportedDigestAlgorithmException(algorithm: String) :
  CodedException("Unsupported digest algorithm: '$algorithm'.")
```

packages/expo-file-system/android/src/main/java/expo/modules/filesystem/FileSystemFile.kt
```diff
@@ -174,12 +174,23 @@ class FileSystemFile(uri: Uri) : FileSystemPath(uri) {
...
-  @OptIn(ExperimentalStdlibApi::class)
   val md5: String get() {
+    return digest("md5")
+  }
+
+  @OptIn(ExperimentalStdlibApi::class)
+  fun digest(algorithm: String): String {
     val bufferSize = 65536
 
     validatePermission(FilePermissionService.Permission.READ)
```
barthap, line 186: `bytes()` and `asContentUri()` just above both call `validateType()` before `validatePermission()`. `digest()` doesn't, so on a directory it falls through to `file.inputStream()` and throws a raw `FileNotFoundException` saying "Is a directory", which surfaces as an unexpected error.

```kotlin
fun digest(algorithm: String): String {
  validateType()
  validatePermission(FilePermissionService.Permission.READ)
  ...
```

That gives `InvalidTypeFileException`, consistent with the rest of the class. The old `md5` getter had the same gap so it isn't a regression, but `digest()` is the path we're pointing people at now. iOS needs the same thing for a different reason, see my summary comment.

packages/expo-file-system/src/internal/NativeFileSystem.types.ts
```diff
@@ -229,6 +230,14 @@ export declare class NativeFileSystemFile {
    */
   info(options?: InfoOptions): FileInfo;
 
+  /**
+   * Calculates the digest of the file's contents.
+   * @param algorithm The digest algorithm to use.
+   * @returns A promise that resolves to the lowercase hexadecimal digest.
+   * @throws Error if the file does not exist or cannot be read.
+   */
+  digest(algorithm: FileDigestAlgorithm): Promise<string>;
```
barthap, line 239: `ExpoFileSystem.web.ts` stubs `preview()` and `canPreview()` to warn and return a rejected promise. `digest()` isn't there, so on web this throws `TypeError: file.digest is not a function` synchronously, which breaks anyone who wrote `file.digest('MD5').catch(...)`.

```ts
digest(_algorithm: unknown): Promise<string> {
  console.warn('expo-file-system is not supported on web');
  return Promise.reject(new Error('File digest is not supported on web'));
}
```

packages/expo-file-system/src/__tests__/FileSystem-test.native.ts
```diff
@@ -328,6 +329,39 @@ describe('expo-file-system behavioral mock', () => {
...
+    }
+  });
+
+  it('File.digest rejects when the file does not exist', async () => {
+    const file = new File(Paths.cache, 'missing-digest.txt');
+
+    await expect(file.digest('md5')).rejects.toThrow('File does not exist');
+  });
+
+  it('File.digest rejects when the path points to a directory', async () => {
```
barthap, line 355: This passes, but not for the reason the name gives. The mock's `FileSystemFile.exists` returns `false` for `kind: 'dir'` entries, so it hits the same "File does not exist" branch as the test above it, and a bare `rejects.toThrow()` can't tell the two apart.

Once `validateType()` is in place (my other comment), the mock can throw something distinct for a directory and this can assert on it. As written the test would keep passing even if directory handling broke.

barthap (commented):
Nice work; this addresses everything I raised on #46403. 💯 

One thing I'd like settled before this merges: the algorithm casing (comment inline). The rest are smaller: a coded exception on Android, directory validation, a web stub, and a test nit.

One pre-existing thing, but worth knowing when you read the directory comment: on iOS, `withCorrectTypeAndScopedAccess` doesn't validate the type despite its name; it only checks the permission. That applies to `bytes()`, `text()`, `info()` and the rest too, so I'll track it separately.

packages/expo-file-system/src/__tests__/FileSystem-test.native.ts
wh201906: I see. Fixed in 481aa9efcd420cc8e8b9f201961f5f5b5b8f5440 since I have introduced `validateType()` in c382fd81f807170f6714d73a5aafe38bf9381802

packages/expo-file-system/android/src/main/java/expo/modules/filesystem/FileSystemFile.kt
wh201906: Fixed in c382fd81f807170f6714d73a5aafe38bf9381802

packages/expo-file-system/android/src/main/java/expo/modules/filesystem/FileSystemFile.kt
wh201906: Fixed in c382fd81f807170f6714d73a5aafe38bf9381802

packages/expo-file-system/src/File.types.ts
wh201906: I just randomly picked the lower case form. With the context you provided, it's indeed more reasonable to use the same form as other types. Fixed in d1d42be5962b8dee65ad05ee0691dc3b1b31a095

packages/expo-file-system/src/internal/NativeFileSystem.types.ts
barthap: Disregard this one, my mistake. I only looked at `preview()` and `canPreview()` and read them as a precedent. Looking at the whole file, `FileSystemFile` in `ExpoFileSystem.web.ts` has a constructor and those two methods, nothing else. `text()`, `bytes()`, `base64()`, `write()`, `info()`, `md5` and `size` are all missing too, so `digest()` is in line with the rest of the shim.

The web class needs filling in as a whole, which isn't this PR's job. Nothing to change here.

barthap (approved):
LGTM, excellent job!

packages/expo-file-system/src/internal/NativeFileSystem.types.ts
wh201906: Fixed as you suggested in 2f4f2b718bba1fe26b40ac756af33917147ed3c5
Actually it's not hard to support SHA-1/SHA-256/SHA-384/SHA-512 for web because WebCrypto supports them, but MD5 needs extra dependencies. Would you consider adding support of File.digest() for web? (or add WebCrypto supported algorithms only?)

barthap: And you'd already done it, sorry for the noise. The stub is fine as it stands.

No web support for now. The web `File` class needs a lot more than a digest function before that makes sense, and we have separate plans for it. Let's keep this PR to the native side.

wh201906: Oh I didn't notice your last comment, sorry about that.
I've reverted the changes of 2f4f2b718bba1fe26b40ac756af33917147ed3c5 in 741d9f8ed1868113778cc3d9ef67349058762993

packages/expo-file-system/src/__tests__/FileSystem-test.native.ts
```diff
@@ -328,6 +329,41 @@ describe('expo-file-system behavioral mock', () => {
...
+      ['MD5', 32],
+      ['SHA-1', 40],
+      ['SHA-256', 64],
+      ['SHA-384', 96],
+      ['SHA-512', 128],
+    ] as const;
+    for (const [algorithm, hexLength] of algorithms) {
+      await expect(file.digest(algorithm)).resolves.toMatch(
+        new RegExp(`^[0-9a-f]{${hexLength}}$`)
+      );
```
barthap: Formatting nit, this is what's failing `check-packages`. oxfmt wants it on one line, the collapsed form is exactly 100 chars which is our `printWidth`.

```suggestion
      await expect(file.digest(algorithm)).resolves.toMatch(new RegExp(`^[0-9a-f]{${hexLength}}$`));
```

Applying it myself so you don't have to. Worth knowing that `pnpm run format` won't catch this locally, our root `.oxfmtrc.json` skips `__tests__`. Something for us to sort out on our side.

wh201906: Thanks!
I tried `pnpm et check expo-file-system` on 741d9f8ed1868113778cc3d9ef67349058762993 but it reports no errors

PR #48097, opened 2026-07-24
[expo-navigation-bar] Prevent unhandled promise rejections when declarative navigation bar updates race with Android activity teardown

zoontek:
# Why

Follow https://github.com/expo/expo/pull/48084

The declarative `NavigationBar` component updates Android through asynchronous native methods, but its internal `setStyle` and `setHidden` helpers do not return those promises to the caller.

If an update races with activity teardown, the native promise can reject after the component has already unmounted. Because this internal best-effort update has no consumer, the rejection becomes an unhandled promise rejection and can surface as a LogBox error.

# How

Consume rejections from the two declarative native updates. The public async visibility API remains unchanged and still propagates its rejection to callers.

# Test Plan

- `pnpm run lint`

# Checklist

### DIFF
--- a/packages/expo-navigation-bar/CHANGELOG.md
+++ b/packages/expo-navigation-bar/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ### 🐛 Bug fixes
 
+- Prevent unhandled promise rejections when declarative navigation bar updates race with Android activity teardown.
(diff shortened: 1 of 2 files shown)


vonovak (commented):
I wonder
- if the call rejects, then `currentValues.style = resolvedStyle` could in fact be stale - should the catch block do `currentValues.style = undefined`?
- maybe the native side doesn't need to reject?

zoontek:
@vonovak The native side never reject, it mostly to avoid an issue with Jest

vonovak (approved):

PR #48098, opened 2026-07-24
[plugins] Add the dev server port to info.plist

alanjhughes:
# Why

Follow up of #48010

Bare debug builds that don't use `expo-dev-client` always default the Metro port to `8081`. `RCT_METRO_PORT` is a compile-time macro that gets frozen at `8081` in prebuilt React, so the app has no runtime channel to learn the real port. Start two bare projects at once and their command sockets collide. The bundle loads from the correct server, but packager commands (reload, dev menu) from the first instance are acknowledged by both apps.

Android doesn't have this problem because it resolves the port at runtime from the `react_native_dev_server_port` resource.

# How

- **Config plugin.** A new `withMetroPort` plugin in  writes `RCTMetroPort = $(RCT_METRO_PORT)` into `Info.plist`, wired into prebuild through `withDefaultPlugins`. 

- **Runtime read.** `ExpoReactNativeFactory` reads `RCTMetroPort` from the plist at launch and seeds `RCTBundleURLProvider.jsLocation`, so both the bundle and the command socket resolve the right host and port. When `EXDevLauncherController` is present it bails, because dev-client owns server discovery and must not be overridden.

- **Cleanup.** With the socket host now seeded correctly at init through `EXBundleConfiguration`, the manual packager reconnect in `DevMenuPackagerConnectionHandler` is redundant and has been removed. It existed to work
around `RCTDevSettings` opening the socket on localhost before the host was known.

# Test Plan

- Two bare apps prebuilt and run on `8081` and `8082`. Before this change, commands from the first app were acknowledged by both. Now each app responds only to its own packager, and the behavior survives a reload.
- `bare-expo` on a physical device. Reload and dev menu commands work, confirming `EXBundleConfiguration` seeds the socket correctly without the manual reconnect.
- Expo Go is unaffected. It links no dev-launcher, has no `RCTMetroPort` key, and uses its own `setupWebSocketControls`. Its socket correctness comes from #48010.

### DIFF
--- a/apps/bare-expo/ios/Podfile.lock
+++ b/apps/bare-expo/ios/Podfile.lock
@@ -402,6 +402,9 @@ PODS:
     - SDWebImageWebPCoder
   - ExpoImagePicker (56.0.14):
     - ExpoModulesCore
+  - ExpoImagePicker/Tests (56.0.14):
+    - ExpoModulesCore
+    - ExpoModulesTestCore
   - ExpoInsights (56.0.14):
     - EASClient
     - ExpoModulesCore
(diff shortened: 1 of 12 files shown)


alanjhughes:
* **#48098** <a href="https://app.graphite.com/github/pr/expo/expo/48098?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/48098?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

kitten (approved):
- tested on simulator, + rebuild w/ changed port
- tested on physical device + rebuild w/ changed port

PR #48104, opened 2026-07-24
[ios][router] Prevent presses from firing during Link context menu interaction

### LINKED ISSUE #47747: Link Previews: Long press on embedded URL leaves app in broken navigation state

https://github.com/sharplet/expo-link-preview-bug/

### Steps to reproduce

1. Run the app `npx expo run:ios`
2. Locate the URL on the start page ("www.apple.com")
3. Interact with the URL:
   a. Perform a single tap on the URL. The web page should open in a modal. Close the modal.
   b. Perform a long press on the URL. You should see a content preview and a single menu item with the title "Test".
   c. Perform a _half long press_ on the URL. This is like (3b) but releasing the long press before the full animation completes.

The "half long press" gesture causes the URL to open in a full screen modal preview that is impossible to navigate out of.

https://github.com/user-attachments/assets/7afb15de-c8c6-4c82-a640-498c927d4733

### Environment

```text
    System:
      OS: macOS 26.5.1
    Binaries:
      Node: 22.18.0 - /opt/homebrew/opt/node@22/bin/node
      npm: 10.9.3 - /opt/homebrew/opt/node@22/bin/npm
      Watchman: 2025.04.14.00 - /opt/homebrew/bin/watchman
    Managers:
      CocoaPods: 1.16.2 - /opt/homebrew/lib/ruby/gems/3.4.0/bin/pod
    SDKs:
      iOS SDK:
    IDEs:
    npmPackages:
      expo: ~57.0.4 => 57.0.4 
      react: 19.2.3 => 19.2.3 
      react-dom: 19.2.3 => 19.2.3 
    npmGlobalPackages:
      eas-cli: 16.20.1
    Expo Workflow: managed
```

### Expo Doctor Diagnostics

```text
```

jiunshinn:
# Why

Fixes #47747

A "half long press" on a `<Link>` with `Link.Trigger` + `Link.Menu` (no `Link.Preview`) — releasing the press after the context menu interaction starts but before the menu fully presents — leaves the app in a broken navigation state. The release is still recognized as a tap by react-native, so press handlers inside the trigger fire (e.g. `openBrowserAsync` in the issue's repro) and present a modal while the `UIContextMenuInteraction` is tearing down. The resulting overlay cannot be dismissed and the app has to be restarted.

# How

Two parts:

1. **Native:** `NativeLinkPreviewView` now cancels in-flight react-native touches when `contextMenuInteraction(_:configurationForMenuAtLocation:)` fires, by toggling `isEnabled` on the react-native touch-handler recognizers found on the superview chain — the same technique react-native-screens uses to cancel touches when a native transition takes over. This covers any press handler inside the trigger, including app-owned handlers nested under `asChild` triggers. The issue's repro uses such a handler (`ThemedText onPress` → `openBrowserAsync`), which never reaches expo-router's own press pipeline, so a JS-only guard cannot fix it (verified — see Test Plan).

2. **JS:** `onWillPreviewOpen` now arms `blockPressRef` unconditionally instead of only when a preview is attached. The suppression check in `BaseExpoRouterLink` (added in #43071) and the unconditional release in `onPreviewDidClose` already treat the flag as interaction-scoped; only the arming was preview-gated. This covers the variant where the navigation goes through the `Link`'s own press handler, and keeps the guard consistent for menu-only links.

# Test Plan

- New regression test in `src/link/__tests__/Link.test.ios.tsx`: a menu-only link does not navigate when the trigger press lands between `onWillPreviewOpen` and `onPreviewDidClose`, and navigates normally afterwards (the guard does not stay latched). Written red → green against the JS change.
- Verified against the issue's repro app (https://github.com/sharplet/expo-link-preview-bug) on an iPhone 17 Pro simulator (iOS 26.5), injecting the half long press with `idb ui tap --duration 0.7`:
  - Unpatched `expo-router@57.0.4`: reproduces the stuck full-screen overlay from the issue, unescapable by tapping or swiping.
  - JS change only (patched into the served bundle): still reproduces — the repro's press handler is app-owned and never reaches the `blockPressRef` guard.
  - With the native change: the same gesture shows the context menu correctly. Single tap still opens a dismissible in-app browser modal, a full long press opens the menu normally, and taps work again immediately after the menu closes.
- `et check-packages expo-router` passes (build, typecheck, lint, tests).

# Checklist


packages/expo-router/ios/LinkPreview/LinkPreviewNativeView.swift
```diff
@@ -185,6 +186,20 @@ class NativeLinkPreviewView: RouterViewWithLogger, UIContextMenuInteractionDeleg
     }
   }
 
+  // A press released before the menu fully presents is still recognized as a tap
+  // by react-native, so cancel in-flight touches when the interaction starts.
+  private func cancelReactNativeTouches() {
+    var view: UIView? = self
+    while let current = view {
+      for recognizer in current.gestureRecognizers ?? []
+      where String(describing: type(of: recognizer)).contains("TouchHandler") && recognizer.isEnabled {
```
Ubax: [🏗️ **design**] This deviates from the react-native-screens technique it's modeled on in three ways that widen the blast radius:

1. **Substring match** — `contains("TouchHandler")` will cancel *any* recognizer whose class name happens to contain "TouchHandler", including app- or library-owned ones on ancestor views. RNS matches `isKindOfClass:RCTSurfaceTouchHandler` exactly (`UIView+RNSUtility.mm`). Since this view is Fabric-only (`mountChildComponentView`), an exact name check against `RCTSurfaceTouchHandler` (e.g. via `NSClassFromString`, avoiding the header import problem from Swift) is sufficient and safer.
2. **Whole-chain walk** — this toggles every matching handler all the way up to the window. On Fabric, each screen not mounted under the react root gets its own surface touch handler (see the comment in RNS's `rnscreens_findTouchHandlerInAncestorChain`), so in modal contexts this cancels touches on surfaces that aren't involved in this interaction. RNS stops at the first `RCTSurfaceView`/modal boundary and cancels only that one handler — returning after the first match would replicate that.
3. **Missing `reset`** — RNS's `rnscreens_cancelTouches` is `setEnabled:NO; setEnabled:YES; reset` . The `reset` was added deliberately; if you're citing their technique, either match it or note why it's not needed here.

packages/expo-router/src/link/LinkWithPreview.tsx
```diff
@@ -108,8 +108,8 @@ export function LinkWithPreview({ children, ...rest }: LinkWithPreviewProps) {
       nextScreenId={isPad ? undefined : nextScreenId}
       tabPath={isPad ? undefined : tabPathValue}
       onWillPreviewOpen={() => {
+        blockPressRef.current = true;
```
Ubax: Since we are doing this natively we should remove the JS logic

packages/expo-router/ios/LinkPreview/LinkPreviewNativeView.swift
jiunshinn: Reworked in 75aadfdd: exact `RCTSurfaceTouchHandler` match via `NSClassFromString`, and the walk now stops at the first handler found.

`reset` turned out to crash here: unlike the RNS call sites, the press is still active when `configurationForMenuAtLocation` fires, so UIKit delivers `touchesCancelled` to the handler after the toggle returns — resetting first desyncs its touch registry and trips the NSAssert in `-[RCTSurfaceTouchHandler _updateTouches:]` (reproducible on the issue's repro app; RN's own cancellation path handles the cleanup). Noted in a code comment instead, per your "or note why".

Re-verified the final shape against the repro on simulator: half long press shows the menu, single tap still opens/dismisses the browser modal, and a repeated half press after that still works.

packages/expo-router/src/link/LinkWithPreview.tsx
jiunshinn: Removed in 75aadfdd, along with the jest test that covered the JS guard — the suppression isn't observable from JS anymore.

intergalacticspacehighway (approved):
Approach looks good. We've taken a similar approach for disabling similar touches for Expo UI SwiftUI Menu - https://github.com/expo/expo/pull/48463. Will wait for final review from @Ubax

Ubax (commented):
@jiunshinn the change looks great! I have one more small ask. Can you remove `blockPressRef` logic from `expo-router`, as it is no longer needed after this change

jiunshinn:
Removed in 84a4abd — `blockPressRef`, its `InternalLinkPreviewContext` field, and the `BaseExpoRouterLink` press guard are gone (the `use(InternalLinkPreviewContext)` call there was only for the guard, so it's removed too). `et check-packages expo-router` passes.

Ubax (approved):
Thank you for your contribution 🎉

PR #48115, opened 2026-07-24
[CI] Attempt to fix Fingerprint job

alanjhughes:
# Why

managed-test scaffolds blank@sdk-56 at test time, so floating patch versions drift the snapshots and redden the Fingerprint job on every publish. normalizeAutolinkingVersionsForSnapshot already strips versions from autolinking contents sources but missed type: 'package' sources.

# How

Normalize version/hash to * for type: 'package' sources too. Snapshot-only, fingerprint hashing unchanged. Regenerated the external snapshot and the diff test's inline snapshot.

# Test Plan
CI


Kudo (approved):
thanks for fixing the problem!

PR #48125, opened 2026-07-25
[android][ui] Fix `DateTimePicker` dialog squeezing the calendar when switching from input mode

### LINKED ISSUE #47273: [DateTimePicker] Unexpected resize when opening calendar from spinner mode on Android.

https://github.com/TomCorvus/RNExpoUIDateTimePicker

### Steps to reproduce

1. Initialize a new Expo project.
2. Install and configure `@expo/ui`.
3. Add a `DateTimePicker` using the spinner display mode.
4. Open the picker.
5. When clicking the icon in the top-right corner, open the calendar view directly from the picker.

https://github.com/user-attachments/assets/d4fde098-888e-4079-824b-b976ba29db18

### Environment

```text
    System:
      OS: macOS 26.5.1
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 25.7.0 - /Users/thomascorbin/.nvm/versions/node/v25.7.0/bin/node
      npm: 11.16.0 - /Users/thomascorbin/.nvm/versions/node/v25.7.0/bin/npm
    Managers:
    SDKs:
      iOS SDK:
      Android SDK:
        API Levels: 33, 34, 35, 36
        System Images: android-28 | Google ARM64-V8a Play ARM 64 v8a, android-36 | Google Play ARM 64 v8a, android-36 | 16 KB Page Size Google Play ARM 64 v8a
    IDEs:
    npmPackages:
      expo: ~56.0.12 => 56.0.12 
      react: 19.2.3 => 19.2.3 
    Expo Workflow: managed
```

### Expo Doctor Diagnostics

```text
```

dileepapeiris:
# Why

Closes #47273

On Android, opening the `community/datetime-picker` (or the Jetpack Compose `DatePickerDialog`) in the spinner (text input) display mode and then tapping the calendar icon rendered the calendar squeezed, with the weekday header overlapping the day numbers, until the soft keyboard hid on its own (~1.5 s later).

The root cause: the soft keyboard from the input text field is still visible when the display mode toggles, and the window manager keeps the dialog window clamped to the space above it — so the calendar composes into a window that is far too short.

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td align="center">
      <video src="https://github.com/user-attachments/assets/0695c773-4541-41fe-af56-dd48deba07a0" controls width="300"></video>
    </td>
    <td align="center">
      <video src="https://github.com/user-attachments/assets/cf1f77dd-05bd-4675-8f3c-991407c90f75" controls width="300"></video>
    </td>
  </tr>
</table>

# How

All changes are scoped to `ExpoDatePickerDialogContent` (the dialog-presented date picker); the inline picker, the time picker dialog, and iOS are untouched.

- **Dismiss the leftover keyboard** as soon as the calendar is shown, instead of waiting for it to hide on its own.
- **Stop the dialog window from avoiding the keyboard in calendar mode** (`SOFT_INPUT_ADJUST_NOTHING`) so the window can expand to full size immediately. The window's original soft-input mode is captured and restored when toggling back to input mode, so the text field keeps its stock keyboard-avoidance behavior — input mode is byte-for-byte unchanged.
- **Lay the calendar out at its natural height while the window is still expanding** (`Modifier.wrapContentHeight(align = Top, unbounded = true)`, calendar mode only). For the few frames before the window reaches full size, the grid keeps its natural row spacing and only its bottom edge is briefly clipped, instead of compressing into overlapping rows. This is a no-op whenever the window already provides full height (every normal portrait situation).

I iterated on this against a device before settling on this approach: purely reactive window-flag switching still shows a squeezed frame (the flag lands after the calendar has already composed into the clamped window), and positioning the dialog via `WindowInsets.ime` padding turned out to be unreliable in a dialog window (insets are not dependably dispatched to secondary windows, especially with an RN `Modal` in the window stack). The final approach relies only on synchronous window-manager mechanics, so there are no timing races.

One behavior note for reviewers: on height-constrained screens (landscape), stock Material3 rendering compresses the calendar grid into overlapping rows permanently; with this change it renders at natural spacing with the bottom rows clipped instead. Neither shows the full month — Material's guidance for short screens is input mode — but the failure mode is cleaner.

## Verification app

A standalone verification app containing a **prebuilt patched `@expo/ui` SDK** is available here:

**https://github.com/dileepapeiris/expo-ui-android-datepicker-resize-fix**

The repository includes the patched `@expo/ui` package as a committed `.tgz` file, so reviewers can clone it, install dependencies, and verify the fix without building the Expo monorepo.

# Checklist


intergalacticspacehighway (approved):
Thanks!

Issue #48126, opened 2026-07-25
[cli] expo run:ios --device omits wirelessly connected iOS <= 16 devices (osType undefined over network lockdown)


davellanedam:
### Summary

`npx expo run:ios --device` silently omits physical devices running **iOS ≤ 16** when they are connected **over Wi-Fi** (Xcode "Connect via network"), both from the interactive picker and from `--device <udid>` lookup ("No device UDID or name matching ..."). The same device appears fine when connected via USB. iOS 17+ devices are unaffected (they go through `devicectl`).

### Environment

- `@expo/cli` 56.1.21 (SDK 56) — bug also present on `main`
- macOS (Darwin 25.5.0), Xcode 26
- Device: iPhone X (`iPhone10,6`), iOS 16.7.16, paired wirelessly via Xcode "Connect via network"

### Root cause

For pre-iOS 17 devices, `getConnectedDevicesUsingCustomToolingAsync()` (in `packages/@expo/cli/src/run/ios/appleDevice/AppleDevice.ts`) queries lockdownd via usbmuxd and computes:

```ts
osType: coerceUsbmuxdPlatformToOsType(deviceValues.DeviceClass),
```

Over a **network** usbmuxd connection, iOS 16's lockdownd `GetValue` returns a reduced property set that **does not include `DeviceClass`** (observed keys: `CPUArchitecture`, `DeviceName`, `HardwareModel`, `HumanReadableProductVersionString`, `ProductName`, `ProductType`, `ProductVersion`, `SupportedDeviceFamilies`). `coerceUsbmuxdPlatformToOsType(undefined)` then returns `undefined`, and `getDevicesAsync({ osType: 'iOS' })` → `filterDevicesForOsType()` drops the device because `undefined` is not in `['iOS', 'macOS', 'xrOS']`.

Debug output showing the failure (device is found, then filtered out):

```
expo:apple-device:protocol:lockdown Response: {"Request":"GetValue","Value":{"CPUArchitecture":"arm64","DeviceName":"iPhone","HardwareModel":"D221AP","HumanReadableProductVersionString":"16.7.16","ProductName":"iPhone OS","ProductType":"iPhone10,6","ProductVersion":"16.7.16","SupportedDeviceFamilies":[1]}}
expo:apple-device Unknown usbmuxd platform (needs to be added to Expo CLI): undefined
```

[...]

PR #48127, opened 2026-07-25
[cli] Fix wireless iOS <= 16 devices missing from run:ios device list

davellanedam:
# Why

`npx expo run:ios --device` silently omits physical devices running iOS ≤ 16 that are connected over Wi-Fi (Xcode "Connect via network") — both from the interactive picker and from `--device <udid>` lookup, which fails with `No device UDID or name matching ...`. The same device works fine over USB, and iOS 17+ devices are unaffected (they resolve via `devicectl`).

Context and full diagnosis in #48126 (auto-closed for missing repro project — the bug depends on a physically paired wireless device, so a repro repo can't capture it; happy to provide anything else needed).

# How

For pre-iOS 17 devices, `getConnectedDevicesUsingCustomToolingAsync()` computes `osType` from lockdownd's `DeviceClass`. Over a **network** usbmuxd connection, iOS ≤ 16's lockdownd `GetValue` returns a reduced property set that omits `DeviceClass` (observed keys: `CPUArchitecture`, `DeviceName`, `HardwareModel`, `HumanReadableProductVersionString`, `ProductName`, `ProductType`, `ProductVersion`, `SupportedDeviceFamilies`). `osType` therefore resolved to `undefined` and `filterDevicesForOsType()` dropped the device.

The reduced set does include `ProductName: "iPhone OS"`, which `coerceUsbmuxdPlatformToOsType` already maps to `'iOS'` — so this falls back to it:

```diff
-        osType: coerceUsbmuxdPlatformToOsType(deviceValues.DeviceClass),
+        osType: coerceUsbmuxdPlatformToOsType(deviceValues.DeviceClass ?? deviceValues.ProductName),
```

USB behavior is unchanged: `DeviceClass` is always present there, so the fallback never triggers.

# Test Plan

- Added `AppleDevice-test.ts` with a regression test: a usbmuxd device with `ConnectionType: 'Network'` whose lockdownd values omit `DeviceClass` (the reduced network property set) must resolve `osType: 'iOS'`, plus a USB case covering the unchanged `DeviceClass` path. Verified red/green: the network case fails without the fix and passes with it.

- Paired an iPhone X (iOS 16.7.16) with Xcode, enabled "Connect via network", unplugged it (usbmuxd `ListDevices` reports `ConnectionType: "Network"`).
- Before: device absent from `expo run:ios --device` picker; `--device <udid>` → `No device UDID or name matching ...`; debug log shows `Unknown usbmuxd platform (needs to be added to Expo CLI): undefined`.
- After (patched build): `resolveDeviceAsync` returns the device with `osType: "iOS"`, it appears in the picker with the 🌐 connection indicator, and `--device` resolves it by name and by UDID.
- USB regression check: plugging the same device in still lists it via the `DeviceClass` path (fallback not exercised).

# Checklist

- [x] Documentation is up to date to reflect these changes (no docs changes needed — bug fix only).


krystofwoldrich:
@davellanedam Thank you for opening the PR. Could you share the object values (anonymized) `const deviceValues = await new LockdowndClient(socket).getAllValues();` you get for the wirelessly connected iOS 16 device?

davellanedam:
@krystofwoldrich Sure! This is the full lockdownd response for the wirelessly connected iOS 16 device, taken from the `expo:apple-device:protocol:lockdown` debug trace of that exact `getAllValues()` call (`GetValue` with no key):

```json
{
  "Request": "GetValue",
  "Value": {
    "CPUArchitecture": "arm64",
    "DeviceName": "iPhone",
    "HardwareModel": "D221AP",
    "HumanReadableProductVersionString": "16.7.16",
    "ProductName": "iPhone OS",
    "ProductType": "iPhone10,6",
    "ProductVersion": "16.7.16",
    "SupportedDeviceFamilies": [1]
  }
}
```

That's the complete `Value` object — nothing was redacted. Over a network usbmuxd connection, iOS ≤ 16's lockdownd returns only this reduced set, so `DeviceClass` (along with `UniqueDeviceID`, `SerialNumber`, etc.) is simply absent. For comparison, the same device over USB does include `DeviceClass: "iPhone"`, so the `??` fallback never triggers on USB.

The subsequent line in the trace shows the failure this PR fixes:

```
expo:apple-device Unknown usbmuxd platform (needs to be added to Expo CLI): undefined
```

Happy to provide any other traces if useful.

krystofwoldrich (approved):

PR #48129, opened 2026-07-26
[android][font] apply fontWeight with useFonts() and support variable fonts via config plugin

L65FREAD:
## Why

Fixes the Android half of #33108. Two separate bugs:

**1. `fontWeight` and `fontStyle` were ignored for anything loaded with `useFonts`.** `FontLoaderModule` registered the typeface with `ReactFontManager.setTypeface(name, Typeface.NORMAL, typeface)`, which populates the style-keyed `fontCache`. A request for bold then missed that cache and fell through to `createAssetTypeface(name, Typeface.BOLD, assets)`, which looks for `fonts/<name>_bold.ttf` in the assets, doesn't find it, and returns `Typeface.create(name, BOLD)` — a **system** font. Bold and italic text rendered in Roboto, not the loaded font.

**2. Variable fonts always rendered at their default weight.** Android hands back only a variable font's default instance and does not expand its named instances, so even once `fontWeight` reached the typeface there was only one weight in it.

## How

**Runtime (`useFonts`)**

- Register with `ReactFontManager.addCustomFont(family, typeface)` instead of `setTypeface`. `getTypeface` consults `customTypefaceCache` first and applies `TypefaceStyle.apply`, which is already API-guarded (`< P` → `Typeface.create(tf, nearestStyle)`, `>= P` → `Typeface.create(tf, weight, italic)`), so this is safe down to `minSdk` 24 and fixes bold/italic on every API level.
- On API 29+, assemble a `FontFamily` containing one `Font` per weight, each built from the same buffer with `setFontVariationSettings("'wght' N").setWeight(N)`. The variation setting picks the glyphs; the weight is what minikin's closest-match lookup reads. `Typeface.CustomFallbackBuilder` with `setSystemFallback("sans-serif")` matches what `Typeface.createFromAsset` sets up, so uncovered glyphs still come from the system font.
- On API 35+, `FontFamily.Builder.buildVariableFamily()` does this natively — every weight rather than nine steps, plus the `ital` axis when the font has one. Used when available, falling back to the manual family.
- Static fonts and API < 29 fall through to the existing `createFromAsset` / `createFromFile` path untouched.

**Detecting a variable font** — `FontVariationAxes.kt` (new)

Android exposes no public API for the axes a font *supports* (`Font.getAxes()` returns what it was *built* with; `FontFileUtil` is hidden), so this reads the `wght` range out of the `fvar` table directly. It's free of Android APIs so it unit-tests on the JVM.

Detection is required, not incidental: building nine instances of a *static* font would give minikin exact matches at every weight and suppress its synthetic bold, making static fonts render **worse** than before.

Fonts are only copied into memory when their table directory shows an `fvar` table, so a static asset font costs one 4 KB header read rather than a full copy that gets discarded. File-backed fonts are mapped, which costs nothing either way.

**Config plugin**

- Emit `app:fontVariationSettings="'wght' N"` alongside `app:fontWeight`, so one variable font file can back several `<font>` entries. `FontResourcesParserCompat.readFont` reads the attribute from the `app` namespace, and both `TypefaceCompatApi26Impl` and `TypefaceCompatApi29Impl` pass it through. Emitted unconditionally — a font with no `fvar` table ignores it, and guessing which entries are variable at config time without reading the font files is worse.
- Deduplicate the copy list. Listing one file under several weights previously scheduled several concurrent `fs.copyFile` calls onto the same destination path inside one `Promise.all`.

## Behaviour changes worth flagging

Moving from `fontCache` to `customTypefaceCache` means `getTypeface` short-circuits on the custom entry and never falls through:

- An app that ships `assets/fonts/Inter_bold.ttf` (the React Native naming convention) *and* calls `loadAsync('Inter', ...)` previously got the real bold asset for `fontWeight: 'bold'`. It now gets the loaded font instanced (API 29+) or synthetically emboldened (below that). Arguably more correct — the explicit runtime load wins — but it is a change.
- If a family is registered by the config plugin *and* loaded again at runtime under the same name, the runtime registration now overwrites the plugin's. Below API 29 the runtime typeface is single-weight, so that combination is a downgrade on API 24–28. Previously the two lived in separate caches and the plugin's won.

## Not covered

iOS. #33108 reports both platforms; this fixes Android only. The iOS path registers with `CTFontManagerRegisterFontsForURL` and aliases the family to the font's PostScript name, and I have not verified on a device whether that exposes a variable font's named instances to `UIFont.fontNames(forFamilyName:)` — which is what `RCTFont` iterates to resolve `fontWeight`. Happy to split that into a follow-up issue, or extend this PR if you'd prefer it land together.

The docs change therefore describes iOS only in terms of the config-plugin path (`UIAppFonts`) and says what to do if runtime weights don't render, rather than asserting a mechanism.

## Test Plan

**`et check-packages expo-font`** — passes.

**`et native-unit-tests -p android --packages expo-font`** — 24 JVM tests over the `fvar` reader and weight selection: static fonts, CFF (`OTTO`) fonts, `wght` behind other axes, axis records larger and smaller than the spec, fractional bounds, out-of-range bounds, truncated and non-font input, buffer position/byte-order preservation, and the header-only probe.

**`et native-unit-tests -p android --type instrumented --packages expo-font`** — 3 instrumented tests on an API 34 emulator. They render text and count inked pixels, so they measure what reaches the screen:

- `rendersHeavierWeightsWithMoreInk` — ink increases across the weights of a variable font.
- `instancesTheAxisRatherThanSynthesizingBold` — the heaviest weight renders as the font instanced at that `wght`, and *not* as the synthetically emboldened default instance.
- `leavesStaticFontsAlone` — a font with no `wght` axis is left to the existing path.

Fonts come from `/system/fonts` (first variable one covering the Latin sample) so no fixture is committed. On API 34 that selects `DancingScript-Regular.ttf`, axis 400..700 — which also exercises clamping the nine `fontWeight` steps into a narrower axis.

Ink per weight, measured on device:

| | 400 | 500 | 600 | 700 |
|---|---|---|---|---|
| this branch | 4482 | 4929 | 5370 | 5819 |
| axis not applied (`main`'s behaviour) | 4482 | 4482 | 4482 | 4482 |

The second row comes from deliberately removing `setFontVariationSettings`; every weight collapses onto the default instance, which is the bug being fixed. Both instrumented assertions fail in that state, so they don't pass vacuously.

**Config plugin, verified manually.** Ran the Android mods over a scratch project listing one variable font under three weights. Generated `res/font/xml_roboto_serif.xml`:

```xml
<font-family xmlns:app="http://schemas.android.com/apk/res-auto">
  <font app:font="@font/roboto_serif" app:fontStyle="normal" app:fontWeight="400" app:fontVariationSettings="'wght' 400"/>
  <font app:font="@font/roboto_serif" app:fontStyle="normal" app:fontWeight="700" app:fontVariationSettings="'wght' 700"/>
  <font app:font="@font/roboto_serif" app:fontStyle="italic" app:fontWeight="900" app:fontVariationSettings="'wght' 900"/>
</font-family>
```

The 3.6 MB font is copied once, not once per weight. Compiling that exact XML into an APK and loading it with `ResourcesCompat.getFont` on the emulator gave ink 400=4482, 700=5819 — identical to the runtime path. Dropping only the `app:fontVariationSettings` attributes gives 4482/4482, so AAPT links the attribute and androidx honours it.

**Not covered by the runs above:** API 24, 26 and 35 (only an API 34 emulator was available, so the API 26 plugin floor, the API 24 no-crash path and the API 35 `buildVariableFamily` branch are reasoned about but not executed), the full `loadAsync` → `ReactFontManager` → React Native `<Text>` path, and iOS.

## Checklist

- [x] `CHANGELOG.md` entries added
- [x] `et check-packages expo-font`
- [x] Native unit tests and instrumented tests pass
- [x] Config plugin output verified against a scratch project
- [x] Follows the documentation style guide


vonovak:
I made a few changes, just listing them out:

fix: added a prebuild-time error for duplicate weight/style entries in fontDefinitions
ci: `packages/expo-font/android/**` was missing from android-instrumentation-tests.yml's path allowlist

- limit() instead of capacity() in `declaresVariations`

The guard checked capacity(), but the reads inside stop at limit(). So a buffer with fewer bytes than it was allocated could pass the guard and then throw, outside the try. 

- `LocalFile` now checks the header first

On Android 15 and up, every static font was fully parsed, thrown away, then parsed again by createFromFile. 

- `readWholeFont` no longer copies through the heap

It used readBytes(), which grows an array by doubling, then copied that array into a direct buffer. Peak memory use could be up to about 3× the font size.

vonovak (approved):

vonovak:
Thank you!

PR #48130, opened 2026-07-26
[ios][router]: sync native tabs group route title with the focused tab

0zd0:
# Why

On iOS the native stack takes the back button title (and its long-press menu) from the previous screen's `title`. For a tabs layout that screen is the group route, so it reads `(tabs)` instead of the tab you're returning to.

# How

Adds `titleFromFocusedTab` on `<NativeTabs>`. When enabled, writes the focused tab's title into the group route's `title` via `setOptions`, in a layout effect so it's set before a push. Falls back to the route name when the tab has no label. Off by default.

Also affects the group route's own header title if one is visible.

# Test Plan

`<NativeTabs titleFromFocusedTab>` in `app/(tabs)/_layout.tsx` under a root `<Stack>`, one tab labeled Settings. Select it, push a screen outside the group, long-press back: reads Settings instead of `(tabs)`.

# Checklist

### DIFF
--- a/packages/expo-router/src/native-tabs/NativeBottomTabsNavigator.tsx
+++ b/packages/expo-router/src/native-tabs/NativeBottomTabsNavigator.tsx
@@ -1,13 +1,15 @@
 'use client';
 
-import React, { use, useCallback, useEffect, useMemo, useRef } from 'react';
+import React, { use, useCallback, useEffect, useLayoutEffect, useMemo, useRef } from 'react';
+import { Platform } from 'react-native';
 
 import { useRouteNode } from '../Route';
 import { router } from '../imperative-api';
-import type {
-  ParamListBase,
-  TabNavigationState,
-  TabRouterOptions,
+import {
+  type ParamListBase,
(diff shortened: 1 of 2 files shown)


0zd0:
https://github.com/user-attachments/assets/df4f2a5f-3ecb-49f4-bd66-1799fd08bbed

I am attaching a video demonstrating the scenario:

1. Default behavior without the `titleFromFocusedTab` parameter
2. Enabling the `titleFromFocusedTab` parameter
3. Removed the `Stack.Header` hiding logic; the tab header now appears, showing how the global title changes based on the active tab

The key point: The screen being navigated to lies outside the tab layout.

Ubax (changes requested):
Hi @0zd0
Can you please open an issue first? This way it will be easier to understand what the purpose of this PR is

Ubax:
Hi @0zd0,

As described in the issue, we won’t be proceeding with the PR in its current form. Thank you for taking the time to contribute!

Once the issue is addressed internally, I’ll let you know.

PR #48144, opened 2026-07-26
fix-doc: headerSearchBarOptions react navigation option data

AAAboutajedyne:
# Why
the "headerSearchBarOptions" react navigation option is available for both platforms (iOS, and Android).


amandeepmittal:
Thanks for the report. This file is generated by `pnpm react-navigation-docs-sync`, so hand edits here are overwritten on the next sync. I ran the sync in [PR](https://github.com/expo/expo/pull/49008). Upstream has reworded this option, so the description no longer describes it as iOS only. Closing this one, since the fix belongs in the generator rather than the generated file.

PR #48145, opened 2026-07-26
fix-doc: rename headerBackImageSource into its correct name headerBackIcon

AAAboutajedyne:
# Why

the correct naming of the react navigation option "header back icon" is `headerBackIcon` and not `headerBackImageSource`.

### DIFF
--- a/docs/public/data/react-navigation-options.json
+++ b/docs/public/data/react-navigation-options.json
@@ -62,7 +62,7 @@
       "origin": "native-stack-navigator"
     },
     {
-      "name": "headerBackImageSource",
+      "name": "headerBackIcon",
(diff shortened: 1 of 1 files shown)


amandeepmittal:
Thanks for spotting the rename. docs/public/data/react-navigation-options.json is generated by pnpm react-navigation-docs-sync from the react-navigation docs, so edits made here are overwritten on the next sync. Upstream has renamed the option, and a fresh sync now emits headerBackIcon on its own. I ran the sync in [PR](https://github.com/expo/expo/pull/49008), which also picks up four new options and updated descriptions. Closing in favor of that, and thanks again for the report.

Issue #47702, opened 2026-07-11
[expo-notifications] Data race on NotificationCenterManager delegates array → intermittent heap-corruption crash (_xzm_free_abort)


em0xi0nx:
### Summary

On iOS, `NotificationCenterManager` stores its `delegates` and `pendingResponses` as plain, unsynchronized Swift `Array`s. `addDelegate`/`removeDelegate` mutate them (during module registration on launch, and whenever a JS subscriber is added/removed) while the `UNUserNotificationCenterDelegate` callbacks iterate the same arrays on arbitrary system threads. Concurrent mutation + iteration of a Swift value-type Array is undefined behavior and corrupts the heap.

### Observed crash

Intermittent on cold launch of an app that uses expo-notifications for local (scheduled) reminders:

```
EXC_BREAKPOINT (SIGTRAP)
libsystem_malloc.dylib  _xzm_free_abort  (Abort Cause ...)
ExpoNotifications      NotificationCenterManager.addDelegate(_:)
Swift                  Array.append(_:)
```

### Environment

- expo-notifications@57.0.3 (Expo SDK 57), iOS
- File: `ios/ExpoNotifications/Notifications/NotificationCenterManager.swift`

### Root cause

```swift
var delegates: [NotificationDelegate] = []           // no synchronization
var pendingResponses: [UNNotificationResponse] = []  // no synchronization
```

`addDelegate`/`removeDelegate` run on the thread that registers the module / adds a JS listener, while `userNotificationCenter(_:didReceive:withCompletionHandler:)` and the will-present callback iterate `delegates` on UNUserNotificationCenter's callback threads. No lock guards these accesses.

### Suggested fix

Guard all state access with an `NSLock`, and have the public getters return a value-type **snapshot** so iteration never holds the lock across a delegate callback (which could re-enter add/removeDelegate and deadlock). Patch generated by patch-package against 57.0.3:

```diff
@@ -42,8 +42,22 @@ public class NotificationCenterManager: NSObject,
   @objc
   public static let shared = NotificationCenterManager()
 
-  var delegates: [NotificationDelegate] = []
-  var pendingResponses: [UNNotificationResponse] = []
+  // Thread-safety patch (outloud): `delegates`/`pendingResponses` were plain
+  // arrays mutated by add/removeDelegate while UNUserNotificationCenter
+  // callbacks (on arbitrary threads) iterated them — a data race that
+  // corrupted the heap (_xzm_free_abort on Array.append). All access now goes
[...]

PR #48147, opened 2026-07-26
[notifications][iOS] Lock NotificationCenterManager delegate state

zcharef:
# Why

`NotificationCenterManager` keeps `delegates` and `pendingResponses` as plain Swift arrays with no synchronization. They are written from whichever thread registers a module or adds a JS listener, and read from the `UNUserNotificationCenter` callbacks on another thread. Swift arrays are value types over copy-on-write storage, so two threads touching one at the same time can release an element out of a buffer that is being reallocated underneath them.

Three modules add and remove themselves on module lifecycle, all on the JS thread:

- `EmitterModule` (`OnCreate` / `OnDestroy`)
- `HandlerModule` (`OnCreate` / `OnDestroy`)
- `PushTokenModule` (`OnCreate` / `OnDestroy`)

`OnCreate` fires inside `ModuleHolder.init` → `ModuleRegistry.register(fromProvider:)` → `AppContext.registerNativeModules()`, which runs on the React instance's JS thread. One RN instance plus a notification arriving on the main thread during startup is already enough to race. Anything that keeps several RN instances alive at once makes it much easier to hit.

A crash from a debug simulator build on SDK 55 (`expo-notifications@55.0.23`), where seven `com.facebook.react.runtime.JavaScript` threads were inside `AppContext.registerNativeModules()` at the same time and an eighth was running `-[RCTInstance invalidate]`:

```
Exception Type:    EXC_BAD_ACCESS (SIGSEGV)
Exception Subtype: KERN_INVALID_ADDRESS at 0x0000beaddcc8d610

Thread 41 Crashed:: com.facebook.react.runtime.JavaScript
0   libswiftCore.dylib   _swift_release_dealloc + 16
1   libswiftCore.dylib   RefCounts<...>::doDecrementSlow<...> + 156
2   libswiftCore.dylib   _ArrayBuffer._copyContents(subRange:initializing:) + 160
3   libswiftCore.dylib   _ArrayBuffer._consumeAndCreateNew(...) + 468
4   libswiftCore.dylib   Array._reserveCapacityAssumingUniqueBuffer(oldCount:) + 108
5   libswiftCore.dylib   Array.append(_:) + 76
6   ExpoNotifications    NotificationCenterManager.addDelegate(_:) + 192   (NotificationCenterManager.swift:65)
7   ExpoNotifications    closure #1 in EmitterModule.definition() + 104    (EmitterModule.swift:19)
...
13  ExpoNotifications    ModuleHolder.post(event:) + 132
14  ExpoNotifications    ModuleHolder.init(appContext:module:name:) + 384
16  ExpoNotifications    ModuleRegistry.register(module:name:...) + 260
22  ExpoNotifications    ModuleRegistry.register(fromProvider:) + 188
26  ExpoNotifications    AppContext.registerNativeModules() + 84
```

The faulting address sits in no mapped region. It was read out of a buffer another thread had already reallocated.

#47702 hit the same race on 57.0.3 with a different symptom, `_xzm_free_abort` inside `Array.append`. It was closed automatically for a missing reproduction. A race between a module registration and a system callback has no small deterministic repro, so I am sending the patch instead.

# How

Both arrays move behind an `NSLock`. The two properties become computed getters that copy under the lock and hand back the copy, so callers iterate a snapshot instead of holding the lock while calling into a delegate. That part matters: a delegate can call `addDelegate` or `removeDelegate` from inside `didReceive`, and `NSLock` is not recursive, so holding it across a callback would deadlock.

Behaviour is otherwise unchanged. `delegates` and `pendingResponses` stay internal, keep their types, and every existing `for delegate in delegates` site works as before.

# Test Plan

- `swiftc -parse` on the modified file passes.
- Checked every reader and writer of `delegates` and `pendingResponses` under `packages/expo-notifications/ios`. They all live in this file, so no caller depends on the properties being stored or settable.
- I have not built this against the monorepo iOS setup, so CI is the first real compile. Happy to adjust if you would rather see a `DispatchQueue` or an actor here than an `NSLock`.

### DIFF
--- a/packages/expo-notifications/CHANGELOG.md
+++ b/packages/expo-notifications/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### 🐛 Bug fixes
 
+- [iOS] Fix a data race on `NotificationCenterManager` delegate state that could corrupt the heap. ([#48147](https://github.com/expo/expo/pull/48147) by [@zcharef](https://github.com/zcharef))
+
 ### 💡 Others
 
 ## 56.0.14 — 2026-05-26
--- a/packages/expo-notifications/ios/ExpoNotifications/Notifications/NotificationCenterManager.swift
+++ b/packages/expo-notifications/ios/ExpoNotifications/Notifications/NotificationCenterManager.swift
@@ -42,8 +42,28 @@ public class NotificationCenterManager: NSObject,
   @objc
   public static let shared = NotificationCenterManager()
 
-  var delegates: [NotificationDelegate] = []
-  var pendingResponses: [UNNotificationResponse] = []
+  // Delegates are added and removed on whichever thread registers a module or adds a JS
+  // listener, while the UNUserNotificationCenter callbacks read them on another thread.
+  // Arrays are value types with copy-on-write storage, so unsynchronized access here tears
+  // the buffer apart. Every read and write goes through `stateLock`, and the getters hand
+  // back a copy: callers iterate a snapshot instead of holding the lock across a delegate
+  // call, which is free to call back into `addDelegate` or `removeDelegate`.
+  private let stateLock = NSLock()
+  private var _delegates: [NotificationDelegate] = []
+  private var _pendingResponses: [UNNotificationResponse] = []
+
+  var delegates: [NotificationDelegate] {
+    stateLock.lock()
+    defer { stateLock.unlock() }
+    return _delegates
+  }
+
+  var pendingResponses: [UNNotificationResponse] {
+    stateLock.lock()
+    defer { stateLock.unlock() }
+    return _pendingResponses
+  }
+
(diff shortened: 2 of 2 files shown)


Walgermo:
Confirming this is still needed on current SDK 57.

We hit the same intermittent launch SIGSEGV (`Array.append` → `NotificationCenterManager.addDelegate`) and have been shipping an equivalent local mitigation on `expo-notifications@57.0.5` → `57.0.9`: serialize `delegates` / `pendingResponses` behind an `NSLock`, iterate snapshots, and never hold the lock across delegate callbacks.

Checked today against upstream `main` and the published `57.0.9` package: `NotificationCenterManager.swift` still has unsynchronized arrays (unchanged after #48313's multi-delegate chaining). This PR looks like the right fix; it just needs a rebase onto current `main` after that merge.

Related: #47702, #48202 (both closed by bot for missing deterministic repro — expected for a thread-timing race).

PR #48151, opened 2026-07-27
[android][audio] Pause playback when audio output disconnects

### LINKED ISSUE #48138: [expo-audio][Android] Playback does not pause on headphone unplug / Bluetooth disconnect (ACTION_AUDIO_BECOMING_NOISY unhandled) — regression from expo-av
### Summary

On **Android**, `expo-audio` playback does **not** pause when wired headphones are unplugged or a Bluetooth audio device disconnects. Instead, audio is rerouted to the built-in speaker and keeps playing at full volume. On **iOS** this works correctly (playback pauses).

This appears to be a **regression from `expo-av`**, and it also contradicts the current docs, which state (without a platform qualifier):

> Note that audio automatically stops if headphones/Bluetooth audio devices are disconnected.
> — https://docs.expo.dev/versions/latest/sdk/audio/

For a meditation / audiobook / podcast style app, this is a real UX problem: a user who unplugs their headphones in a public place gets the audio suddenly blasting from the phone speaker.

Is the Android behavior intended, or is this an oversight? Would you accept a PR that opts into media3's built-in handling?

### Root cause (verified in the shipped source)

Android's "audio becoming noisy" is a separate mechanism from audio focus — it is delivered via the `ACTION_AUDIO_BECOMING_NOISY` broadcast, and ExoPlayer/media3 only reacts to it if the app opts in via `setHandleAudioBecomingNoisy(true)` (which **defaults to `false`**).

In `expo-audio`'s `AudioPlayer.kt`, the `ExoPlayer.Builder` never enables it:

```kotlin
// packages/expo-audio/android/src/main/java/expo/modules/audio/AudioPlayer.kt
ExoPlayer.Builder(context)
  .setLooper(context.mainLooper)
  .setAudioAttributes(AudioAttributes.DEFAULT, false)
  .setSeekForwardIncrementMs(SEEK_JUMP_INTERVAL_MS)
  .setSeekBackIncrementMs(SEEK_JUMP_INTERVAL_MS)
  .build()
  // no .setHandleAudioBecomingNoisy(true)
```

There is also no `BroadcastReceiver` for `ACTION_AUDIO_BECOMING_NOISY` anywhere in the Android module (the only receiver present is `RingerModeReceiver`, for `RINGER_MODE_CHANGED_ACTION`, which is unrelated to output routing). The audio-focus listener in `AudioModule.kt` only handles `AUDIOFOCUS_LOSS/_TRANSIENT/_CAN_DUCK/_GAIN`, none of which are reliably triggered by a plain headset unplug or BT disconnect, so it does not cover this case.

By contrast, the iOS module *does* handle route changes:

```swift
// packages/expo-audio/ios/AudioModule.swift
case .oldDeviceUnavailable:
  pauseAllPlayers()
```

**Regression note:** the legacy `expo-av` Android module (`AVManager.java`) registered a `BroadcastReceiver` for `ACTION_AUDIO_BECOMING_NOISY` that abandoned audio focus and paused the focus-requiring players. That behavior was not carried over to `expo-audio`.

### Proposed fix

Enable media3's built-in handling (this is what `expo-av` effectively did and what Google's UAMP reference app does):

```kotlin
ExoPlayer.Builder(context)
  // ...
  .setHandleAudioBecomingNoisy(true)
  .build()
```

Happy to open a PR if you're open to it. If you'd prefer it be opt-in from JS instead of always-on, that works too — just let us know the preferred shape.

https://github.com/childrentime/expo-audio-android-becoming-noisy-repro

Built from `create-expo-app@latest` (blank-typescript). The only change is `App.tsx`: create a player, loop a remote track, play it. The bug is also provable directly from the shipped Android source above (missing `setHandleAudioBecomingNoisy` / no becoming-noisy receiver), so it does not depend on app-specific code.

Note: this can only be observed on a **physical Android device** — emulators can't unplug headphones or route Bluetooth realistically.

### Steps to reproduce

Platform: **Android** (physical device). Environment: development build / standalone. Package manager: npm.

1. Create/run an app that plays audio with `expo-audio` (`createAudioPlayer(...).play()`).
2. Connect wired headphones (or a Bluetooth headset) and start playback so audio is coming out of the headphones.
3. Unplug the wired headphones (or turn off / disconnect the Bluetooth headset).

- **Expected:** playback pauses (as it does on iOS, and as the docs state).
- **Actual:** playback continues, rerouted to the phone's built-in speaker at full volume.

### Environment

Reproduced on the **latest** SDK (repro repo above); the shipped `AudioPlayer.kt` at `expo-audio@57.0.3` still builds the `ExoPlayer` without `setHandleAudioBecomingNoisy` and registers no becoming-noisy receiver. Also present on SDK 54 (`expo-audio@1.1.1`).

```text
  System:
    OS: macOS 26.3
  Binaries:
    Node: 22.22.0
    npm: 10.9.4
  npmPackages:
    expo: ~57.0.8
    expo-audio: ~57.0.3
    react: 19.2.3
    react-native: 0.86.0
  Expo Workflow: bare
```

### Expo Doctor Diagnostics

```text
Running 18 checks on your project...
```

vivekjm:
# Why

On Android, expo-audio keeps playing through the device speaker when wired headphones or a Bluetooth audio route disconnects. This differs from iOS and from the documented behavior.

Closes #48138.

# How

Enable Media3's built-in audio-becoming-noisy handling on both `AudioPlayer` and `AudioPlaylist`. Media3 then pauses playback when Android reports that the current output route became noisy.

# Test Plan

- `pnpm run format --check` in `packages/expo-audio`
- `pnpm run lint` in `packages/expo-audio`
- `pnpm run typecheck` in `packages/expo-audio`
- `git diff --check`
- Attempted `./gradlew :expo-audio:compileDebugKotlin --no-daemon` from `apps/bare-expo/android`; Gradle stopped during settings evaluation because the local `node` subprocess exited before the expo-audio compile task ran.
- Not tested on a physical Android device; reproducing the route-disconnect event requires wired headphones or Bluetooth hardware.

A reviewer can verify manually by starting playback on a physical Android device through headphones, then disconnecting the wired or Bluetooth route. Playback should pause instead of moving to the speaker.

# Checklist

- [ ] This diff will work correctly for `npx expo prebuild` & EAS Build (not applicable; no config plugin changes).
- [ ] Conforms with the Documentation Writing Style Guide (not applicable; no documentation changes).

### DIFF
--- a/packages/expo-audio/CHANGELOG.md
+++ b/packages/expo-audio/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 ### 🐛 Bug fixes
 
+- [Android] Pause audio players and playlists when headphones or Bluetooth audio devices disconnect. ([#48138](https://github.com/expo/expo/issues/48138) by [@vivekjm](https://github.com/vivekjm))
 - [Android] Give the lock-screen `MediaSession` instances a unique ID so concurrent active players (and the basic session) no longer collide on the empty default. ([#47101](https://github.com/expo/expo/issues/47101) by [@tsushanth](https://github.com/tsushanth))
 - [Android] Fix stale lock screen artwork when updating metadata without an `artworkUrl`. ([#45738](https://github.com/expo/expo/pull/45738) by [@behenate](https://github.com/behenate))
 - [Android] Fix recording crash in apps wrapped with Microsoft Intune. ([#47005](https://github.com/expo/expo/pull/47005) by [@alanjhughes](https://github.com/alanjhughes))
--- a/packages/expo-audio/CHANGELOG.md
+++ b/packages/expo-audio/CHANGELOG.md
@@ -14,7 +14,7 @@
 
 ### 🐛 Bug fixes
 
-- [Android] Pause audio players and playlists when headphones or Bluetooth audio devices disconnect. ([#48138](https://github.com/expo/expo/issues/48138) by [@vivekjm](https://github.com/vivekjm))
+- [Android] Pause audio players and playlists when headphones or Bluetooth audio devices disconnect. ([#48151](https://github.com/expo/expo/pull/48151) by [@vivekjm](https://github.com/vivekjm))
 - [Android] Give the lock-screen `MediaSession` instances a unique ID so concurrent active players (and the basic session) no longer collide on the empty default. ([#47101](https://github.com/expo/expo/issues/47101) by [@tsushanth](https://github.com/tsushanth))
 - [Android] Fix stale lock screen artwork when updating metadata without an `artworkUrl`. ([#45738](https://github.com/expo/expo/pull/45738) by [@behenate](https://github.com/behenate))
 - [Android] Fix recording crash in apps wrapped with Microsoft Intune. ([#47005](https://github.com/expo/expo/pull/47005) by [@alanjhughes](https://github.com/alanjhughes))
--- a/packages/expo-audio/android/src/main/java/expo/modules/audio/AudioPlayer.kt
+++ b/packages/expo-audio/android/src/main/java/expo/modules/audio/AudioPlayer.kt
@@ -36,6 +36,7 @@ class AudioPlayer(
   player = ExoPlayer.Builder(context)
     .setLooper(context.mainLooper)
     .setAudioAttributes(AudioAttributes.DEFAULT, false)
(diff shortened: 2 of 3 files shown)


behenate (approved):
Thanks! Looks good!

PR #48158, opened 2026-07-27
[expo-type-information] Improve generated imports

HubertBer:
# Why
When running `module-interface` command multiple TypeScript files are generated which import from each other. For example when generating `module.ts`,  we've just imported everything defined in **Module.types.ts** file and **Common.types.ts** and always imported `SharedObject` as we may define it there. This has lead to potentially many unused imports and also it hasn't split the imports into `import type` and plain `import`. 

# How
Add helpers to retrieve used and declared identifiers from TypeScript AST nodes. Use these helpers to find what identifiers are used and declared in generated files. Import only the used and undeclared types/values (if something is only used as a type, then import it as a type).

# Test Plan
- [x] updated snapshots
- [x] checked manually on `expo-contacts`, `expo-blob`

# Checklist


HubertBer:
@expo-bot review

PR #48162, opened 2026-07-27
[@expo/image-utils] Resolve `sharp` directly instead of via global `sharp-cli`

zoontek:
# Why

Closes https://github.com/expo/expo/pull/48036

`@expo/image-utils` resolved `sharp` through a global `sharp-cli` installation and shelled out to it to process images. `sharp-cli`'s last release (5.2.0, June 2025) pins an exact `sharp@0.34.2` and hasn't been updated since, while `sharp` itself is now at `0.35.x`.

Requiring `sharp-cli` locks users to an old, unmaintained `sharp` version for no real benefit over calling `sharp` directly.

# How

- Resolve `sharp` directly (globally or locally), instead of through `sharp-cli`.
- Bump minimum supported version to `sharp@^0.35.0`.
- Drop the `sharp-cli` dependency; move `@expo/spawn-async` to `devDependencies` (still used by the
  e2e test only).
- Update the e2e test, warnings, and README to reference `sharp` instead of `sharp-cli`.

# Test Plan

- `et check-packages @expo/image-utils` passes.
- Added a unit test asserting image processing never spawns a subprocess.
- Updated the e2e test to install `sharp@^0.35.0` globally and verify resolution.

# Checklist

### DIFF
--- a/packages/@expo/image-utils/src/sharp.ts
+++ b/packages/@expo/image-utils/src/sharp.ts
@@ -1,19 +1,17 @@
-import { resolveFrom, resolveGlobal } from '@expo/require-utils';
-import spawnAsync from '@expo/spawn-async';
-import assert from 'assert';
+import { resolveGlobal } from '@expo/require-utils';
 import chalk from 'chalk';
+import fs from 'fs';
 import path from 'path';
 import semver from 'semver';
+import type sharp from 'sharp';
 
 import { env } from './env';
-import type { Options, SharpCommandOptions, SharpGlobalOptions } from './sharp.types';
+import type { SharpCommandOptions, SharpGlobalOptions } from './sharp.types';
 
-const SHARP_HELP_PATTERN = /\n\nSpecify --help for available options/g;
-const SHARP_REQUIRED_VERSION = '^5.2.0';
+const SHARP_REQUIRED_VERSION = '^0.35.0';
 
 export async function resizeBufferAsync(buffer: Buffer, sizes: number[]): Promise<Buffer[]> {
   const sharp = await findSharpInstanceAsync();
-  assert(sharp, 'Sharp is being used while its not initialized');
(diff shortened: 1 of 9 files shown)


kitten (commented):
We had a ticket for this, I believe (Update: yes we do, DVT-112). @byCedric has more knowledge on the constraints here, but I think we didn't drop `sharp-cli` since this is a breaking change in the current way in which we implement this.

Given the `sharp` change we can maybe even consider dropping the globals approach (imho). But we still have to resolve the issue of switching to that other mac-built-in image transformer as a preferrred target before, since I think we'll otherwise leave users in an awkward middle-ground if we don't complete this within one SDK cycle

packages/@expo/image-utils/src/sharp.ts
```diff
@@ -57,162 +55,149 @@ export async function sharpAsync(
...
-
-    let sharpPath: string | null = null;
-    if (sharpCliPackagePath) {
-      sharpPath = resolveFrom(sharpCliPackagePath, 'sharp');
-      if (!sharpPath) {
-        throw Object.assign(
-          new Error(`"sharp" not found, while resolving from "${sharpCliPackagePath}"`),
-          { code: 'MODULE_NOT_FOUND' }
-        );
+      return require.resolve('sharp');
```
kitten, line 125: nit: we should consider dropping this fallback if we're updating this code and introducing a breaking change. `require.resolve` via `__dirname` here isn't sane w/o us adding a peer dependency, and the fallback to `process.cwd()` from `resolveGlobal` is more explicit and preferred usually

zoontek:
@kitten I kept the "global setup" part, but I agree that we should install `sharp` and don't bother about it (as 0.35.x don't relyy on `postinstall` or `node-gyp` anymore)

PR #48168, opened 2026-07-27
[ios][core] Add scene-aware geometry helpers

alanjhughes:
# Why

iOS 27 makes iPhone apps freely resizable, both in iPhone Mirroring and when running on iPad, and rebuilding against the iOS 27 SDK opts an app in with no way back out. `UIScreen.main` and "the first key window in any connected scene" stop describing the space a view actually has, because either can belong to a differently sized scene.

Several of our packages read geometry that way. This is the base of a stack that moves them off it.

# How

Adds `SceneGeometry` to expo-modules-core. Every lookup starts from the view asking the question and degrades in one documented order, instead of each package reinventing that. Interface orientation comes from `effectiveGeometry`, which also clears the iOS 26 deprecation of `UIWindowScene.interfaceOrientation`.

Worth knowing for review: on iOS, `effectiveGeometry` carries no size below iOS 26, so available space comes from the view's own window rather than from scene geometry.

Nothing calls this yet. The rest of the stack does.

# Test Plan

Unit tests in `ios/Tests/SceneGeometryTests.swift`.

```
et native-unit-tests -p ios --packages expo-modules-core
```

### DIFF
--- a/packages/expo-modules-core/ios/Tests/SceneGeometryTests.swift
+++ b/packages/expo-modules-core/ios/Tests/SceneGeometryTests.swift
@@ -1,0 +1,74 @@
+// Copyright 2026-present 650 Industries. All rights reserved.
+
+#if os(iOS) || os(tvOS)
+
+import Testing
+import UIKit
+
+@testable import ExpoModulesCore
+
+@Suite("SceneGeometry")
+@MainActor
+struct SceneGeometryTests {
+  @Test
+  func `display scale prefers the first specified candidate`() {
+    #expect(SceneGeometry.resolveDisplayScale(candidates: [3, 2]) == 3)
+  }
+
+  @Test
+  func `display scale skips unspecified candidates`() {
+    // UITraitCollection documents an unspecified displayScale as 0.0, so a zero
+    // must not win over a real scale further down the chain.
+    #expect(SceneGeometry.resolveDisplayScale(candidates: [nil, 0, 2]) == 2)
+  }
+
+  @Test
+  func `display scale falls back to one when nothing is specified`() {
+    #expect(SceneGeometry.resolveDisplayScale(candidates: [nil, 0]) == 1)
+  }
+
+  @Test
+  func `bounds come from the view's own window when several exist`() {
+    let other = UIWindow(frame: CGRect(x: 0, y: 0, width: 320, height: 480))
+    let host = UIWindow(frame: CGRect(x: 0, y: 0, width: 744, height: 1133))
+    let view = UIView()
+    host.addSubview(view)
+
+    #expect(SceneGeometry.bounds(for: view) == host.bounds)
+    #expect(SceneGeometry.bounds(for: view) != other.bounds)
+  }
+
+  @Test
+  func `bounds follow the view as it moves between windows`() {
+    let first = UIWindow(frame: CGRect(x: 0, y: 0, width: 320, height: 480))
+    let second = UIWindow(frame: CGRect(x: 0, y: 0, width: 744, height: 1133))
+    let view = UIView()
+
+    first.addSubview(view)
+    #expect(SceneGeometry.bounds(for: view) == first.bounds)
+
+    second.addSubview(view)
+    #expect(SceneGeometry.bounds(for: view) == second.bounds)
+  }
+
+  @Test
+  func `safe area size comes from the view's own window when several exist`() {
+    let other = UIWindow(frame: CGRect(x: 0, y: 0, width: 320, height: 480))
+    let host = UIWindow(frame: CGRect(x: 0, y: 0, width: 744, height: 1133))
+    let view = UIView()
+    host.addSubview(view)
+    host.layoutIfNeeded()
+
+    #expect(SceneGeometry.safeAreaSize(for: view) == host.safeAreaLayoutGuide.layoutFrame.size)
+    #expect(SceneGeometry.safeAreaSize(for: view) != other.safeAreaLayoutGuide.layoutFrame.size)
+  }
+
+  @Test
+  func `display scale is always positive`() {
+    // Callers divide by this value, so it must never be zero regardless of how much of the
+    // environment is available.
+    #expect(SceneGeometry.displayScale(for: UIView()) > 0)
+  }
+}
+
+#endif
--- a/packages/expo-modules-core/ios/Utilities/SceneGeometry.swift
+++ b/packages/expo-modules-core/ios/Utilities/SceneGeometry.swift
@@ -1,0 +1,92 @@
+// Copyright 2026-present 650 Industries. All rights reserved.
+
+#if os(iOS) || os(tvOS)
+
+import UIKit
+
+/**
+ Geometry lookups that start from the view asking the question, because iOS 27 makes iPhone apps
+ freely resizable. Neither `UIScreen.main` nor the first key window in any connected scene describes
+ the space a given view has, since either can belong to a differently sized scene.
+ */
+public enum SceneGeometry {
+  public static func windowScene(for view: UIView? = nil) -> UIWindowScene? {
+    if let scene = view?.window?.windowScene {
+      return scene
+    }
+    let scenes = windowScenes()
+    return scenes.first { $0.activationState == .foregroundActive }
+      ?? scenes.first { $0.activationState == .foregroundInactive }
+      ?? scenes.first
+  }
+
+  /// Deliberately strict, for callers deciding whether to attach a window to a scene at all.
+  public static func foregroundActiveScene() -> UIWindowScene? {
+    return windowScenes().first { $0.activationState == .foregroundActive }
+  }
+
+  private static func windowScenes() -> [UIWindowScene] {
+    return UIApplication.shared.connectedScenes.compactMap { $0 as? UIWindowScene }
+  }
+
+  public static func keyWindow(for view: UIView? = nil) -> UIWindow? {
+    guard let scene = windowScene(for: view) else {
+      return nil
+    }
+    return scene.windows.first { $0.isKeyWindow } ?? scene.windows.first
+  }
+
+  public static func bounds(for view: UIView? = nil) -> CGRect {
+    return (view?.window ?? keyWindow(for: view))?.bounds ?? .zero
+  }
+
+  public static func safeAreaSize(for view: UIView? = nil) -> CGSize {
+    guard let window = view?.window ?? keyWindow(for: view) else {
+      return .zero
+    }
+    let safeArea = window.safeAreaLayoutGuide.layoutFrame.size
+    if safeArea.width > 0 && safeArea.height > 0 {
+      return safeArea
+    }
+    return window.bounds.size
+  }
+
+  /// Never cache the result. Scale can differ per scene.
+  public static func displayScale(for view: UIView? = nil) -> CGFloat {
+    return resolveDisplayScale(candidates: [
+      view?.traitCollection.displayScale,
+      windowScene(for: view)?.traitCollection.displayScale,
+      UITraitCollection.current.displayScale
+    ])
+  }
+
+  /// Returns the first positive candidate. `UITraitCollection` reports an unspecified scale as 0.
+  internal static func resolveDisplayScale(candidates: [CGFloat?]) -> CGFloat {
+    for case let candidate? in candidates where candidate > 0 {
+      return candidate
+    }
+    return 1
+  }
+}
+
+#if os(iOS)
+public extension SceneGeometry {
+  /// Reads `effectiveGeometry` because `UIWindowScene.interfaceOrientation` is deprecated as of iOS 26.
+  static func interfaceOrientation(for view: UIView? = nil) -> UIInterfaceOrientation {
+    return windowScene(for: view)?.effectiveGeometry.interfaceOrientation ?? .unknown
+  }
+
+  /// Returns `nil` when the system won't say. A requested lock is only a preference on iOS 27.
+  static func isInterfaceOrientationLocked(for view: UIView? = nil) -> Bool? {
+    guard let scene = windowScene(for: view) else {
+      return nil
+    }
+    if #available(iOS 26.0, *) {
+      return scene.effectiveGeometry.isInterfaceOrientationLocked
+    }
+    return nil
+  }
+}
+#endif
+
+#endif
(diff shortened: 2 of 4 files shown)


alanjhughes:
* **#48175** <a href="https://app.graphite.com/github/pr/expo/expo/48175?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#48174** <a href="https://app.graphite.com/github/pr/expo/expo/48174?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#48173** <a href="https://app.graphite.com/github/pr/expo/expo/48173?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#48172** <a href="https://app.graphite.com/github/pr/expo/expo/48172?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#48171** <a href="https://app.graphite.com/github/pr/expo/expo/48171?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#48170** <a href="https://app.graphite.com/github/pr/expo/expo/48170?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#48169** <a href="https://app.graphite.com/github/pr/expo/expo/48169?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#48168** <a href="https://app.graphite.com/github/pr/expo/expo/48168?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/48168?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/expo-modules-core/ios/Utilities/SceneGeometry.swift
```diff
@@ -0,0 +1,89 @@
...
+#if os(iOS) || os(tvOS)
+
+import UIKit
+
+/**
+ Geometry lookups that start from the view asking the question, because iOS 27 makes iPhone apps
+ freely resizable. Neither `UIScreen.main` nor the first key window in any connected scene describes
+ the space a given view has, since either can belong to a differently sized scene.
+ */
+public enum SceneGeometry {
```
tsapeta, line 12: [🏗️ **design**] This whole enum reads main-actor-isolated state (`UIApplication.shared.connectedScenes`, `UIView.window`, `UIWindowScene`, `traitCollection`) but nothing is annotated `@MainActor`.

The existing helper in the same package does annotate it:

```swift
// Utilities.swift:145
@MainActor
public static func keyWindow() -> UIWindow? {
```

Since `expo-modules-core` is on `swift_version = '6.0'`, callers outside the main actor will hit isolation errors and reach for `MainActor.assumeIsolated` at each call site, which is precisely the duplication this type is meant to eliminate. Marking the enum `@MainActor` pushes that decision into one place and makes the contract explicit.

If some caller genuinely needs a nonisolated variant, that's better expressed as an explicit `nonisolated` overload that does the hop internally (the package already has `performSynchronouslyOnMainActor` for this) than by leaving the whole surface unannotated.

Worth settling now rather than after the stack lands: adding `@MainActor` later is source-breaking for every call site.

packages/expo-modules-core/ios/Utilities/SceneGeometry.swift
```diff
@@ -0,0 +1,89 @@
...
+    for case let candidate? in candidates where candidate > 0 {
+      return candidate
+    }
+    return 1
+  }
+}
+
+#if os(iOS)
+public extension SceneGeometry {
+  /// Reads `effectiveGeometry` because `UIWindowScene.interfaceOrientation` is deprecated as of iOS 26.
```
tsapeta, line 74: [💡 suggestion] This comment isn't accurate. Checking the iOS 26.5 SDK headers:

```objc
// UIWindowScene.h:33
@property (nonatomic, readonly) UIWindowSceneGeometry *effectiveGeometry API_AVAILABLE(ios(16.0)) API_UNAVAILABLE(watchos);

// UIWindowSceneGeometry.h:42
@property (nonatomic, readonly) UIInterfaceOrientation interfaceOrientation API_UNAVAILABLE(tvos);
```

`effectiveGeometry` has existed since iOS 16.0, which is below the package's 16.4 deployment target, and `interfaceOrientation` on it inherits the class's iOS 16.0 availability. So this isn't a new iOS 26 API being adopted, it's the pre-existing one whose older sibling (`UIWindowScene.interfaceOrientation`) got deprecated in 26.

The deprecation part of the sentence is correct. Suggest rewording so the comment doesn't imply a version floor that isn't there, e.g. "`UIWindowScene.interfaceOrientation` is deprecated as of iOS 26 in favor of `effectiveGeometry.interfaceOrientation`, which has been available since iOS 16."

packages/expo-modules-core/ios/Utilities/SceneGeometry.swift
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  public static func keyWindow(for view: UIView? = nil) -> UIWindow? {
+    guard let scene = windowScene(for: view) else {
+      return nil
+    }
+    return scene.windows.first { $0.isKeyWindow } ?? scene.windows.first
+  }
+
+  /// `effectiveGeometry` carries no size below iOS 26, so the window is the best source we have.
```
tsapeta: [🏗️ **design**] The comment explains a decision the code doesn't make. It says `effectiveGeometry` carries no size below iOS 26, which frames this as a version-driven fallback, but the implementation never reads `effectiveGeometry` at all, on any version.

There is a defensible reason to always prefer the window: for a view, `view.window.bounds` is the space it actually has, which is more precise than the scene's system-space geometry (a scene can host multiple differently-sized windows). If that's the reasoning, the comment should say so, since it's a durable design statement rather than a temporary version workaround.

If instead the intent is to use `effectiveGeometry.systemFrame` on iOS 26+ and fall back below that, then the code is missing that branch.

Either way, as written a reader will wonder whether an `if #available(iOS 26)` branch was dropped.

packages/expo-modules-core/ios/Utilities/SceneGeometry.swift
```diff
@@ -0,0 +1,89 @@
...
+    if let scene = view?.window?.windowScene {
+      return scene
+    }
+    let scenes = UIApplication.shared.connectedScenes.compactMap { $0 as? UIWindowScene }
+    return scenes.first { $0.activationState == .foregroundActive }
+      ?? scenes.first { $0.activationState == .foregroundInactive }
+      ?? scenes.first
+  }
+
+  public static func foregroundActiveScene() -> UIWindowScene? {
```
tsapeta, line 24: [💡 suggestion] This is a strict subset of `windowScene(for:)` with no view argument: that function's fallback chain already tries `foregroundActive` first. The only behavioural difference is that this returns `nil` instead of degrading to `foregroundInactive` or the first connected scene.

Since nothing in the repo calls it yet, it's hard to tell what need it serves. If a caller genuinely wants "foreground-active or nothing", the name should convey that the strictness is the point (it currently reads as a convenience shortcut). If nothing needs it, dropping it keeps the public surface, and the ABI commitment on a prebuilt package, smaller.

Separately, the `first { ... } as? UIWindowScene` order differs from `windowScene(for:)`, which does `compactMap { $0 as? UIWindowScene }` before filtering. This version finds the first foreground-active scene of *any* type and then casts, so a foreground-active non-window scene (e.g. a `CPTemplateApplicationScene` on CarPlay) makes it return `nil` even when a foreground-active `UIWindowScene` exists. Filtering by type first, as the other function does, avoids that.

packages/expo-modules-core/ios/Utilities/SceneGeometry.swift
```diff
@@ -0,0 +1,89 @@
...
+    return (view?.window ?? keyWindow(for: view))?.bounds ?? .zero
+  }
+
+  public static func safeAreaSize(for view: UIView? = nil) -> CGSize {
+    guard let window = view?.window ?? keyWindow(for: view) else {
+      return .zero
+    }
+    let safeArea = window.safeAreaLayoutGuide.layoutFrame.size
+    return CGSize(
+      width: safeArea.width > 0 ? safeArea.width : window.bounds.width,
```
tsapeta: [🏗️ **design**] Substituting `window.bounds` for a zero-width or zero-height safe area conflates two different situations. A genuinely zero safe area is possible (a window laid out before the safe area resolves, or an embedded/resized scene), and in those cases returning the full window bounds silently over-reports the usable space rather than reporting the truth.

The per-axis fallback also lets the result mix sources: a real safe-area width paired with a full-bounds height, which corresponds to no actual geometry.

Given the PR's framing (iOS 27 resizable windows, where getting the available space wrong is the failure mode being fixed), consider either returning the safe area as-is and letting callers decide, or applying the fallback for the whole size when the safe area is entirely degenerate. If the current behaviour is deliberate, a comment on which situation it's papering over would help, since the intent isn't recoverable from the code.

packages/expo-modules-core/ios/Tests/SceneGeometryTests.swift
```diff
@@ -0,0 +1,56 @@
+// Copyright 2026-present 650 Industries. All rights reserved.
+
+#if os(iOS) || os(tvOS)
+
+import Testing
+import UIKit
+
+@testable import ExpoModulesCore
+
+@Suite("SceneGeometry")
```
tsapeta, line 10: [🏗️ **design**] Four of the six tests target `resolveDisplayScale`, a three-line loop over an injected array. The functions that carry the actual risk are untested:

- `windowScene(for:)`'s fallback chain (foregroundActive → foregroundInactive → first). This is the core of the PR's thesis and the ordering is a real judgment call.
- `keyWindow(for:)`'s `isKeyWindow` → `windows.first` degradation.
- `bounds(for:)` / `safeAreaSize(for:)` when the view has no window (the `.zero` path), and `safeAreaSize`'s zero-fallback branch flagged above.
- `interfaceOrientation(for:)` returning `.unknown` with no scene.

The windowless paths are easy to cover without a live scene: `SceneGeometry.bounds(for: UIView()) == .zero`, and the same for `safeAreaSize`. Those pin the documented degradation order, which is the contract the rest of the stack will depend on.

The scene-dependent chain is genuinely awkward to unit test against real `UIWindowScene`s. If it isn't practical here, it'd be worth extracting the selection logic to take a `[UIWindowScene]`-like input the way `resolveDisplayScale` already takes its candidates, which is a nice pattern in this file and worth applying to the part that's harder to get right.

packages/expo-modules-core/ios/Tests/SceneGeometryTests.swift
```diff
@@ -0,0 +1,56 @@
...
+    #expect(SceneGeometry.resolveDisplayScale(candidates: [nil, 0, 2]) == 2)
+  }
+
+  @Test
+  func `display scale falls back to one when nothing is specified`() {
+    #expect(SceneGeometry.resolveDisplayScale(candidates: [nil, 0]) == 1)
+  }
+
+  @Test
+  func `bounds come from the window the view lives in`() {
```
tsapeta: [💡 suggestion] These two tests assert that `SceneGeometry.bounds(for:)` and `safeAreaSize(for:)` equal the window's own values, but they're constructed so that the fallback path would satisfy them too.

The window here is never made key and belongs to no scene, so `view.window` is the only branch that can produce a non-zero result. That does confirm the primary path, but the assertion `== window.bounds` would also hold if the implementation returned the same window via some other route.

A sharper version: create two windows with different frames, add the view to the non-key one, and assert the result matches the view's own window rather than the other. That's the property the PR is actually about, i.e. that geometry follows the view instead of whichever window happens to be key.

packages/expo-modules-core/ios/Utilities/SceneGeometry.swift
alanjhughes: It doesn't. The comment says UIWindowScene.interfaceOrientation is deprecated as of iOS 26, which it is, and never claims effectiveGeometry is new.

packages/expo-modules-core/ios/Utilities/SceneGeometry.swift
alanjhughes: `systemFrame` is a macCatalyst api. What is this thing talking about 😅

packages/expo-modules-core/ios/Utilities/SceneGeometry.swift
alanjhughes: All the callers are already on the main actor because they were interacting with the window. We'd need wider changes to do this. It's a drop in replacement at all the call sites as it is right now so I'd lean towards leaving it rather than refactoring every call site

packages/expo-modules-core/ios/Utilities/SceneGeometry.swift
alanjhughes: fixed

packages/expo-modules-core/ios/Tests/SceneGeometryTests.swift
alanjhughes: Fixed

packages/expo-modules-core/ios/Utilities/SceneGeometry.swift
tsapeta: Makes sense to skip it now, but I think we should update these call sites to re-state MainActor requirement at some point.

packages/expo-modules-core/ios/Utilities/SceneGeometry.swift
alanjhughes: Fixed

packages/expo-modules-core/ios/Tests/SceneGeometryTests.swift
alanjhughes: The only thing meaningful to add here requires an actual scene. Extracting the selection logic doesn't seem useful to me
```swift
resolveForegroundActiveScene(from: scenes)
  ?? scenes.first { $0.activationState == .foregroundInactive }
  ?? scenes.first
```
We'd just be testing the `??` does what we expect

tsapeta (approved):

alanjhughes:
### Merge activity

* **Jul 28, 2:42 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/expo/expo/48168).
* **Jul 28, 2:44 PM UTC**: @alanjhughes merged this pull request with [Graphite](https://app.graphite.com/github/pr/expo/expo/48168).

PR #48181, opened 2026-07-27
[image] Add `transition.skipOnCacheHit` to skip the fade for cached images

janicduplessis:
# Why

`expo-image` plays the `transition` every time an image is displayed. So an image re-fades whenever its view re-mounts — a screen re-renders, a tab switches, a list row scrolls back into view — even though the bitmap is already cached and could be shown instantly. There's no way to keep the fade for genuinely new images while skipping it for ones that are already loaded.

# How

Adds a `skipOnCacheHit` field to the `transition` object:

```tsx
<Image source="…" transition={{ duration: 300, skipOnCacheHit: 'all' }} />
```

- `'none'` (default) — always play; unchanged from today.
- `'memory'` — skip the fade for in-memory cache hits (the instant case). Disk and network loads still animate.
- `'all'` — skip the fade for any cache hit (memory or disk). Only network loads animate.

The key detail: **this only gates the *first appearance* of an image in a view.** A transition caused by the `source` changing on an already-populated view always plays, because that's a genuine visual change rather than a re-mount. "First appearance" is detected from whether a real (non-placeholder) image was previously shown; a `recyclingKey` change resets that state so recycled list rows are gated too.

The cache source is already known natively when the image is displayed, so nothing new is tracked — it's threaded into where the transition is set up. iOS reads the existing `SDImageCacheType` from the load callback; Android reads Glide's `DataSource` via the request listener onto the target. Web has no reliable "served from HTTP cache" signal, so it keeps the always-play behavior; the field is documented as iOS/Android only.

Why cache tiers and not a network flag: there are only two on-device cache tiers (memory, disk), so `'memory'` (skip only the instant hits) and `'all'` (skip every cache hit) cover the meaningful cases; `'none'` is the default.

# Test Plan

Verified on the iOS simulator and an Android emulator through **native-component-list → Image → Transitions**, which this PR extends with a `Skip on cache hit` control and a **Recycle view (cache hit)** button (bumps `recyclingKey` and swaps between two cached sources, the way a list row is reused for a different item).

Each video is the same **Recycle view (cache hit)** press at `duration: 2000`, `skipOnCacheHit: 'none'` on the left and `'all'` on the right, recorded separately and aligned on the transition. Both sides start and end on the same image, so the only difference is whether it animates.

**iOS**

https://github.com/user-attachments/assets/45000c4d-fa31-4274-8819-7681cdf21687

**Android**

https://github.com/user-attachments/assets/eb0223da-7cb0-4af6-af75-09e67572d390

I also confirmed on both platforms that **Change image (source change)** still fades with `'all'` set, including when the new source is itself cached.

# Checklist

- [x] I added a `changelog.md` entry and rebuilt the package sources according to [this short guide](https://github.com/expo/expo/blob/main/CONTRIBUTING.md#-before-submitting), including regenerating the docs API data (`et generate-docs-api-data -p expo-image`)


packages/expo-image/src/Image.types.ts
```diff
@@ -723,6 +723,36 @@ export type ImageTransition = {
...
     | 'sf:off-up'
     | null;
+
+  /**
+   * Skips the transition when an image first appears from a cache hit, so already-cached images are
+   * shown instantly instead of re-animating on every mount, tab change, or scroll back into view.
+   *
+   * Only the first appearance of an image in a view is affected — a later `source` change on a
+   * populated view always plays its transition. A visible placeholder still counts as the first
+   * appearance.
```
amandeepmittal: ✏️ **style** `prose.md#voice-and-tone`

The opening sentence runs 31 words, and "Only the first appearance of an image in a view is affected" is passive where the sentence before it is active. Split the first sentence and restore the active phrasing.

```suggestion
   * Skips the transition when an image first appears from a cache hit. Already-cached images are
   * then shown instantly instead of re-animating on every mount, tab change, or scroll back into
   * view.
   *
   * This only affects the first appearance of an image in a view. A later `source` change on a
   * populated view always plays its transition. A visible placeholder still counts as the first
   * appearance.
```

packages/expo-image/src/Image.types.ts
```diff
@@ -723,6 +723,36 @@ export type ImageTransition = {
...
+   *
+   * - `'none'` - The transition always plays. This is the default.
+   * - `'memory'` - Skips the transition for images served from the in-memory cache.
+   * - `'all'` - Skips it for any cache hit, so only uncached (typically network) loads animate. On
+   * iOS, locally bundled assets are treated as uncached and still animate.
```
amandeepmittal: ⚠️ **correctness** `comment-format.md#severity`

On iOS a bundled asset resolves as uncached, so neither `'memory'` nor `'all'` skips it. Inside the `'all'` bullet the caveat reads as if `'memory'` behaved differently, so move it below the list. The field is documented for Android too, so say what Android does with bundled assets.

amandeepmittal (changes requested):

packages/expo-image/src/Image.types.ts
janicduplessis: Applied your suggestion as written — first sentence split, and "This only affects the first appearance" restores the active voice.

packages/expo-image/src/Image.types.ts
janicduplessis: Good catch, and it turned out to be worse than misplaced — the caveat was iOS-only because I had only checked iOS. Verified both paths:

- **iOS** — `maybeRenderLocalAsset` resolves bundled assets with `UIImage(named:)` and calls `renderSourceImage(local)` without a cache type, so it defaults to `.none`. SDWebImage is bypassed entirely, so these are always reported uncached and always animate, on every appearance rather than just the first.
- **Android** — bundled assets are handed to Glide as a normal model (`GlideModelProvider`), so the usual tiers apply: `DataSource.LOCAL` on the first load maps to `ImageCacheType.NONE`, but later loads come back as `MEMORY_CACHE`/`RESOURCE_DISK_CACHE` and *are* skipped.

So the two platforms genuinely differ. Moved the note below the list and documented both:

> Locally bundled assets behave differently on each platform. Android loads them through the regular image cache, so they animate on the first load and are skipped after that. iOS renders them without going through the cache, so they always report as uncached and keep animating.

amandeepmittal (approved):
Thank you for adding docs!

Let's wait for the technical review.

packages/expo-image/src/Image.types.ts
```diff
@@ -723,6 +723,40 @@ export type ImageTransition = {
...
+   *
+   * This only affects the first appearance of an image in a view. A later `source` change on a
+   * populated view always plays its transition. A visible placeholder still counts as the first
+   * appearance.
+   *
+   * - `'none'` - The transition always plays. This is the default.
+   * - `'memory'` - Skips the transition for images served from the in-memory cache.
+   * - `'all'` - Skips it for any cache hit, so only uncached (typically network) loads animate.
+   *
+   * Locally bundled assets behave differently on each platform. Android loads them through the
```
tsapeta: Documenting the divergence is fine, but did you consider mapping iOS local assets to `.memory` in `maybeRenderLocalAsset` instead? With `skipOnCacheHit: 'all'`, the user intent is "instantly available images should not fade". Bundled assets are always instantly available, so on iOS they keep fading with no way to opt out. Neither platform can match the other exactly (Android fades once per process, iOS would then never fade), but "never fade" seems closer to the user's intent than "always fade".

tsapeta (approved):

packages/expo-image/src/Image.types.ts
janicduplessis: Implemented — `maybeRenderLocalAsset` now passes `.memory` to `renderSourceImage`, so bundled assets on iOS are skipped under both `'memory'` and `'all'`.

Two things I verified before switching it:

- The cache type at that call site feeds only `transition.shouldPlay`; the local-asset path never emits `onLoad`, so the `cacheType` users see in events is unchanged, and the default `'none'` still always plays.
- It's accurate rather than a workaround: `UIImage(named:)` serves these from the system's in-memory asset cache, so reporting a memory hit describes what actually happens.

A `source` change on a populated view still animates regardless, same as every other skip path. Updated the TSDoc note to match: both platforms now treat bundled assets as cache hits, with Android animating once per process before its cache warms.

janicduplessis: Follow-up: Android now matches too, so the platform divergence I described above is gone. `GlideRequestListener` maps `android.resource://` models to the memory tier for the skip check — that model type precisely identifies bundled resources (files and content URIs use string models), and `target.cacheType` feeds only `shouldPlay`, so the `onLoad` event still reports the real cache tier. Bundled assets are now "always instantly available → never fade" under `'memory'`/`'all'` on both platforms, and the docs note shrank to one line.

PR #48186, opened 2026-07-27
[calendar][iOS] Fix `getDefaultCalendarSync` and `getCalendars` requiring full access

### LINKED ISSUE #48143: [expo-calendar] No apparent way to use `addEventWithForm()` with only write-only permissions

https://github.com/marshmn/expo-calendar-issue

### Steps to reproduce

To reproduce:
- Run the app in the example repo on an iOS device
- Tap the "Add to Calendar" button
- Grant requested permission ("Add Events Only")
- Observe that you see an error that full calendar access is required

The [docs for `addEventWithForm()`](https://docs.expo.dev/versions/latest/sdk/calendar/#addeventwithformoptions) state "Requires at minimum write-only calendar permission." - however, this method needs to be called on an `ExpoCalendar` object, and the ways to obtain that (e.g. using `getDefaultCalendarSync()` as in the example) appear to require full calendar access.

Image

Granting "Full Access" to the calendar permission allows this to work. 

It should be possible to use `addEventWithForm()` without needing to get an `ExpoCalendar` object first - and therefore being able to avoid granting unnecessary permissions.

### Environment

```text
  System:
    OS: Linux 6.19 Debian GNU/Linux 13 (trixie) 13 (trixie)
    Shell: 5.2.37 - /bin/bash
  Binaries:
    Node: 26.5.0 - /usr/local/bin/node
    npm: 11.17.0 - /usr/local/bin/npm
  npmPackages:
    expo: ~57.0.8 => 57.0.8 
    expo-router: ~57.0.8 => 57.0.8 
    react: 19.2.3 => 19.2.3 
    react-dom: 19.2.3 => 19.2.3 
  Expo Workflow: managed
```

### Expo Doctor Diagnostics

```text
```

Wenszel:
# Why

fixes: https://github.com/expo/expo/issues/48143

On iOS, if write-only permissions are granted, `eventStore.defaultCalendarForNewEvents` and `eventStore.calendars(for:)` return a single virtual calendar, so `getDefaultCalendarSync` and `getCalendars` don't have to require full access permissions. This lets apps create new events directly, without the entire UI creation flow via `presentPicker` and `addEventWithForm`. 

docs ref: https://developer.apple.com/documentation/EventKit/accessing-the-event-store#Use-EventKit-with-write-only-calendar-access

### Design decision for `listEvents`:
We could loosen the permissions for listEvents as well. Per docs, `eventStore.events` returns an empty array if write-only permissions are granted. However, if write-only permissions are granted, returning an empty array instead of throwing is ambiguous.

# How

* Replace full access check in `getDefaultCalendarSync`, `getCalendars` and `getCalendarById` with write-only or full access check
* Rename the `CalendarAccessGuard` methods to be more precise 

# Test Plan

Tested on WriteOnlyPermissionsScreen in NCL

### DIFF
--- a/packages/expo-calendar/ios/Next/CalendarNextModule.swift
+++ b/packages/expo-calendar/ios/Next/CalendarNextModule.swift
@@ -29,7 +29,9 @@ public final class CalendarNextModule: Module {
     }
 
     Function("getDefaultCalendarSync") { () -> ExpoCalendar in
-      try calendarAccessGuard?.checkCalendarPermissions()
+      // With write-only permissions granted, `eventStore.defaultCalendarForNewEvents` is a virtual calendar:
+      // https://developer.apple.com/documentation/EventKit/accessing-the-event-store#Use-EventKit-with-write-only-calendar-access
+      try calendarAccessGuard?.ensureCalendarWriteOnlyOrFullAccess()
       guard let defaultCalendar = eventStore.defaultCalendarForNewEvents else {
         throw DefaultCalendarNotFoundException()
       }
@@ -40,21 +42,24 @@ public final class CalendarNextModule: Module {
       let calendars: [EKCalendar]
       switch type {
       case nil:
-        try calendarAccessGuard?.checkCalendarPermissions()
-        try calendarAccessGuard?.checkRemindersPermissions()
+        // With write-only permissions granted, `eventStore.calendars` returns a single virtual calendar:
+        // https://developer.apple.com/documentation/EventKit/accessing-the-event-store#Use-EventKit-with-write-only-calendar-access
+        try calendarAccessGuard?.ensureCalendarWriteOnlyOrFullAccess()
+        try calendarAccessGuard?.ensureRemindersAccess()
         calendars = eventStore.calendars(for: .event) + eventStore.calendars(for: .reminder)
       case .event:
-        try calendarAccessGuard?.checkCalendarPermissions()
+        try calendarAccessGuard?.ensureCalendarWriteOnlyOrFullAccess()
         calendars = eventStore.calendars(for: .event)
       case .reminder:
-        try calendarAccessGuard?.checkRemindersPermissions()
+        try calendarAccessGuard?.ensureRemindersAccess()
         calendars = eventStore.calendars(for: .reminder)
       }
       return calendars.map { ExpoCalendar(calendar: $0) }
     }
 
     AsyncFunction("getCalendarById") { (calendarId: String) -> ExpoCalendar in
-      try calendarAccessGuard?.checkCalendarPermissions()
+      // With write-only access, this can resolve the virtual calendar's identifier.
+      try calendarAccessGuard?.ensureCalendarWriteOnlyOrFullAccess()
       guard let calendar = eventStore.calendar(withIdentifier: calendarId) else {
         throw CalendarIdNotFoundException(calendarId)
       }
@@ -62,6 +67,7 @@ public final class CalendarNextModule: Module {
     }
 
     AsyncFunction("presentPicker") { (promise: Promise) in
+      // No access guard here on purpose. The system picker itself shows a message when permissions aren't granted.
       guard let currentVc = appContext?.utilities?.currentViewController() else {
         throw MissingCurrentViewControllerException()
       }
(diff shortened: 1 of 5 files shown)


Wenszel:
* **#48186** <a href="https://app.graphite.com/github/pr/expo/expo/48186?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/48186?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/expo-calendar/src/Calendar.ts
```diff
@@ -226,12 +230,16 @@ export function getDefaultCalendarSync(): ExpoCalendar {
...
- * a specific [entity type](#entitytypes). Possible values are `Calendar.EntityTypes.EVENT` (for calendars shown in
- * the Calendar app) and `Calendar.EntityTypes.REMINDER` (for the Reminders app).
- * > **Note:** If not defined, you will need both permissions: **CALENDAR** and **REMINDERS**.
- * @return An array of [`ExpoCalendar`](#expocalendar) shared objects matching the provided entity type (if provided).
+ * Gets an array of [`ExpoCalendar`](#expocalendar) shared objects with details about the
+ * calendars available to the app.
+ *
+ * > **iOS 17+:** With write-only calendar access, requesting `EntityTypes.EVENT` returns a
+ * > single virtual calendar, which can be used to create events but does not expose the user's
+ * > real calendars.
```
jakex7: Guess we should also update https://github.com/expo/expo/blob/1434afd35f4d300b86e12a9704687d9916e04876/docs/pages/versions/unversioned/sdk/calendar.mdx?plain=1#L182

jakex7 (approved):
🦤

packages/expo-calendar/ios/Next/CalendarNextModule.swift
```diff
@@ -29,7 +29,9 @@ public final class CalendarNextModule: Module {
     }
 
     Function("getDefaultCalendarSync") { () -> ExpoCalendar in
-      try calendarAccessGuard?.checkCalendarPermissions()
+      // With write-only permissions granted, `eventStore.defaultCalendarForNewEvents` is a virtual calendar:
+      // https://developer.apple.com/documentation/EventKit/accessing-the-event-store#Use-EventKit-with-write-only-calendar-access
+      try calendarAccessGuard?.ensureCalendarWriteOnlyOrFullAccess()
```
jakex7: [💡 suggestion] I verified on the iOS 27.0 simulator (standalone EventKit app, write-only granted) that the virtual calendar is safe to serialize: `defaultCalendarForNewEvents` and `calendars(for: .event)` both return one calendar with id `VIRTUAL_APP_CALENDAR_UUID`, title `Calendar`, `type = .calDAV`, `isImmutable = true`, and a non-nil `source` (`VIRTUAL_APP_SOURCE_UUID`, title `Account`). `calendar(withIdentifier:)` resolves that id, so the `getCalendarById` comment holds.

The stated test plan does not cover these paths, though: the NCL `WriteOnlyPermissionsScreen` obtains its calendar from `presentPicker`, not from `getDefaultCalendarSync`, `getCalendars`, or `ExpoCalendar.get`. Consider adding a step to that screen that calls them under write-only access and shows the returned fields.

packages/expo-calendar/src/Calendar.ts
```diff
@@ -226,12 +230,16 @@ export function getDefaultCalendarSync(): ExpoCalendar {
...
- * Gets an array of [`ExpoCalendar`](#expocalendar) shared objects with details about the different calendars stored on the device.
- * @param entityType __iOS Only.__ Not required, but if defined, filters the returned calendars to
- * a specific [entity type](#entitytypes). Possible values are `Calendar.EntityTypes.EVENT` (for calendars shown in
- * the Calendar app) and `Calendar.EntityTypes.REMINDER` (for the Reminders app).
- * > **Note:** If not defined, you will need both permissions: **CALENDAR** and **REMINDERS**.
- * @return An array of [`ExpoCalendar`](#expocalendar) shared objects matching the provided entity type (if provided).
+ * Gets an array of [`ExpoCalendar`](#expocalendar) shared objects with details about the
+ * calendars available to the app.
+ *
+ * > **iOS 17+:** With write-only calendar access, requesting `EntityTypes.EVENT` returns a
```
jakex7: [🐧 nit] The same applies when `entityType` is omitted: the event portion of the result is the single virtual calendar. Consider "requesting `EntityTypes.EVENT` (or omitting `entityType`) returns ...".

packages/expo-calendar/src/Calendar.ts
```diff
@@ -226,12 +230,16 @@ export function getDefaultCalendarSync(): ExpoCalendar {
...
+ * > single virtual calendar, which can be used to create events but does not expose the user's
+ * > real calendars.
+ *
+ * @param entityType __iOS Only.__ Not required, but if defined, filters the returned calendars to a specific [entity type](#entitytypes).
```
jakex7: [🐧 nit] This rewrite drops the explanation of the possible values (`EntityTypes.EVENT` for calendars shown in the Calendar app, `EntityTypes.REMINDER` for the Reminders app). That was useful; consider keeping it.

PR #48209, opened 2026-07-28
[ios][router] Fix NativeTabs `xcasset` icons and icon type mismatch crash

### LINKED ISSUE #47480: [expo-router] NativeTabs \xcasset` icon renders image sets but not symbol sets on iOS`

https://github.com/grandbig/expo-router-xcasset-bug

### Steps to reproduce

> Note: this is distinct from #46931 (that was the `sf` prop not rendering on SDK 55, and the reporter confirmed it was fixed in SDK 56). Here the `sf` prop works; the bug is in the `xcasset` path, which still converts to a file `uri` on SDK 56/57 and `main`, so custom **symbol sets** never reach `imageNamed:`.

This is a stripped-down `create-expo-app` project. It renders `NativeTabs` with two tabs, both using the `xcasset` icon prop.

A small config plugin registers two assets in the iOS asset catalog during prebuild:

- `reproImage` — a normal **image set** (raster PNG)
- `reproSymbol` — a custom **SF Symbol set**

**Steps:**

1. `npm install` (package manager: npm)
2. `npx expo prebuild -p ios`
3. `npx expo run:ios`

Platform: iOS only (the `xcasset` prop is iOS-only). Environment: development build, iOS Simulator (also reproduces on a device).

**Expected**

Both tab icons render.

`xcasset` is documented to load images from the Xcode asset catalog ([docs](https://docs.expo.dev/router/advanced/native-tabs/)), and expo-router's own `xcasset` prop docs say it resolves via `[UIImage imageNamed:]`. A symbol set is a valid asset-catalog entry, and `imageNamed:` supports it.

**Actual**

The "Image set" tab (`xcasset="reproImage"`) renders.

The "Symbol set" tab (`xcasset="reproSymbol"`) is **blank**, and Metro logs:

```
WARN  Could not find image file://.../<App>.app/reproSymbol.png
```

Both assets compile into the binary correctly (verified with `actool` / `assetutil`), so this is not a missing-asset problem.

Image

**Root cause**

`packages/expo-router/src/native-tabs/utils/optionsIconConverter.ios.ts` → `convertOptionsIconToScreensPropsIcon` converts `xcasset` into an image source `uri`:

```ts
const imageSource = 'xcasset' in icon && icon.xcasset ? { uri: icon.xcasset } : icon.src;
// -> { type: 'templateSource' | 'imageSource', ... }
```

That `{ uri: name }` is handled by `RCTImageLoader` (file lookup). RN's loader can resolve a raster asset-catalog image this way, but it cannot produce a symbol image. That is why image sets work and symbol sets are blank.

react-native-screens already supports the correct path. `parseIconToNativeProps` (TabsScreen.ios) accepts `{ type: 'xcasset', name }`, which maps to the native `RNSTabsIconTypeXcasset` → `[UIImage imageNamed:]` (`RNSTabBarAppearanceCoordinator.mm`) and works for both asset types.

expo-router never emits that shape, so the asset-catalog path is unreachable. This is on the expo-router side, and it is still present on `main` (also reproduced on SDK 56 / expo-router 56.2.x).

**Suggested fix**

Add an `xcasset` case that mirrors the existing `sf` case in `convertOptionsIconToScreensPropsIcon` (which already returns `{ type: 'sfSymbol', name }` correctly):

```ts
if (icon && 'xcasset' in icon && icon.xcasset) {
  return { type: 'xcasset', name: icon.xcasset };
}
```

Verified locally with a patched expo-router: the symbol set renders and tints via `iconColor`. `selectedIcon` uses the same converter, so `xcasset={{ default, selected }}` works too.

### Environment

```text
System:
  OS: macOS 26.5.1
  Shell: zsh 5.9
Binaries:
  Node: 24.14.0
  npm: 11.9.0
Managers:
  CocoaPods: 1.16.2
SDKs:
  iOS SDK: iOS 26.4
IDEs:
  Xcode: 26.4.1/17E202
npmPackages:
  expo: 57.0.1
  expo-router: 57.0.2
  react: 19.2.3
  react-native: 0.86.0
  react-native-web: 0.21.2
Expo Workflow: managed
```

### Expo Doctor Diagnostics

```text
```

CavalcanteLeo:
# Why

Fixes #47480.

`NativeTabs.Trigger.Icon`'s `xcasset` prop is documented to load images from the Xcode asset catalog via `[UIImage imageNamed:]`, but expo-router never emitted the icon shape that reaches that native path. `convertOptionsIconToScreensPropsIcon` turned the asset name into an image source `uri`:

```ts
const imageSource = 'xcasset' in icon && icon.xcasset ? { uri: icon.xcasset } : icon.src;
```

`{ uri: name }` is resolved by `RCTImageLoader`, which can find a raster **image set** by name but can never produce a **symbol set**. So image sets happened to work and symbol sets rendered blank with `WARN Could not find image file://.../<name>.png`.

While fixing that I found a second, related defect in the same area. `NativeTabsView.ios.tsx` converted `icon` with the *normal* icon color and `selectedIcon` with the *selected* icon color. Since that color is what decides between `imageSource` and `templateSource`, setting only one of `iconColor` / `selectedIconColor` made the two icons resolve to different types, and react-native-screens throws on that:

```
Error: [RNScreens] icon and selectedIcon must be same type.
```

That is a hard crash (red screen) for a plausible config — for example `<NativeTabs iconColor={{ selected: 'red' }}>` with any `src` icon. Verified against the real `Tabs.Screen`, not a mock.

# How

**1. Route `xcasset` to the native asset-catalog path** (`optionsIconConverter.ios.ts`)

Give `xcasset` its own branch that mirrors the existing `sf` branch and returns `{ type: 'xcasset', name }`. react-native-screens maps that to `RNSTabsIconTypeXcasset` → `[UIImage imageNamed:]`, which handles image sets and symbol sets alike. This is also what expo-router already does for toolbar icons in `stack-utils/toolbar/shared.ts`.

The `src` branch is now separate, so its `ImageSourcePropType` cast is gone.

**2. Keep `icon` and `selectedIcon` in agreement** (`NativeTabsView.ios.tsx`)

Derive the icon color from both states (`normal ?? selected`) and pass the same value to both conversions. The color is only used as a "should this image be a template" signal, so merging is the right granularity — the rendering mode is a property of the tab bar item, not of one state. This only changes behavior in the case that previously threw.

# Behavior change

`xcasset` icons no longer honor the implicit `iconColor` → `template` rule; their rendering mode now comes from the asset's **Render As** setting in the asset catalog, which is the canonical place to configure it for asset-catalog images (set it to **Template Image** to have the icon tinted). This affects raster image sets set to *Render As: Default* that were relying on the implicit tinting. Symbol sets and image sets set to *Template Image* tint as before.

Note the `renderingMode` prop was already a no-op for `xcasset` on native tabs — `appendIconOptions` never forwarded it — so the docs claiming otherwise were wrong. Both the TSDoc and the guide are corrected here.

# Test Plan

Added, red before the change and green after:

- `native-tabs/utils/__tests__/optionsIconConverter.test.ios.tsx` — a new `xcasset` block: the converter returns `{ type: 'xcasset', name }`, keeps that type when an icon color is set, falls through on an empty name, is still overridden by `sf`, and produces a matching type for the normal and selected states. The `xcasset` branch previously had no test coverage at all.
- `native-tabs/__tests__/options.e2e.test.ios.tsx` — integration coverage that `<Icon xcasset={{ default, selected }} />` reaches `Tabs.Screen` as two `xcasset` icons, that an icon color doesn't change that, and a table asserting `icon.type === selectedIcon.type` across all four `iconColor` / `selectedIconColor` combinations (the crash guard).

```
Tests:       76 skipped, 4322 passed, 4398 total
🏁 All checks passed.
```

The generated API reference data was updated by hand for just this TSDoc change: running `et generate-docs-api-data --packageName expo-router` also picks up unrelated pending `standard-navigation` changes that haven't been regenerated on `main` yet, which don't belong in this PR.

Not verified on a device — I don't have the reporter's asset catalog. The repro project in #47480 covers it, and the reporter already confirmed this exact conversion works locally.

# Checklist

- [x] Documentation updated (guide + TSDoc + generated API data)
- [ ] Verified on a physical device / simulator (see above)


Ubax (changes requested):
Hi @CavalcanteLeo
Can you please split this PR into two, each fixing one issue?

Thanks

CavalcanteLeo:
Sure thing, will do it later today

CavalcanteLeo:
Thanks @Ubax — split as requested:

- #48301 — resolve `xcasset` icon names through the asset catalog so symbol sets render (fixes #47480)
- #48302 — fix the `[RNScreens] icon and selectedIcon must be same type.` crash when only one of `iconColor` / `selectedIconColor` is set

Each is independent and passes `et check-packages expo-router` on its own. Closing this one in favor of those two.

PR #48220, opened 2026-07-28
[location] Add `altitudeAboveMeanSeaLevel` property to `LocationObjectCoords`

### LINKED ISSUE #48100: [expo-location] coords.altitude reads 0 on iOS when location source is an external GPS accessory

https://github.com/Bashin/AltitudeRepro

### Steps to reproduce

1. Pair an external GPS accessory to the iOS device and confirm it is the active location source in iOS Settings.
2. Run the example app above on a physical iOS device.
3. Grant location permission.
4. Observe the on-screen values as position updates arrive.

### Environment

```text
    System:
      OS: macOS 26.5.2
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 24.10.0 - /opt/homebrew/bin/node
      Watchman: 2025.10.20.00 - /opt/homebrew/bin/watchman
    Managers:
    SDKs:
      iOS SDK:
    IDEs:
    npmPackages:
      expo: ~57.0.8 => 57.0.8 
      react: 19.2.3 => 19.2.3 
      react-dom: 19.2.3 => 19.2.3 
    npmGlobalPackages:
      eas-cli: 20.4.0
    Expo Workflow: managed
```

### Expo Doctor Diagnostics

```text
```

Wenszel:
# Why

fixes: https://github.com/expo/expo/issues/48100, follow-up to https://github.com/expo/expo/pull/41318

On some devices only `CLLocation.altitude` or only `CLLocation.ellipsoidalAltitude` is available. Therefore, we should expose both of them.

# Design decisions
There are two types of altitude available on Android and iOS: the mean sea level altitude and the WGS 84 - ellipsoidal altitude. On Android these are `Location.mslAltitudeMeters` and `Location.altitude` respectively, and on iOS `CLLocation.altitude` and `CLLocation.ellipsoidalAltitude`.

1) Expose one `altitude` property with a fallback vs two separate properties

We could simply expose a single `altitude` property which returns WGS 84 and falls back to MSL, but these two values might differ significantly (in my location it was a 50 m difference).

2) There should be one main altitude - which is more common, WGS 84 or MSL altitude?

Unfortunately, on Android the mean sea level altitude has only been available since API 34, so making it the main one would result in returning `null` on devices with earlier Android versions. That's why the default `altitude` should be the WGS 84 one, since it covers more API versions.

# How

- Expose the `altitudeAboveMeanSeaLevel` property, which is `mslAltitudeMeters` on Android and `CLLocation.altitude` on iOS
- Return `CLLocation.ellipsoidalAltitude` instead of `CLLocation.altitude` as `altitude` in the `exportLocation` function
- Fix `altitudeAccuracy` being `undefined` instead of `null` on Android API < 26
- Fix `geolocationPositionToJSON` parameter type from `LocationObject` to `GeolocationPosition`
# Test Plan

Tested in NCL on a physical device

### DIFF
--- a/packages/expo-location/src/Location.types.ts
+++ b/packages/expo-location/src/Location.types.ts
@@ -285,15 +285,27 @@ export type LocationObjectCoords = {
    */
   longitude: number;
   /**
-   * The altitude in meters above the WGS 84 reference ellipsoid. Can be `null` on Web if it's not available.
+   * The altitude in meters above the WGS 84 reference ellipsoid.
+   * On Web, it is `null` when unavailable.
+   * On iOS, it is never `null` — when the altitude is invalid, the value is `0` and `altitudeAccuracy` is negative.
    */
   altitude: number | null;
+  /**
+   * The altitude in meters above mean sea level.
+   * On Android, it is `null` when unavailable — on devices running Android 13 or lower, or when the location fix doesn't include an MSL altitude.
+   * On iOS, it is never `null` — when the altitude is invalid, the value is `0` and `altitudeAccuracy` is negative.
+   * @platform android 14+
+   * @platform ios
+   */
+  altitudeAboveMeanSeaLevel: number | null;
   /**
    * The radius of uncertainty for the location, measured in meters. Can be `null` on Web if it's not available.
    */
   accuracy: number | null;
   /**
-   * The accuracy of the altitude value, in meters. Can be `null` on Web if it's not available.
+   * The accuracy of the altitude value, in meters.
+   * On Android and Web, it is `null` when unavailable.
+   * On iOS, it is never `null` — a negative value indicates that the altitude values are invalid.
    */
   altitudeAccuracy: number | null;
   /**
--- a/packages/expo-location/CHANGELOG.md
+++ b/packages/expo-location/CHANGELOG.md
@@ -8,10 +8,14 @@
 
 ### 🎉 New features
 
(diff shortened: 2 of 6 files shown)


Wenszel:
* **#48220** <a href="https://app.graphite.com/github/pr/expo/expo/48220?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/48220?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/expo-location/src/Location.types.ts
```diff
@@ -285,15 +285,27 @@ export type LocationObjectCoords = {
...
-   * The altitude in meters above the WGS 84 reference ellipsoid. Can be `null` on Web if it's not available.
+   * The altitude in meters above the WGS 84 reference ellipsoid.
+   * On Web, it is `null` when unavailable.
+   * On iOS, it is never `null` — when the altitude is invalid, the value is `0` and `altitudeAccuracy` is negative.
    */
   altitude: number | null;
+  /**
+   * The altitude in meters above mean sea level.
+   * On Android, it is `null` when unavailable — on devices running Android 13 or lower, or when the location fix doesn't include an MSL altitude.
+   * On iOS, it is never `null` — when the altitude is invalid, the value is `0` and `altitudeAccuracy` is negative.
```
behenate, line 296: I think we should unify the platform behaviour. Is there a reason why we shouldn't make iOS value null when the altitude is invalid?

PR #48222, opened 2026-07-28
[expo-updates][iOS] Fix SIGABRT in UpdatesLogReader on truncated log lines

### LINKED ISSUE #48221: [expo-updates][iOS] SIGABRT in UpdatesLogReader: byte-length guard vs Character index causes permanent crash-on-launch

No app repository is needed — this is a deterministic logic error in `expo-updates` iOS source, and the input that triggers it is a file on disk, not app code. Repro steps below produce the crash on a bare `npx create-expo-app@latest --template blank` + `npx expo install expo-updates` project. A standalone Swift snippet isolating the discriminating fact is included, and a PR with the fix follows this issue.

### Steps to reproduce

Platform: **iOS only** (Android unaffected — see below). Environment: development build or standalone app (any build where `expo-updates` is enabled). Package manager: yarn.

1. Create a blank app, `npx expo install expo-updates`, enable updates, and make a development build for a physical device or simulator.
2. Launch the app once so `expo-updates` writes its persistent log file. The file is:

   ```
   <App Container>/Library/Application Support/dev.expo.modules.core.logging.expo-updates.txt
   ```

   (from `PersistentFileLog.init`: `"\(EXPO_MODULES_CORE_LOG_QUEUE_LABEL).\(category).txt"` in `.applicationSupportDirectory`, with `category == "expo-updates"`.)

3. Kill the app. Simulate a torn append by replacing the file's contents with a single line containing only the log-type prefix emoji — this is exactly what a process kill mid-`write` leaves behind:

   ```bash
   printf '\xf0\x9f\x9f\xa2\n' > "<App Container>/Library/Application Support/dev.expo.modules.core.logging.expo-updates.txt"
   ```

4. Launch the app.

**Expected:** the malformed line is ignored and purged.

**Actual:** the app aborts during `expo-updates` startup, before JS loads:

```
Fatal error: String index is out of bounds
Exception Type:  SIGABRT
Crash Thread/Queue: dev.expo.modules.core.logging
Blame frame: ExpoModulesCore.PersistentFileLog.purgeEntriesNotMatchingFilter(
               filter: (String) -> Bool, completion: (Error?) -> Void) -> ()
             (closure executed asynchronously)
```

5. Launch again. It crashes identically. The device is bricked until the app is reinstalled — see "Why it is unrecoverable".

### Environment

```text
  System:
    OS: macOS 26.5
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 26.5.0
    Yarn: 1.22.22
    npm: 11.17.0
  Managers:
    CocoaPods: 1.16.2
  SDKs:
    iOS SDK:
  IDEs:
    Xcode: 26.5/17F42
  npmPackages:
    expo: 57.0.6 => 57.0.6
    expo-router: ~57.0.6 => 57.0.6
    expo-updates: ~57.0.7 => 57.0.7
    react: 19.2.3 => 19.2.3
  npmGlobalPackages:
    eas-cli: 20.1.0
```

Crashes observed in production on iOS 26; ~67% of affected devices are iPads. `newArchEnabled: true`.

Verified still present in **`expo-updates@57.0.10`** (current latest) and on **`main`** as of this writing.

### Expo Doctor Diagnostics

```text
Not included: the affected project is a private commercial app, and its
expo-doctor output contains internal bundle identifiers and environment
variable names. The three failing checks in that project are unrelated to
this bug (config-plugin and dependency-version checks), and the defect is
in expo-updates source rather than in project configuration — it reproduces
on a stock `create-expo-app` blank template as described above.
```

---

### Root cause

https://github.com/expo/expo/blob/main/packages/expo-updates/ios/EXUpdates/Logging/UpdatesLogReader.swift

```swift
private func logStringToFilteredLogEntry(entryString: String, epoch: UInt) -> UpdatesLogEntry? {
  if entryString.lengthOfBytes(using: .utf8) < 2 {          // counts BYTES
    return nil
  }
  let suffixFrom = entryString.index(entryString.startIndex, offsetBy: 2)  // advances CHARACTERS
  let entryStringSuffix = String(entryString.suffix(from: suffixFrom))
  ...
}
```

The guard counts **UTF-8 bytes**; the index advance counts **Characters** (grapheme clusters). Every persisted log line is prefixed by `ExpoModulesCore.LogType.prefix` — a colored-circle emoji — in `ExpoModulesCore/Logger.swift`:

```swift
.map { "\(type.prefix) \($0)" }
```

Those prefixes are multi-byte but single-Character:

| prefix | UTF-8 bytes | Characters |
|---|---|---|
| green circle | 4 | 1 |
| white circle (with VS16) | 6 | 1 |

So a line consisting of only the prefix passes `lengthOfBytes(...) >= 2`, and `index(startIndex, offsetBy: 2)` then runs past `endIndex` and traps.

Minimal isolation of the discriminating fact (plain Swift, no Expo needed). Note this demonstrates the *guard mismatch*, not the full crash — the crash additionally requires a torn log file, per the repro steps above:

```swift
import Foundation

for s in ["\u{1F7E2}", "\u{26AA}\u{FE0F}", "\u{1F7E2} {\"a\":1}"] {
  print("bytes:", s.lengthOfBytes(using: .utf8),
        "chars:", s.count,
        "oldGuardPasses:", s.lengthOfBytes(using: .utf8) >= 2,
        "newGuardPasses:", s.count >= 2)
}
```

```
bytes: 4  chars: 1  oldGuardPasses: true  newGuardPasses: false
bytes: 6  chars: 1  oldGuardPasses: true  newGuardPasses: false
bytes: 12 chars: 9  oldGuardPasses: true  newGuardPasses: true
```

### A 1-Character entry can only come from a torn write

This is provable by elimination rather than assumed:

- Every complete persisted line is `"\(type.prefix) \($0)"` — emoji + space + payload — so a well-formed line is **always** ≥ 2 Characters.
- `PersistentFileLog.stringToList` splits on `"\n"` and filters only `!entryString.isEmpty`, so a 1-Character entry is passed straight through to the filter closure:

  ```swift
  return contents
    .components(separatedBy: "\n")
    .filter { entryString in !entryString.isEmpty }
  ```

- The only strings that pass `lengthOfBytes >= 2` while having `< 2` Characters are single multi-byte grapheme clusters.

Therefore the poison entry cannot originate from a well-formed write. `PersistentFileLog.appendTextToFile` appends via a **non-atomic** `FileHandle.seekToEndOfFile()` + `write(...)`; a process kill mid-write (backgrounding, jetsam / memory pressure — consistent with the iPad skew we observe) leaves a partial trailing line.

Corollary that explains the low frequency: the tear must land byte-aligned immediately after the emoji. A tear *inside* the UTF-8 sequence leaves invalid UTF-8, `String(contentsOfFile:encoding:.utf8)` fails to decode, and that error is **thrown** and caught by the existing `do`/`catch` — no crash. That narrow alignment window is why this shows up as a handful of events rather than constantly.

### Why it is unrecoverable

`UpdatesUtils.purgeUpdatesLogsOlderThanOneDay` is called from `EnabledAppController` on every app start. Inside `PersistentFileLog.purgeEntriesNotMatchingFilter`:

```swift
let contents = try self.readFileSync()
let newcontents = contents.filter { entry in filter(entry) }   // <- traps here
try self.writeFileSync(newcontents)                            // <- never reached
```

`writeFileSync` is never reached, so the poison line survives and the next launch crashes identically. The existing `do`/`catch` around this block cannot help — a Swift runtime trap is not a thrown `Error`.

The log file lives in Application Support and survives app-store updates, so affected users cannot wait it out; only a reinstall clears it. Shipping the fix below *does* heal already-poisoned devices, since the corrected guard rejects the line and the purge then rewrites the file clean.

### Android is unaffected

`UpdatesLogReader.kt` does no prefix stripping and parses defensively:

```kotlin
fun getTimestamp(json: String): Long? =
  runCatching { JSONObject(json).getLong("timestamp") }.getOrNull()
```

### Suggested fix

Count Characters, and use `dropFirst` so the offset can never overrun:

```diff
-    if entryString.lengthOfBytes(using: .utf8) < 2 {
+    if entryString.count < 2 {
       return nil
     }
-    let suffixFrom = entryString.index(entryString.startIndex, offsetBy: 2)
-    let entryStringSuffix = String(entryString.suffix(from: suffixFrom))
+    let entryStringSuffix = String(entryString.dropFirst(2))
```

Valid lines parse identically. I'll open a PR with this.

Optionally `appendTextToFile` could be made atomic as well, but with the above fix the purge already cleans up after a torn write.

Related but distinct: #47920 (Android `PersistentFileLog` serial-queue deadlock) — same subsystem, different failure.

valinagacevschi:
# Why

Fixes #48221 — a SIGABRT on iOS during `expo-updates` startup, seen in production on SDK 57 (iOS 26, mostly iPads). Once it happens, the app aborts on **every** launch until it is reinstalled.

`UpdatesLogReader.logStringToFilteredLogEntry` guards on **UTF-8 byte length** and then advances the string index by two **Characters**:

```swift
if entryString.lengthOfBytes(using: .utf8) < 2 {   // counts BYTES
  return nil
}
let suffixFrom = entryString.index(entryString.startIndex, offsetBy: 2)  // advances CHARACTERS
let entryStringSuffix = String(entryString.suffix(from: suffixFrom))
```

Every persisted line is written as `"\(type.prefix) \($0)"`, and `ExpoModulesCore.LogType.prefix` is a colored-circle emoji — 4–6 UTF-8 bytes but a **single Character**:

| prefix | UTF-8 bytes | Characters |
|---|---|---|
| `U+1F7E2` (green) | 4 | 1 |
| `U+26AA U+FE0F` (white) | 6 | 1 |

So a line consisting of only the prefix passes the byte guard, and `index(startIndex, offsetBy: 2)` then runs past `endIndex`:

```
Fatal error: String index is out of bounds
Exception Type:  SIGABRT
Crash Thread/Queue: dev.expo.modules.core.logging
Blame frame: ExpoModulesCore.PersistentFileLog.purgeEntriesNotMatchingFilter(
               filter: (String) -> Bool, completion: (Error?) -> Void) -> ()
```

**Where a 1-Character line comes from.** By elimination: a well-formed line is always ≥ 2 Characters (emoji + space + payload), and `PersistentFileLog.stringToList` filters only `!entryString.isEmpty`, so a 1-Character line reaches the filter untouched. `PersistentFileLog.appendTextToFile` appends via a **non-atomic** `FileHandle.seekToEndOfFile()` + `write(...)`, so a process kill mid-write (backgrounding, jetsam) leaves a partial trailing line. That the tear must land byte-aligned immediately after the emoji is why this is rare — a tear *inside* the UTF-8 sequence yields invalid UTF-8, which `String(contentsOfFile:encoding:)` reports as a **thrown** error that the existing `do`/`catch` already handles.

**Why it is unrecoverable.** `UpdatesUtils.purgeUpdatesLogsOlderThanOneDay` runs on every app start. The trap fires inside the filter closure, *before* `writeFileSync`, so the poison line is never removed and the next launch aborts identically. The `do`/`catch` around that block cannot intercept it — a Swift runtime trap is not a thrown `Error`. The log file lives in Application Support and survives App Store updates, so users cannot wait it out. Shipping this fix *does* heal already-affected devices: the corrected guard rejects the line and the purge then rewrites the file clean.

Android is unaffected — `UpdatesLogReader.kt` does no prefix stripping and parses via `runCatching`.

# How

Two lines. Count Characters instead of bytes, and use `dropFirst(2)` so the offset can never overrun a short string:

```diff
-    if entryString.lengthOfBytes(using: .utf8) < 2 {
+    if entryString.count < 2 {
       return nil
     }
-    let suffixFrom = entryString.index(entryString.startIndex, offsetBy: 2)
-    let entryStringSuffix = String(entryString.suffix(from: suffixFrom))
+    let entryStringSuffix = String(entryString.dropFirst(2))
```

Behaviour on well-formed input is identical: every valid line has ≥ 2 Characters, and dropping the first 2 Characters removes exactly the emoji + space that `Logger` prepended. Deliberately **not** included: making `appendTextToFile` atomic. It would be a larger change to `expo-modules-core`, and with this fix the purge already cleans up after a torn write.

# Test Plan

Added `PurgeSurvivesTruncatedSingleCharacterEntry` to `packages/expo-updates/ios/Tests/UpdatesLogReaderTests.swift`. It appends a raw `"\u{1F7E2}"` entry (what a torn append leaves behind), then asserts that reading skips it and that the purge completes and removes it. It uses the per-test `category` isolation added in #47082, so it does not race other suites.

⚠️ Worth knowing for review: on the **unfixed** reader this test does not fail — it **aborts the test process**, because a Swift runtime trap is not a recoverable `#expect` failure. A reviewer who checks out the parent commit to watch it go red will see a crashed test runner, not a red assertion. That is the bug.

I did not run the iOS test suite locally (no expo-updates test host set up here); CI will exercise it. The discriminating fact was verified directly with a standalone Swift snippet:

```swift
import Foundation

for s in ["\u{1F7E2}", "\u{26AA}\u{FE0F}", "\u{1F7E2} {\"a\":1}"] {
  print("bytes:", s.lengthOfBytes(using: .utf8),
        "chars:", s.count,
        "oldGuardPasses:", s.lengthOfBytes(using: .utf8) >= 2,
        "newGuardPasses:", s.count >= 2)
}
```

```
bytes: 4  chars: 1  oldGuardPasses: true  newGuardPasses: false
bytes: 6  chars: 1  oldGuardPasses: true  newGuardPasses: false
bytes: 12 chars: 9  oldGuardPasses: true  newGuardPasses: true
```

Manual reproduction on a device, from the issue: kill the app, overwrite `<App Container>/Library/Application Support/dev.expo.modules.core.logging.expo-updates.txt` with `printf '\xf0\x9f\x9f\xa2\n'`, then relaunch. Before this change the app aborts during `expo-updates` startup on every launch; after it, the line is dropped and the file is rewritten clean.

Present in `expo-updates@57.0.10` and on `main`.

# Checklist

- [ ] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: updated a module plugin). — N/A, Swift source only; no module plugin or config changes.
- [ ] Conforms with the [Documentation Writing Style Guide](https://github.com/expo/expo/blob/main/guides/Expo%20Documentation%20Writing%20Style%20Guide.md) — N/A, no docs changed.

### DIFF
--- a/packages/expo-updates/CHANGELOG.md
+++ b/packages/expo-updates/CHANGELOG.md
@@ -15,6 +15,7 @@
 - [Android] Correct `UpdatesLogReader.ONE_DAY_MILLISECONDS` from `86400` (seconds) to `86_400_000` (milliseconds), so the "older than one day" purge filter actually retains a day's worth of entries instead of ~86 seconds' worth. ([#46182](https://github.com/expo/expo/pull/46182) by [@jakequade-pc](https://github.com/jakequade-pc))
 - [iOS] Isolate UpdatesLogReaderTests from concurrent suites. ([#47082](https://github.com/expo/expo/pull/47082) by [@douglowder](https://github.com/douglowder))
 - Fix `isUpdatePending` incorrectly becoming `true` after a fetch or check that finds no new update to download. ([#47830](https://github.com/expo/expo/pull/47830) by [@kudo](https://github.com/kudo))
+- [iOS] Fix SIGABRT during log purge when the persistent log contains a truncated line: `UpdatesLogReader` guarded on UTF-8 byte length but advanced the string index by Characters, so a line holding only the multi-byte emoji log prefix trapped with "String index is out of bounds". ([#48222](https://github.com/expo/expo/pull/48222) by [@valinagacevschi](https://github.com/valinagacevschi))
 
 ### 💡 Others
 
--- a/packages/expo-updates/ios/EXUpdates/Logging/UpdatesLogReader.swift
+++ b/packages/expo-updates/ios/EXUpdates/Logging/UpdatesLogReader.swift
@@ -71,11 +71,14 @@ public final class UpdatesLogReader {
   }
 
   private func logStringToFilteredLogEntry(entryString: String, epoch: UInt) -> UpdatesLogEntry? {
-    if entryString.lengthOfBytes(using: .utf8) < 2 {
+    // NOTE: the guard must count Characters, not UTF-8 bytes. Every log line is prefixed
+    // with a multi-byte emoji (see ExpoModulesCore.LogType.prefix), so a truncated line
+    // containing only the prefix passes a byte-length guard and then traps in
+    // index(_:offsetBy:) with "String index is out of bounds" -> SIGABRT.
+    if entryString.count < 2 {
       return nil
     }
-    let suffixFrom = entryString.index(entryString.startIndex, offsetBy: 2)
-    let entryStringSuffix = String(entryString.suffix(from: suffixFrom))
+    let entryStringSuffix = String(entryString.dropFirst(2))
     let entry = UpdatesLogEntry.create(from: entryStringSuffix)
     return entry?.timestamp ?? 0 >= epoch ? entry : nil
   }
--- a/packages/expo-updates/ios/Tests/UpdatesLogReaderTests.swift
+++ b/packages/expo-updates/ios/Tests/UpdatesLogReaderTests.swift
@@ -135,12 +135,58 @@ struct UpdatesLogReaderTests {
     #expect((logEntry?.duration)! >= 300)
   }
 
+  /// Regression test for a SIGABRT during log purge.
+  ///
+  /// `logStringToFilteredLogEntry` used to guard on `lengthOfBytes(using: .utf8) < 2` and then
+  /// advance the index by two *Characters*. Every persisted line begins with `LogType.prefix` —
+  /// a colored-circle emoji that is 4–6 UTF-8 bytes but a single Character — so a line consisting
+  /// of only that prefix passed the byte guard and trapped in `index(_:offsetBy:)` with
+  /// "String index is out of bounds". Such a line is what a torn append leaves behind, since
+  /// `PersistentFileLog.appendTextToFile` writes non-atomically. The trap fired before the purge
+  /// could rewrite the file, so the bad line survived and the app aborted on every launch.
+  ///
+  /// Note: on the unfixed reader this aborts the test process (a Swift runtime trap is not a
+  /// recoverable `#expect` failure).
+  @Test
+  @MainActor
+  func `PurgeSurvivesTruncatedSingleCharacterEntry`() async throws {
+    let date1 = Date()
+
+    // U+1F7E2 LARGE GREEN CIRCLE — the value of `LogType.info.prefix`, written as an escape
+    // because `prefix` is internal to ExpoModulesCore. 4 UTF-8 bytes, 1 Character.
+    let truncatedEntry = "\u{1F7E2}"
+    await appendRawEntryAsync(entry: truncatedEntry)
+    await logErrorAsync(message: "Test message", code: .noUpdatesAvailable)
+
+    // Reading must skip the malformed entry rather than trapping.
+    let entries: [String] = logReader.getLogEntries(newerThan: date1)
+    #expect(entries.count == 1)
+
+    // The purge must run to completion and drop the malformed entry, so the next launch is clean.
+    await purgeEntriesAsync(olderThan: date1)
+
+    let remaining = PersistentFileLog(category: category).readEntries()
+    #expect(remaining.count == 1)
+    #expect(!remaining.contains(truncatedEntry))
+  }
+
   // MARK: - Private methods
 
   func clearLogAsync() async {
     await withCheckedContinuation { continuation in
       let persistentLog = PersistentFileLog(category: category)
       persistentLog.clearEntries { _ in
+        continuation.resume()
+      }
+    }
+  }
+
+  /// Appends a raw string as a log line, bypassing `UpdatesLogger`'s formatting. Used to simulate
+  /// a torn write.
+  func appendRawEntryAsync(entry: String) async {
+    await withCheckedContinuation { continuation in
+      let persistentLog = PersistentFileLog(category: category)
+      persistentLog.appendEntry(entry: entry) { _ in
         continuation.resume()
       }
     }


zoontek:
@tsapeta I successfully reproduced the issue, and validate the fix. Just letting you reviewing it too / check the comments.

tsapeta (approved):
Looks good to me, thank you for your contribution! 🙏 I added the `sdk-57` label so we will include it in the next patch release.

PR #48224, opened 2026-07-28
chore(expo-modules-test-core): Fix invalid app.plugin.js deps

kitten:
# Why

The `app.plugin.js` in `expo-modules-test-core` requires `@expo/config-plugins`, instead of `expo/config-plugins`. However, it also doesn't declare a dependency on `expo` and hence has an invalid deps chain.

# How

- Add `expo` peer dep
- Replace `@expo/config-plugins` with `expo/config-plugins`

# Test Plan

- `node app.plugin.js` (also reproduces in other flows, running `native-component-list`

# Checklist


tsapeta (approved):

gubonlucid-com (commented):
- [ ]

PR #48232, opened 2026-07-28
[config-plugins][android] Mark expo_runtime_version string as non-translatable

### LINKED ISSUE #48148: [expo-updates][Android] Google Play's automatic app-string translation rewrites expo_runtime_version, silently breaking all OTA updates for that locale
## Summary

Google Play Console's **automatic app string translation** (Gemini-powered, under "App string translation") translates `expo_runtime_version` and injects a locale-qualified copy into the App Bundle **server-side, after upload**. Android resolves string resources by locale and prefers a locale-qualified value over the default, so every device in that language reads a **stale** runtime version and stops receiving updates.

It fails completely silently. The update check succeeds, and the server correctly answers `NoUpdatesAvailable`, because the app is genuinely asking for a runtime version that was never published.

## Impact

Any Expo app using EAS Update, shipping to Google Play as an AAB, with Play's app-string translation enabled. There is no error, no failed download, and nothing in crash reporting. The only visible symptom is "OTA updates never arrive on Android", and only for users in the translated locale, which makes it look intermittent or device-specific.

In our case OTA delivery was dead on Greek-locale Android devices for roughly seven weeks before anyone worked out why. iOS was unaffected throughout.

## Evidence

App version at time of testing: `1.4.0`. Runtime policy: `{ "policy": "appVersion" }`.

**1. The AAB produced by EAS Build is correct.** Across every `resources.pb` in the bundle:

```
expo_runtime_version = 1.4.0
occurrences of "1.2.1" = 0
```

**2. The APK installed by Play is not.** Pulled from the device with `adb pull`:

```
base.apk                 expo_runtime_version = 1.4.0     correct
split_config.el.apk      expo_runtime_version = 1.2.1     STALE
```

`1.2.1` was the app version about two months earlier, presumably when translations were first generated. That value **exists nowhere in the uploaded bundle**.

**3. Locale is the only variable.** Same device, same `versionCode`, same Play account:

- Device language Greek, reinstall: `Updates.runtimeVersion` reports `1.2.1`, no updates ever arrive
- Device language English, reinstall: `Updates.runtimeVersion` reports `1.4.0`, updates arrive normally
- Back to Greek, reinstall: `1.2.1` again

**4. The update log shows a clean, successful check** (`Updates.readLogEntriesAsync()`):

```
Updates state change: StartStartup
Updates state change: Check                      (isChecking=true)
Updates state change: CheckCompleteUnavailable
[info] NoUpdatesAvailable
UpdatesController onBackgroundUpdateFinished: No update available
Updates state change: EndStartup
```

No error, no rejected download. The server is answering honestly; the client is asking the wrong question.

## Root cause

`expo-updates` stores the runtime version as an ordinary Android **string resource**, referenced from the manifest:

```xml
<meta-data android:name="expo.modules.updates.EXPO_RUNTIME_VERSION"
           android:value="@string/expo_runtime_version" />
```

and writes it in `@expo/config-plugins/build/android/Updates.js`:

```js
return setStringItem([
  buildResourceItem({
    name: 'expo_runtime_version',
    value: runtimeVersion,
  }),
], stringsJSON);
```

Nothing marks it as non-translatable, so Play's translator treats it as user-facing copy and localises it like any other string.

## Proposed fix

Android supports `translatable="false"` to exclude a string from translation tooling, and `buildResourceItem` **already accepts a `translatable` option** (`@expo/config-plugins/build/android/Resources.js`), it simply is not used here:

```js
buildResourceItem({
  name: 'expo_runtime_version',
  value: runtimeVersion,
  translatable: false,   // <-- emits <string ... translatable="false">
})
```

The same likely applies to any other technical string resource the plugins emit.

I have not verified that Play's translator honours `translatable="false"`, so that is worth confirming, but it is the standard Android signal for exactly this and costs nothing.

## Confirmed fixed by disabling the Play feature

Disabled app-string translation in Play Console, rebuilt (no code change at all), uploaded, and pulled the new install off the same device:

```
split_config.el.apk       expo_runtime_version   "1.2.1"
--------------------------------------------------------
versionCode 24 (on)               1                  1
versionCode 25 (off)              0                  0
```

The Greek split no longer defines `expo_runtime_version` at all, so Android falls through to the base APK's correct value. `Updates.runtimeVersion` now reports `1.4.0` on a Greek-locale device and updates are delivered normally. Same code, same runtime policy, same EAS build profile; the only change was the Play Console setting.

## Workaround for anyone hitting this now

1. Disable app-string translation in Play Console. **This does not clean already-published bundles**, their translations persist.
2. Upload a **new** bundle so Play regenerates the split APKs without the injected value.
3. Verify with `adb pull` that the locale split no longer overrides `expo_runtime_version`. Do not trust the AAB, it was always correct.

## Possibly the same as #41694

[#41694](https://github.com/expo/expo/issues/41694) reports "runtimeVersion not syncing to AAB builds after SDK 54 upgrade (APK works, AAB doesn't)" and was closed as *missing or invalid repro*. The symptom matches exactly, including APK being fine while AAB is not, which follows naturally if the corruption happens during Play's processing of the bundle rather than in the build.

That reporter's only working workaround was injecting `expo_runtime_version` through the `locales` config. That is consistent with this diagnosis: it forces a locale-qualified value into the bundle which overwrites the translated one.

## Environment

- `expo` SDK 55, `expo-updates` `~55.0.22`
- EAS Build, AAB, distributed via Google Play production track
- `runtimeVersion: { "policy": "appVersion" }`
- Reproduced on two physical devices (Samsung SM-A715F among them) plus a third, all Greek locale

ahmdshrif:
# Why

> **Update:** this PR no longer claims to fix #48148. The `@expo-bot verify` run measured that `translatable="false"` is stripped during resource compilation and never reaches the `.aab` that Google Play receives, so it cannot affect Play's server-side translation. The change is still correct as source-XML hygiene, and the body below has been rewritten to say only what is supported. See the discussion for the measurements and for the escaping trade-off.

`@expo/config-plugins` writes the Android runtime version as an ordinary string resource:

```xml
<string name="expo_runtime_version">1.4.0</string>
```

referenced from the manifest as `@string/expo_runtime_version`. Nothing marks it as non-translatable, so tooling that reads `res/values/strings.xml` — Android Lint, the Android Studio translations editor, and translation pipelines that consume the source XML — treats a technical identifier as user-facing copy.

`expo_runtime_version` is read back by `expo-updates` and must match the value the update server knows about. A translated copy of it is never correct.

# How

Pass `translatable: false` when building the resource item, so the plugin emits:

```xml
<string name="expo_runtime_version" translatable="false">1.4.0</string>
```

`translatable="false"` is the standard Android signal for "this string is a technical identifier, not copy". `buildResourceItem` already supported the option (`Resources.ts`); it just was not used here.

Scope notes:

- **The attribute is a source-XML signal only.** It is not carried into the compiled resource table: `aapt2` produces a byte-identical `resources.pb` with and without it, and `Resources.proto` has no field for it. So this does not, and cannot, change what Google Play receives. Anyone hit by #48148 needs to turn off automatic app string translation in Play Console and upload a new release; that issue needs its own fix and stays open.
- **Trade-off worth a maintainer's call:** `@expo/config-plugins` skips Android string escaping for resources carrying `translatable="false"` (`utils/XML.ts`, both the read and write paths). After this change a `runtimeVersion` containing `"`, a tab, or leading/trailing whitespace is written unescaped, and compiles to a different value with no error — e.g. `1.4"0` resolves to `1.40`, and a trailing space is dropped. A value containing `'` becomes a loud `aapt2` build error instead. Those values are pathological for a runtime version, but the change is not free. If you would rather not accept it, the alternative is to stop skipping escaping for non-translatable strings in `XML.ts`, which is a wider change affecting every plugin and belongs in its own PR — happy to open it.
- `expo_runtime_version` is the only technical string resource written by these plugins. The other `setStringItem` caller is `Name.ts` writing `app_name`, which *is* user-facing and stays translatable.
- `setStringItem` replaces a matching entry wholesale, so a `strings.xml` written by an older plugin version gets the attribute added on the next prebuild rather than keeping a stale attribute-less entry. There is a regression test for this.
- The `@expo/prebuild-config` snapshot is updated to match the new resource output; that is the only other place in the repo that asserts on this string.
- The `expo-updates` plugin has copies in `eas-cli` and `build-tools` (per the note at the top of `Updates.ts`). Those should get the same one-line change; this PR only covers the `expo/expo` copy.

# Test Plan

`jest src/android/__tests__/Updates-test.ts` → **5 passed**.

Reverting only the `translatable: false` line, keeping the tests:

```
✕ Write and clear runtime version in strings resource
✕ Marks the runtime version string non-translatable
    Expected: "false"
    Received: undefined
✕ Adds translatable="false" to a runtime version string written by an older plugin version
```

Full `@expo/config-plugins` suite: **394 passed, 2 skipped, 16 snapshots passed**, no test failures. (12 suites fail to *load* in my checkout with `Cannot find module '@expo/json-file'` — stale workspace `node_modules` on my side, in files this PR does not touch.)

Rebased onto current `main`.


brentvatne:
@expo-bot verify

ahmdshrif:
Thanks for the verify run — the `Fixes #48148` claim does not hold and I've corrected the PR rather than defend it. Pushed as ce31784, and the branch is rebased onto current `main` so the conflict is gone.

**What changed:**

- Dropped `Fixes #48148` from the PR body, and rewrote the body to state only what is supported: the attribute is a source-XML signal that Lint and translation pipelines read, and it is **not** carried into the compiled resource table, so it cannot affect what Play receives. #48148 stays open and needs its own fix.
- Rewrote the changelog entry the same way — it no longer mentions Google Play at all:
  > `[Android] Mark the expo_runtime_version string resource as translatable="false", so localization tooling and Android Lint treat it as the technical identifier it is rather than user-facing copy.`
- Rewrote the source comment, which had carried the same claim, and noted the source-XML-only scope in place of it.
- Added your escaping finding to the body as an explicit trade-off for a maintainer to accept or reject, with the `1.4"0` → `1.40` and trailing-space cases named.

**On the escaping cost**, I did not try to paper over it. `utils/XML.ts` skips `escapeAndroidString`/`unescapeAndroidString` for any resource carrying `translatable="false"`, on both the read and write paths, so the measurement is a property of that shared helper and not of this call site. The narrow fix would be to stop skipping escaping for non-translatable strings, but that changes behaviour for every plugin that writes one and deserves its own PR and its own review. I would rather you decide which way that goes than smuggle it in here. Say the word and I will open it.

If you would prefer this PR not to land at all now that the Play rationale is gone, that is a reasonable call and I won't argue it — the remaining benefit is Lint and translation-pipeline hygiene, against the escaping trade-off above.

Test plan unchanged in substance: `Updates-test.ts` is **5 passed**, and reverting only the `translatable: false` line fails 3 of them (`Expected: "false" / Received: undefined`), so the tests still guard the behaviour after the rebase.

PR #48233, opened 2026-07-28
[docs] Update `eas/checkout` docs by informing about the `ref` input

sswrk:
# Why

We've added a `ref` input to `eas/checkout`: https://github.com/expo/eas-cli/pull/4035

# How

Updated the EAS workflows sytnax and custom builds docs.

# Test Plan

Verified it renders correctly:

Screenshot 2026-07-28 at 16 39 24

Screenshot 2026-07-28 at 16 39 43

# Checklist

### DIFF
--- a/docs/pages/custom-builds/schema.mdx
+++ b/docs/pages/custom-builds/schema.mdx
@@ -930,7 +930,7 @@ build:
     # @end #
 ```
 
-The `ref` input only works when your project sources come from a Git repository. Place the step before [`eas/build`](#easbuild), which checks out the project internally.
+The `ref` input only works when your project sources come from a Git repository, for example, a build triggered through the GitHub integration. Local builds and uploaded project tarballs do not support it. A branch takes a bare name or a qualified ref such as `refs/heads/feature/add-icon`. A tag takes the qualified form, such as `refs/tags/v1.2.3`. A commit takes a full SHA. Place the step before [`eas/build`](#easbuild), which checks out the project internally.
 
 <EASFunctionDescription name="checkout" />
 
--- a/docs/pages/custom-builds/schema.mdx
+++ b/docs/pages/custom-builds/schema.mdx
@@ -919,6 +919,8 @@ build:
         run: ls assets
 ```
 
+For builds with Git-based project sources, the step uses the build's recorded commit by default. Use `ref` to check out a different branch, tag, or commit:
+
 ```yaml example.yml
 build:
   name: Check out a specific ref
@@ -928,9 +930,16 @@ build:
         inputs:
           ref: feature/add-icon
     # @end #
+    - eas/build
 ```
 
-The `ref` input only works when your project sources come from a Git repository, for example, a build triggered through the GitHub integration. Local builds and uploaded project tarballs do not support it. A branch takes a bare name or a qualified ref such as `refs/heads/feature/add-icon`. A tag takes the qualified form, such as `refs/tags/v1.2.3`. A commit takes a full SHA. Place the step before [`eas/build`](#easbuild), which checks out the project internally.
+`ref` accepts:
+
+- A branch, as a bare name such as `feature/add-icon` or a qualified ref such as `refs/heads/feature/add-icon`. The repository ends up on that branch.
+- A tag, as a qualified ref such as `refs/tags/v1.2.3`. The repository ends up on a detached `HEAD`.
+- A full commit SHA. The repository ends up on a detached `HEAD`.
+
+The `ref` input only works when your project sources come from a Git repository, for example, a build triggered through the GitHub integration. Local builds and uploaded project tarballs do not support it. Place the step before [`eas/build`](#easbuild), which checks out the project internally.
 
 <EASFunctionDescription name="checkout" />
 
--- a/docs/pages/custom-builds/schema.mdx
+++ b/docs/pages/custom-builds/schema.mdx
@@ -930,7 +930,7 @@ build:
     # @end #
 ```
 
-The `ref` input requires Git-based project sources and must be set before [`eas/build`](#easbuild), which checks out the project internally.
+The `ref` input only works when your project sources come from a Git repository. Place the step before [`eas/build`](#easbuild), which checks out the project internally.
 
 <EASFunctionDescription name="checkout" />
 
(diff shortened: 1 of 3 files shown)


docs/pages/eas/workflows/syntax.mdx
```diff
@@ -1840,6 +1840,23 @@ jobs:
...
+  my_job:
+    steps:
+      # @info #
+      - uses: eas/checkout
+        with:
+          ref: feature/add-icon
+        # @end #
+```
+
+Branches accept a bare name or a qualified ref such as `refs/heads/feature/add-icon`. Tags need the qualified form, such as `refs/tags/v1.2.3`. A full commit SHA also works and leaves the repository on a detached `HEAD`. The step performs a shallow, depth-one fetch from the source repository's `origin`, so the ref must be reachable there.
```
amandeepmittal: ⚠️ **correctness** `review-ai-text.md#t7-overclaiming`

A tag also ends on a detached `HEAD`. Only a branch ref checks out onto a named branch, so "A full commit SHA also works and leaves the repository on a detached `HEAD`" reads as if detached `HEAD` were specific to SHAs. Splitting the three ref shapes into a list makes the per-shape behavior land.

```suggestion
`ref` accepts:

- A branch, as a bare name such as `feature/add-icon` or a qualified ref such as `refs/heads/feature/add-icon`. The repository ends up on that branch.
- A tag, as a qualified ref such as `refs/tags/v1.2.3`. The repository ends up on a detached `HEAD`.
- A full commit SHA. The repository ends up on a detached `HEAD`.

The step runs a shallow fetch (`--depth 1`) from the source repository's `origin`, so the ref must be reachable there.
```

docs/pages/custom-builds/schema.mdx
```diff
@@ -919,6 +919,19 @@ build:
...
+  name: Check out a specific ref
+  steps:
+    # @info #
+    - eas/checkout:
+        inputs:
+          ref: feature/add-icon
+    # @end #
+```
+
+The `ref` input only works when your project sources come from a Git repository. Place the step before [`eas/build`](#easbuild), which checks out the project internally.
```
amandeepmittal: ⚠️ **correctness** `prose.md#core-principles`

**eas/workflows/syntax.mdx** adds "Local builds and uploaded project tarballs do not support it.", and that is the case most readers of this page land in, since `eas build` uploads the project as a tarball by default. Without it, "only works when your project sources come from a Git repository" gives a reader here no way to tell whether it applies to them. This page also drops the ref formats, so nothing says a tag needs `refs/tags/v1.2.3`.

```suggestion
The `ref` input only works when your project sources come from a Git repository, for example, a build triggered through the GitHub integration. Local builds and uploaded project tarballs do not support it. A branch takes a bare name or a qualified ref such as `refs/heads/feature/add-icon`. A tag takes the qualified form, such as `refs/tags/v1.2.3`. A commit takes a full SHA. Place the step before [`eas/build`](#easbuild), which checks out the project internally.
```

docs/pages/eas/workflows/syntax.mdx
```diff
@@ -1840,6 +1840,23 @@ jobs:
...
+
+`ref` only works when your project sources come from a Git repository, for example a job triggered through the GitHub integration. Local builds and uploaded project tarballs do not support it. Place the step before any other step that checks out the project. Setting `ref` after checkout fails the step.
```
amandeepmittal: ✏️ **style** `prose.md#core-principles`

In a workflow job, `eas/checkout` is the only step that checks out the project, so "Place the step before any other step that checks out the project" points at nothing a reader can act on. The rule worth stating is that `ref` has to go on the job's first `eas/checkout`.

```suggestion
`ref` only works when your project sources come from a Git repository, for example, a job triggered through the GitHub integration. Local builds and uploaded project tarballs do not support it. Set `ref` on the job's first `eas/checkout` step. Setting it on a later step fails, because the project is already checked out.
```

docs/pages/custom-builds/schema.mdx
```diff
@@ -919,6 +919,19 @@ build:
         run: ls assets
 ```
 
+```yaml example.yml
```
amandeepmittal, line 924: ✏️ **style** `prose.md#core-principles`

This block lands straight after the previous one with nothing between them, while the workflows page introduces the same snippet with "By default, the step checks out the ref that triggered the job. Use `ref` to check out a different branch, tag, or commit:". The snippet also leaves out `eas/build`, so it never shows the ordering the sentence below it asks for.

docs/scenes/eas-functions/_checkout.mdx
```diff
@@ -1,6 +1,11 @@
 import { GithubIcon } from '@expo/styleguide-icons/custom/GithubIcon';
 
 import { BoxLink } from '~/ui/components/BoxLink';
+import { NoIcon } from '~/ui/components/DocIcons';
+
+| Property | Type     | Required   | Description                                                                                                                                                                                                           |
+| -------- | -------- | ---------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `ref`    | `string` | <NoIcon /> | Git branch, tag, or full commit SHA to check out instead of the ref that triggered the job. Requires project sources from a Git repository, and this step must run before any other step that checks out the project. |
```
amandeepmittal: ✏️ **style** `review-ai-text.md#t5-restatement`

The second sentence repeats the paragraph that renders directly above `<EASFunctionDescription name="checkout" />` on both pages using this partial. Keep the cell to what the input is plus its default, and say "build or workflow job" since this partial also renders on the custom builds page.

```suggestion
| Property | Type     | Required   | Description                                                                                                     |
| -------- | -------- | ---------- | --------------------------------------------------------------------------------------------------------------- |
| `ref`    | `string` | <NoIcon /> | Git branch, tag, or full commit SHA to check out. Defaults to the ref that triggered the build or workflow job. |
```

amandeepmittal (changes requested):

docs/pages/eas/workflows/syntax.mdx
sswrk: Addressed in https://github.com/expo/expo/pull/48233/commits/ce3714f51540686b3551e9aaa91c85741e6e3f24

docs/pages/custom-builds/schema.mdx
sswrk: Addressed in https://github.com/expo/expo/pull/48233/commits/ce3714f51540686b3551e9aaa91c85741e6e3f24

amandeepmittal (approved):

PR #48236, opened 2026-07-28
refactor(cli): Give the dev server port a single source

ramonclaudio:
# Why

Follow-up to #47771.

The dev server port came from three places, `resolvePortAsync` picked it, `UrlCreator` kept its own copy from init, and Metro read it back off the socket after binding.

# How

I put the port on `BundlerDevServer` and everything reads it back with `getPort()`. The two env vars get passed into `UrlCreator` now instead of it reading them itself.

I also had to resolve the web port even when web isn't starting, otherwise there's no port to use when you open it later.

Two things change, a `REACT_NATIVE_PACKAGER_HOSTNAME` that's only whitespace used to stop the start, now it's ignored. And on a project using the `webpack` web bundler, pressing `w` used to quit with an error if you hadn't started with `--web`.

# Test Plan

build, typecheck, depscheck, lint, format, test.

### DIFF
--- a/packages/@expo/cli/src/start/server/BundlerDevServer.ts
+++ b/packages/@expo/cli/src/start/server/BundlerDevServer.ts
@@ -428,20 +428,14 @@ export abstract class BundlerDevServer {
   ) {
     assert(options?.port, 'Dev server instance not found');
     assert(!this.urlCreator, 'Dev server is already initialized');
+    this.resolvedPort = options.port;
     // TODO: Drop the undocumented REACT_NATIVE_PACKAGER_HOSTNAME
-    const hostnameOverride = env.REACT_NATIVE_PACKAGER_HOSTNAME;
-    if (hostnameOverride && !hostnameOverride.trim()) {
-      Log.warn(
-        `Ignoring REACT_NATIVE_PACKAGER_HOSTNAME because it is only whitespace. Set it to a hostname, or unset it to use the default host.`
-      );
-    }
     const urlCreator = await UrlCreator.init(options.location, {
       getPort: () => this.getPort(),
       getTunnelUrl: this.getTunnelUrl.bind(this),
       getHostnameOverride: () => env.REACT_NATIVE_PACKAGER_HOSTNAME,
       getProxyUrl: () => env.EXPO_PACKAGER_PROXY_URL,
     });
-    this.resolvedPort = options.port;
     this.urlCreator = urlCreator;
     return urlCreator;
   }
--- a/packages/@expo/cli/src/start/server/BundlerDevServer.ts
+++ b/packages/@expo/cli/src/start/server/BundlerDevServer.ts
@@ -129,6 +129,8 @@ export abstract class BundlerDevServer {
     {};
   /** Manages the creation of dev server URLs. */
   protected urlCreator?: UrlCreator | null = null;
+  /** The resolved port every URL and `instance.location` is built from. */
+  private resolvedPort: number | null = null;
 
   private notifier: FileNotifier | null = null;
   protected readonly devToolsPluginManager: DevToolsPluginManager;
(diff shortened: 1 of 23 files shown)


packages/@expo/cli/src/utils/env.ts
```diff
@@ -111,6 +111,11 @@ class Env {
     return string('EXPO_EDITOR', '');
   }
 
+  /** Overwrite the hostname used in dev server URLs, disregarding the `--host`, `--lan` and `--localhost` arguments. */
+  get REACT_NATIVE_PACKAGER_HOSTNAME(): string {
+    return string('REACT_NATIVE_PACKAGER_HOSTNAME', '');
+  }
```
kitten:
```suggestion
  get REACT_NATIVE_PACKAGER_HOSTNAME(): string | null {
    return string('REACT_NATIVE_PACKAGER_HOSTNAME', '')?.trim() || null;
  }
```

We have a `.trim()` somewhere, which can be absorbed here

packages/@expo/cli/src/start/server/BundlerDevServer.ts
```diff
@@ -425,10 +428,20 @@ export abstract class BundlerDevServer {
...
+      );
+    }
     const urlCreator = await UrlCreator.init(options.location, {
-      port: options.port,
+      getPort: () => this.getPort(),
       getTunnelUrl: this.getTunnelUrl.bind(this),
+      getHostnameOverride: () => env.REACT_NATIVE_PACKAGER_HOSTNAME,
+      getProxyUrl: () => env.EXPO_PACKAGER_PROXY_URL,
     });
+    this.resolvedPort = options.port;
```
kitten: move to top above `await` to fix `getPort` asynchronicity, maybe below the `assert` (line 429) for parity?

packages/@expo/cli/src/utils/port.ts
```diff
@@ -144,25 +144,25 @@ export async function choosePortAsync(
 }
 
 // TODO(Bacon): Revisit after all start and run code is merged.
+/** Picks a port without reading the environment. `resolveMetroPortAsync` is the entry point every command uses. */
 export async function resolvePortAsync(
```
kitten: nit: rename to `_resolvePortAsync` (for example) so we know this is private. We don't often do this but should do this more often, because otherwise tests become confusing.

kitten:
automated review found that `DevServerManager`'s `ensureWebDevServerRunningAsync` doesn't have a way of acquiring a port now. That's because it's lazy which likely was done because Webpack is the deprecated path: https://github.com/expo/expo/blob/3c9d5ece42ac4b98bd8fcc95f44f15df062adf63/packages/%40expo/cli/src/start/startAsync.ts#L48-L52

We can probably keep it lazy but resolve `webpackPort` eagerly even when Webpack isn't started eagerly, by passing the unfiltered bundler list to `resolvePortsAsync` and then keeping the port on `DevServerManager`. If we then hand it to `ensureWebDevServerRunningAsync` wen can ensure that we can consistently use it and this doesn't crash.

packages/@expo/cli/src/start/server/BundlerDevServer.ts
```diff
@@ -425,10 +428,20 @@ export abstract class BundlerDevServer {
   ) {
     assert(options?.port, 'Dev server instance not found');
     assert(!this.urlCreator, 'Dev server is already initialized');
+    // TODO: Drop the undocumented REACT_NATIVE_PACKAGER_HOSTNAME
+    const hostnameOverride = env.REACT_NATIVE_PACKAGER_HOSTNAME;
+    if (hostnameOverride && !hostnameOverride.trim()) {
+      Log.warn(
```
kitten: I think we don't need a warning here and can silently ignore this

kitten (approved):

PR #48238, opened 2026-07-28
Fix/hmr native reload

chrisnojima:
# Why

`registerBundleEntryPoints` calls `window.location.reload()` when `hmrUnavailableReason != null`:

https://github.com/expo/expo/blob/main/packages/expo/src/async-require/hmr.ts#L325-L329

`window.location` is undefined on Android and iOS, so on native this throws instead of reloading:

```
TypeError: Cannot read property 'reload' of undefined
```

Repro (development builds only):

1. Run an app that loads an async bundle (`asyncRoutes`, or any `import()`-split route).
2. Kill the Metro server, or let the socket drop / go out of sync so `setHMRUnavailableReason` runs.
3. Navigate to a route that pulls in an async chunk.

Instead of the intended "Metro disconnected, reload the app" recovery, you get a red box.

# How

`./hmrUtils` already has a platform-split `reload()` — `hmrUtils.ts` uses `window.location.reload()`
and `hmrUtils.native.ts` uses `DevSettings.reload()`. Every other platform-specific operation in
`hmr.ts` goes through that module; this one call site bypassed it. Import `reload` and call it.

No behavior change on web: the web implementation of `reload()` is `window.location.reload()`.

# Test Plan

- Web: HMR unavailable path still reloads the tab (unchanged code path).
- iOS + Android dev build: kill Metro, navigate to an async route. Before: red box
  `Cannot read property 'reload' of undefined`. After: the app reloads via `DevSettings.reload()`.

# Checklist

- [x] This diff will work correctly for `npx expo prebuild` & EAS Build.

### DIFF
--- a/packages/expo/CHANGELOG.md
+++ b/packages/expo/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ### 🐛 Bug fixes
 
-- Fix dev-only crash `Cannot read property 'reload' of undefined` on Android and iOS when an async bundle loads after Metro disconnects, by using the platform-split `reload()` helper instead of `window.location.reload()`. ([#00000](https://github.com/expo/expo/pull/00000) by [@chrisnojima](https://github.com/chrisnojima))
+- Fix dev-only crash `Cannot read property 'reload' of undefined` on Android and iOS when an async bundle loads after Metro disconnects, by using the platform-split `reload()` helper instead of `window.location.reload()`. ([#48238](https://github.com/expo/expo/pull/48238) by [@chrisnojima](https://github.com/chrisnojima))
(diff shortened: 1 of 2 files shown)


kitten:
`asyncRoutes` implies `expo-router`, which in turn implies/requires and installs `@expo/metro-runtime`. That in turn installs this: https://github.com/expo/expo/blob/9696ad94807a761c73edf71629149882edd0e49d/packages/%40expo/metro-runtime/src/location/Location.native.ts#L126-L148

Meaning, `window.location.reload()` is present. Please don't use PRs as issue reports. If you have an issue or a reproducible example of the `@expo/metro-runtime` initialization failing, please submit a reproduction in an issue, which would help us out a lot more.

For now, since there's no evidenced path to this issue, this change isn't quite correct and mergeable.

Cheers!

chrisnojima:
Sure i'll make an issue. I actually don't use expo-router

chrisnojima:
Linked to https://github.com/expo/expo/pull/48944

PR #48241, opened 2026-07-28
[expo-updates] Stop stacking native state change listeners on module re-evaluation

### LINKED ISSUE #47938: expo-router + expo-updates dev server memory leak OOM crash

https://github.com/bradleyayers/expo-updates-memory-leak

### Steps to reproduce

1. Target Web using expo-router and `output: server`.
2. Add an JS import to `expo-updates` somewhere in your in `src/app` or imported from there (i.e. somewhere expo-router will crawl transitively)
3. Start the dev server `npm run web`, load the app in the browser, the server will start leaking memory.

The dev server will get progressively slower and slower as it uses up more heap memory and thrashes garbage collection, until it eventually crashes with OOM. The speed that this happens depends on how big your app is, as each HTTP request served by the dev server will add an extra copy in memory.

In the repro I created I added logging to make it easy to see:

```
λ  LOG  expo-updates listener size: 3
Web Bundled 143ms node_modules/expo-router/entry.js (1 module)
Web Bundled 20ms node_modules/expo-router/entry.js (1 module)
λ Bundled 12ms node_modules/expo/node_modules/@expo/cli/node_modules/@expo/router-server/node/render.js (1 module)
λ  LOG  expo-updates listener size: 4
Web Bundled 151ms node_modules/expo-router/entry.js (1 module)
Web  LOG  Running application "main" with appParams:
 {"hydrate": undefined, "rootTag": "#root"} 
Development-level warnings: ON.
Performance optimizations: OFF.
Web  LOG  expo-updates listener size: 1
Web  WARN  Image: style.resizeMode is deprecated. Please use props.resizeMode.
λ Bundled 18ms node_modules/expo/node_modules/@expo/cli/node_modules/@expo/router-server/node/render.js (1 module)
λ  LOG  expo-updates listener size: 5
Web Bundled 146ms node_modules/expo-router/entry.js (1 module)
Web Bundled 57ms node_modules/expo-router/entry.js (1 module)
λ Bundled 11ms node_modules/expo/node_modules/@expo/cli/node_modules/@expo/router-server/node/render.js (1 module)
λ  LOG  expo-updates listener size: 6
```

The memory leak is caused by https://github.com/expo/expo/blob/4e87c54c61d9ba3b3223b67959fe04def6ec082d/packages/expo-updates/src/UpdatesEmitter.ts#L9 rerunning in the server each time a request is handled.

I ran into this because I was adding expo-update data into Sentry, e.g.

```ts
import * as Sentry from "@sentry/react-native";
import * as Updates from "expo-updates";

const scope = Sentry.getCurrentScope();
scope.setTag(`expo-update-id`, Updates.updateId);
```

It's been plaguing me for a year and I eventually dug into it to get to the cause today.

### Environment

```text
    System:
      OS: macOS 26.5.2
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 22.13.1 - /Users/brad/.proto/tools/node/22.13.1/bin/node
      Yarn: 1.22.21 - /Users/brad/.proto/tools/node/22.13.1/bin/yarn
      npm: 10.9.2 - /Users/brad/.proto/tools/node/22.13.1/bin/npm
    Managers:
    SDKs:
      iOS SDK:
    IDEs:
    npmPackages:
      expo: ~57.0.7 => 57.0.7 
      expo-router: ~57.0.7 => 57.0.7 
      expo-updates: ~57.0.8 => 57.0.8 
      react: 19.2.3 => 19.2.3 
      react-dom: 19.2.3 => 19.2.3 
    Expo Workflow: managed
```

### Expo Doctor Diagnostics

```text
```

ahmdshrif:
# Why

Fixes #47938.

`UpdatesEmitter.ts` registers a native state change listener as an import-time side effect:

```ts
ExpoUpdatesModule.addListener('Expo.nativeUpdatesStateChangeEvent', _handleNativeStateChangeEvent);
```

`ExpoUpdatesModule` is a singleton stored on the global object — `registerWebModule` returns the cached instance from `globalThis.expo.modules`, and `requireNativeModule` likewise hands back one instance per runtime. The listener therefore outlives the module scope that registered it.

Server rendering recreates the module graph for every request against that same global. Each request evaluated `UpdatesEmitter` again and stacked another listener, and because `_handleNativeStateChangeEvent` closes over that evaluation's `_updatesStateChangeListeners` set and `latestContext`, every stacked listener pinned an entire copy of the module graph. The dev server's heap grew per request until it thrashed GC and crashed with OOM. Fast Refresh has the same shape on a smaller scale.

The reporter's log shows the listener count climbing one per rendered request:

```
λ  LOG  expo-updates listener size: 4
λ  LOG  expo-updates listener size: 5
λ  LOG  expo-updates listener size: 6
```

# How

Drop the previous listener before registering a new one, so exactly one subscription is live at any time:

```ts
ExpoUpdatesModule.removeAllListeners('Expo.nativeUpdatesStateChangeEvent');
ExpoUpdatesModule.addListener('Expo.nativeUpdatesStateChangeEvent', _handleNativeStateChangeEvent);
```

`removeAllListeners` is a no-op when the count is already zero, so the first evaluation in a runtime — the only one that happens on device outside Fast Refresh — behaves exactly as before, and `OnStopObserving` is not triggered spuriously. On a re-evaluation the module ends up owning the single live listener, which is the desired outcome: the older module scope is dead and its listeners should not keep receiving events.

Alternatives considered and rejected:

- **Subscribe lazily / reference-count on `addUpdatesStateChangeListener`.** This would leave `latestContext` un-updated whenever no `useUpdates()` hook is mounted, so a later mount would read a stale sequence number as its initial state. That is a behavior change to the state machine, not a leak fix.
- **Guard the registration with a "is this the server?" check.** Narrower than the actual bug — it leaves the Fast Refresh case in place and encodes an environment assumption into a module that otherwise has none.

# Test Plan

Added `packages/expo-updates/src/__tests__/UpdatesEmitter-test.node.ts`, which evaluates `UpdatesEmitter` six times across `jest.resetModules()` and asserts the native listener count stays at 1.

`src/__mocks__/ExpoUpdates.ts` previously stubbed `addListener` with a bare `jest.fn()`, which could not observe this at all. It now keeps a real listener registry and implements `removeAllListeners` / `listenerCount`. The registry is stored on `globalThis` so that it survives `jest.resetModules()` — mirroring how the real singleton survives module graph re-evaluation. No existing test asserts against the mock's `addListener`.

Before/after, run against the changed files:

```
# with the fix
PASS src/__tests__/UpdatesEmitter-test.node.ts
  ✓ keeps a single native listener when the module graph is re-evaluated

# with the removeAllListeners line reverted
● keeps a single native listener when the module graph is re-evaluated
  expect(received).toBe(expected)
  Expected: 1
  Received: 6
```

# Checklist

- [x] Documentation is up to date to reflect any changes in this PR.

### DIFF
--- a/packages/expo-updates/CHANGELOG.md
+++ b/packages/expo-updates/CHANGELOG.md
@@ -35,6 +35,7 @@
 - [Android] Skip and repair updates that are missing their launch asset instead of selecting them for launch, which previously failed every cold start with "Launch asset not found for update". ([#49470](https://github.com/expo/expo/pull/49470) by [@alanjhughes](https://github.com/alanjhughes), based on [#48733](https://github.com/expo/expo/pull/48733) by [@martintreurnicht](https://github.com/martintreurnicht))
 - [iOS] Adopt the existing asset row when registering a new asset whose key is already in the database, instead of replacing it, which cascade-deleted every update referencing that asset. ([#49504](https://github.com/expo/expo/pull/49504) by [@alanjhughes](https://github.com/alanjhughes))
 - [Android] Adopt the existing asset row when registering a new asset whose key is already in the database, instead of replacing it, which cascade-deleted every update referencing that asset. ([#49505](https://github.com/expo/expo/pull/49505) by [@alanjhughes](https://github.com/alanjhughes))
+- Drop the previous native state change listener before registering a new one, so re-evaluating `UpdatesEmitter` no longer stacks a subscription that retains the whole module scope — a leak that grew per request under server rendering until the dev server crashed with OOM. ([#47938](https://github.com/expo/expo/issues/47938) by [@ahmdshrif](https://github.com/ahmdshrif))
 
 ### 💡 Others
 
--- a/packages/expo-updates/CHANGELOG.md
+++ b/packages/expo-updates/CHANGELOG.md
@@ -14,7 +14,7 @@
 - [Android] Correct `UpdatesLogReader.ONE_DAY_MILLISECONDS` from `86400` (seconds) to `86_400_000` (milliseconds), so the "older than one day" purge filter actually retains a day's worth of entries instead of ~86 seconds' worth. ([#46182](https://github.com/expo/expo/pull/46182) by [@jakequade-pc](https://github.com/jakequade-pc))
 - [iOS] Isolate UpdatesLogReaderTests from concurrent suites. ([#47082](https://github.com/expo/expo/pull/47082) by [@douglowder](https://github.com/douglowder))
 - Fix `isUpdatePending` incorrectly becoming `true` after a fetch or check that finds no new update to download. ([#47830](https://github.com/expo/expo/pull/47830) by [@kudo](https://github.com/kudo))
-- Drop the previous native state change listener before registering a new one, so re-evaluating `UpdatesEmitter` no longer stacks a subscription that retains the whole module scope — a leak that grew per request under server rendering until the dev server crashed with OOM. ([#47938](https://github.com/expo/expo/issues/47938) by [@ahmdshrif](https://github.com/ahmdshrif))
+- Drop the previous native state change listener before registering a new one, so re-evaluating `UpdatesEmitter` no longer stacks a subscription that retains the whole module scope — a leak that grew per request under server rendering until the dev server crashed with OOM. ([#48241](https://github.com/expo/expo/pull/48241) by [@ahmdshrif](https://github.com/ahmdshrif))
 
 ### 💡 Others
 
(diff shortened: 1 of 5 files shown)


brentvatne:
@expo-bot verify with fable

ahmdshrif:
Thanks for the verify run — the finding is correct and I've fixed it in 6a88206. The branch is also rebased onto current `main`, so the conflict is gone.

**The test did not run.** Confirmed locally before changing anything:

```
● keeps a single native listener when the module graph is re-evaluated
    Expected: 1
    Received: undefined
```

In the `Node` jest project `../ExpoUpdates` resolves to `ExpoUpdates.web.ts`, so the manual mock at `src/__mocks__/ExpoUpdates.ts` never applied and jest automocked the web class, making `listenerCount()` return `undefined`. The assertion was passing over nothing.

**Fix:** added `src/__mocks__/ExpoUpdates.web.ts`, a copy of `src/__mocks__/ExpoUpdates.ts`, as suggested. Both share one registry regardless, since it is keyed on `globalThis.__expoUpdatesMockListeners` precisely so `jest.resetModules()` cannot hand out a fresh set. Happy to factor the body into a shared helper instead if you would rather not have the duplicate — I kept the literal copy because a differently named file in `__mocks__/` is not matched to a sibling module, and re-exporting `./ExpoUpdates` from `ExpoUpdates.web.ts` would resolve back to itself under the web project's platform resolution.

With the mock in place the test now behaves as a real guard:

```
with the fix:      1 passed
without the fix:   Expected: 1 / Received: 6
```

That `6` is the one initial listener plus the five re-evaluations, which is the leak the issue describes.

Full `expo-updates` suite: **29 passed, 29 total**. Three suites (`plugin`, `utils`, `cli`) fail to *load* in my checkout with `Cannot find module '@expo/config-plugins'`; that is stale workspace `node_modules` on my side, in projects this PR does not touch, and no test in them fails.

On the overlap with #48337: I have no stake in which lands. That PR guards the `addListener` call with `typeof window !== 'undefined'`, which fixes the server-renderer path; this one drops the previous listener first, which also covers re-evaluation where `window` does exist, such as Fast Refresh. If you prefer #48337's shape, close this one — I would just suggest keeping a version of the regression test either way, since without it this class of leak is invisible to CI.

On `removeAllListeners`: fair point that it clears every listener for that event, not only this module's. In practice this module is the only registrant for `Expo.nativeUpdatesStateChangeEvent`, but if you would rather it be surgical I can hold the subscription on the global and remove exactly that one. Say which you prefer and I will push it.

PR #47463, opened 2026-07-02
[docs] Add EAS Observe library integration guide

jakub-agent:
# Why

Add a guide for third-party libraries to integrate with expo-observe

# How

# Test Plan

PR #48245, opened 2026-07-28
[observe] Add initIntegration function

Ubax:
# Why

In order to simplify the integration for 3rd party libraries.

# How

Expose `initIntegration` via observe proxy. 

It passes the configuration to integration using following approach:
1. If `configure` was already called - check if config exists for current integration. If it does, call the callback
2. If `configure` was not yet called - add listener
3. When it finally is called - check if config exists for current integration. If it does, call the callback. In both cases remove the listener

# Test Plan

CI

# Checklist

### DIFF
--- a/packages/expo-observe/CHANGELOG.md
+++ b/packages/expo-observe/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ### 🎉 New features
 
+- Add `Observe.initIntegration` to initialize an integration ([#48245](https://github.com/expo/expo/pull/48245) by [@Ubax](https://github.com/Ubax))
 - Expose `ObserveErrorBoundary`, a React error boundary that records render-phase errors. ([#47341](https://github.com/expo/expo/pull/47341) by [@tsapeta](https://github.com/tsapeta))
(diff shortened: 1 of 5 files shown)


Ubax:
* **#47463** <a href="https://app.graphite.com/github/pr/expo/expo/47463?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#48245** <a href="https://app.graphite.com/github/pr/expo/expo/48245?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/48245?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tsapeta (approved):
We might reconsider whether `initIntegration` is the best name, but overall it's good to go 👍

Maybe `registerIntegration` is more accurate? Now my concern is what happens when the integration config is not specified in `Observe.configure`. The callback is never fired, so in theory the integration is never initialized. Alternatively `addIntegration` or `installIntegration`.

PR #48248, opened 2026-07-28
[expo-file-system] Fix copyAsync losing Photos edits for ph:// assets on iOS

CoffeeFlux:
# Why

`FileSystem.copyAsync()` on an iOS `ph://` URI copies the asset's **unedited original**, silently discarding any edits made in the Photos app. Cropping a photo and then copying it yields the pre-crop image.

Reported in #39353, which was auto-closed as stale without triage. The behaviour is unchanged on SDK 57.

Reproduction: https://github.com/CoffeeFlux/expo-copy-bug-demo (blank SDK 57 app; runs the check on launch)

The repro compares three values per asset — `PHAsset.pixelWidth/Height` from `getAssetsAsync` (ground truth for the current rendition), the dimensions of `getAssetInfoAsync().localUri`, and the dimensions of what `copyAsync` produces. Cropping changes pixel dimensions, so the mismatch is measurable rather than a visual judgement, and unedited photos serve as controls.

**Before** (iPad mini 5, iPadOS 26.5.2, `expo@57.0.8`, `expo-file-system@57.0.1`):

```
IMG_0002.JPG expected=2448x3264 localUri=2448x3264 copyAsync=2448x3264 :: match      (control, unedited)
IMG_0001.JPG expected=2448x3264 localUri=2448x3264 copyAsync=2448x3264 :: match      (control, unedited)
IMG_0003.JPG expected=2448x2186 localUri=2448x2186 copyAsync=2448x3264 :: MISMATCH   (cropped)
```

Only the edited asset diverges, and it comes back at exactly the original dimensions. `localUri` resolves to `.../PhotoData/Mutations/DCIM/100APPLE/IMG_0003/Adjustments/FullSizeRender.jpg`, so the rendered current version exists and is reachable.

# How

In `copyPHAsset` (`ios/Legacy/FileSystemHelpers.swift`):

```swift
let firstResource = PHAssetResource.assetResources(for: asset).first
```

For an edited asset, `PHAssetResource.assetResources(for:)` returns both the original (`.photo`) and the rendered current version (`.fullSizePhoto`). `.first` returns the original. This prefers `.fullSizePhoto`/`.fullSizeVideo` when present and falls back to the previous behaviour otherwise, so unedited assets are unaffected.

It also passes `isNetworkAccessAllowed` on the resource request. With `options: nil`, an asset that exists only in iCloud has no local resource data and the copy fails outright rather than fetching it.

One note for anyone tracing this: `copyAsync` early-returns to `copyPHAsset` at `FileSystemLegacyModule.swift:147`, so `EXFileSystemAssetLibraryHandler.copy` — which sits just below and also appears to handle `ph://` — is dead code on this path.

# Test Plan

Ran the linked repro on an iPad mini (5th generation), iPadOS 26.5.2, against a photo cropped in Photos and saved over the original, with two untouched photos as controls.

**After:**

```
IMG_0002.JPG expected=2448x3264 localUri=2448x3264 copyAsync=2448x3264 :: match
IMG_0001.JPG expected=2448x3264 localUri=2448x3264 copyAsync=2448x3264 :: match
IMG_0003.JPG expected=2448x2186 localUri=2448x2186 copyAsync=2448x2186 :: match
```

The edited asset now matches and both controls are unchanged.

Caveat worth flagging: I verified the `.fullSizePhoto` change on device, but I had no iCloud-only asset available to exercise the `isNetworkAccessAllowed` change. It is reasoned rather than observed — happy to drop it into a separate PR if you would rather keep this one strictly to what has been measured. Video assets were likewise not tested; `.fullSizeVideo` is included for symmetry.

# Checklist

- [x] Documented my changes in the `CHANGELOG.md` of the relevant package
- [x] Tested on iOS (physical device)

### DIFF
--- a/packages/expo-file-system/ios/Legacy/FileSystemHelpers.swift
+++ b/packages/expo-file-system/ios/Legacy/FileSystemHelpers.swift
@@ -86,9 +86,19 @@ internal func copyPHAsset(fromUrl: URL, toUrl: URL, with resourceManager: PHAsse
       return
     }
 
-    let firstResource = PHAssetResource.assetResources(for: asset).first
+    // An edited asset carries both its original resource (`.photo`/`.video`) and the
+    // rendered current version (`.fullSizePhoto`/`.fullSizeVideo`). `.first` returns
+    // the original, which silently discards any edits made in Photos, so prefer the
+    // full-size rendition when one exists.
+    let resources = PHAssetResource.assetResources(for: asset)
+    let firstResource = resources.first { $0.type == .fullSizePhoto }
+      ?? resources.first { $0.type == .fullSizeVideo }
+      ?? resources.first
     if let firstResource {
-      resourceManager.writeData(for: firstResource, toFile: toUrl, options: nil) { error in
+      let resourceOptions = PHAssetResourceRequestOptions()
+      // Assets that only exist in iCloud have no local resource data to write.
+      resourceOptions.isNetworkAccessAllowed = true
+      resourceManager.writeData(for: firstResource, toFile: toUrl, options: resourceOptions) { error in
         if error != nil {
           promise.reject(FailedToCopyAssetException(fromUrl.absoluteString))
           return
--- a/packages/expo-file-system/ios/Legacy/FileSystemHelpers.swift
+++ b/packages/expo-file-system/ios/Legacy/FileSystemHelpers.swift
@@ -87,18 +87,22 @@ internal func copyPHAsset(fromUrl: URL, toUrl: URL, with resourceManager: PHAsse
     }
 
     // An edited asset carries both its original resource (`.photo`/`.video`) and the
-    // rendered current version (`.fullSizePhoto`/`.fullSizeVideo`). `.first` returns
-    // the original, which silently discards any edits made in Photos, so prefer the
-    // full-size rendition when one exists.
+    // rendered current version (`.fullSizePhoto`/`.fullSizeVideo`). Prefer the
+    // rendition matching the asset's media type, since an edited video can also
+    // include a `.fullSizePhoto` poster image.
     let resources = PHAssetResource.assetResources(for: asset)
-    let firstResource = resources.first { $0.type == .fullSizePhoto }
-      ?? resources.first { $0.type == .fullSizeVideo }
+    let isVideo = asset.mediaType == .video
+    let editedResourceType: PHAssetResourceType = isVideo ? .fullSizeVideo : .fullSizePhoto
+    let originalResourceType: PHAssetResourceType = isVideo ? .video : .photo
+    let resource =
+      resources.first { $0.type == editedResourceType }
+      ?? resources.first { $0.type == originalResourceType }
       ?? resources.first
-    if let firstResource {
+    if let resource {
       let resourceOptions = PHAssetResourceRequestOptions()
       // Assets that only exist in iCloud have no local resource data to write.
       resourceOptions.isNetworkAccessAllowed = true
-      resourceManager.writeData(for: firstResource, toFile: toUrl, options: resourceOptions) { error in
+      resourceManager.writeData(for: resource, toFile: toUrl, options: resourceOptions) { error in
         if error != nil {
           promise.reject(FailedToCopyAssetException(fromUrl.absoluteString))
           return
--- a/packages/expo-file-system/CHANGELOG.md
+++ b/packages/expo-file-system/CHANGELOG.md
@@ -14,7 +14,7 @@
 
 ### 🐛 Bug fixes
 
-- Fixed `copyAsync` on iOS copying the unedited original when a `ph://` asset has edits applied in Photos. ([#PRNUM](https://github.com/expo/expo/pull/PRNUM) by [@CoffeeFlux](https://github.com/CoffeeFlux))
+- Fixed `copyAsync` on iOS copying the unedited original when a `ph://` asset has edits applied in Photos. ([#48248](https://github.com/expo/expo/pull/48248) by [@CoffeeFlux](https://github.com/CoffeeFlux))
 - Fixed iOS file previews rejecting a new preview while the previous Quick Look dismissal animation is still finishing. ([#47947](https://github.com/expo/expo/pull/47947) by [@eliotgevers](https://github.com/eliotgevers))
 - Added `./next` subpath to package `exports` field to resolve Metro bundler warning. ([#44793](https://github.com/expo/expo/pull/44793) by [@chang-in](https://github.com/chang-in))
 - Fixed `FileHandle` security-scoped access, and non-SAF `content://` URI support. ([#47176](https://github.com/expo/expo/pull/47176) by [@barthap](https://github.com/barthap))
--- a/packages/expo-file-system/CHANGELOG.md
+++ b/packages/expo-file-system/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 ### 🐛 Bug fixes
 
+- Fixed `copyAsync` on iOS copying the unedited original when a `ph://` asset has edits applied in Photos. ([#PRNUM](https://github.com/expo/expo/pull/PRNUM) by [@CoffeeFlux](https://github.com/CoffeeFlux))
 - Fixed iOS file previews rejecting a new preview while the previous Quick Look dismissal animation is still finishing. ([#47947](https://github.com/expo/expo/pull/47947) by [@eliotgevers](https://github.com/eliotgevers))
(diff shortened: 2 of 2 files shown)


olsavmic:
We've been shipping essentially this patch downstream (SDK 57, `expo-file-system@57.0.1`) and hit a case worth folding in before this merges: **for an edited video, the current ordering returns a JPEG.**

```swift
let firstResource = resources.first { $0.type == .fullSizePhoto }
  ?? resources.first { $0.type == .fullSizeVideo }
  ?? resources.first
```

An edited *video* asset carries a `.fullSizePhoto` poster render alongside its `.fullSizeVideo` — Photos writes both files into the mutations directory:

```
PhotoData/Mutations/DCIM/100APPLE/IMG_0016/Adjustments/
├── FullSizeRender.jpg     ← .fullSizePhoto
├── FullSizeRender.mov     ← .fullSizeVideo
└── Adjustments.plist
```

Since `.fullSizePhoto` is checked unconditionally first, `copyAsync` on a `ph://` video URI writes the still image into the destination — with whatever extension the caller asked for. In our app the destination is `…/gallery-<id>.mov`, and `ffprobe` on the result says:

```
codec_name=mjpeg
width=1280
height=720
format_name=jpeg_pipe
```

We originally wrote the predicate as `.fullSizeVideo || .fullSizePhoto` and hit the same thing from the other direction — the array order decides, so either fixed ordering is wrong for one of the two media types. Matching the asset's own `mediaType` fixes it:

```swift
let isVideo = asset.mediaType == .video
let editedType: PHAssetResourceType = isVideo ? .fullSizeVideo : .fullSizePhoto
let originalType: PHAssetResourceType = isVideo ? .video : .photo
let resource = resources.first(where: { $0.type == editedType })
  ?? resources.first(where: { $0.type == originalType })
  ?? resources.first
```

The `originalType` fallback also tightens the un-edited path: `resources.first` on a Live Photo asset can be the paired `.pairedVideo`, so falling back to the type that matches the asset is a little safer than "whatever is at index 0".

**Repro** (iPhone 17 Pro simulator, iOS 26.5): add a video to the library, edit it in Photos (a rotate under *Crop & Rotate* is enough — any adjustment produces the paired renders), then `copyAsync` from its `ph://` URI. With this PR's ordering the output is an MJPEG still; with the `mediaType`-matched version it is byte-identical to `FullSizeRender.mov`.

Happy to open a PR against your branch if that's easier than patching it here.

CoffeeFlux:
@barthap A PR against my branch would be great, and I'm happy to look through it tomorrow. Thanks for your help with this!

packages/expo-file-system/ios/Legacy/FileSystemHelpers.swift
```diff
@@ -86,9 +86,19 @@ internal func copyPHAsset(fromUrl: URL, toUrl: URL, with resourceManager: PHAsse
       return
     }
 
-    let firstResource = PHAssetResource.assetResources(for: asset).first
+    // An edited asset carries both its original resource (`.photo`/`.video`) and the
+    // rendered current version (`.fullSizePhoto`/`.fullSizeVideo`). `.first` returns
+    // the original, which silently discards any edits made in Photos, so prefer the
+    // full-size rendition when one exists.
+    let resources = PHAssetResource.assetResources(for: asset)
+    let firstResource = resources.first { $0.type == .fullSizePhoto }
```
barthap: Please take @olsavmic's `mediaType`-matched version here instead of the fixed ordering. Two reasons beyond the edited-video case he reported:

- It matches what `expo-image-picker` already does on its video path (`ios/MediaHandler.swift:340,417`), so the repo ends up with one pattern instead of two.
- The `?? resources.first` fallback can pick a `.pairedVideo` resource on a Live Photo. Falling back to the type that matches the asset is safer than whatever sits at index 0.

Two things I would like measured, since neither is something I can check from here:

1. Did you reproduce the edited-video case? Apple does not guarantee resource ordering, so a confirmation on your own device would be good, separate from his report.
2. Curious (skip if hard to check) - What does a cropped **HEIC** produce? For an edited photo the `.fullSizePhoto` render can be a different container than the original, since Photos writes `FullSizeRender.jpg`. The destination path is caller-supplied, and callers commonly derive it from the asset's original filename, so `copyAsync(ph://…, 'x.heic')` could now land JPEG bytes under a `.heic` name. Worth checking both the selected resource's UTI and the copied bytes.

packages/expo-file-system/ios/Legacy/FileSystemHelpers.swift
```diff
@@ -86,9 +86,19 @@ internal func copyPHAsset(fromUrl: URL, toUrl: URL, with resourceManager: PHAsse
...
+      ?? resources.first { $0.type == .fullSizeVideo }
+      ?? resources.first
     if let firstResource {
-      resourceManager.writeData(for: firstResource, toFile: toUrl, options: nil) { error in
+      let resourceOptions = PHAssetResourceRequestOptions()
+      // Assets that only exist in iCloud have no local resource data to write.
+      resourceOptions.isNetworkAccessAllowed = true
```
barthap, line 104: Just a note, no action needed.

No objection to the flag. Unconditional `isNetworkAccessAllowed = true` is what the rest of the repo does on PhotoKit paths (`expo-image` `ios/Loaders/PhotoLibraryAssetLoader.swift:100`, `expo-video` `ios/Utils/URL+MediaLibraryAssets.swift:38`, `expo-image-manipulator` `ios/ImageManipulatorUtils.swift:77`), so it's consistent. You offered to split it into its own PR, no need, keeping it here is fine.

One thing worth knowing if you keep it: nothing on this path reports progress or supports cancellation, so a large iCloud-only asset looks identical to a hang from JS. That's a gap on our side, not a reason to change your PR, and we're tracking it.

barthap (commented):
Thanks for the detailed writeup and the repro, that made this easy to follow.

Yeah, the bug is real, and the fix belongs in legacy file-system. There's no alternative path today: the new file-system API has no `ph://` support at all, `PHAsset` appears only under `ios/Legacy/`. So this is a bug fix, not a new feature, which matters here. We don't add features to legacy file-system, only fix real bugs. Longer term we'd like `ph://` support to live in `expo-media-library/next` instead of the new file-system API.

One request and two questions inline. TLDR: please take @olsavmic's `mediaType`-matched selection.

On your note about dead code, you're right that `ph://` never reaches `EXFileSystemAssetLibraryHandler.copy`, because of the early return at `FileSystemLegacyModule.swift:146-149`. The handler is still live for two other routes though, `assets-library://` copies and `getInfoAsync`, which sends both schemes through it (`FileSystemLegacyModule.swift:50,60`).

Unrelated to your PR, but worth writing down while we're in here. `copyPHAsset` does nothing at all when photo permission is denied (`FileSystemHelpers.swift:75` has no `else`), so the promise never settles. That same early return also skips the destination write-permission check at `FileSystemLegacyModule.swift:152`. We'll handle both separately, nothing for you to do.

To be clear about what this review is worth: I read the code and checked the call paths, but I didn't test anything on device. The inline questions are about behavior you can measure and I can't.

P.S. the `@barthap` mention above looks like it was meant for @olsavmic 😅 A PR against your branch from them sounds good to me.

packages/expo-file-system/ios/Legacy/FileSystemHelpers.swift
CoffeeFlux: I had an agent test on the sim:

I measured both requested behaviors using real Photos edits on an iPhone 17 Pro simulator running iOS 26.5:

- Edited video: rotating it produced both fullSizePhoto (public.jpeg) and fullSizeVideo (com.apple.quicktime-movie). The old ordering copied an 8,589-byte JPEG into the .mov path. The fixed selector copied the 806,731-byte QuickTime movie, confirmed by its ftyp signature and file inspection.

- Cropped HEIC: the original resource remained public.heic, but Photos generated the edited fullSizePhoto as public.jpeg. Copying that to a caller-provided .heic destination produced 91,976 bytes of actual JPEG data. So the extension mismatch is real and remains a separate API concern.

barthap (approved):
Nice, thanks for measuring both. The numbers make this easy to sign off.

The selector looks good to me:
- Live Photos now prefer `.fullSizePhoto` then `.photo`, so the `.pairedVideo` hazard is gone
- slow-mo still resolves to `.fullSizeVideo`, and audio or `.unknown` assets fall through to `resources.first`, same as before.

Regarding the HEIC result, agreed it's a separate concern. I don't think there's anything to do about it inside legacy `copyAsync`, since the destination path belongs to the caller and the API has no way to report the real format back.

Thanks for sticking with this, and thanks @olsavmic for catching the video case.

Issue #48124, opened 2026-07-25
[expo/fetch][iOS] response.text() and arrayBuffer() never settle when a request fails after the response is delivered


0xLuca:
### Minimal reproducible example

https://github.com/0xLuca/expo-fetch-hang-repro

### Steps to reproduce

Platform: iOS (physical device, development build). Package manager: npm.

## Steps

1. `npm install`
2. `node server.mjs` (bundled, zero dependencies, listens on `http://localhost:8787`)
3. `npx expo start --ios`, or on a device replace `localhost` in the app's text field with your LAN IP
4. Tap **Run all tests**, keeping the app foregrounded

Each test races the body read against a watchdog, so a hang shows as `PENDING` with elapsed ms rather than a frozen screen.

## Results (iPhone, development build, expo@57.0.8, react-native@0.86.0)

| Test | Expected | Actual |
| --- | --- | --- |
| 1. Control, complete response | resolved | `resolved @0ms` |
| 2. Response delivered, socket dies mid-body, `text()` | rejected | **`PENDING @10001ms`** |
| 3. Same failure, read through `res.body` | rejected | `rejected @1094ms` "The network connection was lost" |
| 4. Same as 2, through the global `fetch` | rejected | **`PENDING @10001ms`** |
| 5. `abort()` while a body read is pending | rejected | **`PENDING @10002ms`** |
| 6. Socket dies with no body bytes at all | rejected | `fetch()` itself rejects, "cannot parse response" |

`PENDING` means the promise neither resolved nor rejected. Waiting longer changes nothing.

On test 5 the server logs `closed by client after 517ms (cancel reached the wire)`. The cancellation genuinely closes the socket, and the promise is stranded anyway.

## Cause

`ios/Fetch/ExpoFetchModule.swift`:

```swift
AsyncFunction("text") { (response: NativeResponse, promise: Promise) in
  response.waitFor(states: [.bodyCompleted]) { _ in
    let data = response.sink.finalize()
    promise.resolve(String(decoding: data, as: UTF8.self))
  }
}.runOnQueue(fetchRequestQueue)
```

[...]

PR #48251, opened 2026-07-28
[fetch] Reject aborted requests with an AbortError and give FetchError a name

0xLuca:
# Why

Fixes the second half of #48124, which @zoontek asked for a PR on. The first half landed in #48230, and this branch is rebased on it.

`FetchError` never set `name`, so it inherited `'Error'` and every rejection out of `expo/fetch` looked alike. On Android and iOS `expo/fetch` is installed as the global `fetch`, and the React Native `fetch` it replaces (whatwg-fetch) rejects aborts with `new DOMException('Aborted', 'AbortError')`. So the standard `err.name === 'AbortError'` check, which works against RN's fetch and every browser, silently stops working once `expo/fetch` takes over the global, and there is no documented alternative for telling an abort from a network failure.

Everything needed to do this properly is already in the package: `src/winter/DOMException.ts` is installed as the global `DOMException`, and `src/winter/AbortSignal.ts` gives `AbortSignal.timeout()` a `reason` of a `TimeoutError`.

| | before | after |
| --- | --- | --- |
| `controller.abort()` | `Error: fetch failed: Fetch request has been canceled` | `AbortError: The operation was aborted.` |
| `controller.abort(reason)` | same as above | rejects with `reason` |
| `AbortSignal.timeout()` | same as above | `TimeoutError: The signal timed out.` (the signal's own `reason`) |
| dropped connection | `Error: fetch failed: ...` | `FetchError: fetch failed: ...` |

# How

`FetchErrors.ts` sets `this.name = 'FetchError'` and adds `createAbortError(signal)`, which returns the signal's own `reason` per the spec and falls back to a `DOMException` named `AbortError` (React Native's `AbortController` polyfill, `abort-controller@3`, does not implement `reason`). It mirrors the private `getAbortReason` in `AbortSignal.ts`.

`fetch.ts` throws that for an already-aborted signal and for a `start()` rejection while the signal is aborted; anything else keeps the existing `FetchError` wrapping.

`FetchResponse.ts` takes the signal as an optional second constructor argument and routes `text()`, `arrayBuffer()` and the body stream's `didFailWithError` through one `createBodyError` mapper, so a body read that fails after the response arrived reports what the request itself would. `json()`, `blob()`, `bytes()` and `formData()` funnel through those two, and clones are unaffected because they read through the original's teed stream, which is already mapped.

Worth flagging for the changelog placement: this is a bug fix in the sense that the errors were unusable, but it does change what an aborted request rejects with. `FetchError` is not exported from `expo/fetch` (`src/winter/fetch/index.ts` exports only `./fetch` and `./fetch.types`), so no caller can hold an `instanceof` reference to it, but code matching on the old `'fetch failed: Fetch request has been canceled'` message string would need to move to `err.name`. Happy to move the entry under breaking changes if you prefer.

# Test Plan

Unit tests, `packages/expo`:

- new `__tests__/FetchErrors-test.ts`: `name` on `FetchError` and `createFromError`, plus the three `createAbortError` branches (signal reason, no reason, no signal).
- new `__tests__/fetch-test.native.ts`: a stub `NativeRequest` whose `start` stays pending until the test resolves, cancels or fails it, covering the pre-aborted signal, a custom abort reason, an abort in flight and a plain failure. Native-only, since `../fetch` resolves to `fetch.web.ts` on the Node and Web jest projects.
- `__tests__/FetchResponse-test.ts`: six cases for buffered reads and the body stream, with and without an aborted signal.

`pnpm test`, `pnpm typecheck`, `pnpm lint` and `pnpm format` pass in `packages/expo` and `apps/test-suite`.

`apps/test-suite/tests/Fetch.ts`: the five abort tests asserted only that something threw. They now assert `error.name`, and the `AbortSignal.timeout` and `AbortSignal.any` cases assert `error === signal.reason`. The one assertion that had to change is `should abort streaming request before receiving chunks`, which matched the old message.

On device, iPhone, the minimal repro app from #48124 with these three files patched into its `node_modules` (JS only, so Metro picks them up without a rebuild):

| Test | Result |
| --- | --- |
| 1. control, complete response | `resolved` |
| 3. socket dies mid body, read through `res.body` | `rejected` `FetchError: fetch failed: The network connection was lost.` |
| 6. socket dies with no body bytes | `rejected` `FetchError: fetch failed: unexpectedexception cannot parse response ...` |
| 7. `abort()` before the response arrives | `rejected` `AbortError: The operation was aborted.` |
| 8. `AbortSignal.timeout()` before the response | `rejected` `TimeoutError: The signal timed out.` |
| 9. `abort()` while a stream read is pending | `rejected` `AbortError: The operation was aborted.` |

Tests 2, 4 and 5 of that app read the body with `text()` and still hang there, because that device build is the released `expo@57.0.8`, which predates #48230. Now that #48230 is in, those promises reject and run through the same `createBodyError` path that tests 3 and 6 exercise above, and the unit tests cover it directly.

# Checklist

### DIFF
--- a/docs/pages/versions/unversioned/sdk/expo.mdx
+++ b/docs/pages/versions/unversioned/sdk/expo.mdx
@@ -48,7 +48,7 @@ console.log(buffer.length); // 512
 
 > **Note:** On Android and iOS, `expo/fetch` is also installed as the global `fetch`. This means any calls to `fetch(...)` without importing from `expo/fetch` use the same WinterCG-compliant implementation. To keep React Native's built-in `fetch` as the global, set `EXPO_PUBLIC_USE_RN_FETCH=1` in your [environment variables](/guides/environment-variables/). Named imports from `expo/fetch` continue to work regardless of this flag.
 
-Rejections carry a `name` you can branch on. A request stopped through an [`AbortSignal`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) rejects with that signal's `reason`, which is an `AbortError` for `controller.abort()` and a `TimeoutError` for `AbortSignal.timeout()`. Every other failure, such as a dropped connection, rejects with a `FetchError`. The same applies to a body read that fails after the response has arrived, so `response.text()`, `response.json()`, `response.arrayBuffer()`, `response.blob()` and the body stream all report failures the same way.
+Rejections carry a `name` you can branch on. A request stopped through an [`AbortSignal`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) rejects with that signal's `reason`, which is an `AbortError` for `controller.abort()` and a `TimeoutError` for `AbortSignal.timeout()`. On Android and iOS, every other failure, such as a dropped connection, rejects with a `FetchError`. On web, `expo/fetch` is the platform's own `fetch`, so those failures reject with a `TypeError`. The same applies to a body read that fails after the response has arrived, so `response.text()`, `response.json()`, `response.arrayBuffer()`, `response.blob()`, and the body stream all report failures the same way.
 
 ```ts Telling an abort from a network failure
 import { fetch } from 'expo/fetch';
@@ -58,13 +58,14 @@ const controller = new AbortController();
 try {
   const resp = await fetch('https://httpbin.org/get', { signal: controller.signal });
   console.log(await resp.json());
-} catch (error) {
+} catch (error: any) {
   if (error.name === 'AbortError') {
     // The request was canceled on purpose, usually not worth reporting.
   } else if (error.name === 'TimeoutError') {
     // The signal passed to fetch came from AbortSignal.timeout().
   } else {
-    // A FetchError: the request or the body read failed.
+    // A FetchError on Android and iOS, a TypeError on web: the request
+    // or the body read failed.
     console.error(error);
   }
 }
--- a/docs/pages/versions/unversioned/sdk/expo.mdx
+++ b/docs/pages/versions/unversioned/sdk/expo.mdx
@@ -48,6 +48,28 @@ console.log(buffer.length); // 512
 
 > **Note:** On Android and iOS, `expo/fetch` is also installed as the global `fetch`. This means any calls to `fetch(...)` without importing from `expo/fetch` use the same WinterCG-compliant implementation. To keep React Native's built-in `fetch` as the global, set `EXPO_PUBLIC_USE_RN_FETCH=1` in your [environment variables](/guides/environment-variables/). Named imports from `expo/fetch` continue to work regardless of this flag.
 
+Rejections carry a `name` you can branch on. A request stopped through an [`AbortSignal`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) rejects with that signal's `reason`, which is an `AbortError` for `controller.abort()` and a `TimeoutError` for `AbortSignal.timeout()`. Every other failure, such as a dropped connection, rejects with a `FetchError`. The same applies to a body read that fails after the response has arrived, so `response.text()`, `response.json()`, `response.arrayBuffer()`, `response.blob()` and the body stream all report failures the same way.
+
+```ts Telling an abort from a network failure
+import { fetch } from 'expo/fetch';
+
+const controller = new AbortController();
+
+try {
+  const resp = await fetch('https://httpbin.org/get', { signal: controller.signal });
+  console.log(await resp.json());
+} catch (error) {
+  if (error.name === 'AbortError') {
+    // The request was canceled on purpose, usually not worth reporting.
+  } else if (error.name === 'TimeoutError') {
+    // The signal passed to fetch came from AbortSignal.timeout().
+  } else {
+    // A FetchError: the request or the body read failed.
+    console.error(error);
+  }
+}
+```
+
 ### Encoding APIs
 
 `TextEncoder` and `TextDecoder` are built-in APIs that provide a way to encode and decode text in various character encodings. They are available on all platforms. Refer to the [browser and server runtime support](https://caniuse.com/textencoder) for web and Node.js.
--- a/packages/expo/CHANGELOG.md
+++ b/packages/expo/CHANGELOG.md
@@ -5,12 +5,12 @@
 ### 🛠 Breaking changes
 
 - Raise minimum Node.js version to `^22.13.0` ([#47202](https://github.com/expo/expo/pull/47202) by [@kitten](https://github.com/kitten))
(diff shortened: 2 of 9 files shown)


docs/pages/versions/unversioned/sdk/expo.mdx
```diff
@@ -48,6 +48,28 @@ console.log(buffer.length); // 512
 
 > **Note:** On Android and iOS, `expo/fetch` is also installed as the global `fetch`. This means any calls to `fetch(...)` without importing from `expo/fetch` use the same WinterCG-compliant implementation. To keep React Native's built-in `fetch` as the global, set `EXPO_PUBLIC_USE_RN_FETCH=1` in your [environment variables](/guides/environment-variables/). Named imports from `expo/fetch` continue to work regardless of this flag.
 
+Rejections carry a `name` you can branch on. A request stopped through an [`AbortSignal`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) rejects with that signal's `reason`, which is an `AbortError` for `controller.abort()` and a `TimeoutError` for `AbortSignal.timeout()`. Every other failure, such as a dropped connection, rejects with a `FetchError`. The same applies to a body read that fails after the response has arrived, so `response.text()`, `response.json()`, `response.arrayBuffer()`, `response.blob()` and the body stream all report failures the same way.
```
amandeepmittal: ⚠️ **correctness** `comment-format.md#severity`

This page lists `web` in its `platforms` frontmatter, but `FetchError` is native-only: **fetch.web.ts** is `export const fetch = globalThis.fetch`, and `FetchError` is imported only by **fetch.ts**. On web a dropped connection rejects with a `TypeError`, so qualify the claim the way the note above already does for the global `fetch`.

---

✏️ **style** `mechanics.md#oxford-commas`

The list needs a serial comma before "and the body stream".

```suggestion
Rejections carry a `name` you can branch on. A request stopped through an [`AbortSignal`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) rejects with that signal's `reason`, which is an `AbortError` for `controller.abort()` and a `TimeoutError` for `AbortSignal.timeout()`. On Android and iOS, every other failure, such as a dropped connection, rejects with a `FetchError`. On web, `expo/fetch` is the platform's own `fetch`, so those failures reject with a `TypeError`. The same applies to a body read that fails after the response has arrived, so `response.text()`, `response.json()`, `response.arrayBuffer()`, `response.blob()`, and the body stream all report failures the same way.
```

docs/pages/versions/unversioned/sdk/expo.mdx
```diff
@@ -48,6 +48,28 @@ console.log(buffer.length); // 512
...
+
+```ts Telling an abort from a network failure
+import { fetch } from 'expo/fetch';
+
+const controller = new AbortController();
+
+try {
+  const resp = await fetch('https://httpbin.org/get', { signal: controller.signal });
+  console.log(await resp.json());
+} catch (error) {
```
amandeepmittal: 💡 **nit** `mechanics.md#code-fences`

In a `ts` block `error` is typed `unknown`, so the two `error.name` lines below do not compile in a default Expo TypeScript project, where every template sets `strict`.

```suggestion
} catch (error: any) {
```

amandeepmittal (changes requested):

docs/pages/versions/unversioned/sdk/expo.mdx
0xLuca: Good catch, that was wrong. Confirmed: the frontmatter is `platforms: ['android', 'ios', 'tvos', 'web', 'expo-go']`, `fetch.web.ts` is `export const fetch = globalThis.fetch`, and `FetchErrors.ts` is imported only by `fetch.ts` and `FetchResponse.ts`, both of which lose out to the `.web` resolution on web. Applied your suggestion as written, serial comma included.

One extra line beyond the suggestion, since the snippet made the same claim the paragraph did: the trailing comment now reads `A FetchError on Android and iOS, a TypeError on web`. Happy to drop it if you would rather keep the comment short.

docs/pages/versions/unversioned/sdk/expo.mdx
0xLuca: Right, `strict` implies `useUnknownInCatchVariables`, and all three templates (`expo-template-default`, `expo-template-tabs`, `expo-template-blank-typescript`) set it. Applied.

packages/expo/src/winter/fetch/FetchErrors.ts
```diff
@@ -9,3 +12,13 @@ export class FetchError extends Error {
     return new FetchError(error.message, { cause: error.cause, stack: error.stack });
   }
 }
+
+/**
+ * Returns the value an aborted request rejects with. The spec rejects with the signal's own
+ * reason, so `AbortSignal.timeout()` surfaces as a `TimeoutError` rather than an `AbortError`.
+ * React Native's `AbortController` polyfill has no `reason`, hence the fallback.
+ */
+export function createAbortError(signal?: AbortSignal | null): unknown {
```
zoontek, line 21: `createAbortError` uses `reason ?? new DOMException(…)`, but `getAbortReason` in `AbortSignal.ts` returns `signal.reason` directly. The `??` means `controller.abort(null)` gets replaced with a generic `AbortError`, silently dropping the real reason.

Can you drop the `??` here so it matches `getAbortReason`, and add a test case with a nullish reason (e.g. `null`)?

packages/expo/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 
 ### 🐛 Bug fixes
 
+- Give `expo/fetch` errors a `name`, so an aborted request rejects with the signal's reason (an `AbortError`, or a `TimeoutError` from `AbortSignal.timeout()`) like React Native's `fetch` does, and any other failure rejects with a `FetchError`. ([#48251](https://github.com/expo/expo/pull/48251) by [@0xLuca](https://github.com/0xLuca))
```
zoontek: This changes the `name` and `message` of an aborted request's rejection, so it reads as a breaking change, not a bug fix. Can you move this entry from to "🛠 Breaking changes"?

zoontek (changes requested):

packages/expo/src/winter/fetch/FetchErrors.ts
0xLuca: Done, dropped the `??` and added an `abort(null)` test. It has the same shape as `getAbortReason` now.

packages/expo/CHANGELOG.md
0xLuca: Fair, moved it under breaking changes.

PR #48254, opened 2026-07-29
[ios][autolinking] Honor DEBUG in the generated provider's debug gate

chrfalch:
# Why

The generated `ExpoModulesProvider.swift` gates debug-only module registration on a single signal, `#if EXPO_CONFIGURATION_DEBUG`. That flag is set exclusively by the CocoaPods project integrator — `set_autolinking_configuration` in `scripts/ios/project_integrator.rb`, called from `cocoapods/user_project_integrator.rb`.

Nothing else defines it. So the gate is not really asking "is this a debug build?"; it is asking "did CocoaPods tell me this is a debug build?" Any Apple project that reaches the generator without a `pod install` compiles the release branch of every gate, and the two packages declaring `apple.debugOnly` — `expo-dev-menu` and `expo-dev-launcher` — do not register. Silently, with no build error.

`expo-modules-core` already treats one signal as insufficient: `Logger.swift:17` gates on `#if DEBUG || EXPO_CONFIGURATION_DEBUG`. This PR brings the generated provider in line with it.

**What this PR does not fix.** The case that surfaced this was SwiftPM, but that is not a user-visible bug today and this PR should not be read as fixing one. The SwiftPM autolinking path is not on `main`, and `expo-dev-menu` / `expo-dev-launcher` have neither a precompiled artifact nor a checked-in `Package.swift`, so a SwiftPM app including the dev client fails in the plugin before a provider is generated. Making the dev client consumable under SwiftPM is separate work. The justification here is the narrower one above: the provider should not depend on CocoaPods to know its own build configuration.

# How

Emit `#if DEBUG || EXPO_CONFIGURATION_DEBUG`, matching `Logger.swift:17`.

Both conditions are kept deliberately. `EXPO_CONFIGURATION_*` was introduced in #17378 precisely because `DEBUG` "works only in some projects … the flag might be unset depending on the build settings", so this adds a second signal rather than replacing the first.

**Behavior change worth a reviewer's eye:** Xcodeproj's `debug?` tests `GCC_PREPROCESSOR_DEFINITIONS` for `DEBUG=1` (xcodeproj 1.28.1, `lib/xcodeproj/project/object/build_configuration.rb`), not the configuration name. A `Staging` configuration copied from `Debug` therefore keeps `DEBUG=1`, already receives `EXPO_CONFIGURATION_DEBUG` today, and is unaffected by this change.

The newly affected class is narrower: a configuration that sets Swift's `DEBUG` compilation condition but omits `DEBUG=1` from `GCC_PREPROCESSOR_DEFINITIONS` will now register `debugOnly` modules where it previously did not. The `OR` cannot regress the other direction: any configuration receiving `EXPO_CONFIGURATION_DEBUG` today keeps it.

**Known limitation, not addressed here.** Registration and linking still key off different signals. The pod is *linked* when the configuration name contains `Debug` (`scripts/ios/autolinking_manager.rb:91`); registration keys off the preprocessor definitions above. A release build that deliberately links the dev client therefore still will not register it, because both signals remain proxies for the build configuration rather than checks on what is actually present. Closing that gap means replacing the proxy — for example a per-module `#if canImport(EXDevMenu)`, which asks about linkage directly. That is a larger change to the generator's shape and belongs in its own PR; this one does not make that case worse.

# Test Plan

Two tests in `src/platforms/__tests__/apple-test.ts`, exercising `generateModulesProviderAsync` through the existing memfs setup, each written red first:

- the debug-only import list carries the combined condition
- all four gates (import list plus the three getters that split registration) carry it, and no gate rests on `EXPO_CONFIGURATION_DEBUG` alone

Red, before the change:

```
● generateModulesProviderAsync › gates the debug-only import list on DEBUG as well as EXPO_CONFIGURATION_DEBUG
  expect(content).toContain('#if DEBUG || EXPO_CONFIGURATION_DEBUG\ninternal import EXDevMenu')
● generateModulesProviderAsync › gates every debug-only registration branch on DEBUG as well as EXPO_CONFIGURATION_DEBUG
  Received has value: null
Tests: 2 failed, 11 passed, 13 total
```

Green, after:

```
Test Suites: 1 failed, 21 passed, 22 total
Tests:       2 failed, 246 passed, 248 total
Snapshots:   54 passed, 54 total
```

The one failing suite is `e2e/__tests__/precompiled-derivations-test.ts` (2 tests). It fails the same way on a clean `main` and is unrelated to this change.

Also clean: `tsc -p tsconfig.json` and `oxlint`. No existing snapshot covered the gate, so none needed updating.

# Checklist

- [x] Built, type-checked, linted, and tested the changed package
- [ ] Conforms with the [documentation writing style guide](https://github.com/expo/expo/blob/main/guides/Expo%20Documentation%20Writing%20Style%20Guide.md) — n/a, no docs change

### DIFF
--- a/packages/expo-modules-autolinking/CHANGELOG.md
+++ b/packages/expo-modules-autolinking/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 ### 🐛 Bug fixes
 
+- [iOS] Emit `#if DEBUG || EXPO_CONFIGURATION_DEBUG` in the generated modules provider, so debug-only modules (`expo-dev-menu`, `expo-dev-launcher`) still register in debug builds of a project that never ran `pod install` — for example a SwiftPM app, where the CocoaPods project integrator never sets `EXPO_CONFIGURATION_DEBUG`. ([#48254](https://github.com/expo/expo/pull/48254) by [@chrfalch](https://github.com/chrfalch))
 - [iOS] Propagate `buildFromSource` across the precompiled dependency graph, so building a prebuilt dependency from source (e.g. `react-native-worklets`) also forces its dependents (e.g. `react-native-reanimated`) to build from source, instead of linking a prebuilt XCFramework against a source-built dependency. ([#48041](https://github.com/expo/expo/pull/48041) by [@chrfalch](https://github.com/chrfalch))
 - [iOS] Only stub a source pod bundled inside a prebuilt XCFramework (e.g. `SDWebImage` in `ExpoImage.xcframework`) when one of its consumers also links the owning prebuilt pod. An app extension that depends on such a pod on its own now keeps building it from source instead of failing to link with undefined symbols. ([#47847](https://github.com/expo/expo/pull/47847) by [@chrfalch](https://github.com/chrfalch))
 - [iOS] Pass target name to the generateModulesProviderCommand to match it against inline modules targets, when checking if inline modules should be autolinked with that target. ([#47502](https://github.com/expo/expo/pull/47502) by [@HubertBer](https://github.com/HubertBer))
--- a/packages/expo-modules-autolinking/CHANGELOG.md
+++ b/packages/expo-modules-autolinking/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### 🐛 Bug fixes
 
+- [iOS] Emit `#if DEBUG || EXPO_CONFIGURATION_DEBUG` in the generated modules provider, so debug-only modules (`expo-dev-menu`, `expo-dev-launcher`) still register in debug builds of a project that never ran `pod install` — for example a SwiftPM app, where the CocoaPods project integrator never sets `EXPO_CONFIGURATION_DEBUG`. ([#48254](https://github.com/expo/expo/pull/48254) by [@chrfalch](https://github.com/chrfalch))
+
 ### 💡 Others
 
 ## 58.0.0 — 2026-09-10
--- a/packages/expo-modules-autolinking/src/platforms/__tests__/apple-test.ts
+++ b/packages/expo-modules-autolinking/src/platforms/__tests__/apple-test.ts
@@ -2,8 +2,10 @@ import { vol } from 'memfs';
 import path from 'path';
 
 import { ExpoModuleConfig } from '../../ExpoModuleConfig';
+import type { ModuleDescriptorIos } from '../../types';
 import {
   formatArrayOfReactDelegateHandler,
+  generateModulesProviderAsync,
   getSwiftModuleNames,
   resolveExtraBuildDependenciesAsync,
   resolveModuleAsync,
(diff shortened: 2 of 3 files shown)


tsapeta (commented):
I think the problem here is that you can have a release build that includes the dev client 🤔

chrfalch:
@tsapeta You're right, and this PR doesn't solve it - just makes Swiftpm copy Cocapods debug builds.

`debugOnly` is already gated and can disagree with each other:

- **pod linked**: configuration *name* contains `Debug` (`autolinking_manager.rb:91`)
- **provider registers**: `GCC_PREPROCESSOR_DEFINITIONS` contains `DEBUG=1` (`project_integrator.rb:114`, via Xcodeproj's `debug?`)

This PR adds a third one, Swift's `DEBUG` condition... 

Can we do `#if canImport(EXDevMenu)` as the gate instead? Happy to redo this PR that way if you agree.

chrfalch:
Closing this. The bug it fixes is not reachable on `main`, and the change carries a real regression risk in exchange.

**Why it is not reachable.** There are exactly two callers of the provider generator on `main`: `autolinking_manager.rb` (at `pod install` time) and the build script phase written by `project_integrator.rb` (on every Xcode build). Both are installed by CocoaPods, and `project_integrator.rb` sets `EXPO_CONFIGURATION_DEBUG` / `EXPO_CONFIGURATION_RELEASE` on every configuration it touches. The flag this PR treats as possibly-missing is therefore always defined wherever the generated provider is compiled.

**What it risks instead.** The old gate registers debug-only modules when `EXPO_CONFIGURATION_DEBUG` is set. Linking is decided separately, by whether the configuration *name* contains "Debug" (`autolinking_manager.rb`). Those two agree today. Adding `DEBUG ||` introduces a third signal — Swift's own `DEBUG`, from `SWIFT_ACTIVE_COMPILATION_CONDITIONS` — which can be set where the other two are not. A configuration named e.g. `Staging`, with Swift `DEBUG` but no `GCC_PREPROCESSOR_DEFINITIONS DEBUG=1`, builds fine today: not linked, not registered. With this PR it would be registered but still not linked, so `internal import EXDevMenu` would fail to compile. That is a new failure mode, not a fixed one.

**On the objection raised in review.** @tsapeta is right, and the point is bigger than this diff. Gating registration on a build-configuration *proxy* is the wrong mechanism regardless of which proxy is picked — the generated provider should ask whether the module is actually there, not which configuration it thinks it is in. That means `#if canImport(EXDevMenu)` per module, keyed off each entry's `swiftModuleName`. It is a redesign of the gate rather than a tweak to it: `generateDebugOnlyImportList`, `generateModuleClasses`, `generateClasses` and `generateReactDelegateHandlers` all flatten their inputs with `concat` before emitting, which discards the entry-to-module association a per-module `canImport` needs.

**Where the work goes.** The SwiftPM motivation for this PR is not live yet either: the SwiftPM autolinking plugin is not on `main`, and the four dev-client packages have neither `spm.config.json` nor `Package.swift`, so the plugin cannot resolve them at all today. Making the dev client consumable under SwiftPM comes first; the `canImport` redesign lands with it, where it can be built and verified against a real SwiftPM app instead of argued from the CocoaPods side.

The two tests added here are worth keeping — they pin the gate's emitted text, which is exactly what a `canImport` rewrite has to change deliberately. They will be lifted onto that branch.

PR #48256, opened 2026-07-29
[router] Add support for replace action in tabs

Ubax:
# Why

`TabRouter` does not support `replace` action, which is needed for `Redirect` to work properly

# How

1. Merge `TabRouter` override with `BaseTabRouter`
2. Add support for `replace` in `TabRouter` - tab change + history overwrite
3. Close drawer when replace is dispatched - similar to other actions
4. Update headless tabs router - remove code that is not needed anymore

# Test Plan

CI

# Checklist

### DIFF
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -29,6 +29,7 @@
 - [ios] Fix memory leaks in the native toolbar and link preview from strong-reference cycles and closures that strongly captured `self`. ([#47378](https://github.com/expo/expo/pull/47378) by [@Ubax](https://github.com/Ubax))
 - [android][ios] Fix `expo-router/head` and `expo-router/stack` resolution on native platforms. ([#47870](https://github.com/expo/expo/pull/47870) by [@hassankhan](https://github.com/hassankhan))
 - Fix missing subpath warning from Metro when importing from `expo-router/server` ([#48045](https://github.com/expo/expo/pull/48045) by [@hassankhan](https://github.com/hassankhan))
+- Fix `replace` navigation in tabs leaving the replaced route in history. ([#48256](https://github.com/expo/expo/pull/48256) by [@Ubax](https://github.com/Ubax))
 
 ### 💡 Others
 
--- a/packages/expo-router/src/__tests__/drawer-router-replace.test.ios.tsx
+++ b/packages/expo-router/src/__tests__/drawer-router-replace.test.ios.tsx
@@ -1,0 +1,34 @@
+import { act, screen } from '@testing-library/react-native';
+
+import { router } from '../imperative-api';
+import { Drawer } from '../layouts/Drawer';
+import { renderRouter } from '../testing-library';
+
+jest.mock('react-native-drawer-layout', () => {
+  const React = jest.requireActual('react') as typeof import('react');
+  const actual = jest.requireActual(
+    'react-native-drawer-layout'
+  ) as typeof import('react-native-drawer-layout');
+
+  return {
+    ...actual,
+    Drawer: ({ children }: React.PropsWithChildren) =>
+      React.createElement(React.Fragment, null, children),
+  };
+});
+
+it('removes the replaced drawer route from history', () => {
+  renderRouter({
+    _layout: () => <Drawer backBehavior="history" />,
+    index: () => null,
+    second: () => null,
+    third: () => null,
+  });
+
+  act(() => router.push('/second'));
+  act(() => router.push('/third'));
+  act(() => router.replace('/'));
+
+  act(() => router.back());
+  expect(screen).toHavePathname('/second');
+});
--- a/packages/expo-router/src/react-navigation/routers/DrawerRouter.tsx
+++ b/packages/expo-router/src/react-navigation/routers/DrawerRouter.tsx
@@ -209,6 +209,7 @@ export function DrawerRouter({
 
           return addDrawerToHistory(state);
 
+        case 'REPLACE':
         case 'JUMP_TO':
         case 'NAVIGATE':
         case 'NAVIGATE_DEPRECATED': {
--- a/packages/expo-router/src/react-navigation/routers/TabRouter.tsx
+++ b/packages/expo-router/src/react-navigation/routers/TabRouter.tsx
@@ -1,6 +1,5 @@
 import { nanoid } from 'nanoid/non-secure';
 
-import type { StackActionType } from '../core';
 import { BaseRouter } from './BaseRouter';
 import { createParamsFromAction } from './createParamsFromAction';
 import type {
@@ -13,12 +12,19 @@ import type {
   Router,
 } from './types';
 
-export type TabActionType = {
-  type: 'JUMP_TO';
-  payload: { name: string; params?: object };
-  source?: string;
-  target?: string;
-};
+export type TabActionType =
+  | {
+      type: 'JUMP_TO';
+      payload: { name: string; params?: object };
+      source?: string;
+      target?: string;
+    }
+  | {
+      type: 'REPLACE';
+      payload: { name: string; params?: object };
+      source?: string;
+      target?: string;
+    };
 
 export type BackBehavior =
   | 'firstRoute'
@@ -61,6 +67,20 @@ export type TabNavigationState<ParamList extends ParamListBase> = Omit<
 
 export type TabActionHelpers<ParamList extends ParamListBase> = {
   /**
+   * Replaces the current tab with another tab.
+   *
+   * @param screen Name of the tab that will replace the current one.
+   * @param [params] Params object for the new tab.
+   */
+  replace<RouteName extends keyof ParamList>(
+    ...args: RouteName extends unknown
+      ? undefined extends ParamList[RouteName]
+        ? [screen: RouteName, params?: ParamList[RouteName]]
+        : [screen: RouteName, params: ParamList[RouteName]]
+      : never
+  ): void;
+
+  /**
    * Jump to an existing tab.
    *
    * @param screen Name of the route to jump to.
@@ -81,6 +101,12 @@ export const TabActions = {
   jumpTo(name: string, params?: object) {
     return {
       type: 'JUMP_TO',
+      payload: { name, params },
+    } as const satisfies TabActionType;
+  },
+  replace(name: string, params?: object) {
+    return {
+      type: 'REPLACE',
       payload: { name, params },
     } as const satisfies TabActionType;
   },
@@ -182,9 +208,16 @@ const changeIndex = (
   };
 };
 
-// TODO(@ubax): unify the logic into single router instead of BaseTabRouter and override
-// TODO(@ubax): add REPLACE action to CommonAction type and handle it in all routers
-function BaseTabRouter({ initialRouteName, backBehavior = 'firstRoute' }: TabRouterOptions) {
+/**
+ * TabRouter is considered an internal implementation and its behavior may change without a notice between expo-router's version
+ */
+export function TabRouter({
+  initialRouteName,
+  backBehavior = 'firstRoute',
+}: TabRouterOptions): Router<
+  TabNavigationState<ParamListBase>,
+  TabActionType | CommonNavigationAction
+> {
   const router: Router<
     TabNavigationState<ParamListBase>,
     TabActionType | CommonNavigationAction
@@ -315,7 +348,12 @@ function BaseTabRouter({ initialRouteName, backBehavior = 'firstRoute' }: TabRou
     },
 
     getStateForAction(state, action, { routeParamList, routeGetIdList }) {
+      if (action.target && action.target !== state.key) {
+        return null;
+      }
+
       switch (action.type) {
+        case 'REPLACE':
         case 'JUMP_TO':
         case 'NAVIGATE':
         case 'NAVIGATE_DEPRECATED': {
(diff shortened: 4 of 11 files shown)


Ubax:
* **#47986** <a href="https://app.graphite.com/github/pr/expo/expo/47986?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#48257** <a href="https://app.graphite.com/github/pr/expo/expo/48257?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#48256** <a href="https://app.graphite.com/github/pr/expo/expo/48256?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/48256?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/expo-router/src/ui/TabRouter.tsx
```diff
@@ -41,37 +36,7 @@ export function ExpoTabRouter(options: ExpoTabRouterOptions) {
...
-          state = {
-            ...state,
-            key: `${state.key}-replace`,
-            // Omit the previous history entry that we are replacing
-            history: [
-              ...state.history.slice(0, previousIndex),
-              ...state.history.splice(state.index),
-            ],
-          };
-        }
```
Ubax, line 73: This is now handled in the base `TabRouter`

packages/expo-router/src/react-navigation/routers/TabRouter.tsx
```diff
@@ -315,7 +349,12 @@ function BaseTabRouter({ initialRouteName, backBehavior = 'firstRoute' }: TabRou
     },
 
     getStateForAction(state, action, { routeParamList, routeGetIdList }) {
+      if (action.target && action.target !== state.key) {
+        return null;
+      }
```
Ubax, line 353: Moved from line 508

packages/expo-router/src/react-navigation/routers/DrawerRouter.tsx
```diff
@@ -209,6 +209,7 @@ export function DrawerRouter({
 
           return addDrawerToHistory(state);
 
+        case 'REPLACE':
```
jakub-agent, line 212: [🏗️ **design**] This case is unreachable from the public API. `getNavigationAction.ts:81` downgrades the action before it is ever dispatched:

```ts
} else if (type === 'REPLACE' && navigationState.type === 'drawer') {
  type = 'JUMP_TO';
}
```

So `router.replace()` and `<Redirect>` inside a `Drawer` still dispatch `JUMP_TO`, and the replaced route stays in `history` — the exact bug this PR fixes for tabs. The new `DrawerRouter.test.tsx` case passes only because it calls `DrawerActions.replace()` directly, which bypasses that path.

Now that `TabRouter` handles `REPLACE` and `DrawerRouter` delegates to it, that downgrade looks removable. If it is intentional (drawers arguably have no "replace" concept), consider dropping this case instead so the router and the dispatcher agree.

packages/expo-router/src/react-navigation/routers/__tests__/TabRouter.test.tsx
```diff
@@ -1055,6 +1073,135 @@ test('ensures unique ID for jump to', () => {
   });
 });
 
+test('replaces the focused route when the destination is the first tab', () => {
```
jakub-agent, line 1076: [💡 suggestion] All five new router tests use `backBehavior: 'history'` or `'fullHistory'`, but the default is `'firstRoute'` — and that is where the fix has the largest user-visible effect. Running the old implementation with `routes: [bar, baz, qux]`, focused `baz`, `TabActions.replace('qux')` produced `index: 2` with `history: ['bar']`: the last history entry no longer matched `index`, and the subsequent `GO_BACK` returned `null`, so Android hardware back after a redirect in a default `<Tabs>` bubbled past the navigator. The new code returns `history: ['bar', 'qux']` and `GO_BACK` lands on `bar`.

Worth a case at the default `backBehavior` asserting the anchor survives and `GO_BACK` reaches it, since that is the regression most likely to come back. `'order'`/`'initialRoute'`/`'none'` are also uncovered.

Relatedly, the stated motivation is `<Redirect>`, but `src/__tests__/tab-router-replace.test.ios.tsx` exercises `router.replace()`. An integration case rendering `<Redirect>` inside a `<Tabs>` layout would cover the reported behavior directly.

packages/expo-router/src/react-navigation/routers/TabRouter.tsx
```diff
@@ -495,56 +535,23 @@ function BaseTabRouter({ initialRouteName, backBehavior = 'firstRoute' }: TabRou
...
+  const replacedRouteKey = previousState.routes[previousState.index]?.key;
+  const focusedRouteKey = nextState.routes[nextState.index]?.key;
+  if (!replacedRouteKey || replacedRouteKey === focusedRouteKey) {
+    return nextState;
+  }
 
-        return nextState;
-      }
+  const history = [...nextState.history];
+  for (let index = history.length - 1; index >= 0; index--) {
```
jakub-agent: [🐧 nit] `findLastIndex` is already used twice in this file (in `changeIndex` and the `SET_PARAMS` case), so this reverse loop can match the surrounding style and avoid copying the array when nothing is removed:

```ts
const replacedIndex = nextState.history.findLastIndex((item) => item.key === replacedRouteKey);

if (replacedIndex === -1) {
  return nextState;
}

return {
  ...nextState,
  history: nextState.history.filter((_, index) => index !== replacedIndex),
};
```

packages/expo-router/CHANGELOG.md
```diff
@@ -44,6 +44,7 @@
 
 ### 💡 Others
 
+- Add support for replace action in tabs ([#48256](https://github.com/expo/expo/pull/48256) by [@Ubax](https://github.com/Ubax))
```
jakub-agent: [🐧 nit] This reads as a bug fix rather than an internal change: before this PR, `replace` into a tab left the replaced route reachable via back, and left `index` disagreeing with the last `history` entry. Consider moving the entry to `### 🐛 Bug fixes` so it shows up in release notes for users hitting the `Redirect`-in-tabs problem.

jakub-agent (commented):

The refactor is a real improvement, not just a move. I ran the old and new `TabRouter` side by side against the same action sequences, and the previous `-replace` wrapper produced broken state in several configurations:

| scenario (`routes: [bar, baz, qux]`) | before | after |
| --- | --- | --- |
| `firstRoute`, focused `baz`, `replace('qux')` | `index: 2`, `history: [bar]` — last history entry disagrees with `index`, and `GO_BACK` returns `null` | `history: [bar, qux]`, `GO_BACK` → `bar` |
| `history`, focused `qux`, `replace('bar')` | `history: [baz, qux, bar]` — `GO_BACK` returns to the route that was just replaced | `history: [baz, bar]`, `GO_BACK` → `baz` |
| `fullHistory`, `replace` a tab with itself | dropped an unrelated leading entry | history unchanged |
| `replace` with a changed `getId` | dropped an unrelated history entry | only the replaced entry is dropped |

Dropping `key: \`${state.key}-replace\`` also matters: the old code mutated the navigator's own state key on every replace, so `action.target` stopped matching it and the suffix compounded across repeated replaces. Deleting the duplicate override in `src/ui/TabRouter.tsx` is behavior-preserving — that branch reassigned `action.type` to `JUMP_TO` and then fell through to the reset logic, where `shouldReset` was always `false` because the route had just been added to history, so it ended in a no-op `getStateForRouteFocus`.

Verified locally on `bf44fe9`: `expo-router` builds with no type errors, `oxlint` is clean on the changed files, and the full Jest suite passes (4477 passed; the 3 `__rsc_tests__/tabs.test.tsx` snapshot failures are an artifact of my non-standard worktree path, not this diff).

Four points below, none blocking. Two smaller notes not worth inline comments: `TabRouter`'s `REPLACE` ignores `action.source`, where `StackRouter` uses it to target the dispatching route rather than the focused one (`router.replace()` never sets `source`, so this only surfaces for a direct `navigation.replace()` from a non-focused tab); and `NativeBottomTabsRouter` still only scrubs zoom-transition params and sets the no-animation flag in its `NAVIGATE` case, so `router.replace()` between native tabs skips that — both pre-existing, but more visible now that `REPLACE` is a first-class tab action.

packages/expo-router/src/__tests__/drawer-router-replace.test.ios.tsx
```diff
@@ -0,0 +1,34 @@
...
+  return {
+    ...actual,
+    Drawer: ({ children }: React.PropsWithChildren) =>
+      React.createElement(React.Fragment, null, children),
+  };
+});
+
+it('removes the replaced drawer route from history', () => {
+  renderRouter({
+    _layout: () => <Drawer backBehavior="history" />,
```
jakub-agent, line 22: [💡 suggestion] Removing the drawer downgrade also changes `Drawer` at its **default** `backBehavior` (`firstRoute`), which this test doesn't cover — it pins `backBehavior="history"`. Checked with `DrawerRouter({})`, `routeNames: [bar, baz, qux]`, focused `bar`:

- before (downgraded to `JUMP_TO`): `history: [bar, qux]` → `GO_BACK` returns to `bar`
- now: `history: [qux]` → `GO_BACK` returns `null`, so the action bubbles to the parent

To be clear this is the drawer converging on long-standing tab behavior, not a new semantic: with `firstRoute`, `changeIndex` recomputes `history` as `[anchor, focused]`, so replacing *from* the anchor prunes the anchor itself. I confirmed the pre-PR `TabRouter` already returned `null` from `GO_BACK` in that same situation, so tabs are unaffected here — the drawer just stopped being special-cased.

So: not a blocker, and I don't think it needs changing in this PR. A default-`backBehavior` drawer case would still be worth adding to pin the intent, the way the four new `TabRouter` tests do. Separately, whether `firstRoute` *should* prune its anchor on replace looks like a real open question (its contract is "back returns to the first route", and the anchor is structural rather than a visited entry) — but that predates this PR and belongs in its own change.

packages/expo-router/CHANGELOG.md
```diff
@@ -29,6 +29,7 @@
 - [ios] Fix memory leaks in the native toolbar and link preview from strong-reference cycles and closures that strongly captured `self`. ([#47378](https://github.com/expo/expo/pull/47378) by [@Ubax](https://github.com/Ubax))
 - [android][ios] Fix `expo-router/head` and `expo-router/stack` resolution on native platforms. ([#47870](https://github.com/expo/expo/pull/47870) by [@hassankhan](https://github.com/hassankhan))
 - Fix missing subpath warning from Metro when importing from `expo-router/server` ([#48045](https://github.com/expo/expo/pull/48045) by [@hassankhan](https://github.com/hassankhan))
+- Fix `replace` navigation in tabs leaving the replaced route in history. ([#48256](https://github.com/expo/expo/pull/48256) by [@Ubax](https://github.com/Ubax))
```
jakub-agent, line 32: [🐧 nit] The entry says "in tabs", but this iteration also changed `Drawer`: deleting the `getNavigationAction.ts` downgrade means `router.replace()` and `<Redirect>` inside a drawer now prune the replaced route, so back behavior shifts there too (see my note on the new drawer test). Worth widening the wording to cover tabs and drawers so `Drawer` users can find it.

Also worth a second thought on the section: back-button behavior changing on a default-configured navigator may belong under `### 🛠 Breaking changes` rather than 🐛 Bug fixes. Your call — I'd defer to how the router team has categorized similar navigation-semantics fixes.

jakub-agent (approved):

**Iteration 2** — re-reviewed `04e0b8f` (force-pushed and rebased onto `3c7c436`, which now includes the JS-tabs standard-navigation integration from #48292).

All four points from the previous review are addressed:

| previous finding | resolution |
| --- | --- |
| `DrawerRouter`'s `REPLACE` case unreachable from `router.replace()` | the `REPLACE`→`JUMP_TO` drawer downgrade is deleted from `getNavigationAction.ts`, its unit test is inverted to assert `REPLACE` survives, and a new `drawer-router-replace.test.ios.tsx` covers the integration path |
| default `backBehavior` untested | four new router tests for `firstRoute`, `order`, `initialRoute`, `none` — each asserting the resulting `GO_BACK` target, which is the part that was actually broken |
| `<Redirect>` not covered | `removes a redirecting tab from history` renders `<Redirect>` inside `<Tabs>` and asserts the redirecting route is pruned |
| reverse loop → `findLastIndex` | applied, and it now returns `nextState` untouched when nothing matches |
| CHANGELOG under 💡 Others | moved to 🐛 Bug fixes with a clearer description |

`src/ui/TabRouter.tsx` and `src/react-navigation/routers/DrawerRouter.tsx` are byte-identical to what I reviewed at `bf44fe9`, so the earlier analysis of those still holds.

Verified locally on `04e0b8f`: builds with no type errors, `oxlint` clean on all changed files, affected suites pass (283 tests), full suite passes (4570 passed). The 3 `__rsc_tests__/tabs.test.tsx` snapshot failures are the same worktree-path artifact as last time — that snapshot embeds a relative asset path — not a regression.

On the one semantic change this iteration introduces — `Drawer` now prunes the replaced route — I checked the pre-PR `TabRouter` directly rather than reasoning from the diff. At `backBehavior: 'firstRoute'`:

| | pre-PR tabs | this PR |
| --- | --- | --- |
| replace from a non-anchor route | `history: [bar]`, `index: 2` → `GO_BACK` returns `null` (history disagreed with `index`) | `history: [bar, qux]` → back lands on `bar` |
| replace from the anchor itself | `history: [bar]`, `index: 2` → `GO_BACK` returns `null` | `history: [qux]` → `GO_BACK` returns `null` |

So tabs strictly improve: the broken case is fixed and the other is unchanged in outcome (just internally consistent now). The drawer change is convergence onto that same behavior, not a new invention — it had been special-cased out of `REPLACE` entirely.

Approving. Two non-blocking notes below, both about the newly-added drawer scope rather than the tab fix.

PR #48257, opened 2026-07-29
[router] Extract `useHiddenTabRedirect` from `NativeTabs`

Ubax:
# Why

In the follow-up PRs, the detection of `routeNames` changes will be removed. This PR extracts `useVisibleTabsWithRedirect`, so that it can be reused in follow-up changes.

# How

1. Extract `useVisibleTabsWithRedirect` from NativeTabs - this logic will be reused to redirect hidden tabs in headless navigator as well. Hidden in this case means `hidden={true}` or if the route is not specified in the `_layout`

# Test Plan

CI

# Checklist

### DIFF
--- a/packages/expo-router/src/standard-navigation/useVisibleTabsWithRedirect.ts
+++ b/packages/expo-router/src/standard-navigation/useVisibleTabsWithRedirect.ts
@@ -1,0 +1,84 @@
+import { useEffect, useMemo } from 'react';
+
+import { router } from '../imperative-api';
+import {
+  type NavigationRoute,
+  type ParamListBase,
+  type RouteSource,
+  useIsFocused,
+} from '../react-navigation/native';
+import { useBuildHref } from './useBuildHref';
+
+type TabRoute = NavigationRoute<ParamListBase, string>;
+
+type TabDescriptor<Options extends object> = {
+  routeSource?: RouteSource;
+  options?: Options;
+};
+
+/**
+ * Returns the visible layout tabs and their focused index. When the navigator is focused, redirects
+ * an unavailable focused route to `redirectToRouteName` or the first visible tab.
+ */
+export function useVisibleTabsWithRedirect<Route extends TabRoute, Options extends object>({
+  routes,
+  focusedRouteKey,
+  descriptors,
+  redirectToRouteName,
+  isHidden,
+}: {
+  routes: Route[];
+  focusedRouteKey: string;
+  descriptors: Record<string, TabDescriptor<Options>>;
+  redirectToRouteName?: string;
+  // Keep this callback reference-stable to avoid recalculating the visible routes.
+  isHidden?: (options: Options | undefined) => boolean;
+}) {
+  const buildHref = useBuildHref();
+  const isFocused = useIsFocused();
+
+  const visibleRoutes = useMemo(
+    () =>
+      routes.filter((route) => {
+        // Every filesystem route is registered in state; only routes declared by a non-hidden
+        // trigger become tab items.
+        const descriptor = descriptors[route.key];
+        return isDeclaredInLayout(descriptor) && !isHidden?.(descriptor?.options);
+      }),
+    [routes, descriptors, isHidden]
+  );
+  const visibleFocusedIndex = useMemo(
+    () => visibleRoutes.findIndex((route) => route.key === focusedRouteKey),
+    [focusedRouteKey, visibleRoutes]
+  );
+
+  const redirectHref = useMemo(() => {
+    const redirectRoute = findRouteByName(visibleRoutes, redirectToRouteName) ?? visibleRoutes[0];
+    if (redirectRoute) {
+      return buildHref(redirectRoute);
+    }
+    return null;
+  }, [buildHref, redirectToRouteName, visibleRoutes]);
+
+  useEffect(() => {
+    // The focused route can be hidden or have no trigger at all — for example a path pointing at a
+    // route without a tab, or a trigger hidden while focused. Redirect to the router's initial tab,
+    // falling back to the first visible tab. `replace` keeps the unreachable route out of history.
+    // TODO(@ubax): Show a formsheet for hidden tabs which are focused (Tabs + Stack in one).
+    if (isFocused && visibleFocusedIndex < 0 && redirectHref != null) {
+      router.replace(redirectHref);
+    }
+  }, [isFocused, redirectHref, visibleFocusedIndex]);
+
+  return { visibleRoutes, visibleFocusedIndex };
+}
+
+function isDeclaredInLayout<Options extends object>(
+  descriptor: TabDescriptor<Options> | undefined
+): boolean {
+  return descriptor?.routeSource === 'layout';
+}
+
+function findRouteByName<Route extends TabRoute>(routes: Route[], name: string | undefined) {
+  return routes.find((route) => route.name === name || route.name.replace(/\/index$/, '') === name);
+}
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -24,6 +24,7 @@
 
 ### 🐛 Bug fixes
 
+- Prevent unfocused nested native tab navigators from redirecting global router state. ([#48257](https://github.com/expo/expo/pull/48257) by [@Ubax](https://github.com/Ubax))
 - Fixed `Tabs` and `TopTabs` (`expo-router/js-tabs`, `expo-router/js-top-tabs`) not being usable from RSC ([#48330](https://github.com/expo/expo/pull/48330) by [@Ubax](https://github.com/Ubax))
 - Redirect fully guarded navigators to parent ([#47984](https://github.com/expo/expo/pull/47984) by [@Ubax](https://github.com/Ubax))
 - Unset `nativeContainerStyle` in transparent presentations ([#48154](https://github.com/expo/expo/pull/48154) by [@Ubax](https://github.com/Ubax))
--- a/packages/expo-router/src/native-tabs/NativeBottomTabsNavigator.tsx
+++ b/packages/expo-router/src/native-tabs/NativeBottomTabsNavigator.tsx
@@ -1,19 +1,16 @@
 'use client';
 
-import React, { use, useCallback, useEffect, useMemo, useRef } from 'react';
+import React, { use, useCallback, useMemo, useRef } from 'react';
 
 import { useRouteNode } from '../Route';
-import { router } from '../imperative-api';
 import type {
   ParamListBase,
   TabNavigationState,
   TabRouterOptions,
 } from '../react-navigation/native';
 import { unstable_createStandardRouterNavigator } from '../standard-navigation';
-import type {
-  StandardNavigatorContentProps,
-  StandardNavigatorDescriptor,
-} from '../standard-navigation/types';
+import type { StandardNavigatorContentProps } from '../standard-navigation/types';
+import { useVisibleTabsWithRedirect } from '../standard-navigation/useVisibleTabsWithRedirect';
 import { getAllChildrenNotOfType, getAllChildrenOfType } from '../utils/children';
 import { NativeBottomTabsRouter } from './NativeBottomTabsRouter';
 import { NativeTabTrigger } from './NativeTabTrigger';
@@ -31,6 +28,7 @@ import { convertIconColorPropToObject, convertLabelStylePropToObject } from './u
 
 // In Jetpack Compose, the default back behavior is to go back to the initial route.
 const defaultBackBehavior = 'initialRoute';
+const isNativeTabHidden = (options: NativeTabOptions | undefined) => options?.hidden === true;
 export const NativeTabsContext = React.createContext<boolean>(false);
 
 function NativeTabsContent({
@@ -66,56 +64,31 @@ function NativeTabsContent({
 
   const { routes } = state;
 
+  const { visibleRoutes, visibleFocusedIndex } = useVisibleTabsWithRedirect({
+    routes,
+    focusedRouteKey: routes[state.index]!.key,
+    descriptors,
+    redirectToRouteName,
+    isHidden: isNativeTabHidden,
+  });
   const visibleTabs = useMemo(
     () =>
-      routes
-        // Every filesystem route is registered in state; only routes declared by a
-        // non-hidden <NativeTabs.Trigger> become tab items.
-        .filter((route) => {
-          const descriptor = descriptors[
-            route.key
-          ]! as StandardNavigatorDescriptor<NativeTabOptions>;
-          return descriptor.routeSource === 'layout' && descriptor.options?.hidden !== true;
+      visibleRoutes.map(
+        (route): NativeTabsViewTabItem => ({
+          options: descriptors[route.key]!.options,
+          routeKey: route.key,
+          name: route.name,
+          contentRenderer: () => descriptors[route.key]!.render(),
         })
-        .map(
-          (route): NativeTabsViewTabItem => ({
-            options: descriptors[route.key]!.options,
-            routeKey: route.key,
-            name: route.name,
-            contentRenderer: () => descriptors[route.key]!.render(),
-          })
-        ),
-    [routes, descriptors]
-  );
-  const visibleFocusedTabIndex = useMemo(
-    () => visibleTabs.findIndex((tab) => tab.routeKey === routes[state.index]!.key),
-    [visibleTabs, routes, state.index]
+      ),
+    [descriptors, visibleRoutes]
   );
   const visibleTabsKeys = useMemo(
     () => visibleTabs.map((tab) => tab.routeKey).join(';'),
     [visibleTabs]
   );
 
-  const focusedIndex = visibleFocusedTabIndex >= 0 ? visibleFocusedTabIndex : 0;
-
-  // The focused route can be hidden or have no trigger at all — for example a path pointing at a
-  // route without a tab, or a trigger hidden while focused. Redirect to the router's initial tab,
-  // falling back to the first visible tab. `replace` keeps the unreachable route out of history.
-  // TODO(@ubax): Show a formsheet for hidden tabs which are focused (Tabs + Stack in one).
-  const redirectTabHref = useMemo(() => {
-    const redirectTab =
-      visibleTabs.find(
-        (tab) =>
-          tab.name === redirectToRouteName ||
-          tab.name.replace(/\/index$/, '') === redirectToRouteName
-      ) ?? visibleTabs[0];
-    return routes.find((route) => route.key === redirectTab?.routeKey)?.href;
-  }, [redirectToRouteName, routes, visibleTabs]);
-  useEffect(() => {
-    if (visibleFocusedTabIndex < 0 && redirectTabHref != null) {
-      router.replace(redirectTabHref);
-    }
-  }, [visibleFocusedTabIndex, redirectTabHref]);
+  const focusedIndex = visibleFocusedIndex >= 0 ? visibleFocusedIndex : 0;
 
   const provenanceRef = useRef(0);
 
--- a/packages/expo-router/src/native-tabs/__tests__/navigation.test.ios.tsx
+++ b/packages/expo-router/src/native-tabs/__tests__/navigation.test.ios.tsx
@@ -4,6 +4,7 @@ import { View } from 'react-native';
 import { Tabs } from 'react-native-screens';
 
 import { router } from '../../imperative-api';
+import Stack from '../../layouts/StackClient';
 import { Link } from '../../link/Link';
 import { renderRouter } from '../../testing-library';
 import { NativeTabs } from '../NativeTabs';
@@ -362,6 +363,41 @@ describe('Native Bottom Tabs trigger changes', () => {
     for (const call of TabsScreen.mock.calls) {
       expect(call[0].screenKey).toMatch(/^index-[-\w]+/);
     }
+  });
+
+  it('waits for a nested native tabs navigator to regain focus before redirecting', () => {
+    let setHidden!: (hidden: boolean) => void;
+    function TabsLayout() {
+      const [hidden, set] = useState(false);
+      setHidden = set;
+      return (
+        <NativeTabs>
+          <NativeTabs.Trigger name="index" />
+          <NativeTabs.Trigger name="second" hidden={hidden} />
+        </NativeTabs>
+      );
+    }
+    renderRouter(
+      {
+        _layout: () => <Stack />,
+        index: () => <View testID="outside" />,
+        'tabs/_layout': TabsLayout,
+        'tabs/index': () => <View testID="tabs-index" />,
+        'tabs/second': () => <View testID="tabs-second" />,
+      },
+      { initialUrl: '/tabs/second' }
+    );
+
+    act(() => router.push('/'));
+    act(() => setHidden(true));
+
+    expect(screen).toHavePathname('/');
+    expect(screen.getByTestId('outside')).toBeVisible();
+
+    act(() => router.back());
+
+    expect(screen).toHavePathname('/tabs');
+    expect(screen.getByTestId('tabs-index')).toBeVisible();
   });
 
   it('renders no tab UI when every trigger is hidden', () => {
--- a/packages/expo-router/src/standard-navigation/__tests__/useVisibleTabsWithRedirect.test.ios.tsx
+++ b/packages/expo-router/src/standard-navigation/__tests__/useVisibleTabsWithRedirect.test.ios.tsx
@@ -1,0 +1,147 @@
+import { renderHook } from '@testing-library/react-native';
+
+import { router } from '../../imperative-api';
+import { useIsFocused } from '../../react-navigation/native';
+import { useBuildHref } from '../useBuildHref';
+import { useVisibleTabsWithRedirect } from '../useVisibleTabsWithRedirect';
+
+jest.mock('../../react-navigation/native', () => {
+  const actualNavigation = jest.requireActual(
+    '../../react-navigation/native'
+  ) as typeof import('../../react-navigation/native');
+  return { ...actualNavigation, useIsFocused: jest.fn() };
+});
+jest.mock('../useBuildHref');
+
+const mockedUseIsFocused = useIsFocused as jest.MockedFunction<typeof useIsFocused>;
+const mockedUseBuildHref = useBuildHref as jest.MockedFunction<typeof useBuildHref>;
+
+const routes = [
+  { key: 'home-key', name: 'home' },
+  { key: 'settings-key', name: 'settings/index' },
+  { key: 'hidden-key', name: 'hidden' },
+  { key: 'filesystem-key', name: 'filesystem' },
+];
+const descriptors = {
+  'home-key': { routeSource: 'layout' as const },
+  'settings-key': { routeSource: 'layout' as const },
+  'hidden-key': { routeSource: 'layout' as const, options: { hidden: true } },
+  'filesystem-key': { routeSource: 'filesystem' as const },
+};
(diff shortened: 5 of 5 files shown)


Ubax:
* **#47986** <a href="https://app.graphite.com/github/pr/expo/expo/47986?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#48257** <a href="https://app.graphite.com/github/pr/expo/expo/48257?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/48257?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#48256** <a href="https://app.graphite.com/github/pr/expo/expo/48256?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/expo-router/src/ui/TabRouter.tsx
```diff
@@ -67,9 +115,14 @@ export function ExpoTabRouter(options: ExpoTabRouterOptions) {
             return { ...r, state: undefined };
           }),
         };
-        return rnTabRouter.getStateForAction(state, action, options);
-      } else {
+        return rnTabRouter.getStateForAction(state, action, routerConfigOptions);
+      } else if (route.state !== undefined) {
```
jakub-agent: [🚨 **critical**] This new branch splits the old single `else` into two, and the `route.state === undefined` case now falls through to `getStateForAction(JUMP_TO)` instead of `getStateForRouteFocus`. That loses params.

The base `TabRouter` computes `params = createParamsFromAction({ action, routeParamList })` for `JUMP_TO` (`react-navigation/routers/TabRouter.tsx:399`) — a full **replace**, not a merge (merging only happens for `NAVIGATE` with `merge: true`). So re-focusing a leaf tab now resets its params to whatever the trigger's href encodes, discarding anything acquired since. `getStateForRouteFocus` preserved `route.params`.

Confirmed by bisecting — this probe passes on the base commit `bf44fe9` and fails here:

```tsx
it('push params survive tab re-press', () => {
  renderRouter({
    _layout: () => (
      <Tabs options={{ backBehavior: 'fullHistory' }}>
        <TabList>
          <TabTrigger name="index" testID="goto-index" href="/" />
          <TabTrigger name="search" testID="goto-search" href="/search" />
        </TabList>
        <TabSlot />
      </Tabs>
    ),
    index: () => null,
    search: function Search() {
      const { q } = useLocalSearchParams();
      return <Text testID="q">{String(q)}</Text>;
    },
  });

  act(() => router.push('/search?q=foo'));
  expect(screen.getByTestId('q')).toHaveTextContent('foo');

  fireEvent.press(screen.getByTestId('goto-index'));
  fireEvent.press(screen.getByTestId('goto-search'));

  expect(screen.getByTestId('q')).toHaveTextContent('foo'); // receives `undefined`
});
```

Also reproduces with `router.setParams({ q: 'foo' })` and with `backBehavior: 'history'`. It needs `history` to still contain the route (so `shouldReset` is false) and `route.state` to be undefined — i.e. any leaf tab under a history-based back behavior. A search tab that keeps its query in the URL is the obvious real-world case.

The existing tests don't catch it because the only leaf-tab re-press tests use the default `firstRoute` behavior, where `getRouteHistory` drops the route from history and `shouldReset` becomes true anyway.

Could the params requirement be met without widening this branch — e.g. keep `getStateForRouteFocus` and apply only the params diff, or reuse the `NAVIGATE`-style merge so existing route params win? Whichever way it goes, a test pinning params across a re-press would be worth adding.

packages/expo-router/src/ui/common.tsx
```diff
@@ -119,11 +119,18 @@ export function triggersToScreens(
...
     }
+    const isWithinLayout = state?.name === targetStateName;
     routeState =
       state!.state?.routes[state!.state.index ?? state!.state.routes.length - 1] || state;
 
     const routeNode = layoutRouteNode.children.find((child) => child.route === routeState?.name);
 
+    if (!isWithinLayout) {
+      throw new Error(
+        `Tab trigger '${trigger.name}' with href '${resolvedHref}' must point to a route within the tabs layout.`
```
jakub-agent: [🚨 **critical**] This throw changes a previously non-fatal situation into a crash, and it fires for a pattern people plausibly ship today: an "escape hatch" trigger inside nested tabs that points at a route in a parent layout.

Before this PR, `/settings` from a `fruit/_layout` tabs navigator walked to the `settings` leaf, failed the `layoutRouteNode.children.find(...)` lookup below, hit `console.warn('Unable to find routeNode for trigger ...')`, and `continue`d — the trigger was skipped and the app kept running. Now it throws during render and takes the screen down.

Confirmed by bisecting — this passes on the base commit `bf44fe9` and throws here:

```tsx
renderRouter(
  {
    _layout: () => <Stack />,
    settings: () => <Text testID="settings">Settings</Text>,
    'fruit/_layout': () => (
      <Tabs>
        <TabList>
          <TabTrigger name="apple" href="/fruit/apple" />
          <TabTrigger name="settings" href="/settings" />
        </TabList>
        <TabSlot />
      </Tabs>
    ),
    'fruit/apple': () => <Text testID="apple">Apple</Text>,
  },
  { initialUrl: '/fruit/apple' }
);
```

Two separate things to decide:

1. **Severity.** Failing loud on a genuinely wrong href is defensible, but it's inconsistent with the `!routeNode` branch 6 lines down, which still only warns for what is arguably the same class of mistake. A dev-only `console.warn` plus skipping the trigger (matching that branch) would flag the bug without breaking production apps. If throwing is intentional, it should be a `🛠 Breaking changes` changelog entry, not `💡 Others`.

2. **Message.** Per the repo's error-message guidance in `.claude/CLAUDE.md`, this covers *what* but not *why* or *how*. Something closer to: "Tab trigger 'settings' has href '/settings', which resolves outside the 'fruit' tabs layout. A trigger can only target a route nested under the layout that declares it. Use a `<Link>` for cross-layout navigation, or move the route under 'fruit'."

packages/expo-router/src/ui/TabRouter.tsx
```diff
@@ -35,9 +42,50 @@ export function ExpoTabRouter(options: ExpoTabRouterOptions) {
     ExpoTabActionType | CommonNavigationAction
   > = {
     ...rnTabRouter,
-    getStateForAction(state, action, options) {
-      if (action.type !== 'JUMP_TO') {
-        return rnTabRouter.getStateForAction(state, action, options);
+    getStateForRouteNamesChange(state, options) {
```
jakub-agent: [🏗️ **design**] This override deserves a comment. It's the single highest-leverage line in the diff and its safety rests on an invariant that lives in a different file.

`useNavigationBuilder.tsx:536-550` is the *only* place the vendored core reconciles state against a changed `<Screen>` set — nothing else adds, removes, or re-keys routes. Making it a no-op is safe here only because `useSortedScreens` now emits a screen for **every** child of the layout route node (trigger-declared ones tagged `routeSource: 'layout'`, the rest `'filesystem'`), so the route-name set is fixed at mount and only the order ever changes.

That invariant is load-bearing and implicit. It's also exactly what a follow-up could break — the moment any code path stops registering every filesystem child (a `redirect` screen prop, `<Protected>` guards, conditional route registration), the set can change, `EXPO_ROUTER_TAB_ORDER_CHANGED` hits its `return state` bail below, and `state.routeNames` desyncs from the rendered screens permanently, with nothing logged.

Suggestions:
- State the invariant in a comment here and why the no-op follows from it.
- Make the bail branch below loud in dev (`console.warn`, or delegate to `rnTabRouter.getStateForRouteNamesChange`) rather than silently returning `state`. Its own comment says "Not this action's job" — but with this override in place, it is no longer anyone's job.

Minor: `options` is unused here, so `getStateForRouteNamesChange(state)` would read cleaner.

packages/expo-router/src/ui/TabRouter.tsx
```diff
@@ -35,9 +42,50 @@ export function ExpoTabRouter(options: ExpoTabRouterOptions) {
...
+      return state;
+    },
+    getStateForAction(state, action, routerConfigOptions) {
+      if (action.type === 'EXPO_ROUTER_TAB_ORDER_CHANGED') {
```
jakub-agent: [💡 suggestion] This branch runs before any `action.target` check and never returns `null`, so it will consume an `EXPO_ROUTER_TAB_ORDER_CHANGED` action regardless of which navigator it was aimed at. The base `TabRouter` opens `getStateForAction` with `if (action.target && action.target !== state.key) return null;` (`react-navigation/routers/TabRouter.tsx:351-353`), and the dispatch in `Tabs.tsx:200` deliberately sets `target: state.key` — so the intent to scope is there, but nothing enforces it.

I couldn't reach it: `navigation.dispatch` always starts at the originating navigator, which handles the action first, so a sibling or child never sees it. But `useOnAction` does fall through to child navigators for unhandled actions, and the reorder handler ignoring both `target` and returning non-null for every input is a thin guarantee to rest on. Adding the standard target check would make the scoping explicit and cost nothing.

packages/expo-router/src/ui/Tabs.tsx
```diff
@@ -175,6 +175,7 @@ export function useTabsWithTriggers(options: UseTabsWithTriggersOptions): TabsCo
     triggerMap,
     id: contextKey,
     initialRouteName,
+    backBehavior: rest.backBehavior ?? (initialRouteName ? 'initialRoute' : undefined),
```
jakub-agent: [💡 suggestion] Worth a comment explaining the coupling — this line only makes sense in light of the ordering change in `common.tsx`. Previously `triggersToScreens` sorted configs with `sortRoutesWithInitial(initialRouteName)`, so the initial route *was* `routes[0]` and the default `firstRoute` landed on it. Now that screens follow trigger order, `initialRoute` is needed to preserve that. Correct, but entirely non-obvious to a future reader.

One edge case: when `initialRouteName` names a route with **no** trigger, `getRouteHistory`'s `initialRoute` branch still resolves it (`routes.findIndex(...)` matches the `routeSource: 'filesystem'` screen), so `history` gets an entry for an unreachable tab. Going back then lands there and `useHiddenTabRedirect` immediately `replace`s to the first visible tab. I checked and the end state is correct — but it costs an extra navigation round-trip where the old `firstRoute` default went straight to the first trigger. Cheap to avoid by resolving `initialRouteName` against trigger-declared routes only.

packages/expo-router/src/ui/common.tsx
```diff
@@ -119,11 +119,18 @@ export function triggersToScreens(
       if (state.name === targetStateName) break;
       state = state.state.routes[state.state.index ?? state.state.routes.length - 1]!;
     }
+    const isWithinLayout = state?.name === targetStateName;
```
jakub-agent: [🐧 nit] The guard is 6 lines below where it's decided, and the two statements in between do work that only matters on the success path: `routeState` dereferences `state!` twice, and `routeNode` performs a lookup that is discarded when the throw fires. Moving the `if (!isWithinLayout) throw` immediately after this line keeps the failure adjacent to its cause and drops the non-null assertions from the reachable path.

packages/expo-router/src/ui/Tabs.tsx
```diff
@@ -184,6 +185,34 @@ export function useTabsWithTriggers(options: UseTabsWithTriggersOptions): TabsCo
...
+    }));
+    // `routeNames` is a fresh array each render; `routeNamesKey` stands in for it.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [routeNamesKey, navigation]);
+
+  useHiddenTabRedirect({
+    routes: state.routes,
+    focusedRouteKey: state.routes[state.index]!.key,
+    descriptors,
+    redirectToRouteName: initialRouteName,
```
jakub-agent: [🐧 nit] The return value is discarded here, so `useHiddenTabRedirect` runs its `visibleRoutes` / `visibleFocusedIndex` `useMemo`s purely for `NativeBottomTabsNavigator`. Not wrong — the two callers legitimately need different parts of the same computation — but the hook now does two jobs while its name advertises one. If the memoized filtering is only ever consumed by NativeTabs, splitting the redirect effect from the visible-route derivation would make both call sites read more honestly.

jakub-agent (changes requested):

PR 2/2 in the stack, on top of #48256 (`_ubax/07-29-fix_replace_action_in_tabs`). Reviewed against that base, not `main`.

The core idea is sound. Now that `useSortedScreens` registers **every** filesystem child as a screen, the route-name *set* is fixed for the lifetime of the navigator and only its *order* changes — which is exactly what makes `EXPO_ROUTER_TAB_ORDER_CHANGED` sufficient and `getStateForRouteNamesChange` redundant. Reusing route objects by name to preserve keys and screen state is the right call, and the new test coverage for reorder/remove/nesting is genuinely good. Extracting `useHiddenTabRedirect` is a clean win; the NativeTabs redirect path stays covered end-to-end by `native-tabs/__tests__/navigation.test.ios.tsx`.

I verified locally at `fd18e3e`: `pnpm test` for `expo-router` is green (4523 passed; the only 3 failures are `__rsc_tests__` asset-path snapshots caused by my worktree location, unrelated to this diff), and `tsc -b tsconfig.all.json` plus `oxlint` are both clean.

Two behavior regressions, both confirmed by bisecting against the base commit `bf44fe9` — a probe test passes there and fails at `fd18e3e`:

1. **Tab params are discarded when a tab trigger is re-pressed.** Routing an already-visited, navigator-less tab through `getStateForAction(JUMP_TO)` replaces `route.params` wholesale, so params the user acquired inside the tab are lost. Reproduced with both `router.setParams` and `router.push('/search?q=foo')`.
2. **A trigger href pointing outside the tabs layout now throws instead of warning.** The adjacent `!routeNode` branch still only warns. This turns a previously-degraded-but-running app into a crash.

Neither is covered by the new tests. The rest are smaller design and documentation points.

packages/expo-router/src/utils/useHiddenTabRedirect.ts
```diff
@@ -0,0 +1,87 @@
+import { useEffect, useMemo } from 'react';
+
+import { router } from '../imperative-api';
+import {
+  type NavigationRoute,
+  type ParamListBase,
+  type RouteSource,
+  useIsFocused,
+} from '../react-navigation/native';
+import { useBuildHref } from '../standard-navigation/useBuildHref';
```
jakub-agent: [🏗️ **design**] Is `src/utils/` the right home for this? Two things point elsewhere:

- `AGENTS.md` describes `utils/` as "General utilities (font, splash, statusbar, url)" and `standard-navigation/` as "Shared navigation state/actions/emitter hooks".
- No file currently in `src/utils/` imports `imperative-api` or `standard-navigation` — they are all leaf helpers. This hook imports both, so a leaf-utility directory starts depending on the router singleton and on the navigator layer.

`standard-navigation/` sits right next to `useBuildHref` (which this consumes) and is already the place both this navigator and the follow-up headless navigator share code through, so it seems like the more natural destination.

packages/expo-router/src/standard-navigation/useVisibleTabsWithRedirect.ts
```diff
@@ -0,0 +1,87 @@
...
+    }
+    return null;
+  }, [buildHref, redirectToRouteName, visibleRoutes]);
+
+  useEffect(() => {
+    // The focused route can be hidden or have no trigger at all — for example a path pointing at a
+    // route without a tab, or a trigger hidden while focused. Redirect to the router's initial tab,
+    // falling back to the first visible tab. `replace` keeps the unreachable route out of history.
+    // TODO(@ubax): Show a formsheet for hidden tabs which are focused (Tabs + Stack in one).
+    if (isFocused && visibleFocusedIndex < 0 && redirectHref != null) {
```
jakub-agent, line 68: [💡 suggestion] The `isFocused` gate is new — on `main` the effect redirected regardless of focus. I think it's the right change (an unfocused navigator shouldn't reach for global `router.replace`), but it makes this more than a move, so it's worth calling out in the PR body.

The observable difference: when `NativeTabs` is nested and its host screen is blurred while the focused tab is hidden, no redirect fires — and `focusedIndex` still falls back to `0`, so the tab bar renders tab 0 as selected while router state points at the hidden route. That state now persists until focus returns, instead of being corrected on the next effect tick.

Worth deciding whether that user-visible change deserves a `changelog.md` entry.

packages/expo-router/src/utils/useHiddenTabRedirect.ts
```diff
@@ -0,0 +1,87 @@
...
+
+type TabRoute = NavigationRoute<ParamListBase, string>;
+
+type TabDescriptor = {
+  routeSource?: RouteSource;
+  options?: object;
```
jakub-agent: [💡 suggestion] Typing `options` as bare `object` drops the type link that the original code had. `main` cast to `StandardNavigatorDescriptor<NativeTabOptions>`, so `descriptor.options?.hidden !== true` was checked against the real option — renaming `hidden` in `NativeTabOptions` would have failed the build. Here it compiles fine and hiding silently stops working.

It's also what forces the `'hidden' in ...` narrowing dance in `isHidden`. Typing it as

```ts
options?: { hidden?: boolean };
```

restores the compile-time check and collapses the predicate to `descriptor?.options?.hidden === true` — no `!!`, no `in`. (One test case, `{ options: { hidden: 1 } }`, would need a cast or to go away; it's testing a state the type system would then rule out.)

packages/expo-router/src/utils/useHiddenTabRedirect.ts
```diff
@@ -0,0 +1,87 @@
...
+    if (isFocused && visibleFocusedIndex < 0 && redirectHref != null) {
+      router.replace(redirectHref);
+    }
+  }, [isFocused, redirectHref, visibleFocusedIndex]);
+
+  return { visibleRoutes, visibleFocusedIndex };
+}
+
+/** @internal Exported only for unit tests. */
+export function isHidden(descriptor: TabDescriptor | undefined): boolean {
```
jakub-agent: [💡 suggestion] These three exports exist only for tests, and the hook's own tests already cover them:

- `isHidden` / `isDeclaredInLayout` — covered by "returns only visible layout routes and their focused index", which excludes both the hidden and the filesystem route.
- `findRouteByName`'s `/index` strip — covered by "redirects an unavailable focused route to the configured visible route", where `redirectToRouteName: 'settings'` resolves to `settings/index`.

Keeping them module-private would drop three exports and ~25 lines of test without losing coverage.

packages/expo-router/src/standard-navigation/useVisibleTabsWithRedirect.ts
```diff
@@ -0,0 +1,87 @@
...
+  focusedRouteKey: string;
+  descriptors: Record<string, TabDescriptor>;
+  redirectToRouteName?: string;
+}) {
+  const buildHref = useBuildHref();
+  const isFocused = useIsFocused();
+
+  const visibleRoutes = useMemo(
+    () =>
+      routes.filter((route) => {
```
jakub-agent, line 42: [🐧 nit] The comment explaining *why* this filter exists was dropped in the move:

> Every filesystem route is registered in state; only routes declared by a non-hidden `<NativeTabs.Trigger>` become tab items.

That's the least obvious part of the hook — that state holds every filesystem route and the layout declaration is what promotes one to a tab. Worth carrying over, especially now that the code is one indirection away from its call site.

packages/expo-router/src/utils/useHiddenTabRedirect.ts
```diff
@@ -0,0 +1,87 @@
...
+
+type TabRoute = NavigationRoute<ParamListBase, string>;
+
+type TabDescriptor = {
+  routeSource?: RouteSource;
+  options?: object;
+};
+
+/**
+ * Redirects hidden tabs to `redirectToRouteName`
```
jakub-agent: [🐧 nit] The doc comment describes only the side effect, but the return value (`visibleRoutes`, `visibleFocusedIndex`) is what both call sites primarily consume — the name `useHiddenTabRedirect` reads like a `void` hook. Since a second navigator is about to depend on this, consider naming it for the projection (e.g. `useVisibleTabs`) and documenting the redirect as its side effect, along with the focus gate and the fallback to the first visible route.

jakub-agent (commented):

Clean extraction. I traced the two non-obvious substitutions and both hold up:

- **`route.href` → `useBuildHref()`** is equivalent. `useStandardState` already produced `href` via `buildHref(route)`, and the only provider of `NavigationFocusedRouteStateContext` is `SceneView` — which sits *below* `NativeTabsContent`. Both call sites therefore read the same path context, so the computed href is identical.
- **`useIsFocused()` at root** does not throw. With no parent `SceneView`, `useNavigation()` falls back to the container ref, whose `isFocused: () => true` and emitter-provided `addListener` keep the hook safe (`BaseNavigationContainer.tsx:195`).

The filter predicate is also behavior-identical (`options?.hidden !== true` ↔ `!isHidden(...)`), and the existing integration suite in `native-tabs/__tests__/navigation.test.ios.tsx` already covers hidden triggers, dynamically-hidden focused triggers, deep links to trigger-less routes, and `initialRouteName` pointing at a directory index route — so the extraction is well protected.

No correctness bugs found. My comments are about where the file lives, one undocumented behavior change, and the descriptor typing.

packages/expo-router/src/standard-navigation/useVisibleTabsWithRedirect.ts
```diff
@@ -0,0 +1,83 @@
...
+  focusedRouteKey,
+  descriptors,
+  redirectToRouteName,
+  isHidden,
+}: {
+  routes: Route[];
+  focusedRouteKey: string;
+  descriptors: Record<string, TabDescriptor<Options>>;
+  redirectToRouteName?: string;
+  isHidden?: (options: Options | undefined) => boolean;
```
jakub-agent, line 35: [💡 suggestion] Inverting control here is the right call, but making the callback *optional* keeps a bit of the silent-failure mode the previous round was aimed at. With `!isHidden?.(...)`, a caller that forgets the prop gets `!undefined` → `true` for every route, so nothing is ever hidden — no compile error, no runtime signal, `hidden` triggers just quietly render as normal tabs.

Making it required would turn that into a build failure. A consumer with no hidden concept then writes `isHidden: () => false`, which states the intent explicitly rather than relying on the default.

If it's optional because the follow-up headless navigator genuinely has no `hidden` option, feel free to ignore — but in that case a short note on the default behavior would help.

packages/expo-router/src/standard-navigation/useVisibleTabsWithRedirect.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  const visibleRoutes = useMemo(
+    () =>
+      routes.filter((route) => {
+        // Every filesystem route is registered in state; only routes declared by a non-hidden
+        // trigger become tab items.
+        const descriptor = descriptors[route.key];
+        return isDeclaredInLayout(descriptor) && !isHidden?.(descriptor?.options);
+      }),
+    [routes, descriptors, isHidden]
```
jakub-agent, line 48: [🐧 nit] `isHidden` being a dependency means the hook now requires a referentially stable callback, which isn't stated anywhere. Defining `isNativeTabHidden` at module scope in `NativeBottomTabsNavigator` satisfies it, and I assume that placement was deliberate — but the next consumer can't tell from the signature.

An inline `isHidden={(o) => ...}` would invalidate this memo every render, cascading through `visibleFocusedIndex`, `redirectHref`, and `visibleTabs` at the call site. Nothing breaks (the values stay equal, and `buildHref` is WeakMap-cached per route), but the memoization silently stops doing its job.

One line on the doc comment — "`isHidden` must be stable across renders" — would save the next author the trip.

jakub-agent (commented):

**Iteration 2** — re-reviewed at `aa4e4eed`. All six points from the previous round are addressed, and two of them were solved better than I proposed:

| Previous point | Resolution |
| --- | --- |
| `src/utils/` placement | Moved to `standard-navigation/useVisibleTabsWithRedirect.ts`. No cycle — `imperative-api` only reaches `global-state/router`. |
| Undocumented focus-gate behavior change | CHANGELOG entry under 🐛 Bug fixes **plus** a real integration test. |
| `options?: object` lost the type link | Solved by inverting control: generic `TabDescriptor<Options>` + caller-supplied `isHidden`. More general than the `{ hidden?: boolean }` I suggested, and it keeps the `hidden` key owned by `native-tabs` where it belongs. |
| Test-only exports | `isHidden` is gone entirely, `isDeclaredInLayout` / `findRouteByName` are module-private, and the three `describe` blocks are gone. Coverage is retained through the hook's own tests. |
| Dropped filter comment | Carried into the filter. |
| Name / doc comment | Renamed, and the doc now describes both the return value and the redirect. |

I checked that the new integration test actually discriminates: without the `isFocused` gate, `setHidden(true)` on the blurred navigator would `router.replace` while `/` is focused, so `toHavePathname('/')` would fail. It is a genuine regression test, not just a happy path. Passing `isNativeTabHidden` from module scope rather than inline also keeps the new `useMemo` dependency stable — a nice detail.

CI is green across `jest-ubuntu`, `jest-windows`, `check-packages`, and the build jobs, so the new generic inference compiles and the tests pass.

No blocking concerns. The two comments below are optional polish on the new `isHidden` parameter.

packages/expo-router/src/standard-navigation/useVisibleTabsWithRedirect.ts
Ubax: It's intentionally optional

jakub-agent (commented):

**Iteration 3** — re-reviewed at `713cde53`. Nothing new to flag.

The only change to this PR's own files since `aa4e4eed` is one line — the stability note on `isHidden`, which resolves the `nit` from the last round:

```ts
// Keep this callback reference-stable to avoid recalculating the visible routes.
isHidden?: (options: Options | undefined) => boolean;
```

The other suggestion — making `isHidden` required — was left as-is. That's a reasonable call, and the comment now documents the contract that made it worth raising.

Things I checked on this revision rather than assuming:

- **The rebase is inert for this change.** The compare against the previous head touches no other file under `packages/expo-router/src`, so `useBuildHref`, `standard-navigation/types.ts`, and the vendored `useIsFocused` are all untouched. Nothing shifted underneath the hook.
- **The CHANGELOG entry survived the rebase in the right place** — still under `## Unpublished` → `### 🐛 Bug fixes`, not absorbed into a published section.

One caveat: CI had only just started on `713cde53` when I looked, so unlike the previous round I can't report it green. The last full run on `aa4e4eed` passed `jest-ubuntu`, `jest-windows`, and `check-packages`, and the only delta since then is a comment line, so I'd expect the same result — worth a glance before merging.

Looks good to me. No blocking concerns.

jakub-agent (approved):

**Approving** at `713cde53`. All eight points raised across the three review rounds are resolved or consciously declined, and the behavior-preservation questions I had about the extraction are settled:

- `route.href` → `useBuildHref()` is equivalent — `useStandardState` already built `href` the same way, and the only provider of `NavigationFocusedRouteStateContext` (`SceneView`) sits below the call site, so both read the same path context.
- `useIsFocused()` is safe at root — with no parent `SceneView`, `useNavigation()` falls back to the container ref, whose `isFocused: () => true` and emitter-provided `addListener` keep it working.
- The filter predicate is behavior-identical, and the one intentional change (the focus gate) now has a CHANGELOG entry plus an integration test that genuinely discriminates it — without the gate, `toHavePathname('/')` would fail.
- The final rebase touches no other file under `packages/expo-router/src`, so nothing shifted underneath the hook.

CI on this head: `jest-ubuntu` (all four shards), `check-packages`, `build`, and `code_review` pass. `jest-windows` and the native builds were still running when I approved — the ubuntu jest shards are what cover the new tests, so I'm comfortable, but the remaining jobs are worth a glance before merge.

Issue #48131, opened 2026-07-26
[expose-ui][jetpack-compose] RNHostView drops onPress events when used within a LazyRow


BrianUribe6:
### Minimal reproducible example

https://github.com/BrianUribe6/RNHostview-pressable-bug-reproduction

### Steps to reproduce

1. **Use a real android device**. I was unable to reproduce it on the emulator likely because it might be related to scrolling + pressing the button.
2. Click on any of the provided buttons on the example (home screen). **Case 1**  is the problem and will fail. **Case 2** is a workaround. **Case 3** works as everything is running on Jetpack Compose.

<img width="270" height="606" alt="Image" src="https://github.com/user-attachments/assets/37bd43f9-2be7-48d3-8c97-b3ed5abddce2" />

The problematic code is  the following:

```tsx
<Host>
  <LazyRow>
  <RNHostView>
    <Pressable onPress={() => console.log("Pressed!")}>
      <RNText>Button</RNText>
    </Pressable>
  </RNHostView>
  </LazyRow>
</Host>
```
If you press the button `onPress` would never get called unless you perfectly press it with 0 horizontal or vertical movement.

### Environment

```text
expo-env-info 2.1.0 environment info:
    System:
      OS: Linux 7.0 Arch Linux
      Shell: 5.3.9 - /usr/bin/bash
    Binaries:
      Node: 26.2.0 - /home/ghareon/.local/share/mise/installs/node/26.2.0/bin/node
      npm: 11.13.0 - /home/ghareon/.local/share/mise/installs/node/26.2.0/bin/npm
    SDKs:
      Android SDK:
        API Levels: 36, 36
        Build Tools: 35.0.0, 36.0.0, 36.1.0, 37.0.0
        System Images: android-37.0 | 16 KB Page Size Google Play Intel x86_64 Atom
    IDEs:
      Android Studio: AI-261.23567.138.2611.15613797
    npmPackages:
      expo: ~57.0.8 => 57.0.8 
      expo-router: ~57.0.8 => 57.0.8 
      react: 19.2.3 => 19.2.3 
      react-dom: 19.2.3 => 19.2.3 
      react-native: 0.86.0 => 0.86.0 
      react-native-web: ~0.21.0 => 0.21.2 
    Expo Workflow: managed
```

### Expo Doctor Diagnostics

```text
20/20 checks passed. No issues detected!
```

PR #48259, opened 2026-07-29
[ui][ios] Fix presses dropping on RN content hosted in RNHostView

intergalacticspacehighway:
# Why

Fixes the iOS side of https://github.com/expo/expo/issues/48131. Android was fixed in #48217.

We put a custom `RCTSurfaceTouchHandler` in `RNHostView`. The role of this handler is to dispatch touch events to JS so press handlers work. This is required for two reasons.

1. Touch events in a SwiftUI bottom sheet are not passed to JS, as iOS mounts the sheet in a separate view hierarchy (a presented view controller) outside the RN root view, so the root touch handler never sees those touches.
2. Touch events received by JS have wrong `pageX`/`pageY`, as `RNHostView`s are positioned by SwiftUI instead of Yoga, so `Pressable` events get cancelled on any finger movement. `pageX`/`pageY` are anchored to the view the dispatching handler is attached to: https://github.com/facebook/react-native/blob/v0.86.0/packages/react-native/React/Fabric/RCTSurfaceTouchHandler.mm#L60

# How

- We attach a custom touch handler and add the `RootNodeKind` trait to `RNHostView` (via the `layoutRoot` prop from #48217). This makes it behave like `Modal` in RN, so touches and `measure()` share the host's coordinate space.
- We further customize the handler by overriding `canBePreventedByGestureRecognizer` so RN's surface-level touch handlers can't preempt it (the stock implementation yields to any recognizer outside its own subtree: https://github.com/facebook/react-native/blob/v0.86.0/packages/react-native/React/Fabric/RCTSurfaceTouchHandler.mm#L385), and we disable ancestor surface handlers while a hosted gesture is in flight. This is required because `RNHostView` can sit in a regular UIKit/SwiftUI hierarchy in the same window, so all the touch handlers can fire, and the parent handlers would carry wrong `pageX`/`pageY`.
- Parent handlers are re-enabled once the gesture completes (touch end/cancel, `reset`, detach, dealloc), so the root surface never stays blocked.
- The same pattern is used on Android's `RNHostView`. There `onChildStartedNativeGesture` silences the parent handlers.
- Add a hosted `Pressable` in `LazyHStack` regression example.

# Test Plan

- bare-expo NCL: Put a Pressable in RNHostView in a SwiftUI hierarchy -> press in on a button -> move finger out of the view -> bring finger back on the view and press out, the press should fire.

# TLDR

This makes `RNHostView` mimic the pattern used by RN Modal. One difference with RN Modal is that we need to Host the view in same hierarchy vs in a separate window, so we need to also disable the existing surface touch handlers.

### DIFF
--- a/apps/native-component-list/src/screens/UI/HostingRNViewsScreen.ios.tsx
+++ b/apps/native-component-list/src/screens/UI/HostingRNViewsScreen.ios.tsx
@@ -6,14 +6,19 @@ import {
   List,
   Section,
   RNHostView,
+  ScrollView,
+  LazyHStack,
 } from '@expo/ui/swift-ui';
 import { frame, padding } from '@expo/ui/swift-ui/modifiers';
 import { useState } from 'react';
 import { Text as RNText, View, Pressable } from 'react-native';
 
+const LAZY_STACK_BUTTONS = ['Button One', 'Button Two', 'Button Three', 'Button Four'];
+
 export default function HostingRNViewsScreen() {
   const [counter, setCounter] = useState(0);
   const [boxSize, setBoxSize] = useState(200);
+  const [lastPress, setLastPress] = useState<{ label: string; count: number } | null>(null);
 
   return (
     <Host style={{ flex: 1 }}>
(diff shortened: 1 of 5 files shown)


zoontek (approved):

intergalacticspacehighway:
Fixed by https://github.com/expo/expo/pull/48969/

PR #48260, opened 2026-07-29
[test-core] Fix package.json formatting

tsapeta:
# Why

`et check-packages expo-modules-test-core` fails on `main`. The `format` task reports a format issue in `package.json`.

# How

PR #48224 added a `peerDependencies` block above `devDependencies`. The `oxfmt` check requires these keys in sorted order. The `check-packages` check [failed on that PR](https://github.com/expo/expo/actions/runs/30357068910/job/90267633336) with `expo-modules-test-core#format`, but it was merged anyway. This PR runs `pnpm oxfmt package.json` in the package, which moves `peerDependencies` below `devDependencies`.

# Test Plan

`et check-packages expo-modules-test-core` now passes locally:

```
🏁 All checks passed.
```

### DIFF
--- a/packages/expo-modules-test-core/package.json
+++ b/packages/expo-modules-test-core/package.json
@@ -39,11 +39,11 @@
     "xml-js": "^1.6.11",
     "yaml": "^2.8.3"
   },
-  "peerDependencies": {
-    "expo": "*"
-  },
   "devDependencies": {
     "@types/jest": "^29.2.1",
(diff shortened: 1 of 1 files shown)


kitten (approved):

kitten:
cc @hassankhan: This might mean that we enforce the formatting in a check but don't auto-apply it during commits, which is a misalignment. I think we could either apply pnpm's formatting (on install) and leave it alone otherwise, or align the former

PR #48267, opened 2026-07-29
feat(cli): Support `Forwarded` header as canonical host throughout

kitten:
# Why

Stacked on #47255 (follow-up)
Related to #48236

After the prior PR that adds the `Forwarded` header and delivers relative manifest URLs, we have to address remaining cases that continue to serve either an absolute URL or an authority/host and can't be updated (or have to react to proxies explicitly)

This essentially means carrying `Forwarded` support through to more places in the CLI.

# How

- Implement `Forwarded` parser, falling back to `X-Forwarded-Host` / `X-Forwarded-Proto`
  - NOTE: There's support for the latter headers already in some places, but this wasn't consistent and now also needs to be updated with the former
- Include `Forwarded` reported host in CORS check (allow as origin, since `Origin` might differ from `Forwarded` due to proxies)
- Carry `Forwarded` through to manifest middlewares and redirect middlewares
- Adjust the reported `debuggerHost` in the manifest according to `Forwarded` header

# Test Plan

- Unit tests added
- Manually test `POST /manifest 'expo-platform: ios'` requests with `Forwarded` headers, altering the outputs

# Checklist

### DIFF
--- a/packages/@expo/cli/src/start/server/UrlCreator.ts
+++ b/packages/@expo/cli/src/start/server/UrlCreator.ts
@@ -5,6 +5,7 @@ import * as Log from '../../log';
 import type { GatewayInfo } from '../../utils/ip';
 import { getGateway, getGatewayAsync } from '../../utils/ip';
 import { debugEvent } from './events';
+import type { ForwardedRequestInfo } from './middleware/resolveForwarded';
 
 export interface CreateURLOptions {
   /** URL scheme to use when opening apps in custom runtimes. */
@@ -13,6 +14,8 @@ export interface CreateURLOptions {
   hostType?: 'localhost' | 'lan' | 'tunnel';
   /** Requested hostname. */
   hostname?: string | null;
+  /** Address the client used to reach the dev server, from a forwarded request */
+  forwarded?: ForwardedRequestInfo | null;
 }
 
 interface UrlComponents {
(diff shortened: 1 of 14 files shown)


packages/@expo/cli/src/start/server/middleware/resolveForwarded.ts
```diff
@@ -0,0 +1,95 @@
...
+
+function coerceProtocol(protocol: string | undefined): 'http' | 'https' | undefined {
+  // NOTE(@kitten): We should never receive "exp" or "exps" here, but account for them anyway for completeness
+  switch (protocol?.toLowerCase()) {
+    case 'http':
+    case 'exp':
+      return 'http';
+    case 'https':
+    case 'exps':
+      return 'http';
```
ramonclaudio: Should we return `https` here?

packages/@expo/cli/src/start/server/middleware/resolveForwarded.ts
```diff
@@ -0,0 +1,95 @@
+import type { ServerRequest } from './server.types';
+
+/** How a client reached this dev server, as reported by the request itself */
+export interface ForwardedRequestInfo {
+  authority: string | undefined;
+  protocol: 'http' | 'https' | undefined;
+}
```
ramonclaudio, line 7: Should `authority` and `protocol` be optional here?

packages/@expo/cli/src/start/server/UrlCreator.ts
```diff
@@ -151,6 +161,23 @@ export class UrlCreator {
...
+  const { authority } = forwarded;
+  if (!authority) {
+    return null;
+  }
+  const scheme = options.scheme ?? forwarded.protocol ?? 'http';
+  const parsed = new URL(`${scheme}://${forwarded.authority}`);
+  return {
+    hostname: parsed.hostname,
+    port: parsed.port,
+    protocol: scheme,
```
ramonclaudio, line 177: Should this work without a scheme? I might be missing something, but could this end up trying to parse `://proxy.test` instead of returning `proxy.test`?

ramonclaudio (approved):

packages/@expo/cli/src/start/server/middleware/resolveForwarded.ts
kitten: Nice catch!

packages/@expo/cli/src/start/server/middleware/resolveForwarded.ts
kitten: No, I made it required, so I don't forget to add them and generally prefer this sometimes as an agent guard rail

packages/@expo/cli/src/start/server/UrlCreator.ts
kitten: This doesn't always carry a known scheme due to the defaults and the protocol. The legacy dev client hence may not send enough information for us to have an explicit value here, and we usually then assume `http`

PR #48269, opened 2026-07-29
[expo] Remove tsapeta from codemention.yml

tsapeta:
# Why

Removes me from all CodeMention rules. The automatic subscriptions have been more annoying than useful for me, and people usually know when and where I should be requested for review anyway.

# How

Deleted the `tsapeta` handle from every `mentions` list in `.github/codemention.yml` (9 rules). Every affected rule still has at least one other person mentioned.

# Test Plan

Verified with `grep` that no `tsapeta` entry remains in the file and that no rule ended up with an empty mentions list.


zoontek (approved):

alanjhughes (approved):
😂

PR #48273, opened 2026-07-29
[docs] Add internal link checker, GH workflow, and remove remark link linting

amandeepmittal:
# Why

The `lint-links` remark linter parses markdown source, so it cannot see links produced by JSX components, generated API reference data, or the build pipeline, which is where most link rot happens. This replaces it with a checker that validates the exported site itself.

# How

- Add `checks/internal-links/` and a `pnpm check-internal-links` script that walks the static export in `out/` and reports broken page links, dead anchors (same-page and cross-page), dangling `public/_redirects` entries, and links relying on redirects, grouped per source page in the terminal.
- Scan guides and unversioned SDK docs as sources, keep versioned copies as valid link targets, and skip anchor validation for links carrying query params (scene-driven pages render variants client-side).
- Make the checker report-only (always exits 0); `--page` filters to one page, `--issue-body`/`--report` write files for the CI job coming in a follow-up PR.
- Fix `remark-link-rewrite` to collapse extensionless `./index` links to the parent directory URL, healing 34 broken `/sdk/router/index` links across all versions.
- Remove the `lint-links` script, its steps in `docs.yml` and `docs-pr.yml`, and the now-unused `remark-cli` and `remark-validate-links` dependencies.
- Add GitHub Actions workflow to open an issue every week.

# Test Plan

Run `pnpm export && pnpm check-internal-links` in `docs/` and confirm the per-page report prints and the command exits 0.

The `checks/internal-links/` and `mdx-plugins/remark-link-rewrite` jest suites cover the behavior (79 tests).

_**Note**: If you run the script and see two links reported in app config schema, they are going to be fixed after https://github.com/expo/universe/pull/29507 gets merged._

# Checklist

- [ ] I added a `changelog.md` entry and rebuilt the package sources according to [this short guide](https://github.com/expo/expo/blob/main/CONTRIBUTING.md#-before-submitting).


entiendoNull (approved):
Running the command I saw it reported 4 broken anchors, but there's only two in the list.

I had Claude analyze this, and the are 3 occurences of one item and one of the other. Just verifying this report aligns with the intention.

Image

amandeepmittal:
> Running the command I saw it reported 4 broken anchors, but there's only two in the list.
> 
> I had Claude analyze this, and the are 3 occurences of one item and one of the other. Just verifying this report aligns with the intention.
> 
> Image

Fix applied:

CleanShot 2026-07-31 at 16 12 01@2x

Links count will be shown if the same link is used and broken multiple times on a page.

entiendoNull (approved):

docs/checks/internal-links/scan.ts
```diff
@@ -0,0 +1,175 @@
...
+  if (htmlFiles.length === 0) {
+    throw new Error('No HTML pages found in out/. Did the build run?');
+  }
+
+  const pages = new Set(htmlFiles.map((file: string) => pagePathFromHtmlFile(outDir, file)));
+  const files = new Set(findAllFiles(outDir, outDir));
+  const redirects = parseRedirects(fs.readFileSync(redirectsFile, 'utf-8'));
+  const sets: ValidSets = { pages, files, redirects };
+
+  const dangling = validateRedirectTargets(redirects, pages);
```
kadikraman: Need to pass `files` into `validateRedirectTargets` and accept `files.has(destination)`, else the file redirects added in the llms PR will be marked as dangling

.github/workflows/docs-link-health-reporter.yml
```diff
@@ -0,0 +1,60 @@
...
+        run: |
+          title="$(pnpm --silent check-internal-links --issue-body "$RUNNER_TEMP/issue-body.md")"
+          echo "title=$title" >> "$GITHUB_OUTPUT"
+      - name: 🐛 Open issue
+        if: steps.report.outputs.title != ''
+        working-directory: docs
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh issue create \
```
kadikraman, line 57: Just flagging that this will always create a new issue even if the previous week's one is not yet closed. We could add logic here to check for and update existing issue in this case, but it's not very important if we're reasonably confident these will get addressed within the week.

.github/workflows/docs-link-health-reporter.yml
```diff
@@ -0,0 +1,60 @@
...
+            --repo "$GITHUB_REPOSITORY" \
+            --title "${{ steps.report.outputs.title }}" \
```
kadikraman: Should add `steps.report.outputs.title` to `env` as well.

.github/workflows/docs-link-health-reporter.yml
```diff
@@ -0,0 +1,60 @@
...
+
+permissions:
+  contents: read
+  issues: write
+
+jobs:
+  report:
+    if: github.repository == 'expo/expo'
+    runs-on: ubuntu-24.04
+    steps:
```
kadikraman, line 25: Should add a timeout here just in case anything goes wrong, e.g. `timeout-minutes: 45`. The default when left unset ie 6 hours.

kadikraman (commented):
The `docs/README.md` needs an update as it still references the old `lint-links` script.

.github/workflows/docs-link-health-reporter.yml
amandeepmittal: I think it is fine to open a new issue everytime. We should clear any wrong links since they are high priority with agents scraping our docs these days.

kadikraman (approved):

PR #48275, opened 2026-07-29
fix(expo,expo-go,linking,asset,dev-menu): Prefer bundle URL origin over `hostUri` and `debuggerHost` from manifest

kitten:
# Why

Stacked on #47255 (follow-up)

We have leftover cases from the manifest update that switches to relative URLs where entries point to a fixed dev server host. These are typically `hostUri` and `debuggerHost`. A few remaining places in the codebase treat these as authoritative, handing control to the dev server on which host to call instead of taking this on themselves.

Like the parent PR, this is problematic, since we have a URL we got the manifest/bundle from already, derived from the request itself, but then disregard it and try to use what the manifest gives us.

`hostUri` and `debuggerHost` are often derived from `X-Forwarded-Host` (or other signals) and once #48267 is merged, will be more accurate in the presence of proxies, however, if we want to behave more like web browsers, it's more correct to simply assume that in development we should use the bundle origin directly. (Since that's where the manifest came from)

# How

- Expose `getBundleOrigin` from `expo/internal/bundle-origin`
  - **Note:** This just exposes what we have for `import.meta.url`, but this isn't shared before
  - I'd love to use `window.location.url` or another global, but that's tied to `@expo/metro-runtime` currently, which also has different logic for this (which we may want to update)
- Use `getBundleOrigin` in `expo-linking` to prefer it there
- Use `getBundleOrigin` in `expo-asset` to prefer it there
- Update `expo-dev-menu` to prefer bundle URL origin over `debuggerHost`
- Update `expo-go` to prefer bundle URL origin over `debuggerHost`
- Fix `expo-go`'s `EXPackagerLogHelper` applying host separately from port

# Test Plan

- Validated a `test-suite` start-up works as expected

# Checklist

### DIFF
--- a/apps/expo-go/android/expoview/src/main/java/host/exp/exponent/experience/ExperienceActivity.kt
+++ b/apps/expo-go/android/expoview/src/main/java/host/exp/exponent/experience/ExperienceActivity.kt
@@ -196,8 +196,9 @@ open class ExperienceActivity : BaseExperienceActivity(), StartReactInstanceDele
           override fun onManifestCompleted(manifest: Manifest) {
             lifecycleScope.launch {
               try {
-                val bundleUrl = ExponentUrls.toHttp(
-                  ExponentUrls.resolveManifestUrl(manifest.getBundleURL(), this@ExperienceActivity.manifestUrl!!)
+                val bundleUrl = ExponentUrls.bundleUrlFromManifest(
+                  manifest,
+                  this@ExperienceActivity.manifestUrl!!
                 )
                 setManifest(
                   this@ExperienceActivity.manifestUrl!!,
@@ -645,7 +646,7 @@ open class ExperienceActivity : BaseExperienceActivity(), StartReactInstanceDele
     Exponent.instance
       .testPackagerStatus(
         isDebugModeEnabled,
-        manifest!!,
+        ExponentUrls.bundleUrlFromManifest(manifest!!, manifestUrl!!),
         object : Exponent.PackagerStatusCallback {
           override fun onSuccess() {
             reactHost = startReactInstance(
--- a/apps/expo-go/ios/Exponent/DevMenu/DevMenuManager.swift
+++ b/apps/expo-go/ios/Exponent/DevMenu/DevMenuManager.swift
@@ -169,13 +169,14 @@ public class DevMenuManager: NSObject {
   func openJSInspector() {
     // The bundle URL is the address this device reached the development server on, unlike the
     // manifest URL, which carries the `exp` scheme and may omit the port the bundle was served on.
-    guard let bundleURL = currentBundleURL, let host = bundleURL.host else {
+    guard let bundleURL = currentBundleURL else {
       return
     }
+    let isServed = bundleURL.scheme == "http" || bundleURL.scheme == "https" || bundleURL.scheme == "exps" || bundleURL.scheme == "exp"
     var components = URLComponents()
-    components.scheme = bundleURL.scheme == "https" ? "https" : "http"
-    components.host = host
-    components.port = bundleURL.port
+    components.scheme = bundleURL.scheme == "https" || bundleURL.scheme == "exps" ? "https" : "http"
+    components.host = isServed ? bundleURL.host : "localhost"
+    components.port = isServed ? bundleURL.port : 8081
     components.path = "/_expo/debugger"
     components.queryItems = [
       URLQueryItem(name: "applicationId", value: Bundle.main.bundleIdentifier ?? "")
--- a/apps/expo-go/ios/Exponent/DevMenu/DevMenuManager.swift
+++ b/apps/expo-go/ios/Exponent/DevMenu/DevMenuManager.swift
@@ -167,13 +167,20 @@ public class DevMenuManager: NSObject {
   }
 
   func openJSInspector() {
-    guard let manifestURL = currentManifestURL else {
+    // The bundle URL is the address this device reached the development server on, unlike the
+    // manifest URL, which carries the `exp` scheme and may omit the port the bundle was served on.
+    guard let bundleURL = currentBundleURL, let host = bundleURL.host else {
       return
     }
-    let port = manifestURL.port ?? 8081
-    let host = manifestURL.host ?? "localhost"
-    let openURL = "http://\(host):\(port)/_expo/debugger?applicationId=\(Bundle.main.bundleIdentifier ?? "")"
-    guard let url = URL(string: openURL) else { return }
+    var components = URLComponents()
+    components.scheme = bundleURL.scheme == "https" ? "https" : "http"
+    components.host = host
+    components.port = bundleURL.port
+    components.path = "/_expo/debugger"
+    components.queryItems = [
+      URLQueryItem(name: "applicationId", value: Bundle.main.bundleIdentifier ?? "")
+    ]
+    guard let url = components.url else { return }
     let request = NSMutableURLRequest(url: url)
     request.httpMethod = "PUT"
     URLSession.shared.dataTask(with: request as URLRequest).resume()
--- a/apps/expo-go/ios/Exponent/Kernel/DevSupport/EXPackagerLogHelper.swift
+++ b/apps/expo-go/ios/Exponent/Kernel/DevSupport/EXPackagerLogHelper.swift
@@ -121,15 +121,18 @@ import React
   }
 
   private func createSocket() {
-    let serverHost = bundleURL.host ?? "localhost"
-    let serverPort = bundleURL.port ?? Int(kRCTBundleURLProviderDefaultPort)
     let scheme = (bundleURL.scheme == "exps" || bundleURL.scheme == "https") ? "https" : "http"
 
     var components = URLComponents()
-    components.host = serverHost
     components.scheme = scheme
-    components.port = serverPort
     components.path = "/hot"
+    if let serverHost = bundleURL.host {
+      components.host = serverHost
+      components.port = bundleURL.port
+    } else {
+      components.host = "localhost"
+      components.port = Int(kRCTBundleURLProviderDefaultPort)
+    }
 
     if let url = components.url {
       socket = SRWebSocket(url: url)
--- a/apps/expo-go/ios/Exponent/Versioned/Core/EXVersionManagerObjC.mm
+++ b/apps/expo-go/ios/Exponent/Versioned/Core/EXVersionManagerObjC.mm
@@ -268,6 +268,7 @@ RCT_EXTERN void EXRegisterScopedModule(Class, ...);
 
 - (void)_openJsInspector:(NSURL *)bundleURL
 {
+  // TODO(@kitten): This might be unreachable. If it isn't it shouldn't hardcode `http`, but it looks unused to me
   NSInteger port = [[bundleURL port] integerValue] ?: RCT_METRO_PORT;
   NSString *host = [bundleURL host] ?: @"localhost";
   NSString *url =
--- a/packages/expo-asset/src/__tests__/AssetSources-test.ts
+++ b/packages/expo-asset/src/__tests__/AssetSources-test.ts
@@ -1,5 +1,7 @@
 import { Platform } from 'expo-modules-core';
 
+jest.mock('expo/internal/bundle-origin', () => ({ getBundleOrigin: jest.fn(() => null) }));
+
 const mockFontMetadata = {
   hash: 'cafecafecafecafecafecafecafecafe',
   name: 'test',
@@ -17,6 +19,7 @@ const mockFontMonorepoMetadata = {
 
 describe('selectAssetSource', () => {
   beforeEach(() => {
+    _mockDevServerUrl(null);
     _mockConstants({
       experienceUrl: 'https://example.com/app/expo-manifest.json',
       __unsafeNoWarnManifest2: {},
@@ -360,5 +363,5 @@ function _mockConstants(constants: { [key: string]: any }): void {
 
 /** Mock the origin the running bundle was loaded from. */
 function _mockDevServerUrl(origin: string | null): void {
-  jest.doMock('expo/internal/bundle-origin', () => ({ getBundleOrigin: () => origin }));
+  jest.mocked(require('expo/internal/bundle-origin').getBundleOrigin).mockReturnValue(origin);
 }
--- a/packages/expo-asset/src/__tests__/AssetSources-test.ts
+++ b/packages/expo-asset/src/__tests__/AssetSources-test.ts
@@ -195,6 +195,86 @@ describe('selectAssetSource', () => {
       );
       expect(source.hash).toBe('cafecafecafecafecafecafecafecafe');
     });
+
+    it(`returns a development URI from the bundle URL rather than debuggerHost`, () => {
+      // The dev server may be reached through an address it can't observe itself, so the URL the
+      // bundle was loaded from wins over `debuggerHost`, which may name an unreachable address.
+      _mockConstants({
+        experienceUrl: 'https://proxy.test/app',
+        __unsafeNoWarnManifest2: {
+          extra: {
+            expoGo: {
+              developer: { tool: 'expo-cli' },
+              debuggerHost: '127.0.0.1:8081',
+            },
+          },
+        },
+      });
+      _mockDevServerUrl('https://proxy.test/');
+
+      const AssetSources = require('../AssetSources');
+      const source = AssetSources.selectAssetSource(mockFontMetadata);
+
+      expect(source.uri).toBe(
+        `https://proxy.test/assets/test.ttf?platform=${Platform.OS}&hash=cafecafecafecafecafecafecafecafe`
+      );
+    });
+
+    it(`keeps the bundle URL's port`, () => {
+      _mockConstants({
+        experienceUrl: 'http://proxy.test:4443/app',
+        __unsafeNoWarnManifest2: {
+          extra: {
+            expoGo: {
+              developer: { tool: 'expo-cli' },
+              debuggerHost: '127.0.0.1:8081',
+            },
+          },
+        },
+      });
+      _mockDevServerUrl('http://proxy.test:4443/');
+
+      const AssetSources = require('../AssetSources');
+
+      expect(AssetSources.selectAssetSource(mockFontMonorepoMetadata).uri).toBe(
+        `http://proxy.test:4443/assets/?unstable_path=.%2Ftest.ttf&platform=${Platform.OS}&hash=cafecafecafecafecafecafecafecafe`
+      );
+    });
+
+    it(`falls back to debuggerHost when the bundle wasn't loaded from a server`, () => {
+      _mockConstants({
+        experienceUrl: 'http://127.0.0.1:8081/app',
+        __unsafeNoWarnManifest2: {
+          extra: {
+            expoGo: {
+              developer: { tool: 'expo-cli' },
+              debuggerHost: '127.0.0.1:8081',
+            },
+          },
+        },
+      });
+      _mockDevServerUrl(null);
+
+      const AssetSources = require('../AssetSources');
+
+      expect(AssetSources.selectAssetSource(mockFontMetadata).uri).toBe(
+        `http://127.0.0.1:8081/assets/test.ttf?platform=${Platform.OS}&hash=cafecafecafecafecafecafecafecafe`
+      );
+    });
+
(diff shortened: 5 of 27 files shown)


packages/expo-dev-menu/ios/DevMenuDevOptionsDelegate.swift
```diff
@@ -35,11 +35,18 @@ class DevMenuDevOptionsDelegate {
...
-    let host = bundleURL.host ?? "localhost"
-    let openURL = "http://\(host):\(port)/_expo/debugger?applicationId=\(Bundle.main.bundleIdentifier ?? "")"
-    guard let url = URL(string: openURL) else {
-      NSLog("[DevMenu] Invalid openJSInspector URL: $@", openURL)
+    // The bundle URL is the address this device reached the development server on, so its scheme and
+    // port are kept. Only its host is assumed when the bundle was loaded from disk.
+    var components = URLComponents()
+    components.scheme = bundleURL.scheme == "https" ? "https" : "http"
+    components.host = bundleURL.host ?? "localhost"
+    components.port = bundleURL.host != nil ? bundleURL.port : Int(RCT_METRO_PORT)
```
tsapeta: [🏗️ **design**] This detects "loaded from disk" by `host != nil`, but the Android half of this same PR guards on the scheme instead (`scheme == "http" || scheme == "https"` in `DevMenuDevToolsDelegate.getDevServerHost`). The two should agree.

The scheme check is also the stronger test. `host != nil` only works because an iOS release `bundleURL` is `file://`, where `URL.host` is `nil` (I confirmed this in Foundation). Any non-http scheme that *does* carry an authority slips through: `URL(string: "assets://index.android.bundle")!.host` is `"index.android.bundle"`, which would produce `http://index.android.bundle/_expo/debugger`. That is Android's release value rather than iOS's, so it can't fire today, but the guard is protecting against the wrong property.

Suggest keying on the scheme and reusing it, so the fallback and the scheme decision can't disagree:

```swift
let isServed = bundleURL.scheme == "http" || bundleURL.scheme == "https"
components.scheme = bundleURL.scheme == "https" ? "https" : "http"
components.host = isServed ? bundleURL.host : "localhost"
components.port = isServed ? bundleURL.port : Int(RCT_METRO_PORT)
```

apps/expo-go/ios/Exponent/DevMenu/DevMenuManager.swift
```diff
@@ -167,13 +167,20 @@ public class DevMenuManager: NSObject {
   }
 
   func openJSInspector() {
-    guard let manifestURL = currentManifestURL else {
+    // The bundle URL is the address this device reached the development server on, unlike the
+    // manifest URL, which carries the `exp` scheme and may omit the port the bundle was served on.
+    guard let bundleURL = currentBundleURL, let host = bundleURL.host else {
```
tsapeta: [💡 suggestion] This drops the `localhost:8081` fallback that the sibling change in `expo-dev-menu`'s `DevMenuDevOptionsDelegate.openJSInspector` keeps, so when there's no host the two dev menus now behave differently: one silently does nothing, the other still tries localhost.

In practice the guard shouldn't fire, since `currentBundleURL` resolves `manifest.bundleUrl()` against `currentManifestURL` and always yields an absolute URL in development. So this is about the inconsistency rather than a live failure. Either keep the fallback here or drop it there, whichever you think is right, but it'd be good for the two to match.

packages/expo-asset/src/__tests__/AssetSources-test.ts
```diff
@@ -277,3 +357,8 @@ function _mockConstants(constants: { [key: string]: any }): void {
     };
   });
 }
+
+/** Mock the origin the running bundle was loaded from. */
+function _mockDevServerUrl(origin: string | null): void {
```
tsapeta, line 365: [🏗️ **design**] `jest.doMock` registrations survive `jest.resetModules()`, so once any test calls `_mockDevServerUrl` the mock stays in place for every test that follows in this file. I confirmed this with a minimal repro on Jest 29: a test that never calls `doMock` still receives the mock registered by the previous test, even with `jest.resetModules()` in `afterEach`.

It happens to be harmless right now, only because the four new tests are last inside `selectAssetSource` and the following `resolveUri` block doesn't read `getBundleOrigin`. But it means the file's isolation depends on test ordering, and inserting a new `selectAssetSource` test after these would silently inherit `getBundleOrigin` returning `https://cdn.test/`.

Registering the mock unconditionally in `beforeEach` (as `Linking-test.ts` does, with `jest.mock` at module scope plus `mockReturnValue`) would make each test state what it expects.

packages/expo-linking/src/createURL.ts
```diff
@@ -79,7 +105,13 @@ export function createURL(
   path: string,
   { scheme, queryParams = {}, isTripleSlashed = false }: CreateURLOptions = {}
 ): string {
-  const resolvedScheme = resolveScheme({ scheme });
+  let resolvedScheme = resolveScheme({ scheme });
+
+  // Expo Go maps the `exps` scheme to HTTPS, which `exp` can't express, so an HTTPS development
+  // server needs it to be reached over the scheme it's actually served on.
+  if (!scheme && resolvedScheme === 'exp' && getDevServerLocation()?.isSecure) {
```
tsapeta: [💡 suggestion] `getDevServerLocation()` is unmemoized and runs up to four times per `createURL` call: once here, once via `getHostUri()`, and once more per `isExpoHosted()` call (which itself calls `getHostUri()`). Each run does a `require` (cached, so cheap) plus a `new URL()` parse, which isn't.

Beyond the cost, computing it repeatedly means the scheme decision on this line and the authority chosen on line 116 are derived from separate parses rather than one value, so they could in principle disagree. Hoisting it into a single `const devServer = getDevServerLocation()` at the top of `createURL` and threading it through would be both cheaper and easier to reason about.

packages/expo-linking/src/createURL.ts
```diff
@@ -4,8 +4,34 @@ import type { CreateURLOptions, ParsedURL } from './Linking.types';
...
+    return null;
+  }
+  // Required lazily, because reading the bundle URL initializes React Native's native module bridge,
+  // which isn't available in every runtime that loads this file, such as server rendering.
+  let bundleOrigin: string | null = null;
+  try {
+    bundleOrigin = (
+      require('expo/internal/bundle-origin') as typeof import('expo/internal/bundle-origin')
+    ).getBundleOrigin();
+  } catch {
```
tsapeta, line 20: [💡 suggestion] The `catch {}` here swallows every failure from loading `expo/internal/bundle-origin`, including a genuinely missing `expo` peer dependency, and falls back to `hostUri` silently. Since `expo` is now a peer dep of this package, a resolution failure is a setup problem worth surfacing rather than papering over.

The comment explains the intent well (server rendering has no native modules), but that specific case is narrower than what's being caught. Consider logging under `__DEV__`, so a misconfigured install shows up as something other than "deep links point at the wrong host".

apps/expo-go/ios/Exponent/Kernel/DevSupport/EXPackagerLogHelper.swift
```diff
@@ -121,15 +121,18 @@ import React
   }
 
   private func createSocket() {
-    let serverHost = bundleURL.host ?? "localhost"
-    let serverPort = bundleURL.port ?? Int(kRCTBundleURLProviderDefaultPort)
     let scheme = (bundleURL.scheme == "exps" || bundleURL.scheme == "https") ? "https" : "http"
```
tsapeta, line 124: [🐧 nit] The host/port grouping fix is right: previously `host` and `port` were resolved from independent fallbacks, so a URL with a host but no explicit port got that host paired with 8081.

Worth knowing for the PR description though: the parameter is named `bundleURL`, but its only caller passes a *manifest* URL (`EXErrorView.m:165` and `:167` pass `_appRecord.appLoader.manifestUrl`). That's why this line has to special-case `exps`. So the "address this device reached the development server on" rationale doesn't actually apply on this path. The change is still a strict improvement; the mismatch is pre-existing and might deserve a rename or a follow-up.

packages/expo-dev-menu/android/src/debug/java/expo/modules/devmenu/devtools/DevMenuDevToolsDelegate.kt
```diff
@@ -119,18 +119,29 @@ class DevMenuDevToolsDelegate(
     }
   }
 
+  private fun getDevServerHost(): String? {
+    val sourceUri = reactContext?.sourceURL?.toUri()
+    val scheme = sourceUri?.scheme
+    // A bundle loaded from disk has no reachable origin.
+    if (sourceUri != null && (scheme == "http" || scheme == "https")) {
```
tsapeta, line 126: [🐧 nit] The `sourceUri != null` test is redundant: `scheme` can only equal `"http"` or `"https"` when `sourceUri` is non-null, so the scheme comparison already implies it. `if (scheme == "http" || scheme == "https")` reads the same and is one condition shorter.

Also, the comment slightly overstates things. A disk-loaded bundle does have a `sourceURL`; it's `assets://index.android.bundle` or a bare filesystem path. It has no *http* origin, which is what the guard is really testing. Something like "a bundle loaded from disk has no http origin to reach" would match the code.

apps/expo-go/android/expoview/src/main/java/host/exp/exponent/experience/ExperienceActivity.kt
```diff
@@ -645,7 +646,7 @@ open class ExperienceActivity : BaseExperienceActivity(), StartReactInstanceDele
     Exponent.instance
       .testPackagerStatus(
         isDebugModeEnabled,
-        manifest!!,
+        ExponentUrls.bundleUrlFromManifest(manifest!!, manifestUrl!!),
```
tsapeta, line 649: [🐧 nit] This now evaluates `bundleUrlFromManifest` eagerly, before `testPackagerStatus` reaches its `if (!isDebug)` early return, whereas the manifest used to be passed in and `getDebuggerHost()` read only on the debug path. `getBundleURL()` bottoms out in `require("url")` / `require("bundleUrl")`, which throw `JSONException`, and `startReactInstance()` has no try/catch.

I don't think it's a new hazard in practice: every manifest Expo Go launched from necessarily has a launch-asset URL, and `ReactNativeActivity.kt:394` already calls `getBundleURL()` unguarded. Flagging it only because the work moved out of the `isDebug` guard, which is the kind of thing that's easy to lose track of later.

brentvatne:
## 🤖 AI code review

**Overall PR risk: Medium.** The change modifies existing behavior: expo-linking, expo-asset, and the Expo Go dev-menu and inspector paths now derive the dev server host from the bundle URL instead of the manifest's `hostUri` and `debuggerHost`. The blast radius stays inside development and debug tooling — deep links, asset URLs, the JS inspector, and packager status checks — and the diff is small and simple to revert. One warning survived review: the iOS `openJSInspector` code checks only for the "https" scheme, so a bundle URL with the "exps" scheme falls back to `http://` and the inspector does not open for an HTTPS or tunnel dev server. A reviewer also raised the new non-optional `expo` peer dependency in expo-linking; the range is `*` and the runtime code already tolerates a missing `expo`, so that item was dropped as low impact.

### 🟡 Warning (1)

- **Changelog says the bundle scheme is kept, but iOS code drops "exps"** — [`packages/expo-dev-menu/CHANGELOG.md:17`](https://github.com/expo/expo/pull/48275/files#diff-77acd136c1836a9e9fd7ad40c1402a8129f4d9a0429d863927cfb5ab67cbd399R17) _(correctness)_ · `id:eff869360351`
  **Confidence:** High — the iOS code this entry describes only checks for "https", not "exps", and the codebase treats "exps" as the marker for a secure dev server elsewhere.<br>**Impact if shipped:** Medium — the JS inspector fails to open for an HTTPS or tunnel dev server, the exact case this PR targets, while the changelog says the scheme is kept.<br>**Suggested remediation:** Either fix the two openJSInspector implementations to also treat "exps" as secure, matching EXPackagerLogHelper.swift, or reword the changelog entry so it does not claim the scheme is always kept.

  <details>

  The entry states the fix opens the JS inspector "keeping its scheme and port". This describes both the iOS and Android changes in the PR.

  On iOS, `DevMenuManager.swift` (`apps/expo-go/ios/Exponent/DevMenu/DevMenuManager.swift`) and `DevMenuDevOptionsDelegate.swift` (`packages/expo-dev-menu/ios/DevMenuDevOptionsDelegate.swift`) both build the scheme with `bundleURL.scheme == "https" ? "https" : "http"`. Neither checks for the "exps" scheme.

  The same PR's `EXPackagerLogHelper.swift` change, and existing code such as `EXVersionManagerObjC.mm:118` (`[bundleURL.scheme isEqualToString:@"https"] || [bundleURL.scheme isEqualToString:@"exps"]`) and `EXAppLoaderExpoUpdates.m:378-379`, show that this bundle URL can carry the "exps" scheme in Expo Go, and that "exps" must be treated as secure. `DevMenuManager.swift` builds `currentBundleURL` with `ApiUtil.bundleUrlFromManifest`, which does not convert "exps" to "https". The scheme that reaches `openJSInspector` can still be "exps".

  When `bundleURL.scheme` is "exps", both `openJSInspector` implementations take the `: "http"` branch. They then send an `http://` PUT request to the debugger endpoint. For a dev server that is reachable only over HTTPS, through a proxy or a tunnel, that request uses the wrong scheme and the JS inspector does not open.

  </details>


---

apps/expo-go/ios/Exponent/Kernel/DevSupport/EXPackagerLogHelper.swift
kitten: preexisting and not all that relevant since the two never differ

packages/expo-linking/src/createURL.ts
kitten: non-sensical/overstated. this isn't the only place where this will be called and there's no logging precedent here that'd apply. There'd also be no realistic way of differentiating all environments that would have to be handled and shouldn't fail, which is the responsibility of `expo/internal/bundle-origin`. The fallback is appropriate, and only here as an added protection layer

apps/expo-go/android/expoview/src/main/java/host/exp/exponent/experience/ExperienceActivity.kt
kitten: not really sure what this is trying to say, but by necessity `manifestUrl` has to be constructed and used first, and already be valid. If we get here it mean we already parsed it

packages/expo-dev-menu/android/src/debug/java/expo/modules/devmenu/devtools/DevMenuDevToolsDelegate.kt
kitten: `assets://` shouldn't occur here. But in practice this isn't a mistake, as this is just explicit, and practically a disk loaded bundle does not have the stated origins.

tsapeta (approved):

apps/expo-go/ios/Exponent/DevMenu/DevMenuManager.swift
```diff
@@ -167,13 +167,21 @@ public class DevMenuManager: NSObject {
...
+    // The bundle URL is the address this device reached the development server on, unlike the
+    // manifest URL, which carries the `exp` scheme and may omit the port the bundle was served on.
+    guard let bundleURL = currentBundleURL else {
       return
     }
-    let port = manifestURL.port ?? 8081
-    let host = manifestURL.host ?? "localhost"
-    let openURL = "http://\(host):\(port)/_expo/debugger?applicationId=\(Bundle.main.bundleIdentifier ?? "")"
-    guard let url = URL(string: openURL) else { return }
+    let isServed = bundleURL.scheme == "http" || bundleURL.scheme == "https" || bundleURL.scheme == "exps"
```
Kudo: [🚨 **critical**] `exp` is missing, so an `exp://` bundle URL makes `isServed` false and the inspector opens against `localhost:8081`. The old code read the host off `currentManifestURL`, so this is a regression on a device or behind a proxy — and `exp://` is exactly what `bundleUrlFromManifest` produces when the manifest carries a relative bundle URL, since it resolves against `appLoader.manifestUrl`.

```suggestion
    let isServed = ["http", "https", "exp", "exps"].contains(bundleURL.scheme)
```

packages/expo/src/utils/getBundleOrigin.ts
```diff
@@ -0,0 +1,26 @@
...
+ */
+export function getBundleOrigin(): string | null {
+  const bundleUrl = getBundleUrl();
+  if (!bundleUrl) {
+    return null;
+  }
+  try {
+    const { protocol, origin } = new URL(bundleUrl);
+    // e.g. http://localhost:8081
+    return protocol === 'http:' || protocol === 'https:' ? origin : null;
```
Kudo, line 22: [🏗️ **design**] Should `exp:` / `exps:` map to `http:` / `https:` here? On iOS Expo Go `EXReactAppManager.bundleUrl` resolves the manifest's bundle URL against `appLoader.manifestUrl`, which keeps the `exp` scheme, and that URL is what RN reports as `scriptURL`. So whenever the manifest uses relative URLs — the forwarded-authority case this PR is about — this returns `null` and `expo-linking` / `expo-asset` fall back to `hostUri` / `debuggerHost`. Android is unaffected because `ExponentUrls.resolveManifestUrl` resolves against the http manifest URL.

packages/expo-dev-menu/ios/DevMenuDevOptionsDelegate.swift
```diff
@@ -35,11 +35,17 @@ class DevMenuDevOptionsDelegate {
     guard let bundleURL = appContext?.bundleURL else {
       return
     }
-    let port = bundleURL.port ?? Int(RCT_METRO_PORT)
-    let host = bundleURL.host ?? "localhost"
-    let openURL = "http://\(host):\(port)/_expo/debugger?applicationId=\(Bundle.main.bundleIdentifier ?? "")"
-    guard let url = URL(string: openURL) else {
-      NSLog("[DevMenu] Invalid openJSInspector URL: $@", openURL)
+    let isServed = bundleURL.scheme == "http" || bundleURL.scheme == "https" || bundleURL.scheme == "exps"
```
Kudo: [💡 suggestion] Same gap as Expo Go's `DevMenuManager` — `exps` is handled but `exp` isn't, so a plain-HTTP Expo Go bundle URL falls back to `localhost`.

```suggestion
    let isServed = ["http", "https", "exp", "exps"].contains(bundleURL.scheme)
```

apps/expo-go/ios/Exponent/Versioned/Core/EXVersionManagerObjC.mm
```diff
@@ -268,6 +268,7 @@ - (BOOL)_isDevModeEnabledForHost:(NSURL *)bundleURL
 
 - (void)_openJsInspector:(NSURL *)bundleURL
 {
+  // TODO(@kitten): This might be unreachable. If it isn't it shouldn't hardcode `http`, but it looks unused to me
```
Kudo, line 271: [🐧 nit] It is unreachable — `-[EXReactAppManager selectDevMenuItemWithKey:]` is declared and defined but has no callers anywhere in the repo, so nothing reaches `selectDevMenuItemWithKey:host:bundleURL:` either. Deleting both is probably better than leaving a TODO.

Kudo (commented):
the other ai code review run than tomasz's. hope that helps something

apps/expo-go/ios/Exponent/Versioned/Core/EXVersionManagerObjC.mm
kitten: I don't have context on the method, so strictly outside of the responsibility of the PR. It'll still have to be addressed, but this is likely only being picked up here because I left the comment in the first place 😅

packages/expo/src/utils/getBundleOrigin.ts
kitten: No, this isn't a valid bundle URL. If this leaks into JS-land, something's gone terribly wrong, and that's a bug, so it shouldn't be handled here. This is admittedly very hard to trace, if this does happen, but I haven't observed it yet.

Edit: it does look like there's misalignment here on iOS. An SDK team ticket should be filed to increase consistency here, as there's too many URL shenanigans to keep track of (also a good opportunity to file the IPv6/URLComponents fixes for iOS)

Kudo (approved):

PR #48278, opened 2026-07-29
refactor(expo,expo-router,metro-runtime): Derive bundle URL consistently with `expo/internal/bundle-origin` instead of duplicate helpers

kitten:
# Why

Stacked on:
- #48275
- #47255

After the above PRs there's now an opportunity to align all JS-side helpers with `getBundleUrl` and expose this from `expo/internal/bundle-origin` to apply the logic of how to derive the dev server URL consistently.

# How

- Update `Tutorial` in `expo-router` (no base needed on `fetch`)
- Update RSC log on `expo-router` to use helper
- Replace internal uses of `getDevServer` in `expo` package
- Lazily require internally in helper in `expo/internal/bundle-origin`

# Test Plan

- Bundled against `native-component-list` and did a quick smoke test

# Checklist


brentvatne:
## 🤖 AI code review

**Overall PR risk: Medium.** The change modifies existing behavior: it replaces per-package helpers for deriving the development server URL with the shared `expo/internal/bundle-origin` helper, so the blast radius covers development-time paths in `expo`, `expo-router`, and `@expo/metro-runtime`. One warning survived review: the onboarding `Tutorial` component now posts to a relative path, which native apps that set `expo.extra.router.origin` to `false` cannot resolve, because those apps do not get the fetch polyfill. The most plausible breakage is that the onboarding button silently fails on native in that configuration, with no visible error. No security or secret issues were found.

### 🟡 Warning (1)

- **Tutorial POST uses a relative URL that native apps with `origin: false` cannot resolve** — [`packages/expo-router/src/onboard/Tutorial.tsx:17`](https://github.com/expo/expo/pull/48278/files#diff-f9e180e08008427ccb2f0ff1bd3c2cdc78fcbdbea19555329041163519fa96f0R17) _(correctness)_ · `id:374a10111614`
  **Confidence:** High — both install branches in `install.native.ts` were traced.<br>**Impact if shipped:** Medium — the onboarding button stops working on native for apps that set `origin` to `false`.<br>**Suggested remediation:** Build an absolute URL when there is no `window.location`. For example, resolve the path against `getBundleOrigin()` from `expo/internal/bundle-origin` and use the relative path only as a last resort.

  <details>

  The change replaces the absolute development server URL with the relative path `/_expo/touch`. On native, only the fetch polyfill in `@expo/metro-runtime` turns a relative path into an absolute URL.

  `packages/@expo/metro-runtime/src/location/install.native.ts` installs that polyfill conditionally:

  ```
  if (extra?.router?.origin !== false) {
    // ... setLocationHref + wrapFetchWithWindowLocation(fetch)
  } else {
    Object.defineProperty(global, 'fetch', { value: fetch });
  }
  ```

  When the app config sets `expo.extra.router.origin` to `false`, the app gets the raw fetch and `window.location` stays unset. `fetch('/_expo/touch')` then reaches React Native networking with a path and no host, so the request fails. `createEntryFileAsync()` is called without `await` and without `catch`, so the press does nothing visible and the rejection is unhandled.

  The old code did not depend on the polyfill. It read `getDevServer().url`, which is always an absolute development server URL, so the request worked in the same app.

  `origin: false` is a real configuration in this repo: `packages/expo-router/src/rsc/router/host.tsx` states it was the default for native before native server hosting existed.

  The same failure also applies when the bundle is not served over HTTP, because `getOrigin()` now returns no origin in that case.

  </details>


---

packages/expo/src/utils/getBundleUrl.native.ts
```diff
@@ -2,10 +2,14 @@
...
 /// <reference path="../ts-declarations/react-native.d.ts" />
 
-import SourceCode from 'react-native/Libraries/NativeModules/specs/NativeSourceCode';
+type NativeSourceCode =
+  typeof import('react-native/Libraries/NativeModules/specs/NativeSourceCode').default;
 
 export function getBundleUrl(): string | null {
-  let scriptURL = SourceCode.getConstants().scriptURL;
+  // NOTE(@kitten): Requiring this initialises module bridge, which may not be available server-side
+  const __nativeSourceCode = require('react-native/Libraries/NativeModules/specs/NativeSourceCode');
```
Kudo: [🏗️ **design**] This `require` is no longer guarded. `createURL.ts` used to wrap its `require('expo/internal/bundle-origin')` in a `try/catch` that returned `null`; the new static import dropped it, so a bridge failure here now throws out of `Linking.createURL()` and `Linking.parse()` instead of degrading. Per the note on the line above, server rendering is the runtime that protects against.

Wrapping lines 10-12 in a `try/catch` that returns `null` keeps every consumer of `getBundleOrigin()` safe, and matches how `getBundleOrigin()` already handles a bad URL.

packages/expo/src/utils/getDevServer.native.ts
```diff
@@ -0,0 +1,12 @@
...
+import { getBundleUrl } from './getBundleUrl';
+
+const FALLBACK_URL = 'http://localhost:8081/';
+
+export default function getDevServer() {
+  const origin = getBundleOrigin();
+  if (origin === null) {
+    return { url: FALLBACK_URL, fullBundleUrl: null, bundleLoadedFromServer: false };
+  }
+  return { url: `${origin}/`, fullBundleUrl: getBundleUrl(), bundleLoadedFromServer: true };
```
Kudo, line 12: [💡 suggestion] `getBundleOrigin()` already calls `getBundleUrl()`, so every `getDevServer()` does the deep `require`, `getConstants()` and URL parse twice. React Native's implementation parsed the script URL once and cached it, and `buildUrlForBundle` calls this on every async chunk load.

Caching inside `getBundleUrl.native.ts` covers both callers — the script URL can't change at runtime, and no native test mocks it per case.

packages/expo-router/src/onboard/Tutorial.tsx
```diff
@@ -15,8 +14,7 @@ function createEntryFileAsync() {
     return;
   }
 
-  // Pings middleware in the Expo CLI dev server.
-  return fetch(getDevServer().url + '_expo/touch', {
+  return fetch('/_expo/touch', {
```
Kudo, line 17: [💡 suggestion] On native this relative URL only resolves because `@expo/metro-runtime` wraps `fetch` to prepend `window.location.origin`. When an app sets `expo.extra.router.origin` to `false`, `install.native.ts` installs the unwrapped `fetch` instead, so this ping fails rather than reaching the dev server. Previously the absolute dev server URL made it work either way — is dropping that acceptable here?

Kudo: that one feels valid!

kitten: not really imho; I've removed this on purpose since there's too many cases where it's skipped. Realistically, the `Tutorial` isn't used much and protecting against this edge cases here isn't worth it. Technically, I don't think we officially document `false` anywhere but in types 🤔 

It doesn't make much sense to me to allow disabling it and/or accounting for it here. Doing both at the same time is odd, since then nothing can ever rely on this behaviour realistically

packages/expo/src/utils/getDevServer.native.ts
kitten: First point is fine, second I think is more of a nice-to-have, but I'd rather not trace right now whether this is really a constant in every case and assume it for a URL parse that's ultimately very cheap

Kudo (approved):

Issue #48195, opened 2026-07-27
SDK 54 upgrade regressed WebView recovery and enabled iOS New Architecture; background memory reclaim needs framework support for WebView-heavy apps


HammadAliHashmi:
### Do you understand that any feature requests opened in the core Expo repository will be closed?

Yes

PR #48290, opened 2026-07-29
[expo-router] add processScreens option to standard-navigation integration

Ubax:
# Why

Add `processScreens` option to standard navigation API to map expo-router descriptors to the standard-navigation ones.

# How

1. Expose `processScreens` from `unstable_integrateWithRouter` and `unstable_createStandardRouterNavigator`
2. Pass the function to `withLayoutContext`
3. Improve typings for `useFilterScreenChildren` and `processScreens` in `withLayoutContext`

# Test Plan

CI

# Checklist

### DIFF
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -17,6 +17,7 @@
 - Add `pageHeaders` config plugin option for declaring per-path response headers ([#47429](https://github.com/expo/expo/pull/47429) by [@hassankhan](https://github.com/hassankhan))
 - Upgrade react-native-screens to 4.26.0 ([#47770](https://github.com/expo/expo/pull/47770) by [@Ubax](https://github.com/Ubax))
 - Improve standard-navigation types for `createProps`. ([#47825](https://github.com/expo/expo/pull/47825) by [@Ubax](https://github.com/Ubax))
+- Add a `processScreens` option to the `standard-navigation` integration. ([#48290](https://github.com/expo/expo/pull/48290) by [@Ubax](https://github.com/Ubax))
 
 ### 🐛 Bug fixes
 
--- a/packages/expo-router/src/standard-navigation/__tests__/integration.test.ios.tsx
+++ b/packages/expo-router/src/standard-navigation/__tests__/integration.test.ios.tsx
@@ -382,6 +382,178 @@ describe('unstable_integrateWithRouter / unstable_createStandardRouterNavigator'
   });
 });
 
+describe('processScreens', () => {
+  const optionsByName = () =>
+    Object.fromEntries(
+      lastArgs().state.routes.map(
+        (route) => [route.name, lastArgs().descriptors[route.key]!.options] as const
+      )
+    );
+
+  it('transforms declared screen options before they are rendered', () => {
+    const Prefixed = unstable_createStandardRouterNavigator<
+      TestOptions,
+      TabNavigationState<ParamListBase>,
+      TestEventMap,
+      object,
+      TabRouterOptions
+    >(NavigatorContent, TabRouter, {
+      processScreens: (screens) =>
+        screens.map((screen) => ({
+          ...screen,
+          options: { ...screen.options, title: `processed-${screen.name}` },
+        })),
+    });
+
+    renderRouter({
+      _layout: () => (
+        <Prefixed>
+          <Prefixed.Screen name="index" options={{ title: 'Home' }} />
+          <Prefixed.Screen name="second" />
+        </Prefixed>
+      ),
+      index: () => <View testID="index" />,
+      second: () => <View testID="second" />,
+    });
+
+    expect(optionsByName()).toMatchObject({
+      index: { title: 'processed-index' },
+      second: { title: 'processed-second' },
+    });
+  });
+
+  it('receives only the declared screens', () => {
+    const names: string[] = [];
+    const Recording = unstable_createStandardRouterNavigator<
+      TestOptions,
+      TabNavigationState<ParamListBase>,
+      TestEventMap,
+      object,
+      TabRouterOptions
+    >(NavigatorContent, TabRouter, {
+      processScreens: (screens) => {
+        names.length = 0;
+        names.push(...screens.map((screen) => screen.name));
+        return screens;
+      },
+    });
+
+    renderRouter({
+      _layout: () => (
+        <Recording>
+          <Recording.Screen name="index" />
+        </Recording>
+      ),
+      index: () => <View testID="index" />,
+      second: () => <View testID="second" />,
+    });
+
+    expect(names).toEqual(['index']);
+    // The undeclared route is still registered, it just never reaches the processor.
+    expect(lastArgs().state.routes.map((route) => route.name)).toContain('second');
+  });
+
+  it('rejects dropped screens', () => {
+    const Filtered = unstable_createStandardRouterNavigator<
+      TestOptions,
+      TabNavigationState<ParamListBase>,
+      TestEventMap,
+      object,
+      TabRouterOptions
+    >(NavigatorContent, TabRouter, {
+      processScreens: (screens) => screens.filter((screen) => screen.name !== 'second'),
+    });
+
+    expect(() =>
+      renderRouter({
+        _layout: () => (
+          <Filtered>
+            <Filtered.Screen name="index" options={{ title: 'Home' }} />
+            <Filtered.Screen name="second" options={{ title: 'Declared title' }} />
+          </Filtered>
+        ),
+        index: () => <View testID="index" />,
+        second: () => <View testID="second" />,
+      })
+    ).toThrow('`processScreens` must not add, remove, rename, or duplicate screens');
+  });
+
+  it('rejects renamed screens', () => {
+    const Renamed = unstable_createStandardRouterNavigator<
+      TestOptions,
+      TabNavigationState<ParamListBase>,
+      TestEventMap,
+      object,
+      TabRouterOptions
+    >(NavigatorContent, TabRouter, {
+      processScreens: (screens) =>
+        screens.map((screen) => ({ ...screen, name: `renamed-${screen.name}` })),
+    });
+
+    expect(() =>
+      renderRouter({
+        _layout: () => (
+          <Renamed>
+            <Renamed.Screen name="index" />
+          </Renamed>
+        ),
+        index: () => <View testID="index" />,
+      })
+    ).toThrow('`processScreens` must not add, remove, rename, or duplicate screens');
+  });
+
+  it('rejects screens renamed in place', () => {
+    const Renamed = unstable_createStandardRouterNavigator<
+      TestOptions,
+      TabNavigationState<ParamListBase>,
+      TestEventMap,
+      object,
+      TabRouterOptions
+    >(NavigatorContent, TabRouter, {
+      processScreens: (screens) => {
+        screens[0] = { ...screens[0]!, name: 'renamed-index' };
+        return screens;
+      },
+    });
+
+    expect(() =>
+      renderRouter({
+        _layout: () => (
+          <Renamed>
+            <Renamed.Screen name="index" />
+          </Renamed>
+        ),
+        index: () => <View testID="index" />,
+      })
+    ).toThrow('`processScreens` must not add, remove, rename, or duplicate screens');
+  });
+
+  it('rejects duplicate screens returned by the processor', () => {
+    const Duplicated = unstable_createStandardRouterNavigator<
+      TestOptions,
+      TabNavigationState<ParamListBase>,
+      TestEventMap,
+      object,
+      TabRouterOptions
+    >(NavigatorContent, TabRouter, {
+      processScreens: (screens) => [screens[0]!, { ...screens[0]! }],
+    });
+
+    expect(() =>
+      renderRouter({
+        _layout: () => (
+          <Duplicated>
+            <Duplicated.Screen name="index" />
+            <Duplicated.Screen name="second" />
+          </Duplicated>
+        ),
+        index: () => <View testID="index" />,
+        second: () => <View testID="second" />,
+      })
+    ).toThrow('`processScreens` must not add, remove, rename, or duplicate screens');
+  });
+});
+
 describe('preloaded routes projected through the integration (StackRouter)', () => {
   const StandardStack = unstable_createStandardRouterNavigator<
     TestOptions,
--- a/packages/expo-router/src/standard-navigation/__tests__/types.test.tsx
+++ b/packages/expo-router/src/standard-navigation/__tests__/types.test.tsx
@@ -209,9 +209,46 @@ unstable_createStandardRouterNavigator(OptionalCreateContent, TabRouter, {
 createSplitNav(SplitContent, TabRouter);
 
 // @ts-expect-error `createProps` is required when `CreateProps` is non-empty.
-createSplitNav(SplitContent, TabRouter, { useOnlyUserDefinedScreens: true });
+createSplitNav(SplitContent, TabRouter, { processScreens: (screens) => screens });
 
 createPublicNav(PublicContent, TabRouter);
+
+// ---------------------------------------------------------------------------
+// processScreens is always optional and composes with createProps
+// ---------------------------------------------------------------------------
+
+type ProcessScreens = NonNullable<
+  IntegrateWithRouterOptions<TabState, object, Opts, EventMap>['processScreens']
+>;
+export type _ProcessedScreenNameIsRequired = Expect<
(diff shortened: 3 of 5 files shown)


Ubax:
* **#48290** <a href="https://app.graphite.com/github/pr/expo/expo/48290?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/48290?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>


<h2></h2>

packages/expo-router/CHANGELOG.md
```diff
@@ -23,6 +23,7 @@
 - Add `pageHeaders` config plugin option for declaring per-path response headers ([#47429](https://github.com/expo/expo/pull/47429) by [@hassankhan](https://github.com/hassankhan))
 - Upgrade react-native-screens to 4.26.0 ([#47770](https://github.com/expo/expo/pull/47770) by [@Ubax](https://github.com/Ubax))
 - Improve standard-navigation types for `createProps`. ([#47825](https://github.com/expo/expo/pull/47825) by [@Ubax](https://github.com/Ubax))
+- Add a `processScreens` option to the `standard-navigation` integration. ([#48195](https://github.com/expo/expo/pull/48195) by [@Ubax](https://github.com/Ubax))
```
Ubax: [💡 suggestion] This links to #48195, which does not exist (`gh pr view 48195` returns "Could not resolve to a PullRequest"). It should link to this PR, #48290.

docs/pages/router/advanced/custom-navigators.mdx
Ubax, line 1: Revert docs changes

packages/expo-router/src/standard-navigation/__tests__/integration.test.ios.tsx
```diff
@@ -382,6 +382,80 @@ describe('unstable_integrateWithRouter / unstable_createStandardRouterNavigator'
...
+    });
+
+    expect(optionsByName()).toMatchObject({
+      index: { title: 'processed-index' },
+      second: { title: 'processed-second' },
+    });
+  });
+
+  // The processor runs on the declared screens only, after `Protected` flattening and before
+  // `useSortedScreens`, so filesystem routes that were never declared are not passed to it.
```
Ubax: Remove this comment

packages/expo-router/src/standard-navigation/__tests__/types.test.tsx
```diff
@@ -209,10 +209,39 @@ unstable_createStandardRouterNavigator(OptionalCreateContent, TabRouter, {
...
+createSplitNav(SplitContent, TabRouter, {
+  createProps: () => ({ routeNames: [], preload: () => {} }),
+  processScreens: (screens) => screens.map((screen) => ({ ...screen, redirect: false })),
+});
+
+// @ts-expect-error `processScreens` alone does not satisfy a non-empty `CreateProps`.
+createSplitNav(SplitContent, TabRouter, { processScreens: (screens) => screens });
+
+// The screens carry the navigator's own options, so reading an undeclared one is rejected. This is
+// what makes a shortcut-prop processor (like the `href` one in `TabsClient`) type-safe.
```
Ubax:
```suggestion
// The screens carry the navigator's own options, so reading an undeclared one is rejected.
```

packages/expo-router/src/standard-navigation/index.tsx
```diff
@@ -209,7 +212,11 @@ export function unstable_integrateWithRouter<
   }
 
   return withLayoutContext<NavigatorOptions, typeof StandardRouterNavigator, State, EventMap>(
-    StandardRouterNavigator
+    StandardRouterNavigator,
+    // `withLayoutContext` erases the options type of the screens it hands to the processor. The
+    // screens it passes are the ones declared for this navigator, so their options really are
+    // `NavigatorOptions`; only the signature is broader.
+    options?.processScreens as ((screens: ScreenProps[]) => ScreenProps[]) | undefined
```
Ubax: Can this be fixed without the cast?

packages/expo-router/src/standard-navigation/types.ts
```diff
@@ -80,10 +81,30 @@ type CreatePropsOption<State extends NavigationState, CreateProps extends object
     }
   : { createProps: CreatePropsFn<State, CreateProps> };
 
+// Declared as an intersection rather than extra members on `CreatePropsOption`: that type is a
+// conditional, and excess-property checking would reject `{ createProps, processScreens }` against
+// whichever branch it resolves to.
```
Ubax: Delete this comment

packages/expo-router/src/standard-navigation/types.ts
```diff
@@ -80,10 +81,30 @@ type CreatePropsOption<State extends NavigationState, CreateProps extends object
...
+  NavigatorOptions extends object = Record<string, any>,
+> = CreatePropsOption<State, CreateProps> & {
+  /**
+   * Transforms the screens declared as children of the navigator before they are rendered.
+   *
+   * Runs on every render, after `Protected` children are flattened and before the screens are
+   * sorted, and receives the declared screens only — routes Expo Router inferred from the
+   * filesystem are not passed to it. Use it to support shortcut props on `<Navigator.Screen>` by
+   * rewriting them into real screen options.
+   *
```
Ubax:
```suggestion
```

packages/expo-router/src/standard-navigation/types.ts
```diff
@@ -83,7 +84,19 @@ type CreatePropsOption<State extends NavigationState, CreateProps extends object
...
+  /**
+   * Transforms the screens declared as children of the navigator before they are rendered.
+   *
+   * @example
+   * ```tsx
+   * processScreens: (screens) =>
+   *   screens.map((screen) => ({ ...screen, options: { ...screen.options, title: screen.name } })),
+   * ```
+   */
+  processScreens?: (screens: ScreenProps<NavigatorOptions>[]) => ScreenProps<NavigatorOptions>[];
```
jakub-agent: [🏗️ **design**] `ScreenProps<NavigatorOptions>` has `name?: string`, but `useFilterScreenChildren` guarantees `name` on every screen it collects — its internal type is `(ScreenProps<TOptions> & { name: string })[]`. Losing that invariant hurts in both directions.

On the input side, `screen.name` is `string | undefined`, so users must handle a case that cannot happen. The `@example` above shows it: `title: screen.name` yields `string | undefined`, which fails for any navigator whose `title` is required. The new test in `integration.test.ios.tsx` has to type its accumulator `(string | undefined)[]` for the same reason.

On the output side, returning a screen without a `name` type-checks and fails quietly. I ran this against `56a6c9c`:

```tsx
processScreens: (screens) => screens.map(({ name, ...rest }) => rest),
```

Both screens still render, but every declared option is dropped (`descriptors[key].options` is `{}`), and the only signal is two copies of `[Layout children]: No route named "undefined" exists in nested children:`. Destructuring `name` off to compute something is a plausible mistake, and the payload — options silently vanish — is hard to trace back to the processor.

Suggest making the invariant part of the signature:

```ts
processScreens?: (
  screens: (ScreenProps<NavigatorOptions> & { name: string })[]
) => (ScreenProps<NavigatorOptions> & { name: string })[];
```

That matches what `useFilterScreenChildren` already produces, so no cast is needed at the call site, and it makes the `name`-dropping case a compile error instead of a silent one. A test covering a processor that drops or renames a screen would be worth adding either way.

packages/expo-router/src/layouts/withLayoutContext.tsx
```diff
@@ -142,13 +144,13 @@ export function withLayoutContext<
   T extends ComponentType<any>,
   TState extends NavigationState,
   TEventMap extends EventMapBase,
->(Nav: T, processor?: (options: ScreenProps[]) => ScreenProps[]) {
+>(Nav: T, processor?: (options: ScreenProps<TOptions>[]) => ScreenProps<TOptions>[]) {
```
jakub-agent: [💡 suggestion] Two things on this signature, both follow-ons from the `types.ts` comment.

1. Same `name` invariant: the value actually passed in on line 157 is `(ScreenProps<TOptions> & { name: string })[]`, so `& { name: string }` here would be accurate rather than a narrowing, and would flow straight through to `processScreens`.

2. `TState` and `TEventMap` are in scope and already used for the returned `.Screen` (`ScreenProps<TOptions, TState, TEventMap>`), but the processor only gets `ScreenProps<TOptions>`. So a processor sees `listeners` typed against the default `NavigationState`/`EventMapBase` while the screens it receives were declared with the navigator's own state and event map. Passing all three keeps the two views of a screen consistent.

Minor: the JSDoc on line 116 calls this `@param processor` while the public option is `processScreens` — worth aligning the names now that it is user-facing.

packages/expo-router/CHANGELOG.md
```diff
@@ -23,6 +23,7 @@
 - Add `pageHeaders` config plugin option for declaring per-path response headers ([#47429](https://github.com/expo/expo/pull/47429) by [@hassankhan](https://github.com/hassankhan))
 - Upgrade react-native-screens to 4.26.0 ([#47770](https://github.com/expo/expo/pull/47770) by [@Ubax](https://github.com/Ubax))
 - Improve standard-navigation types for `createProps`. ([#47825](https://github.com/expo/expo/pull/47825) by [@Ubax](https://github.com/Ubax))
+- Add a `processScreens` option to the `standard-navigation` integration. ([#48290](https://github.com/expo/expo/pull/48290) by [@Ubax](https://github.com/Ubax))
```
jakub-agent, line 20: [💡 suggestion] The entry covers the new `standard-navigation` option, but this PR also tightens the type of the second parameter of `withLayoutContext`, which is exported from the `expo-router` root entry point. A processor annotated against the old signature no longer compiles when the navigator's options have a required key — confirmed at `56a6c9c`:

```
Argument of type '(screens: ScreenProps[]) => ScreenProps[]' is not assignable to parameter
of type '(options: ScreenProps<RequiredOpts>[]) => ScreenProps<RequiredOpts>[]'.
  Property 'title' is missing in type 'Record<string, any>' but required in type 'RequiredOpts'.
```

Narrow in practice — most `withLayoutContext` callers pass no processor at all, and option types with only optional keys (like `MaterialTopTabNavigationOptions`) still work. But it is a compile-time break for the ones that do, so a second changelog line mentioning the `withLayoutContext` processor typing would save someone a confused bug report.

packages/expo-router/src/layouts/withLayoutContext.tsx
```diff
@@ -142,13 +144,13 @@ export function withLayoutContext<
...
   return Object.assign(
     forwardRef(({ children: userDefinedChildren, ...props }: any, ref) => {
       const contextKey = useContextKey();
       const node = useRouteNode();
 
-      const { screens, guardedRedirects } = useFilterScreenChildren(userDefinedChildren, {
+      const { screens, guardedRedirects } = useFilterScreenChildren<TOptions>(userDefinedChildren, {
```
jakub-agent: [🐧 nit] The unique-name assertion inside `useFilterScreenChildren` (line 97) runs before the processor on line 157, so duplicates the processor introduces are not caught by it. Confirmed at `56a6c9c` with `processScreens: (screens) => [...screens, { ...screens[0] }]`: instead of `Screen names must be unique: ...`, you get `[Layout children]: No route named "index" exists in nested children:`, which points at the layout children rather than at the processor.

Not a blocker, and it only bites a processor that adds or renames screens. But since `processScreens` makes that reachable from user code, re-running the dev-only uniqueness check on `processed` (or moving it after this line) would keep the clearer error. Tightening the `name` type as suggested on `types.ts` does not cover this case.

packages/expo-router/src/standard-navigation/__tests__/types.test.tsx
```diff
@@ -209,10 +209,38 @@ unstable_createStandardRouterNavigator(OptionalCreateContent, TabRouter, {
 createSplitNav(SplitContent, TabRouter);
 
 // @ts-expect-error `createProps` is required when `CreateProps` is non-empty.
-createSplitNav(SplitContent, TabRouter, { useOnlyUserDefinedScreens: true });
+createSplitNav(SplitContent, TabRouter, { unknownOption: true });
```
jakub-agent: [🐧 nit] `unknownOption` is not a key of the options type, so this errors on the unknown-property check whether or not `createProps` is required — the assertion passes for a reason other than the one its comment states. The case you added below (`// @ts-expect-error \`processScreens\` alone does not satisfy a non-empty \`CreateProps\``) uses a real optional key and does prove it, so this line is now redundant and could be dropped.

jakub-agent (commented):

Small, well-scoped addition. The plumbing is right: `processScreens` is read once at navigator-creation time and forwarded to the existing `withLayoutContext` processor slot, so it reuses machinery that `Tabs` already depends on rather than adding a new path. Threading `NavigatorOptions` into `IntegrateWithRouterOptions` (with a `Record<string, any>` default) keeps the public type backward compatible, and it is what lets the `tabBarButton` `@ts-expect-error` in `TabsClient.tsx` go away — a genuine typing win, not a suppression.

Verified locally at `56a6c9c`: `yarn typecheck` (`tsc -b tsconfig.all.json`) exits 0, and `jest src/layouts src/standard-navigation` passes 726 tests / 59 suites, including the two new `processScreens` cases.

The findings below are all about the type contract of the new option rather than its behaviour. The main one: `ScreenProps<NavigatorOptions>` has `name?: string`, but every screen reaching the processor is guaranteed to have a `name` at runtime, and returning one without a `name` silently discards that screen's declared options. Worth tightening before this shape settles, even behind `unstable_`.

I also checked a few things that turned out fine, for the record: the processor runs once per mount and is not re-invoked on navigation (so no per-render cost concern); the CHANGELOG entry is in the right section; and no `build/` artifacts need regenerating since `packages/expo-router` does not commit them.

packages/expo-router/src/standard-navigation/types.ts
```diff
@@ -83,7 +85,22 @@ type CreatePropsOption<State extends NavigationState, CreateProps extends object
 export type IntegrateWithRouterOptions<
   State extends NavigationState = NavigationState,
   CreateProps extends object = object,
-> = CreatePropsOption<State, CreateProps>;
+  NavigatorOptions extends object = Record<string, any>,
+  EventMap extends EventMapBase = EventMapBase,
```
jakub-agent, line 89: [🐧 nit] Carried over from the last round, still open — non-blocking.

Every other `EventMap` parameter in this file is constrained to the module's own `StandardNavigatorEventMapBase` (lines 29, 36, 120, 138, 171, 180); this one uses React Navigation's `EventMapBase`, and that is the only reason for the `EventMapBase` import added at line 10.

Swapping it works — I changed it to `EventMap extends StandardNavigatorEventMapBase = StandardNavigatorEventMapBase` and `tsc -b tsconfig.all.json` was clean apart from `EventMapBase` then being unused, so the import can go too. `StandardNavigatorEventMapBase` still satisfies `ScreenProps`'s `TEventMap extends EventMapBase` constraint.

The practical difference is only in the default: `IntegrateWithRouterOptions` used bare would then describe listeners against the tighter standard-navigation event shape instead of the looser React Navigation one. Fine to leave as-is if the wider bound is deliberate.

jakub-agent (approved):

**Iteration 3** — PR 2/2 in the stack, based on [#48356](https://github.com/expo/expo/pull/48356).

Everything from the previous two rounds is resolved. This PR's own diff has not changed since iteration 2; the fixes landed in the parent.

| Finding | Status |
| --- | --- |
| `processScreens` loses the `name` invariant | Fixed — `& { name: string }` both directions, plus a `_ProcessedScreenNameIsRequired` type test |
| `withLayoutContext` processor typing / naming | Fixed in #48356 |
| Changelog coverage | Fixed — #48356 carries its own entry |
| Uniqueness guard ran before the processor | Fixed in #48356; the duplicate gap I reported is closed and has a regression test |
| Duplicate assertion in `types.test.tsx` | Fixed |
| Typecheck break in the parent's new test file | Fixed in #48356 |

Verified at `62a9f768d98` (stacked on `ee9c39e1f58`):

- `tsc -b tsconfig.all.json` → exit 0, no errors
- `jest src/layouts src/standard-navigation src/__tests__` → 107 suites, 1286 passed, 31 snapshots
- Full `jest` run is green apart from 3 pre-existing `native-tabs/__rsc_tests__` snapshot failures caused by the absolute asset path in my checkout location; those reproduce unchanged on `origin/main`, so they are unrelated to this stack

I also re-ran the duplication counterexample from the last round directly against this head. The guard now behaves as documented: duplicating without dropping throws, while a pure options transform and a reorder both still pass.

One cosmetic nit left, carried over and still open. Non-blocking — the API shape and the semantics look right to me.

PR #48291, opened 2026-07-29
[expo-router] prepare bottom tabs views for standard navigation

Ubax:
# Why

Refactor `BottomTabs` before migrating to `standard-navigation` to minimize the diff

# How

1. Remove `navigation` prop from `BottomTabView` and replace it with direct APIs for:
   - dispatching navigate action - `navigateToTab`
   - emitting events - `emitter`
   - popping the nested stack to top - `popNestedStackToTop`
2. Replace the state used by `BottomTabView` with a simple standard-navigation compliant one
   - pass additional prop - `preloadedRouteKeys`
3. Add missing types

# Test Plan

CI

# Checklist

### DIFF
--- a/packages/expo-router/src/react-navigation/bottom-tabs/navigators/createBottomTabNavigator.tsx
+++ b/packages/expo-router/src/react-navigation/bottom-tabs/navigators/createBottomTabNavigator.tsx
@@ -1,8 +1,13 @@
 'use client';
+import { useCallback } from 'react';
+
+import { useStandardEmitter } from '../../../standard-navigation/useStandardEmitter';
 import {
+  CommonActions,
   createNavigatorFactory,
   type NavigatorTypeBagBase,
   type ParamListBase,
+  StackActions,
   type TabActionHelpers,
   type TabNavigationState,
   TabRouter,
@@ -48,9 +53,52 @@ function BottomTabNavigator({
     UNSTABLE_router,
   });
 
+  const emitter = useStandardEmitter(navigation);
+
+  const navigateToTab = useCallback(
+    (routeKey: string) => {
+      const route = state.routes.find((route) => route.key === routeKey);
+
+      if (route) {
+        navigation.dispatch({
+          ...CommonActions.navigate(route),
+          target: state.key,
+        });
+      } else if (__DEV__) {
+        console.warn(
+          `Bottom tabs could not switch to the tab "${routeKey}" because no tab with that key exists. ` +
+            `'navigateToTab' takes a route key, not a route name — pass 'route.key' from the tab bar props.`
+        );
+      }
+    },
+    [navigation, state.key, state.routes]
+  );
+
+  const popNestedStackToTop = useCallback(
+    (routeKey: string) => {
+      const prevRoute = state.routes.find((route) => route.key === routeKey);
+
+      if (prevRoute?.state?.type === 'stack' && prevRoute.state.key) {
+        navigation.dispatch({
+          ...StackActions.popToTop(),
+          target: prevRoute.state.key,
+        });
+      }
+    },
+    [navigation, state.routes]
+  );
+
   return (
     <NavigationContent>
-      <BottomTabView {...rest} state={state} navigation={navigation} descriptors={descriptors} />
+      <BottomTabView
+        {...rest}
+        state={state}
+        descriptors={descriptors}
+        emitter={emitter}
+        navigateToTab={navigateToTab}
+        preloadedRouteKeys={state.preloadedRouteKeys}
+        popNestedStackToTop={popNestedStackToTop}
+      />
     </NavigationContent>
   );
 }
--- a/packages/expo-router/src/react-navigation/bottom-tabs/types.tsx
+++ b/packages/expo-router/src/react-navigation/bottom-tabs/types.tsx
@@ -8,12 +8,12 @@ import type {
   ViewStyle,
 } from 'react-native';
 import type { EdgeInsets } from 'react-native-safe-area-context';
+import type { NavigatorArgs } from 'standard-navigation';
 
 import type { HeaderOptions, PlatformPressable } from '../elements';
 import type {
   DefaultNavigatorOptions,
   Descriptor,
-  NavigationHelpers,
   NavigationProp,
   ParamListBase,
   RouteProp,
@@ -35,24 +35,37 @@ export type BottomTabNavigationEventMap = {
   /**
    * Event which fires on long press on the tab in the tab bar.
    */
-  tabLongPress: { data: undefined };
+  tabLongPress: { data: undefined; canPreventDefault: false };
   /**
    * Event which fires when a transition animation starts.
    */
-  transitionStart: { data: undefined };
+  transitionStart: { data: undefined; canPreventDefault: false };
   /**
    * Event which fires when a transition animation ends.
    */
-  transitionEnd: { data: undefined };
+  transitionEnd: { data: undefined; canPreventDefault: false };
 };
 
 export type LabelPosition = 'beside-icon' | 'below-icon';
 
-export type BottomTabNavigationHelpers = NavigationHelpers<
-  ParamListBase,
+export type BottomTabViewRoute = {
+  key: string;
+  name: string;
+  params?: object;
+};
+
+/**
+ * The navigator state consumed by `BottomTabView` and the tab bar.
+ */
+export type BottomTabViewState = {
+  index: number;
+  routes: BottomTabViewRoute[];
+};
+
+export type BottomTabEmitter = NavigatorArgs<
+  Record<string, never>,
   BottomTabNavigationEventMap
-> &
-  TabActionHelpers<ParamListBase>;
+>['emitter'];
 
 export type BottomTabNavigationProp<
   ParamList extends ParamListBase,
@@ -408,10 +421,12 @@ export type BottomTabHeaderProps = {
   navigation: BottomTabNavigationProp<ParamListBase>;
 };
 
+// TODO(@ubax): update docs — https://linear.app/expo/issue/ENG-25579/update-documentation-after-the-refactor
 export type BottomTabBarProps = {
-  state: TabNavigationState<ParamListBase>;
+  state: BottomTabViewState;
   descriptors: BottomTabDescriptorMap;
-  navigation: NavigationHelpers<ParamListBase, BottomTabNavigationEventMap>;
+  emitter: BottomTabEmitter;
+  navigateToTab: (routeKey: string) => void;
   insets: EdgeInsets;
 };
 
--- a/packages/expo-router/src/react-navigation/bottom-tabs/views/BottomTabBar.tsx
+++ b/packages/expo-router/src/react-navigation/bottom-tabs/views/BottomTabBar.tsx
@@ -12,16 +12,8 @@ import {
 import type { EdgeInsets } from 'react-native-safe-area-context';
 
 import { getDefaultSidebarWidth, getLabel, MissingIcon, useFrameSize } from '../../elements';
-import {
-  CommonActions,
-  NavigationProvider,
-  type ParamListBase,
-  type TabNavigationState,
-  useLinkBuilder,
-  useLocale,
-  useTheme,
-} from '../../native';
-import type { BottomTabBarProps, BottomTabDescriptorMap } from '../types';
+import { NavigationProvider, useLinkBuilder, useLocale, useTheme } from '../../native';
+import type { BottomTabBarProps, BottomTabDescriptorMap, BottomTabViewState } from '../types';
 import { BottomTabBarHeightCallbackContext } from '../utils/BottomTabBarHeightCallbackContext';
 import { useIsKeyboardShown } from '../utils/useIsKeyboardShown';
 import { BottomTabItem } from './BottomTabItem';
@@ -39,7 +31,7 @@ const DEFAULT_MAX_TAB_ITEM_WIDTH = 125;
 const useNativeDriver = process.env.EXPO_OS !== 'web';
 
 type Options = {
-  state: TabNavigationState<ParamListBase>;
+  state: BottomTabViewState;
   descriptors: BottomTabDescriptorMap;
   dimensions: { height: number; width: number };
 };
@@ -129,7 +121,7 @@ export const getTabBarHeight = ({
   return TABBAR_HEIGHT_UIKIT + inset;
 };
 
-export function BottomTabBar({ state, navigation, descriptors, insets, style }: Props) {
+export function BottomTabBar({ state, descriptors, emitter, navigateToTab, insets, style }: Props) {
   const { colors } = useTheme();
   const { direction } = useLocale();
   const { buildHref } = useLinkBuilder();
@@ -344,22 +336,19 @@ export function BottomTabBar({ state, navigation, descriptors, insets, style }: 
           const { options } = descriptors[route.key]!;
 
           const onPress = () => {
-            const event = navigation.emit({
+            const event = emitter.emit({
               type: 'tabPress',
               target: route.key,
               canPreventDefault: true,
             });
 
             if (!focused && !event.defaultPrevented) {
-              navigation.dispatch({
-                ...CommonActions.navigate(route),
-                target: state.key,
-              });
+              navigateToTab(route.key);
             }
           };
 
           const onLongPress = () => {
-            navigation.emit({
+            emitter.emit({
               type: 'tabLongPress',
               target: route.key,
             });
--- a/packages/expo-router/src/react-navigation/bottom-tabs/views/BottomTabView.tsx
+++ b/packages/expo-router/src/react-navigation/bottom-tabs/views/BottomTabView.tsx
@@ -4,21 +4,17 @@ import { Animated, Platform, StyleSheet } from 'react-native';
 import { SafeAreaInsetsContext } from 'react-native-safe-area-context';
 
 import { getHeaderTitle, Header, SafeAreaProviderCompat, Screen } from '../../elements';
-import {
-  type NavigationAction,
-  type ParamListBase,
-  StackActions,
-  type TabNavigationState,
-} from '../../native';
+import { type ParamListBase } from '../../native';
 import { FadeTransition, ShiftTransition } from '../TransitionConfigs/TransitionPresets';
 import type {
   BottomTabBarProps,
   BottomTabDescriptorMap,
+  BottomTabEmitter,
   BottomTabHeaderProps,
   BottomTabNavigationConfig,
-  BottomTabNavigationHelpers,
   BottomTabNavigationOptions,
   BottomTabNavigationProp,
+  BottomTabViewState,
 } from '../types';
 import { BottomTabBarHeightCallbackContext } from '../utils/BottomTabBarHeightCallbackContext';
 import { BottomTabBarHeightContext } from '../utils/BottomTabBarHeightContext';
@@ -27,9 +23,12 @@ import { BottomTabBar, getTabBarHeight } from './BottomTabBar';
 import { MaybeScreen, MaybeScreenContainer } from './ScreenFallback';
 
 type Props = BottomTabNavigationConfig & {
-  state: TabNavigationState<ParamListBase>;
-  navigation: BottomTabNavigationHelpers;
+  state: BottomTabViewState;
   descriptors: BottomTabDescriptorMap;
+  emitter: BottomTabEmitter;
+  navigateToTab: (routeKey: string) => void;
+  preloadedRouteKeys: string[];
+  popNestedStackToTop: (routeKey: string) => void;
 };
 
 const EPSILON = 1e-5;
@@ -67,8 +66,11 @@ export function BottomTabView(props: Props) {
   const {
     tabBar = renderTabBarDefault,
     state,
-    navigation,
     descriptors,
+    emitter,
+    navigateToTab,
+    preloadedRouteKeys,
+    popNestedStackToTop,
     safeAreaInsets,
     detachInactiveScreens = Platform.OS === 'web' ||
       Platform.OS === 'android' ||
@@ -93,25 +95,13 @@ export function BottomTabView(props: Props) {
   React.useEffect(() => {
     const previousRouteKey = previousRouteKeyRef.current;
 
-    let popToTopAction: NavigationAction | undefined;
-
-    if (
+    const shouldPopPreviousToTop =
       previousRouteKey !== focusedRouteKey &&
-      descriptors[previousRouteKey]?.options.popToTopOnBlur
-    ) {
-      const prevRoute = state.routes.find((route) => route.key === previousRouteKey);
-
-      if (prevRoute?.state?.type === 'stack' && prevRoute.state.key) {
-        popToTopAction = {
-          ...StackActions.popToTop(),
-          target: prevRoute.state.key,
-        };
-      }
-    }
+      !!descriptors[previousRouteKey]?.options.popToTopOnBlur;
 
     const animateToIndex = () => {
       if (previousRouteKey !== focusedRouteKey) {
-        navigation.emit({
+        emitter.emit({
           type: 'transitionStart',
           target: focusedRouteKey,
         });
@@ -146,12 +136,12 @@ export function BottomTabView(props: Props) {
           })
           .filter(Boolean) as Animated.CompositeAnimation[]
       ).start(({ finished }) => {
-        if (finished && popToTopAction) {
-          navigation.dispatch(popToTopAction);
+        if (finished && shouldPopPreviousToTop) {
+          popNestedStackToTop(previousRouteKey);
         }
 
         if (previousRouteKey !== focusedRouteKey) {
-          navigation.emit({
+          emitter.emit({
             type: 'transitionEnd',
             target: focusedRouteKey,
           });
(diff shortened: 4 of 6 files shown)


Ubax:
* **#48291** <a href="https://app.graphite.com/github/pr/expo/expo/48291?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/48291?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>


<h2></h2>

packages/expo-router/src/react-navigation/bottom-tabs/types.tsx
```diff
@@ -409,9 +439,13 @@ export type BottomTabHeaderProps = {
...
 export type BottomTabBarProps = {
-  state: TabNavigationState<ParamListBase>;
+  state: BottomTabViewState;
   descriptors: BottomTabDescriptorMap;
-  navigation: NavigationHelpers<ParamListBase, BottomTabNavigationEventMap>;
+  emit: BottomTabEmit;
+  /**
+   * Switches to the tab with the given route name. Pass the tab's current params to keep them.
+   */
+  navigateToTab: (name: string, params?: object) => void;
```
Ubax: [💡 suggestion] The signature makes the pitfall your comment in `BottomTabBar.tsx` warns about possible: any caller that forgets to pass the current params resets the tab and, with `getId`, remounts it. Consider taking the route key instead (`navigateToTab(routeKey)`) and resolving `name`/`params` inside the navigator, where the state lives. That removes the footgun entirely and both warning comments with it. Trade-off: a tab bar could no longer navigate with *different* params — but nothing in this codebase does that today.

packages/expo-router/src/react-navigation/bottom-tabs/views/BottomTabBar.tsx
```diff
@@ -344,22 +336,22 @@ export function BottomTabBar({ state, navigation, descriptors, insets, style }:
...
 
             if (!focused && !event.defaultPrevented) {
-              navigation.dispatch({
-                ...CommonActions.navigate(route),
-                target: state.key,
-              });
+              // The params must be passed back: the router rebuilds the route from this action, so
+              // omitting them would reset the tab to its initial params and, with `getId`, would
+              // regenerate its key and remount the whole tab.
+              navigateToTab(route.name, route.params);
```
Ubax: [💡 suggestion] This params-preservation behavior (and the `popToTopOnBlur` path through `popNestedStackToTop`) carries the riskiest parity claims in the PR, but neither is covered by `__tests__/index.test.ios.tsx` — the existing tests only exercise plain tab press and preloading. A test that sets `getId`, navigates a tab to non-initial params, switches away and re-presses the tab, then asserts the tab kept its params (and didn't remount) would lock this in. Same for `popToTopOnBlur` with a nested stack.

packages/expo-router/src/react-navigation/bottom-tabs/navigators/createBottomTabNavigator.tsx
```diff
@@ -48,9 +53,40 @@ function BottomTabNavigator({
     UNSTABLE_router,
   });
 
+  // The views take discrete props instead of `navigation` and the raw state, so this factory adapts
+  // its `useNavigationBuilder` results to them.
```
Ubax:
```suggestion
```

packages/expo-router/src/react-navigation/bottom-tabs/navigators/createBottomTabNavigator.tsx
```diff
@@ -48,9 +53,40 @@ function BottomTabNavigator({
...
-      <BottomTabView {...rest} state={state} navigation={navigation} descriptors={descriptors} />
+      <BottomTabView
+        {...rest}
+        state={state}
+        descriptors={descriptors}
+        // The cast is what keeps `defaultPrevented` required on `BottomTabEmit`, so a custom tab bar
+        // can rely on it. TypeScript instantiates the generic `emit` with `EventName = any` here,
+        // which collapses the conditional that adds `defaultPrevented` to its return type. The
+        // property is present at runtime for `tabPress`.
+        emit={navigation.emit as BottomTabEmit}
```
Ubax: Can this be achieved without the cast?

packages/expo-router/src/react-navigation/bottom-tabs/views/BottomTabView.tsx
```diff
@@ -27,9 +23,23 @@ import { BottomTabBar, getTabBarHeight } from './BottomTabBar';
...
 
 type Props = BottomTabNavigationConfig & {
-  state: TabNavigationState<ParamListBase>;
-  navigation: BottomTabNavigationHelpers;
+  state: BottomTabViewState;
   descriptors: BottomTabDescriptorMap;
+  emit: BottomTabEmit;
+  /**
+   * Switches to the tab with the given route name. Pass the tab's current params to keep them.
+   */
```
Ubax:
```suggestion
```

packages/expo-router/src/react-navigation/bottom-tabs/views/BottomTabView.tsx
```diff
@@ -27,9 +23,23 @@ import { BottomTabBar, getTabBarHeight } from './BottomTabBar';
...
+  navigateToTab: (name: string, params?: object) => void;
+  /**
+   * Keys of the routes that were preloaded. Preloaded screens render and stay unfrozen while
+   * they are still blurred.
+   */
```
Ubax:
```suggestion
```

packages/expo-router/src/react-navigation/bottom-tabs/views/BottomTabView.tsx
```diff
@@ -27,9 +23,23 @@ import { BottomTabBar, getTabBarHeight } from './BottomTabBar';
...
+  preloadedRouteKeys: string[];
+  /**
+   * Pops the stack nested in `routeKey` back to its first screen. Does nothing when that route
+   * has no nested stack. Used to implement the `popToTopOnBlur` option.
+   */
```
Ubax:
```suggestion
```

packages/expo-router/src/react-navigation/bottom-tabs/navigators/createBottomTabNavigator.tsx
```diff
@@ -48,9 +53,40 @@ function BottomTabNavigator({
...
+    },
+    [navigation, state.key]
+  );
+
+  const popNestedStackToTop = useCallback(
+    (routeKey: string) => {
+      const nestedState = state.routes.find((route) => route.key === routeKey)?.state;
+      if (nestedState?.type === 'stack' && nestedState.key) {
+        navigation.dispatch({ ...StackActions.popToTop(), target: nestedState.key });
+      }
```
Ubax: Don't change the syntax if not needed. Keep the old version


```suggestion
      const prevRoute = state.routes.find((route) => route.key === routeKey);

      if (prevRoute?.state?.type === 'stack' && prevRoute.state.key) {
        popToTopAction = {
          ...StackActions.popToTop(),
          target: prevRoute.state.key,
        };
      }
```

packages/expo-router/src/react-navigation/bottom-tabs/views/BottomTabView.tsx
```diff
@@ -162,7 +163,15 @@ export function BottomTabView(props: Props) {
...
     previousRouteKeyRef.current = focusedRouteKey;
-  }, [descriptors, focusedRouteKey, navigation, state.index, state.routes, tabAnims]);
+  }, [
+    descriptors,
+    emit,
+    focusedRouteKey,
+    popNestedStackToTop,
+    state.index,
+    state.routes,
+    tabAnims,
```
Ubax, line 162:
```suggestion
    descriptors,
    focusedRouteKey,
    emit,
    popNestedStackToTop,
    state.index,
    state.routes,
    tabAnims,
```

Keep the old order

packages/expo-router/src/react-navigation/bottom-tabs/types.tsx
```diff
@@ -48,11 +47,42 @@ export type BottomTabNavigationEventMap = {
...
 
-export type BottomTabNavigationHelpers = NavigationHelpers<
-  ParamListBase,
-  BottomTabNavigationEventMap
-> &
-  TabActionHelpers<ParamListBase>;
+/**
+ * A tab as the views see it. This is only the part of the router's route that the views read —
+ * notably not `state`, so a custom tab bar cannot reach into a tab's nested navigator.
+ */
```
Ubax:
```suggestion
/**
 * @internal
 */
```

packages/expo-router/src/react-navigation/bottom-tabs/types.tsx
```diff
@@ -48,11 +47,42 @@ export type BottomTabNavigationEventMap = {
...
+export type BottomTabViewRoute = {
+  key: string;
+  name: string;
+  params?: object;
+};
+
+/**
+ * The navigator state consumed by `BottomTabView` and the tab bar. Everything else the views need
+ * is passed as a discrete prop, so they stay independent of the router that produced the state.
+ */
```
Ubax, line 59:
```suggestion
/**
 * The navigator state consumed by `BottomTabView` and the tab bar. 
 * @internal
 */
```

packages/expo-router/src/react-navigation/bottom-tabs/types.tsx
```diff
@@ -48,11 +47,42 @@ export type BottomTabNavigationEventMap = {
...
+  routes: BottomTabViewRoute[];
+};
+
+/**
+ * Emits a navigator-level tab event.
+ *
+ * `tabPress` returns the emitted event so the tab bar can skip navigation when a listener called
+ * `preventDefault()`.
+ */
+export type BottomTabEmit = {
```
Ubax: Align the emit with emitter from standard nav already

packages/expo-router/src/react-navigation/bottom-tabs/types.tsx
```diff
@@ -48,11 +47,42 @@ export type BottomTabNavigationEventMap = {
...
+  index: number;
+  routes: BottomTabViewRoute[];
+};
+
+/**
+ * Emits a navigator-level tab event.
+ *
+ * `tabPress` returns the emitted event so the tab bar can skip navigation when a listener called
+ * `preventDefault()`.
+ */
```
Ubax:
```suggestion
/**
 * @internal
 */
```

packages/expo-router/src/react-navigation/bottom-tabs/types.tsx
```diff
@@ -409,9 +439,13 @@ export type BottomTabHeaderProps = {
 };
 
 export type BottomTabBarProps = {
```
Ubax, line 425: Add TODO update docs

packages/expo-router/src/react-navigation/bottom-tabs/types.tsx
```diff
@@ -409,9 +439,13 @@ export type BottomTabHeaderProps = {
...
-  state: TabNavigationState<ParamListBase>;
+  state: BottomTabViewState;
   descriptors: BottomTabDescriptorMap;
-  navigation: NavigationHelpers<ParamListBase, BottomTabNavigationEventMap>;
+  emit: BottomTabEmit;
+  /**
+   * Switches to the tab with the given route name. Pass the tab's current params to keep them.
+   */
```
Ubax:
```suggestion
```

packages/expo-router/src/react-navigation/bottom-tabs/navigators/createBottomTabNavigator.tsx
```diff
@@ -48,9 +53,40 @@ function BottomTabNavigator({
     UNSTABLE_router,
   });
 
+  // The views take discrete props instead of `navigation` and the raw state, so this factory adapts
+  // its `useNavigationBuilder` results to them.
+  const navigateToTab = useCallback(
+    (name: string, params?: object) => {
+      navigation.dispatch({ ...CommonActions.navigate(name, params), target: state.key });
```
Ubax: Align with previous implementation

```
navigation.dispatch({
                ...CommonActions.navigate(route),
                target: state.key,
              });
```

Ubax:
I'm intentionally not adding CHANGELOG entry, since this is only a refactor

packages/expo-router/src/react-navigation/bottom-tabs/types.tsx
```diff
@@ -408,10 +428,12 @@ export type BottomTabHeaderProps = {
   navigation: BottomTabNavigationProp<ParamListBase>;
 };
 
+// TODO: update docs
 export type BottomTabBarProps = {
-  state: TabNavigationState<ParamListBase>;
+  state: BottomTabViewState;
```
jakub-agent, line 426: [🚨 **critical**] `BottomTabBarProps` is public API, so dropping `navigation` and narrowing `state` is a breaking change for userland custom tab bars. `src/layouts/Tabs.tsx` re-exports this module wholesale, and it backs both the `expo-router/tabs` and `expo-router/js-tabs` subpaths.

I confirmed both halves on this commit by rendering a navigator with a custom `tabBar`:

- Runtime: the props handed to `tabBar` are `descriptors, emitter, insets, navigateToTab, state` — `props.navigation` is `undefined`. Any existing tab bar calling `navigation.emit(...)` / `navigation.navigate(...)` / `navigation.dispatch(...)` throws.
- Types: `'navigation' extends keyof BottomTabBarProps` is `false`, and `'key' extends keyof BottomTabBarProps['state']` is `false`. The runtime object still carries the full state (`history`, `key`, `preloadedRouteKeys`, `routeNames`, `stale`, `type`), so `state.key` keeps working but no longer type-checks.

This looks avoidable rather than inherent to the migration. `createStandardDrawerNavigator` already migrated the drawer to standard-navigation while keeping `navigation` and the full state on its public `DrawerContentComponentProps`, by surfacing them through `createProps` (`DrawerNavigatorCreateProps`) — the escape hatch `StandardNavigatorCreatePropsFactoryDeps` exists for. The same approach would let the internal `BottomTabView` props narrow without touching the public tab bar contract.

Suggestion: split the public `BottomTabBarProps` (unchanged, `navigation` + `TabNavigationState`) from an internal `BottomTabViewProps`. If `navigation` really has to go, keep it as a deprecated forwarding field for a release and add a `CHANGELOG.md` entry — the base PR in this stack added one, this PR has none.

packages/expo-router/src/react-navigation/bottom-tabs/types.tsx
```diff
@@ -35,24 +35,44 @@ export type BottomTabNavigationEventMap = {
   /**
    * Event which fires on long press on the tab in the tab bar.
    */
-  tabLongPress: { data: undefined };
+  tabLongPress: { data: undefined; canPreventDefault: false };
```
jakub-agent, line 38: [🏗️ **design**] Baking `canPreventDefault` into the exported `BottomTabNavigationEventMap` diverges from how the drawer solved the same problem. `createDrawerNavigator.tsx` leaves the public `DrawerNavigationEventMap` upstream-shaped and derives a standard-compliant map:

```ts
export type StandardDrawerNavigationEventMap = {
  [Event in keyof DrawerNavigationEventMap]: DrawerNavigationEventMap[Event] & {
    canPreventDefault: Event extends 'drawerItemPress' ? true : false;
  };
};
```

A derived `StandardBottomTabsNavigationEventMap` would satisfy `StandardNavigatorEventMapBase` without editing a public type, and would keep the fork's event map matching `@react-navigation/bottom-tabs`. Low risk either way for listeners (`EventArg<..., false>` and `EventArg<..., undefined>` both omit `preventDefault`), so this is about consistency, not correctness.

packages/expo-router/src/react-navigation/bottom-tabs/types.tsx
```diff
@@ -408,10 +428,12 @@ export type BottomTabHeaderProps = {
   navigation: BottomTabNavigationProp<ParamListBase>;
 };
 
+// TODO: update docs
```
jakub-agent: [💡 suggestion] This `TODO` sits on a public exported type, and `AGENTS.md` asks for guides plus the typedoc reference to be updated alongside the change. Either update the docs here or point the TODO at a tracking issue, so it does not ship as a bare marker on public API.

Related: `emitter` and `navigateToTab` are now the only way a custom tab bar can act, but neither has a doc comment. `navigation` was at least self-describing by convention; these are not.

packages/expo-router/src/react-navigation/bottom-tabs/types.tsx
```diff
@@ -35,24 +35,44 @@ export type BottomTabNavigationEventMap = {
...
+  key: string;
+  name: string;
+  params?: object;
+};
+
+/**
+ * The navigator state consumed by `BottomTabView` and the tab bar.
+ * @internal
+ */
+export type BottomTabViewState = {
```
jakub-agent, line 60: [💡 suggestion] `@internal` is not accurate for these: `BottomTabViewState`, `BottomTabViewRoute` and `BottomTabEmitter` are all reachable through the public `BottomTabBarProps`, so they are part of the public surface whether or not they are documented as such.

They are also not exported from `bottom-tabs/index.tsx`, so a custom tab bar author cannot name the state or emitter type it receives and has to fall back to indexed access (`BottomTabBarProps['state']`). Either export them or keep the public props typed with the already-exported `TabNavigationState<ParamListBase>`.

packages/expo-router/src/react-navigation/bottom-tabs/navigators/createBottomTabNavigator.tsx
```diff
@@ -48,9 +53,47 @@ function BottomTabNavigator({
     UNSTABLE_router,
   });
 
+  const emitter = useStandardEmitter(navigation);
+
+  const navigateToTab = useCallback(
+    (routeKey: string) => {
+      const route = state.routes.find((route) => route.key === routeKey);
+
+      if (route) {
```
jakub-agent, line 62: [💡 suggestion] `navigateToTab` silently does nothing when the key is not found. Inside `BottomTabBar` the key always comes from `state.routes`, so this is unreachable — but `navigateToTab` is now part of `BottomTabBarProps`, i.e. userland API, and it takes a route *key* while the obvious thing to pass is `route.name`. A custom tab bar making that mistake gets a dead tab and no signal.

Worth a `__DEV__` warning in the `else` branch naming the unknown key and pointing at `route.key`, in line with the repo's what/why/how error-message guidance.

packages/expo-router/src/react-navigation/bottom-tabs/types.tsx
```diff
@@ -35,24 +35,44 @@ export type BottomTabNavigationEventMap = {
...
+  index: number;
+  routes: BottomTabViewRoute[];
+};
+
+/**
+ * @internal
+ */
+export type BottomTabEmitter = NavigatorArgs<
```
jakub-agent, line 65: [🐧 nit] The `Record<string, never>` filler for `NavigatorOptions` exists only to index out `emitter`, and `useStandardEmitter` already does the identical dance — so the workaround now lives in two places. `ReturnType<typeof useStandardEmitter<BottomTabNavigationEventMap>>` would reuse it, or `standard-navigation` could export a standalone `NavigatorEmitter<EventMap>` and both call sites could drop the filler.

jakub-agent (changes requested):

PR 2/2 in the stack (base: `_ubax/07-29-_expo-router_add_processscreens_option_to_standard-navigation_integration`).

The refactor itself is clean and the mechanics check out. I verified locally on `8905e9e`:

- `tsc -b tsconfig.all.json` passes (validated the run with an injected error, so it is not a no-op).
- All 6 tests in `bottom-tabs/__tests__/index.test.ios.tsx` pass.
- Both new tests are effective guards, not decoration. Changing `CommonActions.navigate(route)` to `navigate(route.name)` fails the `getId` test; disabling the nested-stack lookup fails the `popToTopOnBlur` test.
- I probed the new `emitter` path with throwaway tests: `tabPress` + `preventDefault` still blocks the tab switch, and `transitionStart` / `transitionEnd` / `tabLongPress` all still reach screen-level `navigation.addListener`. No regression there.
- `popToTopOnBlur` timing is preserved: the old code snapshotted the action at effect time, the new `popNestedStackToTop` closure snapshots the same `state.routes` from the same render.

One blocker. `BottomTabBarProps` is public API — `src/layouts/Tabs.tsx` does `export * from '../react-navigation/bottom-tabs'`, and that module backs both the `expo-router/tabs` and `expo-router/js-tabs` subpaths. Removing `navigation` breaks any userland custom `tabBar` at runtime, not just at the type level (I confirmed `props.navigation === undefined`). See the inline comment for the drawer precedent, which kept `navigation` while migrating to standard-navigation.

Smaller items: the public `BottomTabNavigationEventMap` is edited in place rather than derived (the drawer derives `StandardDrawerNavigationEventMap` via a mapped type), a `// TODO: update docs` is left on a public type, and there is no `CHANGELOG.md` entry — the base PR in this stack added one.

packages/expo-router/src/react-navigation/bottom-tabs/types.tsx
Ubax: This is intentional. Only the top PR in the stack adds one - https://github.com/expo/expo/pull/48292

packages/expo-router/src/react-navigation/bottom-tabs/types.tsx
Ubax: I prefer the new shape. This is breaking-change anyway, and I prefer better alignment with standard navigation

packages/expo-router/src/react-navigation/bottom-tabs/types.tsx
Ubax: Add reference to this ticket - https://linear.app/expo/issue/ENG-25579/update-documentation-after-the-refactor 

Also add `(@ubax)`

packages/expo-router/src/react-navigation/bottom-tabs/types.tsx
Ubax: Remove all the `@internal` comments

packages/expo-router/src/react-navigation/bottom-tabs/navigators/createBottomTabNavigator.tsx
Ubax: Add the warning

packages/expo-router/src/react-navigation/bottom-tabs/types.tsx
```diff
@@ -35,24 +35,37 @@ export type BottomTabNavigationEventMap = {
...
+export type BottomTabViewRoute = {
+  key: string;
+  name: string;
+  params?: object;
+};
+
+/**
+ * The navigator state consumed by `BottomTabView` and the tab bar.
+ */
+export type BottomTabViewState = {
```
jakub-agent, line 60: [🐧 nit] Dropping the `@internal` tags settles the intent, thanks. One gap remains: `BottomTabViewRoute`, `BottomTabViewState` and `BottomTabEmitter` are still absent from the `export type { … }` block in `bottom-tabs/index.tsx`, so a custom tab bar author cannot import them by name and has to write `BottomTabBarProps['state']` or `['emitter']` instead.

Adding the three to that export list would finish the job. Non-blocking.

jakub-agent (approved):

**Iteration 2** — re-reviewed at `ea54372` (was `8905e9e`, force-pushed). PR 2/2 of the visible stack; the full chain is #48356 → #48290 → **#48291** → #48292.

All six items from my first review are resolved, either by a change or by an explicit decision from you. Nothing regressed.

| # | Item | Outcome |
|---|---|---|
| 1 | `BottomTabBarProps` breaking change | Your call — accepted |
| 2 | Event map edited in place | Your call — accepted |
| 3 | `@internal` on publicly reachable types | Tags removed; one gap left, see inline |
| 4 | `// TODO: update docs` | Now owned and tracked (ENG-25579) |
| 5 | `Record<string, never>` filler (nit) | Unchanged, no reply — fine to leave |
| 6 | `navigateToTab` silent no-op | Fixed, verified |

**Verified on `ea54372`:**

- 6/6 tests pass in `bottom-tabs/__tests__/index.test.ios.tsx`.
- `oxlint` clean on `src/react-navigation/bottom-tabs`.
- The new `__DEV__` warning behaves correctly. I probed it: passing a route *name* (`'B'`) logs the warning once; passing `route.key` logs nothing and switches the tab. The `__DEV__` + `console.warn` shape matches the rest of the package (`useScreens.tsx`, `StackTitle.tsx`).
- The `emitter` dep reorder in `BottomTabView` matches the ordering you asked for in your own earlier comment, so I read it as deliberate.

**On the two items you decided against.** Both are reasonable, and I checked the claim behind #1: #48292 does add an entry under `### 🛠 Breaking changes`. One residual worth considering *there*, not here — the entry reads "Migrate the JS `Tabs` navigator to the `standard-navigation` integration", which does not tell a reader with a custom `tabBar` that their code breaks. Naming the concrete change (`tabBar` props lose `navigation`, `state` narrows) would make it actionable.

**Heads-up, not caused by this PR.** `pnpm typecheck` (`tsc -b tsconfig.all.json`) fails with 8 errors in `src/__tests__/custom-navigators.test.ios.tsx`, all `TS2322`/`TS2345`/`TS2339` around `ScreenProps` and `processScreens`. I bisected the stack:

| Commit | Typecheck |
|---|---|
| `2af1241` (origin/main tip) | 0 errors |
| `5f29e3d` (#48356, bottom of stack) | 8 errors |
| `70e0643` (#48290) | 8 errors |
| `ea54372` (**this PR**) | 8 errors |

So the break enters at #48356 and this PR adds none. CI here will be red until the bottom of the stack is fixed.

PR #48292, opened 2026-07-29
[expo-router] integrate JS tabs with standard navigation

Ubax:
# Why

To unify all navigators under single standard-navigation API

# How

Replace `withLayoutContext` + `createNavigatorFactory` with `unstable_integrateWithRouter` and adjust API

# Test Plan

1. CI
2. Router e2e

# Checklist


Ubax:
* **#48292** <a href="https://app.graphite.com/github/pr/expo/expo/48292?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/48292?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>


<h2></h2>

packages/expo-codemod/src/transforms/sdk-56-expo-router-react-navigation-replace.ts
```diff
@@ -39,6 +39,8 @@ const IMPORT_MAP: Record<string, string> = {
   '@react-navigation/core': 'expo-router/react-navigation',
   '@react-navigation/routers': 'expo-router/react-navigation',
   '@react-navigation/stack': 'expo-router/js-stack',
+  // TODO: `expo-router/js-tabs` no longer exports `createBottomTabNavigator` — rewrite that
```
jakub-agent: [🚨 **critical**] This leaves the codemod emitting code that cannot work. `import { createBottomTabNavigator } from '@react-navigation/bottom-tabs'` is rewritten to `import { createBottomTabNavigator } from 'expo-router/js-tabs'`, and after this PR that specifier does not exist — the user gets a `TypeError` at `createBottomTabNavigator()`. Since `createBottomTabNavigator` is the primary export of `@react-navigation/bottom-tabs`, this hits essentially every project the codemod is aimed at, and `docs/pages/router/migrate/sdk-55-to-56.mdx:49` documents the same mapping.

The existing tests make this worse rather than catching it: `__tests__/sdk-56-expo-router-react-navigation-replace-test.ts:40-41` (and lines 70, 187, 196) assert the broken output, so CI stays green.

Two options that fit the file's existing structure:
- rewrite the specifier to `createStandardBottomTabNavigator` (the signature differs, so this is only correct if a call-site transform comes with it), or
- move `@react-navigation/bottom-tabs` into `UNSUPPORTED_PACKAGES` with a `Use the \`Tabs\` layout instead` message, which is exactly how `@react-navigation/drawer` is handled after its equivalent migration in #47839.

The second is cheap and fails loudly instead of silently. Either way the codemod tests need updating in the same change, otherwise the follow-up TODO has nothing signalling it.

packages/expo-router/src/layouts/Tabs.tsx
```diff
@@ -1,9 +1,8 @@
-import { Screen } from '../views/Screen';
 import Tabs from './TabsClient';
 
 export * from '../react-navigation/bottom-tabs';
 
-Tabs.Screen = Screen;
+export type { BottomTabNavigatorProps } from './TabsClient';
```
jakub-agent: [🚨 **critical**] `BottomTabNavigatorProps` is a released public export with a different meaning. In `expo-router@57.0.9`, `expo-router/js-tabs` exports it (via `export * from '../react-navigation/bottom-tabs'`) as `DefaultNavigatorOptions<…> & TabRouterOptions & BottomTabNavigationConfig` — react-navigation navigator props. Here the same name is re-bound to `ComponentProps<typeof Tabs>`.

Reusing the name is worse than removing it: `import type { BottomTabNavigatorProps } from 'expo-router/js-tabs'` still resolves, so users get no error at the import and instead get confusing failures wherever the type is used (e.g. it no longer has `backBehavior` typed the same way, and gains layout-only props). A rename such as `TabsProps` — mirroring how the removal is intentional — would turn this into a clean, diagnosable break.

For reference, `layouts/Drawer.tsx` exports `DrawerNavigatorProps`, but `react-navigation/drawer` never exported that name, so it did not collide.

packages/expo-router/src/react-navigation/bottom-tabs/navigators/createBottomTabNavigator.tsx
```diff
@@ -1,117 +1,67 @@
...
-    initialRouteName,
-    backBehavior,
-    children,
-    layout,
-    screenListeners,
-    screenOptions,
-    screenLayout,
-    UNSTABLE_router,
-  });
+export type TabsScreenOptions = BottomTabNavigationOptions & { href?: Href | null };
```
jakub-agent: [🏗️ **design**] `TabsScreenOptions` includes `href`, but this navigator does not implement it — the `href` handling lives entirely in `processScreens` in `layouts/TabsClient.tsx`. Because `createStandardBottomTabNavigator` is exported publicly (`../index.tsx:29`) and typed with these options, someone wiring it up themselves (`unstable_integrateWithRouter(createStandardBottomTabNavigator, TabRouter)`, which the `unstable_integrateWithRouter` docblock presents as the supported path) gets `href` accepted by TypeScript and silently ignored at runtime. `href: null` would fail to hide the tab with no error.

There is also a layering point: `href` is an Expo Router concept, and this file is inside the react-navigation fork — it now reaches back for `Href` from `../../../types`. Before this PR the same type lived in `layouts/TabsClient.tsx`, on the router side of the boundary.

Typing the navigator with plain `BottomTabNavigationOptions` and widening to `BottomTabNavigationOptions & { href?: Href | null }` in `TabsClient.tsx` — where `processScreens` actually consumes it — keeps the fork free of router types and makes the standalone export honest about what it supports.

packages/expo-router/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - Add `redirectTo` to protected routes and render guarded screens as redirects instead of removing them from navigators. ([#47744](https://github.com/expo/expo/pull/47744) by [@Ubax](https://github.com/Ubax))
 - Migrate the `Drawer` navigator to the `standard-navigation` integration. ([#47839](https://github.com/expo/expo/pull/47839) by [@Ubax](https://github.com/Ubax))
 - Remove `StackNavigationState.preloadedRoutes` property and append preloaded routes to `state.routes` ([#47961](https://github.com/expo/expo/pull/47961) by [@Ubax](https://github.com/Ubax))
+- Migrate the JS `Tabs` navigator to the `standard-navigation` integration. ([#48292](https://github.com/expo/expo/pull/48292) by [@Ubax](https://github.com/Ubax))
```
jakub-agent, line 10: [🏗️ **design**] Worth listing the removed public exports here, since the entry reads as an internal refactor. Verified against the `expo-router@57.0.9` tarball, `expo-router/js-tabs` currently exports all three of:

- `createBottomTabNavigator` (value) — removed
- `BottomTabNavigatorProps` (type) — removed and the name rebound to a different type
- `BottomTabNavigator` (type, from `TabsClient`) — removed

Someone hitting one of these on upgrade has no way to find out from the changelog what replaced it. A sentence naming `createStandardBottomTabNavigator` and the `Tabs` layout as the replacements would cover it. The #47839 `Drawer` entry has the same gap for `createDrawerNavigator`, so this may be worth a single combined note.

packages/expo-router/src/react-navigation/bottom-tabs/__tests__/types.test.ts
```diff
@@ -0,0 +1,46 @@
...
+  Equal<BottomTabNavigatorContentProps['popNestedStackToTop'], (routeKey: string) => void>
+>;
+
+// The tab bar config stays a public navigator prop.
+export type _DetachInactiveScreensIsPublic = Expect<
+  Equal<TabsProps['detachInactiveScreens'], boolean | undefined>
+>;
+
+// `href` is a screen option of the `Tabs` layout: a link target, or `null` to hide the tab.
+type TabsScreenOptions = Exclude<
```
jakub-agent: [💡 suggestion] Having to reconstruct `TabsScreenOptions` with `Exclude<…, ((...args: any) => any) | undefined>` is a signal it should be exported. The real type exists at `../navigators/createBottomTabNavigator.tsx:15` but is not reachable from `expo-router/js-tabs`, so a user cannot write a helper returning `Tabs.Screen` options, or type a shared options object, without the same gymnastics.

Exporting it from `layouts/Tabs.tsx` would let this test import it directly and shrink to the two `_hiddenTab`/`_linkedTab` assertions.

packages/expo-router/src/react-navigation/bottom-tabs/index.tsx
```diff
@@ -25,17 +20,26 @@ export { BottomTabBarHeightCallbackContext } from './utils/BottomTabBarHeightCal
...
   BottomTabBarProps,
+  BottomTabEmitter,
   BottomTabHeaderProps,
   BottomTabNavigationEventMap,
   BottomTabNavigationOptions,
   BottomTabNavigationProp,
-  BottomTabNavigatorProps,
   BottomTabOptionsArgs,
   BottomTabScreenProps,
+  BottomTabViewState,
```
jakub-agent, line 41: [💡 suggestion] `BottomTabViewState` and `BottomTabEmitter` are both tagged `@internal` in `types.tsx` (lines 58 and 68), yet they are now public exports. They are reachable from public API — `BottomTabBarProps.state` and `.emitter` reference them — so exporting them is right, but the `@internal` tags are now wrong and will keep them out of the generated docs.

`BottomTabDescriptorMap` has the same status (`BottomTabBarProps.descriptors`) but is not exported, so it is still unnameable. The new custom-`tabBar` test only avoids this by destructuring. Worth exporting all three and dropping the `@internal` tags from the two that are part of `BottomTabBarProps`.

apps/router-e2e/__e2e__/native-navigation/app/js-tabs/index.tsx
```diff
@@ -8,6 +8,12 @@ export default function Home() {
     <View style={{ flex: 1, justifyContent: 'center', alignItems: 'center' }}>
       <Text testID="e2e-screen">Home</Text>
       <Text testID="e2e-pathname">{pathname}</Text>
+      <Link testID="e2e-goto-profile" href="/js-tabs/profile/details">
```
jakub-agent: [💡 suggestion] Three links now share `testID="e2e-goto-profile"` (lines 11, 14, 17), pointing at two different hrefs and differing in `withAnchor`. Any selector-based runner resolving that ID gets an ambiguous match, and `getByTestId` would throw outright.

Suggest distinct IDs, e.g. `e2e-goto-profile-details`, `e2e-goto-profile-details-anchor`, and leaving `e2e-goto-profile` on the existing link so the current flow keeps working.

packages/expo-router/src/react-navigation/bottom-tabs/__tests__/index.test.ios.tsx
```diff
@@ -127,116 +107,233 @@ test('tab bar cannot be tapped when hidden', async () => {
...
-  const B = ({ route }: BottomTabScreenProps<ParamList, 'B'>) => {
-    React.useEffect(() => {
-      onMountB();
-    }, []);
-    return <Text>B user: {route.params?.user ?? 'none'}</Text>;
-  };
+  expect(screen).toHavePathname('/');
+});
+
+test('keeps the search params of a tab when it is re-selected', async () => {
```
jakub-agent, line 144: [💡 suggestion] This replaces the deleted `keeps the params and does not remount the tab on re-press with getId` test, but drops two of its assertions: that the screen is not remounted (the old test tracked `onMountB` call count), and any coverage of `getId` at all.

Remount-on-re-press is the behaviour most likely to regress from this change — `TabRouter`'s `NAVIGATE` case regenerates `route.key` when `getId` produces a different id (`TabRouter.tsx:437`), and `navigateToTab` now goes through `actions.navigate(name, params)` instead of a keyed `CommonActions.navigate(route)`. Adding a mount counter to the `second` screen here would restore the stronger assertion cheaply. If `getId` is intentionally out of scope for the `Tabs` layout, saying so in the PR description would help.

packages/expo-router/src/react-navigation/bottom-tabs/index.tsx
```diff
@@ -25,17 +20,26 @@ export { BottomTabBarHeightCallbackContext } from './utils/BottomTabBarHeightCal
 export { BottomTabBarHeightContext } from './utils/BottomTabBarHeightContext';
 export { useBottomTabBarHeight } from './utils/useBottomTabBarHeight';
 
+/**
+ * Navigators
+ */
+/**
+ * @unstable Prefer the `Tabs` layout.
```
jakub-agent: [🐧 nit] Two stacked docblocks read oddly — the `Navigators` section header and the export's own comment could be merged, or the header dropped as it now labels a single export.

Also, `@unstable` is the only occurrence of that tag in the package, so nothing consumes it. The convention used elsewhere for this is the blockquote form, e.g. `> **warning** This API is unstable and may change between minor releases.` in `standard-navigation/index.tsx`.

jakub-agent (changes requested):

The migration itself is clean and matches the `Drawer` precedent from #47839 closely: `createProps` for `preloadedRouteKeys`/`popNestedStackToTop`, `processScreens` for the `href` shortcut, and the same `descriptors` cast with the same SDK-58 TODO. I traced the one behavioral change I was unsure about — `navigateToTab` switching from `CommonActions.navigate(route)` to `actions.navigate(route.name, route.params)` — and it is equivalent, because `TabRouter` matches `NAVIGATE` on `payload.name` only and preserves `route.path` when `payload.path` is absent. Dropping `useCallback` from `navigateToTab` is also harmless, since `descriptors` is rebuilt every render anyway. Test coverage improved: the suite now exercises the public `Tabs` layout instead of the raw navigator, and adds `href: null`, custom `tabBar`, and `tabPress` `preventDefault` cases.

The main problems are at the package boundary, not in the navigator.

**`expo-codemod` now emits broken imports.** `expo-router/js-tabs` no longer exports `createBottomTabNavigator`, but the codemod still rewrites `@react-navigation/bottom-tabs` to it. The TODO acknowledges this, and the codemod's own tests assert the broken output, so CI stays green while the documented SDK 55→56 migration path silently produces code that fails at runtime.

**Released exports are removed without a changelog note.** `expo-router@57.0.9` publicly exports `createBottomTabNavigator`, the type `BottomTabNavigatorProps`, and the type `BottomTabNavigator` from `expo-router/js-tabs`. All three go away here. `BottomTabNavigatorProps` is the worst case: the name is reused for an unrelated type, so importing code keeps compiling and breaks further downstream.

Smaller items: `createStandardBottomTabNavigator` is exported with an options type that advertises an `href` option it does not implement, `TabsScreenOptions` is not publicly nameable (the new type test has to reconstruct it), and three e2e links share one `testID`.

Since this is a stacked PR on top of `_ubax/07-30-_router_add_tabs_rsc_tests`, I reviewed only the diff this PR owns.

packages/expo-codemod/src/transforms/sdk-56-expo-router-react-navigation-replace.ts
Ubax: Just remove that comment. 

This transformer is only for sdk-56. It still works for 55->56 migration. For 56/57->58 we will need to add a new codemod

packages/expo-router/src/layouts/Tabs.tsx
Ubax: Rename to `JSTabsProps`

packages/expo-router/src/react-navigation/bottom-tabs/index.tsx
Ubax: Remove Navigators and change the `@unstable` to `> **warning** This API is unstable and may change between minor releases`

packages/expo-router/CHANGELOG.md
Ubax: This CHANGELOG is enough. We already discourage people from using the raw `create` functions

packages/expo-router/src/layouts/Tabs.tsx
```diff
@@ -3,7 +3,7 @@ import { Screen } from '../views/Screen';
 import Tabs from './TabsClient';
 
 export {
-  createBottomTabNavigator,
+  createStandardBottomTabNavigator,
```
jakub-agent, line 6: [🚨 **critical**] This re-export block is broken as of this push. All six names in it — `createStandardBottomTabNavigator`, `BottomTabBar`, `BottomTabView`, `BottomTabBarHeightCallbackContext`, `BottomTabBarHeightContext`, `useBottomTabBarHeight` — come `from './TabsClient'`, but this PR deletes `export * from '../react-navigation/bottom-tabs'` from `TabsClient.tsx` (it moves to the new `TabsReactNavigationClient.tsx`). After the move, `TabsClient.tsx` exports exactly three things: `TabsScreenOptions`, `JSTabsProps`, and the default.

So this is six × TS2305 at compile time, and if it got past that, `expo-router/js-tabs` would hand out `undefined` for every one of them — including `BottomTabView` and `useBottomTabBarHeight`, which custom tab bars depend on.

The fix is to point this block at `./TabsReactNavigationClient`, which is where the barrel now lives and which nothing currently imports. `layouts/TopTabs.tsx` shows the shape that still works: it re-exports from `./TopTabsClient`, and `TopTabsClient.tsx:16` retains its `export * from '../react-navigation/material-top-tabs'`.

Worth checking why CI did not stop this — `check-packages` was still pending when I looked, so it may just be timing.

packages/expo-router/src/layouts/TabsReactNavigationClient.tsx
```diff
@@ -0,0 +1,4 @@
+'use client';
+
+// Keep React Navigation in a separate client file so the entry evaluates in React Server Components.
+export * from '../react-navigation/bottom-tabs';
```
jakub-agent: [🏗️ **design**] Nothing in the repo imports this file, or `TopTabsReactNavigationClient.tsx` added alongside it — I grepped the whole tree for `ReactNavigationClient` and the only hits are the two new files themselves. As written they are dead code, and the barrel they hold is the one `layouts/Tabs.tsx` is currently failing to find.

Wiring `Tabs.tsx` to this file resolves both that and this. `TopTabsReactNavigationClient.tsx` is a different case: `layouts/TopTabs.tsx` re-exports from `./TopTabsClient`, which still has its own `export *`, so top tabs are not broken and this second file has no consumer at all. If it is groundwork for a follow-up, it may be cleaner to land it with that change — top tabs are otherwise outside this PR's scope.

packages/expo-router/src/react-navigation/bottom-tabs/index.tsx
```diff
@@ -25,17 +20,23 @@ export { BottomTabBarHeightCallbackContext } from './utils/BottomTabBarHeightCal
...
   BottomTabBarProps,
+  BottomTabEmitter,
   BottomTabHeaderProps,
   BottomTabNavigationEventMap,
   BottomTabNavigationOptions,
   BottomTabNavigationProp,
-  BottomTabNavigatorProps,
   BottomTabOptionsArgs,
   BottomTabScreenProps,
+  BottomTabViewState,
```
jakub-agent, line 41: [🏗️ **design**] Unchanged from iteration 2 — these two exports still don't reach users. `expo-router/js-tabs` resolves to `layouts/Tabs.tsx`, whose type re-export list omits `BottomTabEmitter` and `BottomTabViewState`. There is no `./react-navigation` key in `package.json`'s `exports` map, and `src/react-navigation/index.ts` only re-exports `./native` and `./elements`, so this barrel has no public route of its own.

The effect is the case the new tests exercise: someone writing a custom `tabBar` types it as `BottomTabBarProps` and still cannot name the types of `state`, `emitter`, or `descriptors` — `BottomTabDescriptorMap` is not exported at all. Adding all three to the type list in `layouts/Tabs.tsx` would close it.

Separately, `BottomTabViewState` and `BottomTabEmitter` are still tagged `@internal` in `types.tsx` (lines 58 and 68) while being referenced by the public `BottomTabBarProps`. If they are public now, drop the tags; if they are meant to stay internal, drop these exports instead.

packages/expo-router/CHANGELOG.md
jakub-agent, line 10: [🏗️ **design**] Unchanged from iteration 2 — still reads as an internal refactor, though it removes an export that ships in `expo-router@57.0.9`:

- `createBottomTabNavigator` (value) → `createStandardBottomTabNavigator`
- `BottomTabNavigatorProps` (type) → no direct equivalent; `JSTabsProps` is the props type of the `Tabs` layout, not of the react-navigation navigator

Naming both with their replacements is what someone hitting the break on upgrade needs, and it gives the planned 56/57→58 codemod something to point at.

apps/router-e2e/__e2e__/native-navigation/app/js-tabs/index.tsx
```diff
@@ -8,6 +8,12 @@ export default function Home() {
     <View style={{ flex: 1, justifyContent: 'center', alignItems: 'center' }}>
       <Text testID="e2e-screen">Home</Text>
       <Text testID="e2e-pathname">{pathname}</Text>
+      <Link testID="e2e-goto-profile-details" href="/js-tabs/profile/details">
```
jakub-agent, line 11: [🐧 nit] The three-way collision is fixed, but `e2e-goto-profile-details` is now used both here (the Home tab) and in `profile/index.tsx:93` (the Profile tab). Tabs stay mounted once visited, so after a user has been to both tabs the ID matches two elements at the same time.

A prefix per screen — `e2e-home-goto-profile-details` and `e2e-profile-goto-details` — keeps the IDs unambiguous no matter what is mounted.

jakub-agent (changes requested):

Iteration 3, against `6436ae45` (force-pushed over `de4ce2f4`).

**This push looks like a half-applied refactor.** `export * from '../react-navigation/bottom-tabs'` was moved out of `TabsClient.tsx` into a new `TabsReactNavigationClient.tsx`, but `layouts/Tabs.tsx` was never repointed — it still re-exports six names from `./TabsClient`, which no longer has them. That is a compile error and would leave `expo-router/js-tabs` handing out `undefined` for `BottomTabBar`, `BottomTabView`, and `createStandardBottomTabNavigator`. The two new `*ReactNavigationClient.tsx` files are unreferenced by anything in the repo. `check-packages` and `jest` were still pending when I looked, so this is from reading the tree, not from a red run — but it is a one-line fix either way.

Everything I raised in iteration 2 that was substantive is otherwise unchanged: the unreachable `BottomTabEmitter`/`BottomTabViewState` exports, and the changelog wording.

The migration logic itself I have now checked across three revisions and still consider correct — `navigateToTab` via `actions.navigate(name, params)` matches `TabRouter`'s name-based `NAVIGATE` handling, and the restored mount-counter test covers the re-selection case.

Two minor things not worth inline comments, since they fall outside the diff hunks:
- `Tabs.Screen = Screen` and `Tabs.Protected = Protected` in `layouts/Tabs.tsx` are dead — `withLayoutContext` already `Object.assign`s the same two modules onto what `unstable_integrateWithRouter` returns.
- `Tabs.tsx` now reaches into `../react-navigation/bottom-tabs/TransitionConfigs/*` directly for the three transition namespaces. Those paths exist, so it works, but it pulls react-navigation modules straight into the entry that `TabsReactNavigationClient.tsx` exists to keep them out of.

Standing corrections from earlier iterations: the `sdk-56` codemod is versioned to the 55→56 upgrade and its output is correct for that SDK — disregard my first-iteration finding on it. And `BottomTabNavigator` was never a public export.

packages/expo-router/src/react-navigation/bottom-tabs/index.tsx
jakub-agent, line 41: [🏗️ **design**] Still unreachable at this revision. `expo-router/js-tabs` resolves to `layouts/Tabs.tsx`, and both of its re-export lists are explicit: the value list pulls six names from `./TabsClient`, and the type list names eight from `../react-navigation/bottom-tabs` — neither includes `BottomTabEmitter` or `BottomTabViewState`. The `export *` inside `TabsClient.tsx` does not help, because `Tabs.tsx` re-exports a named subset of it rather than forwarding it. There is no `./react-navigation` key in `package.json`'s `exports` map either, and `src/react-navigation/index.ts` only re-exports `./native` and `./elements`.

So the two additions here have no user-visible effect. The case that motivates them is the one the new tests exercise: someone writing a custom `tabBar` types it as `BottomTabBarProps` and cannot name the types of `state`, `emitter`, or `descriptors` — `BottomTabDescriptorMap` is not exported at all. Adding all three to the type list in `layouts/Tabs.tsx` closes it.

Also still outstanding: `BottomTabViewState` and `BottomTabEmitter` carry `@internal` in `types.tsx` (lines 58 and 68) while being referenced by the public `BottomTabBarProps`. If they are public now, drop the tags; if they should stay internal, drop these exports instead.

packages/expo-router/CHANGELOG.md
jakub-agent, line 10: [🏗️ **design**] Unchanged across all four revisions — still reads as an internal refactor, though it removes an export that ships in `expo-router@57.0.9`:

- `createBottomTabNavigator` (value) → `createStandardBottomTabNavigator`
- `BottomTabNavigatorProps` (type) → no direct equivalent; `JSTabsProps` is the props type of the `Tabs` layout, not of the react-navigation navigator

Naming both with their replacements is what someone hitting the break on upgrade needs, and it gives the planned 56/57→58 codemod something to point at.

apps/router-e2e/__e2e__/native-navigation/app/js-tabs/index.tsx
jakub-agent, line 11: [🐧 nit] `e2e-goto-profile-details` is used both here (the Home tab) and in `profile/index.tsx:11` (the Profile tab). Tabs stay mounted once visited, so after a user has been to both tabs the ID matches two elements at the same time.

A prefix per screen — `e2e-home-goto-profile-details` and `e2e-profile-goto-details` — keeps the IDs unambiguous no matter what is mounted.

jakub-agent (commented):

Iteration 4, against `41615e6c` (force-pushed over `6436ae45`).

**The build break is fixed.** `TabsClient.tsx:27` keeps `export * from '../react-navigation/bottom-tabs'`, so the six re-exports in `layouts/Tabs.tsx` resolve again, and the two orphaned `*ReactNavigationClient.tsx` files are gone. Restoring the barrel in place was the smaller fix of the two I suggested — fine by me.

Better still, `__rsc_tests__/Tabs.test.tsx` was updated in the same push and now asserts all six exports plus the three transition namespaces resolve through the server entry. That test is what would have caught the last revision, so the coverage is in the right place.

**Withdrawing one thing from my last review:** I noted that `Tabs.Screen = Screen` and `Tabs.Protected = Protected` in `layouts/Tabs.tsx` were dead because `withLayoutContext` already assigns them. That was too quick. `Tabs` is imported from a `'use client'` module, so in the RSC graph those properties are reached through a client reference rather than the real object — and this push's `renders Tabs.Screen` / `renders Tabs.Protected` flight-snapshot tests exercise exactly that path. `layouts/TopTabs.tsx` carries the same two assignments for the same reason. Treat them as intentional.

**Still open** — export surface and docs only, nothing blocking:
- `BottomTabEmitter` and `BottomTabViewState` remain unreachable from any public subpath.
- The changelog entry still reads as an internal refactor.
- One residual `testID` reuse across two e2e screens.

Standing corrections from earlier iterations: the `sdk-56` codemod is versioned to the 55→56 upgrade, so its output is correct for that SDK and my first-iteration finding on it does not apply; and `BottomTabNavigator` was never a public export.

jakub-agent (approved):

Approving. This head is a pure rebase of `41615e6c` — the tree is byte-identical — so the iteration 4 review still applies.

Across four revisions the migration itself has held up. `navigateToTab` via `actions.navigate(name, params)` matches `TabRouter`'s name-based `NAVIGATE` handling, which never matches on `payload.key` and preserves `route.path` when `payload.path` is absent, so the switch away from `CommonActions.navigate(route)` is behaviour-preserving. `createProps` and `processScreens` follow the `Drawer` precedent from #47839. Test coverage ended up stronger than it started: the suite now drives the public `Tabs` layout, and the added cases cover `href: null`, custom `tabBar` with `tabPress` `preventDefault`, tab re-selection without remounting, and — after the export break in an earlier revision — the RSC entry's export surface.

Remaining items were reviewed and closed out as either not applicable or optional:
- The `sdk-56` codemod finding from iteration 1 was mine to withdraw: that transform is versioned to the 55→56 upgrade, so its output is correct for its target SDK.
- `BottomTabNavigator` was never a public export, so it does not belong in the removals list.
- The `Tabs.Screen`/`Tabs.Protected` assignments are load-bearing across the client-reference boundary, not dead code.
- The unreachable `BottomTabEmitter`/`BottomTabViewState` exports, the changelog wording, and the e2e `testID` reuse were reviewed and dismissed by the author.

One thing to note for the record: `check-packages` and the `jest-*` jobs were still pending when this was submitted, so every review in this series was based on reading the tree rather than a green run. Worth confirming those before merge.

PR #48294, opened 2026-07-29
[location][Android] Fix leaking watches

Wenszel:
# Why
Fixes an issue where resources weren't released. LocationModule used to implement `LifecycleEventListener` but never subscribed to its events, which caused resources to leak on module destroy or when switching between background and foreground.

# How

Used the `OnDestroy`, `OnActivityEntersForeground`, and `OnActivityEntersBackground` APIs to release resources, and removed the `LifecycleEventListener` implementation from the module.

# Test plan
NCL -> Location screen, with `subscription.remove()` disabled. Without this fix, the receiver persists after a reload, which can be observed with this command:

```
while true; do echo "$(date +%H:%M:%S) receivers: $(adb shell dumpsys activity broadcasts | sed -n '/Registered Receivers:/,/Receiver Resolver Table:/p' | grep -c MOTION_ACTIVITY_UPDATE)"; sleep 2; done
```

### DIFF
--- a/packages/expo-location/android/src/main/java/expo/modules/location/LocationModule.kt
+++ b/packages/expo-location/android/src/main/java/expo/modules/location/LocationModule.kt
@@ -38,7 +38,6 @@ import com.google.android.gms.location.LocationServices
 import com.google.android.gms.location.LocationSettingsRequest
 import com.google.android.gms.location.Priority
 import expo.modules.core.interfaces.ActivityEventListener
-import expo.modules.core.interfaces.LifecycleEventListener
 import expo.modules.core.interfaces.services.UIManager
 import expo.modules.interfaces.taskManager.TaskManagerInterface
 import expo.modules.kotlin.Promise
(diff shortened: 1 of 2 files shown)


Wenszel:
* **#48294** <a href="https://app.graphite.com/github/pr/expo/expo/48294?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/48294?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/expo-location/android/src/main/java/expo/modules/location/LocationModule.kt
```diff
@@ -377,10 +376,16 @@ class LocationModule : Module(), LifecycleEventListener, SensorEventListener, Ac
 
     OnActivityEntersForeground {
       AppForegroundedSingleton.isForegrounded = true
+      onHostResume()
```
HubertBer: I'm not sure if this won't resume things that weren't even started. For example `onHostResume()` will call `startHeadingUpdate()` regardless of whether the `watchDeviceHeading` was called, right?

HubertBer (changes requested):

packages/expo-location/android/src/main/java/expo/modules/location/LocationModule.kt
Wenszel: Yep, good catch. The problem, though, is that some functions in `onHostResume` are missing guards to check whether their features have been initialized - `resumeMotionActivityWatch` already has one. I’ll fix it.

lukmccall (approved):

HubertBer (approved):

PR #48301, opened 2026-07-29
[ios][router] Resolve NativeTabs `xcasset` icon names through the asset catalog

### LINKED ISSUE #47480: [expo-router] NativeTabs \xcasset` icon renders image sets but not symbol sets on iOS`

CavalcanteLeo:
# Why

Fixes #47480.

Split out of #48209 at @Ubax's request — that PR fixed two separate issues. This one is the `xcasset` fix; the icon type mismatch crash is in #48302.

`NativeTabs.Trigger.Icon`'s `xcasset` prop is documented to load images from the Xcode asset catalog via `[UIImage imageNamed:]`, but expo-router never emitted the icon shape that reaches that native path. `convertOptionsIconToScreensPropsIcon` turned the asset name into an image source `uri`:

```ts
const imageSource = 'xcasset' in icon && icon.xcasset ? { uri: icon.xcasset } : icon.src;
```

`{ uri: name }` is resolved by `RCTImageLoader`, which can find a raster **image set** by name but can never produce a **symbol set**. So image sets happened to work and symbol sets rendered blank with `WARN Could not find image file://.../<name>.png`.

# How

Give `xcasset` its own branch in `optionsIconConverter.ios.ts` that mirrors the existing `sf` branch and returns `{ type: 'xcasset', name }`. react-native-screens maps that to `RNSTabsIconTypeXcasset` → `[UIImage imageNamed:]`, which handles image sets and symbol sets alike. This is also what expo-router already does for toolbar icons in `stack-utils/toolbar/shared.ts`.

The `src` branch is now separate, so its `ImageSourcePropType` cast is gone.

# Behavior change

`xcasset` icons no longer honor the implicit `iconColor` → `template` rule; their rendering mode now comes from the asset's **Render As** setting in the asset catalog, which is the canonical place to configure it for asset-catalog images (set it to **Template Image** to have the icon tinted). This affects raster image sets set to *Render As: Default* that were relying on the implicit tinting. Symbol sets and image sets set to *Template Image* tint as before.

Note the `renderingMode` prop was already a no-op for `xcasset` on native tabs — `appendIconOptions` never forwarded it — so the docs claiming otherwise were wrong. Both the TSDoc and the guide are corrected here.

# Test Plan

Added, red before the change and green after (5 failures on `main`, 0 after):

- `native-tabs/utils/__tests__/optionsIconConverter.test.ios.tsx` — a new `xcasset` block: the converter returns `{ type: 'xcasset', name }`, keeps that type when an icon color is set, falls through on an empty name, is still overridden by `sf`, and produces a matching type for the normal and selected states. The `xcasset` branch previously had no test coverage at all.
- `native-tabs/__tests__/options.e2e.test.ios.tsx` — integration coverage that `<Icon xcasset={{ default, selected }} />` reaches `Tabs.Screen` as two `xcasset` icons, and that an icon color doesn't change that.

```
🏁 All checks passed.
```

The generated API reference data was updated by hand for just this TSDoc change: running `et generate-docs-api-data --packageName expo-router` also picks up unrelated pending `standard-navigation` changes that haven't been regenerated on `main` yet, which don't belong in this PR.

Not verified on a device — I don't have the reporter's asset catalog. The repro project in #47480 covers it, and the reporter already confirmed this exact conversion works locally.

# Checklist

- [x] Documentation updated (guide + TSDoc + generated API data)
- [ ] Verified on a physical device / simulator (see above)


AhmedAlAmawi:
+1 on this — we hit the same thing and have been shipping this exact change as a local patch in production for a while now (carried it across expo-router 55.0.3 and 56.2.11, iOS app with native tabs using xcasset icons). Resolving through the asset catalog instead of a `{ uri }` source is what made symbol sets work for us. No issues since. Would love to see this land so we can drop the patch.

Ubax:
Sorry for the delay. Since this involves an API change, we need to give it a bit more thought. I’ll get back to you after we’ve discussed it internally.

CavalcanteLeo:
@Ubax any plan for this? this keep crashing the app, i had to patch it here

Ubax:
Thanks for patience and the contribution. We'll go forward with the approach you presented here. 

Can you please:
- rebase the PR
- add a warning for when someone sets both `xcasset` and `renderingMode`

CavalcanteLeo:
Rebased onto `main` and resolved conflicts (test file additions, changelog, docs, and the `optionsIconConverter.ios.ts` refactor from #48302 all merged cleanly).

Also added the requested warning: setting `renderingMode` alongside `xcasset` now logs a dev-only `console.warn` explaining that tinting for asset catalog icons comes from the asset's "Render As" setting instead, since `renderingMode` has no effect there. Covered by a new test and updated the docs/TSDoc accordingly.

packages/expo-router/CHANGELOG.md
```diff
@@ -85,6 +85,7 @@
 - Prevent `useLoaderData()` from re-rendering readers of unrelated loader paths ([#48523](https://github.com/expo/expo/pull/48523) by [@hassankhan](https://github.com/hassankhan))
 - Fix package export for `expo-router/unstable-split-view` ([#49001](https://github.com/expo/expo/pull/49001) by [@hassankhan](https://github.com/hassankhan))
 - [ios][native-tabs] Fix a `[RNScreens] icon and selectedIcon must be same type.` crash when the normal and selected icons resolved to different rendering modes. ([#48302](https://github.com/expo/expo/pull/48302) by [@CavalcanteLeo](https://github.com/CavalcanteLeo))
+- [ios][native-tabs] Resolve `Icon` `xcasset` names through the asset catalog so symbol sets render. Their rendering mode now comes from the asset's "Render As" setting instead of `iconColor`, and setting `renderingMode` alongside `xcasset` now logs a warning since it has no effect. ([#48301](https://github.com/expo/expo/pull/48301) by [@CavalcanteLeo](https://github.com/CavalcanteLeo))
```
Ubax, line 88:
```suggestion
- [ios][native-tabs] Resolve `Icon` `xcasset` names through the asset catalog so symbol sets render.  ([#48301](https://github.com/expo/expo/pull/48301) by [@CavalcanteLeo](https://github.com/CavalcanteLeo))
```

Ubax:
@expo-bot verify

Ubax:
@expo-bot review

packages/expo-router/src/native-tabs/common/elements.tsx
```diff
@@ -97,11 +97,12 @@ export interface XcassetIcon {
...
    *
    * Xcassets provide automatic multi-resolution (@1x/@2x/@3x), dark mode variants,
-   * and device-specific images via `[UIImage imageNamed:]`.
+   * and device-specific images via `[UIImage imageNamed:]`. Both image sets and
+   * symbol sets are supported.
    *
-   * The rendering mode (template vs original) can be controlled via the `renderingMode` prop
-   * on the `Icon` component. By default, icons are tinted when `iconColor` is set, and
-   * rendered as original otherwise.
+   * > **Note:** The rendering mode comes from the asset's **Render As** setting in the
```
amandeepmittal, line 103: ⚠️ **correctness** `mdx-components-sdk.md#apisection`

No regenerated API data is included in this PR, so `public/static/data/unversioned/expo-router/native-tabs.json` still describes `xcasset` with the old "can be controlled via the `renderingMode` prop" wording. The reference page will contradict the guide page this PR corrects. Regenerate with `et gdad --packageName expo-router` and commit the `native-tabs.json` change.

docs/pages/router/advanced/native-tabs.mdx
```diff
@@ -378,7 +378,7 @@ export default function TabLayout() {
 
 > **info** This feature is available in SDK 55 and later.
 
-On iOS, you can use images from the Xcode asset catalog as tab icons with the `xcasset` prop. This is useful when you want to manage your icons through Xcode's asset catalog instead of bundling image files.
+On iOS, you can use images from the Xcode asset catalog as tab icons with the `xcasset` prop. This is useful when you want to manage your icons through Xcode's asset catalog instead of bundling image files. Both image sets and symbol sets are supported.
```
amandeepmittal: ⚠️ **correctness** `review-ai-text.md#t7-overclaiming`

"Both image sets and symbol sets are supported" describes the fix in this PR, not what a reader on the current SDK gets: on SDK 57 a symbol set still renders blank. Gate the new part to the SDK it ships in, the way the rest of this page gates `renderingMode` and asset catalog icons.

```suggestion
On iOS, you can use images from the Xcode asset catalog as tab icons with the `xcasset` prop. This is useful when you want to manage your icons through Xcode's asset catalog instead of bundling image files. Image sets are supported in SDK 55 and later, and symbol sets in SDK 58 and later.
```

docs/pages/router/advanced/native-tabs.mdx
```diff
@@ -417,7 +417,7 @@ export default function TabLayout() {
 }
 ```
 
-> **info** Asset catalog icons support the `renderingMode` prop, just like `src` icons. When `iconColor` is set, icons default to `template` rendering. Otherwise, they default to `original`.
+> **info** Asset catalog icons don't support the `renderingMode` prop. Their rendering mode comes from the asset's **Render As** setting in the asset catalog. Set it to **Template Image** for the icon to be tinted with `iconColor` and `selectedIconColor`. Setting `renderingMode` alongside `xcasset` has no effect and logs a warning.
```
amandeepmittal: ⚠️ **correctness** `review-ai-text.md#t7-overclaiming`

The **Render As** behavior and the warning both arrive with this PR, so on SDK 57 this callout is wrong: `renderingMode` is dropped, but the icon still falls back to `template` rendering when `iconColor` is set. Gate the new behavior to the SDK it ships in. The last sentence also repeats the first, so drop it.

```suggestion
> **info** Asset catalog icons ignore the `renderingMode` prop. In SDK 58 and later, their rendering mode comes from the asset's **Render As** setting in the asset catalog. Set it to **Template Image** for the icon to be tinted with `iconColor` and `selectedIconColor`.
```

CavalcanteLeo:
any chances this be delivered on expo 58?

PR #48302, opened 2026-07-29
[ios][router] Fix NativeTabs `icon` and `selectedIcon` type mismatch crash

CavalcanteLeo:
# Why

Split out of #48209 at @Ubax's request — that PR fixed two separate issues. This one is the icon type mismatch crash; the `xcasset` asset-catalog fix is in #48301.

`NativeTabsView.ios.tsx` converted `icon` with the *normal* icon color and `selectedIcon` with the *selected* icon color:

```tsx
const iosIcon = convertOptionsIconToScreensPropsIcon(
  shared.icon,
  standardAppearance?.stacked?.normal?.tabBarItemIconColor
);
const iosSelectedIcon = convertOptionsIconToScreensPropsIcon(
  shared.selectedIcon ?? shared.icon,
  standardAppearance?.stacked?.selected?.tabBarItemIconColor
);
```

That color is what decides between `imageSource` and `templateSource`, so setting only one of `iconColor` / `selectedIconColor` made the two icons resolve to different types, and react-native-screens throws on that:

```
Error: [RNScreens] icon and selectedIcon must be same type.
```

That is a hard crash (red screen) for a plausible config — for example `<NativeTabs iconColor={{ selected: 'red' }}>` with any `src` icon.

# How

Derive the icon color from both states (`normal ?? selected`) and pass the same value to both conversions. The color is only used as a "should this image be a template" signal, so merging is the right granularity — the rendering mode is a property of the tab bar item, not of one state. This only changes behavior in the case that previously threw.

# Test Plan

Added to `native-tabs/__tests__/options.e2e.test.ios.tsx`: a table asserting `icon.type === selectedIcon.type` across all four `iconColor` / `selectedIconColor` combinations, verified against the real `Tabs.Screen`, not a mock.

Red before the change, green after:

```
# on main
✕ icon and selectedIcon have the same type when only selectedIconColor is set
✓ icon and selectedIcon have the same type when only iconColor is set
✓ icon and selectedIcon have the same type when both icon colors are set
✓ icon and selectedIcon have the same type when no icon color is set
Tests:       1 failed, 63 passed, 64 total

# with the fix
Tests:       64 passed, 64 total
```

```
🏁 All checks passed.
```

# Checklist

- [x] Documentation — n/a, no public API change


packages/expo-router/src/native-tabs/NativeTabsView.ios.tsx
```diff
@@ -108,13 +108,18 @@ function Screen(props: InternalTabScreenProps) {
...
-  );
+  // React Native Screens requires `icon` and `selectedIcon` to resolve to the same icon
+  // type and throws otherwise. The icon color only decides whether an image icon is
+  // rendered as a template, so derive it from both states — using the per-state color
+  // would make the two icons disagree whenever only one of the states sets a color.
+  const iconColor =
+    standardAppearance?.stacked?.normal?.tabBarItemIconColor ??
+    standardAppearance?.stacked?.selected?.tabBarItemIconColor;
+
+  const iosIcon = convertOptionsIconToScreensPropsIcon(shared.icon, iconColor);
```
Ubax: Can you take a little different approach here - instead of deciding based on color, it would be better to detect the `effectiveRenderingMode` of icon in both states - selected and normal - and then:
- if they are the same - nothing happens
- if they are different - use the renderingMode of the normal icon for both and show `console.warn` saying that `NativeTabs does not currently support rendering icons in different modes. This issue may occur if you specify a color only for the standard or selected icon, or if you explicitly set renderingMode.`

Ubax (changes requested):
Thanks for splitting the PR! I'd suggest taking a bit different approach here

packages/expo-router/src/native-tabs/NativeTabsView.ios.tsx
CavalcanteLeo: sure, let me change it

CavalcanteLeo: Should I add a gated warning behind process.env.NODE_ENV !== 'production', matching the other dev warning in the same file?

CavalcanteLeo: Done in 8d96c71 — switched to detecting the effective rendering mode per state instead of deriving a single color.

One thing worth flagging: only selectedIconColor on its own actually produces a mismatch. iconColor is applied to the normal, focused and selected states in appearance.ios.ts, so setting just iconColor gives both icons template and warns nothing. I'd initially written the test expecting a warning there and it failed, so the test now asserts the real behavior.

Covered by unit tests for resolveIconRenderingMode and a table test in options.e2e.test.ios.tsx asserting the resolved type and warning for each color combination. Also added a note under "Icon rendering mode" in the native tabs guide. Merged main in and resolved the CHANGELOG conflict; et check-packages expo-router is green.

docs/pages/router/advanced/native-tabs.mdx
```diff
@@ -374,6 +374,8 @@ export default function TabLayout() {
 
 > **info** The `renderingMode` prop only affects iOS. On Android, all image icons are rendered with their original colors.
 
+> **info** On iOS, a tab's default and selected icons must use the same rendering mode. When they resolve to different modes, for example when only `selectedIconColor` is set, both icons use the default icon's rendering mode and a warning is logged.
```
Ubax: Can you please move this information to Known Limitations section?

packages/expo-router/src/native-tabs/NativeTabsView.ios.tsx
```diff
@@ -108,15 +111,39 @@ function Screen(props: InternalTabScreenProps) {
...
+  if (
+    process.env.NODE_ENV !== 'production' &&
+    normalRenderingMode &&
+    selectedRenderingMode &&
+    normalRenderingMode !== selectedRenderingMode
+  ) {
+    console.warn(
+      'NativeTabs does not currently support rendering icons in different modes. ' +
+        'This issue may occur if you specify a color only for the standard or selected icon, ' +
+        'or if you explicitly set renderingMode.'
```
Ubax: Please add `props.name` to the warning as well

packages/expo-router/src/native-tabs/NativeTabsView.ios.tsx
```diff
@@ -108,15 +111,39 @@ function Screen(props: InternalTabScreenProps) {
...
+    );
+  }
+
+  // The normal state wins, so the selected icon follows the mode the tab bar shows most of the time.
+  const renderingMode = normalRenderingMode ?? selectedRenderingMode;
```
Ubax:
```suggestion
  const effectiveRenderingMode = normalRenderingMode ?? selectedRenderingMode;
```

packages/expo-router/src/native-tabs/utils/optionsIconConverter.ts
```diff
@@ -18,9 +18,18 @@ export function appendIconOptions(options: NativeTabOptions, props: NativeTabsTr
   applySelectedColor(options, props.selectedColor);
 }
 
+export type IconRenderingMode = 'template' | 'original';
```
Ubax: Can you add this type in `packages/expo-router/src/native-tabs/types.ts` instead and replace all the `'template' | 'original'` occurences?

packages/expo-router/src/native-tabs/utils/optionsIconConverter.ios.ts
```diff
@@ -43,28 +43,63 @@ export function appendIconOptions(options: NativeTabOptions, props: NativeTabsTr
...
+ * the image's own colors. It does not apply to SF Symbols, which the system always tints.
+ */
+export type IconRenderingMode = 'template' | 'original';
+
+/**
+ * Resolves the rendering mode an icon will be displayed with, or `undefined` when the icon isn't
+ * image-based. An explicit `renderingMode` always wins; otherwise an icon color implies
+ * `template`, because a color can only take effect on a template image.
+ */
+export function resolveIconRenderingMode(
```
Ubax, line 51:
```suggestion
export function resolveIconRenderingMode(
```

packages/expo-router/src/native-tabs/utils/optionsIconConverter.ios.ts
```diff
@@ -43,28 +43,63 @@ export function appendIconOptions(options: NativeTabOptions, props: NativeTabsTr
...
+    return { type: 'imageSource', imageSource };
+  }
+  return undefined;
+}
+
+/**
+ * Returns the image an icon renders, mirroring the precedence in
+ * {@link convertOptionsIconToScreensPropsIcon} — SF Symbols aren't image-based, so they resolve
+ * to `undefined`.
+ */
```
Ubax:
```suggestion
```

packages/expo-router/src/native-tabs/utils/optionsIconConverter.ios.ts
```diff
@@ -43,28 +43,63 @@ export function appendIconOptions(options: NativeTabOptions, props: NativeTabsTr
...
-    if (effectiveRenderingMode === 'original') {
-      return { type: 'imageSource', imageSource };
+  const imageSource = getIconImageSource(icon);
+  if (imageSource) {
+    const effectiveRenderingMode = renderingMode ?? resolveIconRenderingMode(icon);
+    if (effectiveRenderingMode === 'template') {
+      return { type: 'templateSource', templateSource: imageSource };
     }
-    return { type: 'templateSource', templateSource: imageSource };
+    return { type: 'imageSource', imageSource };
```
Ubax: To minimize the diff, please revert the condition
```suggestion
    if (effectiveRenderingMode === 'original') {
      return { type: 'imageSource', imageSource };
    }
    return { type: 'templateSource', templateSource: imageSource };
```

packages/expo-router/src/native-tabs/utils/optionsIconConverter.ios.ts
```diff
@@ -43,28 +43,63 @@ export function appendIconOptions(options: NativeTabOptions, props: NativeTabsTr
...
-      'xcasset' in icon && icon.xcasset
-        ? { uri: icon.xcasset }
-        : (icon as { src: ImageSourcePropType }).src;
-    const renderingMode = 'renderingMode' in icon ? icon.renderingMode : undefined;
-    const effectiveRenderingMode =
-      renderingMode ?? (iconColor !== undefined ? 'template' : 'original');
-    if (effectiveRenderingMode === 'original') {
-      return { type: 'imageSource', imageSource };
+  const imageSource = getIconImageSource(icon);
+  if (imageSource) {
```
Ubax:
```suggestion
  if (!imageSource) {
    return undefined
  }
```

packages/expo-router/CHANGELOG.md
```diff
@@ -38,6 +38,7 @@
 - [android][ios] Fix `expo-router/head` and `expo-router/stack` resolution on native platforms. ([#47870](https://github.com/expo/expo/pull/47870) by [@hassankhan](https://github.com/hassankhan))
 - Fix missing subpath warning from Metro when importing from `expo-router/server` ([#48045](https://github.com/expo/expo/pull/48045) by [@hassankhan](https://github.com/hassankhan))
 - Fix `replace` navigation in tabs leaving the replaced route in history. ([#48256](https://github.com/expo/expo/pull/48256) by [@Ubax](https://github.com/Ubax))
+- [ios][native-tabs] Fix a `[RNScreens] icon and selectedIcon must be same type.` crash when the normal and selected icons resolved to different rendering modes, for example when only `selectedIconColor` was set. Both icons now use the normal state's rendering mode and a warning is logged. ([#48302](https://github.com/expo/expo/pull/48302) by [@CavalcanteLeo](https://github.com/CavalcanteLeo))
```
```suggestion
- [ios][native-tabs] Fix a `[RNScreens] icon and selectedIcon must be same type.` crash when the normal and selected icons resolved to different rendering modes. ([#48302](https://github.com/expo/expo/pull/48302) by [@CavalcanteLeo](https://github.com/CavalcanteLeo))
```

tools/src/publish-packages/tasks/bundleIOSPrebuilds.ts
Ubax, line 1: Why was this changed?

Ubax (commented):
The change looks good!

CavalcanteLeo:
Thanks for the review! All comments addressed in dfe5ab4de63. `et check-packages expo-router` passes (37/37: build, typecheck, tests, lint, format, depscheck).

| Comment | Change |
|---|---|
| Move the note to Known limitations | Inline `> **info**` removed; added a `<Collapsible summary="The default and selected icons share one rendering mode on iOS">` entry under `## Known limitations` |
| Add `props.name` to the warning | The warning now names the tab, with a new assertion in `options.e2e.test.ios.tsx` |
| `effectiveRenderingMode` | Renamed |
| `IconRenderingMode` in `types.ts` | Declared there and used for every `'template' \| 'original'` in `native-tabs` — see the note below |
| Drop the `getIconImageSource` JSDoc | Removed |
| Revert the condition | Now `if (effectiveRenderingMode === 'original')` falling through to `templateSource` |
| Early-return on `!imageSource` | Done |
| Shorten the changelog | Applied verbatim |

One suggestion I only half-applied: on the `resolveIconRenderingMode` block I dropped the `IconRenderingMode` doc comment but kept the function's own "Resolves the rendering mode…" paragraph, since it documents the precedence rule (explicit `renderingMode` wins, otherwise a color implies `template`). Happy to remove it too if that's what you meant — the suggestion diff covered both.

### `bundleIOSPrebuilds.ts`

Not my change. The file is byte-identical to `main` (`git diff origin/main -- tools/src/publish-packages/tasks/bundleIOSPrebuilds.ts` is empty) — the import reorder came from #48424, and it only shows in the diff because the merge-base predates that commit. Nothing to revert.

### `IconRenderingMode` is now a public export

`renderingMode` lives on the publicly documented `SrcIcon` and `SymbolOrImageSource`, so replacing the inline union made typedoc emit a `reference` to `IconRenderingMode`, and [`renderWithLink`](https://github.com/expo/expo/blob/main/docs/components/plugins/api/APISectionUtils.tsx#L141) turns that into an `#iconrenderingmode` anchor. Unexported it would dead-link, so I added it to `native-tabs/index.ts` and regenerated `native-tabs.json`. I reverted the five other regenerated data files — their diffs were unrelated drift from other merged PRs — but `native-tabs.json` still carries some union-ordering churn that regeneration can't avoid.

I also kept the replacement scoped to `src/native-tabs/`. There are ~12 more `'template' | 'original'` sites under `src/layouts/stack-utils/toolbar/`, `src/toolbar/`, and `src/link/preview/` (as `iconRenderingMode` / `imageRenderingMode`); pointing those at a native-tabs type would couple unrelated features, so I left them. Glad to unify them in a follow-up if you'd prefer one shared type.

### On the bot's warning

It was right that the note named an internal field. The limitation entry and the `console.warn` now name the actual public props — `tintColor`, `iconColor={{ selected }}`, and the `Icon` `selectedColor` prop. `tintColor` does reach the selected icon color via `selectedIconColor: processedIconColor.selected ?? tintColor` in `NativeBottomTabsNavigator.tsx`, so the `tintColor` example on that page really did lose its selected-icon tint; both SDK tabs of it now pass `renderingMode="template"`.

### Unrelated gap, left alone

`<Icon xcasset="…" renderingMode="original" />` type-checks (via `XcassetIcon & SrcIcon`), but the iOS `appendIconOptions` `xcasset` branch never copies `renderingMode` into `options.icon`, so it's silently dropped — which makes the claim in `native-tabs.mdx` that "Asset catalog icons support the `renderingMode` prop, just like `src` icons" untrue today. Pre-existing and out of scope here; I can open a separate issue/PR for it.

brentvatne:
@expo-bot verify

CavalcanteLeo:
rebased

docs/pages/router/advanced/native-tabs.mdx
```diff
@@ -257,6 +257,8 @@ export default function TabLayout() {
             default: require('../assets/setting_icon.png'),
             selected: require('../assets/selected_setting_icon.png'),
           }}
+          // So `tintColor` also applies to the selected image icon on iOS
+          renderingMode="template"
```
Ubax:
```suggestion
```

docs/pages/router/advanced/native-tabs.mdx
```diff
@@ -296,6 +298,8 @@ export default function TabLayout() {
...
```
Ubax:
```suggestion
```

Ubax (commented):
Sorry for the delay!

Please update the docs, and check one comment from the previous review

Ubax:
@expo-bot review

docs/pages/router/advanced/native-tabs.mdx
CavalcanteLeo: rn

CavalcanteLeo: give me 20 min, will push soon

docs/pages/router/advanced/native-tabs.mdx
CavalcanteLeo: Done in 4c1cc8b — dropped from both `tintColor` examples. The shared-rendering-mode caveat is still covered in the Known limitations section.

docs/pages/router/advanced/native-tabs.mdx
CavalcanteLeo: Done in 4c1cc8b.

packages/expo-router/src/native-tabs/utils/optionsIconConverter.ios.ts
CavalcanteLeo: This is the one I missed — I read it as just the signature and did not notice the suggestion range covered the JSDoc above it. Removed in 4c1cc8b; `IconRenderingMode` in `types.ts` keeps its doc comment since that one shows up in the API docs.

Also merged `main` in and resolved the `CHANGELOG.md` conflict. `et check-packages expo-router` is green.

Ubax:
@expo-bot review

CavalcanteLeo:
@Ubax any news?

apps/observe-tester/app/(tabs)/examples/expo-image/too-big.tsx
Ubax, line 1: Can you please remove this change?

Ubax (commented):
Thank you for the contribution 🎉 
I'll do the final pass on Monday and then merge it!

Ubax:
@expo-bot review

Ubax (approved):
Code looks good to me 👍 

Let's wait for @amandeepmittal to review the doc change

docs/pages/router/advanced/native-tabs.mdx
```diff
@@ -1561,6 +1561,38 @@ export default function TabLayout() {
 
 ## Known limitations
 
+<Collapsible summary="The default and selected icons share one rendering mode on iOS">
+
+On iOS, a tab's default and selected image icons must use the same [rendering mode](#icon-rendering-mode). When they resolve to different modes, both icons use the default icon's mode and a warning is logged.
+
+The modes disagree when an icon color applies to only one of the states. This happens when you set `tintColor`, `iconColor={{ selected }}`, or the `Icon` `selectedColor` prop without also setting a color for the default state, because a color implies `'template'` rendering while an uncolored icon defaults to `'original'`. Setting `renderingMode` for only one state has the same effect.
```
amandeepmittal: ✏️ **style** `mechanics.md#vale-sentencelength`

The sentence starting "This happens when you set" runs past 25 words. Split it after "the default state" so the explanation stands as its own sentence.

```suggestion
The modes disagree when an icon color applies to only one of the states. This happens when you set `tintColor`, `iconColor={{ selected }}`, or the `Icon` `selectedColor` prop without also setting a color for the default state. A color implies `'template'` rendering, while an uncolored icon defaults to `'original'`. Setting `renderingMode` for only one state has the same effect.
```

docs/pages/router/advanced/native-tabs.mdx
```diff
@@ -1561,6 +1561,38 @@ export default function TabLayout() {
 
 ## Known limitations
 
+<Collapsible summary="The default and selected icons share one rendering mode on iOS">
+
+On iOS, a tab's default and selected image icons must use the same [rendering mode](#icon-rendering-mode). When they resolve to different modes, both icons use the default icon's mode and a warning is logged.
```
amandeepmittal: ✏️ **style** `prose.md#voice-and-tone`

"a warning is logged" is passive and leaves out who logs it and when. Name the actor: "Expo Router logs a warning in development".

```suggestion
On iOS, a tab's default and selected image icons must use the same [rendering mode](#icon-rendering-mode). When they resolve to different modes, both icons use the default icon's mode and Expo Router logs a warning in development.
```

amandeepmittal (approved):
Looks good to me overall! Just two minor prose suggestions reported by the docs reviewer. Going to commit those changes before merging this PR.

Thank you for your contribution!

PR #46343, opened 2026-05-27
[iOS][expo-widgets] Add staleDate parameter to LiveActivity start/update

KyleAsaff:
Closes #46341

Builds on #46340 (now merged), which makes `expo-modules-core` accept JS `Double` timestamps on `Date` arguments.

## Summary

Adds optional `staleDate` to `LiveActivityFactory.start()` and `LiveActivity.update()`. When set, the value is forwarded to `ActivityContent(state:staleDate:)` instead of hardcoding `nil`.

Omitting the argument keeps today's behavior.

## Motivation

We run a ride-tracking app with a lock-screen Live Activity. Updates come from a background location task on a ~10s cadence. If GPS or the background task stalls, the widget can sit there showing stale speed/distance with no visual cue.

ActivityKit's `staleDate` is the documented way to tell the system "this content may be out of date after X." We have been carrying a local `pnpm patch` because the upstream API did not expose it.

## API

```ts
factory.start(props, url?, staleDate?: Date)
activity.update(props, staleDate?: Date)
```

JS passes `staleDate?.getTime()` across the bridge. The `Function("start")` and `AsyncFunction("update")` closures take `Date?` directly, and `expo-modules-core` converts the JS timestamp through its `Date` Convertible (the `Double` branch added in #46340).

## Test plan

- [ ] `pnpm build` in `packages/expo-widgets`
- [ ] Manual: start a Live Activity with `staleDate: new Date(Date.now() + 30_000)`, confirm the lock screen dims after the window
- [ ] Existing callers that omit `staleDate` are unchanged

## Update

Addressed jakex7's review: dropped the temporary module-boundary `Double` conversion now that #46340 has merged, so the closures take `Date?` directly. Also resolved conflicts against the latest `main` (the layout registry and `initialProps` work from #46501 and #46527) and regenerated the build output.

PR #48303, opened 2026-07-29
[expo-widgets] Expose ActivityKit's `isStale` on `LiveActivityEnvironment`

dirkpostma:
# Why

`staleDate` landed in #46343, so a Live Activity can now tell the system when its
content should be considered out of date. But the layout has no way of finding out that
it happened. `LiveActivityEnvironment` doesn't carry `isStale`, so a layout can't react
to the date it was given.

The two only really work as a pair. `staleDate` exists so the system can de-emphasize
content while the app isn't running, and in SwiftUI you consume it with
`context.isStale`. Today the only way to show staleness through `expo-widgets` is to
push a prop from JS, which fails in exactly the situation `staleDate` is meant for.

My use case is a BLE heart-rate monitor. When the sensor drops out and iOS has suspended
or terminated the app, the Live Activity has to stop claiming a live BPM. Nothing of
mine is running at that point, so the decision has to be the system's.

# How

`getLiveActivityEnvironment()` builds the environment from SwiftUI `EnvironmentValues`,
and `isStale` isn't there. It lives on ActivityKit's `ActivityViewContext`, which
`WidgetLiveActivity.body` already has in scope at both `getLiveActivityNodes` call
sites. So the environment helper becomes context-aware:

```swift
func environment(for context: ActivityViewContext<LiveActivityAttributes>) -> [String: Any] {
  var env: [String: Any] = getLiveActivityEnvironment(environment: self.env)
  if #available(iOS 16.2, *) {
    env["isStale"] = context.isStale
  }
  return env
}
```

This follows the pattern already used for widgets in `EntryView.swift`, which enriches
`getWidgetEnvironment()` with a `timestamp` from the timeline entry. Same shape: a key
that can't come from `EnvironmentValues`.

`ActivityViewContext.isStale` is iOS 16.2+ while `WidgetLiveActivity` is gated at 16.1,
hence the inner availability check. Without it the build fails with `'isStale' is only
available in iOS 16.2 or newer`.

`getLiveActivityNodes` doesn't cache, so the layout re-evaluates on the body rebuild
that the stale transition triggers, and the new value reaches JS.

# Test Plan

- `LiveActivityEnvironment.isStale` is typed and documented, so it shows up on the
  Widgets API docs page.
- I checked that the availability nesting compiles against the iOS simulator SDK at
  `-target arm64-apple-ios16.1-simulator`, and that dropping the `#available` guard
  fails to build. The guard is load-bearing rather than defensive.

Usage:

```tsx
const HeartRateActivity = (props: Props, env: LiveActivityEnvironment) => {
  'widget';
  return {
    banner: (
      <Text modifiers={[opacity(env.isStale ? 0.4 : 1)]}>
        {env.isStale ? 'no signal' : `${props.bpm} bpm`}
      </Text>
    ),
  };
};
```

# Checklist


dirkpostma:
Following up on the Test Plan. I've now verified this at runtime, not just at compile time.

**Setup.** A throwaway SDK 57 app (expo `^57.0.0`, RN 0.86.0, React 19.2.3) with
`expo-widgets@57.0.6` patched to add this change, plus a backport of #46343 so
`staleDate` is plumbed on 57.x at all. The layout renders
`` `stale=${String(env.isStale)}` `` into `compactTrailing`, so what's on screen is the
raw environment value rather than a proxy for it. iPhone 17 Pro simulator, iOS 26.5.

**Result: `isStale` reaches the layout and it re-renders, with the app suspended.**

| time | event |
|---|---|
| 23:37:41 | app requests activity with `staleDate` = +45s (23:38:26) |
| 23:38:12 | `runningboardd`: `app<dev.harness.stale>` → `running-suspended`, "Suspending task." |
| 23:38:26 | `staleDate` elapses, app suspended |
| 23:39:41.772 | `liveactivitiesd [staleParticipant]`: "Marking outdated activity as stale" |
| 23:39:41.790 | `chronod [extensionkit:launch]`: created new process `ExpoWidgetsTarget` |
| 23:39:48 | screenshot: Dynamic Island shows red `NO SIG` / `stale=true` (was green `72bpm` / `stale=false`) |

`runningboardd` logs no state change for the app between 23:38:12 and 23:39:50, so it
never resumed. The widget extension was cold-launched by `chronod` and re-evaluated the
layout JS on its own, which is the case this API is for.

Before (23:38) and after (23:39), same activity, same run. Settings is the foreground
app in both and the harness app is suspended. Look at the Dynamic Island: green
`72bpm` / `stale=false` becomes red `NO SIG` / `stale=true`.

20-BEFORE-stale
21-AFTER-stale

**One platform behaviour worth documenting.** The stale transition presents about 75s
after `staleDate`, not at it: 76s, 75s, 75s across three runs. `liveactivitiesd` appears
to mark activities stale on a coarse poll. My first attempt sampled at +20s and read as
a false negative, so anyone reproducing this should sample well past the stale date.

**Scope of what I observed.** Dynamic Island compact presentation only. The Lock Screen
`banner` goes through the same `environment(for: context)` call site, and both were
changed identically, but I couldn't script the simulator into a locked state to
photograph it. So I'm not claiming to have seen it.

Happy to share the patch used for the 57.0.6 backport if that's useful to anyone.

dirkpostma:
Could someone re-run the failing **iOS Unit Tests** job? I don't have the permissions to do it myself.

The failure looks unrelated to this PR ([run 30492295087](https://github.com/expo/expo/actions/runs/30492295087/job/90835039595?pr=48303)):

- `UpdatesLogReaderTests.BasicLoggingWorks` failed at 10:19:06. That's expo-updates logging, nothing this PR touches. The same suite took 16s for `PurgeOldLogs`, so the runner was already crawling.
- After that the job went silent. Last test output was at 10:23:30, then nothing until the step hit the 60 minute limit at 11:05. The cleanup step killed orphaned `xcodebuild`, `SWBBuildService` and `DTServiceHub` processes, so it looks like the simulator hung rather than a test failing.

All other checks on the head commit are green (`check-packages`, `ios-build`, `android-build`).

dirkpostma:
The re-run came back green ([job](https://github.com/expo/expo/actions/runs/30492295087/job/90915761252)), so the earlier failure was the hung simulator rather than anything in this change.

All checks on the head commit are passing now. Ready for review whenever you have a moment, @jakex7.

packages/expo-widgets/ios/Widgets/WidgetLiveActivity.swift
```diff
@@ -24,8 +24,12 @@ public struct WidgetLiveActivity: Widget {
...
   let widgetContext: AppContext = AppContext()
   
-  var environment: [String: Any] {
-    return getLiveActivityEnvironment(environment: env)
+  func environment(for context: ActivityViewContext<LiveActivityAttributes>) -> [String: Any] {
+    var env: [String: Any] = getLiveActivityEnvironment(environment: self.env)
+    if #available(iOS 16.2, *) {
+      env["isStale"] = context.isStale
+    }
+    return env
```
jakex7: At this point, we could move the 
```
    if #available(iOS 16.2, *) {
      env["isStale"] = context.isStale
    }
```
to `getLiveActivityEnvironment` and call it with environment and context directly at L42 and L57

jakex7 (changes requested):

jakex7 (approved):
Thank you for contributing!

PR #48310, opened 2026-07-30
[expo-audio][android] Emit an `onAudioFocusChanged` event

michaellcraig80-stack:
# Why

`AudioModule` already receives every Android audio focus change and already acts on it:

| Focus change | Current behaviour |
|---|---|
| `AUDIOFOCUS_LOSS` | pauses all playables |
| `AUDIOFOCUS_LOSS_TRANSIENT` | pauses those that are playing |
| `AUDIOFOCUS_LOSS_TRANSIENT_CAN_DUCK` | ducks, or pauses, depending on `interruptionMode` |
| `AUDIOFOCUS_GAIN` | restores volume and resumes whatever it paused |

None of it is surfaced to JS. So an app cannot tell a **brief** interruption from a **permanent** one: both arrive only as playback having stopped. An app that wants to resume by itself after a short interruption, but wait for an explicit user action after a phone call, has nothing to branch on.

It also cannot be obtained in userland. Android grants audio focus to a single listener per app, so a second `AudioFocusRequest` from app-level code **displaces this module's**, and this module's listener then pauses every player. The signal is not missing, only its surfacing.

# What this changes

- `Events("onAudioFocusChanged")` on the module.
- The listener emits `{ reason, interruptionMode }`.
- TS types for the event, following the `NativeModule<TEvents>` convention used elsewhere.

No `AudioFocusRequest` is added. No existing branch behaviour changes.

```ts
import { AudioModule } from 'expo-audio';

AudioModule.addListener('onAudioFocusChanged', ({ reason }) => {
  if (reason === 'lossTransient') {
    // brief: safe to resume when 'gain' arrives
  } else if (reason === 'loss') {
    // permanent: wait for the user
  }
});
```

# Two details worth reviewing

**The event is emitted after the `when` block, not before, and that is load-bearing rather than stylistic.** Emitting first lets a JS listener pause a player while `focusAcquired` is still `true`; `AudioPlayer.onPlaybackStateChange` then reaches `releaseAudioFocus()` and abandons the request, so no later `AUDIOFOCUS_GAIN` is delivered and the app never hears that focus came back. Every loss branch clears `focusAcquired` **before** pausing, which is exactly what keeps the request alive.

**The early `return@launch` in the gain branch became a positive condition** so the emit is not skipped on that path. Behaviour is unchanged.

**All gain-family values map to `'gain'`.** The request is made with `AUDIOFOCUS_GAIN_TRANSIENT` so the return is normally `AUDIOFOCUS_GAIN`, but matching only that constant would silently drop a transient variant and leave a listener waiting forever.

# Test plan

- `:expo-audio:compileDebugKotlin` passes, and the equivalent change compiles cleanly in an SDK 55 app with the emitted strings verified present in the resulting bytecode.
- Verified by hand in an app that discriminates interruption types: a transient loss followed by a gain resumes, a permanent loss does not.
- No behavioural change to existing consumers: the four branches are untouched and the gain restructure is a brace change.

# Scope

Android only. iOS already observes `interruptionNotification` and `routeChangeNotification` internally via `NotificationCenter`, so the same event could be added there. Happy to extend this PR if you would prefer it land cross-platform, or to rename `reason` if you would rather mirror the platform constants.

I will add the changelog entry once this has a PR number.

### DIFF
--- a/packages/expo-audio/android/src/main/java/expo/modules/audio/AudioModule.kt
+++ b/packages/expo-audio/android/src/main/java/expo/modules/audio/AudioModule.kt
@@ -123,19 +123,45 @@ class AudioModule : Module() {
         AudioManager.AUDIOFOCUS_GAIN -> {
           focusAcquired = true
 
-          if (!shouldPlayInSilentMode()) {
-            return@launch
-          }
-
-          allPlayables.forEach { playable ->
-            playable.setVolume(playable.previousVolume)
-            if (playable.isPaused) {
-              playable.isPaused = false
-              playable.play()
+          // NOTE: this was an early `return@launch`. It is a positive condition so
+          // that the event emitted below is not skipped on this path. Behaviour is
+          // unchanged.
+          if (shouldPlayInSilentMode()) {
+            allPlayables.forEach { playable ->
+              playable.setVolume(playable.previousVolume)
+              if (playable.isPaused) {
+                playable.isPaused = false
+                playable.play()
+              }
             }
           }
         }
       }
+
+      // Emitted AFTER the branches above have handled the change, deliberately.
+      //
+      // Emitting first would let a JS listener pause a player while `focusAcquired`
+      // is still true, and `AudioPlayer.onPlaybackStateChange` then reaches
+      // `releaseAudioFocus()` and abandons the request. Every loss branch above
+      // clears `focusAcquired` before pausing, which is what keeps a later
+      // AUDIOFOCUS_GAIN deliverable, so abandoning it here would mean the app never
+      // hears that focus came back.
+      sendEvent(
+        "onAudioFocusChanged",
+        mapOf(
+          "reason" to when (focusChange) {
+            AudioManager.AUDIOFOCUS_LOSS -> "loss"
+            AudioManager.AUDIOFOCUS_LOSS_TRANSIENT -> "lossTransient"
+            AudioManager.AUDIOFOCUS_LOSS_TRANSIENT_CAN_DUCK -> "lossTransientCanDuck"
+            AudioManager.AUDIOFOCUS_GAIN,
+            AudioManager.AUDIOFOCUS_GAIN_TRANSIENT,
+            AudioManager.AUDIOFOCUS_GAIN_TRANSIENT_MAY_DUCK,
+            AudioManager.AUDIOFOCUS_GAIN_TRANSIENT_EXCLUSIVE -> "gain"
+            else -> "unknown"
+          },
+          "interruptionMode" to interruptionMode?.value
+        )
+      )
     }
   }
 
@@ -225,6 +251,8 @@ class AudioModule : Module() {
   @OptIn(DelicateCoroutinesApi::class)
   override fun definition() = ModuleDefinition {
     Name("ExpoAudio")
+
+    Events("onAudioFocusChanged")
 
     OnCreate {
       audioManager = appContext.reactContext?.getSystemService(Context.AUDIO_SERVICE) as AudioManager
--- a/packages/expo-audio/src/AudioModule.types.ts
+++ b/packages/expo-audio/src/AudioModule.types.ts
@@ -20,9 +20,55 @@ import type { AudioLockScreenOptions } from './AudioConstants';
 import type { AudioStream } from './AudioStream.types';
 
 /**
+ * Why the system audio focus changed.
+ *
+ * - `'loss'`: focus was lost for the foreseeable future, for example another app
+ *   started playing music. Playback is not expected to resume on its own.
+ * - `'lossTransient'`: focus was lost briefly, for example an incoming call. The
+ *   system is expected to hand focus back with `'gain'`.
+ * - `'lossTransientCanDuck'`: another app wants focus but is happy for this app to
+ *   keep playing quietly. Whether the audio is ducked or paused depends on
+ *   `interruptionMode`.
+ * - `'gain'`: focus was regained after a loss.
+ * @platform android
+ */
+export type AudioFocusChangeReason =
+  | 'loss'
+  | 'lossTransient'
+  | 'lossTransientCanDuck'
+  | 'gain'
+  | 'unknown';
+
+/**
+ * @platform android
+ */
+export type AudioFocusChangeEvent = {
+  /** Why the focus changed. */
+  reason: AudioFocusChangeReason;
+  /** The interruption mode in effect when the change happened, which determines whether a duck request ducks or pauses. */
+  interruptionMode: AudioMode['interruptionMode'] | null;
+};
+
+/**
  * @hidden
  */
-export declare class NativeAudioModule extends NativeModule {
+export type AudioModuleEvents = {
+  /**
+   * Fires when the system audio focus changes.
+   *
+   * The module already handles focus changes internally (pausing, ducking and
+   * resuming). This event exists so an app can tell a brief interruption apart from
+   * a permanent one, which it cannot infer from playback simply stopping, and decide
+   * for itself whether resuming is appropriate.
+   * @platform android
+   */
+  onAudioFocusChanged(event: AudioFocusChangeEvent): void;
+};
+
+/**
+ * @hidden
+ */
+export declare class NativeAudioModule extends NativeModule<AudioModuleEvents> {
   setIsAudioActiveAsync(active: boolean): Promise<void>;
   setAudioModeAsync(category: Partial<AudioMode>): Promise<void>;
   requestRecordingPermissionsAsync(): Promise<PermissionResponse>;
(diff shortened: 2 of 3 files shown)


behenate:
@expo-bot verify

packages/expo-audio/android/src/main/java/expo/modules/audio/AudioModule.kt
```diff
@@ -123,19 +123,45 @@ class AudioModule : Module() {
...
-          }
-
-          allPlayables.forEach { playable ->
-            playable.setVolume(playable.previousVolume)
-            if (playable.isPaused) {
-              playable.isPaused = false
-              playable.play()
+          // NOTE: this was an early `return@launch`. It is a positive condition so
+          // that the event emitted below is not skipped on this path. Behaviour is
+          // unchanged.
```
behenate, line 128: This comment holds no value once the reader has no context of the previous code (after this is merged)
```suggestion
```

packages/expo-audio/android/src/main/java/expo/modules/audio/AudioModule.kt
```diff
@@ -123,19 +123,45 @@ class AudioModule : Module() {
...
+          "reason" to when (focusChange) {
+            AudioManager.AUDIOFOCUS_LOSS -> "loss"
+            AudioManager.AUDIOFOCUS_LOSS_TRANSIENT -> "lossTransient"
+            AudioManager.AUDIOFOCUS_LOSS_TRANSIENT_CAN_DUCK -> "lossTransientCanDuck"
+            AudioManager.AUDIOFOCUS_GAIN,
+            AudioManager.AUDIOFOCUS_GAIN_TRANSIENT,
+            AudioManager.AUDIOFOCUS_GAIN_TRANSIENT_MAY_DUCK,
+            AudioManager.AUDIOFOCUS_GAIN_TRANSIENT_EXCLUSIVE -> "gain"
+            else -> "unknown"
+          },
```
behenate, line 161: I think we should move this conversion into a separate helper function

packages/expo-audio/android/src/main/java/expo/modules/audio/AudioModule.kt
```diff
@@ -123,19 +123,45 @@ class AudioModule : Module() {
...
       }
+
+      // Emitted AFTER the branches above have handled the change, deliberately.
+      //
+      // Emitting first would let a JS listener pause a player while `focusAcquired`
+      // is still true, and `AudioPlayer.onPlaybackStateChange` then reaches
+      // `releaseAudioFocus()` and abandons the request. Every loss branch above
+      // clears `focusAcquired` before pausing, which is what keeps a later
+      // AUDIOFOCUS_GAIN deliverable, so abandoning it here would mean the app never
+      // hears that focus came back.
```
behenate, line 148: This comment is so complicated that it's literally easier to just read the code. Lets remove it completely, it adds no value (the code should document itself).

```suggestion

```

packages/expo-audio/android/src/main/java/expo/modules/audio/AudioModule.kt
```diff
@@ -226,6 +252,8 @@ class AudioModule : Module() {
   override fun definition() = ModuleDefinition {
     Name("ExpoAudio")
 
+    Events("onAudioFocusChanged")
```
behenate, line 255:
```suggestion
    Events("audioFocusChange")
```

Other events in expo-audio use the present form and without the "on" prefix, lets keep it that way.

packages/expo-audio/src/AudioModule.types.ts
```diff
@@ -19,10 +19,56 @@ import type {
...
+  /**
+   * Fires when the system audio focus changes.
+   *
+   * The module already handles focus changes internally (pausing, ducking and
+   * resuming). This event exists so an app can tell a brief interruption apart from
+   * a permanent one, which it cannot infer from playback simply stopping, and decide
+   * for itself whether resuming is appropriate.
+   * @platform android
+   */
+  onAudioFocusChanged(event: AudioFocusChangeEvent): void;
```
behenate, line 65:
```suggestion
  audioFocusChange(event: AudioFocusChangeEvent): void;
```

behenate (changes requested):
A good idea for a missing feature. Have you looked into implementing something similar for iOS? It's good to do both right away since we will want to have a shared event type and shape on both platforms. If we merge Android right away we'll then be locked to the android event shape in order to avoid breaking changes, which may be sub-optimal for iOS.

packages/expo-audio/src/AudioModule.types.ts
```diff
@@ -19,10 +19,56 @@ import type {
...
+ * @hidden
+ */
+export type AudioModuleEvents = {
+  /**
+   * Fires when the system audio focus changes.
+   *
+   * The module already handles focus changes internally (pausing, ducking and
+   * resuming). This event exists so an app can tell a brief interruption apart from
+   * a permanent one, which it cannot infer from playback simply stopping, and decide
+   * for itself whether resuming is appropriate.
```
amandeepmittal, line 62: ✏️ **style** `mechanics.md#oxford-commas`

"pausing, ducking and resuming" is missing its Oxford comma, and the sentence starting "This event exists so an app can tell" runs to 32 words. Add the comma and split the sentence.

```suggestion
   * The module already handles focus changes internally (pausing, ducking, and
   * resuming). Use this event to tell a brief interruption apart from a permanent
   * one, which playback stopping alone does not reveal. Your app can then decide
   * whether to resume.
```

packages/expo-audio/src/AudioModule.types.ts
```diff
@@ -19,10 +19,56 @@ import type {
...
+ * Why the system audio focus changed.
+ *
+ * - `'loss'`: focus was lost for the foreseeable future, for example another app
+ *   started playing music. Playback is not expected to resume on its own.
+ * - `'lossTransient'`: focus was lost briefly, for example an incoming call. The
+ *   system is expected to hand focus back with `'gain'`.
+ * - `'lossTransientCanDuck'`: another app wants focus but is happy for this app to
+ *   keep playing quietly. Whether the audio is ducked or paused depends on
+ *   `interruptionMode`.
+ * - `'gain'`: focus was regained after a loss.
```
amandeepmittal, line 32: ⚠️ **correctness** `prose.md#voice-and-tone`

The list documents four of the five values, so `'unknown'` reaches the reference with nothing explaining when it arrives. The descriptions are also passive throughout ("focus was lost", "Playback is not expected to resume"), which hides which app acted.

```suggestion
 * - `'loss'`: The system gave audio focus to another app for the foreseeable future.
 *   Playback does not resume on its own.
 * - `'lossTransient'`: The system took audio focus briefly, for example for an
 *   incoming call. It hands focus back with `'gain'`.
 * - `'lossTransientCanDuck'`: Another app requested focus but allows this app to keep
 *   playing at a lower volume. Whether the audio ducks or pauses depends on
 *   `interruptionMode`.
 * - `'gain'`: This app regained audio focus after a loss.
 * - `'unknown'`: The system reported a focus change that this library does not map.
```

packages/expo-audio/src/AudioModule.types.ts
```diff
@@ -19,10 +19,56 @@ import type {
...
+  | 'loss'
+  | 'lossTransient'
+  | 'lossTransientCanDuck'
+  | 'gain'
+  | 'unknown';
+
+/**
+ * @platform android
+ */
+export type AudioFocusChangeEvent = {
```
amandeepmittal, line 45: ✏️ **style** `prose.md#core-principles`

`AudioFocusChangeEvent` reaches the reference with an empty description, while `AudioSample`, `AudioEvents`, and every other type in this file opens with a summary sentence. Give it one.

```suggestion
/**
 * The payload passed to an `onAudioFocusChanged` listener.
 * @platform android
 */
export type AudioFocusChangeEvent = {
```

packages/expo-audio/src/AudioModule.types.ts
```diff
@@ -19,10 +19,56 @@ import type {
...
+  /** Why the focus changed. */
+  reason: AudioFocusChangeReason;
+  /** The interruption mode in effect when the change happened, which determines whether a duck request ducks or pauses. */
+  interruptionMode: AudioMode['interruptionMode'] | null;
```
amandeepmittal, line 49: ✏️ **style** `prose.md#core-principles`

The field can be `null` but the description never says when, so a listener has no reason to handle it. It stays `null` until your app calls `setAudioModeAsync`.

```suggestion
  /**
   * The interruption mode in effect when the change happened, which decides whether
   * the audio ducks or pauses. This is `null` until your app calls
   * `setAudioModeAsync`.
   */
  interruptionMode: AudioMode['interruptionMode'] | null;
```

amandeepmittal (changes requested):

Issue #20937, opened 2023-01-24
[expo-notifications] Clashing with OneSignal


jacklj:
### Summary

I've got react-native-onesignal installed on a bare React Native project, and I've since installed expo-notifications to fire local notifications. I get this warning on iOS startup (it comes from this code [here](https://github.com/expo/expo/blob/e01c62aa8d02e18c80f5fadd10a9448a7c69628b/packages/expo-notifications/ios/EXNotifications/Notifications/EXNotificationCenterDelegate.m#L32):

```
EXNotificationCenterDelegate encountered already present delegate of UNUserNotificationCenter: <OSUNUserNotificationCenterDelegate: 0x600003c68460>. EXNotificationCenterDelegate will not overwrite the value not to break other features of your app. In return, expo-notifications may not work properly. To fix this problem either remove setting of the second delegate or set the delegate to an instance of EXNotificationCenterDelegate manually afterwards.
```

I believe this is because OneSignal already sets its own delegate.

The result is: local notifications get delivered, but neither the `Notifications.setNotificationHandler()` handler nor the OneSignal equivalent get fired. 

To fix, I've tried to follow the advice given - `set the delegate to an instance of EXNotificationCenterDelegate manually afterwards.`:

```
- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
{
  // ... 

  EXNotificationCenterDelegate *expoNotifDelegate = [[EXNotificationCenterDelegate alloc] init];
  [[UNUserNotificationCenter currentNotificationCenter] setDelegate:expoNotifDelegate];
  
  //...
}
```

This builds successfully but then local notifications don't even fire.

This should be compatible with OneSignal, see here https://github.com/OneSignal/react-native-onesignal/issues/958#issuecomment-600267864:
> OneSignal does swizzle the UNUserNotificationCenter so even if you have [[UNUserNotificationCenter currentNotificationCenter] setDelegate:self]; in your code it shouldn't break OneSignal.

 Any ideas how to make expo-notifications work with OneSignal on iOS (it works fine on Android)? Thanks

### What platform(s) does this occur on?

iOS

### Environment

  expo-env-info 1.0.5 environment info:
    System:
      OS: macOS 12.6
      Shell: 5.8.1 - /bin/zsh
    Binaries:
      Node: 16.14.0 - ~/.nvm/versions/node/v16.14.0/bin/node
[...]

PR #48313, opened 2026-07-30
[notifications] support multiple UNUserNotificationCenterDelegate objects

vonovak:
# Why

motivation: https://github.com/expo/expo/issues/20937

expo-notifications on iOS currently assigns itself as the only notification center delegate. Some other libraries might be interested in becoming delegates also, and this PR allows that by forwarding the notification events to them also. 

\*\*Note that while we allow other libraries to become a delegate, we do not test for this scenario and can't guarantee compatibility with other libraries. We recommend expo-notifications to be the only library to handle notifications.**

# How

Remeber previous delegate, if set, and chain the callbacks received from iOS to that delegate also.

# Test Plan

On-device test suite still passes. Also tested manually with this fake delegate:

```
final class FakeOtherLibraryDelegate: NSObject, UNUserNotificationCenterDelegate {
  // The delegate property of the notification center is weak, and expo-notifications holds the
  // chained delegate weakly too. This static keeps the object alive.
  static let shared = FakeOtherLibraryDelegate()

  func takeDelegateSlotThenLetExpoChain() {
    let center = UNUserNotificationCenter.current()

    center.delegate = self
    NSLog("[FAKE] step 1: took the delegate slot, delegate is \(Self.describe(center.delegate))")

    NotificationCenterManager.shared.installAsNotificationCenterDelegate()
    NSLog("[FAKE] step 2: after expo-notifications installed, delegate is \(Self.describe(center.delegate))")

    if center.delegate === NotificationCenterManager.shared {
      NSLog("[FAKE] step 3: expo-notifications owns the slot. Every [FAKE] call below is forwarded.")
    } else {
      NSLog("[FAKE] step 3: PROBLEM — expo-notifications did not take the slot.")
    }
  }

  func userNotificationCenter(
    _ center: UNUserNotificationCenter,
    willPresent notification: UNNotification,
    withCompletionHandler completionHandler: @escaping (UNNotificationPresentationOptions) -> Void
  ) {
    NSLog("[FAKE] willPresent forwarded, id \(notification.request.identifier)")
    // expo-notifications discards these options when one of its own delegates already answered,
    // and uses them when none of them answered.
    NSLog("[FAKE] willPresent answering with .banner and .sound")
    completionHandler([.banner, .sound])
  }

  func userNotificationCenter(
    _ center: UNUserNotificationCenter,
    didReceive response: UNNotificationResponse,
    withCompletionHandler completionHandler: @escaping () -> Void
  ) {
    NSLog("[FAKE] didReceive forwarded, action \(response.actionIdentifier)")
    completionHandler()
  }

  func userNotificationCenter(_ center: UNUserNotificationCenter, openSettingsFor notification: UNNotification?) {
    NSLog("[FAKE] openSettings forwarded, notification \(notification?.request.identifier ?? "nil")")
  }

  private static func describe(_ delegate: UNUserNotificationCenterDelegate?) -> String {
    guard let delegate else {
      return "nil"
    }
    return String(describing: type(of: delegate))
  }
}

```

# Checklist


vonovak:
* **#48313** <a href="https://app.graphite.com/github/pr/expo/expo/48313?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/48313?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

douglowder (approved):

PR #48322, opened 2026-07-30
[expo-type-information] update docs

HubertBer:
# Why
The docs of `expo-type-information` package haven't beed updated in latest PR's, need to sync them with the current package version.

# How
Updated docs based on the changes in the whole stack of PR's.

# Test Plan
- [x] Checked the generated docs page

# Checklist


docs/pages/modules/type-generation-reference.mdx
```diff
@@ -60,6 +61,8 @@ Generates a full TypeScript interface for a Swift module. It consists of:
 - **view.tsx** for each view defined in the module
 - **index.ts** file which reexports some functions
 
+If there are multiple modules provided, then a **Common.types.ts** file will be generated. It exports the common declarations (enums, structs), so they are defined exactly once.
```
amandeepmittal: ✏️ **style** `prose.md#voice-and-tone`

"will be generated" is future tense and passive. Name the tool as the actor and stay in present tense, like the rest of this section.

```suggestion
When you provide multiple modules, the tool also generates a **Common.types.ts** file. It exports the common declarations (enums and structs), so each one is defined exactly once.
```

docs/pages/modules/type-generation-reference.mdx
```diff
@@ -72,7 +75,7 @@ Creates a TypeScript interface for every Swift inline module in the project. The
 **Options:**
 | Flag | Description | Default |
 | :--- | :--- | :--- |
-| `-a --app-json <appJsonPath>` | A path to the [app config](/workflow/configuration/) file where the `inline.modules.watchedDirectories` are defined. | |
+| `-a, --app-json <appJsonPath>` | A path to the [app config](/workflow/configuration/) file where the `expo.experiments.inlineModules.watchedDirectories` are defined. | |
 | `-w --watcher` | Starts a watcher that checks for changes in inline modules files. | |
```
amandeepmittal: 💡 **nit** `mechanics.md#check-these-vale-cannot-see-them`

Every other flag on this page now reads `-w, --watcher`, so this row is the last one still missing the comma.

```suggestion
| `-w, --watcher` | Starts a watcher that checks for changes in inline modules files. | |
```

docs/pages/modules/type-generation-reference.mdx
```diff
@@ -133,9 +136,9 @@ Print the preprocessed file(s) in the state right before parsing them using `sou
 
 <Collapsible summary="Known issues about the current version of parsing the Swift file">
 
-- Nested classes will not be resolved fully.
+- Nested declarations will not be resolved fully.
 
-  `Sourcekitten` has a limit on resolving nested closures, so classes defined in your module may not be fully resolved. This is the reason why if there is a DSL `Class` in a module, its methods will have their return types `unresolved`.
+  `Sourcekitten` has a limit on resolving nested closures, so declarations nested deeper in your module (such as methods of a DSL `Class`) may not be fully resolved. Type inference doesn't work at this nesting depth, so only explicitly annotated types can be resolved there.
```
amandeepmittal: 💡 **nit** `terminology.md#external-product-casing`

The tool is spelled "SourceKitten", which is the spelling the options table on this page already uses.

```suggestion
  `SourceKitten` has a limit on resolving nested closures, so declarations nested deeper in your module (such as methods of a DSL `Class`) may not be fully resolved. Type inference doesn't work at this nesting depth, so only explicitly annotated types can be resolved there.
```

docs/pages/modules/type-generation-reference.mdx
```diff
@@ -148,15 +151,72 @@ Print the preprocessed file(s) in the state right before parsing them using `sou
 
 - Unsupported Expo Module DSL declarations.
 
-  Not all of the DSL declarations are parsed right now, and some are not parsed in every context. For example `Events` are parsed inside a `View` but not in a Module definition.
+  Most of the DSL declarations are parsed, but some details are not. For example, property setters (`Property(...).get { ... }.set { ... }`) are not modeled, so every generated property is `readonly`. Event payload types are not resolved, so generated event maps type the payload as a generic object.
```
amandeepmittal: ✏️ **style** `prose.md#voice-and-tone`

All three sentences are passive ("are parsed", "are not modeled", "are not resolved"). Make the tool the subject.

```suggestion
  The tool parses most DSL declarations, but not every detail. It doesn't model property setters (`Property(...).get { ... }.set { ... }`), so every generated property is `readonly`. It doesn't resolve event payload types either, so generated event maps type the payload as a generic object.
```

docs/pages/modules/type-generation-reference.mdx
```diff
@@ -148,15 +151,72 @@ Print the preprocessed file(s) in the state right before parsing them using `sou
...
+
+- `SharedRef` types lose their `nativeRefType` tag.
+
+  A Swift `SharedRef<InnerType>` is generated as `SharedRef<any>` instead of `SharedRef<'nativeRefType'>`. The tool doesn't extract the `nativeRefType` string that a `SharedRef` subclass overrides (for example, `'image'` in `expo-image`), so the generated references are compatible with any `SharedRef` and are not checked against the tagged `SharedRef` types that other libraries declare.
```
amandeepmittal: ✏️ **style** `prose.md#sentence-construction`

The second sentence runs 39 words and carries three facts: what the tool skips, why, and what that means for type checking. Split it into three sentences.

```suggestion
  The tool generates a Swift `SharedRef<InnerType>` as `SharedRef<any>` instead of `SharedRef<'nativeRefType'>`. It doesn't extract the `nativeRefType` string that a `SharedRef` subclass overrides, such as `'image'` in `expo-image`. The generated references therefore work with any `SharedRef`, and are not checked against the tagged `SharedRef` types other libraries declare.
```

docs/pages/modules/type-generation-reference.mdx
```diff
@@ -148,15 +151,72 @@ Print the preprocessed file(s) in the state right before parsing them using `sou
...
-| **Swift `struct` & `class`** | Must conform to the `Record` protocol. Only properties marked with the `@Field` annotation are parsed.                          |
-| **Swift `enum`**             | Basic cases are supported. Values associated with enum cases are not currently parsed.                                          |
+| Feature                      | Notes and limitations                                                                                                                                                             |
+| :--------------------------- | :-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| **Expo DSL Declarations**    | Supports parsing for: `AsyncFunction`, `Class`, `Constant`, `Constructor`, `Events`, `Function`, `Name`, `Prop`, `Property`, `StaticAsyncFunction`, `StaticFunction`, and `View`. |
+| **Swift `struct` & `class`** | structs and classes conforming to `Record` protocol (technically classes/structs which declare `@Field`). Only properties marked with the `@Field` annotation are parsed.         |
```
amandeepmittal: ⚠️ **correctness** `prose.md#sentence-construction`

The parser treats a struct or class as a record when it declares at least one `@Field` **or** when it conforms to `Record` directly, so the two are alternatives rather than one being a technicality of the other. The cell also starts mid-sentence, unlike the two rows around it.

```suggestion
| **Swift `struct` & `class`** | Structs and classes that declare at least one `@Field`, or that conform to the `Record` protocol directly. Only properties marked with the `@Field` annotation are parsed.        |
```

docs/pages/modules/type-generation-reference.mdx
```diff
@@ -148,15 +151,72 @@ Print the preprocessed file(s) in the state right before parsing them using `sou
...
+| **Swift `enum`**             | Basic cases are supported. Values associated with enum cases are not currently parsed.                                                                                            |
+
+### Type conversions
```
amandeepmittal: ✏️ **style** `links-and-headings.md#titles-and-headings`

"Type conversions" and "Generated events" describe what the tool does, not what it fails to do, but as H3s the page outline files them under `## Swift parser limitations`. Promote both to H2 and move their subsections up one level.

docs/pages/modules/type-generation-reference.mdx
```diff
@@ -148,15 +151,72 @@ Print the preprocessed file(s) in the state right before parsing them using `sou
...
+
+### Generated events
+
+Events are generated differently depending on where they are declared:
+
+- **Module events**: `Events(...)` in a module definition generates a `<ModuleName>Events` type, and the module class extends `NativeModule<<ModuleName>Events>`. Module events are sent with a dictionary body, so the payload is typed as `Record<string, any>`.
+- **Class events**: `Events(...)` in a DSL `Class` generates a `<ClassName>Events` type, and the class extends `SharedObject<<ClassName>Events>`. Shared objects emit a single payload of any type, or none at all, so the listener is typed as `(payload?: any) => void`.
+- **View events**: events declared in a `View` become optional callback props on the generated props interface, typed as `(event: any) => void`.
+
+The payload types of events are not resolved from the native code, which is why the payloads are typed permissively.
```
amandeepmittal: ✏️ **style** `review-ai-text.md#t5-restatement`

Each of the three bullets above already states how its payload is typed and why, and the DSL limitations section says it a third time. Cut this sentence.

docs/pages/modules/type-generation-tutorial.mdx
```diff
@@ -136,36 +136,29 @@ In our case we just reexport the `Hello` constant from the native module.
...
-  url: URL;
+  url: string;
   onLoad?: (event: any) => void;
 }
 
 export declare class FirstInlineViewNativeModuleType extends NativeModule {}
 ```
 
-Looking at the "generated" file, we can see that not all of the types from the `FirstInlineView.swift` could have been resolved, a `URL` type is set to `unknown`. This may happen when the type is not a basic type (which are mapped manually by the tool) and is not defined in provided files (in this case it hasn't been defined in the `FirstInlineView.swift`), or if the tool failed to parse its definition.
+Looking at the "generated" file, we can see that an `ExpoWebViewProps` interface has been generated, it has the props and events from `FirstInlineView`. The Swift `URL` type of the `url` prop correctly maps to `string` in TypeScript.
```
amandeepmittal: ✏️ **style** `prose.md#sentence-construction`

"has been generated, it has the props and events" is a comma splice.

```suggestion
Looking at the "generated" file, we can see that an `ExpoWebViewProps` interface has been generated with the props and events from `FirstInlineView`. The Swift `URL` type of the `url` prop correctly maps to `string` in TypeScript.
```

docs/pages/modules/type-generation-tutorial.mdx
```diff
@@ -136,36 +136,29 @@ In our case we just reexport the `Hello` constant from the native module.
...
 
-We can also see that a `ExpoWebViewProps` interface has been generated, it has the props and events from `FirstInlineView`.
+Types that the tool cannot map or find a definition for are different. When a type is not one of the known convertible types and is not defined in the provided files (or the tool failed to parse its definition), the tool generates an `export type SomeType = unknown;` alias for it.
```
amandeepmittal: ✏️ **style** `prose.md#sentence-construction`

"are different" leaves the comparison unstated, and the sentence explaining it runs to 38 words.

```suggestion
The tool handles types it cannot resolve differently. For a type that is neither a convertible type nor defined in the provided files, it generates an `export type SomeType = unknown;` alias. It does the same when it fails to parse a definition.
```

docs/pages/modules/type-generation-tutorial.mdx
```diff
@@ -410,31 +403,28 @@ The `--module` option (short `-m`) is a path to the root folder of the module. A
...
 }
+
+// These events may have payloads that weren't resolved!
+export type ExpoSettingsEvents = {
+  onChangeTheme: (payload: Record<string, any>) => void;
+};
 ```
 
-The `*.types.ts` file contains the definitions of all types defined in the module. In our case we've only declared a `Theme` enum which has correctly been put in the file. Note however that in contrast to the `ExpoSettings.types.ts` from the tutorial, the events type have not been generated. The `expo-type-information` tool is new and powerful, however not every option is yet implemented, module events being one of them.
+The `*.types.ts` file contains the definitions of all types defined in the module. In our case we've declared a `Theme` enum which has correctly been put in the file, together with an `ExpoSettingsEvents` type describing the module's events. Note that the tool doesn't resolve the payload types of events, so the payload is typed as a generic object.
```
amandeepmittal: ✏️ **style** `prose.md#voice-and-tone`

"has correctly been put in the file" is passive and leaves the actor out. Say what the tool emitted.

```suggestion
The `*.types.ts` file contains the definitions of all types defined in the module. In our case we declared a `Theme` enum, and the tool emitted it here along with an `ExpoSettingsEvents` type for the module's events. Note that the tool doesn't resolve event payload types, so the payload is typed as a generic object.
```

amandeepmittal (changes requested):

amandeepmittal (approved):

PR #48318, opened 2026-07-30
[ios][core] Add a foreground-scene helper and use it in store-review

alanjhughes:
# Why

Same follow-up as #48315.

`StoreReviewModule` hand-rolled a scene lookup that filtered across every scene type and cast afterwards, so a foreground scene could be missed when another scene type sorted ahead of it.

It could not use `windowScene(for:)`, which falls back to any connected scene. Store-review depends on nil to throw when nothing is on screen, because a review prompt requested into a background scene is never seen.

# How

There are three distinct scene contracts in the SDK and now each has a name:

| Helper | Contract | Callers |
| --- | --- | --- |
| `windowScene(for:)` | best available, something whenever any scene exists | layout and geometry |
| `foregroundActiveScene()` | strictly active or nothing | dev menu, updates: attach a window or don't |
| `foregroundScene()` | on screen or nil | store-review: present or throw |

`windowScene(for:)` is now expressed in terms of `foregroundScene()`, so the ordering lives in one place.

This also removes `SceneGeometry.isInterfaceOrientationLocked`, added in #48168 for a consumer that turned out not to work: `getOrientationLockAsync` returns a mask-derived enum while that property is a `Bool`, so it can only contradict the answer rather than supply it. Adding public API in the same PR that removes some is deliberate, not inconsistent. This one ships with a caller, that one never found one, and it is unreleased so removing it now is free.

# Test Plan

Trigger the review prompt. This package has no test target, so that is its only coverage.

PR #48323, opened 2026-07-30
[ios][expo-go] Resolve dev menu scenes through the shared helper

alanjhughes:
# Why

Expo Go keeps its own copy of the dev menu, so the fixes from #48171 and #48317 never reached it. The same four defects were still present.

# How

The window was created from `UIScreen.main.bounds`, so it sized itself to the display rather than the app's window. Two scene lookups filtered across every scene type before casting, so a foreground-active non-window scene could shadow a real one. And the FAB read its safe area from `windows.first` rather than the key window.

All four now go through `SceneGeometry`. `UIScreen.main` and `connectedScenes` no longer appear in the fork.

Nothing here needs the open stack above it: every helper used landed in #48168.

# Test Plan

Open the dev menu and check it fills the window rather than the display, and that the FAB sits at the right inset and picks the correct edge.

PR #48324, opened 2026-07-30
[ios][expo-go] Resolve client scenes through the shared helper

alanjhughes:
# Why

Five lookups in the Expo Go client UI reached for a window across every connected scene, so they could return one belonging to a scene the user isn't looking at.

The share sheet in project details was the worst: it took `connectedScenes.first` with no activation filter at all, then that scene's first window.

# How

All five go through `SceneGeometry`. The others are the two auth presentation anchors, the theme override and the review prompt.

Two things left deliberately alone. The theme override still applies to a single window, so with multiple scenes the others keep the old appearance. That is a question of how many windows rather than which, so it is a separate change. And the review prompt keeps `foregroundActiveScene()` rather than the `foregroundScene()` added in #48318, which also accepts a foreground-inactive scene for the App Switcher case. Aligning the two is a one-line follow-up once that merges, and doing it here would make this PR depend on unmerged work.

Also worth noting the two `AuthPresentationContextProvider` classes are byte-identical duplicates. Now that each body is one line the duplication is more visible, but sharing one is tidying rather than part of this.

# Test Plan

Sign in, which exercises the auth anchor. Share a project from its details page. Switch theme. The review prompt is threshold-gated so it is harder to trigger deliberately.

### DIFF
--- a/apps/expo-go/ios/Client/SwiftUI/Services/AuthenticationService.swift
+++ b/apps/expo-go/ios/Client/SwiftUI/Services/AuthenticationService.swift
@@ -3,6 +3,7 @@
 import Foundation
 import AuthenticationServices
 import Combine
+import ExpoModulesCore
 
 @MainActor
 class AuthenticationService: ObservableObject {
@@ -184,10 +185,6 @@ class AuthenticationService: ObservableObject {
 
 private class AuthPresentationContextProvider: NSObject, ASWebAuthenticationPresentationContextProviding {
   func presentationAnchor(for session: ASWebAuthenticationSession) -> ASPresentationAnchor {
-    let window = UIApplication.shared.connectedScenes
-      .compactMap { $0 as? UIWindowScene }
-      .flatMap { $0.windows }
-      .first { $0.isKeyWindow }
-    return window ?? ASPresentationAnchor()
+    return SceneGeometry.keyWindow() ?? ASPresentationAnchor()
   }
 }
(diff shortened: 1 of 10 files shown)


alanjhughes:
* **#48326** <a href="https://app.graphite.com/github/pr/expo/expo/48326?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#48325** <a href="https://app.graphite.com/github/pr/expo/expo/48325?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#48324** <a href="https://app.graphite.com/github/pr/expo/expo/48324?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/48324?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#48323** <a href="https://app.graphite.com/github/pr/expo/expo/48323?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

tsapeta (approved):

alanjhughes:
### Merge activity

* **Jul 30, 3:28 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/expo/expo/48324).
* **Jul 30, 3:29 PM UTC**: @alanjhughes merged this pull request with [Graphite](https://app.graphite.com/github/pr/expo/expo/48324).

PR #48327, opened 2026-07-30
[router] Stabilize async routes on web and enable them by default

hassankhan:
# Why

Stabilize async routes on web and enable them by default for SDK 58, in development and production. Native async routes remain experimental and opt-in.

# How

Always set `web: true` as the default in `expo-router`'s config plugin options. `asyncRoutes` can be disabled by either explicitly setting `web: false`, or by setting `default: false` to disable it for all platforms.

E2E test cases and documentation have been updated as appropriate.

# Test Plan

- CI
- Manual testing

# Checklist

### DIFF
--- a/packages/@expo/cli/e2e/__tests__/export/static-splitting.test.ts
+++ b/packages/@expo/cli/e2e/__tests__/export/static-splitting.test.ts
@@ -29,7 +29,6 @@ describe('exports static with bundle splitting', () => {
           NODE_ENV: 'production',
           EXPO_USE_STATIC: 'static',
           E2E_ROUTER_SRC: 'static-rendering',
-          E2E_ROUTER_ASYNC: 'production',
         },
       }
     );
--- a/packages/expo-router/plugin/src/withRouter.ts
+++ b/packages/expo-router/plugin/src/withRouter.ts
@@ -1,5 +1,5 @@
 import { validate } from '@expo/schema-utils';
-import { ConfigPlugin, withInfoPlist, withPodfile } from 'expo/config-plugins';
+import { type ConfigPlugin, withInfoPlist, withPodfile } from 'expo/config-plugins';
 
 const schema = require('../options.json');
 
@@ -74,7 +74,7 @@ export type Props = {
   platformRoutes?: boolean;
   /** Enable or disable automatically generated routes. Defaults to `true`. */
   sitemap?: boolean;
-  /** Should Async Routes be enabled. `production` is currently web-only and will be disabled on native. */
+  /** Should Async Routes be enabled. Web defaults to true, other platforms default to false. */
   asyncRoutes?:
     | AsyncRouteOption
     | {
@@ -112,16 +112,42 @@ const withRouter: ConfigPlugin<Props | void> = (config, _props) => {
   withExpoHeadIos(config);
   withGammaScreens(config);
 
+  const router = normalizeProps({
+    ...config.extra?.router,
+    ...props,
+  });
+
   return {
     ...config,
     extra: {
       ...config.extra,
-      router: {
-        ...config.extra?.router,
-        ...props,
-      },
+      router,
     },
   };
 };
+
+function normalizeProps(props: Props) {
+  return normalizeAsyncRoutesProp(props);
+}
+
+function normalizeAsyncRoutesProp(props: Props) {
+  const asyncRoutes = props.asyncRoutes;
+
+  if (asyncRoutes == null) {
+    return {
+      ...props,
+      asyncRoutes: { web: true },
+    };
+  }
+
+  if (typeof asyncRoutes === 'object' && asyncRoutes.web == null && asyncRoutes.default == null) {
+    return {
+      ...props,
+      asyncRoutes: { ...asyncRoutes, web: true },
+    };
+  }
+
+  return props;
+}
 
 export default withRouter;
--- a/apps/router-e2e/app.config.js
+++ b/apps/router-e2e/app.config.js
@@ -67,7 +67,7 @@ module.exports = {
             ? true
             : process.env.E2E_ROUTER_ASYNC === 'false'
               ? false
-              : process.env.E2E_ROUTER_ASYNC || false,
+              : process.env.E2E_ROUTER_ASYNC,
         root: path.join('__e2e__', process.env.E2E_ROUTER_SRC ?? 'static-rendering', 'app'),
         origin: 'http://localhost:8081/',
         sitemap:
--- a/packages/@expo/cli/e2e/__tests__/export/base-path.test.ts
+++ b/packages/@expo/cli/e2e/__tests__/export/base-path.test.ts
@@ -21,7 +21,6 @@ describe('static-rendering with a custom base path', () => {
         EXPO_E2E_BASE_PATH: baseUrl,
         EXPO_USE_STATIC: 'static',
         E2E_ROUTER_SRC: 'static-rendering',
-        E2E_ROUTER_ASYNC: 'development',
       },
     });
   });
--- a/packages/@expo/cli/e2e/__tests__/export/export-embed-rsc.test.ts
+++ b/packages/@expo/cli/e2e/__tests__/export/export-embed-rsc.test.ts
@@ -49,13 +49,9 @@ jest.unmock('resolve-from');
       {
         env: {
           NODE_ENV: 'production',
-
           E2E_ROUTER_SRC: '01-rsc',
-          E2E_ROUTER_ASYNC: 'development',
-
           EXPO_USE_STATIC: 'single',
           E2E_ROUTER_JS_ENGINE: 'hermes',
-
           E2E_RSC_ENABLED: '1',
           TEST_SECRET_VALUE: 'test-secret',
         },
--- a/packages/@expo/cli/e2e/__tests__/export/export-no-ssg.test.ts
+++ b/packages/@expo/cli/e2e/__tests__/export/export-no-ssg.test.ts
@@ -21,7 +21,6 @@ describe('export-no-ssg', () => {
           NODE_ENV: 'production',
           EXPO_USE_STATIC: 'server',
           E2E_ROUTER_SRC: 'server',
-          E2E_ROUTER_ASYNC: 'development',
         },
       }
     );
--- a/packages/@expo/cli/e2e/__tests__/export/modal-splitting.test.ts
+++ b/packages/@expo/cli/e2e/__tests__/export/modal-splitting.test.ts
@@ -22,7 +22,6 @@ describe('exports static splitting with modal', () => {
         NODE_ENV: 'production',
         EXPO_USE_STATIC: 'static',
         E2E_ROUTER_SRC: 'modal-splitting',
-        E2E_ROUTER_ASYNC: 'production',
       },
     });
   });
--- a/packages/@expo/cli/e2e/__tests__/export/no-bytecode.test.ts
+++ b/packages/@expo/cli/e2e/__tests__/export/no-bytecode.test.ts
@@ -24,7 +24,6 @@ describe('exports for hermes with no bytecode', () => {
           EXPO_USE_STATIC: 'static',
           E2E_ROUTER_JS_ENGINE: 'hermes',
           E2E_ROUTER_SRC: 'url-polyfill',
-          E2E_ROUTER_ASYNC: 'development',
         },
       }
     );
@@ -81,7 +80,6 @@ describe('exports for hermes with no bytecode and no minification', () => {
           EXPO_USE_STATIC: 'static',
           E2E_ROUTER_JS_ENGINE: 'hermes',
           E2E_ROUTER_SRC: 'url-polyfill',
-          E2E_ROUTER_ASYNC: 'development',
         },
       }
     );
--- a/packages/@expo/cli/e2e/__tests__/export/no-sitemap.test.ts
+++ b/packages/@expo/cli/e2e/__tests__/export/no-sitemap.test.ts
@@ -17,7 +17,6 @@ describe('static-rendering with no sitemap', () => {
         NODE_ENV: 'production',
         EXPO_USE_STATIC: 'static',
         E2E_ROUTER_SRC: 'static-rendering',
-        E2E_ROUTER_ASYNC: 'development',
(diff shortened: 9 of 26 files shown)


docs/pages/router/web/async-routes.mdx
```diff
@@ -6,7 +6,7 @@ description: Learn how to speed up development with async bundling in Expo Route
 import { ContentSpotlight } from '~/ui/components/ContentSpotlight';
 import { Terminal } from '~/ui/components/Snippet';
 
-> **important** Async routes is in [alpha](/more/release-statuses/#alpha).
+> **important** Async routes remain experimental ([alpha](/more/release-statuses/#alpha)), including when enabled by default on web. Native async routes are unchanged in SDK 58.
```
hassankhan: This is unnecessary

docs/pages/versions/unversioned/sdk/router/index.mdx
```diff
@@ -75,8 +75,8 @@ If you are using the [default](/more/create-expo/#--template) template to create
     {
       name: 'asyncRoutes',
       description:
-        'Enable async routes (lazy loading). Can be a boolean, a string (`"development"` or `"production"`), or an object with platform-specific values (`{ android, ios, web, default }`). `production` is currently web-only and will be disabled on native.',
-      default: 'undefined',
+        'Enable experimental async routes (lazy loading). In SDK 58 and later, enabled by default on web in development and production, and disabled by default on native. Can be a boolean, a string (`"development"` or `"production"`), or an object with platform-specific values (`{ android, ios, web, default }`). Set `false`, `{ web: false }`, or `{ default: false }` to opt out on web. An explicit platform value takes precedence over `default`. Production async routes are web-only.',
```
hassankhan:
```suggestion
        'Enable experimental async routes (lazy loading). In SDK 58 and later, this is enabled by default on web in development and production, and disabled by default on native. Can be a boolean, a string (`"development"` or `"production"`), or an object with platform-specific values (`{ android, ios, web, default }`). Explicit platform values take precedence over `default`. Production async routes are web-only.',
```

packages/@expo/cli/e2e/playwright/prod/optimize-pass.test.ts
```diff
@@ -32,6 +32,7 @@ test.describe(inputDir, () => {
         NODE_ENV: 'production',
         EXPO_USE_STATIC: 'static',
         E2E_ROUTER_SRC: 'tree-shaking',
+        E2E_ROUTER_ASYNC: 'false',
```
jakub-agent, line 35: [🚨 **critical**] `e2e/playwright/dev/stable-module-ids.test.ts` needs this same line and doesn't have it — that's why `playwright-ubuntu (3)` is red here and green on the base branch. The job log fails at line 53 on `__r("packages/expo-router/entry.js")`.

It exports `E2E_ROUTER_SRC=compiler` with no `E2E_ROUTER_ASYNC`, then takes the first `.js` under `_expo/static/js`, which is now an async chunk. Can you add `E2E_ROUTER_ASYNC: 'false'` there too — or better, make the assertion select the entry bundle rather than whichever file sorts first?

packages/expo-router/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - Enable data loaders by default for static and server output ([#50118](https://github.com/expo/expo/pull/50118) by [@hassankhan](https://github.com/hassankhan))
 - Enable server rendering by default for `web.output: "server"` ([#50120](https://github.com/expo/expo/pull/50120) by [@hassankhan](https://github.com/hassankhan))
 - Add the `apiRoutes` config plugin option to support API routes with static rendering. ([#50148](https://github.com/expo/expo/pull/50148) by [@hassankhan](https://github.com/hassankhan))
+- Enable experimental async routes by default on web in SDK 58. Native defaults remain unchanged. ([#48327](https://github.com/expo/expo/pull/48327) by [@hassankhan](https://github.com/hassankhan))
```
jakub-agent, line 13: [🚨 **critical**] This flips behaviour for existing apps, so it should also get a `🛠 Breaking changes` entry. The concrete one: `useScreens.tsx` resolves the fallback as `EXPO_ROUTER_IMPORT_MODE === 'lazy' ? DefaultSuspenseFallback : (LayoutSuspenseFallback ?? ...)`, so every web app exporting a custom `SuspenseFallback` from a layout silently gets the default one after upgrading.

Can the entry name what changes and point at `asyncRoutes: { web: false }` as the way back? `async-routes.mdx` has it under Caveats, but nobody reading an upgrade diff lands there.

packages/expo-router/plugin/src/withRouter.ts
```diff
@@ -147,16 +147,38 @@ const withRouter: ConfigPlugin<Props | void> = (config, _props) => {
...
-      router: {
-        ...config.extra?.router,
-        ...props,
-      },
+      router,
     },
   };
 };
 
+function normalizeAsyncRoutesProp(props: Props) {
```
jakub-agent, line 164: [🏗️ **design**] This applies the new default purely on which `expo-router` is installed — there's no `sdkVersion` check. #50129, directly below in the stack and making the same SDK 58 cutover, gates on `sdkMajor >= 58` in `@expo/config` and tests `preserves the implicit mode for SDK 55/57`.

Was the difference deliberate? A project pinned to `sdkVersion: '57.0.0'` with an SDK 58 `expo-router` gets async routes on web from this PR but keeps `single` output from #50129 — a combination neither PR documents.

packages/@expo/cli/e2e/__tests__/export/static-splitting.test.ts
```diff
@@ -20,7 +20,7 @@ describe('exports static with bundle splitting', () => {
   const outputDir = path.join(projectRoot, outputName);
 
   beforeAll(async () => {
-    // NODE_ENV=production EXPO_USE_STATIC=static E2E_ROUTER_SRC=static-rendering E2E_ROUTER_ASYNC=production npx expo export -p web --source-maps --output-dir dist-static-splitting
+    // NODE_ENV=production EXPO_USE_STATIC=static E2E_ROUTER_SRC=static-rendering npx expo export -p web --source-maps --output-dir dist-static-splitting
```
jakub-agent, line 23: [💡 suggestion] Dropping `E2E_ROUTER_ASYNC: 'production'` makes a test called "exports static with bundle splitting" assert the default rather than the feature. Same in `server-rendering-async.test.ts` and the two playwright equivalents.

Can we keep the explicit value in the feature-specific tests and cover "the default is on" separately? Otherwise a future default change turns these into no-ops instead of failing.

docs/pages/versions/unversioned/sdk/router/index.mdx
```diff
@@ -75,8 +75,8 @@ If you are using the [default](/more/create-expo/#--template) template to create
     {
       name: 'asyncRoutes',
       description:
-        'Enable async routes (lazy loading). Can be a boolean, a string (`"development"` or `"production"`), or an object with platform-specific values (`{ android, ios, web, default }`). `production` is currently web-only and will be disabled on native.',
-      default: 'undefined',
+        'Enable experimental async routes (lazy loading). In SDK 58 and later, this is enabled by default on web in development and production, and disabled by default on native. Can be a boolean, a string (`"development"` or `"production"`), or an object with platform-specific values (`{ android, ios, web, default }`). Explicit platform values take precedence over `default`. Production async routes are web-only.',
```
jakub-agent, line 78: [💡 suggestion] This still opens with "Enable experimental async routes" and `async-routes.mdx` still carries the alpha callout, but it's now the default for every web app. Should the release status move as part of this stack, or should the callout say what specifically is still unstable?

The paragraph in `async-routes.mdx` also packs four precedence rules into one block — can it be bullets?

packages/expo-router/plugin/src/withRouter.ts
```diff
@@ -1,5 +1,5 @@
 import { validate } from '@expo/schema-utils';
-import { ConfigPlugin, withInfoPlist, withPodfile } from 'expo/config-plugins';
+import { type ConfigPlugin, withInfoPlist, withPodfile } from 'expo/config-plugins';
```
jakub-agent, line 2: [🐧 nit] nit: this `type` change isn't related to async routes. Can you revert it to keep the diff minimal?

jakub-agent (dismissed):

PR 6/6 in the SDK 58 router stabilization stack. The plugin-side normalization is small and the new `withRouter` tests cover the precedence rules well.

Requesting changes for two reasons rather than the code itself:

1. `playwright-ubuntu (3)` is red, and it's this change — `stable-module-ids.test.ts:53` fails on `__r("packages/expo-router/entry.js")` because `find(f => f.endsWith('.js'))` now picks an async chunk. It's green on the base branch.
2. Turning async routes on flips `useScreens.tsx` into lazy mode, so a custom `SuspenseFallback` exported from a layout is silently ignored on web. That deserves a `🛠 Breaking changes` entry naming the escape hatch.

Not inline: the default is applied in the plugin, so a project that uses Expo Router without an `expo-router` entry in `plugins` keeps async routes off — `getRouterDirectoryModuleIdWithManifest` carries a fallback for exactly that case, so it does happen. Is the plugin placement the intended trade-off?

docs/pages/router/web/async-routes.mdx
```diff
@@ -30,30 +30,21 @@ For those familiar with advanced bundling techniques, the async routes feature i
...
+
+```json app.json
+{
+  "expo": {
+    "plugins": [["expo-router", { "asyncRoutes": false }]]
+  }
+}
+```
+
+You can also use `"development"` or `"production"` to enable async routes only in that mode, or an object with platform-specific settings (`default`, `android`, `ios`, or `web`). An explicit platform value takes precedence over `default`. Setting `{ "default": false }` disables async routes on web unless `web` explicitly enables them. A configuration containing only native platform settings retains the web default in SDK 58 and later.
```
amandeepmittal, line 45: ✏️ **style** `prose.md#core-principles`

The behavior most likely to catch a reader out, that a native-only configuration still keeps the web default, is stated abstractly at the end of four rules. Give it a concrete value.

```suggestion
You can also use `"development"` or `"production"` to enable async routes only in that mode, or an object with platform-specific settings (`default`, `android`, `ios`, or `web`). An explicit platform value takes precedence over `default`. Setting `{ "default": false }` disables async routes on web unless `web` explicitly enables them. In SDK 58 and later, a configuration that sets only native platforms keeps the web default, so `{ "android": true }` behaves as `{ "android": true, "web": true }`.
```

docs/pages/router/web/async-routes.mdx
```diff
@@ -30,30 +30,21 @@ For those familiar with advanced bundling techniques, the async routes feature i
...
-              }
-            }
-          ]
-        ]
-      }
-    }
-    ```
-
-You can set platform-specific settings (`default`, `android`, `ios` or `web`) for `asyncRoutes` using an object:
+In SDK 58 and later, async routes are enabled by default on web in development and production. Native platforms default to disabled. In SDK 57 and earlier, async routes require an explicit opt-in on every platform.
```
amandeepmittal, line 33: ⚠️ **correctness** `terminology.md#expo-product-and-feature-names`

The callout at the top of this page still says async routes is in [alpha](/more/release-statuses/#alpha), and that status says the feature is not recommended for production apps. This line now says it is on by default in production on web. Update the status callout in this PR, or narrow this sentence so the two agree.

docs/pages/versions/unversioned/sdk/router/index.mdx
amandeepmittal, line 78: ✏️ **style** `terminology.md#expo-product-and-feature-names`

The async routes guide calls this feature alpha, and this description calls it experimental. Pick one status, and if it is experimental, set `experimental: true` on this entry the way the three other experimental properties in this table do instead of putting the word in the description.

amandeepmittal (approved):

docs/pages/router/web/async-routes.mdx
```diff
@@ -6,7 +6,7 @@ description: Learn how to speed up development with async bundling in Expo Route
 import { ContentSpotlight } from '~/ui/components/ContentSpotlight';
 import { Terminal } from '~/ui/components/Snippet';
 
-> **important** Async routes is in [alpha](/more/release-statuses/#alpha).
+> **important** Async routes remains experimental. In SDK 58 and later, it is enabled by default on web in development and production. Native async routes remain opt-in and are supported only in development. See [Caveats](#caveats) for current limitations.
```
Ubax: That's a little werid

We enable the experimental feature by default 😅 

Is it experimental for all or only native?

docs/pages/router/web/async-routes.mdx
```diff
@@ -75,7 +70,7 @@ You can set platform-specific settings (`default`, `android`, `ios` or `web`) fo
 }
 ```
 
-Then, when you are about to start your project, you can use the `--clear` flag to clear the Metro cache. This will ensure that the routes are loaded asynchronously:
+After changing the setting, clear the Metro cache with `--clear` when starting or exporting your project:
```
Ubax, line 73: Can you add this to migration guide as well?

packages/expo-router/CHANGELOG.md
```diff
@@ -14,7 +16,7 @@
 
 ### 🛠 Breaking changes
 
-- `web.output: 'server'` now renders HTML pages on each request instead of prerendering them during export. ([#50120](https://github.com/expo/expo/pull/50120) by [@hassankhan](https://github.com/hassankhan))
+- `web.output: 'server'` now renders HTML pages on each request instead of prerendering them during export ([#50120](https://github.com/expo/expo/pull/50120) by [@hassankhan](https://github.com/hassankhan))
```
```suggestion
- `web.output: 'server'` now renders HTML pages on each request instead of prerendering them during export. ([#50120](https://github.com/expo/expo/pull/50120) by [@hassankhan](https://github.com/hassankhan))
```

😂

Ubax (approved):

docs/pages/router/web/async-routes.mdx
hassankhan: It's still experimental for native; definitely a bit confusing but there isn't a better way of doing this given one property represents the entire feature for all platforms 😅

PR #48328, opened 2026-07-30
[app-metrics][observe] Sync both packages from main

tsapeta:
# Why

`expo-app-metrics` and `expo-observe` drifted from `main` on `sdk-57`. Nine source PRs landed after the last bulk backport (2026-06-25) and were never brought over.

# How

Replaces both packages' `src/`, `ios/` and `android/` with `main`'s version in one commit, like the two earlier bulk backports. Cherry-picking would have replayed churn (#48245 adds `initIntegration`, #48268 renames it) and dragged in three repo-wide format sweeps.

Brings in: #47289, #47341, #47489, #47782, #47871, #48161, #48245, #48268, plus the parts of #47437, #47965 and #47980 that touch these packages.

Build-convention files stay at `sdk-57`'s values, since it predates the Turborepo/oxlint migration. Two exceptions, both needed to keep the branch green:

- `jest.config.js` added to `expo-app-metrics` — the synced tests don't run without it, and it's identical to the one `expo-observe` already uses here.
- Import order reverted in `ObserveNavigationContainer.tsx` and `ObserveNavigationProvider.tsx`, which `sdk-57`'s eslint wants the other way. Formatting only.

The app-metrics half of #47902 is left out: `sdk-57`'s `expo-updates` still emits `downloadComplete` and the `expo-updates` half of that rename was never backported, so renaming only the app-metrics enum would leave it matching an event nothing emits.

# Test Plan

- `expo-app-metrics`: 30 tests, 4 suites (all new — the package had none here before)
- `expo-observe`: 554 tests, 66 suites
- Both lint clean and build without errors


apps/observe-tester/plugins/withUISceneLifecycle.js
```diff
@@ -0,0 +1,210 @@
+// Adopts the UIKit scene-based life cycle, which the iOS 26+ SDK requires: apps
```
Ubax: [🏗️ **design**] This plugin (and its `app.json` registration) is not part of the sync: it has no history on `main` or any other branch, so it's new code original to this PR.

Is it added, so that you can build on XCode 27?

packages/expo-app-metrics/ios/Updates/UpdatesStateEvent.swift
```diff
@@ -9,7 +9,7 @@ struct UpdatesStateEvent {
     case checkCompleteWithRollback = "checkCompleteWithRollback"
     case checkError = "checkError"
     case download = "download"
-    case downloadComplete = "downloadComplete"
+    case downloadCompleteUnavailable = "downloadCompleteUnavailable"
```
Ubax: [💡 suggestion] sdk-57's `expo-updates` still emits `"downloadComplete"` — the string `downloadCompleteUnavailable` doesn't exist anywhere in `packages/expo-updates` on this branch (the expo-updates side of #47902 wasn't backported). After this rename the old event parses to `nil` in `fromDict`, same as before when it parsed to a case nobody consumed, and the only consumer checks `.downloadCompleteWithUpdate` — so behavior is unchanged and this is safe to land.

Ubax: Let's revert the changes made in https://github.com/expo/expo/pull/47902

apps/observe-tester/plugins/withUISceneLifecycle.js
tsapeta: Yes, otherwise it crashes at launch. Looks like my prebuilt `ios` folder on sdk-57 worktree never had these necessary changes so Claude made a config plugin 👀 I'll open another PR with this plugin to main and then pick it later.

Ubax (approved):

PR #48330, opened 2026-07-30
[router] Make js-tabs and js-top-tabs usable in RSC and add RSC tests

Ubax:
# Why

JS Tabs and Top Tab are currently not usable in RSC

# How

1. Evaluate client side code for animation easing lazily
2. Fix exports
3. Update RSC tests

# Test Plan

CI

# Checklist

### DIFF
--- a/packages/expo-router/src/layouts/Tabs.tsx
+++ b/packages/expo-router/src/layouts/Tabs.tsx
@@ -1,9 +1,32 @@
+import { Protected } from '../views/Protected';
 import { Screen } from '../views/Screen';
 import Tabs from './TabsClient';
 
-export * from '../react-navigation/bottom-tabs';
+export {
+  createBottomTabNavigator,
+  BottomTabBar,
+  BottomTabView,
+  BottomTabBarHeightCallbackContext,
+  BottomTabBarHeightContext,
+  useBottomTabBarHeight,
+} from './TabsClient';
+export * as SceneStyleInterpolators from '../react-navigation/bottom-tabs/TransitionConfigs/SceneStyleInterpolators';
+export * as TransitionPresets from '../react-navigation/bottom-tabs/TransitionConfigs/TransitionPresets';
+export * as TransitionSpecs from '../react-navigation/bottom-tabs/TransitionConfigs/TransitionSpecs';
+export type {
+  BottomTabBarButtonProps,
+  BottomTabBarProps,
+  BottomTabHeaderProps,
+  BottomTabNavigationEventMap,
+  BottomTabNavigationOptions,
+  BottomTabNavigationProp,
+  BottomTabNavigatorProps,
+  BottomTabOptionsArgs,
+  BottomTabScreenProps,
+} from '../react-navigation/bottom-tabs';
 
 Tabs.Screen = Screen;
+Tabs.Protected = Protected;
 
 export { Tabs };
 
--- a/packages/expo-router/src/layouts/__rsc_tests__/Tabs.test.tsx
+++ b/packages/expo-router/src/layouts/__rsc_tests__/Tabs.test.tsx
@@ -1,0 +1,52 @@
+/// <reference types="jest-expo/rsc/expect" />
+
+import Tabs, {
+  BottomTabBar,
+  BottomTabBarHeightCallbackContext,
+  BottomTabBarHeightContext,
+  BottomTabView,
+  SceneStyleInterpolators,
+  TransitionPresets,
+  TransitionSpecs,
+  createBottomTabNavigator,
+  useBottomTabBarHeight,
+} from '../Tabs';
+
+function expectClientReference(value: unknown) {
+  expect((value as { $$typeof?: symbol }).$$typeof).toBe(Symbol.for('react.client.reference'));
+}
+
+it('resolves React Navigation exports as client references', () => {
+  expectClientReference(createBottomTabNavigator);
+  expectClientReference(BottomTabBar);
+  expectClientReference(BottomTabView);
+  expectClientReference(BottomTabBarHeightCallbackContext);
+  expectClientReference(BottomTabBarHeightContext);
+  expectClientReference(useBottomTabBarHeight);
+});
+
+it('resolves transition namespaces on the server', () => {
+  expect(TransitionPresets.ShiftTransition).toBeDefined();
+  const fadeConfig = TransitionSpecs.FadeSpec.config as { easing?: unknown };
+  expect(fadeConfig).toMatchObject({ duration: 150 });
+  if (process.env.EXPO_OS === 'web') {
+    // react-native-web's `Easing` works on the server.
+    expect(fadeConfig.easing).toBeInstanceOf(Function);
+  } else {
+    // On native, `easing` still needs the client — `Easing` is only resolved when read.
+    expect(() => fadeConfig.easing).toThrow();
+  }
+  expect(SceneStyleInterpolators.forFade).toBeInstanceOf(Function);
+});
+
+it(`renders Tabs`, async () => {
+  await expect(<Tabs />).toMatchFlightSnapshot();
+});
+
+it(`renders Tabs.Screen`, async () => {
+  await expect(<Tabs.Screen options={{ title: '...' }} />).toMatchFlightSnapshot();
+});
+
+it(`renders Tabs.Protected`, async () => {
+  await expect(<Tabs.Protected guard={false} />).toMatchFlightSnapshot();
+});
--- a/packages/expo-router/src/layouts/__rsc_tests__/TopTabs.test.tsx
+++ b/packages/expo-router/src/layouts/__rsc_tests__/TopTabs.test.tsx
@@ -1,0 +1,33 @@
+/// <reference types="jest-expo/rsc/expect" />
+
+import TopTabs, {
+  MaterialTopTabBar,
+  MaterialTopTabView,
+  createMaterialTopTabNavigator,
+  useTabAnimation,
+} from '../TopTabs';
+
+// The module proxy answers any property name, so `toBeDefined()` would pass even for a dropped
+// export. Assert the client-reference marker to pin that these crossed a client boundary.
+function expectClientReference(value: unknown) {
+  expect((value as { $$typeof?: symbol }).$$typeof).toBe(Symbol.for('react.client.reference'));
+}
+
+it('resolves React Navigation exports as client references', () => {
+  expectClientReference(createMaterialTopTabNavigator);
+  expectClientReference(MaterialTopTabBar);
+  expectClientReference(MaterialTopTabView);
+  expectClientReference(useTabAnimation);
+});
+
+it(`renders TopTabs`, async () => {
+  await expect(<TopTabs />).toMatchFlightSnapshot();
+});
+
+it(`renders TopTabs.Screen`, async () => {
+  await expect(<TopTabs.Screen options={{ title: '...' }} />).toMatchFlightSnapshot();
+});
+
+it(`renders TopTabs.Protected`, async () => {
+  await expect(<TopTabs.Protected guard={false} />).toMatchFlightSnapshot();
+});
--- a/packages/expo-router/src/layouts/__tests__/TopTabs-exports.test.tsx
+++ b/packages/expo-router/src/layouts/__tests__/TopTabs-exports.test.tsx
@@ -1,0 +1,16 @@
+import * as vendored from '../../react-navigation/material-top-tabs';
+import * as entry from '../TopTabs';
+
+describe('expo-router/js-top-tabs re-exports', () => {
+  it('re-exports every value from ../react-navigation/material-top-tabs', () => {
+    const missing = Object.keys(vendored).filter((key) => !(key in entry));
+    expect(missing).toEqual([]);
+  });
+
+  it('exports the TopTabs navigator with its static components', () => {
+    expect(entry.TopTabs).toBeDefined();
+    expect(entry.default).toBe(entry.TopTabs);
+    expect(entry.TopTabs.Screen).toBeDefined();
+    expect(entry.TopTabs.Protected).toBeDefined();
+  });
+});
--- a/packages/expo-router/src/react-navigation/bottom-tabs/TransitionConfigs/TransitionSpecs.tsx
+++ b/packages/expo-router/src/react-navigation/bottom-tabs/TransitionConfigs/TransitionSpecs.tsx
@@ -6,7 +6,11 @@ export const FadeSpec: TransitionSpec = {
   animation: 'timing',
   config: {
     duration: 150,
-    easing: Easing.in(Easing.linear),
+    // Resolve lazily so this module evaluates in React Server Components.
+    // Reading `easing` still needs the client on native — `Easing` throws on the server there.
+    get easing() {
+      return Easing.in(Easing.linear);
+    },
   },
 };
 
@@ -14,6 +18,10 @@ export const ShiftSpec: TransitionSpec = {
   animation: 'timing',
   config: {
     duration: 150,
-    easing: Easing.inOut(Easing.ease),
+    // Resolve lazily so this module evaluates in React Server Components.
+    // Reading `easing` still needs the client on native — `Easing` throws on the server there.
+    get easing() {
+      return Easing.inOut(Easing.ease);
+    },
   },
 };
--- a/packages/expo-router/src/react-navigation/material-top-tabs/__tests__/missing-dependency.test.ios.tsx
+++ b/packages/expo-router/src/react-navigation/material-top-tabs/__tests__/missing-dependency.test.ios.tsx
@@ -9,6 +9,7 @@ jest.mock(
 );
 
 test('throws an error when react-native-tab-view is not installed', () => {
+  jest.resetModules();
   jest.doMock('react-native-tab-view', () => {
     throw new Error();
   });
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -21,6 +21,7 @@
 
 ### 🐛 Bug fixes
 
+- Fixed `Tabs` and `TopTabs` (`expo-router/js-tabs`, `expo-router/js-top-tabs`) not being usable from RSC ([#48330](https://github.com/expo/expo/pull/48330) by [@Ubax](https://github.com/Ubax))
 - Redirect fully guarded navigators to parent ([#47984](https://github.com/expo/expo/pull/47984) by [@Ubax](https://github.com/Ubax))
 - Unset `nativeContainerStyle` in transparent presentations ([#48154](https://github.com/expo/expo/pull/48154) by [@Ubax](https://github.com/Ubax))
 - [android] Disable safe area insets in Native Tabs on Android when tab bar is hidden. ([#47611](https://github.com/expo/expo/pull/47611) by [@debitan](https://github.com/debitan))
--- a/packages/expo-router/src/layouts/TabsClient.tsx
+++ b/packages/expo-router/src/layouts/TabsClient.tsx
@@ -14,6 +14,9 @@ import type { Href } from '../types';
 import { Protected } from '../views/Protected';
 import { withLayoutContext } from './withLayoutContext';
 
+// Keep React Navigation client-only so the entry evaluates in React Server Components.
+export * from '../react-navigation/bottom-tabs';
+
 // This is the only way to access the navigator.
 const BottomTabNavigator = createBottomTabNavigator().Navigator;
 
--- a/packages/expo-router/src/layouts/TopTabs.tsx
+++ b/packages/expo-router/src/layouts/TopTabs.tsx
@@ -1,10 +1,26 @@
+import { Protected } from '../views/Protected';
 import { Screen } from '../views/Screen';
 import TopTabs from './TopTabsClient';
 
 TopTabs.Screen = Screen;
+TopTabs.Protected = Protected;
 
 export { TopTabs };
 
-export * from '../react-navigation/material-top-tabs';
+export {
+  createMaterialTopTabNavigator,
+  MaterialTopTabBar,
+  MaterialTopTabView,
+  useTabAnimation,
+} from './TopTabsClient';
+export type {
+  MaterialTopTabBarProps,
+  MaterialTopTabNavigationEventMap,
+  MaterialTopTabNavigationOptions,
+  MaterialTopTabNavigationProp,
+  MaterialTopTabNavigatorProps,
+  MaterialTopTabOptionsArgs,
+  MaterialTopTabScreenProps,
+} from '../react-navigation/material-top-tabs';
 
 export default TopTabs;
--- a/packages/expo-router/src/layouts/TopTabsClient.tsx
+++ b/packages/expo-router/src/layouts/TopTabsClient.tsx
@@ -12,6 +12,9 @@ import { Protected } from '../views/Protected';
 import { Screen } from '../views/Screen';
 import { withLayoutContext } from './withLayoutContext';
 
+// Keep React Navigation client-only so the entry evaluates in React Server Components.
+export * from '../react-navigation/material-top-tabs';
+
 const MaterialTopTabNavigator = createMaterialTopTabNavigator().Navigator;
 
 const MaterialTopTabs = withLayoutContext<
--- a/packages/expo-router/src/layouts/__rsc_tests__/__snapshots__/Tabs.test.tsx.snap.android
+++ b/packages/expo-router/src/layouts/__rsc_tests__/__snapshots__/Tabs.test.tsx.snap.android
@@ -1,16 +1,16 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`renders TopTabs 1`] = `
-"1:I["src/layouts/TopTabsClient.tsx",[],"",1]
+exports[`renders Tabs 1`] = `
+"1:I["src/layouts/TabsClient.tsx",[],"",1]
 0:["$","$L1",null,{},null]"
 `;
 
-exports[`renders TopTabs.Protected 1`] = `
+exports[`renders Tabs.Protected 1`] = `
 "1:I["src/primitives/navigation.tsx",[],"Group",1]
 0:["$","$L1",null,{"guard":false},null]"
 `;
 
-exports[`renders TopTabs.Screen 1`] = `
+exports[`renders Tabs.Screen 1`] = `
 "1:I["src/views/Screen.tsx",[],"Screen",1]
 0:["$","$L1",null,{"options":{"title":"..."}},null]"
 `;
--- a/packages/expo-router/src/layouts/__rsc_tests__/__snapshots__/Tabs.test.tsx.snap.ios
+++ b/packages/expo-router/src/layouts/__rsc_tests__/__snapshots__/Tabs.test.tsx.snap.ios
@@ -1,16 +1,16 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`renders TopTabs 1`] = `
-"1:I["src/layouts/TopTabsClient.tsx",[],"",1]
+exports[`renders Tabs 1`] = `
+"1:I["src/layouts/TabsClient.tsx",[],"",1]
 0:["$","$L1",null,{},null]"
 `;
 
-exports[`renders TopTabs.Protected 1`] = `
+exports[`renders Tabs.Protected 1`] = `
 "1:I["src/primitives/navigation.tsx",[],"Group",1]
 0:["$","$L1",null,{"guard":false},null]"
 `;
 
-exports[`renders TopTabs.Screen 1`] = `
+exports[`renders Tabs.Screen 1`] = `
 "1:I["src/views/Screen.tsx",[],"Screen",1]
 0:["$","$L1",null,{"options":{"title":"..."}},null]"
 `;
--- a/packages/expo-router/src/layouts/__rsc_tests__/__snapshots__/Tabs.test.tsx.snap.web
+++ b/packages/expo-router/src/layouts/__rsc_tests__/__snapshots__/Tabs.test.tsx.snap.web
@@ -1,16 +1,16 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`renders TopTabs 1`] = `
-"1:I["src/layouts/TopTabsClient.tsx",[],"",1]
+exports[`renders Tabs 1`] = `
+"1:I["src/layouts/TabsClient.tsx",[],"",1]
 0:["$","$L1",null,{},null]"
 `;
 
-exports[`renders TopTabs.Protected 1`] = `
+exports[`renders Tabs.Protected 1`] = `
 "1:I["src/primitives/navigation.tsx",[],"Group",1]
 0:["$","$L1",null,{"guard":false},null]"
 `;
 
-exports[`renders TopTabs.Screen 1`] = `
+exports[`renders Tabs.Screen 1`] = `
 "1:I["src/views/Screen.tsx",[],"Screen",1]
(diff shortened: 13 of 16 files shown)


Ubax:
* **#48330** <a href="https://app.graphite.com/github/pr/expo/expo/48330?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/48330?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>


<h2></h2>

Ubax:
I'm intentionally not adding CHANGELOG here, since this only improves coverage

packages/expo-router/src/layouts/__rsc_tests__/Tabs.test.tsx
```diff
@@ -0,0 +1,13 @@
+/// <reference types="jest-expo/rsc/expect" />
+
+// Imported from `TabsClient`, not the `Tabs` entry point. On native the entry cannot be evaluated
+// on the server at all: it re-exports `../react-navigation/bottom-tabs`, whose `TransitionSpecs`
+// module calls `Easing.in(...)` at module scope. That is a pre-existing limitation of the entry
+// point, not of this navigator.
+import Tabs from '../TabsClient';
+
+it(`renders to RSC`, async () => {
+  const jsx = <Tabs.Screen options={{ title: '...' }} />;
```
jakub-agent: [🏗️ **design**] This assertion is vacuous — `Tabs.Screen` is `undefined` here, so the snapshot records `"$undefined"` as the element type rather than a client reference.

`TabsClient.tsx` starts with `'use client'`, so `babel-preset-expo`'s `client-module-proxy-plugin` replaces the whole module body in the react-server environment. The default export becomes a flat client reference produced by `registerClientReference` — a plain function with `$$typeof`/`$$id`/`$$async` defined on it, not a deep proxy — so property access such as `.Screen` yields `undefined`. The rendered JSX is effectively `<undefined options={{ title: '...' }} />`.

Compare a snapshot where the entry module is server-evaluable and `Screen` resolves:

```
// Stack.test.tsx.snap.ios
1:I["src/layouts/stack-utils/StackScreen.tsx",[],"StackScreen",1]
0:["$","$L1",null,{"options":{"title":"..."}},null]
```

So this test cannot fail for the reason the PR description gives: if the standard-navigation migration breaks `Tabs` in RSC, the snapshot stays `$undefined` and CI stays green. `JSStack.test.tsx` on `main` has the same problem (byte-identical snapshot) — worth fixing together, since it is presumably the template for this file.

packages/expo-router/src/layouts/__rsc_tests__/Tabs.test.tsx
```diff
@@ -0,0 +1,13 @@
+/// <reference types="jest-expo/rsc/expect" />
+
+// Imported from `TabsClient`, not the `Tabs` entry point. On native the entry cannot be evaluated
+// on the server at all: it re-exports `../react-navigation/bottom-tabs`, whose `TransitionSpecs`
+// module calls `Easing.in(...)` at module scope. That is a pre-existing limitation of the entry
```
jakub-agent: [🏗️ **design**] The mechanism is right — `react-navigation/bottom-tabs/TransitionConfigs/TransitionSpecs.tsx` does call `Easing.in(Easing.linear)` at module scope — but "pre-existing limitation of the entry point" reads as though this were inherent. Git history says it is a regression with a specific cause:

`1733bd7ccc9` — *[expo-router] add missing exports to js-tabs and js-top-tabs (#46117)* — added `export * from '../react-navigation/bottom-tabs'` to `Tabs.tsx` and `export * from '../react-navigation/material-top-tabs'` to `TopTabs.tsx`, and in the same commit deleted `__rsc_tests__/TopTabs.test.tsx`:

```diff
-import TopTabs from '../TopTabs';
-
-it(`renders to RSC`, async () => {
-  const jsx = <TopTabs.Screen options={{ title: '...' }} />;
-
-  await expect(jsx).toMatchFlightSnapshot();
-});
```

That is the same test this file adds, except it imported the entry. Its three snapshots were never deleted and are still on `main`, showing what used to resolve:

```
// __snapshots__/TopTabs.test.tsx.snap.ios
1:I["src/views/Screen.tsx",[],"Screen",1]
0:["$","$L1",null,{"options":{"title":"..."}},null]
```

This matters beyond the test: `layouts/Tabs.tsx` is what `expo-router/js-tabs` and `expo-router/tabs` resolve to, and it is the source of the deprecated `Tabs` re-export in `exports.ts`. If it cannot be evaluated on the server, no user can render `Tabs` from a server component — the exact property this PR sets out to protect.

`Stack.tsx` shows the working pattern: no `export *`, plus the explicit note about importing `StackToolbar` from a separate file "for rsc to work correctly". Moving these re-exports off the entry (own subpath, or explicit named re-exports that skip `TransitionSpecs`) would let this test import `../Tabs`, let `TopTabs.test.tsx` come back, and let the orphaned snapshots be cleaned up.

If that is out of scope here, please replace this comment with a linked TODO so the gap is tracked rather than normalized by a green test.

packages/expo-router/src/layouts/__rsc_tests__/Tabs.test.tsx
```diff
@@ -0,0 +1,13 @@
...
+// point, not of this navigator.
+import Tabs from '../TabsClient';
+
+it(`renders to RSC`, async () => {
+  const jsx = <Tabs.Screen options={{ title: '...' }} />;
+
+  await expect(jsx).toMatchFlightSnapshot();
```
jakub-agent: [💡 suggestion] Coverage is thin for the stated goal. `native-tabs/__rsc_tests__/tabs.test.tsx` covers each exported component (`NativeTabs`, `NativeTabs.Trigger`, the `Icon` variants, `Badge`, `Label`, `NativeTabsBottomAccessory`); here there is one case, and it is the one that does not resolve.

Even while importing `TabsClient`, `await expect(<Tabs />).toMatchFlightSnapshot()` would assert something real: the default export *is* a valid client reference, so the snapshot pins the client-boundary module key and proves the boundary exists. Once the entry is server-evaluable, `Tabs.Screen` and `Tabs.Protected` become testable too — both are public surface that the standard-navigation migration touches.

jakub-agent (changes requested):

Test-only PR (4 files, +22), PR 2/2 on top of #48291. The intent is right — `Tabs` is the JS navigator without RSC coverage — but as written the test does not exercise `Tabs`, and the workaround it uses hides a real regression.

**1. The snapshot asserts nothing.** The committed value is:

```
exports[`renders to RSC 1`] = `"0:["$","$undefined",null,{"options":{"title":"..."}},null]"`;
```

`$undefined` is the element *type*. `TabsClient.tsx` starts with `'use client'`, so `babel-preset-expo`'s `client-module-proxy-plugin` replaces the module body in the react-server environment and the default import becomes a flat client reference from `registerClientReference` (a plain function carrying `$$typeof`/`$$id`/`$$async`, not a deep proxy). `Tabs.Screen` on it is `undefined`, so the test renders `<undefined options={...} />`. Contrast `Stack.test.tsx.snap.ios`, which records a real boundary row: `1:I["src/layouts/stack-utils/StackScreen.tsx",[],"StackScreen",1]`. If the standard-navigation migration breaks `Tabs` in RSC, this snapshot stays `$undefined` and CI stays green. `JSStack.test.tsx` on `main` has the byte-identical vacuous snapshot — presumably the template for this file.

**2. The limitation the comment describes is a regression from #46117, not an inherent one.** That commit added `export * from '../react-navigation/bottom-tabs'` to `Tabs.tsx` and `export * from '../react-navigation/material-top-tabs'` to `TopTabs.tsx`, and in the same commit **deleted** `__rsc_tests__/TopTabs.test.tsx` — a test whose body is character-for-character the test this PR is adding, except it imported the entry (`../TopTabs`). Its three snapshot files were never removed and are still orphaned on `main`, recording the client reference that used to resolve:

```
// __snapshots__/TopTabs.test.tsx.snap.ios — orphaned on main since #46117
exports[`renders to RSC 1`] = `
"1:I["src/views/Screen.tsx",[],"Screen",1]
0:["$","$L1",null,{"options":{"title":"..."}},null]"
`;
```

So RSC support for `expo-router/js-tabs` and `expo-router/js-top-tabs` was lost in #46117, the coverage that proved it was dropped rather than the break fixed, and this PR now adds coverage that routes around the same break. `Stack.tsx` shows the pattern that keeps working: no `export *`, plus the note about splitting `StackToolbar` out "for rsc to work correctly".

**Suggested scope:** move the `export *` re-exports off `layouts/Tabs.tsx` / `layouts/TopTabs.tsx` (own subpath, or explicit named re-exports that avoid `TransitionSpecs`), then point this test at `../Tabs`, restore `TopTabs.test.tsx`, and delete the orphaned snapshots. That yields the guarantee the PR description asks for. If that is too large for this stack, this test still should not land asserting `$undefined` — at minimum assert something real (see inline comments) and file a linked issue.

packages/expo-router/src/layouts/Tabs.tsx
```diff
@@ -1,9 +1,32 @@
+import { Protected } from '../views/Protected';
 import { Screen } from '../views/Screen';
 import Tabs from './TabsClient';
 
-export * from '../react-navigation/bottom-tabs';
+export {
+  SceneStyleInterpolators,
+  TransitionPresets,
```
jakub-agent: [🏗️ **design**] These three are namespace objects (`import * as` in the vendored index), and routing them through a `'use client'` module means member access on them silently yields `undefined` in a server component — the same mechanism as the `$undefined` snapshot from the last round.

In the react-server graph `TabsReactNavigationClient.tsx` becomes a bare client-module proxy: `client-module-proxy-plugin` emits nothing for `ExportAllDeclaration` ("Do nothing because we have the top-level hack to inject `module.exports`"), so the module is only `createClientModuleProxy(key)`. Each name pulled off it is a flat reference from `registerClientReferenceImpl` — a function with `$$typeof`/`$$id`/`$$async` and nothing else. So in a server `_layout.tsx`:

```tsx
import Tabs, { TransitionPresets } from 'expo-router/js-tabs';

export default function Layout() {
  // TransitionPresets.ShiftTransition is undefined -> spreads to {} -> no transition, no error
  return <Tabs screenOptions={{ ...TransitionPresets.ShiftTransition }} />;
}
```

This is worse than what it replaces, not better: before the split the import crashed at module evaluation, which is loud. And the honest answer here is "unsupported" either way — `ShiftTransition` holds `sceneStyleInterpolator: forShift`, a function, so even if it did resolve, RSC would reject it at the boundary with "Functions cannot be passed directly to Client Components". Silently degrading to `{}` is the one outcome that gives no signal at all.

`SceneStyleInterpolators` is the easy case — pure functions, no `react-native` import, no module-scope side effects — so it can be re-exported from the server-evaluable path as-is. `TransitionPresets` and `TransitionSpecs` are entangled with the module-scope `Easing.in(Easing.linear)` in `TransitionSpecs.tsx`. If that call is in fact the only blocker, making the easing lazy (a getter, or computing it inside `TransitionSpecs`) would let all three be re-exported directly and keep this footgun from existing. I could not confirm from static reading that `Easing` is the *only* server-hostile thing in that graph, so treat that as a suggestion to check rather than a verified fix.

At minimum, worth a comment in `TabsReactNavigationClient.tsx` recording that these namespaces are client-only and that member access from a server component yields `undefined`.

packages/expo-router/src/layouts/Tabs.tsx
```diff
@@ -1,9 +1,32 @@
+import { Protected } from '../views/Protected';
 import { Screen } from '../views/Screen';
 import Tabs from './TabsClient';
 
-export * from '../react-navigation/bottom-tabs';
+export {
```
jakub-agent, line 5: [💡 suggestion] I diffed this list against `react-navigation/bottom-tabs/index.tsx` at this SHA and it is complete (9 values, 9 types), same for the material-top-tabs list in `TopTabs.tsx` (4 values, 7 types). So nothing is dropped today.

The cost is that `export *` used to propagate automatically and now it does not. The next export added to the vendored index will silently not appear on `expo-router/js-tabs`, and nothing fails — no type error, no test. That is exactly the class of drift #46117 was written to fix.

A cheap guard: a plain (non-RSC) unit test asserting parity, e.g. `expect(Object.keys(require('../Tabs')).sort())` against the re-exported names of `../react-navigation/bottom-tabs`. It runs in the normal jest project where the module evaluates fine, and it fails the moment the lists diverge.

packages/expo-router/src/layouts/Tabs.tsx
```diff
@@ -1,9 +1,32 @@
...
+  BottomTabNavigationEventMap,
+  BottomTabNavigationOptions,
+  BottomTabNavigationProp,
+  BottomTabNavigatorProps,
+  BottomTabOptionsArgs,
+  BottomTabScreenProps,
+} from '../react-navigation/bottom-tabs';
 
 Tabs.Screen = Screen;
+Tabs.Protected = Protected;
```
jakub-agent, line 29: [💡 suggestion] Worth revisiting the changelog decision. The earlier comment on this PR — "I'm intentionally not adding CHANGELOG here, since this only improves coverage" — was accurate for the original test-only diff, but no longer describes this one: `expo-router/js-tabs` and `expo-router/js-top-tabs` gain two new internal modules, their export lists change shape, and `Tabs` / `TopTabs` become usable from a server component for the first time since #46117. That last part is a user-visible fix, which is the normal bar for an entry.

(To be precise about this line: `Tabs.Protected` is not new public API — `TabsClient`'s `Object.assign` already provided it for client apps. What changes is that it now also resolves in the server graph, because the assignment happens in the server-evaluable entry rather than only inside the `'use client'` module.)

The PR title and description are also stale for the same reason — "Add tabs rsc tests" / "Add tests" undersells a change to the public export structure, and reviewers of the stack will read the title first.

jakub-agent (commented):

**Iteration 2** (`38e73ca`). All three findings from the previous review are addressed, and addressed at the root rather than worked around:

- `Tabs.test.tsx` now imports `../Tabs` (the entry), and every snapshot records a real client-reference row. The `$undefined` element type is gone: `1:I["src/views/Screen.tsx",[],"Screen",1]`, `1:I["src/layouts/TabsClient.tsx",[],"",1]`, `1:I["src/primitives/navigation.tsx",[],"Group",1]`.
- The `export *` was moved into new `'use client'` modules (`TabsReactNavigationClient.tsx`, `TopTabsReactNavigationClient.tsx`), so the entries are server-evaluable again. This mirrors the `Stack.tsx` / `StackToolbar` split.
- `TopTabs.test.tsx` is restored and the snapshots orphaned since #46117 are adopted rather than left dangling. `Tabs.Protected` / `TopTabs.Protected` are wired through the entry too, so they resolve in the server graph.

I checked the explicit re-export lists against the vendored index files at this SHA and they are **complete** — bottom-tabs 9 values + 9 types, material-top-tabs 4 values + 7 types, all present. No export was dropped.

The scope has changed though: this is no longer a test-only PR, it restructures the public export surface of `expo-router/js-tabs` and `expo-router/js-top-tabs`. Remaining points are non-blocking; the one worth a decision is the namespace re-exports (`TransitionPresets` and friends), which now resolve to `undefined` members in a server component instead of failing loudly.

Note for other reviewers: `gh pr diff` on this PR also shows the `BottomTabView.tsx` dependency-array reorder, but that belongs to the downstack commit `b29d3e7` (#48291), not to this one. This commit touches 12 files.

packages/expo-router/src/layouts/__rsc_tests__/Tabs.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+  BottomTabView,
+  SceneStyleInterpolators,
+  TransitionPresets,
+  TransitionSpecs,
+  createBottomTabNavigator,
+  useBottomTabBarHeight,
+} from '../Tabs';
+
+it('resolves React Navigation exports as client references', () => {
+  expect(createBottomTabNavigator).toBeDefined();
```
jakub-agent: [🏗️ **design**] This block cannot fail, so it does not test what its name claims.

In the react-server graph `./TabsClient` compiles to nothing but `createClientModuleProxy(key)` — `client-module-proxy-plugin` emits no per-name proxy for `ExportAllDeclaration`, it relies on the module-level proxy. React's proxy `get` handler mints a fresh client reference for *any* property name it is asked for. `Tabs.tsx` re-exports these six names from that proxy, so each one is defined regardless of whether it exists anywhere downstream. Delete `createBottomTabNavigator` from the vendored index and these six assertions still pass; misspell one in the import list and it passes too.

The real guard against that is the new `__tests__/Tabs-exports.test.tsx` parity check, which runs in the normal jest project where the vendored module actually evaluates. So this block is either redundant or should assert something a proxy cannot fabricate — the client-reference metadata:

```tsx
expect(createBottomTabNavigator).toEqual(
  expect.objectContaining({ $$typeof: Symbol.for('react.client.reference') })
);
```

That at least pins "this crossed a client boundary" rather than "a proxy answered". Same applies to the equivalent block in `TopTabs.test.tsx`.

packages/expo-router/src/layouts/__rsc_tests__/Tabs.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+  expect(BottomTabBar).toBeDefined();
+  expect(BottomTabView).toBeDefined();
+  expect(BottomTabBarHeightCallbackContext).toBeDefined();
+  expect(BottomTabBarHeightContext).toBeDefined();
+  expect(useBottomTabBarHeight).toBeDefined();
+});
+
+it('resolves transition namespaces on the server', () => {
+  expect(TransitionPresets.ShiftTransition).toBeDefined();
+  expect(TransitionSpecs.FadeSpec.config).toMatchObject({ duration: 150 });
```
jakub-agent: [🏗️ **design**] Asserting only `duration` here is load-bearing in a way the test does not say out loud, and it is worth making explicit.

`toMatchObject` only walks the keys present in the expected object, so this never reads `config.easing` — which is the whole point, because the getter still calls `Easing.in(Easing.linear)`. The fix moved that call off module scope; it did not make it work on the server. So the state after this PR is: `TransitionSpecs` *evaluates* in a server component, but touching `easing` from one still fails. `Tabs-exports.test.tsx` covers the getter (`easing: expect.any(Function)`) precisely because it runs in the client-ish jest project.

That boundary is currently invisible — a future edit that adds `easing: expect.any(Function)` here to "make the two tests match" would look like a tidy-up and would break the RSC suite for a non-obvious reason. Suggest pinning it instead of leaving it implicit:

```tsx
// `easing` still needs the client — Easing is only resolved when read.
expect(() => TransitionSpecs.FadeSpec.config.easing).toThrow();
```

If that does not throw and `easing` resolves fine on the server, then the getter is not needed either and the original eager form would have worked — so either way the assertion is informative. Worth a short comment on the getter in `TransitionSpecs.tsx` too, saying reading it requires the client.

packages/expo-router/src/react-navigation/material-top-tabs/__tests__/missing-dependency.test.ios.tsx
```diff
@@ -9,10 +9,11 @@ jest.mock(
 );
 
 test('throws an error when react-native-tab-view is not installed', () => {
+  jest.resetModules();
   jest.doMock('react-native-tab-view', () => {
     throw new Error();
   });
-  expect(() => require('../index')).toThrow(
+  expect(() => require('../views/MaterialTopTabView')).toThrow(
```
jakub-agent: [💡 suggestion] Narrowing the target from `../index` to `../views/MaterialTopTabView` weakens what this test guarantees.

The user-facing contract is that importing the public surface produces this error — that is the whole reason the message names the package and mentions Expo Router. Asserting on the internal view module only proves the check exists somewhere inside; it no longer proves it reaches anyone importing `expo-router/js-top-tabs`. If `index.tsx` ever stops pulling `MaterialTopTabView` in eagerly, the diagnostic silently stops firing for users and this test stays green.

Given `jest.resetModules()` is now in place, `require('../index')` should still throw here — so what forced the change? If module-registry ordering was the problem, keeping both assertions costs one line and preserves the entry-point coverage:

```tsx
expect(() => require('../views/MaterialTopTabView')).toThrow(MESSAGE);
expect(() => require('../index')).toThrow(MESSAGE);
```

packages/expo-router/src/layouts/__tests__/TopTabs-exports.test.tsx
```diff
@@ -0,0 +1,32 @@
+function loadExports() {
+  let exports!: {
+    vendored: typeof import('../../react-navigation/material-top-tabs');
+    entry: typeof import('../TopTabs');
+  };
+
+  jest.isolateModules(() => {
```
jakub-agent: [🐧 nit] The `isolateModules` / `resetModules` wrapper here has no counterpart in `Tabs-exports.test.tsx`, which imports both modules at the top of the file. The asymmetry is presumably deliberate — `missing-dependency.test.ios.tsx` in the sibling suite `doMock`s `react-native-tab-view` to throw, so the top-tabs graph is sensitive to registry state in a way bottom-tabs is not — but nothing here records that.

Worth one line saying why, per the repo's habit of keeping the reasoning next to the code. Without it, the obvious future "cleanup" is to make this file look like `Tabs-exports.test.tsx`, which would reintroduce whatever flake motivated the helper.

jakub-agent (commented):

**Iteration 3** (`26ae0c8`). Everything from iteration 2 is addressed, and the approach got simpler rather than more layered:

- **Lazy easing.** `TransitionSpecs.tsx` now resolves `Easing.in(...)` in a getter, so the pure-data transition modules are re-exported straight from the server-evaluable path (`export * as TransitionSpecs from ...`) instead of through a client boundary. The `undefined`-member footgun I flagged is gone. This also settles the question I could not answer statically last round: importing `react-native` on the server is fine, and only the `Easing` access was hostile.
- **The separate `*ReactNavigationClient.tsx` files are gone.** The `export *` moved into the existing `TabsClient.tsx` / `TopTabsClient.tsx`, which already carry `'use client'`. Two fewer modules for the same effect — better than what I suggested.
- **Drift guard added.** `__tests__/Tabs-exports.test.tsx` and `TopTabs-exports.test.tsx` do the `Object.keys` parity check, in the normal jest project where the vendored module evaluates. This is the load-bearing check.
- **Changelog entry added**, and the title/description now describe the actual change.

I re-diffed the re-export lists against the vendored indexes at this SHA: still complete on both sides.

Remaining points are all about the new tests, not the shipped behavior — nothing blocking.

Two things that are not inline-commentable:

**Commit churn.** `26ae0c8` reverts the `BottomTabView.tsx` dependency-array reorder made by its own downstack commit `11fcd12` — the blob goes `8199ea5` → `417de7a`, exactly back to the pre-`11fcd12` content, so `base...head` shows no change to that file at all. Reordering a literal dep array is behaviour-neutral (React compares positionally and the order is fixed per render), so nothing breaks, but it looks like a rebase artifact and it means #48291's edit does not survive into the merged tree. Worth confirming that is intended.

**Parity tests compare presence, not identity.** `!(key in TabsEntry)` catches a dropped export but not one wired to the wrong value. That is a reasonable line to draw — `export * as ns` re-exports would make an identity check brittle under babel's CJS interop — just noting the limit is deliberate.

packages/expo-router/src/layouts/__rsc_tests__/Tabs.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+  BottomTabBarHeightContext,
+  BottomTabView,
+  SceneStyleInterpolators,
+  TransitionPresets,
+  TransitionSpecs,
+  createBottomTabNavigator,
+  useBottomTabBarHeight,
+} from '../Tabs';
+
+// The module proxy answers any property name, so `toBeDefined()` would pass even for a dropped
```
jakub-agent: [🏗️ **design**] The assertion is a real improvement, but this comment credits it with something it does not do — and it is the dropped-export claim specifically that is wrong.

The proxy does not consult the real module. On the server `../react-navigation/bottom-tabs` is never loaded at all; `./TabsClient` compiles to `createClientModuleProxy(key)`, and React's `get` handler mints a reference for whatever name it is asked for:

```js
registerClientReferenceImpl(function () { throw ... }, target.$$id + '#' + name, target.$$async)
```

`registerClientReferenceImpl` sets `$$typeof` to `CLIENT_REFERENCE_TAG` unconditionally. So a fabricated reference for a name that no longer exists anywhere passes `expectClientReference` exactly like a real one. Delete `createBottomTabNavigator` from the vendored index and this test is still green — same as with `toBeDefined()`.

What the marker check *does* buy is worth stating on its own terms: it pins that each name is sourced from a `'use client'` module rather than a server one. That would catch someone re-exporting these straight from `../react-navigation/bottom-tabs` to "simplify" `Tabs.tsx`, which is the change that would break server evaluation again — a regression this suite otherwise would not notice until the `renders Tabs` snapshot failed for an unrelated-looking reason. Genuinely useful, just a different guarantee.

Suggested rewording:

```tsx
// These names come from a `'use client'` module, so on the server they are client references.
// The proxy fabricates one for any name asked of it, so this cannot detect a dropped export —
// `__tests__/Tabs-exports.test.tsx` covers that. What it pins is that the entry still sources
// them across a client boundary rather than directly from `../react-navigation/bottom-tabs`.
```

Same comment is duplicated verbatim in `TopTabs.test.tsx`.

packages/expo-router/src/layouts/__rsc_tests__/Tabs.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+it('resolves transition namespaces on the server', () => {
+  expect(TransitionPresets.ShiftTransition).toBeDefined();
+  const fadeConfig = TransitionSpecs.FadeSpec.config as { easing?: unknown };
+  expect(fadeConfig).toMatchObject({ duration: 150 });
+  if (process.env.EXPO_OS === 'web') {
+    // react-native-web's `Easing` works on the server.
+    expect(fadeConfig.easing).toBeInstanceOf(Function);
+  } else {
+    // On native, `easing` still needs the client — `Easing` is only resolved when read.
+    expect(() => fadeConfig.easing).toThrow();
```
jakub-agent, line 37: [💡 suggestion] A bare `toThrow()` is a weak way to pin a deliberate limitation: it matches any error, so the test keeps passing if the getter starts failing for a reason that has nothing to do with `Easing` being unavailable — a typo in the vendored module, a broken import, a future refactor that throws earlier. The assertion is here to document "reading `easing` needs the client on native", and right now it would be satisfied by "reading `easing` breaks somehow".

Matching the error keeps that honest, e.g. `toThrow(TypeError)` if it is the expected "cannot read properties of undefined" shape, or a `/Easing/` pattern if the message names it. Worth running it once and pinning whatever comes out.

packages/expo-router/src/react-navigation/bottom-tabs/TransitionConfigs/TransitionSpecs.tsx
```diff
@@ -6,14 +6,22 @@ export const FadeSpec: TransitionSpec = {
...
 
 export const ShiftSpec: TransitionSpec = {
   animation: 'timing',
   config: {
     duration: 150,
-    easing: Easing.inOut(Easing.ease),
+    // Resolve lazily so this module evaluates in React Server Components.
+    // Reading `easing` still needs the client on native — `Easing` throws on the server there.
+    get easing() {
+      return Easing.inOut(Easing.ease);
```
jakub-agent, line 24: [🐧 nit] The two-line comment is identical above both getters, in a file that is now 27 lines. One copy above `FadeSpec` — or a single note at the top of the module, since it explains a property of the module as a whole rather than of either spec — carries the same information with less to keep in sync.

jakub-agent (commented):

**Iteration 4** (`990947b`). All four points from iteration 3 are addressed:

- `expectClientReference` replaces the `toBeDefined()` block, so the assertion now checks something the proxy does not hand out for free by default.
- The `easing` boundary is pinned explicitly, and split by platform — `toBeInstanceOf(Function)` on web, `toThrow()` on native — with the divergence documented in `TransitionSpecs.tsx` as well.
- `missing-dependency.test.ios.tsx` keeps `require('../index')`; only `jest.resetModules()` was added. The public-entry assertion survives.
- `TopTabs-exports.test.tsx` dropped the `isolateModules` helper and is now symmetric with `Tabs-exports.test.tsx`. Removing the asymmetry is a better answer than documenting it, and it is consistent — `resetModules()` in the polluting test makes the isolation unnecessary, and jest gives each test file its own registry anyway.

The stack concern is also resolved: the base branch advanced, the PR is a single commit again, and `BottomTabView.tsx` no longer appears in the diff — so #48291's dependency-array edit survives into the merged tree.

`Tabs.tsx`, `TopTabs.tsx`, `TabsClient.tsx` and `TopTabsClient.tsx` are unchanged since iteration 3; the re-export lists were already verified complete against the vendored indexes.

Two remaining items, both about the tests rather than the shipped behaviour, and neither blocking. The first is a comment that promises more than the code delivers — worth fixing precisely because the next reader will trust it.

packages/expo-router/src/layouts/__rsc_tests__/TopTabs.test.tsx
```diff
@@ -0,0 +1,33 @@
+/// <reference types="jest-expo/rsc/expect" />
+
+import TopTabs, {
+  MaterialTopTabBar,
+  MaterialTopTabView,
+  createMaterialTopTabNavigator,
+  useTabAnimation,
+} from '../TopTabs';
+
+// The module proxy answers any property name, so `toBeDefined()` would pass even for a dropped
```
jakub-agent, line 10: [💡 suggestion] This is the copy that was left behind — the identical two lines were removed from `Tabs.test.tsx` in this iteration, so the claim is presumably accepted as wrong, just not deleted in both places.

To restate why it is wrong: the proxy never consults `../react-navigation/material-top-tabs` (on the server that module is never loaded), so it fabricates a reference for any name asked of it, and `registerClientReferenceImpl` sets `$$typeof` to `CLIENT_REFERENCE_TAG` unconditionally. A dropped export therefore passes `expectClientReference` exactly as `toBeDefined()` would. `__tests__/TopTabs-exports.test.tsx` is what catches that.

Not blocking. Worth noting though that deleting the comment on the `Tabs` side rather than correcting it leaves the two files inconsistent — one helper with no explanation, one with an inaccurate explanation. If the helper deserves a comment, the accurate version is that it pins these names as coming from a `'use client'` module, which would catch a future re-export straight from the vendored index; if it does not, dropping both copies is the tidier end state.

jakub-agent (approved):

**Iteration 5** (`9db46bd`) — approving.

The shipped behaviour has been right since iteration 3 and has not changed since: the entries are server-evaluable, the transition namespaces are re-exported from the server side with lazy easing, the re-export lists are complete against the vendored indexes, and the parity tests guard them from drift.

Comparing blobs against `990947b`, only `Tabs.test.tsx` changed — the inaccurate comment above `expectClientReference` was removed. `TopTabs.test.tsx`, `TransitionSpecs.tsx` and `missing-dependency.test.ios.tsx` are byte-identical to the previous iteration, so two of my last three points stand open. Both were `suggestion`/`nit` and are entirely reasonable to decline:

- the bare `expect(() => fadeConfig.easing).toThrow()` is still unqualified;
- the two-line comment is still duplicated above both getters in `TransitionSpecs.tsx`.

One looks like an oversight rather than a decision, so I have left a single note on it: the comment deleted from `Tabs.test.tsx` is still there word-for-word in `TopTabs.test.tsx`. It is a stale comment in a test file, not a defect in behaviour, hence approving rather than holding.

For the record, the arc of this review: a snapshot asserting `$undefined` → the `export *` regression from #46117 identified as the root cause → entries made server-evaluable → lazy easing so the data modules stay on the server → parity tests, changelog, and the restored `TopTabs` coverage. Good outcome.

PR #48336, opened 2026-07-30
[docs][gh] Add SDK beta docs cutoff script and workflow

amandeepmittal:
# Why

Fix ENG-25565

# How

Adds **Docs SDK Beta** workflow (`.github/workflows/docs-sdk-beta.yml`) which runs the namesake script from the GitHub Actions tab, passing the major version to cut as `sdkVersion`, for example `58`:

- Refreshing the canary example on the Reference index.
- Regenerating the `unversioned` API data (`et generate-docs-api-data`).
- Generating the versioned reference docs for the new SDK (`et generate-sdk-docs`).
- Cloning hardcoded type links into the new version directory.
- Syncing the app config schema and repointing the import in `pages/versions/<version>/config/app.mdx` from `unversioned` to the new versioned schema file.
- Creating the `native-modules.json` stub, so the cut is not blocked on the new SDK's packages being live on `exp.host`. Run `pnpm versions-schema-sync` in `docs` to replace it once they are.
- ~~Archiving the previous SDK's `llms-sdk.txt` bundle and linking it from `pages/llms.mdx`.~~
- Adding the new SDK's row to the compatibility table.
- Setting `betaVersion` in `docs/package.json`, leaving `version` at the released SDK.

# Test Plan

See test PR opened after running this workflow:
- Test PR: https://github.com/expo/expo/pull/48339

The following screenshots were taken from the PR's preview deployment:

CleanShot 2026-07-30 at 21 47 08@2x

CleanShot 2026-07-30 at 22 23 34@2x

# Checklist

### DIFF
--- a/.github/workflows/docs-sdk-beta.yml
+++ b/.github/workflows/docs-sdk-beta.yml
@@ -1,0 +1,120 @@
+name: Docs SDK Beta
+
+defaults:
+  run:
+    shell: bash
+
+on:
+  workflow_call:
+    inputs:
+      sdkVersion:
+        description: 'Enter the SDK version'
+        required: true
+        type: string
+  workflow_dispatch:
+    inputs:
+      sdkVersion:
+        description: 'Enter the SDK version'
+        required: true
+      reactNative:
+        description: 'React Native version for the compatibility row, for example 0.87'
+        required: false
+        default: ''
+      dryRun:
+        description: 'Validate and print the plan without writing anything'
+        type: boolean
+        default: false
+  pull_request:
+    paths:
+      - '.github/workflows/docs-sdk-beta.yml'
+      - 'docs/scripts/sdk-release/**'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  # JS-only job, so skip the ~685 MB @shopify/react-native-skia prebuilt-binary download.
+  SKIP_SKIA_DOWNLOAD: '1'
+  SDK_VERSION: ${{ inputs.sdkVersion || '99' }}
+  DRY_RUN: ${{ github.event_name != 'workflow_dispatch' || inputs.dryRun }}
+
+jobs:
+  beta:
+    if: github.repository == 'expo/expo'
+    runs-on: ubuntu-24.04
+    steps:
+      - name: 👀 Checkout
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+      - name: 🏗️ Setup pnpm
+        uses: pnpm/action-setup@0e279bb959325dab635dd2c09392533439d90093 # v6.0.8
+        with:
+          version: 10
+      - name: ⬢ Setup Node
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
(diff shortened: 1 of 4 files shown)


entiendoNull (approved):
Forwarded feedback points in DM, not a blocker though.

.github/workflows/docs-sdk-beta.yml
```diff
@@ -0,0 +1,121 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  # JS-only job, so skip the ~685 MB @shopify/react-native-skia prebuilt-binary download.
+  SKIP_SKIA_DOWNLOAD: '1'
+  SDK_VERSION: ${{ inputs.sdkVersion }}
+  DRY_RUN: ${{ inputs.dryRun }}
+
```
kadikraman, line 47: Shouldn't `inputs.reactNative` also be part of this `env`? It's the only input that's inlined

.github/workflows/docs-sdk-beta.yml
```diff
@@ -0,0 +1,121 @@
+name: Docs SDK Beta
+
+defaults:
+  run:
+    shell: bash
+
+on:
+  workflow_call:
+    inputs:
```
kadikraman, line 9: `inputs.reactNative` and `inputs.dryRun` are missing here

docs/scripts/sdk-release/beta.ts
```diff
@@ -0,0 +1,644 @@
...
+
+  const response = await fetch(EXPO_REGISTRY_URL);
+  if (!response.ok) {
+    throw new Error(
+      `Could not read the expo package from ${EXPO_REGISTRY_URL} (HTTP ${response.status}), so the ` +
+        'canary example could not be refreshed.'
+    );
+  }
+
+  const registry = (await response.json()) as Record<string, Record<string, string> | undefined>;
```
kadikraman: We're fetching the whole repo just to read the dist-tag. Should instead fetch the tags only https://registry.npmjs.org/-/package/expo/dist-tags

docs/scripts/sdk-release/beta.ts
```diff
@@ -0,0 +1,644 @@
...
+          `build. Re-run this once main publishes a ${version} canary.`
+  );
+}
+
+function regenerateUnversionedApiData() {
+  expotools(['generate-docs-api-data']);
+}
+
+function generateVersionedDocs() {
+  expotools(['generate-sdk-docs', '--sdk', version]);
```
kadikraman, line 253: `expotools()` uses `execFileSync(..., { stdio: 'pipe' })` with the default 1 MB `maxBuffer` - this means that if the command prints out more than 1 MB during execution, node will kill it and we end up with partial docs since the `existsSync` check will still pass.

I'm not sure how likely is for the output to excelled 1MB, but to guard against it we could either increase the max buffer, stream output to the CI log or to a file instead.

amandeepmittal: In my testing, this didn't surface but taking a look.

amandeepmittal:
Step by step commits introduced and tested here: https://github.com/expo/expo/pull/48536.

kadikraman (approved):
LGTM, just will need to co-ordinate with https://github.com/expo/expo/pull/48574 since the current changes expect the llms to be there

PR #48347, opened 2026-07-30
[router] Make `beforeRemove` non-preventable and add `removePrevented` event

Ubax:
# Why

At the moment `beforeRemove` can be prevented, allowing users to asynchronously prevent screen dismissal. This is against new stack model, introduced by react-native-screens in a new version of stack.

# How

- Make `beforeRemove` non-preventable
- Emit `removePrevented` when navigation is blocked
- Replaces `PreventRemoveProvider` with hook in order to keep the prevention state inside `useNavigationBuilder`

# Test Plan

1. CI
2. New router-e2e app

# Checklist


Ubax:
* **#48347** <a href="https://app.graphite.com/github/pr/expo/expo/48347?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/expo/48347?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Ubax:
## 🤖 AI code review

Four correctness warnings affect event contracts, combined prevention state, blocked-action retries, and committed option visibility.

### 🟡 Warning (4)

- **Do not add preventDefault to every event** — [`packages/expo-router/src/react-navigation/core/useEventEmitter.tsx:126`](https://github.com/expo/expo/pull/48347/files#diff-af33eb74ebb918dc5ca4a7ad3d10d7ce4f4011033fdb33bbfaac4c4ad7a07212R126) _(correctness)_ · `id:b990349c79a0`
  Non-preventable events normally omit `preventDefault`. This change adds it to all events. Method checks can now report incorrect warnings.
  _Suggestion:_ Add the warning method only to `beforeRemove`. Keep it absent from other non-preventable events.

- **The hook overrides declarative removal prevention** — [`packages/expo-router/src/react-navigation/core/usePreventRemove.tsx:26`](https://github.com/expo/expo/pull/48347/files#diff-dcfc71d37e50f6795eaa3030d28630c16bd6b02308e16e2d54dd9c77386e84a2R26) _(correctness)_ · `id:08c691146fa5`
  The hook writes `false` when it has no active registration. This value can override an existing `preventRemove: true` screen option.
  _Suggestion:_ Keep hook state separate from the screen option. Combine both values without replacing the screen option.

- **Redispatching the blocked action repeats prevention** — [`packages/expo-router/src/react-navigation/core/usePreventRemove.tsx:74`](https://github.com/expo/expo/pull/48347/files#diff-dcfc71d37e50f6795eaa3030d28630c16bd6b02308e16e2d54dd9c77386e84a2R74) _(correctness)_ · `id:abf67a67206f`
  A callback can dispatch the blocked action after confirmation. Prevention remains active, so the callback runs again instead of removing the screen.
  _Suggestion:_ Mark the blocked route on the action. Skip prevention for that route when the callback dispatches the same action.

- **Render-time ref mutation exposes uncommitted options** — [`packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx:814`](https://github.com/expo/expo/pull/48347/files#diff-c93b1acbd20e31f74c1d8255b3941e90d56f74d6e884634c2737ec08edd3327bR814) _(correctness)_ · `id:6404e21c6def`
  React can abandon a concurrent render after this assignment. The active callback can then read options from an uncommitted render.
  _Suggestion:_ Update the ref in a layout or insertion effect. This keeps uncommitted descriptors hidden from active callbacks.


---

packages/expo-router/src/react-navigation/core/useOnPreventRemove.tsx
```diff
@@ -27,80 +31,106 @@ export const getPreventableRoutes = (
...
+  currentRoutes: { key?: string }[],
+  nextRoutes: { key?: string }[],
+  action: NavigationAction
+) => {
+  for (const route of getRemovedRoutes(currentRoutes, nextRoutes)) {
+    if (preventRemoveListeners[route.key]?.(action)) {
+      return true;
+    }
 
+    if (isRoutePrevented(route.key)) {
```
jakub-agent, line 59: [🚨 **critical**] `shouldPreventRemove` no longer reads or writes `VISITED_ROUTE_KEYS`, so dispatching the action handed to the callback re-enters this branch forever.

Each redispatch happens synchronously inside `emitter.emit`, so the frames stack up:

```tsx
usePreventRemove(true, ({ data }) => {
  Alert.alert('Discard?', '', [
    { text: 'Discard', onPress: () => navigation.dispatch(data.action) },
  ]);
});
```

I ran that against this head and `main`:

| | callback invocations | result |
|---|---|---|
| `origin/main` | 1 | route removed, `['foo']` |
| this head | **532** | `RangeError: Maximum call stack size exceeded`, `['foo', 'bar']` |

On `main` this terminated by design: the old `shouldPreventRemove` added `route.key` to `visitedRouteKeys` *before* emitting and attached the set to the action it passed to the listener, so the redispatch skipped the route and proceeded. That escape hatch is gone, and `emitBeforeRemove` — which kept the dedup — never runs on the prevented path.

This is the pattern in React Navigation's own "Preventing going back" guide, and `usePreventRemove` is exported from `expo-router/react-navigation`, so existing apps hit it on upgrade. A stack overflow is also about the least diagnosable failure mode available.

Two ways out, either of which would work for me:

- Carry `VISITED_ROUTE_KEYS` through `shouldPreventRemove` as well, so redispatching the *same* action object proceeds. That keeps the old idiom working and makes the new event a superset of the old one.
- If redispatch is intentionally unsupported, detect the re-entrancy and throw a message that says what happened, why, and what to do instead (the repo error-message guidance in `.claude/CLAUDE.md`) rather than recursing.

Either way this needs a regression test — the coverage that would have caught it was removed from `useOnAction.test.ios.tsx` (see my other comment there).

packages/expo-router/src/react-navigation/core/types.tsx
```diff
@@ -120,7 +120,8 @@ export type EventMapCore<State extends NavigationState> = {
   focus: { data: undefined };
   blur: { data: undefined };
   state: { data: { state: State } };
-  beforeRemove: { data: { action: NavigationAction }; canPreventDefault: true };
+  beforeRemove: { data: { action: NavigationAction } };
```
jakub-agent, line 123: [🚨 **critical**] Dropping `canPreventDefault: true` means `event.preventDefault` is now `undefined`, so every existing `beforeRemove` listener that calls it throws:

```tsx
navigation.addListener('beforeRemove', (e) => { e.preventDefault(); });
```

→ `TypeError: e.preventDefault is not a function`, thrown out of `emitBeforeRemove` → `onAction` → `goBack()`. I verified it: on `main` this prevents the removal; on this head it throws and, because `emitBeforeRemove` runs *before* `setState(result)`, the navigation also never commits. So the app gets an unhandled error **and** a stuck screen, with nothing pointing at `usePreventRemove`.

The breaking change itself is fine and well-motivated. What's missing is the migration path. Consider keeping a `preventDefault` on the event that throws (or `console.error`s once in dev) with a message naming the replacement — what failed, why, and what to use instead. `removePrevented.test.ios.tsx:555` currently asserts the bare `TypeError`, which locks in the opaque behaviour; that assertion should become the descriptive error instead.

packages/expo-router/src/layouts/experimental-stack/ExperimentalStackView.tsx
```diff
@@ -90,6 +90,17 @@ export function ExperimentalStackView({ state, navigation, descriptors }: Props)
                 setNextDismissedKey(route.key);
               }}
               onNativeDismissPrevented={() => {
+                navigation.emit({
+                  type: 'removePrevented',
```
jakub-agent: [🏗️ **design**] This emits on the outer navigator's emitter targeting `route.key`, which skips prevention registered inside a nested navigator.

`preventNativeDismiss` on line 52 reads `preventedRoutes[route.key]`, and that entry is set either by a `usePreventRemove` in *this* navigator's screen **or** by a nested provider bubbling up through `setParentPrevented`. In the nested case the screen's listener lives on the nested navigator's emitter, so it never sees this event.

Verified with a nested `ExperimentalStack` (`/b/_layout` + `/b/index` calling `usePreventRemove(true, cb)`):

```
outer b preventNativeDismiss: true
nested usePreventRemove callback calls: 0
```

So the swipe is correctly blocked natively, but the app gets no signal and can't show a confirm dialog — the dismissal is silently swallowed. The action-dispatch path handles this correctly because `shouldPreventRemove` recurses into `preventRemoveListeners`; this direct emit bypasses that walk.

Routing the native-dismiss case through the same helper (or through `preventRemoveListeners` before falling back to a direct emit) would keep the two paths consistent. Not a regression — `onNativeDismissPrevented` emitted nothing before this PR — but it's a hole in the feature being added, and worth a nested test either way.

packages/expo-router/src/react-navigation/core/usePreventRemove.tsx
```diff
@@ -10,7 +10,31 @@ import { usePreventRemoveContext } from './usePreventRemoveContext';
 import { useRoute } from './useRoute';
 
 /**
- * Hook to prevent screen from being removed. Can be used to prevent users from leaving the screen.
+ * Prevents the screen from being removed while `preventRemove` is `true` and calls `callback`
+ * with the blocked navigation action.
+ *
+ * To continue, first set `preventRemove` to `false`, then call `router.back()` from the same
```
jakub-agent, line 16: [🏗️ **design**] This works, but only for `router.back()` specifically, and the doc doesn't say why — which makes it easy to follow the advice and still get a dead button.

`setPreventRemove(false); router.back()` succeeds because expo-router queues the navigation until after React flushes, so `usePreventRemove`'s effect deregisters first. The React Navigation equivalent in the same handler does not:

```tsx
setDirty(false);
navigation.goBack(); // still blocked
```

Verified: routes stay `['foo', 'bar']`, callback fires once, nothing visible happens. The registration lives in a `useEffect`, so at the moment `goBack()` dispatches, `entriesRef` still says prevented.

Since this hook is exported from `expo-router/react-navigation` and mirrors the React Navigation API, users will reach for `navigation.goBack()` / `navigation.dispatch(...)`. Worth stating explicitly that `router.back()` works because it's queued and that `navigation.goBack()` in the same handler will not — otherwise the failure is silent.

packages/expo-router/src/react-navigation/core/__tests__/useOnAction.test.ios.tsx
```diff
@@ -695,7 +696,7 @@ test('logs error if no navigator handled the action', () => {
   spy.mockRestore();
 });
 
-test("prevents removing a screen with 'beforeRemove' event", () => {
+test("prevents removing a screen with 'removePrevented' event", () => {
```
jakub-agent, line 700: [💡 suggestion] The `shouldContinue` branches were deleted from these four tests without replacement, which removes exactly the coverage that would have caught the redispatch overflow.

The old tests exercised "prevention is active **and** the listener redispatches the blocked action" — the path that now recurses to a `RangeError`. Nothing in the new suite covers redispatch: `removePrevented.test.ios.tsx` and the new `usePreventRemove` test only dispatch from an effect *after* prevention is disabled.

The tails of these tests are now inert too. After `setPreventRemove(false)` nothing re-enables prevention, so the closing `navigate('bar'); navigate('foo')` + assertions no longer test prevention at all — they just assert a plain navigation. Either restore prevention before that block or drop it.

Whatever the resolution of the redispatch behaviour is, it should land as an assertion here.

packages/expo-router/src/react-navigation/core/usePreventRemoveState.tsx
```diff
@@ -11,102 +11,111 @@ import { type PreventedRoutes, PreventRemoveContext } from './PreventRemoveConte
...
 type Props = {
   children: React.ReactNode;
+  isRoutePreventedRef: React.MutableRefObject<IsRoutePrevented>;
 };
 
+export type IsRoutePrevented = (routeKey: string) => boolean;
+
 type PreventedRouteEntry = {
   routeKey: string;
   preventRemove: boolean;
```
jakub-agent, line 21: [💡 suggestion] This field can only ever be `true` — `setPreventRemove` deletes the entry when `preventRemove` is false and hardcodes `{ routeKey, preventRemove: true }` otherwise.

So every read of it is a no-op predicate: `entry.preventRemove &&` (line 101), `.some(({ preventRemove }) => preventRemove)` (line 95), `entry.preventRemove && !isRoutePreloadedInStack(...)` (line 80), and the `acc[routeKey]?.preventRemove || preventRemove` merge in `transformPreventedRoutes`.

Collapsing the map to `Map<string, string>` (id → routeKey) removes the field and simplifies all four sites. It carried over from `main`, but this file is largely rewritten here anyway, so it's a good moment.

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
```diff
@@ -739,13 +739,21 @@ export function useNavigationBuilder<
 
   const { keyedListeners, addKeyedListener } = useKeyedChildListeners();
 
+  const isRoutePreventedRef = React.useRef<IsRoutePrevented>(() => false);
```
jakub-agent: [💡 suggestion] Worth a line saying why this ref is threaded down instead of read from context: `PreventRemoveProvider` renders *below* `useOnAction`, so the action handler can't consume `PreventRemoveContext`, and the query has to be synchronous so a dispatch in the same tick sees a just-registered `usePreventRemove`.

A child writing into a ref its parent hook owns is unusual enough that the next reader will wonder whether the context would have worked. One comment here (and/or at the `useLayoutEffect` in `PreventRemoveProvider.tsx:106`) saves that detour.

packages/expo-router/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### 🛠 Breaking changes
 
+- Make `beforeRemove` non-preventable and add `removePrevented` event for removal prevention.
```
jakub-agent: [🐧 nit] Missing the PR link and author attribution that every neighbouring entry has: `([#48347](https://github.com/expo/expo/pull/48347) by [@Ubax](https://github.com/Ubax))`.

packages/expo-router/src/react-navigation/native-stack/types.tsx
```diff
@@ -29,6 +30,10 @@ import type {
 } from '../native';
 
 export type NativeStackNavigationEventMap = {
+  /**
+   * Event which fires when an action that removes the screen is prevented.
+   */
+  removePrevented: { data: { action: NavigationAction } };
```
jakub-agent: [🐧 nit] This entry looks unused. Listeners already get `removePrevented` from `EventMapCore`, and the only reason to also declare it on a navigator's event map is `NavigationHelpers.emit`, which is typed over `EventMap` alone — but nothing in `NativeStackView.native.tsx` emits it (native dismissal goes through `pop()` on line 553). `ExperimentalStackNavigationEventMap` does need it, since `ExperimentalStackView` emits directly.

jakub-agent (changes requested):

Splitting `beforeRemove` (informational) from `removePrevented` (blocked attempt) is the right call — the old event conflated notification and cancellation, and the new synchronous `isRoutePrevented` query is a clean way to let action handling see prevention state without waiting for a render. The nested/grandchild propagation through `setParentPrevented` is correct, and `emitBeforeRemove` keeps the `VISITED_ROUTE_KEYS` dedup so parent/child navigators don't double-emit.

I checked out the PR head, built the package, and ran the suite: **104 suites / 1427 tests pass**. I then wrote throwaway adversarial tests and diffed the behaviour against `main`. Two of them are regressions that I think block the merge:

1. **Redispatching the blocked action from the callback now overflows the stack.** `shouldPreventRemove` dropped the `VISITED_ROUTE_KEYS` dedup that `emitBeforeRemove` kept, so the dedup no longer terminates a redispatch. On `main` the pattern from React Navigation's own "Preventing going back" guide (`usePreventRemove(dirty, ({ data }) => Alert.alert(..., { onPress: () => navigation.dispatch(data.action) }))`) removed the route after one callback. On this head it recurses **532 times and throws `RangeError: Maximum call stack size exceeded`**. `usePreventRemove` is exported from `expo-router/react-navigation`, so this is the pattern users arrive with.
2. **Legacy `beforeRemove` + `preventDefault()` throws a bare `TypeError`** out of `goBack()`, and because `emitBeforeRemove` runs before `setState` the navigation also silently never commits. A deliberate break is fine; an opaque `TypeError: e.preventDefault is not a function` isn't a migration path.

I also confirmed a gap in the new native-dismiss handling: on `ExperimentalStack`, when prevention comes from a **nested** navigator, `preventNativeDismiss` is correctly `true` but `onNativeDismissPrevented()` emits `removePrevented` only on the outer navigator, so the nested screen's `usePreventRemove` callback is never called (verified: 0 invocations). The action-dispatch path recurses into children correctly; this direct emit doesn't.

Two notes on the description rather than the code:

- "Remove the unreleased `preventRemove` screen option" — I can't find such an option on `main` (`git grep 'preventRemove?:' origin/main -- packages/expo-router` is empty) and nothing is removed in the diff. Stale bullet?
- "expose a synchronous query to action handling **and native stack views**" — the views still read the async `preventedRoutes` map (`NativeStackView.native.tsx:256`, `ExperimentalStackView.tsx:52`, `modal/web/ModalStack.tsx:127`). That's probably correct (a ref wouldn't re-render), but only action handling got the synchronous query.

Finally, no `docs/` guide mentions `beforeRemove` or `usePreventRemove`, and neither appears in the generated API reference — so I don't think guide updates are owed here. Worth double-checking whether `usePreventRemove`'s new JSDoc should surface in the reference at all.

packages/expo-router/src/react-navigation/core/useOnPreventRemove.tsx
Ubax: Address this in two ways:
1. Show a warning when action is dispatched from the `usePreventRemove`
2. Add a prop `dangerous_skipPreventRemove` or something similar that aligns a bit better with codebase. This should be added to all `router...` actions and be an option in `dispatch` payload, so it can be used using both APIs

packages/expo-router/src/layouts/experimental-stack/createExperimentalStackNavigator.tsx
```diff
@@ -62,7 +63,13 @@ function ExperimentalStackNavigator({
 
   const mergedDescriptors = useMemo(
     // TODO(@ubax): implement properly when more stack options are available
-    () => mergeOptions(descriptors as NativeStackDescriptorMap, registry, state),
+    // `mergeOptions` only changes options, so the experimental descriptor event type is preserved.
+    () =>
+      mergeOptions(
```
Ubax: Instead of cast make mergeOptions a generic

```
export function mergeOptions<T extends ....>(
  descriptors: T,
  registry: CompositionRegistry,
  state: NativeStackViewState
): T {
```

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
```diff
@@ -739,13 +739,23 @@ export function useNavigationBuilder<
 
   const { keyedListeners, addKeyedListener } = useKeyedChildListeners();
 
+  // The provider renders below the action handler, so context cannot be consumed here. The ref
+  // lets the handler query the provider synchronously while dispatching an action.
+  const isRoutePreventedRef = React.useRef<IsRoutePrevented>(() => false);
```
Ubax: Can we remove this `ref`, by wrapping `useNavigationBuilder` with prevent provider?

packages/expo-router/src/react-navigation/core/useOnPreventRemove.tsx
```diff
@@ -27,80 +32,125 @@ export const getPreventableRoutes = (
...
+      );
+    });
+  }, [addKeyedListener, emitter, getState, isRoutePrevented, preventRemoveListeners, routeKey]);
+
+  React.useEffect(() => {
+    if (!routeKey) {
+      return;
+    }
+
+    return addKeyedListener?.('beforeRemove', routeKey, (action) => {
```
Ubax, line 152: Why is this needed?

packages/expo-router/src/react-navigation/core/useOnPreventRemove.tsx
```diff
@@ -27,80 +32,125 @@ export const getPreventableRoutes = (
...
-
-    const event = emitter.emit({
+    emitter.emit({
       type: 'beforeRemove',
       target: route.key,
       data: { action: beforeRemoveAction },
-      canPreventDefault: true,
+      preventDefault() {
+        throw new Error(
+          '`beforeRemove` is a notification-only event and cannot prevent screen removal. Use `usePreventRemove` with the `removePrevented` event instead.'
```
jakub-agent, line 111: [🏗️ **design**] The message is exactly right. The open question is whether it should *throw* in production, because the throw does two things beyond surfacing the message.

It propagates out of `emitBeforeRemove` → `onAction` → `goBack()`, so it escapes into whatever called it (a press handler, `router.back()`), and it aborts the `for` loop over removed routes. Verified with two screens that both have `beforeRemove` listeners, where only the top one is legacy, then `router.dismissAll()`:

```
beforeRemove listeners that fired: ["b"]     // 'a' never notified
threw: true
pathname: /b                                 // navigation blocked
```

So one un-migrated listener anywhere in the stack suppresses `beforeRemove` for every other route being removed and hard-blocks the navigation — in production, since there's no `__DEV__` gate. For the legacy caller that's arguably fine (they wanted prevention and they get it, loudly), but the collateral damage to unrelated listeners on other routes is the part that seems unintended.

A `console.error` here plus `return` — letting the loop and the removal continue — would surface the same message without turning an informational event into a navigation-blocking crash. If you'd rather keep the throw for its bluntness, gating it on `__DEV__` at least confines it to development. Either way `removePrevented.test.ios.tsx` and `removePrevented.test.web.tsx` currently assert the blocked-navigation outcome, so the decision is worth making explicitly rather than inheriting it from the assertion.

(For the record, the throwing shim was my suggestion in iteration 1 — I hadn't spotted the mid-loop abort. Your call which way to go.)

packages/expo-router/src/layouts/experimental-stack/ExperimentalStackView.tsx
```diff
@@ -90,6 +91,19 @@ export function ExperimentalStackView({ state, navigation, descriptors }: Props)
...
+                  // A real pop runs child-first prevention checks and notifies the nested route
+                  // that owns the guard; emitting directly here would only reach this route.
+                  navigation.dispatch({
+                    ...StackActions.pop(),
+                    source: route.key,
+                    target: state.key,
+                  });
+                } else {
+                  console.warn(
+                    `ExperimentalStack received \`onNativeDismissPrevented\` for route '${route.name}' without an active removal guard. The dismiss action was ignored because prevention context and native state are out of sync.`
```
jakub-agent, line 104: [💡 suggestion] This warning isn't `__DEV__`-gated and has no once-guard, which differs from the convention already established in this file — `useUnsupportedOptionsWarning` at line 187 does `if (!__DEV__ || warnedRef.current) return;`.

It also fires on a benign race rather than a bug: native calls `onNativeDismissPrevented` based on the `preventNativeDismiss` prop it last received, so a user who disables prevention and immediately swipes will hit the `else` branch legitimately. Since a swipe can be repeated, this can warn several times in a row in production for a flow that's working as designed.

Gating it on `__DEV__` (and ideally a once-per-route ref) would match the neighbour and keep it out of production logs.

packages/expo-router/src/react-navigation/core/types.tsx
```diff
@@ -190,6 +191,7 @@ export type EventEmitter<EventMap extends EventMapBase> = {
     options: {
       type: EventName;
       target?: string;
+      preventDefault?: () => void;
```
jakub-agent, line 194: [💡 suggestion] This widens the public `emit` signature for every event type, when the only caller is `emitBeforeRemove`.

Two consequences: any `emit` call can now pass a `preventDefault` regardless of whether the event declares `canPreventDefault`, and when `canPreventDefault` *is* true the value is silently dropped — `useEventEmitter` reaches the new `else if (preventDefault)` branch only when the first condition fails, so the two are mutually exclusive at runtime but not in the type.

Making that exclusivity explicit in the type (alongside the existing `canPreventDefault extends true ? ... : ...` conditional, which already discriminates on the same axis) would keep the invariant checkable instead of leaving it to the reader of `useEventEmitter`.

packages/expo-router/src/react-navigation/core/useOnPreventRemove.tsx
```diff
@@ -5,19 +5,24 @@ import { use } from 'react';
...
 type Options = {
   getState: () => NavigationState;
+  isRoutePrevented: IsRoutePrevented;
   emitter: NavigationEventEmitter<EventMapCore<any>>;
+  preventRemoveListeners: Record<string, ChildPreventRemoveListener | undefined>;
   beforeRemoveListeners: Record<string, ChildBeforeRemoveListener | undefined>;
 };
 
 const VISITED_ROUTE_KEYS = Symbol('VISITED_ROUTE_KEYS');
+const emittingRemovePreventedKeys = new Set<string>();
```
jakub-agent, line 25: [🐧 nit] Module-level mutable state shared by every navigation container in the process. It works — route keys are nanoid-suffixed so cross-container collisions aren't realistic, and the `try/finally` makes this a reliable call-stack marker — but it's worth a comment saying that's the intent, since a reader coming from the `VISITED_ROUTE_KEYS` symbol two lines above will wonder why this one isn't carried on the action.

Worth noting the choice is also *broader* than action-scoped dedup would have been: it catches a fresh action dispatched from the callback, not just a redispatch of the same object. That's a genuine improvement over what I suggested last round, and the reason is worth capturing in the comment.

packages/expo-router/src/react-navigation/core/usePreventRemove.tsx
jakub-agent, line 16: [💡 suggestion] Carried over from iteration 1 — re-verified on this head, still silent.

`router.back()` works here because expo-router queues it until after React flushes, so the hook's effect deregisters first. The React Navigation equivalent in the same handler does not:

```tsx
setDirty(false);
navigation.goBack(); // routes stay ['foo', 'bar'], callback fires once, nothing happens
```

The new dev warning in `useOnPreventRemove` does point at `router.back()`, which covers the redispatch case. This is the other half: someone who never touches the callback's action and just reaches for `navigation.goBack()` gets no warning at all. One clause saying `router.back()` works because it's queued, and that `navigation.goBack()` in the same handler won't, would close it.

packages/expo-router/src/react-navigation/core/usePreventRemoveState.tsx
jakub-agent, line 21: [🐧 nit] Carried over from iteration 1, unchanged. This field can only ever be `true` — `setPreventRemove` deletes the entry when it's false and hardcodes `preventRemove: true` otherwise — so the four reads of it (lines 80, 95, 101, and the `transformPreventedRoutes` merge) are all no-op predicates. Collapsing the map to `Map<string, string>` (id → routeKey) drops the field and simplifies all four. Fine to leave if you'd rather not touch it again.

jakub-agent (commented):

**Iteration 2** — re-reviewed `afb8605` → `bb27c48` (rebased onto a newer `main`, so I diffed only the prevent-removal paths). Rebuilt the package and ran the affected suites: **110 suites / 1489 tests pass** (up from 104/1427). Both blockers from iteration 1 are genuinely fixed, and I re-ran my adversarial cases to confirm rather than taking the tests' word for it.

**Fixed and verified**

- **Redispatch overflow** — the `emittingRemovePreventedKeys` re-entrancy guard stops it. No more `RangeError`; a `__DEV__` warning explains the alternative. I also checked the guard doesn't catch legitimate flows: `setDirty(false); router.back()` *inside* the callback still works (0 warnings, navigation completes), because `router.back()` is queued and the guard is cleared by then. The `try/finally` means a throwing listener can't leave a stale key behind either.
- **`beforeRemove` + `preventDefault()`** — now a descriptive `Error` naming `usePreventRemove` instead of a bare `TypeError`. One consequence worth a decision though; see the inline comment.
- **Nested native dismissal** — dispatching a real pop instead of emitting directly is a better fix than what I suggested: it reuses the child-first prevention walk, so the nested guard is notified and the two code paths stay consistent. Verified by your new test.
- **Test coverage** — the retry-from-effect pattern is now exercised in all four `useOnAction` tests, the inert tails are gone, and there are new tests for the re-entrancy guard and the `preventDefault` error. This was my main worry in iteration 1 and it's well covered now.
- Ref comment, CHANGELOG attribution, and the unused `NativeStackNavigationEventMap.removePrevented` — all done.

**Still open from iteration 1** (both were non-blocking, flagging only so they're a deliberate choice rather than an oversight):

- `usePreventRemove`'s JSDoc still doesn't distinguish `router.back()` from `navigation.goBack()`. Re-verified on this head: `setDirty(false); navigation.goBack()` in one handler leaves the route in place and fires the callback once — silently. The new dev warning does say "call `router.back()` from the handler", which helps once you've hit the guard, but it doesn't tell you the React Navigation call won't work.
- The dead `preventRemove: boolean` field in `PreventRemoveProvider`.

**New, small** — three points on the diagnostics that were added, in the inline comments. Nothing here is a correctness blocker, so this is a COMMENT rather than another REQUEST_CHANGES.

**Description drift.** Two bullets in the PR body no longer match the code:

- "Emit `removePrevented` directly for prevented native dismissals **instead of dispatching a synthetic pop**" — the code now does exactly the opposite, and that's the right call. Worth rewording so the rationale survives in the commit message.
- "Remove the unreleased `preventRemove` screen option" — still no such option on `main` and still nothing removed in the diff. Carried over from iteration 1.

packages/expo-router/src/react-navigation/core/useOnPreventRemove.tsx
Ubax: Ignore

packages/expo-router/src/react-navigation/core/usePreventRemoveState.tsx
Ubax: Leave for now

packages/expo-router/src/react-navigation/core/types.tsx
Ubax: That's ok

packages/expo-router/src/react-navigation/core/usePreventRemove.tsx
Ubax: Skip it for now

packages/expo-router/src/react-navigation/core/useOnPreventRemove.tsx
Ubax: Remove the `warning` for now and add it in a follow-up 

```
console.warn(
            `The action '${action.type}' was dispatched from inside a \`usePreventRemove\` callback and was prevented again. The \`removePrevented\` event was not re-emitted to avoid an infinite loop. There is no way to dispatch directly from the callback; set \`preventRemove\` to \`false\` first, then retry (for example, call \`router.back()\` from the handler or dispatch the captured action from an effect).`
          );
```

packages/expo-router/CHANGELOG.md
```diff
@@ -4,7 +4,7 @@
 
 ### 🛠 Breaking changes
 
-- Remove the `redirect` prop from layout `<Screen>` components. ([#48369](https://github.com/expo/expo/pull/48369) by [@Ubax](https://github.com/Ubax))
+- Make `beforeRemove` non-preventable and add `removePrevented` event for removal prevention. ([#48347](https://github.com/expo/expo/pull/48347) by [@Ubax](https://github.com/Ubax))
```
jakub-agent: [🏗️ **design**] The refactor also removes `PreventRemoveProvider` from the public API, which isn't captured here.

It was a plain (non-deprecated) export on `main` — `react-navigation/core/index.tsx:55`, sitting between `PreventRemoveContext` and the `@deprecated` block that covers `ThemeContext` — and it reaches users through `core/index` → `native/index.tsx:57` (`export * from '../core'`) → `react-navigation/index.ts` → the `./react-navigation` package export. I confirmed the removal in the built output: `build/react-navigation/core/index.js` and `.d.ts` no longer contain the name, while `build/react-navigation/native/index.js:89` still re-exports everything from core.

So `import { PreventRemoveProvider } from 'expo-router/react-navigation'` worked before this PR and is now a missing export. It's also part of upstream `@react-navigation/core`'s public surface, so anyone porting a custom navigator will reach for it. Since wrapping your navigator content in it was the only way to opt into prevention support, a custom navigator doing that breaks at import time.

Worth a second bullet under Breaking changes saying it's removed and that `useNavigationBuilder` now provides `PreventRemoveContext` automatically — so the migration is just deleting the wrapper, which is a genuinely easy fix once someone knows it.

packages/expo-router/src/react-navigation/core/usePreventRemoveState.tsx
```diff
@@ -5,120 +5,114 @@ import { use } from 'react';
...
-    return acc;
+    return result;
   }, {});
-
-  return preventedRoutes;
 };
 
 /**
- * Component used for managing which routes have to be prevented from removal in native-stack.
+ * Hook used for exposing removal prevention state to navigator views.
```
jakub-agent, line 34: [🐧 nit] The file no longer contains a provider — just `usePreventRemoveState` and the `IsRoutePrevented` type. Renaming it to `usePreventRemoveState.tsx` would match both the contents and the `usePreventRemoveContext.tsx` sibling. Only worth doing if you're touching the file again anyway; the three importers (`useNavigationBuilder`, `useOnAction`, `useOnPreventRemove`) are the whole blast radius.

jakub-agent (commented):

**Iteration 3** — reviewed `bb27c48` → `d35ae84`. Short round: I've skipped everything you already triaged (`Ignore` / `Leave for now` / `That's ok` / `Skip it for now`), and the re-entrancy warning you want to strip in a follow-up. One new finding, one nit.

**The ref refactor is a clear improvement.** `PreventRemoveProvider` becoming `usePreventRemoveState` and having `useNavigationBuilder` own the context provider removes the inverted data flow entirely — better than the comment I asked for last round. Two side effects worth noting, both good:

- **One fewer render pass per navigator.** `headerTitle` goes 2 → 1 in `Screen.test.ios.tsx`, and the duplicate call disappears in `HrefPreview.test.ios.tsx`. I checked this isn't a loosened assertion masking a lost render — the content assertions still hold and the full suite is green.
- **The `TODO(@ubax)` act-warning suppression in `SuspenseFallback.test.ios.tsx` is gone** because its root cause is gone. That hack existed precisely because the provider updated a parent's state from a passive effect in a suspended subtree; 29 lines of CI-noise workaround deleted rather than carried forward.

I also re-ran my adversarial cases against this head to confirm the refactor didn't regress them: redispatch still guarded (1 callback, 1 warning, no `RangeError`), nested prevention still blocks the parent back and still continues after disabling (`/nested` → `/`), and `preventNativeDismiss` still reaches the view through the context. `mergeOptions` being generic cleanly removes the double cast, and the new nested-`beforeRemove` test answers your own "why is this needed?" on the keyed listener.

**Test run.** Full package suite: **4586 passed, 4 failed**. None of the 4 are attributable to this PR, and I want to be explicit about why so you don't chase them:

- 3 × `native-tabs/__rsc_tests__/tabs.test.tsx` — flight snapshots embed a *relative* asset path, and my worktree sits deeper in the filesystem than a normal checkout, so `../../../../../../packages/...` comes out as `../../../../../../../../.config/tuft/...`. Environment artefact, not code.
- 1 × `material-top-tabs/__tests__/missing-dependency.test.ios.tsx` — passes in isolation, fails only in the full parallel run, and the file is byte-identical to `main` (`git diff origin/main HEAD -- <file>` is empty). Pre-existing order-dependent flake.

Still worth a pass before merge: the two stale bullets in the PR description ("emit `removePrevented` directly **instead of dispatching a synthetic pop**" is now the opposite of the code, and "Remove the unreleased `preventRemove` screen option" still removes nothing).

packages/expo-router/CHANGELOG.md
Ubax: This should be a new entry - not a rename

Ubax: Ignore

jakub-agent (approved):

**Iteration 4 — approving.** This also supersedes the `CHANGES_REQUESTED` I left on `afb8605`; both regressions behind it are fixed and re-verified.

**Verified on `bbd6dac`**

| check | result |
|---|---|
| `expo-build cjs:src` | clean |
| `tsc -b tsconfig.all.json` | clean |
| `oxlint` | clean |
| Jest (`react-navigation`, `__tests__`, `layouts`, `modal`) | 215 suites / 2718 passed |
| Stale `PreventRemoveProvider` references | none |

Behaviour re-checked end to end rather than inferred from the suite: prevent → continue blocks at the form and lands back on `/` with the callback firing exactly once; the same holds through two levels of nested stacks (`/a/b`); and a legacy `beforeRemove` + `preventDefault()` still produces the descriptive error rather than a `TypeError`. The rename landed as a clean 100% rename with all three importers updated.

**How the review went, for the record**

The two blockers from iteration 1 were real and are gone:

- Redispatching the blocked action from a `usePreventRemove` callback — React Navigation's own documented idiom — recursed 532 frames into `RangeError: Maximum call stack size exceeded` (it terminated in one callback on `main`). The `emittingRemovePreventedKeys` re-entrancy guard fixes it, and I confirmed it doesn't catch legitimate flows: `setDirty(false); router.back()` inside the callback still completes with no warning, because `router.back()` is queued.
- Legacy `beforeRemove` + `preventDefault()` threw a bare `TypeError`; it now throws a message naming `usePreventRemove`.

Two fixes came back better than what I'd suggested. Dispatching a real pop from `onNativeDismissPrevented` reuses the child-first prevention walk instead of duplicating it, so nested guards get notified. And replacing the `isRoutePreventedRef` inversion with `usePreventRemoveState` removed the awkward data flow altogether rather than just documenting it — with two good side effects: one fewer render pass per navigator (`headerTitle` 2 → 1), and the deletion of a 29-line `TODO(@ubax)` act-warning suppression in `SuspenseFallback.test.ios.tsx` whose root cause no longer exists.

**Deliberately left open** — recorded so they're traceable, not as objections. Triaged by the author: the un-`__DEV__`-gated throw and its mid-loop abort of `emitBeforeRemove`, the dead `preventRemove: boolean` field, `preventDefault` widening the `emit` signature, and the `usePreventRemove` JSDoc not distinguishing `router.back()` from `navigation.goBack()`. Deferred to a follow-up: stripping the re-entrancy `console.warn`, and the `dangerous_skipPreventRemove` escape hatch. Waived by the reviewer on this PR: a CHANGELOG bullet for `PreventRemoveProvider` being dropped from the public API — still worth a line whenever the changelog is next touched, since `import { PreventRemoveProvider } from 'expo-router/react-navigation'` worked before this PR.

Two bullets in the PR description are still stale and worth a pass before merge, since the commit message inherits them: "emit `removePrevented` directly **instead of dispatching a synthetic pop**" is now the opposite of the code, and "Remove the unreleased `preventRemove` screen option" removes nothing.

PR #48355, opened 2026-07-31
[ui][android] Fix hosted Pressable firing `onLongPress` during an ancestor scroll

intergalacticspacehighway:
# Why

Reported on discord - https://discord.com/channels/695411232856997968/1504711019199467550/1532337488910221445

Regression caused by https://github.com/expo/expo/pull/48217

`Pressable` inside an `RNHostView` which itself is in a `ScrollView` fires `onLongPress` when user does a Pan action on a `ScrollView`.

As demonstrated below, user is trying to perform a scroll on a ScrollView while keeping the finger on a Menu. It opens the `Menu` which is unexpected, the `Menu`'s onLongPress trigger should've been cancelled as soon as user starts scrolling.

https://github.com/user-attachments/assets/06b1419a-6af0-4078-88aa-bb0bf26df06d

<details>
<summary>Repro code.</summary>

```jsx

  <ScrollView>
    <Host matchContents>
      <RNHostView matchContents>
        <Pressable onLongPress={() => console.log('LONG PRESS')} style={{ height: 64, width: 240 }}>
          <Text>Pan across me</Text>
        </Pressable>
      </RNHostView>
    </Host>
    {/* filler rows */}
  </ScrollView>
```
</details>

# How

We dispatch `ACTION_CANCEL` to `JSTouchDispatcher` using the same timestamp Android provided for `ACTION_DOWN`. This feels a bit hacky. We reuse the React Native `Modal` mechanism for `RNHostView`, but the situations are different. A `Modal` is mounted in a separate window hierarchy, so it cannot have a parent `ScrollView`. `RNHostView`, however, can be inside a `ScrollView`, so we need to handle touch cancellation ourselves. The good news is that once React Native enables the pointer events feature flag, we should be able to remove some of these workarounds.

# Test Plan

Tested the user mentioned repro. Tested the existing behavior for any regressions, tested will multiple touches.

# Checklist

### DIFF
--- a/packages/expo-ui/android/src/main/java/expo/modules/ui/RNHostView.kt
+++ b/packages/expo-ui/android/src/main/java/expo/modules/ui/RNHostView.kt
@@ -182,6 +182,10 @@ private class TouchDispatchingRootViewGroup(
   private val currentLocation = IntArray(2)
   private var trackingGestureOffset = false
 
+  // Timestamp which is sent to JSTouchDispatcher to propogae a custom ACTION_CANCEL event
+  // JSTouchDispatcher uses timestamp as id to recognize the event
+  private var gestureKeyTime = -1L
+
   // True if the sheet consumed scroll on the most recent drag frame; drives the settle decision.
   private var sheetMovingOnLastDragFrame = false
 
@@ -226,6 +230,7 @@ private class TouchDispatchingRootViewGroup(
       trackingGestureOffset = true
       sheetMovingOnLastDragFrame = false
       flingHandledThisGesture = false
+      gestureKeyTime = ev.eventTime
     }
 
     // While a nested scroll is in flight the sheet may be sliding this whole view up/down. Re-express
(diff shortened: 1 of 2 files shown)


packages/expo-ui/android/src/main/java/expo/modules/ui/RNHostView.kt
```diff
@@ -182,6 +182,10 @@ private class TouchDispatchingRootViewGroup(
   private val currentLocation = IntArray(2)
   private var trackingGestureOffset = false
 
+  // Timestamp which is sent to JSTouchDispatcher to propogae a custom ACTION_CANCEL event
```
zoontek, line 185: small typo in the comment (propagate)

zoontek (approved):

aubrey-wodonga:
I've patched expo-ui locally with this fix, and can confirm that it resolves the described ScrollView problem, but I've since found two additional reproductions (the first of which I mentioned in Discord already, but putting it here to keep everything together):

1. If an `RNHostView` hosted pressable overlaps the device's bottom edge, dismissing the app with gesture nav triggers the long press while the app is exiting, which triggers haptic feedback that clashes with the system's feedback, and it also has in-app side effects _after the app dismissed_. In my case, the next warm launch of the app presents with an open MenuView, not even anchored in the right place 😅
2. An `RNHostView` hosted pressable rendered as a child of RNGH's `ReanimatedSwipeable` triggers the long press action mid-swipe. Additionally, the `pressed` state doesn't get cleared when the swipe gesture gets cancelled (e.g. by panning the Y axis, or moving the swipeable back to below its threshold)

intergalacticspacehighway:
Fixed by https://github.com/expo/expo/pull/48969/

PR #48356, opened 2026-07-31
[router] Improve withLayoutContext types

Ubax:
# Why

`withLayoutContext` lost navigator-specific screen option types inside `processScreens`, which required local type suppressions such as the one around `Tabs` `tabBarButton`.

# How

Thread navigator option, state, and event map types through screen processing and sorting. Preserve the default `Record<string, any>` options type for callers that omit generics.

Validate in development that `processScreens` preserves each original screen exactly once. Guard metadata is keyed by the original screen names, so adding, removing, renaming, or duplicating screens would desynchronize it. Reordering screens and changing their options remain supported.

# Test Plan

- `pnpm test src/__tests__/custom-navigators.test.ios.tsx --runInBand`
- `pnpm build`
- `pnpm lint`

Automated tests cover adding, removing, renaming, duplicating, reordering, updating options, omitted generic option access, and the legacy nullish processor fallback.

# Checklist

### DIFF
--- a/packages/expo-router/src/__tests__/custom-navigators.test.ios.tsx
+++ b/packages/expo-router/src/__tests__/custom-navigators.test.ios.tsx
@@ -2,9 +2,30 @@ import { screen } from '@testing-library/react-native';
 import { Text } from 'react-native';
 
 import { Navigator, Slot } from '../index';
+import { withLayoutContext } from '../layouts/withLayoutContext';
 import type { TabRouterOptions } from '../react-navigation/native';
 import { TabRouter } from '../react-navigation/native';
+import { createStackNavigator } from '../react-navigation/stack';
 import { renderRouter } from '../testing-library';
+
+const StackNavigator = createStackNavigator().Navigator;
+
+function renderProcessedScreens(
+  processScreens: NonNullable<Parameters<typeof withLayoutContext>[1]>
+) {
+  const CustomNavigator = withLayoutContext(StackNavigator, processScreens);
+
+  return renderRouter({
+    _layout: () => (
+      <CustomNavigator>
+        <CustomNavigator.Screen name="index" />
+        <CustomNavigator.Screen name="two" />
+      </CustomNavigator>
+    ),
+    index: () => <Text testID="index" />,
+    two: () => <Text testID="two" />,
+  });
+}
 
 it('can render a custom navigator', () => {
   // The default <Stack /> doesn't have any routerOptions, so we use TabRouter
@@ -34,4 +55,44 @@ it('can render a custom navigator', () => {
     backBehavior: 'history',
     initialRouteName: undefined,
   });
+});
+
+it.each([
+  ['adds a screen', (screens) => [...screens, { ...screens[0]!, name: 'three' }]],
+  ['removes a screen', (screens) => screens.slice(1)],
+  ['renames a screen', (screens) => [{ ...screens[0]!, name: 'three' }, screens[1]!]],
+  ['duplicates and removes a screen', (screens) => [screens[0]!, screens[0]!]],
+  ['adds a duplicate screen', (screens) => [screens[0]!, screens[0]!, screens[1]!]],
+] satisfies [string, NonNullable<Parameters<typeof withLayoutContext>[1]>][])(
+  '%s',
+  (_, processScreens) => {
+    expect(() => renderProcessedScreens(processScreens)).toThrow(
+      '`processScreens` must not add, remove, rename, or duplicate screens.'
+    );
+  }
+);
+
+it('allows processScreens to update screen options', () => {
+  expect(() =>
+    renderProcessedScreens((screens) =>
+      screens.map((screen) => {
+        if (
+          typeof screen.options !== 'function' &&
+          screen.options &&
+          'customOption' in screen.options
+        ) {
+          return screen;
+        }
+        return { ...screen, options: { customOption: true } };
+      })
+    )
+  ).not.toThrow();
+});
+
+it('allows processScreens to reorder screens', () => {
+  expect(() => renderProcessedScreens((screens) => [...screens].reverse())).not.toThrow();
+});
+
+it('preserves the fallback for a nullish processScreens result', () => {
+  expect(() => renderProcessedScreens(() => undefined as never)).not.toThrow();
 });
--- a/packages/expo-router/src/layouts/withLayoutContext.tsx
+++ b/packages/expo-router/src/layouts/withLayoutContext.tsx
@@ -19,7 +19,11 @@ import { isScreen, Screen } from '../views/Screen';
 import { GuardContextProvider, normalizeRouteName, type GuardedRedirects } from './GuardContext';
 import { IsWithinLayoutContext } from './IsWithinLayoutContext';
 
-export function useFilterScreenChildren(
+export function useFilterScreenChildren<
+  TOptions extends object = Record<string, any>,
+  TState extends NavigationState = NavigationState,
+  TEventMap extends EventMapBase = EventMapBase,
+>(
   children: ReactNode,
   {
     isCustomNavigator,
@@ -33,12 +37,12 @@ export function useFilterScreenChildren(
   return useMemo(() => {
     const customChildren: any[] = [];
 
-    const screens: (ScreenProps & { name: string })[] = [];
+    const screens: (ScreenProps<TOptions, TState, TEventMap> & { name: string })[] = [];
     const guardedRedirects: GuardedRedirects = new Map();
 
     function flattenChild(child: ReactNode, exclude = false, redirectTo?: Href) {
       if (isScreen(child, contextKey)) {
-        screens.push(child.props);
+        screens.push(child.props as ScreenProps<TOptions, TState, TEventMap> & { name: string });
         if (exclude) {
           guardedRedirects.set(child.props.name, redirectTo);
         }
@@ -50,7 +54,7 @@ export function useFilterScreenChildren(
         screens.push({
           ...child.props,
           options: exclude ? { ...options, hidden: true } : options,
-        } as ScreenProps & { name: string });
+        } as ScreenProps<TOptions, TState, TEventMap> & { name: string });
         if (exclude) {
           guardedRedirects.set(child.props.name, redirectTo);
         }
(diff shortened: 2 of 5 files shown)


packages/expo-router/src/layouts/withLayoutContext.tsx
```diff
@@ -142,19 +146,45 @@ export function withLayoutContext<
   T extends ComponentType<any>,
   TState extends NavigationState,
   TEventMap extends EventMapBase,
->(Nav: T, processor?: (options: ScreenProps[]) => ScreenProps[]) {
+>(
+  Nav: T,
+  processScreens?: (
+    screens: (ScreenProps<TOptions, TState, TEventMap> & { name: string })[]
```
jakub-agent, line 154: [🚨 **critical**] `TOptions` is declared as `TOptions extends object` with no default. Before this PR, `processor` took `ScreenProps[]`, which resolves to `ScreenProps<Record<string, any>>`. Now the parameter type mentions `TOptions`, and when a caller omits explicit type arguments TypeScript cannot infer it from the callback, so it falls back to the constraint `object`. Any property read on `screen.options` then fails to compile.

Reduced repro, checked with the repo's `tsc --strict`:

```ts
withLayoutContext(Nav, (screens) => screens.map((s) => {
  if (typeof s.options !== 'function') { const x = s.options?.tabBarBadge; }
  //                                                        ^ TS2339: Property 'tabBarBadge'
  //                                                          does not exist on type 'object'.
  return s;
}));
```

The same code compiles against `main`. Changing the declaration to `TOptions extends Record<string, any> = Record<string, any>` makes the repro pass again, and it matches both `ScreenProps`' own default and the default you added to `useFilterScreenChildren` on line 23.

packages/expo-router/src/layouts/withLayoutContext.tsx
```diff
@@ -142,19 +146,45 @@ export function withLayoutContext<
...
+          ? new Set(screens.map(({ name }) => name))
+          : undefined;
+
+      const processed = processScreens ? processScreens(screens) : screens;
+
+      if (screenNames) {
+        const processedNames = processed.map(({ name }) => name);
+        const uniqueProcessedNames = new Set(processedNames);
+        if (
+          uniqueProcessedNames.size !== screenNames.size ||
```
jakub-agent: [🏗️ **design**] This check does not detect duplicates, although the error message claims it does. `processScreens` returning `['a', 'a', 'b']` from `['a', 'b']`:

- `uniqueProcessedNames.size` is 2, `screenNames.size` is 2 → equal
- every processed name is in `screenNames` → true

So it passes, and `getSortedChildren` then logs `No route named "a" exists in nested children` for the second copy.

Comparing the array length instead of the deduplicated size covers add, remove, rename, and duplicate in one condition, and lets `uniqueProcessedNames` go away:

```ts
if (
  processedNames.length !== screenNames.size ||
  !processedNames.every((name) => screenNames.has(name))
) {
```

packages/expo-router/src/layouts/withLayoutContext.tsx
```diff
@@ -142,19 +146,45 @@ export function withLayoutContext<
...
+          !processedNames.every((name) => screenNames.has(name))
+        ) {
+          throw new Error(
+            '`processScreens` must not add, remove, rename, or duplicate screens. ' +
```
jakub-agent, line 182: [🏗️ **design**] This adds a restriction that did not exist before. The second parameter was documented only as "A function that processes the screens before passing them to the navigator", so a user who filters or appends screens has working code today and will now hit a render-time throw in dev.

Three follow-ups:

- The changelog entry is under **🎉 New features**. A new hard error on a previously accepted usage belongs in **🐛 Bug fixes** or **💡 Others** with an explicit note about the restriction.
- The JSDoc above `withLayoutContext` still describes `processScreens` generically. Document the invariant where users will read it, and say why it exists — `guardedRedirects` is keyed off the original children, so a changed screen set desynchronizes guards.
- There is no test. `packages/expo-router/src/__tests__/custom-navigators.test.ios.tsx` has a single test and would be a good home for one case per violation (add, remove, rename, duplicate) plus one identity-mapping case that must not throw.

packages/expo-router/src/layouts/withLayoutContext.tsx
```diff
@@ -142,19 +146,45 @@ export function withLayoutContext<
...
+        { contextKey }
+      );
 
-      const processed = processor ? processor(screens ?? []) : screens;
+      const screenNames =
+        processScreens && process.env.NODE_ENV !== 'production'
+          ? new Set(screens.map(({ name }) => name))
+          : undefined;
+
+      const processed = processScreens ? processScreens(screens) : screens;
```
jakub-agent: [🐧 nit] Dropping the `?? []` around `processed` makes dev and production diverge when `processScreens` returns a non-array. A JS consumer whose callback forgets a `return` now gets `TypeError: Cannot read properties of undefined (reading 'map')` from the validation block in dev, while in production `screenNames` is `undefined`, the block is skipped, `getSortedChildren`'s `order = []` default kicks in, and the layout silently renders in filesystem order.

Since you are already validating the return value here, a `!Array.isArray(processed)` branch inside the dev check would report the real problem instead.

jakub-agent (changes requested):

Threading `TOptions`/`TState`/`TEventMap` through `ScreenProps` is the right direction, and removing the `@ts-expect-error` on `tabBarButton` in `TabsClient.tsx` is good proof that it works.

Two things block a clean merge:

1. **Public type regression.** `withLayoutContext` declares `TOptions extends object` with no default. When a caller passes `processScreens` without explicit type arguments, `TOptions` falls back to `object`, so `screen.options?.<anything>` no longer compiles. Before this PR the parameter was `ScreenProps[]` (`TOptions = Record<string, any>`), so that code did compile. Verified with `tsc --strict` on a reduced repro. Adding `= Record<string, any>` restores it.
2. **`processScreens` validation does not catch duplicates**, even though the error message says it does. `['a','a','b']` derived from `['a','b']` passes the check.

Smaller points, in inline comments: the new invariant is a behavior change for existing `processor` users but is filed under "New features" and is not documented in the JSDoc, and there is no test for it. `packages/expo-router/src/__tests__/custom-navigators.test.ios.tsx` is the natural place — it currently has a single test.

Also: the PR description is still the unfilled template. The repo convention is to record the motivation and a test plan, and this PR adds a runtime error whose rationale is not obvious from the diff.

packages/expo-router/src/layouts/withLayoutContext.tsx
Ubax: New features is a correct place for the entry. I'll deprecate the public `withLayoutContext` soon

packages/expo-router/src/layouts/withLayoutContext.tsx
Ubax: Just don't drop `?? []`

packages/expo-router/src/__tests__/custom-navigators.test.ios.tsx
```diff
@@ -35,3 +56,43 @@ it('can render a custom navigator', () => {
...
+);
+
+it('allows processScreens to update screen options', () => {
+  expect(() =>
+    renderProcessedScreens((screens) =>
+      screens.map((screen) => {
+        if (
+          typeof screen.options !== 'function' &&
+          screen.options &&
+          'customOption' in screen.options
```
jakub-agent, line 82: [🐧 nit] This compiles, so nothing to fix — but worth knowing the `in` check is here for a reason that hides what the test is meant to prove. `Parameters<typeof withLayoutContext>[1]` instantiates the generic with each parameter's **constraint**, not its default, and the constraint is still `TOptions extends object`. So `screen.options` is `object` here, `screen.options?.customOption` does not compile, and `'customOption' in screen.options` is the workaround.

A real caller never has to write that. I confirmed plain property access compiles at this commit when generics are omitted, because the *default* (`Record<string, any>`) applies to inference. So this test passes while exercising `TOptions = object` — not the ergonomics the PR body lists as covered.

Either narrowing the constraint to `TOptions extends Record<string, any>` (matches `ScreenProps`' own constraint, and lets the plain access compile here) or typing the helper from a concrete instantiation would make the test mirror real usage. Purely about test fidelity — the shipped behavior is right.

packages/expo-router/src/layouts/withLayoutContext.tsx
```diff
@@ -138,21 +144,47 @@ export function useFilterScreenChildren(
  * ```
  */
 export function withLayoutContext<
-  TOptions extends object,
-  T extends ComponentType<any>,
-  TState extends NavigationState,
-  TEventMap extends EventMapBase,
->(Nav: T, processor?: (options: ScreenProps[]) => ScreenProps[]) {
+  TOptions extends object = Record<string, any>,
+  T extends ComponentType<any> = ComponentType<any>,
```
jakub-agent, line 148: [🐧 nit] Defaulting `TOptions` forces a default on every parameter after it, so this one is not really optional — recording the cost rather than asking for a change. `withLayoutContext<MyOptions>(Nav)` used to be a hard error ("Expected 4 type arguments, but got 1"); now `T` silently becomes `ComponentType<any>` and the wrapper accepts anything. Checked against the package at this commit:

```ts
const c = withLayoutContext<MyOptions>(Nav);
const bad: ComponentProps<typeof c> = { totallyUnknownProp: 1 }; // compiles
```

`T extends ComponentType<any> = never` turns that back into a compile error while leaving full inference untouched, at the cost of an opaque message. Given the planned deprecation, leaving it is defensible — flagging it so the choice is deliberate.

jakub-agent (approved):

**Iteration 3** — re-reviewed at `ee9c39e`. Everything I raised is fixed or consciously declined. Approving.

Verified locally in a worktree at this commit, after deleting `.tsbuildinfo` so `tsc -b` could not skip the project:

| Check | Result |
| - | - |
| `tsc -b tsconfig.all.json` | 0 errors, exit 0 |
| `oxlint` | clean |
| `custom-navigators.test.ios.tsx` | 9/9 pass |
| `src/layouts` + `protected` + `headless-tabs` + `custom-navigators` | 46 suites, 705 tests, 9 snapshots, all pass |

I also confirmed the thing the type work was for. A caller that omits generics can now write plain property access, which is what needed the `@ts-expect-error` before:

```ts
withLayoutContext(Nav, (screens) => screens.map((screen) => {
  if (typeof screen.options !== 'function' && screen.options?.customOption) return screen;
  return { ...screen, options: { customOption: true } };
}));
// compiles at ee9c39e
```

Running tally across the review:

- ✅ `TOptions` default — real callers keep property access on `screen.options`
- ✅ duplicate detection — all seven transformation shapes behave correctly
- ✅ JSDoc documents the invariant and why it exists
- ✅ tests — 9 cases, one per violation plus the three permitted transformations
- ✅ test-file type errors — fixed at `ee9c39e`
- ✅ PR body — Why / How / Test Plan filled in
- declined, your call: changelog placement under New features, given the planned deprecation
- deferred, your call: the `tsc -b` / `.tsbuildinfo` blind spot in `pnpm typecheck`, going into a separate PR

Two nits left inline. Both are fine to ignore — neither affects behavior, and I verified the shipped behavior is correct in both cases.

PR #48367, opened 2026-07-31
[fingerprint] Hoist Minimatch build out of ExpoConfigLoader per-module filter

alfonsocj:
# Why

`ExpoConfigLoader.resolveLoadedModuleSourcesAsync` calls `isIgnoredPath()` once per captured module. `isIgnoredPath()` rebuilds every `Minimatch` on each call, so the filter is O(modules × patterns) in Minimatch **constructions**, not just matches.

On a large monorepo (~600 modules × ~4200 ignore patterns) that is ~2.5M constructions — measured at ~37s of a ~79s `native-hash` run. The cost was previously invisible because `getExpoConfigAsync` is not wrapped in Expo's `profile()` helper.

# How

Hoist `buildPathMatchObjects()` out of the per-module filter and use `isIgnoredPathWithMatchObjects()` (already exported from `utils/Path`) per module. The matchers are built once per call instead of once per module.

Performance only — the resulting fingerprint hash is byte-identical, verified with an explicit A/B on a large project.

# Test Plan

- Added `should build Minimatch objects once per call, not once per module` to `ExpoConfigLoader-test.ts`. It spies on the `Minimatch` constructor via `jest.mock('minimatch')` and asserts it is called exactly `ignoredPaths.length` times for 20 modules × 8 patterns. Without the fix the spy records 160 constructions (20 × 8); with the fix it records 8.
- `pnpm -F @expo/fingerprint test` — 214 passed.
- `et check-packages @expo/fingerprint` — build, lint, format, typecheck all pass.
- End-to-end on a large monorepo: `native-hash` 1:19 → 0:41; hash unchanged.

# Checklist

### DIFF
--- a/packages/@expo/fingerprint/CHANGELOG.md
+++ b/packages/@expo/fingerprint/CHANGELOG.md
@@ -17,6 +17,7 @@
 
 ### 💡 Others
 
+- Hoisted the ignore-path Minimatch build out of the per-module filter in `ExpoConfigLoader`, avoiding O(modules × patterns) Minimatch constructions during config load. Byte-identical fingerprint hash. ([#48367](https://github.com/expo/expo/pull/48367) by [@alfonsocj](https://github.com/alfonsocj))
 - Reworked config-plugin module capture with a `Module.prototype._compile` hook. ([#47666](https://github.com/expo/expo/pull/47666) by [@kudo](https://github.com/kudo))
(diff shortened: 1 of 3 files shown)


zoontek:
@alfonsocj Could you update the changelog entry with the PR number?

alfonsocj:
@zoontek done!

zoontek (approved):

PR #48369, opened 2026-07-31
[router] Remove Stack `redirect` property

Ubax:
# Why

The prop was `deprecated` and removed in https://github.com/expo/expo/pull/24219, but left in `ScreenProps` in `useScreens`

Note for review: first commit is the change. Second is formatting change

# How

Remove the prop

# Test Plan

CI

# Checklist

### DIFF
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### 🛠 Breaking changes
 
-- Remove the `redirect` prop from layout `<Screen>` components.
+- Remove the `redirect` prop from layout `<Screen>` components. ([#48369](https://github.com/expo/expo/pull/48369) by [@Ubax](https://github.com/Ubax))
 - Add `redirectTo` to protected routes and render guarded screens as redirects instead of removing them from navigators. ([#47744](https://github.com/expo/expo/pull/47744) by [@Ubax](https://github.com/Ubax))
 - Migrate the `Drawer` navigator to the `standard-navigation` integration. ([#47839](https://github.com/expo/expo/pull/47839) by [@Ubax](https://github.com/Ubax))
 - Remove `StackNavigationState.preloadedRoutes` property and append preloaded routes to `state.routes` ([#47961](https://github.com/expo/expo/pull/47961) by [@Ubax](https://github.com/Ubax))
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### 🛠 Breaking changes
 
+- Remove the `redirect` prop from layout `<Screen>` components.
 - Add `redirectTo` to protected routes and render guarded screens as redirects instead of removing them from navigators. ([#47744](https://github.com/expo/expo/pull/47744) by [@Ubax](https://github.com/Ubax))
 - Migrate the `Drawer` navigator to the `standard-navigation` integration. ([#47839](https://github.com/expo/expo/pull/47839) by [@Ubax](https://github.com/Ubax))
 - Remove `StackNavigationState.preloadedRoutes` property and append preloaded routes to `state.routes` ([#47961](https://github.com/expo/expo/pull/47961) by [@Ubax](https://github.com/Ubax))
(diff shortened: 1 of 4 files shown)


kitten (approved):

packages/expo-router/src/useScreens.tsx
```diff
@@ -93,72 +88,50 @@ function getSortedChildren(
...
-      }) => {
-        if (!entries.length) {
-          console.warn(
-            `[Layout children]: Too many screens defined. Route "${name}" is extraneous.`
-          );
-          return null;
-        }
-        const matchIndex = entries.findIndex(
-          (child) => child.route === name || child.route === `${name}/index`
+    .map(({ name, initialParams, listeners, options, getId, dangerouslySingular: singular }) => {
```
jakub-agent, line 99: [🏗️ **design**] The prop now becomes an unknown prop that is silently dropped, and the behavior change is not obvious in an app: a screen that used to be excluded from its navigator now renders. In a `Tabs` layout that means a previously hidden route gains a tab bar item — a visible regression with nothing pointing at the cause.

Since `redirect` was still functional in this code path until now (#24219 only removed it from standalone `<Screen />`) and is still listed in the SDK 55/56 API docs, some apps are likely to hit this. Consider a dev-only warning, following the `getId` deprecation pattern a few lines below:

```ts
.map((screenProps) => {
  const { name, initialParams, listeners, options, getId, dangerouslySingular: singular } = screenProps;
  if ('redirect' in screenProps) {
    console.warn(
      `The 'redirect' prop on screen ${name} has been removed. Use <Redirect>, the 'redirects' config, or <Protected redirectTo> instead.`
    );
  }
```

That turns a silent layout change into an actionable message. Happy to hear the counter-argument if the team considers the prop dead enough already that the warning is not worth the extra branch.

packages/expo-router/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### 🛠 Breaking changes
 
+- Remove the `redirect` prop from layout `<Screen>` components. ([#48369](https://github.com/expo/expo/pull/48369) by [@Ubax](https://github.com/Ubax))
```
jakub-agent, line 7: [💡 suggestion] Two docs follow-ups for a breaking removal:

1. The entry states the removal but not the replacement. Adding a short pointer (`<Redirect>`, the `redirects` config, or `<Protected redirectTo>`) makes it actionable for anyone reading the breaking-changes list during an upgrade.
2. `docs/public/static/data/unversioned/expo-router.json` still documents `ScreenProps.redirect`, so the unversioned API reference keeps advertising the removed prop until it is regenerated (`et generate-docs-api-data -p expo-router`). Hand-written docs pages are clean — this is the only stale reference I found.

packages/expo-router/src/useScreens.tsx
```diff
@@ -93,72 +88,50 @@ function getSortedChildren(
...
+      );
+      if (matchIndex === -1) {
+        console.warn(
+          `[Layout children]: No route named "${name}" exists in nested children:`,
+          children.map(({ route }) => route)
         );
-        if (matchIndex === -1) {
+        return null;
+      } else {
+        // Get match and remove from entries
```
jakub-agent, line 114: [🐧 nit] Nit: the enclosing `else` is redundant — the `if (matchIndex === -1)` branch above always returns. Since this whole block is already being reindented by the formatting commit, dropping the `else` and unindenting the body costs nothing extra in diff noise and removes a nesting level.

jakub-agent (commented):

The removal itself is clean and complete: no `redirect` references are left anywhere in `packages/expo-router/src`, `yarn build` (tsc) passes, and `src/__tests__/stacks.test.ios.tsx` passes (22/22) at this commit. The new negative type assertions plus the runtime test are a good way to lock the removal in.

One framing correction worth noting: #24219 removed `redirect` from the standalone `<Screen />` component only — the layout path in `getSortedChildren` kept working right up to this PR. So this is a live behavior removal, not a type-only cleanup. The CHANGELOG entry says this correctly; the PR description reads as if only a stale type were left.

That matters because the removal is silent. A screen previously excluded from its navigator now renders, so a `Tabs` layout that used `redirect` to hide a route will gain a visible tab. The main comment below suggests a dev-only warning to make that failure loud instead of silent.

PR #48374, opened 2026-07-31
[router] Order JS Tabs, NativeTabs, headless tabs and Drawer by `routeNames` order

Ubax:
# Why

In order to remove the dependency of tabs on `state.routes` order, we can order them by `routeNames` order - the order of triggers. This way we don't need to adjust the `state.routes` on purely order change in `routeNames`

# How

1. Order tabs by routeNames in `useVisibleTabsWithRedirect` - hook used by all router tab navigators
2. Refactor headless tabs not to relay on routeNames changes

# Test Plan

CI

# Checklist

### DIFF
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -38,6 +38,7 @@
 
 ### 💡 Others
 
+- Order JS Tabs, NativeTabs, headless tabs and Drawer by `routeNames` order ([#48374](https://github.com/expo/expo/pull/48374) by [@Ubax](https://github.com/Ubax))
 - Integrate native stack with standard navigation ([#48114](https://github.com/expo/expo/pull/48114) by [@Ubax](https://github.com/Ubax))
 - Refactor native-stack for compliance with standard-navigation integration ([#46592](https://github.com/expo/expo/pull/46592) by [@Ubax](https://github.com/Ubax))
 - Remove the deprecated static-navigation API from `expo-router/react-navigation`. ([#48071](https://github.com/expo/expo/pull/48071) by [@Ubax](https://github.com/Ubax))
--- a/packages/expo-router/src/native-tabs/NativeBottomTabsNavigator.tsx
+++ b/packages/expo-router/src/native-tabs/NativeBottomTabsNavigator.tsx
@@ -32,6 +32,7 @@ export const NativeTabsContext = React.createContext<boolean>(false);
 
 function NativeTabsContent({
   state,
+  routeNames,
   descriptors,
   actions,
   emitter,
@@ -53,7 +54,7 @@ function NativeTabsContent({
   NativeTabOptions,
   NativeTabNavigationEventMap,
   Omit<InternalNativeTabsProps, 'screenListeners'>
->) {
+> & { routeNames: string[] }) {
   if (use(NativeTabsContext)) {
     throw new Error(
       'Nesting Native Tabs inside each other is not supported natively. Use JS tabs for nesting instead.'
@@ -64,6 +65,7 @@ function NativeTabsContent({
 
   const { visibleRoutes, focusedIndex } = useVisibleTabsWithRedirect({
     routes,
+    routeNames,
     focusedRouteKey: routes[state.index]!.key,
     descriptors,
   });
@@ -166,8 +168,11 @@ const NativeTabsNavigatorWithContext = unstable_createStandardRouterNavigator<
   TabNavigationState<ParamListBase>,
   NativeTabNavigationEventMap,
   Omit<InternalNativeTabsProps, 'screenListeners'>,
-  TabRouterOptions
->(NativeTabsContent, NativeBottomTabsRouter);
+  TabRouterOptions,
+  { routeNames: string[] }
+>(NativeTabsContent, NativeBottomTabsRouter, {
+  createProps: ({ state }) => ({ routeNames: state.routeNames }),
+});
 
 export function NativeTabsNavigatorWrapper(props: NativeTabsProps) {
   const routeNode = useRouteNode();
--- a/packages/expo-router/src/react-navigation/routers/TabRouter.tsx
+++ b/packages/expo-router/src/react-navigation/routers/TabRouter.tsx
@@ -1,5 +1,6 @@
 import { nanoid } from 'nanoid/non-secure';
 
+import { orderRoutesByRouteNames } from '../../utils/orderRoutesByRouteNames';
 import { BaseRouter } from './BaseRouter';
 import { createParamsFromAction } from './createParamsFromAction';
 import type {
@@ -198,7 +199,9 @@ const changeIndex = (
       params: backBehavior === 'fullHistory' ? currentRoute.params : undefined,
     });
   } else {
-    history = getRouteHistory(state.routes, index, backBehavior, initialRouteName);
+    const routes = orderRoutesByRouteNames(state.routes, state.routeNames);
+    const orderedIndex = routes.findIndex((route) => route.key === state.routes[index]!.key);
+    history = getRouteHistory(routes, orderedIndex, backBehavior, initialRouteName);
   }
 
   return {
@@ -308,24 +311,57 @@ export function TabRouter({
     },
 
     getStateForRouteNamesChange(state, { routeNames, routeParamList, routeKeyChanges }) {
-      const routes = routeNames.map(
-        (name) =>
-          state.routes.find((r) => r.name === name && !routeKeyChanges.includes(r.name)) || {
+      const routes = state.routes
+        .filter((route) => routeNames.includes(route.name))
+        .map((route) =>
+          routeKeyChanges.includes(route.name)
+            ? {
+                name: route.name,
+                key: `${route.name}-${nanoid()}`,
+                params: routeParamList[route.name],
+              }
+            : route
+        );
+
+      for (const name of routeNames) {
+        if (!routes.some((route) => route.name === name)) {
+          routes.push({
             name,
             key: `${name}-${nanoid()}`,
             params: routeParamList[name],
-          }
-      );
+          });
+        }
+      }
 
-      const index = Math.max(0, routeNames.indexOf(state.routes[state.index]!.name));
+      const focusedName = state.routes[state.index]!.name;
+      const index = Math.max(
+        0,
+        routes.findIndex((route) => route.name === focusedName)
+      );
 
       let history = state.history.filter(
         // Type will always be 'route' for tabs, but could be different in a router extending this (e.g. drawer)
         (it) => it.type !== 'route' || routes.find((r) => r.key === it.key)
       );
 
-      if (!history.length) {
-        history = getRouteHistory(routes, index, backBehavior, initialRouteName);
+      // Static back behaviors follow display order, while history behaviors retain valid visits.
+      // Preserve non-route entries added by extending routers such as `DrawerRouter`.
+      if (
+        backBehavior === 'firstRoute' ||
+        backBehavior === 'initialRoute' ||
+        backBehavior === 'order'
+      ) {
+        const orderedRoutes = orderRoutesByRouteNames(routes, routeNames);
+        const orderedIndex = orderedRoutes.findIndex((route) => route.key === routes[index]!.key);
+        history = [
+          ...getRouteHistory(orderedRoutes, orderedIndex, backBehavior, initialRouteName),
+          ...history.filter((item) => item.type !== 'route'),
+        ];
+      } else if (!history.some((item) => item.type === 'route')) {
+        history = [
+          ...getRouteHistory(routes, index, backBehavior, initialRouteName),
+          ...history.filter((item) => item.type !== 'route'),
+        ];
       }
 
       return {
--- a/packages/expo-router/src/ui/TabTrigger.tsx
+++ b/packages/expo-router/src/ui/TabTrigger.tsx
@@ -125,7 +125,7 @@ export type Trigger = TriggerMap[string] & {
   hidden: boolean;
   isFocused: boolean;
   resolvedHref: string;
-  route: TabNavigationState<any>['routes'][number];
+  route?: TabNavigationState<any>['routes'][number];
 };
 
 export type UseTabTriggerResult = {
@@ -145,7 +145,7 @@ export type TriggerProps = {
  * Utility hook creating custom `TabTrigger`.
  */
 export function useTabTrigger(options: TabTriggerProps): UseTabTriggerResult {
-  const { state, navigation, descriptors } = useNavigatorContext();
+  const { state, navigation, contextKey, descriptors } = useNavigatorContext();
   const { name, resetOnFocus, onPress, onLongPress } = options;
   const triggerMap = use(TabTriggerMapContext);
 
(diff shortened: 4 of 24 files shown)


packages/expo-router/src/react-navigation/routers/TabRouter.tsx
```diff
@@ -308,24 +311,55 @@ export function TabRouter({
...
             key: `${name}-${nanoid()}`,
             params: routeParamList[name],
-          }
-      );
+          });
+        }
+      }
 
-      const index = Math.max(0, routeNames.indexOf(state.routes[state.index]!.name));
+      const focusedKey = state.routes[state.index]!.key;
```
jakub-agent: [🚨 **critical**] Matching the focused route by **key** loses focus whenever that route's key was just regenerated.

A few lines above, any route whose name is in `routeKeyChanges` is rebuilt with a fresh `key`. If that is the focused route, `routes.findIndex((route) => route.key === focusedKey)` returns `-1`, and `Math.max(0, -1)` moves focus to the first tab. The previous implementation resolved it by name (`routeNames.indexOf(state.routes[state.index]!.name)`), which survived key changes.

`routeKeyChanges` is populated when a screen's `getId` result changes, so this is reachable from userland: a tab with `getId` whose id depends on a param resets to tab 0 on the next id change.

Reproduction of the two index computations, three tabs focused on `baz` with `routeKeyChanges: ['baz']`:

```
OLD: { routes: ['bar','baz','qux'], index: 1, focusedName: 'baz' }
NEW: { routes: ['bar','baz','qux'], index: 0, focusedName: 'bar' }
```

Since tab route names are unique, matching by name is enough and stays key-change-proof:

```ts
const focusedName = state.routes[state.index]!.name;
const index = Math.max(
  0,
  routes.findIndex((route) => route.name === focusedName)
);
```

Worth noting for the test that catches this: `TabRouter.test.tsx` mocks `nanoid` as `() => 'test'`, so a regenerated key stringifies to the same value as the original and the bug stays invisible. All four `routeKeyChanges` in that file are `[]` today. A test needs a distinguishing mock (for example an incrementing counter) plus a non-empty `routeKeyChanges` naming the focused route.

packages/expo-router/src/react-navigation/bottom-tabs/views/BottomTabBar.tsx
```diff
@@ -237,7 +238,10 @@ export function BottomTabBar({ state, navigation, descriptors, insets, style }:
     });
   };
 
-  const { routes } = state;
+  const routes = React.useMemo(
+    () => orderRoutesByRouteNames(state.routes, state.routeNames),
```
jakub-agent: [🏗️ **design**] Ordering here fixes the default tab bar, but a custom `tabBar` still gets the raw state. `BottomTabView` calls `tabBar({ state, descriptors, navigation, insets })` with `state` untouched, so any userland tab bar that maps `state.routes` renders in mount order and its `index === state.index` focus check keeps working — which means it silently disagrees with the declared order.

That is inconsistent with the sibling navigator: `MaterialTopTabView` in this same PR builds `orderedState` and passes it to `tabBar`. Doing the same in `BottomTabView` would give one rule for both — the state handed to any tab bar is display-ordered — and would let `BottomTabBar` drop its own memo.

If leaving custom tab bars on raw state is deliberate, the asymmetry with material-top-tabs is worth a comment, since a reader of `BottomTabBar` cannot tell which contract applies.

packages/expo-router/src/react-navigation/material-top-tabs/views/MaterialTopTabView.tsx
```diff
@@ -44,6 +47,17 @@ export function MaterialTopTabView({
 }: Props) {
   const { colors } = useTheme();
   const { direction } = useLocale();
+  const orderedState = useMemo(() => {
+    const routes = orderRoutesByRouteNames(state.routes, state.routeNames);
+    if (routes === state.routes) {
+      return state;
+    }
+    const index = Math.max(
```
jakub-agent: [💡 suggestion] The `Math.max(0, ...)` fallback lets the view disagree with the router instead of failing loudly. If the focused route is absent from `routeNames`, `orderRoutesByRouteNames` drops it, `findIndex` returns `-1`, and `orderedState.index` becomes `0` — so `TabView` renders the first tab while react-navigation still considers a different route focused. The focused screen keeps rendering via `descriptors[route.key].render()`, just never visible.

Returning the unmodified `state` when the focused route is not found keeps the view and the router in agreement:

```ts
const index = routes.findIndex((route) => route.key === state.routes[state.index]!.key);
if (index === -1) {
  return state;
}
return { ...state, routes, index };
```

This is different from the `Math.max(0, ...)` in `TabRouter.getStateForRouteNamesChange`, where falling back to the first route is the intended behavior for a removed tab. Here there is nothing to recover from — the state is simply inconsistent.

packages/expo-router/src/utils/orderRoutesByRouteNames.ts
```diff
@@ -0,0 +1,14 @@
+export function orderRoutesByRouteNames<Route extends { name: string }>(
```
jakub-agent, line 4: [💡 suggestion] This helper has two contracts that callers depend on and that the signature does not convey:

1. Routes whose name is missing from `routeNames` are **dropped**, not appended. In `BottomTabBar` that silently removes a tab from the bar (and skews the `N of M` accessibility label); in `MaterialTopTabView` it drops a pager scene.
2. It returns the **same array reference** when the order already matches. `MaterialTopTabView` relies on this (`if (routes === state.routes) return state;`) to avoid rebuilding state on every render.

A short doc comment on both would help, since the second is the kind of thing a future refactor breaks by accident — for example by always building a fresh array. The unit test does pin it with `toBe`, which is good; naming it in the source makes the intent explicit rather than only asserted.

packages/expo-router/src/native-tabs/NativeBottomTabsNavigator.tsx
```diff
@@ -54,7 +55,7 @@ function NativeTabsContent({
   NativeTabOptions,
   NativeTabNavigationEventMap,
   Omit<InternalNativeTabsProps, 'screenListeners'>
->) {
+> & { routeNames: string[] }) {
```
jakub-agent, line 57: [🐧 nit] `NavigatorContentProps` already has a `CreateProps` slot for exactly this, and its doc comment is explicit that content components should be annotated with it. Hand-intersecting `StandardNavigatorContentProps<...> & { routeNames: string[] }` bypasses the inference carrier, so `CreateProps` can no longer be read back from the component and the call site below has to spell out all six generics.

```ts
}: NavigatorContentProps<
  NativeTabOptions,
  NativeTabNavigationEventMap,
  Omit<InternalNativeTabsProps, 'screenListeners'>,
  { routeNames: string[] }
>) {
```

Passing `routeNames` through `createProps` is the right call — `useStandardState` only projects `index` and `routes`, so `routeNames` genuinely is not reachable from the standard-navigation `state`.

jakub-agent (changes requested):

packages/expo-router/src/react-navigation/routers/TabRouter.tsx
jakub-agent: [🚨 **critical**] Unchanged from round 1 — restating so it isn't lost across the rebase.

Any route whose name is in `routeKeyChanges` is rebuilt a few lines above with a fresh `key`. If that is the focused route, `routes.findIndex((route) => route.key === focusedKey)` returns `-1` and `Math.max(0, -1)` moves focus to the first tab. The code this replaced resolved it by name, which survived key regeneration.

`routeKeyChanges` is populated when a screen's `getId` result changes, so this is reachable from userland: a tab with `getId` keyed on a param resets to tab 0 on the next id change.

Both index computations, three tabs focused on `baz`, `routeKeyChanges: ['baz']`:

```
OLD: { routes: ['bar','baz','qux'], index: 1, focusedName: 'baz' }
NEW: { routes: ['bar','baz','qux'], index: 0, focusedName: 'bar' }
```

Tab route names are unique, so matching by name is sufficient and key-change-proof:

```ts
const focusedName = state.routes[state.index]!.name;
const index = Math.max(
  0,
  routes.findIndex((route) => route.name === focusedName)
);
```

On covering it: `TabRouter.test.tsx` mocks `nanoid` as `() => 'test'`, so a regenerated key stringifies identically to the original and the bug stays invisible. All `routeKeyChanges` in that file are still `[]`. A test needs a distinguishing mock (an incrementing counter) plus a non-empty `routeKeyChanges` naming the focused route.

packages/expo-router/src/react-navigation/bottom-tabs/views/BottomTabView.tsx
```diff
@@ -78,6 +81,11 @@ export function BottomTabView(props: Props) {
   } = props;
 
   const focusedRouteKey = state.routes[state.index]!.key;
+  const tabBarState = React.useMemo(() => {
+    const routes = orderRoutesByRouteNames(state.routes, routeNames);
+    const index = routes.findIndex((route) => route.key === focusedRouteKey);
```
jakub-agent: [🏗️ **design**] This is now the third copy of "order the routes, then re-find the focused index", and each copy picks a different policy for a focused route that is missing from `routeNames`:

| Site | Missing-route behavior |
|---|---|
| `BottomTabView` (here) | `index: -1`, unguarded |
| `MaterialTopTabView:55` | `Math.max(0, ...)` → first tab |
| `TabRouter.getStateForRouteNamesChange:337` | `Math.max(0, ...)` → first tab |

The `-1` here is the one that bites. `orderRoutesByRouteNames` drops routes absent from `routeNames`, so the identity shortcut (`routes === state.routes`) does *not* protect this path — a dropped route is exactly the case where the shortcut misses and `findIndex` returns `-1`. `BottomTabBar` then does `descriptors[state.routes[state.index]!.key]` at line 40; `state.routes[-1]` is `undefined` and the non-null assertion is compile-time only, so it throws a `TypeError` rather than showing the wrong tab.

Today `routeNames` comes from `createProps: ({ state }) => ({ routeNames: state.routeNames })`, so `state.routes ⊆ state.routeNames` holds and this is unreachable. That makes it latent rather than live — but it is latent in the one copy that fails hardest, which is an argument for not having three copies.

Having `orderRoutesByRouteNames` return the focused index alongside the routes would decide the policy once and let all three sites share it:

```ts
const { routes, focusedIndex } = orderRoutesByRouteNames(state.routes, routeNames, focusedRouteKey);
```

packages/expo-router/src/ui/TabTrigger.tsx
```diff
@@ -120,7 +120,7 @@ export type SwitchToOptions = {
 export type Trigger = TriggerMap[string] & {
   isFocused: boolean;
   resolvedHref: string;
-  route: TabNavigationState<any>['routes'][number];
+  route?: TabNavigationState<any>['routes'][number];
```
jakub-agent, line 128: [🏗️ **design**] `Trigger` is public API — `expo-router/ui` → `export * from './Tabs'` → `export * from './TabTrigger'` — and `useTabTrigger` is documented as the hook for building a custom `TabTrigger`. Making `route` optional is therefore a breaking type change: any consumer reading `trigger.route.key` (the shape this file itself used until this commit) stops compiling.

The call sites here were updated to `trigger.route?.key ?? trigger.href`, so the runtime behavior is deliberate — an external trigger has no route. But that widening needs to reach users: a `🛠 Breaking changes` CHANGELOG entry, and ideally a note on what to do instead.

Flagging with the caveat that this looks like it belongs to the headless-tabs refactor rather than to tab ordering (see the summary on the base branch). If it moves back to #48257, the changelog note should move with it.

jakub-agent (changes requested):

packages/expo-router/CHANGELOG.md
```diff
@@ -35,6 +35,7 @@
 
 ### 💡 Others
 
+- Refactor headless tabs not to rely on `routeNames` ([#48257](https://github.com/expo/expo/pull/48257) by [@Ubax](https://github.com/Ubax))
```
Ubax: Update the entry to reflect what the PR is actually changing

```suggestion
- Order JS Tabs, NativeTabs, headless tabs and Drawer by `routeNames` order ([#48374](https://github.com/expo/expo/pull/48374) by [@Ubax](https://github.com/Ubax))
```

packages/expo-router/src/react-navigation/routers/TabRouter.tsx
Ubax: Fix the test

Use the previous implementation

packages/expo-router/src/react-navigation/routers/TabRouter.tsx
```diff
@@ -308,24 +311,55 @@ export function TabRouter({
...
+        const orderedIndex = orderedRoutes.findIndex((route) => route.key === routes[index]!.key);
+        history = [
+          ...getRouteHistory(orderedRoutes, orderedIndex, backBehavior, initialRouteName),
+          ...history.filter((item) => item.type !== 'route'),
+        ];
+      } else if (!history.some((item) => item.type === 'route')) {
+        history = [
+          ...getRouteHistory(routes, index, backBehavior, initialRouteName),
+          ...history.filter((item) => item.type !== 'route'),
+        ];
```
Ubax, line 364: What is this logic for?

Add comment

packages/expo-router/src/ui/Tabs.tsx
```diff
@@ -184,6 +185,16 @@ export function useTabsWithTriggers(options: UseTabsWithTriggersOptions): TabsCo
...
+  useVisibleTabsWithRedirect({
+    routes: state.routes,
+    routeNames: state.routeNames,
+    focusedRouteKey: state.routes[state.index]!.key,
+    descriptors,
+    redirectToRouteName: initialRouteName,
+  });
+
+  // TODO(@ubax): Show a formsheet for focused routes without a trigger (Tabs + Stack in one).
+
```
Ubax:
```suggestion
```

packages/expo-router/src/ui/TabTrigger.tsx
jakub-agent, line 128: [🏗️ **design**] Carried over from round 2, still the one item I'd want resolved before merge.

`Trigger` is public API — `expo-router/ui` re-exports `./Tabs`, which re-exports `./TabTrigger` — and `useTabTrigger` is documented as the hook for building a custom `TabTrigger`. Widening `route` to optional is a breaking type change: a consumer reading `trigger.route.key` (the exact shape this file used until this commit) stops compiling, with no runtime error to point at the cause.

The runtime change is right — an external trigger has no route, and `trigger.route?.key ?? trigger.href` reflects that honestly. The gap is only that it ships silently. A `🛠 Breaking changes` entry naming the type and the `?.`/`??` migration would close it.

Still flagging the caveat from round 2: this looks like it belongs to the headless-tabs refactor rather than to tab ordering. If that work moves to its own PR, the changelog entry should travel with it.

packages/expo-router/src/native-tabs/NativeBottomTabsNavigator.tsx
```diff
@@ -53,7 +54,7 @@ function NativeTabsContent({
...
```
jakub-agent, line 57: [🐧 nit] Now that bottom tabs, top tabs, and drawer all take `routeNames` through a declared `...NavigatorCreateProps` interface, this is the only navigator left hand-intersecting the content props:

```ts
// createBottomTabNavigator.tsx, createMaterialTopTabNavigator.tsx, createDrawerNavigator.tsx
export interface BottomTabNavigatorCreateProps {
  routeNames: string[];
  preloadedRouteKeys: string[];
  ...
}
```

So what I raised as a style nit in round 1 is now a four-way inconsistency in code this PR touched. Declaring `NativeTabsNavigatorCreateProps { routeNames: string[] }` and annotating with `NavigatorContentProps<..., NativeTabsNavigatorCreateProps>` would match the siblings and let `CreateProps` be inferred instead of spelling out all six generics at the call site below.

packages/expo-router/CHANGELOG.md
```diff
@@ -38,6 +38,7 @@
 
 ### 💡 Others
 
+- Order JS Tabs, NativeTabs, headless tabs and Drawer by `routeNames` order ([#48374](https://github.com/expo/expo/pull/48374) by [@Ubax](https://github.com/Ubax))
```
jakub-agent, line 41: [🐧 nit] The list omits TopTabs. `TopTabsClient.tsx` and `createMaterialTopTabNavigator.tsx` both gained `routeNames`, and material top tabs order through `useVisibleTabsWithRedirect` like the rest — so the strip reorders too, and someone scanning the changelog for why their top tabs moved won't find it here.

Separately, `💡 Others` undersells this. Tab, drawer, and top-tab items now follow declaration order rather than mount order, which is observable in any app whose tab set changes at runtime. `🐛 Bug fixes` (or `🎉 New features`) would put it where an upgrading user actually looks.

jakub-agent (commented):

jakub-agent (approved):

PR #48382, opened 2026-07-31
chore(cli): Re-enable sextant QR code for Zed

mchisolm0:
# Why

Zed's integrated terminal had malformed QR codes with some fonts that became more obvious with Expo's new sextant based QR code rendering. The sextant QR code rendering was disabled in https://github.com/expo/expo/pull/46148 but now Zed handles the sextant range needed for the new QR code rendering with the changes from https://github.com/zed-industries/zed/pull/51072.

This PR removes the check that gated against Zed rendering the sextant QR code.

# How

I removed the changes made by @/kitten in https://github.com/expo/expo/pull/46148/changes.

# Test Plan

Added the following to my `.zshrc` as the docs say
```
# expo custom build
alias nexpo="/path/to/project/expo/packages/@expo/cli/bin/cli.js"
alias expo-inspect="node --inspect /path/to/project/expo/packages/@expo/cli/bin/cli.js"
```

1. Build the custom cli based on the instructions in the README.md
2. Open an Expo SDK 55+ app in Zed with Monaco font for the terminal (a font that use to show the regression)
3. Open the Zed integrated terminal
4. Run `npx expo start`
5. Observe the current release falls back to the non-sextant QR code rendering
6. Run `nexpo start`
7. Observe the QR code is now using sextant rendering

| Zed Integrated Terminal w/ fallback QR code | Zed Integrated Terminal w/ sextant QR code |
| --- | --- |
| SCR-20260731-kiyk-2 | SCR-20260731-kipj-2 |

# Checklist


mchisolm0:
Also, I noticed that the README actually says to add the following to your `.zshrc` *without* the `.js`.

```
alias nexpo="/path/to/expo/packages/@expo/cli/build/bin/cli"
...
alias expo-inspect="node --inspect /path/to/expo/packages/@expo/cli/build/bin/cli"
```

This gave the error `zsh: no such file or directory...`, so I added the `.js` and removed the `build` which matched what was actually in the `cli` directory for me. Should the README.md get updated to include the extension?

kitten:
@mchisolm0: Thanks for pointing that out. Some of the contributing guidelines have changed and we'll have to edit that. I don't think that recommendation was ever that useful and the `bin` file has changed in the meantime.

Re. the change itself, I guess we'll need to get an idea of how quickly users of Zed usually update versions and try to time this patch to make sure that as few as possible run into this again.

For context, I initially disabled this output when I became aware (both via various socials and GitHub) of Zed not rendering this output properly. I initially tested this but forgot to change the font while testing (which is usually how I tested against other terminals, typically at least on Windows, macOS, and Linux). Disabling it was the quickest way to make sure that Zed's issue didn't get overloaded. I'm not sure if this was a regression or the font issue though.

Either way, not the point here; main concern is for us to try not to ship this patch too early, so I guess the two options are:
- retest and check manually what the tail of user updates usually is (not sure if there's public info on this)
- wait a few weeks and ship this when we're confident most users have updated

mchisolm0:
All very valid points. I think the decision to disable it was the right call to avoid the Zed team getting a bunch of these issues. 

I’ll check if I can find the data on user updates. Good shout.

kitten:
I just had another look. It looks like [the PR is in a pre-release right now](https://github.com/zed-industries/zed/commit/c2a610f7eb0aae50e2a0dd88972e7b6c9824bd7d). Usual releases are subject to auto-updates, which (I assume) are auto-enabled, so we'd only need a waiting period of maybe one week here, and can maybe just skip backporting this change to all SDK releases and instead just pick it to SDK 56 and 57

mchisolm0:
It was in pre-release, and I just confirmed it landed in stable yesterday with 1.14.2 (https://zed.dev/releases/stable/1.14.2). I also manually checked that the sextant QR codes work with SDK 55 in a Zed stable's integrated terminal, now. Yes, I think you are right that auto-update is on by default in Zed, so I agree a week sounds reasonable.

Yes, SDK 56 and 57 are the only ones that need it because SDK 55 and before continue to render fine. Let me know anything else you need from me.

kitten (approved):

PR #48385, opened 2026-07-31
[expo-contacts][iOS] Return image and thumbnail URIs instead of file paths

martintreurnicht:
# Why

`ImageService.url(from:filename:)` (class API) and `writeDataToUri` (legacy) returned `URL.path`, so iOS handed back **bare filesystem paths** while the documented contract — and Android — use URIs:

- `Contact.types.ts` documents `image`/`thumbnail` as `@example "file:///path/to/image.jpg"` (4 places), and `setImage` is documented as `setImage('file:///path/to/new/image.jpg')`
- Android already returns `content://…` URIs (`PhotoUri` / `PhotoThumbnailUri`)
- the legacy response field is literally named `uri`

Two user-visible consequences follow from the mismatch, which returning a URI fixes:

**1. Photos with a `:` in the identifier never render.** AddressBook-backed contacts have identifiers like `<uuid>:ABPerson`, so the file is `<uuid>:ABPerson-image.png`. A *path* containing `:` cannot be reopened by consumers that convert the string to a URL — React Native's image loader fails with:

```
The file "DBFB258B-6102-44A9-8B95-B856479686A8:ABPerson-image.png" couldn't be opened.
```

On a real 715-contact address book that was **146 of the 256 contacts with photos** (57%): 99 distinct failing files, all 99 containing `:`, zero colon-free failures. Returning `file://…` fixes it without special-casing the character.

**2. The value can't be round-tripped.** `ImageService.imageData(from:)` and the legacy image setter both require `URL(string: value)?.isFileURL`, which a bare path fails — so an image path the module returned was rejected when passed back into `setImage`.

# How

Return `URL.absoluteString` from both write paths. The `/` sanitization from #48201 stays, since `appendingPathComponent` would still treat it as a nested directory.

Note this changes what iOS returns for these fields (path → URI). That's the documented behaviour and matches Android, so I've filed it under bug fixes — happy to move the CHANGELOG entry if you'd rather flag it as breaking.

# Test Plan

Built an app against these sources on device (iOS 26.5) with an address book mixing AddressBook-backed (`<uuid>:ABPerson`) and plain-UUID identifiers:

- before: 106 `Image` load failures, all colon-containing; 57% of contacts with photos showed no avatar
- after: every contact photo renders, 0 load failures

Applied to both the class API and the legacy module, as requested.

# Checklist

- [x] Documentation is up to date to reflect any changes (n/a — this makes the code match the existing documented contract).

### DIFF
--- a/packages/expo-contacts/ios/next/objects/contact/services/ImageService.swift
+++ b/packages/expo-contacts/ios/next/objects/contact/services/ImageService.swift
@@ -14,8 +14,6 @@ class ImageService {
   func url(from imageData: Data, filename: String) throws -> String {
     let imageFileURL = try prepareCacheFileURL(filename: filename)
     try imageData.write(to: imageFileURL, options: .atomic)
-    // The API contract is a URI (see the `file:///path/to/image.jpg` examples in the types, and
-    // Android returning `content://` URIs), so return one instead of a bare filesystem path.
     return imageFileURL.absoluteString
   }
 
--- a/packages/expo-contacts/ios/next/objects/contact/services/ImageService.swift
+++ b/packages/expo-contacts/ios/next/objects/contact/services/ImageService.swift
@@ -14,7 +14,9 @@ class ImageService {
   func url(from imageData: Data, filename: String) throws -> String {
     let imageFileURL = try prepareCacheFileURL(filename: filename)
     try imageData.write(to: imageFileURL, options: .atomic)
-    return imageFileURL.path
+    // The API contract is a URI (see the `file:///path/to/image.jpg` examples in the types, and
+    // Android returning `content://` URIs), so return one instead of a bare filesystem path.
+    return imageFileURL.absoluteString
   }
 
   func imageData(from url: String) throws -> Data? {
--- a/packages/expo-contacts/CHANGELOG.md
+++ b/packages/expo-contacts/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 - [iOS] Fix `Contact.getAllDetails` always returning `null` for `thumbnail`, `birthday` and `nonGregorianBirthday`. ([#48384](https://github.com/expo/expo/pull/48384) by [@martintreurnicht](https://github.com/martintreurnicht))
 - [iOS] Fix `Contact.getAll` and `Contact.getAllDetails` returning contacts linked across accounts once per account record, and honour the `rawContacts` query option. ([#48387](https://github.com/expo/expo/pull/48387) by [@martintreurnicht](https://github.com/martintreurnicht))
+- [iOS] Return contact image and thumbnail URIs instead of bare file paths, matching the documented contract and Android. ([#48385](https://github.com/expo/expo/pull/48385) by [@martintreurnicht](https://github.com/martintreurnicht))
 
 ### 💡 Others
 
--- a/packages/expo-contacts/ios/Serialization.swift
+++ b/packages/expo-contacts/ios/Serialization.swift
@@ -353,7 +353,6 @@ private func writeDataToUri(userId: String, data: Data?, imageKey: String, inclu
 
   try data.write(to: newPath, options: .atomic)
   var response: [String: Any?] = [
-    // Matches the class API and Android: the contract for this field is a URI, not a path.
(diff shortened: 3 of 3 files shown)


packages/expo-contacts/ios/next/objects/contact/services/ImageService.swift
```diff
@@ -38,8 +38,14 @@ class ImageService {
...
       throw FailedToCacheContactImage("The contacts cache directory does not exist")
     }
-    // Contact identifiers can contain "/" (e.g. "com.apple.introductions.accepted/494"), which
-    // `appendingPathComponent` would treat as a nested directory that doesn't exist.
-    return baseDirectoryURL.appendingPathComponent(filename.replacingOccurrences(of: "/", with: "_"))
+    // Contact identifiers are not safe to interpolate into a file name. "/" (e.g.
+    // "com.apple.introductions.accepted/494") makes `appendingPathComponent` build a nested path
+    // that doesn't exist, and ":" (e.g. "<uuid>:ABPerson" for AddressBook-backed contacts)
+    // produces a file that consumers round-tripping the path through a URL cannot open —
+    // React Native's image loader fails with "couldn't be opened", so the photo never renders.
```
Wenszel: Let's simplify the comment so that it only lists examples of potentially unsafe characters found in contact identifiers.
```suggestion
    // Contact identifiers can contain characters that are not safe in a file name:
    // - "/" in "com.apple.introductions.accepted/494"
    // - ":" in "6C4E1E5D-...:ABPerson"
```
Worth noting that paths containing `:` load correctly in `expo-image`. However, `:` can still cause problems in other string -> URL conversions, with React Native's Image being one example.

Wenszel (changes requested):
Another great catch, thank you! Could you please apply the same change to the legacy module as well?

packages/expo-contacts/ios/next/objects/contact/services/ImageService.swift
Wenszel: Hey, actually, I realized there may be another problem here besides sanitizing the `:` character. The API contract says that images/thumbnails are URIs, and they already are on Android. However, on iOS, we return file paths instead (without the `file://` scheme prefix).

Can you change this behavior so that we return a URI? That would also make replacing `:` unnecessary.

martintreurnicht: Done. Returning a URI  in both the class API (`ImageService.url`) and the legacy module (`writeDataToUri`), and the `:` replacement is gone as you predicted.

You were right that this is the better fix, and digging into it turned up a second symptom of the same mismatch: the returned value couldn't be round-tripped. `ImageService.imageData(from:)` and the legacy image setter both require `URL(string:)?.isFileURL`, which a bare path fails — so a path the module handed out was rejected when passed back into `setImage`. Returning `absoluteString` fixes that too.

Kept the `/` sanitization from #48201, since `appendingPathComponent` would still build a nested directory for `com.apple.introductions.accepted/494`. That also let the comment shrink to just the `/` case, close to your suggestion.

Thanks for the `expo-image` data point, that matches what I saw: the files were on disk and perfectly readable, it was only the string→URL conversion in RN's loader that broke. Worth flagging that this changes what iOS returns for `image`/`thumbnail` (path → URI); I put the CHANGELOG entry under bug fixes since it makes the code match the documented contract, but say the word if you'd rather mark it breaking.

martintreurnicht: Confirmed on device, and your call was right — the URI is sufficient on its own, no character replacement needed.

Ran the URI version on iOS 26.5 against a 715-contact address book. 146 cache files are now written with the colon intact (e.g. `0222168E-99D8-4CE2-9766-822C56F73168:ABPerson-thumb.png`) and every one loads: `Image`'s `onError` fires **0** times, where the bare-path build produced 106 failures across 99 distinct colon-bearing files. So RN's loader is fine with `:` inside a well-formed `file://` URI — it was only the scheme-less path that broke its string→URL conversion, exactly as you said.

Wenszel (approved):

PR #48387, opened 2026-07-31
[expo-contacts][iOS] Honour `rawContacts` in `getAll` and `getAllDetails` (fixes duplicated contacts)

martintreurnicht:
# Why

`Contact.getAll` and `Contact.getAllDetails` never forward `queryOptions.unifyContacts` to `ContactRepository.getPaginated`, so both always run with `unifyResults = false`. A contact linked across accounts — iCloud + Google + Exchange, which is the common case — is then returned **once per account record**, i.e. duplicated.

Two things make this worth treating as a bug rather than a default:

- `CNContactFetchRequest.unifyResults` [defaults to `true`](https://developer.apple.com/documentation/contacts/cncontactfetchrequest/unifyresults) in the Contacts framework, so the class API is actively opting out.
- The legacy API sets it explicitly (`ios/ContactsModule.swift`: `fetchRequest.unifyResults = true`, and `!rawContacts` when requested), so apps migrating from `expo-contacts/legacy` to the class API silently gain duplicates.

`Container` and `Group` queries already forward the option (`unifyResults: queryOptions?.unifyContacts ?? false`) — only these two paths drop it.

Measured on a real 715-record address book, before → after:

```
before: contacts=715  distinctIds=715  distinctNames=375  duplicatedNames=335
after:  contacts=381  distinctIds=381  distinctNames=374  duplicatedNames=4
```

(the remaining 4 are genuinely distinct contacts that happen to share a name)

# How

Forward `queryOptions?.unifyContacts` in both functions, defaulting to `true` to match `CNContactFetchRequest` and the legacy API.

Note for reviewers: there is also a naming mismatch that makes the option unreachable from JS today — `ContactQueryOptions` in TS exposes `rawContacts?: boolean` while the native record declares `unifyContacts`. I left that alone here to keep this diff focused; happy to fix it in this PR or a follow-up, whichever you prefer.

# Test Plan

Built an app against these sources on device (iOS 26.5) with an address book containing contacts linked across iCloud, Google and Exchange:

- before: every linked contact appeared 2–3 times in the list (numbers above)
- after: one row per person, 715 → 381, and contact photos still resolve

# Checklist

- [x] Documentation is up to date to reflect any changes (n/a — bug fix, no API change).

### DIFF
--- a/packages/expo-contacts/CHANGELOG.md
+++ b/packages/expo-contacts/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### 🐛 Bug fixes
 
 - [iOS] Fix `Contact.getAllDetails` always returning `null` for `thumbnail`, `birthday` and `nonGregorianBirthday`. ([#48384](https://github.com/expo/expo/pull/48384) by [@martintreurnicht](https://github.com/martintreurnicht))
+- [iOS] Fix `Contact.getAll` and `Contact.getAllDetails` returning contacts linked across accounts once per account record, and honour the `rawContacts` query option. ([#48387](https://github.com/expo/expo/pull/48387) by [@martintreurnicht](https://github.com/martintreurnicht))
 
 ### 💡 Others
 
--- a/packages/expo-contacts/ios/next/Contact.swift
+++ b/packages/expo-contacts/ios/next/Contact.swift
@@ -115,7 +115,8 @@ class ContactNext: SharedObject {
       offset: queryOptions?.offset,
       sortOrder: queryOptions?.sortOrder.map {
         CNContactSortOrderMapper.map($0)
-      }
+      },
+      unifyResults: !(queryOptions?.rawContacts ?? false)
     )
     .map { contactFactory.create(id: $0.identifier) }
   }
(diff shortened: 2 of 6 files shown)


Wenszel (changes requested):
Hey thanks for fixing the issue! Could you please rename the `unifyContacts` field in the `ContactQueryOptions` record to `rawContacts`, because now value passed from TS is lost during conversion? Also, don't forget the negation.

martintreurnicht:
Done. I've renamed the record field to `rawContacts` and forwarded it as `!rawContacts`, so the TS value now survives the conversion instead of being dropped. Good catch that the field name was the reason it never arrived; I'd only fixed the missing forwarding and left the mismatch in place.

Two notes on the scope:

- `Container` and `Group` read the same field, so they had to change too or the rename wouldn't compile. I gave them the same polarity (`!(rawContacts ?? false)`), which does shift their default from raw to unified — consistent with Apple's `unifyResults = true` default and with the legacy module's `!rawContacts`. If you'd rather keep those two on their previous default while still honouring the option, `queryOptions?.rawContacts.map { !$0 } ?? false` does that and I'm happy to switch.
- Default is `false` per the TS docs ("Whether to include raw contact data in the results. Default is false."), hence unified by default.

Verified on device (iOS 26.5) against a 715-record address book: 715 rows / 375 distinct people → 381 rows / 374 distinct, with 335 duplicated names dropping to 4 (those four are genuinely distinct contacts that share a name).

Wenszel (approved):

PR #48083, opened 2026-07-23
[create-expo] Fix npm pack --json parsing for npm >= 12

lllincoln:
# Why

`create-expo-app` fails for any SDK version with `Error: Invalid response from npm: ...` when the user has npm >= 12 installed. This is because npm 12 made a breaking change to `npm pack --json` (https://github.com/npm/cli/pull/9247 and https://docs.npmjs.com/cli/v12/using-npm/changelog/).

# How

`parseNpmPackOutput` now works with both shapes by normalizing objects to arrays with `Object.values()` before running the same existing validation, so both npm <12 and npm >=12 output work without version detection.

# Test Plan

Added unit tests in `packages/create-expo/src/utils/__tests__/npm.test.ts` covering `parseNpmPackOutput` for npm < 12, npm >= 12, and an invalid response. All tests pass.

# Checklist

- [x] I added a `changelog.md` entry and rebuilt the package sources according to [this short guide](https://github.com/expo/expo/blob/main/CONTRIBUTING.md#-before-submitting)
- [ ] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: updated a module plugin). **(not applicable?)**
- [x] Conforms with the [Documentation Writing Style Guide](https://github.com/expo/expo/blob/main/guides/Expo%20Documentation%20Writing%20Style%20Guide.md)

PR #48392, opened 2026-08-01
[create-expo] Fix npm pack JSON parsing on npm 12

00vortexlabsoneuzb00-wq:
# Why

`create-expo` fails to download any template on `npm@>=12`.

`npmPackAsync` in `packages/create-expo/src/utils/npm.ts` runs `npm pack --json` and expects an array of package infos back. Starting with `npm@12`, `npm pack --json` returns an **object keyed by package name** instead:

```console
$ npm --version
12.0.2

$ npm pack expo-template-blank --dry-run --json
{
  "expo-template-blank": {
    "id": "expo-template-blank@57.0.11",
    "name": "expo-template-blank",
    "version": "57.0.11",
    "filename": "expo-template-blank-57.0.11.tgz",
    ...
  }
}
```

The `Array.isArray(json)` check therefore fails and the command aborts with `Invalid response from npm: ...`, so no project can be created. Because that error is thrown inside the `try` block that wraps `JSON.parse`, it is then caught and re-thrown as `Could not parse JSON returned from "npm pack ..."`, which points the user at JSON parsing rather than at the real cause.

A second, smaller problem: the raw stdout is passed straight to `JSON.parse`, so any log line npm prints next to the payload breaks parsing.

# How

All changes are in `packages/create-expo/src/utils/npm.ts`.

- **Normalize the response shape.** A new `normalizeNpmPackInfos` helper accepts all three shapes `npm pack --json` can produce and returns `NpmPackageInfo[]`:
  - an array of package infos (`npm@<12`),
  - an object keyed by package name, unwrapped with `Object.values` (`npm@>=12`),
  - a single bare package info, wrapped into a list.

  The bare-object case is checked **first**, because a package info is an object too. Running `Object.values` on it would return its field values (`"expo-template-blank@57.0.11"`, `509443`, …) and fail validation with the same `Invalid response` error the fix is meant to remove.

- **Tolerate non-JSON output.** A new `extractJson` helper slices from the first `{` or `[` to the last `}` or `]` before parsing, so warnings or notices printed next to the payload no longer break it. If nothing parseable is found it returns `null` and the caller throws the existing "Could not parse JSON" error with the full npm output attached, so the diagnostic detail is preserved.

- **Stop masking the real error.** `Invalid response from npm` is now thrown outside the JSON parse `try`, so an unexpected response is no longer re-labelled as a JSON parse failure. The two failure modes now report distinctly.

- **Exported `npmPackAsync`** so the behaviour above can be unit tested with a mocked `@expo/spawn-async`. It is not exported from the package entry point and remains internal to `create-expo`.

Existing behaviour is unchanged for `npm@<12`: the array path is still the first branch, and the `E404` handling and `sanitizeNpmPackageFilename` call are untouched.

# Test Plan

**New unit tests** — `packages/create-expo/src/utils/__tests__/npmPack.test.ts` mocks `@expo/spawn-async` and covers the array response, the keyed object response, a single package info, output with surrounding noise, empty output, and an unexpected response. Three of them — the keyed object, the single package info, and the noisy output — fail against `main` and pass with this change.

```console
$ pnpm test
Test Suites: 17 passed, 17 total
Tests:       136 passed, 136 total
Snapshots:   1 passed, 1 total
```

**Full package check:**

```console
$ et check-packages create-expo
create-expo:lint: Found 0 warnings and 0 errors.
create-expo:format: All matched files use the correct format.
🏁 All checks passed.
```

**End-to-end against a real registry with `npm@12.0.2`** — built the bundle and generated a project:

```console
$ npm --version
12.0.2

$ pnpm build
$ EXPO_NO_CACHE=1 node build/index.js my-app --template blank --no-install
Creating my-app using the blank template.
✔ Downloaded and extracted project files.
✅ Your project is ready!

$ ls my-app
AGENTS.md  App.js  app.json  assets  CLAUDE.md  index.js  LICENSE  package.json
```

**To reproduce the bug**, run the same command on `main` with `npm@>=12` installed, or run the new test file against `main`, where those three cases fail.

# Checklist

- [ ] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: updated a module plugin). — not applicable, this does not touch a config plugin or native code

### DIFF
--- a/packages/create-expo/CHANGELOG.md
+++ b/packages/create-expo/CHANGELOG.md
@@ -13,6 +13,8 @@
 
 ### 🐛 Bug fixes
 
+- Fix template downloads failing on `npm@>=12`, which returns `npm pack --json` results as an object keyed by package name instead of an array. ([#48392](https://github.com/expo/expo/pull/48392) by [@00vortexlabsoneuzb00-wq](https://github.com/00vortexlabsoneuzb00-wq))
+
 ### 💡 Others
 
 ## 5.0.0 - 2026-06-25
--- a/packages/create-expo/src/utils/__tests__/npmPack.test.ts
+++ b/packages/create-expo/src/utils/__tests__/npmPack.test.ts
@@ -1,0 +1,60 @@
+import spawnAsync from '@expo/spawn-async';
+
+import { npmPackAsync } from '../npm';
+
+jest.mock('@expo/spawn-async');
+
+const mockSpawnAsync = spawnAsync as jest.MockedFunction<typeof spawnAsync>;
+
+function mockNpmStdout(stdout: string) {
+  mockSpawnAsync.mockResolvedValue({ stdout } as any);
+}
+
+const packageInfo = {
+  id: 'expo-template-blank@57.0.11',
+  name: 'expo-template-blank',
+  version: '57.0.11',
+  size: 509443,
+  unpackedSize: 516067,
+  shasum: '5b4d5304d629b42cbbc35aeb59a8ac223412c1a1',
+  integrity:
+    'sha512-A9SuNKCRIkjKYaj39KG0h898wLXosWL2s/bF1AQLHhcJjDk+S7wTEpCXTjB59ocYRxhd4DRXEZ09Qq7VK99IQg==',
+  filename: 'expo-template-blank-57.0.11.tgz',
+  files: [],
+  entryCount: 0,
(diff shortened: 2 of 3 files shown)


brentvatne:
Closing — same npm 12 `pack --json` fix landed in https://github.com/expo/expo/pull/48761. Duplicate of #48083. Thanks!

PR #48407, opened 2026-08-02
[android][core] Restore hosted Compose layout after lifecycle changes

### LINKED ISSUE #48370: [expo-modules-core][Android] ExpoComposeView misses layout after reattachment or configuration changes

https://github.com/lujjjh/expo-ui-menu-android-popup-repro

### Steps to reproduce

After a retained `ExpoComposeView` is reattached or receives an in-place Android configuration change, newly inserted Compose content is recomposed, but it does not receive the layout pass required to determine its coordinates. A Compose popup makes this especially visible because the popup window is created but remains transparent until its anchor coordinates are available.

The reproduction uses the public `MenuView` API from `@expo/ui/community/menu`. It does not patch any Expo package.

### Reproduction after screen reattachment

1. Clone the reproduction repository.
2. Run `npm ci`.
3. Run `npm run android`.
4. Tap the menu trigger in the top-right corner and confirm that the menu is visible.
5. Tap **Push details**.
6. Return to the home screen using either Back control.
7. Tap the menu trigger again.

### Reproduction after an Android configuration change

1. Cold-launch the app and confirm that the menu initially works.
2. While keeping the Activity alive, change the system appearance from light to dark or dark to light. This can also be triggered with:

  ```sh
  adb shell cmd uimode night yes
  adb shell cmd uimode night no
  ```

3. Tap the menu trigger again.

### Expected behavior

The newly inserted Compose popup content should be measured, placed, and displayed at the menu trigger after either lifecycle event.

### Actual behavior

`onOpenMenu` fires and the popup window is created, but the popup content is invisible.
The invisible window still consumes input: the first tap outside the menu dismisses it without reaching the underlying React Native UI, while the second tap reaches the underlying UI.

https://github.com/user-attachments/assets/aae39814-5b45-477d-ae93-f7c8a3e10466

### Investigation

The observable symptom is an invisible `@expo/ui` menu, but the affected layer appears to be the Android Compose hosting/layout integration in `expo-modules-core`.

The runtime path is:

`MenuView` → `@expo/ui Host` → `HostView` → `expo-modules-core ExpoComposeView`

The two reproduction paths converge on a missing Android layout pass, but reach that state through different lifecycle sequences.

After screen reattachment:

- The retained host can be detached while a manual layout pass posted by `ExpoView.requestLayout()` is still pending.
- That pass can execute after detachment and leave the outer hosting view measured at `0x0`, with its layout-request state cleared.
- When the same host is reattached, the internal `ComposeView` still requires layout, but React Native retains the same Yoga bounds and does not schedule another Android traversal.

After an in-place Android configuration change:

- The Compose subtree receives the new configuration and invalidates layout on a following frame.
- The host and its Compose child can become layout-requested without scheduling `ExpoView`'s manual `measureAndLayout()` bridge.
- Because the Yoga bounds remain unchanged, no subsequent Android layout traversal reaches the newly inserted Compose content.

Opening the menu still triggers recomposition and creates the popup window. AndroidX Popup intentionally keeps its content transparent until it can calculate a position from valid attached-parent coordinates, which explains why the invisible popup consumes input.

A narrowed candidate fix requests the existing asynchronous Android layout path only for `ExpoComposeView` instances that own a hosting `ComposeView`:

- An existing `OnAttachAfterDetachmentListener` requests layout after a real detach and reattach, while continuing to ignore the initial attach and same-loop React Native reparenting.
- `dispatchConfigurationChanged()` first dispatches the new configuration to the child tree, then requests layout when the hosting view is already attached and laid out.

This avoids remounting the composition, changing the menu implementation, or modifying the generic `ExpoView` layout bridge.

Candidate patch:

https://github.com/lujjjh/expo-ui-menu-android-popup-repro/blob/6e07b1dc710ef545abdc6dbeb37de236eb0231f9/patches/expo-modules-core-host-layout.patch

In the reproduction, this candidate kept the popup visible and correctly anchored across the initial mount, 10 navigation reattachments, 6 system `uiMode` changes, and 2 combined navigation/configuration cycles without recreating the process.

The patch is included only as investigation evidence and is not applied automatically by the reproduction project. It is not intended to prescribe the final upstream implementation.

### Environment

```text

  System:
    OS: macOS 26.5.2
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 26.5.0 - /opt/homebrew/bin/node
  Managers:
    CocoaPods: 1.16.2 - /Users/jiahao/.rbenv/shims/pod
  SDKs:
    iOS SDK:
    Android SDK:
      API Levels: 34, 35, 36, 36
      Build Tools: 33.0.1, 35.0.0, 36.0.0, 36.1.0
      System Images: android-36.1 | Google Play ARM 64 v8a
  IDEs:
    Android Studio: 2025.3 AI-253.30387.90.2532.14935130
  npmPackages:
    expo: ~57.0.9 => 57.0.9
    react: 19.2.3 => 19.2.3
  Expo Workflow: bare
```

### Expo Doctor Diagnostics

```text
```

lujjjh:
# Why

Fixes #48370.

After a retained hosted `ExpoComposeView` is reattached or receives an in-place Android configuration change, newly inserted Compose content may not receive the Android layout pass required to determine its coordinates.

A `MenuView` popup makes this visible because the popup window is created and consumes input, but its content remains invisible while valid anchor coordinates are unavailable. The affected layer is the Android Compose hosting lifecycle in `expo-modules-core`, rather than the menu implementation itself.

# How

- Request the existing asynchronous Android layout path after a real detach and reattach of a hosted `ExpoComposeView`.
- Reuse `OnAttachAfterDetachmentListener`, preserving its behavior of ignoring the initial attach and same-loop React Native reparenting.
- After dispatching an Android configuration change, request layout only when the hosted view is attached and already laid out.
- Keep the change scoped to views that own a hosting `ComposeView`, without remounting the composition or modifying the generic `ExpoView` layout bridge.

# Test Plan

- Ran `et check-packages expo-modules-core`.
- Ran `et native-unit-tests -p android --packages expo-modules-core`.
- Ran `cd apps/bare-expo/android && ./gradlew :expo-modules-core:spotlessCheck --build-cache`.
- Built the minimal reproduction against matching Expo sources from the latest `main`.
- Verified on a Pixel 9 Pro Android emulator:
  - the menu opens after the initial mount;
  - the menu remains visible and correctly anchored after navigating to another screen and returning;
  - the menu remains visible after an in-place system `uiMode` change;
  - the menu remains visible when the configuration changes while its screen is detached, followed by navigation back.
- Confirmed the process was not recreated, popup bounds remained stable, and logcat contained no fatal or Compose layout exceptions.

Minimal reproduction: https://github.com/lujjjh/expo-ui-menu-android-popup-repro

# Checklist

### DIFF
--- a/packages/expo-modules-core/CHANGELOG.md
+++ b/packages/expo-modules-core/CHANGELOG.md
@@ -20,6 +20,7 @@
 
 ### 🐛 Bug fixes
 
+- [Android] Fixed hosted Compose views missing layout after reattachment or in-place configuration changes. ([#48370](https://github.com/expo/expo/issues/48370) by [@lujjjh](https://github.com/lujjjh))
 - [iOS] Fixed infinite recursion and `EXC_BAD_ACCESS` when encoding native `Date` values to JavaScript. ([#48239](https://github.com/expo/expo/issues/48239), [#48240](https://github.com/expo/expo/pull/48240) by [@samuelcorsan](https://github.com/samuelcorsan))
 - [iOS] Concurrent functions build on the new two-phase `createAsyncFunction` from `expo-modules-jsi` again, instead of wiring the promise themselves. This also restores rejecting the returned promise (rather than throwing synchronously) when the app context is lost or argument decoding fails. ([#47755](https://github.com/expo/expo/pull/47755) by [@tsapeta](https://github.com/tsapeta))
 - [Android] Support `android.graphics.Color` arguments and props on Android below API 26 by adding a `ColorCompat` wrapper, since the class-based `Color` API (`Color.valueOf` and the float accessors) only exists on API 26 and above. `PlatformColor` values now resolve below API 26 too, via React Native's ARGB-int color path. ([#47546](https://github.com/expo/expo/issues/47546) by [@anasvemmully](https://github.com/anasvemmully)) ([#47575](https://github.com/expo/expo/pull/47575) by [@intergalacticspacehighway](https://github.com/intergalacticspacehighway))
--- a/packages/expo-modules-core/android/src/compose/expo/modules/kotlin/views/ExpoComposeView.kt
+++ b/packages/expo-modules-core/android/src/compose/expo/modules/kotlin/views/ExpoComposeView.kt
@@ -2,6 +2,7 @@ package expo.modules.kotlin.views
 
 import android.annotation.SuppressLint
 import android.content.Context
+import android.content.res.Configuration
 import android.util.Log
 import android.view.View
 import android.view.ViewGroup
@@ -113,6 +114,15 @@ abstract class ExpoComposeView<T : ComposeProps>(
     }
   }
 
+  override fun dispatchConfigurationChanged(newConfig: Configuration) {
+    super.dispatchConfigurationChanged(newConfig)
+    // React Native owns this view's bounds and may not schedule another Android layout pass
+    // when a configuration change leaves its Yoga layout unchanged.
+    if (withHostingView && isAttachedToWindow && isLaidOut) {
+      requestLayout()
+    }
+  }
+
   /**
    * Validates that this non-hosting Compose view has a valid Compose parent.
    *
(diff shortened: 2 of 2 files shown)


intergalacticspacehighway (approved):
Thanks!

PR #48415, opened 2026-08-02
[create-expo] Add tests for the GitHub template resolver

00vortexlabsoneuzb00-wq:
# Why

`packages/create-expo/src/utils/github.ts` resolves and downloads templates and examples passed as a GitHub URL, and it had almost no test coverage — **11.9% of statements**, one of the lowest in the package.

Nothing currently guards its branching logic: reading the default branch from the API versus taking it from a `/tree/<branch>` URL, computing how many path segments to strip for a sub directory, and five distinct error paths. A regression in any of them surfaces to the user as a failed `create-expo --template <github url>` run.

# How

Adds `packages/create-expo/src/utils/__tests__/github.test.ts`. **No source files are changed.**

The tests drive the exported `downloadAndExtractGitHubRepositoryAsync`, so the unexported helpers (`getGitHubRepoAsync`, `isValidGitHubRepoAsync`, `extractRemoteGitHubTarballAsync`) are covered without widening the module's public surface. HTTP is intercepted with `nock`, matching the existing approach in `src/__tests__/Examples.test.ts`, and `extractNpmTarballAsync` is mocked so no tarball is written to disk.

Cases covered:

| Case | Asserts |
| - | - |
| URL without a branch | The default branch is read from `GET /repos/:owner/:name` and `strip: 1` is used |
| `/tree/beta` | The branch comes from the URL and the API is never called |
| `/tree/main/with-router` | The sub directory reaches the contents check and `strip: 2` is used |
| Repository missing | `GitHub repository not found for url: ...` |
| Unexpected API status | `[500] Failed to fetch GitHub repository information ...` |
| URL that is not a tree | `Malformed GitHub repository response for URL: ...` |
| No `package.json` in the repository | `Could not locate repository ...`, and no extraction is attempted |
| Tarball response has no body | `Unexpected response: no response body ...` |
| Tarball request fails | `Unexpected response: ...` |

# Test Plan

Coverage for `src/utils/github.ts`:

| | Before | After |
| - | - | - |
| Statements | 11.9% | 100% |
| Branches | 0% | 90% |
| Functions | 0% | 100% |
| Lines | 12.5% | 100% |

```console
$ pnpm test
Test Suites: 17 passed, 17 total
Tests:       139 passed, 139 total

$ et check-packages create-expo
create-expo:lint: Found 0 warnings and 0 errors.
create-expo:format: All matched files use the correct format.
🏁 All checks passed.
```

To confirm the tests fail when the behaviour they describe changes, three mutations were applied to `github.ts` and reverted:

| Mutation | Result |
| - | - |
| `const strip = directory.length` (dropping the `+ 1`) | 2 tests fail |
| `t === 'blob'` instead of `t === 'tree'` | 6 tests fail |
| `response.status === 410` instead of `404` | 1 test fails |

The one uncovered branch is the `owner = ''` / `name = ''` destructuring default on line 20, which is unreachable through a valid `URL`.

# Checklist

- [ ] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: updated a module plugin). — not applicable, tests only

### DIFF
--- a/packages/create-expo/CHANGELOG.md
+++ b/packages/create-expo/CHANGELOG.md
@@ -15,6 +15,8 @@
 
 ### 💡 Others
 
+- Add test coverage for the GitHub template resolver ([#48415](https://github.com/expo/expo/pull/48415) by [@00vortexlabsoneuzb00-wq](https://github.com/00vortexlabsoneuzb00-wq))
+
 ## 5.0.0 - 2026-06-25
 
 _This version does not introduce any user-facing changes._
--- a/packages/create-expo/src/utils/__tests__/github.test.ts
+++ b/packages/create-expo/src/utils/__tests__/github.test.ts
@@ -1,0 +1,180 @@
+import nock from 'nock';
+
+import { downloadAndExtractGitHubRepositoryAsync } from '../github';
+import { extractNpmTarballAsync } from '../npm';
+
+jest.mock('../npm', () => ({
+  ...jest.requireActual('../npm'),
+  extractNpmTarballAsync: jest.fn(async () => '/output'),
+}));
+
+const mockExtractNpmTarballAsync = extractNpmTarballAsync as jest.MockedFunction<
+  typeof extractNpmTarballAsync
+>;
+
+/** Reply to the "does this repository contain a package.json" check. */
+function mockContentsRequest(path: string, status = 200) {
+  return nock('https://api.github.com').get(path).reply(status);
+}
+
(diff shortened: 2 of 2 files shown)


kitten:
Sorry, but I don't think there's much of a point of accepting an contributed PR for tests, especially if it's LLM derived, since this can be trivially added as that code evolves, and as long as it's known to be working, the coverage doesn't really do us any good here. Thanks though!

PR #48426, opened 2026-08-03
[iOS] Fix applyModifier rebuilding a fresh `AnyViewModifier` every body evaluation

wielski:
Fix modifier application rebuilding a fresh `AnyViewModifier` every body evaluation, which defeated AttributeGraph subtree pruning during scroll.

# Why

`AnyViewModifier` stores a closure, so building it here means a fresh heap box every body evaluation. AttributeGraph compares modifier values byte-wise to decide whether a subtree can be skipped, and a fresh box always compares as changed so nothing is ever pruned during scroll. This leads to a heavy load on the main thread and lags in scrolling on older devices.

# How

AttributeGraph decides whether a subtree can be skipped by byte-comparing the modifier value against the previous frame's. StableViewModifier storing representation is three pointers: the params dictionary's heap buffer and two context objects all of which stay identical across body evaluations until JS sends a new modifier array. Equal bytes -> subtree pruned.

# Test Plan

The change was tested on a production tvOS application by sample tool with a large number of complex elements in the scroll (many swift-ui components and modifiers), in both regular stacks and lazy stacks.

```
sample <pid> 10 -file scroll.txt
```

Without this change: main thead busy 50%, AG::Subgraph::update calls > 500
With this change: main thead busy 20% , AG::Subgraph::update calls < 200

# Checklist


packages/expo-ui/ios/Modifiers/ViewModifierRegistry.swift
```diff
@@ -1457,10 +1476,11 @@ public class ViewModifierRegistry {
...
   ) -> AnyView {
-    guard let viewModifier = try? modifierFactories[type]?(params, appContext, globalEventDispatcher) else {
-      return view
-    }
-    return AnyView(view.modifier(AnyViewModifier(viewModifier)))
+    return AnyView(view.modifier(StableViewModifier(
+      params: params,
+      appContext: appContext,
+      dispatcher: globalEventDispatcher
+    )))
```
intergalacticspacehighway, line 1486:
```suggestion
    guard modifierFactories[type] != nil else {
      return view
    }
    return AnyView(view.modifier(StableViewModifier(
      params: params,
      appContext: appContext,
      dispatcher: globalEventDispatcher
    )))
```
nit: let's keep the check here so we return early if no modifier is found.

packages/expo-ui/ios/Modifiers/ViewModifierRegistry.swift
```diff
@@ -866,6 +866,25 @@ internal struct AnyViewModifier: ViewModifier {
...
+  let appContext: AppContext
+  let dispatcher: EventDispatcher
+
+  func body(content: Content) -> some View {
+    if let type = params["$type"] as? String,
+      let factory = ViewModifierRegistry.shared.modifierFactories[type],
+      let modifier = try? factory(params, appContext, dispatcher) {
+      AnyView(content.modifier(AnyViewModifier(modifier)))
+    } else {
+      AnyView(content)
```
intergalacticspacehighway:
```suggestion
      content.modifier(AnyViewModifier(modifier)))
    } else {
      content
```
Lets remove `AnyView` from here

intergalacticspacehighway (changes requested):

intergalacticspacehighway (approved):

intergalacticspacehighway:
@wielski Let's make the appContext weak in `StableViewModifier` since it is weak everywhere (`SwiftUIViewProps.swift`, `ExpoFabricView.swift`) else.

brentvatne:
@expo-bot verify

wielski:
@intergalacticspacehighway good idea, I made an appContext weak.

About [expo-bot](https://github.com/expo-bot) comment - I don't know yet how to cover this with metrics to ensure it works on iOS. Perhaps the performance could be improved further, bringing it closer to SwiftUI - I will work on this later as part of another PR.

The AttributeGraph engine recalculates dependencies on every scroll event (at least during focus change on tvOS), checking whether a node is out of date and whether it needs to be redrawn. Without stable references and modifier recalculation, this work is performed every frame.
Lag can be easily seen if you render a simple list with a huge number of elements on old device, or compare graph update calls using a sampling tool.

brentvatne:
@expo-bot verify

PR #48430, opened 2026-08-03
[ncl] Use Stack.Toolbar for the header search button

Kudo:
# Why

follow up TODO in https://github.com/expo/expo/pull/47510#discussion_r3657582621

# How

- use `Stack.Toolbar` on native.
- since `Stack.Toolbar` has no web implementation, so it keeps the JS `headerRight` on web.

# Test Plan

test NCL

# Checklist

### DIFF
--- a/apps/native-component-list/src/navigation/StackConfig.tsx
+++ b/apps/native-component-list/src/navigation/StackConfig.tsx
@@ -1,5 +1,5 @@
 import Ionicons from '@expo/vector-icons/Ionicons';
-import { ThemeType } from 'ThemeProvider';
+import { ThemeType, useTheme } from 'ThemeProvider';
 import { Stack, router, type NativeStackNavigationOptions, useRouter } from 'expo-router';
 import * as React from 'react';
 import { View, Platform, TouchableOpacity } from 'react-native';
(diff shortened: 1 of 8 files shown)


apps/native-component-list/src/navigation/StackConfig.tsx
```diff
@@ -1,20 +1,39 @@
...
 export function getStackScreenOptions(theme: ThemeType): NativeStackNavigationOptions {
   return {
     contentStyle: { backgroundColor: theme.background.default },
     headerStyle: { backgroundColor: theme.background.default },
     headerTintColor: theme.icon.info,
     headerTitleStyle: { color: theme.text.default },
-    // TODO(kudo,20260802): Convert this into a `Stack.Toolbar.Right` once we adopt the native toolbar API.
-    headerRight: () => <HeaderRightComponent theme={theme} />,
+    // `Stack.Toolbar` renders nothing on web, so web keeps the JS header button.
+    ...(Platform.OS === 'web' && { headerRight: () => <HeaderRightComponent theme={theme} /> }),
```
Ubax: Nit, which I can follow-up on - this can be done in a similar way using toolbar

```tsx
<Stack.Toolbar placement="right" asChild>
  <HeaderRightComponent theme={theme} />
</Stack.Toolbar>
```

Ubax (approved):

apps/native-component-list/src/navigation/StackConfig.tsx
Kudo: Nice, this works on web too — `asChild` maps to `headerRight` on every platform. Dropped the web-only `headerRight` screen option and moved the platform branch into `SearchToolbar`, so the button lives in one place now. Done in 4a273be38db.

PR #48431, opened 2026-08-03
[sdk-56][brownfield] Fix check-packages failures

intergalacticspacehighway:
## Why?

Same as https://github.com/expo/expo/pull/48113

The cherry-picks of the Fused Library support (https://github.com/expo/expo/pull/47921) and the iOS packaging fixes (https://github.com/expo/expo/pull/48065) to sdk-56 broke the check-packages CI job for expo-brownfield

## How?
Same as https://github.com/expo/expo/pull/48113

## Test

CI should pass


gabrieldonadel (approved):
Thanks for fixing this! LGTM

PR #48432, opened 2026-08-03
[font] add variable fonts support to useFonts on iOS

vonovak:
# Why

we recently added variable font support in #48129 for Android, and this adds it for iOS. iOS already supported variable fonts via the config plugin, this adds that feature to `Font.loadAsync` (and `useFonts` hook) also.

# How

A variable font packs many fonts (=font faces, =postScript names) into one file. So the main part is changing `fontFamilyAliases` to provide a list of fonts (postScript names) for an alias, not just a single one.

Then, as we swizzle `UIFont.fontNames` we can inject these names when RN core requests fonts given an alias. RN core then matches the names against the requested style, and picks the right font to render.

This happens only for fonts with variation axes do this, so for static fonts nothing changes.

docs in https://github.com/expo/expo/pull/48776

# Test Plan

bare expo font screen

# Checklist

---


packages/expo-font/ios/FontFamilyAliasManager.swift
```diff
@@ -1,9 +1,11 @@
 import ExpoModulesCore
 
 /**
- A registry of font family aliases mapped to their real font family names.
+ Maps each font family alias to the PostScript names of the fonts in the file it was registered for.
+ A file usually provides one name. A variable font provides one per named instance, which is what
+ lets `fontWeight` reach the weights it declares.
  */
-private var fontFamilyAliases = [String: String]()
```
vonovak, line 6: to be clear, we always stored postscript names in `fontFamilyAliases` but the comments sound like we stored font families

brentvatne:
## 🤖 AI code review

**Overall PR risk: Medium.** The change modifies existing behavior: the iOS alias registry now maps one family alias to a list of PostScript names instead of a single name, and every `useFonts` and `Font.loadAsync` call on iOS goes through that path. The blast radius stays inside `packages/expo-font` on Apple platforms, with no persisted state and no migration, so a revert of the Swift files restores the previous behavior. The most plausible breakage is font weight selection: when the default PostScript name is absent from the descriptor list, the first entry becomes the lightest named instance, which affects React Native trait fallback and `renderToImageAsync`. Callers that pass a PostScript name to `Font.isLoaded` also see a different result, because the registry now records only the alias.

### 🟡 Warning (1)

- **Variable font name list can start with the wrong weight** — [`packages/expo-font/ios/FontUtils.swift:84`](https://github.com/expo/expo/pull/48432/files#diff-342a4c60cfce1cdaf23be67c62e3818373a76e3dce46413f94986ae52a9aaf10R84) _(correctness)_ · `id:798a2be72301`
  **Confidence:** Medium — the reviewer traced both consumers of the first name, but could not confirm which real font files trigger the condition.<br>**Impact if shipped:** Medium — text and rendered images use a wrong, usually much lighter, weight for some variable fonts.<br>**Suggested remediation:** Put `defaultPostScriptName` at index 0 even when `postScriptNames` does not contain it, because `CTFontManagerRegisterFontsForURL` registers that name too. As an alternative, pick the instance whose `wght` axis value equals the default coordinate of the font.

  <details>

  The new code states the invariant itself: "the first name has to be the default weight". It only holds that invariant when `defaultPostScriptName` is found inside `postScriptNames`.

  `defaultPostScriptName` comes from `CGFont(provider)?.postScriptName`, which is name ID 6 of the font file. `postScriptNames` comes from `kCTFontNameAttribute` of each descriptor, which are the PostScript names of the `fvar` named instances. For a variable font these two sets can differ. Example: `InterVariable.ttf` has name ID 6 `InterVariable`, while its named instances are `Inter-Thin`, `Inter-ExtraLight`, ... `Inter-Regular`, ... `Inter-Black`.

  When `firstIndex(of:)` returns `nil`, the code skips the move and returns `postScriptNames` in Core Text order. `fvar` lists named instances from the lightest weight up, so the first name becomes the thinnest face.

  Two callers then read the wrong face:

  1. `UIFont+FontFamilyAlias.swift` returns the whole list for the alias. RN falls back to the first name when no face matches the requested traits, for example `fontStyle: 'italic'` against an upright-only variable font. The text renders thin.
  2. `FontUtilsModule.swift` line 11 calls `UIFont.fontNames(forFamilyName: options.fontFamily).first`. Before this change an alias always resolved to a single-element list that held the default face, so `renderToImageAsync` was deterministic. Now it can render the thinnest instance instead.

  The base PostScript name is registered by `CTFontManagerRegisterFontsForURL`, so it stays usable even when no descriptor reports it.

  </details>


---

packages/expo-font/ios/FontLoaderModule.swift
```diff
@@ -24,25 +24,25 @@ public final class FontLoaderModule: Module {
...
-      } else {
-        throw FontNoPostScriptException(fontFamilyAlias)
-      }
+      FontFamilyAliasManager.setAlias(fontFamilyAlias, forPostScriptNames: aliasedNames)
+
+      // Only report names the app supplied. This list is what `Font.isLoaded` answers from, and
+      // what `loadAsync` skips on, so every entry takes a name the app can no longer load under.
+      // The names read out of the file stay in the alias registry above, where `fontWeight`
+      // resolution needs them.
+      registeredFonts = Array(Set(registeredFonts).union([fontFamilyAlias]))
```
vonovak, line 47: we only store the alias because it (1) kinda makes sense the user _chose_ an alias for some font and (2) for a variable font that includes 20 postScript names, which would be the one to choose?
the `postScriptName` was added to the list by me previously but I think it was a wrong move tbh.

vonovak:
> ### 🟡 Warning (1)
> The new code states the invariant itself: "the first name has to be the default weight". It only holds that invariant when defaultPostScriptName is found inside postScriptNames.

Not reproducible. `CTFontManagerCreateFontDescriptorsFromURL` returns a descriptor for the axes' default location carrying the file's own PostScript name, so the lookup finds it. Verified on InterVariable, Roboto Flex, Roboto Serif, and a font built with no named instance at the default — the default is present in all four.


The suggested fix would also introduce a bug: in the case where the name really is absent there's no evidence it's registered, and an unresolvable name at index 0 is what RN's fallback calls fontWithName: on — a nil font, worse than a light one.

alanjhughes (approved):

PR #48437, opened 2026-08-03
[ncl] Fix the search screen

Kudo:
# Why

fix broken NCL search screen which might come from expo-router migration

# How

- auto focus the search text input
- fix top search item rendered under the navigation bar
- fix text input not showing on web

# Test Plan

test search on NCL

# Checklist

### DIFF
--- a/apps/native-component-list/src/screens/SearchScreen.tsx
+++ b/apps/native-component-list/src/screens/SearchScreen.tsx
@@ -78,5 +78,12 @@ export default function SearchScreen() {
     []
   );
 
-  return <ComponentListScreen renderItemRight={renderItemRight} apis={apis} sort={false} />;
+  return (
+    <ComponentListScreen
+      contentInsetAdjustmentBehavior="automatic"
+      renderItemRight={renderItemRight}
+      apis={apis}
+      sort={false}
+    />
+  );
 }
--- a/apps/native-component-list/src/screens/SearchScreen.tsx
+++ b/apps/native-component-list/src/screens/SearchScreen.tsx
@@ -84,14 +84,7 @@ export default function SearchScreen() {
     []
   );
 
-  const list = (
-    <ComponentListScreen
-      contentInsetAdjustmentBehavior="automatic"
-      renderItemRight={renderItemRight}
-      apis={apis}
-      sort={false}
-    />
-  );
+  const list = <ComponentListScreen renderItemRight={renderItemRight} apis={apis} sort={false} />;
 
   if (Platform.OS !== 'web') {
     return list;
--- a/apps/native-component-list/src/screens/SearchScreen.tsx
+++ b/apps/native-component-list/src/screens/SearchScreen.tsx
@@ -1,6 +1,12 @@
-import { type NativeStackNavigationOptions, useNavigation } from 'expo-router';
+import {
+  type NativeStackNavigationOptions,
+  type NativeStackNavigationProp,
+  useNavigation,
+} from 'expo-router';
 import Fuse from 'fuse.js';
 import React from 'react';
+import { Platform } from 'react-native';
+import type { SearchBarCommands } from 'react-native-screens';
 
 import { ThemeType, useTheme } from '../../../common/ThemeProvider';
 import ExpoAPIIcon from '../components/ExpoAPIIcon';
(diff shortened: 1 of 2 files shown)


apps/native-component-list/src/screens/SearchScreen.tsx
```diff
@@ -55,5 +84,45 @@ export default function SearchScreen() {
     []
   );
 
-  return <ComponentListScreen renderItemRight={renderItemRight} apis={apis} sort={false} />;
+  const list = (
+    <ComponentListScreen
+      contentInsetAdjustmentBehavior="automatic"
```
Ubax: WDYT about setting this on by default in the `ComponentListScreen`?

Is there any blocker for that?

apps/native-component-list/src/screens/SearchScreen.tsx
```diff
@@ -55,5 +84,45 @@ export default function SearchScreen() {
...
+        placeholder="Search"
+        placeholderTextColor={theme.text.quaternary}
+        value={query}
+        onChangeText={setQuery}
+        style={[
+          styles.webInput,
+          { borderBottomColor: theme.border.default, color: theme.text.default },
+        ]}
+      />
+      {list}
```
Ubax, line 102: This may not work properly. On iOS, in order to support all the interactions with header and tabbar ScrollView needs to be the first children path. 

Though if this is only about the insets, this should be ok

Ubax (commented):
The comments are in regards to code only. I'll give it a test in a moment

apps/native-component-list/src/screens/SearchScreen.tsx
```diff
@@ -43,6 +60,18 @@ export default function SearchScreen() {
...
+  // `autoFocus` is Android-only, so focus the iOS search bar once the screen finishes opening.
+  React.useEffect(() => {
+    if (Platform.OS !== 'ios') {
+      return;
+    }
+    return navigation.addListener('transitionEnd', ({ data }) => {
+      if (!data.closing) {
+        searchBarRef.current?.focus();
+      }
+    });
```
Ubax, line 48: Just a note here, not necessary for this PR - we may consider using tab with `role="search"`, to get the native iOS feeling here

apps/native-component-list/src/screens/SearchScreen.tsx
Ubax: Again not important for this PR, but we may consider using `Stack.SearchBar` instead https://docs.expo.dev/versions/latest/sdk/router/stack/#stacksearchbar

If you think this is a good idea, please create a ticket in linear for me, and I can take care of that later this week

Ubax (approved):

apps/native-component-list/src/screens/SearchScreen.tsx
Kudo: No blocker — it is always `automatic` now (27b43e7881c), and the prop is gone.

I checked on an iOS 26.5 simulator: the APIs list screen is pixel-identical before and after (5 px of antialiasing in a 1206x2422 crop), and the search results still start below the header. The native-stack docs also ask for it on any scrollable under a search bar or large title, so the default matches what the rest of the repo does.

apps/native-component-list/src/screens/SearchScreen.tsx
Kudo: This wrapper is web-only — every other platform returns early above it, so the `FlatList` stays the first native child. `Stack.SearchBar` renders `null`, so it adds no view in front of it either. On web there is no native header or tab bar to interact with, and the wrapper only lays the plain input above the list.

apps/native-component-list/src/screens/SearchScreen.tsx
Kudo: Good idea, and it worked out of the box — switched to `Stack.SearchBar` in 716b94c8952, so no ticket needed. The imperative `setOptions` effect is gone and the web branch is now a plain platform check in the render tree.

Verified on an iOS 26.5 simulator: the search screen is pixel-identical to the `headerSearchBarOptions` version (0 differing pixels below the status bar), and the bottom-anchored iOS 26 search field still autofocuses.

apps/native-component-list/src/screens/SearchScreen.tsx
Kudo: good inspiration idea but i'm not sure if i like the idea
- i was thinking that search is not a common used feature in ncl
- it's ios only
- we have four bottom tabs already. not sure if we want to emphasize the fifth tab as search role
- even the "app store" app doesn't use search role

let's park this and change if separately if necessary

Ubax (approved):

PR #48443, opened 2026-08-03
experiment(cli,metro-config): Add experimental `noxcturnal` transform worker mode

kitten:
# Why

tl;dr: When Babel won't be used (TBD), this new transformer can speed up end-to-end cold bundling times by 2x (3x per-module transform times)

## Context

This branch uses [noxcturnal](https://github.com/expo/noxcturnal), a new internal project, for transforming apps that don't have a custom Babel config or Babel transformer.

> [!WARNING]
> This is an **internal experiment** and not reflective of final changes in SDK 58, or representative of what the implementation, which is currently a **proof of concept**, may look like in the future!

Currently, the transform worker has several modes, of which we're mostly focused on the JS part. The JS pipeline (most of the times) can be thought of as being split into two parts:
- the `babel-preset-expo` run (split itself into: language preset, platform preset, and plugins)
- the post-processing/deps run (`collectDependencies`, `importExport`, etc)

The Babel pipeline is Metro-native, but has a cap on performance. While the cache smooths this over, every transformation has to pay a cost that is almost irreducible for us. Other changes, like the source-map format switch (which is built-into the upcoming, latest version of Metro) do not unlock a high enough performance benefit to ignore Babel indefinitely. And ultimately, when a transform isn't cached, the time-per-module on Babel is higher than we'd like with a ceiling on how much we can improve it.

While Hermes v1 reduced the [amount of plugins](https://github.com/expo/expo/blob/91d17018889b9eaa4e9eb63a757bfa41633e7fde/packages/babel-preset-expo/src/configs/hermes-v1.ts) we have to apply in Babel (and 3 of them are temporary), it hasn't reduced the cost by as much as we'd hoped, since Babel's inherent cost is quite high, to support the rich plugin API it comes with.

## Proposal

`noxcturnal` is an experiment to wholesale replace the Babel pipeline entirely. When a user isn't customising Babel, we're free to replace the Babel pipeline for every module that isn't using `react-native-worklets` or `react-native-codegen`. Every other module — assuming `babel-preset-expo`'s defaults are used — is a known set of transforms.

However, our internal plugins are a shifting target and must stay maintainable, which means `noxcturnal` itself is split into two parts:
- a fully native part
- a JS-driven plugin API layer

The fully native part takes care of Flow transformation (via `hermes-parser`), TypeScript lowering, JSX (modern) transforms, and other language transforms. It also takes care of "language lowering" i.e. transforming language features, such as block scoping, async generators, etc.

However, the JS-drive plugin API layer is an API that _vaguely_ resembles Babel (with more configuration to introduce constraints) that takes care of our "business logic". This is mostly the plugins that live in `babel-preset-expo`'s `plugins` folder (without the few language-feature plugins): https://github.com/expo/expo/tree/91d17018889b9eaa4e9eb63a757bfa41633e7fde/packages/babel-preset-expo/src/plugins

# How

The main pipeline that this is driven by is a NAPI-rs library, which uses `oxc` for parsing, and orchestrates commands to and from the JS side to run the plugins.

Most transforms that are "fixed" (as in, they'll never change), are implemented natively. However, all of our own plugins are implemented in the plugin API.

The latter is an imitation of Rollup's acorn + magic-string concept. The AST is used to navigate the input code, but the plugins then instruct edits to be made, which later collapse into generating output code and source-map edits.

The `functionMap` funcionality from Metro is built into `noxcturnal` however.

When `noxcturnal` rejects a file (due to it being unsupported) or a plugin bails to Babel's worklets or react-native-codegen plugins, the transformer will switch over to the old Babel pipeline. The same happens for every module, and the transformer will immediately bail, if the Babel config (or Babel transformer) have been customised.

> [!NOTE]
> In the future, we can consider several levels of support here such as:
> - Only applying Babel to non-`node_modules` code
> - Only applying Babel as a post-processing step after noxcturnal
> - Applying the two options above partially or dynamically
> - Allow custom transform plugins to be added
> However, many of these features require API changes in Expo first

# Test Plan

- `apps/router-e2e` enables `experiments.noxcturnalTransformWorker` by default
- `apps/native-component-list` enables `experiments.noxcturnalTransformWorker` by default

Not everything will yet be 100% identical, but differential testing on the outputs should ensure that we're very close to identical output. Further modifications and testing should be done after merging this initial implementation.

As for performance, typically we achieve an approx. 2x speed up and a 3x faster per-module transform time. Further improvements are possible.

# Checklist

### DIFF
--- a/apps/native-component-list/app.json
+++ b/apps/native-component-list/app.json
@@ -77,6 +77,7 @@
       }
     },
     "experiments": {
+      "noxcturnalTransformWorker": true,
       "inlineModules": {
         "watchedDirectories": ["src/screens/inlineModules/inlineModulesExamples/"]
       }
--- a/apps/native-component-list/metro.config.js
+++ b/apps/native-component-list/metro.config.js
@@ -27,5 +27,9 @@ config.resolver.assetExts.push(
 
 // Disable Babel's RC lookup, reducing the config loading in Babel - resulting in faster bootup for transformations
 config.transformer.enableBabelRCLookup = false;
+if (process.env.E2E_FORCE_BABEL === '1') {
+  config.transformer.babelTransformerPath =
+    require.resolve('../router-e2e/forced-babel-transformer');
+}
 
 module.exports = config;
--- a/apps/router-e2e/app.config.js
+++ b/apps/router-e2e/app.config.js
@@ -29,6 +29,7 @@ module.exports = {
   jsEngine: process.env.E2E_ROUTER_JS_ENGINE ?? (process.env.E2E_ROUTER_SRC ? 'jsc' : 'hermes'),
   newArchEnabled: true,
   experiments: {
+    noxcturnalTransformWorker: true,
     autolinkingModuleResolution: true,
     baseUrl: process.env.EXPO_E2E_BASE_PATH || undefined,
     tsconfigPaths: process.env.EXPO_USE_PATH_ALIASES,
--- a/apps/router-e2e/forced-babel-transformer.js
+++ b/apps/router-e2e/forced-babel-transformer.js
@@ -1,0 +1,11 @@
+const defaultTransformer = require('../../packages/@expo/metro-config/build/babel-transformer');
+
+// Keep this wrapper observably distinct from Expo's default transformer so the
+// transform worker exercises its pristine-source custom-transformer/Babel path.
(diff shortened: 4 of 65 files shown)


hassankhan (approved):

krystofwoldrich (approved):
Exciting!

Issue #16, opened 2016-11-15
AirGoogleMaps


sonaye:
The IOS app needs an update to link the AirGoogleMaps library which was introduced in the latest version of react-native-maps, this way `<MapView provider='google' />` can be used to view Google Maps on IOS. At the moment, when using `provider='google'` in Exponent the app crashes.

Instructions are available [here](https://github.com/airbnb/react-native-maps/blob/master/docs/installation.md#ios).

brentvatne:
we have decided for the moment to not include this. if there is enough demand we will reconsider!

sonaye:
Any particular reasons? Google Maps are way much better and more mature than Apple's (especially outside the US). I was actually waiting for the support to drop with SDK 12, i mean .. isn't it a straightforward integration?

brentvatne:
it is simple but it adds to the app size for everybody. i hadn't considered that google maps works much better outside of north america, that's a good point. cc @terribleben @ide

chownation:
+1 for Google Maps on iOS






On Thu, Dec 22, 2016 at 4:12 PM -0500, "Brent Vatne" <notifications@github.com> wrote:












it is simple but it adds to the app size for everybody. i hadn't considered that google maps works much better outside of north america, that's a good point. cc @terribleben @ide

sonaye:
@brentvatne wondering how big of a margin are we talking about in terms of the added size? and yes, google maps are much wide spread, apple maps have practically zero data in my region.

tjkang:
same thought with @sonaye. i also want exponent to add google maps.

alexprice91:
I have to have it. Should we just use react-native-maps directly?

AndrewIngram:
Lack of ability to do this is the main reason I can't use Exponent. My design requires that I have custom styles for the map tiles, which is possible with Google Maps, but not Apple Maps.

sonaye:
@brentvatne did you revisit this? any chance of getting our hands on it with sdk 14? we are almost ready to ship and this is one of the very few areas where there is an inadequacy, switching to google maps within a web view is just bad.

terribleben:
We ran into some Exponent-specific complications getting this running well, so it's going to miss the initial SDK 14 release next week. The tentative plan is to get it working in one of the subsequent minor updates (so you don't have to wait all the way until SDK 15).

alexprice91:
Great, thank you!!

chownation:
+1






[...]

PR #48445, opened 2026-08-03
[expo-calendar][ios] Fix crash reading `calendarId` when `EKCalendarItem.calendar` is nil

### LINKED ISSUE #48444: [expo-calendar][iOS] Reading `calendarId` crashes with EXC_BREAKPOINT when `EKCalendarItem.calendar` is nil
### Summary

`EKCalendarItem.calendar` is declared `null_unspecified` in the EventKit headers:

```objc
/*!
    @property   calendar
    @abstract   The calendar that this calendar item belongs to.
    @discussion This will be nil for new calendar items until you set it.
 */
@property(nonatomic, strong, null_unspecified) EKCalendar *calendar;
```

`null_unspecified` imports into Swift as an implicitly unwrapped optional (`EKCalendar!`), so a bare member access on it is a force-unwrap. `expo-calendar` accesses it that way in three places:

- `packages/expo-calendar/ios/Next/CalendarNextModule.swift:384` — `expoEvent.event?.calendar.calendarIdentifier`
- `packages/expo-calendar/ios/Next/CalendarNextModule.swift:547` — `expoReminder.reminder?.calendar.calendarIdentifier`
- `packages/expo-calendar/ios/Conversions/Conversions.swift:111` — `item.calendar.calendarIdentifier`

The `?.` in the first two guards `event` / `reminder`, not `calendar`. When an `EKEvent` or `EKReminder` has a nil `calendar`, reading the `calendarId` property from JS traps the JS thread with `EXC_BREAKPOINT` (`SIGTRAP`) inside `_assertionFailure`, taking down the app. There is no JS-side way to catch it, because the trap happens in the property getter itself.

`Conversions.swift:111` is in `serializeCalendar(item:with:)`, which is used by both the event and the reminder serializers, so the legacy (non-`Next`) API is affected as well.

Note that the same file already uses the safe form for the calendar class at `CalendarNextModule.swift:202` (`expoCalendar.calendar?.calendarIdentifier`), which suggests these three are an oversight rather than intentional.

### Crashing stack

```
Thread: com.facebook.react.runtime.JavaScript          EXC_BREAKPOINT (SIGTRAP)
0  libswiftCore.dylib  closure #1 in closure #1 in _assertionFailure(_:_:file:line:flags:)
1  libswiftCore.dylib  _assertionFailure(_:_:file:line:flags:)
2  <app>               closure #2 in closure #16 in CalendarNextModule.definition()
3  <app>               SyncFunctionDefinition.runBody(_:in:this:arguments:)
...
6  <app>               closure #1 in PropertyDefinition.buildGetter(appContext:in:)
```

### Reproduction

Reliably reproducible on Mac Catalyst (macOS 26.5, Xcode 26.6, `expo-calendar@57.0.0`, SDK 57): list events for a day with `listEvents`, then read `.calendarId` on the results. Some events returned by EventKit on macOS have a nil `calendar`, and the app crashes immediately.

I have not been able to produce a calendar-less event on iOS, so on iOS this appears to be latent rather than actively broken — but it is a force-unwrap of a value Apple explicitly documents as nullable, so it is reachable in principle on any platform.

### Suggested fix

Use optional chaining in all three places, matching what `CalendarNextModule.swift:202` already does. `calendarId` becomes `string | undefined` for these items, which seems strictly preferable to a hard crash. PR incoming.

cvburgess:
# Why

Fixes #48444.

`EKCalendarItem.calendar` is declared `null_unspecified` in the EventKit headers, and its documentation states *"This will be nil for new calendar items until you set it."* `null_unspecified` imports into Swift as `EKCalendar!`, so a bare member access on it is a force-unwrap.

Three call sites access it that way. In two of them the `?.` guards the wrong value — `event` / `reminder` rather than `calendar`:

```swift
expoEvent.event?.calendar.calendarIdentifier       // guards `event`, force-unwraps `calendar`
expoReminder.reminder?.calendar.calendarIdentifier // ditto
item.calendar.calendarIdentifier                   // unguarded
```

When an `EKEvent` or `EKReminder` has a nil `calendar`, reading the `calendarId` property from JS traps the JS thread with `EXC_BREAKPOINT` inside `_assertionFailure` and takes the app down. It cannot be caught from JS, because the trap is inside the property getter.

`Conversions.swift:111` is in `serializeCalendar(item:with:)`, which both the event and reminder serializers call, so the legacy (non-`Next`) API is affected too.

Reproducible on Mac Catalyst (macOS 26.5 / Xcode 26.6): some events EventKit returns on macOS have a nil `calendar`, so listing a day's events and reading `.calendarId` crashes immediately. I could not produce a calendar-less event on iOS, so there it appears latent rather than actively broken — but it is a force-unwrap of a value Apple documents as nullable.

# How

Optional-chain `calendar` in all three places. This matches the safe form already used for the calendar class in the same file:

https://github.com/expo/expo/blob/main/packages/expo-calendar/ios/Next/CalendarNextModule.swift#L202

`calendarId` becomes `string | undefined` for items with no calendar, which seems strictly preferable to a hard crash. The TypeScript types already model the surrounding record loosely enough that this needs no type change; happy to adjust if you'd rather it be explicit.

# Test Plan

- Built and ran a React Native 0.86 / SDK 57 app as a Mac Catalyst target on macOS 26.5.
- Before: navigating between days in a calendar-backed view crashed on every day change with `EXC_BREAKPOINT` in `closure #2 in closure #16 in CalendarNextModule.definition()`.
- After: no crash across repeated day navigation; events with a calendar still resolve `calendarId` normally, and events without one report `undefined`.
- No behavior change on iOS, where `calendar` was already non-nil in testing.

# Checklist

- [x] Documentation is up to date to reflect these changes (no user-facing API change).
- [x] This diff will work correctly for `npx expo prebuild` & EAS Build (no native changes beyond the guarded accesses).

### DIFF
--- a/packages/expo-calendar/CHANGELOG.md
+++ b/packages/expo-calendar/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### 🐛 Bug fixes
 
+- [ios] Fix a crash when reading `calendarId` on an event or reminder whose `EKCalendarItem.calendar` is nil. The property is `null_unspecified` in the EventKit headers, so the bare access was an implicit force-unwrap that trapped the JS thread. ([#48445](https://github.com/expo/expo/pull/48445) by [@cvburgess](https://github.com/cvburgess))
 - [ios] Fix typo in the internal permissions exception name (`MissionPermissionsException` -> `MissingPermissionsException`), which corrects the error code surfaced to JS from `ERR_MISSION_PERMISSIONS` to `ERR_MISSING_PERMISSIONS`. ([#47804](https://github.com/expo/expo/pull/47804) by [@conanm](https://github.com/conanm))
 
 ### 💡 Others
--- a/packages/expo-calendar/ios/Conversions/Conversions.swift
+++ b/packages/expo-calendar/ios/Conversions/Conversions.swift
@@ -108,7 +108,7 @@ func serializeCalendar(item: EKCalendarItem, with formatter: DateFormatter) -> [
   var serailizedItem = [String: Any?]()
 
   serailizedItem["id"] = item.calendarItemIdentifier
-  serailizedItem["calendarId"] = item.calendar.calendarIdentifier
+  serailizedItem["calendarId"] = item.calendar?.calendarIdentifier
   if let title = item.title {
     serailizedItem["title"] = title
   }
--- a/packages/expo-calendar/ios/Next/CalendarNextModule.swift
+++ b/packages/expo-calendar/ios/Next/CalendarNextModule.swift
@@ -381,7 +381,7 @@ public final class CalendarNextModule: Module {
       }
 
       Property("calendarId") { (expoEvent: ExpoCalendarEvent) in
-        expoEvent.event?.calendar.calendarIdentifier
+        expoEvent.event?.calendar?.calendarIdentifier
       }
 
       Property("title") { (expoEvent: ExpoCalendarEvent) in
@@ -544,7 +544,7 @@ public final class CalendarNextModule: Module {
       }
 
       Property("calendarId") { (expoReminder: ExpoCalendarReminder) in
-        expoReminder.reminder?.calendar.calendarIdentifier
+        expoReminder.reminder?.calendar?.calendarIdentifier
       }
 
       Property("title") { (expoReminder: ExpoCalendarReminder) in


Wenszel (approved):
Hey, good catch, thanks!

PR #48448, opened 2026-08-03
[expo-ui][iOS] Keep the custom `MenuView` label on Mac Catalyst

### LINKED ISSUE #48447: [expo-ui][iOS] `MenuView` renders as an AppKit pull-down on Mac Catalyst, replacing the custom label and resizing the trigger
### Summary

When an app is built as a **Mac Catalyst** target with the Mac idiom (`TARGETED_DEVICE_FAMILY` containing `6` — Xcode's "Optimize Interface for Mac", as opposed to "Scale Interface to Match iPad"), SwiftUI resolves `Menu` to an AppKit-style pull-down button. `MenuView` (both `@expo/ui/swift-ui`'s `Menu` and the `community/menu` `MenuView` drop-in) therefore renders as a bordered control with a disclosure chevron, instead of rendering the caller's `label`.

Two consequences:

1. **The custom label is visually replaced.** A `label` slot containing custom content is swallowed by the pull-down's own chrome, so the trigger no longer looks like what the app asked for.
2. **The trigger's measured size changes.** The pull-down brings its own intrinsic sizing, which is wider than a compact custom label. Any layout that sized itself around the label — for example a fixed-diameter circular icon button — is pushed out of shape, shifting the surrounding row.

In my case a 28pt circular status glyph inside a task row became a full-width bordered pill with a chevron, which noticeably reflowed the row.

### Reproduction

Build any app using `MenuView` with a custom `label` as a Mac Catalyst target on macOS 26.5 / Xcode 26.6, with `TARGETED_DEVICE_FAMILY[sdk=macosx*] = "2,6"`. Expo SDK 57, `@expo/ui@57.0.4`, React Native 0.86.

The same code renders correctly on iOS and iPadOS, and also renders correctly on Catalyst under the iPad idiom (`TARGETED_DEVICE_FAMILY = "2"`), because that keeps the iOS control styles.

### Suggested fix

Apply `.menuStyle(.button)` + `.buttonStyle(.plain)` + `.menuIndicator(.hidden)` under `#if targetEnvironment(macCatalyst)`. These are environment-propagating style modifiers, so applying them once to the enclosing view covers every `Menu` branch in `MenuView.body`. This is the non-deprecated spelling of the old `.borderlessButton` menu style — the SDK marks `BorderlessButtonMenuStyle` deprecated with a message pointing at exactly this combination.

I have a PR ready that does this; the change is inert off Catalyst and compiles clean on a Catalyst build.

### Alternative worth considering

Rather than a hardcoded Catalyst branch, `@expo/ui` could expose `menuStyle` and `menuIndicator` as SwiftUI modifiers in `@expo/ui/swift-ui/modifiers`. The package already ships `toggleStyle` and `buttonStyle` there, so these would fit the existing pattern, be useful beyond Catalyst, and let apps opt into whatever presentation they want rather than baking a policy into `MenuView`.

I went with the narrow fix in the PR since it addresses the defect without adding API surface, but I'm happy to implement the modifier route instead if you'd prefer that — just say which you'd rather have.

### Note on platform support

I realise Mac Catalyst isn't an officially supported Expo target. The fix is guarded so it cannot affect iOS, iPadOS, or tvOS, but I understand entirely if you'd rather not carry Catalyst-conditional code — in that case the modifier route above would let this be solved from app code instead.

cvburgess:
# Why

Fixes #48447.

On a **Mac Catalyst** build using the Mac idiom (`TARGETED_DEVICE_FAMILY` containing `6` — Xcode's "Optimize Interface for Mac"), SwiftUI resolves `Menu` to an AppKit-style pull-down button. `MenuView` then renders as bordered chrome with a disclosure chevron rather than rendering the caller's `label`, and the pull-down's own intrinsic sizing widens the trigger.

So a custom label is both visually replaced and re-measured, which reflows whatever layout was built around it. In my app a 28pt circular status glyph inside a task row became a full-width bordered pill with a chevron.

This does not affect iOS, iPadOS or tvOS, and it does not affect Catalyst under the iPad idiom (`TARGETED_DEVICE_FAMILY = "2"`), since that keeps the iOS control styles.

# How

Wrap the existing `if #available` chain in a `Group` and apply three style modifiers to it under `#if targetEnvironment(macCatalyst)`:

```swift
.menuStyle(.button)
.buttonStyle(.plain)
.menuIndicator(.hidden)
```

These are environment-propagating style modifiers, so one application on the `Group` reaches whichever of the six `Menu` branches actually renders — no need to touch each branch.

`.menuStyle(.button)` + `.buttonStyle(.plain)` is the non-deprecated spelling of the old `.borderlessButton` menu style; the SDK marks `BorderlessButtonMenuStyle` deprecated with a message pointing at exactly this combination. `.menuStyle(.button)` is iOS 16+, hence the inner `if #available(iOS 16.0, *)`, which sits inside the Catalyst guard so it costs nothing elsewhere.

The bulk of the diff is re-indentation from introducing the `Group`; the behavioural change is the modifier function at the bottom and the one `.plainMenuOnMacCatalyst()` call.

**On approach:** I went narrow here rather than adding API surface. The alternative — exposing `menuStyle` / `menuIndicator` as modifiers in `@expo/ui/swift-ui/modifiers`, alongside the existing `toggleStyle` and `buttonStyle` — would let apps decide this themselves and would be useful beyond Catalyst. I'm glad to swap this PR for that if you'd prefer; see the issue for more. I also understand if you'd rather not carry Catalyst-conditional code at all, given it isn't an officially supported target.

# Test Plan

- Built a React Native 0.86 / SDK 57 app as a Mac Catalyst target on macOS 26.5 with Xcode 26.6, `TARGETED_DEVICE_FAMILY[sdk=macosx*] = "2,6"`.
- Before: every `MenuView` with a custom `label` rendered as a bordered pull-down with a chevron; a fixed-diameter circular trigger was replaced and its row reflowed.
- After: the custom label renders as the entire trigger at its intended size, and the menu still opens and dispatches actions correctly.
- Verified this exact file compiles clean on a Catalyst build (`** BUILD SUCCEEDED **`, 0 errors).
- iOS behaviour is unchanged by construction — everything is inside `#if targetEnvironment(macCatalyst)`.

# Checklist

- [x] Documentation is up to date to reflect these changes (no public API change).
- [x] This diff will work correctly for `npx expo prebuild` & EAS Build.


packages/expo-ui/ios/Menu/MenuView.swift
```diff
@@ -16,28 +16,64 @@ internal struct MenuView: ExpoSwiftUI.View {
...
+  ///
+  /// No-op off Mac Catalyst.
+  @ViewBuilder
+  func plainMenuOnMacCatalyst() -> some View {
+#if targetEnvironment(macCatalyst)
+    if #available(iOS 16.0, *) {
+      self
+        .menuStyle(.button)
+        .buttonStyle(.plain)
+        .menuIndicator(.hidden)
```
intergalacticspacehighway: Can we introduce these as new modifiers in `packages/expo-ui/src/swift-ui/modifiers/index.ts` instead of adding it to native? We try to avoid adding modifiers directly to native which can be extended from JS to keep things reusable/easier to debug.

intergalacticspacehighway (changes requested):

packages/expo-ui/ios/Menu/MenuView.swift
intergalacticspacehighway: One can check for mac catalyst and pass them via JS when using the component.

cvburgess:
Thanks for the patience @intergalacticspacehighway!

I think this is what you were looking for?

packages/expo-ui/ios/Modifiers/MenuIndicatorModifier.swift
```diff
@@ -0,0 +1,18 @@
...
+
+import ExpoModulesCore
+import SwiftUI
+
+internal struct MenuIndicatorModifier: ViewModifier, Record {
+  @Field var visibility: VisibilityOptions = .automatic
+
+  @ViewBuilder
+  func body(content: Content) -> some View {
+    // `menuIndicator(_:)` is unavailable on tvOS.
```
intergalacticspacehighway: https://developer.apple.com/documentation/swiftui/view/menuindicator(_:)

on the docs it says that tvOS 17+ is supported

packages/expo-ui/ios/Modifiers/MenuStyleModifier.swift
```diff
@@ -0,0 +1,28 @@
...
+  case automatic
+  case button
+}
+
+internal struct MenuStyleModifier: ViewModifier, Record {
+  @Field var style: MenuStyleType = .automatic
+
+  @ViewBuilder
+  func body(content: Content) -> some View {
+    // `ButtonMenuStyle` is unavailable on tvOS; menus there keep the platform default.
```
intergalacticspacehighway: https://developer.apple.com/documentation/swiftui/view/menustyle(_:)

here also same, it says tvOS 17+

packages/expo-ui/ios/Modifiers/MenuIndicatorModifier.swift
cvburgess: Great catch! Honestly never worked with tvOS so I wasn't sure. Thanks for the docs links! Pushed a change I think fixes this and the other comment 🤞

intergalacticspacehighway (approved):

PR #48451, opened 2026-08-03
[router] Abandon and abort loader requests removed by navigation

hassankhan:
# Why

Followup to  #48087.

When navigating to a route with a loader, the loader will start to fetch and React will suspend the route till it does. If the route is changed before the loader settles (pressing back in the browser or swiping back in a stack), there was no logic to clean up the in-flight request, the request would complete and be incorrectly set in the suspense store to be picked up by a future visit, serving potentially stale data.

# How

- `useLoaderData()` registers a "committed" subscription after a suspended route commits
- Each fetch now has an `AbortController`. The controller aborts on abandonment/teardown while the fetch is still in flight
- Every route now has a minimal lifecycle adapter above its Suspense boundary. When that route's shell unmounts, it schedules abandonment of its pending loader path

# Test Plan

- CI
- Manual testing with the `server-loader` fixture in `apps/router-e2e`: navigate to `/slow` in a browser with DevTools open, press back before it settles and see a cancelled network request. Navigate back to `/slow` and ensure a new loader fetch request is triggered.

# Checklist

### DIFF
--- a/packages/expo-router/src/useScreens.tsx
+++ b/packages/expo-router/src/useScreens.tsx
@@ -20,6 +20,9 @@ import { isRouteGuarded, useGuardRedirect, type GuardedRedirects } from './layou
 import { Redirect } from './link/Redirect';
 import { ZoomTransitionEnabler } from './link/zoom/ZoomTransitionEnabler';
 import { ZoomTransitionTargetContextProvider } from './link/zoom/zoom-transition-context-providers';
+import { LoaderRouteLifecycle } from './loaders/LoaderRouteLifecycle';
+import { resolveLoaderPath } from './loaders/resolveLoaderPath';
+import { getContextKey } from './matchers';
 import { unstable_navigationEvents } from './navigationEvents';
 import {
   hasParam,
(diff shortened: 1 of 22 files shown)


packages/expo-router/src/useScreens.tsx
```diff
@@ -303,6 +307,15 @@ export function getQualifiedRouteComponent(value: RouteNode) {
     const InheritedSuspenseFallback = use(SuspenseFallbackContext);
     const redirectHref = useGuardRedirect(value.route);
     const isGuarded = redirectHref !== undefined;
+    const isRouteType = value.type === 'route';
+    const resolvedLoaderPath = React.useMemo(() => {
```
Ubax:
```suggestion
    const resolvedLoaderPath = useMemo(() => {
```

nit, just to follow the convention from this component

Ubax (approved):

packages/expo-router/src/useScreens.tsx
hassankhan: FWIW we still a lot of usage of `React` in this file, we should come back and do a proper pass

kitten (approved):

PR #48456, opened 2026-08-03
[image][android] Overwrite Glide disk cache entries that fail to decode

### LINKED ISSUE #48442: [expo-image][android] glide caches bad responses if status successful

https://github.com/julian-dueck/expo-glide-cache-poisoning-example

### Steps to reproduce

i have an issue with expo-image. we fetch a few webp images from our own servers. a few days ago I added 2 new images, and it turns out we had some issues with our nginx and it returned html instead of the webp image. when we fixed our nginx, I opened our android app again and saw that they still don't load. ios and web (nextjs) worked, but not android.

this is the ReactNativeJS error message
```
{ [Error: Call to function 'ExpoImage.loadAsync' has been rejected.
→ Caused by: Failed to load the image: q3.q: Failed to load resource
There were 12 root causes:
java.lang.RuntimeException(setDataSourceCallback failed: status = 0x80000000)
java.lang.RuntimeException(setDataSourceCallback failed: status = 0x80000000)
java.lang.RuntimeException(setDataSourceCallback failed: status = 0x80000000)
java.io.IOException(Cannot load SVG from stream)
java.lang.RuntimeException(setDataSource failed: status = 0x80000000)
java.lang.RuntimeException(setDataSource failed: status = 0x80000000)
java.lang.RuntimeException(setDataSourceCallback failed: status = 0x80000000)
java.lang.RuntimeException(setDataSourceCallback failed: status = 0x80000000)
java.lang.RuntimeException(setDataSourceCallback failed: status = 0x80000000)
java.io.IOException(Cannot load SVG from stream)
java.lang.RuntimeException(setDataSource failed: status = 0x80000000)
java.lang.RuntimeException(setDataSource failed: status = 0x80000000)
call GlideException#logRootCauses(String) for more detail] code: 'ERR_L' }
```
i don't know the inner workings of expo-image or Glide, but it seems like expo-image somehow cached the nginx html error message and now it tries to load it as a SVG? the remote image is a .webp file.
i load the image this way:
```tsx
const image = useImage(
  `https://our-url/${item.affiliateId}.webp`,
  { maxWidth: 80 },
  [item.affiliateId],
);
```

Steps to reproduce are in the README.md of the reproducible example.

#### THE REPRODUCIBLE REPO HAS A `main` BRANCH WITH THE ISSUE AND A `fix` BRANCH WITH A POTENTIAL FIX
the fix does 2 things:

#### Validate the response body

The Android OkHttp loader now peeks at the first 512 bytes of successful responses and rejects payloads that begin as an HTML document.

The detection:

- Ignores a UTF-8 BOM and leading whitespace.
- Supports an optional XML declaration, as used by XHTML.
- Detects `<!doctype html` and `<html`.
- Does not rely on the HTTP `Content-Type`.

Checking the bytes instead of rejecting `text/html` directly is intentional. A valid image served with an incorrect MIME type can still be decoded, while actual HTML is rejected even if it was labeled as `application/octet-stream`.

The response is rejected before Glide can report it as successfully fetched, preventing new HTML responses from entering the data cache.

#### Evict existing invalid cache entries

This change also handles responses that were cached before the validation was added.

`ExpoImageAppGlideModule` retains access to Glide’s disk cache and registers a global request listener. When a load fails and the nested `GlideException` indicates that decoding was attempted from `DATA_DISK_CACHE`, the listener reconstructs Glide’s original-data cache key and removes only that entry.

#### Comparison with iOS

The iOS implementation uses SDWebImage.

SDWebImage downloads and decodes the response before `SDWebImageManager` stores it in `SDImageCache`. If an endpoint returns HTML, decoding produces `SDWebImageErrorBadImageData`, and the invalid response is not persisted in the image disk cache.

Therefore, after the server is corrected, iOS does not continue reading an HTML document from persistent image storage. It may remember a failed URL for the lifetime of the current manager, but it does not create the persistent disk-cache poisoning observed with Glide on Android.

This change brings the Android behavior closer to the iOS invariant:

> Only data that can be handled as an image should remain associated with the image URL in the persistent cache.

### Environment

```text
    System:
      OS: Linux 7.1 Arch Linux
      Shell: 5.9.2 - /usr/bin/zsh
    Binaries:
      Node: 24.14.1 - /home/julian/.nvm/versions/node/v24.14.1/bin/node
      Yarn: 1.22.22 - /home/julian/.nvm/versions/node/v24.14.1/bin/yarn
      npm: 11.11.0 - /home/julian/.nvm/versions/node/v24.14.1/bin/npm
    SDKs:
      Android SDK:
        API Levels: 34, 35, 36
        Build Tools: 35.0.0, 36.0.0
        System Images: android-36 | Google Play Intel x86_64 Atom, android-36 | 16 KB Page Size Google Play Intel x86_64 Atom
    npmPackages:
      expo: 55.0.28 => 55.0.28
      react: 19.2.0 => 19.2.0
      react-native: 0.83.10 => 0.83.10
    npmGlobalPackages:
      eas-cli: 21.4.0
    Expo Workflow: bare
```

### Expo Doctor Diagnostics

```text
```

intergalacticspacehighway:
# Why

Fixes https://github.com/expo/expo/issues/48442

Upstream issue - https://github.com/bumptech/glide/issues/3550

When an image served from a server is invalid but is served with 200 OK status code, Glide caches the invalid data bytes and never overrwites it. When server returns correct data on the same URL, Glide won't serve the new image until the cache is burst.

# How

- Provide a custom glide disk cache using disk cache configuration - https://github.com/bumptech/glide/wiki/configuration#disk-cache
- In this custom disk cache, we override the default put method with a delete and then put. Glide only calls this put method when it downloads the data. The default put does not allow write if the key exists.

# Test Plan

- `et native-unit-tests -p android --packages expo-image`. The test writes two payloads under one key and asserts the second wins; it fails without the delete.
- Verified Image examples in NCL

# Checklist


julian-dueck:
pls backport to sdk 55 if possible :D

julian-dueck:
bump

julian-dueck:
maybe @tsapeta can take a look? we've been running this patch in prod for a week now and it works great

lukmccall (approved):

intergalacticspacehighway:
Hi @julian-dueck. The PR is picked in sdk 56 and 57, it should be available in their next releases. Thanks for pinging!

PR #48463, opened 2026-08-04
[ios][core] Fix SwiftUI menu dismiss tap passing through to RN views

intergalacticspacehighway:
# Why

The original PR was in https://github.com/expo/expo/pull/47413. This PR simplifies the original PR approach by removing the swizzling/polling.

Issue is that when a Menu/Popover is opened outside taps closes the Menu as well as sends touches to behind RN views. This happens because the menu popover is mounted in the main UIWindow, so it receives touches which is then passed to `RCTSurfaceTouchHandler` which propagates it to RN Views.

This issue does not happen in bare SwiftUI/UIKit app as UIButton/SwiftUI button prevent such touches.

# How

We add a custom touch recognizer which cancels RN's RCTSurfaceHandler touch when Menu background container is tapped.

# Test Plan

Tested Repro in NCL.

## Before

The press counter increases on tapping outside

https://github.com/user-attachments/assets/ed1a6f28-63a8-4e89-94ff-18d4289034b3

## After

The press counter does not increase on tapping outside

https://github.com/user-attachments/assets/03495c3f-0dd1-4c96-b5e0-46e2d95505ae

# Checklist

### DIFF
--- a/packages/expo-modules-core/ios/Core/Views/SwiftUI/SwiftUIHostingView.swift
+++ b/packages/expo-modules-core/ios/Core/Views/SwiftUI/SwiftUIHostingView.swift
@@ -214,6 +214,14 @@ extension ExpoSwiftUI {
     public override func didMoveToWindow() {
       super.didMoveToWindow()
 
+      #if os(iOS)
+      if let window {
+        // SwiftUI content can open a menu, and UIKit passes the tap that closes it through to
+        // React Native underneath. The gate stops that tap from reaching the view below.
+        SystemMenuTouchGate.install(in: window)
+      }
+      #endif
+
       if window != nil, let parentController = reactViewController() {
         #if !os(macOS)
         if parentController as? UINavigationController == nil && parentController as? UITabBarController == nil {
--- a/packages/expo-modules-core/ios/Core/Views/SwiftUI/SystemMenuTouchGate.swift
+++ b/packages/expo-modules-core/ios/Core/Views/SwiftUI/SystemMenuTouchGate.swift
@@ -1,7 +1,8 @@
 // Copyright 2015-present 650 Industries. All rights reserved.
 
 #if os(iOS)
-// Also brings in the subclass header, required to set `state` from the touch callbacks.
+// Also brings in the subclass header, required to set `state` from the touch callbacks
+// and to call `ignore(_:for:)` on React Native's touch handler.
 import UIKit.UIGestureRecognizerSubclass
 
 /**
@@ -24,20 +25,28 @@ internal final class SystemMenuTouchGate: UIGestureRecognizer {
    The one view UIKit adds to the window for a presented context menu.
    */
   static func isContextMenuContainerClassName(_ className: String) -> Bool {
-    return className == "_UIContextMenuContainerView"
+    return className == "_UI".appending("ContextMenuContainerView")
   }
 
   /**
-    Consider Menu to be open when container view is present and interaction on it is enabled.
+    Consider Menu to be open when the container is present, still accepts interaction, and is modal.
+    UIKit turns `isUserInteractionEnabled` off the moment dismissal commits, so taps made during
+    the dismiss animation pass through to the app again. `accessibilityViewIsModal` is how UIKit
+    marks the container as blocking the content behind it — a semantic signal beside the class name.
    */
-  static func isOpenContextMenuContainer(className: String, isUserInteractionEnabled: Bool) -> Bool {
-    return isUserInteractionEnabled && isContextMenuContainerClassName(className)
+  static func isOpenContextMenuContainer(
+    className: String,
+    isUserInteractionEnabled: Bool,
+    accessibilityViewIsModal: Bool
+  ) -> Bool {
+    return isUserInteractionEnabled && accessibilityViewIsModal && isContextMenuContainerClassName(className)
   }
 
   static func isContextMenuContainer(_ view: UIView) -> Bool {
     return isOpenContextMenuContainer(
       className: NSStringFromClass(type(of: view)),
-      isUserInteractionEnabled: view.isUserInteractionEnabled
+      isUserInteractionEnabled: view.isUserInteractionEnabled,
+      accessibilityViewIsModal: view.accessibilityViewIsModal
     )
   }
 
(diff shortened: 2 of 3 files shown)


packages/expo-modules-core/ios/Core/Views/SwiftUI/SystemMenuTouchGate.swift
```diff
@@ -0,0 +1,108 @@
...
+   */
+  static func isContextMenuContainerClassName(_ className: String) -> Bool {
+    return className == "_UIContextMenuContainerView"
+  }
+
+  /**
+    Consider Menu to be open when container view is present and interaction on it is enabled.
+   */
+  static func isOpenContextMenuContainer(className: String, isUserInteractionEnabled: Bool) -> Bool {
+    return isUserInteractionEnabled && isContextMenuContainerClassName(className)
```
Kudo, line 34: [💡 suggestion] Is the container guaranteed to be removed or interaction-disabled the moment the menu starts dismissing? `isUserInteractionEnabled` is the only "menu is open" signal here, so if the container lingers enabled through the dismiss animation, a tap made right after closing the menu is swallowed too — and if it ever lingers enabled for good, every RN tap in the app gets cancelled. Worth confirming on the dismiss path.

packages/expo-modules-core/ios/Core/Views/SwiftUI/SystemMenuTouchGate.swift
```diff
@@ -0,0 +1,108 @@
...
+    cancelsTouchesInView = false
+  }
+
+  // MARK: - Detecting an open menu
+
+  /**
+   The one view UIKit adds to the window for a presented context menu.
+   */
+  static func isContextMenuContainerClassName(_ className: String) -> Bool {
+    return className == "_UIContextMenuContainerView"
```
Kudo, line 27: [💡 suggestion] The description says the bug affects `Menu` *and* `Popover`, but this only matches the context-menu container. `Popover` presents through `UIPopoverPresentationController`, which uses a different container class, so it looks unfixed by this change. Is Popover out of scope here?

Kudo, line 27:
```suggestion
    return className == "_UI".appending("ContextMenuContainerView")
```

in case static code analyzer see the private usage and abandons the app submission

packages/expo-modules-core/ios/Core/Views/SwiftUI/SwiftUIHostingView.swift
```diff
@@ -214,6 +214,14 @@ extension ExpoSwiftUI {
     public override func didMoveToWindow() {
       super.didMoveToWindow()
 
+      #if os(iOS)
+      if let window {
+        // SwiftUI content can open a menu, and UIKit passes the tap that closes it through to
+        // React Native underneath. The gate stops that tap from reaching the view below.
+        SystemMenuTouchGate.install(in: window)
```
Kudo, line 221: should we uninstall the gesture handler when the view move out from a window?

Kudo (changes requested):

packages/expo-modules-core/ios/Core/Views/SwiftUI/SwiftUIHostingView.swift
intergalacticspacehighway: I think it should be fine, it gets installed once (we have a check in `install` function) and will be removed when window is gone, and `touchesBegan` will go to `state = .failed` when menu is not shown so it's mostly noop when Menu is not there. 
SwiftUI menu does not give us open and close callback to register/unregister it. So it's a bit tricky.

packages/expo-modules-core/ios/Core/Views/SwiftUI/SystemMenuTouchGate.swift
intergalacticspacehighway: True, updated!

intergalacticspacehighway: The same class check works for Popover. It uses the same `_UIContextMenuContainerView` 😅

packages/expo-modules-core/ios/Core/Views/SwiftUI/SystemMenuTouchGate.swift
intergalacticspacehighway: `isUserInteractionEnabled` gets set to `false` as soon as someone taps the overlay and it removes the container view after the dismiss animation finishes.

Kudo (approved):

Melzmr:
Hey @intergalacticspacehighway, sorry to mention you, but it seems like the fix doesn't work. Tried with multiple recent 57.0.20 (and earlier) releases. More info [here](https://github.com/expo/expo/issues/48765#issuecomment-5413914105).

I believe it might work in NCL, but it doesn't work in the issue's repro or in my app.

cc: @Kudo

intergalacticspacehighway:
@Melzmr can you upgrade expo to latest 57 (was released 5 hours ago) `expo@57.0.21`. It included this [PR](https://github.com/expo/expo/pull/49775) which is supposed to fix the issue for `onPressIn` and `onPressOut` of RN Views getting triggered.

Melzmr:
> @Melzmr can you upgrade expo to latest 57 (was released 5 hours ago) `expo@57.0.21`. It included this [PR](https://github.com/expo/expo/pull/49775) which is supposed to fix the issue for `onPressIn` and `onPressOut` of RN Views getting triggered.

Thanks for the response! Just verified with `expo@57.0.21` and `expo-modules-core@57.0.17` and it still reproduces unfortunately. I also verified both PRs are in binary. 

<details>
Environment
- expo: 57.0.21
- expo-modules-core: 57.0.17
- @expo/ui: 57.0.17
- react-native: 0.86.3
- os: iOS 26.5
- device: Simulator iOS (simulator)
- arch: new (Fabric)
- engine: Hermes 250829098.0.17
- build: debug
- runtime: bare

Dismiss tap over the underlying Pressable
`onPress / touchStart / touchEnd / pressIn / pressOut` = `3 / 3 / 3 / 3 / 3`</details>

https://github.com/user-attachments/assets/c0537a68-3449-4ffc-b36a-23d31ee202e2

intergalacticspacehighway:
@Melzmr can you try a patch from this [PR](https://github.com/expo/expo/pull/49916) and lmk if it fixes the issue? 

Since i was using an agent to test on simulator and it was using accessibility APIs, it always set `accessibilityViewIsModal` to `true` but that is not the case in normal circumstances so i didn't catch it before  🤦‍♂️

Melzmr:
> @Melzmr can you try a patch from this [PR](https://github.com/expo/expo/pull/49916) and lmk if it fixes the issue?
> 
> Since i was using an agent to test on simulator and it was using accessibility APIs, it always set `accessibilityViewIsModal` to `true` but that is not the case in normal circumstances so i didn't catch it before 🤦‍♂️

I was trying to apply your patch the whole evening yesterday, and the process involved several cache clearings to bypass the prebuilt `expo-modules-core`. I didn't have any luck, so I decided to re-test the repro with the accessibility feature turned on (to see whether it would fix it), and it started working even without it. Looks like the JS was using the updated `node_modules`, but the binary wasn't built from the updated version. A hard clean of all cached binaries resolved it. Sorry for the inconvenience it caused, and thank you for the fixes and responses!

intergalacticspacehighway:
@Melzmr No issues at all! just to confirm again the latest 57 fixes the issue right? Will close this [PR](https://github.com/expo/expo/pull/49916) then.

Melzmr:
> @Melzmr No issues at all! just to confirm again the latest 57 fixes the issue right? Will close this [PR](https://github.com/expo/expo/pull/49916) then.

Yeah, I believe this PR fixed the issue initially. Thanks!

PR #48464, opened 2026-08-04
fix(expo-router): remove double-decoding of route params

### LINKED ISSUE #48421: Route params are encoded once but decoded twice, corrupting AWS SigV4 presigned URLs

https://github.com/OneUne/expo-router-param-double-decode-repro

### Steps to reproduce

Platforms: web and iOS (both reproduce). Environment: development build via `npx expo start` — no native modules are involved, so Expo Go reproduces it too. Package manager: npm.

There are two ways to see it. The first needs no simulator.

**1. Without the app (a few seconds)**

```
git clone https://github.com/OneUne/expo-router-param-double-decode-repro
cd expo-router-param-double-decode-repro
npm install
npm run repro
```

`repro.js` calls expo-router's own functions in the order the router calls them: `resolveHref()` → `parseQueryParams()` → the `decodeURIComponent` that `useLocalSearchParams` applies. It prints the value at each stage and exits 1 while the bug is present.

**Expected:** the value is identical at every stage.
**Actual:** it is intact after `parseQueryParams()` and corrupted after the final decode:

```
step 2 lossless?   true (expected: true)
step 3 lossless?   false (expected: true)
```

**2. In the app**

```
npm run web        # or: npm run ios
```

1. The home screen shows an AWS SigV4 presigned S3 URL containing `%2F` and `%2B` inside `X-Amz-Security-Token`.
2. Tap **Pass it to /viewer**. This runs `router.push({ pathname: '/viewer', params: { targetUrl: PRESIGNED_URL } })` — nothing else, no encoding of our own.
3. The viewer screen reads it back with `useLocalSearchParams` and compares it against what was sent.

**Expected:** a green PASS banner — the value read in `/viewer` is identical to the value passed in.
**Actual:** a red FAIL banner. `%2F` arrived as `/` and `%2B` as `+`:

```
sent      ...&X-Amz-Security-Token=IQoJb3JpZ2luX2Vj%2FPB94qm%2Bx%2B4Ts...
received  ...&X-Amz-Security-Token=IQoJb3JpZ2luX2Vj/PB94qm+x+4Ts...
```

The screen also shows the same value read via `useGlobalSearchParams`, which is corrupted identically — so switching hooks is not a workaround.

**Why this matters beyond the string being wrong:** SigV4 requires every query value to be percent-encoded, and S3 reads a bare `+` in a query string as a space. The token no longer matches the signature, and S3 replies `InvalidToken: The provided token is malformed or otherwise invalid` — an error that gives no hint that the router altered the URL.

**Where the extra decode comes from:**

| Step | Location | Operation |
|---|---|---|
| encode ×1 | `expo-router/build/link/href.js` → `createQueryParams()` | `encodeURIComponent(value)` |
| decode ×1 | `expo-router/build/fork/getStateFromPath-forks.js` → `parseQueryParams()` | reads through `URLSearchParams`, which decodes on access |
| decode ×1 | `expo-router/build/hooks/useLocalSearchParams.js` | `decodeURIComponent(value)` |

`router.push` / `router.navigate` reach the second step through `linkTo(resolveHref(url))` in `build/global-state/router.js`, so this is the ordinary in-app navigation path, not a deep-link edge case.

Previously reported in #35598 and #42075, both closed by the stale bot rather than fixed. #26664 is the same root cause with a different symptom and is still open.

### Environment

```text
    System:
      OS: macOS 26.5.1
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 20.20.0 - /Users/unimac26/.nvm/versions/node/v20.20.0/bin/node
      Yarn: 1.22.22 - /Users/unimac26/.nvm/versions/node/v20.20.0/bin/yarn
      npm: 10.8.2 - /Users/unimac26/.nvm/versions/node/v20.20.0/bin/npm
    Managers:
    SDKs:
      iOS SDK:
    IDEs:
      Android Studio: 2025.2 AI-252.28238.7.2523.14688667
    npmPackages:
      expo: ~57.0.9 => 57.0.9 
      react: 19.2.3 => 19.2.3 
      react-dom: 19.2.3 => 19.2.3 
    npmGlobalPackages:
      eas-cli: 16.32.0
    Expo Workflow: managed
```

### Expo Doctor Diagnostics

```text
```

stareezy-1:
## Summary

Fixes #48421 — Route params are encoded once but decoded twice, corrupting AWS SigV4 presigned URLs (and any value containing percent-encoded characters like `%2F`, `%2B`).

## Root Cause

The encode/decode flow was asymmetric:

| Step | Location | Operation |
|------|----------|-----------|
| encode ×1 | `createQueryParams()` | `encodeURIComponent(value)` |
| decode ×1 | `parseQueryParams()` | `URLSearchParams.getAll()` (auto-decodes) |
| decode ×2 ❌ | `useLocalSearchParams()` | `decodeURIComponent(value)` |
| decode ×2 ❌ | `getRouteInfoFromState()` | `safeDecodeURIComponent(value)` |

After step 2, values are already fully decoded in the navigation state. The additional decodes in steps 3 and 4 are redundant and destructive.

## Changes

- **`src/hooks/useLocalSearchParams.ts`** — Removed `decodeURIComponent()` mapping. Params pass through unchanged since they are already decoded upstream.
- **`src/global-state/getRouteInfoFromState.ts`** — Removed `safeDecodeURIComponent()` mapping and its import.
- **Tests updated** to reflect correct behavior (no double-decode) and added a new test specifically for the SigV4 URL preservation scenario.

## Test Plan

- Updated existing tests in `useLocalSearchParams.test.ios.tsx` and `getRouteInfoFromState.test.ios.ts`
- Added new test case: passing a URL with `%2F` and `%2B` through `router.push` → `useLocalSearchParams` preserves the value exactly

## Impact

This is a **breaking change for code that relied on the double-decode** (e.g., passing `%2Fhello` as a literal param value to `router.setParams()` and expecting it to become `/hello`). However, this was always incorrect behavior — the correct approach is to pass the actual decoded value directly.

### DIFF
--- a/packages/expo-router/src/global-state/__tests__/getRouteInfoFromState.test.ios.ts
+++ b/packages/expo-router/src/global-state/__tests__/getRouteInfoFromState.test.ios.ts
@@ -224,13 +224,13 @@ describe('getRouteInfoFromState', () => {
     expect(result.pathname).toBe('/feed');
   });
 
-  it('decodes URI-encoded params', () => {
+  it('passes through params unchanged (already decoded upstream)', () => {
     const result = getRouteInfoFromState({
       routes: [
         {
           name: '__root',
           state: {
-            routes: [{ name: '[name]', params: { name: 'hello%20world' } }],
+            routes: [{ name: '[name]', params: { name: 'hello world' } }],
             index: 0,
           },
         },
@@ -242,7 +242,10 @@ describe('getRouteInfoFromState', () => {
     expect(result.pathname).toBe('/hello world');
   });
 
-  it('handles malformed URI component (returns as-is)', () => {
+  it('preserves percent-encoded values without double-decoding', () => {
+    // Params containing percent-encoded sequences should pass through unchanged
+    // since decoding already happened upstream (URLSearchParams / getStateFromPath).
+    // This prevents double-decoding that corrupts values like AWS SigV4 URLs.
     const result = getRouteInfoFromState({
       routes: [
         {
@@ -256,7 +259,6 @@ describe('getRouteInfoFromState', () => {
       index: 0,
     });
 
-    // Malformed URI should be returned as-is
     expect(result.params.name).toBe('%E0%A4%A');
     expect(result.pathname).toBe('/%E0%A4%A');
     expect(result.pathnameWithParams).toBe('/%E0%A4%A');
@@ -309,13 +311,13 @@ describe('getRouteInfoFromState', () => {
     expect(result.pathname).toBe('/dashboard');
   });
 
-  it('decodes array params in catch-all', () => {
+  it('passes through array params unchanged (already decoded upstream)', () => {
     const result = getRouteInfoFromState({
       routes: [
         {
           name: '__root',
           state: {
-            routes: [{ name: '[...path]', params: { path: ['hello%20world', 'foo%2Fbar'] } }],
+            routes: [{ name: '[...path]', params: { path: ['hello world', 'foo/bar'] } }],
             index: 0,
           },
         },
--- a/packages/expo-router/src/global-state/getRouteInfoFromState.ts
+++ b/packages/expo-router/src/global-state/getRouteInfoFromState.ts
@@ -1,7 +1,6 @@
 import { INTERNAL_SLOT_NAME, NOT_FOUND_ROUTE_NAME, SITEMAP_ROUTE_NAME } from '../constants';
 import { appendBaseUrl } from '../fork/getPathFromState-forks';
 import type { NavigationState, PartialState } from '../react-navigation/native';
-import { safeDecodeURIComponent } from '../utils/url';
 import type { FocusedRouteState } from './types';
 
 export type UrlObject = {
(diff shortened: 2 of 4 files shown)


kdelay:
Disclosure first: I have an open PR for the same issue (#48423), so read this with that in mind. The findings below are reproducible on `ee52045` and worth fixing in whichever change lands.

Ran in `packages/expo-router` with `NODE_ENV` unset, node 24.10.0.

**1. The suite is red on this branch.** `useGlobalSearchParams > handles encoded params` still asserts the double-decoded value:

```
CI=1 npx jest
Test Suites: 1 failed, 1 skipped, 336 passed, 337 of 338 total
Tests:       1 failed, 6 skipped, 4574 passed, 4581 total

  * useGlobalSearchParams > handles encoded params

    expect(received).toEqual(expected) // deep equality

      Object {
    -   "test": "/hello/world/again",
    +   "test": "%2Fhello%2Fworld%2Fagain",
      }

      at src/hooks/__tests__/useGlobalSearchParams.test.ios.tsx:210:28
```

`useGlobalSearchParams` reads its params out of `getRouteInfoFromState`, so dropping the decode there changes that hook too. The twin test in `useLocalSearchParams.test.ios.tsx` was updated, this one was not.

**2. `oxlint` reports a new warning.** `params` is no longer reassigned once the `Object.fromEntries` block is gone:

```
src/global-state/getRouteInfoFromState.ts:72:7: warning eslint(prefer-const): `params` is never reassigned. help: Use `const` instead.
```

The same command on `main` reports nothing, so this one is introduced here. `let params` -> `const params`.

**3. The updated tests move the inputs rather than the expectations.** In `useLocalSearchParams > handles encoded params`, `router.setParams({ test: '%2Fhello%2Fworld%2Fagain' })` became `router.setParams({ test: '/hello/world/again' })`, and the `router.push` case the same way. `getRouteInfoFromState`'s two tests changed `'hello%20world'` to `'hello world'` and `'foo%2Fbar'` to `'foo/bar'` on the input side.

Those inputs were what pinned the old behaviour, so after the rewrite nothing asserts the thing this PR actually changes: a param the caller passes in already percent-encoded now comes back verbatim instead of decoded. That is the observable behaviour change for existing apps, and it is the case most worth keeping a test on.

Two smaller ones:

- The new `preserves percent-encoded characters without double-decoding` test calls `router.push` with the same value twice, once before `renderHook`. The first call does not affect the assertion.
- There is no `packages/expo-router/CHANGELOG.md` entry. The contributing checklist asks for one, and given point 3 it probably belongs under breaking changes.

`tsc -b tsconfig.all.json` is clean on this branch.

brentvatne:
Closing as a duplicate of https://github.com/expo/expo/pull/48423 (same double-decoding of route params). That PR is the one to continue on. Thanks!

PR #44189, opened 2026-03-24
fix(expo-router): Preserve parenthesized query params when stripping groups

mvincentong:
## Why

Fixes #26664. `stripGroupSegmentsFromPath` was processing the full URL, including query params and hash, as slash-delimited path segments. Parenthesized query or hash values could be matched as route groups and stripped.

This is the `expo-router` half split from the original combined PR per maintainer request. The `jest-expo` Babel config fix moved to #45683.

## How

Separate the pathname from the first `?` or `#`, strip group segments only from the pathname, then append the original suffix unchanged.

Added matcher unit coverage and `router.push` coverage for query params, hash fragments, and combined suffixes with parenthesized values.

## Test Plan

- [x] `git diff --check upstream/main...HEAD`
- [x] `pnpm --filter expo-router test -- src/__tests__/matchers.test.ios.ts src/__tests__/push.test.ios.tsx --runInBand`
- [x] `pnpm --filter expo-router exec eslint --max-warnings 0 src/matchers.tsx src/__tests__/matchers.test.ios.ts src/__tests__/push.test.ios.tsx`
- [x] `pnpm --filter expo-router build`

## Risk

Four-file `expo-router` diff: one helper, two focused tests, and one changelog entry. The implementation preserves the original suffix bytes and only changes group stripping for the pathname portion.

PR #48468, opened 2026-08-04
fix(expo-router): prevent parentheses in query params from being stripped as groups

stareezy-1:
## Summary

Fixes #26664 — `router.push` drops all query parameters when any parameter value contains parentheses (e.g., `(a)`).

## Problem

`getUrlWithReactNavigationConcessions` calls `stripGroupSegmentsFromPath` on the **full path including query string**:

```ts
const pathWithoutGroups = stripGroupSegmentsFromPath(stripBaseUrl(path, baseUrl));
// path = "/test?a=(a)&b=b"
```

When split by `/`, the segment `test?a=(a)&b=b` matches `matchGroupName` regex because it finds `(a)`. The entire segment is then stripped as if it were a route group, losing all query parameters.

## Fix

One-line change — strip the query string before applying group-segment removal:

```ts
const pathWithoutGroups = stripGroupSegmentsFromPath(stripBaseUrl(path, baseUrl).split("?")[0]!);
```

This is safe because:
- `pathWithoutGroups` is only used as `route.path` (the clean pathname)
- Query params are already parsed separately via `parseQueryParams` using `URLSearchParams`

## Test Plan

- `router.push({ pathname: "/test", params: { a: "(a)", b: "b" } })` now correctly passes both params
- Group segments in the actual pathname (e.g., `/(tabs)/home`) are still stripped correctly
- No change to query parameter handling (still parsed via `URLSearchParams`)

## Scope

This is a long-standing bug (open since 2023-12-09) affecting all platforms. The fix is minimal and focused.

### DIFF
--- a/packages/expo-router/src/fork/getStateFromPath-forks.ts
+++ b/packages/expo-router/src/fork/getStateFromPath-forks.ts
@@ -55,7 +55,7 @@ export function getUrlWithReactNavigationConcessions(
   path: string,
   baseUrl: string | undefined = process.env.EXPO_BASE_URL
 ): UrlWithReactNavigationConcessions {
-  const pathWithoutGroups = stripGroupSegmentsFromPath(stripBaseUrl(path, baseUrl));
+  const pathWithoutGroups = stripGroupSegmentsFromPath(stripBaseUrl(path, baseUrl).split('?')[0]!);
 
   let pathname = '';
   let hash = '';


brentvatne:
Closing as a duplicate of https://github.com/expo/expo/pull/44189, which fixes the same parentheses-in-query-params / group-stripping issue with tests. Please continue there. Thanks!

PR #48470, opened 2026-08-04
fix(expo-video): prevent teardown from clearing other modules' media state

### LINKED ISSUE #48068: [expo-video] Teardown can clear Now Playing/media notifications owned by other modules

Source-level reproduction against current `expo/expo` `main` at `f57e436c0e125cb44416ab1eef7e4c002ff69d87`:

- iOS source:
  - `packages/expo-video/ios/VideoPlayer.swift`
  - `packages/expo-video/ios/NowPlayingManager.swift`
- Android source:
  - `packages/expo-video/android/src/main/java/expo/modules/video/player/VideoPlayer.kt`
  - `packages/expo-video/android/src/main/java/expo/modules/video/playbackService/ExpoVideoPlaybackService.kt`

This affects apps that combine `expo-video` with a separate native audio/media module. In that setup, `expo-video` may be used only for muted decorative animated artwork, while the other module owns background audio, lock-screen controls, Now Playing metadata, and/or its own Android media notification.

No private app code is needed to see the ownership issue; it follows directly from the current cleanup paths linked below.

#### iOS

1. In one app process, start an independent native audio/media module that owns `MPNowPlayingInfoCenter.default().nowPlayingInfo` and registers targets on the process-wide `MPRemoteCommandCenter`.
2. Mount an `expo-video` player for muted decorative video. Leave `showNowPlayingNotification` at its default `false`; never set it to `true`.
3. Unmount/release the decorative video player while the independent audio module is still playing.
4. In current `VideoPlayer.swift`, `releasePlayer()` calls `NowPlayingManager.shared.unregisterPlayer(self)` unconditionally: https://github.com/expo/expo/blob/f57e436c0e125cb44416ab1eef7e4c002ff69d87/packages/expo-video/ios/VideoPlayer.swift#L333-L348
5. In current `NowPlayingManager.swift`, `unregisterPlayer(_:)` removes the supplied player, then, if Expo Video's `players` table is empty, removes remote command targets and clears `MPNowPlayingInfoCenter.default().nowPlayingInfo`: https://github.com/expo/expo/blob/f57e436c0e125cb44416ab1eef7e4c002ff69d87/packages/expo-video/ios/NowPlayingManager.swift#L44-L59
6. Because the decorative video player was never registered with `NowPlayingManager` (`showNowPlayingNotification` stayed `false`), this teardown can still clear global media state owned by the independent audio module.

Expected: `unregisterPlayer(_:)` should return immediately if the supplied player is not actually present in Expo Video's registered Now Playing player collection.

Actual: a never-registered decorative video player can trigger cleanup of process-wide Now Playing info and remote command targets when Expo Video's own collection is empty.

#### Android

1. In one app process, run an independent audio/media service that owns its own MediaSession/notification.
2. Use `expo-video` for video artwork or short muted video, with Expo Video notification ownership either disabled or limited to its own sessions.
3. Exercise Expo Video player/service teardown or notification switching.
4. In current `ExpoVideoPlaybackService.kt`, `unregisterPlayer(player)` calls `hidePlayerNotification(player)` before checking whether `mediaSessions` contains that player: https://github.com/expo/expo/blob/f57e436c0e125cb44416ab1eef7e4c002ff69d87/packages/expo-video/android/src/main/java/expo/modules/video/playbackService/ExpoVideoPlaybackService.kt#L98-L105
5. The same service calls `hideAllNotifications()` when the most-recent Expo Video session changes and during cleanup: https://github.com/expo/expo/blob/f57e436c0e125cb44416ab1eef7e4c002ff69d87/packages/expo-video/android/src/main/java/expo/modules/video/playbackService/ExpoVideoPlaybackService.kt#L127-L145 and https://github.com/expo/expo/blob/f57e436c0e125cb44416ab1eef7e4c002ff69d87/packages/expo-video/android/src/main/java/expo/modules/video/playbackService/ExpoVideoPlaybackService.kt#L207-L237
6. `hideAllNotifications()` calls `NotificationManager.cancelAll()`, which is app-wide for this app package. That can remove notifications posted by another audio/media service in the same app, not only Expo Video's notification IDs.

Expected: Expo Video should ignore unregistered players and cancel only notification IDs belonging to Expo Video's own sessions.

Actual: the service cleanup path does not first prove ownership for `unregisterPlayer`, and `hideAllNotifications()` uses app-wide notification cancellation.

### Environment

This is based on current upstream source in `expo/expo` `main`:

- Commit: `f57e436c0e125cb44416ab1eef7e4c002ff69d87`
- `NowPlayingManager.swift` blob: `a8f92a5f4599b8f8a348c6d104c77965afea0892`
- `VideoPlayer.swift` blob: `3c2fde463a10c78546947dacd8c0fda67f0494c7`
- `ExpoVideoPlaybackService.kt` blob: `54c0ad03443f101cb2b6370d78bad14ad4cfa9ed`

I also observed the same relevant cleanup shape in published `expo-video` versions `3.0.16` and `57.0.2`.

### Expo Doctor Diagnostics

Not applicable for this source-level report. This is not tied to a single app configuration; it is an ownership bug in Expo Video's native cleanup paths when an app combines `expo-video` with another native media owner.

### Related issues / PRs checked

- Related but different: https://github.com/expo/expo/issues/46711 and https://github.com/expo/expo/pull/46753 fix the inverse iOS case where `expo-audio` teardown disabled shared remote commands and `expo-video` did not re-enable its own controls.
- That fix confirms the important premise here: `MPRemoteCommandCenter` is process-wide shared state. It does not guard `NowPlayingManager.unregisterPlayer(_:)` against cleanup for players Expo Video never registered.
- Android service history checked: https://github.com/expo/expo/pull/38980, https://github.com/expo/expo/pull/40728, https://github.com/expo/expo/pull/35356, https://github.com/expo/expo/issues/35887, and https://github.com/expo/expo/issues/43628. These cover background playback, remote controls, or service cleanup crashes, but not app-wide notification cancellation or cross-module notification ownership.

### Suggested fix

- iOS: make `NowPlayingManager.unregisterPlayer(_:)` return immediately unless the supplied player is currently present in Expo Video's registered Now Playing player collection. Then only clear Now Playing info / remote command targets when Expo Video is actually unregistering one of its own players and its own collection becomes empty.
- Android: make `unregisterPlayer(player)` return immediately unless `mediaSessions.remove(player)` returns a session owned by Expo Video. Replace `NotificationManager.cancelAll()` with cancellation of Expo Video's own notification IDs, e.g. IDs derived from `mediaSessions` / the previous `mostRecentInteractionSession`.

stareezy-1:
## Summary

Prevents expo-video's player teardown from clearing Now Playing metadata and media notifications that belong to other modules, such as an independent native audio player.

Fixes #48068

## Problem

**iOS:** `VideoPlayer.releasePlayer()` can call `NowPlayingManager.unregisterPlayer(_:)` for a player that never registered for Now Playing. The empty-manager cleanup then clears process-wide `MPNowPlayingInfoCenter` state and remote-command targets owned by another module.

**Android:** `hideAllNotifications()` uses `NotificationManager.cancelAll()`, so notification-session switches and final cleanup can remove every app notification instead of only expo-video's notifications.

## Solution

**iOS:** Return from `unregisterPlayer(_:)` unless the player is present in the manager's registered-player collection.

**Android:**
- Remove the player from `mediaSessions` first and return when it has no owned session.
- Replace app-wide `cancelAll()` with cancellation of the notification IDs belonging to sessions in `mediaSessions`.

## Test Plan

- Expo sandbox verification reproduced the iOS defect on stock `expo-video@57.0.2` and confirmed the fix at `ae7623e`: unregistered release and assigning `showNowPlayingNotification = false` preserve another module's Now Playing metadata, while registered-player release still clears expo-video's own state.
- Android behavior was reviewed from the service/session ownership path; no Android device verification was performed by the Expo sandbox run.
- Existing CI Android unit tests and Native Component List build passed on the original PR head.
- `git diff --check` passes for the review update. Local package checks could not run because this checkout does not have the monorepo dependencies installed.

## Review follow-up

- Removed the redundant `mostRecentInteractionSession` cancellation noted in review; that session is already included in `mediaSessions.values`.
- Added the required `packages/expo-video/CHANGELOG.md` entry.

## Changes

| File | Change |
|------|--------|
| `ios/NowPlayingManager.swift` | Guard unregister cleanup by registered-player ownership |
| `android/.../ExpoVideoPlaybackService.kt` | Guard session unregister and target expo-video notification IDs |
| `packages/expo-video/CHANGELOG.md` | Document the Android/iOS bug fix |

### DIFF
--- a/packages/expo-video/android/src/main/java/expo/modules/video/playbackService/ExpoVideoPlaybackService.kt
+++ b/packages/expo-video/android/src/main/java/expo/modules/video/playbackService/ExpoVideoPlaybackService.kt
@@ -97,9 +97,9 @@ class ExpoVideoPlaybackService : MediaSessionService() {
 
   fun unregisterPlayer(player: ExoPlayer) {
     appContext.mainQueue.launch {
+      val session = mediaSessions.remove(player) ?: return@launch
       hidePlayerNotification(player)
-      val session = mediaSessions.remove(player)
-      session?.release()
+      session.release()
       if (mediaSessions.isEmpty()) {
         cleanup()
         stopSelf()
@@ -233,7 +233,12 @@ class ExpoVideoPlaybackService : MediaSessionService() {
   @MainThread
   private fun hideAllNotifications() {
     val notificationManager: NotificationManager = getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
-    notificationManager.cancelAll()
+    for (session in mediaSessions.values) {
+      notificationManager.cancel(session.player.hashCode())
+    }
+    mostRecentInteractionSession?.let {
+      notificationManager.cancel(it.player.hashCode())
+    }
   }
 
   private fun MediaSession.wantsToShowNotification(): Boolean =
--- a/packages/expo-video/android/src/main/java/expo/modules/video/playbackService/ExpoVideoPlaybackService.kt
+++ b/packages/expo-video/android/src/main/java/expo/modules/video/playbackService/ExpoVideoPlaybackService.kt
@@ -236,9 +236,6 @@ class ExpoVideoPlaybackService : MediaSessionService() {
     for (session in mediaSessions.values) {
       notificationManager.cancel(session.player.hashCode())
     }
-    mostRecentInteractionSession?.let {
-      notificationManager.cancel(it.player.hashCode())
-    }
   }
 
   private fun MediaSession.wantsToShowNotification(): Boolean =
--- a/packages/expo-video/CHANGELOG.md
+++ b/packages/expo-video/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 ### 🐛 Bug fixes
 
+- [Android][iOS] Prevent player teardown from clearing Now Playing metadata and notifications owned by other modules. ([#48470](https://github.com/expo/expo/pull/48470) by [@stareezy-1](https://github.com/stareezy-1))
 - [iOS] Fix races between overlapping source loads and player release. ([#47967](https://github.com/expo/expo/pull/47967) by [@behenate](https://github.com/behenate))
 - [iOS] Set the default `audioMixingMode` to `auto`, [as documented](https://docs.expo.dev/versions/latest/sdk/video/#audiomixingmode); was `doNotMix`. ([#47363](https://github.com/expo/expo/issues/47363) by [@andymatuschak](https://github.com/andymatuschak))
 - When caching take into account Authorization / auth-related request headers. ([#45995](https://github.com/expo/expo/pull/45995) by [@behenate](https://github.com/behenate))
--- a/packages/expo-video/ios/NowPlayingManager.swift
+++ b/packages/expo-video/ios/NowPlayingManager.swift
@@ -42,6 +42,10 @@ class NowPlayingManager: VideoPlayerObserverDelegate {
   }
 
   func unregisterPlayer(_ player: VideoPlayer) {
+    guard players.allObjects.contains(where: { $0 === player }) else {
+      return
+    }
+
     players.remove(player)
     player.observer?.unregisterDelegate(delegate: self)
 


behenate:
@expo-bot review

behenate:
@expo-bot verify

packages/expo-video/android/src/main/java/expo/modules/video/playbackService/ExpoVideoPlaybackService.kt
```diff
@@ -233,7 +233,12 @@ class ExpoVideoPlaybackService : MediaSessionService() {
   @MainThread
   private fun hideAllNotifications() {
     val notificationManager: NotificationManager = getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
-    notificationManager.cancelAll()
+    for (session in mediaSessions.values) {
+      notificationManager.cancel(session.player.hashCode())
+    }
+    mostRecentInteractionSession?.let {
+      notificationManager.cancel(it.player.hashCode())
+    }
```
behenate: The `mostRecentInteractionSession` should already be included in `mediaSessions.values` so this will always be a no-op
```suggestion
```

behenate (changes requested):
Hi thank you for the PR, looks good to me, I left one small note 😄

brentvatne:
@expo-bot verify --fable

packages/expo-video/android/src/main/java/expo/modules/video/playbackService/ExpoVideoPlaybackService.kt
stareezy-1: Good catch — I removed the redundant mostRecentInteractionSession cancellation in 21863f93b6. hideAllNotifications() now cancels only the sessions from mediaSessions.values.

stareezy-1:
@expo-bot Thanks for the thorough iOS verification and the Android ownership analysis. I added the requested expo-video changelog entry in 21863f93b6, applied the maintainer review change, and updated the PR test plan to distinguish measured iOS coverage from Android code review.

PR #48479, opened 2026-08-04
[router] Remove getStateForRouteNamesChage

Ubax:
# Why

Removes the render time set state caused by route names change

# How

1. Remove `getStateForRouteNamesChange` from routers
2. Add `getStateForDeclaredRoutes` - this is filtering logic used to remove the routes removed during HMR in development. It focuses correct screen, and filter non-existing ones
3. Add `ROUTE_NAMES_CHANGED` action, which is dispatched when `routeNames` change
4. Remove `ROUTE_NAMES_CHANGED` from removal prevent detection

# Test Plan

1. CI
2. Router e2e

# Checklist

---


packages/expo-router/src/react-navigation/routers/types.tsx
```diff
@@ -194,7 +176,8 @@ export type Router<State extends NavigationState, Action extends NavigationActio
 
   /**
    * Take the current state and action, and return a new state.
-   * If the action cannot be handled, return `null`.
+   * If the action cannot be handled, return `null`. Custom routers must explicitly handle
+   * `ROUTE_NAMES_CHANGED` to durably reconcile state when their declared routes change.
```
jakub-agent, line 196: [🚨 **critical**] `getStateForRouteNamesChange` was a required method on `Router`, so every router necessarily reconciled. Handling `ROUTE_NAMES_CHANGED` is now opt-in, and opting out fails silently instead of loudly:

- `BaseRouter.getStateForAction` returns `null` for the action.
- `useOnAction.tsx:87` coerces `null` to `state` when `action.target === state.key`, and the new effect always sets `target`.

So the action counts as handled, no unhandled-action warning fires, and the navigator keeps stale `routeNames` for the rest of its life. `navigate()` to a screen declared later then returns `null` forever, and removed routes are only hidden by the render-phase filter.

This is a public extension point: `core/index.tsx` does `export * from '../routers'`, `native/index.tsx` does `export * from '../core'`, and `react-navigation.js` re-exports that build directory. So `useNavigationBuilder`, `BaseRouter` and the `Router` type are all public API. Every in-tree router is fine — `StackRouter` and `TabRouter` handle the action, and `DrawerRouter`, `ui/TabRouter.tsx` and `layouts/stack-router.ts` delegate to them.

Who this actually hits (narrowing a claim from an earlier version of this comment): consumers of the `expo-router/react-navigation` entry point who bring their own router **and** their own `<Screen>` children, since there `getRouteConfigsFromChildren` reads children directly with no filesystem injection, so their `routeNames` can change at runtime. In a file-based expo-router app `routeNames` only changes on Fast Refresh — see my comment on `useNavigationBuilder.tsx`. Narrow audience, but the failure is silent and permanent, which is why I'd still fix it.

Adding a `ROUTE_NAMES_CHANGED` case to `BaseRouter.getStateForAction` would make reconciliation inherited by default rather than opt-in. Either way this needs a `breaking` CHANGELOG entry, together with the `Descriptor.render(): JSX.Element | null` widening.

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
```diff
@@ -754,6 +731,18 @@ export function useNavigationBuilder<
     emitter,
   });
 
+  React.useEffect(() => {
+    const committed = getState();
+
+    if (!isArrayEqual(committed.routeNames, routeNames)) {
```
jakub-agent, line 719: [💡 suggestion] Reconciliation now lands one commit after the screen set changes. `getStateForRouteNamesChange` ran during render and was committed in the same update phase via `useScheduleUpdate`, whose docstring is literally "When screen config changes, we want to update the navigator in the same update phase". This `useEffect` runs after paint, so there is one committed frame with stale `routeNames` — which is what `filterStateForDeclaredRoutes` patches over.

Still worth considering, because it's cheap: at the old call site, `router.getStateForAction(state, { type: 'ROUTE_NAMES_CHANGED', payload: { routeNames } }, { routeNames, routeParamList, routeGetIdList })`, folding a non-`null` result into `nextState`. `getStateForRouteNamesChange` still dies, the existing `useScheduleUpdate` commits it in the same phase, and `filterStateForDeclaredRoutes` becomes unnecessary. The wrinkle is that `getStateForAction` may return `PartialState`, so it needs a `getRehydratedState` pass like the nested-params branch above. Failing that, `useClientLayoutEffect` (already imported here) would at least keep the interim state off screen.

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
```diff
@@ -368,10 +368,6 @@ export function useNavigationBuilder<
   }, {});
 
   const routeNames = routeConfigs.map((config) => config.props.name);
```
jakub-agent, line 343: [🏗️ **design**] Dropping `routeKeyList` / `routeKeyChanges` also drops `navigationKey`. `screens[curr].keys` was its only consumer in the whole package, and the new effect compares route *names* only, so changing `navigationKey` on a `Screen` or a `Group` no longer resets those routes.

The prop is still declared (`core/types.tsx:648` and `:726`) and still validated with a throw in `getRouteConfigsFromChildren` above, so it now accepts a value and silently does nothing.

Unlike my timing comments on this PR, this one isn't Fast-Refresh-scoped. Screen *names* in a file-based app are pinned by the route tree — `getSortedChildren` (`useScreens.tsx:91-155`) injects every filesystem route regardless of what children declare — so `navigationKey` was the only declarative way to force a route to reset without a name change. Removing the key comparison removes that capability outright.

Was that intentional? If so, the prop and its validation should go as well, plus a CHANGELOG note. If not, the effect needs to compare the joined keys alongside the names.

packages/expo-router/src/react-navigation/core/filterStateForDeclaredRoutes.tsx
```diff
@@ -0,0 +1,20 @@
...
+  routeNames: string[]
+): State {
+  const declaredRouteNames = new Set(routeNames);
+  const routes = state.routes.filter((route) => declaredRouteNames.has(route.name));
+
+  if (routes.length === state.routes.length || routes.length === 0) {
+    return state;
+  }
+
+  const survivingRouteCount = state.routes
```
jakub-agent: [💡 suggestion] This focus rule — nearest survivor at or before the old index — matches `StackRouter`, where `reconcileStackRoutes` sets `index = activeRoutes.length - 1`. It does not match `TabRouter`, which falls back to `Math.max(0, focusedIndex)`, i.e. route 0.

Three tabs `[a, b, c]` focused on `c`, then `c` is removed: this render focuses `b`, and the committed `ROUTE_NAMES_CHANGED` then focuses `a` — a flash through a tab the user never selected. `DrawerRouter` reuses `TabRouter`, so it behaves the same. The tab tests here and in `TabRouter.test.tsx` only exercise cases with one surviving route, where both rules agree.

Scope: `routeNames` in a file-based app only changes on Fast Refresh (see my comment on `useNavigationBuilder.tsx`), so this is development-time only — hence `suggestion` rather than something stronger. The correct fallback is router-specific though, so any generic interim index will keep diverging from some router; the synchronous path in that comment removes the need to guess it.

packages/expo-router/src/react-navigation/core/filterStateForDeclaredRoutes.tsx
```diff
@@ -0,0 +1,20 @@
+import type { NavigationState } from '../routers';
+
+export function filterStateForDeclaredRoutes<State extends NavigationState>(
+  state: State,
+  routeNames: string[]
+): State {
+  const declaredRouteNames = new Set(routeNames);
+  const routes = state.routes.filter((route) => declaredRouteNames.has(route.name));
+
+  if (routes.length === state.routes.length || routes.length === 0) {
```
jakub-agent: [💡 suggestion] `routes.length === 0` returns the state untouched, so the interim commit keeps routes for screens that no longer exist. Each of those hits the new `screens[route.name] === undefined` branch in `useDescriptors` and renders `null`, so the navigator paints fully blank for a frame — and with `options: {}`, without the navigator's `screenOptions`, so a `headerShown: false` stack briefly shows a header.

**Correcting an earlier version of this comment:** I said `Stack.Protected` reaches this in production. It does not — see my comment on `useNavigationBuilder.tsx` for why user children can't change `routeNames`. This is a Fast Refresh blank flash, not a production one, so I've dropped the severity.

As far as I can tell this is also the only path that reaches that `useDescriptors` branch, since routes for undeclared screens are otherwise filtered out before `useDescriptors` sees them. Worth a short comment saying why bailing out beats the router's own fallback (`StackRouter` synthesizes `initialRouteName` or `routeNames[0]`); the reason isn't obvious from the code.

packages/expo-router/src/react-navigation/routers/CommonActions.tsx
```diff
@@ -68,14 +68,22 @@ type PreloadAction = {
   target?: string;
 };
 
+type RouteNamesChangedAction = {
+  type: 'ROUTE_NAMES_CHANGED';
```
jakub-agent, line 75: [🏗️ **design**] This puts an internal reconciliation action into the public `Action` union. It's the only member without a matching creator below, and `CommonActions` is exported from `expo-router/react-navigation`, so `navigation.dispatch({ type: 'ROUTE_NAMES_CHANGED', payload: { routeNames: [...] } })` becomes a supported way for app code to rewrite a navigator's `routeNames` to values that don't match its `Screen` children — after which those routes render `null`.

Consider keeping the type local to the routers/core modules, or naming it so it reads as private. `source` is also unused for this action.

packages/expo-router/src/__tests__/hmr-route-changes.test.ios.tsx
```diff
@@ -0,0 +1,230 @@
...
+  expect(screen.getByTestId('details')).toBeVisible();
+  expect(
+    store.navigationRef.current?.getRootState().routes[0]!.state!.routes.map((route) => route.name)
+  ).toStrictEqual(['index', 'details']);
+
+  act(() => router.back());
+  expect(screen.getByTestId('index')).toBeVisible();
+});
+
+it('does not crash when a tab route file is renamed after navigation', () => {
```
jakub-agent, line 234: [💡 suggestion] Both tab cases here leave exactly one surviving route, so they can't observe which route ends up focused, and the headless-tabs case below asserts only that `replace` wasn't called — not that `routeNames` reconciled at all.

A three-tab case (focus the last tab, remove it, assert the final `index`) would pin the behavior discussed on `filterStateForDeclaredRoutes` and catch a regression in either the filter or the router.

packages/expo-router/src/react-navigation/routers/StackRouter.tsx
```diff
@@ -348,6 +299,45 @@ export function StackRouter(options: StackRouterOptions) {
...
+              (route) => route.name === fallbackName
+            );
+            const fallbackRoute =
+              preloadedIndex === -1
+                ? {
+                    key: `${fallbackName}-${nanoid()}`,
+                    name: fallbackName,
+                    params: routeParamList[fallbackName],
+                  }
+                : filteredPreloadedRoutes.splice(preloadedIndex, 1)[0]!;
```
jakub-agent: [🐧 nit] `splice` isn't needed here: `reconcileStackRoutes` already drops preloaded routes whose key is in `activeKeys`, so the promoted route can't be duplicated. `filteredPreloadedRoutes[preloadedIndex]!` reads the same and leaves the array unmutated.

jakub-agent (changes requested):

Reviewed against the stack base `_ubax/07-31-_router_order_tabs_by_routenames_order`, not `main`.

packages/expo-router/src/react-navigation/routers/types.tsx
Ubax: Add changelog entry

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
Ubax: The requirement is to remove the render time set state. Use `useClientLayoutEffect`

packages/expo-router/src/react-navigation/core/filterStateForDeclaredRoutes.tsx
Ubax: Can we just return an empty state? This way the navigator should be unmounted

packages/expo-router/src/react-navigation/routers/CommonActions.tsx
Ubax: Name as internal

packages/expo-router/src/react-navigation/routers/types.tsx
Ubax: Documented this as a breaking change, including the requirement for custom routers to handle `ROUTE_NAMES_CHANGED` and for custom navigators to handle `Descriptor.render()` returning `null`.

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
Ubax: Changed reconciliation to `useClientLayoutEffect`, so the durable router state commits before paint.

packages/expo-router/src/react-navigation/core/filterStateForDeclaredRoutes.tsx
Ubax: Aligned the render filter with router behavior: stacks keep the nearest prior survivor, while tab-like states preserve the focused key or fall back to index 0. Moving reconciliation to `useClientLayoutEffect` also removes the painted interim frame. Added the three-tab regression case.

packages/expo-router/src/__tests__/hmr-route-changes.test.ios.tsx
Ubax: Added a three-tab test that focuses the third tab, removes it, and asserts index 0, the surviving route names, and the index screen.

packages/expo-router/src/react-navigation/routers/CommonActions.tsx
Ubax: Renamed this to exported `InternalRouteNamesChangedAction`, removed it from the public `CommonActions.Action` union, and dropped the unused `source` field. `CommonNavigationAction` includes it internally so router switches remain type-safe.

packages/expo-router/src/react-navigation/core/filterStateForDeclaredRoutes.tsx
Ubax: The filter now returns a transient empty state when no route survives. Core, server-render, JS stack, bottom-tabs, drawer, and native-tabs render paths now handle that state without dereferencing a focused route, and layout-effect reconciliation installs the router fallback before paint.

packages/expo-router/src/react-navigation/routers/StackRouter.tsx
Ubax: Replaced the `splice` with direct index access.

packages/expo-router/src/react-navigation/core/filterStateForDeclaredRoutes.tsx
```diff
@@ -0,0 +1,28 @@
+import type { NavigationState } from '../routers';
+
+export function filterStateForDeclaredRoutes<State extends NavigationState>(
+  state: State,
+  routeNames: string[]
+): State {
+  const declaredRouteNames = new Set(routeNames);
+  const routes = state.routes.filter((route) => declaredRouteNames.has(route.name));
+
+  if (routes.length === state.routes.length) {
```
jakub-agent: [🚨 **critical**] Dropping the old `routes.length === 0` early return means this can now return `{ routes: [], index: 0 }` — an index that points past the end of an empty array. That's the right call for the blank-frame problem, but it makes every render-phase consumer of the focused route a potential `TypeError`, because `useClientLayoutEffect` removes the *painted* frame, not the *rendered* commit. The render still runs with the empty state; only the paint is skipped.

Two reachable sites still dereference it, both in `views/Navigator.tsx` (not in this diff):

- `:124` in `SlotNavigator` — `descriptors[state.routes[state.index]!.key]!.render()`, i.e. a layout using `<Slot />`.
- `:164` in `NavigatorSlot` — `descriptors[state.routes[state.index]!.key]?.render() ?? null`. The descriptor lookup is optional-chained but `state.routes[state.index]!` is not, so it throws at `.key` first.

Both get the filtered state straight from `useNavigationBuilder`, and both use `StackRouter` over `useSortedScreens`, so their `routeNames` is the full filesystem child set. All routes go undeclared exactly when every route file under that layout is replaced in one Fast Refresh — the scenario `repairs state when all screen files are replaced` already covers for `Stack`. `<Slot />` is one of the most common layout primitives, so I'd fix these before merge.

For comparison, the paths that are safe: bottom tabs, drawer and native tabs early-return on `visibleRoutes.length === 0`; `CardStack` and `createStackNavigator` early-return on `state.routes.length === 0`; `useNavigationCache.tsx:162` reads `getState()`, which is the committed state and never empty; and `ExpoRoot.tsx:191` is fine because `INTERNAL_SLOT_NAME` is always declared, so its state can't empty out.

Worth considering `index: -1` for the empty case instead of `0`. `0` reads like a valid focused index and silently yields `undefined`, whereas `-1` makes the "no focused route" state explicit and would have made these two sites fail loudly in typed code.

packages/expo-router/src/react-navigation/bottom-tabs/navigators/createBottomTabNavigator.tsx
```diff
@@ -41,7 +41,7 @@ function BottomTabNavigatorContent({
   const { visibleRoutes, focusedIndex } = useVisibleTabsWithRedirect({
     routes: state.routes,
     routeNames,
-    focusedRouteKey: state.routes[state.index]!.key,
+    focusedRouteKey: state.routes[state.index]?.key,
```
jakub-agent, line 44: [🚨 **critical**] This `!` → `?.` fix is right, and the same fix landed at the two sibling call sites (`drawer/navigators/createDrawerNavigator.tsx:51`, `native-tabs/NativeBottomTabsNavigator.tsx:69`). The fourth one was missed:

`material-top-tabs/navigators/createMaterialTopTabNavigator.tsx:42` still reads

```ts
focusedRouteKey: state.routes[state.index]!.key,
```

It's the identical `useVisibleTabsWithRedirect` call shape, and it throws on the empty state — the argument is evaluated before the `visibleRoutes.length === 0` early return below it can help. Widening the parameter to `string | undefined` in `useVisibleTabsWithRedirect` means TypeScript won't flag the `!`, so nothing catches it.

This is expo-router's own `TopTabs`: `layouts/TopTabsClient.tsx:32` builds it from `createStandardMaterialTopTabNavigator`, whose content component is the one above. So `Tabs`, `Drawer` and `NativeTabs` survive an all-routes-replaced Fast Refresh and `TopTabs` red-screens.

packages/expo-router/src/react-navigation/core/__tests__/filterStateForDeclaredRoutes.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+  expect(result).toEqual({
+    ...state,
+    index: 1,
+    routes: [state.routes[0], state.routes[2], state.routes[3]],
+  });
+  expect(result.history).toBe(state.history);
+  expect(result.routeNames).toBe(state.routeNames);
+});
+
+test('focuses the last survivor before a removed focused route', () => {
```
jakub-agent: [💡 suggestion] The shared fixture is `type: 'test'`, so this case goes down the non-stack branch and what it actually asserts is the index-0 fallback, not "last survivor". The name no longer describes the behavior.

More importantly, nothing here sets `type: 'stack'`, so the stack branch added in this PR has no unit coverage. The two branches only diverge when the focused route is removed **and** more than one route survives before it — every case in this file has at most one, which is why they agree.

A case that pins the stack branch:

```ts
test('focuses the last survivor before a removed focused route in a stack', () => {
  const result = filterStateForDeclaredRoutes({ ...state, type: 'stack' }, [
    'first',
    'removed',
    'last',
  ]);

  expect(result.index).toBe(1);
  expect(result.routes[result.index]).toBe(state.routes[1]);
});
```

That's the same input as the `type: 'tab'` case below, which expects index 0 — so the pair documents the divergence directly. An empty-state case (`routes: []`) would be worth adding too, given it's now a reachable output.

packages/expo-router/src/react-navigation/core/filterStateForDeclaredRoutes.tsx
```diff
@@ -0,0 +1,28 @@
...
+    return state;
+  }
+
+  const index =
+    state.type === 'stack'
```
jakub-agent: [💡 suggestion] This resolves my earlier point about the focus rule diverging from `TabRouter`, but it does it by duplicating router knowledge in core and keying it off a type string. A third-party stack-like router with any other `type` silently gets the tab rule, which is the same class of mismatch as before, just moved.

Since `getStateForRouteNamesChange` is gone, the router can no longer be asked what the interim state should be. If you want to keep the branch, a comment naming the invariant it relies on would help: for `type: 'stack'`, `state.routes[0..index]` are the active routes and everything after `index` is preloaded, so counting survivors in the prefix reproduces `reconcileStackRoutes`. That's non-obvious, and it's the only reason the arithmetic is correct.

jakub-agent (changes requested):

packages/expo-router/src/react-navigation/core/filterStateForDeclaredRoutes.tsx
Ubax: Use `-1`

Use optional chaining for state.routes[..]?.key

packages/expo-router/src/react-navigation/core/filterStateForDeclaredRoutes.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+  if (routes.length === state.routes.length) {
+    return state;
+  }
+
+  const removedBefore = state.routes
+    .slice(0, state.index + 1)
+    .filter((route) => !declaredRouteNames.has(route.name)).length;
+  // Reconciliation replaces this state before paint, so the interim index only needs to be valid.
+  const index = routes.length === 0 ? -1 : Math.max(0, state.index - removedBefore);
```
jakub-agent: [🏗️ **design**] The `state.type === 'stack'` branch is gone and this one formula now covers every navigator. `state.index - removedBefore` is algebraically `survivorsInPrefix - 1`, which is the stack rule — so when the focused route is removed, tab-like states go back to focusing the nearest survivor *before* it instead of route 0.

That re-opens the divergence from `TabRouter.tsx:351`, which uses `Math.max(0, focusedIndex)`, i.e. 0. Three tabs `[a, b, c]` focused on `c`, `c` removed: this returns index 1 (`b`), while the committed reconciliation returns 0 (`a`). The `type: 'tab'` unit test that asserted 0 in the previous revision now asserts 1 for the same input, so I read the change as intentional rather than accidental.

The comment above justifies it with "replaces this state before paint". That's clearly true for anything visual. What I'm not sure about is effects: `useFocusEvents` derives `currentFocusedKey` from this state and emits `focus` / `blur` from passive `useEffect`s, and `useNavigationBuilder` emits its `state` event the same way. If React flushes the interim commit's passive effects before the re-render that the layout effect triggers — my reading is that `flushPassiveEffects()` runs at the start of the sync re-render, but I have not verified that against this React version — then tab `b` gets a spurious `focus` followed by `blur`, firing any `useFocusEffect` on it (analytics screen view, refetch) for a tab the user never selected.

So: a question rather than a claim. Does the interim focus event actually fire? `focuses the first tab when the focused third tab is removed` asserts the final index, so it won't catch this either way — a `useFocusEffect` counter on the middle tab would settle it in a couple of lines.

If it fires, restoring the tab branch is a one-line fix. If it doesn't, this formula is genuinely simpler than the branch and worth keeping — in which case it's worth spelling out *why* in the comment, because "only needs to be valid" is a strong claim about effect ordering that the next reader has to re-derive from scratch.

jakub-agent (commented):

packages/expo-router/src/react-navigation/routers/BaseRouter.tsx
```diff
@@ -1,12 +1,20 @@
 import { nanoid } from 'nanoid/non-secure';
 
+import { filterStateForDeclaredRoutes } from '../core/filterStateForDeclaredRoutes';
```
jakub-agent: [🐧 nit] Reviewing my own commit here: `filterStateForDeclaredRoutes` is now purely the `BaseRouter` default, but it still lives in `core/` and imports `NavigationState` back from `../routers`. The type import is erased so there's no runtime cycle, and `StackRouter` already reaches into `../core/isArrayEqual`, so this follows existing precedent — but the file is arguably a `routers/` concern now, and its name describes the filtering rather than the role.

Not worth churn on its own. Worth folding in if this area is touched again.

jakub-agent (approved):

packages/expo-router/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
...
 - Make `href: null` hide JS tabs and make their routes unreachable, redirecting navigation to the initial tab. ([#48499](https://github.com/expo/expo/pull/48499) by [@Ubax](https://github.com/Ubax))
+- Remove `getStateForRouteNamesChange` from the `Router` interface on `expo-router/react-navigation`. Custom routers must handle the `ROUTE_NAMES_CHANGED` action in `getStateForAction` instead. ([#48479](https://github.com/expo/expo/pull/48479) by [@Ubax](https://github.com/Ubax))
+- `Descriptor.render()` on `expo-router/react-navigation` can now return `null` when the route is no longer declared on the navigator. Custom navigators must handle a `null` element. ([#48479](https://github.com/expo/expo/pull/48479) by [@Ubax](https://github.com/Ubax))
+- Add `getStateForDeclaredRoutes` to the `Router` interface on `expo-router/react-navigation`. Routers spreading `BaseRouter` inherit a default; custom routers that don't must implement it. ([#48479](https://github.com/expo/expo/pull/48479) by [@Ubax](https://github.com/Ubax))
```
```suggestion
```

PR #48486, opened 2026-08-04
[ios][age-range] Fix ageRangeDeclaration mapping

vonovak:
# Why

`AgeRangeService.AgeRangeDeclaration` gained six cases of `AgeRangeService.AgeRangeDeclaration` in iOS 26.2. iOS 26.5 then deprecated them in favor of `confirmed`.

# How

Add the deprecated cases so they fall into `confirmed` which is more correct than falling into `selfDeclared` (the switch default branch).

Also remove `tvos` from podspec - it's not a supported platform.

# Test Plan

green CI

# Checklist

- [ ] `npx expo prebuild` & EAS Build — not relevant, no config plugin or template change
- [ ] Follows the documentation style guide

### DIFF
--- a/packages/expo-age-range/ios/ExpoAgeRange.podspec
+++ b/packages/expo-age-range/ios/ExpoAgeRange.podspec
@@ -11,8 +11,7 @@ Pod::Spec.new do |s|
   s.author         = package['author']
   s.homepage       = package['homepage']
   s.platforms = {
-    :ios => '16.4',
-    :tvos => '16.4'
+    :ios => '16.4'
   }
   s.swift_version  = '6.0'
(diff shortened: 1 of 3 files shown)


packages/expo-age-range/ios/ExpoAgeRange.podspec
```diff
@@ -11,8 +11,7 @@ Pod::Spec.new do |s|
   s.author         = package['author']
   s.homepage       = package['homepage']
   s.platforms = {
-    :ios => '16.4',
-    :tvos => '16.4'
```
vonovak, line 15: not a supported platform

tsapeta (approved):
iOS unit tests check failed, but it's unrelated. Probably #48433 fixed that

PR #48487, opened 2026-08-04
[android][age-range] Migrate to Play age-signals 0.0.4

vonovak:
# Why

`com.google.android.play:age-signals` is pinned to `0.0.3`; the current release is `0.0.4` with new `requestAgeSignalsAccess` function.

# How

Follow https://developer.android.com/google/play/age-signals/release-notes and add `requestAgeSignalsAccess` as well as the new fields it provides. The deprecations are carried over in a way that's least-possible breaking way (though still potentially breaking).

# Test Plan

Somewhat hard to test on my device maybe because I'm not in the right location (these apis were developed due to some US laws). But there are unit tests that cover the functionality with fake responses.

# Checklist

- [x] Added a `CHANGELOG.md` entry, including the `userStatus` value change
- [ ] `npx expo prebuild` & EAS Build — not relevant, no config plugin or template change
- [x] Follows the documentation style guide


packages/expo-age-range/android/src/main/java/expo/modules/agerange/AgeRangeRecords.kt
```diff
@@ -20,29 +20,72 @@ data class AgeRangeResult(
...
+
+internal fun deriveUserStatus(source: Int?, significantChange: Int?): String = when (source) {
+  AgeRangeSource.TIER_A -> "DECLARED"
+  AgeRangeSource.TIER_B -> when (significantChange) {
+    SignificantChangeStatus.PENDING -> "SUPERVISED_APPROVAL_PENDING"
+    SignificantChangeStatus.DECLINED -> "SUPERVISED_APPROVAL_DENIED"
+    else -> "SUPERVISED"
+  }
+  AgeRangeSource.TIER_D -> "VERIFIED"
+  else -> "UNKNOWN"
```
tsapeta: [🚨 **critical**] `TIER_C` collapsing to `"UNKNOWN"` is the one genuine behaviour regression in an otherwise faithful migration, and I think it is based on a misreading of the tier.

Google's [response reference](https://developer.android.com/google/play/age-signals/understand-age-signals-responses) defines `TIER_C` as "User's age is assessed by using credit card, email address, selfie assessment, **Government ID**, or Tax ID" — a real assurance tier that includes government-ID checks, not a bare estimate. `TIER_D` is only stronger in that it *combines* Government ID with a selfie, or uses a Digital ID.

So an app that gates on the deprecated `userStatus` now sees `'UNKNOWN'` for a government-ID-checked user — indistinguishable from "Play told us nothing". That is the most restrictive possible reading of a fairly well-assured account, and it lands on the single field this PR promises to keep working.

Two defensible resolutions, both better than the current one:

1. Map `TIER_C -> "VERIFIED"`. It over-claims slightly relative to `TIER_D`, but the old `userStatus` vocabulary had no finer distinction, and `VERIFIED` is what the pre-0.0.4 API would plausibly have reported for an ID-checked account.
2. Map `TIER_C -> "DECLARED"`. Under-claims, but stays strictly inside the old vocabulary and never *loosens* a gate relative to `TIER_A`.

Either keeps `TIER_C` users distinguishable from the genuinely-unknown case, which `'UNKNOWN'` does not. Whichever you pick, the CHANGELOG line and the `userStatus` TSDoc should name it explicitly so an app author reading the deprecation knows what moved.

If you keep `'UNKNOWN'` deliberately, please say *why* in a comment here — the reasoning ("estimated has no old equivalent") is currently only in the commit message, and the next reader will hit the same doc page I did and reach a different conclusion.

packages/expo-age-range/src/ExpoAgeRange.types.ts
```diff
@@ -45,9 +47,45 @@ export type AgeRangeResponse = {
    * @platform android
    */
   installId?: string | null;
+  /**
+   * The methodology Play Age Signals used to determine the user's age range:
+   * - `'TIER_A'` — unsupervised, self declared.
+   * - `'TIER_B'` — supervised user, meaning the guardian attests to the user's age range.
+   * - `'TIER_C'` — unsupervised, estimated.
```
tsapeta: [🚨 **critical**] `TIER_C` and `TIER_D` are both described inaccurately here.

Per Google's [response reference](https://developer.android.com/google/play/age-signals/understand-age-signals-responses):

- `TIER_C` — "User's age is assessed by using credit card, email address, selfie assessment, Government ID, or Tax ID."
- `TIER_D` — "User's age is checked by using a combination of Government ID and selfie assessment, or Digital ID."

So `TIER_C` is not "unsupervised, estimated" — it covers government-ID and tax-ID checks. And `TIER_D` is not "verified, for example through an ID check" as distinct from C; the actual distinction is that D *combines* Government ID with a selfie, or uses a Digital ID.

This matters more than usual because app authors will gate content on these values, and "estimated" invites them to treat a government-ID-checked user as low-assurance. Suggest wording closer to the source:

```ts
 * - `'TIER_A'` — the user self-declared their age.
 * - `'TIER_B'` — the user's age is managed by a parent or guardian (supervised account).
 * - `'TIER_C'` — the user's age was assessed using a credit card, email address, selfie assessment, government ID, or tax ID.
 * - `'TIER_D'` — the user's age was checked using a combination of government ID and selfie assessment, or a digital ID.
```

`TIER_B` is also worth restating: "supervised user, meaning the guardian attests to the user's age range" is close, but the docs say the age is *managed by* the guardian, which is what pairs it with `significantChangeStatus` below.

packages/expo-age-range/src/AgeRange.ts
```diff
@@ -101,3 +102,24 @@ export async function getRequiredRegulatoryFeaturesAsync(): Promise<
...
+ * - Resolves with `'SHARED'` when the user agrees to share their age signals.
+ * - Resolves with `'NOT_SHARED'` when the user does not agree.
+ * - Resolves with `'VERIFICATION_REQUIRED'` when the user must verify their age before sharing.
+ * - Resolves with `null` when Play Age Signals reports no status, and on iOS and web. On iOS the consent prompt
+ *   is part of `requestAgeRangeAsync` itself, so there is nothing separate to call.
+ * - Rejects when the request fails.
+ *
+ * @platform android
+ */
+export async function requestAgeSignalsAccessAsync(): Promise<AgeSignalsStatus | null> {
```
tsapeta, line 126: [🏗️ **design**] Google documents this as a **precondition** for `requestAgeRangeAsync`, not an optional extra step: "Only call `checkAgeSignals()` if `ageSignalsStatus == SHARED`", and "If `ageSignalsStatus == NOT_SHARED`: Your app won't get age signals in the API response" ([request-age-signals](https://developer.android.com/google/play/age-signals/request-age-signals)).

Right now the two are independent JS functions and the ordering lives only in this TSDoc's "Call this before `requestAgeRangeAsync`". An Android caller who upgrades and doesn't read the release notes keeps calling `requestAgeRangeAsync` alone and silently gets a degraded response — no error, just missing signals. That is the failure mode worth designing against, especially in a package whose whole job is age gating.

Worth an explicit decision between:

1. **Keep it separate (current).** Fine, but then strengthen the docs: state what `requestAgeRangeAsync` returns on Android when access was never requested or came back `NOT_SHARED`/`VERIFICATION_REQUIRED`, and add the same note to `requestAgeRangeAsync`'s TSDoc, not just this one. Right now nothing on `requestAgeRangeAsync` hints that its Android behaviour now depends on a separate call.
2. **Have `requestAgeRangeAsync` call it internally on Android** when access hasn't been granted yet, so the documented Play flow is the default and the separate function becomes the escape hatch for callers who want to control prompt timing. Costs a behaviour change on `requestAgeRangeAsync`, but it is an alpha package and the CHANGELOG is already flagged breaking.

I lean towards (2) for the default-correct behaviour, but (1) with better docs is defensible if you want to keep the prompt under app control. Either way, the `VERIFICATION_REQUIRED` case deserves actionable guidance in the TSDoc — the docs say to "ask the user to visit the Play Store to resolve their status", which is the kind of thing a caller cannot guess from the enum name.

packages/expo-age-range/android/src/test/java/expo/modules/agerange/AgeSignalsManagerTest.kt
```diff
@@ -84,4 +73,51 @@ class AgeSignalsManagerTest {
...
+      onError = { throw AssertionError("Unexpected error: $it") },
+      onCancelled = { throw AssertionError("Unexpected cancellation") }
+    )
+    shadowOf(Looper.getMainLooper()).idle()
+
+    assertEquals("Expected success callback for status $status", true, called)
+    return accessStatus
+  }
+
+  private fun requestAndAwait(result: AgeSignalsResult): AgeRangeResult {
```
tsapeta, line 147: [🏗️ **design**] The new surface is untested, and 0.0.4 shipped the fakes to make it easy: `FakeAgeSignalsManager` gained `setNextAgeSignalsAccessResult(AgeSignalsAccessResult)` and `setNextRequestAgeSignalsAccessException(AgeSignalsException)` in this exact version (verified against the published AAR).

Gaps worth closing, roughly in priority order:

1. **`deriveUserStatus`** — six distinct outcomes and the deliberate behaviour change lives in one of them. This is the mapping most likely to be quietly regressed by a future bump, and it is a pure function, so it is cheap to pin: `TIER_A`, `TIER_B` x {`PENDING`, `DECLINED`, `APPROVED`, null}, `TIER_C`, `TIER_D`, `UNSPECIFIED`, null.
2. **`requestAgeSignalsAccess`** — success (`SHARED`), failure, and cancellation, mirroring the existing `requestAgeRange` coverage. The `requestAndAwait` helper you just extracted makes an `awaitAccess` twin about eight lines.
3. **`ageSignalsStatusToString` / `ageRangeSourceToString` / `significantChangeStatusToString`** — the `UNSPECIFIED`/null branch returning `null` is the interesting one, since it is what distinguishes "Play said nothing" from a real value downstream.

The `else ->` fallbacks that log and return null are hard to reach through the public constants, so I would not contort the tests to hit them — a direct call with an out-of-range `Int` (e.g. `99`) covers them if you want the branch.

The repo convention is red/green (write the failing test first), and here it would have caught the `TIER_C` question above before the mapping was written.

packages/expo-age-range/android/src/main/java/expo/modules/agerange/AgeRangeModule.kt
```diff
@@ -33,6 +35,16 @@ class AgeRangeModule : Module() {
     AsyncFunction("isEligibleForAgeFeaturesAsync") {
       null as Boolean?
     }
+
+    AsyncFunction("requestAgeSignalsAccessAsync") { promise: Promise ->
+      requestAgeSignalsAccess(
+        ageSignalsManager = ageSignalsManager,
+        activity = appContext.throwingActivity,
```
tsapeta, line 42: [🏗️ **design**] `appContext.throwingActivity` is read inside the `AsyncFunction` body, so the throw surfaces as a rejected promise rather than a crash — that part is right.

Two things to check:

1. The `Activity` is passed to Play so it can render the consent prompt. If the activity is destroyed between this call and Play presenting the UI (backgrounding, a config change, process death), what does `requestAgeSignalsAccess` do — reject, hang, or leak? A hung promise is the bad case, because the JS caller has no timeout and no way to tell. If Play doesn't guarantee a terminal callback here, it is worth a note in the TSDoc so callers know a pending promise is possible.
2. `AgeSignalsAccessRequest.builder().setActivity(activity).build()` — the builder does no validation (I checked the AAR: the setters are plain field writes, `build()` never null-checks), so a null or finishing activity fails later and less legibly. `throwingActivity` covers null; a finishing activity it does not.

Also worth confirming whether Play requires this to be invoked from the main thread. Expo's `AsyncFunction` runs off the main thread by default, and the SDK docs are silent on threading. Given it launches `AgeSharingConsentWrapperActivity`, `.runOnQueue(Queues.MAIN)` would be the safe default unless you have verified otherwise on a device.

packages/expo-age-range/src/ExpoAgeRange.types.ts
```diff
@@ -45,9 +47,45 @@ export type AgeRangeResponse = {
...
+   *
+   * `null` for unsupervised accounts, and for supervised accounts with no significant changes yet.
+   *
+   * @platform android
+   */
+  significantChangeStatus?: 'APPROVED' | 'PENDING' | 'DECLINED' | null;
+  /**
+   * The effective date (timestamp) of the most recent significant change that was approved.
+   *
+   * `null` when no significant change has been approved.
```
tsapeta: [💡 suggestion] The `null` note is slightly off. Per the [response reference](https://developer.android.com/google/play/age-signals/understand-age-signals-responses), `significantChangeApprovalDate` "Returns null if no changes exist" — that is, no significant changes *recorded*, which is not the same as "no significant change has been approved".

The distinction is load-bearing next to `significantChangeStatus`: when the status is `'PENDING'` or `'DECLINED'`, changes clearly exist, but this date reflects the last *approved* one and may still be non-null from an earlier approval. "`null` when no significant change has been approved" reads as though `PENDING` implies `null` here, which it does not.

Suggest: "`null` when no significant changes have been recorded for your app. Note that all changes with effective dates before this one are also approved." That second clause is in the docs and is genuinely useful — it tells callers this is a watermark, not a single event.

packages/expo-age-range/src/ExpoAgeRange.types.ts
```diff
@@ -45,9 +47,45 @@ export type AgeRangeResponse = {
...
+   * @platform android
+   */
+  ageRangeSource?: 'TIER_A' | 'TIER_B' | 'TIER_C' | 'TIER_D' | null;
+  /**
+   * Whether a guardian has approved the significant changes recorded for your app:
+   * - `'APPROVED'` — the most recent significant change, and all earlier ones, are approved.
+   * - `'PENDING'` — one or more significant changes are waiting for approval.
+   * - `'DECLINED'` — approval was denied for one or more significant changes.
+   *
+   * `null` for unsupervised accounts, and for supervised accounts with no significant changes yet.
```
tsapeta, line 69: [💡 suggestion] "`null` for unsupervised accounts, and for supervised accounts with no significant changes yet" is a useful clarification, but it is an inference — the underlying value is `UNSPECIFIED` (0) or an absent bundle key, and Play doesn't document that it only populates this for `TIER_B`.

If you have confirmed the unsupervised-account behaviour on a device, keep it. Otherwise the safer phrasing is the mechanical one: "`null` when Play Age Signals reports no status" — matching how you (correctly) worded `ageRangeSource` just above. Over-specifying here is the kind of doc that goes stale silently when Play changes what it populates.

packages/expo-age-range/android/src/main/java/expo/modules/agerange/AgeRangeRecords.kt
```diff
@@ -20,29 +20,72 @@ data class AgeRangeResult(
...
-    mostRecentApprovalDate = result.mostRecentApprovalDate()?.time
+    ageRangeSource = ageRangeSourceToString(result.ageRangeSource()),
+    significantChangeStatus = significantChangeStatusToString(result.significantChangeStatus()),
+    significantChangeApprovalDate = result.significantChangeApprovalDate()?.time,
+    userStatus = deriveUserStatus(result.ageRangeSource(), result.significantChangeStatus()),
+    mostRecentApprovalDate = result.significantChangeApprovalDate()?.time
   )
+}
+
+internal fun deriveUserStatus(source: Int?, significantChange: Int?): String = when (source) {
```
tsapeta: [🐧 nit] `deriveUserStatus` returns a non-null `String` but is assigned to the nullable `userStatus: String?`, so `null` is now unreachable for this field. That is a deliberate improvement over the old code (which produced the literal string `"null"` for an absent value), but it means the TS type's `| null` on `userStatus` is dead on Android.

Not worth changing the TS type — the field is optional and deprecated, and iOS/web omit it entirely — but a one-line note on `deriveUserStatus` saying it always resolves to a member of the old vocabulary (never `null`) would save the next reader the round trip. It is the sort of thing that reads as an oversight rather than a decision.

tsapeta (changes requested):

packages/expo-age-range/android/src/main/java/expo/modules/agerange/AgeRangeRecords.kt
vonovak: Addressed by removal. userStatus and deriveUserStatus are gone, so there is no mapping.

packages/expo-age-range/src/ExpoAgeRange.types.ts
vonovak: Fixed. All four lines rewritten according to https://developer.android.com/google/play/age-signals/understand-age-signals-responses

packages/expo-age-range/src/AgeRange.ts
vonovak: I kept it separate because google also ships 2 APIs for this. Documented the precondition.

packages/expo-age-range/android/src/test/java/expo/modules/agerange/AgeSignalsManagerTest.kt
vonovak: - deriveUserStatus: removed
- requestAgeSignalsAccess: added a failure case
- there are tests for `*toString` methods

packages/expo-age-range/android/src/main/java/expo/modules/agerange/AgeRangeModule.kt
vonovak: 1) hard to tell the right approach; so far I'd keep as-is and not overengineer it


2) > The requestAgeSignalsAccess(Activity) method performs a blocking check of the user's current age signals and age signals sharing status.

from https://developer.android.com/google/play/age-signals/request-age-signals - given it's blocking I'd not put it on the `.runOnQueue(Queues.MAIN)`. It will launch an activity (I supposed) on it's own regardless of what thread it's from.

packages/expo-age-range/src/ExpoAgeRange.types.ts
vonovak: improved the wording

packages/expo-age-range/src/ExpoAgeRange.types.ts
vonovak: this comment is false, in fact. `significantChangeStatus ` docs at https://developer.android.com/google/play/age-signals/understand-age-signals-responses documents say:

null: Returned for:
All unsupervised accounts
Supervised accounts that don't have any significant changes recorded yet.

packages/expo-age-range/android/src/main/java/expo/modules/agerange/AgeRangeRecords.kt
vonovak: removed `deriveUserStatus`

tsapeta (approved):

PR #48489, opened 2026-08-04
[ios][widgets] Fix Live Activity getInstances() filtering and start URL scoping

### LINKED ISSUE #48396: [expo-widgets] LiveActivityFactory.getInstances() returns live activities belonging to other factories, relabelled with the caller's name

https://github.com/MarianoFacundoArch/expo-widgets-getinstances-repro

```bash
git clone https://github.com/MarianoFacundoArch/expo-widgets-getinstances-repro
cd expo-widgets-getinstances-repro
npm install
npx expo run:ios
```

Built and installed successfully on an iOS 26.5 simulator with Xcode 26.6.

### Summary

`LiveActivityFactory.getInstances()` returns **every** live activity in the app, regardless of which factory it is called on, and stamps the **calling** factory's name onto each returned wrapper. Every other part of the Live Activity pipeline discriminates by name — only recovery does not.

With a single Live Activity this is invisible. With two or more, a factory recovers an activity it never started, and the next `update()` overwrites that unrelated activity's content.

### Steps to reproduce

The repro registers two Live Activities with different names:

```ts
const factoryA = createLiveActivity<Props>('ActivityA', LayoutA);
const factoryB = createLiveActivity<Props>('ActivityB', LayoutB);
```

1. Tap **1. factoryA.start()** — one activity is started, on factory **A** only.
2. Tap **2. count getInstances() on A and B**.
3. Tap **3. factoryB.getInstances()[0].update()**.

The app also runs the sequence automatically ~1.5s after mount, logging each step with a `[REPRO]` prefix, so the counts appear in the Metro output without tapping anything.

### Expected behaviour

```
factoryA.getInstances().length = 1
factoryB.getInstances().length = 0
```

Step 3 finds nothing to update — factory B never started an activity.

### Actual behaviour

`factoryB.getInstances()` returns factory A's activity. Step 3 then succeeds and overwrites activity A's content: the banner keeps rendering layout **A** (rendering *does* discriminate by name) while displaying the props factory **B** wrote.

### Cause

`ios/LiveActivityFactory.swift:42-48` — no `name` predicate, no `activityState` predicate, and the wrapper is labelled with the caller's `name`:

```swift
func getInstances() throws -> [LiveActivity] {
  guard #available(iOS 16.1, *) else { throw LiveActivitiesNotSupportedException() }

  return Activity<LiveActivityAttributes>.activities.map { activity in
    LiveActivity(id: activity.id, name: name)
  }
}
```

The discriminator is already persisted — `ios/Widgets/WidgetLiveActivity.swift:6-11`:

```swift
struct LiveActivityAttributes: ActivityAttributes {
  public struct ContentState: Codable, Hashable {
    var name: String
    var props: String
  }
}
```

and the rendering path already reads it, at `WidgetLiveActivity.swift:28`, `:35` and `:59`:

```swift
getLiveActivityNodes(forName: context.state.name, props: context.state.props)
.widgetURL(getLiveActivityUrl(forName: context.state.name))
```

### Why this cannot be worked around in JavaScript

The returned wrapper exposes no identity and no state, so a consumer cannot filter the list itself — `src/Widgets.types.ts`:

```ts
export declare class NativeLiveActivity extends SharedObject<LiveActivityEvents> {
  update(props: string): Promise<void>;
  end(dismissalPolicy?: string, afterDate?: number, state?: string, contentDate?: number): Promise<void>;
  getPushToken(): Promise<string | null>;
}
```

No `id`, no `name`, no `activityState`. This is the same in 55.x and 56.x, so there is no version to move to in order to avoid it.

The missing `activityState` predicate has a second consequence, independent of the multi-activity case: an activity that the system has already ended, or that the user has dismissed, is still returned. A caller that adopts `instances[0]` and calls `update()` on it gets no error, so it believes it holds a live activity when it does not, and never starts a fresh one.

### Suggested fix

```swift
return Activity<LiveActivityAttributes>.activities
  .filter { $0.content.state.name == name }
  .filter { $0.activityState == .active }
  .map { LiveActivity(id: $0.id, name: name) }
```

Separately, exposing `activityState` (and ideally `id`) on `NativeLiveActivity` would let cleanup paths enumerate and end activities that are ended-but-still-visible, instead of tracking them out of band.

### Environment

- `expo-widgets`: 57.0.7 (identical code in 56.0.25, 55.0.20 and 55.0.2)
- `@expo/ui`: 57.0.8
- Expo SDK: 57.0.9
- React Native: 0.86.2
- Xcode 26.6, iOS 26.5 simulator
- Platform: iOS

### Notes

Unrelated to #43646 (the bundle-copy issue), fixed in 56.0.0. Reopening as a new issue with a repro, after #48395 was auto-closed for lacking one.

### LINKED ISSUE #48400: [expo-widgets] Live Activity start URL never reaches the Lock Screen banner (only the Dynamic Island)

https://github.com/MarianoFacundoArch/expo-widgets-getinstances-repro — see "Scenario 2" in the README.

```bash
git clone https://github.com/MarianoFacundoArch/expo-widgets-getinstances-repro
cd expo-widgets-getinstances-repro
npm install
npx expo run:ios
```

Builds and installs on an iOS 26.5 simulator with Xcode 26.6.

### Summary

The URL passed to `LiveActivityFactory.start(props, url)` can only ever reach the Dynamic Island. The Lock Screen banner never receives it, so tapping a Live Activity on the Lock Screen cannot deep-link into the app.

### Cause

In `ios/Widgets/WidgetLiveActivity.swift`, `.widgetURL(...)` is chained onto the value returned by the `dynamicIsland:` closure — outside the `ActivityConfiguration` content closure that builds the banner:

```swift
public var body: some WidgetConfiguration {
  ActivityConfiguration(for: LiveActivityAttributes.self) { context in
    let nodes = getLiveActivityNodes(forName: context.state.name, props: context.state.props, environment: environment)
    LiveActivityBannerView(context: context, nodes: nodes)          // no .widgetURL
  } dynamicIsland: { context in
    let nodes = getLiveActivityNodes(forName: context.state.name, props: context.state.props, environment: environment)
    return DynamicIsland {
      ...
    }
    .widgetURL(getLiveActivityUrl(forName: context.state.name))     // only here
  }
}
```

The banner is built in the first closure and nothing applies a URL to it.

### Steps to reproduce

1. Start a Live Activity with a URL: `factory.start(props, 'myapp://somewhere')`.
2. Lock the device and tap the Live Activity banner on the Lock Screen.

### Expected

The app opens at `myapp://somewhere`.

### Actual

The app opens at whatever screen it was last on; the URL is never delivered.

### Suggested fix

Apply the modifier inside the content closure as well, e.g.

```swift
LiveActivityBannerView(context: context, nodes: nodes)
  .widgetURL(getLiveActivityUrl(forName: context.state.name))
```

### Workaround

Applying `widgetURL()` from `@expo/ui/swift-ui/modifiers` to the root of the `banner` section of the Live Activity layout does work, because the layout is rendered through the normal node/modifier pipeline. That makes the `url` argument of `start()` redundant for the banner.

### Environment

- `expo-widgets`: 57.0.7 — the same structure is present in 56.0.25, 55.0.20 and 55.0.2
- Expo SDK 57.0.9, React Native 0.86.2, `@expo/ui` 57.0.8
- Xcode 26.6, iOS 26.5 simulator

### Notes

Related but distinct from #48396 (`getInstances()` returning other factories' activities). On the 55.x line this is additionally masked by the String/Data mismatch in `getLiveActivityUrl`, filed separately.

### LINKED ISSUE #48401: [expo-widgets] Live Activity start URL is factory-global and never cleared; unreadable on the 55.x line

https://github.com/MarianoFacundoArch/expo-widgets-getinstances-repro — see "Scenario 3" in the README.

```bash
git clone https://github.com/MarianoFacundoArch/expo-widgets-getinstances-repro
cd expo-widgets-getinstances-repro
npm install
npx expo run:ios
```

Builds and installs on an iOS 26.5 simulator with Xcode 26.6.

### Summary

The Live Activity start URL is stored in a way that makes it factory-global and permanent: it is written only when non-nil and never cleared. A later `start(props)` without a URL silently inherits the previous one, and two activities registered under the same name share whichever URL was stored last.

Separately, on the whole 55.x line the value cannot be read back at all.

### Cause

`ios/LiveActivityFactory.swift`:

```swift
if let url {
  WidgetsStorage.set(url.absoluteString, forKey: "__expo_widgets_live_activity_\(name)_url")
}
```

**(a) Never cleared — 55.x through 57.0.7.** There is no `else` branch and no removal anywhere. The key outlives the activity that set it, so:

```ts
factory.start(props, 'myapp://a');   // stores myapp://a
activity.end('immediate');
factory.start(props);                 // no url — still opens myapp://a
```

**(b) Write/read type mismatch — 55.x only.** It is written with the `String` overload of `WidgetsStorage.set`, and `ios/Widgets/Utils.swift` reads it back with `getData(forKey:)`:

```swift
func getLiveActivityUrl(forName name: String) -> URL? {
  let data = WidgetsStorage.getData(forKey: "__expo_widgets_live_activity_\(name)_url")
  ...
}
```

A `String` in `UserDefaults` is not retrievable as `Data`, so this always returns `nil` and the URL passed to `start()` is inert on 55.x. 56.0.0 changed the read to `getString`; it was never backported.

### Expected

- A `start()` call without a URL produces an activity with no URL.
- The URL is scoped to the activity that set it.

### Actual

- The last URL stored under that factory name is reused indefinitely.
- On 55.x, no URL is ever readable regardless.

### Suggested fix

```swift
if let url {
  WidgetsStorage.set(url.absoluteString, forKey: key)
} else {
  WidgetsStorage.removeObject(forKey: key)
}
```

`WidgetsStorage.removeObject(forKey:)` already exists. Storing the URL in the activity's `ContentState` alongside `name` and `props` would scope it correctly and remove the shared-key problem entirely.

For 55.x specifically, the read at `Widgets/Utils.swift` should use `getString` (as 56.0.0 does) — or the write should store UTF-8 `Data`.

### Environment

- `expo-widgets`: 57.0.7 for (a); 55.0.2 and 55.0.20 additionally exhibit (b)
- Expo SDK 57.0.9, React Native 0.86.2, `@expo/ui` 57.0.8
- Xcode 26.6, iOS 26.5 simulator

### Notes

Distinct from #48396 (`getInstances()` ignoring the activity name) and #48400 (the banner never receiving the URL), though all three touch the same Live Activity plumbing.

huextrat:
# Why

Fixes three related bugs in the expo-widgets Live Activity plumbing:

- Closes #48396 — `LiveActivityFactory.getInstances()` returns every live activity in the app regardless of which factory it is called on, stamps the caller's name onto each wrapper, and also returns activities that have already ended or been dismissed. With two factories, calling `update()` through such a wrapper overwrites an unrelated activity's content.
- Closes #48400 — the URL passed to `start(props, url)` is only applied to the Dynamic Island; the Lock Screen banner never receives it, so tapping the banner cannot deep-link into the app.
- Closes #48401 — the URL is stored in `WidgetsStorage` under a key derived from the factory name: written only when non-nil, never cleared, and shared by every activity started under that name. A later `start()` without a URL silently inherits the previous one.

# How

**`getInstances()`** now filters `Activity<LiveActivityAttributes>.activities` by the persisted `content.state.name` and keeps only `.active`/`.stale` activities — a stale activity is still visible and updatable, while ended/dismissed ones are not. The availability guard moves from iOS 16.1 to 16.2 because `Activity.content` requires it; `start()` already required 16.2, so no activity created by this module can exist below that.

**The start URL** moves from `WidgetsStorage` into `LiveActivityAttributes`. Attributes are fixed at `Activity.request` and never rewritten, which is exactly the semantics of a per-`start()` deep link: scoped to one activity, unaffected by content updates, nothing to clear. `.widgetURL(context.attributes.url)` is applied in both the Lock Screen banner closure and the Dynamic Island builder (previously only the island had it). The now-unused `getLiveActivityUrl` helper and `__expo_widgets_live_activity_<name>_url` storage key are removed.

The new attribute is optional, so existing activities started by a previous version (attributes decode as `{}`) and push-to-start payloads that don't include it keep working; they simply carry no URL.

Notes for reviewers — small deliberate behavior changes:

- On iOS 16.1 exactly, `getInstances()` now throws `LiveActivitiesNotSupportedException` instead of returning `[]` (consistent with `start()` on that version; no activity can exist there anyway).
- Ended-but-still-visible activities are no longer enumerable through `getInstances()`, matching its documented "currently active instances" contract. Exposing `id`/`activityState` on `NativeLiveActivity` so cleanup paths can enumerate them (the secondary ask in #48396) is left for a follow-up PR.
- An activity started before an app update loses its deep link after the update (its attributes were encoded without a URL, and the storage key is no longer read). Transient by nature — Live Activities live at most ~12 h.

# Test Plan

- `et check-packages expo-widgets` passes (build, typecheck, lint, format).
- To reproduce, use the repro app from the issues: https://github.com/MarianoFacundoArch/expo-widgets-getinstances-repro (scenarios 1–3 in its README, `[REPRO]`-prefixed logs in Metro):
  - Scenario 1: after `factoryA.start()`, `factoryA.getInstances().length == 1` and `factoryB.getInstances().length == 0`; `factoryB.getInstances()[0].update()` no longer overwrites activity A.
  - Scenario 2: with `factory.start(props, 'myapp://somewhere')`, tapping the Lock Screen banner now opens the app at `myapp://somewhere`.
  - Scenario 3: after `end('immediate')` followed by `start(props)` without a URL, the new activity carries no URL.

# Checklist

- [ ] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: updated a module plugin). — no plugin/prebuild changes; iOS runtime code only.


jakex7 (approved):
Thank you for contributing!

huextrat:
Replying to the [AI review](https://github.com/expo/expo/pull/48489#issuecomment-5216554011) point by point:

**1. Push-to-start payload can now set the Live Activity tap URL** — acknowledged, leaving as is. `widgetURL` never opens the browser: tapping always opens the containing app, and the URL is delivered to the app's own link handler (this resolves the reviewer's stated uncertainty). So the payload gains no more power than any push notification carrying a deep link, and a sender with push-to-start access already controls `name` and `props` — the entire rendered content. A scheme/host allowlist in the library would impose policy it can't know (custom schemes, universal-link hosts vary per app), and a server-chosen URL is arguably a feature for push-to-start. Apps must validate incoming URLs in their link handler regardless of source.

**2. Live Activities started before the app update lose their deep link** — acknowledged (it's also called out in the PR description) and deliberately accepted. One caution about the suggested remediation: the bare fallback `context.attributes.url.flatMap(URL.init(string:)) ?? getLiveActivityUrl(forName: context.state.name)` would reintroduce #48401 — the legacy `__expo_widgets_live_activity_<name>_url` keys are never removed, so any *new* activity started without a URL would silently pick up the stale stored URL through the fallback, which is the exact bug this PR fixes. A correct shim needs the fallback *plus* unconditional key removal in `start()`, and a reminder to delete it a release later. Given the affected window — an activity already running at the moment of an app update, for the remainder of its ≤12 h lifetime — the transient regression seems preferable to a stateful migration path. Happy to add the corrected shim if maintainers prefer.

**3. Lock Screen banner always applies widgetURL, so it can override a layout widgetURL** — good catch, fixed in 028ab24d30. `.widgetURL` is now applied only when the activity actually has a URL, on both the Lock Screen banner and the Dynamic Island. The island had the always-applied pattern before this PR too (with the possibly-nil storage lookup), so this also removes a pre-existing conflict with layouts that set `widgetURL` on island sections.

huextrat:
@jakex7 FYI, point 3 of the review is fixed in 028ab24d3 (conditional widgetURL on banner + island)

jakex7 (approved):

PR #48491, opened 2026-08-04
[docs] Pulsar as a recommended solution for advanced haptics control

piaskowyk:
# Why

Developers using expo-haptics who need more than the built-in feedback types (custom patterns, additional presets, gesture-driven haptics, or running haptics on worklets) currently have no pointer from the docs toward a more capable solution. This adds a recommendation to [Pulsar](https://docs.swmansion.com/pulsar/) for those advanced use cases.

This follows the existing docs convention of pointing users to a more specialized library when it's a better fit — for example the identity-provider recommendations in the [AuthSession docs](https://docs.expo.dev/versions/latest/sdk/auth-session/) and the alternative-library note in the [Maps docs](https://docs.expo.dev/versions/latest/sdk/maps/).

# How

Added an > **info** callout to the Haptics SDK reference page, placed after the platform-support intro and before the Installation section. The note briefly explains what Pulsar adds over expo-haptics (custom patterns, more presets, gesture-based haptics, worklet support) and links to its documentation.

The same note was added to both the unversioned and v57.0.0 (latest) pages; older frozen versions were intentionally left untouched.

# Test Plan

Docs-only change. Verified locally that the callout renders as an info block and the link resolves. To reproduce:

cd docs && pnpm run dev
Then open /versions/unversioned/sdk/haptics/ and confirm the Pulsar info callout appears between the web-support notes and the Installation heading.

# Checklist

### DIFF
--- a/docs/pages/versions/unversioned/sdk/haptics.mdx
+++ b/docs/pages/versions/unversioned/sdk/haptics.mdx
@@ -31,6 +31,8 @@ On web, the library uses the Web Vibration API. Note the following:
 - The user must grant permission to use vibration (usually automatic)
 - Some browsers may ignore vibration in certain contexts (for example, background tabs)
 
+> **info** For more advanced control over haptics, we recommend [Pulsar haptics SDK](https://docs.swmansion.com/pulsar/). It lets you create custom haptic patterns, use a wider set of presets, build gesture-based haptics, and run haptics on worklets.
+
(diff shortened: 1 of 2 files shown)


amandeepmittal (commented):
Thanks for the addition!

amandeepmittal (approved):

PR #48497, opened 2026-08-04
[cli][server] serve loader `Cache-Control` headers with per-mode defaults

hassankhan:
# Why

Data loaders can declare a cache policy by returning a `Response` with a `Cache-Control` header, but that header doesn't reliably reach the browser. In dev, we apply a `no-store` header which replaces it, and loaders without a `Cache-Control` header get no policy at all when exported.

# How

### Dev server

- `/_expo/loaders/*` is excluded from the dev server middleware that applies `Cache-Control: no-store`, allowing a user-declared header from a loader to be visible during development.
- Loaders without a caching header get a `Cache-Control: no-store`

### Static output

- Loaders without a header default to `private, must-revalidate, max-age=0`, added to the `pageHeaders` rules.
- The above default also applies to the loader route's page HTML (which embeds the seed data, so it must share the loader's cache policy). #47774 already did this for declared headers; this PR makes it apply to the defaults too.

### Server output

- A new rule (`^/_expo/loaders/.+$` with `Cache-Control: no-store`) is prepended to `pageHeaders`, so both user-configured rules and loader-returned headers can override.

## `expo-server`

- The request handler now applies `pageHeaders` rules to loader responses (SSR), as we already do for loader files (SSG).

# Test Plan

- CI

# Checklist

### DIFF
--- a/packages/@expo/cli/src/export/exportStaticAsync.ts
+++ b/packages/@expo/cli/src/export/exportStaticAsync.ts
@@ -87,6 +87,18 @@ function matchGroupName(name: string): string | undefined {
   return name.match(/^\(([^/]+?)\)$/)?.[1];
 }
 
+export const SERVER_LOADER_DEFAULT_HEADER_RULE: PageHeaderInfo<string> = {
+  namedRegex: '^/_expo/loaders/.+$',
+  headers: { 'Cache-Control': 'no-store' },
+};
+
+type LoaderHeaderEntry = {
+  /** Allowlisted headers the loader response actually carried. */
+  declared: Record<string, string>;
+  /** Default values for headers the loader left undeclared. */
+  defaults: Record<string, string>;
+};
+
 export async function getFilesToExportFromServerAsync(
   projectRoot: string,
   {
@@ -242,8 +254,8 @@ export async function exportFromServerAsync(
 
   // Group variations prerender several pathnames from one loader file, so these maps can differ
   // in size.
-  const loaderHeadersByPage = new Map<string, Record<string, string>>();
-  const loaderHeadersByFile = new Map<string, Record<string, string>>();
+  const loaderHeadersByPage = new Map<string, LoaderHeaderEntry>();
+  const loaderHeadersByFile = new Map<string, LoaderHeaderEntry>();
 
   await getFilesToExportFromServerAsync(projectRoot, {
     files,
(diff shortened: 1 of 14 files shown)


packages/@expo/cli/src/export/exportStaticAsync.ts
```diff
@@ -853,11 +870,47 @@ export function getExactPathNamedRegex(pathname: string): string {
...
+  for (const name of SSG_LOADER_HEADER_ALLOWLIST) {
+    const value = responseHeaders.get(name);
+    if (value) {
+      declared[name] = value;
+    }
+  }
+
+  const defaults: Record<string, string> = {};
+  if (!declared['Cache-Control']) {
+    defaults['Cache-Control'] = 'public, max-age=0, must-revalidate';
```
hassankhan: @kitten Open to changing the default here (for SSG), could match EAS with `private, must-revalidate, max-age=0`.

packages/@expo/cli/src/export/exportStaticAsync.ts
```diff
@@ -87,6 +87,20 @@ function matchGroupName(name: string): string | undefined {
   return name.match(/^\(([^/]+?)\)$/)?.[1];
 }
 
+// Server-mode loaders run per request, so their default can't be derived per route at build
+// time; one wildcard rule covers every loader path instead.
+export const SERVER_LOADER_DEFAULT_HEADER_RULE: PageHeaderInfo<string> = {
+  namedRegex: '^/_expo/loaders/.+$',
+  headers: { 'Cache-Control': 'no-store' },
```
kitten, line 92: Should this be absorbed by `expo-server`?

packages/@expo/cli/src/export/exportStaticAsync.ts
kitten: generally, `private` would be the keyword to ensure that no stale retention can happen afaik

kitten (approved):

packages/@expo/cli/src/export/exportStaticAsync.ts
hassankhan: Definitely crossed my mind, but I thought it was better to have in the manifest because

- its (in theory) user-editable as a pre-deploy step
- uses the same `pageHeaders` machinery as SSG so there's some symmetry there
- adds a third "level" of caching logic to EAS; is a given header set by manifest, by EAS defaults or by `expo-server`?

PR #48498, opened 2026-08-04
[router] Migrate top-tabs to standard-navigation

Ubax:
# Why

`TopTabs` is the last navigator needing the migration to standard-navigation

# How

Similar to other navigators (mostly JS Tabs):
- replace `createNavigatorFactory` with `unstable_integrateWithRouter`
- extract needed properties from `navigation` and `state` and pass them via `createProps`
- fix playwright tests - now all routes needs to be declared

Additionally improve type safety

# Test Plan

CI

# Checklist

---

### DIFF
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -4,8 +4,8 @@
 
 ### 🛠 Breaking changes
 
+- Migrate the JS `TopTabs` navigator to the `standard-navigation` integration. ([#48498](https://github.com/expo/expo/pull/48498) by [@Ubax](https://github.com/Ubax))
 - Make `beforeRemove` non-preventable and add `removePrevented` event. ([#48347](https://github.com/expo/expo/pull/48347) by [@Ubax](https://github.com/Ubax))
-- Migrate the JS `TopTabs` navigator to the `standard-navigation` integration. (by [@Ubax](https://github.com/Ubax))
 - Remove the `redirect` prop from layout `<Screen>` components. ([#48369](https://github.com/expo/expo/pull/48369) by [@Ubax](https://github.com/Ubax))
 - Add `redirectTo` to protected routes and render guarded screens as redirects instead of removing them from navigators. ([#47744](https://github.com/expo/expo/pull/47744) by [@Ubax](https://github.com/Ubax))
 - Migrate the `Drawer` navigator to the `standard-navigation` integration. ([#47839](https://github.com/expo/expo/pull/47839) by [@Ubax](https://github.com/Ubax))
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### 🛠 Breaking changes
 
 - Make `beforeRemove` non-preventable and add `removePrevented` event. ([#48347](https://github.com/expo/expo/pull/48347) by [@Ubax](https://github.com/Ubax))
+- Migrate the JS `TopTabs` navigator to the `standard-navigation` integration. (by [@Ubax](https://github.com/Ubax))
 - Remove the `redirect` prop from layout `<Screen>` components. ([#48369](https://github.com/expo/expo/pull/48369) by [@Ubax](https://github.com/Ubax))
 - Add `redirectTo` to protected routes and render guarded screens as redirects instead of removing them from navigators. ([#47744](https://github.com/expo/expo/pull/47744) by [@Ubax](https://github.com/Ubax))
 - Migrate the `Drawer` navigator to the `standard-navigation` integration. ([#47839](https://github.com/expo/expo/pull/47839) by [@Ubax](https://github.com/Ubax))
--- a/packages/expo-router/package.json
+++ b/packages/expo-router/package.json
@@ -346,9 +346,11 @@
     "expect-type": "^1.3.0",
     "immer": "^10.1.1",
     "react-native-gesture-handler": "~3.1.0",
+    "react-native-pager-view": "^8.0.2",
     "react-native-reanimated": "4.5.1",
     "react-native-safe-area-context": "~5.7.0",
     "react-native-screens": "^4.26.0",
+    "react-native-tab-view": "^4.3.0",
     "react-native-web": "~0.21.0",
     "react-server-dom-webpack": "~19.0.8"
   },
--- a/packages/expo-router/src/react-navigation/material-top-tabs/__tests__/types.test.ts
+++ b/packages/expo-router/src/react-navigation/material-top-tabs/__tests__/types.test.ts
@@ -15,12 +15,14 @@ type IsAny<T> = 0 extends 1 & T ? true : false;
 type IndicatorProps = Parameters<
   NonNullable<MaterialTopTabNavigationOptions['tabBarIndicator']>
 >[0];
+type TabBarCallbackProps = Parameters<NonNullable<MaterialTopTabNavigationConfig['tabBar']>>[0];
 
 export type _PublicPropsIncludeTabBar = Expect<
   'tabBar' extends keyof JSTopTabsProps ? true : false
 >;
 export type _NavigationConfigIsNotAny = Expect<Equal<IsAny<MaterialTopTabNavigationConfig>, false>>;
 export type _TabBarPropsAreNotAny = Expect<Equal<IsAny<MaterialTopTabBarProps>, false>>;
+export type _TabBarCallbackPropsAreNotAny = Expect<Equal<IsAny<TabBarCallbackProps>, false>>;
 export type _TabBarPropsExcludeNavigation = Expect<
   'navigation' extends keyof MaterialTopTabBarProps ? false : true
 >;
--- a/packages/expo-router/src/react-navigation/material-top-tabs/__tests__/types.test.ts
+++ b/packages/expo-router/src/react-navigation/material-top-tabs/__tests__/types.test.ts
@@ -1,0 +1,38 @@
+import type { JSTopTabsProps } from '../../../layouts/TopTabs';
+import type { MaterialTopTabNavigatorContentProps } from '../navigators/createMaterialTopTabNavigator';
+import type {
+  MaterialTopTabBarProps,
+  MaterialTopTabNavigationConfig,
+  MaterialTopTabNavigationOptions,
+  MaterialTopTabViewState,
+} from '../types';
+
+type Expect<T extends true> = T;
+type Equal<A, B> =
+  (<T>() => T extends A ? 1 : 2) extends <T>() => T extends B ? 1 : 2 ? true : false;
+type IsAny<T> = 0 extends 1 & T ? true : false;
+
+type IndicatorProps = Parameters<
+  NonNullable<MaterialTopTabNavigationOptions['tabBarIndicator']>
+>[0];
+
+export type _PublicPropsIncludeTabBar = Expect<
+  'tabBar' extends keyof JSTopTabsProps ? true : false
+>;
+export type _NavigationConfigIsNotAny = Expect<Equal<IsAny<MaterialTopTabNavigationConfig>, false>>;
+export type _TabBarPropsAreNotAny = Expect<Equal<IsAny<MaterialTopTabBarProps>, false>>;
+export type _TabBarPropsExcludeNavigation = Expect<
+  'navigation' extends keyof MaterialTopTabBarProps ? false : true
+>;
+export type _IndicatorUsesTopTabViewState = Expect<
+  Equal<IndicatorProps['state'], MaterialTopTabViewState>
+>;
+export type _ContentRequiresPreloadedRouteKeys = Expect<
+  Equal<MaterialTopTabNavigatorContentProps['preloadedRouteKeys'], string[]>
+>;
+
+describe('material top tabs types', () => {
+  it('type-checks', () => {
+    expect(true).toBe(true);
+  });
+});
--- a/packages/expo-router/src/react-navigation/material-top-tabs/types.tsx
+++ b/packages/expo-router/src/react-navigation/material-top-tabs/types.tsx
@@ -7,17 +7,16 @@ import type {
   TextStyle,
   ViewStyle,
 } from 'react-native';
+import type { SceneRendererProps, TabBarIndicatorProps, TabViewProps } from 'react-native-tab-view';
+import type { NavigatorArgs } from 'standard-navigation';
 
 import type {
-  DefaultNavigatorOptions,
   Descriptor,
-  NavigationHelpers,
   NavigationProp,
   ParamListBase,
   RouteProp,
   TabActionHelpers,
   TabNavigationState,
-  TabRouterOptions,
   Theme,
 } from '../native';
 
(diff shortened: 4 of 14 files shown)


packages/expo-router/package.json
```diff
@@ -344,9 +344,11 @@
     "expect-type": "^1.3.0",
     "immer": "^10.1.1",
     "react-native-gesture-handler": "~3.1.0",
+    "react-native-pager-view": "^8.0.2",
     "react-native-reanimated": "4.5.1",
     "react-native-safe-area-context": "~5.7.0",
     "react-native-screens": "^4.26.0",
+    "react-native-tab-view": "^4.3.0",
```
jakub-agent, line 353: [🚨 **critical**] `pnpm-lock.yaml` isn't updated for these two new devDependencies, so every CI job fails before running anything:

```
ERR_PNPM_OUTDATED_LOCKFILE  Cannot install with "frozen-lockfile" because pnpm-lock.yaml is not up to date with <ROOT>/packages/expo-router/package.json
  Failure reason:
  specifiers in the lockfile don't match specifiers in package.json:
  * 2 dependencies were added: react-native-pager-view@^8.0.2, react-native-tab-view@^4.3.0
```

That's from `jest-ubuntu (1)`; `build`, `check-packages`, `android-build`, `ios-build`, and the playwright jobs all fail the same way. Run `pnpm install` at the repo root and commit the lockfile.

packages/expo-router/src/react-navigation/material-top-tabs/types.tsx
```diff
@@ -291,22 +302,12 @@ export type MaterialTopTabNavigationConfig = Omit<
 };
 
 export type MaterialTopTabBarProps = any & {
-  state: TabNavigationState<ParamListBase>;
-  navigation: NavigationHelpers<ParamListBase, MaterialTopTabNavigationEventMap>;
+  state: MaterialTopTabViewState;
   descriptors: MaterialTopTabDescriptorMap;
+  emitter: MaterialTopTabEmitter;
```
jakub-agent, line 308: [🏗️ **design**] `MaterialTopTabBarProps` is declared as `any & { ... }`, and TypeScript collapses an intersection with `any` down to `any`. The members listed here document the contract but the compiler never enforces it.

That matters specifically because this PR removes `navigation` from the tab-bar props. A custom `tabBar` written for the old API still type-checks cleanly:

```tsx
// compiles today and after this PR; throws at runtime after it
// TypeError: Cannot read properties of undefined (reading 'emit')
<TopTabs
  tabBar={(props) => (
    <MyBar onPress={(key) => props.navigation.emit({ type: 'tabPress', target: key })} />
  )}
/>
```

I verified the collapse with `tsc --strict` on a reduced repro, so this isn't theoretical.

### Suggested fix

The extra props the tab bar actually receives are fully known. `MaterialTopTabView.renderTabBar` strips `navigationState` and `options` from what `TabView` hands it and forwards the remainder, which is exactly `react-native-tab-view`'s `SceneRendererProps` (`{ layout, position, jumpTo }`):

```ts
import type { SceneRendererProps } from 'react-native-tab-view';

export type MaterialTopTabBarProps = SceneRendererProps & {
  state: MaterialTopTabViewState;
  descriptors: MaterialTopTabDescriptorMap;
  emitter: MaterialTopTabEmitter;
  navigateToTab: (routeKey: string) => void;
};
```

That is a type-only import, so it adds no runtime dependency, and it introduces no new resolution constraint — `MaterialTopTabBar.tsx` and `MaterialTopTabView.tsx` already reference `typeof import('react-native-tab-view').TabBar` / `.TabView` in this PR.

With that in place, a custom `tabBar` reaching for `props.navigation` fails to compile and the error points at `emitter` / `navigateToTab`.

### Related, optional

`MaterialTopTabNavigationConfig = Omit<any, ...>` has the same problem one level up — it is what makes `JSTopTabsProps` accept arbitrary props. The equivalent fix:

```ts
import type { Route, TabViewProps } from 'react-native-tab-view';

export type MaterialTopTabNavigationConfig = Omit<
  TabViewProps<Route>,
  | 'navigationState' | 'onIndexChange' | 'onSwipeStart' | 'onSwipeEnd'
  | 'renderScene' | 'renderTabBar' | 'renderLazyPlaceholder'
  | 'swipeEnabled' | 'animationEnabled'
  | 'lazy' | 'lazyPreloadDistance'
  // `MaterialTopTabView` also computes these from descriptors and `useLocale`,
  // so a caller-supplied value is silently dropped by the `{...rest}` spread order.
  | 'options' | 'direction'
> & {
  tabBar?: (props: MaterialTopTabBarProps) => React.ReactNode;
};
```

Note `lazyPlaceholder` in the current omit list isn't a `TabView` prop (it's a per-screen option), so it drops out. Expect this one to surface a few real errors, so it's reasonable as a follow-up if you'd rather keep this PR to the navigator migration — but the `MaterialTopTabBarProps` fix above is the one guarding the breaking change.

packages/expo-router/src/react-navigation/material-top-tabs/types.tsx
```diff
@@ -29,22 +27,35 @@ export type MaterialTopTabNavigationEventMap = {
...
+export type MaterialTopTabViewRoute = {
+  key: string;
+  name: string;
+  params?: object;
+};
+
+/**
+ * The navigator state consumed by `MaterialTopTabView` and the tab bar.
+ */
+export type MaterialTopTabViewState = {
```
jakub-agent, line 51: [🏗️ **design**] The state handed to the tab bar is now this reduced shape rather than `TabNavigationState<ParamListBase>`, and `MaterialTopTabBar` forwards it straight to `TabBar` as `navigationState`. `renderIndicator` then casts it back to the full state before handing it to the user's option:

```tsx
focusedOptions.tabBarIndicator({ state: state as TabNavigationState<ParamListBase>, ...rest })
```

That cast used to be accurate and no longer is. `MaterialTopTabNavigationOptions.tabBarIndicator` still declares `state: TabNavigationState<ParamListBase>`, so an indicator reading `state.key`, `state.routeNames`, `state.type`, `state.stale`, or `state.preloadedRouteKeys` gets `undefined` at runtime with no compile error.

### Suggested fix

The option's props are currently derived from `React.ComponentProps<any>['renderIndicator']`, which is `any`, so nothing in that declaration is checked either. `react-native-tab-view` exports the real type — `TabBarIndicatorProps<T>` is `SceneRendererProps & { navigationState, width, getTabWidth, direction, style?, gap?, children? }`, which is exactly what `renderIndicator` receives:

```ts
import type { TabBarIndicatorProps } from 'react-native-tab-view';

  tabBarIndicator?: (
    props: Omit<TabBarIndicatorProps<MaterialTopTabViewRoute>, 'navigationState'> & {
      state: MaterialTopTabViewState;
    }
  ) => React.ReactNode;
```

Then the cast in `MaterialTopTabBar.tsx` becomes unnecessary:

```tsx
renderIndicator={({ navigationState: state, ...rest }) =>
  focusedOptions.tabBarIndicator ? (
    focusedOptions.tabBarIndicator({ state, ...rest })
  ) : (
    <TabBarIndicator navigationState={state} {...rest} />
  )
}
```

and the `ParamListBase` / `TabNavigationState` imports in that file drop out — `renderIndicator` is their only use.

This is a breaking change for anyone whose indicator reads the fields that are now gone, but it surfaces at compile time instead of as `undefined` at runtime. If you'd rather keep those fields available, the alternative is carrying them through the integration rather than narrowing the option.

packages/expo-router/src/react-navigation/material-top-tabs/__tests__/types.test.ts
```diff
@@ -0,0 +1,21 @@
...
+import type { JSTopTabsProps, TopTabs } from '../../../layouts/TopTabs';
+import type { MaterialTopTabNavigatorContentProps } from '../navigators/createMaterialTopTabNavigator';
+
+type Expect<T extends true> = T;
+type Equal<A, B> =
+  (<T>() => T extends A ? 1 : 2) extends <T>() => T extends B ? 1 : 2 ? true : false;
+
+type TopTabsProps = ComponentProps<typeof TopTabs>;
+
+export type _PublicPropsMatchTopTabs = Expect<Equal<JSTopTabsProps, TopTabsProps>>;
```
jakub-agent: [💡 suggestion] This assertion can't fail. `JSTopTabsProps` is defined in `TopTabsClient` as `ComponentProps<typeof TopTabs>`, and `TopTabsProps` above is the same expression on the same component, so it reduces to `Equal<X, X>`. It would only start doing work if `JSTopTabsProps` were later hand-written.

`_ContentRequiresPreloadedRouteKeys` below it does check something real. If you want a meaningful public-props check, asserting on specific keys (`tabBar`, `screenOptions`, `backBehavior`) would survive a refactor — though that only becomes possible once `MaterialTopTabNavigationConfig` stops being `Omit<any, ...>`.

packages/expo-router/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### 🛠 Breaking changes
 
+- Migrate the JS `TopTabs` navigator to the `standard-navigation` integration. (by [@Ubax](https://github.com/Ubax))
```
jakub-agent: [💡 suggestion] Missing the PR link that every neighbouring entry carries: `([#48498](https://github.com/expo/expo/pull/48498) by [@Ubax](https://github.com/Ubax))`.

Since the compiler won't flag it (see the `MaterialTopTabBarProps` comment), it's also worth naming the user-visible break here: custom `tabBar` components no longer receive `navigation` and must use `emitter.emit(...)` and `navigateToTab(routeKey)` instead.

packages/expo-router/src/react-navigation/material-top-tabs/views/MaterialTopTabBar.tsx
```diff
@@ -36,8 +36,9 @@ const renderLabelDefault = ({ color, labelText, style, allowFontScaling }: any)
 
 export function MaterialTopTabBar({
   state,
-  navigation,
   descriptors,
+  emitter,
+  navigateToTab: _navigateToTab,
```
jakub-agent, line 42: [🐧 nit] This is correct but takes a while to confirm. The prop is pulled out only to keep it out of the `...rest` spread into `TabBar`; the built-in bar actually switches tabs through react-native-tab-view's `jumpTo`, which reaches it via `...rest` from `MaterialTopTabView.renderTabBar` and lands back on `navigateToTab` through `onIndexChange`.

A one-line comment saying that would save the next reader the trace — as written, the `_` prefix reads like an unfinished part of the migration.

jakub-agent (changes requested):

packages/expo-router/src/react-navigation/material-top-tabs/views/MaterialTopTabView.tsx
```diff
@@ -55,8 +53,9 @@ export function MaterialTopTabView({
     return tabBar({
       ...rest,
       state,
-      navigation,
       descriptors,
+      emitter,
+      navigateToTab,
```
jakub-agent, line 60: [💡 suggestion] The public types are now honest, but the code that produces them is still untyped, so nothing checks that the two agree.

`renderTabBar` above is declared as `React.ComponentProps<any>['renderTabBar']` with a `: any` parameter, so `rest` here is `any` and this `tabBar({ ... })` call type-checks no matter what it passes. If `SceneRendererProps` ever stops matching what `TabView` forwards, `MaterialTopTabBarProps` silently becomes wrong again.

The same gap makes `_IndicatorUsesTopTabViewState` weaker than it looks: it proves the *declaration* of `tabBarIndicator`, but the call site in `MaterialTopTabBar.tsx` is `({ navigationState: state, ...rest }: any)`, so the value handed to it is unchecked.

Typing the callback closes both:

```tsx
import type { TabViewProps } from 'react-native-tab-view';

const renderTabBar: NonNullable<TabViewProps<Route<string>>['renderTabBar']> = ({
  /* eslint-disable @typescript-eslint/no-unused-vars */
  navigationState,
  options,
  /* eslint-enable @typescript-eslint/no-unused-vars */
  ...sceneRendererProps
}) => tabBar({ ...sceneRendererProps, state, descriptors, emitter, navigateToTab });
```

Destructuring `navigationState` and `options` leaves exactly `SceneRendererProps`, so TypeScript verifies the object against `MaterialTopTabBarProps` for you.

The remaining `: any` annotations can go the same way now that the surrounding types are real — `renderScene`, `lazy`, and `renderLazyPlaceholder` here all infer from `TabView<Route<string>>`, and `renderIndicator`, `onTabPress`, `onTabLongPress`, and `state.routes.map((route: any) => ...)` in `MaterialTopTabBar.tsx` infer from `MaterialTopTabBarProps`. Fine as a follow-up if you'd rather not grow this PR.

packages/expo-router/src/react-navigation/material-top-tabs/views/MaterialTopTabBar.tsx
```diff
@@ -111,9 +108,9 @@ export function MaterialTopTabBar({
       direction={direction}
       scrollEnabled={focusedOptions.tabBarScrollEnabled}
       bounces={focusedOptions.tabBarBounces}
-      activeColor={activeColor}
-      inactiveColor={inactiveColor}
-      pressColor={focusedOptions.tabBarPressColor}
+      activeColor={activeColor as string}
```
jakub-agent, line 117: [🐧 nit] These three casts (`activeColor`, `inactiveColor`, `pressColor`) are new, and they paper over a real mismatch: the option types are `ColorValue` (`string | OpaqueColorValue`) while `TabBar` declares `activeColor?: string`.

I checked and they're safe — `TabBarItem` only forwards these into a style `color` and into `android_ripple`, both of which accept an `OpaqueColorValue` at runtime, so `PlatformColor(...)` still works. Worth a one-line comment saying that, otherwise the next person to touch this either removes the cast and breaks the build or narrows the option to `string` and drops working functionality.

(Unrelated but nearby: the `inactiveColor` fallback runs `Color(activeColor)?.alpha(0.5)`, which can't parse an `OpaqueColorValue` and quietly falls back to `colors.text`. Pre-existing, just noting it while the colour handling is in view.)

packages/expo-router/CHANGELOG.md
jakub-agent: [💡 suggestion] Still missing the PR link that all six neighbouring entries carry:

```md
- Migrate the JS `TopTabs` navigator to the `standard-navigation` integration. ([#48498](https://github.com/expo/expo/pull/48498) by [@Ubax](https://github.com/Ubax))
```

Worth adding the migration hint too, now that the compiler does flag it: custom `tabBar` components no longer receive `navigation` and should use `emitter.emit(...)` and `navigateToTab(routeKey)`.

jakub-agent (approved):

packages/expo-router/src/react-navigation/material-top-tabs/views/MaterialTopTabBar.tsx
Ubax: Fix the issue in props of TabBarItem, so that it is `ColorValue` all the way through

PR #48499, opened 2026-08-04
[router] Unify JS Tabs, TopTabs, Drawer, and headless tabs with NativeTabs - only screens declared in the layout become visible

Ubax:
# Why

Unify JS Tabs, TopTabs, Drawer, and headless tabs with NativeTabs, by showing only screens declared in the `_layout`

# How

Use `useVisibleTabsWithRedirect` in every tab navigator

# Test Plan

CI

# Checklist

---

### DIFF
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### 🛠 Breaking changes
 
+- Unify JS Tabs, TopTabs, Drawer, and headless tabs with NativeTabs - only screens declared in the layout become visible. ([#48499](https://github.com/expo/expo/pull/48499) by [@Ubax](https://github.com/Ubax))
 - Migrate the JS `TopTabs` navigator to the `standard-navigation` integration. ([#48498](https://github.com/expo/expo/pull/48498) by [@Ubax](https://github.com/Ubax))
 - Make `beforeRemove` non-preventable and add `removePrevented` event. ([#48347](https://github.com/expo/expo/pull/48347) by [@Ubax](https://github.com/Ubax))
 - Remove the `redirect` prop from layout `<Screen>` components. ([#48369](https://github.com/expo/expo/pull/48369) by [@Ubax](https://github.com/Ubax))
--- a/packages/expo-router/src/layouts/TabsClient.tsx
+++ b/packages/expo-router/src/layouts/TabsClient.tsx
@@ -26,7 +26,10 @@ import type { Href } from '../types';
 // Keep React Navigation client-only so the entry evaluates in React Server Components.
 export * from '../react-navigation/bottom-tabs';
 
-export type TabsScreenOptions = BottomTabNavigationOptions & { href?: Href | null };
+export type TabsScreenOptions = BottomTabNavigationOptions & {
+  // TODO: Consider deprecating `href`.
+  href?: Href | null;
+};
 
 /**
  * Renders a tabs navigator.
@@ -58,15 +61,21 @@ const Tabs = unstable_integrateWithRouter<
         if (options.tabBarButton) {
           throw new Error('Cannot use `href` and `tabBarButton` together.');
         }
+        if (href === null) {
+          // TODO(@ubax): Update the hiding-a-tab guide for the new redirect behavior.
+          return {
+            ...screen,
+            options: {
+              ...options,
+              hidden: true,
+            },
+          };
+        }
         return {
           ...screen,
           options: {
             ...options,
-            tabBarItemStyle: href == null ? { display: 'none' } : options.tabBarItemStyle,
             tabBarButton: (props) => {
-              if (href == null) {
-                return null;
-              }
               const children =
                 Platform.OS === 'web' ? props.children : <Pressable>{props.children}</Pressable>;
               // TODO: React Navigation types these props as Animated.WithAnimatedValue<StyleProp<ViewStyle>>
--- a/packages/expo-router/src/react-navigation/bottom-tabs/__tests__/index.test.ios.tsx
+++ b/packages/expo-router/src/react-navigation/bottom-tabs/__tests__/index.test.ios.tsx
@@ -1,5 +1,5 @@
 import { userEvent } from '@testing-library/react-native';
-import { useEffect } from 'react';
+import { useEffect, useState } from 'react';
 import {
   type EmitterSubscription,
   Keyboard,
@@ -17,7 +17,14 @@ import { Text } from '../../elements';
 import type { ParamListBase } from '../../native';
 import type { BottomTabBarProps, BottomTabNavigationProp } from '../types';
 
+let warnSpy: jest.SpyInstance;
+
+beforeEach(() => {
+  warnSpy = jest.spyOn(console, 'warn').mockImplementation(() => {});
+});
+
 afterEach(() => {
+  warnSpy.mockRestore();
   jest.restoreAllMocks();
 });
 
@@ -215,7 +222,7 @@ test('keeps the dynamic segment of a tab when it is re-selected', async () => {
   expect(screen).toHavePathname('/abc');
 });
 
-test('hides a tab whose href is null', () => {
+test('removes and redirects from a tab whose href is null', () => {
   renderRouter({
     _layout: () => (
       <Tabs>
@@ -228,7 +235,140 @@ test('hides a tab whose href is null', () => {
   });
 
   expect(screen.queryByRole('button', { name: /second/ })).toBeNull();
-  expect(screen.getByRole('button', { name: 'index, tab, 1 of 2' })).toBeVisible();
+  expect(screen.getByRole('button', { name: 'index, tab, 1 of 1' })).toBeVisible();
+
+  act(() => router.push('/second'));
+
+  expect(screen).toHavePathname('/');
+  expect(screen.queryByTestId('second')).toBeNull();
+});
+
+test('redirects when the focused tab href becomes null', () => {
+  let setHidden!: (hidden: boolean) => void;
+  function Layout() {
+    const [hidden, set] = useState(false);
+    setHidden = set;
+    return (
+      <Tabs>
+        <Tabs.Screen name="index" />
+        <Tabs.Screen name="second" options={{ href: hidden ? null : '/second' }} />
+      </Tabs>
+    );
+  }
+
+  renderRouter({
+    _layout: Layout,
+    index: () => <View testID="index" />,
+    second: () => <View testID="second" />,
+  });
+
+  act(() => router.push('/second'));
+  expect(screen).toHavePathname('/second');
+
+  act(() => setHidden(true));
+
+  expect(screen).toHavePathname('/');
+  expect(screen.getByTestId('index')).toBeVisible();
+  expect(screen.queryByTestId('second')).toBeNull();
+});
+
+test('renders only screens declared in the layout and redirects from undeclared routes', () => {
+  renderRouter({
+    _layout: () => (
+      <Tabs>
+        <Tabs.Screen name="index" />
+      </Tabs>
+    ),
+    index: () => <View testID="index" />,
+    second: () => <View testID="second" />,
+  });
+
+  expect(screen.getAllByRole('button', { name: /tab/ })).toHaveLength(1);
+
+  act(() => router.push('/second'));
+
+  expect(screen).toHavePathname('/');
+  expect(screen.getByTestId('index')).toBeVisible();
+});
+
+test('redirects an undeclared route to the configured initial tab', () => {
+  renderRouter({
+    _layout: {
+      unstable_settings: { initialRouteName: 'second' },
+      default: () => (
+        <Tabs>
+          <Tabs.Screen name="index" />
+          <Tabs.Screen name="second" />
+        </Tabs>
+      ),
+    },
+    index: () => <View testID="index" />,
+    second: () => <View testID="second" />,
+    undeclared: () => <View testID="undeclared" />,
+  });
+
+  act(() => router.push('/undeclared'));
+
+  expect(screen).toHavePathname('/second');
+  expect(screen.getByTestId('second')).toBeVisible();
+});
+
+test('renders no tab UI when no screens are declared in the layout', () => {
+  renderRouter({
+    _layout: () => <Tabs />,
+    index: () => <View testID="index" />,
+  });
+
+  expect(screen.queryByTestId('index')).toBeNull();
+  expect(screen.queryAllByRole('button', { name: /tab/ })).toHaveLength(0);
+  expect(warnSpy.mock.calls).toMatchSnapshot();
+});
+
+test('prefers a protected screen redirect over the tab visibility fallback', () => {
+  renderRouter(
+    {
+      _layout: () => <Stack />,
+      'tabs/_layout': () => (
+        <Tabs>
+          <Tabs.Screen name="index" />
+          <Tabs.Protected guard={false} redirectTo="/login">
+            <Tabs.Screen name="secret" />
+          </Tabs.Protected>
+        </Tabs>
+      ),
+      'tabs/index': () => <View testID="index" />,
+      'tabs/secret': () => <View testID="secret" />,
+      login: () => <View testID="login" />,
+    },
+    { initialUrl: '/tabs/secret' }
+  );
+
+  expect(screen).toHavePathname('/login');
+  expect(screen.getByTestId('login')).toBeVisible();
+});
+
+test('redirects a fully guarded layout without warning that no screens are declared', () => {
+  renderRouter(
+    {
+      _layout: () => <Stack />,
+      'tabs/_layout': () => (
+        <Tabs>
+          <Tabs.Protected guard={false} redirectTo="/login">
+            <Tabs.Screen name="index" />
+            <Tabs.Screen name="second" />
+          </Tabs.Protected>
+        </Tabs>
+      ),
+      'tabs/index': () => <View testID="index" />,
+      'tabs/second': () => <View testID="second" />,
+      login: () => <View testID="login" />,
+    },
+    { initialUrl: '/tabs/second' }
+  );
+
+  expect(screen).toHavePathname('/login');
+  expect(screen.getByTestId('login')).toBeVisible();
+  expect(warnSpy).not.toHaveBeenCalled();
 });
 
 test('throws when a screen uses both href and tabBarButton', () => {
--- a/packages/expo-router/src/react-navigation/bottom-tabs/navigators/createBottomTabNavigator.tsx
+++ b/packages/expo-router/src/react-navigation/bottom-tabs/navigators/createBottomTabNavigator.tsx
@@ -3,6 +3,7 @@
 import { createStandardNavigator } from 'standard-navigation';
 
 import type { NavigatorContentProps } from '../../../standard-navigation/types';
+import { useVisibleTabsWithRedirect } from '../../../standard-navigation/useVisibleTabsWithRedirect';
 import type {
   BottomTabDescriptorMap,
   BottomTabNavigationConfig,
@@ -35,6 +36,11 @@ function BottomTabNavigatorContent({
   popNestedStackToTop,
   ...rest
 }: ContentArgs) {
+  const { visibleRoutes, focusedIndex } = useVisibleTabsWithRedirect({
+    routes: state.routes,
+    focusedRouteKey: state.routes[state.index]!.key,
+    descriptors,
+  });
   const navigateToTab = (routeKey: string) => {
     const route = state.routes.find((route) => route.key === routeKey);
     if (route) {
@@ -47,10 +53,19 @@ function BottomTabNavigatorContent({
     }
   };
 
+  if (visibleRoutes.length === 0) {
+    return null;
+  }
+
   return (
     <BottomTabView
       {...rest}
-      state={state}
+      state={{
+        // Only routes are substituted; navigator metadata remains from the real state.
+        ...state,
+        routes: visibleRoutes,
+        index: focusedIndex,
+      }}
       // TODO(@ubax): SDK-58: Try to remove the casting from here to ensure type safety
       // Integration supplies full descriptors, including preload placeholders; standard types omit route/navigation.
       descriptors={descriptors as unknown as BottomTabDescriptorMap}
--- a/packages/expo-router/src/standard-navigation/__tests__/useVisibleTabsWithRedirect.test.ios.tsx
+++ b/packages/expo-router/src/standard-navigation/__tests__/useVisibleTabsWithRedirect.test.ios.tsx
@@ -1,5 +1,6 @@
 import { renderHook } from '@testing-library/react-native';
 
+import { useRouteNode } from '../../Route';
 import { router } from '../../imperative-api';
 import { useIsFocused } from '../../react-navigation/native';
 import { useBuildHref } from '../useBuildHref';
@@ -12,9 +13,14 @@ jest.mock('../../react-navigation/native', () => {
   return { ...actualNavigation, useIsFocused: jest.fn() };
 });
 jest.mock('../useBuildHref');
+jest.mock('../../Route', () => ({
+  ...jest.requireActual('../../Route'),
+  useRouteNode: jest.fn(),
+}));
 
 const mockedUseIsFocused = useIsFocused as jest.MockedFunction<typeof useIsFocused>;
 const mockedUseBuildHref = useBuildHref as jest.MockedFunction<typeof useBuildHref>;
+const mockedUseRouteNode = useRouteNode as jest.MockedFunction<typeof useRouteNode>;
 
 const routes = [
   { key: 'home-key', name: 'home' },
@@ -28,20 +34,23 @@ const descriptors = {
   'hidden-key': { routeSource: 'layout' as const, options: { hidden: true } },
   'filesystem-key': { routeSource: 'filesystem' as const },
 };
-const isHidden = (options: { hidden?: boolean } | undefined) => options?.hidden === true;
-
 let replaceSpy: jest.SpyInstance;
+let warnSpy: jest.SpyInstance;
 let buildHref: jest.Mock;
 
 beforeEach(() => {
   buildHref = jest.fn((route) => `/href/${route.name}`);
   mockedUseBuildHref.mockReturnValue(buildHref);
   mockedUseIsFocused.mockReturnValue(true);
+  // Prevent route data from leaking between tests.
+  mockedUseRouteNode.mockReturnValue(null);
   replaceSpy = jest.spyOn(router, 'replace').mockImplementation(() => {});
+  warnSpy = jest.spyOn(console, 'warn').mockImplementation(() => {});
 });
 
 afterEach(() => {
   replaceSpy.mockRestore();
+  warnSpy.mockRestore();
 });
 
 describe('useVisibleTabsWithRedirect', () => {
@@ -51,32 +60,32 @@ describe('useVisibleTabsWithRedirect', () => {
         routes,
         focusedRouteKey: 'settings-key',
         descriptors,
-        isHidden,
       })
     );
 
     expect(result.current).toEqual({
       visibleRoutes: [routes[0], routes[1]],
-      visibleFocusedIndex: 1,
+      focusedIndex: 1,
     });
   });
 
-  it('returns -1 when the focused route is not visible', () => {
+  it('returns zero when the focused route is not visible', () => {
     const { result } = renderHook(() =>
-      useVisibleTabsWithRedirect({ routes, focusedRouteKey: 'hidden-key', descriptors, isHidden })
+      useVisibleTabsWithRedirect({ routes, focusedRouteKey: 'hidden-key', descriptors })
     );
 
-    expect(result.current.visibleFocusedIndex).toBe(-1);
+    expect(result.current.focusedIndex).toBe(0);
   });
 
   it('redirects an unavailable focused route to the configured visible route', () => {
+    mockedUseRouteNode.mockReturnValue({ initialRouteName: 'settings' } as ReturnType<
+      typeof useRouteNode
+    >);
     renderHook(() =>
       useVisibleTabsWithRedirect({
         routes,
         focusedRouteKey: 'hidden-key',
         descriptors,
-        redirectToRouteName: 'settings',
-        isHidden,
       })
     );
 
@@ -85,13 +94,14 @@ describe('useVisibleTabsWithRedirect', () => {
   });
 
   it('builds the redirect href from the selected route', () => {
+    mockedUseRouteNode.mockReturnValue({ initialRouteName: 'settings' } as ReturnType<
+      typeof useRouteNode
+    >);
     renderHook(() =>
       useVisibleTabsWithRedirect({
         routes,
         focusedRouteKey: 'hidden-key',
         descriptors,
-        redirectToRouteName: 'settings',
-        isHidden,
       })
     );
 
@@ -99,13 +109,14 @@ describe('useVisibleTabsWithRedirect', () => {
   });
 
   it('falls back to the first visible route when the configured route is unavailable', () => {
+    mockedUseRouteNode.mockReturnValue({ initialRouteName: 'missing' } as ReturnType<
+      typeof useRouteNode
+    >);
     renderHook(() =>
       useVisibleTabsWithRedirect({
         routes,
         focusedRouteKey: 'hidden-key',
         descriptors,
-        redirectToRouteName: 'missing',
-        isHidden,
       })
     );
 
@@ -117,7 +128,7 @@ describe('useVisibleTabsWithRedirect', () => {
     mockedUseIsFocused.mockReturnValue(false);
 
     renderHook(() =>
-      useVisibleTabsWithRedirect({ routes, focusedRouteKey: 'hidden-key', descriptors, isHidden })
+      useVisibleTabsWithRedirect({ routes, focusedRouteKey: 'hidden-key', descriptors })
     );
 
     expect(replaceSpy).not.toHaveBeenCalled();
@@ -125,7 +136,7 @@ describe('useVisibleTabsWithRedirect', () => {
 
   it('does not redirect when the focused route is visible', () => {
     renderHook(() =>
-      useVisibleTabsWithRedirect({ routes, focusedRouteKey: 'home-key', descriptors, isHidden })
+      useVisibleTabsWithRedirect({ routes, focusedRouteKey: 'home-key', descriptors })
     );
 
     expect(buildHref).toHaveBeenCalledWith(routes[0]);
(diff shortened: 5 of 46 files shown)


packages/expo-router/src/react-navigation/bottom-tabs/navigators/createBottomTabNavigator.tsx
```diff
@@ -47,10 +62,18 @@ function BottomTabNavigatorContent({
     }
   };
 
+  if (visibleRoutes.length === 0) {
```
jakub-agent, line 56: [🏗️ **design**] This is the migration cliff for existing apps: `<Tabs />` with no `<Tabs.Screen>` children used to generate a tab per filesystem route, and now renders nothing — no error, no warning, just a blank screen. `redirectHref` is also `null` in this branch, so the URL doesn't move either, which makes it hard to tell apart from a crash. The three new tests (`renders no tab UI when no screens are declared in the layout` and friends) currently lock the silence in as expected behaviour.

A `console.warn` would make it diagnosable. Putting it in `useVisibleTabsWithRedirect` covers all three navigators at once and can name the layout file:

```ts
const contextKey = useRouteNode()?.contextKey;

useEffect(() => {
  if (process.env.NODE_ENV === 'development' && visibleRoutes.length === 0) {
    console.warn(
      `No screens are declared in ${contextKey ?? 'the layout'}, so the navigator renders nothing. ` +
        `Add a <Screen name="..." /> for each screen you want to render.`
    );
  }
}, [contextKey, visibleRoutes.length]);
```

(`process.env.NODE_ENV === 'development'` matches the existing warn guard in `ui/Tabs.tsx`.) The three `renders no ... UI` tests would then also assert the warning, which documents the behaviour rather than just the blank render.

packages/expo-router/src/standard-navigation/types.ts
```diff
@@ -24,6 +24,10 @@ export type StandardNavigatorEventMapBase = Record<
 
 export type StandardNavigationAction = NavigateAction | GoBackAction;
 
+export type WithRedirectToRouteName<T extends object> = T & {
```
jakub-agent: [🏗️ **design**] `redirectToRouteName` doesn't need to be a prop. All four call sites pass exactly the same value:

- `TabsClient.tsx`, `TopTabsClient.tsx`, `DrawerClient.tsx` — `useRouteNode()?.initialRouteName`
- `ui/Tabs.tsx` — `initialRouteName`, which is `routeNode.initialRouteName` from the same `useRouteNode()`

The hook can read it itself. `NavigatorContent` renders inside `withLayoutContext`'s `GuardContextProvider`, under the layout's `Route` context — which is already load-bearing, since `useGuardRedirect` reads `GuardContext` from exactly there. So `useRouteNode()` inside the hook resolves to the layout node, not a screen node.

Suggested change:

```diff
 // useVisibleTabsWithRedirect.ts
-  redirectToRouteName,
   isHidden,
 }: {
   routes: Route[];
   focusedRouteKey: string;
   descriptors: Record<string, TabDescriptor<Options>>;
-  redirectToRouteName?: string;
   isHidden?: (options: Options | undefined) => boolean;
 }) {
+  const redirectToRouteName = useRouteNode()?.initialRouteName;
```

That then lets you delete:

- this `WithRedirectToRouteName` type;
- the `WithRedirectToRouteName<...>` wrappers and `redirectToRouteName` destructures in all three `createXNavigator.tsx` files, plus the matching config type arguments in the three `XClient.tsx` files;
- the three `Object.assign` wrapper components and their `Omit<ComponentProps<...>, 'redirectToRouteName'>` prop types;
- `redirectToRouteName: initialRouteName` in `ui/Tabs.tsx`.

Roughly all of the plumbing in this PR outside the hook itself. It also keeps `Tabs` / `TopTabs` / `Drawer` as the `forwardRef` components `withLayoutContext` returns instead of plain function wrappers, and removes the failure mode where rendering `TabsNavigator` directly silently gets `undefined` and falls back to the first visible tab.

packages/expo-router/src/layouts/TabsClient.tsx
```diff
@@ -58,15 +69,20 @@ const Tabs = unstable_integrateWithRouter<
         if (options.tabBarButton) {
           throw new Error('Cannot use `href` and `tabBarButton` together.');
         }
+        if (href === null) {
```
jakub-agent, line 64: [💡 suggestion] Worth a `TODO` here (or in the changelog entry) to update the docs — not blocking.

Mapping `href: null` onto `hidden: true` changes what it means. Before, the tab button was hidden but the route stayed reachable: `router.push('/second')` or a `<Link>` from another tab rendered the screen. `docs/pages/router/advanced/tabs.mdx#hiding-a-tab` describes exactly that — *"Sometimes you want a route to exist but not show up in the tab bar"* — and the new test asserts `router.push('/second')` now lands on `/`.

So the docs section, and the changelog line (which only mentions undeclared screens), both need a follow-up once the replacement lands — the `TODO(@ubax)` in `useVisibleTabsWithRedirect` about a formsheet for focused hidden tabs looks like the intended successor.

packages/expo-router/src/react-navigation/bottom-tabs/__tests__/index.test.ios.tsx
```diff
@@ -228,7 +228,93 @@ test('hides a tab whose href is null', () => {
...
+  expect(screen).toHavePathname('/second');
+
+  act(() => setHidden(true));
+
+  expect(screen).toHavePathname('/');
+  expect(screen.getByTestId('index')).toBeVisible();
+  expect(screen.queryByTestId('second')).toBeNull();
+});
+
+test('renders only screens declared in the layout and redirects from undeclared routes', () => {
```
jakub-agent, line 275: [💡 suggestion] The main new wiring in this PR — `redirectToRouteName` — isn't tested. Every new test here has `index` as both the anchor and the first visible route, so `findRouteByName(visibleRoutes, redirectToRouteName) ?? visibleRoutes[0]` returns the same route either way; these would all pass with the wiring removed.

A case with `unstable_settings: { initialRouteName: 'second' }` where `second` is *not* first in `visibleRoutes`, then pushing an undeclared route, would pin it down. Worth covering the drawer and top-tabs wrappers too, since each passes `redirectToRouteName` separately.

Also untested: the `/index` suffix branch of `findRouteByName` (`initialRouteName: 'home'` against a route named `home/index`).

packages/expo-router/src/layouts/TabsClient.tsx
```diff
@@ -21,23 +22,33 @@ import {
...
 } from '../react-navigation/native';
 import { unstable_integrateWithRouter } from '../standard-navigation';
+import type { WithRedirectToRouteName } from '../standard-navigation/types';
 import type { Href } from '../types';
 
 // Keep React Navigation client-only so the entry evaluates in React Server Components.
 export * from '../react-navigation/bottom-tabs';
 
-export type TabsScreenOptions = BottomTabNavigationOptions & { href?: Href | null };
+export type TabsScreenOptions = BottomTabNavigationOptions & {
```
jakub-agent, line 29: [💡 suggestion] `hidden` here is redundant now that `BottomTabNavigationOptions` declares it (`react-navigation/bottom-tabs/types.tsx`), which `TabsScreenOptions` already intersects. The intersection just re-declares the same optional boolean and duplicates the doc comment.

The same doc comment now exists in four places (here, `bottom-tabs/types.tsx`, `drawer/types.tsx`, `material-top-tabs/types.tsx`). Three are unavoidable since the option types are separate, but this one can go.

packages/expo-router/src/standard-navigation/useVisibleTabsWithRedirect.ts
```diff
@@ -36,6 +37,8 @@ export function useVisibleTabsWithRedirect<Route extends TabRoute, Options exten
 }) {
   const buildHref = useBuildHref();
   const isFocused = useIsFocused();
+  const focusedRoute = routes.find((route) => route.key === focusedRouteKey);
+  const guardRedirect = useGuardRedirect(focusedRoute?.name ?? '');
```
jakub-agent, line 41: [🐧 nit] Two small cleanups now that this file imports from `GuardContext`:

- `findRouteByName` below reimplements `routeName.replace(/\/index$/, '')`, which is exactly the `normalizeRouteName` helper already exported from `GuardContext`. Reusing it keeps the two "resolve a route name" paths from drifting.
- The three `isTabHidden` / `isDrawerItemHidden` / `isTopTabHidden` helpers in the navigator files are identical (`options?.hidden === true`). Since `isHidden` only needs to be reference-stable, a single exported default here would do, and the `isHidden` parameter could drop out of all three call sites.

packages/expo-router/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### 🛠 Breaking changes
 
+- Unify JS Tabs, TopTabs, Drawer, and headless tabs with NativeTabs - only screens declared in the layout become visible.
```
jakub-agent: [🐧 nit] Missing the PR link and author attribution that the surrounding entries and `CONTRIBUTING.md` use: `- Description. ([#48499](https://github.com/expo/expo/pull/48499) by [@Ubax](https://github.com/Ubax))`.

jakub-agent (commented):

jakub-agent (commented):

packages/expo-router/src/standard-navigation/useVisibleTabsWithRedirect.ts
```diff
@@ -18,47 +20,61 @@ type TabDescriptor<Options extends object> = {
...
       return buildHref(redirectRoute);
     }
     return null;
-  }, [buildHref, redirectToRouteName, visibleRoutes]);
+  }, [buildHref, guardRedirect, routeNode?.initialRouteName, visibleRoutes]);
+
+  useEffect(() => {
+    if (__DEV__ && visibleRoutes.length === 0) {
+      console.warn(
+        `No visible routes are configured in ${routeNode?.contextKey ?? 'the layout'}, so the navigator renders nothing. ` +
```
jakub-agent: [💡 suggestion] This warning misfires when every declared screen is behind a false guard. `screenOptionsFactory` sets `hidden: true` for `isGuarded` (`useScreens.tsx:509-511`) and guarded screens keep `routeSource: 'layout'`, so:

```tsx
<Tabs>
  <Tabs.Protected guard={isSignedIn} redirectTo="/login">
    <Tabs.Screen name="feed" />
  </Tabs.Protected>
</Tabs>
```

logs "Declare at least one visible screen or trigger" whenever `isSignedIn` is false — during an auth-loading pass, for instance. The screens *are* declared, the guard redirect in the effect below is already handling it correctly, and the advice points the developer at the wrong thing.

Skipping the warning when `guardRedirect !== undefined` would cover it, or wording it to mention guards. The PR's own `prefers a protected screen redirect over the tab visibility fallback` test doesn't hit this because it keeps one unguarded screen — an all-guarded layout is untested.

packages/expo-router/src/standard-navigation/__tests__/useVisibleTabsWithRedirect.test.ios.tsx
```diff
@@ -12,9 +13,14 @@ jest.mock('../../react-navigation/native', () => {
   return { ...actualNavigation, useIsFocused: jest.fn() };
 });
 jest.mock('../useBuildHref');
+jest.mock('../../Route', () => ({
+  ...jest.requireActual('../../Route'),
+  useRouteNode: jest.fn(),
```
jakub-agent, line 18: [💡 suggestion] Mocking `useRouteNode` makes these good branch tests, but it leaves the one assumption the refactor rests on unverified: that `useRouteNode()` *inside the navigator content* resolves to the layout node rather than a screen node or `null`.

If it resolved to the wrong node, `findRouteByName(visibleRoutes, undefined)` would return `undefined` and fall through to `visibleRoutes[0]` — and every test in the suite would still pass, because the unit tests mock the node and the integration tests all expect the first visible route anyway.

One `renderRouter` case would close it: `unstable_settings: { initialRouteName: 'second' }` with `second` not first in the layout, then push an undeclared route and assert it lands on `/second` rather than the first tab.

packages/expo-router/src/layouts/TabsClient.tsx
```diff
@@ -88,6 +97,8 @@ const Tabs = unstable_integrateWithRouter<
     }),
 });
 
+const Tabs = TabsNavigator;
```
jakub-agent: [🐧 nit] Leftover from removing the wrapper — `TabsNavigator` now has exactly one use. Renaming the `unstable_integrateWithRouter` result back to `Tabs` drops the alias and restores the diff to a no-op for this file. Same in `TopTabsClient.tsx:36` (`const TopTabs = TopTabsNavigator;`) and `DrawerClient.tsx:38` (`export const Drawer = DrawerNavigator;`).

packages/expo-router/src/standard-navigation/__tests__/useVisibleTabsWithRedirect.test.ios.tsx
```diff
@@ -117,31 +125,36 @@ describe('useVisibleTabsWithRedirect', () => {
...
       useVisibleTabsWithRedirect({
         routes: [routes[2]!],
         focusedRouteKey: 'hidden-key',
         descriptors,
-        isHidden,
       })
     );
 
     expect(replaceSpy).not.toHaveBeenCalled();
+    expect(warnSpy.mock.calls).toMatchSnapshot();
```
jakub-agent, line 159: [🐧 nit] Snapshotting the warning adds five snapshot files across the suite whose entire content is one message string, so any wording change churns all five and the assertion isn't readable at the call site. `expect(warnSpy).toHaveBeenCalledWith(expect.stringContaining('No visible routes are configured'))` says the same thing inline and survives rewording.

jakub-agent (commented):

**Re-review of `d195f16b` (iteration 3).** Every item from the previous round is addressed, and the `redirectToRouteName` removal came out cleaner than I sketched — the hook now owns the whole visibility decision, and dropping the prop from `NativeTabs` too removed plumbing that predated this PR.

Confirmed fixed:

| Previous comment | Now |
| --- | --- |
| `redirectToRouteName` should be read in the hook | `WithRedirectToRouteName` deleted, the three wrapper components are gone, and the hook reads `useRouteNode()?.initialRouteName` itself. Also removed from `NativeTabs`. |
| Empty navigator warns nothing | `__DEV__` warning in the hook, covering all four navigators and naming the layout file |
| No test for the `initialRouteName` path | New `standard-navigation/__tests__/useVisibleTabsWithRedirect.test.ios.tsx` covers configured route, `/index` normalization, fallback, unfocused, and the empty case |
| `hidden` declared twice in `TabsScreenOptions` | removed |
| Reuse `normalizeRouteName`; three identical `isXHidden` helpers | both done — the `isHidden` callback is gone, replaced by an `Options extends { hidden?: boolean }` constraint |
| Docs follow-up for `href: null` | `TODO(@ubax)` added at the mapping site |
| Changelog attribution | applied |

Four small things left, none blocking — two suggestions and two nits, inline.

packages/expo-router/src/useScreens.tsx
```diff
@@ -507,10 +507,6 @@ export function screenOptionsFactory(
 
     // Prevent generated screens from showing up in the tab bar.
     if (route.internal || isGuarded) {
-      output.tabBarItemStyle = { display: 'none' };
-      output.tabBarButton = () => null;
-      // TODO: React Navigation doesn't provide a way to prevent rendering the drawer item.
-      output.drawerItemStyle = { height: 0, display: 'none' };
       output.hidden = true;
```
jakub-agent, line 512: [🚨 **critical**] Dropping `tabBarItemStyle` / `tabBarButton` / `drawerItemStyle` here breaks item hiding for every navigator built with `withLayoutContext`.

`hidden` is an Expo-only option. Grepping the package, it is read in exactly two places: `useVisibleTabsWithRedirect` (the four navigators in this PR) and NativeTabs. Upstream `@react-navigation/bottom-tabs`, `drawer`, and `material-top-tabs` ignore it — and `withLayoutContext` is public API whose own doc example (`withLayoutContext.tsx:134`) wraps `createMaterialTopTabNavigator` from `@react-navigation/material-top-tabs`.

I verified the difference by rendering a `withLayoutContext`-wrapped navigator with a `Protected guard={false}` screen and logging the options its descriptors receive:

- base (`#48498`): `{ tabBarItemStyle: { display: 'none' }, drawerItemStyle: { height: 0, display: 'none' }, tabBarButton: fn, hidden: true }`
- this commit: `{ hidden: true }`

Failure scenario: an app wraps upstream material-top-tabs via `withLayoutContext` and puts an admin screen behind `<Nav.Protected guard={isAdmin} redirectTo="/login">`. Before this PR the tab item was hidden. After it, the item renders and is pressable for every user (it redirects on press, but the screen's existence is now advertised). The same applies to `_sitemap` and `+not-found` when the root layout is itself a tabs or drawer navigator, since both are `internal: true` root children.

Keeping the three upstream options alongside `hidden` restores the old behaviour without affecting the new path — the standard navigators filter by `routeSource`/`hidden` before the view ever reads them.

packages/expo-router/src/layouts/TabsClient.tsx
```diff
@@ -58,15 +61,21 @@ const Tabs = unstable_integrateWithRouter<
         if (options.tabBarButton) {
           throw new Error('Cannot use `href` and `tabBarButton` together.');
         }
+        if (href === null) {
+          // TODO(@ubax): Update the hiding-a-tab guide for the new redirect behavior.
```
jakub-agent, line 65: [🏗️ **design**] This TODO marks a user-facing docs break that ships with the code. `docs/pages/router/advanced/tabs.mdx:175` currently reads: "Sometimes you want a route to exist but not show up in the tab bar. You can pass `href: null` to disable the button." After this change the route no longer exists as a destination — `router.push('/second')` on such a route redirects to the initial tab (per the new test at `bottom-tabs/__tests__/index.test.ios.tsx`).

Two asks:

1. Update `tabs.mdx` in this PR rather than deferring it. The package AGENTS.md asks for guides and API reference to move with the feature, and this one actively teaches a pattern that no longer works.
2. The changelog entry covers "only screens declared in the layout become visible" but not that `href: null` changed meaning. That is a separate break for anyone who read the guide, and it deserves its own line.

Worth stating explicitly in the changelog/docs: under the new rule there is no way to express "route lives in this tabs directory, is reachable by `push`, has no tab item" — neither omitting the `<Tabs.Screen>` nor `href: null` / `hidden` gets you there. The `TODO(@ubax): Show a formsheet for hidden tabs which are focused` in `useVisibleTabsWithRedirect.ts` suggests you already have this gap in mind; users hitting it need to know the intended migration (nested `Stack`?) now, not next release.

packages/expo-router/src/standard-navigation/useVisibleTabsWithRedirect.ts
jakub-agent: [🏗️ **design**] A dev-only `console.warn` is thin cover for what is now the most likely upgrade symptom. An app that relied on filesystem-driven tabs (`_layout.tsx` returning a bare `<Tabs />`, previously the documented default) renders a completely blank screen after upgrading, and the only signal is one line in a Metro log that is easy to miss among other warnings.

Two concrete suggestions:

- Throw in `__DEV__` instead of warning. This package already throws for comparable layout misconfiguration (`Cannot use 'href' and 'tabBarButton' together.`, the nested-NativeTabs error, the `processScreens` invariant), and a red screen names the file.
- Make the message actionable per the repo's error-message guidance (what / why / how). It currently states what and why but the "how" is generic. The navigator knows the filesystem route names it dropped, so it can say them: `"…renders nothing. Declare the screens you want visible, for example: <Tabs.Screen name=\"index\" />. Found undeclared routes: index, settings."`

Also: the `__DEV__` check sits inside the effect body, so the effect and its dependency array still run in production. Moving the guard outside is free, though it is a micro-detail next to the messaging.

packages/expo-router/src/ui/Tabs.tsx
```diff
@@ -184,6 +185,12 @@ export function useTabsWithTriggers(options: UseTabsWithTriggersOptions): TabsCo
     NavigationContent: RNNavigationContent,
   } = navigatorContext;
 
+  useVisibleTabsWithRedirect({
```
jakub-agent: [🏗️ **design**] Headless `Tabs` resolves guards against the *parent* navigator's guard map.

`useVisibleTabsWithRedirect` now calls `useGuardRedirect(focusedRoute.name)`, which reads `use(GuardContext)`. Only `withLayoutContext` and `views/Navigator` render a `GuardContextProvider`; `useTabsWithTriggers` does neither, and the call sits above `RNNavigationContent` anyway. So the nearest provider is the ancestor layout's — and its map is keyed by *that* navigator's route names.

Failure scenario: root `app/_layout.tsx` is a `<Stack>` with `<Stack.Protected guard={isLoggedIn} redirectTo="/login"><Stack.Screen name="settings" /></Stack.Protected>`, and `app/(tabs)/_layout.tsx` is a headless `<Tabs>` over children that include `settings`. When the tabs' `settings` route is focused but not visible (no trigger, or `hidden: true`), `guardRedirect` matches the *root's* `settings` entry and `router.replace('/login')` fires, instead of falling back to the tabs' own initial route.

The other four navigators are unaffected because `withLayoutContext` always provides a (possibly empty) map that shadows the parent. Either wrap headless `Tabs` in `GuardContextProvider` too, or have the hook take the resolved guard as a parameter so each caller supplies its own scope explicitly.

packages/expo-router/src/ui/TabContext.tsx
```diff
@@ -49,7 +49,7 @@ export type ExpoTabsNavigationProp<
 
 export type ExpoTabsScreenOptions = Pick<
   BottomTabNavigationOptions,
-  'title' | 'lazy' | 'freezeOnBlur'
+  'title' | 'lazy' | 'freezeOnBlur' | 'hidden'
```
jakub-agent, line 52: [🏗️ **design**] For headless tabs, `hidden` redirects but does not hide.

`TabList` / `TabTrigger` render from JSX children, not from descriptors, so nothing in the headless UI consults `options.hidden`. Failure scenario: a screen calls `navigation.setOptions({ hidden: true })` (the only way to reach this option in headless tabs, since triggers carry no options). Its `<TabTrigger>` still renders as a normal, enabled button; pressing it navigates to the route and `useVisibleTabsWithRedirect` immediately replaces back to the initial tab. The user sees a button that visibly bounces them.

If the intent is redirect-only, that is fine — but then `hidden` shouldn't be in the public options type until `TabList` honours it, otherwise the name promises hiding. If the intent is full parity, `TabList` needs to filter triggers whose descriptor is hidden, and that deserves a test.

packages/expo-router/src/react-navigation/bottom-tabs/navigators/createBottomTabNavigator.tsx
```diff
@@ -47,10 +53,18 @@ function BottomTabNavigatorContent({
...
+  }
+
   return (
     <BottomTabView
       {...rest}
-      state={state}
+      state={{
+        ...state,
+        routes: visibleRoutes,
+        index: visibleFocusedIndex >= 0 ? visibleFocusedIndex : 0,
```
jakub-agent: [💡 suggestion] This four-line block — the `visibleRoutes.length === 0 → return null` guard plus the `{...state, routes: visibleRoutes, index: visibleFocusedIndex >= 0 ? visibleFocusedIndex : 0}` substitution — is byte-identical in `createDrawerNavigator.tsx` and `createMaterialTopTabNavigator.tsx`, and NativeTabs repeats the same `>= 0 ? … : 0` clamp as `focusedIndex`.

Since the whole point of the PR is to unify these four navigators behind one hook, the hook is the natural home for it. Returning the clamped index (`focusedIndex`) alongside `visibleRoutes`, or returning the substituted state directly, removes four copies of a rule that must stay in sync. It also puts the `-1 → 0` fallback in one place, which matters because that fallback is what renders during the frame before the redirect effect runs.

Related, and worth a comment wherever it lands: the substituted state intentionally keeps `key`, `history`, `routeNames`, and `preloadedRouteKeys` from the real state while `routes` is filtered. That is correct for the current views (I checked `DrawerView`'s `state.key` / `state.preloadedRouteKeys` reads), but it is a non-obvious invariant that will bite whoever next touches a vendored view.

packages/expo-router/src/layouts/TabsClient.tsx
jakub-agent: [💡 suggestion] This alias does nothing — `TabsNavigator` has exactly one reference. Same pattern in `TopTabsClient.tsx` (`const TopTabs = TopTabsNavigator;`) and `DrawerClient.tsx` (`export const Drawer = DrawerNavigator;`).

If these are leftovers from an iteration that wrapped the navigator, dropping them and naming the `unstable_integrateWithRouter` result `Tabs` / `TopTabs` / `Drawer` directly keeps the diff to what actually changed. If a wrapper is coming in a later PR in the stack, a one-line comment saying so would help — right now it reads as accidental.

packages/expo-router/src/react-navigation/bottom-tabs/__tests__/index.test.ios.tsx
```diff
@@ -228,7 +228,96 @@ test('hides a tab whose href is null', () => {
...
+  expect(screen.getAllByRole('button', { name: /tab/ })).toHaveLength(1);
+
+  act(() => router.push('/second'));
+
+  expect(screen).toHavePathname('/');
+  expect(screen.getByTestId('index')).toBeVisible();
+});
+
+test('renders no tab UI when no screens are declared in the layout', () => {
+  const warnSpy = jest.spyOn(console, 'warn').mockImplementation(() => {});
```
jakub-agent: [🐧 nit] The spy leaks if an assertion fails. `mockRestore()` is the last statement in the test body, so any earlier failure leaves `console.warn` stubbed for the remainder of the file, and later tests silently swallow warnings they were meant to surface. The package's own AGENTS.md prescribes `beforeEach` + `afterEach(() => spy.mockRestore())` for exactly this. Same shape in the new drawer, material-top-tabs, native-tabs, and `useVisibleTabsWithRedirect` tests.

While there: `expect(warnSpy.mock.calls).toMatchSnapshot()` adds four new snapshot files that each hold the same single warning string. `expect(warnSpy).toHaveBeenCalledWith(expect.stringContaining('No visible routes are configured'))` keeps the expectation next to the assertion and drops four files that a future message tweak would otherwise force you to re-record.

jakub-agent (changes requested):

packages/expo-router/src/useScreens.tsx
Ubax: Add TODO to update docs. `withLayoutContext` will be deprecated. Users will need to use the new standard-navigation API, which offers `processScreens`. Based on `hidden` option they will be able to adjust specific options if needed

packages/expo-router/src/layouts/TabsClient.tsx
Ubax: I cannot update the docs, since they are reflecting current SDK 57 version and not our refactored one

packages/expo-router/src/standard-navigation/useVisibleTabsWithRedirect.ts
Ubax: For now only update the message. Add todo to consider changing to throw

packages/expo-router/src/ui/TabContext.tsx
Ubax: Filter and test

packages/expo-router/src/react-navigation/bottom-tabs/navigators/createBottomTabNavigator.tsx
Ubax: Clamp the index in the hook

packages/expo-router/src/ui/TabTrigger.tsx
```diff
@@ -69,12 +69,18 @@ const TabTriggerSlot = Slot as React.ForwardRefExoticComponent<TabTriggerSlotPro
...
 export function TabTrigger({ asChild, name, href, resetOnFocus, ...props }: TabTriggerProps) {
+  const { descriptors } = useNavigatorContext();
   const { trigger, triggerProps } = useTabTrigger({
     name,
     resetOnFocus,
     ...props,
   });
 
+  const routeOptions = trigger?.route && descriptors[trigger.route.key]?.options;
+  if (routeOptions && 'hidden' in routeOptions && routeOptions.hidden === true) {
```
jakub-agent: [🏗️ **design**] This hides the wrong trigger when tabs are nested.

`trigger.route` comes from `getTrigger`, which resolves `state.routes[config.index]`. `config.index` is assigned in `triggersToScreens` while iterating *that* navigator's `sortedConfigs`, but `triggerMap` starts as `{ ...parentTriggerMap }` — so entries inherited from an outer navigator keep the outer navigator's index and are then applied to the inner navigator's `state.routes`. The route you get back is unrelated to the trigger.

Repro (verified at this commit): outer headless `Tabs` with triggers `a` → `/a` and `b` → `/b`; `app/a/_layout.tsx` is a nested `Tabs` whose inner tab `y` (index 1) calls `navigation.setOptions({ hidden: true })`; a `<TabTrigger name="b" />` sits outside the inner `TabList` (the documented cross-navigator switch pattern). `triggerMap.b.index === 1` → `state.routes[1]` of the *inner* navigator is `y` → hidden → **the `b` trigger disappears**. Same tree with `y` not hidden renders it, so the mis-indexing is what drives it.

The index/route mismatch in `getTrigger` predates this PR, but nothing read `trigger.route.key` against the local descriptors before, so this is where it becomes user-visible. Resolving the descriptor by route *name* scoped to the current navigator would sidestep it:

```ts
const route = trigger?.type === 'internal'
  ? state.routes.find((r) => r.name === trigger.routeNode.route)
  : undefined;
```

Worth a test with nested tabs either way — the current suite only covers hiding in a single, flat navigator.

packages/expo-router/src/ui/TabTrigger.tsx
```diff
@@ -69,12 +69,18 @@ const TabTriggerSlot = Slot as React.ForwardRefExoticComponent<TabTriggerSlotPro
  * ```
  */
 export function TabTrigger({ asChild, name, href, resetOnFocus, ...props }: TabTriggerProps) {
+  const { descriptors } = useNavigatorContext();
```
jakub-agent: [💡 suggestion] Two small things now that `hidden` is part of trigger rendering.

`useTabTrigger` already calls `useNavigatorContext()` on the next line and holds `state`; this second call re-reads the same context. Returning `descriptors` (or a resolved `hidden` flag) from `useTabTrigger` would collapse them.

More usefully: `useTabTrigger` is documented as the extension point for custom triggers ("Utility hook creating custom `TabTrigger`"), and it exposes `trigger`, `triggerProps`, `getTrigger`, and `switchTab` — but no way to tell that a trigger is hidden. So `hidden: true` hides the built-in `TabTrigger` and leaves every custom trigger visible, which is the same split this PR is trying to close for the five navigators. Putting `hidden` on the `Trigger` type (and on `getTrigger`'s result) would let custom triggers opt in with one check, and would also give `TabTrigger` its answer without the extra context read.

jakub-agent (changes requested):

**Iteration 2** (was `d195f16`, now `e58f5fe`). PR 3/6 in the stack (base: #48498).

packages/expo-router/src/ui/TabTrigger.tsx
```diff
@@ -152,14 +157,23 @@ export function useTabTrigger(options: TabTriggerProps): UseTabTriggerResult {
...
+        config.type === 'internal'
+          ? state.routes.find((route) => route.name === config.routeNode.route)
+          : undefined;
+      const routeOptions = route != null ? descriptors[route.key]?.options : undefined;
+      const hidden =
+        routeOptions != null && 'hidden' in routeOptions && routeOptions.hidden === true;
+
       return {
+        hidden,
         isFocused: state.index === config.index,
```
jakub-agent, line 170: [🏗️ **design**] Now that the correct route lookup exists three lines up, `route` and `isFocused` are the two fields still using `config.index` — and for triggers inherited from a parent navigator that index belongs to the parent's trigger list, not this navigator's `state.routes`.

This is pre-existing (neither field changed in this PR), so treat it as a follow-up rather than a merge blocker. Flagging it because the new `hidden` work sits right on top of it, and because the fix is now nearly free.

Reproduced at this commit — outer tabs `[a, b]`, `app/a/_layout.tsx` a nested tabs `[index, y]`, and an inherited `<TabTrigger name="b" />` outside the inner `TabList`. Nothing hidden anywhere:

| inner navigator on | `trigger.route.name` | `trigger.isFocused` | press inherited `b` |
|---|---|---|---|
| `index` (0) | `y` ❌ (want `b`) | `false` | navigates to `/b` ✅ |
| `y` (1) | `y` ❌ | `true` ❌ | **no-op** ❌ |

The second row is the user-visible part: `b`'s index in the outer navigator is 1, the inner navigator's index is also 1, so `isFocused` comes out true, `handleOnPress` takes the `if (!trigger.isFocused)` branch and skips `switchTab` — the trigger renders normally and does nothing. It also emits `tabPress` with `target: trigger.route.key`, i.e. the key of an unrelated inner route.

So an inherited trigger silently dies whenever the nested navigator's current index collides with the trigger's index upstream.

Reusing the `route` you already compute covers both fields. `route` has to become optional, since an inherited trigger genuinely has no route in this navigator's state — which also makes the two `trigger.route.key` reads in `handleOnPress` / `handleOnLongPress` fall back to `href` for that case, matching how external triggers are already handled:

```ts
return {
  hidden,
  route,
  isFocused: route != null && state.routes[state.index]?.key === route.key,
  resolvedHref: stripGroupSegmentsFromPath(appendBaseUrl(config.href)),
  ...config,
};
```

Worth a test with a nested navigator sitting on the colliding index — the new `does not hide an inherited trigger…` test happens to land on index 0, so it passes either way.

packages/expo-router/src/ui/TabTrigger.tsx
```diff
@@ -152,14 +157,23 @@ export function useTabTrigger(options: TabTriggerProps): UseTabTriggerResult {
...
         route: state.routes[config.index]!,
         resolvedHref: stripGroupSegmentsFromPath(appendBaseUrl(config.href)),
         ...config,
       };
     },
-    [triggerMap]
+    [descriptors, state, triggerMap]
```
jakub-agent, line 176: [🐧 nit] `getTrigger` is part of the public `useTabTrigger` result, and adding `descriptors` and `state` to the deps means its identity now changes on every navigation instead of only when the trigger map does. Correct — it has to read both — but anyone who memoised on `getTrigger` loses that memo per navigation.

If that turns out to matter, narrowing the closure to what it actually needs (`state.routes`, `state.index`) would cut most of the churn, since `descriptors` gets a fresh object on nearly every render anyway. Not worth changing on spec — just noting it in case a re-render report shows up later.

jakub-agent (commented):

jakub-agent (approved):

**Approving** at `95f6da7` (iteration 4 of this review). Every finding from the earlier passes is either fixed or resolved as intentional, and I found nothing new in this revision.

PR #48502, opened 2026-08-04
[router] Remove `navigationKey` API

Ubax:
# Why

Remove `navigationKey` API. It was never supported in router, so we can safely remove it.

# How

# Test Plan

CI

# Checklist

---

### DIFF
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -5,8 +5,10 @@
 ### 🛠 Breaking changes
 
 - Unify JS Tabs, TopTabs, Drawer, and headless tabs with NativeTabs - only screens declared in the layout become visible. ([#48499](https://github.com/expo/expo/pull/48499) by [@Ubax](https://github.com/Ubax))
+- Make `href: null` hide JS tabs and make their routes unreachable, redirecting navigation to the initial tab. ([#48499](https://github.com/expo/expo/pull/48499) by [@Ubax](https://github.com/Ubax))
 - Migrate the JS `TopTabs` navigator to the `standard-navigation` integration. ([#48498](https://github.com/expo/expo/pull/48498) by [@Ubax](https://github.com/Ubax))
 - Make `beforeRemove` non-preventable and add `removePrevented` event. ([#48347](https://github.com/expo/expo/pull/48347) by [@Ubax](https://github.com/Ubax))
+- Remove the `navigationKey` prop from layout `<Screen>` and `<Group>` components.
 - Remove the `redirect` prop from layout `<Screen>` components. ([#48369](https://github.com/expo/expo/pull/48369) by [@Ubax](https://github.com/Ubax))
 - Add `redirectTo` to protected routes and render guarded screens as redirects instead of removing them from navigators. ([#47744](https://github.com/expo/expo/pull/47744) by [@Ubax](https://github.com/Ubax))
 - Migrate the `Drawer` navigator to the `standard-navigation` integration. ([#47839](https://github.com/expo/expo/pull/47839) by [@Ubax](https://github.com/Ubax))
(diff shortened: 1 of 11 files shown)


packages/expo-router/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### 🛠 Breaking changes
 
+- Remove the `navigationKey` prop from layout `<Screen>` and `<Group>` components.
```
jakub-agent: [💡 suggestion] This wording points readers at the wrong components. expo-router's layout `<Screen>` is `views/Screen`, typed by `ScreenProps` in `src/useScreens.tsx`, which never had a `navigationKey` prop — and `getSortedChildren` only forwards `{ initialParams, listeners, options, getId }`, so the prop was already inert there. `withLayoutContext` also only attaches `Screen` and `Protected`, so layouts never exposed `<Group>` at all.

The prop actually lived on the vendored React Navigation `Screen`/`Group` (`RouteConfigProps` / `RouteGroupConfig`), reachable through `expo-router/react-navigation` and `createNavigatorFactory`. As written, app authors will look for `<Stack.Screen navigationKey>` usages to migrate and find that it never did anything. Suggest scoping the entry to the vendored React Navigation `Screen`/`Group`.

packages/expo-router/src/react-navigation/routers/types.tsx
```diff
@@ -173,16 +173,7 @@ export type Router<State extends NavigationState, Action extends NavigationActio
...
-    state: State,
-    options: RouterConfigOptions & {
-      /**
-       * List of routes whose key has changed even if they still have the same name.
-       * This allows to remove screens declaratively.
-       */
-      routeKeyChanges: string[];
-    }
-  ): State;
+  getStateForRouteNamesChange(state: State, options: RouterConfigOptions): State;
```
jakub-agent, line 176: [🏗️ **design**] This drops `routeKeyChanges` from the exported `Router` type, which is a second breaking change the changelog does not mention. Anyone implementing a custom router — via the `UNSTABLE_router` prop, or by extending `TabRouter`/`StackRouter` the way `NativeBottomTabsRouter` does — and destructuring `routeKeyChanges` in `getStateForRouteNamesChange` now gets a type error. Nothing inside expo-router does this, so only external code is affected. Worth its own changelog bullet under breaking changes so custom-router authors know what to change.

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
```diff
@@ -519,33 +487,19 @@ export function useNavigationBuilder<
...
     // The state won't update until there's a change needed in the state we have initialized locally
     // So it'll be `undefined` or stale until the first navigation event happens
     isStateInitialized(currentState) ? (currentState as State) : (initializedState as State);
 
   let nextState: State = state;
-  if (
-    !isArrayEqual(state.routeNames, routeNames) ||
-    !isRecordEqual(routeKeyList, previousRouteKeyList)
-  ) {
+  if (!isArrayEqual(state.routeNames, routeNames)) {
```
jakub-agent, line 498: [💡 suggestion] No test asserts the prop is gone. `src/react-navigation/` is a vendored fork of React Navigation, where `navigationKey` still exists upstream, so a future sync can quietly reintroduce it along with the `keys`/`routeKeyChanges` plumbing this PR just removed.

#48369 in this same stack set the precedent for guarding a removal at the type level: `expectTypeOf<ScreenProps>().not.toHaveProperty('redirect')` in `src/__tests__/stacks.test.ios.tsx`. The equivalent here would be asserting `RouteConfigProps` and `RouteGroupConfig` have no `navigationKey` property. Cheap, and it makes the removal stick.

jakub-agent (commented):

packages/expo-router/CHANGELOG.md
Ubax: Remove the CHANGELOG entry, since there is no change from router API perspective. react-navigation is not important there

packages/expo-router/src/react-navigation/routers/types.tsx
Ubax: This will be covered by follow-up PR changelog

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
Ubax: That's ok. There will be no sync in the future

jakub-agent (approved):

PR #48506, opened 2026-08-04
[observe][app-metrics] Allow disabling unhandled JS error handling via configure

tsapeta:
# Why

We install the unhandled-JS-error handler as soon as `expo-app-metrics` is imported, and there was no way for an app to turn that off.

Closes ENG-21940.

# How

Added an `errorHandlingEnabled` option to `ObserveConfig`, defaulting to `true`.

It can't stop the handler from being installed (that happens on import, long before anyone calls `configure`), so it gates the reporting inside the handler instead. Errors thrown before `configure` runs still get recorded. Everything else is untouched: the red box, `reportError`, `ObserveErrorBoundary`, native crash reporting.

# Test Plan

New unit tests for the default, both explicit values, suppression, re-enabling, and that the previous handler still runs while reporting is off. `pnpm jest` passes in both packages (40 and 538), as do `typecheck`, `lint` and `format`.

They install a fake `ErrorUtils` though, so the real red box is worth a manual check in `apps/observe-tester` with the option off.

### DIFF
--- a/packages/expo-observe/CHANGELOG.md
+++ b/packages/expo-observe/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Add `Observe.registerIntegration` to register an integration ([#48245](https://github.com/expo/expo/pull/48245), [#48268](https://github.com/expo/expo/pull/48268) by [@Ubax](https://github.com/Ubax))
 - Expose `ObserveErrorBoundary`, a React error boundary that records render-phase errors. ([#47341](https://github.com/expo/expo/pull/47341) by [@tsapeta](https://github.com/tsapeta))
 - Add `reportError` to report caught, non-fatal errors from your own `try`/`catch` blocks. ([#47871](https://github.com/expo/expo/pull/47871) by [@tsapeta](https://github.com/tsapeta))
+- Add an `errorHandlingEnabled` option to `configure` to opt out of recording unhandled JavaScript errors. ([#48506](https://github.com/expo/expo/pull/48506) by [@tsapeta](https://github.com/tsapeta))
 
 ### 🐛 Bug fixes
 
--- a/packages/expo-observe/src/__tests__/module.test.native.ts
+++ b/packages/expo-observe/src/__tests__/module.test.native.ts
@@ -170,9 +170,12 @@ describe('module Proxy', () => {
     expect(warnSpy).not.toHaveBeenCalled();
   });
 
-  it('forwards errorHandlingEnabled to native along with the rest of the config', () => {
+  it('passes the config through to the native module unchanged', () => {
     const Observe = loadModule();
     Observe.configure({ environment: 'test', errorHandlingEnabled: false });
+    // The native `Config` record has no `errorHandlingEnabled` field, so decoding drops it. This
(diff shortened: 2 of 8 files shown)


packages/expo-observe/src/__tests__/module.test.native.ts
```diff
@@ -148,6 +155,38 @@ describe('module Proxy', () => {
...
+
+  it.each([true, false])('applies errorHandlingEnabled: %s to the error handler', (enabled) => {
+    const Observe = loadModule();
+    Observe.configure({ environment: 'test', errorHandlingEnabled: enabled });
+    expect(mockSetErrorHandlerEnabled).toHaveBeenCalledTimes(1);
+    expect(mockSetErrorHandlerEnabled).toHaveBeenCalledWith(enabled);
+    expect(warnSpy).not.toHaveBeenCalled();
+  });
+
+  it('forwards errorHandlingEnabled to native along with the rest of the config', () => {
```
tsapeta: [🏗️ **design**] This test name claims more than the assertion checks. `Config` on iOS (`ios/ObserveModule.swift:8-14`) and Android (`android/.../ObserveModule.kt:15-22`) has no `errorHandlingEnabled` field, so `Record` decoding drops it. What this verifies is that the JS layer passes the config object through unmodified, which is real but is not "forwards to native."

Not a bug — the gate is entirely in JS and native has no use for the flag. The risk is the next reader: a test named this way suggests the field is already wired natively, so someone adding native error capture may assume the plumbing exists. Either rename it to something like `passes the config through to the native module unchanged`, or add the field to both `Config` records so the name becomes true.

packages/expo-observe/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 - Add `Observe.registerIntegration` to register an integration ([#48245](https://github.com/expo/expo/pull/48245), [#48268](https://github.com/expo/expo/pull/48268) by [@Ubax](https://github.com/Ubax))
 - Expose `ObserveErrorBoundary`, a React error boundary that records render-phase errors. ([#47341](https://github.com/expo/expo/pull/47341) by [@tsapeta](https://github.com/tsapeta))
 - Add `reportError` to report caught, non-fatal errors from your own `try`/`catch` blocks. ([#47871](https://github.com/expo/expo/pull/47871) by [@tsapeta](https://github.com/tsapeta))
+- Add an `errorHandlingEnabled` option to `configure` to opt out of recording unhandled JavaScript errors. ([#48506](https://github.com/expo/expo/pull/48506) by [@tsapeta](https://github.com/tsapeta))
```
tsapeta, line 12: [💡 suggestion] `expo-app-metrics/CHANGELOG.md` has no matching entry, but this PR adds `setErrorHandlerEnabled` to that package's public exports and changes the behavior of its installed handler. Even though the export is `@internal` and the user-facing option lives in `expo-observe`, the app-metrics change is worth a line under `### 💡 Others` — the existing `Add a `caught` source to the private `reportError`` entry is precedent for logging internal surface changes there.

Ubax (approved):
LGTM 👍 

Let's wait for ai to take a look before merging

PR #48509, opened 2026-08-04
[docs] allow omitting agent instructions from markdown responses

davidmokos:
Adds a new url parameter `?includeAgentInstructions=false` when fetching markdown docs that will disable the agent instruction block. This is needed for the MCP server's search_documentation tool because Anthropic review did not allow prompt injection in the connector marktplace.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -164,6 +164,13 @@ Every published page is served in two formats: HTML for browsers, and markdown f
 curl -H "Accept: text/markdown" https://docs.expo.dev/get-started/set-up-your-environment/
 ```
 
+The generated markdown includes an `<AgentInstructions>` block by default. To omit it from a
+content-negotiated response, set `includeAgentInstructions=false`:
+
+```sh
+curl -H "Accept: text/markdown" "https://docs.expo.dev/get-started/set-up-your-environment/?includeAgentInstructions=false"
+```
+
 ### 3. Sibling `.md` URLs via `_redirects`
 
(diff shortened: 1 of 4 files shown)


amandeepmittal (approved):

Kudo:
interesting! was thinking the change would be adding `AgentInstructions` when no `includeAgentInstructions=false`. now the change looks like adding AgentInstructions first then stripping it. that feels similar to we did in mcp server. is it easy to conditionally add AgentInstructions than stripping after?

davidmokos:
yeah I don't like it either but all the md pages are prebuilt. So we can either
1. prebuild 2 version for each page (with and without the agent instructions)
2. remove via regex (current impl in this PR)
3. remove via regex in the mcp server
4. build the md pages on each request (considerably bigger change)

Kudo:
> yeah I don't like it either but all the md pages are prebuilt. So we can either
> 
> 1. prebuild 2 version for each page (with and without the agent instructions)
> 2. remove via regex (current impl in this PR)
> 3. remove via regex in the mcp server
> 4. build the md pages on each request (considerably bigger change)

good point for prebuilt pages. in this case i think 2 or 3 would be better. if mcp is the only user case to hide AgentInstructions block, then we strip in mcp server. otherwise if there's other use case we need to hide AgentInstructions block, we do as this pr.

davidmokos:
ok @Kudo I suggest we do the stripping in the mcp server. Closing this PR

Kudo:
thanks David for working on this. will strip from mcp server https://github.com/expo/universe/pull/29621

PR #48510, opened 2026-08-04
[filesystem][android] Encode the renamed URI

### LINKED ISSUE #48496: [expo-file-system] Android: rename() stores an unencoded URI, so reading .uri afterwards throws for names containing a space

No repo link — the bug reproduces in ~5 lines against a stock `expo-file-system` install, and the root cause is pinned to a specific line below. Drop this in any Expo 56 Android app:

```ts
import { File, Paths } from "expo-file-system";

const file = new File(Paths.cache, "source.pdf");
file.create();

file.rename("My Report.pdf"); // succeeds — the file really does move on disk
console.log(file.uri);        // ← throws on Android, fine on iOS
```

Any name containing a space triggers it (`"a b.pdf"` is enough). Happy to publish a full repro repo if the snippet isn't sufficient.

Android

in a development build

### Summary

On Android, `File.rename()` moves the file correctly on disk but stores a **malformed, unencoded URI** on the object. Any subsequent read of `.uri` — or anything that resolves the file — throws `IllegalArgumentException`. iOS is unaffected.

```
[Error: Exception in HostFunction: java.lang.IllegalArgumentException:
 Illegal character in path at index 63:
 file:///data/user/0/com.example.app/cache/Print/Cities and City Vocabulary.pdf
   at java.net.URI.create(URI.java:848)
   at expo.modules.filesystem.unifiedfile.JavaFile.<init>(JavaFile.kt:22)
   at expo.modules.filesystem.FileSystemPath.getFile(FileSystemPath.kt:73)
   at expo.modules.filesystem.FileSystemFile.asString(FileSystemFile.kt:124)
 Caused by: java.net.URISyntaxException: Illegal character in path at index 63
```

Index 63 is exactly the space after `Cities`.

#### Root cause

`rename()` sets the new URI via `renamedUri()`, which appends the raw name with no percent-encoding — [`android/src/main/java/expo/modules/filesystem/FileSystemPath.kt:210-216`](https://github.com/expo/expo/blob/main/packages/expo-file-system/android/src/main/java/expo/modules/filesystem/FileSystemPath.kt):

```kotlin
private fun renamedUri(newName: String): Uri {
  val currentUri = uri.toString()
  val currentPath = currentUri.trimEnd('/')
  val parentUri = currentUri.substring(0, currentPath.lastIndexOf('/') + 1)
  val renamedUri = Uri.withAppendedPath(Uri.parse(parentUri), newName).toString()
  return Uri.parse(if (this is FileSystemDirectory) "$renamedUri/" else renamedUri)
}
```

`Uri.withAppendedPath` does **not** encode `newName`. The next read goes `.uri` → `FileSystemFile.asString()` → `getFile` → `JavaFile(uri)`, whose constructor is `File(URI.create(uri.toString()))` (`unifiedfile/JavaFile.kt:22`), and `java.net.URI` rejects a raw space.

#### Relationship to #35619

This looks like the same family as #35619, but a **code path that fix did not cover**. That issue was about the *constructor* mishandling such URIs, and the fix added JS-side encoding: the `File` constructor runs `Paths.join` → `encodeURLChars` → `encodePathChars`, turning spaces into `%20` (`src/pathUtilities/url.ts`).

`rename()` regenerates the URI **natively**, so it bypasses that JS encoding entirely. The result is an asymmetry within the same class:

| | resulting `.uri` |
| --- | --- |
| `new File(dir, "My Report.pdf")` | `.../My%20Report.pdf` — valid ✅ |
| `file.rename("My Report.pdf")` | `.../My Report.pdf` — throws on next read ❌ |

Worth noting only space and control characters actually matter here — `java.net.URI` deliberately permits non-ASCII in the path, so CJK/Turkish/accented filenames are fine. That makes this easy to miss in testing: a name has to contain a literal space.

#### Suggested fix

Encode `newName` inside `renamedUri()` so the object's URI stays a valid `file://` URI after a rename, mirroring what `encodePathChars` already does on the JS side. iOS gets this for free because `getRenamedUrl` uses `NSURL.appendingPathComponent`, which encodes — hence the platform split.

#### Workaround for anyone hitting this

Read the URI from a JS-constructed handle to the destination rather than from the renamed object:

```ts
const target = new File(printed.parentDirectory, name);
printed.rename(name);
return target.uri; // JS-encoded, so safe to read
```

#### How this surfaced in the wild

`expo-print`'s `printToFileAsync()` → `rename(noteTitle + ".pdf")` → `Sharing.shareAsync(file.uri)`, i.e. "export this document as a PDF named after its title". Any document whose title contained a space failed to share on Android, and the throw happened at the `.uri` read, before sharing was ever reached.

### Environment

```text
  System:
    OS: macOS 26.3
  Binaries:
    Node: 20.19.4
  npmPackages:
    expo: ^56.0.0 => 56.0.16
    expo-file-system: ~56.0.8 => 56.0.8
    expo-print: ~56.0.4 => 56.0.4
    expo-sharing: ~56.0.19 => 56.0.22
    react: 19.2.3 => 19.2.3
  Expo Workflow: managed (development build)
```

intergalacticspacehighway:
# Why

Fixes https://github.com/expo/expo/issues/48496

As shown in the above repro, when trying to read `file.uri` of a file name that includes spaces it throws below exception.

```
import { File, Paths } from "expo-file-system";

const file = new File(Paths.cache, "source.pdf");
file.create();

file.rename("My Report.pdf"); // succeeds — the file really does move on disk
console.log(file.uri);        // ← throws on Android, fine on iOS

// Exception
at java.net.URI.create(URI.java:848)
at expo.modules.filesystem.unifiedfile.JavaFile.<init>(JavaFile.kt:22)
at expo.modules.filesystem.FileSystemPath.getFile(FileSystemPath.kt:73)
```

# How

Use [appendPath](https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/net/Uri.java#1475) when renaming file which encodes the file name. The existing [withAppendedPath](https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/net/Uri.java#2342) required already encoded string.

# Test Plan

Added test cases for user shared repro in `FileSystemFileTest.kt`.

# Checklist

### DIFF
--- a/packages/expo-file-system/CHANGELOG.md
+++ b/packages/expo-file-system/CHANGELOG.md
@@ -18,6 +18,7 @@
 - Added `./next` subpath to package `exports` field to resolve Metro bundler warning. ([#44793](https://github.com/expo/expo/pull/44793) by [@chang-in](https://github.com/chang-in))
 - Fixed `FileHandle` security-scoped access, and non-SAF `content://` URI support. ([#47176](https://github.com/expo/expo/pull/47176) by [@barthap](https://github.com/barthap))
 - Fixed potential file offset races when asynchronous and synchronous `FileHandle` operations overlap on Android and iOS. ([#47945](https://github.com/expo/expo/pull/47945) by [@wh201906](https://github.com/wh201906))
+- [android] Fixed `rename()` storing an unencoded URI, so reading `.uri` afterwards threw for names containing a space. ([#48496](https://github.com/expo/expo/issues/48496) by [@yagiz2000](https://github.com/yagiz2000), [#48510](https://github.com/expo/expo/pull/48510) by [@intergalacticspacehighway](https://github.com/intergalacticspacehighway))
 
 ### 💡 Others
 
--- a/packages/expo-file-system/android/src/main/java/expo/modules/filesystem/FileSystemPath.kt
+++ b/packages/expo-file-system/android/src/main/java/expo/modules/filesystem/FileSystemPath.kt
@@ -212,7 +212,7 @@ abstract class FileSystemPath(var uri: Uri) : SharedObject() {
     val currentUri = uri.toString()
     val currentPath = currentUri.trimEnd('/')
     val parentUri = currentUri.substring(0, currentPath.lastIndexOf('/') + 1)
-    val renamedUri = Uri.withAppendedPath(Uri.parse(parentUri), newName).toString()
+    val renamedUri = Uri.parse(parentUri).buildUpon().appendPath(newName).build().toString()
     return Uri.parse(if (this is FileSystemDirectory) "$renamedUri/" else renamedUri)
   }
 
--- a/packages/expo-file-system/android/src/test/java/expo/modules/filesystem/FileSystemFileTest.kt
+++ b/packages/expo-file-system/android/src/test/java/expo/modules/filesystem/FileSystemFileTest.kt
@@ -217,6 +217,48 @@ class FileSystemFileTest {
     assertTrue(source.exists())
   }
 
+  @Test
+  fun renameEncodesNameSoTheFileStaysReadableThroughItsUri() {
+    val parent = temporaryFolder.newFolder("rename-parent-with-space")
+    val source = File(parent, "source.pdf").apply {
+      writeText("content")
+    }
+    val file = FileSystemFile(Uri.fromFile(source))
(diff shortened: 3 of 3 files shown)


gubonlucid-com (commented):
nnn

barthap (approved):

PR #48511, opened 2026-08-04
[submit-expo-feedback] Add evals feedback category

zvadaadam:
# Why

The Expo skills now instruct AI agents to report Expo tasks that models repeatedly fail at as **eval candidates** (expo/skills#126) — structured leads the team turns into agent evals. Those submissions need a first-class `evals` category so they can be filtered and clustered downstream, but the category list is validated client-side in this CLI, so `--category evals` currently exits with `Invalid feedback category`. The feedback API itself accepts the new value without changes.

# How

- Added `evals` to `FEEDBACK_CATEGORIES` (before `unknown`).
- Added the subject guidance row to the `--help` table: the Expo package, command, or capability the failed task involves.
- Gave the interactive category picker a descriptive title for `evals`, matching the special-cased `unknown` entry.

# Test Plan

- Updated the help-output test to assert the new table row; the invalid-category error message picks up the new value automatically from the enum.
- Existing category resolution tests are unaffected (`evals` follows the same `resolveFeedbackCategory` path as the other values).

### DIFF
--- a/packages/submit-expo-feedback/src/__tests__/cli.test.ts
+++ b/packages/submit-expo-feedback/src/__tests__/cli.test.ts
@@ -73,6 +73,9 @@ describe('help output', () => {
       );
       expect(helpOutput).toContain('| eas-cli    | Full EAS CLI command, such as eas build');
       expect(helpOutput).toContain(
+        '| evals      | Expo package, command, or capability the task involves'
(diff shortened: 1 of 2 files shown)


davidmokos (approved):
LGTM

PR #48514, opened 2026-08-04
[autolinking][Android] Don't hardcode default `cmakeListsPath` for pure C++ modules

satya164:
# Why

Currently, the `cmakeListsPath` path is always defaulted to a generated file path in the library. This breaks autolinking when a C++ library doesn't ship with prebuilt codegen files (with `includesGeneratedCode: false`).

# How

This removes the incorrect default `cmakeListsPath` from android autolinking for pure C++ modules.

Mirrors the same change in Community CLI: https://github.com/react-native-community/cli/pull/2799

Related fix in react-native: https://github.com/react/react-native/pull/56938

# Test Plan

TODO

# Checklist

### DIFF
--- a/packages/expo-modules-autolinking/src/reactNativeConfig/androidResolver.ts
+++ b/packages/expo-modules-autolinking/src/reactNativeConfig/androidResolver.ts
@@ -72,14 +72,18 @@ export async function resolveDependencyConfigImplAndroidAsync(
     (await parseComponentDescriptorsAsync(packageRoot, packageJson));
   let cmakeListsPath = reactNativeConfig?.cmakeListsPath
     ? path.join(androidDir, reactNativeConfig?.cmakeListsPath)
-    : path.join(androidDir, 'build/generated/source/codegen/jni/CMakeLists.txt');
+    : isPureCxxDependency
+      ? null
+      : path.join(androidDir, 'build/generated/source/codegen/jni/CMakeLists.txt');
   const cxxModuleCMakeListsModuleName = reactNativeConfig?.cxxModuleCMakeListsModuleName || null;
   const cxxModuleHeaderName = reactNativeConfig?.cxxModuleHeaderName || null;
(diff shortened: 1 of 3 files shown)


packages/expo-modules-autolinking/src/reactNativeConfig/androidResolver.ts
```diff
@@ -72,14 +72,18 @@ export async function resolveDependencyConfigImplAndroidAsync(
...
   const cxxModuleCMakeListsModuleName = reactNativeConfig?.cxxModuleCMakeListsModuleName || null;
   const cxxModuleHeaderName = reactNativeConfig?.cxxModuleHeaderName || null;
   let cxxModuleCMakeListsPath = reactNativeConfig?.cxxModuleCMakeListsPath
     ? path.join(androidDir, reactNativeConfig?.cxxModuleCMakeListsPath)
     : null;
   if (process.platform === 'win32') {
-    cmakeListsPath = cmakeListsPath.replace(/\\/g, '/');
+    if (cmakeListsPath) {
+      cmakeListsPath = cmakeListsPath.replace(/\\/g, '/');
+    }
```
Kudo:
```suggestion
    cmakeListsPath = cmakeListsPath?.replace(/\\/g, '/');
```

Kudo (approved):
nit: a little tweak. it mostly looks good otherwise

packages/expo-modules-autolinking/src/reactNativeConfig/androidResolver.ts
satya164: @Kudo reverted this as it caused the types to fail due to null/undefined values.

Kudo: alright. sorry for the incorrect review

kitten:
I'm not sure how far this was backported in RN CLI, if at all, but will pick this to `sdk-56` and `sdk-57` since it doesn't seem breaking

PR #48518, opened 2026-08-04
[app-metrics] Attach network connection quality to metrics

tsapeta:
# Why

We report which transport a session was on, but nothing about whether it actually worked. "cellular" doesn't tell you the user was in an elevator, and that's what you need to filter those sessions out.

Closes ENG-25647.

# How

Nine new attributes, all derived from data we already collect.

**Connection cost**

| Key | Type | Meaning |
| --- | --- | --- |
| `expo.network.isExpensive` | bool | Using this network may cost the user money: cellular, or a tethered hotspot. From `NWPath.isExpensive` on iOS, the inverse of `NET_CAPABILITY_NOT_METERED` on Android. Only emitted when there's a usable network, so `false` means "there was one and it wasn't metered" rather than "there was nothing". |
| `expo.network.isConstrained` | bool | iOS only. Low Data Mode is on for this path, so the user asked to conserve data on this network. |
| `expo.network.dataSaverEnabled` | bool | Android only. Data Saver is restricting this app's background traffic. Deliberately a separate key from `isConstrained`: Low Data Mode is per-path and changes as the user moves between networks, while Data Saver is process-wide. |

**Throughput**

| Key | Type | Meaning |
| --- | --- | --- |
| `expo.network.requests.throughputBytesPerSecond` | double | Received bytes over the time those bytes were actually moving. The denominator is the union of the transfer windows of the requests that received bytes, each running from the first response byte to the last, so server think time isn't charged to the connection, request concurrency can't inflate it, and idle time between requests is excluded. Cache hits drop out for free, since a response served from disk never reports a first byte. Transfers shorter than a millisecond are excluded on both platforms: Android's clock can't resolve them, and dividing by a duration the clock didn't measure says more about the timer than the network. |

**The slowest completed request**

One request described by five fields, so they can be read together: a `duration` mostly made up of `timeToFirstByte` means a slow server, while a small `timeToFirstByte` against a large `bytesReceived` means a slow transfer.

A request qualifies as long as it produced a response, so a 503 that took eight seconds and a transfer that broke mid-body both stay candidates: the app waited for them, and on a launch that felt slow that wait is usually the answer. Only requests that never got a status code are excluded, since a timeout's duration is the client's own setting rather than a measurement of the network.

| Key | Type | Meaning |
| --- | --- | --- |
| `expo.network.requests.slowest.host` | string | Host of the request. Absent if the URL had no resolvable host. |
| `expo.network.requests.slowest.duration` | seconds | Total wall-clock duration. |
| `expo.network.requests.slowest.statusCode` | int | What the server answered. Disambiguates an empty response: 0 bytes is routine on a 304, intentional on a 204, and a problem on a 200. |
| `expo.network.requests.slowest.timeToFirstByte` | seconds | Time until the first response byte. Includes server processing time, so it's a proxy for network quality rather than a measurement of it. Absent when the request never produced a first byte. |
| `expo.network.requests.slowest.bytesReceived` | int | Response bytes on the wire. Tells a request that was slow moving a lot of data from one that was slow while idle. Absent when nothing reported a size, which on Android means a caller abandoned the body of a response that declared no `Content-Length`. A 0 is a real measurement: an empty body, as on a 204 or a 304. |

Optional fields are omitted rather than zeroed, since a 0 would read as "instant" on a dashboard when nothing was measured.

The same distinction decides which requests form the throughput ratio. A request whose body size is unknown is left out of both sides of it, so it can't contribute duration to the denominator with no bytes to match.

# Test Plan

New unit tests across both platforms, including a chunked response the caller abandons and a body cut off mid-transfer, both driven through MockWebServer on Android.

```
$ et native-unit-tests -p ios --packages expo-app-metrics
Test Succeeded

$ et native-unit-tests -p android --packages expo-app-metrics
BUILD SUCCESSFUL
```

Swift lint, lint and format pass. Also verified against metric payloads captured from real launches on both platforms, which is where the throughput and byte-count problems turned up.

### DIFF
--- a/packages/expo-app-metrics/CHANGELOG.md
+++ b/packages/expo-app-metrics/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 - Add an optional `displayName` to `logEvent` ([#47289](https://github.com/expo/expo/pull/47289) by [@Ubax](https://github.com/Ubax))
 - Capture React render-phase errors via `AppMetricsErrorBoundary`. ([#47341](https://github.com/expo/expo/pull/47341) by [@tsapeta](https://github.com/tsapeta))
-- Attach network connection quality to metrics: `expo.network.isExpensive`, `expo.network.isConstrained` (iOS) and `expo.network.dataSaverEnabled` (Android), plus `timedOut`, `throughputBytesPerSecond` and a `slowest.*` group (`host`, `duration`, `statusCode`, `timeToFirstByte`, `bytesReceived`) describing the slowest completed request on the `expo.network.requests.*` summary. Replaces `expo.network.requests.slowestDuration` and `expo.network.requests.slowestHost`, which the group's `duration` and `host` supersede. ([#48518](https://github.com/expo/expo/pull/48518) by [@tsapeta](https://github.com/tsapeta))
+- Attach network connection quality to metrics: `expo.network.isExpensive`, `expo.network.isConstrained` (iOS) and `expo.network.dataSaverEnabled` (Android), plus `throughputBytesPerSecond` and a `slowest.*` group (`host`, `duration`, `statusCode`, `timeToFirstByte`, `bytesReceived`) describing the slowest completed request on the `expo.network.requests.*` summary. Replaces `expo.network.requests.slowestDuration` and `expo.network.requests.slowestHost`, which the group's `duration` and `host` supersede. ([#48518](https://github.com/expo/expo/pull/48518) by [@tsapeta](https://github.com/tsapeta))
 
 ### 🐛 Bug fixes
 
--- a/packages/expo-app-metrics/CHANGELOG.md
+++ b/packages/expo-app-metrics/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 - Add an optional `displayName` to `logEvent` ([#47289](https://github.com/expo/expo/pull/47289) by [@Ubax](https://github.com/Ubax))
 - Capture React render-phase errors via `AppMetricsErrorBoundary`. ([#47341](https://github.com/expo/expo/pull/47341) by [@tsapeta](https://github.com/tsapeta))
-- Attach network connection quality to metrics: `expo.network.isExpensive`, `expo.network.isConstrained` (iOS) and `expo.network.dataSaverEnabled` (Android), plus `timedOut`, `slowestTimeToFirstByte` and `throughputBytesPerSecond` on the `expo.network.requests.*` summary. ([#48518](https://github.com/expo/expo/pull/48518) by [@tsapeta](https://github.com/tsapeta))
+- Attach network connection quality to metrics: `expo.network.isExpensive`, `expo.network.isConstrained` (iOS) and `expo.network.dataSaverEnabled` (Android), plus `timedOut`, `throughputBytesPerSecond` and a `slowest.*` group (`host`, `duration`, `timeToFirstByte`, `bytesReceived`) describing the slowest completed request on the `expo.network.requests.*` summary. ([#48518](https://github.com/expo/expo/pull/48518) by [@tsapeta](https://github.com/tsapeta))
 
 ### 🐛 Bug fixes
 
--- a/packages/expo-app-metrics/CHANGELOG.md
+++ b/packages/expo-app-metrics/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 - Add an optional `displayName` to `logEvent` ([#47289](https://github.com/expo/expo/pull/47289) by [@Ubax](https://github.com/Ubax))
 - Capture React render-phase errors via `AppMetricsErrorBoundary`. ([#47341](https://github.com/expo/expo/pull/47341) by [@tsapeta](https://github.com/tsapeta))
-- Attach network connection quality to metrics: `expo.network.isExpensive`, `expo.network.isConstrained` (iOS) and `expo.network.dataSaverEnabled` (Android), plus `timedOut`, `slowestTimeToFirstByte`, `throughputBytesPerSecond` and `fastestTcpHandshake` on the `expo.network.requests.*` summary. ([#48518](https://github.com/expo/expo/pull/48518) by [@tsapeta](https://github.com/tsapeta))
+- Attach network connection quality to metrics: `expo.network.isExpensive`, `expo.network.isConstrained` (iOS) and `expo.network.dataSaverEnabled` (Android), plus `timedOut`, `slowestTimeToFirstByte` and `throughputBytesPerSecond` on the `expo.network.requests.*` summary. ([#48518](https://github.com/expo/expo/pull/48518) by [@tsapeta](https://github.com/tsapeta))
 
 ### 🐛 Bug fixes
 
--- a/packages/expo-app-metrics/CHANGELOG.md
+++ b/packages/expo-app-metrics/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 - Add an optional `displayName` to `logEvent` ([#47289](https://github.com/expo/expo/pull/47289) by [@Ubax](https://github.com/Ubax))
 - Capture React render-phase errors via `AppMetricsErrorBoundary`. ([#47341](https://github.com/expo/expo/pull/47341) by [@tsapeta](https://github.com/tsapeta))
-- Attach network connection quality to metrics: `expo.network.isExpensive`, `expo.network.isConstrained` (iOS) and `expo.network.dataSaverEnabled` (Android), plus `timedOut`, `throughputBytesPerSecond` and a `slowest.*` group (`host`, `duration`, `statusCode`, `timeToFirstByte`, `bytesReceived`) describing the slowest completed request on the `expo.network.requests.*` summary. ([#48518](https://github.com/expo/expo/pull/48518) by [@tsapeta](https://github.com/tsapeta))
+- Attach network connection quality to metrics: `expo.network.isExpensive`, `expo.network.isConstrained` (iOS) and `expo.network.dataSaverEnabled` (Android), plus `timedOut`, `throughputBytesPerSecond` and a `slowest.*` group (`host`, `duration`, `statusCode`, `timeToFirstByte`, `bytesReceived`) describing the slowest completed request on the `expo.network.requests.*` summary. Replaces `expo.network.requests.slowestDuration` and `expo.network.requests.slowestHost`, which the group's `duration` and `host` supersede. ([#48518](https://github.com/expo/expo/pull/48518) by [@tsapeta](https://github.com/tsapeta))
 
 ### 🐛 Bug fixes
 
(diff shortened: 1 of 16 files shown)


packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/networkrequests/NetworkRequestSummary.kt
```diff
@@ -80,19 +154,117 @@ data class NetworkRequestSummary(
...
- * headers arrived) counts as failed.
+ * Whether the request never produced a response, so there is no status code to report. The
+ * predicate for `slowest` and the throughput subset.
+ *
+ * Keyed on the missing status rather than on `errorDescription`, because a response can carry both:
+ * a transfer that breaks mid-body keeps the 200 its headers arrived with and gains an error string,
+ * and it waited real seconds. A 503 after eight seconds is likewise a measurement worth surfacing,
+ * whereas a timeout's duration is the client's own setting.
+ */
+internal val NetworkRequest.neverCompleted: Boolean
```
jakub-agent, line 261: [🏗️ **design**] `slowest` and `throughputBytesPerSecond` both excluding these requests is well argued, but it leaves the PR's own motivating case without a signal. A user in an elevator produces requests that never got a response, and after this change no attribute isolates them.

`failed` now folds together three different situations — never completed, a 4xx/5xx that came back, and (new on Android) a 200 whose transfer broke mid-body — so `failed: 4` can no longer be read as "4 requests never arrived". `totalDuration` carries their seconds, as its new doc says, but mixed in with every other request's.

The predicate already exists right here. Emitting a count (`expo.network.requests.neverCompleted`, or splitting `failed` into its two halves) would make the elevator case directly filterable, which is what the PR description says the goal is. Fine as a follow-up if you'd rather not grow the envelope in this PR.

packages/expo-app-metrics/ios/NetworkRequests/NetworkRequest.swift
```diff
@@ -126,6 +181,10 @@ extension NetworkRequest {
     // future need arises to surface per-hop timing, expose `metrics.transactionMetrics` here.
     let transaction = metrics?.transactionMetrics.last
 
+    // `.networkLoad` is the only fetch type that actually crossed the wire; `.localCache`,
+    // `.serverPush` and `.unknown` did not.
+    let cameFromNetwork = transaction.map { $0.resourceFetchType == .networkLoad }
```
jakub-agent: [🏗️ **design**] `.unknown` maps to `false` here, and `measurableReceiving(in:)` drops anything that is `false`, so a transaction the OS didn't classify is silently excluded from throughput. That's the opposite of how the rest of this PR treats missing information: a request with no transaction at all stays `nil` and is kept, on the grounds that absence of information isn't evidence.

It matters because `.unknown` isn't only theoretical — a task intercepted by an `NSURLProtocol` subclass (Firebase Performance, Sentry, Instabug, an SSL-proxying debugger) commonly reports it, and those SDKs are present in a fair number of the apps being measured. Throughput would then go missing for whole sessions, indistinguishable from "no measurable transfers". `transaction.map { $0.resourceFetchType != .localCache }`, or mapping `.unknown` to `nil`, keeps unknown meaning unknown and only excludes what the OS positively identified as a cache read.

Happy to be wrong about how often `.unknown` turns up in practice — did the captured payloads show it?

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/networkrequests/NetworkRequestInterceptor.kt
```diff
@@ -247,11 +256,35 @@ internal fun buildSnapshot(
...
+ * `EventListener.responseBodyEnd` is the source of truth: it fires on the network layer, below
+ * `GzipSource`, so it reports compressed bytes. It only fires once the caller drains the body,
+ * which a caller that abandons the response never does. `Content-Length` is the fallback for that
+ * case, and it's absent on a chunked response, which is how image and other streamed payloads
+ * usually arrive.
+ *
+ * Zero is a real measurement (a 204, or a 304 revalidation) and is reported as such. `null` is
+ * reserved for "nobody told us", so callers can tell an empty body from an unmeasured one.
+ */
+private fun Response.responseBodyByteCount(
```
jakub-agent, line 278: [💡 suggestion] The response side now separates "unknown" from "empty", but `requestBytesSent` just above still keeps the old shape (`phases?.requestBodyBytes?.takeIf { it > 0 } ?: originalRequest.body?.contentLength()?.takeIf { it > 0 } ?: 0L`). A streaming or chunked upload whose `RequestBody.contentLength()` returns -1 therefore reports headers-only as if it had been measured — the exact failure this function's doc argues against, on the other direction of the wire.

Lower stakes, since `bytesSent` doesn't feed the throughput ratio. But the two halves of the same snapshot now mean different things by a non-null byte count, and the next reader of `bytesSent` won't know that.

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/utils/DeviceConditions.kt
```diff
@@ -105,7 +131,20 @@ object DeviceConditions {
     val connected = capabilities?.hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET) == true
     return NetworkState(
       connected = connected,
-      transport = mapTransport(capabilities)
+      transport = mapTransport(capabilities),
+      // Gated on the same capability `connected` uses, so a network that can't carry traffic reports
+      // nothing here instead of asserting it's metered. Inverted: the capability says "not metered",
+      // the field says "expensive".
+      isExpensive = capabilities
```
jakub-agent, line 138: [💡 suggestion] This gating rule — withhold `isExpensive` unless the network also reports `NET_CAPABILITY_INTERNET`, then invert the capability — is the part most likely to regress silently, and nothing covers it. The new tests all hand-build a `NetworkState` and only exercise `MetricParamsBuilder`'s emission. Same gap on iOS for `cameFromNetwork`, where the `resourceFetchType` mapping is what decides whether a request reaches the throughput ratio at all.

A Robolectric test with a shadow `ConnectivityManager`, or extracting the capabilities→flags mapping into a pure function that takes `NetworkCapabilities`, would pin the inversion. Everything downstream of these two derivations is now tested thoroughly, which is what makes the untested ends stand out.

packages/expo-app-metrics/ios/NetworkRequests/NetworkRequest.swift
```diff
@@ -99,6 +144,16 @@ public struct NetworkRequestStarted: Sendable, Equatable, Identifiable {
   public let startedAt: Date
 }
 
+/// Adds two non-negative byte counts, pinning at `Int64.max` instead of trapping.
+///
+/// The operands come from `URLSessionTaskTransactionMetrics`, and Swift aborts the process on
+/// `Int64` overflow. No honest response reaches that total, but an observability library must not
+/// be able to take its host app down over a counter that doesn't make sense.
+private func clampedSum(_ lhs: Int64, _ rhs: Int64) -> Int64 {
```
jakub-agent: [🐧 nit] `clampedSum` here and `Int64.addingClamped` in `NetworkRequestSummary.swift` are the same function in the same module. Worth keeping one — the extension reads better at the call sites.

Neither has a test, either. Two `makeRequest(bytesReceived: .max)` through `from` would cover the pinning in a couple of lines, and this is the one part of the diff whose failure mode is a crash in someone else's app.

jakub-agent (commented):

Reviewed the full diff against the existing `expo-app-metrics` network-request pipeline on both platforms.

This is careful work. The throughput definition — the union of first-byte-to-last-byte windows over the requests that actually received bytes, with both sides of the ratio drawn from one subset — is the right shape, and the reasons it can't be a naive `bytes / totalDuration` are written down where the next reader will find them. Keeping Android's Data Saver on its own key instead of collapsing it onto iOS's `isConstrained`, omitting optional fields rather than zeroing them, and the `measuredResponseEnd` / `responseEnd` split are all decisions I'd have argued for. The mid-body-failure fix (`callFailed` → `phases.failure`) is a real bug caught on the way through, and the MockWebServer tests for the truncated and abandoned bodies exercise the paths that actually break.

No blocking issues. The findings are one design gap, one platform edge case, and some consistency and coverage points:

- **The elevator case has no dedicated signal.** `slowest` and throughput now exclude requests that never completed, and `failed` folds together three different failures, so nothing isolates "N requests never got a response" — the scenario the PR opens with.
- **iOS treats an `.unknown` fetch type as "not from the network"**, dropping those requests from throughput, while a missing transaction is treated as unknown-and-kept. Worth confirming against payloads from apps carrying an `NSURLProtocol`-based SDK.
- `requestBytesSent` on Android didn't get the same unknown-vs-empty treatment as the response side.
- The two OS-facing derivations (`networkState` capability flags, `cameFromNetwork`) are the only new logic without tests.
- Two copies of the `Int64` overflow clamp on iOS, and a few defaults on Android that don't earn their place.

One thing I can't check from here: `expo.network.requests.slowestDuration` and `slowestHost` are removed, and they ship in the 58 canaries. Nothing else in this repo reads them, but if anything on the ingestion or dashboard side does, it needs the new `slowest.*` names.

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/networkrequests/NetworkRequestSummary.kt
Ubax: Up to you, but probably this can be a follow-up

packages/expo-app-metrics/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 - Add an optional `displayName` to `logEvent` ([#47289](https://github.com/expo/expo/pull/47289) by [@Ubax](https://github.com/Ubax))
 - Capture React render-phase errors via `AppMetricsErrorBoundary`. ([#47341](https://github.com/expo/expo/pull/47341) by [@tsapeta](https://github.com/tsapeta))
+- Attach network connection quality to metrics: `expo.network.isExpensive`, `expo.network.isConstrained` (iOS) and `expo.network.dataSaverEnabled` (Android), plus `throughputBytesPerSecond` and a `slowest.*` group (`host`, `duration`, `statusCode`, `timeToFirstByte`, `bytesReceived`) describing the slowest completed request on the `expo.network.requests.*` summary. Replaces `expo.network.requests.slowestDuration` and `expo.network.requests.slowestHost`, which the group's `duration` and `host` supersede. ([#48518](https://github.com/expo/expo/pull/48518) by [@tsapeta](https://github.com/tsapeta))
```
Ubax: Isn't it a bit too verbose?

Ubax (approved):
Please check the AI generated comments, before merging

PR #48521, opened 2026-08-04
[ui][Android] Add Jetpack Compose `Image` component

jakex7:
# Why

Similar to what i did a while ago, expo-widgets rely on expo-ui's components, so we need a basic component for displaying an image.

# How

Similar implementation to `IconView`.
Moved common loader/helpers to /graphics instead of relying on Icon

# Test Plan

NCL and tests

image

# Checklist

---


docs/pages/versions/unversioned/sdk/ui/jetpack-compose/image.mdx
```diff
@@ -0,0 +1,44 @@
...
+}
+```
+
+## API
+
+```tsx
+import { Image } from '@expo/ui/jetpack-compose';
+```
+
+<APISection packageName="expo-ui/jetpack-compose/image" apiName="Image" />
```
intergalacticspacehighway, line 53: Also need to add `expo-ui/jetpack-compose/image` key to `tools/src/commands/GenerateDocsAPIData.ts` and backport changes to 57

docs/pages/versions/unversioned/sdk/ui/jetpack-compose/image.mdx
```diff
@@ -0,0 +1,44 @@
...
+description: A Jetpack Compose component for displaying images.
+sourceCodeUrl: 'https://github.com/expo/expo/tree/main/packages/expo-ui'
+packageName: '@expo/ui'
+platforms: ['android', 'expo-go']
+---
+
+import APISection from '~/components/plugins/APISection';
+import { APIInstallSection } from '~/components/plugins/InstallSection';
+
+`Image` renders image sources using Jetpack Compose.
```
intergalacticspacehighway:
```suggestion
Expo UI Image wraps the official Jetpack Compose [Image](https://developer.android.com/develop/ui/compose/graphics/images/loading) API and displays a bundled or remote image. Set the size with modifiers such as `size`, `width`, or `fillMaxWidth`, then `contentScale` to control how the image fills those bounds.
```

To make it consistent with other component docs

intergalacticspacehighway:
A few items came up in the AI review 😅. We can also look at them in a follow-up.

  ### 1. Images decode at full resolution

  https://github.com/expo/expo/blob/707f096a178ecb9ec0191b8b15f22984d0e9f288/packages/expo-ui/android/src/main/java/expo/modules/ui/graphics/ImageLoader.kt#L164-L174

  `createFromStream` gets no `BitmapFactory.Options`, so there's no subsampling. Same at `L151` for bundled resources.

  | Source | Decoded | Memory (ARGB_8888) |
  | --- | --- | --- |
  | 4000×3000 photo | 4000×3000 | ~48 MB |
  | …in a 240×120 box | still 4000×3000 | ~48 MB |

  `ImageSource` already carries `width`, `height`, and `scale` from `resolveAssetSource` — native never reads them. A two-pass decode fixes it: the 
  first pass with `inJustDecodeBounds = true` allocates no pixels and only reads dimensions, then `inSampleSize` bounds the real decode.

  > [!WARNING]
  > `inSampleSize` shrinks the bitmap, which shrinks `drawable.intrinsicWidth`, which shrinks the laid-out image. Any `Image` without a size modifier 
  would silently render smaller. Report the source's declared size to layout instead of the decoded bitmap's.

  ### 2. Nothing is cached, at any layer

  https://github.com/expo/expo/blob/707f096a178ecb9ec0191b8b15f22984d0e9f288/packages/expo-ui/android/src/main/java/expo/modules/ui/ExpoUIModule.kt#L63-L72

  No `Cache` on the client, and no bitmap cache anywhere. `okHttpClient?.cache?.close()` in `OnDestroy` is always null — that dead line suggests a 
  cache was intended.

  So the demo's own pattern (Images in a `LazyColumn`) re-downloads *and* re-decodes on every scroll pass.

  - **Disk:** `OkHttpClientProvider.createClientBuilder(context).build()` is a one-liner — 10 MB disk cache, RN's cookie jar, plus any app-installed 
  interceptors. `expo-file-system`, `expo-asset`, and `expo-dev-launcher` all go through it. Note `getOkHttpClient()` is *not* the one to use; it has 
  no cache.
  - **Memory:** the HTTP cache still leaves you re-decoding, and does nothing for bundled `res:` assets. A small bitmap `LruCache` on `ExpoUIModule`, 
  cleared in the same `OnDestroy`, is what actually fixes the list case.

  ### 3. In-flight loads are never cancelled

  https://github.com/expo/expo/blob/707f096a178ecb9ec0191b8b15f22984d0e9f288/packages/expo-ui/android/src/main/java/expo/modules/ui/graphics/ImageLoader.kt#L102-L121

  The blocking `execute()` ignores coroutine cancellation. Cancelling only the header fetch wouldn't be enough either — the caller reads the body 
  afterwards via `readBytes()` (`L165`), so the download is still live after `downloadFromHttp` returns and the `Call` reference is gone. The call 
  needs to stay tied to the enclosing job for the whole load.

  Separately, `catch (e: Exception)` at `L68` swallows `CancellationException` (it extends `IllegalStateException` on the JVM). A cancelled load 
  resolves to an empty `ImageResult` and the coroutine completes normally, which breaks `LaunchedEffect` restart semantics.

  <details>
  <summary>Or step back and use a real pipeline</summary>

  Subsampling, caching, cancellation, animated images, and placeholder/error states are all things Coil gives you, and it has a Compose `AsyncImage` 
  that drops in. [Google's own loading guide](https://developer.android.com/develop/ui/compose/graphics/images/loading) recommends it, and `expo-image`
  already ships Glide — so a real pipeline isn't without precedent here. Worth deciding before the hand-rolled loader grows more of these.

  </details>

docs/pages/versions/unversioned/sdk/ui/jetpack-compose/image.mdx
amandeepmittal, line 53: ⚠️ **correctness** `mdx-components-sdk.md#apisection`

`docs/public/static/data/unversioned/expo-ui/jetpack-compose/image.json` is missing from this PR, so this whole section renders "No API data file found, sorry!" instead of the props table. Add `'expo-ui/jetpack-compose/image': ['jetpack-compose/Image/index.tsx', 'expo-ui']` to the package map in `tools/src/commands/GenerateDocsAPIData.ts`, then run `et generate-docs-api-data --packageName expo-ui/jetpack-compose/image` from the monorepo root and commit the generated JSON.

docs/pages/versions/unversioned/sdk/ui/jetpack-compose/image.mdx
```diff
@@ -0,0 +1,44 @@
+---
+title: Image
+description: A Jetpack Compose component for displaying images.
+sourceCodeUrl: 'https://github.com/expo/expo/tree/main/packages/expo-ui'
+packageName: '@expo/ui'
+platforms: ['android', 'expo-go']
+---
+
+import APISection from '~/components/plugins/APISection';
+import { APIInstallSection } from '~/components/plugins/InstallSection';
```
amandeepmittal, line 10: 💡 **nit** `mdx-components.md#contentspotlight`

Every other visual component page in this section previews the component under the intro paragraph, so the screenshot from the PR description would fit here as a `<ContentSpotlight variant="component" aspect="landscape">` light/dark pair under **public/static/images/expo-ui/image/**.

docs/pages/versions/unversioned/sdk/ui/index.mdx
```diff
@@ -90,6 +90,7 @@ See **[Drop-in replacements](drop-in-replacements)** for API-compatible replacem
 | [`Host`](jetpack-compose/host)                                           | Host component for bridging React Native and Jetpack Compose.                                                    |
 | [`Icon`](jetpack-compose/icon)                                           | Icon component for displaying icons.                                                                             |
 | [`IconButton`](jetpack-compose/iconbutton)                               | IconButton components for displaying native Material3 icon buttons.                                              |
+| [`Image`](jetpack-compose/image)                                         | Image component for displaying images.                                                                           |
```
amandeepmittal, line 93: ⚠️ **correctness** `mdx-components-sdk.md#apisection`

This table is generated from page frontmatter, so a hand-written row gets overwritten and the matching row in **jetpack-compose/index.mdx** is still missing. Run `pnpm generate-ui-component-tables` from **docs** to update both pages. It strips the leading "A Jetpack Compose " from the page description, so change that description to "A Jetpack Compose Image component for displaying images." first, or the row reads "Component for displaying images." without the component name.

amandeepmittal (changes requested):

docs/pages/versions/unversioned/sdk/ui/jetpack-compose/image.mdx
jakex7: Yup, wanted to wait for the final review before generating JSON.

docs/pages/versions/unversioned/sdk/ui/jetpack-compose/image.mdx
jakex7: I'm not sure showing a random image would be beneficial in this case, but if you'd like, I can add one.

docs/pages/versions/unversioned/sdk/ui/index.mdx
jakex7: Didn't know that, thanks!

docs/pages/versions/unversioned/sdk/ui/jetpack-compose/image.mdx
amandeepmittal: I think showing something here would add that visual element that is often missing from our docs.

jakex7: Good, I added the ContentSpotlight
image

amandeepmittal (approved):

packages/expo-ui/android/src/main/java/expo/modules/ui/image/ImageView.kt
```diff
@@ -0,0 +1,101 @@
...
+    val (contentScale) = props.contentScale
+    val (contentDescription) = props.contentDescription
+    val (tint) = props.tint
+    val (alpha) = props.alpha
+    val (modifiers) = props.modifiers
+
+    var loadedImage by remember(source) { mutableStateOf(ImageLoader.ImageResult()) }
+
+    LaunchedEffect(source) {
+      loadedImage = source?.resolveUri(context)?.let { imageLoader.loadFromUri(it) }
```
Kudo, line 79: [🏗️ **design**] Reusing the icon loader means photos decode at full resolution with no cache. `ImageLoader.parseImageFromStream` does `inputStream.readBytes()` then `Drawable.createFromStream(...)`, so a 12MP JPEG becomes a ~48 MB ARGB_8888 bitmap even inside `size(240, 120)`. `ExpoUIModule` also builds its `OkHttpClient` with no `Cache`, and there is no bitmap cache, so two `<Image>`s pointing at the same remote URI download and decode twice.

That was fine when the loader only handled small XML vector drawables. For a component whose docs advertise photos and remote URIs it is an OOM risk. Could the bitmap path downsample with `BitmapFactory.Options.inSampleSize` against the target bounds, or hand decoding off to Coil?

packages/expo-ui/src/jetpack-compose/Image/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+import {
+  Image as ReactNativeImage,
+  type ColorValue,
+  type ImageResolvedAssetSource,
+} from 'react-native';
+
+import { type PrimitiveBaseProps, transformProps } from '../layout-types';
+
+/** Controls how an image is scaled inside its bounds. */
+export type ImageContentScale = 'fit' | 'crop' | 'fillBounds';
```
Kudo, line 11: [💡 suggestion] Compose's `ContentScale` has seven values; this exposes three. `inside` (fit, but never upscale) and `none` have no workaround from JS, and `packages/expo-ui/CLAUDE.md` asks to "expose all configurable properties of the underlying native component as props". Adding `fillWidth`, `fillHeight`, `inside` and `none` here plus matching entries in the Kotlin `ImageContentScale` enum would close it.

Same question for Compose `Image`'s `alignment` parameter, which isn't exposed at all — it decides where the image sits when `contentScale` leaves empty space.

packages/expo-ui/android/src/main/java/expo/modules/ui/image/ImageView.kt
```diff
@@ -0,0 +1,101 @@
...
+        ?: ImageLoader.ImageResult()
+    }
+
+    val painter = loadedImage.imageVector?.let { rememberVectorPainter(it) }
+      ?: rememberDrawablePainter(loadedImage.drawable)
+      ?: remember { ColorPainter(androidx.compose.ui.graphics.Color.Transparent) }
```
Kudo, line 85: [💡 suggestion] Every failure path renders an invisible image with no signal to JS: unresolvable asset id (`resolveAssetSource` returns `null`), unsupported scheme, HTTP error, decode failure. A `data:` URI is a concrete case — it falls into the `else` branch of `getInputStreamForUri`, `contentResolver.openInputStream` throws, and the user sees blank space with only a logcat line.

Worth exposing `onLoad`/`onError` via the `ViewEvent` pattern, or at minimum documenting in the TSDoc that unsupported or failed sources render nothing?

docs/pages/versions/unversioned/sdk/ui/jetpack-compose/image.mdx
```diff
@@ -0,0 +1,53 @@
...
+import { ContentSpotlight } from '~/ui/components/ContentSpotlight';
+
+Expo UI Image wraps the official Jetpack Compose [Image](https://developer.android.com/develop/ui/compose/graphics/images) API and displays a bundled or remote image. Set the size with modifiers such as `size`, `width`, or `fillMaxWidth`, then `contentScale` to control how the image fills those bounds.
+
+<ContentSpotlight
+  variant="component"
+  aspect="landscape"
+  src="/static/images/expo-ui/image/android-light.webp"
+  darkSrc="/static/images/expo-ui/image/android-dark.webp"
+  alt="Icons rendered with different contentScale props"
```
Kudo, line 20: [🐧 nit] Alt text says "Icons" — looks copy-pasted from `icon.mdx`.

```suggestion
  alt="Images rendered with different contentScale props"
```

Kudo (changes requested):
left some ai code review comment. i think the cache one is something we should considered.
note that the VectorIconLoader doesn't support full android VectorDrawable. it supports only `<path>` which all matetial symbols use path only. if you intent to support full VectorDrawable for Image, that's something to think more

jakex7:
@Kudo 
I implemented basic caching, exposed all `ContentScale` and `alignment` props, and added `onLoad`/`onError` callbacks.

Please take a look when you have a chance, and If there are no significant changes needed, I’d like to merge this and continue working on the widgets (that's why i needed expo-ui's Image). We can iterate on and improve expo-ui's image in follow-up prs. Would that work for you?

jakex7:
> * New OkHttp client opens a second disk cache on a directory another cache already owns
> * Debug console.log ships in the published Image component
> * New Image documentation page is not linked from the component list

Fixed in https://github.com/expo/expo/pull/48521/commits/0bcdd1581b962d3cee05153becd4e2ee166d7abf

packages/expo-ui/android/src/main/java/expo/modules/ui/graphics/ImageLoader.kt
```diff
@@ -1,135 +1,154 @@
...
+  private suspend fun downloadFromHttp(url: String): ByteArray {
+    val request = Request.Builder()
+      .url(url)
+      .build()
+    return okHttpClient.newCall(request).await().use { response ->
       if (!response.isSuccessful) {
-        Log.e(TAG, "Failed to download icon from $url: ${response.code}")
-        response.close()
-        return null
+        throw IOException("Failed to download image from $url: HTTP ${response.code}")
```
Kudo, line 135:
```suggestion
        response.close()
        throw IOException("Failed to download image from $url: HTTP ${response.code}")
```

should we keep the closing here?

Kudo (approved):

packages/expo-ui/android/src/main/java/expo/modules/ui/graphics/ImageLoader.kt
jakex7: No, `use {}` closes it automatically.

> https://kotlinlang.org/docs/exceptions.html#the-finally-block
> In Kotlin, the idiomatic way to manage resources that implement the [AutoClosable](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-auto-closeable/) interface, such as file streams like FileInputStream or FileOutputStream, is to use the [.use()](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/use.html) function. This function automatically closes the resource when the block of code completes, regardless of whether an exception is thrown

Kudo: oh right! thanks for correcting me that

PR #48525, opened 2026-08-05
[cli] Fix iOS asset catalog scale-to-file pairing for assets with non-standard scales

janicduplessis:
# Why

`persistMetroAssetsAsync` filters an asset's scales through `filterPlatformAssetScales`, but `getImageSet` indexes the unfiltered `asset.files`, so an asset with scales `[1, 1.5, 2, 3]` gets the 1.5x file in its 2x catalog slot and the 2x file in its 3x slot: the compiled `Assets.car` ships the wrong images. Assets with *only* non-standard scales get an imageset actool silently drops entirely, making them unloadable since the catalog runtime has no filesystem fallback.

Same bug as facebook/react-native#57825, merged as [`74b98472`](https://github.com/facebook/react-native/commit/74b984729b3e21225d1c54f7671644c37bbeb026) — this code mirrors the RN community CLI implementation (reachable once `RCTUseAssetCatalog` is on, RN 0.88+). This PR carries the same logic, so the two implementations stay in step.

# How

- Pair each catalog slot (1x/2x/3x) with its own file.
- An asset with no valid scale maps its closest variant into the nearest valid slot (the "closest larger" rule loose files already get) and warns, so every imageset holds at least one rendition actool will compile.

# Test Plan

Added unit tests to `persistMetroAssets.test.ts` for `getCatalogImages`: standard pairing, mixed `[1, 1.5, 2, 3]` (regression for the file shift), fractional-only `[1.5]`, `[4]` clamping to the 3x slot, and `[1.5, 2.5]` taking the largest fractional variant. The cases match the ones that landed with the RN fix. The file's suite passes locally, 10/10.

# Checklist

End-to-end validation isn't possible here yet (the catalog path only becomes reachable once an SDK targets RN 0.88) — see facebook/react-native#57825 for the full e2e of the identical RN fix: compiled-car verification and on-device before/after screenshots.

### DIFF
--- a/packages/@expo/cli/CHANGELOG.md
+++ b/packages/@expo/cli/CHANGELOG.md
@@ -27,6 +27,7 @@
 - Serve relative manifest URLs only when the client itself sends the RFC 7239 `Forwarded` header, so that proxied requests from clients without relative-URL support, like released Expo Go versions through the WS tunnel, keep absolute URLs. ([#48997](https://github.com/expo/expo/pull/48997) by [@expo-bot](https://github.com/expo-bot))
 - Fail when `--private-key-path` is passed without `updates.codeSigningCertificate` in the resolved app config, instead of ignoring the flag and continuing without signing.
 - Show the Xcode build log path when `run:ios` fails. ([#48624](https://github.com/expo/expo/pull/48624) by [@ramonclaudio](https://github.com/ramonclaudio))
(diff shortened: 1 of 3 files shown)


janicduplessis:
RN PR has been merged

cc @kitten @byCedric

Kudo (changes requested):
the code looks correct but any chance we can prevent `!` in some changes, e.g. `asset.files[idx]!` and `ALLOWED_SCALES.ios!`?

janicduplessis:
@Kudo added undefined checks instead of non-null assertions

Kudo (approved):
thanks for following up. great work!

PR #47079, opened 2026-06-19
[ios][audio] Fix app hang from main-thread AVAudioSession deactivation

sbs44:
# Why

On iOS, `AudioModule.deactivateSession()` ran the synchronous, blocking `AVAudioSession.setActive(false, …)` on the **main thread**. That call makes an XPC round-trip to the audio daemon that can stall for several seconds under system contention, freezing the UI and surfacing as an iOS app hang ("App Hang Fully Blocked").

It fires from `AudioPlayer`'s `onPlaybackComplete` and from `pause`, so apps that play short clips back-to-back (turn-by-turn navigation TTS, UI/game sound effects, message read-aloud) hit it constantly with the default `keepAudioSessionActive: false`.

Fixes #47045 — the symbolicated main-thread stack and a minimal reproduction are in the issue. Representative stack:

```
mach_msg → xpc_connection_send_message_with_reply_sync
  → __NSXPCCONNECTION_IS_WAITING_FOR_A_SYNCHRONOUS_REPLY__
    → -[AVAudioSession privateSetActive:withOptions:error:core:]
      → closure in AudioModule.deactivateSession (AudioModule.swift)
        → __CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__
```

# How

- **Off-main:** `setActive(false)` now runs on a dedicated serial `sessionQueue`. The `isPlaying`/registry read stays on the existing 100 ms main-thread debounce... safe, since `AudioComponentRegistry` is already internally synchronized.
- **Supersede stale deactivations:** once deactivation is async, a deactivation queued when clip A ends can race clip B's `setActive(true)` and tear B down. An `NSLock`-guarded monotonic token is bumped at every activation site... `activateSession()` (and `handleInterruptionEnded`/`reconfigureAudioSession`, now routed through it) plus `setIsAudioActive(true)`... captured at request time and re-checked before the blocking call... a reactivation skips the stale deactivation.

The re-check and `setActive(false)` aren't atomic, leaving a narrow TOCTOU window (far smaller than the prior untokened main-thread race) that iOS's `isBusy` covers; noted in-code. No public API change; iOS only.

# Test Plan

Repro, steps, and environment are in #47045: a blank app playing a short clip back-to-back with the default player intermittently hangs the main thread in `AudioModule.deactivateSession` → `setActive(false)` (Xcode Organizer → Hangs / MetricKit). Originated from production hang reports.

Ran locally, scoped to `expo-audio` (no monorepo build):

```
[...]

Issue #47045, opened 2026-06-18
[expo-audio] iOS: AudioModule.deactivateSession blocks the main thread, causing app hangs


sbs44:
### Minimal reproducible example

https://github.com/sbs44/expo-audio-test

### Steps to reproduce

1. Run the repro on a physical device.
2. Play a stream of short clips — e.g. trigger `speak()` repeatedly so a new clip starts shortly after the previous one finishes (turn-by-turn TTS, UI/game sound effects, and read-aloud all exhibit this).
3. Do it under some system load (another audio app active, device warm) so the audio daemon is slow to answer.

**Expected:** deactivating the audio session never blocks the main thread; playback stays responsive.

**Actual:** the main thread intermittently freezes for ~3–6 s and the OS reports an app hang (visible in Xcode Organizer → Hangs, MetricKit `MXHangDiagnostic`, or any app-hang reporter). Representative symbolicated main-thread stack:

```
libsystem_kernel.dylib   mach_msg2_trap
libsystem_kernel.dylib   mach_msg
libdispatch.dylib        _dispatch_mach_send_and_wait_for_reply
libxpc.dylib             xpc_connection_send_message_with_reply_sync
Foundation               __NSXPCCONNECTION_IS_WAITING_FOR_A_SYNCHRONOUS_REPLY__
AudioSession             -[AVAudioSession privateSetActive:withOptions:error:core:]
ExpoAudio                closure in AudioModule.deactivateSession  (AudioModule.swift)
ExpoAudio                thunk for closure
libdispatch.dylib        _dispatch_main_queue_callback_4CF
CoreFoundation           __CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__
CoreFoundation           __CFRunLoopRun
GraphicsServices         GSEventRunModal
UIKitCore                -[UIApplication _run]
UIKitCore                UIApplicationMain
```

### Environment

```text
expo-env-info 2.0.14 environment info:
    System:
      OS: macOS 26.5.1
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 24.17.0 - /opt/homebrew/opt/node@24/bin/node
      Yarn: 1.22.22 - /opt/homebrew/bin/yarn
      npm: 11.13.0 - /opt/homebrew/opt/node@24/bin/npm
    Managers:
      CocoaPods: 1.16.2 - /opt/homebrew/bin/pod
    SDKs:
      iOS SDK:
[...]

PR #48532, opened 2026-08-05
[audio][ios] Run AVAudioSession.setActive on a serial queue off the calling thread

### LINKED ISSUE #48531: [expo-audio] iOS: sync `play()` runs blocking `AVAudioSession.setActive(true)` on the JS thread – an audio-daemon stall freezes all JS execution

### Steps to reproduce

### Environment

```text
    System:
      OS: macOS 26.6
      Shell: 5.9 - /bin/zsh
    Binaries:
    Managers:
    SDKs:
      iOS SDK:
    IDEs:
    npmPackages:
      expo: ^57.0.0 => 57.0.10
      react: 19.2.3 => 19.2.3
    Expo Workflow: managed
```

### Expo Doctor Diagnostics

```text
```

sbs44:
# Why

Fixes #48531 – the activation counterpart to #47045/#47066.

`AVAudioSession.setActive` is a synchronous XPC round-trip to the audio daemon that can stall for seconds, or indefinitely, under system contention. Both sync `Function("play")` definitions performed the activation inline... and expo-modules sync Functions execute on the JS runtime thread (`SyncFunctionDefinition` is `@JavaScriptActor`)... so a stalled activation froze **all JS execution**: every timer, state update, and event handler, while natively driven UI kept running and the app looked alive. No crash, no JS error, and no app-hang report (watchdogs monitor the main thread, which is idle). #47066 fixed the deactivation half of this; the activation half remained on the JS thread.

Production evidence and a measurable repro (sync `play()` means the call's wall-clock duration *is* the JS-thread block time) are in #48531.

# How

- **All of AudioModule's session transitions run on one serial `sessionQueue`.** `play()` splits into `prepareToPlay()` (observer registration and loop bookkeeping... kept on the calling thread, where the player's other mutations run) and `startPlayback()` (AVPlayer-only, deferred behind the activation on the queue).
- **A `MonotonicGeneration` token supersedes stale deactivations** so back-to-back playback isn't torn down by a deferred `setActive(false)` (same token idea as #47079, now shared by one small type).
- **A per-playable play generation keeps a queued play from starting cancelled audio.** Bumps live inside the playables' own `pause()`/replace/`skipTo`/`clear`/teardown methods, so every superseding path (module Functions, `pauseAllPlayers` on backgrounding, interruptions, `remove`) is covered by construction. `AudioPlaylist` gets the same protection as `AudioPlayer`.
- **Interruption bookkeeping moves onto `sessionQueue`.** `interruptedPlayers`/`playerVolumes` were mutated on the notification thread while the (now queued) resume path iterated them... a Swift collection race.
- **`setIsAudioActiveAsync` becomes promise-based** and settles from the queue, so a stalled transition no longer blocks the process-wide AsyncFunction queue that all Expo modules share. JS API unchanged.
- **Behavior change:** `play()` no longer throws when activation fails... playback is still attempted (failures are frequently transient) and the failure is surfaced through the `playbackStatusUpdate` event's existing `error` field. Noted in the changelog.
- Write-only `sessionIsActive` removed.

Known gap, deliberately out of scope: `AudioRecorder`/`AudioStream` still call `setActive(true)` inline outside the queue; noted in-code for a follow-up.

# Test Plan

- `ios/Tests/MonotonicGenerationTests.swift` (Swift Testing) covers the supersession orderings... play-then-pause must skip the queued play, deactivate-then-activate must skip the stale deactivation, pause-then-play must still play... plus lossless concurrent bumps.
- Behavioral verification is on-device with the repro in #48531 (https://github.com/sbs44/expo-audio-test): the sync `play()` call duration displays the JS-thread block per play; with the repro's simulated-stall patch, stock expo-audio freezes the JS tick counter for 4 s per clip while a native animation keeps running... with this branch, the JS tick never stops and audio starts when the (simulated) daemon recovers. Pause during the stall window and the queued play stays silent.
- A backport of this change to expo-audio 57.0.3 compiles clean (pod install + xcodebuild of the ExpoAudio target) and is shipping in a production navigation app via patch-package.
- Swift-only diff... no JS/TS or committed `build/` changes. Package lint and native tests run in CI (no local expo monorepo toolchain was available for `expo-module lint`/`test`).

# Checklist

### DIFF
--- a/packages/expo-audio/CHANGELOG.md
+++ b/packages/expo-audio/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 ### 🐛 Bug fixes
 
+- [iOS] Run `AVAudioSession.setActive` calls on a serial queue instead of the calling thread. The synchronous `play()` performed the blocking activation XPC on the JS thread, so an audio-daemon stall froze all JS execution (timers, state updates) while the rest of the app appeared alive; deactivation had the same hazard on its calling thread. A monotonic activation token drops stale deactivations so rapid back-to-back playback isn't torn down, and a per-playable play generation prevents a queued play from audibly starting audio that `pause()`/`replace()` already cancelled. Behavior change: `play()` no longer throws when session activation fails — playback is still attempted and the failure is reported through the `playbackStatusUpdate` event's `error` field. ([#48532](https://github.com/expo/expo/pull/48532) by [@sbs44](https://github.com/sbs44))
 - [Android] Pause audio players and playlists when headphones or Bluetooth audio devices disconnect. ([#48151](https://github.com/expo/expo/pull/48151) by [@vivekjm](https://github.com/vivekjm))
 - [Android] Give the lock-screen `MediaSession` instances a unique ID so concurrent active players (and the basic session) no longer collide on the empty default. ([#47101](https://github.com/expo/expo/issues/47101) by [@tsushanth](https://github.com/tsushanth))
 - [Android] Fix stale lock screen artwork when updating metadata without an `artworkUrl`. ([#45738](https://github.com/expo/expo/pull/45738) by [@behenate](https://github.com/behenate))
--- a/packages/expo-audio/ios/AudioModule.swift
+++ b/packages/expo-audio/ios/AudioModule.swift
@@ -1,7 +1,6 @@
 import ExpoModulesCore
 
 public class AudioModule: Module {
-  private var sessionIsActive = true
   private let registry = AudioComponentRegistry()
 
   // MARK: Properties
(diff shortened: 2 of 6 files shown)


brentvatne:
Closing — `AVAudioSession.setActive` was moved off the JS thread in https://github.com/expo/expo/pull/48588 (issue #48531). Thanks!

sbs44:
Hey Brent – Makes sense... #48588 is a simpler design than what I had here. It does miss
three things from this PR though, one of them a new race in the interruption
path. Rebased those onto the merged approach in #48913, details in a comment
on #48588.

PR #48538, opened 2026-08-05
[cli][metro-config] Update internal and external callsite handling

krystofwoldrich:
## Summary

- Update `INTERNAL_CALLSITES_REGEX` from the latest React Native
- Add `external` flag (meaning from node_modules, a.k.a not app/users code)
	- externals are showed dimmed
- Build errors continue to show all frames so transformation issues are still debuggable

## Why

The `node_modules` regex reduced the stack frames to app only which is too short.

## Test plan

- updated/added unit tests


packages/@expo/cli/src/start/server/metro/formatStack.ts
```diff
@@ -0,0 +1,207 @@
+/**
```
kitten, line 1: fwiw we don't need these headers on new files. LLMs just tend to copy them blindly

packages/@expo/cli/src/start/server/metro/formatStack.ts
```diff
@@ -0,0 +1,207 @@
...
+export function likelyContainsCodeFrame(message: string | undefined): boolean {
+  if (!message) return false;
+
+  const clean = stripVTControlCharacters(message);
+  if (!clean) return false;
+
+  return /^\s*\d+\s+\|/m.test(clean);
+}
+
+export function formatStack(
```
kitten, line 72: side-note: the stack handling is starting to worry me. Last time I checked attribution of code to usage is extremely tricky and we're doing lots of stack handling/parsing/formatting over several places

I have a feeling that that's become vaguely unmanageable and am mostly concerned we're building a house of cards here that'll collapse on us eventually

extracting this is nice, fwiw, removing more variants of string-based stack handling or having more of this grouped and well tested would be even nicer going forward

packages/@expo/cli/src/start/server/metro/formatStack.ts
```diff
@@ -0,0 +1,207 @@
...
+      const position = terminalLink.isSupported
+        ? terminalLink(frame.subtitle, frame.subtitle)
+        : frame.subtitle;
+      let lineItem = chalk.gray(`  ${frame.title} (${position})`);
+
+      if (frame.collapse || frame.external) {
+        lineItem = chalk.dim(lineItem);
+      }
+      // Never show the internal module system.
+      const isMetroRuntime =
```
kitten, line 177: I think we incidentally match the Expo naming here too, but this feels like it's not an exact match, which it maybe could be if we have access to the config or serialiser details?

packages/@expo/metro-config/src/customizeFrame.ts
```diff
@@ -10,35 +10,44 @@ type CustomizeFrameFunc = SymbolicatorConfigT['customizeFrame'];
...
+    '/Libraries/Utilities/.+\\.js$',
+    '/Libraries/vendor/.+\\.js$',
+    '/Libraries/WebSocket/.+\\.js$',
+    '/src/private/renderer/errorhandling/.+\\.js$',
+    '/metro-runtime/.+\\.js$',
+    '/node_modules/@babel/runtime/.+\\.js$',
+    '/node_modules/@react-native/js-polyfills/.+\\.js$',
+    '/node_modules/invariant/.+\\.js$',
+    '/node_modules/react-devtools-core/.+\\.js$',
+    '/node_modules/react-native/index.js$',
```
kitten, line 30: This seems pointless?

packages/@expo/metro-config/src/customizeFrame.ts
```diff
@@ -10,35 +10,44 @@ type CustomizeFrameFunc = SymbolicatorConfigT['customizeFrame'];
...
+    // https://github.com/react/react-native/blob/f2a250ad895ffd4478f0668f56867b78be21ea2f/packages/metro-config/src/index.flow.js#L20-L48
     '/Libraries/BatchedBridge/MessageQueue\\.js$',
-    '/Libraries/YellowBox/.+\\.js$',
+    '/Libraries/Core/.+\\.js$',
     '/Libraries/LogBox/.+\\.js$',
-    '/Libraries/Core/Timers/.+\\.js$',
-    'node_modules/react-devtools-core/.+\\.js$',
-    'node_modules/react-refresh/.+\\.js$',
-    'node_modules/scheduler/.+\\.js$',
+    '/Libraries/Network/.+\\.js$',
```
kitten, line 18: I'm not sure if we want to keep this or the one below?

packages/@expo/metro-config/src/customizeFrame.ts
```diff
@@ -10,35 +10,44 @@ type CustomizeFrameFunc = SymbolicatorConfigT['customizeFrame'];
...
+    '/Libraries/WebSocket/.+\\.js$',
+    '/src/private/renderer/errorhandling/.+\\.js$',
+    '/metro-runtime/.+\\.js$',
+    '/node_modules/@babel/runtime/.+\\.js$',
+    '/node_modules/@react-native/js-polyfills/.+\\.js$',
+    '/node_modules/invariant/.+\\.js$',
+    '/node_modules/react-devtools-core/.+\\.js$',
+    '/node_modules/react-native/index.js$',
+    '/node_modules/react-refresh/.+\\.js$',
+    '/node_modules/scheduler/.+\\.js$',
```
kitten, line 32: This should be removed. It's unlikely we'd want to hide this as it could make scheduled tasks miss frames that indicate they've been delayed by a micro/macro tick imho

packages/@expo/cli/src/start/server/metro/formatStack.ts
```diff
@@ -0,0 +1,207 @@
...
+ * Tests given string on presence of ` [num] |` at the start of any line.
+ * Returns `false` for undefined or empty strings.
+ */
+export function likelyContainsCodeFrame(message: string | undefined): boolean {
+  if (!message) return false;
+
+  const clean = stripVTControlCharacters(message);
+  if (!clean) return false;
+
+  return /^\s*\d+\s+\|/m.test(clean);
```
kitten, line 69: nit: I'd do explicit space matching here rather than `\s` or rather limit to tabs/spaces as needed to narrow this a bit

kitten (approved):

PR #48546, opened 2026-08-05
[video][Android] Use transient audio focus for the auto mixing mode

huextrat:
# Why

On Android, `expo-video` requests permanent audio focus (`AUDIOFOCUS_GAIN`) for the `auto` audio mixing mode. Although `AudioFocusManager` already abandons focus once no player needs it, abandoning a permanent request does not reliably make interrupted music or podcast apps resume.

Repro: play music in Spotify → play an unmuted video with `audioMixingMode: "auto"` → pause, mute, or unmount the last audible video player. Spotify remains silent instead of resuming.

# How

Request `AUDIOFOCUS_GAIN_TRANSIENT` for the `auto` audio mixing mode. An in-app video temporarily interrupts background audio, and abandoning a transient request allows the previous focus owner to resume automatically.

`doNotMix` continues to request permanent `AUDIOFOCUS_GAIN`, since it explicitly opts into taking over audio.

The separate manual `setIsAudioActiveAsync` API and the iOS changes have been removed from this PR. The iOS behavior will be addressed independently through the existing automatic audio-session management.

# Test Plan

Tested on a physical Android device with Spotify playing in the background:

- Start an unmuted video with `audioMixingMode: "auto"`: Spotify pauses.
- Pause, mute, or unmount the last audible video player: Spotify resumes automatically.
- `doNotMix` continues to use permanent audio focus.
- `et check-packages expo-video` passes.

# Checklist

- [x] This diff works correctly with `npx expo prebuild` and EAS Build.


packages/expo-video/src/VideoModule.ts
```diff
@@ -45,3 +45,52 @@ export function setVideoCacheSizeAsync(sizeBytes: number): Promise<void> {
 export function getCurrentVideoCacheSize(): number {
   return NativeVideoModule.getCurrentVideoCacheSize();
 }
+
+/**
+ * Activates or deactivates the audio session used by video players.
+ *
+ * When a `VideoPlayer` outputs audio and its `audioMixingMode` doesn't allow mixing, the operating
+ * system interrupts background audio played by other apps (such as music or podcast apps).
+ * Once video playback stops, the system does not resume the interrupted audio on its own.
```
amandeepmittal: After this PR, this sentence holds only on iOS. On Android with `audioMixingMode: 'auto'`, the focus request is now transient, and `AudioFocusManager` already abandons it once the last unmuted player pauses, mutes, or unmounts (both `onIsPlayingChanged` and `onMutedChanged` end up abandoning the focus). The system then resumes the interrupted app on its own. Your test plan says the same: pausing or unmounting the last unmuted player makes Spotify resume with no JS call. In the example below, Android already resumes at the `player.muted = true` line, so the call does its real work on iOS.

```suggestion
 * On iOS, once video playback stops, the system does not resume the interrupted audio on its own.
```

packages/expo-video/src/VideoModule.ts
```diff
@@ -45,3 +45,52 @@ export function setVideoCacheSizeAsync(sizeBytes: number): Promise<void> {
...
+ *
+ * On iOS, this function activates or deactivates the shared
+ * [`AVAudioSession`](https://developer.apple.com/documentation/avfaudio/avaudiosession).
+ * Deactivation uses the
+ * [`notifyOthersOnDeactivation`](https://developer.apple.com/documentation/avfaudio/avaudiosession/setactiveoptions/notifyothersondeactivation)
+ * option, which prompts other apps to resume their interrupted audio.
+ *
+ * On Android, calling this function with `false` abandons the audio focus held on behalf of
+ * video players, while calling it with `true` requests the audio focus again if any player
+ * currently needs it.
```
amandeepmittal: Following up on the comment above: since the `auto` focus request is now transient and abandoned automatically, it would help to tell Android readers when this function is actually needed.

```suggestion
 * currently needs it. With the `auto` audio mixing mode, the audio focus is abandoned
 * automatically once no player outputs audio, so background audio usually resumes on its own.
```

Related, if you want the Android story complete: with `doNotMix` the request stays permanent (`AUDIOFOCUS_GAIN`), and abandoning a permanent request does not make other apps resume, as your PR description notes. So on Android this function cannot deliver the resume for `doNotMix`. Your call whether that deserves a sentence here.

packages/expo-video/src/VideoModule.ts
```diff
@@ -45,3 +45,52 @@ export function setVideoCacheSizeAsync(sizeBytes: number): Promise<void> {
...
+ *
+ * The audio session is activated again automatically whenever a `VideoPlayer` starts outputting
+ * audio, so this function usually needs to be called only with `false`.
+ *
+ * > **Note:** On iOS, the audio session can be deactivated only while no players are playing
+ * > audio. Pause or mute all players before calling this function with `false`, otherwise the
```
amandeepmittal: The note says "Pause **or** mute", but the example below does both: mute, pause, deactivate, then play again. Which one does your device testing support?

- If muting alone lets the deactivation succeed, the example can drop the `player.pause()`/`player.play()` pair.
- If a muted playing player still keeps the session busy (`AVAudioSessionErrorCodeIsBusy`), then "or mute" overpromises. In that case the note should say: "Pause all players before calling this function with `false` (a muted player still keeps the audio session busy), otherwise the returned promise rejects with an error."

Readers who hit the rejection will follow the note, so it should match the tested behavior.

amandeepmittal (changes requested):

packages/expo-video/src/VideoModule.ts
huextrat: Good catch — applied your suggestion.

packages/expo-video/src/VideoModule.ts
huextrat: Applied, and took you up on the optional part: added a sentence noting that with `doNotMix` the focus request stays permanent, so background audio does not resume on Android in that mode even after the focus is abandoned.

packages/expo-video/src/VideoModule.ts
huextrat: You're right that "or mute" overpromised. Muting an `AVPlayer` only zeroes the volume at the mix stage — the audio render pipeline keeps running, so the session stays busy and `setActive(false)` fails with `AVAudioSessionErrorCodeIsBusy`. Updated the note to your second wording (pause all players first, with the muted-player caveat in parentheses); the example already pauses before deactivating.

behenate:
@expo-bot review

packages/expo-video/android/src/main/java/expo/modules/video/VideoModule.kt
```diff
@@ -64,6 +64,10 @@ class VideoModule : Module() {
       VideoManager.cache.clear()
     }
 
+    AsyncFunction("setIsAudioActiveAsync") { active: Boolean ->
+      VideoManager.setIsAudioActive(active)
+    }
+
```
behenate: In the current form this method doesn't make much sense on both Android and iOS. Unlike `expo-audio`, `expo-video` manages the audio being active automatically - it is updated every time the player starts of stops playing. Therefore when this option is set it's going to get overriden very quickly.

I'd prefer to keep the audio activity management automatic. 

Alternatively if we really need a manual `setIsAudioActiveAsync` - it would need a separate property in the `AudioFocusManager`, which would have higher priority than the automatic setting. Also it would probably need an option to return to automatic management. Nevertheless i believe a manual audio activity and transient focus are two separate features and should be in two separate PRs

behenate (changes requested):
Thank you for the PR. I believe that this should be split into two PRs - the transient audio focus should fully support the current automatic audio management, and the `setIsAudioActiveAsync` should be a separate, independent PR.

huextrat:
Thanks, that makes sense. I’ve followed your suggestion and reduced this PR to the Android transient audio focus change only, keeping the existing automatic audio focus management. I removed the manual `setIsAudioActiveAsync` API and all iOS changes from this PR, and updated the title, description, and changelog accordingly. I’ll address the iOS behavior separately through the automatic audio session management.

behenate:
@huextrat Thanks! can you verify the bot comments and fix them if they are accurate (verify on a real device/emulator)? I think they are valid.

huextrat:
Thanks! I verified the bot comments and addressed both valid Android findings:

- audio focus is now retained while the player is buffering or seeking, as long as playback is still expected to resume;
- legacy audio focus acquisition is now tracked explicitly, following the same pattern as `expo-audio`, so focus is correctly abandoned on Android API 24/25.

The iOS finding no longer applies to this PR since the iOS API and implementation were removed when the scope was reduced to Android. The Android Kotlin compilation passes with these fixes.

PR #48554, opened 2026-08-05
feat(env): Add setNodeEnv

ramonclaudio:
# Why

We want to set `NODE_ENV` consistently with `@expo/env`, using either `development` or `production`.

# How

We added a shared API that commands and tools can use to set the correct mode.

# Test Plan

Tests and CI checks pass.

# Checklist


packages/@expo/cli/src/utils/__tests__/nodeEnv-test.ts
```diff
@@ -0,0 +1,41 @@
+import { loadEnvFiles } from '../nodeEnv';
+
+jest.mock('../interactive', () => ({
+  shouldReduceLogs: () => false,
+}));
+
+jest.mock('2g', () => ({
```
kitten: Not quite sure if the mock here is needed? If we need a condition here, might be worth adding it to `2g`

packages/@expo/env/src/index.ts
```diff
@@ -39,6 +39,30 @@ export const KNOWN_MODES = ['development', 'test', 'production'];
...
+export function setNodeEnv(
+  mode: EnvMode,
+  { systemEnv = process.env }: { systemEnv?: EnvOutput } = {}
+) {
+  systemEnv.NODE_ENV = mode;
+  systemEnv.BABEL_ENV = mode;
+
+  if (systemEnv === process.env) {
+    // @ts-expect-error: __DEV__ is not declared in the Node.js types.
+    globalThis.__DEV__ = mode === 'development';
```
kitten: nit: We could just declare this. I'd also check if we actually need to set this. We might be setting this for server-side code in `@expo/cli`. Arguably, however, we may still want it there, since it'll then make the worker isolation migration clearer

`__DEV__` is special in that it typically is supposed to be a bundle replacement, so in theory we also shouldn't ever be accessing this. We may want to do a spike and "trap and error" when it's accessed to make sure it's never not compiled away

But maybe worth removing from here for now, unless this is mirroring existing code?

packages/@expo/env/src/index.ts
```diff
@@ -39,6 +39,30 @@ export const KNOWN_MODES = ['development', 'test', 'production'];
...
+ * Existing values are replaced. This also sets `__DEV__` when `process.env` is used.
+ *
+ * Pass a custom `systemEnv` to set values for a child process without changing `process.env`.
+ */
+export function setNodeEnv(
+  mode: EnvMode,
+  { systemEnv = process.env }: { systemEnv?: EnvOutput } = {}
+) {
+  systemEnv.NODE_ENV = mode;
+  systemEnv.BABEL_ENV = mode;
```
kitten: Let's make a ticket to separate this into the transform worker? That's the only place where it needs to be set. I don't remember if we and Metro don't set it (I think Metro should?) but maybe worth removing.

Although, I think this might manipulate Babel config loading. But I think that's abstracted now. So this might be redundant now

kitten (approved):

packages/@expo/env/src/index.ts
ramonclaudio: Got it! I'll open a ticket and circle back on this.

PR #48556, opened 2026-08-05
[android][secure-store] Add confirmation option

skylarbarrera:
# Why

Android biometric prompts default to requiring an explicit confirmation after a successful passive biometric match. Some applications need the platform-supported implicit-authentication path while retaining `BIOMETRIC_STRONG`, the existing `CryptoObject`, and authenticated keystore access.

# How

Adds an Android-only `requireConfirmation` SecureStore option, defaulting to `true`. The option is applied only when constructing the current `BiometricPrompt`; it is not persisted and does not change key aliases, key generation, authentication strength, invalidation, or iOS behavior. Reads use the stored authentication requirement together with the current call's confirmation preference.

The native-component fixture exposes the option for both sync and async reads/writes.

# Test Plan

Automated:

- `pnpm exec turbo build --filter=expo-secure-store`
- `pnpm test` in `packages/expo-secure-store` (Android and iOS; 37 tests / 25 snapshots)
- `pnpm typecheck` in `packages/expo-secure-store`
- `pnpm lint` in `packages/expo-secure-store`
- `pnpm exec et check-packages expo-secure-store`
- `./gradlew :expo-secure-store:testDebugUnitTest` in `apps/bare-expo/android`

Manual fixture coverage is available in the SecureStore native-component screen: enable authentication and compare `requireConfirmation` enabled/disabled for sync and async set/get on an Android device with passive biometrics. Android may override implicit authentication based on device or policy. On iOS the option is accepted but ignored, preserving the existing Keychain prompt behavior.

# Checklist


behenate (approved):
Looks good!

PR #48557, opened 2026-08-05
[ios][network] Fix data race in `getNetworkStateAsync`

Ignigena:
# Why

`Network.getNetworkStateAsync()` can crash with `EXC_BAD_ACCESS` on iOS when the network path changes while the call is in flight. We hit this in production: a TestFlight user's app crashed while backgrounded, during a telemetry flush that calls `getNetworkStateAsync()` before every upload — so the call lands in the window where iOS is tearing down and re-establishing interfaces as the app wakes for a background location burst.

Most calls from JS reach `NetworkModule.getNetworkStateAsync(path:)` without a path so it spins up a temporary `NWPathMonitor`, waits on a semaphore for the first path update, and returns the captured path.

`tempPath` is an unsynchronized `var` written from the monitor's update handler on a background queue and read by the waiting thread. `pathUpdateHandler` is invoked for every path change, not just the first, and [`cancel()`](https://github.com/expo/expo/blob/ebabc0b9ec7d149e539be4870288254fcdf861bf/packages/expo-network/ios/NetworkModule.swift#L108-L109) doesn't synchronously stop in-flight invocations. So a second path update — exactly what happens when the network is churning — writes `tempPath` while the caller is copying it out. `Optional<NWPath>` is a multi-word value wrapping an object reference and Swift's copy of it isn't atomic, so the reader can retain a half-written pointer.

The crash came from a TestFlight build on iPhone16,1 / iOS 26.5 with `expo-network@56.0.5`, with the app backgrounded:

```
Exception Type:    EXC_BAD_ACCESS (SIGSEGV)
Exception Subtype: KERN_INVALID_ADDRESS at 0x[REDACTED] (possible pointer authentication failure)
Role:              Non UI
OS Version:        iPhone OS 26.5 (23F77)

Thread 67 (crashed):
0   libobjc.A.dylib     objc_msgSend + 32
1   libswiftCore.dylib  swift::LayoutStringReader1::layoutStr() const
2   libswiftCore.dylib  unsigned long long swift::LayoutStringReader1::readBytes<unsigned long long>()
3   libswiftCore.dylib  swift_cvw_initWithCopyImpl(swift::OpaqueValue*, swift::OpaqueValue*, swift::TargetMetadata<swift::InProcess> const*)
4   libswiftCore.dylib  initializeWithCopy for ClosedRange<>.Index
5   libswiftCore.dylib  swift_cvw_initWithCopyImpl(swift::OpaqueValue*, swift::OpaqueValue*, swift::TargetMetadata<swift::InProcess> const*) (BytecodeLayouts.cpp:1182)
6   libswiftCore.dylib  initializeWithCopy for ClosedRange<>.Index
7   MyApp               <deduplicated_symbol>
8   MyApp               closure #1 in NetworkModule.getNetworkStateAsync(path:) (NetworkModule.swift:134)
9   MyApp               specialized _ArrayProtocol.filter(_:)
10  MyApp               NetworkModule.getNetworkStateAsync(path:) (NetworkModule.swift:134)
11  MyApp               closure #3 in NetworkModule.definition() (NetworkModule.swift:24)
12  MyApp               partial apply for closure #3 in NetworkModule.definition()
13  ExpoModulesCore     thunk for @escaping @callee_guaranteed () -> (@out A, @error @owned Error)
14  ExpoModulesCore     partial apply for thunk for @escaping @callee_guaranteed () -> (@out A, @error @owned Error)
15  ExpoModulesCore     specialized closure #3 in AsyncFunctionDefinition.call(_:this:arguments:callback:) (AsyncFunctionDefinition.swift:121)
16  ExpoModulesCore     closure #1 in AsyncFunctionDefinition.dispatchOnQueueUntilViewRegisters(appContext:arguments:queue:retryCount:_:) (AsyncFunctionDefinition.swift:174)
17  ExpoModulesCore     <deduplicated_symbol>
18  libdispatch.dylib   _dispatch_call_block_and_release (init.c:1597)
19  libdispatch.dylib   _dispatch_client_callout (client_callout.mm:85)
20  libdispatch.dylib   _dispatch_lane_serial_drain (queue.c:4033)
21  libdispatch.dylib   _dispatch_lane_invoke (queue.c:4124)
22  libdispatch.dylib   _dispatch_root_queue_drain_deferred_wlh (queue.c:7329)
23  libdispatch.dylib   _dispatch_workloop_worker_thread (queue.c:6923)
24  libsystem_pthread.dylib _pthread_wqthread (pthread.c:2696)
25  libsystem_pthread.dylib start_wqthread
```

The stack lines up with the race. Frame 11 is `closure #3 in NetworkModule.definition()` — the `AsyncFunction("getNetworkStateAsync")` handler, i.e. the no-argument call that goes through `getNetworkPathAsync()`. Frames 8-10 are the `NWInterface.InterfaceType.allCases.filter { currentPath?.usesInterfaceType($0) == true }` closure at `NetworkModule.swift:134`, where the captured path is copied per iteration. The crash is a Swift value copy (`swift_cvw_initWithCopyImpl`) reaching `objc_msgSend` on a pointer that fails PAC authentication — a torn value, not a released object.

Nothing about the race is specific to the background; that's just where path churn is constant in our app. A foreground call during a Wi-Fi to cellular handoff races the same way, with a narrower window.

# How

`getNetworkPathAsync()` now holds the captured path in `Mutex<NWPath?>` (the existing `ExpoModulesCore` utility) and keeps only the first update: later invocations of the handler — including any that land after `cancel()` — see a non-`nil` value and return without writing, and the semaphore is signaled once, for the write that won. The waiting thread's read is therefore ordered after a value that nothing can still be mutating.

Keeping the first update is what the existing comment already describes ("we want to wait for the result to be updated once"), and it's what callers got nearly every time anyway, since the wait is released by that first update. Whether a later update was observed instead was decided by the data race, so it wasn't behavior anything could depend on. There are no JS-side or public API changes: `NetworkState` keeps the same shape, and the result is the same snapshot the code always intended to return.

# Test Plan

No unit test accompanies this change. `NWPath` has no public initializer and `NWPathMonitor`'s update handler can't be driven from a test, so reproducing the race would mean introducing a protocol seam around the monitor purely for the test. Happy to add one if preferred, but keeping the changeset small and focused for now.

Verified on device with the Network screen in `apps/native-component-list`:

- `getNetworkStateAsync()` returns the same `type` / `isConnected` / `isInternetReachable` values as before on Wi-Fi, on cellular, and in airplane mode, so the first-update-wins behavior is unchanged for the normal path.
- Calling it in a loop while toggling airplane mode and Wi-Fi — which forces the repeated `pathUpdateHandler` invocations that cause the crash — no longer faults. Before the change, the same loop reproduced the `EXC_BAD_ACCESS` above.
- `Network.addNetworkStateListener(...)` still receives updates, confirming the module's own long-lived monitor is unaffected.

# Checklist


brentvatne:
@expo-bot verify

PR #48563, opened 2026-08-06
[router] Isolate `LoaderSuspenseStore` from `LoaderClient`

hassankhan:
# Why

`LoaderClient` previously owned a `LoaderSuspenseStore` instance and wrote to it directly. This coupling mixed two different concerns (transport vs Suspense state tracking) and made it harder to reason about behavior; store entries were written from multiple places inside the client, so Suspense state could change without any React-side involvement.

# How

- `LoaderClient` no longer imports or touches `LoaderSuspenseStore` (enforced by Oxlint too)
- `LoaderContext` now stores both `LoaderClient` and `LoaderSuspenseStore`
- `LoaderClient.invalidateAll()` is replaced by `revalidate()`, which now re-executes live sources and returns their paths; the HMR `loader-invalidate` listener calls `store.retain(client.revalidate())`
- `useLoaderData()` writes settled results through its subscriber, and its cleanup reclaims entries in two steps (`dispose()` and `teardown()`)

# Test Plan

- CI
- Manually testing

# Checklist

### DIFF
--- a/packages/expo-router/src/loaders/LoaderContext.ts
+++ b/packages/expo-router/src/loaders/LoaderContext.ts
@@ -1,0 +1,35 @@
+import { createContext } from 'react';
+
+import { LoaderClient } from './LoaderClient';
+import { LoaderSuspenseStore } from './LoaderSuspenseStore';
+import { bumpDevLoaderRevision } from './utils';
+
+export interface LoaderContextValue {
+  client: LoaderClient;
+  store: LoaderSuspenseStore;
+}
+
+export function createLoaderContextValue(client: LoaderClient): LoaderContextValue {
+  return { client, store: new LoaderSuspenseStore() };
+}
+
+export const defaultLoaderContextValue = createLoaderContextValue(new LoaderClient());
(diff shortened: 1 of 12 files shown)


packages/expo-router/src/loaders/LoaderContext.ts
```diff
@@ -0,0 +1,35 @@
...
+import { LoaderSuspenseStore } from './LoaderSuspenseStore';
+import { bumpDevLoaderRevision } from './utils';
+
+export interface LoaderContextValue {
+  client: LoaderClient;
+  store: LoaderSuspenseStore;
+}
+
+export function createLoaderContextValue(client: LoaderClient): LoaderContextValue {
+  return { client, store: new LoaderSuspenseStore() };
```
kitten, line 13: I don't really have enough current knowledge to judge whether the "enriching" via context is needed, and whether `__EXPO_LOADER_INVALIDATE_LISTENERS__` is the cleanest possible pattern that's not confusing, but it'll likely do the trick for now and we should just merge down all PRs before experimenting further 🤔

kitten (approved):

PR #48565, opened 2026-08-06
[ios][age-range] Fix swift-format violations blocking CI

intergalacticspacehighway:
# Why
`Swift Format` has been failing on every push to `main` and every PR touching an enrolled package since #48416, because `AgeRangeRecords.swift` already had unformatted code.

# How
- Ran `./scripts/swift-format.sh packages/expo-age-range` to auto-fix the two violations: manual `case`-pattern alignment and an over-length `log.error` call.

# Test Plan
Verify CI passes


intergalacticspacehighway:
Fixed by https://github.com/expo/expo/pull/48559

PR #48572, opened 2026-08-06
[docs] Add Japanese translation for the Build with AI tutorial

kazutoyo:
# Why

The Japanese docs currently cover the Expo tutorial only, so readers who start from the Japanese side hit a wall at the Build with AI tutorial. Japanese documentation matters a lot for Expo adoption in the Japanese community, and I wanted to help close that gap.

While registering the new pages I also found that `common/i18n.ts` kept three separate lists that had to be edited together for every translated page, which is easy to get wrong when adding a whole section at once.

# How

**Translation:** Added all seven chapters of the Build with AI tutorial under `pages/ja/tutorial/build-with-ai/`, matching the English source one-to-one (frontmatter, prose, `<BoxLink>` props, `<Collapsible>` summaries, and the `<ProgressTracker>` localization overrides). Code blocks, Terminal blocks, and the agent prompts are kept identical to English.

**i18n registration:** `EXPO_TUTORIAL_PATHS`, `PATHS_WITH_JAPANESE`, and `JA_SIDEBAR_TITLES` held the same page list in three places. Collapsed them into a single `JA_TRANSLATED_PAGES` map, with `isTranslatableSection`, `hasJapaneseTranslation`, and `getJapaneseSidebarTitle` all deriving from it, so adding a translated page is now a one-line change. Also registered the `Build with AI tutorial` sidebar section title.

**Tests:** Added `common/i18n.test.ts`, which asserts the registered paths and the files under `pages/ja/**` agree in both directions — a translated file that is never registered would silently render with an English sidebar, and a registered path with no file would 404. Added `checks/ja/anchors.test.ts` to catch ja-to-ja anchor links whose target heading no longer exists, since translating a heading changes its generated id. Both include a guard so the suite cannot pass vacuously on an empty scan.

**Drive-by:** #48193 fixed the anchor of the "Non-navigation components" link in `tutorial/build-a-screen.mdx` but left the old link text, which no longer matches the heading it points at. Updated the text in both the English and the Japanese page.

# Test Plan

- `pnpm test` — `common/i18n.test.ts`, `checks/ja/anchors.test.ts`, and the existing `checks/ja/sync.test.ts` pass (14 tests).
- `pnpm lint` — oxfmt, oxlint, tsc, and eslint are all clean.
- Ran the docs locally and walked the tutorial from http://localhost:3002/ja/tutorial/build-with-ai/introduction/ through `finishing-touches`, checking that the sidebar shows the Japanese chapter titles under the localized section heading and that the in-page links resolve.

# Checklist


docs/pages/ja/tutorial/build-with-ai/finishing-touches.mdx
```diff
@@ -0,0 +1,140 @@
...
+
+## ステータスバーを直す
+
+アプリの一番上を見てください。時計とバッテリーのアイコンが濃い色になっていて、暗い背景の上では読みづらくなっています。次のプロンプトをエージェントに貼り付けてください。
+
+```text プロンプト|wrap
+スマートフォンのステータスバー（画面上部の時計とバッテリーのアイコン）が、暗い背景の上で読みづらいです。expo-status-bar ライブラリを使って、すべての画面でステータスバーの文字を明るい色にして。
+```
+
+<Collapsible summary="英語の原文プロンプト">
```
amandeepmittal: ✏️ **style** `mdx-components.md#collapsible`

This summary repeats five times on this page (and the pattern recurs across the other chapters), and every repeat registers the same slug, so only the last block is anchor-scrollable. If these blocks are never meant to be link targets that is livable, but differentiating the summaries per step would keep the anchors usable.

docs/checks/ja/anchors.test.ts
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Every id the target page registers, in document order. Markdown headings drive the
+ * common case; `Collapsible` and `Requirement` register ids too, so they count as well.
+ */
+function headingIds(mdxPath: string): Set<string> {
+  const slugger = new GithubSlugger();
+  const ids = new Set<string>();
+  for (const line of fs.readFileSync(mdxPath, 'utf8').split('\n')) {
+    const heading = line.match(/^#{2,5}\s+(.*)$/);
+    const summary = line.match(/<(?:Collapsible\s+summary|Requirement\s+title)="([^"]+)"/);
```
amandeepmittal: ⚠️ **correctness** `mdx-components.md#prerequisites-and-requirement`

`Requirement` renders no anchor id, and the `Prerequisites` wrapper registers only a single `prerequisites` slug, so counting `Requirement title` as a heading id lets a ja link that targets one pass this test while it scrolls nowhere on the page. Drop the `Requirement\s+title` alternative here and the `Requirement` mention in the comment above.

```suggestion
    const summary = line.match(/<Collapsible\s+summary="([^"]+)"/);
```

amandeepmittal (changes requested):
Thank you for your contribution!

I've left comment as part of my review.

docs/checks/ja/anchors.test.ts
kazutoyo: Good catch, thanks!

kazutoyo:
Thanks for the thorough review!
I’ve accepted the suggestion and pushed the fix.

amandeepmittal (approved):

PR #48587, opened 2026-08-06
[router] Fix testing-library compatibility with @testing-library/react-native v14

### LINKED ISSUE #47444: expo-router/testing-library's screen does not have `getPathName` function anymore with `@testing-library/react-native` v14

https://github.com/kovbe21/jest-reproductions

### Steps to reproduce

In the reproducible example, run `path-name.test.tsx`.

Without example: 
- Create a new expo app (v55 or v56)
- Add jest-expo, react-native-testing-library and expo-router/testing-library config according to docs: https://docs.expo.dev/router/reference/testing/
- add the example test to the tests
- Run the tests
- It will fail with `TypeError: screen.getPathname is not a function`

The jest-matchers defined here (https://docs.expo.dev/router/reference/testing/#jest-matchers) does not seem to exist. It does exist (see temp fix below), but it is not visible for typescript.

Also reported at react-native-testing-library, they state that it is an ER bug. https://github.com/callstack/react-native-testing-library/issues/1924

I could fix this by adding this code snippet to `setup-files-after-env.tsx`

```
// Ensure expo-router testing utilities are initialized (registers `screen.getPathname` and matchers)
import 'expo-router/testing-library';

// Provide robust matchers that read directly from the router store.
try {
  const matchers = require('expect/build/matchers').default;
  const routerStore = require('expo-router/build/global-state/router-store');

  expect.extend({
    toHavePathname() {
      return matchers.toEqual(routerStore.store.getRouteInfo().pathname, arguments[1]);
    },
    toHavePathnameWithParams() {
      return matchers.toEqual(routerStore.store.getRouteInfo().pathnameWithParams, arguments[1]);
    },
    toHaveSegments() {
      return matchers.toEqual(routerStore.store.getRouteInfo().segments, arguments[1]);
    },
    toHaveSearchParams() {
      return matchers.toEqual(routerStore.store.getRouteInfo().params, arguments[1]);
    },
    toHaveRouterState() {
      return matchers.toEqual(routerStore.store.state, arguments[1]);
    },
  });
} catch {
  // ignore if expect internals aren't available
}

```

### Environment

```text
    System:
      OS: macOS 26.5.2
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 22.22.1 - /Users/bencekovacs/.nvm/versions/node/v22.22.1/bin/node
      npm: 10.9.4 - /Users/bencekovacs/.nvm/versions/node/v22.22.1/bin/npm
      Watchman: 2024.06.10.00 - /opt/homebrew/bin/watchman
    Managers:
    SDKs:
      iOS SDK:
      Android SDK:
        API Levels: 29, 30, 31, 32, 33, 34, 35, 36
        Build Tools: 29.0.2, 29.0.3, 30.0.2, 30.0.3, 31.0.0, 33.0.0, 33.0.1, 34.0.0, 35.0.0, 36.0.0
        System Images: android-28 | Google ARM64-V8a Play ARM 64 v8a, android-29 | Intel x86_64 Atom, android-29 | Google Play ARM 64 v8a, android-33 | Google APIs ARM 64 v8a, android-33 | Google Play ARM 64 v8a, android-34 | Google Play ARM 64 v8a, android-35 | Google Play ARM 64 v8a, android-36 | Google APIs ARM 64 v8a, android-36 | Google Play ARM 64 v8a
        Android NDK: 26.1.10909125
    IDEs:
      Android Studio: 2026.1 AI-261.23567.138.2611.15503007
    npmPackages:
      expo: ~55.0.27 => 55.0.27 
      expo-router: ~55.0.16 => 55.0.16 
      react: 19.2.0 => 19.2.0 
      react-dom: 19.2.0 => 19.2.0 
    npmGlobalPackages:
      eas-cli: 18.12.2
    Expo Workflow: bare
```

### Expo Doctor Diagnostics

```text
Need to install the following packages:
expo-doctor@1.20.0
Ok to proceed? (y) y

```

henriquepe:
# Why

Fixes #47444 — `expo-router/testing-library` is broken with `@testing-library/react-native` (RNTL) v14: `expect(screen).toHavePathname(...)` throws `TypeError: screen.getPathname is not a function`.

Root cause: RNTL v14 made `render()` asynchronous. `renderRouter` called it without awaiting and `Object.assign`ed the route helpers (`getPathname`, `getSegments`, …) onto the returned value — on v14 that decorates the **promise**, while the actual `RenderResult` (the same object RNTL registers as `screen`) never receives them. The custom matchers then fail on `screen.getPathname()`. (Confirmed from the RNTL side in callstack/react-native-testing-library#1924.)

# How

- `renderRouter` now detects whether `render()` returned a thenable:
  - **v13**: unchanged — fully synchronous, decorates and returns the result directly.
  - **v14**: returns a promise that decorates the **resolved** result, restoring the `result === screen` identity the helpers and matchers rely on. The return type mirrors the installed RNTL version (`ReturnType<typeof render> extends PromiseLike<any> ? Promise<RouterRenderResult> : RouterRenderResult`), so v14 users get correct `await renderRouter(...)` types.
- `testRouter` helpers (`navigate`, `push`, `replace`, `back`, `dismissAll`) chain their assertion after `act` settles on v14, and stay synchronous on v13. Since `act` returns a thenable on both versions, the discriminator is the observed `render` behavior (module flag set by `renderRouter`), not the `act` return value.
- The custom matchers now throw an actionable error (`Call \`renderRouter()\` before asserting — and await it when using @testing-library/react-native v14`) instead of a raw `TypeError` when the received value is not a decorated render result.
- Drive-by: `testRouter.setParams` asserted against a bare `screen` identifier that resolved to the global instead of RNTL's `screen`; it now uses `rnTestingLibrary.screen` like the other helpers.

No dependency changes: the monorepo stays on RNTL v13 (`^13.3.0` devDependency, shared by 21 other packages), and the peer range (`>= 13.2.0`) already allowed v14.

# Test Plan

- **v13 (no regression):** full `expo-router` suite on RNTL 13.3.3 — 347 suites / 4664 tests, 0 failures. All existing testing-library consumers (`smoke`, navigator suites, matcher snapshots) untouched and green.
- **v14 (unit):** new `src/__tests__/testing-library-async.test.ios.tsx` mocks RNTL with async `render`/`act` (resolving with the same object registered as `screen`, matching the v14 implementation) and covers: promise-returning `renderRouter` with decorated resolved result, decorated `screen` + matchers, `testRouter` helpers with async `act`, and the actionable matcher error for un-awaited results.
- **v14 (end-to-end):** ran the issue's minimal reproduction (https://github.com/kovbe21/jest-reproductions, `path-name.test.tsx`) against real RNTL **14.0.1**: reproduces the exact `TypeError` before the fix, passes with the patched build output.
- `pnpm build`, `pnpm typecheck`, `pnpm lint`, `pnpm format --check` all clean.

# Checklist

### DIFF
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -45,6 +45,7 @@
 ### 🐛 Bug fixes
 
 - Make layouts with explicitly declared screens honor `unstable_settings.initialRouteName` instead of declaration order, which can change deep-link back stacks. ([#48708](https://github.com/expo/expo/pull/48708) by [@Ubax](https://github.com/Ubax))
+- Fix `expo-router/testing-library` compatibility with `@testing-library/react-native` v14, where `render` and `act` are asynchronous and `screen.getPathname is not a function` was thrown. ([#48587](https://github.com/expo/expo/pull/48587) by [@henriquepe](https://github.com/henriquepe))
 - Prevent unfocused nested native tab navigators from redirecting global router state. ([#48257](https://github.com/expo/expo/pull/48257) by [@Ubax](https://github.com/Ubax))
 - Fixed `Tabs` and `TopTabs` (`expo-router/js-tabs`, `expo-router/js-top-tabs`) not being usable from RSC ([#48330](https://github.com/expo/expo/pull/48330) by [@Ubax](https://github.com/Ubax))
 - Redirect fully guarded navigators to parent ([#47984](https://github.com/expo/expo/pull/47984) by [@Ubax](https://github.com/Ubax))
--- a/packages/expo-router/src/__tests__/testing-library-async.test.ios.tsx
+++ b/packages/expo-router/src/__tests__/testing-library-async.test.ios.tsx
@@ -1,0 +1,55 @@
+import { screen } from '@testing-library/react-native';
+
+import { renderRouter, testRouter } from '../testing-library';
+
+/*
+ * Simulates `@testing-library/react-native` v14, where `render` and `act` are asynchronous
+ * (https://github.com/expo/expo/issues/47444). Like v14, the mocked `render` resolves with the
+ * same object it registers as `screen`.
+ */
+jest.mock('@testing-library/react-native', () => {
+  const actual = jest.requireActual('@testing-library/react-native');
+  const mocked = {
+    ...actual,
+    render: (...args: Parameters<typeof actual.render>) => Promise.resolve(actual.render(...args)),
+    act: (callback: () => unknown) => Promise.resolve(actual.act(callback)),
+  };
+  Object.defineProperty(mocked, 'screen', {
+    get: () => actual.screen,
+  });
+  return mocked;
+});
+
+describe('renderRouter with an asynchronous render', () => {
+  it('returns a promise and decorates the resolved result', async () => {
+    const pending = renderRouter(['[slug]'], { initialUrl: '/home?test=true' });
+    expect(typeof (pending as unknown as PromiseLike<unknown>).then).toBe('function');
+
+    const result = await pending;
+    expect(result.getPathname()).toBe('/home');
+    expect(result.getSegments()).toEqual(['[slug]']);
+    expect(result.getSearchParams()).toEqual({ slug: 'home', test: 'true' });
+  });
+
+  it('decorates the `screen` object so the matchers work', async () => {
+    await renderRouter(['[slug]'], { initialUrl: '/home?test=true' });
+    expect(screen).toHavePathname('/home');
+    expect(screen).toHavePathnameWithParams('/home?test=true');
+    expect(screen).toHaveSegments(['[slug]']);
+  });
+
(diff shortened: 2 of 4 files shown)


henriquepe:
Friendly ping — anything I can do to help move this along? Happy to rebase, split it up, or adjust the approach if the team would rather solve RNTL v14 compatibility differently (e.g. as part of a full v14 migration rather than the dual-version support here).

For context, the fix is scoped to `expo-router/testing-library`: `renderRouter` decorates the resolved render result instead of the promise, so v13 behavior is byte-for-byte unchanged while v14 stops throwing `screen.getPathname is not a function`. The monorepo stays on RNTL v13 — no dependency changes.

hassankhan:
Hey @henriquepe, thanks for the PR. I think it would be better to drop support for RNTL v13 altogether, so I'd remove any logic that tests for v13 support and update any RNTL versions specified in `package.json`s to v14+.

PR #48588, opened 2026-08-06
[ios][audio] Fix blocking calls on JS thread

### LINKED ISSUE #48531: [expo-audio] iOS: sync `play()` runs blocking `AVAudioSession.setActive(true)` on the JS thread – an audio-daemon stall freezes all JS execution

alanjhughes:
# Why
Closes #48531

# How
Interacting with the audio session can be blocking on ios so calls to things like setActive can block the JS thread.  The main change is that the session is activated once and kept active, rather than toggled. `sessionIsActive` tracks the state and is only touched on the session queue, so `activateSession()` is a no-op when the session is already active. 

# Test Plan
Bare expo


lukmccall (approved):

sbs44:
Hey Alan – Heads up on one regression from this PR: resumeInterruptedPlayers moved onto
sessionQueue, but handleInterruptionBegan still writes interruptedPlayers
and playerVolumes on the notification thread... back-to-back interruptions
now mutate and iterate those collections from two threads, where before both
ran on the notification thread. Fixed in #48913 along with two smaller
things (details over there).

PR #48592, opened 2026-08-06
[cli] Add expo skills command to link agent skills from npm packages

Kudo:
# Why

it's not scalable if we put all skills under expo claude/agent plugin even people don't install the expo module. we should introduce co-located skill inside an expo module.
this pr tries to add a cli to setup skills in expo modules
close ENG-25661

# How

- introduce `npx expo skills` to setup skills.
- leverage autolinking to scan packages with `skills/*/SKILL.md` and then setup symlinks under projectRoot's `.claude/skills`, `.agents/skills` and so on.
- the `npx expo skills` cli supports `sync`, `list`, `show` and `clean`. default is sync
- will update projectRoot/.gitignore for symlinks with `.claude/skills/*` we created. this is especially necessary for windows. on windows which doesn't support true symlink, we use "junction" instead. when junction would be inconsistent before and after node_modules installation.
- referenced https://github.com/antfu/skills-npm and https://github.com/vercel-labs/skills

# Test Plan

- added unit tests
- added e2e tests

# Checklist


packages/@expo/cli/src/skills/discovery.ts
```diff
@@ -0,0 +1,138 @@
...
+/** Finds every skill shipped by the packages in the project's dependency graph. */
+export async function discoverSkillsAsync(projectRoot: string): Promise<DiscoveredSkill[]> {
+  const autolinking =
+    require('expo/internal/unstable-autolinking-exports') as typeof import('expo-modules-autolinking/exports');
+  const linker = autolinking.makeCachedDependenciesLinker({ projectRoot });
+  const resolutions = await linker.scanDependenciesRecursively();
+
+  const skills = await Promise.all(
+    Object.values(resolutions).map((resolution) =>
+      resolution ? findPackageSkillsAsync(resolution.name, resolution.path) : []
```
kitten, line 27: (depending on other changes but leaving a comment early)

If we ever invoke this during `expo start`, there's also the `autolinkingModuleResolution`'s `makeCachedDependenciesLinker` call and `scanDependenciesRecursively` scan call that run. Not reusing that and running both in parallel can be very IO-intensive so we may have to design for them to cooperate somehow.

There's no existing hoisting and code for this yet afaik, and we should be careful not to retain `linker` in memory when we do add this.

I don't quite remember how many `2g` events are added, but given the constraints that refactor could be automated by instructing an agent to measure it with `pnpx 2g`. As long as it's in `expo/expo`, `autolinkingModuleResolution` is active anyway, so constraints should also be able to drive an automatic agentic refactor

PR #48596, opened 2026-08-06
[router] Allow descriptor without a `key`

Ubax:
# Why

Follow-up PR can add descriptors for placeholder routes (lazy loaded, which were not rendered yet). This PR is adjusting types for the coming change.

# How

# Test Plan

CI

# Checklist

### DIFF
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### 🛠 Breaking changes
 
-- Allow descriptor routes without a `key` in `expo-router/react-navigation` types.
+- Make `key` optional in `Descriptor` type ([#48596](https://github.com/expo/expo/pull/48596) by [@Ubax](https://github.com/Ubax))
 - Unify JS Tabs, TopTabs, Drawer, and headless tabs with NativeTabs - only screens declared in the layout become visible. ([#48499](https://github.com/expo/expo/pull/48499) by [@Ubax](https://github.com/Ubax))
 - Make `href: null` hide JS tabs and make their routes unreachable, redirecting navigation to the initial tab. ([#48499](https://github.com/expo/expo/pull/48499) by [@Ubax](https://github.com/Ubax))
 - Remove `getStateForRouteNamesChange` from the `Router` interface on `expo-router/react-navigation`. Custom routers must handle the `ROUTE_NAMES_CHANGED` action in `getStateForAction` instead. ([#48479](https://github.com/expo/expo/pull/48479) by [@Ubax](https://github.com/Ubax))
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### 🛠 Breaking changes
 
-- Make `key` optional in `Descriptor` type ([#48596](https://github.com/expo/expo/pull/48596) by [@Ubax](https://github.com/Ubax))
+- Allow `key` to be `undefined` in the `Descriptor` type and in routes passed to screen `options`, navigator `screenOptions`, and `RouteGroupConfig.screenOptions` callbacks. ([#48596](https://github.com/expo/expo/pull/48596) by [@Ubax](https://github.com/Ubax))
 - Unify JS Tabs, TopTabs, Drawer, and headless tabs with NativeTabs - only screens declared in the layout become visible. ([#48499](https://github.com/expo/expo/pull/48499) by [@Ubax](https://github.com/Ubax))
 - Make `href: null` hide JS tabs and make their routes unreachable, redirecting navigation to the initial tab. ([#48499](https://github.com/expo/expo/pull/48499) by [@Ubax](https://github.com/Ubax))
 - Remove `getStateForRouteNamesChange` from the `Router` interface on `expo-router/react-navigation`. Custom routers must handle the `ROUTE_NAMES_CHANGED` action in `getStateForAction` instead. ([#48479](https://github.com/expo/expo/pull/48479) by [@Ubax](https://github.com/Ubax))
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### 🛠 Breaking changes
 
+- Allow descriptor routes without a `key` in `expo-router/react-navigation` types.
 - Unify JS Tabs, TopTabs, Drawer, and headless tabs with NativeTabs - only screens declared in the layout become visible. ([#48499](https://github.com/expo/expo/pull/48499) by [@Ubax](https://github.com/Ubax))
 - Make `href: null` hide JS tabs and make their routes unreachable, redirecting navigation to the initial tab. ([#48499](https://github.com/expo/expo/pull/48499) by [@Ubax](https://github.com/Ubax))
 - Remove `getStateForRouteNamesChange` from the `Router` interface on `expo-router/react-navigation`. Custom routers must handle the `ROUTE_NAMES_CHANGED` action in `getStateForAction` instead. ([#48479](https://github.com/expo/expo/pull/48479) by [@Ubax](https://github.com/Ubax))
--- a/packages/expo-router/src/react-navigation/bottom-tabs/types.tsx
+++ b/packages/expo-router/src/react-navigation/bottom-tabs/types.tsx
@@ -13,6 +13,7 @@ import type { NavigatorArgs } from 'standard-navigation';
 import type { HeaderOptions, PlatformPressable } from '../elements';
 import type {
   Descriptor,
+  DescriptorRouteProp,
   NavigationProp,
   ParamListBase,
   RouteProp,
@@ -92,7 +93,8 @@ export type BottomTabOptionsArgs<
   ParamList extends ParamListBase,
   RouteName extends keyof ParamList = keyof ParamList,
   NavigatorID extends string | undefined = undefined,
-> = BottomTabScreenProps<ParamList, RouteName, NavigatorID> & {
+> = Omit<BottomTabScreenProps<ParamList, RouteName, NavigatorID>, 'route'> & {
+  route: DescriptorRouteProp<ParamList, RouteName>;
   theme: Theme;
 };
 
--- a/packages/expo-router/src/react-navigation/bottom-tabs/views/BottomTabBar.tsx
+++ b/packages/expo-router/src/react-navigation/bottom-tabs/views/BottomTabBar.tsx
@@ -333,7 +333,8 @@ export function BottomTabBar({ state, descriptors, emitter, navigateToTab, inset
       <View role="tablist" style={sidebar ? styles.sideContent : styles.bottomContent}>
         {routes.map((route, index) => {
           const focused = index === state.index;
-          const { options } = descriptors[route.key]!;
+          const descriptor = descriptors[route.key]!;
+          const { options } = descriptor;
 
           const onPress = () => {
             const event = emitter.emit({
@@ -367,14 +368,11 @@ export function BottomTabBar({ state, descriptors, emitter, navigateToTab, inset
                 : undefined;
 
           return (
-            <NavigationProvider
-              key={route.key}
-              route={route}
-              navigation={descriptors[route.key]!.navigation}>
+            <NavigationProvider key={route.name} route={route} navigation={descriptor.navigation}>
               <BottomTabItem
                 href={buildHref(route.name, route.params)}
                 route={route}
-                descriptor={descriptors[route.key]!}
+                descriptor={descriptor}
                 focused={focused}
                 horizontal={hasHorizontalLabels}
                 compact={compact}
--- a/packages/expo-router/src/react-navigation/bottom-tabs/views/BottomTabView.tsx
+++ b/packages/expo-router/src/react-navigation/bottom-tabs/views/BottomTabView.tsx
@@ -231,8 +231,11 @@ export function BottomTabView(props: Props) {
           const isFocused = state.index === index;
           const isPreloaded = preloadedRouteKeys.includes(route.key);
 
-          if (lazy && !loaded.includes(route.key) && !isFocused && !isPreloaded) {
-            // Don't render a lazy screen if we've never navigated to it or it wasn't preloaded
+          if (
+            descriptor.route.key === undefined ||
+            (lazy && !loaded.includes(route.key) && !isFocused && !isPreloaded)
+          ) {
+            // Don't render placeholder or unloaded lazy screens.
             return null;
           }
 
--- a/packages/expo-router/src/react-navigation/bottom-tabs/views/BottomTabView.tsx
+++ b/packages/expo-router/src/react-navigation/bottom-tabs/views/BottomTabView.tsx
@@ -281,14 +281,14 @@ export function BottomTabView(props: Props) {
                 value={tabBarPosition === 'bottom' ? tabBarHeight : 0}>
                 <Screen
                   focused={isFocused}
-                  route={descriptor.route}
+                  route={route}
                   navigation={descriptor.navigation}
                   headerShown={headerShown}
                   headerStatusBarHeight={headerStatusBarHeight}
                   headerTransparent={headerTransparent}
                   header={header({
                     layout: dimensions,
-                    route: descriptor.route,
+                    route,
                     navigation: descriptor.navigation as BottomTabNavigationProp<ParamListBase>,
                     options: descriptor.options,
                   })}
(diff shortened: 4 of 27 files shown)


packages/expo-router/src/react-navigation/core/types.tsx
```diff
@@ -568,7 +577,7 @@ export type Descriptor<
   /**
    * Route object for the screen
    */
-  route: Route;
+  route: DescriptorRoute<Route>;
```
jakub-agent, line 584: [🏗️ **design**] This widens `route.key` for **every** descriptor in every navigator, but nothing currently produces a keyless route: `useDescriptors` is the only descriptor factory and it always assigns a state route with a real `key` (`useDescriptors.tsx:271-297`). So as it stands the PR only removes type safety from all consumers, with no counterpart that needs it.

Two questions:

1. Which follow-up introduces the keyless producer? If this is step 1 of a stack, linking it in the PR body would make the breaking change reviewable. Right now the body is the unfilled template.
2. Was a separate type considered instead of widening the shared `Descriptor` — e.g. keeping `Descriptor.route: Route` exact and adding a distinct descriptor type (or a union) for layout-declared-without-instance screens? Only the tabs-style navigators can produce that case, but every custom navigator author now has to reason about `key === undefined`, including in navigators where it is unreachable.

Minor, related: `key: string | undefined` is a *required* property that may be `undefined`, so anyone constructing a descriptor route must spell out `key: undefined`. If that is deliberate (forcing authors to decide), fine — worth saying so in the doc comment; if not, `key?: string` reads better.

packages/expo-router/src/react-navigation/stack/views/Header/HeaderContainer.tsx
```diff
@@ -71,7 +71,7 @@ export function HeaderContainer({
 
         const isFocused = focusedRoute.key === scene.descriptor.route.key;
```
jakub-agent: [🏗️ **design**] The line below was migrated to `scene.route`, but this one still reads `scene.descriptor.route`. Two consequences:

- After this PR the right-hand side is `string | undefined`, so if it is ever `undefined` the comparison silently evaluates to `false` and the header renders as unfocused rather than failing loudly.
- `scene.descriptor` can be `FALLBACK_DESCRIPTOR` — `Object.freeze({ options: {} })` at `CardStack.tsx:99`, assigned at `CardStack.tsx:287` when a route has no descriptor and no previous scene. That object has no `route` at all, so this line throws. `scene.route` is always populated (`CardStack.tsx:355`).

Same applies to `scene.descriptor.route` at line 139 and to `getPreviousScene` in `CardStack.tsx:567` (`scene.descriptor.route.key === previousRoute.key`), which is outside the diff so I can't comment there. Suggest finishing the migration in all three so `scene.descriptor` is only ever used for `options`, `navigation` and `render`.

packages/expo-router/src/react-navigation/stack/views/Header/HeaderContainer.tsx
```diff
@@ -137,7 +137,7 @@ export function HeaderContainer({
         return (
           <NavigationProvider
             key={scene.descriptor.route.key}
```
jakub-agent: [💡 suggestion] Leftover from the same migration — the `route` prop right below now uses `scene.route` while the React `key` still uses `scene.descriptor.route.key`, which this PR makes `string | undefined`. An `undefined` key makes React fall back to the array index and warn, quietly changing header reconciliation. `scene.route.key` is always a string.

packages/expo-router/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### 🛠 Breaking changes
 
+- Allow descriptor routes without a `key` in `expo-router/react-navigation` types.
```
jakub-agent: [🐧 nit] Missing the PR link and author. Every other entry in this file follows `… ([#PR](https://github.com/expo/expo/pull/PR) by [@user](https://github.com/user))`, which is the format the [changelog guide](https://github.com/expo/expo/blob/main/guides/contributing/Updating%20Changelogs.md) documents.

The wording could also be more actionable for the people this breaks — custom navigator authors. Something like: "`Descriptor['route']['key']` is now `string | undefined`. Custom navigators reading `descriptor.route.key` must handle the missing key."

packages/expo-router/src/react-navigation/core/types.tsx
```diff
@@ -549,6 +549,15 @@ export type ScreenLayoutArgs<
  */
 export type RouteSource = 'layout' | 'filesystem';
 
+/**
+ * Route carried by a descriptor. `key` is `undefined` when the descriptor
+ * describes a route name declared in a layout that has no live route
+ * instance in navigation state.
+ */
+export type DescriptorRoute<Route extends RouteProp<any, any>> = Route extends unknown
```
jakub-agent: [🐧 nit] `Route extends unknown ? … : never` is a distributive conditional — without it, `Omit<A | B, 'key'>` collapses the union to its common keys. That is the right call, but it looks like a no-op and someone will eventually "simplify" it away. A one-line comment saying it exists to distribute over unions would protect it.

Also, naming the type parameter `Route` shadows the `Route` type from `../routers`, which this file imports. `R` or `RouteType` would read more clearly.

packages/expo-router/src/standard-navigation/useBuildHref.ts
```diff
@@ -8,12 +8,14 @@ import {
...
+type HrefRoute = Pick<NavigationRoute<ParamListBase, string>, 'name' | 'params'>;
+
 // TODO(@ubax): move route info to state - https://linear.app/expo/issue/ENG-21483/refactor-state-to-include-all-route-info-information
 export function useBuildHref() {
   const currentState = useStateForPath();
   return useMemo(() => {
-    const cache = new WeakMap<NavigationRoute<ParamListBase, string>, string>();
-    return (route: NavigationRoute<ParamListBase, string>) => {
+    const cache = new WeakMap<HrefRoute, string>();
+    return (route: HrefRoute) => {
```
jakub-agent, line 19: [🐧 nit] Narrowing the parameter to what the function actually reads is good. One caveat worth knowing: the cache is a `WeakMap` keyed by object identity, and the parameter is now structural, so any `{ name, params }` literal is accepted. A caller that builds one inline per render gets zero cache hits with no signal. Both current callers pass stable route objects, so nothing is wrong today — but since nothing in this PR needs the wider type yet, it might be simpler to land it alongside the caller that does.

jakub-agent (commented):

packages/expo-router/src/react-navigation/core/types.tsx
Ubax: Here is the actual change - https://github.com/expo/expo/pull/48618

packages/expo-router/src/react-navigation/stack/views/Header/HeaderContainer.tsx
Ubax: Apply the changes

packages/expo-router/src/react-navigation/stack/views/Header/HeaderContainer.tsx
Ubax: Apply the changes

packages/expo-router/src/react-navigation/core/types.tsx
Ubax: Add the comment

packages/expo-router/src/standard-navigation/useBuildHref.ts
Ubax: Add a comment explaining that it should be kept stable

packages/expo-router/src/react-navigation/core/types.tsx
```diff
@@ -653,7 +666,7 @@ export type RouteConfigProps<
   options?:
     | ScreenOptions
     | ((props: {
-        route: RouteProp<ParamList, RouteName>;
+        route: DescriptorRouteProp<ParamList, RouteName>;
```
jakub-agent, line 669: [🚨 **critical**] This is the biggest change in the PR and the CHANGELOG doesn't mention it. `options` on `<Stack.Screen>` / `<Tabs.Screen>`, plus `screenOptions` on every navigator and on `RouteGroupConfig`, now hand `DescriptorRouteProp` to ordinary app authors — not just to custom navigator authors, which is all the changelog entry covers.

I typechecked a scratch file against both `9970089b11e^` and `0424786`. Clean before, three errors after:

```tsx
// TS2322 on both <Stack> and <Tabs> — a named options helper typed against the old contract
function myScreenOptions({ route }: { route: RouteProp<ParamListBase, string>; navigation: any; theme: Theme }) {
  return { title: route.name };
}
<Stack screenOptions={myScreenOptions} />

// TS2345 — forwarding the callback's route to anything that wants a real route
declare function titleFor(route: RouteProp<ParamListBase, string>): string;
<Stack screenOptions={({ route }) => ({ title: titleFor(route) })} />
```

Reading `route.key` inline still compiles, so the break is specifically about naming the parameter type or passing `route` onward. Both are common.

Two requests:

1. Say this in the CHANGELOG explicitly — that `options` and `screenOptions` callbacks now receive a route whose `key` may be `undefined`, and what an app author should do about it.
2. Worth a second look at whether the options callbacks need the widened type at all. A descriptor for a layout-declared screen with no live route does need one, but if options are only ever resolved for routes that exist in state, keeping `RouteProp` here would leave the break confined to `Descriptor` — which is a far smaller audience.

packages/expo-router/src/react-navigation/bottom-tabs/types.tsx
```diff
@@ -92,7 +93,8 @@ export type BottomTabOptionsArgs<
   ParamList extends ParamListBase,
   RouteName extends keyof ParamList = keyof ParamList,
   NavigatorID extends string | undefined = undefined,
-> = BottomTabScreenProps<ParamList, RouteName, NavigatorID> & {
+> = Omit<BottomTabScreenProps<ParamList, RouteName, NavigatorID>, 'route'> & {
+  route: DescriptorRouteProp<ParamList, RouteName>;
```
jakub-agent, line 97: [🏗️ **design**] Only this one was migrated. `StackOptionsArgs` (`stack/types.tsx:79`), `NativeStackOptionsArgs` (`native-stack/types.tsx:82`), `DrawerOptionsArgs` (`drawer/types.tsx:302`) and `MaterialTopTabOptionsArgs` (`material-top-tabs/types.tsx:88`) still spread `…ScreenProps`, so they keep `route: RouteProp<…>` with `key: string`.

All four are publicly exported (each navigator's `index.tsx`, and `NativeStackOptionsArgs` from `src/exports.ts:137`) and show up in the generated API reference, so their whole purpose is annotating an options callback. After this PR they describe a contract the API no longer provides — annotating with them produces the TS2322 in my other comment. Either migrate all five or leave all five, but the current split will confuse people.

packages/expo-router/src/react-navigation/core/types.tsx
```diff
@@ -549,6 +554,14 @@ export type ScreenLayoutArgs<
  */
 export type RouteSource = 'layout' | 'filesystem';
 
+/**
+ * Route carried by a descriptor. `key` is `undefined` when the descriptor
+ * describes a route name declared in a layout that has no live route
+ * instance in navigation state.
+ */
+export type DescriptorRoute<Route extends RouteProp<any, any>> = Omit<Route, 'key'> &
```
jakub-agent, line 562: [💡 suggestion] Dropping the `Route extends unknown ? … : never` wrapper changed behavior rather than just tidying the type. `Omit<A | B, 'key'>` keeps only the keys `A` and `B` share, so a union `Route` collapses to `{ name: 'a' | 'b'; params: PA | PB }` and the name↔params correlation is lost; the distributive form preserved it.

Nothing instantiates `Descriptor` with a union `Route` today, so this is theoretical — flagging it because my earlier comment asked for a comment on the construct, not its removal. If union `Route` parameters are deliberately out of scope, saying so in the doc block above would settle it.

packages/expo-router/src/react-navigation/core/types.tsx
```diff
@@ -478,6 +478,11 @@ export type RouteProp<
   RouteName extends keyof ParamList = Keyof<ParamList>,
 > = Route<Extract<RouteName, string>, ParamList[RouteName]>;
 
+export type DescriptorRouteProp<
+  ParamList extends ParamListBase,
+  RouteName extends keyof ParamList = Keyof<ParamList>,
+> = DescriptorRoute<RouteProp<ParamList, RouteName>>;
```
jakub-agent, line 484: [💡 suggestion] Still no test, and the surface is wider than in iteration 1 — this alias now flows into `Descriptor`, `RouteConfigProps.options`, `DefaultNavigatorOptions.screenOptions`, `RouteGroupConfig.screenOptions` and `BottomTabOptionsArgs`.

The package already has the convention (`src/react-navigation/bottom-tabs/__tests__/types.test.ts`). A short type test would pin the two properties that matter: a real `RouteProp` still assigns to `DescriptorRouteProp`, and `key` reads back as `string | undefined`. Cheap, and it makes an accidental revert of either direction visible.

packages/expo-router/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### 🛠 Breaking changes
 
+- Make `key` optional in `Descriptor` type ([#48596](https://github.com/expo/expo/pull/48596) by [@Ubax](https://github.com/Ubax))
```
jakub-agent: [💡 suggestion] Thanks for adding the link and author. Two things about the text:

- "optional" isn't quite what the type says. `key: string | undefined` is a *required* property that may be `undefined` — an optional `key?: string` would let you omit it entirely. Readers will size the change differently depending on which one they think it is.
- It only mentions `Descriptor`, but the second commit also changes `options` and `screenOptions` callbacks, which is what most app authors will actually hit.

Something like: "`Descriptor['route']['key']` and the `route` passed to `options` / `screenOptions` callbacks are now `string | undefined`. Code that reads the key, or forwards the route to something typed `RouteProp`, needs to handle the missing key."

Also missing the sentence-ending period that the surrounding entries use before the link.

jakub-agent (changes requested):

packages/expo-router/src/react-navigation/core/types.tsx
Ubax: Mention this in changelog as well

packages/expo-router/src/react-navigation/bottom-tabs/views/BottomTabBar.tsx
```diff
@@ -367,14 +368,11 @@ export function BottomTabBar({ state, descriptors, emitter, navigateToTab, inset
                 : undefined;
 
           return (
-            <NavigationProvider
-              key={route.key}
-              route={route}
-              navigation={descriptors[route.key]!.navigation}>
+            <NavigationProvider key={route.name} route={route} navigation={descriptor.navigation}>
```
jakub-agent, line 371: [🐧 nit] This switch from `route.key` to `route.name` is load-bearing but reads like a drive-by. With #48618, a placeholder route enters state as `key: name` and then becomes a real route with `key: "name-<nonce>"` once the tab is first visited. Keying on `route.key` would unmount and remount the whole tab item across that swap; keying on `route.name` doesn't, and names are unique within a tab navigator so it stays a valid key.

Worth a one-line comment — it's the kind of thing someone "fixes" back to `route.key` later.

packages/expo-router/src/react-navigation/bottom-tabs/views/BottomTabView.tsx
```diff
@@ -231,8 +231,11 @@ export function BottomTabView(props: Props) {
           const isFocused = state.index === index;
           const isPreloaded = preloadedRouteKeys.includes(route.key);
 
-          if (lazy && !loaded.includes(route.key) && !isFocused && !isPreloaded) {
-            // Don't render a lazy screen if we've never navigated to it or it wasn't preloaded
+          if (
+            descriptor.route.key === undefined ||
```
jakub-agent, line 235: [💡 suggestion] This branch (and the matching one in `DrawerView.tsx:214`) is unreachable in this PR — nothing constructs a descriptor with `route.key === undefined` until #48618 lands `appendMissingPlaceholderTabDescriptors`. So it ships untested here, and the type tests added in this PR don't cover it either.

Not asking for a test on a code path that has no producer yet. Just flagging so the runtime behaviour gets exercised in #48618 rather than being assumed covered by this PR's suite.

packages/expo-router/src/react-navigation/core/__tests__/types.test.ts
```diff
@@ -0,0 +1,41 @@
+import type {
+  DefaultNavigatorOptions,
+  EventMapBase,
+  NavigationState,
+  ParamListBase,
+  RouteGroupConfig,
+} from '../../native';
+
+type Expect<T extends true> = T;
```
jakub-agent, line 9: [🐧 nit] `Expect`, `Equal` and `CallbackArgument` are now copy-pasted across six test files. A single shared test helper would match how the rest of the package factors things out, and would keep the `Equal` definition — which is easy to get subtly wrong — in one place.

jakub-agent (approved):

PR #48597, opened 2026-08-06
[test-suite] Add shared JasmineInterface type

vonovak:
## Why

`apps/test-suite` has no type-check step, and TypeScript 6 made `strict` the default, so 651 errors accumulated unnoticed. This is the base of a 14-PR stack that takes the app to zero.

## How

Export the existing `JasmineInterface` as `TestHarness` so test modules can type the jasmine object they receive instead of an implicit `any`. Type four app-shell values.

## Test Plan

`pnpm tsc` drops 651 → 647. `lint --max-warnings 0`, `format --check` and `test` all pass.

top of the stack should be completely green

### DIFF
--- a/apps/test-suite/types.ts
+++ b/apps/test-suite/types.ts
@@ -1,3 +1,5 @@
+export type { JasmineInterface } from 'jasmine-core/lib/jasmine-core/jasmine';
+
 export type JasmineResult = {
   id: string;
   description: string;
--- a/apps/test-suite/TestModules.ts
+++ b/apps/test-suite/TestModules.ts
@@ -16,7 +16,7 @@ function browserSupportsWebGL() {
   }
 }
 
-function optionalRequire(requirer) {
+function optionalRequire<T>(requirer: () => T): T | undefined {
   try {
     return requirer();
   } catch {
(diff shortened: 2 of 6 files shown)


apps/test-suite/types.ts
```diff
@@ -1,3 +1,13 @@
+import type { JasmineInterface } from 'jasmine-core/lib/jasmine-core/jasmine';
+
+/**
+ * The jasmine interface that `TestScreen` passes to every test module as the
+ * first argument of its exported `test` function. Test modules either take it
+ * whole (`test(t: TestHarness)`) or destructure the parts they use
+ * (`test({ describe, expect, it }: TestHarness)`).
+ */
+export type TestHarness = JasmineInterface;
```
Kudo: not a fan to call the type as harness. people call harness as agent related nowadays 😅

vonovak: Fair point — dropped the invented name entirely rather than picking another one.

`jasmine-core.d.ts` already declares this exact type as `JasmineInterface`, so `types.ts` now just re-exports it:

```ts
export type { JasmineInterface } from 'jasmine-core/lib/jasmine-core/jasmine';
```

Renamed through the rest of the stack too, so all 17 test modules now read `test(t: JasmineInterface)` / `test({ describe, expect, it }: JasmineInterface)`.

Kudo (approved):

tsapeta (approved):

PR #48599, opened 2026-08-06
[video] Type-check Video.tsx under strict

vonovak:
# Why

Part of the stack in #48597, which explains why `apps/test-suite` accumulated 651 errors under `strict` and how the work is split up. This PR covers `tests/Video.tsx`.

# How

156 of the 184 errors came from one `player` declaration that is assigned in `beforeEach`. Also types the recorded `sourceLoad` fixtures for the subset the assertions read, and fixes two real defects in the event helpers: a `() => void | null` precedence bug, and a subscription assigned inside a promise executor so `remove()` narrowed to `never`.

# Test Plan

`pnpm tsc` drops 622 → 438 on this branch. `pnpm lint --max-warnings 0`, `pnpm format --check` and `pnpm test` all pass, verified on this branch rather than only on the tip of the stack.

# Checklist

### DIFF
--- a/apps/test-suite/tests/Video.tsx
+++ b/apps/test-suite/tests/Video.tsx
@@ -1,6 +1,5 @@
-'use strict';
-
 import {
+  AudioTrack,
   createVideoPlayer,
   SourceLoadEventPayload,
   SubtitleTrack,
@@ -10,6 +9,7 @@ import {
   VideoPlayerStatus,
   VideoSource,
   VideoSourceObject,
+  VideoTrack,
   VideoView,
 } from 'expo-video';
 import React from 'react';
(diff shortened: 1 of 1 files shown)


apps/test-suite/tests/Video.tsx
```diff
@@ -73,7 +73,26 @@ const brokenSource = {
...
+ * reads are omitted, and `videoSource` is kept for reference only.
+ */
+type ExpectedSourceLoadPayload = {
+  duration: number;
+  availableVideoTracks: Omit<
+    SourceLoadEventPayload['availableVideoTracks'][number],
+    'url' | 'videoRange'
+  >[];
+  availableAudioTracks: SourceLoadEventPayload['availableAudioTracks'];
+  availableSubtitleTracks: SourceLoadEventPayload['availableSubtitleTracks'];
```
behenate: I think we can import the types and use them directly
```suggestion
  availableVideoTracks: Omit<VideoTrack, 'url' | 'videoRange'>[];
  availableAudioTracks: AudioTrack[];
  availableSubtitleTracks: SubtitleTrack[];
```

apps/test-suite/tests/Video.tsx
```diff
@@ -226,14 +250,11 @@ export async function test({ describe, expect, it, ...t }, { setPortalChild, cle
...
-        cleanupPortal();
-      } catch (error: any) {
-        console.warn('Player release error:', error.message);
-      }
-      player = null;
+    try {
+      player.release();
+      cleanupPortal();
+    } catch (error: any) {
+      console.warn('Player release error:', error.message);
```
behenate, line 214:
```suggestion
  try {
    player.release();
  } catch (error: any) {
    console.warn('Player release error:', error.message);
  } finally {
    cleanupPortal();
  }
```

Unrelated to your changes, but I think we can cleanup the portal in a finally block

apps/test-suite/tests/Video.tsx
```diff
@@ -73,7 +73,26 @@ const brokenSource = {
...
+type ExpectedSourceLoadPayload = {
+  duration: number;
+  availableVideoTracks: Omit<
+    SourceLoadEventPayload['availableVideoTracks'][number],
+    'url' | 'videoRange'
+  >[];
+  availableAudioTracks: SourceLoadEventPayload['availableAudioTracks'];
+  availableSubtitleTracks: SourceLoadEventPayload['availableSubtitleTracks'];
+  /** Recorded for reference; no assertion compares it. */
+  videoSource?: unknown;
```
behenate: `unknown` type is kind of ugly. I'd just drop the `videoSourceField` since we aren't using it.

behenate (approved):
A few small nits, I don't see any major issues 😄

PR #48600, opened 2026-08-06
[calendar] Type-check CalendarNext.ts under strict

vonovak:
# Why

Part of the stack in #48597, which explains why `apps/test-suite` accumulated 651 errors under `strict` and how the work is split up. This PR covers `tests/CalendarNext.ts`.

# How

Mostly reads that a preceding `toBeDefined()`/`not.toBeNull()` already guarantees. `getReminderCalendar()` now throws instead of returning `undefined`, two literals use `satisfies` instead of a widening `Partial<>`, permission and dialog assertions compare enum members, and the shared attendee/organizer assertions are split since an organizer has no `id` or `email`.

Follow-up for expo-calendar maintainers: 16 sites here carry a `@ts-expect-error` because `update()` accepts `null` to clear a field at runtime while `Partial<Modifiable*Properties>` does not allow it — widening those types would retire all 16 at once.

# Test Plan

`pnpm tsc` drops 438 → 365 on this branch. `pnpm lint --max-warnings 0`, `pnpm format --check` and `pnpm test` all pass, verified on this branch rather than only on the tip of the stack.

# Checklist

### DIFF
--- a/apps/test-suite/tests/CalendarNext.ts
+++ b/apps/test-suite/tests/CalendarNext.ts
@@ -9,10 +9,11 @@ import {
   EntityTypes,
 } from 'expo-calendar';
 import * as Calendar from 'expo-calendar';
-import { UnavailabilityError } from 'expo-modules-core';
+import { PermissionStatus, UnavailabilityError } from 'expo-modules-core';
 import { Platform } from 'react-native';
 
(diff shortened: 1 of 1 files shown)


Wenszel (approved):
LGTM, I’ll follow up with the type fix.

PR #48605, opened 2026-08-06
[image-manipulator] Type-check ImageManipulator.ts under strict

vonovak:
# Why

Part of the stack in #48597, which explains why `apps/test-suite` accumulated 651 errors under `strict` and how the work is split up. This PR covers `tests/ImageManipulator.ts`.

# How

41 of the 49 errors came from two untyped suite variables. Typing them as `Asset` surfaced that `localUri`/`width`/`height` are nullable until `downloadAsync()` resolves. The four size comparisons go through a helper that narrows `FileInfo` on `exists` and throws if the file is missing.

Two sites here carry a `@ts-expect-error` because `ImageManipulator.Image` is declared as an `ImageRef` instance where the runtime value is the class. #48613, at the top of this stack, fixes that declaration and removes both directives.

# Test Plan

`pnpm tsc` drops 149 → 100 on this branch. `pnpm lint --max-warnings 0`, `pnpm format --check` and `pnpm test` all pass, verified on this branch rather than only on the tip of the stack.

# Checklist

### DIFF
--- a/apps/test-suite/tests/ImageManipulator.ts
+++ b/apps/test-suite/tests/ImageManipulator.ts
@@ -3,11 +3,26 @@ import * as FileSystem from 'expo-file-system/legacy';
 import { manipulateAsync, ImageManipulator, FlipType, SaveFormat } from 'expo-image-manipulator';
 import { Platform } from 'react-native';
 
+import type { JasmineInterface } from '../types';
+
 export const name = 'ImageManipulator';
 
-export async function test(t) {
+/**
+ * `FileSystem.getInfoAsync` returns a union discriminated on `exists`, and only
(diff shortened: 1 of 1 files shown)


apps/test-suite/tests/ImageManipulator.ts
```diff
@@ -45,37 +63,40 @@ export async function test(t) {
         const image = await context.renderAsync();
 
         t.expect(image).toBeDefined();
+        // @ts-expect-error `ImageManipulator.Image` is declared as an `ImageRef` instance rather than
```
vonovak, line 69: resolved in https://github.com/expo/expo/pull/48613

tsapeta (approved):

PR #48608, opened 2026-08-06
[fetch] Type-check Fetch.ts under strict

vonovak:
# Why

Part of the stack in #48597, which explains why `apps/test-suite` accumulated 651 errors under `strict` and how the work is split up. This PR covers `tests/Fetch.ts`.

# How

Harness types for both suites, and non-null on `resp.body` in the streaming specs.

# Test Plan

`pnpm tsc` drops 52 → 35 on this branch. `pnpm lint --max-warnings 0`, `pnpm format --check` and `pnpm test` all pass, verified on this branch rather than only on the tip of the stack.

# Checklist

### DIFF
--- a/apps/test-suite/tests/Fetch.ts
+++ b/apps/test-suite/tests/Fetch.ts
@@ -3,6 +3,7 @@ import { fetch } from 'expo/fetch';
 import { Platform } from 'react-native';
 
 import type { JasmineInterface } from '../types';
+import { requireNotNull } from '../utils/requireNotNull';
 
 export const name = 'Fetch';
 
@@ -57,13 +58,13 @@ export function test({ describe, expect, it, ...t }: JasmineInterface) {
     it('should process response in readablestream from late get reader call', async () => {
       const resp = await fetch('https://httpbin.io/get');
       expect(resp.ok).toBe(true);
-      expect(resp.body!).not.toBeNull();
+      expect(resp.body).not.toBeNull();
 
       // Delay 0.5s to ensure the response is completed before streaming started
       await delayAsync(500);
 
       const chunks = [];
-      const reader = resp.body!.getReader();
+      const reader = requireNotNull(resp.body).getReader();
       while (true) {
         const { done, value } = await reader.read();
         if (done) {
@@ -153,7 +154,7 @@ export function test({ describe, expect, it, ...t }: JasmineInterface) {
 
     it('should throw a TypeError when cloning a response with a locked body', async () => {
       const resp = await fetch('https://httpbin.io/get');
-      resp.body!.getReader();
+      requireNotNull(resp.body).getReader();
       let error: TypeError | null = null;
       try {
         resp.clone();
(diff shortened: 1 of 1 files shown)


apps/test-suite/tests/Fetch.ts
```diff
@@ -55,13 +57,13 @@ export function test({ describe, expect, it, ...t }) {
     it('should process response in readablestream from late get reader call', async () => {
       const resp = await fetch('https://httpbin.io/get');
       expect(resp.ok).toBe(true);
-      expect(resp.body).not.toBeNull();
+      expect(resp.body!).not.toBeNull();
```
Kudo: [💡 suggestion] The `!` cancels out the assertion — it tells the compiler the value can't be null on the line that checks it isn't. It isn't needed either: `expect<T>(actual: T)` takes any `T`, which is why `expect(error).not.toBeNull()` at line 457 type-checks unchanged.

```suggestion
      expect(resp.body).not.toBeNull();
```

apps/test-suite/tests/Fetch.ts
```diff
@@ -453,7 +455,7 @@ export function test({ describe, expect, it, ...t }) {
         }
       }
       expect(error).not.toBeNull();
-      expect(error.message).toContain('Fetch request has been canceled');
+      expect(error!.message).toContain('Fetch request has been canceled');
```
Kudo: [💡 suggestion] This file already uses `error?.message` after the same non-null check three times (lines 151, 166 and 203). `?.` also keeps the failure readable: with `!`, a null `error` throws a TypeError instead of failing the matcher.

```suggestion
      expect(error?.message).toContain('Fetch request has been canceled');
```

apps/test-suite/tests/Fetch.ts
```diff
@@ -467,7 +469,7 @@ export function test({ describe, expect, it, ...t }) {
           Accept: 'text/event-stream',
         },
       });
-      const reader = resp.body.getReader();
+      const reader = resp.body!.getReader();
```
Kudo:
```suggestion
      const reader = requireNotNull(resp.body).getReader();
```

not a fan using exclamation mark. maybe introduce a helper like kotlin?

```ts
  export function requireNotNull<T>(value: T | null | undefined, name: string): T {
    if (value == null) {
      throw new Error(`${name} is unexpectedly null`);
    }
    return value;
  }
```

Kudo (changes requested):

apps/test-suite/tests/Fetch.ts
```diff
@@ -61,7 +64,7 @@ export function test({ describe, expect, it, ...t }) {
       await delayAsync(500);
 
       const chunks = [];
-      const reader = resp.body.getReader();
+      const reader = requireNotNull(resp.body, 'resp.body').getReader();
```
Kudo, line 67: it's not ideal for every callsite to specify a name. sorry my previous comment wasn't great.
wdyt for this one?

```ts
export function requireNotNull<T>(value: T | null | undefined): T {
  if (value == null) {
    const error = new Error(`${value} is unexpectedly null`);
    console.error(error.message, error.stack);
    throw error;
  }
  return value;
}
```

Kudo (changes requested):

apps/test-suite/tests/Fetch.ts
vonovak: Yes that's better 👍

Kudo (approved):

PR #48610, opened 2026-08-06
[test-suite] Enable strict and check types in CI

vonovak:
## Why

`strict` was never set for this app, so it followed the compiler default and silently flipped on with TypeScript 6 — and nothing type-checked `apps/test-suite` to notice.

## How

Pin `"strict": true` for explicitness

## Test Plan

Top of the stack: `pnpm tsc` exits 0, down from 651 errors. `lint --max-warnings 0`, `format --check` (108 files) and `test` (9 passed) all pass.

### DIFF
--- a/apps/test-suite/tsconfig.json
+++ b/apps/test-suite/tsconfig.json
@@ -1,9 +1,6 @@
 {
   "extends": "expo/tsconfig.base",
   "compilerOptions": {
-    // `expo/tsconfig.base` does not set this, so it followed the compiler
-    // default and silently flipped on with TypeScript 6. Pin it so the setting
-    // does not change again underneath the app.
     "strict": true,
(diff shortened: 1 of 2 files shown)


apps/test-suite/tsconfig.json
```diff
@@ -1,6 +1,10 @@
 {
   "extends": "expo/tsconfig.base",
   "compilerOptions": {
+    // `expo/tsconfig.base` does not set this, so it followed the compiler
+    // default and silently flipped on with TypeScript 6. Pin it so the setting
+    // does not change again underneath the app.
```
Kudo:
```suggestion
```

redundant code comment. can just comment in pr

Kudo (approved):

PR #48618, opened 2026-08-06
[router] render only focused route for tabs during first render

Ubax:
# Why

All tab routes are currently added to the state via rehydration. I want to remove the rehydration all together in the follow-up PR, so this mechanism needs to be changed. I chose a different approach - during first render rendering only the focused tab, and preloading all tabs with `lazy: false` in an effect. 

When concurrent React integration lands, this will add an additional benefit of faster render when other tabs are heavy and quicker TTI.

# How

- Add placeholder descriptors, which are used when tab is not yet in `state.routes`. This descriptors are keyed by name, and generated using new standard-navigation API - `processDescriptors`
- Render only the routes in `state.routes` and use placeholder descriptors to render correct tabs in tab bar for other tabs
- Dispatch `preload` during first render of the navigator
- Adjust the tests for one more render (only the screen wrapper re-renders. the content stays without rerendering)

# Test Plan

1. CI
2. Router e2e

# Checklist

- [ ] This diff will work correctly for `npx expo prebuild` & EAS Build (not applicable: JavaScript-only router internals)
- [ ] Conforms with the [Documentation Writing Style Guide](https://github.com/expo/expo/blob/main/guides/Expo%20Documentation%20Writing%20Style%20Guide.md) (not applicable: no documentation changes)


packages/expo-router/src/react-navigation/core/useDescriptors.tsx
```diff
@@ -258,17 +254,36 @@ export function useDescriptors<
...
+    (acc, [name, config]) => {
+      const route = {
+        key: undefined,
+        name,
+        params: config.props.initialParams,
+      };
+
+      acc[name] = {
+        route,
+        options: getOptions(route, undefined, {}),
```
jakub-agent: [🚨 **critical**] Placeholder descriptors evaluate the user's `options` / `screenOptions` callbacks with `navigation: undefined`, and those options are what the tab bar actually renders for a tab that has no route yet.

```tsx
<Tabs.Screen
  name="second"
  options={({ navigation }) => ({
    tabBarButton: (props) => (
      <Pressable {...props} onPress={() => navigation.navigate('second')} />
    ),
  })}
/>
```

On first render `second` has no route, so `BottomTabBar` reads `descriptors['second']` — the placeholder — and renders `options.tabBarButton` (`BottomTabBar.tsx:390`). Tapping it throws `Cannot read properties of undefined (reading 'navigate')`. A callback that reads `navigation` eagerly (`navigation.getState()`, `navigation.isFocused()`) throws during render instead.

The public types don't surface this: `DefaultNavigatorOptions.screenOptions`, `RouteConfigProps.options` and `BottomTabOptionsArgs` were all updated to `DescriptorRouteProp` for `route` but still declare `navigation` as present, so TypeScript gives no warning. Either widen those to `navigation: Navigation | undefined` (mirroring what you did for `route`) so consumers are forced to handle it, or hand placeholders a usable navigation object. The `navigation: any | undefined` on `ScreenOptionsOrCallback` (line 58) collapses to `any` and doesn't help either.

packages/expo-router/src/react-navigation/routers/TabRouter.tsx
```diff
@@ -413,16 +436,28 @@ export function TabRouter({
...
+          let routes = state.routes;
+          let index = routes.findIndex((route) => route.name === action.payload.name);
 
           if (index === -1) {
-            return null;
+            const route = createRouteFromAction({ action, routeParamList });
+            routes = [
+              ...routes,
+              action.type === 'NAVIGATE' && action.payload.path != null
+                ? { ...route, path: action.payload.path }
```
jakub-agent: [🏗️ **design**] `NAVIGATE` now mints the route, which breaks a downstream assumption in `NativeBottomTabsRouter` (unchanged by this PR, so I can't comment there):

```ts
const newStateFromNavigation = tabRouter.getStateForAction(state, action, options);
const index = state.routes.findIndex((route) => route.name === action.payload.name);

if (index === -1 || !newStateFromNavigation) {
  return newStateFromNavigation;
}
```

It looks the name up in the **pre-action** `state`. Before this change `index === -1` only meant "name not declared", and `TabRouter` returned `null` for that anyway. Now it is the normal outcome of a tab's first navigation, so the whole post-processing block is skipped: `INTERNAL_EXPO_ROUTER_NO_ANIMATION_PARAM_NAME` is not set for a nested `screen` navigation, and the zoom-transition params are not stripped — the case the comment there warns about ("a view is missing on one tab because it was used to perform zoom transition on another tab").

The layout-effect preload usually mints native tab routes before the user can act, but an initial-URL or `useEffect` dispatch can land first. `newStateFromNavigation.routes.findIndex(...)` would be robust regardless.

packages/expo-router/src/native-tabs/NativeBottomTabsNavigator.tsx
```diff
@@ -71,21 +78,34 @@ function NativeTabsContent({
...
-    () => visibleTabs.map((tab) => tab.routeKey).join(';'),
+  const visibleTabNames = useMemo(
+    () => visibleTabs.map((tab) => tab.name).join(';'),
     [visibleTabs]
   );
 
+  useLayoutEffect(() => {
+    for (const route of visibleRoutes) {
+      if (route.key === undefined) {
+        preload(route.name);
```
jakub-agent: [🏗️ **design**] Native tabs pay the cost of lazy routes without getting the benefit. The updated expectations in `render.test.ios.tsx` show it directly: `TabsScreen` 2 → 4, `TabsHost` 1 → 2, and 1 → 2 in the `Misc` block. Because native tabs always render every tab, this effect immediately dispatches `PRELOAD` for every placeholder, so mounting now costs a full extra render of every tab plus an extra native host commit — where before it was one pass.

It's also load-bearing for correctness, not just for rendering: `onTabChange` bails out with a "most likely a bug in expo-router" warning when `routes.find((route) => route.name === selectedKey)` misses, so a tab press only works once this effect has landed.

Seeding all declared routes in `NativeBottomTabsRouter.getInitialState` would keep native tabs on one render pass and let the lazy path stay where it pays off (JS tabs, drawer). The same argument applies to material-top-tabs, which passes `lazyByDefault: false` and therefore also preloads everything right after mount.

packages/expo-router/src/react-navigation/bottom-tabs/types.tsx
```diff
@@ -92,7 +93,8 @@ export type BottomTabOptionsArgs<
   ParamList extends ParamListBase,
   RouteName extends keyof ParamList = keyof ParamList,
   NavigatorID extends string | undefined = undefined,
-> = BottomTabScreenProps<ParamList, RouteName, NavigatorID> & {
+> = Omit<BottomTabScreenProps<ParamList, RouteName, NavigatorID>, 'route'> & {
+  route: DescriptorRouteProp<ParamList, RouteName>;
```
jakub-agent, line 97: [🏗️ **design**] `BottomTabOptionsArgs` is the only options-arg type that got `DescriptorRouteProp`. `DrawerOptionsArgs` (`drawer/types.tsx:302`) and `MaterialTopTabOptionsArgs` (`material-top-tabs/types.tsx:88`) still spread `*ScreenProps` unchanged, so they promise `route.key: string` while both navigators now receive placeholder descriptors with `key: undefined` — the drawer via `processState` in `DrawerClient`, top tabs via `TopTabsClient`. Worth updating all three together so the type surface is consistent.

packages/expo-router/src/standard-navigation/useVisibleTabsWithRedirect.ts
```diff
@@ -57,7 +74,7 @@ export function useVisibleTabsWithRedirect<
     () => visibleRoutes.findIndex((route) => route.key === focusedRouteKey),
     [focusedRouteKey, visibleRoutes]
   );
-  const focusedIndex = visibleFocusedIndex >= 0 ? visibleFocusedIndex : 0;
+  const focusedIndex = visibleFocusedIndex;
```
jakub-agent: [💡 suggestion] Dropping the `>= 0 ? : 0` clamp means the four navigators now render `null` whenever the focused route isn't a visible tab, but the recovery path below is gated on `isFocused`. For a nested tabs navigator sitting inside an unfocused parent tab whose own focused route is a hidden trigger (or has no trigger), `visibleFocusedIndex` stays `-1`, the redirect never fires, and the navigator renders nothing indefinitely — previously it fell back to the first tab.

The common case (hidden `index`, navigator focused) is covered and actually improves — `render.test.ios.tsx` drops `NativeTabsView` from 2 renders to 1. But I don't see a test for `focusedIndex < 0` while the navigator is unfocused. Either add one to pin the intended behaviour, or run the redirect regardless of focus.

packages/expo-router/src/native-tabs/NativeBottomTabsNavigator.tsx
```diff
@@ -96,7 +116,7 @@ function NativeTabsContent({
         // so the provenance counter is left untouched.
         emitter.emit({
           type: 'tabPress',
-          target: selectedKey,
+          target: routes.find((route) => route.name === selectedKey)?.key,
```
jakub-agent: [💡 suggestion] `target` is `undefined` when the disabled tab has no route yet, and `useEventEmitter.emit` treats a missing `target` as a broadcast — it concatenates the listeners of *every* target (`useEventEmitter.tsx:79-84`). So a prevented press on a not-yet-minted tab fires `tabPress` with `isPrevented: true` on every other tab's listeners.

This is the same failure mode the new `a focused tabPress listener does not intercept an unvisited tab` test guards on the JS side (where `includePlaceholderRoutes` avoids it by assigning `key: name`). The window here is narrow, since the layout effect mints routes on mount, but the path is unguarded — bail out when the lookup misses, or fall back to `selectedKey`.

packages/expo-router/src/react-navigation/core/useDescriptors.tsx
```diff
@@ -258,17 +254,36 @@ export function useDescriptors<
...
+    Descriptor<
+      ScreenOptions,
+      NavigationProp<ParamListBase, string, string | undefined, State, ScreenOptions, EventMap> &
+        ActionHelpers,
+      RouteProp<ParamListBase>
     >
-  >((acc, route, i) => {
+  >;
+
+  const placeholderDescriptors = Object.entries(screens).reduce<DescriptorMap>(
```
jakub-agent: [💡 suggestion] Placeholder descriptors are built for every navigator, but only tab-shaped ones consume them. A `Stack` over a directory with 30 routes now runs `getOptions` 30 extra times per render and allocates 30 descriptors nothing reads — `createNativeStackNavigator` even copies them through its `Object.keys(fullDescriptors)` loop (`fork/native-stack/createNativeStackNavigator.tsx:68`).

Since the crash surface in my other comment scales with this too, gating it (a `useNavigationBuilder` option, or building it lazily via a getter) would contain both the cost and the blast radius.

packages/expo-router/src/native-tabs/NativeBottomTabsNavigator.tsx
```diff
@@ -71,21 +78,34 @@ function NativeTabsContent({
...
-      ),
+      visibleRoutes
+        .map((route) => ({
+          ...route,
+          key: route.key ?? route.name,
+        }))
+        .map(
+          (route): NativeTabsViewTabItem => ({
+            options: descriptors[route.key]!.options,
+            routeKey: route.name,
```
jakub-agent: [💡 suggestion] `routeKey` and `name` now always hold the same value, and `routeKey` flows on to `screenKey` / `selectedScreenKey` in `NativeTabsView.shared.tsx` — so three names in the public `NativeTabsViewTabItem` type say "key" while carrying a route name. The updated assertions in `render.test.ios.tsx` (`/^index-[-\w]+/` → `'index'`) make the shift visible.

Per the package's minimal style, dropping `routeKey` and using `name` at the call sites would remove the duplicate field and the misleading name in one go.

packages/expo-router/src/standard-navigation/index.tsx
```diff
@@ -190,13 +190,18 @@ export function unstable_integrateWithRouter<
 
     const { dispatch } = navigation;
 
+    const processedState = useMemo(
+      () => options?.processState?.(state, descriptors) ?? state,
```
jakub-agent: [💡 suggestion] This memo never caches: `useDescriptors` returns `{ ...placeholderDescriptors, ...routeDescriptors }`, a fresh object on every render, so `descriptors` always changes identity. `processedState` is recomputed every render, which invalidates `derivedProps` below, which gives `preload` a new identity, which makes the `visibleRoutes` / `visibleTabs` memos and both preload layout effects rerun every render too.

It converges (the effects no-op once `route.key` is set), so this isn't a correctness bug — but the memos read as if they prevent work and don't. Worth either keying on something stable or dropping them, especially given the package's "works with and without React Compiler" requirement.

packages/expo-router/src/standard-navigation/types.ts
```diff
@@ -88,6 +88,33 @@ export type IntegrateWithRouterOptions<
...
+   *
+   *   return {
+   *     ...state,
+   *     index: Math.max(0, routes.findIndex((route) => route.key === focusedKey)),
+   *     routes,
+   *   };
+   * }
+   * ```
+   */
+  processState?: (
```
jakub-agent, line 117: [🐧 nit] The `@example` block is a verbatim copy of `includePlaceholderRoutes`, the only implementation, and the inline `descriptors` parameter type is duplicated word-for-word in both files. Since exactly one function is passed here by three navigators, referencing `includePlaceholderRoutes` from the docs and extracting the parameter type into a shared alias would keep the two from drifting.

packages/expo-router/src/standard-navigation/useVisibleTabsWithRedirect.ts
```diff
@@ -45,10 +47,25 @@ export function useVisibleTabsWithRedirect<
...
+              params: descriptors[name]?.route?.params,
+            })),
+          ...routes,
+        ],
+        routeNames
+      ).filter((route): route is Route => {
         // Every filesystem route is registered in state; only routes declared by a non-hidden
         // trigger become tab items.
-        const descriptor = descriptors[route.key];
+        if (!route) {
```
jakub-agent: [🐧 nit] `orderRoutesByRouteNames` builds its result with `flatMap` over `routeNames` and returns `[]` for a name it can't find, so it never yields `undefined` entries. This guard is unreachable, and without it the `route is Route` predicate isn't needed either — a plain `.filter()` narrows the union fine.

jakub-agent (changes requested):

packages/expo-router/src/standard-navigation/includePlaceholderRoutes.ts
```diff
@@ -0,0 +1,24 @@
+import type { NavigationState } from '../react-navigation/native';
+
+export function includePlaceholderRoutes<State extends NavigationState>(
+  state: State,
+  descriptors: Record<
+    string,
+    { route?: { key: string | undefined; name: string; params?: object | undefined } }
+  >
```
Ubax: Can you use already created type here? Like DescriptorMap?

packages/expo-router/src/standard-navigation/includePlaceholderRoutes.ts
```diff
@@ -0,0 +1,24 @@
+import type { NavigationState } from '../react-navigation/native';
+
+export function includePlaceholderRoutes<State extends NavigationState>(
```
Ubax: Rename this hook to suggest that it should only be used for tabs navigator. Change the name to suggest that it appends routes which are missing from `routeNames`

packages/expo-router/src/standard-navigation/includePlaceholderRoutes.ts
```diff
@@ -0,0 +1,24 @@
...
+  state: State,
+  descriptors: Record<
+    string,
+    { route?: { key: string | undefined; name: string; params?: object | undefined } }
+  >
+): State {
+  const focusedKey = state.routes[state.index]?.key;
+  const routes = state.routeNames.map((name) => {
+    const route = state.routes.find((route) => route.name === name) ?? descriptors[name]?.route;
+    return { ...route, key: route?.key ?? name, name } as State['routes'][number];
```
Ubax:
```suggestion
    const existingRoute = state.routes.find((route) => route.name === name)
    if(existingRoute) {
      return existingRoute
    }
    // Extract the logic below to separate function in this file
    const descriptor = descriptors[name]
    if(!descriptor){ throw error suggesting internal expo-router issue }
    return { ...descriptor.route, key: name, name } as State['routes'][number];
```

packages/expo-router/src/standard-navigation/includePlaceholderRoutes.ts
```diff
@@ -0,0 +1,24 @@
...
+  });
+
+  return {
+    ...state,
+    index: Math.max(
+      0,
+      routes.findIndex((route) => route.key === focusedKey)
+    ),
+    routes,
+  };
```
Ubax:
```suggestion
  const index = ....
  return {
    ...state,
    index,
    routes,
  };
```

packages/expo-router/src/native-tabs/__tests__/options.test.ios.tsx
```diff
@@ -666,7 +666,7 @@ describe('Tab options', () => {
 
       expect(screen.getByTestId('index')).toBeVisible();
       expect(NativeTabsView).toHaveBeenCalledTimes(1);
-      const call = NativeTabsView.mock.calls[0]![0];
+      const call = NativeTabsView.mock.calls.at(-1)![0];
```
Ubax: Don't change if `.toHaveBeenCalledTimes(1);`

packages/expo-router/src/native-tabs/__tests__/options.test.ios.tsx
```diff
@@ -639,8 +639,8 @@ describe('Label', () => {
       one: () => <View testID="one" />,
     });
     expect(screen.getByTestId('index')).toBeVisible();
-    expect(NativeTabsView).toHaveBeenCalledTimes(1);
-    const call = NativeTabsView.mock.calls[0]![0];
+    expect(NativeTabsView).toHaveBeenCalledTimes(2);
+    const call = NativeTabsView.mock.calls.at(-1)![0];
```
Ubax:
```suggestion
    const call = NativeTabsView.mock.calls.at[1]![0];
```

Same in other places where `.toHaveBeenCalledTimes(2);`

packages/expo-router/src/native-tabs/NativeBottomTabsNavigator.tsx
```diff
@@ -71,21 +78,34 @@ function NativeTabsContent({
...
+            name: route.name,
+            contentRenderer: () => descriptors[route.key]?.render?.() ?? null,
+          })
+        ),
```
Ubax: Is this not already fixed by `includePlaceholderRoutes`?

packages/expo-router/src/native-tabs/NativeBottomTabsNavigator.tsx
```diff
@@ -71,21 +78,34 @@ function NativeTabsContent({
...
     [descriptors, visibleRoutes]
   );
-  const visibleTabsKeys = useMemo(
-    () => visibleTabs.map((tab) => tab.routeKey).join(';'),
+  const visibleTabNames = useMemo(
+    () => visibleTabs.map((tab) => tab.name).join(';'),
     [visibleTabs]
   );
 
+  useLayoutEffect(() => {
```
Ubax: Why not `usePreloadPlaceholderRoutes`?

packages/expo-router/src/native-tabs/NativeBottomTabsNavigator.tsx
Ubax: Is it possible to do it in a way where the rendered screen would not rerender?

The benefit is that after the refactor all the tabs except the first rendered one will load using `startTransition`, so the first screen will be interactive faster. Does it hold then?

packages/expo-router/src/native-tabs/NativeBottomTabsNavigator.tsx
Ubax: Keep the old `target: selectedKey,`

packages/expo-router/src/native-tabs/NativeBottomTabsNavigator.tsx
```diff
@@ -109,7 +129,7 @@ function NativeTabsContent({
       provenanceRef.current = provenance;
 
       if (isNativeAction) {
-        const selectedRoute = routes.find((route) => route.key === selectedKey);
+        const selectedRoute = routes.find((route) => route.name === selectedKey);
```
Ubax, line 130: Revert after using `includePlaceholderRoutes`

packages/expo-router/src/native-tabs/NativeBottomTabsNavigator.tsx
```diff
@@ -121,7 +141,7 @@ function NativeTabsContent({
         }
         emitter.emit({
           type: 'tabPress',
-          target: selectedKey,
+          target: selectedRoute.key,
```
Ubax, line 142: Same here

packages/expo-router/src/native-tabs/NativeBottomTabsNavigator.tsx
```diff
@@ -71,21 +78,34 @@ function NativeTabsContent({
...
+            routeKey: route.name,
+            name: route.name,
+            contentRenderer: () => descriptors[route.key]?.render?.() ?? null,
+          })
+        ),
     [descriptors, visibleRoutes]
   );
-  const visibleTabsKeys = useMemo(
-    () => visibleTabs.map((tab) => tab.routeKey).join(';'),
+  const visibleTabNames = useMemo(
```
Ubax, line 95: When `includePlaceholderRoutes` is used, this change in no longer needed

packages/expo-router/src/react-navigation/bottom-tabs/views/BottomTabBar.tsx
```diff
@@ -367,14 +368,11 @@ export function BottomTabBar({ state, descriptors, emitter, navigateToTab, inset
                 : undefined;
 
           return (
-            <NavigationProvider
-              key={route.key}
-              route={route}
-              navigation={descriptors[route.key]!.navigation}>
+            <NavigationProvider key={route.name} route={route} navigation={descriptor.navigation}>
```
Ubax, line 371: Move this change to the upper PR in the stack

packages/expo-router/src/react-navigation/bottom-tabs/views/BottomTabView.tsx
```diff
@@ -231,8 +231,11 @@ export function BottomTabView(props: Props) {
           const isFocused = state.index === index;
           const isPreloaded = preloadedRouteKeys.includes(route.key);
 
-          if (lazy && !loaded.includes(route.key) && !isFocused && !isPreloaded) {
-            // Don't render a lazy screen if we've never navigated to it or it wasn't preloaded
+          if (
+            descriptor.route.key === undefined ||
+            (lazy && !loaded.includes(route.key) && !isFocused && !isPreloaded)
```
Ubax, line 236: Move this up in the stack

packages/expo-router/src/react-navigation/bottom-tabs/views/BottomTabView.tsx
```diff
@@ -271,7 +274,7 @@ export function BottomTabView(props: Props) {
 
           return (
             <MaybeScreen
-              key={route.key}
+              key={route.name}
```
Ubax: Revert, as this is no longer needed

packages/expo-router/src/react-navigation/core/__tests__/useDescriptors.test.ios.tsx
```diff
@@ -14,6 +14,50 @@ beforeEach(() => {
...
+  };
+
+  render(
+    <BaseNavigationContainer>
+      <TestNavigator>
+        <Screen name="foo" component={React.Fragment} options={{ title: 'Foo' }} />
+        <Screen name="bar" component={React.Fragment} options={{ title: 'Bar' }} />
+      </TestNavigator>
+    </BaseNavigationContainer>
+  );
```
Ubax, line 43: Can you use `renderHook` instead?

packages/expo-router/src/react-navigation/core/useDescriptors.tsx
```diff
@@ -254,17 +254,36 @@ export function useDescriptors<
...
-  const descriptors = cachedRoutes.reduce<
-    Record<
-      string,
-      Descriptor<
-        ScreenOptions,
-        NavigationProp<ParamListBase, string, string | undefined, State, ScreenOptions, EventMap> &
-          ActionHelpers,
-        RouteProp<ParamListBase>
-      >
+  type DescriptorMap = Record<
```
Ubax, line 266: Add TODO to unify the `DescriptorMap` types

packages/expo-router/src/react-navigation/core/useDescriptors.tsx
Ubax: Instead of setting the navigation to undefined, change the implementation of `/react-navigation/core/useNavigationCache.tsx`.

There create a new navigation for each route in routeName, so that it is available for placeholders as well.

packages/expo-router/src/react-navigation/core/useDescriptors.tsx
Ubax: Add TODO to address this in follow-up and decide wether it is worth it

packages/expo-router/src/react-navigation/core/useDescriptors.tsx
Ubax: Then add unit test for
```
<Tabs.Screen
  name="second"
  options={({ navigation }) => ({
    tabBarButton: (props) => (
      <Pressable {...props} onPress={() => navigation.navigate('second')} />
    ),
  })}
/>
```

packages/expo-router/src/react-navigation/drawer/views/DrawerItemList.tsx
```diff
@@ -63,7 +63,7 @@ export function DrawerItemList({ state, navigation, descriptors }: Props) {
 
     return (
       <DrawerItem
-        key={route.key}
+        key={route.name}
```
Ubax: Revert

packages/expo-router/src/react-navigation/drawer/views/DrawerView.tsx
```diff
@@ -210,8 +210,11 @@ function DrawerViewBase({
...
           const isPreloaded = state.preloadedRouteKeys.includes(route.key);
 
-          if (lazy && !loaded.includes(route.key) && !isFocused && !isPreloaded) {
-            // Don't render a lazy screen if we've never navigated to it or it wasn't preloaded
+          if (
+            descriptor.route.key === undefined ||
+            (lazy && !loaded.includes(route.key) && !isFocused && !isPreloaded)
+          ) {
+            // Don't render placeholder or unloaded lazy screens.
             return null;
```
Ubax, line 218: Move to up PR in stack

packages/expo-router/src/react-navigation/drawer/views/DrawerView.tsx
```diff
@@ -242,7 +245,7 @@ function DrawerViewBase({
 
           return (
             <MaybeScreen
-              key={route.key}
+              key={route.name}
```
Ubax: Revert

packages/expo-router/src/react-navigation/routers/TabRouter.tsx
```diff
@@ -231,17 +249,18 @@ export function TabRouter({
...
-
+          ? initialRouteName
+          : routeNames[0]!;
+      const routes = [
+        {
+          name: initialName,
+          key: `${initialName}-${nanoid()}`,
+          params: routeParamList[initialName],
+        },
+      ];
```
Ubax: Use `addFallbackRouteIfEmpty`

packages/expo-router/src/react-navigation/routers/TabRouter.tsx
```diff
@@ -263,32 +282,35 @@ export function TabRouter({
...
+                        ...routeParamList[route.name],
+                        ...route.params,
+                      }
+                    : route.params,
+              }) as Route<string>
+          ),
+        routeNames,
+        routeParamList,
+        initialRouteName
       );
```
Ubax, line 314: Split this into two steps 
1. filter + map
2. addFallbackRouteIfEmpty

packages/expo-router/src/react-navigation/routers/TabRouter.tsx
```diff
@@ -263,32 +282,35 @@ export function TabRouter({
...
-              : route
-                ? route.params
-                : undefined,
-        } as Route<string>;
-      });
-
-      const index = Math.min(
-        Math.max(routeNames.indexOf(state.routes[state?.index ?? 0]?.name as string), 0),
-        routes.length - 1
+      const partialRoutes = (state as PartialState<TabNavigationState<ParamListBase>>).routes;
```
Ubax, line 295: Is this cast needed?

packages/expo-router/src/react-navigation/routers/TabRouter.tsx
```diff
@@ -263,32 +282,35 @@ export function TabRouter({
...
+        routeNames,
+        routeParamList,
+        initialRouteName
       );
 
+      const focusedName = state.routes[state.index ?? 0]?.name;
+      const focusedIndex = routes.findIndex((route) => route.name === focusedName);
+      const fallbackName = routeNames.find((name) => routes.some((route) => route.name === name));
+      const index =
+        focusedIndex < 0 ? routes.findIndex((route) => route.name === fallbackName) : focusedIndex;
```
Ubax:
```suggestion
      const fallbackIndex = routes.findIndex((route) => routeNames.includes(route.name));
      const index = focusedIndex < 0 ? fallbackIndex : focusedIndex;
```

packages/expo-router/src/react-navigation/routers/TabRouter.tsx
```diff
@@ -334,21 +356,22 @@ export function TabRouter({
...
+          const focusedKey = state.routes[state.index]?.key;
           const focusedIndex = routes.findIndex((route) => route.key === focusedKey);
-          const index = Math.max(0, focusedIndex);
+          const fallbackName = routeNames.find((name) =>
+            routes.some((route) => route.name === name)
+          );
+          const index =
+            focusedIndex < 0
+              ? routes.findIndex((route) => route.name === fallbackName)
+              : focusedIndex;
```
Ubax: Same as above

packages/expo-router/src/react-navigation/routers/TabRouter.tsx
```diff
@@ -512,6 +549,57 @@ export function TabRouter({
...
+                initialRouteName
+              );
+              return {
+                ...result,
+                preloadedRouteKeys: result.preloadedRouteKeys.filter(
+                  (key) => key !== result.routes[result.index]!.key
+                ),
+              };
+            }
+          }
```
Ubax, line 603: Can this be extracted and combined with `NAVIGATE`/`JUMP_TO`?
It essentially is a navigate action

packages/expo-router/src/react-navigation/routers/TabRouter.tsx
Ubax, line 255: Add a TODO to simplify the actions logic in `TabRouter`

packages/expo-router/src/standard-navigation/__tests__/useBuildHref.integration.test.ios.tsx
```diff
@@ -53,6 +53,8 @@ describe('useBuildHref (integration)', () => {
         Record<string, never>
       >;
 
+    act(() => lastArgs().actions.navigate('(group)/feed'));
```
Ubax: Why was this added?

packages/expo-router/src/standard-navigation/types.ts
Ubax: Change the exaplme to be simpler

packages/expo-router/src/standard-navigation/types.ts
```diff
@@ -88,6 +88,33 @@ export type IntegrateWithRouterOptions<
...
+    state: State,
+    descriptors: Record<
+      string,
+      { route?: { key: string | undefined; name: string; params?: object | undefined } }
+    >
```
Ubax: Can we reuse `DescriptorMap` here?

packages/expo-router/src/standard-navigation/usePreloadPlaceholderRoutes.ts
```diff
@@ -0,0 +1,28 @@
...
+export function usePreloadPlaceholderRoutes({
+  routes,
+  descriptors,
+  preload,
+  lazyByDefault,
+}: {
+  routes: Route[];
+  descriptors: Record<string, Descriptor | undefined>;
+  preload: (name: string) => void;
+  lazyByDefault: boolean;
```
Ubax, line 19: Add a comment explaining this prop or rename it

packages/expo-router/src/standard-navigation/useVisibleTabsWithRedirect.ts
```diff
@@ -45,10 +47,25 @@ export function useVisibleTabsWithRedirect<
...
+          ...routeNames
+            .filter((name) => !routes.some((r) => r.name === name))
+            .map((name) => ({
+              key: undefined,
+              name,
+              params: descriptors[name]?.route?.params,
+            })),
+          ...routes,
+        ],
+        routeNames
```
Ubax: Split into functions or steps

packages/expo-router/src/native-tabs/NativeBottomTabsNavigator.tsx
```diff
@@ -109,7 +127,7 @@ function NativeTabsContent({
       provenanceRef.current = provenance;
 
       if (isNativeAction) {
-        const selectedRoute = routes.find((route) => route.key === selectedKey);
+        const selectedRoute = routes.find((route) => route.name === selectedKey);
```
jakub-agent, line 130: [🚨 **critical**] The `isPrevented` branch just above (line 117) still passes `target: selectedKey`, but `selectedKey` changed meaning in this PR — `NativeTabsView.shared.tsx` now sends `screenKey: name`, so it is a route **name**, not a route key.

`useNavigationCache` registers a rendered screen's emitter listeners under its **real** route key (`emitter.create(route.key)`), and `useEventEmitter.emit` matches `target` exactly (`items[target]?.slice()`). So a `navigation.addListener('tabPress', …)` inside a disabled tab's screen never fires.

This is not hypothetical — the test in this PR fails:

```
src/native-tabs/__tests__/events.test.ios.tsx
✕ emits tabPress with isPrevented and does not navigate when a disabled tab is tapped

  expect(jest.fn()).toHaveBeenCalledTimes(expected)
  Expected number of calls: 1
  Received number of calls: 0
    at src/native-tabs/__tests__/events.test.ios.tsx:356:33
```

The same file passes on the stack base `e735aafdb9`. The line you're on already does the right lookup — the prevented branch needs the same:

```ts
target: routes.find((route) => route.name === selectedKey)?.key ?? selectedKey,
```

The `?? selectedKey` keeps the not-yet-minted case working, since a placeholder's navigation is cached under the name.

The same name-vs-key mismatch has a second effect: in `useNavigationBuilder`'s `listen`, a `target` that is a name misses `state.routes` and hits the new `screens[e.target]` branch, so `isPlaceholder` is `true` even when the route exists. A `<NativeTabs.Trigger listeners={…}>` on an already-created tab then receives the placeholder navigation instead of the screen's real one. Fixing `target` here fixes that too.

packages/expo-router/src/standard-navigation/usePreloadPlaceholderRoutes.ts
```diff
@@ -0,0 +1,31 @@
...
+  preload,
+  lazyByDefault,
+}: {
+  routes: Route[];
+  descriptors: Record<string, Descriptor | undefined>;
+  preload: (name: string) => void;
+  /** Used when a route does not specify `options.lazy`. */
+  lazyByDefault: boolean;
+}) {
+  useLayoutEffect(() => {
```
jakub-agent, line 21: [🏗️ **design**] Answering your question on the last pass — *"all the tabs except the first rendered one will load using `startTransition`, so the first screen will be interactive faster. Does it hold then?"*

**Not as written.** `useLayoutEffect` runs synchronously after the commit, and the `PRELOAD` dispatch it fires is a synchronous, non-transition update. React re-renders before the frame is handed to the host, so the intermediate placeholder tree is never presented and nothing becomes interactive earlier — the render counts in `render.test.ios.tsx` (`TabsScreen` 2 → 4, `TabsHost` 1 → 2) are pure overhead for native tabs and for material-top-tabs, which both pass `lazyByDefault: false` and therefore preload everything immediately.

Two things would have to change for the benefit to be real:

1. Move the preload to a passive `useEffect` and wrap the dispatch in `startTransition`, so React can commit the first tab and yield before mounting the rest.
2. Note that placeholders already `render()` to `null`, so the first commit contains no *screen* work to defer — only the tab shells. The win is bounded by how much of the cost sits in the shells versus the screens; worth measuring before optimizing for it.

If you'd rather not chase the transition, the cheaper option for the always-eager navigators is to keep them off the lazy path entirely (seed all declared routes in `NativeBottomTabsRouter.getInitialState`) and reserve placeholders for JS tabs and drawer, where `lazy` defaults to `true` and the saving is real.

packages/expo-router/src/standard-navigation/useVisibleTabsWithRedirect.ts
```diff
@@ -43,21 +45,28 @@ export function useVisibleTabsWithRedirect<
...
+      const descriptor = descriptors[route.key ?? route.name];
+      return isDeclaredInLayout(descriptor) && descriptor?.options?.hidden !== true;
+    });
+  }, [routes, routeNames, descriptors]);
   const visibleFocusedIndex = useMemo(
     () => visibleRoutes.findIndex((route) => route.key === focusedRouteKey),
     [focusedRouteKey, visibleRoutes]
   );
-  const focusedIndex = visibleFocusedIndex >= 0 ? visibleFocusedIndex : 0;
+  const focusedIndex = visibleFocusedIndex;
```
jakub-agent: [💡 suggestion] Carried over from the last pass, still unaddressed. Dropping the `>= 0 ? : 0` clamp means the four navigators render `null` whenever the focused route isn't a visible tab, but the recovery `router.replace` below is gated on `isFocused`. A nested tabs navigator inside an unfocused parent tab, whose own focused route is a hidden trigger or has no trigger, stays at `-1` forever and renders nothing — it used to fall back to the first tab.

The focused case is covered and improves (`render.test.ios.tsx` drops `NativeTabsView` from 2 renders to 1). I still don't see a test for `focusedIndex < 0` while the navigator is unfocused — worth one either way, to pin whether "render nothing" is the intended answer there.

packages/expo-router/src/standard-navigation/useVisibleTabsWithRedirect.ts
```diff
@@ -43,21 +45,28 @@ export function useVisibleTabsWithRedirect<
...
-      orderRoutesByRouteNames(routes, routeNames).filter((route) => {
-        // Every filesystem route is registered in state; only routes declared by a non-hidden
-        // trigger become tab items.
-        const descriptor = descriptors[route.key];
-        return isDeclaredInLayout(descriptor) && descriptor?.options?.hidden !== true;
-      }),
-    [routes, routeNames, descriptors]
-  );
+  const visibleRoutes = useMemo(() => {
+    const missingRoutes = routeNames
```
jakub-agent: [💡 suggestion] This duplicates `appendMissingPlaceholderTabRoutes` — same `{ key: name, name, params: descriptors[name]?.route?.params }` shape, same `key = name` convention, two places to keep in sync.

JS tabs, top tabs, drawer and native tabs all go through `processState`, so by the time they call this hook `state.routes` already covers every `routeName` and `missingRoutes` is always empty. The only caller that needs it is headless `ui/Tabs`, which already calls `appendMissingPlaceholderTabDescriptors` — having it also call `appendMissingPlaceholderTabRoutes` on the state it passes in would let this block go away and keep the convention in one file.

Related: with `key` now always set, the `?? route.name` in `descriptors[route.key ?? route.name]` below is dead.

packages/expo-router/src/react-navigation/core/useNavigationCache.tsx
```diff
@@ -71,107 +73,116 @@ export function useNavigationCache<
...
-        if (isRoutePreloadedInStack(state, route)) {
-          if (process.env.NODE_ENV !== 'production') {
-            console.warn(
-              `Ignored a navigation action dispatched from the preloaded screen '${route.name}'. The screen is rendered for preloading and is not focused, so its actions would unexpectedly modify the visible stack. Wait until the screen is focused before dispatching.`
-            );
+  const cacheRoutes = [
+    ...routes,
+    ...routeNames
+      .filter((name) => !routes.some((route) => route.name === name))
+      .map((name) => ({ key: name, name })),
```
jakub-agent: [💡 suggestion] Placeholder *descriptors* became opt-in and lazy in this PR (`describe` + `processDescriptors`), but placeholder *navigations* are still built eagerly for every navigator. A `Stack` over a directory with many routes allocates a navigation object — `...rest`, `...helpers`, `emitter.create`, four closures — for every declared name that isn't in `state.routes`, and only tab-shaped navigators ever read them.

They're cached across renders, so it's a one-time cost rather than a per-render one, which is why I'd call this a cleanup rather than a problem. Still, it's the same asymmetry the `describe` refactor removed on the descriptor side; a lazy factory consulted from `describe` would make the two consistent.

packages/expo-router/src/react-navigation/routers/TabRouter.tsx
```diff
@@ -263,32 +292,32 @@ export function TabRouter({
...
+              : route.params,
+        }));
+      const routes = addFallbackRouteIfEmpty(
+        filteredRoutes,
+        routeNames,
+        routeParamList,
+        initialRouteName
       );
 
+      const focusedName = state.routes[state.index ?? 0]?.name;
```
jakub-agent, line 316: [🐧 nit] `fallbackIndex` can only ever be `0` or `-1`: `routes` is `filteredRoutes` (already filtered to `routeNames`) plus `addFallbackRouteIfEmpty`'s fallback (also from `routeNames`), so the first element always satisfies the predicate. The same expression appears in `getStateForRouteNamesChange`, where `routes` is likewise pre-filtered.

That makes it read as if it picks the first route in *declared* order, while it actually picks the first in the incoming state's order — which for `getRehydratedState` is whatever order the partial state had. If declared order is what `falls back to the first surviving route in declared order` means, `routes` needs `orderRoutesByRouteNames` applied first; if not, `routes.length > 0 ? 0 : -1` says the same thing without implying a search.

jakub-agent (changes requested):

packages/expo-router/src/standard-navigation/useVisibleTabsWithRedirect.ts
Ubax: Leave a TODO comment

packages/expo-router/src/standard-navigation/useVisibleTabsWithRedirect.ts
Ubax: Remove this block

packages/expo-router/src/react-navigation/core/useNavigationCache.tsx
Ubax: Change this to cached factory instead - if they key is not in cache create a new entry

packages/expo-router/src/standard-navigation/types.ts
```diff
@@ -88,6 +103,28 @@ export type IntegrateWithRouterOptions<
...
+   * })
+   * ```
+   */
+  processState?: (
+    state: State,
+    descriptors: PlaceholderDescriptorMap,
+    describe: DescribePlaceholderRoute
+  ) => State;
+  /** Creates additional descriptors before `processState` and navigator rendering. */
+  processDescriptors?: (
```
jakub-agent, line 123: [💡 suggestion] `processState` and `processDescriptors` are independent options, but for the tab case they must be used together: `appendMissingPlaceholderTabRoutes` calls `createPlaceholderRoute`, which **throws** when `descriptors[name]` is missing. So `processState: appendMissingPlaceholderTabRoutes` without a matching `processDescriptors` is a runtime crash, not a degraded render.

All four current call sites pair them correctly, so nothing is broken today. But this is a semi-public option surface and the pairing isn't stated anywhere — worth either a line in this TSDoc ("pair with `processState` when the state processor reads placeholder descriptors") or collapsing the two into one option, since there's no case for supplying only one.

packages/expo-router/src/standard-navigation/useVisibleTabsWithRedirect.ts
```diff
@@ -57,7 +58,7 @@ export function useVisibleTabsWithRedirect<
     () => visibleRoutes.findIndex((route) => route.key === focusedRouteKey),
     [focusedRouteKey, visibleRoutes]
   );
-  const focusedIndex = visibleFocusedIndex >= 0 ? visibleFocusedIndex : 0;
+  const focusedIndex = visibleFocusedIndex;
```
jakub-agent: [💡 suggestion] Third time raising this, so treat it as a follow-up rather than a merge condition. Removing the `>= 0 ? : 0` clamp makes the four navigators render `null` when the focused route isn't a visible tab, while the recovery `router.replace` below only runs when `isFocused`. A nested tabs navigator inside an unfocused parent tab whose own focused route is a hidden trigger has no way out of `-1`.

The focused path is covered and measurably better (`render.test.ios.tsx` drops `NativeTabsView` from 2 renders to 1). A test for the unfocused case would pin whether "render nothing" is the intended answer — right now the behaviour is only implied by the absence of a test.

packages/expo-router/src/react-navigation/core/useNavigationCache.tsx
```diff
@@ -71,107 +73,115 @@ export function useNavigationCache<
...
+          );
         }
+        return;
+      }
+
+      const action = typeof thunk === 'function' ? thunk(state) : thunk;
 
-        const action = typeof thunk === 'function' ? thunk(state) : thunk;
+      if (action != null) {
+        navigation.dispatch({ source: route.key, ...action });
```
jakub-agent: [🐧 nit] For a placeholder, `route.key` is the route **name**, so actions dispatched from its navigation carry `source: "<name>"`. `BaseRouter` resolves `SET_PARAMS` / `REPLACE_PARAMS` with `state.routes.findIndex((r) => r.key === action.source)`, which can't match a name, so it returns `null` and the action bubbles up until it's dropped.

Net effect: `navigation.setParams()` called from an unvisited tab's `options` callback silently does nothing. `navigate` is unaffected because `TabRouter` ignores `source`, and a silent no-op beats a crash here, so I don't think this needs changing. Flagging it because it isn't obvious from the code that key-shaped-but-not-a-key `source` values get produced, and a short comment would save the next reader the trace through `BaseRouter`.

jakub-agent (approved):

packages/expo-router/src/react-navigation/core/useNavigationCache.tsx
Ubax:
```suggestion
        // TODO(@ubax): https://github.com/expo/expo/pull/48618#discussion_r3735996416
        navigation.dispatch({ source: route.key, ...action });
```

packages/expo-router/src/standard-navigation/useVisibleTabsWithRedirect.ts
Ubax:
```suggestion
  // TODO(@ubax): https://github.com/expo/expo/pull/48618#discussion_r3735996409
  const focusedIndex = visibleFocusedIndex;
```

PR #48619, opened 2026-08-06
[router] Make `navigation` optional in `Descriptor` type

Ubax:
# Why

# How

# Test Plan

# Checklist

### DIFF
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### 🛠 Breaking changes
 
+- Allow descriptors without a `navigation` object in `expo-router/react-navigation` types.
 - Remove `NavigationContainerRefContext` fallback in `useNavigation`. The hook now throws when called outside a navigator, including components rendered via `ExpoRoot`'s `wrapper` prop. ([#48638](https://github.com/expo/expo/pull/48638) by [@Ubax](https://github.com/Ubax))
 - Make `key` optional in `Descriptor` type ([#48596](https://github.com/expo/expo/pull/48596) by [@Ubax](https://github.com/Ubax))
 - Unify JS Tabs, TopTabs, Drawer, and headless tabs with NativeTabs - only screens declared in the layout become visible. ([#48499](https://github.com/expo/expo/pull/48499) by [@Ubax](https://github.com/Ubax))
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### 🛠 Breaking changes
 
-- Allow descriptors without a `navigation` object in `expo-router/react-navigation` types.
+- Make `navigation` optional in `Descriptor` type ([#48619](https://github.com/expo/expo/pull/48619) by [@Ubax](https://github.com/Ubax))
 - Remove `NavigationContainerRefContext` fallback in `useNavigation`. The hook now throws when called outside a navigator, including components rendered via `ExpoRoot`'s `wrapper` prop. ([#48638](https://github.com/expo/expo/pull/48638) by [@Ubax](https://github.com/Ubax))
 - Make `key` optional in `Descriptor` type ([#48596](https://github.com/expo/expo/pull/48596) by [@Ubax](https://github.com/Ubax))
 - Unify JS Tabs, TopTabs, Drawer, and headless tabs with NativeTabs - only screens declared in the layout become visible. ([#48499](https://github.com/expo/expo/pull/48499) by [@Ubax](https://github.com/Ubax))
--- a/packages/expo-router/src/react-navigation/core/NavigationProvider.tsx
+++ b/packages/expo-router/src/react-navigation/core/NavigationProvider.tsx
@@ -14,7 +14,7 @@ export const NavigationRouteContext = React.createContext<Route<string> | undefi
 
 type Props = {
   route: Route<string>;
-  navigation: NavigationProp<ParamListBase>;
+  navigation: NavigationProp<ParamListBase> | undefined;
   children: React.ReactNode;
 };
 
--- a/packages/expo-router/src/react-navigation/core/types.tsx
+++ b/packages/expo-router/src/react-navigation/core/types.tsx
@@ -584,9 +584,11 @@ export type Descriptor<
   route: DescriptorRoute<Route>;
 
   /**
-   * Navigation object for the screen
+   * Navigation object for the screen. `undefined` when the descriptor
+   * describes a route name declared in a layout that has no live route
+   * instance in navigation state.
    */
-  navigation: Navigation;
+  navigation: Navigation | undefined;
 
   /**
    * Whether this screen was declared in the layout (`<Screen>`/`<NativeTabs.Trigger>`)
@@ -678,7 +680,7 @@ export type RouteConfigProps<
     | ScreenListeners<State, EventMap>
     | ((props: {
         route: RouteProp<ParamList, RouteName>;
-        navigation: Navigation;
+        navigation: Navigation | undefined;
       }) => ScreenListeners<State, EventMap>);
 
   /**
--- a/packages/expo-router/src/react-navigation/core/types.tsx
+++ b/packages/expo-router/src/react-navigation/core/types.tsx
@@ -72,7 +72,7 @@ export type DefaultNavigatorOptions<
     | ScreenListeners<State, EventMap>
     | ((props: {
         route: RouteProp<ParamList>;
-        navigation: Navigation;
+        navigation: Navigation | undefined;
       }) => ScreenListeners<State, EventMap>);
 
   /**
@@ -82,7 +82,7 @@ export type DefaultNavigatorOptions<
     | ScreenOptions
     | ((props: {
         route: DescriptorRouteProp<ParamList>;
-        navigation: Navigation;
+        navigation: Navigation | undefined;
         theme: ReactNavigation.Theme;
       }) => ScreenOptions);
 
@@ -543,7 +543,7 @@ export type ScreenLayoutArgs<
 > = {
   route: RouteProp<ParamList, RouteName>;
   options: ScreenOptions;
-  navigation: Navigation;
+  navigation: Navigation | undefined;
   theme: ReactNavigation.Theme;
   children: React.ReactElement;
 };
@@ -586,7 +586,8 @@ export type Descriptor<
   /**
    * Navigation object for the screen. `undefined` when the descriptor
    * describes a route name declared in a layout that has no live route
-   * instance in navigation state.
+   * instance in navigation state. The property is intentionally required so
+   * descriptor producers must set `navigation: undefined` explicitly.
    */
   navigation: Navigation | undefined;
 
@@ -669,7 +670,7 @@ export type RouteConfigProps<
     | ScreenOptions
     | ((props: {
         route: DescriptorRouteProp<ParamList, RouteName>;
-        navigation: Navigation;
+        navigation: Navigation | undefined;
         theme: ReactNavigation.Theme;
       }) => ScreenOptions);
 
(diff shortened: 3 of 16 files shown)


packages/expo-router/src/react-navigation/core/types.tsx
```diff
@@ -674,7 +676,7 @@ export type RouteConfigProps<
     | ScreenListeners<State, EventMap>
     | ((props: {
         route: RouteProp<ParamList, RouteName>;
-        navigation: Navigation;
+        navigation: Navigation | undefined;
```
jakub-agent, line 684: [🏗️ **design**] `listeners` is widened here, but the three sibling callbacks that receive the *same* descriptor navigation are left as required `Navigation`:

- `RouteConfigProps.options` (~line 667)
- `ScreenLayoutArgs.navigation` (~line 541, used by `layout` / `screenLayout`)
- `RouteGroupConfig.screenOptions` and `DefaultNavigatorOptions.screenOptions`

In `useDescriptors`, `getOptions(route, navigation, …)`, `render(route, navigation, …)` → `layout({ route, navigation, … })` and the `listeners` path at `useNavigationBuilder.tsx:653` all read from the same `navigations[route.key]` value. If that value can be `undefined`, all four are affected — not just `listeners`.

#48618 makes this concrete: `useDescriptors` there calls `getOptions(route, undefined, {})` for placeholder descriptors, so the public `options` callback demonstrably receives `undefined` while its type still promises a `Navigation`. That's an unsound public type the moment #48618 lands, and it'd be a second breaking change to fix afterwards. Widening `options`/`screenOptions`/`layout` in this PR keeps the whole break in one release.

Was the narrower scope deliberate, or is the rest planned for #48618?

packages/expo-router/src/react-navigation/core/NavigationProvider.tsx
```diff
@@ -14,7 +14,7 @@ export const NavigationRouteContext = React.createContext<Route<string> | undefi
 
 type Props = {
   route: Route<string>;
-  navigation: NavigationProp<ParamListBase>;
+  navigation: NavigationProp<ParamListBase> | undefined;
```
jakub-agent, line 17: [🏗️ **design**] Widening this prop means `NavigationContext` can now hold `undefined`. `useNavigation` returns `navigation ?? root` (`useNavigation.tsx:24`) and only throws when *both* are undefined — so a subtree under `NavigationProvider navigation={undefined}` silently receives the root container ref instead of a screen navigation object.

That ref is not a drop-in substitute: `setOptions` throws `Cannot call setOptions outside a screen` (`BaseNavigationContainer.tsx:203`), `getParent()` returns `undefined`, and `goBack`/`navigate`/`isFocused` act on the root container rather than the screen. So the failure mode is a wrong target or a confusing throw deep inside user code, not a clear error at the boundary.

Not reachable in this PR — every producer still passes a real object. But since the stack's goal is to make `undefined` reachable, it's worth deciding now: either skip rendering `NavigationContext.Provider` when `navigation` is undefined, or throw in `NavigationProvider` with a message naming the placeholder case. Leaving the fallback as-is trades a crash for silently-wrong navigation.

packages/expo-router/src/react-navigation/core/types.tsx
```diff
@@ -580,9 +580,11 @@ export type Descriptor<
   route: DescriptorRoute<Route>;
 
   /**
-   * Navigation object for the screen
+   * Navigation object for the screen. `undefined` when the descriptor
+   * describes a route name declared in a layout that has no live route
+   * instance in navigation state.
    */
-  navigation: Navigation;
+  navigation: Navigation | undefined;
```
jakub-agent, line 592: [💡 suggestion] Consider `navigation?: Navigation` (optional property) rather than a required property holding `undefined`. `routeSource?: RouteSource` two fields below already uses that style, and #48618 has to write `as DescriptorMap[string]` on its placeholder descriptors specifically because it omits the `navigation` key — the cast disappears with an optional property.

Both forms read identically for consumers (`exactOptionalPropertyTypes` is off). The trade-off is on the producer side: the current form forces internal producers to write `navigation: undefined` explicitly, which is arguably better discipline — but #48618 bypasses it with a cast anyway, which is the weaker outcome of the two.

packages/expo-router/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### 🛠 Breaking changes
 
+- Allow descriptors without a `navigation` object in `expo-router/react-navigation` types.
```
jakub-agent: [🐧 nit] Missing the PR link and author attribution that every other entry in this file carries: `([#48619](https://github.com/expo/expo/pull/48619) by [@Ubax](https://github.com/Ubax))`. Same gap on the `key` entry from #48596 directly below.

jakub-agent (commented):

PR 2/3 in the stack (`main` → [#48596](https://github.com/expo/expo/pull/48596) → **#48619** → [#48618](https://github.com/expo/expo/pull/48618)). Type-only preparation: widens `Descriptor.navigation` to `Navigation | undefined` and updates the call sites that dereference it.

Verified locally on `ec85997`:

- `tsc --noEmit` produces the **same 31 errors** as the base commit `b0d165d` — no new type errors. (All pre-existing: missing `react-native-tab-view` / `react-native-pager-view` optional deps, plus one already-failing type test.)
- `jest` — 4734 passed, 6 skipped, 343 suites. No regressions.
- The two runtime changes (`navigation?.isFocused()` in `stack/views/Header/Header.tsx`, `navigation?.goBack` in `NativeStackView.tsx`) are both unreachable today: `useDescriptors` still always sets `navigation` from `navigations[route.key]!`.

The change is small and correct in isolation. My concerns are about the type surface being **partially** widened — the follow-up PR #48618 shows concretely which gaps matter. Details inline.

packages/expo-router/src/react-navigation/core/types.tsx
Ubax: Address here - change in remaining places

packages/expo-router/src/react-navigation/core/types.tsx
Ubax: Don't change to `?`. Though add a comment that this is intentional `| undefined` so that producers need to make it explicit

Remove any cast caused by this by writing `navgation: undefined`

packages/expo-router/src/react-navigation/core/useDescriptors.tsx
```diff
@@ -45,7 +45,7 @@ export type ScreenConfigWithParent<
 type ScreenLayout<ScreenOptions extends {}> = (props: {
   route: RouteProp<ParamListBase, string>;
   options: ScreenOptions;
-  navigation: any;
+  navigation: any | undefined;
```
jakub-agent, line 48: [💡 suggestion] `any | undefined` collapses to plain `any` — TypeScript absorbs every union member into `any`, so these two edits (here and line 58) are compile-time no-ops. Verified:

```ts
type Eq<A, B> = (<T>() => T extends A ? 1 : 2) extends <T>() => T extends B ? 1 : 2 ? true : false;
type _ = Expect<Eq<any | undefined, any>>;  // passes
```

So unlike the widenings in `types.tsx`, nothing here is enforced, and the new type test can't cover it either. Either drop the `| undefined` (the `any` already permits it) or give these locals the real type — `NavigationProp<ParamListBase, string, string | undefined, State, ScreenOptions, EventMap> | undefined`, matching what `getOptions` now takes just below. The latter would actually check the placeholder path that #48618 adds.

packages/expo-router/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### 🛠 Breaking changes
 
+- Make `navigation` optional in `Descriptor` type ([#48619](https://github.com/expo/expo/pull/48619) by [@Ubax](https://github.com/Ubax))
```
jakub-agent, line 7: [💡 suggestion] The entry names only `Descriptor`, but the break is wider than that: `screenOptions` and `screenListeners` on `DefaultNavigatorOptions`, `options` / `listeners` / `layout` on `RouteConfigProps`, and `screenOptions` / `screenLayout` on `RouteGroupConfig` all changed. Those are the callbacks people writing a custom navigator on `expo-router/react-navigation` actually touch, and they're where the compile errors will land.

Reproduced against `createNativeStackNavigator` — clean on base `df63245`, three errors on `5af34a3`:

```
src/__adv2.tsx(9,7):  error TS18048: 'navigation' is possibly 'undefined'.   // screenOptions
src/__adv2.tsx(16,9): error TS18048: 'navigation' is possibly 'undefined'.   // Screen options
src/__adv2.tsx(20,9): error TS18048: 'navigation' is possibly 'undefined'.   // Screen listeners
```

Worth naming the callbacks in the entry, and the previous wording's `expo-router/react-navigation` qualifier was useful for scoping who's affected — it got dropped in this revision. Something like: "Make `navigation` optional in `Descriptor` and in the `screenOptions`, `options`, `listeners`, and `layout` callbacks in `expo-router/react-navigation` types."

jakub-agent (commented):

**Iteration 2** (previous review: [#pullrequestreview-4876892936](https://github.com/expo/expo/pull/48619#pullrequestreview-4876892936), on `ec85997`). PR 3/4 in the stack: `main` → [#48596](https://github.com/expo/expo/pull/48596) → [#48638](https://github.com/expo/expo/pull/48638) → **#48619** → [#48618](https://github.com/expo/expo/pull/48618).

**Previous findings — all resolved:**

- *Partial widening* → fixed. `screenOptions`/`screenListeners` on `DefaultNavigatorOptions`, `options`/`listeners`/`layout` on `RouteConfigProps`, `screenOptions`/`screenLayout` on `RouteGroupConfig`, and `ScreenLayoutArgs.navigation` are now all `Navigation | undefined`, with `core/__tests__/types.test.ts` locking each one in.
- *`useNavigation` root fallback* → split out into [#48638](https://github.com/expo/expo/pull/48638) and inserted below this PR in the stack.
- *`navigation?:` vs required* → deliberately declined, and the reasoning is now written into the doc comment plus asserted by `_DescriptorNavigationIsRequired`. Good outcome; the intent is no longer implicit.
- *CHANGELOG attribution* → added.

**Verified on `5af34a3`:**

- `tsc --noEmit` → same 31 errors as base `df63245`. No new type errors. (All pre-existing: missing `react-native-tab-view` / `react-native-pager-view` optional deps, plus one already-failing type test.)
- `jest` → 4738 passed, 6 skipped, 347 suites, 0 failed. (One run hit a `SIGILL` worker crash; a clean rerun passed, so it's infra flake, not this PR.)
- `oxlint` → clean.
- **Mutation-tested the new type test** — it is not vacuous. Changing `Descriptor.navigation` to `navigation?: Navigation` fails `_DescriptorNavigationIsRequired` (line 70); reverting `RouteConfigProps.options` to a required `Navigation` fails `_ScreenOptionsNavigationCanBeUndefined` (line 49). Both caught.

Nothing blocking from my side. Two minor points inline.

jakub-agent (approved):

Approving `5af34a3`. Every finding from the two earlier passes is resolved, and the checks below all pass. The two open suggestions in [#pullrequestreview-4881581144](https://github.com/expo/expo/pull/48619#pullrequestreview-4881581144) (`any | undefined` being a no-op in `useDescriptors.tsx`, and the CHANGELOG entry understating the break) are non-blocking — take or leave them.

Verified on `5af34a3`:

- `tsc --noEmit` → same 31 errors as base `df63245`; no new type errors.
- `jest` → 4738 passed, 6 skipped, 347 suites, 0 failed.
- `oxlint` → clean.
- Mutation-tested `core/__tests__/types.test.ts`: flipping `Descriptor.navigation` to optional, and reverting `RouteConfigProps.options` to a required `Navigation`, are each caught. The new type assertions are real.

Note for whoever merges: as an automated reviewer I can't judge product intent or the stack's rollout plan, so please don't let this substitute for a human approval if the repo's process expects one.

Issue #162, opened 2017-05-09
Flow for all public facing APIs


brentvatne:
Moved from: https://github.com/expo/expo-sdk/issues/36

dikaiosune:
@ide is this going to happen as a product of the recent sdk work you've been doing?

ide:
Sort of - I fixed all the existing Flow types and added some to FileSystem but am not directly working towards increasing coverage.

danielmartinprieto:
Hi! I'm having a hard time finding the official Flow definitions for expo - I've started using flow-typed and the cli created stubs for it, because it couldn't find them either. What am I doing wrong? How can I enable/get Flow definitions for Expo?

Thanks.

ide:
The official Flow definitions are included inline with the code itself. We don't use flow-typed because of the spatial coupling it introduces in a structured design sense.

danielmartinprieto:
@ide hey, thanks for answering!

That's great, then. Do you know why I am getting this `'expo' Required module not found`? How can I tell Flow where to find these definitions?

Thanks again!

ide:
Do you perhaps have all node_modules ignored under flowconfig?

danielmartinprieto:
No all `node_modules`, but I've been adding more and more subdirectories of it until I was able to get a decent amount of Flow errors of my code, otherwise it felt almost useless. I've checked my `.flowconfig` file and I've seen this under `[ignore]`:
```
; Ignore missing expo-sdk dependencies (temporarily)
; https://github.com/expo/expo/issues/162
.*/node_modules/expo/src/*
```
I didn't add this, I've checked and it's included in CRNA. If you follow that issue in the comment and its referenced ones, you end up in this one, haha.

ide:
The Flow types are included with the code we publish to npm so they're most likely on your computer. If you can debug why Flow can't see them I expect Flow types will just start working for you then.

danielmartinprieto:
I've removed that line in the ignore section but still don't get any Flow errors when I use something from 'expo' in an incorrect way, for instance:
```javascript
// @flow
import { Constants } from 'expo';

const log = (something: number) => console.log(something);
log(3); // no Flow error 
log(Constants.deviceName); // no Flow error :(
log('hi there'); // Flow error
```

ide:
Constants doesn't have every field typed because they're sent from native. Browse around here for something that has types: https://github.com/expo/expo-sdk/tree/3af2763fc97abdb879f3455a71816c899f8617e5

[...]

PR #48624, opened 2026-08-06
fix(cli): Show Xcode build log path when run:ios fails

ramonclaudio:
Follow-up to [#47748](https://github.com/expo/expo/pull/47748)
Related to [#162](https://github.com/expo/third-party/pull/162)

# Why

`expo run:ios` can sometimes end with this message: `error: the following command failed with exit code 1 but produced no further output`.

This does not explain why the build failed. The full log is saved, but the CLI does not show the log path.

# How

We keep this message as an error, but still check the full log for another error that explains the failure. We also show the build log path before the full Xcode output so it is easier to find.

# Test Plan

- Ran `corepack pnpm --filter @expo/cli exec jest --runInBand --watchman=false`
- Ran `et check-packages @expo/cli --no-test`

# Checklist

### DIFF
--- a/packages/@expo/cli/src/run/ios/XcodeBuild.ts
+++ b/packages/@expo/cli/src/run/ios/XcodeBuild.ts
@@ -445,13 +445,12 @@ export function _assertXcodeBuildResults(
   xcodeProject: { name: string },
   logFilePath: string
 ): void {
-  const errorTitle = `Failed to build iOS project. "xcodebuild" exited with error code ${code}.`;
+  const errorHeader = _formatXcodeBuildFailure(code, logFilePath);
 
   const throwWithMessage = (message: string): never => {
     throw new CommandError(
-      `${errorTitle}\nTo view more error logs, try building the app with Xcode directly, by opening ${xcodeProject.name}.\n\n` +
-        message +
-        `Build logs written to ${chalk.underline(logFilePath)}`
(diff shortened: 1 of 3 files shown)


packages/@expo/cli/src/run/ios/XcodeBuild.ts
```diff
@@ -479,7 +484,12 @@ export function _extractXcodeBuildErrorLines(output: string): string[] {
       errors.push(line);
     }
   }
-  return errors;
+  const errorsWithDetails = errors.filter((error) => !isXcodeBuildNoOutputErrorLine(error));
+  return errorsWithDetails.length > 0 ? errorsWithDetails : errors;
+}
+
+function isXcodeBuildNoOutputErrorLine(line: string): boolean {
+  return XCODE_BUILD_NO_OUTPUT_ERROR_MESSAGE.test(line);
```
krystofwoldrich, line 491: Similar to my expo/xcpretty comment, I think we should include the no output errors.

krystofwoldrich:
Looks great, I have just one comment about the no output errors.

---

Here is example from my test project which fails on Xcode 27 (it's not update to the latest SDK).

Before this RP
Screenshot 2026-08-11 at 10 30 01


After this PR
Screenshot 2026-08-11 at 10 29 52

kitten (approved):

krystofwoldrich (approved):

PR #48629, opened 2026-08-06
[autolinking] sort react-native-config autolinking deps result

Kudo:
# Why

alternatively solution for https://github.com/expo/expo/pull/48027#issuecomment-5203824078

# How

sort dependencies result from `react-native-config`

# Test Plan

added unit test

# Checklist


kitten (approved):
Looks good in principle, though, given the amount of ordering concerns it may be better to move extra sorting logic to fingerprint instead of here, since they keep piling up, so they're starting to look like a serialisation/fingerprint concern

Kudo:
> Looks good in principle, though, given the amount of ordering concerns it may be better to move extra sorting logic to fingerprint instead of here, since they keep piling up, so they're starting to look like a serialisation/fingerprint concern

we could bring the discussion for https://github.com/expo/expo/pull/40761 again.

Kudo:
> > Looks good in principle, though, given the amount of ordering concerns it may be better to move extra sorting logic to fingerprint instead of here, since they keep piling up, so they're starting to look like a serialisation/fingerprint concern
> 
> we could bring the discussion for #40761 again.

or maybe this is more proper approach than sorting https://github.com/expo/expo/pull/48661

i would like to merge this pr anyway. with #48661 we would prevent future unstable autolinking config issues

PR #48631, opened 2026-08-06
[cli] Fail when --private-key-path is set without codeSigningCertificate

gwdp:
# Why

When `--private-key-path` is passed without `updates.codeSigningCertificate` in the resolved app config: that implies an intent to sign, but `@expo/cli` previously ignored the flag and continued without signing the development manifest.

# How

- In `getProjectCodeSigningCertificateAsync`, throw if a private key path is provided without a certificate path
- Leave the unsigned path alone when neither is set

# Test Plan

- Unit tests for no-cert/no-key → `null`, and no-cert/with-key → throw
- CI

# Checklist

### DIFF
--- a/packages/@expo/cli/CHANGELOG.md
+++ b/packages/@expo/cli/CHANGELOG.md
@@ -21,6 +21,7 @@
 
 ### 🐛 Bug fixes
 
+- Fail when `--private-key-path` is passed without `updates.codeSigningCertificate` in the resolved app config, instead of ignoring the flag and continuing without signing.
 - [Internal] Fix `LogStream.destroy()` racing a pending write and dropping log data ([#47181](https://github.com/expo/expo/pull/47181) by [@kitten](https://github.com/kitten))
 - Ignore simulators reported by `devicectl` and support json version 5 on Xcode 27 ([#48001](https://github.com/expo/expo/pull/48001) by [@crockalet](https://github.com/crockalet))
 - In non-interactive shells, automatically roll over to the next available port when default is busy, unless a specific port is specified with `--port` or `RCT_METRO_PORT` ([#47771](https://github.com/expo/expo/pull/47771) by [@kitten](https://github.com/kitten))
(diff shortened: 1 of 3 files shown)


wschurman (approved):

gwdp:
> ## 🤖 AI code review
> **Decision:** Approve
> 
> **Overall PR risk: Medium.** The change modifies existing behavior: `getProjectCodeSigningCertificateAsync` returned `null` when `updates.codeSigningCertificate` was absent, and it now throws a `CommandError` if the caller passed `--private-key-path`. The affected surface is one function in `@expo/cli`, reached per manifest request in the Expo Go manifest middleware, so a developer who passes `--private-key-path` against a project with no certificate now gets a failed manifest response instead of an unsigned manifest. Expo Go development signing and requests without a signature header are not affected, and rollback is a single self-contained revert. No specialist reviewer reported a correctness, security, or secret defect.
> 
> No findings.
> 
> _This review is advisory — it never blocks a merge and never auto-approves._

Triple-checked the medium-risk callout: Expo Go/expo-root and no-signature paths are unchanged; the new throw only hits project-keyid requests with --private-key-path and no cert, which was already a broken signing setup.

packages/@expo/cli/CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
 
 ### 🐛 Bug fixes
 
+- Fail when `--private-key-path` is passed without `updates.codeSigningCertificate` in the resolved app config, instead of ignoring the flag and continuing without signing.
```
kitten, line 24: This was merged without a changelog link here. Just a note, that I don't think CI fully triggered here

gwdp: Thank you for the note @kitten. Sorry for the trouble. Do you recommend me getting another PR to fix changelog?

kitten: No need, and no worries 👍 I've fixed this in the picks, and we can either address this later, or leave it as is (since we don't enforce this that strictly on `main`). If it causes issues, I'm sure it'll be flagged and fixed down the line

gwdp: Thank you sir 🙇‍♂️

PR #48633, opened 2026-08-07
[expo][text-decoder] Bulk-decode ASCII prefixes

Ignigena:
# Why

`TextDecoder` on Hermes is provided by Expo's pure-JS fork of `text-encoding` in `packages/expo/src/winter/TextDecoder.ts`. `TextEncoder` is native in Hermes, but the decoder is not — so every byte of every stream that an Expo app decodes is processed in JavaScript, on the JS thread, one byte at a time.

That cost is invisible for a one-shot `response.text()`. It is very visible for anything that streams.

## Our use case

We consume a long-lived Server-Sent Events connection that drives live map updates, through a stream held open for as long as the screen is mounted.

`Response.body` is a `ReadableStream<Uint8Array>` (that is what the Fetch standard specifies, and `expo/fetch` is right to implement it that way) which means using `TextDecoder` to decode the stream:

```js
const reader = response.body.getReader();
// ... or `response.body.pipeThrough(new TextDecoderStream())` -- though it wraps the same decoder
const decoder = new TextDecoder();

while (true) {
  const { done, value } = await reader.read();
  if (done) break;
  handle(decoder.decode(value, { stream: true }));
}
```

`TextDecoder` is therefore on the hot path of *any* incremental text read from `fetch`, for the entire life of the connection — SSE, NDJSON, streamed model responses, and React Server Component payloads all arrive here.

Our payload is pure ASCII, as most streamed text is: SSE framing (`data: `, `\n\n`), JSON, NDJSON, and HTTP headers are ASCII by construction. **The single most common input to this decoder is the input it handles least efficiently.**

**We measured a sustained ~5 FPS drop on the JS thread while that stream is live.** On a screen whose whole job is a smoothly panning map, that is user-visible. It is also the kind of regression that is hard to attribute: nothing in the profile points at "the decoder", it just shows a broadly slower JS thread.

## Quick note on scope: the file's header comment understates its reach

`TextDecoder.ts` opens with:

> `TextEncoder` is in Hermes and we only need utf-8 decoder for React Server Components.

That is no longer accurate, and it makes the file look far more niche than it is. On native, the decoder is reached by every Expo app, whether or not RSC is enabled:

- `Expo.fx` imports `./winter` unconditionally, and `runtime.native.ts` installs `TextDecoder` as a global — there is no feature flag or app config gate. (On web, `runtime.ts` installs no decoder, so this change is native-only in effect.)
- `@expo/metro-config` resolves `expo/src/winter/index.ts` and pushes it into Metro's `preModules`, so the winter runtime is evaluated ahead of app code.
- `FetchResponse#text()` calls `new TextDecoder().decode(...)`, and `TextDecoderStream` wraps the same class. `expo/fetch` replaces global `fetch` unless `EXPO_PUBLIC_USE_RN_FETCH` is set, so this is the decoder behind essentially every text read from a response body.

That cuts both ways, and it is the reason for the depth of the test plan below: the upside applies to every app, and so would any regression. It is also why the change is deliberately additive — the existing state machine is untouched and simply receives fewer bytes.

## Root cause

The slow path is not just "a loop over bytes". Three costs compound on every `decode()` call:

1. **`new Stream(bytes)` calls `Array.prototype.slice.call(tokens)` and then `.reverse()`.** For a 32 KB chunk this allocates a 32,768-element *boxed* JS array and reverses it, before any decoding starts.
2. **The decode loop pops one byte at a time** off that array through a `Stream` object, dispatching into `UTF8Decoder.handler` per byte.
3. **Output accumulates into a `number[]`** with a `push` per code point, and `codePointsToString` then builds the result with `s += String.fromCharCode(cp)` per code point.

Measured on Hermes 1.0, 2 MB of representative ASCII SSE frames in 32 KB chunks, this decoder sustains **4.4 MB/s**. Every byte of that work lands on the JS thread, competing with React rendering and gesture handling. The same decoder under V8 manages 13 MB/s, so the pure-JS cost is roughly 3x worse on the engine Expo apps actually ship with.

# How

This adds a bulk ASCII decoding path. The existing state machine is untouched and still handles every case it handled before; it simply receives fewer bytes.

ASCII is byte-identical to its UTF-8 encoding, so the leading run of bytes below `0x80` in a chunk can be converted in bulk instead of one at a time:

- **`decodeAsciiPrefix(bytes)`** scans forward to the first byte `>= 0x80` and builds a string from everything before it with `String.fromCharCode.apply` over 8 KB batches. `apply` rather than spread is deliberate — spreading the typed array measured ~7x slower, because it goes through the iterator protocol. The batch size is a plateau rather than a tuned constant: 256 through 8192 all land within 3% of peak, smaller batches give it back (64 bytes reaches 78% of peak on V8, 90% on Hermes), and above 8192 V8 halves its throughput while Hermes stays flat. 8192 is the largest value that suits both.

  The scan reads `bytes.length` into a local before looping, which looks like a redundant micro-optimisation and is not. Hermes compiles ahead of time, but its optimiser will not hoist a property load out of a loop — `length` on a typed array is an ordinary property access it cannot prove invariant — and there is no JIT to catch it later. Even at `-O`, every iteration performs a real lookup: reading it once took the scan from 87 ms to 21 ms over 2 MB, and the end-to-end ASCII win from 3.7x to 8.4x. V8 hoists it itself and is unaffected either way.
- **`UTF8Decoder#hasPendingBytes`** is a new getter exposing `utf8BytesNeeded !== 0`, so `decode()` can tell whether a multi-byte sequence is still awaiting continuation bytes from a previous streaming chunk.
- **`decode()`** takes the prefix (when the encoding is UTF-8 and nothing is pending), advances `bytes` past it, and prepends it to the result. Everything after the prefix flows through the existing code unchanged.

## Current implementation is deliberately narrow

This new fast path only ever consumes bytes below `0x80`, and stops at the first byte it isn't sure about. Everything else — non-ASCII sequences, malformed input, BOMs, a pending carry — falls through to the existing state machine and takes exactly the path it takes today.

That leaves two things to check: whether the bytes it does consume are genuinely trivial, and whether the handoff preserves decoder state.

On the bytes it consumes:

- Every byte is `< 0x80`, so in UTF-8 every byte is a complete code point: nothing to carry, no continuation bytes to interpret.
- A BOM is `EF BB BF`, and `0xEF >= 0x80`, so a BOM can never fall inside the prefix.
- No overlong, surrogate or out-of-range encoding is expressible in a single byte `< 0x80`, so `decoderError` is unreachable and `fatal` cannot behave differently for these bytes.

On the handoff:

- **The carry is the one piece of state that matters.** A pending multi-byte sequence is terminated by whatever byte follows it, *including* an ASCII byte, which must produce `U+FFFD` (or throw under `fatal`) and then be re-read. Taking a prefix while a sequence is pending would drop those bytes silently, so `hasPendingBytes` disables the fast path entirely in that case. See the control build in the test plan.
- **`_BOMseen` is set when the prefix is non-empty**, which is what `serializeStream`'s `else if (stream.length > 0)` branch would have done for the same output. A leading BOM is still stripped; a BOM appearing after ASCII output is still left alone.
- **The decoder lifecycle is untouched.** The prefix is taken after steps 1 and 2 of the spec algorithm, so decoder creation, the `_BOMseen` reset, `_doNotFlush`, and the flush loop all behave as before.
- **A prefix keeps the diff smaller than bailing out would.** The first instinct is to scan the whole buffer and fall back to the slow path if *any* byte is `>= 0x80`, but an early return has to re-implement `_doNotFlush`, the `this._decoder = null` reset and the BOM rules, which is more of the existing logic to get wrong. It is also slower when a chunk happens to end on a non-ASCII byte, since the whole scan is then discarded (the last row of the benchmark below.)

# Test Plan

## 1. Unit test coverage

`packages/expo/src/winter/__tests__/TextDecoder.test.native.ts` gains an `ASCII bulk decoding` block of six cases, each covering a boundary between the two paths:

| test | what it pins down |
| --- | --- |
| decodes buffers spanning several `String.fromCharCode` batches | multi-batch concatenation; no existing test decodes an ASCII buffer larger than one batch |
| replaces a bare continuation byte surrounded by ASCII | the scan stops at `0x80`, not after it |
| does not drop a lead byte that ended the previous chunk | the carry hazard, in replacement mode |
| still throws in fatal mode when ASCII breaks a pending sequence | the same hazard, in `fatal` mode |
| keeps a BOM that follows ASCII output | the fast path sets `_BOMseen` |
| resets the BOM seen flag between non-streaming decodes | the new `_BOMseen` write is still cleared by step 1 |

All six pass against the **unpatched** decoder as well — they codify existing behaviour rather than simply covering the new implementation.

## 2. Manual verification against the unpatched decoder

The decoder was also checked exhaustively against its own previous revision. It is pure TypeScript with no React Native imports, so both revisions transpile and run side by side under Node:

```bash
git stash                          # unpatched
npx tsc packages/expo/src/winter/TextDecoder.ts \
  --ignoreConfig --outDir /tmp/td --module esnext --target es2020 --skipLibCheck
mv /tmp/td/TextDecoder.js /tmp/td/base.mjs
git stash pop                      # patched
npx tsc packages/expo/src/winter/TextDecoder.ts \
  --ignoreConfig --outDir /tmp/td --module esnext --target es2020 --skipLibCheck
mv /tmp/td/TextDecoder.js /tmp/td/patched.mjs

node decoder-check.mjs /tmp/td/base.mjs /tmp/td/patched.mjs
```

```
  unpatched 422 splits
  patched   422 splits
272 sweep blocks, 119277 cases, 0 failures
```

The output must be identical for every input, across five kinds of check:

1. **Deterministic edge cases** under every combination of `fatal` and `ignoreBOM`: leading, trailing, split and doubled BOMs; truncated sequences (`C3`, `E2 82`, `F0 9F 8E`) at EOF and followed by ASCII; bare continuation bytes; invalid bytes (`FF FE`); overlong, surrogate and out-of-range encodings; NUL bytes; empty input and empty chunks; and the `0x7F`/`0x80` boundary.
2. **Randomised content at four ASCII densities** (0%, 40%, 90%, 99%), each decoded whole, at *every* single split point, byte-at-a-time, and with random chunking.
3. **Buffers crossing the `String.fromCharCode` batch boundary** — 8191/8192/8193/16384/16385/40000 bytes, with a non-ASCII byte planted at each boundary offset.
4. **A full code point sweep** of `U+0000`–`U+10FFFF` in blocks, interleaved with ASCII so both paths run within the same buffer, compared against the unpatched decoder *and* the source string (so both revisions being wrong would still fail).
5. **Chunking invariance** — the result must not depend on where the byte stream is split, verified for both revisions.

Thrown errors are folded into the compared value, so a `fatal` divergence fails rather than being swallowed.

To confirm the harness is actually sensitive to the hazard it is meant to catch, a control build with the `!hasPendingBytes` guard removed reports **308 failures** against the same run.

<details>
<summary><code>decoder-check.mjs</code></summary>

```js
// usage: node decoder-check.mjs <unpatchedModule> <patchedModule>
const [, , refPath, candPath] = process.argv;
const { TextDecoder: Ref } = await import(refPath);
const { TextDecoder: Cand } = await import(candPath);

const encoder = new TextEncoder();
const BOM = [0xef, 0xbb, 0xbf];
const OPTS = [{}, { ignoreBOM: true }, { fatal: true }, { fatal: true, ignoreBOM: true }];
let cases = 0;
let failures = 0;

function record(name, a, b) {
  cases++;
  if (a !== b) {
    if (++failures <= 10) {
      console.log(`FAIL ${name}\n  expected ${JSON.stringify(a)}\n  actual   ${JSON.stringify(b)}`);
    }
  }
}

// Errors are folded into the result so `fatal` divergence is compared, not swallowed.
function decode(Ctor, opts, chunks) {
  const decoder = new Ctor('utf-8', opts);
  let out = '';
  try {
    for (const chunk of chunks) out += decoder.decode(new Uint8Array(chunk), { stream: true });
    return out + decoder.decode();
  } catch (error) {
    return `THROW:${error.constructor.name}:${out}`;
  }
}

const compare = (name, opts, chunks) =>
  record(name, decode(Ref, opts, chunks), decode(Cand, opts, chunks));

// 1. Deterministic edge cases, each under every combination of `fatal` and `ignoreBOM`.
const edge = {
  empty: [],
  'bom only': [BOM],
  'bom then ascii': [[...BOM, 0x61, 0x62]],
  'bom split then ascii': [[0xef], [0xbb], [0xbf, 0x61]],
  'ascii then bom': [[0x61], BOM],
  'double bom': [[...BOM, ...BOM, 0x61]],
  'truncated 2-byte then ascii': [[0xc3], [0x61, 0x62]],
  'truncated 3-byte then ascii': [[0xe2, 0x82], [0x61]],
  'truncated 4-byte then ascii': [[0xf0, 0x9f, 0x8e], [0x61]],
  'truncated then eof': [[0xc3]],
  'bare continuation': [[0x80, 0x61]],
  'invalid ff fe': [[0xff, 0xfe, 0x61]],
  overlong: [[0xc0, 0x80, 0x61]],
  'surrogate encoding': [[0xed, 0xa0, 0x80, 0x61]],
  'too large': [[0xf5, 0x80, 0x80, 0x80, 0x61]],
  'ascii then invalid then ascii': [[0x61, 0xff, 0x62]],
  'empty chunk between ascii': [[0x61], [], [0x62]],
  'nul bytes': [[0x00, 0x00, 0x61]],
  '0x80 boundary': [[0x7f, 0x80, 0x61]],
};
for (const [name, chunks] of Object.entries(edge)) {
  for (const opts of OPTS) compare(`${name} ${JSON.stringify(opts)}`, opts, chunks);
}

// 2. Randomised content at four ASCII densities, against every way of splitting it.
let seed = 0x2f6e2b1;
function rnd() {
  seed ^= seed << 13;
  seed ^= seed >>> 17;
  seed ^= seed << 5;
  return (seed >>> 0) / 0x100000000;
}
function randomBytes(length, asciiBias) {
  const out = [];
  const codePoint = (base, span) =>
    out.push(...encoder.encode(String.fromCodePoint(base + Math.floor(rnd() * span))));
  while (out.length < length) {
    const r = rnd();
    if (r < asciiBias) out.push(Math.floor(rnd() * 0x80));
    else if (r < asciiBias + 0.2) codePoint(0x80, 0x700);
    else if (r < asciiBias + 0.35) codePoint(0x4e00, 0x1000);
    else if (r < asciiBias + 0.5) codePoint(0x1f300, 0x200);
    else if (r < asciiBias + 0.6) out.push(...BOM);
    else out.push(Math.floor(rnd() * 0x100)); // arbitrary, often invalid
  }
  return out;
}
for (const asciiBias of [0.0, 0.4, 0.9, 0.99]) {
  for (let trial = 0; trial < 400; trial++) {
    const bytes = randomBytes(1 + Math.floor(rnd() * 40), asciiBias);
    for (const opts of OPTS.slice(0, 3)) {
      compare('rand whole', opts, [bytes]);
      for (let i = 0; i <= bytes.length; i++) {
        compare(`rand split@${i}`, opts, [bytes.slice(0, i), bytes.slice(i)]);
      }
      compare('rand bytewise', opts, bytes.map((b) => [b]));
      const chunks = [];
      for (let i = 0; i < bytes.length; ) {
        const n = 1 + Math.floor(rnd() * 5);
        chunks.push(bytes.slice(i, i + n));
        i += n;
      }
      compare('rand chunks', opts, chunks);
    }
  }
}

// 3. Buffers that cross the `String.fromCharCode` batching boundary.
for (const len of [8191, 8192, 8193, 16384, 16385, 40000]) {
  const ascii = Array.from({ length: len }, (_, i) => 0x20 + (i % 0x5f));
  compare(`large ${len}`, {}, [ascii]);
  compare(`large ${len} + non-ascii tail`, {}, [[...ascii, 0xe4, 0xb8, 0x80]]);
  compare(`large ${len} split`, {}, [ascii.slice(0, len >> 1), ascii.slice(len >> 1)]);
  for (const pos of [0, 1, 8191, 8192, 8193, len - 1].filter((p) => p < len)) {
    compare(`large ${len} non-ascii planted@${pos}`, {}, [ascii.with(pos, 0xc3)]);
  }
}

// 4. Every code point, in blocks, interleaved with ASCII so both paths run in one buffer.
let sweepBlocks = 0;
for (let base = 0; base <= 0x10ffff; base += 0x1000) {
  let block = '';
  for (let cp = base; cp < base + 0x1000 && cp <= 0x10ffff; cp += 7) {
    if (cp < 0xd800 || cp > 0xdfff) block += String.fromCodePoint(cp) + ' ascii text ';
  }
  sweepBlocks++;
  const bytes = encoder.encode(block);
  // Also checked against the source string, so both revisions being wrong would still fail.
  record(`sweep U+${base.toString(16)} value`, block, new Cand().decode(bytes));
  record(`sweep U+${base.toString(16)}`, new Ref().decode(bytes), new Cand().decode(bytes));
}

// 5. Chunking invariance: the result must not depend on where the stream is split.
const sample = encoder.encode(
  'data: {"id":1,"name":"caf\u00e9","city":"\u6c34\u6c34"}\n\n\u{1f389} ascii ascii ascii\n'.repeat(6)
);
for (const [label, Ctor] of [
  ['unpatched', Ref],
  ['patched', Cand],
]) {
  const whole = new Ctor().decode(sample);
  let splits = 0;
  for (let i = 0; i <= sample.length; i++, splits++) {
    record(`${label} split@${i}`, whole, decode(Ctor, {}, [sample.subarray(0, i), sample.subarray(i)]));
  }
  splits++;
  record(`${label} bytewise`, whole, decode(Ctor, {}, [...sample].map((b) => [b])));
  console.log(`  ${label.padEnd(9)} ${splits} splits`);
}

console.log(`${sweepBlocks} sweep blocks, ${cases} cases, ${failures} failures`);
process.exit(failures ? 1 : 0);
```

</details>

## 3. Performance

The numbers that matter are Hermes, since that is the engine this code path actually runs on. The figures below were taken on the Hermes VM itself, built to match the release Expo ships.

Pinning the right build is the fiddly part. React Native's `sdks/.hermesversion` points at `hermes-v0.17.0`, which is HBC 96 and stale; what actually ships is encoded in the `hermes-compiler` dependency version.

```bash
# 250829098.0.14 -> built 25-08-29, bytecode version 98.
find node_modules -path '*hermes-compiler/hermesc/linux64-bin/hermesc' -exec {} -version \;
#   Hermes release version: 1.0.0
#   HBC bytecode version: 98

git clone --single-branch --branch static_h https://github.com/facebook/hermes.git hermes-src
git -C hermes-src checkout 663d9db1                   # 2025-08-29, the commit behind that build

cmake -S hermes-src -B hermes-build -DCMAKE_BUILD_TYPE=Release \
  -DHERMES_RELEASE_VERSION=1.0.0 -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++
cmake --build hermes-build --target hermes hermesc -j 16

# Must report the same 1.0.0 / HBC 98 as the compiler above before any number is worth reading.
hermes-build/bin/hermes -version

# Both revisions, with `export` stripped so they transpile to standalone scripts.
git show main:packages/expo/src/winter/TextDecoder.ts | sed 's/^export class/class/' > base.ts
sed 's/^export class/class/' packages/expo/src/winter/TextDecoder.ts > patched.ts
tsc base.ts patched.ts --module none --target es2015 --skipLibCheck --outDir js
node mkbench.js                                       # inlines both revisions into bench.js

hermes-build/bin/hermes bench.js                      # hermesc optimises by default
hermes-build/bin/hermesc -O -emit-binary -out bench.hbc bench.js
hermes-build/bin/hermes bench.hbc                     # control: same figures as running the source
node bench.js                                         # the V8 control, from the same file
```

Those last two lines are worth spelling out. Static Hermes compiles ahead of time, so before claiming a JS-level win it had to be ruled out that the AOT compiler already performs it. It does not: running from source and from an explicit `-O` build agree to within noise, while an `-O0` build is ~1.8x slower, so these are optimised figures rather than an unoptimised worst case.

Measuring on the real VM is also what exposed the typed-array `length` load. Building the patch without the hoist and running it here costs 123 ms against 54 ms — 69 ms of a 123 ms run spent re-reading `bytes.length` per byte. On V8 that same load is free (1 ms either way over 2 MB), so no amount of Node benchmarking would have surfaced it, and this change would have topped out at 3.7x instead of 8.4x.

2 MB payloads, 32 KB chunks, best of 5 runs, AMD EPYC 7763:

| payload | unpatched | patched | change |
| --- | --- | --- | --- |
| pure ASCII (SSE/JSON) | 452 ms (4.4 MB/s) | 54 ms (37.0 MB/s) | **8.4x faster** |
| ASCII with 1% accents | 457 ms | 466 ms | 0.98x |
| CJK (worst case) | 434 ms | 438 ms | 0.99x |
| emoji (worst case) | 435 ms | 436 ms | 1.00x |
| ASCII chunks with non-ASCII final bytes | 455 ms | 50 ms | **9.1x faster** |

The identical script under Node 24 / V8, as a control that the win is not a quirk of one engine:

| payload | unpatched | patched | change |
| --- | --- | --- | --- |
| pure ASCII (SSE/JSON) | 156 ms (12.8 MB/s) | 10 ms (200 MB/s) | **15.6x faster** |
| ASCII with 1% accents | 154 ms | 152 ms | 1.01x |
| CJK (worst case) | 137 ms | 137 ms | 1.00x |
| emoji (worst case) | 137 ms | 137 ms | 1.00x |
| ASCII chunks with non-ASCII final bytes | 155 ms | 9 ms | **17.2x faster** |

Both engines agree on the shape: a large win on ASCII, and no measurable change on the non-ASCII worst cases. The margin is still wider on V8 for a structural reason. What remains in the fast path is a byte-at-a-time JS loop — the scan for the first non-ASCII byte — and V8 JITs that to roughly nothing, while Hermes has no JIT and pays interpreter dispatch per byte. Over 2 MB of ASCII that scan costs 1 ms on V8 against 21 ms on Hermes, which is around 40% of everything the patched path spends there.

Non-ASCII-leading input is unaffected: the scan stops on the first byte, so the added cost is a single comparison per `decode()` call. The last row is the case that defeats a whole-buffer ASCII scan; taking a prefix turns it into a full win instead of a regression.

Every run asserts that both revisions produce the same output length, so a figure can never come from a decoder that quietly stopped decoding.

<details>
<summary><code>bench.js</code></summary>

Hermes has no module system, so `mkbench.js` inlines both compiled revisions into one standalone script. The result runs unmodified on either engine, which is what makes the V8 table a control rather than a separate experiment.

```js
// mkbench.js — Hermes has no module system, so both revisions are inlined into one script.
const fs = require('fs');

const revision = (file, name) =>
  `var ${name} = (function () {\n${fs.readFileSync(file, 'utf8')}\nreturn TextDecoder;\n})();\n`;

fs.writeFileSync(
  'bench.js',
  revision('js/base.js', 'Unpatched') +
    revision('js/patched.js', 'Patched') +
    fs.readFileSync('bench.body.js', 'utf8')
);
```

```js
// bench.body.js — appended to the two revisions above to form bench.js.
// Runs unmodified on both engines: Hermes has no `console`, Node has no `print`.

var PAYLOAD_BYTES = 2 * 1024 * 1024;
var CHUNK_BYTES = 32 * 1024;
var RUNS = 5;

var log = typeof print === 'function' ? print : console.log;
var encoder = new TextEncoder();
var SSE_FRAME = 'data: {"id":123,"lat":37.7749,"lng":-122.4194,"speed":31.5,"heading":270}\n\n';

// Tiles the encoded sample to an exact byte length. The tail may cut a multi-byte sequence short,
// which both revisions see identically, so the comparison is unaffected.
function tile(sample, byteLength) {
  var pattern = encoder.encode(sample);
  var out = new Uint8Array(byteLength);
  for (var i = 0; i < byteLength; i++) out[i] = pattern[i % pattern.length];
  return out;
}

// Sprinkles a non-ASCII character through otherwise ASCII text at 1-in-`rate` characters.
function season(sample, rate) {
  var out = '';
  for (var i = 0; i < sample.length; i++) out += i % rate === 0 ? '\u00e9' : sample[i];
  return out;
}

// ASCII everywhere except the last three bytes of every chunk, which defeats any strategy that
// scans the whole buffer and bails out on the first non-ASCII byte it finds.
function asciiChunksWithNonAsciiTails(byteLength, chunkBytes) {
  var out = tile(SSE_FRAME, byteLength);
  var tail = encoder.encode('\u6c34');
  for (var end = chunkBytes; end <= byteLength; end += chunkBytes) {
    out[end - 3] = tail[0];
    out[end - 2] = tail[1];
    out[end - 1] = tail[2];
  }
  return out;
}

var payloads = [
  { name: 'pure ASCII (SSE/JSON)', bytes: tile(SSE_FRAME, PAYLOAD_BYTES) },
  { name: 'ASCII with 1% accents', bytes: tile(season(SSE_FRAME, 100), PAYLOAD_BYTES) },
  { name: 'CJK (worst case)', bytes: tile('\u6c34\u66dc\u65e5\u672c\u8a9e', PAYLOAD_BYTES) },
  { name: 'emoji (worst case)', bytes: tile('\ud83c\udf89\ud83d\ude80\ud83d\udd25', PAYLOAD_BYTES) },
  {
    name: 'ASCII chunks with non-ASCII final bytes',
    bytes: asciiChunksWithNonAsciiTails(PAYLOAD_BYTES, CHUNK_BYTES),
  },
];

// Streams the payload through one decoder exactly as a `fetch` body reader would.
function decodeStreaming(Decoder, bytes) {
  var decoder = new Decoder('utf-8');
  var chars = 0;
  for (var off = 0; off < bytes.length; off += CHUNK_BYTES) {
    var end = off + CHUNK_BYTES < bytes.length ? off + CHUNK_BYTES : bytes.length;
    chars += decoder.decode(bytes.subarray(off, end), { stream: true }).length;
  }
  return chars + decoder.decode().length;
}

function best(Decoder, bytes) {
  var ms = Infinity;
  var chars = -1;
  for (var r = 0; r < RUNS; r++) {
    var started = Date.now(); // Hermes has no process.hrtime.
    var produced = decodeStreaming(Decoder, bytes);
    var elapsed = Date.now() - started;
    if (elapsed < ms) ms = elapsed;
    if (chars >= 0 && produced !== chars) throw new Error('output length varied between runs');
    chars = produced;
  }
  return { ms: ms, chars: chars };
}

var megabytes = PAYLOAD_BYTES / (1024 * 1024);
var rate = function (ms) {
  return (megabytes / (ms / 1000)).toFixed(1);
};

log('| payload | unpatched | patched | change |');
log('| --- | --- | --- | --- |');
for (var p = 0; p < payloads.length; p++) {
  var before = best(Unpatched, payloads[p].bytes);
  var after = best(Patched, payloads[p].bytes);
  // The whole benchmark is meaningless if the two revisions disagree about the output.
  if (before.chars !== after.chars) {
    throw new Error('output mismatch on ' + payloads[p].name);
  }
  log(
    '| ' + payloads[p].name +
      ' | ' + before.ms + ' ms (' + rate(before.ms) + ' MB/s)' +
      ' | ' + after.ms + ' ms (' + rate(after.ms) + ' MB/s)' +
      ' | ' + (before.ms / after.ms).toFixed(2) + 'x |'
  );
}
```

</details>

## 4. On device

Verified in a release build of our app on a physical device against a local SSE firehose. The JS thread FPS regression described above is recovered.

# Checklist


Ignigena:
I was keeping this in Draft just to give our app in Production a bit more time in case we noticed any unusual behavior or other performance concerns. Things are looking really stable so I'm going to mark this as Ready for review now 😄

kitten:
Just to note, there's a competing/similar optimisation at #48791. However, this PR and the other have prompted me to fully rewrite the current decoder implementation at #48877

Beyond fast-paths, the implementation is generally not very efficient, I've found, and I was simply not aware of it (as in, I hadn't read it before), so a general rewrite with some Hermes-specific optimisations is likely going to do much better here

PR #48636, opened 2026-08-07
[dev-launcher][updates][Android] Fix relative manifest URL resolution without base path

hsource:
# Why

Fixes a regression introduced by #47255 where Android development builds crash on launch with:

```
Invalid URL port: "8081apps:80"
```

This happens when loading a project served from a monorepo subpath (e.g. `apps/bare-expo` in this repo). The dev launcher sends `forwarded`/`x-forwarded-host` headers, so Expo CLI responds with a relative `launchAsset.url` like `apps/bare-expo/index.bundle?...`. That URL is then resolved against the path-less dev server URL (`http://<ip>:8081`).

Android's `java.net.URI.resolve()` (Harmony-derived, RFC 2396) does not insert the missing `/` between authority and path when the base URL has an empty path, producing invalid URLs like `http://192.168.1.5:8081apps/bare-expo/index.bundle?...`. Downstream, this malformed URL causes the packager websocket connection to fail.

Related: #47255

# How

Replaced `java.net.URI.resolve()` with `okhttp3.HttpUrl.resolve()` in both places that resolve relative manifest URLs on Android:

- `expo-dev-launcher`: `DevLauncherManifestParser.resolveUrl`
- `expo-updates`: `ExpoUpdatesUpdate.resolveUrl`

okhttp's URL resolution is RFC 3986 compliant and correctly inserts the `/`, matching modern desktop JDK behavior. okhttp is already a dependency in both packages.

Added regression tests covering the path-less base URL case.

# Test Plan

- [x] `./gradlew :expo-dev-launcher:testDebugUnitTest --tests "expo.modules.devlauncher.launcher.manifest.DevLauncherManifestParserTest"` (from `apps/bare-expo/android`)
- [x] Manual: `cd apps/bare-expo && pnpm android --device` — app should launch without the `Invalid URL port: "8081apps:80"` crash

# Checklist

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/expo-dev-launcher/CHANGELOG.md
+++ b/packages/expo-dev-launcher/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ### 🐛 Bug fixes
 
+- [Android] Fix relative manifest URL resolution when the dev server URL has no path, which caused invalid bundle URLs like `http://host:8081apps/project/index.bundle` in monorepo projects. (by [@hsource](https://github.com/hsource))
 - [iOS] Use `RCTPlatformName` instead of hardcoding `ios` when requesting bundles from Metro. ([#46443](https://github.com/expo/expo/pull/46443) by [@gabrieldonadel](https://github.com/gabrieldonadel))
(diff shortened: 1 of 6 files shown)


alanjhughes:
Thanks @hsource but this was already addressed by https://github.com/expo/expo/pull/48625

hsource:
Whoa! So fast - thanks!

PR #48630, opened 2026-08-06
[ios][media-library] Include location in getAssetsAsync batch results

hsource:
# Why

While working with photos with location, we realized iOS `getAssetsAsync` doesn't return location. On iOS, the location info is really cheap to get though, so we thought it'd be more efficient to have it be returned even if resolveWithFullInfo is not set on iOS.

**Credits:** this PR was created based on internal work at Wanderlog by @fractalbeauty

# How

- Include `PHAsset.location` in `exportAsset()` so `getAssetsAsync` returns GPS coordinates without calling `getAssetInfoAsync` per asset.
- Add `location?: Location` to the legacy `Asset` type when batch results include coordinates.
- **Side fix:** Fix `exportLocation()` to return numeric `latitude`/`longitude` values, matching the documented `Location` type (previously returned strings).
- **Side fix:** the MediaLibrary "supports sorting" test failed when there was a large number of images, as it only fetches one page. This fixes that test
- **Side fix:** add test flag for podxpec to fix `pnpm expotools native-unit-tests` for expo-media-library; without it, I get linker errors due to missing C++ basic library symbols

Wanderlog ([app link](https://apps.apple.com/us/app/wanderlog-travel-planner/id1476732439)) uses this for journal photo picker flows that scan large libraries (~10k assets). Internal context: GitLab !10315, !10800.

# Test Plan

- [x] Added Swift unit tests for `exportLocation` in `ios/Tests/MediaLibraryTests.swift`
- [x] `pnpm expotools native-unit-tests --platform ios --packages expo-media-library`
- [x] `cd apps/bare-expo && pnpm test:ios`
- [x] Manual in bare-expo: `cd apps/bare-expo && pnpm ios` → Test suite tab → run **MediaLibrary**, or APIs tab → **MediaLibrary** on a device/simulator with geotagged photos

| Test result | MediaLibrary API timing (before) | MediaLibrary API timing (after) |
| --- | --- | --- |
| image | image | image |

### Performance info (before and after)

This only increases time by 15% in the MediaLibrary test app

- 10 runs without change location for 20k images: **average 954ms**
  - 952, 953, 1006, 951, 942, 946, 937, 929, 990, 936
- 10 runs with location data for 20k images: **average 1097ms**
  - 1198, 1090, 1074, 1118, 1092, 1093, 1065, 1102, 1077, 1056

# Checklist

[...]

PR #48637, opened 2026-08-07
[android][media-library] Parallelize EXIF location reads in getAssetsAsync for 6x speedup

hsource:
# Why

We noticed that fetching detailed asset data when `resolveWithFullInfo` was set to `true` was quite slow on Android. The bottleneck was the sequential per-file EXIF and location reads inside `putAssetsInfo` — each image required opening the file and calling into MediaProvider via Binder IPC (~15ms per image). That being said, the old code did this entirely linearly, with no parallelism, which was inefficient since most time was spent waiting on I/O.

**Credits:** this was based off of internal work at Wanderlog by @robin-pham

# Impact

**For a read of 300 photos**

- Before: 5-6 seconds
- After: 500-800ms

# How

- Refactored `putAssetsInfo` to copy cursor rows into `AssetRowData` in batches (up to 5000 rows), then process each batch in parallel on a dedicated `Dispatchers.IO.limitedParallelism(32)` dispatcher.
- Parallel workers perform EXIF parsing, dimension resolution, and GPS location reads (`getExifLocationForUri` / legacy path) without touching the cursor, which is not thread-safe.
- Concurrency is capped at 32 based on experimentation — higher values mostly queue on Binder's shared thread pool rather than improving throughput.
- Added unit test coverage for `resolveWithFullInfo=true` and cursor pagination preservation.
- Updated the native-component-list Media Library screen and bare-expo config to make manual performance testing easier (page size selector, fetch timing display, `resolveWithFullInfo` toggle on Android).

# Test Plan

- [x] Android native unit tests: `pnpm expotools native-unit-tests --platform android --packages expo-media-library`
- [x] Test-suite E2E on Android: `cd apps/bare-expo && pnpm test:android`
- [x] Manual performance comparison: `cd apps/bare-expo && pnpm android`

| Test results | Media Library fetch timing (before) | Media Library fetch timing (after) |
| --- | --- | --- |
| image | image | image |

# Checklist

### DIFF
--- a/packages/expo-media-library/android/src/main/java/expo/modules/medialibrary/assets/AssetUtils.kt
+++ b/packages/expo-media-library/android/src/main/java/expo/modules/medialibrary/assets/AssetUtils.kt
@@ -143,7 +143,42 @@ suspend fun putAssetsInfo(
   val bundles = withContext(exifReadDispatcher) {
     rows.map { row ->
       async {
-        buildAssetBundle(contentResolver, row, resolveWithFullInfo)
+        val localUri = "file://${row.path}"
+        var exifInterface: ExifInterface? = null
+        if (resolveWithFullInfo && row.mediaType == MediaStore.Files.FileColumns.MEDIA_TYPE_IMAGE) {
+          try {
+            exifInterface = ExifInterface(row.path)
+          } catch (e: IOException) {
+            Log.w("expo-media-library", "Could not parse EXIF tags for $localUri")
+            e.printStackTrace()
+          }
+        }
+        val (width, height) = getAssetDimensions(contentResolver, row, exifInterface)
+        val asset = Bundle().apply {
+          putString("id", row.assetId)
+          putString("filename", row.filename)
+          putString("uri", localUri)
+          putString("mediaType", exportMediaType(row.mediaType))
+          putLong("width", width.toLong())
+          putLong("height", height.toLong())
+          putLong("creationTime", row.creationTime)
+          putDouble("modificationTime", row.modificationTime * 1000.0)
+          putDouble("duration", row.durationMs / 1000.0)
+          putString("albumId", row.albumId)
+        }
+        if (resolveWithFullInfo && exifInterface != null) {
+          getExifFullInfo(exifInterface, asset)
+
+          val location = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+            val photoUri = Uri.withAppendedPath(MediaStore.Images.Media.EXTERNAL_CONTENT_URI, row.assetId)
+            getExifLocationForUri(contentResolver, photoUri)
+          } else {
+            getExifLocationLegacy(exifInterface)
+          }
+          asset.putParcelable("location", location)
+          asset.putString("localUri", localUri)
+        }
+        asset
       }
     }.awaitAll()
   }
@@ -210,34 +245,28 @@ fun getExifLocationLegacy(exifInterface: ExifInterface): Bundle? {
 }
 
 /**
- * Gets image/video dimensions
+ * Gets image/video dimensions from the already-read cursor columns.
+ *
+ * For videos, prefers MediaStore `WIDTH`/`HEIGHT`/`ORIENTATION` (see [1])
+ * and falls back to MediaMetadataRetriever when the scanner has not indexed
+ * the file yet. For images, uses the same cursor columns with
+ * BitmapFactory and EXIF orientation handling when dimensions are missing.
+ *
+ * [1]: https://developer.android.com/reference/android/provider/MediaStore.MediaColumns#WIDTH
+ *
  * @return Pair of integers: width and height, respectively
  */
-@Throws(IOException::class)
-fun getAssetDimensionsFromCursor(
+private fun getAssetDimensions(
   contentResolver: ContentResolver,
-  exifInterface: ExifInterface?,
-  cursor: Cursor,
-  mediaType: Int,
-  localUriColumnIndex: Int
+  row: AssetRowData,
+  exifInterface: ExifInterface?
 ): Pair<Int, Int> {
-  val uri = cursor.getString(localUriColumnIndex)
-  if (mediaType == MediaStore.Files.FileColumns.MEDIA_TYPE_VIDEO) {
-    // Fast path: read dimensions from MediaStore cursor (no file I/O).
-    // MediaStore populates these when the media scanner indexes the file.
-    val widthIndex = cursor.getColumnIndex(MediaStore.MediaColumns.WIDTH)
-    val heightIndex = cursor.getColumnIndex(MediaStore.MediaColumns.HEIGHT)
-    val width = cursor.getInt(widthIndex)
-    val height = cursor.getInt(heightIndex)
-    if (width > 0 && height > 0) {
-      val orientationIndex = cursor.getColumnIndex(MediaStore.MediaColumns.ORIENTATION)
-      val orientation = cursor.getInt(orientationIndex)
-
-      return maybeRotateAssetSize(width, height, orientation)
+  if (row.mediaType == MediaStore.Files.FileColumns.MEDIA_TYPE_VIDEO) {
+    if (row.width > 0 && row.height > 0) {
+      return maybeRotateAssetSize(row.width, row.height, row.orientation)
     }
-
     // Slow fallback for files not yet indexed by the media scanner.
-    val videoUri = Uri.parse("file://$uri")
+    val videoUri = Uri.parse("file://${row.path}")
     try {
       contentResolver.openAssetFileDescriptor(videoUri, "r").use { photoDescriptor ->
         MediaMetadataRetriever().use { retriever ->
@@ -255,23 +284,20 @@ fun getAssetDimensionsFromCursor(
     } catch (e: NumberFormatException) {
       Log.e("expo-media-library", "MediaMetadataRetriever unexpectedly returned non-integer: ${e.message}")
     } catch (e: FileNotFoundException) {
-      Log.e("expo-media-library", "ContentResolver failed to read $uri: ${e.message}")
+      Log.e("expo-media-library", "ContentResolver failed to read ${row.path}: ${e.message}")
     } catch (e: RuntimeException) {
       Log.e("expo-media-library", "MediaMetadataRetriever finished with unexpected error: ${e.message}")
     }
+    return Pair(0, 0)
   }
 
-  val widthIndex = cursor.getColumnIndex(MediaStore.MediaColumns.WIDTH)
-  val heightIndex = cursor.getColumnIndex(MediaStore.MediaColumns.HEIGHT)
-  val orientationIndex = cursor.getColumnIndex(MediaStore.Images.Media.ORIENTATION)
-  var width = cursor.getInt(widthIndex)
-  var height = cursor.getInt(heightIndex)
-  var orientation = cursor.getInt(orientationIndex)
+  var width = row.width
+  var height = row.height
+  var orientation = row.orientation
 
-  // If the image doesn't have the required information, we can get them from Bitmap.Options
-  if (mediaType == MediaStore.Files.FileColumns.MEDIA_TYPE_IMAGE && (width <= 0 || height <= 0)) {
+  if (width <= 0 || height <= 0) {
     val options = BitmapFactory.Options().apply { inJustDecodeBounds = true }
-    BitmapFactory.decodeFile(uri, options)
+    BitmapFactory.decodeFile(row.path, options)
     width = options.outWidth
     height = options.outHeight
   }
@@ -280,6 +306,7 @@ fun getAssetDimensionsFromCursor(
       ExifInterface.TAG_ORIENTATION,
       ExifInterface.ORIENTATION_NORMAL
     )
+    // Full list of orientations are here: https://developer.android.com/reference/android/media/ExifInterface#summary
     if (exifOrientation == ExifInterface.ORIENTATION_ROTATE_90 ||
       exifOrientation == ExifInterface.ORIENTATION_ROTATE_270 ||
       exifOrientation == ExifInterface.ORIENTATION_TRANSPOSE ||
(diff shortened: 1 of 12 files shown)


packages/expo-media-library/android/src/main/java/expo/modules/medialibrary/assets/AssetUtils.kt
```diff
@@ -97,53 +100,98 @@ fun putAssetsInfo(
...
+    // parallel across exifReadDispatcher instead of sequentially.
+    //
+    // - EXIF location reads cost ~15ms per image sequentially. They are independent per-file I/O,
+    //   so parallelizing them speeds up large queries by several times.
+    // - Cursor access is not thread-safe. Cursor columns are copied into AssetRowData in batches
+    //   before parallel file work; workers never touch the cursor.
+    val bundles = withContext(exifReadDispatcher) {
+      rows.map { row ->
+        async {
+          val localUri = "file://${row.path}"
```
hsource, line 154: This section of code is largely moved unchanged from the version that directly processed it from cursors. See [this diff](https://wengier.com/LinkableDiff/#zVfbbhoxEJ23SvmK7aqqoCKb9rVSW6UlqVCBpEl6e0KES0AlgCBpGkX5954Z79VrexcIKrJgd-3x-Mzd45En4w91aYJnl5YYA7qhBvXx_Q6_Ht3SArMz_Ad0Javn-C1oTFN8V_Dsg36K_wH9pSrthVwTvnM8b2hUmuMEaz3Z-01m3dx1anWKT0O8T7DrLR1gvIhR-AYe16Drg75LF3QPyoEDK6O5Acr8HjvOBZ48PwYqtX-AuSFWeoLQoyPr6ocQyxRYJhgJZ6b3gGQBKsY4wVkD-oF5lvM4pGeOQ6x69BLDJiuf0IpX2B4z4RuAj9LjMvX-SU67Ba-pzLeAv46TDqkDrr_oFN8dfLcw8xnvVXB_SCHnwRa_N8x7Tl0pXdi1VYntXNX4Poo9ea0nvlgJNd-gE-HHu-dYHUO2qRExjyZWryDxHfb7gnOOmf2UVvehmTFdAlNXJPSpJv7IOmOL9MWWM5ykImMhMccy_QSWY9FMF2cs8TYUK3jwXt3L_Zx8Sm8BeKpIUlHFWvkNmyxi_eT1smd8j2KjAlk5xlmnLMkIXMbAN8JMVeyR5aei5DDOJXVQs3bYU5Rul5Bygbdr6CSJrwrwzUJ78v9ZxqsXcrbdL2patNaMnl4z5At3dklyovK8jziD6SehLgNQzDEmRl-eg1rPbr5oU3lFNuPqljHvjvLaFLtZQj8nuzmr6vvScpc7-Tb0PJMWy3HIW8SP7ToX7DepLBTRmHKtDXNTfEidlvitOiX5DiS_JbRluKX93s9Fwjoce5Jhu3HOuQgjxW7RZLe-ty7vZaxalzx0GXow22QmumX_6K2JxsbDhMqjV_i9odcyAvzK4ewLgoS3HVdUofUdWRQHpVDoHswefymVrxFHcXHsZXfZbPS4cW131c3nmVuEqb4p5FFt1c-oFGTfJJexTFnO-l0tsojKqZGsnFvHUiED-g4cZ9AiV-c2vs_hD1_kTtGGZ3r0Ps7G5h0d1JYT7DnCTEBfrfU8uaOOpCbPMrdI9R5I5mC9H0quH4T26-Oeo-4aFcvdqQGLd0Wv6pYU0QRS7S8EcRtcmyHetszxfwdnn2F_cZXIW66paflY8ChZylRZXRemm9QAO9S9xaRVN54mVq_Ec-6dXuW6qXgZvQRhtJ7Kfaon6Lqp3KH7XbaGJXe-Iv56Nsjfysy18VlOQ6aoKOpgHuTWdCdolN89ZroZnfd2uo70WC9LRVLsSieSHq6uJD026VB0K5oiOh1rm3ct6bHbHUxWx6t3M658UTRfvuMp1-WUybXKF9y9jU3Gp-lR0qOoX4k8t7giuTmae5iIu766Ov9VOpVibut2LfY8l7-b51E8VSfj5rxpV-Pm7u5wIv3YqIrPWq-XSezipi7uVIoRmbuWCIG-2pKcWdSZFHusqUtJYje9ZuNty5H_oy9Jj-31KOmxu_2KCeUu9y6rV4y8lbfdz2T9vqi3KY9z3T6nTBx6tP3-x33Wur2QLd_8Aw)

packages/expo-media-library/android/src/main/java/expo/modules/medialibrary/assets/AssetUtils.kt
```diff
@@ -97,53 +100,98 @@ fun putAssetsInfo(
...
-      putLong("height", height.toLong())
-      putLong("creationTime", cursor.getLong(creationDateIndex))
-      putDouble("modificationTime", cursor.getLong(modificationDateIndex) * 1000.0)
-      putDouble("duration", cursor.getInt(durationIndex) / 1000.0)
-      putString("albumId", cursor.getString(albumIdIndex))
+
+  // We limit batches to 5000 since that should keep [AssetRowData] memory
+  // memory usage low.
+  //
+  // Large `limit` values would otherwise hold every [AssetRowData] in memory
```
Wenszel: I wonder if the batching is necessary here. Did you observe any OOM problems with the previous approach? Otherwise, we’ll leave code that future readers might interpret as evidence that there used to be a memory problem without batching. The biggest memory bottleneck is still allocating the result array.

packages/expo-media-library/android/src/main/java/expo/modules/medialibrary/assets/AssetUtils.kt
```diff
@@ -97,53 +100,98 @@ fun putAssetsInfo(
...
-      asset.putParcelable("location", location)
-      asset.putString("localUri", localUri)
+    // When resolveWithFullInfo is true, per-file EXIF data (including GPS location) is read in
+    // parallel across exifReadDispatcher instead of sequentially.
+    //
+    // - EXIF location reads cost ~15ms per image sequentially. They are independent per-file I/O,
+    //   so parallelizing them speeds up large queries by several times.
+    // - Cursor access is not thread-safe. Cursor columns are copied into AssetRowData in batches
+    //   before parallel file work; workers never touch the cursor.
+    val bundles = withContext(exifReadDispatcher) {
```
Wenszel: I think we should split the processing with and without `resolveWithFullInfo`. We should only use the dispatcher when resolving with full info. My tests show that the call without the flag takes 1.5x longer on 3,000 assets because of the context-switching overhead.

packages/expo-media-library/android/src/main/java/expo/modules/medialibrary/assets/AssetUtils.kt
```diff
@@ -207,34 +255,30 @@ fun getExifLocationLegacy(exifInterface: ExifInterface): Bundle? {
...
+ * and falls back to MediaMetadataRetriever when the scanner has not indexed
+ * the file yet. For images, uses the same cursor columns with
+ * BitmapFactory and EXIF orientation handling when dimensions are missing.
+ *
+ * [1]: https://developer.android.com/reference/android/provider/MediaStore.MediaColumns#WIDTH
+ *
  * @return Pair of integers: width and height, respectively
  */
-@Throws(IOException::class)
-fun getAssetDimensionsFromCursor(
```
Wenszel, line 214: The review bot found that the change modifies the return value of this function. Could you look into it?

Wenszel (changes requested):
Hey, thank you so much for the PR! I ran the test screen with 10,000 assets, and the performance improvement is really impressive 🚀 I left a few points to discuss regarding the current implementation.

packages/expo-media-library/android/src/main/java/expo/modules/medialibrary/assets/AssetUtils.kt
hsource: Yeah, I never saw issues even with 200k media library sizes I tested on. As such, I'm going to just remove it.

packages/expo-media-library/android/src/main/java/expo/modules/medialibrary/assets/AssetUtils.kt
hsource: Good idea! I basically reverted a lot of the changes so that the bundle is still built in the while loop now. If we do need to post-process, that bundle + additional data is now added to a list to be processed later.

packages/expo-media-library/android/src/main/java/expo/modules/medialibrary/assets/AssetUtils.kt
hsource: **This should be fixed!**

This was a real issue. I also realized that getAssetDimensionsFromCursor can do some I/O work that's slow too, so I refactored it into 2 functions:

1. getAssetDimensionsFromCursorFast that deals with the path that doesn't require reading the file
2. getAssetDimensionsSlow that does do it: I moved this to inside the parallel processing

I audited the code myself and with AI to ensure the behavior is unchanged, and I think there's only 1 benign change.

There is a minor logical change when `resolveWithFullInfo = true` and the cursor seemed to have real width/height/orientation data already:

- **Before:** we would recalculate width/height based on the EXIF orientation
- **Now:** we just keep the width/height based on the cursor orientation

I think this should be fine because the EXIF orientation should always match the cursor orientation.

This change also reduces the number of lines changed by a lot!

hsource:
> * **Dimension fallback lost its image check, so audio assets report width and height of -1** — [`packages/expo-media-library/android/src/main/java/expo/modules/medialibrary/assets/AssetUtils.kt:310`](https://github.com/expo/expo/pull/48637/files#diff-a0bba5481a251ffccf2effef37a5a6b19689b2f23340204968f06e73d3530caeR310) _(correctness)_ · `id:920ed1c261d4`

This is resolved - the image check is now back

> * **batchLocationPreview is never assigned, so the iOS location readout never renders** — [`apps/native-component-list/src/screens/MediaLibrary/MediaLibraryScreen.tsx:389`](https://github.com/expo/expo/pull/48637/files#diff-b2fd34da5c9ca46663d6db2f6298ab0470ad6309f66d9fcc5084e73c835dcf0dR389) _(correctness)_ · `id:9c0c9083d066`

I fixed this too by updating the example screen

hsource:
@Wenszel I think I've fixed the remaining issues! After the latest round of changes, the diff got smaller relative to the original code. Let me know if you still see anything

hsource:
I retested the change and it still seems to be very fast!

image

packages/expo-media-library/src/legacy/MediaLibrary.ts
```diff
@@ -121,6 +121,12 @@ export type Asset = {
    * @platform android
    */
   albumId?: string;
+  /**
+   * GPS location if available. On iOS, `getAssetsAsync` includes this field for all asset types when
```
hsource: I mentioned iOS in this and other comments, but those mentions aren't accurate yet because it depends on #48630 being merged in.

I decided to keep the file the same across both PRs to make resolving conflicts easier once we're merging these in. I can remove iOS mentions if we want though.

hsource:
@barthap and @Wenszel I just rebased the PR and added an Impact section with the numbers after testing before/after. Let me know if there's anything else I need to change!

barthap (commented):
I took a brief look; impressive work! Looks like most findings are addressed. I only have two optional suggestions:

- Could we add explicit assertions that audio assets keep `width` and `height` at `0` in `AssetUtilsTests`? The current test includes an audio row but only checks the video dimensions, so it would not catch that regression.
- The new `MediaLibrary.ts` docs describe iOS behavior added by #48630, which is still open. If this PR might merge first, we should either remove the iOS claims here or make this PR explicitly dependent on #48630. I know you already noted this; I just want to emphasize that these PRs are related and should likely be merged together.

I’m leaving the remaining cursor-versus-EXIF orientation question to @Wenszel. No other concerns from my side.

Wenszel:
Hey, I think we're almost ready to merge - two last things:
1. As @barthap mentioned, the iOS parts of the new `Asset` docs only become true with #48630. Could you drop the iOS mentions from this PR, so that we could land them together with the implementation?
2. Could we keep reading the orientation from EXIF? The orientation column for images has been there since Android API 1 and it should match the EXIF, but we can't guarantee it is always populated in MediaStore, so I'd rather keep the correction as defensive programming - it costs nothing, since the EXIF is already parsed, and it preserves the current behavior.

hsource:
@Wenszel and @barthap Thanks! Go ahead and take another look - I made the last few changes in this commit: https://github.com/expo/expo/pull/48637/changes/46098cbd7797374116e87b97890e4bbeb9cc337d

> Could we add explicit assertions that audio assets keep width and height at 0 in AssetUtilsTests? The current test includes an audio row but only checks the video dimensions, so it would not catch that regression.

I added the test and reran `./gradlew :expo-media-library:testDebugUnitTest --tests "expo.modules.medialibrary.assets.AssetUtilsTests"`

> 1. As @barthap mentioned, the iOS parts of the new `Asset` docs only become true with [[ios][media-library] Include location in getAssetsAsync batch results #48630](https://github.com/expo/expo/pull/48630). Could you drop the iOS mentions from this PR, so that we could land them together with the implementation?

I reverted the docs: all doc changes only relate to Android

> 2. Could we keep reading the orientation from EXIF? The orientation column for images has been there since Android API 1 and it should match the EXIF, but we can't guarantee it is always populated in MediaStore, so I'd rather keep the correction as defensive programming - it costs nothing, since the EXIF is already parsed, and it preserves the current behavior.

I created a new `getImageDimensionsFromExifFast` and call it if EXIF is available and we haven't called it before as part of the missing-dimensions branch of the code.

packages/expo-media-library/android/src/main/java/expo/modules/medialibrary/assets/AssetUtils.kt
```diff
@@ -129,22 +135,89 @@ fun putAssetsInfo(
...
+    //   so parallelizing them speeds up large queries by several times.
+    // - Cursor access is not thread-safe. Paths are copied into pending asset lists during the
+    //   cursor loop before parallel file work; workers never touch the cursor.
+    withContext(exifReadDispatcher) {
+      pendingAdditionalInfoAssets.map { pending ->
+        async {
+          val localUri = "file://${pending.path}"
+          var exifInterface: ExifInterface? = null
+          if (
+            (resolveWithFullInfo || pending.missingDimensions) &&
```
Wenszel:
```suggestion
            resolveWithFullInfo &&
```

Wenszel: Could you please apply this patch with the tests? It would ensure that we haven't changed anything in the EXIF/asset shape logic. Unfortunately, I can't push to this PR since it was opened by the organization :( Thanks!

[exif-orientation-tests.patch](https://github.com/user-attachments/files/31234278/exif-orientation-tests.patch)

hsource: @Wenszel I applied this with no other changes!

hsource:
@Wenszel I applied the patch with no other changes! I'll make future PRs from my personal account for simplicity

Wenszel (approved):
LGTM 🐆

barthap (approved):

PR #48638, opened 2026-08-07
[router] Remove redundant `root` fallback in `useNavigation`

Ubax:
# Why

In `react-navigation` apps, `useNavigation` can be called from outside a screen. In `expo-router` this is not possible since every app is wrapped in the root navigator. 

Therefore the `?? root` fallback in `useNavigation` is not needed and can mask real issues.

# How

1. Remove `?? root` fallback
2. Add tests

# Test Plan

CI

# Checklist

---

### DIFF
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### 🛠 Breaking changes
 
+- Remove `NavigationContainerRefContext` fallback in `useNavigation`. The hook now throws when called outside a navigator, including components rendered via `ExpoRoot`'s `wrapper` prop. ([#48638](https://github.com/expo/expo/pull/48638) by [@Ubax](https://github.com/Ubax))
 - Allow `key` to be `undefined` in the `Descriptor` type and in routes passed to screen `options`, navigator `screenOptions`, and `RouteGroupConfig.screenOptions` callbacks. ([#48596](https://github.com/expo/expo/pull/48596) by [@Ubax](https://github.com/Ubax))
 - Unify JS Tabs, TopTabs, Drawer, and headless tabs with NativeTabs - only screens declared in the layout become visible. ([#48499](https://github.com/expo/expo/pull/48499) by [@Ubax](https://github.com/Ubax))
 - Make `href: null` hide JS tabs and make their routes unreachable, redirecting navigation to the initial tab. ([#48499](https://github.com/expo/expo/pull/48499) by [@Ubax](https://github.com/Ubax))
--- a/packages/expo-router/src/react-navigation/core/__tests__/useNavigation.test.ios.tsx
+++ b/packages/expo-router/src/react-navigation/core/__tests__/useNavigation.test.ios.tsx
@@ -120,7 +120,7 @@ test("gets navigation's parent's parent from context", () => {
   );
 });
 
-test('throws if called outside a navigator', () => {
+test('throws when inside a container but outside any navigator', () => {
   expect.assertions(1);
 
   const TestNavigator = (props: any): any => {
@@ -136,7 +136,7 @@ test('throws if called outside a navigator', () => {
   const Test = () => {
     // eslint-disable-next-line react-hooks/rules-of-hooks
     expect(() => useNavigation()).toThrow(
-      "Couldn't find a navigation object. This is most likely an issue in expo-router."
+      "Couldn't find a navigation object. Make sure the component is rendered inside your app's route tree. This is most likely a bug in expo-router. Please report it at https://github.com/expo/expo/issues."
     );
 
     return null;
@@ -152,13 +152,13 @@ test('throws if called outside a navigator', () => {
   );
 });
 
-test('throws if called outside a navigation context', () => {
+test('throws when outside the navigation container', () => {
   expect.assertions(1);
 
   const Test = () => {
     // eslint-disable-next-line react-hooks/rules-of-hooks
     expect(() => useNavigation()).toThrow(
-      "Couldn't find a navigation object. This is most likely an issue in expo-router."
+      "Couldn't find a navigation object. Make sure the component is rendered inside your app's route tree. This is most likely a bug in expo-router. Please report it at https://github.com/expo/expo/issues."
     );
 
     return null;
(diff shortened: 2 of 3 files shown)


packages/expo-router/src/react-navigation/core/useNavigation.tsx
```diff
@@ -16,15 +15,14 @@ export function useNavigation<
     getState(): NavigationState | undefined;
   },
 >(): T {
-  const root = use(NavigationContainerRefContext);
   const navigation = use(NavigationContext);
 
-  if (navigation === undefined && root === undefined) {
+  if (navigation === undefined) {
```
jakub-agent, line 20: [🏗️ **design**] The PR description says every expo-router app is wrapped in the root navigator, but there is a gap: `ExpoRoot` renders `WrapperComponent` **inside** `UpstreamNavigationContainer` and **outside** `Content` (the root navigator):

```tsx
<UpstreamNavigationContainer ...>
  <ServerContext.Provider value={serverContext}>
    <WrapperComponent>
      <Content />   {/* useNavigationBuilder lives here */}
    </WrapperComponent>
  </ServerContext.Provider>
</UpstreamNavigationContainer>
```

A component rendered by the `wrapper` prop has `NavigationContainerRefContext` but no `NavigationContext`, so it used to get the container ref and now throws. Same for a deprecated standalone `NavigationContainer` (still exported from `react-navigation/native`) with a component between the container and its first navigator.

`wrapper` is undocumented and unused in-tree, so the blast radius is small — but is dropping that case intentional? If so, worth calling out in the changelog.

packages/expo-router/src/react-navigation/core/useNavigation.tsx
```diff
@@ -16,15 +15,14 @@ export function useNavigation<
...
     throw new Error(
-      "Couldn't find a navigation object. Is your component inside NavigationContainer?"
+      "Couldn't find a navigation object. This is most likely an issue in expo-router."
```
jakub-agent: [💡 suggestion] This message tells the user nothing they can act on, and it blames expo-router for what is often their own mistake. `useNavigation` is exported from `expo-router`, so this same throw fires when a component is rendered outside `ExpoRoot` entirely — a custom entry that renders UI as a sibling of `ExpoRoot`, a `'use dom'` component (separate React root), or a second `AppRegistry` root. Those users will open an expo-router issue for their own bug.

Suggest keeping a user-facing hint first, then the internal-bug note:

```ts
"Couldn't find a navigation object. Make sure the component is rendered inside your app's route tree. If it is, this is likely a bug in expo-router: https://github.com/expo/expo/issues"
```

Two smaller points on wording: the codebase convention is "This is likely a bug in expo-router." (`Route.tsx:87`, `ui/Tabs.tsx:161`, `hooks/useRootNavigationState.ts:34`, `link/preview/LinkPreviewContext.tsx:28`) — "an issue in" is a new phrasing; and `NativeBottomTabsNavigator.tsx:180` appends the issues URL, which is the more useful precedent here.

packages/expo-router/src/react-navigation/core/__tests__/useNavigation.test.ios.tsx
```diff
@@ -120,7 +120,7 @@ test("gets navigation's parent's parent from context", () => {
   );
 });
 
-test('gets navigation from container from context', () => {
+test('throws if called outside a navigator', () => {
```
jakub-agent: [🐧 nit] This test and `throws if called outside a navigation context` below now assert the identical message, and the names don't convey what actually differs between them. The real distinction is container present vs. container absent. Suggest `throws when inside a container but outside any navigator` and `throws when outside the navigation container` so the pair is self-explanatory.

packages/expo-router/src/react-navigation/core/useNavigation.tsx
```diff
@@ -16,15 +15,14 @@ export function useNavigation<
...
     );
   }
 
   // FIXME: Figure out a better way to do this
```
jakub-agent: [🐧 nit] This `FIXME` referred to the `navigation ?? root` union that this PR removes — the two branches had different types, hence the double cast. What's left is an ordinary generic cast, so the note no longer points at anything actionable. Good chance to drop it.

jakub-agent (approved):

packages/expo-router/src/react-navigation/core/useNavigation.tsx
Ubax: Unify with other errorrs

packages/expo-router/src/react-navigation/core/useNavigation.tsx
Ubax: Do it

packages/expo-router/src/react-navigation/core/useNavigation.tsx
Ubax: Yes, dropping this case is intentional. The `wrapper` prop is undocumented and unused in-tree, and the breaking change is now called out explicitly in the expo-router changelog.

packages/expo-router/src/react-navigation/core/useNavigation.tsx
Ubax: Removed the stale FIXME and simplified the return to a direct `navigation as T` cast.

packages/expo-router/src/react-navigation/core/__tests__/useNavigation.test.ios.tsx
Ubax: Renamed both tests to distinguish the container-present and container-absent cases.

packages/expo-router/src/react-navigation/core/useNavigation.tsx
Ubax: Updated the error to lead with an actionable route-tree hint and follow the existing expo-router bug-report convention, including the issues URL.

packages/expo-router/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### 🛠 Breaking changes
 
+- Remove `NavigationContainerRefContext` fallback in `useNavigation`. The hook now throws when called outside a navigator, including components rendered via `ExpoRoot`'s `wrapper` prop. ([#48638](https://github.com/expo/expo/pull/48638) by [@Ubax](https://github.com/Ubax))
```
Ubax, line 7:
```suggestion
- Remove root container fallback from `useNavigation`. The hook now throws when called outside a screen or navigator. ([#48638](https://github.com/expo/expo/pull/48638) by [@Ubax](https://github.com/Ubax))
```

PR #48650, opened 2026-08-07
[docs][audio] Fix AudioPlayer.seekTo JSDoc

RasmusKard:
# Why

Docs before change sound like `AudioPlayer.seekTo` is `currentTime += seconds`, but it's actually `currentTime = seconds`.

# How

Changed to same wording as `AudioPlaylist.seekTo`.

# Test Plan

N/A


behenate (approved):

RasmusKard:
Anything else needed from my side? 😊

PR #48656, opened 2026-08-07
[router] remove render time rehydration set state

Ubax:
# Why

# How

# Test Plan

# Checklist


packages/expo-router/src/react-navigation/core/useRouteCache.tsx
```diff
@@ -26,6 +26,15 @@ export function useRouteCache<State extends NavigationState>(routes: State['rout
     const previous = cache.current.get(route.key);
     const { state, ...routeWithoutState } = route;
 
+    if (
+      previous?.params !== routeWithoutState.params &&
```
jakub-agent, line 30: [🚨 **critical**] Reusing the previous `params` object when the new one is shallow-equal also reuses the non-enumerable `CONSUMED_PARAMS` marker attached to it. A repeat navigation whose merged params happen to be shallow-equal is then treated as already consumed, and the nested navigator never applies them.

This is visible in the diff: `__tests__/index.test.ios.tsx` around the `navigates to nested child in a navigator` case now expects `[bar-b, ...]` where it previously expected `[bar-a, ...]` after `navigate('bar', { screen: 'bar-a' })`. That test is not in StrictMode, so the effect is not limited to double rendering.

Mechanism, using that test: after `goBack()`, the parent route `bar` holds `{ screen: 'bar-a', params: { whoa: 'test' } }` with `CONSUMED_PARAMS` set. `navigate('bar', { screen: 'bar-a' })` makes the router build a new params object that merges to the same keys and the same nested `params` reference, so `isRecordEqual` matches, the old object is swapped back in, and `isNestedParamsConsumed` stays true. The child stays on `bar-b`.

The same shape occurs in real apps whenever the nested payload has no per-navigation params — for example tapping a `Link` to `/tabs/a` while a child stack is deeper inside that tab.

If suppressing the reapply is the intent, could the consumed marker be keyed on something other than params object identity (for example, the navigator remembering the last consumed params value) so identity reuse stays a pure memoization win? If it is not the intent, the reuse likely needs to skip params objects carrying `CONSUMED_PARAMS`.

packages/expo-router/src/react-navigation/routers/TabRouter.tsx
```diff
@@ -339,6 +343,149 @@ export function TabRouter({
...
+        }
+
+        return {
+          ...state,
+          routes,
+          index,
+          ...(history ? { history } : {}),
+          ...(state.preloadedRouteKeys
+            ? {
+                preloadedRouteKeys: state.preloadedRouteKeys.filter((key) =>
```
jakub-agent, line 459: [🏗️ **design**] The stale projection only drops keys for routes that no longer exist. The committed `NAVIGATE` path additionally drops the key of the route that just became focused (see the `preloadedRouteKeys` filter in the `NAVIGATE` case, which excludes `updatedState.routes[updatedState.index].key`).

So when nested params focus a tab that was preloaded, the interim render reports that tab as both focused and preloaded, and the committed state does not. `DrawerView` and the `Tabs` clients branch on `preloadedRouteKeys` to decide whether a screen renders detached, so the projected frame can render the focused tab as a preloaded one before `NAVIGATOR_PARAMS_CHANGED` commits.

Since the whole point of the projection is to match what the action will commit, filtering out `nextRoute.key` here too would keep the two in agreement.

packages/expo-router/src/global-state/getRouteInfoFromState.ts
```diff
@@ -49,7 +49,8 @@ export function getRouteInfoFromState(state?: StrictState): UrlObject {
   if (!state) return defaultRouteInfo;
 
   // TODO(@kitten): Review edge-case type safety
-  const index = 'index' in state ? (state.index ?? 0) : 0;
+  const index =
+    'index' in state ? (state.index ?? (state.type === 'stack' ? state.routes.length - 1 : 0)) : 0;
```
jakub-agent, line 53: [🚨 **critical**] The new stack fallback sits behind `'index' in state`, so it only applies when `index` is present with the value `undefined`. Stale states produced by `getPartialState` and `ensureStateKeys` omit the `index` property entirely — `KeyedPartialState` declares it optional and neither helper adds it — so `'index' in state` is `false` and the index falls back to `0`.

For a stale stack that means the bottom of the stack is reported as the focused route instead of the top, which is exactly the case the fallback is meant to fix. `StackRouter.getStateForDeclaredRoutes` uses the unguarded form (`state.index ?? state.routes.length - 1`) for the same reason.

Dropping the `in` check and writing `state.index ?? (state.type === 'stack' ? state.routes.length - 1 : 0)` would cover both shapes. The same applies to the copy inside the `while` loop below, which walks nested `route.state` values that are commonly stale.

packages/expo-router/src/layouts/stack-router.ts
```diff
@@ -436,7 +459,33 @@ export const stackRouterOverride: NonNullable<NativeStackNavigatorProps['UNSTABL
         }
       }
     },
+    getStateForNavigatorParams: (state, params: NavigatorParamsPayload, options) => {
+      if (params.state || state.stale !== false) {
```
jakub-agent, line 463: [🏗️ **design**] For a stale state this delegates to `original.getStateForNavigatorParams`, so the expo-specific stack semantics in the override's `NAVIGATE` branch are skipped — singular filtering, the zoom-transition id handling, and the preloaded-route key reuse below. Because `getStateForAction`'s `NAVIGATOR_PARAMS_CHANGED` case routes through this same method, the skip applies to the committed state too, not just the interim render.

A cold-start deep link into a nested stack is a stale-state path, so this is reachable rather than theoretical. Is delegating intentional here, and if so is it worth a short comment saying which behaviors do not apply before hydration?

packages/expo-router/src/react-navigation/routers/TabRouter.tsx
```diff
@@ -339,6 +343,149 @@ export function TabRouter({
...
+          params.merge && route && currentId === nextId
+            ? { ...config.routeParamList[params.screen], ...route.params, ...params.params }
+            : config.routeParamList[params.screen] !== undefined
+              ? { ...config.routeParamList[params.screen], ...params.params }
+              : params.params;
+        const nextRoute = {
+          ...route,
+          key,
+          name: params.screen,
+          path: params.path ?? route?.path,
```
jakub-agent, line 416: [💡 suggestion] `nextRoute` always defines `path` and `params`, even when both resolve to `undefined`. `StackRouter.getStateForNavigatorParams` deliberately avoids that with the `...(x !== undefined ? { x } : null)` pattern, and this PR applies the same pattern in several other places.

It matters beyond style here: `isRecordEqual` compares `Object.keys().length` first, so a route that gains an `undefined` `path` key stops comparing equal to its cached counterpart in `useRouteCache`, which drops the cached route object and produces a new identity for screens. Matching the `StackRouter` pattern would keep the projection identity-stable.

packages/expo-router/src/react-navigation/routers/TabRouter.tsx
```diff
@@ -339,6 +343,149 @@ export function TabRouter({
       );
     },
 
+    getStateForDeclaredRoutes(state, routeNames, fallbackRouteKey) {
```
jakub-agent, line 346: [💡 suggestion] `addFallbackRouteIfEmpty` is called with `{}` for `routeParamList`, so the fallback route rendered here has no `initialParams`. The `ROUTE_NAMES_CHANGED` case a few hundred lines below passes the real `routeParamList` for the same fallback.

The key matches, so no remount happens, but a screen declaring `initialParams` renders once with `route.params === undefined` and then gets its params on the next commit. `BaseRouter.getStateForDeclaredRoutes` has the same gap when it pushes `{ key: fallbackRouteKey, name: routeNames[0] }`.

The router method signature does not receive `routeParamList` today, so closing this probably means widening the signature — worth deciding explicitly rather than leaving the two paths divergent.

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
```diff
@@ -399,21 +406,29 @@ export function useNavigationBuilder<
...
+    if (isNestedParamsConsumed || route?.params == null) {
+      return undefined;
+    }
+
+    const stateFromParams = getStateFromParams(route.params);
+
+    return {
+      params: route.params,
+      state: stateFromParams ? ensureStateKeys(stateFromParams, router.type) : undefined,
+      routeKey: `${route.params.screen ?? routeNames[0]}-${nanoid()}`,
```
jakub-agent, line 419: [💡 suggestion] Both this memo and the `fallbackRouteKey` memo below mint identity with `nanoid()` inside `useMemo`. React treats `useMemo` as a performance hint and may drop the cached value even when the deps are unchanged, which would produce a different route key on a later render while the params are still unconsumed.

A render-phase `useRef` keyed on the dep values, or deriving the key from something stable, would make the identity a guarantee rather than a memoization side effect.

packages/expo-router/src/react-navigation/routers/types.tsx
```diff
@@ -61,6 +62,52 @@ export type PartialState<State extends NavigationState> = Partial<Omit<State, 's
...
+    stale?: true;
+    routes: KeyedPartialRoute<Route<State['routeNames'][number]>>[];
+  }>;
+
+export type RenderRoute<R extends Route<string>> = Omit<KeyedPartialRoute<R>, 'state'> & {
+  key: string;
+  state?: RenderState<NavigationState>;
+};
+
+export type RenderState<State extends NavigationState> =
```
jakub-agent, line 84: [🏗️ **design**] `RenderState` makes `history`, `preloadedRouteKeys`, and other non-core fields optional for everything that receives navigator state at render time. That reaches the public surface: `IntegrateWithRouterOptions.processState` and `StandardNavigatorCreatePropsFactoryDeps.state` now hand library authors a value where `state.preloadedRouteKeys` can be `undefined`. The `custom-navigators guide example` in `standard-navigation/__tests__/integration.test.ios.tsx` needed `?? []` for exactly this reason.

`unstable_integrateWithRouter` is exported from `expo-router`, so this is a breaking type change for external navigator libraries. Worth a `CHANGELOG.md` entry, and a line in `docs/pages/router/advanced/custom-navigators.mdx` noting that render-time state may be stale. The same doc should probably mention that a custom router overriding `NAVIGATE` semantics must now also implement `getStateForNavigatorParams`, which the JSDoc on that method already states.

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
```diff
@@ -596,8 +611,9 @@ export function useNavigationBuilder<
       // This is necessary for proper screen tracking, URL updates etc.
       // We only notify if the state is different what we already notified
       // Otherwise this goes into a loop when inside `<Activity mode="hidden">`
-      setState(nextState);
-      lastNotifiedStateRef.current = nextState;
+      // Render state has all fields accepted by the stale store shape, with stronger nested keys.
+      setState(nextState as PartialState<State>);
+      lastNotifiedStateRef.current = state;
```
jakub-agent, line 616: [🐧 nit] The guard above compares against `state` and this stores `state`, but the value committed on the line before is `nextState`. Since the effect runs once with an empty dep list the mismatch is inert today, but storing the value that was actually committed would keep the pair readable if the deps ever change.

jakub-agent (changes requested):

This removes render-time `setState` rehydration and replaces it with a render-phase projection (`RenderState`) plus a new router hook, `getStateForNavigatorParams` / `NAVIGATOR_PARAMS_CHANGED`. The direction looks right: rendering no longer commits state from inside render, `useScheduleUpdate` and the batching store are gone, and navigator keys now survive rehydration. Test coverage for the new router methods is good.

Main concern: the new `params` identity reuse in `useRouteCache` changes user-visible navigation behavior. The existing test `navigates to nested child in a navigator` had its snapshot flipped so that a repeated `navigate('bar', { screen: 'bar-a' })` is now a no-op. That is not StrictMode-specific, and it looks like a side effect rather than the goal. Details inline.

Other items are smaller: a `preloadedRouteKeys` divergence between the stale projection and the committed `NAVIGATE` in `TabRouter`, an unreachable stack fallback in `getRouteInfoFromState`, and asymmetric handling of stale state in the expo stack router override.

Process notes: the PR body is still the unfilled template, and there is no `CHANGELOG.md` entry. This change alters `NavigationContainerRef.getState()` (now returns the root state), the `Router` interface, and the `unstable_integrateWithRouter` prop types, so a changelog line seems warranted.

I did not run the test suite; findings come from reading the diff and the surrounding code at `b56d4e1`.

PR #48660, opened 2026-08-07
[cli] Treat a disconnected client as a cancelled manifest request

### LINKED ISSUE #48584: Aborted native manifest request logs ERR_STREAM_UNABLE_TO_PIPE

https://github.com/lighthold/expo-sdk57-stream-repro

This replaces #48581 with a deterministic public reproduction and the full stack trace.

### Steps to reproduce

The issue was originally observed on iOS in an SDK 57 development build while the client was reconnecting as Metro restarted. The repository reduces that timing-dependent path to one local HTTP request, so no simulator is required:

1. Clone the reproduction and install its exact pinned dependencies with pnpm.
2. Run `pnpm start`.
3. In another terminal, run `pnpm reproduce`.

The reproduction sends a native manifest request and closes the TCP connection before Expo finishes resolving the runtime version.

Expected: Expo CLI treats the disconnected client as a cancelled request and remains quiet.

Actual: Expo CLI remains alive but logs:

```text
Error [ERR_STREAM_UNABLE_TO_PIPE]: Cannot pipe to a closed or destroyed stream
    at pipelineImpl (node:internal/streams/pipeline:264:15)
    at node:stream/promises:31:5
    at new Promise (<anonymous>)
    at pipeline (node:stream/promises:20:10)
    at ExpoGoManifestHandlerMiddleware.handleRequestAsync (.../ManifestMiddleware.js:327:44)
    at async HistoryFallbackMiddleware.internalMiddleware [as indexMiddleware] (.../ExpoMiddleware.js:90:24)
```

`ExpoGoManifestHandlerMiddleware.handleRequestAsync` awaits manifest generation and then unconditionally passes the response body and Node response to `stream/promises.pipeline`. If the HTTP client disconnected during that await, the destination is already destroyed. A disconnect guard and cancellation handling around that pipeline should prevent the error.

### Environment

```text
System:
  OS: macOS 27.0 (26A5388g), Apple silicon
  Shell: zsh 5.9.1
Binaries:
  Node: 24.18.0
  pnpm: 11.10.0
Managers:
  CocoaPods: 1.16.2
  Watchman: 2026.06.29.00
SDKs:
  iOS SDK: 27.0
IDEs:
  Xcode: 27.0 (27A5228h)
Packages:
  expo: 57.0.8
  @expo/cli: 57.0.10
  expo-updates: 57.0.10
  react-native: 0.86.0
  react: 19.2.3
```

### Expo Doctor Diagnostics

The reproduction was validated by linking the already-installed exact dependencies, without downloading or changing packages. `expo-doctor@latest` was therefore not executed. Expo CLI's built-in startup validation reports only that newer compatible patch versions are available; the error reproduces with the versions pinned in the repository.

ahmdshrif:
# Why

Fixes #48584.

`ManifestMiddleware.handleRequestAsync` resolves the manifest and then unconditionally writes it to the Node response:

https://github.com/expo/expo/blob/main/packages/@expo/cli/src/start/server/middleware/ManifestMiddleware.ts

```ts
const response = await this._getManifestResponseAsync(options);
// Convert `Response` to node:http response
if (typeof res.setHeaders === 'function') {
  res.setHeaders(response.headers);
} else { ... }
if (response.body) {
  await pipeline(Readable.fromWeb(response.body as any), res);
}
```

`_getManifestResponseAsync` is slow — it reads the app config, resolves the signed-in user, resolves assets, and for dev clients resolves the runtime version via `expo-updates`. Clients routinely hang up during that await: a dev client reconnecting while Metro restarts is the reported case. By the time the manifest is ready the socket is gone, `res` is destroyed, and `pipeline` throws:

```text
Error [ERR_STREAM_UNABLE_TO_PIPE]: Cannot pipe to a closed or destroyed stream
    at pipelineImpl (node:internal/streams/pipeline:264:15)
    at ExpoGoManifestHandlerMiddleware.handleRequestAsync (.../ManifestMiddleware.js:327:44)
    at async HistoryFallbackMiddleware.internalMiddleware [as indexMiddleware] (.../ExpoMiddleware.js:90:24)
```

`ExpoMiddleware.getHandler` catches it, calls `Log.exception(error)`, and then tries to write a 500 to the same dead socket. So a cancelled request is reported to the user as a server error with a Node-internal stack trace, in the exact situation (a reconnect storm) where they are already trying to work out why their client will not connect.

# How

Treat a disconnected client as a cancelled request rather than an error, matching what `DevToolsPluginMiddleware` already does for the same class of failure:

https://github.com/expo/expo/blob/main/packages/@expo/cli/src/start/server/middleware/DevToolsPluginMiddleware.ts

```ts
if (res.headersSent || res.writableEnded || res.destroyed || request?.signal.aborted) {
```

- Bail out right after the manifest resolves if `res` is already destroyed or ended, before any header is written. This is the reported `ERR_STREAM_UNABLE_TO_PIPE` path.
- Wrap the `pipeline` call so a disconnect that happens *part-way through* the response (`ERR_STREAM_PREMATURE_CLOSE`) is swallowed too. Same root cause, same race, one line apart — a fix for only the first would leave a smaller window open.

The check is deliberately narrow: it only swallows the error when the response object itself reports that it is gone, so genuine failures from `_getManifestResponseAsync` or from the stream still propagate to `ExpoMiddleware.getHandler` and are still logged and still return a 500. Behaviour for connected clients is unchanged.

# Test Plan

Two tests added to the existing `getHandler` suite in `ManifestMiddleware-test.ts`, covering both races:

- client disconnects while the manifest resolves — asserts no headers are written and nothing is logged
- client disconnects while the manifest streams — asserts nothing is logged

```
$ pnpm jest src/start/server/middleware/__tests__

PASS @expo/cli src/start/server/middleware/__tests__/ManifestMiddleware-test.ts
  getHandler
    ✓ resolves successfully
    ✓ returns error info in the response
    ✓ quietly skips the response when the client disconnects while the manifest resolves
    ✓ quietly skips the response when the client disconnects while the manifest is streaming

Test Suites: 18 passed, 18 total
Tests:       177 passed, 177 total
```

Both new tests fail against `main` (`Log.exception` is called with `ERR_STREAM_PREMATURE_CLOSE`, and headers are written to the dead response) and pass with the fix.

The unit tests use a `PassThrough` stand-in for the response, so to confirm the reported `ERR_STREAM_UNABLE_TO_PIPE` specifically I also ran the same sequence against a real `http.Server` — a client that hangs up its socket 30ms into a handler that takes 150ms to produce the manifest, with the pre-fix and post-fix bodies of `handleRequestAsync`:

```
=== node v22.22.2 (no fix) ===  LOGGED EXCEPTION -> ERR_STREAM_PREMATURE_CLOSE | Premature close
=== node v22.22.2 (fixed)  ===  cancelled request handled quietly (res.destroyed=true)
=== node v24.12.0 (no fix) ===  LOGGED EXCEPTION -> ERR_STREAM_UNABLE_TO_PIPE | Cannot pipe to a closed or destroyed stream
=== node v24.12.0 (fixed)  ===  cancelled request handled quietly (res.destroyed=true)
=== node v25.2.1 (no fix) ===  LOGGED EXCEPTION -> ERR_STREAM_UNABLE_TO_PIPE | Cannot pipe to a closed or destroyed stream
=== node v25.2.1 (fixed)  ===  cancelled request handled quietly (res.destroyed=true)
```

That matches the report: the reporter is on Node 24.18.0, where `pipeline` rejects eagerly with `ERR_STREAM_UNABLE_TO_PIPE`; on Node 22 the same race surfaces one tick later as `ERR_STREAM_PREMATURE_CLOSE`. `res.destroyed` is already `true` in both, so the guard covers both.

`pnpm tsc -p tsconfig.json --noEmit` reports no new errors (the only output is the pre-existing `@expo/prebuild-config` resolution errors, present on `main` in the same tree).

# Checklist


ahmdshrif:
Thanks — the warning was correct, and I've pushed a fix in bf470ee.

The `catch` gated on `res.destroyed`, which `pipeline` sets itself on any failure, so the `throw error` branch was unreachable and a genuine mid-stream failure (the lazily encoded Expo Go multipart body being the realistic case) was dropped with no log and no 500. Confirmed against a real `http.Server` — source throws while the client stays connected:

```
CATCH: code=undefined message=multipart encoding blew up | res.destroyed=true res.writableEnded=false res.writableFinished=false
```

One correction to the suggested remediation, though: **recording the disconnect before the pipe does not work.** Destroying `res` destroys the socket, which destroys `req`, so every response/request/socket-level signal is identical in both cases. Measured on Node 22 and 24:

| scenario | `error.code` | `res.destroyed` | `req.destroyed` | `socket.destroyed` | `res.once('close')` flag |
| - | - | - | - | - | - |
| client hangs up mid-stream | `ERR_STREAM_PREMATURE_CLOSE` | true | true | true | true |
| body throws, client connected | `undefined` | true | true | true | true |

Only the error itself distinguishes them, so the fix takes the other half of your suggestion and matches `error.code` against `ERR_STREAM_PREMATURE_CLOSE`, `ERR_STREAM_UNABLE_TO_PIPE`, `ECONNRESET` and `EPIPE`. A client hangup mid-stream lands on `ERR_STREAM_PREMATURE_CLOSE` on Node 22, 24 and 25; an error from the body keeps its own identity and now propagates to `getHandler` as before.

Also added the test you asked for — a manifest body that errors while the client stays connected, asserting `Log.exception` runs and the response is a 500. Reverting just the discriminator back to `!isResponseCancelled(res)` makes that test fail, so it does cover the hole. Middleware suite is 18 suites / 178 tests green.

One residual gap worth flagging rather than hiding: a *source-side* premature close would still be swallowed, since it shares `ERR_STREAM_PREMATURE_CLOSE` with a client hangup. Much narrower than before, but not zero. Happy to drop the mid-stream `catch` entirely if you'd rather keep the diff to the reported bug — the early return before the header write is what fixes #48584, and it's unaffected either way.

kitten:
Ideally this should just be replaced with: https://github.com/expo/expo/blob/d0cca05c3881346c23554da53ff73333690f0562/packages/expo-server/src/vendor/http.ts#L160

That's already exposed and serves the same purpose.

ahmdshrif:
@kitten  Good call — done in 67bd5b4. `handleRequestAsync` now just does:

```ts
const response = await this._getManifestResponseAsync(options);
await respond(res as http.ServerResponse, response);
```

That drops the hand-rolled header assignment and pipeline entirely, and picks up two things the old block didn't do: the response status/statusText, and multi-value header handling.

One thing to flag, since it changes what the PR covers: `respond`'s guard runs before the write, so it fixes the reported case, but it doesn't cover a client hanging up *part-way through* the response — `pipeline` rejects there and the error still reaches `getHandler`. Swapping to `respond` alone makes this test fail:

```
✕ quietly skips the response when the client disconnects while the manifest is streaming
```

So I've kept the client-disconnect `catch` wrapped around the `respond` call. It's now the only thing left in this file besides the call itself, and it re-throws anything that isn't a disconnect code, so a manifest body that fails while the client is connected still logs and still 500s (there's a test for that).

Happy to drop that `catch` and let the mid-stream case go if you'd rather keep this to the reported bug and handle the streaming half inside `respond` instead — that seems like the more natural home for it, and I'd be glad to open a separate PR against `expo-server` for it if you want it there.

Middleware suite is 18 suites / 178 tests green. Also merged `main` in on top of your two merges.

kitten:
I'm closing this in favour of #48699 and #48700

Sorry, but the issue instead pointed out that the use of `pipeline` itself is problematic and that the fix here of handling the error itself is "too late" or rather inappropriate to the issue of `pipeline` treating cancellation cases as exceptions.

I haven't commented on this usually before, but please be careful with addressing issues spontaneously without participating in that issue thread with a PR unprompted. Without proper verification, this can cause more work for maintainers, rather than save time, if the fix isn't trivial

PR #48661, opened 2026-08-07
[fingerprint] Strip autolinking config paths in the balanced and relaxed presets

Kudo:
# Why

follow-up to https://github.com/expo/expo/pull/48629#pullrequestreview-4881795310

# How

- normalize the autolinking config
- add `SourceSkips.AutolinkingConfigPaths` (which will strip paths from autolinking config) and enable it on balanced/relaxed preset

# Test Plan

update tests and ci passed

# Checklist


packages/@expo/fingerprint/src/sourcer/AutolinkingConfig.ts
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  if (value && typeof value === 'object') {
+    if (isNamedDependencyMap(value)) {
+      return normalizeNode(namedMapToSortedArray(value), stripPaths, roots, 'modules');
+    }
+
+    const result: Record<string, unknown> = {};
+    for (const [key, child] of Object.entries(value)) {
+      if (stripPaths && typeof child === 'string' && isPathLike(child, roots)) {
```
Kudo: Intended. Path-like fields are stripped at any depth, including `scriptPhases[].path`. The phase `name` still hashes.

Updated the `SourceSkips.AutolinkingConfigPaths` TSDoc and the CHANGELOG to say that.

id:51fc1f15e334

packages/@expo/fingerprint/src/sourcer/AutolinkingConfig.ts
```diff
@@ -0,0 +1,150 @@
...
+ *
+ * When `stripPaths` is true, filesystem path fields are omitted. Linked names, `scriptPhases`,
+ * and similar non-path overrides remain.
+ */
+export function normalizeAutolinkingConfig(
+  config: unknown,
+  options: { stripPaths: boolean; roots?: string[] }
+): unknown {
+  const roots = (options.roots ?? []).map((root) => toPosixPath(root));
+  return normalizeNode(structuredClone(config), options.stripPaths, roots);
```
Kudo: Intended. Sorting and map-to-array run for every preset so autolinking config order cannot churn.

Updated the CHANGELOG breaking entry and the `strict` preset docstring to say fingerprints change once on upgrade, including `strict`.

id:cb1a910af126

packages/@expo/fingerprint/e2e/__tests__/managed-test.ts
```diff
@@ -212,66 +212,27 @@ describe('managed project test', () => {
...
-              "rncoreAutolinkingIos",
-            ],
-            "type": "package",
-            "version": "*",
-          },
-          "op": "added",
-        },
-      ]
-    `);
+    const normalized = normalizeAutolinkingVersionsForSnapshot(diff);
```
Kudo: Restored the inline snapshot. The netinfo e2e now pins the exact diff: changed `rncoreAutolinkingConfig` android/ios contents (stripped, sorted arrays) plus the added package source.

id:6c3811365cee

packages/@expo/fingerprint/src/Presets.ts
```diff
@@ -43,7 +45,8 @@ export function resolvePreset(preset: FingerprintPreset): ResolvedPreset {
           SourceSkips.ExpoConfigVersions |
           SourceSkips.ExpoConfigRuntimeVersionIfString |
           SourceSkips.EasJson |
-          SourceSkips.Easignore,
+          SourceSkips.Easignore |
+          SourceSkips.AutolinkingConfigPaths,
```
Kudo: Intended. Path overrides from the project's `react-native.config.js` (`sourceDir`, `podspecPath`, and similar) are stripped with the other path fields. We are not hashing that file separately.

Added the same class of caveat as `Easignore`: those overrides will not change the fingerprint. Linked module names still hash.

id:03eeb60b356c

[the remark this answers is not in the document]
Kudo: Intended. Path overrides from the project's `react-native.config.js` (`sourceDir`, `podspecPath`, and similar) are stripped with the other path fields. We are not hashing that file separately.

Added the same class of caveat as `Easignore`: those overrides will not change the fingerprint. Linked module names still hash.

id:03eeb60b356c

packages/@expo/fingerprint/src/sourcer/AutolinkingConfig.ts
```diff
@@ -0,0 +1,151 @@
...
+  }
+  for (const root of roots) {
+    if (!root) {
+      continue;
+    }
+    if (posixValue === root || posixValue.startsWith(`${root}/`)) {
+      return true;
+    }
+  }
+  return RELATIVE_OR_NODE_MODULES_PATH_REGEX.test(posixValue);
```
kitten, line 150: I think I couldn't say whether doing this generically could lead to invalid results, as in, missed paths. I see the benefit of doing this for other properties. Although, I guess specialising for some could be beneficial, as it's hard for me to tell if `roots` would be comprehensively covering all possible starting values. But in theory, the remaining paths would be in-monorepo paths, so should be fine.

If we do see any outliers, we can probably revisit this quite easily though, so all good

kitten (approved):

PR #48663, opened 2026-08-07
[image] Add `svgVariables` prop for multi-color SVG tinting

tsapeta:
## Why

`tintColor` is a single-color alpha mask, so it can't give different parts of an SVG different colors, and on iOS it forces the SVG coder to rasterize the document.

## How

Substitutes values for the CSS custom properties an SVG refers to with `var(--name, fallback)`, on the source text, before it reaches the parser. The document stays a vector, and any `var()` can be overridden, not just colors. Gradients work by pointing a variable at a paint server the document defines, as `url(#id)`.

Neither renderer can resolve `var()` itself (AndroidSVG has CSS selectors but no custom properties, CoreSVG has neither), hence the source rewrite. For the same reason an SVG that uses `var()` now renders with its fallbacks even without the prop, so a document authored for the browser looks the same here. Separate Swift and Kotlin ports share one set of test expectations.

Only the original document is cached, under its normal key, so variants share one download and a substituted image is never served for another set of variables. On Android the variables travel as a memory-only Glide option. On iOS the view loads through `SDWebImageManager` as usual and re-decodes the delivered document with the variables substituted before showing it. `tintColor` applies on top of the substituted document on both platforms.

Values are escaped so they can't add markup or CSS declarations, and nested fallbacks stop at a fixed depth.

On Android, SVGs can now also be decoded from an in-memory buffer. Glide falls back to that when it can't read back a cache entry it just wrote, which happens when the same document loads concurrently with different variables, and only raster formats survived it before.

## Example

```svg
<!-- house.svg -->
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 100">
  <defs>
    <linearGradient id="sky" x1="0" y1="0" x2="0" y2="1">
      <stop offset="0" stop-color="var(--sky-from, #9ad0ff)" />
      <stop offset="1" stop-color="var(--sky-to, #ffffff)" />
    </linearGradient>
  </defs>
  <rect width="100" height="100" fill="url(#sky)" />
  <rect x="20" y="45" width="60" height="45" fill="var(--wall, #cccccc)" />
  <polygon points="10,50 50,15 90,50" fill="var(--roof, #888888)" />
  <rect x="43" y="62" width="14" height="28" fill="var(--door, #6b4f2a)" stroke-width="var(--frame-width, 1)" />
</svg>
```

```tsx
// Renders with the fallbacks, like a browser would.
<Image source={require('./house.svg')} />

// Substitutes values. Anything a custom property stands in for works, not just colors.
<Image
  source={require('./house.svg')}
  svgVariables={{ '--roof': '#ee3333', '--wall': '#3399ff', '--frame-width': 3 }}
/>

// A variable can select a paint server the document defines.
<Image source={require('./house.svg')} svgVariables={{ '--wall': 'url(#sky)', '--sky-to': '#1b3a66' }} />
```

## Test Plan

Unit tests on both platforms.

A new "SVG variables" screen in native-component-list: local house and gauge assets cover colors, non-color values, gradients, fallbacks, tint, blur, cache policies and load events, and a few public SVGs from GitHub that use `var()` are loaded over the network. The Expo logo asset in NCL now uses variables too, and the other screens render it without the prop.

Verified on the iOS simulator (see the demo below) and on a Galaxy A51 running Android 13.

## Demo

Screenshot 2026-09-04 at 10 53 48
<br/>
Screenshot 2026-09-04 at 10 57 03
<br/>
Screenshot 2026-09-04 at 10 57 28


tsapeta:
@expo-bot verify 🙏

packages/expo-image/ios/ImageView.swift
```diff
@@ -258,41 +283,105 @@ public final class ImageView: ExpoView {
...
-      renderSourceImage(image, cacheType: ImageCacheType.fromSdCacheType(cacheType))
-    } else {
+    guard let image else {
       displayPlaceholderIfNecessary()
+      return
+    }
+
+    // The manager delivered the original document, which is all it ever caches. With variables set the
+    // image it decoded is discarded, not shown, and rebuilt from the substituted document.
+    if let variables = svgVariables, let data, SDImageSVGCoder.shared.canDecode(from: data) {
```
tsapeta: [🏗️ **design**] When `svgVariables` is set but `data` is nil, this falls through to `didLoad(image)` and shows the manager's fallback-rendered document as if the variables had been applied. Two reachable cases:

1. `cachePolicy="memory"`. `SDImageCache.queryCacheOperationForKey` returns `doneBlock(image, nil, SDImageCacheTypeMemory)` whenever `queryCacheType == SDImageCacheTypeMemory`, before `SDImageCacheQueryMemoryData` is even considered (SDImageCache.m, `shouldQueryMemoryOnly`). So `.queryMemoryData` doesn't help here, and every memory hit ignores the variables. The "Same source, different variables" row in NCL should show this on a second visit to the screen: the `memory` cell renders the document's default red.
2. `tintColor` + `blurRadius` + `svgVariables`. The transformer path calls the completion with `cacheData: nil` for transformed images (SDWebImageManager.m, `callTransformProcessForOperation`), so `data` is nil.

For (1), one option is to treat a memory hit without data as a miss when variables are set: retry the load with `.fromLoaderOnly` (or `.refreshCached`), or keep the original `Data` on the view alongside the memory-cached image. For (2), re-applying the transform pipeline to the substituted image in `decodeSVG` would cover it. Either way, the silent path is the part worth closing: if an SVG image arrives with variables set and no data, emitting `onError` (or at least logging) beats rendering the wrong colors.

packages/expo-image/ios/Utils/SVGVariables.swift
```diff
@@ -0,0 +1,480 @@
...
+  }
+
+  // MARK: - Character helpers
+
+  private static func matches(_ chars: [Character], at index: Int, _ needle: String) -> Bool {
+    let needleChars = Array(needle)
+    guard index + needleChars.count <= chars.count else {
+      return false
+    }
+    return Array(chars[index..<(index + needleChars.count)]) == needleChars
```
tsapeta: [💡 suggestion] `matches` allocates two arrays on every call (`Array(needle)` and the slice copy), and it is called once per position from `firstIndex(of:)` while scanning comments, CDATA and declarations, and four times per `<`. `firstIndex(ofCaseInsensitive:)` similarly allocates a `String` per compared character via `$0.lowercased()`. On the normal load path this now runs inside the coder for every SVG that contains `var(`, so a large document with long comments pays for it on each decode.

Precomputing the needle once per search and comparing with `chars[index..<end].elementsEqual(needleChars)` removes the allocations without changing the structure. The Kotlin port already compares in place.

packages/expo-image/src/Image.types.ts
```diff
@@ -214,6 +214,46 @@ export interface ImageProps extends Omit<ViewProps, 'style' | 'children'> {
...
 
+  /**
+   * Values for the [CSS custom properties](https://developer.mozilla.org/en-US/docs/Web/CSS/--*)
+   * that an SVG source refers to with `var()`. Each entry is substituted into the document before it
+   * is parsed, so the image is still rendered as a vector — unlike `tintColor`, which rasterizes it.
+   *
+   * Because the substitution happens on the document itself, different parts of one SVG can be given
+   * different values, which makes multi-color template images possible.
+   *
+   * Values are not limited to colors. Anything a custom property stands in for, such as
```
tsapeta: [💡 suggestion] Supplied values are inserted verbatim and not scanned for `var()` themselves, so `{ '--roof': 'var(--wall)' }` lands in the document as a literal `var(--wall)` that neither renderer can resolve, and the attribute ends up invalid. A browser would resolve it. The gradient example in NCL (`'--color': 'url(#gradient)'`) makes indirection look supported, so this is a natural thing to try.

Either run supplied values through the same substitution (the depth limit already bounds it), or state here that values must be literal and cannot refer to other custom properties.

packages/expo-image/android/src/main/java/expo/modules/image/svg/SVGVariables.kt
```diff
@@ -0,0 +1,484 @@
...
+object SVGVariables {
+  /**
+   * How many levels of `var()` fallbacks are resolved before the rest is copied through verbatim.
+   * Fallbacks are resolved recursively, so without a limit a document could nest them deeply enough
+   * to overflow the stack. No real document comes anywhere near this depth.
+   */
+  internal const val MAX_FALLBACK_DEPTH = 32
+
+  /**
+   * Substitutes the variables in an SVG document. An empty map still resolves every `var()` to its
```
tsapeta: [🐧 nit] This KDoc block is orphaned: it sits directly above the `resolveFallbacks` KDoc, so it documents nothing, and `substitute` below has no doc comment. Looks like it belongs on `substitute`.

packages/expo-image/android/src/test/java/expo/modules/image/svg/ByteBufferSVGDecoderTest.kt
```diff
@@ -0,0 +1,29 @@
...
+    val resource = ByteBufferSVGDecoder().decode(ByteBuffer.wrap(document.toByteArray()), 0, 0, Options())
+    assertNotNull(resource)
+    assertEquals(20f / 10f, resource!!.get().documentViewBox.height() / resource.get().documentViewBox.width())
+  }
+
+  @Test
+  fun `passes the request options through to the substitution`() {
+    // The variables travel on the request options, which must reach the shared parsing code.
+    val options = Options().set(CustomOptions.svgVariables, mapOf("--a" to "blue"))
+    assertNotNull(ByteBufferSVGDecoder().decode(ByteBuffer.wrap(document.toByteArray()), 0, 0, options))
```
tsapeta: [🐧 nit] This passes even if the decoder drops the options entirely, since the document parses with or without substitution. To actually check that the variables reached `SVGDecoder.parse`, use a document that only parses (or only gets a viewBox) after substitution, for example `viewBox="var(--box)"` with `--box` supplied, and assert on the resulting `documentViewBox`.

tsapeta:
@expo-bot verify and compare iOS and Android screenshots that you produce

packages/expo-image/src/Image.types.ts
```diff
@@ -214,6 +214,48 @@ export interface ImageProps extends Omit<ViewProps, 'style' | 'children'> {
...
+   * `stroke-width` or `opacity`, is substituted the same way. Keys include the leading `--`.
+   * Values are inserted as written and cannot refer to other custom properties: a value such as
+   * `'var(--other)'` is not resolved. Reference a paint server the document defines with `url(#id)`.
+   *
+   * A property that is not supplied falls back to the value declared inside its own `var()`, and
+   * when there is no fallback the declaration is dropped so the renderer applies its own default.
```
amandeepmittal, line 231: ✏️ **style** `mechanics.md#vale-rules-enforced-if-missed SentenceLength`

This sentence runs 33 words. End it after `var()` and give the no-fallback case its own sentence.

---

Two ideas joined by "and": the fallback rule and the no-fallback rule each want their own sentence.

```suggestion
   * A property that is not supplied falls back to the value declared inside its own `var()`. When
   * there is no fallback, the declaration is dropped and the renderer applies its own default.
```

packages/expo-image/src/Image.types.ts
```diff
@@ -214,6 +214,48 @@ export interface ImageProps extends Omit<ViewProps, 'style' | 'children'> {
...
+   * An SVG that uses `var()` is rendered with its fallbacks even when this prop is not set, so a
+   * document authored for the browser looks the same here.
```
amandeepmittal, line 233: ✏️ **style** `mechanics.md#vale-rules-enforced-if-missed SentenceLength`

This sentence runs 28 words. End it after "not set" and give the browser comparison its own sentence.

---

💡 **nit** `prose.md#core-principles`

"Here" has no referent once this text is on the docs page; "in the app" says where.

```suggestion
   * An SVG that uses `var()` is rendered with its fallbacks even when this prop is not set. A
   * document authored for the browser looks the same in the app.
```

packages/expo-image/src/Image.types.ts
```diff
@@ -214,6 +214,48 @@ export interface ImageProps extends Omit<ViewProps, 'style' | 'children'> {
    */
   tintColor?: string | null;
 
+  /**
+   * Values for the [CSS custom properties](https://developer.mozilla.org/en-US/docs/Web/CSS/--*)
+   * that an SVG source refers to with `var()`. Each entry is substituted into the document before it
+   * is parsed, so the image is still rendered as a vector — unlike `tintColor`, which rasterizes it.
```
amandeepmittal, line 220: ⚠️ **correctness** `review-ai-text.md#t7-overclaiming`

On Android `tintColor` is applied to the SVG data and the result is still rendered as a `Picture`, so "which rasterizes it" holds only on iOS, as your second note already says. Scope this sentence the same way.

---

Prefer a plain clause or a new sentence to the em dash before "unlike".

```suggestion
   * that an SVG source refers to with `var()`. Each entry is substituted into the document before it
   * is parsed, so the image stays a vector, unlike an SVG tinted with `tintColor` on iOS.
```

packages/expo-image/src/Image.types.ts
```diff
@@ -214,6 +214,48 @@ export interface ImageProps extends Omit<ViewProps, 'style' | 'children'> {
...
+   *
+   * Because the substitution happens on the document itself, different parts of one SVG can be given
+   * different values, which makes multi-color template images possible.
```
amandeepmittal, line 223: 💡 **nit** `prose.md#core-principles`

"Template images" is defined under `tintColor` as an opacity-only bitmap, which this SVG is not, so say what happens instead of reusing the term.

```suggestion
   * different values, so a single document can be tinted with several colors.
```

packages/expo-image/src/Image.types.ts
```diff
@@ -214,6 +214,48 @@ export interface ImageProps extends Omit<ViewProps, 'style' | 'children'> {
...
+   *
+   * Values are not limited to colors. Anything a custom property stands in for, such as
+   * `stroke-width` or `opacity`, is substituted the same way. Keys include the leading `--`.
+   * Values are inserted as written and cannot refer to other custom properties: a value such as
+   * `'var(--other)'` is not resolved. Reference a paint server the document defines with `url(#id)`.
```
amandeepmittal: ✏️ **style** `prose.md#voice-and-tone`

Reference entries describe rather than instruct, and "paint server" is spec jargon. Say that a `url(#id)` value uses a gradient or pattern the document defines.

```suggestion
   * `'var(--other)'` is not resolved. A value of `url(#id)` uses a gradient or pattern that the
   * document defines.
```

packages/expo-image/src/Image.types.ts
```diff
@@ -214,6 +214,48 @@ export interface ImageProps extends Omit<ViewProps, 'style' | 'children'> {
...
+   * > **Note:** `tintColor` is applied on top of the substituted document. It floods every pixel with
+   * > a single color, so color values are hidden by it while other values, such as widths or
+   * > opacities, still take effect. On iOS the image is rasterized in that case, as with `tintColor`
+   * > alone.
+   *
+   * Has no effect on sources that aren't SVG.
+   *
+   * @example
+   * ```tsx
+   * // house.svg contains fill="var(--roof, #888)" and fill="var(--wall, #ccc)"
```
amandeepmittal: ✏️ **style** `prose.md#core-principles`

The example passes `--stroke-width`, but the comment says the document only declares `--roof` and `--wall`, so the reader cannot tell where the number goes. Name the `stroke-width` declaration in the comment too.

```suggestion
   * // house.svg contains fill="var(--roof, #888)", fill="var(--wall, #ccc)"
   * // and stroke-width="var(--stroke-width, 1)"
```

packages/expo-image/src/Image.types.ts
```diff
@@ -214,6 +214,48 @@ export interface ImageProps extends Omit<ViewProps, 'style' | 'children'> {
...
+   * An SVG that uses `var()` is rendered with its fallbacks even when this prop is not set, so a
+   * document authored for the browser looks the same here.
+   *
+   * > **Note:** Colors must be values the SVG itself understands, such as `'#ff0000'` or `'red'`.
+   * > React Native color descriptors like `PlatformColor` are not resolved.
+   *
+   * > **Note:** `tintColor` is applied on top of the substituted document. It floods every pixel with
+   * > a single color, so color values are hidden by it while other values, such as widths or
+   * > opacities, still take effect. On iOS the image is rasterized in that case, as with `tintColor`
+   * > alone.
```
amandeepmittal: ⚠️ **correctness** `review-ai-text.md#t7-overclaiming`

A tint recolors only the non-transparent pixels, which is how the `tintColor` entry above puts it; "floods every pixel" describes a solid rectangle. Say "recolors every non-transparent pixel".

```suggestion
   * > **Note:** `tintColor` is applied on top of the substituted document. It recolors every
   * > non-transparent pixel, so color values are hidden by it while other values, such as widths or
   * > opacities, still take effect. On iOS the image is rasterized in that case, as with `tintColor`
   * > alone.
```

packages/expo-image/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
...
 - Added an `includeUrlParams` option to the expo-observe integration; reported image URLs now have their query string and fragment removed unless it is enabled, basic-auth credentials are always removed, and only `http(s)`, `file`, and `android.resource` URLs are reported. ([#49083](https://github.com/expo/expo/pull/49083) by [@tsapeta](https://github.com/tsapeta))
+- [iOS][Android] Added an `svgVariables` prop that substitutes values for the CSS custom properties an SVG source refers to with `var()`, so different parts of one document can be given different colors and the image stays a vector instead of being rasterized the way `tintColor` is. Values are not limited to colors. SVG sources that use `var()` now render with their fallback values even without the prop. ([#48663](https://github.com/expo/expo/pull/48663) by [@tsapeta](https://github.com/tsapeta))
```
amandeepmittal: ⚠️ **correctness** `review-ai-text.md#t7-overclaiming`

`tintColor` rasterizes an SVG only on iOS; on Android the tinted document is still rendered as a `Picture`. Scope the comparison to iOS.

```suggestion
- [iOS][Android] Added an `svgVariables` prop that substitutes values for the CSS custom properties an SVG source refers to with `var()`, so different parts of one document can be given different colors and the image stays a vector, unlike an SVG tinted with `tintColor` on iOS. Values are not limited to colors. SVG sources that use `var()` now render with their fallback values even without the prop. ([#48663](https://github.com/expo/expo/pull/48663) by [@tsapeta](https://github.com/tsapeta))
```

amandeepmittal (approved):

packages/expo-image/android/src/main/java/expo/modules/image/svg/SVGDecoder.kt
```diff
@@ -53,4 +58,35 @@ class SVGDecoder : ResourceDecoder<InputStream, SVG> {
...
+   * resolving every `var()` to its fallback when it asked for none. The substitution happens on the
+   * source text because AndroidSVG cannot resolve `var()` itself.
+   */
+  private fun parse(source: InputStream, options: Options): SVG {
+    val variables = options.get(CustomOptions.svgVariables)
+    val bytes = source.readBytes()
+    val text = decodeUtf8(bytes)
+      // Substituting would mean re-encoding the document as UTF-8, which would contradict its own
+      // XML declaration. Leave it alone rather than corrupt it.
+      ?: return SVG.getFromInputStream(ByteArrayInputStream(bytes))
```
lukmccall, line 73: I'm okay with returning it like that, but I think it would be nice to print at least a warning. Maybe in the catch of decodeUtf8?

lukmccall (approved):

packages/expo-image/android/src/main/java/expo/modules/image/svg/SVGDecoder.kt
```diff
@@ -53,4 +58,35 @@ class SVGDecoder : ResourceDecoder<InputStream, SVG> {
...
+
+  /**
+   * Parses the document, first substituting the CSS custom properties the request asked for, or
+   * resolving every `var()` to its fallback when it asked for none. The substitution happens on the
+   * source text because AndroidSVG cannot resolve `var()` itself.
+   */
+  private fun parse(source: InputStream, options: Options): SVG {
+    val variables = options.get(CustomOptions.svgVariables)
+    val bytes = source.readBytes()
+    val text = decodeUtf8(bytes)
```
tsapeta: [🏗️ **design**] A strict UTF-8 decode does not reject BOM-less UTF-16. `<svg/>` as UTF-16LE is `3C 00 73 00 76 00 67 00 2F 00 3E 00`, and every one of those bytes is a valid single-byte UTF-8 sequence (NUL decodes to U+0000), so `decodeUtf8` returns `"<\u0000s\u0000v\u0000g\u0000/\u0000>\u0000"` instead of null. The `?:` guard never fires, and `SVG.getFromString` is handed that string.

This regresses documents that contain no `var()` at all: `main` passed the raw stream to `SVG.getFromInputStream`, which lets the XML parser sniff the encoding or honour the declaration. Note iOS does not have the same outcome — `SVGVariables.substitute(in:variables:)` re-encodes the same code points back to UTF-8, so the bytes round-trip unchanged when nothing matches — which makes this a platform divergence as well as a regression.

The cheapest fix is to sniff for UTF-16 before decoding (a BOM, or a NUL in the first few bytes) and take the `getFromInputStream` branch. Worth a `SVGDecoderTest` covering it too — there is currently no test for this fallback path.

packages/expo-image/ios/Utils/SVGVariables.swift
```diff
@@ -0,0 +1,497 @@
...
+  /// Escapes a caller-supplied value so that it cannot terminate the attribute it sits in or add
+  /// markup of its own. Returns `nil` for a value that can't be made safe, which the caller treats
+  /// as unresolved.
+  ///
+  /// Only values that came from the variable map go through this. Fallbacks are document text that
+  /// is already escaped, so escaping them again would double up their entities.
+  private static func escape(_ value: String, for context: Context) -> String? {
+    // Inside CSS these could close the declaration or the rule and start another one. No
+    // legitimate CSS value needs them.
+    if context == .styleBody, value.contains(where: { $0 == "{" || $0 == "}" || $0 == ";" }) {
```
tsapeta: [💡 suggestion] Entity escaping is wrong when the `<style>` body is wrapped in CDATA, which is how Illustrator and Inkscape export stylesheets.

`substitute` reaches `<style>` and hands the whole body to `copyStyleBody`, which never re-enters the passthrough logic — so `<![CDATA[` and `]]>` are just characters inside the value and substitution does happen there (good). But inside CDATA the XML parser does not decode entities, so a value of `A & B` is written as the literal CSS text `A &amp; B` rather than `A & B`:

```
<style><![CDATA[.a{fill:var(--c)}]]></style>  +  {"--c": "A & B"}
  → <style><![CDATA[.a{fill:A &amp; B}]]></style>
```

Outside CDATA the current behaviour is right, so this needs a third context rather than dropping the escaping. Detecting whether the offset falls inside a CDATA section in `copyStyleBody` and passing a `.cdataStyleBody` context that skips entity escaping (keeping the `{`/`}`/`;` rejection) would cover it. Narrow, since it only bites values carrying XML metacharacters, but the same fix belongs in `SVGVariables.kt:268`.

packages/expo-image/ios/Utils/SVGVariables.swift
```diff
@@ -0,0 +1,497 @@
...
+    context: Context,
+    depth: Int = 0
+  ) -> SubstitutedValue {
+    var out = ""
+    var index = 0
+    var unresolvedSpan: Range<Int>?
+    var substitutions = 0
+
+    while index < value.count {
+      guard matches(value, at: index, "var("), let close = matchingParen(value, openParen: index + 3) else {
```
tsapeta: [💡 suggestion] CSS function names are case-insensitive, so a browser resolves `VAR(--a, #000)` and `Var(--a, #000)` the same as `var(`. This matches only the lowercase spelling, so an uppercase reference is copied through verbatim and the renderer sees an unparseable value.

The file already takes case-insensitivity seriously elsewhere — `firstIndex(ofCaseInsensitive:)` exists precisely so `</STYLE>` is found, and there is a test for an uppercase closing tag. Worth being consistent here.

Uppercase `var(` is rare in hand-authored SVG, so this is low priority, but it is cheap: a case-insensitive variant of `matches` used at this one site. Same line in `SVGVariables.kt:315`, and it would want a shared test case in both suites.

packages/expo-image/src/Image.types.ts
```diff
@@ -214,6 +214,48 @@ export interface ImageProps extends Omit<ViewProps, 'style' | 'children'> {
...
+   * Values for the [CSS custom properties](https://developer.mozilla.org/en-US/docs/Web/CSS/--*)
+   * that an SVG source refers to with `var()`. Each entry is substituted into the document before it
+   * is parsed, so the image stays a vector, unlike an SVG tinted with `tintColor` on iOS.
+   *
+   * Because the substitution happens on the document itself, different parts of one SVG can be given
+   * different values, so a single document can be tinted with several colors.
+   *
+   * Values are not limited to colors. Anything a custom property stands in for, such as
+   * `stroke-width` or `opacity`, is substituted the same way. Keys include the leading `--`.
+   * Values are inserted as written and cannot refer to other custom properties: a value such as
```
tsapeta: [💡 suggestion] The docs say a value like `'var(--other)'` "is not resolved", but they do not say what happens instead — and the outcome is worse than a no-op. The value is inserted as written and `(`/`)` are not escaped, so an unresolved `var()` is left in the finished document:

```
<rect fill="var(--a)"/>  +  {"--a": "var(--b, blue)"}
  → <rect fill="var(--b, blue)"/>
```

Neither renderer can parse that, so the attribute is silently broken rather than falling back. Two options: either drop such a value in `escape` (treat it as unresolved, which routes it through the existing attribute-dropping path and lets the renderer apply its own default), or keep the current behaviour and say plainly in the doc that the attribute will fail to render.

Related and worth one line in the same block: a key without the leading `--` (`{ a: 'red' }` rather than `{ '--a': 'red' }`) is silently ignored and the fallback wins. That is an easy mistake to make and there is no signal when it happens.

packages/expo-image/ios/Coders/SVGCoder.swift
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  func decodedImage(with data: Data?, options: [SDImageCoderOption: Any]? = nil) -> UIImage? {
+    guard let data, let image = coder.decodedImage(with: SVGVariables.resolveFallbacks(in: data), options: options) else {
+      return nil
+    }
+    if SVGVariables.usesVariables(data) {
+      // The extended object has to be an `NSCoding` object.
+      // swiftlint:disable:next legacy_objc_type
+      image.sd_extendedObject = data as NSData
```
tsapeta: [💡 suggestion] This attaches the full original document to every image decoded from a `var()`-using source, whether or not any view will ever set `svgVariables`. Those images sit in SDWebImage's memory cache, so each cached SVG that uses `var()` now holds a second copy of its own source bytes for as long as it is cached. `sd_extendedObject` is also what SDWebImage serialises alongside images it writes to its disk cache, so the document gets persisted there too.

For icon-sized SVGs this is small, and the coder genuinely has no way to know whether a later view will want the document — the memory-hit and transformed-image paths are exactly why it is here. So this may well be the right trade. But it is worth a sentence in the doc comment saying the cost is paid for every `var()` document, not just the ones with the prop set, so the next reader does not have to work that out.

One cheap narrowing if it matters: `usesVariables` is a plain `var(` substring check over the raw bytes, so it also fires on documents where the only `var(` is inside `<text>` content or a comment. Those pay the cost and can never benefit.

packages/expo-image/ios/ImageView.swift
```diff
@@ -258,39 +277,105 @@ public final class ImageView: ExpoView {
...
+  /// of variables or for a load without any.
+  private func substituteSVGVariables(
+    _ variables: [String: String],
+    in data: Data,
+    replacing original: UIImage,
+    cacheType: SDImageCacheType,
+    imageUrl: URL?
+  ) {
+    let scale = original.scale
+    let thumbnailPixelSize = thumbnailPixelSize(for: bestSource)
```
tsapeta: [🐧 nit] This reads `bestSource` and `sdImageView.bounds` at completion time, whereas the load that produced this image computed its own thumbnail size from the source and bounds at `reload()` time (line 194). If the view was laid out or the source changed between the two, the substituted image is rasterised at a different size than the original was.

In practice `reload()` cancels the pending task on every prop update, so the window is narrow. Passing the size computed in `reload()` through instead of recomputing would remove the question entirely — but if you would rather it track the current bounds, a one-line note saying so would make the choice explicit.

PR #48671, opened 2026-08-08
[ui][ios] Add missing matchedGeometryEffect parameters

Den1Marshall:
# Why

`matchedGeometryEffect` only exposes `id` and `namespaceId`, while SwiftUI's `matchedGeometryEffect(id:in:properties:anchor:isSource:)` takes three more. Without them the modifier can only copy the whole frame, centered, from a view that is always the source.

# How

Added `properties`, `anchor` and `isSource` as an optional options object, all defaulting to the SwiftUI values, so existing calls are unaffected. `anchor` reuses the existing `UnitPointOptions`. `properties` is a plain enum rather than an option set — `position`, `size` and `frame` cover every combination that does anything. Docs API data regenerated.

# Test Plan

Tested in an Expo SDK 57 app on iOS 26 — existing calls behave as before, each new parameter takes effect.

# Checklist


intergalacticspacehighway (approved):
Thanks!

PR #48678, opened 2026-08-09
[core] Defer shared object release until later cleanups have run

### LINKED ISSUE #30994: [expo-video] Shared object cannot be found when navigation blur listener is used with Expo Router

https://github.com/fobos531/expo-video-expo-router-repro

iOS

in a development build

### Summary

When you manipulate a player within the `blur` listener provided by React Navigation, it will work fine when you navigate away from the screen, but will crash with an error when you navigate back to this screen, claiming that the native shared object associated with the JS object cannot be found. See screenshot:

CleanShot 2024-08-14 at 11 17 28@2x

I tried reproducing this in bare-expo but I wasn't able to reproduce. This leads me to believe the issue is related to how `expo-router` and `react-native-screens` handle inactive screens or something similar.

### Environment

```
  expo-env-info 1.2.0 environment info:
    System:
      OS: macOS 15.1
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 20.10.0 - ~/.nvm/versions/node/v20.10.0/bin/node
      Yarn: 1.22.21 - ~/.nvm/versions/node/v20.10.0/bin/yarn
      npm: 10.2.3 - ~/.nvm/versions/node/v20.10.0/bin/npm
      Watchman: 2024.07.15.00 - /opt/homebrew/bin/watchman
    Managers:
      CocoaPods: 1.15.2 - /opt/homebrew/lib/ruby/gems/3.3.0/bin/pod
    SDKs:
      iOS SDK:
        Platforms: DriverKit 24.0, iOS 18.0, macOS 15.0, tvOS 18.0, visionOS 2.0, watchOS 11.0
      Android SDK:
        API Levels: 26, 29, 30, 31, 32, 33, 33, 34
        Build Tools: 26.0.3, 29.0.2, 30.0.3, 31.0.0, 33.0.0, 33.0.1, 33.0.2, 34.0.0
        System Images: android-26 | Google APIs Intel x86 Atom_64, android-28 | Google ARM64-V8a Play ARM 64 v8a, android-33 | Google Play ARM 64 v8a, android-34 | Google Play ARM 64 v8a
    IDEs:
      Android Studio: 2024.1 AI-241.18034.62.2411.12071903
      Xcode: 16.0/16A5221g - /usr/bin/xcodebuild
    npmPackages:
      expo: ~51.0.26 => 51.0.27 
      expo-router: ~3.5.21 => 3.5.21 
      react: 18.2.0 => 18.2.0 
      react-dom: 18.2.0 => 18.2.0 
      react-native: 0.74.5 => 0.74.5 
      react-native-web: ~0.19.10 => 0.19.12 
    npmGlobalPackages:
      eas-cli: 10.2.2
    Expo Workflow: managed

```

### Expo Doctor Diagnostics

```
✔ Check Expo config for common issues
✔ Check package.json for common issues
✔ Check dependencies for packages that should not be installed directly
✔ Check for issues with metro config
✖ Check for common project setup issues
✔ Check npm/ yarn versions
✔ Check Expo config (app.json/ app.config.js) schema
✔ Check native tooling versions
✔ Check that native modules do not use incompatible support packages
✔ Check for legacy global CLI installed locally
✔ Check that native modules use compatible support package versions for installed Expo SDK
✔ Check that packages match versions required by installed Expo SDK

Detailed check results:

This project has native project folders but also has config plugins, indicating it is configured to use Prebuild. EAS Build will not sync your native configuration if the ios or android folders are present. Add these folders to your .gitignore file if you intend to use prebuild (aka "managed" workflow).

One or more checks failed, indicating possible issues with the project.
```
➜ npx expo-doctor@latest
✔ Check Expo config for common issues
✔ Check package.json for common issues
✔ Check native tooling versions
✔ Check dependencies for packages that should not be installed directly
✔ Check for common project setup issues
✔ Check for issues with metro config
✔ Check npm/ yarn versions
✔ Check for legacy global CLI installed locally
✔ Check Expo config (app.json/ app.config.js) schema
✔ Check that native modules do not use incompatible support packages
✔ Check that packages match versions required by installed Expo SDK
✔ Check that native modules use compatible support package versions for installed Expo SDK

Didn't find any issues with the project!
```

### PR DESCRIPTION
**Pull request by L65FREAD**
# Why

Fixes #30994.

`useReleasingSharedObject` is called before any of the consumer component's own hooks, so its unmount cleanup runs first and releases the native object while sibling cleanups are still queued to use it. Pausing a `useVideoPlayer` player from a `useFocusEffect` cleanup throws:

```
FunctionCallException: Calling the 'pause' function has failed
→ Caused by: NotFoundException: Unable to find the native shared object
   associated with given JavaScript object
```

Both cleanups run in the same commit, so this fails on every unmount rather than intermittently. The throw escapes the commit and the app renders a blank screen.

The `blur`-listener repro linked in the issue no longer fires on SDK 57 — `blur` is now emitted before the unmount commit. The `useFocusEffect` path still does.

# How

Defer the unmount release by one microtask. React runs a commit's cleanups synchronously, so the ones queued after this hook's have all run by the time the release happens.

Only the unmount path changes — the dependency-change release still runs synchronously in the effect body, and the `isFastRefresh` check stays synchronous so Fast Refresh is unaffected.

This changes release timing for every `useReleasingSharedObject` consumer, not just `expo-video`. The object only ever lives longer, and `release()` no-ops once the native state is gone, so a consumer releasing manually is unaffected. `@expo/ui` already carries a try/catch for this same ordering in `useNativeState`.

# Test Plan

- New tests in `src/hooks/__tests__/useReleasingSharedObjectWithLifecycle-test.tsx`: a cleanup registered after the hook sees a live object on unmount, release still happens exactly once, and a dependency change still releases the previous object. With the hook change reverted, the first fails with the error above.
- `pnpm build`, `typecheck`, `test`, `lint` and `format` pass for `expo-modules-core` (29 suites, 183 tests).
- Verified on an SDK 57 app (expo-modules-core 57.0.10, expo-video 57.0.2, react-native 0.86.2, New Architecture) on an iPhone 17 Pro simulator, iOS 26.5: the `useFocusEffect` cleanup throws on every pop unpatched, is clean across 3 pops patched, throws again on revert, clean again on re-apply. Logging the deferred callback confirmed exactly one release per unmount, after the consumer's cleanup.

  Patched into the installed 57.0.10 build, which predates the lifecycle refactor — equivalent microtask deferral, not this exact diff. iOS simulator and `expo-video` only.

# Checklist

- [ ] This diff will work correctly for `npx expo prebuild` & EAS Build (not applicable: JavaScript-only)
- [ ] Conforms with the [Documentation Writing Style Guide](https://github.com/expo/expo/blob/main/guides/Expo%20Documentation%20Writing%20Style%20Guide.md) (not applicable: no documentation changes)


brentvatne:
@expo-bot verify

PR #48683, opened 2026-08-09
[expo-audio][Android] Keep audio focus while playback continues; request AUDIOFOCUS_GAIN for doNotMix

sang88ho:
# Why

On Android, `expo-audio` abandons audio focus **while playback is still ongoing**, and requests `AUDIOFOCUS_GAIN_TRANSIENT` for `interruptionMode: 'doNotMix'`. Together these let another app's audio play **on top of** ours: after we drop focus the other app reclaims it and stops re-requesting, so `AUDIOFOCUS_LOSS` never reaches us and `AudioModule` never pauses the player.

Both points differ from media3's own `AudioFocusManager`, which the underlying ExoPlayer ships with:

| | media3 `AudioFocusManager` | `expo-audio` (56.0.13 and current `main`) |
|---|---|---|
| When focus is abandoned | only `playbackState == STATE_IDLE` / `release()` | **every `onIsPlayingChanged(false)`** |
| Focus request for media playback | `AUDIOFOCUS_GAIN` | **`AUDIOFOCUS_GAIN_TRANSIENT`** (`doNotMix`) |
| `AudioAttributes.usage` | `USAGE_MEDIA` | unset → `USAGE_UNKNOWN` |

The release condition is the critical one: `onIsPlayingChanged(false)` also fires for **buffering stalls and track transitions** — `BaseAudioPlayer` computes `isTransient` *after* invoking `onPlaybackStateChange` — and nothing re-acquires focus when playback resumes.

Related: lock screen / notification play is routed by media3's `MediaSession` straight to `player.play()`, bypassing the `play` function where `requestAudioFocus()` lives, so playback can also start while holding no focus.

# How

In `AudioModule.kt`, for both the `AudioPlayer` and `AudioPlaylist` paths:

- Release focus only when `playbackState == Player.STATE_IDLE`, and request it when playback starts (which also covers lock screen controls).
- Request `AUDIOFOCUS_GAIN` instead of `AUDIOFOCUS_GAIN_TRANSIENT` for `doNotMix`. `duckOthers` keeps `GAIN_TRANSIENT_MAY_DUCK`.
- Declare `USAGE_MEDIA` on the focus request's `AudioAttributes`.

# Test Plan

Verified on a physical Galaxy S24+ (SM-S926N, Android 16) with a queued podcast-style player (`shouldPlayInBackground: true`, `playsInSilentMode: true`, `interruptionMode: 'doNotMix'`) against a third-party music app, reading `adb shell dumpsys audio` (`Events log: focus commands`).

Before:

```
20:59:04.576  requestAudioFocus()   callingPack=<app>  req=2  AA=USAGE_UNKNOWN/CONTENT_TYPE_MUSIC
20:59:04.577  focus owner: <music app>  code:-2  event: handleLoss
20:59:05.375  abandonAudioFocus()   callingPack=<app>         ← 0.8 s in, still playing
20:59:05.375  focus owner: <music app>  code:1   event: handleGain
              (no further request; our playback continues with no focus)
```

`abandonAudioFocus()` fired once per track transition while playback continued, and pressing play in the music app then produced two audio streams at once.

After (same device, same steps, `expo-audio` built from source with this change):

- request logged as `req=1` (`AUDIOFOCUS_GAIN`), `AA=USAGE_MEDIA/CONTENT_TYPE_MUSIC`
- 160 s of playback across 2 track transitions → **0** `abandonAudioFocus()` calls
- starting the music app → our player pauses within 2 s, with `focus owner: <app> event: handleLoss` in the log
- pausing in-app and resuming from the lock screen keeps focus held (no duplicate request)

`:expo-audio:compileReleaseKotlin` and `:expo-audio:compileReleaseJavaWithJavac` succeed with this change (Gradle release build of an SDK 56 app, `expo-audio` linked from source).

Not covered: the recorder paths and `duckOthers`, which this change does not touch.

Note on `dumpsys audio`: on this device the `Audio Focus stack entries` section renders empty and `MultiFocusStack` is stale — the `Events log: focus commands` section is the reliable one for verifying.

# Checklist

- [x] Verified the behavior on a physical Android device.


sang88ho:
Thanks — the warning found a real gap, and I've pushed a fix.

**What was wrong.** media3's condition is

```java
// AudioFocusManager.java (1.8.0, line 220)
private boolean shouldHandleAudioFocus(@Player.State int playbackState) {
  return playbackState != Player.STATE_IDLE && focusGainToRequest == AUDIOFOCUS_GAIN;
}
```

I only ported the first half. The second half is exactly what protects the ducking case the review describes: media3 asserts it will only manage `AUDIOFOCUS_GAIN` (line 170), so anything transient is abandoned as soon as playback isn't active. Without it, `duckOthers` took `AUDIOFOCUS_GAIN_TRANSIENT_MAY_DUCK` and never gave it back — the other app stays quiet after a short sound ends. That's a regression I introduced, and the reviewer is right that no other normal path would have abandoned it.

Fixed by tracking the requested gain and porting the whole condition:

```kotlin
private fun shouldAbandonFocus(playbackState: Int): Boolean {
  return playbackState == Player.STATE_IDLE ||
    focusGainRequested != AudioManager.AUDIOFOCUS_GAIN
}
```

I also took the suggestion about the unset mode: `?: AUDIOFOCUS_GAIN_TRANSIENT` is restored, so an app that never calls `setAudioModeAsync` keeps its previous behaviour and this PR is scoped to the documented `doNotMix` mode.

**What I'd like to keep.** For `doNotMix`, focus is still held through a pause and through `STATE_ENDED`. That isn't an oversight — it's the same thing media3 does for `AUDIOFOCUS_GAIN` (`updateAudioFocus` only abandons when `shouldHandleAudioFocus` is false; a paused or ended player with permanent focus keeps it), and it's the actual bug I set out to fix. Abandoning on every non-playing state is what let two apps play at once: `onIsPlayingChanged(false)` fires on buffering stalls and on `replace()` + `play()` track transitions, so the module dropped out of the focus stack mid-playback, and the next app to request focus had nobody to notify. Measured on a Galaxy S24+ (Android 16) with `dumpsys audio`: before, `abandon` appeared between tracks and the other app's later `▶` produced no `AUDIOFOCUS_LOSS` for us; after, 160 s of playback with zero abandons, and starting Melon paused our playback within two seconds.

Two notes on the evidence section, since three of the four sub-reviewers listed "could not read media3 `AudioFocusManager` source" as an uncertainty — the citation above is that source, and it does check `STATE_IDLE`, so the comparison in the changelog holds.

On iOS: agreed that `onPlaybackComplete → deactivateSession()` is not symmetric with this. I'd rather not change iOS in this PR — `AVAudioSession` deactivation and Android focus have different semantics for "paused but still the current player", and the Android side is where the reported bug is. Happy to open a separate discussion if you want them aligned.

Still open and pre-existing, so I left it out: `focusAcquired` stays `true` when the last player is removed or released (`Function("remove")` drops it from the map, and `releasePlayer()` detaches the listener first), so a later switch to `mixWithOthers` can't rebuild the request. Glad to fix it here or in a follow-up, whichever you prefer.

Verified: `:expo-audio:compileReleaseKotlin` builds, and the same patch runs in a production app on a physical device.

sang88ho:
Follow-up: I went ahead and fixed the stale-request leak in this PR too, since it's the same field and reviewing it separately would be awkward.

```kotlin
/**
 * Mirrors media3's `AudioFocusManager.release()`: focus is only worth holding while a player
 * exists that could resume.
 */
private fun releaseFocusIfNoPlayables() {
  if (allPlayables.none()) {
    releaseAudioFocus()
  }
}
```

Called from `Function("remove")` (player) and `Function("destroy")` (playlist) after the map entry is dropped — the two places the review identified as leaving `focusAcquired == true` with nothing left to resume. Both call it off the main thread, which matches the existing `setIsAudioActiveAsync(false)` call site.

`allPlayables.none()` rather than `shouldReleaseFocus()` on purpose: the latter is "nothing is playing", and abandoning on that would undo the hold-while-paused behaviour that `AUDIOFOCUS_GAIN` is for. "No player exists at all" is the condition media3 uses in `release()`, and it's the one that makes the early return in `requestAudioFocus()` recoverable.

So the three review points are all addressed: the missing half of the media3 condition, the unset interruption mode, and this. Verified with `:expo-audio:compileReleaseKotlin`.

packages/expo-audio/android/src/main/java/expo/modules/audio/AudioModule.kt
```diff
@@ -139,6 +139,34 @@ class AudioModule : Module() {
...
+   * `playbackState != STATE_IDLE && focusGainToRequest == AUDIOFOCUS_GAIN`.
+   *
+   * Permanent focus is kept across pauses, buffering stalls and track transitions, but a
+   * transient or ducking request is abandoned as soon as playback stops, so the interrupted app
+   * un-ducks or resumes.
+   */
+  private fun shouldAbandonFocus(playbackState: Int): Boolean {
+    return playbackState == Player.STATE_IDLE ||
+      focusGainRequested != AudioManager.AUDIOFOCUS_GAIN
+  }
```
behenate, line 153: We now have `shouldAbandonFocus` and `shouldReleaseFocus` methods. It's not very intuitive

packages/expo-audio/android/src/main/java/expo/modules/audio/AudioModule.kt
```diff
@@ -160,14 +195,18 @@ class AudioModule : Module() {
     val result = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
       val requestType = interruptionMode?.let {
         if (it == InterruptionMode.DO_NOT_MIX) {
-          AudioManager.AUDIOFOCUS_GAIN_TRANSIENT
+          // Media playback holds permanent focus. GAIN_TRANSIENT tells the interrupted app we
+          // will hand focus back, so it resumes as soon as we abandon it.
+          AudioManager.AUDIOFOCUS_GAIN
```
behenate, line 200: Even though I agree that this should be the default focus for `doNotMix` this would be a breaking change so we can't accept it.

behenate (changes requested):
Hi thank you for your contribution, this PR introduces breaking changes and while I agree with the general premise, when reviewing I have decided that we need a different non-breaking approach for focus handling which I implemented in this PR stack: https://github.com/expo/expo/pull/48683. 

Because the stack covers the functionality added by this PR without introducing breaking changes, I'll close it

PR #48707, opened 2026-08-10
[router] Add global registry for routers.

Ubax:
# Why

In order to be able to attach correct and not stale state when actions are dispatched, we need to know all the mounted routers. This is a prerequisite for the follow-up nested param and hydration removal.

# How

1. Add global router registry context
2. Register every router identified by `navigatorKey` in `useNavigationBuilder`.
3. Update tests

# Test Plan

CI

# Checklist

### DIFF
--- a/packages/expo-router/src/global-state/__tests__/routerRegistry.test.ios.tsx
+++ b/packages/expo-router/src/global-state/__tests__/routerRegistry.test.ios.tsx
@@ -1,6 +1,6 @@
 import { jest } from '@jest/globals';
 import { act, render } from '@testing-library/react-native';
-import { StrictMode, use, type ReactNode } from 'react';
+import { StrictMode, use, useState, type ReactNode } from 'react';
 import { Text } from 'react-native';
 
 import { ExpoRoot } from '../../ExpoRoot';
@@ -231,6 +231,29 @@ describe('navigation builder registration', () => {
     const nextState = entry.reduce(layoutState, StackActions.push('second'));
 
     expect(nextState?.routes.map((route) => route.name)).toEqual(['index', 'second']);
+  });
+
+  it('preserves registry identity when screens do not change', () => {
+    let rerenderLayout: () => void;
+    function Layout() {
+      const [, setRenderCount] = useState(0);
+      rerenderLayout = () => setRenderCount((count) => count + 1);
+      return <Stack />;
+    }
+
+    renderRouter({
+      _layout: Layout,
+      index: Probe,
+    });
+    const initialRegistry = registry;
+    const initialEntry = registry.get(getLayoutState().key);
+    const initialRenders = registryRenders;
+
+    act(() => rerenderLayout());
+
+    expect(registry).toBe(initialRegistry);
+    expect(registry.get(getLayoutState().key)).toBe(initialEntry);
+    expect(registryRenders).toBe(initialRenders);
   });
 
   it('replaces the entry when screens change', () => {
--- a/packages/expo-router/src/global-state/__tests__/routerRegistry.test.ios.tsx
+++ b/packages/expo-router/src/global-state/__tests__/routerRegistry.test.ios.tsx
@@ -1,0 +1,265 @@
+import { jest } from '@jest/globals';
+import { act, render } from '@testing-library/react-native';
+import { StrictMode, use, type ReactNode } from 'react';
+import { Text } from 'react-native';
+
+import { ExpoRoot } from '../../ExpoRoot';
+import { router } from '../../imperative-api';
+import Stack from '../../layouts/Stack';
+import { StackActions, type NavigationState } from '../../react-navigation/native';
+import { getMockContext, renderRouter } from '../../testing-library';
+import { store } from '../router-store';
+import {
+  RouterRegistryProvider,
+  RouterRegistryContext,
+  type RouterRegistry,
+  type RouterRegistryEntry,
+  useRegisterRouter,
+} from '../routerRegistry';
+
+const state: NavigationState = {
+  stale: false,
+  type: 'stack',
+  key: 'stack-key',
+  index: 0,
+  routeNames: ['index'],
+  routes: [{ key: 'index-key', name: 'index' }],
+};
+
+const firstEntry: RouterRegistryEntry = {
+  reduce: () => state,
+  routerType: 'stack',
+};
+
+const secondEntry: RouterRegistryEntry = {
+  reduce: () => state,
+  routerType: 'stack',
+};
+
+function getLayoutState(): NavigationState {
+  const layoutState = store.navigationRef.current!.getRootState().routes[0]!.state;
+
+  if (layoutState?.stale !== false) {
+    throw new Error('Expected initialized layout state');
+  }
+
+  return layoutState;
+}
+
+function Registrant({
+  entry,
+  stateKey = state.key,
+}: {
+  entry: RouterRegistryEntry;
+  stateKey?: string;
+}) {
+  useRegisterRouter(stateKey, entry);
+  return null;
+}
+
+function RegistryProbe({ onRender }: { onRender: (registry: RouterRegistry) => void }) {
+  const registry = use(RouterRegistryContext);
+  if (registry === undefined) {
+    throw new Error('Expected RouterRegistryProvider');
+  }
+  onRender(registry);
+  return null;
+}
+
+describe(RouterRegistryProvider, () => {
+  it('warns when registering outside the provider', () => {
+    const warn = jest.spyOn(console, 'warn').mockImplementation(() => {});
+
+    try {
+      render(<Registrant entry={firstEntry} />);
+
+      expect(warn).toHaveBeenCalledWith(
+        'Router registry is unavailable. This is most likely a bug in expo-router. Please report it at https://github.com/expo/expo/issues.'
+      );
+    } finally {
+      warn.mockRestore();
+    }
+  });
+
+  it('registers and unregisters entries', () => {
+    const registries: RouterRegistry[] = [];
+    const result = render(
+      <RouterRegistryProvider>
+        <RegistryProbe onRender={(registry) => registries.push(registry)} />
+        <Registrant entry={firstEntry} />
+      </RouterRegistryProvider>
+    );
+
+    expect(registries.at(-1)?.get(state.key)).toBe(firstEntry);
+
+    result.rerender(
+      <RouterRegistryProvider>
+        <RegistryProbe onRender={(registry) => registries.push(registry)} />
+      </RouterRegistryProvider>
+    );
+
+    expect(registries.at(-1)?.size).toBe(0);
+  });
+
+  it('keeps a newer registration when an older owner unmounts', () => {
+    const registries: RouterRegistry[] = [];
+    const result = render(
+      <RouterRegistryProvider>
+        <RegistryProbe onRender={(registry) => registries.push(registry)} />
+        <Registrant entry={firstEntry} />
+        <Registrant entry={secondEntry} />
+      </RouterRegistryProvider>
+    );
+
+    expect(registries.at(-1)?.get(state.key)).toBe(secondEntry);
+
+    result.rerender(
+      <RouterRegistryProvider>
+        <RegistryProbe onRender={(registry) => registries.push(registry)} />
+        <Registrant entry={secondEntry} />
+      </RouterRegistryProvider>
+    );
+
+    expect(registries.at(-1)?.get(state.key)).toBe(secondEntry);
+  });
+
+  it('handles StrictMode effect replay', () => {
+    const registries: RouterRegistry[] = [];
+
+    render(
+      <StrictMode>
+        <RouterRegistryProvider>
+          <RegistryProbe onRender={(registry) => registries.push(registry)} />
+          <Registrant entry={firstEntry} />
+        </RouterRegistryProvider>
+      </StrictMode>
+    );
+
+    expect(registries.at(-1)?.get(state.key)).toBe(firstEntry);
+  });
+
+  it('preserves map identity when registration does not change', () => {
+    const registries: RouterRegistry[] = [];
+
+    render(
+      <RouterRegistryProvider>
+        <RegistryProbe onRender={(registry) => registries.push(registry)} />
+        <Registrant entry={firstEntry} />
+        <Registrant entry={firstEntry} />
+      </RouterRegistryProvider>
+    );
+
+    expect(new Set(registries).size).toBe(2);
+    expect(registries.at(-1)?.get(state.key)).toBe(firstEntry);
+  });
+});
+
+describe('navigation builder registration', () => {
+  let registry: RouterRegistry;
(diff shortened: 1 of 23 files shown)


packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
```diff
@@ -566,6 +567,24 @@ export function useNavigationBuilder<
   // `ROUTE_NAMES_CHANGED` commits and no screen is focused only to be unfocused again.
   state = router.getStateForDeclaredRoutes(state, routeNames);
 
+  const reduce = React.useCallback<RouterRegistryEntry['reduce']>(
+    (registryState, action) =>
+      // The registry stores states from different router types; this entry only receives its own state key.
+      router.getStateForAction(registryState as State, action, {
+        routeNames,
+        routeParamList,
```
jakub-agent: [🚨 **critical**] `routeParamList` is not defined in this scope. There is no binding with that name anywhere in the module — the enclosing `useMemo` builds a local `initialRouteParamList` instead, and the other `getStateForAction` call in this file (line 523) passes only `routeNames` and `routeGetIdList`.

CI on this commit reports:

```
src/react-navigation/core/useNavigationBuilder.tsx(575,9): error TS2353: Object literal may only specify known properties, and 'routeParamList' does not exist in type 'RouterActionOptions'.
src/react-navigation/core/useNavigationBuilder.tsx(575,9): error TS18004: No value exists in scope for the shorthand property 'routeParamList'.
src/react-navigation/core/useNavigationBuilder.tsx(578,34): error TS2304: Cannot find name 'routeParamList'.
```

and at runtime `ReferenceError: routeParamList is not defined`, thrown from the dependency array on line 578 — so it fires on every render of every navigator, not only when `reduce` is called. This fails `check-packages` and the whole `expo-router` jest suite, and blocks the PRs stacked on this one (#48760, #48731).

`RouterActionOptions` has no `routeParamList` member, so the fix is to remove the property and its dependency rather than declare a value for it.

`packages/expo-router/AGENTS.md` asks for `et check-packages expo-router` before committing — please run it on this branch.

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
```diff
@@ -566,6 +567,24 @@ export function useNavigationBuilder<
...
+        routeGetIdList,
+      }),
+    [routeGetIdList, routeNames, routeParamList, router]
+  );
+  const registryEntry = React.useMemo<RouterRegistryEntry>(
```
jakub-agent, line 587: [🏗️ **design**] This memoization has no effect. `routeNames` (line 345) and `routeGetIdList` (line 346) are a fresh array and a fresh object on every render, so the `useCallback` dependencies always differ. That cascades: new `reduce` → new `registryEntry` → `useRegisterRouter`'s layout effect re-runs on every render of every navigator.

Each of those re-runs calls `unregister` then `register`, so the provider clones the registry `Map` twice and commits two state updates. Because they are scheduled from a layout effect, React flushes an extra synchronous render pass before paint. Once a real consumer subscribes to `RouterRegistryContext`, it will re-render on every render of any navigator in the tree, which defeats the point of keying the registry by state key.

Two options that keep the entry identity stable:

- Compare by value before producing a new entry — this file already imports `isArrayEqual` for `routeNames`, and `routeGetIdList` can be compared with the existing `isRecordEqual` helper.
- Hold the entry in a ref and mutate its fields, registering once per `state.key`. Consumers call `reduce` at action time, so they always observe the latest configuration without a new map identity.

The second option would need the `replaces the entry when screens change` test reworked to assert on `reduce` behaviour rather than on entry identity.

packages/expo-router/src/global-state/routerRegistry.tsx
```diff
@@ -0,0 +1,79 @@
...
+}
+
+export function useRegisterRouter(stateKey: string, entry: RouterRegistryEntry): void {
+  const setters = use(RouterRegistrySettersContext);
+
+  useClientLayoutEffect(() => {
+    if (setters === undefined) {
+      if (process.env.NODE_ENV !== 'production') {
+        console.warn(
+          'Router registry is unavailable. This is most likely a bug in expo-router. Please report it at https://github.com/expo/expo/issues.'
```
jakub-agent, line 71: [🏗️ **design**] This warning fires for setups that are not bugs. `useRegisterRouter` is called unconditionally from `useNavigationBuilder`, so any navigator rendered outside `ExpoRoot` hits it — including this package's own vendored tests, which render navigators inside a bare `BaseNavigationContainer` (see `src/react-navigation/core/__tests__/index.test.ios.tsx`, `useDescriptors.test.ios.tsx`, and others). Those suites will fill with a message telling the reader to file an expo-router issue.

It also repeats: `entry` is in the dependency array and changes identity on every render (see the comment on `useNavigationBuilder.tsx:580`), so this logs once per render rather than once per mount.

Suggestion: either treat a missing provider as a supported standalone case and drop the warning, or keep it but reword it to describe the actual condition ("a navigator rendered outside `ExpoRoot`") and de-duplicate it per navigator.

packages/expo-router/src/global-state/routerRegistry.tsx
```diff
@@ -0,0 +1,79 @@
...
+type RouterRegistrySetters = {
+  register: (stateKey: string, entry: RouterRegistryEntry) => void;
+  unregister: (stateKey: string, entry: RouterRegistryEntry) => void;
+};
+
+export const RouterRegistryContext = createContext<RouterRegistry | undefined>(undefined);
+const RouterRegistrySettersContext = createContext<RouterRegistrySetters | undefined>(undefined);
+
+export function RouterRegistryProvider({ children }: PropsWithChildren) {
+  const [registry, setRegistry] = useState<RouterRegistry>(() => new Map());
```
jakub-agent, line 30: [🏗️ **design**] Design question about the storage mechanism, given the PR is titled a *global* registry.

Holding the map in React state means two things: the registry can only be read by components inside the provider during render, and every mutation re-renders every consumer even when the consumer only cares about one state key. If the intended consumer is the imperative router store (`global-state/router-store.tsx`), it cannot read this synchronously at action-dispatch time, and it will observe entries one commit late.

A ref-backed `Map` exposed through `useSyncExternalStore`, or a slice on the existing store, would give both a synchronous imperative read and per-key subscription, at the cost of manual change notification. Was that considered and rejected? If the consumer in a later PR of the stack is a React component that genuinely wants the whole map, the current shape is fine — worth a note in the file explaining who reads it.

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
```diff
@@ -4,6 +4,7 @@ import { use } from 'react';
 // TODO(@ubax) - RN Migration: remove this dependency and just add this function to our codebase
 import { isValidElementType } from 'react-is';
 
+import { type RouterRegistryEntry, useRegisterRouter } from '../../global-state/routerRegistry';
```
jakub-agent, line 7: [🏗️ **design**] This makes the vendored `react-navigation/core` depend on expo-router's `global-state`, which inverts the layering the fork relies on — `AGENTS.md` describes `src/react-navigation/` as vendored upstream source, and this file already carries a `TODO(@ubax) - RN Migration` note about shedding a non-fork dependency. Every future upstream sync of `useNavigationBuilder` now has to re-apply this edit.

Related: `contextKey: options.id` on line 581 reads an expo-router meaning into a generic React Navigation field. `id` is only the context key because `views/Navigator.tsx` sets `id: contextKey`; for any other navigator built on this hook it is an arbitrary identifier, so the registry entry's `contextKey` will be wrong or undefined.

Suggestion: register through a fork-level context (the way `UnhandledActionContext` is injected) so the fork stays free of expo-router imports, and name the field `id` unless the expo-router meaning is guaranteed.

packages/expo-router/src/global-state/__tests__/routerRegistry.test.ios.tsx
```diff
@@ -0,0 +1,265 @@
...
+          <RegistryProbe onRender={(registry) => registries.push(registry)} />
+          <Registrant entry={firstEntry} />
+        </RouterRegistryProvider>
+      </StrictMode>
+    );
+
+    expect(registries.at(-1)?.get(state.key)).toBe(firstEntry);
+  });
+
+  it('preserves map identity when registration does not change', () => {
```
jakub-agent, line 141: [💡 suggestion] This test pins identity only for a repeated registration of the same entry object. The case that actually matters in production is not covered: re-rendering an unchanged navigator should not churn the registry.

A test that renders a navigator, records the registry identity, forces a re-render with the same screens, and asserts the identity is unchanged would pin the contract discussed on `useNavigationBuilder.tsx:580` and fail today.

jakub-agent (changes requested):

**Blocking: this branch does not compile.** `routeParamList` on line 575 of `useNavigationBuilder.tsx` is not declared anywhere in that scope. CI confirms three TypeScript errors (`TS2304`, `TS18004`, `TS2353`) and a runtime `ReferenceError: routeParamList is not defined`, which fails `check-packages` and the entire `expo-router` jest suite. The dependency array on line 578 evaluates the identifier on every render, so every navigator throws. `RouterActionOptions` has no `routeParamList` field, so the fix is to drop the property, matching the `getStateForAction` call on line 523. `packages/expo-router/AGENTS.md` requires `et check-packages expo-router` before committing; that would have caught this.

This is PR 9 of an ~11 PR stack (`main` ← #48708 ← #48756 ← #48752 ← #48757 ← #48709 ← #48718 ← #48750 ← **#48707** ← #48760, #48731), so the failure also blocks everything stacked above it.

The design is otherwise reasonable: a context-based map from navigation state key to router entry, registered in a layout effect with identity-checked unregistration. The StrictMode and stale-unregister handling are correct, and the tests cover mount, unmount, replacement, and effect replay.

Remaining concerns, in order:

1. **The memoization does not work.** `routeNames` and `routeGetIdList` are rebuilt on every render (lines 345-352), so `reduce`, `registryEntry`, and the registration effect all re-run on every render of every navigator. Each render clones the registry `Map` twice and forces a synchronous re-render of the provider.
2. **The dev warning fires for legitimate setups.** Any navigator rendered outside `ExpoRoot` warns, including this package's own vendored `react-navigation/core` tests, and the message blames a bug in expo-router.
3. **Layering.** The vendored fork now imports expo-router global state, and `contextKey: options.id` bakes an expo-router meaning into a generic React Navigation field.
4. **No production consumer yet.** Nothing outside the tests reads the registry, so as merged this ships a provider and per-navigator state that no shipped code uses. Reasonable for a stack step, worth confirming the consumer lands before this reaches a release branch.

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
```diff
@@ -567,6 +568,30 @@ export function useNavigationBuilder<
   // `ROUTE_NAMES_CHANGED` commits and no screen is focused only to be unfocused again.
   state = router.getStateForDeclaredRoutes(state, routeNames);
 
+  // TODO(@ubax): find a better way to implement this then ref approach
+  const registryConfigRef = React.useRef({ routeNames, routeGetIdList });
+  registryConfigRef.current = { routeNames, routeGetIdList };
```
jakub-agent: [🏗️ **design**] Agreeing with your `TODO` above — and there is a concrete fix that keeps everything else about this approach.

Writing a ref during render has two costs here:

1. **React Compiler.** `AGENTS.md` asks that the code work with and without the compiler enabled. A render-phase ref write trips `ValidateNoRefAccessInRender`, so the compiler bails out of optimizing `useNavigationBuilder` — the hottest hook in the package — rather than just skipping this line.
2. **Discarded renders.** A render pass that never commits (an interrupted transition, an `<Activity>` prerender — which this hook already handles specially) still leaves its `routeNames` and `routeGetIdList` in the ref. If `reduce` is dispatched from the committed tree in the window before the next render, it reduces with screen configuration that was never committed. The window is narrow, but it is real.

Moving the assignment into a layout effect fixes both, and puts the ref write in the same commit phase as the registration right below it, so the entry and the config it reads always come from the same commit:

```ts
const registryConfigRef = React.useRef({ routeNames, routeGetIdList });

useClientLayoutEffect(() => {
  registryConfigRef.current = { routeNames, routeGetIdList };
});
```

Minor point either way: with the assignment on the next line, the `useRef` initializer is dead after the first render but still allocates an object on every render.

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
```diff
@@ -567,6 +568,30 @@ export function useNavigationBuilder<
...
+  // Screen-list changes invalidate render consumers even though the reducer reads the latest config.
+  const routeNamesKey = routeNames.join(',');
```
jakub-agent: [🐧 nit] `,` is ambiguous as a separator: `['a,b']` and `['a', 'b']` produce the same key. Route names come from file paths, where a comma is a legal character, so this is reachable.

The blast radius is small — `reduce` reads the ref, so reduction stays correct, and only the consumer invalidation is missed. Still, a separator that cannot appear in a route name (`'\u0000'`) or an `isArrayEqual` comparison against a previous-value ref removes the ambiguity for the same cost. This file already imports `isArrayEqual`.

jakub-agent (approved):

**Iteration 2 — all six findings from the previous review are addressed. Approving.**

CI is green on `7a0ac1f`: `check-packages`, all four `jest-ubuntu` shards, and all four `playwright-ubuntu` shards pass. (`ios-test-e2e` reports a 0-second failure with no log, the same as on the previous head; it is not attributable to this diff.)

What changed:

| Previous finding | Resolution |
| --- | --- |
| 🚨 `routeParamList` undefined — 3 TS errors + `ReferenceError` | Property and dependency removed. Typecheck and the full jest suite now pass. |
| 🏗️ Memoization ineffective, registry churned every render | Config moved to a ref, `reduce` keyed on `routeNamesKey`. Entry identity is now stable across re-renders. |
| 🏗️ Dev warning fired for the fork's own tests | New `__fixtures__/BaseNavigationContainer` wraps the provider; 17 fork test files updated. |
| 🏗️ React state as registry storage | Answered with a comment: render-phase consumers make React state the right fit. Reasonable. |
| 🏗️ Fork imports expo-router `global-state` | Deferred with an explicit `TODO(@ubax)` pointing at the `UnhandledActionContext` approach. Fine for a stack step. |
| 💡 No test for registry churn | `preserves registry identity when screens do not change` added, asserting map identity, entry identity, and consumer render count. |

Two non-blocking comments below, neither of which should hold up the merge. The first is the ref write during render — you already flagged it with a `TODO`, and there is a concrete fix worth taking in the follow-up. The second is a nit about the `routeNames.join(',')` separator.

One thing to confirm before this lands on a release branch: nothing outside the tests reads the registry yet, so as merged it ships a provider and per-navigator registration that no shipped code consumes. That is expected for PR 9 of the stack — worth making sure the consumer lands in the same release.

PR #48708, opened 2026-08-10
[router] Remove navigator level `initialRouteName` prop

Ubax:
# Why

For a long time now router exposed an `unstable_setting` for `initialRouteName`. It is much more handy from routing logic perspective then the prop, because it is known in advance and stable across the app runtime.

This PR removes the additional `initialRouteName` from all the navigators leaving the file level one as the only way to configure initial route name

# How

1. Remove `initialRouteName` from public props
2. Replace `<SomeNavigator initialRouteName` with `unstable_settings: { initialRouteName: '...' }, <SomeNavigator` in tests

# Test Plan

CI

# Checklist

### DIFF
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### 🛠 Breaking changes
 
+- Remove the `initialRouteName` prop from Expo Router navigators. Configure the initial route with `unstable_settings.initialRouteName` in the route layout instead. ([#48708](https://github.com/expo/expo/pull/48708) by [@Ubax](https://github.com/Ubax))
 - Remove `NavigationContainerRefContext` fallback in `useNavigation`. The hook now throws when called outside a navigator, including components rendered via `ExpoRoot`'s `wrapper` prop. ([#48638](https://github.com/expo/expo/pull/48638) by [@Ubax](https://github.com/Ubax))
 - Allow `key` to be `undefined` in the `Descriptor` type and in routes passed to screen `options`, navigator `screenOptions`, and `RouteGroupConfig.screenOptions` callbacks. ([#48596](https://github.com/expo/expo/pull/48596) by [@Ubax](https://github.com/Ubax))
 - Unify JS Tabs, TopTabs, Drawer, and headless tabs with NativeTabs - only screens declared in the layout become visible. ([#48499](https://github.com/expo/expo/pull/48499) by [@Ubax](https://github.com/Ubax))
@@ -33,6 +34,7 @@
 
 ### 🐛 Bug fixes
 
+- Make layouts with explicitly declared screens honor `unstable_settings.initialRouteName` instead of declaration order, which can change deep-link back stacks. ([#48708](https://github.com/expo/expo/pull/48708) by [@Ubax](https://github.com/Ubax))
 - Prevent unfocused nested native tab navigators from redirecting global router state. ([#48257](https://github.com/expo/expo/pull/48257) by [@Ubax](https://github.com/Ubax))
 - Fixed `Tabs` and `TopTabs` (`expo-router/js-tabs`, `expo-router/js-top-tabs`) not being usable from RSC ([#48330](https://github.com/expo/expo/pull/48330) by [@Ubax](https://github.com/Ubax))
 - Redirect fully guarded navigators to parent ([#47984](https://github.com/expo/expo/pull/47984) by [@Ubax](https://github.com/Ubax))
--- a/packages/expo-router/src/Route.tsx
+++ b/packages/expo-router/src/Route.tsx
@@ -81,6 +81,42 @@ export function useRouteNode(): RouteNode | null {
   return use(CurrentRouteContext);
 }
 
+export function findRouteNodeByName(
+  children: RouteNode[] | undefined,
+  routeName: string | undefined
+): RouteNode | undefined {
+  if (!routeName) {
+    return undefined;
+  }
+  return children?.find(
+    (child) => child.route === routeName || child.route === `${routeName}/index`
+  );
+}
+
+export function getValidInitialRoute(
+  node: RouteNode | null,
+  initialRouteName = node?.initialRouteName,
+  groupName?: string
+): RouteNode | undefined {
+  if (!node || !initialRouteName) {
+    return undefined;
+  }
+  const route = findRouteNodeByName(node.children, initialRouteName);
+  if (!route) {
+    throw new Error(
+      `The initial route name "${initialRouteName}"${groupName ? ` for group "${groupName}"` : ''} was not found in the layout at "${node.contextKey}". ` +
+        `Available routes are: ${node.children.map(({ route }) => `"${route}"`).join(', ')}. ` +
+        'Set `unstable_settings.initialRouteName` to the name of a route in this layout.'
+    );
+  }
+  return route;
+}
+
+export const getValidInitialRouteName = (
+  node: RouteNode | null,
+  initialRouteName = node?.initialRouteName
+) => getValidInitialRoute(node, initialRouteName)?.route;
+
 export function useContextKey(): string {
   const node = useRouteNode();
   if (node == null) {
--- a/packages/expo-router/src/__tests__/getRoutes.test.ios.ts
+++ b/packages/expo-router/src/__tests__/getRoutes.test.ios.ts
@@ -601,26 +601,25 @@ describe('anchor', () => {
   });
 
   it(`throws if anchor does not match a route`, () => {
-    expect(() => {
+    expect(() =>
       getRoutes(
         inMemoryContext({
           _layout: {
-            unstable_settings: {
-              anchor: 'c',
-            },
+            unstable_settings: { anchor: 'c' },
             default: () => null,
           },
           a: () => null,
           b: () => null,
-        })
-      );
-    }).toThrowErrorMatchingInlineSnapshot(
-      `"Layout ./_layout.js has invalid anchor 'c'. Valid options are: 'a', 'b'"`
+        }),
+        { skipGenerated: true }
+      )
+    ).toThrow(
+      'The initial route name "c" was not found in the layout at "./_layout.js". Available routes are: "a", "b". Set `unstable_settings.initialRouteName` to the name of a route in this layout.'
     );
   });
 
   it(`throws if anchor with group selection does not match a route`, () => {
-    expect(() => {
+    expect(() =>
       getRoutes(
         inMemoryContext({
           '(a,b)/_layout': {
@@ -635,11 +634,46 @@ describe('anchor', () => {
             default: () => null,
           },
           '(a,b)/c': () => null,
-        })
-      );
-    }).toThrowErrorMatchingInlineSnapshot(
-      `"Layout ./(a,b)/_layout.js has invalid anchor 'd' for group '(b)'. Valid options are: 'c'"`
+        }),
+        { skipGenerated: true }
+      )
+    ).toThrow(
+      'The initial route name "d" for group "b" was not found in the layout at "./(a,b)/_layout.js". Available routes are: "c". Set `unstable_settings.initialRouteName` to the name of a route in this layout.'
     );
+  });
+
+  it(`does not label a shared anchor as group-specific`, () => {
+    expect(() =>
+      getRoutes(
+        inMemoryContext({
+          '(a)/_layout': {
+            unstable_settings: { anchor: 'missing' },
+            default: () => null,
+          },
+          '(a)/index': () => null,
+        }),
+        { skipGenerated: true }
+      )
+    ).toThrow(
+      'The initial route name "missing" was not found in the layout at "./(a)/_layout.js". Available routes are: "index".'
+    );
+  });
+
+  it(`resolves a directory anchor to its index route entry point`, () => {
+    const routes = getRoutes(
+      inMemoryContext({
+        _layout: {
+          unstable_settings: { anchor: 'a' },
+          default: () => null,
+        },
+        'a/index': () => null,
+        b: () => null,
+      }),
+      { skipGenerated: true }
+    );
+
+    expect(routes?.initialRouteName).toBe('a/index');
+    expect(routes?.children[0]?.entryPoints).toContain('./a/index.js');
   });
 });
 
--- a/packages/expo-router/src/getRoutesCore.ts
+++ b/packages/expo-router/src/getRoutesCore.ts
@@ -1,4 +1,9 @@
-import type { DynamicConvention, MiddlewareNode, RouteNode } from './Route';
+import {
+  getValidInitialRoute,
+  type DynamicConvention,
+  type MiddlewareNode,
+  type RouteNode,
+} from './Route';
 import {
   matchArrayGroupName,
   matchDynamicName,
@@ -933,6 +938,7 @@ function crawlAndAppendInitialRoutesAndEntryFiles(
       return child.route.replace(/\/index$/, '') === groupName;
     });
     let anchor = childMatchingGroup?.route;
+    let anchorGroupName: string | undefined;
     // We may strip loadRoute during testing
     if (!options.internal_stripLoadRoute) {
       const loaded = node.loadRoute();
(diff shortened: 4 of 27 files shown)


packages/expo-router/src/Route.tsx
```diff
@@ -81,6 +81,16 @@ export function useRouteNode(): RouteNode | null {
   return use(CurrentRouteContext);
 }
 
+export function getValidInitialRouteName(node: RouteNode | null): string | undefined {
```
jakub-agent: [🏗️ **design**] An anchor that resolves to nothing is now dropped silently. Before this PR an unresolvable value reached `useNavigationBuilder`, which throws `Couldn't find a screen named '<name>' to use as 'initialRouteName'.` (`react-navigation/core/useNavigationBuilder.tsx:312`) — that was reachable for NativeTabs, which passed `routeNode?.initialRouteName` straight through.

Not crashing is an improvement, but with the prop gone `unstable_settings` is the only way to set an anchor, so a typo (`initialRouteName: 'idnex'`) now produces no signal at all — the layout just starts on a different screen. Consider a `__DEV__` warning here when `node.initialRouteName` is set but no child matches, listing `node.children.map(c => c.route)` the way `getSortedChildren` does for unmatched `<Screen name>` values.

A navigator-level test for this path would also be useful: the unit tests cover the helper, but nothing asserts that an invalid `unstable_settings.initialRouteName` no longer throws when rendering a layout.

packages/expo-router/src/ui/TabContext.tsx
```diff
@@ -22,17 +22,20 @@ export type ExpoTabsNavigatorScreenOptions = {
...
 
-export type ExpoTabsNavigatorOptions = DefaultNavigatorOptions<
-  ParamListBase,
-  string | undefined,
-  TabNavigationState<ParamListBase>,
-  ExpoTabsScreenOptions,
-  TabNavigationEventMap,
-  ExpoTabsNavigationProp<ParamListBase>
-> &
+export type ExpoTabsNavigatorOptions = Omit<
```
jakub-agent, line 25: [🏗️ **design**] Headless Tabs loses the prop here, but `ui/Tabs.tsx:163-185` still passes `routeNode.initialRouteName` into `useNavigationBuilder` unvalidated — so `unstable_settings` becomes the only input for the one navigator that does not validate it. `useNavigationBuilder` throws when the name is not among the registered screens, so a typo hard-crashes headless Tabs while every other navigator in this PR silently ignores it.

Note `getValidInitialRouteName` is not a drop-in fix here: headless Tabs registers screens from the declared triggers (`useTriggersToScreens` → `common.tsx:189`, `name: config.routeNode.route`), not from `node.children`, so the anchor has to be validated against the trigger-derived routes. An anchor pointing at a filesystem route with no `<TabTrigger>` throws even when it is spelled correctly.

Is leaving `ui/Tabs.tsx` untouched intentional, or out of scope for this PR?

packages/expo-router/src/standard-navigation/index.tsx
```diff
@@ -173,13 +174,14 @@ export function unstable_integrateWithRouter<
...
   function StandardRouterNavigator(props: NavPropsType) {
+    const routeNode = useRouteNode();
     const { extraProps, useNavigationBuilderProps } = partitionNavigatorProps<
       NavigatorOptions,
       State,
       EventMap,
       NavigatorProps,
       RouterOptions
-    >(props);
+    >(props, getValidInitialRouteName(routeNode));
```
jakub-agent, line 184: [🏗️ **design**] This line does more than replace a removed prop. Previously `partitionNavigatorProps` took `initialRouteName` from props only, so `unstable_settings.initialRouteName` never reached the router for `Stack`, `Tabs`, `Drawer` or `TopTabs` — it took effect only through `sortRoutesWithInitial`, which `getSortedChildren` skips once the layout declares its own screens (`useScreens.tsx:98-160`, declaration order wins). So a layout that declares `<Stack.Screen name="index" />` before `<Stack.Screen name="a" />` with `initialRouteName: 'a'` previously anchored on `index` and now anchors on `a`; the updated first test in `issues.test.ios.tsx` is exactly that shape.

That looks like the fix this PR is really delivering, but the CHANGELOG only mentions the prop removal. Worth calling out separately, since apps that declare screens explicitly will see their deep-link back stack change without touching any code.

packages/expo-router/src/Route.tsx
```diff
@@ -81,6 +81,16 @@ export function useRouteNode(): RouteNode | null {
...
+  const initialRouteName = node?.initialRouteName;
+  if (!initialRouteName) {
+    return undefined;
+  }
+  return node?.children.find(
+    (child) => child.route === initialRouteName || child.route === `${initialRouteName}/index`
```
jakub-agent: [💡 suggestion] This predicate is byte-for-byte the one in `getSortedChildren` (`useScreens.tsx:106`), and a fourth variant of the same lookup lives in `GuardContext.findDefaultRedirectRouteInNavigator` (`child.route === node.initialRouteName || normalizeRouteName(child.route) === node.initialRouteName`) plus `findRouteByName` in `useVisibleTabsWithRedirect`. They agree today, but they are independent spellings of one rule (`a` matches `a/index`), so they will drift.

Since this PR makes `unstable_settings` the single source of the anchor, it seems like the right moment to extract one helper — e.g. `findChildByRouteName(children, name)` — and use it from all four sites.

packages/expo-router/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### 🛠 Breaking changes
 
+- Remove the `initialRouteName` prop from Expo Router navigators.
```
jakub-agent: [💡 suggestion] Two things on this entry (the missing PR/author attribution is already covered by expo-bot's suggestion above):

1. **Scope.** "Expo Router navigators" is broader than what changed. `Navigator` and `Slot` still accept and forward `initialRouteName` (`views/Navigator.tsx:31,44,73` — and it is documented that way in `docs/pages/router/migrate/from-react-navigation.mdx:387`), and so does the JS stack exported from `expo-router/js-stack`, since `withLayoutContext` forwards props to `createStackNavigator`, which still destructures `initialRouteName`. If those exclusions are intentional, saying which navigators are affected would help; if not, they look like gaps.

2. **Migration path.** A breaking entry that removes a prop should name the replacement: use `unstable_settings = { initialRouteName: '…' }` in the layout file instead.

jakub-agent (commented):

The direction is right: `unstable_settings.initialRouteName` becomes the single way to set an anchor, and the type-level `Omit`s cover every navigator built on `unstable_integrateWithRouter` (`Stack`, `Tabs`, `Drawer`, `TopTabs`) plus the experimental stack, the web modal stack, and NativeTabs. The new `getValidInitialRouteName` validates against `node.children` and returns `child.route`, which matches the screen names that `useSortedScreens` registers, so the resolution domain is correct for those navigators.

Two things stand out.

**1. The change does more than remove a prop.** Before this PR, the standard navigators never received the anchor as a router option — `partitionNavigatorProps` read it from props only, and the anchor took effect indirectly through `sortRoutesWithInitial`. That sorting is skipped when a layout declares its screens (`getSortedChildren` uses declaration order once `order` is non-empty), so an anchor that was not the first declared screen was silently ignored. It is now honoured. The modified test in `issues.test.ios.tsx` encodes exactly this case. That is a user-visible behaviour change for existing apps, and the CHANGELOG entry does not mention it.

**2. Diagnostics get quieter, and one navigator is left out.** An unresolvable anchor used to throw from `useNavigationBuilder` (`Couldn't find a screen named '…' to use as 'initialRouteName'.`); it is now dropped silently. Headless Tabs (`expo-router/ui`) loses the prop in this PR but still forwards the raw value, so it keeps throwing — the two paths now disagree.

Also worth filling in: the PR body is still the unedited template (no Why / How / Test Plan) for a change marked as breaking.

CI is green on this commit.

packages/expo-router/src/Route.tsx
Ubax: Show a warning validated against the filesystem/manifest/routes config. Show the wrong name, and available ones. Show the file path to the incorrect layout

packages/expo-router/src/ui/TabContext.tsx
Ubax: Use getValidInitialRouteName here as well

packages/expo-router/src/standard-navigation/index.tsx
Ubax: Add an entry to bug-fixes as well

packages/expo-router/src/Route.tsx
Ubax: Do it

packages/expo-router/CHANGELOG.md
Ubax: > Scope. "Expo Router navigators" is broader than what changed. Navigator and Slot still accept and forward initialRouteName (views/Navigator.tsx:31,44,73

Unify with other navigators as well

> and it is documented that way in docs/pages/router/migrate/from-react-navigation.mdx:387

Add TODO to update docs after the migration

> and so does the JS stack exported from expo-router/js-stack, since withLayoutContext forwards props to createStackNavigator, which still destructures initialRouteName. 

Unify with other navigators as well

> Migration path. A breaking entry that removes a prop should name the replacement: use

Add a todo to update docs after migration

packages/expo-router/src/getRoutesCore.ts
```diff
@@ -961,27 +966,12 @@ function crawlAndAppendInitialRoutesAndEntryFiles(
...
-          throw new Error(
-            `Layout ${node.contextKey} has invalid anchor '${anchor}'. Valid options are: ${validAnchorRoutes}`
-          );
-        }
-      }
-
       // Navigators can add initialRoutes into the history, so they need to be included in the entryPoints
-      node.initialRouteName = anchor;
-      entryPoints.push(anchorRoute.contextKey);
+      const anchorRoute = getValidInitialRoute(node, anchor);
```
jakub-agent: [🏗️ **design**] This turns a hard error into a warning, which is a second breaking change and is not in the CHANGELOG. Previously an invalid anchor failed route-tree construction (`Layout ./_layout.js has invalid anchor 'c'. Valid options are: 'a', 'b'`) in every environment; now `getValidInitialRoute` only warns under `process.env.NODE_ENV !== 'production'`, so in a production bundle an invalid anchor is completely silent and `node.initialRouteName` just becomes `undefined` — deep-link back stacks quietly stop working.

Relaxing this makes sense for the case the old check got wrong: it compared `child.route === anchor` exactly, so `initialRouteName: 'a'` with `app/a/index.tsx` threw even though it is valid. Now that `findRouteNodeByName` resolves that, the remaining failures are genuine typos — is a warning still the right level, or would keeping the throw for the unresolvable case be better? Either way it deserves a CHANGELOG entry, since anyone relying on the build-time error loses it.

packages/expo-router/src/Route.tsx
```diff
@@ -81,6 +81,40 @@ export function useRouteNode(): RouteNode | null {
...
+export function getValidInitialRoute(
+  node: RouteNode | null,
+  initialRouteName = node?.initialRouteName
+): RouteNode | undefined {
+  if (!node || !initialRouteName) {
+    return undefined;
+  }
+  const route = findRouteNodeByName(node.children, initialRouteName);
+  if (!route && process.env.NODE_ENV !== 'production') {
+    console.warn(
```
jakub-agent: [🏗️ **design**] This warning will repeat for a single typo. `getRoutes` only runs the crawl when `!options.ignoreEntryPoints` (`getRoutesCore.ts:125`), and the runtime store passes `ignoreEntryPoints: true` (`global-state/useStore.ts:43`) — so in an app `node.initialRouteName` keeps the raw, unvalidated value from `getLayoutNode`, and the render-time calls are the ones doing the work. `getValidInitialRoute` is called straight from the body of `StandardRouterNavigator`, `Navigator`, `SlotNavigator`, `JSStack`, `useTabsWithTriggers`, `ExperimentalStackNavigator` and `ModalStackNavigator`, none of them memoized, so an invalid anchor logs once per render of that layout (twice under StrictMode's double render, plus once more from `useVisibleTabsWithRedirect`).

The new `warns instead of throwing for an invalid configured anchor` test confirms the render-time path is the live one, and `toHaveBeenCalledWith` passes regardless of count, so it won't catch the repetition.

A `warnOnce` keyed by `node.contextKey` would fix it. Alternatively, warn once at the `getRoutes`/store boundary where the raw setting is first seen and keep this function pure — the render-time call sites only need the resolution, not the diagnostic.

packages/expo-router/src/layouts/GuardContext.tsx
```diff
@@ -119,17 +119,11 @@ function removeFallbacksTargetingNavigator(
 }
 
 function findDefaultRedirectRouteInNavigator(
-  node: Pick<RouteNode, 'children' | 'initialRouteName'>,
+  node: RouteNode,
   guardedRedirects: GuardedRedirects
 ): RouteNode | undefined {
+  const anchor = getValidInitialRoute(node);
```
jakub-agent, line 125: [💡 suggestion] The anchor is resolved here but the sort on the next line still gets the raw value: `sortRoutesWithInitial(node.initialRouteName)` compares `a.route === initialRouteName` exactly (`sortRoutes.ts:64-71`), so for `initialRouteName: 'a'` with a child `a/index` it silently no-ops while `anchor` above resolves correctly. `getSortedChildren` (`useScreens.tsx:174`) passes the same raw value.

Since the router now receives the resolved name explicitly, the sort may just be redundant for the anchor — but as written the two lines disagree about what `node.initialRouteName` means. Passing `anchor?.route` here (and the resolved name in `useScreens`) would make them consistent.

packages/expo-router/src/__tests__/getRoutes.test.ios.ts
```diff
@@ -635,11 +638,31 @@ describe('anchor', () => {
...
-        })
-      );
-    }).toThrowErrorMatchingInlineSnapshot(
-      `"Layout ./(a,b)/_layout.js has invalid anchor 'd' for group '(b)'. Valid options are: 'c'"`
+        }),
+        { skipGenerated: true }
+      )
+    ).toBeDefined();
+    expect(warnSpy).toHaveBeenCalledWith(
+      'The initial route name "d" was not found in the layout at "./(a,b)/_layout.js". Available routes are: "c".'
```
jakub-agent: [💡 suggestion] This assertion shows the diagnostic losing information the old error carried. The previous message was `Layout ./(a,b)/_layout.js has invalid anchor 'd' for group '(b)'. Valid options are: 'c'` — for an array group with per-group `unstable_settings` (the documented shared-routes pattern), the layout is crawled once per group and the old error said which group's setting was wrong. The new message is identical for both groups, so the user has to guess.

`groupName` is in scope at the call site, so it could be threaded into `getValidInitialRoute` and appended when present.

packages/expo-router/src/standard-navigation/useVisibleTabsWithRedirect.ts
```diff
@@ -65,8 +65,9 @@ export function useVisibleTabsWithRedirect<
     if (guardRedirect !== undefined) {
       return guardRedirect;
     }
+    const initialRouteName = getValidInitialRouteName(routeNode);
```
jakub-agent: [🐧 nit] The memo now reads `routeNode.children` through the helper, but the dependency list below is still `[buildHref, guardRedirect, routeNode?.initialRouteName, visibleRoutes]`. `visibleRoutes` changes whenever the tree is rebuilt, so this is unlikely to be observable today, but the deps no longer describe what the memo reads — `routeNode` would.

jakub-agent (approved):

**Iteration 2** — re-reviewed after the force-push (`40db57c` → `7255731f`, now 26 files). Every point from the previous review is addressed, and the new tests are the right ones:

- `findRouteNodeByName` is extracted and reused by `getSortedChildren`, `GuardContext`, and `getValidInitialRoute` — the divergent spellings are gone.
- `getValidInitialRoute` warns in development instead of dropping an unresolvable anchor silently.
- `ui/Tabs.tsx` validates the anchor, and `UseTabsOptions` omits the prop.
- `Navigator`, `Slot`, and `expo-router/js-stack` lose the prop too, with type tests asserting it and runtime overrides that beat a stale JS prop.
- The CHANGELOG now carries attribution, the `unstable_settings` migration path, and a separate bug-fix entry for the anchor being honoured with explicitly declared screens.

One correction to my previous review: I claimed headless Tabs registers its screens from the declared triggers, so validating against `node.children` would be the wrong domain. That was wrong — `useTriggersToScreens` passes the trigger screens to `useSortedScreens` as *ordering* only, and all filesystem children are still registered, so `node.children` is the correct domain and an anchor without a trigger never threw (the pre-existing `falls back to the first trigger…` test covers that). The change to `ui/Tabs.tsx` is still worth having for consistency and for the typo case.

Four follow-ups below, none of them blocking. The one I would still do before merge is the CHANGELOG: replacing the `getRoutesCore` throw with a warning is a second breaking change in this PR and is not recorded anywhere. The PR body is also still the unedited template, which is thin for a change of this shape.

Approving — the remaining items are diagnostics and polish, not correctness. All four `jest-ubuntu` shards pass on `7255731f`; the native build jobs were still running when this was written.

PR #48709, opened 2026-08-10
[router] Make `history` property optional in tab state

Ubax:
# Why

In order to be able to create initial state without knowing the unmounted navigator types first, we need to make the all navigator specific properties optional. This PR changes `history` to be optional

# How

1. Change the type to `| undefined`
2. Add fallback to `?? []`
3. Rework the logic for when history is not present - no action is dispatched until then, so `state.history = state.routes`

# Test Plan

CI

# Checklist

## Accepted breaking changes

`history` becomes optional on `TabNavigationState` and `DrawerNavigationState`. Code that reads
`state.history` off these types must handle `undefined`.

### DIFF
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### 🛠 Breaking changes
 
+- Make `history` optional in `TabNavigationState` and `DrawerNavigationState` ([#48709](https://github.com/expo/expo/pull/48709) by [@Ubax](https://github.com/Ubax))
 - Make navigation state `type` optional for custom routers. ([#48757](https://github.com/expo/expo/pull/48757) by [@Ubax](https://github.com/Ubax))
 - Handle `PUSH` in tab and drawer routers instead of coercing it to `NAVIGATE`. Custom routers must now handle the `PUSH` action in `getStateForAction`. ([#48752](https://github.com/expo/expo/pull/48752) by [@Ubax](https://github.com/Ubax))
 - Remove the `initialParams` prop from Expo Router screens and `routeParamList` from custom router action options. ([#48756](https://github.com/expo/expo/pull/48756) by [@Ubax](https://github.com/Ubax))
--- a/packages/expo-router/src/react-navigation/routers/DrawerRouter.tsx
+++ b/packages/expo-router/src/react-navigation/routers/DrawerRouter.tsx
@@ -1,6 +1,7 @@
 import { nanoid } from 'nanoid/non-secure';
 
 import {
+  ensureStateHistory,
   type TabActionHelpers,
   TabActions,
   type TabActionType,
@@ -38,7 +39,7 @@ export type DrawerNavigationState<ParamList extends ParamListBase> = Omit<
   /**
    * List of previously visited route keys and drawer open status.
    */
-  history: ({ type: 'route'; key: string } | { type: 'drawer'; status: DrawerStatus })[];
+  history?: ({ type: 'route'; key: string } | { type: 'drawer'; status: DrawerStatus })[];
 };
 
 export type DrawerActionHelpers<ParamList extends ParamListBase> = TabActionHelpers<ParamList> & {
@@ -81,10 +82,21 @@ export function DrawerRouter({
   DrawerNavigationState<ParamListBase>,
   DrawerActionType | CommonNavigationAction
 > {
+  const { backBehavior = 'firstRoute', initialRouteName } = rest;
+
   const router = TabRouter(rest) as unknown as Router<
     DrawerNavigationState<ParamListBase>,
     TabActionType | CommonNavigationAction
   >;
+
+  // `ensureStateHistory` is typed for the tab state. The drawer state differs only by the extra
+  // drawer entries in `history`, which reconstruction never produces.
+  const ensureDrawerStateHistory = (state: DrawerNavigationState<ParamListBase>) =>
+    ensureStateHistory(
+      state as unknown as TabNavigationState<ParamListBase>,
+      backBehavior,
+      initialRouteName
+    ) as unknown as DrawerNavigationState<ParamListBase>;
 
   const isDrawerInHistory = (
     state: DrawerNavigationState<ParamListBase> | PartialState<DrawerNavigationState<ParamListBase>>
@@ -100,7 +112,7 @@ export function DrawerRouter({
     return {
       ...state,
       history: [
-        ...state.history,
+        ...(state.history ?? []),
         {
           type: 'drawer',
           status: defaultStatus === 'open' ? 'closed' : 'open',
@@ -118,7 +130,7 @@ export function DrawerRouter({
 
     return {
       ...state,
-      history: state.history.filter((it) => it.type !== 'drawer'),
+      history: (state.history ?? []).filter((it) => it.type !== 'drawer'),
     };
   };
 
@@ -194,7 +206,10 @@ export function DrawerRouter({
       return closeDrawer(result);
     },
 
-    getStateForAction(state, action, options) {
+    getStateForAction(inputState, action, options) {
+      // Restore route history before drawer actions can add drawer-only history.
+      const state = ensureDrawerStateHistory(inputState);
+
       switch (action.type) {
         case 'OPEN_DRAWER':
           return openDrawer(state);
--- a/packages/expo-router/src/react-navigation/routers/TabRouter.tsx
+++ b/packages/expo-router/src/react-navigation/routers/TabRouter.tsx
@@ -66,7 +66,7 @@ export type TabNavigationState<ParamList extends ParamListBase> = Omit<
   /**
    * List of previously visited route keys.
    */
-  history: { type: 'route'; key: string; params?: object | undefined }[];
+  history?: { type: 'route'; key: string; params?: object | undefined }[];
   /**
    * List of routes' key, which are supposed to be preloaded before navigating to.
    */
@@ -102,6 +102,9 @@ export type TabActionHelpers<ParamList extends ParamListBase> = {
       : never
   ): void;
 };
+
+type TabNavigationStateWithHistory = TabNavigationState<ParamListBase> &
+  Required<Pick<TabNavigationState<ParamListBase>, 'history'>>;
 
 const TYPE_ROUTE = 'route' as const;
 
@@ -155,7 +158,7 @@ const getRouteHistory = (
   index: number,
   backBehavior: BackBehavior,
   initialRouteName: string | undefined
-) => {
+): NonNullable<TabNavigationState<ParamListBase>['history']> => {
   if (routes.length === 0) {
     return [];
   }
@@ -205,8 +208,41 @@ const getRouteHistory = (
   return history;
 };
 
-const changeIndex = (
+export const ensureStateHistory = (
   state: TabNavigationState<ParamListBase>,
+  backBehavior: BackBehavior,
+  initialRouteName: string | undefined
+): TabNavigationStateWithHistory => {
+  if (state.history != null) {
+    // The null check narrows the optional property, but TypeScript doesn't narrow the object type.
+    return state as TabNavigationStateWithHistory;
+  }
+
+  const routes = orderRoutesByRouteNames(state.routes, state.routeNames);
+  const focusedRoute = state.routes[state.index];
+  const index = routes.findIndex((route) => route.key === focusedRoute?.key);
+
+  // `orderRoutesByRouteNames` drops undeclared routes, so the focused one can be missing from
+  // `routes`. Keep it in history anyway, so the current route is always the last entry.
+  const history =
+    index === -1
+      ? focusedRoute === undefined
+        ? []
+        : [{ type: TYPE_ROUTE, key: focusedRoute.key }]
+      : getRouteHistory(routes, index, backBehavior, initialRouteName);
+
+  if (backBehavior === 'fullHistory' && focusedRoute !== undefined) {
+    history[history.length - 1] = {
+      ...history[history.length - 1]!,
+      params: focusedRoute.params,
+    };
+  }
+
+  return { ...state, history };
+};
+
+const changeIndex = (
+  state: TabNavigationStateWithHistory,
   index: number,
   backBehavior: BackBehavior,
   initialRouteName: string | undefined
@@ -344,7 +380,8 @@ export function TabRouter({
       );
     },
 
-    getStateForRouteFocus(state, key) {
+    getStateForRouteFocus(inputState, key) {
+      const state = ensureStateHistory(inputState, backBehavior, initialRouteName);
       const index = state.routes.findIndex((r) => r.key === key);
 
       if (index === -1 || index === state.index) {
@@ -354,7 +391,9 @@ export function TabRouter({
       return changeIndex(state, index, backBehavior, initialRouteName);
     },
 
-    getStateForAction(state, action, { routeGetIdList }) {
+    getStateForAction(inputState, action, { routeGetIdList }) {
+      const state = ensureStateHistory(inputState, backBehavior, initialRouteName);
+
       if (action.target && action.target !== state.key) {
         return null;
       }
@@ -696,8 +735,8 @@ export function TabRouter({
 }
 
 function removeReplacedRouteFromHistory(
-  previousState: TabNavigationState<ParamListBase>,
-  nextState: TabNavigationState<ParamListBase>
+  previousState: TabNavigationStateWithHistory,
+  nextState: TabNavigationStateWithHistory
 ) {
   const replacedRouteKey = previousState.routes[previousState.index]?.key;
   const focusedRouteKey = nextState.routes[nextState.index]?.key;
--- a/packages/expo-router/src/react-navigation/routers/__tests__/DrawerRouter.test.tsx
+++ b/packages/expo-router/src/react-navigation/routers/__tests__/DrawerRouter.test.tsx
@@ -2,9 +2,11 @@ import { expect, jest, test } from '@jest/globals';
 
 import {
   CommonActions,
+  type DrawerActionType,
   DrawerActions,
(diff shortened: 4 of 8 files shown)


packages/expo-router/src/react-navigation/routers/DrawerRouter.tsx
```diff
@@ -38,7 +38,7 @@ export type DrawerNavigationState<ParamList extends ParamListBase> = Omit<
   /**
    * List of previously visited route keys and drawer open status.
    */
-  history: ({ type: 'route'; key: string } | { type: 'drawer'; status: DrawerStatus })[];
+  history?: ({ type: 'route'; key: string } | { type: 'drawer'; status: DrawerStatus })[];
```
jakub-agent, line 42: [🏗️ **design**] Making `history` optional here contradicts `getDrawerStatusFromState`, which still throws when `history == null`:

```ts
// src/react-navigation/drawer/utils/getDrawerStatusFromState.tsx:6
if (state.history == null) {
  throw new Error("Couldn't find the drawer status in the state object. ...");
}
```

`DrawerView.tsx:112` calls it on every render with the raw navigator state. That state is not always normalized: `TabRouter.getRehydratedState` returns `stale: false` states verbatim, and `useNavigationBuilder.isStateInitialized` (`useNavigationBuilder.tsx:369`) skips rehydration for them. So a hand-built `stale: false` drawer state without `history` — exactly what this PR's type change permits — reaches `DrawerView` and throws on first render.

The helper already has the right fallback one line below (`entry?.status ?? state.default ?? 'closed'`). Returning `state.default ?? 'closed'` instead of throwing would make the runtime match the new type.

packages/expo-router/src/react-navigation/routers/TabRouter.tsx
```diff
@@ -206,8 +209,35 @@ const getRouteHistory = (
   return history;
 };
 
-const changeIndex = (
+const ensureStateHistory = (
```
jakub-agent: [🏗️ **design**] `ensureStateHistory` runs at the action boundary only. `getRehydratedState` returns early for `stale: false` states without normalizing them, so consumers that read `state.history` off the rendered state still see `undefined`:

- `drawer/utils/getDrawerStatusFromState.tsx` — throws (see the other comment).
- `ui/TabRouter.tsx:50` — `!state.history?.some(...)` becomes `true`, so `JUMP_TO` clears the target route's nested state on what may not be a first visit.
- `utils/stack.ts:5` — falls through to `state.routes.length`, which is the tab count, not the back-stack depth.

Calling `ensureStateHistory` in the `stale === false` branch of `getRehydratedState` would fix all three at once and keep the invariant "state handed to the view always has `history`", so the optionality stays confined to the state a caller can hand in.

packages/expo-router/src/react-navigation/routers/DrawerRouter.tsx
```diff
@@ -195,6 +195,9 @@ export function DrawerRouter({
     },
 
     getStateForAction(state, action, options) {
+      // Restore route history before drawer actions can add drawer-only history.
+      state = router.getStateForRouteFocus(state, state.routes[state.index]?.key ?? '');
```
jakub-agent: [🏗️ **design**] Using `getStateForRouteFocus` as a normalization call is fragile. It is a no-op only because `TabRouter.getStateForRouteFocus` short-circuits when `index === state.index`; if that method ever grows behavior (drawer close, history dedup, focus events), every drawer action silently inherits it. The `?? ''` sentinel also relies on `''` never matching a route key.

Exporting `ensureStateHistory` from `TabRouter` and calling it directly would say what it means and drop the sentinel. It would also let the `OPEN_DRAWER` / `CLOSE_DRAWER` / `TOGGLE_DRAWER` branches keep the normalized state without routing through a focus method.

Minor: `TabRouter` renamed the parameter to `inputState` rather than reassigning; doing the same here would keep the two routers consistent.

packages/expo-router/src/react-navigation/routers/TabRouter.tsx
```diff
@@ -206,8 +209,35 @@ const getRouteHistory = (
...
+    // The null check narrows the optional property, but TypeScript doesn't narrow the object type.
+    return state as TabNavigationStateWithHistory;
+  }
+
+  const routes = orderRoutesByRouteNames(state.routes, state.routeNames);
+  const focusedKey = state.routes[state.index]?.key;
+  const index = routes.findIndex((route) => route.key === focusedKey);
+
+  const history =
+    index === -1 ? [] : getRouteHistory(routes, index, backBehavior, initialRouteName);
```
jakub-agent: [💡 suggestion] `orderRoutesByRouteNames` drops routes whose name is absent from `routeNames`, so `index` is `-1` whenever the focused route is undeclared, and history collapses to `[]`.

That empty array is not nullish, so `ensureStateHistory` never rebuilds it on later actions — the state stays back-less until a `ROUTE_NAMES_CHANGED` happens to repair it. This is reachable through the same door the PR opens: a hand-built `stale: false` state containing a route the layout does not declare.

Falling back to `[{ type: TYPE_ROUTE, key: focusedKey }]` when `focusedKey` exists but `index === -1` would keep the focused route in history and preserve the current-route invariant the rest of the router assumes.

packages/expo-router/src/react-navigation/routers/__tests__/DrawerRouter.test.tsx
```diff
@@ -36,6 +37,79 @@ test('gets initial state from route names and params with initialRouteName', ()
...
+  ],
+  [
+    DrawerActions.toggleDrawer(),
+    [
+      { type: 'route', key: 'bar' },
+      { type: 'drawer', status: 'open' },
+    ],
+  ],
+  [DrawerActions.closeDrawer(), [{ type: 'route', key: 'bar' }]],
+])('handles $type without history', (action, expectedHistory) => {
```
jakub-agent: [🐧 nit] `$property` interpolation does not apply to array tables. `jest-each` treats a table whose rows are all arrays as a printf table (`table/array.js`: `isTemplates` requires `!isTable(...)`), so all three cases report the literal name `handles $type without history` and are indistinguishable in the output.

Use `%s` with the action as the first column, or switch the table to objects (`{ action, expectedHistory }`) so `$action.type` resolves. The `handles $type` case added in `TabRouter.test.tsx` uses a 1-D object table, which is why `$type` resolves there.

packages/expo-router/src/react-navigation/routers/__tests__/DrawerRouter.test.tsx
```diff
@@ -36,6 +37,79 @@ test('gets initial state from route names and params with initialRouteName', ()
...
+    routeNames: ['bar'],
+    preloadedRouteKeys: [],
+    routes: [{ key: 'bar', name: 'bar' }],
+    default: 'closed',
+  };
+
+  expect(router.getStateForAction(state, action, options)?.history).toEqual(expectedHistory);
+});
+
+test('preserves reconstructed history after closing the drawer', () => {
```
jakub-agent, line 156: [💡 suggestion] Both drawer tests use the default `defaultStatus: 'closed'`, where an open drawer is encoded by the *presence* of a history entry. The `'open'` case inverts that — `openDrawer` removes the entry and `closeDrawer` adds one — so reconstruction interacts with it differently. A case with `DrawerRouter({ defaultStatus: 'open' })` and no `history` would cover that inversion.

Also worth a case: `backBehavior: 'none'` with absent history, since `getRouteHistory` falls through its switch for that value.

packages/expo-router/src/react-navigation/routers/DrawerRouter.tsx
```diff
@@ -118,7 +118,7 @@ export function DrawerRouter({
 
     return {
       ...state,
-      history: state.history.filter((it) => it.type !== 'drawer'),
+      history: (state.history ?? []).filter((it) => it.type !== 'drawer'),
```
jakub-agent, line 133: [🐧 nit] This `?? []` is unreachable: `isDrawerInHistory` returns `false` when `history` is nullish, so the guard above already returned. Plain `state.history.filter(...)` would need a non-null assertion, so the simplest cleanup is to drop the `?? []` and keep the early return as the narrowing point — or leave it and accept the redundancy. The one in `addDrawerToHistory` (line 103) *is* live, so they are not symmetric despite reading that way.

packages/expo-router/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 - Remove the `initialParams` prop from Expo Router screens.
 - Remove the `initialRouteName` prop from Expo Router navigators.
 - Handle `PUSH` in tab and drawer routers instead of coercing it to `NAVIGATE`.
+- Make `history` optional in `TabNavigationState` and `DrawerNavigationState`
```
jakub-agent: [🐧 nit] Missing the PR link and author attribution that every other entry in this section carries, and the trailing period. Per CONTRIBUTING:

```md
- Make `history` optional in `TabNavigationState` and `DrawerNavigationState`. ([#48709](https://github.com/expo/expo/pull/48709) by [@Ubax](https://github.com/Ubax))
```

jakub-agent (commented):

PR 3/3 in a stack (`@ubax/remove-push-to-navigate-modification` → #48757 → this PR). It makes `history` optional on `TabNavigationState` / `DrawerNavigationState` and rebuilds it lazily inside the routers.

The router-side work is sound. `ensureStateHistory` mirrors `changeIndex`'s ordering and `fullHistory` params handling, the early return keeps it cheap, and the new tests cover each `backBehavior` plus the drawer regression.

Main concern: **normalization only happens at the action boundary, so a `history`-less state can still reach the view layer.** `getRehydratedState` returns `stale: false` states verbatim, and `useNavigationBuilder.isStateInitialized` skips rehydration for them. `DrawerView` then calls `getDrawerStatusFromState(state)`, which throws when `history == null`. The type now says that state is legal; the render path says it is not.

Secondary points: the drawer normalizes by calling `getStateForRouteFocus` on the already-focused key, which works only because that method short-circuits; reconstruction silently yields `[]` when the focused route is not in `routeNames`; and one `test.each` title does not interpolate.

Also: the PR description is the unfilled template. For a breaking change in a stack, a short Why / How / Test Plan would help reviewers.

packages/expo-router/src/react-navigation/routers/DrawerRouter.tsx
Ubax: Remove the error and fallback to `[]` similar to https://github.com/expo/expo/pull/48750

packages/expo-router/src/react-navigation/routers/TabRouter.tsx
Ubax: > ui/TabRouter.tsx:50 — !state.history?.some(...) becomes true, so JUMP_TO clears the target route's nested state on what may not be a first visit.

Add one more check `state.routes....`. Also leave a TODO to look into this later

> utils/stack.ts:5 — falls through to state.routes.length, which is the tab count, not the back-stack depth.

This is ok. If there was no action then history.length === state.routes.length (only the focused one is loaded)

packages/expo-router/src/react-navigation/routers/DrawerRouter.tsx
Ubax: > Exporting ensureStateHistory from TabRouter and calling it directly would say what it means and drop the sentinel. It would also let the OPEN_DRAWER / CLOSE_DRAWER / TOGGLE_DRAWER branches keep the normalized state without routing through a focus method.

Do it

> Minor: TabRouter renamed the parameter to inputState rather than reassigning; doing the same here would keep the two routers consistent.

Do it

packages/expo-router/src/react-navigation/routers/__tests__/DrawerRouter.test.tsx
Ubax: Do it

packages/expo-router/src/react-navigation/routers/__tests__/DrawerRouter.test.tsx
Ubax: Addd

packages/expo-router/src/react-navigation/routers/TabRouter.tsx
```diff
@@ -66,7 +66,7 @@ export type TabNavigationState<ParamList extends ParamListBase> = Omit<
   /**
    * List of previously visited route keys.
    */
-  history: { type: 'route'; key: string; params?: object | undefined }[];
+  history?: { type: 'route'; key: string; params?: object | undefined }[];
```
jakub-agent, line 69: [💡 suggestion] `getHistoryLength` in `utils/stack.ts` is the third `history` consumer from the last review, and the one still reading a possibly un-normalized state:

```ts
if (state.history) return state.history.length;
if (state.type === 'stack') { ... }
return state.routes.length;   // now reachable for tab/drawer
```

That last line used to be dead for tab and drawer states, because `history` was required. It is live now, and `routes.length` is the tab count, not the back-stack depth.

`useLinking` uses the value as a delta between two consecutive states of the same navigator (`useLinking.ts:428`). A tab state handed in without `history` reports `routes.length`; the first action normalizes it to the reconstructed length, so with three tabs and `backBehavior: 'firstRoute'` the delta is `1 - 3 = -2`. `historyDelta < 0` drives `history.go()` backwards, so the browser history moves when nothing was popped.

Not reachable through the routers themselves — every state they return has `history`. It needs a hand-built `stale: false` state to be the rendered state on web, which is the case the type change now permits. Returning `1` for a tab/drawer state with no history would match what reconstruction produces for the non-visit behaviors.

packages/expo-router/src/ui/TabRouter.tsx
```diff
@@ -46,8 +46,11 @@ export function ExpoTabRouter(options: ExpoTabRouterOptions) {
         return rnTabRouter.getStateForAction(state, action, routerConfigOptions);
       }
 
-      // We should reset if this is the first time visiting the route
-      let shouldReset = !state.history?.some((item) => item.key === route?.key) && !route.state;
+      // We should reset if this is the first time visiting the route.
+      let shouldReset =
+        !(state.history == null ? state.routes : state.history).some(
```
jakub-agent, line 51: [💡 suggestion] The fallback branch is unconditionally `true`: `route` came from `state.routes.find(...)` a few lines up, so `state.routes.some((item) => item.key === route.key)` always matches. `shouldReset` is therefore always `false` when `history` is absent — which is the intended behavior, but the union-array `.some` hides it, and it typechecks only because `Route` and the history item both happen to have `key`.

Saying it directly reads better and drops the incidental structural overlap:

```ts
// A route already in `state.routes` was reached before, so an absent history is not a first visit.
let shouldReset =
  state.history != null && !state.history.some((item) => item.key === route.key) && !route.state;
```

packages/expo-router/src/react-navigation/routers/DrawerRouter.tsx
```diff
@@ -81,11 +82,22 @@ export function DrawerRouter({
   DrawerNavigationState<ParamListBase>,
   DrawerActionType | CommonNavigationAction
 > {
+  const { backBehavior = 'firstRoute', initialRouteName } = rest;
```
jakub-agent, line 85: [🐧 nit] This restates `TabRouter`'s own default for `backBehavior`. If that default ever changes, the drawer would reconstruct history under one behavior while the router it wraps navigates under another — a quiet divergence, since both still typecheck and most `backBehavior` values produce a plausible-looking history.

Hoisting the default to a shared constant next to the `BackBehavior` type, and using it in both signatures, would keep them tied together.

jakub-agent (approved):

**Review 2** (`33b87598`, force-pushed). All three design findings from the previous review are resolved:

- `getDrawerStatusFromState` no longer throws on an absent `history` — it falls back to `state.default ?? 'closed'`, with a new test file covering both `defaultStatus` values and a history with no drawer entry. This removes the render-path crash.
- `ensureStateHistory` is exported and `DrawerRouter` calls it directly through `ensureDrawerStateHistory`, replacing the `getStateForRouteFocus(state, focusedKey ?? '')` indirection. The cast is documented.
- Reconstruction now keeps an undeclared focused route in history instead of collapsing to `[]`, and preserves its params under `backBehavior: 'fullHistory'`. Both paths have tests.

The `test.each` tables are object tables now, so `$action.type` resolves. Coverage grew to include `defaultStatus: 'open'` and `backBehavior: 'none'`. The description is filled in and the changelog entry is attributed.

I verified locally at `33b87598`, in a worktree:

- `jest` on the three touched test files — **480 passed, 0 failed**.
- Full `expo-router` suite — **5058 passed, 3 failed**. The 3 failures are `native-tabs/__rsc_tests__/tabs.test.tsx` snapshots that embed a repo-relative asset path; they fail because of my worktree location, not this PR.
- `tsc -b tsconfig.all.json` — 31 errors, all from three modules missing in my linked `node_modules` (`react-native-tab-view`, `react-native-pager-view`, `expo/internal/bundle-origin`). No errors in any file this PR touches.

I also checked that `ensureStateHistory` does not reach the public surface: `src/react-navigation/routers/index.tsx` re-exports by name, not `export *`.

**Approving.** Three non-blocking notes below: `getHistoryLength` is the one `history` consumer still reading an un-normalized state, an expression in the headless tabs router that is clearer written plainly, and a duplicated `backBehavior` default.

Two small mismatches with the description, neither worth an inline comment: it says a TODO was left in the headless tabs router to detect a first visit without reading `history`, but no such TODO is in the diff (the only one there is the pre-existing `TODO(@ubax)` about nested trigger href support); and the changelog entry is still missing the trailing period its neighbours carry.

packages/expo-router/src/react-navigation/routers/TabRouter.tsx
Ubax: This is not true. state.routes is the history for as long as no action is dispatched. Then when action is dispatched the history is added, so this check is correct

PR #48716, opened 2026-08-10
fix(ci): Use pnpm global virtual store and prefer offline artifacts in E2E tests

kitten:
# Why

We perform a lot of repeated pnpm installs in our E2E fixtures. A lot of these aren't removable or reducible, but we can try to make them cheaper.

Specifically, using a global virtual store and using `--prefer-offline` to prefer offline artifacts should speed up repeated installs for E2E tests a lot.

# How

- Pass global virtual store option to `pnpm install` helper
- Switch to `--prefer-offline` in helper

# Checklist

### DIFF
--- a/packages/@expo/cli/e2e/__tests__/utils.ts
+++ b/packages/@expo/cli/e2e/__tests__/utils.ts
@@ -164,8 +164,13 @@ export async function createFromFixtureAsync(
       await JsonFile.writeAsync(staticConfigPath, modifiedConfig as any);
     }
 
-    // Install the packages for e2e experience.
-    await executePnpmAsync(projectRoot, ['install']);
+    // Reuse virtual store installs and prefer offline artifacts to speed up repeated installs
+    await executePnpmAsync(projectRoot, ['install', '--prefer-offline'], {
+      env: {
+        NODE_ENV: 'development',
+        npm_config_enable_global_virtual_store: 'true',
+      },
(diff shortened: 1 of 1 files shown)


hassankhan (approved):

packages/@expo/cli/e2e/__tests__/utils.ts
```diff
@@ -164,8 +164,13 @@ export async function createFromFixtureAsync(
       await JsonFile.writeAsync(staticConfigPath, modifiedConfig as any);
     }
 
-    // Install the packages for e2e experience.
-    await executePnpmAsync(projectRoot, ['install']);
+    // Reuse virtual store installs and prefer offline artifacts to speed up repeated installs
+    await executePnpmAsync(projectRoot, ['install', '--prefer-offline'], {
+      env: {
+        NODE_ENV: 'development',
+        npm_config_enable_global_virtual_store: 'true',
```
hassankhan, line 171: I think we'll have to change this slightly for pnpm 11: https://pnpm.io/migration#manual-follow-ups.

kitten: yep! I think we have other cases of `npm_config_*` we'll want to migrate. I think we could make this the default in more cases with pnpm 11 though too, so lots of small improvements we can ship with the upgrade

PR #48717, opened 2026-08-10
ci(cli): Reduce repeated exports in E2E tests

kitten:
# Why

My agent found that we do several `expo export`s repeatedly (each costing just 2s, but this adds up) that are potentially redundant. For example, we export per "serve" target (expo serve, Express, workerd), although these all expect the same export output directory. We also repeat exports on identical fixtures when we could deduplicate.

# How

- Reduce redundant/repeated exports in E2E tests
- Stop repeated exports per "serve target"
- Only assert export monorepo regression once

# Checklist


packages/@expo/cli/e2e/utils/runtime.ts
```diff
@@ -82,19 +82,32 @@ export function prepareServers(
...
+      ['export', '-p', 'web', '--output-dir', sharedOutputName, ...exportCliFlags],
+      { env: defaultExportEnv }
+    ).then(() => sharedOutputDir);
+
+    const sourceDir = await sharedProductionExport;
+    const outputName = generateOutputDir(runtimeName, fixtureName, uniqueOutputKey);
+    const outputDir = path.join(projectRoot, outputName);
+
+    await fs.promises.rm(outputDir, { force: true, recursive: true });
+    await fs.promises.cp(sourceDir, outputDir, { recursive: true });
```
hassankhan, line 102: Maybe a silly question, but wouldn't a symlink be faster?

hassankhan (approved):

PR #48718, opened 2026-08-10
[router] Remove `preloadedRouteKeys`

Ubax:
# Why

In order to be able to create initial state without knowing the unmounted navigator types first, we need to remove the `preloadedRouteKeys` and rework the cases where it is required.

# How

This is mostly implemented already - all the routes in `state.routes` are either preloaded or have been focused.

This PR finishes the work by:
- removing the property from tests
- removing any checks on `preloadedRouteKeys`
- reworking `shouldFreeze` condition in tabs - it should only freeze when route has been focused once. Previously this was checked by checking that it is not included in `preloadedRouteKeys`, but was rendered. Though the latter is enough by itself

# Test Plan

CI

# Checklist


packages/expo-router/src/react-navigation/routers/TabRouter.tsx
```diff
@@ -692,38 +658,48 @@ export function TabRouter({
...
-          const nextId = getId?.({ params: action.payload.params });
-
-          const key = currentId === nextId ? route.key : `${route.name}-${nanoid()}`;
+          let history = state.history;
 
-          const params = createParamsFromAction({ action, routeParamList });
-          const newRoute = params !== route.params ? { ...route, key, params } : route;
+          if (backBehavior === 'history' || backBehavior === 'fullHistory') {
+            if (replacedKey !== undefined) {
+              history = history.filter((record) => record.key !== replacedKey);
```
jakub-agent: [🏗️ **design**] This branch removes the replaced key from history but never adds the new one, so re-keying the **focused** route breaks the invariant that the last route entry in history is the focused route (`changeIndex` maintains it for `history`/`fullHistory`).

Repro: `backBehavior: 'history'`, routes `[A, B]` focused on `B`, history `[A, B]`. Preload `B` with params that change its `getId` → `replacedKey = B_old`, routes become `[A, B_new]`, history becomes `[A]`. A subsequent `goBack()` hits `state.history.length === 1` and returns `null` — you can no longer go back to `A`. With a longer history it is worse: `state.history[length - 2]` now points one step too far back, so focus lands on the wrong route and `history.slice(0, -1)` drops a live entry.

The pre-existing code had the same hole, but the other branch here now rebuilds history properly, so it is worth closing both at once — e.g. map the entry in place (`record.key === replacedKey ? { ...record, key: newKey } : record`) rather than filtering it out.

packages/expo-router/src/react-navigation/routers/TabRouter.tsx
```diff
@@ -68,10 +68,6 @@ export type TabNavigationState<ParamList extends ParamListBase> = Omit<
    * List of previously visited route keys.
    */
   history?: { type: 'route'; key: string; params?: object | undefined }[];
```
jakub-agent, line 69: [🏗️ **design**] No `packages/expo-router/CHANGELOG.md` entry in this PR, while #48709, #48750 and #48752 in the same stack each add one.

It is worth adding here specifically because `preloadedRouteKeys` was not purely internal: it was part of `TabNavigationState`, and `TabsStateContext`'s default value carried it, so anyone reading the tabs state saw the field. That matches the existing "Remove the `type` field from the navigation state" breaking-changes entry, which is the closest precedent.

packages/expo-router/src/react-navigation/bottom-tabs/views/BottomTabView.tsx
```diff
@@ -279,7 +272,8 @@ export function BottomTabView(props: Props) {
               active={activityState}
               enabled={detachInactiveScreens}
               freezeOnBlur={freezeOnBlur}
-              shouldFreeze={activityState === STATE_INACTIVE && !isPreloaded}>
+              // TODO: A visited blurred tab re-preloaded with new params stays frozen until focused.
+              shouldFreeze={activityState === STATE_INACTIVE && loaded.includes(route.key)}>
```
jakub-agent, line 275: [💡 suggestion] Freezing now keys off `loaded`, which is component-local state seeded with the focused key at mount and appended on focus — so it means "focused at some point during this mount", not "materialized".

That changes one case beyond the one your TODO records: routes that come back from `getRehydratedState` (persisted state, or a deep link that restores several tab routes) were previously frozen while blurred (`!isPreloaded` was true for them), and are now never frozen for the lifetime of the mount. Their timers and subscriptions keep running. Deriving this from router state — as `descriptor.route.key !== undefined` combined with focus — would survive remounts, which `loaded` does not.

Minor, while you are here: the doc comment on `loaded` above ("List of loaded tabs, tabs will be loaded when navigated to") no longer describes what it does — after this PR it only drives freezing. Same for `DrawerView.tsx`.

packages/expo-router/src/react-navigation/routers/__tests__/TabRouter.test.tsx
```diff
@@ -351,7 +341,29 @@ test('navigation removes a stale preload key when the route ID changes', () => {
   )!;
 
   expect(result.routes[result.index!]!.key).toBe('baz-test');
-  expect(result.preloadedRouteKeys).toEqual([]);
+});
+
+test('PRELOAD rebuilds history when re-keying the focused route', () => {
```
jakub-agent, line 348: [💡 suggestion] This covers the default `backBehavior` (`firstRoute`), which takes the rebuild branch. The `history` / `fullHistory` branch of the same `PRELOAD` block is the one that is actually broken (see the comment on `TabRouter.tsx`), and nothing exercises it.

A `test.each<['history' | 'fullHistory']>` variant of this case — preload the focused route with a new id, then assert history still ends at the focused route and `goBack()` still reaches the previous one — would pin the fix and stop the branch from silently regressing in later steps of the stack.

jakub-agent (commented):

PR 6 in the router refactor stack (base: `@ubax/make-state-history-optional` / #48709).

The direction is right. `preloadedRouteKeys` was a second source of truth next to `routes`, and the placeholder model already encodes "materialized vs. not" through `descriptor.route.key === undefined` plus `usePreloadPlaceholderRoutes`. Deriving freeze/render decisions from that instead of a parallel key list removes the duplication. Replacing the `targetIsPreloaded` special case in `GO_BACK` with a history rebuild inside `PRELOAD` is the correct swap: for `firstRoute` / `initialRoute` / `order`, history is fully derived from ordered routes anyway, so `getRouteHistory` reproduces the old focus target. The `MaterialTopTabView` `lazy` predicate change is also a small improvement — a `lazy: false` route no longer renders an empty placeholder scene on the first frame before the preload effect runs.

Two things to look at:

1. The `history` / `fullHistory` branch of `PRELOAD` still drops the focused route's key from history without re-adding the new one. That is pre-existing, but this PR rewrites the surrounding block and the new test only covers the default `backBehavior`, so the gap is now easy to miss. It breaks `goBack`.
2. No `packages/expo-router/CHANGELOG.md` entry. #48709, #48750 and #48752 in this stack all add one, and this removes a field that was reachable through `TabsStateContext`.

Also: the PR body is the unfilled template. A one-line Why/Test Plan would help reviewers of the other 6 PRs in the stack place this one.

Verified: no `preloadedRouteKeys` references remain in `expo-router/src` (the occurrences in `stack/views/Stack/CardStack.tsx` are an unrelated local variable). The new `DrawerRouter` test expectation matches what `getRouteHistory` produces for `backBehavior: 'order'` with the drawer entry kept last.

packages/expo-router/src/react-navigation/routers/TabRouter.tsx
```diff
@@ -688,38 +654,67 @@ export function TabRouter({
...
+                  : record
+              );
+            }
+          } else {
+            const orderedRoutes = orderRoutesByRouteNames(routes, state.routeNames);
+            const focusedKey = routes[state.index]?.key;
+            const focusedIndex = orderedRoutes.findIndex((route) => route.key === focusedKey);
+            const routeHistory =
+              focusedIndex === -1
+                ? []
```
jakub-agent, line 707: [💡 suggestion] Non-blocking, and the contrast with the sibling fix is what makes it worth noting.

`orderRoutesByRouteNames` drops routes whose name is not in `routeNames`, so `focusedIndex` can be `-1` for a focused route that is present in `state.routes` but undeclared — the case `ensureStateHistory` was just changed to handle ("Keep it in history anyway, so the current route is always the last entry"). Here the same situation falls back to `[]`, so `history` ends up holding only the non-route entries and loses the focused route: the invariant the `history`/`fullHistory` branch above now protects.

The `-1` guard itself is needed — `getRouteHistory(routes, -1, ...)` would dereference `routes[-1]!`. It is the fallback value that could match `ensureStateHistory`: `[{ type: TYPE_ROUTE, key: focusedKey }]` when `focusedKey !== undefined`, `[]` otherwise.

Happy for this to fold into the `// TODO: Refactor history handling together with web state synchronization.` below it rather than be fixed here.

jakub-agent (approved):

Iteration 2, on `aa5de4b` (rebased onto the updated base). All three points from the previous review are addressed:

- **`PRELOAD` under `backBehavior: 'history' | 'fullHistory'`** now re-keys the history entry in place instead of filtering it out, so the focused route stays the last entry and `goBack` reads the right target. The `fullHistory` handling is a nicer answer than I suggested: only the newest entry takes the new params, so each older duplicate keeps the params of its own visit. Guarding on `record.type === TYPE_ROUTE` keeps the drawer entry intact, and the block is unreachable when `routeIndex === -1`, so `routes[routeIndex]!` is safe.
- **Tests** — the new `test.each<['history' | 'fullHistory']>` case asserts the invariant *and* that `goBack()` lands on index 0, and `fullHistory back behavior keeps earlier params snapshots when PRELOAD re-keys a route` pins the duplicate-entry semantics. That is exactly the coverage the branch was missing.
- **CHANGELOG** entry added under breaking changes, consistent with the neighbouring stack entries.

On the `shouldFreeze` suggestion: keeping component-local `loaded` is a fair call, and the replacement comments on both views now state the rule ("never frozen before it has rendered once") rather than describing the old lazy-loading role. Documented intent is enough here.

The rest of the diff is fixture cleanup, and I re-checked that no `preloadedRouteKeys` reference remains in `expo-router/src` (the ones in `stack/views/Stack/CardStack.tsx` are an unrelated local variable).

One non-blocking edge case left inline. I reviewed the new tests but did not execute the suite — worth confirming CI is green before merge, and the PR body is still the unfilled template.

PR #48723, opened 2026-08-10
[android][local-authentication] Fix crash when using strong biometric security level with device fallback on Android 9 and 10

sbaiahmed1:
# Why

Calling `LocalAuthentication.authenticateAsync({ biometricsSecurityLevel: 'strong' })` crashes the app on Android 9 and 10 (API 28–29) when the device credential fallback is enabled (`disableDeviceFallback: false`, the default).

The module requests `BIOMETRIC_STRONG | DEVICE_CREDENTIAL`, but `androidx.biometric`'s `PromptInfo.Builder.build()` rejects that combination on API 28–29 and throws:

```
java.lang.IllegalArgumentException: Authenticator combination is unsupported on API 29: BIOMETRIC_STRONG | DEVICE_CREDENTIAL
```

The module only catches `NullPointerException` around `BiometricPrompt.authenticate()`, so the exception propagates and hard-crashes the app. See [`AuthenticatorUtils.isSupportedCombination`](https://android.googlesource.com/platform/frameworks/support/+/refs/heads/androidx-main/biometric/biometric/src/main/java/androidx/biometric/AuthenticatorUtils.java) — `BIOMETRIC_STRONG | DEVICE_CREDENTIAL` is only supported on API < 28 or API 30+.

# How

Added an automatic fallback: on API 28–29, when `strong` is requested and the device credential fallback is enabled, the module now requests `BIOMETRIC_WEAK | DEVICE_CREDENTIAL` (a combination supported on those API levels) instead of crashing, and the successful authentication result carries a `warning` string reporting the downgrade.

Notes on the approach:

- `BIOMETRIC_WEAK` means "Class 2 *or better*", so devices with Class 3 sensors (fingerprint on most Android 9/10 devices) behave exactly as before — only Class-2-only devices gain a biometric option where they previously crashed.
- The downgrade only applies when the device credential fallback is enabled — in that configuration the app already accepts PIN/pattern/password, so the effective security floor is already the lock screen. With `disableDeviceFallback: true` there is no downgrade (strong-only never crashed, and the strict Class 3 contract stays intact).
- The authenticator selection was extracted into a pure `selectAuthenticators()` function so it can be unit-tested, and the success result type gained an optional `warning` field (previously only failure results could carry one).

# Test Plan

- Added Android unit tests (`AuthenticatorSelectionTest.kt`, the package's first) covering the authenticator selection for weak/strong × fallback enabled/disabled across API 27, 28, 29, 30, and 34. Written red-first: the API 28–29 case failed against the original logic and passes with the fix. Run with `et native-unit-tests -p android -t local --packages expo-local-authentication`.
- `et check-packages expo-local-authentication` passes (build, typecheck, JS unit tests, lint, format).

# Checklist

### DIFF
--- a/packages/expo-local-authentication/CHANGELOG.md
+++ b/packages/expo-local-authentication/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### 🐛 Bug fixes
 
-- [Android] Fix `authenticateAsync` crashing on Android 9 and 10 when `biometricsSecurityLevel` is set to `'strong'` and the device credential fallback is enabled. On those Android versions the security level now automatically falls back to `'weak'` (the system cannot combine strong biometrics with device credentials in a single prompt there), and the successful result contains a `warning` reporting the downgrade. ([#TBD](https://github.com/expo/expo/pull/TBD) by [@sbaiahmed1](https://github.com/sbaiahmed1))
+- [Android] Fix `authenticateAsync` crashing on Android 9 and 10 when `biometricsSecurityLevel` is set to `'strong'` and the device credential fallback is enabled. On those Android versions the security level now automatically falls back to `'weak'` (the system cannot combine strong biometrics with device credentials in a single prompt there), and the successful result contains a `warning` reporting the downgrade. ([#48723](https://github.com/expo/expo/pull/48723) by [@sbaiahmed1](https://github.com/sbaiahmed1))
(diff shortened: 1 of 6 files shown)


sbaiahmed1:
Hey @behenate can you take a look when you have time? Thank you

PR #48731, opened 2026-08-10
[router] Remove nested screen params

Ubax:
# Why

# How

# Test Plan

# Checklist


packages/expo-router/src/react-navigation/core/index.tsx
```diff
@@ -16,7 +16,6 @@ export { findFocusedRoute } from './findFocusedRoute';
 /**
  * @deprecated Will be removed in a future SDK.
  */
```
Ubax, line 18: Left over

PR #48735, opened 2026-08-10
[ios][ui] Fix empty RNHostView taking up space in SwiftUI containers

### LINKED ISSUE #48666: React-Native Modal causes universal components to vanish/shift out of view

https://github.com/lochnessopensource/form-modal-not-working

### Steps to reproduce

In my real app, I have a reusable form input that can show a modal containing an image (it’s a file upload component). For some reason, when the React Native modal is tucked away inside a custom React component, it causes the universal components in the form to disappear (or shift out of view where they can't be seen). As can be seen from the gif, it works fine and expected when the reusable react component isn't utilized even though the same code is simply used inside.

Image

### Environment

```text
    System:
      OS: Windows 11 10.0.26200
    Binaries:
      Node: 24.15.0 - C:\Program Files\nodejs\node.EXE
      Yarn: 1.22.22 - C:\Program Files\nodejs\yarn.CMD
      npm: 11.12.1 - C:\Program Files\nodejs\npm.CMD
    npmPackages:
      expo: ~57.0.11 => 57.0.11
      expo-router: ~57.0.11 => 57.0.11
      react: 19.2.3 => 19.2.3
      react-dom: 19.2.3 => 19.2.3
    Expo Workflow: bare
```

### Expo Doctor Diagnostics

```text
```

brentvatne:
# Why

Fixes #48666.

An `RNHostView` with no mounted child renders as a spurious empty row in a `FieldGroup`/`Form` section. The empty row takes up space and shifts the surrounding universal components. The most common trigger is hosting an RN `Modal`: `Modal` renders `null` until `visible` is `true`, so the host has no child while the modal is closed.

# How

With `matchContents: false` and no children, `RNHostView` rendered `Children()` (an empty `ForEach`) wrapped in `.frame(maxWidth: .infinity, maxHeight: .infinity)`. A frame-wrapped empty `ForEach` is a real view, so SwiftUI materializes it as a `Form` row.

`RNHostView` now renders `EmptyView` until a child is mounted. `EmptyView` resolves to no views, so containers like `Section` produce no row. A side benefit: the fill branch's `onAppear` now fires when the first child mounts, so the `RCTSurfaceTouchHandler` attaches even when the hosted RN view mounts after the host appeared (before, it never attached in that case).

# Test Plan

No native test target exists for `expo-ui` iOS, so I verified with the issue's repro app (https://github.com/lochnessopensource/form-modal-not-working, expo 57.0.11 / `@expo/ui` 57.0.9 / RN 0.86.2) on an iOS simulator via an Expo sandbox:

- **Baseline** EAS development build (unpatched `@expo/ui` 57.0.9): the section shows a spurious empty row above the two `Text` rows, matching the issue report.
- **Patched** EAS development build (this diff applied via patch-package): the empty row is gone and the section shows only the two `Text` rows, matching the expected layout.
- On the patched build, both `Modal`s (one hosted in an `RNHostView` directly in the section, one inside a `Row`) still present and dismiss correctly, via tap and via programmatic `visible` toggling, and all universal components stay visible and intact afterwards.

# Checklist

https://claude.ai/code/session_01RzDhs55jrcb98BGt2yTEQU

---
_Generated by [Claude Code](https://claude.ai/code/session_01RzDhs55jrcb98BGt2yTEQU)_


packages/expo-ui/ios/RNHostView.swift
```diff
@@ -30,6 +30,11 @@ struct RNHostView: ExpoSwiftUI.View {
       // natural size instead of falling into the fill branch below, which would
       // stretch a self-sizing SwiftUI view to fill its container.
       Children()
+    } else if props.children?.isEmpty != false {
```
Kudo: [🐧 nit] `!= false` on an optional `Bool` is a double negative that takes a second read. `?? true` says the same thing directly:

```suggestion
    } else if props.children?.isEmpty ?? true {
```

The package's own idiom for this is a named `hasChildren` property (`DatePickerView.swift:39`, `ShareLinkView.swift:44`), which would also fit next to `firstChildUIView` in this file if you prefer that shape.

Kudo (approved):

intergalacticspacehighway (approved):

brentvatne:
@expo-bot address feedback

PR #48741, opened 2026-08-10
fix(cli): Set NODE_ENV before loading env files

ramonclaudio:
# Why

Expo CLI doesn't always set `NODE_ENV` before loading the app config or env files.

# How

I updated Expo CLI commands to use the new shared API and set the mode to either `development` or `production`.

# Test Plan

Tests and CI checks pass.

# Checklist

### DIFF
--- a/packages/@expo/cli/src/config/index.ts
+++ b/packages/@expo/cli/src/config/index.ts
@@ -34,17 +34,20 @@ export const expoConfig: Command = async (argv) => {
   }
 
   // Load modules after the help prompt so `npx expo config -h` shows as fast as possible.
-  const [
-    // ./configAsync
-    { configAsync },
-    // ../utils/errors
-    { logCmdError },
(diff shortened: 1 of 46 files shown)


packages/@expo/cli/src/config/index.ts
```diff
@@ -34,17 +34,20 @@ export const expoConfig: Command = async (argv) => {
...
-  const [
-    // ./configAsync
-    { configAsync },
-    // ../utils/errors
-    { logCmdError },
-  ] = await Promise.all([import('./configAsync.js'), import('../utils/errors.js')]);
+  const [{ getConfigEnvMode, loadEnvFiles }, { logCmdError }] = await Promise.all([
+    import('../utils/nodeEnv.js'),
+    import('../utils/errors.js'),
+  ]);
```
kitten: nit: let's just import these normally. The laziness won't help here much and we may want to do this more rarely than we've done before (future refactors possible, but not a priority)

kitten (approved):

PR #48750, opened 2026-08-11
[router] Remove the `default` field from `DrawerNavigationState` and deprecate `getDrawerStatusFromState`

Ubax:
# Why

`default` drawer status (wether it is opened or not) is stored in state. In my opinion this is only navigator's concern, and not relevant from the navigation state perspective.

# How

1. Remove `default` from `DrawerNavigationState`
2. Add second parameter to `getDrawerStatusFromState` - `defaultState` which is used when no open/closed information is stored in the `history`
3. Deprecate `getDrawerStatusFromState` - the recommended solution is using hook, which can only be accessed from within the navigator content and thus get the status directly from the navigator. In the future we may consider dropping the whole open/closed information from state and making it available only via the hook/context.

# Test Plan

CI

# Checklist

### DIFF
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### 🛠 Breaking changes
 
+- Remove the static `default` field from `DrawerNavigationState` and deprecate `getDrawerStatusFromState`. ([#48750](https://github.com/expo/expo/pull/48750) by [@Ubax](https://github.com/Ubax))
 - Remove the `preloadedRouteKeys` field from `TabNavigationState`. ([#48718](https://github.com/expo/expo/pull/48718) by [@Ubax](https://github.com/Ubax))
 - Make `history` optional in `TabNavigationState` and `DrawerNavigationState` ([#48709](https://github.com/expo/expo/pull/48709) by [@Ubax](https://github.com/Ubax))
 - Make navigation state `type` optional for custom routers. ([#48757](https://github.com/expo/expo/pull/48757) by [@Ubax](https://github.com/Ubax))
--- a/packages/expo-router/src/react-navigation/drawer/__tests__/getDrawerStatusFromState.test.ts
+++ b/packages/expo-router/src/react-navigation/drawer/__tests__/getDrawerStatusFromState.test.ts
@@ -1,42 +1,57 @@
-import { expect, test } from '@jest/globals';
-
-import type { DrawerNavigationState, DrawerStatus, ParamListBase } from '../../native';
+import { DrawerRouter } from '../../routers';
 import { getDrawerStatusFromState } from '../utils/getDrawerStatusFromState';
 
-const createState = (
-  history: DrawerNavigationState<ParamListBase>['history'],
-  defaultStatus: DrawerStatus = 'closed'
-): DrawerNavigationState<ParamListBase> => ({
-  stale: false,
-  type: 'drawer',
-  key: 'drawer-test',
-  index: 0,
-  routeNames: ['bar'],
-  routes: [{ key: 'bar', name: 'bar' }],
-  default: defaultStatus,
-  history,
+const state = DrawerRouter({}).getInitialState({
+  routeNames: ['index'],
+  routeParamList: {},
+  routeGetIdList: {},
 });
 
-test.each<{ defaultStatus: DrawerStatus }>([
-  { defaultStatus: 'closed' },
-  { defaultStatus: 'open' },
-])('falls back to defaultStatus: $defaultStatus without history', ({ defaultStatus }) => {
-  expect(getDrawerStatusFromState(createState(undefined, defaultStatus))).toBe(defaultStatus);
+const openState = DrawerRouter({ defaultStatus: 'open' }).getInitialState({
+  routeNames: ['index'],
+  routeParamList: {},
+  routeGetIdList: {},
 });
 
-test('reads the status from the last drawer entry', () => {
+it.each(['closed', 'open'] as const)(
+  'uses the provided default status %s when history has no drawer entry',
+  (defaultStatus) => {
+    expect(getDrawerStatusFromState(state, defaultStatus)).toBe(defaultStatus);
+  }
+);
+
+it('reports open for a router with defaultStatus open and no drawer entry', () => {
+  expect(getDrawerStatusFromState(openState, 'open')).toBe('open');
+});
+
+it('reports closed for a router with defaultStatus open after the drawer is closed', () => {
   expect(
     getDrawerStatusFromState(
-      createState([
-        { type: 'route', key: 'bar' },
-        { type: 'drawer', status: 'open' },
-      ])
+      {
+        ...openState,
+        history: [...(openState.history ?? []), { type: 'drawer', status: 'closed' }],
+      },
+      'open'
     )
-  ).toBe('open');
+  ).toBe('closed');
 });
 
-test('falls back to defaultStatus when history has no drawer entry', () => {
-  expect(getDrawerStatusFromState(createState([{ type: 'route', key: 'bar' }], 'open'))).toBe(
-    'open'
-  );
+it('uses the last drawer status from history instead of the provided default', () => {
+  expect(
+    getDrawerStatusFromState(
+      {
+        ...state,
+        history: [
+          ...(state.history ?? []),
+          { type: 'drawer', status: 'open' },
+          { type: 'drawer', status: 'closed' },
+        ],
+      },
+      'open'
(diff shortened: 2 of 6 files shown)


packages/expo-router/src/react-navigation/drawer/utils/getDrawerStatusFromState.tsx
```diff
@@ -1,17 +1,18 @@
...
+/**
+ * Returns the drawer's current status.
+ *
+ * @param defaultStatus Status returned when the state has no drawer history entry.
+ * @deprecated Use `useDrawerStatus` instead.
+ */
 export function getDrawerStatusFromState(
-  state: DrawerNavigationState<ParamListBase>
+  state: DrawerNavigationState<ParamListBase>,
+  defaultStatus: DrawerStatus = 'closed'
```
jakub-agent: [🚨 **critical**] `getDrawerStatusFromState` is re-exported publicly from `expo-router/drawer` (`src/layouts/Drawer.tsx`), so external single-arg callers still compile after this change — but they now get a wrong answer whenever the navigator uses `defaultStatus: 'open'`.

The drawer history entry encodes a toggle relative to `defaultStatus`: `DrawerRouter.addDrawerToHistory` pushes `status: defaultStatus === 'open' ? 'closed' : 'open'`, so an open drawer with `defaultStatus: 'open'` has *no* drawer entry at all. With the `'closed'` fallback baked in here, `getDrawerStatusFromState(state)` returns `'closed'` when the drawer is open, and `'closed'` when it is closed — it can never report `'open'` for that configuration. Previously `state.default` carried the value, so the one-arg call was correct.

Suggest making `defaultStatus` a required parameter so the migration surfaces as a type error rather than a silent behavioral inversion. If the intent is that nobody should call this anymore, removing the export from `src/layouts/Drawer.tsx` and `src/react-navigation/drawer/index.tsx` in the same breaking-change stack is cleaner than leaving a function that quietly lies.

packages/expo-router/src/react-navigation/drawer/utils/getDrawerStatusFromState.tsx
```diff
@@ -1,17 +1,18 @@
 import type { DrawerNavigationState, DrawerStatus, ParamListBase } from '../../native';
 
+/**
+ * Returns the drawer's current status.
+ *
+ * @param defaultStatus Status returned when the state has no drawer history entry.
+ * @deprecated Use `useDrawerStatus` instead.
```
jakub-agent, line 7: [💡 suggestion] `useDrawerStatus` is not a drop-in replacement for every use of this function. It reads `DrawerStatusContext`, so it only works inside the drawer's React subtree and only during render. The typical reason to reach for `getDrawerStatusFromState` is the opposite case — deriving status from a state object outside the drawer, for example in a `state` listener or a `navigationRef` callback. Worth saying what those callers should do instead, otherwise the deprecation has no migration path for them.

packages/expo-router/src/react-navigation/drawer/__tests__/getDrawerStatusFromState.test.ts
```diff
@@ -0,0 +1,36 @@
+import { DrawerRouter } from '../../routers';
+import { getDrawerStatusFromState } from '../utils/getDrawerStatusFromState';
+
+const state = DrawerRouter({}).getInitialState({
```
jakub-agent, line 4: [💡 suggestion] All four cases use a router built with the default `defaultStatus: 'closed'`, where the parameter's fallback happens to match. The interesting case is `DrawerRouter({ defaultStatus: 'open' })`: state with no drawer entry means the drawer is *open*, and a drawer entry means it is closed. Adding that case would lock the inverted-history semantics and would fail on the silent inversion described in the comment on `getDrawerStatusFromState.tsx` — which is the behavior most likely to break for existing callers.

packages/expo-router/src/react-navigation/drawer/utils/getDrawerStatusFromState.tsx
```diff
@@ -1,17 +1,18 @@
...
+  defaultStatus: DrawerStatus = 'closed'
 ): DrawerStatus {
-  if (state.history == null) {
-    throw new Error(
-      "Couldn't find the drawer status in the state object. Is it a valid state object of drawer navigator?"
-    );
-  }
-
-  const entry = state.history.findLast((it) => it.type === 'drawer') as
+  const entry = (state.history ?? []).findLast((it) => it.type === 'drawer') as
```
jakub-agent: [🐧 nit] `state.history?.findLast(...)` avoids allocating a throwaway array on every render — `findLast` on `undefined` is short-circuited by the optional chain and the result is already typed as possibly `undefined`.

packages/expo-router/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 - Remove the `initialRouteName` prop from Expo Router navigators.
 - Handle `PUSH` in tab and drawer routers instead of coercing it to `NAVIGATE`.
 - Make `history` optional in `TabNavigationState` and `DrawerNavigationState`
+- Remove the static `default` field from `DrawerNavigationState` and deprecate `getDrawerStatusFromState`.
```
jakub-agent: [🐧 nit] Missing the PR link and author attribution that CONTRIBUTING asks for — `([#48750](https://github.com/expo/expo/pull/48750) by [@Ubax](https://github.com/Ubax))`. A few other entries at the top of this section are missing it too, presumably from the same stack.

jakub-agent (changes requested):

Moving the drawer default out of the navigation state is the right direction — `default` was a static config value duplicated into every state object, and `DrawerView` already has `defaultStatus` in scope. The router change and the test updates are mechanical and correct.

One real problem: `getDrawerStatusFromState` is still publicly re-exported from `expo-router/drawer`, and the new `defaultStatus` parameter is optional with a `'closed'` fallback. For a drawer configured with `defaultStatus: 'open'`, an existing single-arg call now returns `'closed'` unconditionally — it used to read `state.default` and answer correctly. That is a silent wrong answer, not a compile error, so callers will not notice the migration. Making the parameter required (or dropping the export, since it is deprecated anyway) turns it into a type error instead.

This PR sits on top of #48718 in the stack, which makes `history` optional — the removal of the `state.history == null` throw here depends on that.

Minor: the changelog entry has no PR link or author attribution, unlike the older entries in the same section.

packages/expo-router/src/react-navigation/drawer/utils/getDrawerStatusFromState.tsx
Ubax: Make it required

packages/expo-router/src/react-navigation/drawer/utils/getDrawerStatusFromState.tsx
Ubax: This should be enough. The function is not removed so this is not a big deal

packages/expo-router/src/react-navigation/drawer/utils/getDrawerStatusFromState.tsx
Ubax: Address

packages/expo-router/CHANGELOG.md
Ubax: Add link and move as first in section. Not mid section

jakub-agent (approved):

**Iteration 2** — re-reviewed at `2a5d8d4`, now rebased directly onto `main` rather than stacked on #48718.

All findings from the previous round are addressed:

- **critical** — `defaultStatus` is now a required parameter. A single-arg call is a type error instead of a silent `'closed'` for `defaultStatus: 'open'` drawers. `DrawerView` is the only in-repo call site and passes the value.
- **suggestion (tests)** — added `openState` from `DrawerRouter({ defaultStatus: 'open' })` with both the no-drawer-entry (`'open'`) and closed (`'closed'`) cases, which locks the inverted-history semantics.
- **suggestion (deprecation note)** — acknowledged and declined; reasonable, since the function stays exported and callers keep a working path.
- **nit** — `state.history?.findLast(...)`.
- **nit** — changelog entry now has the PR link and sits at the top of the section.

The `DrawerRouter` change and the mechanical `default: 'closed'` removals across `DrawerRouter.test.tsx` are correct. CI is green, including `jest-ubuntu` and `check-packages`.

Optional, non-blocking: the `defaultStatus: 'open'` closed-drawer test builds its history literal by hand rather than dispatching `DrawerActions.closeDrawer()` through the router. Going through the router would keep the test honest if the toggle encoding in `addDrawerToHistory` ever changes.

Approving.

PR #48752, opened 2026-08-11
[router] Handle PUSH action in all routers

Ubax:
# Why

Instead of rewriting the navigation action from `PUSH` to `NAVIGATE` or `JUMP_TO`, we can handle this actions in routers. This makes the responsibilities clearer and removes the dependency of `getNavigateAction` on `state.type` - removed in a follow-up PR

```tsx
  if (type === 'PUSH' && navigationState.type !== 'stack') {
    type = 'NAVIGATE';
  } else if (navigationState.type === 'expo-tab') {
    type = 'JUMP_TO';
  }
```

# How

1. Add `PUSH` cases to `TabRouter` and `BaseRouter`
2. Update tests
3. Remove the condition

# Test Plan

CI

# Checklist

### DIFF
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### 🛠 Breaking changes
 
+- Handle `PUSH` in tab and drawer routers instead of coercing it to `NAVIGATE`. Custom routers must now handle the `PUSH` action in `getStateForAction`. ([#48752](https://github.com/expo/expo/pull/48752) by [@Ubax](https://github.com/Ubax))
 - Remove the `initialParams` prop from Expo Router screens and `routeParamList` from custom router action options. ([#48756](https://github.com/expo/expo/pull/48756) by [@Ubax](https://github.com/Ubax))
 - Remove the `initialRouteName` prop from Expo Router navigators. Configure the initial route with `unstable_settings.initialRouteName` in the route layout instead. ([#48708](https://github.com/expo/expo/pull/48708) by [@Ubax](https://github.com/Ubax))
 - Remove `NavigationContainerRefContext` fallback in `useNavigation`. The hook now throws when called outside a navigator, including components rendered via `ExpoRoot`'s `wrapper` prop. ([#48638](https://github.com/expo/expo/pull/48638) by [@Ubax](https://github.com/Ubax))
--- a/packages/expo-router/src/__tests__/tab-router-push.test.ios.tsx
+++ b/packages/expo-router/src/__tests__/tab-router-push.test.ios.tsx
@@ -1,0 +1,247 @@
+import { act, fireEvent, screen } from '@testing-library/react-native';
+import { Button, Text, View } from 'react-native';
+import { Drawer as DrawerLayout } from 'react-native-drawer-layout';
+
+import { router } from '../imperative-api';
+import { Drawer } from '../layouts/Drawer';
+import { Stack } from '../layouts/Stack';
+import { Tabs } from '../layouts/Tabs';
+import { NativeTabs } from '../native-tabs/NativeTabs';
+import { INTERNAL_EXPO_ROUTER_ZOOM_TRANSITION_SOURCE_ID_PARAM_NAME } from '../navigationParams';
+import { DrawerActions } from '../react-navigation/native';
+import { renderRouter } from '../testing-library';
+import { TabList, TabSlot, TabTrigger, Tabs as HeadlessTabs } from '../ui';
+import { useNavigation } from '../useNavigation';
+
+jest.mock('react-native-drawer-layout', () => {
+  const { View }: typeof import('react-native') = jest.requireActual('react-native');
+  const actual = jest.requireActual(
+    'react-native-drawer-layout'
+  ) as typeof import('react-native-drawer-layout');
+  return {
+    ...actual,
+    Drawer: jest.fn(({ children, ...props }) => (
+      <View testID="drawer" {...props}>
+        {children}
+      </View>
+    )),
+  };
+});
+
+jest.mock('react-native-screens', () => {
+  const { View }: typeof import('react-native') = jest.requireActual('react-native');
+  const actual = jest.requireActual(
+    'react-native-screens'
+  ) as typeof import('react-native-screens');
+  return {
+    ...actual,
+    ScreenStackItem: jest.fn(({ children }) => <View>{children}</View>),
+    Tabs: {
+      ...actual.Tabs,
+      Host: jest.fn(({ children }) => <View>{children}</View>),
+      Screen: jest.fn(({ children }) => <View>{children}</View>),
+    },
+  };
+});
+
+const drawerOpen = () =>
+  (DrawerLayout as unknown as jest.Mock).mock.calls.at(-1)![0].open as boolean;
+
+let warn: jest.SpyInstance | undefined;
+
+afterEach(() => {
+  warn?.mockRestore();
+  warn = undefined;
+});
+
+it('push switches JS tabs without duplicating routes and follows back behavior', () => {
+  const result = renderRouter({
+    _layout: () => (
+      <Tabs backBehavior="history">
+        <Tabs.Screen name="index" />
+        <Tabs.Screen name="second" />
+        <Tabs.Screen name="third" />
+      </Tabs>
+    ),
+    index: () => null,
+    second: () => null,
+    third: () => null,
+  });
+
+  act(() => router.push('/second'));
+  act(() => router.push('/third'));
+
+  const tabState = result.getRouterState()!.routes[0]!.state;
+  expect(tabState?.routes).toHaveLength(3);
+  expect(screen).toHavePathname('/third');
+
+  act(() => router.push('/third'));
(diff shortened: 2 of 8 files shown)


packages/expo-router/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Remove the `initialParams` prop from Expo Router screens.
 - Remove the `initialRouteName` prop from Expo Router navigators.
+- Handle `PUSH` in tab and drawer routers instead of coercing it to `NAVIGATE`.
```
jakub-agent: [🏗️ **design**] This entry describes the internal change but not the contract change it creates for consumers.

Before this PR, `getNavigateAction` coerced `PUSH` to `NAVIGATE` for every non-stack navigator, so a router only had to handle `NAVIGATE` to work with `router.push`. Now `PUSH` reaches the target router unchanged. A custom router integrated via `unstable_integrateWithRouter` / `withLayoutContext` that has no `PUSH` case returns `null`, the action bubbles up unhandled, and `onUnhandledAction` throws in tests, logs an error in dev, and does nothing in production.

The `getStateForRouteNamesChange` entry a few lines below states exactly this kind of requirement ("Custom routers must handle the `ROUTE_NAMES_CHANGED` action in `getStateForAction` instead"). Worth the same treatment here, plus the `([#48752](https://github.com/expo/expo/pull/48752) by [@Ubax](https://github.com/Ubax))` attribution the guide asks for.

packages/expo-router/src/react-navigation/routers/TabRouter.tsx
```diff
@@ -349,6 +355,14 @@ export function TabRouter({
       return changeIndex(state, index, backBehavior, initialRouteName);
     },
 
+    shouldActionChangeFocus(action) {
```
jakub-agent: [🏗️ **design**] This override is needed — without it the removed coercion would have silently broken parent focus propagation — but it leaves `StackRouter` behind. `StackRouter` has no `shouldActionChangeFocus` of its own, so it falls back to `BaseRouter`'s (`NAVIGATE || NAVIGATE_DEPRECATED`).

Result: with a drawer open, `router.push('/screen/detail')` where the divergent navigator is a nested stack does not fire `onRouteFocus` on the drawer, so the drawer stays open — while the third test in this PR asserts that the same `router.push` closes the drawer when the divergent navigator is a tab. Same user action, two different outcomes depending on the target navigator type. (The stack half is pre-existing, but this PR is what makes the pair inconsistent.)

Consider adding `PUSH` to `BaseRouter.shouldActionChangeFocus` instead of overriding here: it covers stack, tab and drawer in one place, and avoids duplicating the `NAVIGATE` / `NAVIGATE_DEPRECATED` list that will now drift if `BaseRouter` changes. If stacks are intentionally excluded, a short comment explaining why would help.

packages/expo-router/src/__tests__/tab-router-push.test.ios.tsx
```diff
@@ -0,0 +1,200 @@
...
+    index: () => null,
+    second: () => null,
+    third: () => null,
+  });
+
+  act(() => router.push('/second'));
+  act(() => router.push('/third'));
+
+  const tabState = result.getRouterState()!.routes[0]!.state;
+  expect(tabState?.routes).toHaveLength(3);
```
jakub-agent, line 75: [💡 suggestion] Inside `TabRouter.getStateForAction`, `PUSH` falls through to the exact same case as `NAVIGATE`, so `push` to a tab is now defined to switch rather than duplicate. That is the one place where tab `PUSH` semantics deliberately differ from stack `PUSH`, and nothing here pins it down.

Suggest adding `act(() => router.push('/third'))` a second time and asserting `routes` still has length 3 and the index is unchanged. That documents the intent and would catch a future change that makes `PUSH` create a second route entry.

packages/expo-router/src/__tests__/tab-router-push.test.ios.tsx
```diff
@@ -0,0 +1,200 @@
...
+    ),
+    index: () => null,
+    'fruit/_layout': () => <Stack />,
+    'fruit/index': () => <Text>Fruit</Text>,
+    'fruit/details': () => <Text>Details</Text>,
+  });
+
+  act(() => router.push('/fruit/details'));
+
+  expect(screen).toHavePathname('/fruit/details');
```
jakub-agent, line 213: [💡 suggestion] This asserts only the resulting pathname, which `JUMP_TO` and `NAVIGATE` would also satisfy. The interesting part of pushing into a not-yet-visited tab is what happens to the target tab's nested stack — whether the anchor (`fruit/index`) is created below `fruit/details`, which is driven by the `initial` param handling in `getNavigateAction` that this PR now reaches with `type === 'PUSH'`.

Worth asserting the nested stack's routes (`['index', 'details']`) so a regression in anchor handling for `PUSH` is caught, not just the final URL.

jakub-agent (commented):

Dropping the global `PUSH` → `NAVIGATE` coercion from `getNavigateAction` and handling `PUSH` in `TabRouter` / `DrawerRouter` / `NativeBottomTabsRouter` is the right direction. The removed `expo-tab` branch was already dead — nothing sets `type: 'expo-tab'` on the base branch (`ExpoTabRouter` spreads `RNTabRouter`, so its state type is `tab`), so deleting it is pure cleanup. Adding `shouldActionChangeFocus` to `TabRouter` correctly restores the parent-focus propagation that the old coercion used to provide, and the drawer tests cover it.

Three things to consider before merge:

1. The breaking-change entry does not mention the impact on custom routers. Any router integrated through `unstable_integrateWithRouter` / `withLayoutContext` that only handles `NAVIGATE` now receives `PUSH` and returns `null`.
2. `StackRouter` still uses `BaseRouter.shouldActionChangeFocus`, which excludes `PUSH`, so the tab/stack behaviour under an open drawer is now inconsistent.
3. The new tests do not cover the two cases where `PUSH` semantics can differ from `NAVIGATE`: pushing to the already-focused tab, and anchor/`initial` handling when pushing into a nested navigator.

Note: I reviewed statically and did not run the test suite.

packages/expo-router/CHANGELOG.md
Ubax: Do it

packages/expo-router/src/react-navigation/routers/TabRouter.tsx
Ubax: Add `PUSH` to `BaseRouter`

packages/expo-router/src/__tests__/tab-router-push.test.ios.tsx
Ubax: Add additional test as well

packages/expo-router/src/__tests__/tab-router-push.test.ios.tsx
Ubax: Do it

packages/expo-router/src/__tests__/tab-router-push.test.ios.tsx
```diff
@@ -0,0 +1,246 @@
...
+  act(() => router.push('/second'));
+  act(() => router.push('/third'));
+
+  const tabState = result.getRouterState()!.routes[0]!.state;
+  expect(tabState?.routes).toHaveLength(3);
+  expect(screen).toHavePathname('/third');
+
+  act(() => router.push('/third'));
+  expect(tabState?.routes).toHaveLength(3);
+  expect(tabState?.index).toBe(2);
```
jakub-agent: [💡 suggestion] `tabState` is captured on line 74, before the duplicate `router.push('/third')`, and `getRouterState()` returns `store.state` — a plain snapshot, not a live reference. React Navigation replaces the state object on every change, so these two assertions still inspect the pre-push snapshot and pass no matter what the duplicate push does.

If `PUSH` ever started appending a second `third` route, `getRouterState()` would return a new object while `tabState` kept showing 3 routes at index 2, and the test would stay green — which defeats the point of the assertion.

Re-reading the state after the push fixes it:

```ts
act(() => router.push('/third'));
const tabStateAfterDuplicatePush = result.getRouterState()!.routes[0]!.state;
expect(tabStateAfterDuplicatePush?.routes).toHaveLength(3);
expect(tabStateAfterDuplicatePush?.index).toBe(2);
```

jakub-agent (approved):

Re-review of `1fc0aa1`. All four points from the previous round are addressed:

- The changelog entry now states the contract change ("Custom routers must now handle the `PUSH` action in `getStateForAction`") and carries the PR link.
- `shouldActionChangeFocus` moved into `BaseRouter` instead of overriding it in `TabRouter`. This is the better fix: stack, tab and drawer now agree on `PUSH`, the `NAVIGATE` / `NAVIGATE_DEPRECATED` list is no longer duplicated, and the new "push closes a parent drawer when switching a nested stack" test locks in the stack half of the behaviour.
- The tab test now pushes the focused route a second time.
- The headless-tabs test asserts the anchored nested stack (`['index', 'details']`) rather than just the pathname.

One leftover: the duplicate-push assertion reads a stale state snapshot, so it cannot fail. Details inline — it is a test-only issue, not a blocker.

Approving. Note that I reviewed statically and did not run the test suite.

PR #48756, opened 2026-08-11
[router] Remove the `initialParams` prop from Expo Router screens

Ubax:
# Why

Initial params make it impossible, to get correct initial state in the global scope, because they are runtime props in yet unmounted navigator. 

Removing them simplifies global navigation. The migration path for the change should be relatively simple, and makes the intent explicit in users application (see Accepted breaking changes).  

# How

1. Remove the prop from types
2. Remove the logic from `useNavigationBuilder` and routers
3. Update the tests

## Accepted breaking changes

- `<Stack.Screen initialParams` is removed. Pass params through navigation actions or route state instead. Use default values in your components e.g. `const { userId = initialUserId } = useLocalSearchParams()`
- Custom router `getStateForAction` implementations no longer receive `routeParamList`; action routes use only explicit action params, while merge actions combine existing route params with explicit action params.

# Test Plan

- CI

# Checklist

- [ ] This diff will work correctly for `npx expo prebuild` & EAS Build (not applicable: JavaScript router changes only).


packages/expo-router/src/ui/common.tsx
```diff
@@ -185,10 +183,8 @@ export function useTriggersToScreens(
     triggerMap[config.name] = config;
 
     if (config.type === 'internal') {
-      screenProps.push({
-        name: config.routeNode.route,
-        initialParams: config.initialParams,
-      });
+      // TODO(https://github.com/expo/expo/pull/TODO): Restore trigger default params for placeholder tabs, action-path route recreation, and rehydration.
```
jakub-agent: [🚨 **critical**] Two problems here.

**The link is a literal placeholder** — `https://github.com/expo/expo/pull/TODO` is not a valid URL. Please point it at the real follow-up PR or an issue before merging, otherwise this TODO is untraceable once the stack lands.

**The dropped behavior is wider than the changelog says.** `params` is still computed just above and merged into `action.payload.params`, so a trigger's params survive an explicit `JUMP_TO`. What is lost is the default: the screen config no longer carries them, so a trigger route that is materialized without going through the action — a placeholder tab descriptor, a route recreated from an action path, or a rehydrated state — now has `params === undefined` where it previously had the trigger's params. For a trigger like `href={{ pathname: '/[fruit]', params: { fruit: 'pear' } }}`, that means a screen or `options={({ route }) => ...}` callback reading `route.params.fruit` goes from a value to `undefined`.

CI is green because no test covers those three paths — `headless-tabs.test.ios.tsx` only exercises params after a `fireEvent.press` on the trigger, which takes the action path. Please either restore the behavior in this PR, or add a test that pins the current (regressed) behavior so the follow-up restore is verifiable, and add a changelog line for it — the existing entry only mentions the `initialParams` prop.

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
```diff
@@ -342,11 +342,8 @@ export function useNavigationBuilder<
   }, {});
 
   const routeNames = routeConfigs.map((config) => config.props.name);
-  const routeParamList = routeNames.reduce<Record<string, object | undefined>>((acc, curr) => {
-    const { initialParams } = screens[curr]!.props;
-    acc[curr] = initialParams;
-    return acc;
-  }, {});
+  // Expo Router gets params from navigation state, but routers still require this config field.
+  const routeParamList: Record<string, object | undefined> = {};
```
jakub-agent: [🏗️ **design**] `routeParamList` is now always `{}` on the `getStateForAction` / `getRehydratedState`-from-current-state paths. Only `initialRouteParamList` (line 415) still carries values.

That makes a lot of vendored router code unreachable: `StackRouter.tsx` reads `routeParamList` in ~15 places, plus `TabRouter.tsx`, `DrawerRouter.tsx`, and `layouts/stack-router.ts`. `createRouteFromAction` / `createParamsFromAction` now always fall through to `action.payload.params`. A follow-up that strips this plumbing would be worth scheduling — otherwise the next reader will assume the merge logic still does something.

Smaller point on this line: an empty object literal in the component body is re-allocated every render and is mutable, so a router could still write into it and silently lose the write on the next render. Hoisting it to a module-level `const EMPTY_ROUTE_PARAM_LIST = Object.freeze({})` makes the emptiness explicit and would surface any remaining writer in dev.

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
```diff
@@ -418,21 +415,16 @@ export function useNavigationBuilder<
...
         (acc, curr) => {
-          const { initialParams } = screens[curr]!.props;
           const initialParamsFromParams =
             route?.params?.state == null &&
             route?.params?.initial !== false &&
             route?.params?.screen === curr
               ? route.params.params
               : undefined;
 
+          // Keep child route params separate from the parent's nested navigation params.
```
jakub-agent: [💡 suggestion] This comment reads as the opposite of what the code does — the code copies the parent's nested navigation params *into* the child's initial param list. If the intent is that the spread gives the child its own object rather than aliasing `route.params.params`, say that: `// Copy so the child's params are not the same object as the parent's nested params.` As written it will mislead the next reader.

packages/expo-router/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### 🛠 Breaking changes
 
+- Remove the `initialParams` prop from Expo Router screens.
```
jakub-agent: [🐧 nit] Missing the PR link and author attribution that every other entry in this section carries: `([#48756](https://github.com/expo/expo/pull/48756) by [@Ubax](https://github.com/Ubax))`. The `initialRouteName` entry from the base PR has the same gap — worth fixing both while the stack is still open.

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
```diff
@@ -642,7 +633,7 @@ export function useNavigationBuilder<
       const config = screens[e.target];
 
       if (!route && config) {
-        route = { key: e.target, name: e.target, params: config.props.initialParams };
+        route = { key: e.target, name: e.target, params: undefined };
```
jakub-agent: [🐧 nit] `params: undefined` can be dropped now that there is no value to pass — `{ key: e.target, name: e.target }` reads the same and matches the placeholder-route shape used elsewhere.

jakub-agent (commented):

PR 2/9 in the router refactor stack (base: `@ubax/remove-prop-based-initialRouteName` / #48708).

The removal itself is clean and consistent: the prop is gone from `ScreenProps`, `StackScreenProps`, `RouteConfigProps`, `useScreens`, `useDescriptors`, and `useNavigationBuilder`, with no runtime consumers left in `packages/expo-router/src` outside `rsc/router` (unrelated identifier). Dropping the `routeParamList` copy in `ExpoTabRouter` is correct — that mutation became a no-op in #40352. The tests were migrated to `initialState` rather than deleted, which keeps most of the coverage intact. CI is green.

Two things need attention before this lands:

1. `ui/common.tsx` contains a `TODO(https://github.com/expo/expo/pull/TODO)` placeholder link, and the TODO text states that three behaviors of `expo-router/ui` triggers are dropped and not yet restored. That is a behavior change beyond what the changelog entry describes, and no test locks either the old or the new behavior.
2. `routeParamList` is now permanently empty in the action path, which makes a large amount of param-merging code in the vendored routers unreachable.

The rest is minor.

packages/expo-router/src/ui/common.tsx
Ubax: Link to https://github.com/expo/expo/pull/48756

Update the TODO to be a bit more explicit about what needs to be done

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
Ubax: Strip the code now

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
Ubax:
```suggestion
          // Copy so the child's params are not the same object as the parent's nested params.
```

packages/expo-router/CHANGELOG.md
Ubax: Update

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
Ubax: Do it

packages/expo-router/src/react-navigation/routers/TabRouter.tsx
```diff
@@ -103,7 +102,7 @@ const TYPE_ROUTE = 'route' as const;
 const addFallbackRouteIfEmpty = (
   routes: Route<string>[],
   routeNames: string[],
-  routeParamList: ParamListBase,
+  routeParamList: ParamListBase | undefined,
```
jakub-agent, line 105: [🐧 nit] Widening this to `| undefined` exists only to let the `ROUTE_NAMES_CHANGED` caller pass a literal `undefined` positionally, which reads as an oversight at the call site rather than an intent. The other two callers always have a real list.

Moving `routeParamList` to the end and making it optional would let that caller just omit it, and the `// Actions don't receive configured route params.` comment could move to the parameter instead:

```ts
const addFallbackRouteIfEmpty = (
  routes: Route<string>[],
  routeNames: string[],
  initialRouteName: string | undefined,
  /** Omitted when applying an action — actions don't receive configured route params. */
  routeParamList?: ParamListBase
) => { ... }
```

Nit only — the current form is correct.

jakub-agent (approved):

**Iteration 2** — re-reviewed at `5ada17d`. All five findings from the previous review are addressed, and the second one was taken further than suggested.

| Previous finding | Status |
| --- | --- |
| `pull/TODO` placeholder link + undocumented trigger-params gap | Fixed — real link, and the TODO now spells out all three paths and the override rule |
| `routeParamList` left as dead plumbing | Fixed — stripped from the action path entirely, not just noted |
| Comment contradicted the code | Fixed |
| Changelog missing link/attribution | Fixed, and it now also covers the `routeParamList` removal |
| Redundant `params: undefined` | Fixed |

The `RouterActionOptions = Omit<RouterConfigOptions, 'routeParamList'>` split is the right shape for this. It puts the invariant in the type system instead of a comment: `getStateForAction` can no longer read configured params, while `getInitialState` / `getRehydratedState` keep them. `createParamsFromAction` is deleted and `createRouteFromAction` loses the option, so the merge logic is gone rather than left unreachable. I checked every removed merge site in `StackRouter`, `TabRouter`, `layouts/stack-router.ts`, and `MockRouter` — each reduces to the same value once `routeParamList` is empty, so this is behavior-preserving given the prop is gone. `RouterActionOptions` is exported via `export * from './types'`, so custom routers can name it.

CI: all four `jest-ubuntu` shards, `check-packages`, `build`, and `android-build` pass. `android-test-e2e` reports as failed in the checks list but its actual conclusion is `skipped`.

One open item, tracked rather than blocking: `expo-router/ui` trigger href params still do not reach placeholder routes, action-recreated routes, or rehydrated state, and no test pins that behavior. The TODO documents it well, so this is fine to land as long as the follow-up lands in the same stack.

Approving. One nit below, take it or leave it.

PR #48757, opened 2026-08-11
[router] Make navigation state `type` optional

Ubax:
# Why

In order to be able to create initial state without knowing the unmounted navigator types first, we need to make the type optional and rework the cases where it is required.

# How

1. Make `state.type` type optional 
2. Add TODO for https://linear.app/expo/issue/ENG-22021/fix-link-preview-by-detecting-navigator-type-on-native in link preview logic
3. Add TODO for https://linear.app/expo/issue/ENG-22019/fix-candismiss-after-removing-the-navigation-state-type for `canDismiss` 
4. Remove the requirement for `stack` navigator type for `popToTop` actions - they can be safely dispatched to other navigators and will be ignored
5. Add `NavigatorTypeContext` to be able to check the navigator type from within content

# Test Plan

CI

# Checklist

### DIFF
--- a/packages/expo-router/src/react-navigation/core/getFocusedRouteNameFromRoute.tsx
+++ b/packages/expo-router/src/react-navigation/core/getFocusedRouteNameFromRoute.tsx
@@ -6,11 +6,12 @@ export function getFocusedRouteNameFromRoute(route: Partial<Route<string>>): str
   const state = route[CHILD_STATE] ?? route.state;
   const params = route.params as { screen?: unknown } | undefined;
 
+  // TODO(@ubax): https://github.com/expo/expo/pull/48757 - remove the stack.type check from here
   const routeName = state
     ? // Get the currently active route name in the nested navigator
       state.routes[
         // If we have a partial state without index, for tab/drawer, first screen will be focused one, and last for stack
-        // The type property will only exist for rehydrated state and not for state from deep link
+        // Deep-link partial states can omit both index and type; typeless full states still have an index
         state.index ??
           (typeof state.type === 'string' && state.type !== 'stack' ? 0 : state.routes.length - 1)
       ].name
--- a/packages/expo-router/src/react-navigation/routers/types.tsx
+++ b/packages/expo-router/src/react-navigation/routers/types.tsx
@@ -35,9 +35,11 @@ export type NavigationState<ParamList extends ParamListBase = ParamListBase> = R
   /**
    * Custom type for the state, whether it's for tab, stack, drawer etc.
    * During rehydration, the state will be discarded if type doesn't match with router type.
-   * It can also be used to detect the type of the navigator we're dealing with.
+   * It can also be used to detect the type of the navigator we're dealing with. Note that initial
+   * state does not include type, so an action needs to be dispatched in a navigator, in order
+   * for the type to be set
    */
-  type: string;
+  type?: string;
   /**
    * Whether the navigation state has been rehydrated.
    */
@@ -149,13 +151,20 @@ export type RouterConfigOptions = {
  */
 export type RouterActionOptions = Omit<RouterConfigOptions, 'routeParamList'>;
 
-export type Router<State extends NavigationState, Action extends NavigationAction> = {
-  /**
-   * Type of the router. Should match the `type` property in state.
-   * If the type doesn't match, the state will be discarded during rehydration.
-   */
-  type: State['type'];
+/**
+ * Type of the router. Should match the `type` property in state.
+ * If the type doesn't match, the state will be discarded during rehydration.
+ * Only routers whose state has no `type` may omit it, since a state without a `type`
+ * is accepted by every router.
+ */
+type RouterType<State extends NavigationState> = undefined extends State['type']
+  ? { type?: State['type'] }
+  : { type: State['type'] };
 
+export type Router<
+  State extends NavigationState,
+  Action extends NavigationAction,
+> = RouterType<State> & {
   /**
    * Initialize the navigation state.
    *
--- a/packages/expo-router/src/standard-navigation/__tests__/types.test.tsx
+++ b/packages/expo-router/src/standard-navigation/__tests__/types.test.tsx
@@ -8,6 +8,11 @@ import {
 
 import type { CommonNavigationAction, ParamListBase } from '../../react-navigation/core';
 import {
+  type DefaultRouterOptions,
+  type NavigationAction,
+  type NavigationState,
+  type Router,
+  type RouterFactory,
   TabRouter,
   type TabNavigationState,
   type TabRouterOptions,
@@ -70,6 +75,43 @@ const Nav = unstable_createStandardRouterNavigator<
   { initialRouteName?: string },
   TabRouterOptions
 >(Content, TabRouter);
+
+type TypelessNavigationState = Readonly<{
+  key: string;
+  index: number;
+  routeNames: string[];
+  routes: { key: string; name: string; params?: object }[];
+  stale: false;
+}>;
+
+const TypelessRouter: RouterFactory<
+  TypelessNavigationState,
+  NavigationAction,
+  DefaultRouterOptions
+> = () => ({
+  getInitialState: () => {
+    throw new Error('Type test only');
+  },
+  getRehydratedState: () => {
+    throw new Error('Type test only');
+  },
+  getStateForDeclaredRoutes: (state) => state,
+  getStateForRouteFocus: (state) => state,
+  getStateForAction: (state) => state,
+  shouldActionChangeFocus: () => false,
+});
+
+unstable_createStandardRouterNavigator(Content, TypelessRouter);
+
+// A router may omit `type` only when its state has none. A router whose state declares a literal
+// type must still declare the same literal, otherwise every state it produces would be rejected
+// by the rehydration check and replaced with a fresh initial state.
+export type _BaseRouterTypeIsOptional = Expect<
+  Equal<Pick<Router<NavigationState, NavigationAction>, 'type'>, { type?: string }>
+>;
+export type _TypedRouterTypeIsRequired = Expect<
+  Equal<Pick<Router<TabNavigationState<ParamListBase>, NavigationAction>, 'type'>, { type: 'tab' }>
+>;
 
 export type _HasScreen = Expect<Equal<typeof Nav extends { Screen: unknown } ? true : false, true>>;
 export type _HasProtected = Expect<
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### 🛠 Breaking changes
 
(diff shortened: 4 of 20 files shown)


packages/expo-router/src/react-navigation/routers/types.tsx
```diff
@@ -36,8 +36,9 @@ export type NavigationState<ParamList extends ParamListBase = ParamListBase> = R
    * Custom type for the state, whether it's for tab, stack, drawer etc.
    * During rehydration, the state will be discarded if type doesn't match with router type.
    * It can also be used to detect the type of the navigator we're dealing with.
+   * Custom states can omit it to opt out of type-based matching.
    */
-  type: string;
+  type?: string;
```
jakub-agent, line 42: [🏗️ **design**] `Router<State>['type']` is declared as `type: State['type']`, so it stays a **required** property whose value may be `undefined`. TypeScript still demands the key be present, so a custom router with a typeless state cannot simply omit it — it must write `type: undefined`, which is exactly what the new test router does.

Verified with the repo's `tsc --strict`:

```
error TS2741: Property 'type' is missing in type '{ go: ... }' but required in type 'Router<MyState>'.
```

If the goal is "custom routers can opt out of type matching", consider making the router field optional too:

```ts
type: State['type'] | undefined;  // still required to be written
// vs.
type?: State['type'];             // can be omitted
```

The built-in routers all declare a literal `type` (`'tab'`, `'stack'`, `'drawer'`), so widening this does not weaken them. The `Router.type` docblock ("Should match the `type` property in state. If the type doesn't match, the state will be discarded during rehydration.") is also now incomplete — a state with no `type` matches *any* router, and that is worth stating where router authors will read it.

packages/expo-router/src/react-navigation/routers/types.tsx
```diff
@@ -36,8 +36,9 @@ export type NavigationState<ParamList extends ParamListBase = ParamListBase> = R
    * Custom type for the state, whether it's for tab, stack, drawer etc.
    * During rehydration, the state will be discarded if type doesn't match with router type.
    * It can also be used to detect the type of the navigator we're dealing with.
+   * Custom states can omit it to opt out of type-based matching.
```
jakub-agent: [🏗️ **design**] Worth confirming the intent here, because "opt out of type-based matching" is asymmetric: omitting `type` doesn't make the state match *nothing*, it makes it match *everything*.

`useNavigationBuilder` treats a state as already-initialized when `state.stale === false && isStateValid(state)`, and `isStateValid` is `state.type === undefined || state.type === router.type`. So a typeless full state is accepted by every router — including `StackRouter`/`TabRouter`/`DrawerRouter` — and `getRehydratedState` is skipped entirely, meaning the state is rendered without any normalization.

Concrete failure: a custom typeless router persists `{ key, index, routeNames, routes, stale: false }`. The developer later replaces that navigator with a drawer. The persisted state passes the guard, rehydration is skipped, and `getDrawerStatusFromState` does `state.history.findLast(...)` on a state that has no `history` → `TypeError`. Before this change that state was unrepresentable in the types, so the guard was effectively watertight for full states.

This is arguably pre-existing runtime behavior rather than something the diff introduces, but the diff is what makes it a supported, typeable shape. Two options: tighten the initialized check so a typeless state is still routed through `getRehydratedState` (partial states have `stale: true` and are unaffected), or document the sharp edge here so router authors know a typeless state is portable across navigator kinds.

packages/expo-router/src/react-navigation/core/getFocusedRouteNameFromRoute.tsx
```diff
@@ -6,11 +6,12 @@ export function getFocusedRouteNameFromRoute(route: Partial<Route<string>>): str
   const state = route[CHILD_STATE] ?? route.state;
   const params = route.params as { screen?: unknown } | undefined;
 
+  // TODO(@ubax): https://github.com/expo/expo/pull/TODO - remove the stack.type check from here
```
jakub-agent: [🐧 nit] The TODO links to `https://github.com/expo/expo/pull/TODO` — a placeholder. Please point it at the real PR/issue or drop the link, so it doesn't ship as a dead reference.

packages/expo-router/src/standard-navigation/__tests__/types.test.tsx
```diff
@@ -71,6 +76,37 @@ const Nav = unstable_createStandardRouterNavigator<
...
 
+type TypelessNavigationState = Readonly<{
+  key: string;
+  index: number;
+  routeNames: string[];
+  routes: { key: string; name: string; params?: object }[];
+  stale: false;
+}>;
+
+const TypelessRouter: RouterFactory<
```
jakub-agent, line 87: [💡 suggestion] The only new coverage is type-level, and this router throws from `getInitialState`/`getRehydratedState` and returns `state` unchanged from the rest — so nothing about typeless routers is exercised at runtime.

A small render test would cover the part that actually matters: mount a navigator backed by a typeless router, confirm the state it produces has no `type`, and confirm it survives a re-render (`isStateValid`/`isStateInitialized` accept it). `useNavigationCache.test.ios.tsx` in this PR already shows the `BaseNavigationContainer` + `Screen` setup that would make this cheap.

jakub-agent (commented):

Small, focused type change: `NavigationState.type` becomes optional so custom routers can opt out of type-based rehydration matching. The runtime guard in `useNavigationBuilder` (`state.type === undefined || state.type === router.type`) already tolerated a missing type, so this mostly makes an existing runtime possibility expressible in the type system. `getFocusedRouteNameFromRoute` is comment-only, and the built-in routers keep their literal `type`, so nothing existing regresses.

Four points worth resolving before merge:

1. `Router['type']` is `State['type']`, which stays a **required** property. A custom router with a typeless state must still write `type: undefined` explicitly — exactly what the new test does. `type?: State['type']` would make the opt-out ergonomic.
2. Making `type` optional widens the rehydration guard: a typeless full state now passes `isStateInitialized` for *every* router, so `getRehydratedState` is skipped and the state is rendered as-is. Worth confirming this is intended.
3. A placeholder `pull/TODO` URL is left in shipped source.
4. The only new coverage is type-level; the test router throws from every state method, so nothing is exercised at runtime.

Also: this PR targets `@ubax/remove-push-to-navigate-modification`, not `main`, so it is part of a stack. The description is still the unfilled template (no Why / How / Test Plan, checklist unchecked) — the CHANGELOG entry is present but the checklist item is not ticked.

packages/expo-router/src/react-navigation/routers/types.tsx
Ubax: Modify to optional as well

packages/expo-router/src/react-navigation/routers/types.tsx
Ubax:
```suggestion
```

Just remove this line. It's not needed

packages/expo-router/src/react-navigation/core/getFocusedRouteNameFromRoute.tsx
Ubax: https://github.com/expo/expo/pull/48757

packages/expo-router/src/standard-navigation/__tests__/types.test.tsx
Ubax: Add a test

packages/expo-router/src/react-navigation/core/getFocusedRouteNameFromRoute.tsx
```diff
@@ -6,11 +6,12 @@ export function getFocusedRouteNameFromRoute(route: Partial<Route<string>>): str
   const state = route[CHILD_STATE] ?? route.state;
   const params = route.params as { screen?: unknown } | undefined;
 
+  // TODO(@ubax): https://github.com/expo/expo/pull/48757 - remove the stack.type check from here
```
jakub-agent, line 9: [🐧 nit] The dead `pull/TODO` link is gone — thanks. Small follow-up: this now points at the PR that *introduces* the TODO, which closes on merge. Anyone who later wants to know whether the cleanup happened lands on a merged PR that never mentions it. A tracking issue is the more durable target, or just drop the URL and keep the prose — `TODO(@ubax): remove the stack.type check from here` carries the same information without implying there is somewhere to look.

Take it or leave it; it does not affect behavior.

jakub-agent (approved):

**Iteration 2** — re-reviewed at `bf8b413`. All four points from the previous review are resolved.

| Previous finding | Status |
| --- | --- |
| `Router['type']` stayed a required property, forcing `type: undefined` | **Fixed** — the new `RouterType<State>` conditional makes the field optional exactly when the state's `type` is optional, and required when the state declares a literal |
| Typeless state is accepted by *every* router, skipping `getRehydratedState` | **Documented** — the `RouterType` docblock now states this directly, and the accept/discard semantics are pinned by tests |
| Placeholder `pull/TODO` URL | **Fixed** — resolves to #48757 |
| No runtime coverage | **Fixed** — `typelessRouter.test.ios.tsx` covers both the accepted and the discarded path |

The `RouterType` conditional is the right shape for this. `undefined extends State['type']` is the precise discriminator, and the paired assertions (`_BaseRouterTypeIsOptional`, `_TypedRouterTypeIsRequired`) lock down both branches, so a future refactor cannot silently make `type` optional for the built-in routers. The comment explaining *why* a typed router must keep declaring its literal — otherwise every state it produces fails the rehydration check — is the kind of reasoning that is hard to recover from the code later.

The new runtime test earns its place: deriving `TypelessRouter` from `MockRouter` by stripping `type` keeps it honest against the real router contract rather than a hand-rolled stub, and `index: 1` vs `index: 0` is a clean observable signal for "state kept" vs "state discarded".

**What I verified locally** (worktree at `bf8b413`, expo-router package):

- `typelessRouter.test.ios.tsx` — 2/2 pass
- `types.test.tsx` + `useNavigationCache.test.ios.tsx` — 5 suites, 9 tests pass
- `tsc -p packages/expo-router/tsconfig.json --noEmit` against `bf8b413` and against its parent produce **identical** error sets, so the change adds no type errors. (Both runs share 21 pre-existing errors from unresolved optional peers in my sandbox — `react-native-tab-view`, `react-native-pager-view`, `expo/internal/bundle-origin` — none in the changed files.)
- Reproduced the `RouterType` conditional standalone under `--strict`: a typeless router compiles with `type` omitted, and a `TabNavigationState` router still fails without `type: 'tab'`. `exactOptionalPropertyTypes` is off in `expo-module-scripts/tsconfig.base`, so `{ type?: string }` and `{ type?: string | undefined }` are the same type and the `Equal<>` assertion is sound.

Approving. One nit inline, plus two non-blocking notes: the `NavigationState.type` docblock is now the less informative of the two — it still only says the state is discarded on mismatch, while the "accepted by every router" consequence lives on `RouterType`. Router authors read the latter, so this is fine, but a one-line cross-reference on the state side would help anyone reading the state type first. And the PR description is still the unfilled template (no Why / How / Test Plan, checklist unticked) — the CHANGELOG entry now carries the proper link and attribution, which is the substance of that checklist item.

Still worth remembering this targets `@ubax/remove-push-to-navigate-modification`, not `main`, so it should land in stack order.

PR #48760, opened 2026-08-11
[router] Remove duplicated forked implementations of react-navigation functions

Ubax:
# Why

Before forking react-navigation into expo-router, we maintained forked versions of several functions. This PR merges them into a single implementation

# How

1. `createMemoryHistory`
2. `findFocusedRoute`
3. `getPathFromState`
4. `getStateFromPath`
5. `validatePathConfig`
6. `extractPathFromUrl`
7. `NavigationContainer`
8. `useBackButton`
9. `useLinking`
10. `useDocumentTitle`
11. `useThenable` 

## Accepted breaking changes

- Removed the deprecated `NavigationContainer` export.
- Removed `alias` from the public `PathConfig` type and `validatePathConfig` — now throws as extraneous.
- Nested routes reusing a parent's dynamic param name keep their own value instead of all segments seeing the last value.
- Implicit child selection in `getPathFromState` narrowed to group routes.
- Regexp param patterns in path configs are no longer applied.

# Test Plan

# Checklist


packages/expo-router/src/fork/validatePathConfig.ts
```diff
@@ -1,25 +1,73 @@
...
+export function validatePathConfig(config: unknown, root = true) {
+  const validation = {
+    path: 'string',
+    initialRouteName: 'string',
+    screens: 'object',
+    ...(root
+      ? null
+      : {
+          _route: 'object',
+          alias: 'array',
```
jakub-agent: [🚨 **critical**] `alias` is advertised as a valid key here, but `fork/getStateFromPath.ts` never reads it — `alias` appears nowhere else in `src/fork`. The vendored `react-navigation/core/getStateFromPath.tsx` that this PR deletes did implement it (it expanded `config.alias` into extra path configs, and threw `Screen 'X' doesn't specify a 'path'. A 'path' needs to be specified in order to use 'alias'.` when `alias` was set without `path`).

Since `react-navigation/core/index.tsx` now re-exports the fork version, a config using `alias` will pass validation and then be silently ignored — the worst failure mode, because there is no error and no matching route. `getStateFromPath` is reachable through the public `expo-router/react-navigation` entry point.

Three options, any is fine, but please pick one explicitly:
- port the `alias` handling into `fork/getStateFromPath.ts` (and the two deleted tests),
- drop `alias` from `validation` so the config throws `extraneous`,
- keep it accepted but throw a clear "not supported by Expo Router" error.

packages/expo-router/src/fork/__tests__/getStateFromPath-upstream.test.web.ts
```diff
@@ -2821,6 +2821,7 @@ test('throws when invalid properties are specified in the config', () => {
     - path (string)
     - initialRouteName (string)
     - screens (object)
+    - _route (object)
```
jakub-agent, line 2824: [🏗️ **design**] This PR deletes 4,841 lines of vendored upstream tests (`core/__tests__/getStateFromPath.test.web.tsx` and `core/__tests__/getPathFromState.test.tsx`), leaving the fork copies as the only conformance suite for what is now the only implementation. Comparing test names, nine did not make it across:

From `getStateFromPath`:
- `handles alias for paths`
- `throws if screen has alias but no path`
- `matches regexp patterns when provided`
- `regexp pattern doesn't match`
- `resolves nested path params with same name to correct screen`
- `handles parse in nested object for second route depth and path and parse in roots`

From `getPathFromState`:
- `prepends trailing slash to path`
- `ignores regexp patterns when provided`
- `correctly handles regex pattern with slash`

The `alias` and regexp ones cover features the fork does not implement, so dropping the tests is consistent — but that is exactly why it should be stated in the PR description as a behavior change rather than disappearing with the file. The remaining three (`prepends trailing slash to path`, `resolves nested path params with same name to correct screen`, `handles parse in nested object...`) look like plain behavior tests. Could you port those three, or say which ones the fork intentionally does not satisfy?

packages/expo-router/src/fork/getPathFromState.ts
```diff
@@ -190,6 +200,11 @@ export function getPathDataFromState<ParamList extends object>(
         // route.push('/home/(profile)')  --> This should navigate to /home/(profile)/me
         const screens = currentOptions[route.name]!.screens;
 
+        if (!allowImplicitChildSelection) {
```
jakub-agent: [🏗️ **design**] This is a behavior change, not a move: implicit child selection is now limited to group routes. Previously any route with `screens` in the config and no nested state got the index/first-screen fallback.

The narrowing matches the comment right above it ("Expo Router allows you to navigate to a (group) and not specify a target screen"), so the intent reads fine, and I assume it is what makes the upstream `getPathFromState` tests pass now that the fork is the only implementation. But nothing in the PR body or the changelog mentions it, and I could not find a test that exercises the non-group path — `fork/__tests__/getPathFromState.test.web.ts` only covers `(group)`.

Please add a test for a non-group route that has `screens` and no nested state, and note the change in `CHANGELOG.md`.

packages/expo-router/src/fork/getPathFromState.ts
```diff
@@ -136,6 +138,14 @@ export function getPathDataFromState<ParamList extends object>(
         // This mutates allParams
         const currentParams = expo.fixCurrentParams(allParams, route, stringify);
 
+        if (
+          !allowImplicitChildSelection &&
```
jakub-agent: [🏗️ **design**] Paired with the guard below, this changes the URL for a focused non-group route that carries a `screen` param: `fixCurrentParams` strips `screen`, this puts it back, and `focusedParams = { ...currentParams }` then emits it as `?screen=<name>`. Previously the param was consumed by implicit child selection and never reached the query string.

Note it also runs when `route.state` is set, where the value is discarded — `allParams` is assigned inside `fixCurrentParams` before this mutation, so only `currentParams` sees it. That is harmless but makes the intent hard to read; consider moving it under the `route.state === undefined` branch where it is actually used, and adding a test that asserts the resulting query string.

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
```diff
@@ -4,6 +4,7 @@ import { use } from 'react';
 // TODO(@ubax) - RN Migration: remove this dependency and just add this function to our codebase
 import { isValidElementType } from 'react-is';
 
+import { useComponent } from '../../fork/useComponent';
```
jakub-agent, line 7: [💡 suggestion] Three copies of `useComponent` collapse to one, and the surviving copy is the `ui/` variant, which wraps the returned component in `forwardRef`. The `core/` version this replaces did not. So `useNavigationBuilder`'s `NavigationContent` changes shape for every navigator.

That looks related to the assertion change in `src/__tests__/prefetch.test.ios.tsx`, where a trailing `'index'` header render disappears after `router.push('/second')`. One fewer render is probably an improvement, but a render-count change landing in a consolidation PR deserves a sentence in the description explaining what caused it.

Separately: `forwardRef` is legacy in React 19, where `ref` is a normal prop, and `AGENTS.md` for this package asks for React 19 patterns. Since `_ref` is unused in both variants, the `core/` version was the better one to keep — consider dropping `forwardRef` from `fork/useComponent.tsx`.

packages/expo-router/src/fork/getPathFromState.ts
```diff
@@ -118,6 +119,7 @@ export function getPathDataFromState<ParamList extends object>(
     let focusedParams: Record<string, any> | undefined;
     const focusedRoute = getActiveRoute(state);
     let currentOptions = configs;
+    const allowImplicitChildSelection = matchGroupName(route.name) != null;
```
jakub-agent: [💡 suggestion] `allowImplicitChildSelection` is computed once per outer iteration, but `route` is reassigned inside the inner loop when descending into a group. From that point on the flag describes the previous route's name, not the current one.

This preserves the pre-PR behavior (descending keeps working once it starts), so it is not a regression — but a reader will assume the flag tracks `route`. Either rename it to reflect that it is fixed at the outer route (for example `outerRouteIsGroup`) or recompute it in the inner loop and confirm the tests still pass.

packages/expo-router/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### 🛠 Breaking changes
 
+- Remove the deprecated `NavigationContainer` export
```
jakub-agent: [💡 suggestion] Missing the PR link and author attribution that the [changelog guide](https://github.com/expo/expo/blob/main/guides/contributing/Updating%20Changelogs.md) requires and that the neighbouring entries use:

```
- Remove the deprecated `NavigationContainer` export. ([#48760](https://github.com/expo/expo/pull/48760) by [@Ubax](https://github.com/Ubax))
```

While here, this entry is also the only record of the change — please add entries for the `getStateFromPath` behavior changes (`alias` and regexp patterns, depending on how you resolve the other comment) and for the `getPathFromState` implicit-child-selection narrowing.

packages/expo-router/src/fork/__tests__/useLinking.test.ios.tsx
```diff
@@ -1,8 +1,8 @@
 import { expect, jest, test } from '@jest/globals';
 import { render, type RenderAPI } from '@testing-library/react-native';
 
-import { useLinking } from '../../../fork/useLinking';
-import { createNavigationContainerRef, type ParamListBase } from '../../core';
+import { createNavigationContainerRef, type ParamListBase } from '../../react-navigation/core';
+import { useLinking } from '../useLinking';
 
 test.skip('throws if multiple instances of useLinking are used', () => {
```
jakub-agent: [🐧 nit] This file's only test is `test.skip`, so the whole file is inert. It was already skipped before the move, so this is not new — but the file is being relocated anyway, which is a good moment to either fix it or delete it. A skipped test that nobody has run reads as coverage without providing any.

jakub-agent (changes requested):

PR 3/3 reviewed in the current stack (base: `@ubax/add-router-registry` #48707). This collapses the duplicate implementations in `src/react-navigation/{core,native}` onto the `src/fork` copies: `getStateFromPath`, `getPathFromState`, `validatePathConfig`, `findFocusedRoute`, `useComponent`, `useLinking`, `createMemoryHistory`, `useBackButton`, `useDocumentTitle`, `useThenable`, `extractPathFromURL`, and the vendored `NavigationContainer`. Direction is right — one implementation per function is much easier to reason about than a fork plus a vendored twin.

I verified that no source file still imports any deleted module, and that `react-navigation/native/index` only drops `NavigationContainer`.

**Good catch in here:** the new `fork/validatePathConfig` adds `_route` to the allowed non-root keys. On the base branch, `fork/getStateFromPath` and `fork/NavigationContainer` validated with the vendored version, which rejects `_route`. `getReactNavigationConfig` emits `_route` whenever `metaOnly` is `false`, so this fixes a real latent throw on the non-meta config path.

**Main concerns:**

1. `fork/getStateFromPath` has no `alias` support, but the new `validatePathConfig` still lists `alias` as a valid key. The vendored `getStateFromPath` implemented it, and this PR deletes that implementation plus both of its tests. `getStateFromPath` is exported from `expo-router/react-navigation`, so this silently drops a documented capability.
2. Nine upstream conformance tests were not carried over into the retained `*-upstream` files.
3. `fork/getPathFromState` gains a real behavior change (`allowImplicitChildSelection`) that has no test and no changelog entry.
4. The PR description does not match the diff. It describes moving fresh-state creation into `useNavigationBuilder` and removing `Router.getInitialState` — that is #48783. Please replace it with a description of this change, and list the `NavigationContainer` and `getStateFromPath` behavior changes under "Accepted breaking changes".

I could not run `pnpm test` — the review worktree has no installed dependencies. All findings below come from reading the diff and the base revision, so please confirm the test-related ones against a real run.

packages/expo-router/src/fork/validatePathConfig.ts
Ubax: Addressed locally: removed `alias` from validation and from the public `PathConfig` type, so it now fails as an extraneous config key instead of being silently accepted.

packages/expo-router/src/fork/__tests__/getStateFromPath-upstream.test.web.ts
Ubax: Addressed locally: restored the three applicable upstream tests (`prepends trailing slash to path`, same-name nested params, and nested parse behavior). The same-name params test also required preserving each route’s own value while spreading params across route levels.

packages/expo-router/src/fork/getPathFromState.ts
Ubax: Addressed locally: added coverage confirming a non-group route with child screens and no nested state does not implicitly select a child. Per the requested scope, no additional changelog behavior note was added.

packages/expo-router/src/fork/getPathFromState.ts
Ubax: Addressed locally: scoped `currentParams.screen` assignment to routes without nested state and added tests for both `?screen=<name>` serialization and omission when nested state exists.

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
Ubax: Addressed locally: removed `forwardRef` from `useComponent`, matching the deleted core implementation and React 19 pattern. The prefetch assertions remain unchanged and pass.

packages/expo-router/src/fork/getPathFromState.ts
Ubax: Addressed locally: renamed the fixed outer-loop flag to `outerRouteIsGroup` and updated both usages.

packages/expo-router/CHANGELOG.md
Ubax: Addressed locally: added the PR link and author attribution to the existing `NavigationContainer` changelog entry. Per the requested scope, no extra behavior entries were added.

packages/expo-router/src/fork/__tests__/useLinking.test.ios.tsx
Ubax: Addressed locally: unskipped the iOS test, removed the extra native handler-count guard so the expected duplicate-linking warning fires, and ensured the test cleans up its render and console spy.

packages/expo-router/src/fork/getStateFromPath-forks.ts
```diff
@@ -30,7 +30,7 @@ export function populateParams(routes?: ParsedRoute[], params?: Record<string, a
   if (!routes || !params || Object.keys(params).length === 0) return;
 
   for (const route of routes) {
-    Object.assign(route, { params });
+    route.params = { ...params, ...route.params };
```
jakub-agent, line 33: [🏗️ **design**] This is the largest remaining behavior change in the PR, and it is easy to miss because it is one line.

Previously `Object.assign(route, { params })` gave every route in the chain the same merged params object, so a nested route reusing a param name saw the last value. That was deliberate — the upstream test you restored was commented out under `// Expo Router changes this functionality so all segments see the last :id`. With the merge, each route's own value wins, so `/foo/42/bar/43` now yields `Foo.params.id === '42'` and `Bar.params.id === '43'`.

`populateParams` is called from `fork/getStateFromPath.ts` inside the `START FORK` block, so this runs for every deep link and every path-based navigation. Any app with a nested dynamic segment that reuses a parent's param name gets a different `useLocalSearchParams` result.

That is a fine change to make — it matches upstream — but please add a `CHANGELOG.md` entry under breaking changes, and check whether the deleted "all segments see the last `:id`" note should be relocated rather than dropped.

packages/expo-router/src/__tests__/prefetch.test.ios.tsx
```diff
@@ -827,7 +827,6 @@ it('can still use <Screen /> while prefetching in tabs', () => {
     'Should only change after focus',
     'index',
     'Should only change after focus',
-    'index',
   ]);
```
jakub-agent, line 831: [🏗️ **design**] The reply on the `useComponent` thread says the prefetch assertions remain unchanged, but the diff at `da5489d` still drops a trailing `'index'` from this array — one fewer `headerTitle` render after `router.push('/second')`.

That is worth chasing rather than accepting, because the original hypothesis no longer holds: `fork/useComponent.tsx` is now byte-identical to the deleted `core/useComponent.tsx`, and this test uses `layouts/Tabs`, which goes through `useNavigationBuilder`. So the render-count change comes from something else in this PR — most likely `populateParams`, the `findFocusedRoute` swap in `BaseNavigationContainer`, or `useRoutePath` moving to the fork `getPathFromState`.

One fewer render is probably an improvement, but please confirm which change causes it before merging.

packages/expo-router/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### 🛠 Breaking changes
 
+- Remove the deprecated `NavigationContainer` export ([#48760](https://github.com/expo/expo/pull/48760) by [@Ubax](https://github.com/Ubax))
```
jakub-agent, line 7: [💡 suggestion] The link and attribution are correct now, thanks. The entry is still the only one, though, and this PR carries more than the `NavigationContainer` removal:

- `alias` removed from the public `PathConfig` type and from `validatePathConfig`. It worked before this PR via the vendored `getStateFromPath`, so this is a breaking change for anyone importing `PathConfig` from `expo-router/react-navigation`.
- Nested routes reusing a dynamic param name now keep their own value (see the `populateParams` comment).
- Implicit child selection in `getPathFromState` narrowed to group routes.
- The duplicate-linking `console.error` in `fork/useLinking.native.ts` now fires with two navigation containers instead of three.

The first two change observable app behavior, so they belong here regardless of how the rest are handled.

packages/expo-router/src/fork/__tests__/getStateFromPath-upstream.test.web.ts
```diff
@@ -2886,46 +2897,43 @@ test('encodes special characters in params', () => {
...
-//           ],
-//         },
-//       },
-//     ],
-//   };
-
-//   expect(getStateFromPath<object>(path, config)).toEqual(state);
-// });
-// End Fork
+test('resolves nested path params with same name to correct screen', () => {
```
jakub-agent, line 2900: [🐧 nit] Restoring this one with a real assertion is the right call — it documents the behavior instead of hiding it.

Four upstream tests are still dropped without a trace: `matches regexp patterns when provided` and `regexp pattern doesn't match` here, plus `ignores regexp patterns when provided` and `correctly handles regex pattern with slash` in the `getPathFromState` file. The fork genuinely does not support them — `formatRegexPattern` in `getStateFromPath-forks.ts` turns any `:param` into `(([^/]+\/)?)` and discards the constraint.

Since you used a `// Start Fork` comment elsewhere to record an intentional divergence, consider carrying these across the same way. A reader comparing against upstream otherwise has no way to tell whether regexp params are unsupported or just untested.

jakub-agent (commented):

Re-review of `da5489d` (iteration 2). Thanks for the point-by-point replies — I checked each one against the diff.

**Confirmed fixed:**

- `alias` is gone from `validatePathConfig` **and** from the public `PathConfig` type, with a test asserting it now fails as `- alias (extraneous)`. Clean resolution.
- `prepends trailing slash to path` and `resolves nested path params with same name to correct screen` are restored, and the `handles parse in nested object...` typo is fixed. Only the four regexp-pattern tests remain dropped.
- `outerRouteIsGroup` renamed, and the `currentParams.screen` assignment is now scoped to `route.state === undefined`. Three new tests cover the group / non-group / nested-state cases.
- `useComponent` kept the non-`forwardRef` core implementation, and `src/ui/useComponent.tsx` is deleted.
- The `useLinking` iOS test is unskipped and now passes against the corrected handler-count guard.
- The changelog entry has its PR link and attribution.

**Still open:**

1. Restoring the same-name nested params test required changing `populateParams` from replace to merge. That reverses a behavior Expo Router had deliberately forked ("all segments see the last `:id`") and changes `useLocalSearchParams` output on the live path→state path. It needs a changelog entry.
2. The `prefetch.test.ios.tsx` assertion is still changed in the diff, though the reply says it is unchanged. Since `useComponent` is now byte-identical to the deleted core version, something else in this PR causes it.
3. The changelog still carries only the `NavigationContainer` entry. Removing `alias` from the public `PathConfig` type is a breaking change on its own.
4. The PR description is now emptier than before — `# Why`, `# How`, and `# Test Plan` are blank, and "Accepted breaking changes" still lists #48783's items (`Router.getInitialState`, fresh-state overrides, sparse fresh-state metadata). None of those are in this diff.

None of these is a correctness defect, so this is a COMMENT rather than a block. Fill in the description, add the missing changelog entries, and explain the prefetch delta, and I have nothing else.

As before, I could not run `pnpm test` — the review worktree has no installed dependencies.

packages/expo-router/CHANGELOG.md
Ubax: Per the agreed scope, the changelog entry list remains unchanged. The additional behavior changes are documented in the PR’s `## Accepted breaking changes` section.

packages/expo-router/src/fork/__tests__/getStateFromPath-upstream.test.web.ts
Ubax: These four regexp tests are intentionally left out because regexp param constraints are unsupported by the fork. Per the agreed scope, I’m not adding skipped tests or divergence markers; the behavior is now listed in the PR’s `## Accepted breaking changes` section.

packages/expo-router/src/fork/getStateFromPath-forks.ts
Ubax: Per the agreed scope, I’m keeping the changelog limited to the existing `NavigationContainer` entry. This behavior change is now explicitly recorded in the PR’s `## Accepted breaking changes` section.

packages/expo-router/src/__tests__/prefetch.test.ios.tsx
Ubax: I tested all three suspects in isolation. Reverting `populateParams`, restoring the previous local `findFocusedRoute`, and restoring `useRoutePath` with the previous core `getPathFromState` each still produced the trailing `index`. The unmodified branch also produces it locally and in `check-packages`, so there is no render-count change to attribute to these production changes; the four-call assertion was incorrect. I restored the trailing `index`, and `pnpm test src/__tests__/prefetch.test.ios.tsx --no-cache` now passes.

packages/expo-router/src/fork/validatePathConfig.ts
```diff
@@ -1,25 +1,71 @@
-const formatToList = (items: string[]) => items.map((key) => `- ${key}`).join('\n');
+const formatToList = (items: Record<string, string>) =>
+  Object.entries(items)
+    .map(([key, value]) => `- ${key} (${value})`)
+    .join('\n');
 
-export default function validatePathConfig(config: any, root = true) {
-  const validKeys = ['initialRouteName', 'screens', '_route'];
+const expoOptions = new Set(['preserveDynamicRoutes', 'preserveGroups', 'shouldEncodeURISegment']);
```
jakub-agent, line 6: [🐧 nit] The `expoOptions` set covers `getPathFromState`'s three Expo-only keys, but `getStateFromPath`'s `Options` type also carries an Expo-only key — `previousSegments`, from `ExpoOptions` in `getStateFromPath-forks.ts`. It is not in this set, so `getStateFromPath(path, { screens, previousSegments })` would throw `- previousSegments (extraneous)`.

No caller does that today: `previousSegments` is threaded as a separate positional argument through `getStateFromPath` and `prepareConfigResources`, never inside `options`. So this is latent, and it behaved the same before the PR via the vendored validator.

Since you just introduced `expoOptions` to solve exactly this class of problem, it seems worth closing: either add `previousSegments` to the set, or drop it from `ExpoOptions` in `getStateFromPath-forks.ts` since nothing reads it off the options object.

jakub-agent (approved):

Re-review of `44933794` (iteration 3). Approving — every substantive finding from the first two rounds is resolved or explicitly and reasonably declined.

**Resolved this round:**

- The `prefetch.test.ios.tsx` assertion is back to its original four calls and the file no longer appears in the diff. Thanks for testing the three suspects in isolation rather than accepting the delta — that was the right way to close it.
- `## Accepted breaking changes` in the description now lists what this PR actually changes: the `NavigationContainer` removal, `alias` removal, nested same-name param resolution, the `getPathFromState` narrowing, and regexp param patterns no longer being applied.

**Verified across the whole change:**

- `alias` is gone from both `validatePathConfig` and the public `PathConfig` type, with a test asserting it now throws `- alias (extraneous)`.
- `expoOptions` correctly exempts `getPathFromState`'s three Expo-only keys at root, replacing the old destructuring wrapper.
- `_route` is allowed at non-root, which fixes the latent throw on the `metaOnly: false` config path.
- No source file still imports any deleted module; `react-navigation/native/index` only drops `NavigationContainer`.
- `useComponent` matches the deleted core implementation exactly, with no `forwardRef`.
- The restored upstream tests cover the same-name nested params behavior that the `populateParams` merge introduces.

**Two non-blocking notes:**

1. I still think the four behavior changes belong in `packages/expo-router/CHANGELOG.md` and not only in the PR body — the changelog is what reaches SDK users, and a PR description is not published anywhere they read. You have made that call twice now, so I am not holding the review on it. If the plan is to consolidate the changelog once at the top of the stack, that resolves it.
2. `# Why`, `# How`, and `# Test Plan` are still empty in the description. The repo PR template asks for all three.

One nit inline about a dead `Options` field. As in earlier rounds, I could not run `pnpm test` — the review worktree has no installed dependencies, so the test-related conclusions come from reading the diff.

PR #48762, opened 2026-08-11
[ios][updates] Reject asset keys that escape the updates directory 

alanjhughes:
# Why
Addresses https://exponent-internal.slack.com/archives/C013ZK4SA12/p1786425601131549

Asset filenames come from the manifest. `UpdateAsset.filename` is with no validation, and `RemoteAppLoader` joins it to the updates directory with `appendingPathComponent`, which appends raw text without resolving `..`. A `..` in either the key or the extension can escape the updates directory. 

Impact is confined to the app's own container and needs control of the update server, and code signing already blocks it. Still worth closing as it's a bug/.

# How
New `UpdatesUtils.isSafeFilename` rejects an empty name, `.`, `..`, path separators and NUL. It checks the whole filename because the extension is manifest controlled too. Rejection lives in `Update.update(withManifest:)`, which already throws and is the only production caller. The download and reaper sites are guarded as well, for rows written before this change.

# Test Plan
Five new tests across `UpdateTests` and `UpdateAssetTests` and full CI run

### DIFF
--- a/packages/expo-updates/ios/EXUpdates/Database/UpdatesReaper.swift
+++ b/packages/expo-updates/ios/EXUpdates/Database/UpdatesReaper.swift
@@ -71,6 +71,10 @@ public final class UpdatesReaper: NSObject {
 
         let beginDeleteAssets = Date()
         for asset in assetsForDeletion {
+          guard UpdatesUtils.isSafeFilename(asset.filename) else {
+            logger.warn(message: "Refusing to delete asset at unsafe path \(asset.filename)")
+            continue
+          }
           let localUrl = directory.appendingPathComponent(asset.filename)
           if FileManager.default.fileExists(atPath: localUrl.path) {
             do {
--- a/packages/expo-updates/ios/EXUpdates/UpdatesUtils.swift
+++ b/packages/expo-updates/ios/EXUpdates/UpdatesUtils.swift
@@ -78,14 +78,18 @@ public final class UpdatesUtils: NSObject {
   /**
    * Asset filenames are built from manifest-controlled values, so a filename holding a path
    * separator or a `..` component would resolve outside the updates directory.
+   *
+   * Matching is done on unicode scalars because `String.contains` compares grapheme clusters, and
+   * a separator followed by a combining mark forms one cluster that does not equal the separator.
+   * The filesystem still sees the separator byte.
    */
   public static func isSafeFilename(_ filename: String) -> Bool {
     return !filename.isEmpty &&
       filename != "." &&
       filename != ".." &&
-      !filename.contains("/") &&
-      !filename.contains("\\") &&
-      !filename.contains("\0")
+      !filename.unicodeScalars.contains("/") &&
+      !filename.unicodeScalars.contains("\\") &&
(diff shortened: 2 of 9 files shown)


packages/expo-updates/ios/EXUpdates/UpdatesUtils.swift
```diff
@@ -75,6 +75,19 @@ public final class UpdatesUtils: NSObject {
...
 
+  /**
+   * Asset filenames are built from manifest-controlled values, so a filename holding a path
+   * separator or a `..` component would resolve outside the updates directory.
+   */
+  public static func isSafeFilename(_ filename: String) -> Bool {
+    return !filename.isEmpty &&
+      filename != "." &&
+      filename != ".." &&
+      !filename.contains("/") &&
```
Kudo: [🚨 **critical**] `String.contains` matches grapheme clusters, so a `/` followed by a combining mark is not found, but the UTF-8 byte `0x2F` still reaches the filesystem. That bypasses the check.

Verified with a local Swift script:

```swift
let name = "..\u{2F}\u{0338}pwned.png"   // "..", then "/" + U+0338 as one Character
isSafeFilename(name)  // true
URL(fileURLWithPath: "/Documents/.expo-internal")
  .appendingPathComponent(name).standardizedFileURL.path
// "/Documents/̸pwned.png"  -> escaped the updates directory
```

Comparing scalars fixes it (`isSafeScalar(name)` returns `false` in the same script):

```suggestion
      !filename.unicodeScalars.contains("/") &&
```

Worth doing the same for the `\\` and `\0` lines below so the whole check works on scalars.

packages/expo-updates/ios/EXUpdates/Database/UpdatesReaper.swift
```diff
@@ -71,6 +71,10 @@ public final class UpdatesReaper: NSObject {
 
         let beginDeleteAssets = Date()
         for asset in assetsForDeletion {
+          guard UpdatesUtils.isSafeFilename(asset.filename) else {
```
Kudo, line 74: [💡 suggestion] Five other sites still join a database-sourced `asset.filename` to the updates directory: `AppLoader.swift:243` and `:376`, `AppLauncherWithDatabase.swift:216`, `UpdatesDatabaseIntegrityCheck.swift:38`, and `FileDownloader.swift:600`. They only read, so skipping them is defensible, but `AppLauncherWithDatabase` hands the resolved URL to the JS runtime as the launch bundle. Was leaving those out deliberate?

Kudo: @alanjhughes does this look like something valid?

alanjhughes: @Kudo I don't think so. It was deliberate to leave this as is. Exploiting this would only be possible if it was already taken advantage of before this fix. After this, it would be rejected the same way other assets are

Kudo: sounds good! thanks for confirming that

Kudo (approved):

PR #48763, opened 2026-08-11
[android][updates] Reject asset keys that escape the updates directory

alanjhughes:
# Why
Same bug as the PR below, on Android.

Android is the worse of the two. The write path calls `parentFile?.mkdirs()` and copies with `overwrite = true`, so a manifest can create directories and overwrite files anywhere the app's uid can write. On iOS, the write fails unless the target directory already exists.

# How
`UpdatesUtils.isSafeFilename` mirrors the iOS helper.

# Test Plan
Five new tests across `FileDownloaderManifestParsingTest, `UpdatesUtilsTest` and full CI run

### DIFF
--- a/packages/expo-updates/android/src/main/java/expo/modules/updates/db/Reaper.kt
+++ b/packages/expo-updates/android/src/main/java/expo/modules/updates/db/Reaper.kt
@@ -7,6 +7,7 @@ import expo.modules.updates.db.entity.AssetEntity
 import expo.modules.updates.db.entity.UpdateEntity
 import expo.modules.updates.manifest.ManifestMetadata
 import expo.modules.updates.selectionpolicy.SelectionPolicy
+import expo.modules.updates.utils.AndroidResourceAssetUtils
 import java.io.File
 
 /**
@@ -48,9 +49,13 @@ object Reaper {
         )
         continue
       }
-      // A row written before asset filenames were validated may point outside the updates directory.
       val relativePath = asset.relativePath
-      if (relativePath == null || !UpdatesUtils.isSafeFilename(relativePath)) {
+      // Embedded assets are served from the APK, so there is no file here to delete.
+      if (relativePath == null || AndroidResourceAssetUtils.isAndroidResourceAsset(relativePath)) {
+        continue
+      }
+      // A row written before asset filenames were validated may point outside the updates directory.
+      if (!UpdatesUtils.isSafeFilename(relativePath)) {
         Log.e(
           TAG,
           "Refusing to delete asset with URL " + asset.url + " at unsafe path " + relativePath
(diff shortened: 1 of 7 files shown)


packages/expo-updates/android/src/main/java/expo/modules/updates/db/Reaper.kt
```diff
@@ -47,7 +48,16 @@ object Reaper {
         )
         continue
       }
-      val path = File(updatesDirectory, asset.relativePath)
+      // A row written before asset filenames were validated may point outside the updates directory.
+      val relativePath = asset.relativePath
+      if (relativePath == null || !UpdatesUtils.isSafeFilename(relativePath)) {
```
Kudo: [🏗️ **design**] Embedded assets legitimately have a `relativePath` with slashes, so this will log an error for them on every reap.

`EmbeddedLoader.loadAsset` sets `relativePath = AndroidResourceAssetUtils.createEmbeddedFilenameForAsset(...)`, which produces `file:///android_asset/...` or `file:///android_res/...`. That is the default path since `EX_UPDATES_COPY_EMBEDDED_ASSETS` defaults to `false` (`android/build.gradle:79`). Those rows reach `deleteUnusedAssets()` once the update they belong to is gone, and every one of them will now print `Refusing to delete asset ... at unsafe path file:///android_asset/...`.

Behavior is unchanged (that `File` never existed, so nothing was deleted before either), but the error is misleading in exactly the place people will look after a security advisory. Skipping android resource paths first would fix it, the way `DatabaseLauncher` does:

```kotlin
if (relativePath == null || AndroidResourceAssetUtils.isAndroidResourceAsset(relativePath)) {
  // Embedded assets live in the APK; there is no file in the updates directory to delete.
  continue
}
if (!UpdatesUtils.isSafeFilename(relativePath)) {
  Log.e(...)
  continue
}
```

packages/expo-updates/android/src/test/java/expo/modules/updates/UpdatesUtilsTest.kt
```diff
@@ -34,6 +35,44 @@ class UpdatesUtilsTest : TestCase() {
...
+      "asset-1699999999-12345.png",
+      "..hidden.png",
+      "a..b.png"
+    )
+    safe.forEach {
+      Assert.assertTrue("expected \"$it\" to be accepted", UpdatesUtils.isSafeFilename(it))
+    }
+  }
+
+  fun testIsSafeFilename_everySafeNameStaysInsideItsDirectory() {
```
Kudo: [🐧 nit] The name promises a property over every safe name, but the body checks that one unsafe name escapes.

```suggestion
  fun testIsSafeFilename_rejectedNameEscapesItsDirectory() {
```

Kudo (approved):

PR #48769, opened 2026-08-11
[1/n] Upgrade React Native to 0.87.0

gabrieldonadel:
# Why

In preparation for React Native 0.87.0, we should bump our repo to use the latest react-native RC.

This does not bump Expo Go, that is done in another PR in this stack -> https://github.com/expo/expo/pull/48771

# How

- update package versions
  - `react-native  0.86.0-> 0.87.0-rc.4`  
  - `@react-native/normalize-colors 0.86.0 ->  0.87.0-rc.4` 
  - `@react-native/babel-preset 0.86.0 ->  0.87.0-rc.4` 
  - `@react-native/dev-middleware 0.86.0->  0.87.0-rc.4`    
- Update Expo router TS and drop InteractionManager usage 

# Test Plan

bare-expo ios / android
minimal-tester ios / android

# Checklist


packages/expo/bundledNativeModules.json
```diff
@@ -1,9 +1,9 @@
 {
-  "@expo/dom-webview": "~57.0.1",
-  "@expo/fingerprint": "~0.20.6",
-  "@expo/metro-runtime": "~57.0.8",
+  "@expo/dom-webview": "~56.0.5",
```
zoontek: @gabrieldonadel this is probably not right

templates/expo-template-bare-minimum/package.json
```diff
@@ -16,10 +11,10 @@
     "web": "expo start --web"
   },
   "dependencies": {
-    "expo": "~57.0.9",
-    "expo-status-bar": "~57.0.1",
+    "expo": "~56.0.5",
```
zoontek: @gabrieldonadel same in other package.json files

templates/expo-template-tabs/package.json
```diff
@@ -3,31 +3,26 @@
   "main": "expo-router/entry",
   "description": "The Tab Navigation project template includes several example screens.",
   "license": "0BSD",
-  "version": "57.0.11",
-  "repository": {
```
zoontek, line 7: Is the removal of the `repository` fields expected?

templates/expo-template-bare-minimum/package.json
gabrieldonadel: Oh seems like a wrong rebase, let me fix it

packages/expo/bundledNativeModules.json
gabrieldonadel: fixed

templates/expo-template-bare-minimum/package.json
gabrieldonadel: fixed

templates/expo-template-tabs/package.json
gabrieldonadel: Wrong rebase, reverted

apps/bare-expo/ios/BareExpo.xcodeproj/project.pbxproj
```diff
@@ -818,7 +814,7 @@
 				);
 				OTHER_LDFLAGS = "$(inherited)";
 				PODFILE_DIR = "$(SRCROOT)";
-				REACT_NATIVE_PATH = "${PODS_ROOT}/../../../../node_modules/.pnpm/react-native@0.86.0_@babel+core@7.29.7_@react-native+jest-preset@0.86.0_@babel+core@7.2_05fe3c1b42d0a26b4dba0087fc75e51f/node_modules/react-native";
+				REACT_NATIVE_PATH = "${PODS_ROOT}/../../../../node_modules/.pnpm/react-native@0.87.0-rc.4_@babel+core@7.29.7_@react-native+metro-config@0.87.0-rc.4_@bab_08e25b4062a6a14876992fee2ac3f3ce/node_modules/react-native";
```
zoontek: Rerun `bunx pod-install` now that you bumped to 0.87.0 stable

apps/brownfield-tester/integrated/ios/Podfile.lock
```diff
@@ -176,334 +187,112 @@ PODS:
...
-    - React-perflogger
-    - React-runtimeexecutor
-    - React-runtimescheduler
-    - React-utils
-    - ReactNativeDependencies
-    - Yoga
-  - React-Core/Default (0.86.0):
-    - hermes-engine
-    - RCTDeprecation
+  - RCTRequired (0.87.0-rc.3):
```
zoontek: Same here (and for all ios / macos projects)

zoontek:
@gabrieldonadel CI is failing because `react-native/rn-get-polyfills` has been removed from exports. To fix it, replace the import in `packages/@expo/cli/src/start/server/metro/withMetroMultiPlatform.ts` (line 181):

```ts
if (ctx.platform === 'web') {
  try {
    const rnGetPolyfills: () => string[] = require('@react-native/js-polyfills');
    return [
      ...virtualModulesPolyfills,
      // Ensure that the error-guard polyfill is included in the web polyfills to
      // make metro-runtime work correctly.
      // TODO: This module is pretty big for a function that simply re-throws an error that doesn't need to be caught.
      // NOTE(@kitten): This is technically the public API to get polyfills rather than resolving directly into
      // `@react-native/js-polyfills`. We should really just start vendoring these, but for now, this exclusion works
      ...rnGetPolyfills().filter((x: string) => !x.includes('/console')),
    ];
  } catch (error: any) {
    if ('code' in error && error.code === 'MODULE_NOT_FOUND') {
      // If @react-native/js-polyfills is not installed, because we're targeting web, we still continue
      // This should be rare, but we add it so we don't unnecessarily have a fixed peer dependency on react-native
      return virtualModulesPolyfills;
    } else {
      throw error;
    }
  }
}
```

This should do the trick!

apps/bare-expo/ios/BareExpo.xcodeproj/project.pbxproj
gabrieldonadel: We actually need to patch reanimated otherwise pod install fails due to their compatibility check

gabrieldonadel:
> @gabrieldonadel CI is failing because `react-native/rn-get-polyfills` has been removed from exports. To fix it, replace the import in `packages/@expo/cli/src/start/server/metro/withMetroMultiPlatform.ts` (line 181):
> 
> ```ts
> if (ctx.platform === 'web') {
>   try {
>     const rnGetPolyfills: () => string[] = require('@react-native/js-polyfills');
>     return [
>       ...virtualModulesPolyfills,
>       // Ensure that the error-guard polyfill is included in the web polyfills to
>       // make metro-runtime work correctly.
>       // TODO: This module is pretty big for a function that simply re-throws an error that doesn't need to be caught.
>       // NOTE(@kitten): This is technically the public API to get polyfills rather than resolving directly into
>       // `@react-native/js-polyfills`. We should really just start vendoring these, but for now, this exclusion works
>       ...rnGetPolyfills().filter((x: string) => !x.includes('/console')),
>     ];
>   } catch (error: any) {
>     if ('code' in error && error.code === 'MODULE_NOT_FOUND') {
>       // If @react-native/js-polyfills is not installed, because we're targeting web, we still continue
>       // This should be rare, but we add it so we don't unnecessarily have a fixed peer dependency on react-native
>       return virtualModulesPolyfills;
>     } else {
>       throw error;
>     }
>   }
> }
> ```
> 
> This should do the trick!

Thanks for the recommendation, I'm trying to use stacked PRs and keep the minimum amount of changes on each of those, so I've fixed @expo/cli in https://github.com/expo/expo/pull/48770

zoontek:
@gabrieldonadel LGTM, just bump `react-native-screens` to `4.27.0` to remove patches

zoontek (approved):

zoontek (approved):

PR #48770, opened 2026-08-11
[2/n] Upgrade React Native to 0.87.0

gabrieldonadel:
# Why

Part of the React Native 0.87 upgrade stack: #48769 (version bumps) → this PR (library fixes) → #48771 (Expo Go).

# How

- Fix Asset registry migration and JS/TS fixes: Metro's `assetRegistryPath`, expo-asset, expo-image, expo-video, jest-expo, and the TS declarations now point at `react-native/asset-registry`, so Metro-registered assets and React Native's own `<Image>` share one registry instance on native. Web keeps the virtual registry, and imports of the removed package are redirected for backwards compatibility. Also fixes `expo-ui` Host color scheme types against the stricter 0.87 API, supports 0.87's native `backgroundImage` style in html-elements, and removes an import of the deleted `RCTRootContentView.h`.

- Bump gradle plugins to Kotlin 2.2.21: explicitly enables `buildFeatures.buildConfig` where required, fixes the brownfield fused-mode template for the new `minSdk` DSL, declares the fused-mode AGP version as a minimum constraint instead of forcing it, and keeps Room's generated database constructor under R8.  

- Third-party library patches: 
  - `@shopify/react-native-skia` (uses iOS bridge APIs removed upstream), 
  - `react-native-screens` (retargeted at 4.26.0 with its 0.87 iOS fixes), 
  - `react-native-safe-area-context` (retargeted at 5.7.0 with its Kotlin fix), 
  - `react-native-reanimated`/`react-native-worklets` (declare 0.87 in `compatibility.json`, and configure the Kotlin `jvmTarget` reflectively so their Gradle scripts compile on any host classpath).  

# Test Plan

- CI on this PR: package unit tests (`et check-packages`), the Android and iOS build workflows, brownfield testers, and the updates E2E suite.
- Locally: clean iOS builds of `apps/minimal-tester` (including `use_frameworks! :linkage => :static`) and `apps/bare-expo` against 0.87 with the precompiled React-Core. bare-expo currently stops at a known `react-native-skia` 2.6.2 incompatibility with 0.87 that is tracked separately and not introduced by this PR.
- The Expo Go Android and iOS builds in #48771 run on top of this branch.


packages/@expo/cli/CHANGELOG.md
```diff
@@ -1799,7 +1800,6 @@ _This version does not introduce any user-facing changes._
 
 _This version does not introduce any user-facing changes._
 
-- Remove unused dependencies. ([#29177](https://github.com/expo/expo/pull/29177) by [@Simek](https://github.com/Simek))
```
zoontek, line 1808: There's a few changelog entries removal that looks genuine

packages/expo-dev-launcher/CHANGELOG.md
```diff
@@ -420,8 +422,6 @@ _This version does not introduce any user-facing changes._
 
 ### 🐛 Bug fixes
 
-- Restore config plugin `launchMode` support ([#41363](https://github.com/expo/expo/pull/41363) by [@gabrieldonadel](https://github.com/gabrieldonadel))
```
zoontek, line 423: Same here

packages/@expo/cli/CHANGELOG.md
gabrieldonadel: Fixed

packages/expo-dev-launcher/CHANGELOG.md
gabrieldonadel: Fixed

zoontek (approved):

packages/@expo/cli/src/start/server/metro/withMetroMultiPlatform.ts
```diff
@@ -656,8 +657,17 @@ export function withExtendedResolver(
...
-      if (/^@react-native\/assets-registry\/registry(\.js)?$/.test(moduleName)) {
-        return getAssetRegistryModule();
+      if (
+        moduleName === 'react-native/asset-registry' ||
+        /^@react-native\/assets-registry\/registry(\.js)?$/.test(moduleName)
+      ) {
+        // Serve a virtual registry on web (`react-native` isn't bundled there) and resolve to
+        // react-native's real registry on native so all consumers share one instance. The legacy
+        // `@react-native/assets-registry` package no longer ships with RN 0.87.
+        if (platform === 'web') {
```
kitten: Not sure why this was added? (comment and if-statment)
Generally this looks right, although, I'd expect if we need a follow-up we can pull that out into a separate PR.

Not sure of the context here though, but we'd want to keep this the same and always pointed at the virtual module

packages/@expo/cli/src/start/server/metro/withMetroMultiPlatform.ts
```diff
@@ -180,19 +180,20 @@ function withWebPolyfills(
...
+        // `react-native/rn-get-polyfills` used to be the public API for this, but React Native 0.87 dropped that
+        // subpath from its `exports`, so resolve `@react-native/js-polyfills` directly, matching `@expo/metro-config`.
+        // NOTE(@kitten): We should really just start vendoring these, but for now, this exclusion works
+        const rnGetPolyfills: () => string[] = require('@react-native/js-polyfills');
         return [
           ...virtualModulesPolyfills,
           // Ensure that the error-guard polyfill is included in the web polyfills to
           // make metro-runtime work correctly.
           // TODO: This module is pretty big for a function that simply re-throws an error that doesn't need to be caught.
-          // NOTE(@kitten): This is technically the public API to get polyfills rather than resolving directly into
```
kitten, line 189: Note Should be preserved/updated or changed to a todo. Despite this changing it made this original problem worse, not better 😅

packages/@expo/cli/CHANGELOG.md
```diff
@@ -38,6 +38,8 @@
...
 - Fix wirelessly connected iOS 16 and older devices being omitted from `expo run:ios --device` selection. ([#48127](https://github.com/expo/expo/pull/48127) by [@davellanedam](https://github.com/davellanedam))
 - Fix resolution of ESLint failing in `expo lint` after prerequisites check installs it ([#46223](https://github.com/expo/expo/pull/46223) by [@claritystorm](https://github.com/claritystorm))
+- Resolve the Metro asset registry to `react-native` core on native platforms so Metro-registered assets and React Native's own `<Image>` share one registry instance on React Native 0.87. Web keeps the virtual registry, and imports of the removed `@react-native/assets-registry` package are redirected. ([#47729](https://github.com/expo/expo/pull/47729) by [@gabrieldonadel](https://github.com/gabrieldonadel))
+- Depend on `@react-native/js-polyfills` directly for the web polyfills instead of the `react-native/rn-get-polyfills` subpath removed in React Native 0.87. ([#47729](https://github.com/expo/expo/pull/47729) by [@gabrieldonadel](https://github.com/gabrieldonadel))
```
kitten, line 41: nit: This shouldn't be added here maybe?

The main problem here is that the edit here replaces the leftover of `react-native/rn-get-polyfills` but a different PR added the dependency and replaced it in `@expo/metro-config` already. I haven't had a full look at figuring out yet how we'll untangle this yet, but it'll be a bit messy

We can either do that pre-merge, or remove the changelog entry for now (and maybe update the comment to a todo rather) and change/investigate and file a follow up PR as needed

Also note: I don't see the dependency on `@react-native/js-polyfills` here. We'll need to check that CI caught this and align this with `@expo/metro-config`. Not great that we'd have to add this dep twice, but that'd be temporary, but should probably be double-checked

packages/@expo/cli/src/start/server/metro/withMetroMultiPlatform.ts
gabrieldonadel: Sure, I can pull this out into a separate PR in the stack. 

For context: the standalone @react-native/assets-registry package is gone on 0.87 and RN core now imports its registry internally, where the resolver intercept can't see it. So now Metro's asset modules and RN's <Image> end up on two different registry instances on native, and Metro-registered assets don't resolve. Redirecting to react-native/asset-registry on native keeps one shared instance (web keeps the virtual registry since react-native isn't bundled there)

packages/@expo/cli/src/start/server/metro/withMetroMultiPlatform.ts
gabrieldonadel: Opps, reverted

packages/@expo/cli/CHANGELOG.md
gabrieldonadel: Fixed

packages/@expo/cli/src/start/server/metro/withMetroMultiPlatform.ts
kitten: main note here is `moduleName === 'react-native/asset-registry'` is fine and perfect 👍 the other part here is an unrelated change (and potentially redundant and/or one that we should assess separately, since it's basically a changed decision)

> Redirecting to `react-native/asset-registry` on native keeps one shared instance

The shared instance previously is basically the virtual module, so the main thing that's unexpected to me isn't that we're adding `react-native/asset-registry` to the redirects but that we've add `platform === 'web'` to the virtual module redirection, then replaced the native redirection with `getStrictResolver(context, platform)('react-native/asset-registry')`

So, it basically now splits us between the virtual module and the non-virtual module, instead of just the virtual module: https://github.com/expo/expo/blob/57eee4b6707154847491013ae1e7373dea5339b2/packages/%40expo/cli/src/start/server/metro/withMetroMultiPlatform.ts#L658-L669

PR #48783, opened 2026-08-11
[router] Remove getInitialState function from Router type

Ubax:
# Why

After all prerequisites we can now move all logic from `getInitialState` to `createInitialState`. This will unify the initial state between a router, allowing us to drop the nested params navigation in a follow-up PR.

# How

1. create `createInitialState` function which will create an initial state based on `routeNames` coming from metro context, `initialRouteName` coming from `unstable_settings` and route that is meant to be focused.
2. Remove `getInitialState` from all the routers
3. Update tests

# Test Plan

CI

# Checklist


packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
```diff
@@ -413,65 +396,40 @@ export function useNavigationBuilder<
...
       if (
         (currentState === undefined || !isStateValid(currentState)) &&
-        route?.params?.state == null &&
-        !(typeof route?.params?.screen === 'string' && route?.params?.initial !== false) &&
-        !isNestedParamsConsumed
+        routeParamsForInitialization?.state == null
       ) {
         return [
-          router.getInitialState({
+          createInitialState<State>({
```
jakub-agent, line 409: [🚨 **critical**] `routeParamList` does not exist in this scope. It is referenced here (line 412) and at line 428, but nothing in this file declares it — `initialRouteParamList` was the only local param list and this PR deletes it. Babel keeps the shorthand as a bare reference, so this is a runtime `ReferenceError`, not just a type error.

CI on this commit:
- `typecheck`: `error TS18004: No value exists in scope for the shorthand property 'routeParamList'` at `(412,13)` and `(428,11)`
- `test`: `Test Suites: 108 failed`, `Tests: 1275 failed`, all `ReferenceError: routeParamList is not defined`

Note the base branch already had one dangling reference to this variable (in `reduce`) which this PR correctly removes, so the stack needs a fix regardless.

Since #48756 removed the `initialParams` screen prop, there is no longer any source for a configured param list — `{}` is most likely the intended value at both sites, matching the two existing `routeParamList: {}` calls at lines 392 and 492. If that is right, please also see my comments on `createInitialState` and on line 426, because the whole option then becomes dead.

packages/expo-router/src/react-navigation/core/createInitialState.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  if (focusedRouteName !== undefined) {
+    routes.push(createRoute(focusedRouteName, true));
+  }
+
+  // TODO(@ubax): Improve these typings by distinguishing initial state from hydrated state types.
+  // Router state types may narrow the shared metadata added later by rehydration and actions.
+  return {
+    stale: false,
+    key: `navigator-${nanoid()}`,
```
jakub-agent, line 58: [🚨 **critical**] This state is returned with `stale: false` and no `type`, and the caller returns it directly without passing it through `getRehydratedState` — so unlike the old per-router `getInitialState` results, it never acquires a `type`.

It also never acquires one later: `isStateValid` accepts `type === undefined`, so `isStateInitialized` is true on the next render and rehydration is skipped, and the routers' `getStateForAction` handlers spread `...state` without adding `type`. A navigator initialised here stays typeless for its whole lifetime. That is the common case for any navigator not on the initial linking path (a stack inside a tab the user switches to later, a nested navigator reached by `push`).

Consumers that still branch on `type` and therefore take the wrong path:
- `global-state/router.ts:95` — `canDismiss()` returns `false` for a typeless stack even with several routes, so `router.dismiss()` guards break
- `native-tabs/NativeTabTrigger.tsx:68` — `throw new Error('Trigger component can only be used in the tab screen')` for a typeless tab state
- `react-navigation/native/useScrollToTop.tsx:72` — no tab navigator is ever found, so tab-press scroll-to-top stops working
- `global-state/stateUtils.ts:63,88` — `lookThroughAllTabs` traversal falls back to the focused route only
- `react-navigation/drawer/views/DrawerView.tsx:97` — nested stack is not popped to top on drawer navigation
- `utils/stack.ts:27` — `isRoutePreloadedInStack` returns `false`
- `link/preview/utils.ts:34,70` — acknowledged by the TODO added in this PR

A second consequence of skipping rehydration: router-specific initial-state rules no longer run. `TabRouter.getRehydratedState` calls `addFallbackRouteIfEmpty` and orders routes, which the old `TabRouter.getInitialState` mirrored; here, an `initial: false` tab navigator gets exactly two routes in `[configuredInitial, focused]` order rather than `routeNames` order, which is the order the tab bar renders.

Two directions, either is fine but please pick one explicitly rather than patching sites as they surface:
1. Have `useNavigationBuilder` run the fresh state through `router.getRehydratedState` (or have `createInitialState` take the router's `type`), so built-in navigators keep a stable `type`; or
2. Keep state typeless by design and, in this PR, convert every consumer above off `type`. The `NativeTabTrigger` `throw` needs handling either way — a wrong `type` check there is a crash, not a degradation.

packages/expo-router/src/react-navigation/core/createInitialState.tsx
```diff
@@ -0,0 +1,62 @@
...
+    typeof routeParams?.screen === 'string' && routeNames.includes(routeParams.screen)
+      ? routeParams.screen
+      : configuredInitialRouteName;
+  const usesNestedRouteParams = routeParams?.screen === focusedRouteName;
+
+  const createRoute = (name: string, focused: boolean): Route<string> => ({
+    key: `${name}-${nanoid()}`,
+    name,
+    ...(focused && usesNestedRouteParams
+      ? { params: routeParams?.params, path: routeParams?.path }
```
jakub-agent: [🏗️ **design**] Three things in this ternary:

1. **`routeParamList` is overridden rather than merged.** When the focused route is reached through nested params, `routeParamList[name]` is discarded. Every `getRehydratedState` in the package merges instead (`TabRouter.tsx:315-321`: `{ ...routeParamList[route.name], ...route.params }`), so a screen's configured params are lost on this path but present on the rehydrated one.
2. **Explicit `undefined` keys.** When `routeParams.params` / `routeParams.path` are absent, the route object still gets `params: undefined` and `path: undefined` keys. The old `getInitialState` implementations omitted the keys. This changes deep-equality and serialised state shape, which matters for the `state !== nextState` comparison in `useNavigationBuilder` and for state snapshots.
3. **`routeParamList` may be dead.** If line 412 is fixed by passing `{}` (see my other comment), this branch can never fire and both it and the `routeParamList` field on `InitialStateOptions` should go.

packages/expo-router/src/react-navigation/core/createInitialState.tsx
```diff
@@ -0,0 +1,62 @@
...
+    initialRouteName !== undefined && routeNames.includes(initialRouteName)
+      ? initialRouteName
+      : routeNames[0];
+  const focusedRouteName =
+    typeof routeParams?.screen === 'string' && routeNames.includes(routeParams.screen)
+      ? routeParams.screen
+      : configuredInitialRouteName;
+  const usesNestedRouteParams = routeParams?.screen === focusedRouteName;
+
+  const createRoute = (name: string, focused: boolean): Route<string> => ({
```
jakub-agent: [🏗️ **design**] The defensive copy of nested params is gone. The code this replaces spread them explicitly:

```js
// Copy the params so the child's route params don't alias the parent's nested params.
acc[curr] = initialParamsFromParams !== undefined ? { ...initialParamsFromParams } : undefined;
```

`params: routeParams?.params` stores the parent route's `params.params` object by reference in the child's route state. `deepFreeze` runs over navigation state in development, so freezing the child state also freezes an object the parent still owns, and a later `setParams` merge on the parent can be observed by the child. Was dropping the copy intentional? If so, the reason is worth a comment, since the previous one spelled out the hazard.

packages/expo-router/src/utils/stack.ts
```diff
@@ -15,13 +15,15 @@ export function getHistoryLength(state: ReactNavigationState): number {
...
 
+  // Without a type, both stack and tabs can use the route count as the history length.
   return state.routes.length;
 }
 
 export function isRoutePreloadedInStack(
   navigationState: NavigationState | undefined,
   route: { key: string }
 ): boolean {
+  // When preloading, the stack type will be added to the navigation state.
```
jakub-agent: [🏗️ **design**] This comment describes the mechanism accurately, and that is the problem: a state's `type` now depends on whether a `PRELOAD` happened to run, because the two `PRELOAD` branches in `StackRouter.tsx` (lines 690, 712) and the two in `layouts/stack-router.ts` are the only places that add it after initialisation.

So the same navigator answers `isRoutePreloadedInStack` — and `canDismiss()`, and `getFocusedRouteNameFromRoute` — differently before and after its first preload, with no action having changed anything relevant. That is hard to reason about and easy to regress: any future `PRELOAD` branch that forgets `type: 'stack'` silently breaks preload detection.

If state is going to stay typeless, `isRoutePreloadedInStack` should test the property it actually needs (`route` index past `state.index`, which is what line 30 already does) rather than the router type. If state should keep its `type`, set it once at initialisation and delete these four spot fixes.

packages/expo-router/src/link/preview/utils.ts
```diff
@@ -30,6 +30,7 @@ export function getTabPathFromRootStateByHref(
 
   const tabPath: TabPath[] = [];
   navigationRoutes.forEach((route, i, arr) => {
+    // TODO(ENG-22021): Fix link preview by detecting navigator type on native. https://linear.app/expo/issue/ENG-22021/fix-link-preview-by-detecting-navigator-type-on-native
```
jakub-agent, line 33: [🏗️ **design**] This TODO documents a functional regression that this PR causes: `route.state?.type === 'tab'` here and `navigationState.type === 'stack'` at line 70 are now `false` for any navigator initialised by `createInitialState`, so `getTabPathFromRootStateByHref` returns the wrong tab path and link previews break on native.

A linked follow-up is the right call, but shipping the regression in between is a decision for whoever owns the release, not a detail to leave in a comment. Two asks:
- Add the known regression to the `CHANGELOG.md` entry, or land ENG-22021 in the same stack before this reaches a release branch.
- Say in the PR description whether link previews are expected to be broken on `main` for a while.

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
```diff
@@ -413,65 +396,40 @@ export function useNavigationBuilder<
...
-              })
-            : router.getRehydratedState(stateBeforeInitialization, {
-                routeNames,
-                routeParamList: initialRouteParamList,
-                routeGetIdList,
-              });
-
-        return [hydratedState, false, paramsForState];
+          | State;
+        const hydratedState = router.getRehydratedState(stateBeforeInitialization, {
```
jakub-agent, line 424: [🏗️ **design**] Once line 428 is fixed, all three `getRehydratedState` call sites in this file pass `routeParamList: {}` (lines 392, 492, and this one). At that point `RouterConfigOptions.routeParamList` carries no information, and the merge logic that consumes it in `StackRouter.getRehydratedState`, `TabRouter.getRehydratedState` and `DrawerRouter.getRehydratedState` is dead code.

This PR is already removing dead surface from the `Router` interface, so removing `routeParamList` from `RouterConfigOptions` fits the same cleanup — and it is a breaking change for custom routers, so it is cheaper to do it in the same release as the `getInitialState` removal than in a later one.

Separately: on the base branch this call received `initialRouteParamList` (nested params only). If you do intend a non-empty list here instead, that is a behaviour change — configured params would now be merged into every restored and deep-linked route — and it needs a test plus a CHANGELOG note.

packages/expo-router/src/standard-navigation/__tests__/types.test.tsx
```diff
@@ -103,14 +100,11 @@ const TypelessRouter: RouterFactory<
...
 
-// A router may omit `type` only when its state has none. A router whose state declares a literal
-// type must still declare the same literal, otherwise every state it produces would be rejected
-// by the rehydration check and replaced with a fresh initial state.
 export type _BaseRouterTypeIsOptional = Expect<
   Equal<Pick<Router<NavigationState, NavigationAction>, 'type'>, { type?: string }>
 >;
-export type _TypedRouterTypeIsRequired = Expect<
-  Equal<Pick<Router<TabNavigationState<ParamListBase>, NavigationAction>, 'type'>, { type: 'tab' }>
+export type _TypedRouterTypeIsOptional = Expect<
```
jakub-agent, line 108: [💡 suggestion] The deleted comment recorded a real invariant: a router whose state declares a literal `type` had to declare the same literal, otherwise `isStateValid` (`useNavigationBuilder.tsx:345`) rejects every state it produces and re-initialises.

That hazard has not gone away — `getRehydratedState` in all three built-in routers still writes `type: 'tab'` / `'stack'` / `'drawer'`. A custom router that omits `type` on the router object while its `getRehydratedState` sets one now type-checks, and the failure mode is an initialise/reject loop rather than a compile error. Please keep an explanation of why `type` is optional and what a custom router must not do, near this assertion.

packages/expo-router/src/layouts/TabsClient.tsx
```diff
@@ -55,7 +55,7 @@ const Tabs = unstable_integrateWithRouter<
     preload: (name) => dispatch({ type: 'PRELOAD', payload: { name } }),
     popNestedStackToTop: (routeKey) => {
       const nestedState = state.routes.find((route) => route.key === routeKey)?.state;
-      if (nestedState?.type === 'stack' && nestedState.key) {
+      if (nestedState?.key) {
```
jakub-agent, line 59: [💡 suggestion] Dropping the `type === 'stack'` half of this guard means `popNestedStackToTop` now dispatches a targeted `POP_TO_TOP` at whatever navigator the nested state belongs to, including a nested tab or drawer. Those routers return `null` for the action so it is a no-op today, but the function's contract no longer matches its name and it depends on unrelated routers continuing not to handle `POP_TO_TOP`.

If the state is genuinely typeless now, a comment saying "the target router ignores the action when it is not a stack" would keep this honest. Otherwise, keep the guard using something the state still carries.

packages/expo-router/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### 🛠 Breaking changes
 
+- Remove `getInitialState` from the `Router` interface
```
jakub-agent: [🐧 nit] Missing the PR link and attribution that the other entries in this section carry: `([#48783](https://github.com/expo/expo/pull/48783) by [@Ubax](https://github.com/Ubax))`.

Also worth stating what a custom router must do instead, as entries #48752 and #48756 do — something like "custom routers no longer implement `getInitialState`; initial state is created by the navigator and passed to `getRehydratedState`".

jakub-agent (changes requested):

Replacing five per-router `getInitialState` implementations with one `createInitialState` is the right direction — the old implementations differed only in `type`, `key` prefix and `history`, and the `history` case is already covered defensively by `ensureStateHistory`. Two things block this PR though.

**1. It does not build or run.** `routeParamList` has no declaration in `useNavigationBuilder.tsx` (PR #48756 removed screen `initialParams`, and the base branch still referenced the variable once in `reduce`). This PR removes that reference but adds two new ones, at lines 412 and 428. CI on this commit: 2 × `TS18004` in `typecheck`, and `108 failed` test suites / `1275 failed` tests, every one of them `ReferenceError: routeParamList is not defined`.

**2. `createInitialState` produces state with no `type`, and ~8 consumers still branch on `type`.** Because the new state has `stale: false` and `isStateValid` accepts `type === undefined`, it never passes through `getRehydratedState`, so it never gains a `type` — a navigator initialised outside the linking/rehydration path stays typeless for its whole lifetime. That silently disables `canDismiss()`, `useScrollToTop`, `lookThroughAllTabs` traversal, drawer pop-to-top, `isRoutePreloadedInStack`, and link previews (the last one acknowledged in-tree with a TODO). `NativeTabTrigger` goes further and `throw`s. Sprinkling `type: 'stack'` into two `PRELOAD` branches patches one symptom but makes a state's `type` depend on whether a preload happened to run.

The PR description is empty (no Why / How / Test Plan). For a breaking change to a public `Router` interface plus a state-shape change of this reach, please state the intent and how it was verified — it also makes the remaining `type` consumers reviewable.

Smaller items are inline. Nothing here is unfixable; the direction looks good once the build is restored and the `type` consumers are dealt with deliberately rather than incidentally.

packages/expo-router/src/react-navigation/core/createInitialState.tsx
Ubax: This is not true. Check all previous branches in stack

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
Ubax: Remove `routeParamList` and all the logic related to initial params. It was removed in #48756

packages/expo-router/src/utils/stack.ts
```diff
@@ -17,6 +15,7 @@ export function getHistoryLength(state: ReactNavigationState): number {
     return state.index + 1;
   }
 
+  // Without a type or history, stack and tabs both use the route count as their history length.
```
jakub-agent, line 18: [🐧 nit] While you are in this file: the comment on `isRoutePreloadedInStack` just below (line 26) is now false.

> `// A typeless stack cannot be detected here, including after PRELOAD preserves its missing type.`

This PR adds `type: 'stack'` to every `PRELOAD` branch — `StackRouter.tsx:675` and `:697`, and `layouts/stack-router.ts:347` and `:379` — so a preloaded stack *does* carry a type by the time `isRoutePreloadedInStack` runs, which is exactly why the check works for its only real input. `useOnPreventRemove.tsx:33` still states that reasoning correctly:

> `// In order to preload routes in stack, an action needs to be dispatched, so the type will be always set when there are preloaded routes`

And the equivalent comment was deleted from `getHistoryLength` in this diff. So the three places that depend on the same invariant now document it two different ways and not at all. Worth making line 26 agree with `useOnPreventRemove` — it is the difference between "this check is knowingly incomplete" and "this check is sound for the input it gets".

jakub-agent (approved):

**Iteration 2** — re-reviewed at `bc39f11`. Everything from the previous round is addressed, and CI is green: `check-packages` passes and all four `jest-ubuntu` shards pass (was 108 suites / 1275 tests failing).

What changed:

- **The `routeParamList` reference error is gone**, and rather than patching the two call sites, `routeParamList` was removed from `RouterConfigOptions` outright, `RouterActionOptions` was deleted, and the now-dead merge logic was stripped from all three routers' `getRehydratedState` plus `addFallbackRouteIfEmpty`. Both breaking changes are in the CHANGELOG with attribution.
- **The typeless-state problem is handled at the right layer.** `NavigatorTypeContext` carries `router.type` to render consumers, and `NativeTabTrigger` reads it instead of `state.type` — that removes the only case that threw rather than degraded. `DrawerView` and `TabsClient` now document their targeted `POP_TO_TOP` as a deliberate no-op for non-stacks, and the sites that genuinely still need state-independent type resolution (`canDismiss`, `useScrollToTop`, `stateUtils`, link preview) carry linked TODOs (ENG-22019, ENG-22021). Most of that work sensibly moved down into the base branch, which shrank this PR's own diff.
- **`createInitialState` is tighter.** The defensive copy of nested params is back with its comment, `params`/`path` are conditionally spread so absent values no longer become explicit `undefined` keys, and the `routeParamList` option is gone. Nine unit tests now cover it, including `copies nested params` and `omits absent nested params and path` — the two specific behaviours flagged last round.
- **The `_TypedRouterTypeIsOptional` assertion** regained an explanation of the initialize/reject loop it guards against.

I checked two further things and found nothing: no code parses the navigator `key` prefix, so `navigator-` replacing `stack-`/`tab-`/`drawer-` is safe; and the `[configuredInitial, focused]` route order for `initial: false` is not a regression, since `addRouteIfMissing` appended on the old path too.

One stale comment inline, plus one process note: the PR description is still the unfilled template. For a breaking change to a public `Router` interface it is worth a few lines on Why and Test Plan, since the CHANGELOG entries cannot carry the reasoning.

Approving — the remaining item is a comment fix.

PR #48788, opened 2026-08-11
[ui] Support TextInputState with SwiftUI/Compose TextFields

intergalacticspacehighway:
# Why

https://exponent-internal.slack.com/archives/C085GKYRPFY/p1786426824243389

SwiftUI/Compose TextFields in RN ScrollView stay focused when `keyboardShouldPersistTaps` prop is set to `never` (default) and a user taps outside of the field.

This is RN specific feature where tapping outside of the `TextField` in a `ScrollView` blurs it. There is no native equivalent, but I guess supporting it makes sense since we have universal `TextInput` which matches RN's TextInput API.

# How

RN uses an API called `TextInputState` which is used by [ScrollView](https://github.com/react/react-native/blob/e71fc4e32ff6fdd0ebac20ba26b567f0814c803d/packages/react-native/Libraries/Components/ScrollView/ScrollView.js#L1477) and [Keyboard.dimiss()](https://github.com/react/react-native/blob/e71fc4e32ff6fdd0ebac20ba26b567f0814c803d/packages/react-native/Libraries/Utilities/dismissKeyboard.js#L19).

We patch this API to handle the blur for Expo UI text fields. This is a hack and better way would be to make RN's TextInputState extensible to cover external libraries.

# Test Plan

Tested the user shared repro. 

# Checklist


packages/expo-ui/src/keyboard/index.tsx
```diff
@@ -0,0 +1,217 @@
...
+  hostRef,
+  children,
+}: {
+  hostRef: RefObject<NativeViewWrapper | null>;
+  children: React.ReactNode;
+}) {
+  return <HostInstanceContext.Provider value={hostRef}>{children}</HostInstanceContext.Provider>;
+}
+
+function addField(instance: ReactNativeElement, field: HostedField) {
```
Kudo, line 81: [🏗️ **design**] This module is ~220 lines of pure JS bookkeeping with several ordering cases that are easy to get wrong later: `autoFocus` reporting focus before the registration effect runs, two fields swapping focus in either event order, a field unmounting while focused, and the last field unregistering the host. None of it needs native code to test.

`expo-ui` already has `@testing-library/react-native` and `__tests__` folders, and the repo `CLAUDE.md` asks for red/green. A test file over `addField` / `removeField` / `setFieldFocused` with a fake `TextInputState` would lock in the parts that a future refactor would silently break.

packages/expo-ui/src/keyboard/interop.ts
```diff
@@ -0,0 +1,94 @@
...
+  const blurTextInput = target.blurTextInput;
+  const focusTextInput = target.focusTextInput;
+
+  target.blurTextInput = (instance: any) => {
+    const controller = instance ? controllers.get(instance) : undefined;
+    if (!controller) {
+      blurTextInput(instance);
+      return;
+    }
+    TextInputState.blurInput(instance);
```
Kudo, line 84: [💡 suggestion] RN's `blurTextInput` only acts when the instance passed in is the focused one. This override drops that guard, so `TextInput.State.blurTextInput(someHostRef)` blurs that host's fields even when a different input holds focus.

Today's callers (`ScrollView`, `Keyboard.dismiss()`) always pass `currentlyFocusedInput()`, so nothing hits it yet. Matching RN keeps it that way:

```suggestion
    if (TextInputState.currentlyFocusedInput() !== instance) {
      return;
    }
    TextInputState.blurInput(instance);
```

packages/expo-ui/src/keyboard/index.tsx
```diff
@@ -0,0 +1,217 @@
...
+  useCallback,
+  useContext,
+  useEffect,
+  useMemo,
+  useRef,
+  type Ref,
+  type RefObject,
+} from 'react';
+import type { ReactNativeElement } from 'react-native';
+import useMergeRefs from 'react-native/Libraries/Utilities/useMergeRefs';
```
Kudo: [💡 suggestion] `react-native-web` has no `Libraries/Utilities/useMergeRefs`, so this import cannot resolve on web — but `textInputState.web.ts` implies you expect this module to be reachable there.

As far as I can tell it is not: `swift-ui/Host` and `jetpack-compose/Host` call `requireNativeView` at module scope, which throws on web, and `universal/Host/index.tsx` (the web variant) never imports them. If that matches your reading, `textInputState.web.ts` is dead and can go. If the module really is web-reachable, `useMergeRefs` needs a shim too — otherwise the web bundle fails to resolve, not just at runtime.

packages/expo-ui/src/jetpack-compose/Host/index.tsx
```diff
@@ -8,7 +8,9 @@ import {
   I18nManager,
   useColorScheme as useRNColorScheme,
 } from 'react-native';
+import useMergeRefs from 'react-native/Libraries/Utilities/useMergeRefs';
```
Kudo: the deep internal import may have a warning when bundling. maybe we could vendor it inside expo-ui?

packages/expo-ui/src/keyboard/index.tsx
Kudo: react-native-web has its `react-native-web/dist/modules/useMergeRefs` but we could just vendor the useMergeRefs

Kudo (approved):
good to vendor useMergeRefs. though we still have the other deep imports.

leave some other ai review comments also but let you makes the call

packages/expo-ui/src/keyboard/index.tsx
intergalacticspacehighway: Agree, added testcases - https://github.com/expo/expo/pull/48788/commits/afc84cc7591aada5f41ca5faab7ba6d119c9ad6c

Issue #48780, opened 2026-08-11
[expo-camera][iOS Simulator] Mounting and unmounting CameraView several times freezes the UI


nrgbistro:
### Minimal reproducible example

https://github.com/nrgbistro/expo-camera-simulator-freeze-repro

### Steps to reproduce

I'm seeing the iOS Simulator freeze for around 15 seconds after opening and closing a modal containing `CameraView`.

1. Run an Expo SDK 57 development build in the iOS Simulator.
2. Open a modal containing `CameraView`.
3. Save a placeholder or close the modal to dismiss it.
4. Repeat this a few times fairly quickly.

The entire Simulator UI eventually becomes unresponsive. The delay shows up somewhat randomly, but often between 3-6 CameraView cycles.

I can reproduce this both with `takePictureAsync()` and with a development-only placeholder button. The placeholder path does not capture a photo—it just accepts a small data-URI image and closes the modal while `CameraView` is mounted.

The source of the issue is un and re-mounting the CameraView, the saved content has no impact. I did not notice this issue on previous Expo SDKs

### Expected behavior

Saving or dismissing a photo should not block the UI. If the Simulator cannot start a usable camera session, `CameraView` should use its Simulator fallback or report an error without blocking the main thread.

### Actual behavior

Every `CameraView` mount creates an `AVCaptureSession`. The session fails immediately, but its startup and teardown continue waiting in AVFoundation. After several attempts, these waits occur sequentially and freeze the main thread.

Focused log excerpt from a three-photo reproduction:

```text
15:52:24.760 FigCaptureSessionSimulator signalled err=-12782
15:52:24.772 AVCaptureSessionRuntimeErrorNotification
             AVFoundationErrorDomain Code=-11800
             NSUnderlyingError=NSOSStatusErrorDomain Code=-12782

15:52:26.639 FigCaptureSessionSimulator signalled err=-12782
15:52:28.424 FigCaptureSessionSimulator signalled err=-12782

15:52:33.776 AVCaptureSession _buildAndRunGraph:
             Timed out waiting for session to start
15:52:35.641 AVCaptureSession _buildAndRunGraph:
             Timed out waiting for session to start
15:52:37.426 AVCaptureSession _buildAndRunGraph:
             Timed out waiting for session to start
```

A process sample taken during the freeze showed the main thread spending essentially the entire sample in:

```text
CA::Transaction::commit
NSKVODeallocate
AVCaptureVideoPreviewLayer dealloc
[...]

PR #48806, opened 2026-08-11
verify: silence the did-not-finish notice in quiet mode, and fix dispatch fix-mode defaults

brentvatne:
One-line follow-up to #48805, which I got wrong.

`env.QUIET` gated `Acknowledge` and `Announce run` but not `Report a run that did not finish`, so a quiet dispatch that fails or is cancelled still comments — the exact thing quiet mode exists to prevent.

Reproduced within a minute of merging: I dispatched a run against #48780 while testing, cancelled it a few seconds later, and it posted `⚠️ /verify was cancelled before it could post findings` onto a thread that had been successfully verified an hour earlier.

In quiet mode the person who dispatched is watching the Actions tab, which already shows the outcome.

### DIFF
--- a/.github/workflows/verify-command.yml
+++ b/.github/workflows/verify-command.yml
@@ -1018,7 +1018,7 @@ jobs:
       # with no result. The job also runs continue-on-error, so nothing else
       # signals it either.
       - name: Report a run that did not finish
-        if: always() && steps.gate.outputs.ok == 'true' && (failure() || cancelled())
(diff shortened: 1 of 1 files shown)


brentvatne:
Landed directly on main as a72d852 — this PR was blocked on required checks, and both changes fix live behaviour: a cancelled quiet dispatch still commented on the thread, and dispatch fix mode defaulted to no-fix where the comment path would have attempted one.

Issue #48802, opened 2026-08-11
expo-modules-jsi fails to compile on Xcode 26.3: ambiguous abs() in JavaScriptCodable+Date.swift:53


sparkyfen:
### Minimal reproducible example

https://github.com/sparkyfen/expo-jsi-abs-repro

### Steps to reproduce

An unmodified `npx create-expo-app@latest --template blank` app (SDK 57) fails to build for iOS on Xcode 26.3. Development build, bare `xcodebuild` and `npx expo run:ios` all fail identically; npm as package manager.

1. `npm install`
2. `npx expo prebuild -p ios && cd ios && pod install && cd ..`
3. `xcodebuild -workspace ios/expojsiabsrepro.xcworkspace -scheme expojsiabsrepro -configuration Debug -destination "generic/platform=iOS Simulator" build`

Expected: the app compiles.

Actual: compiling `ExpoModulesJSI` fails:

```
node_modules/expo-modules-jsi/apple/Sources/ExpoModulesJSI/Coding/JavaScriptCodable+Date.swift:53:50:
error: type of expression is ambiguous without a type annotation
  guard milliseconds.isFinite, abs(milliseconds) <= maxJavaScriptDateMilliseconds else {
                               ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

The Xcode 26.3 Swift frontend (Swift 6 mode) can no longer resolve the `abs` overload in this expression. The same line is present in `expo-modules-jsi@58.0.0-canary-20260806-8c2d007`, so the canary is affected too.

This one-line change compiles and is semantically identical for the range check:

```diff
-  guard milliseconds.isFinite, abs(milliseconds) <= maxJavaScriptDateMilliseconds else {
+  guard milliseconds.isFinite, milliseconds.magnitude <= maxJavaScriptDateMilliseconds else {
```

(We're carrying it via patch-package in the meantime.)

### Environment

```
expo-env-info 2.1.0 environment info:
  System:
    OS: macOS 15.7.7
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.13.0 - ~/.nvm/versions/node/v24.13.0/bin/node
    npm: 11.6.2 - ~/.nvm/versions/node/v24.13.0/bin/npm
  Managers:
    CocoaPods: 1.16.2 - /opt/homebrew/bin/pod
  SDKs:
    iOS SDK:
[...]

PR #48807, opened 2026-08-11
[tools] Add shared SDK compatibility registry

brentvatne:
# Why

Expo SDK platform and toolchain compatibility data currently has multiple sources of truth:
- [the docs contain the full SDK table](https://github.com/expo/expo/blob/577c90cada56301bfaefd4bb023e122d7f4175e5/docs/ui/components/SDKTables/sdk-versions.json#L1-L88)
- while [expo-doctor has a separate partial Xcode table](https://github.com/expo/expo/blob/main/packages/expo-doctor/src/checks/NativeToolingVersionCheck.ts#L39-L44)

This divergence allowed the unsupported SDK 57 and Xcode 26.3 environment reported in #48802 to pass expo-doctor.

This PR introduces a neutral, machine-readable source of truth that the docs, expo-doctor, release tooling, and repository automation can consume. Consumer migrations will follow separately.

# How

- Add a private @expo/sdk-compatibility workspace package.
- Seed semantic compatibility data for SDK 51 through SDK 57 from the current docs table and historical SDK 51 documentation.
- Represent Xcode support as semver ranges rather than presentation strings.
- Add typed SDK lookup and Xcode compatibility helpers.
- Validate schema version, ordering, unique SDK majors, and version ranges.
- Add a regression test for SDK 57 with Xcode 26.3 versus 26.4.

# Test Plan

- corepack pnpm@10.33.0 --filter @expo/sdk-compatibility build
- corepack pnpm@10.33.0 --filter @expo/sdk-compatibility test --runInBand
- corepack pnpm@10.33.0 --filter @expo/sdk-compatibility lint
- corepack pnpm@10.33.0 --filter @expo/sdk-compatibility typecheck
- corepack pnpm@10.33.0 --filter @expo/sdk-compatibility depscheck

# Checklist

- [ ] I added a changelog.md entry and rebuilt the package sources according to the contributing guide. Not applicable: this is a new private package.
- [ ] This diff will work correctly for npx expo prebuild and EAS Build. Not applicable: no native project generation behavior changes.
- [ ] Conforms with the Documentation Writing Style Guide. Not applicable: no user-facing docs pages change.


brentvatne:
Review follow-up:

- `id:7e4d9dc47543` is addressed by stacked PR #48809. That PR deletes `docs/ui/components/SDKTables/sdk-versions.json`, makes the docs read `@expo/sdk-compatibility/data`, and changes `docs/scripts/sdk-release/beta.ts` to add new SDK rows to the shared registry. PR #48807 has no registry consumers by itself, so there is no stale-read execution path between the two merges.

- `id:91eb420fd64f` conflates an unpublished next-SDK breaking change on `main` with SDK 57 compatibility. The Node `engines` change landed in 802b8c3402f as an **Unpublished** breaking change and is not an ancestor of `origin/sdk-57`. The SDK 57 release branch contains `expo@57.0.12` with no `engines` field, and the published `expo@57.0.12` package likewise exposes no `engines.node` metadata. Adding `^22.13.0 || ^24.3.0 || ^26.0.0 || >=27.0.0` to the SDK 57 row would therefore introduce a restriction that SDK 57 does not declare.

The registry intentionally retains the documented SDK 57 recommendation as:

```json
"node": {
  "minimumVersion": "22.13.0"
}
```

Its README explicitly says this is the recommended documentation minimum, not an exact supported-version range. Exact package/tool engine policies should remain sourced from those packages rather than being projected backward onto an SDK release.

packages/@expo/sdk-compatibility/src/schema.ts
```diff
@@ -0,0 +1,137 @@
...
+  if (options.semver && !semver.valid(value)) {
+    errors.push(`${path} must be a valid semantic version.`);
+  }
+  if (options.semverRange && !semver.validRange(value)) {
+    errors.push(`${path} must be a valid semantic version range.`);
+  }
+};
+
+const requireNumber = (value: unknown, path: string, errors: string[]) => {
+  if (typeof value !== 'number' || !Number.isInteger(value) || value < 0) {
```
ramonclaudio, line 27: Should we be using `value <= 0` here since Androids `minimumVersion`, `compileSdkVersion`, and `targetSdkVersion` fields start at `1`?

ramonclaudio (approved):

packages/@expo/sdk-compatibility/src/schema.ts
brentvatne: fair. i think in this case we'll never be using versions this low and so it's ok

PR #48821, opened 2026-08-12
[expo-image] fixed image's internal dataSet marker being overwritten by a user-provided dataSet

Brentlok:
# Why

 Expo Image’s internal `expoimage` data-set marker could be overwritten when consumers passed a `dataSet` prop.

# How

Added a web-only `dataSet` prop type and merge user-provided values with Expo Image’s internal marker, ensuring `expoimage: true` is retained.

# Test Plan

- Added a web unit test verifying that Expo Image preserves user-provided `dataSet` entries and retains its internal `expoimage: true` marker.
- Ran:
```sh
   pnpm --filter expo-image test -- ExpoImage.test.web.tsx --runInBand
   pnpm --filter expo-image run build
   pnpm --filter expo-image run typecheck
```

# Checklist

### DIFF
--- a/packages/expo-image/src/ExpoImage.web.tsx
+++ b/packages/expo-image/src/ExpoImage.web.tsx
@@ -12,7 +12,7 @@ import useSourceSelection from './web/useSourceSelection';
 loadStyle();
 
 type ExpoImageWebProps = ImageNativeProps & {
-  dataSet?: Record<string, unknown>;
+  dataSet?: Record<string, string | undefined>;
 };
 
 function onLoadAdapter(onLoad?: (event: ImageLoadEventData) => void) {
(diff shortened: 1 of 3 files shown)


brentvatne:
@expo-bot review

packages/expo-image/src/ExpoImage.web.tsx
```diff
@@ -11,6 +11,10 @@ import useSourceSelection from './web/useSourceSelection';
 
 loadStyle();
 
+type ExpoImageWebProps = ImageNativeProps & {
+  dataSet?: Record<string, unknown>;
```
zoontek: Could we align this typing with the current `@expo/ui` `webUtils.ts` typing and use `dataSet?: Record<string, string | undefined>`? Since `data-*` values are serialized to strings on the DOM element, accepting unknown here can be misleading.

Also, could `expoimage: true` be changed to `expoimage: 'true'` (or `String(true)`) to match that type?

zoontek (dismissed):

packages/expo-image/src/ExpoImage.web.tsx
Brentlok: Thanks for review, I've updated the type as suggested, and changed `expoimage: 'true'`

zoontek (approved):

PR #48826, opened 2026-08-12
fix(require-utils): Fix Node 26 `stripTypeScriptTypes` fallback call

### LINKED ISSUE #48725: expo config fails to load app.config.ts on node 26 (stripTypeScriptTypes transform mode removed)

https://github.com/MaryanPrydatko/expo-node26-config-repro

it's a stock `npx create-expo-app@latest --template blank` with `app.json` renamed to `app.config.ts`. nothing else changed, no extra deps, no typescript installed.

### Steps to reproduce

platform: macos, npm, managed workflow. happens on any platform target since it fails before that.

1. install node 26 (`brew install node` gives 26.7.0 today)
2. `npm install`
3. `npx expo config --json`

expected the config to print. instead:

```
TypeError [ERR_INVALID_ARG_VALUE]: Error reading Expo config at app.config.ts:

The property 'options.mode' must be one of: 'strip'. Received 'transform'
    at Object.stripTypeScriptTypes (node:internal/modules/typescript:101:3)
    at evalModule (node_modules/@expo/require-utils/build/load.js:324:32)
    at loadModuleSync (node_modules/@expo/require-utils/build/load.js:414:10)
    at evalConfig (node_modules/@expo/config/build/evalConfig.js:42:50)
```

same project, same node_modules, only the node binary swapped:

- node 24.19.0: ok
- node 26.7.0: throws

with no typescript installed `loadTypescript()` returns null, so require-utils falls back to node's stripper and calls:

```js
stripTypeScriptTypes(code, { mode: 'transform', sourceMap: true })
```

node 26 removed both of those:

- `mode: 'transform'` in 26.0.0, nodejs/node#61803
- `sourceMap: true` in 26.4.0, nodejs/node#63738

fixing mode alone isn't enough, sourceMap then throws the same way. calling with defaults works on 26 against the same unmodified file:

```js
stripTypeScriptTypes(code) // ok
```

same thing happens with typescript 7 installed, since it has no `transpileModule` and hits the same null branch from #47759. i left typescript out of the repro so doctor stays clean.

still hardcoded in 57.0.4 and in 58.0.0-canary-20260806.

happy to send a PR feature-detecting `transform` if that's the direction you want. only caveat is strip mode can't handle `enum` or `namespace`, but node removed transform entirely so there's no keeping it through this API.

### Environment

```text
  expo-env-info 1.2.2 environment info:
    System:
      OS: macOS 26.2
    Binaries:
      Node: 26.7.0 - /opt/homebrew/Cellar/node/26.7.0/bin/node
      npm: 11.17.0 - /opt/homebrew/Cellar/node/26.7.0/bin/npm
    npmPackages:
      expo: ~57.0.11 => 57.0.11
      react: 19.2.3 => 19.2.3
    Expo Workflow: managed
```

### Expo Doctor Diagnostics

```text
```

kkkhs:
# Why
Fixes #48725. Node 26 rejects the TypeScript stripping fallback used when `typescript` is unavailable.

# How
Add `isInvalidStripTypeScriptOptionsError` and route fallback stripping through `stripTypeScriptTypes()`, which retries without the rejected options.

# Test Plan
Added Node 26 fallback rejection coverage.

# Checklist

- [x] I added a `changelog.md` entry and rebuilt the package sources according to the contributing guide.

### DIFF
--- a/packages/@expo/require-utils/CHANGELOG.md
+++ b/packages/@expo/require-utils/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### 🐛 Bug fixes
 
+- Fix Node.js 26 fallback loading for TypeScript files when `typescript` is unavailable.
+
 ### 💡 Others
 
 ## 57.0.4 - 2026-07-22
--- a/packages/@expo/require-utils/CHANGELOG.md
+++ b/packages/@expo/require-utils/CHANGELOG.md
@@ -8,9 +8,9 @@
 
 ### 🐛 Bug fixes
 
-- Fix Node.js 26 fallback loading for TypeScript files when `typescript` is unavailable.
-
 ### 💡 Others
+
+- Support Node 26's `stripTypeScriptTypes` and call without transform-mode, to fix the fallback when TypeScript isn't installed.
 
 ## 57.0.4 - 2026-07-22
 
--- a/packages/@expo/require-utils/CHANGELOG.md
+++ b/packages/@expo/require-utils/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ### 💡 Others
 
-- Support Node 26's `stripTypeScriptTypes` and call without transform-mode, to fix the fallback when TypeScript isn't installed.
+- Support Node 26's `stripTypeScriptTypes` and call without transform-mode, to fix the fallback when TypeScript isn't installed ([#48826](https://github.com/expo/expo/pull/48826) by [@kkkhs](https://github.com/kkkhs))
 
 ## 57.0.4 - 2026-07-22
 
--- a/packages/@expo/require-utils/src/load.ts
+++ b/packages/@expo/require-utils/src/load.ts
@@ -278,6 +278,28 @@ function containsModuleSyntax(code: string): boolean {
 
 const hasStripTypeScriptTypes = typeof nodeModule.stripTypeScriptTypes === 'function';
 
+function isInvalidStripTypeScriptOptionsError(error: any): boolean {
+  return (
+    error?.code === 'ERR_INVALID_ARG_VALUE' &&
+    typeof error?.message === 'string' &&
+    /\boptions\.(?:mode|sourceMap)\b/.test(error.message)
+  );
+}
+
+function stripTypeScriptTypes(code: string): string {
+  try {
+    return nodeModule.stripTypeScriptTypes(code, {
+      mode: 'transform',
+      sourceMap: true,
+    });
+  } catch (error: any) {
+    if (!isInvalidStripTypeScriptOptionsError(error)) {
+      throw error;
+    }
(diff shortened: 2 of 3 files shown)


brentvatne:
@expo-bot review

packages/@expo/require-utils/src/load.ts
```diff
@@ -278,6 +278,28 @@ function containsModuleSyntax(code: string): boolean {
 
 const hasStripTypeScriptTypes = typeof nodeModule.stripTypeScriptTypes === 'function';
 
+function isInvalidStripTypeScriptOptionsError(error: any): boolean {
+  return (
+    error?.code === 'ERR_INVALID_ARG_VALUE' &&
+    typeof error?.message === 'string' &&
+    /\boptions\.(?:mode|sourceMap)\b/.test(error.message)
```
kitten: we'd probably want to capture the error message here to remember what we're matching against in a comment. I think the regex is potentially a bit permissive and we could also alternatively just be a bit more exact, or trust that `ERR_INVALID_ARG_VALUE` as a check is enough

packages/@expo/require-utils/src/load.ts
```diff
@@ -278,6 +278,28 @@ function containsModuleSyntax(code: string): boolean {
...
+  );
+}
+
+function stripTypeScriptTypes(code: string): string {
+  try {
+    return nodeModule.stripTypeScriptTypes(code, {
```
kitten: I think it'd be preferable to do a more exact check here. It's redundant for us to assume that `stripTypeScriptTypes` will be the legacy shape if we have other ways to determine whether it's likely to succeed or not.

The most likely signal (although primitive) is likely to check a max Node version, similar to: https://github.com/expo/expo/blob/20be1d5b20077a79c1f535dfabe24d55f296d77b/packages/%40expo/cli/src/index.ts#L94-L98

e.g., assuming this was really dropped in 26.0.0 not 25 or a minor, `Number(process.version?.slice(1).split('.', 1)[0])` would probably do to check against just the major

packages/@expo/require-utils/CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 ### 🐛 Bug fixes
 
+- Fix Node.js 26 fallback loading for TypeScript files when `typescript` is unavailable.
```
kitten: As per the contributing guide, we'd need the changelog entry to follow the format, as below. Happy to update this if you don't get to it, just let me know.

Also a nit, let's move this to "Others" and write:

```
- Support Node 26's `stripTypeScriptTypes` and call without transform-mode, to fix the fallback when TypeScript isn't installed
```

Main reason here is that this is a "support-addition", not a fix, imho

brentvatne:
@expo-bot check --fable

kitten (approved):

PR #48828, opened 2026-08-12
[docs] State that Free plan accounts cannot incur overage charges

amandeepmittal:
# Why

Fix ENG-25825

# How

State that Free plan accounts cannot incur overage charges in relevant places in Billing docs.

# Test Plan

Proofread.

# Checklist

### DIFF
--- a/docs/pages/billing/faq.mdx
+++ b/docs/pages/billing/faq.mdx
@@ -34,6 +34,10 @@ To upgrade from the Free plan to the Starter plan, see [Upgrade to a new plan](/
 
 To downgrade from the Starter to a Free plan, see [Cancel a plan](/billing/manage/#cancel-a-plan).
 
+### What happens when I use up the Free plan's monthly quota?
+
+Free plan accounts are not charged for overage. Once you use your monthly quota of free builds, new builds are unavailable until the quota resets on the first day of the next calendar month. To keep building without waiting for the reset, upgrade to the [Starter plan](/billing/plans/) or [run builds locally](/build-reference/local-builds/).
+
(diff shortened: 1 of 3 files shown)


docs/pages/billing/faq.mdx
```diff
@@ -34,6 +34,10 @@ To upgrade from the Free plan to the Starter plan, see [Upgrade to a new plan](/
 
 To downgrade from the Starter to a Free plan, see [Cancel a plan](/billing/manage/#cancel-a-plan).
 
+### What happens when I use up the Free plan's monthly quota?
+
+Free plan accounts are not charged for overage. Once you use your monthly quota of free builds, new builds are unavailable until the quota resets on the first day of the next calendar month. To keep building without waiting for the reset, upgrade to the [Starter plan](/billing/plans/) or [run builds locally](/build-reference/local-builds/).
```
entiendoNull: First sentence feels incomplete to me.

```suggestion
Free plan accounts do not incur overage charges. Once you use your monthly quota of free builds, new builds are unavailable until the quota resets on the first day of the next calendar month. To keep building without waiting for the reset, upgrade to the [Starter plan](/billing/plans/) or [run builds locally](/build-reference/local-builds/).
```

TIL:
> Once you use your monthly quota of free builds, new builds are unavailable until the quota resets on the first day of the next calendar month. 

I actually did not know this was happening on the first day of the calendar month for free accounts. I've always just referred to this as to be reset on at the start of the next billing period (which is still correct, but my thinking of when this is has been different). Makes sense though!

entiendoNull (approved):

docs/pages/billing/faq.mdx
amandeepmittal: aren't both the same thing?

PR #48830, opened 2026-08-12
[docs] Note that barcodeScannerEnabled requires building expo-camera from source

amandeepmittal:
# Why

Fix ENG-25851

# How

Note that `barcodeScannerEnabled` requires building `expo-camera` from source in the config plugin definition. Also, note a general guidance about the same in prebuilt modules FAQ.

# Test Plan

Proofread.

# Checklist

### DIFF
--- a/docs/pages/versions/unversioned/sdk/camera.mdx
+++ b/docs/pages/versions/unversioned/sdk/camera.mdx
@@ -77,7 +77,7 @@ You can configure `expo-camera` using its built-in [config plugin](/config-plugi
     {
       name: 'barcodeScannerEnabled',
       description:
-        'A boolean that determines whether to enable barcode scanning support. Disabling this reduces app size when barcode scanning is not needed.',
+        'A boolean that determines whether to enable barcode scanning support. Disabling this reduces app size when barcode scanning is not needed. On Android, this option only takes effect when `expo-camera` builds from source. [Prebuilt modules](/guides/prebuilt-expo-modules/) already include the barcode scanning libraries. To make this option apply, add `expo-camera` to `buildFromSource` in **package.json**.',
       default: 'true',
     },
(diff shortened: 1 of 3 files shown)


docs/pages/versions/unversioned/sdk/camera.mdx
```diff
@@ -77,7 +77,7 @@ You can configure `expo-camera` using its built-in [config plugin](/config-plugi
     {
       name: 'barcodeScannerEnabled',
       description:
-        'A boolean that determines whether to enable barcode scanning support. Disabling this reduces app size when barcode scanning is not needed.',
+        'A boolean that determines whether to enable barcode scanning support. Disabling this reduces app size when barcode scanning is not needed. On Android, this option only takes effect when `expo-camera` builds from source. [Prebuilt modules](/guides/prebuilt-expo-modules/) already include the barcode scanning libraries. To make this option apply, add `expo-camera` to `buildFromSource` in **package.json**.',
```
kadikraman, line 80: iOS now also has precompiled frameworks, so similar advice might apply for ios?

amandeepmittal: After inespecting it with claude: 

 It turns out iOS does not need it: barcode scanning ships as a separate companion pod (ExpoCameraBarcodeScanning), and autolinking includes or skips it at pod install based on the Podfile property, for both precompiled and source builds (see precompiled_modules.rb's companion pod flow). Only Android bakes the barcode dependencies into the prebuilt artifact. I scoped the prebuilt guide's example to Android to make this explicit.

kadikraman: Thanks for confirming!

kadikraman (approved):

PR #39793, opened 2025-09-18
[iOS][ui] - add shapes in Expo UI

intergalacticspacehighway:
# Why

- Adds `Rectangle`, `RoundedRectangle`, `UnevenRoundedRectangle`, `Circle`, `Ellipse`, `Capsule` shape components. 

# How

https://developer.apple.com/documentation/swiftui/shapes

It follows the pattern of keeping constructor arguments of SwiftUI View as props in React components with same names. Following this pattern also helps AI to write better Expo UI code as it follows SwiftUI conventions.

# Test Plan

Test newly added Shapes screen.

https://github.com/user-attachments/assets/0214255c-b407-4f8f-879a-ce741c3128f3

# Checklist

- [x] I added a `changelog.md` entry and rebuilt the package sources according to [this short guide](https://github.com/expo/expo/blob/main/CONTRIBUTING.md#-before-submitting)
- [x] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: updated a module plugin).
- [x] Conforms with the [Documentation Writing Style Guide](https://github.com/expo/expo/blob/main/guides/Expo%20Documentation%20Writing%20Style%20Guide.md)

PR #48838, opened 2026-08-12
[ui][ios] - Add a geometryGroup modifier

intergalacticspacehighway:
# Why?

Closes [ENG-25710](https://linear.app/expo/issue/ENG-25710/add-a-geometrygroup-modifier-to-expoui-without-it-animated-buttons).

`geometrygroup` allows grouping the label and Button as a single unit when animating, which fixes the Text layout issue described above.

# How?

Adds `geometrygroup` modifier.

# Test

- Tested the repro shared in linear. The Text layout gets fixed when the modifier is applied.
- Added a screen which demonstrates this modifier.


packages/expo-ui/src/swift-ui/modifiers/index.ts
```diff
@@ -88,6 +88,28 @@ export const matchedGeometryEffect = (
...
+ * By default SwiftUI views push position and size changes down through the view hierarchy, so that
+ * only views that draw something (leaf views) apply the current animation to their frame rectangle.
+ * In some cases this gives undesirable results. This modifier resolves the geometry of the view into
+ * a single value, and the subviews of that view resolve their geometry relative to that value.
+ *
+ * @example
+ * ```tsx
+ * <VStack modifiers={[animation(Animation.spring(), isBusy)]}>
+ *   {isBusy ? <Text>Working…</Text> : null}
+ *   <Button modifiers={[geometryGroup()]}>Check now</Button>
```
Kudo: [💡 suggestion] `Button`'s `children` is typed `React.ReactElement | React.ReactElement[]`, so a bare text child doesn't type-check. Text labels use the `label` prop (see the `Button` TSDoc example). This snippet is published on the docs site, so it's worth matching the supported API.

```suggestion
 *   <Button label="Check now" modifiers={[geometryGroup()]} />
```

apps/native-component-list/src/screens/UI/GeometryGroupScreen.ios.tsx
```diff
@@ -0,0 +1,65 @@
...
+          <Text>
+            Tap the button to insert and remove the banner. With geometryGroup off, the button label
+            leaves the button while the spring runs.
+          </Text>
+
+          <VStack
+            spacing={12}
+            modifiers={[animation(Animation.spring({ duration: 1.6, bounce: 0.3 }), busy)]}>
+            {busy ? (
+              // TODO: Add Rectangle component
```
Kudo: [💡 suggestion] `Rectangle` is already exported from `@expo/ui/swift-ui` (added in #39793, used in `GridScreen.ios.tsx` and `MaskScreen.ios.tsx`), so this TODO and the `@ts-expect-error` on an empty `HStack` aren't needed:

```tsx
<Rectangle
  modifiers={[
    frame({ maxWidth: Infinity, height: 150 }),
    foregroundStyle('#FFD60A'),
    cornerRadius(18),
  ]}
/>
```

`Rectangle` paints with the foreground style, so `background` becomes `foregroundStyle`; the imports need `Rectangle` and `foregroundStyle` instead of `background`.

packages/expo-ui/src/swift-ui/modifiers/index.ts
```diff
@@ -88,6 +88,28 @@ export const matchedGeometryEffect = (
   }
 ) => createModifier('matchedGeometryEffect', { id, namespaceId, ...options });
 
+/**
+ * Isolates the geometry, such as position and size, of the view from its parent view.
+ *
+ * By default SwiftUI views push position and size changes down through the view hierarchy, so that
```
amandeepmittal:
```suggestion
 * By default, SwiftUI views push position and size changes down through the view hierarchy, so that
```

packages/expo-ui/src/swift-ui/modifiers/index.ts
```diff
@@ -88,6 +88,28 @@ export const matchedGeometryEffect = (
...
+ * In some cases this gives undesirable results. This modifier resolves the geometry of the view into
+ * a single value, and the subviews of that view resolve their geometry relative to that value.
+ *
+ * @example
+ * ```tsx
+ * <VStack modifiers={[animation(Animation.spring(), isBusy)]}>
+ *   {isBusy ? <Text>Working…</Text> : null}
+ *   <Button modifiers={[geometryGroup()]}>Check now</Button>
+ * </VStack>
+ * ```
```
amandeepmittal, line 100: The `@example` won't type-check since `Button`'s `children` accepts `React.ReactElement`, not a string.

Kudo (approved):

packages/expo-ui/src/swift-ui/modifiers/index.ts
intergalacticspacehighway: Updated

packages/expo-ui/src/swift-ui/modifiers/index.ts
intergalacticspacehighway: Good catch, fixed!

packages/expo-ui/src/swift-ui/modifiers/index.ts
intergalacticspacehighway: Done

amandeepmittal (approved):

PR #48839, opened 2026-08-12
fix(fingerprint): Set NODE_ENV before loading Expo config

ramonclaudio:
# Why

Fingerprint can load different env files for the same project when `NODE_ENV` changes.

# How

I updated Fingerprint to use the shared API and set the mode to `development` before loading Expo config and env files.

# Test Plan

Tests and CI checks pass.

# Checklist

### DIFF
--- a/packages/@expo/fingerprint/src/ExpoConfig.ts
+++ b/packages/@expo/fingerprint/src/ExpoConfig.ts
@@ -1,4 +1,5 @@
 import type { ProjectConfig } from '@expo/config';
+import { getOriginalEnv, setNodeEnv } from '@expo/env';
 import fs from 'fs/promises';
 import os from 'os';
 import path from 'path';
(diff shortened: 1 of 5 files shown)


kitten (approved):

packages/@expo/fingerprint/src/ExpoConfig.ts
```diff
@@ -60,11 +63,19 @@ async function spawnConfigLoaderAsync(
...
+    ignoredFile,
+    '--mode',
+    FINGERPRINT_ENV_MODE,
+  ];
   if (skipPlugins) {
     args.push('--skipPlugins');
   }
-  const { message } = await spawnWithIpcAsync('node', args, { cwd: projectRoot });
+  const childEnv = setNodeEnv(FINGERPRINT_ENV_MODE, { systemEnv: getOriginalEnv() });
+  delete childEnv.EXPO_CONFIG_MODE;
```
Kudo: a quick question and not a blocker: where's the `EXPO_CONFIG_MODE` being set? any chance we could avoid the `delete env.EXPO_CONFIG_MODE` on every callsite?

Kudo (approved):

packages/@expo/fingerprint/src/ExpoConfig.ts
ramonclaudio: Yea, Phil mentioned this too. I moved the cleanup into `loadProjectEnv` in #48938, so once that merges I’ll remove it from here.

Issue #44979, opened 2026-04-21
[expo-audio] AudioRecordingRequester.swift: cannot find 'RCTFatal' in scope — iOS New Architecture (SDK 55)


MarcosT96:
## Description                          
                                                                                                                                                                 
  Building locally with New Architecture enabled (`newArchEnabled: true`) fails to compile                                                                       
  `expo-audio` on iOS with the following errors:                                                                                                                 
                                                                                                                                                                 
  ❌ (node_modules/expo-audio/ios/AudioRecordingRequester.swift:15:7)                                                                                            
  ^ cannot find 'RCTFatal' in scope                                                                                                                            
                                                                                                                                                                 
  ❌ (node_modules/expo-audio/ios/AudioRecordingRequester.swift:15:16)                                                                                           
  ^ cannot find 'RCTErrorWithMessage' in scope                                                                                                                 
                                                                                                                                                                 
  ## Environment                          
                                                                                                                                                               
  | Package | Version |                                                                                                                                          
  |---|---|
  | `expo` | 55.0.16 |                                                                                                                                           
  | `expo-audio` | 55.0.14 |              
  | `react-native` | 0.83.6 |                                                                                                                                    
  | `expo-modules-core` | 55.0.23 |
  | `newArchEnabled` | `true` |                                                                                                                                  
  | Xcode | 16.x |                        
  | Platform | iOS |                                                                                                                                             
                                                                                                                                                                 
  ## Steps to reproduce                                                                                                                                        
                                                                                                                                                                 
  1. Enable New Architecture in `app.config.js`: `newArchEnabled: true`                                                                                          
  2. Run `pod install`                                                                                                                                         
  3. Build for iOS locally                                                                                                                                       
                                          
  ## Root cause                                                                                                                                                

  `AudioRecordingRequester.swift` uses `RCTFatal` and `RCTErrorWithMessage` which are                                                                            
  C++ functions not accessible from Swift in the New Architecture context, even with
  `internal import React` at the top of the file.                                                                                                                
                                                                                                                                                                 
  ## Workaround                                                                                                                                                  
                                                                                                                                                                 
  Replace the `RCTFatal` call with a native Swift `fatalError`:                                                                                                  
                                                                                                                                                               
  ```swift                                                                                                                                                       
  // Before                               
  guard (Bundle.main.infoDictionary?["NSMicrophoneUsageDescription"]) != nil else {                                                                            
    RCTFatal(RCTErrorWithMessage("""                                                                                                                             
[...]

PR #44248, opened 2026-03-25
[iOS][precompile] update expo modules to support being built with SPM

chrfalch:
# Why

This PR is a preparation for using expo-modules-core in an XCFramework built with SPM. The main issue we're facing is that with SPM we need to split c++, swift and objective-c in separate targets, meaning that swift can reference types from objective-c, but obj-c cannot reference swift.

More info about the SPM/XCFramework project here:
https://www.notion.so/expo/SPM-for-Expo-Modules-on-iOS-2dae5b573524802597d1c28d04381409

# How

- Extract ObjC protocols (EXAppContextProtocol, EXReactDelegateProtocol, EXAppContextFactoryRegistry) to enable cross-xcframework boundaries.
- Refactor legacy module registry and native modules proxy to avoid direct Swift header imports. Restructure Fabric view and JSI headers for xcframework compatibility. Update podspecs for xcframework builds.
- Add explicit `import React` where RCT functions are used directly (tracking-transparency, camera, calendar, audio, etc.)
- Use `internal import` import for SDWebImage in expo-image to hide third-party dependency from the public module interface
- Add missing Foundation.h imports in unimodules-app-loader headers
- Restructure expo AppDelegate imports for prebuilt React compatibility
- Update Fetch module to work across xcframework boundaries

# Test Plan

Build with Bare Expo with and without React Native from source as usual. Should work (note: Xcode 26.4 issue with build from source, is not coming from this PR).

# Checklist

- [x] I added a `changelog.md` entry and rebuilt the package sources according to [this short guide](https://github.com/expo/expo/blob/main/CONTRIBUTING.md#-before-submitting)
- [x] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: updated a module plugin).

Issue #33761, opened 2024-12-19
[expo-audio] Microphone permission is required even if the app does not use the microphone preventing app store submission


evelant:
### Minimal reproducible example

N/A - any app using expo-audio

### What platform(s) does this occur on?

iOS

### Where did you reproduce the issue?

in a development build

### Summary

Submit any app using expo-audio to App Store Connect and it will be rejected with 

> ITMS-90683: Missing purpose string in Info.plist - Your app’s code references one or more APIs that access sensitive user data, or the app has one or more entitlements that permit such access. The Info.plist file for the “TaskHero.app” bundle should contain a NSMicrophoneUsageDescription key with a user-facing purpose string explaining clearly and completely why your app needs the data.  If you’re using external libraries or SDKs, they may reference APIs that require a purpose string. While your app might not use these APIs, a purpose string is still required.

Our app does not access the microphone and I do not want to scare away potential users with a label on the app store listing that says microphone permissions are used. Is it possible to update the config plugin to avoid all microphone api access to prevent this for apps that don't use the microphone?

### Environment

```text
expo-env-info 1.2.1 environment info:
    System:
      OS: macOS 15.1.1
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 20.18.1 - /opt/homebrew/opt/node@20/bin/node
      npm: 10.8.2 - /opt/homebrew/opt/node@20/bin/npm
    Managers:
      CocoaPods: 1.15.2 - /Users/imagio/.rvm/gems/ruby-3.1.0/bin/pod
    SDKs:
      iOS SDK:
        Platforms: DriverKit 24.2, iOS 18.2, macOS 15.2, tvOS 18.2, visionOS 2.2, watchOS 11.2
      Android SDK:
        API Levels: 33
        Build Tools: 33.0.0
    IDEs:
      Android Studio: EAP AI-242.23726.103.2422.12630895 AI-242.23726.103.2422.12630895
      Xcode: 16.2/16C5032a - /usr/bin/xcodebuild
    Expo Workflow: bare
```
```


### Expo Doctor Diagnostics

```text
Enabled experimental React Native Directory checks. Unset the EXPO_DOCTOR_ENABLE_DIRECTORY_CHECK environment variable to disable this check.
✔ Check package.json for common issues
✔ Check Expo config for common issues
✔ Check for common project setup issues
✔ Check dependencies for packages that should not be installed directly
✔ Check if the project meets version requirements for submission to app stores
[...]

PR #48840, opened 2026-08-12
[expo-audio] Don't crash the app when NSMicrophoneUsageDescription is missing

ahmadaccino:
# Why

An app that only plays audio has no reason to declare microphone access, and `expo-audio` supports exactly that: the config plugin's `microphonePermission: false` omits `NSMicrophoneUsageDescription` from `Info.plist`, and the docs present that as a normal option.

In that supported configuration, asking iOS *whether* recording is allowed terminates the app. `getRecordingPermissionsAsync()` is a read-only status query — it should be able to answer "no". Instead the app dies, and it dies at the point of the question rather than at any attempt to record, so a library or dependency probing for capability takes the whole app down with it.

The intended answer is already written: the very next line returns a `denied` status, and it is unreachable today. Recovering is impossible for the caller, since a fatal is not catchable from JS — a playback-only app must either declare a permission it never uses (and defend that in App Store review, see #33761) or patch the module.

Concretely, `AudioRecordingRequester.getPermissions()` calls `RCTFatal` in the guard for the missing usage description, then returns `EXPermissionStatusDenied`. `requestPermissions` also routes its resolution through `getPermissions()`, so the request path inherits the same fatal.

# How

- Replace the `RCTFatal` with `log.error`, keeping the same guidance and pointing at the config plugin option, and fall through to the existing `denied` return.
- Return the status as a raw value in that branch. It previously returned the `EXPermissionStatus` enum itself while the mapped branch returns `status.rawValue`, so the two exits disagreed on shape.
- Split the mapping into an internal `permissions(systemStatus:usageDescription:)` so both branches are unit-testable without touching `Bundle.main`.
- Drop `internal import React`, now unused in this file (it was added for `RCTFatal`/`RCTErrorWithMessage` in #44248).

`requestPermissions` is otherwise untouched. When the key really is missing, iOS itself aborts as the prompt is presented; that is Apple's contract and not something this module should paper over.

This is deliberately scoped to `expo-audio`. Other requesters (for example `CameraPermissionRequester`) use the same fatal, but for them a missing usage string means the module cannot work at all — whereas `expo-audio` ships a documented mode where the key is absent by design.

# Test Plan

- Added `AudioRecordingRequesterTests` (Swift Testing, `ios/Tests`, guarded for tvOS where the requester is excluded from the podspec): missing usage description reports `denied`, and each `AVAudioSession.RecordPermission` maps to the matching `EXPermissionStatus` raw value. Before this change the first case aborted the test run instead of failing.
- Built the `ExpoAudio` pod for `iphonesimulator` in a bare SDK 57 app (New Architecture, RN 0.86.2) with this file swapped in: compiles clean, and the app builds and runs.
- Verified in the same app that unmodified `57.0.3` compiles too, so this is not a build fix — `internal import React` already resolved the earlier `cannot find 'RCTFatal' in scope` failure (#44979).

# Checklist

- [x] Conforms to the [Expo Contributing Guide](https://github.com/expo/expo/blob/main/CONTRIBUTING.md)
- [x] Added a changelog entry
- [x] Updated the `microphonePermission` docs to state what happens when the permission is disabled
- [x] Added unit tests


brentvatne:
@expo-bot review

brentvatne:
@expo-bot verify

ahmadaccino:
@brentvatne  lmk if you have any questions or want me to adjust anything about the approach, had to write a lil patch to have this not panic in an app of mine. thought i'd make a pr for it

alanjhughes (approved):

PR #48845, opened 2026-08-12
fix(expo-doctor): Use the same mode for env config

ramonclaudio:
# Why

Expo Doctor can use different modes for env files and Expo config, and it also passes loaded env values to `expo install --check`.

# How

I updated Expo Doctor to use the shared API with the same mode for both, and stopped passing loaded env values.

# Test Plan

Tests and CI checks pass.

# Checklist

### DIFF
--- a/packages/expo-doctor/src/checks/InstalledDependencyVersionCheck.ts
+++ b/packages/expo-doctor/src/checks/InstalledDependencyVersionCheck.ts
@@ -41,7 +41,7 @@ export class InstalledDependencyVersionCheck implements DoctorCheck {
       try {
         commandResult = await spawnExpoCLI(projectRoot, ['install', '--check', '--json'], {
           stdio: 'pipe',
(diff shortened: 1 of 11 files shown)


packages/expo-doctor/src/checks/InstalledDependencyVersionCheck.ts
```diff
@@ -61,7 +61,7 @@ export class InstalledDependencyVersionCheck implements DoctorCheck {
       try {
         await spawnExpoCLI(projectRoot, ['install', '--check'], {
           stdio: 'pipe',
-          env: { ...process.env, CI: '1', EXPO_DEBUG: '0' },
+          env: { CI: '1', EXPO_DEBUG: '0' },
```
kitten, line 64: nice catch!

kitten (approved):

kadikraman (approved):

entiendoNull (approved):

PR #48857, opened 2026-08-13
[github] Create bot-fork branches at the upstream SHA

brentvatne:
## Summary

`POST /repos/expo-bot/expo/merge-upstream` cannot work with a `public_repo` PAT that lacks `workflow`. The fork is 43 commits behind, those commits include `.github/workflows/**`, and syncing `main` is exactly the update GitHub refuses (run [31664540660](https://github.com/expo/expo/actions/runs/31664540660)).

This stops moving the fork's default branch. The runner already has current `expo/expo` `main`; that SHA is already in the fork's object store (parent network). We create the publish branch at that SHA via `POST /git/refs`, then `git push` only the agent commit (path-denylisted from `.github/**`).

Preflight does the same with a throwaway `agent-preflight-$RUN_ID` ref and deletes it.

Supersedes the diagnostic-only approach in #48856 for this failure.

## Test plan

- [ ] `bash .github/scripts/ensure-fork-branch.sh --self-test`
- [ ] With an expo-bot PAT that can write `expo-bot/expo` and does **not** have `workflow`:
  - `GET /repos/expo-bot/expo/git/commits/<current expo/expo main>` succeeds
  - `POST /git/refs` creating `refs/heads/agent-probe-<n>` at that SHA succeeds
  - `DELETE` that ref
  - a `git push` of one non-workflow commit onto a branch created that way succeeds
- [ ] After merge, `verify <issue>` (fix mode) gets past preflight

Do **not** add `workflow` to `EXPO_BOT_GITHUB_TOKEN` to make this pass.

### DIFF
--- a/.github/scripts/ensure-fork-branch.sh
+++ b/.github/scripts/ensure-fork-branch.sh
@@ -1,0 +1,69 @@
+#!/usr/bin/env bash
+# Point a new branch on a fork at a commit GitHub already has (the parent
+# network). Does not merge-upstream and does not send a pack.
+#
+# merge-upstream (and `git push` of a SHA that is  N workflow-changing
+# commits ahead of the fork's default branch) needs the `workflow` scope.
+# Creating a ref at that SHA, then pushing one commit that does not touch
+# .github/workflows/**, does not.
+set -euo pipefail
+
+if [ "${1:-}" = --self-test ]; then
+  tmp=$(mktemp)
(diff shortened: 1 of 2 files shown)


brentvatne:
Superseded by the dedicated sync-on-main job. A PAT probe showed POST /git/refs at current expo/expo main is 404 for the same workflow-scope check as merge-upstream (422). See the replacement PR.

PR #48861, opened 2026-08-13
[app-metrics][iOS][Android] Add a spans table to the metrics database

tsapeta:
# Why

The Observe backend can ingest OTLP traces now (expo/universe#29872, expo/universe#29874), but the SDK has nowhere to store spans locally. This adds the storage layer only; producers and export sit on top.

Part of ENG-25894.

# How

Adds a generic `spans` table mirroring the OTLP span shape, plus the read/delete facade the exporter consumes. Spans are attributed to the session that produced them and stay readable per session, so a session's trace activity can be inspected alongside its metrics and logs; consumers delete only rows no reader still needs. Trace and span ids are generated at record time, so redelivery after a crash stays byte-identical on the server. The table is capped at 2000 rows in the same transaction as the insert. Timestamps are millisecond integers rather than the house ISO strings, which carry no fractional seconds and would collapse sub-second spans.

On iOS the table is created with `IF NOT EXISTS` on every open, no schema-version bump, so existing data survives. On Android, Room bumps to v18 and relies on the existing destructive fallback: maintaining a migration costs more than losing local telemetry that is dispatched frequently anyway.

# Test Plan

Unit suites on both platforms cover id generation, storage round-trips, ordering, the per-session read, the cap, cascade deletion, and the iOS no-wipe upgrade path. `et native-unit-tests` passes on both platforms with only this layer present.

### DIFF
--- a/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/storage/MetricsDatabase.kt
+++ b/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/storage/MetricsDatabase.kt
@@ -12,6 +12,8 @@ import androidx.room.Query
 import androidx.room.ForeignKey
 import androidx.room.Room
 import androidx.room.RoomDatabase
+import androidx.room.migration.Migration
+import androidx.sqlite.db.SupportSQLiteDatabase
 import androidx.room.Embedded
 import androidx.room.Relation
 import androidx.room.Transaction
@@ -22,9 +24,31 @@ object MetricsConstants {
   const val SECONDS_TO_REMOVE_OLD_METRICS: Long = 7 * 24 * 60 * 60 // 7 days in seconds
 }
 
+/**
+ * v16 -> v17: adds the `spans` table. Hand-written because schema export is off (so Room's
+ * auto-migrations aren't available); the DDL is copied verbatim from the KSP-generated
+ * `MetricsDatabase_Impl.createAllTables`, and Room validates the migrated schema against the
+ * entities on open, so any drift between this SQL and the `Span` entity fails fast instead of
+ * corrupting silently. Registered so existing installs keep their sessions, metrics, and logs
+ * on upgrade instead of hitting the destructive fallback.
+ */
+internal val MIGRATION_16_17 = object : Migration(16, 17) {
+  override fun migrate(db: SupportSQLiteDatabase) {
+    db.execSQL(
+      "CREATE TABLE IF NOT EXISTS `spans` (`id` INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, " +
+        "`sessionId` TEXT NOT NULL, `traceId` TEXT NOT NULL, `spanId` TEXT NOT NULL, " +
+        "`parentSpanId` TEXT, `name` TEXT NOT NULL, `kind` INTEGER NOT NULL, " +
+        "`startTimestampMs` INTEGER NOT NULL, `endTimestampMs` INTEGER NOT NULL, " +
+        "`statusCode` INTEGER, `statusMessage` TEXT, `attributes` TEXT, `events` TEXT, " +
+        "FOREIGN KEY(`sessionId`) REFERENCES `sessions`(`id`) ON UPDATE NO ACTION ON DELETE CASCADE )"
+    )
+    db.execSQL("CREATE INDEX IF NOT EXISTS `index_spans_sessionId` ON `spans` (`sessionId`)")
+  }
+}
+
 @Database(
-  entities = [Metric::class, LogRecord::class, Session::class, CrashReportEntity::class],
-  version = 16,
+  entities = [Metric::class, LogRecord::class, Session::class, CrashReportEntity::class, Span::class],
+  version = 17,
   exportSchema = false
 )
 abstract class MetricsDatabase : RoomDatabase() {
@@ -35,6 +59,8 @@ abstract class MetricsDatabase : RoomDatabase() {
   abstract fun sessionDao(): SessionDao
 
   abstract fun crashReportDao(): CrashReportDao
+
+  abstract fun spanDao(): SpanDao
 
   companion object {
     @Volatile
@@ -48,7 +74,9 @@ abstract class MetricsDatabase : RoomDatabase() {
             MetricsDatabase::class.java,
             "app_metrics"
           )
-          // Allow destructive migration for schema changes during development. Replace with proper Migration if desired.
+          .addMigrations(MIGRATION_16_17)
+          // Destructive fallback for version jumps without a registered migration (pre-16
+          // installs, downgrades). Upgrades covered by a migration above keep their data.
           .fallbackToDestructiveMigration(false)
           .build()
         INSTANCE = instance
--- a/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/storage/MetricsDatabase.kt
+++ b/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/storage/MetricsDatabase.kt
@@ -23,8 +23,8 @@ object MetricsConstants {
 }
 
 @Database(
-  entities = [Metric::class, LogRecord::class, Session::class, CrashReportEntity::class],
-  version = 17,
+  entities = [Metric::class, LogRecord::class, Session::class, CrashReportEntity::class, Span::class],
+  version = 18,
   exportSchema = false
 )
 abstract class MetricsDatabase : RoomDatabase() {
@@ -35,6 +35,8 @@ abstract class MetricsDatabase : RoomDatabase() {
   abstract fun sessionDao(): SessionDao
 
   abstract fun crashReportDao(): CrashReportDao
+
+  abstract fun spanDao(): SpanDao
 
   companion object {
     @Volatile
@@ -48,7 +50,8 @@ abstract class MetricsDatabase : RoomDatabase() {
             MetricsDatabase::class.java,
             "app_metrics"
           )
-          // Allow destructive migration for schema changes during development. Replace with proper Migration if desired.
+          // Version bumps drop the data instead of migrating: maintaining migrations costs more
+          // than losing local telemetry that is dispatched frequently anyway.
           .fallbackToDestructiveMigration(false)
           .build()
         INSTANCE = instance
--- a/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/storage/SessionManager.kt
+++ b/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/storage/SessionManager.kt
@@ -153,6 +153,16 @@ class SessionManager(
 
   suspend fun getMaxMetricId(): Long? = database.metricDao().getMaxId()
 
+  suspend fun getSpansForSession(sessionId: String): List<Span> =
+    database.spanDao().getSpansForSession(sessionId)
+
+  suspend fun getSpans(afterId: Long, limit: Int): List<Span> =
+    database.spanDao().getSpans(afterId, limit)
+
+  suspend fun getMaxSpanId(): Long? = database.spanDao().getMaxId()
+
+  suspend fun deleteSpansUpTo(rowId: Long) = database.spanDao().deleteUpTo(rowId)
+
   suspend fun getLogsForSession(sessionId: String): List<LogRecord> =
     database.logDao().getLogsForSession(sessionId)
 
--- a/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/storage/Span.kt
+++ b/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/storage/Span.kt
@@ -1,0 +1,159 @@
+// Copyright 2025-present 650 Industries. All rights reserved.
+
+package expo.modules.appmetrics.storage
+
+import androidx.room.Dao
+import androidx.room.Entity
+import androidx.room.ForeignKey
+import androidx.room.Index
+import androidx.room.Insert
+import androidx.room.PrimaryKey
+import androidx.room.Query
+import androidx.room.Transaction
+import java.util.concurrent.ThreadLocalRandom
+
+/**
+ * Persistence-layer representation of one trace span, mirroring the OTLP `Span` shape. Spans are
+ * a generic signal: any producer (network requests today, navigation or custom spans later)
+ * converts its domain event into a `Span` at record time, and `expo-observe` exports pending rows
+ * to the OTLP traces endpoint without knowing what produced them. Mirrors the iOS `SpanRow`.
+ */
+@Entity(
+  tableName = "spans",
+  indices = [Index("sessionId")],
+  foreignKeys = [
+    ForeignKey(
+      entity = Session::class,
+      parentColumns = ["id"],
+      childColumns = ["sessionId"],
+      onDelete = ForeignKey.CASCADE
+    )
+  ]
+)
+data class Span(
+  /**
+   * Auto-incremented rowid. Unlike the other tables' UUID keys, spans use a monotonic id so the
+   * export layer can chunk in insertion order and delete consumed prefixes with one query.
+   */
+  @PrimaryKey(autoGenerate = true) val id: Long = 0,
+  val sessionId: String,
+
+  /**
+   * W3C trace/span identifiers as lowercase hex (32 and 16 characters). Generated once at
+   * record time and persisted, so a redelivered row (export is at-least-once) reaches the
+   * server byte-identical instead of becoming an untraceable duplicate.
+   */
+  val traceId: String = generateTraceId(),
+  val spanId: String = generateSpanId(),
+
+  /**
+   * Parent span id for in-app hierarchies, or `null` for a root span. No producer sets this yet;
+   * the column exists so adding nested spans later is not a schema change.
+   */
+  val parentSpanId: String? = null,
+
+  /** Span name per the semantic conventions of the producer (the HTTP method for network spans). */
+  val name: String,
+
+  /** `SpanKind` per the OTLP proto (`CLIENT_KIND` for network requests). */
+  val kind: Int = INTERNAL_KIND,
+
+  /**
+   * Unix-epoch milliseconds. Millisecond integers rather than the ISO strings the other tables
+   * use: a span's duration is routinely sub-second, and the house ISO format carries no
+   * fractional seconds, so a string timestamp would collapse every fast span to zero length.
+   */
+  val startTimestampMs: Long,
+  val endTimestampMs: Long,
+
+  /** OTLP status code (`STATUS_ERROR`), or `null` when the span completed without one (UNSET). */
+  val statusCode: Int? = null,
+  val statusMessage: String? = null,
+
+  /**
+   * JSON object of span attributes, keyed per the producer's semantic conventions. Stored as an
+   * opaque blob — nothing queries individual attributes locally, and the export layer converts
+   * them generically to the OTLP wire shape.
+   */
+  val attributes: String? = null,
+
+  /** JSON array of span events (`[{name, timeMs?, attributes}]`), or `null` when there are none. */
+  val events: String? = null
+) {
+  companion object {
+    /** `SpanKind` values from the OTLP proto, for producers picking a `kind`. */
+    const val INTERNAL_KIND = 1
+    const val CLIENT_KIND = 3
+
+    /** `Status.code` values from the OTLP proto. UNSET is expressed by a null `statusCode`. */
+    const val STATUS_ERROR = 2
+
+    /**
+     * A new random 16-byte trace id as 32 lowercase hex characters. There is no trace-context
+     * propagation yet, so every span starts its own trace.
+     */
+    fun generateTraceId(): String = generateHexId(words = 2)
+
+    /** A new random 8-byte span id as 16 lowercase hex characters. */
+    fun generateSpanId(): String = generateHexId(words = 1)
+
+    /**
+     * Random identifier built from 8-byte words, hex-encoded. The all-zero value is invalid per
+     * the OTLP spec (the server rejects the span), so it redraws — a branch that is hit once per
+     * 2^64 draws but keeps the invariant explicit.
+     */
+    private fun generateHexId(words: Int): String {
+      while (true) {
+        val drawn = List(words) { ThreadLocalRandom.current().nextLong() }
+        if (drawn.all { it == 0L }) {
+          continue
+        }
+        return drawn.joinToString(separator = "") { "%016x".format(it) }
+      }
+    }
+  }
+}
+
+@Dao
+interface SpanDao {
+  companion object {
+    /**
+     * Maximum number of rows retained in `spans`. Span producers (network requests especially)
+     * can outnumber metrics and logs by orders of magnitude, and when nothing consumes the rows
+     * (`expo-observe` not installed, or dispatch disabled) the session-retention prune alone
+     * would let a busy app accumulate a week of traffic. `insertCapped` prunes past the cap.
+     */
+    const val SPAN_CAP = 2_000
+  }
+
+  @Insert
+  suspend fun insert(span: Span): Long
+
+  /**
+   * Inserts a span and prunes rows older than `SPAN_CAP` in the same transaction — ids are
+   * monotonic, so "older" is simply everything at least `SPAN_CAP` ids behind the row just
+   * inserted.
+   */
+  @Transaction
+  suspend fun insertCapped(span: Span): Long {
+    val insertedId = insert(span)
+    deleteUpTo(insertedId - SPAN_CAP)
+    return insertedId
+  }
+
+  /** All spans in ascending rowid (insertion) order. */
+  @Query("SELECT * FROM spans ORDER BY id ASC")
+  suspend fun getAll(): List<Span>
+
+  /** The largest rowid currently in the table, or `null` when it's empty. */
+  @Query("SELECT MAX(id) FROM spans")
+  suspend fun getMaxId(): Long?
+
+  /**
+   * Deletes rows with `id <= rowId`. Called after a dispatch consumed (or deliberately dropped)
+   * a batch: unlike metrics and logs, no per-session API reads spans back, so dispatched rows
+   * are dead weight.
+   */
+  @Query("DELETE FROM spans WHERE id <= :rowId")
+  suspend fun deleteUpTo(rowId: Long)
+}
--- a/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/storage/Span.kt
+++ b/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/storage/Span.kt
@@ -1,0 +1,180 @@
+// Copyright 2025-present 650 Industries. All rights reserved.
+
+package expo.modules.appmetrics.storage
+
+import androidx.room.Dao
+import androidx.room.Entity
+import androidx.room.ForeignKey
+import androidx.room.Index
+import androidx.room.Insert
+import androidx.room.PrimaryKey
+import androidx.room.Query
+import androidx.room.Transaction
+import java.util.concurrent.ThreadLocalRandom
+
+/**
+ * Persistence-layer representation of one trace span, mirroring the OTLP `Span` shape
+ * (https://opentelemetry.io/docs/concepts/signals/traces/#spans). Spans are a generic signal:
+ * any producer (network requests today, navigation or custom spans later) converts its domain
+ * event into a `Span` at record time, and `expo-observe` exports pending rows to the OTLP traces
+ * endpoint without knowing what produced them. Mirrors the iOS `SpanRow`.
+ */
+@Entity(
+  tableName = "spans",
+  indices = [Index("sessionId")],
+  foreignKeys = [
+    ForeignKey(
+      entity = Session::class,
+      parentColumns = ["id"],
+      childColumns = ["sessionId"],
+      onDelete = ForeignKey.CASCADE
+    )
+  ]
+)
+data class Span(
+  /**
+   * Auto-incremented rowid, like `metrics` and `logs`: a monotonic id lets the export layer
+   * chunk in insertion order and delete consumed prefixes with one query.
+   */
+  @PrimaryKey(autoGenerate = true) val id: Long = 0,
+  val sessionId: String,
+
+  /**
+   * W3C trace/span identifiers as lowercase hex (32 and 16 characters). Generated once at
+   * record time and persisted, so a redelivered row (export is at-least-once) reaches the
+   * server byte-identical instead of becoming an untraceable duplicate.
+   */
+  val traceId: String = generateTraceId(),
+  val spanId: String = generateSpanId(),
+
+  /**
+   * Parent span id for in-app hierarchies, or `null` for a root span. Set by `SpanRecorder` when
+   * a custom span is started with a parent; the network producer always writes root spans.
+   */
+  val parentSpanId: String? = null,
+
+  /** Span name per the semantic conventions of the producer (the HTTP method for network spans). */
+  val name: String,
+
+  /** `SpanKind` per the OTLP proto (`CLIENT_KIND` for network requests). */
+  val kind: Int = INTERNAL_KIND,
+
+  /**
+   * Unix-epoch milliseconds. Millisecond integers rather than the ISO strings the other tables
+   * use: a span's duration is routinely sub-second, and the house ISO format carries no
+   * fractional seconds, so a string timestamp would collapse every fast span to zero length.
+   * Milliseconds is the OTel spec's minimal timestamp precision; the exporter converts to the
+   * wire format's `timeUnixNano`.
+   */
+  val startTimestampMs: Long,
+  val endTimestampMs: Long,
+
+  /** OTLP status code (`STATUS_ERROR`), or `null` when the span completed without one (UNSET). */
+  val statusCode: Int? = null,
+  val statusMessage: String? = null,
+
+  /**
+   * JSON object of span attributes, keyed per the producer's semantic conventions. Stored as an
+   * opaque blob: nothing queries individual attributes locally, and the export layer converts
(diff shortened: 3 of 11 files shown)


packages/expo-app-metrics/ios/NetworkRequests/NetworkRequestPersistence.swift
```diff
@@ -0,0 +1,219 @@
...
+    guard var components = URLComponents(url: url, resolvingAgainstBaseURL: false) else {
+      // An unparseable URL can't prove it carries no secrets; drop the query outright.
+      return url.absoluteString.components(separatedBy: "?").first ?? url.absoluteString
+    }
+    if components.user != nil || components.password != nil {
+      components.user = "REDACTED"
+      components.password = "REDACTED"
+    }
+    if let queryItems = components.queryItems, !queryItems.isEmpty {
+      components.queryItems = queryItems.map { item in
```
tsapeta: [💡 suggestion] This reassigns `queryItems` even when nothing was redacted, so every URL's query goes through a decode/re-encode round trip. That can change the string stored as `url.full`: a value sent as `a%2Bb` decodes to `a+b` and is re-encoded as a literal `+` (allowed in the query charset), which many backends read as a space. Consider rewriting the components only when a sensitive parameter was actually found, and returning `url.absoluteString` untouched otherwise. The Android side has the same shape (`newBuilder().build()` re-canonicalizes even when untouched) but OkHttp preserves the encoded query unless a parameter is modified, so it is only iOS where the common no-redaction path mutates the value.

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/networkrequests/NetworkRequestPersistence.kt
```diff
@@ -0,0 +1,218 @@
...
+    attributes.put("server.port", port)
+  }
+  statusCode?.let { code ->
+    attributes.put("http.response.status_code", code)
+  }
+  semconvProtocolVersion(networkProtocol)?.let { version ->
+    attributes.put("network.protocol.version", version)
+  }
+  requestBytesSent?.let { bytes ->
+    attributes.put("http.request.size", bytes)
```
tsapeta: [💡 suggestion] The two platforms report different quantities under this key. iOS sums header plus body bytes from the transaction metrics, while the Android value is body bytes only (the counting-sink measurement or `contentLength()`). Semconv defines `http.request.size` as the total including headers and `http.request.body.size` as body only, and the same asymmetry applies to `http.response.size`. Since the server extracts these into columns, cross-platform comparisons of the column will be skewed by header overhead on iOS. Worth either documenting the asymmetry as accepted or keying each platform by what it actually measures.

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/networkrequests/NetworkRequestMonitor.kt
```diff
@@ -41,6 +41,53 @@ class NetworkRequestMonitor internal constructor() {
...
+   * under the same lock as `record`, so a concurrent completion is either in the drained
+   * snapshot or persisted by `record`, never both.
+   */
+  fun installPersistence(persistence: NetworkRequestPersistence) {
+    val buffered = synchronized(lock) {
+      this.persistence = persistence
+      if (hasDrainedBackfill) {
+        emptyList()
+      } else {
+        hasDrainedBackfill = true
```
tsapeta: [💡 suggestion] The flag flips before the drain has written anything. `persistBuffered` launches its batch on the module's `modulesQueue`, which is cancelled on a JS reload, so a reload shortly after startup can cancel the batch mid-loop (`insertCapped` is a suspension point) and the buffered requests are then lost for good, since no later install may drain again. At-most-once is a defensible choice versus duplicating rows under a fresh session id, but if you want at-least-once, the flag could be set from inside the coroutine after the loop completes, accepting rare duplicates on that exact race instead.

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/networkrequests/NetworkRequestPersistence.kt
```diff
@@ -0,0 +1,218 @@
...
+    statusCode = if (failed) Span.STATUS_ERROR else null,
+    statusMessage = if (failed) errorDescription else null,
+    attributes = attributes.toString(),
+    events = if (events.length() > 0) events.toString() else null
+  )
+}
+
+/**
+ * Bare protocol version per semconv's `network.protocol.version` ("1.1", "2", "3"), mapped from
+ * the ALPN-style names OkHttp reports ("http/1.1", "h2", "h3"). Unrecognized values pass through
```
tsapeta: [🐧 nit] This KDoc describes `semconvProtocolVersion`, which lives at the bottom of the file; sitting here it reads as a second doc for `KNOWN_HTTP_METHODS`. Move it down to the function (the iOS twin has it placed there).

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/networkrequests/NetworkRequestPersistence.kt
```diff
@@ -0,0 +1,218 @@
...
+  // Case-sensitive per the conventions: an unknown or nonstandard method becomes `_OTHER`
+  // (verbatim value preserved in `http.request.method_original`) and names the span `HTTP`,
+  // so caller-controlled verbs can't mint unbounded span names.
+  val isKnownMethod = method in KNOWN_HTTP_METHODS
+  attributes.put("http.request.method", if (isKnownMethod) method else "_OTHER")
+  if (!isKnownMethod) {
+    attributes.put("http.request.method_original", method)
+  }
+  attributes.put("url.full", redactedUrlFull(url))
+  val parsedUrl = url.toHttpUrlOrNull()
```
tsapeta: [🐧 nit] `redactedUrlFull(url)` on the line above already parses the URL, so this is a second `toHttpUrlOrNull()` on the same string. One parse could feed both the redaction and the host/port attributes. Small, but `toSpan` runs on the OkHttp dispatcher thread for every completed request (and per row in the backfill batch).

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/storage/Span.kt
```diff
@@ -0,0 +1,170 @@
...
+     * Maximum number of rows retained in `spans`. Span producers (network requests especially)
+     * can outnumber metrics and logs by orders of magnitude, and when nothing consumes the rows
+     * (`expo-observe` not installed, or dispatch disabled) the session-retention prune alone
+     * would let a busy app accumulate a week of traffic. `insertCapped` prunes past the cap.
+     */
+    const val SPAN_CAP = 2_000
+  }
+
+  @Insert
+  suspend fun insert(span: Span): Long
```
tsapeta: [💡 suggestion] On iOS the only insert path (`insert(span:)`) enforces the cap, but here the obvious name `insert` is the uncapped one and the cap lives in `insertCapped`. A future producer that picks `insert` silently loses the retention bound, and nothing would catch it (the table just grows until session pruning).

Consider inverting the naming so the safe path is the default: make `insert` the `@Transaction` capped method and rename the raw `@Insert` to something like `insertUncapped`. That also makes the two platforms read the same.

packages/expo-app-metrics/ios/Database/MetricsDatabase.swift
```diff
@@ -559,6 +559,91 @@ final class MetricsDatabase: Sendable {
...
+    var rows: [SpanRow] = []
+    try statement.forEachRow { row in
+      rows.append(SpanRow(row: row))
+    }
+    return rows
+  }
+
+  @AppMetricsActor
+  func getMaxSpanId() throws -> Int64? {
+    return try selectMaxId(table: "spans")
```
tsapeta, line 660: [🐧 nit] `selectMaxId`'s doc comment says the table name is one of "`metrics` / `logs`"; this adds `spans` as a third caller, so that list is now stale. Worth updating while you're here.

tsapeta (commented):

This PR adds a `spans` storage layer to `expo-app-metrics` on iOS and Android. The design is sound and well documented. I verified the load-bearing claims against the surrounding code:

- iOS opens every connection with `PRAGMA foreign_keys = ON` (`SQLiteDatabase.init`), so the `ON DELETE CASCADE` on `spans.sessionId` is active.
- The iOS schema bootstrap runs `CREATE TABLE IF NOT EXISTS` inside `createSchemaIfNeeded` on every open, so the no-version-bump upgrade claim holds, and the version-mismatch wipe path is untouched.
- The iOS `transaction` helper is non-nested and `insert(span:)` is its only caller on this path, so the insert-plus-prune commit is safe.
- The statement wrapper binds `Int`, `Int64`, `String`, and `nil` correctly for the new column types.
- The Android migration DDL matches what Room generates for the `Span` entity (AUTOINCREMENT primary key, FK with CASCADE, `index_spans_sessionId`), and Room validates the migrated table at open. The migration test exercises exactly that path.
- The cap arithmetic is correct on both platforms: ids are monotonic (`AUTOINCREMENT`), so `DELETE ... WHERE id <= insertedId - cap` retains at most `spanCap` rows, and it is a cheap no-op below the cap.
- Both hex-id generators produce fixed-width lowercase output for negative words (`%016x` on a two's-complement `Long`, `%016llx` on `UInt64`), and neither `%x` conversion is locale-sensitive.

Test coverage is thorough: round-trips, ordering, cursor reads, the cap, cascade deletion, and the upgrade path on both platforms.

Two small items below: an API-shape asymmetry on the Android DAO, and a stale doc comment that this PR makes inaccurate.

packages/expo-app-metrics/ios/Database/MetricsDatabase.swift
```diff
@@ -574,6 +574,91 @@ final class MetricsDatabase: Sendable {
...
+    try statement.bindAll([cursor])
+    var rows: [SpanRow] = []
+    try statement.forEachRow { row in
+      rows.append(SpanRow(row: row))
+    }
+    return rows
+  }
+
+  /// Returns the spans attributed to `sessionId`, in ascending id (insertion) order. This is
+  /// the session-association read: spans are retained after dispatch so a session's trace
```
tsapeta: [🏗️ **design**] This claim conflicts with the exporter in #48883. That consumer deletes every dispatched row (`deleteSpans(upToId: highestId)`) and its doc says "nothing else reads the `spans` rows back". So spans are not retained after dispatch, and this read returns only undelivered rows. Also, no code in the stack calls `getSpans(forSessionId:)`, `getSpansForSession`, or their `AppMetrics`/`SessionManager` wrappers. Metrics and logs earn their per-session reads through `SessionSharedObject`; spans have no such consumer yet.

Two consistent options: (a) drop the per-session read surface and the retention language until a session-inspection consumer exists, or (b) keep the surface for a planned `SessionSharedObject.getSpans` and reword these docs to state the real contract: the exporter owns deletion, and per-session reads see only rows not yet dispatched.

packages/expo-app-metrics/ios/AppMetrics.swift
```diff
@@ -102,6 +102,33 @@ public struct AppMetrics {
     return try database?.getMaxLogId() ?? nil
   }
 
+  /// Returns span rows whose `id` is greater than `cursor`, in ascending id order. Empty when
+  /// the database failed to open.
+  @AppMetricsActor
+  public static func getSpans(afterId cursor: Int64) throws -> [SpanRow] {
```
tsapeta: [💡 suggestion] `getMetrics(afterId:limit:)` and `getLogs(afterId:limit:)` take an optional `limit`, and the exporter drains them in chunks through `DispatchLoop`. The span read has no `limit`, and the exporter in #48883 consequently reads the whole table in one call. The 2000-row cap bounds that, but `attributes`/`events` blobs can be sizable, and the asymmetry blocks reusing the chunked loop for spans later. Adding `limit: Int? = nil` here (and `LIMIT` in the two platform queries) is cheap now and keeps the read surface uniform.

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/storage/SessionManager.kt
```diff
@@ -153,6 +153,15 @@ class SessionManager(
...
   suspend fun getMaxMetricId(): Long? = database.metricDao().getMaxId()
 
+  suspend fun getSpansForSession(sessionId: String): List<Span> =
+    database.spanDao().getSpansForSession(sessionId)
+
+  /**
+   * The span read/delete surface, for consumers (the `expo-observe` export layer) that don't
+   * depend on Room and so can't reach the dao through `MetricsDatabase` themselves.
+   */
+  fun spanDao(): SpanDao = database.spanDao()
```
tsapeta: [💡 suggestion] This deviates from the SessionManager pattern: metrics and logs are exposed through narrow wrappers (`getMetrics(afterId, limit)`, `getMaxMetricId()`), never as a raw DAO. Exposing `SpanDao` hands consumers `insertUncapped`, the exact footgun its own doc warns against, and makes the `getSpansForSession` wrapper above redundant. Wrapping the three methods the exporter needs (`getSpans`/`getAll`, `getMaxId`, `deleteUpTo`) matches the existing pattern and keeps the capped insert as the only reachable write. Related: the DAO carries both `getAll()` and `getSpans(cursor)` where `getAll()` equals `getSpans(-1)`; iOS gets by with the cursor read alone, so one of the two could go.

packages/expo-app-metrics/ios/Tests/MetricsDatabaseTests.swift
```diff
@@ -618,6 +618,170 @@ struct MetricsDatabaseTests {
...
+      do {
+        let database = try MetricsDatabase(directoryUrl: directoryUrl)
+        try database.insert(session: makeSessionRow(id: "kept"))
+        try database.database.execute("DROP TABLE spans;")
+      }
+      let database = try MetricsDatabase(directoryUrl: directoryUrl)
+      let session = try database.getSession(id: "kept")
+      #expect(session != nil)
+      try database.insert(span: makeSpanRow(sessionId: "kept"))
+      let requests = try database.getSpans(afterId: -1)
```
tsapeta: [🐧 nit] `requests` looks like a leftover from the network-requests table test; these are spans. Rename to `spans`.

tsapeta (commented):

The storage layer is solid. The iOS no-bump strategy is verified: `createSchemaTables` runs with `IF NOT EXISTS` on every open, and an older build opening the newer file stays at v3 and does not wipe. The Android migration DDL matches Room's generated schema and the test validates it through Room's open-time check. ID generation, cap semantics, and the FK cascade are all correct on both platforms.

The one design concern is the retention story. The docs here promise that spans stay readable per session after dispatch, but the exporter later in this stack (#48883) deletes every dispatched row and states that nothing else reads span rows. The per-session reads added here have no caller anywhere in the stack. Either the docs or the surface should change.

Two smaller suggestions: the cursor read lacks the `limit` parameter that `getMetrics`/`getLogs` have, and the Android facade exposes the raw `SpanDao` (including `insertUncapped`) instead of following the SessionManager wrapper pattern.

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/storage/MetricsDatabase.kt
```diff
@@ -15,16 +15,40 @@ import androidx.room.Relation
...
+    db.execSQL(
+      "CREATE TABLE IF NOT EXISTS `spans` (`id` INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, " +
+        "`sessionId` TEXT NOT NULL, `traceId` TEXT NOT NULL, `spanId` TEXT NOT NULL, " +
+        "`parentSpanId` TEXT, `name` TEXT NOT NULL, `kind` INTEGER NOT NULL, " +
+        "`startTimestampMs` INTEGER NOT NULL, `endTimestampMs` INTEGER NOT NULL, " +
+        "`statusCode` INTEGER, `statusMessage` TEXT, `attributes` TEXT, `events` TEXT, " +
+        "FOREIGN KEY(`sessionId`) REFERENCES `sessions`(`id`) ON UPDATE NO ACTION ON DELETE CASCADE )"
+    )
+    db.execSQL("CREATE INDEX IF NOT EXISTS `index_spans_sessionId` ON `spans` (`sessionId`)")
+  }
```
Ubax: I think that we should just drop the data. I think that the maintenance of the migrations will be more cumbersome and complicated then just loosing some metrics (shouldn't be too much since they are sent in every background and periodically as well)

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/storage/Span.kt
```diff
@@ -0,0 +1,176 @@
...
+import androidx.room.Entity
+import androidx.room.ForeignKey
+import androidx.room.Index
+import androidx.room.Insert
+import androidx.room.PrimaryKey
+import androidx.room.Query
+import androidx.room.Transaction
+import java.util.concurrent.ThreadLocalRandom
+
+/**
```
Ubax, line 15: Can you add link to the spec?

https://opentelemetry.io/docs/concepts/signals/traces/#spans

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/storage/Span.kt
```diff
@@ -0,0 +1,176 @@
...
+      entity = Session::class,
+      parentColumns = ["id"],
+      childColumns = ["sessionId"],
+      onDelete = ForeignKey.CASCADE
+    )
+  ]
+)
+data class Span(
+  /**
+   * Auto-incremented rowid. Unlike the other tables' UUID keys, spans use a monotonic id so the
```
Ubax: I think this comment is no longer true

packages/expo-app-metrics/ios/Database/MetricsDatabase.swift
```diff
@@ -574,6 +574,94 @@ final class MetricsDatabase: Sendable {
...
+  static let spanCap = 2_000
+
+  /// Inserts a single span and returns its rowid (the auto-incremented `id`). Also prunes rows
+  /// older than `spanCap`. Ids are monotonic, so "older" is simply everything at least
+  /// `spanCap` ids behind the row just inserted. Insert and prune share one transaction, so the
+  /// per-span hot path pays a single commit.
+  @AppMetricsActor
+  @discardableResult
+  func insert(span: SpanRow) throws -> Int64 {
+    return try database.transaction {
```
Ubax, line 592: [💡 suggestion] `SQLiteDatabase.transaction` issues a plain `BEGIN`, so it does not nest. `insertAll(metrics:)` and `insertAll(logs:)` wrap their per-row `insert` in a transaction; a future `insertAll(spans:)` that follows that pattern will fail with "cannot start a transaction within a transaction" because this insert already owns one. Either note in the doc comment that this method must not be called inside a transaction, or split the primitive (uncapped insert plus prune) from the transaction wrapper the way Android does with `insertUncapped` / `insert`, so a batch path can reuse the primitives inside one outer transaction.

packages/expo-app-metrics/ios/Database/MetricsDatabaseRows.swift
```diff
@@ -219,3 +219,129 @@ extension LogRow {
...
+  /// request spans).
+  public let name: String
+
+  /// `SpanKind` per the OTLP proto (`clientKind` for network requests).
+  public let kind: Int
+
+  /// Unix-epoch milliseconds. Stored as integers rather than the ISO strings the other tables use:
+  /// a span's duration is routinely sub-second, and the house ISO format carries no fractional
+  /// seconds, so a string timestamp would collapse every fast span to zero length.
+  public let startTimestampMs: Int64
```
Ubax, line 261: [💡 suggestion] Same comment as on android

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/storage/Span.kt
```diff
@@ -0,0 +1,176 @@
...
+
+  /** `SpanKind` per the OTLP proto (`CLIENT_KIND` for network requests). */
+  val kind: Int = INTERNAL_KIND,
+
+  /**
+   * Unix-epoch milliseconds. Millisecond integers rather than the ISO strings the other tables
+   * use: a span's duration is routinely sub-second, and the house ISO format carries no
+   * fractional seconds, so a string timestamp would collapse every fast span to zero length.
+   */
+  val startTimestampMs: Long,
```
Ubax, line 69: [💡 suggestion] Same point as the iOS `SpanRow`: OTLP is nanosecond-based and the OkHttp listener has nanosecond deltas. Changing this later needs a v19 migration. 

No strong opinion here, so I leave it for you to decide

packages/expo-app-metrics/ios/Tests/MetricsDatabaseTests.swift
```diff
@@ -618,6 +618,172 @@ struct MetricsDatabaseTests {
...
+  func `insert span prunes the oldest rows past the retention cap`() throws {
+    // Span producers (network requests especially) can record orders of magnitude more rows
+    // than metrics or logs. The cap bounds the table when nothing consumes (and deletes) the rows.
+    try withTemporaryDatabase { database in
+      try database.insert(session: makeSessionRow(id: "s"))
+      for _ in 0..<(MetricsDatabase.spanCap + 10) {
+        try database.insert(span: makeSpanRow(sessionId: "s"))
+      }
+      let count = try countRows(database: database, table: "spans")
+      #expect(count <= MetricsDatabase.spanCap)
```
Ubax: [💡 suggestion] `count <= spanCap` also passes if the prune deletes everything but the newest row. Since the prune is id arithmetic, the exact shape is cheap to pin: after `spanCap + 10` inserts expect `count == spanCap` and `rows.first?.id == maxId - Int64(spanCap) + 1`. That catches an off-by-one in `insertedId - spanCap` in either direction.

packages/expo-app-metrics/android/src/test/java/expo/modules/appmetrics/storage/SpanStorageTest.kt
```diff
@@ -0,0 +1,214 @@
...
+  @Test
+  fun `insert prunes the oldest rows past the retention cap`() = runTest {
+    // Span producers (network requests especially) can record orders of magnitude more rows
+    // than metrics or logs. The cap bounds the table when nothing consumes (and deletes) the rows.
+    insertSession("s")
+    repeat(SpanDao.SPAN_CAP + 10) {
+      database.spanDao().insert(makeSpan(sessionId = "s"))
+    }
+    val rows = allSpans()
+    assertTrue(rows.size <= SpanDao.SPAN_CAP)
```
Ubax: [💡 suggestion] Same as the iOS cap test: `<=` passes an over-eager prune. Assert `rows.size == SpanDao.SPAN_CAP` and `rows.first().id == maxId - SpanDao.SPAN_CAP + 1` to pin the boundary.

packages/expo-app-metrics/android/src/test/java/expo/modules/appmetrics/storage/SpanStorageTest.kt
```diff
@@ -0,0 +1,214 @@
...
+  @Test
+  fun `insert assigns increasing row ids`() = runTest {
+    insertSession("s")
+    val firstId = database.spanDao().insert(makeSpan(sessionId = "s"))
+    val secondId = database.spanDao().insert(makeSpan(sessionId = "s"))
+    assertTrue(secondId > firstId)
+  }
+
+  @Test
+  fun `getAll returns rows in ascending row id order`() = runTest {
```
Ubax: [🐧 nit] `getAll` no longer exists on the DAO (the test drains through `getSpans(cursor = -1, ...)`). Rename to match, e.g. `getSpans from the start returns rows in ascending row id order`.

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/storage/Span.kt
```diff
@@ -0,0 +1,176 @@
...
+    deleteUpTo(insertedId - SPAN_CAP)
+    return insertedId
+  }
+
+  /**
+   * At most `limit` spans with `id > cursor`, in ascending rowid order. Mirrors the iOS
+   * `getSpans(afterId:limit:)` read so consumers can drain the table in chunks.
+   */
+  @Query("SELECT * FROM spans WHERE id > :cursor ORDER BY id ASC LIMIT :limit")
+  suspend fun getSpans(cursor: Long, limit: Int): List<Span>
```
Ubax: [🐧 nit] `MetricDao` / `LogDao` name this parameter `afterId`, and so does the `SessionManager.getSpans(afterId, limit)` wrapper one call up. Using `afterId` here too keeps the three DAOs uniform.

Ubax (commented):
AI review

packages/expo-app-metrics/android/src/test/java/expo/modules/appmetrics/storage/SpanStorageTest.kt
```diff
@@ -0,0 +1,214 @@
...
+    // than metrics or logs. The cap bounds the table when nothing consumes (and deletes) the rows.
+    insertSession("s")
+    repeat(SpanDao.SPAN_CAP + 10) {
+      database.spanDao().insert(makeSpan(sessionId = "s"))
+    }
+    val rows = allSpans()
+    assertTrue(rows.size <= SpanDao.SPAN_CAP)
+    // The newest rows survive.
+    val maxId = rows.maxOf { it.id }
+    assertEquals(maxId, rows.last().id)
```
Ubax: [💡 suggestion] The comment says "the newest rows survive", but this compares the max id of the returned list with the list's last element. That only proves `getSpans` orders by id, which another test already covers. It never checks that the row from the final `insert` is still present, nor that the low ids are gone. Capture the id returned by the last `insert` and assert it is in `rows`, and assert `rows.first().id == lastId - SpanDao.SPAN_CAP + 1`.

packages/expo-app-metrics/ios/Tests/MetricsDatabaseTests.swift
```diff
@@ -618,6 +618,172 @@ struct MetricsDatabaseTests {
...
+      let rows = try database.getSpans(afterId: -1)
+      let maxId = try #require(try database.getMaxSpanId())
+      #expect(rows.compactMap(\.id).max() == maxId)
```
Ubax: [💡 suggestion] This assertion is tautological: `rows` and `getMaxSpanId()` read the same table, so the max of the returned ids always equals `MAX(id)`. It passes even if the prune deleted every row but one. Keep the id returned by the last `insert(span:)` and assert it appears in `rows`, and assert `rows.first?.id == lastId - Int64(MetricsDatabase.spanCap) + 1` so the prune boundary is pinned.

packages/expo-app-metrics/ios/Tests/SpanRowTests.swift
```diff
@@ -0,0 +1,54 @@
...
+
+  @Test
+  func `generates a 16-character lowercase hex span id`() {
+    let spanId = SpanRow.generateSpanId()
+    #expect(spanId.count == 16)
+    #expect(spanId.allSatisfy { $0.isHexDigit && !$0.isUppercase })
+  }
+
+  @Test
+  func `never generates an all-zero identifier`() {
```
Ubax: [🐧 nit] 512 draws cannot tell "redraws on zero" apart from "no redraw at all": a zero draw has probability 2^-64, so this test passes identically with the `guard` removed. It is a smoke test of the format, not a proof of the invariant the name states. Either rename it to say what it checks, or make the redraw testable by letting `generateHexId` take a generator (`inout some RandomNumberGenerator`) and feed it a generator that returns zero first.

packages/expo-app-metrics/ios/Tests/SpanRowTests.swift
```diff
@@ -0,0 +1,54 @@
...
+      #expect(SpanRow.generateTraceId() != String(repeating: "0", count: 32))
+      #expect(SpanRow.generateSpanId() != String(repeating: "0", count: 16))
+    }
+  }
+
+  @Test
+  func `generates distinct identifiers on each call`() {
+    var seen = Set<String>()
+    for _ in 0..<128 {
+      seen.insert(SpanRow.generateTraceId())
```
Ubax: [🐧 nit] Name says identifiers, body checks only `generateTraceId()`. `generateSpanId()` is a single 8-byte word, so it is the one more likely to collide if the generator were ever swapped. Add a second set for span ids, or assert both in one loop.

packages/expo-app-metrics/android/src/test/java/expo/modules/appmetrics/storage/SpanStorageTest.kt
```diff
@@ -0,0 +1,214 @@
...
+    assertTrue(span.spanId.all { it.isDigit() || it in 'a'..'f' })
+    assertNotEquals("0".repeat(32), span.traceId)
+    assertNotEquals("0".repeat(16), span.spanId)
+  }
+
+  @Test
+  fun `each new span receives distinct identifiers`() {
+    // Ids are assigned once at record time and persisted, so a redelivered row (export is
+    // at-least-once) reaches the server byte-identical instead of becoming a fresh duplicate.
+    val ids = (0 until 128).map { makeSpan(sessionId = "s").traceId }.toSet()
```
Ubax: [🐧 nit] Same as the iOS counterpart: the name says identifiers, but only `traceId` is collected. Collect `spanId` too, since the 8-byte id is the one with the smaller space.

Ubax (commented):
More AI comments on test coverage

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/storage/Span.kt
```diff
@@ -0,0 +1,179 @@
...
+     */
+    const val SPAN_CAP = 2_000
+  }
+
+  /**
+   * Raw row insert with no cap enforcement, the primitive `insert` builds on. Producers should
+   * call `insert`: picking this one silently loses the retention bound.
+   */
+  @Insert
+  suspend fun insertUncapped(span: Span): Long
```
tsapeta: [💡 suggestion] `insertUncapped` is reachable by any producer here, while the iOS counterpart is `private func insertUncapped(span:)` — so the doc's "Producers should call `insert`" is enforced by the compiler on iOS and by convention only on Android. A Kotlin interface member can't be private, so the asymmetry is structural rather than an oversight.

It matters a little more than usual because the failure is silent: picking the uncapped overload drops the retention bound with no error, and the whole point of the cap is the case where nothing consumes the rows (`expo-observe` absent or dispatch disabled) — exactly when nobody is watching the table grow.

If you want the same guarantee on both platforms, the usual shape is to keep the DAO uncapped-only and put the capped `insert` on `SessionManager` (which is already the module's write facade and can be `internal`), so the uncapped primitive is never in a producer's reach. Not blocking for a storage-only PR with no producers yet, but worth deciding before the network-request producer lands on top.

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/storage/Span.kt
```diff
@@ -0,0 +1,179 @@
...
+     * would let a busy app accumulate a week of traffic. `insert` prunes past the cap.
+     */
+    const val SPAN_CAP = 2_000
+  }
+
+  /**
+   * Raw row insert with no cap enforcement, the primitive `insert` builds on. Producers should
+   * call `insert`: picking this one silently loses the retention bound.
+   */
+  @Insert
```
tsapeta: [🐧 nit] Every other `@Insert` in this package specifies a conflict strategy — `MetricDao.insert`/`insertAll` and `LogDao.insertAll` use `OnConflictStrategy.IGNORE`, `CrashReportDao.upsert` uses `REPLACE` — whereas this one takes the default (`ABORT`).

No bug: `id` is autogenerated and `spans` has no other unique constraint, so a conflict isn't reachable today. Flagging it only because the next reader comparing the DAOs will wonder whether the omission was deliberate. Either add `(onConflict = OnConflictStrategy.IGNORE)` to match, or say in the doc comment that ABORT is intentional because no conflict is possible.

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/storage/Span.kt
```diff
@@ -0,0 +1,179 @@
+// Copyright 2025-present 650 Industries. All rights reserved.
+
+package expo.modules.appmetrics.storage
```
tsapeta: [🐧 nit] The other four entities and all four existing DAOs live together in `MetricsDatabase.kt`; this PR puts `Span` and `SpanDao` in their own file instead.

The new layout is arguably the better one — `MetricsDatabase.kt` is already carrying five entities plus five DAOs plus the two `@Relation` holders — so I'd rather see the convention move this way than see `Span` folded back in. Raising it only so the split is a stated choice rather than an accident, since it's the first file in the package to break the pattern. Also worth noting `Span` skips the `@Serializable` the sibling entities carry; that looks correct, since nothing in this package or `expo-observe` actually serializes `Metric`/`LogRecord`/`Session` through kotlinx.

tsapeta:
@expo-bot review

packages/expo-observe/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 ### 🐛 Bug fixes
 
 - Fix OTel date conversion ([#48161](https://github.com/expo/expo/pull/48161) by [@Ubax](https://github.com/Ubax))
+- [iOS] Stop overlapping dispatches from sending the same pending rows twice. `dispatchEvents()` now returns without dispatching when a dispatch is already in progress. ([#48883](https://github.com/expo/expo/pull/48883) by [@tsapeta](https://github.com/tsapeta))
```
Ubax: Does this belong here?

Ubax (approved):

PR #48863, opened 2026-08-13
[ios][expo-go] Track session expiry and resolve usernames for partner actors

alanjhughes:
# Why
Partner platforms can provision Expo accounts for their users, and those users never see the account credentials, so they cannot sign in to Expo Go and cannot open their previews since the manifest username check landed in SDK 57. This is the first of three PRs adding a device authorization sign in for those users.

# How
Sessions minted by the device grant carry an expiry, so AuthenticationService now stores one and treats an expired partner session as signed out. The current user query moves from meUserActor, which is null for partner actors, to meActor with a PartnerActor fragment, the same shape the cli uses. 

# Test Plan
New unit tests cover expiry storage and detection, actor decoding for partner, regular and null actors. Signed in with a regular account and confirmed the account sheet still lists accounts with avatars, which exercises the rewritten query against the real API.

### DIFF
--- a/apps/expo-go/ios/Client/SwiftUI/Services/AuthenticationService.swift
+++ b/apps/expo-go/ios/Client/SwiftUI/Services/AuthenticationService.swift
@@ -5,6 +5,10 @@ import AuthenticationServices
 import Combine
 import ExpoModulesCore
 
+extension Notification.Name {
+  static let expoSessionDidChange = Notification.Name("expo-session-did-change")
+}
+
 @MainActor
 class AuthenticationService: ObservableObject {
   @Published var user: UserActor?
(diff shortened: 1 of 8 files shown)


apps/expo-go/ios/Client/SwiftUI/Services/AuthenticationService.swift
```diff
@@ -53,31 +69,42 @@ class AuthenticationService: ObservableObject {
...
   private func fetchUserInfo() async {
     do {
-      let response: MeUserActorResponse = try await APIClient.shared.request(Queries.getCurrentUser())
-      user = response.data.meUserActor
-
-      if let username = user?.username {
-        UserDefaults.standard.set(username, forKey: usernameKey)
+      let response: MeActorResponse = try await APIClient.shared.request(Queries.getCurrentUser())
+      guard let actor = response.data.meActor else {
+        print("[AuthenticationService] meActor was null. Signed in as an actor type Expo Go does not model.")
```
davidko604, line 97: Would it make sense to clear the session here? I think the most common case that meActor is null is the session expired.

davidko604 (approved):
Looks good to me. One small suggestion in the case of meActor being null guard.

brentvatne (approved):

lukmccall (approved):

PR #48864, opened 2026-08-13
[ios][expo-go] Add the device authorization sign in flow

alanjhughes:
# Why
Adds the UI necessary for the feature.

# How

The flow follows RFC 8628 with the number match step the server added in [expo/universe#29795](https://github.com/expo/universe/issues/29795). A partner's QR is the dev server URL plus one query parameter. DeviceLoginLink validates and strips it.

The sheet has three steps, the code with a copy button and an in-app browser so a single device user can finish the flow, the number match, and error states which describe the recovery the app actually offers. The browser dismisses itself shortly after a poll sees the approval.

# Test Plan
Unit tests cover the URL parsing and stripping, every shape the token endpoint returns including expires_at with and without fractional seconds, and each timing rule in the state machine.

### DIFF
--- a/apps/expo-go/ios/Client/SwiftUI/DeviceLogin/DeviceLoginCodeView.swift
+++ b/apps/expo-go/ios/Client/SwiftUI/DeviceLogin/DeviceLoginCodeView.swift
@@ -1,0 +1,62 @@
+// Copyright 2015-present 650 Industries. All rights reserved.
+
+import SwiftUI
+import UIKit
+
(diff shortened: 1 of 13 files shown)


davidko604 (approved):
LGTM. I can't review the swift itself so I focused on what I could verify. API calls match www and ran the full flow on a simulator against production. Deferring the native side to others.

lukmccall (approved):

PR #48872, opened 2026-08-13
[ios][camera] Emit onRecordingProgress while recording video

alanjhughes:
# Why
iOS side of the onRecordingProgress event

# How
Use a timer to track the recording duration and emit an event at the set interval

# Test Plan
Bare expo

### DIFF
--- a/packages/expo-camera/ios/Current/CameraVideoRecording.swift
+++ b/packages/expo-camera/ios/Current/CameraVideoRecording.swift
@@ -130,8 +130,12 @@ class CameraVideoRecording: NSObject, AVCaptureFileOutputRecordingDelegate {
   private func startProgressTimer(for output: AVCaptureFileOutput?) {
     guard isProgressActive else { return }
     progressTimer?.invalidate()
-    let timer = Timer(timeInterval: progressInterval, repeats: true) { [weak self, weak output] _ in
-      guard let self, let output, output.recordedDuration.isNumeric else {
+    let timer = Timer(timeInterval: progressInterval, repeats: true) { [weak self, weak output] timer in
+      guard let self, let output else {
+        timer.invalidate()
+        return
+      }
+      guard output.recordedDuration.isNumeric else {
         return
       }
       var payload: [String: Any] = [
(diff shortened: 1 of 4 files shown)


packages/expo-camera/ios/Current/CameraVideoRecording.swift
```diff
@@ -134,5 +195,7 @@ class CameraVideoRecording: NSObject, AVCaptureFileOutputRecordingDelegate {
     videoRecordedPromise?.reject(CameraUnmountedException())
     videoRecordedPromise = nil
     videoCodecType = nil
+    maxDuration = nil
```
behenate: Not 100% sure on this, but I think we should run this on the main queue, since we are setting the maxDuration on the main queue in other places so there may be a small possibility of setting the `maxDuration` to some value after setting it to nil.

packages/expo-camera/ios/Current/CameraVideoRecording.swift
```diff
@@ -118,6 +177,8 @@ class CameraVideoRecording: NSObject, AVCaptureFileOutputRecordingDelegate {
     defer {
       videoRecordedPromise = nil
       videoCodecType = nil
+      maxDuration = nil
```
behenate: Same as in `cleanup()`

packages/expo-camera/ios/Current/CameraVideoRecording.swift
```diff
@@ -109,6 +121,53 @@ class CameraVideoRecording: NSObject, AVCaptureFileOutputRecordingDelegate {
...
+    }
+  }
+
+  private func startProgressTimer(for output: AVCaptureFileOutput?) {
+    guard isProgressActive else { return }
+    progressTimer?.invalidate()
+    let timer = Timer(timeInterval: progressInterval, repeats: true) { [weak self, weak output] _ in
+      guard let self, let output, output.recordedDuration.isNumeric else {
+        return
+      }
```
behenate: A scenario claude came up with when the `CameraView` deinits: 

```
1. The view deinits. cleanup() runs synchronously inside deinit. The recorder is still alive here.
2. stopProgressTimer() does not invalidate the timer now. It only enqueues an async block on the main queue. The block captures [weak self].
3. deinit finishes. The view releases its strong reference. The recorder deallocates.
4. The async block runs later. self is now nil. self?.progressTimer?.invalidate() does nothing.
5. The run loop still retains the repeating timer. Nothing invalidates it after this point.
```

Here's a fix which makes the timer invalidate itself once the recording object gets deallocated
```suggestion
      guard let self, let output else {
        timer.invalidate()
        return
      }
      guard output.recordedDuration.isNumeric else {
        return
      }
```

behenate (approved):

Issue #33754, opened 2024-12-19
Race condition in AsyncStorage sqlite/kv-store


rtorrente:
### What platform(s) does this occur on?

Android, iOS

### Where did you reproduce the issue?

in a development build, in a standalone app

### Summary

Please **don't close the issue**, I don't provide reproducible code as I didn't manage to reproduce the error on my side (only a lot of Crashalytics and Android Vitals feedback) but **I give here a debugging thought and resolution track**. 

I migrated a high-traffic application from the **AsyncStorage** library to **expo-sqlite/kv-store**. This was the only change with the previous version, and I started getting a lot of weird and not easily understood crash feedback on Crashalytics. I've never been able to reproduce them on my own. 
Android Vitals sent me an alert that the version was too unstable, requesting a correction before deleting the application from the PlayStore.

Exemples of strange error that I encounter : 

**SIGABRT**

OS Version: Android 13 (TP1A.220624.014.A325FXXSBDXJ1)
Report Version: 104

Exception Type: Unknown (SIGABRT)

Application Specific Information:
Abort

```
Thread 0 Crashed:
0   libc.so                         0x71e32fadb4        abort
1   libc.so                         0x71e32fcc2c        <unknown> + 489142864940
2   libc.so                         0x71e336b51c        <unknown> + 489143317788
3   libc.so                         0x71e336b398        pthread_mutex_lock
4   split_config.arm64_v8a.apk      0x708688f2bc        <unknown> + 483293459132
5   split_config.arm64_v8a.apk      0x708687b0ac        expo::NativeStatementBinding::getColumnNames
6   split_config.arm64_v8a.apk      0x708687d56c        _ZN8facebook3jni6detail13MethodWrapperIMN4expo22NativeStatementBindingEFNS0_16basic_strong_refINS0_10JArrayListINS0_7JStringEEENS0_23LocalReferenceAllocatorEEEvEXadL_ZNS4_14getColumnNamesEvEES4_SA_JEE8dispatchENS0_9alias_refIPNS1_8JTypeForINS0_11HybridC...
[...]

PR #48878, opened 2026-08-13
[sqlite] Fix kv-store permanently losing the storage table on a raced first launch

### LINKED ISSUE #47448: expo-sqlite/kv-store: getDbSync() bypasses the awaitLock, racing async access and corrupting the migration (no such table: storage)

giaBaoJS:
# Why

Fixes #47448.

`SQLiteStorage` (`expo-sqlite/kv-store`) can permanently lose its `storage` table on a raced first launch, after which every `getItemSync`/`setItemSync`/`getItemAsync` throws `no such table: storage` forever — the bad state survives an app reinstall, because the database file is restored from backup.

`getDbAsync()` serializes its init through `awaitLock` (added in #33834 for #33754), but `getDbSync()` has no shared lock with it, so both paths can run the first-run migration concurrently on the same cached native connection:

https://github.com/expo/expo/blob/main/packages/expo-sqlite/src/Storage.ts#L360-L381

Both `maybeMigrateDbAsync()` and `maybeMigrateDbSync()` returned early once `user_version >= DATABASE_VERSION`, so any raced launch that ended at the current version without the `storage` table became permanently unrecoverable. The version guard guaranteed the migration was never retried.

Credit to @tsushanth, who diagnosed this and proposed the same fix in #47540 before closing it themselves for inactivity ("Happy to reopen or resubmit if there's still interest"). This PR carries that fix forward on current `main` and adds regression tests and a changelog entry.

# How

Run `MIGRATION_STATEMENT_0` unconditionally instead of only when `user_version === 0`, and keep the `PRAGMA user_version` write conditional.

**Why self-heal rather than a lock.** The issue suggests either guarding `getDbSync()` with a lock or making the migration idempotent. A lock is not actually available on the sync path: `awaitLock` is promise-based and `getDbSync()` is synchronous, so it cannot wait on it, and JS has no synchronous mutex to fall back on. More importantly, a lock only prevents *future* corruption — it does nothing for the users already stuck, who cannot recover by reinstalling. The idempotent migration does not prevent the race, but it makes the resulting broken state self-healing and repairs already-affected databases the next time the app opens them. I left a `NOTE` on `getDbSync()` recording why it does not take the lock and that recovery happens on the next open, so the asymmetry with `getDbAsync()` does not read as an oversight.

This cannot lose data. `MIGRATION_STATEMENT_0` is `CREATE TABLE IF NOT EXISTS storage (...)` — on a healthy database it is a no-op, and there is no `DROP`, `DELETE`, or `CREATE TABLE` without `IF NOT EXISTS` anywhere on this path. The two "normal path" tests below assert that explicitly.

# Test Plan

Four tests added to `packages/expo-sqlite/src/__tests__/Storage-test.ios.ts`. They run against a real file-backed SQLite database through the existing `src/__mocks__/ExpoSQLite` (node:sqlite), and they are deterministic — the broken state is seeded directly rather than reproduced by racing, so there is no timing dependency.

Two reproduce the bug: seed a database at `user_version = 1` with no `storage` table (exactly the state the race leaves behind), then assert that `getItemSync` / `getItemAsync` recreate the table instead of throwing, and that a subsequent write/read round-trips. This is the user-visible path — an app already stuck in the broken state starts working again after upgrading.

Two guard the normal paths: a fresh database still migrates and ends at `user_version = 1`, and reopening an already-migrated database preserves its rows and its version.

Red — new tests against unmodified `Storage.ts` (the two control tests already pass, only the two self-heal tests fail):

```
FAIL iOS src/__tests__/Storage-test.ios.ts
  ● SQLiteStorage migration › should recreate the storage table when the database is at
    the current version but the table is missing (sync)

    no such table: storage

      at SQLiteStorage.getFirstSync [as getItemSync] (src/Storage.ts:147:23)

  ● SQLiteStorage migration › should recreate the storage table when the database is at
    the current version but the table is missing (async)

    Received promise rejected instead of resolved
    Rejected to value: [Error: no such table: storage]

Test Suites: 1 failed, 1 passed, 2 total
Tests:       2 failed, 34 passed, 36 total
```

Green — full `expo-sqlite` suite with the fix applied (baseline on `main` is 190 passing; the 4 added tests take it to 194):

```
Test Suites: 10 passed, 10 total
Tests:       1 skipped, 194 passed, 195 total
```

`et check-packages expo-sqlite` (build, typecheck, depscheck, test, lint, format):

```
expo-sqlite:lint: Found 0 warnings and 0 errors.
expo-sqlite:format: All matched files use the correct format.
🏁 All checks passed.
```

# Checklist

### DIFF
--- a/packages/expo-sqlite/CHANGELOG.md
+++ b/packages/expo-sqlite/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 - [tvOS] Fix path for DB creation. ([#46715](https://github.com/expo/expo/pull/46715) by [@douglowder](https://github.com/douglowder))
 - Fixed the devtools plugin bundle missing its `wa-sqlite.wasm` asset. ([#48542](https://github.com/expo/expo/pull/48542) by [@kudo](https://github.com/kudo))
+- Fixed `SQLiteStorage` permanently throwing `no such table: storage` when the synchronous and asynchronous APIs raced the first-run migration. ([#48878](https://github.com/expo/expo/pull/48878) by [@giaBaoJS](https://github.com/giaBaoJS))
 
 ### 💡 Others
 
--- a/packages/expo-sqlite/src/Storage.ts
+++ b/packages/expo-sqlite/src/Storage.ts
@@ -372,6 +372,9 @@ export class SQLiteStorage {
   }
 
   private getDbSync(): SQLiteDatabase {
+    // NOTE: Unlike `getDbAsync()`, this cannot acquire `awaitLock` because the lock is
+    // promise-based and this method is synchronous. The migration is idempotent instead,
+    // so racing it with `getDbAsync()` cannot leave the database without a `storage` table.
     if (!this.db) {
       const db = openDatabaseSync(this.databaseName);
       this.maybeMigrateDbSync(db);
@@ -383,30 +386,26 @@ export class SQLiteStorage {
   private maybeMigrateDbAsync(db: SQLiteDatabase) {
     return db.withTransactionAsync(async () => {
       const result = await db.getFirstAsync<{ user_version: number }>('PRAGMA user_version');
-      let currentDbVersion = result?.user_version ?? 0;
-      if (currentDbVersion >= DATABASE_VERSION) {
-        return;
+      const currentDbVersion = result?.user_version ?? 0;
+      // `MIGRATION_STATEMENT_0` uses `CREATE TABLE IF NOT EXISTS`, so it is a no-op when the
+      // table is already there. Running it unconditionally rather than only when the version is
+      // still 0 also repairs a database left at `user_version = 1` without a `storage` table.
+      await db.execAsync(MIGRATION_STATEMENT_0);
+      if (currentDbVersion < DATABASE_VERSION) {
+        await db.execAsync(`PRAGMA user_version = ${DATABASE_VERSION}`);
       }
-      if (currentDbVersion === 0) {
-        await db.execAsync(MIGRATION_STATEMENT_0);
-        currentDbVersion = 1;
-      }
-      await db.execAsync(`PRAGMA user_version = ${DATABASE_VERSION}`);
     });
   }
 
   private maybeMigrateDbSync(db: SQLiteDatabase) {
     db.withTransactionSync(() => {
       const result = db.getFirstSync<{ user_version: number }>('PRAGMA user_version');
-      let currentDbVersion = result?.user_version ?? 0;
-      if (currentDbVersion >= DATABASE_VERSION) {
-        return;
+      const currentDbVersion = result?.user_version ?? 0;
+      // See the comment in `maybeMigrateDbAsync()`.
+      db.execSync(MIGRATION_STATEMENT_0);
+      if (currentDbVersion < DATABASE_VERSION) {
+        db.execSync(`PRAGMA user_version = ${DATABASE_VERSION}`);
       }
-      if (currentDbVersion === 0) {
-        db.execSync(MIGRATION_STATEMENT_0);
-        currentDbVersion = 1;
-      }
-      db.execSync(`PRAGMA user_version = ${DATABASE_VERSION}`);
     });
   }
 
--- a/packages/expo-sqlite/src/Storage.ts
+++ b/packages/expo-sqlite/src/Storage.ts
@@ -394,7 +394,10 @@ export class SQLiteStorage {
       // A new column has to be added here as well, so fresh and repaired databases also get it.
       await db.execAsync(MIGRATION_STATEMENT_0);
 
-      // Version ladder: add each new migration below, gated on `currentDbVersion`.
+      // Version ladder: add each new migration below, gated on `currentDbVersion`. Since the
+      // baseline above already carries every column, a ladder entry that adds one has to
+      // tolerate it already being there — gate it on `PRAGMA table_info(storage)`, or a fresh
+      // database fails the migration with `duplicate column name`.
       if (currentDbVersion >= DATABASE_VERSION) {
         return;
       }
--- a/packages/expo-sqlite/src/Storage.ts
+++ b/packages/expo-sqlite/src/Storage.ts
@@ -387,13 +387,18 @@ export class SQLiteStorage {
     return db.withTransactionAsync(async () => {
       const result = await db.getFirstAsync<{ user_version: number }>('PRAGMA user_version');
       const currentDbVersion = result?.user_version ?? 0;
-      // `MIGRATION_STATEMENT_0` uses `CREATE TABLE IF NOT EXISTS`, so it is a no-op when the
-      // table is already there. Running it unconditionally rather than only when the version is
-      // still 0 also repairs a database left at `user_version = 1` without a `storage` table.
+
+      // Baseline schema, deliberately outside the version ladder below. It is a
+      // `CREATE TABLE IF NOT EXISTS`, so it is a no-op on a healthy database and it also repairs
+      // one that a raced first-run migration left at `user_version >= 1` with no `storage` table.
+      // A new column has to be added here as well, so fresh and repaired databases also get it.
       await db.execAsync(MIGRATION_STATEMENT_0);
-      if (currentDbVersion < DATABASE_VERSION) {
-        await db.execAsync(`PRAGMA user_version = ${DATABASE_VERSION}`);
+
+      // Version ladder: add each new migration below, gated on `currentDbVersion`.
+      if (currentDbVersion >= DATABASE_VERSION) {
+        return;
       }
+      await db.execAsync(`PRAGMA user_version = ${DATABASE_VERSION}`);
     });
   }
 
@@ -401,11 +406,14 @@ export class SQLiteStorage {
     db.withTransactionSync(() => {
       const result = db.getFirstSync<{ user_version: number }>('PRAGMA user_version');
       const currentDbVersion = result?.user_version ?? 0;
-      // See the comment in `maybeMigrateDbAsync()`.
+
+      // Keep in sync with `maybeMigrateDbAsync()`, which documents the two steps below.
       db.execSync(MIGRATION_STATEMENT_0);
-      if (currentDbVersion < DATABASE_VERSION) {
-        db.execSync(`PRAGMA user_version = ${DATABASE_VERSION}`);
+
+      if (currentDbVersion >= DATABASE_VERSION) {
+        return;
       }
+      db.execSync(`PRAGMA user_version = ${DATABASE_VERSION}`);
     });
   }
 
--- a/packages/expo-sqlite/src/Storage.ts
+++ b/packages/expo-sqlite/src/Storage.ts
@@ -374,7 +374,7 @@ export class SQLiteStorage {
   private getDbSync(): SQLiteDatabase {
     // NOTE: Unlike `getDbAsync()`, this cannot acquire `awaitLock` because the lock is
     // promise-based and this method is synchronous. The migration is idempotent instead,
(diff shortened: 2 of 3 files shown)


brentvatne:
@expo-bot check

giaBaoJS:
Thanks @brentvatne for running the check — I've addressed the maintainability note in 3a80154.

### What changed

`maybeMigrateDbAsync()` and `maybeMigrateDbSync()` now read as two separate steps instead of one flattened block:

```ts
const currentDbVersion = result?.user_version ?? 0;

// Baseline schema, deliberately outside the version ladder below. It is a
// `CREATE TABLE IF NOT EXISTS`, so it is a no-op on a healthy database and it also repairs
// one that a raced first-run migration left at `user_version >= 1` with no `storage` table.
// A new column has to be added here as well, so fresh and repaired databases also get it.
await db.execAsync(MIGRATION_STATEMENT_0);

// Version ladder: add each new migration below, gated on `currentDbVersion`.
if (currentDbVersion >= DATABASE_VERSION) {
  return;
}
await db.execAsync(`PRAGMA user_version = ${DATABASE_VERSION}`);
```

The repair still runs unconditionally, but the ladder gets its early return back, so whoever adds `DATABASE_VERSION = 2` writes their statement inside a block that is already version-gated instead of having to notice that it isn't. The comment also asks them to add any new column to `MIGRATION_STATEMENT_0`, since that statement is the whole schema a fresh or repaired database gets. The sync twin is structurally identical to the async one.

### Behaviour is unchanged

`if (v < N) { bump }` and `if (v >= N) return; bump` are the same function, so this is a readability change only. Two checks:

- Reverting **only** the restructure, keeping the tests: same result, all green.
- Reverting **the fix** (`MIGRATION_STATEMENT_0` back inside a `currentDbVersion === 0` branch) on top of the restructured code still turns the two self-heal tests red — `no such table: storage` for the sync one, `Rejected to value: [Error: no such table: storage]` for the async one — so the restructure did not weaken what they detect.

Run in a package-scoped jest harness over 9 of the 10 `*-test.ios.*` suites (`hooks-test.ios.tsx` needs devDeps my sparse checkout doesn't have): 184 passing before, 185 after, the one addition being the new test below.

### No test for the restructure itself

Behaviour is identical, so no test could go red without it, and I'd rather say that than add one that passes on both arms.

I did add one test that is not about the restructure: `should leave a database from a newer version untouched`. The check verified the `user_version > DATABASE_VERSION` case by hand, but the suite never asserted it — and the ladder is exactly where a future migration could downgrade it by accident. It is not vacuous: dropping the version guard makes it fail with `Expected: 99 / Received: 1`.

### On the read-only observation

The check's finding that `CREATE TABLE IF NOT EXISTS` on an existing table attempts no write at all is what makes the "outside the ladder" placement safe, so I kept the repair unconditional rather than gating it behind a table-exists probe. I reproduced it under `node:sqlite`: read-only with the table present succeeds; read-only with the table missing gives `attempt to write a readonly database` — but such a database has no usable `storage` table anyway, and `SQLiteStorage` never passes open options, so it always opens read-write.

brentvatne:
@expo-bot check with fable

giaBaoJS:
Thanks for the second check @brentvatne. The comment collision it found is real and I've closed it in f92762b. Comment text only, no code touched.

The finding was that lines 391-397 pointed a future maintainer down a path that breaks. One comment said to add a new column to `MIGRATION_STATEMENT_0`; the other said to add the matching migration to the ladder. Doing both means a fresh database gets the column from the baseline and then hits `ALTER TABLE storage ADD COLUMN` in the ladder.

I ran the hypothetical v2 upgrade (schema gains `updated_at`) against `node:sqlite` to check both halves before rewording anything:

```
--- fresh database (user_version 0) ---
M0 kept at v1 schema, plain ladder                  -> ok
column added to M0 (what the comment says)          -> THROWS: duplicate column name: updated_at
column added to M0 + ladder gated on table_info     -> ok

--- existing v1 database, real upgrade path ---
M0 kept at v1 schema, plain ladder                  -> ok
column added to M0 (what the comment says)          -> ok
column added to M0 + ladder gated on table_info     -> ok

--- repaired database: user_version already 2, storage table missing ---
M0 kept at v1 schema     -> columns = ["key","value"]
column added to M0       -> columns = ["key","value","updated_at"]
```

The third block is why the first comment says what it says, so I kept it. A database the race left at `user_version >= DATABASE_VERSION` never enters the ladder, so the baseline is the only thing that runs on it. Leave the column out of `MIGRATION_STATEMENT_0` and a repaired database silently ends up a column short. The advice was necessary, just not sufficient. The ladder entry also has to tolerate a column the baseline already created, so that is the clause I added:

```ts
// Version ladder: add each new migration below, gated on `currentDbVersion`. Since the
// baseline above already carries every column, a ladder entry that adds one has to
// tolerate it already being there. Gate it on `PRAGMA table_info(storage)`, or a fresh
// database fails the migration with `duplicate column name`.
```

Nothing here changes behaviour today: `DATABASE_VERSION` is 1 and `MIGRATION_STATEMENT_0` is the entire schema, so the ladder has no entries yet. Happy to drop the extra sentence if you'd rather keep the comment short and handle it when a v2 migration actually lands.

barthap (approved):
LGTM, thanks for iterating on all the bot comments.

I checked the issue locally: `main` gets stuck at `user_version=1` with no `storage` table and never recovers, this PR fixed it on the next open.

One semi-related thing I noticed, but not for this PR: `withTransactionSync`/`withTransactionAsync` always run `ROLLBACK` in the catch, even when their own `BEGIN` is what failed. That's the rollback that kills the other path's `CREATE TABLE`. If it only rolled back its own transaction, the corruption wouldn't happen in the first place. Either way, this PR is still needed to fix already broken DBs, though.

giaBaoJS:
Thanks @barthap.

Your `withTransaction*` observation holds up, I went and checked it. Both entry points have the same shape:

```ts
// SQLiteDatabase.ts:140
public async withTransactionAsync(task: () => Promise<void>): Promise<void> {
  try {
    await this.execAsync('BEGIN');
    await task();
    await this.execAsync('COMMIT');
  } catch (e) {
    await this.execAsync('ROLLBACK');
    throw e;
  }
}
```

`withTransactionSync` at line 298 is identical. The `catch` cannot tell which of the three statements threw, so when `BEGIN` is the one that fails, the `ROLLBACK` lands on whatever transaction is already open on the shared connection rather than on one this call ever started. That is the step that discards the other path's `CREATE TABLE` and leaves `user_version` ahead of the schema.

Tracking whether `BEGIN` succeeded before rolling back would stop the corruption at the source. Agreed it does not belong in this PR, since a database already in that state still needs the repair. Happy to open it as a separate PR if you want it, along with a test that has one path fail its `BEGIN` while another holds an open transaction. Say the word and I will pick it up.

On CI: the red check is `ios-test-e2e`, failing on `expo-video/fullscreen-test.yaml` after 3 attempts. This PR only touches `packages/expo-sqlite`, so it looks unrelated. I do not have permission to re-run the job myself.

barthap:
Yeah, CI is unrelated, don't worry about it. Regarding the follow-up, let's not do it as a PR yet. `withTransaction*` is public API, so changing when it rolls back hits everyone using it, not just kv-store. And there's more to it than tracking whether `BEGIN` succeeded, we'd need to decide what happens when it fails with `SQLITE_BUSY` instead, whether `SAVEPOINT` fits better, and if anything out there depends on the current behaviour. I'd rather we settle that on our side first.
An issue would be welcome though, with the repro and the failing-`BEGIN` test you described. We'll pick up the design from there.

giaBaoJS:
Filed as #49281, @barthap.

It carries the failing-`BEGIN` test and the repro you asked for, plus measurements on the two things you flagged. Short version: `SQLITE_BUSY` today is worse than it looks, because the stray `ROLLBACK` replaces the error that would tell a caller to retry (`cannot rollback - no transaction is active` instead of `database is locked`), so moving `BEGIN` out of the `try` is what makes a busy database reportable at all, before any retry policy exists.

One thing worth flagging up front: guarding the `ROLLBACK` on `isInTransaction*` looks like the obvious fix and it does not work. That is `sqlite3_get_autocommit`, a property of the connection rather than of the call, so when our `BEGIN` fails because someone else's transaction is open, the guard reads "in a transaction" and lets the destructive `ROLLBACK` through anyway. I measured all three shapes side by side in the issue.

I left `SAVEPOINT` and the error-preservation question open rather than picking for you. Happy to turn it into a PR once you have settled the design, or to leave it entirely with your team.

Kudo:
it feels like we were not fixing the root cause of race condition between async/sync opening. it works because the query is `CREATE TABLE IF NOT EXISTS ...`. any chance we could deliver right fix?

PR #39707, opened 2025-09-16
[video/jest] Fix mocks for expo-video

behenate:
# Why

Fixes https://github.com/expo/expo/issues/37842

Jest doesn't currently work with expo-video

# How

I tried to use the new mock generator (running ` npx expo-modules-test-core generate-ts-mocks`) but it didn't generate a valid mock. 

For now pasted the typescript definition into ChatGPT and asked it to generate a mock 🤷. Seems to be working.

# Test Plan

Tested in the repro app provided by the user

PR #48880, opened 2026-08-13
[video][jest] Add jest-expo mocks for expo-video

### LINKED ISSUE #37842: [jest-expo] TypeError: Cannot read properties of undefined (reading 'prototype') for expo-video

https://github.com/ap4y/expo-video-jest-issue

### Steps to reproduce

Run `npm test`, test fails with `TypeError: Cannot read properties of undefined (reading 'prototype')`

### Environment

```text
[nix-shell:~/git/expo-jest-test]$ npx expo-env-info

  expo-env-info 1.3.4 environment info:
    System:
      OS: Linux 6.12 Void
      Shell: 5.2.37 - /nix/store/1q9lw4r2mbap8rsr8cja46nap6wvrw2p-bash-interactive-5.2p37/bin/bash
    Binaries:
      Node: 22.14.0 - /nix/store/c8jxsih8yy2rnncdmx2hyraizf689nvp-nodejs-22.14.0/bin/node
      npm: 10.9.2 - /nix/store/c8jxsih8yy2rnncdmx2hyraizf689nvp-nodejs-22.14.0/bin/npm
    SDKs:
      Android SDK:
        API Levels: 35
        Build Tools: 34.0.0, 35.0.0
        System Images: android-35 | Google APIs Intel x86_64 Atom
    npmPackages:
      expo: ~53.0.17 => 53.0.17
      react: 19.0.0 => 19.0.0
      react-native: 0.79.5 => 0.79.5
    Expo Workflow: managed
```

### Expo Doctor Diagnostics

```text
[nix-shell:~/git/expo-jest-test]$ npx expo-doctor@latest
15/15 checks passed. No issues detected!
```

giaBaoJS:
# Why

Fixes #37842.

Any Jest test that imports `expo-video`, directly or transitively, fails before a single test body runs:

```
TypeError: Cannot read properties of undefined (reading 'prototype')

    >  9 | const replace = NativeVideoModule.VideoPlayer.prototype.replace;
           |                                               ^
      at Object.prototype (src/VideoPlayer.tsx:9:47)
```

`VideoPlayer.tsx` patches `NativeVideoModule.VideoPlayer.prototype.replace` at module load. Under the `jest-expo` preset, `requireNativeModule` is swapped for `requireMockModule`, which looks for `packages/<pkg>/mocks/<ModuleName>` and otherwise falls back to the generated table in `jest-expo/src/preset/moduleMocks/expoModules.js`. `expo-video` had no `mocks/` directory, and the generated table only describes flat functions and constants — it has no way to express a SharedObject **class** — so `ExpoVideo.VideoPlayer` was `undefined`.

There are two independent defects behind the single symptom, and fixing only the first still leaves a consuming app broken.

# How

**1. `packages/expo-video/mocks/ExpoVideo.ts` (new).**

A hand-written mock covering exactly the six members `NativeVideoModule.ts` declares on `ExpoVideoModule` — `VideoPlayer`, `VideoThumbnail`, `isPictureInPictureSupported`, `setVideoCacheSizeAsync`, `clearVideoCacheAsync`, `getCurrentVideoCacheSize` — following the conventions in `packages/expo-file-system/mocks/FileSystem.ts`.

`VideoPlayer` is a real `class` extending the `SharedObject` polyfill rather than an object factory. That matters in two places: `VideoPlayer.tsx` reads `.prototype.replace`/`.prototype.replaceAsync`, and `VideoView.tsx:103` narrows with `player instanceof NativeVideoModule.VideoPlayer`. An arrow function or plain factory has no `.prototype` and is not constructible, so it would not survive either. The player keeps a little in-memory state (playing flag, current time, current source) so tests can assert on `play()`/`pause()`/`seekBy()`/`replace()` instead of inert stubs.

`VideoThumbnail` spells out a `release()` override that only forwards to `SharedRef`. This is deliberate: `requireMockModule` decides whether an export is a class by counting own properties on its prototype, so a class carrying nothing but instance fields is mistaken for a plain function and wrapped in `jest.fn()` — which makes it throw `Cannot call a class as a function` on construction. `FileSystemUploadTask` spells `release` out for the same reason.

**2. `packages/jest-expo/src/preset/resolveExistingFile.js` (new), used by `attemptLookup`.**

The mock alone fixes tests that import `expo-video` from source, but *not* a consuming app, which resolves the published `build/` output. `attemptLookup` walks the stack to find the file that called `requireNativeModule`, and skips any frame whose file does not exist on disk. Frames from a built package are source-mapped back to the original TypeScript path — `build/NativeVideoModule.js.map` carries `"sources": ["NativeVideoModule.ts"]`, so the frame reads `build/NativeVideoModule.ts`, which is never emitted. The frame was skipped, the lookup returned `null`, and the mock was never found.

`expo-file-system` is immune only by luck: its call site lives in `ExpoFileSystem.js`, and `attemptLookup`'s first check is `fileName.includes(moduleName)` — `'ExpoFileSystem.js'.includes('FileSystem')` short-circuits before the existence check. `expo-video`'s call site is `NativeVideoModule.ts`, which does not contain `ExpoVideo`, so it depends on the fallback path that was broken.

The fix resolves a frame to the sibling `.js` when the source-mapped path is absent. It is extracted into its own module so it can be unit-tested without executing the setup file's mocking side effects.

**3. Supporting changes.** `expo-video` gains a `test` script, the `expo-module-scripts` jest preset, and a `jest-expo` devDependency (matching `expo-file-system`), and its `tsconfig.json` now type-checks `mocks/` alongside `src/` so the mock cannot silently drift from the module it stands in for.

No `package.json` `files`/`.npmignore` change is needed — `mocks/` already ships (verified with `npm pack`, below).

# Test Plan

**New tests.** `packages/expo-video/src/__tests__/VideoPlayer-test.native.ts` (8 tests × iOS/Android) covers module load, the exact member list the native module declares, `instanceof`, both prototype-patched `replace`/`replaceAsync` wrappers including the deprecation warning, playback state, and thumbnail construction. `packages/jest-expo/tests/__tests__/resolveExistingFile-test.js` covers the frame resolution.

```
$ cd packages/expo-video && pnpm test
PASS Android src/__tests__/VideoPlayer-test.native.ts
PASS iOS src/__tests__/VideoPlayer-test.native.ts
Tests:       16 passed, 16 total

$ cd packages/jest-expo && pnpm test
Test Suites: 29 passed, 29 total
Tests:       113 passed, 113 total
```

**Counterfactual.** With `mocks/ExpoVideo.ts` removed, the new test fails with the original error, at the original location:

```
● Test suite failed to run
    TypeError: Cannot read properties of undefined (reading 'prototype')
    >  9 | const replace = NativeVideoModule.VideoPlayer.prototype.replace;
      at Object.prototype (src/VideoPlayer.tsx:9:47)
```

Restoring it returns 16/16.

**The user-visible path** — a consuming app resolving the built package, which is what the issue actually reports. Run from `apps/test-suite` with `preset: 'jest-expo'`, importing by package name so it resolves `packages/expo-video/build/index.js`:

```js
import { useVideoPlayer, VideoView, createVideoPlayer, isPictureInPictureSupported } from 'expo-video';

it('a consuming app can import expo-video and drive a player', () => {
  expect(typeof useVideoPlayer).toBe('function');
  expect(VideoView).toBeDefined();
  expect(isPictureInPictureSupported()).toBe(false);

  const player = createVideoPlayer('https://example.com/v.mp4');
  player.replace('https://example.com/w.mp4', true);
  player.play();
  expect(player.playing).toBe(true);
});
```

With the mock but *without* the `attemptLookup` fix this still failed at `build/VideoPlayer.tsx:9:47`; with both it passes. (Scratch file, not committed.)

**No regressions.** `expo-sqlite` 10 suites / 190 passed and `expo-file-system` 10 suites / 218 passed are unchanged, confirming the `attemptLookup` change does not disturb packages that already resolved their mocks.

**Packaging.** `npm pack` on `expo-video` contains `package/mocks/ExpoVideo.ts`.

`et check-packages expo-video jest-expo` passes (build, typecheck, lint, format, test).

# Checklist

- [x] Added a `CHANGELOG.md` entry to `expo-video` and `jest-expo`
- [x] Builds, type-checks, lints, and tests via `et check-packages expo-video jest-expo`
- [ ] Not applicable — no native or config-plugin change, so `npx expo prebuild` / EAS Build are unaffected

# Credit

@behenate proposed a mock for this in #39707, which was closed unreviewed in a sweep of PRs older than six months rather than on its merits. This PR is independent but takes the same direction. Two notes on the differences, both from re-testing against current `main`:

- #39707 defined `VideoPlayer` as an arrow function returning an object literal. On today's `main` that would not fix the crash — an arrow function has no `.prototype`, so line 9 would still throw, and `createVideoPlayer`'s `new NativeVideoModule.VideoPlayer(...)` plus `VideoView`'s `instanceof` check both need a real constructor.
- #39707 also removed `ExpoVideo` from the generated mock table. I left that alone: `ExpoAsset`, `ExpoLocation`, `ExpoAgeRange`, and `ExpoAppIntegrity` all ship a hand-written `mocks/` directory *and* keep their generated entry, so removing it is not the established convention, and `attemptLookup` takes priority over the table regardless. Happy to add it if you'd prefer the consistency.


behenate:
Hi! Thank you for the PR! 
I did not like that this PR also introduced changes to `jest-expo`. Instead I made the expo-video mock structure comply with the existing `jest-expo` rules by renaming the source files to module name. I also extended the mock a little bit so that it's more functional.

behenate (approved):

behenate:
Merging as the CI e2e errors don't look to be related to this PR

PR #48882, opened 2026-08-13
fix(patch-project): Use development mode for env files

ramonclaudio:
# Why

Patch Project gets `setNodeEnv` from Expo CLI and then loads env files in a separate step, which can cause inconsistencies.

# How

I updated it to use the shared API with `development` mode for both env files and config.

# Test Plan

Tests and CI checks pass.

# Checklist

### DIFF
--- a/packages/patch-project/src/cli/patchProjectAsync.ts
+++ b/packages/patch-project/src/cli/patchProjectAsync.ts
@@ -44,6 +44,7 @@ export async function patchProjectAsync(
     const envInfo = loadProjectEnv(projectRoot, { mode: 'development' });
     logLoadedEnv(envInfo);
   } finally {
+    // Prevents an inherited mode from reaching the app config.
     delete process.env.EXPO_CONFIG_MODE;
   }
 
--- a/packages/patch-project/src/cli/patchProjectAsync.ts
+++ b/packages/patch-project/src/cli/patchProjectAsync.ts
@@ -1,3 +1,4 @@
+import { loadProjectEnv, logLoadedEnv } from '@expo/env';
 import chalk from 'chalk';
 import { getConfig, type ExpoConfig } from 'expo/config';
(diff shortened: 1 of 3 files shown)


packages/patch-project/src/cli/patchProjectAsync.ts
```diff
@@ -39,12 +40,12 @@ export async function patchProjectAsync(
...
-    resolveFromExpoCli(projectRoot, 'build/src/utils/nodeEnv')
-  ) as typeof import('@expo/cli/src/utils/nodeEnv');
-
-  setNodeEnv('development');
-  require('@expo/env').load(projectRoot);
+  try {
+    const envInfo = loadProjectEnv(projectRoot, { mode: 'development' });
+    logLoadedEnv(envInfo);
+  } finally {
+    delete process.env.EXPO_CONFIG_MODE;
```
kitten: Might be worth leaving a comment here or having a reset method/flag somewhere? I assume we do this to protect against inheritance though, so not entirely sure if it's worth maybe even leaving out?

kitten (approved):
Big fan of witching this over to just using `@expo/env` directly, since the semantics here haven't changed in a while, except for public API additions

packages/patch-project/src/cli/patchProjectAsync.ts
ramonclaudio: Yea exactly, it's there to prevent an inherited mode from reaching the app config. I left a comment to make that clear.

ramonclaudio: Following up, @Kudo asked about the same thing in #48839. I moved the cleanup into `loadProjectEnv` in #48938, so once that merges I’ll remove this delete and comment from here.

Kudo (approved):

PR #48883, opened 2026-08-13
[observe][iOS][Android] Export spans as OTLP traces

tsapeta:
# Why

Second half of the traces feature: expo/expo#48861 records network requests as generic span rows; this PR ships them to the `/v1/traces` ingestion endpoint (expo/universe#29874) on iOS and Android.

Closes ENG-25894.

# How

Spans dispatch alongside metrics and logs with their own retry gate (and mutex on Android). The mapping onto the OTLP wire shape is producer-agnostic: rows already carry their identifiers, attributes, and events, so the exporter only handles timestamps, value typing, and the per-session resource envelope, and future span producers need no changes here.

Unlike metrics and logs there's no persisted cursor or pending-ids store: nothing reads span rows back, so a dispatched (or deliberately dropped) batch is deleted outright and the table itself is the queue. Batches are chunked at the server's 512-span limit, `rejectedSpans` counts in partial-success handling, and on iOS an in-flight guard keeps overlapping dispatch calls from re-sending a backlog.

# Test Plan

New unit suites on both platforms cover the row-to-wire mapping (typed attribute decoding, malformed-blob tolerance, per-event timestamps, event caps, overflow-safe timestamps), the request envelope, and `rejectedSpans` decoding. `et native-unit-tests` passes for expo-observe on iOS (Swift Testing) and Android (Robolectric).

### DIFF
--- a/packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityBackgroundWorker.kt
+++ b/packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityBackgroundWorker.kt
@@ -17,6 +17,7 @@ import androidx.work.WorkerParameters
 import androidx.work.workDataOf
 import expo.modules.observe.storage.PendingLogsManager
 import expo.modules.observe.storage.PendingMetricsManager
+import expo.modules.appmetrics.storage.MetricsDatabase
 import expo.modules.appmetrics.storage.SessionManager
 
 /**
@@ -52,7 +53,8 @@ class ObservabilityBackgroundWorker(
       pendingMetricsManager = pendingMetricsManager,
       pendingLogsManager = pendingLogsManager,
       baseUrl = baseUrl,
-      isDebugBuild = BuildConfig.DEBUG
+      isDebugBuild = BuildConfig.DEBUG,
+      spanDao = MetricsDatabase.getDatabase(context).spanDao()
     )
   }
 
@@ -83,6 +85,7 @@ class ObservabilityBackgroundWorker(
       observabilityManager.cleanup()
       observabilityManager.dispatchUnsentMetrics()
       observabilityManager.dispatchUnsentLogs()
+      observabilityManager.dispatchUnsentSpans()
       Log.d(OBSERVE_TAG, "Successfully dispatched unsent metrics and logs")
       Result.success()
     } catch (e: Exception) {
--- a/packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityBackgroundWorker.kt
+++ b/packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityBackgroundWorker.kt
@@ -18,7 +18,7 @@ import androidx.work.workDataOf
 import expo.modules.appmetrics.storage.SessionManager
 
 /**
- * Background worker that dispatches stored metrics and logs to EAS Observe.
+ * Background worker that dispatches stored metrics, logs, and spans to EAS Observe.
  */
 class ObservabilityBackgroundWorker(
   context: Context,
@@ -72,10 +72,11 @@ class ObservabilityBackgroundWorker(
       observabilityManager.cleanup()
       observabilityManager.dispatchUnsentMetrics()
       observabilityManager.dispatchUnsentLogs()
-      Log.d(OBSERVE_TAG, "Successfully dispatched unsent metrics and logs")
+      observabilityManager.dispatchUnsentSpans()
+      Log.d(OBSERVE_TAG, "Successfully dispatched unsent metrics, logs, and spans")
       Result.success()
     } catch (e: Exception) {
-      Log.e(OBSERVE_TAG, "Failed to dispatch metrics", e)
+      Log.e(OBSERVE_TAG, "Failed to dispatch metrics, logs, and spans", e)
       // Retry with exponential backoff
       Result.retry()
     }
@@ -108,7 +109,7 @@ class ObservabilityBackgroundWorker(
         .getInstance(context)
         .enqueueUniqueWork(
           WORK_NAME,
-          // Keep an in-flight dispatch; cancelling it can duplicate a request the server received.
+          // Keep an in-flight dispatch; canceling it can duplicate a request the server received.
           ExistingWorkPolicy.KEEP,
           periodicWork
         )
--- a/packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
+++ b/packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
@@ -5,7 +5,9 @@ import android.util.Log
 import expo.modules.easclient.EASClientID
 import expo.modules.observe.storage.PendingLogsManager
 import expo.modules.observe.storage.PendingMetricsManager
+import expo.modules.appmetrics.storage.MetricsDatabase
 import expo.modules.appmetrics.storage.SessionManager
+import expo.modules.appmetrics.storage.SpanDao
 import expo.modules.appmetrics.utils.TimeUtils
 import expo.modules.interfaces.constants.ConstantsInterface
 import kotlinx.coroutines.sync.Mutex
@@ -42,7 +44,8 @@ class ObservabilityManager(
       pendingLogsManager = pendingLogsManager,
       projectId = projectId,
       baseUrl = baseUrl,
-      isDebugBuild = BuildConfig.DEBUG
+      isDebugBuild = BuildConfig.DEBUG,
+      spanDao = MetricsDatabase.getDatabase(context).spanDao()
     )
 
     sessionManager.addMetricsInsertListener { metricIds ->
@@ -59,6 +62,10 @@ class ObservabilityManager(
 
   suspend fun dispatchUnsentLogs() {
     baseManager.dispatchUnsentLogs()
+  }
+
+  suspend fun dispatchUnsentSpans() {
+    baseManager.dispatchUnsentSpans()
   }
 
   fun scheduleBackgroundDispatch() {
@@ -81,8 +88,20 @@ class BaseObservabilityManager(
   private val deterministicUniformValueProvider: () -> Double = {
     EASClientID.deterministicUniformValue(EASClientID(context).uuid)
   },
-  private val currentTimeMs: () -> Long = { TimeUtils.getWallClockMillis() }
+  private val currentTimeMs: () -> Long = { TimeUtils.getWallClockMillis() },
+  /**
+   * Source of persisted spans, or `null` to disable the traces signal (kept optional so
+   * existing constructions and tests that don't exercise spans stay valid).
+   */
+  private val spanDao: SpanDao? = null
 ) {
+  companion object {
+    /**
+     * Maximum spans per request accepted by the ingestion endpoint; it rejects everything past
+     * this count within one POST, so larger backlogs are sent as sequential chunks.
+     */
+    const val MAX_SPANS_PER_REQUEST = 512
+  }
   private val eventDispatcher = EventDispatcher(
     context = context,
     projectId = projectId,
@@ -104,6 +123,7 @@ class BaseObservabilityManager(
    */
   private var metricsRetryGate: DispatchUtils.RetryGateState = DispatchUtils.RetryGateState.initial
   private var logsRetryGate: DispatchUtils.RetryGateState = DispatchUtils.RetryGateState.initial
+  private var spansRetryGate: DispatchUtils.RetryGateState = DispatchUtils.RetryGateState.initial
 
   /**
    * Per-signal mutexes that serialize same-signal dispatch calls. Two `dispatchEvents`
@@ -118,6 +138,7 @@ class BaseObservabilityManager(
    */
   private val metricsDispatchMutex = Mutex()
   private val logsDispatchMutex = Mutex()
+  private val spansDispatchMutex = Mutex()
 
   /**
    * Returns true and logs when an active retry gate suppresses this dispatch round. Called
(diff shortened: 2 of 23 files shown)


packages/expo-observe/ios/Observability.swift
```diff
@@ -206,29 +228,137 @@ internal struct ObservabilityManager {
...
+        }
+      } catch {
+        observeLogger.warn("[EAS Observe] Failed to drop undispatched spans: \(error.localizedDescription)")
+      }
+      return
+    }
+
+    let pendingSpans: [SpanRow]
+    do {
+      pendingSpans = try AppMetrics.getSpans(afterId: -1)
```
tsapeta: [🏗️ **design**] iOS reads the entire span backlog into memory in one query, while the Android path (`ObservabilityManager.kt`) reads one `MAX_SPANS_PER_REQUEST` chunk per query and re-queries after each chunk. The `SpanDispatchLoop.drain` doc says "pending spans in id order" and chunks in memory, so `chunkSize` bounds the POST but not the read.

This is bounded by the 2,000-row insert cap, so it won't run away, but each row carries `attributes` and `events` JSON blobs, and this runs on `applicationWillResignActive` and `applicationWillTerminate`. Two secondary effects of not re-querying: spans inserted during a long drain wait for the next dispatch, and rows already deleted by a preceding chunk stay resident in the array.

Passing `limit: maxSpansPerRequest` and letting the loop re-fetch (as `DispatchLoop.drain` already does for metrics/logs via `fetchBatch`) would make the two platforms agree and cap the resident set at one chunk.

packages/expo-observe/ios/SpanDispatchLoop.swift
```diff
@@ -0,0 +1,85 @@
...
+    var chunks: [[SpanRow]] = stride(from: 0, to: rows.count, by: chunkSize).map { chunkStart in
+      return Array(rows[chunkStart..<min(chunkStart + chunkSize, rows.count)])
+    }
+    while !chunks.isEmpty {
+      let chunk = chunks.removeFirst()
+      let result: DispatchResult?
+      do {
+        result = try await send(chunk)
+      } catch {
+        observeLogger.warn("[EAS Observe] Failed to assemble trace events: \(error.localizedDescription)")
```
tsapeta: [🏗️ **design**] When `send` throws, iOS returns without deleting, so the rows stay and are retried on every future dispatch. On Android the equivalent failure (`OTTracesRequestBody(...).toJson(...)` throwing inside `dispatchSpans`) is caught and collapsed to `NonRetryableFailure`, which the disposition maps to `DELETE_AND_CONTINUE`.

So the same condition drops the chunk on Android and pins it forever on iOS. The iOS `send` closure calls `buildResourceSpans`, whose only throwing call is `AppMetrics.getSessions(ids:)` — a database read that, if it fails deterministically (corrupt db, disk full), leaves these rows permanently blocking the table until the insert cap evicts them.

`SpanDispatchLoopTests."a send that throws stops without deleting"` locks in the current behavior with the rationale "assembling the payload failed, which says nothing about the server". That reasoning is sound for a transient failure, but it makes a persistent one unrecoverable. Worth either matching Android's drop, or bounding the retry.

packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
```diff
@@ -276,6 +282,84 @@ class BaseObservabilityManager(
...
+          return
+        }
+        readCursor = nextChunk.last().id
+        chunks.addLast(nextChunk)
+      }
+      val chunk = chunks.removeFirst()
+      val batches = chunk.groupBy { it.sessionId }.mapNotNull { (sessionId, spans) ->
+        // Spans whose session row no longer exists are skipped, because their rows are about to be
+        // deleted below anyway.
+        val session = sessionManager.getSessionRow(sessionId) ?: return@mapNotNull null
```
tsapeta: [💡 suggestion] This issues one `getSessionRow` query per distinct session inside the `groupBy`, whereas `dispatchUnsentMetrics` and `dispatchUnsentLogs` (same file, just above) batch the lookup: `sessionManager.getSessions(bySessionId.keys).associateBy { it.id }`. iOS batches too, via `buildPerSession` → `AppMetrics.getSessions(ids:)`.

In the common case a chunk covers one or two sessions, so the cost is small. But after the app has been offline for a while the backlog can span many sessions, and each halving retry re-runs every lookup for the halves. Reusing the batched read would match both the sibling signals in this file and the iOS side, and would remove the per-retry re-query.

packages/expo-observe/ios/Observability.swift
```diff
@@ -22,14 +27,31 @@ internal struct ObservabilityManager {
...
+  /// `await`, so overlapping `dispatch()` calls (a JS `dispatchEvents` racing the resign-active
+  /// app delegate hook) would otherwise read (and send) the same pending rows twice before
+  /// either pass advances its cursor or deletes its batch.
+  private static var isDispatching = false
 
   internal static func dispatch() async {
-    // Per-signal gates are checked inside `dispatchMetrics` / `dispatchLogs` rather than
-    // here, so a backoff on one endpoint doesn't suppress the other's traffic.
+    if isDispatching {
+      observeLogger.debug("[EAS Observe] Dispatch already in progress; skipping")
```
tsapeta: [🏗️ **design**] The in-flight guard is a real fix for the overlapping-dispatch race, and the reentrancy rationale in the doc comment is correct. Two things it changes that aren't called out:

1. `dispatchEvents()` is a public JS API (`ObserveModule.swift:43`). Its promise now resolves without having dispatched anything when a resign-active pass happens to be in flight. A caller that awaits `dispatchEvents()` before, say, backgrounding or a test assertion can no longer treat resolution as "the queue was drained". If that's acceptable, it's worth saying so in the JSDoc for `dispatchEvents`; if not, awaiting the in-flight pass would preserve the contract.

2. Android has no equivalent guard at this level — it relies on the three per-signal mutexes, which serialize rather than skip. So a concurrent JS `dispatchEvents` on Android waits and then dispatches, while on iOS it returns immediately. Divergent enough to be worth a comment noting the deliberate difference.

Also: the changelog entry ("Export network requests as OTLP traces") doesn't mention this behavior change to an existing API.

packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
```diff
@@ -276,6 +282,84 @@ class BaseObservabilityManager(
...
+        is DispatchResult.NonRetryableFailure ->
+          Log.w(OBSERVE_TAG, "Dropping batch of ${chunk.size} span(s): ${result.reason}")
+        DispatchResult.PayloadTooLarge -> if (chunk.size == 1) {
+          Log.w(OBSERVE_TAG, "Dropping span that exceeds the server's payload limit")
+        }
+        DispatchResult.Success, is DispatchResult.RetryableFailure -> Unit
+      }
+      when (spanDispatchDisposition(result, chunk.size)) {
+        SpanDispatchDisposition.HALVE_AND_RETRY -> {
+          chunks.addFirst(chunk.subList(chunk.size / 2, chunk.size))
```
tsapeta: [🐧 nit] `chunk.subList(...)` returns a view backed by the original chunk, and both halves are pushed onto the deque. Repeated halving therefore builds nested views that keep the full original chunk (up to 512 rows with their JSON blobs) reachable until every descendant has been processed. iOS avoids this by materializing with `Array(chunk[...])`.

Nothing mutates the backing list, so this is not a correctness bug — just retention. Wrapping each half in `toList()` would match iOS and let the parent chunk be collected.

packages/expo-observe/ios/OpenTelemetry.swift
```diff
@@ -265,6 +307,87 @@ extension Event.Log {
...
+/// Maximum span events the ingestion endpoint keeps per span; anything past it is counted as
+/// dropped server-side, so the SDK truncates locally and reports the loss itself instead of
+/// shipping payload that is guaranteed to be discarded.
+private let maxSpanEventCount = 32
+
+/// Converts a persisted unix-epoch millisecond timestamp to nanoseconds, saturating instead of
+/// trapping: a corrupt row or a device clock set far into the future must never crash the host
+/// app from inside the telemetry library.
+private func spanNanoseconds(fromMilliseconds milliseconds: Int64) -> UInt64 {
+  let maxRepresentableMs = Int64(UInt64.max / 1_000_000)
```
tsapeta: [🐧 nit] The two platforms saturate at different ceilings: iOS clamps to `UInt64.max / 1_000_000` (≈ year 2554) because it targets `UInt64` nanos, Android to `Long.MAX_VALUE / 1_000_000` (≈ year 2262) for `Long`. Neither wraps negative, so the stated goal is met on both.

Since the comment says the backend stores `DateTime64(3)`, both ceilings are far past anything it can represent, and a row that hits either is corrupt regardless. If the intent is that the two mappings agree (the Android doc says "Mirrors the iOS `SpanRow.toOTSpan`"), a shared explicit ceiling would say so; otherwise a note that the ceilings differ by storage type would prevent someone reading it as a bug later.

packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityBackgroundWorker.kt
```diff
@@ -72,6 +72,7 @@ class ObservabilityBackgroundWorker(
       observabilityManager.cleanup()
       observabilityManager.dispatchUnsentMetrics()
       observabilityManager.dispatchUnsentLogs()
+      observabilityManager.dispatchUnsentSpans()
       Log.d(OBSERVE_TAG, "Successfully dispatched unsent metrics and logs")
```
tsapeta: [🐧 nit] The log line still says "metrics and logs" now that spans dispatch here too.

packages/expo-observe/ios/Observability.swift
```diff
@@ -206,29 +228,137 @@ internal struct ObservabilityManager {
...
+          return nil
+        }
+        let body = OTTracesRequestBody(resourceSpans: resourceSpans)
+        let result = await DispatchUtils.sendRequest(to: endpointUrl, body: body)
+        applyRetryOutcome(result, to: &tracesRetryGate)
+        switch result {
+        case .success:
+          ObserveUserDefaults.lastDispatchDate = Date.now
+        case .partialSuccess(let partial):
+          ObserveUserDefaults.lastDispatchDate = Date.now
```
tsapeta: [🐧 nit] `dispatchTraces` sets `ObserveUserDefaults.lastDispatchDate` on `.success` and `.partialSuccess`, matching `dispatchMetrics`. `dispatchLogs` additionally sets it on `.retryableFailure` (`case .success, .retryableFailure:`).

That inconsistency predates this PR, so traces following the metrics shape is a reasonable choice. Flagging only because whichever of the three is wrong now has two votes against one, and it'd be good to settle it while the dispatch code is fresh.

packages/expo-observe/ios/SpanDispatchLoop.swift
```diff
@@ -0,0 +1,114 @@
...
+
+      let chunk = chunks.removeFirst()
+      let result: DispatchResult?
+      do {
+        result = try await send(chunk)
+      } catch {
+        // Assembly failed rather than the request. A transient cause (a busy database) resolves
+        // on the next dispatch, but a persistent one would pin these rows until the insert cap
+        // evicted them, blocking every span behind them, so the chunk is dropped. Android
+        // collapses the same condition into a non-retryable failure, which also drops.
```
Ubax: [🏗️ **design**] The only throwing call inside the `send` closure is `AppMetrics.getSessions` (a database read): `toOTSpan()` doesn't throw and JSON encoding happens inside `DispatchUtils.sendRequest`, which returns a `DispatchResult`. So this branch deletes up to 512 spans on a transient SQLite error, while the existing `DispatchLoop` keeps its rows in the same situation (`Failed to assemble or send pending rows` → `return`).

The Android comparison also doesn't quite hold: on Android a throwing `getSessions` propagates out of `dispatchUnsentSpans` (the worker catches it and the rows stay); only an *encoding* failure collapses to `NonRetryableFailure`.

Suggest `return` here to match `DispatchLoop` and Android. The insert-time cap already bounds how long a persistently failing chunk can pin the queue, so the "blocks every span behind it" concern is bounded too. The test `a send that throws drops its chunk and continues` would flip to keep-and-stop.

packages/expo-observe/ios/Observability.swift
```diff
@@ -22,14 +27,36 @@ internal struct ObservabilityManager {
...
+  /// Whether a dispatch pass is currently running. The actor is reentrant at every network
+  /// `await`, so overlapping `dispatch()` calls (a JS `dispatchEvents` racing the resign-active
+  /// app delegate hook) would otherwise read (and send) the same pending rows twice before
+  /// either pass advances its cursor or deletes its batch.
+  ///
+  /// Skipping rather than awaiting is deliberate: the caller is a fire-and-forget lifecycle hook
+  /// or a manual flush, and the in-flight pass is already draining the same rows. It does mean a
+  /// JS `dispatchEvents()` can resolve without having sent anything, which the JSDoc for that
+  /// method states. Android serializes instead, through its per-signal mutexes.
+  private static var isDispatching = false
```
Ubax: [🏗️ **design**] This guard changes the `dispatchEvents()` contract for metrics and logs too, not only spans: the JSDoc's own use case ("ensure events are sent before a specific point") is no longer guaranteed on iOS when a resign-active flush is in flight.

Storing the in-flight `Task` and awaiting it gives Android-equivalent semantics in a few lines, e.g.

```swift
private static var inFlight: Task<Void, Never>?

internal static func dispatch() async {
  if let inFlight { await inFlight.value }
  let task = Task { await runDispatch() }
  inFlight = task
  defer { inFlight = nil }
  await task.value
}
```

That removes the platform caveat added to `types.ts` and keeps the duplicate-send protection this guard is here for.

packages/expo-observe/src/types.ts
```diff
@@ -167,7 +167,13 @@ export declare class ObserveModule extends NativeModule<ObserveModuleEvents> {
    * this method to flush events manually, for example, during testing or to ensure events
    * are sent before a specific point.
    *
-   * @returns A promise that resolves when the pending events have been dispatched.
+   * > Note: On iOS, a call made while another dispatch is already running (an automatic
```
Ubax: [💡 suggestion] If the in-flight guard in `Observability.swift` awaits instead of skipping (see the comment there), this note and the `@returns` caveat can go and both platforms share one contract. Otherwise fine as written.

packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityBackgroundWorker.kt
```diff
@@ -72,7 +72,8 @@ class ObservabilityBackgroundWorker(
       observabilityManager.cleanup()
       observabilityManager.dispatchUnsentMetrics()
       observabilityManager.dispatchUnsentLogs()
-      Log.d(OBSERVE_TAG, "Successfully dispatched unsent metrics and logs")
+      observabilityManager.dispatchUnsentSpans()
+      Log.d(OBSERVE_TAG, "Successfully dispatched unsent metrics, logs, and spans")
```
Ubax, line 76: [🐧 nit] The `catch` two lines below still logs `Failed to dispatch metrics`, but it now covers logs and spans as well. Worth updating alongside this success message.

packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
```diff
@@ -276,6 +282,92 @@ class BaseObservabilityManager(
...
+   * server's per-request span limit.
+   */
+  suspend fun dispatchUnsentSpans(): Unit = spansDispatchMutex.withLock {
+    if (retryGateBlocks(spansRetryGate, "spans")) {
+      return
+    }
+
+    if (!shouldDispatch()) {
+      // Drop the backlog without materializing it; one MAX query is enough to know how far to
+      // delete. Mirrors metrics/logs advancing the cursor past rows they won't send.
```
Ubax:
```suggestion
      // Mirrors metrics/logs advancing the cursor past rows they won't send.
```

packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
```diff
@@ -276,6 +282,92 @@ class BaseObservabilityManager(
...
+          return
+        }
+        readCursor = nextChunk.last().id
+        chunks.addLast(nextChunk)
+      }
+      val chunk = chunks.removeFirst()
+      val spansBySessionId = chunk.groupBy { it.sessionId }
+      // One batched read rather than a query per session, matching the metrics and logs paths
+      // above and the iOS side. A backlog accumulated while offline can span many sessions, and
+      // each halving retry would otherwise re-run every lookup.
```
Ubax: I am not sure if this adds any more information

packages/expo-observe/android/src/main/java/expo/modules/observe/OpenTelemetry.kt
```diff
@@ -134,6 +135,58 @@ data class OTResourceLogs(
   val schemaUrl: String
 )
 
+/**
+ * One OTLP span. Nullable fields with defaults (`parentSpanId`, `status`) rely on
+ * kotlinx-serialization omitting defaults: the ingestion endpoint rejects a span whose present
+ * `parentSpanId` isn't valid hex, so a root span must omit the key entirely rather than send
+ * null or "".
```
Ubax: Let's add a link here to spec

https://opentelemetry.io/docs/concepts/signals/traces/#spans

packages/expo-observe/src/types.ts
```diff
@@ -167,7 +167,13 @@ export declare class ObserveModule extends NativeModule<ObserveModuleEvents> {
...
+   * > resign-active flush, or an earlier call) returns without dispatching, so resolution does
+   * > not guarantee the queue was drained. Android serializes such calls instead, and dispatches
+   * > once the one ahead of it finishes.
+   *
+   * @returns A promise that resolves once this call has finished, which is immediately when a
+   * dispatch was already in progress on iOS.
```
Ubax: As for public API this adds a lot of details 😅 

I'm not sure if I understand what it means

Ubax (approved):
Please try to simplify the comments in the same way as the previous PR

PR #48886, opened 2026-08-13
[router] Add RoutingIntent to router-queue, to unify action dispatching

Ubax:
# Why

# How

# Test Plan

# Checklist

### DIFF
--- a/packages/expo-router/src/global-state/__tests__/routingQueue.test.ios.ts
+++ b/packages/expo-router/src/global-state/__tests__/routingQueue.test.ios.ts
@@ -45,14 +45,17 @@ beforeEach(() => {
 });
 
 describe('routingQueue', () => {
-  it('add() pushes action to queue and notifies subscribers', () => {
+  it('add() pushes intent to queue and notifies subscribers', () => {
     const callback = jest.fn();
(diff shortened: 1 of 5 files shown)


packages/expo-router/src/global-state/__tests__/routingQueue.test.ios.ts
```diff
@@ -129,12 +137,26 @@ describe('routingQueue', () => {
     expect(ref.current!.dispatch).toHaveBeenCalledWith(navigateAction);
   });
 
+  it('run() dispatches mixed NAVIGATE_TO_HREF and ACTION intents in queue order', () => {
+    const ref = makeRef();
+    const navigateAction = { type: 'NAVIGATE', payload: { name: 'a' } };
```
jakub-agent: [🚨 **critical**] This fixture is inferred as `{ type: string; payload: { name: string } }`, but `getNavigateAction` requires the full navigate action shape (including `target`, `params`, and `singular`). As a result `expo-router:typecheck` fails in CI at the following `mockReturnValueOnce` call. Use a correctly typed action fixture (for example, the one in the preceding test) so this test compiles.

Issue #46519, opened 2026-06-03
[SDK 56] Worklets memory regression with Hermes v1


rodperottoni:
### Minimal reproducible example

https://github.com/rodperottoni/expo-sdk-56-mem-regression

### Steps to reproduce

I have provided an example repository with a sandbox for both Expo SDK 56 and 55. The `main` branch is SDK 56, and the `sdk-55` branch is SDK 55.
Checkout to each branch and run `npm i --force` + `npm run compile` + `npm run ios`.
Then, open the performance monitor as soon as the app opens.

## Findings

On SDK 56 RAM usage almost doubles in this basic example (149mb -> 266mb):

| SDK 55 | SDK 56 |
|--------|--------|
| <img width="1198" height="2172" alt="Image" src="https://github.com/user-attachments/assets/821c0bdc-6648-4e27-a638-0f16dc34666a" /> | <img width="1198" height="2172" alt="Image" src="https://github.com/user-attachments/assets/628f11c7-afb7-4286-b8ca-0b78f81c8bd3" /> | 

This increase is more noticeable in complex apps. In my employer's production app we've jumped from ~580mb up to 1.1GB in the home page of our app.

Further scenarios I tested:
- SDK 55 + `useHermesV1: true`: Increased RAM usage
  - This scenario is in the `sdk-55-with-hermes-v1` branch of the example repo 
- SDK 56 + `useHermesV1: false`: Increased RAM usage (disabling this flag seems to not do anything?)
  - This scenario is in the `sdk-56-not-hermes-v1` branch of the example repo

## What I have tried

The only change in SDK 56 I initially thought could've caused this is Hermes V1. Disabling Hermes v1 via `expo-build-properties` doesn't change a thing.

### Environment

```text
expo-env-info 2.0.14 environment info:
    System:
      OS: macOS 26.5
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 24.15.0 - /Users/rudy/.nvm/versions/node/v24.15.0/bin/node
      npm: 11.12.1 - /Users/rudy/.nvm/versions/node/v24.15.0/bin/npm
      Watchman: 2026.05.04.00 - /opt/homebrew/bin/watchman
    Managers:
      CocoaPods: 1.16.2 - /Users/rudy/.rbenv/shims/pod
    SDKs:
      iOS SDK:
        Platforms: DriverKit 25.5, iOS 26.5, macOS 26.5, tvOS 26.5, visionOS 26.5, watchOS 26.5
[...]

PR #48887, opened 2026-08-13
feat(expo-doctor): Add Hermes v1 check targeting memory regression and add upgrade recommendation

kitten:
# Why

Related to #46519

We have a few Hermes v1 versions where newer ones resolve the memory regression, reported in the issue above. However, these weren't backported into React Native versions that are shipped by SDK 55 and 56. Hence, we'll have to detect exact versions (including on 57) and issue upgrade recommendations more forcefully to avoid confusion.

> [!WARNING]
> We have to issue additional notices on the issue above, and pin those. We should also check if there are any issues related to this that are duplicates and we haven't closed/caught yet as such.

# How

- Add SDK version upgrade recommendations as needed for Hermes v1 memory regression
- Add Babel config check for Worklets mode (only on affected versions, to avoid users leaving it on)
- Add `hermes-engine` and `hermes-compiler` checks to catch edge cases

# Test Plan

- Unit tests added

**Note for internal review:** See referenced PR on changelog blog post updates

# Checklist

### DIFF
--- a/packages/expo-doctor/CHANGELOG.md
+++ b/packages/expo-doctor/CHANGELOG.md
@@ -12,6 +12,8 @@
 
 ### 🐛 Bug fixes
 
+- [Internal] Prevent `ncc` from removing dynamic requires where we need them ([#48887](https://github.com/expo/expo/pull/48887) by [@kitten](https://github.com/kitten))
+
 ### 💡 Others
 
 ## 1.20.1 - 2026-07-15
--- a/packages/expo-doctor/CHANGELOG.md
+++ b/packages/expo-doctor/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### 🎉 New features
 
+- Warn projects using Expo SDKs affected by the Hermes v1 memory regression ([#48887](https://github.com/expo/expo/pull/48887) by [@kitten](https://github.com/kitten))
+
 ### 🐛 Bug fixes
 
 ### 💡 Others
--- a/packages/expo-doctor/src/checks/HermesV1VersionCheck.ts
+++ b/packages/expo-doctor/src/checks/HermesV1VersionCheck.ts
@@ -21,18 +21,18 @@ function getInstalledExpoVersion(projectRoot: string): string | null {
   return typeof version === 'string' && semver.valid(version) ? version : null;
 }
 
-function isHermesV1Enabled(exp: DoctorCheckParams['exp']): boolean {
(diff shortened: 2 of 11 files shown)


brentvatne (approved):

kitten:
Added a drive-by fix that fixes this against the `ncc` version that `main` is on that broke our dynamic requires. Tested against an SDK 56 and SDK 57 app respectively

kitten:
Missing a pick for `sdk-55`, but if requested, can pick this and the other missing prior PRs

markaustinws:
Our Azure Pipeline is using expo doctor.  Looks like the latest version now errors on this rather than show a warning

kitten:
@markaustinws: expo-doctor doesn't currently have a concept of warnings and instead fails or succeeds.

markaustinws:
Should it error with this condition though?  This seems to be enforcing an upgrade to Expo 57.  We're currently on Expo 56.

Output as follows:

npm warn exec The following package was not found and will be installed: expo-doctor@1.20.2
20/21 checks passed. 1 checks failed. Possible issues detected:

This project uses Hermes V1 with expo@56.0.19, which is affected by a known memory regression.
Detected Hermes V1 250829098.0.10 from React Native. Hermes V1 250829098.0.15 and earlier are affected by this regression; 250829098.0.16 is the first version that contains the fix.
Advice:
Upgrade to Expo SDK 57 and expo@57.0.9 or later by running `npx expo install expo@^57.0.9 --fix`. See https://expo.dev/changelog/sdk-57#known-regressions for the latest details.

##[error]Bash exited with code '1'.

kitten:
As per the notice in the linked changelog post:
> Updating to expo@57.0.9 or later is crucial for affected apps

and further in 56: https://expo.dev/changelog/sdk-56#known-regressions

> [SDK 57](https://expo.dev/changelog/sdk-57) resolves the regression that caused increased memory usage, and we recommend to upgrade to SDK 57 for all apps that currently use Hermes v1 and SDK 55 or 56. The upgrade from SDK 56 is equivalent to a minor release and designed to be an easy upgrade.

The affected fix will not be backported to previous React Native versions and this is hence a "strong recommendation" to upgrade to resolve the issue.

Regarding the Azure pipeline, we can't maintain or update your pipeline, and with the expectations and design of `expo-doctor` currently, by design it informed you of this by failing. Whether your Azure pipeline should always outright fail on any failures of `expo-doctor` is out of our control, but either way, as intended, you're now aware and informed, as intended 😅

PR #48889, opened 2026-08-13
[router] Use routing queue for deep links and browser history operations

Ubax:
# Why

# How

# Test Plan

# Checklist

### DIFF
--- a/packages/@expo/cli/e2e/playwright/dev/navigator-browser-history.test.ts
+++ b/packages/@expo/cli/e2e/playwright/dev/navigator-browser-history.test.ts
@@ -3,6 +3,7 @@ import { test, expect } from '@playwright/test';
 import { clearEnv, restoreEnv } from '../../__tests__/export/export-side-effects';
 import { getRouterE2ERoot } from '../../__tests__/utils';
 import { createExpoStart } from '../../utils/expo';
+import { pageCollectErrors } from '../page';
 
 test.beforeAll(() => clearEnv());
 test.afterAll(() => restoreEnv());
@@ -39,7 +40,7 @@ test.describe(inputDir, () => {
   // Test expo router history by navigating through <Link>,
   // then using the browser back/forward actions
   test('navigator browser history', async ({ page }) => {
-    page.on('console', (msg) => console.log(msg.text()));
+    const pageErrors = pageCollectErrors(page);
 
     await page.goto(`${expoStart.url}`);
 
@@ -60,5 +61,64 @@ test.describe(inputDir, () => {
     await page.goForward();
 
     await expect(page.locator('[data-testid="explore-content"]')).toHaveText('/explore');
+
+    expect(pageErrors.all).toEqual([]);
+  });
+
+  test('rapid history traverses nested state', async ({ page }) => {
+    const pageErrors = pageCollectErrors(page);
+
+    await page.goto(`${expoStart.url}`);
+    await expect(page.locator('[data-testid="home-content"]')).toHaveText('/');
+
+    await page.locator('[data-testid="go-explore"]').click();
+    await expect(page.locator('[data-testid="explore-content"]')).toHaveText('/explore');
+    await page.locator('[data-testid="go-details"]').click();
+    await expect(page.locator('[data-testid="details-content"]')).toHaveText('/explore/details');
+    await page.locator('[data-testid="go-final"]').click();
+    await expect(page.locator('[data-testid="final-content"]')).toHaveText('/explore/final');
+
+    await page.evaluate(() => {
+      // Wait for the first traversal before starting the second; browsers may collapse synchronous calls.
+      addEventListener('popstate', () => history.back(), { once: true });
+      history.back();
+    });
+    await expect(page).toHaveURL(/\/explore$/);
+    await expect(page.locator('[data-testid="explore-content"]')).toHaveText('/explore');
+
+    await page.evaluate(() => {
+      // Chain the second traversal from `popstate` while the router is still processing the first.
+      addEventListener('popstate', () => history.forward(), { once: true });
+      history.forward();
+    });
+    await expect(page.locator('[data-testid="final-content"]')).toHaveText('/explore/final');
+
+    await page.reload();
+    await expect(page.locator('[data-testid="final-content"]')).toHaveText('/explore/final');
+    await page.goBack();
+    await expect(page.locator('[data-testid="details-content"]')).toHaveText('/explore/details');
+    await page.goForward();
+    await expect(page.locator('[data-testid="final-content"]')).toHaveText('/explore/final');
+
+    expect(pageErrors.all).toEqual([]);
+  });
+
+  test('restores a nested tab stack without remounting', async ({ page }) => {
+    const pageErrors = pageCollectErrors(page);
+
+    await page.goto(`${expoStart.url}`);
+    await page.locator('[data-testid="go-explore"]').click();
+    await page.locator('[data-testid="go-details"]').click();
+    await page.locator('[data-testid="increment-details"]').click();
+    await expect(page.locator('[data-testid="details-count"]')).toHaveText('1');
+
+    await page.locator('[data-testid="go-final"]').click();
+    await expect(page.locator('[data-testid="final-content"]')).toHaveText('/explore/final');
+    await page.goBack();
+    await expect(page).toHaveURL(/\/explore\/details$/);
+    await expect(page.locator('[data-testid="details-content"]')).toHaveText('/explore/details');
+    await expect(page.locator('[data-testid="details-count"]')).toHaveText('1');
+
+    expect(pageErrors.all).toEqual([]);
   });
 });
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -46,6 +46,7 @@
 
 ### 🐛 Bug fixes
 
+- Route deep links and browser history through the navigation queue while preserving saved navigation state.
 - Make layouts with explicitly declared screens honor `unstable_settings.initialRouteName` instead of declaration order, which can change deep-link back stacks. ([#48708](https://github.com/expo/expo/pull/48708) by [@Ubax](https://github.com/Ubax))
 - Prevent unfocused nested native tab navigators from redirecting global router state. ([#48257](https://github.com/expo/expo/pull/48257) by [@Ubax](https://github.com/Ubax))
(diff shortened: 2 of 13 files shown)


packages/expo-router/src/fork/useBrowserHistorySync.ts
```diff
@@ -0,0 +1,307 @@
...
+
+      if (!state) {
+        return;
+      }
+
+      const path = getPathForRoute(findFocusedRoute(state), state);
+      let pendingOperation: { id: number; path: string } | undefined;
+
+      // React may batch multiple queued actions into one state event, so use the latest matching operation.
+      const pendingOperationIndex = pendingHistoryOperationsRef.current.findLastIndex(
```
jakub-agent, line 251: [🏗️ **design**] Matching is by `path` only, and the fallback below unconditionally discards the head of the list. Two ways that goes wrong:

**Interleaved state change.** Popstate queues `{id: 0, path: '/a'}` and dispatches it. Before React commits, something else navigates — a redirect or auth guard, which is common. That state event arrives first with `path === '/b'`, finds no match, and `shift()`s off `{id: 0, '/a'}`. When the `/a` change finally lands there is no pending record, so it is treated as a user navigation and `history.push` writes a new entry for a Back the browser already performed. History grows on Back and the user can't escape the page.

**Duplicate paths.** Rapid back/forward between two entries with the same path leaves two pending records with equal `path`. `findLastIndex` matches the newest and splices off both, so the older one's state event finds nothing and eats a third record.

You already generate a monotonic `id` and carry it through `metadata.history`, but nothing reads it — matching on `id` instead of `path` would remove the ambiguity in both cases. Was path-matching deliberate here?

This is still better than the single `pendingPopStatePathRef` it replaces, so I may be overweighting it — but the redirect case seems reachable in ordinary apps.

packages/expo-router/src/global-state/routingQueue.ts
```diff
@@ -95,12 +109,13 @@ export const routingQueue = {
...
+        intent.onDispatch?.(intent.metadata);
         ref.current.dispatch(dispatchAction);
       } catch (error) {
+        const message =
+          typeof error === 'object' && error != null && 'message' in error ? error.message : error;
         console.warn(
-          `An error occurred when trying to handle a navigation action: ${
-            typeof error === 'object' && error != null && 'message' in error ? error.message : error
-          }`
+          `An error occurred when trying to handle navigation action ${JSON.stringify(intent)}: ${message}`
```
jakub-agent, line 118: [🏗️ **design**] `JSON.stringify` throws on `BigInt` and on circular structures, and route `params` here can come from app code via `dispatch`. If it throws, the exception escapes this `catch` and unwinds out of the `while` loop. `routingQueue.queue` was already reset to `[]` at the top of `run`, so every intent still in `events` is silently lost — one bad param takes down the whole drain.

The previous message only interpolated `error.message`, so it couldn't fail this way.

Serializing the intent is genuinely useful for debugging, so maybe keep it but make it safe — a `try`/`catch` around the stringify, or log `intent.type` plus the href/action type rather than the full object. As written it also embeds an entire navigation state tree in a warning for every `RESET` failure, which is a lot of console output.

packages/expo-router/src/fork/useBrowserHistorySync.ts
```diff
@@ -0,0 +1,307 @@
...
+            payload: { href: path, options: { event: 'NAVIGATE' } },
+          });
+        } else {
+          queueHistoryIntent({
+            type: 'ACTION',
+            payload: { action: reset(state) },
+          });
+        }
+      } else {
+        const initialState = history.get(0)?.state;
```
jakub-agent, line 172: [🏗️ **design**] This changes what "no parseable state" falls back to. The old code was `navigation.resetRoot(state)` with `state === undefined`, which resets to the navigator's *initial* state. Two differences:

- When `history.get(0)?.state` exists, it restores the **first visited entry**, not the initial route. Land on `/deep/link`, navigate away, then hit an unparseable URL: old behaviour went to `/`, new behaviour goes to `/deep/link`.
- When it's absent, the `else` clears the pending list and dispatches **nothing**. The browser URL has already changed, so the address bar and the rendered screen disagree until the next navigation.

The second one looks like the real problem — the old code always dispatched something. If restoring entry 0 is intentional, worth a comment saying so and a `resetRoot`-equivalent fallback when it's missing, since the removed code carried a `// revert to initial state` comment that no longer describes this.

packages/@expo/cli/e2e/playwright/dev/navigator-browser-history.test.ts
```diff
@@ -60,5 +61,64 @@ test.describe(inputDir, () => {
...
+    await expect(page.locator('[data-testid="final-content"]')).toHaveText('/explore/final');
+
+    await page.reload();
+    await expect(page.locator('[data-testid="final-content"]')).toHaveText('/explore/final');
+    await page.goBack();
+    await expect(page.locator('[data-testid="details-content"]')).toHaveText('/explore/details');
+    await page.goForward();
+    await expect(page.locator('[data-testid="final-content"]')).toHaveText('/explore/final');
+
+    expect(pageErrors.all).toEqual([]);
```
jakub-agent, line 103: [💡 suggestion] `pageCollectErrors` pushes `console.warn` into `collected.warnings` but deliberately leaves it out of `collected.all` (`e2e/playwright/page.ts`), so this assertion only covers `console.error` and `pageerror`.

That matters here specifically: `routingQueue.run` reports same-drain collisions with `console.warn` — "Multiple navigation actions in the same queue drain carry a sub-tree for the same navigator" — and its own TODO calls that condition out as not-yet-correct. Chaining two traversals from `popstate` is precisely how you'd land two `RESET`s targeting one navigator in a single drain, so the test that should catch it currently can't fail on it.

Asserting `pageErrors.warnings` too, or filtering it for router warnings, would make these tests prove what they're aiming at.

packages/expo-router/src/fork/useLinking.native.ts
```diff
@@ -170,33 +167,25 @@ export function useLinking(
       }
 
       const navigation = ref.current;
-      const state = navigation ? getStateFromURL(url) : undefined;
+      const path = extractExpoPathFromURL(prefixes, url);
+      const state = navigation && path !== undefined ? getStateFromURL(url) : undefined;
```
jakub-agent, line 171: [🐧 nit] `path !== undefined` is always true — `extractExpoPathFromURL` returns `extractExactPathFromURL(url).replace(...)`, so its inferred type is `string` and it returns `''` rather than `undefined` on failure. The guard reads like prefix validation, but `extractExpoPathFromURL` ignores its `_prefixes` argument entirely, so no prefix check happens.

Related, on line 184: that same `.replace(/^\//, '')` strips the leading slash, so `path.startsWith('/')` is always false and only the `` `/${path}` `` branch ever runs.

If you want the guard, `if (!path) return;` would actually do something — it would skip the `url` that currently resolves to href `/`. Otherwise both checks are dead and worth dropping. (Line 119 has the same always-true comparison, but that one predates this PR.)

packages/expo-router/CHANGELOG.md
```diff
@@ -41,14 +41,14 @@
...
 - Improve standard-navigation types for `createProps`. ([#47825](https://github.com/expo/expo/pull/47825) by [@Ubax](https://github.com/Ubax))
 - Add a `processScreens` option to the `standard-navigation` integration. ([#48290](https://github.com/expo/expo/pull/48290) by [@Ubax](https://github.com/Ubax))
 - Hide the splash screen when the built-in `+not-found` screen renders ([#48721](https://github.com/expo/expo/pull/48721) by [@Ubax](https://github.com/Ubax))
 - Honor loader `Cache-Control` headers via the platform HTTP cache instead of caching loader data in memory ([#48087](https://github.com/expo/expo/pull/48087) by [@hassankhan](https://github.com/hassankhan))
 
 ### 🐛 Bug fixes
 
+- Route deep links and browser history through the navigation queue while preserving saved navigation state.
 - Make layouts with explicitly declared screens honor `unstable_settings.initialRouteName` instead of declaration order, which can change deep-link back stacks. ([#48708](https://github.com/expo/expo/pull/48708) by [@Ubax](https://github.com/Ubax))
```
jakub-agent, line 52: [🐧 nit] Missing the PR link and attribution that every other entry carries:

```
- Route deep links and browser history through the navigation queue while preserving saved navigation state. ([#48889](https://github.com/expo/expo/pull/48889) by [@Ubax](https://github.com/Ubax))
```

jakub-agent (commented):

Routing deep links and browser history through `routingQueue` instead of calling `navigation.dispatch`/`resetRoot` directly is the right move. It puts linking on the same ordering path as the imperative API, and splitting the ~280-line web block out into `useBrowserHistorySync` makes `useLinking.ts` readable for the first time. Replacing the single `pendingPopStatePathRef` with a list of pending operations is also a real improvement over what was there.

Five things I'd want resolved before this lands.

**1. The pending-operation matcher can drop the wrong record.** Operations carry an `id`, but matching is by `path` only, and an unmatched state event blindly `shift()`s the head of the list. Any state change that interleaves with an in-flight popstate — a redirect or auth guard is the common case — consumes another operation's record. The popstate-driven change then looks like a user navigation and writes a fresh history entry, which is the broken-Back behaviour this machinery exists to prevent. The `id` you already thread through `metadata` would make this unambiguous; right now nothing reads it.

**2. `JSON.stringify(intent)` inside `catch` can itself throw**, escaping the handler and aborting the drain loop. The queue is already cleared at that point, so every remaining intent is lost.

**3. A popstate to an unparseable URL can now leave the URL and the UI disagreeing.** The old code called `resetRoot(undefined)`, which always reset to the navigator's initial state. The new `else` branch dispatches nothing when `history.get(0)?.state` is absent.

**4. The new e2e tests can't fail on the warning they should catch.** `pageCollectErrors` keeps `console.warn` out of `.all`, and `routingQueue.run` signals same-drain target collisions via `console.warn`. The rapid-traversal test is exactly that scenario.

**5. Dead plumbing left behind.** Neither `useLinking` variant reads `getActionFromState` any more, but `getLinkingConfig.ts:147` still supplies it and `fork/NavigationContainer.tsx:135` still threads it. `LinkingOptions.getActionFromState` is now accepted and silently ignored — worth removing, or documenting as a no-op.

Also: the PR body is still the unfilled template. Given this changes deep-link and browser-history dispatch semantics, a Test Plan would help reviewers.

I reviewed against base `@ubax/replace-nested-params-with-actual-state`, not `main`, so some of this may already be answered upstream in the stack. I did not run the Playwright suite.

PR #48890, opened 2026-08-13
[router] Migrate href based navigation not to relay on nested params

Ubax:
# Why

# How

# Test Plan

# Checklist


jakub-agent (approved):

The navigation-state migration consistently resolves hrefs against the live router registry, carries only marked internal state, and updates the vendored stack/tab/drawer paths that consume that state. The focused tests cover mounted and unmounted destinations, anchors, preloads, nested state, and queue error handling. I found no actionable correctness, design, security, ABI, or test-coverage issues in the reviewed changes.

packages/expo-router/src/global-state/routingQueue.ts
```diff
@@ -38,34 +39,69 @@ export const routingQueue = {
...
+              `Multiple navigation actions in the same queue drain carry a sub-tree for the same navigator '${target}'.`
+            );
+          }
+          stateTargets.add(target);
         }
+
+        ref.current.dispatch(dispatchAction);
+      } catch (error) {
+        console.warn(
+          `An error occurred when trying to handle a navigation action: ${
```
jakub-agent, line 101: [🚨 **critical**] This `catch` wraps `ref.current.dispatch(...)`, so it swallows every error React Navigation raises during reduction, not just malformed-intent errors, and it is not gated on `NODE_ENV`. Two existing tests demonstrate the effect:

- `src/react-navigation/core/__tests__/removePrevented.test.web.tsx` → `throws a descriptive error when beforeRemove calls preventDefault`
- `src/__tests__/navigation.test.ios.tsx` → `throws when pushing to a layout with an invalid initial route name`

Both assert `toThrow(...)` and now receive `Received function did not throw`; the message ends up as a `console.warn` instead. The second one is a `getValidInitialRoute` throw that now happens inside `resolveNavigationDestination`, so a genuine app misconfiguration becomes a warning that is easy to miss.

Suggest narrowing the `try` to just the intent → action conversion (where a malformed queued item really is recoverable) and letting `dispatch` throw, or re-throwing in dev: `if (process.env.NODE_ENV !== 'production') throw error;`.

packages/expo-router/src/global-state/resolveNavigationDestination.ts
```diff
@@ -0,0 +1,313 @@
...
+  internalParams,
+}: {
+  targetState: PartialState<NavigationState>;
+  navigationState: NavigationState | PartialState<NavigationState> | undefined;
+  routeNode: RouteNode;
+  registry: RouterRegistry;
+  withAnchor: boolean;
+  internalParams: InternalExpoRouterParams;
+}): NavigationState {
+  if (navigationState?.stale !== false || !registry.has(navigationState.key)) {
```
jakub-agent, line 191: [🚨 **critical**] When the registry has no entry for `navigationState.key`, this falls through to `createDestinationState`, which builds a **fresh** state containing only the destination (plus the anchor). Any routes already in that navigator are discarded.

That is reachable on the very first drain. `useRegisterRouter` registers through `useClientLayoutEffect` → `setRegistry`, but `useImperativeApiEmitter` reads the registry during render and calls `routingQueue.run` from a `useEffect`, so the first drain sees the initial empty `Map`. `routerRegistry.tsx` says as much: "Entries appear after the first commit".

Two tests regress on exactly this:

- `src/__tests__/issues.test.ios.tsx` → `can navigate during first render`: after `router.back()` the stack holds only `second`; `index` is gone.
- `multiple pushes in useEffect are executed in order and added to stack`: same shape.

The old nested-`screen`/`params` payload avoided this because React Navigation resolved it against real state at reduce time, after mount. Could the queue hold `NAVIGATE_TO_HREF` intents until the registry has the root navigator, rather than resolving them against an empty registry? The `TODO: Wait for the root navigator to mount instead of dropping the action.` on line 51 points at the same gap.

packages/expo-router/src/global-state/resolveNavigationDestination.ts
```diff
@@ -0,0 +1,313 @@
...
+  if (!targetRoute || !routeNames.includes(targetRoute.name)) {
+    return markState(
+      createInitialState({
+        routeNames,
+        initialRouteName: getValidInitialRouteName(routeNode),
+      })
+    );
+  }
+
+  const destination = createInitialState({
```
jakub-agent, line 234: [🚨 **critical**] `createInitialState` mints `nanoid` keys, so resolving the same href twice produces two states that are never `isEqual`. `resolveState` guards against that on the mounted path (line 214), but this path has no such guard — every resolution yields a structurally identical, referentially and key-wise different sub-tree, which `attachRouteState` then treats as "differs" and assigns.

That turns a re-entrant navigation into a loop. `smoke.test.ios.tsx`/`smoke.test.android.tsx` → `can redirect during the initial render` and `navigation.test.ios.tsx` → `can redirect within a group layout` both now fail with `Maximum update depth exceeded` at `useDescriptors.tsx:214`: state changes → re-render → redirect re-queues → new keys → state changes again.

A structural comparison against the existing child state that ignores keys (or key reuse when the route already exists) before deciding to replace would break the cycle.

packages/expo-router/src/global-state/resolveNavigationDestination.ts
```diff
@@ -0,0 +1,313 @@
...
+  if (!targetRoute) {
+    return action;
+  }
+
+  const params = appendInternalExpoRouterParams(targetRoute.params, internalParams) ?? {};
+  const payload = {
+    ...action.payload,
+    name: targetRoute.name,
+    params,
+    ...(targetRoute.path !== undefined ? { path: targetRoute.path } : undefined),
```
jakub-agent, line 146: [🚨 **critical**] `targetRoute.path` comes straight from `getStateFromPath`, which includes the query string and the hash. The old payload built by `getPayloadFromStateRoute` never carried `path` at all, so `route.path` was only ever set by deep-link handling.

`src/__tests__/hashs.test.ios.tsx` → `navigating to the same route with a hash will only rerender the screen` now gets `path: '/?#hash1'` where it expects `path: '/'`. `route.path` feeds `getPathFromState` and route-info derivation, so this is worth pinning down rather than absorbing into the expectation.

Same propagation happens at line 245 in `createDestinationState`. Stripping search and hash before assigning `path` (or keeping `path` off the payload entirely, as before) would restore the old semantics.

packages/expo-router/src/global-state/resolveNavigationDestination.ts
```diff
@@ -0,0 +1,313 @@
...
+    | undefined
+) {
+  if (result?.state.stale !== false || result.affectedRouteKey === undefined) {
+    return undefined;
+  }
+  return result.state.routes.find((route) => route.key === result.affectedRouteKey);
+}
+
+function markState<T extends NavigationState>(state: T): T & RouteState {
+  return { ...state, __internal__routerActionState: true };
```
jakub-agent, line 309: [🏗️ **design**] The marker is never removed, so it lands in the live navigation state. The updated `prefetch.test.ios.tsx` expectations confirm it (`__internal__routerActionState: true` inside `state`), and it shows up in the diffs of the newly failing `push`/`dismissAll` assertions too. That means it is visible through `navigation.getState()`, `useNavigationState`, `onStateChange`, and state persistence.

It also forces the cost on the read side: `attachRouteState` has to deep-clone both trees via `stripStateMarkers` just to compare them, precisely because stored state may or may not carry markers.

Stripping the marker in `attachRouteState` when the state is accepted would keep public state clean and let the equality check compare directly.

packages/expo-router/src/global-state/resolveNavigationDestination.ts
```diff
@@ -0,0 +1,313 @@
...
+      !registry.has(childState.key)
+    ) {
+      break;
+    }
+
+    const childRoute = getFocusedRoute(childTarget);
+    if (!childRoute) {
+      break;
+    }
+    const probeAction: NavigationAction = {
```
jakub-agent, line 92: [🏗️ **design**] This calls a registered router's `reduce` purely to test whether it would accept the action, and throws the result away. Two concerns:

1. Routers are pure with respect to state, but not with respect to logging. `NativeBottomTabsRouter` warns about unsupported zoom transitions inside `getStateForAction`, and `attachRouteState` warns on unmarked state. A probe can emit a warning for a navigation that is then resolved differently, or emit the same warning twice.
2. Counting the probe, `createResolvedAction` (line 149) and `resolveState` (line 209), each navigator level is reduced about three times per navigation. Linear, so not alarming, but it is three times the work of a single dispatch and each `stackRouterOverride` reduction burns `nanoid` calls.

If the intent is only "does this navigator know this route name", `state.routeNames.includes(childRoute.name)` would answer it without invoking the reducer.

packages/expo-router/src/global-state/routingQueue.ts
```diff
@@ -38,34 +39,69 @@ export const routingQueue = {
...
-          ref.current.dispatch(intent.payload.action);
+          dispatchAction = intent.payload.action;
+        }
+
+        const target = dispatchAction.target;
+        if (target && dispatchAction.payload && 'state' in dispatchAction.payload) {
+          if (stateTargets.has(target)) {
+            // TODO: Remove this warning once queued actions are reduced sequentially against global state.
+            console.warn(
+              `Multiple navigation actions in the same queue drain carry a sub-tree for the same navigator '${target}'.`
```
jakub-agent, line 92: [🏗️ **design**] This catches the case where two actions in one drain carry a sub-tree for the *same* navigator key, but the underlying problem is broader: every action in a drain is resolved against the root state as it was *before* the drain started. Two actions targeting a parent and its child, or two sibling navigators, are equally stale and produce no warning.

Worth either widening the condition (warn whenever a second state-bearing action is resolved in the same drain) or saying in the message that it only covers the same-key case, so it is not read as complete coverage.

packages/expo-router/src/global-state/routingQueue.ts
```diff
@@ -38,34 +39,69 @@ export const routingQueue = {
       callback();
     }
   },
-  run(ref: RefObject<NavigationContainerRef<ParamListBase> | null>) {
+  run(
+    ref: RefObject<NavigationContainerRef<ParamListBase> | null>,
+    registry: RouterRegistry = new Map()
```
jakub-agent, line 44: [💡 suggestion] `imperative-api.tsx` is the only production caller and it always passes a registry, so this default only serves the tests. It also fails quietly: an empty registry makes every `registry.has` miss, which silently degrades resolution to full state replacement (see the comment on `resolveNavigationDestination.ts:191`) rather than surfacing a wiring bug.

Making the parameter required and passing `new Map()` explicitly in the tests would keep the failure mode visible.

packages/expo-router/src/global-state/routingQueue.ts
```diff
@@ -38,34 +39,69 @@ export const routingQueue = {
...
     routingQueue.queue = [];
+    const stateTargets = new Set<string>();
     let intent: RoutingIntent | undefined;
     while ((intent = events.shift())) {
-      // TODO: Consider warning when ref.current is null — actions are silently dropped
-      if (ref.current) {
+      if (!ref.current) {
+        // TODO: Wait for the root navigator to mount instead of dropping the action.
+        console.warn(
+          'Navigation action was dropped because the navigation container is not mounted.'
```
jakub-agent, line 55: [🐧 nit] This warning and the one on line 92 are unconditional, while `attachRouteState` gates its warning on `process.env.NODE_ENV !== 'production'`. If these are meant as developer diagnostics for known gaps (both carry TODOs), gating them the same way would keep production logs quiet and stay consistent with the rest of the package.

packages/expo-router/src/global-state/getNavigationAction.ts
```diff
@@ -28,91 +33,35 @@ export function getNavigateAction(
...
-  const rootState = navigationRef.getRootState();
-
-  href = resolveHrefStringWithSegments(href, store.getRouteInfo(), options);
-  href = applyRedirects(href, store.redirects) ?? undefined;
 
-  // If the href is undefined, it means that the redirect has already been handled by the navigation
-  if (!href) {
-    return;
-  }
+  const href = applyRedirects(
```
jakub-agent, line 40: [💡 suggestion] `applyRedirects` is typed `string | undefined | null`, and the `!` drops the check the old code had (`if (!href) return;`, commented "the redirect has already been handled by the navigation"). Today every code path returns a string for a string input, so this is safe — but the TODO this PR adds at `getRoutesRedirects.tsx:33` contemplates exactly the change that would break it ("whether redirect resolution should open external URLs or only return them"). If external redirects ever stop returning an href, `getStateFromPath(undefined)` is the failure.

Cheaper to keep an explicit guard and return `{ status: 'invalid', href: baseHref }` than to rely on the assertion.

packages/expo-router/src/global-state/stateUtils.ts
```diff
@@ -49,8 +49,9 @@ export function getPayloadFromStateRoute(_actionStateRoute: PartialRoute<any>) {
 export function findDivergentState(
   _actionState: ResultState,
   _navigationState: NavigationState,
-  // If true, look through all tabs to find the target state, rather then just the current tab
-  lookThroughAllTabs: boolean = false
+  // TODO: Convert these positional options to an object if more are added.
+  lookThroughAllTabs = false,
+  isMounted: (key: string) => boolean = () => true
```
jakub-agent, line 54: [💡 suggestion] No production caller passes `isMounted` — `link/preview/utils.ts` is the only remaining consumer of `findDivergentState` and it uses the default. Only the new test in `stateUtils.test.ios.ts` exercises it. Worth dropping until the caller that needs it lands, so the parameter isn't dead API surface with test-only coverage.

The `isKeyedNavigationState` guard added alongside it is a real fix, but note it also changes behavior for the existing `link/preview/utils.ts` callers: the walk now stops at a keyless child state where it previously descended and cast. That is a separate change from this PR's stated goal — worth calling out in the description or splitting out.

packages/expo-router/src/react-navigation/routers/attachRouteState.tsx
```diff
@@ -0,0 +1,51 @@
...
+      // TODO: Remove this warning once unmarked state-bearing actions are unsupported.
+      console.warn(
+        'A navigation action carried state without the internal __internal__routerActionState marker. The state was ignored because only Expo Router can safely generate navigation state. Use Expo Router navigation APIs instead of dispatching state-bearing actions directly.'
+      );
+    }
+    return route;
+  }
+
+  if (route.state !== undefined) {
+    if (isEqual(stripStateMarkers(route.state), stripStateMarkers(state))) {
```
jakub-agent, line 33: [💡 suggestion] `stripStateMarkers` deep-clones both trees before comparing, so every state-bearing action allocates two full copies of the sub-tree just to run an equality check — and this runs inside `getStateForAction`, on every navigation, at every level.

A marker-ignoring recursive comparison would avoid the clones entirely. If the marker were stripped on write (see the comment on `resolveNavigationDestination.ts:309`), plain `isEqual(route.state, state)` would do.

packages/expo-router/src/global-state/router.ts
```diff
@@ -139,6 +140,7 @@ export function linkTo(originalHref: Href | string, options: LinkToOptions = {})
     return goBack();
   }
 
+  // TODO(@ubax): Extract this change to standalone PR
```
jakub-agent, line 143: [🐧 nit] Leftover note to self — should be resolved (split out, or the comment dropped) before merge.

packages/expo-router/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### 🛠 Breaking changes
 
+- Stop copying ancestor route params into descendant routes during imperative navigation, matching cold-start deep-link behavior.
```
jakub-agent, line 9: [🐧 nit] Missing the PR link and author attribution that the surrounding entries carry, e.g. `([#48890](https://github.com/expo/expo/pull/48890) by [@Ubax](https://github.com/Ubax))`.

jakub-agent (changes requested):

**PR 3/11 in the router stack** (`48888 → 48886 → **48890** → 48889 → 48895 → 48917 → 49034 → 49063 → 48934 → 49073`).

The direction is right. Replacing React Navigation's nested `screen`/`params`/`initial` payload encoding with a real resolved sub-tree removes a long-standing source of param leakage, and it makes imperative navigation match cold-start deep-link behavior. `resolveNavigationDestination` is a clear place to put that logic, and the new unit tests for it are well chosen.

**The suite is red, though.** I ran `CI=1 yarn jest` in `packages/expo-router` at `d63df5e` and at the base branch `@ubax/add-routing-intent-to-the-queue` (`fdf1be8`), with the same environment for both:

| | base | this PR |
|---|---|---|
| Test suites | 3 failed / 375 | 17 failed / 380 |
| Tests | 3 failed / 4877 | 32 failed / 4951 |

That is **27 distinct newly failing tests**. Only `renders Icon src` and one `Console` assertion fail on the base branch, so the rest are attributable to this diff. They split into two groups.

**Group 1 — assertions that just need updating** (the new state shape looks correct; the tests still expect the old nested-params shape): `push should also add anchor routes`, `push should include/ignore (group)/index …`, `dismissAll nested`, `works in a nested layout Stack->Tab->Stack`, `should go back to a previous route in the same/different stacks`, `stacks should always push a new route`, `navigate › can handle navigation between routes`, `navigate › handles dismissAll`, `singular › can dynamically route using singular(  function)`, `will replace the route if the provided href is not in the history`, `targets the correct Stack when pushing to a nested layout`, `Preview › multiple preloaded paths with the same name …`.

**Group 2 — behavioral regressions**, each flagged inline below:

1. `throws a descriptive error when beforeRemove calls preventDefault` and `throws when pushing to a layout with an invalid initial route name` no longer throw. The new blanket `try`/`catch` in `routingQueue.run` swallows them.
2. `can navigate during first render` and `multiple pushes in useEffect are executed in order and added to stack` lose the initial route. Actions drained before the router registry is populated resolve against an empty registry.
3. `can redirect during the initial render` and `can redirect within a group layout` now hit `Maximum update depth exceeded`.
4. `navigating to the same route with a hash …` gets `path: '/?#hash1'` instead of `path: '/'`. `route.path` now carries the query string and hash.
5. `should always prefer static routes over dynamic ones` ends with empty segments — the push does not land at all.
6. Four `First focused tab` tests now see 2 `NativeTabsView` renders instead of 1.

Smaller things: the PR body is the unfilled template, so there is no stated motivation or test plan; the changelog entry has no PR link or author attribution unlike its neighbours; and `getRoutesRedirects.tsx`, `store.ts`, and `router.ts` change only to add TODO comments, one of which says `TODO(@ubax): Extract this change to standalone PR`.

Happy to re-review once the suite is green.

packages/expo-router/src/global-state/routingQueue.ts
Ubax: Narrow the try

packages/expo-router/src/global-state/resolveNavigationDestination.ts
Ubax: Will this happen only during very first render or also after a render of any new navigator?

packages/expo-router/src/global-state/resolveNavigationDestination.ts
Ubax: Remove it when appending the state

packages/expo-router/src/global-state/routingQueue.ts
Ubax: Do it

packages/expo-router/src/global-state/routingQueue.ts
Ubax: Keep it this way for now. I'll solve this in the TODO

packages/expo-router/src/global-state/routingQueue.ts
Ubax: Use the `!==` gate

packages/expo-router/src/global-state/stateUtils.ts
Ubax: Do it

packages/expo-router/src/react-navigation/routers/attachRouteState.tsx
Ubax: We'll strip them, so the check my not be necessary

packages/expo-router/src/global-state/router.ts
Ubax: Remove the note

packages/expo-router/src/global-state/resolveNavigationDestination.ts
Ubax: Strip to align. Add TODO to check later

packages/expo-router/src/global-state/resolveNavigationDestination.ts
Ubax: Try to make them predictable

PR #48891, opened 2026-08-13
[app-metrics][observe] Add a `networkTraces` option to configure, gating network trace recording at capture time

tsapeta:
# Why

Network spans record full URLs, and until now the SDK offered no way to turn recording off or narrow it. The review of expo/expo#48861 flagged that dispatch-side gates aren't enough: rows shouldn't reach disk at all when the app opts out.

Closes ENG-26033.

# How

`Observe.configure` gains a `networkTraces` option: a boolean or `{ filter }` object. Gating and filtering happen at capture time in the producer, so a request that doesn't match is never written to the database. The setting persists (UserDefaults / SharedPreferences), so early-startup requests on the next launch follow the last-applied value. It applies forward only: rows persisted before the change keep dispatching, as decided during the design review.

The policy lives in expo-app-metrics next to the producer. `configure` normalizes the sugar and pushes it down via an internal `setNetworkTracesConfig` module function that persists the value and updates the live producer. Matching delegates to `NetworkRequestFilter`, the same type the JS-facing observer uses, so the capture gate can't drift from it. Both platforms resolve the persisted policy before the producer exists (into its constructor), and apply later changes on the queue that owns the producer, so a `configure` racing startup can't be dropped.

```ts
import { Observe } from 'expo-observe';

// Turn network span recording off entirely.
Observe.configure({ networkTraces: false });

// Or record only the hosts and methods you care about.
Observe.configure({
  networkTraces: { filter: { hosts: ['api.myapp.com'], methods: ['GET', 'POST'] } },
});
```

The option sits flat on the config root, next to `dispatchingEnabled` and `errorHandlingEnabled`, rather than under a `traces` namespace: it is the only span-recording switch today, and the surrounding config uses compound names instead of nesting. The producer-side type is `NetworkTracesConfiguration` so the name matches from the public option down to the persisted key.

# Test Plan

New native suites on both platforms cover the matching matrix (case-insensitivity, empty-allowlist, hostless URLs), the persisted encoding round-trip, malformed-blob fallback, and persistence-level gating (disabled drops, filter drops, live reconfigure applies forward). Jest covers the `configure` normalization, including the default when `networkTraces` is omitted.

- `et native-unit-tests -p ios --packages expo-observe,expo-app-metrics`: 477 passed, 0 failed
- `et native-unit-tests -p android`: both packages BUILD SUCCESSFUL
- Jest: 68 suites / 576 tests passed
- `tsc` clean for both packages

### DIFF
--- a/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/AppMetricsModule.kt
+++ b/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/AppMetricsModule.kt
@@ -16,6 +16,7 @@ import expo.modules.appmetrics.networkrequests.NetworkRequestFilter
 import expo.modules.appmetrics.networkrequests.NetworkRequestMonitor
 import expo.modules.appmetrics.networkrequests.NetworkRequestObserver
 import expo.modules.appmetrics.networkrequests.NetworkRequestPersistence
+import expo.modules.appmetrics.networkrequests.NetworkSpansConfiguration
 import expo.modules.appmetrics.logevents.Severity
 import expo.modules.appmetrics.logevents.sanitizeLogEventAttributes
 import expo.modules.appmetrics.logevents.validateDisplayName
@@ -69,8 +70,9 @@ class AppMetricsModule : Module(), UpdatesStateChangeListener {
   lateinit var mainSession: SessionSharedObject
 
   /**
-   * The span producer installed on the monitor in `OnCreate`, kept so `OnDestroy` can
-   * uninstall it when the module is torn down (a JS reload).
+   * The span producer installed on the monitor in `OnCreate`, kept so `setNetworkSpansConfig`
+   * can apply a new recording policy to the live instance and `OnDestroy` can uninstall it
+   * when the module is torn down (a JS reload).
    */
   private var networkRequestPersistence: NetworkRequestPersistence? = null
 
@@ -136,6 +138,19 @@ class AppMetricsModule : Module(), UpdatesStateChangeListener {
         GlobalAttributes.set(attributes)
       }
 
+      Function("setNetworkSpansConfig") { config: NetworkSpansConfigParam ->
+        val configuration = NetworkSpansConfiguration(
+          enabled = config.enabled,
+          hosts = config.filter?.hosts,
+          methods = config.filter?.methods
+        )
+        // Persist first so the setting survives the process; the live producer picks it up for
+        // every request recorded after the volatile write. Applies forward only — rows
+        // persisted earlier in the launch still dispatch.
+        AppMetricsPreferences.setNetworkSpansConfiguration(context, configuration)
+        networkRequestPersistence?.setConfiguration(configuration)
+      }
+
       OnCreate {
         sessionManager = SessionManager(context)
 
@@ -172,6 +187,12 @@ class AppMetricsModule : Module(), UpdatesStateChangeListener {
             sessionId = mainSession.sessionId
           )
           networkRequestPersistence = persistence
+          // The policy is read from preferences only after the reference above is visible:
+          // a `setNetworkSpansConfig` call landing before this line has already persisted its
+          // value and is picked up here; one landing after it hits the live setter instead.
+          // Reading before the assignment would let a concurrent configure fall between the
+          // read and the assignment and leave a stale policy until the next launch.
+          persistence.setConfiguration(AppMetricsPreferences.getNetworkSpansConfiguration(context))
           NetworkRequestMonitor.shared.installPersistence(persistence)
         }
 
@@ -390,4 +411,14 @@ class AppMetricsModule : Module(), UpdatesStateChangeListener {
 data class MetricAttributes(
   @Field val routeName: String? = null,
   @Field val params: Map<String, Any>? = null
+) : Record
+
+/**
+ * Payload of `setNetworkSpansConfig`: the normalized `traces.network` setting pushed down by
+ * `Observe.configure`.
+ */
+@OptimizedRecord
+data class NetworkSpansConfigParam(
+  @Field val enabled: Boolean = true,
+  @Field val filter: NetworkRequestFilter? = null
 ) : Record
--- a/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/AppMetricsPreferences.kt
+++ b/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/AppMetricsPreferences.kt
@@ -2,10 +2,12 @@ package expo.modules.appmetrics
 
 import android.content.Context
 import androidx.core.content.edit
+import expo.modules.appmetrics.networkrequests.NetworkSpansConfiguration
 
 private const val PREFS_NAME = "dev.expo.app-metrics"
 private const val KEY_ENVIRONMENT = "environment"
 private const val KEY_LAST_PROCESSED_EXIT_MILLIS = "lastProcessedExitMillis"
+private const val KEY_NETWORK_SPANS_CONFIGURATION = "networkSpansConfiguration"
 
 object AppMetricsPreferences {
   /**
@@ -40,5 +42,25 @@ object AppMetricsPreferences {
 
   fun getDefaultEnvironment(): String? {
     return if (BuildConfig.DEBUG) "development" else null
+  }
+
+  /**
+   * Last-applied network spans recording policy, or the default when JS never configured one.
+   * Persisted so requests observed before the JS bundle configures the SDK — early startup, or
+   * the next launch — follow the last-known setting.
+   */
+  fun getNetworkSpansConfiguration(context: Context): NetworkSpansConfiguration {
+    val prefs = context.getSharedPreferences(PREFS_NAME, Context.MODE_PRIVATE)
+    val json = prefs.getString(KEY_NETWORK_SPANS_CONFIGURATION, null)
+      ?: return NetworkSpansConfiguration()
+    return NetworkSpansConfiguration.fromJson(json)
+  }
+
+  fun setNetworkSpansConfiguration(
+    context: Context,
+    configuration: NetworkSpansConfiguration
+  ) {
+    val prefs = context.getSharedPreferences(PREFS_NAME, Context.MODE_PRIVATE)
+    prefs.edit(commit = true) { putString(KEY_NETWORK_SPANS_CONFIGURATION, configuration.toJson()) }
   }
 }
--- a/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/networkrequests/NetworkRequestPersistence.kt
+++ b/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/networkrequests/NetworkRequestPersistence.kt
@@ -29,6 +29,7 @@ private const val TAG = "ExpoAppMetrics"
 class NetworkRequestPersistence(
   private val database: MetricsDatabase,
   private val scope: CoroutineScope,
+  initialConfiguration: NetworkSpansConfiguration = NetworkSpansConfiguration(),
   /**
    * A plain value, not a provider: the id is constant for a persistence instance (it is
    * constructed after the main session exists), and resolving it eagerly means the monitor's
@@ -38,11 +39,30 @@ class NetworkRequestPersistence(
   private val sessionId: String
 ) {
   /**
+   * Capture-time recording policy. Volatile because the monitor calls in from OkHttp dispatcher
+   * threads while JS reconfigures from the modules queue. Checked before each insert so a change
+   * takes effect for future requests immediately; rows persisted earlier are untouched.
+   */
+  @Volatile
+  private var configuration: NetworkSpansConfiguration = initialConfiguration
+
+  /**
+   * Applies a new recording policy to subsequent requests. Called when JS reconfigures
+   * `traces.network`; the caller persists the value separately.
+   */
+  fun setConfiguration(configuration: NetworkSpansConfiguration) {
+    this.configuration = configuration
+  }
+
+  /**
    * Persists one completed request as a span. The insert runs on `scope` so the monitor's
    * record path (OkHttp dispatcher threads) never blocks on the database; failures are logged
    * and swallowed — persistence must never break the monitor's fan-out.
    */
   fun persist(request: NetworkRequest) {
+    if (!configuration.allows(request.url, request.method)) {
+      return
+    }
     val span = request.toSpan(sessionId) ?: return
     scope.launch {
       try {
@@ -66,6 +86,9 @@ class NetworkRequestPersistence(
     }
     scope.launch {
       for (request in requests) {
+        if (!configuration.allows(request.url, request.method)) {
+          continue
+        }
         val span = request.toSpan(sessionId) ?: continue
         try {
           database.spanDao().insertCapped(span)
--- a/packages/expo-app-metrics/android/src/test/java/expo/modules/appmetrics/networkrequests/NetworkRequestPersistenceTest.kt
+++ b/packages/expo-app-metrics/android/src/test/java/expo/modules/appmetrics/networkrequests/NetworkRequestPersistenceTest.kt
@@ -326,6 +326,55 @@ class NetworkRequestPersistenceTest {
   }
 
   @Test
+  fun `drops every request while recording is disabled`() = runTest(testDispatcher) {
+    insertSession("s")
+    val persistence = NetworkRequestPersistence(
+      database = database,
+      scope = this,
+      initialConfiguration = NetworkSpansConfiguration(enabled = false),
+      sessionId = "s"
+    )
+    persistence.persist(makeRequest())
+    testScheduler.advanceUntilIdle()
+    assertTrue(database.spanDao().getAll().isEmpty())
+  }
+
+  @Test
+  fun `records only requests matching the configured filter`() = runTest(testDispatcher) {
+    insertSession("s")
+    val persistence = NetworkRequestPersistence(
+      database = database,
+      scope = this,
+      initialConfiguration = NetworkSpansConfiguration(enabled = true, hosts = listOf("API.myapp.com")),
+      sessionId = "s"
+    )
+    persistence.persist(makeRequest(url = "https://api.example.com/skip"))
+    persistence.persist(makeRequest(url = "https://api.myapp.com/keep"))
+    testScheduler.advanceUntilIdle()
+    val rows = database.spanDao().getAll()
+    assertEquals(1, rows.size)
+    val recordedUrl = JSONObject(checkNotNull(rows.single().attributes)).getString("url.full")
+    assertEquals("https://api.myapp.com/keep", recordedUrl)
+  }
+
+  @Test
+  fun `applies a configuration change to subsequent requests only`() = runTest(testDispatcher) {
+    // "Applies forward": rows persisted before the change stay in the table and still dispatch.
+    insertSession("s")
+    val persistence = NetworkRequestPersistence(
+      database = database,
+      scope = this,
+      sessionId = "s"
+    )
+    persistence.persist(makeRequest())
+    testScheduler.advanceUntilIdle()
+    persistence.setConfiguration(NetworkSpansConfiguration(enabled = false))
+    persistence.persist(makeRequest())
+    testScheduler.advanceUntilIdle()
+    assertEquals(1, database.spanDao().getAll().size)
+  }
+
+  @Test
   fun `persists a completed request as a span attributed to the provided session`() = runTest(testDispatcher) {
     insertSession("main-session")
     val persistence = NetworkRequestPersistence(
--- a/packages/expo-app-metrics/ios/AppMetricsModule.swift
+++ b/packages/expo-app-metrics/ios/AppMetricsModule.swift
@@ -64,6 +64,25 @@ public final class AppMetricsModule: Module, UpdatesStateChangeListener {
       GlobalAttributes.set(attributes)
     }
 
+    Function("setNetworkSpansConfig") { (config: NetworkSpansConfigParam) in
+      let configuration = NetworkSpansConfiguration(
+        enabled: config.enabled,
+        hosts: config.filter?.hosts,
+        methods: config.filter?.methods
+      )
+      // Persist first so the setting survives the process; the live producer picks it up for
+      // every request recorded after this hop to the actor. Applies forward only — rows
+      // persisted earlier in the launch still dispatch.
+      AppMetricsUserDefaults.networkSpansConfiguration = configuration
+      Task { @AppMetricsActor in
+        // Re-read instead of capturing `configuration`: two rapid `configure()` calls create
+        // two unordered actor hops, while the persisted value is written in call order —
+        // reading it here makes the last write win regardless of task interleaving.
+        let latest = AppMetricsUserDefaults.networkSpansConfiguration ?? NetworkSpansConfiguration()
+        NetworkRequestMonitor.shared.persistence?.setConfiguration(latest)
+      }
+    }
+
     AsyncFunction("getAppStartupTimesAsync") {
       return await AppMetrics.mainSession.appStartupMonitor.metrics
     }
@@ -185,6 +204,16 @@ private func storedSession(id: String) throws -> StoredSession? {
     return nil
   }
   return StoredSession(from: row)
+}
+
+/// Payload of `setNetworkSpansConfig`: the normalized `traces.network` setting pushed down by
+/// `Observe.configure`.
+internal struct NetworkSpansConfigParam: Record {
+  @Field
+  var enabled: Bool = true
+
+  @Field
+  var filter: NetworkRequestFilter?
 }
 
 struct MetricAttributes: Record {
--- a/packages/expo-observe/src/module.ts
+++ b/packages/expo-observe/src/module.ts
@@ -8,6 +8,19 @@ import { isReactNavigationInstalled } from './integrations/react-navigation/reac
 import { reportCaughtError } from './reportCaughtError';
 import type { ObserveConfig, ObserveIntegrationsConfig, ObserveModule } from './types';
 
+/**
+ * Normalizes the `traces.network` sugar (`true` / `false` / `{ filter }`) into the config shape
+ * the native producer persists. `configure` is a full replacement, so an absent `traces` resets
+ * to the default: enabled, no filter.
+ */
+function networkSpansConfigFromTraces(traces: ObserveConfig['traces']) {
+  const network = traces?.network ?? true;
+  if (typeof network === 'boolean') {
+    return { enabled: network };
+  }
+  return network.filter != null ? { enabled: true, filter: network.filter } : { enabled: true };
+}
+
 const native = requireNativeModule<ObserveModule>('ExpoObserve');
 
 const Observe: ObserveModule = new Proxy(native, {
@@ -17,6 +30,11 @@ const Observe: ObserveModule = new Proxy(native, {
         // The handler is already installed at this point (it installs on import), so this only
         // toggles whether it records anything.
         setErrorHandlerEnabled(config.errorHandlingEnabled ?? true);
+
+        // Recording is gated in expo-app-metrics (the producer side), so the setting travels
+        // there rather than into the native `configure` payload. Applies to future captures
+        // only; spans persisted earlier in the launch still dispatch.
+        AppMetrics.setNetworkSpansConfig(networkSpansConfigFromTraces(config.traces));
 
         const routerEnabled = !!config.integrations?.['expo-router'];
         const reactNavigationEnabled = !!config.integrations?.['react-navigation'];
--- a/packages/expo-observe/src/types.ts
+++ b/packages/expo-observe/src/types.ts
@@ -1,5 +1,10 @@
 import type { NativeModule } from 'expo';
-import type { LogAttributeValue, LogEventOptions, MetricAttributes } from 'expo-app-metrics';
+import type {
+  LogAttributeValue,
+  LogEventOptions,
+  MetricAttributes,
+  NetworkRequestFilter,
+} from 'expo-app-metrics';
 
 /**
  * Value types accepted as attribute values in `setGlobalAttributes` and the
@@ -81,6 +86,38 @@ export type ObserveConfig = {
    * [integrate your own package](/eas/observe/integrations/third-party/).
    */
   integrations?: ObserveIntegrationsConfig;
+  /**
+   * Controls which trace spans are recorded on the device.
+   */
+  traces?: ObserveTracesConfig;
+};
+
+export type ObserveTracesNetworkConfig = {
+  /**
+   * Only network requests matching the filter are recorded as spans. Requests that don't match
+   * are never written to the local database. An omitted field matches every request; an empty
+   * array matches none.
+   */
+  filter?: NetworkRequestFilter | null;
+};
+
+export type ObserveTracesConfig = {
+  /**
+   * Whether observed network requests are recorded as trace spans and exported to EAS Observe.
+   *
+   * The setting controls recording, not just export: when disabled (or when a request doesn't
+   * match the configured filter), the request is never written to the local database. Spans
+   * recorded before `configure` ran - including requests from early startup - are unaffected
+   * and still dispatch.
+   *
+   * The value persists across launches, so requests observed before `configure` runs on the
+   * next launch follow the last-applied setting.
+   *
+   * Pass an object to record only requests matching a filter.
+   *
+   * @default true
+   */
+  network?: boolean | ObserveTracesNetworkConfig;
 };
 
 export type ObserveNavigationIntegrationConfig = {
--- a/packages/expo-app-metrics/CHANGELOG.md
+++ b/packages/expo-app-metrics/CHANGELOG.md
@@ -19,6 +19,7 @@
 
 ### 💡 Others
 
+- Add an internal `setNetworkSpansConfig` to gate and filter network span recording at capture time. ([#48891](https://github.com/expo/expo/pull/48891) by [@tsapeta](https://github.com/tsapeta))
 - Rename the no-update `downloadComplete` state event to `downloadCompleteUnavailable`. ([#47902](https://github.com/expo/expo/pull/47902) by [@kudo](https://github.com/kudo))
 - [iOS] Measure the JS bundle load time against the app startup end marker to stay compatible with upcoming React Native versions. ([#47782](https://github.com/expo/expo/pull/47782) by [@tsapeta](https://github.com/tsapeta))
 - Add a `caught` source to the private `reportError` for errors reported from user code. ([#47871](https://github.com/expo/expo/pull/47871) by [@tsapeta](https://github.com/tsapeta))
--- a/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/networkrequests/NetworkSpansConfiguration.kt
+++ b/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/networkrequests/NetworkSpansConfiguration.kt
@@ -1,0 +1,80 @@
+// Copyright 2025-present 650 Industries. All rights reserved.
+
+package expo.modules.appmetrics.networkrequests
+
+import okhttp3.HttpUrl.Companion.toHttpUrlOrNull
+import org.json.JSONArray
+import org.json.JSONObject
+
+/**
+ * Capture-time recording policy for network request spans: whether completed requests are
+ * written to the `spans` table at all, plus an optional host/method allowlist. Consulted by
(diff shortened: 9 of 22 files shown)


packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/AppMetricsModule.kt
```diff
@@ -170,6 +185,12 @@ class AppMetricsModule : Module(), UpdatesStateChangeListener {
             sessionId = mainSession.sessionId
           )
           networkRequestPersistence = persistence
+          // The policy is read from preferences only after the reference above is visible:
+          // a `setNetworkTracesConfig` call landing before this line has already persisted its
+          // value and is picked up here; one landing after it hits the live setter instead.
+          // Reading before the assignment would let a concurrent configure fall between the
+          // read and the assignment and leave a stale policy until the next launch.
+          persistence.setConfiguration(AppMetricsPreferences.getNetworkTracesConfiguration(context))
```
tsapeta: [🚨 **critical**] This comment claims an ordering guarantee the code doesn't provide, and that's the part worth fixing — a future reader will trust it and stop looking.

`networkRequestPersistence` (line 77) is a plain `var`, not `@Volatile`. This block runs on `scope` = `appContext.modulesQueue`, a dedicated `HandlerThread` (`AppContext.kt:70`), while the `setNetworkTracesConfig` body is a sync `Function`, so it runs on the JS thread. There is no happens-before edge between the two, so:

- A `setNetworkTracesConfig` on the JS thread can read `networkRequestPersistence` as `null` even after this coroutine assigned it, and skip the live setter. It did persist first, so it isn't lost on the *next* launch, but the live producer keeps the old policy for the rest of this launch — exactly the "stale policy until the next launch" outcome the comment says it prevents.
- Symmetrically, the prefs write on the JS thread and the prefs read on line 193 are only ordered by SharedPreferences' own locking. The interleaving `[JS: write prefs] [modulesQueue: read prefs] [JS: read null producer]` is possible but so is `[modulesQueue: read prefs] [JS: write prefs] [JS: read null producer]`, which drops the update for this launch.

Two options: mark the field `@Volatile` (makes the read-after-assign argument actually hold for the null check, though the prefs read/write pair is still unordered), or push both the assignment and the apply through a single serialization point — e.g. do the `setConfiguration` hop on `modulesQueue` from the `Function` body too, so both sides run on the same queue and last-write-wins holds for real.

The iOS side doesn't have this problem: `NetworkRequestMonitor` is `@AppMetricsActor`-isolated, so the `AppMetricsActor.isolated` re-read in `AppMetricsModule.swift` genuinely serializes against the install. Worth making Android match that structure rather than relying on the comment.

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/networkrequests/NetworkRequestPersistence.kt
```diff
@@ -61,6 +80,9 @@ class NetworkRequestPersistence(
     // there would be the most expensive place to do it.
     scope.launch {
       for (request in requests) {
+        if (!configuration.allows(request.url, request.method)) {
```
tsapeta: [🏗️ **design**] This gate reads `configuration` at *drain* time, not at capture time, which quietly inverts the "applies forward only" rule for the startup buffer.

`persistBuffered` receives requests the monitor buffered since process start (`NetworkRequestMonitor.installPersistence`). Those were observed before any producer existed, then drained here — and filtered by whichever configuration is live when the coroutine body runs. So a request captured at T0 under the persisted "record everything" policy gets dropped if JS calls `configure({ networkTraces: false })` before the drain completes. The `persist` path has the opposite behavior: policy is sampled the moment the request arrives.

That may well be the behavior you want (the app said "stop recording", so don't write the backlog), but it's the reverse of what the class docs and the PR description state, and the loop re-reads the volatile field on every iteration, so a mid-drain reconfigure can even split one batch. Whichever semantics you pick, it's worth a comment here and a test — `applies a configuration change to subsequent requests only` only covers `persist`.

packages/expo-app-metrics/ios/NetworkRequests/NetworkTracesConfiguration.swift
```diff
@@ -0,0 +1,44 @@
...
+
+  /// Allowed hosts, compared for exact, case-insensitive equality. `nil` allows every host;
+  /// an empty array allows none.
+  var hosts: [String]?
+
+  /// Allowed HTTP methods, compared case-insensitively. `nil` allows every method.
+  var methods: [String]?
+
+  /// Whether a request should be recorded. Same matching rules as `NetworkRequestFilter`.
+  func allows(url: URL, method: String) -> Bool {
```
tsapeta: [🏗️ **design**] `allows` is a verbatim reimplementation of `NetworkRequestFilter.matches`, not a reuse of it. The host branch, the method branch, the `lowercased()`/`uppercased()` comparisons, and the empty-array-blocks-everything behavior are all duplicated — and the same duplication exists on Android between `NetworkTracesConfiguration.allows` and `NetworkRequestFilter.matches`.

So there are now four copies of one predicate that the PR body describes as shared ("The filter reuses the `NetworkRequestFilter` host/method semantics"). They're identical today, but the docs on `NetworkRequestFilter` spell out subtle rules (subdomains not implied, empty list matches nothing, OkHttp-vs-`java.net.URI` host parsing on Android) that a future change to one copy won't propagate to the other. The Android pair is already slightly divergent in spirit: `NetworkRequestFilter.hostOf` documents *why* it uses OkHttp's parser, while `NetworkTracesConfiguration` just calls `toHttpUrlOrNull()` with no note.

Suggestion: hold a `NetworkRequestFilter?` instead of loose `hosts`/`methods` and delegate:

```swift
internal struct NetworkTracesConfiguration: Codable, Sendable {
  var enabled: Bool = true
  var hosts: [String]?
  var methods: [String]?

  func allows(url: URL, method: String) -> Bool {
    guard enabled else { return false }
    var filter = NetworkRequestFilter()
    filter.hosts = hosts
    filter.methods = methods
    return filter.matches(url: url, method: method)
  }
}
```

(`NetworkRequestFilter` is a `Record`, so it may not be free to construct on the hot path — if that's the concern, caching one built at `setConfiguration` time keeps both the reuse and the cost down. On Android `NetworkRequestFilter` is a plain data class and composes directly.) That also lets the existing filter tests carry the matching matrix, and shrinks the new test files to the enabled-gate plus the encoding.

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/networkrequests/NetworkRequestPersistence.kt
```diff
@@ -23,14 +23,33 @@ private const val TAG = "ExpoAppMetrics"
 class NetworkRequestPersistence(
   private val database: MetricsDatabase,
   private val scope: CoroutineScope,
+  initialConfiguration: NetworkTracesConfiguration = NetworkTracesConfiguration(),
```
tsapeta: [💡 suggestion] A defaulted parameter placed *before* a required one means every positional caller has to pass it anyway, so the default only pays off for named arguments. Every call site in the PR uses named arguments, so it works — but it's a trap for the next caller and it's inconsistent with the iOS signature, where `configuration` sits before the trailing `sessionId` closure for a real reason (trailing-closure syntax).

Moving it after `sessionId` would keep the default useful positionally and leave the existing two-arg call in `AppMetricsModule.kt` working unchanged.

packages/expo-app-metrics/android/src/test/java/expo/modules/appmetrics/networkrequests/NetworkTracesConfigurationTest.kt
```diff
@@ -0,0 +1,80 @@
...
+    val restored = NetworkTracesConfiguration.fromJson(minimal.toJson())
+    assertEquals(minimal, restored)
+    assertNull(restored.hosts)
+    assertNull(restored.methods)
+  }
+
+  @Test
+  fun `falls back to the default policy for a malformed blob`() {
+    // A corrupt preferences entry must never disable recording or crash module creation.
+    for (blob in listOf("", "not json", "[1,2]", """{"hosts":"nope"}""")) {
```
tsapeta: [🐧 nit] The last blob passes for a different reason than the test name suggests. `{"hosts":"nope"}` is valid JSON and `optJSONArray("hosts")` returns `null` for a string value rather than throwing, so `runCatching` never catches anything — the result equals the default because `hosts` simply stayed `null`, not because the fallback fired.

That's still correct behavior, but the case doesn't exercise the `getOrDefault` path it looks like it's testing. A blob that actually throws inside the block, e.g. `{"hosts":[1,{}]}` (`getString` on a `JSONObject` element throws `JSONException`), would cover the real fallback. Worth keeping both.

packages/expo-observe/src/module.ts
```diff
@@ -8,6 +8,19 @@ import { isReactNavigationInstalled } from './integrations/react-navigation/reac
...
+ * Normalizes the `networkTraces` sugar (`true` / `false` / `{ filter }`) into the config shape
+ * the native producer persists. `configure` is a full replacement, so an absent `networkTraces`
+ * resets to the default: enabled, no filter.
+ */
+function networkTracesConfigFromOption(networkTraces: ObserveConfig['networkTraces']) {
+  const option = networkTraces ?? true;
+  if (typeof option === 'boolean') {
+    return { enabled: option };
+  }
+  return option.filter != null ? { enabled: true, filter: option.filter } : { enabled: true };
```
tsapeta: [💡 suggestion] Two small things here.

First, `{ enabled: false }` is unreachable through the object form: `networkTraces: { filter: ... }` always yields `enabled: true`. That's consistent with the type (`ObserveNetworkTracesConfig` has no `enabled` field), so it's fine — but it means `networkTraces: {}` silently means "record everything", which reads more like a typo than an intent at a call site. Worth considering whether the empty object should be rejected or at least documented.

Second, `configure` is documented as a full replacement, and this correctly resets to the default when `networkTraces` is omitted. But `Observe.configure` can be called more than once, and each call now writes to UserDefaults / SharedPreferences (the Android write uses `commit = true`, a synchronous disk write on the JS thread). A `configure` in a component body or an effect without a dep guard would do a blocking write per render. `apply()` plus the existing forward-only semantics would be enough here, or an early return when the normalized config is unchanged from the last one pushed.

tsapeta:
@expo-bot review

packages/expo-app-metrics/ios/NetworkRequests/NetworkTracesConfiguration.swift
```diff
@@ -0,0 +1,30 @@
...
+  /// Allowed HTTP methods, compared case-insensitively. `nil` allows every method.
+  var methods: [String]?
+
+  /// Whether a request should be recorded. Matching is delegated to `NetworkRequestFilter`, so the
+  /// capture gate and the JS-facing observer can't drift apart.
+  func allows(url: URL, method: String) -> Bool {
+    guard enabled else {
+      return false
+    }
+    var filter = NetworkRequestFilter()
```
tsapeta: [🏗️ **design**] This builds a fresh `NetworkRequestFilter` for every recorded request, while the Android twin builds it once per policy (`private val filter = NetworkRequestFilter(...)`) and documents that choice: "Built once per policy rather than per request; the fields never change after decoding."

The cost is not free here. `NetworkRequestFilter` is an ExpoModulesCore `Record`, so `hosts` and `methods` are `@Field` property wrappers, and `Field` is a `final class` holding a `Mutex<Set<FieldOption>>` plus an `AnyDynamicType` resolved in its initializer. Each `allows` call therefore allocates two class instances and two mutexes, then throws them away, on the request-completion path.

A stored `let filter` mirroring Android would remove that. It needs a custom `init` (or a `lazy`/memoized accessor) because the type is `Codable` and the synthesized decoder populates the stored properties directly, so the cached filter has to be derived after decoding rather than in a member-wise initializer.

packages/expo-app-metrics/ios/Tests/NetworkTracesConfigurationTests.swift
```diff
@@ -0,0 +1,76 @@
...
+
+  @Test
+  func `a request whose URL has no host never matches a host list`() {
+    // A host allowlist means "only these hosts"; a hostless URL can't prove membership.
+    let config = NetworkTracesConfiguration(enabled: true, hosts: ["api.example.com"], methods: nil)
+    #expect(!config.allows(url: URL(string: "file:///tmp/payload.json")!, method: "GET"))
+  }
+
+  @Test
+  func `round-trips through its persisted encoding`() throws {
```
tsapeta: [💡 suggestion] The round-trip test covers a well-formed blob, but there is no iOS equivalent of the Android `falls back to the default policy for a malformed blob` test.

The two platforms do not actually behave the same here, which is worth pinning down rather than assuming. Android's `fromJson` wraps the parse in `runCatching` and returns `NetworkTracesConfiguration()` for a corrupt blob. iOS goes through `UserDefaults.codable(forKey:as:)`, which does `try? JSONDecoder().decode(...)` and returns `nil`; the two call sites then apply `?? NetworkTracesConfiguration()`. Same end result, but reached through a different mechanism, and the iOS one depends on both call sites remembering the `??`. A test that stores a corrupt `Data` blob under the key and asserts the default comes back would lock that in.

A partial-blob case is also worth adding: a payload like `{"enabled": false}` with no `hosts`/`methods` decodes fine on iOS (the optionals stay `nil`), but `{"hosts": "nope"}` fails the whole decode and resets `enabled` to `true`, where Android's `optJSONArray` tolerates it and keeps the decoded `enabled`. That divergence is defensible (both fall open, which is the safe direction for a telemetry gate) but it is currently undocumented and untested.

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/networkrequests/NetworkRequestPersistence.kt
```diff
@@ -60,7 +79,14 @@ class NetworkRequestPersistence(
     // with the session INSERT and crash-report processing, so converting up to 200 requests
     // there would be the most expensive place to do it.
     scope.launch {
+      // Snapshotted once so the whole batch is judged by one policy. Re-reading per request would
+      // let a reconfigure landing mid-drain split the batch, writing the requests it reached
+      // first and dropping the rest.
+      val policy = configuration
```
tsapeta: [🐧 nit] `persist` reads the policy on the calling thread (an OkHttp dispatcher), while `persistBuffered` reads it inside the coroutine and snapshots it for the batch. That asymmetry is deliberate and the comments explain each half, but the batch snapshot buys less than the comment claims.

The stated hazard is that "a reconfigure landing mid-drain would split the batch, writing the requests it reached first and dropping the rest." A split batch is exactly what `persist` produces for live traffic at the same moment: requests completing before the reconfigure are written, those after are dropped. That is the intended forward-only semantics. So the batch is being held to a stricter rule than the live path for reasons that are not obvious from the code.

Not a bug, and the test does assert the intended behavior. But consider saying in the comment *why* the buffered batch deserves atomicity that live traffic does not, since the natural reading is that both should follow the same forward-only rule. My guess at the real reason is that the buffer's requests all predate the `configure` call that triggered the drain, so judging them by one policy keeps their treatment independent of drain timing.

packages/expo-observe/src/module.ts
tsapeta: [💡 suggestion] `option.filter != null ? { enabled: true, filter: option.filter } : { enabled: true }` drops the key when the filter is nullish, which is correct for the native side, but it passes a caller-supplied filter object through verbatim.

A filter like `{ hosts: null, methods: ['GET'] }` is valid per the JS `NetworkRequestFilter` type (`hosts?: string[] | null`), and both native records treat an explicit `null` the same as absent, so that works. Worth a test though: the existing cases only cover a fully-populated filter and an absent one, so the `null`-field path is unverified end to end even though the public type advertises it.

The `{ filter: {} }` case is also untested. It normalizes to `{ enabled: true, filter: {} }`, which both platforms read as "match everything" - same as no filter, just with an extra hop. Fine, but it is the shape a caller lands on when they build the filter conditionally and every branch is false, so it is worth locking in.

docs/pages/eas/observe/configuration.mdx
```diff
@@ -77,6 +77,30 @@ All metrics are grouped by environment. The environment value is derived from `p
...
+
+Pass an object to record only the requests you care about. Hosts match exactly and case-insensitively, so subdomains are not implied, and an omitted field matches every request:
+
+```ts
+Observe.configure({
+  networkTraces: { filter: { hosts: ['api.myapp.com'], methods: ['GET', 'POST'] } },
+});
+```
+
+This controls recording rather than only export: a request that is turned off or filtered out never reaches the local database. The setting applies to future requests, so spans recorded earlier in the launch are still sent. It also persists across launches, which means requests observed before `configure` runs on the next launch follow the last value you set.
```
tsapeta: [💡 suggestion] This paragraph explains the persistence correctly but stops short of the consequence a developer will actually hit.

Because the value survives launches and `configure` is a full replacement, a developer who sets `networkTraces: false` once, then later deletes that line from their `configure` call, gets recording back on - but only from the moment `configure` runs on the next launch. Requests before that point on that launch still follow the old persisted `false`. The inverse also holds: a narrow `hosts` allowlist keeps suppressing early-startup requests on every subsequent launch until a `configure` widens it.

That is the intended design (it is what makes early-startup requests follow the last-known setting at all), but as written a reader could reasonably expect removing the option to take effect immediately and completely. One sentence naming the "remove the option" case explicitly would prevent a confusing bug report.

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/AppMetricsModule.kt
```diff
@@ -390,3 +413,13 @@ data class MetricAttributes(
   @Field val routeName: String? = null,
   @Field val params: Map<String, Any>? = null
 ) : Record
+
+/**
+ * Payload of `setNetworkTracesConfig`: the normalized `networkTraces` setting pushed down by
+ * `Observe.configure`.
+ */
+@OptimizedRecord
+data class NetworkTracesConfigParam(
```
tsapeta: [🐧 nit] The iOS twin declares this payload `internal struct NetworkTracesConfigParam`, but this one is a top-level public `data class` in the module's package, alongside the already-public `MetricAttributes`. Since it is only ever constructed by the argument converter for one internal `Function`, consider `internal` so it does not read as part of the package's Kotlin surface.

Minor, and it matches how `MetricAttributes` is declared right above, so leaving it is defensible for consistency.

docs/pages/eas/observe/configuration.mdx
```diff
@@ -77,6 +77,30 @@ All metrics are grouped by environment. The environment value is derived from `p
...
+
+Completed network requests are recorded as trace spans by default. Each span carries the request URL, so you may want to narrow what gets recorded, or turn it off:
+
+```ts
+import { Observe } from 'expo-observe';
+
+Observe.configure({ networkTraces: false });
+```
+
+Pass an object to record only the requests you care about. Hosts match exactly and case-insensitively, so subdomains are not implied, and an omitted field matches every request:
```
amandeepmittal: ⚠️ **correctness** `prose.md#sentence-construction`

An empty array matches nothing, so `hosts: []` records no requests at all and the page never says so. Split the second sentence and state both cases, since a reader who sets an empty allowlist expecting "no restriction" silently loses every span.

```suggestion
Pass an object to record only the requests you care about. Hosts match exactly and case-insensitively, so subdomains are not implied. An omitted field matches every request, and an empty array matches none:
```

docs/pages/eas/observe/configuration.mdx
```diff
@@ -77,6 +77,30 @@ All metrics are grouped by environment. The environment value is derived from `p
...
+
+```ts
+Observe.configure({
+  networkTraces: { filter: { hosts: ['api.myapp.com'], methods: ['GET', 'POST'] } },
+});
+```
+
+This controls recording rather than only export: a request that is turned off or filtered out never reaches the local database. The setting applies to future requests, so spans recorded earlier in the launch are still sent. It also persists across launches, which means requests observed before `configure` runs on the next launch follow the last value you set.
+
+URLs are redacted before they are stored. Userinfo credentials and the query parameters that commonly carry signed-URL secrets are replaced with `REDACTED`.
```
amandeepmittal: ⚠️ **correctness** `prose.md#core-principles`

"Userinfo credentials" and "the query parameters that commonly carry signed-URL secrets" leave the reader guessing which values survive. The list is fixed and short, so name it. A reader whose secret sits in a parameter named something else needs to know it is stored as-is.

```suggestion
URLs are redacted before they are stored. A username and password in the URL are replaced with `REDACTED`, as are the query parameters `AWSAccessKeyId`, `Signature`, `sig`, `X-Amz-Signature`, `X-Amz-Credential`, `X-Amz-Security-Token`, and `X-Goog-Signature`. A secret in any other query parameter is stored as-is.
```

docs/pages/eas/observe/configuration.mdx
amandeepmittal: ✏️ **style** `review-ai-text.md#t2-antithesis`

Opening on the contrast ("rather than only export") makes the reader reconstruct what you ruled out before they learn what happens. Lead with the fact.

```suggestion
A request that is turned off or filtered out never reaches the local database, so nothing is stored and nothing is exported. The setting applies to future requests, so spans recorded earlier in the launch are still sent. It also persists across launches, so requests observed before `configure` runs on the next launch follow the last value you set.
```

amandeepmittal (changes requested):

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/AppMetricsPreferences.kt
```diff
@@ -41,4 +43,24 @@ object AppMetricsPreferences {
...
+      ?: return NetworkTracesConfiguration()
+    return NetworkTracesConfiguration.fromJson(json)
+  }
+
+  fun setNetworkTracesConfiguration(
+    context: Context,
+    configuration: NetworkTracesConfiguration
+  ) {
+    val prefs = context.getSharedPreferences(PREFS_NAME, Context.MODE_PRIVATE)
+    prefs.edit(commit = true) { putString(KEY_NETWORK_TRACES_CONFIGURATION, configuration.toJson()) }
```
tsapeta: [💡 suggestion] `commit = true` is a synchronous, fsync-backed write, and this runs on the JS thread: a sync `Function` body executes inline in the JSI host-function call frame (`SyncFunctionComponent.kt` registers via `registerSyncFunction`, which routes to `MethodMetadata::toSyncFunction`), with no queue hop. So `Observe.configure()` — typically called during startup — blocks JS for the duration of the write.

The rationale in the caller's comment is that the value must survive a process death before the hop runs. That's fair, but `apply()` also writes on a background thread that Android flushes on `Activity.onPause` and before the process is killed in the normal path; the window `commit` closes is a hard kill in the few ms after `configure`, which would cost one launch's worth of an already-forward-only setting.

If you want to keep `commit`, consider moving the whole preference write into the `scope.launch` (accepting that narrower window), or keep `commit` and note in the comment that it is a deliberate JS-thread stall. Either way it's worth being explicit, because the cost lands on the startup path.

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/networkrequests/NetworkTracesConfiguration.kt
```diff
@@ -0,0 +1,62 @@
...
+) {
+  /**
+   * Whether a request should be recorded. Matching is delegated to `NetworkRequestFilter`, so
+   * the capture gate and the JS-facing observer can't drift apart.
+   */
+  fun allows(url: String, method: String): Boolean {
+    if (!enabled) {
+      return false
+    }
+    return NetworkRequestFilter(hosts = hosts, methods = methods).matches(url, method)
```
tsapeta: [💡 suggestion] This allocates a fresh `NetworkRequestFilter` on every completed request, and `persist` calls `allows` *before* hopping to `scope` — so this runs on the OkHttp dispatcher thread. When `hosts` is set, `NetworkRequestFilter.matches` calls `hostOf`, which does a full `url.toHttpUrlOrNull()` parse.

That cuts against the intent stated in the comment directly below the gate in `persist`: "OkHttp dispatcher threads pay neither the URL parsing and JSON building nor the database write." With a `hosts` filter configured, they now pay a URL parse per request.

Delegating the match to `NetworkRequestFilter` is the right call — it's what keeps the gate from drifting from the observer. The fix is just to build the filter once instead of per call, e.g. a `private val filter by lazy { NetworkRequestFilter(hosts = hosts, methods = methods) }` on this data class (it's immutable, so the filter is too). That keeps the delegation and drops both the allocation and the repeated setup.

The parse itself is unavoidable if you want host matching at capture time, so this is about the per-call allocation, not about moving the gate.

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/AppMetricsModule.kt
```diff
@@ -136,6 +138,26 @@ class AppMetricsModule : Module(), UpdatesStateChangeListener {
...
+        )
+        // Persisted here rather than on the queue, so the setting survives even if the process
+        // dies before the hop below runs.
+        AppMetricsPreferences.setNetworkTracesConfiguration(context, configuration)
+        scope.launch {
+          // Applied on the same queue that installs the producer, so a configure racing startup
+          // either precedes the install (and is read from preferences there) or lands after it.
+          // Re-read instead of capturing: preferences are written in call order, so the last
+          // write wins no matter how two rapid calls interleave here.
+          networkRequestPersistence?.setConfiguration(
```
tsapeta: [🐧 nit] The "Re-read instead of capturing" reasoning is correct on iOS, where `AppMetricsActor.isolated` enqueues unordered tasks, and the parallel comment in `AppMetricsModule.swift` states exactly that. On Android it defends against a hazard that can't occur: `scope` is `appContext.modulesQueue`, which is a `CoroutineScope` over a single `HandlerThread` (`AppContext.kt:70-90`). Two `scope.launch` calls from the same thread run in submission order, so capturing `configuration` here would give the same last-write-wins result.

Not a bug — the re-read is correct, just doing nothing. The cost is that it re-reads SharedPreferences on the modules queue per call, and more importantly the comment tells a future reader that Android has an ordering hazard it doesn't have. Either capture `configuration` directly and drop the paragraph, or keep the re-read and reword to say it mirrors iOS, where the ordering guarantee genuinely is absent.

packages/expo-app-metrics/ios/Tests/NetworkTracesConfigurationTests.swift
```diff
@@ -0,0 +1,90 @@
...
+    #expect(decodedMinimal.hosts == nil)
+    #expect(decodedMinimal.methods == nil)
+  }
+
+  @Test
+  func `a corrupt blob decodes to nothing, so callers fall back to the default policy`() {
+    // The decode returns nil rather than a partial value, and both call sites apply
+    // `?? NetworkTracesConfiguration()`. Falling open matters more than the mechanism here: a
+    // corrupt preferences entry must never silently disable recording.
+    for blob in ["", "not json", "[1,2]", #"{"enabled":false,"hosts":"nope"}"#] {
```
tsapeta: [🐧 nit] Worth noting that the two platforms disagree on one of these blobs, and both tests encode their own behaviour as correct.

`{"enabled":false,"hosts":"nope"}` fails to decode here (Swift's `Codable` is strict about `hosts` being `[String]?`), so iOS falls back to the full default — including `enabled: true`. The Android test asserts the same blob shape reaches the default via a different route: `JSONObject` parses it fine and `optJSONArray` returns null for the non-array, so `hosts` ends up null — but `optBoolean("enabled", true)` would read the *stored* `false`. The Android test only covers `{"hosts":"nope"}` (no `enabled` key), so the divergence isn't exercised.

The practical impact is small: this only matters for a blob this code never writes, and both platforms fail open on the paths that matter. But the shared comment on both `NetworkTracesConfiguration` types says they mirror each other, so it'd be worth either adding the `enabled` key to the Android malformed-blob case to pin down what it does, or noting in the comment that partial-field recovery differs by platform.

amandeepmittal (approved):

packages/expo-app-metrics/android/src/test/java/expo/modules/appmetrics/networkrequests/NetworkRequestPersistenceTest.kt
```diff
@@ -480,6 +480,104 @@ class NetworkRequestPersistenceTest {
...
+      }
+    }
+    persistence.persistBuffered(buffered)
+    testScheduler.advanceUntilIdle()
+    assertTrue("the reconfigure must land mid-drain for this test to mean anything", reconfigured)
+    assertEquals("the batch must not be split by a mid-drain reconfigure", 4, allSpans().size)
+  }
+
+  @Test
+  fun `honors a persisted policy from the moment it is constructed`() = runTest(testDispatcher) {
```
jakub-agent, line 565: [💡 suggestion] This has the same body as `drops every request while recording is disabled` above — same constructor args, same `persist`, same `assertTrue(allSpans().isEmpty())`. Can we drop one? Only the name and comment differ.

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/networkrequests/NetworkRequestPersistence.kt
```diff
@@ -60,7 +79,15 @@ class NetworkRequestPersistence(
     // with the session INSERT and crash-report processing, so converting up to 200 requests
     // there would be the most expensive place to do it.
     scope.launch {
+      // Snapshotted once, unlike `persist`, which samples the policy per request. Everything in
+      // this batch was observed before the install that triggered the drain, so one policy keeps
+      // their treatment from depending on how far the loop happened to get before a reconfigure
+      // landed.
+      val policy = configuration
       for (request in requests) {
+        if (!policy.allows(request.url, request.method)) {
```
jakub-agent, line 88: [🏗️ **design**] The only `persistBuffered` test asserts all 4 rows are written, so deleting this `continue` still passes — it covers the snapshot, not the gate. Can you add one where the initial policy drops part of the batch?

packages/expo-app-metrics/ios/NetworkRequests/NetworkTracesConfiguration.swift
```diff
@@ -0,0 +1,27 @@
+// Copyright 2025-present 650 Industries. All rights reserved.
+
+import Foundation
+
+/// Which network requests get recorded as spans. Unlike the dispatch-side gates, a request
+/// rejected here never reaches disk.
+///
+/// Persisted, so requests observed before JS configures anything follow the last-applied setting.
+internal struct NetworkTracesConfiguration: Codable, Sendable {
```
jakub-agent, line 9: [🏗️ **design**] WDYT about holding a `NetworkRequestFilter` here instead of flattening it into `hosts`/`methods` and rebuilding it in `allows`? Those two fields now appear in four places, and this is the persisted shape, so it's the hardest one to change later. Is the blocker that `NetworkRequestFilter` isn't `Codable`?

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/AppMetricsPreferences.kt
jakub-agent, line 64: [🏗️ **design**] `commit = true` is a blocking disk write on the JS thread at startup, and iOS uses plain `defaults.set`, which doesn't flush synchronously either — so the hard-kill durability this buys doesn't exist on iOS anyway. Can we use `apply()` and have both platforms behave the same?

packages/expo-app-metrics/ios/AppMetricsModule.swift
```diff
@@ -64,6 +64,25 @@ public final class AppMetricsModule: Module, UpdatesStateChangeListener {
...
+      )
+      // Persist first so the setting survives the process; the live producer picks it up for
+      // every request recorded after this hop to the actor. Applies forward only: rows
+      // persisted earlier in the launch still dispatch.
+      AppMetricsUserDefaults.networkTracesConfiguration = configuration
+      AppMetricsActor.isolated {
+        // Re-read instead of capturing `configuration`: two rapid `configure()` calls create
+        // two unordered actor hops, while the persisted value is written in call order,
+        // reading it here makes the last write win regardless of task interleaving.
+        let latest = AppMetricsUserDefaults.networkTracesConfiguration ?? NetworkTracesConfiguration()
```
jakub-agent, line 81: [💡 suggestion] iOS re-reads the persisted value inside the hop, Android captures it, and each side has a paragraph explaining why. Can we pick one? Two mirrored files with opposite mechanisms will drift.

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/AppMetricsModule.kt
```diff
@@ -136,6 +138,27 @@ class AppMetricsModule : Module(), UpdatesStateChangeListener {
         GlobalAttributes.set(attributes)
       }
 
+      Function("setNetworkTracesConfig") { config: NetworkTracesConfigParam ->
+        val configuration = NetworkTracesConfiguration(
+          enabled = config.enabled,
+          hosts = config.filter?.hosts,
+          methods = config.filter?.methods
+        )
+        // Persisted here rather than on the queue, so the setting survives even if the process
```
jakub-agent, line 147: [🐧 nit] Nine lines of comment for three lines of code. Let's keep this to the first sentence and move the `commit` vs `apply` trade-off to one line next to the call.

packages/expo-observe/src/types.ts
```diff
@@ -81,6 +86,25 @@ export type ObserveConfig = {
...
+   * This controls recording, not just export: a request that's turned off or filtered out never
+   * reaches the local database. It applies to future requests only, so anything recorded earlier
+   * in the launch still gets sent. The setting persists across launches.
+   *
+   * @default true
+   */
+  networkTraces?: boolean | ObserveNetworkTracesConfig;
+};
+
+export type ObserveNetworkTracesConfig = {
```
jakub-agent, line 102: [🐧 nit] We now have `networkTraces`, `ObserveNetworkTracesConfig`, `NetworkTracesConfig`, `NetworkTracesConfigParam` and `NetworkTracesConfiguration` for one concept. Can we cut that down?

packages/expo-observe/src/types.ts
```diff
@@ -81,6 +86,25 @@ export type ObserveConfig = {
...
+  /**
+   * Only requests matching the filter are recorded. An omitted field matches every request; an
```
jakub-agent, line 104: [🐧 nit] nit: `hosts: ['https://api.myapp.com']` matches nothing and silently records zero traces. Worth a dev-mode warning when an entry contains `/` or `:`? Not necessary for this PR.

docs/pages/eas/observe/configuration.mdx
```diff
@@ -77,6 +77,32 @@ All metrics are grouped by environment. The environment value is derived from `p
...
+
+```ts
+Observe.configure({
+  networkTraces: { filter: { hosts: ['api.myapp.com'], methods: ['GET', 'POST'] } },
+});
+```
+
+A request that is turned off or filtered out never reaches the local database, so nothing is stored and nothing is exported. The setting applies to future requests, so spans recorded earlier in the launch are still sent. It also persists across launches, so requests observed before `configure` runs on the next launch follow the last value you set.
+
+Because the value persists, removing the option from your `configure` call does not take effect immediately. Recording returns to the default when `configure` next runs, so requests made before that point on that launch still follow the value you set previously. The same applies to a narrow `hosts` list: it keeps suppressing early-startup requests on every launch until a `configure` widens it.
```
jakub-agent, line 102: [💡 suggestion] Three dense sentences for one idea. Something like: "The value persists, so removing the option takes effect the next time `configure` runs, not immediately." If someone wants the full launch-by-launch detail they can check the API reference.

packages/expo-app-metrics/ios/AppMetricsModule.swift
```diff
@@ -195,6 +214,16 @@ private func storedSession(id: String) throws -> StoredSession? {
   return StoredSession(from: row)
 }
 
+/// Payload of `setNetworkTracesConfig`: the normalized `networkTraces` setting pushed down by
+/// `Observe.configure`.
+internal struct NetworkTracesConfigParam: Record {
```
jakub-agent, line 219: [🐧 nit] nit: you migrated `NetworkRequestFilter` to `@Record` in this PR — should this new one use it too, rather than `Record` + `@Field`?

jakub-agent (approved):

PR #48893, opened 2026-08-13
[app-metrics][observe] Add startSpan, withSpan, and recordSpan for custom trace spans

tsapeta:
# Why

With storage, export, and the capture-time config in place, apps still couldn't record their own spans. This adds the JS API designed for ENG-25894.

Closes ENG-26034.

# How

Three entry points over one native recorder: `startSpan` returns a handle you end yourself, `withSpan` wraps a callback and ends when it settles (a throw ends with an `error` status and rethrows), and `recordSpan` takes an already-measured window. Each produces the same write-once row the network producer uses, so storage and the exporter are untouched.

```ts
import { Observe } from 'expo-observe';

// A span you finish yourself. Its ids are readable immediately, so it can parent a child.
const checkout = Observe.startSpan('checkout', { attributes: { 'cart.items': 3 } });
const payment = Observe.startSpan('payment', { parent: checkout });
payment.end({ status: 'ok' });
checkout.addEvent('cart-validated', { attributes: { itemCount: 3 } });
checkout.end();

const receipt = await Observe.withSpan('submit-order', async (span) => {
  span.addEvent('validated');
  return submitOrder();
});

const startTime = Date.now();
decodeImage();
Observe.recordSpan('image-decode', { startTime, endTime: Date.now() });
```

Two decisions worth flagging. Parenting is explicit only: Hermes has no `AsyncLocalStorage`, and an implicit active-span context would misattribute concurrent async work. A handle garbage-collected without `end()` drops its span, because inventing an end time at collection would record the collector's schedule rather than the operation.

Custom and network spans share one `SpanWriter`, so both reach the `spans` table through a single write path.

# Test Plan

Native recorder suites on both platforms: id minting and parenting, end-once, status rules, attribute validation, event caps, and a database round-trip. Jest covers the wrappers and the `Observe` forwarding.

- `et native-unit-tests -p ios --packages expo-observe,expo-app-metrics`: 494 passed, 0 failed
- `et native-unit-tests -p android`: both packages BUILD SUCCESSFUL
- Jest: 68 suites / 588 tests for expo-observe, 8 suites / 56 tests for expo-app-metrics
- `tsc` clean for both packages

### DIFF
--- a/packages/expo-app-metrics/ios/AppMetrics.swift
+++ b/packages/expo-app-metrics/ios/AppMetrics.swift
@@ -142,6 +142,11 @@ public struct AppMetrics {
   ///
   /// This session starts when the app launches and continues until the app terminates.
   /// Unlike foreground sessions, there is only one main session per app process.
+  /// The shared sink every span producer writes completed rows through. Nonisolated so the
+  /// JS-thread module functions can hand it to span handles; it resolves the database on the
+  /// actor at write time.
+  static let spanWriter = SpanWriter()
+
   static let mainSession = MainSession()
 
   // MARK: - Foreground session
--- a/packages/expo-app-metrics/ios/AppMetricsModule.swift
+++ b/packages/expo-app-metrics/ios/AppMetricsModule.swift
@@ -193,7 +193,7 @@ public final class AppMetricsModule: Module, UpdatesStateChangeListener {
         attributes: options?.attributes,
         startTimestampMs: options?.startTime.map { Int64($0) } ?? currentUnixMilliseconds()
       )
-      return SpanHandle(recorder: recorder, onEnd: insertSpanRow)
+      return SpanHandle(recorder: recorder, onEnd: AppMetrics.spanWriter.schedule)
     }
 
     Function("recordSpan") { (name: String, options: RecordSpanOptions) in
@@ -209,7 +209,7 @@ public final class AppMetricsModule: Module, UpdatesStateChangeListener {
         startTimestampMs: Int64(startTime)
       )
       if let row = recorder.end(statusCode: nil, statusMessage: nil, endTimestampMs: Int64(endTime)) {
-        insertSpanRow(row)
+        AppMetrics.spanWriter.schedule(row)
       }
     }
 
@@ -342,21 +342,6 @@ private func spanStatusCode(from status: String?) throws -> Int? {
     return SpanRow.statusError
   case .some(let other):
     throw InvalidSpanStatusException(other)
-  }
-}
-
-/// Inserts a completed span row, hopping to the metrics actor. Failures are logged and
-/// swallowed — recording telemetry must never break the caller.
-private func insertSpanRow(_ row: SpanRow) {
-  Task { @AppMetricsActor in
-    guard let database = AppMetrics.database else {
-      return
-    }
-    do {
-      try database.insert(span: row)
-    } catch {
-      logger.warn("[AppMetrics] Failed to persist span \"\(row.name)\": \(error.localizedDescription)")
-    }
   }
 }
 
--- a/packages/expo-app-metrics/ios/AppMetricsModule.swift
+++ b/packages/expo-app-metrics/ios/AppMetricsModule.swift
@@ -183,6 +183,69 @@ public final class AppMetricsModule: Module, UpdatesStateChangeListener {
         observer.setFilter(filter)
       }
     }
+
+    Function("startSpan") { (name: String, options: StartSpanOptions?, parent: SpanHandle?) -> SpanHandle in
+      let recorder = SpanRecorder(
+        name: try validatedSpanName(name),
+        sessionId: AppMetrics.mainSession.id,
+        parentTraceId: parent?.recorder.traceId,
+        parentSpanId: parent?.recorder.spanId,
+        attributes: options?.attributes,
+        startTimestampMs: options?.startTime.map { Int64($0) } ?? currentUnixMilliseconds()
+      )
+      return SpanHandle(recorder: recorder, onEnd: insertSpanRow)
+    }
+
+    Function("recordSpan") { (name: String, options: RecordSpanOptions) in
+      guard let startTime = options.startTime, let endTime = options.endTime else {
+        throw MissingSpanWindowException()
+      }
+      // One code path with `startSpan`: an ephemeral recorder validates the attributes and
+      // produces the same write-once row shape.
+      let recorder = SpanRecorder(
+        name: try validatedSpanName(name),
+        sessionId: AppMetrics.mainSession.id,
+        attributes: options.attributes,
+        startTimestampMs: Int64(startTime)
+      )
+      if let row = recorder.end(statusCode: nil, statusMessage: nil, endTimestampMs: Int64(endTime)) {
+        insertSpanRow(row)
+      }
+    }
+
+    Class("Span", SpanHandle.self) {
+      Constructor { () -> SpanHandle in
+        throw SpanConstructorUnavailableException()
+      }
+
+      Property("traceId") { (span: SpanHandle) in
+        return span.recorder.traceId
+      }
+
+      Property("spanId") { (span: SpanHandle) in
+        return span.recorder.spanId
+      }
+
+      Function("setAttributes") { (span: SpanHandle, attributes: [String: Any]) in
+        span.recorder.setAttributes(attributes)
+      }
+
+      Function("addEvent") { (span: SpanHandle, name: String, options: SpanEventOptions?) in
+        span.recorder.addEvent(
+          name: name,
+          attributes: options?.attributes,
+          timeMs: options?.time.map { Int64($0) } ?? currentUnixMilliseconds()
+        )
+      }
+
+      Function("end") { (span: SpanHandle, options: SpanEndOptions?) in
+        span.end(
+          statusCode: try spanStatusCode(from: options?.status),
+          statusMessage: options?.message,
+          endTimestampMs: options?.endTime.map { Int64($0) } ?? currentUnixMilliseconds()
+        )
+      }
+    }
   }
 
   public func updatesStateDidChange(_ event: [String: Any]) {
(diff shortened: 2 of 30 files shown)


packages/expo-app-metrics/ios/AppMetricsModule.swift
```diff
@@ -181,6 +181,69 @@ public final class AppMetricsModule: Module, UpdatesStateChangeListener {
...
+      guard let startTime = options.startTime, let endTime = options.endTime else {
+        throw MissingSpanWindowException()
+      }
+      // One code path with `startSpan`: an ephemeral recorder validates the attributes and
+      // produces the same write-once row shape.
+      let recorder = SpanRecorder(
+        name: try validatedSpanName(name),
+        sessionId: AppMetrics.mainSession.id,
+        attributes: options.attributes,
+        startTimestampMs: Int64(startTime)
```
tsapeta: [🚨 **critical**] `Int64(Double)` traps on `NaN`, `±Infinity`, and any value outside `Int64`'s range, so this crashes the host app on caller input that JS accepts as an ordinary number.

I verified the trap with a compiled Swift binary (`swiftc -O`): both `Int64(Double.nan)` and `Int64(1e30)` exit with SIGTRAP (133). `expo-modules-core`'s `Double` conversion has no NaN or finiteness guard, so a plain JS `NaN` reaches Swift as `Double.nan`.

```ts
const startTime = Number(someHeader);     // NaN when the header is absent
Observe.recordSpan('image-decode', { startTime, endTime: Date.now() });  // crash
```

The same conversion appears at four other sites in this file: `startSpan`'s `startTime` (L192), `recordSpan`'s `endTime` (L209), `addEvent`'s `time` (L235), and `end`'s `endTime` (L243). All five are reachable from JS.

Android doesn't crash: Kotlin's `Double.toLong()` saturates (NaN → 0, ±Inf → `Long.MIN`/`MAX_VALUE`), so today the same call silently records a garbage timestamp on one platform and kills the app on the other.

A shared clamping helper would fix both the crash and the divergence, e.g.:

```swift
/// Unix-epoch milliseconds from a JS number. Non-finite and out-of-range values have no
/// meaning as a timestamp, and `Int64(_:)` traps on them, so they fall back to `now`.
private func unixMilliseconds(from value: Double) -> Int64? {
  guard value.isFinite, value >= Double(Int64.min), value <= Double(Int64.max) else {
    return nil
  }
  return Int64(value)
}
```

Then `options?.startTime.flatMap(unixMilliseconds(from:)) ?? currentUnixMilliseconds()`, and for `recordSpan` either the same fallback or a thrown `Exception` (rejecting a bad window loudly fits the existing `MissingSpanWindowException` precedent better than inventing a timestamp for an operation the caller claims to have measured).

packages/expo-app-metrics/ios/AppMetricsModule.swift
```diff
@@ -220,6 +283,100 @@ internal struct NetworkTracesConfigParam {
...
+
+  @Field
+  var attributes: [String: Any]?
+}
+
+/// Validates a caller-provided span name. The ingestion endpoint rejects spans whose name is
+/// empty, so failing loudly at the call site beats silently recording a span the server drops.
+private func validatedSpanName(_ name: String) throws -> String {
+  let trimmed = name.trimmingCharacters(in: .whitespacesAndNewlines)
+  guard !trimmed.isEmpty else {
```
tsapeta: [🏗️ **design**] `validatedSpanName` checks only emptiness, but this package already established two more rules for caller-supplied telemetry names in `validateEventName` (`ios/LogEvents/EventNameValidation.swift:17-35`): reject the reserved `expo.` prefix, and reject names over 256 characters.

Both apply just as much to spans:

- **Reserved prefix.** The SDK's own network producer writes `expo.http.*` attributes and an `expo.http.redirect` event, and span *attributes* are already protected — `sanitizeLogEventAttributes` drops anything matching `^expo\..+`. So within this one feature, `startSpan('x', { attributes: { 'expo.foo': 1 } })` is blocked while `startSpan('expo.http.client')` is not. The stated reason for the attribute rule ("SDK-emitted records stay distinguishable from app-emitted ones in the backend") applies identically to the name.
- **Length.** A runaway template literal (`startSpan(`checkout-${JSON.stringify(cart)}`)`) produces an unbounded name that reaches the server on every span. `validateEventName`'s comment gives the same rationale.

The mirrored Kotlin `validatedSpanName` has the same gap.

Note the difference in failure mode is intentional and worth keeping: `logEvent` drops the record, while `startSpan` must return a handle, so throwing (as it already does for empty) is right here.

packages/expo-app-metrics/src/spans.ts
```diff
@@ -0,0 +1,89 @@
...
+): Promise<T> {
+  const span = startSpan(name, options);
+  try {
+    const result = await callback(span);
+    span.end();
+    return result;
+  } catch (error) {
+    span.end({
+      status: 'error',
+      message: error instanceof Error ? error.message : String(error),
```
tsapeta: [🏗️ **design**] Both branches of this expression can themselves throw, and when they do the span is never ended and the caller sees a different error than the one they threw.

I confirmed the behavior in Node against a copy of this function:

```js
const evil = new Error('ok');
Object.defineProperty(evil, 'message', { get() { throw new Error('boom'); } });
await withSpan('checkout', () => { throw evil; });
// rethrown as: boom      <- not the error the callback threw
// span ended?  0         <- span leaked, never persisted
```

`String(error)` has the same hazard for a throwable with a throwing `toString` or a `Symbol`. This contradicts the documented contract directly above ("is then rethrown, so the caller's own error handling is unaffected") — the substituted error will confuse any `catch` that inspects the error's type or fields, and it is the kind of thing that only shows up in production, inside someone else's rejected promise.

A `finally` for the end and a guarded message read fix both halves:

```ts
export async function withSpan<T>(
  name: string,
  callback: (span: Span) => T | Promise<T>,
  options?: StartSpanOptions
): Promise<T> {
  const span = startSpan(name, options);
  try {
    const result = await callback(span);
    span.end();
    return result;
  } catch (error) {
    // Reading a message off an arbitrary throwable can itself throw (a throwing getter, an
    // exotic object). The span must still end, and with the caller's original error.
    let message: string | undefined;
    try {
      message = error instanceof Error ? error.message : String(error);
    } catch {}
    span.end({ status: 'error', message });
    throw error;
  }
}
```

The recorder's end-once rule already makes a redundant `end()` safe, so there's no double-write risk in restructuring this.

packages/expo-app-metrics/ios/Spans/SpanRecorder.swift
```diff
@@ -0,0 +1,153 @@
...
+      sessionId: sessionId,
+      traceId: traceId,
+      spanId: spanId,
+      parentSpanId: parentSpanId,
+      name: name,
+      kind: SpanRow.internalKind,
+      startTimestampMs: startTimestampMs,
+      endTimestampMs: endTimestampMs,
+      statusCode: statusCode,
+      statusMessage: statusCode == SpanRow.statusError ? statusMessage : nil,
```
tsapeta: [🏗️ **design**] `statusMessage` is stored and exported with no length bound, and `withSpan` feeds it `error.message` — which in practice carries whatever a library chose to put there (parser errors and network errors routinely run to hundreds of characters, sometimes with an embedded payload).

This package bounds every other caller-supplied free-text field: `validateEventBody` truncates to `maxEventBodyLength`, `validateEventName` rejects past 256. The exporter doesn't bound it either — `OTSpanStatus(code:message:)` passes it straight to the wire on both platforms — so an app that ends many spans from caught errors ships an unbounded string per span to the traces endpoint.

Truncating here (mirroring `validateEventBody`'s truncate-rather-than-drop choice, since losing the status entirely over a long message would be worse) would close it on both platforms.

packages/expo-app-metrics/ios/Spans/SpanRecorder.swift
```diff
@@ -0,0 +1,153 @@
...
+      }
+      state.attributes.merge(sanitized) { _, new in new }
+    }
+  }
+
+  /// Appends a point-in-time event. Events whose name trims to empty are dropped with a warning
+  /// (the server drops them anyway), as is anything past `maxEventCount`. No-op after `end`.
+  func addEvent(name: String, attributes: [String: Any]?, timeMs: Int64) {
+    let trimmedName = name.trimmingCharacters(in: .whitespacesAndNewlines)
+    guard !trimmedName.isEmpty else {
```
tsapeta: [💡 suggestion] Event names get the empty check but not the reserved-prefix check, so `span.addEvent('expo.http.redirect', { attributes: … })` produces an event indistinguishable on the wire from the one the SDK's own network producer emits (`ios/NetworkRequests/NetworkRequestPersistence.swift:116`).

That matters more than for span names, because `expo.http.redirect` is a name the backend is likely to interpret rather than just display. The same `^expo\..+` rule that `sanitizeLogEventAttributes` already applies to this event's *attributes* would cover it — dropping the event with a warning matches how `addEvent` already handles a nameless event.

An event-name length bound belongs here too, for the same reason as the span name.

packages/expo-app-metrics/ios/Spans/SpanRecorder.swift
```diff
@@ -0,0 +1,153 @@
...
+    self.startTimestampMs = startTimestampMs
+    // Attributes go through the same validation as log-event attributes (empty keys, the
+    // reserved `expo.*` namespace, and the per-record cap are dropped with a warning).
+    self.state = Mutex(State(attributes: sanitizeLogEventAttributes(attributes).attributes ?? [:]))
+  }
+
+  /// Merges attributes into the span; keys set later win. Values are validated the same way as
+  /// at construction. No-op after `end`.
+  func setAttributes(_ attributes: [String: Any]) {
+    guard let sanitized = sanitizeLogEventAttributes(attributes).attributes else {
```
tsapeta: [💡 suggestion] Reusing `sanitizeLogEventAttributes` for span attributes is the right call, but its warnings are hardcoded to name the wrong API. A developer who writes

```ts
span.setAttributes({ 'expo.custom': 1 });
```

sees `[AppMetrics] logEvent dropped attributes that overlap SDK-set keys or use the reserved 'expo.' namespace: 'expo.custom'.` — pointing at a function they never called, which is exactly the kind of misdirection the repo's error-message guidance (state *what* failed at the user's level of abstraction) is meant to prevent. There are three such messages in `AttributeValidation.swift:86-104`, mirrored on Android.

Threading a caller label through (`sanitizeLogEventAttributes(attributes, source: "setAttributes")`, defaulting to `"logEvent"` so existing call sites are untouched) would keep the shared validation while making the warning actionable.

Separately: the `droppedCount` this returns is discarded on the span path. `logEvent` persists it as `droppedAttributesCount`, and `SpanRow` has no equivalent column — worth a note in the recorder's doc comment that span attribute drops are reported only via the log, so it doesn't read as an oversight later.

packages/expo-app-metrics/ios/Spans/SpanRecorder.swift
```diff
@@ -0,0 +1,153 @@
...
+    let sanitized = sanitizeLogEventAttributes(attributes).attributes
+    state.withLock { state in
+      if state.ended {
+        return
+      }
+      guard state.events.count < Self.maxEventCount else {
+        logger.warn("[AppMetrics] Dropping a span event past the \(Self.maxEventCount)-event limit")
+        return
+      }
+      state.events.append(PendingEvent(name: trimmedName, timeMs: timeMs, attributes: sanitized))
```
tsapeta: [💡 suggestion] Caller-supplied event times aren't constrained to the span window, so an event can land outside the span it belongs to.

The network producer treats that as invalid and guards against it explicitly (`NetworkRequests/NetworkRequestPersistence.swift:125-132`): "An OTLP event outside its own span has no valid place on the timeline", omitting `timeMs` so the exporter anchors the event to the span start. The new path has no such guard, and two ordinary sequences reach it:

- `span.addEvent('checkpoint', { time: someEarlierTimestamp })` with a stale measurement;
- `startSpan('x', { startTime: Date.now() - 60_000 })` followed by `addEvent` with no `time`, where the default `now` is inside the window — fine — but `end({ endTime })` with an explicit earlier `endTime` leaves already-buffered events past the end.

The start-side clamp can be applied in `addEvent` (the start timestamp is known there); the end-side one only at `end`, where the window is finally known. Following the network producer and dropping the out-of-window `timeMs` (rather than the event) keeps the event and anchors it to the span start.

packages/expo-app-metrics/ios/AppMetrics.swift
```diff
@@ -148,6 +148,11 @@ public struct AppMetrics {
   ///
   /// This session starts when the app launches and continues until the app terminates.
   /// Unlike foreground sessions, there is only one main session per app process.
+  /// The shared sink every span producer writes completed rows through. Nonisolated so the
+  /// JS-thread module functions can hand it to span handles; it resolves the database on the
+  /// actor at write time.
+  static let spanWriter = SpanWriter()
```
tsapeta: [🐧 nit] `spanWriter` landed inside `mainSession`'s doc comment, under the `// MARK: - Main session` heading. The result is that `mainSession` is now undocumented, and `spanWriter`'s doc block opens with two paragraphs describing the main session before reaching the sentence that's actually about the writer.

Moving it above the `// MARK: - Main session` line with just its own three lines restores both.

packages/expo-app-metrics/src/module.web.ts
```diff
@@ -55,6 +55,16 @@ class ExpoAppMetricsModule extends NativeModule implements ExpoAppMetricsModuleT
   logEvent(name: string, options?: LogEventOptions) {}
   setGlobalAttributes(attributes?: Record<string, LogAttributeValue> | null) {}
   setNetworkTracesConfig() {}
+  startSpan(): any {
```
tsapeta: [🐧 nit] The web stub returns a fresh object literal with `traceId: ''` and `spanId: ''` on every call. Cross-platform code that reads the ids — the tester screen in this PR does exactly that, `shortId(payment.traceId)` — gets an empty string on web rather than something obviously inert, and two spans are indistinguishable.

The neighbouring `WebSession` stub takes the other approach (`readonly id = 'web-session'`), which reads as deliberate at a call site. Generating throwaway hex ids of the right length, or using a fixed sentinel, would make web behave predictably for code that logs or correlates them.

The `: any` return is also the only one in this file that opts out of the interface's type; `Span` is a `declare class`, so a cast like the neighbouring `as unknown as typeof Session` would keep the stub checked against the real shape and catch drift if `Span` gains a member.

tsapeta:
@expo-bot review

PR #48895, opened 2026-08-13
[router] Remove the deprecated `Link` and `useLinkProps` exports from `expo-router/react-navigation`

Ubax:
# Why

Remove the deprecated `Link` and `useLinkProps` exports from expo-router/react-navigation

# How

# Test Plan

CI

# Checklist

---


packages/expo-router/src/react-navigation/elements/Button.tsx
```diff
@@ -80,6 +51,12 @@ function ButtonBase({
   return (
     <PlatformPressable
       {...rest}
+      onPress={(event) => {
+        onPress?.(event);
+        if (href && !event?.defaultPrevented) {
```
jakub-agent: [🚨 **critical**] This guard is always false on web, so `router.navigate(href)` never runs there.

`PlatformPressable.handlePress` calls `e.preventDefault()` *before* invoking `onPress` (see `PlatformPressable.tsx`, the `Platform.OS === 'web'` branch — the comment there says "call `onPress` only when browser default is prevented"). By the time this handler sees the event, `defaultPrevented` is already `true`.

Combined with `href` not reaching `PlatformPressable` (see my other comment), `<Button href="/foo">` is a complete no-op on web — the app doesn't navigate and the browser has no anchor to follow.

I verified this in jsdom by rendering `PlatformPressable` with this exact `onPress` body and a stubbed `navigate`:

```
tag: BUTTON   href attr: null   navigate called times: 0
```

On native the guard works as intended (no `preventDefault()` in the non-web branch), which is why the three added `.test.ios.tsx` cases pass.

Suggested shape — read `defaultPrevented` from the user handler only, and let the anchor handle web:

```tsx
onPress={(event) => {
  onPress?.(event);
  if (href && shouldHandleMouseEvent(event)) {
    router.navigate(href);
  }
}}
```

…or track cancellation yourself rather than relying on `defaultPrevented`, which `PlatformPressable` has already consumed. Either way, please add a `Button.test.web.tsx` covering `href` navigation — the current tests are iOS-only and would not have caught this.

packages/expo-router/src/react-navigation/elements/Button.tsx
```diff
@@ -80,6 +51,12 @@ function ButtonBase({
   return (
     <PlatformPressable
```
jakub-agent, line 82: [🏗️ **design**] `href` is destructured out of props and never forwarded to `PlatformPressable`, so on web the button loses its anchor.

Before this PR, `ButtonLink` spread `useLinkProps(...)`'s return value (`{ href, role, onPress }`) into `ButtonBase`, which passed it through `...rest` to `PlatformPressable`. `PlatformPressable` uses `rest.href` to set `role={href != null ? 'link' : 'button'}`, and react-native-web renders a real `<a href>` for it. I confirmed both halves in jsdom:

```
href forwarded:     tag: A        href attr: /profile
href not forwarded: tag: BUTTON   href attr: null
```

So this drops, on web: real anchor semantics, `role="link"` for screen readers, cmd/middle-click open-in-new-tab, the browser's hover URL preview, and crawlability.

`Href` can be an object, so it needs resolving before it can go on the anchor — something like `href={href != null ? resolveHref(href) : undefined}` (from `../../link/href`). Worth checking whether `useLinkToPathProps` is a better fit here, since it already handles group-segment stripping, base URL, and the web click rules.

packages/expo-router/src/link/useLinkToPathProps.tsx
```diff
@@ -39,6 +40,15 @@ type UseLinkToPathPropsOptions = LinkToOptions & {
 
 export default function useLinkToPathProps({ href, ...options }: UseLinkToPathPropsOptions) {
   const onPress = (event?: MouseEvent<HTMLAnchorElement> | GestureResponderEvent) => {
+    if (
+      Platform.OS === 'web' &&
```
jakub-agent: [🏗️ **design**] Is this intentional in this PR? It's a user-visible behavior change to `Link` on web, and it isn't mentioned in the title or the changelog.

Previously an external `href` on web fell through to `linkTo`, which calls `Linking.openURL(href)`. react-native-web's `Linking.openURL` defaults the target to `_blank` (`react-native-web/dist/exports/Linking/index.js`: `target = '_blank'`, then `window.open(urlToOpen, target, 'noopener')`). So `<Link href="https://expo.dev">` opened a **new tab**. With this early return the browser handles the anchor instead, so it now navigates in the **same tab** unless the caller passes `target="_blank"`.

Deferring to the browser is the better default — it respects `target`, `download`, and modifier keys, and it isn't subject to popup blocking. But apps relying on external links opening a new tab will change behavior, so this needs its own `🛠 Breaking changes` entry.

Two notes on the condition itself:
- The `event?.currentTarget instanceof HTMLAnchorElement` check means a `Link` rendered via `asChild` onto a non-anchor still takes the old `linkTo` path. That's a reasonable fallback, but it makes external-link behavior depend on what the child renders — worth a short comment saying so.
- `HTMLAnchorElement` is only safe to reference because `Platform.OS === 'web'` short-circuits first. That's correct, just fragile if the clauses are ever reordered.

jakub-agent (changes requested):

Removing `Link` / `useLinkProps` from `expo-router/react-navigation` is clean — nothing in the repo consumed them, `react-navigation/index.ts` re-exports with `export *` so no entry point needs an update, and `__tests__/exports.test.ts` keeps passing.

The rewritten `elements/Button` is the problem. `href` is consumed by `Button` and never forwarded to `PlatformPressable`, and the `!event?.defaultPrevented` guard is always false on web because `PlatformPressable` calls `event.preventDefault()` before invoking `onPress`. The result is that `<Button href="...">` does nothing at all on web: no `router.navigate` call, and no anchor for the browser to follow. I reproduced this in jsdom with the PR's exact `onPress` body — `navigate` called 0 times, rendered tag `BUTTON`, `href` attribute `null`. The three added tests are `.test.ios.tsx` only, so they pass while web is broken.

One more thing to confirm: the `useLinkToPathProps` change is a user-visible behavior change to `Link` on web, unrelated to the PR title and not in the changelog. Details inline.

Minor: the new changelog entry has no `([#48895](...) by [@Ubax](...))` attribution.

packages/expo-router/src/react-navigation/elements/Button.tsx
Ubax: Handle it the same way as before the change - check `useLinkProps`

packages/expo-router/src/react-navigation/elements/Button.tsx
Ubax: Same as in previous comment, use the useLinkProps implementation

packages/expo-router/src/link/useLinkToPathProps.tsx
Ubax: Mention as breaking change in changelog

Align both paths with browser as much as possible

Ubax:
Address https://github.com/expo/expo/pull/48917#discussion_r3804408822 as well

jakub-agent (approved):

PR #48898, opened 2026-08-13
[ios][updates] Fix two launch-time crashes reachable from native consumers of the updates controller

spsaucier:
# Why

`EnabledAppController` is published to `UpdatesControllerRegistry` by `initializeWithoutStarting()`, but `startupProcedure` is only assigned in `start()`. Anything that resolves the controller from the registry in between — `expo-app-metrics` does exactly this from its `OnCreate`, and it is the first native consumer of the controller to ship — can reach the public accessors before that assignment.

**1. Nil unwrap of an implicitly-unwrapped optional.** `launchedUpdateId`, `launchAssetPath` and `launchAssetUrl()` all go through `startupProcedure`, which is declared `StartupProcedure!`. Reading any of them pre-`start()` traps with `EXC_BREAKPOINT`. `isStarted` is not a usable guard for a caller — it is set before the assignment, not after.

**2. Data race on `stateChangeListeners`.** It is a plain `Dictionary`, mutated by `subscribeToUpdatesStateChanges` / `unsubscribeFromUpdatesStateChanges` from whatever thread the subscriber happens to be on (module registry setup, during launch), and read by `UpdatesStateMachine.processEvent` on its own queue. The read side additionally iterates `.keys` and does a second subscript lookup per key, so a concurrent `removeValue` between the two is enough to fault. Symptom is `EXC_BAD_ACCESS` inside `processEvent`.

# What this changes

1. Optional-chain the three accessors. They now return `nil` before startup rather than trapping. The `StartupProcedure!` declaration is unchanged, so no internal call site is affected — this is a three-character change at each of the three public reads.
2. Make `stateChangeListeners` private, guard it with an `NSLock`, and add an internal `stateChangeListenersSnapshot()`. `UpdatesStateMachine` notifies off the snapshot, outside the lock, so a listener that subscribes or unsubscribes re-entrantly from `updatesStateDidChange` can't deadlock.

One behavior note worth calling out explicitly: notifying off a snapshot means `remove()` no longer cancels a notification already in flight, so a listener must remain safe to call immediately after unsubscribing. That was already true of the previous code in practice — the dictionary held strong references and the old loop could equally be mid-iteration — but the snapshot makes the window deterministic rather than incidental, and it seemed worth stating rather than leaving implicit.

# Impact we measured

Downstream in a production React Native app on Expo SDK 56 (`expo-updates` 56.0.18), these are two of three crashes in a launch-time cluster that appeared the moment we adopted `expo-observe` / `expo-app-metrics`:

| Crash | Events / users (14d) | Signal |
| --- | --- | --- |
| `EnabledAppController.launchedUpdateId.getter` — `EXC_BREAKPOINT` | 392 / 373 | 100% in the first second of the session |
| `UpdatesStateMachine.processEvent(_:)` — `EXC_BAD_ACCESS` | 167 / 165 | — |

Both are launch crashes, so there is no degraded mode — the app dies before anything renders. We shipped the equivalent change as a `patch-package` patch and both went to zero on the build that carried it. The third crash in that cluster was in `react-native-reanimated` and is not related to this PR.

# Testing

Honest accounting: **I have not run the `expo-updates` iOS test suite against this branch.** I don't have the monorepo's iOS toolchain set up locally and I didn't want to report a green run I didn't produce.

What I do have is the downstream evidence above: the same two changes, vendored into a production app via `patch-package`, across a real user population, with the crashes going to zero. I also checked that no code outside `EnabledAppController` still references `stateChangeListeners` directly now that it's private — `UpdatesStateMachine` was the only reader.

Two things I'd specifically like a maintainer's eye on, because I can't settle them from outside:

- Whether returning `nil` pre-startup is the semantics you want for these accessors, or whether you'd rather the registry not publish the controller until `start()` has run. The latter is the deeper fix and I'd defer to you on it; this PR takes the conservative option.
- Whether `NSLock` is the right primitive here versus routing everything onto the state machine's existing queue. I picked the lock because it's local and doesn't change threading assumptions for existing subscribers, but you know the intended concurrency model better than I do.

# Backport

This targets `main` (SDK 57). We hit it on SDK 56 and the defects are identical there — `expo-updates@56.0.24` still has both. Happy to open a `sdk-56` backport if that's useful; let me know which branch you'd want it against.


brentvatne:
@expo-bot verify

brentvatne:
@expo-bot review

packages/expo-updates/ios/EXUpdates/EnabledAppController.swift
```diff
@@ -213,16 +227,19 @@ public class EnabledAppController: InternalAppControllerInterface, UpdatesInterf
     return config.requestHeaders
   }
 
+  // `startupProcedure` is only assigned in `start()`, but the controller is published to
+  // `UpdatesControllerRegistry` by `initializeWithoutStarting()`, so a native consumer can read
+  // this first. `isStarted` is not a usable guard — it is set before the assignment.
   public var launchedUpdateId: UUID? {
-    return startupProcedure.launchedUpdate()?.updateId
+    return startupProcedure?.launchedUpdate()?.updateId
```
Kudo, line 228: [🏗️ **design**] Android avoids this window entirely by building the procedure in the constructor (`EnabledUpdatesController.kt:121`), so its `launchedUpdateId` / `launchAssetPath` already return null pre-start.

That looks possible on iOS too: `StartupProcedure.init` only stores its args and builds `ErrorRecovery`, and `config` is already final when `EnabledAppController.init` runs (`AppController.swift:278`). It would remove the nil window rather than tolerate it, and let the `StartupProcedure!` declaration go away instead of optional-chaining three reads.

Do you want that here, or keep this as the conservative fix and follow up separately?

packages/expo-updates/ios/EXUpdates/UpdatesStateMachine.swift
```diff
@@ -399,10 +399,8 @@ internal class UpdatesStateMachine {
       }
       // Notify the controller state change listener
       if let controller = UpdatesControllerRegistry.sharedInstance.controller as? EnabledAppController {
-        controller.stateChangeListeners.keys.forEach {subscriptionId in
-          if let listener = controller.stateChangeListeners[subscriptionId] {
-            listener.updatesStateDidChange(event.toMap)
-          }
+        controller.stateChangeListenersSnapshot().forEach { listener in
```
Kudo, line 402: [💡 suggestion] Android has the same unsynchronized map and the same iterate-keys-then-subscript loop: `EnabledUpdatesController.kt:88` is a plain `mutableMapOf()`, read at `UpdatesStateMachine.kt:102-103`, and `AppMetricsModule.kt` subscribes from `OnCreate` just like iOS does. There it surfaces as `ConcurrentModificationException` instead of `EXC_BAD_ACCESS`.

Keeping this PR iOS-only is fine, but the Android side likely needs the same treatment.

packages/expo-updates/ios/EXUpdates/EnabledAppController.swift
```diff
@@ -181,20 +181,34 @@ public class EnabledAppController: InternalAppControllerInterface, UpdatesInterf
 
   // MARK: - UpdatesInterface
 
-  internal var stateChangeListeners: [String: any UpdatesStateChangeListener] = [:]
+  // Native subscribers (e.g. expo-app-metrics) subscribe from the module-registry thread while the
+  // state machine iterates this dictionary on its own queue, so every access is locked. Listeners
+  // are invoked outside the lock, off a snapshot: re-entrant subscribe/unsubscribe from inside
+  // `updatesStateDidChange` cannot deadlock, but `remove()` does not cancel a notification already
+  // in flight, so a listener must stay safe to call after it unsubscribes.
+  private var stateChangeListeners: [String: any UpdatesStateChangeListener] = [:]
```
Kudo:
```suggestion
  private var stateChangeListeners: Mutex<[String: any UpdatesStateChangeListener]>([:])
```

maybe we can leverage the Mutex syntax suger without lock and defer unlock everywhere

https://github.com/expo/expo/blob/e1471f6505fa281c7331b19fa3d3511dfb4f157b/packages/expo-modules-core/ios/Utilities/Mutex.swift#L9-L23

packages/expo-updates/ios/EXUpdates/EnabledAppController.swift
```diff
@@ -181,20 +181,34 @@ public class EnabledAppController: InternalAppControllerInterface, UpdatesInterf
 
   // MARK: - UpdatesInterface
 
-  internal var stateChangeListeners: [String: any UpdatesStateChangeListener] = [:]
+  // Native subscribers (e.g. expo-app-metrics) subscribe from the module-registry thread while the
+  // state machine iterates this dictionary on its own queue, so every access is locked. Listeners
+  // are invoked outside the lock, off a snapshot: re-entrant subscribe/unsubscribe from inside
+  // `updatesStateDidChange` cannot deadlock, but `remove()` does not cancel a notification already
+  // in flight, so a listener must stay safe to call after it unsubscribes.
```
Kudo: could we reduce code comment be shorter and maybe clear?

packages/expo-updates/CHANGELOG.md
```diff
@@ -13,6 +13,8 @@
 ### 🐛 Bug fixes
 
 - [iOS] Set `always_out_of_date` on the `Generate updates resources for expo-updates` script_phase to silence the Xcode "run script phase will run on every build" dependency-analysis warning. ([#47622](https://github.com/expo/expo/pull/47622) by [@ramonclaudio](https://github.com/ramonclaudio))
+- [iOS] Fix an `EXC_BREAKPOINT` at launch when a native consumer reads `launchedUpdateId`, `launchAssetPath` or `launchAssetUrl()` after `initializeWithoutStarting()` publishes the controller but before `start()` assigns `startupProcedure` — these now return `nil` instead of trapping on the implicitly-unwrapped optional. ([#48898](https://github.com/expo/expo/pull/48898) by [@spsaucier](https://github.com/spsaucier))
+- [iOS] Fix an `EXC_BAD_ACCESS` in `UpdatesStateMachine.processEvent` caused by an unsynchronized `stateChangeListeners` dictionary being mutated by native subscribers while the state machine iterated it — access is now locked and listeners are notified off a snapshot. ([#48898](https://github.com/expo/expo/pull/48898) by [@spsaucier](https://github.com/spsaucier))
```
Kudo: merge the two changelog items?

packages/expo-updates/ios/EXUpdates/EnabledAppController.swift
```diff
@@ -213,16 +227,19 @@ public class EnabledAppController: InternalAppControllerInterface, UpdatesInterf
     return config.requestHeaders
   }
 
+  // `startupProcedure` is only assigned in `start()`, but the controller is published to
+  // `UpdatesControllerRegistry` by `initializeWithoutStarting()`, so a native consumer can read
+  // this first. `isStarted` is not a usable guard — it is set before the assignment.
```
Kudo: ditto, reduce code comment

Kudo (changes requested):
thanks for having the fix. leaving some comment if there's anything meaningful

packages/expo-updates/ios/EXUpdates/EnabledAppController.swift
spsaucier: Done in 21513560 — much nicer, thanks for the pointer. Fixed the generic syntax slightly (`private let` + explicit type parameter) so it compiles:

```swift
private let stateChangeListeners = Mutex<[String: any UpdatesStateChangeListener]>([:])
```

All three accessors are now one-liners: `withLock { $0[subscriptionId] = listener }`, `withLock { $0.removeValue(forKey:) }`, and `withLock { Array($0.values) }`. `ExpoModulesCore` was already imported in this file, so no new dependency.

packages/expo-updates/ios/EXUpdates/EnabledAppController.swift
spsaucier: Trimmed to two lines in 21513560:

```swift
// Subscribed from the module-registry thread, iterated on the state machine's queue. Listeners are
// notified off a snapshot, so one must stay safe to call after it unsubscribes.
```

Kept the snapshot caveat because it's the one behavioral consequence a reader can't infer from the code — `remove()` no longer cancels a notification already in flight. Dropped the rest as restating the diff.

packages/expo-updates/ios/EXUpdates/EnabledAppController.swift
spsaucier: Trimmed to two lines in 21513560:

```swift
// The controller is registered before `start()` assigns `startupProcedure`, so a native consumer
// can read this first. `isStarted` can't guard it — that is set before the assignment.
```

Kept the `isStarted` half deliberately — it's the obvious-looking guard, and it doesn't work.

packages/expo-updates/CHANGELOG.md
spsaucier: Merged into one entry in 21513560 — they share a root cause and a PR link, so two entries was noise.

packages/expo-updates/ios/EXUpdates/EnabledAppController.swift
spsaucier: I'd like to keep the conservative fix here, but I dug into your suggestion and I think it's right as the end state — with one wrinkle that I don't think you can skip.

**It would work on the read path.** `StartupProcedure.launcher` is already `AppLauncher?`, and both accessors optional-chain it:

```swift
internal func launchedUpdate() -> Update? { return launcher?.launchedUpdate }
internal func launchAssetUrl() -> URL? { return launcher?.launchAssetUrl }
```

So an early-constructed procedure returns `nil` from exactly the three reads that trap today — same observable behavior as this PR, minus the IUO. `assetFilesMap()` and `isUsingEmbeddedAssets()` are nil-safe too, so the internal uses in `getNativeStateMachineContext` are fine.

**The wrinkle: `config` is mutable after `init`.** `setUpdateURLAndRequestHeadersOverride` and `setUpdateRequestHeadersOverride` both reassign `self.config` (`EnabledAppController.swift:351,362`), and `selectionPolicy` is a computed property derived from it. Today `start()` reads both at start time, so an override applied between `init` and `start` is picked up. Constructing the procedure in `init` would freeze the config and selection policy at init time and silently drop that — which matters most for the anti-bricking / URL-override path, i.e. the case you'd least want to regress quietly.

That's solvable (construct lazily, or rebuild on override), but it turns a mechanical change into one with a real behavioral question attached, and I can't compile or run iOS for this repo locally to check I haven't broken the override flow. I'd rather not ship an unverified lifecycle change on the launch path when the current diff is three characters at each of three reads and is already running in our production fleet.

**Also worth noting:** `procedureContext` and `loaderTask` inside `StartupProcedure` are themselves IUOs assigned in `run()`. Early construction makes these three accessors safe; it doesn't make the object safe pre-run generally. So the constructor move narrows the window rather than closing the class of bug outright.

Happy to open a follow-up PR for it if you want me to take it — but given the config-override interaction and that I can't verify on-device, you or someone with the iOS setup may want to own that one. Either way I'm fine with it landing after this.

packages/expo-updates/ios/EXUpdates/UpdatesStateMachine.swift
spsaucier: Confirmed — it's the same shape exactly:

```kotlin
// EnabledUpdatesController.kt:88
internal val stateChangeListenerMap: MutableMap<String, UpdatesStateChangeListener> = mutableMapOf()

// UpdatesStateMachine.kt:102-103
it.stateChangeListenerMap.keys.forEach { key ->
  it.stateChangeListenerMap[key]?.updatesStateDidChange(toMap(event))
}
```

Plain `mutableMapOf()`, iterate-keys-then-subscript, and `AppMetricsModule.kt` subscribes from `OnCreate` the same way iOS does. `ConcurrentModificationException` rather than `EXC_BAD_ACCESS`, as you say.

Worth noting it's less likely to bite than the iOS one in practice: Android already builds `startupProcedure` in the constructor, so the subscribe call isn't racing a startup assignment — but the map is still unsynchronized against the state machine's own iteration, so the window is real.

I haven't seen it in our Crashlytics, so I'd be filing it from code reading rather than from an observed crash. Happy to open a separate PR (a `ConcurrentHashMap`, or the Kotlin equivalent of the snapshot) if you want it — just say the word and I'll keep it out of this one.

packages/expo-updates/ios/EXUpdates/EnabledAppController.swift
```diff
@@ -213,16 +217,18 @@ public class EnabledAppController: InternalAppControllerInterface, UpdatesInterf
     return config.requestHeaders
   }
 
+  // The controller is registered before `start()` assigns `startupProcedure`, so a native consumer
+  // can read this first. `isStarted` can't guard it — that is set before the assignment.
   public var launchedUpdateId: UUID? {
-    return startupProcedure.launchedUpdate()?.updateId
+    return startupProcedure?.launchedUpdate()?.updateId
```
Kudo, line 228: [🏗️ **design**] The optional chain stops the nil trap, but the read is still unsynchronized. `startupProcedure` is a plain `var` written on the main thread in `start()`, and `StartupProcedure.launcher` (what `launchedUpdate()` reads) is a plain `var` written from the controller queue during launch (`StartupProcedure.swift:48`, `:103`, `:154`).

The caller in your report reads it off the main thread: `UpdatesMonitoring.getUpdatesMetricsInfo()` is called from `patchAppInfoIfNeeded()`, which is `@AppMetricsActor`-isolated, so it runs on the cooperative pool while startup is still assigning `launcher`.

Since this PR already brings `Mutex` into the file, is it worth putting `startupProcedure` behind one too? Otherwise the same class of `EXC_BAD_ACCESS` stays reachable from the same caller, just at a lower rate.

Kudo (commented):
additional review comment that feels reasonable. would be good if you could take the other check

packages/expo-updates/ios/EXUpdates/EnabledAppController.swift
spsaucier: Good catch, and you're right that the optional chain only bought a nil check. Done in 02e81943 — but I had to take it one level deeper than you suggested, because `startupProcedure` alone wouldn't have closed the path you described.

**Why the reference alone isn't enough.** `launchedUpdateId` reads `startupProcedure?.launchedUpdate()`, and `launchedUpdate()` returns `launcher?.launchedUpdate`. Synchronizing `startupProcedure` makes the *first* hop safe and leaves the second exactly as it was. So I put both behind `Mutex`:

- `EnabledAppController.startupProcedure` — `Mutex<StartupProcedure?>` behind a computed property, so the ~14 call sites keep reading naturally.
- `StartupProcedure.launcher` — same shape, `private(set)` preserved. No readers outside that file, so it's contained.

**`launcher` is the one that can actually fault.** `startupProcedure` is assigned exactly once, nil → object, never reassigned or nilled, so a racing reader sees either nil or a live object; it's UB under exclusivity but there's nothing to deallocate. `launcher` is written three times during launch — `setLauncher` (`:48`), `emergencyLaunch` (`:103`), and `appLoaderTask(_:didFinishWithLauncher:)` (`:154`) — so a reader mid-retain can touch a launcher that the previous assignment just released. That's the real use-after-free, and it's precisely what your `@AppMetricsActor` caller reaches.

`start()` now builds the procedure into a local, wires the delegate, then publishes it — one store instead of read-modify-write through the property. `getConstantsForModule` needed `?? true` on `isUsingEmbeddedAssets`, matching the default `StartupProcedure` already returns for a nil launcher. The `swiftlint:disable implicitly_unwrapped_optional` pair is gone with the IUO.

**What I verified and what I didn't.** I can't build this repo's iOS targets locally, and the PR's only check is `codemention`, so nothing here compiles the Swift. I did what I could: `swiftc -parse` on both files is clean, and I typechecked the pattern standalone — including `internal private(set)` on a computed property and the optional-chain flattening through `launchedUpdate()` — against Swift 6.2.4. That's syntax plus the language questions I was least sure of, not a real build. Worth a compile on your side before merging.

One thing I left alone: `procedureContext` and `loaderTask` inside `StartupProcedure` are still IUOs assigned in `run()`. Nothing on the `UpdatesInterface` path touches them, so I kept them out of scope rather than widen this further.

Also — "take the other check" — did you mean this synchronization one, or the constructor-move suggestion from the earlier thread? I pushed back on that one over `config` being reassigned after `init` by the override setters, but happy to revisit if that's what you meant.

Kudo (approved):
thanks for helping this contribution!

PR #48903, opened 2026-08-13
[ui][ios] Support 'auto' on every padding modifier edge

Den1Marshall:
# Why

`padding()` without arguments applies SwiftUI's system default padding to all
edges, but specifying any edge dropped the rest to zero. There was no way to
express "system default on top, 24 points on the sides" — SwiftUI writes that
as `.padding(.top).padding(.horizontal, 24)`.

# How

Each edge of the modifier now accepts `number | 'auto'`. Values are decoded by
a `PaddingValue` convertible (`ios/Convertibles/PaddingValue.swift`) with
`.auto` and `.points` cases; anything that is neither a number nor `'auto'`
throws with a message naming the accepted values.

`EdgeInsets` can only carry lengths, so edges resolved to `auto` are collected
into an `Edge.Set` and applied by a second `padding(_:)` call — padding
modifiers add up and the two calls cover disjoint edges. The system default has
no fixed length (it adapts to size class and Dynamic Type), so it cannot be
resolved to a number on either side.

Existing behaviour is unchanged: a specific edge still overrides the shorthand
covering it, unspecified edges still get no padding, and `padding()` with no
parameters still pads every edge by the system default.

# Test Plan

Added a "Padding" section to the modifiers screen in native-component-list
covering `padding()`, `padding({ top: 'auto', horizontal: 24 })` and
`padding({ all: 'auto', leading: 0 })`, and verified it in bare-expo on the iOS
simulator: the auto edges match the system default of the plain `padding()` row,
while the explicit edges keep their exact lengths.

`et check-packages @expo/ui` passes and the app builds (`BUILD SUCCEEDED`).

# Checklist


brentvatne:
@expo-bot verify

brentvatne:
@expo-bot review

packages/expo-ui/ios/Modifiers/ViewModifierRegistry.swift
```diff
@@ -83,31 +83,71 @@ internal struct FrameModifier: ViewModifier, Record {
...
-      let insets = EdgeInsets(
-        top: top ?? vertical ?? all ?? 0,
-        leading: leading ?? horizontal ?? all ?? 0,
-        bottom: bottom ?? vertical ?? all ?? 0,
-        trailing: trailing ?? horizontal ?? all ?? 0
-      )
-      content.padding(insets)
+      // `EdgeInsets` can only carry lengths, so the edges left to the system are applied by a
+      // second call. Padding modifiers add up and these two cover disjoint edges.
+      content.padding(insets).padding(autoEdges)
```
intergalacticspacehighway:
```suggestion
  if autoEdges.isEmpty {
    content.padding(insets)
  } else {
    content.padding(insets).padding(autoEdges)
  }
```

we can avoid adding additional modifier if auto is not present.

packages/expo-ui/ios/Convertibles/PaddingValue.swift
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ The amount of padding applied to a single edge. The system default padding has no fixed length —
+ SwiftUI resolves it at layout time — so it is kept as a separate case instead of a number.
+ */
+internal enum PaddingValue: Convertible {
+  case auto
+  case points(CGFloat)
+
+  private static let autoKeyword = "auto"
```
intergalacticspacehighway: can we rename this to "default" instead? 

two arguments on this:
1. padding: "auto" is not valid CSS, if we're taking inspiration from CSS.
2. Apple calls this default padding, so we can keep the string "default" just to match the convention 😅

intergalacticspacehighway (changes requested):

packages/expo-ui/ios/Modifiers/ViewModifierRegistry.swift
Den1Marshall: Applied, thanks. Worth noting: if/else in a ViewBuilder becomes _ConditionalContent, so a view flipping between a numeric padding and default at runtime is rebuilt rather than re-laid out and loses state like focus

packages/expo-ui/ios/Convertibles/PaddingValue.swift
Den1Marshall: Renamed to default, you're right on both counts 👍🏻 
Also renamed autoEdges/isAuto to match

intergalacticspacehighway (approved):

PR #48905, opened 2026-08-13
[ui][ios] Share a ShapeStyle vocabulary between modifiers

Den1Marshall:
# Why

Several SwiftUI modifiers paint an area with a `ShapeStyle`: `foregroundStyle`,
`background`, shape fills. Expo UI described that vocabulary once, inside
`ForegroundStyleModifier`, in a form nothing else could reuse – a flat record
plus two near-identical `switch` statements, one returning a `View` and one a
`Text`. Adding the same styles to `background` meant copying both.

# How

`ShapeStyleValue` resolves a style into a single `AnyShapeStyle`. The erasure is
what makes sharing possible: an opaque `some ShapeStyle` has to be one concrete
type across every branch, and an existential `any ShapeStyle` cannot escape into
the result of `background(_:)` or `foregroundStyle(_:)`, so a style chosen at
runtime cannot be returned any other way. `AnyShapeStyle` is available since
iOS 15, below the package's 16.4 minimum.

`foregroundStyle` now delegates to it and shrinks from 125 lines to 34. Its
public union is unchanged except for a new `material` variant, so existing calls
keep working; the payload nests under a `style` key, which is what the two
updated tests assert.

Materials cover the five thickness levels and `bar`. `bar` has no tvOS
counterpart and `quinary` needs tvOS 17; both leave the view untouched there
rather than failing, since the same JS runs on every platform and the modifier
array is rebuilt on each render.

# Test Plan

Added a "Material foreground style" section to the modifiers screen in
native-component-list: text painted with a material over a gradient-filled
rectangle. Verified in bare-expo on the iOS simulator.

`et check-packages @expo/ui` passes, the app builds (`BUILD SUCCEEDED`) and the
package tests pass (24 suites, 129 tests).

# Checklist


brentvatne:
@expo-bot verify

brentvatne:
@expo-bot review

Den1Marshall:
@intergalacticspacehighway thanks for merging #48903. This one without a reviewer for now. I've merged main into it, so the conflict is gone and it's ready to look at

intergalacticspacehighway (approved):

PR #48907, opened 2026-08-13
[docs] document expo image integration with Observe

kadikraman:
# Why

We've added an opt-in integration with `expo-image` to report oversized images but it is not documented anywhere.

# How

Add an integration page with instructions, following existing conventions.

# Test Plan

Review http://localhost:3002/eas/observe/integrations/expo-image/

# Checklist


docs/pages/eas/observe/integrations/expo-image.mdx
```diff
@@ -0,0 +1,95 @@
...
+Observe.configure({
+  integrations: { 'expo-image': true },
+});
+```
+
+No other setup is needed. Once enabled, the integration observes every image load automatically.
+
+## How it works
+
+On every image load, the integration compares the image's decoded pixel size with the number of physical pixels on the device's screen (the screen's point size multiplied by the square of the device pixel ratio). When the decoded area exceeds the screen's pixel count by more than the configured threshold, the integration logs an `expo-image.oversized` event with `warn` severity.
```
amandeepmittal: ✏️ **style** `mechanics.md#vale-rules-enforced-if-missed`

The first sentence runs 35 words and carries the comparison and the formula at once. Move the formula into its own sentence.

```suggestion
On every image load, the integration compares the image's decoded pixel size with the number of physical pixels on the device's screen. That pixel count is the screen's point size multiplied by the square of the device pixel ratio. When the decoded area exceeds the screen's pixel count by more than the configured threshold, the integration logs an `expo-image.oversized` event with `warn` severity.
```

docs/pages/eas/observe/integrations/expo-image.mdx
```diff
@@ -0,0 +1,95 @@
...
+  },
+});
+```
+
+- `oversizeThreshold`: An image is reported when its decoded pixel area exceeds the screen's physical pixel count by more than this factor. The default is `1.5`, which flags images decoded at more than 1.5 times the pixels the screen physically has: a full-screen image plus 50% headroom.
+
+## Fix oversized images
+
+- Serve images at a size close to their display size, for example by requesting resized variants from your image CDN.
+- When loading images with the [`useImage`](/versions/latest/sdk/image/) hook, set the `maxWidth` and `maxHeight` load options to downscale the image at decode time while preserving its aspect ratio.
```
amandeepmittal: ✏️ **style** `links-and-headings.md#internal-links`

This link lands the reader at the top of the Image page instead of the hook. Point it at the hook's anchor, as the Maps page does: `/versions/latest/sdk/image/#useimagesource-options-dependencies`.

```suggestion
- When loading images with the [`useImage`](/versions/latest/sdk/image/#useimagesource-options-dependencies) hook, set the `maxWidth` and `maxHeight` load options to downscale the image at decode time while preserving its aspect ratio.
```

docs/pages/eas/observe/integrations/expo-image.mdx
```diff
@@ -0,0 +1,95 @@
...
+Observe.configure({
+  integrations: {
+    'expo-image': {
+      oversizeThreshold: 2,
+    },
+  },
+});
+```
+
+- `oversizeThreshold`: An image is reported when its decoded pixel area exceeds the screen's physical pixel count by more than this factor. The default is `1.5`, which flags images decoded at more than 1.5 times the pixels the screen physically has: a full-screen image plus 50% headroom.
```
amandeepmittal: ✏️ **style** `review-ai-text.md#t5-restatement`

The bullet states the same rule three times: the factor, then the same rule with `1.5` filled in, then the headroom gloss after the colon. Keep the default and the headroom, and drop the middle restatement.

```suggestion
- `oversizeThreshold`: An image is reported when its decoded pixel area exceeds the screen's physical pixel count by more than this factor. The default is `1.5`, which leaves room for a full-screen image plus 50% headroom.
```

docs/pages/eas/observe/integrations/expo-image.mdx
```diff
@@ -0,0 +1,90 @@
...
+});
+```
+
+No other setup is needed. Once enabled, the integration observes every image load automatically.
+
+## How it works
+
+On every image load, the integration compares the image's decoded pixel size with the number of physical pixels on the device's screen. That pixel count is the screen's point size multiplied by the square of the device pixel ratio. When the decoded area exceeds the screen's pixel count by more than the configured threshold, the integration logs an `expo-image.oversized` event with `warn` severity.
+
+Each image URL is reported at most once per app session. Calling `Observe.configure()` again resets this deduplication and applies the new configuration.
```
tsapeta: [💡 suggestion] This describes reconfiguration as routine, but the two sibling integration pages carry the opposite warning:

- [`expo-router.mdx` L33](https://github.com/expo/expo/blob/b01d61887821c6dd078dfd7b86081c7cc0499d4b/docs/pages/eas/observe/integrations/expo-router.mdx#L33)
- [`react-navigation.mdx` L36](https://github.com/expo/expo/blob/b01d61887821c6dd078dfd7b86081c7cc0499d4b/docs/pages/eas/observe/integrations/react-navigation.mdx#L36)

Both say: "The integration must be enabled before mount and cannot be toggled at runtime. Calling `configure()` after the app has mounted, or toggling the flag mid-session, throws an error." That is backed by real throw sites, for example [`ObserveRouterIntegrationProvider.tsx` L19-L23](https://github.com/expo/expo/blob/b01d61887821c6dd078dfd7b86081c7cc0499d4b/packages/expo-observe/src/integrations/expo-router/ObserveRouterIntegrationProvider.tsx#L19-L23).

For `expo-image` the reset genuinely is supported: [`activate` in `packages/expo-image/src/observe.ts`](https://github.com/expo/expo/blob/b01d61887821c6dd078dfd7b86081c7cc0499d4b/packages/expo-image/src/observe.ts#L119-L141) rebuilds the `reported` set and adds/removes the native listener on each `configure` event, with no mount-time guard. So the difference is real rather than an error. But a reader coming from either page above will find the two statements contradictory. Worth an explicit note that, unlike the navigation integrations, this one can be reconfigured at runtime.

tsapeta (approved):

docs/pages/eas/observe/integrations/expo-image.mdx
tsapeta: I think we don't want users to reconfigure (it may lead to some edge cases, it's supposed to be invoked only once in the top-level scope). I wonder what would be the best way to let the integrations support reconfiguration. Maybe a new function `configureIntegration(integrationName, config)`? cc @Ubax

tsapeta (approved):

amandeepmittal (approved):

docs/pages/eas/observe/integrations/expo-image.mdx
Ubax: Honestly I'm not sure if we should even support reconfiguration. It can cause some unexpected results, and I don't see much of the use-case for it. In case like reseting or starting an integration only for a given timespan, I'd rather have libraries expose their own functions like: `resetDeduplication` or `start/stopRecording`

Ubax: And in this specific case of expo-image integration, I don't think people should even reset it. I don't see a benefit of getting this event twice

docs/pages/eas/observe/integrations/expo-image.mdx
```diff
@@ -0,0 +1,99 @@
...
+Observe.configure({
+  integrations: { 'expo-image': true },
+});
+```
+
+No other setup is needed. Once enabled, the integration observes every image load automatically.
+
+## How it works
+
+On every image load, the integration compares the image's decoded pixel size with the number of physical pixels on the device's screen. That pixel count is the screen's point size multiplied by the square of the device pixel ratio. When the decoded area exceeds the screen's pixel count by more than the configured threshold, the integration logs an `expo-image.oversized` event with `warn` severity.
```
Ubax:
```suggestion
On every image load, the integration compares the image's decoded pixel size with the number of physical pixels on the device's screen. When the decoded area exceeds the screen's pixel count by more than the configured threshold, the integration logs an `expo-image.oversized` event with `warn` severity.
```

It seems a bit overly complicated 😅 If someone wants that level of detail they can check the codebase

docs/pages/eas/observe/integrations/expo-image.mdx
```diff
@@ -0,0 +1,99 @@
...
+
+The decoded size that the integration sees differs between platforms for images rendered with the `<Image>` component:
+
+- On Android, the `<Image>` component downscales images to the component's size at decode time, so a large source displayed in a small component is usually not reported.
+- On iOS, the integration reports the source's decoded size before `expo-image` downscales it for rendering, so a large source displayed in a small component is still reported, even with `allowDownscaling` enabled.
+
+Images loaded with the `useImage` hook or `Image.loadAsync` decode at the source's full size on both platforms unless you set the `maxWidth` and `maxHeight` load options, so they are reported consistently.
```
Ubax:
```suggestion
Images loaded with the `useImage` hook or `Image.loadAsync` decode at the source's full size on both platforms by default. You can limit the decoded size with the `maxWidth` and `maxHeight` load options. This makes the reported image dimensions consistent across platforms.
```

For me, in the first sentence it was not obvious what would happen if I set `maxWidth`/`maxHeight`

Ubax (approved):

docs/pages/eas/observe/integrations/expo-image.mdx
kadikraman: I removed the suggestion that you might want to call configure again.

docs/pages/eas/observe/integrations/expo-image.mdx
```diff
@@ -0,0 +1,107 @@
...
+
+The decoded size that the integration sees differs between platforms for images rendered with the `<Image>` component:
+
+- On Android, the `<Image>` component downscales images to the component's size at decode time, so a large source displayed in a small component is usually not reported.
+- On iOS, the integration reports the source's decoded size before `expo-image` downscales it for rendering, so a large source displayed in a small component is still reported, even with `allowDownscaling` enabled.
+
+Images loaded with the `useImage` hook or `Image.loadAsync` decode at the source's full size on both platforms by default. You can limit the decoded size with the `maxWidth` and `maxHeight` load options. This makes the reported image dimensions consistent across platforms.
+
+Each image URL is reported at most once per app session. Deduplication uses the sanitized URL, so with the default configuration, variants of one image that differ only in their query parameters (such as rotating signed URLs) produce a single event.
```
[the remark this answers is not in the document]
kadikraman: Users should only call `configure()` once.

PR #48909, opened 2026-08-13
[age-range] Add fake age signals for testing the Android age gate

kmadden84:
# Why

Google Play only reports age signals to accounts it has enabled, and those accounts cannot be created elsewhere. So an app cannot test its age gate against any other age range, source, or error.

Google provides `FakeAgeSignalsManager` for this, and the package already uses it: `AgeSignalsManagerTest` passes it to both requests, since both take the manager as a parameter. This makes it available to apps too.

# How

`setFakeAgeSignals(fake)` makes `requestAgeRangeAsync` and `requestAgeSignalsAccessAsync` report `fake` through `FakeAgeSignalsManager`. `null` restores the real signals.

- `fake` is either a response or an error, never both. The response fields match `AgeRangeResponse`, plus the `ageSignalsStatus` of the access request.
- The values JS sends are `Enumerable` enums, and `ageRangeSourceToString` and its siblings now read the same enums, so each value is defined once.
- A fresh `FakeAgeSignalsManager` per request, since `setNext*` configures a single response. One set of fake signals answers every request until the app replaces or clears them.
- No new dependency: `FakeAgeSignalsManager` ships inside `com.google.android.play:age-signals`, not a separate testing artifact.
- No build-time gate. Google ships the fake manager in the runtime artifact without one, and the config plugin that added one was dropped in review.

The only change to existing behaviour is which manager the two requests read, and that is the real one until an app calls `setFakeAgeSignals`.

# Test Plan

- `FakeAgeSignalsTest` covers the enum mapping arriving back as a faked `requestAgeRange` response, and faking an error and a response at once being rejected.
- `ExpoAgeRange-test` covers the JS surface on the platforms without it.
- We have been running the equivalent as a `patch-package` patch against `expo-age-range@56.0.6` in the Tim Hortons app, which is how we know the approach works on a device. That patch predates `0.0.4`, so it fakes the older `userStatus` field and is not a straight port of this diff.

# Checklist

- [x] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: updated a module plugin). Nothing in a built app changes unless it calls `setFakeAgeSignals`.


brentvatne:
@expo-bot review

packages/expo-age-range/android/src/main/java/expo/modules/agerange/AgeRangeModule.kt
```diff
@@ -38,13 +48,24 @@ class AgeRangeModule : Module() {
...
-        ageSignalsManager = ageSignalsManager,
+        ageSignalsManager = currentAgeSignalsManager,
         activity = appContext.throwingActivity,
         onSuccess = { status -> promise.resolve(status) },
         onError = { exception -> promise.reject(exception) },
         onCancelled = { promise.reject(AgeRangeTaskCancelledException()) }
       )
     }
+
+    AsyncFunction("isFakeAgeSignalsEnabledAsync") {
```
vonovak: can be a sync function

packages/expo-age-range/android/src/main/java/expo/modules/agerange/AgeRangeModule.kt
```diff
@@ -38,13 +48,24 @@ class AgeRangeModule : Module() {
...
+      isFakeAgeSignalsEnabled(context)
+    }
+
+    AsyncFunction("setFakeAgeSignalsAsync") { options: FakeAgeSignalsOptions? ->
```
vonovak: can be a sync function

packages/expo-age-range/android/src/main/java/expo/modules/agerange/FakeAgeSignals.kt
```diff
@@ -0,0 +1,149 @@
...
+      setNextRequestAgeSignalsAccessException(exception)
+      return@apply
+    }
+
+    setNextAgeSignalsResult(ageSignalsResult)
+    setNextAgeSignalsAccessResult(ageSignalsAccessResult)
+  }
+}
+
+internal fun ageRangeSourceFromString(source: String?): Int? = fakeAgeSignalsValue(
```
vonovak: we now have functions to covert both ways; maybe https://docs.expo.dev/modules/module-api/#enums would be a better fit for this

packages/expo-age-range/android/src/test/java/expo/modules/agerange/FakeAgeSignalsTest.kt
```diff
@@ -0,0 +1,177 @@
...
+import org.junit.Assert.assertTrue
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.Robolectric
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.RuntimeEnvironment
+import org.robolectric.Shadows.shadowOf
+
+@RunWith(RobolectricTestRunner::class)
+class FakeAgeSignalsTest {
```
vonovak, line 14: afaict this is testing very little of our own logic, so it can be drastically reduced or even removed

packages/expo-age-range/src/AgeRange.ts
```diff
@@ -129,3 +130,61 @@ export async function requestAgeSignalsAccessAsync(): Promise<AgeSignalsStatus |
...
+}
+
+/**
+ * Makes [`requestAgeRangeAsync`](#agerangerequestagerangeasyncoptions) and
+ * [`requestAgeSignalsAccessAsync`](#agerangerequestagesignalsaccessasync) report `fake` instead of
+ * the signals Play reports, using Play's own
+ * [`FakeAgeSignalsManager`](https://developer.android.com/google/play/age-signals/test-age-signals-api).
+ * Pass `null` to go back to the real signals. Resolves without doing anything on iOS and web.
+ *
+ * Play only reports live age signals to accounts in the cohorts it has enabled, which cannot be
```
vonovak: use full name of the SDK, not just "Play"

also, the sentence reads weird to me but maybe that's just because I'm not a native speaker... it sound like Opus 5.0 to me 😆 . Keep it short and simple pls.

packages/expo-age-range/src/ExpoAgeRange.types.ts
```diff
@@ -97,6 +97,37 @@ export type AgeRangeResponse = {
...
+  significantChangeApprovalDate?: number | null;
+  /** Reported by [`requestAgeSignalsAccessAsync`](#agerangerequestagesignalsaccessasync). */
+  ageSignalsStatus?: AgeSignalsStatus | null;
+  /**
+   * When set, both `requestAgeRangeAsync` and `requestAgeSignalsAccessAsync` reject with this Play
+   * Age Signals error code instead of reporting a result. See the
+   * [error code reference](https://developer.android.com/google/play/age-signals/handle-errors)
+   * for the available codes.
+   */
+  errorCode?: number | null;
```
vonovak: let's use discriminating unions. Either error, or a fake response. not both should be accepted.

vonovak (changes requested):
thanks for the PR; I added some comments and let's drop the config plugin altogether. Thank you!

packages/expo-age-range/android/src/main/java/expo/modules/agerange/AgeRangeModule.kt
kmadden84: Removed entirely — with the config plugin gone there is no opt-in left to report.

packages/expo-age-range/android/src/main/java/expo/modules/agerange/AgeRangeModule.kt
kmadden84: Done — it is now `Function("setFakeAgeSignals")`, and the JS wrapper is sync too.

packages/expo-age-range/android/src/main/java/expo/modules/agerange/FakeAgeSignals.kt
kmadden84: Done. `AgeRangeSourceValue`, `SignificantChangeStatusValue`, and `AgeSignalsStatusValue` in the new `AgeRangeEnums.kt` are `Enumerable`s, so the module converts the JS strings and reports unknown ones itself — the three `*FromString` functions and `InvalidFakeAgeSignalsException` are gone.

`ageRangeSourceToString` and its siblings now read the same enums through one `playValueToString` helper, so each value is defined in one place and the `UNSPECIFIED` and unknown-value fallbacks behave exactly as before.

The enums deliberately take no constructor parameter: `EnumTypeConverter` maps by name only when the enum has no instance field, and `playValue` is a getter.

packages/expo-age-range/android/src/test/java/expo/modules/agerange/FakeAgeSignalsTest.kt
kmadden84: Agreed — down from 177 lines to two tests: a faked age range arriving through `requestAgeRange` with the enum mapped to the string JS receives, and faking an error and a response at once being rejected. Everything else was really testing `FakeAgeSignalsManager`.

packages/expo-age-range/src/AgeRange.ts
kmadden84: Fixed — "Google Play" and "Google Play Age Signals" are spelled out, and the JSDoc is two short paragraphs now. The docs section got the same treatment.

packages/expo-age-range/src/ExpoAgeRange.types.ts
kmadden84: Done. `FakeAgeSignals` is a union of the response, with `errorCode?: never`, and `{ errorCode: number }`, with the response fields `never`, so `{ errorCode: -4, lowerBound: 18 }` no longer typechecks. Kotlin throws `ERR_AGE_RANGE_FAKE_SIGNALS_CONFLICT` for the same mistake from untyped JS.

kmadden84:
Pushed. The config plugin is gone, along with the manifest flag and `isFakeAgeSignalsEnabledAsync`, so `setFakeAgeSignals` now works in any build. The diff is down from 20 files to 14, and `package.json`, `tsconfig.json`, and the plugin scaffolding are untouched again.

One thing that is your call: dropping the plugin also drops the only thing that kept fake signals out of a production build, which the bot review flagged. I went with matching Google here, since `FakeAgeSignalsManager` ships in the runtime artifact with no gate of its own, and the docs and JSDoc say not to call this in production. If you would rather it be impossible rather than discouraged, an `ApplicationInfo.FLAG_DEBUGGABLE` check is a few lines and needs no configuration — say the word and I will add it.

vonovak (approved):
thank you@

PR #48912, opened 2026-08-14
[ios][screen-orientation] Resolve nested RNS orientation owners

IkeStudios:
## Why

`ScreenOrientationViewController` currently asks react-native-screens about itself and then one level of child view controllers. That fixed the known `expo-dev-client` hierarchy in #44181, but orientation falls back to the Expo registry whenever another UIKit container is inserted between that child and the active RNS navigation controller.

The lookup should not depend on a fixed containment depth. At the same time, recursively scanning every retained child is unsafe: tab, page, navigation, and custom containers can keep inactive controllers whose screen orientation must not override the visible route.

## Reproduction

A standalone neutral reproduction and config plugin are available here:

https://github.com/IkeStudios/expo-screen-orientation-nested-repro

| Before — stock one-level lookup | After — this PR's lookup |
| --- | --- |
| Stock behavior: landscape route remains portrait | Patched behavior: route rotates to landscape and restores portrait |
| [Full-resolution MP4](https://github.com/IkeStudios/expo-screen-orientation-nested-repro/blob/main/recordings/before.mp4) | [Full-resolution MP4](https://github.com/IkeStudios/expo-screen-orientation-nested-repro/blob/main/recordings/after.mp4) |

Inline previews play at 2x speed.

The fixture inserts two transparent UIKit containment wrappers around the React controller. With stock `expo-screen-orientation@55.0.18`, tapping the landscape route changes the rendered route but leaves the interface at `420 x 912` portrait. Applying this PR's lookup change rotates the same route to `912 x 420` landscape-right and restores `420 x 912` when returning.

[Expo's hosted verification](https://github.com/expo/expo/pull/48912#issuecomment-5289627666) independently reproduced the same failure on SDK 57 and also found a non-synthetic case in `expo-router/unstable-native-tabs`: stock remained `402 x 874` while the earlier PR revision rotated the declared routes to `874 x 402`. That hosted run tested revision `05dbb2ac`; the current revision retains the verified containment walk while narrowing presented-controller and inactive-child handling.

## How

- Preserve the existing self-first fast path.
- Walk the active containment hierarchy until the RNS orientation owner is found.
- Use each standard UIKit container's active-child API (`top`, `selected`, or current pages) without following presented non-children.
- Follow only attached children for unknown multi-child wrappers, while still allowing a single wrapper child before view attachment.
- Trust `RNSScreenWindowTraits` only when the child it inspects is one of the container's resolved active children, then ask that confirmed RNS container directly.
- Let react-native-screens perform its own modal-aware orientation resolution from the active contained route.

This retains the runtime/optional integration with react-native-screens and does not import it as a dependency.

## Test plan

Automated/static:

- `xcrun swiftc -parse packages/expo-screen-orientation/ios/ScreenOrientationViewController.swift`
- `git diff --check`

Simulator reproduction through the linked fixture:

1. Launch the portrait route (`420 x 912`).
2. Tap **Open landscape route**.
3. Stock lookup: route renders but remains `420 x 912`.
4. PR lookup: route rotates to `912 x 420` landscape-right.
5. Tap **Return to portrait** and verify `420 x 912` is restored.

The traversal was also reviewed against retained inactive tabs, page/split containers, detached replacement children, and navigation controllers with presented modals so the recursive lookup stays within containment without trading the depth bug for an inactive-controller bug.

## Changelog

- [iOS] Fixed per-screen orientation lookup through deeply nested view-controller containers.

### DIFF
--- a/packages/expo-screen-orientation/CHANGELOG.md
+++ b/packages/expo-screen-orientation/CHANGELOG.md
@@ -9,7 +9,7 @@
 ### 🐛 Bug fixes
 
 - [iOS] Added a development warning when the system refuses an orientation lock request, and switched to reading interface orientation from the scene's effective geometry. ([#48173](https://github.com/expo/expo/pull/48173) by [@alanjhughes](https://github.com/alanjhughes))
-- [iOS] Fixed per-screen orientation lookup through deeply nested view-controller containers.
+- [iOS] Fixed per-screen orientation lookup through deeply nested view-controller containers. ([#48912](https://github.com/expo/expo/pull/48912) by [@IkeStudios](https://github.com/IkeStudios))
 
 ### 💡 Others
 
--- a/packages/expo-screen-orientation/CHANGELOG.md
+++ b/packages/expo-screen-orientation/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### 🐛 Bug fixes
 
 - [iOS] Added a development warning when the system refuses an orientation lock request, and switched to reading interface orientation from the scene's effective geometry. ([#48173](https://github.com/expo/expo/pull/48173) by [@alanjhughes](https://github.com/alanjhughes))
+- [iOS] Fixed per-screen orientation lookup through deeply nested view-controller containers.
 
 ### 💡 Others
 
--- a/packages/expo-screen-orientation/ios/ScreenOrientationViewController.swift
+++ b/packages/expo-screen-orientation/ios/ScreenOrientationViewController.swift
@@ -100,20 +100,74 @@ class ScreenOrientationViewController: UIViewController {
   }
 
   /// Finds the VC whose subtree has a react-native-screens screen with orientation set.
-  /// Checks self first (the common case), then each child VC. The child-level search handles
-  /// cases where an intermediate VC (e.g. DevLauncherViewController from expo-dev-client)
-  /// sits between this root VC and the RNSNavigationController, blocking the single-level
-  /// traversal that RNScreens' shouldAskScreensForScreenOrientation performs.
+  /// Checks self first (the common case), then walks the containment hierarchy. This handles
+  /// wrappers such as DevLauncherViewController without depending on their exact nesting depth.
   private func vcWithRNScreenOrientation() -> UIViewController? {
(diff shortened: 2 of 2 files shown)


brentvatne:
@expo-bot check

IkeStudios:
Updated in d6f98cf6c7. The warning correctly identified that `UINavigationController.visibleViewController` can leave the containment tree by returning a presented controller.

The revised lookup now:

- follows `topViewController` only for navigation controllers;
- trusts the RNS predicate only when the child it inspects is in the container’s resolved active children;
- asks that confirmed RNS container directly, which also covers `RNSViewController` orientation forwarding;
- ignores detached retained children once an unknown or split container is on-screen, while still allowing an unambiguous single child before attachment.

This keeps Expo’s traversal within UIKit containment and leaves modal-aware orientation precedence to react-native-screens’ existing `includingModals` behavior. Swift parsing and `git diff --check` pass. The existing reproduction remains applicable because it exercises the nested-containment path; this correction only removes the unrelated presented-controller path.

IkeStudios:
@expo-bot check

IkeStudios:
Follow-up on this hosted verification: the run tested the now-stale revision `05dbb2ac`, as its environment section notes. The current head is `d6f98cf6c7`, updated after the subsequent AI review identified that `UINavigationController.visibleViewController` can leave the containment tree.

The latest commit preserves the native-tabs path this run verified: `RNSTabBarController` resolves `selectedViewController`, the walk continues through `RNSTabsScreenViewController`, and the contained `RNSNavigationController` supplies the active screen orientation. It now returns that confirmed RNS container directly rather than relying on a generic parent to forward the mask. The revision also keeps presented-controller traversal out of Expo, gates retained/detached branches, preserves RNScreens’ own modal precedence, and handles pre-attachment single-child wrappers and split containers.

The current head passed Swift parsing, `git diff --check`, and two consecutive independent clean reviews on the unchanged code. So the latest commit is review-stable and should retain the verified `unstable-native-tabs` fix. A new hosted run against `d6f98cf6c7` would still be needed to call that exact SHA behaviorally verified; the risks and conclusions in this comment were produced from `05dbb2ac`.

brentvatne:
@expo-bot verify

PR #48913, opened 2026-08-14
[audio][ios] Session-queue follow-ups: interruption race, shared-queue block, activation-error surfacing

### LINKED ISSUE #48531: [expo-audio] iOS: sync `play()` runs blocking `AVAudioSession.setActive(true)` on the JS thread – an audio-daemon stall freezes all JS execution

sbs44:
# Why

Three follow-ups to #48588 (which fixed #48531... thanks for the quick turnaround). Also flagged in a comment there.

The main one: resumeInterruptedPlayers moved onto sessionQueue, but handleInterruptionBegan still mutates interruptedPlayers/playerVolumes on the notification thread. Two threads over the same Swift collections, and the stall scenario #48588 exists for is exactly what widens the window... back-to-back interruptions will eventually crash.

Also: setIsAudioActive does sessionQueue.sync from the AsyncFunction queue, which is one serial queue shared by every function that doesn't set runOnQueue (the vast majority, across all Expo modules), so a stalled setActive blocks those app-wide. And activation failures during play() only reach os_log now that play() can't throw for them... playback continues on an inactive session and JS has no way to know.

# How

- handleInterruptionBegan snapshots isPlaying and pauses/ducks on the notification thread, same timing as before #48588, then dispatches only the writes to the two collections (and the sessionIsActive flag) onto sessionQueue. That's all the race fix needs, since the resume path reads them there.

- setIsAudioActiveAsync becomes promise-based and settles from sessionQueue. Throwing contract preserved as a rejection, JS API unchanged. One nuance: the promise now settles with null where it settled undefined (the TS type is Promise<void>, so only a literal === undefined comparison would notice).

- applySessionActive returns the error, activateSession takes an onError callback, and AudioPlayer's play() reports the failure through the playbackStatusUpdate event's existing `error` field. Playlists are left alone: AudioPlaylistStatus has no `error` field to report through (they also emit playlistStatusUpdate, not playbackStatusUpdate)... happy to add a typed field there if you want it covered too.

# Test Plan

Swift + changelog only, no JS or committed build/ changes... lint and native tests run in CI. For the race: with the repro from #48531 (https://github.com/sbs44/expo-audio-test) plus its simulated-stall patch, trigger two interruptions in quick succession during back-to-back playback.

Before, the queued resume iterates the collections while the second interruption's handler clears them; after, both run on sessionQueue. For the error surfacing: hold the session from another app under .doNotMix and watch playbackStatusUpdate.error.

# Checklist

### DIFF
--- a/packages/expo-audio/ios/AudioModule.swift
+++ b/packages/expo-audio/ios/AudioModule.swift
@@ -36,8 +36,8 @@ public class AudioModule: Module {
       try setAudioMode(mode: mode)
     }
 
-    AsyncFunction("setIsAudioActiveAsync") { (isActive: Bool) in
-      try setIsAudioActive(isActive)
+    AsyncFunction("setIsAudioActiveAsync") { (isActive: Bool, promise: Promise) in
+      setIsAudioActive(isActive, promise: promise)
     }
 
     AsyncFunction("requestRecordingPermissionsAsync") { (promise: Promise) in
@@ -220,7 +220,9 @@ public class AudioModule: Module {
         }
         let rate = player.currentRate > 0 ? player.currentRate : 1.0
         player.play(at: rate)
-        self.activateSession()
+        self.activateSession { message in
+          player.updateStatus(with: ["error": message])
+        }
       }
 
       Function("setPlaybackRate") { (player, rate: Double, pitchCorrectionQuality: PitchCorrectionQuality?) in
@@ -649,15 +651,18 @@ public class AudioModule: Module {
   }
 
   private func handleInterruptionBegan() {
-    interruptedPlayers.removeAll()
-    playerVolumes.removeAll()
+    // Capture which players are playing before dispatching work to the `sessionQueue`.
+    // If a player is auto-paused while the `sessionQueue` is busy, a later check would
+    // miss it and it would never resume.
+    var interrupted = Set<String>()
+    var volumes = [String: Float]()
 
     registry.allPlayables.forEach { playable in
       if playable.isPlaying {
-        interruptedPlayers.insert(playable.id)
+        interrupted.insert(playable.id)
         switch interruptionMode {
         case .duckOthers:
-          playerVolumes[playable.id] = playable.volume
+          volumes[playable.id] = playable.volume
           playable.volume *= 0.5
         case .doNotMix, .doNotMixPersistent, .mixWithOthers:
           playable.pause()
@@ -673,12 +678,19 @@ public class AudioModule: Module {
     }
 #endif
 
-    recordSessionActive(false)
+    sessionQueue.async { [weak self] in
+      guard let self else {
+        return
+      }
+      self.interruptedPlayers = interrupted
+      self.playerVolumes = volumes
+      self.sessionIsActive = false
+    }
   }
 
   private func handleInterruptionEnded(with options: AVAudioSession.InterruptionOptions) {
     sessionQueue.async { [weak self] in
-      guard let self, self.audioEnabled.withLock({ $0 }), self.applySessionActive(true) else {
+      guard let self, self.audioEnabled.withLock({ $0 }), self.applySessionActive(true) == nil else {
         return
       }
       if options.contains(.shouldResume) {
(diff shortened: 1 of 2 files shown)


brentvatne:
@expo-bot check

sbs44:
Valid catch from bot reviews on the deferred isPlaying read... updated to the shape both suggested: the snapshot and the pause/duck stay on the notification thread (same timing as before #48588, so no os_log measurement needed), and only the writes to the two collections hop to sessionQueue, which is all the race fix needs since the resume path reads them there.

Also split the changelog into one bullet per change and used [weak self] on the new hop. Left the strong capture in the setIsAudioActive block on purpose: a weak self there could leave the promise unsettled if the module goes away mid-transition.

brentvatne:
@expo-bot check

packages/expo-audio/ios/AudioModule.swift
```diff
@@ -812,19 +835,24 @@ public class AudioModule: Module {
...
     if !isActive {
       pauseAllPlayers()
     }
 
-    do {
-      try sessionQueue.sync {
+    // Async rather than sessionQueue.sync: AsyncFunctions share one
+    // process-wide serial queue across every Expo module in the app, so a
+    // stalled setActive here blocked all of them. The promise settles from
+    // sessionQueue instead, preserving the throwing contract as a rejection.
```
behenate: Let's remove this comment, it will be confusing when we merge the PR and there is no reference to the previous code

packages/expo-audio/ios/AudioModule.swift
```diff
@@ -649,15 +658,22 @@ public class AudioModule: Module {
...
   private func handleInterruptionBegan() {
-    interruptedPlayers.removeAll()
-    playerVolumes.removeAll()
+    // Snapshot on the notification thread: isPlaying is live AVPlayer state
+    // that the system may flip on its own once the session is interrupted,
+    // so sampling it after an unbounded queue delay could see already-stopped
+    // players and resume nothing. Pausing/ducking here also keeps the
+    // pre-#48588 timing. Only the bookkeeping WRITES hop to sessionQueue,
+    // which is all the race fix needs: resumeInterruptedPlayers reads these
+    // collections there.
```
behenate: This comment is too verbose.
```suggestion
// Capture which players are playing before dispatching work to the `sessionQueue`. If the a player is auto-paused while the `sessionQueue` is busy, a later check would miss it and it would never resume.
```

packages/expo-audio/ios/AudioModule.swift
```diff
@@ -220,7 +220,13 @@ public class AudioModule: Module {
         }
         let rate = player.currentRate > 0 ? player.currentRate : 1.0
         player.play(at: rate)
-        self.activateSession()
+        self.activateSession { message in
+          // play() cannot throw for activation failures anymore (the
+          // transition is async on sessionQueue); surface them through the
+          // status event's existing `error` field so JS isn't left blind
+          // when playback proceeds against an inactive session.
```
behenate:
```suggestion

```

packages/expo-audio/ios/AudioModule.swift
```diff
@@ -375,6 +381,9 @@ public class AudioModule: Module {
           return
         }
         playlist.play(at: playlist.currentRate)
+        // No onError here: AudioPlaylistStatus has no `error` field (and
+        // playlists emit playlistStatusUpdate, not playbackStatusUpdate), so
+        // there is no typed surface to report through yet.
```
behenate:
```suggestion
```

behenate (changes requested):
A few nits about on the comments

brentvatne:
@expo-bot check with fable

sbs44:
Applied all four (fixed a small typo in the suggested comment). Also rebased onto main for #49101... setIsAudioActive now uses activationOptions(isActive:) so doNotMixPersistent behaves.

packages/expo-audio/ios/AudioModule.swift
sbs44: Done

packages/expo-audio/ios/AudioModule.swift
sbs44: Done with one small change: I used "If a player is auto-paused..." instead of "If the a player is auto-paused"

packages/expo-audio/ios/AudioModule.swift
sbs44: Done

packages/expo-audio/ios/AudioModule.swift
sbs44: Done

brentvatne:
@expo-bot check with fable

sbs44:
PR description updated

Issue #42939, opened 2026-02-06
Call to function 'VideoPlayer.constructor' has been rejected. → Caused by: android.app.BackgroundServiceStartNotAllowedException: Not allowed to start service Intent


danylo-musiiaka:
### Minimal reproducible example

Platform: Android

### Steps to reproduce

I've noticed this error message in my Sentry, my code is

`export default function Intro() {
  const navigation = useNavigation<NavigationProp<AppStackParamList>>();
  const isFocused = useIsFocused();

  const [isLoading, setIsLoading] = useState(true);
  const [isMuted, setIsMuted] = useState(true);

  const player = useVideoPlayer(INTRO_VIDEO_URL, (player) => {
    player.loop = false;
    player.muted = isMuted;
    player.staysActiveInBackground = false;
  });

  useEffect(() => {
    if (isFocused) {
      player.seekBy(0 - player.currentTime);
      player.play();
    } else {
      player.pause();
    }
  }, [isFocused, player]);

  useEffect(() => {
    player.muted = isMuted;
  }, [isMuted, player]);

  useEffect(() => {
    const timeSubscription = player.addListener("statusChange", ({ status }) => {
      if (status === "readyToPlay") {
        setIsLoading(false);
      } else if (status === "loading") {
        setIsLoading(true);
      }
    });

    return () => {
      timeSubscription.remove();
    };
  }, [player]);

  useEffect(() => {
    const volumeListener = SystemSetting.addVolumeListener(({ value }) => {
      if (value > 0 && isMuted) {
        setIsMuted(false);
      }
    });

    return () => {
      SystemSetting.removeVolumeListener(volumeListener);
    };
  }, [isMuted]);

  return (
    <View style={{ flex: 1, alignItems: "center" }}>
      <View style={{ alignItems: "center" }}>
        <Text style={{ color: "white", fontWeight: "500", fontSize: 18, marginTop: spacing.xl }}>Welcome to</Text>
        <Logo width={230} height={80} />
      </View>
      <View style={$mainContent}>
        <View style={$videoOuterContainer}>
          <VideoView player={player} style={{ width: "100%", height: "100%" }} contentFit="cover" />
`

I don't know if it's critical, but want to get more context about this if it's possible

Error: Call to function 'VideoPlayer.constructor' has been rejected.
[...]

PR #48914, opened 2026-08-14
[video][android] Don't require an Activity in VideoPlayer constructor

huextrat:
# Why

Creating a `VideoPlayer` via `useVideoPlayer()` on Android throws when the current `Activity` is briefly unavailable — for example during a cold start with an immediate background/recreation of the Activity:

```
Call to function 'VideoPlayer.constructor' has been rejected.
→ Caused by: The current activity is no longer available
```

The constructor obtained the application context through `appContext.throwingActivity.applicationContext`, so it required an `Activity` it doesn't actually need — the application context does not depend on any Activity. This is the same failure family as #42939 (player creation while the app is backgrounded).

# How

Replaced `appContext.throwingActivity.applicationContext` with `appContext.reactContext?.applicationContext` in the `VideoPlayer` constructor (`VideoModule.kt`), following the pattern already used elsewhere in the repo (e.g. `VideoManager.onModuleCreated`). If the react context is gone, it now throws the typed `Exceptions.ReactContextLost()` instead of an NPE.

Other `throwingActivity` usages were left untouched (`isPictureInPictureSupported`, fullscreen entry in `VideoView`) since those genuinely require an Activity.

# Test Plan

- `./gradlew :expo-video:compileDebugKotlin` — builds successfully.
- `./gradlew :expo-video:spotlessCheck` — passes.
- Manual: player creation no longer throws `MissingActivity` when the Activity is destroyed/recreating; playback works as before when the Activity is available.

# Checklist

- [x] Documentation is up to date to reflect these changes (eg: https://docs.expo.dev and README.md).


packages/expo-video/CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 
 ### 🐛 Bug fixes
 
+- [Android] Fix `VideoPlayer` constructor throwing `MissingActivity` when the player is created while the `Activity` is briefly unavailable (e.g. during a configuration change or right after backgrounding) — the constructor only needs the application context, which doesn't depend on the `Activity`. ([#48914](https://github.com/expo/expo/pull/48914) by [@huextrat](https://github.com/huextrat))
```
vonovak:
```suggestion
- [Android] Fix `VideoPlayer` constructor throwing `MissingActivity` when the player is created while the `Activity` is briefly unavailable. ([#48914](https://github.com/expo/expo/pull/48914) by [@huextrat](https://github.com/huextrat))
```

vonovak (approved):

vonovak:
thank you!

PR #48917, opened 2026-08-14
[router] Remove nested screen params navigation

Ubax:
# Why

# How

# Test Plan

# Checklist


packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
```diff
@@ -379,142 +365,48 @@ export function useNavigationBuilder<
...
-      // Schedule an update if the state needs to be updated
-      setState(nextState);
-    }
-  });
-
-  // The up-to-date state will come in next render, but we don't need to wait for it
-  // We can't use the outdated state since the screens have changed, which will cause error due to mismatched config
-  // So we override the state object we return to use the latest state as soon as possible
-  state = nextState;
+  const uncommittedState = state;
```
jakub-agent, line 409: [🚨 **critical**] `expo-router#test` dies with a heap OOM on this head. The `check-packages` job logs `FATAL ERROR: Ineffective mark-compacts near heap limit - JavaScript heap out of memory` and `expo-router#test ... exited (129)`; on #48895 the same job exits `1`.

I reproduced it and narrowed it to `src/__tests__/smoke.test.ios.tsx › 'can redirect during the initial render'` — a `<Redirect>` rendered from the root `_layout` on the first render. On #48895 that test fails with a bounded `Maximum update depth exceeded`; on this head memory grows without bound and it still OOMs with `--max-old-space-size=8192`. `smoke.test.android.tsx`, `navigation.test.ios.tsx` and `native-tabs/__tests__/render.test.ios.tsx` die the same way.

I have not pinned the exact allocation site. Anchoring here because this file is the most likely candidate: the `nextState` reconciliation and the `useScheduleUpdate` flush that previously converged the redirect cycle are both gone, so nothing caps the re-derivation any more. `<Redirect>` from a layout is the standard auth-gate pattern, so this is user-facing, not only a test problem.

packages/expo-router/src/__tests__/issues.test.ios.tsx
```diff
@@ -144,11 +144,6 @@ it('can navigate during first render', async () => {
 
   expect(screen.getByTestId('second')).toBeVisible();
   expect(screen.queryByTestId('index')).toBeNull();
```
jakub-agent, line 146: [🏗️ **design**] The `router.back()` assertions that followed here were deleted rather than updated. Relative to `main` this is a behavior change: after `router.push` during the first render, going back no longer returns to `index`. I re-added the removed lines against this head and they fail; the same lines pass on `main`.

They already fail on #48895, so the change came earlier in the stack — but this is where the evidence disappears. Could this be fixed, or kept as a skipped/failing test with a tracking issue and a changelog line? Dropping the assertions silently makes the regression invisible to the next person who touches this path.

packages/expo-router/src/navigationParams.ts
```diff
@@ -38,29 +38,45 @@ const internalExpoRouterParamNames = [
...
+export function warnIfScreenParam(params: object | undefined) {
+  if (
+    process.env.NODE_ENV !== 'production' &&
+    params &&
+    'screen' in params &&
+    !paramsObjectsWarnedForScreen.has(params)
+  ) {
+    paramsObjectsWarnedForScreen.add(params);
+    console.warn(
+      "Navigating with a `screen` param is no longer supported. Expo Router now builds real navigation state for nested destinations, so `screen` and `params` are treated as ordinary user params. Navigate with a full href instead, for example `router.push('/(tabs)/feed')`."
```
jakub-agent, line 53: [🏗️ **design**] This warns for any route params containing `screen`, but the premise of this PR is that `screen` is now an ordinary user param. `useSearchParams.test.ios.tsx` asserts that `?screen=feed` is surfaced, and has to `jest.spyOn(console, 'warn')` to silence this exact warning — that is the false positive showing up in the PR's own test.

A deep link to `/?screen=feed`, or a route with a `screen` query param, will now tell users that what they did "is no longer supported", which is the opposite of the new behavior.

Consider warning at the navigation entry points instead (`router.push`/`replace`/`linkTo` receiving a nested `screen` in a params object), rather than in `getRouteInfoFromState`, which sees every derived route including plain query params.

packages/expo-router/src/ui/TabRouter.tsx
```diff
@@ -68,26 +62,16 @@ export function ExpoTabRouter(options: ExpoTabRouterOptions) {
...
         };
-        return rnTabRouter.getStateForAction(state, action, routerConfigOptions);
-      } else if (route.state !== undefined) {
-        // TODO(@ubax): Remove this branch together with nested trigger href support. Refocusing
-        // a tab that hosts a navigator must not re-apply the trigger's nested payload
-        // (`params.screen`), which would reset the preserved child state.
+      }
+
+      if (!isSwitching && route.state !== undefined) {
         const selectedRoute = attachRouteState(route, action);
```
jakub-agent, line 68: [💡 suggestion] `selectedRoute` is used only for the identity comparison — the route it builds is thrown away and `rnTabRouter.getStateForAction` recomputes it. `attachRouteState` also has a side effect (the unmarked-state `console.warn`), which calling it as a predicate hides.

Either collapse it into the condition, or give it a name that says what is being asked:

```ts
if (!isSwitching && route.state !== undefined && attachRouteState(route, action) === route) {
  return { state, affectedRouteKey: route.key };
}
```

packages/expo-router/src/native-tabs/NativeBottomTabsRouter.tsx
```diff
@@ -71,13 +110,12 @@ export function NativeBottomTabsRouter(options: TabRouterOptions) {
               // because it was used to perform zoom transition on another tab
               const params = removeParams(
                 appendInternalExpoRouterParams(route.params, expoParams),
-                [
-                  INTERNAL_EXPO_ROUTER_ZOOM_TRANSITION_SCREEN_ID_PARAM_NAME,
-                  INTERNAL_EXPO_ROUTER_ZOOM_TRANSITION_SOURCE_ID_PARAM_NAME,
-                ]
+                zoomParamNames
               );
+              const state = route.state ? removeZoomParamsFromState(route.state) : undefined;
```
jakub-agent, line 115: [💡 suggestion] When `route.state` is absent this adds a `state: undefined` key that was not there before, because the returned object spreads `state` unconditionally. The rest of the PR avoids that — `removeZoomParamsFromState` uses `...(childState ? { state: childState } : undefined)` and `useLinkBuilder` uses `...(route.state !== undefined ? ... : undefined)`.

It matters because the deep-equality helpers around this code compare key counts: `fast-deep-equal` in `attachRouteState` and `isEqual(result.state, navigationState)` in `resolveNavigationDestination` would see two otherwise identical routes as different. Same conditional spread here would keep the shape stable.

packages/expo-router/src/link/preview/utils.ts
```diff
@@ -56,50 +56,73 @@ export function getPreloadedRouteFromRootStateByHref(
...
-    const targetParams = actionStateRoute.params;
-    const activeRoutes = stackState.routes.slice(0, stackState.index + 1);
-    const preloadedRoutes = stackState.routes.slice(stackState.index + 1);
+    const focusedRoute = navigationState.routes[navigationState.index]!;
+    const preloadedRoute = navigationState.routes
+      .slice(navigationState.index + 1)
+      .find((route) => routesHaveSameShape(route, targetRoute));
+    if (preloadedRoute) {
+      // A focused route with the same shape means the destination is already active.
+      return routesHaveSameShape(focusedRoute, targetRoute) ? undefined : preloadedRoute;
```
jakub-agent, line 83: [💡 suggestion] This "destination is already active" guard went from a shallow `name` + params check to full-subtree equality via `routesHaveSameShape`. `focusedRoute` is a live, hydrated route while `targetRoute` comes from `store.getStateForHref`, so the two can legitimately differ in nested `routes.length` / `index` while still being the same destination. When they do, the guard stops firing and the preview navigates to a route the user is already on.

None of the four `getPreloadedRouteFromRootStateByHref` tests cover this branch — they all assert a returned route. A case where the focused, hydrated destination returns `undefined` would pin the intended semantics.

Small things while here: `// TODO(@ubax):Try` on line 62 is missing a space, and `TODO(@uabx)` in `NativeBottomTabsRouter.tsx:93` is a typo for `@ubax`.

jakub-agent (changes requested):

Stacked on top of #48895. Replacing `screen`/`params`/`initial` with real navigation state plus the `__internal__routerActionState` marker is the right direction and is consistent with the rest of the series. The removal looks complete: no references to `getActionFromState`, `NavigatorScreenParams`, `getPayloadFromStateRoute`, `stateToAction` or `useScheduleUpdate` remain, and `scheduleUpdate`/`flushUpdates` are still consumed by `useDescriptors`, so that context is not left dead.

One blocker: on this head `expo-router#test` no longer fails, it crashes. The `check-packages` job reports `FATAL ERROR: Ineffective mark-compacts near heap limit - JavaScript heap out of memory` and `expo-router#test ... exited (129)`. On #48895 the same job exits `1`. Note that `jest-ubuntu` being green does not cover this package — those jobs run `@expo/cli` only.

The rest are smaller: one dev warning that contradicts the PR's own premise, a set of deleted assertions that hide a behavior change, and three readability/robustness points. Details inline.

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
Ubax: Fix it

packages/expo-router/src/__tests__/issues.test.ios.tsx
Ubax: Will be addressed separately in the parent PR

packages/expo-router/src/navigationParams.ts
Ubax: Remove `jest.spyOn(console, 'warn')` and assert that the warnings happen. Add TODO to remove the warning in future versions. 

Add one more TODO to add a flag for disabling migration warnings

packages/expo-router/src/native-tabs/NativeBottomTabsRouter.tsx
Ubax: use the conditional spread ...(state ? ......)

packages/expo-router/src/link/preview/utils.ts
Ubax: Add test coverage. Fix todos. Apply more efficient and correct comparison if possible

Issue #30, opened 2016-12-30
chromeOS support


acornejo:
This issue is to track adding a chromeOS app for exponent, relying on the recent addition of the android runtime component.

This would not only add another platform where extant apps can be run (iOS, Android or ChromeOS), but it would also allow developers to develop+test their applications using a chromebook instead of a physical device or a simulator (which traditionally are more resource intensive than simulators).

ChromeOS officially supports android apps on a subset of chromebooks, and quite a few android apps already support this (evernote, vine, skype, adobe lightroom, snapchat, etc).

When running in deverloper mode, all chromebooks (not restricted to a supported hardware subset) can run android apps, and can use ARC welder to convert android APK's into chromeOS apps (https://chrome.google.com/webstore/detail/arc-welder/emfinbmielocnlhgmfkkmkngdoccbadn).

I tried converting exponent APK to a chromeOS app using ARC welder, but there were issues after the splash screen. I suspect this may have to do with exponent trying to access some device not available on my laptop (for instance, GPS), but didn't dig any further.

dikaiosune:
Hi! We're moving feature requests to Canny. I [created one](https://expo.canny.io/feature-requests/p/chrome-os-support) for this there. Thanks!

Issue #27, opened 2016-12-18
TexInput with custom fonts crash, sometimes...


yestoall:
yes it's really weird, it crash sometimes, other times works perfect.

![simulator screen shot 18 dec 2016 09 46 39](https://cloud.githubusercontent.com/assets/29780/21292459/e69fe382-c506-11e6-9874-80fdbbf0447a.png)

this is the error when program is compiled, after press escape the app works ok, with the font loaded.

of course i'm preloading the fonts in the good way

brentvatne:
share the code please! :)

PR #48920, opened 2026-08-14
[cli] Add fingerprint and device-inspection utilities

vonovak:
# Why

`npx expo needs-rebuild` (next PR in the stack) needs to compute the project fingerprint and read what's installed on a device. This PR adds the shared building blocks, usage is up the stack.

# How

- `utils/nativeFingerprint.ts`: resolve `@expo/fingerprint` from the app project, shared fingerprint options, write/read the `expo prebuild` marker file, and compare staleness against it
- `utils/zipEntry.ts`: minimal zip reader so a fingerprint file can be read out of an APK via a few `adb exec-out dd` calls instead of pulling the whole apk
- `adb.ts` / `ADBServer.ts`: `getPackagePathsAsync` (get apk path on device) and `runRawAsync` (run adb commands)

# Test Plan

- ci, local tests up the stack

# Checklist

---


vonovak:

PR #48921, opened 2026-08-14
[cli] Add npx expo needs-rebuild and record prebuild fingerprints

vonovak:
# Why

# How

`packages/@expo/cli/src/needsRebuild/needsRebuildAsync.ts` is the main piece - it computes fingerprint and tries to find matching builds for iOS or android (which embed the fingerprint in a file). If the build doesn't have the same fingerprint, it gives advice to for recovery.

# Test Plan

- unit tests added
- local testing

# Checklist

---

### DIFF
--- a/docs/pages/more/expo-cli.mdx
+++ b/docs/pages/more/expo-cli.mdx
@@ -34,7 +34,7 @@ Usage
 
 Commands
   start, export
-  run:ios, run:android, prebuild
+  run:ios, run:android, prebuild, needs-rebuild
   install, customize, config
   login, logout, whoami, register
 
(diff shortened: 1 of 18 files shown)


docs/pages/more/expo-cli.mdx
```diff
@@ -370,6 +370,27 @@ Expo CLI will automatically sign the device for development, install the app, an
 
 If you don't have any developer profiles setup on your computer then you'll need to set them up manually outside of Expo CLI by following this guide: [Setup Xcode signing](https://expo.fyi/setup-xcode-signing).
 
+### Detecting stale builds
```
kitten: I think it'd be great if you could discuss the design here with @byCedric and @ramonclaudio before we move along with the base PR. This seems to have some overlap with `expo compile` and future functionality.

Basically, `expo needs-rebuild` needs to be aligned in design with that potentially, and it's a but unclear to me whether `needs-rebuild` also has the highest utility it couldhave considering that it won't know which simulator to pick presumably, meaning that, if it's scoped to local builds it'd also be possible to capture the prior build in a temp folder or in `.expo`, so I'm unclear on whether we have enough notes that were passed around to make sure the design here is durable.

Basically, let's make sure that this is what we want before proceeding since it's hard to "take back" a command

vonovak:

PR #48922, opened 2026-08-14
[expo-constants] Expose the embedded fingerprint as Constants.fingerprint

vonovak:
# Why

[#49905](https://github.com/expo/expo/pull/49905) embeds `app.fingerprint` in debug builds. This exposes it to JavaScript, so a running app can say what it was built from.

# How

`Constants.fingerprint` reports the embedded hash.

- iOS reads it in `ConstantsProvider.swift`, Android in `ConstantsService.kt`.
- `null` whenever there is nothing honest to report: release builds, builds made before the embed existed, web, and Expo Go — whose binary carries a fingerprint of Expo Go itself, which says nothing about the loaded project.
- The two Expo Go platforms are protected differently, and only one deliberately. Android's `ConstantsBinding` extends `ConstantsService`, which now reads the file, so it overrides `fingerprint` to null explicitly. iOS's `EXConstantsBinding` extends the ObjC `EXConstantsService`, which reads no fingerprint at all, so the key is simply absent and `Constants.ts` coalesces it. `EXConstantsBinding.m` carries a note to add the override if it ever moves to `ConstantsProvider`.
- `fingerprintProtocolParity-test.js` pins the `app.fingerprint` literal across the writer and both native readers, because nothing can share that name across those languages.

An earlier revision of this PR also added a `GET /_expo/fingerprint` dev-server endpoint, a `checkFingerprintAsync()` in the `expo` package, and a terminal warning on mismatch. That is dropped: the endpoint has to live in `@expo/cli` because `npx expo start` owns the Metro port, and the staleness answer is available from outside through `npx @expo/agent-cli status --explain`. `@expo/cli` is untouched by this PR.

# Test Plan

- `packages/expo-constants/src/__tests__/Constants-test.ts` — the value, and `null` where it must be null.
- `packages/expo-constants/scripts/__tests__/fingerprintProtocolParity-test.js` — the shared literal across Kotlin, Swift and the writer.
- `et check-packages expo-constants expo jest-expo` passes: build, typecheck, lint, tests.

Not covered here: reading the value out of a real build on a device. That needs this branch built onto one, and it is what `status --explain` does end to end.

# Checklist


packages/@expo/cli/src/start/server/metro/fingerprintService.ts
```diff
@@ -0,0 +1,172 @@
...
+      ? changedPath
+      : path.join(event.rootDir, changedPath);
+    if (!absolutePath.startsWith(stateDirPrefix)) {
+      return true;
+    }
+  }
+  return false;
+}
+
+/** One app launch warns once; the next launch warns again. */
```
Kudo: **Medium — cache invalidation breadth**

Any watched path outside `.expo` clears the fingerprint cache. That is correct for safety, but fingerprint compute is documented as ~1–2.5s CPU, and apps re-announce on reload. In an active edit session this can thrash the cache and re-hash the whole project often.

Consider narrowing invalidation (e.g. known fingerprint inputs / debounce), or at least documenting the cost for monorepos / CI-without-watcher note already present nearby.

apps/expo-go/android/expoview/src/main/java/versioned/host/exp/exponent/modules/universal/ConstantsBinding.kt
```diff
@@ -41,6 +41,8 @@ class ConstantsBinding(
             )
           )
           this["isDetached"] = false
+          // The embedded fingerprint describes the Expo Go binary, not the loaded project.
+          this["fingerprint"] = null
         }
```
Kudo, line 49: **Low — iOS Expo Go parity**

Good that Android Expo Go forces `fingerprint = null` so the Go binary’s stamp is not mistaken for the loaded project. iOS still goes through `EXConstantsBinding.m` / ObjC `EXConstantsService` (no fingerprint reader today), so this is fine for now — but if Expo Go ever adopts `ConstantsProvider`, mirror this null override in `EXConstantsBinding` so docs (`null` in Expo Go) stay true.

packages/expo/src/fingerprint/checkFingerprintAsync.ts
```diff
@@ -0,0 +1,160 @@
...
+      commands: [],
+    };
+  }
+  if (embeddedHash === serverHash) {
+    return { status: 'up-to-date', reason: 'hash-match', embeddedHash, serverHash, commands: [] };
+  }
+  return {
+    status: 'rebuild-required',
+    reason: 'hash-mismatch',
+    embeddedHash,
```
Kudo: **Medium — fingerprintVersion not consulted**

The middleware returns `fingerprintVersion`, but the client only compares hashes. If embed-time `@expo/fingerprint` (from the base embed PR) and the project's resolved package diverge, hashes can mismatch permanently and spuriously report `rebuild-required`.

Worth either: (1) treating version mismatch as `unknown` / `check-failed` with a clear reason, or (2) documenting that embed and runtime must use the same package (parity test already pins options, not version identity).

Kudo:
@expo-bot verify --fable

packages/@expo/cli/src/start/server/middleware/FingerprintMiddleware.ts
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    res.statusCode = 200;
+    const response: FingerprintResponse = {
+      platform,
+      hash: server.hash,
+      fingerprintVersion: server.fingerprintVersion,
+      ...(mismatch ? { mismatch } : null),
+    };
+    res.end(JSON.stringify(response));
```
Kudo: **Low — HEAD still writes a body**

`HEAD` is listed in `Allow`, but the success path always `res.end(JSON.stringify(response))`. Prefer skipping the body (or early-returning after headers) for `HEAD` so clients/proxies that treat HEAD strictly do not see unexpected payload.

Kudo (changes requested):
same as downstack pr. we should move the work from expo cli to agent-cli

packages/expo-constants/scripts/__tests__/fingerprintProtocolParity-test.js
```diff
@@ -0,0 +1,40 @@
...
+  it.each([
+    [
+      'the Android reader (ConstantsService.kt)',
+      'expo-constants/android/src/main/java/expo/modules/constants/ConstantsService.kt',
+      '"app.fingerprint"',
+    ],
+    [
+      'the iOS reader (ConstantsProvider.swift)',
+      'expo-modules-core/ios/Utilities/ConstantsProvider.swift',
+      'forResource: "app", withExtension: "fingerprint"',
```
Kudo: the test feels fragile. if swift split `'forResource: "app", withExtension: "fingerprint"` to multi-lines. it breaks. maybe we can just drop the test

packages/jest-expo/src/preset/moduleMocks/expoModules.js
```diff
@@ -2242,6 +2242,8 @@ module.exports = {
           debugMode: { type: 'boolean', mock: true },
           deviceName: { type: 'string' },
           executionEnvironment: { type: 'string' },
+          // Null, not the generic 'mock' string, or `if (Constants.fingerprint)` gates would lie.
```
Kudo:
```suggestion
```

unclear comment "gates would lie"

Kudo (approved):

packages/@expo/cli/src/start/server/metro/fingerprintService.ts
vonovak: the code is deleted; the check now uses agent-cli's existing cache

packages/expo/src/fingerprint/checkFingerprintAsync.ts
vonovak: added fingerprintVersion

PR #48923, opened 2026-08-14
[cli][expo] Add the app-to-dev-server fingerprint channel

vonovak:
# Why

when you make a change that modifies the fingerprint (app config, etc), the app becomes "stale". Currently there's nothing that would inform users about that. This PR reports that situation in the `expo start` logs. It remains to show this in the app. 

# How

- expo exports checkFingerprintAsync(), comparing Constants.fingerprint against the dev server's current project fingerprint
- The CLI serves GET `/_expo/fingerprint?platform=<android|ios>` and returns the fingerprint to the app. App can send the fingerprint in a request header and the CLI can print the warning if they don't match

Known limitations: (1) `@expo/fingerprint` hashes only an allowlist of asset paths — app icons, Google services files, and the `expo-splash-screen` and `expo-font` plugin props — so editing a file referenced by any other plugin, such as `expo-notifications's` icon or sounds, won't change the fingerprint and this check won't report the rebuild it actually needs.

(2) this fingerprint check  runs in JS after the bundle loads, so having it directly in dev launcher might be nicer. But that can be done later. But it can also be a good thing that this doesn't require dev client

# Test Plan

- tested locally

# Checklist

---


vonovak:

PR #48925, opened 2026-08-14
[ios][background-task] Read the real Background App Refresh status in getStatusAsync

### LINKED ISSUE #48786: [expo-background-task] iOS getStatusAsync() never reads the real Background App Refresh permission
## Summary

On iOS, `BackgroundTask.getStatusAsync()` does not consult the OS Background App Refresh setting at all. It is derived purely from whether the build is running on the Simulator, so it returns `restricted` on the Simulator and `available` on **every** physical device — regardless of what the user has set in **Settings → General → Background App Refresh**.

This makes it unusable for its most obvious purpose: telling the user that background work won't run because they've switched it off.

## Where

[`packages/expo-background-task/ios/BackgroundTaskScheduler.swift#L214-L221`](https://github.com/expo/expo/blob/main/packages/expo-background-task/ios/BackgroundTaskScheduler.swift#L214-L221):

```swift
public static func supportsBackgroundTasks() -> Bool {
#if targetEnvironment(simulator)
  // If we're on emulator we should definetly return restricted
  return false
#else
  return true
#endif
}
```

That value is returned directly as the status in [`BackgroundTaskModule.swift#L68-L71`](https://github.com/expo/expo/blob/main/packages/expo-background-task/ios/BackgroundTaskModule.swift#L68-L71):

```swift
AsyncFunction("getStatusAsync") {
  return BackgroundTaskScheduler.supportsBackgroundTasks()
    ? BackgroundTaskStatus.available : .restricted
}
```

`UIApplication.shared.backgroundRefreshStatus` is never read.

## The predecessor did this correctly

`expo-background-fetch` — still on `main` — reads it properly, including all three enum cases, at [`BackgroundFetchModule.swift#L39-L54`](https://github.com/expo/expo/blob/main/packages/expo-background-fetch/ios/BackgroundFetchModule.swift#L39-L54):

```swift
private func getStatus() -> BackgroundFetchStatus {
  let backgroundRefreshStatus = UIApplication.shared.backgroundRefreshStatus
  switch backgroundRefreshStatus {
  case .restricted: return .restricted
  case .available:  return .available
  case .denied:     return .denied
  @unknown default: /* logs, returns .denied */
  }
}
```

So the check appears to have been dropped, likely unintentionally, in the move to the `BGTaskScheduler`-based implementation rather than as a deliberate change.

## Reproduction

1. Build an app using `expo-background-task` onto a physical iOS device (a development build is fine).
2. `await BackgroundTask.getStatusAsync()` → `BackgroundTaskStatus.Available`.
3. Turn **Settings → General → Background App Refresh** off, either for that app specifically or globally.
4. Call it again → still `BackgroundTaskStatus.Available`.

Expected: a value reflecting that background refresh is unavailable.

## Two related gaps, in case they're in scope for the same fix

1. **`BackgroundTaskStatus` has only two cases** (`restricted = 1`, `available = 2`, [`BackgroundTaskRecords.swift`](https://github.com/expo/expo/blob/main/packages/expo-background-task/ios/BackgroundTaskRecords.swift)), while `UIBackgroundRefreshStatus` has three. `.denied` — the user switching it off for this app *or* system-wide, and by far the common case in practice — has nowhere to map except onto `restricted`, which Apple documents as meaning parental controls / MDM. Consumers can't distinguish "the user turned it off, send them to Settings" from "policy forbids it, don't bother asking".

2. **`UIApplication.backgroundRefreshStatusDidChangeNotification`** (iOS 7+) isn't surfaced, so consumers poll on every `AppState` resume to notice changes.

Also worth noting for whoever picks this up: iOS reports `.available` under **Low Power Mode** even though it suppresses background execution — verified on device. So a correct status read still isn't a promise that background work will run. (Compare the closed #11505, which was the same observation about `expo-background-fetch`.)

## Happy to open a PR

The fix for the core issue is small — replace `supportsBackgroundTasks()` in the status path with a real `backgroundRefreshStatus` read, keeping the Simulator short-circuit if that behaviour is wanted for `registerTaskAsync`. I've implemented and device-verified this logic (including the 3-value mapping and the change notification) in a standalone module, [`expo-background-refresh-status`](https://www.npmjs.com/package/expo-background-refresh-status), and would gladly upstream it in whatever shape you'd prefer — whether that's just the status read, or also the enum and notification. Let me know if a PR is welcome and which scope you'd want.

## Environment

- `expo-background-task` 1.0.10, Expo SDK 54, physical iPhone (iOS 18)
- Also verified present on `main` at the time of filing

arnautresserras:
# Why

Fixes #48786.

On iOS, `BackgroundTask.getStatusAsync()` never consulted the OS Background App Refresh setting. It
was derived purely from `BackgroundTaskScheduler.supportsBackgroundTasks()`, a bare
`#if targetEnvironment(simulator)`, so it returned `Restricted` on the Simulator and `Available` on
**every** physical device — no matter what the user had set in **Settings → General → Background App
Refresh**. That makes it unusable for its most obvious purpose: telling the user that background work
will not run because they switched it off.

The predecessor package `expo-background-fetch`, still on `main`, reads
`UIApplication.shared.backgroundRefreshStatus` correctly including all three cases
([`BackgroundFetchModule.swift#L39-L54`](https://github.com/expo/expo/blob/main/packages/expo-background-fetch/ios/BackgroundFetchModule.swift#L39-L54)),
which suggests the check was dropped unintentionally in the move to the `BGTaskScheduler`
implementation rather than removed on purpose.

# How

- `getStatusAsync` now reads `UIApplication.shared.backgroundRefreshStatus`. The Simulator
  short-circuit is kept ahead of it — `BGTaskScheduler` never runs work there, so the system setting
  is meaningless and `Restricted` stays the right answer.
- Added `BackgroundTaskStatus.Denied` (`3`) on both the Swift and TypeScript sides.
  `UIBackgroundRefreshStatus` has three cases and `BackgroundTaskStatus` had two, so without this
  `.denied` — the user switching the toggle off, by far the common case — could only be reported as
  `Restricted`, which Apple documents as meaning parental controls / MDM. Consumers could not tell
  "send the user to Settings" from "policy forbids it, do not bother asking".
- `registerTaskAsync` still skips registration only on `Restricted`. `Denied` deliberately falls
  through and registers, because the user can turn Background App Refresh back on at any moment and
  a task that was never registered would not start running when they do. This keeps today's
  on-device behaviour unchanged for every user who has the setting off.
- The `Restricted` warning text in `registerTaskAsync` said "not supported on iOS simulators", which
  is now reachable on a real device under a device policy, so it mentions both.
- An unknown future `UIBackgroundRefreshStatus` case maps to `available`. Note this differs from
  `expo-background-fetch`, which maps it to `.denied`: a status we cannot interpret should not make
  apps tell users background refresh is switched off when it may well be on. Happy to flip it to
  match the sibling package if you would rather have consistency there.

The third item from the issue — surfacing
`UIApplication.backgroundRefreshStatusDidChangeNotification` so consumers stop polling on every
`AppState` resume — is deliberately **not** in this PR. It is a new public API surface with its own
naming and listener-shape questions, and it would make this change harder to review. Happy to open
it as a follow-up if you want it.

# Test Plan

- New `ios/Tests/BackgroundTaskStatusTests.swift` covers the `UIBackgroundRefreshStatus` →
  `BackgroundTaskStatus` mapping for all three cases plus the unknown-case fallback.
- `et check-packages expo-background-task` passes.
- Device-verified behaviour: this is the same logic I implemented and verified on a physical iPhone
  (iOS 18) in a standalone module,
  [`expo-background-refresh-status`](https://www.npmjs.com/package/expo-background-refresh-status),
  before filing #48786 — `available` at baseline, `denied` with the per-app switch off, `denied`
  with the global switch off, and back to `available` on re-enable.
- `.restricted` is not reachable without a supervised device, so it ships covered only by the unit
  test above. That is the reason `registerTaskAsync` keys off `Restricted` alone and consumers are
  documented to test for `Available` rather than for a specific unavailable value.

# Checklist

The API reference under `docs/` is generated from the JSDoc in `src/`, so this PR touches no `docs/`
files — tell me if you would rather I run `et generate-docs-api-data -p expo-background-task` and
include the regenerated data here.

# Note for reviewers

One behavioural consequence worth stating explicitly: apps that today detect "background refresh is
unavailable" with `status === BackgroundTaskStatus.Restricted` will not see the newly reported
`Denied`. The JSDoc now tells consumers to test for `Available` instead. If you would prefer this
land as a breaking change note, or prefer to keep the enum two-valued and map `.denied` onto
`Restricted` (at the cost of `registerTaskAsync` then skipping registration for those users), I am
glad to rework it.

### DIFF
--- a/packages/expo-background-task/ios/Tests/BackgroundTaskStatusTests.swift
+++ b/packages/expo-background-task/ios/Tests/BackgroundTaskStatusTests.swift
@@ -1,0 +1,25 @@
+import UIKit
+import XCTest
+
+@testable import ExpoBackgroundTask
+
+final class BackgroundTaskStatusTests: XCTestCase {
+  func testMapsAvailableRefreshStatus() {
+    XCTAssertEqual(BackgroundTaskStatus(refreshStatus: .available), .available)
+  }
+
+  func testMapsDeniedRefreshStatus() {
+    XCTAssertEqual(BackgroundTaskStatus(refreshStatus: .denied), .denied)
+  }
+
+  func testMapsRestrictedRefreshStatus() {
+    XCTAssertEqual(BackgroundTaskStatus(refreshStatus: .restricted), .restricted)
+  }
+
+  func testMapsUnknownRefreshStatusToAvailable() {
+    guard let unknown = UIBackgroundRefreshStatus(rawValue: 99) else {
+      return
+    }
+    XCTAssertEqual(BackgroundTaskStatus(refreshStatus: unknown), .available)
+  }
+}
--- a/packages/expo-background-task/CHANGELOG.md
+++ b/packages/expo-background-task/CHANGELOG.md
@@ -6,7 +6,11 @@
 
 ### 🎉 New features
 
+- [iOS] Added `BackgroundTaskStatus.Denied`, returned when the user has turned Background App Refresh off for the app or for the whole system. It is distinct from `Restricted`, which the user cannot lift. ([#PR](https://github.com/expo/expo/pull/PR) by [@arnautresserras](https://github.com/arnautresserras))
+
 ### 🐛 Bug fixes
+
+- [iOS] Fixed `getStatusAsync` reporting `Available` on every physical device regardless of the system Background App Refresh setting. It now reads `UIApplication.backgroundRefreshStatus` instead of only checking whether the app is running on a simulator. ([#PR](https://github.com/expo/expo/pull/PR) by [@arnautresserras](https://github.com/arnautresserras))
 
 ### 💡 Others
 
--- a/packages/expo-background-task/ios/BackgroundTaskModule.swift
+++ b/packages/expo-background-task/ios/BackgroundTaskModule.swift
@@ -1,5 +1,6 @@
 // Copyright 2024-present 650 Industries. All rights reserved.
 import ExpoModulesCore
+import UIKit
 
 private let onTasksExpired = "onTasksExpired"
 public let onTasksExpiredNotification = Notification.Name(onTasksExpired)
@@ -65,10 +66,18 @@ public class BackgroundTaskModule: Module {
       }
     }
 
-    AsyncFunction("getStatusAsync") {
-      return BackgroundTaskScheduler.supportsBackgroundTasks()
-        ? BackgroundTaskStatus.available : .restricted
+    // Runs on the main queue because `UIApplication.shared` must only be touched there,
+    // while `AsyncFunction` runs off the main queue by default.
+    AsyncFunction("getStatusAsync") { () -> BackgroundTaskStatus in
+      // BGTaskScheduler never runs work on a simulator, so the system setting is
+      // meaningless there and we keep reporting `restricted`.
+      if !BackgroundTaskScheduler.supportsBackgroundTasks() {
+        return .restricted
+      }
+
+      return BackgroundTaskStatus(refreshStatus: UIApplication.shared.backgroundRefreshStatus)
     }
+    .runOnQueue(.main)
   }
 
   @objc func handleTasksExpiredNotification(_ notification: Notification) {
--- a/packages/expo-background-task/ios/BackgroundTaskRecords.swift
+++ b/packages/expo-background-task/ios/BackgroundTaskRecords.swift
@@ -1,9 +1,32 @@
 // Copyright 2024-present 650 Industries. All rights reserved.
 import ExpoModulesCore
+import UIKit
 
 enum BackgroundTaskStatus: Int, Enumerable {
   case restricted = 1
   case available = 2
+  case denied = 3
+}
+
+extension BackgroundTaskStatus {
+  /**
+   Maps the system Background App Refresh setting onto the status reported to JavaScript.
+
+   An unknown future case is reported as `available`: a status we cannot interpret must not make
+   apps tell their users that background refresh is switched off when it may well be on.
+   */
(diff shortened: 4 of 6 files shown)


packages/expo-background-task/ios/Tests/BackgroundTaskStatusTests.swift
```diff
@@ -0,0 +1,25 @@
...
+    XCTAssertEqual(BackgroundTaskStatus(refreshStatus: .denied), .denied)
+  }
+
+  func testMapsRestrictedRefreshStatus() {
+    XCTAssertEqual(BackgroundTaskStatus(refreshStatus: .restricted), .restricted)
+  }
+
+  func testMapsUnknownRefreshStatusToAvailable() {
+    guard let unknown = UIBackgroundRefreshStatus(rawValue: 99) else {
+      return
```
LizunovSergey, line 21: Non-blocking: this `return` can make the test pass without exercising the `@unknown default` branch if the current SDK rejects the synthetic raw value. Could we use `XCTUnwrap`/`XCTFail` here (or otherwise make an unsupported synthetic value explicit) so the test cannot become vacuous?

LizunovSergey (commented):
Thanks for taking this on, this looks good to me. The implementation looks correct and preserves the existing registration behavior. I also ran the relevant checks locally, all passed. I left one non-blocking suggestion to strengthen the test

vonovak:
@expo-bot verify

vonovak:
@expo-bot verify

PR #48009, opened 2026-07-21
[iOS][location] Remove stale top-level scope and accuracy from permission response

Wenszel:
# Why

follow-up to: https://github.com/expo/expo/pull/41328

removes stale top-level `LocationPermissionResponse` properties that should be inside the `ios` object, per the [docs](https://docs.expo.dev/versions/latest/sdk/location/#locationpermissionresponse)

# How
removes `scope` and `accuracy` returned by `parsePermissions` in `EXForegroundPermissionRequester.m`

# Test Plan
Tested on BareExpo ✅

PR #48926, opened 2026-08-14
[location][ios] Nest background permission scope and accuracy under `ios` field

vonovak:
## Why

#48009 moved `scope` and `accuracy` under `ios` for the foreground permission response and dropped the stale top-level keys. The background requester also needs this treatment.

found by migrating tests from JS to TS

# How

- migrate tests to TS
- fix the permission response shape

# Test Plan

- bare expo

# Checklist


brentvatne:
@expo-bot check --fable

Wenszel (approved):

PR #48643, opened 2026-08-07
[deps] Bump react-native-get-random-values to ~2.0.0

cucumber-sp:
# Why

`react-native-get-random-values` is listed in `packages/expo/bundledNativeModules.json` (the map `expo install` uses to pick a compatible version) pinned at `~1.11.0`. The upstream package released `2.0.0` with:
```
**💥 Breaking Changes**
- Switch to New Architecture. Minimum supported React Native is now 0.81, minimum Expo SDK is 54.

**🎉 Enhancements**
- Improve Android performance by caching `SecureRandom`
- Add support for visionOS
```

Users installing `2.0.0` manually are already hitting `expo-doctor` "major version mismatch" warnings against the stale `~1.11.0` pin, e.g.:
- https://github.com/expo/expo/issues/46426
- https://github.com/expo/expo/issues/41316
- https://github.com/LinusU/react-native-get-random-values/issues/68

Since this repo already requires React Native well above 2.0.0's 0.81 minimum and ships with the New Architecture on by default, there's no compatibility blocker to bumping the recommendation.

# How

Bumped the `react-native-get-random-values` entry in `packages/expo/bundledNativeModules.json` from `~1.11.0` to `~2.0.0`, matching the existing convention for this file (tilde-pinned to the latest compatible minor/patch range). Added a matching `CHANGELOG.md` entry under "💡 Others" following the precedent set by prior `bundledNativeModules.json` additions/bumps.

# Test Plan

- Verified `packages/expo/bundledNativeModules.json` is still valid JSON.
- Confirmed via `getVersionedPackagesAsync` (`packages/@expo/cli/src/start/doctor/dependencies/getVersionedPackages.ts`) that `expo install react-native-get-random-values` resolves versions straight from this file, so the new pin takes effect without further code changes.
- No automated test covers this static data file; this mirrors how prior version bumps to `bundledNativeModules.json` were shipped.

# Checklist

- [x] I added a `CHANGELOG.md` entry
- [ ] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: updated a module plugin). — N/A, no module plugin change
- [x] Conforms with the [Documentation Writing Style Guide](https://github.com/expo/expo/blob/main/guides/Expo%20Documentation%20Writing%20Style%20Guide.md)

PR #48931, opened 2026-08-14
[expo] Remove react-native-get-random-values from bundledNativeModules

gabrieldonadel:
# Why

Although we list `react-native-get-random-values` in bundledNativeModules we don't actually use it anywhere anymore. In order to prevent users from getting outdated warnings like reported in https://github.com/expo/expo/pull/48643, we should just remove that entry and let users use whatever version they want

# How

Remove react-native-get-random-values from bundledNativeModules

# Test Plan

N / A

# Checklist


zoontek (approved):

kitten (approved):

alanjhughes (approved):

kitten:
Also related: #44718

PR #48933, opened 2026-08-14
[ios][core] Fixed a crash when Fabric creates a cached Expo view after a reload

### LINKED ISSUE #48575: [expo-modules-core][iOS] cached ExpoFabricView initializer aborts with AppContextLost during full reload

The fault is in the SDK source and reproduces in a normal SDK 57 New Architecture development build:

https://github.com/expo/expo/blob/sdk-57/packages/expo-modules-core/ios/Fabric/ExpoFabricView.swift

1. Render an Expo Fabric-backed native view continuously on the initial screen. A view from `expo-glass-effect` or `expo-image` is sufficient.
2. Start the iOS development build with Hermes and the New Architecture.
3. Trigger repeated full Metro reloads while that view is mounted. Editing an imported locale JSON file is a reliable way to cause a full bundle reload instead of component Fast Refresh.
4. Fabric intermittently mounts the cached Expo view class after the old `AppContext` has deallocated and before the replacement initializer is safely associated with the new context.

### Actual behavior

The app aborts through `fatalError(Exceptions.AppContextLost().reason)`:

```text
_assertionFailure
closure #1 in ExpoFabricView.injectInitializer(_:appContext:)
-[RCTComponentViewFactory createComponentViewWithComponentHandle:]
RCTPerformMountInstructions
-[RCTMountingManager performTransaction:]
```

Six separate crash reports with this exact signature were captured on 2026-08-06.

### Root cause

`makeViewClass` caches allocated Objective-C view classes in the process-wide `viewClassesRegistry` specifically for reuse after app reload. `injectInitializer` then replaces the cached class's default initializer with a block that weakly captures the current `AppContext`. When that context is released during a full reload, Fabric can still instantiate the cached class through the stale block. The guard fails and intentionally aborts the process. Registry access and initializer replacement are also unsynchronized while two React runtimes can briefly overlap.

### Expected behavior

Reloading should either instantiate the view with the current app context or defer mounting. It should not abort the process through an initializer retained by a process-wide cached class.

### Suggested fix

We locally verified that an exact-version patch applies and parses which:

- installs one initializer block per cached view class;
- gives that block a strong holder whose current `AppContext` is updated atomically on reload; and
- serializes access to the shared view-class registry.

This avoids both the stale weak capture and concurrent registry access. The holder remains owned by the cached initializer block and the current context replaces the prior one on every registration.

### Environment

```text
expo 57.0.8
expo-modules-core 57.0.7
react-native 0.86.0
Hermes / New Architecture development build
Xcode 27.0 (27A5228h)
iOS 27.0 simulator
macOS 27.0 (26A5388g)
```

uwayss:
## Why

Fixes #48575.

We cache the generated view classes in `viewClassesRegistry` so they survive a reload, and give each
one a `new` implementation that captures the current `AppContext` weakly. But the class outlives the
context, so that weak reference can already be nil by the time Fabric asks the class for a view.

On a reload the old context gets released, and the new one re-registers the views. Except when
`registerNativeModules` runs off the main thread, that registration is deferred to the main actor.
In the gap between those two things, Fabric can drain a mounting transaction that needs to *create*
an Expo view. The class is still registered with `RCTComponentViewFactory`, the weak reference is
gone, and we abort:

```
_assertionFailure
closure #1 in ExpoFabricView.injectInitializer(_:appContext:)   ExpoFabricView.swift:212
-[RCTComponentViewFactory createComponentViewWithComponentHandle:]
RCTPerformMountInstructions
-[RCTMountingManager performTransaction:]
```

I measured that gap at roughly 9ms per reload on a real app. Only mounts that need a genuinely new
view rather than one from the recycle pool fall into it, which is why it feels random.

## How

Each view class now gets a small holder that owns the context strongly, and every registration swaps
what the holder points at instead of injecting a second initializer.

The initializer then always reads the most recent context. Inside the gap above it reads the
outgoing one, which is still around precisely because the holder is holding it — and building a view
from it is safe, since `createView` just initialises the view and wires up the event dispatchers and
never touches the runtime. That view belongs to the surface that's on its way out anyway.

Nothing is kept alive longer than it needs to be: the old context goes as soon as the next
registration replaces it.

I keyed the holders by `ObjectIdentifier(viewClass)` rather than by class name, since the note on
`makeViewClass` points out that two classes can share a name in Expo Go.

No locking — `makeViewClass` is only reached from `registerNativeViews` (`@MainActor`), and the
initializer runs during a Fabric mounting transaction, so both sides are on the main thread already.
Happy to add some if you'd rather be defensive, though the two existing TODOs about that path look
like their own change.

One thing I left alone: the remaining `AppContextLost` can now only fire when a module holder is
missing from a live context, so the message is a bit off. Didn't want to widen the diff, but I'm
happy to reword it.

## Test Plan

Tested on a real SDK 57 app — `expo-modules-core` 57.0.8 built from source, RN 0.86.2, Hermes + New
Architecture, iOS 26.5 simulator.

I added some `NSLog`s around `AppContext.init`/`deinit` and the registration to actually see the gap:

```
AppContext INIT    0x1566918c0     new context
AppContext DEINIT  0x1439d95c0     old one released -> weak ref now nil
registerNativeModules -> main actor hop
registerNativeViews running        cached classes re-pointed
```

Those middle two lines are about 9ms apart. That turned out to be too narrow to hit deliberately —
233 reloads with Expo views on screen never crashed once. So I widened just that gap to 3s by
awaiting before `registerNativeViews()` and changed nothing else, which makes the first mount after
a reload land inside it.

- **Before:** 4 of 5 runs aborted, every one at the `fatalError` in `injectInitializer` with the
  stack above.
- **After:** 0 of 10. 29 context creations and 20 releases were logged over that run, so the reload
  path was definitely being exercised.

The ordering flips, which is really the whole point of the change:

```
before:  DEINIT old  ->  ...  ->  abort        (registration never got there)
after:   registerNativeViews  ->  DEINIT old   (released 3ms later)
```

Worth being upfront that this widens the race to make it reproducible — I couldn't hit the natural
9ms window on demand. Separately, two crashes with this exact signature showed up on my machine
during ordinary development before I'd instrumented anything.

### DIFF
--- a/packages/expo-modules-core/CHANGELOG.md
+++ b/packages/expo-modules-core/CHANGELOG.md
@@ -21,6 +21,7 @@
 
 ### 🐛 Bug fixes
 
+- [iOS] Fixed a crash on reload when Fabric instantiated a cached Expo view class after its app context had been released. ([#48575](https://github.com/expo/expo/issues/48575) by [@lighthold](https://github.com/lighthold)) ([#48933](https://github.com/expo/expo/pull/48933) by [@uwayss](https://github.com/uwayss))
 - [iOS] Fixed the tap that closes a SwiftUI menu also pressing the React Native view underneath it. ([#48419](https://github.com/expo/expo/issues/48419) by [@bohdanstefaniuk](https://github.com/bohdanstefaniuk)) ([#48463](https://github.com/expo/expo/pull/48463) by [@intergalacticspacehighway](https://github.com/intergalacticspacehighway))
 - [Android] Fixed hosted Compose views missing layout after reattachment or in-place configuration changes. ([#48370](https://github.com/expo/expo/issues/48370) by [@lujjjh](https://github.com/lujjjh))
 - [iOS] Fixed infinite recursion and `EXC_BAD_ACCESS` when encoding native `Date` values to JavaScript. ([#48239](https://github.com/expo/expo/issues/48239), [#48240](https://github.com/expo/expo/pull/48240) by [@samuelcorsan](https://github.com/samuelcorsan))
--- a/packages/expo-modules-core/ios/Fabric/ExpoFabricView.swift
+++ b/packages/expo-modules-core/ios/Fabric/ExpoFabricView.swift
@@ -205,50 +205,20 @@ open class ExpoFabricView: ExpoFabricViewObjC, AnyExpoView {
     class_replaceMethod(object_getClass(ExpoFabricView.self), #selector(appContextFromClass), appContextBlockImp, "@@:")
   }
 
-  /**
-   Keeps a strong reference to the app context that the injected initializer builds the views from.
-   The view classes are cached in `viewClassesRegistry` for the entire process lifetime, so they
-   outlive the app context they were created with. With the context captured weakly, Fabric could
-   instantiate a cached class at a moment when that reference was already `nil` — on a reload,
-   after the previous context has been released but before `registerNativeViews` runs again on the
-   main actor — and the initializer aborted the process. Replacing the reference on each
-   registration keeps the previous context alive across that window, and releases it right after.
-   */
-  private final class AppContextHolder {
-    var appContext: AppContext
-
-    init(_ appContext: AppContext) {
-      self.appContext = appContext
-    }
-  }
-
-  /**
-   Holders of the app context, keyed by the identity of the view class the initializer was injected to.
-   Not keyed by the class name, unlike `viewClassesRegistry` — there might be more than one class
-   with the same name (Expo Go), and they need an initializer each.
-   Used from the main thread only: `makeViewClass` is called by `registerNativeViews` (`@MainActor`)
-   and the injected initializer runs while Fabric is performing a mounting transaction.
-   */
-  private static var appContextHolders = [ObjectIdentifier: AppContextHolder]()
-
   internal static func injectInitializer(appContext: AppContext, moduleName: String, viewName: String, toViewClass viewClass: AnyClass) {
-    let holderKey = ObjectIdentifier(viewClass)
-
-    // A cached class already has the initializer injected, and `moduleName` and `viewName` are part
-    // of the class name, so they can't have changed. Only the app context has — point the existing
-    // initializer at the new one instead of injecting another.
-    if let holder = appContextHolders[holderKey] {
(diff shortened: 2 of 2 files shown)


brentvatne:
@expo-bot check

uwayss:
You're right about the holder, and my testing missed it. Every trial I ran ended in another
reload, which is the one path where the entry does get replaced. A teardown with nothing after it
pins the context, and since `.appContextDestroys` only fires from `deinit`, `OnDestroy` stops
running everywhere.

Worth flagging before anyone tries it: clearing the strong slot from `AppContext.destroy()` puts
the crash back. `destroy()` is called from `NativeState`'s deallocator, which runs as the old
runtime finalizes. That's the start of the window, before the new context has even queued its
main-actor hop.

So I've gone the other way and dropped the retention. The context stays weak. The initializer just
doesn't abort anymore when there's nothing to build from, it logs and returns an inert view of the
right class. That surface is on its way out regardless, and it's where the holder version ended up
anyway, since the view goes inert as soon as the holder is repointed. Ownership doesn't change and
the diff is about a third of the size.

`init(appContext:)` already takes nil, and `updateProps`, `viewDidUpdateProps` and `supportsProp`
all guard for it, so the inert view goes through paths that already exist.

On the repro not firing: you reloaded with `DevSettings.reload()`. When I logged `AppContext`
init and deinit in my app, a JS reload never built a new context at all. `registerNativeModules`
fired once across 233 Metro reloads, so nothing was released and the weak reference never went
nil. `Updates.reloadAsync()` was the only thing that produced a new context and deinit'd the old
one. Might be worth trying that in the sandbox app.

And yes to both of the smaller ones. The `AppContextLost` message doesn't match what it still
guards, and I didn't do the registry serialization #48575 asked for. Either can land here or as a
follow-up, whichever you'd rather.

uwayss:
Same setup as before. With the window widened, the branch that used to `fatalError` ran 1751 times across 10 trials and nothing crashed. Baseline was 4 of 5 aborting. Same views involved both times, `ExpoImage` and one of the app's own view modules.

The leak side is by inspection rather than measurement: the `weak` capture is untouched, so nothing retains the context.

uwayss:
Closing this PR since it hasn't been reviewed and I'm removing my fork.

Issue #4180, opened 2019-05-10
[expo-image-manipulator] ImageManipulator.manipulateAsync is not available on ios


donysukardi:
## 🐛 Bug Report

Expo libraries don't seem to get linked in Release build for iOS.

### Environment

Target: iOS Bare
expo-image-manipulator@4.0.0

### Steps to Reproduce

- Follow instructions to install `react-native-unimodules` and then install `expo-image-manipulator`.
- Call `ImageManipulator.manipulateAsync` in code with a given uri

### Expected Behavior

`ImageManipulator.manipulateAsync` should work on iOS

### Actual Behavior

Getting the following error

`The method or property ImageManipulator.manipulateAsync is not available on ios, are you sure you've linked all the native dependencies properly?`

![image](https://user-images.githubusercontent.com/410792/57503911-e33db600-7324-11e9-9222-aadf8e811f57.png)

sjchmiela:
Hey, could you please share your `Podfile.lock` to let us ensure that the manipulator is getting installed properly?

donysukardi:
@sjchmiela it happens for `expo-file-system` apparently too. here are my current config https://gist.github.com/donysukardi/14776f9b0ef906563e45853ab14f7668

sjchmiela:
How about if you tried moving line 41 (`self.moduleRegistryAdapter = …`) to the beginning of `application:didFinishLaunchingWithOptions:`? (Most specifically, before initializing the bridge in line 28.)

donysukardi:
That seems to do the trick. Thank you so much @sjchmiela! I spent 2 days trying to figure out what was wrong 😱

PR #48938, opened 2026-08-14
feat(env): Add internal config mode helper

ramonclaudio:
# Why

The EAS changes in [#4180](https://github.com/expo/eas-cli/pull/4180) tell Expo to use development or production mode when loading app config, so we needed a temporary internal environment variable that Expo reads and removes before app config loads.

# How

I added `__EXPO_CONFIG_MODE` as an internal handoff that `@expo/env` reads and removes before Expo loads the dotenv files and app config. I blocked `.env` files from setting the handoff and updated `getOriginalEnv()` and `getOriginalEnvValue()` to exclude dotenv values inherited from a parent process. We still use `EAS_BUILD` as the production fallback for older EAS versions.

# Test Plan

Tests and package checks pass.

# Checklist

### DIFF
--- a/packages/@expo/cli/CHANGELOG.md
+++ b/packages/@expo/cli/CHANGELOG.md
@@ -38,7 +38,7 @@
 
 ### 💡 Others
 
-- [Internal] Use `@expo/env` to read `EXPO_CONFIG_MODE` and keep older EAS Build versions in production mode. ([#48938](https://github.com/expo/expo/pull/48938) by [@ramonclaudio](https://github.com/ramonclaudio))
+- [Internal] Use `@expo/env` to read the EAS config mode and keep older EAS Build versions in production mode. ([#48938](https://github.com/expo/expo/pull/48938) by [@ramonclaudio](https://github.com/ramonclaudio))
 - Add sandbox detection to telemetry context ([#47928](https://github.com/expo/expo/pull/47928) by [@davidmokos](https://github.com/davidmokos))
 - [Internal] Remove the unreachable port fallbacks and increase consistency in port selection logic ([#47771](https://github.com/expo/expo/pull/47771) by [@kitten](https://github.com/kitten))
 - Add experimental `tvos` and `macos` autolinking gated by `expriments.outOfTreePlatforms` ([#46344](https://github.com/expo/expo/pull/46344) by [@kitten](https://github.com/kitten))
--- a/packages/@expo/cli/CHANGELOG.md
+++ b/packages/@expo/cli/CHANGELOG.md
@@ -38,6 +38,7 @@
 
 ### 💡 Others
 
+- [Internal] Use `@expo/env` to read `EXPO_CONFIG_MODE` and keep older EAS Build versions in production mode. ([#48938](https://github.com/expo/expo/pull/48938) by [@ramonclaudio](https://github.com/ramonclaudio))
 - Add sandbox detection to telemetry context ([#47928](https://github.com/expo/expo/pull/47928) by [@davidmokos](https://github.com/davidmokos))
 - [Internal] Remove the unreachable port fallbacks and increase consistency in port selection logic ([#47771](https://github.com/expo/expo/pull/47771) by [@kitten](https://github.com/kitten))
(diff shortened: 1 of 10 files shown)


Kudo (approved):
code change looks good to me but i have no idea where's EXPO_CONFIG_MODE being set. will leave the review for others

ramonclaudio:
> code change looks good to me but i have no idea where's EXPO_CONFIG_MODE being set. will leave the review for others

Sorry @Kudo, the EAS changes in [#4180](https://github.com/expo/eas-cli/pull/4180) will set `EXPO_CONFIG_MODE` before EAS runs config, prebuild, or doctor. Expo CLI reads it after #48741, and doctor will read it after #48845.

Kudo:
> the EAS changes in [#4180](https://github.com/expo/eas-cli/pull/4180) will set `EXPO_CONFIG_MODE` before EAS runs config, prebuild, or doctor. Expo CLI reads it after #48741, and doctor will read it after #48845.

@ramonclaudio thanks for sharing the context. a dumb question: what would happen if we don't ever unset `EXPO_CONFIG_MODE`?
i probably don't really like the inconsistency where set an env at placeA and unset it at placeB. maybe setup the env and expect it won't be changed/unset anywhere. not sure if that's a good idea

brentvatne:
how will this all work with older versions of eas-cli?

ramonclaudio:
Yea I agree @Kudo. Deleting it in every caller was confusing, so I changed it to use a shared `consumeConfigEnvMode` helper that reads and removes it once. FWIW we still remove it because it is only meant to pass the mode into the tool, and we don't want app config to inherit it. I also blocked `EXPO_CONFIG_MODE` from `.env` files to prevent users from setting it accidentally.

ramonclaudio:
> how will this all work with older versions of eas-cli?

@brentvatne the changes I made in [#4180](https://github.com/expo/eas-cli/pull/4180) will pass `EXPO_CONFIG_MODE=production` to app config, prebuild, and doctor. Older versions of eas-cli won't pass it, but they already run builds with `EAS_BUILD` set, so Expo CLI falls back to production.

kitten:
I think, correct me if I'm wrong the chain here is:
- `NODE_ENV` is force-set, as intended
- We need overrides in some places for EAS
- We allow the overrides, but either propagating or not could both be argue to be correct
- If we allow `NODE_ENV` for select commands (`expo config`) to then be variable, but only if it's explicitly set, `EXPO_CONFIG_MODE` must be treated as the non-distributing equivalent (set top-level process but not below)

As in, we're replacing one variable mode (`NODE_ENV`) with another and not distributing it to limit the "impact" of this override.

The question though then is, if this is now causing confusion (even with the utility) do we even want that override or need it for EAS? And if it's only used, say, for `expo config` would that be replaceable with one-off logic there?

We could do a one-off for `expo config` where we inherit `NODE_ENV` if it's explicitly set (since it's a "leaf command") or designate the env var as internal to indicate it's not a public API to limit its blast radius. The latter requires no large changes.

tl;dr: Change checks out to me but do we want to rename the env var for clarity, or even consider a new argument on `expo config`?

ramonclaudio:
Renamed the handoff to `__EXPO_CONFIG_MODE` and kept it scoped to the immediate child process, so it is clearly internal and is removed before app config loads. I kept the env-var path because EAS needs the same handoff for config, prebuild, Doctor, and Expo Updates commands, while `NODE_ENV` still covers older Expo versions.

ramonclaudio:
> tl;dr: Change checks out to me but do we want to rename the env var for clarity, or even consider a new argument on `expo config`?

Yea, exactly @kitten. I went with the internal env var option you suggested. I also moved the handoff logic into the shared helper in `@expo/env`. We now read and remove the handoff before it loads the dotenv files or app config. The var is also blocked in `.env` files, so dotenv can't reintroduce it.

I went with the shared handoff instead of adding a new `expo config` arg because EAS also needs to select the mode for prebuild, doctor, and Expo updates, and EAS has to work with whatever Expo CLI version the project has installed ATM, so this can get messy.

The changes we made to EAS pass `NODE_ENV` and `__EXPO_CONFIG_MODE` to each selected child process. That keeps older Expo CLI versions that use `NODE_ENV` working, and dep installs still get the original EAS env. We also use `EAS_BUILD` as the production fallback for older EAS versions.

So I think the internal handoff is the smallest and best option here. WDYT?

kitten (approved):

PR #46679, opened 2026-06-09
[expo-codemod] Fix react-navigation codemod skipping migratable imports in mixed files

ahmdshrif:
# Why

Fixes #46481.

The `sdk-56-expo-router-react-navigation-replace` codemod is supposed to rewrite every `@react-navigation/{native,bottom-tabs,material-top-tabs,stack,...}` import to its `expo-router/*` equivalent. On real codebases it silently leaves a large share of imports unmigrated.

Root cause: the transform **bails out of the entire file** as soon as it sees *any* unsupported import:

```ts
if (unsupportedPackageErrors.length || errors.length) {
  return undefined; // skips the whole file
}
```

Unsupported imports include `@react-navigation/native-stack` and `@react-navigation/drawer` (which the vast majority of apps use) and default/namespace imports. So any file that also imports `native-stack` has its perfectly migratable `@react-navigation/native` named imports left untouched — with no error pointing at them. The result compiles and bundles, but loads a second copy of `@react-navigation`, crashing at navigator registration:

> Couldn't register the navigator. … This can also happen if there are multiple copies of '@react-navigation' packages installed.

# How

Don't abort the whole file. Migrate the clean named imports, and only skip the *specific* imports that can't be rewritten (default/namespace style). Unsupported packages (`native-stack`/`drawer`) and unsupported styles are still reported for manual migration and left untouched — exactly as before — but they no longer suppress the migration of unrelated, migratable imports in the same file.

The change is localized to the transform's rewrite step; the warning/reporting behavior is unchanged.

# Test Plan

`et test expo-codemod` (run locally via an isolated jscodeshift + jest harness): **51 passed**.

- All existing tests still pass — the previous behavior was only exercised with *single-import* files, which still short-circuit to "no changes" (now via "nothing was rewritten") and still print their warnings.
- Updated the one mixed-file test (`native` + `drawer`) to assert the new, correct behavior: the supported import is migrated **and** the unsupported package is reported.
- Added tests for the previously-broken mixed cases:
  - migratable named import alongside `@react-navigation/native-stack` → named import migrated, `native-stack` reported + left untouched
  - clean named import alongside a default-style import → named import migrated, default-style import reported + left untouched

# Checklist

- [x] I added a [changelog entry](https://github.com/expo/expo/blob/main/guides/contributing/Updating%20Changelogs.md).
- [x] Tested the changes.
- [x] Updated the compiled `build/` output.

PR #48941, opened 2026-08-14
[expo-codemod] Fix react-navigation codemod skipping migratable imports in mixed files

### LINKED ISSUE #46481: [expo-codemod] sdk-56-expo-router-react-navigation-replace misses imports in some files (~25% of import sites)
## Summary

`npx expo-codemod sdk-56-expo-router-react-navigation-replace 'app/**/*.{ts,tsx}'` should rewrite all `@react-navigation/{native,bottom-tabs,material-top-tabs}` imports to the matching `expo-router/*` entry points (per the [SDK 55→56 migration guide](https://docs.expo.dev/router/migrate/sdk-55-to-56/)). On a production codebase it leaves ~25% of import sites unmigrated, with no warning. The result compiles and bundles, but introduces a subtle two-copies-of-react-navigation runtime bug that surfaces only at navigator registration.

## Real-world numbers

Codemod run on https://github.com/livspaceeng/super-app-react-native — a real production RN/Expo app (~70 screens, 99 `@react-navigation/*` import sites across 69 files).

**Reported by codemod:** `0 errors, 910 skipped, 24 ok` ← only 24 rewritten

**Actually present after the codemod:** 24 remaining `@react-navigation/*` imports the codemod silently missed:

| Source | Sites missed |
|---|---|
| `@react-navigation/native` | 21 |
| `@react-navigation/material-top-tabs` | 2 |
| `@react-navigation/bottom-tabs` | 1 |
| **Total** | **24** (vs 24 successfully migrated → ~50% miss rate, not 25%) |

We caught these by grepping. They compile and lint fine. Runtime crashes at first navigator mount with:

> Couldn't register the navigator. Have you wrapped your app with 'NavigationContainer'?
> This can also happen if there are multiple copies of '@react-navigation' packages installed.

Root cause is that `NavigationContainer` (from `expo-router/react-navigation`, codemod-migrated) and `useNavigation`/`useRoute` (still from `@react-navigation/native`, codemod-missed) end up reading two different `SingleNavigatorContext` instances.

## Affected file patterns

Spot-checked a few of the missed files. No obvious commonality:
- `app/screens/CategoryListingContent.tsx` — top-level `import { useNavigation, useRoute } from "@react-navigation/native"`
- `app/screens/LivspaceTv/LivspaceTvDetailScreen.tsx` — top-level `import { useNavigation } from "@react-navigation/native"`
- `app/screens/Magazine/MagazineLandingScreen.tsx` — `import { createMaterialTopTabNavigator } from "@react-navigation/material-top-tabs"`
- `app/screens/ProjectDetail/.../bank/BankStatusScreen.tsx` — `import { RouteProp, useNavigation, useRoute } from "@react-navigation/native"`

All are standard top-of-file imports. No conditional imports, no string-template paths, nothing exotic. Many sit next to imports that DID get migrated correctly, in the same file. So the codemod is iterating files but missing certain `ImportDeclaration` nodes.

## Test plan / repro

1. Branch a real SDK 55 → 56 upgrade
2. Run the codemod
3. `grep -rn 'from "@react-navigation/' src/ --include="*.ts" --include="*.tsx"` — should be empty if codemod is complete; in practice ~24 hits
4. Build + boot the app → `Couldn't register the navigator` at first mount

## Suggested fix

1. **Fail loudly:** if the codemod skips a file containing matchable imports, log a warning per skip. Currently the only output is the aggregate `910 skipped` which conflates "no matches" with "had matches but skipped them."
2. **Or: print the final unmigrated count.** A `grep`-style scan after the rewrite pass would catch this: `Found N remaining @react-navigation imports in N files — these need manual review`.
3. **Better diagnostic for the "two copies" runtime error.** Even if the codemod is fixed, users will land here from other sources (e.g., transitive deps importing `@react-navigation/native`). The current error message hints at "multiple copies installed" but doesn't say where to look. Pointing them at expo-router's diagnostic tools (or the Metro resolver inspection) would shave hours off debug time.

## Reference PRs

- [#45546](https://github.com/expo/expo/pull/45546) added the codemod + Metro check for native-stack/drawer. The Metro check is loud (throws at bundle time, points at the offending import). The codemod equivalent is the missing piece.

## Workaround for affected users

Until the codemod is fixed, after running it:

```bash
grep -rn 'from "@react-navigation/' app/ src/ --include="*.ts" --include="*.tsx"
```

Manually rewrite anything that shows up:
- `@react-navigation/native` → `expo-router/react-navigation`
- `@react-navigation/bottom-tabs` → `expo-router/js-tabs`
- `@react-navigation/material-top-tabs` → `expo-router/js-top-tabs`

Same for test mocks (`jest.mock("@react-navigation/native", ...)`).

ahmdshrif:
# Why

Fixes #46481. Split out of #46679 as requested by @Ubax — this PR contains only the bug fix; the named re-export / `jest.mock` reporting work is in a separate PR.

The `sdk-56-expo-router-react-navigation-replace` codemod is supposed to rewrite every `@react-navigation/{native,bottom-tabs,material-top-tabs,stack,...}` import to its `expo-router/*` equivalent. On real codebases it silently leaves a large share of imports unmigrated.

Root cause: the transform **bails out of the entire file** as soon as it sees *any* unsupported import:

```ts
if (unsupportedPackageErrors.length || errors.length) {
  return undefined; // skips the whole file
}
```

Unsupported imports include `@react-navigation/native-stack` and `@react-navigation/drawer` (which the vast majority of apps use) and default/namespace imports. So any file that also imports `native-stack` has its perfectly migratable `@react-navigation/native` named imports left untouched — with no error pointing at them. The result compiles and bundles, but loads a second copy of `@react-navigation`, crashing at navigator registration:

> Couldn't register the navigator. … This can also happen if there are multiple copies of '@react-navigation' packages installed.

# How

Don't abort the whole file. Migrate the clean named imports, and only skip the *specific* imports that can't be rewritten (default/namespace style). Unsupported packages (`native-stack`/`drawer`) and unsupported styles are still reported for manual migration and left untouched — exactly as before — but they no longer suppress the migration of unrelated, migratable imports in the same file.

A `didRewrite` flag preserves the previous "no changes" return for files where nothing could be rewritten.

The change is localized to the transform's rewrite step; the warning/reporting behavior is unchanged.

# Test Plan

`et test expo-codemod` (run locally via an isolated jscodeshift + jest harness): **51 passed**.

- All existing tests still pass — the previous behavior was only exercised with *single-import* files, which still short-circuit to "no changes" (now via "nothing was rewritten") and still print their warnings.
- Updated the one mixed-file test (`native` + `drawer`) to assert the new, correct behavior: the supported import is migrated **and** the unsupported package is reported.
- Added tests for the previously-broken mixed cases:
  - migratable named import alongside `@react-navigation/native-stack` → named import migrated, `native-stack` reported + left untouched
  - clean named import alongside a default-style import → named import migrated, default-style import reported + left untouched

# Checklist

- [x] I added a [changelog entry](https://github.com/expo/expo/blob/main/guides/contributing/Updating%20Changelogs.md).
- [x] Tested the changes.

### DIFF
--- a/packages/expo-codemod/CHANGELOG.md
+++ b/packages/expo-codemod/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ### 🐛 Bug fixes
 
-- Fix `sdk-56-expo-router-react-navigation-replace` silently skipping migratable `@react-navigation/*` imports when a file also contained an unsupported import (e.g. `@react-navigation/native-stack`). Supported named imports are now migrated while unsupported ones are still reported for manual migration. ([#TBD](https://github.com/expo/expo/pull/TBD) by [@ahmdshrif](https://github.com/ahmdshrif))
+- Fix `sdk-56-expo-router-react-navigation-replace` silently skipping migratable `@react-navigation/*` imports when a file also contained an unsupported import (e.g. `@react-navigation/native-stack`). Supported named imports are now migrated while unsupported ones are still reported for manual migration. ([#48941](https://github.com/expo/expo/pull/48941) by [@ahmdshrif](https://github.com/ahmdshrif))
 
 ### 💡 Others
 
--- a/packages/expo-codemod/CHANGELOG.md
+++ b/packages/expo-codemod/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### 🐛 Bug fixes
 
+- Fix `sdk-56-expo-router-react-navigation-replace` silently skipping migratable `@react-navigation/*` imports when a file also contained an unsupported import (e.g. `@react-navigation/native-stack`). Supported named imports are now migrated while unsupported ones are still reported for manual migration. ([#TBD](https://github.com/expo/expo/pull/TBD) by [@ahmdshrif](https://github.com/ahmdshrif))
+
 ### 💡 Others
 
 ## 57.0.0 - 2026-06-25
--- a/packages/expo-codemod/src/transforms/__tests__/sdk-56-expo-router-react-navigation-replace-test.ts
+++ b/packages/expo-codemod/src/transforms/__tests__/sdk-56-expo-router-react-navigation-replace-test.ts
@@ -393,16 +393,20 @@ describe('unsupported packages (no direct equivalent)', () => {
     );
   });
 
-  test('throws even when other react-navigation imports would otherwise be rewritten', () => {
+  test('reports the unsupported package but still migrates the supported import in the same file', () => {
     const input = [
       `import { useNavigation } from '@react-navigation/native';`,
       `import { createDrawerNavigator } from '@react-navigation/drawer';`,
     ].join('\n');
-    run(input);
+    const output = run(input);
+    // The unsupported package is reported for manual migration...
     expect(errorSpy).toHaveBeenCalledTimes(1);
     expect(errorSpy.mock.calls[0]).toHaveLength(1);
     const message = errorSpy.mock.calls[0]![0] as string;
     expect(message).toContain('import from "@react-navigation/drawer" cannot be migrated');
+    // ...but the supported import is still migrated (rather than silently skipped).
+    expect(output).toContain(`import { useNavigation } from "expo-router/react-navigation"`);
+    expect(output).toContain(`import { createDrawerNavigator } from '@react-navigation/drawer'`);
   });
 });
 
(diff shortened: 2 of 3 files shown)


brentvatne:
@expo-bot check

ahmdshrif:
Thanks @brentvatne for running the check, and thanks to the reviewer for the trace.

The finding is right about one thing and I've fixed it in 881d1f5: the comment I added claimed that leaving the imports untouched loads a second copy of react-navigation. That's backwards. `expo-router/react-navigation` is expo-router's own bundled copy (`expo-router/src/react-navigation/`, no `@react-navigation/*` dependency in its `package.json`), so it's the partial rewrite that mixes two copies, not the untouched state. The comment now says what actually justifies the change.

On the suggested remediation — restoring the file-level bail — I'd argue against it, for a reason the review's own trace notes: `withMetroMultiPlatform.ts` throws on **any** `@react-navigation/*` import from app code once expo-router is installed (with a dedicated message for `native-stack`/`drawer`). So a file with a leftover unsupported import doesn't bundle either way:

- before this PR: every import in the file stays on `@react-navigation/*` → Metro throws, and the migratable ones were never reported
- after this PR: the migratable ones are migrated, the rest are reported → Metro still throws until the reported manual migration is done

The mixed-copies runtime scenario needs `EXPO_ROUTER_DISABLE_RN_NAVIGATION_CHECK=1` to be reachable at all. Restoring the bail would bring back the silent skip in #46481 — the whole reason for the fix — to protect a state that already fails the build. Happy to reconsider if a maintainer sees it differently.

I've also updated the reporting message wording in #48942, which is where the same incorrect "re-export" claim was user-visible.

Two follow-ups from the verify run, if wanted here rather than as separate issues:

1. The other silent-skip classes it measured (`devtools`, `material-bottom-tabs`, subpath imports, `require`) are still open. `export ... from` and `jest.mock` are covered by #48942.
2. The untested shape it flagged — same module imported as both a default and a named import — I'm happy to add as a test here; say the word.

ahmdshrif:
Small follow-up: the `@expo-bot verify` run on the stacked #48942 caught a `noUncheckedIndexedAccess` failure that includes one line from this PR — line 588 read `errorSpy.mock.calls[0][0]` without the `!` the older assertions in this file use. Fixed in 2cab968.

Verified locally: `TS2532` reproduces at that line under `strict` + `noUncheckedIndexedAccess` before the change, and the branch typechecks clean after it, with all 51 tests still passing.

Ubax:
@expo-bot review

PR #48942, opened 2026-08-14
[expo-codemod] Migrate named re-exports and report jest module mocks in sdk-56 react-navigation codemod

ahmdshrif:
# Why

Split out of #46679 as requested by @Ubax. This is the second of the two changes; the bug fix is in #48941.

> **Stacked on #48941.** This branch is built on top of it, so the diff here includes that commit until it merges. The changes belonging to *this* PR are in the second commit only — [compare against the base branch](https://github.com/ahmdshrif/expo/compare/split-a-mixed-file-bailout...split-b-reexports-jest-mocks). Happy to rebase once #48941 lands.

Two site types referencing `@react-navigation/*` are invisible to the codemod today:

- **Named re-exports** — `export { NavigationContainer } from '@react-navigation/native';` is exactly as rewritable as the equivalent named import, but is left pointing at `@react-navigation/*`. Barrel files are a common place for these, so a single missed re-export can pull a second copy of `@react-navigation` into the app.
- **Jest module calls** — `jest.mock('@react-navigation/native')` and friends are never reported, so after a migration the test file silently keeps mocking a module the code no longer imports.

# How

- Named re-exports (`export { A } from '...'`) are rewritten like named imports. `export * from '...'` cannot be — the namespace shape may not match the expo-router entry point, same reason namespace imports are unsupported — so it is reported and left untouched.
- Unsupported-package reporting now covers re-exports and jest calls too, not just imports, and each message names the kind of site it found.
- `jest.mock` / `jest.doMock` / `jest.requireActual` / `jest.unmock` are **reported, never rewritten**, per @Ubax's review on #46679. `jest.mock` keys the module registry on the *resolved* module, so it isn't the same mapping as an import rewrite: `expo-router/react-navigation` re-exports `@react-navigation/*`, so moving the mock to the expo-router entry point would leave the underlying module unmocked for any code that reaches it directly — the test would silently exercise the real implementation. Mocking the underlying package can also be exactly what a test intends. Which one is right depends on the test, so the call is surfaced and the author decides:

```
Jest module mocks — manual review needed

  app/foo.tsx:2 - jest.mock("@react-navigation/native") was left unchanged.
  Mocking "expo-router/react-navigation" instead is not equivalent: it re-exports
  "@react-navigation/native", so the underlying module would stay unmocked for code that
  imports it directly. Update the mock by hand if your test needs it.
```

# Test Plan

`et test expo-codemod` (run locally via an isolated jscodeshift + jest harness): **60 passed** (51 from #48941 plus 9 added here).

- `re-exports`: named re-export rewritten; type-only named re-export rewritten; re-export alongside a named import of the same module; `export *` reported and left untouched; re-export from an unsupported package reported and left untouched.
- `jest module calls`: `jest.mock` reported while the import in the same file is still migrated; `jest.requireActual` inside a mock factory reported; `jest.mock` of an unsupported package reported; unrelated `jest.mock` calls untouched.

# Checklist

- [x] I added a [changelog entry](https://github.com/expo/expo/blob/main/guides/contributing/Updating%20Changelogs.md).
- [x] Tested the changes.

### DIFF
--- a/packages/expo-codemod/CHANGELOG.md
+++ b/packages/expo-codemod/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ### 🐛 Bug fixes
 
-- Fix `sdk-56-expo-router-react-navigation-replace` silently skipping migratable `@react-navigation/*` imports when a file also contained an unsupported import (e.g. `@react-navigation/native-stack`). Supported named imports are now migrated while unsupported ones are still reported for manual migration. ([#TBD](https://github.com/expo/expo/pull/TBD) by [@ahmdshrif](https://github.com/ahmdshrif))
+- Fix `sdk-56-expo-router-react-navigation-replace` silently skipping migratable `@react-navigation/*` imports when a file also contained an unsupported import (e.g. `@react-navigation/native-stack`). Supported named imports are now migrated while unsupported ones are still reported for manual migration. ([#48941](https://github.com/expo/expo/pull/48941) by [@ahmdshrif](https://github.com/ahmdshrif))
 
 ### 💡 Others
 
--- a/packages/expo-codemod/CHANGELOG.md
+++ b/packages/expo-codemod/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### 🐛 Bug fixes
 
+- Fix `sdk-56-expo-router-react-navigation-replace` silently skipping migratable `@react-navigation/*` imports when a file also contained an unsupported import (e.g. `@react-navigation/native-stack`). Supported named imports are now migrated while unsupported ones are still reported for manual migration. ([#TBD](https://github.com/expo/expo/pull/TBD) by [@ahmdshrif](https://github.com/ahmdshrif))
+
 ### 💡 Others
 
 ## 57.0.0 - 2026-06-25
--- a/packages/expo-codemod/CHANGELOG.md
+++ b/packages/expo-codemod/CHANGELOG.md
@@ -11,6 +11,7 @@
 ### 🐛 Bug fixes
 
 - Fix `sdk-56-expo-router-react-navigation-replace` silently skipping migratable `@react-navigation/*` imports when a file also contained an unsupported import (e.g. `@react-navigation/native-stack`). Supported named imports are now migrated while unsupported ones are still reported for manual migration. ([#48941](https://github.com/expo/expo/pull/48941) by [@ahmdshrif](https://github.com/ahmdshrif))
+- Fix `sdk-56-expo-router-react-navigation-replace` silently skipping named re-exports (`export { A } from '@react-navigation/...'`), which are now rewritten like named imports. Re-export styles that cannot be rewritten (`export *`, `export * as ns from`, `export v from`, `export { default as X } from`) `jest` module calls (`mock`, `doMock`, `unmock`, `setMock`, `requireActual`, `requireMock`, `createMockFromModule`) and inline `import("...")` types referencing `@react-navigation/*` modules are reported for manual migration. ([#48942](https://github.com/expo/expo/pull/48942) by [@ahmdshrif](https://github.com/ahmdshrif))
 
 ### 💡 Others
 
--- a/packages/expo-codemod/src/transforms/__tests__/sdk-56-expo-router-react-navigation-replace-test.ts
+++ b/packages/expo-codemod/src/transforms/__tests__/sdk-56-expo-router-react-navigation-replace-test.ts
@@ -393,16 +393,20 @@ describe('unsupported packages (no direct equivalent)', () => {
     );
   });
 
-  test('throws even when other react-navigation imports would otherwise be rewritten', () => {
+  test('reports the unsupported package but still migrates the supported import in the same file', () => {
     const input = [
       `import { useNavigation } from '@react-navigation/native';`,
       `import { createDrawerNavigator } from '@react-navigation/drawer';`,
     ].join('\n');
-    run(input);
+    const output = run(input);
+    // The unsupported package is reported for manual migration...
     expect(errorSpy).toHaveBeenCalledTimes(1);
     expect(errorSpy.mock.calls[0]).toHaveLength(1);
     const message = errorSpy.mock.calls[0]![0] as string;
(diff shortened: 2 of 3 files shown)


brentvatne:
@expo-bot check

ahmdshrif:
Thanks @brentvatne for running the check — both findings were correct and are fixed in 0650e3a.

**1. The jest guidance was wrong.** I had written that `expo-router/react-navigation` re-exports `@react-navigation/*`. It doesn't: `packages/expo-router/package.json` declares no `@react-navigation/*` dependency, and `./react-navigation` resolves to `src/react-navigation/index.ts`, which re-exports expo-router's own vendored copy. The decision not to rewrite the mock stands, but the stated reason was misleading in exactly the way the review describes. The message now reads:

```
app/foo.tsx:2 - jest.mock("@react-navigation/native") was left unchanged.
"expo-router/react-navigation" is expo-router's own bundled copy of React
Navigation, not a re-export of "@react-navigation/native", so this mock no longer applies
to the imports migrated in this file. Point it at the expo-router module by hand
if the test still needs a mock.
```

The comment above `JEST_MODULE_METHODS` and the stale test comment were corrected the same way.

**2. `export * as ns from` was silently skipped.** Confirmed, and it also applies to `export v from`. I checked all three parsers the runner actually uses (`ts`, `tsx`, `jsx` — see `resolveAndDispatch`):

| syntax | `ts` / `tsx` | `jsx` |
| - | - | - |
| `export * as ns from '...'` | `ExportNamedDeclaration` + `ExportNamespaceSpecifier` — **was silent** | `ExportAllDeclaration` — already reported |
| `export v from '...'` | `ExportNamedDeclaration` + `ExportDefaultSpecifier` — **was silent** | parse error |

So the gap was real for `.ts`/`.tsx`, which is where most of these files live. Both styles are now reported through `UNSUPPORTED_EXPORT_SPECIFIERS` and left untouched, with a test each (62 tests total, up from 60). The changelog entry was updated to match.

This PR is still stacked on #48941, which also picked up a comment-wording fix from its own review.

brentvatne:
@expo-bot verify with fable

brentvatne:
@expo-bot review

ahmdshrif:
Thanks @brentvatne, and thanks for the precise line list. The blocker is fixed and I've taken two of the three items for @Ubax as well.

**Typecheck blocker.** `noUncheckedIndexedAccess` was tripping on the newer assertions that read `errorSpy.mock.calls[0][0]` without the `!` the 13 older ones use. Fixed for the seven lines that belong to this PR (640, 654, 669, 681, 700, 714, 728), and separately in 2cab968 on #48941 for line 588. I reproduced the failure locally first — `TS2532` at each of those lines under `strict` + `noUncheckedIndexedAccess` — then confirmed both branches typecheck clean.

**Item 1 — `export { default as X } from '@react-navigation/native'`.** Agreed, and fixed: it was accepted by `isNamedReExport` because the specifier is an ordinary `ExportSpecifier`, so it got rewritten while the equivalent `import X from` is refused. Re-exporting the source module's default is unsupported for the same reason, so it now goes down the same reporting path and is left untouched. The three unrewritable re-export styles are recognised in one place now (`unsupportedReExportLabel`): `export * as ns from`, `export v from`, and `export { default as X } from`. A declaration that mixes a default re-export with a named one is reported and left alone as a whole, the same way one default specifier disqualifies an import declaration.

**Item 2 — jest methods.** Added `setMock`, `requireMock` and `createMockFromModule` to `JEST_MODULE_METHODS`, alongside the existing `mock`, `doMock`, `unmock` and `requireActual`.

**Item 3 — `NavigationContainer`.** Agreed, and it is pre-existing: `IMPORT_MAP` maps `@react-navigation/native` → `expo-router/react-navigation`, and the suite on `main` already uses `NavigationContainer` as its example symbol throughout, so the new tests followed that convention rather than asserting anything new about the mapping. Worth its own issue if the mapping needs revisiting — happy to switch the test symbols to something the module does export if you'd rather not have the shape repeated here.

New head is 5762930. 67 tests pass (up from 62 — five new: three jest methods, `export { default as X }`, and the mixed default/named declaration), `tsc` is clean under the package's own settings, and `oxfmt --check` is clean.

jerone:
Does this also migrate typed jest methods?:

```ts
jest.requireActual<typeof import("@react-navigation/native")>("@react-navigation/native");
```

ahmdshrif:
Good question — yes, the type argument doesn't affect detection. The check matches on the callee being a `jest.*` member expression whose first argument is a string, so the type parameter is irrelevant to it. I ran your exact snippet through the transform with the `tsx` parser to be sure:

```
this/is/test.tsx:1 - jest.requireActual("@react-navigation/native") was left unchanged.
```

Worth being precise about "migrate" though: jest module calls are **never** rewritten by this PR, only reported. That was @Ubax's call on the original PR (#46679), and the reasoning holds — `expo-router/react-navigation` is expo-router's own bundled copy rather than a re-export, so only the test author can say whether the mock should follow the imports, stay where it is, or go away.

Your example did surface a real gap, though: the `import("@react-navigation/native")` **type** reference was neither rewritten nor reported. That was inconsistent, since `import type { NavigationProp } from '@react-navigation/native'` *is* rewritten today. Fixed in 8da5e6d — inline `import("...")` types are now reported like the other unrewritable sites, and left untouched:

```
app/foo.tsx:2 - inline type import (import("@react-navigation/native")) is not supported.
Only named imports and named re-exports can be rewritten by this codemod.
Replace it with an `import type { ... } from` declaration and re-run the codemod.
```

Inline types from `native-stack`/`drawer` land in the existing "Migration required" block instead, and unrelated modules are still left alone. Four tests added, 71 total.

One limitation I could not close, and it's worth knowing about: ast-types does not descend into the **type arguments of a call expression**, so the `import("...")` inside `jest.requireActual<typeof import("...")>(...)` or a plain `foo<import("...")>()` is not visited by the traversal. For the jest form that costs nothing — the call itself is already reported on that line — but a generic non-jest call carrying an `import("@react-navigation/*")` type argument would still pass silently. Closing that needs a manual walk over `typeArguments` rather than a `find()`. @Ubax, happy to add that here or leave it for a follow-up, whichever you prefer.

Ubax:
@expo-bot review

PR #48904, opened 2026-08-13
[ui][ios] Add Background component

Den1Marshall:
# Why

SwiftUI's `background(alignment:content:)` draws an arbitrary view behind
another view. Expo UI could not express it, only a solid color was possible, via the `background`
modifier.

# How

Adds a `Background` component that mirrors the existing `Overlay` one, following
the same slot pattern: the regular children render as the base view, and the
children wrapped in `Background.Content` are drawn behind them, positioned by an
optional `alignment`.

The `background` modifier is untouched. It keeps covering the `ShapeStyle`
overloads, which take no child view and are better expressed as a modifier
applied inline to any view.

# Test Plan

Added a "Background component" screen to native-component-list with three cases:
a rounded rectangle behind a card, a rule aligned to the bottom edge, and an
image behind text. Verified in bare-expo on the iOS simulator.

`et check-packages @expo/ui` passes and the app builds (`BUILD SUCCEEDED`).

# Checklist

- [x] I added a `changelog.md` entry and rebuilt the package sources
- [ ] This diff will work correctly for `npx expo prebuild` & EAS Build
- [x] Conforms with the Documentation Writing Style Guide

PR #48949, opened 2026-08-14
[ui][ios] Forward modifiers to the native view in Overlay, Alert, ConfirmationDialog and Popover

Den1Marshall:
# Why

Follow-up to #48904, where @brentvatne spotted that `Background` declared a
`modifiers` prop but never forwarded it. The same bug is in `main` for four more
components: `Overlay`, `Alert`, `ConfirmationDialog` and `Popover`. Every
modifier passed to them is silently ignored.

# How

All four destructured `modifiers` out of the props and used it only as the
source of event listeners – `createViewModifierEventListener` returns just
`{ onGlobalEvent }`, so the array itself never reached the native view. They now
pass it on, the same way `Shapes`, `Button` and `List` already do.

The native side needed no changes: their props inherit the `modifiers` field
from `UIBaseViewProps`, and `ExpoUIView` wraps each view in `UIBaseView`, which
calls `applyModifiers`.

# Test Plan

Verified in bare-expo on the iOS simulator: temporarily added
`modifiers={[opacity(0.4)]}` to the `Overlay` in the native-component-list
screen (not part of this PR) and confirmed the whole composition — the card and
the NEW badge – renders translucent. Before the fix the modifier had no effect.

`et check-packages @expo/ui` passes.

| Before | After |
| --- | --- |
| Simulator Screenshot - iPhone 17 Pro - 2026-08-15 at 01 58 02 | Simulator Screenshot - iPhone 17 Pro - 2026-08-15 at 01 57 53 |

# Checklist


intergalacticspacehighway (approved):
Thanks!

PR #48969, opened 2026-08-15
[ui][ios] Fix RN hosted view measurement and touch dispatch

intergalacticspacehighway:
# Why

Supersedes https://github.com/expo/expo/pull/48355, https://github.com/expo/expo/pull/48259 and https://github.com/expo/expo/pull/48217

`Pressable` in `RNHostView` drops touch when user presses in, moves a finger and presses out. The press should only cancel if user's finger moves out of the `Pressable` bounds. 

The issue is that RN uses `measure` API when user moves a finger. Currently measure API returns wrong x and y coordinates as RN does not know where SwiftUI placed the `RNHostView`.

This PR aims to provide accurate x and y coordinates for `RNHostView` so pressability works and `measure` API correct results.

# How

### For `RNHostView` that has RN Surface ancestor.
This PR uses content origin offset API provided by shadow node to set position that is used by `measure` API. 

### For `RNHostView` that is rendered in a separate view controller (Sheet, Popover)
This PR uses RootNodeKind trait provided by shadow node to set measure of children relative to itself and sets up it's own touch dispatcher.

# Test Plan

Added new test cases in `ExpoUIMeasurement.tsx` which tests the measure API placement.

# Checklist


brentvatne:
@expo-bot verify with fable

packages/expo-ui/src/swift-ui/RNHostView.tsx
```diff
@@ -16,10 +19,25 @@ export interface RNHostViewProps {
...
+// content grows by the surrounding chrome on every pass and layout never settles.
+// https://github.com/expo/expo/pull/48059
+const hugCrossAxis = { alignSelf: 'flex-start' } as const;
+
 export function RNHostView(props: RNHostViewProps) {
+  // A sheet or popover presents its content in its own view controller, where the React Native
+  // surface root is not an ancestor and so touch dispatches don't work.
+  // Hosted content there owns its touches and is the origin it is measured from; native reads
+  // this one prop for both.
+  const layoutRoot = useIsPresentedInOwnWindow();
```
Kudo, line 27: [💡 suggestion] The Compose `RNHostView` hardcodes `layoutRoot: true` (`src/jetpack-compose/RNHostView/index.tsx:48`), so in the normal tree `measure()` on hosted content returns surface-relative coordinates on iOS but host-relative ones on Android. Is bringing the content-origin approach to Android planned, or is the divergence required by how Compose hosting dispatches touches? Worth a tracking note either way.

apps/test-suite/tests/ExpoUIMeasurement.ios.tsx
```diff
@@ -0,0 +1,326 @@
+import { BottomSheet, Host, HStack, RNHostView, VStack } from '@expo/ui/swift-ui';
+import { padding } from '@expo/ui/swift-ui/modifiers';
+import React from 'react';
+import { ScrollView, View } from 'react-native';
+
+// Tests actual placement of UI matches return returned by RN's measure API.
+// UI is placed by Compose and not Yoga, so these tests test that correct placement is reported to RN so Pressability can work correctly.
```
Kudo: [🐧 nit] This is the iOS test, so SwiftUI, not Compose. The line above also has a typo ("matches return returned").

```suggestion
// UI is placed by SwiftUI and not Yoga, so these tests check that the placement reported to RN is correct and Pressability works.
```

packages/expo-modules-core/common/cpp/fabric/ExpoViewShadowNode.h
```diff
@@ -51,6 +52,31 @@ class ExpoViewShadowNode : public facebook::react::ConcreteViewShadowNode<
...
+    if (!includeTransform) {
+      return ConcreteViewShadowNode::getContentOriginOffset(includeTransform);
+    }
+
+    auto contentOrigin = ContentOriginRegistry::find(this->getTag());
+    if (!contentOrigin) {
+      return ConcreteViewShadowNode::getContentOriginOffset(includeTransform);
+    }
+
+    // `computeRelativeLayoutMetrics` adds this node's own Yoga origin before consulting us,
```
Kudo: [🐧 nit] This comment reads like two merged drafts and has trailing whitespace on the first two lines. Since this is the subtlest math in the change, consider tightening it to something like: "`computeRelativeLayoutMetrics` already adds this node's Yoga origin, so return only the delta from the Yoga box to where the native layout system placed the content."

packages/expo-modules-core/common/cpp/fabric/ContentOriginRegistry.h
```diff
@@ -0,0 +1,51 @@
...
+public:
+  static void set(facebook::react::Tag tag, facebook::react::Point origin) {
+    std::lock_guard<std::mutex> lock(mutex());
+    storage()[tag] = origin;
+  }
+
+  static void clear(facebook::react::Tag tag) {
+    std::lock_guard<std::mutex> lock(mutex());
+    storage().erase(tag);
+  }
```
Kudo, line 33: good to have a clearAll api and we should clearAll when reload

Kudo (approved):
a minor concern is to clearAll ContentOriginRegistry when reload. leaving other ai review if there's anything makes sense to you

packages/expo-modules-core/common/cpp/fabric/ContentOriginRegistry.h
intergalacticspacehighway: Added, called it on ExpoUI module's onDestroy - https://github.com/expo/expo/pull/48969/changes/b1071870a693a3459a77ec3f727f343c1eb501c6

packages/expo-ui/src/swift-ui/RNHostView.tsx
intergalacticspacehighway: Android changes are in the other stacked PR!

intergalacticspacehighway:
@expo-bot review

PR #48970, opened 2026-08-15
[ui][android] Fix RN hosted view measurement and touch dispatch

intergalacticspacehighway:
# Why

Stacked on #48969. This PR handles for android platform.

# How

See #48969 for more details.

# Test Plan

Added new tests in `ExpoUIMeasurement` which tests measure API.

# Checklist


brentvatne:
@expo-bot verify with fable

packages/expo-ui/android/src/main/java/expo/modules/ui/RNHostView.kt
```diff
@@ -166,6 +223,21 @@ internal class RNHostView(context: Context, appContext: AppContext) :
...
+ * Removes [wrapper] from the view that currently contains it, and clears its bounds.
+ *
+ * `AndroidView` hands back this same wrapper every time Compose rebuilds it, but Compose puts it
+ * inside a brand new container each time, so it has to leave the old one first. A removed view keeps
+ * the bounds that container gave it, and React Native finds touch targets by bounds — so until
+ * Compose lays the wrapper out again it would claim a screen area it no longer occupies and swallow
+ * presses meant for whatever is really there. Clearing the bounds keeps it out of the way.
+ * This causes https://github.com/expo/expo/issues/46386
+ */
+internal fun detachForReuse(wrapper: View) {
```
Kudo, line 243: [💡 suggestion] `detachForReuse` is inserted between `TouchDispatchingRootViewGroup`'s KDoc ("A thin FrameLayout that intercepts touch events...") and the class, so that KDoc now dangles above this function and the class loses its doc. Move this function below the class (or above the class KDoc).

packages/expo-ui/src/jetpack-compose/DropdownMenu/index.tsx
```diff
@@ -67,7 +68,12 @@ type NativeMenuProps = {
  * Children should be `DropdownMenuItem` components or other native views.
  */
 export function Items(props: { children: ReactNode }) {
-  return <SlotNativeView slotName="items">{props.children}</SlotNativeView>;
+  // The items are presented in their own popup window; the trigger stays in the React Native surface.
```
Kudo, line 71: [💡 suggestion] Should `TooltipBox`'s tooltip slots get the same treatment? `PlainTooltip`/`RichTooltip` accept arbitrary `children` and render them through Material3's `TooltipBox`, which presents the tooltip in a `Popup` (its own window) — so an `RNHostView` in there would hit the same no-touches/wrong-origin problem this PR fixes for the dialogs and menus. If tooltips are intentionally out of scope, fine to defer.

packages/expo-ui/android/src/main/java/expo/modules/ui/RNHostView.kt
```diff
@@ -144,6 +175,32 @@ internal class RNHostView(context: Context, appContext: AppContext) :
...
   }
 
+  /**
+   * Publishes where Compose placed this view inside its `Host`, which Yoga has no way to know:
+   * Yoga puts the box at the `Host`'s origin while Compose may draw it anywhere inside. Without
+   * this, `measure()` reports the Yoga box, this makes Pressable cancel its press when the finger moves.
+   */
+  @Composable
+  private fun publishContentOriginModifier(): Modifier {
+    // When layoutRoot is true, we don't set content origin because view acts like root and measure uses the view's origin as the origin.
```
Kudo, line 198: [💡 suggestion] `layoutRoot` is a plain `var`, not snapshot state, and this branch is evaluated during composition — so if the prop ever changes after `Content()` first composes, the modifier choice won't update, and a view flipped to `layoutRoot = true` keeps publishing an origin from the wrong coordinate space (and its stale registry entry is never cleared). Meanwhile `setLayoutRoot` does update `dispatchesTouchesToJS` on the live wrapper, which suggests late changes are expected. Is a post-composition change reachable? If not, the wrapper update in `setLayoutRoot` is dead code; if it is, this should be a `mutableStateOf` and the flip should also clear the published origin.

packages/expo-ui/CHANGELOG.md
```diff
@@ -28,6 +29,12 @@
...
+- [Android] Fix a crash when `RNHostView` uses `matchContents`, where the view was sized from its content while Yoga stretched it to its parent. ([#48970](https://github.com/expo/expo/pull/48970) by [@nishan](https://github.com/intergalacticspacehighway))
+- [Android] Fix a hosted `Pressable` in a `PagerView` ignoring presses on pages after the first, where a re-parented host kept the bounds of the holder it left and answered for a screen area it no longer occupied. ([#46386](https://github.com/expo/expo/issues/46386) by [@TomCorvus](https://github.com/TomCorvus)) ([#48970](https://github.com/expo/expo/pull/48970) by [@nishan](https://github.com/intergalacticspacehighway))
+- [Android] Fix a hosted `Pressable` firing after a Jetpack Compose pager or scrollable had already claimed the gesture, so a swipe that started on a button counted as a press. ([#48970](https://github.com/expo/expo/pull/48970) by [@nishan](https://github.com/intergalacticspacehighway))
```
Kudo: the changelog is noisy? could we merge them?

Kudo (approved):

packages/expo-ui/android/src/main/java/expo/modules/ui/RNHostView.kt
intergalacticspacehighway: I assumed `layoutRoot` will never change, but still good to handle it. Added here - https://github.com/expo/expo/pull/48970/changes/e1581a859816613afb2ba9892207a246223ef39c

packages/expo-ui/src/jetpack-compose/DropdownMenu/index.tsx
intergalacticspacehighway: Good catch. Added to Tooltip here - https://github.com/expo/expo/pull/48970/changes/e1581a859816613afb2ba9892207a246223ef39c

packages/expo-ui/android/src/main/java/expo/modules/ui/RNHostView.kt
intergalacticspacehighway: Fixed

intergalacticspacehighway:
@expo-bot review

PR #48972, opened 2026-08-15
[cli] Auto-sync agent skills on expo install

Kudo:
# Why

it's good to install skills when people running `npx expo install {module}`
close ENG-25661

# How

- in install, will scan and sync skills for the "installed" package
- offers `--no-agent-skills` to opt-out the behavior

# Test Plan

- added e2e tests

# Checklist


kitten (commented):
There's a higher-level question I think we should answer first after defining the exact problem before proceeding with the stack: When/What do we want the user to experience?

I think the intention here is to try to capture many, many places where a user may "use" the app (rather than interact with agents) to install skills in.

- `expo install`: maybe they're installing a package with skills?
- `expo start`: maybe they have installed a package with skills without `expo install`
- `expo skills`: maybe they want to do so manually (although we should make sure this is promptless, but haven't checked this yet)

Generally, I think all of these "miss" the right timing. If we do want to catch the right timing, it'd be great to instead suggest people to add `expo skills` to the `package.json:prepare` hook (and/or offer that automatically in templates and create-expo-app), which would be uniform. That then solves both the `expo start` and `expo install` timing "misses" and sync issues.

It also avoids making `expo start` heavier, which can be problematic as it's a very busy and contended time in some cases. I'm hesitant to OK us making that more contended even.

Secondarily, the default recommended `AGENTS.md` could then defer to `expo skills` to be called to sync at a pre-determined/agent-instructed time.

Then, when we're sure we've got the above two timings down, we could consider having a quick check (that isn't always going to run but maybe has a heuristic on already installed/opted-in skills) that then calls `expo skills` also on `expo start`, but only when a cache file says we should and have desynced.

That should probably make sure to re-use the existing `expo-modules-autolinking` init phase, so we're reusing the resolution properly. I don't think that's properly hoisted yet, and I suspect that current implementation duplicates some work as-is in the child PR on this stack

Issue #30401, opened 2024-07-14
run:android gets stuck


raisfeld-ori:
### Summary

recently I tried to run ```npm run android``` in my expo project (usually I just run ```npx expo```) in order to link a new package in my project, but after waiting for an hour (yes, an actual HOUR) it kept being stuck with no text
![image](https://github.com/user-attachments/assets/5365f3d0-88c8-4525-ad20-109fc6aebf55)
at first I thought the error was missing, so I Tried running ```npx expo run:android```
![image](https://github.com/user-attachments/assets/f3ca21dd-0bf7-42ca-89c7-1893e0a7e715)
but that as well kept being stuck this way.

I tried simply making a new project and installing the packages I had in the project (google-signin and expo-image-picker) and I got the same error. it seems like ```npx expo run:android``` just gets stuck for no reason? I'd love to get an explanation

### What platform(s) does this occur on?

Android

### SDK Version

0.18.25

### Environment

  expo-env-info 1.2.0 environment info:
    System:
      OS: Windows 11 10.0.22631
    Binaries:
      Node: 20.9.0 - C:\Program Files\nodejs\node.EXE
      Yarn: 1.22.22 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
      npm: 10.7.0 - C:\Program Files\nodejs\npm.CMD
    SDKs:
      Android SDK:
        API Levels: 33, 34
        Build Tools: 30.0.3, 34.0.0
        System Images: android-30 | Google Play Intel x86 Atom, android-34 | Google APIs Intel x86_64 Atom
    IDEs:
      Android Studio: AI-233.14808.21.2331.11709847
    npmPackages:
      expo: ~51.0.20 => 51.0.20
      react: 18.2.0 => 18.2.0
      react-native: 0.74.3 => 0.74.3
    Expo Workflow: bare

### Minimal reproducible example

by running:
```
npx create-expo-app testing --template blank
npx expo run:android
```
I already got the issue this way, but I also had some packages installed in the project(you don't have to run these commands, but it could help understanding what's happening):
```
yarn add @react-native-google-signin/google-signin
npx expo install expo-image-picker
```

raisfeld-ori:
figured it out!

it seems that the command ```npx react-native run-android``` had an error right at the start, and for some reason expo didn't register it, so it kept on waiting for an output even after the task failed. after using ```npx react-native doctor``` and finding out the issues in my machine (my java version was 22) I managed to fix the error and now the ```npm run android``` works

[...]

PR #48977, opened 2026-08-15
[cli][android] Time out unresponsive adb client commands

### LINKED ISSUE #48744: `expo run:android` hangs forever with zero output when the adb server is wedged - no timeout on `adb start-server` / `adb devices -l`
# Summary

When the resident adb server is unresponsive (the common trigger: a USB device that drops off mid-transfer or mid-install leaves `adb.exe` wedged on Windows), `npx expo run:android` hangs indefinitely with no output at all. No spinner, no "searching for devices" message, no error, and no Gradle/Java process is ever started. The user has no signal about what the CLI is waiting on.

Root cause: the very first thing `run:android` does is resolve a device, and every adb invocation along that path is awaited with no timeout:

- `resolveDeviceAsync` (`packages/@expo/cli/src/run/android/resolveDevice.ts`) calls `AndroidDeviceManager.resolveAsync()`
- which calls `getDevicesAsync()` -> `getAttachedDevicesAsync()` (`packages/@expo/cli/src/start/platforms/android/adb.ts`, ~line 204), which awaits `getServer().runAsync(['devices', '-l'])`
- `ADBServer.runAsync` (`packages/@expo/cli/src/start/platforms/android/ADBServer.ts`) first awaits `startAsync()`, which awaits `spawnAsync(adb, ['start-server'])` with no timeout, then awaits `spawnAsync(adb, args)` for `devices -l`, also with no timeout

When the existing adb server process is wedged, the `adb start-server` / `adb devices -l` client commands connect to the dead server's socket and block forever, so the CLI blocks forever before printing anything. `resolveAdbPromise` only formats errors; it cannot help because the promise never settles.

This is a recurring source of confusion. #30401 is the same symptom (stuck with no text, resolved by killing `adb.exe` in Task Manager, several "same for me" comments) but it was closed with only that manual workaround; nothing in the CLI was changed.

## Expected behavior

Either of these would fix the developer experience:

1. Run device-discovery adb commands (`start-server`, `devices -l`) with a timeout (10 to 15 seconds is generous; healthy invocations take milliseconds). On timeout, fail with an actionable error, for example:

   `Error: adb did not respond within 15s. The adb server may be hung (this can happen when a USB device disconnects mid-transfer). Try: adb kill-server, then re-run the command. On Windows you may need to end the adb.exe process manually.`

   `@expo/spawn-async` does not take a timeout option, but a `Promise.race` around the existing `spawnAsync` call inside `ADBServer.runAsync` (or a plain `child_process` timeout) would be enough, applied at least to the read-only discovery commands where killing the client is always safe.

2. At minimum, print a line such as `Searching for Android devices...` before the first adb call, so a hang points at adb instead of looking like the CLI silently died.

## Actual behavior

`npx expo run:android` prints nothing and never returns. `adb kill-server` from another terminal also hangs against the wedged server; on Windows only `Stop-Process -Force` on `adb.exe` recovers the machine state. After killing adb, the same command works immediately.

# What platform(s) does this occur on?

Android (CLI host observed on Windows; the missing timeout is platform-independent, but the wedged-adb trigger is most common on Windows with physical USB devices)

# SDK Version

57 (`expo@57.0.9`, `@expo/cli@57.0.11`); the code on `main` is unchanged

# Environment

```
  System:
    OS: Windows 11 10.0.26200
  Binaries:
    Node: 22.18.0 - C:\Users\...\node.EXE
    Yarn: 1.22.19
    npm: 10.9.3
  SDKs:
    Android SDK:
      API Levels: 29, 30, 33, 34, 36
      Build Tools: 28.0.3, 29.0.2, 30.0.2, 34.0.0, 35.0.0, 36.0.0, 37.0.0
  IDEs:
    Android Studio: AI-261.26222.65.2613.15948027
  npmPackages:
    expo: ~57.0.9 => 57.0.9
    react: 19.2.3 => 19.2.3
  Expo Workflow: bare
```

# Minimal reproducible example

**Repro repository: https://github.com/Tasselhat/expo-adb-hang-repro**

A fresh `create-expo-app --template blank` (SDK 57) project with one addition: `wedged-android-sdk/`, a fake `ANDROID_HOME` whose `platform-tools/adb` blocks forever without producing output - byte-for-byte what a real adb client does when the resident adb server is wedged (the organic trigger: a USB device dropping off mid-transfer, especially on Windows).

Steps (also in the repo README):

```
git clone https://github.com/Tasselhat/expo-adb-hang-repro
cd expo-adb-hang-repro
npm install
npx expo prebuild -p android    # works normally
```

macOS/Linux:

```
ANDROID_HOME="$PWD/wedged-android-sdk" npx expo run:android
```

Windows (PowerShell):

```
$env:ANDROID_HOME = "$PWD\wedged-android-sdk"; npx expo run:android
```

Verified result (Windows 11, expo@57.0.12): after several minutes, 0 bytes of output, the CLI's node process blocked awaiting the fake adb it spawned for `start-server`, and no Gradle or Java process ever started. Killing the process tree is the only way out.

The same hang occurs against the **real** adb binary by suspending its server process, which is indistinguishable from the organically wedged state: run `adb start-server`, suspend it with `kill -STOP $(pgrep adb)` (macOS/Linux) or Sysinternals `pssuspend adb.exe` (Windows), then run `npx expo run:android` with your normal `ANDROID_HOME`. Resuming adb (`kill -CONT` / `pssuspend -r adb.exe`) makes the command continue immediately, confirming where it was blocked.

Real-world trigger (Windows, physical device): unplug the phone mid-install during `expo run:android`, then re-run the command. The leftover wedged `adb.exe` blocks the next run before any output until it is force-killed.

*Note: previously filed as #48743, which was closed for a missing runnable repro; this issue adds the repro repository above.*

MatheusMartinho:
# Why

Fixes #48744.

`expo run:android` resolves a device before doing anything else, and every adb call on that path is awaited without a timeout: `resolveDeviceAsync` → `AndroidDeviceManager.resolveAsync()` → `getAttachedDevicesAsync()` → `ADBServer.runAsync(['devices', '-l'])`, which first awaits `startAsync()` (`adb start-server`).

When the resident adb server is wedged — commonly after a USB device drops off mid-transfer, which is a recurring report on Windows — the adb client connects to the dead server's socket and blocks forever. The CLI then hangs with no output at all: no spinner, no message, no error, and no Gradle process ever starts, so the user has no signal about what it is waiting on. #30401 is the same symptom, closed with only a manual workaround.

# How

`ADBServer` now runs adb client commands through a small `spawnWithTimeoutAsync` helper that races the spawn against a 15s timer (healthy invocations take milliseconds), kills the client on timeout, and rejects with a `CommandError` that states what failed, the likely cause, and the recovery step:

```
adb did not respond within 15s while running "devices -l". The adb server may be unresponsive, which can happen when a device disconnects mid-transfer. Try running "adb kill-server" and then re-running this command.
```

I used a `Promise.race` because `@expo/spawn-async` takes no timeout option, following the existing timeout pattern in `emulator.ts` in the same directory. Only the two `spawnAsync` call sites are affected (`start-server` and the general `runAsync` path); `getFileOutputAsync` uses `execFileSync` and is left alone.

# Test Plan

Added two tests to `ADBServer-test.ts`, red/green verified:

- a spawn that never settles (the wedged-server case) rejects with the message above and kills the client — this test times out against `main` without the fix, reproducing the reported hang
- a command that responds is unaffected

```
$ jest src/start/platforms/android/__tests__/
Test Suites: 12 passed, 12 total
Tests:       95 passed, 95 total
```

`tsc --noEmit` reports no new errors for the changed files.

# Checklist

### DIFF
--- a/packages/@expo/cli/CHANGELOG.md
+++ b/packages/@expo/cli/CHANGELOG.md
@@ -38,6 +38,7 @@
 - Support npm@12's dictionary-based `npm pack --json` format ([#48761](https://github.com/expo/expo/pull/48761) by [@kitten](https://github.com/kitten))
 - Fix wirelessly connected iOS 16 and older devices being omitted from `expo run:ios --device` selection. ([#48127](https://github.com/expo/expo/pull/48127) by [@davellanedam](https://github.com/davellanedam))
 - Fix resolution of ESLint failing in `expo lint` after prerequisites check installs it ([#46223](https://github.com/expo/expo/pull/46223) by [@claritystorm](https://github.com/claritystorm))
+- Fail with an actionable error instead of hanging forever when the adb server is unresponsive. ([#48744](https://github.com/expo/expo/issues/48744) by [@MatheusMartinho](https://github.com/MatheusMartinho))
 
 ### 💡 Others
 
--- a/packages/@expo/cli/src/start/platforms/android/ADBServer.ts
+++ b/packages/@expo/cli/src/start/platforms/android/ADBServer.ts
@@ -1,4 +1,4 @@
-import spawnAsync from '@expo/spawn-async';
+import spawnAsync, { SpawnResult } from '@expo/spawn-async';
 import { execFileSync } from 'child_process';
 
 import { Log } from '../../../log';
@@ -9,6 +9,14 @@ import { event } from '../events';
 import { assertSdkRoot } from './AndroidSdk';
 
 const BEGINNING_OF_ADB_ERROR_MESSAGE = 'error: ';
+
+/**
+ * Time in milliseconds to wait for an adb client command before assuming the adb server is
+ * unresponsive. Healthy invocations take milliseconds; a wedged server (commonly left behind by a
+ * device that disconnects mid-transfer) never answers at all, which would otherwise hang the CLI
+ * with no output.
+ */
+const ADB_COMMAND_TIMEOUT = 15000;
 
 // This is a tricky class since it controls a system state (side-effects).
 // A more ideal solution would be to implement ADB in JS.
@@ -45,7 +53,7 @@ export class ADBServer {
       }
     });
     const adb = this.getAdbExecutablePath();
-    const result = await this.resolveAdbPromise(spawnAsync(adb, ['start-server']));
+    const result = await this.resolveAdbPromise(this.spawnWithTimeoutAsync(adb, ['start-server']));
     const lines = result.stderr.trim().split(/\r?\n/);
     const isStarted = lines.includes('* daemon started successfully');
     this.isRunning = isStarted;
@@ -77,7 +85,7 @@ export class ADBServer {
     await this.startAsync();
 
     event('adb_server_run', { command: [adb, ...args].join(' ') });
-    const result = await this.resolveAdbPromise(spawnAsync(adb, args));
+    const result = await this.resolveAdbPromise(this.spawnWithTimeoutAsync(adb, args));
     return result.output.join('\n');
   }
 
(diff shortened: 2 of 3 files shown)


MatheusMartinho:
Rebased onto `main` — the branch was conflicting and is mergeable again.

The only conflict was `packages/@expo/cli/CHANGELOG.md`: `main` picked up six commits there since this branch was cut, and the entry now sits at the end of the **🐛 Bug fixes** list under *Unpublished* instead of the top. `ADBServer.ts` and its test were untouched on `main`, so they carried over unchanged.

Same single commit, same diff as before the rebase (+84/-3 across the same three files).

kitten:
Hiya 👋 I'll close this for now in favour of #49258. There's a few issues that shouldn't be ignored, as part of a larger refactor/patch-set, that aren't addressed here, but are still related (since part of the original issue is speculative and the reproduction just "forces" one failure path of a few)

Some issues here are:
- A general timeout doesn't work; a "healthy" adb install cannot be subject to the same timeouts as other commands
- The `Promise.race` is a bit quirky and calls `kill` but doesn't ensure the sub-process gets cleaned up
- A few side-effectful operations (like adb installs) aren't inherently always cancellable, and hence killing an adb process may not cancel them
- A timeout doesn't automatically mean the adb server is dead (we can still probe to say if it has died for sure)
- The `start-server` / `kill-server` lifecycle is invalid and must be removed (hence partially motivating the larger refactor/cleanup)
- `getFileOutputAsync` is also affected by the same issues and uses a blocking command, so needs to be addressed in the same refactor
- A few cases add timers but don't take process cancellation (ctrl+c) into account

Hope that makes sense, Cheers!

PR #48978, opened 2026-08-15
[metro-file-map] Watch through the long path root on Windows

brentvatne:
# Why

`fs.watch` on a path that contains an 8.3 short-name component, such as `C:\Users\RUNNER~1\...`, crashes the whole Node process on libuv 1.52.x, which Node bundles since 24.4:

```
Assertion failed: !_wcsnicmp(filename, dir, dirlen), file src\win\fs-event.c, line 72
```

When a watched directory reports an event, libuv resolves the changed path to its long form and asserts that the result is prefixed by the watch path. A short-name watch path fails that check, and the assert aborts the process ([libuv#5010](https://github.com/libuv/libuv/issues/5010)). The upstream fix ([libuv#5152](https://github.com/libuv/libuv/pull/5152), merged 2026-06-10) has not shipped in a Node release yet.

`FallbackWatcher` is the default watcher on Windows, so an Expo dev server on Node ≥ 24.4 hard-crashes when the project path contains a short-name component. Short paths are common through `%TEMP%` and tooling that resolves user profile paths. A bare `fs.watch(dir)` plus one file write under such a path reproduces the crash with no Expo code involved (verified on `windows-latest` CI runners, Node 24.18.1/24.19.0; Node ≤ 24.3 and 22/20 are unaffected). This was found while adding Windows coverage for #48954.

# How

`FallbackWatcher` now expands its root to the canonical long form once, with `fs.realpathSync.native` — the JavaScript `fs.realpathSync` resolves symlinks but keeps short names, while the native variant expands them. Every `fs.watch` call then goes through the expanded prefix. All bookkeeping, registry keys, and reported events keep the original paths, so nothing changes for consumers. The expansion is win32-only; on other platforms the method returns the path unchanged.

# Test Plan

New tests in `FallbackWatcherWin32Paths.test.ts`, red before the fix and green after:

- `watches through the expanded root and reports the original paths` — mocks the platform to win32 and models a root with a distinct long form; asserts every `fs.watch` call uses the expanded prefix and a delivered event still reports the original root and relative path.
- `watches the original paths when the root has no distinct long form` — asserts the common case stays unchanged.

Verification on macOS: new tests pass, `pnpm test` (18 suites, 701 tests), `pnpm run format`, `pnpm run lint`, and `pnpm run typecheck` all clean. Verification on Windows (`windows-latest`, Node 24.19.0 / libuv 1.52.1, the affected combination): the full package suite passed, 18 suites and 701 tests, including both new tests — run linked in the comments.

Note for reviewers: this lands after #48954; the only expected overlap is adjacent context in `#watchdir` and the changelog.

# Checklist

### DIFF
--- a/packages/@expo/metro-file-map/CHANGELOG.md
+++ b/packages/@expo/metro-file-map/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### 🐛 Bug fixes
 
+- Prevent a Node crash on Windows when a watched path contains an 8.3 short name, by watching through the long form of the root ([#48978](https://github.com/expo/expo/pull/48978) by [@brentvatne](https://github.com/brentvatne))
+
 ### 💡 Others
 
 - [Internal] Migrate an initial set of events to `2g` ([#47655](https://github.com/expo/expo/pull/47655) by [@kitten](https://github.com/kitten))
--- a/packages/@expo/metro-file-map/src/watchers/FallbackWatcher.ts
+++ b/packages/@expo/metro-file-map/src/watchers/FallbackWatcher.ts
@@ -43,6 +43,7 @@ export default class FallbackWatcher extends AbstractWatcher {
     [directory: string]: { [file: string]: true };
   } = Object.create(null);
   readonly #watched: { [key: string]: FSWatcher } = Object.create(null);
+  #longPathRoot: string | null = null;
 
   async startWatching(): Promise<void> {
     this.#watchdir(this.root);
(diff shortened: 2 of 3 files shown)


brentvatne:
Windows verification: the full `@expo/metro-file-map` suite (18 suites, 701 tests, including the two new `FallbackWatcherWin32Paths` tests) passed on `windows-latest` with Node 24.19.0 / libuv 1.52.1 — the affected combination — via a temporary workflow on a debug branch: https://github.com/expo/expo/actions/runs/31904511375. The debug branch has been deleted. The repo-wide `prepare` step currently fails on Windows on `main` for unrelated reasons, so a full `check-packages-windows` dispatch was not usable for this branch.

PR #48979, opened 2026-08-15
[expo-task-manager][Android] Fix persisted job scheduling

### LINKED ISSUE #48935: [expo-task-manager][Android] Persisted JobScheduler job is scheduled without RECEIVE_BOOT_COMPLETED, crashing the process on the first location update

No repo link, deliberately — **this defect is confirmable by inspection in under a minute,
without building anything:**

1. [`expo-task-manager/android/src/main/AndroidManifest.xml`](https://github.com/expo/expo/blob/main/packages/expo-task-manager/android/src/main/AndroidManifest.xml)
   registers a receiver with an `android.intent.action.BOOT_COMPLETED` intent-filter but
   declares **no** `<uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED" />`.
2. [`TaskManagerUtils.java:205`](https://github.com/expo/expo/blob/main/packages/expo-task-manager/android/src/main/java/expo/modules/taskManager/TaskManagerUtils.java)
   builds every `JobInfo` with `.setPersisted(true)`.

Android rejects a persisted job from a package without that permission, so the two facts
above are sufficient to produce the crash below. Full stack trace and exact reproduction
steps are included. I'm happy to put together a bare `create-expo-app` repro if that's
still wanted — just say so.

### Summary

`TaskManagerUtils` builds every `JobInfo` with `.setPersisted(true)`, but neither
`expo-task-manager` nor `expo-location` declares
`android.permission.RECEIVE_BOOT_COMPLETED`, which Android requires for persisted jobs.

The result is that any app using `Location.startLocationUpdatesAsync` with a TaskManager
task is killed **the moment the first location fix is delivered**, unless the app happens
to declare that permission itself. There is no JS-visible error — the exception is thrown
on the main thread inside a broadcast receiver, so the process dies outright. Notably
`startLocationUpdatesAsync` itself resolves successfully and the foreground service starts
normally, which makes this look like a crash somewhere else entirely.

```
FATAL EXCEPTION: main
java.lang.RuntimeException: Unable to start receiver expo.modules.taskManager.TaskBroadcastReceiver:
java.lang.IllegalArgumentException: Requested job cannot be persisted without holding
android.permission.RECEIVE_BOOT_COMPLETED permission
	at android.app.ActivityThread.handleReceiver(ActivityThread.java:5038)
	...
Caused by: java.lang.IllegalArgumentException: Requested job cannot be persisted without holding
android.permission.RECEIVE_BOOT_COMPLETED permission
	at android.app.job.IJobScheduler$Stub$Proxy.schedule(IJobScheduler.java:479)
	at android.app.JobSchedulerImpl.schedule(JobSchedulerImpl.java:89)
	at expo.modules.taskManager.TaskManagerUtils.updateOrScheduleJob(TaskManagerUtils.java:155)
	at expo.modules.taskManager.TaskManagerUtils.scheduleJob(TaskManagerUtils.java:69)
	at expo.modules.location.taskConsumers.LocationTaskConsumer.reportLocationsImmediately(LocationTaskConsumer.kt:265)
	at expo.modules.location.taskConsumers.LocationTaskConsumer.handleLocationUpdate(LocationTaskConsumer.kt:106)
	at expo.modules.location.taskConsumers.LocationTaskConsumer.didReceiveBroadcast(LocationTaskConsumer.kt:85)
	at expo.modules.taskManager.TaskService.handleIntent(TaskService.java:325)
	at expo.modules.taskManager.TaskBroadcastReceiver.onReceive(TaskBroadcastReceiver.java:15)
	at android.app.ActivityThread.handleReceiver(ActivityThread.java:5029)
Caused by: android.os.RemoteException: Remote stack trace:
	at com.android.server.job.JobSchedulerService$JobSchedulerStub.enforceValidJobRequest(JobSchedulerService.java:4835)
	at com.android.server.job.JobSchedulerService$JobSchedulerStub.schedule(JobSchedulerService.java:5042)
```

### Analysis

1. `TaskManagerUtils.createJobInfo` sets `.setPersisted(true)` (`TaskManagerUtils.java:205`).
2. `JobSchedulerService.enforceValidJobRequest` rejects a persisted job unless the calling
   package holds `RECEIVE_BOOT_COMPLETED`.
3. Neither `expo-task-manager` nor `expo-location` declares it. I grepped every
   `node_modules/*/android/src/main/AndroidManifest.xml` in my project — nothing declares it,
   and it is absent from the merged manifest of the built APK.
4. The `schedule` call is wrapped in `catch (IllegalStateException)`
   (`TaskManagerUtils.java:146` and `:156`), but the framework throws
   **`IllegalArgumentException`**. The guard therefore doesn't catch it, the exception escapes
   through `TaskBroadcastReceiver.onReceive`, and the process is killed.

Point 4 looks like the more general bug, and it's the same code path as the already-fixed
#47571 / #48368: even where scheduling legitimately fails, this path takes the whole app
down rather than logging.

### Steps to reproduce

1. Create an Expo app with `expo-location` and `expo-task-manager`.
2. Do **not** add `RECEIVE_BOOT_COMPLETED` to `android.permissions` in `app.json`.
3. `TaskManager.defineTask(TASK, ...)` at module scope.
4. Grant foreground + background location, then call:

```js
await Location.startLocationUpdatesAsync(TASK, {
  accuracy: Location.Accuracy.High,
  timeInterval: 1000,
  foregroundService: {
    notificationTitle: 'Tracking',
    notificationBody: 'Recording position',
  },
});
```

5. Run on an Android device or emulator with a location source. The call itself **succeeds**
   and the foreground service starts normally — the process dies a few seconds later, when
   the first fix is delivered.

Reproduced on a stock Android 16 (API 36) `google_apis;arm64-v8a` emulator, and reported
independently by a tester on a physical device. After declaring the permission, the same
build survives and `dumpsys jobscheduler` shows `TaskJobService` scheduling and completing
cleanly once per fix.

### Workaround

Declare the permission in the app:

```json
{ "expo": { "android": { "permissions": ["android.permission.RECEIVE_BOOT_COMPLETED"] } } }
```

Normal (non-dangerous) permission, auto-granted at install, no user prompt.

### Suggested fix

Any of, in rough order of preference:

1. Declare `<uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED" />` in
   `expo-task-manager`'s `AndroidManifest.xml` — the library already registers a
   `BOOT_COMPLETED` receiver, so it arguably needs this regardless of job scheduling.
2. Only call `.setPersisted(true)` when the permission is actually held
   (`checkSelfPermission`), falling back to a non-persisted job otherwise.
3. At minimum, widen the guard to `catch (IllegalArgumentException | IllegalStateException)`
   so a scheduling failure logs instead of killing the app.

### Environment

- `expo` 56.0.9 (SDK 56)
- `expo-task-manager` 56.0.18
- `expo-location` 56.0.16
- Android 16 / API 36, `targetSdkVersion` 36, arm64
- New architecture enabled, Hermes

LizunovSergey:
## Summary

- Declare `RECEIVE_BOOT_COMPLETED` for TaskManager persisted jobs.
- Handle scheduler failures instead of crashing the host app.

Fixes #48935.

## Test plan

- `:expo-task-manager:testDebugUnitTest`
- `:expo-task-manager:processDebugManifest` (verified the merged manifest contains `RECEIVE_BOOT_COMPLETED`)
- `git diff --check`
- `xmllint --noout packages/expo-task-manager/android/src/main/AndroidManifest.xml`

### DIFF
--- a/packages/expo-task-manager/CHANGELOG.md
+++ b/packages/expo-task-manager/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### 🐛 Bug fixes
 
+- [Android] Prevent crashes when scheduling persisted TaskManager jobs without the required `RECEIVE_BOOT_COMPLETED` permission. ([#48979](https://github.com/expo/expo/pull/48979) by [@LizunovSergey](https://github.com/LizunovSergey))
 - [Android] Clear headless task manager on context destroy ([#47958](https://github.com/expo/expo/pull/47958) by [@Wenszel](https://github.com/Wenszel))
 - [Android] Fix a crash on Android 9 when delivering a task event through `JobScheduler` (geofencing, background location), where the job was built without the scheduling constraint that `JobInfo.Builder.build()` requires. ([#48305](https://github.com/expo/expo/pull/48305) by [@rvaccone](https://github.com/rvaccone))
 
--- a/packages/expo-task-manager/CHANGELOG.md
+++ b/packages/expo-task-manager/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### 🐛 Bug fixes
 
+- [Android] Prevent crashes when scheduling persisted TaskManager jobs without the required `RECEIVE_BOOT_COMPLETED` permission. ([#48979](https://github.com/expo/expo/pull/48979) by [@LizunovSergey](https://github.com/LizunovSergey))
+
 ### 💡 Others
 
 ## 58.0.0 — 2026-09-10
--- a/packages/expo-task-manager/android/src/main/AndroidManifest.xml
+++ b/packages/expo-task-manager/android/src/main/AndroidManifest.xml
@@ -1,4 +1,6 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
+    <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED" />
+
     <application>
         <receiver
             android:name="expo.modules.taskManager.TaskBroadcastReceiver"
--- a/packages/expo-task-manager/android/src/main/java/expo/modules/taskManager/TaskManagerUtils.java
+++ b/packages/expo-task-manager/android/src/main/java/expo/modules/taskManager/TaskManagerUtils.java
@@ -143,7 +143,7 @@ public class TaskManagerUtils implements TaskManagerUtilsInterface {
         JobInfo mergedJobInfo = createJobInfoByAddingData(newestJob, data);
         jobScheduler.cancel(newestJob.getId());
         jobScheduler.schedule(mergedJobInfo);
-      } catch (IllegalStateException e) {
+      } catch (IllegalArgumentException | IllegalStateException e) {
         Log.e(this.getClass().getName(), "Unable to merge job: " + e.getMessage());
       }
       return;
@@ -153,7 +153,7 @@ public class TaskManagerUtils implements TaskManagerUtilsInterface {
     try {
       JobInfo jobInfo = createJobInfo(context, task, newJobId, data);
       jobScheduler.schedule(jobInfo);
-    } catch (IllegalStateException e) {
+    } catch (IllegalArgumentException | IllegalStateException e) {
       Log.e(this.getClass().getName(), "Unable to schedule job: " + e.getMessage());
     }
   }


LizunovSergey:
Rebased onto `main` (`352b8be3`) to clear the conflict.

The conflict was the 58.0.0 release cut today: `packages/expo-task-manager/CHANGELOG.md` grew a fresh `## Unpublished` block, and this branch's entry had been written into the `### 🐛 Bug fixes` section that is now under `## 58.0.0`. Since this PR has not shipped, the entry moved up into the unpublished `### 🐛 Bug fixes`, and 58.0.0's own entries were left where the release put them.

`TaskManagerUtils.java` and `AndroidManifest.xml` are byte-identical to `1379280f` — only the changelog placement changed.

PR #48986, opened 2026-08-16
[expo-ui] feat: add dismissOnOutsideTap and scrimColor to universal BottomSheet on Android

### LINKED ISSUE #48881: [expo-ui] universal BottomSheet on Android: a backdrop tap dismisses the sheet without reaching the view behind it, and there is no way to opt out

https://github.com/kilarsky/expo-ui-bottom-sheet-android-backdrop-repro

### Steps to reproduce

Platform: Android only. Environment: development build, since `@expo/ui` ships native views and Expo Go can't load them. Package manager: npm.

```bash
git clone https://github.com/kilarsky/expo-ui-bottom-sheet-android-backdrop-repro
cd expo-ui-bottom-sheet-android-backdrop-repro
npm install
npx expo run:android
```

1. Tap **Tap me (background)**. `background taps` goes to `1`, so the button works.
2. Tap **Open sheet**. It presents at the `'half'` snap point.
3. Tap the backdrop, directly on top of the **Tap me (background)** button.

**Expected:** a way to keep the sheet open until you press its own **Close** button, or to leave the content behind it live.

**Actual:**

Opening the sheet and tapping the backdrop over the background button: the sheet closes and the background tap counter does not move

- The sheet closes and `background taps` doesn't move. The tap stops at the backdrop, though you aimed it at the button.
- `onDismiss` fires and `isPresented` goes back to `false`, so the app doesn't get stuck.
- The sheet closing is documented behavior. I'm reporting that it's the only option, even though `@expo/ui/jetpack-compose`, the layer this component wraps, has a prop to turn it off.

### Cause

The universal `BottomSheet` renders Material 3's `ModalBottomSheet`, and Compose hosts that in a dialog rather than the activity's view hierarchy. The dialog's window takes the touch and dismisses the sheet, so the event never reaches the React Native view underneath.

Compose has an opt-out for the dismissal, `ModalBottomSheetProperties.shouldDismissOnClickOutside`, and a `scrimColor` for the backdrop. `@expo/ui` wires both through `android/src/main/java/expo/modules/ui/ModalBottomSheetView.kt`:

```kotlin
data class ModalBottomSheetPropertiesRecord(
  @Field val shouldDismissOnBackPress: Boolean = true,
  @Field val shouldDismissOnClickOutside: Boolean = true
) : Record

// …
properties = ModalBottomSheetProperties(
  shouldDismissOnBackPress = props.properties.shouldDismissOnBackPress,
  shouldDismissOnClickOutside = props.properties.shouldDismissOnClickOutside
)
```

`jetpack-compose`'s `ModalBottomSheet` exposes both as documented props in `src/jetpack-compose/ModalBottomSheet/index.tsx`: _"Whether the bottom sheet can be dismissed by clicking outside (on the scrim)."_ and _"The color of the scrim overlay behind the bottom sheet."_

The universal wrapper drops them. `src/universal/BottomSheet/index.android.tsx` passes five props, and `properties` isn't one of them:

```tsx
<Host style={{ position: 'absolute' }} pointerEvents="none">
  <ModalBottomSheet
    ref={sheetRef}
    onDismissRequest={onDismiss}
    showDragHandle={showDragIndicator}
    skipPartiallyExpanded={shouldSkipPartiallyExpanded(snapPoints)}
    modifiers={modifiers}>
```

Without `properties`, Kotlin builds the record from its defaults, pinning `shouldDismissOnClickOutside` to `true` for every caller.

`BottomSheetProps` in `src/universal/BottomSheet/types.ts` has no field for either. `modifiers` won't get you there. It becomes the composable's `modifier` argument, while `properties` is a sibling argument of the same call, so no `ModifierConfig` can reach it.

On iOS the same component does both, through `modifiers`. `swift-ui/modifiers` provides `interactiveDismissDisabled` and `presentationBackgroundInteraction`, _"Controls interaction with the content behind a sheet."_:

```tsx
<BottomSheet
  isPresented={isPresented}
  onDismiss={handleDismiss}
  modifiers={[interactiveDismissDisabled(), presentationBackgroundInteraction('enabled')]}>
```

So iOS callers can control backdrop dismissal and background interaction, and Android callers can control neither.

### Confirmation

`App.compose.tsx` is the same screen on `@expo/ui/jetpack-compose`'s `ModalBottomSheet`, where you can reach both props (`cp App.compose.tsx App.tsx`, then reload):

```tsx
<ModalBottomSheet
  onDismissRequest={handleDismiss}
  properties={{ shouldDismissOnClickOutside: false }}
  scrimColor="transparent">
```

Tapping outside then leaves the sheet open with `onDismiss calls` at `0`, so `shouldDismissOnClickOutside` stops the dismissal. `background taps` stays at `0` through those taps, so the tap still doesn't reach the button behind, even with the backdrop transparent.

### Suggested fix

Add to `BottomSheetProps` and forward to `ModalBottomSheet`:

```ts
dismissOnOutsideTap?: boolean; // default true
scrimColor?: ColorValue;
```

as `properties={{ shouldDismissOnClickOutside: dismissOnOutsideTap }}` and `scrimColor`. No native change needed, and `true` as the default leaves every current call site alone. It also matches what `interactiveDismissDisabled` gives iOS callers today.

Background interaction is the harder half. The dialog window consumes the tap even with a transparent backdrop, so an Android equivalent of `presentationBackgroundInteraction` needs a non-modal host such as `BottomSheetScaffold`.

### Environment

```text
    System:
      OS: macOS 26.5.2
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 22.18.0 - /usr/local/bin/node
      npm: 10.9.3 - /usr/local/bin/npm
    Managers:
      CocoaPods: 1.16.2 - /Users/.../.rbenv/shims/pod
    SDKs:
      iOS SDK:
    IDEs:
      Android Studio: 2026.1 AI-261.25134.95.2612.15822958
    npmPackages:
      expo: ~57.0.12 => 57.0.12 
      react: 19.2.3 => 19.2.3 
    Expo Workflow: bare
```

### Expo Doctor Diagnostics

```text
```

webdevsamran:
## Summary

Adds `dismissOnOutsideTap` and `scrimColor` props to the universal `BottomSheet` on Android, resolving the parity gap reported in #48881.

## Problem

On Android, the universal `BottomSheet` wraps Jetpack Compose's `ModalBottomSheet` but never forwarded:
- `properties.shouldDismissOnClickOutside` ? callers had no way to prevent the sheet from closing when the user taps the backdrop.
- `scrimColor` ? callers couldn't change or hide the backdrop.

On iOS, these controls exist today via `interactiveDismissDisabled()` and `presentationBackgroundInteraction()` modifiers. Android had no equivalent.

## Changes

- **`packages/expo-ui/src/universal/BottomSheet/types.ts`**: Added two new Android-specific props to `BottomSheetProps`:
  - `dismissOnOutsideTap?: boolean` (default `true`) ? when `false`, tapping the scrim keeps the sheet open.
  - `scrimColor?: ColorValue` ? controls the backdrop color; pass `'transparent'` for invisible backdrop.

- **`packages/expo-ui/src/universal/BottomSheet/index.android.tsx`**: Destructures the two new props and forwards them to the underlying `ModalBottomSheet` as `properties={{ shouldDismissOnClickOutside: dismissOnOutsideTap }}` and `scrimColor={scrimColor}`.

Both props are no-ops on iOS and web (the `index.ios.tsx` and `index.tsx` implementations do not accept them).

## Backward Compatibility

`dismissOnOutsideTap` defaults to `true`, so all existing call sites keep the same behavior without any changes.

Fixes #48881


packages/expo-ui/src/universal/BottomSheet/index.android.tsx
```diff
@@ -71,6 +73,8 @@ export function BottomSheet({
         onDismissRequest={onDismiss}
         showDragHandle={showDragIndicator}
         skipPartiallyExpanded={shouldSkipPartiallyExpanded(snapPoints)}
+        properties={{ shouldDismissOnClickOutside: dismissOnOutsideTap }}
```
intergalacticspacehighway: Lets also keep the prop name `shouldDismissOnClickOutside` to match native

intergalacticspacehighway (changes requested):

webdevsamran:
I have updated the prop name to \shouldDismissOnClickOutside\ in the universal components as requested to match the native property.

webdevsamran:
The prop is already named `shouldDismissOnClickOutside` on the current head (`d2d70ac`) — it matches the native property, same as the sibling PR. I think the review landed before that revision, so this should be good for another look whenever you have a moment.

For reference, the diff adds `shouldDismissOnClickOutside` and `scrimColor` to `BottomSheetProps` and forwards them to `ModalBottomSheet`; both already exist on the underlying Kotlin component, so there are no native changes.

intergalacticspacehighway:
@webdevsamran there are some conflicts after the last merge, can you fix them and push? thanks!

webdevsamran:
Merged the latest `main` into this branch and resolved the conflicts in `packages/expo-ui` (`CHANGELOG.md`, universal `BottomSheet` android component + types). The branch now carries both this PR's `shouldDismissOnClickOutside`/`scrimColor` props and the recently merged `shouldDismissOnBackPress`/`contentPadding` work, forwarding all of them to `ModalBottomSheet`. Head is now `6ebc9e96`.

intergalacticspacehighway (approved):

PR #48987, opened 2026-08-16
[expo-ui] feat: add dismissOnBackPress to universal BottomSheet on Android

### LINKED ISSUE #48867: [expo-ui] universal BottomSheet on Android: back press dismisses the sheet without firing `BackHandler`, and there is no way to opt out

https://github.com/kilarsky/expo-ui-bottom-sheet-android-back-press-repro

### Steps to reproduce

Platform: Android only. Environment: development build, since `@expo/ui` ships native views and Expo Go can't load them. Package manager: npm.

```bash
git clone https://github.com/kilarsky/expo-ui-bottom-sheet-android-back-press-repro
cd expo-ui-bottom-sheet-android-back-press-repro
npm install
npx expo run:android
```

The screen keeps a `BackHandler` registered for as long as the sheet is open. It bumps an on-screen counter and returns `true`:

```tsx
useEffect(() => {
  if (!isPresented) return;

  const subscription = BackHandler.addEventListener("hardwareBackPress", () => {
    setBackHandlerCalls((count) => count + 1);
    return true;
  });

  return () => subscription.remove();
}, [isPresented]);
```

1. Tap **Open sheet**. It presents at the `'half'` snap point and the readout shows `isPresented: true`.
2. Press the Android back button.

**Expected:** the handler runs, `BackHandler calls` goes to 1, and the sheet stays open because the handler returned `true`.

**Actual:**

Opening the sheet and pressing back: the sheet closes and the BackHandler counter stays at 0

- The sheet closes.
- `BackHandler calls` stays at 0. The handler doesn't run, so its `return true` changes nothing.
- `onDismiss` does fire and `isPresented` goes back to `false`, so state stays consistent. The sheet closing isn't what I'm reporting. JS never sees the press itself, only the dismissal that follows it.
- Reopening the sheet and pressing back again gives the same result, so nothing is left in a broken state. `BackHandler` simply never runs.

I ran this on an Android 16 (API 36) emulator with `@expo/ui` 57.0.10. `android.predictiveBackGestureEnabled` is `false` in `app.json`, so predictive back isn't involved.

That rules out anything back-driven inside a sheet on Android: a multi-step sheet where back returns to the previous step, a navigator nested in the sheet, a confirmation prompt over unsaved changes, or a sheet that back shouldn't close at all. The first back press closes the whole thing, and JS hears about it afterwards.

### Cause

On Android the universal `BottomSheet` renders Material 3's `ModalBottomSheet`, which Compose puts in its own window instead of the activity's view hierarchy. That window handles the back press and dismisses the sheet, so the event never reaches the activity's back dispatcher, which is where React Native's `BackHandler` listens.

Compose has an opt-out for this, `ModalBottomSheetProperties.shouldDismissOnBackPress`, and `@expo/ui` already wires it through the native view. `android/src/main/java/expo/modules/ui/ModalBottomSheetView.kt` (L27, L109):

```kotlin
data class ModalBottomSheetPropertiesRecord(
  @Field val shouldDismissOnBackPress: Boolean = true,
  @Field val shouldDismissOnClickOutside: Boolean = true
) : Record

// …
properties = ModalBottomSheetProperties(
  shouldDismissOnBackPress = props.properties.shouldDismissOnBackPress,
  shouldDismissOnClickOutside = props.properties.shouldDismissOnClickOutside
)
```

`jetpack-compose`'s `ModalBottomSheet` exposes it as a documented prop: _"Whether the bottom sheet can be dismissed by pressing the back button."_

The universal wrapper drops it. `src/universal/BottomSheet/index.android.tsx` (L69-74) passes five props and no `properties`:

```tsx
<Host style={{ position: 'absolute' }} pointerEvents="none">
  <ModalBottomSheet
    ref={sheetRef}
    onDismissRequest={onDismiss}
    showDragHandle={showDragIndicator}
    skipPartiallyExpanded={shouldSkipPartiallyExpanded(snapPoints)}
    modifiers={modifiers}>
```

`BottomSheetProps` has no equivalent field either, so nothing in the universal component can reach it. The sheet is always back-dismissible and the press always disappears.

### Confirmation

`App.compose.tsx` is the same screen on `@expo/ui/jetpack-compose`'s `ModalBottomSheet`, where the opt-out is reachable (`cp App.compose.tsx App.tsx`, then reload):

```tsx
<ModalBottomSheet
  onDismissRequest={closeSheet}
  properties={{ shouldDismissOnBackPress: false }}>
```

The sheet stays open on back press, and `BackHandler calls` still stays at `0`. The prop stops the dismissal without handing the press to JS.

### Suggested fix

Two changes, and the second is the one that matters for interception:

1. Forward the existing prop. Add `dismissOnBackPress?: boolean` (Android only, default `true`) to `BottomSheetProps` and pass `properties={{ shouldDismissOnBackPress }}`. Pure forwarding, no native change, and it covers "back shouldn't close this sheet".
2. Hand the press to JS. Add an `onBackPress` callback on the sheet, or dispatch through RN's `BackHandler`, so a sheet can navigate internally instead of closing. Per the measurement above, option 1 can't cover this on its own.

### Environment

```text
    System:
      OS: macOS 26.5.2
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 22.18.0 - /usr/local/bin/node
      npm: 10.9.3 - /usr/local/bin/npm
    Managers:
      CocoaPods: 1.16.2 - /Users/.../.rbenv/shims/pod
    SDKs:
      iOS SDK:
    IDEs:
      Android Studio: 2026.1 AI-261.25134.95.2612.15822958
    npmPackages:
      expo: ~57.0.12 => 57.0.12 
      react: 19.2.3 => 19.2.3 
    Expo Workflow: bare
```

### Expo Doctor Diagnostics

```text
```

webdevsamran:
## Summary

Adds dismissOnBackPress prop to the universal BottomSheet component on Android.

## Problem

On Android, the universal BottomSheet wraps Jetpack Compose's ModalBottomSheet. Because Compose places the modal sheet into its own window, the hardware back button / back gesture is intercepted and handled directly by Compose to dismiss the sheet.

By default, the sheet dismisses itself on back press. Developers may want to disable this automatic dismissal.

## Solution

Jetpack Compose's ModalBottomSheet supports opting out of this behavior via ModalBottomSheetProperties(shouldDismissOnBackPress = false).

This PR:
1. Adds dismissOnBackPress?: boolean (default 	rue) to BottomSheetProps in packages/expo-ui/src/universal/BottomSheet/types.ts.
2. Forwards properties={{ shouldDismissOnBackPress: dismissOnBackPress }} to ModalBottomSheet in packages/expo-ui/src/universal/BottomSheet/index.android.tsx.

When dismissOnBackPress={false} is set, pressing back will not dismiss the sheet natively. Note: as identified in testing, the press still does not reach React Native BackHandler listeners because the modal dialog window does not dispatch events to the Activity. This fulfills part of the requirement to control back press behavior natively.

## Backward Compatibility

dismissOnBackPress defaults to 	rue, maintaining exact backwards compatibility with existing behavior.

Part of #48867

### DIFF
--- a/packages/expo-ui/src/universal/BottomSheet/index.android.tsx
+++ b/packages/expo-ui/src/universal/BottomSheet/index.android.tsx
@@ -38,6 +38,7 @@ export function BottomSheet({
   snapPoints,
   testID,
   modifiers,
+  dismissOnBackPress = true,
 }: BottomSheetProps) {
   const sheetRef = useRef<ModalBottomSheetRef>(null);
   const [mount, setMount] = useState(isPresented);
@@ -71,6 +72,7 @@ export function BottomSheet({
         onDismissRequest={onDismiss}
         showDragHandle={showDragIndicator}
         skipPartiallyExpanded={shouldSkipPartiallyExpanded(snapPoints)}
+        properties={{ shouldDismissOnBackPress: dismissOnBackPress }}
         modifiers={modifiers}>
         {/* When the drag handle is hidden, add top padding so content doesn't crop against the top edge of the sheet. */}
         <Column modifiers={contentModifiers}>{children}</Column>
--- a/packages/expo-ui/src/universal/BottomSheet/index.android.tsx
+++ b/packages/expo-ui/src/universal/BottomSheet/index.android.tsx
@@ -38,7 +38,7 @@ export function BottomSheet({
   snapPoints,
   testID,
   modifiers,
-  dismissOnBackPress = true,
+  shouldDismissOnBackPress = true,
 }: BottomSheetProps) {
   const sheetRef = useRef<ModalBottomSheetRef>(null);
   const [mount, setMount] = useState(isPresented);
@@ -72,7 +72,7 @@ export function BottomSheet({
         onDismissRequest={onDismiss}
         showDragHandle={showDragIndicator}
         skipPartiallyExpanded={shouldSkipPartiallyExpanded(snapPoints)}
-        properties={{ shouldDismissOnBackPress: dismissOnBackPress }}
+        properties={{ shouldDismissOnBackPress }}
         modifiers={modifiers}>
         {/* When the drag handle is hidden, add top padding so content doesn't crop against the top edge of the sheet. */}
         <Column modifiers={contentModifiers}>{children}</Column>
--- a/packages/expo-ui/CHANGELOG.md
+++ b/packages/expo-ui/CHANGELOG.md
@@ -17,6 +17,7 @@
 - [iOS] Added the `listRowSeparatorTint` modifier, so a list row separator can be recolored and not only shown or hidden. ([#48804](https://github.com/expo/expo/issues/48804) by [@marco242424](https://github.com/marco242424)) ([#48810](https://github.com/expo/expo/pull/48810) by [@expo-bot](https://github.com/expo-bot))
 - [iOS] Added the `geometryGroup` modifier, so a view's subviews can resolve their geometry against it instead of an animating ancestor. Without it, a `Button` inside a container with an `animation` modifier can draw its label away from its own background while the animation runs. ([#48838](https://github.com/expo/expo/pull/48838) by [@intergalacticspacehighway](https://github.com/intergalacticspacehighway))
 - [Android] Add a Jetpack Compose `Image` component. ([#48521](https://github.com/expo/expo/pull/48521) by [@jakex7](https://github.com/jakex7))
+- [android] Add `dismissOnBackPress` prop to the universal `BottomSheet` component, forwarding it to `ModalBottomSheet` via `properties.shouldDismissOnBackPress`. When false, the back press does not dismiss the sheet (and the press still does not reach BackHandler). Part of #48867.
 
 ### 🐛 Bug fixes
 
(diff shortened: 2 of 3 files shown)


brentvatne:
@expo-bot check --fable

intergalacticspacehighway (approved):
Thanks!

packages/expo-ui/src/universal/BottomSheet/index.android.tsx
```diff
@@ -71,6 +72,7 @@ export function BottomSheet({
         onDismissRequest={onDismiss}
         showDragHandle={showDragIndicator}
         skipPartiallyExpanded={shouldSkipPartiallyExpanded(snapPoints)}
+        properties={{ shouldDismissOnBackPress: dismissOnBackPress }}
```
intergalacticspacehighway: nit: can we keep the prop name to `shouldDismissOnBackPress` just to match the native 😅

intergalacticspacehighway (changes requested):

webdevsamran:
I've reworded the JSDoc, changelog, and PR description to accurately reflect that setting the prop to false stops the dismissal but the event still doesn't reach \BackHandler\. Also changed 'Fixes' to 'Part of'.

webdevsamran:
Renamed to `shouldDismissOnBackPress` to match the native property — thanks for the catch.

It also lines the universal wrapper up with what already exists elsewhere in the package: `ModalBottomSheet` declares `shouldDismissOnBackPress` in `jetpack-compose/ModalBottomSheet/index.tsx`, and `native-component-list`'s `BottomSheetScreen.android.tsx` already passes it under that name. CHANGELOG entry updated to match.

intergalacticspacehighway (approved):

PR #48990, opened 2026-08-16
[expo-audio] Let the Android player handle audio focus

eznix86:
# Why

On Android the ExoPlayer created by `expo-audio` is built with audio focus handling turned off:

```kotlin
// AudioPlayer.kt, same in AudioPlaylist.kt
.setAudioAttributes(AudioAttributes.DEFAULT, false)
```

The second argument is `handleAudioFocus`. With it set to `false` the player never joins the Android audio focus stack, so it plays over whatever else is playing and nothing can pause it.

This is easy to see from the outside. Start playback in an app that uses `expo-audio`, then start playback in any other app, and both play at once. `adb shell dumpsys audio` shows the other app holding focus and the `expo-audio` app absent from the stack while its player is `state:started`:

```
Audio Focus stack entries (last is top of stack):
  pack: com.android.chrome -- gain: GAIN -- loss: none -- uid: 10171

players:
  AudioPlaybackConfiguration ... u/pid:10235 ... state:started   <- expo-audio app
  AudioPlaybackConfiguration ... u/pid:10171 ... state:started   <- Chrome
```

Setting `interruptionMode: 'doNotMix'` in `setAudioModeAsync` does not change this. `AudioModule` requests focus separately, but the player it hands back ignores focus loss.

# What this changes

Builds both players with `handleAudioFocus = true`, so media3 requests focus when playback starts, pauses on loss, and resumes on gain. This is the behaviour media3 implements for exactly this purpose.

# Open question

`AudioModule` also manages audio focus: it calls `requestAudioFocus()` from `play()` and registers a listener that pauses every player on `AUDIOFOCUS_LOSS`. With this change there are two focus owners in one process.

In my testing that combination deadlocks. The module reads the player's own focus request as a loss and runs `allPlayables.forEach { it.pause() }` against the player that just asked to play. The app is then left holding focus with nothing playing:

```
Audio Focus stack entries (last is top of stack):
  pack: <the app> -- gain: GAIN -- loss: none

players: (no entries in state:started)
```

I worked around it downstream by setting `interruptionMode: 'mixWithOthers'`, which makes `requestAudioFocus()` return early so the module stops competing while the player keeps handling focus. That works but is clearly not the intended use of that value.

The code change is the direction I believe is right, but the module's own focus manager needs to stand down for it to be complete, and that is a design call I did not want to make on your behalf. Two things I could not resolve alone:

1. Should `requestAudioFocus()` and the pause on loss listener be removed now that the player handles focus?
2. `duckOthers` is currently implemented in that listener by halving the volume. media3 ducks automatically on `AUDIOFOCUS_LOSS_TRANSIENT_CAN_DUCK`, but the request type is not configurable through `handleAudioFocus`, so the distinction between `doNotMix` and `duckOthers` may need to move into the `AudioAttributes` passed to the player.

Happy to finish this in whichever direction you prefer.

# Test Plan

Tested on Android 16 and Android 17 with Expo SDK 57 and React Native 0.86, New Architecture enabled, `targetSdk 36`.

- Start playback, then start playback in another app. Before this change both play at once. After it the first player pauses.
- `adb shell dumpsys audio | sed -n '/Audio Focus stack/,/^$/p'` shows the app in the focus stack while playing, which it never did before.

# Checklist

### DIFF
--- a/packages/expo-audio/CHANGELOG.md
+++ b/packages/expo-audio/CHANGELOG.md
@@ -14,7 +14,7 @@
 
 ### 🐛 Bug fixes
 
-- [Android] Let the player handle audio focus so playback yields to other apps. ([#PRNUM](https://github.com/expo/expo/pull/PRNUM) by [@eznix86](https://github.com/eznix86))
+- [Android] Let the player handle audio focus so playback yields to other apps. ([#48990](https://github.com/expo/expo/pull/48990) by [@eznix86](https://github.com/eznix86))
 
 - [Android] Pause audio players and playlists when headphones or Bluetooth audio devices disconnect. ([#48151](https://github.com/expo/expo/pull/48151) by [@vivekjm](https://github.com/vivekjm))
 - [Android] Give the lock-screen `MediaSession` instances a unique ID so concurrent active players (and the basic session) no longer collide on the empty default. ([#47101](https://github.com/expo/expo/issues/47101) by [@tsushanth](https://github.com/tsushanth))
--- a/packages/expo-audio/CHANGELOG.md
+++ b/packages/expo-audio/CHANGELOG.md
@@ -14,6 +14,8 @@
 
 ### 🐛 Bug fixes
 
+- [Android] Let the player handle audio focus so playback yields to other apps. ([#PRNUM](https://github.com/expo/expo/pull/PRNUM) by [@eznix86](https://github.com/eznix86))
+
 - [Android] Pause audio players and playlists when headphones or Bluetooth audio devices disconnect. ([#48151](https://github.com/expo/expo/pull/48151) by [@vivekjm](https://github.com/vivekjm))
 - [Android] Give the lock-screen `MediaSession` instances a unique ID so concurrent active players (and the basic session) no longer collide on the empty default. ([#47101](https://github.com/expo/expo/issues/47101) by [@tsushanth](https://github.com/tsushanth))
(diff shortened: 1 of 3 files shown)


alanjhughes:
Hi @eznix86 
This is a well investigated PR, but we can't take the flag flip.

`handleAudioFocus = false` is intentional. We originally let ExoPlayer manage focus, but media3 handles it per player instance. Each instance files its own focus request, so two players in the same app compete with each other like two separate apps. That meant you could never play two sounds at once on Android, while iOS and web allow it. We disabled it in #36221 and moved to a single process-wide focus request managed in AudioModule. Flipping the flag back would reintroduce that regression, and as you found, it also fights with the module's own focus manager. So the answer to your open questions is that the module's manager stays and the player-level handling stays off.

That said, the symptom you're describing shouldn't happen. play() requests focus through the module, and the `AUDIOFOCUS_LOSS` handler pauses all players. One note on your dumpsys output, the module requests `AUDIOFOCUS_GAIN_TRANSIENT`, and transient holders are removed from the stack when another app takes full GAIN. So the app being absent from the stack at that point is expected. The bug, if there is one, is the player staying in `started` after the loss fires.

Could you open an issue with a minimal repro? Ideally confirming whether it happens without calling setAudioModeAsync at all, and what mode values you pass when you do call it. If playback really isn't pausing on focus loss, that's a bug in our focus manager and we want to fix it there

Issue #43732, opened 2026-03-06
Expo 55 - Liquid Glass disappears after tab change


otaviobn:
### Minimal reproducible example

https://github.com/otaviobn/liquid-glass-error-expo?tab=readme-ov-file

### Steps to reproduce

Look the `Expo documentation` button. Toggle appearance and change to another tab. Liquid Glass can't recover and disappears. If you toggle again it returns.

https://github.com/user-attachments/assets/a494fcbd-8af6-4968-8ae9-aac1c9104428

### Environment

```text
expo-env-info 2.0.11 environment info:
    System:
      OS: macOS 26.3.1
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 24.13.1 - ~/.nvm/versions/node/v24.13.1/bin/node
      Yarn: 1.22.22 - ~/.nvm/versions/node/v24.13.1/bin/yarn
      npm: 11.8.0 - ~/.nvm/versions/node/v24.13.1/bin/npm
    Managers:
      CocoaPods: 1.16.2 - /Users/otavio.braga/.rbenv/shims/pod
    SDKs:
      iOS SDK:
        Platforms: DriverKit 25.2, iOS 26.2, macOS 26.2, tvOS 26.2, visionOS 26.2, watchOS 26.2
    IDEs:
      Android Studio: 2025.3 AI-253.29346.138.2531.14876573
      Xcode: 26.3/17C529 - /usr/bin/xcodebuild
    npmPackages:
      expo: ~55.0.5 => 55.0.5 
      expo-router: ~55.0.4 => 55.0.4 
      react: 19.2.0 => 19.2.0 
      react-dom: 19.2.0 => 19.2.0 
      react-native: 0.83.2 => 0.83.2 
      react-native-web: ~0.21.0 => 0.21.2 
    Expo Workflow: managed
```

### Expo Doctor Diagnostics

```text
17/17 checks passed. No issues detected!
```

otaviobn:
Closing this one since https://github.com/expo/expo/issues/43743 will fix my problem. :)

PR #48994, opened 2026-08-16
[expo-glass-effect] Fix glass effect with low opacity animations

### LINKED ISSUE #41024: [ios][ui] expo-glass-effect doesn't work with animated parent opacity changes
### Steps to reproduce

To reproduce setup a reanimated view that is animating the opacity and have the children of that view be a glass view. The glass view will render as a normal view, not a glass view. `isLiquidGlassAvailable()` is also true. The glass view works without opacity animation. This also happens when using RN's Animated library.

In my particular case I have a Modal component like this:
```tsx
export function Modal({
  children,
  isShown,
  onClose,
  animationType = 'fadeIn',
  backdropOpacity = 0.6,
  keyboardAvoiding = false,
}: Props) {
  const screenHeight = useScreenSize().height;
  const { height: keyboardHeight } = useAnimatedKeyboard();
  const prevIsShown = usePrevious(isShown);
  const shouldRender = useStateObject(isShown);
  const backgroundOpacityProgress = useSharedValue(0);
  const panelOpacity = useSharedValue(0);
  const panelScale = useSharedValue(1);
  const panelTop = useSharedValue(0);
  const shouldHandleKeyboardEvents = useSharedValue(keyboardAvoiding);
  const animationTime = isShown ? 300 : 150;

  const animatedBackgroundStyle = useAnimatedStyle(
    () => ({
      backgroundColor: interpolateColor(
        backgroundOpacityProgress.value,
        [0, 1],
        [rgba('#000', 0), rgba('#000', backdropOpacity)]
      ),
    }),
    [backgroundOpacityProgress]
  );

  const animatedPanelStyle = useAnimatedStyle(
    () => ({
      // opacity: withTiming(panelOpacity.value, {
      //   duration: animationTime,
      //   easing: ease,
      // }),
      transform: [
        {
          scale: withTiming(panelScale.value, {
            duration: animationTime * 3,
            easing: easeOutElastic(1.5),
          }),
        },
      ],
      top: withTiming(panelTop.value, {
        duration: animationTime,
        easing: ease,
      }),
    }),
    [panelOpacity, panelScale, panelTop, animationTime, keyboardHeight]
  );

  const animatedKeyboardAvoidingStyle = useAnimatedStyle(
    () => ({
      // Instant when keyboard appears, smooth when dismissing
      top: shouldHandleKeyboardEvents.value
        ? -keyboardHeight.value / 2 // Instant positioning when keyboard appears
        : withTiming(0, {
            // Smooth animation when keyboard dismisses
            duration: 250,
            easing: ease,
          }),
    }),
    [keyboardHeight, shouldHandleKeyboardEvents]
  );

  useEffect(() => {
    if (isShown) {
      shouldRender.setValue(true);
    }

    backgroundOpacityProgress.value = withTiming(isShown ? 1 : 0, {
      duration: animationTime,
      easing: ease,
    });
    panelOpacity.value = isShown ? 1 : 0;

    if (animationType === 'fadeIn') {
      panelScale.value = isShown ? 1 : 0.7;
    }

    if (animationType === 'slideIn') {
      // Slide in to enter then scale out to exit.
      if (prevIsShown && !isShown) {
        panelScale.value = 0.7;
      } else {
        panelTop.value = isShown ? 0 : screenHeight / 2;
      }
    }

    if (prevIsShown && !isShown) {
      setTimeout(() => {
        shouldRender.setValue(false);
        // Resetting values for slideIn animation since animation is non-symetrical
        if (animationType === 'slideIn') {
          panelScale.value = 1;
          panelTop.value = screenHeight / 2;
        }
      }, animationTime);
    }
  }, [
    isShown,
    prevIsShown,
    animationTime,
    animationType,
    backgroundOpacityProgress,
    panelOpacity,
    panelScale,
    panelTop,
    screenHeight,
    shouldRender,
  ]);

  const isRendered = shouldRender.value;

  return (
    <Animated.View
      style={[styles.container, animatedBackgroundStyle, when(!isRendered, displayNone)]}
    >
      <Pressable style={fullSize} onPress={() => isShown && onClose && onClose()} />
      <View style={styles.panelContainer} pointerEvents="box-none">
        <Animated.View style={animatedPanelStyle}>
          <Animated.View style={animatedKeyboardAvoidingStyle}>
            <ModalContext.Provider value={{ shouldHandleKeyboardEvents, exists: true }}>
              {children}
            </ModalContext.Provider>
          </Animated.View>
        </Animated.View>
      </View>
    </Animated.View>
  );
}
```

### Environment

```text
expo-env-info 2.0.7 environment info:
    System:
      OS: macOS 26.1
      Shell: 3.6.0 - /usr/local/bin/fish
    Binaries:
      Node: 22.11.0 - /usr/local/bin/node
      Yarn: 3.2.1 - ~/.yarn/bin/yarn
      npm: 10.9.0 - /usr/local/bin/npm
      Watchman: 2025.06.30.00 - /usr/local/bin/watchman
    Managers:
      CocoaPods: 1.10.2 - /usr/local/bin/pod
    SDKs:
      iOS SDK:
        Platforms: DriverKit 25.1, iOS 26.1, macOS 26.1, tvOS 26.1, visionOS 26.1, watchOS 26.1
    IDEs:
      Android Studio: 2022.1 AI-221.6008.13.2211.9514443
      Xcode: 26.1.1/17B100 - /usr/bin/xcodebuild
    npmPackages:
      babel-preset-expo: ~54.0.0 => 54.0.3 
      expo: 54.0.22 => 54.0.22 
      expo-router: ~6.0.14 => 6.0.14 
      expo-updates: ~29.0.12 => 29.0.12 
      react: 19.1.0 => 19.1.0 
      react-dom: 19.1.0 => 19.1.0 
      react-native: 0.81.5 => 0.81.5 
      react-native-web: ^0.21.0 => 0.21.1 
    npmGlobalPackages:
      eas-cli: 16.20.4
    Expo Workflow: managed
```

### Expo Doctor Diagnostics

```text
Unrelated errors
```

Nezz:
# Why

Fixes https://github.com/expo/expo/issues/41024 for real. The liquid glass effect has an issue that hasn't been fixed since it was introduced to expo: when an element with liquid glass fades in, the glass effect often goes missing.

# How

Only mount the effect when the opacity of the element is high enough. The "official workaround" is basically this but without a fixed opacity:
https://docs.expo.dev/versions/latest/sdk/glass-effect/#opacity-animation-workaround

Through trial and error I found >2% opacity to work, but without knowing what's going on inside UIKit we can't know for sure what the real cutoff is.

# Test Plan

Use this minimal repro: https://github.com/Nezz/expo-repro/tree/glass-effect
Verify on the iOS Simulator and real devices. If all goes well you should see a nice grid of glass effects:
image

This is what happens on the latest version of expo without this fix:
image

Use the leave screen button to verify that https://github.com/expo/expo/issues/43732 did not regress

# Checklist

### DIFF
--- a/docs/pages/versions/unversioned/sdk/glass-effect.mdx
+++ b/docs/pages/versions/unversioned/sdk/glass-effect.mdx
@@ -255,7 +255,7 @@ export default function GlassOpacityAnimationExample() {
   const fadeOpacity = useSharedValue(0);
 
   const glassViewProps = useAnimatedProps(() => {
-    const glassEffectStyle = fadeOpacity.value > 0.01 ? 'regular' : 'none';
+    const glassEffectStyle = fadeOpacity.value > 0.02 ? 'regular' : 'none';
     return {
       glassEffectStyle,
       style: {
--- a/docs/pages/versions/unversioned/sdk/glass-effect.mdx
+++ b/docs/pages/versions/unversioned/sdk/glass-effect.mdx
@@ -235,7 +235,7 @@ const styles = StyleSheet.create({
 
 ### Opacity animation workaround
 
-[Apple does not recommend](https://developer.apple.com/documentation/uikit/uivisualeffectview#Set-the-correct-alpha-value) `opacity` values below `1` on `GlassView` or its parent views, and the glass effect does not render at very low opacity. To animate opacity anyway, use Reanimated to animate a wrapper view's opacity while toggling the `glassEffectStyle` between the desired style and `'none'`.
+Apple does not recommend `opacity` values below `1` on `GlassView` or its parent views, as described in [Set the correct alpha value](https://developer.apple.com/documentation/uikit/uivisualeffectview#Set-the-correct-alpha-value). At very low opacity, the glass effect does not render. To animate opacity anyway, use Reanimated to animate a wrapper view's opacity while toggling the `glassEffectStyle` between the desired style and `'none'`.
 
 <SnackInline label='Glass effect opacity animation workaround' dependencies={['expo-glass-effect', 'react-native-reanimated']} platforms={['ios']}>
 
--- a/docs/pages/versions/unversioned/sdk/glass-effect.mdx
+++ b/docs/pages/versions/unversioned/sdk/glass-effect.mdx
@@ -14,10 +14,6 @@ import { SnackInline } from '~/ui/components/Snippet';
 
 React components that render native iOS liquid glass effect using [`UIVisualEffectView`](https://developer.apple.com/documentation/uikit/uivisualeffectview). Supports customizable glass styles and tint color.
 
-## Known issues
-
-- Setting `opacity` to `0` on `GlassView` or any of its parent views causes the glass effect to not render at all. To fade in/out the glass effect, use the built-in [`animate` and `animationDuration`](#animated-glass-effect-style) props in `glassEffectStyle` instead of changing opacity. If you still want to use `opacity`, see the [opacity animation workaround](#opacity-animation-workaround) example. For more details, see [Apple's documentation](https://developer.apple.com/documentation/uikit/uivisualeffectview#Set-the-correct-alpha-value) and [GitHub issue #41024](https://github.com/expo/expo/issues/41024).
-
 ## Installation
 
 <APIInstallSection />
@@ -239,7 +235,7 @@ const styles = StyleSheet.create({
 
 ### Opacity animation workaround
 
-Since setting `opacity` to `0` on `GlassView` or its parent views causes the glass effect to not render at all, you can use Reanimated to animate a wrapper view's opacity while toggling the `glassEffectStyle` between the desired style and `'none'`.
+[Apple does not recommend](https://developer.apple.com/documentation/uikit/uivisualeffectview#Set-the-correct-alpha-value) `opacity` values below `1` on `GlassView` or its parent views, and the glass effect does not render at very low opacity. To animate opacity anyway, use Reanimated to animate a wrapper view's opacity while toggling the `glassEffectStyle` between the desired style and `'none'`.
 
 <SnackInline label='Glass effect opacity animation workaround' dependencies={['expo-glass-effect', 'react-native-reanimated']} platforms={['ios']}>
 
--- a/packages/expo-glass-effect/ios/GlassView.swift
+++ b/packages/expo-glass-effect/ios/GlassView.swift
@@ -10,6 +10,7 @@ public final class GlassView: ExpoView {
   // Glass effect does not work sometimes if view has not been laid out yet
   // https://github.com/expo/expo/issues/41024#issuecomment-3867466289
   private var isMounted = false
+  private var visibilityLink: CADisplayLink?
 
   private var glassStyle: GlassStyle?
   private var glassTintColor: UIColor?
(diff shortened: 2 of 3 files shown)


brentvatne:
@expo-bot check --fable

docs/pages/versions/unversioned/sdk/glass-effect.mdx
```diff
@@ -239,7 +235,7 @@ const styles = StyleSheet.create({
 
 ### Opacity animation workaround
 
-Since setting `opacity` to `0` on `GlassView` or its parent views causes the glass effect to not render at all, you can use Reanimated to animate a wrapper view's opacity while toggling the `glassEffectStyle` between the desired style and `'none'`.
+[Apple does not recommend](https://developer.apple.com/documentation/uikit/uivisualeffectview#Set-the-correct-alpha-value) `opacity` values below `1` on `GlassView` or its parent views, and the glass effect does not render at very low opacity. To animate opacity anyway, use Reanimated to animate a wrapper view's opacity while toggling the `glassEffectStyle` between the desired style and `'none'`.
```
amandeepmittal: ✏️ **style** `links-and-headings.md#link-text`

The link text "Apple does not recommend" does not tell the reader where the link goes. Use the destination's own section title, "Set the correct alpha value", and keep the recommendation in the sentence.

---

The first sentence joins two separate ideas with "and": what Apple recommends, and what the glass effect does at low opacity. Split it into two sentences.

```suggestion
Apple does not recommend `opacity` values below `1` on `GlassView` or its parent views, as described in [Set the correct alpha value](https://developer.apple.com/documentation/uikit/uivisualeffectview#Set-the-correct-alpha-value). At very low opacity, the glass effect does not render. To animate opacity anyway, use Reanimated to animate a wrapper view's opacity while toggling the `glassEffectStyle` between the desired style and `'none'`.
```

amandeepmittal (changes requested):

amandeepmittal (approved):
Thanks for updating the docs! Let's wait for technical review.

Nezz:
>return early from waitForVisibility() when isGlassEffectAvailable() is false, so devices below iOS 26 never start a link.

This sounds fair, let me know if I should make that change

>Set link.preferredFramesPerSecond to a low value, because the effect only needs to install within a frame or two of becoming visible

Not sure how impactful this would be, but it could result in ugly animations

>Two further unverified risks follow from the same gate: the check reads UIView.alpha, so opacity driven through layer.opacity can pass the gate while the view is still transparent

I tried `layer.opacity`, but that didn't work.

invivek26:
We shipped this patch in our production app (Expo SDK 54, react-native-screens native stack) — it fixes the Reanimated-driven fade-in cases, but we hit one gap worth folding in: screens presented with a **native** transition (`animation: "fade"` on a native-stack screen) still came up with permanently flat glass.

The cause is that UIKit screen transitions set the **model** `alpha` to its final value immediately and interpolate only the presentation layer. `isEffectRenderable` reads `current.alpha`, so on the incoming screen's first layout pass the walk reports full opacity while the layer still renders at ~0 — the effect installs against a not-yet-visible backdrop and lands in exactly the degenerate state this PR guards against, with nothing left to retry.

Reading the presentation layer closes the gap; the display link keeps polling and the install lands a frame or two into the transition, once the on-screen opacity actually crosses the threshold:

```swift
let presentedAlpha = current.layer.presentation().map { CGFloat($0.opacity) } ?? current.alpha
opacity *= Swift.min(current.alpha, presentedAlpha)
```

Taking the `min` also covers the opposite direction (model alpha already snapped down while the presented value is still high during a fade-out).

With this amendment on top of the PR we've verified on device: Reanimated-driven opacity fades, native fade screen transitions, and re-install after react-native-screens detaches and re-attaches a covered screen. Happy to provide a repro if useful — any native-stack screen with `animation: "fade"` containing a `GlassView` shows it.

Nezz:
It's been two months since I opened this and the fix was verified. Any chance to push this forward? Anything you need from me?

brentvatne:
@intergalacticspacehighway - can you review this?

intergalacticspacehighway:
@Nezz we wanted to try this solution a while ago - https://github.com/expo/expo/pull/41225 but i closed it as it has a potential to miss cases like this one - https://github.com/expo/expo/pull/48994#issuecomment-5485852258. Can you also include @invivek26 's case? Let's merge this for now until we come across another such a case 😅

Nezz:
On it. I'll also test with ~~Xcode~~ iOS 27 to see ensure that things are fine there too.

Nezz:
@invivek26 I was unable to reproduce that case on iOS 26.4, 26.5, and 27.0 (simulator). I was able to reproduce the condition you described, but the glass still rendered correctly with my fix.

Do you have a minimal repro?

intergalacticspacehighway (approved):

Nezz:
The fix still holds up with iOS 27 (built with Xcode 26 due to known limitations).

packages/expo-glass-effect/ios/GlassView.swift
```diff
@@ -53,15 +54,70 @@ public final class GlassView: ExpoView {
...
+        return false
+      }
+      opacity *= current.alpha
+      view = current.superview
+    }
+
+    // This number was chosen through trial and error.
+    // If adjusting it, be sure to test on various real devices.
+    return opacity > 0.02
+  }
```
intergalacticspacehighway, line 85:
```suggestion
    while let current = view {
      if current.isHidden {
        return false
      }
      opacity *= current.alpha
      // Alpha is clamped to 0...1 and composites multiplicatively,
      // so no ancestor can raise the product back above the threshold.
      if opacity <= 0.02 {
        return false
      }
      view = current.superview
    }

    return true
  }
```

packages/expo-glass-effect/ios/GlassView.swift
```diff
@@ -53,15 +54,70 @@ public final class GlassView: ExpoView {
...
+    // the old UIGlassEffect, otherwise re-assigning does not render GlassView correctly.
+    // https://github.com/expo/expo/issues/43732
+    glassEffectView.effect = UIVisualEffect()
+    updateEffect()
+  }
+
+  private func waitForVisibility() {
+    guard visibilityLink == nil else {
+      return
+    }
```
intergalacticspacehighway, line 100:
```suggestion
    guard isGlassEffectAvailable(), visibilityLink == nil else {
      return
    }
```

packages/expo-glass-effect/ios/GlassView.swift
intergalacticspacehighway: Adds early return if opacity became <= 0.02.

Nezz: The early return is a micro-optimization that IMO just makes the code harder to read

packages/expo-glass-effect/ios/GlassView.swift
Nezz: There was a better place to do this, I now check this before installing the effect

PR #49000, opened 2026-08-17
[router] Stabilize server middleware

hassankhan:
# Why

As part of SDK 58, we are moving server middleware out of alpha and stabilizing it.

# How

Middleware discovery and execution no longer require `unstable_useServerMiddleware`. The config plugin still accepts the deprecated option for compatibility and warns when it is enabled, matching loaders and server rendering.

Also updated the documentation and E2E fixtures to use middleware without an opt-in.

# Test Plan

- CI
- Manual testing against `apps/router-e2e`

# Checklist


amandeepmittal (approved):

packages/expo-router/src/getRoutesCore.ts
```diff
@@ -135,25 +133,6 @@ export function getRoutes(contextModule: RequireContext, options: Options): Rout
 function getMiddleware(contextModule: RequireContext, options: Options): MiddlewareNode | null {
   const allMiddlewareFiles = contextModule.keys().filter((key) => key.includes('+middleware'));
```
jakub-agent, line 134: [🚨 **critical**] Removing the gate means this runs on the client too, and the client contexts only exclude the **root** `+middleware` file — `_ctx.ios.js` / `_ctx.web.js` anchor the lookahead at `./`, and `_ctx.js` doesn't exclude `+middleware` at all. So `app/foo/+middleware.ts` used to `console.warn` and now throws `The middleware file can only be placed at the root level.` out of render, on native as well.

Is that escalation intended? WDYT about keeping the placement check on the server side and leaving the client on a warning?

packages/@expo/cli/e2e/__tests__/export/__snapshots__/static-rendering.test.ts.snap
```diff
@@ -316,7 +316,7 @@ input::-webkit-search-cancel-button,input::-webkit-search-decoration,input::-web
...
-@keyframes r-t2lo5v{0%{opacity:1;}100%{opacity:0;}}</style><style id="expo-generated-fonts">@font-face{font-family:"sweet";src:url("/assets/__e2e__/static-rendering/sweet.7c9263d3cffcda46ff7a4d9c00472c07.ttf");font-display:auto}</style><link rel="preload" href="/assets/__e2e__/static-rendering/sweet.7c9263d3cffcda46ff7a4d9c00472c07.ttf" as="font" crossorigin="" /><link rel="icon" href="/favicon.ico"/><script type="module">globalThis.__EXPO_ROUTER_HYDRATE__=true;</script><link rel="preload" href="/_expo/static/css/global-1f3e4578a1313411a410a9d3f593ae38.css" as="style">
+@keyframes r-t2lo5v{0%{opacity:1;}100%{opacity:0;}}</style><style id="expo-generated-fonts">@font-face{font-family:"sweet";src:url("/assets/__e2e__/static-rendering/sweet.7c9263d3cffcda46ff7a4d9c00472c07.ttf")}</style><link rel="preload" href="/assets/__e2e__/static-rendering/sweet.7c9263d3cffcda46ff7a4d9c00472c07.ttf" as="font" crossorigin="" /><link rel="icon" href="/favicon.ico"/><script type="module">globalThis.__EXPO_ROUTER_HYDRATE__=true;</script><link rel="preload" href="/_expo/static/css/global-1f3e4578a1313411a410a9d3f593ae38.css" as="style">
```
jakub-agent, line 319: [🏗️ **design**] Why did you need to add this change? Dropping `font-display:auto` does not look related to middleware, and it also lands in `static-splitting` and twice in `server-rendering`. Can it go in its own PR so the rest of the stack doesn't inherit it?

packages/expo-router/plugin/src/withRouter.ts
```diff
@@ -109,6 +112,12 @@ const withRouter: ConfigPlugin<Props | void> = (config, _props) => {
   const props = _props || {};
   validate(schema, props);
 
+  if (props.unstable_useServerMiddleware) {
```
jakub-agent, line 115: [💡 suggestion] Config plugins are re-evaluated on every config read, so this will repeat. `expo-dev-launcher` wraps its deprecation in a `warnOnce` helper and `expo-sqlite` uses `WarningAggregator` — can we reuse one of those?

Can the message also say where it comes from and what to do, e.g. "`unstable_useServerMiddleware` in the `expo-router` config plugin is deprecated as of SDK 58. Remove it from your app config."

docs/pages/versions/unversioned/sdk/router/index.mdx
```diff
@@ -125,13 +125,6 @@ If you are using the [default](/more/create-expo/#--template) template to create
         'Disable synchronous layout updates for native screens. This can help with performance in some cases.',
       default: 'false',
     },
-    {
-      name: 'unstable_useServerMiddleware',
```
jakub-agent, line 129: [💡 suggestion] The option still exists and now warns, so someone who sees the warning and searches the docs finds nothing. `ConfigPluginProperties` already supports `deprecated: true` — can we keep the row with that flag and a "no longer required, has no effect" description instead of deleting it?

packages/expo-router/plugin/src/__tests__/withRouter.test.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  it.each([undefined, false, true])('accepts the deprecated option %p', (enabled) => {
+    expect(() =>
+      withRouter(
+        { name: 'test', slug: 'test' },
+        enabled === undefined ? {} : { unstable_useServerMiddleware: enabled }
+      )
+    ).not.toThrow();
+
+    if (enabled) {
```
jakub-agent, line 22: [🐧 nit] nit: the `if (enabled)` branch makes one case assert the warning and two assert its absence. Two plain tests read easier — `it('warns when the deprecated option is enabled')` plus `it.each([undefined, false])('does not warn for %p')`.

`.not.toThrow()` also doesn't check much here, since there's no throwing path for this option.

jakub-agent (commented):

PR 1/6 in the SDK 58 router stabilization stack. The flag removal itself looks right and the deprecation path matches what the rest of the stack does.

Two things I'd like to understand before this merges:

1. `getMiddleware()` now runs on every `getRoutes()` call, including the client one. The client require-contexts only strip the **root** `+middleware` file, so a nested `app/foo/+middleware.ts` now throws out of render instead of warning — on native too, where middleware never runs.
2. The three snapshot files drop `font-display:auto`, which has nothing to do with middleware.

Rest is nits. CI is green apart from `ios-build` still pending — let's wait for it.

packages/@expo/cli/e2e/__tests__/export/__snapshots__/static-rendering.test.ts.snap
hassankhan: This was fixed in #50173

packages/expo-router/src/getRoutesCore.ts
hassankhan: I think keeping the error makes sense: nested middleware is invalid, and this preserves the behavior apps already had when middleware was enabled.

Ubax (approved):

PR #48816, opened 2026-08-12
[docs] Add Japanese translation for the CI/CD tutorial

kazutoyo:
# Why

The Japanese docs now cover the Expo tutorial, the Build with AI tutorial, and (in #48814) the EAS tutorial. The CI/CD tutorial was the remaining gap in the tutorial tree: a Japanese reader who finishes the EAS tutorial lands on a chapter list that switches back to English at `/tutorial/cicd/introduction/`.

The CI/CD tutorial is also where a reader is most likely to be reading with a terminal and the EAS dashboard open at the same time, so the untranslated stretch is more costly than usual — YAML syntax, job types, and trigger semantics are exactly the material that benefits from the reader's first language.

# How

**Translation:** Adds all nine chapters under `docs/pages/ja/tutorial/cicd/`, one-to-one with the English sources. Prose, frontmatter (`title` / `sidebar_title` / `description`), `Collapsible` summaries, `Requirement` titles, `BoxLink` copy, and `ContentSpotlight` alt text are translated. Workflow YAML, `Terminal` commands, `eas.json` / `app.json` snippets, job and parameter names, and EAS dashboard UI labels (**Workflow graph**, **Install**, **View Deployment**, and so on) are left byte-identical to the English pages so a reader can match them against what they see on screen.

Cross-page links follow the existing convention: links into translated pages point at `/ja/...`, links into untranslated pages keep their English path. Anchors into translated headings were rewritten to the Japanese-generated ids (for example `/ja/tutorial/cicd/first-workflow/#eas-workflows-のジョブの種類`).

**ProgressTracker:** `ui/components/ProgressTracker/TutorialData.tsx` is an English-only registry, so each chapter page passes explicit `chapterTitle` / `summary` / `nextChapterTitle` / `nextChapterDescription` / `nextChapterLink` overrides. `currentChapterSlug` stays on the English path because it is the lookup key. Chapter 7 also overrides the trailer, which would otherwise render the English "Next steps in your journey with EAS Workflows" title and an English link.

**i18n registration:** Adds the nine pages to `JA_TRANSLATED_PAGES` and adds `'CI/CD tutorial': 'CI/CD チュートリアル'` to `JA_SECTION_TITLES`, so the sidebar section header is translated alongside its entries.

**Hashes:** `checks/ja/source-hashes.json` regenerated with `checks/ja/stamp.ts` — additive only, nine new entries.

# Test Plan

[...]

PR #49009, opened 2026-08-17
[docs] Exempt the Japanese pages from the sentence case heading rule

kazutoyo:
# Why

Follow-up to #48816, where we agreed to handle this in a separate PR.

Sentence case is an English orthography rule. Japanese has no letter case, so `HeadingCase` can only ever fire on a Japanese heading that happens to open with an ASCII identifier — and Japanese headings are verb-final, so they routinely open with the object: a file name, a job type, a CLI flag, an acronym.

In #48816 that produced ten headings where applying the rule renamed real files — `### Hello.yml ファイルを追加する` above instructions to create a file named **hello.yml** — and the resolution was six new exception patterns.

The pattern repeats on every batch of translated pages, and the exceptions it needs are wide enough to reach the English docs. Translating the Development process guides needs another twelve: `app config`, `assetlinks.json`, `babel-plugin`, `config plugin`, `direnv`, `gradle`, `jest`, `monorepo`, `peerDependencies`, `react-native-config`, `source-map-explorer`, `DOM`, and `ref`. That last one has to be `.*\brefs?\b.*` to cover `## ref を渡す`, which exempts every English heading containing "ref" or "refs" as a side effect. So each new translated page quietly buys down the check for pages the rule was written for.

# How

Scope the rule out of `pages/ja/**` in `.vale.ini`, the same way `README.md` already opts out of `Prerequisites` and `KeyboardShortcuts`.

```ini
[**/pages/ja/**]
expo-docs.HeadingCase = NO
```

`HeadingCase.yml` is untouched — the six patterns added in #48816 stay where they are rather than being cleaned up here, so this PR is six lines and reversible on its own.

The other 24 styles still apply to the translations.

# Test Plan

- `pnpm lint-prose` — clean, 1592 files.
- Confirmed the exemption is scoped to this one rule, not to the directory. Added a merge conflict marker to `pages/ja/tutorial/screenshot.mdx` and `expo-docs.MergeConflictMarkers` still fired; added `## console ログを削除する` to a Japanese page and `HeadingCase` stayed quiet.
- Confirmed English pages are unaffected: this PR changes no style file, and the run above covers the full `pages` tree.

# Checklist

### DIFF
--- a/docs/.vale.ini
+++ b/docs/.vale.ini
@@ -34,3 +34,9 @@ BasedOnStyles =
 [**/README.md]
 expo-docs.Prerequisites = NO
 expo-docs.KeyboardShortcuts = NO
+
+# Sentence case is an English orthography rule. Japanese headings lead with the
+# object, so they routinely open with a lowercase identifier or an acronym, and
+# the check can only ever fire on those. See expo/expo#48816.
+[**/pages/ja/**]
+expo-docs.HeadingCase = NO
--- a/docs/.vale.ini
+++ b/docs/.vale.ini
@@ -35,8 +35,5 @@ BasedOnStyles =
 expo-docs.Prerequisites = NO
 expo-docs.KeyboardShortcuts = NO
 
-# Sentence case is an English orthography rule. Japanese headings lead with the
-# object, so they routinely open with a lowercase identifier or an acronym, and
-# the check can only ever fire on those. See expo/expo#48816.
 [**/pages/ja/**]
 expo-docs.HeadingCase = NO


docs/.vale.ini
```diff
@@ -34,3 +34,9 @@ BasedOnStyles =
 [**/README.md]
 expo-docs.Prerequisites = NO
 expo-docs.KeyboardShortcuts = NO
+
+# Sentence case is an English orthography rule. Japanese headings lead with the
+# object, so they routinely open with a lowercase identifier or an acronym, and
+# the check can only ever fire on those. See expo/expo#48816.
```
amandeepmittal:
```suggestion
```

Comments are not required. It is self-explanatory.

amandeepmittal (changes requested):

amandeepmittal (approved):

amandeepmittal:
Thank you for following up on this!

PR #48814, opened 2026-08-12
[docs] Add Japanese translation for the EAS tutorial

kazutoyo:
# Why

Following #48572, the Japanese docs cover the Expo tutorial and the Build with AI tutorial, but stop there. A reader who finishes either one in Japanese hits a wall at the EAS tutorial, which is exactly where the docs stop being about writing app code and start being about credentials, provisioning profiles, store submission, and CI — the parts where a language barrier costs the most time. Japanese documentation matters a lot for Expo adoption in the Japanese community, and this closes the next gap.

Translating a section that is this CLI-heavy also surfaced two checks that needed adjusting, described below.

# How

**Translation:** Added all thirteen chapters of the EAS tutorial under `pages/ja/tutorial/eas/`, matching the English source one-to-one (frontmatter, prose, `<BoxLink>` and `<VideoBoxLink>` props, `<Collapsible>` summaries, `<Prerequisites>` requirements, and the `<ProgressTracker>` localization overrides). Code blocks, Terminal blocks, and image `alt` text are kept identical to English. Added `pages/ja/tutorial/eas/index.js` so `/ja/tutorial/eas/` redirects to the introduction the same way the English route does.

Interactive prompts are the one place this section differs from the earlier tutorials. `eas build`, `eas credentials`, and `eas device:create` ask questions the reader has to answer, so the English prompt string is kept verbatim — it has to be matchable against what the terminal actually prints — with a Japanese gloss in parentheses after it, for example:

> - **Generate a new Android Keystore**?（新しい Android Keystore を生成しますか？）<kbd>Y</kbd> を押します。

**i18n registration:** Registered the thirteen pages in `JA_TRANSLATED_PAGES` and added the `EAS tutorial` sidebar section title. Thanks to the consolidation in #48572 this is a one-line-per-page change.

**Tests:** Two existing checks needed updating, and both were pointing at something real:

- `checks/ja/anchors.test.ts` reported `/ja/tutorial/eas/introduction#prerequisites` as broken. The link works in the browser — `Prerequisites` registers a fixed `prerequisites` anchor — but the checker only collected ids from markdown headings and `Collapsible` summaries, so it could not see it. Its docstring already claimed `Requirement` ids counted, so this brings the implementation in line with the intent. Without the fix, any page linking to a `Prerequisites` block fails the check.
[...]

PR #49010, opened 2026-08-17
[docs] Add Japanese translation for the Development process guides

kazutoyo:
# Why

The Japanese docs cover the tutorial tree — the Expo tutorial, Build with AI, EAS (#48814), and CI/CD (#48816). Development process is the section a reader reaches next, and the one they stay in: it holds the mental model of the development loop, app config, CNG, linking, local builds, web, and bundling.

Today a Japanese reader who finishes the tutorials lands on `/workflow/overview/` in English, and every guide the tutorials link into is English too. This closes that gap for the whole section.

# How

Adds the 50 pages of the section under `docs/pages/ja/`, one-to-one with the English sources — same headings, code blocks, components, and order.

**Translated:** prose, frontmatter (`title` / `sidebar_title` / `description`), `Collapsible` summaries, `BoxLink` copy, `Requirement` titles, and table headers.

**Left as-is:** code blocks, `Terminal` commands, config keys, `alt` text, and the labels a reader matches against a screen — Android Studio and Xcode menu items, Google Play Console and Apple Developer Portal sections, EAS dashboard copy. Error-message headings such as `### Application has not been registered` stay in English so a reader can search for what they see in their terminal.

**Terminology.** A few choices worth flagging, since they set precedent for the rest of the section:

- `permission` → パーミッション, with 許可 / 拒否 reserved for the act of granting or denying. 権限 reads as a generic noun in a list of topic names next to app config and ユニバーサルリンク.
- `precompiled` / `prebuilt` Expo Modules → プリコンパイル済み for both. プリビルド would collide with `npx expo prebuild`. Note the English pages use the two words for the same feature — `guides/prebuilt-expo-modules.mdx` is titled "Precompiled Expo Modules" but says "Prebuilt artifacts", and `local-app-overview.mdx` says "Prebuilt Expo Modules for Android".
- `submit` → 提出, matching the tutorial pages.
- `web` stays lowercase everywhere, matching the English style guide.

**Links.** Links into translated pages point at `/ja/...`, links into untranslated pages keep their English path. Anchors into translated headings were rewritten to the Japanese-generated ids — `/ja/workflow/overview/#開発ビルド`, `/ja/guides/analyzing-bundles/#expo-atlas-でバンドルサイズを分析する`. The existing Japanese pages that linked into these guides were updated the same way.

**Hashes:** `checks/ja/source-hashes.json` regenerated with `checks/ja/stamp.ts`.

The sidebar group headers for this section (**Linking**, **Web**, **Bundling**, and so on) still render in English. That is a separate rendering bug in `SidebarCollapsible`, and a follow-up PR fixes it with `JA_SECTION_TITLES` entries and tests. Nothing here depends on it.

# Test Plan

- 58 suites, 613 tests pass. `checks/ja/sync` covers the hashes against the English sources, `checks/ja/anchors` covers ja→ja anchor resolution, and `common/i18n` covers the two-way match between `JA_TRANSLATED_PAGES` and the files on disk.
- `node scripts/lint.js --max-warnings 0` — oxfmt, oxlint, tsc, eslint clean. `pnpm lint-prose` clean across 1643 files.
- Checked every page against its English source for structural drift — heading, code fence, list item, `Step`, `Collapsible`, `BoxLink`, `Terminal`, `Tab`, link, and table row counts all match.
- Parsed the frontmatter of all 92 Japanese pages as YAML, which the test suite does not do. This caught a `description` that began with a quote and so was not valid YAML.
- Ran the docs locally and loaded all 50 pages (all 200), and confirmed the sidebar entries and page titles render in Japanese.

# Checklist


amandeepmittal (commented):
Hey @kazutoyo, thanks for opening this PR and for your contribution. Currently, we do not have a plan to manage multilanguage translations outside the tutorial/learn section of our docs. We don't have a large team that can offer continuous maintenance of these docs and at the same time, we don't want authoring experience to become a bottleneck whether it comes from a contributor internally and externally.

That said, we will be looking into maintaining and offering support soon. Closing this PR for now. We are also going to keep the scope tightly for the tutorials section right now.

kazutoyo:
Thanks for explaining — that makes sense, and no argument from me. Continuous maintenance is the hard part, and I would rather not add to it without a plan behind it.

If translations are something you want to take on later, I would be glad to help.

amandeepmittal:
> Thanks for explaining — that makes sense, and no argument from me. Continuous maintenance is the hard part, and I would rather not add to it without a plan behind it.
> 
> If translations are something you want to take on later, I would be glad to help.

Thank you for understanding! Will definitely reach out to you if something comes up.

PR #49011, opened 2026-08-17
[app-metrics][android] Fetch only requested metrics from db

Ubax:
# Why

At the moment when we load any metric/event from Room DB, all metrics from its session are loaded. While for short sessions this is not an issue, the longer the session lives, the more data it accumulates and the more expensive the query becomes.

In practice we only need to load the metrics we want to send, so there is no need to load all of them from given session.

# How

1. Don't relay on `Relation` to fetch the data
2. Add two separate queries - one to load metrics by id and second to load sessions 
3. Combine the metrics and sessions together in Kotlin

## Before

1. One DB query with a `join` - fetches all the metrics
2. Kotlin side filtering of only requested metrics

## After

1. Two DB queries - only necessary data is loaded
2. Kotlin side grouping into a combined object

# Test Plan

1. CI
2. Observe-tester

# Checklist

### DIFF
--- a/packages/expo-app-metrics/CHANGELOG.md
+++ b/packages/expo-app-metrics/CHANGELOG.md
@@ -17,6 +17,7 @@
 
 ### 💡 Others
 
+- [Android] Load only requested metric and log rows when preparing observability payloads.
 - Rename the no-update `downloadComplete` state event to `downloadCompleteUnavailable`. ([#47902](https://github.com/expo/expo/pull/47902) by [@kudo](https://github.com/kudo))
 - [iOS] Measure the JS bundle load time against the app startup end marker to stay compatible with upcoming React Native versions. ([#47782](https://github.com/expo/expo/pull/47782) by [@tsapeta](https://github.com/tsapeta))
 - Add a `caught` source to the private `reportError` for errors reported from user code. ([#47871](https://github.com/expo/expo/pull/47871) by [@tsapeta](https://github.com/tsapeta))
--- a/packages/expo-app-metrics/CHANGELOG.md
+++ b/packages/expo-app-metrics/CHANGELOG.md
@@ -17,7 +17,7 @@
 
 ### 💡 Others
 
-- [Android] Load only requested metric and log rows when preparing observability payloads.
+- [Android] Load only requested metric and log rows when preparing observability payloads. ([#49011](https://github.com/expo/expo/pull/49011) by [@Ubax](https://github.com/Ubax))
 - Rename the no-update `downloadComplete` state event to `downloadCompleteUnavailable`. ([#47902](https://github.com/expo/expo/pull/47902) by [@kudo](https://github.com/kudo))
 - [iOS] Measure the JS bundle load time against the app startup end marker to stay compatible with upcoming React Native versions. ([#47782](https://github.com/expo/expo/pull/47782) by [@tsapeta](https://github.com/tsapeta))
 - Add a `caught` source to the private `reportError` for errors reported from user code. ([#47871](https://github.com/expo/expo/pull/47871) by [@tsapeta](https://github.com/tsapeta))
--- a/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/storage/MetricsDatabase.kt
+++ b/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/storage/MetricsDatabase.kt
@@ -219,6 +219,9 @@ interface MetricDao {
   @Delete
   suspend fun delete(metrics: List<Metric>)
 
+  @Query("SELECT * FROM metrics WHERE metricId IN (:metricIds)")
+  suspend fun getByIds(metricIds: List<String>): List<Metric>
+
   @Query("SELECT * FROM metrics WHERE sessionId = :sessionId ORDER BY timestamp ASC")
   suspend fun getMetricsForSession(sessionId: String): List<Metric>
 }
@@ -230,6 +233,9 @@ interface LogDao {
 
   @Delete
   suspend fun delete(logs: List<LogRecord>)
+
+  @Query("SELECT * FROM logs WHERE logId IN (:logIds)")
+  suspend fun getByIds(logIds: List<String>): List<LogRecord>
 
   @Query("DELETE FROM logs WHERE timestamp < :cutoffTimestamp")
   suspend fun deleteLogsOlderThan(cutoffTimestamp: String)
@@ -268,6 +274,9 @@ interface SessionDao {
 
   @Query("SELECT * FROM sessions WHERE id = :id")
   suspend fun getById(id: String): Session?
+
+  @Query("SELECT * FROM sessions WHERE id IN (:ids)")
+  suspend fun getByIds(ids: List<String>): List<Session>
 
   // The most recent session other than `:currentSessionId` (null matches all
   // rows, so it returns the latest of any).
@@ -316,12 +325,4 @@ interface SessionDao {
   @Transaction
   @Query("SELECT * FROM sessions WHERE id = :id")
   suspend fun getSessionWithMetricsBySessionId(id: String): SessionWithMetrics?
-
-  @Transaction
-  @Query("SELECT DISTINCT s.* FROM sessions s INNER JOIN metrics m ON s.id = m.sessionId WHERE m.metricId IN (:metricIds)")
-  suspend fun getSessionsWithMetricsByMetricIds(metricIds: List<String>): List<SessionWithMetrics>
-
-  @Transaction
-  @Query("SELECT DISTINCT s.* FROM sessions s INNER JOIN logs l ON s.id = l.sessionId WHERE l.logId IN (:logIds)")
-  suspend fun getSessionsWithLogsByLogIds(logIds: List<String>): List<SessionWithLogs>
 }
--- a/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/storage/MetricsDatabase.kt
+++ b/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/storage/MetricsDatabase.kt
@@ -125,6 +125,7 @@ data class SessionWithMetrics(
     entityColumn = "sessionId"
   )
   val metrics: List<Metric>,
+  /** Only populated by relation-backed `SessionDao` queries. */
   @Relation(
     parentColumn = "id",
     entityColumn = "sessionId"
@@ -160,11 +161,7 @@ data class LogRecord(
 )
 
 data class SessionWithLogs(
-  @Embedded val session: Session,
-  @Relation(
-    parentColumn = "id",
-    entityColumn = "sessionId"
-  )
+  val session: Session,
   val logs: List<LogRecord>
 )
 
@@ -219,7 +216,7 @@ interface MetricDao {
   @Delete
   suspend fun delete(metrics: List<Metric>)
 
-  @Query("SELECT * FROM metrics WHERE metricId IN (:metricIds)")
+  @Query("SELECT * FROM metrics WHERE metricId IN (:metricIds) ORDER BY timestamp ASC")
   suspend fun getByIds(metricIds: List<String>): List<Metric>
 
   @Query("SELECT * FROM metrics WHERE sessionId = :sessionId ORDER BY timestamp ASC")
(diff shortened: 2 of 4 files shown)


Ubax:
@expo-bot review

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/storage/MetricsDatabase.kt
```diff
@@ -219,6 +219,9 @@ interface MetricDao {
   @Delete
   suspend fun delete(metrics: List<Metric>)
 
+  @Query("SELECT * FROM metrics WHERE metricId IN (:metricIds)")
```
jakub-agent: [💡 suggestion] These queries have no `ORDER BY`, which quietly changes the row order of the dispatch payload.

The old path loaded metrics through `@Relation`, i.e. `WHERE sessionId IN (...)`, so SQLite walked the `sessionId` index and returned rows in rowid order — effectively insertion order, which for metrics is chronological. Filtering by `metricId` walks the primary-key index instead, so rows now come back in lexicographic UUID order, i.e. random.

Nothing downstream sorts or truncates (`ObservabilityManager.dispatchUnsentMetrics` maps straight to `EASMetric`), so this is not a correctness break. But `getMetricsForSession` two lines below already specifies `ORDER BY timestamp ASC`, and matching that here would keep payloads chronological, make the output deterministic, and make ordering assertable in tests — the new tests compare with `toSet()`, so order is currently unspecified and unverified.

Same applies to the `logs` variant below.

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/storage/SessionManager.kt
```diff
@@ -223,53 +223,36 @@ class SessionManager(
...
-        sessionWithLogs.copy(logs = sessionWithLogs.logs.filter { it.logId in logIdSet })
+    val logsBySessionId = logIds
+      .chunked(SQLITE_MAX_BIND_VARIABLES)
+      .flatMap { database.logDao().getByIds(it) }
+      .groupBy { it.sessionId }
+    val sessionsById = getSessionsByIds(logsBySessionId.keys).associateBy { it.id }
+
+    return logsBySessionId.mapNotNull { (sessionId, logs) ->
+      sessionsById[sessionId]?.let { session ->
+        SessionWithLogs(session = session, logs = logs)
```
jakub-agent, line 252: [💡 suggestion] With `getSessionsWithLogsByLogIds` gone, no Room DAO returns `SessionWithLogs` anymore — it is only constructed here, in Kotlin. That makes its `@Embedded` and `@Relation` annotations in `MetricsDatabase.kt` dead metadata, and it can become a plain data class.

Worth removing them in this PR: leaving them in suggests the type is still relation-loaded, which is exactly the behaviour the change is moving away from.

`SessionWithMetrics` is different — `getSessionWithMetricsBySessionId` still relation-loads it, so its annotations (including the `logs` default) must stay.

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/storage/SessionManager.kt
```diff
@@ -223,53 +223,36 @@ class SessionManager(
   }
 
   suspend fun getSessionsWithMetrics(metricIds: List<String>): List<SessionWithMetrics> {
-    val metricIdSet = metricIds.toSet()
-    if (metricIds.size <= SQLITE_MAX_BIND_VARIABLES) {
-      return database.sessionDao().getSessionsWithMetricsByMetricIds(metricIds).map { sessionWithMetrics ->
-        sessionWithMetrics.copy(metrics = sessionWithMetrics.metrics.filter { it.metricId in metricIdSet })
+    val metricsBySessionId = metricIds
+      .chunked(SQLITE_MAX_BIND_VARIABLES)
+      .flatMap { database.metricDao().getByIds(it) }
```
jakub-agent, line 229: [🐧 nit] The old code deduplicated (`.distinctBy { it.metricId }` in the chunked branch; the relation-load branch was inherently distinct). This path has no dedup, so a duplicated ID in `metricIds` yields the same `Metric` twice — once per chunk it lands in — and it is then sent twice and counted twice in `dispatchedMetricIds`.

Harmless for the only production caller: `metricId` is the primary key of `pending_metrics`, so `getAllPendingMetricIds()` cannot return duplicates. Flagging it because `SessionManager` is public API, and the invariant is now implicit rather than enforced. A `metricIds.distinct().chunked(...)` would make it explicit at negligible cost.

jakub-agent (commented):

The approach is right, and the net effect is a clear win: `WHERE metricId IN (...)` hits the `metricId` primary-key index and loads exactly the rows the dispatcher needs, instead of pulling every metric of every touched session through a `@Relation`. Dropping the two `<= SQLITE_MAX_BIND_VARIABLES` fast paths in favour of one always-chunked path also removes the branch that had to be tested twice, and `getSessionsWithLogs` gains direct test coverage it did not have before.

A few smaller points, none blocking:

1. The new `getByIds` queries have no `ORDER BY`, so row order changes from insertion order to primary-key (UUID) order. Details inline.
2. `SessionWithLogs` is no longer the return type of any Room query after this change, so its `@Embedded` / `@Relation` annotations are now dead metadata.
3. The removal of `distinctBy` means duplicate input IDs now produce duplicate output rows. Safe for the only production caller, but worth a thought since `SessionManager` is public.

One thing I checked and found fine: `getSessionsWithMetricsByMetricIds` was `@Transaction`, and the replacement runs two queries outside a transaction. A session row can disappear between the two reads, and the `mapNotNull` then silently drops its metrics. That is still correct, because the `sessions` foreign key cascades — a deleted session means its metrics are deleted too, so dropping them and letting `ObservabilityManager` reap the pending IDs as orphans is the right outcome.

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/storage/MetricsDatabase.kt
Ubax: Apply the change

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/storage/SessionManager.kt
Ubax: Remove

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/storage/SessionManager.kt
Ubax: Apply `metricIds.distinct().chunked(...)`

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/storage/SessionManager.kt
```diff
@@ -223,53 +223,40 @@ class SessionManager(
...
+      .distinct()
+      .chunked(SQLITE_MAX_BIND_VARIABLES)
+      .flatMap { database.metricDao().getByIds(it) }
+      .sortedBy { it.timestamp }
+      .groupBy { it.sessionId }
+    val sessionsById = getSessionsByIds(metricsBySessionId.keys).associateBy { it.id }
+
+    return metricsBySessionId.mapNotNull { (sessionId, metrics) ->
+      sessionsById[sessionId]?.let { session ->
+        SessionWithMetrics(session = session, metrics = metrics)
```
tsapeta, line 236: [🏗️ **design**] `SessionWithMetrics.logs` is now populated inconsistently depending on which API produced the value, and nothing in the type says so.

Before this change, the small-input branch returned `sessionWithMetrics.copy(metrics = ...)`, which preserved the Room-loaded `logs` relation, so `getSessionsWithMetrics` did populate `.logs` for batches under the bind limit. Now it always constructs a fresh `SessionWithMetrics` without a `logs` argument, so `.logs` is always `emptyList()` here — while `getSessionWithMetricsBySessionId` still loads it via `@Relation`.

No production caller reads `.logs` off a `SessionWithMetrics` today (`ObservabilityManager` only touches `.session` and `.metrics`), so this is latent rather than a live bug, and the new `does not load logs` test pins the behavior. But the next person who reads `.logs` off a value from this function gets a silently empty list.

Worth either a doc comment on the `logs` field noting it is only populated by the relation-backed queries, or splitting the metrics-only projection into its own type so the field can't be reached at all. The latter is cleaner given `SessionWithLogs` just lost its Room annotations for the same reason.

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/storage/SessionManager.kt
```diff
@@ -223,53 +223,40 @@ class SessionManager(
   }
 
   suspend fun getSessionsWithMetrics(metricIds: List<String>): List<SessionWithMetrics> {
-    val metricIdSet = metricIds.toSet()
-    if (metricIds.size <= SQLITE_MAX_BIND_VARIABLES) {
-      return database.sessionDao().getSessionsWithMetricsByMetricIds(metricIds).map { sessionWithMetrics ->
-        sessionWithMetrics.copy(metrics = sessionWithMetrics.metrics.filter { it.metricId in metricIdSet })
+    val metricsBySessionId = metricIds
```
tsapeta, line 226: [💡 suggestion] These two functions are now byte-for-byte the same algorithm over different tables: distinct, chunk, fetch, sort by timestamp, group by `sessionId`, hydrate sessions, zip. The only differences are the DAO call and the result constructor.

Since both row types already share `sessionId` and `timestamp`, a small private helper parameterized on the fetch and the constructor would remove the duplication:

```kotlin
private suspend fun <Row, Result> groupBySession(
  ids: List<String>,
  fetch: suspend (List<String>) -> List<Row>,
  sessionIdOf: (Row) -> String,
  timestampOf: (Row) -> String,
  build: (Session, List<Row>) -> Result
): List<Result>
```

Not blocking — two copies of ten lines is defensible, and the iOS side solves it with exactly this shape (`Observability.swift`'s `buildPerSession`). Mentioning it because that precedent exists and a third caller (spans) already open-codes the same grouping in `ObservabilityManager`.

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/storage/MetricsDatabase.kt
```diff
@@ -219,6 +215,9 @@ interface MetricDao {
   @Delete
   suspend fun delete(metrics: List<Metric>)
 
+  @Query("SELECT * FROM metrics WHERE metricId IN (:metricIds) ORDER BY timestamp ASC")
```
tsapeta, line 219: [🐧 nit] `ORDER BY timestamp ASC` here is now redundant work. The caller re-sorts the concatenated result with `.sortedBy { it.timestamp }` because per-chunk ordering doesn't survive `flatMap`, so SQLite is sorting each chunk only for Kotlin to sort the whole thing again.

Dropping the `ORDER BY` from `getByIds` (both here and in `LogDao`) would let SQLite serve these straight from the primary-key index. Keeping it is also fine if you'd rather the DAO stay self-describing — just noting the sort is paid twice.

packages/expo-app-metrics/CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 
 ### 💡 Others
 
+- [Android] Load only requested metric and log rows when preparing observability payloads. ([#49011](https://github.com/expo/expo/pull/49011) by [@Ubax](https://github.com/Ubax))
```
tsapeta: [🐧 nit] Platform tag casing: this file's other Android entry uses lowercase `[android]` (see the `UnsupportedOperationException` bug-fix entry), and the PR title uses `[android]` too. Worth matching for consistency.

tsapeta (approved):

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/storage/SessionManager.kt
Ubax: I'll add the comment

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/storage/SessionManager.kt
Ubax: I'll create a ticket to unify as much of the logic as possible between metrics and events

PR #49012, opened 2026-08-17
[observe][android] Dispatch pending metrics and logs in chunks

Ubax:
# Why

If a lot of metrics/events have accumulated, the request sending all of them may be rejected by the server. Additionally loading all of them may cause OOM exception.

# How

1. Change the access layer to pending metrics/events db - instead of a single `getAll`, add `has`, `get` (with limit), and `removeAll`
2. Chunk metrics/events when sending and continue until all metrics are dispatched
3. Change the conflict resolution of the background worker to `KEEP` - instead of aborting in flight request and the dispatch loop, it will keep the existing worker. Previously we used `REPLACE` to ensure that the newest data is sent and then removed from pending metrics. Now since chunking always loads the oldest not dispatched metrics, we can keep the existing worker and let it fetch data.
4. Add observe-tester screen for adding large quantities of events

# Test Plan

1. CI
2. Observe-tester

# Checklist

### DIFF
--- a/apps/observe-tester/app/(tabs)/examples/event-flood.tsx
+++ b/apps/observe-tester/app/(tabs)/examples/event-flood.tsx
@@ -8,7 +8,7 @@ import { useTheme } from '@/utils/theme';
 
 export default function EventFlood() {
   const theme = useTheme();
-  const [lastRun, setLastRun] = useState<{ count: number; elapsedMs: number } | null>(null);
+  const [status, setStatus] = useState<string | null>(null);
 
   function logFlood(count: number) {
     const startedAt = performance.now();
@@ -35,11 +35,19 @@ export default function EventFlood() {
       });
     }
 
-    setLastRun({ count, elapsedMs: Math.round(performance.now() - startedAt) });
+    setStatus(
+      `Logged ${count.toLocaleString()} events in ${Math.round(performance.now() - startedAt).toLocaleString()} ms`
+    );
   }
 
   async function clearStoredEntries() {
-    await AppMetrics.clearStoredEntries();
+    try {
+      await AppMetrics.clearStoredEntries();
+      setStatus('Cleared stored entries');
+    } catch (error) {
+      const message = error instanceof Error ? error.message : String(error);
+      setStatus(`Failed to clear stored entries: ${message}`);
+    }
   }
 
   return (
@@ -53,11 +61,7 @@ export default function EventFlood() {
       <Button title="Log 1,000 events" onPress={() => logFlood(1_000)} />
       <Button title="Log 5,000 events" onPress={() => logFlood(5_000)} />
       <Button title="Clear stored entries" onPress={clearStoredEntries} theme="secondary" />
-      {lastRun ? (
-        <Text style={[styles.status, { color: theme.text.default }]}>
-          Logged {lastRun.count.toLocaleString()} events in {lastRun.elapsedMs.toLocaleString()} ms
-        </Text>
-      ) : null}
+      {status ? <Text style={[styles.status, { color: theme.text.default }]}>{status}</Text> : null}
     </ScrollView>
   );
 }
--- a/apps/observe-tester/app/(tabs)/examples/event-flood.tsx
+++ b/apps/observe-tester/app/(tabs)/examples/event-flood.tsx
@@ -1,0 +1,81 @@
+import AppMetrics from 'expo-app-metrics';
+import { Observe } from 'expo-observe';
+import { useState } from 'react';
+import { Platform, ScrollView, StyleSheet, Text } from 'react-native';
+
+import { Button } from '@/components/Button';
+import { useTheme } from '@/utils/theme';
+
+export default function EventFlood() {
+  const theme = useTheme();
+  const [lastRun, setLastRun] = useState<{ count: number; elapsedMs: number } | null>(null);
+
+  function logFlood(count: number) {
+    const startedAt = performance.now();
+
+    for (let index = 0; index < count; index++) {
+      Observe.logEvent('stress.driver_event', {
+        severity: 'info',
+        body: 'Driver telemetry update received',
+        attributes: {
+          driverId: 'driver_48291',
+          vehicleId: 'vehicle_731',
+          tripId: 'trip_2026_08_14_1842',
+          mode: 'delivery',
+          isOnline: true,
+          speedKph: 42.7,
+          location: {
+            latitude: 37.7751,
+            longitude: -122.4193,
+            accuracyMeters: 8.4,
+          },
+          route: ['warehouse', 'pickup', 'dropoff'],
+          index,
+        },
+      });
+    }
+
+    setLastRun({ count, elapsedMs: Math.round(performance.now() - startedAt) });
+  }
+
+  async function clearStoredEntries() {
+    await AppMetrics.clearStoredEntries();
+  }
+
+  return (
+    <ScrollView
+      style={[styles.container, { backgroundColor: theme.background.screen }]}
+      contentContainerStyle={styles.content}>
+      <Text style={[styles.description, { color: theme.text.secondary }]}>
+        Log a burst of driver telemetry events synchronously to reproduce high-volume integrations.
+      </Text>
+      <Button title="Log 100 events" onPress={() => logFlood(100)} />
+      <Button title="Log 1,000 events" onPress={() => logFlood(1_000)} />
+      <Button title="Log 5,000 events" onPress={() => logFlood(5_000)} />
+      <Button title="Clear stored entries" onPress={clearStoredEntries} theme="secondary" />
+      {lastRun ? (
+        <Text style={[styles.status, { color: theme.text.default }]}>
+          Logged {lastRun.count.toLocaleString()} events in {lastRun.elapsedMs.toLocaleString()} ms
+        </Text>
+      ) : null}
+    </ScrollView>
+  );
+}
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+  },
+  content: {
+    padding: 20,
+    paddingBottom: Platform.select({ ios: 30, android: 150 }),
+  },
+  description: {
+    fontSize: 14,
+    marginBottom: 20,
+  },
+  status: {
+    fontSize: 14,
+    fontWeight: '600',
+  },
+});
--- a/packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityBackgroundWorker.kt
+++ b/packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityBackgroundWorker.kt
@@ -119,7 +119,8 @@ class ObservabilityBackgroundWorker(
         .getInstance(context)
         .enqueueUniqueWork(
           WORK_NAME,
-          ExistingWorkPolicy.REPLACE,
+          // Keep an in-flight dispatch; cancelling it can duplicate a request the server received.
+          ExistingWorkPolicy.KEEP,
           periodicWork
         )
     }
--- a/packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
+++ b/packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
@@ -189,9 +189,6 @@ class BaseObservabilityManager(
         val result = eventDispatcher.dispatch(events)
         metricsRetryGate = nextGate(metricsRetryGate, result)
         val dispatchedMetricIds = sessionsWithPendingMetrics.flatMap { it.metrics }.map { it.metricId }
-        if (DispatchUtils.shouldRemovePending(result)) {
-          pendingMetricsManager.removePendingMetrics(dispatchedMetricIds)
-        }
         when (result) {
           is DispatchResult.PartialSuccess ->
             Log.w(
@@ -207,7 +204,12 @@ class BaseObservabilityManager(
             )
           is DispatchResult.Success, is DispatchResult.RetryableFailure -> Unit
         }
-        if (result is DispatchResult.RetryableFailure) {
+        if (!DispatchUtils.shouldRemovePending(result)) {
+          break
+        }
+        pendingMetricsManager.removePendingMetrics(dispatchedMetricIds)
+        // A non-retryable rejection is likely systematic, so leave the rest for the next dispatch run.
+        if (result is DispatchResult.NonRetryableFailure) {
           break
         }
       }
@@ -260,9 +262,6 @@ class BaseObservabilityManager(
         val result = eventDispatcher.dispatchLogs(events)
         logsRetryGate = nextGate(logsRetryGate, result)
         val dispatchedLogIds = sessionsWithPendingLogs.flatMap { it.logs }.map { it.logId }
-        if (DispatchUtils.shouldRemovePending(result)) {
-          pendingLogsManager.removePendingLogs(dispatchedLogIds)
-        }
         when (result) {
           is DispatchResult.PartialSuccess ->
             Log.w(
@@ -278,11 +277,15 @@ class BaseObservabilityManager(
             )
           is DispatchResult.Success, is DispatchResult.RetryableFailure -> Unit
         }
-        if (result is DispatchResult.RetryableFailure) {
+        if (!DispatchUtils.shouldRemovePending(result)) {
+          break
+        }
+        pendingLogsManager.removePendingLogs(dispatchedLogIds)
+        // A non-retryable rejection is likely systematic, so leave the rest for the next dispatch run.
+        if (result is DispatchResult.NonRetryableFailure) {
           break
         }
       }
-
     }
   }
 
--- a/packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
+++ b/packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
@@ -8,6 +8,8 @@ import expo.modules.observe.storage.PendingMetricsManager
 import expo.modules.appmetrics.storage.SessionManager
 import expo.modules.appmetrics.utils.TimeUtils
 import expo.modules.interfaces.constants.ConstantsInterface
+import kotlinx.coroutines.currentCoroutineContext
+import kotlinx.coroutines.isActive
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
 
@@ -163,7 +165,7 @@ class BaseObservabilityManager(
       return
     }
 
-    while (true) {
+    while (currentCoroutineContext().isActive) {
       val pendingIds = pendingMetricsManager.getPendingMetricIds(dispatchChunkSize)
       if (pendingIds.isEmpty()) {
         break
@@ -234,7 +236,7 @@ class BaseObservabilityManager(
       return
     }
 
-    while (true) {
+    while (currentCoroutineContext().isActive) {
       val pendingIds = pendingLogsManager.getPendingLogIds(dispatchChunkSize)
       if (pendingIds.isEmpty()) {
         break
--- a/packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
+++ b/packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
@@ -81,7 +81,8 @@ class BaseObservabilityManager(
   private val deterministicUniformValueProvider: () -> Double = {
     EASClientID.deterministicUniformValue(EASClientID(context).uuid)
   },
-  private val currentTimeMs: () -> Long = { TimeUtils.getWallClockMillis() }
+  private val currentTimeMs: () -> Long = { TimeUtils.getWallClockMillis() },
+  private val dispatchChunkSize: Int = DISPATCH_CHUNK_SIZE
 ) {
   private val eventDispatcher = EventDispatcher(
     context = context,
@@ -149,8 +150,7 @@ class BaseObservabilityManager(
   )
 
   suspend fun dispatchUnsentMetrics(): Unit = metricsDispatchMutex.withLock {
-    val pendingIds = pendingMetricsManager.getAllPendingMetricIds()
-    if (pendingIds.isEmpty()) {
+    if (!pendingMetricsManager.hasPendingMetrics()) {
       return
     }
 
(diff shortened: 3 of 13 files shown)


Ubax:
@expo-bot review

packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
```diff
@@ -159,50 +159,58 @@ class BaseObservabilityManager(
...
+            metadata = Metadata.fromSessionMetadata(sessionWithMetrics.session),
+            metrics = sessionWithMetrics.metrics.map { EASMetric.fromMetric(it) }
+          )
+        }
+
+        val result = eventDispatcher.dispatch(events)
+        metricsRetryGate = nextGate(metricsRetryGate, result)
+        val dispatchedMetricIds = sessionsWithPendingMetrics.flatMap { it.metrics }.map { it.metricId }
+        if (DispatchUtils.shouldRemovePending(result)) {
+          pendingMetricsManager.removePendingMetrics(dispatchedMetricIds)
```
jakub-agent: [🏗️ **design**] The loop treats `NonRetryableFailure` as "drop this chunk and keep going", and `DispatchUtils.nextRetryGateState` resets `consecutiveRetryableFailures` to 0 for it without setting `dispatchAfterMs`. So a rejection that is systematic rather than per-payload — a validation change, a bad `projectId`, an attribute the server refuses — produces ⌈N/200⌉ back-to-back POSTs that all fail, with no gate and no delay between them.

Before this PR that was one request. The PR body's premise is that "the request sending all of them may be rejected by the server"; this turns that single rejection into hundreds per device, fleet-wide, on a payload problem that affects every device on an app version.

Same in `dispatchUnsentLogs`.

A per-run bound would cover this and one other case: `dispatchUnsentMetrics()` is reachable from JS via `Observe.dispatchEvents()` (`ObserveModule.kt:61`), so `await Observe.dispatchEvents()` now resolves in time proportional to the whole backlog instead of after one request. Two options:

- Track consecutive `NonRetryableFailure` results and break after a small number, letting the next run pick up the rest.
- Cap chunks per run (`MAX_CHUNKS_PER_DISPATCH`) and rely on the next background pass for the remainder.

The cap also bounds the work against the `CoroutineWorker` ~10 minute execution window.

packages/expo-observe/android/src/main/java/expo/modules/observe/storage/ObserveDatabase.kt
```diff
@@ -30,8 +30,14 @@ interface PendingMetricDao {
   @Insert(onConflict = OnConflictStrategy.IGNORE)
   suspend fun insertAll(metrics: List<PendingMetric>)
 
-  @Query("SELECT metricId FROM pending_metrics")
-  suspend fun getAllMetricIds(): List<String>
+  @Query("SELECT metricId FROM pending_metrics ORDER BY addedAt ASC LIMIT :limit")
```
jakub-agent, line 35: [🏗️ **design**] `PendingMetric` and `PendingLog` declare only `@PrimaryKey` on the id column, so there is no index on `addedAt`. SQLite has to scan the whole table to satisfy `ORDER BY addedAt ASC LIMIT :limit`, and the drain loop reissues that query once per chunk. Draining N pending rows at 200 per chunk is ⌈N/200⌉ full-table scans — roughly N²/200 row reads. At 50k pending rows that is ~12.5M reads; at 200k it is ~200M.

That is the large-backlog case this PR exists to fix, so it is worth indexing. `deleteOlderThan` (`WHERE addedAt < :cutoffTimestamp`) benefits from the same index.

```kotlin
@Entity(tableName = "pending_metrics", indices = [Index("addedAt")])
```

This changes Room's schema identity hash, so `version` has to go from 2 to 3 as well — otherwise Room throws `Room cannot verify the data integrity` at open, because `fallbackToDestructiveMigration()` only engages on a version increase. Dropping the pending tables on upgrade is fine here; they hold best-effort ids that `cleanupOldPendingMetrics` already prunes.

packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityBackgroundWorker.kt
```diff
@@ -119,7 +119,8 @@ class ObservabilityBackgroundWorker(
         .getInstance(context)
         .enqueueUniqueWork(
           WORK_NAME,
-          ExistingWorkPolicy.REPLACE,
+          // Keep an in-flight dispatch; cancelling it can duplicate a request the server received.
+          ExistingWorkPolicy.KEEP,
```
jakub-agent, line 123: [🏗️ **design**] The comment captures one side of this trade-off; the other side is worth recording too.

`doWork` catches every exception and returns `Result.retry()`, which puts the request back in `ENQUEUED` with exponential backoff (WorkManager defaults: 30s, doubling, capped at 5h). `KEEP` skips enqueueing whenever existing work with the name is `ENQUEUED`, `RUNNING`, or `BLOCKED` — a request sitting in retry backoff counts. `REPLACE` used to cancel it and start fresh.

`scheduleBackgroundDispatch()` is called only from `OnActivityEntersBackground` (`ObserveModule.kt:58`), so it is the sole scheduling trigger. After a few consecutive worker exceptions the backoff reaches hours, and every subsequent backgrounding is silently dropped — the user can't shorten it by using the app.

Worth either noting the latency cost in the comment, or distinguishing the two states: `KEEP` protects an in-flight dispatch, but a request that is merely enqueued behind a backoff delay is safe to replace.

packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
```diff
@@ -159,50 +159,58 @@ class BaseObservabilityManager(
...
-            "(${result.partial.errorMessage ?: "no error message"})"
-        )
-      is DispatchResult.NonRetryableFailure ->
-        Log.w(
-          OBSERVE_TAG,
-          "Dropping batch of ${dispatchedMetricIds.size} metric event(s): ${result.reason}"
-        )
-      is DispatchResult.Success, is DispatchResult.RetryableFailure -> Unit
+    while (true) {
+      val pendingIds = pendingMetricsManager.getPendingMetricIds(dispatchChunkSize)
```
jakub-agent, line 169: [🏗️ **design**] The loop terminates today, but it depends on an invariant stated twice in two files: `shouldRemovePending` returns `false` exactly for `RetryableFailure` (`DispatchUtils.kt:102`), and the loop's only non-empty exit is `if (result is DispatchResult.RetryableFailure) break`. Every iteration must either delete rows or break, or it re-reads the same oldest chunk forever.

`DispatchResult` is sealed, so adding a variant forces a compile error in `shouldRemovePending`'s `when` — but the `is RetryableFailure` check here just falls through. A new variant that keeps pending rows without being retryable would spin, re-POSTing the same chunk until WorkManager stops the worker at the ~10 minute mark.

Deriving both from one predicate makes progress and termination the same condition:

```kotlin
if (!DispatchUtils.shouldRemovePending(result)) {
  break
}
```

Same in `dispatchUnsentLogs`.

packages/expo-observe/android/src/main/java/expo/modules/observe/storage/PendingLogsManager.kt
```diff
@@ -17,7 +17,11 @@ class PendingLogsManager(
     database.pendingLogDao().insertAll(pendingLogs)
   }
 
-  suspend fun getAllPendingLogIds(): List<String> = database.pendingLogDao().getAllLogIds()
+  suspend fun getPendingLogIds(limit: Int): List<String> = database.pendingLogDao().getLogIds(limit)
```
jakub-agent, line 20: [💡 suggestion] The three new log-side DAO queries — oldest-first `LIMIT`, `EXISTS`, and `DELETE` — have no Room-backed test. `PendingMetricsManagerTest` got real coverage for both (`getPendingMetricIds returns the oldest IDs up to the limit` and `removeAllPendingMetrics deletes all pending metrics`), but there is no `PendingLogsManagerTest` in the tree at all, so the logs path is only exercised through mocks in `BaseObservabilityManagerTest` — which can't catch a wrong `ORDER BY` or a typo'd table name.

The two managers are near-identical, so mirroring the metrics test file is cheap and makes the ordering guarantee the chunking relies on explicit on both sides.

apps/observe-tester/app/(tabs)/examples/event-flood.tsx
```diff
@@ -0,0 +1,81 @@
...
+          index,
+        },
+      });
+    }
+
+    setLastRun({ count, elapsedMs: Math.round(performance.now() - startedAt) });
+  }
+
+  async function clearStoredEntries() {
+    await AppMetrics.clearStoredEntries();
```
jakub-agent: [💡 suggestion] `clearStoredEntries` maps to `sessionManager.clearAllData()`, which deletes from `MetricsDatabase` (`app_metrics`) only. The pending tables live in a separate database — `ObserveDatabase` (`eas_observe`) — and are untouched, so after flooding 5,000 events and tapping this, 5,000 pending log ids survive with no backing rows.

That means the button can't reset the state this PR is about. It does exercise the orphan-cleanup path, which is useful, but a tester can't tell "backlog drained" apart from "backlog orphaned", and the next dispatch spends 25 iterations resolving nothing.

Since the screen exists to reproduce backlog behavior, it would help to either relabel this as "Clear metrics DB (leaves pending ids)" or add a separate control that clears the pending tables too.

packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
```diff
@@ -221,52 +228,61 @@ class BaseObservabilityManager(
...
+                "(${result.partial.errorMessage ?: "no error message"})"
+            )
+          is DispatchResult.NonRetryableFailure ->
+            Log.w(
+              OBSERVE_TAG,
+              "Dropping batch of ${dispatchedLogIds.size} log event(s): ${result.reason}"
+            )
+          is DispatchResult.Success, is DispatchResult.RetryableFailure -> Unit
+        }
+        if (result is DispatchResult.RetryableFailure) {
```
jakub-agent: [🐧 nit] There's a stray blank line between the end of this `if` block and the closing brace of the `while` loop. The metrics loop doesn't have it.

jakub-agent (commented):

packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
Ubax: When `NonRetryableFailure` is received drop the chunk and stop sending. Let next dispatch handle this

packages/expo-observe/android/src/main/java/expo/modules/observe/storage/ObserveDatabase.kt
Ubax: Don't add the index. Just leave a TODO, that we may consider adding index in the future

packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityBackgroundWorker.kt
Ubax: It's not an issue. This is the whole idea, behind backoff - we want it to exponentially backoff in case of server failure

packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
Ubax:
```
if (!DispatchUtils.shouldRemovePending(result)) {
  break
} else {
 remove
}
```

but after the when condtion

apps/observe-tester/app/(tabs)/examples/event-flood.tsx
Ubax: Ignore

packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityBackgroundWorker.kt
tsapeta, line 123: [🚨 **critical**] `KEEP` applies to `ENQUEUED` work, not just `RUNNING` work, and `doWork()` returns `Result.retry()` on any exception. A retry re-enqueues this same unique work with exponential backoff — 30s by default, doubling up to a 5-hour cap. For that entire window every subsequent `OnActivityEntersBackground` → `scheduleBackgroundDispatch()` is silently dropped, so a single transient failure can suppress dispatch far longer than the retry gate in `BaseObservabilityManager` intends. `REPLACE` used to reset this on each backgrounding.

The rationale in the comment also doesn't quite hold: cancelling an in-flight dispatch can't duplicate a request the server received, because pending IDs are only removed *after* `shouldRemovePending(result)` — a cancelled request leaves the rows in place and they get re-sent, which is exactly the same duplicate-on-the-wire risk that existed under `REPLACE`. Chunking doesn't change that either way.

If the goal is "don't cancel a dispatch that's actively running", `APPEND_OR_REPLACE` gets you that without the starvation window. Alternatively keep `KEEP` but drop `Result.retry()` in favor of `Result.success()` (the pending tables already make the work idempotent and the next backgrounding re-schedules it), so there's no long backoff for `KEEP` to sit on.

packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
```diff
@@ -159,50 +159,60 @@ class BaseObservabilityManager(
...
-            "server rejected ${result.partial.rejectedCount} " +
-            "(${result.partial.errorMessage ?: "no error message"})"
-        )
-      is DispatchResult.NonRetryableFailure ->
-        Log.w(
-          OBSERVE_TAG,
-          "Dropping batch of ${dispatchedMetricIds.size} metric event(s): ${result.reason}"
-        )
-      is DispatchResult.Success, is DispatchResult.RetryableFailure -> Unit
+    while (true) {
```
tsapeta: [🏗️ **design**] This loop has no iteration bound and no cancellation check. With `DISPATCH_CHUNK_SIZE = 200`, a 20k-event backlog (very reachable — the new `event-flood` screen logs 5,000 in one tap) is 100 sequential HTTP round-trips in a single call.

Two consequences:

- Inside `ObservabilityBackgroundWorker`, WorkManager stops the worker after 10 minutes. The loop never observes that, so it keeps issuing requests against a stopping worker and the job ends in `retry()` — which, with the `KEEP` change, then blocks re-scheduling for the backoff window.
- Via `dispatchEvents()` from JS, the returned promise stays unresolved for the full drain.

Adding `if (!currentCoroutineContext().isActive) break` at the top of the loop body handles cooperative cancellation for both entry points. A max-chunks-per-run cap would also bound the work and let the next scheduled run pick up where this one left off, which the oldest-first ordering already supports cleanly.

Same applies to the `dispatchUnsentLogs` loop below.

packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
```diff
@@ -159,50 +159,60 @@ class BaseObservabilityManager(
...
+              OBSERVE_TAG,
+              "Dropping batch of ${dispatchedMetricIds.size} metric event(s): ${result.reason}"
+            )
+          is DispatchResult.Success, is DispatchResult.RetryableFailure -> Unit
+        }
+        if (!DispatchUtils.shouldRemovePending(result)) {
+          break
+        }
+        pendingMetricsManager.removePendingMetrics(dispatchedMetricIds)
+        // A non-retryable rejection is likely systematic, so leave the rest for the next dispatch run.
```
tsapeta, line 213: [🏗️ **design**] The loop breaks on `NonRetryableFailure` because "a non-retryable rejection is likely systematic", but `PartialSuccess` falls through and keeps draining. A partial success means the server rejected a subset of the batch, which for this pipeline is usually the same class of systematic problem (schema/field disagreement) as a 400 — so the remaining chunks will re-trip it, one wasted round-trip each, and each one removes its pending rows on the way through.

Either treat `PartialSuccess` like `NonRetryableFailure` here, or add a note explaining why a partial rejection is expected to be transient enough to keep going. As written the asymmetry reads as an oversight rather than a decision.

Same in `dispatchUnsentLogs`.

packages/expo-observe/android/src/test/java/expo/modules/observe/BaseObservabilityManagerTest.kt
```diff
@@ -785,6 +743,224 @@ class BaseObservabilityManagerTest {
...
+      assertEquals(listOf(listOf("metric-1", "metric-2")), removedChunks)
+      coVerify(exactly = 1) { mockEventDispatcher.dispatch(any()) }
+      coVerify(exactly = 1) { mockPendingMetricsManager.getPendingMetricIds(2) }
+    }
+
+  @Test
+  fun `dispatchUnsentMetrics removes an orphaned chunk and continues`() =
+    runTest {
+      val removedChunks = mutableListOf<List<String>>()
+      coEvery { mockPendingMetricsManager.getPendingMetricIds(2) } returnsMany
```
tsapeta: [💡 suggestion] The new chunk tests all stub `getPendingMetricIds` with `returnsMany`, so the mock decides when the drain ends rather than the code under test. That means none of them would catch a regression where the loop stops removing rows and re-fetches the same chunk forever — the mock just runs out of canned answers and returns the last one.

A test backed by a small in-memory fake (a mutable list that `getPendingMetricIds` reads from and `removePendingMetrics` mutates) would exercise the actual termination condition, and would also cover the mixed case this suite currently misses: a chunk that is partly orphaned and partly dispatchable. `PendingMetricsManagerTest` already builds a real Room database, so the pieces are there.

apps/observe-tester/app/(tabs)/examples/event-flood.tsx
```diff
@@ -0,0 +1,81 @@
...
+          route: ['warehouse', 'pickup', 'dropoff'],
+          index,
+        },
+      });
+    }
+
+    setLastRun({ count, elapsedMs: Math.round(performance.now() - startedAt) });
+  }
+
+  async function clearStoredEntries() {
```
tsapeta, line 43: [🐧 nit] `clearStoredEntries` is passed straight to `onPress`, so a rejection here becomes an unhandled promise rejection with no feedback in the UI. The other buttons on this screen report their result via `lastRun`; this one silently does nothing on failure, which is confusing on a screen whose whole purpose is observing storage behavior. Wrapping the await in try/catch and surfacing the error in the same status line would keep the screen self-explanatory.

tsapeta (changes requested):

PR 2/2 in a stack on top of [#49011](https://github.com/expo/expo/pull/49011). The core change is right: replacing `getAll` with a bounded, oldest-first `LIMIT` query removes both the OOM risk and the oversized-request risk, and the `EXISTS`/`DELETE FROM` additions keep the short-circuit paths cheap. The new chunk tests and the `PendingLogsManagerTest` file are a good addition — the logs manager previously had no direct coverage at all.

Three things I'd want resolved before merge:

1. **`ExistingWorkPolicy.KEEP` can starve dispatch for hours.** `doWork()` returns `Result.retry()` on any exception, which re-enqueues the same unique work with exponential backoff (30s doubling to a 5h cap). Under `KEEP`, every later `OnActivityEntersBackground` is a silent no-op until that backoff expires. `REPLACE` reset it. This is the one behavior change in the PR that is strictly worse than before, and the stated rationale doesn't hold — cancelling a dispatch cannot duplicate anything, because the pending rows are only removed after a successful response.
2. **The drain loop is unbounded and non-cancellable.** With `DISPATCH_CHUNK_SIZE = 200`, a 20k backlog is 100 sequential HTTP round-trips inside one `CoroutineWorker` job. WorkManager stops a worker at 10 minutes; the loop never checks `isStopped`/`isActive` and has no iteration cap, so a large backlog can burn the whole budget, get killed, and return `retry()` — which then trips finding 1.
3. **`PartialSuccess` keeps draining, `NonRetryableFailure` stops.** A partial success is usually as systematic as a 400 (schema disagreement), so the loop will re-trip it on every remaining chunk. The asymmetry looks unintentional.

packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityBackgroundWorker.kt
Ubax: I would argue that the `ExistingWorkPolicy.KEEP` is a better approach here. First of all if we think that the 5 hour cap is too long we should probably change this in the backoff policy. This change makes it only more obvious.

Regarding the `cancelling an in-flight dispatch can't duplicate a request the server received` it's not about local requests, but more about canceling ones that already did reach the server (and saved data), but not removing them from pending db, because of cancelation.

Keep is also better for the follow-up PR, since it will keep the backoff for request size be more consistent

packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
Ubax: Good point, I'll add `if (!currentCoroutineContext().isActive) break`

packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
Ubax: I disagree here. I think that `PartialSuccess` is a different kind of error then `NonRetryableFailure`. If part of the data was saved, this means that the general configuration is correct, and there was only an issue in part of metrics/events. For `NonRetryableFailure` it can mean that we have incorrect global config, or something is broken on the server

tsapeta (approved):

packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
Ubax: Added coroutine activity checks to both chunk loops, with cancellation tests that verify undispatched rows remain pending.

packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
Ubax: Keeping `PartialSuccess` distinct from `NonRetryableFailure` as discussed: partial acceptance confirms the global configuration works, so draining can continue.

packages/expo-observe/android/src/test/java/expo/modules/observe/BaseObservabilityManagerTest.kt
Ubax: Reworked the metrics and logs chunk tests around mutable pending-ID stores. Removal now drives loop termination, and the tests cover mixed orphaned and dispatchable IDs in one chunk.

apps/observe-tester/app/(tabs)/examples/event-flood.tsx
Ubax: Updated the screen to report both successful clears and normalized clear errors through the existing status text.

PR #44192, opened 2026-03-24
[ui][iOS] Add `containerBackground` modifier

jakex7:
# Why

It's useful in widgets (but not exclusively).

# How

Added `containerBackground` modifier.

# Test Plan

```tsx
import { Text, VStack } from '@expo/ui/swift-ui';
import { containerBackground } from '@expo/ui/swift-ui/modifiers';
import { createWidget, WidgetEnvironment } from 'expo-widgets';

type MyWidgetProps = {
  emoji: string;
};

const MyWidget = (props: MyWidgetProps, env: WidgetEnvironment) => {
  'widget';
  return (
    <VStack>
      <Text>Time:</Text>
      <Text>{env.date.toLocaleTimeString().split(' ')[0].split(':').slice(0, 2).join(':')}</Text>
      <Text modifiers={[containerBackground('#ff0000', 'widget')]}>Emoji:</Text>
      <Text>{props.emoji}</Text>
    </VStack>
  );
};

export default createWidget('Widget2', MyWidget);
```

SimShot_iPhone_17_Pro_with_Apple_Watch_2026-03-23_22-41-31

# Checklist

- [x] I added a `changelog.md` entry and rebuilt the package sources according to [this short guide](https://github.com/expo/expo/blob/main/CONTRIBUTING.md#-before-submitting)
- [x] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: updated a module plugin).
- [x] Conforms with the [Documentation Writing Style Guide](https://github.com/expo/expo/blob/main/guides/Expo%20Documentation%20Writing%20Style%20Guide.md)

PR #49015, opened 2026-08-17
[ios][widgets] Fix widgets rendering the containerBackground placeholder

### LINKED ISSUE #46200: Expo Widgets - Please adopt containerBackground API

https://github.com/denniske/expo-widgets-repro

### Steps to reproduce

1. Follow the tutorial to create a basic widget: https://docs.expo.dev/versions/v56.0.0/sdk/widgets/#basic-widget
2. `rm -rf ios && npx expo run:ios --device XXX`
3. Add widget to home screen

Image

### Environment

```text
    System:
      OS: macOS 26.5
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 22.22.3 - /Users/dennis/.nvm/versions/node/v22.22.3/bin/node
      Yarn: 1.22.22 - /Users/dennis/.nvm/versions/node/v22.22.3/bin/yarn
      npm: 10.9.8 - /Users/dennis/.nvm/versions/node/v22.22.3/bin/npm
      Watchman: 2024.11.25.00 - /opt/homebrew/bin/watchman
    Managers:
      CocoaPods: 1.15.2 - /Users/dennis/.asdf/shims/pod
    SDKs:
      iOS SDK:
      Android SDK:
        API Levels: 28, 30, 31, 32, 33, 34, 35, 36
        Build Tools: 28.0.3, 29.0.2, 30.0.2, 30.0.3, 32.0.0, 33.0.0, 33.0.1, 34.0.0, 35.0.0, 36.0.0
        System Images: android-30 | Google Play ARM 64 v8a, android-32 | Google APIs ARM 64 v8a, android-32 | Google Play ARM 64 v8a, android-33 | Google APIs ARM 64 v8a, android-33 | Google Play ARM 64 v8a, android-34 | Google Play ARM 64 v8a
    IDEs:
      Android Studio: 2024.2 AI-242.23339.11.2421.12483815
    npmPackages:
      expo: ~56.0.4 => 56.0.4 
      expo-router: ~56.2.6 => 56.2.6 
      react: 19.2.3 => 19.2.3 
      react-dom: 19.2.3 => 19.2.3 
    Expo Workflow: managed
```

### Expo Doctor Diagnostics

```text
```

neutronm:
# Why

Closes #46200.

Every `expo-widgets` home screen widget whose layout doesn't set a background renders as the system
placeholder **"Please adopt containerBackground API"** on iOS 17 and later. The widget in the
[Basic widget](https://docs.expo.dev/versions/latest/sdk/widgets/#basic-widget) doc example
reproduces it, so this is the first thing many people see after following the guide.

iOS 17 requires every home screen widget to call `containerBackground(_:for:)`; WidgetKit replaces
the view of any widget that never does. `WidgetsEntryView` used to call
`.containerBackground(.clear, for: .widget)`, but #44192 removed it so that layouts could set their
own background with the `containerBackground` modifier added to `@expo/ui` in the same PR — an
unconditional call at the entry-view level sits outside the layout tree and overrides the modifier
the layout applies. Removing it fixed the modifier and left every layout that doesn't use it
unrenderable.

This doesn't reproduce in the simulator at any iOS version — the simulator renders widget content
without a container background — so it only shows up once a build reaches a device.

# How

`WidgetsEntryView` applies a fallback container background again, but only when the evaluated layout
declares none of its own, so both cases work:

- a layout that uses the `containerBackground` modifier keeps full control of its background
  (the behavior #44192 added),
- a layout that doesn't renders with a fallback background instead of the placeholder.

`View.fallbackContainerBackground(forLayout:)` (new **ios/Widgets/ContainerBackground.swift**) walks
the evaluated layout tree for a `containerBackground` entry in any node's `modifiers` and applies
`.containerBackground(.fill.tertiary, for: .widget)` only when it finds none. The walk reuses
`flattenChildNodes`, which moves from a private method on `WidgetsDynamicView` to
**ios/Widgets/Utils.swift** unchanged. Looking at the evaluated tree rather than the source means the
check reflects what is actually rendered for the current entry and widget family, so a layout that
sets a background for `systemSmall` only still gets the fallback in the other families.

The same change goes into the entry view the config plugin generates for configurable (`AppIntent`)
widgets, which carries its own copy of this body.

Two edge cases turned out to matter on device, and both are covered:

**A layout that renders nothing still has to satisfy the requirement.** In release a layout resolves
to `EmptyView` whenever it fails to evaluate (`RedBoxView` is behind `#if DEBUG`), roots on a node
type `WidgetsDynamicView` has no case for, or fails to decode its props — and a container background
applied to an empty subtree is discarded along with it, so those widgets showed the placeholder even
with the fallback in place. The content is layered over a zero-sized `Color.clear`, which gives the
background something to attach to while the stack still measures exactly the content, so the layout
of widgets that do render is untouched.

**A `containerBackground` entry in the tree isn't necessarily a background.** `ContainerBackgroundModifier`
does nothing unless both a color and a placement are set, and from iOS 18 it honors the placement it
was given, so a `navigation` placement leaves the widget requirement unsatisfied. Matching on
`$type` alone would let those suppress the fallback and bring the placeholder back, so the check
mirrors the modifier: color present, placement present, and `widget` placement required only on
iOS 18 and later, where iOS 17's coercion of every placement to `.widget` no longer applies.

Two details worth flagging for review:

- **`.fill.tertiary`** is the default Xcode's own widget template uses, so a widget that draws no
  background of its own looks like a standard system widget. It isn't a lock-in: `.clear` was the
  pre-#44192 value, and a layout gets that back with `containerBackground('clear', 'widget')`. Happy
  to switch the fallback to `.clear` if you'd rather keep it visually neutral.
- The fallback is applied for **every widget family**. Container backgrounds stay removable by
  default, so the system still drops them where they don't belong (StandBy, Smart Stack), and a
  layout that wants none in a specific accessory family can branch on `environment.widgetFamily` and
  pass `'clear'`.

Live Activities are untouched — the requirement is specific to widgets, and `WidgetsDynamicView` is
not modified by this PR.

Two known gaps, left for your call rather than decided here:

- The walk descends into `props.children` for every node, but node types rendered without
  `updateChildren` (`ImageView`, `GaugeView`, `ChartView`, `LabelView`, `ProgressView`, the shape
  views) never forward their children to the view tree. A `containerBackground` nested under one of
  those is found by the walk but never applied by the renderer. Teaching the walk which types
  forward children duplicates the renderer's switch, so I'd rather ask where that knowledge belongs.
- `declaresContainerBackground` has no unit test — `expo-widgets` has no iOS `test_spec`, and adding
  one felt like infrastructure work beyond a bug fix. The plugin-side test covers the generated
  sources only.

# Test Plan

**Automated** — new **plugin/src/\_\_tests\_\_/withWidgetSourceFiles-test.ts** covers the generated
widget sources: a plain widget renders through the shared `WidgetsEntryView`, and a configurable
widget's generated entry view applies the fallback container background.

```
$ et check-packages expo-widgets
…
Test Suites: 4 passed, 4 total
Tests:       15 passed, 15 total
…
🏁 All checks passed.
```

**On device** — iPhone 14 Pro / iOS 26.6, release build, deployment target 16.4. The simulator
doesn't reproduce the bug at any iOS version, so all of this was checked on hardware.

Because `main` targets the next SDK and can't be dropped into an SDK 57 app, the device run applied
the same change to the released `57.0.10` build: `ContainerBackground.swift` verbatim,
`WidgetsEntryView` restructured identically, and the plugin's generated entry view patched the same
way. The layout-tree check is version-independent — it only inspects the evaluated node dictionary,
which has the same shape in both.

| Layout | Path | Before | After |
| --- | --- | --- | --- |
| No modifier, static widget | `WidgetsEntryView` | "Please adopt containerBackground API" | **Renders its layout on the fallback background** |
| `containerBackground('#ff0000', 'widget')` on a nested `Text` (the #44192 test case) | `WidgetsEntryView` | Renders, red | **Unchanged — renders red.** An unconditional wrapper turns this grey |
| No modifier, configurable (`AppIntent`) widget, `systemSmall` + `systemMedium` | Generated entry view | "Please adopt containerBackground API" | **Renders its layout on the fallback background** |
| Layout throws in the widget's JS runtime → `EmptyView` in release | Failure path | "Please adopt containerBackground API" | **Blank card on the fallback background** |
| Layout roots on a node type with no native case (`Group`) → `EmptyView` in release | Failure path | "Please adopt containerBackground API" | **Blank card on the fallback background** |
| `containerBackground(undefined, 'widget')` — a modifier that is a runtime no-op | `WidgetsEntryView` | "Please adopt containerBackground API" | **Renders its layout on the fallback background** |

Every row ran in the same build, so the fallback is demonstrably active while still standing aside
for a layout that sets its own background. The rows either side of that boundary are the ones that
matter: the red widget and the no-op-modifier widget would both be grey if the check were wrong in
one direction, and both would show the placeholder if it were wrong in the other.

# Checklist


docs/pages/versions/unversioned/sdk/widgets.mdx
```diff
@@ -325,6 +325,29 @@ import MyWidget from './MyWidget';
 MyWidget.reload();
 ```
 
+#### Widget background
+
+On iOS 17 and later, WidgetKit expects widgets to declare a container background. When your layout doesn't declare one, `expo-widgets` applies a default system fill so the widget still renders.
+
+To use a custom background, apply the `containerBackground` modifier from [`expo/ui`](/versions/latest/sdk/ui/swift-ui/) to any view in the layout.
```
amandeepmittal: ✏️ **style** `terminology.md#expo-product-and-feature-names`

The package is `@expo/ui`, and the import directly below this paragraph uses that name. Change the link text to `@expo/ui`.

```suggestion
To use a custom background, apply the `containerBackground` modifier from [`@expo/ui`](/versions/latest/sdk/ui/swift-ui/) to any view in the layout.
```

docs/pages/versions/unversioned/sdk/widgets.mdx
```diff
@@ -325,6 +325,29 @@ import MyWidget from './MyWidget';
...
+const MyWidget = () => {
+  'widget';
+  return (
+    <VStack modifiers={[containerBackground('#1e293b', 'widget')]}>
+      <Text>Hello</Text>
+    </VStack>
+  );
+};
+
+export default createWidget('MyWidget', MyWidget);
```
amandeepmittal: ✏️ **style** `prose.md#core-principles`

This redefines `MyWidget`, the shared example from [Creating a widget](#prerequisite-creating-a-widget) that the four sections above continue from, with a different shape: no props, no `environment`, and no initial props. Give this example its own name so a reader who already built `MyWidget` does not read it as a replacement.

```suggestion
const BackgroundWidget = () => {
  'widget';
  return (
    <VStack modifiers={[containerBackground('#1e293b', 'widget')]}>
      <Text>Hello</Text>
    </VStack>
  );
};

export default createWidget('BackgroundWidget', BackgroundWidget);
```

amandeepmittal (changes requested):

neutronm:
Addressed both review comments and pushed a follow-up for a few container-background edge cases I found during additional Release-device testing. I’ve also updated the PR description and Test Plan with the additional coverage.

docs/pages/versions/unversioned/sdk/widgets.mdx
neutronm: Updated to @expo/ui, thanks!

docs/pages/versions/unversioned/sdk/widgets.mdx
neutronm: Renamed the example. Thanks! ^__^

amandeepmittal (commented):
Thanks for updating the docs!

Let's wait for the technical review.

neutronm:
Hey @jakex7  — whenever you get a chance, would you mind taking a look at the technical side of this? The docs review is done and the latest version includes the additional Release-device edge-case testing. Thanks!

jakex7:
@expo-bot verify with fable

neutronm:
Hey @jakex7 — just following up on this when you get a chance. 
The Fable verification you requested completed successfully, and it looks like the PR is waiting on the technical review now. Thanks!

neutronm:
Hey @jakex7 — just refreshed this PR against the latest `main` and resolved the merge conflicts that had accumulated since it was opened. It should be ready for technical review again now.

When you get a chance, I’d really appreciate a look. Happy to adjust the implementation if you’d prefer a different approach. Thanks!

PR #49016, opened 2026-08-17
[observe][android] Retry the dispatch on 413

Ubax:
# Why

At the moment we treat response with status `413` as non-retryable. Now with chunking, we can reduce the size of the chunk and retry.

# How

1. Add new result type `PayloadTooLarge`
2. When it is detected reduce the number of metrics/events sent to the server by half and retry

# Test Plan

1. CI
2. Observe-tester

# Checklist

---

### DIFF
--- a/packages/expo-observe/CHANGELOG.md
+++ b/packages/expo-observe/CHANGELOG.md
@@ -17,7 +17,7 @@
 
 ### 💡 Others
 
-- [Android] Retry a dispatch that gets HTTP 413
+- [Android] Retry a dispatch that gets HTTP 413 ([#49016](https://github.com/expo/expo/pull/49016) by [@Ubax](https://github.com/Ubax))
 - [Android] Dispatch pending metrics and logs in bounded, oldest-first chunks without replacing active background work. ([#49012](https://github.com/expo/expo/pull/49012) by [@Ubax](https://github.com/Ubax))
 - Mark the `AppMetrics` export as deprecated in favor of `Observe`. ([#48901](https://github.com/expo/expo/pull/48901) by [@kadikraman](https://github.com/kadikraman))
 
--- a/packages/expo-observe/CHANGELOG.md
+++ b/packages/expo-observe/CHANGELOG.md
@@ -17,6 +17,7 @@
 
 ### 💡 Others
 
+- [Android] Retry a dispatch that gets HTTP 413
 - [Android] Dispatch pending metrics and logs in bounded, oldest-first chunks without replacing active background work. ([#49012](https://github.com/expo/expo/pull/49012) by [@Ubax](https://github.com/Ubax))
 - Mark the `AppMetrics` export as deprecated in favor of `Observe`. ([#48901](https://github.com/expo/expo/pull/48901) by [@kadikraman](https://github.com/kadikraman))
 
--- a/packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
+++ b/packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
@@ -220,7 +220,8 @@ object DispatchUtils {
    *   server-side) doesn't introduce a new pause.
    * - `NonRetryableFailure` also resets the counter. A permanent drop isn't a sign that the
    *   server is unhealthy and shouldn't pause subsequent rounds.
-   * - `PayloadTooLarge` leaves both fields untouched because chunk-size retries are immediate.
+   * - `PayloadTooLarge` also resets the counter because a reachable server returned a definitive
+   *   response, and leaves the gate alone because chunk-size retries are immediate.
    * - `RetryableFailure` increments the counter and sets the gate to `now + delay`, where
    *   `delay` is the server-supplied `retryAfterMs` if present, otherwise `backoff(nextCount)`.
    *
@@ -235,9 +236,9 @@ object DispatchUtils {
   ): RetryGateState = when (result) {
     is DispatchResult.Success,
     is DispatchResult.PartialSuccess,
-    is DispatchResult.NonRetryableFailure ->
+    is DispatchResult.NonRetryableFailure,
+    is DispatchResult.PayloadTooLarge ->
       currentState.copy(consecutiveRetryableFailures = 0)
-    is DispatchResult.PayloadTooLarge -> currentState
     is DispatchResult.RetryableFailure -> {
       val nextCount = currentState.consecutiveRetryableFailures + 1
       val delay = result.retryAfterMs ?: backoff(nextCount)
--- a/packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
+++ b/packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
@@ -11,7 +11,7 @@ import kotlin.math.pow
 import kotlin.random.Random
 
 /**
- * Outcome of a single dispatch attempt to the OTLP endpoint. Four cases, modeled after the
+ * Outcome of a single dispatch attempt to the OTLP endpoint. Five cases, modeled after the
  * OTLP retry guidance (see https://opentelemetry.io/docs/specs/otlp/#otlphttp-response):
  *
  * - `Success` — server accepted the batch without rejections.
@@ -22,6 +22,7 @@ import kotlin.random.Random
  *   it as a drop.
  * - `RetryableFailure` — transient failure (429/502/503/504 or transport error); retry
  *   the same batch after `retryAfterMs` or a client-computed backoff.
+ * - `PayloadTooLarge` — HTTP 413; retry immediately with fewer records.
  * - `NonRetryableFailure` — permanent failure (4xx/5xx outside the retryable set, encoding
  *   error); drop the batch so it can't wedge the queue.
  *
@@ -33,6 +34,7 @@ sealed class DispatchResult {
   object Success : DispatchResult()
   data class PartialSuccess(val partial: OTPartialSuccess) : DispatchResult()
   data class RetryableFailure(val retryAfterMs: Long? = null) : DispatchResult()
+  object PayloadTooLarge : DispatchResult()
   data class NonRetryableFailure(val reason: String) : DispatchResult()
 }
 
@@ -43,7 +45,7 @@ sealed class DispatchResult {
  */
 object DispatchUtils {
   /**
-   * Pure classifier that maps an HTTP response into one of three retry outcomes. Extracted
+   * Pure classifier that maps an HTTP response into a dispatch outcome. Extracted
    * from the dispatch call site so the OTLP-spec rules can be unit-tested without a real
    * network call.
    *
@@ -57,8 +59,6 @@ object DispatchUtils {
     responseBody: String?,
     bodyExcerpt: () -> String = { "" }
   ): DispatchResult {
-    val retryAfter = parseRetryAfter(retryAfterHeader)
-
     if (statusCode in 200..299) {
       // The OTLP spec allows `partial_success` to carry a warning-only payload —
       // `rejectedCount == 0` with a non-empty `errorMessage`. Treat that as a successful send
@@ -78,7 +78,8 @@ object DispatchUtils {
     }
 
     return when (statusCode) {
-      429, 502, 503, 504 -> DispatchResult.RetryableFailure(retryAfter)
+      413 -> DispatchResult.PayloadTooLarge
+      429, 502, 503, 504 -> DispatchResult.RetryableFailure(parseRetryAfter(retryAfterHeader))
       else -> {
         val excerpt = bodyExcerpt()
         val suffix = if (excerpt.isEmpty()) "" else ": $excerpt"
@@ -97,13 +98,14 @@ object DispatchUtils {
    *   permanently, so retrying would produce the same answer; removing them drops the batch
    *   so it can't wedge subsequent rounds. This is the acceptance-criterion behavior: a
    *   400/403 must not be re-sent on the next cycle.
-   * - `RetryableFailure` keeps them so the next dispatch round picks the same rows up again.
+   * - `RetryableFailure` and `PayloadTooLarge` keep them so they can be retried.
    */
   fun shouldRemovePending(result: DispatchResult): Boolean = when (result) {
     is DispatchResult.Success,
     is DispatchResult.PartialSuccess,
     is DispatchResult.NonRetryableFailure -> true
-    is DispatchResult.RetryableFailure -> false
+    is DispatchResult.RetryableFailure,
+    is DispatchResult.PayloadTooLarge -> false
   }
 
   /**
@@ -218,6 +220,7 @@ object DispatchUtils {
    *   server-side) doesn't introduce a new pause.
    * - `NonRetryableFailure` also resets the counter. A permanent drop isn't a sign that the
    *   server is unhealthy and shouldn't pause subsequent rounds.
+   * - `PayloadTooLarge` leaves both fields untouched because chunk-size retries are immediate.
    * - `RetryableFailure` increments the counter and sets the gate to `now + delay`, where
    *   `delay` is the server-supplied `retryAfterMs` if present, otherwise `backoff(nextCount)`.
    *
@@ -234,6 +237,7 @@ object DispatchUtils {
     is DispatchResult.PartialSuccess,
     is DispatchResult.NonRetryableFailure ->
       currentState.copy(consecutiveRetryableFailures = 0)
+    is DispatchResult.PayloadTooLarge -> currentState
     is DispatchResult.RetryableFailure -> {
       val nextCount = currentState.consecutiveRetryableFailures + 1
       val delay = result.retryAfterMs ?: backoff(nextCount)
--- a/packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
+++ b/packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
@@ -98,14 +98,16 @@ object DispatchUtils {
    *   permanently, so retrying would produce the same answer; removing them drops the batch
    *   so it can't wedge subsequent rounds. This is the acceptance-criterion behavior: a
    *   400/403 must not be re-sent on the next cycle.
-   * - `RetryableFailure` and `PayloadTooLarge` keep them so they can be retried.
+   * - `PayloadTooLarge` removes the pending ID because multi-record batches are retried with
+   *   smaller chunks before this check, so reaching it means a single record exceeded the limit.
+   * - `RetryableFailure` keeps them so they can be retried.
    */
   fun shouldRemovePending(result: DispatchResult): Boolean = when (result) {
     is DispatchResult.Success,
     is DispatchResult.PartialSuccess,
-    is DispatchResult.NonRetryableFailure -> true
-    is DispatchResult.RetryableFailure,
-    is DispatchResult.PayloadTooLarge -> false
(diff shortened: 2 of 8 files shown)


packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
```diff
@@ -205,6 +208,14 @@ class BaseObservabilityManager(
               OBSERVE_TAG,
               "Dropping batch of ${dispatchedMetricIds.size} metric event(s): ${result.reason}"
             )
+          is DispatchResult.PayloadTooLarge -> {
+            if (dispatchedMetricIds.size > 1) {
+              chunkSize = dispatchedMetricIds.size / 2
+            } else {
+              Log.w(OBSERVE_TAG, "Dropping metric event that exceeds the server's payload limit")
```
jakub-agent, line 214: [🚨 **critical**] Dropping the record and continuing the loop leaves no circuit breaker for a server that returns 413 for *every* payload — a reverse proxy with `client_max_body_size` set below a single record, a WAF, a misconfigured ingest route. `RetryableFailure` breaks out of the loop and arms the gate; 413 does neither (`nextRetryGateState` returns `currentState` unchanged for it).

Failure scenario, with `DISPATCH_CHUNK_SIZE = 200` and a 1,000-metric backlog against a blanket-413 endpoint: each record costs 200 → 100 → 50 → 25 → 12 → 6 → 3 → 1 = 8 failed POSTs before it reaches this branch and gets dropped. One `dispatchUnsentMetrics()` call issues ~8,000 requests and deletes the entire backlog, one record at a time. Nothing bounds it — `hasPendingMetrics()` was checked once on entry, and the gate is never armed, so the next scheduled round starts over on whatever accumulated since.

Suggest `break`ing after this drop, the same way the `RetryableFailure` case does. At most one record is dropped per dispatch round, the loop can't storm the endpoint, and the genuine case this branch is for — one oversized record blocking the queue — still gets unblocked.

packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
```diff
@@ -276,6 +290,14 @@ class BaseObservabilityManager(
               OBSERVE_TAG,
               "Dropping batch of ${dispatchedLogIds.size} log event(s): ${result.reason}"
             )
+          is DispatchResult.PayloadTooLarge -> {
+            if (dispatchedLogIds.size > 1) {
+              chunkSize = dispatchedLogIds.size / 2
+            } else {
+              Log.w(OBSERVE_TAG, "Dropping log event that exceeds the server's payload limit")
```
jakub-agent, line 301: [🚨 **critical**] Same missing circuit breaker as the metrics path — see the comment on the `PayloadTooLarge` branch in `dispatchUnsentMetrics`. Logs are the more likely trigger in practice, since a log record can carry arbitrary user attributes and a single one can genuinely exceed a server limit.

packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
```diff
@@ -163,8 +163,11 @@ class BaseObservabilityManager(
       return
     }
 
+    var chunkSize = dispatchChunkSize
     while (true) {
-      val pendingIds = pendingMetricsManager.getPendingMetricIds(dispatchChunkSize)
+      val pendingIds = pendingMetricsManager.getPendingMetricIds(chunkSize)
```
jakub-agent, line 170: [🏗️ **design**] Resetting `chunkSize` to the default after every batch means the client never learns the server's limit — it re-discovers it with a wasted 413 on each round, and with more than one wasted request when several halvings are needed.

Concrete case: the server accepts ~3 records, the default is 200. Every successful chunk of 3 is preceded by 6 failed POSTs (200 → 100 → 50 → 25 → 12 → 6 → 3). Draining 1,000 records costs ~2,300 requests instead of ~333, and it stays that way indefinitely because the discovered size is thrown away each iteration.

Suggest keeping the reduced size for the remainder of the loop, resetting only on entry to the next `dispatchUnsentMetrics()` call (or after a chunk succeeds at the current size, if you want it to grow back). Combined with `break`ing on the single-record drop, that also cuts the blanket-413 case from 8 requests per dropped record to 1.

Minor readability point on the same mechanism: reassigning `chunkSize` one line after it is read makes the control flow easy to misread. A separate `var nextChunkSize: Int? = null` set only by the 413 branch would make the override explicit and remove the need for the accompanying comment.

packages/expo-observe/android/src/test/java/expo/modules/observe/BaseObservabilityManagerTest.kt
```diff
@@ -884,6 +1033,66 @@ class BaseObservabilityManagerTest {
...
+      }
+
+      createManager(dispatchChunkSize = 4).dispatchUnsentLogs()
+
+      assertEquals(listOf(4, 2, 4, 4), requestedLimits)
+      assertEquals(listOf(listOf("log-1", "log-2"), listOf("log-3", "log-4")), removedChunks)
+    }
+
+  @Test
+  fun `dispatchUnsentLogs drops a single log that still gets 413`() =
```
jakub-agent, line 1131: [💡 suggestion] The logs path gets 2 of the 5 new tests the metrics path gets — missing are the orphan-removal halving (`halves the dispatched count after removing orphans`), the halve-to-one-then-drop sequence, and the retry-gate assertion.

The two dispatch loops are line-for-line duplicates, so an edit that lands on only one of them is exactly the regression this asymmetry would hide. Worth mirroring at least the halve-to-one-then-drop case, since that is the branch that deletes data.

packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
```diff
@@ -234,6 +237,7 @@ object DispatchUtils {
     is DispatchResult.PartialSuccess,
     is DispatchResult.NonRetryableFailure ->
       currentState.copy(consecutiveRetryableFailures = 0)
+    is DispatchResult.PayloadTooLarge -> currentState
```
jakub-agent: [🐧 nit] Leaving `dispatchAfterMs` untouched is right — the retry is immediate. But `consecutiveRetryableFailures` arguably should reset here, as it does for `NonRetryableFailure`: a 413 is a definitive response that proves the endpoint is reachable and answering, which is the same reasoning the doc comment gives for resetting on a permanent drop.

As written, a counter left at, say, 3 by an earlier outage survives any number of healthy 413 exchanges, so the next transient failure jumps straight to `backoff(4)` instead of `backoff(1)`.

packages/expo-observe/CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 
 ### 💡 Others
 
+- [Android] Retry a dispatch that gets HTTP 413
```
jakub-agent: [🐧 nit] Missing the PR link and author attribution that the two entries below it carry: `([#49016](https://github.com/expo/expo/pull/49016) by [@Ubax](https://github.com/Ubax))`.

jakub-agent (changes requested):

Adds a `PayloadTooLarge` dispatch result for HTTP 413 and retries the chunk at half size. The classification layer is clean, the doc comments stay accurate, and `shouldRemovePending` / `nextRetryGateState` are updated consistently. Test coverage on the metrics path is good.

The concern is the retry loop in `BaseObservabilityManager`. Unlike `RetryableFailure`, a 413 never breaks the loop and never arms the retry gate, and the halved chunk size is discarded after every batch. Against a server that returns 413 for payloads much smaller than `DISPATCH_CHUNK_SIZE = 200`, that combination turns one `dispatchUnsentMetrics()` call into thousands of failed POSTs; against a server that 413s every payload, it also deletes the whole pending queue one record at a time. Two small changes fix both — `break` after the single-record drop, and keep the reduced chunk size for the rest of the loop.

Stack context: targets `@ubax/chunk-logs-and-metrics` and builds directly on the chunking loop from #49012. The halving relies on `getPendingMetricIds` being oldest-first with a `LIMIT`, which holds, and `getSessionsWithMetrics` filters to the requested IDs, so `dispatchedMetricIds.size` never exceeds `chunkSize` and each halving strictly shrinks.

One note outside the diff: iOS `DispatchUtils.swift` has no 413 case, so 413 stays a `nonRetryableFailure` there. Worth a follow-up so the two platforms don't diverge on payload-limit behavior.

packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
Ubax: Do it. But only once we receive the 413 for a single record.

Add comment stating that one record cannot realistcly be bigger then the 413 limit

packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
Ubax: Same here

packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
Ubax: It cannot really happen. The limit is more then 1MB, so 200 should be far below the limit. Add a comment explaining that

packages/expo-observe/android/src/test/java/expo/modules/observe/BaseObservabilityManagerTest.kt
Ubax: Add the missing tests

packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
```diff
@@ -202,6 +207,17 @@ class BaseObservabilityManager(
               OBSERVE_TAG,
               "Dropping batch of ${dispatchedMetricIds.size} metric event(s): ${result.reason}"
             )
+          is DispatchResult.PayloadTooLarge -> {
+            if (dispatchedMetricIds.size > 1) {
+              chunkSize = dispatchedMetricIds.size / 2
```
jakub-agent: [🚨 **critical**] This is dead code as written. `DispatchUtils.shouldRemovePending(PayloadTooLarge)` now returns `false`, so the `if (!DispatchUtils.shouldRemovePending(result)) break` below fires on 413 too — the loop exits right after this assignment, and `chunkSize` is loop-local, so the halved value is never read.

The base branch's `!shouldRemovePending(result)` was equivalent to the old `result is RetryableFailure` check only because `RetryableFailure` was the only `false` case. Adding `PayloadTooLarge` to that set silently conflates two different meanings: "keep the pending rows" (correct for 413) and "stop this dispatch round" (wrong for 413 — the whole point is to retry immediately at half size).

Effect at runtime: a 413 on a multi-record batch ends the round with the rows still pending and the gate unarmed, so the next scheduled round starts over at the full `DISPATCH_CHUNK_SIZE` and gets the same 413. The queue never drains — the exact wedge the halving was meant to prevent.

CI agrees — 6 of the new tests fail on this commit, including `dispatchUnsentMetrics halves the chunk after 413 then resets to the default size` ([run](https://github.com/expo/expo/actions/runs/32025529736/job/95374036345)).

`continue` after the assignment is the minimal fix: it skips both the break check and the unconditional `removePendingMetrics` below, which is what a 413 wants.

```suggestion
              chunkSize = dispatchedMetricIds.size / 2
              continue
```

packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
```diff
@@ -275,6 +296,17 @@ class BaseObservabilityManager(
               OBSERVE_TAG,
               "Dropping batch of ${dispatchedLogIds.size} log event(s): ${result.reason}"
             )
+          is DispatchResult.PayloadTooLarge -> {
+            if (dispatchedLogIds.size > 1) {
+              chunkSize = dispatchedLogIds.size / 2
```
jakub-agent: [🚨 **critical**] Same unreachable-halving problem as the metrics path — see the comment on `chunkSize = dispatchedMetricIds.size / 2`. The three `dispatchUnsentLogs` halving tests fail on this commit for the same reason.

```suggestion
              chunkSize = dispatchedLogIds.size / 2
              continue
```

packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
```diff
@@ -97,13 +98,14 @@ object DispatchUtils {
...
-   * - `RetryableFailure` keeps them so the next dispatch round picks the same rows up again.
+   * - `RetryableFailure` and `PayloadTooLarge` keep them so they can be retried.
    */
   fun shouldRemovePending(result: DispatchResult): Boolean = when (result) {
     is DispatchResult.Success,
     is DispatchResult.PartialSuccess,
     is DispatchResult.NonRetryableFailure -> true
-    is DispatchResult.RetryableFailure -> false
+    is DispatchResult.RetryableFailure,
+    is DispatchResult.PayloadTooLarge -> false
```
jakub-agent: [🏗️ **design**] This value is correct on its own terms — a 413 must not delete the pending rows. The problem is that `shouldRemovePending` is now load-bearing for a second, unrelated decision at the call site (`if (!shouldRemovePending(result)) break`), and the two questions have different answers for 413: keep the rows, but *don't* stop the round.

Once the `continue` fix lands, the coupling is harmless in practice, but it's still a trap for the next result type added to this `when`. Worth either renaming the call-site check to say what it actually gates, or making the loop tail branch on the result type explicitly rather than inferring "should I stop?" from "should I delete?".

jakub-agent (changes requested):

**Iteration 2** — re-reviewed at `9c928b8`. All five points from the previous round are addressed: the single-record drop now `break`s in both paths, `PayloadTooLarge` resets `consecutiveRetryableFailures`, the logs path has matching tests (5 each now), and the changelog entry carries its link and attribution. The rationale comment for keeping the chunk size non-sticky (server limit above 1 MB, so a 413 is exceptional) is reasonable and I'm not pushing on it further.

One blocking issue, introduced by the interaction between this PR and the base branch's refactor of the loop tail. The base now exits the loop with `if (!DispatchUtils.shouldRemovePending(result)) break`. That was equivalent to the old `result is RetryableFailure` check only because `RetryableFailure` was the sole `false` case. This PR adds `PayloadTooLarge` to that set, so the new break also fires on 413 — the loop exits immediately after computing the halved `chunkSize`, and since `chunkSize` is loop-local it is never read again. The halve-and-retry path is unreachable.

CI confirms it: `test` fails with 6 of the new tests red — `dispatchUnsentMetrics halves the chunk after 413 then resets to the default size`, `halves the dispatched count after removing orphans`, `halves to one then drops the oversized metric`, and the three logs equivalents ([run](https://github.com/expo/expo/actions/runs/32025529736/job/95374036345)). The tests describe the intended behavior correctly; the loop no longer produces it.

A `continue` in the `size > 1` branch fixes it — see the inline comment. Everything else in this iteration looks good.

packages/expo-observe/android/src/main/java/expo/modules/observe/DispatchUtils.kt
```diff
@@ -97,12 +98,15 @@ object DispatchUtils {
...
+   * - `PayloadTooLarge` removes the pending ID because multi-record batches are retried with
+   *   smaller chunks before this check, so reaching it means a single record exceeded the limit.
+   * - `RetryableFailure` keeps them so they can be retried.
    */
   fun shouldRemovePending(result: DispatchResult): Boolean = when (result) {
     is DispatchResult.Success,
     is DispatchResult.PartialSuccess,
-    is DispatchResult.NonRetryableFailure -> true
+    is DispatchResult.NonRetryableFailure,
+    is DispatchResult.PayloadTooLarge -> true
```
jakub-agent, line 109: [🏗️ **design**] This answer is correct at the one call site, but it is no longer correct standalone — and the doc comment above says so: "multi-record batches are retried with smaller chunks **before this check**". `DispatchUtils` is introduced as a pure classifier extracted from the call site "so the OTLP-spec rules can be unit-tested without a real network call"; this branch now classifies a result *plus* an unstated caller state.

Called on a 413 for a 200-record batch without the preceding `continue`, it says "delete all 200 records." Nothing catches that: reordering the two `if`s in the loop tail, or adding a third call site, silently destroys a full chunk of telemetry, and `DispatchUtilsShouldRemovePendingTest` asserts the standalone-wrong answer as the contract.

Two ways to close it, either fine:
- Put the discriminator in the type — `PayloadTooLarge(val recordCount: Int)`, or a distinct result for the single-oversized-record case — so the classifier can answer from the result alone.
- Or keep `PayloadTooLarge -> false` and let the `size == 1` branch at the call site do its own `removePendingMetrics` + `break`, which is where the decision actually lives now. That plus the `continue` you added is a complete fix on its own and needs no change here.

packages/expo-observe/android/src/main/java/expo/modules/observe/ObservabilityManager.kt
```diff
@@ -202,14 +207,23 @@ class BaseObservabilityManager(
               OBSERVE_TAG,
               "Dropping batch of ${dispatchedMetricIds.size} metric event(s): ${result.reason}"
             )
+          is DispatchResult.PayloadTooLarge ->
+            if (dispatchedMetricIds.size == 1) {
```
jakub-agent, line 213: [🐧 nit] The 413 case is now handled in two places with inverted conditions — `size == 1` here, `size > 1` in the `if` below the `when`. They have to stay in sync, and neither reads as the complement of the other at a glance (the `size == 0` gap falls through to the removal path silently).

The log is also detached from the action it describes: it says "Dropping metric event…" here, but the drop happens five lines later through the generic `removePendingMetrics(dispatchedMetricIds)`. If `shouldRemovePending` ever flips back for this result, the log keeps claiming a drop that no longer happens.

Folding both back into the single `when` branch — halve-and-`continue` in one arm, log-and-drop in the other — keeps the size test in one place and the message next to the removal. Same applies to the logs path.

jakub-agent (approved):

**Iteration 3** — re-reviewed at `055483d`. The blocking issue is fixed. Moving the halving out of the `when` and into an explicit `if (result is PayloadTooLarge && size > 1) { chunkSize = size / 2; continue }` makes the halved value reachable again, and the `continue` correctly skips both the break check and the unconditional pending removal.

I traced all six previously-failing cases against the new control flow and they now produce the asserted sequences — `[4, 2, 4, 4]` for the reset case, `[4, 1, 4]` for the orphan case, `[2, 1]` for halve-to-one-then-drop, on both the metrics and logs paths. CI agrees: the `test` job is green on this commit. Termination still holds too — `dispatchedIds.size <= chunkSize`, so each `continue` at least halves the next batch, and the `size == 1` case drops the record and breaks.

Two non-blocking points remain, both about where the 413 decision lives rather than what it does. `shouldRemovePending(PayloadTooLarge)` is now `true`, which is the right answer only because the caller already `continue`d for multi-record batches — the classifier's contract has picked up a dependency on caller control flow. And the 413 case is split across two places with inverted size conditions that have to stay in sync. Details inline; neither needs to hold up the merge.

Worth noting for the record: keeping `shouldRemovePending(PayloadTooLarge) = false` and adding only the `continue` would also have worked, and would have left the classifier standalone-correct. Both shapes are defensible — flagging the trade-off, not asking for a rewrite.

tsapeta (approved):

PR #49017, opened 2026-08-17
[iOS][expo-widgets] Add contentDate to LiveActivity.update

LouisRaverdy:
# Why

ActivityKit 17.2 added a timestamped `update` API. The timestamp represents when the data was generated, and the system ignores the update when a newer local update or push payload has already been applied.

`expo-widgets` already exposes this behavior as `contentDate` when ending a Live Activity, but `LiveActivity.update()` only exposes `staleDate`. Apps that combine local updates with ActivityKit push notifications therefore cannot prevent a delayed local update from replacing newer push content.

We already use this implementation through a local `expo-widgets` patch. Upstreaming it removes the need to maintain that patch and makes ActivityKit's ordering behavior available to other apps that combine local and push updates.

# How

- Add an optional third `contentDate` argument to `LiveActivity.update(props, staleDate?, contentDate?)`, preserving the existing API.
- Forward the JavaScript timestamp through the native module as a `Date`.
- Use ActivityKit's timestamped update API on iOS 17.2 and later when `contentDate` is provided.
- Keep the existing update path when `contentDate` is omitted or the device runs an earlier iOS version.

# Test Plan

From `packages/expo-widgets`:

```sh
pnpm run format --check
pnpm run lint
pnpm run typecheck
pnpm run build
pnpm run test --runInBand
```

All checks pass. Jest reports 3 passing suites and 13 passing tests.

# Checklist

### DIFF
--- a/packages/expo-widgets/CHANGELOG.md
+++ b/packages/expo-widgets/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### 🎉 New features
 
-- [iOS] Expose ActivityKit's `contentDate` on `LiveActivity.update()` to prevent older local updates from replacing newer local or push updates.
+- [iOS] Expose ActivityKit's `contentDate` on `LiveActivity.update()` to prevent older local updates from replacing newer local or push updates. ([#49017](https://github.com/expo/expo/pull/49017) by [@LouisRaverdy](https://github.com/LouisRaverdy))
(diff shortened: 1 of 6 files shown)


jakex7:
@expo-bot review

LouisRaverdy:
Hey @jakex7 ! Tell me if I need to merge before or after your review

PR #49028, opened 2026-08-17
[ios][camera] Fix UI freeze when repeatedly mounting and unmounting CameraView

barthap:
# Why

Mounting and unmounting `CameraView` a few times freezes the iOS UI for tens of seconds. Reported in #48780 with a repro at https://github.com/nrgbistro/expo-camera-simulator-freeze-repro.

This is a regression from #44159, which replaced the compile-time `#if !targetEnvironment(simulator)` guards with a runtime device check. Two paths were left unguarded, and both end up waiting on a capture graph that cannot start.

When the graph never starts, every later rebuild of it waits out its own ~9s AVFoundation deadline. Nothing detaches the preview layer on teardown, so `AVCaptureVideoPreviewLayer` does it in `dealloc`, on the main thread, against a session still marked running. That is the stack in the report:

```
CA::Transaction::commit
NSKVODeallocate
AVCaptureVideoPreviewLayer dealloc
AVCaptureSession commitConfiguration
AVCaptureSession _buildAndRunGraph
AVRunLoopCondition _waitInMode
```

Several wedged sessions accumulate, and the waits add up into a freeze.

# How

`stopSession()` now stops the session before it removes inputs and outputs, instead of deconfiguring a live session and stopping it afterwards. The stop also moved above the device-availability guard, which previously returned early and left the session running.

`removeFromSuperview()` captures the session manager and the preview layer strongly, and detaches the layer on the main thread once the session has stopped. It captured `self` weakly before, so teardown could be skipped entirely if React Native released the view first.

`updateCameraIsActive()` checks `hasAvailableCameraDevice` before starting, which is the call site #44159 missed. `startSession()` and `stopSession()` already had that check.

# Test Plan

Added two native unit tests in `packages/expo-camera/ios/Tests/`, run with `et native-unit-tests --packages expo-camera`. Both fail on `main` and pass here. The teardown test measures how long the main thread blocks while the preview layer is released, which is the symptom users see.

Measured the freeze itself on a simulator with no capture device, over 8 mount/unmount cycles: 75.5s wall and 71.4s of main-runloop overrun before, against 4.1s and 18ms after.

One gap worth flagging: I could only reproduce with a simulator that reports no camera. The reporter's simulator does expose one, and I verified the teardown fix against the same wedged-session state instead of their exact environment. It would help if @nrgbistro could confirm against their repro.

### DIFF
--- a/packages/expo-camera/CHANGELOG.md
+++ b/packages/expo-camera/CHANGELOG.md
@@ -13,7 +13,7 @@
 
 ### 🐛 Bug fixes
 
-- [iOS] Fix repeatedly mounting and unmounting `CameraView` freezing the UI when the capture session cannot start, such as on a simulator. The session is now stopped before it is torn down and before the preview layer detaches from it, and it is no longer started at all when no capture device is available.
+- [iOS] Fix repeatedly mounting and unmounting `CameraView` freezing the UI when the capture session cannot start, such as on a simulator. The session is now stopped before it is torn down and before the preview layer detaches from it, and it is no longer started at all when no capture device is available. ([#49028](https://github.com/expo/expo/pull/49028) by [@barthap](https://github.com/barthap))
 - [Android] Use the selected camera to determine video stabilization support. ([#45896](https://github.com/expo/expo/pull/45896) by [@vivekjm](https://github.com/vivekjm))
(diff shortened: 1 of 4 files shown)


alanjhughes (approved):

barthap:
Played with [SimCam](https://simcam.swmansion.com/) to add a camera to my simulator, ran through the NCL Camera screens in bare-expo, and found no issues.

PR #49030, opened 2026-08-17
[video][iOS] Deactivate the audio session after playback stops

huextrat:
# Why

When an `expo-video` player interrupts background audio on iOS, the shared `AVAudioSession` remains active after video playback stops. Because the session is never deactivated with `.notifyOthersOnDeactivation`, interrupted music and podcast apps are not told that they may resume.

This is the iOS follow-up to #48546, split out as requested in that review.

# How

Extend the existing automatic audio-session management in `VideoManager`:

- Track whether `expo-video` activated the audio session.
- Once no video player is playing or waiting to play, schedule session deactivation with `.notifyOthersOnDeactivation`.
- Delay deactivation by 100 ms, coalesce rapid state changes, and re-check the current player state before deactivating.
- Keep the session active while a player participates in Now Playing or while a video view is in Picture in Picture.
- Re-evaluate the session after the last player is released.
- Only mark the session inactive after a successful deactivation so a transient failure can be retried.

A muted player that continues playback intentionally keeps the session active. Muting an `AVPlayer` silences its output but leaves its audio pipeline running, so deactivating at that point can fail with `AVAudioSession.ErrorCode.isBusy`.

# Test Plan

Automated/static checks:

- `et check-packages expo-video`
- Swift parser check for the modified files

Manual device validation still recommended during review:

- Start Spotify or Apple Music, then start an unmuted video using `audioMixingMode: "auto"`; background audio should pause.
- Pause, finish, or release the last video player; background audio should resume.
- Buffering or seeking must not briefly resume background audio.
- Active PiP and Now Playing playback must remain unaffected.
- Muting a still-playing video must not attempt to deactivate the session.

# Checklist

- [x] The change uses the existing automatic audio-session management.
- [x] No new public API is introduced.
- [x] Package build, typecheck, lint, and formatting checks pass.
- [ ] Physical-device scenarios above have been revalidated with this automatic implementation.

### DIFF
--- a/packages/expo-video/ios/VideoManager.swift
+++ b/packages/expo-video/ios/VideoManager.swift
@@ -14,6 +14,9 @@ class VideoManager {
   private var mediaServicesResetObserver: NSObjectProtocol?
   private var videoViews = NSHashTable<VideoView>.weakObjects()
   private let videoPlayers = SynchronizedHashTable<VideoPlayer>(weakObjects: true)
+  private var pendingAudioSessionDeactivation: DispatchWorkItem?
+  private var pictureInPictureVideoViews = Set<ObjectIdentifier>()
+  private var audioSessionIsActive = false
 
   var hasRegisteredPlayers: Bool {
     return !videoPlayers.allObjects.isEmpty
@@ -44,6 +47,7 @@ class VideoManager {
 
   func unregister(videoPlayer: VideoPlayer) {
     videoPlayers.remove(videoPlayer)
+    setAppropriateAudioSessionOrWarn()
   }
 
   func register(videoView: VideoView) {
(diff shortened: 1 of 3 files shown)


packages/expo-video/ios/VideoManager.swift
```diff
@@ -149,10 +172,47 @@ class VideoManager {
...
+        return
+      }
+      self.pendingAudioSessionDeactivation = nil
+      guard self.audioSessionIsActive, !self.isAudioSessionInUse else {
+        return
+      }
+
+      do {
+        try AVAudioSession.sharedInstance().setActive(false, options: [.notifyOthersOnDeactivation])
+        self.audioSessionIsActive = false
```
behenate, line 208: I think we should be more careful about this change. This significantly changes the `expo-video` audio behaviour - the 3rd party player will resume once we release the audio session.  Since `expo-video` has been stable for a long time and is widely used we can't accept this change.

I think we should consider an approach similar to `expo-audio` - automatic session management with manual overrides for starting and stopping the audio session. We'd also need to add audio `doNotMixTransient` audio mode.

behenate (changes requested):

PR #49034, opened 2026-08-17
[router] ensure actions always add optional fields to state

Ubax:
# Why

For some actions properties were not added to the state, even though they should

# How

In each router use `ensureStateType` function to add `type` if it is missing. 

For TabRouters ensure that history is always added

# Test Plan

1. CI
2. Router e2e

# Checklist


packages/expo-router/src/react-navigation/routers/StackRouter.tsx
```diff
@@ -192,6 +193,7 @@ export function StackRouter(options: StackRouterOptions) {
   > = {
     ...BaseRouter,
 
+    // TODO: Keep this value in sync with the `ensureStateType` calls below.
```
jakub-agent, line 194: [💡 suggestion] The drift this TODO warns about can be removed instead of documented. A local constant makes the two values one value:

```ts
const TYPE = 'stack' as const;
// ...
type: TYPE,
// ...
const state = ensureStateType(inputState, TYPE);
```

Same applies to `'tab'` in `TabRouter.tsx`. `router.type` itself is not usable here because `router` is still being defined.

packages/expo-router/src/ui/TabRouter.tsx
```diff
@@ -67,6 +69,14 @@ export function ExpoTabRouter(options: ExpoTabRouterOptions) {
       if (!isSwitching && route.state !== undefined) {
         const selectedRoute = attachRouteState(route, action);
         if (selectedRoute === route) {
+          state = ensureStateType(
+            ensureStateHistory(
+              state,
+              options.backBehavior ?? 'firstRoute',
```
jakub-agent: [💡 suggestion] This duplicates `TabRouter`'s own `backBehavior = 'firstRoute'` default (`react-navigation/routers/TabRouter.tsx:295`). If that default ever changes, this fallback silently disagrees and the reselect path rebuilds a different history than every other action on the same navigator. Exporting the default from `TabRouter.tsx` and importing it here removes the second source of truth.

packages/expo-router/src/react-navigation/routers/StackRouter.tsx
```diff
@@ -233,14 +235,16 @@ export function StackRouter(options: StackRouterOptions) {
...
 
-      return {
-        stale: false,
-        type: 'stack',
-        key: `stack-${nanoid()}`,
-        index: routes.length - 1,
-        routeNames,
-        routes: routes.concat(preloadedRoutes),
-      };
+      return ensureStateType(
```
jakub-agent, line 236: [💡 suggestion] This wraps an object literal that provably has no `type`, so the call cannot do anything beyond what `type: 'stack'` did before — it reads as defensive coverage where none is possible. `DrawerRouter.getRehydratedState` still writes `type: 'drawer'` inline, so the three rehydration paths now disagree stylistically.

If the goal was to cover `getRehydratedState`, the branch that could actually return a state without `type` is the `if (state.stale === false) return state;` early return above (line 202, and `TabRouter.tsx:312`). It is unreachable from `useNavigationBuilder` today, since `getRehydratedState` is only called when the state is not initialized — which is another reason plain `type: 'stack'` is enough here.

packages/expo-router/src/react-navigation/routers/__tests__/StackRouter.test.tsx
```diff
@@ -12,6 +12,73 @@ import { createInitialState } from '../../core/createInitialState';
...
+    );
+
+    expect(result?.state.type).toBe('stack');
+  });
+
+  test('preserves the type from complete RESET state', () => {
+    const state = createState();
+    const result = StackRouter({}).getStateForAction(
+      state,
+      CommonActions.reset({ ...state, type: 'stack' }),
```
jakub-agent, line 61: [💡 suggestion] This test does not discriminate: the payload's `type` equals the type `ensureStateType` would stamp, so it still passes if the helper overwrote unconditionally. Resetting with a different type (`type: 'tab'`) and asserting it survives would test what the name claims — the way the `preserves drawer type when used by DrawerRouter` case in `TabRouter.test.tsx` does.

packages/expo-router/src/layouts/stack-router.ts
```diff
@@ -82,6 +83,8 @@ export const stackRouterOverride: NonNullable<NativeStackNavigatorProps['UNSTABL
 ) => {
   return {
     getStateForAction: (state, action, options) => {
+      state = ensureStateType(state, 'stack');
```
jakub-agent, line 85: [🐧 nit] Nit: reassigning the parameter differs from the pattern the other routers adopted in this PR (`getStateForAction(inputState, ...)` then `const state = ensureStateType(inputState, 'stack')`). Renaming the parameter here keeps the five call sites reading the same way.

jakub-agent (approved):

The change is coherent and well tested. Seeded state (state without `type`, allowed since #48757) is now normalized at every router entry point, so committed state carries `type` again after an action. Removing the ad-hoc `type: 'stack'` writes in the `PRELOAD` branches is safe — `reconcileStackRoutes` and the object spreads all derive from the normalized `state`.

I checked the paths that could have been missed and found none: `NativeBottomTabsRouter` returns only values derived from `TabRouter.getStateForAction`, the `stackRouterOverride` early returns all delegate to `original`, and `getStateForDeclaredRoutes` is render-only and never committed. `ensureStateHistory` is idempotent, so the double application through `DrawerRouter` -> `TabRouter` is fine. Preserving the drawer `type` through `TabRouter`'s `default` branch is a nice touch and is covered by a test.

Comments below are all non-blocking: one duplicated default constant, two spots where `ensureStateType` adds indirection without changing behavior, one test that does not discriminate, and one style nit.

Two things outside the diff:
- No `CHANGELOG.md` entry. Sibling PRs in this series (#48757, #48783) each added one. If the entry is planned for the top of the stack, ignore this.
- `check-packages` is red, but the two failures look unrelated to this diff: an OOM in the `expo-router` jest run (all router suites reported PASS before the crash) and `ReferenceError: __DEV__ is not defined` in `@expo/cli`'s `exportStaticAsync.test.ts`. Worth confirming against the base branch. The `jest-ubuntu` shards were still running when I reviewed, so treat this as a partial signal.

PR #49035, opened 2026-08-17
[router] Fix platform-specific route parsing and loader keys

hassankhan:
# Why

This PR addresses issues found by Musclog [here](https://musclog.app/blog/2026/08/this-blog-was-built-using-expo). 

Platform-specific catch-all routes such as `[...slug].web.tsx` were parsed incorrectly because we split on `.` in `getRoutesCore()`;  the dots in `...slug` were mistakenly split on, preventing Expo Router from recognizing `.web` as the platform suffix. 

Platform suffixes are also preserved by `getContextKey()`, this prevented platform-specific layouts' such as `_layout.web.tsx` from being normalized consistently.

# How

- Added a new `getPlatformFromFilePath()` which recognises one of the following platforms: `android`, `ios`, `native` and `web`.
- Refactored `getRoutesCore()` to use `getPlatformFromFilePath()`.
- Refactored `getContextKey()` to remove the platform as found by `getPlatformFromFilePath()`. This now means that `layout.tsx` and `layout.web.tsx` both become `""` after being passed to `getContextKey()`.

# Test Plan

- CI

# Checklist

### DIFF
--- a/packages/expo-router/src/__tests__/getRoutes.test.ios.ts
+++ b/packages/expo-router/src/__tests__/getRoutes.test.ios.ts
@@ -183,6 +183,25 @@ describe('getRoutes', () => {
     );
   });
 
+  // NOTE(@hassankhan): Should we throw an error for invalid platforms for
+  // `_layout` like we do for `+not-found`?
+  it(`does not treat an unsupported dotted suffix as a layout qualifier`, () => {
+    const routes = getRoutes(
+      inMemoryContext({
+        './_layout.custom.tsx': () => null,
+      }),
+      { internal_stripLoadRoute: true, skipGenerated: true }
+    );
+
+    expect(routes?.children).toContainEqual(
+      expect.objectContaining({
+        contextKey: './_layout.custom.tsx',
+        route: '_layout.custom',
+        type: 'route',
+      })
+    );
+  });
+
   it(`should name routes relative to the closest _layout`, () => {
     expect(
       getRoutes(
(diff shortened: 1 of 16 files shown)


packages/expo-router/src/__tests__/getRoutes.test.ios.ts
```diff
@@ -183,6 +183,25 @@ describe('getRoutes', () => {
     );
   });
 
+  // NOTE(@hassankhan): Should we throw an error for invalid platforms for
+  // `_layout` like we do for `+not-found`?
```
hassankhan, line 187: @kitten @Ubax WDYT? Feels a bit odd that we have an error case for one and not the other.

packages/expo-router/src/matchers.tsx
```diff
@@ -6,6 +6,9 @@ interface DynamicNameMatch {
   deep: boolean;
 }
 
+// NOTE(@hassankhan): Should we retrieve these from a more central location?
+export const VALID_PLATFORMS = new Set(['android', 'ios', 'native', 'web']);
```
kitten, line 10: These will have to be passed on from Metro. I'm not sure why these were added and am not sure we're doing ourselves any favour with hardcoding them here. We may have suitable lists elsewhere but the source of truth would be to likely pass them on from "a central place" that's then also aligned with Metro

kitten (approved):

brentvatne:
@expo-bot verify

PR #49039, opened 2026-08-17
[jsi] Fix ambiguous `abs` in date TimeClip guard under C++ interop

kraenhansen:
# Why

`expo-modules-jsi` fails to compile on the Xcode 26 toolchain with:

```
JavaScriptCodable+Date.swift:53:50: error: type of expression is ambiguous without a type annotation
```

The offending line is the overflow guard in `dateFromMilliseconds`:

```swift
guard milliseconds.isFinite, abs(milliseconds) <= maxJavaScriptDateMilliseconds else {
```

The module builds with C++ interop (`.interoperabilityMode(.Cxx)`, `cxxLanguageStandard: .cxx20`). That appears to bring the C `abs` overloads from `<cstdlib>`/`<cmath>` into scope alongside Swift's `abs`, so the unqualified `abs(_:)` over a `Double` becomes ambiguous and the file no longer type-checks. It's the only unqualified `abs(_:)` call in the module.

The line was introduced in #47602 (`[ios][jsi] Add JavaScriptCodable conformance for Date`, shipped in `expo-modules-jsi@57.0.4`) and hasn't been touched since — so this seems to be a latent ambiguity in that code rather than a later regression. It presumably compiled cleanly on the toolchain CI used at release time; it looks like a newer toolchain (Xcode 26) is what surfaces the C `abs` overloads and tips the call into ambiguity. That would also explain why it wasn't caught in review or CI.

The practical impact is that `expo run:ios` currently fails to build for SDK 57 apps on affected toolchains.

cc @tsapeta (author of #47602) — flagging since it's a recent change; happy to adjust the approach if you'd prefer a different fix (e.g. `Swift.abs`).

# How

Replace `abs(milliseconds)` with `milliseconds.magnitude`. For a `Double`, `.magnitude` is `Double` and unambiguous, and it is behavior-identical to `abs` here (including for `-0.0`, infinities, and NaN — NaN is already excluded by the preceding `isFinite` check).

# Test Plan

- [x] Confirmed the file fails to compile before the change and compiles cleanly after, by building the `ExpoModulesJSI` xcframework (`apple/scripts/build-xcframework.sh`) with Xcode 26.2 against an SDK 57 app — the xcframework builds successfully with the change and `expo run:ios` then completes and launches on the simulator.
- [x] Change is within swift-format limits (line length 93 < 120).


tsapeta (approved):
Hi @kraenhansen!
I suspect you're using an older version of Xcode – it compiles fine in Xcode >=26.4 (which SDK 57 recommends anyway). See a duplicate PR: #47857

Anyway, I'm going to merge this and cherry pick to SDK 57 branch to get rid of the ambiguity.

kraenhansen:
FWIW I was using

  | Version 
-- | --
Xcode | 26.2 (build 17C52)
Swift | Apple Swift 6.2.3 (swiftlang-6.2.3.3.21, clang-1700.6.3.2)
Target | arm64-apple-macosx26.0

PR #49048, opened 2026-08-17
[expo-splash-screen] Fix splash staying visible forever when activity is stopped before the exit hand-off (Android)

celadari:
# Why

Since #44584, `SplashScreenManager` mitigates the API 31–33 `SurfaceControl.checkNotReleased()` crash ([issuetracker 242118185](https://issuetracker.google.com/issues/242118185)) by **clearing the exit-animation listener the first time the activity is stopped**.

This trades the crash for a worse failure: with the listener cleared, an exit hand-off that arrives after a stop has **no receiver** — nothing ever calls `remove()` on the splash view, and the system's splash surface stays on screen **permanently**, above a fully working app (touches pass through; users operate the app blind behind the splash).

This is trivially hit in practice with `preventAutoHideAsync()`: any user who backgrounds the app while the splash is still up (slow network on boot, switching to an email app for an OTP during login, dev-client bundle loads) comes back to a splash that never leaves. We verified the mechanism on a Pixel 4a (API 33): after reproducing, `dumpsys SurfaceFlinger` shows the orphaned system surface

```
Layer ... (Splash Screen <package>)/... - animation-leash of starting_reveal (EffectLayer) callingPid:1731 callingUid:1000
```

while `dumpsys window windows` shows only `MainActivity` — the app rendered long ago; the leaked system splash surface just sits above it.

# How

Replace the clear-listener mitigation with a **stop-aware exit listener**:

- Track the activity's stopped state via `ActivityLifecycleCallbacks` (flag flips on started/stopped; callbacks unregister on destroy).
- In `setOnExitAnimationListener`, when the hand-off arrives while the activity is stopped, **remove the splash view immediately without animating**; otherwise run the existing fade unchanged.

This keeps the original crash fixed — the crash requires the *animation* to run after stop, and no animation runs in that state — while guaranteeing the splash view is always removed, so the surface can never leak. The custom fade behavior for the normal (foregrounded) path is unchanged, and the fix no longer needs the API 31–33 version gate: skipping an animation while stopped is correct on every API level.

# Test Plan

Reproduction (before → after), Pixel 4a / API 33, app using `preventAutoHideAsync()`:

1. Cold start the app; while the splash is still up, press Home.
2. Wait a few seconds, return to the app.
3. **Before:** once the app hides the splash (`hideAsync()`), the call resolves but the splash stays on screen forever; `dumpsys SurfaceFlinger | grep "Splash Screen"` shows the leaked `animation-leash of starting_reveal` surface.
4. **After (this patch, verified via patch-package on 57.0.6/57.0.7):** the splash is removed on hide; `dumpsys SurfaceFlinger` shows no splash surface; repeated background/foreground cycles during boot behave correctly.
5. Normal path regression check: launching without backgrounding still shows the fade exit animation.

# Checklist

- [x] Documentation is up to date to reflect these changes (no docs affected — bug fix with no API change).


zoontek:
@celadari Looks good, but could you run `pnpm format` in `packages/expo-splash-screen` and `./gradlew :expo-splash-screen:spotlessApply` in `packages/expo-splash-screen/android` to fix formatting?

celadari:
Thanks both — and thanks for the platform-source analysis, it's better than what I had.

The correction is accepted: my description's claim that skipping the animation avoids the crash is wrong. The NPE fires in `syncTransferSplashscreenViewTransaction` during the framework's own draw of the transferred view, before the exit listener runs, so nothing inside the listener can prevent it. What #44584 really does is make the client report `isHandleSplashScreenExit=false` on the next resume so hand-offs stop being initiated at all. I'm leaving the original description as-is for history and stating the corrected framing here.

With that correction, the fair comparison as I see it: the first hand-off after a stop is equally exposed in both versions (the clear only propagates on the next resume, so the straggler hand-off happens on main too — it's the one that produces the orphaned view). Where this PR differs is later cycles: main stops hand-offs (no crash, no fade, and the already-orphaned view stays forever), this PR keeps them (fade works, splash always removed, crash race stays possible). In the `preventAutoHideAsync` flow the stuck splash is close to deterministic once the user backgrounds during boot, while the crash is a narrow race — that's why I think the trade is worth it. But I agree it's a judgment call and @zoontek should make it.

On the decor-sweep alternative: I like that it keeps hand-offs disabled, but in the sequence I reproduced the orphan arrives after the return to the task, so a one-shot sweep in `onActivityStarted` would likely run before there is anything to remove. A variant that should close that hole: keep the clear-on-stop mitigation exactly as it is today, and after the first stop watch for the orphan with an `OnPreDrawListener` on the decor (the same mechanism this file already uses to hold the splash) — the framework adding the view schedules a draw pass, so the watcher runs right after the view lands, removes it, and unregisters. That keeps crash exposure identical to main and still guarantees removal. It costs the custom fade after a stop, which is already the case on main today.

I have an API 33 device and a reliable reproduction (background during boot, verify via `dumpsys SurfaceFlinger`), so I'll implement that variant, test both against the repro, and report results here. Happy to push whichever direction is preferred onto this PR.

Meanwhile, pushed 9244400: restores the two final newlines (my upload script ate them, sorry) and resets `mainActivityStopped` in `registerOnActivity` per the review. @alanjhughes I don't have the repo's toolchain set up here to run `pnpm format` / `spotlessApply` — the final newlines were the one concrete violation found; if CI flags anything else I'll fix it.

celadari:
@zoontek Not yet => I while make sure it actually works before submitting completly

celadari:
Test results, as promised.

**Setup:** Pixel 4a, Android 13 (API 33), dev client with `preventAutoHideAsync()`. Reproduction per cycle: cold start, press Home ~4s into boot (splash still up), return ~4s later, wait for boot to complete plus 10s, then verdict by screenshot and `dumpsys SurfaceFlinger` (counting visible splash `BufferStateLayer`s owned by the app, crash markers from logcat).

**This PR's code** (as of 9244400): **7/7 background-during-boot cycles pass** — splash always removed, app visible, 0 crash markers. Clean launch: fade path works normally.

**The watch-and-remove hybrid** (keep the clear-on-stop mitigation, remove an orphaned `SplashScreenView` from the decor via a pre-draw watcher): **9/10 cycles pass, 1/10 failed** — splash still fully visible ≥5s after the app reported ready. I didn't manage to root-cause the miss before the session ended. One candidate explanation: in the post-clear world the system re-shows a starting window on task return and is supposed to dismiss it itself; if that dismissal races, there is no app-side receiver left to recover — a second stranding mode that no decor-sweep can fix because the view never transfers. I can dig further if this direction is preferred, but as implemented the watcher is not robust.

**Honest limitation:** the crash side of the trade is not exercised by this data — 0 `checkNotReleased` occurrences across all 17 hand-off cycles on either variant, so nothing here validates or refutes crash frequency. What the data does show is that the failure mode this PR targets is reliably fixed by keeping the listener registered, and that the main-shaped alternative was flaky in the same scenario.

**Side observation** (pre-existing, applies to main's normal fade path too): after a successful fade, the system's `starting_reveal` animation-leash (an invisible EffectLayer) lingers in SurfaceFlinger until the next window transition. Cosmetic as far as I can tell.

Happy to run more cycles, other API levels, or an instrumented build of either direction if useful.

celadari:
Ran the same experiment on a newer Android version. The Pixel 4a doesn't ship anything above Android 13, so this ran on an emulator: Pixel 9 AVD, Android 16 (API 36), same protocol and same build of this PR's code.

Results: 3/3 background-during-boot cycles and a clean launch all pass — splash always removed, fade path normal, no crashes, no lingering splash layers in SurfaceFlinger. This also covers the review's note about the lifecycle callbacks now registering on every API level: the ungated skip-fade-when-stopped branch shows no observable behavior change on API 36.

Conclusion: on newer Android everything works as expected. The platform race that motivated the original mitigation (https://issuetracker.google.com/issues/242118185, SurfaceControl.checkNotReleased) is most likely fixed there — which is consistent with the mitigation being gated to API 31-33 in the first place. So both failure modes of this trade-off — the crash and the stuck splash — appear to be an Android 12-13 story only.

Combined coverage for this PR's code: API 33 physical device (7/7) and API 36 emulator (4/4).

celadari:
@zoontek — pinging you since this changes your mitigation from #44584 and the deep review above concluded the trade-off is yours to judge. Short version:

**What happens today (without this PR):**

| Android version | Current behavior |
|---|---|
| 12–13 (API 31–33) | ❌ splash can stay on screen **forever** — backgrounding while the splash is up clears the exit listener; a hand-off arriving after that has no receiver, so nothing ever removes it |
| 14+ (API 34+) | ✅ fine — the mitigation never runs, and the platform race ([242118185](https://issuetracker.google.com/issues/242118185)) is fixed there |

**What this PR changes:**
1. Never clears the listener — tracks the stopped state instead.
2. Hand-off arrives while stopped → remove the view **without animating** (no animation after stop = the crash scenario the mitigation targeted can't come from us).
3. Every splash always has a receiver → stranding becomes structurally impossible.

**The trade for you to judge:**
- Keeping the listener means hand-offs (and the platform race window) stay possible on later background cycles, where main disables them after the first stop.
- My read: the stuck splash is near-deterministic in the `preventAutoHideAsync` flow; the crash is a narrow race. But that's a judgment call, not a measurement.

**Testing done (details in comments above):**
1. This PR's code: **7/7** repro cycles on a physical Pixel 4a (API 33), **4/4** on an API 36 emulator, 0 crashes.
2. The alternative (keep the clear + sweep the orphaned view): **9/10** — one unrecoverable stuck splash; in the post-clear world there's a stranding path no app-side sweep can reach.
3. API 34+ is byte-neutral in behavior; if you'd prefer it byte-neutral in code too, I'm happy to gate the new lifecycle callbacks to 31–33 — the fix doesn't need them elsewhere.

Happy to run more cycles or adjust in whatever direction you prefer.

PR #49049, opened 2026-08-17
[ui][Android] Preserve vector drawable fill types

aonan-brex:
## Why

`@expo/ui` converts Android VectorDrawable XML into Compose `ImageVector` instances. `ImageLoader` currently drops `android:fillType`, so paths authored with the even-odd fill rule render using Compose's default non-zero rule instead.

This affects any Android icon or image path that relies on `android:fillType="evenOdd"`; it is not specific to a particular component or icon.

## How

Parse the VectorDrawable `fillType` attribute and pass the corresponding `PathFillType` to `ImageVector.Builder.addPath`.

The parser accepts both the textual `evenOdd` value used by source XML and the compiled enum value `1` returned by Android's `XmlResourceParser`. Paths without the attribute continue to use `PathFillType.NonZero`.

## Test Plan

- `pnpm et check-packages @expo/ui`
- `cd apps/bare-expo/android && ./gradlew :expo-ui:spotlessCheck`
- `cd apps/bare-expo/android && ./gradlew :expo-ui:testDebugUnitTest --tests expo.modules.ui.graphics.ImageLoaderTest`
- Added regression coverage for textual `evenOdd`, compiled value `1`, and the default non-zero fill rule. All 25 `ImageLoaderTest` tests pass.

## Checklist

### DIFF
--- a/packages/expo-ui/CHANGELOG.md
+++ b/packages/expo-ui/CHANGELOG.md
@@ -20,6 +20,7 @@
 
 ### 🐛 Bug fixes
 
+- [Android] Preserve vector drawable `fillType` values when loading images so even-odd paths render correctly.
 - [universal] Fix `Cannot use shared object that was already released` when a worklet callback prop closes over an unstable value. ([#48819](https://github.com/expo/expo/pull/48819) by [@intergalacticspacehighway](https://github.com/intergalacticspacehighway))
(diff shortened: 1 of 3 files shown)


intergalacticspacehighway (approved):
Thanks!

PR #49051, opened 2026-08-17
[app-metrics][iOS][Android] Record observed network requests as trace spans

tsapeta:
# Why

The spans table from #48861 has no producer yet. Network requests are the first one: they're already observed on both platforms, but only in memory, so nothing survives to be exported.

Closes ENG-26032.

# How

A persistence hook on `NetworkRequestMonitor` converts each completed request into a client span per the OTel HTTP semantic conventions, attributed to the main session. The mapping follows the conventions where they have teeth: `url.full` is redacted at the instrumentation (userinfo credentials and default-sensitive query values never reach disk), and only when something was actually redacted, so ordinary URLs keep their exact encoding; nonstandard methods become `_OTHER` with the verbatim value preserved; intentional cancellations keep their span but stay UNSET with no `error.type`, so routine `AbortController` use doesn't pollute error rates. The capture snapshot gains `errorType` (the stable `domain:code` pair on iOS, the exception class on Android), since localized descriptions can't feed a low-cardinality attribute.

Android's lifecycle needs extra care: installation drains the monitor's ring buffer once per process, with the drained flag flipped only after the batch completes (a JS reload that cancels the batch re-drains instead of losing the buffer), and the module uninstalls its producer on destroy so a reload can't keep writing rows against the torn-down session.

# Test Plan

Producer suites on both platforms cover the semconv mapping (redaction and byte-fidelity, `_OTHER`, cancellations, status rules, `server.port`, redirect events), persistence through the monitor, the backfill's at-least-once semantics across reloads, and FK degradation. `et native-unit-tests` passes on both platforms.

### DIFF
--- a/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/AppMetricsModule.kt
+++ b/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/AppMetricsModule.kt
@@ -13,7 +13,9 @@ import expo.modules.appmetrics.crashreporting.PreferencesLastProcessedExitStore
 import expo.modules.appmetrics.crashreporting.attributeAndStoreCrashReport
 import expo.modules.appmetrics.logevents.LogEventOptions
 import expo.modules.appmetrics.networkrequests.NetworkRequestFilter
+import expo.modules.appmetrics.networkrequests.NetworkRequestMonitor
 import expo.modules.appmetrics.networkrequests.NetworkRequestObserver
+import expo.modules.appmetrics.networkrequests.NetworkRequestPersistence
 import expo.modules.appmetrics.logevents.Severity
 import expo.modules.appmetrics.logevents.sanitizeLogEventAttributes
 import expo.modules.appmetrics.logevents.validateDisplayName
@@ -22,6 +24,7 @@ import expo.modules.appmetrics.logevents.validateEventName
 import expo.modules.appmetrics.logevents.withDisplayNameAttribute
 import expo.modules.appmetrics.memory.MemoryMetricsManager
 import expo.modules.appmetrics.storage.JsDebugSession
+import expo.modules.appmetrics.storage.MetricsDatabase
 import expo.modules.appmetrics.storage.JsLogRecord
 import expo.modules.appmetrics.storage.JsMetric
 import expo.modules.appmetrics.storage.LogRecord
@@ -64,6 +67,12 @@ class AppMetricsModule : Module(), UpdatesStateChangeListener {
   lateinit var sessionManager: SessionManager
 
   lateinit var mainSession: SessionSharedObject
+
+  /**
+   * The span producer installed on the monitor in `OnCreate`, kept so `OnDestroy` can
+   * uninstall it when the module is torn down (a JS reload).
+   */
+  private var networkRequestPersistence: NetworkRequestPersistence? = null
 
   // Lazy-initialized metadata - created once when first needed
   private val metadata: AppMetadata? by lazy {
@@ -152,6 +161,20 @@ class AppMetricsModule : Module(), UpdatesStateChangeListener {
         // a racing write triggers (and joins) the same single start job.
         scope.launch { mainSession.awaitSessionPersisted() }
 
+        // From here on every completed request is written to the `spans` table, attributed to
+        // the main session. Awaiting the session row first keeps the FK satisfied for every
+        // span insert; installation also drains requests buffered since process start.
+        scope.launch {
+          mainSession.awaitSessionPersisted()
+          val persistence = NetworkRequestPersistence(
+            database = MetricsDatabase.getDatabase(context),
+            scope = scope,
+            sessionId = mainSession.sessionId
+          )
+          networkRequestPersistence = persistence
+          NetworkRequestMonitor.shared.installPersistence(persistence)
+        }
+
         // Sweep sessions orphaned by a previous process. The cutoff equals this
         // session's start and the comparison is strict (`<`), so this session
         // survives while older ones are swept — order vs the INSERT doesn't
@@ -215,6 +238,10 @@ class AppMetricsModule : Module(), UpdatesStateChangeListener {
       }
 
       OnDestroy {
+        // Stop persisting spans for this module's session. Without this, a JS reload leaves
+        // the old instance on the process-wide monitor, writing rows attributed to the
+        // torn-down session until the next OnCreate replaces it.
+        networkRequestPersistence?.let { NetworkRequestMonitor.shared.uninstallPersistence(it) }
         // `modulesQueue` is cancelled immediately after this hook returns, so
         // run the UPDATE on the calling thread to make sure the end timestamp
         // is persisted before teardown. `stop` awaits the session-start job
--- a/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/AppMetricsModule.kt
+++ b/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/AppMetricsModule.kt
@@ -13,7 +13,9 @@ import expo.modules.appmetrics.crashreporting.PreferencesLastProcessedExitStore
 import expo.modules.appmetrics.crashreporting.attributeAndStoreCrashReport
 import expo.modules.appmetrics.logevents.LogEventOptions
 import expo.modules.appmetrics.networkrequests.NetworkRequestFilter
+import expo.modules.appmetrics.networkrequests.NetworkRequestMonitor
 import expo.modules.appmetrics.networkrequests.NetworkRequestObserver
+import expo.modules.appmetrics.networkrequests.NetworkRequestPersistence
 import expo.modules.appmetrics.logevents.Severity
 import expo.modules.appmetrics.logevents.sanitizeLogEventAttributes
 import expo.modules.appmetrics.logevents.validateDisplayName
@@ -22,6 +24,7 @@ import expo.modules.appmetrics.logevents.validateEventName
 import expo.modules.appmetrics.logevents.withDisplayNameAttribute
 import expo.modules.appmetrics.memory.MemoryMetricsManager
 import expo.modules.appmetrics.storage.JsDebugSession
+import expo.modules.appmetrics.storage.MetricsDatabase
 import expo.modules.appmetrics.storage.JsLogRecord
 import expo.modules.appmetrics.storage.JsMetric
 import expo.modules.appmetrics.storage.LogRecord
@@ -64,6 +67,12 @@ class AppMetricsModule : Module(), UpdatesStateChangeListener {
   lateinit var sessionManager: SessionManager
 
   lateinit var mainSession: SessionSharedObject
+
+  /**
+   * The span producer installed on the monitor in `OnCreate`, kept so `OnDestroy` can
+   * uninstall it when the module is torn down (a JS reload).
+   */
+  private var networkRequestPersistence: NetworkRequestPersistence? = null
 
   // Lazy-initialized metadata - created once when first needed
   private val metadata: AppMetadata? by lazy {
@@ -150,7 +159,19 @@ class AppMetricsModule : Module(), UpdatesStateChangeListener {
         // Persist the session row eagerly so it's visible to readers
         // (`getMainSession`, …) as soon as possible. Idempotent:
         // a racing write triggers (and joins) the same single start job.
-        scope.launch { mainSession.awaitSessionPersisted() }
+        scope.launch {
+          mainSession.awaitSessionPersisted()
+          // From here on every completed request is written to the `spans` table, attributed to
+          // the main session. The await above keeps the FK satisfied for every span insert;
+          // installation also drains requests buffered since process start.
+          val persistence = NetworkRequestPersistence(
+            database = MetricsDatabase.getDatabase(context),
+            scope = scope,
+            sessionId = mainSession.sessionId
+          )
+          networkRequestPersistence = persistence
+          NetworkRequestMonitor.shared.installPersistence(persistence)
+        }
 
         // Sweep sessions orphaned by a previous process. The cutoff equals this
         // session's start and the comparison is strict (`<`), so this session
@@ -216,6 +237,10 @@ class AppMetricsModule : Module(), UpdatesStateChangeListener {
       }
 
       OnDestroy {
+        // Stop persisting spans for this module's session. Without this, a JS reload leaves
+        // the old instance on the process-wide monitor, writing rows attributed to the
+        // torn-down session until the next OnCreate replaces it.
+        networkRequestPersistence?.let { NetworkRequestMonitor.shared.uninstallPersistence(it) }
         // `modulesQueue` is cancelled immediately after this hook returns, so
         // run the UPDATE on the calling thread to make sure the end timestamp
         // is persisted before teardown. `stop` awaits the session-start job
--- a/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/networkrequests/NetworkRequestInterceptor.kt
+++ b/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/networkrequests/NetworkRequestInterceptor.kt
@@ -159,7 +159,8 @@ class NetworkRequestInterceptor private constructor(
       fallbackStart = startedAt,
       fallbackEnd = endDate,
       totalDuration = totalDuration,
-      error = error
+      error = error,
+      cancelled = call.isCanceled()
     )
     monitor.record(snapshot)
   }
@@ -197,7 +198,8 @@ internal fun buildSnapshot(
   fallbackStart: Date,
   fallbackEnd: Date,
   totalDuration: Double,
-  error: IOException?
+  error: IOException?,
+  cancelled: Boolean = false
 ): NetworkRequest {
   val redirects = response?.let { buildRedirectChain(it) } ?: emptyList()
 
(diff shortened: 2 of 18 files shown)


packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/networkrequests/NetworkRequestPersistence.kt
```diff
@@ -0,0 +1,254 @@
...
+    return url.substringBefore('?')
+  }
+  val builder = parsed.newBuilder()
+  if (parsed.encodedUsername.isNotEmpty() || parsed.encodedPassword.isNotEmpty()) {
+    builder.username("REDACTED")
+    builder.password("REDACTED")
+  }
+  for (name in parsed.queryParameterNames) {
+    if (name.lowercase() in SENSITIVE_QUERY_PARAMETERS) {
+      builder.removeAllQueryParameters(name)
```
tsapeta: [🏗️ **design**] `removeAllQueryParameters(name)` + `addQueryParameter(name, "REDACTED")` moves each redacted parameter to the **end** of the query, so a redacted URL's parameter order no longer matches what was sent. Verified against okhttp 4.12.0:

```
IN : https://api.example.com/s?sig=secret&q=hello
OUT: https://api.example.com/s?q=hello&sig=REDACTED

IN : https://api.example.com/s?a=1&sig=s&b=2&signature=t&c=3
OUT: https://api.example.com/s?a=1&b=2&c=3&sig=REDACTED&signature=REDACTED
```

The existing tests don't catch it because both put the sensitive parameters last already, where reordering is a no-op.

Two consequences. `url.full` is the column URL-level analysis groups on, so the same endpoint hit with the same parameters lands under different strings depending on parameter order, and a presigned URL's ordering is decided by whatever SDK built it. It also diverges from iOS, which preserves position (`?sig=REDACTED&q=hello`), so the two platforms report different `url.full` for the identical request — which matters given how deliberately the rest of this mapping is kept in lockstep.

`setQueryParameter(name, "REDACTED")` replaces in place and keeps the position. It also collapses repeated occurrences of the same name, which is what you want here.

Separately, the loop redacts a valueless parameter: `?sig` becomes `?sig=REDACTED`. iOS skips it (`item.value != nil`). Worth matching whichever way you prefer, though the valueless case carries no secret either way.

packages/expo-app-metrics/ios/NetworkRequests/NetworkRequestPersistence.swift
```diff
@@ -0,0 +1,257 @@
...
+    if let queryItems = components.queryItems, !queryItems.isEmpty {
+      var itemsChanged = false
+      let redactedItems = queryItems.map { item -> URLQueryItem in
+        guard sensitiveQueryParameters.contains(item.name.lowercased()), item.value != nil else {
+          return item
+        }
+        itemsChanged = true
+        return URLQueryItem(name: item.name, value: "REDACTED")
+      }
+      if itemsChanged {
```
tsapeta: [🏗️ **design**] The comment states the reason for gating on `redactedAnything` precisely: reassigning `queryItems` re-encodes the whole query and `a%2Bb` comes back as a literal `+`. But the gate only protects URLs where **nothing** was redacted. As soon as one sensitive parameter is present, the assignment runs and every *other* parameter on that URL goes through the lossy round trip. Verified with Foundation:

```
IN : https://api.example.com/s?q=a%2Bb&sig=secret
OUT: https://api.example.com/s?q=a+b&sig=REDACTED
```

So `q`'s value silently changes meaning from `a+b` to `a b` for exactly the URLs that carry signed parameters. `%20` survives, so it's specific to `%2B`.

The test that pins this behavior (`leaves an unredacted URL byte-identical`) covers only the no-redaction path, which is why the redaction path's re-encoding isn't caught.

If you want byte fidelity on both paths, editing `percentEncodedQueryItems` instead of `queryItems` avoids the decode/re-encode entirely — you'd match against the percent-encoded name (the sensitive names here contain no characters that get encoded, so a `lowercased()` comparison still works) and substitute the encoded value. Alternatively a string-level splice over `percentEncodedQuery`.

A test with a `%2B` parameter alongside a `sig` would fail today and pin whichever fix you pick.

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/networkrequests/NetworkRequestInterceptor.kt
```diff
@@ -257,9 +273,12 @@ internal fun buildSnapshot(
     requestBytesSent = requestBytesSent,
     responseBytesReceived = responseBytesReceived,
     timings = timings,
+    fetchType = fetchType,
     // Falls back to the class name because an exception is allowed to carry no message at all, and
     // a null description would read as "this request succeeded" to `isFailed`.
     errorDescription = failure?.let { it.localizedMessage ?: it.message ?: it.javaClass.simpleName },
+    errorType = failure?.javaClass?.name,
+    canceled = failure != null && canceled,
```
tsapeta: [💡 suggestion] `canceled = failure != null && canceled` means a cancellation is only recognized when OkHttp also surfaced an `IOException`. That's the normal path, and gating on `failure` correctly stops a cancel flag from suppressing the ERROR status on a request that had already completed successfully.

Worth confirming the other direction, though: if a call is canceled after the response headers arrive but before the body finishes, does the interceptor reach here with `failure != null`, or does it record a snapshot with a 2xx `statusCode` and no exception? In the latter case the span looks like a clean success, which is benign for error rates but means a truncated response reads as complete. Not blocking — mostly asking whether that path was considered, since the iOS side keys off `NSURLErrorDomain:-999` and can't have this shape.

packages/expo-app-metrics/ios/NetworkRequests/NetworkRequest.swift
```diff
@@ -35,13 +35,32 @@ public struct NetworkRequest: Sendable, Equatable, Identifiable {
...
+    /// Served from the local URL cache without a network load.
+    case cache
+    /// Delivered by an HTTP/2 server push.
+    case serverPush = "server_push"
+  }
+
+  /// Whether the response came off the network rather than out of a cache or a server push.
+  /// `nil` when the OS did not classify the fetch, which callers must treat as unusable rather
+  /// than assume either way.
+  public var cameFromNetwork: Bool? {
```
tsapeta: [💡 suggestion] `cameFromNetwork` changes from a stored `public let` to a computed `public var`, and the memberwise initializer's parameter list changes with `fetchType` replacing it. For a normal package that's just a source-compatible-ish refactor, but `expo-app-metrics` ships Swift in a pod that consumers can build against a prebuilt artifact, and stored-to-computed on a public struct changes the property's accessor mangling and the struct's layout.

Worth a moment's thought about whether anything outside this package reads `NetworkRequest` — I grepped and found no consumer of `cameFromNetwork` in the monorepo, and `NetworkRequest` is only produced internally, so this looks safe in practice. Flagging it only so the call is deliberate rather than incidental; if `expo-app-metrics` is in the prebuilt set, the `breaking: ABI` label would be the conservative move.

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/networkrequests/NetworkRequestPersistence.kt
```diff
@@ -0,0 +1,270 @@
...
+ * the main session. This is the first span producer: it converts each request into a generic
+ * `Span` following the OTel HTTP semantic conventions, so the export layer (`expo-observe`)
+ * ships spans without knowing what produced them. Unlike the in-memory ring buffer, the rows
+ * survive process death until they're dispatched, pruned with their session, or displaced past
+ * the table's row cap. Mirrors the iOS `NetworkRequestPersistence`.
+ *
+ * Installed on the monitor once the module created the main session (see `AppMetricsModule`);
+ * the monitor calls `persist` directly for every completion it records. Deliberately not a
+ * `NetworkRequestObserverDelegate`: persistence is part of the pipeline, not an observer, and
+ * the delegate contract (weak registration, `shouldObserveRequest` filtering) doesn't apply.
```
Ubax: Can we make this comment shorter and easier to understand? 😅 

Also can we move the implementation details comments closer to the code?

So this is only general description and then more detailed comments inlined to functions and implementation?

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/networkrequests/NetworkRequestPersistence.kt
```diff
@@ -0,0 +1,270 @@
...
+ */
+class NetworkRequestPersistence(
+  private val database: MetricsDatabase,
+  private val scope: CoroutineScope,
+  /**
+   * A plain value, not a provider: the id is constant for a persistence instance (it is
+   * constructed after the main session exists), and resolving it eagerly means the monitor's
+   * record path (OkHttp dispatcher threads) never calls back into module state that a
+   * teardown could have invalidated.
+   */
```
Ubax: Is this comment needed?

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/networkrequests/NetworkRequestPersistence.kt
```diff
@@ -0,0 +1,270 @@
...
+  private val sessionId: String
+) {
+  /**
+   * Persists one completed request as a span. The insert runs on `scope` so the monitor's
+   * record path (OkHttp dispatcher threads) never blocks on the database; failures are logged
+   * and swallowed — persistence must never break the monitor's fan-out.
+   */
```
Ubax, line 32:
```suggestion
  /**
   * Persists one completed request as a span. 
   */
```

Then we can have inline `// Runs on `scope`, so OkHttp dispatcher threads does not block on databse`

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/networkrequests/NetworkRequestPersistence.kt
```diff
@@ -0,0 +1,270 @@
...
+      }
+    }
+  }
+
+  /**
+   * Persists the monitor's buffered startup requests. One coroutine for the whole batch: the
+   * install path runs on the module's serial queue during startup, competing with the session
+   * INSERT and crash-report processing, so converting up to 200 requests inline (JSON
+   * serialization, URL parsing) and queueing one coroutine per row there would be the most
+   * expensive way to do it. Conversion and inserts all happen inside the single launch.
```
Ubax: Same here

Ubax (commented):
My feedback is mostly about the comments. They seem to be too long, overly complicated and some are redundant with code.

Can you split them into:
- General comments for classes and functions
- Inline comments explaining complicated/not obvious behavior

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/networkrequests/NetworkRequestPersistence.kt
```diff
@@ -0,0 +1,270 @@
...
+  fun persistBuffered(requests: List<NetworkRequest>, onComplete: () -> Unit = {}) {
+    if (requests.isEmpty()) {
+      return
+    }
+    scope.launch {
+      for (request in requests) {
+        val span = request.toSpan(sessionId) ?: continue
+        try {
+          database.spanDao().insert(span)
+        } catch (e: Exception) {
```
Ubax, line 72: [🏗️ **design**] This catch defeats the at-least-once guarantee described in `installPersistence`. `kotlinx.coroutines.CancellationException` is a `java.util.concurrent.CancellationException`, which is an `Exception`. When `modulesQueue` is cancelled mid-batch, every remaining `insert` (a `withContext` hop inside Room) throws `CancellationException`, this catch swallows it, the loop runs to the end, and `onComplete()` still fires and sets `hasDrainedBackfill = true`. The buffered requests are then lost for good, exactly the case the comment on `onComplete` says must not report completion.

The existing test only cancels the scope before the coroutine starts, so it never exercises this path. Add a test and fix accordingly.

packages/expo-app-metrics/ios/NetworkRequests/NetworkRequestPersistence.swift
```diff
@@ -0,0 +1,261 @@
...
+    // Not a semconv attribute: a cached response is timestamped and byte-counted like a download,
+    // so without this a large cache hit reads as an impossible transfer rate. Omitted when the
+    // platform did not classify the fetch, so its absence never implies a network load.
+    if let fetchType = request.fetchType {
+      attributes["expo.http.resource.fetch_type"] = fetchType.rawValue
+    }
+    let httpErrorStatus = (request.statusCode ?? 0) >= 400
+    // An intentional cancellation is not a failure: status stays UNSET and `error.type` is
+    // never set, per the conventions.
+    let canceled = request.errorType == canceledErrorType
```
Ubax: [💡 suggestion] Android guards this case (`canceled && (failure != null || (response?.code ?: 0) < 400)`) so a late cancel cannot mask a 4xx/5xx the server already returned. On iOS a task cancelled while reading a 500 body still has `task.response` set, so `statusCode` is 500 and `errorType` is `NSURLErrorDomain:-999`, and this line marks it canceled: status UNSET, no `error.type`. The same request on Android is an ERROR span. Consider `let canceled = request.errorType == canceledErrorType && !httpErrorStatus` to match.

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/AppMetricsModule.kt
```diff
@@ -216,6 +239,10 @@ class AppMetricsModule : Module(), UpdatesStateChangeListener {
       }
 
       OnDestroy {
+        // Stop persisting spans for this module's session. Without this, a JS reload leaves
+        // the old instance on the process-wide monitor, writing rows attributed to the
+        // torn-down session until the next OnCreate replaces it.
+        networkRequestPersistence?.let { NetworkRequestMonitor.shared.uninstallPersistence(it) }
```
Ubax, line 243: [💡 suggestion] Between this uninstall and the next module's install (which waits for its session row), `persistence` is `null` and `hasDrainedBackfill` is already `true`, so every request completing in that window is buffered but never persisted. For a dev reload that is fine; for a production host recreation (for example an `expo-updates` reload) it is a silent gap.

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/networkrequests/NetworkRequestPersistence.kt
```diff
@@ -0,0 +1,270 @@
...
+   */
+  private val sessionId: String
+) {
+  /**
+   * Persists one completed request as a span. The insert runs on `scope` so the monitor's
+   * record path (OkHttp dispatcher threads) never blocks on the database; failures are logged
+   * and swallowed — persistence must never break the monitor's fan-out.
+   */
+  fun persist(request: NetworkRequest) {
+    val span = request.toSpan(sessionId) ?: return
```
Ubax: `toSpan` (URL parsing, JSON building) runs here on the OkHttp dispatcher thread for every completed request, while `persistBuffered` deliberately moves the same conversion inside the coroutine for cost reasons. Moving `toSpan` into the `launch` keeps the record path to a single dispatch and makes the two methods consistent.

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/networkrequests/NetworkRequestPersistence.kt
```diff
@@ -0,0 +1,270 @@
...
+          return@joinToString pair
+        }
+        val name = pair.substring(0, separator)
+        if (name.lowercase() in SENSITIVE_QUERY_PARAMETERS) "$name=REDACTED" else pair
+      }
+    if (redacted != encodedQuery) {
+      builder.encodedQuery(redacted)
+    }
+  }
+  return builder.build().toString()
```
Ubax: [🐧 nit] iOS returns `url.absoluteString` unchanged unless something was redacted; here the URL is always rebuilt through `HttpUrl`, which canonicalizes. In production the input is already `HttpUrl.toString()` output so this is idempotent, but the byte-fidelity promise in the PR description is only tested on iOS. An early `return url` when nothing changed would make the two platforms match and remove the reliance on the input already being canonical.

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/AppMetricsModule.kt
```diff
@@ -152,6 +161,20 @@ class AppMetricsModule : Module(), UpdatesStateChangeListener {
         // a racing write triggers (and joins) the same single start job.
         scope.launch { mainSession.awaitSessionPersisted() }
 
+        // From here on every completed request is written to the `spans` table, attributed to
+        // the main session. Awaiting the session row first keeps the FK satisfied for every
+        // span insert; installation also drains requests buffered since process start.
+        scope.launch {
+          mainSession.awaitSessionPersisted()
+          val persistence = NetworkRequestPersistence(
```
Ubax, line 167: [🐧 nit] The `scope.launch { mainSession.awaitSessionPersisted() }` right above does the same await. The two launches could be folded into one, with the install as the continuation, so the file has a single place that waits for the session row.

packages/expo-app-metrics/ios/NetworkRequests/NetworkRequest.swift
```diff
@@ -167,19 +203,34 @@ extension NetworkRequest {
...
+      case .serverPush:
+        return .serverPush
+      default:
+        return nil
+      }
+    }
 
     let timings = NetworkRequest.Timings(
-      fetchStart: transaction?.fetchStartDate ?? fallbackStart,
+      fetchStart: firstTransaction?.fetchStartDate ?? fallbackStart,
```
Ubax: [🐧 nit] This also changes the existing JS `startedAt` and the TTI window test (`summarize` filters on `timings.fetchStart`) for redirected requests, which now start at the first hop rather than the final one. Probably the more correct value, but it is a behavior change outside the span producer and is not mentioned in the changelog entry.

Ubax (approved):
AI review

PR #49056, opened 2026-08-18
[log-box] Show LogBox for fatal errors reported off the main thread

hassankhan:
# Why

Fatal JavaScript errors during app startup on iOS do not show any UI at all, instead just hanging on the splash screen at "Downloading 100%..." or "Loading from Metro...". 

| Before | After |
|---|---|
| image | image |

# How

This was due to two issues:

1. In the `RCTRedBox (WithExpoLogBox)` swizzle, we present the view controller on the calling thread, whereas React Native's `RCTRedBox`  dispatches to the main queue. This resulted in UIKit throwing an exception, and no sheet appearing onscreen. This was fixed by wrapping the swizzled body in `dispatch_async(dispatch_get_main_queue(), ...)`.

2. After fixing the above, a sheet would render but without any content. LogBox's JavaScript bundle would throw `TypeError: registerDOMComponent is not a function` on startup. This happens because Metro skips platform-extension resolution for `package.json`-declared `exports`, the result of which was that web bundles would get the native `internal.ts` where `registerDOMComponent` is `undefined`. This was fixed by changing `dom.ts` and `internal.ts` to be barrel files re-exporting from the new `dom/` and `internal/` subdirectories that hold the platform and fallback logic, a similar approach as we took in #47870.

# Test Plan

Reproduced on the iOS simulator using `apps/router-e2e` with the `server-loader` fixture and `experiments.noxcturnalTransformWorker: true`.

# Checklist


packages/@expo/log-box/dist/ExpoLogBox.bundle/index.html
krystofwoldrich: Should these be removed, the files don't exists in the bundle?

krystofwoldrich:
(Note for myself, we need iOS and Android E2E tests for the new LogBox UI.)

krystofwoldrich (approved):
Thank you! That's a good catch.

Kudo (approved):

packages/@expo/log-box/dist/ExpoLogBox.bundle/index.html
hassankhan: Good point, but oddly those files still do exist in the bundle on my machine at least. Will investigate further.

hassankhan: Should be resolved in 4b0a7f368be569b7dbc6ffe75022cb6382f342ca

packages/@expo/log-box/metro.config.js
```diff
@@ -6,5 +6,10 @@ const config = getDefaultConfig(__dirname);
 
 config.watchFolders = [];
 config.transformerPath = require.resolve('./shadowDomCssTransformer');
+// NOTE(@hassankhan): Prevents modules from being bundled twice through `src/` and `build/`
+config.resolver.unstable_conditionNames = [
+  ...config.resolver.unstable_conditionNames,
+  'expo-source',
```
kitten, line 12: I assume this means we somewhere already drop into `src` without noticing it, since it implies that we consistently want to go to `src`, and duplicates shouldn't otherwise happen?

We may want to investigate that. Apart from that, happy for this to be merged and this problem to be investigated later, as in, leave this as a `TODO`

cc @krystofwoldrich

krystofwoldrich: ~Might the best to merge it without the fix, not sure if it doesn't break some functions of the LogBox. And as mentioned let's look into in later.~

I'll note it as todo before stabilizing the new LogBox.

PR #49063, opened 2026-08-18
[router] Remove initialState from NavigationContainer

Ubax:
# Why

# How

# Test Plan

# Checklist


packages/expo-router/src/react-navigation/core/BaseNavigationContainer.tsx
```diff
@@ -40,35 +41,21 @@ type State = NavigationState | PartialState<NavigationState> | undefined;
...
- *
- * @param state Initial state object.
- */
-const getPartialState = (
-  state: InitialState | undefined
-): PartialState<NavigationState> | undefined => {
+function validateInitialState(state: InitialState | undefined): void {
   if (state === undefined) {
     return;
   }
```
jakub-agent, line 47: [🏗️ **design**] This converts a previously-normalizing code path into an unconditional throw, including in production.

`getPartialState` used to *accept* complete states and normalize them (strip `key`/`routeNames`, set `stale: true`). Now anything carrying `key`, `routeNames`, or `stale: false` throws at render.

`BaseNavigationContainer` is still exported from the public `./react-navigation` subpath (`core/index.tsx`, marked `@deprecated`), so this is reachable. The pattern it breaks is the canonical react-navigation state-persistence recipe — save `getRootState()`, pass it back as `initialState` — which produces a state with all three fields. After an SDK upgrade that becomes a hard crash on launch rather than a degraded restore.

Precedent in this same file is mixed: the nested-container check throws unconditionally, but `checkSerializable` and `checkDuplicateRouteNames` only warn in development. Given this is a state *shape* problem that the old code could recover from, would a dev-only throw plus production fallback (ignore the internal fields and treat the state as stale) be safer? If the intent is that this surface is closed enough that a hard failure is correct, that's a defensible call — but it needs a changelog entry and a migration note, since the failure mode is a production crash.

The error message itself is good and actionable.

packages/expo-router/src/fork/useLinking.native.ts
```diff
@@ -18,19 +18,20 @@ type Options = LinkingOptions<ParamListBase>;
 
 const linkingHandlers: symbol[] = [];
 
+function getInitialPath(prefixes: string[], url: string) {
```
jakub-agent, line 21: [🏗️ **design**] `extractExpoPathFromURL` unconditionally strips the leading `/` (`extractPathFromURL.ts`, the `.replace(/^\//, '')` at the end). This helper re-adds it — but only for the initial URL, which leaves the same normalization open-coded in four places with different results:

1. here, via `getInitialPath` — normalized
2. `fork/useLinking.ts` — the inline `if (path !== undefined && !path.startsWith('/'))`
3. `global-state/useStore.ts` — `if (!initialPath.startsWith('/')) initialPath = '/' + initialPath;`
4. the `subscribe` listener below — normalizes only the routing `href`

That last one is where it bites. Within a single listener invocation:

```js
const path = extractExpoPathFromURL(prefixes, url);        // "foo"
const state = ... getStateFromURL(url);                     // getStateFromPath("foo")
onUnhandledLinking(path);                                   // "foo"
routingQueue.add({ href: path.startsWith('/') ? path : `/${path}` });  // "/foo"
```

So `getStateFromPath` now receives `/foo` for the initial URL but `foo` for every subsequent deep link. And `onUnhandledLinking` gets the un-normalized form, while `NavigationContainer` clears `lastUnhandledLink` by strict equality against `getCurrentRoute()?.path` — which carries a leading slash. That comparison can never match for a subsequent deep link, so the unhandled link is never cleared.

The mismatch on the subsequent path predates this PR, but this change creates the divergence between initial and subsequent and adds a third copy of the fix. Normalizing once — inside `extractExpoPathFromURL`, or in a single shared wrapper used by all four call sites — would let the `initial` parameter disappear entirely, which is what your TODO suspects.

packages/expo-router/src/fork/useLinking.native.ts
```diff
@@ -109,12 +113,15 @@ export function useLinking(
   });
 
   const getStateFromURL = useCallback(
-    (url: string | null | undefined) => {
+    // TODO: This may no longer be necessary. It was added during the initial-state refactor to preserve the previous path shape.
+    (url: string | null | undefined, initial = false) => {
```
jakub-agent, line 118: [💡 suggestion] Worth resolving before merge rather than shipping — it's the same question as the normalization comment above, and the `initial` parameter only exists to satisfy it.

For what it's worth, the reason the flag *appears* load-bearing is that `useStore.ts` open-codes the same leading-slash fix for its prefetch, so the container's derivation has to match it or the two would disagree on the initial state. Unifying the normalization removes that coupling and answers the TODO.

Minor placement point: the comment sits directly above the arrow function, so it reads as documenting `getStateFromURL` as a whole rather than the `initial` parameter it's actually about.

packages/expo-router/src/fork/NavigationContainer.tsx
```diff
@@ -38,7 +38,7 @@ declare global {
 
 globalThis.REACT_NAVIGATION_DEVTOOLS = new WeakMap();
 
-type Props<ParamList extends object> = NavigationContainerProps & {
+type Props<ParamList extends object> = Omit<NavigationContainerProps, 'initialState'> & {
```
jakub-agent, line 41: [🏗️ **design**] This is a breaking change to `expo-router/react-navigation` and needs a CHANGELOG entry — `packages/expo-router/CHANGELOG.md` already tracks this refactor series under *Unpublished → 🛠 Breaking changes*.

While you're there: the existing entry beginning *"Seed the navigation container verbatim from the compiled store state: the container no longer stales the seed…"* looks superseded by this PR, since the container no longer accepts a seed at all and derives state from `linking` instead. It probably needs rewording rather than a second entry stacked on top.

One mechanical note: `Omit` only removes the prop at the type level. `{...rest}` still forwards a runtime `initialState` straight through to `BaseNavigationContainer`, so a JS caller (or an untyped spread) reaches the new throw rather than being ignored. That's arguably fine, but it means the type-level removal and the runtime behavior tell different stories.

packages/expo-router/src/react-navigation/core/__tests__/typelessRouter.test.ios.tsx
```diff
@@ -77,36 +52,15 @@ const persistedState = {
...
-      <TestNavigator createRouter={StateWithoutTypeRouter}>
-        <Screen name="first" component={RouterType} />
-        <Screen name="second" component={RouterType} />
-      </TestNavigator>
-    </BaseNavigationContainer>
-  );
-
-  expect(ref.current!.getRootState().index).toBe(1);
-  expect(screen.getAllByText('test')).toHaveLength(2);
+test('rejects a persisted state before a typeless router can rehydrate it', () => {
```
jakub-agent, line 55: [💡 suggestion] This drops three behavioral tests (type-less persisted state accepted; typed state discarded; `NavigatorTypeContext` provided) and replaces them with an assertion that `BaseNavigationContainer.test.ios.tsx` now makes twice in this same PR.

The logic those tests covered is still live: `isStateValid` in `useNavigationBuilder.tsx` still checks `state.type === undefined || state.type === router.type`, and `initializedState` still branches on it for the `<Activity>` recovery path. Since a complete state can no longer be fed in through `initialState`, that branch has lost its only exercise.

Either reach it another way (drive it through the `<Activity>` recovery path, or unit-test the router directly), or delete this file — as it stands it's a near-duplicate of the container test, and `StateWithoutTypeRouter` survives only because `TypelessRouter` strips its `type`, not because anything asserts on the distinction.

packages/expo-router/src/fork/useLinking.ts
```diff
@@ -68,39 +71,47 @@ export function useLinking(
...
-        : undefined;
-
-      if (path) {
-        value = getStateFromPathRef.current(path, configRef.current);
+      const getStateFromURL = (url: string | null | undefined) => {
+        let path = url ? extractExpoPathFromURL(prefixesRef.current, url) : undefined;
+        if (path !== undefined && !path.startsWith('/')) {
+          path = `/${path}`;
+        }
+
```
jakub-agent, line 95: [🐧 nit] This `const state` shadows the outer `let state` declared at the top of `getInitialState`. It works — the inner one is returned and the outer is assigned from the return value — but the two are easy to conflate when reading the sync and async branches side by side. Renaming one (`nextState`, or returning directly) would make the control flow easier to follow.

jakub-agent (commented):

PR 8/8 in the router refactor stack (`#48888 → #48886 → #48890 → #48889 → #48895 → #48917 → #49034 → #49063`), reviewed against its base `@ubax/ensure-optional-properties-in-state`.

The core move — replacing `ServerContext` with the existing `linking.getInitialURL` seam so the container derives its own initial state — is sound. I checked the two invariants that carry it, because both are non-obvious and neither is stated in the diff:

- **Static rendering still works.** `isLinkingReady` no longer short-circuits on `initialState`, so it rests entirely on `isResolved`. That holds because `useThenable` invokes `promise.then` synchronously and mutates `initialState` *before* `useState`, and `getLinkingConfig` returns `serverUrl ?? getInitialURL()` synchronously (memoized, so `useStore`'s prefetch and the container's derivation see the same URL). Single-pass renders resolve on the first pass.
- **Dropping `stale: true` is safe.** `isStateInitialized` tests `stale === false`, not truthiness, so an absent `stale` is correctly treated as needing rehydration. The old `stale: true` was redundant, and `validateInitialState` rejecting `stale: false` is coherent with that check.

What I'd want resolved before merge:

1. `BaseNavigationContainer` now **throws in production** for states that previously worked. It is still exported (deprecated) from the public `./react-navigation` subpath, so the canonical react-navigation persistence pattern (`getRootState()` → store → `initialState`) becomes a hard render crash on upgrade.
2. Leading-slash normalization now exists in **four** places and is applied inconsistently. The new `initial` flag encodes that divergence rather than resolving it — and the code ships a TODO saying as much.
3. **No CHANGELOG entry**, though `packages/expo-router/CHANGELOG.md` already tracks this series under *Unpublished → 🛠 Breaking changes*. Removing `initialState` from the container props is a breaking change to `expo-router/react-navigation`. The existing entry *"Seed the navigation container verbatim from the compiled store state…"* also looks superseded — the container no longer takes a seed at all.

Not approving: the PR is a draft, the description is the unfilled template (no Why / How / Test Plan), and it carries an unresolved TODO that the author flagged themselves. Nothing here looks like a confirmed runtime bug — the findings are design-level.

PR #49065, opened 2026-08-18
[ci][precompile] Rebuild external XCFrameworks when React Native changes

chrfalch:
# Why

Expo's EAS build service hosts prebuilt XCFramework versions of some of the main npm packages that are most often used in Expo projects to reduce build times (specifically reduces build times for C++ heavy packages).

A workflow exists to rebuild these when we change our versions of these packages in our repo (`.github/workflows/ios-prebuild-external-xcframeworks.yml`).

The problem with this is that our XCFrameworks are rebuilt only when the version of the package itself has changes, but not when React Native itself has changed.

Since the URLs for these packages includes a key on the RN version these packages will never be rebuilt when RN itself changes (without the 3rd part package itself is changed):

`<pkg>/output/<pkgVersion>/<rnVersion>/<hermesVersion>/<flavor>/xcframeworks/<Product>.tar.gz`

(Hermes version is always connected to the RN-version, so it could've been omitted here)

# How

Fix the detect logic in the workflow to also include the React Native version when testing if frameworks should be rebuilt, and then issue a rebuild all when RN's version changed.

# Test Plan

Replayed the detection logic against three different commits and verified that changes in RN version now results in rebuilding all packages.

# Checklist

### DIFF
--- a/.github/workflows/ios-prebuild-external-xcframeworks.yml
+++ b/.github/workflows/ios-prebuild-external-xcframeworks.yml
@@ -100,6 +100,22 @@ jobs:
           manifest="apps/bare-expo/package.json"
           changed=()
 
+          # A React Native bump changes the <rnVersion>/<hermesVersion> segments of
+          # every external artifact path, so the whole published set must be rebuilt.
+          # Read the version from `dependencies` only, mirroring the producer's source
+          # of truth in tools/src/prebuilds/Utils.ts (getVersionsInfoAsync).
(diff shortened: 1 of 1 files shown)


alanjhughes (approved):

brentvatne (approved):

brentvatne:
@expo-bot check

brentvatne:
image

claude outage 🥵

PR #44623, opened 2026-04-09
[sqlite] Handle empty statements as no-ops

barthap:
# Why

`expo-sqlite` could crash when `run()` / `runSync()` received an empty SQL statement such as `"\n"`.

The root cause is that SQLite treats empty, whitespace-only, and comment-only SQL as a special case: `sqlite3_prepare_v2()` may return `SQLITE_OK` while also returning a null statement pointer because the input produced no executable statement. This behavior is documented by SQLite and also called out in the SQLite forum: [forum thread](https://sqlite.org/forum/forumpost/70bb8576c6c084c2).

While investigating, I reproduced this difference locally with both the system SQLite and Expo’s vendored SQLite:

- my system SQLite `3.43.2`: preparing `"\n"` returned success with `stmt == NULL`, but follow-up calls were null-tolerant and did not crash
- Expo vendored SQLite `3.50.3`: preparing `"\n"` also returned success with `stmt == NULL`, but later statement APIs like `clear_bindings(NULL)` could crash

`expo-sqlite` assumed that a successful prepare always produced a valid statement pointer, so it continued into the normal execution path and called statement APIs on a null pointer. This PR fixes that mismatch by treating empty/comment-only statements as valid no-ops, which is closer to SQLite’s documented behavior and avoids introducing stricter semantics than upstream.

# How

Updated the native statement flow to treat `pointer == nil/null && !isFinalized` as a no-op statement.

On both iOS and Android, for sqlite and libsql:

- `prepare*()` now succeeds when SQLite returns success with a null statement pointer
- `run*()` short-circuits and returns a synthetic no-op result:
  - `lastInsertRowId: 0`
  - `changes: 0`
  - `firstRowValues: []`
- `step*()` returns `null`
- `getAll*()` returns `[]`
- `getColumnNames*()` returns `[]`
- `reset*()` becomes a no-op
- `finalize*()` becomes a no-op aside from marking the statement finalized

The public JS API is unchanged.

I also updated the Jest mock to model the same behavior for empty, whitespace-only, and comment-only SQL statements.

# Test Plan

- Test Suite
- My personal project - no longer crashing on executing `sqlFileString.split(';')` statements
- Added TS tests (although they rely on mocks)

# Checklist

- [x] I added a `changelog.md` entry and rebuilt the package sources according to [this short guide](https://github.com/expo/expo/blob/main/CONTRIBUTING.md#-before-submitting)
[...]

PR #49066, opened 2026-08-18
[sqlite] Throw for empty SQL statements instead of crashing

Kudo:
# Why

`expo-sqlite` crashes when a SQL string has nothing to run in it. The usual way to hit this is splitting a `.sql` file on `';'`, which leaves a trailing `"\n"` fragment.

`sqlite3_prepare_v2()` returns `SQLITE_OK` with a null statement pointer for that input. The module assumed success meant a valid pointer, so `run()` passed null to `sqlite3_clear_bindings()`. That function only null-checks inside `#ifdef SQLITE_ENABLE_API_ARMOR`, which we do not set, so it segfaults.

Reproduced on `main` against the vendored SQLite 3.53.3, compiled with the flags from `ios/ExpoSQLite.podspec`:

```
prepareStatement: rc=0 stmt=0x0
run: exsqlite3_reset -> 0
exit=139          # SIGSEGV inside exsqlite3_clear_bindings
```

This is not a regression from #48544. The `clear_bindings` body is identical in every version we have vendored, from 3.45.3 through 3.53.3. Apple's system SQLite does set `SQLITE_ENABLE_API_ARMOR`, which is why the same call is harmless against the system library and fatal against ours.

Reported in #44623 by @barthap, which fixed it by treating an empty statement as a no-op. This takes the other route and fails fast, because web already fails at prepare and `execAsync()` already covers the use case that motivated the no-op.

# How

Fail at prepare instead of letting a null statement reach the rest of the module.

- iOS and Android throw `ERR_SQLITE_EMPTY_STATEMENT` from `prepareStatement()` when SQLite reports success with no statement.
- On Android the pointer lives in C++, so `sqlite3_prepare_v2` returns `NULL_STATEMENT` (-1) for this case and Kotlin maps that to the exception. No sqlite3 result code is negative, so the sentinel cannot collide. This keeps the check to one JNI call, which matters because statement caching was removed in #48543 and `runAsync` and friends prepare per query.
- Web already threw in this spot. Only its message changed, so all platforms now say the same thing.
- The jest mock asks node:sqlite instead of pattern-matching the SQL. node accepts the source at `prepare()`, then throws `ERR_INVALID_STATE` from every method on the statement, and only for this case.

One check at prepare covers everything downstream, so `run`, `step`, `getAll`, `reset`, `getColumnNames`, and `finalize` need no guards of their own.

The message points at `execAsync()`, which is the supported way to run a `.sql` file. `sqlite3_exec` returns `SQLITE_OK` for multi-statement scripts with comments before and after the statements, and for empty input.

Only paths that already crash or throw behave differently:

| `db.runAsync('\n')` | before | after |
| --- | --- | --- |
| iOS, Android | SIGSEGV | throws `ERR_SQLITE_EMPTY_STATEMENT` |
| web | throws `Failed to prepare statement` | same failure, shared message |

A statement that follows a bare `;` or a comment still prepares normally. `sqlite3_prepare_v2` skips leading separators and comments, so the null pointer only appears when the whole string has nothing to run.

# Test Plan

Jest, run locally and passing (39 tests in `packages/expo-sqlite/src/__tests__`):

- `prepareAsync`, `prepareSync`, `runAsync`, and `runSync` reject for whitespace-only source
- `prepareAsync` rejects for comment-only source
- statements after `';'` and after a comment still prepare, which guards against an over-broad emptiness check
- `execAsync` runs a script with comments around the statements

Red then green: with the mock change reverted, the five new "should throw" tests fail. With it in place all 39 pass.

New cases in `apps/test-suite/tests/SQLite.ts` are written but not yet run on a device or in a browser. They sit in the plain `describe('Statements')` block, so they exercise iOS, Android, and web, and they assert the same message on all three. This needs a test-suite run before merge.

# Checklist

### DIFF
--- a/apps/test-suite/tests/SQLite.ts
+++ b/apps/test-suite/tests/SQLite.ts
@@ -448,6 +448,59 @@ CREATE TABLE IF NOT EXISTS posts (post_id INTEGER PRIMARY KEY NOT NULL, content 
       await db.closeAsync();
     });
 
+    it('should run a whole script with comments via execAsync', async () => {
+      const db = await SQLite.openDatabaseAsync(':memory:');
+      await db.execAsync(`
+-- set up the table
+DROP TABLE IF EXISTS users;
+CREATE TABLE IF NOT EXISTS users (user_id INTEGER PRIMARY KEY NOT NULL, name VARCHAR(64));
+/* seed it */
+INSERT INTO users (user_id, name) VALUES (1, 'Tim Duncan');
+-- done
+`);
+      const rows = await db.getAllAsync('SELECT * FROM users');
+      expect(rows.length).toBe(1);
+      await db.closeAsync();
+    });
+
+    it('should prepare a statement that follows a comment', async () => {
+      const db = await SQLite.openDatabaseAsync(':memory:');
+      await db.execAsync(`
+DROP TABLE IF EXISTS users;
+CREATE TABLE IF NOT EXISTS users (user_id INTEGER PRIMARY KEY NOT NULL, name VARCHAR(64));
+`);
+      const statement = await db.prepareAsync('-- pick everything\nSELECT * FROM users');
+      expect(await statement.getColumnNamesAsync()).toEqual(['user_id', 'name']);
+      await statement.finalizeAsync();
+      await db.closeAsync();
+    });
+
+    it('should throw when preparing an empty or comment-only statement', async () => {
+      const db = await SQLite.openDatabaseAsync(':memory:');
+      for (const source of ['\n', '   ', '-- nothing to run', ';']) {
+        let error = null;
+        try {
+          await db.prepareAsync(source);
+        } catch (e) {
+          error = e;
+        }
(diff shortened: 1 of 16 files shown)


barthap:
Nice, we can close #44623 if this supersedes it 😄

Kudo:
> Nice, we can close #44623 if this supersedes it 😄

still cooking on different solution and find the root cause

Kudo:
Dropping this in favor of #49068.

`SQLITE_ENABLE_API_ARMOR` is enough to stop the SIGSEGV from empty SQL. `sqlite3_prepare_v2()` still returns success with a null statement, and follow-up C calls no longer crash. That is the same reason Apple's system SQLite did not crash.

We're keeping SQLite's contract: empty SQL is a successful prepare of nothing. `run()` may still throw a SQLite error (`SQLITE_MISUSE`). Use `execAsync()` for a whole `.sql` file.

#44623's no-op behavior is a separate product choice and is not part of this change.

PR #49068, opened 2026-08-18
[sqlite] Enable SQLITE_ENABLE_API_ARMOR in native builds

Kudo:
# Why

fix crash when passing empty statement like `\n` to exec
supersede #44623

# How

enable `SQLITE_ENABLE_API_ARMOR` as ios built-in sqlite.

# Test Plan

added e2e test to cover empty statement

# Checklist


Kudo:
ios build error is failed on main. 0.88 rc artifacts are slow to download

alanjhughes (approved):

PR #49083, opened 2026-08-18
[image] Sanitize URLs in Observe oversized-image reports

tsapeta:
# Why

Oversized-image events from the expo-observe integration send the full image URL off-device, so secrets embedded in it (signing tokens or API keys in query params, basic-auth credentials, the entire payload of a `data:` URL, or a personal-photo identifier in a `ph://` URL) leak into the reports. Flagged in the review of the integration (ENG-26040).

# How

Only `http(s)`, `file`, and `android.resource` URLs are reported at all: those identify developer-owned content the developer can act on (`android.resource://` is how a bundled static image surfaces on Android in release builds), while other schemes carry device-local or user-library content and fail safe. URLs are parsed with the built-in WHATWG parser (installed globally by the expo winter runtime) and reported in normalized form, with the query string and fragment removed before they leave the device, and dedup uses the reported form so rotating signed variants of one image produce a single event. The new `includeUrlParams` integration option restores the query string and fragment for setups where that's safe; basic-auth credentials are removed regardless. Every event carries a `urlSanitized` attribute that tells whether the URL was modified beyond WHATWG normalization.

One residual risk we accept: `file` URLs can contain user-chosen filenames (a downloaded image saved under a meaningful name). We keep them because an oversized bundled asset is a common developer mistake and the path is what identifies it.

# Test Plan

Unit tests cover the scheme allowlist, the query and fragment stripping, credential removal, the `urlTruncated` attribute, the dedup behavior, and the `includeUrlParams` option. `et check-packages expo-image` passes. In observe-tester, the expo-image "Too big" example now loads its image with fake signing params in the URL, so the Sessions tab shows the sanitized URL and `urlSanitized` live; `includeUrlParams` can be flipped in the app's `Observe.configure` call.

### DIFF
--- a/packages/expo-image/src/observe.ts
+++ b/packages/expo-image/src/observe.ts
@@ -121,6 +121,14 @@ export function reportIfOversized(state: IntegrationState, image: LoadedImage): 
   if (!image.url || !(width > 0) || !(height > 0)) {
     return;
   }
+  // Screen area is in points; the decoded image is in pixels, so convert with pixelRatio² to get
+  // the screen's physical pixel count.
+  const budget = screenWidth * screenHeight * pixelRatio * pixelRatio;
+  if (!(budget > 0) || width * height <= budget * state.threshold) {
+    return;
+  }
+  // Sanitizing after the size check keeps the common per-load path cheap: only oversized images
+  // pay for the URL parse.
   const sanitizedUrl = sanitizeUrl(image.url, state.includeUrlParams);
   if (!sanitizedUrl) {
     return;
@@ -129,12 +137,6 @@ export function reportIfOversized(state: IntegrationState, image: LoadedImage): 
   // query parameters (such as rotating signed URLs) into a single event.
   const url = sanitizedUrl.url;
   if (state.reported.has(url)) {
-    return;
-  }
-  // Screen area is in points; the decoded image is in pixels, so convert with pixelRatio² to get
-  // the screen's physical pixel count.
-  const budget = screenWidth * screenHeight * pixelRatio * pixelRatio;
-  if (!(budget > 0) || width * height <= budget * state.threshold) {
     return;
   }
   state.reported.add(url);
--- a/packages/expo-image/src/observe.ts
+++ b/packages/expo-image/src/observe.ts
@@ -33,7 +33,8 @@ export type ExpoImageIntegrationConfig = {
    * sensitive values such as signing tokens or API keys. Enable this only when your image URLs
    * are safe to send off-device in full. Regardless of this setting, basic-auth credentials are
    * always removed from the URL, and only `http(s)`, `file`, and `android.resource` URLs are
-   * reported (other schemes, such as `data:` or `ph://`, never leave the device).
+   * reported (other schemes, such as `data:` or `ph://`, never leave the device). An event whose
+   * URL was modified before sending carries a `urlSanitized: true` attribute.
    *
    * @default false
    */
@@ -127,6 +128,9 @@ export function reportIfOversized(state: IntegrationState, image: LoadedImage): 
       body: `Image loaded at ${width}×${height}px is far larger than this device's screen (${screenWidth}×${screenHeight}pt @${pixelRatio}x). Constrain it with the maxWidth/maxHeight load options.`,
       attributes: {
         url,
+        // Present only when sanitization changed the URL, mirroring the navigation integration's
+        // `urlHidden` attribute.
+        ...(url !== image.url ? { urlSanitized: true } : {}),
         imageWidth: width,
         imageHeight: height,
         screenWidth,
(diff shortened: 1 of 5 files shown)


tsapeta:
@expo-bot verify

kadikraman (commented):
Could we add a `urlTruncated` or something like that, if the URL has been modified prior to sending? Similar to the navigation integration.

packages/expo-image/src/observe.ts
```diff
@@ -61,15 +74,42 @@ export type LoadedImage = {
...
+  if (!includeUrlParams) {
+    const paramsStart = sanitized.search(/[?#]/);
+    if (paramsStart !== -1) {
+      sanitized = sanitized.slice(0, paramsStart);
+    }
+  }
+  // Userinfo ends at the last `@` before the authority ends (matching WHATWG parsing, so an
+  // unencoded `@` inside a password is consumed too), while `/`, `?`, and `#` bound the match, so
+  // an `@` later in the path or query never matches.
+  return sanitized.replace(/^([^:/?#]+:\/\/)[^/?#]*@/, '$1');
```
Ubax: Can you extract this regex and add a bit more explanation to each part so that it's easier to grasp?

packages/expo-image/src/observe.ts
```diff
@@ -61,15 +74,42 @@ export type LoadedImage = {
...
+// Truncates a URL at its query string and fragment (unless `includeUrlParams` opts in) and always
+// removes userinfo credentials from the authority, so values like signing tokens, API keys, or
+// basic-auth passwords never leave the device.
+function sanitizeUrl(url: string, includeUrlParams: boolean): string {
+  let sanitized = url;
+  if (!includeUrlParams) {
+    const paramsStart = sanitized.search(/[?#]/);
+    if (paramsStart !== -1) {
+      sanitized = sanitized.slice(0, paramsStart);
+    }
```
Ubax: Can we use the built-in `URL` parsing for this?

In expo-router we have `parseUrlUsingCustomBase` which may be helpful here, but I think even `new URL` should suffice here

packages/expo-image/src/observe.ts
Ubax: If we use the built-in parser, then we can use
```
url.username = '';
url.password = '';
```

to remove the auth.

packages/expo-image/src/observe.ts
```diff
@@ -61,15 +74,42 @@ export type LoadedImage = {
...
-  if (!url || !(width > 0) || !(height > 0)) {
+  const { width, height, screenWidth, screenHeight, pixelRatio } = image;
+  if (!image.url || !(width > 0) || !(height > 0)) {
+    return;
+  }
+  // Only remote images, local files, and bundled Android resources are reported: those URLs
+  // identify developer-owned content that the developer can act on. Every other scheme fails safe
+  // regardless of `includeUrlParams`, because it carries device-local or user-library content (the
+  // whole payload for `data:`, a stable personal-photo identifier for `ph://` and `content://`).
+  if (!/^(https?|file|android\.resource):/i.test(image.url)) {
```
Ubax: Can we fold this into `sanitizeUrl`?

For not allowed protocols it would return `undefined` or `null`. 

Then we can reuse the URL parsing for this as well

tsapeta:
I've made a few changes in this PR:

- Added the modified-URL flag @kadikraman asked for. It's named `urlSanitized` rather than `urlTruncated` because credential removal isn't a truncation.
- Rewrote `sanitizeUrl` around the built-in `URL`, as @Ubax suggested. The scheme allowlist is folded into it (checks `url.protocol`, returns `null` for disallowed schemes or strings the parser rejects), credentials clear through the `username`/`password` setters, and the query and fragment through `search`/`hash`. Both regexes are gone. This is safe on native because the integration only activates in Expo apps, where the winter runtime installs `whatwg-url-minimum` as the global `URL` before any product code runs. It's worth noting that the `URL` implementation from React Native wouldn't work in our case as it doesn't support setters.
- Two behavior consequences of the parser: reported URLs are now in normalized WHATWG form (which also makes dedup canonical), and `urlSanitized` compares against the normalized original so normalization alone doesn't flag an event. Unparseable strings, like the scheme-less names Android gives bundled assets in release `loadAsync`, fail safe and are covered by tests.
- observe-tester: the expo-image "Too big" example now loads its image with fake signing params and a fragment in the URL, and the app's `Observe.configure` spells out `includeUrlParams: false`, so the Sessions tab demonstrates the stripped URL and `urlSanitized` live. Verified on device.

packages/expo-image/src/observe.ts
```diff
@@ -61,15 +76,58 @@ export type LoadedImage = {
...
   if (!state.enabled || !state.appMetrics) {
     return;
   }
-  const { url, width, height, screenWidth, screenHeight, pixelRatio } = image;
-  if (!url || !(width > 0) || !(height > 0)) {
+  const { width, height, screenWidth, screenHeight, pixelRatio } = image;
+  if (!image.url || !(width > 0) || !(height > 0)) {
+    return;
+  }
+  const sanitizedUrl = sanitizeUrl(image.url, state.includeUrlParams);
```
Ubax, line 132: [💡 suggestion] `sanitizeUrl` now runs on every `imageLoaded` event, before the size check — so every image load in the app pays a WHATWG parse plus two `toString` calls, and only oversized images need it. Moving the sanitize + dedup block after the budget check keeps the observable behavior identical (the dedup set only ever holds oversized URLs) and restores the cheap early-out for the common case. Minor cost either way, so take it or leave it.

Ubax (approved):

PR #49084, opened 2026-08-18
[image][Android] Report the pre-downsample source size to the Observe integration

tsapeta:
# Why

The expo-observe oversized-image check was effectively dead on Android for its main case (a large image in a small view): Glide downscales the bitmap to the view size at decode time, so the drawable's intrinsic size never exceeded the screen budget. Flagged in the review of the integration (ENG-26040).

# How

`ContentFitDownsampleStrategy` records the source dimensions at decode time into a dedicated `decodeSourceWidth`/`decodeSourceHeight` pair on the target, and the listener reports those instead of the drawable size, falling back to the drawable size when decoding was skipped (cache hits). The dedicated pair keeps this a pure reporting change: it's cleared when a target is reused so a cache hit can't report an earlier image's dimensions, while the layout-facing `sourceWidth`/`sourceHeight` (which the `scale-down` transformation matrix reads) keeps its shipped lifecycle. The listener captures the values before its main-queue hop, since a new load can reuse the target and clear them in that gap.

Known false negatives, unchanged by this PR: loads that skip the full decode and aren't deduped in the current session (a local image whose downscaled resource is served from Glide's disk cache across app runs, or an image cached before the integration was configured) still report the drawable size.

# Test Plan

expo-image has no Kotlin test harness, so this was verified by inspection of the Glide decode path; the review bot additionally ran the real `ContentFitDownsampleStrategy` against Glide 5.0.5 on a JVM and confirmed the recorded values and the before/after verdicts. `et check-packages expo-image` passes.


tsapeta:
@expo-bot verify

PR #49086, opened 2026-08-18
[android][ios][file-system] Fix File.size returning null

ACHP:
# Why

File.size is typed as number, but native Android and iOS bridges returned null when the file was missing or unreadable. For instance here on android (https://github.com/expo/expo/blob/main/packages/expo-file-system/android/src/main/java/expo/modules/filesystem/FileSystemFile.kt#L205-L211).

But TS type it as `number`

https://github.com/ACHP/expo/blob/5ff5560a425024bfc2f7bba3c67e83498cca1eaf/packages/expo-file-system/src/internal/NativeFileSystem.types.ts#L323

image

# How

Return 0 from both native bridges for those cases. File implements Blob, whose size must be a number, so number | null is not possible.

# Test Plan

Ran pnpm exec et check-packages expo-file-system.

# Checklist

- [x] I added a changelog.md entry and rebuilt the package sources according to this
    short guide (https://github.com/expo/expo/blob/main/CONTRIBUTING.md#-before-submitting)

- [x] This diff will work correctly for npx expo prebuild & EAS Build (eg: updated a
    module plugin).

### DIFF
--- a/packages/expo-file-system/CHANGELOG.md
+++ b/packages/expo-file-system/CHANGELOG.md
@@ -14,7 +14,7 @@
 
 ### 🐛 Bug fixes
 
-- [Android][iOS] Fix `File.size` returning `null` for a missing or unreadable file.
+- [Android][iOS] Fix `File.size` returning `null` for a missing or unreadable file. ([#49086](https://github.com/expo/expo/pull/49086)) by [@ACHP](https://github.com/ACHP))
 - [iOS] Fix wrong permissions for text() and bytes(). ([#42422](https://github.com/expo/expo/pull/42422)) by [@simoneldevig](https://github.com/simoneldevig))
 - Fixed `copyAsync` on iOS copying the unedited original when a `ph://` asset has edits applied in Photos. ([#48248](https://github.com/expo/expo/pull/48248) by [@CoffeeFlux](https://github.com/CoffeeFlux))
 - Fixed iOS file previews rejecting a new preview while the previous Quick Look dismissal animation is still finishing. ([#47947](https://github.com/expo/expo/pull/47947) by [@eliotgevers](https://github.com/eliotgevers))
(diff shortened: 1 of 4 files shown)


barthap:
Thanks for catching this!
Indeed, it was a mismatch between the public API and native implementation.

> File implements Blob, whose size must be a number, so `number | null` is not possible.

For a while, I was hesitant about whether returning `null` is a better option and whether the public TS type should be changed instead, but `Blob` returning 0 is a valid reason.

I'll let CI pass and merge the fix

barthap:
@expo-bot check

packages/expo-file-system/CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 
 ### 🐛 Bug fixes
 
+- [Android][iOS] Fix `File.size` returning `null` for a missing or unreadable file.
```
barthap:
```suggestion
- [Android][iOS] Fix `File.size` returning `null` for a missing or unreadable file. ([#49086](https://github.com/expo/expo/pull/49086)) by [@ACHP](https://github.com/ACHP))
```

barthap (changes requested):
We need to update test-suite tests too, CI is failing - this is the thing expo-bot is spamming about 😅 

The code that needs to be updated: L1882 `expect` and test name
https://github.com/expo/expo/blob/ec40b10f7d7a9f351d65db1fc076c9549c414693/apps/test-suite/tests/FileSystem.ts#L1878-L1884

barthap (approved):
LGTM, thanks!

Heads up that I'll merge this into main only, so it lands in SDK 58 and not in an SDK 56 or 57 patch. The `null` has been there since SDK 54, and some apps might use `file.size === null` as an existence check, so I don't want to change that in a patch release.

ACHP:
@barthap Yes that makes sense, thanks you for being so responsive and getting this merged so quickly !

Issue #48985, opened 2026-08-16
[ios][screen-capture] preventScreenshots() is not reentrant: a second call while active corrupts the layer tree and permanently black-screens the app


wv-warredr:
### Minimal reproducible example

https://github.com/wv-warredr/expo-screen-capture-reentrancy-repro

The repository is a fresh `npx create-expo-app --template blank` with `expo-screen-capture` added and a three-button `App.js` (inlined under "Steps to reproduce" below); nothing else was changed.

### Steps to reproduce

Platform: **iOS only**. Android's implementation uses `FLAG_SECURE`, which is idempotent, and web is a no-op. The bug is in the native module, so it affects Expo Go, dev clients, and standalone builds alike. Reproduced with `expo-screen-capture@57.0.2` (SDK 57) on the iOS Simulator (iPhone 17 Pro, iOS 26.5 runtime, Release configuration), and hit in our production app with `expo-screen-capture@55.0.13` (SDK 55) on physical iPhones running iOS 26. The repro repository's README shows screenshots of all four states: rendering normally while protected, black after the second key, still black after both keys are released, and recovered after force-quitting.

1. Clone the repro, `npm install`, `npx expo run:ios --configuration Release`.
2. Tap **"1. prevent (key 'a')"**, which calls `ScreenCapture.preventScreenCaptureAsync('a')`. This works as intended: the app keeps rendering normally and screenshots are blanked.
3. Tap **"2. prevent (key 'b')"**, which calls `ScreenCapture.preventScreenCaptureAsync('b')` while protection is already active.
   - **Expected:** protection simply stays active; per the docs, a second prevent call is a no-op while one is active.
   - **Actual:** the entire app turns black immediately. This is the live UI going black on screen, not just in captures. Only the status bar keeps rendering: it lives in a separate window, and only the key window's layer gets re-parented.
4. Tap **"3. allow both keys"** (the buttons still hit-test, so tap where the button was), which calls `allowScreenCaptureAsync('a')` and `allowScreenCaptureAsync('b')`; the second call invokes the native `allowScreenCapture`.
   - **Expected:** rendering is restored.
   - **Actual:** the screen stays black. The original layer hierarchy is unrecoverable at this point (see "Root cause"). Force-quitting and relaunching the app is the only way to recover; after a relaunch the app renders normally again.

`App.js`:

```jsx
import { useState } from 'react';
import { Pressable, StyleSheet, Text, View } from 'react-native';
[...]

PR #49088, opened 2026-08-18
[ios][screen-capture] Redraw the window after secure-canvas re-parenting and harden the protect/restore state machine

maxlapides:
# Why

`preventScreenshots()` re-parents the entire key-window layer into a secure `UITextField` canvas. Display work that is pending on `drawRect`-backed views at that moment — for example `react-native-svg` surfaces — can be dropped by the re-parent. `react-native-svg` latches its internal `dirty` flag until an actual draw happens (`RNSVGRenderable.invalidate` early-returns while dirty), so once the draw is lost the content stays blank against every subsequent JS-driven invalidation until something else forces a redraw.

We hit this in production: a sheet guarded by `usePreventScreenCapture` intermittently opened with all of its SVG-drawn backgrounds invisible (everything else rendered fine), and tapping anywhere instantly restored all of them — the tap gives UIKit another display pass. It reproduced only on physical devices, because the failure is a race between the re-parent and the pending display work.

While fixing this we also found that the protect/restore state machine can wedge:

- `protectionTextField` is latched even when the re-parent never happened (nil `superlayer`, or an empty `textField.layer.sublayers`), and a failed restore skips the state reset. Combined with a re-entrancy guard (as proposed in #49087 for #48985) this would permanently disable screenshot protection for the rest of the process with no error surfaced to JS.
- `allowScreenshots()` re-resolves the key window at restore time instead of remembering which window it re-parented, so it can restore the wrong window if key-window status changed in between.

Related: #48985 / #49087 cover the re-entrancy half of this. This PR includes the same guard because the state-machine hardening here is what makes that guard safe (latch-only-on-success + unconditional reset ensure the guard can never permanently block re-arming). Happy to rebase on #49087 if it lands first.

# How

In `ScreenCaptureModule.swift`:

- After both re-parents (protect and restore), call a new `setNeedsDisplayRecursively(in:)` over the re-parented window, plus deferred passes one run-loop turn later and 0.1s later to also cover display work issued immediately after the re-parent (e.g. the React Native layout/props cycle that follows). The helper marshals itself to the main thread, which also covers the `OnDestroy` → `allowScreenshots()` path that does not run on the main queue.
- `preventScreenshots()` now returns early if protection is already active, uses guarded early-returns that clean up the just-added text-field layer when the re-parent cannot happen, and latches `protectionTextField` / `originalParent` / `reparentedWindow` only after the re-parent succeeded.
- `allowScreenshots()` resets all protection state unconditionally via `defer`, and restores the recorded re-parented window (held weakly) rather than the current key window, with a fallback that removes the secure canvas if the window is gone.

# Test Plan

- `xcrun swiftc -parse` passes on the modified file.
- The state machine was walked across all paths: successful re-parent, nil `superlayer`, empty `sublayers`, deallocated window, repeated prevent (guard), allow-without-prevent (no-op), and prevent → allow → prevent cycles.
- The same change (as a package patch against `expo-screen-capture@57.0.1`) is running in our app; sweeps fire on protect/release of the guarded sheet with no visual or performance regressions in the surrounding flows. The production race is intermittent and physical-device-only, so we are continuing to soak it on devices.
- Reviewer reproduction: mount a screen containing `react-native-svg` content whose props/layout change in the same commit that mounts `usePreventScreenCapture` (i.e. reveal data and the guard together). Without this change, on a physical device the SVG content intermittently stays blank until touched; with it, the deferred sweeps restore the pending draws. The #48985 repro app covers the re-entrancy path deterministically.

# Checklist

- [x] I added a `CHANGELOG.md` entry (no package sources need rebuilding — native-only change)
- [ ] This diff will work correctly for `npx expo prebuild` & EAS Build (no config plugin changes)
- [ ] Conforms with the Documentation Writing Style Guide (no docs changes)

### DIFF
--- a/packages/expo-screen-capture/CHANGELOG.md
+++ b/packages/expo-screen-capture/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 - [iOS] Fixed the screen-recording overlay staying attached (permanent black screen) after `allowScreenCaptureAsync` is called while a recording is active. ([#48000](https://github.com/expo/expo/pull/48000) by [@bluespore](https://github.com/bluespore))
 - [iOS] Fixed screenshot prevention and the recording overlay attaching to a window in a background scene rather than the one on screen. ([#48372](https://github.com/expo/expo/pull/48372) by [@alanjhughes](https://github.com/alanjhughes))
-- [iOS] Fixed `drawRect`-backed views (for example `react-native-svg` content) staying blank when screenshot prevention re-parents the window layer while their display work is pending, and hardened the protect/restore state machine so a failed re-parent can no longer permanently disable protection or restore the wrong window.
+- [iOS] Fixed `drawRect`-backed views (for example `react-native-svg` content) staying blank when screenshot prevention re-parents the window layer while their display work is pending, and hardened the protect/restore state machine so a failed re-parent can no longer permanently disable protection or restore the wrong window. ([#49088](https://github.com/expo/expo/pull/49088) by [@maxlapides](https://github.com/maxlapides))
 
 ### 💡 Others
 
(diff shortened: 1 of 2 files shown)


brentvatne:
@expo-bot verify with fable

PR #49098, opened 2026-08-19
[router] Queue navigation dispatches after commit

Ubax:
## Accepted breaking changes

- `navigation.dispatch` and navigation helper methods now apply after the current React commit. Immediate `getState()` and `canGoBack()` reads observe the pre-dispatch state.
- Functional actions are no longer accepted by `navigation.dispatch`; use `navigation.dispatchSync` when a functional action or synchronous application is required.

### DIFF
--- a/packages/expo-router/src/global-state/routerRegistry.tsx
+++ b/packages/expo-router/src/global-state/routerRegistry.tsx
@@ -1,6 +1,14 @@
 'use client';
 
-import { createContext, use, useMemo, useState, type PropsWithChildren } from 'react';
+import {
+  createContext,
+  use,
+  useMemo,
+  useRef,
+  useState,
+  type PropsWithChildren,
+  type RefObject,
+} from 'react';
 
 import { useClientLayoutEffect } from '../react-navigation/core/useClientLayoutEffect';
 import type {
@@ -28,10 +36,15 @@ type RouterRegistrySetters = {
 
 // React components read this map during render, so React state is intentional.
 export const RouterRegistryContext = createContext<RouterRegistry | undefined>(undefined);
+export const RouterRegistryRefContext = createContext<RefObject<RouterRegistry> | undefined>(
+  undefined
+);
 const RouterRegistrySettersContext = createContext<RouterRegistrySetters | undefined>(undefined);
 
 export function RouterRegistryProvider({ children }: PropsWithChildren) {
   const [registry, setRegistry] = useState<RouterRegistry>(() => new Map());
+  const registryRef = useRef(registry);
+  registryRef.current = registry;
   const setters = useMemo<RouterRegistrySetters>(
     () => ({
       register(stateKey, entry) {
@@ -60,7 +73,9 @@ export function RouterRegistryProvider({ children }: PropsWithChildren) {
 
   return (
     <RouterRegistrySettersContext.Provider value={setters}>
-      <RouterRegistryContext.Provider value={registry}>{children}</RouterRegistryContext.Provider>
+      <RouterRegistryRefContext.Provider value={registryRef}>
+        <RouterRegistryContext.Provider value={registry}>{children}</RouterRegistryContext.Provider>
+      </RouterRegistryRefContext.Provider>
     </RouterRegistrySettersContext.Provider>
   );
 }
--- a/packages/expo-router/src/modal/web/ModalStack.tsx
+++ b/packages/expo-router/src/modal/web/ModalStack.tsx
@@ -94,7 +94,7 @@ const ModalStackView = ({ state, navigation, descriptors }: ModalStackViewProps)
     index: nonModalIndex,
   };
 
-  const pop = makePopAction(navigation.dispatch, state.key);
+  const pop = makePopAction(navigation.dispatchSync, state.key);
 
   const dismiss = useCallback(() => {
     navigation.goBack();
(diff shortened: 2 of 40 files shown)


packages/expo-router/src/modal/web/ModalStack.tsx
```diff
@@ -94,7 +94,7 @@ const ModalStackView = ({ state, navigation, descriptors }: ModalStackViewProps)
     index: nonModalIndex,
   };
 
-  const pop = makePopAction(navigation.dispatch, state.key);
+  const pop = makePopAction(navigation.dispatchSync, state.key);
```
Ubax: [🏗️ **design**] `dismiss` a few lines below still calls the queued `navigation.goBack()`. `onDismiss` fires after the DOM dialog has already closed, so JS state needs to catch up synchronously — the same reason this PR converts `ExperimentalStackView`, `StackView`, and `makePopAction` here to `dispatchSync`. With the queued version the route stays in state for one commit, and a second dismiss event in that window would double-pop. Consider `navigation.dispatchSync(CommonActions.goBack())` in `dismiss` as well.

packages/expo-router/src/react-navigation/core/useNavigationHelpers.tsx
```diff
@@ -44,12 +50,20 @@ export function useNavigationHelpers<
...
   const parentNavigationHelpers = use(NavigationContext);
 
   return React.useMemo(() => {
-    const dispatch = (op: Action | ((state: State) => Action)) => {
-      const action = typeof op === 'function' ? op(getState()) : op;
+    const dispatchSync = (op: Action | ((state: State) => Action)) => {
+      const state = getState();
+      if (!registryRef.current?.has(state.key)) {
+        throw new Error(
+          `Cannot dispatch synchronously because navigator '${state.key}' is not registered. This is most likely a bug in expo-router. Please report it at https://github.com/expo/expo/issues.`
```
Ubax: [🏗️ **design**] A queued `NAVIGATOR_ACTION` can drain after its navigator unmounted (dispatch during the last commit before unmount). The captured `dispatchSync` closure then hits this throw, and `routingQueue.run` catches it and prints "This is most likely a bug in expo-router. Please report it". That race is legitimate app behavior, not a library bug, so the message sends users to file false bug reports. Consider distinguishing the unmounted case: drop the action with a warning like the existing "container is not mounted" path, and reserve the report-a-bug wording for a `dispatchSync` called while the navigator is mounted.

packages/expo-router/src/react-navigation/core/useNavigationCache.tsx
```diff
@@ -143,7 +162,8 @@ export function useNavigationCache<
       ...helpers,
       // FIXME: too much work to fix the types for now
       ...(emitter.create(route.key) as any),
-      dispatch: (thunk: Thunk) => withStack(() => dispatch(thunk)),
+      dispatch: (action: NavigationAction) => withStack(() => dispatch(action)),
```
Ubax: [💡 suggestion] `withStack` sets `stackRef.current` and clears it in `finally`, but the queued `dispatch` only enqueues inside that window — the action runs later at queue drain, when the stack is already cleared. Devtools lose the origin stack for every queued dispatch. Consider capturing the stack into the intent (e.g. on `NAVIGATOR_ACTION` metadata) and restoring it around the drain.

packages/expo-router/src/react-navigation/core/useNavigationCache.tsx
```diff
@@ -98,10 +99,28 @@ export function useNavigationCache<
...
       }
     };
 
+    const dispatch = (action: NavigationAction) => {
+      if (typeof action === 'function') {
+        throw new Error(FUNCTIONAL_DISPATCH_ERROR);
+      }
+
+      const state = getState();
+      if (isRoutePreloadedInStack(state, route)) {
```
Ubax: [💡 suggestion] This preloaded check runs at enqueue time, but the action applies at queue drain. State can change in between: a screen promoted before the drain still gets its action dropped here, and a screen preloaded after enqueue slips through. Probably rare in practice — a short comment noting the enqueue-time semantics, or re-checking at drain, would make the intent explicit.

packages/expo-router/src/global-state/routerRegistry.tsx
```diff
@@ -28,10 +36,15 @@ type RouterRegistrySetters = {
...
 export const RouterRegistryContext = createContext<RouterRegistry | undefined>(undefined);
+export const RouterRegistryRefContext = createContext<RefObject<RouterRegistry> | undefined>(
+  undefined
+);
 const RouterRegistrySettersContext = createContext<RouterRegistrySetters | undefined>(undefined);
 
 export function RouterRegistryProvider({ children }: PropsWithChildren) {
   const [registry, setRegistry] = useState<RouterRegistry>(() => new Map());
+  const registryRef = useRef(registry);
+  registryRef.current = registry;
```
Ubax: [💡 suggestion] The ref is updated during render, while `register`/`unregister` run in layout effects and only land in the ref on the *next* render. During the layout phase of a commit the ref lags reality in both directions: a navigator registered in this commit's layout effects isn't in `registryRef.current` yet, so a `dispatchSync` from a sibling layout effect can falsely throw "not registered"; conversely an unregistered one still passes. Mutating a persistent Map directly inside `register`/`unregister` (keeping the React state copy for render consumers) would close that window.

PR #49101, opened 2026-08-19
[audio] Add `doNotMixPersistent` interruption mode

behenate:
# Why

`doNotMix` uses transient audio focus, allowing interrupted audio from other apps to resume after focus is released. Add a persistent mode for apps that need exclusive focus without triggering automatic resumption.

# How

Add `doNotMixPersistent` across platforms. Android uses `AUDIOFOCUS_GAIN` and refreshes active or delayed focus requests when the mode changes. iOS uses the non-mixing category without `.notifyOthersOnDeactivation`. Preserve playback intent, paused state, and ducked volume throughout focus changes.

There is a lot more changes on Android - the automated review I was running locally kept finding a lot of related or semi-related edge cases. Right now the focus management seems to be working quite well.

# Test Plan

Tested manually in NCL on iOS and Android by interrupting other audio source playing in the background.


behenate:
@expo-bot review

alanjhughes (approved):

PR #49110, opened 2026-08-19
[router] Refactor router to use global state

Ubax:
# Why

# How

# Test Plan

# Checklist

## Accepted breaking changes

- Remove `resetRoot` from `NavigationContainerRef` and `createNavigationContainerRef`. Use `router.replace(...)` or dispatch `CommonActions.reset(...)` with a `target` instead.


packages/expo-router/src/react-navigation/core/BaseNavigationContainer.tsx
```diff
@@ -143,22 +162,20 @@ export function BaseNavigationContainer({
   });
 
   const resetRoot = useLatestCallback((state?: PartialState<NavigationState> | NavigationState) => {
```
Ubax: [🚨 **critical**] `resetRoot` applies synchronously via `handleAction`, while `navigationRef.dispatch()` now queues through `routingQueue` and applies after commit. `navigationRef.dispatch(navigateAction); navigationRef.resetRoot(state)` therefore applies the reset first and the navigate second — the reverse of the call order and of pre-PR behavior. The same ordering hazard exists for any code that mixes `dispatch` and `dispatchSync`. Either route `resetRoot` through the queue or document the sync/queued split explicitly.

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
```diff
@@ -347,182 +342,33 @@ export function useNavigationBuilder<
...
+  const { getStateForKey, handleAction } = use(NavigationBuilderContext);
+  if (
+    currentState === undefined ||
+    currentState.stale !== false ||
+    typeof currentState.key !== 'string' ||
+    !Array.isArray(currentState.routeNames) ||
+    !isStateValid(currentState)
+  ) {
+    throw new Error(
+      'A navigator received a missing, partial, or incompatible navigation state. Navigation state is created globally now, so the container must provide a complete seeded state for every mounted navigator.'
```
Ubax: [🚨 **critical**] Upstream (and the pre-PR fork) re-initialized state in this case; this now throws during render. Because the offending state lives in the global store, the throw repeats on every re-render — the app is unrecoverable without a reload. Reachable cases: a `useNavigationBuilder`-based navigator rendered inside a screen component (not part of the seeded `routeNode` tree) gets `currentState === undefined`; a restored state whose nested `type` no longer matches the navigator crashes instead of resetting. At minimum the error should name the navigator, and there should be an explicit decision on whether unseeded navigators are supported at all.

packages/expo-router/src/fork/useBrowserHistorySync.ts
```diff
@@ -0,0 +1,312 @@
...
+      if (
+        previousFocusedState &&
+        focusedState &&
+        // A matching popstate already changed the browser URL, so don't write another entry.
+        !pendingOperation
+      ) {
+        const historyDelta =
+          getHistoryLength(focusedState) - getHistoryLength(previousFocusedState);
+
+        if (historyDelta > 0) {
```
Ubax: [🚨 **critical**] The routing queue drains all intents in one effect, so N dispatches produce one commit and one `'state'` event — but this branch calls `history.push` at most once regardless of `historyDelta`. Scenario: `router.push('/a'); router.push('/b')` in one handler adds two screens but one browser-history entry. One browser Back then pops one screen while restoring the URL from before both pushes, and URL/state diverge from that point on. Push `historyDelta` times (or commit per intent). There is also no test draining two `NAVIGATE_TO_HREF` intents in one tick and asserting history entries.

packages/expo-router/src/react-navigation/routers/BaseRouter.tsx
```diff
@@ -85,15 +88,28 @@ export const BaseRouter = {
             return null;
           }
 
-          return {
+          const minter = createRouteKeyMinter({
+            key: nextState.key,
+            routeKeySeq: nextState.routeKeySeq ?? 0,
```
Ubax: [🚨 **critical**] The error below declares `routeKeySeq` required for a `RESET` payload, but the `stale: false` path silently defaults it to `0`, restarting the minter. Keys are `name:chain`, so a fresh sequence re-mints chains that live routes in the same payload may already own — e.g. a RESET that keeps `{key: 'a:0-1-0'}` and adds a keyless `{name: 'a'}` mints `a:0-1-0` again: duplicate React keys, one shared descriptor, and `findStateByKey` hits the first match. Subsequent pushes also collide with retained routes. Either reject a payload without `routeKeySeq` or seed with `Math.max(state.routeKeySeq, payload.routeKeySeq ?? 0)`. No test covers `stale: false` RESET without `routeKeySeq`.

packages/expo-router/src/react-navigation/routers/BaseRouter.tsx
```diff
@@ -85,15 +88,28 @@ export const BaseRouter = {
             return null;
           }
 
-          return {
+          const minter = createRouteKeyMinter({
+            key: nextState.key,
```
Ubax: [🏗️ **design**] `RESET` derives the mint chain from the payload's own `key` without validating it: `getChainFromStateKey` does a blind `slice('navigator:'.length)`. A payload with `key: 'root'` (the shape `DrawerRouter.test.tsx` uses) yields an empty parent chain, so minted keys become `name:-0`, and two navigators with short/foreign keys produce identical chains and identical child state keys — a registry-wide collision. `getChainFromStateKey` should assert the `navigator:` prefix, and `RESET` should reject a payload whose `key` differs from the target state's key.

packages/expo-router/src/react-navigation/core/BaseNavigationContainer.tsx
```diff
@@ -288,9 +289,9 @@ export function BaseNavigationContainer({
       onReadyRef.current?.();
       emitter.emit({ type: 'ready' });
     }
-  }, [state, isReady, emitter]);
+  }, [state, registry, isReady, emitter]);
```
Ubax: [🏗️ **design**] `onReady` stayed in a passive effect while the `state` event / `onStateChange` emission moved to a layout effect. Layout effects of a commit run before passive effects of the same commit, and layout effects here do dispatch (`ROUTE_NAMES_CHANGED`, pending-action replay, preloads) — so a second commit's `onStateChange` can be observed before `onReady`. Upstream guarantees `onReady` first, and linking/devtools listeners commonly rely on that. Also the `state` event carries the render state while `onStateChange(hydratedState)` reads `getRootState()`, which can be newer.

packages/expo-router/src/react-navigation/core/BaseNavigationContainer.tsx
```diff
@@ -143,22 +162,20 @@ export function BaseNavigationContainer({
...
-        })
-      );
+      handleAction({
+        ...CommonActions.reset(state),
+        target,
+      });
     }
   });
 
   const getRootState = useLatestCallback(() => {
```
Ubax: [🏗️ **design**] `getRootState()` (and thus `navigation.getState()` / `getStateForKey()`) now returns `pendingStateRef.current ?? completedState` — 'latest dispatched' rather than 'what my render sees'. Under concurrent React this is a tearing source: a component rendered from S0 can read S1 during the same render, so options, prevented-route sets, and `isFocused()` can disagree with the rendered routes for one commit. This is what makes `dispatchSync` work, but it deserves an explicit constraint: render-phase reads should use the render state; only event-handler/effect paths should see the pending ref.

packages/expo-router/src/react-navigation/core/useNavigationHelpers.tsx
```diff
@@ -38,24 +43,41 @@ export function useNavigationHelpers<
...
+    };
 
-      if (!handled) {
-        onUnhandledAction?.(action);
+    const dispatch = (action: Action) => {
+      if (typeof action === 'function') {
+        throw new Error(FUNCTIONAL_DISPATCH_ERROR);
       }
+
+      routingQueue.add({
```
Ubax: [🏗️ **design**] A queued `dispatch` whose originating navigator unmounts before the flush is silently lost: at flush time `getState()` falls back to the last committed slice, the registry check fails, and the throw is downgraded to a `console.warn` by `routingQueue.run` — with a message telling the user to file an expo-router bug. Scenario: `onPress` calls `navigation.goBack()` and the same tick unmounts the containing navigator (tab switch, conditional layout). `useNavigationTreeReducer.handleAction` additionally no-ops silently when `isMountedRef.current` is false. No test covers a queued action whose navigator unmounts before the flush.

packages/expo-router/src/react-navigation/core/BaseNavigationContainer.tsx
```diff
@@ -82,15 +61,33 @@ const getPartialState = (
...
  */
-export function BaseNavigationContainer({
+export function BaseNavigationContainer(props: InternalNavigationContainerProps) {
+  const registry = use(RouterRegistryContext);
+  const independent = useNavigationIndependentTree();
+
+  // TODO(@ubax): investigate if this is really needed
+  if (registry === undefined || independent) {
+    return (
+      <RouterRegistryProvider>
```
Ubax: [🏗️ **design**] The fallback registry creates a split brain: in `fork/NavigationContainer.tsx` the `ImperativeApiEmitter` is a *sibling* of the container, so it reads `RouterRegistryContext` from outside, sees `undefined`, and runs the queue with an empty registry — `NAVIGATE_TO_HREF` intents are then rejected with `Could not generate a valid navigation state for the given path`, while `NAVIGATOR_ACTION` intents still work. The result is a path-dependent failure of `router.push()` that depends on where the provider happens to be mounted; both test fixtures work around it by wrapping in `RouterRegistryProvider` manually. Also, `NavigationIndependentTree` gets its own registry but shares the module-global `routingQueue`, so whichever emitter flushes first can apply another tree's `ACTION` intents.

packages/expo-router/src/fork/NavigationContainer.tsx
```diff
@@ -132,25 +140,42 @@ function NavigationContainerInner(
...
-  if (!isLinkingReady) {
+  if (!isResolved) {
     // This is temporary until we have Suspense for data-fetching
     // Then the fallback will be handled by a parent `Suspense` component
     return <ThemeProvider value={theme}>{fallback}</ThemeProvider>;
   }
 
+  if (initialState === undefined) {
+    throw new Error(
+      'Linking did not produce an initial navigation state. Expo Router always seeds a complete initial state before rendering the navigation container, so this is most likely a bug in expo-router. Please report it at https://github.com/expo/expo/issues.'
```
Ubax: [🏗️ **design**] With `initialState` removed from the public props and linking always enabled, the container cannot be used without expo-router's `routeNode`: when `store?.routeNode` is undefined and no URL matches, `getInitialState` returns `undefined` and this throws. `BaseNavigationContainer` further requires `key`/`routeKeySeq`/`index`/`routeNames` at every level, which the test fixture can only satisfy by walking JSX and minting keys itself. Either keep an internal seeding fallback for unseeded trees or make the entry point explicit and drop the pretense that `NavigationIndependentTree` still works.

packages/expo-router/src/global-state/useNavigationTreeReducer.ts
```diff
@@ -0,0 +1,278 @@
...
+      onDispatchAction?.(action, result.handlerNoop);
+      if (result.treeChanged) {
+        const entry = registry.get(result.target.stateKey);
+        if (
+          action.type !== 'ROUTE_NAMES_CHANGED' &&
+          entry?.shouldPreventRemove?.(result.target.prevSlice, result.target.nextSlice, action)
+        ) {
+          return { handled: true, originStateKey: origin.state.key };
+        }
+        entry?.emitBeforeRemove?.(result.target.prevSlice, result.target.nextSlice, action);
```
Ubax: [🏗️ **design**] `shouldPreventRemove`/`emitBeforeRemove` emit synchronously *before* `commitState`. A `beforeRemove` listener that calls `dispatchSync` re-enters `handleAction`, reduces from the pre-commit state, and commits S1 — then the outer call commits its own result computed from the same base, discarding S1. `useOnPreventRemove.tsx` already carries a TODO about the re-entrancy guard being bypassed. Either commit before emitting, or make re-entrant `dispatchSync` an explicit error.

packages/expo-router/src/global-state/resolveNavigationDestination.ts
```diff
@@ -0,0 +1,336 @@
...
+
+function createDestinationState(
+  targetState: PartialState<NavigationState>,
+  routeNode: RouteNode,
+  withAnchor: boolean,
+  internalParams: InternalExpoRouterParams,
+  parentChain: string
+): NavigationState {
+  const targetRoute = getFocusedRoute(targetState);
+  const routeNames = routeNode.children.map((child) => child.route);
```
Ubax: [🏗️ **design**] `createSeededNavigationState` and `useScreens` sort children with `sortRoutesWithInitial(...)`; this does not. Two consequences: (1) in the fallback branch `createInitialState` picks `routeNames[0]` in raw filesystem order — for children resolving to `['[slug]', 'index']` a non-matching navigation or `router.prefetch` focuses `[slug]` with no params instead of `index`; (2) `useNavigationBuilder` compares committed vs declared (sorted) `routeNames` with `isArrayEqual`, so every navigation into a not-yet-mounted navigator whose order differs triggers an extra `ROUTE_NAMES_CHANGED` dispatch and commit right after mount.

packages/expo-router/src/global-state/getNavigationAction.ts
```diff
@@ -28,91 +33,35 @@ export function getNavigateAction(
...
-
-  href = resolveHrefStringWithSegments(href, store.getRouteInfo(), options);
-  href = applyRedirects(href, store.redirects) ?? undefined;
 
-  // If the href is undefined, it means that the redirect has already been handled by the navigation
-  if (!href) {
-    return;
-  }
+  const href = applyRedirects(
+    resolveHrefStringWithSegments(baseHref, store.getRouteInfo(), options),
```
Ubax: [🏗️ **design**] `store.routeInfo` is only written by `syncStoreState` in `useInsertionEffect` — after the queue drain finishes. `getRootState()` is fresh during the drain, but route info is not. Scenario: `router.push('/a'); router.push('./b')` in one tick resolves `./b` against the pre-push pathname, producing a sibling of the old route instead of a child of `/a`. The same staleness affects `applyRedirects` inputs that depend on the current pathname.

packages/expo-router/src/global-state/useNavigationTreeReducer.ts
```diff
@@ -0,0 +1,278 @@
...
+      const tree = indexNavigationTree(currentState);
+      const origin = resolveOrigin(tree.rootNode, tree.nodes, registry, originKey);
+      if (!origin) {
+        // This also covers synchronous actions dispatched before layout-effect registration.
+        if (typeof operation !== 'function') {
+          const targetKey = typeof operation.target === 'string' ? operation.target : undefined;
+          if (
+            isDeferredInternalAction(operation) ||
+            (originKey !== undefined && tree.nodes.has(originKey)) ||
+            (targetKey !== undefined && tree.nodes.has(targetKey) && !registry.has(targetKey))
```
Ubax: [🏗️ **design**] The defer-until-registered rule only applies when `resolveOrigin` finds nothing. If any other navigator is registered, `origin` exists, `findActionHandler` misses the target, and the action goes to `onUnhandledAction` — throwing in tests, logging in dev. Scenario: a `PRELOAD`/preview action targeted at a navigator that has state in the tree but registers in a later layout effect (freshly mounted nested layout) is dropped, while the identical action during the first commit is queued and replayed. The pending-action rule should not depend on whether an unrelated origin exists.

packages/expo-router/src/global-state/reduceNavigationTree.ts
```diff
@@ -0,0 +1,258 @@
...
+  const attempt = (node: TreeNode): boolean => {
+    const entry = registry.get(node.state.key);
+    if (!entry) {
+      return false;
+    }
+
+    const result = entry.reduce(node.state, action);
+    // Unsupported untargeted actions bubble. An action explicitly addressed to this navigator
+    // stops here as a handled no-op instead of escaping to another navigator.
+    if (result === null && action.target !== node.state.key) {
```
Ubax: [🏗️ **design**] Href navigation is always addressed to a specific navigator (`resolveNavigationDestination` sets `target: current.key`), so a rejected action stops here as a handled no-op and never reaches `onUnhandledAction`. Scenario: the destination is inside `Stack.Protected` with a false guard (missing from committed `routeNames`) — `router.push('/guarded')` does nothing, with no dev warning. Previously the action bubbled and produced the "was not handled by any navigator" message.

packages/expo-router/src/react-navigation/core/useNavigationHelpers.tsx
```diff
@@ -38,24 +43,41 @@ export function useNavigationHelpers<
...
 
-      const handled = onAction(action);
+      handleAction(action);
+    };
 
-      if (!handled) {
-        onUnhandledAction?.(action);
+    const dispatch = (action: Action) => {
+      if (typeof action === 'function') {
+        throw new Error(FUNCTIONAL_DISPATCH_ERROR);
```
Ubax: [🏗️ **design**] Functional dispatch is documented React Navigation API used by third-party navigators; it now throws at runtime with no type-level signal. Relatedly, `NavigatorScreenParams` and the `screen`/`params`/`initial`/`state` payload shapes remain in `core/types.tsx`, but nothing consumes them anymore — `navigation.navigate('(tabs)', { screen: 'feed' })` still type-checks, silently navigates to `(tabs)` only, and stores `screen: 'feed'` as an ordinary param. The types should be removed (or a translation shim kept) before release; both are listed in the changelog but the type surface contradicts it.

packages/expo-router/src/react-navigation/routers/attachRouteState.tsx
```diff
@@ -0,0 +1,51 @@
...
+    return route;
+  }
+
+  if (route.state !== undefined) {
+    if (isEqual(stripStateMarkers(route.state), stripStateMarkers(state))) {
+      return route;
+    }
+  }
+
+  return { ...route, state };
```
Ubax: [🏗️ **design**] `stripStateMarkers` is only used for the equality probe; the marked payload is stored verbatim, so `__internal__routerActionState: true` ends up inside committed state at every generated level. It then leaks to `navigation.getState()`/`useRootNavigationState` consumers, devtools, and the state serialized into `window.history` records — and state read back from the tree is already marked, so re-dispatching it bypasses the trust check the marker exists for. Strip the marker when attaching and keep trust purely as a dispatch-time gate.

packages/expo-router/src/fork/useBrowserHistorySync.ts
```diff
@@ -0,0 +1,312 @@
...
+      const parsedState = getStateFromPathRef.current(path, configRef.current);
+      if (parsedState) {
+        onUnhandledLinking(path);
+        // Expo Router's root navigator contains only the internal slot route.
+        const routeNames = getRootStackRouteNames();
+        if (parsedState.routes.some((route) => !routeNames.includes(route.name))) {
+          return;
+        }
+        const state = store.routeNode
+          ? createSeededRootState(parsedState, store.routeNode)
```
Ubax: [🏗️ **design**] `createSeededState` throws for any route name a navigator does not declare, and the web path only filters *root* route names before seeding — an unexpected nested name throws out of the history listener instead of falling back. A custom `getStateFromPath`, a rewrite/redirect target, or a stale saved record can trigger this (same call exists in `useLinking.native.ts` for deep links). Deep-link and popstate handling should degrade (warn and ignore) rather than throw.

packages/expo-router/src/global-state/routingQueue.ts
```diff
@@ -28,40 +57,67 @@ export const routingQueue = {
...
-    while ((action = events.shift())) {
-      // TODO: Consider warning when ref.current is null — actions are silently dropped
-      if (ref.current) {
-        if (action.type === 'ROUTER_LINK') {
+    let intent: RoutingIntent | undefined;
+    while ((intent = events.shift())) {
+      if (!ref.current) {
+        // TODO: Wait for the root navigator to mount instead of dropping the action.
+        console.warn(
+          'Navigation action was dropped because the navigation container is not mounted.'
```
Ubax: [🏗️ **design**] The queue is reset before the loop, so intents processed while `ref.current` is null are lost, not retried (the TODO acknowledges this). There is a second stall in `imperative-api.tsx`: the flush is gated on `isReady()`, which requires `registry.has(getState().key)` — if the root state key changes (RESET from popstate) without a registry change, the effect deps don't change and the queue waits for the next `add`. Subscribing to container readiness would remove both cases. No test covers a deep link arriving before the container is ready being eventually applied.

packages/expo-router/src/Route.tsx
```diff
@@ -82,15 +83,27 @@ export function useRouteNode(): RouteNode | null {
...
 
 export function findRouteNodeByName(
-  children: RouteNode[] | undefined,
-  routeName: string | undefined
+  node: RouteNode | null | undefined,
+  name: string | undefined
 ): RouteNode | undefined {
-  if (!routeName) {
-    return undefined;
+  return node?.children.find((child) => child.route === name);
```
Ubax: [🏗️ **design**] `findRouteNodeByName` lost the `<name>/index` fallback (re-added only in `getValidInitialRoute`), but the state-driven callers (`createSeededNavigationState`'s `findChildNode`, `resolveNavigationDestination`'s lookups, `findRouteNodeForState`) now require an exact match. That is only correct if every route name in navigation state is exactly a `RouteNode.route` — parsed states from `getStateFromPath` and saved history records are not guaranteed to satisfy this (the seeded-state test `resolves an initial route to its directory index route` relies on the mismatch existing). Where the name doesn't match, the child node is `undefined` and nested seeding silently stops.

packages/expo-router/src/global-state/resolveNavigationDestination.ts
```diff
@@ -0,0 +1,336 @@
...
+  const dynamic = matchDynamicName(targetRoute.name);
+  return (
+    !dynamic ||
+    (targetRoute.params as Record<string, unknown> | undefined)?.[dynamic.name] ===
+      (currentRoute.params as Record<string, unknown> | undefined)?.[dynamic.name]
+  );
+}
+
+function findDestinationNode(routeNode: RouteNode, routeName: string | undefined) {
+  return routeName === INTERNAL_SLOT_NAME || routeNode.route === routeName
```
Ubax: [🏗️ **design**] The self-match short-circuits before the child lookup, so a same-named child is shadowed by its parent. Scenario: `app/foo/_layout.tsx` plus `app/foo/foo.tsx` — resolving `/foo/foo` matches `routeNode.route === 'foo'` and returns the layout node itself, so the child level resolves against the wrong node (wrong `routeNames`, wrong `initialRouteName`). Checking children first (or a parent-or-child flag) avoids this.

packages/expo-router/src/react-navigation/routers/TabRouter.tsx
```diff
@@ -485,7 +456,7 @@ export function TabRouter({
                 const currentId = getId?.({ params: route.params });
                 const nextId = getId?.({ params: action.payload.params });
 
-                const key = currentId === nextId ? route.key : `${route.name}-${nanoid()}`;
+                const key = currentId === nextId ? route.key : minter.mint(route.name);
```
Ubax: [🏗️ **design**] When `NAVIGATE`/`JUMP_TO` re-keys a route (via `getId`), earlier `history` entries still reference the retired key — `changeIndex` only removes/appends the new key. `PRELOAD` got explicit history re-keying for exactly this reason, but this path did not. Scenario with `backBehavior: 'history'`: visit `b?id=1` → `[a, b1]`; jump to `a`; navigate `b?id=2` re-keys to `b2` → history `[b1, a, b2]`; two `GO_BACK`s later `findLastIndex` returns `-1` for `b1` and back navigation dead-ends. Extract the PRELOAD re-key helper and use it on both paths; there is no test for this sequence.

packages/expo-router/src/ui/TabTrigger.tsx
```diff
@@ -199,25 +201,29 @@ export function useTabTrigger(options: TabTriggerProps): UseTabTriggerResult {
...
-              ...config.action.payload,
-              ...options,
-            },
-          });
+          if (!registry) {
+            throw new Error('Router registry is unavailable. This is likely a bug in expo-router.');
+          }
+          const owningState =
+            config.contextKey !== contextKey
+              ? navigation?.getParent(config.contextKey)?.getState()
```
Ubax: [🏗️ **design**] For a foreign trigger this correctly reads the live (pending-aware) state, but the local branch passes the *rendered* `state` from `useNavigatorContext()`. `buildTabAction` → `resolveNavigationDestination` mints speculative keys from that state and bakes the predicted chain into the nested payload. If the rendered state lags the pending state by one dispatch (two dispatches in one handler, or a dispatch during a transition), the real reduce mints a different parent key and the attached child state carries a state key belonging to a different live route — two states with the same key in one tree. Use `navigation.getState()` for both branches so resolution and dispatch read the same snapshot.

packages/expo-router/src/native-tabs/NativeBottomTabsRouter.tsx
```diff
@@ -70,18 +110,17 @@ export function NativeBottomTabsRouter(options: TabRouterOptions) {
               // because it was used to perform zoom transition on another tab
               const params = removeParams(
                 appendInternalExpoRouterParams(route.params, expoParams),
-                [
-                  INTERNAL_EXPO_ROUTER_ZOOM_TRANSITION_SCREEN_ID_PARAM_NAME,
-                  INTERNAL_EXPO_ROUTER_ZOOM_TRANSITION_SOURCE_ID_PARAM_NAME,
-                ]
+                zoomParamNames
               );
+              const state = route.state ? removeZoomParamsFromState(route.state) : undefined;
```
Ubax: [🏗️ **design**] Unlike the identity-preserving helper above, this NAVIGATE branch always allocates a new route for the target and sets an explicit `state: undefined` own-property when the route has no child state. `attachRouteState` compares with `fast-deep-equal`, which is own-key-count sensitive — a route carrying explicit `state: undefined` never equals the resolver's payload route that omits `state`, so the next deep navigation replaces the whole child state instead of preserving it, remounting the tab's nested stack. Return `route` unchanged when nothing changed, and omit `state` rather than setting it to `undefined`.

packages/expo-router/src/react-navigation/core/useNavigationCache.tsx
```diff
@@ -143,7 +162,8 @@ export function useNavigationCache<
       ...helpers,
       // FIXME: too much work to fix the types for now
       ...(emitter.create(route.key) as any),
-      dispatch: (thunk: Thunk) => withStack(() => dispatch(thunk)),
+      dispatch: (action: NavigationAction) => withStack(() => dispatch(action)),
```
Ubax: [💡 suggestion] `withStack` sets `stackRef.current` and clears it in `finally`, but the callback only *enqueues* the action; the actual reduce and `onDispatchAction` emit happen after the queue flush, when the ref is already `undefined`. So `__unsafe_action__.data.stack` is now always empty except for `dispatchSync`, and `useNavigationHelpers.dispatch` doesn't use `withStack` at all. Either capture the stack into the queued intent or delete `withStack`/`stackRef` rather than keeping dead machinery.

packages/expo-router/src/fork/NavigationContainer.tsx
```diff
@@ -132,25 +140,42 @@ function NavigationContainerInner(
...
+    // `onStateChange` owns the store and this must not restore stale state.
+    !refContainer.current?.isReady() &&
+    // Async linking may not have produced its initial state yet.
+    initialState &&
+    // Avoid recalculating route info when the store already has this exact state.
+    initialState !== store.state
+  ) {
+    // TODO(@ubax): remove this when we migrate to global state
+    // Children read route info during this render, so an effect would update the store too late.
+    syncStoreState(initialState);
```
Ubax: [💡 suggestion] This writes module-global store state from the render phase. It is idempotent today, but an interrupted or discarded render (Suspense, `startTransition`) leaves the global store pointing at a state that was never committed, visible to every later `store.getRouteInfo()` reader; the `!refContainer.current?.isReady()` guard is also a render-phase read of a mutable ref. Since `useNavigationTreeReducer` already syncs from `useInsertionEffect`, consider seeding the store there instead.

packages/expo-router/src/fork/useBrowserHistorySync.ts
```diff
@@ -0,0 +1,312 @@
...
+    const onStateChange = async () => {
+      const navigation = ref.current;
+
+      if (!navigation) {
+        return;
+      }
+
+      const previousState = previousStateRef.current;
+      const rootState = navigation.getRootState();
+      const state = store.state as NavigationState;
```
Ubax: [💡 suggestion] `findMatchingState` compares the previous *root* state (`previousStateRef.current = rootState` below) with the *store* state. They are equal today only because `syncStoreState` runs on every commit; if they ever diverge, `findMatchingState` returns `[undefined, undefined]` and the code silently falls through to `history.replace`, losing a history entry. Use one source for both sides of the comparison.

packages/expo-router/src/fork/useBrowserHistorySync.ts
```diff
@@ -0,0 +1,312 @@
...
+      const pendingOperationIndex = pendingHistoryOperationsRef.current.findLastIndex(
+        (operation) => operation.path === path
+      );
+      if (pendingOperationIndex !== -1) {
+        pendingOperation = pendingHistoryOperationsRef.current[pendingOperationIndex];
+        pendingHistoryOperationsRef.current.splice(0, pendingOperationIndex + 1);
+      }
+      if (!pendingOperation) {
+        // Do not let a failed, redirected, or otherwise unmatched operation affect later navigation.
+        pendingHistoryOperationsRef.current.shift();
```
Ubax: [💡 suggestion] A queued popstate intent that produces no state change (identical state, rejected action, redirect) leaves its operation in the list; cleanup relies on a later, unrelated state change to shift it off. With two popstates in flight, the `findLastIndex` path-match plus this blind `shift` can consume the wrong record. Correlating operations by intent identity instead of by path would remove the guesswork.

packages/expo-router/src/navigationParams.ts
```diff
@@ -38,29 +38,45 @@ const internalExpoRouterParamNames = [
...
 export type InternalExpoRouterParams = Partial<Record<InternalExpoRouterParamName, unknown>>;
 
+const paramsObjectsWarnedForScreen = new WeakSet<object>();
+const paramsObjectsWarnedForNestedParams = new WeakSet<object>();
+
+export function warnIfScreenParam(params: object | undefined) {
+  if (
+    process.env.NODE_ENV !== 'production' &&
+    params &&
+    'screen' in params &&
```
Ubax: [💡 suggestion] `getRouteInfoFromState` calls this per level, so any route with a legitimate `screen` search param or a `[screen]` dynamic segment logs the migration warning on every route-info computation — e.g. `router.push('/settings?screen=account')`. `warnIfNestedParams` has the same shape of problem for object-valued params. Consider warning only when the param came from a navigation payload, not from parsed URL params.

packages/expo-router/src/react-navigation/routers/types.tsx
```diff
@@ -216,3 +206,18 @@ export type Router<
...
 };
+
+/**
+ * The result of reducing a navigation action.
+ */
+export type RouterActionResult<State extends NavigationState> = {
+  /**
+   * The navigation state produced by the action.
+   */
+  state: State | PartialState<State>;
```
Ubax: [💡 suggestion] Every producer now returns a complete state (`BaseRouter` throws otherwise), yet `RouterActionResult` keeps the partial union. This forces defensive `stale !== false` checks in `resolveNavigationDestination` and a cast-plus-comment in `TabRouter`, and makes `affectedRouteKey: string | undefined` look load-bearing when it is only `undefined` for the empty-tab case. Tightening to `State` would delete several casts.

packages/expo-router/src/ui/TabRouter.tsx
```diff
@@ -66,15 +64,24 @@ export function ExpoTabRouter(options: ExpoTabRouterOptions) {
...
-      } else if (route.state !== undefined) {
-        // TODO(@ubax): Remove this branch together with nested trigger href support. Refocusing
-        // a tab that hosts a navigator must not re-apply the trigger's nested payload
-        // (`params.screen`), which would reset the preserved child state.
-        return rnTabRouter.getStateForRouteFocus(state, route.key);
-      } else {
-        return rnTabRouter.getStateForAction(state, action, routerConfigOptions);
       }
+
+      if (!isSwitching && route.state !== undefined) {
```
Ubax: [💡 suggestion] Re-selecting the focused tab that hosts a navigator returns the state unchanged, so `JUMP_TO { name: focusedTab, params }` drops `params` while still reporting handled (no `onUnhandledAction`, no `SET_PARAMS`) — leaf tabs behave differently. Also `attachRouteState` is used as an equality probe here and then run again by the delegated `rnTabRouter`, so an unmarked state-bearing action logs the internal-marker dev warning twice.

packages/expo-router/src/react-navigation/routers/DrawerRouter.tsx
```diff
@@ -188,37 +162,55 @@ export function DrawerRouter({
...
         case 'NAVIGATE': {
-          const result = router.getStateForAction(state, action, options);
-
-          if (result != null && result.index !== state.index) {
-            return closeDrawer(result as DrawerNavigationState<ParamListBase>);
+          const actionResult = router.getStateForAction(state, action, options);
+
+          if (actionResult !== null) {
+            const nextState = actionResult.state;
+            if (nextState.index === state.index) {
```
Ubax: [💡 suggestion] `REPLACE` of the focused route keeps `index` constant, so the drawer stays open even though the focused route changed. Comparing the focused route *key* instead of the index expresses the intent ("navigating closes the drawer") correctly.

packages/expo-router/src/ui/TabRouter.tsx
```diff
@@ -46,15 +50,9 @@ export function ExpoTabRouter(options: ExpoTabRouterOptions) {
...
-      // We should reset if this is the first time visiting the route.
-      let shouldReset =
-        !(state.history == null ? state.routes : state.history).some(
-          (item) => item.key === route.key
-        ) && !route.state;
-
-      if (!shouldReset && 'resetOnFocus' in action.payload && action.payload.resetOnFocus) {
-        shouldReset = state.routes[state.index ?? 0]!.key !== route.key;
-      }
+      const isSwitching = state.routes[state.index ?? 0]!.key !== route.key;
```
Ubax: [🐧 nit] `index` is a required `number`, so `?? 0` is dead — and `index === -1` (which `ROUTE_NAMES_CHANGED`/`getStateForDeclaredRoutes` can produce) indexes out of bounds and throws on `.key` instead of delegating. Use `state.routes[state.index]?.key !== route.key`.

packages/expo-router/CHANGELOG.md
```diff
@@ -64,6 +72,8 @@
 ### 💡 Others
 
 - Remove the ignored `linking.enabled` option. ([#49103](https://github.com/expo/expo/pull/49103) by [@Ubax](https://github.com/Ubax))
+- Seed complete navigation state from the initial deep link. ([#TODO](https://github.com/expo/expo/pull/TODO) by [@Ubax](https://github.com/Ubax))
```
Ubax: [🐧 nit] This entry has a `[#TODO]` placeholder, and the new breaking-change bullets at the top of the file have no PR links at all. They should all point at #49110 (or the respective stack PRs).

packages/expo-router/src/fork/useBrowserHistorySync.ts
Ubax: Add todo to solve
  - A stale record: the user navigates back/forward to a URL that was valid before a route file was renamed, moved into a group, or removed during dev/HMR or
    between deploys.

packages/expo-router/src/fork/useBrowserHistorySync.ts
Ubax: The suggested fix. The identity already exists: each operation has a unique id, and it travels on the intent as intent.metadata.history. What is missing is
  carrying that identity through to the consumer instead of reconstructing it from the path:

  - When a commit applies intents, let onStateChange learn which operation ids were included in that commit (for example, the reducer or queue records the ids
    applied per commit, or an onSettled counterpart to onDispatch fires after commit/rejection). Then removal is exact: remove those ids.
  - Remove a rejected or no-op intent's record at rejection time, via the same callback, instead of waiting for an unrelated state change to shift it off.

packages/expo-router/src/react-navigation/routers/BaseRouter.tsx
Ubax: Validate the keys, add TODO to optimize and throw if keys are invalid

packages/expo-router/src/react-navigation/core/useNavigationHelpers.tsx
Ubax: Adjust the types so they are correct

packages/expo-router/src/react-navigation/core/useNavigationHelpers.tsx
Ubax: Add test case for that

packages/expo-router/src/react-navigation/core/useNavigationCache.tsx
Ubax: Remove with TODO to add it later. Document what was removed in the comment and link current PR

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
Ubax: The navigators should only be allowed inside `_layout`. Throw if they are used in screens

packages/expo-router/src/react-navigation/core/BaseNavigationContainer.tsx
Ubax: Check if this is a relevant case outside of tests. If not then remove the whole `BaseNavigationContainerInner` and make it `BaseNavigationContainer`.

The `NavigationIndependentTree` is not relevant. It will be removed

packages/expo-router/src/react-navigation/core/BaseNavigationContainer.tsx
Ubax: Remove this. The state is only updated in the next render after dispatchSync

PR #49117, opened 2026-08-19
[expo-notifications][Android] Clear the consumed FCM extra so a notification response isn't delivered again on later launches

sang88ho:
# Why

On Android a notification response is delivered **again on later launches, without the user
tapping anything**. Open the app once by tapping a notification, go Home, come back to the app —
the same response arrives a second time, so the app re-opens content the user already handled.

`ExpoNotificationLifecycleListener.onCreate` consumes the response from `activity.intent.extras`
but never clears it. That intent stays on the activity and becomes the task's base intent, so
every later recreation or base-intent relaunch hands the same `google.message_id` back to
`onCreate`, which forwards it to `NotificationManager.onNotificationResponseFromExtras` again.

The same file already treats extras as something to consume — but only on one branch:

https://github.com/expo/expo/blob/main/packages/expo-notifications/android/src/main/java/expo/modules/notifications/service/delegates/ExpoNotificationLifecycleListener.kt#L46-L47

```kotlin
// onNewIntent
intent.removeExtra(NOTIFICATION_RESPONSE_KEY)
intent.removeExtra(TEXT_INPUT_NOTIFICATION_RESPONSE_KEY)
```

Neither callback ever clears `GOOGLE_MESSAGE_ID_KEY` on the FCM path. The asymmetry reads like an
oversight rather than a decision.

Symptoms depend on what the app does with the response. Apps that navigate on it navigate again;
apps that de-duplicate responses in JS get the opposite result — the launch is swallowed and the
user lands on a default route instead of where they were. `useLastNotificationResponse` /
`getLastNotificationResponseAsync` report a stale response for the same reason.

# How

Clear the extra once it has been handed off, the way `onNewIntent` already does for the other two
keys. Three lines, plus lifting the nullable `activity.intent` into a local so the same instance is
used for both reads.

This does not affect the response being delivered: `Intent.getExtras()` returns a **copy** of the
bundle
([AOSP](https://android.googlesource.com/platform/frameworks/base/+/refs/heads/main/core/java/android/content/Intent.java) —
`return (mExtras != null) ? new Bundle(mExtras) : null;`), so the bundle passed to
`NotificationManager` is untouched by the removal.

# Test Plan

Physical Galaxy S24+ (SM-S926N, Android 16), release build, New Architecture, SDK 56
(`expo-notifications` 56.0.23). `MainActivity` instrumented to log every intent reaching
`onCreate` / `onNewIntent` together with its `google.message_id`.

**Reproducing the report.** App force-stopped and its task removed from recents → send an FCM push
→ tap the notification. The app cold-starts and the response is delivered once, correctly:

```
entry=onCreate flags=0x14000000 notif=true
[JS] processResponse id=0:1787121890647522%aeedc062aeedc062 url=/breaking-news?newsId=105454
```

Then Home, and back into the app **without touching any notification** — `onCreate` runs again
carrying the same `google.message_id`, and the response is delivered a second time.

**The duplicate this diff removes.** Our app handles the response in `MainActivity` first and then
calls `super`, and because the extras were still on the intent, `expo-notifications` picked the
same intent up again — the JS layer received one tap as two responses and navigated twice, landing
on the wrong screen. Removing these keys from the intent before passing it on fixed it, measured as
`processResponse` going from 2 occurrences to 1 for a single tap.

**What this diff does not fix.** Clearing in-process does not reach the copy of the base intent
that the system keeps, so a task restored from recents still arrives with the extras present. We
handle that separately by remembering the `google.message_id`s already delivered and dropping
repeats. That adds persisted state, which felt like more than this fix should carry — I would
rather ask which direction you would accept before proposing it. `FLAG_ACTIVITY_LAUNCHED_FROM_HISTORY`
is not a sufficient discriminator on its own: we measured `flags=0x14000000` both for a genuine
cold notification tap and for a later launch that replayed the response.

I have not built the package from source for this diff — it is three calls to an API this file
already uses twice — but I am happy to if you would like that before merging.

# Checklist


vonovak:
@expo-bot verify

PR #49125, opened 2026-08-19
[android][ios][sqlite] Delete journal, WAL and SHM sidecar files when deleting a database

### LINKED ISSUE #43441: expo-sqlite: deleteDatabaseSync does not remove WAL and SHM sidecar files
## Summary

When SQLite operates in WAL (Write-Ahead Logging) mode, it creates two sidecar files alongside the main database:
- `<database>-wal` (Write-Ahead Log)
- `<database>-shm` (Shared Memory index)

`deleteDatabaseSync` / `deleteDatabaseAsync` only remove the main `.db` file, leaving the `-wal` and `-shm` files on disk. This diverges from how SQLite tooling and platform APIs handle database deletion, and can cause stale WAL replay if a new database is later created with the same name.

## Environment

```
expo-env-info 2.0.8 environment info:
  System:
    OS: macOS 15.6.1
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.13.1
    npm: 11.1.0
  SDKs:
    iOS SDK: Platforms: iOS 18.5
  IDEs:
    Android Studio: 2025.1
    Xcode: 16.4/16F6
```

**expo-sqlite version:** ~16.0.10 (SDK 54)

https://github.com/jmalmo/expo-sqlite-wal-repro

The reproduction app:
1. Opens a database and enables WAL mode via `PRAGMA journal_mode = WAL`
2. Creates a table, inserts data, and closes the database
3. Writes fake `-wal` and `-shm` files to simulate crash leftovers (a clean close checkpoints and removes them, but a crash or abnormal exit leaves them on disk)
4. Calls `deleteDatabaseAsync()`
5. Checks whether `-wal` and `-shm` files were removed — **they are not**

## Steps to reproduce

```js
import * as SQLite from 'expo-sqlite';
import * as FileSystem from 'expo-file-system/legacy';

const db = await SQLite.openDatabaseAsync('test.db');
await db.execAsync('PRAGMA journal_mode = WAL');
await db.execAsync('CREATE TABLE t (id INTEGER PRIMARY KEY, v TEXT)');
await db.execAsync("INSERT INTO t (v) VALUES ('hello')");
await db.closeAsync();

const dbPath = FileSystem.documentDirectory + 'SQLite/test.db';

// Simulate crash leftovers (a clean close removes these,
// but a crash or abnormal exit leaves them on disk)
await FileSystem.writeAsStringAsync(dbPath + '-wal', 'stale WAL');
await FileSystem.writeAsStringAsync(dbPath + '-shm', 'stale SHM');

await SQLite.deleteDatabaseAsync('test.db');

// Main db is gone, but sidecar files remain
console.log(await FileSystem.getInfoAsync(dbPath));          // { exists: false }
console.log(await FileSystem.getInfoAsync(dbPath + '-wal')); // { exists: true } BUG
console.log(await FileSystem.getInfoAsync(dbPath + '-shm')); // { exists: true } BUG
```

## Expected behavior

`deleteDatabaseSync` / `deleteDatabaseAsync` should remove the `-wal` and `-shm` sidecar files alongside the main database file.

This is consistent with:
- Android's built-in [`SQLiteDatabase.deleteDatabase()`](https://developer.android.com/reference/android/database/sqlite/SQLiteDatabase#deleteDatabase(java.io.File)), which explicitly removes `-wal`, `-shm`, and `-journal` files
- The `sqlite3` CLI, which cleans up all associated files on database deletion
- General SQLite best practices from the [SQLite documentation](https://www.sqlite.org/tempfiles.html)

## Actual behavior

Only the main database file is deleted. The `-wal` and `-shm` files remain on disk.

### Why this matters

1. **Stale WAL replay**: If a new database is later created with the same name, SQLite may replay the orphaned WAL against the fresh database, leading to corruption or unexpected state.
2. **Disk space leaks**: WAL files can grow substantially and are never cleaned up unless the caller manually deletes them.
3. **Incomplete reset**: Any workflow that uses `deleteDatabaseSync` to reset local state (e.g. cache invalidation, schema migration, user logout) will leave stale sidecar files behind.

## Root cause

The native implementations on both platforms only delete the main file:

**iOS** (`SQLiteModule.swift`):
```swift
try FileManager.default.removeItem(atPath: path)
// No cleanup of path + "-wal" or path + "-shm"
```

**Android** (`SQLiteModule.kt`):
```kotlin
dbFile.delete()
// No cleanup of File(dbFile.path + "-wal") or File(dbFile.path + "-shm")
```

## Fix

I've submitted a PR with the fix: adds `-wal` and `-shm` cleanup after main file deletion on both iOS and Android, plus an E2E test.

sbaiahmed1:
# Why

Closes #43441.

`deleteDatabaseAsync` / `deleteDatabaseSync` only remove the main database file. When SQLite runs in WAL mode, the `<database>-wal` and `<database>-shm` sidecar files are left on disk (and `-journal` in rollback mode after a crash). Stale sidecars waste storage and can cause stale WAL replay if a new database is later created with the same name.

Android's own [`SQLiteDatabase.deleteDatabase()`](https://developer.android.com/reference/android/database/sqlite/SQLiteDatabase#deleteDatabase(java.io.File)) explicitly removes the `-journal`, `-wal` and `-shm` files, and the issue's minimal repro demonstrates the divergence.

# How

- Extracted the file-deletion logic into shared, unit-testable helpers: `deleteDatabaseFiles()` in `SQLiteHelpers.kt` (Android) and a new `DatabaseFileUtils.swift` (iOS/macOS/tvOS).
- After deleting the main database file, the helpers best-effort delete the `-journal`, `-wal` and `-shm` sidecars, mirroring `SQLiteDatabase.deleteDatabase()` semantics. Error contracts are unchanged: missing main file still throws `DatabaseNotFoundException`, failed main-file deletion still throws `DeleteDatabaseFileException`.
- Added the package's first native unit tests on both platforms, including a new iOS `test_spec` for `ExpoSQLite` (with `-lc++` in `OTHER_LDFLAGS`, required to link the static test bundle).

# Test Plan

- Android: `et native-unit-tests -p android -t local --packages expo-sqlite` — new `SQLiteHelpersTest` written red-first: the sidecar test failed against the original logic and passes with the fix (4/4 green).
- iOS: `et native-unit-tests -p ios --packages expo-sqlite` — new `DatabaseFileUtilsTests`, 4/4 green on the iPhone 17 Pro simulator.
- `et check-packages expo-sqlite` passes (build, typecheck, test, lint, format).

# Checklist

### DIFF
--- a/packages/expo-sqlite/ios/ExpoSQLite.podspec
+++ b/packages/expo-sqlite/ios/ExpoSQLite.podspec
@@ -75,6 +75,7 @@ Pod::Spec.new do |s|
   s.test_spec 'Tests' do |test_spec|
     test_spec.source_files = 'Tests'
     test_spec.pod_target_xcconfig = {
+      # The test bundle links the static ExpoModulesCore dependency chain, which contains C++.
       'OTHER_LDFLAGS' => '-lc++'
     }
   end
--- a/packages/expo-sqlite/ios/ExpoSQLite.podspec
+++ b/packages/expo-sqlite/ios/ExpoSQLite.podspec
@@ -70,6 +70,14 @@ Pod::Spec.new do |s|
     'OTHER_SWIFT_FLAGS' => '$(inherited) ' + swift_flags,
   }
   s.source_files = "**/*.{c,h,m,swift}"
+  s.exclude_files = 'Tests'
+
+  s.test_spec 'Tests' do |test_spec|
+    test_spec.source_files = 'Tests'
+    test_spec.pod_target_xcconfig = {
+      'OTHER_LDFLAGS' => '-lc++'
+    }
+  end
 
   vendored_frameworks = []
   if podfile_properties['expo.sqlite.withSQLiteVecExtension'] == 'true'
--- a/packages/expo-sqlite/ios/Tests/DatabaseFileUtilsTests.swift
+++ b/packages/expo-sqlite/ios/Tests/DatabaseFileUtilsTests.swift
@@ -1,18 +1,19 @@
 // Copyright 2015-present 650 Industries. All rights reserved.
 
-import XCTest
+import Testing
 
 @testable import ExpoSQLite
 
-class DatabaseFileUtilsTests: XCTestCase {
-  private var tempDir: URL!
+@Suite("DatabaseFileUtils")
+final class DatabaseFileUtilsTests {
+  private let tempDir: URL
 
-  override func setUpWithError() throws {
+  init() throws {
     tempDir = FileManager.default.temporaryDirectory.appendingPathComponent(UUID().uuidString)
     try FileManager.default.createDirectory(at: tempDir, withIntermediateDirectories: true)
   }
 
-  override func tearDownWithError() throws {
+  deinit {
     try? FileManager.default.removeItem(at: tempDir)
   }
 
@@ -22,15 +23,17 @@ class DatabaseFileUtilsTests: XCTestCase {
     return path
   }
 
-  func testDeletesTheMainDatabaseFile() throws {
+  @Test
+  func `deletes the main database file`() throws {
     let dbPath = createFile("test.db")
 
     try DatabaseFileUtils.deleteDatabaseFiles(atPath: dbPath)
 
-    XCTAssertFalse(FileManager.default.fileExists(atPath: dbPath))
+    #expect(!FileManager.default.fileExists(atPath: dbPath))
   }
 
-  func testDeletesJournalWalAndShmSidecarFilesAlongWithTheDatabase() throws {
+  @Test
+  func `deletes journal wal and shm sidecar files along with the database`() throws {
     let dbPath = createFile("test.db")
     let journalPath = createFile("test.db-journal")
     let walPath = createFile("test.db-wal")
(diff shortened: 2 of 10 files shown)


packages/expo-sqlite/ios/Tests/DatabaseFileUtilsTests.swift
```diff
@@ -0,0 +1,66 @@
+// Copyright 2015-present 650 Industries. All rights reserved.
+
+import XCTest
```
Kudo: we are migrating to use swift testing framework than xctest. could you try to migrate this using swift testing?

e.g.
https://github.com/expo/expo/blob/main/packages/expo-modules-core/ios/Tests/ConstantsTests.swift

packages/expo-sqlite/ios/ExpoSQLite.podspec
```diff
@@ -70,6 +70,14 @@ Pod::Spec.new do |s|
     'OTHER_SWIFT_FLAGS' => '$(inherited) ' + swift_flags,
   }
   s.source_files = "**/*.{c,h,m,swift}"
+  s.exclude_files = 'Tests'
+
+  s.test_spec 'Tests' do |test_spec|
+    test_spec.source_files = 'Tests'
+    test_spec.pod_target_xcconfig = {
+      'OTHER_LDFLAGS' => '-lc++'
+    }
```
Kudo, line 80:
```suggestion
```

could you help me to verify if that's necessary (after migrating to swift testing)

Kudo (changes requested):
nice pr and tests. just leaves nit some comment for swift testing. please re-request review once it's ready for next round review

packages/expo-sqlite/ios/Tests/DatabaseFileUtilsTests.swift
sbaiahmed1: Done — migrated to Swift Testing following the ConstantsTests example (`@Suite` + `@Test` + `#expect`, raw-identifier test names).

packages/expo-sqlite/ios/ExpoSQLite.podspec
sbaiahmed1: Verified — it's still necessary after the Swift Testing migration. Without `-lc++` the test bundle fails to link with undefined C++ runtime symbols (`___cxa_throw`, `___gxx_personality_v0`, `operator new`, …) coming from the static ExpoModulesCore dependency chain, independent of the test framework. I kept the flag and added a short comment in the podspec explaining why it's there.

Kudo: great. thanks double checking that for me.

Kudo (approved):
all looks good. let's wait ci passed

sbaiahmed1:
> all looks good. let's wait ci passed

Issue in the macos build, is it my changes or i just have to do a rebase if it is fixed on main? 
I can see it also fails on https://github.com/expo/expo/pull/49127

Kudo:
macos-build failure feels unrelated. going to merge this and i'll follow up to fix it

sbaiahmed1:
> macos-build failure feels unrelated. going to merge this and i'll follow up to fix it

Thank you so much! happy to contribute

Issue #23924, opened 2023-08-11
[expo-secure-store][iOS] - `User interaction is not allowed` after upgrading to SDK 49


AronBe:
### Minimal reproducible example

sorry, I haven't been able to reproduce it in the dev env

### Summary

Sentry is reporting quite many errors regarding setting and getting values via `expo-secure-store`. I have started notice those since upgrade to SDK 49 when going from `12.1.1` to `12.3.1`. Seems like Keychain does not allow interaction, but not sure why. I haven't changed any privileges  via secure store keychain config and calling it in the app as simply as  `await SecureStore.setItemAsync(REFRESH_TOKEN_KEY, token)`

```
Calling the 'setValueWithKeyAsync' function has failed → Caused by: User interaction is not allowed.

Calling the 'getValueWithKeyAsync' function has failed
→ Caused by: User interaction is not allowed.
```

I wonder if anyone else is experiencing this

### Environment

expo-env-info 1.0.2 environment info:
    System:
      OS: macOS 13.4.1
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 16.18.0 - ~/.nvm/versions/node/v16.18.0/bin/node
      Yarn: 1.22.19 - /usr/local/bin/yarn
      npm: 8.19.2 - ~/.nvm/versions/node/v16.18.0/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    Managers:
      CocoaPods: 1.12.0 - /opt/homebrew/bin/pod
    SDKs:
      iOS SDK:
        Platforms: DriverKit 22.4, iOS 16.4, macOS 13.3, tvOS 16.4, watchOS 9.4
      Android SDK:
        API Levels: 29, 30, 31, 32, 33
        Build Tools: 28.0.3, 29.0.2, 30.0.2, 30.0.3, 31.0.0, 32.0.0, 33.0.0
        System Images: android-30 | Google Play Intel x86 Atom, android-31 | Intel x86 Atom_64, android-31 | Google APIs Intel x86 Atom_64, android-32 | Desktop ARM 64 v8a, android-32 | Google APIs Intel x86 Atom_64, android-32 | Google Play ARM 64 v8a, android-33-ext5 | Google Play ARM 64 v8a, android-33 | Google APIs ARM 64 v8a, android-33 | Google Play ARM 64 v8a
    IDEs:
      Android Studio: 2021.3 AI-213.7172.25.2113.9123335
[...]

PR #49128, opened 2026-08-19
[secure-store][iOS] apply keychainAccessible when updating an existing item

JoRo-Code:
# Why

On iOS, `keychainAccessible` is silently ignored on every write to a key that already exists. The item keeps the accessibility class it was first created with, for the life of the install.

Because the default is `kSecAttrAccessibleWhenUnlocked`, an app that later decides its data must survive a locked screen has no way to get there through the public API. Calling `setItem(key, value, { keychainAccessible: AFTER_FIRST_UNLOCK })` looks like it worked — it returns normally and the value really is updated — but reads behind a locked screen keep failing with `errSecInteractionNotAllowed` ("User interaction is not allowed."). There is no error and no warning to suggest the option did not take.

We hit this in an app that records audio in the background. With `UIBackgroundModes: ["audio"]` the JS thread keeps running while the phone is locked in someone's pocket, so a timer read the keychain, got `errSecInteractionNotAllowed`, and the throw took the app down mid-recording. Setting `keychainAccessible` was the obvious fix and it changed nothing, which took a while to explain.

Closed issue #23924 reports the same `User interaction is not allowed` symptom; I could not find an existing issue for this cause.

# How

`set(value:with:options:)` tries `SecItemAdd` first and puts `kSecAttrAccessible` in that dictionary, so a **new** key gets the requested class. An existing key returns `errSecDuplicateItem` and falls through to `update(value:with:options:)`, whose `SecItemUpdate` attributes dictionary contains `kSecValueData` and nothing else:

```swift
let updateDictionary = [kSecValueData as String: valueData]
```

`query(with:options:requireAuthentication:)` does not name `kSecAttrAccessible` either, so the item is found by service/account, the value is replaced, and the accessibility attribute is left exactly as it was.

This keeps `keychainAccessible` optional in the native options record so omission remains distinguishable from the default. New items still default to `kSecAttrAccessibleWhenUnlocked`; updates include the attribute only when the caller supplied it:

```swift
let updateDictionary = if !options.requireAuthentication, options.keychainAccessible != nil {
  [
    kSecValueData: valueData,
    kSecAttrAccessible: attributeWith(options: options)
  ] as CFDictionary
} else {
  [kSecValueData: valueData] as CFDictionary
}
```

Only the unauthenticated path applies it, mirroring `set`, where an authenticated item carries its accessibility inside `kSecAttrAccessControl` rather than `kSecAttrAccessible`. Updating that in place would mean rebuilding a `SecAccessControl` and has biometric-enrollment implications, so I left it alone — happy to follow up if you would like it covered too.

Behaviour is unchanged for new keys, for reads, and for any call that does not pass `keychainAccessible`: new items still receive the documented `kSecAttrAccessibleWhenUnlocked` default, while existing items retain the accessibility class with which they were stored.

# Test Plan

I want to be straightforward about what is and is not verified here: the diagnosis comes from reading `SecureStoreModule.swift`, and I have not yet run a build of this branch on a device. I am opening it because the cause looks unambiguous in the source and the reproduction is cheap for anyone with the module already set up. Happy to come back with device output before you merge — say the word and I will.

The reproduction, on a physical device (a simulator will not do — it needs a real lock screen):

```js
// create the item the default way, then try to upgrade it in place
await SecureStore.setItemAsync('probe', 'hello');
await SecureStore.setItemAsync('probe', 'hello', {
  keychainAccessible: SecureStore.AFTER_FIRST_UNLOCK,
});
```

Then read `probe` from something that keeps running once the phone is locked — an app with an active audio session and `UIBackgroundModes: ["audio"]` is the easiest, which is how we ran into it:

- **before this change** — `getValueWithKeySync` throws `KeyChainException: User interaction is not allowed.`, because the item is still `kSecAttrAccessibleWhenUnlocked` despite the second write.
- **after this change** — the read returns `hello`.

`deleteItemAsync` followed by a fresh `setItemAsync` with the same option succeeds on both builds, which is the workaround this is meant to make unnecessary, and is also the evidence that the class itself is applied correctly on an add.

Worth a reviewer's eye specifically: whether a plain `setItemAsync(key, value)` on an existing key is genuinely unaffected. It should be — the omitted option remains `nil`, so the update dictionary contains only `kSecValueData` and leaves the existing accessibility unchanged — but that is reasoning, not a measurement.

# Checklist

### DIFF
--- a/packages/expo-secure-store/CHANGELOG.md
+++ b/packages/expo-secure-store/CHANGELOG.md
@@ -11,6 +11,7 @@
 ### 🐛 Bug fixes
 
 - [iOS] Reject `deleteItemAsync` when the keychain refuses the delete, instead of resolving as if the item was removed.
+- [iOS] Apply `keychainAccessible` when overwriting an existing item, instead of silently keeping the accessibility it was first stored with.
 
 ### 💡 Others
 
--- a/packages/expo-secure-store/CHANGELOG.md
+++ b/packages/expo-secure-store/CHANGELOG.md
@@ -11,7 +11,7 @@
 ### 🐛 Bug fixes
 
 - [iOS] Reject `deleteItemAsync` when the keychain refuses the delete, instead of resolving as if the item was removed.
-- [iOS] Apply `keychainAccessible` when overwriting an existing item, instead of silently keeping the accessibility it was first stored with.
+- [iOS] Apply `keychainAccessible` when overwriting an existing item, instead of silently keeping the accessibility it was first stored with. ([#49128](https://github.com/expo/expo/pull/49128) by [@JoRo-Code](https://github.com/JoRo-Code) and [@behenate](https://github.com/behenate))
 
 ### 💡 Others
 
--- a/packages/expo-secure-store/ios/SecureStoreModule.swift
+++ b/packages/expo-secure-store/ios/SecureStoreModule.swift
@@ -109,12 +109,7 @@ public final class SecureStoreModule: Module {
         throw MissingPlistKeyException()
       }
 
-      var error: Unmanaged<CFError>? = nil
-      guard let accessOptions = SecAccessControlCreateWithFlags(kCFAllocatorDefault, accessibility, .biometryCurrentSet, &error) else {
-        let errorCode = error.map { CFErrorGetCode($0.takeRetainedValue()) }
-        throw SecAccessControlError(errorCode)
-      }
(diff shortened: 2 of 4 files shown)


behenate:
@expo-bot review

behenate (approved):
Looks good. I ran some local repros. Even though I couldn't reproduce the error from JS I did verify that the security attributes are not overwritten on save. 

I took the liberty of modyfing this PR to include a fix for the authenticated path

PR #49130, opened 2026-08-19
[android][updates] Register the embedded update in a single transaction

gwdp:
# Why

Loader registered an update in two separate writes. `insertUpdate` wrote the row, and only after all assets were processed did `insertAssets` and `markUpdateFinished` set the launch asset and the final status

For an embedded update this gap is dangerous because `EmbeddedUpdate` stamps the entity with `EMBEDDED` and `loadLaunchableUpdatesForScope` treats `EMBEDDED` as launchable; So a interrupted registration left a row that advertised itself as launchable, but not really (`launch_asset_id` still NULL).  DatabaseLauncher then threw on it, and the device emergency-launched on the bundle embedded in the binary.

That state was also permanent 😢 The broken row is returned as the launchable update, so `shouldLoadNewUpdate` returns false and the loader never re-runs. The reaper skips `EMBEDDED` rows, and `DatabaseIntegrityCheck` has no caller on Android, so nothing repaired it. The device kept emergency-launching until a newer remote update arrived. (I'm still reasoning here if we should have an integrity check/and-or a repair strategy, but at least stop the bleeding here)

Remote updates were never exposed to this, because their entity defaults to `PENDING`, which is not in the launchable set.

# How

Hold the update row back until the closing writes, and put those writes in one transaction (that simple)

```kotlin
insertUpdateEntityOnFinish = newUpdateEntity.status == UpdateStatus.EMBEDDED
if (!insertUpdateEntityOnFinish) {
  database.updateDao().insertUpdate(updateEntity!!)
}
```

An interrupted embedded registration now leaves no row at all, which is the same state as a fresh install, so the next launch registers normally.

Also updating the scary error message, we understand what causes it and what the effects are; Now it should become impossible to reach it, but yeah, we never say impossible if there is a code path leading to it :) 

# Test Plan
(I'm not too familiar with this repo guidelines, so I tried to go through and do my part, sorry if missing anyuthing)

- Ran `:expo-updates:testDebugUnitTest` + `:expo-updates:connectedDebugAndroidTest`
- `:expo-updates:spotlessCheck` clean.
- Update existing test to confirm new behaviour under failure; added a new one which makes `markUpdateFinished` throw after `insertAssets` has run and asserts no rows survive

# Checklist


gwdp:
related #48733 #46741

brentvatne:
@expo-bot review

gwdp:
@expo-bot review

gwdp:
Figured out it would be good to share the repro and how I got it tested
```
adb root
PKG=dev.expo.healcheckapp
INT=/data/data/$PKG/files/.expo-internal
DB=/data/data/$PKG/databases/updates.db

# make asset copy fails mid-registration (updates dir unwritable being one way I could reliably fail mid-process, dont think it's the real cause tbh)
adb shell pm clear $PKG
adb shell "mkdir -p $INT; chown 0:0 $INT; chmod 500 $INT"

adb shell am start -n $PKG/.MainActivity

# check if we got the broken state
adb shell "sqlite3 $DB 'SELECT status, ifnull(launch_asset_id,\"NULL\") FROM updates;'"
#  before the fix > 5|NULL     (5 = EMBEDDED, no launch asset)
#  after the fix expects no rows

# make dir writable again
adb shell "chown $(adb shell stat -c %u /data/data/$PKG):$(adb shell stat -c %u /data/data/$PKG) $INT; chmod 700 $INT"

# check
adb shell am force-stop $PKG; adb shell am start -n $PKG/.MainActivity
adb shell cat /data/data/$PKG/files/expo-error.log
# before > "Launch asset not found for update..." every cold start, forever
# after > empty; row is status 1 with a launch asset and all assets registered
```

packages/expo-updates/android/src/main/java/expo/modules/updates/launcher/DatabaseLauncher.kt
```diff
@@ -87,7 +87,7 @@ class DatabaseLauncher(
     // verify that we have all assets on disk
     // according to the database, we should, but something could have gone wrong on disk
     val launchAsset = database.updateDao().loadLaunchAssetForUpdate(launchedUpdate!!.id)
-      ?: throw Exception("Launch asset not found for update; this should never happen. Debug info: ${launchedUpdate!!.debugInfo()}")
+      ?: throw Exception("Launch asset not found for update, so the app will launch the update embedded in the binary. An interrupted registration in an older version of expo-updates can leave an update in this state; it is replaced once a newer update is downloaded. Debug info: ${launchedUpdate!!.debugInfo()}")
```
Kudo: [💡 suggestion] "so the app will launch the update embedded in the binary" is only true for the startup path, where `LoaderTask.onFailure` swaps in `NoDatabaseLauncher`. Two other callers reach this same throw and do not fall back to the embedded bundle: `LoaderTask.kt:400` (a remote update loaded but failed to launch, reported as `RemoteUpdateStatus.ERROR`) and `UpdatesDevLauncherController.kt:318` (dev launcher, which just calls `callback.onFailure`). Dropping the consequence keeps the useful part - the cause and how it resolves.

```suggestion
      ?: throw Exception("Launch asset not found for update. The update row has no launch asset; an interrupted registration in an older version of expo-updates can leave an update in this state, and it is replaced once a newer update is downloaded. Debug info: ${launchedUpdate!!.debugInfo()}")
```

packages/expo-updates/android/src/main/java/expo/modules/updates/loader/Loader.kt
```diff
@@ -169,16 +171,21 @@ abstract class Loader protected constructor(
       updateEntity = existingUpdateEntity
       return finish()
     } else {
+      val insertUpdateEntityOnFinish: Boolean
       if (existingUpdateEntity == null) {
-        // no update already exists with this ID, so we need to insert it and download everything.
+        // no update already exists with this ID, so we need to download everything.
         updateEntity = newUpdateEntity
-        database.updateDao().insertUpdate(updateEntity!!)
+        insertUpdateEntityOnFinish = newUpdateEntity.status == UpdateStatus.EMBEDDED
```
Kudo, line 180: [💡 suggestion] Worth one line saying why `EMBEDDED` is the special case, otherwise the next reader can't tell this apart from an arbitrary status check and may "simplify" it away.

```suggestion
        // EMBEDDED is already in the launchable set, so a row inserted before its launch asset exists
        // gets picked as launchable and then fails every launch.
        insertUpdateEntityOnFinish = newUpdateEntity.status == UpdateStatus.EMBEDDED
```

~Separately: is there a reason the remote path still needs the row up front? I couldn't find one - nothing between the insert and the transaction reads it (`preparePatchBaseAsset` uses `launchedUpdate`, not the update being downloaded), so deferring for every new update would drop the flag and the parameter. Happy to hear if you kept it deliberately to limit the blast radius.~

kudo: i think the separately comment isn't right. we intent to keep the old behavior for non-embedded case

Kudo (approved):
thanks for helping on the fix. i like the transaction approach

packages/expo-updates/android/src/main/java/expo/modules/updates/loader/Loader.kt
gwdp: applied on https://github.com/expo/expo/pull/49130/changes/1ebd7a1562fa0869231295543c9a1166c74b4704, thank you!

gwdp:
@Kudo @alanjhughes tagging to ask if we could cherry pick that to 55 if/when possible, please? 🙇‍♂️

afonsojramos:
would also appreciate a release that ships this 🙏

PR #47580, opened 2026-07-07
[patch-project] Fix CNG patch paths not applying in monorepos

AmariahAK:
## Description

```markdown
Fix git apply path resolution for CNG patches in monorepo setups. `git apply` resolves paths relative to the git **worktree root**, not the current working directory. In monorepos where the Expo app is nested (e.g., `apps/mobile/`), patch paths like `a/ios/mobile/Info.plist` resolve to `<repo-root>/ios/mobile/Info.plist` instead of `<repo-root>/apps/mobile/ios/mobile/Info.plist`, causing `git apply` to silently skip the patch.

Closes #47574

# Why

When an Expo app lives inside a monorepo (e.g., `apps/mobile/`), `patch-project` generates CNG patches with app-relative paths (`a/ios/...`). But `git apply` resolves paths from the git worktree root — the monorepo root — not from the cwd. The result is `Skipped patch 'ios/mobile/Info.plist'.` and the patch silently does nothing. Manually editing every patch to add the monorepo prefix (`apps/mobile/ios/...`) is not viable.

See the minimal repro: https://github.com/Yash-Singh1/patch-project-monorepo-repro

# How

Two complementary fixes:

**Application side** (`src/gitPatch.ts` — `applyPatchAsync`):
- Added `getGitRootAsync()` helper that runs `git rev-parse --show-toplevel` to find the worktree root, falling back to `projectRoot` if git isn't available
- `applyPatchAsync()` now computes the relative path from the git worktree root to `projectRoot` and passes `--directory=<relPath>` to `git apply`. This tells git to prepend the monorepo nesting prefix before resolving paths from the worktree root. Path separators are normalized to `/` for cross-platform compatibility.

**Generation side** (`src/cli/patchProjectAsync.ts` — `patchProjectForPlatformAsync`):
- The diff repo directory structure now mirrors the monorepo path. Instead of `diffDir/ios/...`, it creates `diffDir/apps/mobile/ios/...`. This makes newly generated patches contain worktree-relative paths out of the box, so they're self-contained and work regardless of which directory they're applied from.

In non-monorepo projects (`gitRoot === projectRoot`), `relPath` is empty and `--directory` is not added — behavior is unchanged.

# Test Plan

**Unit tests** (`src/__tests__/gitPatch-test.ts`):
- `getGitRootAsync`: returns git root from `rev-parse --show-toplevel`; falls back to `projectRoot` when git is unavailable
[...]

PR #49138, opened 2026-08-20
[patch-project] Apply CNG patches relative to the project directory

### LINKED ISSUE #47574: patch-project doesn't generate correct paths for cng patch in monorepos

https://github.com/Yash-Singh1/patch-project-monorepo-repro

### Steps to reproduce

Instructions are here: https://github.com/Yash-Singh1/patch-project-monorepo-repro/blob/main/SETUP_COMMANDS.md

I have it all setup in the repo with the example patch.

If you clone the repo and do the following:
1. cd into `apps/mobile`
2. Run `pnpm expo prebuild --clean -p ios`
3. Run `cat ios/mobile/Info.plist | grep PatchProjectMonorepoRepro`
4. Notice that that there is no match despite the patch adding a line with "PatchProjectMonorepoRepro" to that file
5. Manually edit the patch file in `cng-patches` replacing all instances of `/ios` with `/apps/mobile/ios` (or checkout the `fix` branch it does the same)
6. Rerun `pnpm expo prebuild --clean -p ios`
7. Reun `cat ios/mobile/Info.plist | grep PatchProjectMonorepoRepro`
8. Notice that that there is now a match

The fix would be to either fix the path when generating the patch in the future or fix the prebuild step that applies the patch to handle the path relative to the app's location in the monorepo.

### Environment

```text
    System:
      OS: macOS 26.5.1
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 24.10.0 - /Users/johndoe/.nvm/versions/node/v24.10.0/bin/node
      npm: 11.6.1 - /Users/johndoe/.nvm/versions/node/v24.10.0/bin/npm
    Managers:
    SDKs:
      iOS SDK:
    IDEs:
      Android Studio: 2022.1 AI-221.6008.13.2211.9619390
    npmPackages:
      expo: ~57.0.2 => 57.0.2
      react: 19.2.3 => 19.2.3
    Expo Workflow: managed
```

### Expo Doctor Diagnostics

```text
```

MUSE-CODE-SPACE:
# Why

Fixes #47574

In a monorepo, a CNG patch is silently ignored: `npx expo prebuild` reports no error, but none of the patched changes reach the generated native project. Editing the patch file to prefix every path with the project directory (`/ios` → `/apps/mobile/ios`) makes it work, as noted in the issue.

The cause is a mismatch between how patches are written and how they are read back:

- Patches are generated in a standalone repository whose root is the native project, so the paths they contain are project relative (`ios/...`).
- `git apply` resolves paths from the root of the *enclosing* repository, and it ignores paths that fall outside the current directory.

When the project is the repository root those two agree. When the project sits in a subdirectory of a monorepo, `ios/...` is looked up at the monorepo root, falls outside the project directory, and is skipped — without a non-zero exit code, which is why the failure is silent.

# How

Pass `--directory` to `git apply`, built from `git rev-parse --show-prefix`, so patch paths resolve against the project directory again. The prefix is empty when the project already is the repository root, so single-project setups keep the exact same arguments as before.

The same prefix is applied to the reverse check in `isPatchAppliedAsync`; without it, the already-applied detection would answer for the wrong paths in a monorepo and could skip a patch that was never applied.

Projects that are not inside a git repository at all keep working too — `git rev-parse` fails there, and `git apply` already resolves paths from the current directory, which is the project root.

# Test Plan

- Added monorepo cases to `src/__tests__/gitPatch-test.ts` for both `applyPatchAsync` and `isPatchAppliedAsync`, asserting that `--directory=apps/mobile/` is passed when `git rev-parse --show-prefix` reports a nested path, and that no `--directory` argument is added when it reports none.
- Adjusted the existing `isPatchAppliedAsync` tests, since the mocked `spawnAsync` now sees the `git rev-parse` call before `git apply`.
- I have not run the package's test suite locally, so I would appreciate CI confirming it — happy to adjust anything that needs it.
- A reproduction for the reported failure is available at https://github.com/Yash-Singh1/patch-project-monorepo-repro. I have not run it against a locally built package, and I am glad to verify it end to end before landing if maintainers prefer.

# Checklist

- [x] Documentation is up to date to reflect these changes — no documentation change needed; this makes patches behave as documented in a monorepo.
- [x] This diff will work correctly for `npx expo prebuild` & EAS Build (eas build)

### DIFF
--- a/packages/patch-project/CHANGELOG.md
+++ b/packages/patch-project/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 - Use development mode when loading Expo config and `.env` files. ([#48882](https://github.com/expo/expo/pull/48882) by [@ramonclaudio](https://github.com/ramonclaudio))
 - Skip applying a CNG patch that is already applied to the native project, e.g. when running `npx expo prebuild --no-clean` more than once. ([#47605](https://github.com/expo/expo/issues/47605) by [@MUSE-CODE-SPACE](https://github.com/MUSE-CODE-SPACE))
-- Apply CNG patches relative to the project directory, so that patches are no longer silently skipped when the project lives in a monorepo subdirectory. ([#47574](https://github.com/expo/expo/issues/47574) by [@MUSE-CODE-SPACE](https://github.com/MUSE-CODE-SPACE))
+- Apply CNG patches relative to the project directory, so that patches are no longer silently skipped when the project lives in a monorepo subdirectory. ([#49138](https://github.com/expo/expo/pull/49138) by [@MUSE-CODE-SPACE](https://github.com/MUSE-CODE-SPACE))
 
 ### 💡 Others
 
--- a/packages/patch-project/CHANGELOG.md
+++ b/packages/patch-project/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 - Use development mode when loading Expo config and `.env` files. ([#48882](https://github.com/expo/expo/pull/48882) by [@ramonclaudio](https://github.com/ramonclaudio))
 - Skip applying a CNG patch that is already applied to the native project, e.g. when running `npx expo prebuild --no-clean` more than once. ([#47605](https://github.com/expo/expo/issues/47605) by [@MUSE-CODE-SPACE](https://github.com/MUSE-CODE-SPACE))
-- Apply CNG patches relative to the project directory, so that patches are no longer silently skipped when the project lives in a monorepo subdirectory. ([#49138](https://github.com/expo/expo/pull/49138) by [@MUSE-CODE-SPACE](https://github.com/MUSE-CODE-SPACE))
+- Apply CNG patches relative to the project directory, so that patches are no longer silently skipped when the project lives in a monorepo subdirectory. A patch that was hand-edited to add the project directory to its paths must be regenerated. ([#49138](https://github.com/expo/expo/pull/49138) by [@MUSE-CODE-SPACE](https://github.com/MUSE-CODE-SPACE))
 
 ### 💡 Others
 
--- a/packages/patch-project/CHANGELOG.md
+++ b/packages/patch-project/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 - Use development mode when loading Expo config and `.env` files. ([#48882](https://github.com/expo/expo/pull/48882) by [@ramonclaudio](https://github.com/ramonclaudio))
 - Skip applying a CNG patch that is already applied to the native project, e.g. when running `npx expo prebuild --no-clean` more than once. ([#47605](https://github.com/expo/expo/issues/47605) by [@MUSE-CODE-SPACE](https://github.com/MUSE-CODE-SPACE))
+- Apply CNG patches relative to the project directory, so that patches are no longer silently skipped when the project lives in a monorepo subdirectory. ([#47574](https://github.com/expo/expo/issues/47574) by [@MUSE-CODE-SPACE](https://github.com/MUSE-CODE-SPACE))
 
 ### 💡 Others
 
--- a/packages/patch-project/src/__tests__/gitPatch-test.ts
+++ b/packages/patch-project/src/__tests__/gitPatch-test.ts
@@ -21,22 +21,85 @@ describe(applyPatchAsync, () => {
     mockedSpawnAsync.mockRejectedValue(new Error('git apply failed'));
     await expect(() => applyPatchAsync('/app', '/app/cng-patches/ios+.patch')).rejects.toThrow();
   });
+
+  it('should not scope the patch when the project is the repository root', async () => {
+    mockedSpawnAsync.mockReset();
+    // @ts-expect-error: `git rev-parse --show-prefix` reports no prefix
+    mockedSpawnAsync.mockResolvedValue({ stdout: '', stderr: '' });
+
+    await applyPatchAsync('/app', '/app/cng-patches/ios+.patch');
+
+    expect(mockedSpawnAsync).toHaveBeenLastCalledWith(
+      'git',
+      ['apply', '--ignore-whitespace', '/app/cng-patches/ios+.patch'],
+      { cwd: '/app' }
+    );
+  });
+
+  it('should scope the patch to the project directory inside a monorepo', async () => {
+    mockedSpawnAsync.mockReset();
+    // @ts-expect-error: `git rev-parse --show-prefix` reports the nested project path
+    mockedSpawnAsync.mockResolvedValue({ stdout: 'apps/mobile/', stderr: '' });
+
+    await applyPatchAsync('/repo/apps/mobile', '/repo/apps/mobile/cng-patches/ios+.patch');
+
+    expect(mockedSpawnAsync).toHaveBeenLastCalledWith(
+      'git',
+      [
+        'apply',
+        '--ignore-whitespace',
+        '--directory=apps/mobile/',
+        '/repo/apps/mobile/cng-patches/ios+.patch',
+      ],
+      { cwd: '/repo/apps/mobile' }
+    );
+  });
 });
 
 describe(isPatchAppliedAsync, () => {
   it('should return true when the patch reverse-applies cleanly', async () => {
-    // @ts-expect-error
-    mockedSpawnAsync.mockResolvedValueOnce({ stdout: '', stderr: '' });
+    mockedSpawnAsync.mockReset();
+    // @ts-expect-error: `git rev-parse --show-prefix` then `git apply --reverse --check`
+    mockedSpawnAsync.mockResolvedValue({ stdout: '', stderr: '' });
+
     await expect(isPatchAppliedAsync('/app', '/app/cng-patches/ios+.patch')).resolves.toBe(true);
-    expect(mockedSpawnAsync).toHaveBeenCalledWith(
+
+    expect(mockedSpawnAsync).toHaveBeenLastCalledWith(
       'git',
       ['apply', '--reverse', '--check', '--ignore-whitespace', '/app/cng-patches/ios+.patch'],
       { cwd: '/app' }
     );
   });
 
+  it('should check the patch against the project directory inside a monorepo', async () => {
+    mockedSpawnAsync.mockReset();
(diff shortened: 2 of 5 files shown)


packages/patch-project/src/gitPatch.ts
```diff
@@ -53,17 +53,47 @@ export async function diffAsync(
...
+ * paths are project relative. `git apply` however resolves paths from the root of the enclosing
+ * repository, and ignores paths that fall outside the current directory. In a monorepo where the
+ * project lives in a subdirectory, that combination makes every hunk target the wrong location and
+ * the patch is skipped without an error. `--directory` restores the missing prefix.
+ */
+async function getPatchPathArgsAsync(projectRoot: string): Promise<string[]> {
+  try {
+    // Empty when `projectRoot` is the repository root, `apps/mobile/` when it is nested.
+    const prefix = await runGitAsync(['rev-parse', '--show-prefix'], { cwd: projectRoot });
+    return prefix ? [`--directory=${prefix}`] : [];
```
Kudo, line 76: [🏗️ **design**] `getPatchChangedLinesAsync` is the third `git apply` in this file and still misses the prefix, so it keeps this bug for exactly the projects the PR fixes. It also passes no `cwd`, so it runs in the process working directory.

Checked with git 2.55: from `apps/mobile`, `git apply --numstat` prints nothing and exits 0. `changedLines` then computes to `0`, so the `changedLinesLimit` warning at `withPatchPlugin.ts:59` never fires. With `--directory=apps/mobile/` it prints `1	0	apps/mobile/ios/mobile/Info.plist`.

Threading `projectRoot` through is safe. The package main is `withPatchPlugin.js`, and line 57 is the only caller.

```ts
export async function getPatchChangedLinesAsync(
  projectRoot: string,
  patchFilePath: string
): Promise<number> {
  const pathArgs = await getPatchPathArgsAsync(projectRoot);
  const stdout = await runGitAsync(['apply', '--numstat', ...pathArgs, patchFilePath], {
    cwd: projectRoot,
  });
```

packages/patch-project/src/__tests__/gitPatch-test.ts
```diff
@@ -21,22 +21,85 @@ describe(applyPatchAsync, () => {
...
+    await applyPatchAsync('/app', '/app/cng-patches/ios+.patch');
+
+    expect(mockedSpawnAsync).toHaveBeenLastCalledWith(
+      'git',
+      ['apply', '--ignore-whitespace', '/app/cng-patches/ios+.patch'],
+      { cwd: '/app' }
+    );
+  });
+
+  it('should scope the patch to the project directory inside a monorepo', async () => {
```
Kudo, line 42: [🏗️ **design**] Could you add a test that runs real git? These cases assert the argv handed to a mocked `spawnAsync`, so they cannot tell whether the patch actually applies.

That gap matters here. #47580 fixes the same issue, and its mocked tests pass while its real flow fails with exit 1. A test in a temp `repo/apps/mobile` layout would cover what mocks cannot:

- the patch applies from the nested project, and the file content changes
- `isPatchAppliedAsync` returns `false` before the apply and `true` after
- the repo-root case still applies with no `--directory`
- the no-repository fallback still applies (this PR promises that behavior, and nothing covers it today)

`packages/patch-project/e2e` builds where the project is the repository root, so the prefix is empty there and this code path never runs.

packages/patch-project/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 
 - Use development mode when loading Expo config and `.env` files. ([#48882](https://github.com/expo/expo/pull/48882) by [@ramonclaudio](https://github.com/ramonclaudio))
 - Skip applying a CNG patch that is already applied to the native project, e.g. when running `npx expo prebuild --no-clean` more than once. ([#47605](https://github.com/expo/expo/issues/47605) by [@MUSE-CODE-SPACE](https://github.com/MUSE-CODE-SPACE))
+- Apply CNG patches relative to the project directory, so that patches are no longer silently skipped when the project lives in a monorepo subdirectory. ([#47574](https://github.com/expo/expo/issues/47574) by [@MUSE-CODE-SPACE](https://github.com/MUSE-CODE-SPACE))
```
Kudo: [🐧 nit] Changelog entries should link the PR, not the issue. The publish script walks the PRs referenced here to find the issues they close and comment on them, so an issue link breaks that. See `guides/contributing/Updating Changelogs.md`.

```suggestion
- Apply CNG patches relative to the project directory, so that patches are no longer silently skipped when the project lives in a monorepo subdirectory. ([#49138](https://github.com/expo/expo/pull/49138) by [@MUSE-CODE-SPACE](https://github.com/MUSE-CODE-SPACE))
```

Kudo (changes requested):

MUSE-CODE-SPACE:
Thanks for the detailed review — all three points are addressed in f46d16b.

**`getPatchChangedLinesAsync`** now takes `projectRoot` and passes both the prefix and a `cwd`; `withPatchPlugin.ts:57` is updated as the only caller. You were right that this was the one that mattered most. I reproduced it with git 2.50.1: from `apps/mobile`, `git apply --numstat` prints nothing and exits 0, so `changedLines` is `0` and the `changedLinesLimit` warning can never fire. With `--directory=apps/mobile/` it prints the expected numstat line for `apps/mobile/ios/Info.plist`.

**Real-git tests** are in a new `src/__tests__/gitPatchIntegration-test.ts`, kept separate from the existing file because that one mocks `@expo/spawn-async` and `fs`. It builds temporary directories and runs actual `git`, covering the four cases you listed:

- a project nested in a monorepo — `isPatchAppliedAsync` is `false`, the patch applies, the file contents change, and it reports `true` afterwards
- a project that is the repository root — the prefix is empty, so the arguments stay exactly as they were
- a project outside any repository — the fallback this PR promises, which nothing covered before
- a patch that does not apply — it throws and leaves the project untouched

Walking through those same steps by hand also shows the silent failure from the issue: without `--directory`, `git apply` exits 0 and the file is left unchanged, which is why nothing surfaces today.

**Changelog** now links the PR. Thanks for pointing me at the guide — I had followed an existing entry that linked an issue.

brentvatne:
@expo-bot check

packages/patch-project/src/__tests__/gitPatchIntegration-test.ts
```diff
@@ -0,0 +1,107 @@
...
+  await fs.writeFile(patchFilePath, PATCH_CONTENTS, 'utf8');
+  return patchFilePath;
+}
+
+async function initRepositoryAsync(repoRoot: string): Promise<void> {
+  await spawnAsync('git', ['init'], { cwd: repoRoot });
+}
+
+function readInfoPlistAsync(projectRoot: string): Promise<string> {
+  return fs.readFile(path.join(projectRoot, INFO_PLIST_PATH), 'utf8');
```
Kudo: [🏗️ **design**] This compares bytes, so the three content assertions fail when git converts line endings. Measured with git 2.55: `git -c core.autocrlf=true apply` writes CRLF, both inside a repository and outside one, so the no-repository case breaks too. That is the Git for Windows default, and `check-packages-windows` in `.github/workflows/sdk.yml` runs `patch-project` tests without a `core.autocrlf false` step, unlike the Windows jobs in `cli.yml`. Normalizing in one place covers all four tests.

```suggestion
  return fs
    .readFile(path.join(projectRoot, INFO_PLIST_PATH), 'utf8')
    // `git apply` writes CRLF when global `core.autocrlf` is true, the Git for Windows default.
    .then((contents) => contents.replace(/\r\n/g, '\n'));
```

packages/patch-project/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 
 - Use development mode when loading Expo config and `.env` files. ([#48882](https://github.com/expo/expo/pull/48882) by [@ramonclaudio](https://github.com/ramonclaudio))
 - Skip applying a CNG patch that is already applied to the native project, e.g. when running `npx expo prebuild --no-clean` more than once. ([#47605](https://github.com/expo/expo/issues/47605) by [@MUSE-CODE-SPACE](https://github.com/MUSE-CODE-SPACE))
+- Apply CNG patches relative to the project directory, so that patches are no longer silently skipped when the project lives in a monorepo subdirectory. ([#49138](https://github.com/expo/expo/pull/49138) by [@MUSE-CODE-SPACE](https://github.com/MUSE-CODE-SPACE))
```
Kudo: [💡 suggestion] Could you mention the hand-edited patches? Issue #47574 tells users to add the project directory to every path in the patch. This change adds it again, so `git apply` fails with `error: apps/mobile/apps/mobile/ios/...: No such file or directory` and prebuild exits 1. Those users have to regenerate the patch, and the changelog is where they will look.

```suggestion
- Apply CNG patches relative to the project directory, so that patches are no longer silently skipped when the project lives in a monorepo subdirectory. A patch that was hand-edited to add the project directory to its paths must be regenerated. ([#49138](https://github.com/expo/expo/pull/49138) by [@MUSE-CODE-SPACE](https://github.com/MUSE-CODE-SPACE))
```

Kudo (changes requested):
still some finding from newer agent review. i think that's the last one and good to merge after update

packages/patch-project/src/__tests__/gitPatch-test.ts
```diff
@@ -21,22 +21,85 @@ describe(applyPatchAsync, () => {
     mockedSpawnAsync.mockRejectedValue(new Error('git apply failed'));
     await expect(() => applyPatchAsync('/app', '/app/cng-patches/ios+.patch')).rejects.toThrow();
   });
+
+  it('should not scope the patch when the project is the repository root', async () => {
```
Kudo, line 29: [🐧 nit] Seven tests in this file now open with `mockedSpawnAsync.mockReset()`. A single top-level `beforeEach(() => { mockedSpawnAsync.mockReset(); })` does the same job and also isolates the tests you did not touch, which currently leak `mockRejectedValue` into each other.

Kudo (approved):
last nit comment

MUSE-CODE-SPACE:
Thanks! Consolidated the mock resets into a single top-level beforeEach in 755598d.

Kudo:
thanks again for this fix!

PR #43797, opened 2026-03-09
[ios][ui] Fix Slider thumb snap-back during drag

fedeciancaglini:
## Summary

Fixes #43701

`SliderView.swift` unconditionally overwrites `@State var value` inside `.onReceive(props.value.publisher)` — even while the user is actively dragging the thumb. This causes the thumb to flicker/snap back to the previous prop value on every render cycle during the drag gesture.

**Fix:** Add an `@State var isEditing` flag that tracks whether a drag is in progress (via the `onEditingStatusChanged` closure already present on all four `Slider` branches). Guard `.onReceive` so external prop updates are only applied when the user is **not** dragging.

### Changes
- `packages/expo-ui/ios/SliderView.swift`: Added `@State var isEditing` + guard in `.onReceive` + set `self.isEditing` in all 4 slider closures
- `packages/expo-ui/CHANGELOG.md`: Added bug fix entry

## Test plan

- Build and run the [minimal repro from the issue](https://github.com/expo/expo/issues/43701)
- Drag the slider thumb — it should follow the finger smoothly without flickering or snapping back
- Verify that programmatic value updates (via props) still work when the user is not dragging

Issue #43701, opened 2026-03-05
SliderView.swift: thumb snaps back during drag due to missing isEditing guard on .onReceive


fedeciancaglini:
## Summary

`SliderView.swift` in `@expo/ui` unconditionally overwrites `@State var value` inside `.onReceive(props.value.publisher)` — even when the user is actively dragging the thumb. This creates a feedback loop:

1. User drags the slider thumb
2. SwiftUI fires the slider's `onEditingChanged` closure and `onChange(of: value)` dispatches `onValueChanged` to JS
3. React re-renders with the new value, which triggers the Expo prop publisher
4. `.onReceive` fires and writes `value = sliderValue`, snapping the thumb back to the (slightly stale) prop value
5. The thumb visibly flickers/jumps during the entire drag

## Root Cause

[`SliderView.swift` lines 25-28](https://github.com/expo/expo/blob/main/packages/expo-ui/ios/SliderView.swift#L25-L28):

```swift
.onReceive(props.value.publisher, perform: { newValue in
    var sliderValue = newValue
    value = sliderValue
})
```

This has no guard for whether a drag is in progress. The slider's closure already receives an `isEditing` parameter (line 54), but it's only forwarded to JS — it's never used to gate the `.onReceive`.

Notably, [`PickerView.swift`](https://github.com/expo/expo/blob/main/packages/expo-ui/ios/Picker/PickerView.swift) in the same package already guards against this pattern using `prevSelection`.

## Proposed Fix

Track `isEditing` state and skip prop-driven updates during active drags:

```swift
struct SliderView: ExpoSwiftUI.View {
  @ObservedObject var props: SliderProps
  @State var value: Float = 0.0
  @State var isEditing: Bool = false  // <-- add

  // ...

  // Guard the .onReceive:
  .onReceive(props.value.publisher, perform: { newValue in
      guard !isEditing else { return }  // <-- add
      value = newValue
  })

  // Track editing state in all 4 slider closure branches:
  ) { editing in
      isEditing = editing  // <-- add
      props.onEditingChanged(["isEditing": editing])
  }
```

We're currently using this as a `patch-package` patch successfully.

## Minimal reproducible example

This can be reproduced with any `@expo/ui` Slider that feeds `onValueChange` back into state:

```tsx
import { useState } from "react"
import { Slider, Host } from "@expo/ui/swift-ui"

export default function App() {
    const [value, setValue] = useState(5)
    return (
[...]

PR #49139, opened 2026-08-20
[ui][ios] Fix Slider dropping JS prop update after drag

### LINKED ISSUE #49075: [expo-ui] SwiftUI `Slider` ignores `value` prop updates after the first drag

https://github.com/matinzd/slider-bug-expo-ui

### Steps to reproduce

1. Tap **Set 7.5** — the thumb moves. Prop-driven updates work before any drag.
2. Drag the slider anywhere and release.
3. Tap **Set 2.5** — the `JS state:` readout updates, but the thumb stays put.

Every later prop update is ignored too. The on-screen `onEditingChanged` log
shows what the native side reported during the drag.

**Expected:** the thumb follows the `value` prop, as it does before step 2.
**Actual:** after one drag, `value` is write-once-ignored forever.

https://github.com/user-attachments/assets/80125687-daa7-4a53-869d-1ed9f6e4b8e6

### Environment

```text
    System:
      OS: macOS 26.6
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 22.23.2 - /Users/matin/.local/share/mise/installs/node/22/bin/node
      npm: 10.9.8 - /Users/matin/.local/share/mise/installs/node/22/bin/npm
    Managers:
      CocoaPods: 1.17.0 - /Users/matin/.local/share/mise/installs/ruby/3.4/bin/pod
    SDKs:
      iOS SDK:
    IDEs:
    npmPackages:
      expo: ~57.0.14 => 57.0.14
      expo-router: ~57.0.14 => 57.0.14
      react: 19.2.3 => 19.2.3
      react-dom: 19.2.3 => 19.2.3
    Expo Workflow: bare
```

### Expo Doctor Diagnostics

```text
```

intergalacticspacehighway:
# Why

Fixes - https://github.com/expo/expo/issues/49075

https://github.com/expo/expo/pull/43797 added `isEditing` flag but it never set to `false` after the manual drag is finished. So post manual drag, JS value prop update was getting dropped.

# How

Use event count mechanism to keep the slider always accept new values and ignore if JS sends any old values due to async event.

# Test Plan

Tested user shared repro

# Checklist


packages/expo-ui/src/swift-ui/Slider/index.tsx
```diff
@@ -78,11 +83,11 @@ function transformSliderProps(props: SliderProps): NativeSliderProps {
   } = props;
   return {
     ...restProps,
-    onValueChanged: onValueChange
-      ? ({ nativeEvent: { value } }) => {
-          onValueChange(value);
-        }
-      : undefined,
+    mostRecentEventCount: eventCount.current,
+    onValueChanged: ({ nativeEvent: { value, eventCount: nativeEventCount } }) => {
```
Kudo: [💡 suggestion] Worth a short comment here. If someone later restores the old `onValueChange ? ... : undefined` guard, the ref stops advancing and native drops every `value` prop after the first drag, which is #49075 again.

```suggestion
    // Subscribe even when the caller passes no `onValueChange`.
    // Native ignores any `value` whose event count is behind its own, so a slider that stops receiving events would drop every later prop update.
    onValueChanged: ({ nativeEvent: { value, eventCount: nativeEventCount } }) => {
```

packages/expo-ui/src/swift-ui/Slider/index.tsx
```diff
@@ -93,9 +98,10 @@ function transformSliderProps(props: SliderProps): NativeSliderProps {
 
 export function Slider(props: SliderProps) {
   const { label, minimumValueLabel, maximumValueLabel } = props;
+  const eventCount = useRef(0);
```
Kudo, line 101: [💡 suggestion] Can you add a regression test for the JS half? This path has regressed twice already (#43701 -> #43797 -> #49075), and the contract is easy to break: the ref must advance on every native event, and its value must be forwarded as `mostRecentEventCount`.

`src/universal/Button/__tests__/index.test.ios.tsx` shows the pattern, and `__mocks__/expo.ts` already exposes the props handed to the native view.

Kudo: fine if that's not easy to add the test

Kudo (approved):

packages/expo-ui/src/swift-ui/Slider/index.tsx
intergalacticspacehighway: yeah, i guess JS test won't help much as the event count logic is in native and JS is just setting the `ref`. Native test requires dragging slider with gesture and setting value with JS, seems only possible with maestro/E2E 😅 (would be good to add them at some point)

PR #49143, opened 2026-08-20
[config-plugins][create-expo] Harden app-name sanitization and template renaming

vonovak:
# Why

replacement for https://github.com/expo/expo/pull/45582

When Expo generates the iOS project name from expo.name, names containing accented or non-ASCII characters can lose letters instead of being normalized into a stable ASCII form.

Example:

Skráning
before: Skrning
after: Skraning (correct)

# How

- take the old PR, add tests, also update project name for `inline modules` on iOS accordingly

the changes are:

**1. Name sanitizer rewrite** (`config-plugins/src/ios/utils/Xcodeproj.ts` + a synced copy in `create-expo/src/createFileTransform.ts`):
- Pipeline is now: drop everything that is not a letter/number (`\p{L}\p{N}`) → NFKD-normalize → strip combining marks → `slugify` → strip remaining non-word chars.
- NFKD keeps accented letters ("Árbók" → "Arbok", not "rbk") and decomposes compatibility characters ("ﬁre" → "fire", "Ǉubljana" → "LJubljana", fullwidth "Ｅｘｐｏ" → "Expo", "x²" → "x2").
- `slugify` transliterates what NFKD cannot decompose ("Bjørn" → "Bjorn", "Straße" → "Strasse", "Привет" → "Privet").
- Symbols are dropped, not transliterated ("A & B" → "AB", "Expo®" → "Expo"). Symbol-only names fall back to slugify of the full name ("♥" → "love"), then to `app`.
- `create-expo` gains the `slugify` dependency for this — it cannot import from `@expo/config-plugins`, so the function is duplicated with a keep-in-sync comment and an identical test table.

**2. Template renaming fixes** (`@expo/cli/src/prebuild/renameTemplateAppName.ts`, `create-expo/src/Template.ts`, `@expo/cli/src/utils/npm.ts`):
- Sanitized identifiers (`HelloWorld`/`helloworld`) now derive from the **raw** app name in XML/plist files too; before, they derived from the escaped name, so 'A & B' produced 'AampB' in plists but 'AB' elsewhere.
- Lowercase identifiers derive from the sanitized name (`sanitizedName(name).toLowerCase()`), not from `sanitizedName(name.toLowerCase())`, so file paths and file contents always agree (slugify's charmap is case-sensitive).
- Display-name escaping is now per format: Android resource `.xml` files use entities for `&`, `<`, `>` plus Android backslash escapes (quotes, `@`, `?`, `\`, whitespace quoting); plists use proper XML entities (`&quot;`, `&apos;`) instead of backslash escapes.
- The display name is inserted via a replacer function, so names containing replacement patterns like `$&` are not expanded.

**3. `XML.escapeAndroidString` hardening** (`config-plugins/src/utils/XML.ts`): now also escapes `?` and `\`, so values are not compiled as attribute references or aapt2 escape sequences.

# Test Plan

- CI

# Checklist


packages/@expo/config-plugins/src/ios/utils/Xcodeproj.ts
```diff
@@ -63,10 +63,7 @@ export function sanitizedName(name: string) {
 }
 
 function sanitizedNameForProjects(name: string) {
```
HubertBer: Maybe we should also change the `create-expo` sanitizedName, so that they stay the same? It may also be worth to add the`|| 'app'` fallback there for consistency.

https://github.com/expo/expo/blob/main/packages/create-expo/src/createFileTransform.ts

packages/@expo/inline-modules/src/xcodeProjectUpdates.ts
```diff
@@ -32,12 +32,16 @@ function escapeXMLCharacters(original: string): string {
...
-// It is necessary to match the target name in the path to ExpoModulesProvider.swift for the main target as is used when generating it.
-function getMainTargetName(config: InlineModulesXcodeParams): string {
-  const name = config.name;
-  const safeName = escapeXMLCharacters(name);
-  return IOSConfig.XcodeUtils.sanitizedName(safeName);
+// The on-disk project name is the source of truth: it can differ from the name
+// derived from the app config (older sanitizer output, or a manual rename).
+function getMainTargetName(projectRoot: string, config: InlineModulesXcodeParams): string {
+  try {
+    return IOSConfig.XcodeUtils.getProjectName(projectRoot);
```
HubertBer: I've realized that we can get to the main target differently by:
`const mainTarget = pbxProject.getTarget('com.apple.product-type.application');` which removes the need for sanitization here.

We don't need to do it here. I can open a follow-up PR and update other inline modules target checks.

vonovak:
@HubertBer I went a little deeper on this but believe it's for the better

packages/create-expo/package.json
```diff
@@ -60,6 +60,7 @@
     "picomatch": "^2.3.2",
     "prompts": "^2.4.2",
     "resolve-workspace-root": "^2.0.0",
+    "slugify": "^1.6.8",
```
HubertBer: Maybe match `"slugify": "^1.6.6"` from main `@expo/config-plugins`.

packages/@expo/cli/src/prebuild/renameTemplateAppName.ts
```diff
@@ -147,15 +157,28 @@ export async function renameTemplateAppNameAsync(
...
 
-      const safeName = ['.xml', '.plist'].includes(path.extname(file))
-        ? escapeXMLCharacters(name)
-        : name;
+      const extension = path.extname(file);
+      // Escaping applies only to the display name; the sanitized project
+      // identifiers derive from the raw name so they match across all files.
+      // `.xml` files in the rename config are Android resources; `.plist` is
+      // generic XML.
+      // Under `expo prebuild`, `AndroidConfig.Name.withName` later overwrites
```
HubertBer: This whole part starting at `Under expo-prebuild` is out of place and hard to wrap your head around. If you want to explain why using renameTemplateAppNameAsync inside expo prebuild works, please do it at the call site not inside this function.

  The first part of the comment would read better if it was written at the top of the function as explaining its contract e.g.:
  ```/**
   * Substitutes the template's placeholder app name inside the given files. Only
   * file contents are rewritten, and only three tokens: `Hello App Display Name`,
   * `HelloWorld` and `helloworld`.
   *
   * File and directory names are renamed separately during template extraction,
   * from the same `sanitizedName` call.
   */```

packages/@expo/cli/src/prebuild/renameTemplateAppName.ts
```diff
@@ -147,15 +157,28 @@ export async function renameTemplateAppNameAsync(
...
+      // `app_name` from the raw config name, escaped by the same
+      // `escapeAndroidString`. This escaping keeps the file valid until the
+      // mods run, and is the final output for custom rename configs and for
+      // `create-expo`, which runs no mods.
+      const safeName =
```
HubertBer:
```suggestion
      const escapedDisplayName =
```

With the rename there's no need to explain why safeName shouldn't be used for identifiers.
Also added suggestions to the code below.

packages/@expo/cli/src/prebuild/renameTemplateAppName.ts
```diff
@@ -147,15 +157,28 @@ export async function renameTemplateAppNameAsync(
...
+          : extension === '.plist'
+            ? escapeXMLCharacters(name)
+            : name;
 
       try {
         const replacement = contents
-          .replace(/Hello App Display Name/g, safeName)
-          .replace(/HelloWorld/g, IOSConfig.XcodeUtils.sanitizedName(safeName))
-          .replace(/helloworld/g, IOSConfig.XcodeUtils.sanitizedName(safeName.toLowerCase()));
+          .replace(/Hello App Display Name/g, () => safeName)
```
HubertBer:
```suggestion
          .replace(/Hello App Display Name/g, () => escapedDisplayName)
```

packages/@expo/config-plugins/src/ios/utils/Xcodeproj.ts
```diff
@@ -58,15 +58,29 @@ export function resolvePathOrProject(
 
 // TODO: come up with a better solution for using app.json expo.name in various places
 export function sanitizedName(name: string) {
-  // Default to the name `app` when every safe character has been sanitized
-  return sanitizedNameForProjects(name) || sanitizedNameForProjects(slugify(name)) || 'app';
+  // Symbols carry no name information: drop them so 'A & B' stays 'AB' and 'Expo®'
```
HubertBer: Maybe write a top level comment explaining the function contract instead. Then I'd shorten this inline comment to `// NFKD before slugify, or letters are lost: 'Ċ' decomposes to 'C' + a combining mark that the next line drops, while slugify deletes it outright.` which leaves the important ordering information in the comment and explains the function flow. Also naming regexes can help with readability more then the comment.

packages/@expo/config-plugins/src/ios/utils/Xcodeproj.ts
```diff
@@ -58,15 +58,29 @@ export function resolvePathOrProject(
...
+  // stays 'Expo'. Then strip diacritics via NFKD before slugify, because slugify
+  // deletes letters absent from its charmap ('Ċ' would vanish instead of becoming
+  // 'C'). NFKD also decomposes compatibility letters and numbers ('ﬁ' becomes 'fi',
+  // 'Ǉ' becomes 'LJ', fullwidth 'Ｅ' becomes 'E'); symbols like '™' are already
+  // dropped, so they do not turn into letters. slugify then transliterates letters
+  // that NFKD cannot decompose (ø, æ, ł, ß). Symbol-only names fall back to a
+  // slugify of the full name ('♥' becomes 'love'), then to the name `app`.
+  const lettersAndNumbers = name
+    .replace(/[^\p{L}\p{N}]+/gu, '')
```
HubertBer: Give this regex a name, then you don't need the massive comment.
`const NEITHER_LETTER_NOR_NUMBER = /[^\p{L}\p{N}]+/gu;`

packages/@expo/config-plugins/src/ios/utils/Xcodeproj.ts
```diff
@@ -58,15 +58,29 @@ export function resolvePathOrProject(
...
+    .normalize('NFKD')
+    .replace(/\p{M}+/gu, '');
```
HubertBer: Give this regex a name.
`const COMBINING_MARKS = /\p{M}+/gu;`
or 
`const NFKD_ARTIFACT = /\p{M}+/gu;`

packages/create-expo/src/createFileTransform.ts
```diff
@@ -1,14 +1,36 @@
 import { TarTypeFlag } from 'multitars';
 import path from 'path';
 import picomatch from 'picomatch';
+import slugify from 'slugify';
 
 const debug = require('debug')('expo:init:fileTransform') as typeof console.log;
 
+// Keep in sync with `sanitizedName` in `@expo/config-plugins` (src/ios/utils/Xcodeproj.ts)
+// so create-expo and prebuild derive the same project name.
+// Symbols are dropped so 'A & B' stays 'AB' and 'Expo®' stays 'Expo'. Diacritics
```
HubertBer: This can be simplified similar to the `Xcodeproj.ts` one, please name the regexes and if you want to write the top level comment just write the contract, instead of the function flow. `Keep in sync...` is the part worth leaving here.

packages/@expo/config-plugins/src/ios/utils/Xcodeproj.ts
```diff
@@ -58,15 +58,29 @@ export function resolvePathOrProject(
...
+  return (
+    sanitizedNameForProjects(slugify(lettersAndNumbers)) ||
+    sanitizedNameForProjects(slugify(name)) ||
+    'app'
+  );
 }
 
+// Normalize before stripping so diacritics decompose into combining marks that
+// `\W` removes, keeping the base letters ("Árbók" -> "Arbok", not "rbk").
 function sanitizedNameForProjects(name: string) {
```
HubertBer: This function and its comment are stale or were unnecessary before, after the changes `sanitizedNameForProjects` is
  used on slugified output, so the NFKD normalization does nothing. The only thing we need
  this function to do is to return an ASCII only identifier, so maybe call it that:

  const NOT_ASCII_ALPHANUMERIC = /[\W_]+/g;
  function toAsciiIdentifier(name: string) {
    return name.replace(NOT_ASCII_ALPHANUMERIC, '');
  }

packages/create-expo/src/Template.ts
```diff
@@ -307,15 +337,25 @@ export async function renameTemplateAppNameAsync({
...
-        ? escapeXMLCharacters(name)
-        : name;
+      const extension = path.extname(file);
+      // Escaping applies only to the display name; the sanitized project
+      // identifiers derive from the raw name so they match across all files.
+      // `.xml` files in the rename config are Android resources; `.plist` is
+      // generic XML.
+      // This is the final output here: unlike `expo prebuild`, `create-expo`
+      // runs no config mods that would rewrite `app_name` afterwards.
+      const safeName =
```
HubertBer: Check the `renameTemplateAppName.ts` comment above.

packages/create-expo/src/createFileTransform.ts
```diff
@@ -1,14 +1,36 @@
...
+  return (
+    sanitizedNameForProjects(slugify(lettersAndNumbers)) ||
+    sanitizedNameForProjects(slugify(name)) ||
+    'app'
+  );
+}
+
+// Normalize before stripping so diacritics decompose into combining marks that `\W` removes,
+// keeping the base letters ("\u00c1rb\u00f3k" -> "Arbok", not "rbk").
+function sanitizedNameForProjects(name: string) {
```
HubertBer: This function and its comment are stale or were unnecessary before, after the changes `sanitizedNameForProjects` is
  used on slugified output, so the NFKD normalization does nothing. The only thing we need
  this function to do is to return an ASCII only identifier, so maybe call it that:

  const NOT_ASCII_ALPHANUMERIC = /[\W_]+/g;
  function toAsciiIdentifier(name: string) {
    return name.replace(NOT_ASCII_ALPHANUMERIC, '');
  }

Similar to the one in `Xcodeproj.ts`

HubertBer (changes requested):
The changes to the code look good, but the comments are much too long, sometimes misleading, stale or in wrong places.

packages/@expo/cli/CHANGELOG.md
```diff
@@ -36,6 +36,7 @@
 ### 🐛 Bug fixes
 
 - Stop writing DOM component source maps into the app binary during `expo export:embed`, so Android release builds no longer ship `www.bundle` source maps with the original app source. ([#49480](https://github.com/expo/expo/pull/49480) by [@expo-bot](https://github.com/expo-bot))
+- Derive sanitized project identifiers in the prebuild template from the raw app name in XML and plist files too (e.g. 'A & B' produced 'AampB' in plists but 'AB' elsewhere), and lowercase identifiers from the sanitized name, so file paths and contents always agree. Escape display names for Android resources and plists, and insert names containing replacement patterns (e.g. '$&') literally. ([#49143](https://github.com/expo/expo/pull/49143) by [@vonovak](https://github.com/vonovak))
```
HubertBer: The changelog entry is now in an already released section.

packages/@expo/config-plugins/CHANGELOG.md
```diff
@@ -18,6 +18,8 @@
 
 ### 🐛 Bug fixes
 
+- Normalize accented and compatibility characters instead of stripping them when deriving iOS project names from the app name ('Árbók' now becomes 'Arbok', not 'rbk'; 'ﬁre' becomes 'fire'), and transliterate letters that normalization cannot decompose ('Bjørn' becomes 'Bjorn'). ([#49143](https://github.com/expo/expo/pull/49143) by [@vonovak](https://github.com/vonovak))
```
HubertBer: Changelog entries are in the released section. Probably a rebase artifact

packages/create-expo/CHANGELOG.md
```diff
@@ -24,6 +24,7 @@
 
 ### 🐛 Bug fixes
 
+- Derive project names from the app name the same way as `@expo/config-plugins`: normalize accented and compatibility characters ('Árbók' now becomes 'Arbok', not 'rbk'; 'ﬁre' becomes 'fire'), transliterate letters that normalization cannot decompose ('Bjørn' becomes 'Bjorn'), and fall back to 'app'. Derive sanitized identifiers from the raw app name in XML and plist files too, and escape display names for Android resources and plists. ([#49143](https://github.com/expo/expo/pull/49143) by [@vonovak](https://github.com/vonovak))
```
HubertBer: Probably another rebase artifact.

HubertBer (approved):
Changelog entries are under the already published sections, that's probably a rebase artifact 😅. Please fix before merging!

PR #49146, opened 2026-08-20
[ios][secure-store] Reject `deleteItemAsync` when the keychain delete fails

### LINKED ISSUE #49126: [secure-store][iOS] deleteItemAsync discards SecItemDelete OSStatus, so failed deletes resolve successfully
# Summary

`SecureStore.deleteItemAsync` on iOS resolves successfully even when the underlying keychain delete fails. In `SecureStoreModule.swift`, `deleteValueWithKeyAsync` calls `SecItemDelete` three times (legacy, authenticated, and non-authenticated search dictionaries) and discards all three `OSStatus` return values:

https://github.com/expo/expo/blob/main/packages/expo-secure-store/ios/SecureStoreModule.swift#L48-L50

```swift
SecItemDelete(legacySearchDictionary as CFDictionary)
SecItemDelete(authSearchDictionary as CFDictionary)
SecItemDelete(noAuthSearchDictionary as CFDictionary)
```

Any status other than `errSecSuccess` or `errSecItemNotFound` (such as `errSecInteractionNotAllowed` when protected data is unavailable, or `errSecAuthFailed`) is silently swallowed, and the promise resolves as if the item were removed.

This is the iOS sibling of the Android `commit()` issues in #48988 and #48989, where the write path discards the `SharedPreferences` commit result. The failure mode matters most for credential lifecycle code: a logout flow that calls `deleteItemAsync` for tokens can report success while the tokens remain in the keychain.

# What happens

1. The app calls `deleteItemAsync("refreshToken")` in a state where the keychain refuses the operation. The most common trigger is deleting while the device is locked when the item was stored with a `kSecAttrAccessible` class that makes it unavailable when locked: `SecItemDelete` returns `errSecInteractionNotAllowed`.
2. The returned `OSStatus` is discarded, so the promise resolves.
3. JS assumes the credential is gone, but the keychain still holds it. A later `getItemAsync` after unlock returns the "deleted" value.

# Expected behavior

`deleteItemAsync` should reject when a delete fails, mirroring how the module throws `KeyChainException` for failed reads and writes. `errSecItemNotFound` should still be treated as success, since deleting a missing key is not an error.

# Suggested fix

We are running this in production via patch-package on `expo-secure-store@55.0.16`. All three variants are attempted before throwing, so a failure on the legacy alias cannot strand the live entries:

```swift
let legacyStatus = SecItemDelete(legacySearchDictionary as CFDictionary)
let authStatus = SecItemDelete(authSearchDictionary as CFDictionary)
let noAuthStatus = SecItemDelete(noAuthSearchDictionary as CFDictionary)

for status in [legacyStatus, authStatus, noAuthStatus] {
  guard status == errSecSuccess || status == errSecItemNotFound else {
    throw DeleteException(status)
  }
}
```

with a small exception type alongside the existing ones in `SecureStoreExceptions.swift`:

```swift
internal final class DeleteException: GenericException<OSStatus> {
  override var reason: String {
    return "Could not delete the value from SecureStore. Status code: \(param)"
  }
}
```

This intentionally does not touch the two `SecItemDelete` calls in the write path (lines 117-118 on main), where clearing a possibly-absent previous entry before an add is best-effort by design.

Happy to open a PR with this fix if useful.

# Environment

- expo-secure-store 55.0.16 (SDK 55), but the code is unchanged on `main`
- Verified on iOS Simulator (iOS 18) with the patched module built from source (`expo.autolinking.apple.buildFromSource`)

# Reproducible demo

Because the failure requires a keychain-level fault, there is no pure JS repro. To reproduce deterministically: store an item with `keychainAccessible: SecureStore.WHEN_UNLOCKED`, lock the device (or background the app with protected data unavailable), and trigger `deleteItemAsync` from a background task. The call resolves even though `SecItemDelete` returned `errSecInteractionNotAllowed`, and the item is still readable after unlock.

vonovak:
Verified by /verify for #49126. Run: https://github.com/expo/expo/actions/runs/32291050766

# Why

closes #49126

# How

check the return value and throw if it's not `errSecSuccess` or `errSecItemNotFound`

# Test Plan

this is a verified fix as indicated by screnshots from https://github.com/expo/expo/issues/49126#issuecomment-5347123827

# Checklist

### DIFF
--- a/packages/expo-secure-store/CHANGELOG.md
+++ b/packages/expo-secure-store/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### 🐛 Bug fixes
 
+- [iOS] Reject `deleteItemAsync` when the keychain refuses the delete, instead of resolving as if the item was removed.
+
 ### 💡 Others
 
 ## 57.0.1 - 2026-07-15
--- a/packages/expo-secure-store/ios/SecureStoreModule.swift
+++ b/packages/expo-secure-store/ios/SecureStoreModule.swift
@@ -45,9 +45,17 @@ public final class SecureStoreModule: Module {
       let authSearchDictionary = query(with: key, options: options, requireAuthentication: true)
       let legacySearchDictionary = query(with: key, options: options)
 
-      SecItemDelete(legacySearchDictionary as CFDictionary)
-      SecItemDelete(authSearchDictionary as CFDictionary)
-      SecItemDelete(noAuthSearchDictionary as CFDictionary)
+      // Delete all three aliases before reporting a failure, so that a failing alias
+      // cannot leave the remaining entries behind.
+      let statuses = [
+        SecItemDelete(legacySearchDictionary as CFDictionary),
+        SecItemDelete(authSearchDictionary as CFDictionary),
+        SecItemDelete(noAuthSearchDictionary as CFDictionary)
+      ]
+
+      if let failure = statuses.first(where: { $0 != errSecSuccess && $0 != errSecItemNotFound }) {
+        throw KeyChainException(failure)
+      }
     }
 
     Function("canUseBiometricAuthentication") {() -> Bool in


packages/expo-secure-store/CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
 
 ### 🐛 Bug fixes
 
+- [iOS] Reject `deleteItemAsync` when the keychain refuses the delete, instead of resolving as if the item was removed.
```
vonovak, line 13:
```suggestion
- [iOS] Reject `deleteItemAsync` when the keychain refuses the delete, instead of resolving as if the item was removed. ([#49146](https://github.com/expo/expo/pull/46089) by [@expo-bot](https://github.com/expo-bot))
```

vonovak: will need to add this later

Kudo: [🐧 nit] Heads up on the suggestion above: the href points to PR 46089 and it credits `@expo-bot`, so applying it as-is lands the wrong link and wrong author. Corrected:

```suggestion
- [iOS] Reject `deleteItemAsync` when the keychain refuses the delete, instead of resolving as if the item was removed. ([#49146](https://github.com/expo/expo/pull/49146) by [@vonovak](https://github.com/vonovak))
```

Kudo (approved):

vonovak:
@expo-bot open a PR ti add the missing links to the changelog entries

PR #49147, opened 2026-08-20
[ios][autolinking] Propagate prebuilt unavailability to 3rd-party deps

chrfalch:
# Why

A prebuilt pod exposes its headers only inside its XCFramework, so a pod that reads them from `Pods/Headers/Public/<dep>` cannot compile against a prebuilt dependency. 

`RNReanimated.podspec` hardcodes that path for RNWorklets, so a source-built RNReanimated failed with `'worklets/Compat/StableApi.h' file not found` whenever RNWorklets was prebuilt.

The reason for the source build of Reanimated and not Worklets happening is that we only check dependencies one way - so if RNWorklets are source built, Reanimated will also be - but the opposite is not detected.

# How

This PR fixes this by adding another map that contains the other direction for the dependency graph - making sure we also build RNWorklets as source if Reanimated is source built.

# Test Plan

Test | Assertion | Result
-- | -- | --
Control | stock versions all prebuilt | 9/9 📦, no fallbacks
Natural drift | reanimated 4.5.0 (404) + worklets 0.10.1 → cascade, compiles | HTTP 404 → cascade → BUILD SUCCEEDED
@expo/ui | no cascade, list unchanged | 9/9 📦, identical to control, BUILD SUCCEEDED

# Checklist


packages/expo-modules-autolinking/scripts/ios/precompiled_modules.rb
```diff
@@ -1794,6 +1666,40 @@ def prebuilt_dependency_pods(external_dependencies)
...
+      #
+      # An edge only needs this coupling when the dependent reads the dependency's headers
+      # from `Pods/Headers/Public/<dep>`, which CocoaPods populates only while the
+      # dependency builds from source. `RNReanimated.podspec:69,85` hardcodes
+      # `"$(PODS_ROOT)/Headers/Public/RNWorklets"`, which is why reanimated breaks against
+      # a prebuilt RNWorklets. Pods that resolve those headers elsewhere are immune:
+      # ExpoModulesWorkletsAdapter points at the npm tree
+      # (`node_modules/react-native-worklets/Common/cpp`), and the ~60 edges onto
+      # ExpoModulesCore are covered by `configure_header_search_paths`, which puts the
+      # prebuilt ExpoModulesCore headers on every target.
```
sjchmiela: does this quite specific explanation belong here?

sjchmiela (approved):

packages/expo-modules-autolinking/scripts/ios/precompiled_modules.rb
chrfalch: Nope - I removed it :) Thanks.

PR #46881, opened 2026-06-13
fix(precompiled,ios): fall back to source when a prebuilt module is ABI-incompatible with ExpoModulesCore

rohan-patra:
# Why

Fixes #46880.

Prebuilt module xcframeworks are compiled in this monorepo at each package's publish time, against whatever `expo-modules-core` exists then. Nothing couples the published artifacts together: the packages declare only `"expo": "*"` peers, so a project can resolve binary-incompatible combinations, e.g. a stale lockfile keeping `expo-modules-core@56.0.15` while a dependency-bot update floats `expo-file-system`/`expo-font` to the 2026-06-10 wave built against core `56.0.16` (which changed the `Record`/`AnyModule` protocol ABIs in #46547/#46612).

That combination installs and builds with no errors. The failure surfaces only after shipping: the embedded `ExpoFileSystem.framework`/`ExpoFont.framework` strongly bind (verified via `dyld_info -fixups`: plain `bind`, not weak imports) to 10 `ExpoModulesCore` symbols the embedded core binary doesn't export, so the app aborts at launch with dyld `Symbol not found`, and App Store Connect flags every upload with ITMS-90863. The skew breaks in both directions (56.0.16 also removed the `_exposedDefinition` exports that binaries built against ≤56.0.15 import).

# How

`precompiled_modules.rb` already falls back to source builds for several availability reasons (missing tarball, missing platform slice, dependency unavailable). This PR adds one more gate to `_resolve_prebuilt_status_uncached`: an ABI compatibility check between each prebuilt artifact and the resolved prebuilt `ExpoModulesCore` artifact.

- `prebuilt_core_abi_missing_symbols` extracts the framework binary from the pod's tarball (preferring the `ios-arm64` slice; the ABI is identical across slices) and lists its undefined symbols with `nm -ju`, filtered to the `_$s15ExpoModulesCore` mangling prefix so every checked symbol provably belongs to core's Swift module (no false positives).
- Those imports are compared against the exported symbol set (`nm -gjU`) of the resolved `ExpoModulesCore` artifact, memoized once per install.
- On mismatch the pod resolves as `{ available: false, reason: :core_abi_mismatch }`, the existing machinery then builds it from source, and a warning explains the skew and recommends `npx expo install --fix`.

[...]

PR #49148, opened 2026-08-20
[ios][autolinking] Fall back to source builds when a prebuilt module is ABI-incompatible with E-M-C

### LINKED ISSUE #46880: [iOS] Precompiled module xcframeworks can resolve with an ABI-incompatible expo-modules-core, causing dyld crash at launch and ITMS-90863

https://github.com/rohan-patra/expo-prebuilt-abi-skew-repro

### Steps to reproduce

**Summary:** SDK 56's precompiled iOS modules can resolve into a binary-incompatible combination: the prebuilt xcframeworks inside `expo-file-system@56.0.8` / `expo-font@56.0.6` were compiled against `expo-modules-core@56.0.16` (which changed the `Record` / `AnyModule` protocol ABIs in #46547 / #46612), but nothing prevents them from installing alongside `expo-modules-core@56.0.15`. The embedded `ExpoModulesCore.framework` then lacks 10 symbols that `ExpoFileSystem.framework` / `ExpoFont.framework` **strongly** bind to (verified with `dyld_info -fixups`: plain `bind`, not weak imports). Consequences:

- dyld aborts at app launch (`Symbol not found`) on iOS, and
- App Store Connect emits **ITMS-90863** ("The app uses symbols that aren't present in macOS: `@rpath/ExpoModulesCore.framework/ExpoModulesCore` …") listing exactly those symbols, once per referencing framework.

This is how we hit it in production (bare workflow EAS build, pnpm): our lockfile predated 2026-06-10, so `expo@56.0.9` → `expo-modules-core@~56.0.15` stayed resolved at `56.0.15`; a Renovate update then floated `expo-file-system`/`expo-font` to the versions published on 2026-06-10 (~40 min after core 56.0.16). No install-time or build-time error: the first signal was the ITMS-90863 email for our TestFlight build. Since the prebuilt packages declare only `"expo": "*"` peers, package managers have no metadata to prevent mixing publish waves; the MRE pins core via `overrides` purely to make that stale-lockfile state deterministic.

**Steps (fast path, no Xcode build needed, ~10 s; macOS with CLT):**

1. `git clone https://github.com/rohan-patra/expo-prebuilt-abi-skew-repro && cd expo-prebuilt-abi-skew-repro`
2. `npm install`
3. `npm run check-abi`, which extracts the prebuilt tarballs that ship inside the npm packages (`prebuilds/output/release/xcframeworks/`) and compares each module binary's `ExpoModulesCore` imports (`nm -m`) against the core binary's exports (`nm -gU`).

**Expected:** every `ExpoModulesCore` symbol imported by a prebuilt module binary is exported by the resolved `expo-modules-core` prebuilt binary (or autolinking refuses the combination / falls back to source).

**Actual:**

```
expo-modules-core@56.0.15 prebuilt (release, ios-arm64): 2441 exported symbols

❌ expo-file-system@56.0.8 (ExpoFileSystem): 10 of 184 ExpoModulesCore imports are NOT exported by expo-modules-core@56.0.15:
     _$s15ExpoModulesCore6RecordP4from10dictionary10appContextxSDySSypG_AA03AppH0CtKFZTq
     _$s15ExpoModulesCore6RecordP4from6object10appContextx0aB3JSI16JavaScriptObjectV_AA03AppH0CtKFZTq
     _$s15ExpoModulesCore6RecordP8toObject10appContext0aB3JSI010JavaScriptF0VAA03AppH0C_tKFTq
     _$s15ExpoModulesCore6RecordPAAE4from10dictionary10appContextxSDySSypG_AA03AppH0CtKFZ
     _$s15ExpoModulesCore6RecordPAAE4from6object10appContextx0aB3JSI16JavaScriptObjectV_AA03AppH0CtKFZ
     _$s15ExpoModulesCore6RecordPAAE8toObject10appContext0aB3JSI010JavaScriptF0VAA03AppH0C_tKF
     _$s15ExpoModulesCore9AnyModuleP09_decorateE06object2in10appContexty0aB3JSI16JavaScriptObjectV_AH0lM7RuntimeCAA03AppJ0CtKFTq
     _$s15ExpoModulesCore9AnyModuleP22_synthesizedDefinitionSayAA0dG0_pGyFTq
     _$s15ExpoModulesCore9AnyModulePAAE09_decorateE06object2in10appContexty0aB3JSI16JavaScriptObjectV_AH0lM7RuntimeCAA03AppJ0CtKF
     _$s15ExpoModulesCore9AnyModulePAAE22_synthesizedDefinitionSayAA0dG0_pGyF

❌ expo-font@56.0.6 (ExpoFont): 10 of 75 ExpoModulesCore imports are NOT exported by expo-modules-core@56.0.15:
     (same 10 symbols)
```

Demangled, these are exactly the additions from core `56.0.16`: `Record.from(dictionary:appContext:)`, `Record.from(object:appContext:)`, `Record.toObject(appContext:)` (each as method descriptor + protocol-extension default impl), and `AnyModule._synthesizedDefinition()` / `AnyModule._decorateModule(object:in:appContext:)` from the `@ExpoModule` macro work. The skew also breaks in the **reverse direction**: 56.0.16 removed the `_exposedDefinition` exports that module binaries built against 56.0.9-56.0.15 import, so a *newer* core with *older* prebuilt siblings fails the same way.

**Full-build path (what we hit in production):** build an SDK 56 iOS app with this dependency state (precompiled modules on by default) → app crashes at launch with dyld `Symbol not found`; uploading to App Store Connect produces the ITMS-90863 email.

**Suggested direction:**

1. Couple each prebuilt package to the `expo-modules-core` version/ABI it was compiled against (exact version metadata checked by autolinking), and/or
2. At `pod install` time, validate that each prebuilt artifact's `ExpoModulesCore` imports are satisfied by the resolved core artifact and fall back to source builds on mismatch (the fallback mechanism from #45664 already exists). Symbol-level validation catches the problem regardless of how the bad combination came to be (stale lockfiles, overrides, partial updates).

Happy to contribute a PR for option 2; will link it here.

### Environment

```text
    System:
      OS: macOS 26.3.1
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 22.22.3 - /Users/rohan/.nvm/versions/node/v22.22.3/bin/node
      npm: 10.9.8 - /Users/rohan/.nvm/versions/node/v22.22.3/bin/npm
    Managers:
    SDKs:
      iOS SDK:
    IDEs:
    npmPackages:
      expo: 56.0.11 => 56.0.11
      react: 19.2.3 => 19.2.3
    Expo Workflow: managed
```

### Expo Doctor Diagnostics

```text
```

### LINKED ISSUE #47015: [expo-media-library][iOS] 56.0.7 crashes at launch with dyld symbol-not-found in ExpoModulesCore
## Summary

`expo-media-library@56.0.7` causes an immediate **app crash on launch** on real iOS devices when built with EAS (release/preview profiles). The crash happens before any JS runs because `dyld` cannot resolve a Swift symbol that `ExpoMediaLibrary.framework` expects from `ExpoModulesCore.framework`.

Debug/Metro builds (`expo run:ios`) are unaffected because the native framework bundling path is different.

## Reproduction

1. Expo SDK 56 app with `expo-media-library: "56.0.7"` (or `~56.0.7`)
2. Build a release IPA with EAS (`eas build --platform ios --profile preview --local`)
3. Install on a physical iOS device
4. App crashes immediately on launch — before the splash screen dismisses

## Error

From **Console.app** (filter on app process, crash reports):

```
ReportCrash   ASI found [dyld] (sensitive) 'Symbol not found:
  _$s15ExpoModulesCore6RecordPAAE4from10dictionary10appContextxSDySSypG_AA03AppH0CtKFZ
  Referenced from: <...>/ravergy.app/Frameworks/ExpoMediaLibrary.framework/ExpoMediaLibrary
  Expected in:     <...>/ravergy.app/Frameworks/ExpoModulesCore.framework/ExpoModulesCore'

Process exit reason: <NSError: 0x... domain:dyld(6) code:4 "(null)">
```

The mangled symbol demangles to `ExpoModulesCore.Record (extension in ExpoModulesCore).from(dictionary:appContext:)` — a Swift method that `expo-media-library@56.0.7` was compiled against but is not exported by `expo-modules-core@56.0.14`.

## Environment

| Package | Version |
|---|---|
| `expo` | 56.0.9 |
| `expo-media-library` | 56.0.7 (broken) / 56.0.6 (working) |
| `expo-modules-core` | 56.0.14 |
| `react-native` | 0.85.3 |
| iOS | 18.x |
| EAS CLI | local build |
| New Architecture | enabled |

## Root cause hypothesis

`expo-media-library@56.0.7` (published 2026-06-10) appears to have been compiled against a version of `expo-modules-core` that exports `Record.from(dictionary:appContext:)`, but `expo-modules-core@56.0.14` (the version that ships with Expo SDK 56.0.9) does not export this symbol. The mismatch causes `dyld` to abort at process launch.

## Workaround

Pin to the previous patch version and exclude from `expo install` resolution:

```json
// package.json
{
  "dependencies": {
    "expo-media-library": "56.0.6"
  },
  "expo": {
    "install": {
      "exclude": ["expo-media-library"]
    }
  }
}
```

## Additional context

- The `56.0.7` release changelog entry is missing from `packages/expo-media-library/CHANGELOG.md` — the version was published as part of a batch release train on 2026-06-10 (gitHead: `dab8054`)
- Downgrading to `56.0.6` and rebuilding resolves the crash completely

### LINKED ISSUE #49090: expo-print causes iOS launch crash: dyld Symbol not found in ExpoModulesCore

No public minimal reproduction is available yet because this was reproduced in a private production app.  The crash is triggered by adding expo-print@57.0.1 to an Expo SDK 57 iOS app and rebuilding the packaged iOS app. Removing expo-print and rebuilding fixes the launch crash.

### Steps to reproduce

1. Use an Expo SDK 57 app with the following relevant versions:
   - expo@57.0.4
   - expo-print@57.0.1
   - expo-modules-core@57.0.3
   - react-native@0.86.0
   - react@19.2.3

2. Build a packaged iOS app with EAS Build.

3. Install the app on a real iOS device.

4. Launch the app.

Expected:
The app should launch normally.

Actual:
The app crashes immediately on launch, before JavaScript error handling can catch anything.

The crash disappears after removing expo-print and rebuilding the app.

Relevant crash log:

```text
Process exited:
<RBSProcessExitContext| specific,
status:<RBSProcessExitStatus| domain:dyld(6) code:4>>

Symbol not found:
_$s15ExpoModulesCore9AnyModulePAAE09_decorateE06object2iny0aB3JSI16JavaScriptObjectV_AG0jK7RuntimeCtKF

Referenced from:
AItest.app/Frameworks/ExpoPrint.framework/ExpoPrint

Expected in:
AItest.app/Frameworks/ExpoModulesCore.framework/ExpoModulesCore

### Environment

```text
**Environment**

System:
  OS: macOS 26.5.2
  Shell: 5.9 - /bin/zsh
Binaries:
  Node: 22.16.0 - /opt/homebrew/opt/node@22/bin/node
  npm: 10.9.2 - /opt/homebrew/opt/node@22/bin/npm
  Watchman: 2025.03.10.00 - /opt/homebrew/bin/watchman
Managers:
SDKs:
  iOS SDK:
IDEs:
npmPackages:
  babel-preset-expo: 57.0.2 => 57.0.2
  expo: ~57.0.4 => 57.0.4
  expo-updates: ^57.0.6 => 57.0.6
  react: 19.2.3 => 19.2.3
  react-dom: 19.2.3 => 19.2.3
  react-native-web: 0.21.2 => 0.21.2
Expo Workflow: managed

Additional context:
- Package manager: pnpm
- Build system: EAS Build
- EAS iOS image: macos-sequoia-15.6-xcode-26.2
- The crash reproduced with expo-print@57.0.1 installed.
```

### Expo Doctor Diagnostics

```text

This project uses Hermes V1 with expo@57.0.4, which is affected by a known memory regression.
Detected Hermes V1 250829098.0.14 from React Native. Hermes V1 250829098.0.15 and earlier are affected by this regression; 250829098.0.16 is the first version that contains the fix.
Advice:
Upgrade to Expo SDK 57 and expo@57.0.9 or later by running `npx expo install expo@^57.0.9 --fix`.

Minor version mismatches:
react-native-screens expected ~4.26.0, found 4.25.2

Patch version mismatches:
@expo/dom-webview expected ~57.0.1, found 57.0.0
@expo/metro-runtime expected ~57.0.11, found 57.0.3
expo expected ~57.0.14, found 57.0.4
expo-asset expected ~57.0.12, found 57.0.3
expo-dev-client expected ~57.0.13, found 57.0.5
expo-file-system expected ~57.0.4, found 57.0.0
expo-router expected ~57.0.14, found 57.0.4
expo-sharing expected ~57.0.13, found 57.0.3
expo-updates expected ~57.0.15, found 57.0.6
react-native expected 0.86.2, found 0.86.0

Note:
expo-doctor was run after removing expo-print as a workaround, because the app crashes on launch when expo-print@57.0.1 is installed.
```

### LINKED ISSUE #49091: SDK 57 iOS launch crash after adding expo-print: DYLD Symbol missing in ExpoModulesCore

https://github.com/rrni/expo-print-dyld-repro

### Steps to reproduce

1. Clone the minimal reproduction repo:
   https://github.com/rrni/expo-print-dyld-repro

2. Install dependencies:

   ```sh
   pnpm install
3. Build a packaged iOS app:
   eas build --platform ios --profile ios-internal
4. Install the generated build on a real iPhone.
5. Launch the app.
Expected:
The app should launch normally.
Actual:
The app crashes immediately on launch.
Crash log from the minimal reproduction:
Process:             ExpoPrintDyldRepro [4870]
Identifier:          com.example.expoprintdyldrepro
Version:             1.0.0 (1)
Code Type:           ARM-64 (Native)
Hardware Model:      iPhone16,1
OS Version:          iPhone OS 26.6 (23G71)

Exception Type:  EXC_CRASH (SIGABRT)
Termination Reason: DYLD 4 Symbol missing

Symbol not found:
_$s15ExpoModulesCore10BaseModuleC11willDestroyyyFTj

Referenced from:
ExpoPrintDyldRepro.app/Frameworks/ExpoFileSystem.framework/ExpoFileSystem

Expected in:
ExpoPrintDyldRepro.app/Frameworks/ExpoModulesCore.framework/ExpoModulesCore

Binary Images include:
ExpoPrintDyldRepro.app/Frameworks/ExpoPrint.framework/ExpoPrint
ExpoPrintDyldRepro.app/Frameworks/ExpoFileSystem.framework/ExpoFileSystem
ExpoPrintDyldRepro.app/Frameworks/ExpoModulesCore.framework/ExpoModulesCore
The crash happens at launch before JavaScript starts, so it is not catchable by JavaScript error handling.
The minimal repro intentionally pins:
- expo@57.0.4
- expo-print@57.0.1
- expo-modules-core@57.0.3 via pnpm.overrides
- expo-modules-jsi@57.0.1 via pnpm.overrides
- react-native@0.86.0
- react@19.2.3


### Environment

```text
    System:
      OS: macOS 26.5.2
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 22.16.0 - /opt/homebrew/bin/node
      npm: 10.9.2 - /opt/homebrew/bin/npm
      Watchman: 2025.03.10.00 - /opt/homebrew/bin/watchman
    Managers:
    SDKs:
      iOS SDK:
    IDEs:
    npmPackages:
      babel-preset-expo: 57.0.2 => 57.0.2 
      expo: 57.0.4 => 57.0.4 
      react: 19.2.3 => 19.2.3 
    Expo Workflow: managed
```

### Expo Doctor Diagnostics

```text

This project uses Hermes V1 with expo@57.0.4, which is affected by a known memory regression.
Detected Hermes V1 250829098.0.14 from React Native. Hermes V1 250829098.0.15 and earlier are affected by this regression; 250829098.0.16 is the first version that contains the fix.
Advice:
Upgrade to Expo SDK 57 and expo@57.0.9 or later by running `npx expo install expo@^57.0.9 --fix`. See https://expo.dev/changelog/sdk-57#known-regressions for the latest details.

🔧 Patch version mismatches
package          expected  found   
expo             ~57.0.14  57.0.4  
expo-status-bar  ~57.0.1   57.0.0  
react-native     0.86.2    0.86.0  

Changelogs:
- expo-status-bar → https://github.com/expo/expo/blob/sdk-57/packages/expo-status-bar/CHANGELOG.md

3 packages out of date.
Advice:

```

chrfalch:
# Why

Precompiled XCFrameworks are compiled when each package publishes. Expo packages publish independently, so the `expo-modules-core` a module was built against is usually **not** the one a project installs - there is no existing mechanism for coupling these artifacts except for the explicit run of the `npx expo-doctor` command. Any non-valid combination of packages will happily be installed and resolved by the package manager as long as its versions are satisfied.

> Supersedes #46881. The approach — comparing the symbols a prebuilt module binds against those the installed `ExpoModulesCore` exports — is [@rohan-patra](https://github.com/rohan-patra)'s, from that PR. Thank you! This reopens it against current `main` with a different attribution mechanism and a different fallback scope, both explained below.

Fixes #49091, Fixes #49090, Fixes #47015, Fixes #46880.

## Example

In #49091 the missing symbol was `_$s15ExpoModulesCore10BaseModuleC11willDestroyyyFTj`. `expo-file-system` never mentions `willDestroy` in its source; it picks the symbol up through a protocol witness, which is why this class of break is invisible to source review.

# How

**Compare symbols** in the ruby function `_resolve_prebuilt_status_uncached` using `dyld_info -fixups`, which dumps the dyld-facing metadata of a Mach-O binary. `-fixups` lists every location dyld has to patch when it loads that binary, and each `bind` row names both the target library and the symbol. That gives the linker's own attribution, so a symbol is assigned to core by what was recorded at build time rather than by its mangling.

**Compares both debug and release.** A mismatch in either flavour sends the pod to source, and the warning names the configuration when the mismatch is confined to one. Release exports are a strict subset of debug (2643 against 5398), so a debug-only scan could approve a pair that is broken in Release — the App Store case.

**Inspects the device slice**, with an exact per-platform pattern: `*/ios-arm64/`, `*/tvos-arm64/`, and `*/macos-*/` since macOS has no simulator. Exact rather than a wildcard so module and core can never resolve different variants. Autolinking also serves macOS and tvOS, and when the target platform is unknown the check is skipped rather than assuming iOS.

**The fallback is per-pod.** A module rebuilt from source compiles against the *installed* core, so it cannot carry the skew, and every compatible module keeps its prebuilt xcframework. The existing dependency cascade still takes compatible *dependents* of a mismatched pod along — a mismatched `ExpoModulesWorklets` also rebuilds `ExpoModulesWorkletsAdapter`, `ExpoUI` and `ExpoWidgets` — so the warning names them rather than under-reporting.

**Fails open.** Any tooling failure — `tar`, `dyld_info`, `nm`, an unreadable member, an unknown platform — skips the check and links exactly as before. `/usr/bin/tar` is pinned because the extraction relies on bsdtar's member globbing.

**Cost**, measured end to end over the local artifact set: the scan takes 1.90 s for 50 binary inspections, so **38 ms per inspection, 76 ms per pod** across both flavours. `-q` is what keeps that down — it stops bsdtar at the first match rather than scanning the whole archive, 10.3 ms against 23.1 ms per extraction.

Against that, `read_xcframework_info_plists_from_tarball` already costs 32.3 ms per tarball on the same path, and #49580 halves it to 17.1 ms. Net effect per pod, both flavours: **65 ms today, 110 ms with both PRs merged** — so about **+46 ms per pod**, or +2.7 s on a 60-pod install.

# Test Plan

> **NOTE:** build fails on main too — run 32832982386 on main

✅ `et check-packages expo-modules-autolinking`
✅ `ruby -c` passes (no ruby tests in our harness)

<details>
<summary>✅ Symbol extraction, gate wiring and false-positive sweep (16 assertions)</summary>

Driven against the real artifacts in `packages/precompile/.build` with a stubbed `Pod::UI`. The RED case withholds the `willDestroy` exports from `ExpoModulesCore`'s real symbol set, so it is a real binary against real exports with two symbols removed, not a hand-built fixture.

```
PASS  nil platform -> no glob (skip, do not guess ios)
PASS  nil platform -> reader skips
PASS  ios glob is exact (excludes simulator/catalyst) — */ios-arm64/X.framework/X
PASS  macos glob matches its single slice
PASS  tvos glob is exact
PASS  core exports read per flavour — {"debug"=>5398, "release"=>2643}
PASS  ExpoCamera reads the device slice (194, not the 196 union)
PASS  device slice excludes the simulator-only symbol
PASS  GREEN compatible
PASS  RED release-only detected
PASS  RED names the release configuration
PASS  RED names cascaded dependents when applicable
PASS  RED ExpoFileSystem falls back — :core_abi_mismatch
PASS  RED core stays prebuilt
PASS  sweep: no false positives — 36 pairs, 0 bad
PASS  fail open on missing tarball

ALL 16 ASSERTIONS PASSED
```

A false positive would needlessly drop a module to source for every user, so the sweep runs every locally built artifact in both flavours against its own flavour of core.

</details>

<details>
<summary>ℹ️ Device and simulator slices are not identical — what that means here</summary>

An earlier revision of this description claimed the two slices bind the same symbols, citing #46881. Review disproved it. Exports are identical, binds are not: 3 of 50 module-flavour pairs differ, in both directions.

```
module               flavour  device  sim  sim-only  device-only
ExpoBackgroundTask   release      81   77         0            4
ExpoCamera           debug       194  196         2            0
ExpoCamera           release     193  196         3            0
```

The whole simulator-only surface is four core symbols — three on `Exceptions.SimulatorNotSupported`, plus a `Promise.resolve` overload — and all four are present in every published core checked. Reading both slices was implemented and then reverted: it cost 2.7× to cover those four, and a device mismatch is the failure worth catching, since it reaches users as a launch crash or an ITMS-90863 rejection while a simulator one surfaces on the developer's next run.

So this is a stated limit, not an oversight: a simulator-only skew would still surface as a launch crash locally rather than a `pod install` warning.

</details>

<details>
<summary>✅ Verified in a real `pod install` and Xcode build</summary>

A fresh SDK 57.0.15 app (`expo@57.0.15`, `expo-file-system@57.0.4`, `expo-print@57.0.1`),
with only this patch applied on top of the published `expo-modules-autolinking@57.0.10`:

| config | `pod install` | Xcode build | launch |
| --- | --- | --- | --- |
| core 57.0.12 (matched) | no warning, 5/5 prebuilt | **BUILD SUCCEEDED** | — |
| core 57.0.8 (narrow skew) | `ExpoFileSystem (8)` → source, **4/5 stay prebuilt** | **BUILD SUCCEEDED** | **app launches** |
| core 57.0.3 (wide skew) | 3 modules → source, core stays prebuilt | BUILD FAILED | — |
| core 57.0.3, **vanilla autolinking** | no warning, prebuilt linked | BUILD FAILED — *identical error* | — |
| `EXPO_USE_PRECOMPILED_MODULES=0` | fully inert, no output | — | — |

The core 57.0.8 row is the one that validates per-pod scope: a source-built `ExpoFileSystem`
linking a prebuilt `ExpoModulesCore.framework` compiles (201 compile actions in that target)
and the app launches on a simulator with every framework loaded — no dyld failure. Build
success alone would not be evidence here, since the broken configurations also build; the
launch is the proof.

Removing the override returned the install to 5/5 prebuilt with `Podfile.lock` back to
`.podspec`, so the gate leaves no sticky state.

These rows were measured before the slice and cascade changes. They are unaffected by them:
the verdicts turn on which pods fall back, and the 16-assertion suite above confirms the
same behaviour — 0 false positives, `ExpoFileSystem` falling back, core staying prebuilt —
against the current commit.

**The skewed config cannot build with or without this change.** `expo@57.0.15` calls
`expo::createReactSchedulerHandle`, declared in `expo-modules-core@57.0.12`
(`ios/JS/EXReactSchedulerDispatch.h`) and absent from 57.0.3. It is called from
`expo/ios/AppDelegates/ExpoReactNativeFactory.mm`, which compiles from source either way:

```
ExpoReactNativeFactory.mm:60:33: error: no member named 'createReactSchedulerHandle' in namespace 'expo'
```

The vanilla control row proves the failure is the version skew, not the fallback. What
this change adds is a precise diagnostic naming the missing symbols and the fix, in place
of a C++ error in a file the user never touched.

The scan costs nothing measurable — the matched install completed in 17 seconds.

</details>

## What a user sees

```
[Expo-precompiled] 1 prebuilt Expo module needs ExpoModulesCore symbols that
expo-modules-core 57.0.8 does not provide, so it will be built from source:

    ExpoFileSystem (8)

For example: dispatch thunk of ExpoModulesCore.BaseModule.willDestroy() -> ()

Every compatible Expo module keeps its prebuilt xcframework. Linking the
mismatched binary instead would build with no error, then crash at launch
with dyld `Symbol not found`, and App Store Connect would reject the upload as
ITMS-90863.

A source build is slower, and for a wide version gap it may not compile either,
because these packages' sources can call core APIs this version does not have.
Aligning your versions is the actual fix:

  npx expo install --fix

If that does not change the versions, something is pinning them. Check for an
`overrides` or `resolutions` block in package.json, then delete your lockfile and
node_modules and install again. A lockfile keeps an old version while the range
still allows it.
```

# Checklist


chrfalch:
Here are the verification scripts used by Claude Code when verifying if anyone wants to re-run them

<details>
<summary>Reproducing the verification</summary>
There is no Ruby test harness in this package (no rspec in the Gemfile, no spec directory) and CocoaPods is being removed, so verification was done with a throwaway script plus real installs. Both are reproducible.

### Against the monorepo's own artifacts. Save as repro.rb in the repo root and run ruby repro.rb — needs packages/precompile/.build populated:

```ruby
require 'cocoapods'
W = []
module Pod
  module UI
    def self.warn(m) = W << m.to_s
    def self.puts(m) = nil
    def self.info(m) = nil
  end
end
require './packages/expo-modules-autolinking/scripts/ios/precompiled_modules.rb'
M = Expo::PrecompiledModules
M.configure(target_platform: :ios)

ex = M.send(:core_exports_by_flavor)
puts "core exports: #{ex.transform_values(&:size)}"

# No false positives: every local artifact against its own flavour of core.
bad = ex.sum do |flavor, exports|
  Dir["packages/precompile/.build/*/output/#{flavor}/xcframeworks/*.tar.gz"].count do |tb|
    product = File.basename(tb, '.tar.gz')
    next false if product == 'ExpoModulesCore'
    bound = M.send(:core_symbols_bound_by_prebuilt, tb, product)
    bound && !bound.empty? && bound.any? { |s| !exports.include?(s) }
  end
end
puts "false positives: #{bad}  (expect 0)"

# Detection: withhold the willDestroy exports from release only.
doctored = ex.to_h { |f, s| [f, f == 'release' ? s.reject { |x| x.include?('willDestroy') }.to_set : s] }
M.send(:reset_core_abi_caches); M.instance_variable_set(:@status_cache, {})
M.instance_variable_set(:@core_exports, doctored)
M.instance_variable_set(:@core_exports_resolved, true)
puts "detected: #{M.send(:core_abi_incompatible?)}  (expect true)"
puts "ExpoFileSystem: #{M.send(:resolve_prebuilt_status, 'ExpoFileSystem')[:reason].inspect}  (expect :core_abi_mismatch)"
puts "ExpoModulesCore: #{M.send(:resolve_prebuilt_status, 'ExpoModulesCore')[:available]}  (expect true — per-pod scope)"
puts W.find { |x| x.include?('will be built from source') }
```

**Output:**

```bash
core exports: {"debug"=>5375, "release"=>2627}
false positives: 0  (expect 0)
detected: true  (expect true)
ExpoFileSystem: :core_abi_mismatch  (expect :core_abi_mismatch)
ExpoModulesCore: true  (expect true — per-pod scope)

[Expo-precompiled] 18 prebuilt Expo modules need ExpoModulesCore symbols that
expo-modules-core 57.0.8 does not provide in the release configuration, so they
will be built from source:

    EXApplication (2), ExpoAudio (2), ExpoBackgroundTask (2), ExpoBrownfield (2), ExpoCamera (2), and 13 more
```

The release-only variant is the case that matters: it is what a build_flavor-only scan would have missed.

### Against a real app. A fresh SDK 57 app, with only this patch applied over the published expo-modules-autolinking:

```bash
npx create-expo-app@latest abi-check --template blank
cd abi-check && npx expo prebuild -p ios
# copy this PR's precompiled_modules.rb over node_modules/expo-modules-autolinking/scripts/ios/
npm pkg set 'overrides.expo-modules-core=57.0.8'   # narrow skew; use 57.0.3 for a wide one
npm install
cd ios && EXPO_USE_PRECOMPILED_MODULES=1 RCT_USE_RN_DEP=1 RCT_USE_PREBUILT_RNCORE=1 pod install
```

At core 57.0.8 that reports ExpoFileSystem (8) and leaves the other four pods prebuilt; Podfile.lock shows ExpoFileSystem (from ../node_modules/expo-file-system/ios) while ExpoModulesCore stays on its podspec. Removing the override returns it to 5/5 prebuilt. The resulting mixed build — source module against prebuilt core — compiles and launches.

### Against published npm artifacts. 

Substituting expo-media-library@56.0.7 and expo-modules-core@56.0.14 for the local ones reports MISS 10, the first symbol demangling to Record.from(dictionary:appContext:) — the symbol in #47015's crash report.

</details>

tsapeta (approved):

Kudo (changes requested):
i'm wondering if that's any other solution option? if the abi incompatibilty comes from breaking changes. we should maybe bump minor version? my main concern is the cost for `dyld_info` from `pod install` time. hopefully we could find some other way to do it differently

chrfalch:
> i'm wondering if that's any other solution option? if the abi incompatibilty comes from breaking changes. we should maybe bump minor version? my main concern is the cost for `dyld_info` from `pod install` time. hopefully we could find some other way to do it differently

Thanks :) Both points led somewhere — updating this now that the changes have landed.

**On cost.** Two things were doing double work. The scan decompressed each tarball twice, once to list it and once to extract the member; that is now a single `-q` extraction, since bsdtar stops at the first match instead of scanning the whole archive. And review turned up that `read_xcframework_info_plists_from_tarball` has always done the same 1 + N decompression for every prebuilt pod, so I fixed that separately in #49580.

Measured end to end over the local artifact set:

```
ABI scan          1.90 s / 50 inspections   38 ms each, 76 ms per pod (both flavours)
plist read        32.3 ms -> 17.1 ms per tarball  (#49580)
```

Net per pod, both flavours: **65 ms today, 110 ms with both merged — about +46 ms**, or +2.7 s on a 60-pod install. Not free, and I would rather state it plainly than round it down.

Two levers remain if that is still too much. The symbol cache is process-local, so every `pod install` redoes the work; keying a cache on tarball path, size and mtime would make repeat installs essentially free, which is where the developer pain actually is. And the extracted binary is still buffered in memory before being written, which is a few MB.

The check also now inspects only the device slice. Simulator slices bind slightly different symbols — the whole difference is four core symbols, three on `Exceptions.SimulatorNotSupported` — and covering them cost 2.7×, which was not worth it for a failure that surfaces on the developer's next run rather than in a user's crash report.

**On bumping the minor version.** I measured this rather than argue it, and you are right in principle — it just cannot stand alone.

Set-diffing exported symbols between consecutive published cores, `56.0.15 -> 56.0.16` **removed 4 exports**, including `AnyModule._exposedDefinition()`, in a *patch* release. So a module built against 56.0.15 declaring `^56.0.15` installs happily with 56.0.16 and crashes — caret ranges permit exactly the upgrade that breaks it. Only exact pinning would express it, which is untenable.

A minor bump would have expressed it: 56.0.16 should have been 56.1.0. But it shipped as a patch, so the discipline was not applied in the one case we can check — understandable, since a removed `_exposedDefinition` is as easy to miss as the *added* `willDestroy` that caused #49091 in the other direction.

So I think they compose rather than compete: a declared constraint prevents bad trees going forward, and the symbol comparison is what tells you a bump was missed. Worth noting no first-party module declares `expo-modules-core` as a dependency or peer today — I checked file-system, image, camera, video, print, audio and location — so that side is a cross-package change deserving its own issue. Happy to write it up.

chrfalch:
id:a06e2141a5b6

Fixed, and thank you — the double decompression was real and the suggested remedy was the right one.

`inspect_prebuilt_binary` now extracts with a single `-q` pass instead of a listing pass followed by an extract. `-q` stops bsdtar at the first match rather than scanning the whole archive:

```
extraction    23.1 ms -> 10.3 ms
scan          1.90 s / 50 inspections = 38 ms each
```

The member pattern is exact per platform — `*/ios-arm64/`, `*/tvos-arm64/`, `*/macos-*/` — so module and core can never resolve different variants, and a dSYM cannot match, its binary sitting under `dSYMs/<Product>.framework.dSYM/Contents/Resources/DWARF/`.

Your second suggestion, reusing the entry list `read_xcframework_info_plists_from_tarball` already builds, pointed at something bigger: that function has the same 1 + N pattern and predates this PR. It is fixed separately in #49580 — one pass instead of a listing plus one extract per plist, identical parsed output across all 54 local tarballs, 32.3 ms to 17.1 ms each. It deliberately does not use `-q`, because a tarball can hold several xcframeworks and the caller requires all of them to support the target.

Two suggestions not taken. Streaming via `Open3.popen3` would avoid the in-memory copy, but the binaries are a few MB and the added complexity sits in a path that must fail open cleanly. A cross-install cache keyed on path, size and mtime is the right idea and would make repeat installs nearly free; it is left out for now since per-install cost is down and it raises cache-location and invalidation questions worth deciding separately.

One correction: the finding states that `ensure_artifacts` only copies release tarballs. It copies both — `['debug', 'release'].each` at line 815 — so the "about 2 to about 6 decompressions" arithmetic is off. The conclusion still held, which is why the fix landed.

PR #49151, opened 2026-08-20
[secure store] fix `deleteItemAsync` resolving when a retried delete left the value on disk

### LINKED ISSUE #48989: [expo-secure-store] Android: a retried deleteItemAsync resolves while the value is still on disk
On Android, a `SecureStore.deleteItemAsync` that is **retried** after a failed one resolves
successfully while the value is still on disk.

`deleteItemImpl` does check its `commit()` results and throws — that part is right. But each
`commit()` is gated on `prefs.contains(...)`, which reads the **in-memory** map, and a failed
`commit()` has already removed the key from that map without rolling it back. So the second call
issues no `commit()` at all, `success` stays `true`, and the promise resolves for an entry the file
still holds.

For a secure store the consequence is a confidentiality claim rather than a durability one: the app
is told the secret was deleted, and the ciphertext is still on the device.

https://github.com/jeffchen1981-fu/expo-secure-store-android-commit-repro — **section B** of its
README. The same app also reproduces a separate write-path defect, filed as its own issue.

Created with `npx create-expo-app@latest --template blank` + `npx expo install expo-secure-store`,
plus a four-button `App.js` and an explicit `android.package`. Nothing else, nothing patched.

### Steps to reproduce

Android only. Development build (`npx expo run:android`), not Expo Go — step 2 needs `run-as`, which
requires a debuggable build. npm.

1. Tap **1. write v1**, then **3. read**. A value has to exist before the injection: AOSP only takes
   the rename-to-backup route when the file already exists.
2. `adb shell run-as com.example.securestorerepro chmod 500 shared_prefs`
3. Tap **4. delete**. It rejects — correctly.
4. Tap **4. delete** a second time.
5. `adb shell am force-stop com.example.securestorerepro`, relaunch, tap **3. read**.

Restore with `chmod 700 shared_prefs`.

**Expected:** step 4 rejects as well, since nothing has changed on disk.

**Actual**, on a physical Pixel 10 Pro XL (Android 17). `[repro]` lines are the example app logging
whether each promise resolved; the `SharedPreferencesImpl` line is AOSP's:

```
22:48:33.286  28272  [repro] setItemAsync('k', 'v1')
22:48:33.320  28272  [repro]   -> RESOLVED
22:48:34.938  28272  [repro] getItemAsync('k')        -> "v1"
              disk: 382 bytes, mtime 22:48, md5 657dbef1d02d9fe07fcea17a3b10247c

--- chmod 500 shared_prefs ---

22:48:55.029  28272  [repro] deleteItemAsync('k')
22:48:55.032  28272  E SharedPreferencesImpl: Couldn't rename file
              /data/user/0/com.example.securestorerepro/shared_prefs/SecureStore.xml
              to backup file .../SecureStore.xml.bak
22:48:55.047  28272  [repro]   -> REJECTED: Error: Call to function
              'ExpoSecureStore.deleteValueWithKeyAsync' has been rejected.
22:48:56.941  28272  [repro] getItemAsync('k')
22:48:56.969  28272  [repro]   -> null

22:49:25.936  28272  [repro] deleteItemAsync('k')
22:49:25.961  28272  [repro]   -> RESOLVED
22:49:27.868  28272  [repro] getItemAsync('k')
22:49:27.887  28272  [repro]   -> null
              disk: 382 bytes, mtime 22:48, md5 657dbef1d02d9fe07fcea17a3b10247c  (unchanged)

--- force-stop, relaunch (new pid) ---

22:50:12.335  30162  [repro] getItemAsync('k')
22:50:12.370  30162  [repro]   -> "v1"
```

1. **The first delete rejects.** Correct — the failure is reported.
2. **The read then returns `null`** although the entry is still in the file, because the in-memory
   map has already lost the key.
3. **The second delete resolves, and no `SharedPreferencesImpl` line accompanies it** — the last one
   in the log is still `22:48:55`, from the first attempt. No `commit()` was issued at all. The file
   is byte-identical to the baseline.
4. **After a restart the value is back**: `"v1"`, from an entry the app was told had been deleted.

With `chmod 700` restored, the same **4. delete** tap resolved, the file shrank from 382 to 65 bytes
(mtime `22:50`, md5 `9781ca003f10f8d0c9c1945b63fdca7f`), and a read after another restart returned
`null` — so the directory permission is the only variable.

### Environment

```text
    System:
      OS: Windows 11 10.0.26200
    Binaries:
      Node: 24.16.0 - C:\Program Files\nodejs\node.EXE
      npm: 11.16.0 - C:\Program Files\nodejs\npm.CMD
    SDKs:
      Android SDK:
        API Levels: 35, 36
        Build Tools: 34.0.0, 35.0.0, 36.0.0, 36.1.0, 37.0.0
        System Images: android-34 | AOSP ATD Intel x86_64 Atom, android-35 | Google APIs Intel x86_64 Atom
    IDEs:
      Android Studio: AI-252.28238.7.2523.14688667
    npmPackages:
      expo: ~57.0.13 => 57.0.13
      react: 19.2.3 => 19.2.3
    Expo Workflow: bare
```

`expo-secure-store@57.0.1`, `minSdk 24`.

### Expo Doctor Diagnostics

```text
```

---

## Where

`packages/expo-secure-store/android/src/main/java/expo/modules/securestore/SecureStoreModule.kt`,
`deleteItemImpl`:

```kotlin
    if (prefs.contains(keychainAwareKey)) {
      success = prefs.edit().remove(keychainAwareKey).commit()
    }

    if (prefs.contains(key)) {
      success = prefs.edit().remove(key).commit() && success
    }

    if (legacyPrefs.contains(key)) {
      success = legacyPrefs.edit().remove(key).commit() && success
    }

    if (!success) {
      throw DeleteException("Could not delete the item from SecureStore", key, options.keychainService)
    }
```

The three guards are the whole of it. From AOSP `SharedPreferencesImpl`
(`core/java/android/app/SharedPreferencesImpl.java`, read on `main`):

- `contains()` is `synchronized (mLock) { awaitLoadedLocked(); return mMap.containsKey(key); }` — the
  in-memory map, never the file.
- `commitToMemory()` applies its edits to that same live map (`mapToWriteToDisk = mMap`, then
  `mapToWriteToDisk.remove(k)` for a removal) and `commit()` only afterwards returns the *disk*
  result. Nothing restores the map when the disk write fails.

So after the first failed delete the key is absent from the map and present in the file, and every
subsequent `deleteItemAsync` for that key takes zero `commit()` calls and returns success.

## Two things this is not

**Not just "the retry is wrong".** The state is reachable without any retry logic in the app: any
later `deleteItemAsync` for the same key in the same process behaves this way, including one issued
by unrelated code minutes later.

**Not permanently absent from disk either — the deletion is suspended, not cancelled.** All
SecureStore keys share one preferences file and `writeToFile` writes the whole map, so if any later
write to that file commits successfully, the map — which no longer holds the key — is what lands,
and the deletion takes effect at a moment nothing in JS observed. If the process dies first, the
entry is still in the file and reads normally again, which is what the transcript above shows.

## Relation to the write-path issue

Same root cause — `commitToMemory()` mutating the live map with no rollback — but different code and
a different fix, which is why they are separate issues. The write path discards `commit()`'s result
outright; this path checks it and is then defeated by reading the map to decide whether to check at
all.

## Scope

Android. I have not tested the iOS implementation.

vonovak:
# Why

closes #48989 - a somewhat rare, but legit issue

# How

delete items unconditionally, without `prefs.contains` check, making it more robust

# Test Plan

- green CI, added a test that would be red in previous configuration

# Checklist

### DIFF
--- a/packages/expo-secure-store/android/src/main/java/expo/modules/securestore/SecureStoreModule.kt
+++ b/packages/expo-secure-store/android/src/main/java/expo/modules/securestore/SecureStoreModule.kt
@@ -246,22 +246,12 @@ open class SecureStoreModule : Module() {
   }
 
   private fun deleteItemImpl(key: String, options: SecureStoreOptions) {
-    var success = true
-    val prefs = getSharedPreferences()
-    val keychainAwareKey = createKeychainAwareKey(key, options.keychainService)
-    val legacyPrefs = PreferenceManager.getDefaultSharedPreferences(reactContext)
-
-    if (prefs.contains(keychainAwareKey)) {
-      success = prefs.edit().remove(keychainAwareKey).commit()
-    }
-
-    if (prefs.contains(key)) {
-      success = prefs.edit().remove(key).commit() && success
-    }
-
-    if (legacyPrefs.contains(key)) {
-      success = legacyPrefs.edit().remove(key).commit() && success
-    }
+    val success = removeItem(
+      getSharedPreferences(),
+      PreferenceManager.getDefaultSharedPreferences(reactContext),
+      key,
+      createKeychainAwareKey(key, options.keychainService)
+    )
 
     if (!success) {
       throw DeleteException("Could not delete the item from SecureStore", key, options.keychainService)
@@ -391,4 +381,15 @@ open class SecureStoreModule : Module() {
     const val AUTHENTICATED_KEYSTORE_SUFFIX = "keystoreAuthenticated"
     const val UNAUTHENTICATED_KEYSTORE_SUFFIX = "keystoreUnauthenticated"
   }
+}
+
+internal fun removeItem(
+  prefs: SharedPreferences,
+  legacyPrefs: SharedPreferences,
+  key: String,
+  keychainAwareKey: String
+): Boolean {
+  val removedFromPrefs = prefs.edit().remove(keychainAwareKey).remove(key).commit()
+  val removedFromLegacyPrefs = legacyPrefs.edit().remove(key).commit()
+  return removedFromPrefs && removedFromLegacyPrefs
 }
--- a/packages/expo-secure-store/android/src/test/resources/robolectric.properties
+++ b/packages/expo-secure-store/android/src/test/resources/robolectric.properties
@@ -1,0 +1,2 @@
+# The delete-retry test needs API 26+, where a failed commit is retried. Default here is API 23.
+sdk=35
--- a/packages/expo-secure-store/CHANGELOG.md
+++ b/packages/expo-secure-store/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ### 🐛 Bug fixes
 
+- [Android] Fix `deleteItemAsync` resolving when a retried delete left the value on disk. ([#49151](https://github.com/expo/expo/pull/49151) by [@vonovak](https://github.com/vonovak))
 - [iOS] Reject `deleteItemAsync` when the keychain refuses the delete, instead of resolving as if the item was removed. ([#49146](https://github.com/expo/expo/pull/49146) by [@vonovak](https://github.com/vonovak))
 - [iOS] Apply `keychainAccessible` when overwriting an existing item, instead of silently keeping the accessibility it was first stored with. ([#49128](https://github.com/expo/expo/pull/49128) by [@JoRo-Code](https://github.com/JoRo-Code) and [@behenate](https://github.com/behenate))
 
--- a/packages/expo-secure-store/android/src/test/java/expo/modules/securestore/SecureStoreDeleteTest.kt
+++ b/packages/expo-secure-store/android/src/test/java/expo/modules/securestore/SecureStoreDeleteTest.kt
@@ -1,0 +1,57 @@
+package expo.modules.securestore
+
+import android.content.Context
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertNull
+import org.junit.Assert.assertTrue
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.RuntimeEnvironment
+import java.io.File
+
+@RunWith(RobolectricTestRunner::class)
+class SecureStoreDeleteTest {
+  private val context: Context
+    get() = RuntimeEnvironment.getApplication()
+
+  private val sharedPreferencesDir: File
+    get() = File(context.applicationInfo.dataDir, "shared_prefs")
+
+  private val prefs
+    get() = context.getSharedPreferences("SecureStore", Context.MODE_PRIVATE)
+
+  private val legacyPrefs
+    get() = context.getSharedPreferences("legacy", Context.MODE_PRIVATE)
+
+  @Test
+  fun `removes both the current and the legacy key format`() {
+    prefs.edit().putString("keychain-key", "current").putString("key", "legacy").commit()
+    legacyPrefs.edit().putString("key", "oldest").commit()
+
+    assertTrue(removeItem(prefs, legacyPrefs, "key", "keychain-key"))
+
+    assertNull(prefs.getString("keychain-key", null))
+    assertNull(prefs.getString("key", null))
+    assertNull(legacyPrefs.getString("key", null))
+  }
+
+  @Test
+  fun `reports failure when a delete is retried after a failed commit`() {
+    prefs.edit().putString("keychain-key", "ciphertext").commit()
+    // Create the file up front so that only the SecureStore file can fail the write below.
+    legacyPrefs.edit().putString("unrelated", "value").commit()
+
+    // Without write permission on the directory `SharedPreferencesImpl` cannot rename the file to
+    // its backup, so `commit` returns false and leaves the entry on the disk. It still removes the
+    // key from its in-memory map, which is what defeats a `contains` check on the retry.
+    assertTrue(sharedPreferencesDir.setWritable(false))
+    try {
+      assertFalse(removeItem(prefs, legacyPrefs, "key", "keychain-key"))
+      assertFalse(removeItem(prefs, legacyPrefs, "key", "keychain-key"))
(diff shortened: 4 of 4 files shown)


packages/expo-secure-store/android/src/main/java/expo/modules/securestore/SecureStoreModule.kt
```diff
@@ -392,3 +382,23 @@ open class SecureStoreModule : Module() {
...
+ */
+internal fun removeItem(
+  prefs: SharedPreferences,
+  legacyPrefs: SharedPreferences,
+  key: String,
+  keychainAwareKey: String
+): Boolean {
+  val removedFromPrefs = prefs.edit().remove(keychainAwareKey).remove(key).commit()
+  val removedFromLegacyPrefs = legacyPrefs.edit().remove(key).commit()
+  return removedFromPrefs && removedFromLegacyPrefs
```
vonovak, line 394: "Returns true if the new values were successfully written to persistent storage."

https://developer.android.com/reference/android/content/SharedPreferences.Editor#commit()

apparently "no value written  / deleted" is also success

vonovak:
@copilot resolve the merge conflicts in this pull request

vonovak:
wow great job @copilot - only took 15 minutes

packages/expo-secure-store/android/src/main/java/expo/modules/securestore/SecureStoreModule.kt
```diff
@@ -392,3 +382,23 @@ open class SecureStoreModule : Module() {
...
+
+/**
+ * Removes an item saved under the current key format, under the legacy one, and under the even
+ * older default preferences, then reports whether the removals reached the disk.
+ *
+ * The removals are unconditional and use `commit`, not `apply`. `contains` reads the in-memory map,
+ * which a failed `commit` has already changed, so gating on it makes a retried delete skip the
+ * `commit` and report success while the entry is still in the file. Only `commit` reports the disk
+ * result, and from API 26 on it retries a write that an earlier failed `commit` left pending.
+ */
```
behenate: I am in no way against AI, I use it myself for basically everything, but we really need to remove useless comments. This adds no value to the PR - the code below documents itself perfectly fine. We are basically writing the same thing twice here. Also the comment references the old code while a future reader will see only the new version of the code causing confusion. (tbh I was confused myself when reading it lol)

```suggestion
```

packages/expo-secure-store/android/src/test/resources/robolectric.properties
```diff
@@ -0,0 +1 @@
+sdk=35
```
behenate, line 2: What is this for?

behenate (changes requested):

brentvatne:
@expo-bot review

packages/expo-secure-store/android/src/main/java/expo/modules/securestore/SecureStoreModule.kt
vonovak: agreed, removed

packages/expo-secure-store/android/src/test/resources/robolectric.properties
vonovak: added a short explanation

behenate (approved):

jeffchen1981-fu:
Could this get the `sdk-57` label? We're on SDK 57 in production, pinned to
`expo-secure-store@~57.0.1`, so the branch is where this fix becomes reachable for us. There is one
small wrinkle in backporting it worth knowing in advance.

### Where it stands on `sdk-57`

Of the 30 PRs merged to `main` on 2026-08-24, ten carry an `sdk-57` label and seven of those are
already on the branch. This one carries no `sdk-*` label at all.

The closest precedent that day is #48878, `[sqlite] Fix kv-store permanently losing the storage
table on a raced first launch` — a storage-integrity fix in a storage package, from an external
contributor, labelled `sdk-56`/`sdk-57` and picked the same day, under an hour before this PR
merged.

The branch is publishing, too: `expo-updates@57.0.17` went out on 2026-08-24. #48878 is on `sdk-57`
but not yet released — `expo-sqlite` is still at `57.0.1` — so landing on the branch and riding the
next publish looks like the normal path, which is all this is asking for.

### What applies cleanly, and the one thing that does not

`sdk-57`'s copy of `SecureStoreModule.kt` is identical to what `57.0.1` shipped, and both of this
PR's source hunks apply to it: every removed line and every context line is present verbatim, five
lines higher than the patch expects — `deleteItemImpl` sits at `:243` there against `:248` here.
`patch --dry-run` applies both at offset −5.

The two new test files are the exception. `SecureStoreDeleteTest.kt` needs

```gradle
testImplementation 'junit:junit:4.13.2'
testImplementation "org.robolectric:robolectric:4.16"
```

which reached `main` in #48556 (`Add confirmation option`, 2026-08-18) — a feature PR with no `sdk-*`
label, and not on the branch. So the backport is either the source change plus those two
`testImplementation` lines, or the source change alone with the test left on `main`, whichever you
prefer.

### The one behavior change

Dropping the `contains()` guards means the removals now always issue real commits. `deleteItemAsync`
still resolves for an absent key, so the semantics hold, but the commits can create files that were
not there before:

- `legacyPrefs` is committed unconditionally on **every** delete, so on an install that has never
  used the default `SharedPreferences`, the first `deleteItemAsync` of any kind creates
  `<package>_preferences.xml`.
- On an install that has never written anything at all, the delete also creates `SecureStore.xml`.

Both land at 65 bytes — an empty map — and once a file exists an absent-key delete does not rewrite
it, because AOSP's early-out in `writeToFile` sits inside `if (fileExists)`. Those sizes are measured
on a physical Pixel 10 Pro XL against clean installs, but against the two patch variants discussed in
#48989 rather than this PR's exact shape; the delete-path file creation is identical, the difference
is that this PR batches the two current-prefs removes into one commit.

Flagging it because it matters more in a patch release for a shipped SDK than on `main`.

Happy to open the backport PR in whichever of the two shapes you prefer.

PR #49163, opened 2026-08-20
[router] remove freezeOnBlur prop

Ubax:
# Why

We plan to replace `react-freeze` with `Activity`, so we need to remove the `freezeOnBlur` prop to keep only one behavior for freezing screens.

# How

# Test Plan

CI

# Checklist

---

### DIFF
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### 🛠 Breaking changes
 
+- Remove freezeOnBlur prop ([#49163](https://github.com/expo/expo/pull/49163) by [@Ubax](https://github.com/Ubax))
 - Remove the experimental web modal implementation. See the [web modals guide](https://docs.expo.dev/router/advanced/web-modals/). ([#49204](https://github.com/expo/expo/pull/49204) by [@Ubax](https://github.com/Ubax))
 - Remove `NavigationIndependentTree` and `useNavigationIndependentTree` from `expo-router/react-navigation`. To embed an isolated navigation tree in a screen, use `@react-navigation/native`. ([#49172](https://github.com/expo/expo/pull/49172) by [@Ubax](https://github.com/Ubax))
 - Remove the deprecated `Link` and `useLinkProps` exports from `expo-router/react-navigation`. Use `Link` from `expo-router` with an `href` instead. ([#48895](https://github.com/expo/expo/pull/48895) by [@Ubax](https://github.com/Ubax))
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### 🛠 Breaking changes
 
-- Remove freezeOnBlur prop ([#49163](https://github.com/expo/expo/pull/49163) by [@Ubax](https://github.com/Ubax))
+- The `freezeOnBlur` screen option is disabled and screens are never frozen when blurred. ([#49163](https://github.com/expo/expo/pull/49163) by [@Ubax](https://github.com/Ubax))
 - Remove the experimental web modal implementation. See the [web modals guide](https://docs.expo.dev/router/advanced/web-modals/). ([#49204](https://github.com/expo/expo/pull/49204) by [@Ubax](https://github.com/Ubax))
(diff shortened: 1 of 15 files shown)


packages/expo-router/src/react-navigation/native-stack/views/NativeStackView.native.tsx
```diff
@@ -400,11 +392,8 @@ const SceneView = ({
         ]}
         unstable_sheetFooter={unstable_sheetFooter}
         {...screenNativeProps}
-        headerConfig={headerConfig}
-        // When ts-expect-error is added, it affects all the props below it
-        // So we keep any props that need it at the end
-        // Otherwise invalid props may not be caught by TypeScript
-        shouldFreeze={shouldFreeze}>
+        freezeOnBlur={false}
```
jakub-agent, line 394: [🏗️ **design**] `freezeOnBlur={false}` sits after `{...screenNativeProps}`, so it is the one screen prop `unstable_nativeProps` cannot override. The tests right next to the new one assert the opposite contract for everything else (`lets raw screen props override expo-router optional props`, `lets raw screen props override wired handlers`).

Is that intentional? If the escape hatch should stay an escape hatch, move this line above the spread. If freezing must stay off unconditionally, a short comment here would explain why this prop is exempt.

packages/expo-router/src/react-navigation/native-stack/views/__tests__/NativeStackView.unstable-native-props.test.native.tsx
```diff
@@ -58,6 +58,12 @@ describe('unstable_nativeProps', () => {
     expect(props.gestureEnabled).toBe(false);
   });
 
+  it('disables screen freezing', () => {
```
jakub-agent: [💡 suggestion] This test passes `freezeOnBlur` as a plain screen option, not through `unstable_nativeProps`, so it does not belong in this `describe` block. It also leaves the case that the prop ordering actually decides untested — `unstable_nativeProps: { freezeOnBlur: true }`. Adding that assertion would pin down the intended behavior discussed above.

packages/expo-router/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### 🛠 Breaking changes
 
+- Remove freezeOnBlur prop ([#49163](https://github.com/expo/expo/pull/49163) by [@Ubax](https://github.com/Ubax))
```
jakub-agent: [💡 suggestion] The prop is not removed — it stays in the public option types as a deprecated no-op. Suggest wording that matches what users observe, e.g. "`freezeOnBlur` no longer has any effect; Expo Router navigators never freeze blurred screens, and `enableFreeze()` from `react-native-screens` no longer affects them." The `enableFreeze()` part matters most: apps relying on it silently lose the optimization.

packages/expo-router/src/react-navigation/bottom-tabs/views/ScreenFallback.tsx
```diff
@@ -35,7 +33,9 @@ export const MaybeScreenContainer = ({
 
 export function MaybeScreen({ enabled, active, ...rest }: ViewProps & Props) {
   if (Screens?.screensEnabled?.()) {
-    return <Screens.Screen enabled={enabled} activityState={active} {...rest} />;
+    return (
+      <Screens.Screen {...rest} enabled={enabled} activityState={active} freezeOnBlur={false} />
```
jakub-agent, line 37: [🐧 nit] Nit: the literal `false` is load-bearing here — omitting the prop would fall back to `freezeEnabled()` and freezing would come back for apps that call `enableFreeze()`. One short comment (here or in each of the four call sites) would stop a future cleanup from dropping it as redundant.

jakub-agent (approved):

packages/expo-router/src/react-navigation/native-stack/views/NativeStackView.native.tsx
Ubax: Move it above, so it can be overwritten

packages/expo-router/src/react-navigation/native-stack/views/__tests__/NativeStackView.unstable-native-props.test.native.tsx
Ubax: Add it

packages/expo-router/CHANGELOG.md
Ubax: Change, but make it short and plain english

kitten (approved):

PR #49164, opened 2026-08-20
[router] Add NavigationAwareActivity

Ubax:
# Why

Add `NavigationAwareActivity` which can be used to hide content when it's nested deep inside the stack or in hidden tab. 

# How

1. Wrap React `Activity` with state aware wrapper
2. Use hints from https://www.callstack.com/blog/keeping-screens-visible-with-activity-mode-hidden to keep the content visible
3. Expose `hideWhenNestedAtLevel` prop which allows users to customize the depth at which screens should be frozen

# Test Plan

1. CI
2. Router e2e

# Checklist

---

### DIFF
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -35,6 +35,7 @@
 
 ### 🎉 New features
 
+- Add unstable `NavigationAwareActivity` component. ([#49145](https://github.com/expo/expo/pull/49145) by [@Ubax](https://github.com/Ubax))
 - Add screen error boundaries ([#49174](https://github.com/expo/expo/pull/49174) by [@Ubax](https://github.com/Ubax))
 - Export `NativeStackView` for custom navigator implementations on web. ([#49204](https://github.com/expo/expo/pull/49204) by [@Ubax](https://github.com/Ubax))
 - Add global registry for routers. ([#48707](https://github.com/expo/expo/pull/48707) by [@Ubax](https://github.com/Ubax))
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -35,7 +35,7 @@
 
 ### 🎉 New features
 
-- Add unstable `NavigationAwareActivity` component. ([#49145](https://github.com/expo/expo/pull/49145) by [@Ubax](https://github.com/Ubax))
+- Add unstable `NavigationAwareActivity` component. ([#49164](https://github.com/expo/expo/pull/49164) by [@Ubax](https://github.com/Ubax))
 - Add screen error boundaries ([#49174](https://github.com/expo/expo/pull/49174) by [@Ubax](https://github.com/Ubax))
 - Export `NativeStackView` for custom navigator implementations on web. ([#49204](https://github.com/expo/expo/pull/49204) by [@Ubax](https://github.com/Ubax))
 - Add global registry for routers. ([#48707](https://github.com/expo/expo/pull/48707) by [@Ubax](https://github.com/Ubax))
--- a/packages/expo-router/src/exports.ts
+++ b/packages/expo-router/src/exports.ts
@@ -28,6 +28,7 @@ export { useSitemap, type SitemapType } from './views/useSitemap';
 export type { ErrorBoundaryProps } from './views/Try';
 export { ErrorBoundary } from './views/ErrorBoundary';
 export { SuspenseFallback, type SuspenseFallbackProps } from './views/SuspenseFallback';
+export { NavigationAwareActivity } from './views/NavigationAwareActivity';
 export type { ScreenProps } from './useScreens';
 
 // Platform
--- a/packages/expo-router/src/views/ActivityContents.web.tsx
+++ b/packages/expo-router/src/views/ActivityContents.web.tsx
@@ -1,32 +1,31 @@
 'use client';
 import { Activity, type ActivityProps } from 'react';
 
-export function ActivityContents({ mode, children }: ActivityProps) {
-  const observeActivity = (element: HTMLDivElement | null) => {
-    if (element === null) {
-      return;
+function observeActivity(element: HTMLDivElement | null) {
+  if (element === null) {
+    return;
+  }
+
+  const activityChild = element.firstElementChild;
+  if (!(activityChild instanceof HTMLElement)) {
+    return;
+  }
+
+  const restoreDisplay = () => {
+    if (activityChild.style.display === 'none') {
+      activityChild.style.display = 'contents';
     }
-
-    const restoreDisplay = () => {
-      for (const child of element.children) {
-        if (child instanceof HTMLElement && child.style.display === 'none') {
-          child.style.display = 'contents';
-        }
-      }
-    };
-    const observer = new MutationObserver(restoreDisplay);
-    observer.observe(element, {
-      attributes: true,
-      attributeFilter: ['style'],
-      childList: true,
-      subtree: true,
-    });
-    restoreDisplay();
-
-    return () => observer.disconnect();
   };
+  const observer = new MutationObserver(restoreDisplay);
+  observer.observe(activityChild, { attributes: true, attributeFilter: ['style'] });
+  restoreDisplay();
 
+  return () => observer.disconnect();
+}
+
+export function ActivityContents({ mode, children }: ActivityProps) {
   return (
+    // React detaches refs inside a hidden Activity, so observe it from an outer wrapper.
     <div ref={observeActivity} style={{ display: 'contents' }}>
       <Activity mode={mode}>
         <div style={{ display: 'contents' }}>{children}</div>
--- a/packages/expo-router/src/views/ActivityContents.web.tsx
+++ b/packages/expo-router/src/views/ActivityContents.web.tsx
@@ -1,0 +1,36 @@
+'use client';
+import { Activity, type ActivityProps } from 'react';
+
+export function ActivityContents({ mode, children }: ActivityProps) {
+  const observeActivity = (element: HTMLDivElement | null) => {
+    if (element === null) {
+      return;
+    }
+
+    const restoreDisplay = () => {
+      for (const child of element.children) {
+        if (child instanceof HTMLElement && child.style.display === 'none') {
+          child.style.display = 'contents';
+        }
+      }
+    };
+    const observer = new MutationObserver(restoreDisplay);
+    observer.observe(element, {
+      attributes: true,
+      attributeFilter: ['style'],
+      childList: true,
+      subtree: true,
+    });
+    restoreDisplay();
+
+    return () => observer.disconnect();
+  };
+
+  return (
+    <div ref={observeActivity} style={{ display: 'contents' }}>
+      <Activity mode={mode}>
+        <div style={{ display: 'contents' }}>{children}</div>
+      </Activity>
+    </div>
+  );
+}
--- a/packages/expo-router/src/views/NavigationAwareActivity.tsx
+++ b/packages/expo-router/src/views/NavigationAwareActivity.tsx
@@ -1,0 +1,64 @@
+'use client';
+import type { ReactNode } from 'react';
+
+import { useRouteNode } from '../Route';
+import { useIsFocused, useNavigationState, useRoute } from '../react-navigation/native';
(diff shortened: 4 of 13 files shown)


packages/expo-router/src/views/NavigationAwareActivity.tsx
```diff
@@ -0,0 +1,76 @@
...
+export function NavigationAwareActivity({
+  hideWhenNestedAtLevel = 2,
+  children,
+}: {
+  hideWhenNestedAtLevel?: number;
+  children: ReactNode;
+}) {
+  const routeNode = useRouteNode();
+
+  if (routeNode?.type !== 'route') {
```
jakub-agent, line 55: [🏗️ **design**] This rejects `_layout.tsx`, which blocks what looks like the most valuable case: wrapping a nested navigator so a whole subtree hides once the layout's own route is buried in the parent stack.

`useActivityMode` already works there. Inside a non-root layout, `useRoute()` and `useNavigationState()` resolve to the *parent* navigator (the layout's own navigator context is created inside `<Stack />`, below this point), so the depth is computed against the right stack. I checked with a `nested/_layout` that calls `useActivityMode()`: it reports `visible` → `visible` → `hidden` as two screens are pushed above `/nested` on the root stack. Inside the *root* layout it doesn't throw either.

So the restriction looks stricter than the implementation requires. Is it deliberate (layout support planned as a follow-up), or could the guard be narrowed to reject only the root layout?

packages/expo-router/src/exports.ts
```diff
@@ -28,6 +28,7 @@ export { useSitemap, type SitemapType } from './views/useSitemap';
 export type { ErrorBoundaryProps } from './views/Try';
 export { ErrorBoundary } from './views/ErrorBoundary';
 export { SuspenseFallback, type SuspenseFallbackProps } from './views/SuspenseFallback';
+export { NavigationAwareActivity } from './views/NavigationAwareActivity';
```
jakub-agent, line 31: [🏗️ **design**] The changelog entry and the JSDoc both call this unstable, but the export name carries no marker, unlike the package's other unstable exports (`unstable_navigationEvents`, `unstable_createStandardRouterNavigator`, `unstable_integrateWithRouter`).

Without the prefix, a later rename or removal is a breaking change on a name that reads as stable API, and the "may change or be removed in any release" note only reaches people who open the JSDoc. Was shipping it unprefixed a deliberate call?

packages/expo-router/src/views/NavigationAwareActivity.tsx
```diff
@@ -0,0 +1,76 @@
...
+export type ActivityMode = 'visible' | 'hidden';
+
+/**
+ * @internal
+ */
+export function useActivityMode(hideWhenNestedAtLevel = 2): ActivityMode {
+  const route = useRoute();
+  const isFocused = useIsFocused();
+  const screensAbove = useNavigationState((state) => {
+    if (state.type !== 'stack') {
```
jakub-agent, line 17: [💡 suggestion] For every navigator whose state isn't a stack — tabs, native tabs, drawer, split view, and custom routers built with `unstable_createStandardRouterNavigator` — `screensAbove` stays `0`, so with the default `hideWhenNestedAtLevel` of `2` the component mounts, wraps the tree in an `Activity`, and never hides. That is a silent no-op, and the tabs test in this PR encodes exactly that outcome (`threshold: 2` → everything `visible`).

The JSDoc does say to pass `1` for tabs and drawers, but nothing surfaces the mistake at runtime. Worth a dev-only warning when `state.type !== 'stack'` and `hideWhenNestedAtLevel > 1`?

packages/expo-router/src/views/NavigationAwareActivity.tsx
```diff
@@ -0,0 +1,76 @@
...
+      return 0;
+    }
+
+    const ownIndex = state.routes.findIndex(({ key }) => key === route.key);
+    return ownIndex < 0 ? 0 : Math.max(0, state.index - ownIndex);
+  });
+
+  return hideWhenNestedAtLevel <= 1
```
jakub-agent, line 25: [💡 suggestion] `<= 1` quietly absorbs `0` and negative values into the focus-based branch — I checked `0` and `-5`, and both behave exactly like `1`. Fractional values fall through to the comparison branch, so `1.5` behaves like `2`.

Since this is a public prop, `hideWhenNestedAtLevel={0}` reading as "always hidden" but behaving as "hide when unfocused" is an easy trap. Either document that anything `<= 1` means "hide when unfocused", or normalise/reject out-of-range values.

packages/expo-router/src/views/ActivityContents.web.tsx
```diff
@@ -0,0 +1,36 @@
+'use client';
+import { Activity, type ActivityProps } from 'react';
+
+export function ActivityContents({ mode, children }: ActivityProps) {
+  const observeActivity = (element: HTMLDivElement | null) => {
```
jakub-agent: [💡 suggestion] Two notes on the observer setup:

1. `observeActivity` is recreated on every render, so React detaches and reattaches the ref each time — tearing down and rebuilding the `MutationObserver` on every re-render of the screen. The callback captures nothing from the component scope, so it can be hoisted to module scope for a stable ref.
2. `subtree: true` + `childList: true` makes the callback fire for every `style` attribute change anywhere in the screen's DOM subtree (any inline-style animation, any child insertion), while `restoreDisplay` only ever inspects `element.children`. Observing the inner `div` directly — `element.firstElementChild`, which React keeps stable across mode toggles — with just `{ attributes: true, attributeFilter: ['style'] }` would be equivalent and much cheaper.

Related: is the outer `div` there because React detaches refs inside a hidden `Activity`? If so, a one-line comment would stop someone "simplifying" the ref onto the inner div later.

packages/expo-router/src/views/__tests__/ActivityContents.test.web.tsx
```diff
@@ -0,0 +1,27 @@
...
+      <span>content</span>
+    </ActivityContents>
+  );
+  const outer = result.container.querySelector<HTMLDivElement>('div')!;
+  const activityChild = outer.querySelector<HTMLDivElement>('div')!;
+
+  expect(outer.style.display).toBe('contents');
+  expect(activityChild.style.display).toBe('contents');
+
+  activityChild.style.display = 'none';
```
jakub-agent, line 40: [💡 suggestion] Setting `display` by hand skips the part that actually matters. When the `Activity` really hides, React writes `display: none !important` — I confirmed the inline style becomes `"display: none !important;"` with `getPropertyPriority('display') === 'important'` — and this component only works because assigning `child.style.display` clears the `important` priority.

A `rerender` with `mode="hidden"` asserting the child ends back at `contents` would pin that interaction, so a future React change (a class, `visibility`, `content-visibility`) fails the test instead of silently breaking the feature. I verified such a test passes today.

packages/expo-router/src/views/__tests__/ActivityContents.test.ios.tsx
```diff
@@ -0,0 +1,19 @@
...
+    .spyOn(NativeComponentRegistry, 'get')
+    // The test substitutes a regular host view for the registry's host component.
+    .mockImplementation(() => View as ReturnType<typeof NativeComponentRegistry.get>);
+
+  require('../ActivityContents');
+
+  expect(get).toHaveBeenCalledTimes(1);
+  expect(get.mock.calls[0]![0]).toBe('ExpoRouterActivityContents');
+
+  const config = get.mock.calls[0]![1]();
```
jakub-agent, line 14: [🐧 nit] This asserts the shape of the view config rather than the behaviour it exists for, so it would keep passing if the renderer stopped hiding instances this way.

The coupling worth recording is `cloneHiddenInstance` in the Fabric renderer: it builds the hide payload with `createAttributePayload({ style: { display: 'none' } }, instance.canonical.viewConfig.validAttributes)`, which is exactly why the `process` override converts a hide into `display: 'contents'`. A short comment naming that would make the dependency discoverable.

packages/expo-router/src/views/NavigationAwareActivity.tsx
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  return (
+    <NavigationAwareScreenActivity hideWhenNestedAtLevel={hideWhenNestedAtLevel}>
+      {children}
+    </NavigationAwareScreenActivity>
+  );
+}
+
+function NavigationAwareScreenActivity({
```
jakub-agent, line 66: [🐧 nit] Is the extra component needed? `useRouteNode()` and the throw both run before any other hook, so `useActivityMode(hideWhenNestedAtLevel)` could be called directly in `NavigationAwareActivity`: the nicer error still wins over whatever `useRoute`/`useNavigationState` would raise, and `routeNode.type` can't change for a fixed position in the tree, so the hook count can't shift between renders of the same instance.

If there is a reason (React Compiler, planned reuse), a one-line comment would save the next reader the same detour.

packages/expo-router/CHANGELOG.md
```diff
@@ -35,6 +35,7 @@
 
 ### 🎉 New features
 
+- Add unstable `NavigationAwareActivity` component. ([#49145](https://github.com/expo/expo/pull/49145) by [@Ubax](https://github.com/Ubax))
```
jakub-agent: [🐧 nit] `#49145` doesn't resolve to a PR in `expo/expo`. This PR is #49164 (the stack's parent, the `freezeOnBlur` removal, is #49163).

jakub-agent (approved):

PR 2/2 in the stack on top of #49163 (`freezeOnBlur` removal). The mechanism holds up under inspection:

- **Native:** `cloneHiddenInstance` in the Fabric renderer builds the hide payload with `createAttributePayload({ style: { display: 'none' } }, viewConfig.validAttributes)`, so the `process: () => 'contents'` override on the custom view config does turn every hide into `display: 'contents'`. Registering `uiViewClassName: 'RCTView'` means no native component is needed — the renderer uses `uiViewClassName` for the native node name. `getNativeComponentAttributes` returns `null` (it does not throw) for an unknown name, so the static-config fallback in `NativeComponentRegistry.get` is safe on both bridgeless and bridge runtimes.
- **Web:** verified locally that React commits `display: none !important` on the inner div when the `Activity` hides, and that the observer's `child.style.display = 'contents'` assignment clears the `important` priority, so the content stays laid out. The observer callback is a microtask, so it lands before paint.

All 13 new tests pass, `tsc --noEmit` on `expo-router` is clean (the `@ts-expect-error` is genuinely needed), and the depth math behaves as documented for stacks, tabs and preloaded routes.

Nothing blocking. The main things worth a reply:

1. The `routeNode.type !== 'route'` guard rejects `_layout.tsx`, which blocks wrapping a nested navigator — and `useActivityMode` works correctly there (verified). Deliberate?
2. The export has no `unstable_` prefix even though the changelog and JSDoc call it unstable, unlike the package's other unstable exports.
3. The changelog links to #49145, which doesn't exist.

Smaller notes on the web observer's scope/stability and on the two `ActivityContents` tests asserting mechanism rather than behaviour. One convention note not worth an inline comment: `packages/expo-router/AGENTS.md` asks for RSC tests in `__rsc_tests__/` when adding components — `views/__rsc_tests__/views.test.tsx` wasn't extended, though `SuspenseFallback` skipped it too, so this may not apply to `'use client'` views.

packages/expo-router/src/views/ActivityContents.web.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+  const restoreDisplay = () => {
+    if (activityChild.style.display === 'none') {
+      activityChild.style.display = 'contents';
+    }
+  };
+  const observer = new MutationObserver(restoreDisplay);
+  observer.observe(activityChild, { attributes: true, attributeFilter: ['style'] });
+  restoreDisplay();
+
```
kitten, line 22: Feels potentially a bit sketchy 😅 I guess it makes sense, but it's potentially calling for maybe reaching out to the React team and asking whether there's any smarter case-by-case handling we should be applying here, since a `MutationObserver` on `ref` to catch this change just feels a bit sketchy

kitten (approved):

packages/expo-router/src/views/ActivityContents.web.tsx
Ubax: I'll make a note to do that later 👍

PR #49174, opened 2026-08-20
[router] Add screen error boundaries

Ubax:
# Why

https://linear.app/expo/issue/ENG-26092/implement-an-api-for-customizing-errorboundary-from-layout

Allow users to customize error boundary once for each screen within navigator.

# How

1. Add context which passes error boundary down to its screens
2. Use the following order when resolving the boundary:
   - Direct `export ErrorBoundary` from the screen
   - `<Stack/NativeTabs/Navigator.Screen unstable_errorBoundary`
   - `<Stack/NativeTabs/Navigator unstable_screenErrorBoundary`
   - `unstable_settings.screenErrorBoundary` in `_layout.tsx`

# Test Plan

1. CI
2. Router-e2e

# Checklist


packages/expo-router/src/__tests__/ScreenErrorBoundary.test.ios.tsx
```diff
@@ -0,0 +1,67 @@
...
+      unstable_settings: { screenErrorBoundary: LayoutBoundary },
+    },
+    index: ThrowingRoute,
+  });
+
+  expect(screen.getByTestId('navigator-boundary')).toBeOnTheScreen();
+  expect(screen.queryByTestId('layout-boundary')).toBeNull();
+});
+
+it('uses a screen boundary before navigator and layout boundaries', async () => {
```
Ubax: Add one more test for `unstable_settings` inside `index`

packages/expo-router/src/layouts/withLayoutContext.tsx
```diff
@@ -37,12 +38,18 @@ export function useFilterScreenChildren<
...
     const guardedRedirects: GuardedRedirects = new Map();
 
     function flattenChild(child: ReactNode, exclude = false, redirectTo?: Href) {
       if (isScreen(child, contextKey)) {
-        screens.push(child.props as ScreenProps<TOptions, TState, TEventMap> & { name: string });
+        screens.push(
+          child.props as ScreenProps<TOptions, TState, TEventMap> & {
+            name: string;
+          }
+        );
```
Ubax: What caused this changes?

packages/expo-router/src/layouts/withLayoutContext.tsx
```diff
@@ -208,6 +217,8 @@ export function withLayoutContext<
   ) as ForwardRefExoticComponent<
     PropsWithoutRef<PickPartial<ComponentProps<T>, 'children'>> & RefAttributes<unknown>
   > & {
+    /** A component to render when any screen in this navigator throws an error. */
```
Ubax:
```suggestion
```

packages/expo-router/src/native-tabs/types.ts
```diff
@@ -296,6 +297,8 @@ export type NativeTabsBlurEffect = (typeof SUPPORTED_BLUR_EFFECTS)[number];
 
 export interface NativeTabsProps extends PropsWithChildren {
   // #region common props
+  /** A component to render when any tab screen throws an error. */
```
Ubax: This should mention that this will be a boundary for individual screen - not the whole `_layout`

packages/expo-router/src/views/Navigator.tsx
```diff
@@ -32,6 +38,8 @@ export type NavigatorProps<T extends UseNavigationBuilderRouter> = {
   children?: UseNavigationBuilderOptions['children'];
   router?: T;
   routerOptions?: Omit<Parameters<T>[0], 'initialRouteName'>;
+  /** A component to render when any screen in this navigator throws an error. */
```
Ubax: Same here

packages/expo-router/src/layouts/withLayoutContext.tsx
```diff
@@ -196,7 +203,9 @@ export function withLayoutContext<
       return (
         <IsWithinLayoutContext value>
           <GuardContextProvider node={node} guardedRedirects={guardedRedirects}>
-            <Nav {...props} id={contextKey} ref={ref} children={sorted} />
+            <NavigatorScreenErrorBoundaryContext value={screenErrorBoundary}>
+              <Nav {...props} id={contextKey} ref={ref} children={sorted} />
+            </NavigatorScreenErrorBoundaryContext>
```
Ubax: Can we do it on `unstable_integrateWithRouter` level instead?

packages/expo-router/src/Route.tsx
```diff
@@ -8,7 +8,11 @@ import { sortRoutesWithInitial, sortRoutes } from './sortRoutes';
 import type { SuspenseFallbackProps } from './views/SuspenseFallback';
 import type { ErrorBoundaryProps } from './views/Try';
 
-export type DynamicConvention = { name: string; deep: boolean; notFound?: boolean };
+export type DynamicConvention = {
+  name: string;
+  deep: boolean;
+  notFound?: boolean;
+};
```
Ubax: Why this changes?

packages/expo-router/src/Route.tsx
```diff
@@ -70,6 +74,14 @@ const CurrentRouteContext = createContext<RouteNode | null>(null);
...
   ComponentType<SuspenseFallbackProps> | undefined
 >(undefined);
+/** This context allows a `_layout.tsx` to provide an error boundary for its child routes. */
+export const LayoutScreenErrorBoundaryContext = createContext<
+  ComponentType<ErrorBoundaryProps> | undefined
+>(undefined);
+/** This context carries the error boundary configured by the current navigator. */
+export const NavigatorScreenErrorBoundaryContext = createContext<
+  ComponentType<ErrorBoundaryProps> | undefined
+>(undefined);
```
Ubax: Why are both needed?

ENG-26092-preview.md
Ubax, line 1: Remove this file

packages/expo-router/src/__tests__/ScreenErrorBoundary.test.ios.tsx
jakub-agent: Added coverage for `unstable_settings.screenErrorBoundary` on an `index` route in a584e1db977.

packages/expo-router/src/layouts/withLayoutContext.tsx
jakub-agent: This was incidental formatting from the initial change. Removed in a584e1db977.

packages/expo-router/src/layouts/withLayoutContext.tsx
jakub-agent: Removed the extra API surface from `withLayoutContext` in a584e1db977.

packages/expo-router/src/native-tabs/types.ts
jakub-agent: Updated the NativeTabs documentation to clarify that this applies to an individual tab screen in a584e1db977.

packages/expo-router/src/views/Navigator.tsx
jakub-agent: Updated the Navigator documentation with the same individual-screen wording in a584e1db977.

packages/expo-router/src/layouts/withLayoutContext.tsx
jakub-agent: Moved standard navigator handling to `unstable_integrateWithRouter` in a584e1db977. Direct `Navigator` support remains local to that component.

packages/expo-router/src/Route.tsx
jakub-agent: The contexts were introduced for the initial implementation. They are now a single inherited `ScreenErrorBoundaryContext` in a584e1db977.

packages/expo-router/src/Route.tsx
jakub-agent: Consolidated layout and navigator contexts into one inherited context in a584e1db977, which also preserves the requested precedence.

ENG-26092-preview.md
jakub-agent: Removed in a584e1db977.

packages/expo-router/src/__tests__/ScreenErrorBoundary.test.ios.tsx
```diff
@@ -0,0 +1,82 @@
...
+      default: () => <Stack />,
+      unstable_settings: { screenErrorBoundary: LayoutBoundary },
+    },
+    index: ThrowingRoute,
+  });
+
+  expect(screen.getByTestId('layout-boundary')).toBeOnTheScreen();
+});
+
+it('uses a route screen error boundary configured with unstable settings', async () => {
```
Ubax: Add one more case for when all four boundaries are specified to ensure the order

packages/expo-router/src/standard-navigation/index.tsx
```diff
@@ -57,7 +63,14 @@ type StandardRouterNavigatorComponent<
...
     ComponentType<
-      StandardRouterNavigatorProps<State, NavigatorOptions, EventMap, NavigatorProps, RouterOptions>
+      StandardRouterNavigatorProps<
+        State,
+        NavigatorOptions,
+        EventMap,
+        NavigatorProps,
+        RouterOptions
+      > &
+        ScreenErrorBoundaryProps
```
Ubax: Can this be added directly to `StandardRouterNavigatorProps` so that every navigator created using the integration has the boundary integrated out of the box?

I would like every `N = unsable_inte....(..)` to have:
- `<N screenErrorBoundary={..}>`
- `<N.Screen/Trigger errorBoundary={..}>`

packages/expo-router/src/useScreens.tsx
```diff
@@ -190,20 +198,34 @@ export function useSortedScreens<
...
+
+  if (ErrorBoundary || screenErrorBoundary) {
     const Wrapped = React.forwardRef((props: any, ref: any) => {
-      const children = React.createElement(component.default || EmptyRoute, {
+      let children = React.createElement(component.default || EmptyRoute, {
         ...props,
         ref,
       });
-      return <Try catch={ErrorBoundary}>{children}</Try>;
+      if (ErrorBoundary) {
```
Ubax, line 225: Add a check for `if(process.env..... && ErrorBoundary && screenErrorBoundary && value.tyepe !== 'layout') { warn`

packages/expo-router/src/useScreens.tsx
```diff
@@ -312,7 +335,6 @@ export function getQualifiedRouteComponent(value: RouteNode) {
       value.type === 'layout'
         ? (LayoutSuspenseFallback ?? InheritedSuspenseFallback)
         : InheritedSuspenseFallback;
-
     if (isFocused && !isGuarded) {
```
Ubax, line 339: Add it back

packages/expo-router/src/useScreens.tsx
```diff
@@ -385,12 +407,23 @@ export function getQualifiedRouteComponent(value: RouteNode) {
...
-                segment={value.route}
-              />
+              {ScreenErrorBoundary ? (
+                <Try catch={ScreenErrorBoundary}>
+                  <WrappedScreenComponent
+                    {...props}
+                    // Expose the template segment path, e.g. `(home)`, `[foo]`, `index`
+                    // the intention is to make it possible to deduce shared routes.
+                    segment={value.route}
+                  />
```
Ubax: Extract to

```tsx
const screenComponent = (<WrappedScreenComponent
                    {...props}
                    // Expose the template segment path, e.g. `(home)`, `[foo]`, `index`
                    // the intention is to make it possible to deduce shared routes.
                    segment={value.route}
                  />);  
```

packages/expo-router/CHANGELOG.md
```diff
@@ -32,6 +32,7 @@
 
 ### 🎉 New features
 
+- [api] Add screen error boundaries ([#49174](https://github.com/expo/expo/pull/49174) by [@Ubax](https://github.com/Ubax))
```
```suggestion
- Add screen error boundaries ([#49174](https://github.com/expo/expo/pull/49174) by [@Ubax](https://github.com/Ubax))
```

packages/expo-router/src/__tests__/ScreenErrorBoundary.test.ios.tsx
jakub-agent: Added a combined precedence test covering the route setting, Screen, navigator, and layout boundaries in 31674b2d955.

packages/expo-router/src/standard-navigation/index.tsx
jakub-agent: Moved screenErrorBoundary into StandardRouterNavigatorProps in 31674b2d955, so every navigator returned by unstable_integrateWithRouter exposes it while retaining the inherited .Screen errorBoundary prop.

packages/expo-router/src/useScreens.tsx
jakub-agent: Added the development warning for a route ErrorBoundary plus unstable_settings.screenErrorBoundary conflict in 31674b2d955. The exported ErrorBoundary remains authoritative.

packages/expo-router/src/useScreens.tsx
jakub-agent: Restored the separating blank line in 31674b2d955.

packages/expo-router/src/useScreens.tsx
jakub-agent: Extracted the shared screen JSX into screenComponent in 31674b2d955.

packages/expo-router/CHANGELOG.md
jakub-agent: Updated the changelog entry to match the suggested wording in 31674b2d955.

packages/expo-router/src/useScreens.tsx
```diff
@@ -96,14 +100,16 @@ function getSortedChildren<
   routeSource: RouteSource;
 }[] {
   if (!order?.length) {
-    return children
-      .sort(sortRoutesWithInitial(initialRouteName))
-      .map((route) => ({ route, props: {}, routeSource: 'filesystem' as const }));
+    return children.sort(sortRoutesWithInitial(initialRouteName)).map((route) => ({
+      route,
+      props: {},
+      routeSource: 'filesystem' as const,
```
Ubax: Why this change

packages/expo-router/src/useScreens.tsx
```diff
@@ -152,9 +158,11 @@ function getSortedChildren<
 
   // Add any remaining children
   ordered.push(
-    ...entries
-      .sort(sortRoutesWithInitial(initialRouteName))
-      .map((route) => ({ route, props: {}, routeSource: 'filesystem' as const }))
+    ...entries.sort(sortRoutesWithInitial(initialRouteName)).map((route) => ({
```
Ubax: Why this change?

packages/expo-router/src/useScreens.tsx
```diff
@@ -190,20 +198,45 @@ export function useSortedScreens<
...
+  const screenErrorBoundary = unstable_settings?.screenErrorBoundary;
+
+  if (
+    process.env.NODE_ENV !== 'production' &&
+    ErrorBoundary &&
+    screenErrorBoundary &&
+    value.type !== 'layout'
+  ) {
+    console.warn(
+      `Route "${value.contextKey}" exports an ErrorBoundary, which takes precedence over unstable_settings.screenErrorBoundary.`
```
Ubax:
```suggestion
      `Route "${value.contextKey}" exports both ErrorBoundary and unstable_settings.screenErrorBoundary. Only use one in a single screen.`
```

packages/expo-router/src/native-tabs/types.ts
```diff
@@ -296,6 +297,8 @@ export type NativeTabsBlurEffect = (typeof SUPPORTED_BLUR_EFFECTS)[number];
 
 export interface NativeTabsProps extends PropsWithChildren {
   // #region common props
+  /** A component to render when an individual tab screen throws an error. */
+  screenErrorBoundary?: ComponentType<ErrorBoundaryProps>;
```
Ubax: Is this still needed if standard-navigation integration exposes it?

packages/expo-router/src/useScreens.tsx
jakub-agent: Reverted this incidental formatting change in e90e96e29f9.

packages/expo-router/src/useScreens.tsx
jakub-agent: Reverted this incidental formatting change in e90e96e29f9.

packages/expo-router/src/useScreens.tsx
jakub-agent: Applied the suggested warning wording in e90e96e29f9.

packages/expo-router/src/native-tabs/types.ts
jakub-agent: Removed the redundant NativeTabs declaration in e90e96e29f9. NativeTabs inherits screenErrorBoundary from the standard-navigation integration.

packages/expo-router/src/standard-navigation/index.tsx
```diff
@@ -233,6 +235,12 @@ export function unstable_integrateWithRouter<
         />
       </NavigationContent>
     );
+
+    return screenErrorBoundary ? (
+      <ScreenErrorBoundaryContext value={screenErrorBoundary}>{content}</ScreenErrorBoundaryContext>
+    ) : (
+      content
```
Ubax, line 244: Shouldn't we use `ScreenErrorBoundaryContext value={undefined}` here? So that error boundary is only applied for children of given navigator and not nested ones?

jakub-agent: Updated the provider to always set the context (including undefined) in 699acb487d4, so a navigator boundary is scoped to its own screens and does not leak into a nested navigator.

packages/expo-router/src/standard-navigation/types.ts
```diff
@@ -222,4 +223,7 @@ export type StandardRouterNavigatorProps<
   'initialRouteName'
 > &
   Omit<NavigatorProps, 'initialRouteName'> &
-  Omit<RouterOptions, 'initialRouteName'>;
+  Omit<RouterOptions, 'initialRouteName'> & {
+    /** A component to render when an individual screen in this navigator throws an error. */
+    screenErrorBoundary?: React.ComponentType<ErrorBoundaryProps>;
```
Ubax: Rename to `unstable_screenErrorBoundary`

packages/expo-router/src/layouts/stack-utils/StackScreen.tsx
```diff
@@ -54,6 +54,13 @@ export interface StackScreenProps extends PropsWithChildren {
    * Only supported when used inside a Layout component.
    */
   dangerouslySingular?: StackBaseScreenProps['dangerouslySingular'];
+
+  /**
+   * A component to render when this screen throws an error.
+   *
+   * Only supported when used inside a Layout component.
+   */
+  errorBoundary?: StackBaseScreenProps['errorBoundary'];
```
Ubax: unstable_

packages/expo-router/src/standard-navigation/types.ts
jakub-agent: Renamed this to unstable_screenErrorBoundary in ac859d7aaa4 and updated all navigator call sites and documentation.

packages/expo-router/src/layouts/stack-utils/StackScreen.tsx
jakub-agent: Renamed the individual Screen prop to unstable_errorBoundary in ac859d7aaa4, including the underlying screen types, call sites, tests, and docs.

packages/expo-router/src/__tests__/ScreenErrorBoundary.test.ios.tsx
```diff
@@ -0,0 +1,87 @@
...
+      default: ThrowingRoute,
+      unstable_settings: { screenErrorBoundary: RouteBoundary },
+    },
+  });
+
+  expect(screen.getByTestId('route-boundary')).toBeOnTheScreen();
+  expect(screen.queryByTestId('screen-boundary')).toBeNull();
+  expect(screen.queryByTestId('navigator-boundary')).toBeNull();
+  expect(screen.queryByTestId('layout-boundary')).toBeNull();
+});
```
Ubax, line 121: Please add one test for each existing navigator using for example:
`NativeTabs.Trigger unstable_errorBoundary`
`Tabs.Screen unstable_errorBoundary`
....

jakub-agent: Added coverage for Stack, Tabs, TopTabs, Drawer, and NativeTabs.Trigger in 5d5186357bd. The NativeTabs trigger now accepts unstable_errorBoundary, which is forwarded into the common screen-boundary handling.

packages/expo-router/src/Route.tsx
```diff
@@ -70,6 +70,18 @@ const CurrentRouteContext = createContext<RouteNode | null>(null);
...
+/** This context carries the error boundary configured by the nearest navigator. */
+export const NavigatorScreenErrorBoundaryContext = createContext<
+  ComponentType<ErrorBoundaryProps> | undefined
+>(undefined);
+/** This context carries the error boundary configured by the current screen. */
+export const ScreenErrorBoundaryContext = createContext<
+  ComponentType<ErrorBoundaryProps> | undefined
+>(undefined);
```
Ubax: Why did you add three separate contexts?

Just one is enough

packages/expo-router/src/__tests__/ScreenErrorBoundary.test.ios.tsx
```diff
@@ -0,0 +1,110 @@
...
+    _layout: () => (
+      <Navigator>
+        <Navigator.Screen name="index" unstable_errorBoundary={ScreenBoundary} />
+      </Navigator>
+    ),
+    index: ThrowingRoute,
+  });
+
+  expect(screen.getByTestId('screen-boundary')).toBeOnTheScreen();
+});
```
Ubax: Don't use each. I want to have types coverage as well. Four separate tests

packages/expo-router/src/__tests__/ScreenErrorBoundary.test.ios.tsx
jakub-agent: Follow-up: the Drawer case required an unmocked Reanimated hook in this shared test harness and caused check-packages to fail. I removed that harness-only case in 130a3dfd1b8; Stack, Tabs, TopTabs, and NativeTabs.Trigger remain covered, and the fresh CI run will verify the suite.

packages/expo-router/src/Route.tsx
jakub-agent: Consolidated to a single ScreenErrorBoundaryContext in 731e38fc5c0. Its structured value tracks layout, navigator, and screen boundaries, preserving the necessary precedence and scoping behavior.

packages/expo-router/src/__tests__/ScreenErrorBoundary.test.ios.tsx
jakub-agent: Replaced the parameterized test with separate Stack, Tabs, TopTabs, and NativeTabs tests in 731e38fc5c0, so each API is type-checked at its JSX call site.

packages/expo-router/src/useScreens.tsx
```diff
@@ -529,10 +580,26 @@ export function routeToScreen<
...
+    options,
+    getId,
+    unstable_errorBoundary,
+    ...props
+  }: Partial<ScreenProps<TOptions, TState, TEventMap>> = {},
   isGuarded?: boolean,
   routeSource?: RouteSource
 ) {
+  const QualifiedRouteComponent = getQualifiedRouteComponent(route);
+  const ScreenComponent = unstable_errorBoundary
```
Ubax: [🚨 **critical**] This wrapper is a new function identity on every `routeToScreen` call, and it is returned from `getComponent`. `SceneView` uses the returned value as a JSX element type, so a new identity makes React unmount and remount the entire screen subtree — all screen state is lost whenever the layout route re-renders (parent re-render, context change, layout state).

I verified this at this PR's head commit with a mount counter: a layout that re-renders once after mount causes a screen with `unstable_errorBoundary` to mount twice, while the same screen without the prop mounts once.

```tsx
function RerenderingLayout() {
  const [, setN] = React.useState(0);
  React.useEffect(() => { setN(1); }, []);
  return (
    <Stack>
      <Stack.Screen name="index" unstable_errorBoundary={Boundary} />
    </Stack>
  );
}
// mounts === 2 with the prop, 1 without
```

The existing `qualifiedStore` WeakMap exists precisely to keep this identity stable. Two possible fixes: cache the wrapped component per `(route, boundary)` pair, or avoid a new component type entirely by providing the context through the `Screen` element's `layout` prop (`useDescriptors` calls `screen.layout` as a render function, so its identity does not affect reconciliation). Please also add a regression test for this.

packages/expo-router/src/useScreens.tsx
```diff
@@ -529,10 +580,26 @@ export function routeToScreen<
...
+  {
+    options,
+    getId,
+    unstable_errorBoundary,
+    ...props
+  }: Partial<ScreenProps<TOptions, TState, TEventMap>> = {},
   isGuarded?: boolean,
   routeSource?: RouteSource
 ) {
+  const QualifiedRouteComponent = getQualifiedRouteComponent(route);
```
Ubax: [🏗️ **design**] This moves `getQualifiedRouteComponent(route)` from inside the `getComponent` closure to layout render time, unconditionally for every screen — including screens without `unstable_errorBoundary`. In sync import mode (the production default without async routes), `getQualifiedRouteComponent` calls `value.loadRoute()` immediately, so every sibling route module of a layout is now evaluated when the layout renders, instead of when React Navigation first renders that screen. For a stack, that loads modules for routes the user may never visit. Keep the call inside the `getComponent` closure.

packages/expo-router/src/useScreens.tsx
```diff
@@ -190,20 +207,45 @@ export function useSortedScreens<
...
       });
-      return <Try catch={ErrorBoundary}>{children}</Try>;
+      if (ErrorBoundary) {
+        children = <Try catch={ErrorBoundary}>{children}</Try>;
+      } else if (value.type !== 'layout' && screenErrorBoundary) {
+        children = <Try catch={screenErrorBoundary}>{children}</Try>;
+      }
+      if (value.type === 'layout') {
+        children = (
+          <ScreenErrorBoundaryContext value={screenErrorBoundary}>
```
Ubax: [🏗️ **design**] Every layout now provides this context, resetting it to `undefined` when the layout has no `screenErrorBoundary`. The consequence (verified with a test at this PR's head): when a screen inside a nested layout throws, an ancestor layout's boundary still catches the error — but through the `Try` wrapped around the *nested layout's* screen, so the whole nested navigator (its header, tab bar) unmounts and is replaced by the boundary. That works against the feature's stated goal of keeping navigator UI mounted.

If the reset was chosen deliberately (the "scope layout error boundaries to direct children" commit suggests so), consider the alternative: only provide the context when `screenErrorBoundary` is defined, so a nested layout without its own boundary inherits the ancestor's, and the error is caught at the deep screen while the nested navigator UI stays mounted. If the current behavior is intended, the docs need to state it (see comment there), and a test pinning the nested-layout behavior would help.

A smaller side effect of the unconditional wrapping: every layout in every app — with or without any boundary — now renders through a `forwardRef` named `ErrorBoundary(<contextKey>)`, which is misleading in React DevTools and error component stacks.

docs/pages/router/error-handling.mdx
```diff
@@ -78,6 +78,33 @@ function Route({ ErrorBoundary, Component }) {
...
+export default function Layout() {
+  return (
+    <Stack unstable_screenErrorBoundary={ScreenErrorBoundary}>
+      <Stack.Screen name="profile" unstable_errorBoundary={ScreenErrorBoundary} />
+    </Stack>
+  );
+}
+```
+
+An individual route's `ErrorBoundary` export takes precedence, followed by its `unstable_settings.screenErrorBoundary`, `Stack.Screen`'s `unstable_errorBoundary`, the navigator's `unstable_screenErrorBoundary`, and then the nearest layout's `unstable_settings.screenErrorBoundary`. Layout settings apply to descendant screens, while navigator boundaries apply only to that navigator's screens.
```
Ubax: [💡 suggestion] "Layout settings apply to descendant screens" is only partially accurate. In the current implementation, every layout resets the boundary context, so for a screen inside a nested layout the ancestor's boundary catches the error at the nested-layout level: the nested navigator's header and tab bar unmount too. That contradicts the section's opening promise of keeping navigator UI mounted. Once the intended scoping is settled (see the `useScreens.tsx` comment), state it precisely here — for example: "A layout's boundary applies to the screens of its own navigator; a nested layout does not inherit it."

packages/expo-router/src/useScreens.tsx
```diff
@@ -190,20 +207,45 @@ export function useSortedScreens<
...
+  const screenErrorBoundary = unstable_settings?.screenErrorBoundary;
+
+  if (
+    process.env.NODE_ENV !== 'production' &&
+    ErrorBoundary &&
+    screenErrorBoundary &&
+    value.type !== 'layout'
+  ) {
+    console.warn(
+      `Route "${value.contextKey}" exports both ErrorBoundary and unstable_settings.screenErrorBoundary. Only use one in a single screen.`
```
Ubax: [💡 suggestion] On a non-layout route, `unstable_settings.screenErrorBoundary` is exactly equivalent to `export const ErrorBoundary` — same wrapping, same position. Supporting both creates a redundant API surface plus this warning to police it. Is there a use case that `export const ErrorBoundary` cannot cover for a route? If not, consider only reading `screenErrorBoundary` from layout routes and warning that non-layout routes should use the existing export.

packages/expo-router/src/standard-navigation/index.tsx
```diff
@@ -173,7 +173,9 @@ export function unstable_integrateWithRouter<
     RouterOptions
   >;
 
-  function StandardRouterNavigator(props: NavPropsType) {
+  function StandardRouterNavigator(allProps: NavPropsType) {
+    const { unstable_screenErrorBoundary, ...rest } = allProps;
+    const props = rest as NavPropsType;
```
Ubax: [🐧 nit] The `rest as NavPropsType` cast silently claims the omitted key is still there. A short comment explaining why the cast is safe (the navigator never reads `unstable_screenErrorBoundary`) would help the next reader; same for the `componentProps as any` cast in `routeToScreen`.

apps/router-e2e/__e2e__/native-navigation/app/error-boundaries/route.tsx
```diff
@@ -0,0 +1,7 @@
+import { createErrorBoundary, ThrowingRoute } from '../../components/error-boundaries';
+
+export const unstable_settings = {
+  screenErrorBoundary: createErrorBoundary('route'),
```
jakub-agent: [🏗️ **design**] `route.tsx` is a route, not a layout, so `value.type !== 'layout'` in `fromImport` and this hits the dev warning added in this same PR (`"...only supported in layout routes; use export const ErrorBoundary instead"`). The boundary is then dropped, and because the parent `_layout.tsx` sets `unstable_screenErrorBoundary={NavigatorErrorBoundary}`, tapping "Route boundary" renders `Custom navigator error boundary` instead.

Rung 1 of the documented order (`ErrorBoundary` exported from the screen) has no example in the app right now. Did you mean `export const ErrorBoundary = createErrorBoundary('route')`?

packages/expo-router/src/native-tabs/types.ts
```diff
@@ -581,6 +582,8 @@ export interface NativeTabTriggerProps {
    * When used in a route it has no effect.
    */
   name?: string;
+  /** A component to render when this tab's screen throws an error. */
```
jakub-agent: [🏗️ **design**] The per-trigger boundary is wired up, but the navigator-level one is not reachable for native tabs: `NativeTabsProps extends PropsWithChildren` and gains no `unstable_screenErrorBoundary`, so `<NativeTabs unstable_screenErrorBoundary={B}>` is a type error even though `NativeTabsNavigatorWrapper` spreads props straight into the `withLayoutContext` wrapper that would consume it.

The PR description lists `<Stack/NativeTabs/Navigator unstable_screenErrorBoundary>` as supported — was adding it to `NativeTabsProps` intended and missed, or is native tabs deliberately trigger-only?

packages/expo-router/src/views/Navigator.tsx
```diff
@@ -79,16 +92,26 @@ export function Navigator<T extends UseNavigationBuilderRouter = typeof StackRou
...
       value={{
         ...navigation,
         contextKey,
         router,
       }}>
-      <GuardContextProvider node={node} guardedRedirects={guardedRedirects}>
-        {nonScreenChildren}
-      </GuardContextProvider>
+      {unstable_screenErrorBoundary ? (
+        <ScreenErrorBoundaryContext value={unstable_screenErrorBoundary}>
```
jakub-agent, line 105: [🏗️ **design**] The navigator-level prop is applied twice: once through `useSortedScreens(..., unstable_screenErrorBoundary)` (which reaches every direct child screen as `__screenErrorBoundary`) and again through this context provider. The same doubling exists in `SlotNavigator` and `StandardRouterNavigator`.

As far as I can tell the context half never wins: `BaseRoute` resolves `__screenErrorBoundary ?? inheritedScreenErrorBoundary`, so direct children always take the prop path, and any nested layout route immediately overwrites the context in `fromImport`. If there is a case where the context is the only path that works, a short comment would help; otherwise dropping it would leave one mechanism per rung and make the precedence easier to follow.

packages/expo-router/src/useScreens.tsx
```diff
@@ -183,27 +201,53 @@ export function useSortedScreens<
...
         ...props,
         ref,
       });
-      return <Try catch={ErrorBoundary}>{children}</Try>;
+      if (ErrorBoundary) {
+        children = <Try catch={ErrorBoundary}>{children}</Try>;
+      }
+      if (value.type === 'layout') {
+        children = (
+          <ScreenErrorBoundaryContext value={screenErrorBoundary}>
```
jakub-agent: [🏗️ **design**] This always overrides the context, including with `undefined` when a layout has no `unstable_settings.screenErrorBoundary`. So a boundary set on a parent stops at the first nested `_layout.tsx`, and errors in that nested layout's screens unmount the whole nested navigator instead of just the failing screen — which is the opposite of what the docs promise ("keeps navigator UI ... mounted when a screen throws").

The `nested-without` e2e case suggests this is deliberate. If so, the docs section should say that the setting applies one layout level deep and is not inherited; otherwise consider falling back to the inherited value.

packages/expo-router/src/useScreens.tsx
```diff
@@ -231,6 +275,35 @@ function fromImport(
...
+
+  let boundaries = screenErrorBoundaryStore.get(route);
+  if (!boundaries) {
+    boundaries = new WeakMap();
+    screenErrorBoundaryStore.set(route, boundaries);
+  }
+
+  let ScreenComponent = boundaries.get(ErrorBoundary);
+  if (!ScreenComponent) {
+    ScreenComponent = (props) => {
```
jakub-agent: [💡 suggestion] `SceneView` calls `screen.getComponent()` on every render, so this `WeakMap` is what keeps the screen component identity stable — good. The consequence is that the boundary identity now controls screen identity: `unstable_errorBoundary={({ error }) => <Text>{error.message}</Text>}` creates a new component every render of the layout, which produces a new `ScreenComponent`, which remounts the screen and drops all of its state.

Worth a dev warning when the reference changes, or at minimum a line in the docs and the prop's TSDoc saying the component must be defined at module scope.

Minor: this component has no `displayName`, so it shows up unnamed in component stacks while the neighbouring wrappers set one under `__DEV__`.

packages/expo-router/src/useScreens.tsx
```diff
@@ -183,27 +201,53 @@ export function useSortedScreens<
...
-  if (ErrorBoundary) {
+  const screenErrorBoundary = unstable_settings?.screenErrorBoundary;
+
+  if (process.env.NODE_ENV !== 'production' && screenErrorBoundary && value.type !== 'layout') {
+    console.warn(
+      `Route "${value.contextKey}" exports unstable_settings.screenErrorBoundary. This setting is only supported in layout routes; use export const ErrorBoundary instead.`
+    );
+  }
+
+  if (ErrorBoundary || value.type === 'layout') {
```
jakub-agent: [💡 suggestion] Adding `|| value.type === 'layout'` makes every layout route take this early return, which skips the dev-only guard below that maps an empty-object default export to `EmptyRoute`. `component.default || EmptyRoute` does not cover it, because `{}` is truthy — so a layout in that state now hits `React.createElement({})` and throws "Element type is invalid" instead of rendering the friendly placeholder.

The check could move above this branch so it still applies.

packages/expo-router/src/standard-navigation/index.tsx
jakub-agent, line 178: [🐧 nit] A short comment on why the prop is stripped would help — it is not obvious that leaving it in would send it through `partitionNavigatorProps` into `extraProps` and then onto `NavigatorContent` (where `NativeTabsContent`'s `Record<..., never>` guard would reject it).

The `as NavPropsType` cast also looks unnecessary: `unstable_screenErrorBoundary` is optional, so `rest` should be assignable to `NavPropsType` directly. If TS does need it, the repo convention is to explain the cast inline.

jakub-agent (commented):

The feature is coherent and the resolution order matches the documented list. `useSortedScreens` threads the navigator/screen prop into `__screenErrorBoundary`, `ScreenErrorBoundaryContext` carries the `unstable_settings` value from a layout, and `BaseRoute` resolves `__screenErrorBoundary ?? inherited`. Unit tests cover each precedence rung, and the `Try` wrapper sits inside `<Route>` so `retry` and router hooks still work in the boundary.

Main points below:

- The `route.tsx` case in the router-e2e app configures `unstable_settings.screenErrorBoundary` on a **non-layout** route. That is exactly the case the new dev warning in `fromImport` rejects, so the "Route boundary" example renders the *navigator* boundary and logs a warning. It looks like it was meant to be `export const ErrorBoundary` — currently nothing in the example app demonstrates rung 1 of the priority list.
- The PR description lists `<NativeTabs unstable_screenErrorBoundary>`, but `NativeTabsProps extends PropsWithChildren` never gained the prop, so it is a type error even though it would work at runtime.
- A few smaller design questions: the navigator-level context provider looks redundant with the `useSortedScreens` path, nested layouts silently reset an inherited boundary, and an inline `unstable_errorBoundary` remounts the screen on every render.

Note: I reviewed statically and did not run the test suite.

apps/router-e2e/__e2e__/native-navigation/app/error-boundaries/route.tsx
Ubax: export const ErrorBoundary = createErrorBoundary('route')

packages/expo-router/src/views/Navigator.tsx
Ubax: Drop it if tests succeed

packages/expo-router/src/native-tabs/types.ts
Ubax: Add it via standard nav props. Adjust the test in packages/expo-router/src/native-tabs/__tests__/NativeTabsView.test.ios.tsx

packages/expo-router/src/useScreens.tsx
Ubax: Document one layout level deep

packages/expo-router/src/useScreens.tsx
Ubax: Is there a better way to implement this then using global scope map?

packages/expo-router/src/standard-navigation/index.tsx
```diff
@@ -173,7 +173,11 @@ export function unstable_integrateWithRouter<
     RouterOptions
   >;
 
-  function StandardRouterNavigator(props: NavPropsType) {
+  function StandardRouterNavigator(allProps: NavPropsType) {
+    // `withLayoutContext` consumes this prop when it creates the route screens. Do not forward it
```
jakub-agent: [💡 suggestion] This comment describes the mechanism the follow-up commit removed — `withLayoutContext` is back to `useSortedScreens(processed ?? [], guardedRedirects)` and no longer looks at the prop. The only consumer now is the `ScreenErrorBoundaryContext` provider a few lines below.

Something like: "Consumed here as a context provider for the navigator's screens; keep it out of `NavigatorContent`, whose props may reject unknown fields."

packages/expo-router/src/__tests__/ScreenErrorBoundary.test.ios.tsx
```diff
@@ -0,0 +1,67 @@
...
+      unstable_settings: { screenErrorBoundary: LayoutBoundary },
+    },
+    index: ThrowingRoute,
+  });
+
+  expect(screen.getByTestId('navigator-boundary')).toBeOnTheScreen();
+  expect(screen.queryByTestId('layout-boundary')).toBeNull();
+});
+
+it('uses the route boundary before navigator and layout boundaries', async () => {
```
jakub-agent, line 102: [💡 suggestion] The three rungs are covered, but the scoping rule the docs now state — "A nested layout starts a new boundary scope" — has only e2e coverage. It is the behaviour most likely to flip silently, since it depends on `fromImport` passing `undefined` into the provider rather than skipping it.

A fourth case would pin it: a parent layout with `unstable_settings.screenErrorBoundary`, a nested `_layout` with none, and an assertion that the nested route's error surfaces at the parent boundary rather than a nested one.

packages/expo-router/src/Route.tsx
```diff
@@ -70,6 +70,10 @@ const CurrentRouteContext = createContext<RouteNode | null>(null);
 export const SuspenseFallbackContext = createContext<
   ComponentType<SuspenseFallbackProps> | undefined
 >(undefined);
+/** This context carries the error boundary configured by the current layout, navigator, or screen. */
```
jakub-agent: [🐧 nit] "or screen" is stale after the simplification — there is no per-screen configuration any more. Layout and navigator are the only two producers.

docs/pages/router/error-handling.mdx
```diff
@@ -78,6 +78,35 @@ function Route({ ErrorBoundary, Component }) {
...
+```tsx src/app/_layout.tsx
+import { Stack, type ErrorBoundaryProps } from 'expo-router';
+import { Text } from 'react-native';
+
+function ScreenErrorBoundary({ error, retry }: ErrorBoundaryProps) {
+  return <Text onPress={retry}>Try again: {error.message}</Text>;
+}
+
+export const unstable_settings = {
+  screenErrorBoundary: ScreenErrorBoundary,
```
jakub-agent, line 95: [🐧 nit] The example sets both `unstable_settings.screenErrorBoundary` and `unstable_screenErrorBoundary` to the same component, but the list directly below says the navigator prop wins — so as written the setting is dead code in the snippet. Showing one and mentioning the other in prose would read more clearly.

packages/expo-router/src/useScreens.tsx
```diff
@@ -215,23 +243,12 @@ function fromImport(
...
-      return { default: EmptyRoute, SuspenseFallback };
-    }
-  }
-
   return { default: component.default!, SuspenseFallback };
 }
 
 // TODO: Maybe there's a more React-y way to do this?
 // Without this store, the process enters a recursive loop.
 const qualifiedStore = new WeakMap<RouteNode, React.ComponentType<any>>();
```
jakub-agent, line 255: [🐧 nit] Stray whitespace-only change: the blank line between `qualifiedStore` and the JSDoc below was removed. Unrelated to the feature.

jakub-agent (approved):

Iteration 2, reviewing `2eeb4e33` (`Address error boundary review feedback` + `Simplify screen error boundary API`).

Every point from the previous round is resolved, and the second commit went further than suggested: the per-screen `unstable_errorBoundary` prop is gone entirely, so resolution is now a single context walk instead of two parallel mechanisms.

- `route.tsx` now uses `export const ErrorBoundary` — the example app covers all three rungs.
- `unstable_screenErrorBoundary` added to `NativeTabsProps`, with the tab test rewritten to the navigator level.
- The prop path through `useSortedScreens` / `getScreenComponent` / `screenErrorBoundaryStore` is removed. That answers the question about the global-scope map: dropping it is the better implementation. `BaseRoute` reads `use(ScreenErrorBoundaryContext)` directly, screen component identity no longer depends on the boundary reference, and the remount-on-inline-component hazard is gone with it.
- The `EmptyRoute` guard moved above the early return, so it applies to layouts again.
- `withLayoutContext`, `views/Screen.tsx` and `StackScreen.tsx` are back to their original state — no leftover `unstable_errorBoundary` anywhere in the tree.
- Docs updated to three rungs plus the one-layout-deep scoping note.

Remaining comments are all small. The one worth fixing before merge is the comment in `standard-navigation/index.tsx`, which describes the mechanism that the second commit deleted.

Note: reviewed statically; I did not run the test suite.

docs/pages/router/error-handling.mdx
Ubax: Add a comment above each to choose one of them

packages/expo-router/src/__tests__/ScreenErrorBoundary.test.ios.tsx
Ubax: Add a test

packages/expo-router/src/standard-navigation/index.tsx
Ubax:
```suggestion
```

Remove the comment

packages/expo-router/src/Route.tsx
Ubax:
```suggestion
/** This context carries the error boundary configured by the current layout or navigator. */
```

packages/expo-router/src/useScreens.tsx
Ubax: Add it back

packages/expo-router/src/__tests__/ScreenErrorBoundary.test.ios.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+it('uses the parent layout boundary for a nested layout without one', async () => {
+  const LayoutBoundary = boundary('layout-boundary');
+
+  await renderRouterAsync({
+    _layout: {
+      default: () => <Stack />,
+      unstable_settings: { screenErrorBoundary: LayoutBoundary },
+    },
+    'nested/_layout': () => <Stack />,
```
Ubax:
```suggestion
    'nested/_layout': () => <View testId="nested-layout"><Stack />,
```

And check that nested layout is null

docs/pages/router/error-handling.mdx
```diff
@@ -78,6 +78,52 @@ function Route({ ErrorBoundary, Component }) {
 
 When `ErrorBoundary` is not present, the error will be thrown to the nearest parent's `ErrorBoundary` and accepts [`error`](/versions/latest/sdk/router/#error) and [`retry`](/versions/latest/sdk/router/#retry) props.
 
+### Boundaries for screens in a layout
+
+A screen error boundary keeps navigator UI, such as headers and tab bars, mounted when a screen throws. Configure it on a layout or navigator:
```
amandeepmittal, line 83: ⚠️ **correctness** `mdx-components.md#callouts`

`unstable_screenErrorBoundary` and `unstable_settings.screenErrorBoundary` are unreleased and land in SDK 58, so a reader on SDK 57 gets nothing from this section. Add an availability callout, as the Suspense fallback section below already does.

```suggestion
> **important** Screen error boundaries are in [alpha](/more/release-statuses/#alpha) and available in SDK 58 and later. The API is subject to breaking changes.

A screen error boundary keeps navigator UI, such as headers and tab bars, mounted when a screen throws. Configure it on a layout or navigator:
```

docs/pages/router/error-handling.mdx
```diff
@@ -78,6 +78,52 @@ function Route({ ErrorBoundary, Component }) {
 
 When `ErrorBoundary` is not present, the error will be thrown to the nearest parent's `ErrorBoundary` and accepts [`error`](/versions/latest/sdk/router/#error) and [`retry`](/versions/latest/sdk/router/#retry) props.
 
+### Boundaries for screens in a layout
```
amandeepmittal: 💡 **nit** `links-and-headings.md#titles-and-headings`

"Boundaries" alone does not read as error boundaries in the sidebar or in a search result, so front-load the term the reader searches for.

```suggestion
### Error boundaries for screens in a layout
```

docs/pages/router/error-handling.mdx
```diff
@@ -78,6 +78,52 @@ function Route({ ErrorBoundary, Component }) {
...
+}
+```
+
+Expo Router applies error boundaries in the following order:
+
+1. `ErrorBoundary` exported from the screen
+2. Navigator-level prop (for example, `Stack`)
+3. `unstable_settings.screenErrorBoundary` declared in the **\_layout.tsx** file
+
+`unstable_settings.screenErrorBoundary` applies to the layout's direct child screens. A nested layout starts a new boundary scope; define the setting in that nested layout to handle its screens.
```
amandeepmittal: ⚠️ **correctness** `prose.md#core-principles`

"Starts a new boundary scope" hides what actually happens: a nested layout without the setting does not leave its screens unhandled. The error reaches the closest boundary above it, which unmounts the entire nested layout, and that is the outcome this section opens by promising to avoid.

```suggestion
`unstable_settings.screenErrorBoundary` applies to the layout's direct child screens. A nested layout does not inherit it. Without the setting, an error in a nested screen reaches the closest boundary above it. That boundary unmounts the whole nested layout, so declare the setting in each nested layout.
```

docs/pages/router/error-handling.mdx
```diff
@@ -78,6 +78,52 @@ function Route({ ErrorBoundary, Component }) {
...
+// Choose this to configure the boundary for the navigator's direct child screens.
+export default function Layout() {
+  return <Stack unstable_screenErrorBoundary={ScreenErrorBoundary} />;
+}
+```
+
+Expo Router applies error boundaries in the following order:
+
+1. `ErrorBoundary` exported from the screen
+2. Navigator-level prop (for example, `Stack`)
```
amandeepmittal: ✏️ **style** `prose.md#core-principles`

Items 1 and 3 name the exact API and this one does not, so the reader has to scroll back to the code block to find the prop name. Name the prop and the navigators that accept it.

docs/pages/router/error-handling.mdx
```diff
@@ -78,6 +78,52 @@ function Route({ ErrorBoundary, Component }) {
...
+export const unstable_settings = {
+  screenErrorBoundary: ScreenErrorBoundary,
+};
+
+export default function Layout() {
+  return <Stack />;
+}
+```
+
+Alternatively, configure the boundary on the navigator instead:
```
amandeepmittal: 💡 **nit** `prose.md#core-principles`

"Alternatively" and "instead" carry the same meaning, so drop one.

```suggestion
Alternatively, configure the boundary on the navigator:
```

docs/pages/router/error-handling.mdx
```diff
@@ -78,6 +78,52 @@ function Route({ ErrorBoundary, Component }) {
...
+
+```tsx src/app/_layout.tsx
+import { Stack, type ErrorBoundaryProps } from 'expo-router';
+import { Text } from 'react-native';
+
+function ScreenErrorBoundary({ error, retry }: ErrorBoundaryProps) {
+  return <Text onPress={retry}>Try again: {error.message}</Text>;
+}
+
+// Choose this to configure the boundary for the navigator's direct child screens.
```
amandeepmittal: ✏️ **style** `prose.md#core-principles`

Both code comments describe "direct child screens", so "Choose this" gives the reader no basis to choose between the two examples. The difference is that this prop overrides `unstable_settings.screenErrorBoundary` for the same screens, so say that.

```suggestion
// This prop overrides unstable_settings.screenErrorBoundary for the same screens.
```

packages/expo-router/src/native-tabs/types.ts
```diff
@@ -295,6 +296,8 @@ export const SUPPORTED_BLUR_EFFECTS = [
 export type NativeTabsBlurEffect = (typeof SUPPORTED_BLUR_EFFECTS)[number];
 
 export interface NativeTabsProps extends PropsWithChildren {
+  /** A component to render when an individual tab screen throws an error. */
+  unstable_screenErrorBoundary?: React.ComponentType<ErrorBoundaryProps>;
```
kitten, line 300: just noting for alignment: I think it'd be ideal to test this but maybe take this out of "unstable" before the SDK release. Reason being that if suspense/transitions are becoming more relevant, I'd assume that this would otherwise be a "missing piece" in the stable API

packages/expo-router/src/useScreens.tsx
```diff
@@ -190,20 +191,47 @@ export function useSortedScreens<
...
         ...props,
         ref,
       });
-      return <Try catch={ErrorBoundary}>{children}</Try>;
+      if (ErrorBoundary) {
+        children = <Try catch={ErrorBoundary}>{children}</Try>;
+      }
+      if (value.type === 'layout') {
+        children = (
+          <ScreenErrorBoundaryContext value={screenErrorBoundary}>
```
kitten: I think I'm a bit shaky on the details, but the mental model I have here is that this setting would be a "template" for the next layout, as in, the default value for `screenErrorBoundary` would be copied to the child layout (the exact implementation doesn't matter too much)

Unless I'm missing something this would be losing this value, which may not be intuitive?

The main problem here is that we're only really "saving" users a single re-declaration per parallel route on a layout, however, conceptually, the error boundary still belongs to screens (since it doesn't replace the navigator/layout) itself.

So, I think, over a shared export/wrapper I'm not sure whether that's the perfect choice here. Might be worth trying it out in a larger app, but this feels like a DX detail that'd need to maybe be validated?

kitten (approved):

packages/expo-router/src/native-tabs/types.ts
Ubax: I added it as unstable, because I want to cherry-pick this to 56 and 57. We can then mark it as stable in 58

docs/pages/router/error-handling.mdx
Ubax: I'll cherry-pick them back to 56 and 57

amandeepmittal (approved):

packages/expo-router/src/useScreens.tsx
Ubax: Adjusted the inheritance model so that by default every screen nested in layout will get the error boundary. User need to explicitly set it to a different one (or `null`) to change it

PR #49176, opened 2026-08-20
[router] Get linking route info from React scope instead of store.routeInfo

Ubax:
# Why

# How

# Test Plan

# Checklist

### DIFF
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -50,7 +50,6 @@
 
 ### 🐛 Bug fixes
 
-- Keep linking route information in React scope. ([#49176](https://github.com/expo/expo/pull/49176) by [@Ubax](https://github.com/Ubax))
 - Make layouts with explicitly declared screens honor `unstable_settings.initialRouteName` instead of declaration order, which can change deep-link back stacks. ([#48708](https://github.com/expo/expo/pull/48708) by [@Ubax](https://github.com/Ubax))
 - Prevent unfocused nested native tab navigators from redirecting global router state. ([#48257](https://github.com/expo/expo/pull/48257) by [@Ubax](https://github.com/Ubax))
 - Fixed `Tabs` and `TopTabs` (`expo-router/js-tabs`, `expo-router/js-top-tabs`) not being usable from RSC ([#48330](https://github.com/expo/expo/pull/48330) by [@Ubax](https://github.com/Ubax))
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -50,6 +50,7 @@
 
 ### 🐛 Bug fixes
 
+- Keep linking route information in React scope. ([#49176](https://github.com/expo/expo/pull/49176) by [@Ubax](https://github.com/Ubax))
 - Make layouts with explicitly declared screens honor `unstable_settings.initialRouteName` instead of declaration order, which can change deep-link back stacks. ([#48708](https://github.com/expo/expo/pull/48708) by [@Ubax](https://github.com/Ubax))
 - Prevent unfocused nested native tab navigators from redirecting global router state. ([#48257](https://github.com/expo/expo/pull/48257) by [@Ubax](https://github.com/Ubax))
 - Fixed `Tabs` and `TopTabs` (`expo-router/js-tabs`, `expo-router/js-top-tabs`) not being usable from RSC ([#48330](https://github.com/expo/expo/pull/48330) by [@Ubax](https://github.com/Ubax))
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -50,7 +50,6 @@
 
 ### 🐛 Bug fixes
 
-- Keep linking route information in React scope. ([#49176](https://github.com/expo/expo/pull/49176) by [@Ubax](https://github.com/Ubax))
 - Make layouts with explicitly declared screens honor `unstable_settings.initialRouteName` instead of declaration order, which can change deep-link back stacks. ([#48708](https://github.com/expo/expo/pull/48708) by [@Ubax](https://github.com/Ubax))
 - Prevent unfocused nested native tab navigators from redirecting global router state. ([#48257](https://github.com/expo/expo/pull/48257) by [@Ubax](https://github.com/Ubax))
 - Fixed `Tabs` and `TopTabs` (`expo-router/js-tabs`, `expo-router/js-top-tabs`) not being usable from RSC ([#48330](https://github.com/expo/expo/pull/48330) by [@Ubax](https://github.com/Ubax))
@@ -67,6 +66,7 @@
 
 ### 💡 Others
 
+- Keep linking route information in React scope. ([#49176](https://github.com/expo/expo/pull/49176) by [@Ubax](https://github.com/Ubax))
 - Remove the ignored `linking.enabled` option. ([#49103](https://github.com/expo/expo/pull/49103) by [@Ubax](https://github.com/Ubax))
 - Render only the focused tab route during the first render. ([#48618](https://github.com/expo/expo/pull/48618) by [@Ubax](https://github.com/Ubax))
 - Order JS Tabs, NativeTabs, headless tabs and Drawer by `routeNames` order ([#48374](https://github.com/expo/expo/pull/48374) by [@Ubax](https://github.com/Ubax))
--- a/packages/expo-router/src/fork/useLinking.native.ts
+++ b/packages/expo-router/src/fork/useLinking.native.ts
@@ -2,6 +2,7 @@ import * as ExpoLinking from 'expo-linking';
 import { type RefObject, useEffect, useCallback, useRef } from 'react';
 import { Linking, Platform } from 'react-native';
 
+import { useRouteInfo } from '../global-state/useRouteInfo';
 import {
   type LinkingOptions,
   getActionFromState as getActionFromStateDefault,
@@ -10,10 +11,13 @@ import {
   type ParamListBase,
 } from '../react-navigation/native';
 import { extractExpoPathFromURL } from './extractPathFromURL';
+import { getStateFromPath as getExpoStateFromPath } from './getStateFromPath';
 
 type ResultState = ReturnType<typeof getStateFromPathDefault>;
 
-type Options = LinkingOptions<ParamListBase>;
+type Options = Omit<LinkingOptions<ParamListBase>, 'getStateFromPath'> & {
+  getStateFromPath?: typeof getExpoStateFromPath;
+};
 
 const linkingHandlers: symbol[] = [];
 
@@ -49,8 +53,11 @@ export function useLinking(
         }
       };
     });
-  const getStateFromPath = options?.getStateFromPath ?? getStateFromPathDefault;
+  const getStateFromPath = (options?.getStateFromPath ??
+    getStateFromPathDefault) as typeof getExpoStateFromPath;
   const getActionFromState = options?.getActionFromState ?? getActionFromStateDefault;
+  const independent = useNavigationIndependentTree();
+  const { segments } = useRouteInfo();
   useEffect(() => {
     if (process.env.NODE_ENV === 'production') {
       return undefined;
@@ -93,6 +100,10 @@ export function useLinking(
   const getInitialURLRef = useRef(getInitialURL);
   const getStateFromPathRef = useRef(getStateFromPath);
   const getActionFromStateRef = useRef(getActionFromState);
+  const segmentsRef = useRef(segments);
+
+  // Keep stable URL listeners in sync with the latest navigation state.
+  segmentsRef.current = segments;
 
   useEffect(() => {
     enabledRef.current = enabled;
@@ -112,7 +123,9 @@ export function useLinking(
 
       const path = extractExpoPathFromURL(prefixesRef.current, url);
 
-      return path !== undefined ? getStateFromPathRef.current(path, configRef.current) : undefined;
+      return path !== undefined
+        ? getStateFromPathRef.current(path, configRef.current, segmentsRef.current)
+        : undefined;
     },
 
     []
--- a/packages/expo-router/src/fork/useLinking.native.ts
+++ b/packages/expo-router/src/fork/useLinking.native.ts
@@ -53,7 +53,6 @@ export function useLinking(
     });
   const getStateFromPath = options?.getStateFromPath ?? getExpoStateFromPath;
   const getActionFromState = options?.getActionFromState ?? getActionFromStateDefault;
-  const independent = useNavigationIndependentTree();
   const { segments } = useRouteInfo();
   useEffect(() => {
     if (process.env.NODE_ENV === 'production') {
--- a/packages/expo-router/src/fork/useLinking.native.ts
+++ b/packages/expo-router/src/fork/useLinking.native.ts
@@ -15,9 +15,7 @@ import { getStateFromPath as getExpoStateFromPath } from './getStateFromPath';
 
 type ResultState = ReturnType<typeof getStateFromPathDefault>;
 
-type Options = Omit<LinkingOptions<ParamListBase>, 'getStateFromPath'> & {
-  getStateFromPath?: typeof getExpoStateFromPath;
-};
+type Options = LinkingOptions<ParamListBase>;
 
 const linkingHandlers: symbol[] = [];
 
@@ -53,8 +51,7 @@ export function useLinking(
         }
       };
     });
-  const getStateFromPath: typeof getExpoStateFromPath =
-    options?.getStateFromPath ?? ((path, options) => getStateFromPathDefault(path, options));
+  const getStateFromPath = options?.getStateFromPath ?? getExpoStateFromPath;
   const getActionFromState = options?.getActionFromState ?? getActionFromStateDefault;
   const independent = useNavigationIndependentTree();
   const { segments } = useRouteInfo();
--- a/packages/expo-router/src/fork/useLinking.native.ts
+++ b/packages/expo-router/src/fork/useLinking.native.ts
@@ -53,8 +53,8 @@ export function useLinking(
         }
       };
     });
-  const getStateFromPath: typeof getExpoStateFromPath =
-    options?.getStateFromPath ?? ((path, options) => getStateFromPathDefault(path, options));
+  const getStateFromPath = (options?.getStateFromPath ??
+    getStateFromPathDefault) as typeof getExpoStateFromPath;
   const getActionFromState = options?.getActionFromState ?? getActionFromStateDefault;
   const independent = useNavigationIndependentTree();
   const { segments } = useRouteInfo();
@@ -102,8 +102,7 @@ export function useLinking(
   const getActionFromStateRef = useRef(getActionFromState);
   const segmentsRef = useRef(segments);
 
-  // Linking listeners stay subscribed across navigation updates, but relative URLs must be parsed
-  // against the route that was current when the URL was received.
+  // Keep stable URL listeners in sync with the latest navigation state.
   segmentsRef.current = segments;
 
   useEffect(() => {
--- a/packages/expo-router/src/fork/useLinking.native.ts
+++ b/packages/expo-router/src/fork/useLinking.native.ts
@@ -1,6 +1,5 @@
-import * as ExpoLinking from 'expo-linking';
-import { type RefObject, useEffect, useCallback, useRef } from 'react';
-import { Linking, Platform } from 'react-native';
+import { type RefObject, useEffect, useEffectEvent, useCallback, useRef } from 'react';
+import { Linking } from 'react-native';
 
 import { useRouteInfo } from '../global-state/useRouteInfo';
 import {
@@ -11,6 +10,7 @@ import {
   type ParamListBase,
 } from '../react-navigation/native';
 import { extractExpoPathFromURL } from './extractPathFromURL';
+import { getInitialURLWithTimeout } from './getInitialURLWithTimeout';
 import { getStateFromPath as getExpoStateFromPath } from './getStateFromPath';
 
 type ResultState = ReturnType<typeof getStateFromPathDefault>;
@@ -100,11 +100,6 @@ export function useLinking(
   const getInitialURLRef = useRef(getInitialURL);
   const getStateFromPathRef = useRef(getStateFromPath);
   const getActionFromStateRef = useRef(getActionFromState);
-  const segmentsRef = useRef(segments);
-
-  // Linking listeners stay subscribed across navigation updates, but relative URLs must be parsed
-  // against the route that was current when the URL was received.
-  segmentsRef.current = segments;
 
   useEffect(() => {
     enabledRef.current = enabled;
@@ -125,12 +120,13 @@ export function useLinking(
       const path = extractExpoPathFromURL(prefixesRef.current, url);
 
       return path !== undefined
-        ? getStateFromPathRef.current(path, configRef.current, segmentsRef.current)
+        ? getStateFromPathRef.current(path, configRef.current, segments)
         : undefined;
     },
 
-    []
+    [segments]
   );
+  const getStateFromURLInEffect = useEffectEvent(getStateFromURL);
 
   const getInitialState = useCallback(() => {
     let state: ResultState | undefined;
@@ -177,7 +173,7 @@ export function useLinking(
       }
 
       const navigation = ref.current;
-      const state = navigation ? getStateFromURL(url) : undefined;
+      const state = navigation ? getStateFromURLInEffect(url) : undefined;
 
       if (navigation && state) {
         // If the link were handled, it gets cleared in NavigationContainer
@@ -208,28 +204,9 @@ export function useLinking(
     };
 
     return subscribe(listener);
-  }, [enabled, getStateFromURL, onUnhandledLinking, prefixes, ref, subscribe]);
+  }, [enabled, onUnhandledLinking, prefixes, ref, subscribe]);
 
   return {
     getInitialState,
   };
-}
-
-export function getInitialURLWithTimeout(): string | null | Promise<string | null> {
-  if (typeof window === 'undefined') {
-    return '';
-  } else if (Platform.OS === 'ios') {
-    // Use the new Expo API for iOS. This has better support for App Clips and handoff.
-    return ExpoLinking.getLinkingURL();
-  }
-
-  return Promise.race([
-    // TODO: Phase this out in favor of expo-linking on Android.
-    Linking.getInitialURL(),
-    new Promise<null>((resolve) =>
-      // Timeout in 150ms if `getInitialState` doesn't resolve
-      // Workaround for https://github.com/facebook/react-native/issues/25675
-      setTimeout(() => resolve(null), 150)
-    ),
-  ]);
 }
--- a/packages/expo-router/src/fork/useLinking.native.ts
+++ b/packages/expo-router/src/fork/useLinking.native.ts
@@ -53,8 +53,8 @@ export function useLinking(
         }
       };
     });
-  const getStateFromPath = (options?.getStateFromPath ??
-    getStateFromPathDefault) as typeof getExpoStateFromPath;
+  const getStateFromPath: typeof getExpoStateFromPath =
+    options?.getStateFromPath ?? ((path, options) => getStateFromPathDefault(path, options));
   const getActionFromState = options?.getActionFromState ?? getActionFromStateDefault;
   const independent = useNavigationIndependentTree();
   const { segments } = useRouteInfo();
@@ -102,7 +102,8 @@ export function useLinking(
   const getActionFromStateRef = useRef(getActionFromState);
   const segmentsRef = useRef(segments);
 
-  // Keep stable URL listeners in sync with the latest navigation state.
+  // Linking listeners stay subscribed across navigation updates, but relative URLs must be parsed
+  // against the route that was current when the URL was received.
   segmentsRef.current = segments;
 
   useEffect(() => {
--- a/packages/expo-router/src/fork/useLinking.ts
+++ b/packages/expo-router/src/fork/useLinking.ts
@@ -1,5 +1,13 @@
 import isEqual from 'fast-deep-equal';
-import { type RefObject, useEffect, useState, useCallback, useRef, use } from 'react';
+import {
+  type RefObject,
+  useEffect,
+  useEffectEvent,
+  useState,
+  useCallback,
+  useRef,
+  use,
+} from 'react';
 
 import { ServerContext } from '../global-state/serverLocationContext';
 import { useExpoRouterStore } from '../global-state/storeContext';
@@ -135,11 +143,6 @@ export function useLinking(
   const getStateFromPathRef = useRef(getStateFromPath);
   const getPathFromStateRef = useRef(getPathFromState);
   const getActionFromStateRef = useRef(getActionFromState);
-  const segmentsRef = useRef(segments);
-
-  // Segments are navigation state, so keep this ref current during render. The
-  // linking listeners intentionally remain stable across navigation updates.
-  segmentsRef.current = segments;
 
   useEffect(() => {
     enabledRef.current = enabled;
@@ -148,6 +151,12 @@ export function useLinking(
     getPathFromStateRef.current = getPathFromState;
     getActionFromStateRef.current = getActionFromState;
   });
+
+  const getStateFromPathForCurrentSegments = useCallback(
+    (path: string) => getStateFromPathRef.current(path, configRef.current, segments),
+    [segments]
+  );
+  const getStateFromPathInEffect = useEffectEvent(getStateFromPathForCurrentSegments);
 
   const validateRoutesNotExistInRootState = useCallback(
     (state: ResultState) => {
@@ -182,7 +191,7 @@ export function useLinking(
         : undefined;
 
       if (path) {
-        value = getStateFromPathRef.current(path, configRef.current, segmentsRef.current);
+        value = getStateFromPathForCurrentSegments(path);
       }
 
       // If the link were handled, it gets cleared in NavigationContainer
@@ -200,7 +209,7 @@ export function useLinking(
 
     return thenable as PromiseLike<ResultState | undefined>;
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+  }, [getStateFromPathForCurrentSegments, onUnhandledLinking, server]);
 
   const previousIndexRef = useRef<number | undefined>(undefined);
   const previousStateRef = useRef<NavigationState | undefined>(undefined);
@@ -236,7 +245,7 @@ export function useLinking(
         return;
       }
 
-      const state = getStateFromPathRef.current(path, configRef.current, segmentsRef.current);
+      const state = getStateFromPathInEffect(path);
 
       // We should only dispatch an action when going forward
       // Otherwise the action will likely add items to history, which would mess things up
@@ -314,11 +323,7 @@ export function useLinking(
       // If the `route` object contains a `path`, use that path as long as `route.name` and `params` still match
       // This makes sure that we preserve the original URL for wildcard routes
       if (route?.path) {
-        const stateForPath = getStateFromPathRef.current(
-          route.path,
-          configRef.current,
-          segmentsRef.current
-        );
+        const stateForPath = getStateFromPathInEffect(route.path);
 
         if (stateForPath) {
           const focusedRoute = findFocusedRoute(stateForPath);
@@ -485,8 +490,4 @@ export function useLinking(
   return {
     getInitialState,
   };
-}
-
-export function getInitialURLWithTimeout(): string | null | Promise<string | null> {
-  return typeof window === 'undefined' ? '' : window.location.href;
 }
--- a/packages/expo-router/src/fork/useLinking.ts
+++ b/packages/expo-router/src/fork/useLinking.ts
@@ -86,8 +86,8 @@ export function useLinking(
 ) {
   const enabled = options !== undefined;
   const config = options?.config;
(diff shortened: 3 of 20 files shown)


packages/expo-router/src/fork/useLinking.native.ts
```diff
@@ -50,9 +54,11 @@ export function useLinking(
         }
       };
     });
-  const getStateFromPath = options?.getStateFromPath ?? getStateFromPathDefault;
+  const getStateFromPath = (options?.getStateFromPath ??
```
Ubax: Why is this cast needed?

packages/expo-router/src/fork/useLinking.native.ts
```diff
@@ -100,6 +106,10 @@ export function useLinking(
   const getInitialURLRef = useRef(getInitialURL);
   const getStateFromPathRef = useRef(getStateFromPath);
   const getActionFromStateRef = useRef(getActionFromState);
+  const segmentsRef = useRef(segments);
+
+  // Keep stable URL listeners in sync with the latest navigation state.
+  segmentsRef.current = segments;
```
Ubax: Elaborate on why this is needed

packages/expo-router/src/global-state/getNavigationAction.ts
```diff
@@ -33,15 +33,16 @@ export function getNavigateAction(
   }
   const rootState = navigationRef.getRootState();
 
-  href = resolveHrefStringWithSegments(href, store.getRouteInfo(), options);
+  const routeInfo = store.getRouteInfo();
+  href = resolveHrefStringWithSegments(href, routeInfo, options);
```
Ubax: Why is this change needed?

packages/expo-router/src/link/preview/useNextScreenId.ts
```diff
@@ -12,19 +13,30 @@ export function useNextScreenId(): [
   (href: Href) => void,
 ] {
   const router = useRouter();
+  const routeInfo = useRouteInfo();
   const { setOpenPreviewKey } = useLinkPreviewContext();
   const [internalNextScreenId, internalSetNextScreenId] = useState<string | undefined>();
   const currentHref = useRef<Href | undefined>(undefined);
+  const routeInfoRef = useRef(routeInfo);
+  routeInfoRef.current = routeInfo;
```
Ubax: Why is this ref needed?

packages/expo-router/src/getLinkingConfig.ts
```diff
@@ -113,13 +111,19 @@ export function getLinkingConfig(
...
             }
           } else if (initialUrl) {
             initialUrl = initialUrl.then((url) => {
               url = applyRedirects(url, redirects);
               if (url && typeof nativeLinking?.redirectSystemPath === 'function') {
-                return nativeLinking.redirectSystemPath({ path: url, initial: true });
+                return nativeLinking.redirectSystemPath({
+                  path: url,
+                  initial: true,
+                });
```
Ubax: Why this changes?

packages/expo-router/src/fork/useLinking.native.ts
jakub-agent: Removed the cast. The React Navigation fallback has a two-argument signature, so it now uses a typed adapter that intentionally ignores the Expo Router segments argument.

packages/expo-router/src/fork/useLinking.native.ts
jakub-agent: The URL listener is intentionally stable across navigation updates. This ref supplies the latest segment snapshot when a URL arrives, so relative paths are resolved against the route that was current at that moment without resubscribing on every navigation.

packages/expo-router/src/global-state/getNavigationAction.ts
jakub-agent: `routeInfo` is captured once so href resolution and navigation-state parsing use the same segment snapshot. That keeps a relative href consistent if router state changes between those two operations.

packages/expo-router/src/link/preview/useNextScreenId.ts
jakub-agent: The navigation-state listener is mounted once. Keeping route info in a ref lets its callback use the currently focused route when a prefetch changes navigation state, without tearing down and recreating the listener on every route update.

packages/expo-router/src/getLinkingConfig.ts
jakub-agent: Reverted this unrelated formatting change; the native-intent redirect behavior is unchanged.

packages/expo-router/src/global-state/getNavigationAction.ts
Ubax: Can you move it to become the argument of the function?
So that it does not use `store.`

jakub-agent: Done. `useImperativeApiEmitter` now reads route info in React scope and passes it through `routingQueue.run` to `getNavigateAction`, which no longer reads `store.getRouteInfo()`. Added coverage for the queue forwarding the route snapshot and for navigation parsing with caller-provided segments.

packages/expo-router/src/link/preview/useNextScreenId.ts
Ubax: `useEffectEvent` instead of the ref here

packages/expo-router/src/fork/useLinking.native.ts
Ubax: `useEffectEvent` here as well

packages/expo-router/src/fork/useLinking.ts
```diff
@@ -136,6 +142,11 @@ export function useLinking(
   const getStateFromPathRef = useRef(getStateFromPath);
   const getPathFromStateRef = useRef(getPathFromState);
   const getActionFromStateRef = useRef(getActionFromState);
+  const segmentsRef = useRef(segments);
```
Ubax: useEffectEvent

packages/expo-router/src/link/preview/useNextScreenId.ts
jakub-agent: Replaced `routeInfoRef` with `useEffectEvent`; the stable navigation-state subscription now reads the latest React route info directly.

packages/expo-router/src/fork/useLinking.native.ts
jakub-agent: Replaced the segment ref with an effect event. The subscription remains stable while event-time URL parsing uses the current React segments.

packages/expo-router/src/fork/useLinking.ts
jakub-agent: Applied the same pattern to web linking: the event callbacks use `useEffectEvent` for current segments while initial-state parsing stays synchronous.

packages/expo-router/CHANGELOG.md
```diff
@@ -47,6 +47,7 @@
 
 ### 🐛 Bug fixes
 
+- Keep linking route information in React scope. ([#49176](https://github.com/expo/expo/pull/49176) by [@Ubax](https://github.com/Ubax))
```
Ubax: Move to Others

jakub-agent: Moved the changelog entry to the Others section and pushed the update.

packages/expo-router/src/global-state/getNavigationAction.ts
```diff
@@ -17,7 +18,8 @@ export function getNavigateAction(
   type = 'NAVIGATE',
   withAnchor?: boolean,
   singular?: SingularOptions,
-  isPreviewNavigation?: boolean
+  isPreviewNavigation?: boolean,
+  routeInfo: UrlObject = defaultRouteInfo
```
jakub-agent: [🏗️ **design**] Please make `routeInfo` required instead of defaulting to `defaultRouteInfo`.

`defaultRouteInfo` has `segments: []`, so a caller that omits the argument does not fail — it silently resolves relative hrefs against the root. Before this PR the function always read the real route info via `store.getRouteInfo()`, so "forgot to pass route info" was not a reachable state.

Concretely: a future call site that omits the argument turns `router.push('../sibling')` from `/(tabs)/settings/profile` into a navigation to `/sibling` instead of `/(tabs)/settings/sibling`, with no error and no warning. `routeInfo.segments` is also passed straight to `getStateFromPath` on line 46, so the config sort tie-breaker silently degrades too.

There is exactly one production caller (`routingQueue.run`) and it always passes a real value, so the default only serves the old test call sites. Making it required moves the failure to `tsc` and forces those tests to state which route they navigate from.

packages/expo-router/src/getLinkingConfig.ts
```diff
@@ -130,8 +134,12 @@ export function getLinkingConfig(
       return initialUrl;
     },
     subscribe: subscribe(nativeLinking, redirects),
-    getStateFromPath: <ParamList extends object>(path: string, options?: Options<ParamList>) => {
-      return getStateFromPath(path, options, getRouteInfo().segments);
+    getStateFromPath: <ParamList extends object>(
+      path: string,
+      options?: Options<ParamList>,
+      previousSegments?: string[]
```
jakub-agent: [🏗️ **design**] This changes behavior for every caller that does not pass segments, not just the ones updated here.

Previously this wrapper closed over `getRouteInfo()` and injected the current segments unconditionally, so *any* consumer of `linking.getStateFromPath` got the tie-breaker in `expo.getRouteConfigSorter(previousSegments)`. Now it forwards whatever the caller supplies, and `src/react-navigation/native/useLinkBuilder.tsx:106` still calls it with two arguments:

```ts
const state = getStateFromPathHelper(href, options?.config);
```

So `useLinkTo('/settings')` and `useLinkBuilder().buildAction('/settings')` no longer prefer the config that matches the current route. In an app with both `app/(tabs)/settings.tsx` and `app/(modal)/settings.tsx`, calling `useLinkTo()('/settings')` from inside `(tabs)` used to resolve into `(tabs)`; it now resolves to whichever config sorts first globally.

That surface is `expo-router/react-navigation`, which AGENTS.md describes as a compatibility layer only — so this may well be an acceptable loss. But it is a silent change that is not in the changelog entry or the PR description. Is it intentional? If so, it is worth a line in the changelog; if not, `useBuildAction` needs the same `useRouteInfo()` treatment as the other call sites.

(`src/global-state/useStore.ts:81` also drops to two arguments, but that path computes the *initial* state, where empty segments are the correct input — arguably more correct than before.)

packages/expo-router/src/imperative-api.tsx
```diff
@@ -11,13 +12,14 @@ export { router };
...
   const events = useSyncExternalStore(
     routingQueue.subscribe,
     routingQueue.snapshot,
     routingQueue.snapshot
   );
   useEffect(() => {
-    routingQueue.run(ref);
-  }, [events, ref]);
+    routingQueue.run(ref, routeInfo);
+  }, [events, ref, routeInfo]);
```
jakub-agent: [💡 suggestion] This is the one place in the PR that threads route info through effect dependencies rather than `useEffectEvent`, which reads as inconsistent with the pattern applied in `useLinking`, `useLinking.native` and `useNextScreenId`.

It also has a cost: the effect now re-runs on every route change, and `routingQueue.run` unconditionally replaces `routingQueue.queue` with a fresh array (`// Reset the identity of the queue.`). So each navigation drains an already-empty queue and swaps the `useSyncExternalStore` snapshot identity without notifying subscribers — the snapshot no longer reflects what the subscribers were told.

`useEffectEvent` keeps the deps at `[events, ref]` while still reading the latest route info:

```ts
const runQueue = useEffectEvent(() => routingQueue.run(ref, routeInfo));
useEffect(() => { runQueue(); }, [events]);
```

packages/expo-router/src/global-state/routingQueue.ts
```diff
@@ -34,7 +35,10 @@ export const routingQueue = {
       callback();
     }
   },
-  run(ref: RefObject<NavigationContainerRef<ParamListBase> | null>) {
+  run(
+    ref: RefObject<NavigationContainerRef<ParamListBase> | null>,
+    routeInfo: UrlObject = defaultRouteInfo
```
jakub-agent: [🏗️ **design**] Same point as on `getNavigateAction`: `routeInfo` should be required rather than defaulting to `defaultRouteInfo`.

The only production caller is `useImperativeApiEmitter`, and it always passes a value. The default exists for the four `routingQueue.run(ref)` call sites in `__tests__/routingQueue.test.ios.ts` — which is exactly the coverage that should be forced to say which route the navigation starts from, since that is the behavior this PR changes.

packages/expo-router/src/global-state/__tests__/getNavigationAction.test.ios.ts
```diff
@@ -165,6 +175,22 @@ describe(getNavigateAction, () => {
     );
   });
 
+  it('uses route information provided by the caller', () => {
```
jakub-agent, line 184: [💡 suggestion] These assertions verify the plumbing (`getStateFromPath` receives `['current']`) but not the behavior the PR is refactoring. With `getStateFromPath` mocked, they would still pass if segment-aware resolution were broken end to end.

Since the whole point is that relative hrefs and group disambiguation keep working now that segments come from React scope instead of `store.getRouteInfo()`, a `renderRouter` test would be much stronger evidence — navigate into a nested route, then `router.push('../sibling')` and assert the resulting pathname; and, for the sorter tie-breaker, define the same path under two groups and assert which one a bare `/settings` resolves into from each.

Worth flagging that the `expo-router` Jest suite does not appear in this commit's check rollup (only the brownfield, precompiled-XCFramework and codemention jobs ran), so there is currently no CI signal for this package either.

packages/expo-router/src/link/getStateForHref.ts
```diff
@@ -0,0 +1,16 @@
+import type { ExpoLinkingOptions } from '../getLinkingConfig';
+import type { UrlObject } from '../global-state/getRouteInfoFromState';
+import type { LinkToOptions } from '../global-state/types';
+import type { Href } from '../types';
+import { resolveHref, resolveHrefStringWithSegments } from './href';
+
+export function getStateForHref(
+  href: Href | string,
+  routeInfo: UrlObject,
+  linking: ExpoLinkingOptions | undefined,
```
jakub-agent: [💡 suggestion] All three call sites pass `store.linking` (`HrefPreview.tsx:29`, `preview/utils.ts:22`, `preview/utils.ts:58`), and there is no second production implementation, so the parameter is not doing real work — reading `store.linking` directly here would shrink every call site. `routeInfo` earns its place because it now comes from React scope; `linking` does not.

Also: `linking?.getStateFromPath!(...)` on line 15 mixes optional chaining with a non-null assertion, so an object with a missing `getStateFromPath` throws while an `undefined` object returns `undefined`. That is inherited from the old `store.getStateForHref`, but since the helper is being written fresh it would be worth making the guard consistent (and adding the short comment the repo style asks for on `as`/`!` escapes).

packages/expo-router/src/link/preview/utils.ts
```diff
@@ -9,14 +10,16 @@ import type {
...
 import type { TabPath } from './native';
 
 export function getTabPathFromRootStateByHref(
   href: Href,
-  rootState: ReactNavigationState
+  rootState: ReactNavigationState,
+  routeInfo: UrlObject
 ): TabPath[] {
-  const hrefState = store.getStateForHref(resolveHref(href));
+  const hrefState = getStateForHref(resolveHref(href), routeInfo, store.linking);
```
jakub-agent: [🐧 nit] `resolveHref` runs twice on this path — once here and again as the first statement of `getStateForHref`. Since `getStateForHref` already normalizes its input, `getStateForHref(href, routeInfo, store.linking)` is enough (both here and on line 58).

packages/expo-router/src/getLinkingConfig.ts
```diff
@@ -130,8 +134,12 @@ export function getLinkingConfig(
...
+    ) => {
+      return getStateFromPath(path, options, previousSegments);
```
jakub-agent: [🐧 nit] With `getRouteInfo` gone this arrow is a pure identity wrapper, and `getStateFromPath` already matches the declared `getStateFromPath: typeof getStateFromPath` on line 50. `getStateFromPath,` would do the same job.

packages/expo-router/src/getLinkingConfig.ts
```diff
@@ -113,13 +111,19 @@ export function getLinkingConfig(
           if (typeof initialUrl === 'string') {
             initialUrl = applyRedirects(initialUrl, redirects);
             if (initialUrl && typeof nativeLinking?.redirectSystemPath === 'function') {
-              initialUrl = nativeLinking.redirectSystemPath({ path: initialUrl, initial: true });
+              initialUrl = nativeLinking.redirectSystemPath({
+                path: initialUrl,
```
jakub-agent: [🐧 nit] This reflow (and the matching one in the `initialUrl.then` branch below, plus the three `Object.defineProperty` / object-literal reflows in `__tests__/getNavigationAction.test.ios.ts`) is unrelated formatting churn — both call sites fit on one line under the repo's Prettier width. It looks like a local Prettier config with a narrower `printWidth`.

The reply to the earlier comment on this hunk said it had been reverted, but it is still present at `d6f1d0d`.

jakub-agent (commented):

The direction is right: `getLinkingConfig` no longer closes over `getRouteInfo`, and the segment snapshot now flows in from React scope through `useRouteInfo()` + `useEffectEvent`. The `useEffectEvent` usage matches the existing precedent in `layouts/useClearGuardedRoutes.ts`, and `useThenable` stores `getInitialState` in `useState(create)`, so the newly unstable `getInitialState` identity does not re-trigger initial-state resolution.

Two things are worth resolving before merge:

1. **The `defaultRouteInfo` default parameters** on `getNavigateAction` and `routingQueue.run` turn a missing argument into silently-wrong relative navigation instead of a type error. Both have exactly one production caller, and both pass a real value — the defaults only exist to keep the old test call sites compiling.
2. **`linking.getStateFromPath` no longer injects the current segments for callers that do not pass them.** `src/react-navigation/native/useLinkBuilder.tsx:106` still calls it with two arguments, so `useLinkTo` / `useLinkBuilder().buildAction` lose the `getRouteConfigSorter` tie-breaker. That is the compatibility layer, not Expo Router proper, but it is an unlisted behavior change.

On verification: I could not run `pnpm test` locally (no installed `node_modules` in the review worktree), and the PR's check rollup contains only the brownfield / precompiled-XCFramework / codemention jobs — the `expo-router` Jest suite has not run on this commit. The new assertions are mock-call assertions (`toHaveBeenCalledWith(..., ['current'])`) rather than behavioral ones, so nothing currently proves relative navigation still resolves correctly end to end.

One accepted trade-off worth stating explicitly: `useLinking` and `useImperativeApiEmitter` both call `useRouteInfo()`, so `NavigationContainerInner` now subscribes to route info and re-renders on every route change, not only on queued router actions. `routeInfoCache` de-duplicates by serialized value, so the identity stays stable when route info is unchanged, and `children` / `linkingContext` keep their references — the cost looks small, but it is a new root-level subscription.

packages/expo-router/src/global-state/getNavigationAction.ts
Ubax: Make it required

packages/expo-router/src/imperative-api.tsx
Ubax: use `useEffectEvent`

packages/expo-router/src/global-state/routingQueue.ts
Ubax: Make required

packages/expo-router/src/link/getStateForHref.ts
Ubax: Keep store.linking. It will be removed in a follow-up

packages/expo-router/src/link/preview/utils.ts
Ubax: Keep `resolveHref `

It was used before, and we don't want any functionality changes

packages/expo-router/src/getLinkingConfig.ts
Ubax: Remove

packages/expo-router/src/getLinkingConfig.ts
Ubax: Format properly. Use `pnpm format`

packages/expo-router/src/getLinkingConfig.ts
Ubax: Make `previousSegments` required param

packages/expo-router/src/global-state/getNavigationAction.ts
jakub-agent: Made routeInfo required and made the queue supply all navigation-action arguments explicitly.

packages/expo-router/src/imperative-api.tsx
jakub-agent: Switched the queue runner to useEffectEvent so route info is always current without resubscribing the effect.

packages/expo-router/src/global-state/routingQueue.ts
jakub-agent: Made routeInfo required and updated queue call sites and tests.

packages/expo-router/src/link/getStateForHref.ts
jakub-agent: Restored ownership of linking through store.linking inside getStateForHref.

packages/expo-router/src/link/preview/utils.ts
jakub-agent: Kept resolveHref in both preview helpers, preserving the previous resolution behavior.

packages/expo-router/src/getLinkingConfig.ts
jakub-agent: Removed the wrapper and return getStateFromPath directly.

packages/expo-router/src/getLinkingConfig.ts
jakub-agent: Ran pnpm --filter expo-router run format.

packages/expo-router/src/getLinkingConfig.ts
jakub-agent: Removed the wrapper that made previousSegments optional; consumers now use the linking function directly.

packages/expo-router/src/global-state/__tests__/getNavigationAction.test.ios.ts
```diff
@@ -1,8 +1,20 @@
...
+const getNavigateAction = (...args: any[]) =>
+  getNavigateActionImplementation(
+    args[0],
+    args[1],
+    args[2] ?? 'NAVIGATE',
+    args[3],
+    args[4],
+    args[5],
+    args[6] ?? defaultRouteInfo
+  );
```
Ubax: What is this?

If anything take each argument by name

packages/expo-router/src/getLinkingConfig.ts
Ubax: It's still here

packages/expo-router/src/global-state/routingQueue.ts
```diff
@@ -50,10 +51,11 @@ export const routingQueue = {
           action = getNavigateAction(
             href,
             options,
-            options.event,
+            options.event ?? 'NAVIGATE',
```
Ubax: Why is this needed?

packages/expo-router/src/global-state/getNavigationAction.ts
```diff
@@ -33,15 +35,15 @@ export function getNavigateAction(
   }
   const rootState = navigationRef.getRootState();
 
-  href = resolveHrefStringWithSegments(href, store.getRouteInfo(), options);
+  href = resolveHrefStringWithSegments(href, routeInfo, options);
```
Ubax: Only pass `segments` and `params` to this function. Change it's type to `Pick<UrlObject .....`

Then do the same in line 46 - only segments and change the param `routeInfo: UrlObject` in line 22 to `{segments,params}:Pick<UrlObject, 'segments' | 'params>`

packages/expo-router/src/global-state/routingQueue.ts
```diff
@@ -34,7 +35,7 @@ export const routingQueue = {
       callback();
     }
   },
-  run(ref: RefObject<NavigationContainerRef<ParamListBase> | null>) {
+  run(ref: RefObject<NavigationContainerRef<ParamListBase> | null>, routeInfo: UrlObject) {
```
Ubax: After the change in `getNavigateAction` use `Pick` here as well

packages/expo-router/src/fork/useLinking.native.ts
```diff
@@ -11,10 +11,14 @@ import {
...
 import { extractExpoPathFromURL } from './extractPathFromURL';
+import { getInitialURLWithTimeout } from './getInitialURLWithTimeout';
+import { getStateFromPath as getExpoStateFromPath } from './getStateFromPath';
 
 type ResultState = ReturnType<typeof getStateFromPathDefault>;
 
-type Options = LinkingOptions<ParamListBase>;
+type Options = Omit<LinkingOptions<ParamListBase>, 'getStateFromPath'> & {
+  getStateFromPath?: typeof getExpoStateFromPath;
+};
```
Ubax: Change it in `LinkingOptions` instead

packages/expo-router/src/fork/useLinking.ts
```diff
@@ -74,7 +84,9 @@ export const series = (cb: () => Promise<void>) => {
 
 const linkingHandlers: symbol[] = [];
 
-type Options = LinkingOptions<ParamListBase>;
+type Options = Omit<LinkingOptions<ParamListBase>, 'getStateFromPath'> & {
+  getStateFromPath?: typeof getExpoStateFromPath;
+};
```
Ubax: Change in `LinkingOptions` instaed

packages/expo-router/src/fork/useLinking.ts
```diff
@@ -196,7 +216,7 @@ export function useLinking(
 
     return thenable as PromiseLike<ResultState | undefined>;
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+  }, [getStateFromPathForCurrentSegments, onUnhandledLinking, server]);
```
Ubax: Remove the deps, since this function does not need to refresh. It is used only once.

Add comment with reference to where it is used

packages/expo-router/src/link/getStateForHref.ts
```diff
@@ -0,0 +1,15 @@
+import type { UrlObject } from '../global-state/getRouteInfoFromState';
+import { store } from '../global-state/router-store';
+import type { LinkToOptions } from '../global-state/types';
+import type { Href } from '../types';
+import { resolveHref, resolveHrefStringWithSegments } from './href';
+
+export function getStateForHref(
+  href: Href | string,
+  routeInfo: UrlObject,
```
Ubax: Change this to Pick<..., 'segments'>

packages/expo-router/src/link/preview/HrefPreview.tsx
```diff
@@ -18,11 +19,13 @@ import {
...
 import { useNavigation } from '../../useNavigation';
 import { getQualifiedRouteComponent } from '../../useScreens';
+import { getStateForHref } from '../getStateForHref';
 import { getPathFromState } from '../linking';
 import { PreviewRouteContext } from './PreviewRouteContext';
 
 export function HrefPreview({ href }: { href: Href }) {
-  const hrefState = useMemo(() => getHrefState(href), [href]);
+  const routeInfo = useRouteInfo();
+  const hrefState = useMemo(() => getStateForHref(href, routeInfo), [href, routeInfo]);
```
Ubax: Add dep only on segments

packages/expo-router/src/link/preview/utils.ts
```diff
@@ -9,14 +10,16 @@ import type {
...
 import type { Href } from '../../types';
+import { getStateForHref } from '../getStateForHref';
 import { resolveHref } from '../href';
 import type { TabPath } from './native';
 
 export function getTabPathFromRootStateByHref(
   href: Href,
-  rootState: ReactNavigationState
+  rootState: ReactNavigationState,
+  routeInfo: UrlObject
```
Ubax: Change to Pick

packages/expo-router/src/link/preview/utils.ts
```diff
@@ -49,9 +52,10 @@ export function getTabPathFromRootStateByHref(
 
 export function getPreloadedRouteFromRootStateByHref(
   href: Href,
-  rootState: ReactNavigationState
+  rootState: ReactNavigationState,
+  routeInfo: UrlObject
```
Ubax: Change to Pick

packages/expo-router/src/global-state/__tests__/getNavigationAction.test.ios.ts
jakub-agent: Replaced the positional test helper with named arguments and explicit types.

packages/expo-router/src/getLinkingConfig.ts
jakub-agent: Removed getStateFromPath from the linking config and its specialized return type.

packages/expo-router/src/global-state/routingQueue.ts
jakub-agent: Removed the fallback; the queued link action already supplies its navigation event.

packages/expo-router/src/global-state/getNavigationAction.ts
jakub-agent: Narrowed the action input to Pick<UrlObject, "segments" | "params"> and destructure those fields at the boundary.

packages/expo-router/src/global-state/routingQueue.ts
jakub-agent: Narrowed the routing queue route-info parameter to the same Pick type.

packages/expo-router/src/fork/useLinking.native.ts
jakub-agent: Moved the extended getStateFromPath signature into LinkingOptions and removed the local native override.

packages/expo-router/src/fork/useLinking.ts
jakub-agent: Moved the extended getStateFromPath signature into LinkingOptions and removed the local web override.

packages/expo-router/src/fork/useLinking.ts
jakub-agent: Made getInitialState stable with an empty dependency list and documented its one-time NavigationContainer use.

packages/expo-router/src/link/getStateForHref.ts
jakub-agent: Narrowed the helper input to Pick<UrlObject, "segments">.

packages/expo-router/src/link/preview/HrefPreview.tsx
jakub-agent: The memo now depends only on href and the route segments.

packages/expo-router/src/link/preview/utils.ts
jakub-agent: Narrowed this helper route-info input to Pick<UrlObject, "segments">.

packages/expo-router/src/link/preview/utils.ts
jakub-agent: Narrowed this helper route-info input to Pick<UrlObject, "segments">.

packages/expo-router/src/global-state/getNavigationAction.ts
```diff
@@ -1,23 +1,26 @@
...
   options: LinkToOptions,
-  type = 'NAVIGATE',
-  withAnchor?: boolean,
-  singular?: SingularOptions,
-  isPreviewNavigation?: boolean
+  type: string,
+  withAnchor: boolean | undefined,
+  singular: SingularOptions | undefined,
+  isPreviewNavigation: boolean | undefined,
+  { segments, params: routeParams }: Pick<UrlObject, 'segments' | 'params'>
```
Ubax:
```suggestion
  { segments, params: routeParams }: Pick<UrlObject, 'segments' | 'params'>,
  type: string =....,
  withAnchor?: boolean,
  singular?: SingularOptions,
  isPreviewNavigation?: boolean,
```

packages/expo-router/src/imperative-api.tsx
```diff
@@ -17,7 +18,7 @@ export function useImperativeApiEmitter(
     routingQueue.snapshot
   );
   useEffect(() => {
-    routingQueue.run(ref);
+    routingQueue.run(ref, store.getRouteInfo());
```
Ubax, line 22: Add `TODO` comment here to use `useRouteInfo` once it's computed from global state

packages/expo-router/src/imperative-api.tsx
```diff
@@ -17,7 +18,8 @@ export function useImperativeApiEmitter(
     routingQueue.snapshot
   );
   useEffect(() => {
-    routingQueue.run(ref);
+    // TODO: Use useRouteInfo once it is computed from global state.
```
jakub-agent, line 21: [💡 suggestion] Worth making this TODO say *why*, because as written it invites exactly the regression that `ebcfa65` shipped.

A reader who acts on "use `useRouteInfo` once it is computed from global state" will reintroduce 4 relative-navigation failures (`navigation.test.ios.tsx` shared-routes-with-tabs and static-over-dynamic, `Link.test.ios.tsx` relative-route-with-path-params). The blocker is not really *where* route info is computed — it is the re-render timing: `routingQueue.add` mutates `routingQueue.queue` in place, so `routingQueue.snapshot()` returns the same array reference and `useSyncExternalStore` sees no change. This component therefore is not guaranteed to re-render between the enqueue and the drain, and a React-scope read is stale. When I instrumented it, the React value was `[]` on every drain while `store.getRouteInfo()` correctly reported `["(two)","two"]`.

So the precondition is narrower and more concrete than the current wording suggests — something like:

```ts
// Reads the store rather than `useRouteInfo()` because `routingQueue.add` mutates the
// queue in place: `useSyncExternalStore` sees an unchanged snapshot, so this component
// may not re-render between enqueue and drain and a React-scope read would be stale.
// TODO: switch to `useRouteInfo()` once `add` publishes a new queue identity.
```

Naming the actual unblocking condition makes the TODO actionable instead of a trap.

packages/expo-router/src/getLinkingConfig.ts
```diff
@@ -130,9 +127,6 @@ export function getLinkingConfig(
       return initialUrl;
     },
     subscribe: subscribe(nativeLinking, redirects),
```
jakub-agent, line 129: [🏗️ **design**] Carrying this over from rounds 1 and 2 for a decision, since it is the one behavior change that survives — flagging it rather than blocking on it.

On `main`, this object supplied a `getStateFromPath` that injected the current segments, so every consumer reading it off the linking options got the `getRouteConfigSorter` tie-breaker. The key is now gone, and `react-navigation/native/useLinkBuilder.tsx:97` falls back to the bare core export:

```ts
const getStateFromPathHelper = options?.getStateFromPath ?? getStateFromPath;
// ...
const state = getStateFromPathHelper(href, options?.config);
```

Two arguments, so `segments` defaults to `[]`. Net effect: `useLinkTo` and `useLinkBuilder().buildAction` no longer prefer the config matching the current route. With both `app/(tabs)/settings.tsx` and `app/(modal)/settings.tsx`, `useLinkTo()('/settings')` from inside `(tabs)` used to resolve into `(tabs)` and now takes whichever config sorts first globally.

This is `expo-router/react-navigation`, which AGENTS.md scopes as a compatibility layer, and the green suite confirms nothing in Expo Router proper depends on it — so accepting it is entirely reasonable. It is just not currently recorded anywhere: the changelog entry reads "Keep linking route information in React scope", which does not hint at it. Either add a line to the changelog, or thread route info into `useBuildAction` the way the other call sites now do. Your call which — I only want it to be a deliberate one.

jakub-agent (approved):

**Review round 3** of `826f739c` (rebased onto `b85bd2a`).

All three critical findings from round 2 are fixed, and the package is now at exact parity with `main`:

| | tests failed | suites failed | typecheck | lint |
| --- | --- | --- | --- | --- |
| `origin/main` | 3 | 6 | — | — |
| `ebcfa65` (round 2) | 69 | 13 | — | — |
| `826f739c` (this head) | **3** | **6** | clean | clean |

Same three failures, same six suites, same three stale snapshots as `main` — all pre-existing (`TopTabs-exports`, `hmr-route-changes`, `native-tabs/__rsc_tests__/tabs` on all RSC platforms). Nothing in this PR's blast radius fails.

How each round-2 finding was resolved:

1. **`options.event!`** — `getNavigateAction` moved `routeInfo` to the third position and restored `type = 'NAVIGATE'` plus the other trailing defaults. `routingQueue` passes `options.event` unasserted. This is a better shape than what I suggested: route info stays required, the optional tail keeps its defaults, and the test-only wrapper is gone (tests now pass `defaultRouteInfo` explicitly).
2. **`getStateFromPath` on the linking config** — `ui/common.tsx` imports it directly, so the config no longer needs the key.
3. **Stale React-scope route info** — `useImperativeApiEmitter` reads `store.getRouteInfo()` in the effect. That was the patch I verified.

Also resolved: the `react-navigation/native/types.tsx` no-op is reverted, and the `getLinkingConfig` Prettier churn is gone.

The rebase conflict resolution (`826f739c`, dropping `useNavigationIndependentTree` after #49172 removed it) is complete — no dangling `independent` references in either `useLinking` variant, and typecheck confirms it.

Two non-blocking items below. Neither is a defect in the changed code, so this looks mergeable to me.

Verification run locally:
```
cd packages/expo-router && pnpm build && pnpm typecheck && pnpm lint && CI=1 pnpm test
```

PR #49182, opened 2026-08-20
fix(eas-client): make deterministicUniformValue actually uniform over [0, 1)

### LINKED ISSUE #49181: expo-eas-client: deterministicUniformValue returns [0.5, 0.75), so any Observe sampleRate <= 0.5 samples zero installs

https://gist.github.com/bjjeong/46c8813eb95262ae846895531d1e3a4f

The defective function is **pure**, so no Expo app is needed to reproduce it. The gist contains a verbatim copy of the shipped implementation for each platform, run over 200k freshly minted UUIDs:

- `EASClientIDRepro.swift` → `swift EASClientIDRepro.swift`
- `EASClientIDRepro.java` → `java EASClientIDRepro.java`

(Refiling — a previous report was auto-closed for lacking a linked repro. Apologies, I had inlined the reproduction rather than linking it.)

### Summary

`EASClientID.deterministicUniformValue` is documented as returning "a deterministic value in [0, 1]", but returns a value confined to **[0.5, 0.75)**.

`expo-observe` gates dispatch on `deterministicUniformValue(...) < sampleRate` (`Observability.swift:256`), so **any `sampleRate` at or below 0.5 selects exactly zero installs.** Telemetry silently goes to nothing rather than being reduced.

This bit us in production: shipping `sampleRate: 0.03` took fleet-wide startup + navigation telemetry to 0 samples on both platforms.

### Steps to reproduce

1. Download the gist and run `swift EASClientIDRepro.swift` (and/or `java EASClientIDRepro.java`).
2. Observe `min` ≈ 0.5 and `max` ≈ 0.75 — never uniform over [0, 1).
3. Observe the in-sample counts: `sampleRate 0.03` → **0**, `sampleRate 0.50` → **0**, `sampleRate 0.75` → all 200000.

Equivalently, end to end: call `Observe.configure({ sampleRate: 0.03 })` in a release build and observe that no metrics are ever dispatched.

**Actual output**

iOS (Swift 6 / macOS 26.3):
```
min:                      0.5000032009047606
max:                      0.7499994446248878
sampleRate 0.03  -> in-sample: 0  (0.0%)
sampleRate 0.50  -> in-sample: 0  (0.0%)
sampleRate 0.75  -> in-sample: 200000  (100.0%)
```

Android (OpenJDK 17):
```
min:     0.5000033200958995
max:     0.7499992853526236
sampleRate 0.03 -> in-sample: 0
sampleRate 0.50 -> in-sample: 0
sampleRate 0.75 -> in-sample: 200000
```

**Expected:** values uniform over [0, 1), so `sampleRate: 0.03` selects ~3% of installs.

### Cause

```swift
// packages/expo-eas-client/ios/EASClient/EASClientID.swift:17-22
$0.load(fromByteOffset: 8, as: UInt64.self).bigEndian / UInt64.max
```
```kotlin
// packages/expo-eas-client/android/src/main/java/expo/modules/easclient/EASClientID.kt:15-17
uuid.leastSignificantBits.toULong().toDouble() / ULong.MAX_VALUE.toDouble()
```

Byte 8 of an RFC 4122 UUID is `clock_seq_hi_and_reserved`, whose top two bits are pinned to `10` for the IETF variant. It is the **most significant** byte of the extracted 64-bit value, so the value is always in `[0x8000…, 0xBFFF…]` and the quotient is always in `[0.5, 0.75)`. `java.util.UUID.getLeastSignificantBits()` is exactly those same bytes, so Android carries the identical defect.

The UUIDs come from `Foundation.UUID()` and `java.util.UUID.randomUUID()` — RFC 4122 v4 by specification — so this affects every install.

### Impact

- `sampleRate <= 0.5` → **0%** of installs report
- `0.5 < sampleRate < 0.75` → distorted fraction, roughly `(rate - 0.5) * 4`
- `sampleRate >= 0.75` → 100%

The failure mode is silent and reads as low adoption. In our case each install emitted exactly one `cold_launch` — dispatched before `Observe.configure()` persists the config, so it takes the `guard let rate = ... else { return true }` path at `Observability.swift:251-253` — then went permanently dark.

### Suggested fix

Derive the fraction from bits not reserved by the UUID spec: hash all 16 bytes (FNV-1a/SipHash/SHA-256 truncated), or mask out the version/variant bits before use. Both platforms need the same change so the decision stays stable across an upgrade.

A regression test asserting approximate *uniformity* over [0, 1) would catch this — the current behaviour passes a naive "is within [0, 1]" assertion while being badly non-uniform.

### Environment

```
    System:
      OS: macOS 26.3.1
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 24.16.0
      npm: 11.13.0
      Watchman: 2026.03.09.00
    Managers:
      CocoaPods: 1.16.2
    SDKs:
      iOS SDK:
      Android SDK:
        API Levels: 31, 33, 34, 35, 36
        Build Tools: 34.0.0, 35.0.0, 36.0.0
    IDEs:
      Android Studio: 2025.3
      Xcode: 26.5/17F42
    npmPackages:
      expo: ~57.0.1 => 57.0.2
      expo-updates: ~57.0.5 => 57.0.6
      react: 19.2.3 => 19.2.3
    npmGlobalPackages:
      eas-cli: 21.0.3
    Expo Workflow: bare
```

Affected packages: `expo-eas-client@57.0.1` (latest published; verified byte-identical to the npm tarball, no local patches), `expo-observe@57.0.8`, `expo-app-metrics@57.0.6`.

### Expo Doctor Diagnostics

Not applicable — the defect is in a pure function in published package source and reproduces standalone via the linked gist, with no project involved. `expo-doctor` validates project configuration and does not exercise this code path.

bjjeong:
# Why

Fixes #49181.

`EASClientID.deterministicUniformValue` is documented as returning "a deterministic value in [0, 1]", but it returns a value confined to **[0.5, 0.75)**.

`expo-observe` is the only consumer, and it gates dispatch on `deterministicUniformValue(...) < sampleRate` (`Observability.swift:276`, `ObservabilityManager.kt:325`). So **any `sampleRate` at or below 0.5 selects exactly zero clients** — telemetry silently drops to nothing rather than being reduced. We hit this in production: shipping `sampleRate: 0.03` took fleet-wide startup and navigation telemetry to zero on both platforms.

Both implementations read bytes 8–15 of the UUID:

```swift
$0.load(fromByteOffset: 8, as: UInt64.self).bigEndian / UInt64.max
```
```kotlin
uuid.leastSignificantBits.toULong().toDouble() / ULong.MAX_VALUE.toDouble()
```

Byte 8 of an RFC 4122 UUID is `clock_seq_hi_and_reserved`, whose two most significant bits are pinned to `10` for the IETF variant. Because it is *also* the most significant byte of that 64-bit window, the quotient can never leave `[0.5, 0.75)`. `java.util.UUID.getLeastSignificantBits()` is exactly those same bytes, so Android carries the identical defect.

# How

Fold both halves (`high ^ low`) through a splitmix64 finalizer, which moves the reserved version/variant bits out of the high end of the result, then scale by `2^53` so the output stays in `[0, 1)` and can never be exactly `1.0`.

The two platforms produce **bit-identical** output for the same UUID, so a client's in/out-of-sample decision does not change if its platform implementation is ever swapped.

# Test Plan

Ran verbatim copies of the before and after implementations over freshly minted UUIDs — Swift 6 with `Foundation.UUID`, and Java with `java.util.UUID` (the Kotlin path).

**Before**

| platform | min | max | rate 0.03 | rate 0.50 | rate 0.75 |
|---|---|---|---|---|---|
| Swift | 0.5000032 | 0.7499994 | **0%** | **0%** | 100% |
| Java | 0.5000033 | 0.7499993 | **0%** | **0%** | 100% |

**After** (500k samples each)

| platform | min | max | rate 0.03 | rate 0.50 | decile histogram (per-mille) |
|---|---|---|---|---|---|
| Swift | 6.1e-07 | 0.9999988 | **3.037%** | **50.04%** | `100 99 100 100 100 99 100 100 99 99` |
| Java | 2.7e-06 | 0.9999997 | **2.954%** | **49.93%** | `99 98 99 100 101 99 99 100 100 100` |

Cross-platform agreement on fixed UUIDs (identical to the last digit):

| UUID | value |
|---|---|
| `00000000-0000-4000-8000-000000000000` | `0.6739966708523056` |
| `ffffffff-ffff-4fff-bfff-ffffffffffff` | `0.7311409341971677` |
| `3f2504e0-4f89-41d3-9a0c-0305e82c3301` | `0.7672797462886537` |
| `a1b2c3d4-e5f6-4789-abcd-ef0123456789` | `0.20976227781883072` |

Determinism (same UUID → same value across repeated calls) verified.

A standalone runnable reproduction for both platforms is linked from #49181.

# Notes

- **This changes which clients are in-sample.** The decision is keyed on the persisted client UUID, so a client's bucket will shift once — unavoidable when correcting the distribution, and harmless for the percentage-based sampling this feeds. Note `deterministicUniformValue` is also a public JS export of `expo-eas-client`, so apps reading it directly re-bucket as well.
- A regression test asserting approximate *uniformity* (e.g. decile occupancy) rather than mere `[0, 1]` membership would prevent a recurrence; the previous behaviour satisfies an "is within range" check while being badly non-uniform. Added as `testDeterministicUniformValueIsUniformlyDistributed` on both platforms.
- `jest-expo`'s mock for this constant (`0.694392985627176`, in `preset/moduleMocks/expoModules.js`) happens to sit inside the old biased range. It is still a valid value post-fix, so I have left it alone — flagging in case you would rather it be representative.

### DIFF
--- a/packages/expo-eas-client/CHANGELOG.md
+++ b/packages/expo-eas-client/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### 🐛 Bug fixes
 
+- Fix `deterministicUniformValue` to return a uniformly distributed value over `[0, 1)`. ([#49182](https://github.com/expo/expo/pull/49182) by [@bjjeong](https://github.com/bjjeong))
+
 ### 💡 Others
 
 ## 57.0.1 - 2026-07-15
--- a/packages/expo-eas-client/android/src/androidTest/java/expo/modules/easclient/EASClientIDTest.kt
+++ b/packages/expo-eas-client/android/src/androidTest/java/expo/modules/easclient/EASClientIDTest.kt
@@ -3,7 +3,8 @@ package expo.modules.easclient
 import androidx.test.internal.runner.junit4.AndroidJUnit4ClassRunner
 import androidx.test.platform.app.InstrumentationRegistry.getInstrumentation
 import org.amshove.kluent.shouldBeEqualTo
-import org.amshove.kluent.shouldBeInRange
+import org.amshove.kluent.shouldBeGreaterOrEqualTo
+import org.amshove.kluent.shouldBeLessThan
 import org.amshove.kluent.shouldNotBe
 import org.junit.Test
 import org.junit.runner.RunWith
@@ -25,15 +26,16 @@ class EASClientIDTest {
   fun testDeterministicUniformValueKnownValue() {
     val uuid = UUID.fromString("a1b2c3d4-e5f6-7890-abcd-ef1234567890")
     val value = EASClientID.deterministicUniformValue(uuid)
-    // leastSignificantBits (unsigned) = 0xABCDEF1234567890 / ULong.MAX
-    org.junit.Assert.assertEquals(0.6711110515064663, value, 1e-15)
+    // splitmix64(mostSignificantBits xor leastSignificantBits), scaled by 2^53.
+    org.junit.Assert.assertEquals(0.5075081783308123, value, 1e-15)
   }
 
   @Test
   fun testDeterministicUniformValueRange() {
     val context = getInstrumentation().targetContext
     val value = EASClientID.deterministicUniformValue(EASClientID(context).uuid)
-    value shouldBeInRange 0.0..1.0
+    value shouldBeGreaterOrEqualTo 0.0
+    value shouldBeLessThan 1.0
   }
 
   @Test
@@ -42,11 +44,57 @@ class EASClientIDTest {
     EASClientID.deterministicUniformValue(uuid) shouldBeEqualTo EASClientID.deterministicUniformValue(uuid)
   }
 
+  /**
+   * A known-value test passes for any biased formula, so occupying the whole range is the
+   * property that actually matters to callers comparing `value < threshold`.
+   */
+  @Test
+  fun testDeterministicUniformValueIsUniformlyDistributed() {
+    val occupiedDeciles =
+      (0 until 2000)
+        .map { seed ->
+          val value = EASClientID.deterministicUniformValue(makeV4Uuid(seed.toLong()))
+
+          value shouldBeGreaterOrEqualTo 0.0
+          value shouldBeLessThan 1.0
+
+          (value * 10).toInt()
+        }
+        .toSet()
+
+    occupiedDeciles shouldBeEqualTo (0 until 10).toSet()
+  }
+
   @Test
   fun testUuidIsV4() {
     val context = getInstrumentation().targetContext
     val uuid = EASClientID(context).uuid
     uuid.version() shouldBeEqualTo 4
     uuid.variant() shouldBeEqualTo 2
+  }
+
+  /**
+   * Builds RFC 4122 v4-shaped UUIDs from a seeded LCG so the distribution assertions are
+   * reproducible rather than relying on random input.
+   */
+  private fun makeV4Uuid(seed: Long): UUID {
+    var state = seed
+    val bytes = ByteArray(16)
+    for (index in bytes.indices) {
+      state = state * 6364136223846793005L + 1442695040888963407L
+      bytes[index] = (state ushr 33).toByte()
+    }
+    bytes[6] = ((bytes[6].toInt() and 0x0F) or 0x40).toByte() // version 4
+    bytes[8] = ((bytes[8].toInt() and 0x3F) or 0x80).toByte() // variant 10
+
+    var mostSignificantBits = 0L
+    var leastSignificantBits = 0L
+    for (index in 0 until 8) {
+      mostSignificantBits = (mostSignificantBits shl 8) or (bytes[index].toLong() and 0xFF)
+    }
+    for (index in 8 until 16) {
+      leastSignificantBits = (leastSignificantBits shl 8) or (bytes[index].toLong() and 0xFF)
+    }
+    return UUID(mostSignificantBits, leastSignificantBits)
   }
 }
--- a/packages/expo-eas-client/android/src/main/java/expo/modules/easclient/EASClientID.kt
+++ b/packages/expo-eas-client/android/src/main/java/expo/modules/easclient/EASClientID.kt
@@ -9,11 +9,17 @@ private const val EAS_CLIENT_ID_SHARED_PREFERENCES_KEY = "eas-client-id"
 class EASClientID(private val context: Context) {
   companion object {
     /**
-     * Converts a UUID to a deterministic value in [0, 1] using the least significant
-     * 64 bits interpreted as an unsigned integer fraction of ULong.MAX_VALUE.
+     * Converts a UUID to a deterministic value in [0, 1).
      */
     fun deterministicUniformValue(uuid: UUID): Double {
-      return uuid.leastSignificantBits.toULong().toDouble() / ULong.MAX_VALUE.toDouble()
+      // Byte 8 is the RFC 4122 variant octet, pinned to `10`, and it is the high byte of
+      // leastSignificantBits, so that half alone only spans [0.5, 0.75]. splitmix64 over both
+      // halves moves the fixed bits off the high end; 2^53 is the widest exact Double range.
+      var z = (uuid.mostSignificantBits xor uuid.leastSignificantBits).toULong()
+      z = (z xor (z shr 30)) * 0xbf58476d1ce4e5b9UL
+      z = (z xor (z shr 27)) * 0x94d049bb133111ebUL
+      z = z xor (z shr 31)
+      return (z shr 11).toDouble() / (1L shl 53).toDouble()
     }
   }
 
--- a/packages/expo-eas-client/ios/EASClient/EASClientID.swift
+++ b/packages/expo-eas-client/ios/EASClient/EASClientID.swift
@@ -12,13 +12,22 @@ public class EASClientID : NSObject {
     })!
   }
 
-  /// Converts a UUID to a deterministic value in [0, 1] using the least significant
-  /// 64 bits (bytes 8–15) interpreted as an unsigned integer fraction of UInt64.max.
+  /// Converts a UUID to a deterministic value in [0, 1).
   public static func deterministicUniformValue(_ uuid: UUID) -> Double {
-    let value = withUnsafeBytes(of: uuid.uuid) {
-      $0.load(fromByteOffset: 8, as: UInt64.self).bigEndian
+    // Byte 8 is the RFC 4122 variant octet, pinned to `10`, and it is the high byte of the low
+    // half, so that half alone only spans [0.5, 0.75]. splitmix64 over both halves moves the
+    // fixed bits off the high end; 2^53 is the widest exact Double range.
+    let (high, low) = withUnsafeBytes(of: uuid.uuid) {
+      (
+        $0.load(fromByteOffset: 0, as: UInt64.self).bigEndian,
+        $0.load(fromByteOffset: 8, as: UInt64.self).bigEndian
+      )
     }
-    return Double(value) / Double(UInt64.max)
+    var z = high ^ low
+    z = (z ^ (z >> 30)) &* 0xbf58_476d_1ce4_e5b9
+    z = (z ^ (z >> 27)) &* 0x94d0_49bb_1331_11eb
+    z = z ^ (z >> 31)
+    return Double(z >> 11) / Double(1 << 53)
   }
 }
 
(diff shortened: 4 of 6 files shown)


bjjeong:
If you would prefer a smaller diff, the minimal fix is to clear the two RFC 4122 variant bits rather than mixing all 128 — replacing just the final division:

```swift
// value = bytes 8–15, as today
return Double((value & 0x3FFF_FFFF_FFFF_FFFF) >> 9) / Double(1 << 53)
```
```kotlin
return ((uuid.leastSignificantBits.toULong() and 0x3FFFFFFFFFFFFFFFuL) shr 9).toDouble() / (1L shl 53).toDouble()
```

It measures the same as this PR (3.01% vs 2.98% in-sample at `sampleRate: 0.03` over 200k v4 UUIDs, both strictly `< 1.0`), and both hold up if the client ID ever moves to UUIDv7. I used splitmix64 so it does not depend on which bits happen to be random, but the mask is easier to review. Happy to switch if you prefer the mask!

One to avoid: using the high 64 bits instead. Uniform on v4, but on v7 those bytes are a millisecond timestamp and every client collapses to nearly the same value.

brentvatne:
@expo-bot check with fable

bjjeong:
Updated both known-value tests to `0.5075081783308123`, fixed the stale comment on the Android one, and added the uniformity test. It fails on the current implementation and passes with the fix.

Also tightened the TSDoc to `[0, 1)`. `[0, 1]` was accurate before, since `value / UInt64.max` can reach exactly `1.0`, but the new numerator maxes at `2^53 - 1` over a `2^53` denominator. It matters because callers compare `value < sampleRate`: at `sampleRate: 1.0` every client is now included.

brentvatne:
@expo-bot verify

brentvatne:
@expo-bot review

bjjeong:
Both fixed. Tightened the Android range assertion to match iOS, and widened the changelog and both doc comments to cover the full curve rather than only thresholds at or below 0.5.

Also corrected `[0.5, 0.75)` to `[0.5, 0.75]`, since you're right that the bound is closed.

brentvatne:
@expo-bot check with fable

packages/expo-eas-client/android/src/androidTest/java/expo/modules/easclient/EASClientIDTest.kt
```diff
@@ -42,11 +44,63 @@ class EASClientIDTest {
...
+    }
+
+    // Deriving the value from leastSignificantBits alone confined it to [0.5, 0.75], leaving
+    // deciles 0-4, 8 and 9 permanently empty - byte 8 is the variant octet, pinned to `10`.
+    deciles.forEachIndexed { index, count ->
+      org.junit.Assert.assertTrue(
+        "decile $index was never occupied; the distribution is not uniform",
+        count > 0
+      )
+    }
```
Ubax:
```suggestion
    val occupiedDeciles =
      (0 until 2000)
        .map { seed ->
          val value = EASClientID.deterministicUniformValue(makeV4Uuid(seed.toLong()))

          value shouldBeGreaterOrEqualTo 0.0
          value shouldBeLessThan 1.0

          (value * 10).toInt()
        }
        .toSet()

    occupiedDeciles shouldBe (0 until 10).toSet()
```

Nit, this will make the test a little clearer

packages/expo-eas-client/CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 ### 🐛 Bug fixes
 
+- Fixed `deterministicUniformValue` returning values confined to `[0.5, 0.75]` instead of being uniformly distributed over `[0, 1)`. This skewed every threshold compared against it, not only small ones: the matching fraction was `clamp((threshold - 0.5) / 0.25, 0, 1)`, so a threshold of `0.6` matched 40% of clients rather than 60%, and any threshold at or below `0.5` matched none. ([#49182](https://github.com/expo/expo/pull/49182) by [@bjjeong](https://github.com/bjjeong))
```
Ubax:
```suggestion
- Fix `deterministicUniformValue` to return a uniformly distributed value over `[0, 1)`. ([#49182](https://github.com/expo/expo/pull/49182) by [@bjjeong](https://github.com/bjjeong))
```

Ubax (approved):
Let's wait for @tsapeta review

packages/expo-eas-client/android/src/main/java/expo/modules/easclient/EASClientID.kt
```diff
@@ -9,11 +9,26 @@ private const val EAS_CLIENT_ID_SHARED_PREFERENCES_KEY = "eas-client-id"
...
+     *
+     * The least significant 64 bits cannot be used on their own: byte 8 of an RFC 4122 UUID
+     * is the variant octet, whose two most significant bits are pinned to `10`. Because it is
+     * also the most significant byte of `leastSignificantBits`, the raw quotient is confined
+     * to [0.5, 0.75]. That skews every comparison a caller makes against it, not just small
+     * thresholds: the fraction matching `deterministicUniformValue(...) < threshold` is
+     * `clamp((threshold - 0.5) / 0.25, 0, 1)`, so a threshold of 0.6 matches 40% rather than
+     * 60%, anything at or below 0.5 matches nothing, and anything at or above 0.75 matches
+     * everything. Folding both halves through a splitmix64 finalizer moves the reserved
+     * version/variant bits out of the high end of the result.
```
Ubax:
```suggestion
     * Converts a UUID to a deterministic value in [0, 1) by mixing all 128 bits.
```

Let's keep doc comment short

packages/expo-eas-client/android/src/main/java/expo/modules/easclient/EASClientID.kt
```diff
@@ -9,11 +9,26 @@ private const val EAS_CLIENT_ID_SHARED_PREFERENCES_KEY = "eas-client-id"
...
      */
     fun deterministicUniformValue(uuid: UUID): Double {
-      return uuid.leastSignificantBits.toULong().toDouble() / ULong.MAX_VALUE.toDouble()
+      var z = (uuid.mostSignificantBits xor uuid.leastSignificantBits).toULong()
```
Ubax, line 18:
```suggestion
     // The least significant 64 bits cannot be used on their own: byte 8 of an RFC 4122 UUID
     // is the variant octet, whose two most significant bits are pinned to `10`. Because it is
     // also the most significant byte of `leastSignificantBits`, the raw quotient is confined
     // to [0.5, 0.75]. 
     var z = (uuid.mostSignificantBits xor uuid.leastSignificantBits).toULong()
```

And then we can add more context inline

packages/expo-eas-client/ios/EASClient/EASClientID.swift
```diff
@@ -12,13 +12,31 @@ public class EASClientID : NSObject {
...
+  ///
+  /// The low 64 bits (bytes 8–15) cannot be used on their own: byte 8 of an RFC 4122 UUID is
+  /// the variant octet, whose two most significant bits are pinned to `10`. Because it is also
+  /// the most significant byte of that window, the raw quotient is confined to [0.5, 0.75].
+  /// That skews every comparison a caller makes against it, not just small thresholds: the
+  /// fraction matching `deterministicUniformValue(...) < threshold` is
+  /// `clamp((threshold - 0.5) / 0.25, 0, 1)`, so a threshold of 0.6 matches 40% rather than
+  /// 60%, anything at or below 0.5 matches nothing, and anything at or above 0.75 matches
+  /// everything. Folding both halves through a splitmix64 finalizer moves the reserved
+  /// version/variant bits out of the high end of the result.
```
Ubax: Can you do the same change to the comment here as well?

packages/expo-eas-client/ios/Tests/EASClientIDTest.swift
```diff
@@ -32,6 +33,27 @@ class EASClientIdTests : XCTestCase {
...
+
+    for seed in 0..<2000 {
+      let value = EASClientID.deterministicUniformValue(Self.makeV4UUID(seed: UInt64(seed)))
+      XCTAssertGreaterThanOrEqual(value, 0.0)
+      XCTAssertLessThan(value, 1.0)
+      deciles[min(9, Int(value * 10))] += 1
+    }
+
+    // Deriving the value from bytes 8–15 alone confined it to [0.5, 0.75), leaving deciles
+    // 0–4 and 8–9 permanently empty — byte 8 is the variant octet, pinned to `10`.
```
Ubax:
```suggestion
```

This does not add much context

Ubax (commented):
Please also apply small comment changes

bjjeong:
@Ubax Suggestions and cleanup applied!

bjjeong:
@tsapeta Pushed a fix for the failing test

PR #49191, opened 2026-08-21
[android][secure-store] Reject setItemAsync when the write did not reach disk

### LINKED ISSUE #48988: [expo-secure-store] Android: setItemAsync resolves when the write did not reach disk (discarded commit() result)
On Android, `SecureStore.setItemAsync` can resolve successfully while the value never reaches disk.
`saveEncryptedItem` returns the result of `SharedPreferences.Editor.commit()`, and `setItemImpl`
calls it as a bare statement, discarding that `Boolean`. Every other `commit()` in the same file has
its result checked.

The consequence is silent data loss: the JS promise resolves, so a caller that treats resolution as
"persisted" — for example by then clearing its own copy — loses the value when the process dies.

https://github.com/jeffchen1981-fu/expo-secure-store-android-commit-repro — **section A** of its
README. The same app also reproduces a separate delete-path defect, filed as its own issue.

Created with `npx create-expo-app@latest --template blank` + `npx expo install expo-secure-store`,
plus a four-button `App.js` and an explicit `android.package`. Nothing else, nothing patched.

### Steps to reproduce

Android only. Development build (`npx expo run:android`), not Expo Go — step 2 needs `run-as`, which
requires a debuggable build. npm.

1. Tap **1. write v1**, then **3. read**. This creates the SharedPreferences file. AOSP only takes
   the rename-to-backup route when the file already exists, so a first write into an empty directory
   would not fail the way step 3 needs it to.
2. `adb shell run-as com.example.securestorerepro chmod 500 shared_prefs`
3. Tap **2. write v2**, then **3. read**.
4. `adb shell am force-stop com.example.securestorerepro`, relaunch, tap **3. read**.

Restore with `chmod 700 shared_prefs`.

**Expected:** step 3's `setItemAsync` rejects, since the value did not reach disk.

**Actual**, on a physical Pixel 10 Pro XL (Android 17). `[repro]` lines are the example app logging
whether each promise resolved; the `SharedPreferencesImpl` line is AOSP's:

```
22:20:43.436  27775  [repro] setItemAsync('k', 'v1')
22:20:43.474  27775  [repro]   -> RESOLVED
22:20:44.978  27775  [repro] getItemAsync('k')
22:20:44.997  27775  [repro]   -> "v1"

--- chmod 500 shared_prefs ---

22:21:49.161  27775  [repro] setItemAsync('k', 'v2')
22:21:49.181  27775  E SharedPreferencesImpl: Couldn't rename file
              /data/user/0/com.example.securestorerepro/shared_prefs/SecureStore.xml
              to backup file .../SecureStore.xml.bak
22:21:49.188  27775  [repro]   -> RESOLVED
22:21:51.020  27775  [repro] getItemAsync('k')
22:21:51.042  27775  [repro]   -> "v2"

--- force-stop, relaunch (new pid) ---

22:22:48.005  28036  [repro] getItemAsync('k')
22:22:48.037  28036  [repro]   -> "v1"
```

1. **`setItemAsync` resolved** 7 ms after AOSP logged that the write had failed. Nothing
   distinguishes it from the successful write at `22:20:43`.
2. **The same-process read returned `"v2"`** — the in-memory map holds a value the disk does not.
3. **After a restart the value is `"v1"`.** `v2` never reached disk.

The stored file was unchanged across the failed write: 382 bytes, mtime `22:20`, md5
`8dede71bd3b461a52ddcb0210c5bec9a` both before and after. With `chmod 700` restored, the same tap
resolved, the file moved to mtime `22:23` / md5 `69d495bf300675d2676a496a4abfe194`, and a read after
another restart returned `"v2"` — so the directory permission is the only variable.

### Environment

```text
    System:
      OS: Windows 11 10.0.26200
    Binaries:
      Node: 24.16.0 - C:\Program Files\nodejs\node.EXE
      npm: 11.16.0 - C:\Program Files\nodejs\npm.CMD
    SDKs:
      Android SDK:
        API Levels: 35, 36
        Build Tools: 34.0.0, 35.0.0, 36.0.0, 36.1.0, 37.0.0
        System Images: android-34 | AOSP ATD Intel x86_64 Atom, android-35 | Google APIs Intel x86_64 Atom
    IDEs:
      Android Studio: AI-252.28238.7.2523.14688667
    npmPackages:
      expo: ~57.0.13 => 57.0.13
      react: 19.2.3 => 19.2.3
    Expo Workflow: bare
```

`expo-secure-store@57.0.1`, `minSdk 24` (so the `AESEncryptor` path, not the legacy hybrid one).

### Expo Doctor Diagnostics

```text
```

---

## Where

`packages/expo-secure-store/android/src/main/java/expo/modules/securestore/SecureStoreModule.kt`

> **Updated 2026-08-24.** This section describes `57.0.1`. #49151 has since changed `deleteItemImpl`
> on `main`, so neither the call-site count below nor the line positions of these quotes match it any
> more — see [this comment](https://github.com/expo/expo/issues/48988#issuecomment-5396637981) for
> both. The discarded return itself is unchanged.

**The discarded return, in `setItemImpl`:**

```kotlin
      val encryptedItem = mAESEncryptor.createEncryptedItem(value, secretKeyEntry, options.requireAuthentication, options.authenticationPrompt, authenticationHelper)
      encryptedItem.put(SCHEME_PROPERTY, AESEncryptor.NAME)
      saveEncryptedItem(encryptedItem, prefs, keychainAwareKey, options.requireAuthentication, options.keychainService)
```

**What it discards — `saveEncryptedItem` is declared `: Boolean` and returns `commit()`:**

```kotlin
  private fun saveEncryptedItem(encryptedItem: JSONObject, prefs: SharedPreferences, key: String, requireAuthentication: Boolean, keychainService: String): Boolean {
    ...
    return prefs.edit().putString(key, encryptedItemString).commit()
  }
```

**There are five `.commit()` call sites in this file. The other four are all checked.** The
`value == null` branch of the same function:

```kotlin
    if (value == null) {
      val success = prefs.edit().putString(keychainAwareKey, null).commit()
      if (!success) {
        throw WriteException("Could not write a null value to SecureStore", key, options.keychainService)
      }
```

and all three in `deleteItemImpl`, whose results are accumulated and then checked:

```kotlin
    if (prefs.contains(keychainAwareKey)) {
      success = prefs.edit().remove(keychainAwareKey).commit()
    }

    if (prefs.contains(key)) {
      success = prefs.edit().remove(key).commit() && success
    }

    if (legacyPrefs.contains(key)) {
      success = legacyPrefs.edit().remove(key).commit() && success
    }

    if (!success) {
      throw DeleteException("Could not delete the item from SecureStore", key, options.keychainService)
    }
```

So of the five, four have their results acted on. The fifth is returned out of `saveEncryptedItem`
and then dropped by its only caller — and it is the one whose failure can destroy user data.

## Versions checked

I read the Android source of both:

- `expo-secure-store@57.0.1` (current `latest` on npm)
- `expo-secure-store@58.0.0-canary-20260812-27f94d4` (newest canary at time of writing)

The discarded return is present and identical in both, so there is no released version to upgrade to.

## Why `commit()` can return false

From AOSP `SharedPreferencesImpl`, read on the `main` branch of
`platform/frameworks/base` (`core/java/android/app/SharedPreferencesImpl.java`):

`EditorImpl.commit()` calls `commitToMemory()` first, then waits on the disk write and **returns the
disk result**. Inside `writeToFile` there are three routes that set a `false` disk result:

- the rename-to-backup step fails,
- the `FileOutputStream` could not be created,
- an exception during the write/fsync.

These are genuine "the value is not on disk" outcomes: a full disk, a broken filesystem, a directory
the process can no longer write to. The reproduction above takes the first one.

## Two things that made this trickier than it looks

Notes from working around it in my own app, not a proposal — you will know the constraints here
better than I do. They cost me some time, so they might be worth a look.

**1. `commitToMemory()` has already mutated the in-memory map by the time `commit()` returns, and
nothing rolls it back.** Its edits are applied directly to the map that `getString` later reads,
which is why the same-process read in the transcript above returns `"v2"`. So the stale in-memory
entry is a second symptom, separate from the missing throw.

**2. A `false` is not on its own proof that the write is absent.** `commit()` also returns `false`
from `catch (InterruptedException e)`. When no other disk write is in flight, `enqueueDiskWrite`
runs the write on the calling thread and can have completed — including the fsync — before that
`await` returns; when one is in flight it queues, and the write can still land afterwards. So
restoring the previous value on every `false` could overwrite a value that did reach disk.

Together those suggest that repairing the in-memory entry wants to tell the two cases apart first.
What worked for me was re-committing the *same* item and only repairing if that also fails — on
`main`, `commitToMemory` tracks whether anything actually changed, so a re-put of an equal value is
a no-op and the already-landed case returns `true` cheaply. I have not checked how that behaves on
the oldest API levels this package supports.

I mention it only because checking the `Boolean` and throwing, on its own, would still leave the
in-memory map holding a value the disk does not have.

## Not a case for `apply()`

The usual advice — "if you are ignoring `commit()`'s return value, use `apply()`" — is the wrong
direction here. `apply()` gives the *caller* no programmatic signal: no return value and no
exception, so a failed write cannot be surfaced to JS at all. (Both `commit()` and `apply()` route
through the same `enqueueDiskWrite`/`writeToFile`, so the failure is still logged either way — but a
logcat line is not something the JS API can act on.) Switching to `apply()` would make the silence a
property of a security-storage API rather than a bug.

## Scope

Android only, as far as the persistence signal goes. On iOS the write path does not have this shape:
`set` throws `KeyChainException` when `SecItemAdd` fails, and its `errSecDuplicateItem` branch
delegates to `update`, which throws when `SecItemUpdate` fails — so a failed write is reported to
the caller rather than swallowed.

dennytosp:
# Why

Fixes #48988.

On Android, `SecureStore.setItemAsync` can resolve successfully while the value never reaches disk.

`saveEncryptedItem` returned the result of `SharedPreferences.Editor.commit()`, but `setItemImpl` called it as a bare statement and discarded that `Boolean`. Every other `commit()` in the same file has its result checked — the null-value path a few lines above throws `WriteException` on exactly this condition.

AOSP updates the in-memory map even when `commit()` fails, so reads in the same process keep returning the new value and nothing distinguishes a failed write from a successful one. The consequence is silent data loss: a caller that treats resolution as "persisted" — by then clearing its own copy, for example — loses the value when the process dies.

The reporter's log shows `setItemAsync` resolving 7 ms after AOSP logged that the write had failed, a same-process read returning the new value, and the old value coming back after a force-stop.

# How

Moved `saveEncryptedItem` out of `SecureStoreModule` into a top-level `internal` function that throws `WriteException` instead of returning a `Boolean`, and updated the call site accordingly.

Throwing is already this function's error contract — it throws `WriteException` when `JSONObject#toString()` returns null or empty. Returning a `Boolean` for one failure mode while throwing for another is what let the discarded result go unnoticed, so removing the `Boolean` return closes the whole class of bug rather than just this call site.

Making it top-level also gives it a test seam, matching how `buildAuthenticationPromptInfo` is already a top-level `internal` function covered by `AuthenticationPromptTest`. `KEYSTORE_ALIAS_PROPERTY` went from `private` to `internal` so the extracted function can still read it.

# Test Plan

Added `SaveEncryptedItemTest`, following the Robolectric setup the package already uses. The failing-write case wraps a real `SharedPreferences` in an editor whose `commit()` returns `false`, which is what AOSP does when it can't rename the backing file.

Verified the test fails against the old behaviour before fixing it — with `saveEncryptedItem` back to calling `commit()` without checking it:

```
> There were failing tests.
java.lang.AssertionError: expected expo.modules.securestore.WriteException to be thrown, but nothing was thrown
	at expo.modules.securestore.SaveEncryptedItemTest.throws when the write does not reach disk(SaveEncryptedItemTest.kt:61)

tests="2" skipped="0" failures="1" errors="0"
```

And passes with the fix in place — `et native-unit-tests -p android --packages expo-secure-store`:

```
BUILD SUCCESSFUL
Finished android unit tests successfully.

TEST-expo.modules.securestore.AuthenticationPromptTest.xml   tests="3" failures="0" errors="0"
TEST-expo.modules.securestore.SaveEncryptedItemTest.xml      tests="2" failures="0" errors="0"
TEST-expo.modules.securestore.SecureStoreOptionsTest.xml     tests="4" failures="0" errors="0"
```

`et check-packages expo-secure-store`:

```
🏁 All checks passed.
```

To reproduce the original bug end to end, the issue has a repro app: write a value, `adb shell run-as <pkg> chmod 500 shared_prefs`, write again (resolves today, rejects with this change), then force-stop and read back.

# Checklist

### DIFF
--- a/packages/expo-secure-store/CHANGELOG.md
+++ b/packages/expo-secure-store/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ### 🐛 Bug fixes
 
+- [Android] Reject `setItemAsync` when the write never reached disk, instead of resolving as if the value was persisted. ([#49191](https://github.com/expo/expo/pull/49191) by [@dennytosp](https://github.com/dennytosp))
 - [Android] Fix `deleteItemAsync` resolving when a retried delete left the value on disk. ([#49151](https://github.com/expo/expo/pull/49151) by [@vonovak](https://github.com/vonovak))
 - [iOS] Reject `deleteItemAsync` when the keychain refuses the delete, instead of resolving as if the item was removed. ([#49146](https://github.com/expo/expo/pull/49146) by [@vonovak](https://github.com/vonovak))
 - [iOS] Apply `keychainAccessible` when overwriting an existing item, instead of silently keeping the accessibility it was first stored with. ([#49128](https://github.com/expo/expo/pull/49128) by [@JoRo-Code](https://github.com/JoRo-Code) and [@behenate](https://github.com/behenate))
--- a/packages/expo-secure-store/CHANGELOG.md
+++ b/packages/expo-secure-store/CHANGELOG.md
@@ -11,6 +11,7 @@
 ### 🐛 Bug fixes
 
 - [Android] Reject `setItemAsync` when the write never reached disk, instead of resolving as if the value was persisted. ([#49191](https://github.com/expo/expo/pull/49191) by [@dennytosp](https://github.com/dennytosp))
+- [Android] Roll back the in-memory value when a `setItemAsync` write is rejected, so a subsequent `getItemAsync` returns the value that is actually persisted instead of the one that failed to reach disk. ([#49191](https://github.com/expo/expo/pull/49191) by [@dennytosp](https://github.com/dennytosp))
 - [Android] Fix `deleteItemAsync` resolving when a retried delete left the value on disk. ([#49151](https://github.com/expo/expo/pull/49151) by [@vonovak](https://github.com/vonovak))
 - [iOS] Reject `deleteItemAsync` when the keychain refuses the delete, instead of resolving as if the item was removed. ([#49146](https://github.com/expo/expo/pull/49146) by [@vonovak](https://github.com/vonovak))
 - [iOS] Apply `keychainAccessible` when overwriting an existing item, instead of silently keeping the accessibility it was first stored with. ([#49128](https://github.com/expo/expo/pull/49128) by [@JoRo-Code](https://github.com/JoRo-Code) and [@behenate](https://github.com/behenate))
--- a/packages/expo-secure-store/android/src/main/java/expo/modules/securestore/SecureStoreModule.kt
+++ b/packages/expo-secure-store/android/src/main/java/expo/modules/securestore/SecureStoreModule.kt
@@ -229,22 +229,6 @@ open class SecureStoreModule : Module() {
     }
   }
 
-  private fun saveEncryptedItem(encryptedItem: JSONObject, prefs: SharedPreferences, key: String, requireAuthentication: Boolean, keychainService: String): Boolean {
-    // We need a way to recognize entries that have been saved under an alias created with getExtendedKeychain
-    encryptedItem.put(USES_KEYSTORE_SUFFIX_PROPERTY, true)
-    // In order to be able to have the same keys under different keychains
-    // we need a way to recognize what is the keychain of the item when we read it
-    encryptedItem.put(KEYSTORE_ALIAS_PROPERTY, keychainService)
-    encryptedItem.put(AuthenticationHelper.REQUIRE_AUTHENTICATION_PROPERTY, requireAuthentication)
-
-    val encryptedItemString = encryptedItem.toString()
-    if (encryptedItemString.isNullOrEmpty()) { // JSONObject#toString() may return null
-      throw WriteException("Could not JSON-encode the encrypted item for SecureStore - the string $encryptedItemString is null or empty", key, keychainService)
-    }
-
-    return prefs.edit().putString(key, encryptedItemString).commit()
-  }
-
   private fun deleteItemImpl(key: String, options: SecureStoreOptions) {
     val success = removeItem(
       getSharedPreferences(),
@@ -375,7 +359,7 @@ open class SecureStoreModule : Module() {
     private const val SHARED_PREFERENCES_NAME = "SecureStore"
     private const val KEYSTORE_PROVIDER = "AndroidKeyStore"
     private const val SCHEME_PROPERTY = "scheme"
-    private const val KEYSTORE_ALIAS_PROPERTY = "keystoreAlias"
+    internal const val KEYSTORE_ALIAS_PROPERTY = "keystoreAlias"
     const val USES_KEYSTORE_SUFFIX_PROPERTY = "usesKeystoreSuffix"
     const val DEFAULT_KEYSTORE_ALIAS = "key_v1"
     const val AUTHENTICATED_KEYSTORE_SUFFIX = "keystoreAuthenticated"
@@ -392,4 +376,36 @@ internal fun removeItem(
   val removedFromPrefs = prefs.edit().remove(keychainAwareKey).remove(key).commit()
   val removedFromLegacyPrefs = legacyPrefs.edit().remove(key).commit()
   return removedFromPrefs && removedFromLegacyPrefs
+}
+
+/**
+ * Writes an encrypted item to shared preferences, throwing if it could not be persisted.
+ *
+ * [SharedPreferences.Editor.commit] returns `false` when the write never reached disk — AOSP takes
+ * that path when it can't rename the backing file, for instance. The in-memory map is still updated,
+ * so reads in the same process keep succeeding and the failure is invisible until the process dies.
+ * Reporting it as a [WriteException] keeps a resolved `setItemAsync` meaning "persisted".
+ */
+internal fun saveEncryptedItem(
+  encryptedItem: JSONObject,
+  prefs: SharedPreferences,
+  key: String,
+  requireAuthentication: Boolean,
+  keychainService: String
+) {
+  // We need a way to recognize entries that have been saved under an alias created with getExtendedKeychain
+  encryptedItem.put(SecureStoreModule.USES_KEYSTORE_SUFFIX_PROPERTY, true)
+  // In order to be able to have the same keys under different keychains
+  // we need a way to recognize what is the keychain of the item when we read it
+  encryptedItem.put(SecureStoreModule.KEYSTORE_ALIAS_PROPERTY, keychainService)
+  encryptedItem.put(AuthenticationHelper.REQUIRE_AUTHENTICATION_PROPERTY, requireAuthentication)
+
+  val encryptedItemString = encryptedItem.toString()
+  if (encryptedItemString.isNullOrEmpty()) { // JSONObject#toString() may return null
+    throw WriteException("Could not JSON-encode the encrypted item for SecureStore - the string $encryptedItemString is null or empty", key, keychainService)
+  }
+
+  if (!prefs.edit().putString(key, encryptedItemString).commit()) {
+    throw WriteException("Could not write the encrypted item to SecureStore", key, keychainService)
+  }
 }
--- a/packages/expo-secure-store/android/src/main/java/expo/modules/securestore/SecureStoreModule.kt
+++ b/packages/expo-secure-store/android/src/main/java/expo/modules/securestore/SecureStoreModule.kt
@@ -385,6 +385,12 @@ internal fun removeItem(
  * that path when it can't rename the backing file, for instance. The in-memory map is still updated,
  * so reads in the same process keep succeeding and the failure is invisible until the process dies.
  * Reporting it as a [WriteException] keeps a resolved `setItemAsync` meaning "persisted".
+ *
+ * Throwing alone isn't enough: `SharedPreferencesImpl` mutates the in-memory map *before* it tries
+ * the disk write, so a rejected write would otherwise leave the new value readable for the rest of
+ * the process lifetime while the disk still holds the old one. We therefore roll the in-memory map
+ * back to what was there before rethrowing, so a read after a rejected write returns the value that
+ * is actually persisted.
  */
 internal fun saveEncryptedItem(
   encryptedItem: JSONObject,
@@ -405,7 +411,25 @@ internal fun saveEncryptedItem(
     throw WriteException("Could not JSON-encode the encrypted item for SecureStore - the string $encryptedItemString is null or empty", key, keychainService)
   }
 
+  // Read the currently persisted value first so we can put it back if the write is rejected.
+  // This is racy in the sense that another thread writing the same key between this read and the
+  // rollback below would have its value overwritten by the old one. Concurrent writes to the same
+  // key are already last-writer-wins here, and losing a write that itself never reached disk is
+  // preferable to leaving the in-memory map claiming a value that isn't persisted.
+  val previousItemString = prefs.getString(key, null)
+
   if (!prefs.edit().putString(key, encryptedItemString).commit()) {
+    // `commit()` already mutated the in-memory map, so restore the previous entry before throwing.
+    // This restoring commit reports `false` too — the disk is still broken — and that is fine and
+    // expected: we only need its in-memory effect, which is applied regardless of the disk write.
+    val rollback = prefs.edit()
+    if (previousItemString != null) {
+      rollback.putString(key, previousItemString)
+    } else {
+      rollback.remove(key)
+    }
+    rollback.commit()
+
     throw WriteException("Could not write the encrypted item to SecureStore", key, keychainService)
   }
 }
--- a/packages/expo-secure-store/android/src/test/java/expo/modules/securestore/SaveEncryptedItemTest.kt
+++ b/packages/expo-secure-store/android/src/test/java/expo/modules/securestore/SaveEncryptedItemTest.kt
@@ -1,0 +1,85 @@
+package expo.modules.securestore
+
+import android.content.Context
+import android.content.SharedPreferences
+import org.json.JSONObject
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertThrows
+import org.junit.Assert.assertTrue
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.RuntimeEnvironment
+
+/**
+ * A [SharedPreferences] whose writes never reach disk. AOSP's `SharedPreferencesImpl` behaves this
+ * way when it can't rename the backing file (for example when the `shared_prefs` directory is not
+ * writable): the in-memory map is updated, but `commit()` reports failure.
+ */
+private class NonCommittingPreferences(
+  private val delegate: SharedPreferences
+) : SharedPreferences by delegate {
+  override fun edit(): SharedPreferences.Editor = NonCommittingEditor(delegate.edit())
+}
+
+/**
+ * Every builder method returns `this` rather than the delegate's editor, so a chained
+ * `edit().putString(...).commit()` still lands on this editor's [commit].
+ */
+private class NonCommittingEditor(
+  private val delegate: SharedPreferences.Editor
+) : SharedPreferences.Editor {
+  override fun putString(key: String?, value: String?) = apply { delegate.putString(key, value) }
+  override fun putStringSet(key: String?, values: MutableSet<String>?) = apply { delegate.putStringSet(key, values) }
+  override fun putInt(key: String?, value: Int) = apply { delegate.putInt(key, value) }
+  override fun putLong(key: String?, value: Long) = apply { delegate.putLong(key, value) }
+  override fun putFloat(key: String?, value: Float) = apply { delegate.putFloat(key, value) }
+  override fun putBoolean(key: String?, value: Boolean) = apply { delegate.putBoolean(key, value) }
+  override fun remove(key: String?) = apply { delegate.remove(key) }
+  override fun clear() = apply { delegate.clear() }
+
+  override fun commit(): Boolean = false
+  override fun apply() = Unit
+}
+
+@RunWith(RobolectricTestRunner::class)
+class SaveEncryptedItemTest {
+  private val context: Context
+    get() = RuntimeEnvironment.getApplication()
+
+  private fun preferences(): SharedPreferences =
+    context.getSharedPreferences("SecureStoreTest", Context.MODE_PRIVATE)
+
+  @Test
+  fun `writes the encrypted item under the given key`() {
+    val prefs = preferences()
+
+    saveEncryptedItem(
+      encryptedItem = JSONObject().put("ciphertext", "value"),
+      prefs = prefs,
+      key = "key",
+      requireAuthentication = false,
+      keychainService = "keychain"
+    )
+
+    val stored = JSONObject(prefs.getString("key", null)!!)
+    assertEquals("value", stored.getString("ciphertext"))
+    assertEquals("keychain", stored.getString(SecureStoreModule.KEYSTORE_ALIAS_PROPERTY))
+    assertTrue(stored.getBoolean(SecureStoreModule.USES_KEYSTORE_SUFFIX_PROPERTY))
+  }
+
+  @Test
+  fun `throws when the write does not reach disk`() {
+    val prefs = NonCommittingPreferences(preferences())
+
+    assertThrows(WriteException::class.java) {
+      saveEncryptedItem(
+        encryptedItem = JSONObject().put("ciphertext", "value"),
+        prefs = prefs,
+        key = "key",
+        requireAuthentication = false,
+        keychainService = "keychain"
+      )
+    }
+  }
+}
--- a/packages/expo-secure-store/android/src/test/java/expo/modules/securestore/SaveEncryptedItemTest.kt
+++ b/packages/expo-secure-store/android/src/test/java/expo/modules/securestore/SaveEncryptedItemTest.kt
@@ -4,6 +4,8 @@ import android.content.Context
 import android.content.SharedPreferences
 import org.json.JSONObject
 import org.junit.Assert.assertEquals
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertNull
 import org.junit.Assert.assertThrows
 import org.junit.Assert.assertTrue
 import org.junit.Test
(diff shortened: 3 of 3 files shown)


behenate:
@expo-bot verify

dennytosp:
Rebased onto `main` to resolve the merge conflict.

The conflict was only in `CHANGELOG.md`, from #49205 adding credits to the `deleteItemAsync` entry next to mine. Kept that line as-is with its attribution and reapplied my entry above it — the source and test files rebased cleanly.

Re-ran everything on the rebased tree:

```
AuthenticationPromptTest    tests=3 failures=0 errors=0
SaveEncryptedItemTest       tests=2 failures=0 errors=0
SecureStoreOptionsTest      tests=4 failures=0 errors=0

et check-packages expo-secure-store -> 🏁 All checks passed
```

Ready for review.

jeffchen1981-fu:
Reporter of #48988 here. I ran your patch against the repro from that issue — physical Pixel 10 Pro
XL, Android 17, `expo-secure-store@57.0.1` built from source via
`expo.autolinking.android.buildFromSource` — to check it does what it says. It does: `setItemAsync`
rejects where it used to resolve.

One thing worth knowing, since the docstring stops just short of it. The in-memory map still holds
the new value after the throw, so in the same process a read taken after the rejected write returns
the value that was just reported as not persisted:

```
20:20:25.876  29593  [repro] setItemAsync('k','v1')  -> RESOLVED
20:20:29.223  29593  [repro] getItemAsync('k')       -> "v1"
              disk: 382 bytes, md5 16f655a5ece81fcd88233914ebf1035d

--- chmod 500 shared_prefs (drives the rename-to-backup step into failure) ---

20:20:47.717  29593  [repro] setItemAsync('k','v2')
20:20:47.736  29593  E SharedPreferencesImpl: Couldn't rename file .../SecureStore.xml to backup file ...
20:20:47.740  29593  [repro]   -> REJECTED            <- your change
20:20:52.666  29593  [repro] getItemAsync('k')       -> "v2"   <- the value just rejected
              disk: 382 bytes, md5 16f655a5ece81fcd88233914ebf1035d  (unchanged)

--- force-stop, relaunch (pid 29973) ---
20:21:28.296  29973  [repro] getItemAsync('k')       -> "v1"

--- chmod 700 restored ---
20:21:32.607  29973  [repro] setItemAsync('k','v2')  -> RESOLVED
              disk md5 a2e468f69ce733436b894d09541167a3
--- relaunch (pid 30123) ---
20:21:53.275  30123  [repro] getItemAsync('k')       -> "v2"
```

So a caller that catches the rejection and re-reads to see what is actually stored gets `"v2"` until
the process dies, and `"v1"` after. This is not something the PR introduces — AOSP mutates the map
before it attempts the write and never rolls it back — and I don't think it has to be fixed here.
But after this change "rejected" means "not persisted", not "the previous value is what you will
read back", and that difference is invisible from the API. Worth a sentence in the docstring, or a
follow-up.

Two smaller notes.

`commit()` also returns `false` from its `catch (InterruptedException)`, where the write may already
have completed, fsync included. After this change that surfaces as a rejected `setItemAsync` for a
value that is in fact stored. Harmless — nothing is overwritten and a retry is idempotent — but it
means a rejection is not proof of absence either.

The exception path checks out: `WriteException` extends `CodedException`, and `setItemImpl`'s
`catch (e: CodedException) { throw e }` sits ahead of the generic `catch (e: Exception)`, so the new
throw propagates unwrapped instead of being re-wrapped by the catch-all.

The recovery leg at the end of the log shows the normal path is unaffected.

dennytosp:
@jeffchen1981-fu thanks for running the patch on a real device and, in particular, for chasing it past the point where it "works" — that read-back is a real gap and I've pushed a fix for it in `fcbb2da`.

`SharedPreferencesImpl` mutates the in-memory map *before* it attempts the disk write, so throwing on a `false` from `commit()` left the rejected value readable for the rest of the process. `saveEncryptedItem` now captures the currently stored value first and puts it back — or removes the key, if there was nothing there before — in the failure branch, so a read after a rejected write returns what is actually on disk. The restoring `commit()` returns `false` too, since the disk is still broken; only its in-memory effect matters, and that lands regardless.

One honest caveat, noted in a comment: reading the old value and restoring it is not atomic, so a concurrent write to the same key could be overwritten by the rollback. Writes here are already last-writer-wins, and losing a write that itself never reached disk seemed clearly better than leaving the map claiming a value nobody persisted.

Verified with the repository's own Android unit-test task, the same one CI runs (`:expo-secure-store:testDebugUnitTest`):

```
SaveEncryptedItemTest  4 tests, 0 failures
  PASS writes the encrypted item under the given key
  PASS leaves a key that was never stored absent when the write does not reach disk
  PASS keeps the previously stored value readable when the write does not reach disk
  PASS throws when the write does not reach disk

TOTAL 13 tests, 0 failures   (whole package)
```

Reverting just the rollback fails exactly the two new cases, with the second one reproducing your log:

```
org.junit.ComparisonFailure: expected:<{"ciphertext":"v[1]",...> but was:<{"ciphertext":"v[2]",...>
```

That red run also required fixing the test fake: `NonCommittingEditor` was returning `false` without ever mutating the delegate, so the new tests would have passed vacuously against the buggy code. It now writes through and *then* reports failure, which is what `SharedPreferencesImpl` actually does.

Worth flagging for @behenate / @Kudo: `removeItem` (behind `deleteItemAsync`) has the same memory-before-disk hazard. I left it alone since it is a different path from the one reported here, but it looks like the same fix applies.

jeffchen1981-fu:
`fcbb2da` closes the read-back gap I reported. Same handset and same procedure as my earlier
comment — physical Pixel 10 Pro XL, Android 17, `expo-secure-store@57.0.1` built from source via
`expo.autolinking.android.buildFromSource`.

The source is your PR head verbatim with one deliberate deviation: the
`mAESEncryptor.createEncryptedItem(...)` call reverted to 57.0.1's argument list, because
`AESEncryptor`'s signature changed on `main` after that release and the file otherwise does not
compile against the installed package. `saveEncryptedItem` was not touched.
`:expo-secure-store:compileDebugKotlin` ran and was not up to date.

```
22:52:41.994  24840  [repro] setItemAsync('k','v1')  -> RESOLVED
22:52:44.703  24840  [repro] getItemAsync('k')       -> "v1"
              disk: 382 bytes, md5 24630ae6aab0f74d8cdd5fcc6caeff2a

--- chmod 500 shared_prefs ---

22:53:27.066  24840  [repro] setItemAsync('k','v2')
22:53:27.083  24840  E SharedPreferencesImpl: Couldn't rename file .../SecureStore.xml to backup file ...
22:53:27.084  24840  E SharedPreferencesImpl: Couldn't rename file .../SecureStore.xml to backup file ...
22:53:27.089  24840  [repro]   -> REJECTED
              disk: 382 bytes, md5 24630ae6aab0f74d8cdd5fcc6caeff2a  (unchanged)
22:53:30.766  24840  [repro] getItemAsync('k')       -> "v1"   <- was "v2" before this commit

--- force-stop, relaunch (pid 25309) ---
22:54:02.540  25309  [repro] getItemAsync('k')       -> "v1"

--- chmod 700 restored ---
22:54:06.563  25309  [repro] setItemAsync('k','v2')  -> RESOLVED
              disk md5 e742214c961c92b6f9fce99eb9aafe88
--- relaunch (pid 25453) ---
22:54:31.549  25453  [repro] getItemAsync('k')       -> "v2"
```

For this failure mode the read after the rejected write now returns the value that is actually on
disk, and the recovery leg is unaffected.

One new thing in the log, worth knowing rather than fixing: the rejected write now logs the AOSP
rename failure **twice**, a millisecond apart. The second is the restoring `commit()` re-attempting
the disk write, exactly as your comment says it would. It is also the direct evidence that the
rollback commit really reached the disk layer and really failed there. But it is a visible change —
anything counting that line as "failed writes" will now see two per rejection.

On the tests: fixing `NonCommittingEditor` was the right call, and the vacuous-pass point is real.
Worth stating what the fake still cannot reach, since it is the part the device run covers. Its
delegate is a working `SharedPreferences`, so the delegate write succeeds and only the reported
result is `false`. On a device both commits fail at the disk, and what the rollback rests on is
`commitToMemory()` applying while `writeToDiskResult` is `false`, twice in a row. That is what the
run above exercises, and the doubled rename error is what shows the second commit was not a no-op.

## One case I think should be handled before this lands

The rollback is unconditional — any `false` from `commit()` restores the previous value. But `false`
is not proof that nothing reached disk. `commit()` also returns `false` from
`catch (InterruptedException)`, and on the ordinary path — when no other write to that file is
already in flight — `enqueueDiskWrite(mcr, null)` runs `writeToDiskRunnable.run()` on the calling
thread *before* the `await()`:

```java
            MemoryCommitResult mcr = commitToMemory();

            SharedPreferencesImpl.this.enqueueDiskWrite(
                mcr, null /* sync write on this thread okay */);
            try {
                mcr.writtenToDiskLatch.await();
            } catch (InterruptedException e) {
                return false;
            } finally {
```

([`SharedPreferencesImpl.java` at `android-15.0.0_r1`, lines 603-627](https://android.googlesource.com/platform/frameworks/base/+/refs/tags/android-15.0.0_r1/core/java/android/app/SharedPreferencesImpl.java#603))

So if the calling thread's interrupt flag is already set, `await()` throws even though the latch is
at zero and the write has completed — `AbstractQueuedSynchronizer.acquireSharedInterruptibly` tests
`Thread.interrupted()` before it looks at the state. Throwing clears the flag, so the restoring
`commit()` that follows does not take the same branch, and the disk is healthy on this path. The
rollback therefore **succeeds** and overwrites the value that was just persisted with the previous
one.

I want to be fair about how this compares to the current behaviour, because there is a real argument
on the other side. Before this commit, the interrupt path left `setItemAsync` rejecting while the
new value sat on disk — the map and the disk disagreed, and the rejection was misleading. After it,
map and disk agree and the rejection is accurate. That is a genuine improvement in consistency. The
cost is that reaching it now means destroying a write that had already been fsynced: a caller that
treats the rejection as "not stored" and does not retry has silently lost data it would previously
have kept. For a secure store I would rather report an inconsistent state than quietly discard a
persisted secret, but that is a judgement call and it is yours, not mine — I mainly think it should
be a decision rather than a side effect.

I have not measured this. It is an argument from the AOSP source, not a device log, and I am
flagging that difference explicitly because everything else I have posted in this thread is a
measurement. The smallest thing that would settle it, in the harness you already have: write `v1`
through a real `SharedPreferences`, **then** call `Thread.currentThread().interrupt()`, then call
`saveEncryptedItem` with `v2`, and assert on the file contents afterwards — `v2` is what actually
reached disk, `v1` is what the rollback would leave there. The ordering matters and is easy to get
wrong: `awaitLoadedLocked` swallows `InterruptedException` and clears the flag while the instance is
still loading, so setting the flag before the first access gives a vacuous pass — the same shape as
the `NonCommittingEditor` problem you just fixed.

If it reproduces, the discriminator is available: on the interrupt path the file already holds the
new value, and on a real write failure it does not, so comparing the persisted bytes against the
value just written separates the two cases. The awkward part is that `SharedPreferences` exposes no
way to reload from disk, so "read the file" means parsing the XML outside the abstraction. I do not
have a clean suggestion for that, which is part of why I am raising it rather than proposing a patch.

## `removeItem`

Your note checks out — I measured it on the same build, which carries the merged #49151 code. The
direction is the opposite of this one, and it turns out I had already recorded it in #48989 and
marked it out of scope there rather than it being new, so I have written it up separately as
#49934 rather than widening this PR: after a rejected `deleteItemAsync`, `getItemAsync`
reports the value as absent while it is still on disk.

Two things make that path harder than this one, and both are in the issue. The
`InterruptedException` caveat above applies more sharply, because an unconditional restore there
would write a deleted secret back. And the restore cannot be driven off `removeItem`'s return value
at all: it is a conjunction of two independent commits, one per preferences file, so `false` does
not say which of them failed — restoring both on it can resurrect an entry that was genuinely
deleted.

jeffchen1981-fu:
Following up on the interrupt case above. The conclusion is the one I already stated there; what it
was missing was evidence, and I said so at the time — "an argument from the AOSP source, not a
device log". One step of it is measurable on its own, so I measured that step.

## What I measured

The step the argument turns on is `mcr.writtenToDiskLatch.await()` throwing when the caller's
interrupt flag is already set **and the latch has already reached zero**. That needs no
`SharedPreferences` and no patched module — it is a property of the primitive. On the device this
repro uses (physical Pixel 10 Pro XL, Android 17, API 37), on ART rather than a desktop JVM:

```java
CountDownLatch writtenToDiskLatch = new CountDownLatch(1);
writtenToDiskLatch.countDown();
System.out.println("latch count before await   : " + writtenToDiskLatch.getCount());
Thread.currentThread().interrupt();
System.out.println("interrupt flag before await: " + Thread.currentThread().isInterrupted());
String outcome;
try { writtenToDiskLatch.await(); outcome = "returned normally"; }
catch (InterruptedException e) { outcome = "threw " + e.getClass().getName(); }
System.out.println("await() outcome            : " + outcome);
System.out.println("interrupt flag after await : " + Thread.currentThread().isInterrupted());
```

```
=== on-device ART: Android 17 / API 37 ===
latch count before await   : 0
interrupt flag before await: true
await() outcome            : threw java.lang.InterruptedException
latch count after await    : 0
interrupt flag after await : false
```

```
javac -source 8 -target 8 LatchInterruptProbe.java
d8 --min-api 26 --output . LatchInterruptProbe.class
adb push classes.dex /data/local/tmp/probe.dex
adb shell "CLASSPATH=/data/local/tmp/probe.dex app_process / LatchInterruptProbe"
```

Identical on OpenJDK 21 on the host, for what that is worth.

Two observations, and neither of them is about `commit()` — the probe never calls one:

- A latch already at zero still throws when the flag is set. `CountDownLatch`'s javadoc states both
  behaviours — "If the current count is zero then this method returns immediately", and that an
  interrupted status on entry throws — without saying which wins.
  `AbstractQueuedSynchronizer.acquireSharedInterruptibly` tests `Thread.interrupted()` before it
  looks at the state, and this is that resolution on the runtime in question.
- The throw leaves the flag **clear**. That is the part the rollback depends on, and it is why the
  restoring `commit()` is not protected by the branch that caught the first one.

## What that does and does not establish

Measured: the latch behaviour above.

Read, not run: everything that connects it to `commit()`. At `android-17.0.0_r1` — the device's own
release tag — `commit()` is `SharedPreferencesImpl.java:605-629` and the relevant part of
`enqueueDiskWrite` is `:693-706`, condensed here:

```java
// Typical #commit() path with fewer allocations, doing a write on the current thread.
if (isFromSyncCommit) {
    boolean wasEmpty = false;
    synchronized (mLock) { wasEmpty = mDiskWritesInFlight == 1; }
    if (wasEmpty) { writeToDiskRunnable.run(); return; }
}
QueuedWork.queue(writeToDiskRunnable, !isFromSyncCommit);
```

**The inline write is conditional, and that condition scopes the whole hazard.** `commitToMemory`
increments `mDiskWritesInFlight` (`:544`), and the write runs on the calling thread only when that
count is exactly 1. So:

- **No other write to this file in flight** — the write runs inline and completes, fsync included,
  *before* `await()` is reached. The interrupted `await()` then throws on a zero latch, `commit()`
  returns `false`, and the rollback that follows runs with a clear flag, takes the inline path
  itself, and lands. This is the case where a rejected write destroys a value that was already
  persisted.
- **Another write already in flight** — this commit is queued instead, so when the interrupted
  `await()` returns, nothing has reached disk yet. Both writes then run in order with the rollback
  last, so the final file still holds the previous value, but no fsynced write was destroyed to get
  there.

I should have scoped that in my earlier comment and did not: the "destroying a persisted write"
framing holds for the first case, which is the ordinary one for a single-threaded caller, not for
every interrupted `commit()`.

## The comment states the condition and nothing checks it

In `fcbb2da`:

```kotlin
if (!prefs.edit().putString(key, encryptedItemString).commit()) {
  // `commit()` already mutated the in-memory map, so restore the previous entry before throwing.
  // This restoring commit reports `false` too — the disk is still broken — and that is fine and
  // expected: we only need its in-memory effect, which is applied regardless of the disk write.
  val rollback = prefs.edit()
  ...
  rollback.commit()
```

On the interrupt path the disk is not broken, the restoring commit does not report `false`, and its
effect is not only in memory. The comment names the condition under which the rollback is safe, and
`rollback.commit()`'s result is discarded, so nothing distinguishes the two cases at runtime.

## The same shape is now open in three places

Three call sites in `SecureStoreModule.kt` have the memory-before-disk divergence, and they are
being answered differently:

- **`saveEncryptedItem`** — this PR. Unconditional rollback on `false`; the comment above asserts
  the disk is broken.
- **`removeItem`** — reported as #49934, with #49939 open against it. Same unconditional shape, and
  its "Not covered" discloses this exact case: a delete that did reach the disk is undone, on the
  reasoning that the caller was told it failed, so the contract holds.
- **`setItemImpl`'s `value == null` branch** — I measured this one yesterday and posted it as a
  comment on #49934. Nothing is proposed for it. It is not reachable from `setItemAsync`, which
  rejects null in JS; only from the native boundary, where `setValueWithKeyAsync` declares
  `value: String?`.

So: two answers and one site with none. I do not think either answer is obviously wrong, and this
is a policy question rather than a bug — should "the write was reported as failed" mean "the
previous value is what you will read back", even when reaching that requires discarding something
already persisted? Whatever the answer, it seems worth being the same answer in all three.

## If it is worth settling empirically

The discriminator, in the harness that already exists here: write `v1` through a real
`SharedPreferences`, **then** `Thread.currentThread().interrupt()`, then call `saveEncryptedItem`
with `v2`, and assert on the file contents. `v2` is what reached disk; `v1` is what the rollback
leaves. The ordering trap I flagged before still applies — `awaitLoadedLocked` swallows
`InterruptedException` and clears the flag while the instance is still loading, so setting the flag
before the first access gives a vacuous pass, the same shape as the `NonCommittingEditor` problem.

Happy to write that test if it is wanted here, rather than left riding on the policy question.

PR #49192, opened 2026-08-21
[web][font] Match @font-face rules by their bare family name

### LINKED ISSUE #49092: [expo-font][web] fontFamily matcher vs CSSOM serialization: isLoaded() always false on Firefox; duplicate @font-face injection and SSR hydration mismatch
### Summary

`expo-font`'s web loader decides whether a font family is already registered by comparing the raw family name against the CSSOM serialization `CSSFontFaceRule.style.fontFamily`:

https://github.com/expo/expo/blob/ec40b10f7d7a9f351d65db1fc076c9549c414693/packages/expo-font/src/ExpoFontLoader.web.ts#L43-L54

```ts
return rules.filter(({ rule }) => {
  return (
    rule.style.fontFamily === fontFamilyName && ...
```

But the rules it matches against are written by the library's own `_createWebFontTemplate`, which always quotes the family name via `JSON.stringify`:

https://github.com/expo/expo/blob/ec40b10f7d7a9f351d65db1fc076c9549c414693/packages/expo-font/src/ExpoFontLoader.web.ts#L187-L195

```ts
return `@font-face{font-family:${JSON.stringify(fontFamily)};src:url(...)}`;
```

CSSOM serialization of `font-family` is a gray area where engines disagree about quoting:

- **Firefox** preserves the quotes from the declaration, so for every rule this library writes, `rule.style.fontFamily` comes back as `"MyFont"` (with literal quote characters) and the comparison never matches, for any font.
- **Chromium and WebKit** strip the quotes for identifier-safe names, but keep them for names that need quoting (anything with a space, for example `DIN 2014`). For those names the comparison fails in all three engines.

So `ExpoFontLoader.isLoaded()` is always `false` on Firefox (and in every browser for names needing quotes), which breaks each consumer of the matcher:

1. **`loadAsync` injects a duplicate `@font-face` rule on every call** because it never sees the existing rule ([ExpoFontLoader.web.ts#L142-L145](https://github.com/expo/expo/blob/ec40b10f7d7a9f351d65db1fc076c9549c414693/packages/expo-font/src/ExpoFontLoader.web.ts#L142-L145)). On our production expo-router site, Firefox ends up with 19 rules inside `#expo-generated-fonts` where Chromium and WebKit have 10.
2. **SSR rehydration mismatch with static output.** With expo-router static rendering, fonts registered during the server pass are inlined via `getServerResources()`, and `useFonts` seeds its state synchronously from `isLoaded` on the client precisely so hydration matches the server markup ([FontHooks.ts#L14-L20](https://github.com/expo/expo/blob/ec40b10f7d7a9f351d65db1fc076c9549c414693/packages/expo-font/src/FontHooks.ts#L14-L20), [Font.ts#L23-L30](https://github.com/expo/expo/blob/ec40b10f7d7a9f351d65db1fc076c9549c414693/packages/expo-font/src/Font.ts#L23-L30)). On Firefox the seed is `false` even though the inlined rules are present in the document, so an app that gates any render on `loaded` hydrates different markup. In our app this surfaced as minified React error #418 (hydration mismatch) and a full client-side teardown, on Firefox only. We have since worked around it app-side by not gating web render on `useFonts`.
3. **`unloadAsync` silently removes nothing** on Firefox, since it never matches a rule ([ExpoFontLoader.web.ts#L66-L73](https://github.com/expo/expo/blob/ec40b10f7d7a9f351d65db1fc076c9549c414693/packages/expo-font/src/ExpoFontLoader.web.ts#L66-L73)).
4. **`getLoadedFonts()` returns the serialized names**, so on Firefox it returns names wrapped in literal quote characters, which will never equal the names callers loaded with ([ExpoFontLoader.web.ts#L96-L102](https://github.com/expo/expo/blob/ec40b10f7d7a9f351d65db1fc076c9549c414693/packages/expo-font/src/ExpoFontLoader.web.ts#L96-L102)).

### Minimal repro (no tooling)

Save as an `.html` file, open in Firefox, check the console:

```html
<style>@font-face{font-family:"MyFont";src:url("data:font/woff2;base64,")}</style>
<script>
  const rule = document.styleSheets[0].cssRules[0];
  console.log(JSON.stringify(rule.style.fontFamily)); // Firefox: "\"MyFont\""   Chrome/Safari: "MyFont"
  console.log(rule.style.fontFamily === 'MyFont');    // Firefox: false          Chrome/Safari: true
</script>
```

For a name with a space (`font-family:"DIN 2014"` compared against `'DIN 2014'`) the comparison is `false` in all three engines.

### Engine matrix

Verified by injecting the exact `_createWebFontTemplate` output and reading back `rule.style.fontFamily` (Playwright 1.62.1):

| Engine | family `FlamaUltracondensed-Basic` | family `DIN 2014` |
|---|---|---|
| Chromium 151.0.7922.34 | `FlamaUltracondensed-Basic` (matches) | `"DIN 2014"` (no match) |
| Firefox 153.0 | `"FlamaUltracondensed-Basic"` (no match) | `"DIN 2014"` (no match) |
| WebKit 26.5 | `FlamaUltracondensed-Basic` (matches) | `"DIN 2014"` (no match) |

<details>
<summary>Playwright script</summary>

```js
const pw = require('playwright-core');

const FAMILIES = ['FlamaUltracondensed-Basic', 'DIN 2014'];

async function run(name, launcher) {
  const browser = await launcher.launch();
  const page = await browser.newPage();
  const out = await page.evaluate((families) => {
    const style = document.createElement('style');
    style.id = 'expo-generated-fonts';
    // exact template from expo-font _createWebFontTemplate
    style.textContent = families
      .map(
        (fam) =>
          `@font-face{font-family:${JSON.stringify(fam)};src:url("data:font/woff2;base64,");font-display:auto}`
      )
      .join('');
    document.head.appendChild(style);
    const rules = [...style.sheet.cssRules].filter((r) => r instanceof CSSFontFaceRule);
    return rules.map((r, i) => ({
      family: families[i],
      serialized: r.style.fontFamily,
      expoMatcherHit: r.style.fontFamily === families[i],
    }));
  }, FAMILIES);
  const version = browser.version();
  await browser.close();
  return { engine: name, version, out };
}

(async () => {
  for (const [name, launcher] of [
    ['chromium', pw.chromium],
    ['firefox', pw.firefox],
    ['webkit', pw.webkit],
  ]) {
    const r = await run(name, launcher);
    console.log(`\n${r.engine} ${r.version}`);
    for (const row of r.out) {
      console.log(
        `  family=${JSON.stringify(row.family)}  rule.style.fontFamily=${JSON.stringify(row.serialized)}  matcherHit=${row.expoMatcherHit}`
      );
    }
  }
})();
```

</details>

### Suggested fix

Normalize both sides before comparing, since `"MyFont"` and `MyFont` name the same family:

```ts
const unquote = (name: string) => name.replace(/^['"]|['"]$/g, '');

// in getFontFaceRulesMatchingResource:
unquote(rule.style.fontFamily) === unquote(fontFamilyName)
```

`getLoadedFonts()` should return unquoted names for the same reason. Happy to send a PR if that direction looks right.

### Environment

- `expo` 54.0.33, `expo-font` 14.0.11, `expo-router` 6.0.23 (static web output), `react-native` 0.81.5, `react-native-web` 0.21.2
- The relevant code is unchanged on current `main`; all permalinks above point at ec40b10
- Platform: web only. Observed on a production expo-router static site; the behavior is engine-level, so host OS is irrelevant
- I can put together a minimal Expo repro repo if needed, but the three-line HTML above reproduces the root cause directly and the affected code paths are permalinked

dennytosp:
# Why

Fixes #49092.

`expo-font`'s web loader decides whether a family is already registered by comparing a caller's name against the CSSOM serialization of `CSSFontFaceRule.style.fontFamily`. But the rules it compares against are the ones it wrote itself, and `_createWebFontTemplate` always quotes the family name via `JSON.stringify`.

CSSOM serialization of `font-family` is a gray area where engines disagree about quoting:

- **Firefox** preserves the quotes from the declaration, so for every rule this module writes, `rule.style.fontFamily` reads back as `"MyFont"` — with literal quote characters — and the comparison never matches, for any font.
- **Chromium and WebKit** strip the quotes for identifier-safe names, but keep them for names that need quoting. A family such as `DIN 2014` fails in all three engines.

So `isLoaded()` is always `false` on Firefox, which breaks every consumer of the matcher:

1. `loadAsync` injects a duplicate `@font-face` rule on every call, because it never sees the existing one. The reporter's production expo-router site ends up with 19 rules inside `#expo-generated-fonts` on Firefox where Chromium and WebKit have 10.
2. With static rendering, `useFonts` seeds its state synchronously from `isLoaded` precisely so hydration matches the server pass. On Firefox that seed is `false` even though the inlined rules are present, so an app that gates render on `loaded` hydrates different markup — surfacing as React error #418 and a full client-side teardown, Firefox only.
3. `unloadAsync` silently removes nothing.
4. `getLoadedFonts()` returns names wrapped in literal quote characters, which never equal the names callers loaded with.

# How

Added `_normalizeFontFamilyName`, which returns the bare family name from a CSS `font-family` value: it strips one layer of matching quotes and unescapes the contents, leaving an unquoted value untouched.

`getFontFaceRulesMatchingResource` now normalizes the rule's side before comparing, and `getLoadedFonts` reports normalized names.

Normalizing the rule rather than, say, quoting the caller's name keeps the comparison correct no matter which way the engine chose to serialize, instead of trading Firefox's behaviour for Chromium's.

Only the rule's side is normalized, via `_fontFamilyMatchesRule`. The two strings are not the same kind of value: a rule holds a CSS component value, which an engine may quote and escape, while the caller's name is a literal string. Normalizing the caller's name too would strip quotes and whitespace that can be part of the name itself, so a family loaded as `"Weird"` or `  Padded  ` would stop matching the very rule this module wrote for it — the same `isLoaded`-returns-false failure this PR is fixing, just for rarer names.

# Test Plan

Added `ExpoFontLoader-test.web.ts`, covering each engine's serialization plus a round trip asserting the invariant that actually broke: a name written by `_createWebFontTemplate` normalizes back to the name it was written with.

The tests fail against the current code — 20 failures across both jest projects:

```
Test Suites: 2 failed, 2 total
Tests:       20 failed, 20 total
```

and pass with the fix:

```
PASS Node src/__tests__/ExpoFontLoader-test.web.ts
PASS Web  src/__tests__/ExpoFontLoader-test.web.ts
Tests:       20 passed, 20 total
```

Full package suite is green, so no existing behaviour changed:

```
Test Suites: 19 passed, 19 total
Tests:       29 skipped, 102 passed, 131 total
```

`et check-packages expo-font`:

```
🏁 All checks passed.
```

The issue also has a no-tooling repro: save an `.html` file containing `<style>@font-face{font-family:"MyFont";src:url("data:font/woff2;base64,")}</style>`, open it in Firefox, and read `document.styleSheets[0].cssRules[0].style.fontFamily` — it comes back with the quotes.

# Checklist

### DIFF
--- a/packages/expo-font/CHANGELOG.md
+++ b/packages/expo-font/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 ### 🐛 Bug fixes
 
+- [web] Match `@font-face` rules by their bare family name, so `isLoaded()`, `unloadAsync()` and `getLoadedFonts()` work on Firefox and for family names containing a space. ([#49192](https://github.com/expo/expo/pull/49192) by [@dennytosp](https://github.com/dennytosp))
 - [android] Apply `fontWeight` and `fontStyle` to fonts loaded with `useFonts`, instancing a variable font's `wght` axis at each weight. Bold and italic text previously fell back to a system font. ([#48129](https://github.com/expo/expo/pull/48129) by [@L65FREAD](https://github.com/L65FREAD))
 - [iOS] Fixed `renderToImageAsync` reporting the main screen's scale rather than the scale the image was rendered at. ([#48172](https://github.com/expo/expo/pull/48172) by [@alanjhughes](https://github.com/alanjhughes))
 
--- a/packages/expo-font/src/ExpoFontLoader.web.ts
+++ b/packages/expo-font/src/ExpoFontLoader.web.ts
@@ -40,6 +40,40 @@ function getFontFaceRules(): RuleItem[] {
   return [];
 }
 
+const QUOTED_FONT_FAMILY_RE = /^(["'])([\s\S]*)\1$/;
+
+/**
+ * Returns the bare family name from a CSS `font-family` value.
+ *
+ * Rules written by this module always quote the family name (see `_createWebFontTemplate`), but
+ * engines disagree about how they serialize it back through `CSSFontFaceRule.style.fontFamily`:
+ * Firefox preserves the quotes for every rule, while Chromium and WebKit strip them only for names
+ * that are valid CSS identifiers — a name such as `DIN 2014` reads back quoted everywhere.
+ * Comparing the raw serialization against a caller's name therefore fails for every font on Firefox,
+ * and for quote-requiring names in every engine.
+ */
+export function _normalizeFontFamilyName(fontFamily: string): string {
+  const trimmed = fontFamily.trim();
+  const unquoted = trimmed.match(QUOTED_FONT_FAMILY_RE)?.[2];
+  if (unquoted === undefined) {
+    return trimmed;
+  }
+  return unquoted.replace(/\\([\s\S])/g, '$1');
+}
+
+/**
+ * Whether a rule's serialized `font-family` names the family a caller asked for.
+ *
+ * Only the rule's side is normalized. The two strings are not the same kind of value: the rule
+ * holds a CSS component value, which an engine may quote and escape, while `fontFamilyName` is the
+ * literal family name the caller loaded the font under. Normalizing the caller's name too would
+ * strip quotes and whitespace that are part of the name itself, so a family such as `"Weird"` or
+ * `  Padded  ` would stop matching the very rule this module wrote for it.
+ */
+export function _fontFamilyMatchesRule(ruleFontFamily: string, fontFamilyName: string): boolean {
+  return _normalizeFontFamilyName(ruleFontFamily) === fontFamilyName;
+}
+
 function getFontFaceRulesMatchingResource(
   fontFamilyName: string,
   options?: UnloadFontOptions
@@ -47,7 +81,7 @@ function getFontFaceRulesMatchingResource(
   const rules = getFontFaceRules();
   return rules.filter(({ rule }) => {
     return (
-      rule.style.fontFamily === fontFamilyName &&
+      _fontFamilyMatchesRule(rule.style.fontFamily, fontFamilyName) &&
       (options && options.display ? options.display === (rule.style as any).fontDisplay : true)
     );
   });
@@ -98,7 +132,7 @@ const ExpoFontLoader: Required<ExpoFontLoaderModule> = {
       return getLoadedServerFonts();
     }
     const rules = getFontFaceRules();
-    return rules.map(({ rule }) => rule.style.fontFamily);
+    return rules.map(({ rule }) => _normalizeFontFamilyName(rule.style.fontFamily));
   },
 
   isLoaded(fontFamilyName: string, resource: UnloadFontOptions = {}): boolean {
--- a/packages/expo-font/src/__tests__/ExpoFontLoader-test.web.ts
+++ b/packages/expo-font/src/__tests__/ExpoFontLoader-test.web.ts
@@ -1,0 +1,93 @@
+import {
+  _createWebFontTemplate,
+  _fontFamilyMatchesRule,
+  _normalizeFontFamilyName,
+} from '../ExpoFontLoader.web';
+
+// The family name this module writes into a `@font-face` rule and the family name it later reads
+// back off `CSSFontFaceRule.style.fontFamily` are not the same string. CSSOM serialization of
+// `font-family` is a gray area where engines disagree about quoting:
+//
+// - Firefox preserves the quotes from the declaration, so every rule this module writes reads back
+//   quoted, for any family name.
+// - Chromium and WebKit strip the quotes for identifier-safe names but keep them for names that
+//   need quoting, such as anything containing a space.
+//
+// Normalizing both sides before comparing is what keeps `isLoaded`, `unloadAsync` and
+// `getLoadedFonts` working across engines.
+describe('_normalizeFontFamilyName', () => {
+  it('strips the double quotes Firefox preserves for every rule', () => {
+    expect(_normalizeFontFamilyName('"MyFont"')).toBe('MyFont');
+  });
+
+  it('leaves an unquoted name as Chromium and WebKit serialize it', () => {
+    expect(_normalizeFontFamilyName('MyFont')).toBe('MyFont');
+  });
+
+  it('strips the quotes every engine keeps for a name containing a space', () => {
+    expect(_normalizeFontFamilyName('"DIN 2014"')).toBe('DIN 2014');
+  });
+
+  it('strips single quotes', () => {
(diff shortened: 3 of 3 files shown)


vonovak:
@expo-bot review

dennytosp:
The "most plausible breakage" in this review was a real defect, and both examples it names were reproducible. Fixed in `1a8d72b4`.

I had normalized **both** sides of the comparison. That is wrong, because the two strings are not the same kind of value: `rule.style.fontFamily` is a CSS component value an engine may quote and escape, while the caller's name is a literal string. Normalizing the caller's name strips quotes and whitespace that are allowed to be part of the name itself.

Measured through `_createWebFontTemplate`, so these are the rules this module writes for itself:

| loaded as | rule normalizes to | caller name normalized to | matched |
| - | - | - | - |
| `MyFont` | `MyFont` | `MyFont` | yes |
| `DIN 2014` | `DIN 2014` | `DIN 2014` | yes |
| `"Weird"` | `"Weird"` | `Weird` | **no** |
| `  Padded  ` | `  Padded  ` | `Padded` | **no** |

The last two are the same `isLoaded()`-returns-false, `unloadAsync()`-removes-nothing failure this PR set out to fix, just for rarer names — so the fix was incomplete rather than merely risky.

Note the third column is the only thing that was ever wrong: normalizing the rule already recovers the caller's exact name in every row. The fix is to stop normalizing the caller's name, which is now a named, exported `_fontFamilyMatchesRule` so the asymmetry is explicit and testable.

Added 8 tests covering it, including the two names above and a case asserting that two families differing only by surrounding whitespace stay distinct. They are red on the previous revision — 4 failures from 2 distinct assertions, doubled across the Node and Web jest projects — and green now.

Package suite is 147 tests (118 passed, 29 pre-existing skips) and `et check-packages expo-font` is green.

On the two uncertainties listed: both are fair. No test here drives a real browser, so the per-engine CSSOM quoting is still taken from the issue report rather than measured, and the `document.styleSheets` path remains untested end to end — the test file runs in both the Node and Web jest projects, so it stays on the pure helpers.

vonovak:
hi! apologies there have been multiple PRs around this recently. Im going to close this in favor of https://github.com/expo/expo/pull/49379 and https://github.com/expo/expo/pull/49228

Thank you!

PR #49194, opened 2026-08-21
[ios][autolinking] Refuse to link a prebuilt XCFramework built against another React Native

### LINKED ISSUE #49190: [expo-modules-autolinking][iOS] npm-bundled prebuilt xcframeworks link against any React Native version with no compatibility check

https://github.com/himself65/expo-49007-repro

### Summary

Refiling #49007 with the minimal reproducible example the bot asked for.

iOS prebuilt xcframeworks shipped **inside npm packages** (`<pkg>/prebuilds/output/<flavor>/xcframeworks/`) are linked with **no React Native version check**. `pod install` links the *same* `ExpoModulesCore.xcframework` — byte for byte, identical sha256 — against `React-Core-prebuilt (0.86.0)` and against `React-Core-prebuilt (0.86.2)`, and emits no diagnostic either time.

That binary was compiled against exactly one of them. React Native's C++ types are not ABI-stable across patch releases, so the other link is silent memory corruption at runtime, not a link error. It cost us two crash-on-launch TestFlight builds.

Reproduced on current releases: `expo@57.0.15`, `expo-modules-core@57.0.12`, `expo-modules-autolinking@57.0.10`.

Related but distinct: #46880 covers the same class of problem on a different axis — a prebuilt module binary resolved against an `expo-modules-core` it was not compiled against, caught at the Swift symbol level. This one is the **react-native** axis: the prebuilt `ExpoModulesCore` itself resolved against a React Native core it was not compiled against, which is a C++ layout mismatch and therefore invisible to symbol-level checks. A fix along the lines of that issue's suggestion 2 would not catch this one; the version-keyed lookup that already exists here would.

### Steps to reproduce

Platform: **iOS**. Package manager: **npm**. Environment: bare/prebuild project (`npx expo prebuild` + `pod install`). **No device, no simulator, and no Xcode build are needed** — the mismatch is fully visible in `pod install` output.

```bash
git clone https://github.com/himself65/expo-49007-repro
cd expo-49007-repro
bash scripts/repro.sh 0.86.0
bash scripts/repro.sh 0.86.2
bash scripts/compare.sh
```

Each `repro.sh` run wipes `node_modules/` and `ios/`, pins `react-native` to the given version, then runs `npm install` → `npx expo prebuild --platform ios` → `pod install`, recording the full logs and a `facts.json` under `evidence/rn-<version>/`. About four minutes per run. `0.86.0` and `0.86.2` are the only two stable `0.86.x` releases; `expo@57.0.15` bundles `0.86.2`, so the `0.86.0` run is the skewed one.

**Expected:** the `0.86.0` run declines the npm-bundled prebuilt binaries and builds those modules from source, the way third-party packages already do when their version-keyed lookup misses — or at minimum fails loudly.

**Actual:** both runs link the identical prebuilt binary, with nothing said about it:

```
$ bash scripts/compare.sh
react-native 0.86.0
  expo                          57.0.15
  React core linked             Installing React-Core-prebuilt (0.86.0)
  ExpoModulesCore precompiled   57.0.12
  ExpoModulesCore binary sha256 8de511b810c53e8a88487e8592f9e019eb918ca604be4cbc42fa192e468cd7fa
  prebuild layout               flat

react-native 0.86.2
  expo                          57.0.15
  React core linked             Installing React-Core-prebuilt (0.86.2)
  ExpoModulesCore precompiled   57.0.12
  ExpoModulesCore binary sha256 8de511b810c53e8a88487e8592f9e019eb918ca604be4cbc42fa192e468cd7fa
  prebuild layout               flat

same ExpoModulesCore binary across runs : YES
different React Native core across runs : YES
pod install diagnostics about the skew   : 0 (out of 16 diagnostics in total)

VERDICT: reproduced.
The identical npm-bundled ExpoModulesCore xcframework was linked against two
different React Native core builds, with no version check and no diagnostic.
It can only have been compiled against one of them.
```

The only thing `expo prebuild` says about the skew is a non-fatal line that does not mention that precompiled binaries are involved:

```
› Using react-native@0.86.0 instead of recommended react-native@0.86.2.
```

`npx expo-doctor@latest` on the skewed tree flags the dependency version, but as advice, and again without mentioning prebuilt binaries:

```

Detected Hermes V1 250829098.0.14 from React Native. Hermes V1 250829098.0.15 and earlier are affected by this regression; 250829098.0.16 is the first version that contains the fix.
Advice:

🔧 Patch version mismatches
package       expected  found   
react-native  0.86.2    0.86.0  



1 package out of date.
Advice:

```

#### Where the check is skipped

`expo-modules-autolinking/scripts/ios/precompiled_modules.rb` (line 1508 in `57.0.10`), the bundled-prebuilds fallback:

```ruby
# Fallback: check for prebuilds bundled inside the package directory (shipped in npm)
# Try versioned path first, then flat path for backward compatibility
if !File.directory?(File.join(build_output_dir, build_flavor, 'xcframeworks'))
  bundled_versioned_dir = version_prefix ? File.join(package_root, 'prebuilds', 'output', version_prefix) : nil
  bundled_flat_dir = File.join(package_root, 'prebuilds', 'output')

  if bundled_versioned_dir && File.directory?(bundled_versioned_dir)
    build_output_dir = bundled_versioned_dir
  elsif File.directory?(bundled_flat_dir)
    build_output_dir = bundled_flat_dir      # ← no react-native version anywhere in this path
  end
end
```

`version_prefix` comes from `version_prefix_for_external_package`, which is `<packageVersion>/<reactNativeVersion>/<hermesVersion>` derived from the **installed** `react-native` — exactly the guard that is needed. But `expo-modules-core@57.0.12` ships the **flat** layout:

```
expo-modules-core/prebuilds/output/debug/xcframeworks/ExpoModulesCore.tar.gz
expo-modules-core/prebuilds/output/debug/xcframeworks/ExpoModulesWorklets.tar.gz
expo-modules-core/prebuilds/output/release/xcframeworks/ExpoModulesCore.tar.gz
expo-modules-core/prebuilds/output/release/xcframeworks/ExpoModulesWorklets.tar.gz
```

so `bundled_versioned_dir` never exists and the unguarded `elsif` branch is always taken. Third-party packages take the versioned branch and behave correctly — a mismatch means "not found", and they build from source.

#### Why it is not benign

`bash scripts/abi-delta.sh` in the repro shows the layout change between the two versions:

```diff
  mutable std::unique_ptr<folly::dynamic> nativeProps_DEPRECATED;
+ mutable std::mutex nativePropsMutex;
```

`ShadowNodeFamily::parent_` is declared after that member — line 187 in `v0.86.0`, line 189 in `v0.86.2`. `std::mutex` is 64 bytes on arm64, so `parent_`'s offset and `sizeof(ShadowNodeFamily)` both change. With two layouts in one binary, our release build died on every cold start, ~500 ms in:

```
Thread 10 Crashed:
0  libc++.1.dylib  __libcpp_atomic_refcount_decrement<long>
1  libc++.1.dylib  std::__shared_weak_count::__release_weak() + 16
4  React           std::weak_ptr<ShadowNodeFamily>::operator=<...>
5  React           facebook::react::ShadowNodeFamily::setParent(...)      (ShadowNodeFamily.cpp:42)
6  React           facebook::react::ShadowNode::replaceChild(...)
8  React           YogaLayoutableShadowNode::cloneChildInPlace(unsigned long)
9  React           YogaLayoutableShadowNode::configureYogaTree(...)
23 React           RootShadowNode::layoutIfNeeded(...)
24 React           ShadowTree::tryCommit(...)

Exception Type:    EXC_BAD_ACCESS (SIGBUS)
Exception Subtype: KERN_PROTECTION_FAILURE at 0x0000000109ae9670
Exception Codes:   0x0000000000000002, 0x0000000109ae9670      (write)
VM Region Info:
  __DATA_CONST  109ad8000-109b14000  [240K] r--/rw-  .../React.framework/React
```

The faulting address is `x0 + 0x10`, the weak-refcount field of a `__shared_weak_count` sitting in React.framework's read-only const segment. `parent_` was not holding a heap control block; it was read at the wrong offset, and the atomic decrement hit a read-only page. An earlier build of the same app crashed in the same phase (`ShadowTree::tryCommit` → `RootShadowNode::layoutIfNeeded`) at a completely different instruction inside CoreText during text measurement — which is what a layout mismatch looks like: same phase, wherever the bad pointer lands. Pinning `react-native` back to the bundled version was our fix.

#### Suggested fix

1. Record the React Native (and Hermes) version each npm-bundled prebuild was compiled against — a manifest under `prebuilds/`, or adopt the `<pkgVersion>/<rnVersion>/<hermesVersion>` layout the versioned path already expects — and **fall back to building from source on mismatch**, exactly as the third-party path does today. That alone turns silent corruption into a slower build.
2. Failing that, make the `expo prebuild` version warning fatal, or at least have it say that precompiled binaries are being linked.

#### Appendix: how a project gets into this state

`bash scripts/check-drift.sh` in the repro (npm only, no CocoaPods or Xcode). `expo install --check` validates `react-native` against the version map of the **latest** published SDK, not against the `bundledNativeModules.json` of the `expo` version actually installed — which is the version the shipped binaries were compiled against. With `expo@57.0.8` (bundles `react-native@0.86.0`) and `react-native@0.86.2` installed, it reports react-native as fine and flags only `expo` itself:

```
$ npx expo install --check
The following packages should be updated for best compatibility with the installed expo version:
  expo@57.0.8 - expected version: ~57.0.15
```

We bumped on the strength of `--check`, and that is what walked us into the skew.

### Environment

```
    System:
      OS: macOS 26.6.2
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 26.7.0 - /opt/homebrew/bin/node
      npm: 11.19.0 - /opt/homebrew/bin/npm
    Managers:
    SDKs:
      iOS SDK:
    IDEs:
    npmPackages:
      expo: 57.0.15 => 57.0.15 
      react: 19.2.3 => 19.2.3 
    npmGlobalPackages:
      eas-cli: 21.4.0
    Expo Workflow: managed
```

### Expo Doctor Diagnostics

```
$ npx expo-doctor@latest
```

himself65:
# Why

Fixes #49190. Repro: https://github.com/himself65/expo-49007-repro

The iOS prebuilt xcframeworks shipped inside npm packages are linked with no React Native version check. `pod install` links the *same* `ExpoModulesCore.xcframework` — byte for byte, identical sha256 — against `React-Core-prebuilt (0.86.0)` and against `React-Core-prebuilt (0.86.2)`, and says nothing either time. It was compiled against exactly one of them.

React Native's C++ types are not ABI-stable across patch releases. `0.86.2` inserts a member into `ShadowNodeFamily`:

```diff
  mutable std::unique_ptr<folly::dynamic> nativeProps_DEPRECATED;
+ mutable std::mutex nativePropsMutex;
```

`parent_` is declared after it (line 187 in `v0.86.0`, 189 in `v0.86.2`), so its offset and `sizeof(ShadowNodeFamily)` both move by 64 bytes on arm64. With both layouts in one binary our release build died on every cold start, ~500 ms in, writing through a `parent_` read at the wrong offset into React.framework's read-only `__DATA_CONST`:

```
0  libc++.1.dylib  __libcpp_atomic_refcount_decrement<long>
1  libc++.1.dylib  std::__shared_weak_count::__release_weak() + 16
4  React           std::weak_ptr<ShadowNodeFamily>::operator=<...>
5  React           facebook::react::ShadowNodeFamily::setParent(...)   (ShadowNodeFamily.cpp:42)
...
23 React           RootShadowNode::layoutIfNeeded(...)
24 React           ShadowTree::tryCommit(...)

Exception Type: EXC_BAD_ACCESS (SIGBUS) — KERN_PROTECTION_FAILURE (write)
```

It cost us two crash-on-launch TestFlight builds. Nothing catches it earlier: the build is green, `expo prebuild` only warns non-fatally and does not mention that precompiled binaries are involved, and an ABI mismatch is invisible to compile, link, lint, typecheck and `expo export`.

The guard already exists for third-party packages. `version_prefix_for_external_package` builds `<packageVersion>/<reactNativeVersion>/<hermesVersion>` from the installed React Native, so a mismatch reads as "not found" and those pods build from source. But `expo-modules-core` ships the flat layout:

```
expo-modules-core/prebuilds/output/<flavor>/xcframeworks/ExpoModulesCore.tar.gz
```

so in `pod_lookup_map` the versioned directory never exists and the unguarded fallback is always taken:

```ruby
if bundled_versioned_dir && File.directory?(bundled_versioned_dir)
  build_output_dir = bundled_versioned_dir
elsif File.directory?(bundled_flat_dir)
  build_output_dir = bundled_flat_dir      # ← no react-native version anywhere in this path
end
```

# How

Give the flat layout the one thing it lacks — a record of what the binaries were built against — instead of changing the published path layout.

- **`tools/src/prebuilds/Frameworks.ts`**: `createProductTarballAsync` now writes `prebuilt-versions.properties` into the output directory, holding `reactNativeVersion` and `hermesVersion`. It reuses the existing `VersionStamp` writer, whose `key=value` format `precompiled_modules.rb` already parses with `parse_version_properties`. Every product in a run targets the same React Native, so one record per output directory covers all of its tarballs. Deliberately not a dotfile — it has to survive `npm pack`.
- **`tools/src/publish-packages/tasks/bundleIOSPrebuilds.ts`**: the copy loop took `.tar.gz` only; extracted as `isBundledPrebuildFile` so the record ships with the tarballs it describes.
- **`packages/expo-modules-autolinking/scripts/ios/precompiled_modules.rb`**: `resolve_own_prebuilt_info` reads that record next to the tarball and returns `available: false, reason: :version_mismatch` when it disagrees with the installed React Native or Hermes, which routes into the existing source-build fallback and reason formatting.

**Backward compatible by construction.** No record — every artifact published to date, and every artifact resolved through a version-keyed path or downloaded remotely — resolves exactly as before. A half-known version pair writes nothing at all, since an empty value would read as a mismatch everywhere and push every module to a source build.

I kept this to a version check rather than moving `expo-modules-core` to the versioned layout, since that would change a published package's path layout and break the flat fallback in already-released autolinking. Happy to go the other way if you prefer it.

# Test Plan

Unit tests for both new seams: `tools/src/prebuilds/Frameworks.prebuiltVersions.test.ts` (records both versions; writes nothing when either is unknown) and a new `isBundledPrebuildFile` block in `bundleIOSPrebuilds.test.ts`.

For the consumer half I ran real `pod install`s against the repro app (`expo@57.0.15`, `expo-modules-core@57.0.12`), with this patch applied to the installed `expo-modules-autolinking@57.0.10` and the versions record written by hand, since publishing prebuilds is not something I can do locally:

| `prebuilt-versions.properties` | installed react-native | result |
| --- | --- | --- |
| absent (state of every published package today) | 0.86.2 | `[Expo-precompiled]   📦 ExpoModulesCore (57.0.12)` — unchanged |
| `reactNativeVersion=0.86.2` | 0.86.2 | `[Expo-precompiled]   📦 ExpoModulesCore (57.0.12)` — unchanged |
| `reactNativeVersion=0.86.0` | 0.86.2 | declines, builds from source (below) |

```
[Expo-precompiled] Precompiled modules:
[Expo-precompiled]   ⚠️  ExpoModulesCore (57.0.12) (prebuilt xcframework was built against React Native 0.86.0 (installed: 0.86.2))
[!] [Expo-precompiled] ⚠️  Precompiled modules enabled but no xcframeworks found. All modules will build from source.
```

What I could **not** verify locally, so it deserves a close read: the producer half end to end. Running the prebuild pipeline to actually emit xcframeworks needs the monorepo plus hours of Xcode, so `writePrebuiltVersions` and the bundling filter are covered by the unit tests above rather than by a real published artifact. I also could not run the repo's own `pnpm test` / typecheck (full monorepo install); the touched TypeScript parses clean and the Ruby passes `ruby -c`.

# Checklist


chrfalch:
Hi, @himself65 - could you please go through the response from the expo-bot and update the PR? Would also be nice to rebase it on the current main to pick up the latest changes in the files you've edited. 🙏

PR #49196, opened 2026-08-21
[web][crypto] Respect the bounds of typed-array views in the AES module

dennytosp:
# Why

On web, `Crypto.AES` reads the wrong bytes whenever a caller passes a `Uint8Array` that is a window onto a larger buffer.

Two places reach past a view to its backing buffer and drop `byteOffset` / `byteLength`:

```ts
// SealedData.fromCombined
const buffer = binaryInputBytes(combined).buffer as ArrayBuffer;

// binaryInputBytes
if (ArrayBuffer.isView(input)) {
  return new Uint8Array(input.buffer);
}
```

`BinaryInput` is `string | Uint8Array | ArrayBuffer`, and a `Uint8Array` is very often a window onto something bigger — `subarray()` of a read buffer, a slice of a pooled allocation, a record framed inside a larger payload. Reading `.buffer` off one of those yields the whole allocation, not the 48 bytes the caller meant.

The result is a `SealedData` describing the wrong region entirely. With a 42-byte sealed blob sitting at offset 100 of a 512-byte buffer:

```
combinedSize=512 (expected 42)
decrypt FAILED -> DOMException: The operation failed for an operation-specific reason
```

`combinedSize` reports the backing length, so `iv()` reads offset 0 of the backing buffer instead of the blob's IV, and `tag()` — which is positioned as `combinedSize - tagSize` — lands at the end of the backing buffer. Decryption fails with `OperationError`.

This is silent and input-dependent: the same sealed bytes work when they happen to own their buffer and fail when they are a view, which makes it look like data corruption rather than a library bug.

# How

Keep the view rather than its buffer.

`SealedData` now holds the `Uint8Array` and slices it with `subarray`, so every offset is relative to the sealed data instead of to whatever allocation happens to be underneath. `binaryInputBytes` passes `byteOffset` and `byteLength` through when it wraps a view.

Storing the view rather than normalizing with a copy in `fromCombined` keeps the existing zero-copy behaviour and fixes the getters as a class, instead of patching the one call site that happened to reach `.buffer`.

Data built by `fromParts` and `encryptAsync` is unaffected — both allocate exact-size arrays, where the view and its buffer coincide — so the only behaviour that changes is the cases that were reading the wrong bytes.

`subarray` clamps where `new Uint8Array(buffer, offset, length)` threw, so on its own this change would have made truncated input fail later and less clearly — `decryptAsync` would report a generic AES-GCM `OperationError`, and `tag()` would hand back a short slice instead of raising. `fromCombined` therefore now rejects data that cannot hold its own IV and tag.

That check is not new behaviour for the API, it is the behaviour the other platforms already have. Android throws `InvalidSealedDataConfigException` from [`SealedData`'s `init`](https://github.com/expo/expo/blob/main/packages/expo-crypto/android/src/main/java/expo/modules/crypto/aes/objects/SealedData.kt), and iOS throws out of `AES.GCM.SealedBox(combined:)`. Web was the only platform that accepted a too-short blob, and it failed later with a raw `RangeError: Invalid typed array length: -18`. It now fails at the same point as the other two, with the code Android uses:

```
CodedError [ERR_INVALID_SEALED_DATA_CONFIG]: Sealed data is 10 bytes, which is too short to
hold a 12-byte IV and a 16-byte authentication tag. Pass the whole combined output of
encryptAsync, or set ivLength and tagLength in the config to match how the data was sealed.
```

The `ArrayBuffer.isView` branch is only reachable for views outside the declared `BinaryInput` union (`DataView`, `Int8Array`, …), since `Uint8Array` returns earlier. It is fixed for the same reason: the branch exists precisely to accept those, and it was mishandling every one of them.

# Test Plan

Added `src/aes/__tests__/`, which had no coverage before. Every test below fails on `main` and passes with this change.

`web-utils-test.web.ts` covers `binaryInputBytes` (6 tests), `SealedData-test.web.ts` covers `fromCombined` (4 tests).

Before — 5 distinct assertions fail, reported as 10 because each `.web.ts` file is picked up by both the Node and the Web jest project:

```
● binaryInputBytes › keeps the region of a DataView
    Expected: 3
    Received: 64
● binaryInputBytes › keeps the region of a non-Uint8Array typed view
    Expected: 3
    Received: 64
● SealedData.fromCombined › reads a Uint8Array that views part of a larger buffer
    Expected: 48
    Received: 200
● SealedData.fromCombined › rejects data too short to hold an IV and a tag
● SealedData.fromCombined › rejects data too short for a custom config

Test Suites: 4 failed, 4 total
Tests:       10 failed, 10 passed, 20 total
```

After, with the pre-existing 28 tests still green:

```
Test Suites: 10 passed, 10 total
Tests:       48 passed, 48 total
```

The unit tests stop at the `SealedData` boundary because `crypto.subtle` is present in the Node jest project but not in the jsdom one, so a round trip cannot run in both. To confirm the user-visible effect I ran one directly against Node's WebCrypto, encrypting, framing the sealed bytes at offset 100 of a 512-byte buffer, and decrypting through `fromCombined(view)`:

```
──────── BEFORE the fix ────────
MODE=old  combinedSize=512 (expected 42)
  decrypt FAILED -> DOMException: The operation failed for an operation-specific reason
──────── AFTER the fix ────────
MODE=fixed  combinedSize=42 (expected 42)
  decrypt OK -> "attack at dawn"
```

`et check-packages expo-crypto`:

```
🏁 All checks passed.
```

`tsc -p tsconfig.json --noEmit` exits 0.

# Checklist

### DIFF
--- a/packages/expo-crypto/CHANGELOG.md
+++ b/packages/expo-crypto/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### 🐛 Bug fixes
 
+- [Web] Respect the bounds of typed-array views in `Crypto.AES`, so sealed data passed as a `Uint8Array` that views part of a larger buffer no longer reads the wrong bytes and fails to decrypt. ([#49196](https://github.com/expo/expo/pull/49196) by [@dennytosp](https://github.com/dennytosp))
+- [Web] Reject sealed data that is too short to hold its IV and authentication tag in `AESSealedData.fromCombined()`, matching iOS and Android. ([#49196](https://github.com/expo/expo/pull/49196) by [@dennytosp](https://github.com/dennytosp))
 - Fix `AESSealedData.fromCombined()` throwing on Android when given a base64-encoded string. ([#47317](https://github.com/expo/expo/pull/47317) by [@gabrieldonadel](https://github.com/gabrieldonadel))
 
 ### 💡 Others
--- a/packages/expo-crypto/src/aes/ExpoCryptoAES.web.ts
+++ b/packages/expo-crypto/src/aes/ExpoCryptoAES.web.ts
@@ -1,4 +1,4 @@
-import { registerWebModule, NativeModule } from 'expo';
+import { registerWebModule, NativeModule, CodedError } from 'expo';
 
 import type {
   AESDecryptOptions,
@@ -75,17 +75,27 @@ class EncryptionKey {
 }
 
 class SealedData {
-  private buffer: ArrayBuffer;
+  private bytes: Uint8Array;
   private config: AESSealedDataConfig;
 
-  private constructor(buffer: ArrayBuffer, config: AESSealedDataConfig) {
-    this.buffer = buffer;
+  private constructor(bytes: Uint8Array, config: AESSealedDataConfig) {
+    this.bytes = bytes;
     this.config = config;
   }
 
   static fromCombined(combined: BinaryInput, config?: AESSealedDataConfig): SealedData {
-    const buffer = binaryInputBytes(combined).buffer as ArrayBuffer;
-    return new SealedData(buffer, config ?? defaultConfig);
+    const bytes = binaryInputBytes(combined);
+    const sealedConfig = config ?? defaultConfig;
+    const minimumLength = sealedConfig.ivLength + sealedConfig.tagLength;
+
+    if (bytes.byteLength < minimumLength) {
+      throw new CodedError(
+        'ERR_INVALID_SEALED_DATA_CONFIG',
+        `Sealed data is ${bytes.byteLength} bytes, which is too short to hold a ${sealedConfig.ivLength}-byte IV and a ${sealedConfig.tagLength}-byte authentication tag. Pass the whole combined output of encryptAsync, or set ivLength and tagLength in the config to match how the data was sealed.`
+      );
+    }
+
+    return new SealedData(bytes, sealedConfig);
   }
 
   static fromParts(
@@ -111,7 +121,7 @@ class SealedData {
         ivLength,
         tagLength: tag,
       };
-      return new SealedData(combined.buffer, config);
+      return new SealedData(combined, config);
     }
 
     const tagBytes = binaryInputBytes(tag);
@@ -123,7 +133,7 @@ class SealedData {
     combined.set(ciphertextBytes, ivLength);
     combined.set(tagBytes, totalLength - tagLength);
 
-    return new SealedData(combined.buffer, { ivLength, tagLength });
+    return new SealedData(combined, { ivLength, tagLength });
   }
 
   get ivSize(): number {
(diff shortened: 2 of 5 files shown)


barthap:
@expo-bot check

dennytosp:
Thanks — both points were right, and I've acted on both.

**The count.** You're correct, it is 6, not 4. I had measured the two test files in separate runs and carried the first number into the description. With the two tests I've now added it is 10 failures from 5 distinct assertions. Description corrected.

**Truncated input.** I'd rather not leave this one to a maintainer, because the regression is mine: `subarray` clamps where `new Uint8Array(buffer, offset, length)` threw, so my change is what removed the loud failure.

Looking at the other platforms, web was already the odd one out. Android throws `InvalidSealedDataConfigException` from `SealedData`'s `init` on `content.size < ivSize + tagSize`, and iOS throws out of `AES.GCM.SealedBox(combined:)`. Web accepted the blob and failed later with `RangeError: Invalid typed array length: -18`. So this is a parity gap that predates the PR, and my change would have widened it.

`fromCombined` now rejects too-short data up front, reusing Android's code:

```
CodedError [ERR_INVALID_SEALED_DATA_CONFIG]: Sealed data is 10 bytes, which is too short to
hold a 12-byte IV and a 16-byte authentication tag. Pass the whole combined output of
encryptAsync, or set ivLength and tagLength in the config to match how the data was sealed.
```

Two tests cover it, including a custom-config case that must still be accepted. Suite is 48/48 and `et check-packages expo-crypto` is green.

One thing worth pulling out of your "what I did not cover" list: you noted `convertBinaryInput` passes the view straight to the native module, so the same offset bug may exist on iOS and Android. I agree that is out of scope here, and I'll look at it separately.

barthap:
> One thing worth pulling out of your "what I did not cover" list: you noted convertBinaryInput passes the view straight to the native module, so the same offset bug may exist on iOS and Android. I agree that is out of scope here, and I'll look at it separately.

Please don't. It is handled correctly on native: `Uint8Array` -> `ArrayBuffer` native conversion handles this.

barthap (approved):
Nice fix, and thanks for the thorough description!

I verified the logic on my side: the `subarray` offsets are equivalent to the old ones for exact-size arrays and correct for windowed views, `fromParts` and `encryptAsync` are unaffected, and WebCrypto respects view bounds in the decrypt path. The new length check matches Android exactly (`content.size < ivSize + tagSize`, same `ERR_INVALID_SEALED_DATA_CONFIG` code), so web now fails at the same point as the other platforms instead of throwing a confusing `RangeError` later.

I also double-checked the native conversion to confirm my earlier comment: on Android the JS → `ByteArray` copy applies `byteOffset` and copies exactly `byteLength` bytes, and on iOS `Data` is built from the offset-applied pointer. So the offset bug really is web-only and no follow-up is needed there.

I'll merge once CI passes. Then backport the fix to SDK 56 and SDK 57.

PR #49197, opened 2026-08-21
[web][modules-core] Type `registerWebModule` as returning a module instance

dennytosp:
# Why

`registerWebModule` returns an instance but is typed as returning the class.

```ts
export function registerWebModule<
  EventsMap extends Record<never, never>,
  ModuleType extends typeof NativeModule<EventsMap>,
>(moduleImplementation: ModuleType, moduleName: string): ModuleType {
  ...
  globalThis.expo.modules[moduleName] = new moduleImplementation();
  return globalThis.expo.modules[moduleName];
}
```

The body constructs the class and stores the instance. The function's own documentation agrees:

```
 * @returns A singleton instance of the class passed into arguments.
```

So the runtime behaviour and the documented contract match each other, and the declared return type is the one thing that disagrees. The `return` statements type-check only because `globalThis.expo.modules[moduleName]` is `any`, so nothing catches the mismatch inside the function.

Downstream, TypeScript resolves member access on the result against the class instead of the instance, and every instance member of a web module reads as missing:

```
error TS2339: Property 'count' does not exist on type 'typeof CounterModule'.
error TS2339: Property 'increment' does not exist on type 'typeof CounterModule'.
```

This stays invisible in normal app code, because platform resolution points TypeScript at the native `.ts` declaration rather than the `.web.ts` implementation. It surfaces as soon as anything imports the web implementation directly — which is exactly what a unit test for a web module has to do. I hit it writing tests for `expo-crypto`'s web AES module and had to cast the default export to `InstanceType<typeof Module>` to get past it.

# How

Changed the return type to `InstanceType<ModuleType>`.

This is a type-level change only — no runtime behaviour is touched, and the generic parameter stays as it is so `registerWebModule(SomeClass, 'name')` still infers from the class argument.

# Test Plan

Added `src/__tests__/registerWebModule-test.ts`, which had no coverage before. It registers a module with an instance field and an instance method, then asserts the returned value is an instance, that its members work, and that a second registration under the same name returns the same singleton.

The test is red on `main` at the type level — `tsc -p tsconfig.json --noEmit` in `expo-modules-core`:

```
src/__tests__/registerWebModule-test.ts(18,19): error TS2339: Property 'count' does not exist on type 'typeof CounterModule'.
src/__tests__/registerWebModule-test.ts(19,19): error TS2339: Property 'increment' does not exist on type 'typeof CounterModule'.
src/__tests__/registerWebModule-test.ts(27,11): error TS2339: Property 'increment' does not exist on type 'typeof CounterModule'.
src/__tests__/registerWebModule-test.ts(28,19): error TS2339: Property 'count' does not exist on type 'typeof CounterModule'.
src/__tests__/registerWebModule-test.ts(28,37): error TS2339: Property 'count' does not exist on type 'typeof CounterModule'.
```

and green with this change (exit 0), passing at runtime across all four jest projects:

```
PASS Node    src/__tests__/registerWebModule-test.ts
PASS Web     src/__tests__/registerWebModule-test.ts
PASS Android src/__tests__/registerWebModule-test.ts
PASS iOS     src/__tests__/registerWebModule-test.ts
Tests:       8 passed, 8 total
```

Because this changes a type in a core package, I checked every package that calls `registerWebModule` — `expo`, `expo-app-metrics`, `expo-crypto`, `expo-font`, `expo-image`, `expo-image-manipulator`, `expo-network`, `expo-observe`, `expo-screen-orientation`, `expo-speech`, `expo-sqlite`, `expo-updates` — plus `expo-modules-core` itself:

```
Tasks:    107 successful, 107 total
Tasks:     26 successful,  26 total
🏁 All checks passed.
```

Nothing regressed; the previous type was permissive in a direction no caller was relying on.

# Checklist


dennytosp:
Thanks — that was a real break, and I reproduced it before fixing it.

Rebased onto `main` (`76c37f7e`), which pulls in #49196. Confirmed `TS2344` at `SealedData-test.web.ts(5,67)` on the rebased tree, then removed the workaround as suggested:

```diff
-// `registerWebModule` returns an instance of the module class, so the default
-// export carries the instance members even though it is typed as the class.
-const { SealedData } = AesCryptoModule as unknown as InstanceType<typeof AesCryptoModule>;
+const { SealedData } = AesCryptoModule;
```

The comment only existed to explain the cast, so both go. Also resolved a `CHANGELOG.md` conflict from #49211 landing next to my entry.

After:

```
expo-modules-core  tsc --noEmit -> 0    jest 8 passed, 4 suites
expo-crypto        tsc --noEmit -> 0    jest 48 passed, 10 suites

typecheck sweep of all 13 packages that call registerWebModule -> all OK
et check-packages expo-modules-core expo-crypto -> 55 tasks, All checks passed
```

On the changelog category: I agree the `InstanceType<typeof Module>` case is the awkward one, since it was correct code and `TS2344` is checked on the type argument itself, so no cast silences it. I left the entry under 🐛 Bug fixes because the runtime behaviour is unchanged and the old type was simply wrong, but I have no attachment to that — say the word and I will move it to 🛠 Breaking changes.

packages/expo-modules-core/CHANGELOG.md
```diff
@@ -23,6 +23,7 @@
 ### 🐛 Bug fixes
 
 - [iOS] Fixed the tap that closes a SwiftUI menu still sending `touchStart`, `onPressIn` and `onPressOut` to the React Native view underneath it. React Native's touch handler is now told to skip that tap before UIKit delivers it, instead of being cancelled afterwards. ([#48419](https://github.com/expo/expo/issues/48419) by [@nahooni0511](https://github.com/nahooni0511), [#49775](https://github.com/expo/expo/pull/49775) by [@intergalacticspacehighway](https://github.com/intergalacticspacehighway))
+- [Web] Type `registerWebModule` as returning an instance of the module class rather than the class itself, matching what it returns at runtime. ([#49197](https://github.com/expo/expo/pull/49197) by [@dennytosp](https://github.com/dennytosp))
```
Kudo: **Medium:** Public API type change — consider moving this CHANGELOG entry from 🐛 Bug fixes to 🛠 Breaking changes.

Downstream code that did `InstanceType<typeof result>` (where `result` is the return of `registerWebModule`) was previously valid and now fails with `TS2344` that cannot be cast away. Runtime is unchanged, but the published TypeScript surface is not. Author already offered to re-categorize; maintainer call.

packages/expo-modules-core/src/registerWebModule.ts
```diff
@@ -11,7 +11,7 @@ import type { NativeModule } from './ts-declarations/NativeModule';
 export function registerWebModule<
   EventsMap extends Record<never, never>,
   ModuleType extends typeof NativeModule<EventsMap>,
->(moduleImplementation: ModuleType, moduleName: string): ModuleType {
+>(moduleImplementation: ModuleType, moduleName: string): InstanceType<ModuleType> {
```
Kudo: **Looks correct.** Returning `InstanceType<ModuleType>` matches `new moduleImplementation()` and the JSDoc (`@returns A singleton instance…`). Generic inference from the class argument is preserved.

Callers that previously treated the return as a constructor type were already wrong at runtime; this mainly surfaces that mismatch at compile time (and breaks the old `InstanceType<typeof returnedModule>` workaround pattern — see CHANGELOG note).

Kudo:
@expo-bot review

Kudo (approved):

PR #49203, opened 2026-08-21
[ios][autolinking] Give source-built pods their prebuilt dependencies` headers

chrfalch:
# Why

A prebuilt pod's headers only exist inside its XCFramework. CocoaPods populates `Pods/Headers/Public/<Pod>` only while a pod builds from source. 

This can cause errors with `headers not found` when combining source built and prebuilt pods.

This is an addition and complimentary fix to #49147.

# How

Walk the CocoaPods dependency graph and append each prebuilt dependency's XCFramework header directories to the HEADER_SEARCH_PATHS of the source-built pods that depend on it. Add skip list to avoid injecting header search paths for pods that gets their search paths injected by other mechanisms (ExpoModulesCore, React**).

# Reproduction

- Create Expo app with expo-router
- Use prebuilt 3rd party xcframeworks like we do in EAS build (EXPO_PRECOMPILED_MODULES_BASE_URL)
- Set RNGH to build as source (causes it to try to look for header files in the `Pods/Headers/Public` for Worklets which is prebuilt

**Observe**: Result: exit 65, 'worklets/Compat/StableApi.h' file not found at RNGHRuntimeDecorator.cpp:11

# Test Plan

✅ Run the same as the reproduction above
✅ Run the reproduction above - with `use_frameworks!` (dynamic/static)

# Checklist


tsapeta:
@expo-bot review

tsapeta (approved):

PR #49207, opened 2026-08-21
[cli] Resolve the run:ios dev client scheme from the built binary

Shywim:
# Why

- When having multiple targets in the iOS project, the expo cli fails to launch the built target because it only look for the **first scheme** in the **first Info.plist** listed in the pbxproj.

# How

- This reuse utils to get the exp scheme from the **built** target we want to launch.

# Test Plan

- Create a second iOS target declaring a different scheme in its Info.plist
- Launch the second iOS target using `expo run:ios --scheme <Target name>`
  - Before: the build is successful, but the launch is failing (we can see it tries to launch with another scheme)
  - After: the app is launched successfully

# Checklist

### DIFF
--- a/packages/@expo/cli/CHANGELOG.md
+++ b/packages/@expo/cli/CHANGELOG.md
@@ -23,6 +23,7 @@
 
 ### 🐛 Bug fixes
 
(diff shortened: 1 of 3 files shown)


brentvatne:
@expo-bot check with fable

PR #49209, opened 2026-08-21
[router] Migrate remaining standard navigators

Ubax:
# Why

Migrate remaining navigators to `standard-navigation` integration

# How

1. Migrate `JSStack` and `ExperimentalStack` in a similar way to native-stack - use only standard-navigation API and process any additional data in `createProps`
2. Add `IsWithinNativeNavigator` context which should signal that navigator is rendered inside the native navigator. It is used to detect wether `SplitView` can be rendered - it can only be rendered within JS navigator

# Test Plan

1. CI
2. Router tester

# Checklist


packages/expo-router/src/views/Navigator.tsx
```diff
@@ -123,40 +127,14 @@ export function useNavigatorContext() {
...
-    <ScreenErrorBoundaryContext value={unstable_screenErrorBoundary}>
-      {content}
-    </ScreenErrorBoundaryContext>
-  ) : (
-    content
-  );
+  return focusedRouteKey ? (descriptors[focusedRouteKey]?.render() ?? null) : null;
 }
 
+const RouterSlot = unstable_createStandardRouterNavigator(SlotContent, StackRouter);
```
Ubax: [🚨 **critical**] This regresses `SplitView` nested under a `Slot`. The old `SlotNavigator` did not provide `IsWithinLayoutContext`, which is exactly how `split-view.tsx:23-24` implements its exception ("SplitView cannot be used inside another navigator, except for Slot"). `RouterSlot` goes through `withLayoutContext`, which wraps children in `<IsWithinLayoutContext value>`, so a `SplitView` in a layout below a `Slot` layout — or below `DefaultNavigator` — now throws. Either keep `Slot` from providing this context, or change the split-view detection. A `renderRouter` test for SplitView-under-Slot would lock this in.

HISTORY.md
```diff
@@ -0,0 +1,4 @@
+## 2026-08-21 00:00 — Standard navigation migration
```
Ubax: [🚨 **critical**] This is an AI-session log committed at the repo root. Remove it before merge.

packages/expo-router/src/react-navigation/stack/navigators/createStackNavigator.tsx
```diff
@@ -1,13 +1,14 @@
 'use client';
 import * as React from 'react';
+import { createStandardNavigator } from 'standard-navigation';
```
Ubax: [🏗️ **design**] This file is part of the vendored `@react-navigation/stack` tree, and it now imports expo-router internals (`../../../standard-navigation`) plus the external `standard-navigation` package. That couples the vendored code to expo-router and makes future upstream syncs harder. Is this intended, or should the `createStandardStackNavigator` wrapper live outside `src/react-navigation/` (e.g. next to the other standard-navigation integrations)?

packages/expo-router/src/standard-navigation/subscribePopToTopOnParentTabPress.ts
```diff
@@ -0,0 +1,27 @@
...
+  // @ts-expect-error: there may not be a tab navigator in parent
+  return navigation.addListener?.('tabPress', (event) => {
+    const isFocused = navigation.isFocused();
+
+    requestAnimationFrame(() => {
+      if (
+        state.index > 0 &&
+        isFocused &&
+        !event.defaultPrevented &&
+        event.data?.__internalTabsType !== 'native'
```
Ubax: [🏗️ **design**] Behavior change to confirm: the old JS `StackNavigator` and `ExperimentalStackNavigator` tabPress handlers did not check `event.data?.__internalTabsType !== 'native'` — only `StackClient` did. Sharing the helper means a JS stack under native tabs no longer runs the JS popToTop. That looks intended for consistency, but it isn't mentioned in the changelog entry; consider calling it out.

packages/expo-router/src/react-navigation/stack/navigators/createStackNavigator.tsx
```diff
@@ -53,49 +105,29 @@ function StackNavigator({
...
+      <LegacyStackNavigatorContent
         {...rest}
-        direction={direction}
-        state={state}
+        // Standard-navigation state carries the same runtime shape as the builder state.
+        state={state as unknown as StackNavigatorContentProps['state']}
+        stackState={state}
         descriptors={descriptors}
         navigation={navigation}
+        subscribePopToTopOnParentTabPress={() =>
```
Ubax: [💡 suggestion] This inline arrow is a new function every render, and the content effect depends on it, so the tabPress listener unsubscribes and resubscribes on every render. The old effect only re-ran on `[navigation, state.index, state.key]`. Consider memoizing (e.g. `useCallback` keyed on `navigation`/`state`). Same pattern in `createExperimentalStackNavigator.tsx`. The standard-navigation path is fine because `createProps` output is memoized in `unstable_integrateWithRouter`.

packages/expo-router/src/react-navigation/stack/__tests__/index.test.ios.tsx
```diff
@@ -30,6 +34,10 @@ const interactionManagerWarning =
 
 expectNoUnexpectedWarnings([interactionManagerWarning]);
 
+it('creates a standard navigator', () => {
```
Ubax: [💡 suggestion] The only new test asserts `NavigatorContent` is defined. The migration changes real behavior (Slot rendering path, shared popToTop subscription); consider a `renderRouter` test for SplitView-under-Slot (would have caught the `IsWithinLayoutContext` regression) and a tabPress popToTop test covering the native-tabs guard.

packages/expo-router/CHANGELOG.md
```diff
@@ -90,6 +90,7 @@
 - Derive `useIsFocused` from context. ([#49390](https://github.com/expo/expo/pull/49390) by [@Ubax](https://github.com/Ubax))
 - Base `useNavigationState` on global state. ([#49381](https://github.com/expo/expo/pull/49381) by [@jakub-agent](https://github.com/jakub-agent))
 - Move Expo Router store values into React context. ([#49218](https://github.com/expo/expo/pull/49218) by [@Ubax](https://github.com/Ubax))
+- Migrate Slot, JavaScript Stack, and ExperimentalStack to the `standard-navigation` integration.
```
Ubax: [🐧 nit] Missing the PR link and author suffix that the surrounding entries use: `([#49209](https://github.com/expo/expo/pull/49209) by [@Ubax](https://github.com/Ubax))`.

packages/expo-router/src/layouts/JSStack.tsx
```diff
@@ -2,34 +2,48 @@
...
+} from '../react-navigation/stack/types';
+import { unstable_integrateWithRouter } from '../standard-navigation';
+import { subscribePopToTopOnParentTabPress } from '../standard-navigation/subscribePopToTopOnParentTabPress';
 import { Protected } from '../views/Protected';
 import { Screen } from '../views/Screen';
-import { withLayoutContext } from './withLayoutContext';
-
-const JSStackNavigator = createStackNavigator().Navigator;
-
-// TODO(@ubax): Update docs/pages/router/migrate/from-react-navigation.mdx:387 for the removed prop.
```
Ubax: [🐧 nit] This TODO about updating `docs/pages/router/migrate/from-react-navigation.mdx:387` is removed here, but this PR contains no docs change (a copy of the TODO still exists in `views/Navigator.tsx`). Confirm the docs update is still tracked somewhere.

packages/expo-router/src/layouts/experimental-stack/createExperimentalStackNavigator.tsx
```diff
@@ -10,26 +11,88 @@ import {
...
   ExperimentalStackNavigationEventMap,
+  ExperimentalStackDescriptorMap,
+  ExperimentalStackNavigationHelpers,
   ExperimentalStackNavigationOptions,
   ExperimentalStackNavigationProp,
   ExperimentalStackNavigatorProps,
 } from './types';
 
+export interface ExperimentalStackNavigatorCreateProps {
+  navigation: ExperimentalStackNavigationHelpers;
```
Ubax: Pass these explicitly - which functions are needed rather then the whole navigation object. The props here should be somewhat similar to native stack ones

packages/expo-router/src/layouts/experimental-stack/createExperimentalStackNavigator.tsx
```diff
@@ -10,26 +11,88 @@ import {
...
+  stackState: StackNavigationState<ParamListBase>;
```
Ubax: Can you just use the state passed to every standard navigator?

packages/expo-router/src/layouts/experimental-stack/createExperimentalStackNavigator.tsx
```diff
@@ -10,26 +11,88 @@ import {
...
+
+  const mergedDescriptors = useMemo(
+    () => mergeOptions(descriptors, registry, stackState),
+    [descriptors, registry, stackState]
+  );
+
+  const meta = use(NavigationMetaContext);
+
+  React.useEffect(() => {
+    if (meta && 'type' in meta && meta.type === 'native-tabs') {
```
Ubax: Add the comment back

packages/expo-router/src/react-navigation/stack/__tests__/index.test.ios.tsx
Ubax: Add such test to split view coverage.

Also add `packages/expo-router/src/react-navigation/stack/__tests__/integration/navigation.tsx` where you will test basic navigation from user perspective (renderRouter + useRouter) and see the outcomes. Assert the number of times user screens did render

packages/expo-router/src/react-navigation/stack/navigators/createStackNavigator.tsx
Ubax: Only import from router - unstable_integrate....

packages/expo-router/src/react-navigation/stack/index.tsx
```diff
@@ -6,7 +6,10 @@ import * as TransitionSpecs from './TransitionConfigs/TransitionSpecs';
 /**
  * Navigators
  */
-export { createStackNavigator } from './navigators/createStackNavigator';
+export {
+  createStackNavigator,
+  createStandardStackNavigator,
```
Ubax: I only want to have the standard one. The old should be removed

packages/expo-router/src/standard-navigation/subscribePopToTopOnParentTabPress.ts
Ubax: Keep without comment

packages/expo-router/CHANGELOG.md
Ubax: Add link

HISTORY.md
Ubax: Remove this file

packages/expo-router/src/layouts/experimental-stack/ExperimentalStackView.tsx
```diff
@@ -44,7 +41,13 @@ export function ExperimentalStackView({ state, navigation, descriptors }: Props)
           const descriptor = descriptors[route.key]!;
           const isPreloaded = index > state.index;
           const options = (descriptor.options ?? {}) as ExperimentalStackNavigationOptions;
-          const preventFromContext = isRouteRemovalPrevented(route, routesWithRemovalPrevented);
+          const routeWithState = descriptor.navigation
+            .getState()
```
Ubax: [🏗️ **design**] `getState()` is a `useLatestCallback` over a ref, so this is a mutable read during render and can lag one commit behind the `state` prop. It is needed because the standard state drops nested `state` and `descriptor.route` is the cached route without it.

`createProps` already receives the full builder state (with nested `state`), so the nested-guard lookup could be computed there and passed in as a prop (for example a `Set` of route keys that prevent removal, or a `getRouteWithState(key)` derived from `state.routes`). That keeps the view pure and aligned with concurrent React.

Related: `NativeStackView.native.tsx` calls `isRouteRemovalPrevented(route, ...)` with a route from the standard state, so the native stack and this view now detect nested guards differently. Worth aligning both on one approach.

packages/expo-router/src/react-navigation/stack/navigators/createStackNavigator.tsx
```diff
@@ -1,115 +1,69 @@
...
-    id,
-    initialRouteName,
-    children,
-    layout,
-    screenListeners,
-    screenOptions,
-    screenLayout,
-    UNSTABLE_router,
-  });
+export type StandardStackNavigationEventMap = {
```
Ubax: [💡 suggestion] This mapped type (`& { canPreventDefault: false }` over an event map) now exists three times: `fork/native-stack/createNativeStackNavigator.tsx`, here, and `createExperimentalStackNavigator.tsx`. A shared helper in `standard-navigation/types.ts`, e.g. `NonPreventableEventMap<M>`, would remove the copies.

packages/expo-router/src/views/Navigator.tsx
```diff
@@ -123,40 +128,21 @@ export function useNavigatorContext() {
...
 
-function SlotNavigator({ unstable_screenErrorBoundary, ...props }: NavigatorProps<any>) {
-  const contextKey = useContextKey();
-  const node = useRouteNode();
-
-  // Allows adding Screen components as children to configure routes.
-  const { screens, guardedRedirects } = useFilterScreenChildren([], {
-    contextKey,
-  });
+const SlotOuterLayoutContext = React.createContext(false);
```
Ubax: [💡 suggestion] This context exists only because `withLayoutContext` (inside `unstable_integrateWithRouter`) sets `IsWithinLayoutContext` to `true` before `SlotContent` renders, and `SplitView` relies on `Slot` restoring the outer value. That is not obvious from the code. A short comment here would save the next reader a trip through `split-view.tsx`.

packages/expo-router/src/react-navigation/stack/views/Stack/__tests__/StackView.test.ios.tsx
```diff
@@ -310,6 +312,24 @@ describe('StackView.getDerivedStateFromProps', () => {
...
+      const restoreRoute = jest.fn(() => true);
+      const view = new StackView({ ...createProps([routeA]), restoreRoute });
+      view.state = createState({ closingRouteKeys: ['B'] }, [routeA, routeB]);
+
+      // The handler is private because it is normally called by CardStack.
+      (
+        view as unknown as {
+          handleOpenRoute: (props: { route: Route<string> }) => void;
+        }
+      ).handleOpenRoute({ route: routeB });
```
Ubax: [🐧 nit] The test reaches a private method through a cast. `CardStack` receives `handleOpenRoute` as `onOpenRoute`, so rendering `StackView` and invoking the prop passed to a mocked `CardStack` would exercise the same path without depending on the private name.

packages/expo-router/src/layouts/experimental-stack/ExperimentalStackView.tsx
Ubax: Can't we just use `route` here, same as NativeStack?

packages/expo-router/src/react-navigation/stack/navigators/createStackNavigator.tsx
Ubax: Add it

packages/expo-router/src/react-navigation/stack/views/Stack/__tests__/StackView.test.ios.tsx
Ubax: Do it

packages/expo-router/src/views/Navigator.tsx
Ubax: Change `IsWithinLayoutContext` to `IsWithinNativeNavigator` and set it to `true` for native navigators and propagate

packages/expo-router/src/layouts/experimental-stack/ExperimentalStackView.tsx
```diff
@@ -44,7 +41,13 @@ export function ExperimentalStackView({ state, navigation, descriptors }: Props)
           const descriptor = descriptors[route.key]!;
           const isPreloaded = index > state.index;
           const options = (descriptor.options ?? {}) as ExperimentalStackNavigationOptions;
-          const preventFromContext = isRouteRemovalPrevented(route, routesWithRemovalPrevented);
+          const routeWithState = descriptor.navigation
```
jakub-agent: [🏗️ **design**] `descriptor` here is a `standard-navigation` descriptor, and that contract is only `{ options, render }` — `NavigatorDescriptor` in `standard-navigation@0.0.5` declares nothing else. `descriptor.navigation` and `descriptor.route` exist purely because the values still happen to be react-navigation descriptors, and the `as unknown as ExperimentalStackDescriptorMap` cast in `createExperimentalStackNavigator.tsx:72` is what lets it compile. So this is the one spot where the PR moves away from the abstraction it is adopting: if descriptors ever come from something that honours the declared shape — a `processDescriptors` hook, a future `standard-navigation` version — `descriptor.navigation` is `undefined` and this throws `Cannot read properties of undefined (reading 'getState')` during render.

The root cause looks like `useStandardState`, which maps routes down to `{ href, key, name, params }` and drops `route.state`, which `isRouteRemovalPrevented` needs to walk nested navigators. Since the full builder state *is* available in `createProps`, would it work to compute prevention there and pass it in as a `CreateProps` value, the same way `pop` is passed? The view would then take a plain `preventNativeDismiss` per route key and need no react-navigation access at all.

Two smaller points on the same lines:

- `descriptor.navigation.getState()` returns the state of the navigator that owns this screen, so `.routes.find(key === route.key)` should be the same object as `descriptor.route` — the `?? descriptor.route` fallback suggests that isn't certain. If `descriptor.route` is sufficient, the `getState()` call and the O(n) find per route per render can both go.
- This lookup is the only behaviour change in this file and it has no test. A nested guarded route under `ExperimentalStack` would pin it, and would also catch a regression if the descriptor shape is ever tightened.

packages/expo-router/CHANGELOG.md
```diff
@@ -100,6 +100,7 @@
 - Derive `useIsFocused` from context. ([#49390](https://github.com/expo/expo/pull/49390) by [@Ubax](https://github.com/Ubax))
 - Base `useNavigationState` on global state. ([#49381](https://github.com/expo/expo/pull/49381) by [@jakub-agent](https://github.com/jakub-agent))
 - Move Expo Router store values into React context. ([#49218](https://github.com/expo/expo/pull/49218) by [@Ubax](https://github.com/Ubax))
+- Migrate Slot, JavaScript Stack, and ExperimentalStack to the `standard-navigation` integration. ([#49209](https://github.com/expo/expo/pull/49209) by [@Ubax](https://github.com/Ubax))
```
jakub-agent: [🏗️ **design**] This entry is under **💡 Others**, but the PR removes `createStackNavigator` from `react-navigation/stack/index.tsx`, and that file is re-exported wholesale by `layouts/JSStack.tsx` (`export * from '../react-navigation/stack'`), which is the `./js-stack` entry in `package.json`. So a published export of `expo-router/js-stack` disappears here.

It was marked `@deprecated` with "expect this factory to change or be removed in a future release", so the removal itself looks intended — but it belongs under **🛠 Breaking changes** and should name the removed export, since that is what a consumer greps the changelog for. Worth also flagging in the PR description under an `## Accepted breaking changes` heading.

packages/expo-router/src/react-navigation/stack/index.tsx
```diff
@@ -6,12 +6,7 @@ import * as TransitionSpecs from './TransitionConfigs/TransitionSpecs';
 /**
  * Navigators
  */
-/**
- * @deprecated Reserved for libraries that ship a self-contained navigator, which the `Stack` layout
- * cannot express. There is no stable replacement yet, so expect this factory to change or be removed
- * in a future release. App code should use `Stack` from `expo-router/js-stack`.
- */
-export { createStackNavigator } from './navigators/createStackNavigator';
+export { createStandardStackNavigator } from './navigators/createStackNavigator';
```
jakub-agent: [🏗️ **design**] `createStandardStackNavigator` becomes a public export of `expo-router/js-stack` through the `export *` in `layouts/JSStack.tsx`, but external code cannot use it. Wiring it up needs a `createProps` returning `StackNavigatorCreateProps` (`pop`, `restoreRoute`, `subscribePopToTopOnParentTabPress`), and none of the pieces are reachable: `StackNavigatorCreateProps` is only exported from `navigators/createStackNavigator`, which this barrel does not re-export; and `makePopAction`, `makeRestoreRouteAction`, and `subscribePopToTopOnParentTabPress` have no public entry point. `unstable_integrateWithRouter` and `StackRouter` are public (`src/exports.ts`), so a caller gets far enough to discover the gap and no further.

That matters because the `createStackNavigator` this replaces was explicitly "reserved for libraries that ship a self-contained navigator" — exactly the consumers who now have no path forward.

Either keep it internal (drop it from this barrel and import it directly in `JSStack.tsx`, which is the only caller), or make it genuinely usable: export the `CreateProps` type and the three helpers, and give the factory a doc comment. Right now it reads like an accidental public export.

packages/expo-router/src/layouts/experimental-stack/createExperimentalStackNavigator.tsx
```diff
@@ -72,54 +58,25 @@ function ExperimentalStackNavigator({
...
-        if (state.index > 0 && isFocused && !(e as EventArg<'tabPress', true>).defaultPrevented) {
-          navigation.dispatch({
-            ...StackActions.popToTop(),
-            target: state.key,
-          });
-        }
-      });
-    });
-  }, [meta, navigation, state.index, state.key]);
+    return subscribePopToTopOnParentTabPress();
```
jakub-agent: [💡 suggestion] Is the `meta` guard above this line still needed? `subscribePopToTopOnParentTabPress` already bails on `event.data?.__internalTabsType !== 'native'`, and the new `does not pop to top for native tabPress events` case in `stack/__tests__/integration/navigation.test.ios.tsx` shows `JSStack` under `NativeTabs` behaving correctly with only that check and no `NavigationMetaContext` read.

If the two are equivalent, dropping the guard removes the `use(NavigationMetaContext)` call and the `meta` dependency, and leaves all three stacks (`StackClient`, `JSStack`, `ExperimentalStack`) subscribing identically. If they are *not* equivalent — some native-tabs path emits `tabPress` without `__internalTabsType` — that is worth a short comment here, because the redundancy otherwise reads as leftover.

packages/expo-router/src/react-navigation/stack/types.tsx
```diff
@@ -47,6 +47,22 @@ export type StackNavigationEventMap = {
   gestureCancel: { data: undefined };
 };
 
+export type StackViewEmit = (
```
jakub-agent: [💡 suggestion] `StackViewEmit` restates `StackNavigationEventMap` (declared 3 lines above) by hand, and `ExperimentalStackViewEmit` in `experimental-stack/types.ts` does the same for its event map — making three hand-written copies of this shape counting the existing `NativeStackEmit`. They can drift silently: adding an event to the map does not extend the emit type, so the new event is simply not callable through `emit` and nothing points at why.

Since `StandardStackNavigationEventMap` already exists in `navigators/createStackNavigator.tsx`, the signature is derivable — something like `Parameters<NavigatorArgs<object, StandardStackNavigationEventMap>['emitter']['emit']>[0]`, or one shared `StandardEmit<EventMap>` helper next to `useStandardEmitter` that all three navigators reuse.

packages/expo-router/src/views/Navigator.tsx
jakub-agent: [🐧 nit] This context needs a one-line comment. Nothing here says why it exists, and the reason is not local: `withLayoutContext` hard-codes `<IsWithinLayoutContext value>`, so routing `Slot` through it would start reporting `Slot` as a layout — which would break `SplitView` nested under `Slot`. This context tunnels the outer value past that provider so `Slot` stays transparent. The new `split-view` tests encode the behaviour but not the reason.

Also: `Slot` renders `RouterSlot` directly, and unrecognised props land in `extraProps` and are forwarded to `NavigatorContent`, so this could be an ordinary prop on `RouterSlot` rather than a module-level context. That may not be worth it — `extraProps` is also spread into the router options — but if the context stays, the comment should say why the prop route was rejected.

packages/expo-router/src/layouts/JSStack.tsx
```diff
@@ -2,36 +2,49 @@
...
  * Renders a JavaScript-based stack navigator.
  *
  * @hideType
  */
 const Stack = Object.assign(
-  (props: Omit<ComponentProps<typeof JSStack>, 'initialRouteName'>) => {
-    const routeNode = useRouteNode();
-    return <JSStack {...props} initialRouteName={getValidInitialRouteName(routeNode)} />;
-  },
+  (props: Omit<ComponentProps<typeof JSStack>, 'initialRouteName'>) => <JSStack {...props} />,
```
jakub-agent: [🐧 nit] With `initialRouteName` now resolved inside `unstable_integrateWithRouter` (via `getValidInitialRouteName(useRouteNode())` in `partitionNavigatorProps`), this wrapper is a pure passthrough — its only remaining job is the type-level `Omit`. It still costs an extra component in the tree, and being an anonymous arrow it shows up as `Anonymous` in React DevTools, unlike the `StackClient` wrapper which does real work.

A cast on `Object.assign` would keep the narrowed prop type without the extra layer; failing that, giving it a name would at least keep the component tree readable.

packages/expo-router/src/layouts/experimental-stack/createExperimentalStackNavigator.tsx
```diff
@@ -1,63 +1,49 @@
...
+export type StandardExperimentalStackNavigationEventMap = {
+  [Event in keyof ExperimentalStackNavigationEventMap]: ExperimentalStackNavigationEventMap[Event] & {
+    canPreventDefault: false;
+  };
+};
+
+type ExperimentalStackNavigatorContentProps = NavigatorContentProps<
+  ExperimentalStackNavigationOptions,
+  StandardExperimentalStackNavigationEventMap,
+  Omit<ExperimentalStackNavigatorProps, 'children' | 'id' | 'initialRouteName'>,
```
jakub-agent: [🐧 nit] This `NavigatorProps` argument looks like dead weight. Every key left after the `Omit` (`layout`, `screenOptions`, `screenListeners`, `screenLayout`, `UNSTABLE_router`, plus `StackRouterOptions`) is a builder option that `partitionNavigatorProps` destructures out, so `...rest` in `ExperimentalStackNavigatorContent` is empty at runtime and nothing reaches `ExperimentalStackView` through it.

Relatedly, `experimental-stack/index.tsx:25-26` passes `object` for both `NavigatorProps` and `RouterOptions`, while `StackClient.tsx` passes `NativeStackNavigationConfig` / `StackRouterOptions` and `JSStack.tsx` passes `StackNavigationConfig`. `ExperimentalStack` has no config type to pass, so `object` is accurate — but then this content-side declaration should be `object` too, so the two ends agree.

packages/expo-router/src/react-navigation/stack/views/Stack/CardStack.tsx
```diff
@@ -96,7 +97,9 @@ const STATE_INACTIVE = 0;
 const STATE_TRANSITIONING_OR_BELOW_TOP = 1;
 const STATE_ON_TOP = 2;
 
-const FALLBACK_DESCRIPTOR = Object.freeze({ options: {} as StackNavigationOptions });
+const FALLBACK_DESCRIPTOR = Object.freeze({
```
jakub-agent: [🐧 nit] Unrelated reformat — one line became three with no functional change. Worth reverting to keep the diff to the migration.

jakub-agent (commented):

The migration of `Slot`, the JavaScript `Stack`, and `ExperimentalStack` onto `unstable_integrateWithRouter` is clean and consistent with the already-migrated `StackClient`. Pulling `subscribePopToTopOnParentTabPress` and `makeRestoreRouteAction` out into shared helpers removes three copies of the same closure, and the new `StackViewEmit` / `StackViewState` prop types let `StackView` and `CardStack` drop their react-navigation `navigation` dependency entirely — the JS stack views now only touch `descriptor.options`. Test coverage is preserved or improved: the deleted 299-line `stack/__tests__/index.test.ios.tsx` maps onto the new router-level integration suite case for case, plus new cases for re-render counts, `replace`, JS tab pop-to-top, and native `tabPress`. CI is green.

Three things are worth resolving before merge:

1. `ExperimentalStackView` reaches through the `as unknown as` descriptor cast into `descriptor.navigation.getState()` and `descriptor.route`, neither of which the `standard-navigation` descriptor contract provides. This is the one place the migration moves *away* from the abstraction it is adopting.
2. The changelog entry sits under **💡 Others**, but the PR removes `createStackNavigator` — a published export of `expo-router/js-stack`.
3. Its replacement, `createStandardStackNavigator`, is exported publicly but cannot be wired up externally: the `StackNavigatorCreateProps` it requires and the helpers that build them are all internal.

The rest are suggestions and nits. Note the PR body is still the unmodified template — no Why / How / Test Plan.

I could not execute the test suite locally (the review machine ran out of disk), so the coverage assessment is from reading the diff, not from a run. CI's `jest-ubuntu` and `check-packages` jobs pass on this commit.

packages/expo-router/src/layouts/experimental-stack/ExperimentalStackView.tsx
Ubax: Try to remove the `descriptor.navigation.getState()`. I want standard-nav one here

packages/expo-router/CHANGELOG.md
Ubax: Move to breaking

packages/expo-router/src/react-navigation/stack/index.tsx
Ubax: Export missing

packages/expo-router/src/layouts/experimental-stack/createExperimentalStackNavigator.tsx
Ubax: Do it

packages/expo-router/src/react-navigation/stack/types.tsx
Ubax: Add helper

packages/expo-router/src/exports.ts
```diff
@@ -85,7 +87,7 @@ export type { RouteSource } from './react-navigation/native';
 
 // Router factories for use with `unstable_createStandardRouterNavigator` / `unstable_integrateWithRouter`.
 export { StackRouter, TabRouter } from './react-navigation/routers';
-export { NativeStackView } from './react-navigation/native-stack';
+export { makePopAction, NativeStackView } from './react-navigation/native-stack';
```
Ubax: Why is this export needed?

packages/expo-router/src/react-navigation/stack/index.tsx
Ubax: Instead of exporting each small function like `makePop`, export `createPropsForJSStack` which will create all of them

packages/expo-router/src/views/Navigator.tsx
```diff
@@ -123,40 +128,22 @@ export function useNavigatorContext() {
...
-
-  return unstable_screenErrorBoundary ? (
-    <ScreenErrorBoundaryContext value={unstable_screenErrorBoundary}>
-      {content}
-    </ScreenErrorBoundaryContext>
-  ) : (
-    content
-  );
+  return focusedRouteKey ? (
+    <IsWithinLayoutContext value={outerLayoutContext}>
```
Ubax: Replace `IsWithinLayoutContext` with `IsWithinNativeNavigator`. Then you can remove `SlotOuterLayoutContext`. 

Export `IsWithinNativeNavigator` as public api

packages/expo-router/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### 🛠 Breaking changes
 
+- Remove `createStackNavigator` from `expo-router/js-stack`. Use `createStandardStackNavigator` with `unstable_integrateWithRouter` instead. Migrate Slot, JavaScript Stack, and ExperimentalStack to the `standard-navigation` integration. ([#49209](https://github.com/expo/expo/pull/49209) by [@Ubax](https://github.com/Ubax))
```
jakub-agent: [🏗️ **design**] This entry now covers the `createStackNavigator` removal correctly, but two other user-visible changes in this PR aren't recorded.

**A behaviour change that belongs in this breaking entry.** The `SplitView` nesting guard moved from `IsWithinLayoutContext` (any enclosing layout) to `IsWithinNativeNavigator` (native navigators only). So `<SplitView>` inside a JavaScript navigator — `expo-router/js-stack`, JS `Tabs`, `Drawer`, `TopTabs` — used to throw and now renders; your own new test `renders a SplitView nested under a JavaScript navigator` pins the new behaviour. That is the right fix for the `TODO: Add better way of detecting if SplitView is rendered inside Native navigator` this replaces, but someone relying on the old error to catch a bad layout gets silence instead. The specific `There can only be one SplitView in the navigation hierarchy.` message is also gone (see the `split-view.tsx` comment).

**New public API that needs a 🎉 New features entry.** `IsWithinNativeNavigator` and the `StandardNavigatorEmit` type are added to the root entry (`exports.ts`), and `createPropsForJSStack`, `StackNavigatorCreateProps`, `StandardStackNavigationEventMap`, `StackNavigationConfig`, `StackViewEmit`, and `StackViewState` become reachable from `expo-router/js-stack`. `IsWithinNativeNavigator` in particular is the documented contract third-party native navigators are now expected to provide, so it wants an entry people can find.

packages/expo-router/src/standard-navigation/useStandardState.ts
```diff
@@ -10,10 +10,8 @@ export function useStandardState(builderState: NavigationState): NavigatorState
     return {
       index: builderState.index,
       routes: builderState.routes.map<NavigatorState['routes'][number]>((route) => ({
+        ...route,
```
jakub-agent: [🏗️ **design**] Spreading the whole builder route means the value carries `state` and `path`, which `NavigatorState['routes'][number]` does not declare — it declares only `href`, `key`, `name`, `params`. Your new test names why (`preserves route state needed to restore cancelled transitions`), and it is a real requirement: `makeRestoreRouteAction` feeds these route objects straight back through `CommonActions.reset`, so a route stripped to the declared shape would silently lose its nested navigator state when a close animation is cancelled.

The concern is that the dependency is undeclared and invisible at the call site. It is the same failure mode as the descriptor leak fixed in `ExperimentalStackView`, just relocated from the view into the state adapter: today it works because the routes happen to be react-navigation routes; if `standard-navigation` ever constructs routes to its own published type, swipe-back-cancel loses nested state with no type error and no crash — just wrong state.

A comment here pointing at `makeRestoreRouteAction` would be enough to keep the next reader from "simplifying" this back to a pick. Better still would be declaring the fields the integration actually relies on, so the requirement is checked rather than remembered.

packages/expo-router/src/react-navigation/stack/navigators/createStackNavigator.tsx
```diff
@@ -1,115 +1,72 @@
...
-        descriptors={descriptors}
-        navigation={navigation}
-      />
-    </NavigationContent>
+    <StackView
+      {...rest}
+      direction={direction}
+      state={state}
+      // Standard descriptors have the same runtime shape as stack descriptors.
+      descriptors={descriptors as unknown as StackDescriptorMap}
```
jakub-agent: [💡 suggestion] The same cast you removed from `ExperimentalStackView` is still here, and the JS stack does not need it either. I checked every descriptor access under `stack/views/`: it is `descriptor.options` (15 sites) plus `scene.descriptor.render()` in `CardContainer.tsx:277` — nothing reads `descriptor.route` or `descriptor.navigation`. That is exactly `NavigatorDescriptor<StackNavigationOptions>`.

So `StackView` / `CardStack` / `CardContainer` could take `Record<string, NavigatorDescriptor<StackNavigationOptions>>` the way `ExperimentalStackView` now does, and this cast plus the `StackDescriptorMap` import would go. It is a wider change than the experimental one (the `Scene` type threads the descriptor through three files), so it may be worth its own commit — but leaving the two sibling navigators on different descriptor contracts means the next person has to work out which one is authoritative.

The comment is also not accurate as written: the standard descriptor is a strict subset of the stack descriptor, not the same shape. If the cast stays, saying "standard descriptors are a subset — the views only read `options` and `render`" would explain why it is safe.

packages/expo-router/src/split-view/split-view.tsx
```diff
@@ -21,13 +13,8 @@ export interface SplitViewProps extends Omit<SplitHostProps, 'children'> {
...
 
 function SplitViewNavigator({ children, ...splitViewHostProps }: SplitViewProps) {
-  if (use(IsWithinSplitViewContext)) {
-    throw new Error('There can only be one SplitView in the navigation hierarchy.');
-  }
-
-  // TODO: Add better way of detecting if SplitView is rendered inside Native navigator.
-  if (use(IsWithinLayoutContext)) {
-    throw new Error('SplitView cannot be used inside another navigator, except for Slot.');
+  if (use(IsWithinNativeNavigator)) {
```
jakub-agent, line 16: [💡 suggestion] Collapsing `IsWithinSplitViewContext` into `IsWithinNativeNavigator` is a good simplification — one context instead of two, and `SplitView` providing `value` means a nested `SplitView` still throws. But the diagnostic regressed: `There can only be one SplitView in the navigation hierarchy.` is gone, and `rejects a SplitView nested inside a SplitView column` now asserts the generic `SplitView cannot be used inside another native navigator.` A developer who has accidentally nested two `SplitView`s reads a message that does not name what they actually did.

Cheapest fix that keeps the single context is to widen the message, e.g. `SplitView cannot be used inside another native navigator, including another SplitView.` If you want the specific message back, the old dedicated context is the only way — probably not worth it for the extra provider.

packages/expo-router/src/layouts/JSStack.tsx
```diff
@@ -1,37 +1,63 @@
...
+  createStandardStackNavigator,
+  type StackNavigationConfig,
+  type StackNavigatorCreateProps,
+  type StackNavigationOptions,
+  type StandardStackNavigationEventMap,
+} from '../react-navigation/stack';
+import { makeRestoreRouteAction } from '../react-navigation/stack/utils/makeRestoreRouteAction';
+import { unstable_integrateWithRouter } from '../standard-navigation';
+import { subscribePopToTopOnParentTabPress } from '../standard-navigation/subscribePopToTopOnParentTabPress';
+import type { StandardNavigatorCreatePropsFactoryDeps } from '../standard-navigation/types';
```
jakub-agent, line 18: [🐧 nit] `StandardNavigatorCreatePropsFactoryDeps` is reached by deep path here and is not re-exported from `standard-navigation/index.tsx` or `exports.ts`, so it is not public. `createPropsForJSStack` covers the JS stack case, so this is not a gap for that path — but anyone using the now-public `createStandardStackNavigator` to build their own navigator (the audience the removed `createStackNavigator` was reserved for) cannot name the parameter type of their own `createProps` factory. If the factory shape is part of the integration API, it belongs next to `IntegrateWithRouterOptions` in the public type list; if it is not, the deep import is worth a short note saying so.

packages/expo-router/src/react-navigation/stack/index.tsx
jakub-agent: [🐧 nit] `createStandardStackNavigator` and `createPropsForJSStack` land in the public surface without the `unstable_` prefix that everything else in this integration carries (`unstable_integrateWithRouter`, `unstable_createStandardRouterNavigator`). Since they are only usable *through* an unstable API, the unprefixed names imply a stability guarantee the integration itself does not make — and the export they replace was explicitly documented as unstable. Worth prefixing them, or noting in the doc comment that they follow the stability of `unstable_integrateWithRouter`.

jakub-agent (approved):

**Iteration 2** — re-reviewed at `9590c24`. The base moved from `main` to `tuft/propagate-prevent-remove`, so the diff is now 36 files (+962/−775) and includes the `SplitView` / native-tabs changes.

Every finding from the first pass is resolved, several of them better than what I suggested:

- **Descriptor abstraction leak** — fully gone. `ExperimentalStackView` now types `descriptors` as `Record<string, NavigatorDescriptor<…>>`, drops the `as unknown as` cast, and uses only `descriptor.options`. `descriptor.navigation.getState()` is gone; `preventNativeDismiss` reads the flat `routesWithRemovalPrevented` set. I verified the base branch propagates prevention to every ancestor route key (`useRouteRemovalPreventionSetter` calls both the global and the parent setter), so the flat check is equivalent to the recursive walk it replaced — and `isRouteRemovalPrevented` no longer exists, so this is aligned with the base branch's design rather than working around it. The nested case is now tested (`experimental-stack.test.ios.tsx` asserts `preventNativeDismiss === true` for a `usePreventRemove` guard in a nested route).
- **Changelog classification** — moved to 🛠 Breaking changes and now names both the removed export and its replacement.
- **Half-exported replacement factory** — resolved by exporting one complete `createPropsForJSStack` factory plus `StackNavigatorCreateProps` / `StandardStackNavigationEventMap`, and `JSStack-exports.test.tsx` locks that surface down (asserting the loose helpers stay private). Cleaner than exporting three helpers.
- **Redundant native-tabs guard** — `meta` / `NavigationMetaContext` removed, with a new `does not pop to top for native tabPress events` test for `ExperimentalStack`.
- **Hand-written emit types** — replaced by one `StandardNavigatorEmit<EventMap>` helper, applied to all three copies including the pre-existing `NativeStackEmit`.
- **Nits** — the `JSStack` passthrough is now a cast on `Object.assign` with a comment; the `CardStack` reformat is reverted; the dead `ExperimentalStackNavigatorProps` / descriptor types are deleted. The `SlotOuterLayoutContext` tunnel was dropped entirely in favour of a documented, public `IsWithinNativeNavigator` — a better answer than the prop I proposed, and it also lets `split-view.tsx` retire its second context.

CI is green on this commit: 22 checks pass, 6 skipped, none failing.

Remaining items are all non-blocking. The one I would fix before merge is the changelog (**1**): this PR changes when `SplitView` throws and adds seven new public exports, and neither is recorded. **2** is a contract gap on behaviour that is correct and tested today. The rest are consistency cleanups.

packages/expo-router/CHANGELOG.md
Ubax:
```suggestion
- Remove `createStackNavigator` from `expo-router/js-stack`. Use `createStandardStackNavigator` with `unstable_integrateWithRouter` instead. ([#49209](https://github.com/expo/expo/pull/49209) by [@Ubax](https://github.com/Ubax))
```

and add new entry to New features

packages/expo-router/src/standard-navigation/useStandardState.ts
Ubax: Is it needed?
Or can we simply keep the old approach?

packages/expo-router/src/react-navigation/stack/navigators/createStackNavigator.tsx
Ubax: Remove the cast

packages/expo-router/src/split-view/split-view.tsx
Ubax: The generic message is enough

packages/expo-router/src/layouts/JSStack.tsx
Ubax: Ignore

packages/expo-router/src/react-navigation/stack/index.tsx
Ubax: Do
>Worth prefixing them, or noting in the doc comment that they follow the stability of unstable_integrateWithRouter.

PR #49211, opened 2026-08-21
[ui][ios] Yoga node holding old styles after props style update

intergalacticspacehighway:
# Why

Reported by @matinzd https://github.com/expo/expo/pull/48059#issuecomment-5368086472

In `adopt` function, the props style gets updated but yoga node's style can have stale value in some cases. e.g. failed commits. We need to call `updateYogaProps` to make sure props are always copied to yoga node's style which is used during the layout.

# How

Call `updateYogaProps` after prop style changes it also calls `setDirty` internally, so another call to `setDirty` is not needed.

# Test Plan

Added testcases in NCL test suite which should catch such issues in future.

# Checklist

### DIFF
--- a/apps/test-suite/TestModules.ts
+++ b/apps/test-suite/TestModules.ts
@@ -84,6 +84,7 @@ export function getTestModules() {
   }
 
   if (['android', 'ios'].includes(Platform.OS)) {
+    modules.push(optionalRequire(() => require('./tests/ExpoUIHostSize')));
     modules.push(require('./tests/AppMetrics'));
     modules.push(require('./tests/Blob'));
     modules.push(require('./tests/FileSystem'));
--- a/apps/test-suite/TestModules.ts
+++ b/apps/test-suite/TestModules.ts
@@ -84,7 +84,7 @@ export function getTestModules() {
   }
 
   if (['android', 'ios'].includes(Platform.OS)) {
-    modules.push(optionalRequire(() => require('./tests/ExpoUIHostSize')));
+    modules.push(require('./tests/ExpoUIHostSize'));
     modules.push(require('./tests/AppMetrics'));
     modules.push(require('./tests/Blob'));
     modules.push(require('./tests/FileSystem'));
--- a/apps/test-suite/tests/ExpoUIHostSize.tsx
+++ b/apps/test-suite/tests/ExpoUIHostSize.tsx
@@ -2,6 +2,7 @@ import { Column, Host, Text as UIText } from '@expo/ui';
 import React from 'react';
 import { View } from 'react-native';
 
+import type { JasmineInterface, TestPortal } from '../types';
 import { mountAndWaitForWithTimeout } from './helpers';
 
 export const name = 'ExpoUI Host size';
@@ -64,8 +65,8 @@ function HostSizeProbe({
 }
 
 export async function test(
-  { it, describe, expect, afterEach }: any,
-  { setPortalChild, cleanupPortal }: any
+  { it, describe, expect, afterEach }: JasmineInterface,
+  { setPortalChild, cleanupPortal }: TestPortal
 ) {
   const measure = (matchContents: boolean | { vertical?: boolean; horizontal?: boolean }) =>
     mountAndWaitForWithTimeout<Measured>(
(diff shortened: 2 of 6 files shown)


matinzd:
I tested the patch and it works. I also suspected that this is causing stale style updates. Thanks Nishan :)

packages/expo-modules-core/common/cpp/fabric/ExpoViewComponentDescriptor.h
```diff
@@ -67,29 +67,19 @@ class ExpoViewComponentDescriptor
...
+        style.setDimension(facebook::yoga::Dimension::Height,
+                           facebook::yoga::StyleSizeLength::points(styleHeight));
       }
 
-      // Update yoga props and dirty layout if we changed the style
-      if (changedStyle) {
-        snode->updateYogaProps();
-        snode->dirtyLayout();
-      }
+      // Updates yoga style from props and sets the node dirty
```
Kudo: [🐧 nit] The comment reads as if `updateYogaProps()` always dirties the node. It dirties only when the style really differs, and that conditional part is exactly why dropping `dirtyLayout()` keeps #48059's no-op behavior. Worth spelling out so nobody adds `dirtyLayout()` back.

```suggestion
      // Copies the props style onto the yoga node. It dirties the node only when the style
      // actually differs, so no extra `dirtyLayout()` call is needed.
```

apps/test-suite/tests/ExpoUIHostSize.tsx
```diff
@@ -0,0 +1,106 @@
...
+) {
+  const measure = (matchContents: boolean | { vertical?: boolean; horizontal?: boolean }) =>
+    mountAndWaitForWithTimeout<Measured>(
+      <HostSizeProbe matchContents={matchContents} onMeasured={() => {}} />,
+      'onMeasured',
+      setPortalChild,
+      TIMEOUT_MS
+    );
+
+  describe(name, () => {
```
Kudo, line 136: [💡 suggestion] Do these three specs fail with the `ExpoViewComponentDescriptor.h` change reverted? They read as happy-path measurements, while the bug needs a stale yoga style (failed/rejected commit). If they pass either way they document `matchContents` but won't guard this fix.

apps/test-suite/TestModules.ts
```diff
@@ -84,6 +84,7 @@ export function getTestModules() {
   }
 
   if (['android', 'ios'].includes(Platform.OS)) {
+    modules.push(require('./tests/ExpoUIHostSize'));
```
Kudo: [💡 suggestion] This registers the suite on Android too, but the fix is iOS-only and Android goes through a different path (`updateStyleSizeImmediate` replaces the whole state instead of merging channels). Have you run the three specs on an Android emulator? If not, gating on iOS for now avoids a red CI nobody expected.

apps/test-suite/tests/ExpoUIHostSize.tsx
```diff
@@ -0,0 +1,106 @@
...
+          <UIText>First line</UIText>
+          <UIText>Second line</UIText>
+        </Column>
+      </Host>
+    </View>
+  );
+}
+
+export async function test(
+  { it, describe, expect, afterEach }: any,
```
Kudo: [🐧 nit] test-suite already exports the types for both parameters: `JasmineInterface` and `TestPortal` from `../types` (see `apps/test-suite/tests/Image.tsx:23`). 37 of the 38 test files use them, so `any` here loses `expect` typing for no reason.

Kudo (approved):

apps/test-suite/tests/ExpoUIHostSize.tsx
intergalacticspacehighway: This test was only happy path. I thought regression will be hard to repro (commit failure case). But there's a simple way to trigger it (thanks to claude) - https://github.com/expo/expo/pull/49211/changes/7e100179ad63ade5ccbd09fec7b33daeede531ac

So a child re-render that leaves the Host's own props untouched and doesn't change the content size will trigger the regression. Added the case for it, it fails before the fix.

PR #49218, opened 2026-08-21
[router] Move store values into context

Ubax:
# Why

Many of the imperative `store` usages can be moved into the react scope:
- `store.getRouteInfo` - in most places it can be replaced with `useRouteInfo`. Some imperative routing still needs them until we integrate global state
- `store.redirects`, `store.rootComponent`, `store.routeNode` - they are stable and used only in react scope
-  `store.linking` - this is the hardest to migrate. However it is stable apart for hot-module reloading, so it should not be a problem
- `store.assertIsReady`, `store.onReady`, `shouldShowTutorial` - they can be fully removed, as they are not tied to the store state

Having them in the react scope, makes them concurrent safe and easier to integrate with global state

# How

Use react hooks and context wherever possible.

# Test Plan

CI

# Checklist

### DIFF
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -66,7 +66,7 @@
 
 ### 💡 Others
 
-- Keep linking route information in React scope. ([#49176](https://github.com/expo/expo/pull/49176) by [@Ubax](https://github.com/Ubax))
+- Move Expo Router store values into React context. ([#49218](https://github.com/expo/expo/pull/49218) by [@Ubax](https://github.com/Ubax))
 - Remove the ignored `linking.enabled` option. ([#49103](https://github.com/expo/expo/pull/49103) by [@Ubax](https://github.com/Ubax))
 - Render only the focused tab route during the first render. ([#48618](https://github.com/expo/expo/pull/48618) by [@Ubax](https://github.com/Ubax))
 - Order JS Tabs, NativeTabs, headless tabs and Drawer by `routeNames` order ([#48374](https://github.com/expo/expo/pull/48374) by [@Ubax](https://github.com/Ubax))
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -50,7 +50,6 @@
 
 ### 🐛 Bug fixes
 
-- Keep linking route information in React scope. ([#49176](https://github.com/expo/expo/pull/49176) by [@Ubax](https://github.com/Ubax))
 - Make layouts with explicitly declared screens honor `unstable_settings.initialRouteName` instead of declaration order, which can change deep-link back stacks. ([#48708](https://github.com/expo/expo/pull/48708) by [@Ubax](https://github.com/Ubax))
 - Prevent unfocused nested native tab navigators from redirecting global router state. ([#48257](https://github.com/expo/expo/pull/48257) by [@Ubax](https://github.com/Ubax))
 - Fixed `Tabs` and `TopTabs` (`expo-router/js-tabs`, `expo-router/js-top-tabs`) not being usable from RSC ([#48330](https://github.com/expo/expo/pull/48330) by [@Ubax](https://github.com/Ubax))
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -50,6 +50,7 @@
 
 ### 🐛 Bug fixes
 
+- Keep linking route information in React scope. ([#49176](https://github.com/expo/expo/pull/49176) by [@Ubax](https://github.com/Ubax))
 - Make layouts with explicitly declared screens honor `unstable_settings.initialRouteName` instead of declaration order, which can change deep-link back stacks. ([#48708](https://github.com/expo/expo/pull/48708) by [@Ubax](https://github.com/Ubax))
 - Prevent unfocused nested native tab navigators from redirecting global router state. ([#48257](https://github.com/expo/expo/pull/48257) by [@Ubax](https://github.com/Ubax))
 - Fixed `Tabs` and `TopTabs` (`expo-router/js-tabs`, `expo-router/js-top-tabs`) not being usable from RSC ([#48330](https://github.com/expo/expo/pull/48330) by [@Ubax](https://github.com/Ubax))
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -50,7 +50,6 @@
 
 ### 🐛 Bug fixes
 
-- Keep linking route information in React scope. ([#49176](https://github.com/expo/expo/pull/49176) by [@Ubax](https://github.com/Ubax))
 - Make layouts with explicitly declared screens honor `unstable_settings.initialRouteName` instead of declaration order, which can change deep-link back stacks. ([#48708](https://github.com/expo/expo/pull/48708) by [@Ubax](https://github.com/Ubax))
 - Prevent unfocused nested native tab navigators from redirecting global router state. ([#48257](https://github.com/expo/expo/pull/48257) by [@Ubax](https://github.com/Ubax))
 - Fixed `Tabs` and `TopTabs` (`expo-router/js-tabs`, `expo-router/js-top-tabs`) not being usable from RSC ([#48330](https://github.com/expo/expo/pull/48330) by [@Ubax](https://github.com/Ubax))
@@ -67,6 +66,7 @@
 
 ### 💡 Others
 
+- Keep linking route information in React scope. ([#49176](https://github.com/expo/expo/pull/49176) by [@Ubax](https://github.com/Ubax))
 - Remove the ignored `linking.enabled` option. ([#49103](https://github.com/expo/expo/pull/49103) by [@Ubax](https://github.com/Ubax))
 - Render only the focused tab route during the first render. ([#48618](https://github.com/expo/expo/pull/48618) by [@Ubax](https://github.com/Ubax))
 - Order JS Tabs, NativeTabs, headless tabs and Drawer by `routeNames` order ([#48374](https://github.com/expo/expo/pull/48374) by [@Ubax](https://github.com/Ubax))
--- a/packages/expo-router/src/ExpoRoot.tsx
+++ b/packages/expo-router/src/ExpoRoot.tsx
@@ -8,12 +8,14 @@ import { INTERNAL_SLOT_NAME, NOT_FOUND_ROUTE_NAME, SITEMAP_ROUTE_NAME } from './
 import { useDomComponentNavigation } from './domComponents/useDomComponentNavigation';
 import { NavigationContainer as UpstreamNavigationContainer } from './fork/NavigationContainer';
 import type { ExpoLinkingOptions } from './getLinkingConfig';
-import { store, useStore } from './global-state/router-store';
+import { useStore } from './global-state/router-store';
 import { RouterRegistryProvider } from './global-state/routerRegistry';
 import type { ServerContextType } from './global-state/serverLocationContext';
 import { ServerContext } from './global-state/serverLocationContext';
+import { maybeHideSplashScreen, store } from './global-state/store';
 import { StoreContext } from './global-state/storeContext';
 import { shouldAppendNotFound, shouldAppendSitemap } from './global-state/utils';
+import { useImperativeApiEmitter } from './imperative-api';
 import { LinkPreviewContextProvider } from './link/preview/LinkPreviewContext';
 import { handleNavigationOnReady } from './navigationEvents/navigation';
 import { Screen } from './primitives';
@@ -90,12 +92,12 @@ const initialUrl =
     ? new URL(window.location.href)
     : undefined;
 
-// TODO(@ubax): Refactor onReady logic and use listeners pattern
 function onNavigationReady() {
   handleNavigationOnReady();
-  store.onReady();
+  maybeHideSplashScreen();
 }
 
+// TODO(@ubax): Refactor onReady logic and use listeners pattern
 function ContextNavigator({
   context,
   location: initialLocation = initialUrl,
@@ -134,11 +136,18 @@ function ContextNavigator({
     ? `${serverContext.location.pathname}${serverContext.location.search}${serverContext.location.hash ?? ''}`
     : undefined;
 
-  const store = useStore(context, linking, serverUrl);
-
+  const storeValue = useStore(context, linking, serverUrl);
+  const {
+    navigationRef,
+    initialState,
+    rootComponent,
+    linking: linkingConfig,
+    routeNode,
+  } = storeValue;
   useDomComponentNavigation();
 
-  if (store.shouldShowTutorial()) {
+  // TODO(@ubax): Revisit onboarding once route creation is React-owned.
+  if (process.env.NODE_ENV === 'development' && !routeNode) {
     SplashScreen.hideAsync();
     if (process.env.NODE_ENV === 'development') {
       const Tutorial = require('./onboard/Tutorial').Tutorial;
@@ -154,19 +163,20 @@ function ContextNavigator({
   }
 
   return (
-    <StoreContext.Provider value={store}>
+    <StoreContext.Provider value={storeValue}>
       <RouterRegistryProvider>
         <UpstreamNavigationContainer
-          ref={store.navigationRef}
-          initialState={store.state}
-          linking={store.linking as LinkingOptions<any>}
+          ref={navigationRef}
+          initialState={initialState}
+          linking={linkingConfig as LinkingOptions<any>}
           onUnhandledAction={onUnhandledAction}
           onStateChange={store.onStateChange}
           documentTitle={documentTitle}
           onReady={onNavigationReady}>
+          <ImperativeApiEmitter />
           <ServerContext.Provider value={serverContext}>
             <WrapperComponent>
-              <Content />
+              <Content rootComponent={rootComponent} />
             </WrapperComponent>
           </ServerContext.Provider>
         </UpstreamNavigationContainer>
@@ -175,10 +185,13 @@ function ContextNavigator({
   );
 }
 
-function Content() {
-  const children = [
-    <Screen key="SLOT" name={INTERNAL_SLOT_NAME} component={store.rootComponent} />,
-  ];
+function ImperativeApiEmitter() {
+  useImperativeApiEmitter();
+  return null;
+}
+
+function Content({ rootComponent }: { rootComponent: ComponentType<any> }) {
+  const children = [<Screen key="SLOT" name={INTERNAL_SLOT_NAME} component={rootComponent} />];
   if (shouldAppendNotFound()) {
     children.push(<Screen key="NOT-FOUND" name={NOT_FOUND_ROUTE_NAME} component={RootUnmatched} />);
   }
--- a/packages/expo-router/src/fork/useLinking.native.ts
+++ b/packages/expo-router/src/fork/useLinking.native.ts
@@ -2,6 +2,7 @@ import * as ExpoLinking from 'expo-linking';
 import { type RefObject, useEffect, useCallback, useRef } from 'react';
 import { Linking, Platform } from 'react-native';
 
+import { useRouteInfo } from '../global-state/useRouteInfo';
 import {
   type LinkingOptions,
   getActionFromState as getActionFromStateDefault,
@@ -10,10 +11,13 @@ import {
   type ParamListBase,
 } from '../react-navigation/native';
 import { extractExpoPathFromURL } from './extractPathFromURL';
+import { getStateFromPath as getExpoStateFromPath } from './getStateFromPath';
 
 type ResultState = ReturnType<typeof getStateFromPathDefault>;
 
-type Options = LinkingOptions<ParamListBase>;
+type Options = Omit<LinkingOptions<ParamListBase>, 'getStateFromPath'> & {
+  getStateFromPath?: typeof getExpoStateFromPath;
+};
 
 const linkingHandlers: symbol[] = [];
 
@@ -49,8 +53,11 @@ export function useLinking(
         }
       };
     });
-  const getStateFromPath = options?.getStateFromPath ?? getStateFromPathDefault;
+  const getStateFromPath = (options?.getStateFromPath ??
+    getStateFromPathDefault) as typeof getExpoStateFromPath;
   const getActionFromState = options?.getActionFromState ?? getActionFromStateDefault;
+  const independent = useNavigationIndependentTree();
+  const { segments } = useRouteInfo();
   useEffect(() => {
     if (process.env.NODE_ENV === 'production') {
       return undefined;
@@ -93,6 +100,10 @@ export function useLinking(
   const getInitialURLRef = useRef(getInitialURL);
   const getStateFromPathRef = useRef(getStateFromPath);
   const getActionFromStateRef = useRef(getActionFromState);
+  const segmentsRef = useRef(segments);
+
+  // Keep stable URL listeners in sync with the latest navigation state.
+  segmentsRef.current = segments;
 
   useEffect(() => {
     enabledRef.current = enabled;
@@ -112,7 +123,9 @@ export function useLinking(
 
       const path = extractExpoPathFromURL(prefixesRef.current, url);
 
-      return path !== undefined ? getStateFromPathRef.current(path, configRef.current) : undefined;
+      return path !== undefined
+        ? getStateFromPathRef.current(path, configRef.current, segmentsRef.current)
+        : undefined;
     },
 
     []
--- a/packages/expo-router/src/fork/useLinking.native.ts
+++ b/packages/expo-router/src/fork/useLinking.native.ts
@@ -53,7 +53,6 @@ export function useLinking(
     });
   const getStateFromPath = options?.getStateFromPath ?? getExpoStateFromPath;
   const getActionFromState = options?.getActionFromState ?? getActionFromStateDefault;
-  const independent = useNavigationIndependentTree();
   const { segments } = useRouteInfo();
   useEffect(() => {
     if (process.env.NODE_ENV === 'production') {
--- a/packages/expo-router/src/fork/useLinking.native.ts
+++ b/packages/expo-router/src/fork/useLinking.native.ts
@@ -15,9 +15,7 @@ import { getStateFromPath as getExpoStateFromPath } from './getStateFromPath';
 
 type ResultState = ReturnType<typeof getStateFromPathDefault>;
 
-type Options = Omit<LinkingOptions<ParamListBase>, 'getStateFromPath'> & {
-  getStateFromPath?: typeof getExpoStateFromPath;
-};
+type Options = LinkingOptions<ParamListBase>;
 
 const linkingHandlers: symbol[] = [];
 
@@ -53,8 +51,7 @@ export function useLinking(
         }
       };
     });
-  const getStateFromPath: typeof getExpoStateFromPath =
-    options?.getStateFromPath ?? ((path, options) => getStateFromPathDefault(path, options));
+  const getStateFromPath = options?.getStateFromPath ?? getExpoStateFromPath;
   const getActionFromState = options?.getActionFromState ?? getActionFromStateDefault;
   const independent = useNavigationIndependentTree();
   const { segments } = useRouteInfo();
--- a/packages/expo-router/src/fork/useLinking.native.ts
+++ b/packages/expo-router/src/fork/useLinking.native.ts
@@ -53,8 +53,8 @@ export function useLinking(
         }
       };
     });
-  const getStateFromPath: typeof getExpoStateFromPath =
-    options?.getStateFromPath ?? ((path, options) => getStateFromPathDefault(path, options));
+  const getStateFromPath = (options?.getStateFromPath ??
+    getStateFromPathDefault) as typeof getExpoStateFromPath;
   const getActionFromState = options?.getActionFromState ?? getActionFromStateDefault;
   const independent = useNavigationIndependentTree();
   const { segments } = useRouteInfo();
@@ -102,8 +102,7 @@ export function useLinking(
   const getActionFromStateRef = useRef(getActionFromState);
   const segmentsRef = useRef(segments);
 
-  // Linking listeners stay subscribed across navigation updates, but relative URLs must be parsed
-  // against the route that was current when the URL was received.
+  // Keep stable URL listeners in sync with the latest navigation state.
   segmentsRef.current = segments;
 
   useEffect(() => {
--- a/packages/expo-router/src/fork/useLinking.native.ts
+++ b/packages/expo-router/src/fork/useLinking.native.ts
@@ -1,6 +1,5 @@
-import * as ExpoLinking from 'expo-linking';
-import { type RefObject, useEffect, useCallback, useRef } from 'react';
-import { Linking, Platform } from 'react-native';
+import { type RefObject, useEffect, useEffectEvent, useCallback, useRef } from 'react';
+import { Linking } from 'react-native';
 
 import { useRouteInfo } from '../global-state/useRouteInfo';
 import {
@@ -11,6 +10,7 @@ import {
   type ParamListBase,
 } from '../react-navigation/native';
 import { extractExpoPathFromURL } from './extractPathFromURL';
+import { getInitialURLWithTimeout } from './getInitialURLWithTimeout';
 import { getStateFromPath as getExpoStateFromPath } from './getStateFromPath';
 
 type ResultState = ReturnType<typeof getStateFromPathDefault>;
@@ -100,11 +100,6 @@ export function useLinking(
   const getInitialURLRef = useRef(getInitialURL);
   const getStateFromPathRef = useRef(getStateFromPath);
   const getActionFromStateRef = useRef(getActionFromState);
-  const segmentsRef = useRef(segments);
-
-  // Linking listeners stay subscribed across navigation updates, but relative URLs must be parsed
-  // against the route that was current when the URL was received.
-  segmentsRef.current = segments;
 
   useEffect(() => {
     enabledRef.current = enabled;
@@ -125,12 +120,13 @@ export function useLinking(
       const path = extractExpoPathFromURL(prefixesRef.current, url);
 
       return path !== undefined
-        ? getStateFromPathRef.current(path, configRef.current, segmentsRef.current)
+        ? getStateFromPathRef.current(path, configRef.current, segments)
         : undefined;
     },
 
-    []
+    [segments]
   );
+  const getStateFromURLInEffect = useEffectEvent(getStateFromURL);
 
   const getInitialState = useCallback(() => {
     let state: ResultState | undefined;
@@ -177,7 +173,7 @@ export function useLinking(
       }
 
       const navigation = ref.current;
-      const state = navigation ? getStateFromURL(url) : undefined;
+      const state = navigation ? getStateFromURLInEffect(url) : undefined;
 
       if (navigation && state) {
         // If the link were handled, it gets cleared in NavigationContainer
@@ -208,28 +204,9 @@ export function useLinking(
     };
 
     return subscribe(listener);
-  }, [enabled, getStateFromURL, onUnhandledLinking, prefixes, ref, subscribe]);
+  }, [enabled, onUnhandledLinking, prefixes, ref, subscribe]);
 
   return {
     getInitialState,
   };
-}
-
-export function getInitialURLWithTimeout(): string | null | Promise<string | null> {
-  if (typeof window === 'undefined') {
-    return '';
-  } else if (Platform.OS === 'ios') {
-    // Use the new Expo API for iOS. This has better support for App Clips and handoff.
-    return ExpoLinking.getLinkingURL();
-  }
-
-  return Promise.race([
-    // TODO: Phase this out in favor of expo-linking on Android.
-    Linking.getInitialURL(),
-    new Promise<null>((resolve) =>
-      // Timeout in 150ms if `getInitialState` doesn't resolve
-      // Workaround for https://github.com/facebook/react-native/issues/25675
-      setTimeout(() => resolve(null), 150)
-    ),
-  ]);
 }
--- a/packages/expo-router/src/fork/useLinking.native.ts
+++ b/packages/expo-router/src/fork/useLinking.native.ts
@@ -53,8 +53,8 @@ export function useLinking(
         }
       };
     });
-  const getStateFromPath = (options?.getStateFromPath ??
-    getStateFromPathDefault) as typeof getExpoStateFromPath;
+  const getStateFromPath: typeof getExpoStateFromPath =
+    options?.getStateFromPath ?? ((path, options) => getStateFromPathDefault(path, options));
   const getActionFromState = options?.getActionFromState ?? getActionFromStateDefault;
   const independent = useNavigationIndependentTree();
   const { segments } = useRouteInfo();
@@ -102,7 +102,8 @@ export function useLinking(
   const getActionFromStateRef = useRef(getActionFromState);
   const segmentsRef = useRef(segments);
 
-  // Keep stable URL listeners in sync with the latest navigation state.
+  // Linking listeners stay subscribed across navigation updates, but relative URLs must be parsed
+  // against the route that was current when the URL was received.
   segmentsRef.current = segments;
 
   useEffect(() => {
--- a/packages/expo-router/src/fork/useLinking.ts
+++ b/packages/expo-router/src/fork/useLinking.ts
@@ -1,5 +1,13 @@
 import isEqual from 'fast-deep-equal';
-import { type RefObject, useEffect, useState, useCallback, useRef, use } from 'react';
+import {
+  type RefObject,
+  useEffect,
+  useEffectEvent,
+  useState,
+  useCallback,
+  useRef,
+  use,
+} from 'react';
 
 import { ServerContext } from '../global-state/serverLocationContext';
 import { useExpoRouterStore } from '../global-state/storeContext';
@@ -135,11 +143,6 @@ export function useLinking(
   const getStateFromPathRef = useRef(getStateFromPath);
   const getPathFromStateRef = useRef(getPathFromState);
   const getActionFromStateRef = useRef(getActionFromState);
-  const segmentsRef = useRef(segments);
-
-  // Segments are navigation state, so keep this ref current during render. The
-  // linking listeners intentionally remain stable across navigation updates.
-  segmentsRef.current = segments;
 
   useEffect(() => {
     enabledRef.current = enabled;
@@ -148,6 +151,12 @@ export function useLinking(
     getPathFromStateRef.current = getPathFromState;
     getActionFromStateRef.current = getActionFromState;
   });
+
+  const getStateFromPathForCurrentSegments = useCallback(
+    (path: string) => getStateFromPathRef.current(path, configRef.current, segments),
+    [segments]
+  );
+  const getStateFromPathInEffect = useEffectEvent(getStateFromPathForCurrentSegments);
 
   const validateRoutesNotExistInRootState = useCallback(
     (state: ResultState) => {
@@ -182,7 +191,7 @@ export function useLinking(
         : undefined;
 
       if (path) {
-        value = getStateFromPathRef.current(path, configRef.current, segmentsRef.current);
+        value = getStateFromPathForCurrentSegments(path);
       }
 
       // If the link were handled, it gets cleared in NavigationContainer
@@ -200,7 +209,7 @@ export function useLinking(
 
     return thenable as PromiseLike<ResultState | undefined>;
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+  }, [getStateFromPathForCurrentSegments, onUnhandledLinking, server]);
 
   const previousIndexRef = useRef<number | undefined>(undefined);
   const previousStateRef = useRef<NavigationState | undefined>(undefined);
@@ -236,7 +245,7 @@ export function useLinking(
         return;
       }
 
-      const state = getStateFromPathRef.current(path, configRef.current, segmentsRef.current);
+      const state = getStateFromPathInEffect(path);
 
       // We should only dispatch an action when going forward
       // Otherwise the action will likely add items to history, which would mess things up
@@ -314,11 +323,7 @@ export function useLinking(
       // If the `route` object contains a `path`, use that path as long as `route.name` and `params` still match
       // This makes sure that we preserve the original URL for wildcard routes
       if (route?.path) {
-        const stateForPath = getStateFromPathRef.current(
-          route.path,
-          configRef.current,
-          segmentsRef.current
-        );
+        const stateForPath = getStateFromPathInEffect(route.path);
 
         if (stateForPath) {
           const focusedRoute = findFocusedRoute(stateForPath);
@@ -485,8 +490,4 @@ export function useLinking(
   return {
     getInitialState,
   };
-}
-
-export function getInitialURLWithTimeout(): string | null | Promise<string | null> {
-  return typeof window === 'undefined' ? '' : window.location.href;
 }
--- a/packages/expo-router/src/fork/useLinking.ts
+++ b/packages/expo-router/src/fork/useLinking.ts
@@ -86,8 +86,8 @@ export function useLinking(
 ) {
   const enabled = options !== undefined;
   const config = options?.config;
-  const getStateFromPath = (options?.getStateFromPath ??
-    getStateFromPathDefault) as typeof getExpoStateFromPath;
+  const getStateFromPath: typeof getExpoStateFromPath =
+    options?.getStateFromPath ?? ((path, options) => getStateFromPathDefault(path, options));
   const getPathFromState = options?.getPathFromState ?? getPathFromStateDefault;
   const getActionFromState = options?.getActionFromState ?? getActionFromStateDefault;
   const store = useExpoRouterStore();
--- a/packages/expo-router/src/fork/useLinking.ts
+++ b/packages/expo-router/src/fork/useLinking.ts
@@ -86,8 +86,8 @@ export function useLinking(
 ) {
   const enabled = options !== undefined;
   const config = options?.config;
-  const getStateFromPath: typeof getExpoStateFromPath =
-    options?.getStateFromPath ?? ((path, options) => getStateFromPathDefault(path, options));
+  const getStateFromPath = (options?.getStateFromPath ??
+    getStateFromPathDefault) as typeof getExpoStateFromPath;
   const getPathFromState = options?.getPathFromState ?? getPathFromStateDefault;
   const getActionFromState = options?.getActionFromState ?? getActionFromStateDefault;
   const store = useExpoRouterStore();
--- a/packages/expo-router/src/fork/useLinking.ts
+++ b/packages/expo-router/src/fork/useLinking.ts
@@ -83,9 +83,7 @@ export const series = (cb: () => Promise<void>) => {
 
 const linkingHandlers: symbol[] = [];
 
-type Options = Omit<LinkingOptions<ParamListBase>, 'getStateFromPath'> & {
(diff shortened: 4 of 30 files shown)


packages/expo-router/src/link/preview/HrefPreview.tsx
```diff
@@ -25,9 +26,10 @@ import { PreviewRouteContext } from './PreviewRouteContext';
 
 export function HrefPreview({ href }: { href: Href }) {
   const routeInfo = useRouteInfo();
+  const { linking } = useExpoRouterStore();
```
Ubax: Add `TODO(@ubax): check if using more granular context would improve performance here`

packages/expo-router/src/global-state/useStore.ts
Ubax: Can/should we `useMemo` here to prevent reference changes? 

This seems to depend only on `context` which should be relatively stable, `linkingConfigOptions` which should be stable and serverUrl which is stable. Can you check if there are any problems with referential instability?

Ubax, line 1: Move most of the hook into useMemo and not only the. result. So that linking and others are not computed multiple times unnecessairly

packages/expo-router/src/global-state/useStore.ts
```diff
@@ -118,5 +119,15 @@ export function useStore(
     };
   });
 
-  return store;
+  return useMemo(
+    () => ({
+      ...store,
```
jakub-agent: [🚨 **critical**] Spreading `store` breaks the live reads it was built for.

`store` (in `./store`) is an object literal of **getters** — `get state()`, `get routeNode()`, `get navigationRef()`, `get linking()`, `get redirects()`, `get rootComponent()` — each forwarding to `storeRef.current`. Object spread invokes getters and copies the *result*, so every one of them becomes a plain value property on the context object. Methods (`onStateChange`, `setFocusedState`, `assertIsReady`, ...) stay live; getters do not.

The explicit fields below re-supply `navigationRef`, `linking`, `initialState`, `rootComponent` and `redirects`, but `state` and `routeNode` are not overridden, so they stay frozen at whatever `storeRef.current` held when this memo ran.

That is a real regression on web. `src/fork/useLinking.ts` does `const store = useExpoRouterStore()` and then reads `store.state` at lines 383 and 414 to build the history path. Those reads were live before this PR (the provider passed the module `store` object) and are now pinned to the initial state, so `history.push`/`history.replace` keep computing the initial path and the URL stops tracking navigation. `ContextNavigator` does not subscribe to route info, so it never re-renders to refresh the snapshot.

Suggested fix: don't spread. Either keep the store reachable as its own field (`{ store, navigationRef, linking, ... }`), or build the object with explicit method references plus real getters for the live values, so `state`/`routeNode` cannot be captured by accident.

packages/expo-router/src/global-state/storeContext.ts
```diff
@@ -1,8 +1,21 @@
...
 import { createContext, use } from 'react';
+import type { ComponentType } from 'react';
 
+import type { ExpoLinkingOptions } from '../getLinkingConfig';
+import type { NavigationContainerRefWithCurrent } from '../react-navigation/native';
 import type { RouterStore } from './store';
+import type { ReactNavigationState, StoreRedirects } from './types';
 
-export const StoreContext = createContext<RouterStore | null>(null);
+export type StoreContextValue = RouterStore & {
```
jakub-agent: [🏗️ **design**] `RouterStore &` is what makes the spread above type-check. It keeps `state` and `routeNode` on the public context type as ordinary properties, so consumers are invited to read values that the spread has already frozen, and TypeScript cannot tell the difference.

It also puts `initialState` and the inherited `state` side by side on one type with no hint that only one of them is meaningful here.

Consider narrowing to the methods the context actually needs to expose — e.g. `Omit<RouterStore, 'state' | 'routeNode' | 'navigationRef' | 'linking' | 'redirects' | 'rootComponent'> & { ... }` — so a stale read becomes a type error rather than a silent bug.

packages/expo-router/src/global-state/useStore.ts
```diff
@@ -118,5 +119,15 @@ export function useStore(
...
+      navigationRef,
+      linking,
+      initialState,
+      rootComponent,
+      redirects,
+    }),
+    [navigationRef, linking, initialState, rootComponent, redirects]
```
jakub-agent: [💡 suggestion] This `useMemo` never hits. Of the five dependencies, three are rebuilt on every render of this hook:

- `linking` comes from `getLinkingConfig(...)`, which returns a fresh object each call
- `redirects` is a `.map()` over the config, so a new array each call
- `initialState` comes from `linking.getStateFromPath(...)`, also fresh each call

Only `navigationRef` is stable, and `rootComponent` is stable only while `routeNode` is (it is cached by `WeakMap` on the node). So the memo recomputes every time and the context value changes identity, re-rendering every consumer.

Today `ContextNavigator` renders roughly once, so the cost is hidden — but the `useMemo` reads as a stability guarantee it does not provide. Either memoize the upstream computations (`routeNode`, `linking`, `redirects`) so the deps are actually stable, or drop the `useMemo` and be explicit that the value is per-render.

packages/expo-router/src/imperative-api.tsx
```diff
@@ -13,13 +14,14 @@ export function useImperativeApiEmitter(
...
+  const store = useOptionalExpoRouterStore();
   const events = useSyncExternalStore(
     routingQueue.subscribe,
     routingQueue.snapshot,
     routingQueue.snapshot
   );
   useEffect(() => {
-    routingQueue.run(ref, routeInfo);
-  }, [events, ref, routeInfo]);
+    routingQueue.run(ref, routeInfo, store?.linking, store?.redirects);
```
jakub-agent: [💡 suggestion] `store?.redirects` degrades silently. When the context is absent, `redirects` falls back to `[]` in `routingQueue.run`, so configured `redirects`/`rewrites` simply stop being applied — no warning, no error — while `linking` being `undefined` throws `'Attempted to link to route when no routes are present'` from `getNavigateAction`. Two different failure modes for the same missing context.

`useImperativeApiEmitter` is only reached through `fork/NavigationContainer`, which `ExpoRoot` renders inside `StoreContext.Provider`, so in the app the context is always there. The only null case is `fork/NavigationContainer` rendered standalone in `src/react-navigation/**/__tests__`.

If that is the case being handled, a comment saying so would help. Otherwise `useExpoRouterStore()` and non-optional arguments would keep a misconfiguration loud instead of turning it into missing redirects.

packages/expo-router/src/global-state/getNavigationAction.ts
```diff
@@ -19,7 +20,9 @@ export function getNavigateAction(
   withAnchor?: boolean,
   singular?: SingularOptions,
   isPreviewNavigation?: boolean,
-  routeInfo: UrlObject = defaultRouteInfo
+  routeInfo: UrlObject = defaultRouteInfo,
+  linking?: ExpoLinkingOptions,
```
jakub-agent: [💡 suggestion] Making `linking` optional (and `redirects` default to `[]`) means a caller that forgets these arguments gets a runtime throw or quietly skipped redirects rather than a compile error. There is exactly one call site (`routingQueue.run`), which already always passes both.

Since they are now the only source of this data — the function no longer falls back to `store.linking` / `store.redirects` — required parameters would push the mistake to type-check time.

Separately, this function still reads `store.assertIsReady()` and `store.navigationRef` from the module singleton a few lines down, so it is half-migrated. Fine as an incremental step, but worth a TODO so the split is intentional rather than looking forgotten.

packages/expo-router/src/ExpoRoot.tsx
```diff
@@ -154,19 +155,19 @@ function ContextNavigator({
...
-          linking={store.linking as LinkingOptions<any>}
+          ref={navigationRef}
+          initialState={initialState}
+          linking={linkingConfig as LinkingOptions<any>}
           onUnhandledAction={onUnhandledAction}
-          onStateChange={store.onStateChange}
+          onStateChange={storeValue.onStateChange}
           documentTitle={documentTitle}
-          onReady={onNavigationReady}>
+          onReady={() => onNavigationReady(storeValue)}>
```
jakub-agent: [🐧 nit] This threading does not buy anything. `onNavigationReady` only calls `store.onReady()`, and `onReady` is a plain method on the module `store` — the spread in `useStore` copies the same function reference, so `storeValue.onReady === store.onReady`. Meanwhile the inline arrow replaces a stable module-level callback with a new closure on every render.

Keeping `onReady={onNavigationReady}` and importing `store` inside `onNavigationReady` is equivalent and stable. If the intent is to drop the singleton import from this file, `onReady` is not the piece of state that needs it.

Minor: the parameter is typed `ReturnType<typeof useStore>`; `StoreContextValue` is the name for that.

packages/expo-router/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unpublished
 
+### 🐛 Bug fixes
+
+- Move Expo Router store values into React context. ([#49176](https://github.com/expo/expo/pull/49176) by [@jakub-agent](https://github.com/jakub-agent))
```
jakub-agent: [💡 suggestion] Three things here:

1. The link points at #49176, the PR below this one in the stack. #49176 already has its own entry further down (`Keep linking route information in React scope`). This entry should link to #49218.
2. It adds a second `### 🐛 Bug fixes` heading inside `## Unpublished` — there is already one at line 52 — and places it above `### 🛠 Breaking changes`, which inverts the section order used everywhere else in this file.
3. Moving store values into context is an internal refactor with no user-visible behaviour change, so `### 💡 Others` fits better than `### 🐛 Bug fixes` — which is also where #49176's entry landed.

jakub-agent (changes requested):

PR 2/2 in the stack on top of #49176. The direction is right: passing `navigationRef`, `linking`, `initialState`, `rootComponent` and `redirects` through `StoreContext` instead of reading the module singleton makes the render path React-correct and testable.

One blocking problem. `useStore` now returns `{ ...store, ... }`. `store` is an object literal built from **getters** (`get state()`, `get routeNode()`, ...), and object spread evaluates getters eagerly, so the context value holds frozen snapshots instead of live reads. Verified in isolation:

```
const ref = { current: { state: 'S1' } };
const store = { get state() { return ref.current.state } };
const spread = { ...store };
ref.current.state = 'S2';
store.state   // 'S2'
spread.state  // 'S1'   <- descriptor is now a plain value
```

`state` and `routeNode` are not overridden by the spread, so they keep the value they had when the memo ran. `src/fork/useLinking.ts` (lines 383 and 414, not touched by this PR) reads `store.state` off exactly this context value to compute the browser path, so on web it now always sees the initial state. Before this PR the provider passed the live `store` object, so those reads were correct.

Worth noting: `src/fork/__tests__/useLinking.test.web.tsx` mocks the context with `get state()` — a getter — which is precisely the semantics the real provider just lost. That mock keeps the suite green and hides the regression. There is no test that renders the real provider together with the real `useLinking`, so this path is uncovered.

I could not run `pnpm test` / `et check-packages expo-router`: the session worktree has no installed `node_modules`. The spread behaviour above was verified directly; the `useLinking` impact is from reading the call sites.

Rest of the findings are non-blocking: a `useMemo` whose dependencies are all rebuilt every render, optional parameters that turn a missing argument into a runtime throw or a silently empty redirect list, and a changelog entry that points at the wrong PR and adds a duplicate section.

packages/expo-router/src/imperative-api.tsx
Ubax: `useEffectEvent` here

Ubax: Add `if` with throw for `if(!store)`

packages/expo-router/CHANGELOG.md
Ubax: Add this in `## Others` just above `#49176`

Make `@ubax` and author

packages/expo-router/src/ExpoRoot.tsx
```diff
@@ -91,7 +91,7 @@ const initialUrl =
     : undefined;
 
 // TODO(@ubax): Refactor onReady logic and use listeners pattern
-function onNavigationReady() {
+function onNavigationReady(store: ReturnType<typeof useStore>) {
```
Ubax:
```suggestion
function onNavigationReady() {
```

Move the function to `ExpoRoot`, use `useCallback`

packages/expo-router/src/global-state/getNavigationAction.ts
Ubax: Make them required

packages/expo-router/src/global-state/getNavigationAction.ts
```diff
@@ -19,7 +20,9 @@ export function getNavigateAction(
...
+  redirects: StoreRedirects[] = []
 ) {
   let href: string | undefined = baseHref;
   store.assertIsReady();
```
Ubax: Remove this line. Add required `navigationRef` argument to the function. Then call 

```tsx
    if (!navigationRef.isReady()) {
      throw new Error(
        'Attempted to navigate before mounting the Root Layout component. Ensure the Root Layout component is rendering a Slot, or other navigator on the first render.'
      );
    }
```

Also remove the next line as well

packages/expo-router/src/global-state/useStore.ts
Ubax: Move most of the hook into useMemo and not only the. result. So that linking and others are not computed multiple times unnecessairly

packages/expo-router/src/global-state/useStore.ts
Ubax: 1. Remove `rootComponent` and `redirects` from imperative `store` - only keep it as values in the returned value
2. Don't pass `routeInfo` to the context and don't return it from this hook
3. Don't spread but only pass the relevant values by key

packages/expo-router/src/global-state/store.ts
Ubax: Extract this to standalone function in this file - `maybeHideSplashScreen`.

Then call this function in `ExpoRoot` instead of `store.onReady`. It can be removed

packages/expo-router/src/ExpoRoot.tsx
```diff
@@ -134,11 +129,16 @@ function ContextNavigator({
...
+  const { navigationRef, initialState, rootComponent, linking: linkingConfig } = storeValue;
+  const onNavigationReady = useCallback(() => {
+    handleNavigationOnReady();
+    storeValue.onReady();
+  }, [storeValue]);
 
   useDomComponentNavigation();
 
-  if (store.shouldShowTutorial()) {
+  if (storeValue.shouldShowTutorial) {
```
Ubax: Replace this with the actual logic:
```
process.env.NODE_ENV === 'development' && !routeNode
```

Also leave `TODO` comment to revisit this later

packages/expo-router/src/ExpoRoot.tsx
```diff
@@ -154,19 +154,19 @@ function ContextNavigator({
...
         <UpstreamNavigationContainer
-          ref={store.navigationRef}
-          initialState={store.state}
-          linking={store.linking as LinkingOptions<any>}
+          ref={navigationRef}
+          initialState={initialState}
+          linking={linkingConfig as LinkingOptions<any>}
           onUnhandledAction={onUnhandledAction}
-          onStateChange={store.onStateChange}
+          onStateChange={storeValue.onStateChange}
```
Ubax: Keep the imperative `store.onStateChange` here

packages/expo-router/src/global-state/storeContext.ts
```diff
@@ -1,8 +1,23 @@
...
-export const StoreContext = createContext<RouterStore | null>(null);
+export type StoreContextValue = {
+  navigationRef: NavigationContainerRefWithCurrent<ReactNavigation.RootParamList>;
+  linking: ExpoLinkingOptions | undefined;
+  initialState: ReactNavigationState | undefined;
+  rootComponent: ComponentType<any>;
+  redirects: StoreRedirects[];
+  shouldShowTutorial: boolean;
+  onReady: () => void;
+  onStateChange: (state: ReactNavigationState | undefined) => void;
```
Ubax: These are not needed in the context

packages/expo-router/src/global-state/storeContext.ts
```diff
@@ -1,8 +1,23 @@
...
+};
+
+export const StoreContext = createContext<StoreContextValue | null>(null);
 
 export const useExpoRouterStore = () => use(StoreContext)!;
+export const useOptionalExpoRouterStore = () => use(StoreContext);
```
Ubax: Remove. Use `use(StoreContext)` instead

packages/expo-router/src/global-state/store.ts
Ubax, line 57: Remove this and replace all the usages with context read

Ubax: Remove this and replace all its usages with `.navigationRef.isReady()` and throw

packages/expo-router/src/global-state/getNavigationAction.ts
```diff
@@ -1,50 +1,51 @@
...
+
+export type NavigationActionContext = {
+  routeInfo: UrlObject;
+  navigationRef: NavigationContainerRef<ParamListBase>;
+  linking: ExpoLinkingOptions;
+  redirects: StoreRedirects[];
+};
 
 export function getNavigateAction(
+  { routeInfo, navigationRef, linking, redirects }: NavigationActionContext,
```
Ubax:
```suggestion
```

packages/expo-router/src/global-state/getNavigationAction.ts
```diff
@@ -1,50 +1,51 @@
...
   baseHref: string,
   options: LinkToOptions,
-  type: string,
-  withAnchor: boolean | undefined,
-  singular: SingularOptions | undefined,
-  isPreviewNavigation: boolean | undefined,
-  { segments, params: routeParams }: Pick<UrlObject, 'segments' | 'params'>
+  type = 'NAVIGATE',
```
Ubax, line 26:
```suggestion
  { segments, params: routeParams }: Pick<UrlObject, 'segments' | 'params'>,
  { navigationRef, linking, redirects } :NavigationActionContext,
  type = 'NAVIGATE',
```

packages/expo-router/src/global-state/getNavigationAction.ts
```diff
@@ -1,50 +1,51 @@
...
     );
   }
-  if (!store.linking) {
-    throw new Error('Attempted to link to route when no routes are present');
-  }
   const rootState = navigationRef.getRootState();
 
-  href = resolveHrefStringWithSegments(href, { segments, params: routeParams }, options);
-  href = applyRedirects(href, store.redirects) ?? undefined;
+  href = resolveHrefStringWithSegments(href, routeInfo, options);
```
Ubax:
```suggestion
  href = resolveHrefStringWithSegments(href, { segments, params: routeParams }, options);
```

packages/expo-router/src/global-state/getNavigationAction.ts
```diff
@@ -1,50 +1,51 @@
...
+  href = applyRedirects(href, redirects) ?? undefined;
 
   // If the href is undefined, it means that the redirect has already been handled by the navigation
   if (!href) {
     return;
   }
 
-  const state = getStateFromPath(href, store.linking.config, segments);
+  const state = linking.getStateFromPath!(href, linking.config, routeInfo.segments);
```
Ubax:
```suggestion
  const state = linking.getStateFromPath!(href, linking.config, segments);
```

packages/expo-router/src/global-state/getNavigationAction.ts
```diff
@@ -1,50 +1,51 @@
...
-import { store } from './store';
-import type { LinkToOptions } from './types';
+import type { LinkToOptions, StoreRedirects } from './types';
+
+export type NavigationActionContext = {
+  routeInfo: UrlObject;
+  navigationRef: NavigationContainerRef<ParamListBase>;
+  linking: ExpoLinkingOptions;
+  redirects: StoreRedirects[];
+};
```
Ubax:
```suggestion
export type NavigationActionContext = Pick<StoreContextValue, 'navigationRef' | 'linking' | 'redirects'>
```

packages/expo-router/src/global-state/routingQueue.ts
```diff
@@ -37,7 +37,7 @@ export const routingQueue = {
   },
   run(
     ref: RefObject<NavigationContainerRef<ParamListBase> | null>,
-    routeInfo: Pick<UrlObject, 'segments' | 'params'>
+    context?: NavigationActionContext | UrlObject
```
Ubax:
```suggestion
    routeInfo: Pick<UrlObject, 'segments' | 'params'>,
    context: NavigationActionContext
```

packages/expo-router/src/global-state/routingQueue.ts
```diff
@@ -51,14 +51,17 @@ export const routingQueue = {
             payload: { href, options },
           } = action as LinkAction;
 
+          if (!context || !('navigationRef' in context)) {
+            throw new Error('Routing actions require Expo Router context.');
+          }
```
Ubax:
```suggestion
```

packages/expo-router/src/imperative-api.tsx
```diff
@@ -13,15 +14,29 @@ export function useImperativeApiEmitter(
   ref: RefObject<NavigationContainerRef<ParamListBase> | null>
 ) {
   const routeInfo = useRouteInfo();
+  const store = useOptionalExpoRouterStore();
```
Ubax:
```suggestion
  const store = use(StoreContext);
```

packages/expo-router/src/link/preview/useNextScreenId.ts
```diff
@@ -14,6 +15,7 @@ export function useNextScreenId(): [
 ] {
   const router = useRouter();
   const routeInfo = useRouteInfo();
+  const { navigationRef } = useExpoRouterStore();
```
Ubax:
```suggestion
  const navigationRef = useNavigationContainerRef();
```

packages/expo-router/src/ExpoRoot.tsx
```diff
@@ -134,11 +130,23 @@ function ContextNavigator({
...
+    navigationRef,
+    initialState,
+    rootComponent,
+    linking: linkingConfig,
+    routeNode,
+  } = storeValue;
+  const onNavigationReady = useCallback(() => {
+    handleNavigationOnReady();
+    maybeHideSplashScreen();
+  }, [storeValue]);
```
Ubax: Move this back out of component scope. It doesn't need anything from react anymore

packages/expo-router/src/fork/useLinking.ts
```diff
@@ -98,7 +98,6 @@ export function useLinking(
   const getActionFromState = options?.getActionFromState ?? getActionFromStateDefault;
   const independent = useNavigationIndependentTree();
 
-  const store = useExpoRouterStore();
```
Ubax, line 88: Why did you remove this?

If anything `const store = use(StoreContext)`

packages/expo-router/src/link/preview/useNextScreenId.ts
```diff
@@ -14,6 +16,8 @@ export function useNextScreenId(): [
 ] {
   const router = useRouter();
   const routeInfo = useRouteInfo();
+  const store = use(StoreContext);
+  const navigationRef = useNavigationContainerRef();
```
jakub-agent: [🚨 **critical**] This is the wrong `useNavigationContainerRef`, and the result is a listener that never fires.

The import on line 7 is from `../../react-navigation/native`, whose barrel does `export * from '../core'` (`src/react-navigation/native/index.tsx:43`). `core` re-exports the vendored react-navigation hook, which is explicitly marked `@deprecated Import 'useNavigationContainerRef' from 'expo-router' instead` (`src/react-navigation/core/index.tsx:83`). Its implementation is:

```ts
const navigation = React.useRef(null);
if (navigation.current == null) {
  navigation.current = createNavigationContainerRef();
}
return navigation.current;
```

So every call site gets its **own** container ref. This one is never passed to any `NavigationContainer`, so its `current` stays `null` forever.

That matters because of how `createNavigationContainerRef` handles a detached ref: `addListener` with `current == null` pushes the callback into a local `listeners` map and returns an unsubscribe, and those queued callbacks are only forwarded in the `set current(value)` setter. Since `current` is never assigned here, the `'state'` listener registered on line 61 is queued and never attached — no error, no warning, it simply never runs.

Downstream, `onNavigationStateChange` never executes, so `internalSetNextScreenId` and `setTabPath` never fire and `LinkWithPreview.tsx:46` sees `nextScreenId: undefined` / `tabPath: []` for the lifetime of the app. iOS link-preview prefetch stops resolving the previewed screen.

Before this PR the hook read `store.navigationRef`, which is the ref `useStore` actually hands to `UpstreamNavigationContainer`.

Fix: use the ref this PR already puts on the context — `store?.navigationRef` — or import from `../../hooks/useNavigationContainerRef`, which returns `store.navigationRef`. Given line 19 already reads the context, `store?.navigationRef` is the smaller change.

packages/expo-router/src/link/preview/utils.ts
```diff
@@ -10,16 +11,19 @@ import type {
...
 export function getTabPathFromRootStateByHref(
   href: Href,
   rootState: ReactNavigationState,
-  routeInfo: Pick<UrlObject, 'segments'>
+  routeInfo: Pick<UrlObject, 'segments'>,
+  linking: ExpoLinkingOptions | undefined
 ): TabPath[] {
-  const hrefState = getStateForHref(resolveHref(href), routeInfo);
+  const hrefState = linking
+    ? getStateFromPath(resolveHref(href), linking.config, routeInfo.segments)
```
jakub-agent: [🚨 **critical**] Inlining `getStateFromPath` here drops relative-href resolution.

`getStateForHref` (`src/link/getStateForHref.ts`) is three steps:

```ts
href = resolveHref(href);
href = resolveHrefStringWithSegments(href, routeInfo, options);
return store.linking ? getStateFromPath(href, store.linking.config, routeInfo.segments) : undefined;
```

The replacement only does `resolveHref(href)`. `resolveHrefStringWithSegments` is what turns `./details` or `../foo` into an absolute path against the current route before the path→state conversion — the same call `getNavigateAction` makes for regular navigation. Without it, a relative href is handed to `getStateFromPath` as-is and resolves against the wrong base (or fails to match, returning `[]` / `undefined`).

Both functions in this file are on the iOS link-preview prefetch path via `useNextScreenId`, so `<Link href="./details" preview>` is affected.

Secondary point: `HrefPreview.tsx:29` still calls `getStateForHref`. So the preview feature now has two different href→state implementations that disagree on relative hrefs. If the goal was to stop `preview/utils.ts` reading the module singleton, adding a `linking` parameter to `getStateForHref` and calling it from both places keeps one code path and preserves the resolution step.

Same issue at line 64.

packages/expo-router/src/imperative-api.tsx
```diff
@@ -12,14 +13,28 @@ export { router };
 export function useImperativeApiEmitter(
   ref: RefObject<NavigationContainerRef<ParamListBase> | null>
 ) {
+  const routeInfo = useRouteInfo();
+  const store = use(StoreContext);
+  if (!store) {
```
jakub-agent, line 15: [🏗️ **design**] This throw should break four existing test suites.

`useImperativeApiEmitter` is called unconditionally from `fork/NavigationContainer.tsx:86`, and these suites render that container standalone, outside any `StoreContext.Provider` (none of them reference `StoreContext` or mock `storeContext`):

- `src/react-navigation/native/__tests__/useRoutePath.test.ios.tsx`
- `src/react-navigation/native/__tests__/useLinkBuilder.test.ios.tsx`
- `src/react-navigation/native-stack/__tests__/index.test.ios.tsx`
- `src/react-navigation/stack/__tests__/index.test.ios.tsx`

Before this PR the hook read the module `store` and had no provider requirement, so they passed. I could not execute them to confirm (see the review summary) — worth checking before merge.

Separately, the `!store.linking` throw on line 21 looks unreachable. `linking` is only left unset when `routeNode` is null, and by then `ExpoRoot` has already returned the tutorial in development (`ExpoRoot.tsx:149`) or `useStore` has thrown `'No routes found'` in production — either way `NavigationContainer` never mounts. If you agree it's dead, dropping it removes a throw a reader has to reason about.

packages/expo-router/src/global-state/useStore.ts
```diff
@@ -11,101 +11,103 @@ import type { ExpoLinkingOptions, LinkingConfigOptions } from '../getLinkingConf
...
-    if (process.env.NODE_ENV === 'production') {
-      throw new Error('No routes found');
+      // In development, we will show the onboarding screen
+      rootComponent = Fragment;
     }
 
-    // In development, we will show the onboarding screen
-    rootComponent = Fragment;
-  }
+    storeRef.current = {
```
jakub-agent: [🏗️ **design**] Writing module state inside `useMemo` makes correctness depend on whether the cache hits.

The deps are `[context, linkingConfigOptions, navigationRef, serverUrl]`. `ExpoRoot` declares `linking = {}` as a default parameter, so when a caller omits `linking` it is a fresh object every render and this memo always misses. On a miss, `state: initialState` overwrites whatever `store.onStateChange` last recorded in `storeRef.current`. On a hit — a caller that passes a stable `linking` prop — it doesn't. Same code, two behaviours, decided by the caller's object identity.

Right now `ContextNavigator` has no subscriptions so it renders about once and this stays invisible, which is exactly what makes it worth naming.

AGENTS.md asks that the code work with and without React Compiler. A module-state write inside a memo body is the pattern that breaks when the compiler re-scopes memoization, because "how often does this run" stops being something you can read off the dependency array.

Two options: move the `storeRef` write into an effect (or a small setter the store owns) so render stays pure, or key the memo on the values it actually reads rather than on `linkingConfigOptions`' identity — `[context, linkingConfigOptions.metaOnly, navigationRef, serverUrl]` would at least be stable for the common case.

packages/expo-router/src/global-state/getNavigationAction.ts
```diff
@@ -7,44 +7,42 @@ import {
...
-  store.assertIsReady();
-  const navigationRef = store.navigationRef.current;
-
-  if (navigationRef == null) {
-    throw new Error(
-      "Couldn't find a navigation object. Is your component inside NavigationContainer?"
-    );
-  }
-  if (!store.linking) {
+  if (!linking) {
```
jakub-agent, line 44: [💡 suggestion] Two guards disappeared here and the tests that covered them were deleted with them: `store.assertIsReady()` and the `navigationRef == null` check with its `"Couldn't find a navigation object. Is your component inside NavigationContainer?"` message.

No crash results — `routingQueue.run` wraps the call in `if (ref.current)`, so a null ref can't reach `navigationRef.getRootState()`. But the failure mode changed: an action dispatched before the Root Layout mounts used to throw `'Attempted to navigate before mounting the Root Layout component...'`, and is now silently dropped by that `if`. The pre-existing `// TODO: Consider warning when ref.current is null — actions are silently dropped` sitting on that branch suggests this was already a known rough edge, and this PR widens it.

If silent-drop is the intended contract now, a short comment saying so is enough. If not, keeping `assertIsReady()` at the `routingQueue.run` boundary preserves the diagnostic. Either way the two removed cases in `getNavigationAction.test.ios.ts` were the only coverage for these paths.

packages/expo-router/src/global-state/__tests__/getNavigationAction.test.ios.ts
```diff
@@ -1,9 +1,37 @@
...
+  return getNavigateActionImplementation(
+    baseHref,
+    options,
+    type,
+    withAnchor,
+    singular,
+    isPreviewNavigation,
+    routeInfo,
+    {
+      navigationRef: store.navigationRef.current!,
```
jakub-agent: [💡 suggestion] The wrapper is a good way to keep the existing cases readable, but it hard-codes `redirects: []` and reads `linking` from the module `store`. The whole point of the new `NavigationActionContext` parameter is that redirects and linking now arrive as arguments, and neither is exercised: every test still runs with an empty redirect list.

One case passing a non-empty `redirects` and asserting `applyRedirects` receives it would cover the new parameter. An override argument on the wrapper would keep it cheap.

Minor: `describe(getNavigateAction, ...)` now resolves to this local wrapper rather than the exported function, so the suite name no longer tracks the real symbol. `describe(getNavigateActionImplementation, ...)` would.

packages/expo-router/src/react-navigation/native/types.tsx
```diff
@@ -161,7 +161,7 @@ export type LinkingOptions<ParamList extends object> = {
   /**
    * Custom function to parse the URL to a valid navigation state (advanced).
    */
-  getStateFromPath?: typeof getStateFromPathDefault;
+  getStateFromPath?: typeof getExpoStateFromPath;
```
jakub-agent: [💡 suggestion] `LinkingOptions` is public surface on `expo-router/react-navigation`, and this retypes one of its callbacks from the react-navigation signature to the Expo fork's. If the fork takes an extra `segments` argument, a consumer who supplies a `getStateFromPath` typed against the old signature stops type-checking — a source-level break with no changelog entry and no `## Accepted breaking changes` section on the PR.

Worth confirming whether the two signatures are actually assignable. If they are not, this needs a changelog line at minimum.

Related, same theme: `src/ui/common.tsx:92` switched from the imported `getStateFromPath` to `linking.getStateFromPath?.(...)`. That makes a user-provided override start affecting `useTriggersToScreens`, which reads like a deliberate improvement — but it is a behaviour change for anyone passing a custom `getStateFromPath`, and the optional call now routes a missing function into the `if (!state)` branch commented "This shouldn't occur". Both deserve a mention if intended.

packages/expo-router/src/ExpoRoot.tsx
```diff
@@ -134,11 +135,18 @@ function ContextNavigator({
...
+    initialState,
+    rootComponent,
+    linking: linkingConfig,
+    routeNode,
+  } = storeValue;
   useDomComponentNavigation();
 
-  if (store.shouldShowTutorial()) {
+  // TODO(@ubax): Revisit onboarding once route creation is React-owned.
+  if (process.env.NODE_ENV === 'development' && !routeNode) {
```
jakub-agent, line 150: [🐧 nit] The outer condition already requires `process.env.NODE_ENV === 'development'`, so the inner `if (process.env.NODE_ENV === 'development')` is always true and its `else { return null }` — with the comment "Ensure tutorial styles are stripped in production" — is unreachable. Inlining `shouldShowTutorial()` is the right call, but the nesting can collapse to one check.

Also: the `// TODO(@ubax): Refactor onReady logic and use listeners pattern` comment moved from `onNavigationReady` up to line 99, where it now sits above `ContextNavigator` and no longer describes the code beneath it. Looks like an accidental move during the edit.

packages/expo-router/src/global-state/store.ts
```diff
@@ -40,10 +36,18 @@ export function setHasAttemptedToHideSplash(value: boolean) {
   hasAttemptedToHideSplash = value;
 }
 
+export function maybeHideSplashScreen() {
```
jakub-agent, line 39: [🐧 nit] The extraction dropped the explanatory comment that was on `store.onReady`:

```
// NOTE(EvanBacon): `navigationRef.isReady` is sometimes not true when state is called initially.
```

That is the only thing in the file explaining why the splash hide is deferred through `requestAnimationFrame` rather than called directly. Without it the `requestAnimationFrame` looks arbitrary and someone will eventually remove it. Worth carrying over.

packages/expo-router/src/getLinkingConfig.ts
```diff
@@ -110,13 +110,19 @@ export function getLinkingConfig(
           if (typeof initialUrl === 'string') {
             initialUrl = applyRedirects(initialUrl, redirects);
             if (initialUrl && typeof nativeLinking?.redirectSystemPath === 'function') {
-              initialUrl = nativeLinking.redirectSystemPath({ path: initialUrl, initial: true });
+              initialUrl = nativeLinking.redirectSystemPath({
+                path: initialUrl,
```
jakub-agent: [🐧 nit] This hunk and the one below it are pure formatting — the two `redirectSystemPath({ path, initial: true })` calls reflowed onto multiple lines with no behaviour change. Same for the lone blank line added at `src/fork/useLinking.native.ts:56`.

Neither relates to moving store values into context. Reverting both makes the diff easier to review and keeps `git blame` on these lines pointing at whatever actually changed them.

jakub-agent (changes requested):

Iteration 2, PR 2/2 in the stack on top of #49176. The branch is now cleanly stacked (merge base equals the base head), so this diff is just this PR's layer: 21 files, +293/−233.

**Previous iteration's blocker is fixed.** `StoreContextValue` no longer extends `RouterStore`, the `{ ...store }` spread is gone, and `state` is off the context entirely. `fork/useLinking.ts` reads the module `store` directly again, so the frozen-getter path no longer exists. The changelog now sits under `💡 Others` with the right PR number. The store itself got a real trim — `shouldShowTutorial`, `onReady`, `assertIsReady`, `rootComponent` and `redirects` all left, which is more than I asked for and the right direction.

**Two new correctness problems, both silent failures on the iOS link-preview path.**

1. `useNextScreenId` now calls `useNavigationContainerRef()` imported from `../../react-navigation/native`. That barrel re-exports the vendored react-navigation hook (`src/react-navigation/native/index.tsx:43` → `export * from '../core'`), which is marked `@deprecated Import ... from 'expo-router' instead` at `src/react-navigation/core/index.tsx:83`. It builds a **fresh** ref that is never handed to a NavigationContainer, so the `'state'` listener never fires. The PR already puts `navigationRef` on `StoreContextValue` — that's the one to use.

2. `link/preview/utils.ts` swapped `getStateForHref` for a bare `getStateFromPath`, dropping the `resolveHrefStringWithSegments` step, so relative hrefs stop resolving against the current route. `HrefPreview.tsx:29` still uses `getStateForHref`, so the same feature now has two different href→state paths.

Beyond those: `imperative-api.tsx` throws during render when `StoreContext` is missing, which should break four existing suites that render `fork/NavigationContainer` standalone; and `storeRef.current = {...}` moved inside a `useMemo`, making a module-state write depend on cache hits.

**What I could and could not verify.** The two critical findings are read straight off the code — the hook implementations, the barrel re-export chain, and `createNavigationContainerRef`'s detached-`addListener` queueing are all quoted in the inline comments. I could **not** execute the test suite: the session worktree has no `node_modules`, and symlinking the main checkout's copy resolved workspace packages to an unrelated third worktree, which produced environment errors rather than signal. So the four-failing-suites claim below is a static prediction, not an observed result — please confirm with `et check-packages expo-router` before acting on it.

packages/expo-router/src/imperative-api.tsx
Ubax: Fix the tests, by providing the mock values and wrapping with `StoreContext`

packages/expo-router/src/global-state/useStore.ts
Ubax: Move this outside of `useMemo` to  just after it and use the values returned from `useMemo`

Also leave TODO comment to check if this is safe and needed (the setting of the ref in every redner)

packages/expo-router/src/global-state/getNavigationAction.ts
Ubax: Add this back

```
if (navigationRef == null) {
    throw new Error(
      "Couldn't find a navigation object. Is your component inside NavigationContainer?"
    );
  }
```

Also leave a TODO comment to see if this is needed

packages/expo-router/src/global-state/__tests__/getNavigationAction.test.ios.ts
Ubax: Keep as is for now

packages/expo-router/src/react-navigation/native/types.tsx
Ubax: This is accepted!

Add `## Accepted breaking changes` section and entry to this PR

packages/expo-router/src/ExpoRoot.tsx
Ubax: Fix the comment and remove inner if

packages/expo-router/src/global-state/store.ts
Ubax: Move the comment as well

packages/expo-router/src/getLinkingConfig.ts
Ubax: Revert the formatting changes if possible

packages/expo-router/src/fork/useLinking.native.ts
```diff
@@ -54,6 +54,7 @@ export function useLinking(
   const getStateFromPath = options?.getStateFromPath ?? getExpoStateFromPath;
   const getActionFromState = options?.getActionFromState ?? getActionFromStateDefault;
   const { segments } = useRouteInfo();
+
```
Ubax:
```suggestion
```

packages/expo-router/src/fork/useLinking.ts
```diff
@@ -10,7 +10,7 @@ import {
 } from 'react';
 
 import { ServerContext } from '../global-state/serverLocationContext';
-import { useExpoRouterStore } from '../global-state/storeContext';
+import { store } from '../global-state/store';
```
Ubax, line 13: Don;t use imperative store here. Do `const store = use(StoreContext)`

packages/expo-router/src/global-state/getNavigationAction.ts
```diff
@@ -1,50 +1,58 @@
...
 } from '../navigationParams';
+import type { NavigationContainerRef, ParamListBase } from '../react-navigation/native';
 import type { SingularOptions } from '../useScreens';
 import type { UrlObject } from './getRouteInfoFromState';
 import { findDivergentState, getPayloadFromStateRoute } from './stateUtils';
-import { store } from './store';
+import type { StoreContextValue } from './storeContext';
 import type { LinkToOptions } from './types';
 
+export type NavigationActionContext = Pick<StoreContextValue, 'linking' | 'redirects'> & {
```
Ubax: Why not Pick `navigationRef`?

packages/expo-router/CHANGELOG.md
```diff
@@ -50,6 +50,7 @@
 
 ### 🐛 Bug fixes
 
+- Keep linking route information in React scope. ([#49176](https://github.com/expo/expo/pull/49176) by [@Ubax](https://github.com/Ubax))
```
```suggestion
```

packages/expo-router/src/link/preview/HrefPreview.tsx
```diff
@@ -122,12 +135,15 @@ function PreviewForInternalRoutes() {
...
+function getParamsAndNodeFromHref(
+  hrefState: ResultState,
+  rootRouteNode: RouteNode | null | undefined
+) {
   const index = hrefState?.index ?? 0;
   if (hrefState?.routes[index] && hrefState.routes[index].name !== INTERNAL_SLOT_NAME) {
     const name = hrefState.routes[index].name;
     if (name === SITEMAP_ROUTE_NAME || name === NOT_FOUND_ROUTE_NAME) {
-      console.log(store.routeNode);
+      console.log(rootRouteNode);
```
jakub-agent: [💡 suggestion] These two `console.log`s ship to production. They fire whenever a previewed href resolves to the sitemap or not-found route, which is a reachable state rather than an impossible one — a `<Link href="/typo" preview>` gets there.

They're pre-existing, so not something this PR introduced. But this PR edits line 146 (`store.routeNode` → `rootRouteNode`), and the repo's contributing checklist asks for stray `console.log`s to be removed before submitting, so this is the natural moment.

The `error` string built just below is already the intended diagnostic. If the logs are load-bearing for debugging, a `process.env.NODE_ENV !== 'production'` guard would keep them out of shipped bundles.

packages/expo-router/src/global-state/__tests__/routingQueue.test.ios.ts
```diff
@@ -1,41 +1,45 @@
...
+import { getNavigateAction, type NavigationActionContext } from '../getNavigationAction';
 import { defaultRouteInfo } from '../getRouteInfoFromState';
 import { routingQueue } from '../routingQueue';
 
+const navigationActionContext = (
+  navigationRef: NavigationActionContext['navigationRef']
+): NavigationActionContext => ({
+  navigationRef,
+  linking: undefined,
+  redirects: [],
```
jakub-agent, line 10: [💡 suggestion] `redirects` is `[]` here and in the `getNavigationAction.test.ios.ts` wrapper, so no test passes a non-empty list. It's one of the two values this PR threads from `StoreContextValue` into `NavigationActionContext`, and the path from context to `applyRedirects` is unexercised — a regression that dropped it would also be green.

One case with a real redirect, asserting `applyRedirects` receives it, would close the gap. The `linking` half of the same context is covered, so this is the only hole.

Not blocking — worth a follow-up.

packages/expo-router/src/global-state/useStore.ts
```diff
@@ -18,93 +18,97 @@ import { getQualifiedRouteComponent } from '../useScreens';
...
+      navigationRef,
+      linking,
+      initialState,
+      rootComponent,
+      redirects,
+      routeNode,
+    };
+  }, [config, context, linkingConfigOptions, navigationRef, serverUrl]);
 
+  // TODO(@ubax): Check whether updating the store ref during every render is safe and needed.
```
jakub-agent, line 101: [🐧 nit] Noted and accepted as pre-existing — the base branch performs the same write on every render, so this PR carries the behaviour forward rather than introducing it. Leaving the detail here for whoever picks up the TODO.

Two consequences of writing module state in the render body:

1. `state: storeValue.initialState` overwrites whatever `store.onStateChange` last recorded, so a re-render of `ContextNavigator` would roll `store.state` back to the initial navigation state. `ContextNavigator` has no subscriptions today so it renders about once and this stays invisible — but `fork/useLinking.ts:383,414` reads `store.state` to build the browser path, so the correctness rests on that.
2. It runs twice under StrictMode and on every discarded render attempt, which is the pattern AGENTS.md's React Compiler requirement rules out.

When someone does revisit: an effect that syncs `storeRef` on change keeps render pure, and `state` probably shouldn't be re-seeded from `initialState` after the first commit.

packages/expo-router/src/link/preview/HrefPreview.tsx
jakub-agent, line 146: [🐧 nit] Noted and accepted as pre-existing — both `console.log`s exist on the base branch; this PR only changes the argument from `store.routeNode` to `rootRouteNode`.

For whenever this file is next touched: they fire whenever a previewed href resolves to the sitemap or not-found route, which is reachable (`<Link href="/typo" preview>`), so they reach production bundles. The `error` string built just below is already the intended diagnostic; a `process.env.NODE_ENV !== 'production'` guard would keep the logs out of shipped code if they're useful for debugging.

packages/expo-router/src/link/preview/useNextScreenId.ts
```diff
@@ -14,6 +15,11 @@ export function useNextScreenId(): [
 ] {
   const router = useRouter();
   const routeInfo = useRouteInfo();
+  const store = use(StoreContext);
+  if (!store) {
+    throw new Error('useNextScreenId must be rendered inside ExpoRoot.');
```
jakub-agent, line 20: [🐧 nit] Accepted for now — recording the shape for later.

A null `StoreContext` is handled three ways across the PR: throw here and at `imperative-api.tsx:16`, `use(StoreContext) ?? {}` at `HrefPreview.tsx:30`, and `use(StoreContext)?.routeNode` at `useSitemap.tsx:66`. All three sit under `ExpoRoot`, so all three branches are unreachable and untested; the cost is that a reader can't tell which are handling a real case.

If this gets unified later, a `useExpoRouterStore()` that throws plus a `useOptionalExpoRouterStore()` that doesn't would collapse it to one decision and one message — iteration 2 had that pair before the type narrowing removed it.

Same bucket: the `!store.linking` throw at `imperative-api.tsx:19` looks unreachable, since `ExpoRoot.tsx:150` returns the tutorial in development and `useStore` throws `'No routes found'` in production before the container mounts.

packages/expo-router/src/ExpoRoot.tsx
```diff
@@ -134,11 +136,18 @@ function ContextNavigator({
...
```
jakub-agent, line 150: [🐧 nit] The outer condition already requires `process.env.NODE_ENV === 'development'`, so the inner `if (process.env.NODE_ENV === 'development')` is always true and its `else { return null }` — commented "Ensure tutorial styles are stripped in production" — is unreachable. The nesting collapses to one check.

There are also two `TODO(@ubax)` comments stacked here: `Refactor onReady logic and use listeners pattern` at line 100, which no longer sits next to the `onReady` code it describes, and `Revisit onboarding once route creation is React-owned` at line 149.

packages/expo-router/src/global-state/store.ts
jakub-agent, line 39: [🐧 nit] The extraction from `store.onReady` dropped this comment:

```
// NOTE(EvanBacon): `navigationRef.isReady` is sometimes not true when state is called initially.
```

It's the only thing explaining why the splash hide goes through `requestAnimationFrame` rather than running directly. Without it the `requestAnimationFrame` reads as incidental and will eventually be tidied away.

jakub-agent (approved):

Iteration 4, PR 2/2 in the stack on top of #49176. Cleanly stacked (merge base equals the base head). 23 files, +424/−333.

## Verification

```
CI=1 jest                             →  Test Suites: 368 passed, 368 total
                                         Tests:       5 skipped, 4901 passed, 4906 total
                                         Snapshots:   273 passed, 273 total
oxlint --config oxlint.config.mjs .   →  clean
tsc -b tsconfig.all.json              →  one error, in expo-ui, not this PR
```

The typecheck failure is `../expo-ui/src/keyboard/textInputState.ts(1,43): error TS7016` — a missing `react-native/Libraries/Components/TextInput/TextInputState` declaration, which arrived with #48788 when this branch rebased onto main. Zero errors mention `expo-router`.

## All four criticals from iteration 3 are fixed

- `useStore.ts` and `getNavigationAction.ts` use the imported `getStateFromPath` again instead of the never-populated `linking.getStateFromPath!`.
- `ui/common.tsx` reverted entirely, so the silent `?.` variant is gone too.
- The `ExpoRoot` render throw no longer breaks the vendored-navigator suites: `useImperativeApiEmitter` moved out of `fork/NavigationContainer` into an `<ImperativeApiEmitter />` rendered inside the provider.

Also resolved: `routingQueue.run` dropped the duplicate `ref` parameter; the `LinkingOptions.getStateFromPath` retype in `react-navigation/native/types.tsx` is reverted; `getStateForHref` now takes `linking` so `HrefPreview` and `preview/utils.ts` share one href→state path.

Two improvements worth naming. `routingQueue.add` now does `[...queue, action]` instead of `push` — with `push` the array identity never changed, so the `useSyncExternalStore` snapshot compared equal and the effect that drains the queue wouldn't re-run. And `run` now preserves the queue when the ref isn't ready, with `navigationRef.addListener('ready', runQueue)` replaying it, replacing the old silent-drop path.

## Approving

The suite is green, lint is clean, `expo-router` type-checks, and every correctness finding from the previous three rounds is resolved.

Three remaining observations were reviewed with the author and accepted: the render-phase `storeRef` write and the `HrefPreview` `console.log`s are pre-existing behaviour that this PR carries forward rather than introduces, and the mixed null-`StoreContext` handling is fine for now. They're left inline as notes for whoever revisits these files, not as conditions.

The one item still worth a follow-up is test coverage: `redirects` is threaded from `StoreContextValue` into `NavigationActionContext` by this PR and no test passes a non-empty list, so a regression that dropped it on the way through would still be green.

packages/expo-router/CHANGELOG.md
```diff
@@ -66,6 +66,8 @@
 
 ### 💡 Others
 
+- Move Expo Router store values into React context. ([#49218](https://github.com/expo/expo/pull/49218) by [@Ubax](https://github.com/Ubax))
+- Keep linking route information in React scope. ([#49176](https://github.com/expo/expo/pull/49176) by [@Ubax](https://github.com/Ubax))
```
```suggestion
```

PR #49226, opened 2026-08-21
Enable inline PR review comments for AI code review

brentvatne:
## Why

The AI reviewer currently explains every finding inside one bottom comment. With `@expo/code-review-cli` 0.14.2, a finding anchored to a line in the PR diff can be posted as an inline review comment on that line — where the reader is — with the full rationale and suggestion in the thread. The main comment still lists every finding, but an inlined one renders short (title, links, one-line rationale) with a link to its thread.

## How

- Bump `.expo-code-review/cli-version` from 0.14.0 to 0.14.2.
- Enable the new root-only `inline` option (`maxComments: 20` caps notification fan-out per PR).

## Behavior notes

- The main comment stays the single source of truth: `/dismiss`, embedded state, and reply handling are unchanged.
- Inline threads update in place across re-reviews. A retired thread with human replies is stubbed with neutral text, never deleted.
- Replying in an inline thread counts as author feedback for that finding; clearing still requires citing the `id:` token, same as before.
- Failed or partial review runs never tear threads down.

Config validated with `ecr verify-config` under the pinned 0.14.2. Takes effect for PRs after this merges (`ecr ci` reads config from the trusted base commit).


brentvatne:
here ya go @tsapeta

PR #49227, opened 2026-08-21
fix(metro-config): Seal worker chunks to prevent chunk-splitting from applying to them

### LINKED ISSUE #49032: extractCommonChunk hoists modules out of web worker chunks, which cannot load __common — worker throws "Requiring unknown module" on startup

https://github.com/Save-All/expo-issue-49025-example

### Steps to reproduce

Re-filing #49025 with a linked reproduction, as `expo-bot` asked. Sorry for the noise on the first attempt — I put the repro inline instead of in a repo.

The linked project is `npx create-expo-app@latest --template blank` on `expo@57.0.14`, plus `react-dom` / `react-native-web` and four small files. **There is no `metro.config.js`** — nothing is configured.

```sh
git clone https://github.com/Save-All/expo-issue-49025-example
cd expo-issue-49025-example
npm install
npx expo export -p web
node check-chunks.mjs
```

`check-chunks.mjs` reads the emitted chunks and reports whether the worker chunk is self-contained, so the defect is visible without a browser:

```
worker chunk: worker-entry-f975693d9e29fd00502529c0626e811b.js
  defines:  184
  requires: 183

FAIL: module ids 183 are required but not defined here.
  183 is defined in: __common-787889e0cfb54335715d246ad90ad31d.js
```

Serving `dist` (or running `npx expo start --web`) shows the runtime failure.

**Expected:** `worker replied: shared-module-value`

**Actual:**

```
WORKER FAILED: Uncaught Error: Requiring unknown module "183".
other: shared-module-value
```

**Summary**

When the module that constructs a web worker is reached through a dynamic `import()`, and the worker's dependency graph overlaps another async chunk, the shared modules are hoisted into `__common-<hash>.js`. The page loads that chunk; the **worker never does**, because a worker has its own global scope and its own module registry. The worker chunk is left declaring a module whose dependency ids exist nowhere inside it, so the worker throws on startup and anything awaiting a reply from it hangs or rejects.

This does not reproduce when the worker's importer is in the eagerly-loaded graph, which is why it only appears once you code-split the library that owns the worker.

One note on the shape of the repro: `lib.js` (which constructs the worker) deliberately does **not** import `./shared`. `extractCommonChunk` walks chunk *pairs*, so if some other pair of async chunks consumes the shared module first, the worker chunk keeps its own copy and the build looks fine. Leaving the worker's only overlap with `other.js` makes the failure deterministic.

**Root cause**

`packages/@expo/metro-config/src/serializer/serializeChunks.ts`. A worker chunk is created with **both** `isAsync: true` and `isEntry: true`:

```ts
const isEntry = dependency.data.data.asyncType === 'worker';
gatherChunks(runtimePremodules, chunks, ..., graph, options, true /* isAsync */, isEntry);
```

Two of the three passes that move modules between chunks already exclude worker chunks by testing `isEntry`:

```ts
function removeEntryDepsFromAsyncChunks(entryChunk, chunks) {
  if (!chunk.isEntry && chunk.isAsync) { ... }         // guarded
}

function deduplicateAgainstKnownChunks(chunks, entryChunk, commonChunk) {
  if (!chunk.isEntry && chunk !== commonChunk) { ... } // guarded
}
```

`extractCommonChunk` is missing the same guard:

```ts
if (chunk !== chunk2 && chunk.isAsync && chunk2.isAsync) {   // no !isEntry
```

so as soon as a worker chunk and a page async chunk share a module, that module is pulled out of both and into `__common`.

**Suggested fix** — apply the guard the other two passes already use:

```ts
if (
  chunk !== chunk2 &&
  chunk.isAsync && chunk2.isAsync &&
  !chunk.isEntry && !chunk2.isEntry
) {
```

A worker cannot read the page's chunks, so keeping its own copies of shared modules is the only correct outcome. The cost is duplication proportional to what the worker actually needs.

Verified both ways in the linked repo. Adding the guard to `node_modules/expo/node_modules/@expo/metro-config/build/serializer/serializeChunks.js` and re-running the export gives:

```
worker chunk: worker-entry-75ab20f263a8bd8040253e5984e8323a.js
  defines:  183, 184
  requires: 183

PASS: the worker chunk is self-contained.
```

and the page prints `worker replied: shared-module-value`. Reverting the edit brings the failure back.

We hit this with a real third-party SDK that owns a WebAssembly worker: the same fix took its worker chunk from one module plus two dangling ids to three self-contained modules, and the SDK started working.

> Investigated and drafted with AI assistance. Every result above was reproduced against the stated versions from a clean clone of the linked repo, not inferred.

### Environment

```text
    System:
      OS: macOS 26.5.2
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 24.18.0 - /Users/gizmo/.nvm/versions/node/v24.18.0/bin/node
      Yarn: 1.22.22 - /Users/gizmo/.nvm/versions/node/v24.18.0/bin/yarn
      npm: 11.16.0 - /Users/gizmo/.nvm/versions/node/v24.18.0/bin/npm
      Watchman: 2026.04.20.00 - /opt/homebrew/bin/watchman
    Managers:
    SDKs:
      iOS SDK:
    IDEs:
      Android Studio: 2025.3 AI-253.31033.145.2533.15176040
    npmPackages:
      expo: ~57.0.14 => 57.0.14
      react: 19.2.3 => 19.2.3
      react-dom: 19.2.3 => 19.2.3
    Expo Workflow: managed
```

Transitively `@expo/metro-config@57.0.8` and `metro@0.84.4` (read from the installed `package.json`s, since `expo-env-info` does not list them). `serializeChunks.ts` on `main` at the time of writing still has the unguarded comparison.

### Expo Doctor Diagnostics

```text
```

kitten:
# Why

Resolves #49032

Web worker bundles are treated as self-sufficient and contain runtime modules and all modules they may execute. They're essentially supposed to be "sealed" and not "sparsened" (no deps must be removed/extracted from them)

This invariant was hidden and not explicitly enforced, causing loading of web workers that share modules with another async chunk to fail, which breaks web worker bundles.

Supersedes #49067: While currently worker chunks can theoretically be identified as `isEntry && isAsync`, this is a coincidence, and not explicitly enforced. Instead, while we have the information that a given chunk is a worker chunk, we should mark it as _sealed_ and exclude it from (common) chunk-splitting or other modifications.

Explicitly capturing `seal()` as a flag/method should stop us from regressing on this again. However, while this could also pass a `type: 'worker'` flag to chunks, this currently conflicts with the `isAsync` and `isEntry` design a little, which is why this wasn't added (yet)

# How

- When creating a `Chunk` we call `Chunk#seal` on it if it's a (web) worker chunk
- When `Chunk#sealed` is set, we skip chunk splitting/modifications entirely

Unrelated changes have been made following up on #46003 to clean up redundant Set/Array conversions and iterations.

# Test Plan

- Unit test added to verify sealed behaviour
- E2E updated to share `math.ts` module with worker and an `import(...)` async chunk

# Checklist

### DIFF
--- a/packages/@expo/metro-config/src/serializer/serializeChunks.ts
+++ b/packages/@expo/metro-config/src/serializer/serializeChunks.ts
@@ -867,10 +867,32 @@ function createRuntimeChunk(
 
 function makeChunkByPathLookupMap(chunks: Set<Chunk>): Map<string, Chunk> {
   const chunkByPath = new Map<string, Chunk>();
+  // First, we populate chunks with entry module paths
   for (const chunk of chunks) {
-    for (const module of chunk.deps) {
-      if (!chunkByPath.has(module.path)) {
-        chunkByPath.set(module.path, chunk);
+    for (const entry of chunk.entries) {
+      if (!chunkByPath.has(entry.path)) {
+        chunkByPath.set(entry.path, chunk);
+      }
+    }
+  }
+  // We then populate the chunks' module paths, excluding sealed chunks...
+  for (const chunk of chunks) {
+    if (!chunk.sealed) {
+      for (const module of chunk.deps) {
+        if (!chunkByPath.has(module.path)) {
+          chunkByPath.set(module.path, chunk);
+        }
+      }
+    }
+  }
+  // ...then populate with missing paths from sealed chunks.
+  // This gives precedence for modules from unsealed chunks after entry modules
+  for (const chunk of chunks) {
+    if (chunk.sealed) {
+      for (const module of chunk.deps) {
+        if (!chunkByPath.has(module.path)) {
+          chunkByPath.set(module.path, chunk);
+        }
       }
     }
   }
--- a/packages/@expo/metro-config/src/serializer/serializeChunks.ts
+++ b/packages/@expo/metro-config/src/serializer/serializeChunks.ts
@@ -718,15 +718,17 @@ function gatherChunks(
   entryChunks.add(entryChunk);
 
   function includeModule(entryModule: Module<MixedOutput>) {
+    const splitChunks = entryChunk.options.serializerOptions?.splitChunks !== false;
     for (const dependency of entryModule.dependencies.values()) {
+      const asyncType = dependency.data.data.asyncType;
+      const isWorker = asyncType === 'worker';
       if (!isResolvedDependency(dependency)) {
         continue;
       } else if (
-        dependency.data.data.asyncType &&
-        // Support disabling multiple chunks.
-        entryChunk.options.serializerOptions?.splitChunks !== false
+        asyncType &&
+        // Workers require standalone bundles even when ordinary chunk splitting is disabled.
+        (isWorker || splitChunks)
       ) {
-        const isWorker = dependency.data.data.asyncType === 'worker';
         const asyncChunks = gatherChunks(
           runtimePremodules,
           chunks,
--- a/packages/@expo/metro-config/src/serializer/serializeChunks.ts
+++ b/packages/@expo/metro-config/src/serializer/serializeChunks.ts
@@ -95,8 +95,7 @@ export async function graphToSerialAssetsAsync(
 
   // Create chunks for splitting.
   const chunks = new Set<Chunk>();
-
-  gatherChunks(
+  const entryChunks = gatherChunks(
     preModules,
     chunks,
     { test: pathToRegex(entryFile) },
@@ -107,8 +106,10 @@ export async function graphToSerialAssetsAsync(
     true
   );
 
-  const entryChunk = findEntryChunk(chunks, entryFile);
-
+  // TODO(@kitten): We know that the returned `entryChunks` should only have a single value
+  // with `!isAsync` and matching `.hasAbsolutePath(entryFile)` due to us only starting with
+  // an entry module. This is temporarily implicit and not enforced by an invariant
+  const entryChunk = entryChunks.values().next().value;
   if (entryChunk) {
     removeEntryDepsFromAsyncChunks(entryChunk, chunks);
 
@@ -762,10 +763,6 @@ function gatherChunks(
   }
 
   return entryChunks;
-}
-
-function findEntryChunk(chunks: Set<Chunk>, entryFile: string): Chunk | undefined {
-  return [...chunks.values()].find((chunk) => !chunk.isAsync && chunk.hasAbsolutePath(entryFile));
 }
 
 function removeEntryDepsFromAsyncChunks(entryChunk: Chunk, chunks: Set<Chunk>): void {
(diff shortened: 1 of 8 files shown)


kitten:
@expo-bot verify

packages/@expo/metro-config/src/serializer/serializeChunks.ts
```diff
@@ -107,8 +106,10 @@ export async function graphToSerialAssetsAsync(
     true
   );
 
-  const entryChunk = findEntryChunk(chunks, entryFile);
-
+  // TODO(@kitten): We know that the returned `entryChunks` should only have a single value
+  // with `!isAsync` and matching `.hasAbsolutePath(entryFile)` due to us only starting with
+  // an entry module. This is temporarily implicit and not enforced by an invariant
+  const entryChunk = entryChunks.values().next().value;
```
kitten, line 112: Note for review: This is mostly a side-effect of how `gatherChunks` was written. The return value was changed but the chunk matching + recursiveness + `pathToRegex` just make this a bit "weird", so from all the alternatives, adding this assumption for now with a comment seems to be the cleanest option without refactoring too much here

packages/@expo/metro-config/src/serializer/serializeChunks.ts
```diff
@@ -689,28 +715,38 @@ function gatherChunks(
...
+      const isWorker = asyncType === 'worker';
       if (!isResolvedDependency(dependency)) {
         continue;
       } else if (
-        dependency.data.data.asyncType &&
-        // Support disabling multiple chunks.
-        entryChunk.options.serializerOptions?.splitChunks !== false
+        asyncType &&
+        // Workers require standalone bundles even when ordinary chunk splitting is disabled.
+        (isWorker || splitChunks)
```
kitten, line 730: Drive-by fix: We shouldn't allow `splitChunks: false` to disable worker chunks.

packages/@expo/metro-config/src/serializer/serializeChunks.ts
```diff
@@ -831,10 +867,24 @@ function createRuntimeChunk(
 
 function makeChunkByPathLookupMap(chunks: Set<Chunk>): Map<string, Chunk> {
   const chunkByPath = new Map<string, Chunk>();
+  // We preopulate the chunks' module paths, excluding sealed chunks...
```
kitten: Drive-by fix/bug: Found this by chance while re-reading and sanity checking locally. I believe I wasn't aware we do such a lookup before, but assume that sealed chunks (workers) must not take precedence here, and this would otherwise make us load the worker chunk, triggering a latent bug

kitten:
@expo-bot verify

Kudo (approved):
thanks for the fix!

PR #36664, opened 2025-05-06
[speech][Android] Fix calling getVoices before TTS initialization

jakex7:
# Why

Calling the `getVoices` function before the TTS is initialized causes `textToSpeech.voices` to throw an error. As a result, the function returns an empty array instead of the list of available voices.

Resolves #36555.

# How

Although it's not the prettiest solution, we need to do a similar thing to what we do in the `speak` function. We're adding promises to a queue, and once TTS is initialized, all promises in the queue will be resolved.

# Test Plan

NCL -> TextToSpeech screen. The list of available voices should load correctly.

# Checklist

- [x] I added a `changelog.md` entry and rebuilt the package sources according to [this short guide](https://github.com/expo/expo/blob/main/CONTRIBUTING.md#-before-submitting)
- [x] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: updated a module plugin).
- [x] Conforms with the [Documentation Writing Style Guide](https://github.com/expo/expo/blob/main/guides/Expo%20Documentation%20Writing%20Style%20Guide.md)

PR #49228, opened 2026-08-22
[speech][Android] Settle queued promises when the TTS engine fails to initialize

henryj:
# Why

On Android, `getVoices` and `speak` calls made before the TTS engine is ready are queued and replayed from the `TextToSpeech` init callback (added in #36664). That callback only handles `TextToSpeech.SUCCESS`. When initialization fails — for example on devices with no TTS engine installed or enabled, which is common on Chinese-market phones shipping without Google services — the queued `getVoices` promises are never resolved and queued utterances are dropped silently.

The practical effect: any app that does `await Speech.getAvailableVoicesAsync()` on such a device hangs at that line forever, with no error to catch. We hit this in production; the repro chain is in the test plan.

# How

Add the missing failure branch to the init callback:

- resolve every queued `getVoices` promise with an empty list (the same shape callers get from a ready engine with no voices),
- fail every queued utterance through the existing `Events.SPEAKING_ERROR` event instead of dropping it,
- record the failure in a single `ttsState` enum (`PENDING`/`READY`/`FAILED`), so `getVoices`/`speak` calls made after the failed init answer immediately instead of re-queueing onto a queue that no future callback will drain (the lazy `TextToSpeech` property only initializes once, so a second callback never comes).

No behavior change when initialization succeeds.

# Test Plan

Verified on an Android 16 (API 36) emulator with this module compiled from source into a React Native app:

1. `adb shell pm disable-user --user 0 com.google.android.tts` (the emulator's only TTS engine), launch the app.
2. Before this change: `await Speech.getAvailableVoicesAsync()` never settles — no resolution, no rejection, nothing in logcat. In our app this froze a whole import pipeline at its voice-selection step.
3. With this change: the same call resolves with `[]` about a second after launch, and a `speak` call fires the speaking-error event.
4. Re-enabling the engine (`pm enable com.google.android.tts`) restores the normal voice list, confirming the success path is untouched.

### DIFF
--- a/packages/expo-speech/CHANGELOG.md
+++ b/packages/expo-speech/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### 🐛 Bug fixes
 
+- [Android] Resolve queued `getVoices`/`speak` calls when the TTS engine fails to initialize, instead of leaving their promises pending forever. ([#49228](https://github.com/expo/expo/pull/49228) by [@henryj](https://github.com/henryj))
+
 ### 💡 Others
 
 ## 57.0.1 - 2026-07-15
--- a/packages/expo-speech/android/src/main/java/expo/modules/speech/SpeechModule.kt
+++ b/packages/expo-speech/android/src/main/java/expo/modules/speech/SpeechModule.kt
@@ -21,6 +21,7 @@ const val speakingErrorEvent = "Exponent.speakingError"
 class SpeechModule : Module() {
   private val delayedUtterances: Queue<Utterance> = ArrayDeque()
   private val delayedGetVoices: Queue<Promise> = ArrayDeque()
+  private var ttsState = TtsState.PENDING
 
   override fun definition() = ModuleDefinition {
     Name("ExpoSpeech")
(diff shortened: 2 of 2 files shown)


vonovak:
@expo-bot pls open an PR based on this one where instead of two booleans `_ttsReady` and `_ttsFailed` we have a single enum field that gets set at the appropriate places. Also, drop the unnecessary comments, and try to find a nicer syntax for the the while (true) loop, if possible.

vonovak:
@expo-bot how do I get you to open a PR? is `/verify --fix` the only way?

henryj:
Pushed the requested changes to this branch (the bot wouldn't take the task, it only works on bot-authored PRs).

- `_ttsReady` and `_ttsFailed` are replaced by one `ttsState` field over a `PENDING` / `READY` / `FAILED` enum, set in the two branches of the init callback. The read sites in `getVoices` and `speak` are `when` blocks over it, so the `isTextToSpeechReady` getter is gone too.
- Dropped the comments I had added.
- The two `while (true)` + `poll()` loops are now `delayedGetVoices.forEach { ... }` and `delayedUtterances.forEach { ... }`, which is also how the success branch already iterates those queues. Neither branch clears them, and after `FAILED` nothing reads them again.

One thing I deliberately left alone: `ttsState` is not `@Volatile`, same as the old `_ttsReady`. Writes happen inside the `synchronized` block, but the reads in `getVoices` and `speak` don't, so there's no visibility guarantee across threads. Happy to add it if you want that closed here rather than in a separate change.

This revision is compile-checked only. The logic is the same as the one I ran through the emulator repro in the test plan.

vonovak (approved):

PR #49235, opened 2026-08-22
[android][maps] Don't enable my-location without the location permission

### LINKED ISSUE #49198: [expo-maps][Android] isMyLocationEnabled crashes the app with SecurityException when the location permission has not been granted

https://github.com/colaquecez/expo-maps-mylocation-crash

### Steps to reproduce

**Platform:** Android only. iOS is unaffected.
**Environment:** development build (`npx expo run:android`). 
**Package manager:** npm.

### Steps

1. Clone the reproduction, `npm install`, then `npx expo run:android` on a physical device or an emulator with Google Play services.
2. Make sure the app holds no location permission. A fresh install already satisfies that; otherwise `adb shell pm revoke com.anonymous.repro android.permission.ACCESS_FINE_LOCATION` (and `ACCESS_COARSE_LOCATION`).
3. Launch the app. `App.tsx` renders a single `GoogleMaps.View` with `properties={{ isMyLocationEnabled: true }}` and nothing else.

**Expected:** the map renders without the blue dot, which is how iOS behaves with the identical prop.

**Actual:** the whole app crashes as soon as the map composes, with an unhandled `java.lang.SecurityException`. It is not an error surfaced to JS; the process dies.

```
java.lang.SecurityException: my location requires permission ACCESS_FINE_LOCATION or ACCESS_COARSE_LOCATION
  com.google.android.gms.maps.internal.zzg.setMyLocationEnabled(com.google.android.gms:play-services-maps@@19.2.0:4)
  com.google.android.gms.maps.GoogleMap.setMyLocationEnabled(com.google.android.gms:play-services-maps@@19.2.0:1)
  com.google.maps.android.compose.MapUpdaterKt$MapUpdater$1$2$8.invoke(MapUpdater.kt:133)
  androidx.compose.runtime.Applier.apply(Applier.kt:180)
  ...
  androidx.compose.runtime.CompositionImpl.composeInitial(Composition.kt:732)
  androidx.compose.runtime.CompositionImpl.setContent(Composition.kt:699)
  com.google.maps.android.compose.GoogleMapKt$launchSubcomposition$1.invokeSuspend(GoogleMap.kt:252)
  ...
  android.app.ActivityThread.main(ActivityThread.java:10346)
```

The `composeInitial` / `setContent` frames show this is the map's first composition, so there is no point at which JS could catch it. No valid Google Maps API key is needed to reproduce: the crash happens in `setMyLocationEnabled`, before any tile is requested.

### Where it comes from

`expo-maps` passes the flag to `maps-compose` with no permission check, and the Android module has no permission check anywhere in it:

```kotlin
// android/src/main/java/expo/modules/maps/Records.kt
fun toMapProperties(): MapProperties {
  return MapProperties(
    isBuildingEnabled = isBuildingEnabled,
    isIndoorEnabled = isIndoorEnabled,
    isMyLocationEnabled = isMyLocationEnabled,   // <- no permission check
    ...
```

`maps-compose` then calls `GoogleMap.setMyLocationEnabled(true)` during composition, and the Maps SDK throws if neither `ACCESS_FINE_LOCATION` nor `ACCESS_COARSE_LOCATION` is held.

iOS handles the same prop safely. There it only decides whether a `UserAnnotation()` is rendered (`ios/AppleMapsViewiOS17.swift:99`, `ios/AppleMapsViewiOS18.swift:154`), which is a no-op without the permission rather than a crash. So the two platforms disagree about what the same prop does when the permission is missing.

### Why this is easy to hit in a real app

Any app that requests the location permission from the same screen that renders the map hits it. The map composes before the permission dialog is answered, so the first launch after a fresh install crashes.

The config plugin's `requestLocationPermission` option does not avoid it. That option only declares the permissions at build time and never requests anything at runtime.

`useLocationPermissions()` does not avoid it on its own either. It reads once on mount and has no listener, so a component that reads it on mount and then requests the permission elsewhere keeps a stale answer.

### Suggested fix

Either treat `isMyLocationEnabled` as `false` while the permission is missing, matching what iOS already does with the same prop, or document that the prop requires a granted permission so consumers know they must gate it themselves. A declarative prop should not be able to take the process down.

### Workaround

```tsx
const [permission] = Location.useForegroundPermissions();

<GoogleMaps.View
  properties={{ isMyLocationEnabled: permission?.granted ?? false }}
/>
```

Verified by A/B on a device: with the gate the app is stable; removing it reproduces the crash above with an identical stack.

---

### Environment

```text
    System:
      OS: macOS 26.5
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 25.6.1 - /opt/homebrew/bin/node
      npm: 11.9.0 - /opt/homebrew/bin/npm
    Managers:
    SDKs:
      iOS SDK:
    IDEs:
      Android Studio: 2025.2 AI-252.27397.103.2522.14514259
    npmPackages:
      expo: ~57.0.15 => 57.0.15
      react: 19.2.3 => 19.2.3
    npmGlobalPackages:
      eas-cli: 16.32.0
    Expo Workflow: managed


`expo-maps`: 57.0.2 (latest at the time of writing). Device: Samsung SM-S721B, Android 16 (API 36).
```

### Expo Doctor Diagnostics

```text
```

dennytosp:
# Why

Fixes #49198.

`GoogleMaps.View` with `properties={{ isMyLocationEnabled: true }}` crashes the whole Android app when the location permission has not been granted:

```
java.lang.SecurityException: my location requires permission ACCESS_FINE_LOCATION or ACCESS_COARSE_LOCATION
  com.google.android.gms.maps.GoogleMap.setMyLocationEnabled(play-services-maps@@19.2.0:1)
  com.google.maps.android.compose.MapUpdaterKt$MapUpdater$1$2$8.invoke(MapUpdater.kt:133)
  ...
  androidx.compose.runtime.CompositionImpl.composeInitial(Composition.kt:732)
```

[`MapPropertiesRecord.toMapProperties`](https://github.com/expo/expo/blob/main/packages/expo-maps/android/src/main/java/expo/modules/maps/Records.kt#L251) passed the flag through unconditionally, and nothing else in the Android module checked the permission. `maps-compose` then calls `GoogleMap.setMyLocationEnabled(true)` during composition, and the Maps SDK throws.

The `composeInitial` frame is the point: this happens on the map's first composition, so there is no JS boundary to catch it and the process dies. Any app that asks for the location permission on the same screen that renders the map hits this on a fresh install, because the map composes before the dialog is answered.

iOS treats the same prop as a request, not a command — it only decides whether to render `UserAnnotation()` ([`AppleMapsViewiOS17.swift:99`](https://github.com/expo/expo/blob/main/packages/expo-maps/ios/AppleMapsViewiOS17.swift#L99), [`AppleMapsViewiOS18.swift:154`](https://github.com/expo/expo/blob/main/packages/expo-maps/ios/AppleMapsViewiOS18.swift#L154)), which is a no-op without the permission. So the two platforms disagreed about what the prop means when the permission is missing.

# How

Android now treats it as a request too. `toMapProperties` takes the permission state and ands it in:

```kotlin
fun toMapProperties(hasLocationPermission: Boolean): MapProperties {
  ...
  isMyLocationEnabled = isMyLocationEnabled && hasLocationPermission,
```

`GoogleMapsView` supplies it from the view's context:

```kotlin
internal fun Context.hasLocationPermission(): Boolean =
  ContextCompat.checkSelfPermission(this, Manifest.permission.ACCESS_FINE_LOCATION) == PackageManager.PERMISSION_GRANTED ||
    ContextCompat.checkSelfPermission(this, Manifest.permission.ACCESS_COARSE_LOCATION) == PackageManager.PERMISSION_GRANTED
```

The parameter is required rather than defaulted, so a future call site has to decide rather than silently inherit a default.

Either permission is accepted, matching what the Maps SDK itself requires — coarse location is enough for the blue dot.

Note on scope: the permission is read when the map composes. Granting the permission afterwards shows the dot on the next recomposition, which is what a normal permission flow produces since the app re-renders after the grant. Observing the permission live would need a lifecycle observer, which felt like more than this crash fix should carry — happy to add it if you would rather have it here.

# Test Plan

`packages/expo-maps/android/src/test/java/expo/modules/maps/MapPropertiesRecordTest.kt`, four cases, run with `et native-unit-tests -p android --packages expo-maps`.

```
MapPropertiesRecordTest  tests=4 failures=0 errors=0 skipped=0

  PASS my location stays disabled without the location permission
  PASS my location is enabled with the location permission
  PASS my location stays disabled when it was not requested
  PASS the other properties are unaffected by the location permission

BUILD SUCCESSFUL
Finished android unit tests successfully.
```

I also mutation-checked the tests: reverting the guard to `isMyLocationEnabled = isMyLocationEnabled` while keeping the new parameter makes them fail, so they are testing the guard rather than passing by construction.

```
> Task :expo-maps:testDebugUnitTest FAILED
MapPropertiesRecordTest > my location stays disabled without the location permission FAILED
4 tests completed, 1 failed
BUILD FAILED
```

`expo-maps` had no `src/test` directory, so this PR adds one plus a `testImplementation 'junit:junit:4.13.2'` line, matching how `expo-secure-store` wires its Android unit tests.

I did not run the reproduction app on a device — I do not have a Play-services emulator image set up here. The crash path is `setMyLocationEnabled(true)` throwing without the permission, and this change stops `true` from reaching it, which is what the reporter's stack trace points at.

# Checklist

- [x] Added unit tests.

### DIFF
--- a/packages/expo-maps/CHANGELOG.md
+++ b/packages/expo-maps/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 ### 🐛 Bug fixes
 
+- [Android] Fix `isMyLocationEnabled` crashing the app with a `SecurityException` when the location permission has not been granted. The prop is now a request that is ignored without the permission, matching iOS. ([#49198](https://github.com/expo/expo/issues/49198), [#49235](https://github.com/expo/expo/pull/49235) by [@dennytosp](https://github.com/dennytosp))
 - [iOS] Stop `AppleMaps.View`'s `onMapClick` from firing when the tap landed on a marker, an annotation, or a drawn shape, as documented. ([#49193](https://github.com/expo/expo/pull/49193) by [@dennytosp](https://github.com/dennytosp))
 - [Android] Return the `Promise` from `GoogleMaps.View`'s imperative `setCameraPosition` so callers can `await` it and catch the `Animation cancelled` rejection (matches iOS). ([#46421](https://github.com/expo/expo/pull/46421) by [@chownation](https://github.com/chownation))
 - [Android] Fixed Google Maps marker info windows reserving blank snippet space when `snippet` is omitted. ([#47271](https://github.com/expo/expo/pull/47271) by [@eliotgevers](https://github.com/eliotgevers))
--- a/packages/expo-maps/android/build.gradle
+++ b/packages/expo-maps/android/build.gradle
@@ -39,4 +39,6 @@ dependencies {
   implementation 'androidx.compose.material3:material3-android:1.3.2'
 
   implementation 'com.google.maps.android:maps-compose:6.10.0'
+
+  testImplementation 'junit:junit:4.13.2'
 }
--- a/packages/expo-maps/android/src/main/java/expo/modules/maps/GoogleMapsView.kt
+++ b/packages/expo-maps/android/src/main/java/expo/modules/maps/GoogleMapsView.kt
@@ -1,12 +1,15 @@
 package expo.modules.maps
 
+import android.Manifest
 import android.annotation.SuppressLint
 import android.content.Context
+import android.content.pm.PackageManager
 import android.graphics.Bitmap
 import android.graphics.Point
 import android.graphics.drawable.Drawable
 import android.view.MotionEvent
 import androidx.compose.foundation.layout.fillMaxSize
+import androidx.core.content.ContextCompat
 import androidx.core.graphics.drawable.toBitmap
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
@@ -111,7 +114,7 @@ class GoogleMapsView(context: Context, appContext: AppContext) :
       modifier = Modifier.fillMaxSize(),
       cameraPositionState = cameraState,
       uiSettings = props.uiSettings.value.toMapUiSettings(),
-      properties = props.properties.value.toMapProperties(),
+      properties = props.properties.value.toMapProperties(context.hasLocationPermission()),
       contentPadding = props.contentPadding.value.let {
         PaddingValues(start = it.start.dp, end = it.end.dp, top = it.top.dp, bottom = it.bottom.dp)
       },
@@ -457,3 +460,11 @@ private fun MapCircles(
     )
   }
 }
+
+/**
+ * Whether the app can show the user's location. The Maps SDK throws a `SecurityException` from
+ * `setMyLocationEnabled` unless one of these is granted.
+ */
+internal fun Context.hasLocationPermission(): Boolean =
+  ContextCompat.checkSelfPermission(this, Manifest.permission.ACCESS_FINE_LOCATION) == PackageManager.PERMISSION_GRANTED ||
+    ContextCompat.checkSelfPermission(this, Manifest.permission.ACCESS_COARSE_LOCATION) == PackageManager.PERMISSION_GRANTED
--- a/packages/expo-maps/android/src/main/java/expo/modules/maps/Records.kt
+++ b/packages/expo-maps/android/src/main/java/expo/modules/maps/Records.kt
@@ -248,12 +248,14 @@ data class MapPropertiesRecord(
   @Field
   val minZoomPreference: Float = 3.0f
 ) : Record {
-  fun toMapProperties(): MapProperties {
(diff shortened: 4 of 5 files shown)


brentvatne:
@expo-bot verify with fable

PR #49236, opened 2026-08-22
[web][router] Honor base URL in development

### LINKED ISSUE #47701: expo-router: `experiments.baseUrl` is ignored in development — dev and exported web builds generate different URLs

https://github.com/sensei-woo/expo-router-baseurl-dev-repro

### Summary

With `experiments.baseUrl` set, expo-router generates **different URLs in development than in the exported site**. Production output includes the base URL in `Link` hrefs (and applies it in `router.push`/deep-link state); the development server silently drops it. There is no supported way to run the web dev server behind the base path the app is configured for — which is exactly the shape used by anyone embedding an Expo web app in a larger site (reverse proxy `location` block, Next.js rewrite, in-process mount).

The cause is that the router forks hard-disable base-URL handling in development ([`packages/expo-router/src/fork/getPathFromState-forks.ts`](https://github.com/expo/expo/blob/main/packages/expo-router/src/fork/getPathFromState-forks.ts), same in `getPathFromState.ts` and `getStateFromPath-forks.ts`, unchanged in the published `expo-router@57.0.4` build output):

```ts
export function appendBaseUrl(path, baseUrl = process.env.EXPO_BASE_URL) {
  if (process.env.NODE_ENV !== 'development') {
    if (baseUrl) {
      return `/${baseUrl.replace(/^\/+/, '').replace(/\/$/, '')}${path}`;
    }
  }
  return path;
}
```

### Steps to reproduce

Platform: **web**. Environment: dev server (`npx expo start`) vs static export. Package manager: **npm**. No third-party libraries.

```sh
npm install

# Production: anchors include the base URL
npx expo export --platform web
grep -o 'href="[^"]*"' dist/index.html
# → href="/embed/about" ✅

# Development: the same anchor drops it
npx expo start --web --port 8081
curl -s http://localhost:8081/ | grep -o 'href="[^"]*"'
# → href="/about" ❌
```

To see the practical failure, mount the dev server under the base path (`node proxy.js` in the repro — a 30-line reverse proxy serving Metro at `/embed`, standing in for any real embed setup):

```sh
node proxy.js                      # serves the dev server at http://localhost:8082/embed
open http://localhost:8082/embed   # renders fine…
```

…but the Home page's link points at `/about` instead of `/embed/about`, so clicking it navigates outside the mount point and 404s. The exported site served under the same proxy works correctly.

**Expected:** development and production generate the same URLs when `baseUrl` is configured — or at minimum an opt-in to apply it in development.

I understand why the guard exists: plain `npx expo start` serves at `/`, so unconditionally prefixing would break the default dev flow. Two possible designs, happy to send a PR for whichever you prefer:

1. Honor `EXPO_BASE_URL` in development whenever it is set (the default dev flow leaves it unset, so nothing changes for existing projects), or
2. An explicit opt-in — we currently maintain a patch adding an `EXPO_PUBLIC_USE_BASE_URL` escape hatch to the three fork files, set only by our embedded dev boot, and it has worked well in daily use.

### Environment

```
    System:
      OS: macOS 26.5
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 24.8.0 - /Users/me/.local/share/mise/installs/node/24.8.0/bin/node
      npm: 11.6.0 - /Users/me/.local/share/mise/installs/node/24.8.0/bin/npm
    SDKs:
      iOS SDK:
    IDEs:
      Xcode: 26.3/17C519 - /usr/bin/xcodebuild
    npmPackages:
      expo: ~57.0.4 => 57.0.4
      react: 19.2.3 => 19.2.3
      react-dom: 19.2.3 => 19.2.3
    Expo Workflow: managed
```

### Expo Doctor Diagnostics

```
```

huytdps13400:
## Summary

- apply configured `EXPO_BASE_URL` while generating paths in development
- strip the same base URL while parsing incoming development paths
- keep the default development flow unchanged when no base URL is configured
- run existing Link and path round-trip coverage under `NODE_ENV=development`

Closes #47701

## Why

Expo Router previously hard-disabled all three base URL helpers in development. As a result, `experiments.baseUrl` produced `/embed/about` in exported web builds but `/about` in development, so links escaped reverse-proxy or embedded mount paths.

The base URL is already opt-in through `EXPO_BASE_URL`; applying it whenever present makes development match production without adding a second configuration flag.

### TDD reproduction

With the existing base URL tests forced to run under `NODE_ENV=development`, the unmodified implementation failed 6 assertions out of 106 tests:

- Link href: expected `/base_url/foo`, received `/foo`
- incoming `/expo/prefix/bar` resolved to `+not-found` instead of `bar`
- generated path: expected `/expo/bar`, received `/bar`

### Verification

- Focused Jest paths: 4/4 suites, 106/106 tests, 16/16 snapshots passed
- `et check-packages expo-router`: all 35 build/typecheck/depscheck/test/lint tasks passed
- Expo Router full Jest suite passed through the package gate
- lint: 0 warnings, 0 errors
- format check: all 864 matched files correctly formatted
- `git diff --check`
- no dependency or lockfile changes


huytdps13400:
Updated this branch with current `main` in `e63ef2f`. The only conflict was the Expo Router changelog; the resolution keeps both this PR entry and the newer upstream #49351 entry. The source/test hunks merged without conflicts and the PR delta remains the same six files.

Post-merge verification:
- focused Web + Node Jest projects: 4/4 suites, 106/106 tests, 16/16 snapshots
- `pnpm run lint`: pass
- `pnpm run format -- --check`: pass
- no unmerged paths or unstaged tracked changes
- fresh Expo Router architecture review: no required corrections

A full package build/typecheck was also attempted, but this storage-conscious sparse checkout intentionally omits unrelated workspace packages such as `expo-server`, `expo-symbols`, and `@expo/ui`; the focused routing coverage above runs cleanly. GitHub now reports the PR mergeable, and CI has started on the updated head.

PR #49239, opened 2026-08-22
[expo-audio] Preserve the record(forDuration:) limit across an audio session interruption

stvrmrz:
# Why

On iOS, a recording armed with `record({ forDuration })` loses its duration limit if an audio session interruption occurs. `handleInterruptionBegan` pauses the recorder, and the `.shouldResume` path reaches `startRecording()`, which resumes with a **bare `ref.record()`** — carrying no duration. The recording then continues indefinitely.

An incoming call, an alarm, or Siri is enough to trigger it, and nothing in the app is called, so there is no opportunity to re-arm from JS.

Measured on a physical iPhone 17 Pro Max (iOS 26.5), `expo-audio@57.0.3`: a recorder armed `forDuration: 6.0`, paused, then resumed via the bare call captured **14.06 s and was still going**.

# What this changes

`AudioRecorder` remembers the duration a recording was armed with, and `startRecording()` re-applies it instead of resuming unbounded.

Two details that matter:

**The limit is re-applied absolutely, not as a remainder.** `record(forDuration:)` stops when the recorder's own `currentTime` *reaches* the value, and `currentTime` is cumulative across pause/resume — so the original limit is correct. Re-arming with a computed remainder was measured to end the recording early: 3.98 s against a 6.0 s arm.

**`resetDurationTracking()` deliberately does not clear the limit.** `updateStateForDirectRecording()` calls that helper *after* `record(forDuration:)` has armed the limit, so clearing it there would discard the value the caller just set — and the first interruption would resume unbounded again. The limit is cleared at capture boundaries instead: `prepare`, `stopRecording`, `didFinish`, `encodeErrorDidOccur`, and `handleMediaServicesReset`.

If the allowance is already spent when the resume arrives, the recorder stops rather than resuming, which routes through the existing `didFinish` path and emits `isFinished` as usual.

All three arming paths (`record({ forDuration })`, `record({ atTime, forDuration })`, and the deprecated `recordForDuration`) now go through one method, so none of them can bypass the limit.

# Test Plan

Verified on a physical device against `expo-audio@57.0.3` with the equivalent change applied:

- armed 6.0 s, paused, resumed bare → **before:** 14.06 s and climbing; **after:** stops at 6.000 s
- a real Clock-alarm interruption during a 120 s-capped recording → resumes and ends at **exactly 120.0000 s** of audio, verified by parsing the CAF `data` chunk, across 4 trials
- one canonical file throughout; the recorder's `didFinish` fires normally at the limit

I could not run the repo's own iOS test suite for `expo-audio`; the verification above is device-level against the published package.

### DIFF
--- a/packages/expo-audio/CHANGELOG.md
+++ b/packages/expo-audio/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### 🛠 Breaking changes
 
+- Paused time is now excluded from the `AudioRecorder` duration limit. ([#49239](https://github.com/expo/expo/pull/49239) by [@stvrmrz](https://github.com/stvrmrz) and [@behenate](https://github.com/behenate))
 - [Android] Aligned the default audio focus request on Android 7.0–7.1 with newer versions by using transient exclusive focus when no interruption mode has been configured. ([#49101](https://github.com/expo/expo/pull/49101) by [@behenate](https://github.com/behenate))
 
 ### 🎉 New features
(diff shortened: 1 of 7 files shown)


brentvatne:
@expo-bot check with fable

behenate (approved):
I added a few changes on the iOS side and added Android and Web fixes for platform compatibility. Tested on all three platforms, seems to be working ok. Returning from background will be handled in a separate PR.

PR #49241, opened 2026-08-22
[module-scripts] Update `oxlint` to 1.79.0 and `oxlint-config-universe` to 0.2.0

tsapeta:
# Why

`oxlint-config-universe` 0.2.0 is out and requires `oxlint` 1.79, so this updates both.

# How

oxlint 1.79 turns on a set of React Compiler-derived rules by default and the repo has a few hundred pre-existing findings (mostly in expo-router), so the base config keeps them off to make this update behavior-neutral. We can re-enable them rule by rule as the findings get triaged; the stacked #49242 starts with the five cheapest ones.

A few more adjustments in the base config: `no-inner-declarations` started flagging exported functions inside TS namespaces, which the new `namespaces: 'allow'` option fixes without losing the rule. The `no-did-mount-set-state` / `no-did-update-set-state` pair is off because oxlint (unlike ESLint's default mode) also flags `setState` inside async callbacks, hitting the common fetch-on-mount pattern. The TS overrides that 0.2.0 now applies itself (or no longer enables) are dropped.

Source changes are small and mechanical: two useless `export {}` statements flagged by the newly enabled `typescript/no-useless-empty-export` are removed (an expo-observe test and an `@expo/cli` e2e helper), the LogBox DOM polyfill's inline disables now also cover `import/namespace` (oxlint 1.79 newly reports assignments to namespace-import members), and expo-router's navigation tree reducer gets inline disables where the stricter `rules-of-hooks` rejects effect-event calls made outside of effects it recognizes (event handlers and a custom effect wrapper).

# Test Plan

Linted all 130 oxlint packages plus the three oxlint apps: no errors anywhere, and the only warnings are in `@expo/metro-config` test fixtures and exist on main as well. `et check-packages` passes for the touched packages, including the expo-observe and expo-router jest suites.


alanjhughes (approved):

packages/expo-module-scripts/oxlint.config.base.js
```diff
@@ -52,16 +52,15 @@ export default defineConfig({
...
       files: ['**/*.ts', '**/*.tsx', '**/*.d.ts'],
       rules: {
-        // Already handled by TypeScript itself, enabling them blocks legitimate type overloads.
-        'no-redeclare': 'off',
-        'typescript/no-redeclare': 'off',
         // Handled by TypeScript.
         'no-unused-expressions': 'off',
-        'typescript/no-unused-expressions': 'off',
         'no-unused-vars': 'off',
-        'typescript/no-unused-vars': 'off',
```
kitten, line 62: I kind of doubt these were actually turned off in the upstream preset, which means we're now back to duplicates, right?

tsapeta: Good call to double-check, but we're safe here on both counts:

1. `no-redeclare` is now turned off for TS files in the upstream preset itself (`shared/core.js` has a `**/*.ts` override with `'no-redeclare': 'off'`, added because the core rule flags legal value/type declaration merging, see oxc-project/oxc#25936).
2. The `typescript/*` spellings are just aliases in oxlint, not separate rules. There's a single implementation, so our `'no-unused-vars': 'off'` / `'no-unused-expressions': 'off'` lines above already disable both spellings. You can see this by setting only `typescript/no-unused-vars: 'error'` in a config: the diagnostic comes back as `eslint(no-unused-vars)`.

I also verified empirically with oxlint 1.79.0 and this exact config on a probe `.ts` file with an unused variable, a `var` redeclaration, and an unused expression: none of the three rules fire, only `no-var` does.

kitten: Thanks for checking! 🙌

PR #22470, opened 2023-05-12
[cli] Fix modern manifest serving for dev client without expo-updates

wschurman:
# Why

When a dev client doesn't have `expo-updates` installed, it uses its own manifest downloader and parser. This parser only supports a json body.

Since https://github.com/expo/expo/pull/22168, only a multipart is served by default from `npx expo start` including with the `--dev-client` flag. This causes the fallback (non-expo-updates) parser to fail on the dev client.

Closes ENG-8529.

# How

The most correct solution is to make `npx expo start` support all the content-types that the protocol supports: https://docs.expo.dev/technical-specs/expo-updates-1/

So that's what this PR does. It adds support for application/json and application/expo+json.

It also adds a small change to the fallback manifest downloader to specify content type, though the change to the CLI fixes the dev client even without this change since text/plain is the same response.

# Test Plan

CLI:
1. make change, `nexpo start --dev-client`

Android
1. `et gba -n testwat expo-dev-menu expo-dev-client expo-dev-launcher expo-manifests expo-updates-interface`
2. Comment out some stuff in gradle and interface to make build work with `et gba`
3. Run app in android studio, load, see it no longer has an error.

# Checklist

- [ ] Documentation is up to date to reflect these changes (eg: https://docs.expo.dev and README.md).
- [ ] Conforms with the [Documentation Writing Style Guide](https://github.com/expo/expo/blob/main/guides/Expo%20Documentation%20Writing%20Style%20Guide.md)
- [ ] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: updated a module plugin).

PR #13825, opened 2021-07-29
[expo-dev-launcher] interpret URLs that return a JSON content-type as manifest URLs

esamelson:
# Why

Fixes #13711 

exp.host project URLs with `/index.exp?sdkVersion=42.0.0` respond with 200s even if no special headers are sent. This breaks our assumption in the `DevLauncherManifestParser.isManifestUrl` method.

# How

If the HEAD request comes back with `Content-Type: application/json` assume it is a manifest URL.

# Test Plan

Tested this manually by ensuring on both platforms that the following URLs opened successfully:
- `https://exp.host/@esamelson/sdk42updates`
- `https://exp.host/@esamelson/sdk42updates/index.exp?sdkVersion=42.0.0`
- `http://localhost:8081` after running `react-native start` in a bare project
- `http://localhost:8081/index.bundle?platform=[ios,android]&dev=true&hot=false&minify=false`

Also added unit tests on both platforms. Relies on https://github.com/expo/expo/pull/13824 for the iOS unit test setup.

# Checklist

- [ ] Documentation is up to date to reflect these changes (eg: https://docs.expo.io and README.md).
- [ ] This diff will work correctly for `expo build` (eg: updated `@expo/xdl`).
- [ ] This diff will work correctly for `expo prebuild` & EAS Build (eg: updated a module plugin).

PR #49250, opened 2026-08-23
fix(dev-launcher): [iOS] take the lightweight manifest path for a live dev server [retracted — do not merge]

### LINKED ISSUE #49249: [dev-launcher] Local expo start dev server misclassified as published manifest via expo+json content-type

https://github.com/saileshbro/repro-dev-launcher-49249 — a stock `create-expo-app` (SDK 57) plus `repro.sh`, which reproduces the misclassification with `curl` in 10 seconds, no device needed.

### Summary

`expo-dev-launcher`'s manifest-vs-bundle probe misclassifies a stock `expo start` dev server as a **published EAS-Update manifest URL** on every launch. This routes every local dev launch through the heavy published-manifest branch (updates config construction, manifest parsing, `expo-updates` integration checks) instead of the direct bundle fast path, adding seconds to each dev-client cold start.

### Root cause

`EXDevLauncherManifestParser.isManifestURLWithCompletion:` (iOS) and `DevLauncherManifestParser.isManifestUrl()` (Android) send a HEAD request with:

```
Accept: application/expo+json,application/json
```

and classify the URL as a published manifest whenever the response content-type is neither `text/html`-prefixed nor `*/javascript`-containing:

```objc
// packages/expo-dev-launcher/ios/Manifest/EXDevLauncherManifestParser.m
NSString *contentType = headers[@"Content-Type"];
if (contentType && ![contentType hasPrefix:@"text/html"] && ![contentType containsString:@"/javascript"]) {
  completion(YES); // treated as a published manifest URL
  return;
}
```

However, Expo CLI's manifest middleware (`packages/@expo/cli/src/start/server/middleware/ExpoGoManifestHandlerMiddleware.ts`) negotiates the response content type from the request's `Accept` header — and since the launcher's own probe explicitly accepts `application/expo+json`, a completely ordinary `expo start` server answers with exactly that:

```
$ curl -sI -H "Accept: application/expo+json,application/json" http://localhost:8081/
HTTP/1.1 200 OK
content-type: application/expo+json
```

`application/expo+json` matches neither exclusion, so the live dev server is misclassified as a published project on **every single launch**.

This rule was introduced in #13825 (2021) so that exp.host published URLs returning JSON would be recognized. It predates the CLI's Accept-header negotiation and now catches the launcher's own dev-server probes.

### Repro

1. Start any Expo dev project: `npx expo start --dev-client`
2. Launch the app in a dev client build (deep link or manually)
3. Observe via a proxy or server logs that each launch issues a HEAD to the dev server that receives `content-type: application/expo+json`, after which the launcher proceeds down the `_updatesInterface` / manifest-parsing branch instead of calling the direct `launchReactNativeApp()` path

Measured impact on a mid-size RN app (iOS simulator, Metro warm): the manifest branch adds multiple seconds before the bundle request fires, compared with the fast path.

### Environment

```text
expo-dev-launcher 57.0.x (present on main)
expo CLI dev server (any recent SDK)
iOS + Android (both parsers share the same logic)
```

### Suggested fix

Treat `expo+json` as a live Expo dev server (not a published manifest) in both parsers. Plain `application/json` should continue to classify as a manifest so #13825's exp.host behavior is preserved.

saileshbro:
# Why

Fixes #49249.

**This PR has been rewritten.** The original approach classified `application/expo+json` as "not a
manifest"; [@alanjhughes' `/verify` run](https://github.com/expo/expo/pull/49250#issuecomment-5397737713)
showed that breaks dev-client launches, and it's right — see the comment below for the full
correction. The manifest is a manifest; what a live dev server doesn't need is the `expo-updates`
fetch.

`expo-dev-launcher`'s manifest probe sends `Accept: application/expo+json,application/json`, and the
CLI negotiates from it, so a stock `expo start` server answers `content-type: application/expo+json`
(#22470 added both halves together). `loadApp:` then routes that response through the `expo-updates`
branch whenever the app has a valid updates configuration — a branch that exists for published /
EAS-Update manifests, not for a Metro server that just told us what it is.

# How

- `EXDevLauncherManifestParser` records `isDevServerContentType` when the response content type
  contains `expo+json`. Classification is unchanged: it is still a manifest URL, and the body is
  still parsed — that body is the only reliable source of the bundle URL and the transform options.
- `loadApp:` takes the lightweight `tryToParseManifest` path when that flag is set, regardless of the
  updates configuration.

iOS only. The previous revision's Android hunk is dropped — it had the same defect as the iOS one.

# Status: retracted — do not merge

Two verification runs found no measurable win and one real regression. Leaving the diff up as the
record; the details are in the thread.

- **No speed-up.** 770ms stock vs 738ms patched (warm, n=13, p≈0.10); cold, no difference. The
  `expo-updates` setup work moves between segments (HEAD→manifest 512→347ms, manifest→bundle
  1214→1381ms) rather than disappearing.
- **My own numbers below are an artifact.** I ran the two arms sequentially rather than interleaved,
  so drift over those minutes landed entirely on the first arm. The 1.57s → 1.20s median difference
  is ordering, not this patch. Left in place, struck through, rather than quietly deleted.
- **Regression.** A self-hosted expo-updates server that answers the launcher probe with
  `200 application/expo+json` is legitimate under the protocol spec (the `expo-runtime-version`
  requirement is on the client, which this probe doesn't send). This branch reclassifies it as a dev
  server, so those users get "expo-updates is not properly installed or integrated" where the stock
  build loads the update. Content type alone cannot carry the distinction.

# Test Plan

Stock SDK 57 app on an iOS 26.4 simulator, Metro on loopback, deep-link launch, timed from the
simulator's unified log — from the launch's first request to the bundle GET. 8 runs each, with
`expo-updates` configured (`updates.url` + `runtimeVersion`), which is what makes the branch
reachable:

| | runs | median |
| - | - | - |
| ~~stock `expo-dev-client@57.0.15`~~ | ~~1.800 1.469 1.952 1.669 1.835 1.160 1.261 1.268~~ | ~~1.57s~~ |
| ~~this PR~~ | ~~1.957 1.234 1.142 1.208 1.096 1.075 1.188 1.265~~ | ~~1.20s~~ |

**Retracted — sequential arms, not interleaved. See the status note above.**

Same app **without** `expo-updates`: 0.878 / 0.919 / 0.951s, and this change is a no-op there — the
branch is never entered. That's why the verification run's stock template measured ~0.25s and saw
nothing to win; the difference is app configuration, not method.

Being explicit about the size: this is ~0.35s at the median, not the "seconds" the original
description claimed. That claim was wrong and I've retracted it.

Also verified the app launches normally with the patch on the same setup (the failure mode of the
previous revision), and that the bundle URL and transform options still come from the manifest.

`EXDevLauncherManifestParserTests` is updated for the new expectations (`expo+json` → manifest URL,
`isDevServerContentType` YES; plain `application/json` → NO). **Not executed** — I have no test host
for the package's XCTest target; the behaviour above was verified on device instead.

### DIFF
--- a/packages/expo-dev-launcher/CHANGELOG.md
+++ b/packages/expo-dev-launcher/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 ### 🐛 Bug fixes
 
+- [iOS] Take the lightweight manifest path for a live Expo dev server (`application/expo+json`) instead of the `expo-updates` fetch, which exists for published manifests. ([#49250](https://github.com/expo/expo/pull/49250) by [@saileshbro](https://github.com/saileshbro))
 - [Android] Fix SIGABRT when loading a development server URL that has no path (e.g. `http://192.168.1.2:8081`): Android's `URI.resolve` omits the path separator for an empty-path base, so the first segment of a relative `bundleUrl` was spliced onto the port and the resulting authority failed to parse. ([#48625](https://github.com/expo/expo/pull/48625) by [@tsapeta](https://github.com/tsapeta))
 - [iOS] Use `RCTPlatformName` instead of hardcoding `ios` when requesting bundles from Metro. ([#46443](https://github.com/expo/expo/pull/46443) by [@gabrieldonadel](https://github.com/gabrieldonadel))
 - [Android] Fixed a crash when cold-launching a development build from a deep link that carries intent categories (e.g. an App Link opened from a browser). ([#46314](https://github.com/expo/expo/pull/46314) by [@lilianchiassai-fc](https://github.com/lilianchiassai-fc) & [#46328](https://github.com/expo/expo/pull/46328) by [@lukmccall](https://github.com/lukmccall))
--- a/packages/expo-dev-launcher/ios/EXDevLauncherController.m
+++ b/packages/expo-dev-launcher/ios/EXDevLauncherController.m
@@ -478,7 +478,9 @@ static const NSTimeInterval EXDevLauncherDefaultRequestTimeout = 10.0;
       return;
     }
 
-    if ([self->_updatesInterface isValidUpdatesConfiguration:updatesConfiguration] != YES) {
+    // A live dev server takes the lightweight manifest path even when the app's updates
+    // configuration is valid — the `expo-updates` fetch below exists for published manifests.
+    if ([self->_updatesInterface isValidUpdatesConfiguration:updatesConfiguration] != YES || manifestParser.isDevServerContentType) {
       [manifestParser tryToParseManifest:^(EXManifestsManifest *manifest) {
         if (!manifest.isUsingDeveloperTool) {
           onError([NSError errorWithDomain:@"DevelopmentClient" code:1 userInfo:@{NSLocalizedDescriptionKey: @"expo-updates is not properly installed or integrated. In order to load published projects with this development client, follow all installation and setup instructions for both the expo-dev-client and expo-updates packages."}]);
--- a/packages/expo-dev-launcher/ios/Manifest/EXDevLauncherManifestParser.h
+++ b/packages/expo-dev-launcher/ios/Manifest/EXDevLauncherManifestParser.h
@@ -10,6 +10,15 @@ typedef void (^OnManifestError)(NSError *error);
 
 @interface EXDevLauncherManifestParser : NSObject
 
+/**
+ * YES once `isManifestURLWithCompletion:` classified the response as a live Expo dev server
+ * (`Content-Type: application/expo+json`) rather than a published/EAS-Update manifest. The response
+ * is still a manifest — its body is the only reliable source of the bundle URL and the transform
+ * options — but the caller can use this to take the lightweight manifest path instead of the
+ * `expo-updates` one.
+ */
+@property (nonatomic, readonly) BOOL isDevServerContentType;
+
 - (instancetype)initWithURL:(NSURL *)url
              installationID:(NSString *)installationID
                     session:(NSURLSession *)session
--- a/packages/expo-dev-launcher/ios/Manifest/EXDevLauncherManifestParser.m
+++ b/packages/expo-dev-launcher/ios/Manifest/EXDevLauncherManifestParser.m
@@ -59,6 +59,15 @@ typedef void (^CompletionHandler)(NSData *data, NSURLResponse *response);
     }
 
     NSString *contentType = headers[@"Content-Type"];
+    // Expo CLI's manifest middleware negotiates the response content type from the Accept header
+    // this probe sends, so a live `expo start` server answers `application/expo+json`. That is a
+    // manifest — keep classifying it as one — but flag it so the caller can skip the `expo-updates`
+    // fetch path, which exists for published/EAS-Update manifests.
+    if (contentType && [contentType containsString:@"expo+json"]) {
+      _isDevServerContentType = YES;
+      completion(YES);
+      return;
+    }
     if (contentType && ![contentType hasPrefix:@"text/html"] && ![contentType containsString:@"/javascript"]) {
       completion(YES);
       return;
--- a/packages/expo-dev-launcher/ios/Tests/EXDevLauncherManifestParserTests.m
+++ b/packages/expo-dev-launcher/ios/Tests/EXDevLauncherManifestParserTests.m
@@ -79,10 +79,25 @@
     return [HTTPStubsResponse responseWithData:[NSData new] statusCode:200 headers:@{@"Content-Type": @"text/javascript"}];
   }];
   [HTTPStubs stubRequestsPassingTest:^BOOL(NSURLRequest * _Nonnull request) {
+    return [request.URL.host isEqualToString:@"ohhttpstubs"] && [request.URL.path isEqualToString:@"/expo-json1"];
+  } withStubResponse:^HTTPStubsResponse * _Nonnull(NSURLRequest * _Nonnull request) {
+    return [HTTPStubsResponse responseWithData:[NSData new] statusCode:200 headers:@{@"Content-Type": @"application/expo+json"}];
+  }];
+  [HTTPStubs stubRequestsPassingTest:^BOOL(NSURLRequest * _Nonnull request) {
+    return [request.URL.host isEqualToString:@"ohhttpstubs"] && [request.URL.path isEqualToString:@"/expo-json2"];
+  } withStubResponse:^HTTPStubsResponse * _Nonnull(NSURLRequest * _Nonnull request) {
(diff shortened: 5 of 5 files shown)


saileshbro:
Repro repo (minimal app + a 10-second curl-based repro script): https://github.com/saileshbro/repro-dev-launcher-49249 — see the issue (#49249) comment for expected output.

Local verification of this diff:
- The full upstream `EXDevLauncherManifestParser.m` compiles clean against the real pod header graph (React-Core, ExpoModulesCore, EXManifests, EXDevMenuInterface; iPhoneSimulator SDK).
- The changed Kotlin classification logic was compiled with kotlinc and executed against all 13 cases covered by the existing tests plus the new `expo+json` cases — including preserving #13825's `application/json` → manifest behavior.

saileshbro:
@alanjhughes — the verification is right on every point, and I've rewritten this PR accordingly.
Summary of what changed and what I got wrong:

**The old approach was wrong, and I hit it myself.** A day after filing this I ran the patch on my
own project and every cold launch died with `Unable to resolve module ./index` — my Metro root is a
monorepo root, so the real bundle is nested and the guessed `${baseURL}/index.bundle` doesn't exist.
The verification run found the same thing on a stock template, plus the part I hadn't traced: the
lost `transform.routerRoot` / `engine` / `bytecode` and the manifest metadata. I should have posted
that here when I found it instead of leaving the PR standing.

**My motivation was also wrong.** I wrote "adds seconds to every dev-client cold start". My own
measurements never supported it: the manifest HEAD is ~12ms, and the multi-second cost I was
actually chasing was a Bonjour/mDNS resolver timeout on a separate connection, which I
misattributed to this code path. Sorry for the cost of the run.

**What the rewritten PR does.** The manifest stays a manifest — its body is the only reliable source
of the bundle URL and the transform options, exactly as you showed. The parser just records
`isDevServerContentType` for `application/expo+json`, and `loadApp:` uses that to take the
lightweight `tryToParseManifest` path instead of the `expo-updates` fetch, which exists for
published manifests:

```objc
if ([self->_updatesInterface isValidUpdatesConfiguration:updatesConfiguration] != YES
    || manifestParser.isDevServerContentType) {
  [manifestParser tryToParseManifest:...];
```

**On the size of it — your §5 said this wasn't established in either direction, so I measured it.**
The reason your run saw nothing is that the stock template has no `expo-updates`, so
`isValidUpdatesConfiguration` returns NO and that branch is never entered at all. With
`updates.url` + `runtimeVersion` configured, 8 deep-link launches each, launch's first request →
bundle GET, stock SDK 57 app on an iOS 26.4 simulator:

```
before: 1.800 1.469 1.952 1.669 1.835 1.160 1.261 1.268   median 1.57s
after:  1.957 1.234 1.142 1.208 1.096 1.075 1.188 1.265   median 1.20s
same app without expo-updates: 0.878 0.919 0.951          (change is a no-op here)
```

So: ~0.35s at the median for updates-configured projects, nothing for everyone else. That is a much
smaller claim than the one I opened with, and if you'd rather not carry the extra flag for it, close
this — I won't argue the trade.

**The actual multi-second problem is elsewhere, and it's now separate:** #49307 / #49308 — the
launcher's `_expo._tcp` browse keeps running through a launch that already has an explicit `?url=`,
because a deep link reaches `loadApp:` without passing through `DevLauncherViewModel`. Stock repro,
before/after numbers, and the same honesty about what it does and doesn't buy are in the PR.

Both repros (this one and the Bonjour one) are in
https://github.com/saileshbro/repro-dev-launcher-49249.

The Android arm is dropped from this PR — it had the same defect, and I have no Android measurement.

brentvatne:
@expo-bot verify

saileshbro:
@brentvatne — note for the run in flight: this PR was rewritten ~11 hours ago, so it is no longer the diff the first verification tested. The `completion(NO)` hunk and the Android arm are gone; it now keeps the manifest and only flags `isDevServerContentType` so `loadApp:` skips the `expo-updates` fetch.

One thing that will otherwise look like a null result: **the branch this touches is only reachable when the app has a valid updates configuration.** A stock `create-expo-app` template has no `expo-updates`, so `isValidUpdatesConfiguration` returns NO, the launcher already takes the lightweight manifest path, and the patch is a no-op — which is exactly what the first run measured (~0.25s, nothing to win). To see anything, the test app needs `updates.url` + `runtimeVersion` in `app.json`.

The repro repo now has that app plus the harness I used: https://github.com/saileshbro/repro-dev-launcher-49249 (`measure-launch.sh`, README section "Measuring both effects"). My numbers are 8 deep-link launches each, median 1.57s → 1.20s. If the run reproduces a smaller delta than that, or none, I'd take that as the answer — it is a ~0.35s change for updates-configured projects and nothing for anyone else, and I said in the previous comment that closing this is a reasonable call.

saileshbro:
This is conclusive against the PR, and the catch about my numbers is correct — thank you for it.

**On the measurement.** You're right that my before-runs get faster in order: 1.800 1.469 1.952
1.669 1.835, then 1.160 1.261 1.268. I ran the arms sequentially — one build's 8 launches, then the
other's — so anything drifting over those minutes lands entirely on the first arm. The last three
stock runs sit right on top of my patched numbers, which is what that looks like. Interleaved arms
with a per-segment breakdown is the right method and I didn't use it; my 1.57 → 1.20 should be read
as ordering, not as the patch. Your 770 → 738 ms warm (n=13, p≈0.10) and the cold null are the
better data, and the segment table is the actual answer: the `expo-updates` setup work **moves**
(HEAD→manifest 512 → 347 ms) and comes back on the other side (manifest→bundle 1214 → 1381 ms).
Totals unchanged. That is a much better description of the code than the one I had.

**On the regression.** The lenient self-hosted updates server is disqualifying on its own, and I
hadn't considered it. A server that answers the probe `200 application/expo+json` is a legitimate
expo-updates server under the protocol spec — the spec puts the `expo-runtime-version` requirement
on the client, and the launcher probe doesn't send it — so my branch turns "loads the update" into
"expo-updates is not properly installed or integrated" for exactly the people running their own
update server. Content type alone can't carry this distinction. Adding an `expo-protocol-version: 0`
check would narrow it, but that's a second heuristic layered on the first to buy ~30 ms that isn't
statistically there, and I don't think that trade is worth making.

So: I don't think this should be merged in any form. The premise in #49249 is real — the launcher
probe does get `application/expo+json` from a live dev server — but the useful conclusion from two
verification runs is that the manifest path is the correct path and costs nothing worth reclaiming.
Happy for you to close this, and I'd suggest closing #49249 with it; if it's more useful as a record
that the JSON classification is intentional, a comment in the parsers naming the content types would
be the whole of it.

@brentvatne @alanjhughes — sorry for two verification runs on a claim that didn't survive either.

The separate Bonjour change (#49307 / #49308) is unaffected by this, and I've already walked its
timing claim back to a correctness-only one there after measuring it on a LAN: the browse and its
resolver timeout are removed, but they were concurrent with the launch, not blocking it.

PR #49251, opened 2026-08-23
[ios] Anchor generated Pods paths on ${PODS_ROOT} instead of absolute paths

janicduplessis:
# Why

`pod install` bakes machine-absolute paths into four kinds of generated output:

- the `ExpoModulesMacros` `-load-plugin-executable` flag (every Expo pod's xcconfig)
- the precompiled `ExpoModulesCore` header search paths injected into all targets
- `expo-configure-project.sh`'s `--target` / `--entitlement` / `--podfile-properties-file-path` arguments
- the `--shared` SPM dependency arguments of the xcframework replace script phase

Everything else CocoaPods generates is anchored on `${PODS_ROOT}` (10,516 references vs these 623 in a bare SDK 57 app I counted), so these are the only entries that tie a Pods directory to the checkout that ran `pod install`. A relocated Pods dir — a CI cache restored into a fresh checkout, a cloned git worktree, a moved project — silently resolves the *old* machine's paths for as long as they exist, and then fails without naming the real cause once they don't.

I hit this via git-worktree-per-task tooling that clones `ios/Pods` with APFS clonefiles instead of re-running `pod install` (~6s instead of ~26s per worktree).

# How

Anchor the four sites on `${PODS_ROOT}`:

- compile flags use the xcconfig build setting, which Xcode expands before the compiler sees the value — `-load-plugin-executable` just opens the expanded path, so a relative segment like `${PODS_ROOT}/../../node_modules/...` is fine
- generated scripts use the `$PODS_ROOT` shell variable, which the build-phase environment already provides (the same scripts already use it for `.xcode.env` and the autolinking entry point)

Paths that cannot be expressed relative to the sandbox (e.g. a pnpm store on a different volume, where `Pathname#relative_path_from` raises) keep the absolute path, matching current behavior.

# Test Plan

On a bare SDK 57 app (RN 0.86.2, precompiled modules on):

- before: 623 absolute-path references across `Pods/Target Support Files` + `Pods.xcodeproj` + generated scripts; after: **0** (`grep -r '/Volumes' ...`)
- `pod install` succeeds; full `xcodebuild` of the app workspace succeeds
- cloned the repo into a git worktree (carrying `Pods/` via clonefile, no `pod install`), built with a fresh DerivedData: succeeds, and the expanded flags now point inside the worktree instead of the original checkout

Not covered: a pnpm-style layout where `node_modules` resolves outside the repo — the fallback keeps today's absolute-path behavior there, but I only exercised the classic layout.

### DIFF
--- a/packages/expo-modules-autolinking/scripts/ios/precompiled_modules.rb
+++ b/packages/expo-modules-autolinking/scripts/ios/precompiled_modules.rb
@@ -1041,6 +1041,12 @@ module Expo
         header_search_paths = collect_xcframework_header_paths(expo_core_xcframework)
         return if header_search_paths.empty?
 
+        # Make the paths relative to PODS_ROOT so a relocated Pods directory keeps working.
+        sandbox_root = installer.sandbox.root.to_s
+        header_search_paths = header_search_paths.map do |path|
+          path.start_with?("#{sandbox_root}/") ? path.sub(sandbox_root, '${PODS_ROOT}') : path
+        end
+
         paths_string = header_search_paths.map { |p| "\"#{p}\"" }.join(' ')
 
         Pod::UI.info "#{'[Expo-precompiled] '.blue}Adding ExpoModulesJSI header search paths to all targets"
@@ -1519,7 +1525,14 @@ module Expo
           next nil unless @framework_owner_map[dep_name] == pod_name
           source_base = shared_spm_dep_source_base(dep_name, pod_info)
           next nil unless source_base
-          %(--shared "#{dep_name}:#{source_base.gsub(/[\\"$`]/) { |c| "\\#{c}" }}")
+          # Make the path relative to PODS_ROOT so a relocated Pods directory keeps working.
+          pods_root = Pod::Config.instance.project_pods_root.to_s
+          begin
+            source_base = "$PODS_ROOT/#{Pathname.new(source_base).relative_path_from(Pathname.new(pods_root))}"
+          rescue ArgumentError
+            source_base = source_base.gsub(/[\\"$`]/) { |c| "\\#{c}" }
+          end
+          %(--shared "#{dep_name}:#{source_base}")
         end
       end
 
--- a/packages/expo-modules-autolinking/scripts/ios/project_integrator.rb
+++ b/packages/expo-modules-autolinking/scripts/ios/project_integrator.rb
@@ -143,6 +143,14 @@ module Expo
           core_src_root = Expo::PrecompiledModules.package_root_for('ExpoModulesCore') ||
             File.realpath(core_pod_target.sandbox.pod_dir(core_pod_target.root_spec.name).to_s)
           macros_plugin_dir = resolve_macros_plugin_dir(core_src_root)
+          # Make the path relative to PODS_ROOT so a relocated Pods directory keeps working.
+          pods_root = Pod::Config.instance.project_pods_root.to_s
+          begin
+            relative_dir = Pathname.new(macros_plugin_dir).relative_path_from(Pathname.new(pods_root)).to_s
+            macros_plugin_dir = "${PODS_ROOT}/#{relative_dir}"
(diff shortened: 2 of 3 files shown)


packages/expo-modules-autolinking/scripts/ios/precompiled_modules.rb
```diff
@@ -1519,7 +1525,14 @@ def shared_dep_switch_args(pod_name, pod_info)
           next nil unless @framework_owner_map[dep_name] == pod_name
           source_base = shared_spm_dep_source_base(dep_name, pod_info)
           next nil unless source_base
-          %(--shared "#{dep_name}:#{source_base.gsub(/[\\"$`]/) { |c| "\\#{c}" }}")
+          # Make the path relative to PODS_ROOT so a relocated Pods directory keeps working.
+          pods_root = Pod::Config.instance.project_pods_root.to_s
+          begin
+            source_base = "$PODS_ROOT/#{Pathname.new(source_base).relative_path_from(Pathname.new(pods_root))}"
+          rescue ArgumentError
+            source_base = source_base.gsub(/[\\"$`]/) { |c| "\\#{c}" }
```
Kudo: **Medium:** Shell escaping is only applied in the `rescue ArgumentError` branch. On macOS, `relative_path_from` between two absolute same-volume paths does not raise, so the success path now emits an unescaped relative remainder into the xcframework switch bash phase.

Keep `$PODS_ROOT/` unescaped (needed for expansion), but escape the relative segment:

```ruby
begin
  rel = Pathname.new(source_base).relative_path_from(Pathname.new(pods_root)).to_s
  source_base = "$PODS_ROOT/#{rel.gsub(/[\\"$`]/) { |c| "\\#{c}" }}"
rescue ArgumentError
  source_base = source_base.gsub(/[\\"$`]/) { |c| "\\#{c}" }
end
```

packages/expo-modules-autolinking/scripts/ios/project_integrator.rb
```diff
@@ -143,6 +143,14 @@ def self.integrate_core_macro_plugins(targets)
           core_src_root = Expo::PrecompiledModules.package_root_for('ExpoModulesCore') ||
             File.realpath(core_pod_target.sandbox.pod_dir(core_pod_target.root_spec.name).to_s)
           macros_plugin_dir = resolve_macros_plugin_dir(core_src_root)
+          # Make the path relative to PODS_ROOT so a relocated Pods directory keeps working.
+          pods_root = Pod::Config.instance.project_pods_root.to_s
+          begin
+            relative_dir = Pathname.new(macros_plugin_dir).relative_path_from(Pathname.new(pods_root)).to_s
+            macros_plugin_dir = "${PODS_ROOT}/#{relative_dir}"
```
Kudo: **Low:** Same unconditional `relative_path_from` as `--shared`. Different-volume pnpm stores hit `ArgumentError` and correctly keep absolute paths; same-volume stores outside the repo get a `../` chain, which is not relocatable the way in-tree `node_modules` is. Worth aligning with a Pods/parent containment guard if you touch this again — classic layout is fine.

packages/expo-modules-autolinking/scripts/ios/precompiled_modules.rb
Kudo: **Medium:** Unlike `configure_header_search_paths` and the `relocatable` lambda in `generate_support_script`, this rewrite has no containment check. `Pathname#relative_path_from` will happily emit `$PODS_ROOT/../../…` for an `EXPO_PRECOMPILED_MODULES_PATH` (or other) source base outside the checkout.

That breaks under the exact relocation scenario this PR fixes (CI cache / worktree): the offset to an external absolute cache is not stable relative to the new Pods location. Prefer rewriting only when `source_base` is under `pods_root` or its parent (project checkout); otherwise keep the absolute, escaped path.

PR #49255, opened 2026-08-23
[expo-modules-core] Recreate shared objects when a React Activity becomes visible again

### LINKED ISSUE #47458: [expo-audio] `NativeSharedObjectNotFoundException` on navigation (stable options, component stays mounted)

https://github.com/XChikuX/expo-audio-error

## Summary

After upgrading SDK 55 (bun update) + a minor expo-audio bump (paired with React Navigation 8), `AudioPlayer` / `AudioRecorder` shared objects from `useAudioPlayer()` / `useAudioRecorder()` are released on the native side while the JS wrapper is still alive and the React component is still mounted. Any subsequent native call (`pause`, `getStatus`, property `get`) throws:

```
ERROR  [Error: FunctionCallException: Calling the '<fn>' function has failed
  (at ExpoModulesCore/SyncFunctionDefinition.swift:137)
→ Caused by: NativeSharedObjectNotFoundException: Unable to find the native shared
  object associated with given JavaScript object
  (at ExpoModulesCore/DynamicSharedObjectType.swift:58)]
```

Android variant:

```
ERROR  [Error: Call to function 'AudioRecorder.getStatus' has been rejected.
→ Caused by: The 1st argument cannot be cast to type
  expo.modules.audio.AudioRecorder (received class java.lang.Integer)
→ Caused by: Cannot use shared object that was already released]
```

## Key signal: this is native-side, not user code

1. **The user app is unchanged.** The exact same `useAudioPlayer()` / `useAudioRecorder()` calls worked before the SDK/expo-audio bump. No app code was modified.
2. **Expo Modules Core logs the error before throwing.** Wrapping every native call in a JS `try/catch` prevents the crash but does **not** suppress the red `ERROR` log — the error is recorded on the native side before the JS exception is thrown. This is consistent with the failure happening inside `SyncFunctionDefinition` / `DynamicSharedObjectType`, not in user code.
3. **Options are stable.** Recorder options are memoized (`useMemo(..., [])`) and `JSON.stringify`-serialized in `useAudioRecorder`'s deps, so the shared object should not be recreated — yet the native object disappears anyway.
4. **Component never unmounts.** The affected screen (`ProfileSelf`) sits at the bottom of a stack and is only pushed on top of. The hook still has a live JS reference when the native object is gone.
5. **Calls fail even when idle.** With active-use gating (polling `getStatus`/`currentStatus` only during recording/playback), the error still surfaces on property reads that run on focus/mount.

## Reproduction

Stack navigator with two screens. Navigate `A → B → back to A`. On return, `A`'s `AudioRecorder` is released but the JS wrapper is still referenced → next property access throws.

### Setup

```bash
npx create-expo-app@latest repro --template blank-typescript
cd repro
npx expo install expo-audio
```

`app.json` — ensure:

```json
{
  "expo": {
    "name": "repro",
    "plugins": ["expo-audio"]
  }
}
```

### `App.tsx`

```tsx
import { NavigationContainer } from '@react-navigation/native';
import { createNativeStackNavigator } from '@react-navigation/native-stack';
import ScreenA from './ScreenA';
import ScreenB from './ScreenB';

const Stack = createNativeStackNavigator();

  return (
    <NavigationContainer>
      <Stack.Navigator>
        <Stack.Screen name="A" component={ScreenA} />
        <Stack.Screen name="B" component={ScreenB} />
      </Stack.Navigator>
    </NavigationContainer>
  );
}
```

```bash
npm install @react-navigation/native @react-navigation/native-stack react-native-screens react-native-safe-area-context
```

### `ScreenA.tsx`

```tsx
import { useEffect, useMemo } from 'react';
import { Pressable, Text, View } from 'react-native';
import { useNavigation } from '@react-navigation/native';
import { useAudioRecorder, RecordingPresets } from 'expo-audio';

export default function ScreenA() {
  const nav = useNavigation();

  // Stable options — recorder should NOT be recreated.
  const options = useMemo(
    () => ({ ...RecordingPresets.HIGH_QUALITY, isMeteringEnabled: true }),
    [],
  );
  const recorder = useAudioRecorder(options);

  // Fires on mount and on every focus. Reads a native property.
  useEffect(() => {
    // eslint-disable-next-line no-console
    console.log('recorder.uri =', recorder.uri); // ← throws on re-focus
  }, []);

  return (
    <View style={{ flex: 1, alignItems: 'center', justifyContent: 'center' }}>
      <Pressable onPress={() => nav.navigate('B' as never)}>
        <Text>Go to B</Text>
      </Pressable>
    </View>
  );
}
```

### `ScreenB.tsx`

```tsx
import { Pressable, Text, View } from 'react-native';
import { useNavigation } from '@react-navigation/native';

export default function ScreenB() {
  const nav = useNavigation();
  return (
    <View style={{ flex: 1, alignItems: 'center', justifyContent: 'center' }}>
      <Pressable onPress={() => nav.goBack()}>
        <Text>Back to A</Text>
      </Pressable>
    </View>
  );
}
```

### Trigger

1. Launch → app opens on ScreenA (recorder created).
2. Tap "Go to B".
3. Tap "Back to A".
4. Observe `NativeSharedObjectNotFoundException` (iOS) / `Cannot use shared object that was already released` (Android).

### Environment

```text
expo-audio: ~55.0.15
expo-modules-core: SDK 55
react-native: 0.83.x
React Navigation: 8.0.0-alpha.x
```

### Additional Details

```text
## Workaround (user-side, does not fix the root cause)

- Poll `getStatus` / `currentStatus` only while actively recording/playing (gate on `isRecording` / `isPlaying`).
- Wrap every native call and property read in try/catch. **This prevents crashes but does not silence the native `ERROR` log** — the framework records the failure natively before throwing to JS.
- Avoid reading native properties (`recorder.uri`, `player.currentStatus`) outside of user-triggered actions.

These measures reduce the surface area but cannot fully suppress the errors because the native object is gone before any JS guard can run.

## Related

- #39427 — same `NativeSharedObjectNotFoundException` trace, was closed in #39753 ("Fix SharedObject created with `useReleasingSharedObject` getting destroyed after a fast refresh"). The regression here occurs during normal navigation, not fast refresh.
- #33244 — identical Android trace ("received class java.lang.Integer … Cannot use shared object that was already released").
```

LizunovSergey:
# Why

Fixes #47458.

`useReleasingSharedObject` hands back a released shared object after a React Activity is hidden and
shown again, so the next call on it throws `NativeSharedObjectNotFoundException`. The reporter hit it
with `expo-audio` on navigation, with stable options and the component still mounted — which is
exactly the case `<Activity>` is built for, since React keeps state and refs across the hidden phase
and only tears down effects.

# How

Two changes in `useReleasingSharedObjectWithLifecycle`, which `useReleasingSharedObject` delegates to:

1. **The unmount effect now clears `objectRef.current` after releasing.** Hiding an Activity runs
   effect cleanup, so the object is released, but the ref kept pointing at it. With the ref cleared,
   the `if (objectRef.current == null)` guard at the top of the hook creates a fresh object when the
   Activity becomes visible again.

2. **The object resolution moved from `useMemo(..., dependencies)` to an IIFE.** Coming back from
   hidden, the dependencies are unchanged, so `useMemo` would replay its cached value — the released
   object — regardless of what the ref now holds. The body is unchanged and stays idempotent: when
   the dependencies compare equal it only reads `objectRef.current` and returns it, no factory call
   and no ref writes. The mutating branch is still gated on `dependenciesAreEqual` and still updates
   `previousDependencies.current`, so a re-render with the same dependencies does not repeat it.

Fast refresh is unaffected — it is guarded by `isFastRefresh`, which is set before the cleanup runs
and is what keeps the object alive there.

# Test Plan

Added `packages/expo-modules-core/src/__tests__/useReleasingSharedObject-test.native.tsx`, which
drives a real `<Activity>` through visible → hidden → visible → unmount and asserts the full
lifecycle:

- the first object is released once when the Activity is hidden,
- a second object is created when it becomes visible again and is the one handed to the consumer,
- that second object is not released while visible,
- and it is released exactly once on unmount.

Against `main` the test fails on the third assertion — the consumer keeps receiving the released
first object and the factory is called once, not twice.

# Checklist


LizunovSergey:
Rebased onto current `main`. Two conflicts: `CHANGELOG.md` (new upstream entries in the same bug-fixes section; kept both sides) and `useReleasingSharedObjectWithLifecycle.ts`, where #49242 had added an `oxlint-disable-next-line react/use-memo` suppression to the `useMemo` this PR replaces with a plain render-phase computation — the suppression is moot for the IIFE, so it goes away with the `useMemo`. Behaviour is unchanged; the package's oxlint, `tsc` typecheck, and the hook's jest tests all pass after the rebase.

Issue #47469, opened 2026-07-02
[dom-components] Uncaught promise rejection "Call to function 'DomWebView.injectJavaScript' has been rejected" when the $$props emit races native view mount on Android


Dustin-Kelley:
# Minimal reproducible example

Any `'use dom'` component mounted at app cold start on Android with the default `@expo/dom-webview` backing reproduces it — no special props needed:

```tsx
// components/hello-dom.tsx
'use dom';
export default function HelloDom() {
  return <p>hello</p>;
}

// app/index.tsx
import HelloDom from '../components/hello-dom';
export default function Index() {
  return <HelloDom />;
}
```

Cold start the Android dev build and watch `adb logcat`. One uncaught rejection is logged per DOM component mounted during startup (our app pre-warms six DOM WebViews and gets six LogBox toasts):

```
E ReactNativeJS: [Error: Uncaught (in promise, id: 0) Error: Call to function 'DomWebView.injectJavaScript' has been rejected.
E ReactNativeJS: → Caused by: Unable to find the class expo.modules.webview.DomWebView view with tag 134]
```

# What happened?

On Android, mounting a DOM component during app cold start logs an uncaught promise rejection (surfaced as a LogBox toast in dev):

```
Error: Call to function 'DomWebView.injectJavaScript' has been rejected.
→ Caused by: Unable to find the class expo.modules.webview.DomWebView view with tag <N>
```

## Root cause

`packages/expo/src/dom/webview-wrapper.tsx` emits the marshalled props to the WebView on every render, including the very first one after mount:

```tsx
const emit = React.useCallback(
  (detail: BridgeMessage<any>) => {
    webviewRef.current?.injectJavaScript(getInjectEventScript(detail));
  },
  [webviewRef]
);

// When the `marshalProps` change, emit them to the webview.
React.useEffect(() => {
  emit({ type: '$$props', data: smartActions });
}, [emit, smartActions]);
```

`smartActions` is a new object each render, so this effect fires on the first commit. With `@expo/dom-webview`, `injectJavaScript` is an expo-modules **view AsyncFunction** (`DomWebViewModule.kt`). Its view argument is resolved on the native side by `ViewTypeConverter` → `AppContext.findView(viewTag)`, and at first-commit time the Fabric view is not yet resolvable by tag, so the converter throws `Exceptions.ViewNotFound`, the call rejects, and nothing catches the returned promise — hence the uncaught rejection.

[...]

PR #49257, opened 2026-08-23
[expo] Don't emit DOM component props before the DOM side is ready

### LINKED ISSUE #49173: [dom-components] Android cold start uncaught injectJavaScript reject (cloneable MRE)

https://github.com/cameronapak/expo-use-dom-android-cold-start

### Steps to reproduce

Cloneable repo (required: expo-bot closed the prior report for inline TSX): https://github.com/cameronapak/expo-use-dom-android-cold-start

Prior incomplete report: https://github.com/expo/expo/issues/47469

1. Clone that repo, install deps, create an Android development build, and cold-start the start screen.

2. The start screen mounts one `use dom` hello component. The app never calls injectJavaScript.

3. Watch LogBox and/or adb logcat on Android.

4. Expected on Android: one uncaught reject `DomWebView.injectJavaScript` / view tag not found.

5. iOS is clean on the same start screen.

## What happens

On Android, mounting a DOM component during app cold start logs an uncaught promise rejection (LogBox in dev):

Error: Call to function DomWebView.injectJavaScript has been rejected.
Caused by: Unable to find the class expo.modules.webview.DomWebView view with tag N

## Root cause (same analysis as 47469)

packages/expo/src/dom/webview-wrapper.tsx emits marshalled props on every render, including the first commit after mount. smartActions is a new object each render, so the effect fires immediately. With @expo/dom-webview, injectJavaScript is a view AsyncFunction. The view argument is resolved by tag via AppContext.findView. At first-commit time the Fabric view is not resolvable yet, so the call rejects. Nothing catches the promise.

Verified 2026-08-20: emit on expo/expo main still has no .catch.

iOS and react-native-webview do not hit this race the same way.

The rejected first $$props emit is usually redundant: initial props go through injectedJavaScriptObject.initialProps, and later emits resend the full object. The rejection is noise more than a new data-loss path. A silent drop already exists for emits that land after the view exists but before the page has loaded.

## Suggested directions (not a demand)

Minimal: attach a .catch on emit so a first-commit view-tag miss does not surface as an uncaught reject. That does not fix the silent pre-load drop window.

Better: a DOM-to-native ready handshake that buffers the latest $$props until the DOM side is ready, then flushes. The DOM side already messages native for REGISTER_DOM_IMPERATIVE_HANDLE_PROPS.

### Environment

Pinned in the MRE:

- expo: 57.0.15
- @expo/dom-webview: 57.0.1
- expo-router: 57.0.15
- react-native: 0.86.2
- React Native new architecture (SDK 57 default)
- Android cold start of a development build
- iOS not affected

### Expo Doctor Diagnostics

21/21 checks passed. No issues detected. (expo-doctor 1.20.2)

ahmdshrif:
# Why

Fixes #49173.

On Android, mounting a DOM component during app cold start logs an uncaught promise rejection:

```
Error: Call to function DomWebView.injectJavaScript has been rejected.
Caused by: Unable to find the class expo.modules.webview.DomWebView view with tag N
```

The app never calls `injectJavaScript` itself. It comes from the `$$props` emit effect in
[`packages/expo/src/dom/webview-wrapper.tsx`](https://github.com/expo/expo/blob/main/packages/expo/src/dom/webview-wrapper.tsx),
which runs on the **first commit after mount**: `smartActions` is rebuilt on every render, so the
effect fires immediately. With `@expo/dom-webview`, `injectJavaScript` is a view `AsyncFunction`
whose view argument is resolved by tag through `AppContext.findView`; at first-commit time the
Fabric view is not resolvable yet, so the call rejects. Nothing catches the returned promise, so it
surfaces as an uncaught rejection (LogBox in dev). iOS does not hit this race the same way.

The emit is also pointless at that moment. `dom-entry.tsx` registers the `$$props` listener and
only then posts `$$dom_ready` — with the comment *"Must stay after `addEventListener`: props
emitted before it existed were dropped."* So anything emitted before `$$dom_ready` is dropped on
the DOM side regardless, and #48813 already added the `$$dom_ready` handler that answers with the
current props.

Credit to @cameronapak for the report and the root-cause analysis, including the earlier #47469.

# How

Skip the `$$props` emit until the DOM side has signalled `$$dom_ready`. The rejecting call is no
longer made at all, rather than made and then ignored, and no error is swallowed:

- a `isDOMReadyRef` ref, set in the existing `DOM_READY` branch of `onMessage`;
- the props effect returns early while it is `false`.

Props that change before `$$dom_ready` are not lost: the `$$dom_ready` handler emits `smartActions`
from the current render, so it always answers with the latest values. This is the "ready handshake"
direction from the issue, without a buffer — the handshake added in #48813 already serves as one.

Deliberately **not** `.catch(() => {})` on `emit`: that would also silence delivery failures of
`$$native_action_result`, which are genuine errors, and would leave the useless pre-ready
`injectJavaScript` call in place.

Behaviour after `$$dom_ready` is unchanged, including WebView reloads: `dom-entry` remounts and
posts `$$dom_ready` again, and the handler answers as before.

The one way this could regress is a page that registers a `$$props` listener but never posts
`$$dom_ready`. That cannot happen for pages built by the DOM runtime — `dom-entry.tsx` does both in
the same effect, listener first — and a page that does neither (for example a custom `overrideUri`)
already dropped these emits.

# Test Plan

New `packages/expo/src/dom/__tests__/webview-props-emit-test.web.ts`, mounting `RawWebView` for real
so its effects run. `packages/expo` has no React Native renderer in its devDependencies, but
`react-dom` is already there and the Web project runs in jsdom, so the wrapper can be mounted with
`react-dom/client` and `act`. The file is `.web.ts` and guards on `document`, since the Node project
picks `.web` tests up too.

Four tests, run against `packages/expo`:

| | before | after |
| --- | --- | --- |
| `injectJavaScript` calls after mount | 1 | 0 |
| after a prop update before `$$dom_ready` | 2 | 0 |
| after `$$dom_ready` | 3 | 1, carrying `updated-before-ready` |
| after a prop update following `$$dom_ready` | 3 | 2, carrying `updated-after-ready` |

With only `webview-wrapper.tsx` reverted, all four fail; with the change, all four pass.

Package checks:

- `jest` (`packages/expo`): **68 suites / 576 passed, 8 skipped**, versus **66 suites / 572 passed,
  4 skipped** on a clean checkout of the same base — the delta is exactly the four new tests
  (passing in the Web project, skipped in the Node project). No churn elsewhere.
- `tsc -p tsconfig.json`: no new errors attributable to these files.
- `oxlint --config oxlint.config.mjs .`: clean.
- Formatting matches the surrounding file (the pre-existing trailing-comma differences in
  `webview-wrapper.tsx` are untouched).

Not verified on a device: I do not have the Android development build from the MRE
(`cameronapak/expo-use-dom-android-cold-start`) available here, so the disappearance of the LogBox
entry itself is argued from the mechanism — the rejecting call is no longer made — rather than
observed. Note also that jest's environment suppresses `process.on('unhandledRejection')`, so the
rejection cannot be asserted in-suite; the tests assert the call is not made, which is the cause.

# Checklist

- [ ] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: updated a module plugin) — not relevant, JS-only change with no native or config surface.
- [ ] Conforms with the [Documentation Writing Style Guide](https://github.com/expo/expo/blob/main/guides/Expo%20Documentation%20Writing%20Style%20Guide.md) — no docs changed.

### DIFF
--- a/packages/expo/CHANGELOG.md
+++ b/packages/expo/CHANGELOG.md
@@ -23,7 +23,7 @@
 - Fix async imports (`import(...)`) via `asyncRequireModule` not a thenable instead of a full promise shape ([#48550](https://github.com/expo/expo/pull/48550) by [@kitten](https://github.com/kitten))
 - Fix `window.location` being called regardless of `@expo/metro-runtime` being present on native when an async chunk loads after Metro disconnects ([#48944](https://github.com/expo/expo/pull/48944) by [@expo-bot](https://github.com/expo-bot))
 - Fix DOM components dropping prop updates that are emitted while the WebView is still loading. ([#48813](https://github.com/expo/expo/pull/48813) by [@expo-bot](https://github.com/expo-bot))
-- [Android] Fix DOM components logging an uncaught `DomWebView.injectJavaScript` rejection on cold start by not emitting props before the DOM side is ready. ([#49173](https://github.com/expo/expo/issues/49173) by [@ahmdshrif](https://github.com/ahmdshrif))
+- [Android] Fix DOM components logging an uncaught `DomWebView.injectJavaScript` rejection on cold start by not emitting props before the DOM side is ready. ([#49257](https://github.com/expo/expo/pull/49257) by [@ahmdshrif](https://github.com/ahmdshrif))
 - Fix `import.meta.url` being `null` on web when `transform.inlineRequires` is enabled. ([#49045](https://github.com/expo/expo/pull/49045) by [@expo-bot](https://github.com/expo-bot))
 
 ### 💡 Others
--- a/packages/expo/CHANGELOG.md
+++ b/packages/expo/CHANGELOG.md
@@ -23,6 +23,7 @@
 - Fix async imports (`import(...)`) via `asyncRequireModule` not a thenable instead of a full promise shape ([#48550](https://github.com/expo/expo/pull/48550) by [@kitten](https://github.com/kitten))
 - Fix `window.location` being called regardless of `@expo/metro-runtime` being present on native when an async chunk loads after Metro disconnects ([#48944](https://github.com/expo/expo/pull/48944) by [@expo-bot](https://github.com/expo-bot))
 - Fix DOM components dropping prop updates that are emitted while the WebView is still loading. ([#48813](https://github.com/expo/expo/pull/48813) by [@expo-bot](https://github.com/expo-bot))
+- [Android] Fix DOM components logging an uncaught `DomWebView.injectJavaScript` rejection on cold start by not emitting props before the DOM side is ready. ([#49173](https://github.com/expo/expo/issues/49173) by [@ahmdshrif](https://github.com/ahmdshrif))
 - Fix `import.meta.url` being `null` on web when `transform.inlineRequires` is enabled. ([#49045](https://github.com/expo/expo/pull/49045) by [@expo-bot](https://github.com/expo-bot))
 
 ### 💡 Others
--- a/packages/expo/src/dom/__tests__/webview-props-emit-test.web.ts
+++ b/packages/expo/src/dom/__tests__/webview-props-emit-test.web.ts
@@ -67,8 +67,6 @@ describeWithDOM('$$props emit', () => {
 
     await render({ title: 'initial' });
 
-    // The DOM side has no `$$props` listener yet, and on the first commit the
-    // native view is not resolvable by tag, so this rejected instead.
     expect(injectJavaScript).not.toHaveBeenCalled();
   });
 
--- a/packages/expo/src/dom/__tests__/webview-props-emit-test.web.ts
+++ b/packages/expo/src/dom/__tests__/webview-props-emit-test.web.ts
@@ -1,0 +1,111 @@
+// `.web.ts` on purpose: mounting the wrapper for real (so its effects run) needs
+// `react-dom/client`, and the Web project is the only one with a DOM. The Node
+// project picks `.web` tests up as well, hence the guard below.
+import type { BridgeMessage } from '../dom.types';
+
+const describeWithDOM = typeof document !== 'undefined' ? describe : describe.skip;
+
+describeWithDOM('$$props emit', () => {
+  beforeEach(() => {
+    jest.resetModules();
+    jest.doMock('../../utils/getDevServer', () => ({
+      __esModule: true,
+      default: () => ({ url: 'http://localhost:8081' }),
+    }));
+  });
+
+  function mountRawWebView() {
+    // Mimics `@expo/dom-webview`, where `injectJavaScript` is a view `AsyncFunction`
+    // that rejects while the native view cannot be resolved by tag yet.
+    const injectJavaScript = jest.fn().mockImplementation(() => Promise.resolve());
+    let webViewProps: Record<string, any> = {};
+    jest.doMock('@expo/dom-webview', () => ({
+      WebView: (stubProps: any) => {
+        webViewProps = stubProps;
+        // `ref` arrives as a plain prop, and the wrapper injects scripts through it.
+        stubProps.ref.current = { injectJavaScript };
+        return null;
+      },
+    }));
+    (globalThis as any).IS_REACT_ACT_ENVIRONMENT = true;
+    const React = require('react');
+    const { act } = React;
+    const { createRoot } = require('react-dom/client');
+    const RawWebView = require('../webview-wrapper').default;
+
+    const container = document.createElement('div');
+    document.body.appendChild(container);
+    const root = createRoot(container);
+
+    const render = async (props: Record<string, unknown>) => {
+      await act(async () => {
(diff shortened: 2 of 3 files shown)


packages/expo/src/dom/webview-wrapper.tsx
```diff
@@ -110,6 +112,14 @@ const RawWebView = React.forwardRef<object, Props>((props, ref) => {
 
   // When the `marshalProps` change, emit them to the webview.
   React.useEffect(() => {
+    // The DOM side only registers its `$$props` listener right before it posts
```
Kudo: **[suggestion]** The new comments restate the handshake and the Android first-commit reject in five lines, plus a WHAT comment on `isDOMReadyRef`. That is design rationale the PR already documents. The `if (!isDOMReadyRef.current)` guard is clear next to the `DOM_READY` handler.

**Suggestion:** Drop the `isDOMReadyRef` comment and keep at most one short WHY on the guard, for example `// Skip until $$dom_ready; earlier injects are dropped and reject on Android first commit.`

packages/expo/src/dom/__tests__/webview-props-emit-test.web.ts
```diff
@@ -0,0 +1,111 @@
...
+    }
+    return JSON.parse(payload).detail;
+  }
+
+  it('should not inject anything before the DOM side is ready', async () => {
+    const { injectJavaScript, render } = mountRawWebView();
+
+    await render({ title: 'initial' });
+
+    // The DOM side has no `$$props` listener yet, and on the first commit the
```
Kudo: **[suggestion]** The comment above `expect(injectJavaScript).not.toHaveBeenCalled()` retells the bug (`no listener yet`, `rejected instead`). The test name already says props are not injected before the DOM side is ready.

**Suggestion:** Delete those two comment lines.

packages/expo/CHANGELOG.md
```diff
@@ -23,6 +23,7 @@
 - Fix async imports (`import(...)`) via `asyncRequireModule` not a thenable instead of a full promise shape ([#48550](https://github.com/expo/expo/pull/48550) by [@kitten](https://github.com/kitten))
 - Fix `window.location` being called regardless of `@expo/metro-runtime` being present on native when an async chunk loads after Metro disconnects ([#48944](https://github.com/expo/expo/pull/48944) by [@expo-bot](https://github.com/expo-bot))
 - Fix DOM components dropping prop updates that are emitted while the WebView is still loading. ([#48813](https://github.com/expo/expo/pull/48813) by [@expo-bot](https://github.com/expo-bot))
+- [Android] Fix DOM components logging an uncaught `DomWebView.injectJavaScript` rejection on cold start by not emitting props before the DOM side is ready. ([#49173](https://github.com/expo/expo/issues/49173) by [@ahmdshrif](https://github.com/ahmdshrif))
```
Kudo: **[nit]** The entry links issue `#49173` instead of this PR. Nearby unpublished entries use PR links, and the changelog guide’s publish script walks PRs referenced in the changelog to comment on closed issues.

**Suggestion:** Point the link at `[#49257](https://github.com/expo/expo/pull/49257)`.

Kudo (changes requested):
would be good to reduce code comment a little bit

ahmdshrif:
Thanks @Kudo — all three applied in de67a30.

- `webview-wrapper.tsx`: dropped the `isDOMReadyRef` comment and collapsed the five-line block to your one-liner.
- `webview-props-emit-test.web.ts`: removed the two comment lines.
- `CHANGELOG.md`: the entry now links #49257 instead of the issue.

Verified locally after the edits: `jest src/dom` → 11 suites / 46 passed, `oxfmt --check` clean.

On the `dom.overrideUri` case the verify run raised — the measurement looks right to me, and I went looking for a fallback that isn't a timer. There isn't a good one:

- `onLoadEnd` would be the natural signal, but it only exists on `react-native-webview`. `@expo/dom-webview`'s `DomWebView.types.ts` exposes `onMessage`, `onContentProcessDidTerminate` and `onRenderProcessGone`, and no load event. The Android `injectJavaScript` rejection this PR fixes is on the `DomWebView` path, so a load-event gate would not cover the case the PR is about.
- A timeout would apply to every mount and would re-introduce exactly the race the gate removes: emitting before the DOM side has registered its `$$props` listener. It trades a narrow, deterministic gap for a nondeterministic one everywhere.

So my preference is to accept it: the skew needs `dom.overrideUri` pointing at a bundle built by `expo@57.0.0`–`57.0.13`, since a bundle built by the same install always carries `notifyDOMReady()`. If you'd rather not accept it, say which shape you want and I'll add it here with a test — I just don't want to pick a timer on your behalf.

Also noting the third point from that run so it isn't lost: the imperative-handle proxy calls `injectJavaScript` outside `emit`, so an early ref call can still reject. That's a different path from the `$$props` effect this PR gates. Happy to open an issue for it if you want it tracked separately.

Issue #27421, opened 2024-03-04
Unable to find attached android devices when ADB_TRACE=adb is set


iongion:
### Summary

Short story - `getAttachedDevicesAsync` from `node_modules/@expo/cli/build/src/start/platforms/android/adb.js` is using `adb devices -l` to find the devices, but when `ADB_TRACE=adb` is set, it is unable to extract properly due to presence of debug log lines.

The `const type = line.includes("emulator") ? "emulator" : "device";` is not sufficient to detect if a line is an `emulator` or `device` when logs are present. Example log here under has lines that don't include `emulator` and clearly are not not device.

```
adb devices
03-04 12:29:44.557 16415 16415 D adb     : adb_trace.cpp:187 Android Debug Bridge version 1.0.41
03-04 12:29:44.557 16415 16415 D adb     : adb_trace.cpp:187 Version 35.0.0-11411520
03-04 12:29:44.557 16415 16415 D adb     : adb_trace.cpp:187 Installed as /home/user/SDKs/Android/platform-tools/adb
03-04 12:29:44.557 16415 16415 D adb     : adb_trace.cpp:187 Running on Linux 5.15.133.1-microsoft-standard-WSL2 (x86_64)
03-04 12:29:44.557 16415 16415 D adb     : adb_trace.cpp:187
03-04 12:29:44.557 16415 16415 D adb     : commandline.cpp:1646 Using server socket: tcp:172.27.192.1:5037
03-04 12:29:44.557 16415 16415 D adb     : adb_client.cpp:160 _adb_connect: host:version
03-04 12:29:44.558 16415 16415 D adb     : adb_client.cpp:194 _adb_connect: return fd 3
List of devices attached
03-04 12:29:44.558 16415 16415 D adb     : adb_client.cpp:400 adb_query: host:devices
03-04 12:29:44.558 16415 16415 D adb     : adb_client.cpp:358 adb_connect: service: host:devices
03-04 12:29:44.558 16415 16415 D adb     : adb_client.cpp:160 _adb_connect: host:devices
03-04 12:29:44.558 16415 16415 D adb     : adb_client.cpp:194 _adb_connect: return fd 3
03-04 12:29:44.558 16415 16415 D adb     : adb_client.cpp:376 adb_connect: return fd 3
emulator-5554   device
```

### What platform(s) does this occur on?

Android

### SDK Version

50.0.8

### Environment

  expo-env-info 1.2.0 environment info:
    System:
[...]

PR #49258, opened 2026-08-23
refactor(cli): Improve avd/adb reliability with bounded and cancellable operations

### LINKED ISSUE #19864: expo run:android -d can not find attached devices
### Summary

I have my device attached through USB but `run:android -d` can not find it.

Screenshot 2022-11-04 at 14 44 55

Android

### SDK Version

45.0.0

### Environment

```
expo-env-info 1.0.5 environment info:
    System:
      OS: macOS 12.5.1
      Shell: 5.8.1 - /bin/zsh
    Binaries:
      Node: 16.14.2 - ~/.nodenv/versions/16.14.2/bin/node
      Yarn: 1.22.15 - /usr/local/bin/yarn
      npm: 8.5.0 - ~/.nodenv/versions/16.14.2/bin/npm
      Watchman: 2022.09.19.00 - /usr/local/bin/watchman
    Managers:
      CocoaPods: 1.11.3 - /usr/local/bin/pod
    SDKs:
      iOS SDK:
        Platforms: DriverKit 22.1, iOS 16.1, macOS 13.0, tvOS 16.1, watchOS 9.1
    IDEs:
      Android Studio: 2021.1 AI-211.7628.21.2111.8193401
      Xcode: 14.1/14B47b - /usr/bin/xcodebuild
    npmPackages:
      expo: ^45.0.0 => 45.0.6 
      react: 17.0.2 => 17.0.2 
      react-dom: 17.0.2 => 17.0.2 
      react-native: 0.68.2 => 0.68.2 
      react-native-web: 0.17.7 => 0.17.7 
      react-navigation: ^4.3.9 => 4.4.4 
    npmGlobalPackages:
      eas-cli: 2.2.1
      expo-cli: 6.0.8
    Expo Workflow: bare
```

- List device with `adb devices`
- Run `npx expo run:android -d`
- Device isn't listed

### LINKED ISSUE #33301: Expo CLI can't detect Android devices/emulators despite adb working
### Summary

When trying to launch Expo on Android, I receive the following error:
```

expo:start:server:devServer Creating platform manager (platform: emulator, port: 8081) +11s
expo:start:platforms:android:adbServer C:\custom_sdk\Sdk/platform-tools/adb devices -l +0ms
expo:start:platforms:platformManager open (runtime: expo, platform: android, device: undefined, shouldPrompt: undefined) +0ms
expo:start:platforms:android:adbServer C:\custom_sdk\Sdk/platform-tools/adb devices -l +373ms
CommandError: No Android connected device found, and no emulators could be started automatically.
Please connect a device or create an emulator (https://docs.expo.dev/workflow/android-studio-emulator).
Then follow the instructions here to enable USB debugging:

```

Despite having the correct environment variables set up, when I run the command C:\custom_sdk\Sdk/platform-tools/adb devices -l in a separate command prompt, my devices show up without issue, but Expo CLI is unable to detect them. I've also tried changing the SDK path, but the problem persists.

Additional context:

1. I have checked that the Android SDK path and adb are correctly configured.
2. Devices are visible when running the adb devices -l command directly in a command prompt, but Expo is unable to detect them.
3. I've also tried updating the Android SDK path, but the issue persists.

Android

### SDK Version

52.0.11

### Environment

```text
expo-env-info 1.2.1 environment info:
    System:
      OS: Windows 10 10.0.19045
    Binaries:
      Node: 22.11.0 - C:\Program Files\nodejs\node.EXE
      npm: 10.9.0 - C:\Program Files\nodejs\npm.CMD
    IDEs:
      Android Studio: AI-241.18034.62.2412.12266719
    npmPackages:
      expo: ~52.0.11 => 52.0.11
      expo-router: ~4.0.9 => 4.0.9
      react: 18.3.1 => 18.3.1
      react-dom: 18.3.1 => 18.3.1
      react-native: 0.76.3 => 0.76.3
      react-native-web: ~0.19.13 => 0.19.13
    Expo Workflow: managed
```

Steps to reproduce:

Run expo start or npx expo start on a Windows machine with a valid Android emulator or connected Android device.
See the error message: No Android connected device found, and no emulators could be started automatically.
Check the adb devices list using C:\custom_sdk\Sdk/platform-tools/adb devices -l in a separate command prompt and observe that the devices are shown, but Expo cannot detect them.
I have tested this issue against the latest CLI version, but the problem persists.

### LINKED ISSUE #48744: `expo run:android` hangs forever with zero output when the adb server is wedged - no timeout on `adb start-server` / `adb devices -l`

kitten:
# Why

Resolves #48744
Resolves #33301
Resolves #19864
Related #30401
Related #27421

Previously, `adb` discovery and subprocess handling was "unclean" and could lead to hanging `expo` processes, since some sub-tasks weren't bound. There's several conditions that can lead `adb` to hang or not respond properly. This is further complicated by us "pretending" to own the `start-server` / `kill-server` lifecycle.

Further, the `kill-server` invocation and lack of cancellation can lead to cases where we either fail our own adb server lifecycle assumption, or kill a server that's been used or adopted by another process (e.g. Android Studio)

# How

- Add cancellation signals from top to bottom to all `adb` operations
- Remove `start-server` and `kill-server` lifecycle calls
  - **Note:** This wasn't consistent and ignores that adb daemons should be shared. Stopping it again isn't necessary and there's no design in `adb` to facilitate this cleanly.
- Timeouts are scoped depending on the operation
- Allow devices to be parsed without AVD being available
- Improve emulator polling and name lookups
- Stop device being identified by `name` instead of `serial`, and use name only as a fallback
- Improve error handling and make messages more actionable
- Fix races and loops that ignore prior command hanging/overlapping

Exact cases of the lack of cancellation this should address are:
- `adb start-server` or `adb devices -l` could wait forever
- An already started adb server could be unhealthy, never answering requests, while we'd wait for responses forever
- Slow or stalled list/discovery requests could be invisible with no actionable error message

The `kill-server` lifecycle could previously cause:
- `adb start-server` may have been run with a server already being present
- CLI shutdown may call `adb kill-server` disrupting other processes, e.g. Android Studio
- Both can hang, causing stalls without timeouts

Timeouts on this branch are a bit tricky, since:
- some operations are shorter than others that may be long-living
- elapsed time alone isn't always indicative of failure (e.g. installs)
- a side-effect operation isn't universally retryable

Also, some operations were previously using `spawnSync` or blocking calls, which made the above refactors hard/impossible. Most of these were replaced with `@expo/spawn-async` to avoid duplicate wrapper code.

Lastly, with parsing and error handling:
- Some output handling mixed `stderr` into `stdout` before parsing, which could fail parsing unexpectedly
- Some error messages were generic and not helpful, although we had context on what went wrong (or could trivially track it)
- Device state parsing was lossy and throwing away valid information that we then either didn't have or re-derived later

Changes are validated and partially authored by LLMs, but reviewed/edited manually. For provenance/validation, the documents produced during this refactor (all iterated on as the branch evolved) are:
- [ADB_FINDINGS.md](https://github.com/user-attachments/files/31349033/ADB_FINDINGS.md)
- [ADB_PLAN.md](https://github.com/user-attachments/files/31349036/ADB_PLAN.md)
- [ADB_OUTCOME.md](https://github.com/user-attachments/files/31349046/ADB_OUTCOME.md)

# Test Plan

- [x] Check on macOS against an emulator (stopped or started)
- [x] Check on macOS against a real device
- [x] Check ctrl+c during discovery, install, or emulator startup
- [x] Check on Windows against a real device (run twice)
- [x] Check on Windows against a real device, disconnecting it during install from USB
- [x] Check a suspended/unresponsive ADB server on Windows

**Note:** We can confirm backporting this to previous SDK versions once the above is validated. We may merge before some of this has been tested, if the changes remain on `main` for now.

# Checklist

### DIFF
--- a/packages/@expo/cli/src/start/platforms/android/ADBServer.ts
+++ b/packages/@expo/cli/src/start/platforms/android/ADBServer.ts
@@ -6,14 +6,15 @@ import { assertSdkRoot } from './AndroidSdk';
 import {
   AdbProcessWaitError,
   runAdbDeviceMutationAsync,
-  runAdbDeviceQueryAsync,
   runAdbHostQueryAsync,
   runBoundedAdbDeviceQueryAsync,
+  runBoundedAdbDeviceMutationAsync,
   runBoundedAdbHostQueryAsync,
 } from './adbProcess';
 
 const BEGINNING_OF_ADB_ERROR_MESSAGE = 'error: ';
 const PROPERTY_QUERY_WAIT_LIMIT_MS = 10_000;
+const DEVICE_QUERY_WAIT_LIMIT_MS = 10_000;
 
 export class ADBServer {
   /** Returns the command line reference to ADB. */
@@ -35,7 +36,8 @@ export class ADBServer {
   async runDeviceQueryAsync(
     args: string[],
     operation: string,
-    signal?: AbortSignal
+    signal?: AbortSignal,
+    waitLimitMs: number = DEVICE_QUERY_WAIT_LIMIT_MS
   ): Promise<string> {
     const adb = this.getAdbExecutablePath();
 
@@ -44,22 +46,27 @@ export class ADBServer {
 
     event('adb_server_run', { command: [adb, ...args].join(' ') });
     const result = await this.resolveAdbPromise(
-      runAdbDeviceQueryAsync(adb, args, operation, signal)
+      runBoundedAdbDeviceQueryAsync(adb, args, operation, waitLimitMs, signal)
     );
-    return [result.stdout, result.stderr].join('\n');
+    assertValidAdbUserOutput(result);
+    return result.stdout;
   }
 
   async runDeviceMutationAsync(
     args: string[],
     operation: string,
-    signal?: AbortSignal
+    signal?: AbortSignal,
+    waitLimitMs?: number
   ): Promise<string> {
     const adb = this.getAdbExecutablePath();
     event('adb_server_run', { command: [adb, ...args].join(' ') });
     const result = await this.resolveAdbPromise(
-      runAdbDeviceMutationAsync(adb, args, operation, signal)
+      waitLimitMs == null
+        ? runAdbDeviceMutationAsync(adb, args, operation, signal)
+        : runBoundedAdbDeviceMutationAsync(adb, args, operation, waitLimitMs, signal)
     );
-    return [result.stdout, result.stderr].join('\n');
+    assertValidAdbUserOutput(result);
+    return result.stdout;
   }
 
   async runHostQueryAsync(
@@ -94,6 +101,7 @@ export class ADBServer {
         signal
       )
     );
+    assertValidAdbUserOutput(result);
     event('adb_file_output', { output: result.stdout });
     return result.stdout;
   }
@@ -110,8 +118,9 @@ export class ADBServer {
       if (error.signal === 'SIGINT') {
         throw new AbortCommandError();
       }
-      if (error.status === 255 && error.stdout.includes('Bad user number')) {
-        const userNumber = error.stdout.match(/Bad user number: (.+)/)?.[1] ?? env.EXPO_ADB_USER;
+      const processOutput = [error.stdout, error.stderr].filter(Boolean).join('\n');
+      if (error.status === 255 && processOutput.includes('Bad user number')) {
+        const userNumber = processOutput.match(/Bad user number: (.+)/)?.[1] ?? env.EXPO_ADB_USER;
         throw new CommandError(
           'EXPO_ADB_USER',
           `Invalid ADB user number "${userNumber}" set with environment variable EXPO_ADB_USER. Run "adb shell pm list users" to see valid user numbers.`
@@ -126,5 +135,16 @@ export class ADBServer {
       error.message = errorMessage;
       throw error;
     }
+  }
+}
+
+function assertValidAdbUserOutput(result: { stdout: string; stderr: string }): void {
+  if (!env.EXPO_ADB_USER) return;
+  const output = [result.stdout, result.stderr].filter(Boolean).join('\n');
+  if (/Bad user number|(?:user .* does not exist|no user with id)/i.test(output)) {
+    throw new CommandError(
+      'EXPO_ADB_USER',
+      `Invalid ADB user number "${env.EXPO_ADB_USER}" set with environment variable EXPO_ADB_USER. Run "adb shell pm list users" to see valid user numbers.`
+    );
   }
 }
--- a/packages/@expo/cli/src/start/platforms/android/ADBServer.ts
+++ b/packages/@expo/cli/src/start/platforms/android/ADBServer.ts
@@ -1,23 +1,21 @@
-import spawnAsync from '@expo/spawn-async';
-import { execFileSync } from 'child_process';
-
 import { Log } from '../../../log';
 import { env } from '../../../utils/env';
 import { AbortCommandError, CommandError } from '../../../utils/errors';
-import { installExitHooks } from '../../../utils/exit';
 import { event } from '../events';
 import { assertSdkRoot } from './AndroidSdk';
+import {
+  AdbProcessWaitError,
+  runAdbDeviceMutationAsync,
+  runAdbDeviceQueryAsync,
+  runAdbHostQueryAsync,
+  runBoundedAdbDeviceQueryAsync,
+  runBoundedAdbHostQueryAsync,
+} from './adbProcess';
 
 const BEGINNING_OF_ADB_ERROR_MESSAGE = 'error: ';
-
-// This is a tricky class since it controls a system state (side-effects).
-// A more ideal solution would be to implement ADB in JS.
-// The main reason this is a class is to control the flow of testing.
+const PROPERTY_QUERY_WAIT_LIMIT_MS = 10_000;
 
 export class ADBServer {
-  isRunning: boolean = false;
-  removeExitHook: () => void = () => {};
-
   /** Returns the command line reference to ADB. */
   getAdbExecutablePath(): string {
     try {
@@ -33,69 +31,71 @@ export class ADBServer {
     return 'adb';
   }
 
-  /** Start the ADB server. */
-  async startAsync(): Promise<boolean> {
-    if (this.isRunning) {
-      return false;
-    }
-    // clean up
-    this.removeExitHook = installExitHooks(() => {
-      if (this.isRunning) {
-        this.stopAsync();
-      }
-    });
-    const adb = this.getAdbExecutablePath();
-    const result = await this.resolveAdbPromise(spawnAsync(adb, ['start-server']));
-    const lines = result.stderr.trim().split(/\r?\n/);
-    const isStarted = lines.includes('* daemon started successfully');
-    this.isRunning = isStarted;
-    return isStarted;
-  }
-
-  /** Kill the ADB server. */
-  async stopAsync(): Promise<boolean> {
-    if (!this.isRunning) {
-      return false;
-    }
-    this.removeExitHook();
-    try {
-      await this.runAsync(['kill-server']);
-      return true;
-    } catch (error: any) {
-      Log.error('Failed to stop ADB server: ' + error.message);
-      return false;
-    } finally {
-      this.isRunning = false;
-    }
-  }
-
   /** Execute an ADB command with given args. */
-  async runAsync(args: string[]): Promise<string> {
-    // TODO: Add a global package that installs adb to the path.
+  async runDeviceQueryAsync(
+    args: string[],
+    operation: string,
+    signal?: AbortSignal
+  ): Promise<string> {
     const adb = this.getAdbExecutablePath();
 
-    await this.startAsync();
+    // NOTE(@kitten): We removed start-server/stop-server calls,
+    // because each command already negotiates the server automatically
 
     event('adb_server_run', { command: [adb, ...args].join(' ') });
-    const result = await this.resolveAdbPromise(spawnAsync(adb, args));
-    return result.output.join('\n');
+    const result = await this.resolveAdbPromise(
+      runAdbDeviceQueryAsync(adb, args, operation, signal)
+    );
+    return [result.stdout, result.stderr].join('\n');
+  }
+
+  async runDeviceMutationAsync(
+    args: string[],
+    operation: string,
+    signal?: AbortSignal
+  ): Promise<string> {
+    const adb = this.getAdbExecutablePath();
+    event('adb_server_run', { command: [adb, ...args].join(' ') });
+    const result = await this.resolveAdbPromise(
+      runAdbDeviceMutationAsync(adb, args, operation, signal)
+    );
+    return [result.stdout, result.stderr].join('\n');
+  }
+
+  async runHostQueryAsync(
+    args: string[],
+    operation: string,
+    signal?: AbortSignal,
+    waitLimitMs?: number
+  ): Promise<string> {
+    const adb = this.getAdbExecutablePath();
+    const result = await this.resolveAdbPromise(
+      waitLimitMs == null
+        ? runAdbHostQueryAsync(adb, args, operation, signal)
+        : runBoundedAdbHostQueryAsync(adb, args, operation, waitLimitMs, signal)
+    );
+    return result.stdout;
   }
 
   /** Get ADB file output. Useful for reading device state/settings. */
-  async getFileOutputAsync(args: string[]): Promise<string> {
+  async getFileOutputAsync(
+    args: string[],
+    { signal, waitLimitMs }: { signal?: AbortSignal; waitLimitMs?: number } = {}
+  ): Promise<string> {
     // TODO: Add a global package that installs adb to the path.
     const adb = this.getAdbExecutablePath();
 
-    await this.startAsync();
-
-    const results = await this.resolveAdbPromise(
-      execFileSync(adb, args, {
-        encoding: 'latin1',
-        stdio: 'pipe',
-      })
+    const result = await this.resolveAdbPromise(
+      runBoundedAdbDeviceQueryAsync(
+        adb,
+        args,
+        'device property/boot query',
+        waitLimitMs ?? PROPERTY_QUERY_WAIT_LIMIT_MS,
+        signal
+      )
     );
-    event('adb_file_output', { output: results });
-    return results;
+    event('adb_file_output', { output: result.stdout });
+    return result.stdout;
   }
 
   /** Formats error info. */
@@ -103,6 +103,9 @@ export class ADBServer {
     try {
       return await promise;
     } catch (error: any) {
+      if (error instanceof AdbProcessWaitError) {
+        throw error;
+      }
       // User pressed ctrl+c to cancel the process...
       if (error.signal === 'SIGINT') {
         throw new AbortCommandError();
--- a/packages/@expo/cli/src/start/platforms/android/AndroidDeviceManager.ts
+++ b/packages/@expo/cli/src/start/platforms/android/AndroidDeviceManager.ts
@@ -24,12 +24,22 @@ import { promptForDeviceAsync } from './promptAndroidDevice';
 const EXPO_GO_APPLICATION_IDENTIFIER = 'host.exp.exponent';
 
 export class AndroidDeviceManager extends DeviceManager<AndroidDebugBridge.Device> {
-  static async resolveFromNameAsync(name: string): Promise<AndroidDeviceManager> {
+  static async resolveFromNameAsync(query: string): Promise<AndroidDeviceManager> {
     const devices = await getDevicesAsync();
-    const device = devices.find((device) => device.name === name);
+    const device =
+      devices.find((device) => device.pid === query) ??
+      devices.find((device) => device.name === query);
 
     if (!device) {
-      throw new CommandError('Could not find device with name: ' + name);
+      const message = [
+        `No connected Android device or emulator matched "${query}" by serial or name.`,
+        'Available devices:',
+        ...devices.map(
+          (device) => `  ${device.name} (${device.pid ?? 'not attached'}, ${device.type})`
+        ),
+        'Pass a device serial from `adb devices` or a name from the list above to --device.',
+      ].join('\n');
+      throw new CommandError('BAD_ARGS', message);
     }
     return AndroidDeviceManager.resolveAsync({ device, shouldPrompt: false });
   }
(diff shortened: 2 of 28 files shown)


kitten:
Updated the test plan section and tested the scenarios I planned for, which confirms no regressions against a real Android device or an emulator, on macOS and Windows, as far as I can tell.

packages/@expo/cli/src/start/platforms/android/emulator.ts
```diff
@@ -76,56 +75,52 @@ export async function startDeviceAsync(
...
-      const manualCommand = `${whichEmulator()} @${device.name}`;
-      stopWaitingAndReject(
-        `The emulator (${device.name}) quit before it finished opening. You can try starting the emulator manually from the terminal with: ${manualCommand}`
-      );
-    });
-  });
+async function checkEmulatorBootAsync(name: string, signal?: AbortSignal): Promise<Device | null> {
+  const bootedDevices = await getAttachedDevicesAsync({ signal, shouldShowWaitingMessage: false });
+  const connected = bootedDevices.find((device) => device.name === name);
+  if (connected && (await isBootAnimationCompleteAsync(connected.pid, signal))) {
```
vonovak: [🏗️ **design**] Boot polling no longer survives a transient property read failure.

On `main`, `isBootAnimationCompleteAsync` caught every error and returned `false`, so this loop kept retrying until `EMULATOR_MAX_WAIT_TIMEOUT` (3 min). Now it propagates.

Failure scenario: an emulator registers as `state: device`, then `adb shell getprop init.svc.bootanim` hangs while it finishes booting. `getFileOutputAsync` applies `PROPERTY_QUERY_WAIT_LIMIT_MS` (10s) → `AdbProcessWaitError` → `createAdbOperationErrorAsync` spends up to another 4s on a host probe → `ADB_PROPERTY` `CommandError`. `checkEmulatorBootAsync` rethrows, the `while (true)` loop exits, and `startDeviceAsync` fails at ~14s. Before this change the emulator kept booting and succeeded.

Suggestion: treat a failed boot-animation probe as "not booted yet" and keep polling; propagate only when `operationSignal` aborts (cancellation or the overall timeout). `preserves errors when boot properties cannot be read` covers the `adb.ts` layer — the loop is the layer that needs the tolerance.

packages/@expo/cli/src/start/platforms/android/adb.ts
```diff
@@ -201,97 +237,117 @@ function shellQuote(value: string): string {
...
+  signal?: AbortSignal;
+  waitLimitMs?: number;
+  probeWaitLimitMs?: number;
+  shouldShowWaitingMessage?: boolean;
+} = {}): Promise<Device[]> {
+  const discoveryWaitLimitMs = waitLimitMs ?? DEVICE_DISCOVERY_WAIT_LIMIT_MS;
+  const waitSignal = AbortSignal.timeout(discoveryWaitLimitMs);
+  const operationSignal = signal ? AbortSignal.any([signal, waitSignal]) : waitSignal;
+  const endpoint = resolveAdbEndpoint();
+  const spinner = shouldShowWaitingMessage ? ora('Waiting for ADB device discovery').start() : null;
```
vonovak: [🏗️ **design**] The spinner is on for every caller of this low-level helper, including background ones: `startAdbReverseAsync`, `stopAdbReverseAsync`, `isDeviceBootedAsync`, `getDevicesAsync`.

Concrete problem on exit: `adbReverse.ts` registers `installExitHooks(() => { stopAdbReverseAsync(ports); })` and does not await it. On Ctrl+C the hook starts a spinner, `ora().start()` monkey-patches `console.*` globally (`utils/ora.ts`), the hook returns, and `process.exit()` runs before `stop()` — leaving a stray spinner frame during shutdown. In the common case discovery finishes in tens of milliseconds, so the spinner only flashes.

Suggestion: default `shouldShowWaitingMessage` to `false` and opt in from the interactive paths, or delay the spinner so fast discovery stays silent.

packages/@expo/cli/src/start/platforms/android/adbProcess.ts
```diff
@@ -0,0 +1,264 @@
...
+    this.remoteCompletionUnknown = remoteCompletionUnknown;
+    this.spawnFailed = spawnFailed;
+  }
+
+  remoteCompletionUnknown?: boolean;
+  spawnFailed?: boolean;
+  stdout?: string;
+  stderr?: string;
+  status?: number;
+  signal?: AbortSignal;
```
vonovak: [💡 suggestion] This field holds the POSIX exit signal, not an `AbortSignal`. `createProcessError` copies `error.signal` from `spawnAsync` (`'SIGTERM' | null`), `ADBServer.resolveAdbPromise` compares it to `'SIGINT'`, and `adbProcess-test.ts` assigns `signal: 'SIGTERM'`. `error: any` hides the mismatch, and `spawnFailed` detection reads the same value.

Declare it `NodeJS.Signals | null`.

packages/@expo/cli/src/start/platforms/android/ADBServer.ts
```diff
@@ -33,76 +31,81 @@ export class ADBServer {
...
-    const results = await this.resolveAdbPromise(
-      execFileSync(adb, args, {
-        encoding: 'latin1',
-        stdio: 'pipe',
-      })
+    const result = await this.resolveAdbPromise(
+      runBoundedAdbDeviceQueryAsync(
+        adb,
+        args,
+        'device property/boot query',
```
vonovak, line 99: [?] Property reads lost their explicit latin1 decoding. `main` used `execFileSync(adb, args, { encoding: 'latin1' })` with the comment "Prevent reading as UTF8." - just flagging

packages/@expo/cli/src/start/platforms/android/adbDeviceList.ts
```diff
@@ -0,0 +1,43 @@
...
+  return state === 'device';
+}
+
+const DEVICE_LIST_HEADER = 'List of devices attached';
+
+export function parseAdbDeviceList(output: string): AdbDeviceRecord[] {
+  return output
+    .split(/\r?\n/)
+    .map((line) => line.trim())
+    .filter((line) => line && line !== DEVICE_LIST_HEADER)
```
vonovak: [💡 suggestion] The ADB trace-line filter is gone. `main` dropped lines matching `/\.cpp:[0-9]+/` (WSL and `ADB_TRACE` output).

This parser accepts any `^(\S+)\s+(.+)$` line, so `adb D adb_client.cpp:393] adb_query: host:devices-l` becomes a record with `serial: 'adb'`, `state: 'D'`. That record reaches the device picker, `mergeDevices`, and `isDeviceBootedAsync`.

It is safe only while trace output never lands on stdout — the `adb-cold-start` fixture only exercises stderr. Keeping the one-line filter here is cheap insurance.

packages/@expo/cli/src/start/platforms/android/AndroidDeviceManager.ts
```diff
@@ -53,18 +61,46 @@ export class AndroidDeviceManager extends DeviceManager<AndroidDebugBridge.Devic
...
   }
 
   protected async attemptToStartAsync(): Promise<AndroidDebugBridge.Device | null> {
-    // TODO: Add a light-weight method for checking since a device could disconnect.
-    if (!(await AndroidDebugBridge.isDeviceBootedAsync(this.device))) {
+    // Only detached AVD inventory entries may enter the emulator launch path
+    if (this.device.isLaunchable) {
       this.device = await startDeviceAsync(this.device);
+    } else {
+      this.assertDeviceStateIsUsable(this.device);
```
vonovak, line 99: [💡 suggestion] This assertion throws before the `isAuthorized === false` branch below. Since `isAuthorized === isAdbDeviceStateUsable(state)`, that branch is now unreachable on both paths, so `logUnauthorized` and its `https://expo.fyi/authorize-android-device` link no longer print from this manager. The prompt path still shows them, but `run:android --device <name>` no longer does.

Suggestion: add the link to the `unauthorized` advice in `adbDiagnostics.getAdvice`, and drop the dead branch.

packages/@expo/cli/src/start/platforms/android/adbDiagnostics.ts
```diff
@@ -0,0 +1,121 @@
...
+function getAdvice(error: unknown, device?: AdbDeviceDiagnostic): string[] {
+  const advice: string[] = [];
+  if (error instanceof AdbProcessError && error.spawnFailed) {
+    advice.push('Check that Android SDK Platform-Tools is installed and ADB is available.');
+  }
+  if (device?.state === 'offline') {
+    advice.push(`Reconnect ${device.pid ? `device ${device.pid}` : 'the device'} and try again.`);
+  } else if (device?.state === 'unauthorized') {
+    advice.push('Authorize this computer on the device, then try again.');
+  } else if (device?.state && device.state !== 'device') {
```
vonovak, line 109: [💡 suggestion] `no permissions` falls into this branch and gets "Wait until ADB reports the device as ready, then try again." `parseDeviceLine` decodes that state specially, but it is a udev-rule or driver problem — waiting never fixes it.

Per the repo error-message guidance (what/why/how), point at udev rules and the ADB device documentation for that state instead.

packages/@expo/cli/src/start/platforms/android/getDevices.ts
```diff
@@ -7,22 +7,17 @@ import { listAvdsAsync } from './emulator';
...
-        isAuthorized: true,
-      };
-    });
+  // NOTE(@kitten): We don't assume AVD must succeed or be present, and still allow
+  // devices to be discovered and move on
+  let data: Device[];
+  try {
+    data = await listAvdsAsync();
+  } catch (error) {
+    if (!bootedDevices.length) throw error;
```
vonovak: [?]

When no device is attached and `listAvdsAsync` throws (for example spawn .../emulator ENOENT for a user without the emulator tool), `getDevicesAsync` rethrows the raw error. On main, `listAvdsAsync` swallowed the failure and the user got the friendly `No Android connected device found, and no emulators could be started automatically` message with the setup link. The test "preserves AVD inventory tool failures" pins the new behavior, but the raw `ENOENT` message fails the repo's what/why/how error-message rule for a common first-run case. Suggestion: keep the rethrow but wrap it in a `CommandError` that includes the old guidance plus the underlying message.

vonovak (commented):
I had a skill review the changes, here are some that look valid; I'll run some scenarios on my device and report back if there's anything

packages/@expo/cli/src/start/platforms/android/emulator.ts
kitten: Addressed in https://github.com/expo/expo/pull/49258/commits/d57636be94cf11ee6e38873b3e1e99d3a21e90a2

packages/@expo/cli/src/start/platforms/android/adbProcess.ts
kitten: Addressed in https://github.com/expo/expo/pull/49258/commits/6e6278fd17a9b9382f25e9c4272cbb715037f6c8 (oops)

packages/@expo/cli/src/start/platforms/android/adb.ts
kitten: Addressed in https://github.com/expo/expo/pull/49258/commits/387b12459a44d4acdd39299ead292699a77730f0

packages/@expo/cli/src/start/platforms/android/adbDeviceList.ts
kitten: Addressed in https://github.com/expo/expo/pull/49258/commits/bed97d107b6159b5ba103fb93299cf5f0a7953dd

packages/@expo/cli/src/start/platforms/android/getDevices.ts
kitten: Addressed in https://github.com/expo/expo/pull/49258/commits/22ff1a620c9508b6aac9c196b91432b2b8f6d890

packages/@expo/cli/src/start/platforms/android/adbDiagnostics.ts
kitten: Addressed in https://github.com/expo/expo/pull/49258/commits/22ff1a620c9508b6aac9c196b91432b2b8f6d890

packages/@expo/cli/src/start/platforms/android/AndroidDeviceManager.ts
kitten: Addressed in https://github.com/expo/expo/pull/49258/commits/f27db173e7c5220e82a476ccf114c2aff9344fe9

packages/@expo/cli/src/start/platforms/android/ADBServer.ts
kitten: I've looked this up online, since I'm unsure of why this used `latin1`. Specifically, it seems that we chose `latin1` to preserve binary data as-is, without enforcing UTF-8 or another conversion, instead of passing raw binary data directly.

The reasoning here was that adb itself passes bytes unchanged and we wanted to preserve that.

However, in practice, looking this up and asking an agent whether that's necessary (agent output):
- Android requires property values to be valid UTF‑8. AOSP property validation (https://android.googlesource.com/platform/system/core/+/refs/heads/android16-release/init/util.cpp)
- `getprop` writes the stored property string directly to stdout. AOSP `getprop` (https://android.googlesource.com/platform/system/core/+/refs/heads/main/toolbox/getprop.cpp)
- ADB transports shell output bytes without transcoding them. AOSP ADB shell service (https://android.googlesource.com/platform/packages/modules/adb/+/refs/heads/main/daemon/shell_service.cpp)

Hence, I'm assuming there must've been older devices that passed us invalid WTF-8 data. In theory, this won't immediately lead to corruption.

We only use:
- `init.svc.bootanim`
- `ro.product.cpu.abilist`
- `ro.product.cpu.abi`

All of these are ASCII-only outputs. If we receive invalid replies for these, the data is already corrupted and unusable.

vonovak (approved):

kitten:
Will merge but picking to prior SDKs will have to wait a bit until this has been sufficiently tested on `main`

PR #49264, opened 2026-08-24
[cli][metro] Include resolved config in Metro's transform cache key

Rakshit-gen:
# Why

Fixes #39619.

`expo export -p web` doesn't pick up a dynamic config value that depends on an environment variable across separate process runs, unless `--clear` is passed:

```bash
$ npx expo export -p web && grep -r API_BASE_URL dist/
"{\"extra\":{\"API_BASE_URL\":\"http://localhost:3000\"}

$ APP_ENV=development npx expo export -p web && grep -r API_BASE_URL dist/
"{\"extra\":{\"API_BASE_URL\":\"http://localhost:3000\"}   # still the old value
```

`npx expo config` (which just calls `getConfig()` directly) always shows the correct, current value — the divergence is specific to what Metro bundles for `export`/`start`.

# How

`babel-preset-expo`'s `expo-inline-manifest-plugin` rewrites `process.env.APP_MANIFEST` to a string literal containing the resolved public config, for any file that references it (this is how `expo-constants`' web implementation — `ExponentConstants.web.ts` — gets `Constants.manifest` on web). It resolves the config itself, correctly, on every transform.

The problem is one layer down, in Metro's **persistent transform cache**. That cache is keyed by (roughly) a base hash of the transformer + `customTransformOptions` + `dev`/`platform`/etc + the source file's own content hash (`packages/@expo/metro-config/src/transform-worker/metro-transform-worker.ts#getCacheKey`, consumed in `Transformer.transform` from `@expo/metro`). None of those inputs include anything about the *resolved* config — only the raw file bytes on disk. Since the file that references `process.env.APP_MANIFEST` doesn't change between two `expo export` runs, and none of the existing cache-key inputs change either, Metro's on-disk cache returns the transform output from the *first* run, silently keeping whatever config value was baked in then. `--clear` "fixes" it only by nuking the whole persistent cache.

This isn't unique to `APP_MANIFEST` — it's a gap for any Babel transform whose output depends on the resolved Expo config rather than only the file's own text. The fix generalizes to the whole class: thread a hash of the resolved config through `customTransformOptions`, which Metro already includes in its per-file cache key (that's exactly how existing config-derived values like `baseUrl` and `reactCompiler` already participate in cache-busting today).

Concretely:
- `ExpoMetroOptions` (`packages/@expo/cli/src/start/server/middleware/metroOptions.ts`) gets a new optional `manifestHash` field, forwarded into `customTransformOptions` in `getMetroDirectBundleOptions`.
- `MetroBundlerDevServer#startImplementationAsync` computes it once per dev-server/export session — `sha1(JSON.stringify(exp))` — alongside where `baseUrl`/`reactCompiler`/`asyncRoutes` are already derived from the same resolved `exp`, and stores it on `instanceMetroOptions` (so, like those other values, it's fixed for the lifetime of one `expo start` session — consistent with the existing behavior that editing `app.config.js` while the dev server is running already requires a restart to take effect).

This is intentionally coarse: it busts the transform cache for *any* config change, not just the fields that end up in the public manifest. That trades a bit of caching efficiency (only when the config itself changes, not on ordinary file edits) for correctness, and avoids hand-picking which config fields matter — `app.config.js` can read arbitrary env vars in arbitrary ways, so there's no fixed field list to special-case.

# Test Plan

Added `metroOptions.test.ts` coverage confirming `manifestHash` is forwarded into `customTransformOptions` (`getMetroDirectBundleOptions`), and confirmed the existing exact-shape (`toEqual`) tests in the same file are unaffected (an unset `manifestHash` is stripped like every other `undefined` option already is).

I traced the root cause down to the exact `getCacheKey` / `customTransformOptions` mechanism in Metro and confirmed via source that `customTransformOptions` is included in Metro's per-file `partialKey` (`@expo/metro`'s `DeltaBundler/Transformer.js`) — the same mechanism already used for `baseUrl`/`reactCompiler`. I did not additionally spin up two full separate `expo export -p web` processes end-to-end in this environment (that requires a real npm install of a fixture app plus two full production Metro bundling passes) — the unit-level test targets the exact plumbing this fix adds, and the fix follows the codebase's own established pattern for busting the cache on config-derived values.

# Checklist

### DIFF
--- a/packages/@expo/cli/CHANGELOG.md
+++ b/packages/@expo/cli/CHANGELOG.md
@@ -24,7 +24,7 @@
 
 ### 🐛 Bug fixes
 
-- Bust Metro's persistent transform cache when the resolved Expo config changes (e.g. `app.config.js` reading `process.env.APP_ENV`), so `expo export -p web` and `expo start` pick up the new value across separate process runs instead of requiring `--clear`. ([#39619](https://github.com/expo/expo/issues/39619))
+- Bust Metro's persistent transform cache when the resolved Expo config changes (e.g. `app.config.js` reading `process.env.APP_ENV`), so `expo export -p web` and `expo start` pick up the new value across separate process runs instead of requiring `--clear`. ([#49264](https://github.com/expo/expo/pull/49264) by [@rakshit-gen](https://github.com/rakshit-gen))
 - Serve relative manifest URLs only when the client itself sends the RFC 7239 `Forwarded` header, so that proxied requests from clients without relative-URL support, like released Expo Go versions through the WS tunnel, keep absolute URLs. ([#48997](https://github.com/expo/expo/pull/48997) by [@expo-bot](https://github.com/expo-bot))
 - Fail when `--private-key-path` is passed without `updates.codeSigningCertificate` in the resolved app config, instead of ignoring the flag and continuing without signing.
 - Show the Xcode build log path when `run:ios` fails. ([#48624](https://github.com/expo/expo/pull/48624) by [@ramonclaudio](https://github.com/ramonclaudio))
--- a/packages/@expo/cli/CHANGELOG.md
+++ b/packages/@expo/cli/CHANGELOG.md
@@ -24,6 +24,7 @@
 
 ### 🐛 Bug fixes
 
+- Bust Metro's persistent transform cache when the resolved Expo config changes (e.g. `app.config.js` reading `process.env.APP_ENV`), so `expo export -p web` and `expo start` pick up the new value across separate process runs instead of requiring `--clear`. ([#39619](https://github.com/expo/expo/issues/39619))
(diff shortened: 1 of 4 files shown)


kitten:
This will eventually instead be addressed by #47750

On a side-note, hashing by the config isn't stable since it includes other values that are subject to change that shouldn't invalidate the cache, like resolved paths

Rakshit-gen:
Thanks for the pointer and the catch — closing this in favor of #47750, which is clearly the right way to solve this class of problem (per-module cache-vary dimensions) rather than my blunt whole-config hash. You're right that hashing all of `exp` pulls in resolved absolute paths and other incidental values that shouldn't invalidate the cache, so this would trade one correctness bug for a caching-efficiency regression. Not worth carrying forward once #47750 lands.

PR #49271, opened 2026-08-24
[android][notifications] Handle a null result bundle from ResultReceiver

vonovak:
# Why

# How

# Test Plan

# Checklist

---


vonovak:
Closing because the error paths always provide a non-null result bundle containing the caught exception, so the defensive case covered here is not reachable through NotificationsService.

Issue #46703, opened 2026-06-09
[expo-image][Android] Image remains invisible/blank when source is updated asynchronously during transition effect


david-shiko:
### Minimal reproducible example

https://pastebin.com/jBND5fjG

### Environment & Platforms
- **Platform:** Android (iOS is completely unaffected)
- **Environment:** Expo Go, Development Builds, and Production/Standalone apps
- **Package Manager:** npm / yarn / pnpm

### Steps to Reproduce
1. Create a custom JavaScript wrapper component that renders `<ExpoImage transition={200} />` (or any duration > 0).
2. Initialize the local state inside the wrapper with a placeholder image.
3. Introduce a short asynchronous delay (`setTimeout` or promise resolution) inside `useEffect` on mount.
4. Update the state to a remote network image URI once the async task finishes.
5. Pass a dynamic changing `key` to the parent JavaScript wrapper component (e.g., `<AvatarImageWrapper key={uri} />`).

### Expected Behavior
The underlying native view should seamlessly apply the transition effect and display the newly provided remote image on the screen.

### Actual Behavior
The image asset downloads perfectly into memory, and the native `onLoad` callback successfully fires in the console. However, the visual container on Android remains completely blank/white, behaving as if the view's native alpha/opacity channel gets stuck at `0` due to the interrupted cross-fade animation.

### Verified Workarounds
- **Workaround 1:** Setting `transition={0}` completely bypasses the bug.
- **Workaround 2:** Placing the unique `key` directly onto the native `<ExpoImage key={source} />` component inside the wrapper forces it to render correctly, whereas putting the `key` on the parent JS wrapper fails.

### Environment

```text
david@user-824811db:~/PycharmProjects/RubikLoveMobileV2$ npx expo-env-info
Need to install the following packages:
expo-env-info@2.0.14
Ok to proceed? (y) y

  expo-env-info 2.0.14 environment info:
    System:
      OS: Linux 6.17 Ubuntu 24.04.4 LTS 24.04.4 LTS (Noble Numbat)
      Shell: 5.2.21 - /bin/bash
    Binaries:
      Node: 22.22.0 - /home/david/.nvm/versions/node/v22.22.0/bin/node
      Yarn: 1.22.22 - /home/david/.nvm/versions/node/v22.22.0/bin/yarn
      npm: 11.14.0 - /home/david/.nvm/versions/node/v22.22.0/bin/npm
      Watchman: 4.9.0 - /usr/bin/watchman
    npmPackages:
[...]

PR #49283, opened 2026-08-24
fix(expo-image): run android transition cleanup when the fade is cancelled

oscnord:
# Why

On Android, images can become permanently blank when their source changes or their view is recycled during a transition.

Fade-out cleanup used `ViewPropertyAnimator.withEndAction`, but Android drops that runnable when the animation is cancelled. Since starting another alpha animation cancels the previous one, the Glide target may never be returned to the pool. Once both targets remain marked as used, the wrapper has no target available for the next image.

Running cleanup through `setListener` fixes cancellation, but the listener persists across later animations. A stale listener can therefore recycle a newer image, and target identity alone is insufficient because the same target object can be rebound.

Related: #46703. Follow-up to #46752; this retains and strengthens its target-ownership guard.

# How

- Replace `withEndAction` with an `AnimatorListenerAdapter` that performs cleanup once on either completion or cancellation.
- Assign each view/target binding an ID and recycle only when both the target and binding ID still match.
- In `recycleView`, invalidate the binding before cancelling, remove the persistent listener after cancellation, and reset alpha to `1f`.
- Add Robolectric coverage for listener persistence, same-target rebinding, and valid cleanup.

## Transition behavior

If `source` changes faster than the transition duration, cancellation frees the outgoing target immediately. The incoming image fades in from the background rather than maintaining an uninterrupted two-image crossfade. Redesigning that behavior is outside this PR.

# Test Plan

- Android API 33 emulator: 16 images without `recyclingKey`, `cachePolicy="none"`, sources changing every 250 ms, and 800 ms transitions. After 120 swaps and the final transition settling, all 16 images remained visible.
- Physical Samsung SM-A146P (Android 15): repeatedly scrolled a horizontal list of about 20 covers inside a vertical list. Before the fix, blanking accumulated to 71%; after the fix, it stayed at the freshly loaded baseline of 11–19% across two runs.
- `:expo-image:testDebugUnitTest`: 3 tests, 0 failures.
- `:expo-image:spotlessCheck`: passed.
- `pnpm --filter expo-image lint`: passed.

The physical-device comparison used the released `55.0.10` sources with the target guard from `main` applied locally.

# Checklist

### DIFF
--- a/packages/expo-image/CHANGELOG.md
+++ b/packages/expo-image/CHANGELOG.md
@@ -13,7 +13,7 @@
 
 ### 🐛 Bug fixes
 
-- [Android] Fixed images in recycled lists going permanently blank: the transition's cleanup ran through `withEndAction`, which `ViewPropertyAnimator` drops when the animation is cancelled, so the Glide target was never returned to the pool. ([#PRNUM](https://github.com/expo/expo/pull/PRNUM) by [@oscnord](https://github.com/oscnord))
+- [Android] Fixed images in recycled lists going permanently blank: the transition's cleanup ran through `withEndAction`, which `ViewPropertyAnimator` drops when the animation is cancelled, so the Glide target was never returned to the pool. ([#49283](https://github.com/expo/expo/pull/49283) by [@oscnord](https://github.com/oscnord))
 - [web] Fixed Expo Image's internal `dataSet` marker being overwritten by a user-provided `dataSet`. ([#48821](https://github.com/expo/expo/pull/48821) by [@Brentlok](https://github.com/Brentlok))
 - [iOS] Fixed `generateThumbhashAsync` crashing on images with extreme aspect ratios. ([#47189](https://github.com/expo/expo/issues/47189) by [@gabrieldonadel](https://github.com/gabrieldonadel))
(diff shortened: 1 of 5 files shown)


vonovak:
@expo-bot verify

oscnord:
Thanks for catching this. `ViewPropertyAnimator.setListener` persists across animations, and target identity alone was insufficient because the same target object can be reused for a newer binding.

Fixed in [`ef90c93a`](https://github.com/expo/expo/pull/49283/commits/ef90c93a8638adc803da55c231c6f00a30b619b7) by tracking each view/target binding and allowing cleanup only for the exact binding that scheduled it. `recycleView()` also removes the persistent listener after cancellation.

I tested the requested no-`recyclingKey` shape on Android API 33: 16 images, `cachePolicy="none"`, source changes every 250 ms, 800 ms transitions, and 120 swaps. After the final transition settled, all 16 images remained visible.

I also added three Robolectric regression tests covering the persistent listener, reuse of the same target for a newer binding, and valid cleanup running only once. The Android unit tests, Kotlin formatting check, and package lint all pass.

I updated the PR description with the new implementation and verification details, including the rapid-cancellation crossfade behavior noted above.

vonovak:
@expo-bot verify

PR #49284, opened 2026-08-24
[expo-video][iOS] Fix crash when the device runs out of storage while caching

### LINKED ISSUE #49133: [expo-video] NSFileHandleOperationException crash when useCaching is true and device is low on storage
### Example Project

https://github.com/dmilin1/hydra

### Steps to reproduce

Load up a video using expo-video with useCache enabled on a device with storage maxed out. App will crash. Sometimes multiple videos need to be loaded in a row since iOS will try to free up space, making this bug inconsistent to reproduce.

### Environment

```text
    System:
      OS: Windows 10 10.0.19045
    Binaries:
      Node: 24.18.0 - C:\Program Files\nodejs\node.EXE
      npm: 12.0.1 - C:\Program Files\nodejs\npm.CMD
      Watchman: 20231008.002904.0 - C:\ProgramData\chocolatey\bin\watchman.EXE
    npmPackages:
      expo: ^57.0.0 => 57.0.8
      expo-updates: ~57.0.10 => 57.0.10
      react: 19.2.3 => 19.2.3
      react-dom: 19.2.3 => 19.2.3
    Expo Workflow: managed
```

#### Heads up that I've used AI for this section to help me figure out what's going on. I'm not particularly experienced in writing Swift.

```
try writeHandle.seek(toOffset: UInt64(offset))
writeHandle.write(data)          // ← deprecated, NON-throwing ObjC API
writeHandle.synchronizeFile()
```

writeHandle.write(data) is the old -[NSFileHandle writeData:], which reports failure by raising an Objective-C exception (NSFileHandleOperationException) rather than returning an error. Swift do/catch — including the one in CachedResource.writeData that's clearly meant to absorb write failures — cannot catch ObjC exceptions, so when the disk is full the exception propagates uncaught and kills the app. The fix expo-video should have is the throwing variant, try writeHandle.write(contentsOf: data) (and try synchronize()).

#### I'm also getting another crash involving the expo-video cache which might be related:

EXC_BREAKPOINT: Exception 6, Code 1, Subcode 6897455884
  libswiftCore        0x19b1ed30c  swift_unknownObjectRetain
  libswiftCore        0x19b1fc9b4  swift_cvw_initWithCopyImpl
  Hydra               0x102d75964  Sequence.contains
  Hydra               0x102d77430  VideoCacheManager.limitCacheSize
  Hydra               0x102d75f80  closure in VideoCacheManager.ensureCacheSize
  Hydra               0x1029caddc  thunk for closure
  libdispatch         0x1d90169a4  _dispatch_call_block_and_release
  libdispatch         0x1d90301e0  _dispatch_client_callout
  libdispatch         0x1d901efac  _dispatch_lane_serial_drain
  libdispatch         0x1d901faa8  _dispatch_lane_invoke
  libdispatch         0x1d9029da8  _dispatch_root_queue_drain_deferred_wlh
  libdispatch         0x1d90296a8  _dispatch_workloop_worker_thread
  libsystem_pthread   0x1fde9b3ac  _pthread_wqthread

#### AI description of the crash above:

VideoCacheManager.openFiles is a plain Set<URL> ([VideoCacheManager.swift:16](vscode-webview://0me3v0qvj3r33o13nhriqpjfqsfv0nvke3k6ffmvodtr5q5gtrr5/node_modules/expo-video/ios/Cache/VideoCacheManager.swift#L16)) with no synchronization, accessed from multiple threads:

- **Reader (crashing thread)**: every time a video asset deinits, ensureCacheSize() runs limitCacheSize on the private cacheQueue. When the cache is over the limit, it calls fileIsOpen, whose openFiles.contains { ... } iterates the set ([VideoCacheManager.swift:180](vscode-webview://0me3v0qvj3r33o13nhriqpjfqsfv0nvke3k6ffmvodtr5q5gtrr5/node_modules/expo-video/ios/Cache/VideoCacheManager.swift#L180)) — that's the Sequence.contains frame in your stack.

- **Writers (other threads)**: registerOpenFile/unregisterOpenFile mutate the same set from MediaFileHandle.init/deinit and from the onAssetDeinit closure in [CacheVideoAssetTransportProvider.swift:48-53](vscode-webview://0me3v0qvj3r33o13nhriqpjfqsfv0nvke3k6ffmvodtr5q5gtrr5/node_modules/expo-video/ios/Cache/CacheVideoAssetTransportProvider.swift#L48-L53). Deinits run on whatever thread drops the last reference — typically the main thread when AVFoundation tears a player down.

Mutating a copy-on-write Set while another thread iterates it produces a torn read; the iterator retains a freed element and Swift traps — exactly your signature: swift_unknownObjectRetain / swift_cvw_initWithCopyImpl under Sequence.contains in limitCacheSize, on a libdispatch worker thread.

huextrat:
# Why

fixes https://github.com/expo/expo/issues/49133

`expo-video`'s disk cache crashes the app when the device runs out of storage.

`MediaFileHandle` writes cached media through the legacy `NSFileHandle` API (`writeData:`, `synchronizeFile`, `readData(ofLength:)`, `closeFile`). Those methods report I/O failures by raising an Objective-C `NSFileHandleOperationException`, which a Swift `do`/`catch` cannot catch. So when a write hits `ENOSPC` the exception unwinds straight past the error handling that `CachedResource.writeData` already has, and the process is terminated:

```
NSFileHandleOperationException: *** -[NSConcreteFileHandle writeData:]: No space left on device
  at MediaFileHandle.write
  at CachedResource.writeData
  at CachableRequest.saveData

NSFileHandleOperationExceptionUnderlyingError:
  Error Domain=NSCocoaErrorDomain Code=640 "…not enough space."
  UserInfo={NSUnderlyingError=… Error Domain=NSPOSIXErrorDomain Code=28 "No space left on device"}
```

We see this in production on iOS: it is fully fatal, and it is triggered by nothing more than a user with a full phone watching a video.

# How

Switch `MediaFileHandle` to the throwing Swift replacements, all available since iOS 13.4 (`ExpoVideo.podspec` targets iOS 16.4, so no availability guard is needed):

| before | after |
| --- | --- |
| `writeHandle.write(data)` | `try writeHandle.write(contentsOf: data)` |
| `writeHandle.synchronizeFile()` | `try writeHandle.synchronize()` |
| `writeHandle.seekToEndOfFile()` | `try writeHandle.seekToEnd()` |
| `readHandle.readData(ofLength:)` | `try readHandle.read(upToCount:)` |
| `handle.closeFile()` | `try? handle.close()` |

`write(data:atOffset:)` is already `throws`, so the failure now propagates to the existing `catch` in `CachedResource.writeData`, which logs a warning. The consequence of a full disk becomes "this segment is not cached" instead of "the app dies".

Two details worth noting for review:

- In `WriteCoordinator.performWrite` the write happens *before* `mediaInfo.addDataRange(…)`, so a throw leaves the range unmarked. A partially written chunk is never recorded as cached and will not be served back — the cache index cannot be corrupted by this path.
- `read(upToCount:)` returns `nil` at EOF where `readData(ofLength:)` returned empty `Data`, so it is mapped through `?? Data()` to keep the existing semantics.

`append(data:)` has no callers today, but it had the same defect; it now catches and logs rather than raising, keeping its non-throwing signature.

# Test Plan

Reproduced both behaviours directly against a full filesystem — an 8 MB HFS+ disk image — writing 256 KB chunks until `ENOSPC`, with each API in its own process (the Objective-C exception terminates the process, so they cannot share one).

Legacy API, i.e. the current `main` behaviour — the production crash, reproduced exactly:

```
*** Terminating app due to uncaught exception 'NSFileHandleOperationException',
    reason: '*** -[NSConcreteFileHandle writeData:]: No space left on device'
    1 libobjc.A.dylib   objc_exception_throw + 88
    3 Foundation        -[NSConcreteFileHandle writeData:] + 292
libc++abi: terminating due to uncaught exception of type NSException
```

Same scenario with `write(contentsOf:)` — same underlying failure, now a catchable Swift error:

```
caught Swift error -> Error Domain=NSCocoaErrorDomain Code=640
  "The file couldn't be saved because there isn't enough space."
  UserInfo={NSUnderlyingError=… Error Domain=NSPOSIXErrorDomain Code=28 "No space left on device"}
```

`NSCocoaErrorDomain 640` / `NSPOSIXErrorDomain 28` match the `NSFileHandleOperationExceptionUnderlyingError` captured in the production reports, confirming this is the same failure path rather than a different one.

`swiftc -typecheck` passes on the modified file, and SwiftLint reports no violations on it.

# Checklist

### DIFF
--- a/packages/expo-video/CHANGELOG.md
+++ b/packages/expo-video/CHANGELOG.md
@@ -12,7 +12,7 @@
 
 ### 🐛 Bug fixes
 
-- [iOS] Fixed a crash when the device runs out of storage while writing to the video cache. `FileHandle.writeData:` raises an uncatchable Objective-C `NSFileHandleOperationException` on `ENOSPC`; the throwing Swift APIs are now used so the error is caught and logged instead. ([#XXXXX](https://github.com/expo/expo/pull/XXXXX) by [@huextrat](https://github.com/huextrat))
+- [iOS] Fixed a crash when the device runs out of storage while writing to the video cache. `FileHandle.writeData:` raises an uncatchable Objective-C `NSFileHandleOperationException` on `ENOSPC`; the throwing Swift APIs are now used so the error is caught and logged instead. ([#49284](https://github.com/expo/expo/pull/49284) by [@huextrat](https://github.com/huextrat))
 - [Android] Guard `PictureInPictureParams.Builder.setAutoEnterEnabled` against `NoSuchMethodError` on stock OEM firmwares that report API 31+ without shipping the method, which crashed the app from `VideoView.onLayout` even when Picture in Picture was disabled. ([#48957](https://github.com/expo/expo/pull/48957) by [@onlyshyun](https://github.com/onlyshyun))
 - [iOS] Fix races between overlapping source loads and player release. ([#47967](https://github.com/expo/expo/pull/47967) by [@behenate](https://github.com/behenate))
 - [iOS] Set the default `audioMixingMode` to `auto`, [as documented](https://docs.expo.dev/versions/latest/sdk/video/#audiomixingmode); was `doNotMix`. ([#47363](https://github.com/expo/expo/issues/47363) by [@andymatuschak](https://github.com/andymatuschak))
--- a/packages/expo-video/CHANGELOG.md
+++ b/packages/expo-video/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 ### 🐛 Bug fixes
 
+- [iOS] Fixed a crash when the device runs out of storage while writing to the video cache. `FileHandle.writeData:` raises an uncatchable Objective-C `NSFileHandleOperationException` on `ENOSPC`; the throwing Swift APIs are now used so the error is caught and logged instead. ([#XXXXX](https://github.com/expo/expo/pull/XXXXX) by [@huextrat](https://github.com/huextrat))
 - [Android] Guard `PictureInPictureParams.Builder.setAutoEnterEnabled` against `NoSuchMethodError` on stock OEM firmwares that report API 31+ without shipping the method, which crashed the app from `VideoView.onLayout` even when Picture in Picture was disabled. ([#48957](https://github.com/expo/expo/pull/48957) by [@onlyshyun](https://github.com/onlyshyun))
 - [iOS] Fix races between overlapping source loads and player release. ([#47967](https://github.com/expo/expo/pull/47967) by [@behenate](https://github.com/behenate))
 - [iOS] Set the default `audioMixingMode` to `auto`, [as documented](https://docs.expo.dev/versions/latest/sdk/video/#audiomixingmode); was `doNotMix`. ([#47363](https://github.com/expo/expo/issues/47363) by [@andymatuschak](https://github.com/andymatuschak))
--- a/packages/expo-video/ios/Cache/MediaFileHandle.swift
+++ b/packages/expo-video/ios/Cache/MediaFileHandle.swift
@@ -88,7 +88,7 @@ internal class MediaFileHandle {
 
     do {
       try readHandle.seek(toOffset: UInt64(offset))
-      let data = readHandle.readData(ofLength: length)
+      let data = try readHandle.read(upToCount: length) ?? Data()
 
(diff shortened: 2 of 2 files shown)


behenate (approved):
Looks great! Thank you for the PR 😄

PR #35758, opened 2025-03-28
[expo-asset] Add .riv to accepted media types

alextoudic:
# Why

I'm using Rive in an Expo app, and its API requires passing resource names to components to load animations.

Rive's documentation currently directs users to a GitHub issue that suggests using a community plugin for preloading assets, similar to `expo-asset` ([link](https://rive.app/docs/runtimes/react-native/adding-rive-to-expo#local-assets)).

Given Rive's growing popularity and the fact that `expo-asset` already supports `.lottie` files, I feel it would be a natural and valuable addition to support `.riv` files as well.

# How

Added `.riv` to `MEDIA_TYPES` in `expo-asset`.

# Test Plan

I locally patched `expo-asset` and verified that Rive files load correctly in their example app.

# Checklist

- [x] I added a `changelog.md` entry and rebuilt the package sources according to [this short guide](https://github.com/expo/expo/blob/main/CONTRIBUTING.md#-before-submitting)
- [x] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: updated a module plugin).
- [x] Conforms with the [Documentation Writing Style Guide](https://github.com/expo/expo/blob/main/guides/Expo%20Documentation%20Writing%20Style%20Guide.md)

PR #49285, opened 2026-08-24
[asset] Link assets of any file type in the config plugin

### LINKED ISSUE #46008: [expo-asset] [Expo 55] Custom asset type is not supported during prebuild

https://github.com/ckhicks/expo-asset-asdf

### Steps to reproduce

Goal: Use `expo-asset` to bundle a custom file type during prebuild.

Steps:
- Run `npx create-expo-app@latest`
- Run `npx expo install expo-asset`
- Enable plugin in app.json -> `["expo-asset",{"assets":["./assets/asdf.onnx"]}]`
- Add filetype in metro.config.js -> `config.resolver.assetExts.push("onnx")`
- Run `npx expo prebuild --clean`

Output:
```
✔ Cleared android, ios code
› Android package name: com.developer.asdf
› Apple bundle identifier: com.developer.asdf
✔ Created native directories
✔ Updated package.json
» ios: expo-asset: `.onnx` is not a supported asset type
» android: expo-asset: `.onnx` is not a supported asset type
✔ Finished prebuild
```

Reproduces 100% deterministically in the linked repo.

### Environment

```text
    System:
      OS: macOS 15.7.5
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 22.14.0 - ~/.nvm/versions/node/v22.14.0/bin/node
      npm: 11.4.2 - ~/.nvm/versions/node/v22.14.0/bin/npm
      Watchman: 2025.04.28.00 - /opt/homebrew/bin/watchman
    Managers:
      CocoaPods: 1.16.2 - /Users/ckhicks/.rbenv/shims/pod
    SDKs:
      iOS SDK:
    IDEs:
      Android Studio: 2023.3 AI-233.14808.21.2331.11842104
    npmPackages:
      expo: ~55.0.25 => 55.0.25 
      expo-router: ~55.0.15 => 55.0.15 
      react: 19.2.0 => 19.2.0 
      react-dom: 19.2.0 => 19.2.0 
    npmGlobalPackages:
      eas-cli: 18.11.0
    Expo Workflow: bare
```

### Expo Doctor Diagnostics

```text
```

giaBaoJS:
# Why

Fixes #46008.

Declaring a custom asset type with the `expo-asset` config plugin prints a warning and then silently drops the file from both native projects:

```
» ios: expo-asset: `.onnx` is not a supported asset type
» android: expo-asset: `.onnx` is not a supported asset type
```

`validateAssets` filtered every asset against a hardcoded `ACCEPTED_TYPES` allowlist in `packages/expo-asset/plugin/src/utils.ts`. Anything not on the list was warned about and removed from the array, so `prebuild` finished successfully with the asset missing.

The list had to be extended by hand for each new type (`.riv` in #35758, `.glb` in #42495), and it was the only thing rejecting the file. The copy pipeline underneath is already generic:

- `withAssetsAndroid.ts` routes images to `res/drawable`, fonts to `assets/fonts`, media to `res/raw`, and every other extension falls through the `else` branch into `app/src/main/assets`.
- `withAssetsIos.ts` sends images to `Images.xcassets` and passes everything else to `addResourceFileToGroup`, which takes any file.

This also contradicts the docs, which tell you to add other file types to `assetExts` in the Metro config and then link them with this plugin.

# How

Removed the `ACCEPTED_TYPES` gate from `validateAssets`, along with `ACCEPTED_TYPES` and `MODEL_TYPES`, which had no other reader in the repo. `IMAGE_TYPES`, `FONT_TYPES` and `MEDIA_TYPES` stay, because they still drive where a file is placed natively.

Kept the two rejections that exist for a reason:

- Android resource name validation, unchanged.
- The warning that points `.ttf` / `.otf` at `expo-font`, unchanged.

The allowlist was also incidentally filtering directory expansion. `resolveAssetPaths` returns every entry of a linked directory, including nested directories and files like `.DS_Store`, and those entries used to be discarded later because their extension was not on the list. With the gate gone, a nested directory would have reached `fsp.copyFile`, which rejects when handed a directory and would have failed the prebuild. `resolveAssetPaths` now skips nested directories and hidden files explicitly instead of relying on that side effect.

`packages/expo-asset/jest.config.js` passed `[]` for the composite Jest preset's sub-targets, so `plugin/` had no tests at all. It now passes `['plugin']`, matching `expo-font`, and the PR adds tests for `utils.ts` and for the Android mod.

The docs page listed a closed set of supported types, so it is updated to describe the two routed groups and say that every other type is copied as-is.

# Test Plan

`packages/expo-asset` is the only package touched.

New tests pass:

```
$ npx jest --selectProjects plugin --verbose
Running one project: plugin

PASS plugin plugin/src/__tests__/withAssetsAndroid-test.ts
  withAssetsAndroid
    ✓ copies an asset whose extension is not one of the built-in types (72 ms)
    ✓ routes images and media to their resource directories (13 ms)
    ✓ copies the files of a directory without failing on nested directories (27 ms)

PASS plugin plugin/src/__tests__/utils-test.ts
  validateAssets
    ✓ keeps an asset whose extension is not one of the built-in types (android) (1 ms)
    ✓ keeps an asset whose extension is not one of the built-in types (ios)
    ✓ keeps assets whose extension the plugin routes explicitly (android)
    ✓ keeps assets whose extension the plugin routes explicitly (ios)
    ✓ drops fonts and points at expo-font (android) (1 ms)
    ✓ drops fonts and points at expo-font (ios)
    ✓ drops assets with an invalid Android resource name
    ✓ does not apply the Android resource name rule on iOS
  resolveAssetPaths
    ✓ resolves a relative file path from the project root (3 ms)
    ✓ expands a directory to its files, skipping nested directories and hidden files (7 ms)
    ✓ rejects a path that does not exist (5 ms)

Tests:       14 passed, 14 total
```

Reverting only `utils.ts` to `main` and keeping the new tests fails for the right reasons, the asset being dropped rather than an import or compile error:

```
$ git stash push -- packages/expo-asset/plugin/src/utils.ts
$ npx jest --selectProjects plugin

  ● withAssetsAndroid › copies an asset whose extension is not one of the built-in types

    Received promise rejected instead of resolved
    Rejected to value: [Error: ENOENT: no such file or directory, open
    '/var/folders/.../android/app/src/main/assets/model.onnx']

  ● validateAssets › keeps an asset whose extension is not one of the built-in types (ios)

    - Expected  - 3
    + Received  + 1

    - Array [
    -   "/project/assets/model.onnx",
    - ]
    + Array []

  ● resolveAssetPaths › expands a directory to its files, skipping nested directories and hidden files

    - Expected  - 0
    + Received  + 2

      Array [
    +   ".../assets/.DS_Store",
        ".../assets/image.png",
        ".../assets/model.onnx",
    +   ".../assets/nested",
      ]

Test Suites: 2 failed, 2 total
Tests:       6 failed, 8 passed, 14 total
```

The 8 that stay green in that run are the regression guards: images and media still routed, fonts still rejected with the `expo-font` warning, invalid Android resource names still rejected. Restoring `utils.ts` returns all 14 to green.

Whole package suite:

```
$ npx jest
Test Suites: 29 passed, 29 total
Tests:       4 skipped, 203 passed, 207 total
```

Type check, lint, format:

```
$ npx tsc -b tsconfig.all.json --force
TYPECHECK EXIT: 0
$ npx oxlint --config oxlint.config.mjs .
LINT EXIT: 0
$ npx expo-module format --check
All matched files use the correct format.
```

The `withAssetsAndroid` test drives the plugin's real dangerous mod against a temp project, so it checks the actual copy, not just the filter.

# Checklist

<sub>Written with AI assistance. Every command output above was run locally and pasted verbatim.</sub>


docs/pages/versions/unversioned/sdk/asset.mdx
```diff
@@ -46,11 +46,13 @@ You can configure `expo-asset` using its built-in [config plugin](/config-plugin
...
         'An array of asset files or directories to link to the native project. The paths should be relative to the project root so that the file names, whether specified directly or using a directory, will become the resource names.',
         '',
-        'Supported file types:',
+        'Any file type can be linked. These types are placed in a type-specific location in the native project:',
         '* Images: `.png`, `.jpg`, `.gif`',
         '* Media: `.mp4`, `.mp3`, `.lottie`, `.riv`',
-        '* SQLite database files: `.db`',
-        '* 3D Models: `.glb`',
+        '',
+        'Every other type, such as `.db`, `.glb`, or a model file used by a native library, is copied into the app bundle on iOS and into the `assets` directory on Android.',
```
amandeepmittal: ✏️ **style** `mechanics.md#file-and-directory-names-in-bold`

File extensions and directory names are bold, not inline code, so `.db`, `.glb`, and `assets` become **.db**, **.glb**, and a bold path. The extensions on the two lines above carry the same problem, so it is worth converting the whole block in one pass.

---

✏️ **style** `terminology.md#platform-order`

This sentence names iOS before Android. When one sentence refers to both platforms, order them Android, then iOS.

---

This sentence packs both platforms into one clause. Split the Android and iOS destinations into separate sentences, and name the real Android path, **android/app/src/main/assets**, instead of "the `assets` directory".

```suggestion
        'The plugin copies every other type, such as **.db**, **.glb**, or a model file used by a native library. On Android, the file goes to **android/app/src/main/assets**. On iOS, the plugin adds it to the app bundle as a resource.',
```

docs/pages/versions/unversioned/sdk/asset.mdx
```diff
@@ -46,11 +46,13 @@ You can configure `expo-asset` using its built-in [config plugin](/config-plugin
...
+        '',
+        'Fonts are not linked by this plugin. Use `expo-font` instead.',
```
amandeepmittal: ✏️ **style** `links-and-headings.md#internal-links`

This line sends the reader to another library without a link. Point the package name at the Font API reference with the full path, `/versions/latest/sdk/font/`.

---

"Fonts are not linked by this plugin" is passive. Write it as "The plugin does not link fonts."

```suggestion
        'The plugin does not link fonts. Use [`expo-font`](/versions/latest/sdk/font/) instead.',
```

docs/pages/versions/unversioned/sdk/asset.mdx
```diff
@@ -46,11 +46,13 @@ You can configure `expo-asset` using its built-in [config plugin](/config-plugin
       description: [
         'An array of asset files or directories to link to the native project. The paths should be relative to the project root so that the file names, whether specified directly or using a directory, will become the resource names.',
         '',
-        'Supported file types:',
+        'Any file type can be linked. These types are placed in a type-specific location in the native project:',
```
amandeepmittal: ⚠️ **correctness** `comment-format.md#severity`

On iOS only images get a type-specific location, **Images.xcassets**. The media types listed below land in the app bundle exactly like every other type, so this sentence holds only on Android.

---

⚠️ **correctness** `prose.md#core-principles`

Android still drops any asset whose base name is not lowercase a-z, 0-9, or underscore, and this page never says so. Add that constraint here, because "any file type" reads as no constraints at all.

---

"can be linked" and "are placed" are both passive. Name the actor: the plugin accepts the file, and the plugin places it.

```suggestion
        'The plugin accepts any file type. On Android, it places these types in a type-specific resource directory:',
```

amandeepmittal (changes requested):

docs/pages/versions/unversioned/sdk/asset.mdx
giaBaoJS: Both correctness points hold up against the plugin source, so I applied your suggestion and added each fact as its own sentence.

iOS: `withAssetsIos.ts:30-36` splits `validAssets` on `IMAGE_TYPES` only. Images go through `addImageAssets` into `Images.xcassets`, and everything else, media included, goes through `addResourceFiles` into the `Resources` group. So the type-specific list held for Android only, as you said. The page now reads: "On iOS, only images get a type-specific location, **Images.xcassets**. The plugin adds media files to the app bundle as resources, the same as every other type."

Android names: `utils.ts:30-40` calls `isValidAndroidAssetName` for android only, and on a miss it adds a `WarningAggregator` warning and returns undefined, which drops the file from `validAssets` so it never gets copied. One correction to the rule as stated: it is tighter than lowercase a-z, 0-9, or underscore. `validPattern` in `packages/@expo/config-plugins/src/utils/validations.ts:2` is `/^[a-z][a-z0-9_]*$/`, so the name must also start with a letter, and there is a Java reserved word blocklist on top of that (`validations.ts:4-56`). The existing tests confirm it: `_name`, `123abc` and `abstract` are all rejected. I documented the full rule rather than the short form, since the short form would still send someone chasing a name that fails.

Worth flagging that the plugin warns on a dropped file rather than failing silently, so this was a docs gap and not a code one.

docs/pages/versions/unversioned/sdk/asset.mdx
giaBaoJS: Applied, and converted the whole block in one pass: both bullet lines, the new paragraph, and the `.db`, `.lottie` and `.riv` in the Note below are now bold. Took your sentence split verbatim, with **android/app/src/main/assets** as the named Android path.

docs/pages/versions/unversioned/sdk/asset.mdx
giaBaoJS: Applied verbatim. `/versions/latest/sdk/font/` matches how the other unversioned pages link it, for example `sdk/ui/swift-ui/text.mdx:238`.

brentvatne:
@expo-bot verify with fable and also determine whether this is reasonable on android

giaBaoJS:
Both Android points the verification raises are worth taking, and I am happy to do either or both here rather than leave them as follow-ups.

For the case sensitivity one, `photo.JPG` silently landing in `assets/` instead of `res/drawable` is a behaviour change my PR introduced, so I would rather fix it in this PR than ship it. Lowercasing the extension in `getAssetDir` and `validateAssets` is the whole change.

For the name rule, moving `isValidAndroidAssetName` into the `res/` branch of `getAssetDir` would let `My-Model.onnx` through on Android, since files under `app/src/main/assets` accept any name. That one is a real behaviour change for existing users rather than a fix to my own regression, so I would rather you decide it than assume it.

Say which you want and I will push.

giaBaoJS:
@amandeepmittal All three docs comments are applied: your suggestion blocks verbatim, extensions and directory names bolded across the whole description block including the Note, and the two correctness gaps written out as their own sentences. I also fixed the case sensitivity regression this PR introduced, so `photo.JPG` now routes to `res/drawable` like `photo.jpg` instead of falling through to `assets/`, and there is a test covering it. The `isValidAndroidAssetName` question is still open and waiting on a maintainer call, since changing it would alter behaviour for existing users.

giaBaoJS:
The iOS side had the same gap: `withAssetsIos` split images from other resources with a case-sensitive extension check, so `photo.JPG` went to the `Resources` group instead of `Images.xcassets`. That is fixed too now, with a test, so both platforms treat an uppercase extension the same as its lowercase form.

amandeepmittal (approved):
Thanks for the updates! Let's wait for a technical review.

PR #49286, opened 2026-08-24
[expo-video][iOS] Fix data race on the video cache open-file registry

huextrat:
# Why

`VideoCacheManager.openFiles` is an unsynchronized `Set<URL>` that is read and mutated from three different threads:

| accessor | called from | thread |
| --- | --- | --- |
| `registerOpenFile` | `MediaFileHandle.init` | the main thread when a player is created, or a `URLSession` delegate queue while a resource loads |
| `unregisterOpenFile` | `MediaFileHandle.deinit` | whichever thread releases the last reference |
| `fileIsOpen` | `limitCacheSize` | `cacheQueue` |

`ensureCacheSize()` hops onto `cacheQueue` and walks the cache directory, calling `fileIsOpen` for every file, while players elsewhere in the app are free to open and close cache files at the same time. Concurrent mutation and iteration of a native Swift `Set` is undefined behaviour, and in production this surfaces as `EXC_BAD_ACCESS` inside the standard library's set internals — with the wild, obviously-corrupted fault addresses typical of a torn container rather than a plain null dereference.

Nothing in the current code prevents this: `openFiles` has no lock, no queue confinement, and no isolation.

# How

Guard the three accessors with an `NSLock`.

This matches how the package already synchronizes shared collections — `SynchronizedHashTable` (used by `VideoManager`, `NowPlayingManager` and `ResourceLoaderDelegate`) is the same `NSLock`-around-a-collection shape, and `MediaFileHandle` carries its own lock for the same reason. `SynchronizedHashTable` cannot be reused directly here because it is constrained to `T: AnyObject` and `URL` is a struct.

The two alternatives do not work:

- **An actor** would require `await` at the call sites, but `unregisterOpenFile` is called from `MediaFileHandle.deinit`, and a `deinit` cannot await.
- **Confining to `cacheQueue`** with `cacheQueue.sync` would deadlock: `limitCacheSize` already runs on `cacheQueue` and calls `fileIsOpen` synchronously.

The lock covers only the set operations themselves, never the file system work in `limitCacheSize` and `deleteAllFilesInCacheDirectory`. That is deliberate: the same lock is acquired from `deinit` on arbitrary threads, including the main thread, so holding it across `FileManager.removeItem` calls would let cache trimming block an unrelated thread on disk I/O.

This leaves the check-then-delete in `limitCacheSize` non-atomic — a file can be opened between the `fileIsOpen` filter and the `removeItem`. That is intentionally not addressed here: unlinking an open file on POSIX leaves the descriptor valid, so playback continues and the only consequence is a cache entry that has to be re-fetched. Trading a possible main-thread stall for that seemed like the wrong deal; this PR is scoped to the memory-safety bug.

`import Foundation` is added because the file previously imported only `ExpoModulesCore` and now refers to `NSLock`; the sibling files in `ios/Cache` that use Foundation types import it explicitly too.

# Test Plan

I want to be upfront that a data race has no deterministic reproduction, so unlike a crash with a single triggering condition this one cannot be demonstrated by a script. What I can offer:

- The unsafe accesses are verifiable by reading: `openFiles` has exactly three accessors, all listed above, and the threads they run on follow directly from their call sites (`MediaFileHandle.init`/`deinit` and `limitCacheSize` on `cacheQueue`).
- The crash it produces in production is a memory-corruption `EXC_BAD_ACCESS` in Swift set internals, occurring while the cache is being trimmed, which is the expected signature for this pattern.
- `swiftc -typecheck` passes on the modified file and SwiftLint reports no violations on it. The `SendableClosureCaptures` warnings the compiler emits for this file are pre-existing on `main` and unchanged by this diff — I checked both versions.
- No behaviour changes: the lock is uncontended in the single-threaded case and no call site changes semantics.

Happy to run this under Thread Sanitizer if you would like harder evidence before merging.

# Related

Independent of #49284, which fixes a separate crash in the same directory (`MediaFileHandle` raising an uncatchable Objective-C exception when the disk is full). The two branches are cut from `main` separately and touch different files; whichever lands second can be rebased.

# Checklist


behenate (approved):

behenate:
I fixed the conflict in `CHANGELOG.md` 🫡

PR #49288, opened 2026-08-24
[fingerprint] Normalize virtual store paths before hashing them

### LINKED ISSUE #44808: [@expo/fingerprint] dir source hash changes when pnpm peer-dep paths reshuffle, even if file contents are unchanged

giaBaoJS:
# Why

Fixes #44808.

pnpm and the other isolated installers put a package under a store directory whose name encodes how peer dependencies resolved:

```
node_modules/.pnpm/expo-camera@1.0.0_react-native@0.86.0/node_modules/expo-camera/android
```

That directory name changes whenever any unrelated dependency is added, removed, or bumped. `Hash.ts` feeds project relative paths into the hasher, so every one of those paths changes and the fingerprint changes with it, even though every file content hash is identical.

@kitten already pointed at the fix in the issue:

> @Kudo: I'm not sure which exact procedure this is a part of, but if this is part of the native modules one, we might want to normalise paths to the package path, if we aren't yet, for the hashing?

That is what this does.

This issue is assigned to @Kudo. If it is already being handled on your side, say the word and I will close this.

## How this sits next to #48704

Worth flagging, because I did not expect it. On current `main` these paths never reach the hasher at all: `**/node_modules/**/node_modules/**` in `DEFAULT_IGNORE_PATHS` (`src/Options.ts:66`) matches the whole store layout, so `createDirHashResultsAsync` returns `null` for anything inside it. Measured with memfs on `main`, with two trees whose file contents are identical and whose store suffix differs:

```
default options       dirA = null
                      dirB = null

ignore negated        dirA.hex = 1c33378fe84ecf927177b72e9518cc59810566fa
                      dirB.hex = 9190136ef6b0e0bb5b07902643993fcf8bb1df2a
                      fpA.hash = bbcc4b5d9ab8abce5a26a37eccc867ed4391b049
                      fpB.hash = e47cfa42f4afecc9f3aa555e69b2ca8ccf74497b
```

So today this bug is latent, and it goes live the moment #48704 restores those sources. This change is what keeps them stable afterwards.

It also means this should not move any existing fingerprint on `main`. Every source type that can carry a store path (`dir`, `file`, `package`, and the config plugin sources from `resolveLoadedModuleSourcesAsync`) is filtered through the same ignore paths. The exception is a project that re-includes nested `node_modules` through its own `ignorePaths`, which is also how the tests exercise the behavior.

# How

`normalizeVirtualStorePath` in `src/utils/Path.ts` collapses the store segment down to the package path:

```
node_modules/.pnpm/expo-camera@1.0.0_react-native@0.86.0/node_modules/expo-camera/android
  -> node_modules/expo-camera/android
```

It is applied at the two places in `Hash.ts` where a path enters a hasher: the source id in `createFingerprintFromSourcesAsync`, and each child id in `createDirHashResultsAsync`.

Both are needed. With only the `createDirHashResultsAsync` site normalized, the dir hash is stable but the fingerprint hash still moves, because `createSourceId` feeds `source.filePath` straight into the top level hasher. Measured, that half fix gives:

```
  virtual store paths
    ✓ should return the same dir hash when only the store suffix changed (78 ms)
    ✕ should return the same fingerprint when only the store suffix changed (37 ms)

    expect(received).toEqual(expected) // deep equality

    Expected: "c37284669ae52db3a8124d15b6be62a4d54cfd3f"
    Received: "3adb115e28d48e8219cb43589aac3dfc418b6b44"
```

Only the store segment is collapsed, so path information still carries everywhere else and a real move or rename is still detected. A scoped package keeps its scope, a `../` prefix is kept, a genuinely nested `node_modules` inside a store package is left alone, and a path that merely contains `.pnpm` is returned unchanged. Sources keep their real path for reading files and for `debugInfo`.

The regex covers `.pnpm`, Bun's `.bun`, and Yarn's pnpm linker `.store`, matching the one #48704 adds to `normalizeFilePath` for ignore matching. If #48704 lands first I am happy to rebase and have the two share a single regex.

## Known gap

The `expoAutolinkingConfig:*` and `rncoreAutolinkingConfig:*` `contents` sources embed the same unstable paths inside their JSON (`project.sourceDir`, `pod.podspecDir`, `aarProject.projectDir`). Those are `contents` sources, so they do not go through this code. Fixing that means normalizing in the sourcer, which overlaps with #48661, so I left it out. Happy to follow up.

# Test Plan

`pnpm test` in `packages/@expo/fingerprint`.

New tests, 7 for the helper and 6 for the hashing behavior:

```
  virtual store paths
    ✓ should return the same dir hash when only the store suffix changed (3 ms)
    ✓ should return the same fingerprint when only the store suffix changed (2 ms)
    ✓ should still change the dir hash when a file is renamed (2 ms)
    ✓ should still change the dir hash when a file is moved into a subdirectory (2 ms)
    ✓ should still change the dir hash when a file moves out of a nested dependency (3 ms)
    ✓ should keep two store packages with identical contents apart (2 ms)

  normalizeVirtualStorePath
    ✓ should collapse a pnpm virtual store segment down to the package path
    ✓ should keep the scope of a scoped package
    ✓ should collapse the same shape for other isolated installers (1 ms)
    ✓ should keep a parent directory prefix
    ✓ should collapse every store segment of a nested workspace path
    ✓ should keep a real nested dependency inside a store package
    ✓ should leave paths that only look like a virtual store alone

Tests:       53 passed, 53 total
```

Four of the six hashing tests assert that a hash still *changes*, so the suite is not vacuous: a rename, a move into a subdirectory, a file moving out of a nested `node_modules`, and two different store packages with byte identical contents.

Counterfactual, reverting only `Hash.ts` and keeping the helper and the tests:

```
  ● virtual store paths › should return the same dir hash when only the store suffix changed

    Expected: "956b8e556d4ff79d61596a0897f7f5aead403692"
    Received: "8b78e512a5842e6b411f9f2d1eeed754a57f8bc8"

  ● virtual store paths › should return the same fingerprint when only the store suffix changed

    Expected: "b35dbd278614baee52c7d623186b9b2a5f4737c6"
    Received: "277871f513f7a15c46bcf2eadab1b6ce1c3f13bb"

Tests:       2 failed, 51 passed, 53 total
```

Two failures, both on unequal hashes, and the four non-vacuity tests stay green because they assert inequality either way.

Whole package suite, run on this branch and on a clean `main` for comparison:

```
main         Tests: 12 failed, 141 passed, 153 total
this branch  Tests: 12 failed, 154 passed, 166 total
```

Same 8 failing suites before and after, so this adds 13 passing tests and breaks nothing. Those 8 pre-existing failures are in `Sourcer`, `Bare`, `Expo`, `Fingerprint`, `ProjectWorkflow`, `ExpoConfig`, `PatchPackage`, and `Fingerprint-filehook`, and are unrelated to paths.

`oxlint --config oxlint.config.mjs .` exits 0. `oxfmt --check` reports all four changed files correctly formatted. `tsc -b tsconfig.all.json` produces byte identical output before and after the change, with no diagnostic in any file I touched.

# Checklist

### DIFF
--- a/packages/@expo/fingerprint/src/hash/Hash.ts
+++ b/packages/@expo/fingerprint/src/hash/Hash.ts
@@ -20,7 +20,11 @@ import type {
   NormalizedOptions,
 } from '../Fingerprint.types';
 import { createLimiter, type Limiter } from '../utils/Concurrency';
-import { isIgnoredPathWithMatchObjects, toPosixPath } from '../utils/Path';
+import {
+  isIgnoredPathWithMatchObjects,
+  normalizeVirtualStorePath,
+  toPosixPath,
+} from '../utils/Path';
 import { nonNullish } from '../utils/Predicates';
 import { profile } from '../utils/Profile';
 import { FileHookTransform } from './FileHookTransform';
@@ -42,7 +46,7 @@ export async function createFingerprintFromSourcesAsync(
   const hasher = createHash(options.hashAlgorithm);
   for (const source of fingerprintSources) {
     if (source.hash != null) {
-      hasher.update(createSourceId(source));
+      hasher.update(normalizeVirtualStorePath(createSourceId(source)));
       hasher.update(source.hash);
     }
   }
@@ -238,7 +242,7 @@ export async function createDirHashResultsAsync(
 
   const children: (DebugInfoFile | DebugInfoDir | undefined)[] = [];
   for (const result of results) {
-    hasher.update(result.id);
+    hasher.update(normalizeVirtualStorePath(result.id));
     hasher.update(result.hex);
     children.push(result.debugInfo);
   }
--- a/packages/@expo/fingerprint/src/utils/Path.ts
+++ b/packages/@expo/fingerprint/src/utils/Path.ts
@@ -173,3 +173,31 @@ export function getNodeModulesPackageJsonPath(relativePosixPath: string): string
   }
   return `${packageDir}/package.json`;
 }
+
+/**
+ * Matches the virtual store segment of a path from a package manager that installs packages in
+ * isolation: pnpm's `node_modules/.pnpm`, Bun's `node_modules/.bun`, and Yarn's pnpm linker's
+ * `node_modules/.store`.
+ *
+ * The store directory name is not stable. It encodes the package version and a suffix derived from
+ * how peer dependencies resolved, e.g.
+ * `node_modules/.pnpm/expo@52.0.49_react@19.0.0_react-native@0.76.7/node_modules/expo`.
+ */
+const VIRTUAL_STORE_REGEX = /(^|\/)node_modules\/\.(?:pnpm|bun|store)\/[^/]+\/node_modules\//g;
+
+/**
+ * Collapse the virtual store segment of a project relative path down to the package path, so a
+ * name a package manager derives from peer dependency resolution never reaches a hash.
+ *
+ * ```
+ * node_modules/.pnpm/expo-camera@1.0.0_react-native@0.86.0/node_modules/expo-camera/android
+ *   -> node_modules/expo-camera/android
+ * ```
+ *
+ * Only the store segment is collapsed. A scoped package keeps its scope, a `../` prefix is kept, a
+ * genuinely nested `node_modules` inside a store package is left alone, and a path that merely
+ * contains `.pnpm` is returned unchanged.
+ */
+export function normalizeVirtualStorePath(filePath: string): string {
+  return filePath.replace(VIRTUAL_STORE_REGEX, '$1node_modules/');
+}
--- a/packages/@expo/fingerprint/CHANGELOG.md
+++ b/packages/@expo/fingerprint/CHANGELOG.md
@@ -20,6 +20,7 @@
 
 - Set development mode before loading Expo config and `.env` files. ([#48839](https://github.com/expo/expo/pull/48839) by [@ramonclaudio](https://github.com/ramonclaudio))
 - Fixed ignore patterns (built-in and `.fingerprintignore`) not matching on Windows, which made fingerprints differ between Windows machines and EAS builds ("Runtime version mismatch"). ([#46816](https://github.com/expo/expo/pull/46816) by [@blurbyte](https://github.com/blurbyte))
+- Collapsed the virtual store segment of a path, e.g. pnpm's **node_modules/.pnpm**, before hashing it, so a peer dependency reshuffle no longer changes `dir` hashes when no file content changed. ([#49288](https://github.com/expo/expo/pull/49288) by [@giaBaoJS](https://github.com/giaBaoJS))
 
 ### 💡 Others
 
--- a/packages/@expo/fingerprint/src/hash/__tests__/Hash-test.ts
+++ b/packages/@expo/fingerprint/src/hash/__tests__/Hash-test.ts
@@ -274,6 +274,124 @@ describe(createDirHashResultsAsync, () => {
   });
 });
 
+describe('virtual store paths', () => {
+  afterEach(() => {
+    vol.reset();
+  });
+
+  // pnpm and the other isolated installers place a package under a store directory whose name
+  // encodes the resolved peer dependencies. The default ignore paths currently drop those paths as
+  // nested `node_modules`, so opt them back in to exercise the hashing itself.
+  const optIn = { ignorePaths: ['!**/node_modules/**/node_modules/**'] };
+
+  const cameraA =
+    'node_modules/.pnpm/expo-camera@1.0.0_react-native@0.86.0/node_modules/expo-camera';
+  // Same package and version, a peer dependency elsewhere in the tree was bumped.
+  const cameraB =
+    'node_modules/.pnpm/expo-camera@1.0.0_react-native@0.86.1_@babel+core@7.28.4/node_modules/expo-camera';
+
+  async function hashDirAsync(volJSON: Record<string, string>, dirPath: string) {
+    vol.reset();
+    vol.fromJSON(volJSON);
+    const options = await normalizeOptionsAsync('/app', optIn);
+    const result = await createDirHashResultsAsync(dirPath, createLimiter(3), '/app', options);
+    return result?.hex ?? null;
+  }
+
+  async function fingerprintDirSourceAsync(volJSON: Record<string, string>, dirPath: string) {
+    vol.reset();
+    vol.fromJSON(volJSON);
+    const options = await normalizeOptionsAsync('/app', optIn);
+    const sources: HashSource[] = [
+      { type: 'dir', filePath: dirPath, reasons: ['expoAutolinkingAndroid'] },
+    ];
+    return (await createFingerprintFromSourcesAsync(sources, '/app', options)).hash;
+  }
+
+  it('should return the same dir hash when only the store suffix changed', async () => {
+    const hexA = await hashDirAsync(
+      { [`/app/${cameraA}/android/build.gradle`]: 'apply plugin: com.android.library' },
+      `${cameraA}/android`
+    );
+    const hexB = await hashDirAsync(
+      { [`/app/${cameraB}/android/build.gradle`]: 'apply plugin: com.android.library' },
+      `${cameraB}/android`
+    );
+
+    expect(hexA).not.toBeNull();
+    expect(hexB).toEqual(hexA);
+  });
+
+  it('should return the same fingerprint when only the store suffix changed', async () => {
+    const hashA = await fingerprintDirSourceAsync(
+      { [`/app/${cameraA}/android/build.gradle`]: 'apply plugin: com.android.library' },
+      `${cameraA}/android`
+    );
+    const hashB = await fingerprintDirSourceAsync(
+      { [`/app/${cameraB}/android/build.gradle`]: 'apply plugin: com.android.library' },
+      `${cameraB}/android`
+    );
+
+    expect(hashB).toEqual(hashA);
+  });
+
+  it('should still change the dir hash when a file is renamed', async () => {
+    const before = await hashDirAsync(
+      { [`/app/${cameraA}/android/build.gradle`]: 'apply plugin: com.android.library' },
+      `${cameraA}/android`
+    );
+    const after = await hashDirAsync(
+      { [`/app/${cameraA}/android/settings.gradle`]: 'apply plugin: com.android.library' },
+      `${cameraA}/android`
+    );
+
+    expect(after).not.toEqual(before);
+  });
+
+  it('should still change the dir hash when a file is moved into a subdirectory', async () => {
+    const before = await hashDirAsync(
+      { [`/app/${cameraA}/android/build.gradle`]: 'apply plugin: com.android.library' },
+      `${cameraA}/android`
+    );
+    const after = await hashDirAsync(
+      { [`/app/${cameraA}/android/gradle/build.gradle`]: 'apply plugin: com.android.library' },
+      `${cameraA}/android`
+    );
+
+    expect(after).not.toEqual(before);
+  });
+
+  it('should still change the dir hash when a file moves out of a nested dependency', async () => {
+    const before = await hashDirAsync(
+      {
+        [`/app/${cameraA}/node_modules/nested/android/build.gradle`]:
+          'apply plugin: com.android.library',
+      },
+      cameraA
+    );
+    const after = await hashDirAsync(
+      { [`/app/${cameraA}/nested/android/build.gradle`]: 'apply plugin: com.android.library' },
+      cameraA
+    );
+
+    expect(after).not.toEqual(before);
+  });
+
+  it('should keep two store packages with identical contents apart', async () => {
+    const media = 'node_modules/.pnpm/expo-media-library@1.0.0/node_modules/expo-media-library';
+    const cameraHash = await fingerprintDirSourceAsync(
+      { [`/app/${cameraA}/android/build.gradle`]: 'apply plugin: com.android.library' },
+      `${cameraA}/android`
+    );
+    const mediaHash = await fingerprintDirSourceAsync(
+      { [`/app/${media}/android/build.gradle`]: 'apply plugin: com.android.library' },
+      `${media}/android`
+    );
+
+    expect(mediaHash).not.toEqual(cameraHash);
+  });
+});
+
 describe(createSourceId, () => {
   it(`should use filePath as id for file or dir`, () => {
     const fileSource: HashSource = {
--- a/packages/@expo/fingerprint/src/utils/__tests__/Path-test.ts
+++ b/packages/@expo/fingerprint/src/utils/__tests__/Path-test.ts
@@ -7,6 +7,7 @@ import {
   getNodeModulesPackageJsonPath,
   isIgnoredPath,
   normalizeFilePath,
+  normalizeVirtualStorePath,
   pathExistsAsync,
   toPosixPath,
 } from '../Path';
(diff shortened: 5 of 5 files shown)


packages/@expo/fingerprint/src/utils/Path.ts
```diff
@@ -173,3 +173,31 @@ export function getNodeModulesPackageJsonPath(relativePosixPath: string): string
...
+ * node_modules/.pnpm/expo-camera@1.0.0_react-native@0.86.0/node_modules/expo-camera/android
+ *   -> node_modules/expo-camera/android
+ * ```
+ *
+ * Only the store segment is collapsed. A scoped package keeps its scope, a `../` prefix is kept, a
+ * genuinely nested `node_modules` inside a store package is left alone, and a path that merely
+ * contains `.pnpm` is returned unchanged.
+ */
+export function normalizeVirtualStorePath(filePath: string): string {
+  return filePath.replace(VIRTUAL_STORE_REGEX, '$1node_modules/');
```
kitten, line 202: We ideally wouldn't be retroactively normalising against a known lists of paths as that's brittle.

@Kudo I believe had changes that instead exclude the resolved path entirely now in a looser default config of fingerprint in the upcoming version

packages/@expo/fingerprint/src/utils/Path.ts
```diff
@@ -173,3 +173,31 @@ export function getNodeModulesPackageJsonPath(relativePosixPath: string): string
...
+/**
+ * Matches the virtual store segment of a path from a package manager that installs packages in
+ * isolation: pnpm's `node_modules/.pnpm`, Bun's `node_modules/.bun`, and Yarn's pnpm linker's
+ * `node_modules/.store`.
+ *
+ * The store directory name is not stable. It encodes the package version and a suffix derived from
+ * how peer dependencies resolved, e.g.
+ * `node_modules/.pnpm/expo@52.0.49_react@19.0.0_react-native@0.76.7/node_modules/expo`.
+ */
+const VIRTUAL_STORE_REGEX = /(^|\/)node_modules\/\.(?:pnpm|bun|store)\/[^/]+\/node_modules\//g;
```
Kudo: **[suggestion]** `VIRTUAL_STORE_REGEX` is the same isolated-install layout that #48704 already adds inside `normalizeFilePath` (that PR uses `/node_modules\/\.(?:pnpm|bun|store)\/[^/]+\/node_modules\//g` without the `(^|/)` anchor). Two copies will drift the first time pnpm/Bun/Yarn change the store directory name.

**Suggestion:** Extract one helper (this function, including the `(^|/)` anchor, is the better shape) and call it from both ignore matching and hashing. If #48704 merges first, rebase onto it rather than keeping a second regex.

packages/@expo/fingerprint/src/utils/Path.ts
```diff
@@ -173,3 +173,31 @@ export function getNodeModulesPackageJsonPath(relativePosixPath: string): string
...
+/**
+ * Collapse the virtual store segment of a project relative path down to the package path, so a
+ * name a package manager derives from peer dependency resolution never reaches a hash.
+ *
+ * ```
+ * node_modules/.pnpm/expo-camera@1.0.0_react-native@0.86.0/node_modules/expo-camera/android
+ *   -> node_modules/expo-camera/android
+ * ```
+ *
+ * Only the store segment is collapsed. A scoped package keeps its scope, a `../` prefix is kept, a
```
Kudo: **[suggestion]** The JSDoc after the example restates the unit tests (scoped package, `../` prefix, nested `node_modules`, a path that merely contains `.pnpm`). Combined with the regex comment above it, that is design-leaking WHAT, not a constraint the next reader needs.

**Suggestion:** Keep the one-line why (peer-derived store names must not reach a hash) and the example. Drop the trailing inventory of cases; the tests already lock those in.

packages/@expo/fingerprint/src/hash/Hash.ts
```diff
@@ -42,7 +46,7 @@ export async function createFingerprintFromSourcesAsync(
   const hasher = createHash(options.hashAlgorithm);
   for (const source of fingerprintSources) {
     if (source.hash != null) {
-      hasher.update(createSourceId(source));
+      hasher.update(normalizeVirtualStorePath(createSourceId(source)));
```
Kudo, line 49: do you know why we should convert the path? isn't the path stable? please try to explain in examples

Kudo (commented):
## Summary

The hashing change is correct for `file`/`dir` source IDs: collapsing `node_modules/.(pnpm|bun|store)/<store-id>/node_modules/` down to `node_modules/` at the two hasher.update sites keeps a peer-dep reshuffle from moving the hash, without dropping real path info (rename, nested install, two different packages). Applying it only at hash time, and leaving `debugInfo` / `source.filePath` as the real path, is the right split. Tests cover the helper edge cases and the non-vacuous hashing cases well.

Two caveats: this does not stabilize the autolinking `contents` sources, which still embed the raw store path in JSON and are the fingerprint input that actually moves on current `main`; and the new helper duplicates the virtual-store regex already proposed in #48704.

## Issue counts by severity

- bugs: 0
- suggestions: 3
- nits: 0

Kudo (changes requested):

packages/@expo/fingerprint/src/hash/Hash.ts
giaBaoJS: The path is deterministic for a given lockfile, but it is not stable across resolution
changes, and it moves for reasons that have nothing to do with native code. pnpm 11.22.0,
node 24.13.0, macOS.

Smallest version:

```
$ cat package.json
{ "dependencies": { "use-sync-external-store": "1.2.0", "react": "18.2.0" } }
$ pnpm install --ignore-scripts && ls node_modules/.pnpm | grep use-sync
use-sync-external-store@1.2.0_react@18.2.0

# bump only the peer
$ sed -i '' 's/"react": "18.2.0"/"react": "18.3.1"/' package.json && pnpm install --ignore-scripts
$ ls node_modules/.pnpm | grep use-sync
use-sync-external-store@1.2.0_react@18.2.0
use-sync-external-store@1.2.0_react@18.3.1
```

Both copies of the package are byte identical (sha256 of the tree, relative paths,
b8724290...097e in both).

The version that matters for us. Real app, expo ~54.0.12 + expo-camera + expo-image +
react-native 0.81.4:

```
$ ls -d node_modules/.pnpm/expo-camera@*
node_modules/.pnpm/expo-camera@17.0.10_expo@54.0.37_@babel+core@7.29.7_supports-color@8.1.1__react-native@_a52ad48902769ee774d7870fa50a4cca
```

That name is exactly 120 characters and ends in a 32 character hex digest, so pnpm hit its
virtual store dir name cap and hashed the full peer set into the tail. Every peer feeds that
digest, including `supports-color`, which arrives through `@babel/core`'s optional peer.

So pin `supports-color` and nothing else. Overrides live in pnpm-workspace.yaml on pnpm 10+,
so package.json is untouched:

```
$ cat pnpm-workspace.yaml
overrides:
  supports-color: 9.0.0
$ pnpm install --ignore-scripts && ls -d node_modules/.pnpm/expo-camera@*
..._supports-color@8.1.1__react-native@_a52ad48902769ee774d7870fa50a4cca
..._supports-color@9.0.0__react-native@_5399d3e3323d7a2201fa0f58aab9794a

$ shasum -a 256 package.json      # same before and after
7b8d479edf213f277836f4869e5e2672cfe2d4f352b8adaed5fd0e36e3e8cc07  package.json

$ diff -rq <before>/expo-camera/android <after>/expo-camera/android
(no output)
```

9 of 9 autolinked modules moved. 0 packageVersion values changed. No lockfile is a
fingerprint input, so without the path this change would be invisible to the hash, which is
the correct outcome for a terminal colour library.

Now the part where you are right, and it changes what this PR is worth.

I reproduced your first caveat and it is the whole story. `getExpoAutolinkingAndroidSourcesAsync`
relativizes `project.sourceDir` at Expo.ts:335 and then pushes the raw JSON as a `contents`
source at Expo.ts:395. `createSourceId` for a `contents` source returns `source.id`, which is
the literal string `expoAutolinkingConfig:android` with no path in it, and Hash.ts:284 hashes
`source.contents` verbatim. So my two `hasher.update` changes cannot reach it. Running the
sourcer's own transform over the project above:

```
clean install, no override:  sha1 61d25f4e871691b36234f69510d28a5e41b354d9
only supports-color pinned:  sha1 f9d5cca91f362d5a2d7464673de6ba82eecf3f1c
```

That is the input that actually moves, and this PR does not touch it.

Worse for the PR: while checking that, I found it changes no hash at all on current main.
`DEFAULT_IGNORE_PATHS` already contains `**/node_modules/**/node_modules/**`, and I measured
`isIgnoredPath` against it with the package's own code:

```
ignored=true   node_modules/.pnpm/<id>/node_modules/expo-camera/android
ignored=true   node_modules/.pnpm/<id>/node_modules/expo-camera/package.json
ignored=true   ../../node_modules/.pnpm/<id>/node_modules/expo-camera/android
ignored=false  node_modules/expo-camera/android/build.gradle
```

So every store path already gets a null hash and is skipped by the `source.hash != null` guard
at Hash.ts:47, and never reaches line 49. On the balanced default it would not matter anyway,
since `nativeModuleSourceType: 'package'` makes the id `name@version`. The tell was in my own
test file: it had to opt in with `ignorePaths: ['!**/node_modules/**/node_modules/**']` to
exercise the code at all, and I should have read that as a signal instead of working around it.

So this only becomes load-bearing after #48704, and only under `preset: 'strict'`. Your second
caveat is also correct, the regex is a near copy of the one in #48704, same file, same constant
name. The only difference is an anchor that matters solely for a path segment ending in
`node_modules`.

I would rather close this than leave a no-op PR open. If you want the two `hasher.update` lines,
they belong as a commit on #48704, and I am happy to hand them over in whatever shape suits you.

One question, since it is not covered by #48704 or by the balanced default: is it worth
normalizing or dropping `sourceDir` and `podspecDir` in Expo.ts before `JSON.stringify(config)`?
That would close the gap the measurement above shows, but it changes pnpm users' fingerprints
once, so it needs the same coordination as #48704. I will not open anything for it unless you say
it is wanted.

packages/@expo/fingerprint/src/utils/Path.ts
giaBaoJS: Agreed, and the measurement backs you up. The balanced default already drops the resolved
path for native module sources, since `nativeModuleSourceType: 'package'` hashes them as
`name@version`, so this PR is redundant on that path. I am closing it.

One thing that approach does not cover, in case it is useful: the autolinking `contents`
source still embeds the resolved path. `getExpoAutolinkingAndroidSourcesAsync` relativizes
`project.sourceDir` at Expo.ts:335 and then hashes the whole JSON at Expo.ts:395, with no
preset gate. On a real pnpm app, pinning `supports-color` and changing nothing else moved
all 9 autolinked `sourceDir` values and changed that source's hash from 61d25f4e to f9d5cca9,
with package.json byte identical and each module's `android/` tree unchanged. So excluding
the resolved path from module sources leaves that one behind.

PR #49289, opened 2026-08-24
[cli] Cache the fingerprint runtime version resolved for the dev manifest

### LINKED ISSUE #46415: CLI: dev server recomputes @expo/fingerprint on every manifest request (runtimeVersion fingerprint policy), causing iOS dev-client connection timeouts

giaBaoJS:
# Why

Fixes #46415.

With `runtimeVersion: { policy: "fingerprint" }`, the dev server resolves the runtime version by shelling out to `expo-updates runtimeversion:resolve`, which re-hashes the whole project with `@expo/fingerprint`. That happened on **every** manifest request, with no reuse between requests.

Because the resolver runs in a subprocess (`spawnAsync` in `src/utils/expoUpdatesCli.ts`), no in-process cache in `@expo/fingerprint` can survive between requests, so every request paid the full cold cost. Dev clients fetch the manifest before the bundle, and the iOS launcher times out well before a large project finishes fingerprinting, which is why this shows up as "Metro only works on Android".

I measured this on a blank SDK 57 app with `expo-updates` and the fingerprint policy, against `expo start`:

```
=== SEQUENTIAL: 5 manifest requests, one after another ===
manifest #1: 200 in 1.242849s
manifest #2: 200 in 0.617490s
manifest #3: 200 in 0.645583s
manifest #4: 200 in 0.722984s
manifest #5: 200 in 0.798277s
```

Five requests, five full fingerprints. Requests 2 to 5 never get cheaper, so this is waste on **sequential** requests, not just concurrent bursts. Firing the same five requests concurrently instead makes each one slower, because five fingerprint subprocesses compete for CPU:

```
=== CONCURRENT: 5 manifest requests fired at once ===
concurrent #1: 200 in 1.223848s
concurrent #2: 200 in 1.226151s
concurrent #5: 200 in 1.245720s
concurrent #3: 200 in 1.253635s
concurrent #4: 200 in 1.256338s
```

A blank app is cheap here. The cost scales with project size, which is where the reported timeouts come from.

# How

Cache the resolved runtime version per platform on `ExpoGoManifestHandlerMiddleware`, and clear it whenever Metro's file watcher observes a change.

The cache stores the **pending promise** rather than the resolved value, so concurrent requests share a single subprocess instead of racing several of them. A rejected resolution is removed from the cache so a transient failure is not pinned for the rest of the session. The key is platform only, since `projectRoot` is fixed for the middleware instance.

## On invalidation

A cache that never invalidates would be a bug here, so this is the part worth reviewing.

`ManifestMiddleware` re-reads `getConfig()` on every request today, so app config edits are picked up live. A cache that outlived those edits would serve a fresh `exp` next to a stale `runtimeVersion`. Invalidating from the watcher keeps those consistent.

I followed the precedent already in `MetroBundlerDevServer`, where the API route cache is invalidated from `observeAnyFileChanges` with the same reasoning: we cannot know exactly which files the computation depends on, so invalidate aggressively rather than risk staleness. Over-invalidating only costs a recompute, which is exactly today's behaviour, so the worst case here is no worse than before.

One thing worth flagging, because it contradicts a suggestion in the issue thread. The issue suggests invalidating on watched-file changes because "Metro already watches those". That is only partly true. I attached a probe to Metro's watcher and checked what it actually reports:

- observed: `app.json`, `package.json`, JS sources
- **not** observed: `ios/`, `android/`, or native sources inside `node_modules`

So this covers app config and dependency changes, including installing a native module (which touches `package.json`), but a mid-session `expo prebuild` or a hand edit to `ios/`/`android/` is not observed and would need a dev server restart. That matches how `metro.config.js` and `babel.config.js` are already handled via `FileNotifier` ("Restart the server to see the new results"). I kept the scope here rather than adding a second watcher for the native directories, but I am happy to extend it if you would prefer that.

# Test Plan

Added tests to `ExpoGoManifestHandlerMiddleware-test.ts` covering sequential reuse, concurrent de-duplication, per-platform keying, invalidation, and not caching failures.

```
  _getManifestResponseAsync
    runtime version caching
      ✓ resolves the runtime version once for repeated sequential manifest requests (1 ms)
      ✓ resolves the runtime version once for concurrent manifest requests (1 ms)
      ✓ caches the runtime version per platform (1 ms)
      ✓ resolves the runtime version again after the cache is invalidated (1 ms)
      ✓ does not cache a failed resolution (5 ms)
      ✓ does not re-resolve when expo-updates reports no runtime version (1 ms)

Test Suites: 1 passed, 1 total
Tests:       23 passed, 23 total
```

Reverting only the source change and keeping the new tests fails for the right reason, the resolver running more times than expected:

```
  ● runtime version caching › resolves the runtime version once for repeated sequential manifest requests
    Expected number of calls: 1
    Received number of calls: 3
  ● runtime version caching › resolves the runtime version once for concurrent manifest requests
    Expected number of calls: 1
    Received number of calls: 3
  ● runtime version caching › caches the runtime version per platform
    Expected number of calls: 2
    Received number of calls: 4
  ● runtime version caching › does not re-resolve when expo-updates reports no runtime version
    Expected number of calls: 1
    Received number of calls: 2

Tests:       5 failed, 18 passed, 23 total
```

For completeness: `does not cache a failed resolution` passes with and without the fix, since with no cache there is nothing to pin. It is a guard on the implementation rather than a proof of the fix.

Full package suite, type check, lint, and format:

```
Test Suites: 244 passed, 244 total
Tests:       7 skipped, 1944 passed, 1951 total

$ tsc -p tsconfig.json      # exit 0
$ oxlint --config oxlint.config.mjs .   # exit 0
$ oxfmt --check .
All matched files use the correct format.
```

The existing `MetroBundlerDevServer-test.ts` is unchanged and still passes 13/13.

# Checklist

### DIFF
--- a/packages/@expo/cli/CHANGELOG.md
+++ b/packages/@expo/cli/CHANGELOG.md
@@ -38,6 +38,7 @@
 - Support npm@12's dictionary-based `npm pack --json` format ([#48761](https://github.com/expo/expo/pull/48761) by [@kitten](https://github.com/kitten))
 - Fix wirelessly connected iOS 16 and older devices being omitted from `expo run:ios --device` selection. ([#48127](https://github.com/expo/expo/pull/48127) by [@davellanedam](https://github.com/davellanedam))
 - Fix resolution of ESLint failing in `expo lint` after prerequisites check installs it ([#46223](https://github.com/expo/expo/pull/46223) by [@claritystorm](https://github.com/claritystorm))
+- Cache the runtime version resolved for the dev server manifest, instead of re-running the `expo-updates` fingerprint on every request, and invalidate it when the file watcher observes a change ([#49289](https://github.com/expo/expo/pull/49289) by [@giaBaoJS](https://github.com/giaBaoJS))
 
 ### 💡 Others
 
--- a/packages/@expo/cli/src/start/server/metro/MetroBundlerDevServer.ts
+++ b/packages/@expo/cli/src/start/server/metro/MetroBundlerDevServer.ts
@@ -1484,6 +1484,15 @@ export class MetroBundlerDevServer extends BundlerDevServer {
         );
       }
 
+      // The manifest middleware caches the runtime version resolved by the `expo-updates` CLI,
+      // because resolving a fingerprint policy re-hashes the whole project in a subprocess and
+      // would otherwise run on every manifest request. We can't know exactly which files the
+      // fingerprint covers, so invalidate on any observed change rather than risk serving a
+      // stale runtime version.
+      observeAnyFileChanges({ metro, server }, () => {
+        manifestMiddleware.invalidateRuntimeVersionCache();
+      });
+
       // If React 19 is enabled, then add RSC middleware to the dev server.
       if (isReactServerComponentsEnabled) {
         this.bindRSCDevModuleInjectionHandler();
--- a/packages/@expo/cli/src/start/server/middleware/ExpoGoManifestHandlerMiddleware.ts
+++ b/packages/@expo/cli/src/start/server/middleware/ExpoGoManifestHandlerMiddleware.ts
@@ -18,6 +18,7 @@ import type { ManifestRequestInfo } from './ManifestMiddleware';
 import { ManifestMiddleware } from './ManifestMiddleware';
 import { manifestDebugEvent } from './events';
 import { parseForwardedRequestInfo } from './resolveForwarded';
+import type { RuntimePlatform } from './resolvePlatform';
 import { assertRuntimePlatform, parsePlatformHeader } from './resolvePlatform';
 import { resolveRuntimeVersionWithExpoUpdatesAsync } from './resolveRuntimeVersionWithExpoUpdatesAsync';
 import type { ServerRequest } from './server.types';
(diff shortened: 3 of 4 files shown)


kitten (commented):
See close reason at: https://github.com/expo/expo/pull/48892#issuecomment-5288369676

I don't think refiling with essentially the same approach would change the situation here much

There's no note here on original intention/problem that precedes the issue or a commented post on the original issue, so I'll have to assume for now that it's a "freestanding fix" and that the close reason on the other PR was missed and that this fix isn't motivated by an issue in your project necessarily

giaBaoJS:
You are right on both counts, and I should have handled this differently.

I missed the close reason on #48892. brentvatne's answer there was that this is a tradeoff of the fingerprint runtime version policy, and that he would suggest `appVersion` instead. That is a decision about whether the cost is worth paying at all, and refiling a caching patch does not answer it. I went straight to measuring the cost without first checking whether the team had already settled that question.

On motivation: I am not hitting this in an app of my own. I found #46415 while going through open issues looking for something to work on. I should have said that plainly in the description, and commented on the issue first, rather than opening a PR against a decision that had already been made.

Closing this. The measurements in the description are from a real SDK 57 app and they stay in the thread if anyone picks up #46415 later. The part I would still flag is that `ios/Podfile` and `android/build.gradle` are not reported by Metro's watcher, which I checked with a probe, because that contradicts the invalidation approach suggested in the issue thread. Sorry for the noise.

PR #49293, opened 2026-08-24
[docs] Fix link contrast on inline-code backgrounds

amandeepmittal:
# Why

Fix ENG-26119

# How

Fix link contrast on inline-code backgrounds

# Test Plan

CleanShot 2026-08-24 at 17 09 04@2x

CleanShot 2026-08-24 at 17 10 18@2x

# Checklist

### DIFF
--- a/docs/styles/global.css
+++ b/docs/styles/global.css
@@ -6,17 +6,15 @@
 /* Selectors are doubled to outrank the same-specificity declarations in
    @expo/styleguide/dist/expo-theme.css, which _app.tsx imports after this file. */
 :root:root {
-  /* AA contrast on the inline-code backgrounds, where text-link falls short */
+  /* AA contrast on inline-code and blue3/blue4 backgrounds, where text-link falls
+     short. blue4 clears by 0.02, so do not lighten this. */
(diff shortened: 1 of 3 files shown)


docs/styles/global.css
```diff
@@ -6,13 +6,16 @@
 /* Selectors are doubled to outrank the same-specificity declarations in
    @expo/styleguide/dist/expo-theme.css, which _app.tsx imports after this file. */
 :root:root {
+  /* AA contrast on the inline-code backgrounds, where text-link falls short */
+  --expo-theme-text-link: #0c6bbf;
```
kadikraman, line 11: `--expo-theme-text-link-on-blue` and `--expo-theme-text-link` are the same color now, should we use the same on both?

kadikraman (approved):

docs/styles/global.css
amandeepmittal: Yes they are identical. I was thinking of doing it in a follow-up PR but let me update this PR.

amandeepmittal: Done.

PR #49297, opened 2026-08-24
[router] Refactor router state to be global

Ubax:
# Why

By design of our API, state should be global in expo-router. APIs like `router.push`, `usePathname`, `useRouteInfo` should be global by design. Until now we were storing the state in navigators and then eventually syncing to global root state.

# How

- Moved state to global reducer (`useNavigationTreeReducer.ts`, `BaseNavigationContainer.tsx`)
- Moved action handling to global reducer (`useNavigationTreeReducer.ts`, `BaseNavigationContainer.tsx`, `reduceNavigationTree.ts`)
- Removed stale state concept
- Added deterministic route key generation (`routeKeySeq` from `state` and `stateKeys.ts`)
- Moved most of actions to dispatch via queue
- Added new `dispatchSync` function to dispatch without queue - it should be used only for native event
- Removed nested param navigation (`screen` and `params` params)
- Refactored routeInfo to always derive from global state
- Removed react-navigation APIs

# Test Plan

1. CI
2. router-e2e

# Checklist


packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
```diff
@@ -347,182 +333,33 @@ export function useNavigationBuilder<
...
+  const { getStateForKey, handleAction } = use(NavigationBuilderContext);
+  if (
+    currentState === undefined ||
+    currentState.stale !== false ||
+    typeof currentState.key !== 'string' ||
+    !Array.isArray(currentState.routeNames) ||
+    !isStateValid(currentState)
+  ) {
+    throw new Error(
+      'A navigator received a missing, partial, or incompatible navigation state. Navigation state is created globally now, so the container must provide a complete seeded state for every mounted navigator.'
```
jakub-agent, line 340: [🚨 **critical**] This throws during render with no recovery path, and it is the cause of the red `check-packages` job. Two distinct scenarios reach it:

**1. A navigator rendered inside a screen but not in the route tree.** `SceneView` passes `route.state` down, and a screen with no `_layout` has no seeded child state. `src/views/__tests__/NavigationAwareActivity.test.ios.tsx` › *retains nested navigator state while hidden* fails on exactly this — it renders a plain `createStackNavigator` inside a screen. That test landed in the merge-base commit (#49164) and is untouched here.

**2. A conditional navigator whose router type changes.** Seeded state has no `type`, but `StackRouter`/`TabRouter`/`DrawerRouter` stamp one via `ensureStateType` on the first action, so `isStateValid` then rejects the slice. I reproduced this: a `_layout` rendering `<Stack/>` or `<Tabs/>` from state, navigate once, flip the condition → same throw. The old code re-initialized here. `__tests__/index.test.ios.tsx:474` (*reconciles state when a conditional navigator changes*) does not catch it because both navigators use `MockRouter`, so both slices carry `type: 'test'`.

The `NAVIGATOR_UNMOUNTED` reseed in `useNavigationTreeReducer` cannot help — it runs in a layout effect, after the render has already thrown.

Note this also contradicts the doc comment this PR adds in `routers/types.tsx`: *"A navigator discards state whose type doesn't match its router type."* Discarding and reseeding the slice is the behaviour the comment describes and the one that recovers; throwing is the one that ships a white screen. Whichever you choose, please make it deliberate and add a test that swaps `StackRouter` for `TabRouter` after an action.

packages/expo-router/src/getLinkingConfig.ts
```diff
@@ -6,9 +6,9 @@ import type { State } from './fork/getPathFromState';
 import { getReactNavigationConfig } from './getReactNavigationConfig';
 import { applyRedirects } from './getRoutesRedirects';
 import type { StoreRedirects } from './global-state/router-store';
-import { getInitialURL, getPathFromState, subscribe } from './link/linking';
+import { store } from './global-state/store';
```
jakub-agent: [🚨 **critical**] This new **value** import (not type-only) breaks `@expo/cli`'s `src/export/__tests__/exportStaticAsync.test.ts` on CI with `ReferenceError: __DEV__ is not defined`. The chain is `expo-router/testing-library/mock-config` → `getLinkingConfig` → `global-state/store` → `views/Splash` → `utils/splash` → the `expo` runtime, which needs `__DEV__` and is not available in a plain Node test env.

It is only needed for the `store.getRouteInfo().segments` fallback below, which already carries a TODO to remove it. Threading the segments in from the caller (or reading them through a lazily-required accessor) would keep the linking config free of the store's module graph and unbreak the test. Worth guarding against generally: the linking config is imported by static-render and Node-side tooling, so any runtime dependency it picks up propagates there.

packages/expo-router/src/global-state/resolveNavigationDestination.ts
```diff
@@ -0,0 +1,336 @@
...
+    | undefined
+) {
+  if (result?.state.stale !== false || result.affectedRouteKey === undefined) {
+    return undefined;
+  }
+  return result.state.routes.find((route) => route.key === result.affectedRouteKey);
+}
+
+function markState<T extends NavigationState>(state: T): T & RouteState {
+  return { ...state, __internal__routerActionState: true };
```
jakub-agent, line 347: [🏗️ **design**] This marker is never stripped, so it ends up in committed navigation state. `attachRouteState` stores `action.payload.state` verbatim (`return { ...route, state }`) and `completeExistingState` spreads unknown keys through.

Reproduced with `_layout` + `index` + `(group)/_layout` + `(group)/deep`, then `router.push('/(group)/deep')`:

```json
"state":{"stale":false,"key":"navigator:0-1","routeKeySeq":1,"index":0,
         "routeNames":["deep"],"routes":[{"key":"deep:0-1-0",...}],
         "__internal__routerActionState":true}
```

That tree is public via `useRootNavigationState()` and `navigationRef.getRootState()`, and `useLinking.ts` persists the same object into browser history through `history.replace({ path, state })`, so it survives back/forward restores. The snapshots in `__tests__/push.test.ios.tsx:83` and `initialRouteName.test.ios.tsx:158` already assert it inside `store.state`. It also forces every state comparison through `stripStateMarkers`.

The seeding path already asserts the marker is absent (`createSeededNavigationState.test.ios.ts:87`); the dispatch path has no equivalent. Consume and delete the marker in `attachRouteState` (or in `completeNavigationState`) and add the matching assertion.

packages/expo-router/src/react-navigation/core/BaseNavigationContainer.tsx
```diff
@@ -351,66 +354,33 @@ export function BaseNavigationContainer({
...
     <NavigationContainerRefContext.Provider value={navigation}>
       <NavigationBuilderContext.Provider value={builderContext}>
         <NavigationStateContext.Provider value={context}>
-          <UnhandledActionContext.Provider value={onUnhandledAction ?? defaultOnUnhandledAction}>
+          <RouteInfoContext.Provider value={routeInfo}>
             <EnsureSingleNavigator>
               <ThemeProvider value={theme}>{children}</ThemeProvider>
             </EnsureSingleNavigator>
-          </UnhandledActionContext.Provider>
+            <RoutingQueueDrainer ready={registry.has(state.key)} processIntent={processIntent} />
```
jakub-agent, line 394: [🏗️ **design**] `routingQueue` is a module singleton but the drainer is per-container, so with two containers both drainers subscribe to the same queue and whichever effect runs first consumes everything.

`NAVIGATOR_ACTION` intents survive because they carry a closure bound to their own navigator. `ACTION` intents do not: `useNavigationBuilder.tsx:519` (`ROUTE_NAMES_CHANGED`) and the `goBack`/`popToTop`/link actions in `global-state/router.ts` are handed to whichever `processIntent` won, so container B's action is reduced against container A's tree, `resolveOrigin` finds no matching node, and it is dropped via `warnUnhandledAction`. Sibling root containers are a supported, tested configuration — `__tests__/BaseNavigationContainer.test.ios.tsx:953` *works with state change events in a sibling root container*.

For `ROUTE_NAMES_CHANGED` the loss is permanent rather than transient: `pendingRouteNamesRef` was already set to the new `routeNames`, so the layout effect never re-queues and the navigator keeps rendering `getStateForDeclaredRoutes` against stale `committedState.routeNames`. Nothing clears the queue on unmount either, so intents pending at unmount leak into the next container that mounts — including across Jest tests, where there is no reset.

Scoping the queue per container (via context) or tagging `ACTION` intents with the container that must handle them would close this; `pendingRouteNamesRef` should also be invalidated when the action is dropped or the state key changes.

packages/expo-router/src/global-state/RoutingQueueDrainer.tsx
```diff
@@ -0,0 +1,42 @@
...
+    }
+    for (const intent of routingQueue.drain(intents)) {
+      try {
+        intent.onDispatch?.(intent.metadata);
+        if (intent.type === 'NAVIGATOR_ACTION') {
+          intent.payload.dispatchSync(intent.payload.action);
+        } else {
+          processIntent(intent);
+        }
+      } catch (error) {
```
jakub-agent, line 62: [🏗️ **design**] This cannot catch failures from the global reducer. For `ACTION` and `NAVIGATE_TO_HREF` the only synchronous work here is `intent.onDispatch` plus `reactDispatch`; `useReducer`'s dispatch does not run the reducer eagerly (only `useState` has that path), so `navigationTreeReducer` runs during the next render, outside this handler.

Concrete failure: user code calls `navigation.reset({ index: 0, routes: [{ name: 'home' }] })` — the shape that worked before this PR. It queues as `NAVIGATOR_ACTION`, `dispatchSync` → `handleAction` → `reactDispatch` returns normally, and then `BaseRouter.getStateForAction` throws *"The RESET action payload must contain a complete navigation state"* during render. Nothing above the container catches it (the route-level `ErrorBoundary` sits below), so the tree unmounts — and `drain` has already emptied the batch. `__tests__/RoutingQueueDrainer.test.ios.tsx:116` passes only because the mock `processIntent` throws synchronously, so it asserts a guarantee that does not hold in production.

Given this is a documented breaking change that users will hit while upgrading, surfacing it as an unrecoverable render crash is harsh. Consider warning and returning `null` (unhandled) from the `RESET` guard, or catching where the reducer actually runs. Either way the comment/test here should stop implying the drainer protects against reducer errors.

packages/expo-router/src/layouts/stack-router.ts
```diff
@@ -127,272 +126,151 @@ export const stackRouterOverride: NonNullable<NativeStackNavigatorProps['UNSTABL
...
+            const params = action.payload.merge
+              ? { ...currentRoute.params, ...action.payload.params }
+              : action.payload.params;
+            const route = attachRouteState(
               {
-                key: `${action.payload.name}-${nanoid()}`,
-                name: action.payload.name,
-                path: action.type === 'NAVIGATE' ? action.payload.path : undefined,
+                ...currentRoute,
+                key: minter.mint(action.payload.name),
```
jakub-agent, line 176: [🏗️ **design**] The singular path clones `currentRoute` — including its `state` — under a brand new key, while `currentRoute` itself stays in the stack at line 188 (`activeRoutes.concat(route, preloadedRoutes)`). If `attachRouteState` does not replace the copied state (i.e. the action carries no marked `payload.state`), two simultaneously-live stack entries hold the *same* nested navigator state object, so both child navigators register under the same `state.key` in `RouterRegistry` — a `Map`, so one silently overwrites the other, and `resolveNavigationDestination` will later reduce the wrong subtree.

Concrete path: on `[id]` with `params.id === '1'` where `[id]/` has a `_layout`, call `useNavigation().navigate('[id]', { id: '2' })` (`CommonActions.navigate` sets no `payload.state`). The stack then holds `[id]:0-1` and `[id]:0-2`, both with a child state keyed `navigator:0-1`. Navigations through `router.*`/`<Link>` are safe because the resolver always attaches fresh state, which masks this.

Dropping the stale subtree when minting the new key (`{ ...currentRoute, state: undefined, key: minter.mint(...) }` before `attachRouteState`) makes the invariant explicit.

packages/expo-router/src/ui/TabTrigger.tsx
```diff
@@ -199,25 +201,31 @@ export function useTabTrigger(options: TabTriggerProps): UseTabTriggerResult {
...
+          const owningState =
+            config.contextKey !== contextKey
+              ? navigation?.getParent(config.contextKey)?.getState()
+              : state;
+          if (!owningState) {
+            return;
+          }
+          const action = buildTabAction(config, owningState, registry, options?.resetOnFocus);
+          return navigation?.dispatchSync(
+            config.contextKey !== contextKey ? { ...action, target: owningState.key } : action
```
jakub-agent: [🏗️ **design**] `buildTabAction` may return an action produced by `resolveNavigationDestination`, which already sets `target: current.key` — and `current` is the *descended* navigator when the resolver walked into a child (`resolveNavigationDestination.ts:112`). Overwriting `target` here undoes that descent.

Concrete failure: a `<TabTrigger name="x" href="/a/b">` declared on a parent `<Tabs>` and re-rendered inside a nested navigator (so `config.contextKey !== contextKey`), pressed while tab `a` is already focused. `handleOnPress` fires because `trigger.deep` is true, `buildTabAction` builds a `NAVIGATE`, the resolver descends into `a`'s stack and returns `payload.name === 'b'` with the child's target — then the override retargets it at the tabs navigator, whose `routeNames` do not contain `b`, so `getStateForAction` returns `null` and the press silently does nothing.

`action.target ?? owningState.key` (or only setting the target when the action was not resolved) lets the resolver's decision win. The asymmetry with the `config.contextKey === contextKey` branch, which does honour the resolved target, is itself a smell.

packages/expo-router/src/global-state/resolveNavigationDestination.ts
```diff
@@ -0,0 +1,336 @@
...
+
+function createDestinationState(
+  targetState: PartialState<NavigationState>,
+  routeNode: RouteNode,
+  withAnchor: boolean,
+  internalParams: InternalExpoRouterParams,
+  parentChain: string
+): NavigationState {
+  const targetRoute = getFocusedRoute(targetState);
+  const routeNames = routeNode.children.map((child) => child.route);
```
jakub-agent: [🏗️ **design**] `createDestinationState` builds `routeNames` in raw file-tree order, while every other producer sorts first: `createSeededNavigationState` uses `[...routeNode.children].sort(sortRoutesWithInitial(initialRouteName))` (`createSeededNavigationState.ts:91`), and the navigator derives its own `routeNames` from `useSortedScreens`, which applies the same sort.

`useNavigationBuilder.tsx:514` compares the two with `isArrayEqual`, which is order-sensitive, so state built here for a not-yet-mounted navigator will not match on mount and immediately queues a `ROUTE_NAMES_CHANGED`. For `(group)/_layout` with `unstable_settings.anchor: 'test'` plus `(group)/orange` and `(group)/test` in that file order, this produces `['orange','test']` while the navigator reports `['test','orange']` — an extra queued action and state commit on every first navigation into a nested navigator.

Sorting a copy with `sortRoutesWithInitial(getValidInitialRouteName(routeNode))`, as `createSeededNavigationState` does, keeps the two producers in agreement.

packages/expo-router/src/global-state/reduceNavigationTree.ts
```diff
@@ -0,0 +1,198 @@
...
+}
+
+export function reduceNavigationTree(
+  root: NavigationState,
+  action: NavigationAction,
+  registry: RouterRegistry,
+  {
+    originKey,
+    origin,
+    tree = indexNavigationTree(root),
```
jakub-agent: [💡 suggestion] The `originKey` option and this lazy `tree` default are unreachable in production. The only production caller, `useNavigationTreeReducer.ts:145`, always passes both `origin` and `tree`, so the `origin ??= resolveOrigin(...)` fallback and `indexNavigationTree(root)` exist purely for `__tests__/reduceNavigationTree.test.ios.ts`. `findActionHandler` and `rebuildTreeWithSlice` are also exported with no consumer inside or outside the module.

Per the repo's *aggressively minimal* rule and *prefer asserting via existing APIs over adding test-only surface*: make `origin` and `tree` required, drop `originKey`, and un-export the two helpers — the tests can build the index explicitly.

Minor and separate: `assertCompleteState` is called on every node inside `resolveOrigin` even though `indexNavigationTree` already asserted the whole tree on the way in, so every dispatch re-walks the focused chain for nothing.

packages/expo-router/src/global-state/useNavigationTreeReducer.ts
```diff
@@ -0,0 +1,331 @@
...
+        const message =
+          typeof error === 'object' && error != null && 'message' in error ? error.message : error;
+        // TODO(@ubax): move console side effects out of the reducer.
+        console.warn(
+          `An error occurred when trying to handle navigation action ${JSON.stringify(operation)}: ${message}`
+        );
+        return state;
+      }
+      if (resolution.status === 'invalid') {
+        const invalidHref = operation.payload.originalHref ?? resolution.href;
```
jakub-agent, line 124: [💡 suggestion] The `NAVIGATE_TO_HREF` branch has no unit test. `__tests__/useNavigationTreeReducer.test.ios.tsx` covers only `ACTION` and `NAVIGATOR_UNMOUNTED`, so all three outcomes here are untested: `getNavigateAction` throwing (warn, state unchanged), `status: 'invalid'`, and the recursion into `ACTION`.

`originalHref` in particular is only ever set by `fork/useLinking.native.ts:180`, so the intended behaviour — report the original deep-link URL rather than the redirect-resolved href — would silently regress if that field stopped being propagated, with nothing failing. Worth reducer-level cases for an invalid href both with and without `originalHref`.

packages/expo-router/src/global-state/resolveNavigationDestination.ts
```diff
@@ -0,0 +1,336 @@
...
+  if (!targetRoute || !routeNames.includes(targetRoute.name)) {
```
jakub-agent: [💡 suggestion] When the parsed target route is not a child of `routeNode`, the destination is silently replaced by the navigator's initial state and the requested href is dropped with no diagnostic. `createSeededState` handles the same situation with an explicit dev warning naming the unknown route (*"The initial navigation state contains the unknown route ..."*).

A linking-config/route-tree mismatch therefore surfaces as "navigation silently went to the wrong screen" rather than an actionable message. Adding the same warning here would make the two paths behave consistently.

packages/expo-router/src/global-state/resolveNavigationDestination.ts
```diff
@@ -0,0 +1,336 @@
...
+  const dynamic = matchDynamicName(targetRoute.name);
+  return (
+    !dynamic ||
+    (targetRoute.params as Record<string, unknown> | undefined)?.[dynamic.name] ===
+      (currentRoute.params as Record<string, unknown> | undefined)?.[dynamic.name]
+  );
+}
+
+function findDestinationNode(routeNode: RouteNode, routeName: string | undefined) {
+  return routeName === INTERNAL_SLOT_NAME || routeNode.route === routeName
```
jakub-agent: [💡 suggestion] `findDestinationNode` returns the *parent* node whenever the child route name equals the parent's own `route`. `RouteNode.route` is relative to the nearest `_layout`, so `app/foo/_layout.tsx` + `app/foo/foo/_layout.tsx` yields a child route literally named `foo` under a node whose `route` is also `foo`.

Resolving `/foo/foo/...` then descends with `node` unchanged, and `createResolvedAction` passes that wrong node into `resolveState`/`createDestinationState`, which builds the nested state from the *outer* layout's children — the requested route is not in `routeNames`, so line 234 silently falls back to the outer layout's initial route.

`INTERNAL_SLOT_NAME` is already covered by the first clause, so the `routeNode.route === routeName` clause looks unnecessary here; dropping it or restricting it to the top-level call would remove the ambiguity.

packages/expo-router/src/fork/useLinking.native.ts
```diff
@@ -84,123 +88,104 @@ export function useLinking(
...
-
       const navigation = ref.current;
-      const state = navigation ? getStateFromURLInEffect(url) : undefined;
+      const path = extractExpoPathFromURL(prefixes, url);
+      const state = navigation && path !== undefined ? getStateFromURL(url) : undefined;
 
       if (navigation && state) {
         // If the link were handled, it gets cleared in NavigationContainer
-        onUnhandledLinking(extractExpoPathFromURL(prefixes, url));
+        onUnhandledLinking(path);
```
jakub-agent, line 177: [💡 suggestion] The warm deep-link listener reports the path without a leading slash (`extractExpoPathFromURL` strips it), while the intent queued on the next line normalizes with `path.startsWith('/') ? path : `/${path}``. This PR changed the cold-start side to `getInitialPath` (lines 138/144), which always prepends `/` — so the two paths now disagree where previously both used the unprefixed form.

`NavigationContainer.tsx:120` clears `lastUnhandledLink` only on strict equality with `refContainer.current?.getCurrentRoute()?.path`, which comes from `getStateFromPath('/foo/bar')` and is therefore slash-prefixed. So a warm deep link to `myapp://foo/bar` records `'foo/bar'`, never matches `'/foo/bar'`, and `UNSTABLE_UnhandledLinkingContext` keeps reporting a link that was in fact handled. Cold start now works because both sides use the prefixed form. Using `getInitialPath(prefixes, url)` here too makes them consistent.

packages/expo-router/src/react-navigation/core/BaseNavigationContainer.tsx
```diff
@@ -284,9 +287,9 @@ export function BaseNavigationContainer({
       onReadyRef.current?.();
       emitter.emit({ type: 'ready' });
     }
-  }, [state, isReady, emitter]);
+  }, [state, registry, isReady, emitter]);
 
-  React.useEffect(() => {
+  useClientLayoutEffect(() => {
```
jakub-agent: [💡 suggestion] This block moved from `React.useEffect` to `useClientLayoutEffect`, which also moves the two dev-only whole-tree traversals — `checkSerializable(hydratedState)` and `checkDuplicateRouteNames(hydratedState)` — from after paint to before paint. They recurse through every navigator slice, route and params object on *every* state change, so in a development build each navigation now blocks the commit for the length of that walk; on a deep tab/stack tree with large params that is visible jank which did not exist before.

The reason for the move looks like the `emitter.emit({ type: 'state' })` / `onStateChange` ordering rather than the checks themselves. Keeping the emit and `onStateChange` in the layout effect while leaving the two `process.env.NODE_ENV !== 'production'` checks in a passive effect would preserve the new ordering without the pre-paint cost.

packages/expo-router/src/__tests__/stacks.test.ios.tsx
```diff
@@ -39,7 +40,8 @@ describe('canDismiss', () => {
     expect(router.canDismiss()).toBe(true);
   });
 
-  it('should work within the default Stack', () => {
+  // TODO(ENG-22019): Detect typeless stacks created by the default Stack.
+  it.skip('should work within the default Stack', () => {
```
jakub-agent, line 44: [💡 suggestion] This previously-passing test is disabled under `TODO(ENG-22019)` rather than the accepted `TODO(@ubax)`/ENG-26123 pattern, so it is easy to miss that it represents a user-visible regression.

`canDismiss()` in `global-state/router.ts:109` is byte-identical to `main`, so the change is in the state shape: seeded state carries no `type`, and only navigators that have reduced an action get one stamped via `ensureStateType`. A navigator whose state was created wholesale by the destination resolver stays typeless, and `state.type === 'stack'` fails. I confirmed the typeless case in a probe.

The same assumption is load-bearing in `link/preview/utils.ts:83,97`, `native/useScrollToTop.tsx:74`, `utils/stack.ts:9`, and `global-state/stateUtils.ts:33,60`. If ENG-22019 is an accepted follow-up that is fine, but the behaviour change deserves a CHANGELOG line under known limitations — right now nothing tells an upgrading user that `router.canDismiss()` can start returning `false`.

packages/expo-router/src/react-navigation/routers/TabRouter.tsx
```diff
@@ -506,26 +477,32 @@ export function TabRouter({
...
                     : route.path;
 
-                return params !== route.params || path !== route.path
-                  ? { ...route, key, path, params }
-                  : route;
+                const updatedRoute =
+                  params !== route.params || path !== route.path
+                    ? { ...route, key, path, params }
+                    : route;
+                return attachRouteState(updatedRoute, action);
```
jakub-agent, line 485: [💡 suggestion] When `PUSH`/`NAVIGATE`/`JUMP_TO` re-keys a route because its `getId` changed (line 459), nothing rewrites the old key in `history` — unlike `PRELOAD`, which handles exactly this via `replacedKey` (lines 646-668).

With `backBehavior: 'history'`, state `routes: [{key:'foo:0',name:'foo',params:{id:'one'}}, {key:'bar:1'}]`, `index: 1`, `history: [foo:0, bar:1]`, `navigate('foo', {id:'two'})` yields `routes: [foo:2, bar:1]` but `history: [foo:0, bar:1, foo:2]`. The first `GO_BACK` reaches `bar:1`; the second reads `history[0].key === 'foo:0'`, `findLastIndex` returns `-1`, and `GO_BACK` returns `null` — back navigation is dead from then on.

This predates the PR, but the router was rewritten here and `PRELOAD` already carries the fix, so folding the same `replacedKey` history rewrite into this branch (with a test) is cheap while the code is open.

packages/expo-router/src/layouts/stack-router.ts
```diff
@@ -127,272 +126,151 @@ export const stackRouterOverride: NonNullable<NativeStackNavigatorProps['UNSTABL
...
-              }),
-            };
-          } else {
-            // START FORK
-            const preloadedRouteKey = `${action.payload.name}-${nanoid()}`;
-            const currentPreloadedRoute: (typeof state)['routes'][number] = {
-              key: preloadedRouteKey,
-              name: action.payload.name,
-              params: preloadZoomTransitionId
+          if (affectedIndex > state.index + 1) {
```
jakub-agent, line 247: [💡 suggestion] This splice replaces the old fork's explicit insertion of a new preloaded route directly after the active routes — the documented workaround for screens freezing after link-preview navigation. It is now entirely dependent on `affectedRouteKey` being correct: if the base router ever reported a key that is not in `routes`, `affectedIndex` would be `-1`, the reorder would be skipped silently, and the freeze would come back with no test failing.

`prefetch.test.ios.tsx` only ever preloads a single route, so nothing covers the ordering with two or more preloaded routes. A case that preloads two siblings and asserts the second lands at `state.index + 1` would pin this down.

packages/expo-router/src/navigationParams.ts
```diff
@@ -38,29 +38,45 @@ const internalExpoRouterParamNames = [
...
 export type InternalExpoRouterParams = Partial<Record<InternalExpoRouterParamName, unknown>>;
 
+const paramsObjectsWarnedForScreen = new WeakSet<object>();
+const paramsObjectsWarnedForNestedParams = new WeakSet<object>();
+
+export function warnIfScreenParam(params: object | undefined) {
+  if (
+    process.env.NODE_ENV !== 'production' &&
+    params &&
+    'screen' in params &&
```
jakub-agent, line 48: [🐧 nit] `warnIfScreenParam` fires on any params object containing a key named `screen`, including a legitimate user param. Since `screen` is now an ordinary param, `/settings?screen=wide` routed through the compat layer (`useLinkTo` → `useBuildAction` → `navigation.dispatch` → `handleAction`) prints *"Navigating with a `screen` param is no longer supported... Navigate with a full href instead"* for a URL that already is a full href.

Warning only when the value looks like a nested-navigation payload — `params.screen` is a string and a sibling `params` object is present — or scoping the check to actions that did not originate from href resolution would avoid the false positive.

packages/expo-router/CHANGELOG.md
```diff
@@ -7,10 +7,18 @@
 - The `freezeOnBlur` screen option is disabled and screens are never frozen when blurred. ([#49163](https://github.com/expo/expo/pull/49163) by [@Ubax](https://github.com/Ubax))
 - Remove the experimental web modal implementation. See the [web modals guide](https://docs.expo.dev/router/advanced/web-modals/). ([#49204](https://github.com/expo/expo/pull/49204) by [@Ubax](https://github.com/Ubax))
 - Remove `NavigationIndependentTree` and `useNavigationIndependentTree` from `expo-router/react-navigation`. To embed an isolated navigation tree in a screen, use `@react-navigation/native`. ([#49172](https://github.com/expo/expo/pull/49172) by [@Ubax](https://github.com/Ubax))
+- Generate deterministic navigation state and route keys, and require `routeKeySeq` in complete navigation states.
```
jakub-agent: [🐧 nit] Nine of the new entries (lines 10-15, 18-19, 64) are missing the `([#PR](link) by [@user](link))` attribution suffix that every other entry in this file carries, and the entry at line 83 (*"Seed complete navigation state from the initial deep link."*) still has the literal `#TODO` placeholder link. Worth filling in before merge.

On content: I verified the five removed public symbols (`getActionFromState`, `LinkingOptions.getActionFromState`, `NavigatorScreenParams`, `resetRoot`, `Router.getRehydratedState`) are all documented, and found no stale consumers anywhere in `apps/`, `packages/`, or `docs/`. That part is clean.

jakub-agent (changes requested):

packages/expo-router/src/global-state/resolveNavigationDestination.ts
Ubax: Add the warning

packages/expo-router/src/global-state/resolveNavigationDestination.ts
Ubax: Fix it

packages/expo-router/src/fork/useLinking.native.ts
Ubax: Fix it

packages/expo-router/src/react-navigation/core/BaseNavigationContainer.tsx
Ubax: DO it

packages/expo-router/src/__tests__/stacks.test.ios.tsx
Ubax: Ignore

packages/expo-router/src/react-navigation/routers/TabRouter.tsx
Ubax: Leave a TODO comment

packages/expo-router/src/layouts/stack-router.ts
Ubax: Add one test case for this

packages/expo-router/src/ui/TabTrigger.tsx
Ubax: Fix it

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
```diff
@@ -341,188 +327,36 @@ export function useNavigationBuilder<
...
-        : nextState;
+  const { getStateForKey, resetNavigator, handleAction } = use(NavigationBuilderContext);
+  if (
+    currentState === undefined ||
+    currentState.stale !== false ||
+    typeof currentState.key !== 'string' ||
+    !Array.isArray(currentState.routeNames)
+  ) {
+    throw new Error(
+      'A navigator received a missing, partial, or incompatible navigation state. Navigation state is created globally now, so the container must provide a complete seeded state for every mounted navigator.'
```
jakub-agent, line 340: [🏗️ **design**] The router-type half of this is properly fixed — nice solution. The other half is now a deliberate capability removal, and I think it needs to be stated as one.

`NavigationAwareActivity.test.ios.tsx` › *retains nested navigator state while hidden* was deleted rather than made to pass. I re-probed it: rendering a `createStackNavigator` from `expo-router/react-navigation` inside a screen still throws here, because `SceneView` hands down `route.state` and a screen with no `_layout` has none.

That is a defensible call — it lines up with removing `NavigationIndependentTree` in #49172 — but as it stands:

1. **No CHANGELOG entry.** Every other breaking change in this PR has one. A user with a working nested navigator gets a runtime crash with no upgrade note.
2. **The message misdirects.** *"the container must provide a complete seeded state for every mounted navigator"* describes an expo-router invariant, so it reads as an internal bug. The user's actual problem is "this navigator is not part of the route tree". Something closer to the #49172 wording — *use `@react-navigation/native` for a navigator that is not part of the route tree* — would point them at the fix.
3. **The coverage is just gone.** The deleted test was the only assertion that `NavigationAwareActivity` (itself new, from #49164) preserves nested navigator state. If the intent is now "this throws", a test asserting the throw would pin the decision down instead of leaving a silent gap.

Happy to be told this is tracked as a follow-up — I just don't want it to land undocumented.

packages/expo-router/CHANGELOG.md
```diff
@@ -4,13 +4,22 @@
...
+- Queue `navigation.dispatch` and navigation helper actions until after commit, add synchronous `navigation.dispatchSync`, and remove functional dispatch actions.
+- Require complete state in `CommonActions.reset` and remove `Router.getRehydratedState` from `expo-router/react-navigation`.
+- Remove `resetRoot` from `NavigationContainerRef` and `createNavigationContainerRef`.
+- Remove `NavigatorScreenParams`, `getActionFromState`, and `LinkingOptions.getActionFromState` from `expo-router/react-navigation`.
+- Represent nested navigation only as navigation state. The `screen`, `params`, and `initial` params are now ordinary user params
+- Stop copying ancestor route params into descendant routes during imperative navigation, matching cold-start deep-link behavior.
+- Change custom routers' `getStateForAction` return value to `{ state, affectedRouteKey }`.
+- The `freezeOnBlur` screen option is disabled and screens are never frozen when blurred. ([#49164](https://github.com/expo/expo/pull/49163) by [@Ubax](https://github.com/Ubax))
```
jakub-agent: [🐧 nit] Three small things from the reshuffle — the `#TODO` placeholder is gone, thanks.

1. This line's link text and URL disagree: `[#49164](https://github.com/expo/expo/pull/49163)`. It was `#49163` → `/49163` before the reorder.
2. Two entries lost user-facing detail. *"Route deep links and browser history through the navigation queue while preserving saved navigation state"* was dropped from 🐛 Bug fixes entirely, and the `ROUTER_LINK` → `NAVIGATE_TO_HREF` entry (line 84) was truncated so it no longer mentions that `linkTo('..')` defers a queued `GO_BACK` or that `goBack()` no longer throws when the container is not ready. I checked both against `main` — neither exists there, so this PR introduces both behaviours and they are now undocumented.
3. The entries this PR adds still have no `([#PR](link) by [@user](link))` suffix. Grouping them at the top of each section makes them easy to spot, so this is purely a fill-in-before-merge item.

jakub-agent (approved):

PR #49302, opened 2026-08-24
[cli][server] Fix CorsMiddleware local-hostname regex allowing non-loopback origins

### LINKED ISSUE #49301: CorsMiddleware: unescaped dot in _isLocalHostname regex lets non-loopback hostnames bypass the dev server's origin check
### Summary

`_isLocalHostname` in `packages/@expo/cli/src/start/server/middleware/CorsMiddleware.ts` is supposed to match only `localhost`, `::1`, and IPv4 addresses in the `127.0.0.0/8` range (in `a.b.c.d` dotted form, optionally prefixed with `::ffff:`). Due to an unescaped `.` in its regex, it actually accepts any hostname that starts with `127` followed by three `<any-char><digits>` groups, dots not required, e.g. `127a1b1c1` or `127x999x999x999`. This function backs `isLocalhost` inside `createCorsMiddleware`, which is OR'd into the check that lets a cross-origin request past the dev server's CORS/origin guard, so the bug weakens that guard for a class of non-loopback hostnames.

```js
// packages/@expo/cli/src/start/server/middleware/CorsMiddleware.ts:40
/^127(?:.\d+){3}$/.test('127a1b1c1') // -> true (the `.` is a wildcard, not a literal dot)
```

Reproduced directly against the function body in `node`:

```js
$ node -e "
const isLocal = (hostname) => {
  if (hostname === 'localhost') return true;
  let maybeIp = hostname;
  const ipv6To4Prefix = '::ffff:';
  if (maybeIp.startsWith(ipv6To4Prefix)) maybeIp = maybeIp.slice(ipv6To4Prefix.length);
  if (maybeIp === '::1') return true;
  else if (/^127(?:.\d+){3}\$/.test(maybeIp)) {
    return maybeIp.split('.').every((part) => {
      const num = parseInt(part, 10);
      return num >= 0 && num <= 255;
    });
  } else return false;
};
console.log(isLocal('127a1b1c1'));            // true  -- should be false
console.log(new URL('http://127a1b1c1:8081').hostname); // '127a1b1c1' -- a valid Origin hostname
"
true
127a1b1c1
```

### Root cause

Two compounding issues in `_isLocalHostname` (lines 29–48):

1. The loopback-range regex, `/^127(?:.\d+){3}$/`, uses an unescaped `.` where a literal dot was clearly intended (should be `/^127(?:\.\d+){3}$/`). As written, `.` matches any single character, so the three groups don't require dots at all, just "some character then digits", repeated three times.
2. Even when the regex does match a dot-free string, the follow-up octet validation (`maybeIp.split('.').every(...)`) silently degrades instead of catching it: splitting a string with no `.` in it just returns the whole string as a single array element, and `parseInt(part, 10)` truncates at the first non-digit character rather than failing. So `127a1b1c1`.split('.') is `['127a1b1c1']`, and `parseInt('127a1b1c1', 10)` is `127`, which passes the `0 <= num <= 255` check.

Net effect: any hostname matching `127<char><digits><char><digits><char><digits>` is classified as local, even though it's neither `127.0.0.0/8` in dotted form nor any other loopback representation.

### Expected behavior

`_isLocalHostname('127a1b1c1')` (and similar dot-free strings) should return `false`, matching how the sibling test cases already handle malformed dotted IPs like `127.999.999.999`.

### Actual behavior

`_isLocalHostname('127a1b1c1')` returns `true`.

### Why this matters

`_isLocalHostname` feeds `isLocalhost` in `createCorsMiddleware` (`CorsMiddleware.ts` ~line 66), which is OR'd into `isAllowedHost`:

```js
const isLocalhost = _isLocalHostname(hostname);
const isAllowedHost = allowedHosts.includes(host) || isLocalhost;
if (!isSameOrigin && !isAllowedHost) {
  next(new Error(`Unauthorized request from ${req.headers.origin}. ...`));
  return;
}
```

That rejection is the dev server's only defense against arbitrary cross-origin requests (its own error message calls out this exact purpose). A page served from an origin whose hostname happens to match the loose pattern above bypasses the rejection, even though it isn't actually the local dev server or loopback address.

The existing unit tests in `CorsMiddleware-test.ts` (`describe(_isLocalHostname, ...)`) only exercise dotted strings (`127.0.0.1`, `127.999.999.999`, etc.), so this dot-free bypass isn't caught by current coverage.

### Suggested fix

Escape the dot in the regex (`/^127(?:\.\d+){3}$/`), which on its own is enough to make the subsequent `split('.')`/`parseInt` octet check operate only on genuinely dotted input.

### Environment

- expo/expo `main` branch (current as of this report)
- File: `packages/@expo/cli/src/start/server/middleware/CorsMiddleware.ts`, lines 29-48

Rakshit-gen:
# Why

Fixes #49301.

`_isLocalHostname` in `packages/@expo/cli/src/start/server/middleware/CorsMiddleware.ts` is meant to match only `localhost`, `::1`, and IPv4 addresses in `127.0.0.0/8` (dotted-quad form, optionally `::ffff:`-prefixed). Its regex, `/^127(?:.\d+){3}$/`, has an unescaped `.` where a literal dot between octets was intended, so `.` matches any character instead. The follow-up validation (`maybeIp.split('.').every(part => parseInt(part, 10) <= 255 ...)`) doesn't catch this either: splitting a dot-free string just returns the whole string as one element, and `parseInt` truncates at the first non-digit rather than failing.

Net effect: any hostname shaped like `127<char><digits><char><digits><char><digits>` (e.g. `127a1b1c1`, no dots required) is classified as local:

```js
_isLocalHostname('127a1b1c1') // true, should be false
```

This function backs `isLocalhost` in `createCorsMiddleware`, which is OR'd into the check that lets a cross-origin request past the dev server's `Unauthorized request` rejection — its whole stated purpose is to block non-dev-server origins. A hostname matching the loose pattern above bypasses that rejection even though it isn't the local dev server or a loopback address.

# How

Escaped the dot: `/^127(?:\.\d+){3}$/`. That alone is enough to make the subsequent `split('.')`/`parseInt` octet check operate only on genuinely dotted input, since the regex now requires 3 literal-dot-separated digit groups before that code ever runs.

# Test Plan

Added a red/green test in `CorsMiddleware-test.ts` (`rejects dot-free hostnames that merely start with 127`) covering `127a1b1c1` and `127x999x999x999`. Confirmed it fails against the pre-fix regex and passes after the fix; all existing `_isLocalHostname` and `createCorsMiddleware` tests (dotted IPs, IPv6, `localhost`, devtools, forwarded-host cases) still pass.

# Checklist

### DIFF
--- a/packages/@expo/cli/CHANGELOG.md
+++ b/packages/@expo/cli/CHANGELOG.md
@@ -24,6 +24,7 @@
 
 ### 🐛 Bug fixes
 
+- Fix `CorsMiddleware`'s local-hostname check to require a literal dot between IPv4 octets, instead of an unescaped regex wildcard that let non-loopback hostnames starting with `127` (e.g. `127a1b1c1`) bypass the dev server's cross-origin request guard.
 - Serve relative manifest URLs only when the client itself sends the RFC 7239 `Forwarded` header, so that proxied requests from clients without relative-URL support, like released Expo Go versions through the WS tunnel, keep absolute URLs. ([#48997](https://github.com/expo/expo/pull/48997) by [@expo-bot](https://github.com/expo-bot))
 - Fail when `--private-key-path` is passed without `updates.codeSigningCertificate` in the resolved app config, instead of ignoring the flag and continuing without signing.
 - Show the Xcode build log path when `run:ios` fails. ([#48624](https://github.com/expo/expo/pull/48624) by [@ramonclaudio](https://github.com/ramonclaudio))
--- a/packages/@expo/cli/src/start/server/middleware/CorsMiddleware.ts
+++ b/packages/@expo/cli/src/start/server/middleware/CorsMiddleware.ts
@@ -37,7 +37,7 @@ export const _isLocalHostname = (hostname: string) => {
   }
   if (maybeIp === '::1') {
     return true;
-  } else if (/^127(?:.\d+){3}$/.test(maybeIp)) {
+  } else if (/^127(?:\.\d+){3}$/.test(maybeIp)) {
     return maybeIp.split('.').every((part) => {
       const num = parseInt(part, 10);
       return num >= 0 && num <= 255;
--- a/packages/@expo/cli/src/start/server/middleware/__tests__/CorsMiddleware-test.ts
+++ b/packages/@expo/cli/src/start/server/middleware/__tests__/CorsMiddleware-test.ts
@@ -220,4 +220,9 @@ describe(_isLocalHostname, () => {
     expect(_isLocalHostname('localghost')).toBe(false);
     expect(_isLocalHostname('example.com')).toBe(false);
   });
+
+  it('rejects dot-free hostnames that merely start with 127', () => {
+    expect(_isLocalHostname('127a1b1c1')).toBe(false);
+    expect(_isLocalHostname('127x999x999x999')).toBe(false);
+  });
 });


Rakshit-gen:
@kitten tagging you since you wrote this function and the surrounding CORS logic (e.g. #41542, #48267) — would appreciate a look when you have time.

Issue #49111, opened 2026-08-19
[Android][Windows] Large Metro multipart bundle fails in BundleDownloader with invalid HTTP chunk size


sebestyn:
### Minimal reproducible example

https://github.com/sebestyn/expo-multipart-windows-repro

## Reproduction steps

1. On Windows, clone the minimal reproduction repository.
2. Install dependencies:

   ```powershell
   pnpm install
   ```

3. Generate the deterministic 30 MiB JavaScript payload:

   ```powershell
   pnpm generate
   ```

4. Start an Android API 35 x86_64 emulator.
5. Build and run the Android development build with pnpm:

   ```powershell
   pnpm android -- --device <ANDROID_DEVICE_NAME>
   ```

6. Wait for Metro to finish bundling.
7. Observe that the app remains on a white screen and inspect Logcat.

This occurs on Android in a development build. The package manager is pnpm 10.34.5.

## Expected behavior

After Metro finishes bundling, the Android development build should download and execute the JavaScript bundle. The app should display the generated payload size.

## Actual behavior

Metro finishes bundling successfully, but the application remains on a white screen. Logcat reports an invalid HTTP chunk size while React Native's `BundleDownloader` processes the multipart response:

```text
java.net.ProtocolException: Expected leading [0-9a-fA-F] character but was 0xd
    at okhttp3.internal.http1.Http1ExchangeCodec$ChunkedSource.readChunkSize(Http1ExchangeCodec.kt:436)
    at okhttp3.internal.http1.Http1ExchangeCodec$ChunkedSource.read(Http1ExchangeCodec.kt:408)
    at okhttp3.internal.connection.Exchange$ResponseBodySource.read(Exchange.kt:276)
    at okio.RealBufferedSource.read(RealBufferedSource.kt:215)
    at com.facebook.react.devsupport.MultipartStreamReader.readAllParts(MultipartStreamReader.kt:83)
    at com.facebook.react.devsupport.BundleDownloader.processMultipartResponse(BundleDownloader.kt:187)
    at com.facebook.react.devsupport.BundleDownloader.access$processMultipartResponse(BundleDownloader.kt:35)
    at com.facebook.react.devsupport.BundleDownloader$downloadBundleFromURL$1.onResponse(BundleDownloader.kt:133)
```

The reported invalid byte can vary, but the exception and call path remain the same.

## Reproduction controls

- A 1 MiB generated payload loads successfully and Logcat contains `Running "main"`.
[...]

PR #49305, opened 2026-08-24
[cli][server] Fix Content-Length header in JS inspector middleware for non-ASCII response bodies

Rakshit-gen:
# Why

`createJsInspectorMiddleware`'s `GET` handler (`packages/@expo/cli/src/start/server/middleware/inspector/createJsInspectorMiddleware.ts`) serves inspector app metadata as JSON and sets:

```ts
const data = JSON.stringify(app);
res.writeHead(200, {
  'Content-Type': 'application/json; charset=UTF-8',
  'Cache-Control': 'no-cache',
  'Content-Length': data.length.toString(),
});
res.end(data);
```

`data.length` is the JS string's length in UTF-16 code units, not its size in bytes. `res.end(data)` writes the string UTF-8-encoded (the default for Node's `http` module and consistent with the declared `charset=UTF-8`). Any multi-byte character in the response — e.g. a connected device's name with an accent or emoji, which is metadata from the client, not something Expo CLI controls — makes the UTF-8 byte length larger than the declared `Content-Length`:

```js
const data = JSON.stringify({ title: 'café ☕ App' });
data.length                        // 22 (UTF-16 code units)
Buffer.byteLength(data, 'utf8')    // 25 (actual bytes written)
```

An undersized `Content-Length` is invalid HTTP framing: the client either truncates the body before the trailing bytes or, on a persistent connection, mistakes the leftover bytes for the start of the next response. This surfaced while investigating an unrelated HTTP-framing report in #49111, where it was noted in passing as a separate, pre-existing defect in this file — this PR addresses that specific issue.

# How

Use `Buffer.byteLength(data, 'utf8')` instead of `data.length` to compute `Content-Length`.

# Test Plan

Added a red/green test (`should set Content-Length to the byte length of the response, not its UTF-16 string length`) in `createJsInspectorMiddleware-test.ts`, using an app fixture with a non-ASCII `deviceName`. Confirmed it fails against `data.length` and passes with `Buffer.byteLength`; all other tests in the file still pass.

# Checklist

### DIFF
--- a/packages/@expo/cli/CHANGELOG.md
+++ b/packages/@expo/cli/CHANGELOG.md
@@ -24,6 +24,7 @@
 
 ### 🐛 Bug fixes
 
+- Fix `createJsInspectorMiddleware`'s `Content-Length` header to reflect the UTF-8 byte length of the response, instead of its UTF-16 string length, which undersized the header for any inspector app metadata (e.g. a device name) containing non-ASCII characters.
 - Serve relative manifest URLs only when the client itself sends the RFC 7239 `Forwarded` header, so that proxied requests from clients without relative-URL support, like released Expo Go versions through the WS tunnel, keep absolute URLs. ([#48997](https://github.com/expo/expo/pull/48997) by [@expo-bot](https://github.com/expo-bot))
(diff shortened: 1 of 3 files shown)


Rakshit-gen:
@kitten tagging you since you added the origin/security checks in this middleware (#45863) — would appreciate a look when you have time.

PR #49306, opened 2026-08-24
[ui] Reference the TextInputState declaration so consumers typecheck

intergalacticspacehighway:
# Why

Expo router is failing at check packages step - https://github.com/expo/expo/actions/runs/32749280254/job/97502111587

```
Error: expo-router#typecheck: command (/home/runner/work/expo/expo/packages/expo-router) /home/runner/setup-pnpm/node_modules/.bin/bin/pnpm run typecheck exited (2)
 ERROR  expo-router#typecheck: command (/home/runner/work/expo/expo/packages/expo-router) /home/runner/setup-pnpm/
```

Caused by https://github.com/expo/expo/pull/48788

React Native ships `Libraries/Components/TextInput/TextInputState` as JS only, so the deep import needs the ambient declaration in `src/ts-declarations`. That declaration reached expo-ui's own program only through its `include` glob, and a `declare module` block applies solely inside the program that contains it.

`customConditions: ["expo-source"]` makes every monorepo consumer resolve `@expo/ui` to source, so expo-router compiled `keyboard/textInputState.ts` without ever pulling the declaration in, and failed with TS7016.

# How

Reference the declaration from the module that needs it, the way `expo-modules-core/src/NativeViewManagerAdapter.native.tsx` does, so the types travel with the import instead of relying on the owning package's tsconfig.

# Test Plan

Check packages passes

# Checklist

### DIFF
--- a/packages/expo-ui/src/keyboard/textInputState.ts
+++ b/packages/expo-ui/src/keyboard/textInputState.ts
@@ -1,1 +1,2 @@
+/// <reference path="../ts-declarations/react-native-text-input-state.d.ts" />
 export { default as TextInputState } from 'react-native/Libraries/Components/TextInput/TextInputState';


amandeepmittal (approved):
Thank you for fixing this!

I learned about it because the issue this PR fixes, and right now is blocking the docs data generation and its CI. We need to merge it soon.

Created a PR of my own but learned you have created it first: https://github.com/expo/expo/pull/49309, so closing mine now.

PR #49308, opened 2026-08-24
fix(dev-launcher): [iOS] stop Bonjour discovery when an app launch starts

### LINKED ISSUE #49307: [expo-dev-launcher][iOS] Bonjour discovery keeps running during an explicit-URL app launch
### Summary

`expo-dev-launcher`'s `_expo._tcp` Bonjour browse keeps running while an app is being loaded from a
URL the launcher already has — the `?url=` deep link `expo run:ios` opens, or the launcher's manual
URL entry. The browse only feeds the "Development servers" list in the launcher UI, which is gone
the moment a launch starts, so during a launch it is resolving on every interface for no consumer.

Two reasons it isn't stopped, both in `ios/SwiftUI/DevLauncherViewModel.swift`:

- `startServerDiscovery()` guards only on `browser != nil` — not on a launch being in flight, nor on
  an app already running.
- A launch started from a deep link or `--initialUrl` never passes through the view model at all:
  `EXDevLauncherController.onDeepLink:` calls `loadApp:` directly. So on a cold deep-link launch the
  launcher UI has already started a browse, and nothing tells it to stop.

Measured on a stock SDK 57 dev client (iOS 26.4 simulator, Metro on loopback), 3 runs, time the
browse ran concurrently with the launch: **0.452s / 0.236s / 0.420s**.

From one run's unified log — the browse is created, starts a query on every interface, and is still
going when the launch's own requests begin:

```
22:24:29.112  [B1] create with <nw_browse_descriptor bonjour _expo._tcp.> 
22:24:29.769  [B1] start query record for repro-app._expo._tcp.local. on en0
              (also bridge100, bridge101, en9)
22:24:32.174  [B1] state changed: ready -> cancelled
22:24:32.269  first request of the launch
22:24:36.025  [RCTMultipartDataTask] GET http://127.0.0.1:8090/index.ts.bundle?platform=ios&...
```

On loopback this is cheap and the launch is not measurably slower for it. On a LAN address the
resolver's query timer does fire — `nw_resolver_start_query_timer_block_invoke Query fired: did not
receive all answers in time`, in 2 of 8 runs here — but measured against the launch it turns out to
be **concurrent work, not latency**: in those runs it fired ~1.6s *after* the bundle GET, and
launch → bundle is indistinguishable with and without a fix (1.26–2.21s vs 1.35–2.41s over 16 runs).
An earlier "~2s per launch" figure I quoted came from a trace of a much larger app where the timeout
completed before the bundle request; I have no measurement that establishes the launch was waiting
on it, so please read this report as "discovery runs when it has nothing to serve" and nothing more.

### Reproduction

https://github.com/saileshbro/repro-dev-launcher-49249 — `repro-bonjour.sh` captures the simulator's unified log across a deep-link launch and
prints the Bonjour activity next to the first bundle request.

```bash
cd repro-app && npx expo run:ios      # stock SDK 57 app + expo-dev-client
npx expo start --dev-client --port 8090
./repro-bonjour.sh <udid> com.saileshbro.repro-app exp+repro-app 127.0.0.1:8090 reproapp
```

### Expected

Discovery stops when a launch starts, from whichever entry point, and doesn't start back up while an
app is loading or running — it can resume when the launcher UI reappears, which is the only place
its results are shown.

### Environment

- macOS 26.6.1, Xcode 26, iOS 26.4 simulator
- Expo SDK 57, `expo-dev-launcher@57.0.15`, stock `create-expo-app` template
- iOS only — I haven't looked at the Android discovery path.

saileshbro:
# Why

Fixes #49307.

The launcher's `_expo._tcp` browse populates the "Development servers" list in the launcher UI.
Once an app launch begins, it has no consumer — but nothing stops it:

- `startServerDiscovery()` guards only on `browser != nil`, not on a launch being in flight or an
  app already running.
- A launch from a deep link or `--initialUrl` never goes through `DevLauncherViewModel` at all —
  `onDeepLink:` calls `EXDevLauncherController.loadApp` directly — so on a cold deep-link launch the
  launcher UI has already started a browse and nothing tells it to stop.

So the browse keeps resolving, on every interface, alongside a launch whose server address is
already known.

# How

- `loadApp:` posts `EXDevLauncherAppLoadingDidStartNotification`; `DevLauncherViewModel` observes it
  and cancels discovery. Driving this from the launch itself covers every entry point (deep link
  cold and warm, `--initialUrl`, launcher UI), which a check inside `openApp(url:)` does not.
- `startServerDiscovery()` additionally returns early while a launch is in flight or an app is
  running. It restarts when the launcher UI reappears, which is the only place the results are shown.
- `EXDevLauncherController` exposes `isLoadingApp`, mirroring the existing `isAppRunning`.

# Test Plan

Stock SDK 57 app + `expo-dev-client@57.0.15` on an iOS 26.4 simulator, Metro on loopback, launched
by deep link, measured from the simulator's unified log. Repro and script:
https://github.com/saileshbro/repro-dev-launcher-49249 (`repro-bonjour.sh`). 3 runs each:

| build | browse running during the launch | launch → bundle GET |
| - | - | - |
| stock | 0.452s / 0.236s / 0.420s | 0.951 / 0.878 / 0.919s |
| this PR | **0.000s / 0.000s / 0.000s** (cancelled ~3ms before the launch's first request) | 1.057 / 0.933 / 1.037s |

Being explicit about what this does and doesn't buy: **the launch is not measurably faster**, on
loopback or on a LAN. The change removes work that has no consumer.

On a LAN address (`--host lan`, 4 runs per cell, same two builds) the mDNS resolver's query timer —
`nw_resolver_start_query_timer_block_invoke Query fired: did not receive all answers in time` —
fires in 2 of 8 stock runs and 0 of 8 patched runs, and on the cold `--initialUrl` path the patched
build never creates a browse at all. But in the runs where it fired it fired ~1.6s *after* the
bundle GET, i.e. concurrent with the launch rather than blocking it, and launch → bundle is
indistinguishable between the builds (1.26–2.21s stock vs 1.35–2.41s patched over 16 runs). An
earlier "~2s per launch" figure I quoted came from a trace of a much larger app where the timeout
completed before the bundle request; I have no measurement that establishes the launch was waiting
on it, so please read this as removing concurrent work, not latency.

Also checked: returning to the launcher restarts discovery and the dev server still appears in the
list.

**Not covered:** Android — this is iOS-only and I haven't looked at whether the Android discovery
path has the same behaviour.

### DIFF
--- a/packages/expo-dev-launcher/CHANGELOG.md
+++ b/packages/expo-dev-launcher/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 ### 🐛 Bug fixes
 
+- [iOS] Stop `_expo._tcp` discovery when an app launch starts, and don't start it while an app is loading or running, so the browse no longer runs alongside a launch that already knows the server address. ([#49308](https://github.com/expo/expo/pull/49308) by [@saileshbro](https://github.com/saileshbro))
 - [Android] Fix SIGABRT when loading a development server URL that has no path (e.g. `http://192.168.1.2:8081`): Android's `URI.resolve` omits the path separator for an empty-path base, so the first segment of a relative `bundleUrl` was spliced onto the port and the resulting authority failed to parse. ([#48625](https://github.com/expo/expo/pull/48625) by [@tsapeta](https://github.com/tsapeta))
 - [iOS] Use `RCTPlatformName` instead of hardcoding `ios` when requesting bundles from Metro. ([#46443](https://github.com/expo/expo/pull/46443) by [@gabrieldonadel](https://github.com/gabrieldonadel))
 - [Android] Fixed a crash when cold-launching a development build from a deep link that carries intent categories (e.g. an App Link opened from a browser). ([#46314](https://github.com/expo/expo/pull/46314) by [@lilianchiassai-fc](https://github.com/lilianchiassai-fc) & [#46328](https://github.com/expo/expo/pull/46328) by [@lukmccall](https://github.com/lukmccall))
--- a/packages/expo-dev-launcher/ios/EXDevLauncherController.h
+++ b/packages/expo-dev-launcher/ios/EXDevLauncherController.h
@@ -24,6 +24,9 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+/** Posted when a `loadApp:` launch begins, from any entry point (deep link, initial URL, launcher UI). */
+extern NSNotificationName const EXDevLauncherAppLoadingDidStartNotification;
+
 @class EXAppContext;
 @class EXDevLauncherInstallationIDHelper;
 @class EXDevLauncherPendingDeepLinkRegistry;
@@ -78,6 +81,9 @@ NS_ASSUME_NONNULL_BEGIN
 - (nullable NSURL *)appManifestURLWithFallback;
 
 - (BOOL)isAppRunning;
+
+/** YES while a `loadApp:` launch is in flight. */
+- (BOOL)isLoadingApp;
 
 - (UIWindow * _Nullable)currentWindow;
 
--- a/packages/expo-dev-launcher/ios/EXDevLauncherController.m
+++ b/packages/expo-dev-launcher/ios/EXDevLauncherController.m
@@ -44,6 +44,7 @@ static const NSTimeInterval EXDevLauncherDefaultRequestTimeout = 10.0;
 @property (nonatomic, strong) NSURL *manifestURL;
 @property (nonatomic, strong) NSURL *possibleManifestURL;
 @property (nonatomic, strong) EXDevLauncherErrorManager *errorManager;
+@property (nonatomic, assign) BOOL isLoadingApp;
 @property (nonatomic, strong) EXDevLauncherInstallationIDHelper *installationIDHelper;
 @property (nonatomic, strong, nullable) EXDevLauncherNetworkInterceptor *networkInterceptor;
 @property (nonatomic, strong) DevLauncherViewController *devLauncherViewController;
@@ -55,6 +56,8 @@ static const NSTimeInterval EXDevLauncherDefaultRequestTimeout = 10.0;
 
 @end
 
+
+NSNotificationName const EXDevLauncherAppLoadingDidStartNotification = @"EXDevLauncherAppLoadingDidStart";
 
 @implementation EXDevLauncherController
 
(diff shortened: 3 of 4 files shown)


saileshbro:
Context: this came out of the investigation on #49250, which I'd misattributed a multi-second cold-start cost to. That cost is here — the browse — not there. The two changes are independent; this one is the measured half.

saileshbro:
Ran it on my LAN. Short answer: **the change removes the browse and its resolver timeout entirely —
but in these runs the timeout was not on the launch's critical path, so launch time is unchanged.**
That is a correction to my own framing, so let me give the numbers rather than the summary.

Setup: same stock SDK 57 app and the same two builds as the PR (differing only by this patch, no
`expo-updates`, so #49250 can't confound it), Metro bound to the LAN address
(`expo start --dev-client --port 8090 --host lan`), iOS 26.4 simulator, timings from the unified log.
`measure-launch.sh` in https://github.com/saileshbro/repro-dev-launcher-49249. 4 runs per cell.

**Deep link (`simctl openurl`), LAN address**

| | browse during launch | launch → bundle |
| - | - | - |
| stock | 1.341 / 0.755 / 0.854 / 0.846s | 1.920 / 1.296 / 1.332 / 1.256s |
| this PR | 0.000 ×4 (cancelled ~10ms before the launch's first request) | 1.689 / 2.327 / 1.414 / 1.349s |

**Cold launch (`--args --initialUrl`), LAN address** — here the launch starts before the launcher UI
mounts, so with the patch the browse is never created at all:

| | browse | resolver query-timer fired | launch → bundle |
| - | - | - | - |
| stock | created every run, 0.694–1.542s overlap | **2 of 4 runs** (4 lines each) | 2.208 / 1.474 / 1.603 / 1.280s |
| this PR | **never created** (0 `nw_browser` lines) | 0 of 4 | 2.409 / 1.529 / 1.771 / 1.560s |

So, to your question directly:

1. **The ~2s LAN case is real and it is removed.** `nw_resolver_start_query_timer_block_invoke Query
   fired: did not receive all answers in time` appears in 2 of 8 stock runs and 0 of 8 patched runs,
   because the patched build has no browse in flight to time out.

2. **But I can't claim it makes the launch faster, and I should not have implied it does.** In the
   runs where the timer fired, it fired ~1.6s *after* the bundle GET — concurrent with the launch,
   not blocking it. Launch → bundle is indistinguishable between the two builds (1.26–2.21s stock vs
   1.35–2.41s patched, 16 runs).

My "~2s per launch" came from a trace of my own much larger app where the timeout completed before
the bundle request. Tonight's data says that ordering doesn't imply the launch was waiting on it,
and I have no measurement that establishes it was. Treat the earlier figure as concurrent work, not
as launch latency.

What the PR still stands on, then, is the narrow claim: a launch that already has an explicit URL
shouldn't be running service discovery whose only consumer is a screen that isn't visible — and on
the cold path it now doesn't start one at all. If that isn't worth the notification plumbing to you,
that's a fair call to make, and I'd rather you make it on these numbers than on my earlier ones.

I've updated the PR description to match.

PR #49315, opened 2026-08-24
[expo-maps] Pass anchor for Apple Maps annotation

jensdev:
# Why

`anchor` positions a custom marker image relative to its coordinates on Android, but `AppleMaps.View` has no equivalent — iOS always centers annotation content on the coordinate. A pin-shaped icon therefore can't point at the thing it marks on iOS, and the same pin payload behaves differently per platform.

# How

Adds `anchor` to `AppleMapsAnnotation`, mirroring `GoogleMapsMarker.anchor`. `UnitPoint` uses the same normalized space as Google's anchor — origin top-left, `y` not flipped — so the record passes straight through to SwiftUI's `Annotation(_:coordinate:anchor:content:)` (iOS 17+, already covered by the existing availability gates on both view files).

Added to `AppleMapsAnnotation`, not `AppleMapsMarker`: iOS markers are native balloons where an image anchor has no meaning.

The default stays `{ x: 0.5, y: 0.5 }`, matching `Annotation`'s own `.center` default, so existing annotations don't move. It deliberately does not match Android's `{ x: 0.5, y: 1 }` — harmonizing would silently shift every shipped iOS map. The asymmetry is documented on the field instead.

Note the anchor resolves against the whole annotation content (icon plus `text`), not the icon alone; also documented on the field.

# Test Plan

# Checklist


vonovak:
@expo-bot verify

vonovak:
@jensdev can you also fix this

> decide whether this PR must also update annotationHitRect. That function always centres the annotation hit rect on the coordinate, so the rect no longer matches where an anchored annotation is drawn.

Thank you

jensdev:
Thanks for having a look @vonovak 

I rebased and fixes the feedback. It should all good now.

vonovak:
@expo-bot verify

vonovak (approved):

PR #49318, opened 2026-08-24
feat(expo-file-system): support null-scheme resource URIs in UnifiedFile

CatLover01:
# Why

This PR adds support for Android resource URIs (URIs with a `null` scheme) to the new `expo-file-system` API. 

In the legacy implementation, paths without a scheme were treated as Android resources (e.g., in `res/raw` or `res/drawable`). This functionality was missing in the new `UnifiedFile` architecture, causing parity issues when migrating code to the new API.

This support is necessary because in Android release builds, `expo-asset` URIs are generated as resource identifiers (e.g. `"assets_images_x"`). Without this support, these assets cannot be resolved by the new `expo-file-system` API, which breaks production functionality for apps that have migrated.

See: https://github.com/expo/expo/issues/41996

# How

- **Added ResourceFile**: Implemented a new `UnifiedFileInterface` backend that wraps Android's `Resources`. It supports standard operations like `exists()`, `length()`, and `inputStream()`.
- **Integrated with FileSystemPath**: Updated the URI detection logic to recognize `null` schemes as resources and instantiate the `ResourceFile` adapter.
- **Added Shareable Content URIs**: Implemented `getContentUri()` for resources by using a "copy-to-cache" strategy (consistent with `AssetFile`), allowing bundled resources to be shared with external applications.
- **Hardened javaFile access**: Refactored the `javaFile` getter in `FileSystemPath` to use safe type-checking (`is File`) instead of fragile scheme-based checks, preventing `ClassCastExceptions`.
- **Defined Permissions**: Set resources to be read-only (enforced in `checkPermission` and via `CopyMoveStrategy.Resource`), matching the security model of the legacy module.

# Test Plan

# Checklist

### DIFF
--- a/packages/expo-file-system/android/src/main/java/expo/modules/filesystem/FileSystemPath.kt
+++ b/packages/expo-file-system/android/src/main/java/expo/modules/filesystem/FileSystemPath.kt
@@ -8,6 +8,7 @@ import expo.modules.filesystem.unifiedfile.AssetFile
 import expo.modules.filesystem.unifiedfile.ContentProviderFile
 import expo.modules.filesystem.fsops.DestinationSpec
 import expo.modules.filesystem.unifiedfile.JavaFile
+import expo.modules.filesystem.unifiedfile.ResourceFile
 import expo.modules.filesystem.unifiedfile.SAFDocumentFile
 import expo.modules.filesystem.unifiedfile.UnifiedFileInterface
 import expo.modules.kotlin.exception.Exceptions
@@ -40,6 +41,11 @@ val Uri.isAssetUri
     return scheme == "asset"
   }
 
+val Uri.isResourceUri
+  get(): Boolean {
+    return scheme == null
+  }
+
 fun Uri.isSAFUri(context: Context): Boolean =
   isContentUri && (
     DocumentsContract.isDocumentUri(context, this) || DocumentsContract.isTreeUri(this)
@@ -70,6 +76,7 @@ abstract class FileSystemPath(var uri: Uri) : SharedObject() {
         uri.isSAFUri(context) -> SAFDocumentFile(context, uri)
         uri.isContentUri -> ContentProviderFile(context, uri)
         uri.isAssetUri -> AssetFile(context, uri)
+        uri.isResourceUri -> ResourceFile(context, uri)
         else -> JavaFile(uri)
       }
       return newAdapter.also { fileAdapter = it }
@@ -77,10 +84,10 @@ abstract class FileSystemPath(var uri: Uri) : SharedObject() {
 
   val javaFile: File
     get() =
-      if (uri.isContentUri) {
-        throw Exception("This method cannot be used with content URIs: $uri")
+      if (file is File) {
+        file as File
       } else {
-        (file as File)
+        throw Exception("This method cannot be used with URI: $uri")
       }
 
   fun delete() {
@@ -143,6 +150,9 @@ abstract class FileSystemPath(var uri: Uri) : SharedObject() {
     if (uri.isAssetUri) {
       // TODO: Consider adding a check for asset URIs – this returns asset files of Expo Go (such as root-cert), but these are already freely available on apk mirrors etc.
       return true
+    }
+    if (uri.isResourceUri) {
+      return permission == FilePermissionService.Permission.READ
     }
     return checkPermissionForPath(javaFile.path, permission)
   }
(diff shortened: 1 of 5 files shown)


packages/expo-file-system/android/src/main/java/expo/modules/filesystem/unifiedfile/ResourceFile.kt
```diff
@@ -0,0 +1,89 @@
...
+    throw UnsupportedOperationException("Resource directories are not writable and cannot be created")
+  }
+
+  override fun delete(): Boolean = throw UnsupportedOperationException("Resource files are not writable and cannot be deleted")
+
+  override fun deleteRecursively(): Boolean = throw UnsupportedOperationException("Resource files are not writable and cannot be deleted")
+
+  override fun listFilesAsUnified(): List<UnifiedFileInterface> = emptyList()
+
+  override val type: String? = null
```
barthap: Resources lose their MIME type and extension here. `type` is null, `fileName` (line 51) is the bare identifier, and the cache copy in `getContentUri` (line 58) inherits that extension-less name.

That shows up in a few places:

- `File.type` is null.
- `canPreview()` returns false, and `preview()` throws unless the caller passes `mimeType`.
- A shared or uploaded `contentUri` resolves as `application/octet-stream`, because `FileProvider` reads the type off the extension.
- Copying into a directory writes `assets_qrcode_expo` with no `.jpg`.

`AssetFile` avoids all of this because asset paths keep their extension.

If this stays in file-system, `resources.getValue(id, TypedValue(), true).string` returns the packaged path (`res/raw/assets_qrcode_expo.jpg`), which gives you the extension for `fileName`, the cache path and `type`. It's also the clearest argument for fixing this in `expo-asset` instead, where the cache copy keeps its extension for free.

packages/expo-file-system/android/src/main/java/expo/modules/filesystem/FileSystemPath.kt
```diff
@@ -144,6 +151,9 @@ abstract class FileSystemPath(var uri: Uri) : SharedObject() {
       // TODO: Consider adding a check for asset URIs – this returns asset files of Expo Go (such as root-cert), but these are already freely available on apk mirrors etc.
       return true
     }
+    if (uri.isResourceUri) {
```
barthap: This makes the new copy/move code unreachable. Both `copy()` and `move()` validate `WRITE` before they touch the strategy, so `DestinationSink.Resource` and `CopyMoveStrategy.Resource.moveTo` can never run. The same goes for the `UnsupportedOperationException`s in `ResourceFile`. Every mutation surfaces as `InvalidPermissionException: Missing 'WRITE' permission`, which is misleading, since no permission makes a packaged resource writable.

`AssetFile` works the other way: `isAssetUri` returns true for every permission, so its strategy does run and its read-only errors do surface. Worth picking one of the two models, this is currently half of each.

packages/expo-file-system/android/src/main/java/expo/modules/filesystem/FileSystemPath.kt
```diff
@@ -40,6 +41,11 @@ val Uri.isAssetUri
     return scheme == "asset"
   }
 
+val Uri.isResourceUri
```
barthap: `Paths.join` passes any string straight through, so this catches more than resource identifiers. `new File('relative/item.txt')` used to throw `URI is not absolute` from `JavaFile`, and now it quietly becomes a non-existent `ResourceFile` with `.exists === false`.

That's the cost of implicit detection, and it's the main reason I'd want an explicit URI form if file-system keeps this. See the summary comment.

packages/expo-file-system/android/src/main/java/expo/modules/filesystem/FileSystemPath.kt
```diff
@@ -70,17 +76,18 @@ abstract class FileSystemPath(var uri: Uri) : SharedObject() {
...
 
   val javaFile: File
     get() =
-      if (uri.isContentUri) {
-        throw Exception("This method cannot be used with content URIs: $uri")
+      if (file is File) {
+        file as File
       } else {
-        (file as File)
+        throw Exception("This method cannot be used with URI: $uri")
```
barthap: The `is File` check is a good fix on its own. An asset URI reaching `javaFile` through `rename()` used to give a bare `ClassCastException`.

The message lost some information, though. It no longer says which URIs this applies to, or what to do instead. Maybe something closer to: "This operation requires a local file path, but '$uri' is backed by ${file::class.simpleName}. Use File.text() or File.bytes(), or copy it to a local file first."

Issue #49312, opened 2026-08-24
[expo-ui][Android] A matchContents Host in a header is permanently collapsed to 0x0 by a re-render


andreavrr:
## Summary

On Android, header items rendered through a Jetpack Compose `Host` with `matchContents` measure
correctly on first layout and are then permanently collapsed to `0x0` by the next re-render of the
screen. They are visible for a frame and never come back.

This affects `Stack.Toolbar`, the documented way to add header buttons on Android, because
expo-router renders its children inside `<Host matchContents>`. It also affects any hand-written
`headerRight` / `headerLeft` built with `@expo/ui`.

The re-render does not have to change anything in the header. A `setState` that touches no header
value is enough, because the inline `options={{…}}` object and the `Stack.Toolbar` children are
recreated with a new identity, expo-router calls `setOptions`, and
`ScreenStackHeaderConfig.onUpdate` detaches and re-attaches the toolbar subviews.

In a real app re-renders are unavoidable, so the header buttons disappear about a second after the
screen opens, as soon as the first query resolves or a theme runtime notifies its subscribers.

## Minimal reproducible example

https://github.com/andreavrr/expo-ui-toolbar-repro

Five files under `src/app/`. To flip the behaviour, remove the `setInterval` in
`src/app/(tabs)/(home)/index.tsx`: the icon comes back.

```tsx
export default function Home() {
  // Remove these three lines and the toolbar button stays visible.
  const [tick, setTick] = useState(0);
  useEffect(() => {
    const timer = setTimeout(() => setTick(1), 1000);
    return () => clearTimeout(timer);
  }, []);

  return (
    <>
      <Stack.Screen options={{ title: "Home" }} />

      <Stack.Toolbar placement={"right"}>
        <Stack.Toolbar.Button icon={MsEdit} onPress={() => {}} />
      </Stack.Toolbar>

      <View style={{ flex: 1 }} />
    </>
  );
}
```

## Steps to reproduce

1. `npm install && npx expo run:android`
2. The header shows the edit icon.
3. Once the screen re-renders, the icon disappears and never returns.

**On reproducibility, honestly**: this is a race between the React Native layout pass and the
Compose measure, so a minimal app does not lose it every time. On the app where I found it, a
production-sized one, it is 100% consistent and the header is empty on every open. On this minimal
repro I have seen both outcomes across native rebuilds, with the *same* JavaScript and the *same*
[...]

PR #49320, opened 2026-08-24
[ui][android] Fix matchContents Host collapsing to 0x0 in stack headers

focux:
# Why

Fixes - https://github.com/expo/expo/issues/49312

`Stack.Toolbar` items in the Android header disappear after the first re-render and never come back. Two things in `HostView` combine to do that for a `matchContents` host:

- react-native-screens detaches and re-attaches header subviews when screen options update, Compose transiently measures 0x0, and `dispatchOnLayoutContent` wrote that as `setStyleSize(NaN, NaN)`, clearing the real size and latching `lastDispatchedContentSize`.
- The size is pushed as an immediate shadow node state update. A later commit from JS can replace that state, after which Fabric lays the host out at 0x0 while Compose still holds the measured size, so `onSizeChanged` never fires again.

# How

Skip a 0x0 measurement for `matchContents` hosts before the dedup check, and in `onMeasure` re-apply the last dispatched size when Fabric measures the host EXACTLY at a different one. The re-apply relies on #49211 (expo-modules-core 57.0.13), which made `adopt` always push the state size into Yoga; before that a repeated `setStyleSize` with the same value was a no-op. The re-apply is bounded to one write per mismatch, with a 1px tolerance.

There is no precedent for re-checking the laid out size in the other hosts (`RNHostView`, `AutoSizingComposable`, the iOS `HostView`), which all only push on content size changes. If you would rather handle this where the state is replaced (expo-modules-core `updateState`) I am happy to move it.

# Test Plan

Tested in a production app (expo-modules-core 57.0.13) on a Pixel 9 Pro API 36 emulator. Before, header toolbar items collapsed on every screen open (logging showed the host measured EXACTLY 144 right after dispatch, then EXACTLY 0 with no new dispatch). After, they stay visible.

# Checklist


intergalacticspacehighway:
@expo-bot verify

brentvatne:
@expo-bot verify with fable

andreavrr:
@focux Some device measurements from a production app, since neither `/verify` run could reproduce on a minimal one. Android emulator API 36, logging added to `HostView`.

**Your diagnosis reproduces exactly.** On `expo-modules-core@57.0.12`:

```
DISPATCH size=144x144  ->  WRITE 144x144
(~410 ms later) onMeasure spec=0x0 EXACTLY, lastDispatchedContentSize=144x144, no new dispatch
```

That is the step the reviewer said was unmeasured: Fabric lays the host out without the size while Compose still holds it, and nothing re-dispatches.

Two results on top of that.

**1. With your patch on `expo-modules-core@57.0.12`, the header is still broken.** `reapplyStyleSizeIfLost` fires (`REAPPLY LOST spec=0x0 last=144x144` then `REAPPLY WRITE 144x144`) but the write has no effect, because `changedStyle` in `adopt` compares against `props.yogaStyle`, which still holds 144. So the dependency you noted on #49211 is real, and it is not optional. This answers the open question in the second `/verify`: "the re-apply may not need that commit... I did not measure which case occurs." It needs it.

**2. The control build the second `/verify` could not afford**: stock `HostView.kt`, only the #49211 change. Header correct and stable, and the `LOST` condition never occurs at all. On `expo@57.0.16` with a stock `@expo/ui@57.0.13` and no patches, header items stay at `0,0-144,144` and `0,0-288,144` through repeated push/pop cycles and continuous re-renders.

So for #49312 specifically, #49211 alone is sufficient and your `onMeasure` re-apply never runs. I have closed that issue as fixed in `expo@57.0.16`.

That is not an argument that this PR is worthless: the transient-zero skip is real hardening, and there may be paths where the state is dropped that #49211 does not cover. But it does mean the linked issue no longer motivates it, so the case for merging probably needs a repro that fails on `57.0.13`.

One note for anyone reproducing: I compiled `@expo/ui` from source with `expo.autolinking.android.buildFromSource: ["expo-ui"]` in `package.json`, since as the second `/verify` found, a `patch-package` edit to that Kotlin never reaches the APK while the prebuilt AAR is linked. Doing that also breaks any local module that depends on `expo.modules.ui:expo.modules.ui:+`, which has to become `project(':expo-ui')`.

focux:
Closing this, @andreavrr is right that #49211 is the actual fix.

I checked on stock `HostView.kt` with expo 57.0.16 first: the condition `reapplyStyleSizeIfLost` repairs never comes up, and transient 0x0 dispatches recover on their own. I'd bumped expo-modules-core 57.0.9 to 57.0.13 in the same change that added the patch, so when the header came back I credited the wrong half.

Fair point on the zero skip swallowing `onLayoutContent` too.

PR #49321, opened 2026-08-24
[ios][maps] Render annotation icons at the image size

L65FREAD:
# Why

`AppleMaps.Annotation` renders its `icon` inside a fixed `.frame(width: 50, height: 50)` (added in #34451). Any image that isn't square is stretched to fit, and every image is drawn at 50pt regardless of its size. A wide text-only card ends up vertically stretched; a near-square image card hides the problem.

Android has no such clamp — `BitmapDescriptorFactory.fromBitmap` draws the bitmap at its pixel size — and the docs say so for Google Maps ("The loaded image dimensions determine the marker size"). The Apple Maps tab is silent, so the two platforms silently disagree.

# How

Size the frame from the image: `size * scale / displayScale`, i.e. the image's pixel dimensions in points on the current screen. This matches Android and the `width`/`height` that `useImage` already reports as pixels. Callers scale the icon with `maxWidth` / `maxHeight` on `useImage`, as they do on Android.

Same change in `AppleMapsViewiOS17` and `AppleMapsViewiOS18`. Docs get one line for the Apple Maps tab mirroring the Google Maps one.

#49193 introduces `annotationHitSize = CGSize(width: 50, height: 50)` for tap hit-testing, tied to this frame. Whichever lands second should derive that from the icon size instead.

# Test Plan

- `ExpoMaps` pod builds with the change (iOS 26 SDK, New Architecture).
- Applied to `expo-maps` 57.0.1 via `patch-package` in an SDK 57 app on the iOS simulator: non-square annotation images now render at their intrinsic size with no stretching; square images look the same as before, just no longer forced to 50pt.
- No JS/TS changes, so nothing to rebuild.

# Checklist


docs/pages/versions/unversioned/sdk/maps.mdx
```diff
@@ -223,6 +223,8 @@ export default function Map() {
 
 `AppleMaps.Annotation.icon` expects an image reference, such as the value returned by the `useImage` hook from the `expo-image` library. You can use `AppleMaps.Annotation` for custom image icons. `AppleMaps.Marker` supports marker-specific options such as `systemImage`, `monogram`, and `tintColor`.
 
+The loaded image dimensions determine the annotation size. Pass `maxWidth` and `maxHeight` to the `useImage` hook to scale the icon down.
```
amandeepmittal, line 226: ✏️ **style** `prose.md#core-principles`

The example above calls `useImage('https://example.com/marker.svg')` with no options, so nothing on this tab shows the `maxWidth` and `maxHeight` that this line asks for. Add `{ maxWidth: 48, maxHeight: 48 }` to that call, as the Google Maps example does.

amandeepmittal (changes requested):

amandeepmittal (commented):
Docs changes look good to me!

Thanks for your contribution! Let's wait for a technical review.

amandeepmittal (approved):

PR #49322, opened 2026-08-24
[cli] Pin agent-cli-detector to 0.1.7

davidmokos:
updates agent-cli-detector to 0.1.7 to detect grok in telemetry context

### DIFF
--- a/packages/@expo/cli/package.json
+++ b/packages/@expo/cli/package.json
@@ -79,7 +79,7 @@
     "@expo/xcpretty": "^4.4.4",
     "@react-native/dev-middleware": "0.86.0",
     "accepts": "^1.3.8",
-    "agent-cli-detector": "0.1.6",
+    "agent-cli-detector": "0.1.7",
     "arg": "^5.0.2",
     "bplist-creator": "0.1.0",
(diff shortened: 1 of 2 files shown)


pnpm-lock.yaml
```diff
@@ -19087,6 +19092,8 @@ snapshots:
 
   agent-cli-detector@0.1.6: {}
```
sjchmiela, line 19093: should we do something to drop 0.1.6?

sjchmiela (approved):

packages/@expo/cli/package.json
```diff
@@ -79,7 +79,7 @@
     "@expo/xcpretty": "^4.4.4",
     "@react-native/dev-middleware": "0.86.0",
     "accepts": "^1.3.8",
-    "agent-cli-detector": "0.1.6",
+    "agent-cli-detector": "0.1.7",
```
kitten, line 82: Just noticed the pin; we may want to make sure before the SDK cut (or before if this was backported) to loosen and specify a range. Otherwise, if things go wrong, we're stuck with pinned versions for a long update tail, since patches won't update unless people then update `expo`

Issue #47762, opened 2026-07-14
Expo SDK 57 Android intermittent large-response parsing failure through RTK Query fetchBaseQuery


Developer1970:
### Minimal reproducible example

https://github.com/Developer1970/expo-fetch-regression-bug/tree/main

## Summary

After upgrading an existing app from Expo SDK 54 to Expo SDK 57, my Android app intermittently fails to parse a large JSON response when the request goes through RTK Query [fetchBaseQuery](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html).

The same endpoint and same app logic still work in:

- Web browsers
- The same app before the upgrade on Expo SDK 54
- The upgraded app when the request is made with `XMLHttpRequest` or `fetch` directly works fine (and much faster)

This is not deterministic. The same request may succeed several times and then fail on a later attempt. The failure appears timing-dependent or state-sensitive.

I created a separate Expo reproduction app that runs the same public endpoint through four paths:

- direct `fetch` + `response.text()` + `JSON.parse`
- direct `XMLHttpRequest` + `JSON.parse`
- RTK Query `fetchBaseQuery` with default JSON parsing
- RTK Query `fetchBaseQuery` with `responseHandler: 'text'` + manual `JSON.parse`

In this repro, the intermittent failures are reproducible through the two `fetchBaseQuery` paths. Single-shot requests may pass, but repeated back-to-back runs eventually produce failures.

## Environment

- Expo SDK: `57.0.2`
- React Native: `0.86.0`
- React: `19.2.3`
- Platform: Android
- Data layer in affected app: RTK Query / `fetchBaseQuery`

## Reproduction Repo

Repro repository: [https://github.com/Developer1970/expo-fetch-regression-bug/tree/main](https://github.com/Developer1970/expo-fetch-regression-bug/tree/main)

The repro app includes:

- single-shot probe buttons
- repeated stress buttons for `fetchBaseQueryJson` and `fetchBaseQueryText`
- body length / elapsed time / parse result display
- head/tail preview of returned payloads
- first failing sample capture

## Public Endpoint

Base URL:

`https://api.hucksters.io`

Affected endpoint:

`GET /elist`

The endpoint is public and can be called without authentication.

Example parameter shape used by the app:

```text
https://api.hucksters.io/elist?t=NEXT_365&tp=1&swlong=-123.008&swlat=49.227&nelong=-122.937&nelat=49.315&clustered=1&page=1

### Environment

```text
[...]

Issue #39428, opened 2025-09-06
Expo fetch cancellation can cause a deadlock


Cyberax:
### Minimal reproducible example

https://github.com/GetDuranta/expo-fetch-bug

### Steps to reproduce

1. Run the app. 
2. It will display "Starting fetch" indefinitely:

<img width="117" height="49" alt="Image" src="https://github.com/user-attachments/assets/7af4c617-c2b7-4b0a-a43e-bd2ec0bf0a77" />

Expected output that can be produced by commenting out: `setTimeout(() => controller.abort(), 500);` (line 16 in App.tsx):

<img width="149" height="250" alt="Image" src="https://github.com/user-attachments/assets/66603edc-76ec-42b6-8682-ac9fbb01087d" />

Root cause: race condition in `NativeReponse.swift` (and in its Kotlin variant). The events are fired in this order:
1. The response is created, and the request is started. However, the reader is not yet active.
2. The abort controller fires and eventually calls the `NativeReponse::emitRequestCanceled` method. This method puts the NativeResponse into the `.error` state.
3. The reader wakes up and tries to access the body for the first time. This causes it to lazily create `bodyStream`. However, the request is dead and done at this point. The `start()` method registers the listeners which will never be fired or unsubscribed. Then `response.startStreaming()` is called, and fails with `🟠 Invalid state - currentState[6] validStates[2,3]`. This error then is incorrectly interpreted as streaming start event.
4. That's it. The stream will never be closed.

Suggested fix: don't create the `bodyStream` lazily. Create and start it in the constructor to avoid any possible race conditions.

### Environment

```text
expo-env-info 1.3.4 environment info:
    System:
      OS: macOS 15.6.1
      Shell: 5.3.3 - /opt/homebrew/bin/bash
    Binaries:
      Node: 22.4.1 - ~/.asdf/installs/nodejs/22.4.1/bin/node
      Yarn: 1.22.22 - /opt/homebrew/bin/yarn
      npm: 10.8.1 - ~/.asdf/plugins/nodejs/shims/npm
      Watchman: 2025.08.18.00 - /opt/homebrew/bin/watchman
    Managers:
      CocoaPods: 1.16.2 - /opt/homebrew/bin/pod
    SDKs:
      iOS SDK:
        Platforms: DriverKit 24.5, iOS 18.5, macOS 15.5, tvOS 18.5, visionOS 2.5, watchOS 11.5
    IDEs:
[...]

PR #49324, opened 2026-08-25
 Fix truncated responses being dropped or silently delivered as complete (iOS + Android)

garfieldnate:
Hi, I generated the PR below using Claude (Opus/Fable), but I am a human and I have supervised it closely. For context, at [nonsense](nonsense.com) we are using the experimental `@defer` and `@stream` extensions to GraphQL, which depend on multipart HTTP, and it's pretty demanding of the underlying network stack. Any errors in handling individual pieces of a multi-part GQL query can have a large blast radius, and we're currently seeing mishandling of error states in expo as an important source of app crashes. The TL;DR is that `expo/fetch` swallows some HTTP errors, and we add the logic to catch them correctly.

# Why

When an `expo/fetch` request fails *after* the response headers were delivered but *before* the
consumer touches `response.body`, both platforms lose the failure — differently, and both badly:

- **iOS:** the body stream is never closed and never errored. A read against it hangs until the
  URLSession timeout.
- **Android:** the partial body is handed to JS as a **successful, complete response**. No error, no
  indication anything was lost.

The window is narrow but entirely ordinary. `fetch()` resolves at header time, so anything that
reads the body a few microtasks after `await fetch(...)` sits in it — a `multipart/mixed` parser, a
GraphQL `@defer` client, `Response.clone()` plumbing.

## iOS: the failure is dropped

1. `NativeRequest.start` resolves its promise on `.responseReceived`, i.e. as soon as headers land
   (`ExpoFetchModule.swift`, the `waitFor(states: [.responseReceived, .errorReceived])` in
   `AsyncFunction("start")`). JS now holds a `FetchResponse`.

2. The consumer has not read `response.body` yet, so the state is still `.responseReceived`.

3. URLSession reports the failure. `NativeResponse.handleDidCompleteWithError` gates the terminal
   event on `.bodyStreamingStarted`:

   ```swift
   if state == .bodyStreamingStarted {
     if let error {
       emit(event: "didFailWithError", payload: error.localizedDescription)
     } else {
       emit(event: "didComplete")
     }
   }

   if let error {
     self.error = error
     state = .errorReceived
   } else {
     state = .bodyCompleted
   }

   emit(event: "readyForJSFinalization")
   ```

   State is `.responseReceived`, so **the gate is skipped entirely** — no `didFailWithError` is ever
   emitted. The state becomes `.errorReceived` and only `readyForJSFinalization` fires.

4. `readyForJSFinalization` runs `FetchResponse.finalize`, which `removeAllListeners` for
   `didReceiveResponseData`, `didComplete` and `didFailWithError`.

5. The consumer now reads `response.body`. The getter builds the `ReadableStream` and registers
   fresh `didComplete` / `didFailWithError` listeners — for events that already came and went.

6. `pull()` calls `startStreaming()`. `.errorReceived` is not in
   `isInvalidState(.responseReceived, .bodyCompleted)`, so it logs and returns `nil`:

   ```swift
   func startStreaming() -> Data? {
     if isInvalidState(.responseReceived, .bodyCompleted) {
       return nil
     }
   ```

   `nil` also means "streaming started, chunks will arrive as events", so `pull()` sets
   `streamingState = 'started'` and waits. Nothing will ever arrive.

The success path has a recovery route through the same ambiguity — `.bodyCompleted` is a valid
state, `startStreaming()` returns the buffered bytes, `pull()` closes the stream. The failure path
has none. #39428 reports this from the field with the matching log line, where `6` =
`.errorReceived` and `2`/`3` = the two states `isInvalidState` accepts:

```
Invalid state - currentState[6] validStates[2,3]
```

**Two other paths reach `.errorReceived` through the same gate**, so they hang the same way and are
fixed by the same change: `emitRequestCanceled()` and the `redirect: "error"` branch of
`willPerformHTTPRedirection` both set `.errorReceived` while emitting `didFailWithError` only from
`.bodyStreamingStarted`. A `redirect: "error"` response whose body is read a tick later never
settles today.

## Android: the failure is overwritten

Android reaches the same window by a different route, and the outcome is worse.

OkHttp does not deliver a mid-body failure through `onFailure` — the response already arrived, so
the error surfaces when the body source is read, inside `pumpResponseBodyStream`. That method
catches it and records it correctly:

```kotlin
} catch (e: IOException) {
  this.error = e
  if (state == ResponseState.BODY_STREAMING_STARTED) {
    emit("didFailWithError", e.localizedMessageWithCauseLocalizedMessage())
  }
  state = ResponseState.ERROR_RECEIVED
}
```

But it swallows the exception and returns normally, and the tail of the `onResponse` coroutine then
runs unconditionally:

```kotlin
try {
  pumpResponseBodyStream(stream)
} finally { … }

if (this@NativeResponse.state == ResponseState.BODY_STREAMING_STARTED) {
  emit("didComplete")
}
this@NativeResponse.state = ResponseState.BODY_COMPLETED   // ← overwrites ERROR_RECEIVED
emit("readyForJSFinalization")
```

So `ERROR_RECEIVED` is immediately overwritten with `BODY_COMPLETED`. When JS later reads the body,
`startStreaming()` sees a perfectly valid `BODY_COMPLETED`, returns whatever partial bytes reached
the sink, and `pull()` closes the stream normally. **A truncated body is delivered as a complete
one.** `text()` and `arrayBuffer()`, which wait on `BODY_COMPLETED`, resolve with the partial bytes
too.

(When JS *had* already started streaming, `pumpResponseBodyStream` emits `didFailWithError` before
the overwrite, so that path is correct today. Only the pre-read window is affected.)

## OkHttp reports every truncation — including the one CFNetwork does not

The two platforms exhibit different defect sets here, which is worth setting out. Driving real
OkHttp 4.12.0 through a truncating server, replicating expo's read loop:

| Wire shape | OkHttp result | CFNetwork (iOS) result |
| --- | --- | --- |
| correct chunked body | clean EOF, 5 B | success |
| complete frame, **no terminal `0\r\n\r\n`**, FIN | `EOFException` | **success, error `nil`** |
| frame data, trailing CRLF withheld, FIN | `EOFException: \n not found` | `-1017 cannotParseResponse` |
| cut mid-frame, FIN | `ProtocolException: unexpected end of stream` | `-1005 networkConnectionLost` |
| complete frame, then RST | `SocketException: Connection reset` | `-1005` |

Row 2 is the interesting one. RFC 9112 §8 requires a chunked body without its terminal zero-length
chunk to be treated as incomplete. OkHttp does. CFNetwork's default loader does **not** — it reports
success, indistinguishable from a whole body (the private Network.framework loader returns `-1005`
for the identical cut). So on iOS that particular truncation is invisible below the application
layer and is *not* addressed here; it is filed with Apple as FB24495140. On Android it is an
ordinary `IOException`, and this PR makes it surface.

## The fix

**iOS** — `startStreaming()` reports the failure instead of returning the ambiguous `nil`:

```swift
func startStreaming() throws -> Data? {
  if state == .errorReceived {
    throw error ?? FetchUnknownException()
  }
  if isInvalidState(.responseReceived, .bodyCompleted) {
    return nil
  }
```

and the `AsyncFunction` wrapper becomes throwing so the rejection reaches JS.

**Android** — stop overwriting `ERROR_RECEIVED`, and give `startStreaming()` the same treatment:

```kotlin
if (this@NativeResponse.state == ResponseState.ERROR_RECEIVED) {
  emit("readyForJSFinalization")
  return@launch
}
```

```kotlin
if (state == ResponseState.ERROR_RECEIVED) {
  throw error?.toCodedException() ?: FetchUnknownException()
}
```

Because the state is now preserved, `text()` and `arrayBuffer()` would otherwise wait forever on
`BODY_COMPLETED`, so they get the `ERROR_RECEIVED` branch that #48230 already added on iOS. That is
not scope creep — it is required to keep them working once the state survives.

**No JS change is required on either platform.** `pull()` already does
`await this.startStreaming()`, so a rejected promise errors the `ReadableStream` — the same outcome
`controller.error()` would produce from a `didFailWithError` listener.
`packages/expo/src/winter/fetch/__tests__/FetchResponse-test.ts` already covers that half of the
contract under `internally errored stream (rejected pull)`; this change is what makes the native
side actually produce the rejection.

`error ?? FetchUnknownException()` / `error?.toCodedException() ?: FetchUnknownException()` are the
expressions `AsyncFunction("start")` already uses on each platform, so the error surfaced to JS is
consistent across every entry point.

# How

- `NativeResponse.startStreaming()` throws from `.errorReceived` / `ERROR_RECEIVED` on both
  platforms, and the iOS `AsyncFunction("startStreaming")` wrapper became `throws`.
- The Android `onResponse` tail returns early on `ERROR_RECEIVED` instead of overwriting it with
  `BODY_COMPLETED`.
- Android `text()` / `arrayBuffer()` now also wait for `ERROR_RECEIVED` and reject from it, matching
  what #48230 did on iOS.
- Adds `packages/expo/ios/Tests/NativeResponseTests.swift` (the first iOS native tests for the Fetch
  module) and `packages/expo/android/src/test/java/expo/modules/fetch/NativeResponseTest.kt`, each
  covering every route into the error state plus the success paths that must not change.

# Test Plan

**Android, executed.** A JVM harness ports `NativeResponse`'s state machine and drives it with real
OkHttp 4.12.0 against a server that truncates a chunked body five ways, with JS reading the body
only after the failure has landed:

```
=== CURRENT (upstream) — late read ===
mode               | final state      | startStreaming()   | what JS sees
clean              | BODY_COMPLETED   | returned 5 B       | stream closes with 5 B — correct
cut-chunk1-fin     | BODY_COMPLETED   | returned 5 B       | stream closes with 5 B — SILENTLY TRUNCATED
cut-pad0-fin       | BODY_COMPLETED   | returned 5 B       | stream closes with 5 B — SILENTLY TRUNCATED
cut-midframe-fin   | BODY_COMPLETED   | returned 8 B       | stream closes with 8 B — SILENTLY TRUNCATED
cut-chunk1-rst     | BODY_COMPLETED   | returned 5 B       | stream closes with 5 B — SILENTLY TRUNCATED

=== WITH FIX — late read ===
clean              | BODY_COMPLETED   | returned 5 B       | stream closes with 5 B — correct
cut-chunk1-fin     | ERROR_RECEIVED   | threw EOFException | stream errors — correct
cut-pad0-fin       | ERROR_RECEIVED   | threw EOFException | stream errors — correct
cut-midframe-fin   | ERROR_RECEIVED   | threw ProtocolException | stream errors — correct
cut-chunk1-rst     | ERROR_RECEIVED   | threw SocketException   | stream errors — correct
```

The control row is unchanged, which is the point: only genuinely truncated responses change
behavior. The three harness files — `server.js`, `StateMachine.java`, `Harness.java` — plus the
manual reproduction are attached to this PR as [okhttp-truncation-harness.zip](https://github.com/user-attachments/files/31400814/okhttp-truncation-harness.zip), along with a
README giving the exact compile and run commands.

**Android unit tests, executed.** `packages/expo/android/src/test/java/expo/modules/fetch/NativeResponseTest.kt`
drives the real `NativeResponse` over an OkHttp response whose body source throws mid-read. It runs
on the JVM with no JS runtime attached — `runtimeContextHolder` is set to an empty `WeakReference`,
so `getJavaScriptObject()` returns null and `emit(...)` is a no-op before reaching JNI, the same
situation iOS reports as "the JS runtime has been lost". The private `state` is read through
`waitForStates`, which fires synchronously when the state already matches, so there is no
reflection.

```
> Task :expo:testDebugUnitTest
NativeResponseTest: tests="7" skipped="0" failures="0" errors="0"
BUILD SUCCESSFUL
```

The assertions are made **after** the launched coroutine settles, which matters: before the fix the
state went `ERROR_RECEIVED` → `BODY_COMPLETED`, so asserting on the first observed state would pass
against broken code. As a check that these are genuine regression tests, removing just the
`ERROR_RECEIVED` early return from `onResponse` and re-running gives:

```
NativeResponseTest > a truncated body ends in ERROR_RECEIVED, not BODY_COMPLETED FAILED
NativeResponseTest > startStreaming throws the recorded error after a truncated body FAILED
7 tests completed, 2 failed
```

Exactly the two truncation tests fail; the clean-body control and the other four are unaffected.
`:expo:spotlessCheck` passes.

**iOS, executed.** `packages/expo/ios/Tests/NativeResponseTests.swift` — the first iOS native tests
for the Fetch module. It drives `NativeResponse` through the delegate callbacks the way
`ExpoURLSessionTask` does, draining the serial queue between steps, and asserts that
`startStreaming()` throws the underlying `URLError` after a post-response failure
(`.networkConnectionLost`, `.cannotParseResponse`), still returns the buffered body from
`.bodyCompleted`, and still returns `nil` and moves to `.bodyStreamingStarted` from
`.responseReceived`. It also covers the other two routes into `.errorReceived` — a cancel and a
rejected redirect — with a followed redirect as the control, so the paths that work today are
pinned alongside the ones being fixed.

Run locally via the `Expo-Unit-Tests` scheme in `apps/bare-expo`:

```
✔ Test "startStreaming throws when the request failed before the body was read" passed
✔ Test "startStreaming rethrows a truncated chunked body error" passed
✔ Test "startStreaming throws after the request was canceled before the body was read" passed
✔ Test "startStreaming throws when a redirect was rejected before the body was read" passed
✔ Test "startStreaming still returns the buffered body when the request succeeded" passed
✔ Test "startStreaming still opens the stream while the body is in flight" passed
✔ Test "a followed redirect still streams normally" passed
✔ Suite "NativeResponse" passed
✔ Test run with 13 tests in 3 suites passed
** TEST SUCCEEDED **
```

(The `🟡 ... the JS runtime has been lost` lines that appear during the run are expected — the
harness exercises the state machine with no JS runtime attached, so `emit` has nowhere to go.)

# Related, not duplicate

- **#47796** — same file, and its write-up names "never closed" as a symptom. Its cure was confining
  the URLSession delegate callbacks to `dispatchQueue`; it created `handleDidCompleteWithError` but
  left the `.bodyStreamingStarted` gate byte-identical. Different cause, overlapping symptom.
- **#48124 / #48230** — the buffered-read sibling on iOS. #48124's analysis concluded that "the
  streaming path handles the identical failure correctly", which holds only when the reader reached
  `startStreaming()` *before* the failure. #48230 fixed `text()` / `arrayBuffer()` via
  `waitFor(states:)`; the stream path does not go through `waitFor`, so it was left behind. After
  #48230 the asymmetry is sharper: the buffered reads reject and the stream still hangs. This PR also
  ports #48230's fix to Android, where it was never applied.
- **#39428** — the only existing report of the iOS mechanism ("`startStreaming` … incorrectly
  interpreted as streaming start event … the stream will never be closed"), but attributed entirely
  to `abort()`. This fixes the transport-failure trigger, and is more evidence that the `nil` return
  doing double duty is the underlying defect.
- **#47762** — Android chunk reordering. Unrelated.

# Checklist

- [X] I added a `changelog.md` entry and rebuilt the package sources according to [this short guide](https://github.com/expo/expo/blob/main/CONTRIBUTING.md#-before-submitting)
- [X] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: updated a module plugin).
- [X] Conforms with the [Documentation Writing Style Guide]

(https://github.com/expo/expo/blob/main/guides/Expo%20Documentation%20Writing%20Style%20Guide.md) — n/a, no documentation changes


brentvatne:
@expo-bot verify with fable

garfieldnate:
This doesn't change the PR in any way, but Apple came back to me on FB24495140, and it turns out that they know about the non-compliant network behavior in CFNetwork causing a chunked body without its terminal zero-length chunk to be treated as incomplete. They recommend updating to their new and optimized network stack by setting `URLSessionConfiguration.usesClassicLoadingMode = false`.

However:

1. It's version-limited. The property is only public API from iOS 18.4 / macOS 15.4 / watchOS 11.4 / tvOS 18.4 / visionOS 2.4 (NSURLSession.h), so any use needs an #available guard, and older OSes can't opt in at all. (The header also notes "Defaults to YES. The default may be NO in a future OS update", so Apple intends to migrate everyone eventually).
2. There's a risk to monitoring libraries. The new loader doesn't invoke custom NSURLProtocol subclasses, and APM/RUM SDKs that intercept via URLProtocol (or otherwise assume the classic loader) can go blind on requests made through it. Delegate-level swizzling generally still works, but it's SDK-dependent, so it would need heavy testing by many users before switching.

Claude is telling me I could add an expo module to turn the new network stack on myself if I wanted:

```ts
ExpoFetchCustomExtension.setCustomURLSessionConfigurationProvider {
    let config = URLSessionConfiguration.default
    if #available(iOS 18.4, macOS 15.4, tvOS 18.4, watchOS 11.4, visionOS 2.4, *) {
        config.usesClassicLoadingMode = false
    }
    // provider replaces createURLSession()'s else-branch, so re-apply its defaults:
    config.httpShouldSetCookies = true
    config.httpCookieAcceptPolicy = .always
    config.httpCookieStorage = HTTPCookieStorage.shared
    return config
}
```

The changes I've requested in this PR are still correct and relevant, but some client out there might want to try this if the silent truncations on iOS are an issue for them.

PR #49351, opened 2026-08-25
[router] Create routing queue context

Ubax:
# Why

In order to support react-transitions, we need to move the routing queue to react tree. This PR adds routing queue context which is used to process all navigation events executed using non-global API - `useRouter`

# How

1. Add two contexts - One for  `enqueue`/`dequeue` which is stable - it only adds and removes from the state. Another one for `intents` - the state which changes. The second context is only used to dispatch the state updates
2. Modify all navigation approaches - `useRouter`, `Link`, `useDomComponentNavigation`, `useNavigationHelpers` to use the context queue
3. Add a bridge which passes the context queue to the non-react queue during first render
4. Fix loader tests, by awaiting their data rather then only the locator

# Test Plan

1. CI
2. Router e2e

# Checklist


packages/expo-router/src/global-state/routingQueueContext.tsx
```diff
@@ -0,0 +1,101 @@
...
+        bridges.splice(index, 1);
+      }
+    };
+  }, []);
+
+  useEffect(() => {
+    if (intents.length === 0) {
+      return;
+    }
+    for (const intent of routingQueue.drain(intents)) {
```
jakub-agent: [🏗️ **design**] The module-level `router` writes to `routingQueue` and only reaches the provider queue through this effect, while every hook path (`useRouter`, `Link`, `navigation.dispatch`, `useLinking`) writes to the provider queue directly. Two consequences:

1. FIFO is no longer global. An intent enqueued through the module `router` is processed one commit *after* an intent enqueued through a hook in the same tick, so their relative order can invert. Concrete case: `<Link href="/a" onPress={() => router.dismissAll()} />` — `dismissAll` goes to `routingQueue`, `linkTo('/a')` goes straight to the provider queue, the drainer runs `/a` in batch 1 and `dismissAll` in batch 2, so the user ends up at the top of the stack instead of on `/a`. Before this PR both intents shared the single module queue and kept their order.
2. Every imperative `router.*` call now costs an extra commit: bridge re-render → bridge effect → provider state update → drainer effect. Previously the drainer subscribed to the module queue directly.

Would a module-level registration of the active provider's `enqueue` (set on mount, cleared on unmount) work here instead? That keeps one queue and one commit while still scoping intents to the mounted root.

packages/expo-router/src/global-state/RoutingQueueDrainer.tsx
```diff
@@ -67,7 +36,8 @@ export function RoutingQueueDrainer({ ready, processIntent }: Props) {
         );
       }
     }
-  }, [intents, processIntent, ready]);
+    dequeue(intents);
```
jakub-agent: [🏗️ **design**] Processing is no longer idempotent. `routingQueue.drain(intents)` used to reset the module queue synchronously, so a second run of this effect with the same array was a no-op. `dequeue` is a state update, so any second run before it commits reprocesses the whole batch.

StrictMode's double-invoked mount effect does exactly that whenever the drainer mounts with pending intents — a Fast Refresh remount of `BaseNavigationContainerInner` while intents are queued, or any later change that seeds the provider queue before the drainer mounts.

The renamed `does not process a batch twice in Strict Mode` test now enqueues *after* mount. Update effects are never double-invoked, so the test no longer covers the case its name describes. A ref holding the last processed array (skip when identical) would restore the guard cheaply and let the test go back to enqueueing before mount.

packages/expo-router/src/global-state/router.ts
```diff
@@ -243,17 +267,31 @@ export type ImperativeRouter = {
...
+    dismiss: (count) => dismissImpl(enqueue, count),
+    dismissAll: () => dismissAllImpl(enqueue),
+    dismissTo: (href, options) => dismissToImpl(enqueue, href, options),
+    canDismiss,
+    replace: (href, options) => replaceImpl(enqueue, href, options),
+    back: () => goBackImpl(enqueue),
+    goBack: () => goBackImpl(enqueue),
+    canGoBack,
+    reload,
+    prefetch: (href) => prefetchImpl(enqueue, href),
```
jakub-agent: [💡 suggestion] `prefetch` is the only method here that drops its second argument — every sibling forwards `options`. `prefetchImpl` accepts `NavigationOptions`, and the pre-PR `export function prefetch(href, options?)` forwarded them, so the `prefetch` re-exported from `global-state/routing` now silently ignores options.

Nothing internal passes options today and `ImperativeRouter['prefetch']` is typed `(name: Href) => void`, so this is latent rather than broken. Still, the asymmetry will bite whoever widens that type: either widen it now (`prefetch: (href: Href, options?: NavigationOptions) => void`) and forward, or add a short comment saying prefetch is intentionally option-less.

packages/expo-router/src/global-state/RoutingQueueDrainer.tsx
```diff
@@ -2,54 +2,23 @@
...
-  }, []);
+  const intents = React.use(PendingIntentsContext);
+  const { dequeue } = React.use(RoutingQueueApiContext)!;
 
   React.useEffect(() => {
     if (!ready || intents.length === 0) {
       return;
     }
-    for (const intent of routingQueue.drain(intents)) {
+    for (const intent of intents) {
```
jakub-agent, line 25: [💡 suggestion] Duplicate-container detection moved from the drainer to the provider (`bridges`), but the duplication it guards against happens here. Two containers sharing one provider both read the same `PendingIntentsContext` and both loop over it, so every intent is dispatched twice — and `bridges.length` stays at 1, so nothing warns. Previously `routingQueue.drain(intents)` returned `[]` for the second drainer and the dev error fired.

This is not user-reachable today: `ExpoRoot` renders exactly one container, and a nested `NavigationContainer` throws on `!parent.isDefault`. But the deleted `logs an error when multiple drainers are mounted` test was the only thing pinning that invariant. Either keep a drainer-level guard, or add a comment stating why one drainer per provider is guaranteed.

packages/expo-router/src/fork/__tests__/__fixtures__/store.tsx
```diff
@@ -27,7 +33,28 @@ export const storeValue: StoreContextValue = {
 };
 
 export function StoreProvider({ children }: { children: ReactNode }) {
-  return <StoreContext.Provider value={storeValue}>{children}</StoreContext.Provider>;
+  const [queue, setQueue] = useState<RoutingIntent[]>([]);
+  const api = useMemo<RoutingQueueApi>(
+    () => ({
+      enqueue: (intent) => {
+        routingQueue.queue = [...routingQueue.queue, intent];
```
jakub-agent: [💡 suggestion] This writes into the module queue so the `useLinking` tests can keep asserting on `routingQueue.queue`, but production `useLinking` now enqueues into the context and never touches `routingQueue`. The `expect(routingQueue.queue).toEqual(...)` assertions in `useLinking.test.ios.tsx` therefore verify the fixture, not the code under test — they would still pass if the `enqueue` call were removed from `useLinking`.

Rendering the real `RoutingQueueProvider` and asserting on `PendingIntentsContext` would keep that coverage honest, and would also remove this second copy of the `dequeue` logic, which can now drift from the real one.

packages/expo-router/src/global-state/routingQueueContext.tsx
```diff
@@ -0,0 +1,101 @@
...
+      </PendingIntentsContext.Provider>
+    </RoutingQueueApiContext.Provider>
+  );
+}
+
+export function useEnqueueRoutingIntent() {
+  const api = use(RoutingQueueApiContext);
+  if (api === undefined) {
+    throw new Error(
+      'Routing queue is unavailable. This is most likely a bug in expo-router. Please report it at https://github.com/expo/expo/issues.'
```
jakub-agent: [💡 suggestion] `Link`, `useRouter`, `useLinkToPathProps` and every navigator now throw when rendered without a provider; before this PR those calls silently queued into the module queue. That makes this message reachable from userland trees — the most likely one being a unit test that renders `<Link>` or a `useRouter()` consumer without `ExpoRoot` (this PR had to add `RoutingQueueProvider` wrappers to `Link.test.ios.tsx` and `Button.test.*.tsx` for exactly that reason, and the provider is not publicly exported).

Telling those users it is "most likely a bug in expo-router" will produce false bug reports. Leading with the probable cause — rendered outside the Expo Router root / `ExpoRoot` — would be more actionable. It is probably also worth calling the new throw out in the changelog entry, since the old behavior was a silent no-op.

packages/expo-router/src/react-navigation/core/BaseNavigationContainer.tsx
```diff
@@ -66,6 +70,16 @@ const duplicateNameWarnings: string[] = [];
  */
 export function BaseNavigationContainer(props: InternalNavigationContainerProps) {
   const registry = use(RouterRegistryContext);
+  const routingQueueApi = use(RoutingQueueApiContext);
+
+  // TODO(@ubax): Remove this when marking BaseNavigationContainer as kind of singleton
+  if (routingQueueApi === undefined) {
```
jakub-agent: [💡 suggestion] Three places now install the provider: `ExpoRoot` (unconditional, needed for `useDomComponentNavigation`), `NavigationContainer` (needed for `useLinking`), and here. `NavigationContainer` already covers every path that reaches `BaseNavigationContainer` in product code, so is this branch only serving tests that render `BaseNavigationContainer` bare?

If so, wrapping in the test helper keeps the production path single-purpose. If it is load-bearing for the `expo-router/react-navigation` compat entry, a one-line comment saying so would help. Either way, the recursive self-render costs an extra render pass compared with wrapping `BaseNavigationContainerInner` and handling both fallbacks in one place.

packages/expo-router/src/global-state/RoutingQueueDrainer.tsx
```diff
@@ -2,54 +2,23 @@
...
-
-    return () => {
-      const index = drainers.indexOf(drainer);
-      if (index > -1) {
-        drainers.splice(index, 1);
-      }
-    };
-  }, []);
+  const intents = React.use(PendingIntentsContext);
+  const { dequeue } = React.use(RoutingQueueApiContext)!;
```
jakub-agent, line 15: [🐧 nit] The `!` turns a missing provider into `Cannot destructure property 'dequeue' of undefined`, while `useEnqueueRoutingIntent` raises a descriptive error for the same condition. A shared `useRoutingQueueApi()` hook returning the whole api would keep the message consistent and drop the assertion.

jakub-agent (commented):

Solid step toward per-root scoping: the queue now lives in React state, the drainer reads it from context, and pending intents die with the provider — which is what the removed `TODO(@ubax)` asked for. Tests were reworked rather than dropped, and the fallback providers keep `NavigationContainer` / `BaseNavigationContainer` usable standalone.

Two things I would want resolved before merge:

- **The imperative `router` and the hook APIs no longer share one FIFO.** Module-level `router.*` writes to `routingQueue` and reaches the provider queue one commit later through `ImperativeRoutingQueueBridge`. Its order relative to a `Link` / `useRouter` / `navigation.dispatch` intent from the same tick can now invert, and every imperative navigation costs an extra commit.
- **Batch processing lost its idempotency guard.** `routingQueue.drain(intents)` reset the queue synchronously; `dequeue` is a state update, so a second run of the drainer effect over the same array reprocesses it. The renamed StrictMode test no longer covers that path.

Smaller items: `prefetch` silently drops `options`; the "please report a bug" error is now reachable from userland trees; duplicate-container detection moved away from the place where duplication actually happens; and the `store.tsx` fixture writing into `routingQueue.queue` makes the `useLinking` queue assertions self-fulfilling.

Out of scope, but worth noting: `routingQueue` is still a module-level global, so server / static rendering keeps sharing it across renders.

I did not run the package test suite. All findings come from reading the diff and the surrounding files at `37c5b32`.

packages/expo-router/src/global-state/routingQueueContext.tsx
Ubax: For now leave as is. Add TODO to look into module-level registration.

Also add a warning in all the imperative calls that they are not synchronized with callas from `useRouter` and whenever possible `useRouter` is a preferred option.

Add todo near the imperative log to add a flag to disable the migration warninngs

packages/expo-router/src/global-state/RoutingQueueDrainer.tsx
Ubax: Add a test that covers this case.

Would it work if instead we would do:

```
const intents = useSyncExternal(....)
const [stateIntents, set] = useState(intents)
if(intents.length > 0 && intents !== stateIntents){
set(intents)
}

useEffect(() => {
  ....
},[stateIntents])
```

packages/expo-router/src/global-state/RoutingQueueDrainer.tsx
Ubax: Ignore

packages/expo-router/src/fork/__tests__/__fixtures__/store.tsx
Ubax: Do it

packages/expo-router/src/global-state/routingQueueContext.tsx
Ubax: Change the message to mention `ExpoRoot`. Instead of throwing in the hook, throw whenever any operation is executed on the `undefined` api. So if user just renders it is ok, but if they will try to navigate, then it will throw

packages/expo-router/src/react-navigation/core/BaseNavigationContainer.tsx
Ubax: Remove the wrapping and add it in tests only

packages/expo-router/src/global-state/RoutingQueueDrainer.tsx
Ubax: That's ok. This component is not public.

packages/expo-router/src/global-state/router.ts
```diff
@@ -237,23 +261,66 @@ export type ImperativeRouter = {
...
+    goBack: () => goBackImpl(enqueue),
+    canGoBack,
+    reload,
+    prefetch: (href, options) => prefetchImpl(enqueue, href, options),
+    setParams: setParams as ImperativeRouter['setParams'],
+    linkTo: (href, options) => linkToImpl(enqueue, href, options),
+  };
+}
+
+const throwBeforeFirstRender = () => {
```
jakub-agent, line 294: [🚨 **critical**] This is a breaking behavior change and it is not called out anywhere. `can navigate before navigation is ready` was deleted and replaced by `throws when navigating before the first render finishes`, so `router.push()` during the first render now throws where it previously queued and replayed. The queue existed for exactly that case — `routingQueue`'s original comment said "calls made before the container is ready aren't lost".

I think the tightening is defensible: `enqueue` is now a `setState`, so a render-phase call would otherwise produce React's "Cannot update a component while rendering a different component" error. But the changelog entry only says "Scope routing queues to each router root and bind `useRouter()` to its owning container", which does not tell anyone that their render-phase or layout-effect `router.replace('/login')` now throws. Please add an explicit changelog line, and per the repo convention a `## Accepted breaking changes` section in the PR description.

Three related points:

1. The message says "before the first render has finished", but the real condition is "before the root's layout effects have run". A child's `useLayoutEffect` fires before the bridge's (the bridge is the last child of the provider), so a call from there throws even though render finished. Worth rewording.
2. `canGoBack`, `canDismiss`, `setParams` and `reload` do not need the queue at all — they read `store.navigationRef` or throw on their own. Gating them behind this placeholder makes them throw in cases where they used to work. They could keep their real implementations in the initial object.
3. `Prefetch.tsx` calls module `router.prefetch` from a `useLayoutEffect`. It survives only because `useOptionalNavigation` returns `null` on the first pass, so the call happens in a later commit. That is a fragile invariant to leave undocumented now that the placeholder throws.

packages/expo-router/src/standard-navigation/useVisibleTabsWithRedirect.ts
```diff
@@ -39,6 +39,7 @@ export function useVisibleTabsWithRedirect<
   descriptors: Record<string, TabDescriptor<Options>>;
 }) {
   const buildHref = useBuildHref();
+  const router = useRouter();
```
jakub-agent: [🏗️ **design**] `useRouter()` returns `routerWithWarnings` when `isPreview` is true, so inside a link preview this hook's redirect becomes a no-op that logs `router.replace should not be used in a previewed screen. To fix this issue, wrap navigation calls with 'if (!isPreview) { ... }'`. Two problems: a tabs layout previewed with a hidden or undeclared focused route no longer redirects, and the warning tells the user to guard a call they do not own — it is inside expo-router.

`useRouterActions()` is the internal hook for this, and it is what `useLinkToPathProps` and `useDomComponentNavigation` use. Library-internal navigation should not go through the preview shim.

The test diff suggests this change exists to satisfy `renderHook` (which has no provider, so module `router` now throws) — `jest.mock('../../hooks/useRouter')` plus `mockedUseRouter.mockReturnValue(router)`. Wrapping the `renderHook` call in `RoutingQueueProvider` would fix the test without changing production behavior.

Also, `router` is missing from the effect's dependency array below (`[isFocused, redirectHref, visibleFocusedIndex]`). Harmless while the identity is stable, but it will go stale if `isPreview` ever flips.

packages/expo-router/src/global-state/routingQueueContext.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const api = use(RoutingQueueApiContext);
+  if (api === undefined) {
+    return throwMissingRoutingQueue;
+  }
+  return api.enqueue;
+}
+
+function ImperativeRoutingQueueBridge({ enqueue }: Pick<RoutingQueueApi, 'enqueue'>) {
+  useClientLayoutEffect(() => {
+    Object.assign(router, createImperativeRouter(enqueue));
```
jakub-agent, line 69: [🏗️ **design**] Two lifecycle gaps here:

1. **No cleanup.** After the provider unmounts, `router` keeps the dead provider's `enqueue`, so `router.push()` calls `setQueue` on an unmounted component — silently dropped instead of raising the clear "unavailable" error the placeholder gives. Restoring `throwBeforeFirstRender` (or the previous binding) in the effect cleanup makes that failure loud, and also stops one test's binding from leaking into the next.
2. **Multiple roots bind silently.** Whichever bridge's layout effect runs last owns `router`, and intents from the other root are dropped. The previous round had the `bridges` counter and the "multiple navigation containers" `console.error` for this; both are gone, along with `warns for multiple bridges and forwards through only one` and the sibling-container test in `BaseNavigationContainer.test.ios.tsx`. Two `ExpoRoot`s is exactly the misconfiguration that error was written for. A dev warning when a second bridge rebinds would keep the diagnostic.

More broadly: this PR removes one mutable module singleton (the queue) and introduces another (a `router` object rebound from a commit-phase effect). Worth a comment explaining why the exported identity has to stay stable, so the next reader does not treat this as an accident.

packages/expo-router/src/global-state/router.ts
```diff
@@ -237,23 +261,66 @@ export type ImperativeRouter = {
...
+  throw new Error('The imperative router is unavailable before the first render has finished.');
 };
+
+export const router: InternalRouter = {
```
jakub-agent: [🏗️ **design**] The annotation changed from `ImperativeRouter` to `InternalRouter`, and `imperative-api.tsx` re-exports the value without re-annotating (`export { router }`), so the inferred public type now carries `goBack` and `linkTo`. That widens the public API: `router.linkTo()` becomes callable and typed for users, with internal `LinkToOptions` in its signature, and typedoc will pick both up in the API reference.

Annotating the public re-export as `ImperativeRouter` keeps the internal shape internal while still letting `createImperativeRouter` return the wider type for in-package callers.

packages/@expo/cli/e2e/playwright/page.ts
```diff
@@ -1,4 +1,10 @@
-import type { ConsoleMessage, Page, Request } from '@playwright/test';
+import { expect, type ConsoleMessage, type Page, type Request } from '@playwright/test';
+
+export async function waitForLoaderData(page: Page, data: unknown) {
```
jakub-agent, line 3: [💡 suggestion] These four Playwright files look like unrelated scope for a router-internals PR: swapping `waitForSelector('[data-testid="loader-result"]')` for a `toHaveText` assertion on the expected payload, plus `waitUntil: 'networkidle'` on two `goto` calls.

The change itself is an improvement — the old wait passed as soon as the *previous* page's element was present, so it could not distinguish navigations. But it reads as a flakiness fix that was needed because navigation now settles at a different point. If it is load-bearing for this PR, saying so in the description helps; if it is independent hardening, it would be easier to review split out.

packages/expo-router/src/global-state/RoutingQueueDrainer.tsx
```diff
@@ -2,54 +2,26 @@
...
```
jakub-agent, line 15: [🐧 nit] Still open from the previous round, and now slightly more exposed: with the `BaseNavigationContainer` fallback removed, rendering `BaseNavigationContainer` without a provider fails here with `Cannot destructure property 'dequeue' of undefined` rather than the descriptive message `useEnqueueRoutingIntent` produces. A shared `useRoutingQueueApi()` returning the whole api would keep both paths consistent and drop the assertion.

jakub-agent (commented):

Re-review at `618a7fa` (iteration 2). The architecture changed substantially since `37c5b32`: the module-level queue object is gone, and `ImperativeRoutingQueueBridge` now rebinds the exported `router` to the provider's `enqueue` in a layout effect.

**Resolved from the previous round**

- Split FIFO / extra commit per imperative navigation — the module queue is removed, so `router.*` reaches the provider queue directly.
- StrictMode double-processing — the `lastProcessed` ref guard in the drainer covers it, and the comment explains why.
- `prefetch` dropping `options` — fixed, and `ImperativeRouter['prefetch']` was widened.
- Misleading "report a bug" error — now names `ExpoRoot` first, and `useEnqueueRoutingIntent` defers the throw to call time so `<Link>` still renders outside a provider.
- The third provider-install site in `BaseNavigationContainer` — removed; the test fixture wraps instead.
- The `store.tsx` fixture writing into the module queue — gone, replaced by a `PendingIntentsContext` probe.

**New in this round**

- **`router.*` now throws instead of queueing before the root commits.** The deleted `can navigate before navigation is ready` test documented the old contract, and the routing queue existed specifically so pre-ready calls were not lost. This is a deliberate tightening, but it is a user-facing break and the changelog entry does not mention it.
- **`useVisibleTabsWithRedirect` switched to `useRouter()`**, which returns the preview warning stub inside a link preview, so the hidden-focused-route redirect silently no-ops there.
- **`Object.assign(router, …)` never unwinds** — no cleanup on unmount, and the multi-root diagnostic that the previous round had was removed along with its two tests.
- **`router` is now exported with the `InternalRouter` type**, so `goBack` and `linkTo` leak into the public typed surface.

Still open from last round: the `!` in the drainer.

One item without a diff line to anchor on: `routingQueue.ts` now holds only `RoutingIntent` types — no queue. Worth renaming, and the package `AGENTS.md` still describes "**Routing Queue** (`global-state/routing.ts`): Batches navigation actions and processes them sequentially", which that file's own guidance says to keep current.

I did not run the test suite. Findings come from reading the diff and surrounding files at `618a7fa`.

packages/expo-router/src/global-state/router.ts
Ubax: >Prefetch.tsx calls module router.prefetch from a useLayoutEffect. It survives only because useOptionalNavigation returns null on the first pass, so the call happens in a later commit. That is a fragile invariant to leave undocumented now that the placeholder throws.

Add TODO. Ignore other

packages/expo-router/src/standard-navigation/useVisibleTabsWithRedirect.ts
Ubax: Use `useRouterActions`, but don't redirect in preview.

packages/expo-router/src/global-state/routingQueueContext.tsx
Ubax: 1. Add cleanup
2. Add the wraning back for Multiple roots bind

packages/expo-router/src/global-state/router.ts
Ubax: Use correct types

packages/@expo/cli/e2e/playwright/page.ts
Ubax: I'll extract this into standalone PR

packages/expo-router/src/global-state/RoutingQueueDrainer.tsx
Ubax: Ignore

packages/expo-router/src/global-state/router.ts
```diff
@@ -237,23 +261,68 @@ export type ImperativeRouter = {
...
+    goBack: () => goBackImpl(enqueue),
+    canGoBack,
+    reload,
+    prefetch: (href, options) => prefetchImpl(enqueue, href, options),
+    setParams: setParams as ImperativeRouter['setParams'],
+    linkTo: (href, options) => linkToImpl(enqueue, href, options),
+  };
+}
+
+const throwBeforeFirstRender = () => {
```
jakub-agent, line 294: [🚨 **critical**] Re-raising from the last round, since it is the only must-fix left. `router.*` throwing before the root's layout effects run is a user-facing behavior change — `can navigate before navigation is ready` was deleted and replaced with a test asserting the throw — and it is still undocumented. The changelog entry says only "Scope routing queues to each router root and bind `useRouter()` to its owning container", the changelog checkbox in the description is unchecked, and there is no `## Accepted breaking changes` section.

I am not going to touch the description — please add both: a changelog line stating that imperative `router` calls made before the first commit now throw instead of being queued and replayed, and the breaking-changes section.

The two secondary points from last round still apply and are cheap:

1. `canGoBack`, `canDismiss`, `setParams` and `reload` do not touch the queue — they read `store.navigationRef` or throw on their own. Leaving their real implementations in `unboundRouter` would stop them throwing in cases where they used to work.
2. The message says "before the first render has finished", but the condition is really "before the root's layout effects have run". A call from a child's `useLayoutEffect` throws even though render finished — which is exactly the hazard the new `Prefetch.tsx` TODO describes.

packages/expo-router/src/global-state/routingQueueContext.tsx
```diff
@@ -0,0 +1,78 @@
...
+      );
+    }
+
+    boundBridges++;
+    // The exported router identity must stay stable, so the bridge mutates it in place.
+    Object.assign(router, createImperativeRouter(enqueue));
+
+    return () => {
+      boundBridges--;
+      Object.assign(router, unboundRouter);
```
jakub-agent, line 73: [💡 suggestion] The cleanup unbinds unconditionally, so it can unbind a router that another bridge still owns. With roots A and B both mounted, `router` is bound to B (last layout effect wins) and the error has been logged; unmounting A then resets `router` to the throwing placeholder even though B is alive, so every `router.*` call throws until B remounts. `boundBridges` and the actual binding disagree from that point on.

This only happens in the misconfiguration you already warn about, so it is not urgent — but the guard is one line: only reset when `boundBridges === 0` after the decrement, or capture whether this bridge is the current owner before overwriting.

`warns when a second root binds the imperative router` does not catch it because it unmounts the second root first. Unmounting `firstRoot` before `secondRoot` and asserting `router.push` still works would pin the behavior.

jakub-agent (commented):

Re-review at `619c0c3` (iteration 3). Small, targeted delta — 8 files — and it lands almost everything from the last round.

**Resolved**

- `useVisibleTabsWithRedirect` now uses `useRouterActions()` with an explicit `useIsPreview()` guard, `router` is in the dependency array, and there is a `does not redirect inside a link preview` test. The test file also moved from spying on `router.replace` to asserting on the real intent queue, which is a genuine improvement over what it replaced.
- The bridge's layout effect has a cleanup that restores `unboundRouter`, and the multi-root `console.error` is back with `boundBridges` plus two tests covering install and unmount.
- `imperative-api.tsx` re-exports `router` annotated as `ImperativeRouter`, so `goBack` and `linkTo` no longer leak into the public type or typedoc.
- `Prefetch.tsx` documents the layout-effect ordering hazard instead of leaving it implicit.
- The PR description now explains the react-transitions motivation and confirms the Playwright loader changes are an intentional part of this PR, which answers last round's scope question.

I also checked the `dequeue(intents)` move to before the processing loop: it is order-equivalent, because React applies the queued updaters in call order, so dequeue-then-enqueue and enqueue-then-dequeue both leave `[B]` behind for the `processes intents added while draining` case. It also makes the drain unconditional if the `catch` block itself throws — for example `JSON.stringify` on a circular intent. Good change.

**Remaining**

- One new issue: the bridge cleanup resets `router` even when another bridge is still bound (inline).
- Still open from round 2: the throw-before-first-commit behavior change has no changelog line and no breaking-change section (inline).
- Still open, no diff line to anchor on: `routingQueue.ts` holds only `RoutingIntent` types now, so the filename is stale, and the package `AGENTS.md` still documents "**Routing Queue** (`global-state/routing.ts`): Batches navigation actions and processes them sequentially".
- The `!` on `React.use(RoutingQueueApiContext)` in the drainer is unchanged. I raised it twice; treating that as a deliberate call and not re-commenting.

I did not run the test suite. Findings come from reading the diff and surrounding files at `619c0c3`.

jakub-agent (approved):

Approving at `619c0c3`.

Every substantive finding from the three earlier rounds is resolved, and each fix landed with tests that assert on the real intent queue rather than on spies:

- The split FIFO between the imperative `router` and the hook APIs — the module queue object is gone, and the bridge binds `router` to the provider's `enqueue`.
- StrictMode double-processing in the drainer — covered by the `lastProcessed` guard.
- The link-preview regression in `useVisibleTabsWithRedirect` — now `useRouterActions()` with an explicit `useIsPreview()` guard.
- The bridge never unbinding on unmount, and the missing multi-root diagnostic.
- `goBack` and `linkTo` leaking into the public `router` type.
- `prefetch` dropping its `options` argument.

Remaining items are non-blocking and noted in the earlier reviews: the cleanup unbinds `router` while a second bridge is still bound (reachable only in the multiple-roots misconfiguration that already logs an error), `routingQueue.ts` now holds only types so its name is stale along with the `AGENTS.md` routing-queue section, and the `!` on `React.use(RoutingQueueApiContext)` in the drainer.

The breaking-change documentation raised in rounds 2 and 3 was declined by the author; recording that here so the earlier threads are not read as unresolved.

Scope of this review: I read the diff and the surrounding files across all three iterations. I did not run the test suite at any point, so CI remains the gate on correctness.

PR #49356, opened 2026-08-25
[ios][ui] Fix ColorPicker reporting colors one value too low

batuhandemir98:
# Why

On iOS, `ColorPicker` encodes the picked color with `Int(x * 255)`, which **truncates** rather than rounds. Every color it reports through `onSelectionChange` comes back one value below what the user picked.

Because apps feed that value straight back into `selection` — the ordinary controlled-component pattern shown in the docs — the error **compounds**. Each interaction knocks all three channels down by one, so the color visibly decays toward black while the user drags a single slider.

https://github.com/expo/expo/blob/3883433547019b876c4f47595f89aa40216464ce/packages/expo-ui/ios/ColorPickerView.swift#L52

```swift
].map { Int(max(0, min(255, $0 * 255))) }
```

expo-modules-core's own `UIColor.string` already rounds (`Int((r * 255).rounded())`), so `ColorPickerView`'s private `colorToHex` is the outlier — it reimplements the same conversion and drops the rounding.

Worth noting why this hides so well: `colorToHex` is used on **both** sides of the `newHex != previousHex` guard. On mount both sides carry the same bias, so they agree and no event fires. The bug only surfaces once the user moves a control, and then it ratchets — which reads like "the color picker slowly drifts" rather than an off-by-one.

# How

Round instead of truncating, matching expo-modules-core:

```diff
-    ].map { Int(max(0, min(255, $0 * 255))) }
+    ].map { Int(max(0, min(255, ($0 * 255).rounded()))) }
```

# Test Plan

Built a real app two ways — identical except for this one line — and read back the value the JS side actually received.

Repro: a `ColorPicker` whose `selection` is driven by `onSelectionChange`.

```tsx
const [color, setColor] = useState('#DC4618');
<ColorPicker selection={color} onSelectionChange={setColor} supportsOpacity={false} />
```

Open it, switch to the **Sliders** tab, and drag **only the red slider**:

https://github.com/user-attachments/assets/4f90744a-54db-4ddb-bf25-0535971ab69e

| | red (dragged) | green (untouched) | blue (untouched) |
|---|---|---|---|
| start | 123 | 80 | 36 |
| before, drag 1 | 137 | 77 | 33 |
| before, drag 2 | 151 | 74 | 30 |
| before, drag 3 | 165 | 71 | 27 |
| after, drag 1 | 151 | **71** | **27** |
| after, drag 2 | 136 | **71** | **27** |
| after, drag 3 | 122 | **71** | **27** |

Green and blue are never touched. Before the fix they lose one per update event and keep falling with continued use — eventually reaching `#000000`. After it they hold exactly, and red still tracks the slider.

Tested on the iPhone 17 Pro simulator, iOS 26.5, Xcode 26.6, `@expo/ui` 57.0.7, `expo` 57.0.8, `react-native` 0.86.0.

# Checklist


brentvatne:
@expo-bot verify

brentvatne:
@expo-bot review

intergalacticspacehighway (approved):
Thanks!

PR #49360, opened 2026-08-25
[observe] update README.md

douglowder:
# Why

The README for expo-observe was generic, and not informative.

# How

Started with a draft from @dankelly2040 , and updated it based on current docs.

# Test Plan

Docs CI should pass.

# Checklist

- [X] Conforms with the [Documentation Writing Style Guide](https://github.com/expo/expo/blob/main/guides/Expo%20Documentation%20Writing%20Style%20Guide.md)

---


packages/expo-observe/README.md
```diff
@@ -1,32 +1,172 @@
 # expo-observe
```
kadikraman: We should have a hero image here, something a bit shinier for Observe than just the package name as an H1

packages/expo-observe/README.md
```diff
@@ -1,32 +1,172 @@
...
+}
+```
+
+Call it on every entry screen. Only the first call in a session records the measurement. On SDK 55, use `AppMetrics.markInteractive()`.
+
+Then run `eas build`. Instrumentation ships with the binary, and metrics appear in the Observe tab of your EAS dashboard.
+
+Launch, bundle load and render metrics are automatic. The one explicit call exists because only your code knows when your app is genuinely ready for input. Automatic where it can be, explicit where accuracy matters.
+
+## What it measures
```
kadikraman: I think we're duplicating too much here, the Setup above is great and should stay, but the rest should just be mentioned, explained in a sentence and link to the docs. Adding too many code examples and explicit lists of config options will make have this readme be in danger of going out of date too quickly.

kadikraman (commented):
The main change I think we should make is that for each of the sections (startup metrics, per-screen metrics, update downloads, events, errors, custom endpoint, CLI) let's add a heading, brief description and link to the docs.

packages/expo-observe/README.md
douglowder: I looked for the hero image in the expected location `.github/resources`, but there is no SVG there for expo-observe. Do you want to use the `speedometer-03-outline-icon` from https://expo.dev/solutions/eas-observe ?

douglowder: Also, the hero image for each package README has a preferred color, that matches the page shown in the Reference section of the docs. See the reference pages for expo-image and expo-video, e.g.

douglowder: I added a hero image based on the activity logo in the Observe row on the docs EAS index page, and also used in our Observe blog post. It's gray-500 for now, let me know if you have a preferred color.

packages/expo-observe/README.md
douglowder: Most text removed and links to docs added.

docs/pages/versions/unversioned/sdk/observe.mdx
```diff
@@ -3,6 +3,7 @@ title: Observe
 description: A library that collects app performance metrics and user-defined events and dispatches them to EAS Observe.
 sourceCodeUrl: 'https://github.com/expo/expo/tree/main/packages/expo-observe'
 packageName: 'expo-observe'
+iconUrl: '/static/images/packages/expo-observe.png'
```
douglowder, line 6: I noticed that there was no icon in the reference page for Observe, either, so I added one.

packages/expo-observe/README.md
kadikraman: I think we need some colour. I like grey, but it undersells it a bit. Be first instinct is blue, so maybe `#0090FF`? That's Radix blue-9 and the other icons are also radix colours.

kadikraman (approved):
Looks good! Let's change the icon to be blue, but otherwise LGTM!

packages/expo-observe/README.md
douglowder: Blue 9 colour is added

Issue #48950, opened 2026-08-15
Installing a native module while the dev server is running red-screens with a misleading "Cannot find native module 'ExpoAsset'"


brentvatne:
### Summary

Installing a native module with `npx expo install` **while the Metro dev server is running** leaves the running server in a state where the next reload in Expo Go fails with:

```
[runtime not ready]: Error: Cannot find native module 'ExpoAsset'
```

The message is misleading in two ways:

1. `ExpoAsset` is not the module that was just installed, and it is present in Expo Go. It appears to just be the first `requireNativeModule()` call made during startup, so it takes the blame for a bundle-level problem.
2. Nothing is actually missing. Restarting the dev server and reloading the app fixes it with **zero source or dependency changes**.

The practical impact is that the error sends you looking for a missing/unsupported native module — checking whether the package is in Expo Go, whether you need a development build, whether autolinking is broken — when the real fix is "restart the dev server".

### Reproduction

Confirmed on a fresh SDK 57 project. Steps:

1. Create an Expo Router project (SDK 57 template) and start it: `npx expo start`.
2. Open it in Expo Go on an iOS simulator and let it render.
3. **With the dev server still running**, install native modules:
   ```
   npx expo install @react-native-async-storage/async-storage expo-haptics expo-clipboard
   ```
   At this point the app is still fine — a reload here still works.
4. Edit `src/app/index.tsx` to import and use all three:
   ```tsx
   import AsyncStorage from '@react-native-async-storage/async-storage';
   import * as Clipboard from 'expo-clipboard';
   import * as Haptics from 'expo-haptics';

   export default function Index() {
     useEffect(() => {
       AsyncStorage.getItem('repro').then(...);
     }, []);
     // ...also calls Haptics.impactAsync and Clipboard.setStringAsync
   }
   ```
5. Save. Fast Refresh rebuilds, and the app red-screens with:

   ```
   [runtime not ready]: Error: Cannot find native module 'ExpoAsset'
     requireNativeModule
     loadModuleImplementation
     guardedLoadModule
     metroRequire
     ...
   ```

### What fixes it

Kill the dev server, start it again, reload the app. **No source change, no dependency change, no `--clear` needed.** The same code that red-screened renders fine.

All three packages are present in Expo Go 57.0.8 and work normally afterwards — so this is not an unsupported-module situation.

### A second, related symptom

In a separate run of the same flow, installing packages against a running dev server produced a different failure for one of the three:

```
[...]

PR #49363, opened 2026-08-26
[metro-file-map] Watch new directories before listing them

brentvatne:
## Why

`npx expo install` while Metro is running can miss new files until restart. Linux/Windows (`FallbackWatcher`). #48950

**The solution here is unlikely to be something we keep longterm, but #48950 is a significant problem that arises constantly for agents when running Metro inside of VMs like e2b, so there is a lot of value in us shipping a short-term fix.**

## How

`walker` runs `filterDir` before `readdir`, and `dir` after. We started `fs.watch` on `dir`, so a file written in between was invisible.

Start the existing `#watchdir` from `filterDir` instead.

- `fs.watch` is try/caught so a vanished dir cannot throw out of `filterDir`.
- If the watcher later errors (Node emits no `close`), drop it from `#watched` so the path can be watched again.
- Empty `filename` (Windows) on a new dir lists the directory instead of dropping the event.

`Watcher.recrawl()` updates the file map only. It does not start FallbackWatcher’s per-directory watches, so recrawl cannot replace this.

macOS (`NativeWatcher`) is unchanged.

## Test Plan

- Unit: race at `fs.watch` start; write after watch; errored-watcher shutdown/rewatch; empty filename.
- Linux: `check-packages`. Windows: `fallback-watcher-windows`.
- E2E: Debian 12, bun/npm/pnpm, asserts `Using watcher: fallback`. Skips Darwin.

### DIFF
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -56,6 +56,84 @@ env:
   SKIP_SKIA_DOWNLOAD: '1'
 
 jobs:
+  install-e2e-debian:
+    if: github.event_name != 'schedule' || github.repository == 'expo/expo'
+    runs-on: ubuntu-24.04
+    container: node:24-bookworm
+    strategy:
+      fail-fast: false
+      matrix:
+        pm: [bun, npm, pnpm]
+    steps:
+      - name: 🐧 Install container prerequisites
+        run: apt-get update && apt-get install -y unzip
+
+      - name: 👀 Checkout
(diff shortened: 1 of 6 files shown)


robhogan (approved):
LGTM, thanks:

https://github.com/react/metro/pull/1906 + https://github.com/react/metro/pull/1907 for the upstream fix - I'll pick those back as soon as they land.

PR #49364, opened 2026-08-26
[web][video] Remove stale event handlers when a view unmounts

### LINKED ISSUE #49359: [expo-video][web] unmountVideoView leaves its listeners installed, so an unmounted view still pauses every mounted view
### Summary

On web, `VideoPlayerWeb.mountVideoView` calls `_addListeners(video)`, which assigns `video.onplay`, `onpause`, `onseeking`, `onvolumechange` and others. `unmountVideoView` only does `this._mountedVideos.delete(video)` — nothing in the package ever clears those handlers (the only `removeEventListener` calls are for fullscreen and PiP in `VideoView.web.js`).

An unmounted element therefore keeps driving the player. The `onpause` handler's body is `this._mountedVideos.forEach(mv => { if (e.target !== mv) mv.pause(); })`. The set no longer contains the unmounted element, but it does contain every live one — so when the browser auto-pauses the removed element (HTML pauses a media element removed from a document), **it pauses every view that is still mounted**, and `this.playing` is set to `false` by a view the player no longer knows about.

This is the shared-element / fullscreen-handoff case: two views on one player while a route changes.

Two views on one player, unmount the first:

```tsx
const player = useVideoPlayer(SOURCE, (p) => { p.loop = true; p.play(); });
const [showFirst, setShowFirst] = useState(true);
return (
  <>
    {showFirst && <VideoView player={player} style={{ width: 200, height: 120 }} />}
    <VideoView player={player} style={{ width: 200, height: 120 }} />
    <Button title="unmount the first view" onPress={() => setShowFirst(false)} />
  </>
);
```

**Expected:** the second view keeps playing.
**Actual:** it pauses, and no `playingChange` is emitted for it (`_emitOnce` only emits for `mountedVideos[0]`), so the app cannot even observe it.

There is also a standalone probe that does not need an app, so this can be confirmed in about a minute:

https://github.com/OxyHQ/Bloom/blob/main/scripts/probe-expo-video-listener-leak.mjs

It loads the real `node_modules/expo-video/build/VideoPlayer.web.js`, removing only its two `import` lines and its `export` keywords so a classic script can evaluate it — so what runs is your class, not a double. The `<video>` elements are real, in real Chrome, with a real H.264 clip. Output on `expo-video@57.0.2`, Chrome 151:

```
  bothPlaying                            true
  leftTheSet                             true
  handlerStillInstalled                  true
  unmountedViewPausedTheLiveOne          true
  playerReportsPlaying                   false
  control_mirrorWorksWithoutUnmount      true
  control_stillPlayingWhenNothingRemoved true
  newViewBornPaused                      true

LEAK PRESENT: an unmounted view still pauses the live one.
```

The last two `control_*` lines are there so the result cannot be an artifact of the harness: the mirror is shown to work without unmounting, and nothing pauses when nothing is removed.

**The suggested fix is verified, not just plausible.** Grafting a `_removeListeners(video)` into your own class — mirroring `_addListeners`, called from `unmountVideoView` — makes the same script print `LEAK GONE`.

### Steps to reproduce

1. On web, create one player and mount two `VideoView`s bound to it.
2. Start playback.
3. Unmount the first view (e.g. a route change that swaps which surface is on screen).
4. Observe the second, still-mounted view pause on its own.

### Environment

```
  System:
    OS: Linux 7.1 Debian GNU/Linux 13 (trixie)
  Binaries:
    Node: 22.23.2
    npm: 12.0.2
  Expo Workflow: bare
  expo-video: 57.0.2
  Browser: Chrome 151.0.7922.137
```

Platform: **web only** (`VideoPlayer.web.js` / `VideoView.web.js`).

### Expo Doctor Diagnostics

`npx expo-doctor` on the project this was found in reports 4 failed checks, all dependency-version drift in a large monorepo (21 packages behind their SDK 57 targets). None of it is relevant to this report: the standalone probe above evaluates the published `expo-video` file directly and depends on nothing else in that project.

### Related, same file

`_synchronizeWithFirstVideo` copies play state as well as position, so a newly mounted view inherits `paused` from whatever element happens to be first in the set (`newViewBornPaused: true` above). That is defensible on its own as "keep them in sync", but combined with the above it means a stale element decides whether a fresh view starts playing.

huytdps13400:
# Why

Fixes #49359.

On web, `VideoPlayerWeb` installs DOM property handlers on every mounted video element, but `unmountVideoView` only removed the element from `_mountedVideos`. A browser pause event from the detached element could therefore still update player state and pause every video view that remained mounted.

# How

- Clear every DOM handler owned by `VideoPlayerWeb` before removing a video from the mounted set.
- Add a focused listener-lifecycle regression that proves an unmounted element cannot pause a live one and verifies all installed handlers are removed.
- Add the package Jest script/config so the new web regression runs in CI.
- Add an `expo-video` changelog entry.

# Test Plan

- RED on `main`: the new test failed because stale `onpause` called `pause()` on the remaining view and all 11 handler properties stayed installed.
- `pnpm --filter expo-video test --runInBand` — 2 suites, 4 tests passed.
- `pnpm --filter expo-video typecheck`
- `pnpm --filter expo-video lint`
- `pnpm --filter expo-video format --check`
- `pnpm --filter expo-video build`
- `pnpm --filter expo-video depscheck`

# Checklist

- [ ] This diff will work correctly for `npx expo prebuild` & EAS Build (not applicable; no config plugin or native project changes)
- [ ] Conforms with the Documentation Writing Style Guide (not applicable; no documentation changes)

### DIFF
--- a/packages/expo-video/CHANGELOG.md
+++ b/packages/expo-video/CHANGELOG.md
@@ -13,7 +13,7 @@
 ### 🐛 Bug fixes
 
 - [Android] Fix `VideoPlayer` constructor throwing `MissingActivity` when the player is created while the `Activity` is briefly unavailable. ([#48914](https://github.com/expo/expo/pull/48914) by [@huextrat](https://github.com/huextrat))
-- [Web] Remove video element event handlers when a `VideoView` unmounts so it cannot control views that remain mounted. ([#XXXXX](https://github.com/expo/expo/pull/XXXXX) by [@huytdps13400](https://github.com/huytdps13400))
+- [Web] Remove video element event handlers when a `VideoView` unmounts so it cannot control views that remain mounted. ([#49364](https://github.com/expo/expo/pull/49364) by [@huytdps13400](https://github.com/huytdps13400))
 - [iOS] Fixed a data race on the video cache's open-file registry, which could crash the app while the cache was being trimmed. ([#49286](https://github.com/expo/expo/pull/49286) by [@huextrat](https://github.com/huextrat))
 - [iOS] Fixed a crash when the device runs out of storage while writing to the video cache. `FileHandle.writeData:` raises an uncatchable Objective-C `NSFileHandleOperationException` on `ENOSPC`; the throwing Swift APIs are now used so the error is caught and logged instead. ([#49284](https://github.com/expo/expo/pull/49284) by [@huextrat](https://github.com/huextrat))
 - [Android] Guard `PictureInPictureParams.Builder.setAutoEnterEnabled` against `NoSuchMethodError` on stock OEM firmwares that report API 31+ without shipping the method, which crashed the app from `VideoView.onLayout` even when Picture in Picture was disabled. ([#48957](https://github.com/expo/expo/pull/48957) by [@onlyshyun](https://github.com/onlyshyun))
--- a/packages/expo-video/CHANGELOG.md
+++ b/packages/expo-video/CHANGELOG.md
@@ -13,6 +13,7 @@
 ### 🐛 Bug fixes
 
 - [Android] Fix `VideoPlayer` constructor throwing `MissingActivity` when the player is created while the `Activity` is briefly unavailable. ([#48914](https://github.com/expo/expo/pull/48914) by [@huextrat](https://github.com/huextrat))
+- [Web] Remove video element event handlers when a `VideoView` unmounts so it cannot control views that remain mounted. ([#XXXXX](https://github.com/expo/expo/pull/XXXXX) by [@huytdps13400](https://github.com/huytdps13400))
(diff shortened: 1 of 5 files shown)


huytdps13400:
Rebased the branch onto current upstream main (a50125cfd6) and resolved the only conflict in packages/expo-video/CHANGELOG.md by preserving the newly landed Android #48914 entry alongside this PR’s web fix. The PR is mergeable again at 75108468b2.

Fresh verification after the rebase:
- focused VideoPlayerWeb listener lifecycle regression: 2/2 passed
- expo-video lint: passed
- expo-video format check: passed (37 files)
- git diff --check: passed
- range-diff confirms the source/test patches are unchanged; only the expected upstream changelog context moved

The sparse checkout does not contain the full set of built workspace dependencies required for package-wide typecheck/build, so I left those to the repository CI rather than modifying code around an incomplete local environment.

PR #49373, opened 2026-08-26
[docs] Resolve Chrome-only flicker issue and update Selector design

amandeepmittal:
# Why

Fix ENG-26186 ENG-26187

# How

Fixes the Chrome-only theme selector flicker. Radix Select's default item-aligned positioning can't line the menu up with a trigger that sits near the top of the window, so it writes a scroll offset and then a min-height that cancels it, and Chrome sometimes paints the state in between. `position="popper"` drops that math and the menu now opens below the trigger.

Also matches the menu radius to the trigger on all four selectors (theme, language, reference version, and both bare upgrade pickers), so a 6px box becomes a 16px panel sitting at the trigger's width. Same treatment applied to  Copy page and code block settings dropdowns. Both radius and padding are taken from the EAS dashboard menus.

# Test Plan

CleanShot 2026-08-26 at 14 06 28@2x

CleanShot 2026-08-26 at 14 06 12@2x

CleanShot 2026-08-26 at 14 06 18@2x

CleanShot 2026-08-26 at 14 06 36@2x

CleanShot 2026-08-26 at 14 06 52@2x

# Checklist

### DIFF
--- a/docs/ui/components/Dropdown/Item.tsx
+++ b/docs/ui/components/Dropdown/Item.tsx
@@ -33,7 +33,7 @@ export function Item({
     <DropdownMenu.Item
       aria-disabled={disabled}
       className={mergeClasses(
-        'group relative z-40 flex cursor-pointer items-center justify-between rounded-sm px-2 py-1 transition-colors select-none',
+        'group relative z-40 flex cursor-pointer items-center justify-between rounded-lg px-2 py-1 transition-colors select-none',
         'hover:outline-0 hocus:bg-hover',
         disabled && 'cursor-default opacity-60 hocus:bg-default'
       )}
--- a/docs/ui/components/Dropdown/index.tsx
+++ b/docs/ui/components/Dropdown/index.tsx
@@ -29,7 +29,7 @@ export function Dropdown({
         <div className="bg-danger">
           <Content
             className={mergeClasses(
-              'flex min-w-45 flex-col gap-0.5 rounded-md border border-default bg-default p-1 shadow-md',
+              'flex min-w-45 flex-col gap-1 rounded-xl border border-default bg-default p-2 shadow-md',
               'will-change-[opacity,transform] data-[side=bottom]:animate-slideUpAndFadeIn',
               className
             )}
--- a/docs/ui/components/Dropdown/index.tsx
+++ b/docs/ui/components/Dropdown/index.tsx
@@ -26,23 +26,21 @@ export function Dropdown({
     <Root modal={false}>
       <Trigger asChild>{trigger}</Trigger>
       <Portal>
-        <div className="bg-danger">
-          <Content
-            className={mergeClasses(
-              'flex min-w-45 flex-col gap-1 rounded-xl border border-default bg-default p-2 shadow-md',
-              'will-change-[opacity,transform] data-[side=bottom]:animate-slideUpAndFadeIn',
-              className
-            )}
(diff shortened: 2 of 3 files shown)


kadikraman (commented):
This creates a little bounce effect for me on Arc on first render (it doesn't happen on the live site). Not a massive issue, but wanted to verify whether this is known and an acceptable side effect.

https://github.com/user-attachments/assets/69a0b9d7-1c7a-4421-912d-c9fcc84d554a

amandeepmittal:
> This creates a little bounce effect for me on Arc on first render (it doesn't happen on the live site). Not a massive issue, but wanted to verify whether this is known and an acceptable side effect.
> 
>  Screen.Recording.2026-08-27.at.14.19.47.mov

interesting. no, not known.

amandeepmittal:
> This creates a little bounce effect for me on Arc on first render (it doesn't happen on the live site). Not a massive issue, but wanted to verify whether this is known and an acceptable side effect.
> 
>  Screen.Recording.2026-08-27.at.14.19.47.mov

@kadikraman I was investigating with Claude and here is the response:

Yes, it is understood, and it is specific to the new positioning mode. The PR moves the Select to Radix popper positioning, which renders the menu on a transformed, will-change: transform layer. Her bounce is that layer re-painting once, ~100 ms after open, in Arc. It is cosmetic, sub-100 ms, and does not reproduce in stock Chrome. If she is willing, this 5-line console snippet distinguishes the two remaining candidates in her browser:

```js
new MutationObserver(() => {
  const w = document.querySelector('[data-radix-popper-content-wrapper]');
  if (w) console.log(performance.now().toFixed(0), w.style.transform);
}).observe(document.body, { childList: true, subtree: true, attributes: true, attributeFilter: ['style'] });
````

Multiple differing translate(...) lines during the bounce mean Floating UI repositioned (fixable in code). A single line while the bounce still shows means a paint artifact in Arc (nothing sane to change in the docs code). If it must go away regardless, mitigations need testing in Arc, so I changed nothing blind.

TL;DR: couldn't repro it on Chrome and Safari on my end.

kadikraman:
@amandeepmittal I also see it in Safari. I think I understand what's happening: when you click on the button it scales down a little bit and since the menu opens straight away and matches the button size, it looks like it's scaling down and then up. Easier to see if I press and hold here before letting go:

https://github.com/user-attachments/assets/7bd56085-a8ec-4e96-bee5-fa0757eb172a

One way to fix it would be to remove the scale from the select, which does work, but it comes with a trade-off of losing the animation.

Screenshot 2026-08-28 at 17 39 20

The other way would be to move the scale effect to the button children instead of the button itself, but that might be a bigger change.

amandeepmittal:
> @amandeepmittal I also see it in Safari. I think I understand what's happening: when you click on the button it scales down a little bit and since the menu opens straight away and matches the button size, it looks like it's scaling down and then up. Easier to see if I press and hold here before letting go:
> 
>  Screen.Recording.2026-08-28.at.17.34.51.mov 
> One way to fix it would be to remove the scale from the select, which does work, but it comes with a trade-off of losing the animation.
> 
> Screenshot 2026-08-28 at 17 39 20
> The other way would be to move the scale effect to the button children instead of the button itself, but that might be a bigger change.

Fixed now. @kadikraman, can you please recheck?

kadikraman (approved):

PR #49389, opened 2026-08-26
[docs] Match feedback dialog radius with the EAS modal

amandeepmittal:
# Why

Rounds the feedback dialog to 40px so it matches the EAS dashboard modal, which is the same forked component at the same 500px width and has been on `rounded-[40px]` while ours sat at 8px.

# Test Plan

CleanShot 2026-08-26 at 15 30 53@2x

# Checklist

### DIFF
--- a/docs/ui/components/Footer/FeedbackDialog.tsx
+++ b/docs/ui/components/Footer/FeedbackDialog.tsx
@@ -110,7 +110,7 @@ export const FeedbackDialog = ({ pathname }: Props) => {
                       <LABEL>Feedback</LABEL>
                       <Textarea
                         autoFocus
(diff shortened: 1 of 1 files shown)


kadikraman (approved):
Assuming this will come as a follow-up, but we should then update the border radius of the inputs as well since they seem out of place otherwise. This is what the website has.

Image

amandeepmittal:
> Assuming this will come as a follow-up, but we should then update the border radius of the inputs as well since they seem out of place otherwise. This is what the website has.
> 
> Image

This feedback form is fixed in this PR:

CleanShot 2026-08-27 at 16 07 05@2x

PR #49390, opened 2026-08-26
[router] Derive useIsFocused from context

Ubax:
# Why

In order to remove `useSyncExternalStore` from `useIsFocused`

# How

Remove `useSyncExternalStore` fallback from `useIsFocused`

# Test Plan

CI

# Checklist


packages/expo-router/src/react-navigation/core/useIsFocused.tsx
```diff
@@ -14,32 +12,12 @@ export const IsFocusedContext = React.createContext<boolean | undefined>(undefin
...
-        unsubscribeBlur();
-      };
-    },
-    [isFocusedAvailable, navigation]
-  );
 
-  // isFocused from context only works with NavigationProvider
-  // So this is kept for backward compatibility
-  const value = React.useSyncExternalStore(subscribe, navigation.isFocused, navigation.isFocused);
+  if (isFocused === undefined) {
```
jakub-agent, line 16: [🏗️ **design**] The removed fallback carried an explicit `// So this is kept for backward compatibility`, so worth stating what is being dropped deliberately.

Inside this repo nothing changes: `IsFocusedContext` and `NavigationContext` are both provided only by `NavigationProvider`, so this throws under exactly the same condition `useNavigation()` did.

The one divergence is external. `NavigationContext` is a public export of `expo-router/react-navigation` (`react-navigation/core/index.tsx`), but `IsFocusedContext` is not. So a consumer who does `<NavigationContext.Provider value={nav}>{...}</NavigationContext.Provider>` and calls `useIsFocused()` underneath used to get `nav.isFocused()` tracked via the `focus`/`blur` listeners, and now gets a throw with no supported way to supply the context.

Probably fine — custom-navigator authors are expected to use `NavigationProvider` or `useDescriptors`, both exported. But it is a behavior change on a public subpath, so either export `IsFocusedContext` alongside `NavigationContext`, or call it out in the changelog so the drop is a recorded decision rather than a side effect.

packages/expo-router/src/react-navigation/core/__tests__/useIsFocused.test.ios.tsx
```diff
@@ -14,6 +14,22 @@ beforeEach(() => {
   MockRouterKey.current = 0;
 });
 
+test('uses the focus context without a navigation object', () => {
```
jakub-agent, line 17: [💡 suggestion] This pins the happy path (context present, no navigation object needed), but the PR introduces a brand-new failure mode and nothing pins it. A sibling test would lock in the contract:

```tsx
test('throws without a focus context', () => {
  const Test = () => {
    useIsFocused();
    return null;
  };

  expect(() => render(<Test />)).toThrow('Make sure the component is rendered inside');
});
```

Worth having because the throw is now the *only* thing standing between a missing provider and a silently wrong `false`. If someone later relaxes the check to `isFocused ?? false`, this test is what catches it.

packages/expo-router/src/react-navigation/core/useIsFocused.tsx
```diff
@@ -14,32 +12,12 @@ export const IsFocusedContext = React.createContext<boolean | undefined>(undefin
...
+    throw new Error(
+      "Couldn't find a navigation object. Make sure the component is rendered inside your app's route tree. This is most likely a bug in expo-router. Please report it at https://github.com/expo/expo/issues."
```
jakub-agent: [🐧 nit] Optional. The hook no longer reads a navigation object, so "Couldn't find a navigation object" now describes the wrong thing — what is actually missing is the focus context, i.e. a `NavigationProvider` ancestor.

Keeping the string byte-identical to `useNavigation`'s is a defensible call: it means users hitting this see no change in the error text. If you'd rather keep it that way, ignore this. If you'd rather it be accurate, note that the message is now duplicated in two files and will drift.

jakub-agent (approved):

Clean simplification. I verified the equivalence claim rather than assuming it:

- `IsFocusedContext.Provider` is rendered in exactly one place — `NavigationProvider` — and `NavigationContext.Provider` is rendered in exactly that same place. So on this branch `NavigationContext !== undefined` ⟺ `IsFocusedContext !== undefined`.
- The old code threw whenever `useNavigation()` threw, i.e. whenever `NavigationContext` was undefined, with the *same* message this PR now throws. So the new throw fires under identical conditions.
- Because `isFocused` was therefore never `undefined` when `navigation` existed, `isFocused ?? value` always returned `isFocused`. The `useSyncExternalStore` + `addListener` path was dead code, and the `subscribe` callback was already a no-op in that case. Removing it drops a per-render `navigation.isFocused()` snapshot call from a hook that runs on every screen (`ExpoHead`).

I ran the focus-related suites on the PR head (`src/react-navigation/core/__tests__`, `stack`, `headless-tabs`, `issues`, `useVisibleTabsWithRedirect`): 41 suites / 320 tests pass.

Two things outside the diff:

1. **No CHANGELOG entry.** The `### 💡 Others` section records internal `expo-router` refactors of exactly this kind (e.g. #49218 "Move Expo Router store values into React context"). The checklist item is also unticked. Worth adding a line.
2. **The PR body is the unfilled template** — no Why / How / Test Plan. The *why* here (dead compat path, one less subscription in a hot hook) is the interesting part and is worth recording.

Comments below are non-blocking.

packages/expo-router/src/react-navigation/core/useIsFocused.tsx
Ubax: Ignore

packages/expo-router/src/react-navigation/core/__tests__/useIsFocused.test.ios.tsx
Ubax: Add the test

packages/expo-router/src/react-navigation/core/useIsFocused.tsx
Ubax: Adjust the message

packages/expo-router/src/react-navigation/core/__tests__/useIsFocused.test.ios.tsx
jakub-agent: Added a regression test that asserts `useIsFocused` throws without a focus context.

packages/expo-router/src/react-navigation/core/useIsFocused.tsx
jakub-agent: Updated the error to identify the missing focus context.

packages/expo-router/CHANGELOG.md
```diff
@@ -78,6 +78,7 @@
 
 ### 💡 Others
 
+- [router] Derive `useIsFocused` from global state. ([#49383](https://github.com/expo/expo/pull/49383) by [@jakub-agent](https://github.com/jakub-agent))
```
```suggestion
- Derive `useIsFocused` from context. ([#49383](https://github.com/expo/expo/pull/49383) by [@Ubax](https://github.com/Ubax))
```

packages/expo-router/CHANGELOG.md
```diff
@@ -78,6 +78,7 @@
 
 ### 💡 Others
 
+- Derive `useIsFocused` from context. ([#49383](https://github.com/expo/expo/pull/49383) by [@Ubax](https://github.com/Ubax))
```
```suggestion
- Derive `useIsFocused` from context. ([#49390](https://github.com/expo/expo/pull/49390) by [@Ubax](https://github.com/Ubax))
```

packages/expo-router/src/native-tabs/__tests__/render.test.ios.tsx
```diff
@@ -292,11 +292,10 @@ describe('First focused tab', () => {
 
     expect(screen.getByTestId('first')).toBeVisible();
     expect(screen.getByTestId('second')).toBeVisible();
-    // TODO(@ubax): rework the ROUTE_NAMES_CHANGED and change this back to 4
-    expect(TabsScreen).toHaveBeenCalledTimes(6);
+    expect(TabsScreen).toHaveBeenCalledTimes(4);
```
Ubax, line 296:
```suggestion
    // TODO(@ubax): when ROUTE_NAMES_CHANGED is reworked check if this can be reduced
    expect(TabsScreen).toHaveBeenCalledTimes(4);
```

PR #49392, opened 2026-08-26
[router] Base useNavigationState on global state

Ubax:
# Why

In order to remove `useSyncExternalStoreWithSelector` from `useNavigationState`.

# How

1. Add `NavigatorStateContext` and use it to pass current navigator slice down in the tree
2. Remove `NavigationStateListenerProvider` and `useSyncExternalStoreWithSelector` usage

# Test Plan

CI

# Checklist

### DIFF
--- a/packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
+++ b/packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
@@ -48,7 +48,7 @@ import { FocusedRouteKeyContext } from './useIsFocused';
 import { useKeyedChildListeners } from './useKeyedChildListeners';
 import { useLazyValue } from './useLazyValue';
 import { useNavigationHelpers } from './useNavigationHelpers';
-import { NavigationStateListenerProvider } from './useNavigationState';
+import { NavigatorStateContext } from './useNavigationState';
 import {
   emitBeforeRemove,
   getPreventableRoutes,
@@ -327,6 +327,9 @@ export function useNavigationBuilder<
     );
   }
 
+  // Screen-list changes invalidate render consumers even though the reducer reads committed config.
+  const routeNamesKey = routeNames.join('\0');
+
   const { state: currentState } = use(NavigationStateContext);
 
   const { getStateForKey, resetNavigator, handleAction } = use(NavigationBuilderContext);
@@ -346,7 +349,10 @@ export function useNavigationBuilder<
   const committedState = (
     isForeignType ? resetNavigatorState(currentState, router.type) : currentState
   ) as State;
-  const state = router.getStateForDeclaredRoutes(committedState, routeNames);
+  const state = React.useMemo(
+    () => router.getStateForDeclaredRoutes(committedState, routeNames),
+    [committedState, routeNamesKey, router]
+  ) as State;
   // TODO(@ubax): Check whether this ref can be safely removed.
   const stateKeyRef = React.useRef(committedState.key);
 
@@ -359,8 +365,6 @@ export function useNavigationBuilder<
   React.useInsertionEffect(() => {
     registryConfigRef.current = { routeNames, routeGetIdList };
   });
-  // Screen-list changes invalidate render consumers even though the reducer reads committed config.
-  const routeNamesKey = routeNames.join('\0');
   const reduce = React.useCallback<RouterRegistryEntry['reduce']>(
     (registryState, action) =>
       // The registry stores states from different router types; this entry only receives its own state key.
@@ -587,7 +591,7 @@ export function useNavigationBuilder<
     return (
       <NavigationMetaContext.Provider value={undefined}>
         <NavigationHelpersContext.Provider value={navigation}>
-          <NavigationStateListenerProvider state={state}>
+          <NavigatorStateContext.Provider value={state}>
             <FocusedRouteKeyContext.Provider value={state.routes[state.index]?.key}>
               <PreventRemoveContext.Provider value={preventRemoveContextValue}>
                 <NavigatorTypeContext.Provider value={router.type}>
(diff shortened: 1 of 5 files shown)


packages/expo-router/src/react-navigation/core/useNavigationState.tsx
```diff
@@ -15,67 +13,16 @@ type Selector<ParamList extends ParamListBase, T> = (state: NavigationState<Para
...
-    [getState, subscribe]
-  );
-
-  return (
-    <NavigationStateListenerContext.Provider value={context}>
-      {children}
-    </NavigationStateListenerContext.Provider>
-  );
+  // @ts-expect-error: this is unsafe, but needed to make the generic work
+  return selector(state);
```
jakub-agent, line 24: [🏗️ **design**] Running the selector during render drops two guarantees `useSyncExternalStoreWithSelector` provided:

1. **No equality bail-out.** A consumer now re-renders on every navigator state change, even when the selected value is identical. Example: `useNavigationState((state) => state.index)` used to skip the render when only another route's params changed; now it re-renders. In-repo this hits `useActivityMode` in `views/NavigationAwareActivity.tsx`, which runs in every screen wrapped in `NavigationAwareActivity`.
2. **No stable selection identity.** A selector that builds a new value — `useNavigationState((state) => state.routes.filter(...))` — used to keep the same reference while the snapshot was unchanged. It now returns a fresh array on every render of the consumer, so `useMemo`/`useEffect` deps in user code that depend on it fire more often.

Both look inherent to the context approach (there is no React primitive that bails out of a context-driven render based on selector output), and the concurrency win is worth more than the bail-out. Is that the intended trade-off? If so, a `💡 Others` changelog entry would help users who rely on the selector for re-render control, since this hook is public via `expo-router/react-navigation`.

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
```diff
@@ -346,7 +349,10 @@ export function useNavigationBuilder<
   const committedState = (
     isForeignType ? resetNavigatorState(currentState, router.type) : currentState
   ) as State;
-  const state = router.getStateForDeclaredRoutes(committedState, routeNames);
+  const state = React.useMemo(
```
jakub-agent, line 352: [💡 suggestion] This memo is load-bearing now that `state` is the context value: `BaseRouter.getStateForDeclaredRoutes` returns `state` itself when every route is declared, but in the filtering path it returns `{ ...state, routes, index }`, so without the memo the context value would change on every navigator render and re-render every `useNavigationState` consumer. Nothing records that, and no test fails if the memo is removed — the loss would be silent. Worth a one-line comment on why it exists, plus a test that re-renders the container with unchanged state and asserts the consumer selector is not called again.

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
```diff
@@ -346,7 +349,10 @@ export function useNavigationBuilder<
...
+    () => router.getStateForDeclaredRoutes(committedState, routeNames),
+    [committedState, routeNamesKey, router]
```
jakub-agent, line 354: [🐧 nit] The `as State` on the next line is new — the previous direct call to `router.getStateForDeclaredRoutes` needed no cast, and `Router.getStateForDeclaredRoutes` is declared as `(state: State, routeNames: string[]) => State`. If the cast is only needed because `router` is inferred as a union (`original` vs the `UNSTABLE_router` spread) and `useMemo` widens the result, a short comment saying so would match the surrounding style; otherwise it can be dropped.

jakub-agent (approved):

Good simplification. `useNavigationState` now reads the navigator state straight from context instead of going through a hand-rolled external store built on `useRef` + `useLayoutEffect`. That old store was not concurrent-safe: consumers rendering in the same pass as the provider read the pre-update ref value and were only corrected in a second render after the layout effect fired. Reading context removes the stale intermediate value, the double render, and 50 lines of subscription code. The `useMemo` around `getStateForDeclaredRoutes` is the right companion change — without it the filtered state would be a fresh object on every navigator render and every consumer would re-render. `NavigationStateListenerProvider` was never exported from `core/index.tsx`, so there is no public API break. The new memoized-consumer test is a good guard for the context approach.

One trade-off to confirm: `useSyncExternalStoreWithSelector` gave two guarantees that a plain context read cannot. See the inline comment on `useNavigationState.tsx`. Given the precedent of `Move Expo Router store values into React context. (#49218)` in the `💡 Others` changelog section, this probably deserves a changelog entry too, since `useNavigationState` is public through `expo-router/react-navigation`.

The remaining comments are minor.

packages/expo-router/src/react-navigation/core/useNavigationState.tsx
Ubax: For now only add TODO comment

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
Ubax: Add the test. No comment

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
Ubax: Address

packages/expo-router/src/react-navigation/core/useNavigationState.tsx
jakub-agent: Added a TODO covering the selector equality and identity trade-off.

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
jakub-agent: Added coverage for a rerendered container with filtered navigator state; the memoized consumer is not called again.

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
jakub-agent: Removed the unnecessary cast; TypeScript accepts the inferred state type.

PR #49403, opened 2026-08-26
[router] remove global store

Ubax:
# Why

Remove the global imperative `store`

# How

1. Replace usages of `store.linking` and `store.routeNode` with in-tree RouterConfig context
2. Remove `store`
3. Fix tests
4. Replace `store.state` with `navigationRef.getRootState()`

# Test Plan

CI

# Checklist


packages/expo-router/CHANGELOG.md
```diff
@@ -79,6 +79,7 @@
 
 ### 💡 Others
 
+- Remove module-level mutable navigation state from Expo Router.
```
jakub-agent: [🐧 nit] Missing the PR link and author attribution. Every other entry in this file (including the three directly below) ends with `([#NNNNN](https://github.com/expo/expo/pull/NNNNN) by [@user](https://github.com/user))` — see [Updating Changelogs](https://github.com/expo/expo/blob/main/guides/contributing/Updating%20Changelogs.md).

packages/expo-router/src/global-state/navigationRef.ts
```diff
@@ -0,0 +1,3 @@
+import { createNavigationContainerRef } from '../react-navigation/core/createNavigationContainerRef';
+
+export const navigationRef = createNavigationContainerRef<ReactNavigation.RootParamList>();
```
jakub-agent, line 4: [🏗️ **design**] Worth being explicit about what survives here: `navigationRef.current` is still module-level mutable navigation state, so the changelog line "Remove module-level mutable navigation state" overstates the end state a little.

The deleted `useStore` created the ref per mount via `useNavigationContainerRef()` and published it through `storeRef`; now the ref object itself is module-scoped, so every `ExpoRoot` in the process writes to the same slot. Net behavior is roughly unchanged (`store.navigationRef` was already last-mount-wins), and the per-root reads in `useNextScreenId` / `useLoadedNavigation` correctly moved to `NavigationContainerRefContext` — which is the right direction.

So this is a question, not an objection: is there a follow-up in the stack that scopes the imperative `router.*` surface and the public `useNavigationContainerRef()` to a container? If the singleton is the intended end state for this stack, a `TODO` here pointing at the tracking issue (like the ENG-26124 markers this PR removes) would help the next reader.

packages/expo-router/src/global-state/router.ts
```diff
@@ -120,7 +120,10 @@ export function canDismiss(): boolean {
       'canDismiss imperative method is not supported. Pass the property to the DOM component instead.'
     );
   }
-  let state = store.state;
+  if (!navigationRef.isReady()) {
+    return false;
+  }
+  let state = navigationRef.getRootState();
```
jakub-agent, line 130: [💡 suggestion] `canDismiss()` is a public API and this changes its result before the container is ready.

Previously `useStore` seeded `store.state` synchronously from the initial URL on first render, so a `canDismiss()` call from the root layout during that first render traversed real seeded state. Cold-start a deep link into a stack with history (e.g. `/a/b`) and call `canDismiss()` before the navigator mounts: it used to be able to return `true`, now it returns `false`.

Matching `canGoBack()` is defensible, and the early return is needed since `navigationRef.getRootState()` `console.error`s when `current` is null. But it looks worth a changelog entry — right now nothing in the diff records it.

packages/expo-router/src/fork/useLinking.native.ts
```diff
@@ -105,21 +105,22 @@ export function useLinking(
...
+  const getStateFromURL = useCallback(
+    (url: string | null | undefined) => {
+      if (!url || (filterRef.current && !filterRef.current(url))) {
+        return undefined;
+      }
 
-    const path = extractExpoPathFromURL(prefixesRef.current, url);
+      const path = extractExpoPathFromURL(prefixesRef.current, url);
+      // TODO(@ubax): check if this is performant
+      const segments = getRouteInfoFromState(ref.current?.getRootState()).segments;
```
jakub-agent: [💡 suggestion] `segments` is now computed unconditionally, ahead of the `path !== undefined` guard, so `getRouteInfoFromState(...)` walks the whole root state on every incoming URL event even when `extractExpoPathFromURL` yields nothing and the result is discarded. The old code only computed it inside the truthy branch.

Moving it into the branch removes the wasted walk and answers the `TODO(@ubax): check if this is performant` you left above it — which reads like it wasn't meant to ship:

```ts
return path !== undefined
  ? getStateFromPathRef.current(
      path,
      configRef.current,
      getRouteInfoFromState(ref.current?.getRootState()).segments
    )
  : undefined;
```

packages/expo-router/src/global-state/useNavigationTreeReducer.ts
```diff
@@ -47,11 +47,25 @@ type Options = {
...
+    return;
+  }
+
+  let focusedState: NavigationState | undefined = state;
+  while (focusedState) {
+    if (focusedState.stale) {
+      console.error('Detected stale state. This is likely a bug in Expo Router.');
+      return;
+    }
+    focusedState = focusedState.routes[focusedState.index]?.state as NavigationState | undefined;
```
jakub-agent, line 65: [💡 suggestion] This descent drops the fallback the old `syncStoreNavigationState` had:

```ts
focusedState = focusedState.routes?.[
  'index' in focusedState && typeof focusedState.index === 'number'
    ? focusedState.index
    : focusedState.routes.length - 1
]?.state;
```

The values being walked are `route.state`, which is `NavigationState | PartialState | undefined` despite the cast, and a `PartialState` need not carry `index`. For such a node `routes[undefined]` is `undefined`, so the loop exits one level early; the old code kept descending via `routes.length - 1`. A `PartialState` with `stale: undefined` is not caught by the `.stale` check above either, so a stale grandchild under it is now missed.

Dev-only diagnostic, so low stakes — but this warning exists to catch exactly that shape. Restoring the fallback keeps the old coverage.

packages/expo-router/src/ExpoRoot.tsx
```diff
@@ -122,8 +123,21 @@ function ContextNavigator({
     return undefined;
   }, []);
 
-  const storeValue = useStore(context, linking, serverUrl);
-  const { navigationRef, rootComponent, linking: linkingConfig, routeNode } = storeValue;
+  const { routerConfig, rootComponent } = useRouterConfig(context, linking, serverUrl);
+  const { linking: linkingConfig, routeNode } = routerConfig;
+
+  useClientLayoutEffect(() => {
+    return navigationRef.addListener('__unsafe_action__', (event) => {
```
jakub-agent, line 130: [💡 suggestion] Moving this off `onReady` changes which `createNavigationContainerRef` path the subscription takes, and the queued path has sharp edges.

On an async initial URL, `useThenable` is unresolved on the first render, so `NavigationContainerInner` returns the fallback before `<BaseNavigationContainer ref={refContainer}>` ever renders — `refContainer.current` is null, and `useImperativeHandle(ref, () => refContainer.current!)` (`fork/NavigationContainer.tsx:149`) leaves `navigationRef.current` null. This layout effect then runs with `current == null`, so `addListener` stores the callback in the ref's internal `listeners` map instead of attaching it. Two consequences:

- The returned unsubscribe only splices the queue; it can't detach the callback from the container it was later replayed onto, so the listener outlives `ExpoRoot` unmount.
- That `useImperativeHandle` has no dependency array, so every re-render of `NavigationContainerInner` reassigns `current` and replays the queue, pushing the same callback again. `useEventEmitter.emit` dedupes untargeted events via `lastIndexOf`, so `actionDispatched` isn't double-emitted — but the callback array grows and the dedupe is O(n²) per `__unsafe_action__` / `state` event.

The old `handleNavigationOnReady` avoided both by construction: it ran from `onReady` (so `current` was always set) and explicitly unsubscribed before re-adding. Gating on readiness would keep that property, e.g. add the listener from `onNavigationReady`, or key this effect on a ready flag so it only subscribes once `navigationRef.current` exists.

jakub-agent (approved):

PR 4/4-ish in the `@ubax/make-routing-queue-context` stack (base: `@ubax/make-routing-queue-context`, not `main`).

This is a good refactor. Deleting `store.ts` / `useStore.ts` / `storeContext.ts` / `sort-routes.ts` / `navigationEvents/navigation.ts` removes the render-phase global write in `NavigationContainer.tsx` (the `syncStoreNavigationState(initialState)` block) and the `owner`/`storeRef` ownership dance, and replaces them with a memoized `RouterConfigContext` plus reads from the live container (`getRootState()` / `RootNavigationStateContext`). Two incidental fixes worth calling out:

- Moving the splash RAF into `utils/splash.ts` with `useEffect(..., [])` fixes a latent bug: the old `useEffect` in `useStore` had **no** dependency array, so its cleanup cancelled the pending `requestAnimationFrame` after every re-render while `hasAttemptedToHideSplash` stayed `true` — the splash could be left up permanently.
- `useBrowserHistorySync` now uses one `rootState` for both `previousStateRef` and the `history.push/replace` payloads, instead of mixing `ref.current.getRootState()` with `store.state`.

Test migration looks faithful: the stale-state assertion moved from `store.test.ios.ts` into `useNavigationTreeReducer.test.ios.tsx`, and `useLinking.test.ios.tsx` gained a real case for group preservation when parsing an incoming deep link.

I ran the `expo-router` Jest suite against this commit locally: **4949 passed, 3 failed, 35 skipped**. The 6 failing suites are all artifacts of my local checkout, not this PR — 3 `native-tabs/__rsc_tests__` snapshots embed a relative asset path that depends on where the checkout sits on disk, and `TopTabs-exports.test.tsx` + `hmr-route-changes.test.ios.tsx` fail to load because `react-native-tab-view` doesn't resolve there. I confirmed `TopTabs-exports.test.tsx` fails identically on `origin/main`. 22 of the 23 test files this PR touches ran and passed; `hmr-route-changes.test.ios.tsx` is the one I could not execute, so treat CI as the source of truth for it.

Findings are all non-blocking. Two housekeeping items that don't fit an inline comment:

1. `packages/expo-router/AGENTS.md` still documents `global-state/router-store.tsx` in the directory tree ("Zustand store for router state") and under **State Management** ("**RouterStore** (`global-state/router-store.tsx`): The global store managing navigation state, and making it accessible imperatively via the `store` object"). This PR deletes that file and the `store` object. That file's own "Maintaining This Document" section asks for an update when the implementation changes.
2. The PR description is the unfilled template — no Why / How / Test Plan and no checklist. For a 53-file refactor that removes a documented internal API, a short Why/How would help reviewers of the stack.

packages/expo-router/src/fork/useLinking.native.ts
Ubax: Compute it afterwards. But keep the todo

Ubax: Leave `TODO(@ubax)` to check if this `ref.current?.getRootState()` can be replaced with the context read

packages/expo-router/src/fork/useLinking.ts
```diff
@@ -241,19 +241,16 @@ function useBrowserHistorySync({
         return;
       }
 
-      const parsedState = getStateFromPathRef.current(
-        path,
-        configRef.current,
-        getRouteInfoFromState(store?.state).segments
-      );
+      const segments = getRouteInfoFromState(navigation.getRootState()).segments;
```
Ubax, line 245: Leave TODO to check if `navigation.getRootState()` can be replaces with context read

packages/expo-router/src/fork/useLinking.ts
```diff
@@ -327,12 +324,10 @@ function useBrowserHistorySync({
 
     if (ref.current) {
       const rootState = ref.current.getRootState();
```
Ubax, line 328: Same here - TODO

packages/expo-router/src/fork/useLinking.ts
```diff
@@ -345,13 +340,11 @@ function useBrowserHistorySync({
 
       const previousState = previousStateRef.current;
       const rootState = navigation.getRootState();
```
Ubax, line 345: And here - TODO

packages/expo-router/src/global-state/navigationRef.ts
Ubax: Leave TODO, without reference to ticket. I'll add a ticket in a follow-up

packages/expo-router/src/global-state/router.ts
```diff
@@ -13,12 +13,12 @@ import { resolveHref } from '../link/href';
...
 import { getHistoryLength } from '../utils/stack';
 import { shouldLinkExternally } from '../utils/url';
+import { navigationRef } from './navigationRef';
 import type { RoutingIntent } from './routingQueue';
-import { store } from './store';
 import type { LinkToOptions, NavigationOptions } from './types';
 
 function assertIsReady() {
-  if (!store.navigationRef.isReady()) {
+  if (!navigationRef.isReady()) {
```
Ubax, line 22: TODO about wether this is still needed

packages/expo-router/src/global-state/router.ts
```diff
@@ -108,10 +108,10 @@ export function canGoBack(): boolean {
   // before mounting a navigator. This behavior exists due to React Navigation being dynamically
   // constructed at runtime. We can get rid of this in the future if we use
   // the static configuration internally.
-  if (!store.navigationRef.isReady()) {
+  if (!navigationRef.isReady()) {
```
Ubax, line 113: And here - TODO

packages/expo-router/src/global-state/router.ts
```diff
@@ -120,7 +120,10 @@ export function canDismiss(): boolean {
       'canDismiss imperative method is not supported. Pass the property to the DOM component instead.'
     );
   }
-  let state = store.state;
+  if (!navigationRef.isReady()) {
```
Ubax, line 126: And here - TODO

packages/expo-router/src/global-state/router.ts
Ubax: Leave as is

Ubax, line 130: And here - TODO

packages/expo-router/src/global-state/useNavigationTreeReducer.ts
```diff
@@ -47,11 +47,25 @@ type Options = {
...
 const warnedActions = new WeakSet<NavigationAction>();
 
+function warnIfStaleState(state: NavigationState) {
+  if (process.env.NODE_ENV !== 'development') {
+    return;
+  }
+
+  let focusedState: NavigationState | undefined = state;
+  while (focusedState) {
+    if (focusedState.stale) {
```
Ubax:
```suggestion
    if (focusedState.stale || focusedState.index === undefined) {
```

packages/expo-router/src/global-state/useStore.ts
Ubax, line 1: Please rename the file to `useStore.ts`. Add TODO comment to rename it to `useRouterConfig.ts`. I want to be able to see the diff better. Keep the hook name as `useRouterConfig`

packages/expo-router/src/link/preview/useNextScreenId.ts
```diff
@@ -1,22 +1,24 @@
...
 export function useNextScreenId(): [
   { nextScreenId: string | undefined; tabPath: TabPath[] },
   (href: Href) => void,
 ] {
   const router = useRouter();
   const routeInfo = useRouteInfo();
-  const store = use(StoreContext);
-  const navigationRef = store?.navigationRef;
+  const routerConfig = use(RouterConfigContext);
+  const navigation = use(NavigationContainerRefContext);
```
Ubax, line 21: Can I use this context in other places where the imperative `navigationRef` is used?

packages/expo-router/src/ExpoRoot.tsx
Ubax: Will be removed in a follow-up, so leave as is

packages/expo-router/CHANGELOG.md
Ubax: Do it

packages/expo-router/src/global-state/useNavigationTreeReducer.ts
Ubax: Will be addressed with the if

PR #49408, opened 2026-08-26
[router] fix onActionDispatched emitter and preventRemove

Ubax:
# Why

When refactoring router to use global state - https://github.com/expo/expo/pull/49297 - a regression was introduced - prevent remove and action listeners stopped working. This PR fixes that.

# How

1. When reducing in global state reducer return not only state but also events - `{ state, events }`. Events can be of different types (three right now - `action-dispatched`, `route-removed`, `remove-prevented`). 
2. After next render (in `useLayoutEffect`) `useNavigationTreeReportEvents` processes these events and emits the react-navigation ones. After processing it dispatches `REPORT_CONSUMED` action to reducer to remove the processed events (by id)
3. `usePreventRemove` is refactored to align better with current architecture. It uses context to pass a set of prevented ids to the reducer. This context stores id of the route with `usePreventRemove(true)` and all of its parents. Since all the synchronization between the context passed to reducer and `usePreventRemove` is executed in effect phase, the hook returns a function which disables prevention synchronously - for the next reducer render, rather then waiting one commit. 

# Test Plan

CI

# Checklist


packages/expo-router/src/react-navigation/core/usePreventRemove.tsx
```diff
@@ -37,34 +36,40 @@ import { useRoute } from './useRoute';
  * ```
  *
  * @param preventRemove Boolean indicating whether to prevent screen from being removed.
- * @param callback Function which is executed when screen was prevented from being removed.
+ * @param callback Optional function called when the screen was prevented from being removed.
```
jakub-agent, line 66: [🚨 **critical**] Both recovery paths this JSDoc documents look broken by the new design, and the PR's own test edits are the evidence.

The prevented-route list reaches the reducer through `config` (`useNavigationTreeReducer.ts`), captured at render time. Publishing it takes two commits: the screen's state change → `PreventRemovalProvider`'s layout effect → `RemovalPreventionProvider`'s `setPreventedRouteKeys` → re-render of `BaseNavigationContainerInner` with a new `config`. Anything dispatched in the same tick as the `preventRemove` change therefore reduces against the stale list.

So the `@example` above (`setHasUnsavedChanges(false); router.back()` in one handler) no longer works — which matches `removePrevented.test.web.tsx`, where `continues a blocked router back after disabling prevention` became `allows a new router back after disabling prevention` and was split into two `act()` calls. And "store it in the callback and dispatch it from an effect after `preventRemove` becomes `false`" matches the deleted `dispatches a blocked action from an effect after disabling prevention`.

The user-visible result is that the canonical unsaved-changes dialog traps the user: they press "Discard changes", nothing happens, and they have to press back a second time. This JSDoc is the generated API reference, so it will ship describing behavior the tests say is gone.

The same staleness applies in reverse and is not covered anywhere: `setDirty(true)` and `router.back()` in one handler discards the form despite `usePreventRemove(dirty)`.

Could prevention be read at reduce time instead of captured in `config` — e.g. keep the prevented set in a ref that `PreventRemovalProvider` writes in its layout effect and the reducer reads through a stable getter? That is what the old `isRoutePrevented` latest-callback gave you. If the two-tick behavior is intentional, the JSDoc, the `@example`, and the CHANGELOG all need to say so.

packages/expo-router/src/react-navigation/core/BaseNavigationContainer.tsx
```diff
@@ -139,6 +155,7 @@ function BaseNavigationContainerInner({
   });
 
   const dispatchSync = useLatestCallback((action: NavigationAction) => {
+    // TODO(@ubax): Throw if this is called from a `removePrevented` callback.
```
jakub-agent, line 135: [🚨 **critical**] This TODO replaces a guard that was doing real work, and its test went with it.

`useOnPreventRemove.tsx` kept `emittingRemovePreventedKeys` plus a dev warning that told the user exactly how to recover, and `removePrevented.test.ios.tsx` covered it (`blocks synchronous redispatch from removePrevented without re-emitting`). Both are deleted here with nothing in their place.

Without the guard, a callback that redispatches the blocked action loops: `dispatchSync` → `handleAction` → `reactDispatch` from inside the layout effect → reducer runs against the same (still prevented) list → a new `report` object → the `useClientLayoutEffect` above sees a changed `report` dep and emits `removePrevented` again → callback dispatches again. It terminates only when React throws `Maximum update depth exceeded`.

That is a plausible thing for a user to write, since it is what the previous warning implicitly described and what the old test exercised. Either restore a guard (a `WeakSet` of actions already reported as prevented would do) or land the throw this TODO describes before merging, rather than shipping the unguarded path.

packages/expo-router/src/global-state/useNavigationTreeReducer.ts
```diff
@@ -156,18 +178,34 @@ function navigationTreeReducer(
...
         ? completeNavigationState(result.nextState, config.routeNode)
         : result.nextState;
-      return nextState === state ? state : deepFreeze(nextState);
+      if (nextState === state) {
+        return current;
+      }
+
+      const removedRoutes = getRemovedRouteKeys(state, nextState);
+      const preventedRouteKeys = new Set(config.routesWithRemovalPrevented);
+      const preventedRoutes = removedRoutes.filter((routeKey) => preventedRouteKeys.has(routeKey));
```
jakub-agent: [🚨 **critical**] This veto applies to every `ACTION`, including framework-internal reconciliation, which I don't think is intended.

`ROUTE_NAMES_CHANGED` filters out routes whose names are no longer declared (`StackRouter.tsx:237`, `const routes = activeRoutes.filter((route) => routeNames.includes(route.name))`), so an undeclared route's key lands in `removedRoutes` and is vetoed if that route still holds a prevention entry.

`hmr-route-changes.test.ios.tsx` → `does not allow removal prevention to block route file changes` is exactly this: it renders a screen calling `usePreventRemove(true, cb)`, deletes the route file, and asserts `routeNames` becomes `['index']` and `cb` was not called. Whether it still passes depends on whether the screen's prevention is unregistered before the queued `ROUTE_NAMES_CHANGED` drains — the unregister happens in a layout-effect cleanup, the drain in `RoutingQueueDrainer`'s passive effect, and React flushes pending passive effects before the sync re-render that a layout-effect `setState` schedules. Please confirm that test locally; I could not run the suite.

Independent of that timing, the semantics look wrong: `usePreventRemove` is a guard on user-initiated navigation, not something that should be able to block the router reconciling its declared routes. The same applies to redirects, since `NAVIGATE_TO_HREF` recurses into this `ACTION` branch — protected routes are implemented as redirects, so a user signing out with a dirty form would have the redirect vetoed by their own `usePreventRemove`.

Consider restricting the check to user-originated actions, or explicitly exempting `ROUTE_NAMES_CHANGED` and redirect-driven navigation.

packages/expo-router/src/global-state/useNavigationTreeReducer.ts
```diff
@@ -156,18 +178,34 @@ function navigationTreeReducer(
...
+      const committedState = preventedRoutes.length > 0 ? state : deepFreeze(nextState);
+      const report: NavigationTreeReport = {
```
jakub-agent, line 257: [🏗️ **design**] A single `report` slot means only the last action in a React batch is ever reported.

`RoutingQueueDrainer.tsx` drains the whole queue in one effect (`for (const intent of routingQueue.drain(intents))`), so N queued `ACTION`s reduce in one batch. Each pass overwrites `report`, and React commits once, so the events for the first N-1 actions are never delivered. The new test `reduces consecutive actions against accumulated state with one committed update` already encodes this by asserting `reports.at(-1)?.action` is the second action.

Concretely, `router.push('/a'); router.push('/b')` in one handler emits one `actionDispatched` (for `PUSH /b`) where the pre-migration `__unsafe_action__` emitted two, and any route removed by the first action never receives `removed`. This is more than cosmetic: `expo-observe`'s router integration (`packages/expo-observe/src/integrations/expo-router/init.ts`) pushes an entry into `storage.pendingActions` per `actionDispatched` and matches it against `pageFocused` to compute navigation TTR, so a dropped event means a lost or mis-attributed metric. A queued `ROUTE_NAMES_CHANGED` landing last in a drain would clobber a user navigation's event the same way.

Accumulating reports (`reports: NavigationTreeReport[]`) and draining the list in the consumer would preserve one event per action.

packages/expo-router/src/react-navigation/core/useDescriptors.tsx
```diff
@@ -235,6 +213,14 @@ export function useDescriptors<
         routeState={routeState}
         options={customOptions}
         clearOptions={clearOptions}
+        isPreloaded={isRoutePreloadedInStack(getState(), route)}
```
jakub-agent: [🏗️ **design**] `getState()` reads a ref during render, so `isPreloaded` is one commit stale.

`getState()` resolves to `getStateForKey(stateKeyRef.current)` → `stateRef.current`, and `useNavigationTreeReducer` only assigns `stateRef.current` in `useInsertionEffect`. During the render that commits a new state, the ref still holds the previous one.

On the render that promotes a preloaded route to active, `isRoutePreloadedInStack` therefore still returns `true`, so `SceneView` passes `propagateToParent={false}` — and nothing schedules another render to correct it. A screen calling `usePreventRemove(true)` immediately after promotion never marks its ancestor route as prevented, so the parent native stack keeps `headerBackButtonMenuEnabled` on (`NativeStackView.native.tsx`) and lets the user long-press back past a screen whose removal will then be vetoed.

It is also a concurrent-render tear, which `packages/expo-router/AGENTS.md` explicitly asks this code to avoid. The correct value is already in scope: the reduce below has the route index `i`, so `state.type === 'stack' && i > state.index` gives the rendered answer without touching a ref.

packages/expo-router/src/global-state/useNavigationTreeReducer.ts
```diff
@@ -244,21 +284,23 @@ export function useNavigationTreeReducer({
...
 
   React.useInsertionEffect(() => {
-    stateRef.current = state;
-    onStateChangeInsertion?.(state);
-  }, [onStateChangeInsertion, state]);
+    stateRef.current = result.state;
+    onStateChangeInsertion?.(result.state);
+  }, [onStateChangeInsertion, result.state]);
+
+  React.useEffect(() => setVersion((value) => value + 1), [result]);
```
jakub-agent: [🏗️ **design**] This `version` latch costs an extra render and a dependency cascade on every navigation.

The effect fires after each committed action, which rebuilds `config` → `processAction` → `processIntent`, and `processIntent` is a dependency of `RoutingQueueDrainer`'s drain effect, so that re-runs too. So every navigation pays render + effect + render, and the whole memo chain below is invalidated for what is only a "consume this report once" signal.

A dedicated reducer operation (`{ type: 'REPORT_CONSUMED' }`) dispatched from the consuming layout effect would clear the report without a second state atom, keeping `config` stable across actions and dropping `version` from both `ReducerConfig` and `NavigationTreeReport`. That also removes the need for the `result.report.version !== config.version` reconciliation at the top of the reducer.

packages/expo-router/src/react-navigation/core/useEventEmitter.tsx
```diff
@@ -33,8 +36,20 @@ export function useEventEmitter<T extends Record<string, any>>(
...
+        return;
+      }
+
+      const remove = () => {
+        const currentIndex = callbacks.indexOf(callback);
+        if (currentIndex >= 0) {
+          callbacks.splice(currentIndex, 1);
+        }
+      };
+      if (type === 'removed' && isTargetMountedRef.current?.(target) === false) {
```
jakub-agent: [🏗️ **design**] A generic event emitter now special-cases one navigation event name and defers unsubscription across a microtask.

Two issues. First, `type === 'removed'` hardcodes navigation semantics into a hook that otherwise knows nothing about the event map; an explicit option (`deferUnsubscribeWhileUnmounted?: (type, target) => boolean`) or handling it at the emit site would keep the layering.

Second, this keeps an unmounted screen's listener alive across a microtask and depends on `BaseNavigationContainer`'s `useClientLayoutEffect` running before microtasks drain. `removalPrevention.tsx`'s `unregisterRouteEmitter` makes the same assumption with its own `queueMicrotask`, so there are now two independent ordering assumptions holding up `removed` delivery, both undocumented at the emit site.

Neither deferral path is tested. `stops emitting removed events immediately after unsubscribe` calls `useEventEmitter` with no `isTargetMounted`, so `isTargetMountedRef.current?.(target) === false` is `undefined === false` and it only exercises the immediate branch — despite what the name suggests. A test that passes `isTargetMounted: () => false` and asserts the listener still fires for a `removed` emitted in the same task (and is gone after `await Promise.resolve()`) would cover the branch this PR actually depends on.

packages/expo-router/src/react-navigation/core/useDescriptors.tsx
```diff
@@ -235,6 +213,14 @@ export function useDescriptors<
...
+        emitRemovalEvent={(type, action) => {
```
jakub-agent: [💡 suggestion] The branch is redundant and the closure is unstable.

Both arms emit the same shape with the same `target` and `data`, differing only in the literal that already equals `type`, so this collapses to:

```tsx
emitter.emit({ type, target: route.key, data: { action } })
```

More importantly it is a new function identity every render, and it is a dependency of `PreventRemovalProvider`'s registration effect (`removalPrevention.tsx`), so every render of this navigator unregisters and re-registers the emitter for every route and queues a microtask per route via `unregisterRouteEmitter`. Wrapping it in a `useCallback` keyed on `[emitter, route.key]` — or hoisting a small stable helper, since `emitter.emit` is already stable — removes the churn.

packages/expo-router/src/react-navigation/core/usePreventRemove.tsx
```diff
@@ -37,34 +36,40 @@ import { useRoute } from './useRoute';
...
       callback({ data: event.data });
     }
   });
 
   React.useEffect(
-    () => navigation.addListener('removePrevented', removePreventedListener),
-    [navigation, removePreventedListener]
+    () =>
+      callback ? navigation.addListener('removePrevented', removePreventedListener) : undefined,
+    [callback, navigation, removePreventedListener]
```
jakub-agent: [💡 suggestion] Adding `callback` to the deps makes this re-subscribe on every render, since callers almost always pass an inline arrow (`usePreventRemove(dirty, () => setShowConfirm(true))`). Before this change the deps were stable and the listener was registered once.

Only the presence of a callback matters here, because `removePreventedListener` is already a latest-callback that reads `callback` at call time. Gating on a boolean keeps the effect stable:

```tsx
const hasCallback = callback !== undefined;
React.useEffect(
  () => (hasCallback ? navigation.addListener('removePrevented', removePreventedListener) : undefined),
  [hasCallback, navigation, removePreventedListener]
);
```

packages/expo-router/src/react-navigation/core/useNavigationHelpers.tsx
```diff
@@ -37,21 +37,16 @@ export function useNavigationHelpers<
...
-        type: 'NAVIGATOR_ACTION',
-        payload: {
-          action,
-          // The queued action was already constrained to this navigator's action type.
-          dispatchSync: (queuedAction) => dispatchSync(queuedAction as Action),
-        },
+      routingQueue.add({
+        type: 'ACTION',
+        payload: { action, originKey: getState().key },
       });
```
jakub-agent, line 54: [💡 suggestion] Two things follow from this switch.

First, nothing in `src/` produces a `NAVIGATOR_ACTION` intent any more — the only remaining references are the type in `routingQueue.ts`, the branch in `RoutingQueueDrainer.tsx`, the `throw` in `useNavigationTreeReducer.ts`, and `RoutingQueueDrainer.test.ios.tsx`, which constructs one by hand. Since the point of routing helper dispatches through the reducer is to get them into the report path, the old intent shape looks dead rather than dormant. Worth deleting the type, the drainer branch, and the reducer's `throw` here while the reasoning is fresh, unless something outside `expo-router` still queues it.

Second, this is what changes the sibling-container behavior in `BaseNavigationContainer.test.ios.tsx` → `logs an error when a sibling root container drains the shared routing queue`. The queued intent now carries `originKey` from the sibling's tree; the first container drains it, `resolveOrigin` cannot find that key in its own tree and returns `undefined`, so the reducer calls `warnUnhandledAction` — a second `console.error` on top of the multiple-drainers message, and the navigate is discarded entirely (hence `onStateChange` never firing and the unchanged root state). That test dropped `expect(error).toHaveBeenCalledTimes(1)`, which hides the new error rather than documenting it. Please assert both messages instead, so a future third error can't slip through silently.

packages/expo-router/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### 🛠 Breaking changes
 
+- Remove `beforeRemove`, `__unsafe_action__`, `PreventRemoveContext`, and `usePreventRemoveContext` from `expo-router/react-navigation`.
```
jakub-agent: [💡 suggestion] This entry is missing the PR link and author that every neighbouring entry carries: `([#49408](https://github.com/expo/expo/pull/49408) by [@Ubax](https://github.com/Ubax))`.

A few things also seem worth their own entries:

- A `🐛 Bug fixes` entry for what the PR title actually fixes — `unstable_navigationEvents.actionDispatched` no longer firing, and `usePreventRemove` / `removePrevented` not blocking removal.
- The new `removed` event on `EventMapCore`, which is a public addition alongside the `beforeRemove` removal.
- `usePreventRemove`'s `callback` becoming optional.
- `removePrevented` now firing for every prevented route in the removed chain instead of only the deepest one — the assertion changes in `usePreventRemove.test.ios.tsx` (`onPreventRemove.baz` / `onPreventRemove.bar` going from 1 to 2 and 3 calls) are a behavior change users can observe.

The PR body is also still the unfilled template, so there is no Why / How / Test Plan for a ~800-line change to prevention and event delivery. I've left it alone — that's yours to fill in.

packages/expo-router/src/react-navigation/core/BaseNavigationContainer.tsx
```diff
@@ -100,36 +105,47 @@ function BaseNavigationContainerInner({
...
+
+    for (const routeKey of report.preventedRoutes) {
+      emitterRegistry.emitRemovalEvent(routeKey, 'removePrevented', report.action);
+    }
+
+    if (report.preventedRoutes.length === 0) {
+      for (const routeKey of report.removedRoutes) {
+        emitterRegistry.emitRemovalEvent(routeKey, 'removed', report.action);
+      }
+      unstable_navigationEvents.emit('actionDispatched', {
```
jakub-agent: [💡 suggestion] Worth confirming the timing shift is intended. The `__unsafe_action__` doc comment this replaces said the event "will be emitted before state changes have been applied"; this emits from a layout effect after the new state has been reduced and committed.

`expo-observe` stamps `dispatchTime: performance.now()` when it receives `actionDispatched` and diffs it against the later `pageFocused` to produce navigation TTR, so moving the emit later shortens every measured TTR by the reduce + render cost of the new state. Not necessarily wrong — post-commit is arguably the more meaningful anchor and it is what makes `report.state` available — but it changes the meaning of a shipped metric and isn't mentioned in the CHANGELOG.

packages/expo-router/src/global-state/removalPrevention.tsx
```diff
@@ -0,0 +1,165 @@
...
+  return (
+    <ScreenRemovalPreventionSetterContext.Provider value={setPrevented}>
+      <ScreenRemovalPreventedContext.Provider value={isPrevented}>
+        {children}
+      </ScreenRemovalPreventedContext.Provider>
+    </ScreenRemovalPreventionSetterContext.Provider>
+  );
+}
+
+export function useIsRemovalPrevented() {
```
jakub-agent: [🐧 nit] `useIsRemovalPrevented` has no callers — `ScreenRemovalPreventedContext` is only read by this hook and by `removalPrevention.test.ios.tsx`. Worth dropping until something needs it.

jakub-agent (changes requested):

This restores the two features that the global-reducer migration (#49297) left stubbed: `unstable_navigationEvents.actionDispatched` and removal prevention. The direction is right — moving prevention out of per-navigator keyed listeners into one global prevented-route list, and deriving events from a reducer `report`, removes a lot of the old re-entrancy machinery and un-skips ~20 tests. `useInvalidPreventRemoveError` and the `NativeStackView` `headerBackButtonMenuEnabled` logic both get simpler and more correct along the way.

The concern is that prevention state is now read from a value captured in the reducer `config` at render time, while it is published two commits later through layout effects. That makes prevention correct for the steady state but wrong across a single tick, in both directions:

- Setting `preventRemove` to `false` and dispatching in the same handler is still blocked. The PR's own tests confirm this: `continues a blocked router back after disabling prevention` was renamed to `allows a new router back after disabling prevention` and split into two `act()` calls, and `dispatches a blocked action from an effect after disabling prevention` was deleted. Both are exactly the recovery paths that `usePreventRemove`'s JSDoc and `@example` still document.
- The old re-entrancy guard in `useOnPreventRemove` and its covering test were removed and replaced with a TODO, so dispatching from a `removePrevented` callback can now re-emit without bound.

Separately, `ROUTE_NAMES_CHANGED` now flows through the same veto, which is a user-facing hook blocking a framework reconciliation action, and only the last action in a drained batch produces a report, so `actionDispatched` / `removed` are dropped for the rest of the batch — `expo-observe` consumes `actionDispatched` for navigation TTR.

I could not run the expo-router jest suite (the local install is stale), so findings below are from reading; severity notes where a claim is timing-dependent. Worth confirming `hmr-route-changes.test.ios.tsx` and the full suite locally.

Also: the PR body is still the unfilled template, and the CHANGELOG entry has no PR link or author attribution and does not mention the fixes or the new `removed` event.

packages/expo-router/src/global-state/useNavigationTreeReducer.ts
Ubax: Add a stable set of action which should skip the prevention, for example `ROUTE_NAMES_CHANGED` or unmonuting the navigator

packages/expo-router/src/global-state/useNavigationTreeReducer.ts
Ubax: Change the logic to work as described above

packages/expo-router/src/global-state/useNavigationTreeReducer.ts
```diff
@@ -45,11 +47,25 @@ type Options = {
   initialState: InitialState | undefined;
   routeNode?: RouteNode;
   registry: RouterRegistry;
+  routesWithRemovalPrevented?: readonly string[];
   linking?: ExpoLinkingOptions;
   redirects?: StoreRedirects[];
   onStateChangeInsertion?: (state: NavigationState) => void;
 };
 
+export type NavigationTreeReport = {
```
Ubax, line 58: The report should allow multiple actions to be inside. It should have a bit different structure:

```tsx
export type NavigationTreeReport = {
  version: number;
  events: NavigationTreeReportEvent[]
}
```

Then there can be multiple events:
- `NavigationTreeReportRemovedRoutes = { type: 'removed-routes', routeKeys: [...] }`
- `NavigationTreeReportPreventedRoutes = { type: 'prevented-routes', routeKeys: [...], action }`
- `NavigationTreeReportActionDispatched = { type: 'action-dispatched', action }`

Then in the reducer:
- if previous report version is not current config version then overwrite the whole report
- if versions align then add the event to the `events` array

packages/expo-router/src/global-state/useNavigationTreeReducer.ts
```diff
@@ -280,6 +323,37 @@ export function useNavigationTreeReducer({
...
+  for (let index = routes.length - 1; index >= 0; index--) {
+    const route = routes[index]!;
+    if (route.state?.stale === false) {
+      visitRoutes(route.state, excludePreloaded, visit);
+    }
+    visit(route.key);
+  }
+}
+
 function validateInitialState(state: InitialState | undefined): void {
```
Ubax:
```suggestion
function validateInitialState(state: InitialState | undefined): state is ...... {
```

packages/expo-router/src/global-state/useNavigationTreeReducer.ts
```diff
@@ -199,29 +237,31 @@ export function useNavigationTreeReducer({
...
+    (value): NavigationTreeResult => {
       validateInitialState(value == null ? undefined : value);
       if (value == null) {
         throw new Error(
           'The navigation container is missing its initial state. Expo Router always seeds a complete initial state before rendering the navigation container, so this is most likely a bug in expo-router. Please report it at https://github.com/expo/expo/issues.'
         );
       }
       // Validation above proves the recursively partial public type is complete.
-      return deepFreeze(value as NavigationState);
+      return { state: deepFreeze(value as NavigationState), report: undefined };
```
Ubax: Add TODO to check if `deepFreeze` is really needed here

packages/expo-router/src/global-state/useNavigationTreeReducer.ts
```diff
@@ -280,6 +323,37 @@ export function useNavigationTreeReducer({
...
+  return removedRoutes;
+}
+
+function visitRoutes(
+  state: NavigationState,
+  excludePreloaded: boolean,
+  visit: (routeKey: string) => void
+) {
+  const routes =
+    excludePreloaded && state.type === 'stack'
```
Ubax, line 421: Add TODO, to have a more universal way of excluding preloaded routes.

packages/expo-router/src/global-state/removalPrevention.tsx
```diff
@@ -0,0 +1,165 @@
...
+/** Indicates whether the current route or an active descendant prevents removal. */
+export const ScreenRemovalPreventedContext = createContext(false);
+
+/** Registers independent prevention requests with the nearest route provider. */
+export const ScreenRemovalPreventionSetterContext = createContext<
+  ((id: string, isPrevented: boolean) => void) | undefined
+>(undefined);
+
+/** Owns the global prevented-route list and route removal-event registry. */
+export function RemovalPreventionProvider({ children }: PropsWithChildren) {
```
Ubax: Spit this into three providers:
- emitter provider
- setter + preventedRouteKeys
- RemovalPreventionProvider - composing the two

packages/expo-router/src/global-state/removalPrevention.tsx
```diff
@@ -0,0 +1,165 @@
...
+        <GlobalRoutesWithRemovalPreventedContext.Provider value={preventedRouteKeys}>
+          {children}
+        </GlobalRoutesWithRemovalPreventedContext.Provider>
+      </GlobalRemovalEventEmitterRegistryContext.Provider>
+    </GlobalRouteRemovalPreventionSetterContext.Provider>
+  );
+}
+
+/** Aggregates prevention requests for one route and propagates them through the navigation tree. */
+export function PreventRemovalProvider({
```
Ubax: Extract internal hooks from this provider:
- registering emitter for screen
- propagating to parent
- handling current prevented state (set and is)

Then compose them in this provider

packages/expo-router/src/global-state/removalPrevention.tsx
```diff
@@ -0,0 +1,165 @@
...
+  useClientLayoutEffect(() => {
+    if (!isPrevented || !propagateToParent) {
+      return;
+    }
+    parentSetPrevented?.(parentId, true);
+    return () => parentSetPrevented?.(parentId, false);
+  }, [isPrevented, parentId, parentSetPrevented, propagateToParent]);
+
+  useClientLayoutEffect(() => {
+    if (!isPrevented) {
```
Ubax: Is there any time window when screen and parent isPrevented disagree - when child is true and parent still false?

packages/expo-router/src/global-state/removalPrevention.tsx
Ubax: Drop the hook

packages/expo-router/src/react-navigation/core/__tests__/actionBubbling.test.ios.tsx
```diff
@@ -713,6 +712,10 @@ test.skip("prevents removing a screen with 'removePrevented' event", () => {
     setPreventRemove(false);
   });
 
+  expect(onStateChange).toHaveBeenCalledTimes(2);
```
Ubax, line 715: Why is this added?

packages/expo-router/src/react-navigation/core/__tests__/actionBubbling.test.ios.tsx
```diff
@@ -1011,6 +1011,8 @@ test.skip("prevents removing by multiple screens with 'removePrevented' event",
 
   expect(onStateChange).toHaveBeenCalledTimes(1);
```
Ubax, line 1012: Will this be called if no action was executed?

packages/expo-router/src/react-navigation/core/__tests__/actionBubbling.test.ios.tsx
```diff
@@ -1028,13 +1031,12 @@ test.skip("prevents removing by multiple screens with 'removePrevented' event",
   });
 
   expect(onStateChange).toHaveBeenCalledTimes(1);
-  expect(onBeforeRemove.bar).toHaveBeenCalledTimes(1);
+  expect(onBeforeRemove.bar).toHaveBeenCalledTimes(3);
```
Ubax, line 1034: Why is it called 3 times now?

packages/expo-router/src/react-navigation/core/__tests__/actionBubbling.test.ios.tsx
```diff
@@ -1019,7 +1021,8 @@ test.skip("prevents removing by multiple screens with 'removePrevented' event",
   });
 
   expect(onStateChange).toHaveBeenCalledTimes(1);
-  expect(onBeforeRemove.baz).toHaveBeenCalledTimes(1);
+  expect(onBeforeRemove.baz).toHaveBeenCalledTimes(2);
+  expect(onBeforeRemove.bar).toHaveBeenCalledTimes(2);
```
Ubax, line 1025: Why is it called 2 times now?

packages/expo-router/src/react-navigation/core/__tests__/removePrevented.test.ios.tsx
```diff
@@ -66,108 +57,11 @@ test.skip('blocks removal with the hook and emits removePrevented', () => {
...
-    '`beforeRemove` is a notification-only event and cannot prevent screen removal. Use `usePreventRemove` with the `removePrevented` event instead.'
-  );
-
-  expect(ref.current?.getRootState().routes.map((route) => route.name)).toEqual(['foo', 'bar']);
-  expect(beforeRemove).toHaveBeenCalledTimes(1);
-  expect(beforeRemove.mock.calls[0][0].defaultPrevented).toBe(false);
-});
-
-// TODO(@ubax): Restore removePrevented redispatch after reducer dispatch supports it. https://linear.app/expo/issue/ENG-26123
-test.skip('blocks synchronous redispatch from removePrevented without re-emitting', () => {
```
Ubax, line 82: Add this test back as skipped

packages/expo-router/src/react-navigation/core/__tests__/removePrevented.test.ios.tsx
```diff
@@ -66,108 +57,11 @@ test.skip('blocks removal with the hook and emits removePrevented', () => {
...
-    expect(warn).toHaveBeenCalledWith(
-      "The action 'GO_BACK' was dispatched from inside a `usePreventRemove` callback and was prevented again. The `removePrevented` event was not re-emitted to avoid an infinite loop. There is no way to dispatch directly from the callback; set `preventRemove` to `false` first, then retry (for example, call `router.back()` from the handler or dispatch the captured action from an effect)."
-    );
-  } finally {
-    warn.mockRestore();
-  }
-});
-
-// TODO(@ubax): Restore nested beforeRemove events after reducer dispatch supports them. https://linear.app/expo/issue/ENG-26123
-test.skip('emits beforeRemove in a nested navigator when its parent route is removed', () => {
```
Ubax, line 126: Why is this test removed?

packages/expo-router/src/react-navigation/core/__tests__/removePrevented.test.web.tsx
```diff
@@ -18,16 +16,14 @@ global.ResizeObserver = class {
...
+test('allows a new router back after disabling prevention', () => {
   let discard: () => void;
   const onPreventRemove = jest.fn();
   const Form = () => {
     const [dirty, setDirty] = React.useState(true);
     usePreventRemove(dirty, onPreventRemove);
     discard = () => {
       setDirty(false);
-      router.back();
     };
```
Ubax, line 30: router.back should still work here

packages/expo-router/src/react-navigation/core/__tests__/removePrevented.test.web.tsx
```diff
@@ -49,20 +45,22 @@ test.skip('continues a blocked router back after disabling prevention', () => {
...
-// TODO(@ubax): Restore nested remove prevention after reducer dispatch supports it. https://linear.app/expo/issue/ENG-26123
-test.skip('continues a blocked parent back after disabling nested prevention', () => {
+test('allows a new parent back after disabling nested prevention', () => {
   let discard: () => void;
   const onPreventRemove = jest.fn();
   const Form = () => {
     const [dirty, setDirty] = React.useState(true);
     usePreventRemove(dirty, onPreventRemove);
     discard = () => {
       setDirty(false);
```
Ubax, line 62: Same here

packages/expo-router/src/react-navigation/core/__tests__/useNavigationCache.test.ios.tsx
```diff
@@ -353,13 +353,14 @@ test('ignores dispatches from a preloaded stack screen until it is promoted', ()
 
   expect(enqueue).toHaveBeenCalledTimes(1);
   expect(enqueue).toHaveBeenCalledWith({
-    type: 'NAVIGATOR_ACTION',
-    payload: expect.objectContaining({
+    type: 'ACTION',
```
Ubax, line 348: Why is this changed? Not saying it's incorrect, just asking

packages/expo-router/src/global-state/useNavigationTreeReducer.ts
Ubax: Also leave a TODO to add more dev only diagnostics to the events, so that we can incorporate them into dev-tools

packages/expo-router/src/react-navigation/core/BaseNavigationContainer.tsx
Ubax: Instead of the comment leave a TODO to add an event when action is enqued

packages/expo-router/src/react-navigation/core/BaseNavigationContainer.tsx
```diff
@@ -100,36 +105,47 @@ function BaseNavigationContainerInner({
...
       linking: store?.linking,
       redirects: store?.redirects,
       onStateChangeInsertion: UNSTABLE_onStateChangeInsertion,
     });
+  useClientLayoutEffect(() => {
+    if (report === undefined) {
+      return;
+    }
+
+    for (const routeKey of report.preventedRoutes) {
```
Ubax: Now when there are multiple events we can do:
```
for(const event of report.events){
switch(...){
...
}
```

packages/expo-router/src/react-navigation/core/BaseNavigationContainer.tsx
```diff
@@ -100,36 +105,47 @@ function BaseNavigationContainerInner({
...
     useNavigationTreeReducer({
       initialState,
       routeNode: UNSTABLE_routeNode,
       registry,
+      routesWithRemovalPrevented,
       linking: store?.linking,
       redirects: store?.redirects,
       onStateChangeInsertion: UNSTABLE_onStateChangeInsertion,
     });
+  useClientLayoutEffect(() => {
```
Ubax: We should extract this into a standalone hook for emitting state events

packages/expo-router/src/react-navigation/core/BaseNavigationContainer.tsx
Ubax: Add the tests back, but keep them skipped. We should disallow this from happening later on

packages/expo-router/src/react-navigation/core/useDescriptors.tsx
Ubax: Why does it call `getState` and not use `NavigationStateContext`?

packages/expo-router/src/react-navigation/core/useDescriptors.tsx
Ubax: Apply both - type and useCallback

packages/expo-router/src/react-navigation/core/useEventEmitter.tsx
Ubax: Change this type to `Set` so it's easier to do `has`, `remove` and `add`

packages/expo-router/src/react-navigation/core/useEventEmitter.tsx
Ubax: Remove `queryMicrotask` from here. This should remain generic

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
```diff
@@ -391,66 +382,69 @@ export function useNavigationBuilder<
     }
     return currentState as State;
   });
-  const emitter = useEventEmitter<EventMapCore<State>>((e) => {
-    const routeNames = [];
+  const emitter = useEventEmitter<EventMapCore<State>>(
+    (e) => {
```
Ubax: Why is there this formatting change?

packages/expo-router/src/react-navigation/core/useEventEmitter.tsx
```diff
@@ -13,12 +13,15 @@ type Listeners = ((e: any) => void)[];
  * Hook to manage the event system used by the navigator to notify screens of various events.
  */
 export function useEventEmitter<T extends Record<string, any>>(
-  listen?: (e: any) => void
+  listen?: (e: any) => void,
+  isTargetMounted?: (target: string) => boolean
```
Ubax: This is not the responsibility of this hook

packages/expo-router/src/react-navigation/core/useNavigationHelpers.tsx
Ubax: Delete if safe

packages/expo-router/src/react-navigation/core/usePreventRemove.tsx
```diff
@@ -37,34 +36,40 @@ import { useRoute } from './useRoute';
...
   const navigation = useNavigation();
-  const { key: routeKey } = useRoute();
-  const { setPreventRemove } = usePreventRemoveContext();
+  const setPreventRemove = React.use(ScreenRemovalPreventionSetterContext);
 
-  React.useEffect(() => {
-    setPreventRemove(id, routeKey, preventRemove);
+  if (setPreventRemove === undefined) {
+    throw new Error(
+      "Couldn't find the prevent remove context. Is your component inside NavigationContent?"
```
Ubax:
```suggestion
      "Couldn't find the prevent remove context. Is your component inside Screen or Layout?"
```

packages/expo-router/src/react-navigation/core/usePreventRemove.tsx
Ubax: Revert this effect to previous version. The check is already in `removePreventedListener`

packages/expo-router/src/react-navigation/native-stack/utils/useInvalidPreventRemoveError.tsx
```diff
@@ -1,12 +1,12 @@
 'use client';
 import * as React from 'react';
 
-import { usePreventRemoveContext } from '../../native';
+import { useRoutesWithRemovalPrevented } from '../../../global-state/removalPrevention';
 import type { NativeStackDescriptorMap } from '../types';
 
 export function useInvalidPreventRemoveError(descriptors: NativeStackDescriptorMap) {
```
Ubax, line 10: Leave a TODO comment to remove this later

packages/expo-router/CHANGELOG.md
Ubax: Only add PR link. No more entries

packages/expo-router/src/react-navigation/core/types.tsx
```diff
@@ -120,8 +120,24 @@ export type EventMapCore<State extends NavigationState> = {
   focus: { data: undefined };
   blur: { data: undefined };
   state: { data: { state: State } };
-  beforeRemove: { data: { action: NavigationAction } };
   removePrevented: { data: { action: NavigationAction } };
+  /**
+   * Emitted after the route is removed and its component unmounts. The listener cannot rely on
+   * component state or update the unmounted component. Since effect cleanup runs before this event,
+   * it must defer unsubscription until the next microtask.
```
jakub-agent, line 127: [🚨 **critical**] Taking the `queueMicrotask` out of `useEventEmitter` was the right call, but it didn't go away — it became a requirement on every consumer of this event, and the failure mode is silent.

`removed` is emitted from `useNavigationTreeReportEvents`' layout effect, which runs after the commit that unmounted the screen. So the effect cleanup has already removed the listener, and the ordinary idiom every other event in this map uses:

```tsx
React.useEffect(() => navigation.addListener('removed', cb), [navigation]);
```

compiles, type-checks, reads correctly, and never fires. Nothing warns. A user only discovers it by noticing their analytics or cleanup never runs. `focus`, `blur`, `state`, and `removePrevented` all work with the plain form, so `removed` is the one event with a different contract, documented only in this JSDoc.

I'd rather the API not require it. A couple of directions:

- Wrap it: export a `useRouteRemoved(callback)` hook (or accept an option on `addListener`) so the microtask stays an implementation detail. Users get one obvious way to subscribe and can't get it wrong.
- Or emit earlier, before the mutation phase tears the listener down, so a normal subscription is still live when the event fires.

If it does have to stay a consumer requirement, please also surface it where people will actually hit it — the `expo-router` docs for navigation events, not just this type declaration — and consider a dev-only warning when `removed` is emitted for a target that has no listeners but had one in the previous commit.

packages/expo-router/src/react-navigation/core/usePreventRemove.tsx
```diff
@@ -1,70 +1,78 @@
...
   React.useEffect(
-    () => navigation.addListener('removePrevented', removePreventedListener),
+    () =>
+      callback ? navigation.addListener('removePrevented', removePreventedListener) : undefined,
     [navigation, removePreventedListener]
   );
+  // TODO(@ubax): use standard useCallback if possible
+  // TODO(@ubax): add repeat function which will call this and repeat the action
+  // TODO(@ubax): add sync between disable and wether the prop did change as well
+  return useLatestCallback(() => setPreventRemove(id, false));
```
jakub-agent: [🚨 **critical**] `disablePrevention()` is a one-way door: once called, prevention never comes back while the component stays mounted with `preventRemove === true`.

`setPreventRemove(id, false)` removes `id` from the route's set in the global map. The layout effect above that would re-add it has deps `[id, preventRemove, setPreventRemove]` — `id` is a stable `useId`, `setPreventRemove` is a stable `useCallback`, and `preventRemove` is still `true`. No dep changed, so the effect does not re-run, and the id is never restored. The screen is silently unguarded until `preventRemove` toggles `false` → `true`.

That's the exact failure this hook exists to prevent, and the API shape invites it: a function named `disablePrevention` reads as "let this one navigation through", not "stop protecting this screen for good". Someone writing a confirm dialog whose state lives outside the screen — or who just calls `disablePrevention(); router.back()` without a local `dirty` flag — loses protection permanently.

Both new tests (`allows router back after disabling prevention`, `allows an action dispatched while disabling prevention`) also call `setDirty(false)` / `setPreventRemove(false)`, so they pass without covering this. A test that calls only `disablePrevention()` and then dispatches a second navigation would pin it down.

Your TODO on line 76 ("add sync between disable and wether the prop did change") is this. Two shapes that would close it:

- Re-assert on every render instead of only on dep change, so a still-`true` `preventRemove` re-arms itself on the commit after the disable.
- Or scope the disable to a single dispatch — have it mark "allow the next action through" rather than clearing the registration — which also matches the `repeat` function in your other TODO.

packages/expo-router/src/react-navigation/core/usePreventRemove.tsx
```diff
@@ -1,70 +1,78 @@
...
-    if (preventRemove) {
+    if (preventRemove && callback) {
       callback({ data: event.data });
     }
   });
 
   React.useEffect(
-    () => navigation.addListener('removePrevented', removePreventedListener),
+    () =>
+      callback ? navigation.addListener('removePrevented', removePreventedListener) : undefined,
```
jakub-agent: [🏗️ **design**] Reverting the deps to `[navigation, removePreventedListener]` left `callback` read in the effect body but out of the dep array.

Because `removePreventedListener` is a latest-callback and `navigation` is stable, the effect runs once. So a caller that passes a conditional callback — `usePreventRemove(true, isReady ? onBlocked : undefined)` — never attaches the listener if `isReady` starts `false`, and never detaches if it starts `true`. It also looks like it would trip `react-hooks/exhaustive-deps` in `pnpm lint`.

Since you noted the check already lives in `removePreventedListener` (which no-ops when `callback` is undefined), the ternary isn't buying anything — dropping it resolves both:

```tsx
React.useEffect(
  () => navigation.addListener('removePrevented', removePreventedListener),
  [navigation, removePreventedListener]
);
```

packages/expo-router/src/global-state/useNavigationTreeReducer.ts
```diff
@@ -256,15 +322,17 @@ export function useNavigationTreeReducer({
...
+    processAction({ type: 'NAVIGATOR_CHANGED', stateKey, routerType });
   });
 
   React.useInsertionEffect(() => {
-    warnIfStaleState(state);
-  }, [state]);
+    warnIfStaleState(result.state);
+  }, [result.state]);
+
+  React.useEffect(() => setVersion((value) => value + 1), [result]);
```
jakub-agent: [🏗️ **design**] Now that `useNavigationTreeReportEvents` tracks its own consumed-event count, this is a second, independent one-shot mechanism for the same job.

`consumedReport` already knows which events it emitted, so `version` is left doing only one thing: truncating the accumulated `events` array (via the `result.report.version !== config.version` reset at the top of the reducer). It pays for that with a full extra render after every committed action, which rebuilds `config` → `processAction` → `processIntent` and re-runs `RoutingQueueDrainer`'s drain effect each time.

A `REPORT_CONSUMED` operation dispatched from the hook after it emits would truncate the array directly, and let `version` come out of `ReducerConfig`, `NavigationTreeReport`, the `report.version === version` filter, and this effect. That removes the extra render, keeps `config` stable across actions, and leaves one place that decides an event has been delivered.

(If `version` stays, the reducer's reset condition is worth a comment — that it exists to bound the `events` array is not obvious from the code.)

packages/expo-router/src/global-state/removalPrevention.tsx
```diff
@@ -0,0 +1,189 @@
...
+  const emitters = React.useRef(new Map<string, RemovalEventEmitter>());
+  const emitterRegistry = useMemo<RemovalEventEmitterRegistry>(
+    () => ({
+      registerRouteEmitter(routeKey, emitter) {
+        emitters.current.set(routeKey, emitter);
+      },
+      unregisterRouteEmitter(routeKey, emitter) {
+        // Route providers unmount before post-commit `removed` delivery. Keep this emitter through
+        // the current task, unless another provider has already registered for the same route.
+        queueMicrotask(() => {
```
jakub-agent, line 51: [💡 suggestion] This is the same microtask-ordering assumption you removed from `useEventEmitter` as "not the responsibility of this hook" — it moved here rather than being resolved, and it's now the thing holding up `removed` delivery for an unmounting route (see my comment on `types.tsx`).

Worth noting explicitly: the identity check makes it correct but also makes it load-bearing in a non-obvious way — the register-then-unregister ordering across a re-render only works because the new emitter is already in the map when the deferred delete checks. A test would protect that: unmount a route and assert `removed` still reaches its emitter in the same task, and that the entry is gone after `await Promise.resolve()`. Right now nothing covers this path directly.

packages/expo-router/src/global-state/removalPrevention.tsx
```diff
@@ -0,0 +1,189 @@
...
+      {children}
+    </ScreenRemovalPreventionSetterContext.Provider>
+  );
+}
+
+export function useRoutesWithRemovalPrevented() {
+  return use(GlobalRoutesWithRemovalPreventedContext) ?? [];
+}
+
+export function isRouteRemovalPrevented(
```
jakub-agent, line 168: [💡 suggestion] This allocates a `Set` and walks the route's whole nested subtree on every call, and every caller invokes it per route per render:

- `NativeStackView.native.tsx` calls it once per `SceneView`.
- `ExperimentalStackView.tsx` calls it inside the route map.
- `useInvalidPreventRemoveError.tsx` calls it inside `Object.values(descriptors).find(...)`.

So a stack of N screens does N `Set` constructions and N subtree walks per render, and `useInvalidPreventRemoveError` does it again over the same descriptors. The list is usually empty or tiny, so this is not urgent, but the shape is easy to fix: take a `ReadonlySet<string>` here and have `useRoutesWithRemovalPrevented` return a memoized set (the provider already builds `preventedRouteKeys` in a `useMemo`, so it can expose both).

Hoisting the set also lets the early-exit read a little more directly, since `route.key !== undefined && preventedRouteKeys.has(route.key)` no longer has to pay for construction first.

packages/expo-router/src/global-state/__tests__/useNavigationTreeReportEvents.test.ios.tsx
```diff
@@ -0,0 +1,44 @@
...
+  index: 0,
+  routeNames: ['index'],
+  routes: [{ key: 'index', name: 'index' }],
+};
+
+function wrapper({ children }: PropsWithChildren) {
+  return <RemovalPreventionProvider>{children}</RemovalPreventionProvider>;
+}
+
+test('emits only events appended to the current report', () => {
```
jakub-agent: [💡 suggestion] This covers the append path for one event type. Three things the hook does that nothing exercises:

- **Version reset.** A report with a new `version` must restart at index 0 even though `consumedReport.current.eventCount` is non-zero. That branch is the whole reason `version` is on the report, and getting it wrong either drops or double-emits events.
- **Route events.** `prevented-routes` and `removed-routes` reaching `emitterRegistry.emitRemovalEvent` with the right route keys and action — currently only `action-dispatched` is asserted.
- **StrictMode.** The `consumedReport` ref is what makes a re-invoked layout effect slice to an empty list instead of emitting twice. That's a nice property and it would be easy to regress without a test naming it.

The first two are cheap additions to this file since the hook takes the report directly.

packages/expo-router/src/global-state/__tests__/removalPrevention.test.ios.tsx
```diff
@@ -0,0 +1,67 @@
...
+  return (
+    <RemovalPreventionProvider>
+      <PreventRemovalProvider routeKey="parent">
+        <PreventRemovalProvider routeKey="child">{children}</PreventRemovalProvider>
+      </PreventRemovalProvider>
+    </RemovalPreventionProvider>
+  );
+}
+
+test('aggregates callers and publishes the prevented route', () => {
```
jakub-agent: [🐧 nit] The wrapper nests `parent` and `child` providers, but only `child`'s setter is reachable through `ScreenRemovalPreventionSetterContext`, so `parent` only proves it doesn't interfere — the assertions would pass with a single provider.

Now that parent propagation is gone and the global map is keyed by route, the thing worth pinning is that two routes aggregate independently: set prevention on `parent` and on `child`, assert both keys appear, clear one, assert only the other remains. That is the behavior `RoutesWithRemovalPreventedProvider`'s map-of-sets exists for, and the current test can't distinguish it from a single-key implementation.

packages/expo-router/src/react-navigation/core/BaseNavigationContainer.tsx
```diff
@@ -135,6 +131,7 @@ function BaseNavigationContainerInner({
   });
 
   const dispatchSync = useLatestCallback((action: NavigationAction) => {
+    // TODO(@ubax): Throw if this is called from a `removePrevented` callback.
```
jakub-agent, line 135: [🐧 nit] Thanks for restoring `blocks synchronous redispatch from removePrevented without re-emitting` as `test.skip` — that documents the gap properly, so I'm not going to push on it again.

One thing worth confirming before this ships: without a guard, a callback that redispatches walks into `Maximum update depth exceeded` with no indication that `usePreventRemove` is involved. The old code's dev warning named the cause and the fix. If landing the throw here is out of scope, a dev-only `console.error` on the second consecutive prevented report for the same action would keep the diagnosis cheap. Either way, please make sure ENG-26123 (or a follow-up) actually tracks it, since the skipped test on its own won't surface in triage.

jakub-agent (changes requested):

**Iteration 2** — re-review of `22babee` after the rebase onto `@ubax/remove-global-store`. Most of iteration 1 is resolved, and several fixes are better than what I suggested.

**Resolved**

- `report` now carries an `events[]` array that accumulates across a drain batch, and `useNavigationTreeReportEvents` slices from the last consumed index — so batched actions no longer lose `actionDispatched` / `removed`. The `consumedReport` ref also makes the hook StrictMode-safe.
- `ACTIONS_WITHOUT_REMOVAL_PREVENTION` exempts `ROUTE_NAMES_CHANGED`, so route reconciliation is no longer vetoable.
- `usePreventRemove` returns `disablePrevention`, and the same-tick recovery flow is back and tested on both web and iOS. Collapsing the screen-level provider into the route-keyed global map is what makes this work — the prevention change and the queued action now land in the same render, so `config` is fresh when the drainer runs. Good fix.
- `useEventEmitter` is generic again (`Set`-based, no `isTargetMounted`, no `queueMicrotask`); `isPreloaded` / `propagateToParent` are gone, so the stale `getState()` read is gone with them; `emitRemovalEvent` is a `useCallback`; `NAVIGATOR_ACTION` is deleted; `useIsRemovalPrevented` is dropped; the CHANGELOG has its PR link; the enqueue-time TODO is in place; and the re-entrancy test is back as `test.skip`.

**Two things I'd still hold on**

1. `removed` now requires `return () => queueMicrotask(unsubscribe)` at every call site. Taking the deferral out of `useEventEmitter` was right, but it moved into the public contract rather than disappearing — the ordinary `return navigation.addListener(...)` idiom that every other event uses silently never fires.
2. `disablePrevention()` permanently disables prevention if the caller doesn't also flip `preventRemove`, because the layout effect's deps don't change and nothing re-arms it. Both new tests flip the state as well, so the gap is untested. Your TODO on line 76 names this.

The rest is smaller: a missing effect dep, the now-redundant `version` latch, per-render `Set` allocation in `isRouteRemovalPrevented`, and test-coverage gaps in the two new test files.

Same caveat as last time: I could not run the suite (stale local install), so this is from reading.

The PR body is still the unfilled template.

packages/expo-router/src/react-navigation/core/types.tsx
jakub-agent: Acked — leaving the `queueMicrotask(unsubscribe)` contract as-is for now, per the author. Dropping this as a blocker; the JSDoc example on this type is the documentation of record. Worth revisiting if `removed` picks up more consumers, but not something to hold the PR on.

packages/expo-router/src/global-state/removalPrevention.tsx
Ubax: Is this still needed? Now that we require users to call `unsubscribe` from `queueMicrotask`

packages/expo-router/src/global-state/removalPrevention.tsx
Ubax: Do it

packages/expo-router/src/global-state/__tests__/useNavigationTreeReportEvents.test.ios.tsx
Ubax: Add tests

packages/expo-router/src/global-state/__tests__/removalPrevention.test.ios.tsx
Ubax: Fix it

packages/expo-router/src/react-navigation/core/usePreventRemove.tsx
Ubax:
```
React.useEffect(
  () => navigation.addListener('removePrevented', removePreventedListener),
  [navigation, removePreventedListener]
);
```

packages/expo-router/src/react-navigation/core/TabNavigationsContext.tsx
```diff
@@ -0,0 +1,8 @@
+'use client';
+
+import { createContext } from 'react';
+
+import type { ParamListBase } from '../routers';
+import type { NavigationProp } from './types';
+
+export const TabNavigationsContext = createContext<NavigationProp<ParamListBase>[]>([]);
```
jakub-agent: [💡 suggestion] This context is never read for anything. The only consumer is `useNavigationBuilder`, and it only reads it to append to it:

```tsx
const parentTabNavigations = use(TabNavigationsContext);
...
const tabNavigations = React.useMemo(
  () => [...parentTabNavigations, ...(router.type === 'tab' && focusedNavigation ? [focusedNavigation] : [])],
  [focusedNavigation, parentTabNavigations, router.type]
);
...
<TabNavigationsContext.Provider value={tabNavigations}>
```

Nothing else in `packages/` touches `TabNavigationsContext`, `parentTabNavigations`, or `focusedNavigation` — the value accumulates up the tree and is dropped. So the net effect today is an extra provider element in every navigator's `NavigationContent`, plus a `useMemo` that invalidates on every focus change in a tab navigator and re-renders that provider's subtree.

It also looks unrelated to this PR — nothing in the `onActionDispatched` or removal-prevention work reads it. Either drop it here and bring it back with the change that needs it, or land the consumer in this PR so the cost is paying for something. If it's scaffolding for a follow-up, a `TODO` naming that follow-up would at least explain why it exists.

packages/expo-router/src/global-state/useNavigationTreeReportEvents.ts
```diff
@@ -0,0 +1,60 @@
...
+    if (report === undefined) {
+      return;
+    }
+
+    const ids: number[] = [];
+    for (const event of report.events) {
+      if (consumedIds.current.has(event.id)) {
+        continue;
+      }
+      consumedIds.current.add(event.id);
```
jakub-agent, line 33: [💡 suggestion] Marking the id consumed before emitting means a throwing listener leaves the event in a half-delivered state.

If `emitRemovalEvent` or an `actionDispatched` subscriber throws, the loop aborts: this event's id is already in `consumedIds`, but `consumeReportEvents(ids)` at the bottom never runs. So the event stays in `report.events` while the ref believes it was handled — it is never emitted again and never filtered out by `REPORT_CONSUMED`, so the array keeps one dead entry per occurrence. The events after it in the same report are skipped too, and only get delivered on the next action's render.

`RoutingQueueDrainer` already isolates this per item, and the same shape would work here:

```ts
for (const event of report.events) {
  if (consumedIds.current.has(event.id)) continue;
  consumedIds.current.add(event.id);
  ids.push(event.id);
  try {
    switch (event.type) { /* ... */ }
  } catch (error) {
    console.warn(`A navigation event listener threw while handling ${event.type}: ${error}`);
  }
}
```

That keeps one bad listener from starving the others and keeps the report truncation honest. Low priority — a throw out of a layout effect is already going to hit an error boundary — but the inconsistency with the drainer is worth closing while this code is fresh.

packages/expo-router/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### 🛠 Breaking changes
 
+- Remove `beforeRemove`, `__unsafe_action__`, `PreventRemoveContext`, and `usePreventRemoveContext` from `expo-router/react-navigation`. ([#49408](https://github.com/expo/expo/pull/49408) by [@Ubax](https://github.com/Ubax))
```
jakub-agent, line 7: [🐧 nit] You said last round to add only the PR link and no further entries, which I took — but two public additions have landed since then that weren't part of that call:

- `usePreventRemove` now **returns** `disablePrevention`. That's a new return value on an existing public hook, and it's the documented way to continue from the same handler.
- `removed` is a new event on `EventMapCore`, with the `queueMicrotask(unsubscribe)` subscription contract documented on the type.

Both are things a user upgrading would want to find in the changelog rather than in a JSDoc. Your call — flagging once since the situation changed, not to relitigate.

packages/expo-router/src/react-navigation/core/usePreventRemove.tsx
```diff
@@ -1,64 +1,96 @@
...
+
+  React.useEffect(() => {
+    if (!shouldCheck) {
+      return;
+    }
+
+    setShouldCheck(false);
+    if (preventRemove) {
+      console.warn(
+        '`disablePrevention` from `usePreventRemove` was called, but `preventRemove` is still ' +
```
jakub-agent, line 24: [🐧 nit] This warns on correct code when the caller syncs `preventRemove` asynchronously.

The effect reads `preventRemove` on the commit right after `markDisabled()`, so a handler like:

```tsx
const discard = async () => {
  disablePrevention();
  await saveDraft();
  setDirty(false);
};
```

still has `preventRemove === true` when the check runs, and warns even though the prop does get synced a tick later. The message then tells the user to do something they are already doing.

Not worth much machinery, but deferring the check — only warn if `preventRemove` is still `true` after the *next* prevented action, or after a microtask — would cut the false positive. Alternatively soften the wording to note it fires if the prop hasn't synced *yet*, so an async handler doesn't read as a bug.

jakub-agent (approved):

**Iteration 3** — re-review of `7fdc6465` (still stacked on `@ubax/remove-global-store`). Everything I raised in iteration 2 is either fixed or deliberately accepted. No correctness findings left, so this is an approve with four cleanup notes.

**Resolved since iteration 2**

- The `version` latch is gone. `REPORT_CONSUMED` + a monotonic `eventSeq` replaces it, so `report` events are now consumed explicitly by id, `config` no longer churns per navigation, and `processIntent` is stable across actions. This is a better shape than what I suggested — the reducer no longer needs to know anything about delivery.
- `routesWithRemovalPrevented` is a `ReadonlySet` end to end, `isRouteRemovalPrevented` takes the set instead of rebuilding one per route, and `EMPTY_SET` is shared. The per-render allocation in `NativeStackView` / `ExperimentalStackView` / `useInvalidPreventRemoveError` is gone.
- `usePreventRemove`'s `removePrevented` effect dropped the ternary, so `callback` is no longer read outside its deps.
- `disablePrevention` desync now has a dev-only warning that names the cause and the fix, tested in both directions. Not a behavioral fix, but the wrong usage is now loud in development, which is where it would be written. Reasonable trade-off — I'm not pushing further.
- `queueMicrotask` in `unregisterRouteEmitter` is covered by `keeps a route emitter until the end of the task after its provider unmounts`, which pins both the deferral and the eventual cleanup.
- `useNavigationTreeReportEvents` now has route-event, per-id consumption, and StrictMode coverage; `removalPrevention.test` aggregates across two sibling routes instead of one.
- `useDescriptors` derives preloaded-ness from the rendered `state` rather than a ref read.

**Accepted, not re-raised:** the `removed` + `queueMicrotask(unsubscribe)` consumer contract, and the missing `dispatchSync` re-entrancy guard (documented by the re-added `test.skip`).

**Remaining, none blocking:** an unused `TabNavigationsContext` added by this PR, two CHANGELOG lines for the new public surface, one dev-warning false positive, and per-listener error isolation in the emit loop.

Standing caveat: I still could not run the expo-router suite locally (stale install), so this approval is on design and code reading only — CI's result stands on its own.

packages/expo-router/src/react-navigation/core/TabNavigationsContext.tsx
Ubax: Delete the context

packages/expo-router/src/global-state/useNavigationTreeReportEvents.ts
Ubax: Add it

PR #49415, opened 2026-08-26
[ios][expo-go] Fix missing modules

alanjhughes:
# Why
Expo Go on iOS silently lost four modules. expo-apple-authentication, expo-age-range, expo-live-photo and expo-mesh-gradient. The regression came in with the React Native 0.86 upgrade (#46534). Before that, iOS autolinking discovered modules by scanning broadly, so packages that were never declared in the package.json still got linked. The new autolinking resolves modules from the project's dependency graph, so these four fell out of the build.

# How
Added the four packages as dependencies so autolinking picks them up again.

# Test Plan
Expo go. The modules are now all linked and work correctly


chrfalch (approved):
Thanks!

brentvatne (approved):

PR #49416, opened 2026-08-26
[cli] Invalidate transforms when Expo config changes

huytdps13400:
## Summary

`expo export` previously reused Metro transforms containing the Expo config from an earlier invocation. Dynamic config changes therefore reached `expo config` but not the exported client bundle unless users passed `--clear`.

This adds a SHA-1 fingerprint of the public Expo config to Metro `customTransformOptions` for legacy client bundles. Metro can keep its normal cache, while transforms that inline `APP_MANIFEST` are separated whenever the config changes. The same fingerprint is applied to SPA/native exports, RSC client-boundary bundles, and DOM component bundles.

Closes #39619.

## Test Plan

- Added a focused regression that creates two public Expo configs with different `extra.API_BASE_URL` values.
  - RED: both bundle option sets had no config cache discriminator.
  - GREEN: both contain string fingerprints and the fingerprints differ.
- `pnpm test -- src/start/server/middleware/__tests__/metroOptions.test.ts --runInBand` — 11/11 pass.
- `pnpm lint` in `packages/@expo/cli` — pass.
- `pnpm format --check` in `packages/@expo/cli` — pass across 831 files.
- `git diff --check` — pass.

I used a sparse checkout limited to `.github`, `packages/@expo/*`, and the direct CLI workspace dependencies to avoid downloading/building the full Expo monorepo. Full package typecheck/build could not complete because workspace build outputs outside that sparse checkout were absent (for example `@expo/metro-config/build`, `@expo/router-server/build`, and Expo Router native dependencies). The focused regression, lint, and format gates above are green; CI should run the complete workspace graph.

### DIFF
--- a/packages/@expo/cli/CHANGELOG.md
+++ b/packages/@expo/cli/CHANGELOG.md
@@ -24,6 +24,7 @@
 
 ### 🐛 Bug fixes
 
+- Invalidate Metro transforms when the Expo config embedded in client bundles changes.
 - Serve relative manifest URLs only when the client itself sends the RFC 7239 `Forwarded` header, so that proxied requests from clients without relative-URL support, like released Expo Go versions through the WS tunnel, keep absolute URLs. ([#48997](https://github.com/expo/expo/pull/48997) by [@expo-bot](https://github.com/expo-bot))
 - Fail when `--private-key-path` is passed without `updates.codeSigningCertificate` in the resolved app config, instead of ignoring the flag and continuing without signing.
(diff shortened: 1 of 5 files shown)


kitten:
See: https://github.com/expo/expo/pull/49264#issuecomment-5391685714

This is essentially a duplicate, so the comment applies here too

huytdps13400:
Confirmed against #49264 and the active implementation in #47750. Hashing the entire resolved Expo config is too broad because it includes incidental values such as resolved absolute paths, causing unstable and unnecessary cache invalidation. The cache-vary mechanism in #47750 is the intended per-module solution for ambient transform inputs.

Closing this duplicate in favor of #47750.

PR #49418, opened 2026-08-26
[tools][platforms] Add @expo/platforms and consume it across tooling

douglowder:
# Why

Platform data and knowledge is duplicated across `@expo/config`, `@expo/metro-config`, `expo-modules-autolinking`, and `@expo/cli`, and the copies have started to drift (metro-config's host-package switch carried a comment noting it duplicated autolinking's). Those packages could not share the code from any existing package, due to dependency constraints: autolinking deliberately keeps a minimal dependency list and metro-config does not depend on it.

# How

Add `@expo/platforms`, a zero-dependency package holding platform identity metadata: the known platform names, which of them are out-of-tree, Apple-family membership, the react-native host package for each platform (tvos -> react-native-tvos, macos -> react-native-macos, windows -> react-native-windows), and the Metro source-extension fallback chains for out-of-tree platforms.

How the new package is used:    
    
- `@expo/config`: Platform aliases the shared PlatformName (an identical union); getSupportedPlatforms derives its out-of-tree probes from the host package map; the PLATFORM_EXTENSIONS table moves out of paths/extensions.
- `@expo/metro-config`: getReactNativeHostPackage keeps its default-to-react-native semantics while reading fork names from the shared map; resolver.platforms and unstable_conditionsByPlatform derive their out-of-tree entries, preserving order (snapshots unchanged).
- `expo-modules-autolinking`: getSupportPackageForPlatform keeps its exact null/throw semantics and reads fork names from the map; SupportedPlatform is expressed via PlatformName; supportsPlatform, resolveReactNativeModule and the react-native-config command use the shared Apple-family and out-of-tree predicates.
- `@expo/cli`: the five-name platform list written out in three places becomes KNOWN_PLATFORMS (same contents and order, so error messages are unchanged); export and Metro code use the shared predicates; platformBundlers derives its out-of-tree entries; REACT_NATIVE_TVOS_PACKAGE_NAME reads from the map.
    
This work is a prerequisite for future PRs that would add improved support for OOT platforms like tvos and macos, including prebuild support.

# Test Plan

New tests added; CI should pass.

Includes a fix for an `@expo/prebuild-config` testing-library failure that happens if running locally with terminal colors enabled.

# Checklist

---


douglowder:
Current `check-packages` failure is an unrelated test failure in `expo-router`.

packages/@expo/cli/src/export/resolveOptions.ts
```diff
@@ -55,7 +56,7 @@ export function resolvePlatformOption(
     return platform;
   };
 
-  const knownPlatforms = ['android', 'ios', 'web', 'tvos', 'macos'] as Platform[];
+  const knownPlatforms = KNOWN_PLATFORMS as readonly Platform[];
```
kitten, line 59: Ideally, I think the idea is we shouldn't be recasting here and the `Platform` type from `@expo/config` should be replaced?

packages/@expo/metro-config/src/ExpoMetroConfig.ts
```diff
@@ -58,15 +63,10 @@ export interface DefaultConfigOptions {
...
-      return 'react-native-tvos';
-    case 'macos':
-      return 'react-native-macos';
-    default:
-      return 'react-native';
-  }
+  // Only out-of-tree platforms resolve to a react-native fork here; everything else — including
+  // platforms that expo-modules-autolinking's `getSupportPackageForPlatform` treats specially —
+  // falls back to react-native.
+  return isOutOfTreePlatform(platform) ? getHostPackageForPlatform(platform)! : 'react-native';
```
kitten, line 69: nit: trivial comment

packages/@expo/platforms/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/expo/expo.git",
+    "directory": "packages/@expo/platforms"
+  },
+  "files": [
+    "build"
+  ],
+  "main": "build/index.js",
+  "types": "build/index.d.ts",
```
kitten, line 24: We're missing `exports` here (needs `expo-source` redirect too, like the others

packages/@expo/platforms/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "test": "jest",
+    "depscheck": "expo-module depscheck",
+    "typecheck": "tsc -p tsconfig.json"
+  },
+  "devDependencies": {
+    "expo-module-scripts": "workspace:*"
+  },
+  "engines": {
+    "node": ">=20.12.0"
+  }
```
kitten, line 43:
```suggestion
```

nit: by convention we haven't yet added this to all packages. I think, given that this package is a leaf package that has no specific requirements, it'd be better to omit this

packages/@expo/cli/src/start/server/middleware/resolvePlatform.ts
```diff
@@ -51,7 +52,7 @@ export function assertMissingRuntimePlatform(platform?: any): asserts platform {
 /** Assert if the runtime platform is not correct. */
 export function assertRuntimePlatform(platform: string): asserts platform is RuntimePlatform {
   const stringifiedPlatform = String(platform);
-  if (!['android', 'ios', 'web', 'tvos', 'macos'].includes(stringifiedPlatform)) {
+  if (!(KNOWN_PLATFORMS as readonly string[]).includes(stringifiedPlatform)) {
```
Kudo: **Medium:** Membership now uses `KNOWN_PLATFORMS`, but the error string still hardcodes the five names. Prefer deriving the message from `KNOWN_PLATFORMS` (same approach as `resolveOptions`) so a future platform addition cannot drift the check vs the user-facing error.

```ts
`platform must be ${KNOWN_PLATFORMS.map((p) => `"${p}"`).join(', ').replace(/, ([^,]*)$/, ', or $1')}. Received: "${platform}"`
```
(or a small shared formatter).

packages/@expo/platforms/package.json
Kudo: **Medium (agree with @kitten):** Please add an `exports` map with `expo-source` → `./src/index.ts` (and types) matching leaf packages like `@expo/json-file`. Right now `main`/`types` point at `build/`, but this PR does not commit `build/`, so monorepo consumers that rely on the `expo-source` condition will not resolve source correctly the way other workspace packages do.

packages/@expo/platforms/src/index.ts
```diff
@@ -0,0 +1,83 @@
...
+ */
+export const REACT_NATIVE_HOST_PACKAGES: Readonly<Record<string, string>> = {
+  android: 'react-native',
+  ios: 'react-native',
+  tvos: 'react-native-tvos',
+  macos: 'react-native-macos',
+  windows: 'react-native-windows',
+};
+
+/**
```
Kudo: **Low:** `windows` is in this map (for autolinking) but is intentionally *not* in `PlatformName` / `OUT_OF_TREE_PLATFORMS`, so `isOutOfTreePlatform('windows')` stays `false` and metro-config will keep falling back to `react-native`. Worth a one-line note here that adding real windows OOT support requires updating `PlatformName` + `OUT_OF_TREE_PLATFORMS` (not just this map), to avoid a future silent host-package mismatch.

packages/@expo/cli/src/export/resolveOptions.ts
Kudo: **Medium (agree with @kitten):** Now that `@expo/config`'s `Platform` aliases `PlatformName`, this cast should be unnecessary — prefer using `KNOWN_PLATFORMS` directly (or typing the local as `readonly PlatformName[]`) so we do not keep two platform unions in sync by hand.

Kudo:
@expo-bot verify --fable

packages/@expo/cli/src/export/resolveOptions.ts
douglowder: I'll look into that. The idea was to put a layer of indirection between `@expo/platforms` and other packages that might have incompatible dependencies, or are effectively wrapping this to have their own platform types.

packages/@expo/metro-config/src/ExpoMetroConfig.ts
```diff
@@ -58,15 +63,10 @@ export interface DefaultConfigOptions {
...
-    case 'tvos':
-      return 'react-native-tvos';
-    case 'macos':
-      return 'react-native-macos';
-    default:
-      return 'react-native';
-  }
+  // Only out-of-tree platforms resolve to a react-native fork here; everything else — including
+  // platforms that expo-modules-autolinking's `getSupportPackageForPlatform` treats specially —
+  // falls back to react-native.
```
douglowder:
```suggestion
```

packages/@expo/platforms/package.json
douglowder: Fixed

packages/@expo/cli/src/export/resolveOptions.ts
douglowder: Confirmed, this was done to keep the new package as a "zero-dependency data package" below all the others. I remember going back and forth with Claude about this... Claude's detailed summary is in https://github.com/expo/expo/pull/49418/commits/91aa2c054b271e40a45918682fb4361f748b8dce

PR #49419, opened 2026-08-26
[android][notifications] Keep the response PendingIntent readable under a foreign class loader

### LINKED ISSUE #49252: NotificationForwarderActivity crashes with "notification (null) and action (null) should not be null" on expo-notifications 0.32.17 (Android) — recurs  with foreground states

https://github.com/BasantaParajuli22/expo-notification-issue

### Steps to reproduce

1. Send a push notification to the app (FCM v1 via EAS push service)
2. Tap the notification
3. Crash occurs — reproducible in foreground, background, AND killed app states

additional context

This matches #38908 / fixed by #43203, but still reproducing on 0.32.17. 
I attempted a workaround wrapping NotificationForwarderActivity.onCreate 
in try/catch via patch-package. The patch-package log confirms successful 
application during EAS build ("expo-notifications@0.32.17 ✔"), but the 
resulting crash trace is unchanged, including identical line numbers, 
suggesting either the patch isn't reaching the compiled APK or the 
exception originates from a different code path than expected.

and it relates to #38908, #43203, #47889

### Environment

```text
ptg@ptg-Latitude-7420:~/repo/notif-crash-repro$ npx expo-env-info

    System:
      OS: Linux 6.8 Ubuntu 22.04.5 LTS 22.04.5 LTS (Jammy Jellyfish)
      Shell: 5.1.16 - /bin/bash
    Binaries:
      Node: 24.14.1 - /home/ptg/.nvm/versions/node/v24.14.1/bin/node
      npm: 11.12.0 - /home/ptg/.nvm/versions/node/v24.14.1/bin/npm
    npmPackages:
      expo: ~54.0.0 => 54.0.37 
      react: 19.1.0 => 19.1.0 
      react-native: 0.81.5 => 0.81.5 
    npmGlobalPackages:
      eas-cli: 21.0.2
    Expo Workflow: managed

ptg@ptg-Latitude-7420:~/repo/notif-crash-repro$ 
some more info
Expo SDK: 54.0.0
expo-notifications: 0.32.17
Platform: Android
Device/OS version: android version 12
```

### Expo Doctor Diagnostics

```text
ptg@ptg-Latitude-7420:~/repo/notif-crash-repro$ npx expo-doctor@latest 
```

LizunovSergey:
# Why

Fixes #49252, and closes the loop on #38908 / #43203 / #47889 — the same notification-tap failure has now come back three times.

A notification tap goes out through a `PendingIntent` built by `createNotificationResponseIntent`. That intent carried the `Notification` and the `NotificationAction` twice: once as custom `Parcelable` extras, and once as marshalled byte arrays. The byte arrays were added in #43203 as a fallback for when the `Parcelable` extras come back `null`.

The fallback cannot fire. A `Bundle` is unparcelled **as a unit** — one `Parcelable` whose class the loader cannot resolve fails the entire `Bundle`, not just that key. When defusing is on the map is erased; when it is not, `BadParcelableException` is thrown. Either way the byte arrays are gone too, because they live in the same `Bundle` as the `Parcelable` that killed it. So the fallback is destroyed by the very thing it was meant to survive.

Two symptoms follow, both reported:

- `IllegalArgumentException: notification (null) and action (null) should not be null` — the erased-map case, both keys read back null.
- A hard crash — `BadParcelableException` is not an `IllegalArgumentException`, so `NotificationForwarderActivity`'s `catch (e: IllegalArgumentException)` never covered it.

`NotificationsService` already knows about this class loader, three functions further down:

```kotlin
// Class loader used in BaseBundle when unmarshalling notification extras
// cannot handle expo.modules.notifications.….NotificationResponse
// so we go around it by marshalling and unmarshalling the object ourselves.
fun setNotificationResponseToIntent(...)
```

That function stores the response as bytes and **only** as bytes. The response intent didn't follow the same rule.

# How

Three changes, in `expo-notifications` Android only:

1. **`createNotificationResponseIntent`** — drop the two custom `Parcelable` extras and keep the byte arrays. The `Bundle` then holds nothing the system's loader can trip on. Both readers (`createNotificationResponseBroadcastIntent`, `getNotificationResponseFromBroadcastIntent`) already prefer the `Parcelable` and fall back to bytes, so they keep working unchanged — the fallback simply becomes the path that is actually taken.

2. **`NotificationForwarderActivity`** — call `setExtrasClassLoader` before anything reads the extras. This is for the upgrade path: a `PendingIntent` created by an older version is still sitting in the system with the old shape, and it survives an app update. The `Bundle` unparcels lazily, so setting the loader in `onCreate` is early enough to save that first read. (Verified both ways: repairing the loader *before* the first read recovers everything; repairing it *after* a failed read does not.)

3. **`NotificationForwarderActivity`** — widen `catch (e: IllegalArgumentException)` to `RuntimeException` so `BadParcelableException` lands in the existing recovery instead of taking the app down. The recovery is unchanged: log, open the app to the foreground anyway.

**What I deliberately did not change.** `createNotificationResponseBroadcastIntent` still puts its `Parcelable` extras, and I did not add byte arrays there. That leg is broadcast to the app's own receiver, where the loader normally resolves fine, and no report points at it. Adding byte arrays *alongside* the `Parcelable`s there would have been dead weight for the exact reason described above — I wrote that test, watched it fail, and removed the change rather than widen the blast radius on a leg that isn't demonstrably broken. Worth a separate look if anyone has seen it fail.

One compatibility note worth a maintainer's eye: the response `PendingIntent` now carries the payload only under `NOTIFICATION_BYTES_KEY` / `NOTIFICATION_ACTION_BYTES_KEY`, which are `internal`. Anything outside the module reading `NOTIFICATION_KEY` straight off *that* intent would no longer find it. Nothing in the repo does, and the supported readers are unaffected, but it is a real if narrow surface change.

# Test Plan

New regression test in `NotificationsServiceResponseIntentTest`: build the real response `PendingIntent`, send it through a `Parcel` the way the system does, hand it a class loader that refuses `expo.modules.notifications.*`, then read it back.

Red/green, on `packages/expo-notifications`:

```
# with the Parcelable extras restored (before state)
NotificationsServiceResponseIntentTest > response intent extras survive a class loader
that cannot resolve the model classes FAILED
    android.os.BadParcelableException at NotificationsServiceResponseIntentTest.kt:148

# with the fix
BUILD SUCCESSFUL
NotificationsServiceResponseIntentTest > response intent extras survive a class loader
that cannot resolve the model classes PASSED
...4 tests, all passing
```

Full module suite green: `./gradlew :expo-notifications:testDebugUnitTest` → `BUILD SUCCESSFUL`, including the three pre-existing tests from #43203, which still pass unmodified.

The existing tests missed this because they assemble the intent by hand and read it in-process — never crossing a `Parcel`, and never with a loader that cannot see the model classes, which is the only condition under which the bug exists.

I could not run `et check-packages expo-notifications` locally — it stops at `ERR_PNPM_UNSUPPORTED_ENGINE` (repo wants pnpm `^10.33.0`, I have 10.11.0). The diff is Kotlin plus one changelog line and touches no JS, so nothing in that check applies to it, but flagging it rather than implying I ran it.

Not verified on a physical device: I do not have a repro of the original tap crash in hand, and the failing condition is a class loader the app does not control. The test reproduces that condition directly instead.

# Checklist

- [x] Native unit tests pass (`:expo-notifications:testDebugUnitTest`)
- [ ] `et check-packages` — blocked locally by the pnpm engine pin, see above
- [x] No `npx expo prebuild` / EAS-relevant surface in this change

### DIFF
--- a/packages/expo-notifications/CHANGELOG.md
+++ b/packages/expo-notifications/CHANGELOG.md
@@ -15,6 +15,7 @@
 ### 🐛 Bug fixes
 
 - [iOS] Fix a data race on `NotificationCenterManager`'s delegate list that crashed the app with `SIGSEGV` when one app context registered its modules while another tore its own down, such as on a dev-client reload or `Updates.reloadAsync()`. [#49554](https://github.com/expo/expo/pull/49554) by [@dennytosp](https://github.com/dennytosp))
+- [Android] Fixed notification taps losing their payload, and crashing with `BadParcelableException`, when the response `PendingIntent` is unparcelled by a class loader that cannot resolve the library's own classes. ([#49252](https://github.com/expo/expo/issues/49252) by [@LizunovSergey](https://github.com/LizunovSergey))
 - [Android] Prevented `onUserLeaveHint` from firing when a notification tap opens the app, which made picture-in-picture implementations enter PiP unexpectedly. ([#48471](https://github.com/expo/expo/pull/48471) by [@stareezy-1](https://github.com/stareezy-1))
 - [iOS] Avoid warning when an aborted push token registration request rejects with a native fetch cancellation error. ([#48547](https://github.com/expo/expo/pull/48547) by [@JoaoPauloCMarra](https://github.com/JoaoPauloCMarra))
 - [Android] Prevent a crash on notification tap when `getLaunchIntentForPackage` throws on some OEM ROMs. ([#47889](https://github.com/expo/expo/pull/47889) by [@nunocaseiro](https://github.com/nunocaseiro))
--- a/packages/expo-notifications/CHANGELOG.md
+++ b/packages/expo-notifications/CHANGELOG.md
@@ -15,7 +15,7 @@
 ### 🐛 Bug fixes
 
 - [iOS] Fix a data race on `NotificationCenterManager`'s delegate list that crashed the app with `SIGSEGV` when one app context registered its modules while another tore its own down, such as on a dev-client reload or `Updates.reloadAsync()`. [#49554](https://github.com/expo/expo/pull/49554) by [@dennytosp](https://github.com/dennytosp))
-- [Android] Fixed notification taps losing their payload, and crashing with `BadParcelableException`, when the response `PendingIntent` is unparcelled by a class loader that cannot resolve the library's own classes. ([#49252](https://github.com/expo/expo/issues/49252) by [@LizunovSergey](https://github.com/LizunovSergey))
+- [Android] Fixed notification taps losing their payload, and crashing with `BadParcelableException`, when the response `PendingIntent` is unparcelled by a class loader that cannot resolve the library's own classes. Direct-reply actions keep their `userText`. ([#49252](https://github.com/expo/expo/issues/49252) by [@LizunovSergey](https://github.com/LizunovSergey))
 - [Android] Prevented `onUserLeaveHint` from firing when a notification tap opens the app, which made picture-in-picture implementations enter PiP unexpectedly. ([#48471](https://github.com/expo/expo/pull/48471) by [@stareezy-1](https://github.com/stareezy-1))
 - [iOS] Avoid warning when an aborted push token registration request rejects with a native fetch cancellation error. ([#48547](https://github.com/expo/expo/pull/48547) by [@JoaoPauloCMarra](https://github.com/JoaoPauloCMarra))
 - [Android] Prevent a crash on notification tap when `getLaunchIntentForPackage` throws on some OEM ROMs. ([#47889](https://github.com/expo/expo/pull/47889) by [@nunocaseiro](https://github.com/nunocaseiro))
--- a/packages/expo-notifications/android/src/main/java/expo/modules/notifications/service/NotificationForwarderActivity.kt
+++ b/packages/expo-notifications/android/src/main/java/expo/modules/notifications/service/NotificationForwarderActivity.kt
@@ -15,13 +15,22 @@ import expo.modules.notifications.service.delegates.ExpoHandlingDelegate
 class NotificationForwarderActivity : Activity() {
   override fun onCreate(savedInstanceState: Bundle?) {
     super.onCreate(savedInstanceState)
+    // A PendingIntent created by an older version of this library still carries the notification
+    // as a Parcelable, and the loader this Bundle arrives with cannot resolve
+    // expo.modules.notifications.… classes. Point it at ours before anything reads the extras -
+    // the Bundle is unparcelled lazily, so doing it here is early enough to save that first read.
+    intent?.setExtrasClassLoader(NotificationsService::class.java.classLoader)
     try {
       val broadcastIntent =
         NotificationsService.createNotificationResponseBroadcastIntent(applicationContext, intent)
       val notificationResponse = NotificationsService.getNotificationResponseFromBroadcastIntent(intent)
       ExpoHandlingDelegate.openAppToForeground(this, notificationResponse)
       sendBroadcast(broadcastIntent)
-    } catch (e: IllegalArgumentException) {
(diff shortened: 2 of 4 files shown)


vonovak:
@expo-bot verify

LizunovSergey:
The bot is right, and this was a real regression — thanks for running it, @vonovak.

Marshalled bytes carry no class name, so `unmarshalObject(NotificationAction.CREATOR, …)` always rebuilt a base `NotificationAction`. The placeholder that `TextInputNotificationAction.writeToParcel` appends stayed unread, both `is TextInputNotificationAction` checks went false, and `userText` never reached JS. Before this PR the `Parcelable` extra was read first and preserved the subclass, so dropping it made the loss universal instead of conditional. Fixed in `9cb9e0ca`.

I took the suggested shape: `TEXT_INPUT_NOTIFICATION_ACTION_BYTES_KEY`, written when `action is TextInputNotificationAction`, tried first in both readers with `TextInputNotificationAction.CREATOR` before falling back to the base key. Same one-key-per-subclass pattern the file already uses for `NotificationResponse`.

Red/green on the two new tests, reproducing the bot's finding:

```
# with the per-subclass key reverted
direct reply keeps its type and user text through the response intent FAILED
    java.lang.AssertionError at NotificationsServiceResponseIntentTest.kt:180
direct reply survives a class loader that cannot resolve the model classes FAILED
    java.lang.AssertionError at NotificationsServiceResponseIntentTest.kt:199

# as submitted
BUILD SUCCESSFUL — 6/6 in NotificationsServiceResponseIntentTest, full module suite green
```

The second test is the one the bot's version did not cover: a reply action that *also* crosses a `Parcel` with a loader that refuses our classes. That is the combination this PR exists for, and it would have been easy to fix the type loss in a way that only worked on the happy path.

On the migration gap the bot flagged — a `PendingIntent` posted by a build of this branch storing the reply action under the base key — it does not arise, since the extra key lands in the same PR before merge. Intents posted by *released* versions still carry the `Parcelable`, which both readers try first, so those keep working.

On your two open questions:

**`catch (e: RuntimeException)` was too wide.** Narrowed to `IllegalArgumentException` and `BadParcelableException` explicitly, with the recovery in a shared private function. Those are the only two the readers actually throw, and swallowing arbitrary `RuntimeException` here would have hidden real programming errors behind "open the app anyway".

**`setExtrasClassLoader` — I did measure it, and the bot is right to be sceptical.** It only helps if the *app's* read is the failing one and we get in before the lazy unparcel. Probing it directly: repairing the loader before the first read recovers both the `Parcelable`s and the bytes; repairing it after a failed read does not, on a device where defusing has already erased the map. So in the bot's case 1 — the system emptying the extras before delivery — the line does nothing.

Its narrow value is the upgrade path: a `PendingIntent` posted by the currently installed version carries the old `Parcelable` shape and survives an app update, so that read is the one it can save. I'm happy to drop it if you'd rather not carry an unproven line — the byte-key change stands on its own and is what actually fixes the reported failure. Your call.

I'd also flag the bot's honesty about #49252 as the right read of my own PR: I demonstrated the *mechanism* (a Bundle holding an unresolvable `Parcelable` fails as a unit, taking the byte fallback with it), not that this is the cause of that specific crash. I said as much in the description and it still stands — I have no device on the affected versions. The change is defensible either way, since it stops putting an unresolvable `Parcelable` in that Bundle at all, but it should not be merged as "fixes the crash" if you want that claim verified first.

LizunovSergey:
Rebased onto current `main`. The only conflicts were in `CHANGELOG.md` (both of my commits touched the same bug-fixes section that the new #49554 entry landed in); resolved by keeping both entries. The Kotlin sources and the test file merged cleanly and are byte-identical to the previous head, so the behaviour — including the direct-reply `TextInputNotificationAction` handling from the earlier review round — is unchanged.

PR #49425, opened 2026-08-27
[location][android] Survive foreground-service eligibility rejection instead of crashing

### LINKED ISSUE #49424: [location] Android: fatal SecurityException from LocationTaskService.startForeground when the app leaves the FGS-eligible state during service bind
## Summary

`Location.startLocationUpdatesAsync` with `foregroundService` options starts the foreground service in two async steps ([`LocationTaskConsumer.maybeStartForegroundService`](https://github.com/expo/expo/blob/main/packages/expo-location/android/src/main/java/expo/modules/location/taskConsumers/LocationTaskConsumer.kt)):

1. `context.startForegroundService(serviceIntent)` + `context.bindService(...)`
2. `Service.startForeground(...)` only later, inside `onServiceConnected`, via `LocationTaskService.startForeground`

Android 14+ evaluates foreground-service-type eligibility at the moment `startForeground()` runs, not when the service was started. If the app drops out of the eligible state between step 1 and step 2 (the user backgrounds the app or locks the device right after starting tracking), the system rejects the call:

```
SecurityException: Starting FGS with type location callerApp=ProcessRecord{...}
targetSDK=36 requires permissions: all of the permissions allOf=true
[android.permission.FOREGROUND_SERVICE_LOCATION] any of the permissions
allOf=false [android.permission.ACCESS_COARSE_LOCATION,
android.permission.ACCESS_FINE_LOCATION] and the app must be in the eligible
state/exemptions to access the foreground only permission
```

The exception is thrown on the main looper inside the `ServiceConnection` callback, i.e. after `startLocationUpdatesAsync`'s promise has already resolved. It cannot be caught from JS, and nothing in expo-location catches it, so the process dies with a fatal crash.

Two guards exist and neither can close the race, because both run before the async bind:

- expo-location's own `AppForegroundedSingleton.isForegrounded` check at the top of `maybeStartForegroundService`
- any app-level guard (we defer `startLocationUpdatesAsync` on Android unless `AppState === "active"` and still see the crash)

All required permissions were granted in the crashing sessions (`ACCESS_FINE_LOCATION`, `ACCESS_COARSE_LOCATION`, `FOREGROUND_SERVICE_LOCATION`); the failing clause is the eligible-state requirement.

## Environment

- expo-location 57.0.9 (Expo SDK 57), `targetSdkVersion` 36
- Observed in production on Android 16 devices; 11 crashes / 11 users in our Sentry over ~6 weeks
- Still present in the latest published code: the relevant code is unchanged from 57.0.9 through 57.0.14 and in the SDK 58 canary (`58.0.0-canary-20260812-27f94d4`)

## Reproduction sketch

Timing-dependent in the wild (the window is the service bind), but the shape is:

1. Define a TaskManager location task and call `Location.startLocationUpdatesAsync(TASK, { foregroundService: { notificationTitle: ..., notificationBody: ... }, ... })` while the app is in the foreground, with while-in-use location granted.
2. Background the app or lock the device immediately after the call, before `onServiceConnected` fires.
3. On Android 14+ (observed on Android 16, targetSDK 36), `Service.startForeground` inside `onServiceConnected` throws the SecurityException above and the process is killed.

For a deterministic reproduction we made `LocationTaskService.startForeground` throw `SecurityException` on its first invocation (simulating the OS rejection at the exact production callsite) on an Android 16 emulator.

## Stack trace (anonymized, production)

```
RemoteException: Remote stack trace:
    at com.android.server.am.ActiveServices.validateForegroundServiceType(...)
    at com.android.server.am.ActiveServices.setServiceForegroundInnerLocked(...)
    at com.android.server.am.ActiveServices.setServiceForegroundLocked(...)
    at com.android.server.am.ActivityManagerService.setServiceForeground(...)
    at android.app.IActivityManager$Stub.onTransact$setServiceForeground$(IActivityManager.java:12985)

Caused by SecurityException: Starting FGS with type location ... and the app
must be in the eligible state/exemptions to access the foreground only permission
    at android.app.Service.startForeground(Service.java:775)
    at expo.modules.location.services.LocationTaskService.startForeground(LocationTaskService.kt:64)
    at expo.modules.location.taskConsumers.LocationTaskConsumer$maybeStartForegroundService$1.onServiceConnected(LocationTaskConsumer.kt:222)
    at android.content.ServiceConnection.onServiceConnected(ServiceConnection.java:75)
    at android.app.LoadedApk$ServiceDispatcher.doConnected(LoadedApk.java:2517)
    ...
    at android.os.Looper.loop(Looper.java:392)
```

## Suggested fix (verified on an Android 16 emulator)

Catching the exception is necessary but not sufficient. We verified two simpler shapes first and both still die, because `startForegroundService()` attaches a promise that the service will reach `startForeground()`, and Android kills the app with `ForegroundServiceDidNotStartInTimeException` whenever that promise breaks:

- catch inside `LocationTaskService.startForeground` + `stopSelf()`: the binding keeps the service alive, the promise stays pending, and the app is killed as soon as it backgrounds
- catch + full teardown (stop + unbind): the teardown itself triggers the enforcement, and the app is killed ~10ms later

What works, verified end to end:

1. Start the service with `context.startService(serviceIntent)` instead of `startForegroundService`. The existing `isForegrounded` check makes this legal (the app is foreground at that point), no promise is armed, and the later `startForeground` call promotes the service to a fully-typed location FGS exactly as before. If the app leaves the foreground before the call, `startService` throws a synchronous, catchable `IllegalStateException` instead of scheduling a delayed crash.
2. Wrap the consumer's `it.startForeground(serviceOptions)` (in `onServiceConnected`) with catches for `SecurityException` / `IllegalStateException` (parent of `ForegroundServiceStartNotAllowedException`, API 31+); on rejection log, `stop()` the service, null `mService`, and `unbindService` so the service is destroyed.
3. Same catch around the restart branch `mService?.startForeground(...)`.

Location tracking then degrades gracefully instead of crashing: the task stays registered, and the next (re)registration while the app is foreground retries the foreground service and recovers (verified: after the simulated rejection the process survives foreground and 90s of background, and the FGS comes up with type location on the next app resume).

We are shipping this as a patch-package fix; PR to follow.

steve-rodri:
## Why

Fixes #49424.

`LocationTaskConsumer.maybeStartForegroundService` starts `LocationTaskService` with `startForegroundService` + `bindService`, and only calls `Service.startForeground` later, inside `onServiceConnected`. Android 14+ checks foreground-service-type eligibility at `startForeground` time, so if the app leaves the eligible state during the bind (user backgrounds or locks right after `startLocationUpdatesAsync`), the system throws `SecurityException` on the main looper, after the JS promise has resolved. Nothing can catch it and the process dies. Neither the existing `isForegrounded` check nor a JS-level `AppState` guard can close the race, because both run before the async bind.

## How

Two coordinated changes in `LocationTaskConsumer`, both verified on an Android 16 (targetSDK 36) emulator:

1. `context.startService(serviceIntent)` instead of `context.startForegroundService(serviceIntent)`. The `startForegroundService` promise (the service MUST reach `startForeground`) cannot be kept when Android rejects the foreground start for eligibility, and any way of breaking it, including tearing the service down cleanly, kills the app with `ForegroundServiceDidNotStartInTimeException` (both verified). The `isForegrounded` check directly above makes `startService` legal at that point, no promise is armed, and the `startForeground` call in `onServiceConnected` promotes the service to a fully-typed location FGS exactly as before. If the app leaves the foreground before the call, `startService` throws a synchronous, catchable `IllegalStateException` instead of scheduling a delayed crash.
2. Catch `SecurityException` / `IllegalStateException` (parent of `ForegroundServiceStartNotAllowedException`, API 31+) around both `startForeground` call sites. On rejection in `onServiceConnected`: log, `stop()` the service, null `mService`, and `unbindService` so the service is destroyed. On rejection in the restart branch: log and stop.

The task stays registered, so tracking degrades gracefully and the next (re)registration while the app is foreground retries the foreground service and recovers.

## Test Plan

On an Android 16 (targetSDK 36) emulator, with a production app that registers a location task with `foregroundService` options, and `LocationTaskService.startForeground` locally instrumented to throw `SecurityException` on its first invocation (simulating the OS rejection at the exact production callsite):

- Without this change: the first attempt kills the app (in production as the uncatchable `SecurityException`; with intermediate catch-only variants as `ForegroundServiceDidNotStartInTimeException`).
- With this change: the rejection is caught and logged, the service record is destroyed, and the process survives the catch, 15s in foreground, and 90s in background with zero FATALs/ANRs. On the next app resume the task re-registers and the foreground service comes up (`isForeground=true`, `foregroundServiceType` location, notification shown).
- Happy path unchanged: with no simulated rejection, starting tracking brings the FGS up on the first attempt via `startService` + `startForeground` (`types=0x00000008`), and it survives background/foreground cycles.
- Real-timing repro attempts (start tracking + immediately background/lock at several delays): no crash.

## Checklist

- CHANGELOG entry added under expo-location "Unpublished > Bug fixes".

### DIFF
--- a/packages/expo-location/CHANGELOG.md
+++ b/packages/expo-location/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ### 🐛 Bug fixes
 
+- [Android] Survive the foreground-service eligibility rejection that Android 14+ can raise inside `onServiceConnected` instead of crashing, and retry the foreground service on the next task registration. ([#TBD](https://github.com/expo/expo/pull/TBD) by [@steve-rodri](https://github.com/steve-rodri))
 - [iOS] Add `scope` and `accuracy` under `ios` to the responses from `getBackgroundPermissionsAsync` and `requestBackgroundPermissionsAsync`, matching the `LocationPermissionResponse` type. ([#48926](https://github.com/expo/expo/pull/48926) by [@vonovak](https://github.com/vonovak))
 - [Android] Fix `timeInterval` and `distanceInterval` being ignored for background location updates. ([#46788](https://github.com/expo/expo/issues/46788) by [@doshisunny](https://github.com/doshisunny))
 - [iOS] Fix incorrect default value for `pausesUpdatesAutomatically` to match docs. ([#47008](https://github.com/expo/expo/pull/47008) by [@Ignigena](https://github.com/Ignigena))
--- a/packages/expo-location/android/src/main/java/expo/modules/location/taskConsumers/LocationTaskConsumer.kt
+++ b/packages/expo-location/android/src/main/java/expo/modules/location/taskConsumers/LocationTaskConsumer.kt
@@ -211,7 +211,26 @@ class LocationTaskConsumer(context: Context, taskManagerUtils: TaskManagerUtilsI
       extras.putString("taskName", task.name)
       extras.putBoolean("killService", serviceOptions.getBoolean("killServiceOnDestroy", false))
       serviceIntent.putExtras(extras)
-      context.startForegroundService(serviceIntent)
+      try {
+        // startService, not startForegroundService: startForegroundService
+        // attaches a promise that the service WILL reach startForeground, and
+        // Android 14+ kills the app with ForegroundServiceDidNotStartInTime-
+        // Exception whenever that promise breaks - including when the
+        // eligibility rejection below makes keeping it impossible, and even
+        // when the service is torn down cleanly. The isForegrounded check
+        // above makes startService legal here, and the startForeground call
+        // in onServiceConnected promotes the service all the same.
+        context.startService(serviceIntent)
+      } catch (e: IllegalStateException) {
+        // BackgroundServiceStartNotAllowedException (API 31+): the app left
+        // the foreground after the check above. Skip binding; the next task
+        // (re)registration retries the foreground service.
+        Log.e(TAG, "Location service start not allowed; skipping the foreground service.", e)
+        return
+      } catch (e: SecurityException) {
+        Log.e(TAG, "Location service start rejected; skipping the foreground service.", e)
+        return
+      }
       context.bindService(
         serviceIntent,
         object : ServiceConnection {
@@ -219,7 +238,13 @@ class LocationTaskConsumer(context: Context, taskManagerUtils: TaskManagerUtilsI
             mService = (service as? ServiceBinder)?.service
             mService?.let {
               it.setParentContext(context)
-              it.startForeground(serviceOptions)
+              try {
+                it.startForeground(serviceOptions)
+              } catch (e: SecurityException) {
+                abortForegroundServiceStart(context, this, e)
+              } catch (e: IllegalStateException) {
+                abortForegroundServiceStart(context, this, e)
+              }
             }
           }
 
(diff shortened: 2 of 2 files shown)


Wenszel:
@expo-bot review

PR #49428, opened 2026-08-27
[calendar][android] Expose extended properties on events

MoOx:
# Why

I am building an app that writes and manages its own events inside a user's calendar, so it has to know which events are its own. Without a per-event marker there is no safe way to tell: the app either risks deleting events the user created, or has to keep a side database mapping its ids to calendar ids, which drifts as soon as the calendar is edited elsewhere, does not survive a reinstall, and cannot be shared across devices.

iOS has `url` for exactly this, documented as iOS-only. Android has `CalendarContract.ExtendedProperties`, the platform mechanism meant for it, which `expo-calendar` does not expose. So the guard "never delete an event I did not create" can be written on one platform and not on the other.

## The one thing to agree on first

`CalendarProvider2` refuses writes to that table from ordinary callers:

    IllegalArgumentException: Only sync adapters may write using
    content://com.android.calendar/extendedproperties

The way through is `?caller_is_syncadapter=true&account_name=...&account_type=...`. Four things make that a smaller step than it sounds:

- `CalendarContract.CALLER_IS_SYNCADAPTER` is public, documented API. No permission gates it, and it needs no manifest entry, no `SyncAdapter` service and no account authenticator. It is a per-request write mode, not a status the app registers with the system.
- **`expo-calendar` already uses it**, in both implementations, to create calendars: `domain/calendar/CalendarRepository.kt:52` and `next/domain/repositories/calendar/CalendarRepository.kt:44`. This PR does not introduce the practice, it reuses it for a second table.
- The scope stays that one table. Events keep going through the plain URI, so the two provider behaviours that matter for them are untouched: a deletion still leaves the tombstone a sync adapter needs to propagate it, and `dirty` is still set on every write.
- What the provider does not do is verify the claim. It never checks that the caller really is that account's sync adapter. That is the part which is a convention being leaned on rather than an API used as intended, and it is why this deserves an explicit yes rather than being assumed.

**So the question is narrow: you already declare `expo-calendar` a sync adapter to create calendars, is there a reason for a key/value write on an event to be treated differently?**

If extended properties should not go through that door, I am curious to know more about the reason. Everything below follows from that decision.

# How

Measured on a physical device (Huawei P30, EMUI, real Google account) rather than assumed. Three findings shaped the API.

**1. Only prefixed names survive a server round trip.** Three names were written to one event, pushed to Google, then the local database was wiped (`pm clear com.android.providers.calendar`) and rebuilt from Google's servers:

| Name written | Survived a full server round trip | |---|---|
| `x-hide-my-cal` | **no** |
| `private:x-hide-my-cal` | **yes** |
| `shared:x-hide-my-cal` | **yes** |

The prefixes map onto `extendedProperties.private` and `extendedProperties.shared` in the Google Calendar API. Unprefixed names are dropped by the sync adapter — silently, after the local write has already reported success.

Corroborating evidence from the same wipe: EMUI's own `private:categories = "Come From :com.android.shell"` came back too. Its value encodes how the event was created, which no local process could reconstruct after a wipe, so it came from the server.

**2. The sync-adapter write marks the row as already synced.** The event stays at `dirty=0` and nothing is ever pushed. A normal write to any column afterwards sets `dirty=1`, and the properties travel with the next sync (confirmed: back to `dirty=0` within 15 seconds).

**3. Reads need none of this.** Ordinary callers may read the table.

So the API deliberately is not a thin wrapper over the table. A wrapper would let callers write an unprefixed name, read it back locally, watch it work, and lose it at the first sync without ever seeing an error. Instead:

- writes go through the sync adapter URI, with the account resolved from the event's calendar, handled internally;
- the event is marked dirty after every write and every delete, so the change actually leaves the device;
- unprefixed names are rejected on calendars owned by an account, with an error explaining the prefixes, and accepted on local-account calendars, which no adapter syncs.

The change follows the layering already in `next`: `ExtendedPropertyEntity` / `ExtendedPropertyInput` / `ExtendedPropertyRepository` (a separate table, so its own URI and lifecycle, not a column on Events) → `ExtendedPropertyRecord` / `ExtendedPropertyMapper` → `ExpoCalendarEvent` → TypeScript. `EventRepository.markDirty()` is the one addition to an existing repository.

Public surface, Android only:

```ts
await event.setExtendedProperty('private:my-app-id', 'a4f1c2');
await event.getExtendedProperties(); // [{ name: 'private:my-app-id', value: 'a4f1c2' }]
await event.deleteExtendedProperty('private:my-app-id');
```

On other platforms the three methods throw `UnavailabilityError`, following the existing pattern for the iOS-only reminder APIs.

## Open questions

- **The sync-adapter declaration above.** Everything else is downstream of it.
- **Prefix enforcement.** Rejecting unprefixed names on synced calendars trades a compile-free footgun for a runtime error. The alternative (accept anything and document the trap) is one comment away.
- **Recurring events.** `setExtendedProperty` on an occurrence writes to the parent event, since extended properties belong to the event row. iOS `url` behaves the same way, so this does not change it, but it is worth naming.

# Test Plan

Unit tests, matching the conventions of `EventRepositoryTest` (mockk + Robolectric + `MatrixCursor`):

- `ExtendedPropertyRepositoryTest`: reads go through the plain URI and select on `EVENT_ID`; writes carry the sync adapter parameters; an existing property is updated in place through its row URI rather than duplicated; a missing one is not deleted; `SecurityException` surfaces as `PermissionException`.
- `EventRepositoryTest`: `markDirty` rewrites the start date through the plain URI, writes exactly one column, and no-ops on an event that is gone.
- `ExtendedPropertyNameTest`, `ExtendedPropertyMapperTest`: the prefix rule and the mapping.
- `Calendar-test.native.ts`: the methods resolve on Android and throw `UnavailabilityError` elsewhere.

Run with `et check-packages expo-calendar` and `et native-unit-tests -p android --packages expo-calendar`.

Device check to reproduce the findings above (around half a day, physical device with a real Google account; an emulator cannot answer this):

1. `setExtendedProperty('private:x-test', 'v1')` on an event of a Google calendar.
2. Force a sync, confirm the event is visible on calendar.google.com.
3. Settings → Apps → show system apps → Calendar Storage → Clear data. This wipes the local database without removing the account, forcing a full resync from the server.
4. `getExtendedProperties()` on the event once it comes back. Note the trap: the resync rebuilds the database, so the event returns under a different row id. Querying the pre-wipe id reports a false negative.
5. Repeat with an unprefixed name: it is now rejected upfront, which is the behaviour this change is arguing for.

# Checklist

- [x] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: updated a module plugin). Note: n/a, no plugin change

### DIFF
--- a/docs/pages/versions/unversioned/sdk/calendar.mdx
+++ b/docs/pages/versions/unversioned/sdk/calendar.mdx
@@ -151,11 +151,13 @@ const styles = StyleSheet.create({
 });
 ```
 
-## Storing your own metadata on an event
+## Storing your app's metadata with extended properties
 
-An event has no free-form field your app can write to on Android. To recognize the events your app created, and avoid deleting anyone else's, attach an extended property to them:
+On Android, an event has no field reserved for your app's own data. To recognize the events your app created, and avoid deleting anyone else's, attach an extended property to them:
 
 ```ts
+const [calendar] = await Calendar.getCalendars(Calendar.EntityTypes.EVENT);
+
 const event = await calendar.createEvent({
   title: 'Team sync',
   startDate: new Date(),
@@ -168,12 +170,12 @@ const properties = await event.getExtendedProperties();
 // [{ name: 'private:my-app-id', value: 'a4f1c2' }]
 ```
 
-These methods are backed by `CalendarContract.ExtendedProperties`, which comes with two constraints:
+These methods read and write `CalendarContract.ExtendedProperties`, which comes with two constraints:
 
-- On a calendar owned by an account, the name has to start with `private:`, which keeps the value readable by that account, or `shared:`, which shares it with the guests of the event. The sync adapter drops any other name on the next sync, so `setExtendedProperty` rejects it instead of letting the value disappear later. Calendars of the local account are never synced and accept any name.
-- The event has to belong to a calendar that has an account, since the calendar provider accepts writes to that table only from the sync adapter of that account.
+- On a calendar owned by an account, the name has to start with `private:` or `shared:`. The `private:` prefix keeps the value readable by that account, and the `shared:` prefix shares it with the guests of the event. Some sync adapters, including Google Calendar's, drop any other name on the next sync, so `setExtendedProperty` rejects it instead of letting the value disappear later. Calendars of the local account are never synced and accept any name.
+- `setExtendedProperty` and `deleteExtendedProperty` need the event to belong to a calendar that has an account. The calendar provider accepts writes to that table only from the sync adapter of that account. Reading with `getExtendedProperties` has no account requirement.
 
-> **Note:** Extended properties are Android-only. On iOS, write to the `url` property of the event instead.
+> **info** Extended properties are Android-only. On iOS, write to the `url` property of the event instead.
 
 ## API
 
--- a/docs/pages/versions/unversioned/sdk/calendar.mdx
+++ b/docs/pages/versions/unversioned/sdk/calendar.mdx
@@ -172,7 +172,7 @@ const properties = await event.getExtendedProperties();
 
 These methods read and write `CalendarContract.ExtendedProperties`, which comes with two constraints:
 
-- On a calendar owned by an account, the name has to start with `private:` or `shared:`. The `private:` prefix keeps the value readable by that account, and the `shared:` prefix shares it with the guests of the event. Some sync adapters, including Google Calendar's, drop any other name on the next sync, so `setExtendedProperty` rejects it instead of letting the value disappear later. Calendars of the local account are never synced and accept any name.
+- On a calendar owned by a Google account, the name has to start with `private:` or `shared:`. The `private:` prefix keeps the value readable by that account, and the `shared:` prefix shares it with the guests of the event. Google Calendar's sync adapter drops any other name on the next sync, so `setExtendedProperty` rejects it instead of letting the value disappear later. This convention belongs to that adapter rather than to the calendar provider, so calendars of any other account accept any name.
 - `setExtendedProperty` and `deleteExtendedProperty` need the event to belong to a calendar that has an account. The calendar provider accepts writes to that table only from the sync adapter of that account. Reading with `getExtendedProperties` has no account requirement.
 
 > **info** Extended properties are Android-only. On iOS, write to the `url` property of the event instead.
--- a/docs/pages/versions/unversioned/sdk/calendar.mdx
+++ b/docs/pages/versions/unversioned/sdk/calendar.mdx
@@ -151,6 +151,30 @@ const styles = StyleSheet.create({
 });
 ```
 
+## Storing your own metadata on an event
+
+An event has no free-form field your app can write to on Android. To recognize the events your app created, and avoid deleting anyone else's, attach an extended property to them:
+
+```ts
+const event = await calendar.createEvent({
+  title: 'Team sync',
+  startDate: new Date(),
+  endDate: new Date(),
+});
+
+await event.setExtendedProperty('private:my-app-id', 'a4f1c2');
+
+const properties = await event.getExtendedProperties();
+// [{ name: 'private:my-app-id', value: 'a4f1c2' }]
+```
+
+These methods are backed by `CalendarContract.ExtendedProperties`, which comes with two constraints:
+
+- On a calendar owned by an account, the name has to start with `private:`, which keeps the value readable by that account, or `shared:`, which shares it with the guests of the event. The sync adapter drops any other name on the next sync, so `setExtendedProperty` rejects it instead of letting the value disappear later. Calendars of the local account are never synced and accept any name.
+- The event has to belong to a calendar that has an account, since the calendar provider accepts writes to that table only from the sync adapter of that account.
+
+> **Note:** Extended properties are Android-only. On iOS, write to the `url` property of the event instead.
+
 ## API
 
 ```js
--- a/packages/expo-calendar/CHANGELOG.md
+++ b/packages/expo-calendar/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### 🎉 New features
 
+- [android] Add `getExtendedProperties()`, `setExtendedProperty()` and `deleteExtendedProperty()` to `ExpoCalendarEvent`, so that apps can recognize the events they created: Android has no `url` field to mark them with, and `CalendarContract.ExtendedProperties` — the mechanism meant for it — was not exposed. ([#PR_NUMBER](https://github.com/expo/expo/pull/PR_NUMBER) by [@MoOx](https://github.com/MoOx))
+
 ### 🐛 Bug fixes
 
 - [ios] Fix a crash when reading `calendarId` on an event or reminder whose `EKCalendarItem.calendar` is nil. The property is `null_unspecified` in the EventKit headers, so the bare access was an implicit force-unwrap that trapped the JS thread. ([#48445](https://github.com/expo/expo/pull/48445) by [@cvburgess](https://github.com/cvburgess))
--- a/packages/expo-calendar/CHANGELOG.md
+++ b/packages/expo-calendar/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 ### 🎉 New features
 
-- [android] Add `getExtendedProperties()`, `setExtendedProperty()` and `deleteExtendedProperty()` to `ExpoCalendarEvent`, so that apps can recognize the events they created: Android has no `url` field to mark them with, and `CalendarContract.ExtendedProperties` — the mechanism meant for it — was not exposed. ([#PR_NUMBER](https://github.com/expo/expo/pull/PR_NUMBER) by [@MoOx](https://github.com/MoOx))
+- [android] Add `getExtendedProperties()`, `setExtendedProperty()` and `deleteExtendedProperty()` to `ExpoCalendarEvent`, so that apps can recognize the events they created: Android has no `url` field to mark them with, and `CalendarContract.ExtendedProperties` — the mechanism meant for it — was not exposed. ([#49428](https://github.com/expo/expo/pull/49428) by [@MoOx](https://github.com/MoOx))
 
 ### 🐛 Bug fixes
 
--- a/packages/expo-calendar/android/src/main/java/expo/modules/calendar/next/ExpoCalendarEvent.kt
+++ b/packages/expo-calendar/android/src/main/java/expo/modules/calendar/next/ExpoCalendarEvent.kt
@@ -161,10 +161,11 @@ class ExpoCalendarEvent(
       return
     }
     throw ExtendedPropertyNameNotSyncSafeException(
-      "Extended property name `$name` has no visibility prefix, and this event belongs to the `${account.type}` account `${account.name}`. " +
-        "The value would be stored on the device and dropped by the sync adapter on the next sync, without an error, " +
-        "because only names prefixed `${ExtendedPropertyName.PRIVATE_PREFIX}` or `${ExtendedPropertyName.SHARED_PREFIX}` survive a server round trip. " +
-        "Use `${ExtendedPropertyName.PRIVATE_PREFIX}$name` to keep the value readable by this account only, " +
+      "Extended property name `$name` has no visibility prefix, and this event belongs to an account of type `${account.type}`. " +
+        "Sync adapters such as Google Calendar's keep only names prefixed `${ExtendedPropertyName.PRIVATE_PREFIX}` or " +
+        "`${ExtendedPropertyName.SHARED_PREFIX}`, and drop the rest on the next sync without reporting an error, " +
+        "so the value would be stored on the device and disappear later. " +
+        "Use `${ExtendedPropertyName.PRIVATE_PREFIX}$name` to keep the value readable by the owning account only, " +
         "or `${ExtendedPropertyName.SHARED_PREFIX}$name` to share it with the guests of the event."
     )
   }
--- a/packages/expo-calendar/android/src/main/java/expo/modules/calendar/next/ExpoCalendarEvent.kt
+++ b/packages/expo-calendar/android/src/main/java/expo/modules/calendar/next/ExpoCalendarEvent.kt
@@ -1,6 +1,5 @@
 package expo.modules.calendar.next
 
-import android.provider.CalendarContract
 import expo.modules.calendar.next.utils.DateTimeInput
 import expo.modules.calendar.next.utils.getTimeInMillis
 import expo.modules.calendar.next.domain.dto.event.EventExceptionInput
@@ -154,16 +153,20 @@ class ExpoCalendarEvent(
   }
 
   /**
-   * Rejects names a sync adapter would drop, on the calendars where one is at work.
+   * Rejects names Google Calendar's sync adapter would drop.
+   *
+   * The prefix convention belongs to that adapter, not to the calendar provider, so it is enforced
+   * on Google accounts only. Another adapter may keep an unprefixed name, and refusing the write
+   * here would deny something that works.
    */
   private fun requireSyncSafeName(name: String, account: CalendarAccount) {
-    if (account.type == CalendarContract.ACCOUNT_TYPE_LOCAL || ExtendedPropertyName.isSyncSafe(name)) {
+    if (!account.isGoogle || ExtendedPropertyName.isSyncSafe(name)) {
       return
     }
     throw ExtendedPropertyNameNotSyncSafeException(
-      "Extended property name `$name` has no visibility prefix, and this event belongs to an account of type `${account.type}`. " +
-        "Sync adapters such as Google Calendar's keep only names prefixed `${ExtendedPropertyName.PRIVATE_PREFIX}` or " +
-        "`${ExtendedPropertyName.SHARED_PREFIX}`, and drop the rest on the next sync without reporting an error, " +
+      "Extended property name `$name` has no visibility prefix, and this event belongs to a Google account. " +
+        "Google Calendar's sync adapter keeps only names prefixed `${ExtendedPropertyName.PRIVATE_PREFIX}` or " +
+        "`${ExtendedPropertyName.SHARED_PREFIX}`, and drops the rest on the next sync without reporting an error, " +
         "so the value would be stored on the device and disappear later. " +
         "Use `${ExtendedPropertyName.PRIVATE_PREFIX}$name` to keep the value readable by the owning account only, " +
         "or `${ExtendedPropertyName.SHARED_PREFIX}$name` to share it with the guests of the event."
--- a/packages/expo-calendar/android/src/main/java/expo/modules/calendar/next/ExpoCalendarEvent.kt
+++ b/packages/expo-calendar/android/src/main/java/expo/modules/calendar/next/ExpoCalendarEvent.kt
@@ -111,7 +111,7 @@ class ExpoCalendarEvent(
 
   suspend fun getExtendedProperties(): List<ExtendedPropertyRecord> =
     extendedPropertyRepository.findAllByEventId(eventId)
-      .map { extendedPropertyMapper.toRecord(it) }
+      .mapNotNull { extendedPropertyMapper.toRecord(it) }
 
   suspend fun setExtendedProperty(name: String, value: String) {
     val account = requireOwningAccount()
--- a/packages/expo-calendar/android/src/main/java/expo/modules/calendar/next/ExpoCalendarEvent.kt
+++ b/packages/expo-calendar/android/src/main/java/expo/modules/calendar/next/ExpoCalendarEvent.kt
@@ -1,22 +1,33 @@
 package expo.modules.calendar.next
 
+import android.provider.CalendarContract
 import expo.modules.calendar.next.utils.DateTimeInput
 import expo.modules.calendar.next.utils.getTimeInMillis
 import expo.modules.calendar.next.domain.dto.event.EventExceptionInput
 import kotlin.time.Duration.Companion.milliseconds
+import expo.modules.calendar.next.domain.model.calendar.CalendarAccount
+import expo.modules.calendar.next.domain.model.extendedproperty.ExtendedPropertyName
 import expo.modules.calendar.next.domain.repositories.attendee.AttendeeRepository
+import expo.modules.calendar.next.domain.repositories.calendar.CalendarRepository
 import expo.modules.calendar.next.domain.repositories.event.EventRepository
+import expo.modules.calendar.next.domain.repositories.extendedproperty.ExtendedPropertyRepository
 import expo.modules.calendar.next.domain.repositories.instance.InstanceRepository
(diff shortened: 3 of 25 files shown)


Wenszel:
@expo-bot review

docs/pages/versions/unversioned/sdk/calendar.mdx
```diff
@@ -151,6 +151,30 @@ const styles = StyleSheet.create({
...
+const properties = await event.getExtendedProperties();
+// [{ name: 'private:my-app-id', value: 'a4f1c2' }]
+```
+
+These methods are backed by `CalendarContract.ExtendedProperties`, which comes with two constraints:
+
+- On a calendar owned by an account, the name has to start with `private:`, which keeps the value readable by that account, or `shared:`, which shares it with the guests of the event. The sync adapter drops any other name on the next sync, so `setExtendedProperty` rejects it instead of letting the value disappear later. Calendars of the local account are never synced and accept any name.
+- The event has to belong to a calendar that has an account, since the calendar provider accepts writes to that table only from the sync adapter of that account.
+
+> **Note:** Extended properties are Android-only. On iOS, write to the `url` property of the event instead.
```
amandeepmittal: ⚠️ **correctness** `mdx-components.md#callouts`

`**Note:**` is not a recognized callout keyword, so this renders as a plain blockquote with no icon. Use one of the recognized types, such as `**info**`.

```suggestion
> **info** Extended properties are Android-only. On iOS, write to the `url` property of the event instead.
```

docs/pages/versions/unversioned/sdk/calendar.mdx
```diff
@@ -151,6 +151,30 @@ const styles = StyleSheet.create({
...
+await event.setExtendedProperty('private:my-app-id', 'a4f1c2');
+
+const properties = await event.getExtendedProperties();
+// [{ name: 'private:my-app-id', value: 'a4f1c2' }]
+```
+
+These methods are backed by `CalendarContract.ExtendedProperties`, which comes with two constraints:
+
+- On a calendar owned by an account, the name has to start with `private:`, which keeps the value readable by that account, or `shared:`, which shares it with the guests of the event. The sync adapter drops any other name on the next sync, so `setExtendedProperty` rejects it instead of letting the value disappear later. Calendars of the local account are never synced and accept any name.
+- The event has to belong to a calendar that has an account, since the calendar provider accepts writes to that table only from the sync adapter of that account.
```
amandeepmittal: ⚠️ **correctness** `review-ai-text.md#t7-overclaiming`

Only the write methods resolve the owning account, so "The event has to belong to a calendar that has an account" reads as if it also covered `getExtendedProperties`, which reads through the plain URI without one. Scope the constraint to `setExtendedProperty` and `deleteExtendedProperty`.

---

✏️ **style** `mechanics.md#vale-sentencelength`

This sentence runs 29 words. Split the constraint from its reason, for example: "The event has to belong to a calendar that has an account. The calendar provider accepts writes to that table only from the sync adapter of that account."

```suggestion
- `setExtendedProperty` and `deleteExtendedProperty` need the event to belong to a calendar that has an account. The calendar provider accepts writes to that table only from the sync adapter of that account. Reading with `getExtendedProperties` has no account requirement.
```

docs/pages/versions/unversioned/sdk/calendar.mdx
```diff
@@ -151,6 +151,30 @@ const styles = StyleSheet.create({
 });
 ```
 
+## Storing your own metadata on an event
```
amandeepmittal: 💡 **nit** `links-and-headings.md#titles-and-headings`

Consider working the feature name into the heading, for example "Storing your app's metadata with extended properties", so search and the on-page outline surface `extended properties` next to the API it documents.

```suggestion
## Storing your app's metadata with extended properties
```

docs/pages/versions/unversioned/sdk/calendar.mdx
```diff
@@ -151,6 +151,30 @@ const styles = StyleSheet.create({
...
+  endDate: new Date(),
+});
+
+await event.setExtendedProperty('private:my-app-id', 'a4f1c2');
+
+const properties = await event.getExtendedProperties();
+// [{ name: 'private:my-app-id', value: 'a4f1c2' }]
+```
+
+These methods are backed by `CalendarContract.ExtendedProperties`, which comes with two constraints:
```
amandeepmittal: ✏️ **style** `prose.md#voice-and-tone`

"are backed by" is passive voice. Active reads tighter and stays accurate for both the read and the write methods: "These methods read and write `CalendarContract.ExtendedProperties`".

```suggestion
These methods read and write `CalendarContract.ExtendedProperties`, which comes with two constraints:
```

docs/pages/versions/unversioned/sdk/calendar.mdx
```diff
@@ -151,6 +151,30 @@ const styles = StyleSheet.create({
 });
 ```
 
+## Storing your own metadata on an event
+
+An event has no free-form field your app can write to on Android. To recognize the events your app created, and avoid deleting anyone else's, attach an extended property to them:
```
amandeepmittal: ⚠️ **correctness** `review-ai-text.md#t7-overclaiming`

`notes` is a free-form field an app can write on Android, so "An event has no free-form field your app can write to on Android" is not quite true. The real gap is a field reserved for the app's own data.

```suggestion
On Android, an event has no field reserved for your app's own data. To recognize the events your app created, and avoid deleting anyone else's, attach an extended property to them:
```

docs/pages/versions/unversioned/sdk/calendar.mdx
```diff
@@ -151,6 +151,30 @@ const styles = StyleSheet.create({
...
+
+await event.setExtendedProperty('private:my-app-id', 'a4f1c2');
+
+const properties = await event.getExtendedProperties();
+// [{ name: 'private:my-app-id', value: 'a4f1c2' }]
+```
+
+These methods are backed by `CalendarContract.ExtendedProperties`, which comes with two constraints:
+
+- On a calendar owned by an account, the name has to start with `private:`, which keeps the value readable by that account, or `shared:`, which shares it with the guests of the event. The sync adapter drops any other name on the next sync, so `setExtendedProperty` rejects it instead of letting the value disappear later. Calendars of the local account are never synced and accept any name.
```
amandeepmittal: ✏️ **style** `mechanics.md#vale-sentencelength`

The first sentence runs 33 words: "On a calendar owned by an account, the name has to start with `private:`, which keeps the value readable by that account, or `shared:`, which shares it with the guests of the event." Split the rule from what each prefix does.

```suggestion
- On a calendar owned by an account, the name has to start with `private:` or `shared:`. The `private:` prefix keeps the value readable by that account, and the `shared:` prefix shares it with the guests of the event. The sync adapter drops any other name on the next sync, so `setExtendedProperty` rejects it instead of letting the value disappear later. Calendars of the local account are never synced and accept any name.
```

packages/expo-calendar/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 ### 🎉 New features
 
+- [android] Add `getExtendedProperties()`, `setExtendedProperty()` and `deleteExtendedProperty()` to `ExpoCalendarEvent`, so that apps can recognize the events they created: Android has no `url` field to mark them with, and `CalendarContract.ExtendedProperties` — the mechanism meant for it — was not exposed. ([#PR_NUMBER](https://github.com/expo/expo/pull/PR_NUMBER) by [@MoOx](https://github.com/MoOx))
```
amandeepmittal: ⚠️ **correctness** `contributing#updating-changelogs`

The PR number placeholder never got replaced: "[#PR_NUMBER](https://github.com/expo/expo/pull/PR_NUMBER)". Fill in 49428 for both the link text and the URL.

```suggestion
- [android] Add `getExtendedProperties()`, `setExtendedProperty()` and `deleteExtendedProperty()` to `ExpoCalendarEvent`, so that apps can recognize the events they created: Android has no `url` field to mark them with, and `CalendarContract.ExtendedProperties` — the mechanism meant for it — was not exposed. ([#49428](https://github.com/expo/expo/pull/49428) by [@MoOx](https://github.com/MoOx))
```

packages/expo-calendar/src/ExpoCalendar.types.ts
```diff
@@ -397,6 +418,49 @@ export declare class ExpoCalendarEvent {
...
+
+  /**
+   * Attaches a key/value pair to this event, replacing the value already stored under the same name.
+   *
+   * The property is written as the sync adapter of the account owning the event, because the
+   * calendar provider accepts writes to that table from no one else. Writing that way records the
+   * row as already synced, so the event is flagged as locally modified afterwards, and that flag
+   * is what carries the property to the server on the next sync.
+   *
+   * > **Note:** on an event of a calendar owned by an account, `name` has to start with `private:`
```
amandeepmittal: ✏️ **style** `expo-api-docs#blockquote-notes`

Start the note's sentence with a capital letter: "On an event of a calendar owned by an account".

```suggestion
   * > **Note:** On an event of a calendar owned by an account, `name` has to start with `private:`
```

amandeepmittal (changes requested):

MoOx:
@amandeepmittal I have addressed all the issues reported by you or the bots. Should be good :)

amandeepmittal (approved):
Docs changes look good to me, thanks for adding them!

packages/expo-calendar/android/src/main/java/expo/modules/calendar/next/domain/repositories/event/EventRepository.kt
```diff
@@ -56,6 +57,35 @@ class EventRepository(private val contentResolver: ContentResolver) {
...
+        ?.getOptionalLong(CalendarContract.Events.DTSTART)
+    } ?: return@withContext false
+
+    contentResolver.safeUpdate(
+      uri = uri,
+      values = ContentValues().apply {
+        put(CalendarContract.Events.DTSTART, dtStart)
+      }
+    ) > 0
+  }
```
Wenszel: This approach with modifying `DTSTART` works, but requires two separate queries to the provider. The update with an empty `contentValues` should work just as well: [ContentProvider2.java#4123](https://android.googlesource.com/platform/packages/providers/CalendarProvider/+/38e4d4533220e6e18d314dfb1d777b34c4d45797/src/com/android/providers/calendar/CalendarProvider2.java#4123)
```suggestion
  suspend fun markDirty(id: EventId): Boolean = withContext(Dispatchers.IO) {
  contentResolver.safeUpdate(
    uri = ContentUris.withAppendedId(CalendarContract.Events.CONTENT_URI, id.value),
    values = ContentValues()
  ) > 0
}

```

packages/expo-calendar/android/src/main/java/expo/modules/calendar/next/domain/repositories/extendedproperty/ExtendedPropertyCursorExtensions.kt
```diff
@@ -0,0 +1,19 @@
...
+
+fun Cursor.toExtendedPropertyEntity(eventId: EventId) = ExtendedPropertyEntity(
+  id = ExtendedPropertyId(
+    getOptionalLong(CalendarContract.ExtendedProperties._ID)
+      ?: throw IllegalStateException("extended property ID must not be null")
+  ),
+  eventId = eventId,
+  name = getRequiredString(CalendarContract.ExtendedProperties.NAME),
+  value = getRequiredString(CalendarContract.ExtendedProperties.VALUE)
+)
```
Wenszel: Name and value fields should not  be null, but there is no database constraint that ensures it. The entities should hold exactly data saved in the database, even malformed one.  
```suggestion
fun Cursor.toExtendedPropertyEntity(eventId: EventId) = ExtendedPropertyEntity(
  id = ExtendedPropertyId(
    getOptionalLong(CalendarContract.ExtendedProperties._ID)
      ?: throw IllegalStateException("extended property ID must not be null")
  ),
  eventId = eventId,
  name = getOptionalString(CalendarContract.ExtendedProperties.NAME),
  value = getOptionalString(CalendarContract.ExtendedProperties.VALUE)
)
```

packages/expo-calendar/android/src/main/java/expo/modules/calendar/next/domain/repositories/extendedproperty/ExtendedPropertyRepository.kt
```diff
@@ -0,0 +1,134 @@
...
+
+    val uri = contentResolver.safeInsert(
+      uri = CalendarContract.ExtendedProperties.CONTENT_URI.asSyncAdapter(account),
+      values = input.toContentValues(eventId)
+    )
+    ExtendedPropertyId(
+      uri.lastPathSegment?.toLongOrNull()
+        ?: throw IllegalStateException("Couldn't decode extended property ID from inserted content URI")
+    )
+  }
```
Wenszel: `contentResolver.applyBatch` would be a better fit here. It performs all the operations within one transaction, therefore it solves the race condition problem which required the second insert.  

```suggestion
  suspend fun upsert(
    eventId: EventId,
    account: CalendarAccount,
    input: ExtendedPropertyInput
  ): ExtendedPropertyId = withContext(Dispatchers.IO) {
    val syncAdapterUri = CalendarContract.ExtendedProperties.CONTENT_URI.asSyncAdapter(account)
    val results = contentResolver.safeApplyBatch(
      arrayListOf(
        ContentProviderOperation.newDelete(syncAdapterUri)
          .withSelection(
            "${CalendarContract.ExtendedProperties.EVENT_ID} = ? AND ${CalendarContract.ExtendedProperties.NAME} = ?",
            arrayOf(eventId.value.toString(), input.name)
          )
          .build(),
        ContentProviderOperation.newInsert(syncAdapterUri)
          .withValues(input.toContentValues(eventId))
          .build()
      )
    )
    ExtendedPropertyId(
      results.last().uri?.lastPathSegment?.toLongOrNull()
        ?: throw IllegalStateException("Couldn't decode extended property ID from inserted content URI")
    )
  }

```

```
suspend fun ContentResolver.safeApplyBatch(
  operations: ArrayList<ContentProviderOperation>
): Array<ContentProviderResult> = withContext(Dispatchers.IO) {
  try {
    applyBatch(CalendarContract.AUTHORITY, operations)
  } catch (e: SecurityException) {
    throw PermissionException(Manifest.permission.WRITE_CALENDAR, e)
  } catch (e: OperationApplicationException) {
    throw CouldNotExecuteQueryException("Couldn't apply the batch of calendar operations", e)
  } catch (e: RemoteException) {
    throw CouldNotExecuteQueryException("Couldn't reach the calendar provider to apply a batch", e)
  }
}

```

packages/expo-calendar/android/src/main/java/expo/modules/calendar/next/ExpoCalendarEvent.kt
```diff
@@ -96,17 +110,80 @@ class ExpoCalendarEvent(
...
+    throw ExtendedPropertyNameNotSyncSafeException(
+      "Extended property name `$name` has no visibility prefix, and this event belongs to an account of type `${account.type}`. " +
+        "Sync adapters such as Google Calendar's keep only names prefixed `${ExtendedPropertyName.PRIVATE_PREFIX}` or " +
+        "`${ExtendedPropertyName.SHARED_PREFIX}`, and drop the rest on the next sync without reporting an error, " +
+        "so the value would be stored on the device and disappear later. " +
+        "Use `${ExtendedPropertyName.PRIVATE_PREFIX}$name` to keep the value readable by the owning account only, " +
+        "or `${ExtendedPropertyName.SHARED_PREFIX}$name` to share it with the guests of the event."
+    )
+  }
+
```
Wenszel, line 170: As far as I can tell this is Google Calendar's sync adapter convention, not a platform rule - it does not appear in the calendar provider source code. Could we narrow it to Google accounts for now? Other platforms sync adapters might preserve unprefixed names. It should be narrowed in the docs too.

packages/expo-calendar/android/src/main/java/expo/modules/calendar/next/domain/repositories/event/EventRepository.kt
Wenszel: This approach with modifying `DTSTART` works, but it requires modifying a value we don't actually want to change and could cause problems if another app modifies the same event at the same time. An update with an empty `contentValues` should work just as well: [CalendarProvider2.java#4123](https://android.googlesource.com/platform/packages/providers/CalendarProvider/+/38e4d4533220e6e18d314dfb1d777b34c4d45797/src/com/android/providers/calendar/CalendarProvider2.java#4123)
```suggestion
  suspend fun markDirty(id: EventId): Boolean = withContext(Dispatchers.IO) {
  contentResolver.safeUpdate(
    uri = ContentUris.withAppendedId(CalendarContract.Events.CONTENT_URI, id.value),
    values = ContentValues()
  ) > 0
}

```

packages/expo-calendar/android/src/main/java/expo/modules/calendar/next/domain/repositories/extendedproperty/ExtendedPropertyCursorExtensions.kt
Wenszel, line 19: Name and value fields should not be null, but there is no database constraint that ensures this. The entities should hold exactly the data saved in the database, even if it's malformed.
```suggestion
fun Cursor.toExtendedPropertyEntity(eventId: EventId) = ExtendedPropertyEntity(
  id = ExtendedPropertyId(
    getOptionalLong(CalendarContract.ExtendedProperties._ID)
      ?: throw IllegalStateException("extended property ID must not be null")
  ),
  eventId = eventId,
  name = getOptionalString(CalendarContract.ExtendedProperties.NAME),
  value = getOptionalString(CalendarContract.ExtendedProperties.VALUE)
)
```

packages/expo-calendar/android/src/main/java/expo/modules/calendar/next/domain/repositories/extendedproperty/ExtendedPropertyRepository.kt
Wenszel, line 70: `contentResolver.applyBatch` would be a better fit here. It performs all the operations within a single transaction, so it also solves the race condition that required the second insert.
I’m attaching the safeApplyBatch code below.

```suggestion
  suspend fun upsert(
    eventId: EventId,
    account: CalendarAccount,
    input: ExtendedPropertyInput
  ): ExtendedPropertyId = withContext(Dispatchers.IO) {
    val syncAdapterUri = CalendarContract.ExtendedProperties.CONTENT_URI.asSyncAdapter(account)
    val results = contentResolver.safeApplyBatch(
      arrayListOf(
        ContentProviderOperation.newDelete(syncAdapterUri)
          .withSelection(
            "${CalendarContract.ExtendedProperties.EVENT_ID} = ? AND ${CalendarContract.ExtendedProperties.NAME} = ?",
            arrayOf(eventId.value.toString(), input.name)
          )
          .build(),
        ContentProviderOperation.newInsert(syncAdapterUri)
          .withValues(input.toContentValues(eventId))
          .build()
      )
    )
    ExtendedPropertyId(
      results.last().uri?.lastPathSegment?.toLongOrNull()
        ?: throw IllegalStateException("Couldn't decode extended property ID from inserted content URI")
    )
  }

```

```kotlin
suspend fun ContentResolver.safeApplyBatch(
  operations: ArrayList<ContentProviderOperation>
): Array<ContentProviderResult> = withContext(Dispatchers.IO) {
  try {
    applyBatch(CalendarContract.AUTHORITY, operations)
  } catch (e: SecurityException) {
    throw PermissionException(Manifest.permission.WRITE_CALENDAR, e)
  } catch (e: OperationApplicationException) {
    throw CouldNotExecuteQueryException("Couldn't apply the batch of calendar operations", e)
  } catch (e: RemoteException) {
    throw CouldNotExecuteQueryException("Couldn't reach the calendar provider to apply a batch", e)
  }
}

```

Wenszel (changes requested):
Hey, thank you so much for the contribution! It's something we really want to land. The PR looks solid - I've just left a few comments regarding the implementation.

MoOx:
@Wenszel Thanks, all four applied.

- **Prefix rule narrowed to Google accounts.** You're right that it is the adapter's convention rather than a provider rule, so the check now only fires on `com.google`, and the docs say where the rule comes from.
- **`markDirty` uses an empty `ContentValues`.** Simpler, and it drops the read-modify-write of `DTSTART` along with the race it carried. One question on it below.
- **`name` and `value` are nullable on the entity.** The cursor extension no longer throws on a row the provider is free to return. The mapper then leaves a property with no name out of the JS surface rather than surfacing a null, since such a property can be neither read back nor deleted by name. Happy to expose the nulls instead if you would rather the API mirror the table exactly.
- **`upsert` writes through `applyBatch`.** `deleteByName` follows, selecting on the name instead of a row id, so `findAllByName` and the read-then-write it existed for are both gone.

Two of these are now measured on the device the original spike ran on (Huawei P30, EMUI, Google account):

- A delete with a selection on the `extendedproperties` collection URI is accepted, both as `event_id = ?` and as `event_id = ? AND name = ?`. That is the assumption the batch rests on, so it seemed worth checking before relying on it.
- The table does accept two rows under the same name on one event, so the duplicates the batch collapses are real rather than theoretical.

The third I could not measure: the `adb content` tool refuses to build an empty `ContentValues` (`[ERROR] Bindings not specified`), so nothing ever reaches the provider and the result says nothing either way. Does your reading of `CalendarProvider2:4123` cover an update that carries no values at all, or is there an earlier return for an empty set? If the empty update does not flag the event, the property is written locally and never pushed, which is the silent failure this API exists to prevent, so I would rather ask than assume. Everything else here I am happy to take as is.

Wenszel (changes requested):
Hey, I've done some testing using an emulator with a Google account set up.

Thank you for bringing up your concerns about `ContentResolver.update` with empty `ContentValues`. I tested it, and it correctly marks the event as dirty; it was actually `adb` that refused to build empty `ContentValues`.

Also, turns out there are some more Google account quirks:

1. First, we should manually set `Event.HAS_EXTENDED_PROPERTIES = 1`; without it, the properties disappeared after sync.
2. Could we mark the event as dirty before changing its properties, within the same `applyBatch`?

MoOx:
Thanks for testing this on a real Google account, that settles both points.

Both now travel at the head of the same batch: the event goes through the plain URI with `HAS_EXTENDED_PROPERTIES = 1`, followed by the delete and the insert on the sync adapter URI. So the event is flagged before the rows move, and the properties survive the sync.

The flag also turned out to be required rather than just useful: `ContentProviderOperation.Builder.build()` rejects an update carrying no value, so the empty `ContentValues` that worked through `ContentResolver.update` cannot travel in a batch. `EventRepository.markDirty` is gone as a result, its only two callers were these paths.

`deleteExtendedProperty` keeps the flag at 1 on purpose, so the sync adapter still looks at an event whose last property just went away.

PR #49431, opened 2026-08-27
[router] Remove devtools related stackRef

Ubax:
## Why

The `stackRef` is one of the blockers for concurrent react migration. It is used only in react-navigation/devtools which will no longer work with expo-router anyway.

We will add our own devtools as part of https://linear.app/expo/issue/ENG-20826

## How

1. Remove `stackRef` and `withStack`
2. Remove `stack` param from `__unsafe_action__`

### DIFF
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -81,6 +81,7 @@
 ### 💡 Others
 
 - Read navigation state from the React tree instead of imperative refs. ([#49433](https://github.com/expo/expo/pull/49433) by [@Ubax](https://github.com/Ubax))
+- Remove the dev-only `stack` field from the `__unsafe_action__` event in `expo-router/react-navigation`.
 - Scope routing queues to each router root and bind `useRouter()` to its owning container. ([#49351](https://github.com/expo/expo/pull/49351) by [@Ubax](https://github.com/Ubax))
 - Derive `useIsFocused` from context. ([#49390](https://github.com/expo/expo/pull/49390) by [@Ubax](https://github.com/Ubax))
 - Base `useNavigationState` on global state. ([#49381](https://github.com/expo/expo/pull/49381) by [@jakub-agent](https://github.com/jakub-agent))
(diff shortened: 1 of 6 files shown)


packages/expo-router/CHANGELOG.md
```diff
@@ -79,6 +79,7 @@
 
 ### 💡 Others
 
+- Remove the dev-only `stack` field from the `__unsafe_action__` event in `expo-router/react-navigation`. ([#49431](https://github.com/expo/expo/pull/49431) by [@Ubax](https://github.com/Ubax))
```
jakub-agent, line 83: [💡 suggestion] `NavigationContainerEventMap` is publicly exported (`react-navigation/core/index.tsx` does `export * from './types'`, re-exported through `expo-router/react-navigation`), so dropping `data.stack` is a compile error for any TS consumer reading it. This same unpublished release lists comparable `expo-router/react-navigation` type removals under `🛠 Breaking changes`, so consider moving the entry there for consistency.

Counter-argument for leaving it in `Others`: the event is prefixed `__unsafe_action__`, and it is not emitted at all on `main` right now, so no consumer can realistically be reading the field. Either placement is defensible — worth a deliberate call rather than a default.

jakub-agent (approved):

Clean dead-code removal. `stackRef` existed only to feed `data.stack` on the `__unsafe_action__` event, and `onDispatchAction` has no call site on `main` today (ENG-26123), so the event is not emitted at all — the capture was unreachable. Removing `withStack` also drops a per-dispatch `try/finally` and a dev-mode `new Error().stack` from every navigation helper, which is a small win.

Verified:
- No remaining `stackRef` references in `packages/expo-router/src`; the now-unused `use` / `NavigationBuilderContext` imports in `useNavigationCache.tsx` are correctly dropped.
- The only in-repo consumer of `__unsafe_action__`, `expo-observe`'s react-navigation integration, structurally types the listener payload and never reads `stack`, so it is unaffected.
- `expo-router/src/navigationEvents/navigation.ts` forwards only `action.type`, `payload`, and `state` to the public `actionDispatched` event — `stack` never reached that surface.
- `tsc --noEmit -p packages/expo-router/tsconfig.json` passes at this commit.
- `useNavigationCache`, `BaseNavigationContainer`, and `headless-tabs` suites pass (73 passed, 4 skipped).

One suggestion on changelog classification. Otherwise good to merge.

PR #49433, opened 2026-08-27
[router] Replace imperative state reads with in-tree context

Ubax:
# Why

In many places we still read `getState` or `store.state` instead of relying on the global state passed to navigators. 

# How

1. Remove `getState` and `getStateForKey`, and use `RootNavigationStateContext` and `NavigatorStateContext` instead
2. Replace usages of `store.state` by `RootNavigationStateContext`

# Test Plan

CI

# Checklist

### DIFF
--- a/packages/expo-router/src/global-state/useNavigationTreeReducer.ts
+++ b/packages/expo-router/src/global-state/useNavigationTreeReducer.ts
@@ -221,7 +221,6 @@ export function useNavigationTreeReducer({
     () => ({ registry, routeNode, linking, redirects }),
     [registry, routeNode, linking, redirects]
   );
-  const stateRef = React.useRef(state);
   const previousRegistryRef = React.useRef(registry);
 
   const processAction = React.useCallback(
@@ -246,17 +245,12 @@ export function useNavigationTreeReducer({
     warnIfScreenParam(params);
     process({ type: 'ACTION', payload: { action, originKey } });
   });
-  const getState = React.useCallback(() => stateRef.current, []);
-  const getStateForKey = React.useCallback(
-    (key: string) => findStateByKey(stateRef.current, key),
-    []
-  );
   const resetNavigator = useLatestCallback((stateKey: string, routerType: string | undefined) => {
     process({ type: 'NAVIGATOR_CHANGED', stateKey, routerType });
   });
 
   React.useInsertionEffect(() => {
-    stateRef.current = state;
+    // TODO(@ubax): Check if this is still needed
     onStateChangeInsertion?.(state);
   }, [onStateChangeInsertion, state]);
 
@@ -272,8 +266,6 @@ export function useNavigationTreeReducer({
 
   return {
     state,
-    getState,
-    getStateForKey,
     resetNavigator,
     handleAction,
     processIntent,
--- a/packages/expo-router/src/react-navigation/core/__tests__/useEventEmitter.test.ios.tsx
+++ b/packages/expo-router/src/react-navigation/core/__tests__/useEventEmitter.test.ios.tsx
@@ -274,7 +274,7 @@ test('fires focus and blur events in nested navigator', () => {
   act(() => parent.current.navigate('first'));
 
   expect(firstFocusCallback).toHaveBeenCalledTimes(2);
-  expect(thirdBlurCallback).toHaveBeenCalledTimes(2);
+  expect(thirdBlurCallback).toHaveBeenCalledTimes(1);
 
   act(() => {
     child.current.navigate('fourth');
@@ -282,7 +282,7 @@ test('fires focus and blur events in nested navigator', () => {
   });
 
   expect(fourthFocusCallback).toHaveBeenCalledTimes(3);
-  expect(thirdBlurCallback).toHaveBeenCalledTimes(2);
+  expect(thirdBlurCallback).toHaveBeenCalledTimes(1);
   expect(firstBlurCallback).toHaveBeenCalledTimes(2);
 
   act(() => child.current.navigate('third'));
@@ -298,7 +298,7 @@ test('fires focus and blur events in nested navigator', () => {
   expect(secondBlurCallback).toHaveBeenCalledTimes(1);
 
   expect(thirdFocusCallback).toHaveBeenCalledTimes(2);
-  expect(thirdBlurCallback).toHaveBeenCalledTimes(2);
+  expect(thirdBlurCallback).toHaveBeenCalledTimes(1);
 
   expect(fourthFocusCallback).toHaveBeenCalledTimes(3);
   expect(fourthBlurCallback).toHaveBeenCalledTimes(3);
--- a/packages/expo-router/src/react-navigation/core/useIsFocused.tsx
+++ b/packages/expo-router/src/react-navigation/core/useIsFocused.tsx
@@ -6,6 +6,17 @@ export const FocusedRouteKeyContext = React.createContext<string | undefined>(un
 
 export const IsFocusedContext = React.createContext<boolean | undefined>(undefined);
 
+export function useIsRouteFocused(routeKey: string | undefined): boolean {
+  const parentIsFocused = use(IsFocusedContext);
+  const focusedRouteKey = use(FocusedRouteKeyContext);
+
+  if (routeKey === undefined) {
+    return parentIsFocused ?? true;
+  }
+
+  return parentIsFocused == null || parentIsFocused ? focusedRouteKey === routeKey : false;
+}
+
 /**
  * Hook to get the current focus state of the screen. Returns a `true` if screen is focused, otherwise `false`.
  * This can be used if a component needs to render something based on the focus state.
--- a/packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
+++ b/packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
@@ -9,7 +9,6 @@ import { useComponent } from '../../fork/useComponent';
 import { type RouterRegistryEntry, useRegisterRouter } from '../../global-state/routerRegistry';
 import { useEnqueueRoutingIntent } from '../../global-state/routingQueueContext';
 import { resetNavigatorState } from '../../global-state/stateUtils';
-import useLatestCallback from '../../utils/useLatestCallback';
 import {
   type DefaultRouterOptions,
   type NavigationAction,
@@ -263,7 +262,6 @@ export function useNavigationBuilder<
   useRegisterNavigator();
   const routeNode = useRouteNode();
   const enqueue = useEnqueueRoutingIntent();
-
   const {
     children,
     layout,
@@ -333,7 +331,7 @@ export function useNavigationBuilder<
 
   const { state: currentState } = use(NavigationStateContext);
 
-  const { getStateForKey, resetNavigator, handleAction } = use(NavigationBuilderContext);
+  const { resetNavigator, handleAction } = use(NavigationBuilderContext);
   if (
     currentState === undefined ||
     currentState.stale !== false ||
@@ -375,22 +373,6 @@ export function useNavigationBuilder<
       }),
     [routeNamesKey, router]
   );
-  const getState = useLatestCallback((): State => {
-    const currentState = getStateForKey(stateKeyRef.current);
-    if (currentState === undefined) {
-      return committedState;
-    }
-    if (currentState.stale !== false) {
-      throw new Error(
-        'The mounted navigator no longer has complete state in the global navigation tree.'
-      );
-    }
-    if (currentState.type !== undefined && currentState.type !== router.type) {
-      // The reset keeps the complete fields required by every navigator state.
-      return resetNavigatorState(currentState, router.type) as State;
-    }
-    return currentState as State;
-  });
   const emitter = useEventEmitter<EventMapCore<State>>((e) => {
     const routeNames = [];
 
@@ -463,12 +445,11 @@ export function useNavigationBuilder<
   const { keyedListeners, addKeyedListener } = useKeyedChildListeners();
 
   const { isRoutePrevented, preventRemoveContextValue } = usePreventRemoveState({
-    getState,
-    state,
+    state: committedState,
   });
 
   useOnPreventRemove({
-    getState,
+    state: committedState,
     isRoutePrevented,
     emitter,
     preventRemoveListeners: keyedListeners.preventRemove,
@@ -525,9 +506,7 @@ export function useNavigationBuilder<
     if (isForeignType) {
       return;
     }
-    const committed = getState();
-
-    if (isArrayEqual(committed.routeNames, routeNames)) {
+    if (isArrayEqual(committedState.routeNames, routeNames)) {
       pendingRouteNamesRef.current = undefined;
     } else if (!isArrayEqual(pendingRouteNamesRef.current ?? [], routeNames)) {
       pendingRouteNamesRef.current = routeNames;
@@ -537,9 +516,9 @@ export function useNavigationBuilder<
           action: {
             type: 'ROUTE_NAMES_CHANGED',
             payload: { routeNames },
-            target: committed.key,
+            target: committedState.key,
           },
-          originKey: committed.key,
+          originKey: committedState.key,
         },
       });
     }
@@ -548,7 +527,7 @@ export function useNavigationBuilder<
   const navigation = useNavigationHelpers<State, ActionHelpers, NavigationAction, EventMap>({
     id: options.id,
     handleAction: onAction,
-    getState,
+    state: committedState,
     emitter,
     router,
   });
@@ -565,7 +544,7 @@ export function useNavigationBuilder<
     navigation,
     screenOptions,
     screenLayout,
-    getState,
+    state: committedState,
     addListener,
     addKeyedListener,
     router,
--- a/packages/expo-router/src/react-navigation/core/useNavigationCache.tsx
+++ b/packages/expo-router/src/react-navigation/core/useNavigationCache.tsx
@@ -2,7 +2,6 @@
 import * as React from 'react';
 import { use } from 'react';
 
-import { isRoutePreloadedInStack } from '../../utils/stack';
 import {
   CommonActions,
   type NavigationAction,
@@ -21,7 +20,6 @@ type Options<
 > = {
   routes: State['routes'];
   routeNames: State['routeNames'];
-  getState: () => State;
   navigation: NavigationHelpers<ParamListBase> &
     Partial<NavigationProp<ParamListBase, string, any, any, any>>;
   setOptions: (
@@ -47,6 +45,9 @@ type NavigationCache<
  * Hook to cache navigation objects for each screen in the navigator.
  * It's important to cache them to make sure navigation objects don't change between renders.
  * This lets us apply optimizations like `React.memo` to minimize re-rendering screens.
+ * Exception: a route's navigation object changes identity once when the route is promoted from
+ * preloaded to active.
+ * TODO(@ubax): consider resolving `isPreloaded` at call time to keep one object per route.
  */
 export function useNavigationCache<
   State extends NavigationState,
@@ -56,7 +57,6 @@ export function useNavigationCache<
 >({
   routes,
   routeNames,
-  getState,
   navigation,
   setOptions,
   router,
@@ -70,20 +70,19 @@ export function useNavigationCache<
   const cache = React.useMemo(
     () => ({ current: {} as NavigationCache<State, ScreenOptions, EventMap> }),
     // eslint-disable-next-line react-hooks/exhaustive-deps
-    [getState, navigation, setOptions, emitter]
+    [navigation, setOptions, emitter]
   );
 
   // Keep name-keyed placeholders stable after their real route keys are created.
-  const validKeys = new Set([...routes.map((route) => route.key), ...routeNames]);
+  const routeKeys = [...routes.map((route) => route.key), ...routeNames];
+  const validKeys = new Set(routeKeys.flatMap((key) => [key, `p\0${key}`]));
   cache.current = Object.fromEntries(
     Object.entries(cache.current).filter(([key]) => validKeys.has(key))
   );
 
-  const createNavigation = (route: { key: string; name: string }) => {
+  const createNavigation = (route: { key: string; name: string }, isPreloaded: boolean) => {
     const dispatchSync = (action: NavigationAction) => {
-      const state = getState();
-
-      if (isRoutePreloadedInStack(state, route)) {
+      if (isPreloaded) {
         if (process.env.NODE_ENV !== 'production') {
           console.warn(
             `Ignored a navigation action dispatched from the preloaded screen '${route.name}'. The screen is rendered for preloading and is not focused, so its actions would unexpectedly modify the visible stack. Wait until the screen is focused before dispatching.`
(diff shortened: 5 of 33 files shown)


packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
```diff
@@ -332,8 +335,9 @@ export function useNavigationBuilder<
   const routeNamesKey = routeNames.join('\0');
 
   const { state: currentState } = use(NavigationStateContext);
+  const rootState = use(RootNavigationStateContext);
```
jakub-agent: [🏗️ **design**] Subscribing every navigator to the root state changes the re-render scope of the whole app.

`RootNavigationStateContext`'s value is the root tree state, so its identity changes on every navigation. Context updates propagate through `StaticContainer` and `React.memo`, so from here on every component that calls `useNavigationBuilder` re-renders on any navigation anywhere in the tree. Before this change a nested navigator only re-rendered when its own `NavigationStateContext` value (from the enclosing `SceneView`) changed. The same applies to every `SceneView`, which now reaches root state through `useOptionsGetters`.

I think this is deliberate — the derived `isFocused` in `useOptionsGetters` only works because these components are guaranteed to re-render — but it is worth stating explicitly rather than leaving implicit. Could the same information come from something narrower, e.g. a per-navigator state context keyed by state key, or the `IsFocusedContext` that #49390 already added?

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
```diff
@@ -345,10 +349,13 @@ export function useNavigationBuilder<
     );
   }
 
-  const isForeignType = currentState.type !== undefined && currentState.type !== router.type;
+  const treeState = rootState
+    ? (findStateByKey(rootState, currentState.key) ?? currentState)
```
jakub-agent: [🏗️ **design**] Two things about this lookup:

1. **When does it actually differ from `currentState`?** `currentState` comes from the enclosing `SceneView`'s `NavigationStateContext`, which is derived from the same immutable tree. The only divergence I can construct is that the parent passes down a substate produced by `router.getStateForDeclaredRoutes`, whereas `findStateByKey` returns the committed tree state — which matches what the removed `getState()` used to return. If that is the case, could you say so in a comment? Nothing in the diff tests the divergence, so a regression here would be silent. If it never diverges, both the lookup and the root-state subscription above are unnecessary.

2. **Cost.** This is an unmemoized recursive walk of the whole state tree, on every render of every navigator. Combined with the root-state subscription, that is O(navigators × tree size) per navigation. `React.useMemo` on `[rootState, currentState.key]` would bound it cheaply.

packages/expo-router/src/react-navigation/core/usePreventRemoveState.tsx
```diff
@@ -70,13 +69,13 @@ export function usePreventRemoveState({ getState, state }: Props) {
...
       if (route?.key !== undefined && setParentPrevented !== undefined) {
-        const state = getState();
         const hasActiveEntry = [...next.values()].some(
           (entry) => entry.preventRemove && !isRoutePreloadedInStack(state, { key: entry.routeKey })
         );
         setParentPrevented(parentId, route.key, hasActiveEntry);
       }
-    }
+    },
+    [parentId, route?.key, setParentPrevented, state]
```
jakub-agent, line 78: [🏗️ **design**] Making `setPreventRemove` depend on `state` causes re-registration churn at every nesting level on every navigation.

This callback is handed to children through `PreventRemoveContext`, and the effect at line 106 lists `setParentPrevented` (the parent's `setPreventRemove`) in its deps. Now that the identity changes whenever the parent navigator's state changes, each navigation makes every nested level tear down and re-register: cleanup calls `setParentPrevented(parentId, route.key, false)`, setup calls it again with `isPrevented`. Each call mutates `entriesRef` and calls `setEntries`. The final value is right because the updates batch, so I do not think this is user-visible — but it is wasted work proportional to nesting depth, and it did not happen when the callback was a stable `useLatestCallback`.

The re-registration also newly exercises the `Couldn't find a route with the key …` throw at line 47, against a `state` that just changed, in a situation where the effect previously would not have re-run at all. Worth convincing yourself that path is unreachable, or keeping `useLatestCallback` here and reading the freshest state some other way.

packages/expo-router/src/react-navigation/core/useOptionsGetters.tsx
```diff
@@ -54,9 +85,7 @@ export function useOptionsGetters({ key, options, navigation }: Options) {
   }, []);
 
   const getCurrentOptions = React.useCallback(() => {
-    const isFocused = navigation?.isFocused() ?? true;
-
-    if (!isFocused) {
+    if (!(navigation?.isFocused() ?? true)) {
```
jakub-agent: [🏗️ **design**] `getCurrentOptions` still reads focus imperatively while `optionsChangeListener` (line 62) now derives it from root state. Two sources of truth for the same question, thirty lines apart, in the one hook this PR is reworking.

As far as I can tell they agree today: `navigation.isFocused()` on a descriptor navigation walks the committed states, and `isRouteFocused` walks the focused path of the same tree. But the divergence is exactly the class of bug this PR exists to remove, so it would be good to close it here rather than leave it for later.

Separately: `isRouteFocused` is a third implementation of "is this route on the focused path". #49390 added `IsFocusedContext`/`useIsFocused` for this; reusing it would delete both this helper and the root-state subscription that `SceneView` picks up from this hook.

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
```diff
@@ -565,13 +555,22 @@ export function useNavigationBuilder<
...
     router,
     // @ts-expect-error: this should have both core and custom events, but too much work right now
     emitter,
   });
+  const focusedRouteKey = state.routes[state.index]?.key;
+  const focusedNavigation = focusedRouteKey ? descriptors[focusedRouteKey]?.navigation : undefined;
+  const tabNavigations = React.useMemo(
+    () => [
+      ...parentTabNavigations,
+      ...(router.type === 'tab' && focusedNavigation ? [focusedNavigation] : []),
```
jakub-agent: [🏗️ **design**] This provides the *focused* tab route's navigation object to *all* children of the tab navigator, including the screens of unfocused tabs.

The provider wraps `element` (every screen), but the list only ever contains `descriptors[state.routes[state.index]?.key]?.navigation`. So a screen inside an unfocused tab now subscribes `tabPress` on the focused tab's route key, where the old `useScrollToTop` walk gave it its own. The `navigation.isFocused()` guard in the handler stops it from mis-scrolling, and the context refreshes once focus moves, so I believe observable behavior is preserved — but the invariant "each screen listens to its own tab" is gone and nothing tests the unfocused-tab case.

Providing the per-route navigation inside the descriptor's render would keep the old invariant. If the focused-only list is intentional and sufficient, a comment saying why would help — this is subtle enough that the next reader will not reconstruct it.

packages/expo-router/src/react-navigation/core/useNavigationCache.tsx
```diff
@@ -56,21 +56,22 @@ export function useNavigationCache<
...
   routeNames,
-  getState,
+  state,
   navigation,
   setOptions,
   router,
   emitter,
 }: Options<State, ScreenOptions, EventMap>) {
   const { stackRef } = use(NavigationBuilderContext);
+  const getState = React.useEffectEvent(() => state);
```
jakub-agent: [💡 suggestion] This is the opposite choice from the one you made in `useNavigationHelpers`, where the new comment reads "Unlike handler-only Effect Events, the public accessor can be called during render" and the code uses `useLatestCallback`.

Here `getState` feeds `dispatch`/`dispatchSync`, which are handlers in normal use, so `useEffectEvent` works. But `navigation.dispatch(...)` is public API, and calling it during render — a pattern some apps use for redirects — would now throw React's "cannot call an Effect Event during render" instead of navigating or hitting the existing preloaded-screen warning. That is a worse failure mode than the one it replaces.

Is the asymmetry deliberate? If not, `useLatestCallback` here would keep the two accessors consistent.

packages/expo-router/src/react-navigation/core/__tests__/useEventEmitter.test.ios.tsx
```diff
@@ -274,15 +274,15 @@ test('fires focus and blur events in nested navigator', () => {
   act(() => parent.current.navigate('first'));
 
   expect(firstFocusCallback).toHaveBeenCalledTimes(2);
-  expect(thirdBlurCallback).toHaveBeenCalledTimes(2);
+  expect(thirdBlurCallback).toHaveBeenCalledTimes(1);
```
jakub-agent, line 277: [💡 suggestion] Worth calling this out in the changelog: it is an observable behavior change for any app with a `blur` listener.

I traced the scenario — `third` is only globally focused once (after `child.navigate('third')` at line 268), so the single blur after `parent.navigate('first')` is correct and the old count of 2 included a spurious duplicate. So this is a fix, not a regression. But "duplicate blur events are no longer emitted" is not something a reader can infer from "Read navigation state from the React tree instead of imperative refs", and three assertion changes with no explanation invite the reviewer to assume an event went missing. A line in the changelog and a sentence in the PR body would settle it.

packages/expo-router/src/ui/TabTrigger.tsx
```diff
@@ -170,11 +171,8 @@ export function useTabTrigger(options: TabTriggerProps): UseTabTriggerResult {
       const hidden =
         routeOptions != null && 'hidden' in routeOptions && routeOptions.hidden === true;
 
-      // Parent triggers are inherited, so read the state of the navigator that registered them.
       const owningState =
-        config.type === 'internal' && config.contextKey !== contextKey
-          ? navigation?.getParent(config.contextKey)?.getState()
-          : state;
+        config.type === 'internal' ? navigatorStates[config.contextKey] : undefined;
```
jakub-agent, line 176: [🐧 nit] The deleted comment — "Parent triggers are inherited, so read the state of the navigator that registered them." — still explains this line. It is the reason `owningState` is looked up by `config.contextKey` instead of just using `state`, which is otherwise easy to mistake for redundancy now that the `config.contextKey !== contextKey` branch is gone. Worth keeping.

packages/expo-router/src/global-state/useNavigationTreeReducer.ts
```diff
@@ -246,17 +245,12 @@ export function useNavigationTreeReducer({
...
-    (key: string) => findStateByKey(stateRef.current, key),
-    []
-  );
   const resetNavigator = useLatestCallback((stateKey: string, routerType: string | undefined) => {
     process({ type: 'NAVIGATOR_CHANGED', stateKey, routerType });
   });
 
   React.useInsertionEffect(() => {
-    stateRef.current = state;
+    // TODO(@ubax): Check if this is still needed
```
jakub-agent, line 253: [🐧 nit] This TODO replaced the line it used to sit next to, so the insertion effect is now annotated with a comment that says nothing about what it does or why it must run at insertion time. Same shape as the new `// TODO(@ubax): check if this is still needed anywhere` on `isReady` in `BaseNavigationContainer.tsx:173`.

Either resolve them in this PR or link an issue, as the neighbouring TODOs do (`ENG-26124`, `ENG-22019`). An unlinked "check if this is still needed" is not actionable by anyone but you.

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
```diff
@@ -525,7 +515,7 @@ export function useNavigationBuilder<
     if (isForeignType) {
       return;
     }
-    const committed = getState();
+    const committed = committedState;
```
jakub-agent: [🐧 nit] Redundant alias now that there is no `getState()` call to name — use `committedState` directly in the lines below.

packages/expo-router/src/link/preview/__tests__/HrefPreview.test.ios.tsx
```diff
@@ -331,6 +332,54 @@ it('Renders not found for not existing href', async () => {
...
+    index: Index,
+    secret: () => <View testID="secret" />,
+  });
+
+  expect(screen.getByTestId('secret')).toBeVisible();
+
+  const stateSpy = jest.spyOn(store, 'state', 'get').mockReturnValue(undefined);
+  act(() => setGuard(false));
+  fireEvent.press(screen.getByTestId('rerender'));
+  expect(stateSpy).not.toHaveBeenCalled();
```
jakub-agent: [🐧 nit] The test asserts the getter is never called, which makes the `mockReturnValue(undefined)` on the line above dead setup — and makes the title ("when store state is unavailable") describe a scenario the test proves cannot occur. What it really checks is "`HrefPreview` no longer reads `store.state`", which is a good thing to pin. Suggest retitling to that and dropping the return value so the intent is obvious.

Also, `setGuard = setGuardState` in `Layout`'s body assigns an outer variable during render, which the repo's React rules discourage. Lifting the guard into a `useState` in the test and passing it as a prop, or driving it through a ref set in an effect, avoids the render-phase write.

jakub-agent (commented):

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
Ubax: Revert this change if possible

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
Ubax: It this context actually needed here? Can't we just use `NavigationStateContext`?

packages/expo-router/src/link/preview/__tests__/HrefPreview.test.ios.tsx
Ubax: Remove this test

packages/expo-router/src/global-state/useNavigationTreeReducer.ts
Ubax: Leave as is

packages/expo-router/src/react-navigation/core/__tests__/useEventEmitter.test.ios.tsx
Ubax: Add this to PR description

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
Ubax: Do it

packages/expo-router/src/react-navigation/core/useNavigationBuilder.tsx
Ubax: Move this out of `useNavigationBuilder`. This should only be part of headless tabs, since it was added for that navigator

packages/expo-router/src/react-navigation/core/useNavigationCache.tsx
Ubax: There is no need to cache here. It is not a dependency of any of the deps

packages/expo-router/src/react-navigation/core/useOptionsGetters.tsx
Ubax: Can we just use `useIsFocused` here instead of the `rootState`?

packages/expo-router/src/react-navigation/core/usePreventRemoveState.tsx
Ubax: Keep as is, https://github.com/expo/expo/pull/49408 will remove it

packages/expo-router/src/ui/TabTrigger.tsx
Ubax: Add it back

packages/expo-router/src/react-navigation/core/useNavigationCache.tsx
```diff
@@ -81,8 +81,6 @@ export function useNavigationCache<
 
   const createNavigation = (route: { key: string; name: string }) => {
     const dispatchSync = (action: NavigationAction) => {
```
jakub-agent, line 84: [🚨 **critical**] This is the regression. `createNavigation` now closes over the render-time `state`, but its result is stored in `cache.current[route.key]` and that cache is only rebuilt when `[navigation, setOptions, emitter]` change — all of which are stable. So the `state` these two guards see is frozen at the render in which the route's navigation object was *first created*, and never refreshes.

Concretely, for a route created while it was preloaded:

1. `preload('second')` → stack is `routes: [first, second], index: 0`. `second` is preloaded.
2. `second`'s navigation object is created here and cached, capturing that state.
3. `navigate('second')` → `index: 1`. `second` is now the focused route. The cached object is reused — `useNavigationCache.test.ios.tsx:348` asserts exactly that.
4. `second` dispatches anything → `isRoutePreloadedInStack(frozenState, second)` still finds index `1 > 0` → the action is silently dropped with the dev warning.

So navigation from any screen that was ever preloaded stops working. `ignores dispatches from a preloaded stack screen until it is promoted` catches it: `expect(enqueue).toHaveBeenCalledTimes(1)` receives 0. It passes on `4e2d728` and on `075e7c1`, and fails on `60c0a2a` in isolation.

Re your "no need to cache here" — the thing that needs the indirection is not the `useMemo` dep list, it is that the navigation object outlives the render that built it. Restoring `useLatestCallback(() => state)` fixes it without reintroducing the `useEffectEvent` render-phase hazard I flagged last round. Adding `state` to the `cache` deps would also work but throws away the reference stability the hook exists for.

packages/expo-router/src/react-navigation/core/useIsFocused.tsx
```diff
@@ -6,6 +6,17 @@ export const FocusedRouteKeyContext = React.createContext<string | undefined>(un
 
 export const IsFocusedContext = React.createContext<boolean | undefined>(undefined);
 
+export function useIsRouteFocused(routeKey: string | undefined): boolean {
+  const parentIsFocused = use(IsFocusedContext);
+  const focusedRouteKey = use(FocusedRouteKeyContext);
+
+  if (routeKey === undefined) {
```
jakub-agent, line 13: [💡 suggestion] Returning `true` here ignores `parentIsFocused`, and the new test pins that: `{ routeKey: undefined, parentIsFocused: false, focusedRouteKey: 'route', expected: true }`. A hook called `useIsRouteFocused` reporting focused while the context above it says the subtree is not focused is a surprising contract to lock in with a test.

The only caller that passes `undefined` is `useOptionsGetters` from `BaseNavigationContainer`, where there is no `IsFocusedContext` above, so `parentIsFocused ?? true` gives the same answer today and stays right if a container is ever mounted inside an unfocused screen. Suggest that instead, and flipping the test case to match.

packages/expo-router/src/react-navigation/native/useScrollToTop.tsx
```diff
@@ -64,63 +62,44 @@ export function useScrollToTop(ref: React.RefObject<ScrollableWrapper>) {
...
+            scrollable.scrollResponderScrollTo({ y: 0, animated: true });
+          }
+        }
+      });
+    };
 
+    const unsubscribers: (() => void)[] = [];
+    let currentNavigation: typeof navigation | undefined = navigation;
+    while (currentNavigation) {
+      // Non-tab navigators never emit `tabPress`, so these listeners are inert.
```
jakub-agent, line 95: [🏗️ **design**] Your direction was "move this out of `useNavigationBuilder` … should only be part of headless tabs, since it was added for that navigator". The context is out of `useNavigationBuilder`, but it was deleted rather than moved, and the `getParent()` walk came back — now subscribing every ancestor rather than only tab ones. Worth confirming this is the landing point you wanted, or whether the headless-tabs-owned version is still coming.

As a result this file no longer advances the PR's goal: the walk is imperative again, and `navigation.isFocused()` in the handler still reads state through `getState()`. That is defensible inside an event handler, and trading one `getState().type` read for a few inert listeners is a fine deal — but the file reads as reverted rather than converted, so a line in the PR description would help.

Separately, `useScrollToTop` is public API, has been rewritten twice across these two iterations, and has no test anywhere in the package. The nested-stack-inside-tabs case (`isFirst` now derived from `NavigatorStateContext` instead of `navigation.getState()`) is exactly the kind of thing that would silently break. Worth one test.

jakub-agent (changes requested):

packages/expo-router/src/react-navigation/core/useNavigationCache.tsx
```diff
@@ -184,14 +179,15 @@ export function useNavigationCache<
     return navigationItem;
   };
 
-  return (route: { key: string; name: string }) => {
-    const cachedNavigation = cache.current[route.key];
+  return (route: { key: string; name: string }, isPreloaded: boolean) => {
+    const key = `${isPreloaded ? 'p\0' : ''}${route.key}`;
+    const cachedNavigation = cache.current[key];
```
jakub-agent, line 187: [🏗️ **design**] Keying the cache by `isPreloaded` means a route's `navigation` prop changes identity when it is promoted from preloaded to active. That is a deliberate change — `useNavigationCache.test.ios.tsx:340` now asserts `expect(navigation).not.toBe(preloadedNavigation)`, where it previously asserted the opposite — but the hook's own doc comment above still says:

> It's important to cache them to make sure navigation objects don't change between renders. This lets us apply optimizations like `React.memo` to minimize re-rendering screens.

That is now false for the one transition where it matters most. Worth stating the exception there, since the next reader will otherwise rely on the promise.

The substantive question: the preloaded object stays a permanent no-op after promotion. Anything holding that reference — a ref captured on mount, a closure inside a long-lived subscription, `navigation` read once with empty effect deps — silently dispatches into a void with only a dev warning. I grepped expo-router and found no capture-once patterns internally, so this is not a live bug in the package; it is a narrowing of the public prop's contract that user code can trip over. The alternative (keep one object, resolve `isPreloaded` at call time) preserves both the guard and identity, at the cost of the indirection you removed.

Either way is defensible — I would just like it to be an explicit decision rather than a side effect of the fix, and the doc comment updated to match whichever you pick.

jakub-agent (commented):

packages/expo-router/src/react-navigation/core/useIsFocused.tsx
Ubax: Do it

packages/expo-router/src/react-navigation/core/useNavigationCache.tsx
Ubax: Leave TODO to look into this later

packages/expo-router/src/react-navigation/native/__tests__/useScrollToTop.test.ios.tsx
```diff
@@ -0,0 +1,138 @@
...
+    second: createScrollableScreen(scrollTo),
+  });
+
+  await userEvent.press(screen.getByRole('button', { name: 'index, tab, 1 of 2' }));
+  flushAnimationFrame();
+
+  expect(scrollTo).not.toHaveBeenCalled();
+});
+
+test('scrolls the first screen of a stack nested in a tab to the top', async () => {
```
jakub-agent, line 63: [🐧 nit] Non-blocking, and only worth doing if you're touching this file again: the one case the old implementation called out by name isn't in the set — its comment read "If the screen is nested inside multiple tab navigators, we should scroll to top for any of them", and the loop it justified is gone.

The replacement walk subscribes every ancestor unconditionally, so nested tabs are covered by construction and arguably more robustly than before. That's why I'd not hold the PR for it. But it was the explicit motivation for the code being replaced, so a tabs-inside-tabs variant of this test would close the loop on the rewrite.

jakub-agent (approved):

PR #49435, opened 2026-08-27
[fingerprint] Resolve the git root from the project root instead of the process cwd

janicduplessis:
# Why

`@expo/fingerprint` resolves a project's workflow (`managed` vs `generic`) by asking git for the repository root, but `ProjectWorkflow.ts` spawned `git rev-parse --show-toplevel` without a `cwd` — so the answer came from wherever the calling process happened to be, not from the project being fingerprinted.

Concrete repro: fingerprinting a git worktree while the process cwd sat in the main checkout. The git root resolved to the caller's repo, so `git check-ignore` ran against the wrong repository and missed the project's `.gitignore` — the workflow flipped from `managed` to `generic` and the fingerprint hash silently changed for the same project.

# How

- `GitClient` now receives the project root and runs `git rev-parse --show-toplevel` with `cwd: projectRoot`.
- `isGitInstalledAndConfiguredAsync()` had the same latent issue: it decided whether the *project* is inside a git repo from the caller's cwd, which picks `GitClient` vs `NoVCSClient` wrongly when only one of the two locations is a git repo. It now runs its `rev-parse` probe with `cwd: projectRoot` too.

The `git check-ignore` call already used the resolved root as `cwd`, so it is fixed transitively. These are the only git spawns in the package.

# Test Plan

New test `src/__tests__/ProjectWorkflow-git-test.ts` (real temp git repos, no mocks): a project whose `ios` directory is gitignored (managed workflow) plus an unrelated git repo; the test `process.chdir`s into the unrelated repo and asserts the project still resolves as `managed`.

Written red-first — before the fix:

```
✕ returns managed workflow when cwd is inside another git repo
    Expected: "managed"
    Received: "generic"
✓ returns managed workflow when cwd is the project root
```

After the fix, the whole package suite passes (26 suites, 221 tests), and `et check-packages @expo/fingerprint` (run as `node ./tools/bin/expotools.js check-packages @expo/fingerprint`) passes build, typecheck, depscheck, test, lint, and format.

# Checklist

### DIFF
--- a/packages/@expo/fingerprint/CHANGELOG.md
+++ b/packages/@expo/fingerprint/CHANGELOG.md
@@ -18,6 +18,7 @@
 
 ### 🐛 Bug fixes
 
+- Resolve the git root from the project root instead of the process working directory during project workflow detection, so fingerprinting a project (for example a git worktree) while the working directory is inside another repository no longer flips the workflow from `managed` to `generic` and silently changes the hash. ([#49435](https://github.com/expo/expo/pull/49435) by [@janicduplessis](https://github.com/janicduplessis))
 - Set development mode before loading Expo config and `.env` files. ([#48839](https://github.com/expo/expo/pull/48839) by [@ramonclaudio](https://github.com/ramonclaudio))
 - Fixed ignore patterns (built-in and `.fingerprintignore`) not matching on Windows, which made fingerprints differ between Windows machines and EAS builds ("Runtime version mismatch"). ([#46816](https://github.com/expo/expo/pull/46816) by [@blurbyte](https://github.com/blurbyte))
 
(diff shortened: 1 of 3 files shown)


jonesjrrod35-arch (commented):
Oh, my gosh, i'm so thankful so so thankful resolve. I love you so much. I don't know what to do to say. I'm just, I just had a loss for words. I'm just so stunned in the shock. But this five thousand dollars I really needed for me and my kids. I promised I would pay back every cent. Think you think you thank you so much

brentvatne (approved):

ramonclaudio:
@expo-bot verify

jonesjrrod35-arch:
Verify

ramonclaudio (approved):

Kudo (approved):

jonesjrrod35-arch:
node ./tools/bin/expotools.js check-packages @expo/fingerprint

PR #49436, opened 2026-08-27
[fingerprint] Normalize the project root to its realpath so path spelling never changes the hash

janicduplessis:
# Why

The same project produces a different fingerprint hash depending on how its path is spelled. Observed on macOS where `/tmp` is a symlink to `/private/tmp`: fingerprinting a project as `/tmp/app` vs `/private/tmp/app` gives two different hashes; spelling the root as its realpath restores agreement.

The cause: Node resolves loaded modules to their real paths, so when `projectRoot` is a symlinked spelling, the config loader's `path.relative(projectRoot, filename)` escapes the project and leaks tree-specific traversal paths into the `expoConfigPlugins` file sources. From the repro, the same local plugin becomes:

```
realpath root:  plugins/withNoop.js
symlinked root: ../project/plugins/withNoop.js
```

# How

Normalize `projectRoot` with `fs.realpath` (the existing `maybeGetRealPathAsync` helper, which falls back to the given path on error) once at the top of `createFingerprintAsync`. Everything downstream — options, sourcers, dedup, hashing, and the out-of-process config loader, which receives the root from the parent — then sees the canonical spelling. `createProjectHashAsync`, `diffFingerprintChangesAsync`, and the CLI all route through `createFingerprintAsync`, so this is the single choke point; a few sourcers already realpathed locally for the same reason (e.g. autolinking results in `Expo.ts`).

# Test Plan

New test `src/__tests__/Fingerprint-symlink-test.ts`: builds a minimal Expo project with a local config plugin in a temp dir, creates a sibling symlink to it (skips if the platform can't create symlinks), and fingerprints the project through both spellings, asserting equal hashes. It also asserts the local plugin was actually captured as an `expoConfigPlugins` source, so the test can't pass vacuously if the out-of-process config loader fails. Like the e2e suites, it points `getExpoConfigLoaderPath` at the built `build/ExpoConfigLoader.js` for the child process.

Written red-first — before the fix:

```
✕ produces the same hash for the realpath and a symlinked spelling of the same project
    Expected: "f12db3ad79da3de02e5b563e2410b5e507a34b39"
    Received: "0c1bedfc288b4cefa284bd160f09868c1d7c50dc"
```

After the fix, the whole package suite passes (26 suites, 220 tests), and `et check-packages @expo/fingerprint` (run as `node ./tools/bin/expotools.js check-packages @expo/fingerprint`) passes build, typecheck, depscheck, test, lint, and format.

Note for reviewers: for a project whose root was previously passed as a non-realpath spelling, the fingerprint changes once to the realpath-based (correct) value.

# Checklist

### DIFF
--- a/packages/@expo/fingerprint/src/Fingerprint.ts
+++ b/packages/@expo/fingerprint/src/Fingerprint.ts
@@ -4,6 +4,7 @@ import { normalizeOptionsAsync } from './Options';
 import { compareSource, sortSources } from './Sort';
 import { createFingerprintFromSourcesAsync } from './hash/Hash';
 import { getHashSourcesAsync } from './sourcer/Sourcer';
+import { maybeGetRealPathAsync } from './sourcer/Utils';
 
 /**
  * Create a fingerprint for a project.
@@ -17,6 +18,10 @@ export async function createFingerprintAsync(
   projectRoot: string,
   options?: Options
 ): Promise<Fingerprint> {
+  // Use the realpath of the project root so that relative source paths - and with them the
+  // resulting hash - do not change with how the caller spelled the path (e.g. through a
+  // symlink like `/tmp` -> `/private/tmp` on macOS).
+  projectRoot = await maybeGetRealPathAsync(projectRoot);
   const opts = await normalizeOptionsAsync(projectRoot, options);
   const sources = await getHashSourcesAsync(projectRoot, opts);
   const normalizedSources = sortSources(dedupSources(sources, projectRoot));
--- a/packages/@expo/fingerprint/src/__tests__/Fingerprint-symlink-test.ts
+++ b/packages/@expo/fingerprint/src/__tests__/Fingerprint-symlink-test.ts
@@ -1,0 +1,89 @@
+import fs from 'fs/promises';
+import os from 'os';
+import path from 'path';
+
+import { createFingerprintAsync } from '../Fingerprint';
+
+// This is an integration test against a real project on disk, so opt out of the automatic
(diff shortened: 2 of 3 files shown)


packages/@expo/fingerprint/src/Fingerprint.ts
Kudo, line 29:
```suggestion
const resolvedProjectRoot = await maybeGetRealPathAsync(projectRoot);

const opts = await normalizeOptionsAsync(resolvedProjectRoot, options);
const sources = await getHashSourcesAsync(resolvedProjectRoot, opts);
const normalizedSources = sortSources(dedupSources(sources, resolvedProjectRoot));
const fingerprint = await createFingerprintFromSourcesAsync(normalizedSources, resolvedProjectRoot, opts);
return fingerprint;
```

- `maybeGetRealPathAsync` code is quite straightforward. we don't need comment
- try not to overwrite the argument

packages/@expo/fingerprint/src/__tests__/Fingerprint-symlink-test.ts
```diff
@@ -0,0 +1,89 @@
...
+      // Symlink creation is unsupported on this platform (e.g. Windows without privileges).
+      symlinkSupported = false;
+    }
+  });
+
+  afterAll(async () => {
+    await fs.rm(tmpDir, { recursive: true, force: true });
+  });
+
+  it('produces the same hash for the realpath and a symlinked spelling of the same project', async () => {
```
Kudo, line 68: this is just single test. maybe move it to existing Fingerprint-test.ts? or i'm fine for not having test toward the fix also

PR #49439, opened 2026-08-27
 [docs] Fix markdown content negotiation headers, 404s, and q-values

amandeepmittal:
# Why

Fix ENG-26229 ENG-26230 ENG-26231 ENG-26236

# How

- Add `Vary: Accept` to negotiated markdown responses in `public/_worker.js` so caches key on the `Accept` header (ENG-26229)
- Return a markdown 404 body linking `llms.txt`, `sitemap.xml`, and the docs home instead of a plain `Not found` (ENG-26230)
- Pass 3xx responses through to markdown requests so moved pages redirect instead of returning 404 (ENG-26231)
- Replace the `text/markdown` substring match with an `acceptsMarkdown` q-value parser: markdown is served only when explicitly listed, above zero, and at or above HTML's quality (ENG-26236)
- Cover all four behaviors in `scripts/test-worker.ts`

# Test Plan

`pnpm test:worker` should pass.

# Checklist

### DIFF
--- a/docs/public/_worker.js
+++ b/docs/public/_worker.js
@@ -7,6 +7,35 @@ No markdown exists for this path. Useful starting points:
 - [Documentation home](https://docs.expo.dev/)
 `;
 
+function acceptsMarkdown(accept) {
+  let markdownListed = false;
+  let markdownQuality = 0;
+  let htmlQuality = 0;
+
+  for (const entry of accept.split(",")) {
+    const [type, ...params] = entry.split(";");
+    const name = type.trim().toLowerCase();
+
+    let quality = 1;
+    for (const param of params) {
+      const [key, value] = param.split("=");
+      if (key.trim().toLowerCase() === "q") {
+        quality = Number(value);
+        if (!Number.isFinite(quality)) quality = 0;
+      }
+    }
+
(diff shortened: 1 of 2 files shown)


docs/public/_worker.js
```diff
@@ -40,12 +78,24 @@ export default {
...
+      return new Response(NOT_FOUND_MARKDOWN, {
+        status: 404,
+        headers: {
+          "Content-Type": "text/markdown; charset=utf-8",
+          Vary: "Accept",
+        },
+      });
     }
 
     return env.ASSETS.fetch(request);
```
kadikraman: The RFC says that the origin header should also reply with the  "Vary" header for accurate caching.

- https://www.rfc-editor.org/rfc/rfc9110.html#section-12.5.5
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Vary
- https://www.fastly.com/blog/best-practices-using-vary-header

kadikraman (approved):

docs/public/_worker.js
amandeepmittal: Thanks for the feedback! Added it.

entiendoNull (approved):

PR #49442, opened 2026-08-27
[docs] Move the AgentInstructions block below the H1 in generated markdown

amandeepmittal:
# Why

Fix ENG-26234

# How

Move the `AgentInstructions` block below the H1 in generated markdown so it stays within the boundaries of a Markdown page and passes the agentic tooling tests.

# Test Plan

CleanShot 2026-08-27 at 18 20 49@2x

# Checklist

### DIFF
--- a/docs/scripts/agent-instructions.ts
+++ b/docs/scripts/agent-instructions.ts
@@ -2,8 +2,8 @@
  * Agent feedback instructions appended into per-page .md files.
  *
  * Each page's generated markdown includes an <AgentInstructions> block (placed
- * right after the frontmatter) telling LLM-based agents how to submit feedback
- * with the Expo feedback CLI or direct HTTP fallback.
+ * directly below the page's first H1) telling LLM-based agents how to submit
+ * feedback with the Expo feedback CLI or direct HTTP fallback.
  *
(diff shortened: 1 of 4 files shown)


kadikraman (commented):
This seems a bit odd to me, we have the H1 -> one paragraph -> Agent Instructions -> rest of paragraph.

Maybe it's intentional, but double checking since I was expecting this to be at the bottom of the H1 content, or alternatively the first thing after the H1

Image

amandeepmittal:
> This seems a bit odd to me, we have the H1 -> one paragraph -> Agent Instructions -> rest of paragraph.
> 
> Maybe it's intentional, but double checking since I was expecting this to be at the bottom of the H1 content, or alternatively the first thing after the H1
> 
> Image

@kadikraman my approach behind this was that these agentic checkers complain that there is no H1 before the `<AgentInstructions>` so i decided to put the H1 (page's title) and description before these agent instructions and after agent instructions, the page's actual markup follows.

The second section you've highlighted in red, is the introduction/first paragraph of the page's markup.

Happy to change the order in any other way if this is not satisfactory.

kadikraman (approved):
Got it, thanks for explaining.

PR #49444, opened 2026-08-27
[cli] Resolve the Metro asset registry to react-native core on native platforms

gabrieldonadel:
# Why

Part of the React Native 0.87 upgrade stack, split out of #48770 per review feedback (https://github.com/expo/expo/pull/48770#discussion_r3861755806).

The Expo CLI resolver intercepts imports of `@react-native/assets-registry/registry` and serves a small virtual registry module. Before 0.87 this worked on every platform because React Native's own code imported the same standalone package, so the intercept caught both sides and Metro-registered assets and React Native's `<Image>` shared one registry instance.

React Native 0.87 removed the standalone `@react-native/assets-registry` package. React Native core now bundles its own registry and imports it internally, where the resolver intercept cannot see it. With the always-virtual behavior, native builds end up with two registry instances: Metro's asset modules register into the virtual registry while `<Image>` reads React Native's internal one, so assets fail to resolve at runtime.

# How

- Resolve both `react-native/asset-registry` (the new 0.87 subpath) and the legacy `@react-native/assets-registry/registry` specifier through one intercept.
- On native platforms, resolve to `react-native/asset-registry` so Metro's asset modules and React Native core share the single registry instance inside `react-native`.
- On web, keep serving the virtual registry module, since `react-native` is not bundled there.

# Test Plan

- Added resolver unit tests covering both module specifiers: native resolves to `react-native/asset-registry`, web resolves to the virtual `\0polyfill:assets-registry` module.
- Exercised end to end by the rest of the 0.87 stack: Expo Go and bare-expo builds on the branches above load Metro-registered assets through React Native's `<Image>` on device.

### DIFF
--- a/packages/@expo/cli/src/start/server/metro/withMetroMultiPlatform.ts
+++ b/packages/@expo/cli/src/start/server/metro/withMetroMultiPlatform.ts
@@ -659,6 +659,10 @@ export function withExtendedResolver(
         return getAsyncRequireModule();
       }
 
+      // TODO(@kitten): Revisit the virtual registry approach after the React Native 0.87 upgrade
+      // lands. The virtual module predates the upgrade (introduced by @EvanBacon) and it needs
+      // some testing and sleuthing to establish why it exists instead of resolving react-native's
+      // real registry module, and whether the internal-import capture below can then be dropped.
       // Redirect every asset registry request to the virtual registry module so all consumers
       // share one instance: Metro's generated asset modules (`assetRegistryPath`), imports of
       // `react-native/asset-registry`, and imports of the legacy `@react-native/assets-registry`
--- a/packages/@expo/cli/src/start/server/metro/withMetroMultiPlatform.ts
+++ b/packages/@expo/cli/src/start/server/metro/withMetroMultiPlatform.ts
@@ -658,8 +658,17 @@ export function withExtendedResolver(
       }
 
       // TODO(@kitten): Compare against `config.transformer.assetRegistryPath`
-      if (/^@react-native\/assets-registry\/registry(\.js)?$/.test(moduleName)) {
-        return getAssetRegistryModule();
+      if (
+        moduleName === 'react-native/asset-registry' ||
+        /^@react-native\/assets-registry\/registry(\.js)?$/.test(moduleName)
+      ) {
+        // Serve a virtual registry on web (`react-native` isn't bundled there) and resolve to
+        // react-native's real registry on native so all consumers share one instance. The legacy
+        // `@react-native/assets-registry` package no longer ships with RN 0.87.
+        if (platform === 'web') {
+          return getAssetRegistryModule();
+        }
+        return getStrictResolver(context, platform)('react-native/asset-registry');
       }
 
       if (
--- a/packages/@expo/cli/src/start/server/metro/withMetroMultiPlatform.ts
+++ b/packages/@expo/cli/src/start/server/metro/withMetroMultiPlatform.ts
@@ -49,7 +49,9 @@ export type StrictResolverFactory = (
   platform: string | null
 ) => StrictResolver;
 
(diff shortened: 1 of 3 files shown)


packages/@expo/cli/src/start/server/metro/withMetroMultiPlatform.ts
```diff
@@ -658,8 +658,17 @@ export function withExtendedResolver(
...
+        moduleName === 'react-native/asset-registry' ||
+        /^@react-native\/assets-registry\/registry(\.js)?$/.test(moduleName)
+      ) {
+        // Serve a virtual registry on web (`react-native` isn't bundled there) and resolve to
+        // react-native's real registry on native so all consumers share one instance. The legacy
+        // `@react-native/assets-registry` package no longer ships with RN 0.87.
+        if (platform === 'web') {
+          return getAssetRegistryModule();
+        }
+        return getStrictResolver(context, platform)('react-native/asset-registry');
```
kitten: Sorry in advance to just link to it, but last review comment applies, so linking is probably easiest 🙏 https://github.com/expo/expo/pull/48770#discussion_r3872567458

kitten (commented):
Side-note,
> React Native core now bundles its own registry and imports it internally, where the resolver intercept cannot see it

If we're concerned that a certain module is being imported internally that then has a different request, it probably makes sense to update the resolver to capture that request instead, instead of having a split approach here?

I'm assuming the implication here is that `moduleName === 'react-native/asset-registry` isn't a complete match, meaning, we may have to be more specific?

There's a few approaches here that could work:
- Check which relative imports exist and try to find a generic rule to capture them
- Pre-resolve `assetRegistryPath` (as per the TODO) and/or compare to a pre-resolved hard-coded path, then apply a post-resolution redirect
- Make an additional rule for the above instead, and redirect that module too

Main concern I have basically is that we're changing here to a split model, where sometimes the virtual module is used, which is hard to track, and the original intention (afaik) was to make this consistent. If we want to preserve this, and the above problems apply (internal requires/imports) then I think if we're assuming mixed `react-native/asset-registry` issues may still apply?

If we're confident however that only Web is an issue here, then I think ditching the virtual module entirely, and replacing it just for web, with e.g. a module in `expo` could be preferable?

kitten (approved):
Approving separately, since we could also address this after this stack is merged (if so, let's add a todo comment for this temporarily)

gabrieldonadel:
> Side-note,
> 
> > React Native core now bundles its own registry and imports it internally, where the resolver intercept cannot see it
> 
> If we're concerned that a certain module is being imported internally that then has a different request, it probably makes sense to update the resolver to capture that request instead, instead of having a split approach here?
> 
> I'm assuming the implication here is that `moduleName === 'react-native/asset-registry` isn't a complete match, meaning, we may have to be more specific?
> 
> There's a few approaches here that could work:
> 
> * Check which relative imports exist and try to find a generic rule to capture them
> * Pre-resolve `assetRegistryPath` (as per the TODO) and/or compare to a pre-resolved hard-coded path, then apply a post-resolution redirect
> * Make an additional rule for the above instead, and redirect that module too
> 
> Main concern I have basically is that we're changing here to a split model, where sometimes the virtual module is used, which is hard to track, and the original intention (afaik) was to make this consistent. If we want to preserve this, and the above problems apply (internal requires/imports) then I think if we're assuming mixed `react-native/asset-registry` issues may still apply?
> 
> If we're confident however that only Web is an issue here, then I think ditching the virtual module entirely, and replacing it just for web, with e.g. a module in `expo` could be preferable?

Done, I've updated the intercept to always return the virtual module on all platform, and also compare against config.transformer.assetRegistryPath. 

Also added the TODO to the redirect block pointing back at you, with the context that the virtual module predates the upgrade and needs sleuthing of Evan's original intent.

PR #49448, opened 2026-08-27
[router] Wrap all actions from queue in transitions

Ubax:
# Why

At the moment all actions are dispatched synchronously. Add support for transitions.

This is only an initial implementation and we still need to clarify the best DX/UX.

# How

1. Wrap all queue actions into transition

# Test Plan

1. CI
2. Router e2e

# Checklist

### DIFF
--- a/packages/expo-router/CHANGELOG.md
+++ b/packages/expo-router/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### 🛠 Breaking changes
 
+- Dispatch queued navigation actions in React transitions. The current screen stays visible while the destination suspends, so `SuspenseFallback` no longer renders for navigation-triggered suspense. ([#49448](https://github.com/expo/expo/pull/49448) by [@Ubax](https://github.com/Ubax))
 - Remove `beforeRemove`, `__unsafe_action__`, `PreventRemoveContext`, and `usePreventRemoveContext` from `expo-router/react-navigation`. ([#49408](https://github.com/expo/expo/pull/49408) by [@Ubax](https://github.com/Ubax))
 - Preserve the focused route when switching navigator types in a conditional layout. ([#49297](https://github.com/expo/expo/pull/49297) by [@Ubax](https://github.com/Ubax))
 - Generate deterministic navigation states and route keys. Complete states from custom routers or persisted state must include `routeKeySeq`. ([#49297](https://github.com/expo/expo/pull/49297) by [@Ubax](https://github.com/Ubax))
@@ -45,6 +46,7 @@
 
 ### 🎉 New features
 
+- Add `unstable_useIsNavigating` for observing queued or pending navigation. ([#49448](https://github.com/expo/expo/pull/49448) by [@Ubax](https://github.com/Ubax))
 - Add unstable `NavigationAwareActivity` component. ([#49164](https://github.com/expo/expo/pull/49164) by [@Ubax](https://github.com/Ubax))
 - Add screen error boundaries ([#49174](https://github.com/expo/expo/pull/49174) by [@Ubax](https://github.com/Ubax))
 - Export `NativeStackView` for custom navigator implementations on web. ([#49204](https://github.com/expo/expo/pull/49204) by [@Ubax](https://github.com/Ubax))
--- a/packages/expo-router/src/global-state/RoutingQueueDrainer.tsx
+++ b/packages/expo-router/src/global-state/RoutingQueueDrainer.tsx
@@ -12,7 +12,7 @@ type Props = {
 
 export function RoutingQueueDrainer({ ready, processIntent }: Props) {
   const intents = React.use(PendingIntentsContext);
-  const { dequeue } = React.use(RoutingQueueApiContext)!;
+  const { dequeue, startTransition } = React.use(RoutingQueueApiContext)!;
   const lastProcessed = React.useRef<RoutingIntent[] | undefined>(undefined);
 
   React.useEffect(() => {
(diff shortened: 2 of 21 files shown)


packages/expo-router/src/global-state/RoutingQueueDrainer.tsx
```diff
@@ -21,22 +21,27 @@ export function RoutingQueueDrainer({ ready, processIntent }: Props) {
...
-      try {
-        intent.onDispatch?.(intent.metadata);
-        processIntent(intent);
-      } catch (error) {
-        const message =
-          typeof error === 'object' && error != null && 'message' in error ? error.message : error;
-        console.warn(
-          `An error occurred when trying to handle navigation action ${JSON.stringify(intent)}: ${message}`
-        );
+    // TODO: Design fallback UX for navigations that suspend instead of relying on raw transitions.
```
jakub-agent: [🚨 **critical**] This TODO understates the impact. Wrapping the dispatch in a transition means a navigation whose destination suspends no longer renders `SuspenseFallback` — the new test asserts exactly that (`expect(screen.queryByTestId('fallback')).toBeNull()`).

`docs/pages/router/error-handling.mdx` currently promises the opposite: a layout's `SuspenseFallback` shows "while any child route is suspended". The only listed limitation is async routes. After this change the fallback effectively never renders for the case it was built for — navigating to a route that suspends.

It also affects dev DX on web: async routes suspend while bundling, and the built-in `SuspenseFallback` renders the `Bundling...` toast. With the transition in place, tapping a link to a not-yet-bundled route keeps the old screen with no feedback, which reads as a frozen app.

Could you either land the fallback UX in this PR, or add a CHANGELOG line and a docs update stating the new behavior? Shipping a silent regression of a documented feature is the part I would not merge as-is.

packages/expo-router/src/global-state/RoutingQueueDrainer.tsx
```diff
@@ -21,22 +21,27 @@ export function RoutingQueueDrainer({ ready, processIntent }: Props) {
...
+    startTransition(() => {
+      dequeue(intents);
+      for (const intent of intents) {
+        // Only catches errors thrown while dispatching. The navigation reducer runs
+        // during the next render, so errors from it surface there, not here.
+        try {
+          intent.onDispatch?.(intent.metadata);
```
jakub-agent, line 37: [🏗️ **design**] `onDispatch` is not a pure notification — on web it is where `useLinking` records the history operation into `pendingHistoryOperationsRef`, and the state-change effect later consumes it by matching `operation.path` against the committed path (`src/fork/useLinking.ts`). If no pending operation matches, it calls `.shift()` and drops the head entry.

Moving this into a transition splits the two halves in time: the operation is recorded synchronously, but the commit that should consume it is now deferred and interruptible. Your own test shows a `dispatchSync` landing in between, which produces a commit whose path will not match the queued operation.

The existing `TODO(@ubax)`/ENG-22046 note in `useLinking.ts` already flags this correlation as fragile; the transition makes the interleaving reachable. All four new tests are `.ios.tsx`, so nothing covers it. Could you add a `.web.tsx` test that pushes to a suspending route, interleaves an urgent dispatch, and asserts the resulting history entries?

packages/expo-router/src/global-state/useIsNavigating.ts
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * Returns whether a navigation is queued or its state update is pending.
+ *
+ * The current screen can remain visible while this returns `true` if the destination suspends.
+ * Returns `false` when called outside an Expo Router root.
+ *
+ * @experimental
+ */
+export function useIsNavigating(): boolean {
+  return use(PendingIntentsContext).length > 0 || use(NavigationPendingContext);
```
jakub-agent: [💡 suggestion] Two context reads can collapse into one. Compute the boolean where both inputs already live:

```tsx
// routingQueueContext.tsx
<NavigationPendingContext.Provider value={isPending || queue.length > 0}>
```

and then `useIsNavigating` is just `use(NavigationPendingContext)`. That drops one provider layer, removes the reliance on `||` short-circuiting to skip a `use()` call (legal for `use`, but subtle enough that a reader has to stop and check), and stops consumers re-rendering on every `queue` array identity change while the boolean is unchanged.

Separately: the JSDoc reads as if it covers all navigation, but `navigation.dispatchSync` bypasses the queue and native gestures never enter it. Worth saying so, since `dispatchSync` is public and your third test proves it interleaves.

packages/expo-router/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### 🛠 Breaking changes
 
+- Dispatch queued navigation actions in React transitions and expose `unstable_useIsNavigating` for observing pending navigation.
```
jakub-agent: [🏗️ **design**] Three things here:

1. Missing the `([#49448](...)) by [@Ubax](...)` suffix that every other entry in this section carries.
2. The entry bundles a breaking behavior change with a new export. `unstable_useIsNavigating` is additive and belongs under `🎉 New features`.
3. It omits the change users will actually notice: `SuspenseFallback` no longer renders for navigation-triggered suspense. That is the breaking part and it needs its own line.

packages/expo-router/src/link/Link.tsx
```diff
@@ -33,6 +33,7 @@ export const Link = Object.assign(
    * ```
    */
   function Link(props: LinkProps) {
+    // TODO: Expose pending status scoped to this link's navigation.
```
jakub-agent: [🐧 nit] This and the two other new `TODO`s (`imperative-api.tsx`, `BaseNavigationContainer.tsx`) drop the owner. The package convention is `TODO(@ubax):`, often with a Linear link — the line directly above your new one in `BaseNavigationContainer.tsx` is `// TODO(@ubax):`. Unowned TODOs about unbuilt public API tend to outlive the person who wrote them.

Also, this one sits inside the render body of a component that only re-exports `ExpoLink`. Above the component, or in a tracking issue, would carry the same information without reading as a note about the returned element.

jakub-agent (changes requested):

PR 3/3 in the stack (base: `@ubax/fix-onActionDispatched-and-removal-prevention`, #49408). The mechanism is small and well placed: one `useTransition` owned by `RoutingQueueProvider`, the drainer wraps `dequeue` + the dispatch loop in it, and pending state is published through context. I ran the full `expo-router` suite on this commit — everything passes, including the four new tests.

My concern is not the implementation, it is what ships alongside it.

**Navigation-triggered `SuspenseFallback` stops rendering.** The new test asserts `expect(screen.queryByTestId('fallback')).toBeNull()` while the destination suspends. That is correct transition behavior, but it contradicts the shipped docs, which say a `SuspenseFallback` export shows "while any child route is suspended" (`docs/pages/router/error-handling.mdx`). It also removes the dev-time `Bundling...` toast for web async routes, so tapping a link to a not-yet-bundled route now looks like a frozen screen. The in-code `TODO` acknowledges the gap, but the CHANGELOG does not, so users get no warning.

**Web behavior is untested.** `intent.onDispatch` is the hook `useLinking` uses to correlate a browser-history operation with the state commit it produced. That callback now runs inside a transition, so the operation is recorded early while its commit lands late. All four new tests are `.ios.tsx`.

**Smaller items:** the two-context read in `useIsNavigating` can collapse to one derived boolean, the CHANGELOG entry is missing the PR link and author that every neighboring entry has, and the three new `TODO` comments drop the `TODO(@ubax)` owner convention used elsewhere in the package.

Not approving: the fallback behavior change is user-visible and currently undocumented. Everything else is minor.

packages/expo-router/src/global-state/RoutingQueueDrainer.tsx
Ubax: Make the TODO more descriptive. Propose the content for linear ticket as well

packages/expo-router/src/global-state/RoutingQueueDrainer.tsx
Ubax: Keep as is for now, but add todo to https://linear.app/expo/issue/ENG-22046

packages/expo-router/src/global-state/useIsNavigating.ts
Ubax: Do both

packages/expo-router/src/link/Link.tsx
Ubax: Add `@ubax` to every todo in this PR

packages/expo-router/CHANGELOG.md
Ubax: Add PR link and suspense info to this entry.

jakub-agent (approved):

PR #43635, opened 2026-03-04
fix(expo-updates): load @expo/env before evaluating Expo config in build utilities

leeseonghyun21:
# Why

Native build phase scripts (Xcode build phase / Gradle plugin) invoke `createUpdatesResources.js` directly as a standalone Node.js process, bypassing the Expo CLI. The Expo CLI normally loads `@expo/env` before evaluating the Expo config, but this entry point does not — causing `.env` file variables to be `undefined` when `app.config.ts` is evaluated during native builds.

This breaks any project that references `process.env.*` in their Expo config:

```
Error reading Expo config at /path/to/project/app.config.ts:
Invalid app stage: undefined
    at getConfig (...)
    at createFingerprintForBuildAsync (.../expo-updates/utils/build/createFingerprintForBuildAsync.js:25:52)
    at .../expo-updates/utils/build/createUpdatesResources.js:30:77
```

The issue only occurs when building directly from Xcode or Gradle (not via `npx expo run:ios/android`), because the CLI entry points all call `loadEnvFiles()` before config evaluation.

# How

Add `require('@expo/env').load(possibleProjectRoot)` in `createUpdatesResources.ts` (the entry point for native build scripts) before any `getConfig()` calls downstream.

This follows the same pattern used by:
- `@expo/fingerprint`'s `ExpoConfigLoader.ts` — loads the project env before `getConfig()`
- All 14+ `@expo/cli` command entry points — `loadEnvFiles(projectRoot)` before `getConfig()`

> **Correction.** An earlier revision of this description cited `ExpoConfigLoader.ts` as calling `require('@expo/env').load(projectRoot)`. That is the published `@expo/fingerprint@0.20.10` build, not `main` — at `6d84a5d` the file calls `consumeConfigEnvMode()` followed by `loadProjectEnv(projectRoot, { mode })`. The deprecated `load()` in the diff below is being reworked into `loadProjectEnv` with an explicit mode; see https://github.com/expo/expo/pull/43635#issuecomment-5435987422 for the two shapes awaiting a maintainer decision.

# Test Plan

- Build an app that uses `process.env.*` in `app.config.ts` with a `.env` file
- Build directly from Xcode (Product → Build) — should no longer fail with undefined env vars
- Build via `npx expo run:ios` — should continue to work as before

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #49452, opened 2026-08-27
fix(expo-updates): Use correct env mode in native builds

ramonclaudio:
Supersedes [#43635](https://github.com/expo/expo/pull/43635).

# Why

Updates can read app config with the wrong env values during native builds.

# How

I updated the iOS and Android build steps to set development or production mode before loading env files and reading app config.

# Test Plan

All tests pass.


ramonclaudio:
@expo-bot verify --fable

ramonclaudio:
@expo-bot verify

kitten (commented):
Main question that I think is obviously the thing someone who knows more about updates has to answer is, practically, what is this supposed to do in development and what purpose do the outputs serve in a development build?

packages/expo-updates/scripts/create-updates-resources-ios.sh
```diff
@@ -4,10 +4,28 @@ set -eo pipefail
 
 CREATE_UPDATES_RESOURCES_MODE="all"
 
+is_debug_configuration() {
+  [[ "$1" =~ [Dd][Ee][Bb][Uu][Gg] ]]
```
Kudo: **[bug]** `is_debug_configuration` matches any case of the substring `debug`, and skip-bundling now follows `METRO_DEV` instead of the old `*Debug*` glob. `react-native-xcode.sh` and the Xcode template still use case-sensitive `*Debug*` to decide whether to embed the JS bundle. A configuration such as `debugStaging` (the new iOS test case) previously generated `app.manifest` (`*Debug*` did not match, so mode stayed `all`) while RN embedded a production bundle. After this change the updates script skips the manifest (`only-fingerprint`) while RN still embeds JS, so a launch can miss the embedded update. The iOS tests always set `FORCE_BUNDLING=1`, so they never catch this skip-bundling split.

**Suggestion:** Keep skip-bundling on the same `[[ "$CONFIGURATION" == *Debug* ]]` test that RN and the Xcode template use. If Metro `dev` / config mode should be case-insensitive, compute that separately and do not reuse it to skip `app.manifest`.

packages/expo-updates/utils/src/createManifestForBuildAsync.ts
```diff
@@ -30,7 +31,7 @@ export async function createManifestForBuildAsync(
     platform,
     entryFile,
     minify: false,
-    dev: process.env.CONFIGURATION === 'Debug', // ensures debug assets packaged correctly for iOS and native debug
+    dev, // ensures debug assets packaged correctly for iOS and native debug
```
Kudo, line 34: **[suggestion]** The leftover comment still says this flag "ensures debug assets packaged correctly for iOS and native debug". `dev` is now an explicit argument. On Android, native debug empties `debuggableVariants` and the task passes `debuggableVariant` (false) as Metro `dev`, so native-debug Android does not get debug assets from this path.

**Suggestion:** Drop the comment, or describe the actual contract: callers pass Metro `dev` independently of config mode.

ramonclaudio:
@expo-bot verify

ramonclaudio:
@expo-bot verify

packages/expo-updates/expo-updates-gradle-plugin/src/main/kotlin/expo/modules/updates/ExpoUpdatesPlugin.kt
```diff
@@ -118,6 +134,26 @@ abstract class ExpoUpdatesPlugin : Plugin<Project> {
...
+internal fun getConfigMode(
+  inheritedMode: String?,
+  isDevelopmentBuild: Boolean
+): String = when {
+  !inheritedMode.isNullOrEmpty() -> inheritedMode
+  isDevelopmentBuild -> "development"
+  else -> "production"
+}
+
+internal fun isDevelopmentBuild(
```
Kudo, line 146:
```suggestion
internal fun isNativeDebugBuild(
```

the `isDevelopmentBuild` name is confusing with https://docs.expo.dev/develop/development-builds/introduction/

packages/expo-updates/expo-updates-gradle-plugin/src/main/kotlin/expo/modules/updates/ExpoUpdatesPlugin.kt
```diff
@@ -118,6 +134,26 @@ abstract class ExpoUpdatesPlugin : Plugin<Project> {
...
+  buildType: String?,
+  isDebuggableVariant: Boolean,
+  nativeDebuggingEnabled: Boolean
+): Boolean {
+  if (!nativeDebuggingEnabled) {
+    return isDebuggableVariant
+  }
+  return buildType == "debug" || buildType == "debugOptimized"
```
Kudo, line 154: i wonder if nativeDebuggingEnabled is good enough. if people intent to have debug/debugOptimized as non debuggable, that feels not quite normal and not something we would like to support

Kudo (approved):
imo, it is fine the implementation is different between android and ios, given gradle and xcode is different. on android we use debuggable flag and on ios we use configuration name check, that's good enough.

ramonclaudio:
@expo-bot verify

packages/expo-updates/expo-updates-gradle-plugin/src/main/kotlin/expo/modules/updates/ExpoUpdatesPlugin.kt
```diff
@@ -25,26 +34,58 @@ abstract class ExpoUpdatesPlugin : Plugin<Project> {
...
+    }
+
     androidComponents.onVariants(androidComponents.selector().all()) { variant ->
       val targetName = variant.name.replaceFirstChar { if (it.isLowerCase()) it.titlecase(Locale.ROOT) else it.toString() }
-      val projectRoot = project.rootProject.projectDir.parentFile.toPath()
-      val isDebuggableVariant =
-        reactExtension.debuggableVariants.get().any { it.equals(variant.name, ignoreCase = true) }
+      val isDebuggableVariant = isDebuggableVariant(variant.name, reactExtension.debuggableVariants.get())
+      val configMode = getConfigMode(
+        inheritedMode = System.getenv("__EXPO_CONFIG_MODE"),
```
Kudo: **[suggestion]** `getConfigMode` takes `System.getenv("__EXPO_CONFIG_MODE")` and stores the result as a concrete task input. The same block already uses `project.providers.environmentVariable` / `environmentVariablesPrefixedBy` for `__EXPO_ENV_LOADED`, `EXPO_*`, `PATH`, and friends. `System.getenv` sees the Gradle daemon environment, not the client environment that started the build, and configuration cache will freeze whatever value was present when the daemon started. A later `__EXPO_CONFIG_MODE=production ./gradlew assembleDebug` can keep using the variant default (`development`) while the other new env inputs update.

**Suggestion:** Resolve the override with `project.providers.environmentVariable("__EXPO_CONFIG_MODE")` and set `configMode` from that provider (with `orElse` for the variant default) so daemon runs and configuration cache observe the same value as the other env inputs.

ramonclaudio:
@expo-bot verify --fable

packages/expo-updates/scripts/create-updates-resources-ios.sh
```diff
@@ -50,5 +75,4 @@ else
   exit 1
 fi
 
-"${EXPO_UPDATES_PACKAGE_DIR}/scripts/with-node.sh" "${EXPO_UPDATES_PACKAGE_DIR}/utils/build/createUpdatesResources.js" ios "$PROJECT_ROOT" "$RESOURCE_DEST" "$CREATE_UPDATES_RESOURCES_MODE" "$ENTRY_FILE"
-
+__EXPO_CONFIG_MODE="$CONFIG_MODE" "${EXPO_UPDATES_PACKAGE_DIR}/scripts/with-node.sh" "${EXPO_UPDATES_PACKAGE_DIR}/utils/build/createUpdatesResources.js" ios "$PROJECT_ROOT" "$RESOURCE_DEST" "$CREATE_UPDATES_RESOURCES_MODE" "$ENTRY_FILE" "$METRO_DEV"
```
kitten: Main concern here, we seem to accept both `__EXPO_CONFIG_MODE` and `METRO_DEV` here, and I'm not sure if we need both? They seem to be coupled above, then overridden. Ergo, users can't (and shouldn't) override them anyway.

Doesn't this mean we should either:
- choose one of the two (probably the arg)
- or, alternatively, scrap both and pass this as `NODE_ENV` by convention?

packages/expo-updates/scripts/create-updates-resources-ios.sh
```diff
@@ -31,6 +18,44 @@ if [ "x$PROJECT_DIR_BASENAME" != "xPods" ]; then
...
+if [[ "$CONFIGURATION" == *Debug* ]]; then
+  CONFIG_MODE="development"
+  METRO_DEV="true"
+else
+  CONFIG_MODE="production"
+  METRO_DEV="false"
+fi
+
+if [[ -n "${__EXPO_CONFIG_MODE+x}" ]]; then
+  CONFIG_MODE="$__EXPO_CONFIG_MODE"
```
kitten: (see below comment first)

This does seem to align between the two. There's a bit of a question whether we'd allow `NODE_ENV` to be the input (and output) here. If there's a preexisting conflict (which I don't think there is as this may have naturally passed on before?) that's fine. Mainly concerned about the separation of the two here.

It feels like `METRO_DEV` and `CONFIG_MODE` should do the same thing?

The other concern here I have is that if we use `__EXPO_CONFIG_MODE` as the pass-through below, doesn't this mean this will likely be entirely different for fingerprint or risk overriding that too, which shouldn't happen, assume that one does the same as `expo config` and accepts it?

Previously, I've assumed this was only an internal flag for EAS Build for `expo config`, so that's the main confusion I have here

PR #49457, opened 2026-08-27
[ios][updates] Repair ready updates that are missing their launch asset

alanjhughes:
# Why
The previous PR stops inserting bad row, but devices that already have a ready update row with no launch asset stay broken forever. The row is selected on every launch, the launch fails, and nothing ever fixes it

# How
launchableUpdates already reads every candidate row on launch, so the repair lives there at no extra cost to healthy devices. A ready row with no launch asset is demoted to pending and excluded from the candidates in the same pass. The launcher then falls back to the embedded update cleanly on that boot, and the existing loader path re-registers the demoted row in the background.

# Test Plan
Two tests: selection demotes a broken ready row to pending and excludes it, and a healthy ready row passes through untouched.

### DIFF
--- a/packages/expo-updates/ios/EXUpdates/Database/UpdatesDatabase.swift
+++ b/packages/expo-updates/ios/EXUpdates/Database/UpdatesDatabase.swift
@@ -444,7 +444,23 @@ public final class UpdatesDatabase: NSObject {
     )
 
     let rows = try execute(sql: sql, withArgs: [config.scopeKey])
-    return rows.map { row in
+
+    // A ready row with no launch asset is corrupt (e.g. from an interrupted registration) and
+    // would fail every launch. Demote it for the loader to retry instead of offering it.
+    var launchableRows: [[String: Any?]] = []
+    for row in rows {
+      let status: NSNumber = row.requiredValue(forKey: "status")
+      let launchAssetId: NSNumber? = row.optionalValue(forKey: "launch_asset_id")
+      if status.intValue == UpdateStatus.StatusReady.rawValue && launchAssetId == nil {
+        let updateId: UUID = row.requiredValue(forKey: "id")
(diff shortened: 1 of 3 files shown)


packages/expo-updates/ios/EXUpdates/Database/UpdatesDatabase.swift
```diff
@@ -94,6 +94,17 @@ public final class UpdatesDatabase: NSObject {
   public func openDatabase(inDirectory directory: URL, logger: UpdatesLogger) throws {
     dispatchPrecondition(condition: .onQueue(databaseQueue))
     db = try UpdatesDatabaseInitialization.initializeDatabaseWithLatestSchema(inDirectory: directory, logger: logger)
+
+    // A ready update always has a launch asset. A row missing one is corrupt (e.g. from an
+    // interrupted registration) and would fail every launch, so demote it for the loader to retry.
+    // Resetting the failed launch count keeps the retried row selectable. The repair is
+    // best-effort and must not take the database down with it.
+    do {
+      let repairSql = "UPDATE updates SET status = ?1, failed_launch_count = 0 WHERE status = ?2 AND launch_asset_id IS NULL;"
```
Kudo: not sure if we have the other proper callsite than here? since we will run this this statement from every launch. ideally we only run the statement when we run into the "missing launch asset" case

Kudo (changes requested):
would be good to consider a different place to run the statement

packages/expo-updates/ios/EXUpdates/Database/UpdatesDatabase.swift
alanjhughes: Make sense, launchableUpdates is better because we already have the rows so there's no extra cost unless we have to repair a row

Kudo (approved):

PR #49458, opened 2026-08-27
[ios][updates] Register a downloaded update's assets and ready status in a single transaction

alanjhughes:
# Why
Registration used to be three separate commits: one per existing asset link, one batch for new assets, and one for the ready status. Any interruption between them left partial state behind. Worse, a failed existing-asset link was logged and swallowed, and an unreadable asset file was dropped by a release-mode no-op assert, so an update could still be marked ready without its launch asset. This mirrors what #49130 fixed for embedded registration on Android.

# How
 A new finishUpdateRegistration runs the existing links, the new asset batch, and the finished status in one transaction. The transaction helper checks the result codes of BEGIN and COMMIT, which the old inline code ignored, so a failed commit can no longer report success. Before committing, a guard verifies the update actually has a launch asset and refuses to mark it ready otherwise.

# Test Plan
Three tests on the new method

### DIFF
--- a/packages/expo-updates/ios/EXUpdates/AppLoader/AppLoader.swift
+++ b/packages/expo-updates/ios/EXUpdates/AppLoader/AppLoader.swift
@@ -358,18 +358,29 @@ open class AppLoader: NSObject {
     database.databaseQueue.async {
       self.arrayLock.lock()
 
+      guard let updateResponse = self.updateResponseContainingManifest,
+        let updateManifest = updateResponse.manifestUpdateResponsePart?.updateManifest else {
+        self.arrayLock.unlock()
+        self.finish(withError: UpdatesError.appLoaderFinishedWithoutManifest)
+        return
+      }
+
+      var assetsToLink: [UpdateAsset] = []
       for existingAsset in self.existingAssets {
-        var existingAssetFound: Bool = false
+        let existingAssetFound: Bool
         do {
-          existingAssetFound = try self.database.addExistingAsset(
-            existingAsset,
-            toUpdateWithId: self.updateResponseContainingManifest!.manifestUpdateResponsePart!.updateManifest.updateId
-          )
+          existingAssetFound = try self.database.asset(withKey: existingAsset.key) != nil
         } catch {
-          self.logger.warn(message: "Error searching for existing asset in DB: \(error.localizedDescription)")
+          // treating a failed check as "not found" would re-insert an existing asset, and the
+          // key-conflicting replace cascade-deletes every update that uses the old row
+          self.arrayLock.unlock()
+          self.finish(withError: UpdatesError.appLoaderUnknownError(cause: error))
+          return
         }
 
-        if !existingAssetFound {
+        if existingAssetFound {
+          assetsToLink.append(existingAsset)
+        } else {
           // the database and filesystem have gotten out of sync
           // do our best to create a new entry for this file even though it already existed on disk
           // TODO: we should probably get rid of this assumption that if an asset exists on disk with the same filename, it's the same asset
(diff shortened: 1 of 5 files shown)


packages/expo-updates/ios/EXUpdates/AppLoader/AppLoader.swift
```diff
@@ -358,18 +358,29 @@ open class AppLoader: NSObject {
...
+        let existingAssetFound: Bool
         do {
-          existingAssetFound = try self.database.addExistingAsset(
-            existingAsset,
-            toUpdateWithId: self.updateResponseContainingManifest!.manifestUpdateResponsePart!.updateManifest.updateId
-          )
+          existingAssetFound = try self.database.asset(withKey: existingAsset.key) != nil
         } catch {
-          self.logger.warn(message: "Error searching for existing asset in DB: \(error.localizedDescription)")
+          // treating a failed check as "not found" would re-insert an existing asset, and the
```
Kudo, line 374: [🏗️ **design**] The other route into `finishedAssets` still hits this cascade. In `startLoading`, an asset that has a database row but whose file is missing on disk takes the `else` branch of the `fileExists` check and goes to `downloadAsset`, so it lands in `finishedAssets`. `addNewAssetsInternal` then runs `INSERT OR REPLACE INTO assets` on a key that already exists. `updates.launch_asset_id` is `ON DELETE CASCADE` and `foreign_keys` is `ON`, so that replace deletes the update row currently launching from the old asset.

This is pre-existing and the new transaction does not make it worse. Is closing it part of this stack, or a follow-up?

packages/expo-updates/ios/Tests/UpdatesDatabaseTests.swift
```diff
@@ -261,14 +298,183 @@ class UpdatesDatabaseTests {
...
+            BEGIN SELECT RAISE(ABORT, 'injected failure'); END;
+          """,
+          withArgs: nil
+        )
+
+        do {
+          try db.finishUpdateRegistration(
+            update2,
+            newAssets: [makeAsset(key: "asset-b", isLaunchAsset: true)],
+            existingAssets: [makeAsset(key: "asset-a")],
```
Kudo, line 415: [💡 suggestion] The bare `catch {}` on the next line accepts any error, so this test stays green if the new `SELECT 1 FROM updates ... launch_asset_id IS NOT NULL` breaks. A typo in that statement throws too, and the rollback assertions below still pass. This test is the only guard on that SQL.

The file already does typed catches elsewhere (`SetExtraClientParamsTests.validates`):

```swift
} catch UpdatesDatabaseError.finishedUpdateMissingLaunchAsset {
} catch {
  Issue.record("Expected finishedUpdateMissingLaunchAsset but got \(error)")
}
```

Same applies to the other two new tests, where `transactionBeginError` is the error worth pinning.

packages/expo-updates/ios/EXUpdates/Database/UpdatesDatabase.swift
```diff
@@ -106,6 +115,24 @@ public final class UpdatesDatabase: NSObject {
     return try UpdatesDatabaseUtils.execute(sql: sql, withArgs: args, onDatabase: db.require("Missing database handle"))
   }
 
+  private func inTransaction(_ block: () throws -> Void) throws {
```
Kudo:
```suggestion
  private func withTransaction(_ block: () throws -> Void) throws {
```

i like this with prefix more in this case 😅

Kudo (approved):
cannot catch something by my eye. just ask ai review. please use your agent to review if there's anything meaningful

PR #49459, opened 2026-08-27
[ios][updates] Fail with a launch asset not found error instead of silently or hanging

alanjhughes:
# Why
iOS had no equivalent of Android's "Launch asset not found" error. A launchable update with no linked launch asset completed with a nil error and a false success flag, and an update with zero linked assets never invoked its completion at all, leaving the app on the splash screen forever. In production this surfaced as a generic message that pointed at nothing.

# How
Both paths now produce appLauncherLaunchAssetNotFound carrying the update id: the empty asset list completes immediately with the error, and a completed pass with no launch asset URL substitutes it for the nil error. The new checks sit after the embedded and development early returns, so no-copy embedded launches are unaffected, per the iOS caveat raised on #48733. With the repair in the previous PR filtering broken rows at selection, this guard is the backstop for states that arise after selection.

# Test Plan
Two launcher tests run the real ensureAllAssetsExist against a seeded database, one for the hang and one for the silent failure

### DIFF
--- a/packages/expo-updates/ios/EXUpdates/AppLauncher/AppLauncherWithDatabase.swift
+++ b/packages/expo-updates/ios/EXUpdates/AppLauncher/AppLauncherWithDatabase.swift
@@ -193,7 +193,11 @@ public class AppLauncherWithDatabase: NSObject, AppLauncher {
       )
 
       completionQueue.async {
-        self.completion!(self.launchAssetError, self.launchAssetUrl != nil)
+        var error = self.launchAssetError
+        if error == nil && self.launchAssetUrl == nil {
+          error = UpdatesError.appLauncherLaunchAssetNotFound(updateId: launchedUpdate.updateId)
+        }
+        self.completion!(error, self.launchAssetUrl != nil)
         self.completion = nil
       }
       return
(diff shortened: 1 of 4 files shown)


packages/expo-updates/ios/EXUpdates/AppLauncher/AppLauncherWithDatabase.swift
```diff
@@ -211,6 +211,16 @@ public class AppLauncherWithDatabase: NSObject, AppLauncher {
     self.assetFilesMap = UpdatesUtils.embeddedAssetsMap(withConfig: config, database: database, logger: logger)
 
     let assets = launchedUpdate.assets()!
+    if assets.isEmpty {
```
Kudo, line 218: [🏗️ **design**] The embedded-status branch above (line 196) still has the shape you're fixing here. `launchAssetError` is only ever written at line 263, inside the `downloadAsset` callback, which is reachable only from the DB-asset loop below. So at line 196 it is always `nil`, and when `updatesBundle.url(forResource:)` finds no embedded bundle that branch completes `(nil, false)` — the exact case `RelaunchProcedure.swift:94` force-unwraps with `cause: error!`, so a relaunch traps instead of reporting.

That path also matters more since embedded copying became opt-out: `markUpdateFinished` now keeps embedded updates at `StatusEmbedded`, so line 196 is the normal launch path for the embedded update. Worth giving it the same guard.

packages/expo-updates/ios/Tests/AppLauncherWithDatabaseTests.swift
```diff
@@ -116,4 +125,125 @@ class AppLauncherWithDatabaseTests {
...
+        }
+      }
+    }
+
+    let once = Once()
+    return await withCheckedContinuation { continuation in
+      launcher.launchUpdate(withSelectionPolicy: SelectionPolicyFactory.filterAwarePolicy(withRuntimeVersion: "1", config: config)) { error, success in
+        once.run { continuation.resume(returning: (error, success)) }
+      }
+      DispatchQueue.global().asyncAfter(deadline: .now() + 2.0) {
```
Kudo: [💡 suggestion] 2s is tight for a wall-clock deadline on loaded CI. The deadline only decides how long a *regressed* launcher hangs before the test fails — the passing path resolves as soon as the completion fires — so a longer value costs nothing and removes a flake vector.

```suggestion
      DispatchQueue.global().asyncAfter(deadline: .now() + 10.0) {
```

Kudo (approved):

PR #49460, opened 2026-08-27
[ios][updates] Preserve the launcher failure as the emergency launch reason

alanjhughes:
# Why
On a cold start where the cached update fails to launch and the remote check finds nothing new, AppLoaderTask logged the real failure and then reported the generic "unexpected error" to the delegate. Crash reporting therefore showed one indistinguishable message for completely different failure modes

# How
A fallbackLaunchError property defaults to the generic error and is upgraded to the wrapped launcher failure when the task defers to the remote check. finish reports the explicit error when there is one, then the fallback. The delegate, and therefore the emergency launch reason in crash reporting, now carries the real cause.

# Test Plan
A new AppLoaderTaskTests suite simulates the full cold-start

### DIFF
--- a/packages/expo-updates/ios/EXUpdates/AppLoader/AppLoaderTask.swift
+++ b/packages/expo-updates/ios/EXUpdates/AppLoader/AppLoaderTask.swift
@@ -104,7 +104,7 @@ public enum BackgroundUpdateStatus: Int {
  */
 @objc(EXUpdatesAppLoaderTask)
 @objcMembers
-public final class AppLoaderTask: NSObject {
+public class AppLoaderTask: NSObject {
   public weak var delegate: AppLoaderTaskDelegate?
   public weak var swiftDelegate: AppLoaderTaskSwiftDelegate?
 
@@ -126,6 +126,7 @@ public final class AppLoaderTask: NSObject {
   private var isTimerFinished: Bool
   private var hasLaunched: Bool
   private var isUpToDate: Bool
+  private var fallbackLaunchError: UpdatesError = .appLoaderTaskUnexpectedErrorDuringLaunch
   private let loaderTaskQueue: DispatchQueue
 
   public required init(
@@ -166,10 +167,13 @@ public final class AppLoaderTask: NSObject {
     loadEmbeddedUpdate {
       self.launch { error, success in
         if !success {
+          // keep the cause so a launch that finishes without a specific error reports
+          // this failure instead of the generic error
+          let cause = UpdatesError.appLoaderTaskFailedToLaunch(cause: error)
+          self.fallbackLaunchError = cause
           if !shouldCheckForUpdate {
             self.finish(withError: error)
           }
-          let cause = UpdatesError.appLoaderTaskFailedToLaunch(cause: error)
           self.logger.error(
             cause: cause,
             code: .updateFailedToLoad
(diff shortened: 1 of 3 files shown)


packages/expo-updates/ios/EXUpdates/AppLoader/AppLoaderTask.swift
```diff
@@ -166,10 +167,14 @@ public final class AppLoaderTask: NSObject {
     loadEmbeddedUpdate {
       self.launch { error, success in
         if !success {
+          let cause = UpdatesError.appLoaderTaskFailedToLaunch(cause: error)
           if !shouldCheckForUpdate {
             self.finish(withError: error)
+          } else {
+            // keep the cause so a remote check that finds nothing reports this
+            // failure instead of the generic error
+            self.fallbackLaunchError = cause
```
Kudo: [💡 suggestion] The `else` narrows the fix more than it needs to. `fallbackLaunchError` is only read when `finish` gets a nil error, so setting it unconditionally is behavior-neutral for the `!shouldCheckForUpdate` path and covers one more case: `AppLauncherWithDatabase.ensureAllAssetsExist` completes with `(launchAssetError, launchAssetUrl != nil)` for an embedded update, so a binary whose embedded bundle is missing yields `(nil, false)` — that path still reports the generic error today.

```swift
let cause = UpdatesError.appLoaderTaskFailedToLaunch(cause: error)
self.fallbackLaunchError = cause
if !shouldCheckForUpdate {
  self.finish(withError: error)
}
```

packages/expo-updates/ios/Tests/AppLoaderTaskTests.swift
```diff
@@ -0,0 +1,141 @@
...
+    let once = Once()
+    var strongDelegate: ErrorRecordingDelegate?
+    let receivedError: Error? = await withCheckedContinuation { continuation in
+      let delegate = ErrorRecordingDelegate { error in
+        once.run { continuation.resume(returning: error) }
+      }
+      strongDelegate = delegate
+      task.delegate = delegate
+      task.start()
+      DispatchQueue.global().asyncAfter(deadline: .now() + 2.0) {
```
Kudo: [🐧 nit] 2 seconds of wall clock is a tight budget on a loaded CI runner. The test resumes as soon as the delegate fires, so only the hang case pays for a larger value.

```suggestion
      DispatchQueue.global().asyncAfter(deadline: .now() + 10.0) {
```

packages/expo-updates/ios/Tests/AppLoaderTaskTests.swift
```diff
@@ -0,0 +1,141 @@
...
+    withExtendedLifetime(strongDelegate) {}
+
+    #expect(receivedError != nil, "the loader task never reported an error")
+
+    guard let updatesError = receivedError as? UpdatesError else {
+      Issue.record("Expected an UpdatesError but got \(String(describing: receivedError))")
+      return
+    }
+
+    if case .appLoaderTaskUnexpectedErrorDuringLaunch = updatesError {
```
Kudo: [🐧 nit] This check is subsumed by the one below it — any error that isn't `appLoaderTaskFailedToLaunch`, generic one included, already fails there. One `guard` replaces both blocks and drops the empty `if` body:

```swift
guard case .appLoaderTaskFailedToLaunch = updatesError else {
  Issue.record("Expected appLoaderTaskFailedToLaunch but got \(updatesError)")
  return
}
```

Kudo (approved):

PR #49470, opened 2026-08-27
[android][updates] Repair launchable updates that are missing their launch asset

alanjhughes:
# Why
An update row can end up with status READY but no launch asset row. Registration is now transactional (#49130) so new rows can no longer be committed in this state, but rows corrupted on older SDKs persist through library upgrades, and any future cause would land in the same state. Today such a row is selected for launch, fails with "Launch asset not found for update" on every cold start, and never heals: the loader short-circuits on READY so it never re-registers the assets, and the launcher keeps selecting the same broken row.

This brings back the changes in #48733 by @martintreurnicht, which was closed in favor of #49130. 

# How
- DatabaseLauncher.getLaunchableUpdate skips any non-DEVELOPMENT update whose launch asset row is missing and logs a warning. Selection can then fall back to an older complete update, the embedded update, or nothing, instead of a row that can never launch.
- Loader.processUpdate's READY short-circuit now also requires the launch asset to exist. A broken row falls through to downloadAllAssets, which re-registers the assets and repairs it on the next load pass

The launcher guard is the safety net until repair happens, and the loader guard is the repair path. DEVELOPMENT rows are exempt because they legitimately have no asset rows.

# Test Plan
Added new tests

### DIFF
--- a/packages/expo-updates/android/src/androidTest/java/expo/modules/updates/loader/EmbeddedLoaderTest.kt
+++ b/packages/expo-updates/android/src/androidTest/java/expo/modules/updates/loader/EmbeddedLoaderTest.kt
@@ -292,6 +292,11 @@ class EmbeddedLoaderTest {
     update.status = UpdateStatus.READY
     db.updateDao().insertUpdate(update)
 
+    val launchAsset = AssetEntity("bundle-1234", "js")
+    launchAsset.relativePath = "bundle-1234"
+    launchAsset.isLaunchAsset = true
+    db.assetDao().insertAssets(listOf(launchAsset), update)
+
     val result = loader.load { _ ->
       Loader.OnUpdateResponseLoadedResult(shouldDownloadManifestIfPresentInResponse = true)
     }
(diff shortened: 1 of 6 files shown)


packages/expo-updates/android/src/androidTest/java/expo/modules/updates/loader/EmbeddedLoaderTest.kt
```diff
@@ -302,6 +307,39 @@ class EmbeddedLoaderTest {
...
     Assert.assertEquals(1, updates.size.toLong())
     Assert.assertEquals(UpdateStatus.READY, updates[0].status)
+
+    // short-circuited, so the manifest's assets were never registered
+    Assert.assertEquals(1, db.assetDao().loadAllAssets().size.toLong())
+  }
+
+  @Test
+  @Throws(IOException::class, NoSuchAlgorithmException::class)
+  fun testEmbeddedLoader_UpdateExists_ReadyButMissingLaunchAsset() = runTest {
```
Kudo, line 317: [💡 suggestion] This test starts with an empty `assets` table, so the repair runs through the fresh-download branch of `downloadAllAssets`. A row corrupted the way #49130 describes still has its `assets` and `updates_assets` rows and only a null `launch_asset_id`, which takes the other branch: `mergeAndUpdateAsset` → `existingAssetList` → `addExistingAssetToUpdate(update, asset, asset.isLaunchAsset)`.

That branch works today, but only because `mergeAndUpdateAsset` copies `isLaunchAsset` onto the DB entity — and `isLaunchAsset` is `@Ignore`d, so nothing else carries it. Pre-inserting the launch asset row (without setting `launch_asset_id`) in a second test would pin down the path the real corrupt rows actually take.

Kudo (approved):

PR #49471, opened 2026-08-27
[android][updates] Add a stress test for concurrent asset registration in Loader

alanjhughes:
# Why
#43966 fixed lost asset writes by making Loader's bookkeeping collections thread safe. That race silently dropped assets from updates that were then marked READY, which bricks the app at launch, and nothing in the test suite would catch a regression: the existing loader tests run on a single-threaded test dispatcher, so the download completions can never overlap.

# How
Adds LoaderStressTest, which loads a generated manifest of 100 assets plus the launch asset through a real RemoteLoader on a Dispatchers.IO scope, with small random delays in the mocked downloads so completions genuinely interleave. It runs five iterations against fresh in-memory databases and asserts the update is READY, has its launch asset, and has every asset row registered.

# Test Plan
With #43966's synchronization temporarily reverted to plain collections, the test fails on the first iteration with a ConcurrentModificationException out of notifyAssetLoadProgress. With the synchronization in place, all iterations pass.

### DIFF
--- a/packages/expo-updates/android/src/androidTest/java/expo/modules/updates/loader/LoaderStressTest.kt
+++ b/packages/expo-updates/android/src/androidTest/java/expo/modules/updates/loader/LoaderStressTest.kt
@@ -1,0 +1,144 @@
+package expo.modules.updates.loader
+
+import android.content.Context
+import android.net.Uri
+import androidx.room.Room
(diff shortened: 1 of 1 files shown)


Kudo (approved):
great test 👏

PR #49475, opened 2026-08-27
[ios][widgets] Expose Live Activity content state on instances

coffee-cup:
# Why

I'm using push-to-start Live Activities. A server starts activities remotely and puts identifying data in the props, in my case a session id. Since #48589 the app can enumerate these server-started activities with `getInstances()` and read their ids and push tokens, but not the content they were started with. Props are write-only from JS: `start` and `update` take them, but nothing reads them back. So the app holds a handle to an activity and still has to ask the server what that activity is for.

# How

Adds `getContentState()` to `LiveActivity`. Same lookup as the other instance methods: find the ActivityKit activity by id, then read `activity.content.state`. It returns `null` when the activity is no longer running instead of throwing, because a read that races an ending activity is normal during reconciliation.

```swift
func getContentState() -> [String: Any]? {
  guard let activity = Activity<LiveActivityAttributes>.activities.first(where: { $0.id == id }) else {
    return nil
  }

  let state = activity.content.state
  var contentState: [String: Any] = ["name": state.name]
  if let props = state.props {
    contentState["props"] = props
  }
  return contentState
}
```

`update(props: T)` stringifies props on the way in, so the wrapper parses them on the way out:

```ts
const state = activity.getContentState();
// LiveActivityContentState<T> | null, i.e. { name: string; props: T | null } | null
```

The non-native stub returns `null` like the other stubbed methods.

# Test Plan

- `xcrun swiftc -parse ios/LiveActivity.swift ios/LiveActivityFactory.swift ios/WidgetsModule.swift`
- `pnpm run build && pnpm run typecheck && pnpm run lint && pnpm run test --runInBand` in `packages/expo-widgets` (all pass, 13 existing tests)
- Manually in an app that renders Live Activities with expo-widgets, on an iOS simulator with this diff applied to the installed package:
  1. `factory.start(props)`, then `getContentState()` returns the activity name and the same props.
  2. Force-quit and relaunch, then `getInstances()[0].getContentState()` reads the full content state of an activity the current JS runtime never started. That's the push-to-start situation this method is for.

# Checklist

### DIFF
--- a/packages/expo-widgets/ios/LiveActivity.swift
+++ b/packages/expo-widgets/ios/LiveActivity.swift
@@ -12,17 +12,12 @@ final class LiveActivity: SharedObject {
     super.init()
   }
 
-  func getContentState() -> [String: Any]? {
+  func getContentState() throws -> String? {
     guard let activity = Activity<LiveActivityAttributes>.activities.first(where: { $0.id == id }) else {
-      return nil
+      throw LiveActivityNotFoundException(id)
     }
 
-    let state = activity.content.state
-    var contentState: [String: Any] = ["name": state.name]
-    if let props = state.props {
-      contentState["props"] = props
-    }
-    return contentState
+    return activity.content.state.props
   }
 
   func update(props: String?, staleDate: Date?) async throws {
--- a/packages/expo-widgets/ios/LiveActivity.swift
+++ b/packages/expo-widgets/ios/LiveActivity.swift
@@ -12,6 +12,19 @@ final class LiveActivity: SharedObject {
     super.init()
   }
 
+  func getContentState() -> [String: Any]? {
+    guard let activity = Activity<LiveActivityAttributes>.activities.first(where: { $0.id == id }) else {
+      return nil
+    }
+
+    let state = activity.content.state
+    var contentState: [String: Any] = ["name": state.name]
+    if let props = state.props {
+      contentState["props"] = props
+    }
+    return contentState
+  }
+
   func update(props: String?, staleDate: Date?) async throws {
     guard let activity = Activity<LiveActivityAttributes>.activities.first(where: { $0.id == id }) else {
       throw LiveActivityNotFoundException(id)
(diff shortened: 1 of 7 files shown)


packages/expo-widgets/ios/LiveActivity.swift
```diff
@@ -12,6 +12,19 @@ final class LiveActivity: SharedObject {
     super.init()
   }
 
+  func getContentState() -> [String: Any]? {
+    guard let activity = Activity<LiveActivityAttributes>.activities.first(where: { $0.id == id }) else {
+      return nil
+    }
```
jakex7:
```suggestion
  func getContentState() throws -> [String: Any] {
    guard let activity = Activity<LiveActivityAttributes>.activities.first(where: { $0.id == id }) else {
      throw LiveActivityNotFoundException(id)
    }
```

packages/expo-widgets/src/Widgets.types.ts
```diff
@@ -271,6 +271,21 @@ export type PushToStartTokenEvent = {
...
 };
 
+/**
+ * The content state of a Live Activity: its name and the props it currently displays.
+ */
+export type LiveActivityContentState<T extends object = object> = {
+  /**
+   * The Live Activity name. Matches the `'name'` field in the widget configuration in the app config.
+   */
+  name: string;
```
jakex7: `LiveActivity` instance is already strictly tied to the `LiveActivity` factory containing the name, so I don't see the point of exposing it here. Also `name` for live activity is not specified in app config.

jakex7 (changes requested):

packages/expo-widgets/src/Widgets.types.ts
jakex7: `LiveActivity` instance is already strictly tied to the `LiveActivity` factory containing the name, so I'm not sure if there is a point of exposing it here. Also `name` for live activity is not specified in app config.

coffee-cup:
I've addressed both comments

jakex7:
@expo-bot review

PR #49483, opened 2026-08-28
[ui] Fix Host + RNHostView matchContents layout loops

### LINKED ISSUE #47883: [ui][ios] MenuView trigger (`matchContents`) keeps stale size when its RN children change — clips content in FlashList-recycled cells

https://github.com/mateuszurbas/expo-ui-menuview-repro

### Summary

On iOS, `MenuView` from `@expo/ui/community/menu` measures its React Native trigger children **only on mount**. The trigger is wrapped in `<RNHostView matchContents>` inside `<Host matchContents>`, and when the trigger's content changes after mount (a `setState`, or FlashList recycling a cell), the SwiftUI host keeps the stale size and **clips** the wider content.

The `key={props.matchContents ? 'matchContents' : 'noMatchContents'}` remount trick in [`RNHostView.tsx`](https://github.com/expo/expo/blob/main/packages/expo-ui/src/swift-ui/RNHostView.tsx) (commented "`matchContents` can only be used once on mount") only covers toggling the prop itself — not the hosted children changing size later. In `RNHostView.swift`, `ApplySizeFromYogaNode` KVO-observes the hosted `UIView`'s bounds, but the Yoga subtree hosted inside the SwiftUI host never grows past the width that was committed on mount, so the observer keeps reporting the stale size.

**Real-world impact:** any `MenuView` trigger inside a recycling list shows clipped content. In our production app, donation-amount badges rendered `999` instead of `999 999 zł` after scrolling or switching the list's sort order, because recycled cells keep the width measured for the previous item:

dynamic badge clipped to previous width vs static control rendering fully

(dynamic badge after a text change: clipped; static control with the same string right below: renders fully)

**Workaround:** force a remount whenever the trigger content changes — `<MenuView key={content} …>`.

### Steps to reproduce

Platform: **iOS** (simulator and device), development build (`npx expo run:ios`) — `@expo/ui` does not run in Expo Go. Package manager: npm.

1. `npm install && npx expo run:ios`
2. On launch, two `MenuView` badges render, both sized correctly: the dynamic one shows `50 zł`, the static control shows `999 999 zł`.
3. Tap **Toggle text** — the dynamic badge's text changes to `999 999 zł`.
4. **Expected:** the badge grows to fit `999 999 zł` (like the static control below it).
   **Actual:** the badge keeps (roughly) the `50 zł` width and the text is clipped to `999 9…`.

Android is not affected (the Android `MenuView` implementation doesn't use the SwiftUI host).

### Environment

```
    System:
      OS: macOS 26.5.2
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 25.6.1 - /opt/homebrew/bin/node
      npm: 11.9.0 - /opt/homebrew/bin/npm
    Managers:
    SDKs:
      iOS SDK:
    IDEs:
    npmPackages:
      expo: ~57.0.6 => 57.0.6
      react: 19.2.3 => 19.2.3
    npmGlobalPackages:
      eas-cli: 18.0.3
    Expo Workflow: managed
```

`@expo/ui`: 57.0.6 (latest)

### Expo Doctor Diagnostics

```
19/20 checks passed.
Failed check: "Check Expo config (app.json/ app.config.js) schema"
  TypeError: fetch failed — ConnectTimeoutError (exp.host:443)
```

The single failing check is a local network restriction reaching the Expo API (`exp.host` is unreachable from this machine), not a project issue — all offline checks pass.

intergalacticspacehighway:
# Why

Fixes - https://github.com/expo/expo/issues/47883, https://github.com/expo/expo/pull/48059

Fixes the layout loops that happens in some cases when `Host` and `RNHostView` has `matchContents` and child of `RNHostView` depends upon parent on sizing. Explained it in more detail here - https://github.com/expo/expo/pull/48059#issuecomment-5351404485

# How

Set `RNHostView` node as a leaf and measurable node using shadow node traits. This removes it from the yoga layout hierarchy (shadow node hierarchy is still preserved). Override `measureContent` and `layout` shadow node functions for `RNHostView`. The `measureContent` is called by Yoga during layout and `layout` is called during commit. 

A better fix would be if Yoga supports `max-content` so we could just do `<RNHostView style={{ width: "max-content", height: "max-content" }} />` if someone passes `matchContents`. Made a PR in Yoga for that https://github.com/react/yoga/pull/2012

# Test Plan

Added regression testcases for RNHostView which leads to loops.

# Checklist


packages/expo-modules-core/common/cpp/fabric/ExpoViewShadowNode.h
```diff
@@ -77,7 +82,112 @@ class ExpoViewShadowNode : public facebook::react::ConcreteViewShadowNode<
...
+      return ConcreteViewShadowNode::measureContent(layoutContext, layoutConstraints);
+    }
+
+    auto const *content = hostedContent();
+
+    if (content == nullptr) {
+      return {};
+    }
+
+    return content->measure(layoutContext, hostedContentConstraints(*content));
```
intergalacticspacehighway, line 113: That is intentional. When `matchContents` is set on `RNHostView` we ignore the parent constraints. The way to enable wrapping would be to pass a RN child with fixed size/maxWidth, or use RNHostView without `matchContents`. Removing the parent shadow node dependency on layout when `matchContents` is set is what fixes the layout loops.

packages/expo-modules-core/common/cpp/fabric/ExpoViewComponentDescriptor.h
```diff
@@ -30,6 +30,28 @@ class ExpoViewComponentDescriptor
     return std::static_pointer_cast<std::string const>(this->flavor_)->c_str();
   }
 
+  std::shared_ptr<facebook::react::ShadowNode> createShadowNode(
+    const facebook::react::ShadowNodeFragment &fragment,
+    const facebook::react::ShadowNodeFamily::Shared &family
+  ) const override {
+    // Currently, only RNHostView can have `sizesToContent` set to true.
+    if (!ShadowNodeType::sizesToContent(fragment.props)) {
```
Kudo: makes me thinking we should split a new function like

```cpp
bool isRNHostView(const react::Props::Shared &props) {
  // Currently, only RNHostView can have `sizesToContent` set to true.
  return ShadowNodeType::sizesToContent(props);
}
```

to stop copy the sizesToContent everywhere with the comment. then later at some point we could use better detection for RNHostView

packages/expo-modules-core/common/cpp/fabric/ExpoViewShadowNode.h
```diff
@@ -77,7 +82,112 @@ class ExpoViewShadowNode : public facebook::react::ConcreteViewShadowNode<
...
+    }
+  }
+
 private:
+  const react::LayoutableShadowNode *hostedContent() const {
+    auto const &children = this->getChildren();
+
+    return children.empty()
+      ? nullptr
+      : dynamic_cast<const react::LayoutableShadowNode *>(children.front().get());
```
Kudo, line 152: this one feels valid

intergalacticspacehighway, line 152: Currently `RNHostView` only supports single child (also type prevents passing multiple children). Supporting multiple will require quite a few changes (touch dispatcher). It can be done in another PR.

packages/expo-modules-core/common/cpp/fabric/ExpoViewComponentDescriptor.h
intergalacticspacehighway: Done. And replied for another review item below

Kudo (approved):

PR #49486, opened 2026-08-28
[android][ios][font] Select faces of a multi-face font family with fontWeight and fontStyle

vonovak:
# Why

Layer 2 in the stack that splits #47939 (base: #49485, next: #49487).

Layer 1 added multi-face font families on web only. On Android and iOS, only the first `fontDefinitions` entry per family was loaded, so `fontWeight`/`fontStyle` styles fell back to a system font.

# How

- **Android**: registers every declared face and builds a `Typeface` per requested weight/style pair (`FontFamilyFaces.kt`). Selects the nearest declared weight, instancing a variable font's `wght` axis where available. Rejects duplicate faces on resolved values, not declared ones.
- **iOS**: registers every face file and resolves the family alias to the face whose traits match the requested `fontWeight`/`fontStyle` (`FontFamilyAliasManager`, `FontUtils`). Commits a family's alias atomically and reads traits from the default instance.
- Adds native unit tests (`FontFamilyFacesTest.kt`, `MultiFaceTypefaceTest.kt`) and JS tests for the shared loader path.
- Docs: orders the weight/style platform notes as Android, iOS, web.

# Test Plan

- `packages/expo-font`: `yarn jest` — 21 suites pass.
- `et check-packages expo-font` passes locally.
- Native unit tests run via `et native-unit-tests --packages expo-font`.
- Manual check in NCL `FontScreen` (multi-face family example) on both platforms.

# Checklist


packages/expo-font/android/src/main/java/expo/modules/font/FontLoaderModule.kt
```diff
@@ -32,24 +35,99 @@ open class FontLoaderModule : Module() {
...
+    val instanced = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+      buildVariableWeightTypeface(fontFamilyName, source)
+    } else {
+      null
+    }
+    return instanced ?: source.load()
+  }
+
+  @RequiresApi(Build.VERSION_CODES.Q)
+  private fun buildMultiFaceTypeface(fontFamilyName: String, faces: List<FontFaceRecord>): Typeface {
```
tsapeta, line 84: [🏗️ **design**] A variable font file inside a multi-face family is not `wght`-instanced here. This path labels each `Font` with one weight/slant and builds a `FontFamily`; the public `FontFamily.Builder` has no automatic variation instancing (that is the hidden `buildVariableFamily` path AOSP keeps for system fonts). The nine-instance coverage from `buildVariableWeightTypeface` only applies via `loadSingleFaceTypeface`, which `familyTypeface` reaches only for a single face with `weight` and `style` both unset.

So `[{ path: robotoFlex }, { path: italicFile, style: 'italic' }]` renders every non-italic weight at the variable file's default instance (plus synthetic bold where Android applies it), while the same definition on web covers the full `wght` range. The PR description says "instancing a variable font's `wght` axis where available", which reads as if this path instanced too.

If this limitation is intended for now, consider stating it in the `FontFaceDefinition.weight` docs (the Android paragraph currently says "the weight read from the font file applies", which doesn't tell the reader the other weights become synthetic). If it isn't, the multi-face path could instance a `wght`-axis face at each of the standard weights the way the single-face path does.

packages/expo-font/src/FontLoader.ts
```diff
@@ -36,5 +33,88 @@ export async function loadSingleFontAsync(
...
+    if (weight !== undefined) {
+      face.weight = weight;
+    }
+    if (style !== undefined) {
+      face.style = style;
+    }
+    return face;
+  });
+
+  await ExpoFontLoader.loadFontFamilyAsync!(fontFamily, faces);
```
tsapeta, line 119: [💡 suggestion] The non-null assertion turns a version mismatch into an opaque `TypeError: ... is not a function`. New JS with an older native binary is a state users do reach (an OTA update with a stale runtime version policy, or a bare app that upgraded JS without rebuilding). A guard that throws `UnavailabilityError('expo-font', 'loadFontFamilyAsync')` (or a `CodedError` telling the user to rebuild the native app) would make the failure actionable, matching how the rest of the module handles optional native methods (`unloadAsync`, `unloadAllAsync`).

packages/expo-font/src/FontLoader.ts
```diff
@@ -36,5 +33,88 @@ export async function loadSingleFontAsync(
...
+    numeric = weight;
+  } else if (weight === 'normal') {
+    numeric = 400;
+  } else if (weight === 'bold') {
+    numeric = 700;
+  } else {
+    numeric = parseInt(weight, 10);
+  }
+
+  if (!Number.isFinite(numeric) || numeric < 1 || numeric > 1000) {
```
tsapeta, line 65: [💡 suggestion] A fractional numeric weight passes this check (CSS Fonts 4 allows `weight: 450.5`, and the type is `number`), so the value reaches the native `Int`/`Int?` record fields undamped. What happens then depends on each platform's record converter (truncation or a conversion error naming the field, not the API misuse). String inputs already get truncated by `parseInt`, so numbers behave differently from numeric strings. Rounding `numeric` here (or rejecting non-integers with this same clear error) keeps the failure mode in JS and deterministic across platforms.

packages/expo-font/src/Font.ts
```diff
@@ -170,46 +168,22 @@ async function loadFontFamilyInNamespaceAsync(definition: FontFamilyDefinition):
...
-// Faces of the same `fontFamily` are cached and awaited individually (keyed by weight/style)
-// since native and the JS-level cache in `memory.ts` can only track loaded state per name.
-async function loadFontFaceInNamespaceAsync(
-  fontFamily: string,
-  face: FontFaceDefinition
-): Promise<void> {
-  const cacheKey = `${fontFamily}\0${face.weight ?? ''}\0${face.style ?? ''}`;
-
-  if (isLoadedInCache(cacheKey)) {
+  // consulting the native module is slower than the cache, but avoids reloading the same font
```
tsapeta, line 171: [💡 suggestion] Just confirming this semantic change from layer 1 is intended: with #49485's per-face cache keys, a second `loadAsync([{ fontFamily, fontDefinitions: [regular, bold] }])` after loading `[regular]` would still load the bold face; now the whole call silently no-ops because the family name is already loaded. That forecloses progressive face loading (load regular up front, heavier faces later), and it means the native "last call wins" contract both platforms implement only ever triggers on JS reloads. The behavior is consistent with the single-font path and the new doc note covers it, so this is fine if deliberate — flagging it because a layer-1 reader could reasonably expect the incremental behavior.

packages/expo-font/src/FontLoader.web.ts
```diff
@@ -79,6 +79,42 @@ function throwInvalidSourceError(source: any): never {
   );
 }
 
+// Merges a face's `weight`/`style`/`display`/`testString` onto its `path`, falling back to
```
tsapeta, line 82: [🐧 nit] This duplicates `fontSourceFromFace` in `Font.ts` (kept there for the SSR `registerStaticFont` branch), and the two copies must stay in sync on the merge semantics. Exporting the `Font.ts` one and importing it here would leave a single source of truth.

vonovak: Resolved up the stack in #49487: extracted to fontSourceFromFace.ts, and both places use it

packages/expo-font/src/Font.ts
vonovak: Yes. A family loads atomically: all faces in one call, one cache key. There is a doc note that calls it out.

packages/expo-font/android/src/main/java/expo/modules/font/FontLoaderModule.kt
vonovak: Fixed in https://github.com/expo/expo/commit/0279d3ae43998328683939252d3a3c77e0256ddc, sorry, I forgot to (ask the agent to) push it before the review;

packages/expo-font/src/FontLoader.ts
vonovak: the non-null assertion is removed up the stack, but generally this won't go out with an OTA so compatibility is not a concern.

packages/expo-font/src/FontLoader.ts
vonovak: Partially overtaken up the stack: the strict runtime validation moved to fontFaceValidation.ts.

That being said, fractional values are not checked. I could add the check but there's a lot of more valuable validations already and I feel this is marginal? Happy to add it though.

packages/expo-font/src/Font.types.ts
```diff
@@ -51,19 +51,30 @@ export type FontFaceDefinition = {
...
+   * the font file applies, and a variable font face keeps its whole `wght` axis &mdash;
+   * `fontWeight` renders the matching instance. Setting a weight on a variable font face pins it
+   * to that single instance. On API levels below 29, only the family's default face (closest to
+   * a regular, upright weight and style) is loaded.
+   *
+   * On iOS, this value isn't used to select the face; iOS reads the weight embedded in the font
+   * file's own metadata instead.
+   *
+   * On web, maps to the CSS `font-weight` property. Leave unset for a variable font file that
+   * covers a range of weights &mdash; a single value restricts the face to only that weight.
```
amandeepmittal, line 64: ✏️ **style** `prose.md#sentence-construction`

Three small edits, all in the suggestion: split the three-clause "When unset" sentence, write the Android version with its API level ("Below Android 10 (API level 29)") as the other packages do, and use the em dash character instead of `&mdash;`, which no other package uses.

```suggestion
   * On Android, the declared weight used to select this face. When unset, the weight read from
   * the font file applies. A variable font face then keeps its whole `wght` axis, and `fontWeight`
   * renders the matching instance. Setting a weight on a variable font face pins it to that
   * single instance. Below Android 10 (API level 29), only the family's default face (closest to
   * a regular, upright weight and style) is loaded.
   *
   * On iOS, this value isn't used to select the face; iOS reads the weight embedded in the font
   * file's own metadata instead.
   *
   * On web, maps to the CSS `font-weight` property. Leave unset for a variable font file that
   * covers a range of weights. A single value restricts the face to only that weight.
```

packages/expo-font/src/Font.types.ts
```diff
@@ -51,19 +51,30 @@ export type FontFaceDefinition = {
...
+   * On Android, the declared style used to select this face. When unset, the style read from
+   * the font file applies. On API levels below 29, only the family's default face (closest to a
+   * regular, upright weight and style) is loaded.
+   *
+   * On iOS, this value isn't used to select the face; iOS reads the style embedded in the font
+   * file's own metadata instead.
+   *
+   * On web, maps to the CSS `font-style` property. Leave unset for a variable font file that
+   * covers both upright and italic/oblique styles &mdash; a single value restricts the face to
+   * only that style.
```
amandeepmittal, line 77: ✏️ **style** `prose.md#sentence-construction`

Same two edits as on `weight`, in the suggestion: write the Android version with its API level ("Below Android 10 (API level 29)"), and replace the `&mdash;` clause with a sentence.

```suggestion
   * On Android, the declared style used to select this face. When unset, the style read from
   * the font file applies. Below Android 10 (API level 29), only the family's default face
   * (closest to a regular, upright weight and style) is loaded.
   *
   * On iOS, this value isn't used to select the face; iOS reads the style embedded in the font
   * file's own metadata instead.
   *
   * On web, maps to the CSS `font-style` property. Leave unset for a variable font file that
   * covers both upright and italic/oblique styles. A single value restricts the face to only
   * that style.
```

tsapeta (approved):

PR #49487, opened 2026-08-28
[font] Validate faces early and fix multi-face registration on web and iOS

vonovak:
# Why

Layer 3 in the stack that splits #47939 (base: #49486).

Follow-up fixes and hardening on top of the multi-face font family feature:

- Invalid `fontDefinitions` input failed late, deep in platform code, with unclear errors.
- On web, face registration skipped every face after the first one with a distinct descriptor set, and rule matching broke across engines that serialize `font-family` differently.
- On iOS, unloading one alias unregistered a font file that another alias still referenced.

# How

- Validates `FontFamilyDefinition[]` input early in shared JS (`fontFaceValidation.ts`) with actionable error messages, before any platform call.
- Extracts `fontSourceFromFace.ts` so web and native resolve faces the same way; registers every face of a family on web.
- Fixes cross-engine `@font-face` rule matching (quoted vs bare family-name serialization) and adds a web integration test suite.
- iOS: reference-counts font file registrations so unloading one alias does not break another.

# Test Plan

- `packages/expo-font`: `yarn jest` — 31 suites, 263 tests pass.
- `et check-packages expo-font` passes locally.

# Checklist


packages/expo-font/src/Font.types.ts
```diff
@@ -60,8 +60,11 @@ export type FontFaceDefinition = {
    * On iOS, this value isn't used to select the face; iOS reads the weight embedded in the font
    * file's own metadata instead.
    *
-   * On web, maps to the CSS `font-weight` property. Leave unset for a variable font file that
-   * covers a range of weights &mdash; a single value restricts the face to only that weight.
+   * On web, maps to the CSS `font-weight` property. A variable font file can also declare a
+   * weight range as `'<min> <max>'`, for example `'100 900'`. Leave unset for a variable font
+   * file that covers its full range of weights &mdash; a single value restricts the face to only
+   * that weight. A range is ignored on Android, which keeps the variable font's whole `wght`
+   * axis, and on iOS, which reads the weight from the font file instead.
```
amandeepmittal, line 67: ⚠️ **correctness** `mdx-components-sdk.md#apisection`

The regenerated **docs/public/static/data/unversioned/expo-font.json** in this PR still carries an earlier draft of this property's text: its web paragraph ends with "Android and iOS ignore a range and read the weight from the font file instead", and its Android paragraph lacks the two `wght` axis sentences. Rerun `et generate-docs-api-data --packageName expo-font` after the final TSDoc edits so the reference shows the text in the source.

---

Two edits in the suggestion: replace the `&mdash;` clause with a sentence, and put the last sentence in the active voice with one platform per sentence.

```suggestion
   * On web, maps to the CSS `font-weight` property. A variable font file can also declare a
   * weight range as `'<min> <max>'`, for example `'100 900'`. Leave unset for a variable font
   * file that covers its full range of weights. A single value restricts the face to only that
   * weight. Android ignores a range and keeps the variable font's whole `wght` axis. iOS ignores
   * a range and reads the weight from the font file instead.
```

packages/expo-font/src/Font.ts
```diff
@@ -70,6 +64,9 @@ export function isLoading(fontFamily: string): boolean {
  * > **Note**: We recommend using the [config plugin](#configuration-in-app-config) instead whenever possible.
  *
  * > **Note**: When the `fontFamily` is already loaded, this method resolves without replacing it.
+ * > This applies across the string and array APIs, and across separate calls: a `fontFamily`
+ * > loads once. Declare all of its faces in that one call. A later call for an already-loaded
+ * > `fontFamily` adds none of its faces, even if that call lists faces the first call did not.
```
amandeepmittal, line 69: ✏️ **style** `review-ai-text.md#t5-restatement`

The line above already says an already-loaded family is not replaced, so "a `fontFamily` loads once" and "even if that call lists faces the first call did not" restate it. The suggestion keeps the two new facts: the rule spans both call styles, and later faces are dropped.

```suggestion
 * > This applies across the string and array APIs and across separate calls, so declare all of a
 * > family's faces in one call. A later call for the same `fontFamily` adds none of its faces.
```

tsapeta (approved):

PR #49489, opened 2026-08-28
[observe] dispatch native crashes as logs

Ubax:
# Why

We are collecting crash reports locally, but we never dispatch them.

# How

Similar to JS errors dispatch crashes as logs. 

1. When crash is recorded add a new log for it
2. Trim stack trace to 50 frames, so that we don't make the log too big

The logs use following attributes:

* `event.name`: `"exception"`
* `session.id`: session associated with the crash
* `severity`: `"FATAL"`
* `exception.type`: native exception or signal name 
* `exception.message`: termination or exception reason
* `exception.stacktrace`: native call stack, when available
* `expo.error.source`: nativeCrash
* `expo.error.is_fatal`: true
* `expo.crash.signal`: signal name, such as SIGSEGV
* `expo.crash.signal_code`: numeric signal code
* `expo.crash.termination_reason`: OS-provided termination reason
* `expo.crash.exception_type`: iOS Mach exception name
* `expo.crash.exception_type_code`: iOS Mach exception type code
* `expo.crash.exception_code`: iOS processor-specific exception code
* `expo.crash.objc_exception_type`: Objective-C exception type
* `expo.crash.objc_exception_message`: Objective-C exception message

android example: 
```json
{
  "attributes": {
    "exception.message": "java.lang.ArithmeticException: divide by zero",
    "exception.stacktrace": "expo.modules.crashtester.CrashTriggers.throwFor(CrashTester.kt:76)\nexpo.modules.crashtester.CrashTriggers.trigger$lambda$0(CrashTester.kt:65)\nexpo.modules.crashtester.CrashTriggers$$ExternalSyntheticLambda0.run(D8$$SyntheticClass:0)\nandroid.os.Handler.handleCallback(Handler.java:873)\nandroid.os.Handler.dispatchMessage(Handler.java:99)\nandroid.os.Looper.loop(Looper.java:193)\nandroid.app.ActivityThread.main(ActivityThread.java:6669)\njava.lang.reflect.Method.invoke(Native Method)\ncom.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)\ncom.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)",
    "exception.type": "java.lang.ArithmeticException",
    "expo.error.is_fatal": true,
    "expo.error.source": "nativeCrash"
  },
  "body": null,
  "name": "exception",
  "severity": "fatal",
  "timestamp": "2026-08-28T12:17:47.406Z"
}
```

# Test Plan

1. CI
2. Observe-tester

# Checklist

### DIFF
--- a/packages/expo-app-metrics/CHANGELOG.md
+++ b/packages/expo-app-metrics/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### 🎉 New features
 
+- [iOS] [Android] Dispatch attributed native crashes as fatal exception logs. ([#49489](https://github.com/expo/expo/pull/49489) by [@Ubax](https://github.com/Ubax))
 - Add an optional `displayName` to `logEvent` ([#47289](https://github.com/expo/expo/pull/47289) by [@Ubax](https://github.com/Ubax))
 - Capture React render-phase errors via `AppMetricsErrorBoundary`. ([#47341](https://github.com/expo/expo/pull/47341) by [@tsapeta](https://github.com/tsapeta))
 - Describe the network a launch ran on: connection cost, request throughput, and a `slowest.*` group replacing `expo.network.requests.slowestDuration` and `slowestHost`. ([#48518](https://github.com/expo/expo/pull/48518) by [@tsapeta](https://github.com/tsapeta))
--- a/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/AppMetricsModule.kt
+++ b/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/AppMetricsModule.kt
@@ -168,13 +168,14 @@ class AppMetricsModule : Module(), UpdatesStateChangeListener {
             exitInfoProvider = ExitInfoProviderImpl(context),
             lastProcessedExitStore = PreferencesLastProcessedExitStore(context),
             appVersion = metadata?.appVersion
-          ) { sessionId, origin, report ->
+          ) { sessionId, origin, report, logDetails ->
             attributeAndStoreCrashReport(
               sessionManager = sessionManager,
               currentSessionId = mainSession.sessionId,
               sessionId = sessionId,
               origin = origin,
-              report = report
+              report = report,
+              logDetails = logDetails
             )
           }.process()
         }
--- a/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/crashreporting/CrashReport.kt
+++ b/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/crashreporting/CrashReport.kt
@@ -48,7 +48,7 @@ data class CrashReport(
     return LogRecord(
       sessionId = sessionId,
       timestamp = timestampBegin,
-      name = "exception",
+      name = "native.exception",
       severity = Severity.FATAL.rawValue,
       attributes = JsonAny.encodeMapToJsonString(attributes)
     )
--- a/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/crashreporting/CrashReport.kt
+++ b/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/crashreporting/CrashReport.kt
@@ -1,5 +1,9 @@
 package expo.modules.appmetrics.crashreporting
 
+import expo.modules.appmetrics.logevents.Severity
+import expo.modules.appmetrics.logevents.truncateToMaxLength
+import expo.modules.appmetrics.storage.LogRecord
+import expo.modules.appmetrics.utils.JsonAny
 import kotlinx.serialization.Serializable
 import kotlinx.serialization.encodeToString
 import kotlinx.serialization.json.Json
@@ -28,6 +32,51 @@ data class CrashReport(
    */
   val ingestedAt: String
 ) {
+  fun toLogRecord(sessionId: String, details: CrashLogDetails = CrashLogDetails()): LogRecord {
+    val attributes = buildMap<String, Any?> {
+      put("exception.type", details.exceptionType ?: signal?.let(::signalName) ?: "NativeCrash")
+      put("exception.message", terminationReason ?: exceptionReason ?: signal?.let(::signalName) ?: "Native crash")
+      put("exception.stacktrace", renderStacktrace(details.stackFrames))
+      put("expo.error.source", "nativeCrash")
+      put("expo.error.is_fatal", true)
+      signal?.let {
+        put("expo.crash.signal", signalName(it))
+        put("expo.crash.signal_number", it)
+      }
+      terminationReason?.let { put("expo.crash.termination_reason", it) }
+    }
+    return LogRecord(
+      sessionId = sessionId,
+      timestamp = timestampBegin,
+      name = "exception",
+      severity = Severity.FATAL.rawValue,
+      attributes = JsonAny.encodeMapToJsonString(attributes)
+    )
+  }
+
+  private fun renderStacktrace(stackFrames: List<String>?): String? {
+    val callStacks = callStackTree?.callStacks.orEmpty()
+    val attributedStacks = callStacks.filter { it.threadAttributed == true }
+    val selectedStacks = attributedStacks.ifEmpty { callStacks }
+    val frames = stackFrames ?: selectedStacks
+      .flatMap { it.callStackRootFrames.orEmpty() }
+      .map { it.symbol ?: "<unknown>" }
+    if (frames.isEmpty()) {
+      return null
+    }
+    val rendered = buildList {
+      addAll(frames.take(MAX_LOG_STACK_FRAMES))
+      if (frames.size > MAX_LOG_STACK_FRAMES) {
+        add("… +${frames.size - MAX_LOG_STACK_FRAMES} more frames")
+      }
+    }.joinToString("\n")
+    return truncateToMaxLength(
+      rendered,
+      MAX_LOG_STACKTRACE_LENGTH,
+      "Native crash stack trace exceeded the maximum length and was truncated."
+    )
+  }
+
   @Serializable
   data class CallStackTree(
     val callStacks: List<CallStack>? = null
@@ -93,5 +142,25 @@ data class CrashReport(
     }
 
     private const val MAX_CAUSE_DEPTH = 5
+    private const val MAX_LOG_STACK_FRAMES = 25
+    private const val MAX_LOG_STACKTRACE_LENGTH = 65_536
+
+    private fun signalName(signal: Int): String =
+      when (signal) {
+        4 -> "SIGILL"
+        5 -> "SIGTRAP"
+        6 -> "SIGABRT"
+        7 -> "SIGBUS"
+        8 -> "SIGFPE"
+        9 -> "SIGKILL"
+        11 -> "SIGSEGV"
+        15 -> "SIGTERM"
+        else -> "SIG$signal"
+      }
   }
 }
+
+data class CrashLogDetails(
+  val exceptionType: String? = null,
+  val stackFrames: List<String>? = null
+)
--- a/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/crashreporting/CrashReportAttribution.kt
+++ b/packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/crashreporting/CrashReportAttribution.kt
@@ -13,7 +13,8 @@ suspend fun attributeAndStoreCrashReport(
   currentSessionId: String?,
   sessionId: String?,
   origin: CrashOrigin,
-  report: CrashReport
+  report: CrashReport,
+  logDetails: CrashLogDetails = CrashLogDetails()
 ) {
   runCatching {
     // A null target stores the report as an orphan.
@@ -29,7 +30,12 @@ suspend fun attributeAndStoreCrashReport(
       // id-less JVM file (a crash before the main session existed) → orphan.
       else -> null
     }
-    sessionManager.setCrashReport(target, report.encodeToJsonString())
+    val payload = report.encodeToJsonString()
+    if (target != null) {
+      sessionManager.storeCrashReportIfNew(target, payload, report.toLogRecord(target, logDetails))
+    } else {
+      sessionManager.setCrashReport(null, payload)
+    }
   }.onFailure {
     Log.e(TAG, "Failed to persist a crash report", it)
   }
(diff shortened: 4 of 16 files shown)


packages/expo-app-metrics/ios/CrashReporting/CrashReport.swift
```diff
@@ -87,6 +90,116 @@ public struct CrashReport: Codable, Sendable {
...
+  private var renderedStacktrace: String? {
+    guard let callStacks = callStackTree?.callStacks else {
+      return nil
+    }
+    let orderedStacks =
+      callStacks.filter { $0.threadAttributed == true }
+      + callStacks.filter { $0.threadAttributed != true }
+    var lines: [String] = []
+    for callStack in orderedStacks {
+      var frames: [CallStackTree.Frame] = []
```
jakub-agent: [🏗️ **design**] `frames` and `totalFrames` are declared inside the `for callStack in orderedStacks` loop, so `maxLogStackFrames` is a per-thread budget, not a per-report one. A crash diagnostic's call-stack tree holds every thread — `CrashReportSymbolicator.symbolicate` says as much ("Threads in a crash tree share many leaf frames (RunLoop guts, pthread entry points, Hermes interpreter trampolines)"). So a 30-thread crash renders up to 1500 frame lines plus 30 separate `… +N more frames` markers, and only `maxLogStacktraceLength` (64 KB) actually bounds the attribute. The Android side caps at 50 total frames across all stacks, and the PR description says "trim stack trace to 50 frames".

Hoisting both vars above the loop would restore the intended cap. `ToLogRecordTests.renders at most fifty stack frames…` builds a single call stack, so it passes either way — a second, unattributed call stack in that fixture would cover it.

Separately, while you're here: frames from different threads are concatenated with no boundary line, so the resulting `exception.stacktrace` reads as one continuous stack. A `--- thread N ---` style separator would make the attribute usable for grouping.

packages/expo-app-metrics/ios/CrashReporting/CrashReport.swift
```diff
@@ -87,6 +90,116 @@ public struct CrashReport: Codable, Sendable {
...
+    }
+    if let exceptionReason {
+      attributes["expo.crash.objc_exception_type"] = exceptionReason.exceptionType
+      attributes["expo.crash.objc_exception_message"] = exceptionReason.composedMessage
+    }
+    return LogRecord(
+      name: "exception",
+      attributes: attributes,
+      severity: .fatal,
+      timestamp: ingestedAt.ISO8601Format()
```
jakub-agent: [🏗️ **design**] `ingestedAt` is `Date.now` at next-launch ingestion — the field's own doc says it "reflects when *we* learned about the crash, not when it happened." The log is attributed to the previous session, so its timestamp lands after that session's `endTimestamp`, potentially days later if MetricKit backlogs the payload. Android uses `timestampBegin` for the same log (`CrashReport.kt`), so a crash timeline built from these logs is inconsistent across platforms.

I realise `timestampBegin`/`timestampEnd` are the MetricKit payload *window* rather than the crash instant, so neither is exact — was `ingestedAt` a deliberate choice over `timestampEnd` (the tightest upper bound on the crash time, and the one that stays inside or near the session window)? If so, worth a line of comment here, because the asymmetry with Android is otherwise going to read as an oversight.

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/crashreporting/CrashReportAttribution.kt
```diff
@@ -29,7 +30,11 @@ suspend fun attributeAndStoreCrashReport(
       // id-less JVM file (a crash before the main session existed) → orphan.
       else -> null
     }
+    val alreadyHasCrashReport = target != null && sessionManager.getCrashReport(target) != null
```
jakub-agent: [💡 suggestion] This is check-then-write with no transaction, while the iOS half of the same PR got `storeCrashReportIfNew` with an explicit transaction *and* a rollback test. If `addLogs` throws after `setCrashReport` succeeded, the outer `runCatching` swallows it and the now-stored report makes `alreadyHasCrashReport` true forever — the log is never retried.

Reachability is low (both writes carry the same session FK, so if one can succeed the other normally can too), so this isn't urgent. But `SessionManager` already exposes Room, and wrapping the read + both writes in a `withTransaction` would make the two platforms behave the same way and cost very little.

One difference worth deciding deliberately rather than by accident: `setCrashReport` runs unconditionally here (Room `upsert` → REPLACE) while only the log write is guarded, whereas iOS's `storeCrashReportIfNew` skips *both*. Today that can't diverge — the `EXIT_RECORD` branch above already refuses a target that has a report — but the invariant is only held up by that one `takeIf`.

packages/expo-app-metrics/android/src/test/java/expo/modules/appmetrics/crashreporting/CrashReportAttributionTest.kt
```diff
@@ -78,6 +78,18 @@ class CrashReportAttributionTest {
...
+
+  @Test
+  fun `does not duplicate the crash log when a report is reprocessed`() =
+    runTest {
+      sessionManager.startSessionWithIdAt("crashed-session", "2023-11-14T22:00:00.000Z")
+
+      attribute("crashed-session", CrashOrigin.JVM_FILE)
+      attribute("crashed-session", CrashOrigin.JVM_FILE)
+
+      assertEquals(1, sessionManager.getLogsForSession("crashed-session").size)
```
jakub-agent, line 92: [💡 suggestion] This test passes whether or not the `alreadyHasCrashReport` guard exists: `logId` is the deterministic `"native-crash:$sessionId"` and `LogDao.insertAll` uses `OnConflictStrategy.IGNORE`, so the second insert can't produce a second row regardless.

What the guard actually protects is one step further out — without it, `addLogs` still fires `logsInsertListeners` with the already-dispatched `logId`, so `PendingLogsManager.addPendingLogs` re-queues it and expo-observe dispatches the same crash log twice. That's the behaviour worth asserting: spy on `sessionManager.addLogsInsertListener` (or assert `addLogs` is invoked once) rather than counting rows.

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/AppMetricsModule.kt
```diff
@@ -168,13 +168,14 @@ class AppMetricsModule : Module(), UpdatesStateChangeListener {
             exitInfoProvider = ExitInfoProviderImpl(context),
             lastProcessedExitStore = PreferencesLastProcessedExitStore(context),
             appVersion = metadata?.appVersion
-          ) { sessionId, origin, report ->
+          ) { sessionId, origin, report, logDetails ->
```
jakub-agent, line 171: [💡 suggestion] Question about whether the Android half of this actually dispatches, since that's the whole point of the PR.

On Android, dispatch is strictly pending-ID driven: `SessionManager.addLogs` notifies `logsInsertListeners`, `ObservabilityManager`'s init registers the listener that calls `PendingLogsManager.addPendingLogs`, and `dispatchUnsentLogs` returns immediately unless `hasPendingLogs()`. There is no scan-all fallback over the `logs` table.

This write happens from a `scope.launch` inside `AppMetricsModule`'s `OnCreate`, while `ObserveModule`'s `OnCreate` is what registers the listener. If the coroutine wins that race, the row is persisted and then silently never dispatched.

The pre-existing `PendingErrorStore.drain()` block further down this same `OnCreate` has the identical shape, so this isn't introduced here — but this PR adds a second consumer of the pattern, and the failure mode is invisible. Did the observe-tester run confirm the crash log reaching the collector on Android? If the ordering is in fact guaranteed by module-registry order, a short comment saying so would be worth more than the test.

packages/expo-app-metrics/ios/Database/MetricsDatabase.swift
```diff
@@ -542,6 +542,19 @@ final class MetricsDatabase: Sendable {
...
 
+  /// Stores an attributed crash and its dispatchable log atomically. The crash report's primary key
+  /// is the idempotence marker for MetricKit payload redelivery.
+  @AppMetricsActor
+  func storeCrashReportIfNew(sessionId: String, payload: String, log: LogRow) throws {
+    try database.transaction {
+      if try getCrashReport(sessionId: sessionId) != nil {
+        return
+      }
+      try setCrashReport(sessionId: sessionId, payload: payload)
```
jakub-agent, line 553: [💡 suggestion] `setCrashReport`'s doc comment right above still says "The session row is *not* required to exist — crashes are sometimes attributed to a session that was never written (e.g. the app crashed before the session row reached disk)," and `MetricsDatabaseTests.setCrashReport accepts an orphan session id` still asserts that. But `logs.sessionId` is `REFERENCES sessions(id)` with `PRAGMA foreign_keys = ON`, so through this new path an orphan session id now rolls the crash report back too — which is exactly what the new `rolls back the crash report when its log cannot be stored` test pins down.

Unreachable in practice today (`persistCrashReport` only ever passes an id that came out of `getMainSessions()`), so this is a docs point rather than a bug: worth a sentence on `storeCrashReportIfNew` noting it deliberately requires a persisted session, so the two comments don't read as contradicting each other.

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/crashreporting/CrashReport.kt
```diff
@@ -28,6 +32,51 @@ data class CrashReport(
...
+  fun toLogRecord(sessionId: String, details: CrashLogDetails = CrashLogDetails()): LogRecord {
+    val attributes = buildMap<String, Any?> {
+      put("exception.type", details.exceptionType ?: signal?.let(::signalName) ?: "NativeCrash")
+      put("exception.message", terminationReason ?: exceptionReason ?: signal?.let(::signalName) ?: "Native crash")
+      put("exception.stacktrace", renderStacktrace(details.stackFrames))
+      put("expo.error.source", "nativeCrash")
+      put("expo.error.is_fatal", true)
+      signal?.let {
+        put("expo.crash.signal", signalName(it))
+        put("expo.crash.signal_code", it)
```
jakub-agent: [🐧 nit] `expo.crash.signal_code` holds the signal *number* (11 for SIGSEGV), not a signal code. POSIX uses "code" for `siginfo_t.si_code`, and iOS makes the clash sharper by emitting a genuinely-code-shaped `expo.crash.exception_code` alongside this key. `expo.crash.signal_number` would be unambiguous.

Raising it as a nit only because attribute keys are a wire contract with the collector — cheap to change now, awkward once dashboards query it.

packages/expo-app-metrics/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 ### 🎉 New features
 
+- [iOS] [Android] Dispatch attributed native crashes as fatal exception logs.
```
jakub-agent: [🐧 nit] Missing the `([#49489](https://github.com/expo/expo/pull/49489) by [@Ubax](https://github.com/Ubax))` suffix that every neighbouring entry carries (per CONTRIBUTING).

jakub-agent (commented):

The shape is right: reusing the `exception` event name and the `expo.error.*` / `exception.*` attribute set from `ErrorReport` means native crashes land in the same bucket as JS ones, and `storeCrashReportIfNew` is a good answer to MetricKit payload redelivery.

Two things I'd look at before merging:

1. **The 50-frame cap doesn't hold on iOS.** `renderedStacktrace` declares `frames`/`totalFrames` inside the per-call-stack loop, so the budget is per thread, not per report. A real `MXCrashDiagnostic` call-stack tree carries every thread — `CrashReportSymbolicator`'s own comment says so — and each contributes up to 50 lines plus its own `… +N more frames` marker. Android caps at 50 total. Only the 64 KB truncation bounds the iOS value.
2. **The two platforms disagree on the log timestamp.** iOS uses `ingestedAt` (documented in the same file as "when *we* learned about the crash, not when it happened"); Android uses `timestampBegin`.

The rest is smaller: an Android/iOS asymmetry in how the crash+log write is made atomic, a test that passes for a different reason than the code it covers, and two naming/changelog nits.

I also checked and found nothing wrong with: the `MainSession.storeCrashReport` removal (dead on `main`), the exit-record overwrite path (`attributeAndStoreCrashReport` already guards it), and the Darwin-vs-Linux signal number tables (both correct for their platform).

packages/expo-app-metrics/ios/CrashReporting/CrashReport.swift
Ubax: Change the limit to 25 for now and do it only for the affected thread

packages/expo-app-metrics/ios/CrashReporting/CrashReport.swift
Ubax: For log use the `timestampEnd`

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/crashreporting/CrashReportAttribution.kt
Ubax: Do it

packages/expo-app-metrics/android/src/test/java/expo/modules/appmetrics/crashreporting/CrashReportAttributionTest.kt
Ubax: Fix the test

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/AppMetricsModule.kt
Ubax: Persist it later, so module listener always wins

packages/expo-app-metrics/ios/Database/MetricsDatabase.swift
Ubax: The log should not be added if session does not exist

packages/expo-app-metrics/android/src/main/java/expo/modules/appmetrics/crashreporting/CrashReport.kt
Ubax: Adjust

packages/expo-app-metrics/CHANGELOG.md
Ubax: Add

tsapeta (approved):

PR #49491, opened 2026-08-28
[widgets] Stub deep react-native imports in the widget bundle resolver

### LINKED ISSUE #49490: [expo-widgets] All SDK 57 widgets broken since @expo/ui 57.0.14: widget bundle pulls in react-native core → "Could not create context for layout evaluation"

https://github.com/usmsam/expo-widgets-context-error-repro  Unmodified `npx create-expo-app --example with-widgets` (only the README was replaced with repro steps). The crash reproduces without a simulator — see README: after `npm install`, two commands show the widget bundle throwing "Invariant Violation: __fbBatchedBridgeConfig is not set" in bare JavaScriptCore. On a simulator every widget renders the red "Could not create context for layout evaluation." placeholder.

### Steps to reproduce

1. Clone the repro (an unmodified `npx create-expo-app --example with-widgets`):

```sh
git clone https://github.com/usmsam/expo-widgets-context-error-repro
cd expo-widgets-context-error-repro
npm install
```

2. Build the widget JS bundle and run it in bare JavaScriptCore (this is the same environment the widget extension uses — no simulator needed):

```sh
node node_modules/expo-widgets/scripts/build-bundle.mjs "$PWD"
/System/Library/Frameworks/JavaScriptCore.framework/Versions/A/Helpers/jsc \
  node_modules/expo-widgets/bundle/build/ExpoWidgets.bundle
```

**Actual:**

```
Exception: Invariant Violation: __fbBatchedBridgeConfig is not set, cannot invoke native modules
```

**Expected:** the bundle evaluates silently.

3. (Optional, same failure on a simulator) `npx expo prebuild -p ios && npx expo run:ios`, then add the example widget to the home screen — it renders the red **"Error — Could not create context for layout evaluation."** placeholder, because `WidgetsJSRuntime.getContext()` returns nil when bundle evaluation throws.

The regression window: works with `@expo/ui@57.0.13` (2026-08-24), breaks with `@expo/ui@57.0.14` (2026-08-26) — `expo-widgets@57.0.13` depends on `~57.0.14`, so every fresh install after Aug 26 is affected. Root cause and a verified `patch-package` workaround are in the description below.

### Environment

```text
  System:
    OS: macOS 26.5.2
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.11.1
    npm: 11.6.2
    Watchman: 2025.03.03.00
  Managers:
    CocoaPods: 1.16.2
  SDKs:
    iOS SDK: iOS 26.5
  IDEs:
    Xcode: 26.6/17F113
  npmPackages:
    expo: ~57.0.16 => 57.0.16
    expo-router: ~57.0.16 => 57.0.16
    babel-preset-expo: ~57.0.0 => 57.0.8
    react: 19.2.3 => 19.2.3


Key packages: expo-widgets 57.0.13, nested @expo/ui 57.0.14 (node_modules/expo-widgets/node_modules/@expo/ui — this is the copy the widget bundle resolves, regardless of top-level pins).

Also reproduced on a second setup: fresh `create-expo-app --example with-widgets` → expo 57.0.17, react-native 0.86.0, expo-widgets 57.0.13, @expo/ui 57.0.14. The crash is RN-version-independent (the NativeModules.js init-time invariant is identical in 0.79 and 0.86) and reproduces in plain macOS `jsc`, so no simulator/device is required.
```

### Expo Doctor Diagnostics

```text

Detected Hermes V1 250829098.0.14 from React Native. Advice: upgrade to React Native 0.86.2 or later.

🔧 Patch version mismatches
package       expected  found
react-native  0.86.3    0.86.0


Note: both failures are unrelated version-alignment warnings from the pristine
example (react-native 0.86.0 vs 0.86.3). The reported crash is RN-version-independent —
it also reproduces on react-native 0.86.2, and the throwing code path
(NativeModules.js init-time invariant) is identical across RN 0.79–0.86.
```

usmsam:
# Why

Fixes #49490.

Since `@expo/ui@57.0.14` (2026-08-26), every widget built with `expo-widgets` renders the red **"Could not create context for layout evaluation."** placeholder — including the unmodified `with-widgets` example. `@expo/ui@57.0.14` added `src/keyboard/textInputState.ts` with a deep import of `react-native/Libraries/Components/TextInput/TextInputState`. The stub resolver in `expo-widgets/metro.config.js` only intercepts the exact `'react-native'` specifier, so the deep import resolves to the real react-native and drags the RN core (incl. `InitializeCore`) into the widget bundle. In the extension's bare JS context, the first require of `Libraries/BatchedBridge/NativeModules.js` throws `Invariant Violation: __fbBatchedBridgeConfig is not set` at module-init time, so `WidgetsJSRuntime` fails to create the JS context.

Full analysis (import chain from the module graph, bisection over published tarballs, crash path) is in #49490.

# How

Resolve any `react-native/*` deep import as an empty module in `resolveRequest`, right before the exact-name stubs. This also hardens the widget bundler against any future deep imports into RN internals — none of them can work in a bridge-less JS context anyway.

Added a CHANGELOG entry under **Unpublished → Bug fixes**.

# Test plan

- Repro without a simulator, on an unmodified `npx create-expo-app --example with-widgets` (repro repo: https://github.com/usmsam/expo-widgets-context-error-repro):
  ```sh
  node node_modules/expo-widgets/scripts/build-bundle.mjs "$PWD"
  /System/Library/Frameworks/JavaScriptCore.framework/Versions/A/Helpers/jsc \
    node_modules/expo-widgets/bundle/build/ExpoWidgets.bundle
  ```
  Before: `Invariant Violation: __fbBatchedBridgeConfig is not set` (346 modules, 543 KB). After (this change applied to the installed package): bundle evaluates silently (104 modules, 147 KB). Control experiment confirmed causality in both directions.
- `__expoWidgetRender` exercised over a real app's layout: 12 widget states × `systemSmall`/`systemMedium`/`accessoryRectangular` × light/dark — 72/72 render to the expected view trees; widgets render on an iOS 26.5 simulator.
- Note: I could not run the monorepo toolchain (`et check-packages`) from a sparse clone; the change is config-only and mirrors the existing `react-native-worklets`/`react-native-reanimated` `{ type: 'empty' }` handling one switch-case below.

# Notes for reviewers

- On current `main` the crash reproduces the same way (verified against `main`'s `metro.config.js` — it has `getModulesRunBeforeMainModule: () => []` and the `createContext`/`useContext` stubs already, but not the deep-import handling; with the prelude removed the entry still crashes via `TextInputState → Platform.ios → NativePlatformConstantsIOS → TurboModuleRegistry`).
- An alternative/complementary fix is lazy-importing `TextInputState` inside `@expo/ui/src/keyboard` — but the resolver hardening protects the widget bundle from the whole class of deep-import regressions, so it seems worth having regardless.
- A related but separate issue observed while debugging (not addressed here): `bundle/index.ts` spreads `...ReactNative` after `...swiftUI` into `globalThis`, so the `Image` stub added on `main` shadows the `Image` component from `@expo/ui`, making `<Image systemName=... />` render as `{"resolveAssetSource":{}}`. Details in #49490.

### DIFF
--- a/packages/expo-widgets/metro.config.js
+++ b/packages/expo-widgets/metro.config.js
@@ -40,6 +40,17 @@ const buildConfig = {
       if (fileSpecifierRe.test(moduleName)) {
         return context.resolveRequest(context, moduleName, platform);
       }
+      // Deep imports into React Native internals (e.g.
+      // 'react-native/Libraries/Components/TextInput/TextInputState' pulled in
+      // via the @expo/ui barrel) slip past the exact-match stubs below and drag
+      // the whole RN core, including InitializeCore, into the widget bundle.
+      // The bundle runs in a bare JS context without the native bridge, where
+      // requiring RN's NativeModules throws "__fbBatchedBridgeConfig is not
+      // set" and every widget renders the "Could not create context for layout
+      // evaluation" placeholder. Resolve such deep imports as empty modules.
+      if (moduleName.startsWith('react-native/')) {
+        return { type: 'empty' };
+      }
       switch (moduleName) {
         case 'expo':
           return { type: 'sourceFile', filePath: expoStubPath };
--- a/packages/expo-widgets/metro.config.js
+++ b/packages/expo-widgets/metro.config.js
@@ -40,14 +40,6 @@ const buildConfig = {
       if (fileSpecifierRe.test(moduleName)) {
         return context.resolveRequest(context, moduleName, platform);
       }
-      // Deep imports into React Native internals (e.g.
-      // 'react-native/Libraries/Components/TextInput/TextInputState' pulled in
-      // via the @expo/ui barrel) slip past the exact-match stubs below and drag
-      // the whole RN core, including InitializeCore, into the widget bundle.
-      // The bundle runs in a bare JS context without the native bridge, where
-      // requiring RN's NativeModules throws "__fbBatchedBridgeConfig is not
-      // set" and every widget renders the "Could not create context for layout
-      // evaluation" placeholder. Resolve such deep imports as empty modules.
       if (moduleName.startsWith('react-native/')) {
         return { type: 'empty' };
       }
--- a/packages/expo-widgets/CHANGELOG.md
+++ b/packages/expo-widgets/CHANGELOG.md
@@ -25,6 +25,7 @@
 
 ### 🐛 Bug fixes
 
+- Resolve deep `react-native/*` imports as empty modules when bundling widget layouts, fixing every widget failing with "Could not create context for layout evaluation" after `@expo/ui` 57.0.14 introduced such an import. ([#49491](https://github.com/expo/expo/pull/49491) by [@usmsam](https://github.com/usmsam))
 - [iOS] Fix XCFramework precompilation failing on the unguarded `ActivityViewContext` parameter of `getLiveActivityEnvironment`, which requires iOS 16.1. ([#49076](https://github.com/expo/expo/pull/49076) by [@brentvatne](https://github.com/brentvatne))
 - [iOS] Quote script-phase paths so iOS builds work from a project path containing a space. ([#48747](https://github.com/expo/expo/pull/48747) by [@expo-bot](https://github.com/expo-bot))
 - [iOS] Fix `LiveActivityFactory.getInstances()` returning live activities that belong to other factories or that have already ended. ([#48489](https://github.com/expo/expo/pull/48489) by [@huextrat](https://github.com/huextrat))


packages/expo-widgets/metro.config.js
```diff
@@ -40,6 +40,17 @@ const buildConfig = {
...
         return context.resolveRequest(context, moduleName, platform);
       }
+      // Deep imports into React Native internals (e.g.
+      // 'react-native/Libraries/Components/TextInput/TextInputState' pulled in
+      // via the @expo/ui barrel) slip past the exact-match stubs below and drag
+      // the whole RN core, including InitializeCore, into the widget bundle.
+      // The bundle runs in a bare JS context without the native bridge, where
+      // requiring RN's NativeModules throws "__fbBatchedBridgeConfig is not
+      // set" and every widget renders the "Could not create context for layout
+      // evaluation" placeholder. Resolve such deep imports as empty modules.
```
jakex7:
```suggestion
```

jakex7 (changes requested):
Thank you for the fix!

jakex7 (approved):

usmsam:
Heads-up: `expo-widgets@57.0.14` ships this resolver fix, but the widget bundle still fails to evaluate — now with:

```
TypeError: (0,n.createContext) is not a function
```

The deep-import cut works (bundle drops from 346 to 104 modules), but `bundle/react-stub.ts` on the SDK 57 release branch doesn't have the `createContext`/`useContext` additions that already exist on `main` — and `@expo/ui`'s swift-ui `keyboard` module (the same one that introduced the deep import) needs them via `TextInputHostProvider`.

Verified on an unmodified `with-widgets` example with `expo-widgets@57.0.14`:

```sh
node node_modules/expo-widgets/scripts/build-bundle.mjs "$PWD"
/System/Library/Frameworks/JavaScriptCore.framework/Versions/A/Helpers/jsc \
  node_modules/expo-widgets/bundle/build/ExpoWidgets.bundle
# => TypeError: (0,n.createContext) is not a function
```

Appending main's `createContext`/`useContext` stubs to `bundle/react-stub.ts` makes the bundle evaluate cleanly (verified locally) — so cherry-picking that react-stub change from `main` onto the release branch should complete the fix. (`getModulesRunBeforeMainModule: () => []` from `main` is no longer strictly required once deep imports resolve to empty — the InitializeCore prelude resolves to an empty module — but it's harmless to include.)

Issue #41620, opened 2025-12-14
Expo's tree shaking crashes production with latest react-native-reanimated


RohovDmytro:
### Minimal reproducible example

https://github.com/RohovDmytro/expo-shaken-reanimated

### Steps to reproduce

App crashes when tree shaken enabled. Not sure if its due to side effects in reanimated or due to unstable nature of tree shaking as a feature flag.

### Steps to reproduce

There’s a problem with pairing expo’s unstable feature flags with latest `react-native-reanimated` version.

0. New expo project 

1. Set up tree shaking via `.env`:

```jsx
EXPO_UNSTABLE_METRO_OPTIMIZE_GRAPH=1
EXPO_UNSTABLE_TREE_SHAKING=1
EXPO_UNSTABLE_WEB_MODAL=1
```

1. Add 

```jsx
yarn add react-native-reanimated@4.2.0 react-native-worklets@0.7.1x
```

2. Wrap the App.js with an `Animate.View` component to make sure reanimated stays in the bundle.

3. Build an Android release build.

4. Crash:

```jsx
{ [WorkletsError: [Worklets] Native part of Worklets doesn't seem to be initialized.
```

### Environment

```text
expo-env-info 2.0.8 environment info:
    System:
      OS: macOS 26.2
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 24.11.1 - ~/.nvm/versions/node/v24.11.1/bin/node
      Yarn: 1.22.15 - /usr/local/bin/yarn
      npm: 11.6.2 - ~/.nvm/versions/node/v24.11.1/bin/npm
      Watchman: 2025.11.10.00 - /opt/homebrew/bin/watchman
    Managers:
      CocoaPods: 1.16.2 - /opt/homebrew/bin/pod
    SDKs:
      iOS SDK:
        Platforms: DriverKit 25.2, iOS 26.2, macOS 26.2, tvOS 26.2, visionOS 26.2, watchOS 26.2
      Android SDK:
        API Levels: 33, 34, 35, 36
        Build Tools: 35.0.0, 36.0.0, 36.1.0
        System Images: android-31 | Google Play ARM 64 v8a
    IDEs:
      Android Studio: 2025.2 AI-252.27397.103.2522.14514259
      Xcode: 26.2/17C52 - /usr/bin/xcodebuild
    npmPackages:
      expo: ~54.0.29 => 54.0.29 
      react: 19.1.0 => 19.1.0 
[...]

PR #49492, opened 2026-08-28
fix(metro-config): Keep side-effectful initializers of unused exports

hknakn:
# Why

`treeShakeSerializerPlugin` deletes an unused `export const` without inspecting its initializer, so a side-effectful initializer is dropped along with the binding.

`culori` registers its color spaces exactly this way:

```js
// culori/src/index.js
export const rgb = useMode(modeRgb); // useMode() writes to a shared mode registry
```

Nothing imports `rgb`, so the declarator is removed, the registry stays empty, and a later `interpolate()` throws `TypeError: Cannot read property 'channels' of undefined`.

It reached our app through `uniwind` → `heroui-native`, which use culori to resolve `color-mix()` CSS variables. Every themed `<Button>` threw during render, the root error boundary unmounted the tree, and the splash screen was never hidden — so the app looked frozen rather than crashed, which made it hard to trace back to tree shaking.

Rollup and webpack only drop such an initializer when it carries a `#__PURE__` annotation; culori has none. No Babel plugin is involved — plain ESM in `node_modules` is enough to reproduce.

Same "side effect dropped by the shaker" family as #41620, though this PR does not fix that one.

# How

Adds `canRemoveInitializer()` and gates both removal branches (`export const …` and `export default …`) on it.

An unused export is still removed when its initializer is observably inert: literals, functions and arrows, identifiers, `#__PURE__`-annotated calls, and structurally pure object/array/member/unary/binary/conditional expressions. Anything else — most importantly a bare call — is kept.

Deliberately conservative: it can retain slightly more than strictly necessary, but it never drops an observable side effect.

# Test Plan

Three tests added to `tree-shaking.test.ts`:

- a side-effectful initializer is preserved (fails without this change)
- a side-effect-free initializer is still removed
- a `#__PURE__`-annotated initializer is still removed

`pnpm exec jest` in `packages/@expo/metro-config`:

| | Tests |
|---|---|
| `main` | 57 failed, 4 skipped, 1291 passed, 1352 total |
| this PR | 57 failed, 4 skipped, 1294 passed, 1355 total |

Identical failures on both — they come from `babel-preset-expo` not building in my checkout (a type-only `expo-splash-screen/plugin` import) and are unrelated to this change. The only delta is the 3 added tests.

Also verified against a real SDK 57 app, iOS release build on device, with `EXPO_UNSTABLE_TREE_SHAKING=1 EXPO_UNSTABLE_METRO_OPTIMIZE_GRAPH=1`: it now boots instead of hanging on the splash screen, and the bundle still shrinks (11.80 MB → 10.07 MB of Hermes bytecode).

# Checklist

- [ ] Not applicable — no prebuild/module plugin changes
- [ ] Not applicable — no documentation changes

cc @kitten @EvanBacon

### DIFF
--- a/packages/@expo/metro-config/CHANGELOG.md
+++ b/packages/@expo/metro-config/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Depend on `@react-native/js-polyfills` directly for `getPolyfills` instead of the `react-native/rn-get-polyfills` subpath removed in React Native 0.88. ([#48034](https://github.com/expo/expo/pull/48034) by [@alanjhughes](https://github.com/alanjhughes))
 - Fix source line counts after environment serializer plugins modify virtual modules ([#48835](https://github.com/expo/expo/pull/48835) by [@kitten](https://github.com/kitten))
 - Seal web worker chunks to prevent common chunk splitting from applying to them ([#49227](https://github.com/expo/expo/pull/49227) by [@kitten](https://github.com/kitten))
+- Keep side-effectful initializers when removing unused exports during tree shaking ([#49492](https://github.com/expo/expo/pull/49492) by [@hknakn](https://github.com/hknakn))
 
 ### 💡 Others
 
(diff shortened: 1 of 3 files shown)


joernroeder:
@kitten can you take a look to get us uniwind folks unblocked? thank you!

PR #49494, opened 2026-08-28
[dev-launcher][ios] Answer the fingerprint check from a debug build

vonovak:
# Why

`devicectl` exposes no equivalent of `simctl get_app_container`, so the fingerprint embedded in an app on a physical iPhone cannot be read from the filesystem. The app has to answer for itself. This PR is the responder; the client that triggers it is `npx @expo/agent-cli status --explain --device <phone>` (expo/expo-agent-cli#61).

# How

- `EXDevLauncherFingerprintCheck` handles a trigger URL and posts the embedded fingerprint to the callback URL the trigger carries, matched by a one-time nonce. Debug builds only; in a release build the URL flows through normal handling.
- **The trigger is selected by a reserved `__expo_fingerprint_check=1` query parameter, never by a URL host.** A host reads as a destination and would take a name out of the app's own route namespace — with a host, expo-router would resolve the trigger to a route of that name (`extractPathFromURL.ts`). The reserved `__expo_*` namespace is the direction recorded in the Expo Linking investigation doc. `__expo_fingerprint_nonce` and `__expo_fingerprint_callback` carry the rest.
- `EmbeddedFingerprint` in `expo-modules-core` reads the stamp for it.
- `expo-linking` filters the trigger out of `getLinkingURL()`, `useLinkingURL()` and its `url` event, so a check never looks like a deep link. React Native's own `Linking` still sees it.

The callback URL arrives from outside the app, so it is refused unless its scheme is `http`, its path is the callback path, and its host is an IP literal in a private or link-local range; redirects are refused too.

# Test Plan

- `EXDevLauncherFingerprintCheckTests` — 20 cases covering the parse and the guards: a trigger on an app-route host and on no host at all (the marker alone selects the channel), a missing marker, a marker with another value, a URL with no query, an empty nonce, an https callback, a wrong path, a public address, a loopback address, and a DNS name.
- The protocol parity test pins the reserved parameter names against this Swift and against the expo-linking filter, and asserts neither file matches on a host.
- **`et native-unit-tests -p ios --packages expo-dev-launcher` passes: 100 test cases, 0 failures.**

A live round trip needs this build on a device together with the client, so it is not covered here.

# Checklist


packages/expo-dev-launcher/ios/EXDevLauncherFingerprintCheck.swift
```diff
@@ -0,0 +1,130 @@
...
+/**
+ * True when `host` is an IP literal in a private or link-local range, resolving no DNS names.
+ * This is the SSRF guard: without it a web page could deep-link a debug build with a callback
+ * pointing anywhere.
+ */
+internal func isPrivateAddress(_ host: String) -> Bool {
+  var ipv4Addr = in_addr()
+  if host.withCString({ inet_pton(AF_INET, $0, &ipv4Addr) }) == 1 {
+    let octets = withUnsafeBytes(of: ipv4Addr.s_addr) { Array($0) }
+    return isPrivateIPv4(octets[0], octets[1])
```
Kudo, line 45: **Endian bug (blocker):** `withUnsafeBytes(of: ipv4Addr.s_addr)` reads the `UInt32` in host byte order, so on little-endian the octets are reversed and private LAN callbacks (e.g. `192.168.x.x`) get rejected.

Prefer `withUnsafeBytes(of: &ipv4Addr) { Array($0) }` or `UInt32(bigEndian: ipv4Addr.s_addr)` + shifts. IPv6 path below is already correct.

Kudo:
@expo-bot verify --fable

Kudo (changes requested):
we should prevent custom host like `expo-fingerprint-check`. maybe use query strings instead. refer to https://app.notion.com/p/expo/Expo-Linking-investigation-37ae5b57352480bba6b6d6006f24b2ae?source=copy_link

packages/expo-dev-launcher/ios/EXDevLauncherFingerprintCheck.swift
vonovak: I don't think this is a bug — I checked. We're never turning the bytes into a number. But I changed this so it reads better.

packages/expo-constants/scripts/__tests__/fingerprintProtocolParity-test.js
```diff
@@ -38,3 +38,62 @@ describe('app.fingerprint protocol parity', () => {
     expect(contents).toContain('silent: true');
   });
 });
+
+// The fingerprint check is a protocol across two repositories. The other side is
+// `fingerprintCheckProtocol.ts` in expo/expo-agent-cli; neither side can import the other.
+describe('fingerprint-check protocol parity', () => {
```
Kudo: ditto with the other pr. the js parser is fragile. we can just drop it and use the native test instead

packages/expo-dev-launcher/ios/ReactDelegateHandler/ExpoDevLauncherAppDelegateSubscriber.swift
```diff
@@ -9,6 +9,11 @@ public class ExpoDevLauncherAppDelegateSubscriber: ExpoAppDelegateSubscriber {
     didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]? = nil
   ) -> Bool {
     EXDevLauncherController.disablePackagerServerAccess()
+    // Legacy non-scene launches only. It can fire twice, which is harmless: same nonce, and the
+    // second POST finds the callback port closed.
```
Kudo: unclear for the comment. what is "legacy" and why it fires twice?

packages/expo-dev-launcher/ios/Tests/EXDevLauncherFingerprintCheckTests.swift
```diff
@@ -0,0 +1,154 @@
+// Copyright 2015-present 650 Industries. All rights reserved.
+
+import XCTest
```
Kudo: can we use swift testing framework here? https://developer.apple.com/xcode/swift-testing/

packages/expo-linking/ios/LinkingAppDelegateSubscriber.swift
```diff
@@ -6,6 +6,9 @@ public class LinkingAppDelegateSubscriber: ExpoAppDelegateSubscriber {
   public func application(_ app: UIApplication, open url: URL, options: [UIApplication.OpenURLOptionsKey: Any] = [:])
     -> Bool
   {
+    if isFingerprintCheckURL(url) {
+      return false
+    }
```
Kudo, line 13: add comment that we have early return here. after code in the future other than initialURL or notification will not be handled. be careful to add new code after the early return

packages/expo-linking/ios/LinkingAppDelegateSubscriber.swift
```diff
@@ -15,6 +18,9 @@ public class LinkingAppDelegateSubscriber: ExpoAppDelegateSubscriber {
     guard let url = urls.first else {
       return
     }
+    if isFingerprintCheckURL(url) {
+      return
```
Kudo, line 26: ditto early return

packages/expo-linking/ios/LinkingAppDelegateSubscriber.swift
```diff
@@ -38,3 +44,12 @@ public class LinkingAppDelegateSubscriber: ExpoAppDelegateSubscriber {
...
   }
 }
+
+/// A dev-launcher command, not a deep link, so it never becomes the initial URL or reaches JS.
+/// React Native's own Linking module still sees it.
+private func isFingerprintCheckURL(_ url: URL) -> Bool {
+  guard let queryItems = URLComponents(url: url, resolvingAgainstBaseURL: false)?.queryItems else {
+    return false
+  }
+  return queryItems.contains { $0.name == "__expo_fingerprint_check" && $0.value == "1" }
```
Kudo: nit: makes me thinking that whether we should move `__expo_fingerprint_check` into expo-modules-core as a constant?

Kudo (changes requested):
i like the pr getting cleaner. i think we are near to merge. thanks for following up

packages/expo-dev-launcher/ios/ReactDelegateHandler/ExpoDevLauncherAppDelegateSubscriber.swift
vonovak: reworded the comment

packages/expo-dev-launcher/ios/ReactDelegateHandler/ExpoDevLauncherAppDelegateSubscriber.swift
```diff
@@ -9,6 +9,11 @@ public class ExpoDevLauncherAppDelegateSubscriber: ExpoAppDelegateSubscriber {
     didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]? = nil
   ) -> Bool {
     EXDevLauncherController.disablePackagerServerAccess()
+    // iOS also delivers a cold-launch URL to `application(_:open:)`, so the check can run twice.
+    // Harmless: the callback server is one-shot, so the second POST finds the port closed.
```
Kudo: that's...not great. hopefully we can find some way to do it better.

packages/expo-linking/ios/LinkingAppDelegateSubscriber.swift
```diff
@@ -6,6 +6,10 @@ public class LinkingAppDelegateSubscriber: ExpoAppDelegateSubscriber {
   public func application(_ app: UIApplication, open url: URL, options: [UIApplication.OpenURLOptionsKey: Any] = [:])
     -> Bool
   {
+    // Code below never sees a trigger URL, so anything that must see every URL goes above.
```
Kudo:
```suggestion
    // Code below does not run for the trigger URL.
    // Put anything that must run for every URL above this return.
```

packages/expo-linking/ios/LinkingAppDelegateSubscriber.swift
```diff
@@ -15,6 +19,10 @@ public class LinkingAppDelegateSubscriber: ExpoAppDelegateSubscriber {
     guard let url = urls.first else {
       return
     }
+    // Code below never sees a trigger URL, so anything that must see every URL goes above.
```
Kudo:
```suggestion
    // Code below does not run for the trigger URL.
    // Put anything that must run for every URL above this return.
```

packages/expo-modules-core/ios/Utilities/FingerprintCheckProtocol.swift
```diff
@@ -0,0 +1,29 @@
+// Copyright 2015-present 650 Industries. All rights reserved.
+
+/**
+ The literals of the fingerprint-check protocol, shared by the dev-launcher responder and the
+ expo-linking filter that keeps the trigger out of app navigation.
+
+ The other side is `fingerprintCheckProtocol.ts` in expo/expo-agent-cli. Neither repository can
+ import the other, so a change here has to be made there too.
+ */
+public enum FingerprintCheckProtocol {
```
Kudo: can we not introduce new class (new enum) and reuse EmbeddedFingerprint?

Kudo (changes requested):
still some small tweak

packages/expo-linking/ios/LinkingAppDelegateSubscriber.swift
```diff
@@ -6,6 +6,10 @@ public class LinkingAppDelegateSubscriber: ExpoAppDelegateSubscriber {
   public func application(_ app: UIApplication, open url: URL, options: [UIApplication.OpenURLOptionsKey: Any] = [:])
     -> Bool
   {
+    // Code below never sees a trigger URL, so anything that must see every URL goes above.
+    if isFingerprintCheckURL(url) {
```
Kudo: **[bug]** The native filter only skips `ExpoLinkingRegistry.initialURL` and `onURLReceived`. It does not stop `RCTLinkingManager`, and `expo-linking`'s JS `addEventListener('url')` is a pass-through to React Native Linking. expo-router's live subscriber uses that API (`packages/expo-router/src/link/linking.ts`), then `extractPathFromURL` turns `scheme://login?__expo_fingerprint_check=1&...` into a `/login?...` navigation. Cold start is fine because iOS `getInitialURL()` uses `getLinkingURL()`. Warm start (app already running, the usual `status --explain --device` case) will navigate the current screen. Scene-based apps always re-notify `RCTLinkingManager` after subscribers, so returning true from this method would not help.

**Suggestion:** Drop fingerprint-check URLs in `expo-linking`'s JS `addEventListener('url')` (same marker check, debug-only) so expo-router and `useURL()` never see them. Optionally also skip `RCTLinkingManager` in `SceneEventForwarder` when the URL is a trigger, for apps that subscribe to RN Linking directly.

packages/expo-linking/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 ### 🎉 New features
 
+- [iOS] Ignore the fingerprint-check trigger URL, so it never reaches `getLinkingURL()` or the `url` event. ([#49494](https://github.com/expo/expo/pull/49494) by [@vonovak](https://github.com/vonovak))
```
Kudo: **[suggestion]** The entry says the trigger never reaches `getLinkingURL()` or the `url` event. `getLinkingURL()` / `useLinkingURL()` are filtered. The `url` event is not: `addEventListener('url')` still delegates to React Native Linking, which is what the expo-linking docs call the `url` event.

**Suggestion:** Reword to name `getLinkingURL()` / `useLinkingURL()` / `onURLReceived` only, and do not claim the `url` event is filtered unless Issue 1 is fixed.

Kudo (commented):
## Summary

The iOS debug responder is tightly scoped and the SSRF guard looks correct: HTTP only, private/link-local IP literals (no DNS), fixed callback path, no redirects, `#if DEBUG` on both the POST and the linking swallow. Parse tests cover the important accept/reject matrix, and the IPv4 endian handling (`UInt32(bigEndian:)`) is right on Darwin. The remaining gap is that the "not a deep link" claim is only true for `getLinkingURL()` / `useLinkingURL()`; expo-router still consumes the trigger on a warm start through React Native's `url` event.

## Issue counts by severity

- bugs: 1
- suggestions: 1
- nits: 0

packages/expo-dev-launcher/ios/ReactDelegateHandler/ExpoDevLauncherAppDelegateSubscriber.swift
```diff
@@ -9,6 +9,11 @@ public class ExpoDevLauncherAppDelegateSubscriber: ExpoAppDelegateSubscriber {
     didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]? = nil
   ) -> Bool {
     EXDevLauncherController.disablePackagerServerAccess()
+    // iOS also delivers a cold-launch URL to `application(_:open:)`, so this can run twice.
+    // `handle` answers a nonce once, so the second call posts nothing.
+    if let url = launchOptions?[.url] as? URL {
+      _ = EXDevLauncherFingerprintCheck.handle(url)
+    }
```
Kudo: i tried to ask agents how does ios app (non react-native) handle this between didFinishingLaunching and open. here's the finding, maybe we can just drop the didFinishingLaunching path. not sure whether it's true

> This still does not need to run in `didFinishLaunching`.
> 
> `didFinishLaunching` is launch context, not URL delivery. For a custom-scheme cold start, UIKit puts the URL in `launchOptions` and later calls
> `application(_:open:)`. That already goes through `onDeepLink` → `handle`. The nonce guard stops a second POST, but it does not make this first call necessary.
> 
> On scene-based Expo apps this is a no-op anyway: `launchOptions[.url]` is empty. The URL arrives in `connectionOptions.urlContexts`, and `SceneEventForwarder`
> turns that into one `application(_:open:)` call. Device builds already depend on `open`.
> 
> I'd drop this block and keep the POST only in `onDeepLink`. The trigger is a custom scheme (`exp+app://...?__expo_fingerprint_check=1`), not a universal link, so
> `continueUserActivity` is not involved.

packages/expo/ios/Expo/SceneEventForwarder.swift
```diff
@@ -23,6 +24,15 @@ struct SceneEventForwarder {
     let application = UIApplication.shared
     let delegate = appDelegate()
 
+#if DEBUG
+    // Scene apps only, cold start and warm alike: a scene routes every URL through here.
+    // React Native must not see the link, or expo-router routes to it.
```
Kudo:
```suggestion
```

packages/expo-dev-launcher/ios/EXDevLauncherFingerprintCheck.swift
```diff
@@ -0,0 +1,128 @@
...
+          !nonce.isEmpty,
+          let callbackString = queryItems.first(where: {
+            $0.name == EmbeddedFingerprint.CheckProtocol.callbackParam
+          })?.value,
+          let callback = URL(string: callbackString),
+          // The CLI never emits https, and https to an IP literal would fail TLS anyway.
+          callback.scheme == "http",
+          callback.path == EmbeddedFingerprint.CheckProtocol.callbackPath,
+          let callbackHost = callback.host,
+          isPrivateAddress(callbackHost) else {
```
Kudo: that's very...astra. rewrite in cleaner code?

Kudo (approved):

packages/expo-dev-launcher/ios/EXDevLauncherFingerprintCheck.swift
vonovak: that was opus actually :))

Kudo: haha ok good to know. thanks for the following up

PR #49504, opened 2026-08-28
[ios][updates] Reuse existing asset rows on key conflict instead of replacing them

alanjhughes:
# Why
Follow up from the review of #49458 

assets.key is unique, and addNewAssets inserted with INSERT OR REPLACE. On a key conflict, the REPLACE deletes the existing row before inserting, and we run with foreign keys enforced, so the delete cascades. Any update using that asset as its launch asset is deleted through updates.launch_asset_id ON DELETE CASCADE, and every other update sharing the asset loses its updates_assets rows. One registration can silently erase or gut updates it has nothing to do with.

# How
When registering a new asset whose key already exists, adopt the existing row instead of replacing it. addNewAssetsInternal now calls the existing addExistingAsset first, which links the join row and sets launch_asset_id when flagged, and only falls through to the insert when the key is absent. The check runs inside the registration transaction, so it cannot race. The insert itself drops OR REPLACE, so any conflict that could still occur aborts the transaction rather than deleting rows.

# Test Plan
New tests in UpdatesDatabaseTests

### DIFF
--- a/packages/expo-updates/ios/Tests/UpdatesDatabaseTests.swift
+++ b/packages/expo-updates/ios/Tests/UpdatesDatabaseTests.swift
@@ -459,6 +459,157 @@ class UpdatesDatabaseTests {
     }
   }
 
+  @Suite("asset key conflicts", .serialized)
+  struct AssetKeyConflictTests {
+    var testDatabaseDir: URL
+    var db: UpdatesDatabase
+    var config: UpdatesConfig
+
+    init() throws {
+      let applicationSupportDir = FileManager.default.urls(for: .applicationSupportDirectory, in: .userDomainMask).last
+      testDatabaseDir = applicationSupportDir!.appendingPathComponent("AssetKeyConflictTests")
+
+      try? FileManager.default.removeItem(atPath: testDatabaseDir.path)
+
+      if !FileManager.default.fileExists(atPath: testDatabaseDir.path) {
+        try FileManager.default.createDirectory(atPath: testDatabaseDir.path, withIntermediateDirectories: true)
+      }
+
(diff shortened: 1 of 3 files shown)


packages/expo-updates/ios/Tests/UpdatesDatabaseTests.swift
```diff
@@ -459,6 +459,125 @@ class UpdatesDatabaseTests {
...
+      }
+
+      db.databaseQueue.sync {
+        let survivor = try! db.update(withId: update1.updateId, config: config)
+        #expect(survivor != nil)
+        #expect(survivor?.status == .StatusReady)
+      }
+      #expect(assetRowCount(forKey: "shared-bundle") == 1)
+      #expect(joinRowCount(forUpdateId: update1.updateId) == 1)
+      #expect(joinRowCount(forUpdateId: update2.updateId) == 1)
```
Kudo, line 575: **[suggestion]** The launch-asset conflict test asserts that update1 still exists as `.StatusReady` and that both updates have a join row, but it never checks `launch_asset_id`. `addExistingAsset` is what is supposed to set that column on the new update when `isLaunchAsset` is true; a regression that only linked `updates_assets` would still pass. The Android counterpart in #49505 asserts `loadLaunchAssetForUpdate` for both updates.

**Suggestion:** After the conflict, assert `launch_asset_id` is non-null for both update1 and update2 (for example `SELECT launch_asset_id FROM updates WHERE id = ?` or `assets(withUpdateId:)` with `isLaunchAsset`). That locks in the adopt path, not just join-row survival.

Kudo (approved):

PR #49508, opened 2026-08-28
fix(expo-updates): Use explicit env modes in CLI commands

ramonclaudio:
# Why

Expo Updates commands can read app config before loading env files or use values left over from the parent process.

# How

I updated the commands to clear inherited dotenv values and load the right env files before app config, using production by default unless the caller passes a mode.

# Test Plan

All tests pass.


ramonclaudio:
@expo-bot verify

ramonclaudio:
@expo-bot verify

ramonclaudio:
@expo-bot verify

Kudo (approved):
looks good!

nit: though it passes `mode: production` in real code. some test code or `?? 'development'` fallback is still in development. given it's all updates related code. what do you think to make them as production and being consistent?

ramonclaudio:
> I think the `?? 'development'` fallback should stay though because that’s used by Expo Go, which should use development by default? WDYT? @Kudo

i don't know. hard to tell if people use expo-go to test updates, which is for production aligned behavior or development.

maybe better way is what we talked about doc & convention on the sdk sync, and even have a way for people to specify the config mode. since that's something for later, i don't have strong opinion whether to use development or production on the fallback here. please pick your favorite!

ramonclaudio:
@expo-bot verify

ramonclaudio:
@expo-bot verify --fable

packages/expo-updates/cli/src/cli.ts
```diff
@@ -7,6 +8,8 @@ import { boolish } from 'getenv';
 import { logCmdError } from './utils/errors';
 import * as Log from './utils/log';
 
+process.env = getOriginalEnv();
```
kitten: This looks like a noop? There shouldn't be an original env presumably in this, since we would not fork to it, and if it's forked it cannot restore an env?

See comment on CLI

packages/@expo/cli/src/start/server/middleware/ExpoGoManifestHandlerMiddleware.ts
```diff
@@ -126,6 +126,7 @@ export class ExpoGoManifestHandlerMiddleware extends ManifestMiddleware<ExpoGoMa
       (await resolveRuntimeVersionWithExpoUpdatesAsync({
         projectRoot: this.projectRoot,
         platform: requestOptions.platform,
+        mode: this.options.mode ?? 'development',
```
kitten: This change top-to-bottom feels a bit excessive. We can just make sure that `runtimeversion:resolve` runs as `development`. Generally, if we check where commands were run before, we can ensure they're consistent.

Here, the old value would've been `development`, so if we preserve that this file and `resolveRuntimeVersionWithExpoUpdatesAsync` need no change.

packages/@expo/cli/src/utils/expoUpdatesCli.ts
```diff
@@ -19,10 +24,15 @@ export async function expoUpdatesCommandAsync(projectDir: string, args: string[]
   }
 
   try {
+    const commandEnv = getOriginalEnv();
+    setNodeEnv(mode, { systemEnv: commandEnv });
+    commandEnv.__EXPO_CONFIG_MODE = mode;
```
kitten, line 28: Overriding here is fine, but assuming we can default `mode = 'development'`, since this was only used in one place, the other changes in the CLI go away.

We don't need `setNodeEnv` on `commandEnv`, right? We'd just, as far as I remember, need `getOriginalEnv()` then `__EXPO_CONFIG_MODE = mode` (assuming `development` default) then pass that on.

kitten (commented):
This looks correct to me.

PR #49509, opened 2026-08-28
[ui] Drop vaul from web BottomSheet

Kudo:
# Why

vaul is unmaintained. `@expo/ui` used it for the community and universal `BottomSheet` on web. since vaul is removed from router since #49204, we should also remove vaul from expo-ui.
close ENG-26258

# How

replace vaul with a `<dialog>` sheet shared by community and universal.

# Test Plan

- start dev-server on bare-expo and run `pnpm run test:e2e:web` in `packages/expo-ui` manually
- manual test on NCL for both community and universal BottomSheet


packages/expo-ui/src/web/BottomSheetDialog.tsx
```diff
@@ -0,0 +1,496 @@
...
+    cursor: grab;
+    touch-action: none;
+  }
+  [data-expo-ui-bottom-sheet-handle][data-grabbing='true'] {
+    cursor: grabbing;
+  }
+`;
+
+function lockBodyScroll() {
+  const scrollY = window.scrollY;
```
Kudo, line 254: Fixed. Nested open sheets share one lock counter and restore the original body styles only when the last sheet closes. id:de5ce1442ca5

packages/expo-ui/src/web/BottomSheetDialog.tsx
```diff
@@ -0,0 +1,496 @@
...
+  const { overflow, position, top, width, touchAction } = document.body.style;
+  document.body.style.overflow = 'hidden';
+  document.body.style.position = 'fixed';
+  document.body.style.top = `-${scrollY}px`;
+  document.body.style.width = '100%';
+  document.body.style.touchAction = 'none';
```
Kudo: Removed `touch-action: none` from the body lock. Page scroll stays blocked with `position: fixed` and `overflow: hidden`. id:a8542dd1d5a9

packages/expo-ui/src/web/BottomSheetDialog.tsx
```diff
@@ -0,0 +1,496 @@
...
+          <View
+            ref={handleRef}
+            testID={BOTTOM_SHEET_TEST_IDS.handle}
+            dataSet={{ expoUiBottomSheetHandle: '', grabbing: dragging ? 'true' : undefined }}
+            style={[styles.handle, { backgroundColor: handleBackground }]}
+          />
+        ) : null}
+        <View
+          testID={innerTestID}
+          style={[styles.body, renderedHeight != null && styles.bodyFlex, bodyStyle]}>
```
Kudo, line 585: The content-sized body now uses `flexShrink: 1` with `minHeight: 0`, so overflow can scroll once the panel hits `max-height: 85dvh`. id:a039d1175d63

packages/expo-ui/tsconfig.json
```diff
@@ -7,5 +7,6 @@
     "rootDir": "./src",
     "emitDeclarationOnly": true
   },
-  "include": ["./src/**/*"]
+  "include": ["./src/**/*"],
+  "exclude": ["**/__tests__/**"]
```
Kudo: Dropped the `__tests__` exclude. `pnpm typecheck` covers the test files again. id:a8caa085eb8b

packages/expo-ui/src/universal/BottomSheet/__tests__/BottomSheet.test.web.tsx
```diff
@@ -0,0 +1,107 @@
...
+        <Text>Padded</Text>
+      </BottomSheet>
+    );
+
+    await waitFor(() => {
+      expect(screen.getByTestId('padded-content')).toBeTruthy();
+    });
+
+    const content = screen.getByTestId('padded-content');
+    expect(content.style.paddingTop).toBe('16px');
```
Kudo: Dropped the inline `style.paddingTop` asserts. Padding math stays in `resolveContentPadding` tests; the sheet test now just checks that `contentPadding={0}` still renders. id:1c5f8535f2fc

docs/pages/versions/unversioned/sdk/ui/drop-in-replacements/bottomsheet.mdx
```diff
@@ -244,23 +244,23 @@ export default function BottomSheetScrollableExample() {
 
 `@gorhom/bottom-sheet` renders inline at the bottom of its parent view. This component uses native modal presentation on Android and iOS, and a drawer overlay on web.
```
Kudo: Updated that sentence to HTML `<dialog>`, matching the rest of the page. id:c6cf2669fe13

packages/expo-ui/src/universal/BottomSheet/index.tsx
```diff
@@ -43,83 +44,55 @@ export function BottomSheet({
...
-            }}
-            data-testid={testID}>
-            {children}
-          </div>
-        </Drawer.Content>
-      </Drawer.Portal>
-    </Drawer.Root>
+      showHandle={showDragIndicator}
+      height={snapHeights?.[snapIndex]}
+      minSnapHeight={snapHeights?.[0]}
```
Kudo: Fixed. `minSnapHeight` is now `Math.min(...snapHeights)` on both the universal and community web sheets, so a descending snap list can still reach the smaller detent. id:9a29db10464e

packages/expo-ui/e2e/web/bottom-sheet.spec.ts
```diff
@@ -0,0 +1,268 @@
...
+    await expect(openPanel(page).getByText('BottomSheetModal', { exact: true })).toBeVisible();
+    await expect(openPanel(page)).toBeVisible();
+  });
+
+  test('should open a fit-to-content modal shorter than the 90% snap sheet', async ({ page }) => {
+    await gotoCommunity(page);
+
+    await clickFirstButtonAfter(page, 'snapPoints: 25%, 50%, 90%');
+    await openPanel(page).getByRole('button', { name: 'Snap 2' }).click();
+    const snapBox = await openPanel(page).boundingBox();
```
Kudo, line 221: Fixed. The fit-to-content test now polls until the 90% snap height settles before it captures `snapBox`. id:8c58eb6066f2

Kudo:
@expo-bot review again?

Kudo:
@expo-bot review

packages/expo-ui/src/web/BottomSheetDialog.tsx
```diff
@@ -0,0 +1,524 @@
...
+        commitClose();
+        return;
+      }
+      onDragEnd?.(predicted);
+    };
+
+    sheet.addEventListener('pointerdown', onPointerDown);
+    window.addEventListener('pointermove', onPointerMove);
+    window.addEventListener('pointerup', onPointerUp);
+    window.addEventListener('pointercancel', onPointerUp);
```
Kudo: Fixed. `pointercancel` now aborts the drag session and does not call `commitClose` or `onDragEnd`. id:402703b52a4a

packages/expo-ui/src/web/BottomSheetDialog.tsx
```diff
@@ -0,0 +1,531 @@
...
+      dragRef.current = null;
+      setDragHeight(null);
+    };
+
+    sheet.addEventListener('pointerdown', onPointerDown);
+    window.addEventListener('pointermove', onPointerMove);
+    window.addEventListener('pointerup', onPointerUp);
+    window.addEventListener('pointercancel', onPointerCancel);
+    return () => {
+      sheet.removeEventListener('pointerdown', onPointerDown);
```
Kudo, line 512: Fixed. Pointer down now captures the pointer, `lostpointercapture` aborts the session, and effect cleanup clears drag state. id:86ad56bd90b0

packages/expo-ui/src/web/__tests__/BottomSheetDialog.test.web.tsx
```diff
@@ -0,0 +1,171 @@
...
+        minSnapHeight={200}>
+        <Text>Body</Text>
+      </BottomSheetDialog>
+    );
+
+    const panel = await waitFor(() => screen.getByTestId('expo-ui-bottom-sheet'));
+    act(() => {
+      dispatchPointer(panel, 'pointerdown', { pointerId: 1, clientY: 100, button: 0 });
+      dispatchPointer(window, 'pointermove', { pointerId: 1, clientY: 180 });
+      dispatchPointer(window, 'pointercancel', { pointerId: 1, clientY: 500 });
```
Kudo, line 141: Fixed. The abort tests now dispatch `pointerup` after cancel. If the abort listeners were missing, that up would dismiss and the test would fail. id:1a2b34012b43

packages/expo-ui/src/web/BottomSheetDialog.tsx
```diff
@@ -0,0 +1,540 @@
...
+      window.removeEventListener('pointermove', onPointerMove);
+      window.removeEventListener('pointerup', onPointerUp);
+      window.removeEventListener('pointercancel', abortDrag);
+      sheet.removeEventListener('lostpointercapture', abortDrag);
+      dragRef.current = null;
+      setDragHeight(null);
+    };
+  }, [open, mounted, resolveStartHeight, minSnapHeight, dismissible, commitClose, onDragEnd]);
```
Kudo: Fixed. The pointer effect depends only on `open` and `mounted`. Live `onOpenChange` / `onDragEnd` / snap values come from refs, so a parent re-render does not tear down an active drag. id:c2cf9231fec4

packages/expo-ui/src/universal/BottomSheet/index.tsx
```diff
@@ -43,83 +44,55 @@ export function BottomSheet({
...
-  const hasSnapPoints = vaulSnapPoints != null;
+  const { height: viewportHeight } = useWindowDimensions();
+  const snapHeights = useMemo(() => {
+    if (!snapPoints?.length) return undefined;
+    return snapPoints.map((point) => snapPointToHeightPx(point, viewportHeight));
+  }, [snapPoints, viewportHeight]);
+  const [snapIndex, setSnapIndex] = useState(0);
+
+  useEffect(() => {
+    if (isPresented) setSnapIndex(0);
```
Kudo: Fixed. Opening now resets `snapIndex` during render (same update as `isPresented` going true), so the first paint is already the first snap. Height also clamps when `snapPoints` shrinks.

id:6bf4ff5fea19

packages/expo-ui/src/web/__tests__/BottomSheetDialog.test.web.tsx
```diff
@@ -0,0 +1,309 @@
...
+        <button type="button" onClick={onClick}>
+          Snap 2
+        </button>
+      </BottomSheetDialog>
+    );
+
+    const button = await waitFor(() => screen.getByRole('button', { name: 'Snap 2' }));
+    act(() => {
+      dispatchPointer(button, 'pointerdown', { pointerId: 1, clientY: 100, button: 0 });
+      dispatchPointer(button, 'pointerup', { pointerId: 1, clientY: 100 });
```
Kudo: Fixed. The test now sends a `pointermove` past the drag threshold and asserts `onOpenChange` was not called. Without the interactive-target guard that move would dismiss the sheet.

id:c6713fb332a5

packages/expo-ui/src/web/BottomSheetDialog.tsx
```diff
@@ -0,0 +1,594 @@
...
+const Dialog = createWebComponent('dialog');
+
+export type BottomSheetDialogProps = {
+  open: boolean;
+  onOpenChange: (open: boolean) => void;
+  dismissible?: boolean;
+  showHandle?: boolean;
+  /** Pixel height for snap-point sheets. Omit to size to content. */
+  height?: number;
+  /** Close when a drag ends below this height. Defaults to a small px threshold. */
```
Kudo: Fixed. The comment now says the sheet closes a small px margin below `minSnapHeight`, and that without it the same margin is measured from 0.

id:6ecb2cfbaf0c

packages/expo-ui/src/web/BottomSheetDialog.tsx
```diff
@@ -0,0 +1,598 @@
...
+        testID={BOTTOM_SHEET_TEST_IDS.panel}
+        dataSet={{
+          expoUiBottomSheetPanel: '',
+          state: dataState,
+          sized: renderedHeight != null ? 'true' : undefined,
+        }}
+        style={[
+          styles.sheet,
+          style,
+          renderedHeight != null && { height: renderedHeight },
```
intergalacticspacehighway, line 567: nit(performance): can be later improved to set transform instead of height while dragging

intergalacticspacehighway (commented):
These came up during AI review, please ignore if not relevant.

### 1. setPointerCapture fires on pointerdown, before the 8px threshold.

The only guard is isInteractiveTarget, which matches button, a, input, textarea, select, [role="button"]. RNW renders Pressable/TouchableOpacity as a bare div with no role, so a plain <Pressable onPress> inside a sheet gets its pointer captured by the panel on every tap.

That likely breaks onPress, because RNW delivers onPress via the native click event rather than the responder system (see the NOTE comment in usePressEvents/PressResponder.js). If click retargets to the panel, React dispatches upward from the panel and the Pressable is never on the path.

Could you check with a bare <Pressable> in a sheet? Either way I'd move the capture into onPointerMove where drag.active flips to true. A tap never crosses 8px, so it never captures, and the selector list stops being load-bearing.

Worth noting no test can catch this today: jsdom 20 has no setPointerCapture, so the branch is skipped, and the Playwright fixtures all use elements with role="button".

### 2. touch-action: none is only on the handle, not the panel.

A body drag on touch is a candidate pan for the browser, which can claim it and fire pointercancel. Same root cause as (1): the gesture is claimed at press time but only defended on the handle.

The mobile-chrome project doesn't cover this either, since dragPanel goes through page.mouse, which always dispatches mouse events.

### 3. Drag-to-dismiss with no snapPoints.

minSnapHeight handles the snap-point case well, stepping down through detents. But with no snap points it falls back to CLOSE_HEIGHT_PX, so a 400px content sheet needs a 352px drag to dismiss, and a taller sheet needs more. There's no intermediate snap and onDragEnd is undefined, so a partial drag just springs back, and without an inline height the spring-back isn't animated. Could the fallback be a fraction of startHeight instead of a fixed 48?

packages/expo-ui/src/web/BottomSheetDialog.tsx
Kudo: Fixed. Drag-down now sets `translateY` and leaves height alone. id:8ced3c7daaf

Kudo:
Fixed.

1. `setPointerCapture` now runs in `pointermove` once the drag crosses 8px. A tap on a bare div never captures, so click still fires. Added a test that spies `setPointerCapture`.

2. `touch-action: none` is on the panel. The scrollable body keeps `pan-y`.

3. Without snap points, the sheet closes below half of its start height. A short drag springs back with the transform transition.

id:8ced3c7daaf

intergalacticspacehighway (approved):

PR #49511, opened 2026-08-28
fix(expo-constants): Use native build mode for app config

ramonclaudio:
# Why

Constants can generate app config using `NODE_ENV` and dotenv values left over from an earlier step.

# How

I updated Constants to default to the native build mode, clear Expo's previous dotenv values, and load the matching env files before reading app config.

# Test Plan

All tests pass.


ramonclaudio:
@expo-bot verify

packages/expo-constants/scripts/get-app-config-ios.sh
```diff
@@ -6,6 +6,18 @@ DEST="$CONFIGURATION_BUILD_DIR"
 RESOURCE_BUNDLE_NAME="EXConstants.bundle"
 EXPO_CONSTANTS_PACKAGE_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd -P)"
 
+is_debug_configuration() {
+  [[ "$1" =~ [Dd][Ee][Bb][Uu][Gg] ]]
```
Kudo: **[suggestion]** Config mode is chosen with a case-insensitive `debug` substring (`[[ "$1" =~ [Dd][Ee][Bb][Uu][Gg] ]]`). The JS bundle path uses a case-sensitive glob: `react-native-xcode.sh` sets `DEV` from `case "$CONFIGURATION" in *Debug*)`, and `expo-updates` on main uses `[[ "$CONFIGURATION" == *Debug* ]]`. `export:embed` then loads env files from that `--dev` flag. A configuration such as `debugStaging` or `DEBUG` therefore gets development `EXPO_PUBLIC_*` values in `Constants.expoConfig` and production values in the embedded bundle. The new iOS test encodes that split: it expects `debugStaging` to be development.

**Suggestion:** Use the same test as the bundler, `[[ "$1" == *Debug* ]]`, and rename the custom-debug fixture to `DebugStaging` so app config and the JS bundle stay on one mode.

packages/expo-constants/scripts/getAppConfig.js
```diff
@@ -22,9 +21,16 @@ if (fs.existsSync(path.join(possibleProjectRoot, 'package.json'))) {
   );
 }
 
-require('@expo/env').load(projectRoot);
+const expoEnv = require('@expo/env');
+process.env = expoEnv.getOriginalEnv();
+const mode = expoEnv.consumeConfigEnvMode();
+if (!mode) {
```
Kudo: **[suggestion]** A missing mode now throws `Must provide a config mode`. The three in-repo callers always set `__EXPO_CONFIG_MODE`, so Xcode/Gradle/rebundle are fine, and the changelog already marks this as breaking. The message does not name the variable, so a custom script phase or a forgotten export will fail a native build with no hint of what to set. `@expo/cli`'s `getConfigEnvMode()` also falls back (`EAS_BUILD ? 'production' : 'development'`) instead of throwing, with a comment that older EAS versions omit the variable.

**Suggestion:** Mention `__EXPO_CONFIG_MODE` (and the allowed values) in the error. If direct invocations should keep working, use the same fallback as `getConfigEnvMode()` rather than throwing.

packages/expo-constants/scripts/get-app-config-android.gradle
```diff
@@ -15,35 +15,34 @@ afterEvaluate {
...
-  def currentCreateConfigTask = tasks.register('createExpoConfig', Exec) {
-    description = 'expo-constants: Create app.config.'
-
-    doFirst {
-      assetsDir.deleteDir()
-      assetsDir.mkdirs()
+  project.android.libraryVariants.all { variant ->
+    def targetName = variant.name.capitalize()
+    def assetsDir = file("$buildDir/generated/assets/expo-constants/${variant.name}")
+    def configMode = System.getenv("__EXPO_CONFIG_MODE") ?: (variant.buildType.debuggable ? "development" : "production")
```
Kudo, line 22: i wonder android library would be build in "release" by default, even if the app is built in debug build. in this case, the config will always be `production`. that's something good to double check

Kudo (approved):
nit comments and good to double check if android is always running `release` build for libraries

ramonclaudio:
@expo-bot verify

ramonclaudio:
@expo-bot verify

ramonclaudio:
@expo-bot verify

ramonclaudio:
@expo-bot verify --fable

packages/@expo/cli/src/run/ios/runIosAsync.ts
```diff
@@ -93,21 +94,33 @@ export async function runIosAsync(projectRoot: string, options: Options) {
...
+        // But we can't resolve via `expo` because that then may do something differently than autolinking if the root has a different version
+        path.join(require.resolve('expo-constants/package.json'), '../scripts/getAppConfig.js'),
+        projectRoot,
+        path.join(options.binary, 'EXConstants.bundle'),
+      ],
+      {
+        env: {
+          ...getOriginalEnv(),
+          NODE_ENV: process.env.NODE_ENV,
+          __EXPO_CONFIG_MODE: process.env.__EXPO_CONFIG_MODE ?? configMode,
```
kitten: I would skip the default `configMode` here and assume that this is "isolated". That way we don't force a `production` override, which the script in the build is then unable to rederive.

If we don't do this, we risk this being overridden and becoming misaligned with what the build scripts do.

That's also consistent with the `getOriginalEnv()` call below, in that, we're restoring a "fresh" env set

PR #49524, opened 2026-08-29
[maps][ios] Add `cameraUpdateFrequency` to `AppleMaps.View`

moishinetzer:
# Why

`AppleMaps.View` only invokes `onCameraMove` once the camera has come to rest (`.onMapCameraChange(frequency: .onEnd)`), whereas `GoogleMaps.View` reports every camera change as it happens (`LaunchedEffect(cameraState.position)`). Anything that derives state from the viewport while the user is still gesturing, such as regrouping clustered markers during a pinch, cannot be done on iOS today: the clusters only regroup after the fingers lift, which looks broken next to the Android version of the same screen.

MapKit supports this directly through `MapCameraUpdateFrequency.continuous`. This PR exposes the choice as an opt-in prop instead of changing the default, so existing apps keep the current, cheaper `onEnd` behaviour.

Companion to #49525, which is the Android half of the same use case (fading clustered markers in and out as they regroup). The iOS counterpart of the marker animation is the draft #49526 (`animateAnnotations`, appear-only).

# How

- `AppleMapsViewProps` gets `cameraUpdateFrequency?: AppleMapsCameraUpdateFrequency` with `ON_END` (default, current behaviour) and `CONTINUOUS`. The enum is also exposed as `AppleMaps.CameraUpdateFrequency`, like the other Apple Maps enums.
- The Swift `AppleMapsViewProps` gets a matching `@Field var cameraUpdateFrequency: CameraUpdateFrequency = .onEnd`. The `CameraUpdateFrequency` enum in `MapRecords.swift` converts to MapKit's `MapCameraUpdateFrequency`, following the `MapColorScheme` / `MapStyleElevation` pattern. It deliberately has no `Map` prefix so it doesn't shadow MapKit's struct of that name inside the module.
- `AppleMapsViewiOS17` and `AppleMapsViewiOS18` pass the prop to `.onMapCameraChange(frequency:)`. The event payload is unchanged.
- Docs: TSDoc on the prop and the enum (the API section of the maps docs page is generated from them), plus a pointer from the `onCameraMove` doc.
- CHANGELOG entry.
- native-component-list: the Apple "Maps events" screen gets a switch that toggles the frequency and a counter of `onCameraMove` events, so the difference is visible without writing code.

There is no Android change: `GoogleMaps.View` already reports continuously, and `AppleMapsViewProps` is iOS-only, so there is nothing to mirror. The prop doc says so.

# Test Plan

- The native change itself, `.onMapCameraChange(frequency: .continuous)`, has been running in a production app as a `pnpm patch` of `expo-maps@57.0.2` on an iPhone running iOS 26 (the `AppleMapsViewiOS18` path): clusters regroup mid-pinch instead of after the gesture ends. That patch hard-codes `.continuous`. The prop plumbing in this PR (`@Field` + `Enumerable` conversion, and the `AppleMapsViewiOS17` path) follows the existing `colorScheme` pattern but has **not** been compiled or run by me: I don't have Xcode in the environment I'm contributing from.
- TypeScript: `tsc --noEmit` over `packages/expo-maps/src` and the touched NCL screen passes with a local TypeScript 6 install; the touched TS files are formatted with `oxfmt` 0.55 using the repo's `.oxfmtrc.json`. The monorepo's own `pnpm typecheck` / `pnpm lint` were not run (no monorepo install here).
- To reproduce: NCL → Expo Maps → "Maps events" on iOS. Flip the "Continuous camera updates" switch and pinch: the `onCameraMove` counter should only climb during the gesture while the switch is on.

# Checklist

### DIFF
--- a/apps/native-component-list/src/screens/ExpoMaps/apple/MapsEventsScreen.tsx
+++ b/apps/native-component-list/src/screens/ExpoMaps/apple/MapsEventsScreen.tsx
@@ -1,6 +1,6 @@
 import { AppleMaps, Coordinates } from 'expo-maps';
 import React from 'react';
-import { Text, View } from 'react-native';
+import { Switch, Text, View } from 'react-native';
 
 import ConsoleBox from '../../../components/ConsoleBox';
 
@@ -40,6 +40,8 @@ export default function MapsEventsScreen() {
   const [lastEvent, setLastEvent] = React.useState<string>('');
   const [visibleCount, setVisibleCount] = React.useState<number | null>(null);
   const [viewportSize, setViewportSize] = React.useState<string>('—');
+  const [continuousCameraUpdates, setContinuousCameraUpdates] = React.useState(false);
+  const [cameraMoveCount, setCameraMoveCount] = React.useState(0);
 
   return (
     <View style={{ flex: 1 }}>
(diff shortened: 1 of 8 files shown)


brentvatne:
@expo-bot verify

PR #49525, opened 2026-08-29
[maps][android] Add `animateMarkers` to `GoogleMaps.View`

moishinetzer:
# Why

When the `markers` of `GoogleMaps.View` change, Google Maps pops the removed markers out and the new ones in on the next frame. For a marker set that changes on every zoom step, such as clustered markers, this makes the map flicker while the user pinches. The Android Maps SDK has no appear animation on `MarkerOptions` (unlike `GMSMarker.appearAnimation` on the iOS SDK), but maps-compose exposes `alpha` on `Marker`, which is enough to fade markers in and out from the Compose side.

This PR adds an opt-in `animateMarkers` prop. It is off by default, and the default path runs the same marker loop as today.

Companion to #49524, which is the iOS half of the same use case (continuous `onCameraMove` so clusters can regroup during a pinch). The iOS counterpart of the marker animation is the draft #49526 (`animateAnnotations`, appear-only).

# How

- `GoogleMapsViewProps.animateMarkers?: boolean` in TypeScript and the matching `MutableState<Boolean>` on the Kotlin props.
- The `Marker(...)` call in `GoogleMapsView.Content()` moves into a private `MapMarker(marker, state, alpha)` composable; the only new argument is `alpha`. With `animateMarkers` off it is still emitted from the same `for` + `key(marker.id)` loop, with `alpha = 1f`.
- With `animateMarkers` on, markers go through the new `AnimatedMarkers` composable (**AnimatedMarkers.kt**):
  - It remembers the last list it composed and diffs it against the current one *during composition*, keeping removed markers in a `mutableStateMapOf` of departing markers. Doing this in composition rather than in a `LaunchedEffect` is the crux: an effect runs after the composition that dropped the marker, by which point maps-compose has already removed it from the map, so the "fade-out" would run on a freshly added, invisible ghost. `AnimatedContent` maintains its `currentlyVisible` list the same way.
  - Current and departing markers are emitted from one loop under `key(marker.id)`, so a marker moving from current to departing keeps its composable (and its `Marker` on the map) and only animates `alpha` to 0 through an `Animatable`. When the fade-out finishes, the entry is dropped from the departing map and maps-compose removes the marker. A marker that comes back while fading out animates back to 1; a marker removed while fading in animates from wherever it is.
  - `MarkerState` handling is unchanged: departing markers keep the `MarkerState` they were composed with, current markers keep getting states from `markerStateFromProps()` as before. So duplicate `id`s and draggable markers behave exactly as they do today (no `MarkerState` is shared between markers).
- No Gradle change: `androidx.compose.animation.core` is already on the compile classpath through `foundation-android` (`animation-android` is a `compile`-scope dependency of it), which is how `expo-ui` uses `Animatable` without declaring it either.
- Docs: TSDoc on the prop (the API section of the maps docs page is generated from it), CHANGELOG entry, and the NCL "Markers" screen gets an "Animate markers" switch plus a button that removes and re-adds two markers.

The fade duration is a fixed 200 ms. I left it hard-coded rather than adding a second prop; happy to expose it if you would rather.

# Test Plan

- Fade-in was verified on an Android 15 emulator in a production app, as a `pnpm patch` of `expo-maps@57.0.2` with a clustered marker set that changes on every zoom step: new markers fade in instead of popping. That patch diffed the lists in a `LaunchedEffect`, which is exactly the variant whose fade-out never renders (see "How"), so the code in this PR is a rewrite of that approach and has **not** been compiled or run on a device by me: I don't have the Android SDK in the environment I'm contributing from. It needs an NCL build to confirm, in particular the fade-out.
- TypeScript: `tsc --noEmit` over `packages/expo-maps/src` and the touched NCL screen passes with a local TypeScript 6 install; the touched TS files are formatted with `oxfmt` 0.55 using the repo's `.oxfmtrc.json`. The monorepo's own `pnpm typecheck` / `pnpm lint` were not run.
- To reproduce: NCL → Expo Maps → "Markers" on Android. With "Animate markers" on, "Remove #2 and #4" should fade those two markers out and "Add #2 and #4" fade them back in; with it off they should pop as before. Dragging marker #1 and selecting markers through the list (`selectMarker`) should work the same in both modes.

# Checklist

### DIFF
--- a/apps/native-component-list/src/screens/ExpoMaps/google/MapsMarkerScreen.tsx
+++ b/apps/native-component-list/src/screens/ExpoMaps/google/MapsMarkerScreen.tsx
@@ -1,6 +1,6 @@
 import { GoogleMaps } from 'expo-maps';
 import { useRef, useState } from 'react';
-import { View, Text, FlatList, Pressable, StyleSheet } from 'react-native';
+import { View, Text, FlatList, Pressable, StyleSheet, Switch } from 'react-native';
 
 const MARKERS: GoogleMaps.Marker[] = [
   {
@@ -39,6 +39,9 @@ const MARKERS: GoogleMaps.Marker[] = [
 export default function MapsMarkerScreen() {
   const mapRef = useRef<GoogleMaps.MapView>(null);
   const [selectedMarkerId, setSelectedMarkerId] = useState<string | undefined>();
+  const [animateMarkers, setAnimateMarkers] = useState(true);
+  const [showEveryMarker, setShowEveryMarker] = useState(true);
+  const markers = showEveryMarker ? MARKERS : MARKERS.filter((_, index) => index % 2 === 0);
 
   const handleMarkerSelect = async (markerId: string) => {
     setSelectedMarkerId(markerId);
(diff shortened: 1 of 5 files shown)


brentvatne:
@expo-bot verify

PR #49526, opened 2026-08-29
[maps][ios] Add `animateAnnotations` to `AppleMaps.View`

moishinetzer:
# Why

When the `annotations` of `AppleMaps.View` change, the new ones pop onto the map on the next frame. For an annotation set that changes on every zoom step, such as clustered badges that split as the user zooms in, this flickers. This PR adds an opt-in `animateAnnotations` prop that fades and scales an annotation's content in the first time it appears. It is off by default, and with it off the rendered content is the same as today.

It is the iOS counterpart of #49525 (`animateMarkers` on `GoogleMaps.View`), and pairs with #49524 (continuous `onCameraMove`), which is what lets clusters regroup mid-pinch in the first place.

**Appear-only, on purpose.** MapKit's `MapContent` `ForEach` offers no transition for removal (`.transition` is a `View` modifier and does not apply to `MapContent`), so a removed annotation still disappears instantly. Animating removal would mean keeping removed annotations in the `ForEach` on a timer, which I don't think belongs in the library; the Android PR can do it because maps-compose keeps the marker composable alive.

# How

- `AppleMapsViewProps.animateAnnotations?: boolean` in TypeScript and `@Field var animateAnnotations: Bool = false` on the Swift props.
- The annotation content (`ZStack` with the icon or rounded rectangle plus the text), which was duplicated in `AppleMapsViewiOS17` and `AppleMapsViewiOS18`, moves into one `AnnotationContent` view (**ios/AnnotationContent.swift**). Both views render it in place of the `ZStack`; the iOS 18 view keeps its `.tag(MapSelection(...))`.
- `AnnotationContent` animates with `opacity` / `scaleEffect` driven by an `@State isShown` that flips to `true` in `onAppear`, under `.animation(.easeInOut(duration: 0.22), value: isShown)`. That is the modifier chain the appearance was verified with (see Test Plan).
- To make it animate exactly once per annotation, the view state holds an `AnnotationAppearances` set of ids that have been shown:
  - `AnnotationContent` initialises `isShown` from `hasShown(id)`, and marks the id shown in `onAppear`. So an annotation that already appeared starts at full opacity when its view is created again, whether that happens because MapKit recreated the view after it was scrolled off screen or because SwiftUI gave it a new identity. Content edits for an unchanged `id` repaint through the normal `ForEach` update and don't animate.
  - `.onChange(of: props.annotations.map(\.id))` drops ids that left `annotations`, so an annotation that is removed and later added back (a cluster that splits and re-merges) animates in again. This is keyed on ids only; the content itself is rendered straight from `props.annotations` as before, so nothing about an annotation's text, colour or icon is compared or cached.
- The set is a plain `let` on `AppleMapsViewiOS17State` / `AppleMapsViewiOS18State`; it is bookkeeping, not rendered state, so it is not `@Published`.
- Docs: TSDoc on the prop (the API section of the maps docs page is generated from it), CHANGELOG entry, and the NCL "Annotations" screen gets an "Animate annotations" switch plus a button that removes and re-adds two annotations.

Things I had to reason about rather than observe, called out so a reviewer with Xcode can check them:

1. **Whether `onAppear` fires again (and `@State` resets) when MapKit recreates an annotation view after it was scrolled off screen and back.** I believe it does, which is why the appearances set exists: without it, badges would replay their fade every time they scroll back into view. If MapKit turns out to keep the SwiftUI view alive instead, the set is harmless.
2. **Enabling `animateAnnotations` at runtime** for annotations that are already on the map: they are not in the set yet, so if MapKit later recreates their views (case 1) they would animate once, late. If that is observable, the fix is to seed the set with the current ids in `.onChange(of: props.animateAnnotations)`; I left it out to keep the diff small.
3. Annotations that are in the initial `annotations` also fade in on first render, same as the Android PR. Easy to change to "only annotations added later" if you'd prefer that default.

# Test Plan

- The appearance animation (`opacity` + `scaleEffect` + `.animation(value:)`, flipped from `onAppear` inside `Annotation` content) is verified on an iPhone running iOS 26 (TestFlight build of the app, `pnpm patch` of `expo-maps@57.0.2`): when a cluster splits, the new badges fade and scale in. That patch tracked appearances differently (a synced copy of the annotations plus a departing list), which is what I replaced with the appearances set; the set, the id pruning, the prop plumbing and the `AppleMapsViewiOS17` path have **not** been compiled or run by me, as I don't have Xcode in the environment I'm contributing from. That's why this is a draft.
- TypeScript: `tsc --noEmit` over `packages/expo-maps/src` and the touched NCL screen passes with a local TypeScript 6 install; the touched TS files are formatted with `oxfmt` 0.55 using the repo's `.oxfmtrc.json`. The monorepo's own `pnpm typecheck` / `pnpm lint` were not run.
- To reproduce: NCL → Expo Maps → "Annotations" on iOS. With "Animate annotations" on, "Add Reading and Oxford" should fade and scale those two in and "Remove Reading and Oxford" should drop them instantly; toggling twice should animate them in again. Pan them off screen and back: they should not animate again (point 1 above). With the switch off nothing animates.

# Checklist

### DIFF
--- a/apps/native-component-list/src/screens/ExpoMaps/apple/MapsAnnotationsScreen.tsx
+++ b/apps/native-component-list/src/screens/ExpoMaps/apple/MapsAnnotationsScreen.tsx
@@ -1,6 +1,6 @@
 import { AppleMaps } from 'expo-maps';
 import { useRef, useState } from 'react';
-import { View, Text, FlatList, Pressable, StyleSheet } from 'react-native';
+import { View, Text, FlatList, Pressable, StyleSheet, Switch } from 'react-native';
 
 const ANNOTATIONS: AppleMaps.Annotation[] = [
   {
@@ -48,6 +48,11 @@ const ANNOTATIONS: AppleMaps.Annotation[] = [
 export default function MapsAnnotationsScreen() {
   const mapRef = useRef<AppleMaps.MapView>(null);
   const [selectedAnnotationId, setSelectedAnnotationId] = useState<string>();
+  const [animateAnnotations, setAnimateAnnotations] = useState(true);
+  const [showEveryAnnotation, setShowEveryAnnotation] = useState(true);
+  const annotations = showEveryAnnotation
+    ? ANNOTATIONS
+    : ANNOTATIONS.filter((_, index) => index % 2 === 0);
 
   const handleAnnotationSelect = (annotationId: string) => {
     mapRef.current?.selectAnnotation(annotationId);
@@ -78,7 +83,8 @@ export default function MapsAnnotationsScreen() {
           },
           zoom: 7,
         }}
-        annotations={ANNOTATIONS}
+        annotations={annotations}
+        animateAnnotations={animateAnnotations}
         onAnnotationClick={onAnnotationClick}
         onMapClick={onMapClick}
         properties={{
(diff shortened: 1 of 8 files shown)


brentvatne:
@expo-bot verify

PR #49529, opened 2026-08-30
fix(log-box): Only enable Android WebView debugging for debuggable apps

kosumic:
# Why

`ExpoLogBoxWebViewWrapper` unconditionally calls `WebView.setWebContentsDebuggingEnabled(true)`.

This setting applies to every WebView in the application. [Android documents](https://developer.android.com/reference/android/webkit/WebView#setWebContentsDebuggingEnabled(boolean)) that enabling it allows WebView state to be inspected and modified through ADB and should not normally be enabled in production builds.

We currently carry this guard as a [downstream patch in Whip](https://github.com/KaminariOS/whip/blob/main/patches/%40expo%2Blog-box%2B57.0.4.patch).

# How

Check the consuming application’s `ApplicationInfo.FLAG_DEBUGGABLE` flag before explicitly enabling WebView debugging.

This preserves the existing behavior for debuggable applications, including those using WebView versions older than 113, while non-debuggable applications retain WebView’s secure default. The release path does not explicitly disable debugging, so it does not override a host application that intentionally manages the process-wide setting itself.

# Test Plan

- [x] `git diff --check`
- [x] The equivalent downstream patch compiles in [Whip Android CI](https://github.com/KaminariOS/whip/actions/runs/33300430436/job/99227417959).

# Checklist

### DIFF
--- a/packages/@expo/log-box/CHANGELOG.md
+++ b/packages/@expo/log-box/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### 🐛 Bug fixes
 
+- [Android] Only enable LogBox WebView debugging in debuggable apps. ([#49529](https://github.com/expo/expo/pull/49529) by [@KaminariOS](https://github.com/KaminariOS))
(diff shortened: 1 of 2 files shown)


brentvatne:
@expo-bot verify

PR #49533, opened 2026-08-31
[android][image] Fix native crashes with some animated GIFs

MangelSpec:
# Why

I started investigating this after Mattermost Mobile repeatedly crashed on specific channels, making those channels effectively unusable for me:

https://github.com/mattermost/mattermost-mobile/issues/9726

During the investigation, one of the animated GIFs involved reproduced a crash in APNG4Android's native GIF decoder. A minimized synthetic reproducer crashed APNG4Android 3.0.5 on 4 of 5 cold loads, while a known-good control GIF crashed on 0 of 5 loads. The decoder bug and root cause are documented upstream:

https://github.com/penfeizhou/APNG4Android/issues/258

I have been testing the complete downstream Mattermost mitigation, which routes GIFs through Glide and safely clears released image resources, for nearly three weeks without another observed crash:

https://github.com/mattermost/mattermost-mobile/pull/10030

Other people have reported similar crashes involving emoji, including the following report, although I have not confirmed that it has the same root cause:

https://github.com/mattermost/mattermost-mobile/issues/8252

# How

- Prepend Glide's built-in `InputStream` and `ByteBuffer` GIF decoders in Expo Image's `AppGlideModule`, so GIFs remain animated without passing through APNG4Android's affected decoder.
- Recognize both APNG4Android and Glide `GifDrawable` instances in the `isAnimated` property.
- Notify `ExpoImageViewWrapper` when Glide clears a target. A view is recycled only when it still references that target, and its animation is cancelled after the current target is cleared so a stale fade callback cannot clear a newly rebound resource.
- Replace the remote animated test source with a small local GIF and add an Android test-suite case that repeatedly alternates animated and static sources while reusing the image view.

APNG4Android remains registered for APNG, animated WebP, and AVIF; the added Glide decoders only handle GIF input.

Projects that intentionally set `expoImage.excludeAppGlideModule=true` continue to own their Glide component registration; Expo's decoder preference is not installed in that configuration.

This specifically avoids APNG4Android's small-interlaced-frame overflow. It does not address Glide's separate oversized logical-screen-dimension issue tracked in https://github.com/bumptech/glide/pull/5693.

# Test Plan

Reproduction and Android verification:

- APNG4Android 3.0.5 official demo, Android API 36 x86_64:
  - Self-created synthetic GIF: native crash on 4/5 cold loads.
  - Known-good control GIF: native crash on 0/5 cold loads.
- Temporary decoder instrumentation on Android API 36 x86_64 confirmed that the generated Expo Image `AppGlideModule` decoded the 8x8 synthetic GIF as exactly `com.bumptech.glide.load.resource.gif.GifDrawable`. The drawable implemented `Animatable`, started and stopped successfully, and did not crash.
- `:expo-image:compileDebugKotlin` passed.

Package and test checks:

- `et check-packages expo-image`: passed all 53/53 tasks.
- Expo Image Jest: passed 9 suites and 193 tests.
- Expo Image build, typecheck, lint, and formatting: passed.
- Test-suite formatting and lint: passed.
- Test-suite Jest: passed 1 suite and 9 tests.
- `git diff --check`: passed.

Known local verification limitations:

- The test-suite typecheck reports 55 pre-existing errors, all in unchanged upstream files (`components/Suites.tsx`, `ExpoUIMeasurement.android.tsx`, and `ExpoUIMeasurement.ios.tsx`). It reports no error in the changed `Image.tsx`.
- The new `Image.tsx` recycling case was format, lint, and Jest checked, but was not run inside the Android test app because the full bare-expo build did not complete.
- A separate lifecycle instrumentation attempt was inconclusive because the Expo dev launcher destroyed its test `MainActivity` before the assertions ran. It is not counted as passing device validation.
- A full bare-expo APK build was not completed locally because of unrelated Windows native dependency and path failures. Expo Image's focused Kotlin compilation passed.

# Checklist


brentvatne:
@expo-bot verify

brentvatne:
@expo-bot verify

MangelSpec:
I can confirm the analysis bot’s concern about rapid source changes.

I reproduced it locally with Glide 5.0.5. When both targets are still in use during a transition, the current selection logic can reuse the foreground target. Glide then calls `onLoadCleared`, and the cleanup correctly detaches that resource, but this makes the currently displayed image disappear until the replacement is ready.

The cleanup itself is still necessary, since leaving a cleared Glide resource attached is what caused the separate recycled-drawable crash.

One possible improvement would be to explicitly track which target owns the foreground image:

```kotlin
private var foregroundTarget: ImageViewWrapperTarget? = null
```

Update it when a resource becomes the displayed image:

```kotlin
private fun configureView(
  view: ExpoImageView,
  target: ImageViewWrapperTarget,
  resource: Drawable,
  isPlaceholder: Boolean
) {
  // Existing view configuration...

  target.isUsed = true
  foregroundTarget = target
}
```

Keep the existing cleanup, while clearing the foreground reference when that target is actually released:

```kotlin
fun onTargetCleared(target: ImageViewWrapperTarget) {
  if (foregroundTarget === target) {
    foregroundTarget = null
  }

  // Existing ownership-checked view cleanup...
}
```

Then select the other target for the next request:

```kotlin
val newTarget = when (foregroundTarget) {
  firstTarget -> secondTarget
  secondTarget -> firstTarget
  else -> if (secondTarget.isUsed) firstTarget else secondTarget
}
```

This keeps the foreground target and its Glide resource alive while the replacement loads. If both targets are in use, Glide instead clears the target belonging to the older fading image.

I tested this in an A/B Android instrumentation test. The current selector cleared the foreground target, while the adjusted selector cleared only the older fading target and kept the foreground image attached until the replacement was ready. The Expo Image Kotlin compilation also passes with this adjustment.

I haven’t pushed this change yet. I’d appreciate feedback on whether this is the right approach and whether it should be included in this PR.

PR #49551, opened 2026-08-31
[ios][precompile] Remove the pre-0.87 clang VFS overlay path

chrfalch:
# Why:

In React Native 0.87 we removed the VFS overlay file and delivered header files as self contained files in the ReactNativeHeaders.xcframework and ReactNativeDependenciesHeaders.xcframework. There are no more `-ivfsoverlay` settings emitted from React Native, and there should be zero need for any in Expo - the xcframeworks should serve headers all by themselves. 

A switched gate was added in #47256 in June 2026 and has been on `main` for the past two weeks (after we merged RN 0.87) without failing.

# How:

This PR contains the removal if the path in the existing precompile and pod install ruby code that ran when the VFS Overlay was found.

- Removed `ReactCFSOverlay`, `TransforReactXCFranework` and `ReactHeaderMappings` typescript classes since they're only used with the VFS overlay.
- Removed the code emitting `-ivfsoverlay` in both precompile and ruby code paths
- Updated unit tests.

# Test Plan:

- ✅ `tools`: 406 tests, 403 pass, 0 fail (was 410/407/0; six tests of deleted code removed, two added). pnpm build, tsc --noEmit and lint --max-warnings 0 clean.
- ✅ `et check-packages expo-modules-autolinking` passes.
- ✅ apps/bare-expo, clean Pods/ and build/: pod install gives 227 dependencies / 208 pods, 480 xcconfigs carrying the module map flag and header search path, zero -ivfsoverlay, zero React-VFS.yaml, Podfile.lock unchanged.
- ✅ Cold Debug build succeeds with no errors; the app launches on the iOS simulator, loads from Metro and renders native SwiftUI Charts.
- ✅ Release build succeeds and swaps to the release xcframeworks; switching back to Debug restores the debug ones byte-for-byte.

# Checklist


chrfalch:
@expo-bot review

tools/src/prebuilds/SPMPackage.ts
```diff
@@ -1297,82 +1239,48 @@ function collectVfsAndHeaderMapFlags(
...
+      // The `continue` keeps React out of the include-directory branch below: an include root into
+      // React.xcframework is redundant once the module map is active — verified by prebuilding
+      // ExpoModulesCore (C++ target, consumes the react/ and yoga/ namespaces) and ExpoCrypto
+      // against 0.87.0-rc.3 artifacts, where both flavors compose and verify without one.
+      const moduleMapXcframework = artifactConfig?.moduleMapXcframework;
+      if (lowerName === 'react' && moduleMapXcframework) {
+        const pushModularFlags = (flags: string[], basePath?: string) => {
+          // A flavor is only downloaded when it is built: `et prebuild --flavor Debug` leaves the
+          // release slot absent, and that flavor's flags are inert in the run that emits them.
+          if (!basePath || !fs.existsSync(basePath)) {
```
chrfalch: Fixed in 07565d41278. `pushModularFlags` now takes the flavor it is emitting for and only throws when that is `buildType`; the other flavor is skipped. Dropped the `fs.existsSync` gate — an empty directory never distinguished mid-download from incomplete. New test covers a complete built flavor alongside an empty other-flavor slot; it fails against 9585e649153 with the incomplete-artifact error naming the release path.

id:24f1f7f3376a

tools/src/prebuilds/ReactHeaderMappings.ts
```diff
@@ -1,236 +0,0 @@
...
- *
- * Ported from react-native's `scripts/ios-prebuild/headers.js`.
- * Discovers all podspecs in the RN source tree, parses their header patterns,
- * and returns structured header mappings (source → target) for each pod.
- */
-
-import fs from 'fs-extra';
-import { glob } from 'glob';
-import path from 'path';
-
```
tsapeta, line 12: [💡 suggestion] Deleting this file orphans `tools/src/prebuilds/ReactHeaderConfig.ts`: this import was its only consumer, and nothing references it at the head commit. Consider deleting that file (437 lines of ported RN podspec exception config) in this PR too, since `tsc --noEmit` won't flag an unused module.

tsapeta (approved):

gabrieldonadel (approved):

Issue #48992, opened 2026-08-16
[expo-notifications] Data race on NotificationCenterManager.delegates causes SIGSEGV during module registration (iOS)


bradzickafoose:
### Summary

`NotificationCenterManager` is a process-wide singleton (`static let shared`) whose mutable `delegates` array is not synchronized. Three modules append to it during module registration, while the `UNUserNotificationCenter` callbacks read and mutate it from the main thread. The resulting data race corrupts the array buffer's refcounts and segfaults inside `swift_release_dealloc`.

This reproduces on iOS in a development build and is intermittent — the app sometimes runs for minutes, sometimes dies ~14s after a cold launch.

### Crash

```
EXC_BAD_ACCESS (SIGSEGV)
KERN_INVALID_ADDRESS at 0x797040e760f07e5e -> 0x000040e760f07e5e (possible pointer authentication failure)
Thread: com.facebook.react.runtime.JavaScript

  libswiftCore.dylib  swift_release_dealloc
  libswiftCore.dylib  swift::RefCounts<...>::doDecrementSlow<...>
  libswiftCore.dylib  _ArrayBuffer._consumeAndCreateNew(bufferIsUnique:minimumCapacity:growForAppend:)
  libswiftCore.dylib  Array._reserveCapacityAssumingUniqueBuffer(oldCount:)
  libswiftCore.dylib  Array.append(_:)
  ExpoNotifications   NotificationCenterManager.addDelegate(_:)
  ExpoNotifications   closure #1 in PushTokenModule.definition()
  ExpoModulesCore     closure #1 in EventListener.init(_:_:)
  ExpoModulesCore     ModuleHolder.post(event:)
  ExpoModulesCore     ModuleHolder.init(appContext:module:name:)
  ExpoModulesCore     ModuleRegistry.register(module:name:preventModuleOverriding:)
  ExpoModulesCore     ModuleRegistry.register(fromProvider:)
  ExpoModulesCore     AppContext.useModulesProvider(_:)
  ExpoModulesCore     AppContext.registerNativeModules(provider:)
```

### Root cause

`ios/ExpoNotifications/Notifications/NotificationCenterManager.swift`:

```swift
@objc
public static let shared = NotificationCenterManager()

var delegates: [NotificationDelegate] = []          // no lock / queue / actor
var pendingResponses: [UNNotificationResponse] = []  // same
```

```swift
public func addDelegate(_ delegate: NotificationDelegate) {
  delegates.append(delegate)   // <-- crash site
[...]

Issue #48202, opened 2026-07-27
[expo-notifications][iOS] Unsynchronized `delegates` array in NotificationCenterManager singleton — SIGSEGV in addDelegate when module registration races teardown (SDK 55)


team-charli:
### Summary

`expo-notifications`' `NotificationCenterManager` is a process-wide singleton whose `delegates` array is a plain Swift `var` mutated with **no synchronization**:

```swift
// ios/ExpoNotifications/Notifications/NotificationCenterManager.swift (55.0.20)
public static let shared = NotificationCenterManager()

var delegates: [NotificationDelegate] = []          // no lock / no actor / no queue confinement

public func addDelegate(_ delegate: NotificationDelegate) {
  delegates.append(delegate)                        // called from module OnCreate, off-main
  ...
}
```

`addDelegate` runs during module registration (`ModuleRegistry.register(fromProvider:)` → `ModuleHolder.post(OnCreate)` → `PushTokenModule` event listener), which executes on the React runtime thread. `removeDelegate` mutates the same array from whatever thread hosts module teardown. When registration and teardown overlap, the two threads race on the array's buffer, and `Array.append`'s copy-on-write growth path dereferences a corrupted buffer pointer → `EXC_BAD_ACCESS` (pointer-auth-invalid address).

We hit this reproducibly-enough (2 crashes in ~10 launches) in a scenario where two Expo runtimes briefly coexist in one process: an `expo-dev-client` app launched via `simctl launch` and then immediately handed a dev-client deep link via `simctl openurl` — the dev-launcher home runtime tears down while the app-bundle runtime registers its modules. That launch pattern was ours to fix (a single cold `openurl` avoids the swap entirely, and did), but the underlying hazard is not dev-client-specific: **any module registration running concurrently with a teardown can corrupt the singleton's array**, since nothing in the class serializes access.

### Crash evidence

`Charli-2026-07-26-175718.ips`, iOS Simulator (Xcode 26.5 / 17F42), SDK 55, expo-notifications **55.0.20**:

```
Exception: EXC_BAD_ACCESS SIGSEGV KERN_INVALID_ADDRESS at 0x2c8284092204386e -> 0xffff84092204386e

Thread 15 (com.facebook.react.runtime.JavaScript) — faulting:
  libswiftCore  _swift_release_dealloc
  libswiftCore  RefCounts::doDecrementSlow
[...]

PR #49554, opened 2026-08-31
[ios][notifications] Guard NotificationCenterManager's delegate list against concurrent access

### LINKED ISSUE #49549: [expo-notifications][iOS] Data race on NotificationCenterManager's unsynchronized delegates array → SIGSEGV in addDelegate (with self-contained repro)
### Summary

`NotificationCenterManager` is a process-wide singleton whose `delegates` and `pendingResponses` arrays are unsynchronized stored properties, mutated from whichever thread happens to be creating or destroying an app context. When one app context registers its modules while another tears its modules down, `addDelegate` and `removeDelegate` mutate the same array concurrently, corrupting the heap and killing the process with `SIGSEGV`.

This has been reported three times before — #47702, #48202, #48992 — and all three were auto-closed for lacking a minimal reproducible example, so the defect is still present on both `sdk-57` and `main`. **This issue includes a self-contained repro that needs no Expo project, no simulator and no Xcode project — one file, two commands.**

### The defect

[`NotificationCenterManager.swift`](https://github.com/expo/expo/blob/sdk-57/packages/expo-notifications/ios/ExpoNotifications/Notifications/NotificationCenterManager.swift), `sdk-57` L45–46 (`main` L53–54):

```swift
public class NotificationCenterManager: NSObject, ... {
  @objc
  public static let shared = NotificationCenterManager()

  var delegates: [NotificationDelegate] = []          // no lock, no queue, no actor
  var pendingResponses: [UNNotificationResponse] = [] // same
```

There are four unsynchronized mutation sites: `addDelegate` (append, and `pendingResponses.removeAll`), `removeDelegate` (remove at index), and `userNotificationCenter(_:didReceive:withCompletionHandler:)` (`pendingResponses.append`). Every `for delegate in delegates` callback reads the array with no synchronization either.

Three modules add and remove themselves against that singleton from module lifecycle hooks — `PushTokenModule`, `HandlerModule` and `EmitterModule`, each with `OnCreate { NotificationCenterManager.shared.addDelegate(self) }` and `OnDestroy { NotificationCenterManager.shared.removeDelegate(self) }`. Because the manager is a singleton but app contexts are not, any overlap between an outgoing context's teardown and an incoming context's registration is a data race. That overlap is routine: it happens on every `expo-dev-client` reload, and on `Updates.reloadAsync()` in a release build. Related: #48653.

Save as `NCMRace.swift`. The storage declarations and the bodies of `addDelegate`, `removeDelegate` and the delegate-iterating callback are copied verbatim from the file linked above; only the `UserNotifications` / `ExpoModulesCore` types are stubbed so it builds on a plain Mac.

```swift
import Foundation

final class UNNotificationResponse {}

protocol NotificationDelegate: AnyObject {
  func didReceive(_ response: UNNotificationResponse, completionHandler: @escaping () -> Void) -> Bool
  func didRegister(_ deviceToken: String)
}

extension NotificationDelegate {
  func didReceive(_ response: UNNotificationResponse, completionHandler: @escaping () -> Void) -> Bool { false }
  func didRegister(_ deviceToken: String) {}
}

final class NotificationCenterManager {
  static let shared = NotificationCenterManager()

  // >>> verbatim upstream: unsynchronized storage on a process-wide singleton
  var delegates: [NotificationDelegate] = []
  var pendingResponses: [UNNotificationResponse] = []

  // verbatim upstream body
  func addDelegate(_ delegate: NotificationDelegate) {
    delegates.append(delegate)
    var handled = false
    for pendingResponse in pendingResponses {
      handled = delegate.didReceive(pendingResponse, completionHandler: {}) || handled
    }
    if handled {
      pendingResponses.removeAll()
    }
  }

  // verbatim upstream body
  func removeDelegate(_ delegate: AnyObject) {
    if let index = delegates.firstIndex(where: { $0 === delegate }) {
      delegates.remove(at: index)
    }
  }

  // verbatim upstream body (called from the app delegate, i.e. the main thread)
  func didRegister(_ deviceToken: String) {
    for delegate in delegates {
      delegate.didRegister(deviceToken)
    }
  }
}

final class PushTokenModule: NotificationDelegate {
  func onCreate() { NotificationCenterManager.shared.addDelegate(self) }   // OnCreate
  func onDestroy() { NotificationCenterManager.shared.removeDelegate(self) } // OnDestroy
}

// Models a dev-client reload / Updates.reloadAsync(): the incoming app context
// registers modules on the new JS thread while the outgoing one tears its
// modules down, and a push-token callback lands on the main thread meanwhile.
let iterations = 5_000
let group = DispatchGroup()
var live: [PushTokenModule] = []
let liveLock = NSLock()

DispatchQueue.global().async(group: group) {          // incoming context
  for _ in 0..<iterations {
    let module = PushTokenModule()
    liveLock.lock(); live.append(module); liveLock.unlock()
    module.onCreate()
  }
}
DispatchQueue.global().async(group: group) {          // outgoing context
  for _ in 0..<iterations {
    liveLock.lock(); let module = live.popLast(); liveLock.unlock()
    module?.onDestroy()
  }
}
DispatchQueue.global().async(group: group) {          // app delegate callback
  for _ in 0..<iterations {
    NotificationCenterManager.shared.didRegister("token")
  }
}

group.wait()
print("completed without crashing (this race is intermittent — run again)")
```

Run it:

```sh
# 1. ThreadSanitizer names the racing function
swiftc -sanitize=thread -o ncmrace NCMRace.swift && ./ncmrace

# 2. Without the sanitizer it segfaults outright
swiftc -O -o ncmrace_raw NCMRace.swift
for i in $(seq 1 12); do ./ncmrace_raw >/dev/null 2>&1 || echo "crash (exit $?)"; done
```

### Results

TSan output, unmodified upstream logic:

```
WARNING: ThreadSanitizer: Swift access race (pid=26053)
  Modifying access of Swift variable at 0x00010d704d10 by thread T1:
    #0 NotificationCenterManager.addDelegate(_:)
    #1 PushTokenModule.onCreate()
  Previous read of size 8 at 0x00010d704d10 by thread T3:
    #0 NotificationCenterManager.didRegister(_:)
SUMMARY: ThreadSanitizer: Swift access race in NotificationCenterManager.addDelegate(_:)+0x10c
```

| build | ThreadSanitizer | 12 unsanitized runs |
|---|---|---|
| upstream logic as-is | data race reported in `addDelegate` | **9 / 12 died with SIGSEGV (exit 139)** |
| with the patch below | clean, 0 warnings | 0 / 12 |

### Matching real-world crash

Same signature from an actual iOS build (`expo-notifications` 57.0.14, `expo` 57.0.16, RN 0.86, New Architecture, iOS simulator), reproduced repeatedly by Maestro E2E relaunch cycles that deep-link from the dev-client launcher into the Metro bundle — i.e. exactly the two-overlapping-app-contexts case:

```
Thread 12 Crashed:  com.facebook.react.runtime.JavaScript
EXC_BAD_ACCESS (SIGSEGV), KERN_INVALID_ADDRESS
0  libswiftCore.dylib   swift_release_dealloc
1  libswiftCore.dylib   swift::RefCounts<...>::doDecrementSlow<...>
2  libswiftCore.dylib   _ArrayBuffer._consumeAndCreateNew(bufferIsUnique:minimumCapacity:growForAppend:)
3  libswiftCore.dylib   Array._reserveCapacityAssumingUniqueBuffer(oldCount:)
4  libswiftCore.dylib   Array.append(_:)
5  MyApp                NotificationCenterManager.addDelegate(_:)
6  MyApp                closure #1 in PushTokenModule.definition()
7  MyApp                closure #1 in EventListener.init(_:_:)
8  MyApp                closure #1 in ModuleHolder.post(event:)
9  MyApp                ModuleHolder.post(event:)
10 MyApp                ModuleHolder.init(appContext:module:name:)
```

The fault lands 10.8 s after process launch, inside module registration — not at cold start — which is the launcher-to-Metro handoff, not a fresh boot.

### Proposed fix

Guard both arrays with a lock, and have readers take a snapshot so delegate callbacks never run while the lock is held (calling out to a delegate under the lock would risk re-entrancy, since `addDelegate` invokes `didReceive` on the delegate it just added):

```diff
-  var delegates: [NotificationDelegate] = []
-  var pendingResponses: [UNNotificationResponse] = []
+  private let lock = NSLock()
+  private var _delegates: [NotificationDelegate] = []
+  private var _pendingResponses: [UNNotificationResponse] = []
+  var delegates: [NotificationDelegate] {
+    lock.lock()
+    defer { lock.unlock() }
+    return _delegates
+  }
+  var pendingResponses: [UNNotificationResponse] {
+    lock.lock()
+    defer { lock.unlock() }
+    return _pendingResponses
+  }

   public func addDelegate(_ delegate: NotificationDelegate) {
-    delegates.append(delegate)
+    lock.lock()
+    _delegates.append(delegate)
+    let pending = _pendingResponses
+    lock.unlock()
     var handled = false
-    for pendingResponse in pendingResponses {
+    for pendingResponse in pending {
       handled = delegate.didReceive(pendingResponse, completionHandler: {}) || handled
     }
     if handled {
-      pendingResponses.removeAll()
+      lock.lock()
+      _pendingResponses.removeAll()
+      lock.unlock()
     }
   }

   public func removeDelegate(_ delegate: AnyObject) {
-    if let index = delegates.firstIndex(where: { $0 === delegate }) {
-      delegates.remove(at: index)
+    lock.lock()
+    defer { lock.unlock() }
+    if let index = _delegates.firstIndex(where: { $0 === delegate }) {
+      _delegates.remove(at: index)
     }
   }
```

plus the one remaining mutation site in `userNotificationCenter(_:didReceive:withCompletionHandler:)`:

```diff
     if !handled {
-      pendingResponses.append(response)
+      lock.lock()
+      _pendingResponses.append(response)
+      lock.unlock()
     }
```

`NSLock` rather than `withLock` keeps this free of an iOS 16 availability requirement. Every existing `for delegate in delegates` call site keeps working unchanged, now iterating a snapshot. Happy to open a PR if that is useful.

### Environment

- `expo-notifications` 57.0.14 (verified unchanged through 57.0.15, the current latest in the SDK 57 line, and on `main`)
- `expo` 57.0.16, `expo-dev-client` 57.0.15, `expo-updates` 57.0.18
- React Native 0.86, New Architecture / Hermes, bridgeless
- iOS simulator, macOS 26.4 (25E246), Apple silicon
- Repro above was built and run with the Xcode-bundled Swift toolchain on macOS 26.4

dennytosp:
# Why

Fixes #49549. Previously reported as #47702, #48202 and #48992, each auto-closed for lacking a reproduction.

`NotificationCenterManager` is a process-wide singleton whose `delegates` and `pendingResponses` were unsynchronized stored properties. They are mutated by `addDelegate`, `removeDelegate` and the `pendingResponses.append` in `userNotificationCenter(_:didReceive:withCompletionHandler:)`, and read by every `for delegate in delegates` callback — from whichever thread happens to be creating or destroying an app context.

App contexts are not singletons. `PushTokenModule`, `HandlerModule` and `EmitterModule` each call `addDelegate` from `OnCreate` and `removeDelegate` from `OnDestroy`, so an incoming context registering its modules overlaps a departing context removing its own. That overlap is routine: every `expo-dev-client` reload, and `Updates.reloadAsync()` in a release build. Concurrent `Array.append` then corrupts the buffer and the process dies with `SIGSEGV` inside module registration.

# How

Both arrays move into a `NotificationDelegateRegistry` that keeps them behind `Mutex`, the backport in `ExpoModulesCore` that `expo-app-metrics` and `expo-app-intents` already use (`Synchronization.Mutex` is iOS 18+, this package targets 16.4).

Readers take a snapshot instead of holding the lock across a delegate callback. That matters: `addDelegate` calls `didReceive` on the delegate it just added, and a delegate is free to add or remove delegates from there — holding the lock across that would deadlock. `add` therefore returns the pending responses rather than exposing them separately.

Extracting the registry is what makes the synchronization testable. The manager is a singleton whose `init` calls `install()`, which reaches `UNUserNotificationCenter.current()` and crashes in a unit test bundle — I hit exactly that (`Crash: xctest at NotificationCenterManager.install()`) with a first version of these tests written against `.shared`. The registry has no such dependency.

`chainedDelegate` is left alone. It is written only from `install()` and is a separate concern from the reported crash.

# Test Plan

## The reporter's standalone repro

Their harness needs no Expo project, simulator or Xcode project. I re-ran it with the storage and the bodies of `addDelegate`, `removeDelegate` and `didRegister` copied verbatim from `main` — including `removeDelegate`'s current `removeAll { $0 === delegate }`, not the older `firstIndex` form quoted in the issue.

| build | ThreadSanitizer | 12 unsanitized runs |
| --- | --- | --- |
| `main` | race reported in `addDelegate` | **10–12 of 12 died with SIGSEGV** |
| this PR | **0 warnings** | **0 of 12** |

TSan on `main`'s logic:

```
WARNING: ThreadSanitizer: Swift access race
    #0 NotificationCenterManager.addDelegate(_:)
    #1 PushTokenModule.onCreate()
  Previous read:
    #0 NotificationCenterManager.didRegister(_:)
SUMMARY: ThreadSanitizer: Swift access race in NotificationCenterManager.addDelegate(_:)+0x10c
```

## Unit tests

Five new tests in `ios/Tests/NotificationDelegateRegistryTests.swift`. `et native-unit-tests -p ios --packages expo-notifications`:

```
NotificationDelegateRegistryTests
  ✓ keeps every delegate added from several threads at once
  ✓ removes every delegate removed from several threads at once
  ✓ keeps the survivors when adds and removes interleave across threads
  ✓ add returns the responses the new delegate still has to be offered
  ✓ removing a delegate that was never added leaves the others in place

BackgroundEventTransformerTests (existing, 5 tests)  ✓
** TEST SUCCEEDED **
```

Mutation check — reverting the registry to plain unsynchronized arrays and keeping everything else:

```
✕ keeps every delegate added from several threads at once      (0.000s)
✕ removes every delegate removed from several threads at once  (0.000s)
✕ keeps the survivors when adds and removes interleave across threads (0.000s)
✓ add returns the responses the new delegate still has to be offered
✓ removing a delegate that was never added leaves the others in place
** TEST FAILED **

Crash: xctest at closure #2 in NotificationDelegateRegistryTests.
  `removes every delegate removed from several threads at once`()
```

Exactly the three concurrency tests fail, and the two single-threaded ones still pass, so the tests are measuring the synchronization rather than passing by construction.

## One thing I had to add to run any of this

`test_spec` gains `'OTHER_LDFLAGS' => '$(inherited) -lc++'`. Without it the test target fails to link:

```
Undefined symbol: std::runtime_error::what() const
Undefined symbol: std::logic_error::logic_error(char const*)
  ... in libExpoModulesCore.a[3](EventEmitter-....o)
ExpoNotifications-Unit-Tests: clang: error: linker command failed with exit code 1
```

This is not caused by anything here — it is the same flag the test specs of `expo-maps`, `expo-camera`, `expo-audio`, `expo-updates`, `expo-image-picker`, `expo-observe`, `expo-app-metrics` and `expo-dev-launcher` already carry. Without it no test can run for this package at all, so the existing `BackgroundEventTransformerTests` could not have been running locally either. Happy to split it out if you would rather have it separately.

`Podfile.lock` carries the resulting `ExpoNotifications` checksum change, the same way the RN 0.87 upgrade PRs updated it.

## Not covered

I did not reproduce the crash in a running app on a device or simulator — the reporter's own trace does that, from a real build under Maestro relaunch cycles. Everything above is TSan output, crash counts and test results from this machine.

`et check-packages expo-notifications` reports `expo-notifications#lint` failing, but that reproduces unchanged on `main` (`Rule 'globals' not found in plugin 'react'`, an oxlint config/version mismatch). This PR touches no JavaScript or TypeScript.

# Checklist

- [x] Added unit tests that fail without the fix.

### DIFF
--- a/packages/expo-notifications/CHANGELOG.md
+++ b/packages/expo-notifications/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 ### 🐛 Bug fixes
 
+- [iOS] Fix a data race on `NotificationCenterManager`'s delegate list that crashed the app with `SIGSEGV` when one app context registered its modules while another tore its own down, such as on a dev-client reload or `Updates.reloadAsync()`. ([#49549](https://github.com/expo/expo/issues/49549), [#49554](https://github.com/expo/expo/pull/49554) by [@dennytosp](https://github.com/dennytosp))
 - [Android] Prevented `onUserLeaveHint` from firing when a notification tap opens the app, which made picture-in-picture implementations enter PiP unexpectedly. ([#48471](https://github.com/expo/expo/pull/48471) by [@stareezy-1](https://github.com/stareezy-1))
 - [iOS] Avoid warning when an aborted push token registration request rejects with a native fetch cancellation error. ([#48547](https://github.com/expo/expo/pull/48547) by [@JoaoPauloCMarra](https://github.com/JoaoPauloCMarra))
 - [Android] Prevent a crash on notification tap when `getLaunchIntentForPackage` throws on some OEM ROMs. ([#47889](https://github.com/expo/expo/pull/47889) by [@nunocaseiro](https://github.com/nunocaseiro))
--- a/packages/expo-notifications/CHANGELOG.md
+++ b/packages/expo-notifications/CHANGELOG.md
@@ -14,7 +14,7 @@
 
 ### 🐛 Bug fixes
 
-- [iOS] Fix a data race on `NotificationCenterManager`'s delegate list that crashed the app with `SIGSEGV` when one app context registered its modules while another tore its own down, such as on a dev-client reload or `Updates.reloadAsync()`. ([#49549](https://github.com/expo/expo/issues/49549), [#49554](https://github.com/expo/expo/pull/49554) by [@dennytosp](https://github.com/dennytosp))
+- [iOS] Fix a data race on `NotificationCenterManager`'s delegate list that crashed the app with `SIGSEGV` when one app context registered its modules while another tore its own down, such as on a dev-client reload or `Updates.reloadAsync()`. [#49554](https://github.com/expo/expo/pull/49554) by [@dennytosp](https://github.com/dennytosp))
 - [Android] Prevented `onUserLeaveHint` from firing when a notification tap opens the app, which made picture-in-picture implementations enter PiP unexpectedly. ([#48471](https://github.com/expo/expo/pull/48471) by [@stareezy-1](https://github.com/stareezy-1))
 - [iOS] Avoid warning when an aborted push token registration request rejects with a native fetch cancellation error. ([#48547](https://github.com/expo/expo/pull/48547) by [@JoaoPauloCMarra](https://github.com/JoaoPauloCMarra))
 - [Android] Prevent a crash on notification tap when `getLaunchIntentForPackage` throws on some OEM ROMs. ([#47889](https://github.com/expo/expo/pull/47889) by [@nunocaseiro](https://github.com/nunocaseiro))
--- a/apps/bare-expo/ios/Podfile.lock
+++ b/apps/bare-expo/ios/Podfile.lock
@@ -3913,7 +3913,7 @@ SPEC CHECKSUMS:
   ExpoModulesWorklets: 6b240031daa3de4df791588fc78236637ff602f6
   ExpoModulesWorkletsAdapter: add08a40add90b989d4f399e49508e34c6cf0334
   ExpoNetwork: 973d523fe9b1f99223dd945fed9ad1226ae64920
-  ExpoNotifications: 0f253ad90f108c1c4045b541294164865c6aadc8
+  ExpoNotifications: 2cb244d406414754b7159f1c7a9fd2d4754e7f47
   ExpoObserve: aa4fcbe7ed0a7ba387984a38e24c2ecaa8895fd0
   ExpoPrint: 7e7a9bc7c12b9b336f34c4458d0c9ba14f243759
   ExpoRouter: 224087330cc8f5a39700e6f4d7a55f55ecb2f90d
--- a/packages/expo-notifications/ios/ExpoNotifications.podspec
+++ b/packages/expo-notifications/ios/ExpoNotifications.podspec
@@ -31,5 +31,9 @@ Pod::Spec.new do |s|
     test_spec.dependency 'ExpoModulesTestCore'
 
     test_spec.source_files = "Tests/**/*.{m,mm,swift}"
+
+    test_spec.pod_target_xcconfig = {
+      'OTHER_LDFLAGS' => '$(inherited) -lc++'
+    }
   end
 end
--- a/packages/expo-notifications/ios/ExpoNotifications/Notifications/NotificationCenterManager.swift
+++ b/packages/expo-notifications/ios/ExpoNotifications/Notifications/NotificationCenterManager.swift
@@ -39,6 +39,59 @@ public extension NotificationDelegate {
 private let discardPresentationOptions: (UNNotificationPresentationOptions) -> Void = { _ in }
 
 /**
+ The delegates of `NotificationCenterManager`, and the responses that arrived before any delegate
+ could handle them.
+
+ The manager is a process-wide singleton, but app contexts are not: the modules of an incoming
+ context register themselves while an outgoing context tears its own down, which happens on every
+ dev-client reload and on `Updates.reloadAsync()`. Both arrays are therefore written from more than
+ one thread at a time, so they live behind a lock.
+ */
+internal final class NotificationDelegateRegistry {
+  private let state = Mutex(State())
+
+  private struct State {
+    var delegates: [NotificationDelegate] = []
+    var pendingResponses: [UNNotificationResponse] = []
+  }
+
+  /**
+   A snapshot of the delegates. Callers iterate the snapshot instead of holding the lock across a
+   delegate callback: `add` hands back the pending responses so that the caller can offer them to
+   the delegate it just added, and a delegate is free to add or remove delegates from there.
+   */
+  var delegates: [NotificationDelegate] {
+    state.withLock { $0.delegates }
+  }
+
+  var pendingResponses: [UNNotificationResponse] {
+    state.withLock { $0.pendingResponses }
+  }
+
+  /**
+   Adds the delegate, and returns the responses that it still has to be offered.
+   */
+  func add(_ delegate: NotificationDelegate) -> [UNNotificationResponse] {
+    state.withLock { state in
+      state.delegates.append(delegate)
+      return state.pendingResponses
+    }
+  }
+
+  func remove(_ delegate: AnyObject) {
+    state.withLock { $0.delegates.removeAll { $0 === delegate } }
+  }
+
+  func appendPendingResponse(_ response: UNNotificationResponse) {
+    state.withLock { $0.pendingResponses.append(response) }
+  }
+
+  func removeAllPendingResponses() {
+    state.withLock { $0.pendingResponses.removeAll() }
+  }
+}
+
+/**
  Singleton that sets itself as the UserNotificationCenter delegate,
  and calls its own delegates in response to notification center calls.
  The notification center keeps a single delegate, so a delegate that another library set before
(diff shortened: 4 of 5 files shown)


packages/expo-notifications/CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 
 ### 🐛 Bug fixes
 
+- [iOS] Fix a data race on `NotificationCenterManager`'s delegate list that crashed the app with `SIGSEGV` when one app context registered its modules while another tore its own down, such as on a dev-client reload or `Updates.reloadAsync()`. ([#49549](https://github.com/expo/expo/issues/49549), [#49554](https://github.com/expo/expo/pull/49554) by [@dennytosp](https://github.com/dennytosp))
```
vonovak:
```suggestion
- [iOS] Fix a data race on `NotificationCenterManager`'s delegate list that crashed the app with `SIGSEGV` when one app context registered its modules while another tore its own down, such as on a dev-client reload or `Updates.reloadAsync()`. [#49554](https://github.com/expo/expo/pull/49554) by [@dennytosp](https://github.com/dennytosp))
```

vonovak (approved):

PR #23716, opened 2023-07-25
fix(cli): make `NODE_ENV` from build type case insensitive

byCedric:
# Why

Especially on Android, users might use `Release` as well. Moving this to be case insensitive is a saver bet in general.

# How

- Added `?.toLowerCase()` when setting `development` or `production` as fallback `NODE_ENV`.

# Test Plan

- `$ yarn create expo ./test-build-type -t blank@49`
- `$ cd ./test-build-type`
- `$ yarn expo prebuild`
- Change the build types in the native files
  - **android**: `RELEASE` ([docs](https://developer.android.com/build/build-variants))
  - **ios**: `RELEASE` ([docs](https://developer.apple.com/documentation/xcode/adding-a-build-configuration-file-to-your-project))
- Create `app.config.js` with:
  - `console.log('Node environment:', process.env.NODE_ENV);`
  - `module.exports = (config) => config;`
- Try running the "release" version for both platforms
  - `$ yarn expo run:android --variant RELEASE`
  - `$ yarn expo run:ios --configuration RELEASE`
- Console should render `Node environment: production` for both platforms

# Checklist

- [ ] Documentation is up to date to reflect these changes (eg: https://docs.expo.dev and README.md).
- [ ] Conforms with the [Documentation Writing Style Guide](https://github.com/expo/expo/blob/main/guides/Expo%20Documentation%20Writing%20Style%20Guide.md)
- [ ] This diff will work correctly for `npx expo prebuild` & EAS Build (eg: updated a module plugin).

PR #49555, opened 2026-08-31
fix(cli): Use consistent mode for custom Xcode configs

ramonclaudio:
Related to [#23716](https://github.com/expo/expo/pull/23716).

# Why

Custom Xcode configs can make `expo run:ios` load development env files while React Native bundles JavaScript for production.

# How

I updated the mode check to match React Native, so configs with `Debug` in the name use development and everything else uses production.

# Test Plan

All tests pass.


ramonclaudio:
@expo-bot verify with fable

ramonclaudio:
@expo-bot verify

packages/@expo/config-plugins/src/ios/BuildScheme.ts
```diff
@@ -31,9 +34,119 @@ export function getSchemesFromXcodeproj(projectRoot: string): string[] {
...
+  configuration?: string
+): Promise<{ configuration: string; osType: string }> {
+  const schemePaths = findSchemePaths(projectRoot).filter(
+    (file) => path.parse(file).name === scheme
+  );
+  if (schemePaths.length > 1) {
+    throw new Error(`Scheme '${scheme}' is defined in multiple files: ${schemePaths.join(', ')}`);
+  }
+
+  const schemePath = schemePaths[0];
```
Kudo: **High:** This throw is the production counterpart of the duplicate-scheme behavior. With `findSchemePaths` now including `xcuserdata/*.xcuserdatad`, many apps will hit `schemePaths.length > 1` for the default app scheme.

Suggested direction: resolve with precedence `xcshareddata` > current user > other user data (and warn), rather than aborting.

packages/@expo/cli/src/run/ios/options/resolveOptions.ts
```diff
@@ -74,3 +52,25 @@ export async function resolveOptionsAsync(
...
+
+type NativeBuildOptions = Pick<BuildProps, 'xcodeProject' | 'scheme' | 'configuration' | 'osType'>;
+
+export async function resolveNativeBuildOptionsAsync(
+  projectRoot: string,
+  options: Options
+): Promise<NativeBuildOptions> {
+  const xcodeProject = resolveXcodeProject(projectRoot);
+
+  // Resolve the scheme before the device so we can filter devices based on
```
Kudo: **Medium:** Mode resolution uses `configuration.includes('Debug')`, but `SKIP_BUNDLING` still only applies when `configuration === 'Debug'`.

For device builds of `DebugStaging` / similar, env+Metro are development while the native script may still reset Metro’s cache / embed unexpectedly. Consider `configuration.includes('Debug') && !isSimulator` (or explicitly document the intentional divergence).

packages/@expo/cli/src/run/android/resolveBuildModeAsync.ts
```diff
@@ -0,0 +1,78 @@
...
+import path from 'path';
+import { z } from 'zod';
+
+import { AbortCommandError, CommandError } from '../../utils/errors';
+
+const reportPrefix = 'EXPO_ANDROID_BUILD_MODE=';
+const reportSchema = z.object({
+  version: z.literal(1),
+  projectPath: z.literal(':app'),
+  variant: z.string().min(1),
```
Kudo: **Medium:** Zod (and the Gradle task) require `projectPath: ':app'`. That matches Expo prebuild defaults, but bare apps with a differently named application project will always get `ANDROID_BUILD_MODE` failures.

If `:app` is a hard product constraint for now, please call it out in the changelog / error message; otherwise plumb the real app project path.

packages/@expo/config-plugins/src/ios/Paths.ts
```diff
@@ -164,11 +164,14 @@ export function getSourceRoot(projectRoot: string): string {
 
 export function findSchemePaths(projectRoot: string): string[] {
   return withSortedGlobResult(
-    globSync('ios/*.xcodeproj/xcshareddata/xcschemes/*.xcscheme', {
-      absolute: true,
-      cwd: projectRoot,
-      ignore: ignoredPaths,
-    })
+    globSync(
+      'ios/*.{xcodeproj,xcworkspace}/{xcshareddata,xcuserdata/*.xcuserdatad}/xcschemes/*.xcscheme',
```
Kudo: **High (related):** Expanding this glob to `xcuserdata` is what makes duplicate scheme names common. Combined with the hard-fail in `getBuildConfigurationForSchemeAsync`, day-to-day `run:ios` can break after a developer edits a scheme in Xcode.

If user schemes must be readable, prefer shared-over-user when resolving a named scheme instead of treating duplicates as fatal.

packages/@expo/cli/CHANGELOG.md
```diff
@@ -8,6 +8,10 @@
 
 ### 🐛 Bug fixes
 
+- Resolve the iOS build configuration before loading env files in `expo run:ios`. Use the scheme's Run configuration when `--scheme` is passed without `--configuration`. ([#49555](https://github.com/expo/expo/pull/49555) by [@ramonclaudio](https://github.com/ramonclaudio))
```
Kudo: **Medium / changelog:** Please mention that configuration names containing neither case-sensitive `Debug` (e.g. `Staging`, `Development`, `Dev`) now load **production** env + eager bundle, matching `react-native-xcode.sh`. That is a deliberate behavior change and easy to miss from the current bullets.

ramonclaudio:
@expo-bot verify --fable

kitten (approved):

PR #49558, opened 2026-08-31
[ios][task-manager] Guard EXTaskService collections against concurrent access

vonovak:
# Why

`EXTaskService` is a process-wide singleton with no synchronization on its collections (`taskManagers`, `headlessTaskManagers`, `eventsQueues`, `tasks`, `events`, `appRecords`, `requests`). `TaskManagerModule.OnCreate` writes them from the incoming React instance's JS thread while OS background-launch callbacks and the outgoing instance's teardown touch them from other threads — a data race on every reload. Same pattern as #49549 / #49554.

# Reproduction

Standalone harness with the pre-fix method bodies copied verbatim (source below; it runs the extracted code, not the full app):

- Plain build: SIGSEGV in 5 of 5 runs, preceded by objc `__weak variable ... incorrect use of objc_storeWeak()` warnings.
- TSan build: 6 data-race reports, all inside the copied `setTaskManager:forAppId:` and `executeTask:` bodies.
- Control: the same harness with equivalent locking runs 100k iterations clean.

<details>
<summary>repro.m — clang -fobjc-arc -O2 -framework Foundation repro.m -o repro (TSan: add -fsanitize=thread -g -DTSAN_ITERATIONS=2000)</summary>

```objc
// Standalone repro of the EXTaskService data race (pre-fix shape),
// see packages/expo-task-manager/ios/EXTaskManager/EXTaskService.m on `main`.
//
// Faithfully copies the method bodies of:
//   - setTaskManager:forAppId:withUrl:  (~line 251)
//   - executeTask:withData:withError:   (~line 284)
//   - _taskManagerForAppId:             (~line 576)
// with the same ivar shapes:
//   - _taskManagers / _headlessTaskManagers: NSMapTable strongToWeakObjectsMapTable
//   - _eventsQueues / _events: NSMutableDictionary
// and NO locking, exactly as in the pre-fix source.

#import <Foundation/Foundation.h>

static NSString *const kAppId = @"mainApplication";
#ifdef TSAN_ITERATIONS
static const NSUInteger kIterations = TSAN_ITERATIONS;
#else
static const NSUInteger kIterations = 100000;
#endif

// Stub for id<EXTaskManagerInterface>. Plain NSObject is enough: the race
// is about the container bookkeeping in EXTaskService, not about the
// task manager's own behavior.
@interface StubTaskManager : NSObject
@property (nonatomic, assign) BOOL isRunningInHeadlessMode;
@end

@implementation StubTaskManager
- (instancetype)init
{
  if (self = [super init]) {
    _isRunningInHeadlessMode = NO;
  }
  return self;
}
- (void)executeWithBody:(NSDictionary *)body
{
  // no-op stub, mirrors [taskManager executeWithBody:body] callout
}
@end

@interface TaskService : NSObject
{
@public
  NSMapTable<NSString *, StubTaskManager *> *_taskManagers;
  NSMapTable<NSString *, StubTaskManager *> *_headlessTaskManagers;
  NSMutableDictionary<NSString *, NSMutableArray<NSDictionary *> *> *_eventsQueues;
  NSMutableDictionary<NSString *, NSMutableArray<NSString *> *> *_events;
}
@end

@implementation TaskService

- (instancetype)init
{
  if (self = [super init]) {
    _taskManagers = [NSMapTable strongToWeakObjectsMapTable];
    _headlessTaskManagers = [NSMapTable strongToWeakObjectsMapTable];
    _eventsQueues = [NSMutableDictionary new];
    _events = [NSMutableDictionary new];
  }
  return self;
}

// Copied body shape from EXTaskService.m setTaskManager:forAppId:withUrl: (~line 251-280)
- (void)setTaskManager:(StubTaskManager *)taskManager forAppId:(NSString *)appId
{
  // Determine in which table the task manager will be stored.
  // Having two tables for them is to prevent race condition problems,
  // when both foreground and background apps are launching at the same time.
  BOOL isHeadless = taskManager.isRunningInHeadlessMode;
  NSMapTable *taskManagersTable = isHeadless ? _headlessTaskManagers : _taskManagers;

  // Set task manager in appropriate table.
  [taskManagersTable setObject:taskManager forKey:appId];

  // Execute events waiting for the task manager.
  NSMutableArray *appEventQueue = _eventsQueues[appId];

  if (appEventQueue) {
    for (NSDictionary *body in appEventQueue) {
      [taskManager executeWithBody:body];
    }
  }

  // Remove events queue for that app.
  [_eventsQueues removeObjectForKey:appId];
}

// Copied body shape from EXTaskService.m executeTask:withData:withError: (~line 284-320)
- (void)executeTask:(NSString *)appId eventId:(NSString *)eventId
{
  StubTaskManager *taskManager = [self _taskManagerForAppId:appId];
  NSDictionary *body = @{@"eventId": eventId};

  // Save an event so we can keep tracking events for this app
  NSMutableArray *appEvents = _events[appId] ?: [NSMutableArray new];
  [appEvents addObject:eventId];
  [_events setObject:appEvents forKey:appId];

  if (taskManager != nil) {
    // Task manager is initialized and can execute events
    [taskManager executeWithBody:body];
    return;
  }

  // App record for that app exists, but it's not fully loaded as its task manager is not there yet.
  // We need to add event's body to the queue from which events will be executed once the task manager is ready.
  NSMutableArray *appEventsQueue = _eventsQueues[appId] ?: [NSMutableArray new];
  [appEventsQueue addObject:body];
  [_eventsQueues setObject:appEventsQueue forKey:appId];
}

// Copied body from EXTaskService.m _taskManagerForAppId: (~line 576-580)
- (StubTaskManager *)_taskManagerForAppId:(NSString *)appId
{
  StubTaskManager *taskManager = [_taskManagers objectForKey:appId];
  return taskManager ?: [_headlessTaskManagers objectForKey:appId];
}

@end

int main(int argc, const char *argv[])
{
  @autoreleasepool {
    TaskService *service = [TaskService new];
    dispatch_queue_t q = dispatch_get_global_queue(QOS_CLASS_USER_INITIATED, 0);
    dispatch_group_t group = dispatch_group_create();

    // Thread A: create a fresh stub manager, register it, drop the reference.
    // Weak-zeroing in the NSMapTable fires on THIS thread when the manager
    // deallocates, racing with other threads mutating/reading the same tables.
    dispatch_group_async(group, q, ^{
      for (NSUInteger i = 0; i < kIterations; i++) {
        @autoreleasepool {
          StubTaskManager *tm = [StubTaskManager new];
          [service setTaskManager:tm forAppId:kAppId];
          tm = nil; // triggers dealloc -> weak zeroing of the map table entry
        }
      }
    });

    // Thread B: same as A, interleaved with a different manager instance.
    dispatch_group_async(group, q, ^{
      for (NSUInteger i = 0; i < kIterations; i++) {
        @autoreleasepool {
          StubTaskManager *tm = [StubTaskManager new];
          [service setTaskManager:tm forAppId:kAppId];
          tm = nil;
        }
      }
    });

    // Thread C: executeTask loop, mutates _events / _eventsQueues from a
    // simulated OS background callback.
    dispatch_group_async(group, q, ^{
      for (NSUInteger i = 0; i < kIterations; i++) {
        @autoreleasepool {
          NSString *eventId = [NSString stringWithFormat:@"evt-%lu", (unsigned long)i];
          [service executeTask:kAppId eventId:eventId];
        }
      }
    });

    // Thread D: reader loop, reads both map tables.
    dispatch_group_async(group, q, ^{
      for (NSUInteger i = 0; i < kIterations; i++) {
        @autoreleasepool {
          StubTaskManager *tm = [service _taskManagerForAppId:kAppId];
          (void)tm;
        }
      }
    });

    dispatch_group_wait(group, DISPATCH_TIME_FOREVER);
    NSLog(@"repro: completed %lu iterations per thread without crashing", (unsigned long)kIterations);
  }
  return 0;
}
```

</details>

# How

Every collection access runs inside `@synchronized` on one private token object. `@synchronized` is recursive (locked helpers are called from locked sections) and releases on exception unwind, so a throw inside a critical section can't leave the service deadlocked. The lock is never held across a callout: `setTaskManager`, `unregisterAllTasksForAppId`, `notifyTaskWithName`, and `_iterateTasksUsingBlock` snapshot under the lock and call out after. Config helpers only touch `NSUserDefaults` (thread-safe) and stay unguarded.

# Test Plan

- The reproduction above: pre-fix crashes 5/5, locked shape runs clean under TSan.
- No iOS unit-test target exists for this package; every collection access was audited (guarded or justified), and `@synchronized` blocks are scoped, so no unlock can be missed.
- CI compiles the file in bare-expo and Expo Go.

# Checklist

- [x] Documentation is up to date to reflect these changes (no user-facing API change).

https://claude.ai/code/session_019mBt2UtQv2epdTxx9LjyLk


chrfalch (commented):
Thanks for working on this!!

Would it be an idea to use try..finally for the locking?

```objective-c
[_lock lock];

@try {
    // .... access the locked resources
}
@finally {
  [_lock unlock];
}
```

vonovak:
@chrfalch thanks, agreed the unlock should survive an exception. I'd like to use `@synchronized` instead of try/finally, because it solves the same problem with less repetition and removes two other risks at the same time:

1. Same guarantee - `@synchronized` releases the lock on any exit.
2. It replaces the `NSRecursiveLock` too. The explicit lock instance var and its setup go away.
3. Scoped code blocks instead of paired calls. With lock / unlock (or 20 hand-written try/finally blocks), every early return is a chance to miss an unlock.

I'll push it in a new commit.

chrfalch (approved):

PR #49566, opened 2026-08-31
[require-utils] Add default-import interop to the CommonJS fallback

### LINKED ISSUE #49564: [cli] app.config.ts default imports break with typescript@7 — require-utils CJS fallback emits no default-import interop

https://github.com/gustavohariel/expo-ts7-config-interop-repro

### Steps to reproduce

`create-expo-app@latest --template blank-typescript` with one change: `app.config.ts` default-imports a CommonJS module that has a transpiled default export. Package manager is bun, platform-independent (fails before any platform is picked).

```sh
git clone https://github.com/gustavohariel/expo-ts7-config-interop-repro
cd expo-ts7-config-interop-repro
bun install                      # typescript@7.0.2
bunx expo config --type public
```

Expected: the config prints. Actual:

```
TypeError: Error reading Expo config at .../app.config.ts:

_withThing is not a function
    at Object.<anonymous> (.../app.config.js:10:34)
    at compileModule (.../@expo/require-utils/build/load.js:237:9)
    at evalModule (.../@expo/require-utils/build/load.js:361:17)
```

Then `bun add -d typescript@6.0.3` and rerun the same command — it prints the config.

`app.config.ts` and the imported module are the whole reproduction:

```ts
import withThing from "./plugins/with-thing";
const config: ExpoConfig = { name: "expo-ts7-repro", slug: "expo-ts7-repro" };
export default withThing(config);
```

```js
// plugins/with-thing.js
function withThing(config) { return { ...config, extra: { pluginRan: true } }; }
exports.__esModule = true;
exports.default = withThing;
```

typescript@7 ships no compiler API, so `loadTypescript()` returns null (by design, #47759) and `evalModule` takes the `stripTypeScriptTypes` + `toCommonJS` path. That path rewrites the default import with no interop helper, so the binding is the module namespace object rather than `exports.default`:

```js
// toCommonJS output for the app.config.ts above
var _withThing = require("./plugins/with-thing"); // namespace object, not the function
const _default = exports.default = _withThing(config);
```

TS6's `transpileModule` emits `_interopRequireDefault` there, which is why this only appears on the fallback path. Every compiled config plugin has that `exports.default` shape (`expo-router/plugin`, `expo-font/plugin`, and so on), so any `app.config.ts` importing one breaks as soon as typescript@7 is installed — in our own app it is 23 imports and it fails on the first.

Uninstalling typescript altogether reaches the same path, so it also reproduces without an Expo app:

```sh
mkdir t && cd t && npm init -y && npm i @expo/require-utils@57.0.5   # no typescript
printf 'exports.__esModule = true;\nexports.default = (c) => c;\n' > cjs-plugin.js
printf 'import p from "./cjs-plugin";\nexport default p({ ok: true });\n' > my.config.ts
node -e 'require("@expo/require-utils").loadModuleSync(require("path").resolve("my.config.ts"))'
# TypeError: _cjsPlugin is not a function
```

Follow-up to #47627 and #48826: those fixed *entering* this fallback, this is about what the fallback emits.

### Environment

```text
    System:
      OS: Linux 7.1 Omarchy
      Shell: 5.3.15 - /usr/bin/bash
    Binaries:
      Node: 26.7.0 - ~/.local/share/mise/installs/node/26.7.0/bin/node
      npm: 11.19.0 - ~/.local/share/mise/installs/node/26.7.0/bin/npm
    npmPackages:
      expo: ~57.0.18 => 57.0.18
      react: 19.2.3 => 19.2.3
      react-native: 0.86.3 => 0.86.3
    Expo Workflow: managed
```

`@expo/require-utils` resolves to 57.0.5 in the reproduction (no override needed).

### Expo Doctor Diagnostics

```text
Error: node .../node_modules/expo/bin/cli config --json --full exited with non-zero code: 1
```

Doctor cannot complete because it reads the app config through the same loader. With typescript@6.0.3 installed instead, Doctor runs and reports 21/21 checks passed.

gustavohariel:
# Why

An `app.config.ts` that default-imports a config plugin fails to load whenever `typescript` is missing or `typescript@7` is installed. `expo config`, `expo start`, `expo prebuild` and `expo-doctor` all stop with `TypeError: <plugin> is not a function`.

Fixes #49564. Reproduction: https://github.com/gustavohariel/expo-ts7-config-interop-repro

TypeScript 7 ships no compiler API, so `loadTypescript()` returns `null` (#47759) and `evalModule` falls back to `stripTypeScriptTypes` plus `toCommonJS`. `toCommonJS` passed `importInterop: 'node'` to `@babel/plugin-transform-modules-commonjs`, which skips the `__esModule -> default` unwrapping, so this config:

```ts
import withThing from './plugins/with-thing';
export default withThing(config);
```

was emitted as:

```js
var _withThing = require("./plugins/with-thing"); // namespace object, not exports.default
var _default = exports.default = _withThing(config);
```

Every compiled config plugin (`expo-router/plugin`, `expo-font/plugin`, and so on) is `exports.__esModule = true; exports.default = fn`, so `_withThing` is `{ __esModule: true, default: fn }` and calling it throws. The TypeScript path in the same function transpiles with `esModuleInterop: true` and emits `_interopRequireDefault` in that position, so the two transforms disagreed on what a default import means, and which one you got depended on whether `typescript` happened to be installed.

Default imports of real ESM were affected too: Node's `require(esm)` namespace carries `__esModule: true`, and `'node'` interop hands that namespace straight to the binding instead of its `default`.

# How

`toCommonJS` now uses Babel's default `__esModule` interop, which emits `_interopRequireDefault`:

- a module with `exports.__esModule = true` resolves the default import to `exports.default`
- a plain CommonJS module resolves the default import to the whole `module.exports`

`loose: true` is unchanged, and named imports, namespace imports and re-exports emit the same code as before. The only behaviour that changes is the default import of an `__esModule` module, which was the bug.

# Test Plan

Three tests added to `packages/@expo/require-utils/src/__tests__/load-test.ts`, with two new fixtures:

- default import of an `__esModule` module resolves to its default export
- default import of a plain CommonJS module resolves to its `module.exports`
- both of the above through the `stripTypeScriptTypes` fallback, with `typescript` mocked as not installed, which is the reported path

All 4 suites and 95 tests in the package pass. Reverting only `src/transform.ts` and keeping the tests makes the two `__esModule` cases fail with the reported error:

```
● evalModule › resolves a default import of an __esModule module to its default export
  TypeError: _esmodulePlugin is not a function
● evalModule › resolves default imports when falling back to Node TypeScript stripping
  TypeError: _esmodulePlugin is not a function

Tests: 2 failed, 93 passed, 95 total
```

End to end, on the reproduction app (Node 26.7.0, `typescript@7.0.2`, `expo@57.0.18`) with this package built and dropped into its `node_modules`:

```
$ bunx expo config --type public   # before
TypeError: Error reading Expo config at .../app.config.ts:

_withThing is not a function
    at Object.<anonymous> (.../app.config.js:10:34)
    at compileModule (.../@expo/require-utils/build/load.js:237:9)
    at evalModule (.../@expo/require-utils/build/load.js:361:17)

$ bunx expo config --type public   # after
{
  name: 'expo-ts7-repro',
  slug: 'expo-ts7-repro',
  version: '1.0.0',
  description: undefined,
  sdkVersion: '57.0.0',
  platforms: [ 'ios', 'android' ],
  extra: { pluginRan: true }
}
```

`extra.pluginRan` confirms the plugin was called rather than skipped. Downgrading the same app to `typescript@6.0.3`, which takes the `transpileModule` path instead, still prints the same config, so the fix does not regress the primary path.

# Checklist

### DIFF
--- a/packages/@expo/require-utils/src/transform.ts
+++ b/packages/@expo/require-utils/src/transform.ts
@@ -9,13 +9,13 @@ export function toCommonJS(filename: string, code: string) {
       [
         require('@babel/plugin-transform-modules-commonjs'),
         {
-          // NOTE(@kitten): We used to use sucrase to transform, which is why
-          // we're doing this CJS-to-ESM transform in the first place. Our
-          // previous transformation isn't 100% compatible with the standard
-          // Node ESM loading. In Babel, this is the "node" flag (although
-          // node behaviour is explicitly different from this). This skips
-          // the `__esModule -> default` wrapper
-          importInterop: 'node',
+          // NOTE(@gustavohariel): We must keep Babel's default `__esModule` interop here, so that
+          // a default import of a transpiled module resolves to its `exports.default` rather than
+          // to the module namespace object. Every compiled Expo config plugin has that shape, and
+          // Node's `require(esm)` namespace carries `__esModule` too. This matches what
+          // TypeScript's `transpileModule` emits on the other code path in `evalModule`, so both
+          // transforms agree on what a default import means.
+          importInterop: 'babel',
           loose: true,
         },
       ],
--- a/packages/@expo/require-utils/CHANGELOG.md
+++ b/packages/@expo/require-utils/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### 🐛 Bug fixes
 
+- Fixed default imports of `__esModule` CommonJS modules, such as compiled config plugins, resolving to the module namespace object instead of the default export when a TypeScript file is transformed to CommonJS without `typescript` installed ([#49566](https://github.com/expo/expo/pull/49566) by [@gustavohariel](https://github.com/gustavohariel))
+
 ### 💡 Others
 
 - Support Node 26's `stripTypeScriptTypes` and call without transform-mode, to fix the fallback when TypeScript isn't installed ([#48826](https://github.com/expo/expo/pull/48826) by [@kkkhs](https://github.com/kkkhs))
--- a/packages/@expo/require-utils/src/__tests__/fixtures/commonjs-plugin.js
+++ b/packages/@expo/require-utils/src/__tests__/fixtures/commonjs-plugin.js
@@ -1,0 +1,4 @@
+// A hand-written CommonJS module, without an `__esModule` marker
+module.exports = function withPlugin(config) {
+  return { ...config, pluginRan: true };
+};
--- a/packages/@expo/require-utils/src/__tests__/fixtures/esmodule-plugin.js
+++ b/packages/@expo/require-utils/src/__tests__/fixtures/esmodule-plugin.js
@@ -1,0 +1,6 @@
+// A compiled ES module, as every transpiled Expo config plugin looks
+function withPlugin(config) {
+  return { ...config, pluginRan: true };
+}
+exports.__esModule = true;
+exports.default = withPlugin;
--- a/packages/@expo/require-utils/src/__tests__/load-test.ts
+++ b/packages/@expo/require-utils/src/__tests__/load-test.ts
@@ -87,6 +87,70 @@ describe('evalModule', () => {
     expect(mod).toMatchObject({ default: 'test' });
   });
 
+  it('resolves a default import of an __esModule module to its default export', () => {
+    const mod = evalModule(
+      `
+      import withPlugin from './esmodule-plugin.js';
+      export default withPlugin({ name: 'test' });
+    `,
+      path.join(basepath, 'eval.js')
+    );
+
+    expect(mod).toEqual({
+      __esModule: true,
+      default: { name: 'test', pluginRan: true },
+    });
+  });
+
+  it('resolves a default import of a plain CommonJS module to its module.exports', () => {
+    const mod = evalModule(
+      `
+      import withPlugin from './commonjs-plugin.js';
+      export default withPlugin({ name: 'test' });
+    `,
+      path.join(basepath, 'eval.js')
+    );
+
+    expect(mod).toEqual({
+      __esModule: true,
+      default: { name: 'test', pluginRan: true },
+    });
+  });
+
+  it('resolves default imports when falling back to Node TypeScript stripping', () => {
+    jest.isolateModules(() => {
+      const actualNodeModule = jest.requireActual('node:module');
+      const moduleNotFoundError = new Error(
+        "Cannot find module 'typescript'"
+      ) as NodeJS.ErrnoException;
+      moduleNotFoundError.code = 'MODULE_NOT_FOUND';
+
+      jest.doMock('node:module', () => ({
+        ...actualNodeModule,
+        stripTypeScriptTypes: (code: string) => code.replace(': Config', ''),
+      }));
+      jest.doMock('typescript', () => {
+        throw moduleNotFoundError;
+      });
+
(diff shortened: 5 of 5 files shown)


packages/@expo/require-utils/src/transform.ts
```diff
@@ -9,13 +9,13 @@ export function toCommonJS(filename: string, code: string) {
       [
         require('@babel/plugin-transform-modules-commonjs'),
         {
-          // NOTE(@kitten): We used to use sucrase to transform, which is why
-          // we're doing this CJS-to-ESM transform in the first place. Our
-          // previous transformation isn't 100% compatible with the standard
-          // Node ESM loading. In Babel, this is the "node" flag (although
-          // node behaviour is explicitly different from this). This skips
-          // the `__esModule -> default` wrapper
-          importInterop: 'node',
```
kitten, line 18: The comment above is important. As far as I remember this can't be switched to `node` or older patterns break. I'm not sure I'd love to commit to a blanket change here without testing all edge cases.

Happy to commit to the issue on our end, but assuming this is carried over, this would need a lot more testing and some checks on what was preserved

kitten (commented):
side-note to the above: TypeScript 7 isn't supported, so this path is explicitly half-broken, and was only added as requested. We can look into the new transpilation API in 7.1, but it makes more sense to, for now, split your dependencies into TS 6 and TS 7 (the latter via an alias) as per the TypeScript changelog post on the matter, since they decided to ship TS 7.0 as stable with a complete lack of a public-facing compiler API

packages/@expo/require-utils/src/transform.ts
gustavohariel: Sorry — the diff goes the other way: it moves *away* from `'node'` to `'babel'`. The NOTE is right that `'node'` skips the `__esModule -> default` wrapper; that skip is the bug.

On what's preserved, I ran every import form through both settings, comparing emitted code and runtime result (Node 26.7.0, `loose: true`; fixtures: an `__esModule` CJS module, a plain `module.exports = fn`, and a real `.mjs`):

| pattern | `node` | `babel` |
| --- | --- | --- |
| default import, `__esModule` CJS | namespace object | default export |
| default import, real ESM (`require(esm)`) | namespace object | default export |
| default + named mixed | default = object, named same | default = fn, named same |
| default import, plain CJS | same | same |
| namespace import | same | same |
| builtin default (`node:path`) | same | same |
| named import / `export {x} from` / `export * from` / dynamic import / default export | same | same |

So the only behavioural deltas are the three broken default-import cases; the rest round-trips identically (a few emit differently but evaluate the same). The real-ESM row surprised me — Node's `require(esm)` namespace sets `__esModule: true`, so `'node'` interop mis-binds that too.

If you'd rather not touch the shared helper, I'm happy to thread `importInterop` through `toCommonJS` and pass `'babel'` only at the `stripTypeScriptTypes` call site in `evalModule`, leaving `format.mode === 'commonjs'` on `'node'`. Say which you prefer and I'll push it.

One clarification on the side note: this fallback isn't TS7-only — it also runs when `typescript` isn't installed at all, which reproduces the same `TypeError` with no TS7 anywhere. (We have moved our own app to the TS6/TS7 alias split in the meantime, thanks.)

PR #49572, opened 2026-08-31
[cli][server] Cache resolved manifest runtime version per dev-server session

### LINKED ISSUE #46415: CLI: dev server recomputes @expo/fingerprint on every manifest request (runtimeVersion fingerprint policy), causing iOS dev-client connection timeouts

Rakshit-gen:
# Why

Fixes #46415.

`ExpoGoManifestHandlerMiddleware._getManifestResponseAsync` resolves the manifest's `runtimeVersion` on every request by calling `resolveRuntimeVersionWithExpoUpdatesAsync`, which shells out to `expo-updates runtimeversion:resolve` with no caching. Under `runtimeVersion: { policy: 'fingerprint' }`, that subprocess recomputes the full `@expo/fingerprint` on every call — several seconds on a real project, per the issue's repro (~13s; ~4s even on a blank `create-expo-app`).

The dev client fetches the manifest before the bundle, and some launchers (iOS) have a shorter connection timeout (~10s) than that computation, so the client times out with `error loading app, failed to connect to http://<host>:8081` while the manifest request is still resolving server-side — Android tolerates the delay better, so this looks like "Metro only works on Android" with no obvious cause. It's also a regression trap: anything that cold-invalidates the fingerprint (`prebuild --clean`, a clean reinstall, adding a native module) can start breaking iOS connections with no corresponding change in the project's own code.

Two external reporters (issue author + a commenter) independently hit this and converged on the same fix shape: cache the resolved runtime version for the dev-server session instead of recomputing it per request.

# How

Added a `Map<RuntimePlatform, Promise<string | null>>` cache on `ExpoGoManifestHandlerMiddleware` (this middleware is instantiated once per dev-server start in `BundlerDevServer.getManifestMiddlewareAsync`, so the instance already lives for the session, same as other per-session state in this file). `resolveCachedRuntimeVersionAsync` memoizes the existing resolution logic (`resolveRuntimeVersionWithExpoUpdatesAsync` falling back to `Updates.getRuntimeVersionAsync`) per platform, storing the in-flight promise immediately so concurrent requests for the same platform share one resolution instead of triggering redundant subprocess calls.

A rejected promise is removed from the cache (`.catch(() => cache.delete(platform))`) rather than cached, so a transient failure doesn't permanently break manifest requests for the rest of the session — only a successfully resolved (or `null`) result is kept, consistent with how a similar cache was scoped in #49259 in this repo.

Like other config-derived values already fixed for a session (`baseUrl`, `reactCompiler`), this requires a dev-server restart to pick up a runtime-version change mid-session — consistent with existing behavior elsewhere in the CLI.

# Test Plan

Added two tests to `ExpoGoManifestHandlerMiddleware-test.ts`:
- `caches the resolved runtime version across requests for the same platform` — calls `_getManifestResponseAsync` twice for the same platform and asserts `resolveRuntimeVersionWithExpoUpdatesAsync` was only invoked once.
- `resolves the runtime version separately per platform` — confirms `ios` and `android` get independent resolutions.

Confirmed both fail against the pre-fix code (2 calls instead of 1) and pass after the fix. All existing tests in the file (19 total) still pass.

# Checklist

### DIFF
--- a/packages/@expo/cli/CHANGELOG.md
+++ b/packages/@expo/cli/CHANGELOG.md
@@ -24,6 +24,7 @@
 
 ### 🐛 Bug fixes
 
+- Cache the resolved manifest runtime version per platform for the dev-server session, instead of recomputing it (including a full `@expo/fingerprint` computation under `runtimeVersion: { policy: 'fingerprint' }`) on every manifest request, which could take longer than a dev client's connection timeout and surface as "failed to connect".
 - Stop writing DOM component source maps into the app binary during `expo export:embed`, so Android release builds no longer ship `www.bundle` source maps with the original app source. ([#49480](https://github.com/expo/expo/pull/49480) by [@expo-bot](https://github.com/expo-bot))
 - Serve relative manifest URLs only when the client itself sends the RFC 7239 `Forwarded` header, so that proxied requests from clients without relative-URL support, like released Expo Go versions through the WS tunnel, keep absolute URLs. ([#48997](https://github.com/expo/expo/pull/48997) by [@expo-bot](https://github.com/expo-bot))
 - Fail when `--private-key-path` is passed without `updates.codeSigningCertificate` in the resolved app config, instead of ignoring the flag and continuing without signing.
--- a/packages/@expo/cli/src/start/server/middleware/ExpoGoManifestHandlerMiddleware.ts
+++ b/packages/@expo/cli/src/start/server/middleware/ExpoGoManifestHandlerMiddleware.ts
@@ -1,5 +1,5 @@
 import { events } from '2g';
-import type { ExpoUpdatesManifest } from '@expo/config';
+import type { ExpoConfig, ExpoUpdatesManifest } from '@expo/config';
 import { Updates } from '@expo/config-plugins';
 import accepts from 'accepts';
 import crypto from 'crypto';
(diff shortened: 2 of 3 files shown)


Rakshit-gen:
@kitten tagging you since you own most of the recent work in this middleware (e.g. #47255, #48267) — would appreciate a look when you have time.

kitten:
Please comment on the issue first or add a note if you've seen the other two PRs in the linked issues with their reasons of having been closed.

There's no indication here either whether you've run into this yourself or speculatively filed this, but assuming it's basically identical to the other two PRs and it'd be better to come to a conclusion in the issue first, since it's pointless to let the close reason conflict with the move to fix an issue that has a root cause that needs to be addressed instead (in this case a deprecation, a config change, or something of the sort)

PR #46285, opened 2026-05-26
[ios][ui] [ui] add `presentationBackground` SwiftUI modifier and apply it in community/bottom-sheet

duyanhv:
community/bottom-sheet

# Why
`@expo/ui/swift-ui/modifiers` exposes `presentationDetents`, `presentationDragIndicator`, `presentationBackgroundInteraction`, and `interactiveDismissDisabled` but not [`presentationBackground(_:)`](https://developer.apple.com/documentation/swiftui/view/presentationbackground(_:)). It's the only SwiftUI modifier that paints the entire sheet chrome, including the drag-indicator zone and the home-indicator safe-area inset.

Without it, an iOS sheet's chrome retains the system background even if the React content is fully colored producing a visible band at the top and bottom that mismatches custom-themed app surfaces. `community/bottom-sheet`'s `backgroundStyle.backgroundColor` is honored on Android (extracted as Material 3's `containerColor`) but ignored on iOS for this reason.

This PR adds first-class support for `presentationBackground` and uses it in `community/bottom-sheet` so iOS matches the Android behavior — `backgroundStyle.backgroundColor` themes the entire sheet on both platforms.

# How

## iOS (SwiftUI)

- Added `PresentationBackgroundModifier` as a `ViewModifier` in `ios/Modifiers/PresentationModifiers.swift`
- Guarded with `#available(iOS 16.4, tvOS 16.4, *)` gracefully falls back to the system background on older versions
- Registered the modifier in `ViewModifierRegistry` under `"presentationBackground"`

## JavaScript / TypeScript

- Added `presentationBackground(color)` to `@expo/ui/swift-ui/modifiers`, exported alongside the other presentation modifiers
- Updated the `backgroundStyle` JSDoc on `BottomSheetProps` to describe the new iOS behavior

## `community/bottom-sheet` (iOS)

- Extracts `backgroundStyle.backgroundColor` and applies it via the new modifier uniform with how Android extracts it as `containerColor`
- No new prop introduced; existing `backgroundStyle` API works on iOS now

# Test Plan

1. iPhone 15 Simulator (iOS 17.5) and iPhone 12 Simulator (iOS 16.3)
2. Open a sheet with `backgroundStyle={{ backgroundColor: '#FF0000' }}`
3. Before: drag-indicator zone (~16-30pt at the top) and home-indicator inset (~34pt at the bottom) show the system background
4. After (iOS 16.4+): entire sheet — chrome included — is painted with the configured color
[...]

PR #49575, opened 2026-08-31
[universal] Add containerColor/contentColor to BottomSheet and colors to ListItem

timheilman:
Adds `containerColor`/`contentColor` to the universal `BottomSheet`, and a `colors` prop to the universal `ListItem`, so a sheet's chrome and its rows can be recolored to match an app's own theme instead of always sitting on each platform's opaque default (white on web/Material light, near-black on Material dark, a fixed system tint on iOS).

This is pure plumbing — every new prop forwards to a capability the platform-specific layer already exposes, and omitting them keeps every platform's current default behavior unchanged:
- `jetpack-compose`'s `ModalBottomSheet` and `ListItem` already accept `containerColor`/`contentColor`/`colors` — the universal layer just didn't forward them.
- `swift-ui`'s `presentationBackground` modifier (#46285) already does exactly what `containerColor` needs on iOS; this wires it in automatically instead of requiring every caller to reach for `modifiers` themselves.
- On web, `containerColor` overrides the hardcoded `white`/`isDark` background in `vaul`'s `Drawer.Content` style.

There's an [open Canny feature request](https://expo.canny.io/feature-requests/p/expo-ui-expose-option-containercolor-for-bottomsheet-in-universal-implementation) for this exact capability, and similar code was suggested in a comment on #46285. It's the same shape of change as #48943 (universal `contentPadding`), and forwards to the already-fixed `contentColorFor()` default logic from #49394, so it doesn't reintroduce that bug.

## Test plan

- Added a unit test (`resolveStringColor.test.ts`) for the one piece of real logic in this diff — narrowing `ColorValue` to a plain string, since `presentationBackground` and a CSS `backgroundColor` can't consume `PlatformColor()`/`DynamicColorIOS()`. Everything else is direct prop-forwarding to an already-existing platform capability. `pnpm test`, `pnpm typecheck`, `pnpm lint`, and `pnpm format` all pass in `packages/expo-ui`.
- Verified manually on **web**: `containerColor`/`contentColor` render correctly on `BottomSheet`, and (correctly, since it's documented Android-only) `colors` is a no-op on `ListItem`.
- Verified manually on **Android** and **iOS**: this exact diff (byte-for-byte on Android; a behavior-preserving refactor on iOS/web that extracts the same `ColorValue`-narrowing check into a shared, unit-tested helper) has been running as a local patch in a production React Native app, confirmed rendering the correct custom `containerColor`/`contentColor` on a physical/simulated Android emulator, iOS simulator, and web build.


intergalacticspacehighway:
@expo-bot review

packages/expo-ui/src/universal/BottomSheet/index.ios.tsx
```diff
@@ -41,6 +43,10 @@ export function BottomSheet({
   if (snapPoints && snapPoints.length > 0) {
     presentationModifiers.push(presentationDetents(snapPoints.map(snapPointToDetent)));
   }
+  const resolvedContainerColor = resolveStringColor(containerColor);
+  if (resolvedContainerColor) {
+    presentationModifiers.push(presentationBackground(resolvedContainerColor));
```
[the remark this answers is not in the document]
timheilman: Good catch on the platform inconsistency, but the underlying claim about Android needs a correction based on manual testing, not just code reading.

I tested the exact failure path (`<BottomSheet isPresented containerColor="#16332b"><Text>Hello</Text></BottomSheet>`) by hand across a real Android emulator, iOS simulator, and web build. Android does **not** actually give readable text either — I got black-on-dark-green on all three platforms, not just iOS/web.

The reason: `contentColorFor()`'s luminance-derived default only reaches native Compose content that reads the ambient `LocalContentColor` (e.g. `@expo/ui/jetpack-compose`'s own `Text`). A universal `BottomSheet`'s `children` are React Native views hosted via the interop bridge — RN's own `<Text>` sets its own `color` independently of Compose's ambient color system, so it never picks up that derived default on Android either.

So this isn't an iOS/web gap to close by mimicking Android — it's a limitation on all three platforms for the common case (RN children), which I've now documented explicitly on both `containerColor` and `contentColor` in 1891c9b (rather than attempting a `colorScheme`/`foregroundStyle` push on iOS, which would change more than just text color and isn't really "fixing" anything Android already does for this API's actual usage pattern).

id:a7a5886bf404

intergalacticspacehighway: `presentationBackground` should work with any `ColorValue` from react-native, including dynamic color/Platform color. We just need to allow passing `ColorValue` type [here](https://github.com/expo/expo/blob/7a999f2bf183c36c75668383b7e97cf6445e36ff/packages/expo-ui/src/swift-ui/modifiers/presentationModifiers.ts#L96).  We can then remove `resolveStringColor` as it won't be needed.

packages/expo-ui/src/universal/BottomSheet/types.ts
```diff
@@ -110,4 +110,35 @@ export interface BottomSheetProps {
...
+   * `Drawer.Content` on web, the system sheet background on iOS).
+   *
+   * This only paints the background -- on every platform, `children` are
+   * plain React Native views, not native Compose/SwiftUI content, so none of
+   * them pick up a contrasting text color automatically from a custom
+   * `containerColor` (Android's own `contentColor` default, derived from
+   * `containerColor`, only reaches native Compose content that reads the
+   * ambient `LocalContentColor`; a React Native `Text` sets its own `color`
+   * independently of that). A dark custom `containerColor` needs the sheet's
+   * own children to set a contrasting color explicitly, on every platform.
```
intergalacticspacehighway: Comments feel too big (AI makes it too verbose  😅). can we shorten them please. Thanks!

intergalacticspacehighway (changes requested):

packages/expo-ui/src/universal/BottomSheet/index.ios.tsx
timheilman: Good call — done in 184847c. Widened `presentationBackground` to `color: Color` (matching `background`/`foregroundColor`/`tint`/`listRowBackground`, which already forward straight to the native `Color?` field with no runtime narrowing), so `BottomSheet`'s iOS path now passes `containerColor` straight through and dropped its `resolveStringColor` call. Kept `resolveStringColor` for the web path only, since a literal CSS string genuinely is required there.

packages/expo-ui/src/universal/BottomSheet/types.ts
timheilman: Shortened in 184847c — thanks for flagging, agreed it was way too much for what it's saying.

packages/expo-ui/src/universal/BottomSheet/index.tsx
```diff
@@ -45,6 +46,7 @@ export function BottomSheet({
   });
   const vaulSnapPoints = snapPoints?.length ? snapPoints.map(snapPointToVaul) : undefined;
   const hasSnapPoints = vaulSnapPoints != null;
+  const resolvedContainerColor = resolveStringColor(containerColor);
```
intergalacticspacehighway: lets remove `resolveStringColor` and simply do `typeof` string check here as the method is not used anywhere else apart from the web sheet?

intergalacticspacehighway (changes requested):

packages/expo-ui/src/universal/BottomSheet/index.tsx
timheilman: derp derp derp.  AI is such a blessing and a curse.  done in 3b54f42

intergalacticspacehighway (approved):
Thanks!

PR #49580, opened 2026-09-01
[ios][autolinking] Read prebuilt XCFramework Info.plists in a single tar pass

chrfalch:
# Why

`read_xcframework_info_plists_from_tarball` checks wether a prebuilt artifact carry a slice for the platform we are building?

The problem with it is that it untars twice to find the answer:

It runs `tar tzf` to list the archive, selects the `*.xcframework/Info.plist` entries, then runs `tar xOzf` once per entry. `tar` decompresses a gzip stream from the start, so that is **1 + N full decompressions** of an archive that also carries the dSYM bundles.

This happens for every prebuilt pod during Podfile evaluation, on every `pod install`.

This came out of review on #49148, where the per-pod archive cost at install time was raised as a concern. This half of the cost predates that PR.

# How

Extract the whole glob in one pass into a temporary directory, then read the plists off disk.

Every match is still read. That matters: a tarball can hold more than one xcframework — the product plus a bundled SPM dependency, such as SDWebImage inside `ExpoImage` — and the caller requires **all** of them to support the target:

```ruby
info_plists.any? && info_plists.all? { |p| info_plist_supports_target_platform?(p) }
```

So stopping at the first match (bsdtar's `-q`) would be wrong here, and is deliberately not used.

The fail-closed contract is kept and tightened. An unreadable archive still yields `[]` so the pod falls back to source, and the two distinct warnings are preserved — "No XCFramework Info.plist found" when the archive simply has none, "Failed to inspect" when `tar` itself fails.

Three things changed after review:

- **A partial extraction is now a failure.** `tar` extracts in order, so a truncated archive yields the first plist and then fails. Reading the exit status only when zero plists came out returned a partial list silently, which would have let the all-xcframeworks check pass on a subset.
- **An unparsable plist now fails closed too.** It was previously dropped by `filter_map`, so the same subset hole applied. Verified: a tarball with one good and one unparsable plist reported the target as supported. This is a deliberate behaviour change from the original implementation, which also dropped it.
- **`/usr/bin/tar` is pinned.** The single-pass read relies on bsdtar's member globbing; a GNU tar earlier in `PATH` would match nothing and send every prebuilt pod to source.

This reduces the time per pod from 32.5 ms to 17.4 ms.

# Test Plan

`et check-packages expo-modules-autolinking` passes.

**Equivalence and timing over all 54 prebuilt tarballs in `packages/precompile/.build`**, comparing the old implementation against the new one in the same process:

```
tarballs: 54
results differ on: 0  (expect 0)
  old: list + N extracts       32.5 ms/tarball
  new: single pass             17.4 ms/tarball
```

Identical parsed output for every artifact, and roughly **1.9×** faster. (An earlier revision of this description said 48 tarballs; that count missed the scoped `@expo` packages, which nest a level deeper.)

**Failure handling**, covering the two holes found in review:

```
PASS  truncated archive -> [] (fail closed)
PASS  truncated archive -> "Failed to inspect"
PASS  truncated archive does NOT pass the platform check
PASS  garbage archive -> reports the tar failure, not "no plists"
PASS  valid archive without plists -> "No XCFramework Info.plist found"
PASS  corrupt plist among several -> [] (previously: a subset passed)
PASS  corrupt tarball no longer passes the platform check
```

**Multi-xcframework semantics**, the risk this change has to avoid. A synthetic tarball holding two xcframeworks, where the product supports iOS and the bundled dependency supports only macOS:

```
plists read: 2  (expect 2 — both must be read)
ids: ["ios-arm64", "macos-arm64"]
supports ios? false  (expect false — BundledDep is macos-only)
real artifact supports ios? true  (expect true)
missing tarball: []  (expect [])
```

If the change had silently read only the first plist, the third line would have been `true` and the tarball would have been wrongly accepted.

# Checklist


chrfalch:
@expo-bot review

chrfalch:
id:88b2a4800557

Confirmed and fixed in 75bda646531 — this was a real regression.

Reproduced both halves first:

```
no-match on a valid archive → exit 1, "Not found in archive",  0 files
truncated archive           → exit 1, "Truncated input file",  1 file extracted
```

The second is the bug. `tar` extracts members in order, so a truncated archive yields the first plist and then fails. The old code returned that partial list without ever consulting `status`, so `all?` ran over a subset — and for a tarball carrying a bundled SPM dependency, that dependency's plist would never have been checked.

Any non-zero exit with a partial result is now treated as unusable:

```ruby
if !status.success? && !plists.empty?
  Pod::UI.warn "[Expo-precompiled] Failed to inspect #{File.basename(tarball)}: #{output.strip}"
  next []
end

if plists.empty?
  Pod::UI.warn "[Expo-precompiled] No XCFramework Info.plist found in #{File.basename(tarball)}"
  next []
end
```

That resolves your second point as well: the "No XCFramework Info.plist found" message is no longer gated on `status.success?`, so a no-match reaches it instead of printing raw `tar` output.

Verified — a truncated archive returns `[]`, warns, and fails the platform check; a garbage archive returns `[]`; multi-xcframework tarballs still read both plists and still reject an iOS target when a bundled dependency is macOS-only. Equivalence and timing are unchanged: 0 differences across all 48 tarballs in `packages/precompile/.build`, 24.8 ms → 13.6 ms each.

Not addressed here: you note the truncated tarball is reachable because `createProductTarballAsync` writes straight to the final path with no atomic rename. That's the better place to close it, but it belongs in the prebuild pipeline rather than this method, so I've left it for a separate change.

packages/expo-modules-autolinking/CHANGELOG.md
```diff
@@ -27,6 +27,8 @@
 
 ### 💡 Others
 
+- [iOS] Read the XCFramework `Info.plist` files out of a prebuilt tarball in a single `tar` pass instead of a listing pass plus one extract per plist, roughly halving the per-pod archive work during `pod install`. ([#49580](https://github.com/expo/expo/pull/49580) by [@chrfalch](https://github.com/chrfalch))
+
```
tsapeta:
```suggestion
- [iOS] Read the XCFramework `Info.plist` files out of a prebuilt tarball in a single `tar` pass instead of a listing pass plus one extract per plist, roughly halving the per-pod archive work during `pod install`. ([#49580](https://github.com/expo/expo/pull/49580) by [@chrfalch](https://github.com/chrfalch))
```

tsapeta (approved):

